PR #139, opened 2026-05-07
fix(controller): emit env vars from Secret env-mappings annotation

matoushavlena:
## Summary

The controller's `credentialEnvVars` only hardcoded env vars for two cases — `secret-type=anthropic` and `secret-type=connection` with a GitHub host. Every other Secret carrying an `agent-platform.ai/env-mappings` annotation (e.g. a generic GitHub PAT mapped to `GH_TOKEN`) had its mapping silently dropped, so the agent pod never saw the env var. Harnesses that gate on env presence then refused to dispatch.

- Parse `agent-platform.ai/env-mappings` on every Secret and emit each `(envName, placeholder)` pair as a pod env var.
- Per-type fallbacks (anthropic → `ANTHROPIC_API_KEY` / `CLAUDE_CODE_OAUTH_TOKEN`; connection + github host → `GH_TOKEN`) remain for legacy Secrets without the annotation; a `seen` set dedupes overlap.
- Malformed annotation JSON is logged via `slog.Warn` instead of swallowed, so a bad annotation is findable in controller logs instead of silently producing a pod missing env vars.
- `fork_resources.go` inherits the fix transparently (same `credentialEnvVars` function).

## Test plan

- [x] `go test ./pkg/reconciler/ -run TestCredentialEnvVars` — three new tests cover the happy path, dedup across annotation + per-type hardcoded path, and malformed-JSON fallback
- [x] `mise run check` — lint, vet, build, helm, kubeconform all clean
- [ ] Manual: rebuild controller image (`mise run cluster:install`), grant a generic GitHub PAT Secret with `envMappings: [{envName: "GH_TOKEN", ...}]` to an instance, restart the agent pod, confirm `env | grep GH_TOKEN` shows `dummy-placeholder` and `git clone` of a private repo through the gateway succeeds

### DIFF
--- a/packages/controller/pkg/reconciler/envoy.go
+++ b/packages/controller/pkg/reconciler/envoy.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"sort"
 	"strings"
@@ -36,6 +37,7 @@ const (
 	envoyHostPatternAnn   = "agent-platform.ai/host-pattern"
 	envoyHeaderNameAnn    = "agent-platform.ai/injection-header-name"
 	envoyAuthModeAnn      = "agent-platform.ai/auth-mode"
+	envoyEnvMappingsAnn   = "agent-platform.ai/env-mappings"
 	// Per-agent grant annotations stamped by the api-server on the
 	// instance ConfigMap. The controller reads them on every reconcile
 	// and intersects with the owner's credential Secret list. Both lists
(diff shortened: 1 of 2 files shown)


xjacka:
🛡️ **Humr** — Code Review @ \`ebb3abc\`

## PR #139: fix(controller): emit env vars from Secret env-mappings annotation
**Author:** matoushavlena | **Branch:** fix/agent-env-mappings-from-secret → main | **Changes:** +86 −5 (2 files)

### Summary
Fixes a bug in the controller's `credentialEnvVars` function where the `agent-platform.ai/env-mappings` Secret annotation was being ignored — meaning generic Secrets with user-declared env var mappings stamped by the api-server never produced env vars in agent pods. The fix reads the JSON annotation, emits placeholder env vars from it, deduplicates against the legacy type-switch fallbacks, and logs a warning on malformed annotation JSON rather than crashing the reconcile loop.

### Findings
- 🟢 **Suggestion:** The `seen` dedup map silently drops duplicate `envName` entries — whether from duplicate keys within one Secret's annotation array, or cross-Secret collisions. A `slog.Debug` call on silent drop would help operators trace "why isn't my env var appearing" without digging into reconciler logs. Low priority since the api-server controls the annotation format and shouldn't emit duplicates. (`packages/controller/pkg/reconciler/envoy.go:184-189`)
- ✅ **Looks good:** Malformed annotation JSON logs a structured warning via `slog.Warn` (with namespace/secret/error) and falls through to the legacy type-switch fallbacks — reconcile loop stability preserved and verified by `TestCredentialEnvVars_MalformedAnnotationFallsBackCleanly`.
- ✅ **Looks good:** Three focused tests cover the critical paths: basic annotation emission for generic Secrets, cross-path dedup between annotation and hardcoded type-switch, and malformed-JSON fallback.

### Verdict
APPROVE — Well-scoped bug fix with clear defensive error handling and solid test coverage; the only minor gap is a missing diagnostic log for silently-deduped env names.

---

pilartomas (approved):

PR #155, opened 2026-05-11
revert(helm): drop explicit-null probe rendering (#154)

pilartomas:
## Summary
Revert #154. Rendering `readinessProbe: null` / `livenessProbe: null` doesn't get translated by Helm's three-way merge into a field deletion — the patch reaches K8s as `livenessProbe: {}` and validation rejects it (`must specify a handler type`). Omitting the field entirely (the #153 behavior) does work.

## Repro
Minimal chart with a Deployment whose container has `livenessProbe` and `readinessProbe`. `helm install`, then upgrade to a chart that:

- renders `livenessProbe: null` → `UPGRADE FAILED: Deployment ... is invalid: ... must specify a handler type` (matches the prod symptom).
- omits the probe fields → upgrade succeeds, probes deleted from the deployment.

## If your cluster is already wedged
If a prior upgrade left the deployment with the empty-probe state, repair it manually before re-running helm:
```
kubectl patch deploy dam-platform-keycloak -n <ns> --type=strategic \
  -p '{"spec":{"template":{"spec":{"containers":[{"name":"keycloak","livenessProbe":null,"readinessProbe":null}]}}}}'
```

## Test plan
- [x] `mise run helm:check:lint`
- [x] `mise run helm:check:render` (kubeconform clean)
- [x] `helm template . --set probes.enabled=false | grep -cE 'readinessProbe|livenessProbe'` → 0
- [x] `helm template . | grep -cE 'readinessProbe|livenessProbe'` → 10
- [x] End-to-end repro against local k3s — `helm install` (probes) → `helm upgrade` (probe fields omitted) → succeeds, probes removed from the live deployment

### DIFF
--- a/deploy/helm/platform/templates/apiserver/app.yaml
+++ b/deploy/helm/platform/templates/apiserver/app.yaml
@@ -228,12 +228,6 @@ spec:
               port: http
             initialDelaySeconds: 10
             periodSeconds: 30
-          {{- else }}
-          # Explicit null (not omission) so a `helm upgrade` from a release
-          # that previously had probes generates a delete patch; otherwise
-          # K8s rejects the upgrade with "must specify a handler type".
-          readinessProbe: null
-          livenessProbe: null
           {{- end }}
           resources:
             {{- toYaml .Values.apiServer.resources | nindent 12 }}
--- a/deploy/helm/platform/templates/keycloak/app.yaml
+++ b/deploy/helm/platform/templates/keycloak/app.yaml
@@ -109,9 +109,6 @@ spec:
               port: 9000
             initialDelaySeconds: 60
             periodSeconds: 30
-          {{- else }}
-          readinessProbe: null
-          livenessProbe: null
           {{- end }}
           resources:
             {{- toYaml .Values.keycloak.resources | nindent 12 }}
--- a/deploy/helm/platform/templates/postgres.yaml
+++ b/deploy/helm/platform/templates/postgres.yaml
@@ -101,9 +101,6 @@ spec:
             initialDelaySeconds: 15
             periodSeconds: 15
             timeoutSeconds: 3
-          {{- else }}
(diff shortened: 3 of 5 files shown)


xjacka:
🛡️ **Humr** — Code Review @ `00992b0`

## PR #155: revert(helm): drop explicit-null probe rendering (#154)
**Author:** pilartomas | **Branch:** fix/revert-probe-null → main | **Changes:** +0 −18 (5 files)

### Summary
Reverts the explicit-null probe rendering introduced in PR #154 across all five Helm templates (apiserver, keycloak, postgres, redis, ui). The deleted code rendered `readinessProbe: null` and `livenessProbe: null` explicitly when probes are disabled, which was intended to generate a delete patch during `helm upgrade`.

### Findings
- 🟡 **Warning:** Reverting explicit-null probe rendering reintroduces the upgrade regression it was written to prevent. When upgrading from a release that had probes enabled to one that disables them, Kubernetes may reject the change with "must specify a handler type" because the probes exist in the live manifest but are absent (rather than null) in the new template. (`deploy/helm/platform/templates/apiserver/app.yaml`, `keycloak/app.yaml`, `postgres.yaml`, `redis.yaml`, `ui/app.yaml`)
- 🟢 **Suggestion:** If the explicit-null approach caused a different problem (e.g., Helm rendering `null` as a literal string or a validator rejecting null-valued fields), document the tradeoff in a comment or PR description so future contributors understand why the safer workaround was dropped.
- ✅ **Looks good:** The change is small, focused, and mechanically correct — the `{{- else }}` blocks and their contents are cleanly removed without disrupting surrounding template logic.

### Verdict
COMMENT — The revert is syntactically clean, but the removed code guarded against a known Kubernetes upgrade-path bug; the PR should document why that risk is now acceptable or how it's being mitigated.

---

PR #172, opened 2026-05-12
Feat: local project import

### LINKED ISSUE #73: Import local project context into agent workspace
## Problem

When a developer moves from running Claude Code locally to running it in DAM, they lose all their accumulated project context - CLAUDE.md, .claude/memory/ files, .claude/settings.json, custom skills. There's no way to bring that context into the agent's workspace. The developer has to start from scratch or manually copy files into the pod.
                   
For the May 18 demo, we want to show a seamless "local -> cloud" migration: take your existing local project context and import it into a DAM agent instance so the agent picks up right where you left off.

## Scope
 
- One-time import of local project context files into an agent instance's /workspace
- Files to support: CLAUDE.md, .claude/ directory (memory, settings, skills)
- Should work from the CLI (dam import <instance> ./path)
- Not a sync - this is a one-shot upload, not an ongoing bidirectional sync

## Open questions

- Should this also support uploading arbitrary files/directories, or just the known Claude Code context files?

jjeliga:
Adds a one-shot, atomic file-import primitive that lets users bring local context (CLAUDE.md, .claude/, custom skills, source folders) into a Platform agent's /home/agent at creation time or any time later. Same multipart contract for browser and future CLI — see DRAFT-file-import.md

# What ships

## Server primitive (agent-runtime)

POST /api/import — stream-extract a .tar / .tar.gz to a per-instance staging dir on the PVC, then atomically swap it into /home/agent under mode=replace or mode=merge.
POST /api/import-preflight — returns top-level conflicts so clients can drive an OS-style Replace / Merge / Cancel UX without uploading first.
Path safety: symlinks, absolute paths, .., and platform-reserved segments (.triggers, .initialized) are refused; bundle is rejected on first violation, staging dir cleaned up.
Single-flight: concurrent imports on the same instance return 409.
Inactivity timeout (30s) bounds stuck uploads without capping total size.
Boot-time sweeper reaps stale .import-staging-* dirs older than 1h (recovery from extract-vs-finalize crashes).

## api-server proxy

POST /api/instances/:id/import and /import-preflight — ownership-checked, then proxied verbatim to the agent-runtime; body is streamed (no buffering).
Wakes hibernated pods through the standard reachability primitive before forwarding.

## UI

New "Import local context (optional)" section in the Add Agent dialog — pick a folder, pick files, or drag-and-drop (mixed folders + files in one go). Bundle is built browser-side (USTAR + CompressionStream("gzip")); a single .tar / .tar.gz / .tgz passes through verbatim without re-wrapping.
Files-panel gains a "Drop folder" path that uses preflight to surface Replace / Merge conflict modal.
Ergonomic client-side filters: node_modules/, .venv/, __pycache__/, .DS_Store, plus the server-reserved segments stripped before upload.

## Bug fixes folded in
Auth: postBundle / importPreflight now go through authFetch — they were calling plain fetch with credentials: "include", which the Bearer-only middleware on /api/* rejected with 401. Manifested as "Error: unauthorized" on running-agent imports and silent failures during agent creation.
Create-flow race: import now runs before setAgentAccess / setAgentConnections. Those mutate the instance ConfigMap's grant annotations, which the controller applies by re-rendering pod env (i.e. deleting and recreating the pod). Importing after them raced the pod swap and surfaced as instance unreachable. The PVC outlives the pod swap, so importing first leaves the files in place.
UI latency: agents/instances list queries invalidate immediately after instances.create, so the agent tile appears right away instead of after the 30–60s post-create flow (import + grants).

## Known limitations
100-byte tar name limit on the browser-side packer (no USTAR prefix or PAX yet). Deep paths get rejected loudly client-side rather than truncated. A future CLI using Node's tar package would not hit this — it uses PAX automatically.
Import is one-shot and bundled; symlinks, ACLs, and long-path extensions would be a versioned successor, not an inline change.

## Related issues:
Closes #73


xjacka:
🛡️ **Humr** — Code Review @ \`6b69ede\`

## PR #172: Feat: local project import
**Author:** jjeliga | **Branch:** feat/73-local-project-import → main | **Changes:** +1508 −115 (22 files)

### Summary
Implements full-stack local project import: users can drag-and-drop or pick folders/files during agent creation or from the files panel, and the bundle is uploaded as a tar.gz to the new instance's \`/home/agent\`. The agent-runtime gains a streaming multipart handler (\`busboy\`), path-safe extraction with traversal/symlink/reserved-segment rejection, atomic-ish replace/merge finalize logic, preflight conflict detection, and a crash-recovery sweeper. The UI builds a USTAR tar.gz in the browser without an npm library and adds a conflict modal for the files-panel flow.

### Findings
- 🟡 **Warning:** \`STAGING_PREFIX = ".import-staging-"\` is defined independently in both \`http.ts\` and \`sweeper.ts\`. If one drifts the sweeper silently stops reclaiming crash-leftover staging dirs — the exact failure mode the sweeper exists to prevent. Extract to a shared constant in a co-located \`constants.ts\`. (\`packages/agent-runtime/src/modules/import/http.ts\`, \`sweeper.ts\`)
- 🟡 **Warning:** No bundle size cap at the api-server proxy layer. The 30s inactivity timeout only catches stalled connections; a large-but-continuously-streaming upload can fill the PVC. Consider a \`Content-Length\` reject threshold or a streaming byte counter at the proxy boundary. (\`packages/api-server/src/apps/api-server/app.ts:803\`)
- 🟡 **Warning:** The ADR states agent-runtime "atomically swaps" the bundle into \`/home/agent\`, but \`finalize.ts\` replace mode issues separate \`rm -rf\` and \`rename\` calls per top-level entry across two sequential loops — not a single atomic transaction. A crash between the loops leaves the destination partially cleared. Document this limitation in the ADR (or use a double-rename swap: stage → \`.import-new\`, old → \`.import-old\`, new → final, rm old). (\`packages/agent-runtime/src/modules/import/finalize.ts:385-392\`)
- 🟢 **Suggestion:** \`buildBundle\` in \`import-bundle.ts\` reads \`arrayBuffer()\` calls sequentially (\`for...of\` + \`await\`). For many small files, \`await Promise.all(entries.map(e => e.file.arrayBuffer()))\` then a separate build pass would reduce total wall time. (\`packages/ui/src/modules/files/api/import-bundle.ts:1405\`)
- ✅ **Looks good:** Path traversal, absolute-path, symlink, and reserved-segment protections in \`extract.ts\` are solid; all four attack vectors have dedicated unit tests (\`import-extract.test.ts\`).
- ✅ **Looks good:** Single-flight \`activeImport\` guard (409 on concurrent import) is clean and covers the race in replace mode.
- ✅ **Looks good:** Import ordering in \`useCreateAgent\` — before \`setAgentAccess\`/\`setAgentConnections\` — is well-explained in the comment and correctly avoids the pod-restart race window.
- ✅ **Looks good:** The \`platform-base\` Dockerfile refactor (\`pnpm deploy --prod\`) eliminates the previously-duplicated dependency list; \`platform-base/package.json\` is now a stub, which is correct.
- ✅ **Looks good:** The sweeper correctly limits itself to \`homeDir\` entries matching the prefix and uses \`mtime\` for age gating — won't touch live staging dirs or unrelated files.

### Verdict
APPROVE — Core security model (extraction validation, ownership check, single-flight guard) is solid and well-tested. The three warnings (STAGING_PREFIX duplication, no size cap, ADR atomicity claim) are worth a follow-up but don't block the feature.

---

xjacka:
🛡️ **Humr** — Code Review @ \`c53c48e\`

## PR #172: Feat: local project import
**Author:** jjeliga | **Branch:** feat/73-local-project-import → main | **Changes:** +2280 −590 (34 files)

### Summary
Implements full-stack local project import: users can drag-and-drop or pick folders/files during agent creation or from the files panel, and the bundle is uploaded as a tar.gz to the new instance's `/home/agent`. The agent-runtime gains a streaming multipart handler (`busboy`), path-safe extraction with traversal/symlink/reserved-segment rejection, atomic-ish replace/merge finalize logic, preflight conflict detection, and a crash-recovery sweeper. The UI builds a USTAR tar in the browser and adds a conflict modal for the files-panel flow.

### Changes since last review
Previous HEAD: `6b69ede` (2026-05-12T09:41:44Z) — verdict APPROVE

- ✅ **Fixed:** `STAGING_PREFIX` duplication (`http.ts`, `sweeper.ts`) — extracted to `packages/agent-runtime/src/modules/import/constants.ts`; both files now import from it
- ✅ **Fixed:** No bundle size cap at proxy layer — `maxImportBundleBytes` config added (`api-server/src/config.ts`) with both Content-Length header reject and a streaming byte counter `Transform` in the proxy body pipe (`api-server/src/apps/api-server/app.ts`); Helm values expose it to operators with a default 5 GiB
- ✅ **Fixed:** ADR atomicity claim — `docs/adrs/DRAFT-file-import.md` now explicitly documents "Atomicity, scoped" in Consequences, including the per-entry rm+rename crash window and the reason `renameat2(RENAME_EXCHANGE)` was not used
- 🆕 **New:** `requestTimeout = 0` set globally on both `agent-runtime/src/server.ts` and `api-server/src/apps/api-server/app.ts`
- 🆕 **New:** `import-conflict-modal.tsx` heading hardcoded as "Folder already exists" regardless of conflict type

### Findings
- 🟡 **Warning:** `server.requestTimeout = 0` is set globally on both `packages/agent-runtime/src/server.ts` and `packages/api-server/src/apps/api-server/app.ts:1493-1497`. The comments correctly justify disabling it for the `/api/import` route (multi-GB uploads exceed the 5-minute Node default), but this removes the body-read timeout for every other route on both servers — tRPC mutations, file uploads via the existing files surface, `/healthz`, etc. A slow client can hold any route's body open indefinitely. `headersTimeout = 60_000` still protects the headers phase, but it doesn't bound body streaming on non-import routes. Consider scoping the timeout override to the import handler only (e.g., override `req.socket.setTimeout(0)` inside `handleImport` after the import-specific inactivity timer is armed, then reinstate it on close).
- 🟢 **Suggestion:** `import-conflict-modal.tsx:2472` hardcodes the heading "Folder already exists" even when the conflicts are regular files (e.g., `CLAUDE.md`). The body copy handles the single-vs-plural case correctly but the heading is always wrong for file-only conflicts. A simple fix: `{conflicts.length === 1 && !conflicts[0].includes("/") ? "File already exists" : "Conflicts detected"}` or similar.
- ✅ **Looks good:** All three prior warnings resolved cleanly in this revision — `STAGING_PREFIX` is a single shared constant, the size cap is enforced both at the Content-Length header level and with a streaming byte counter that catches lying clients, and the ADR now openly states the atomicity scope.
- ✅ **Looks good:** Path traversal, absolute-path, symlink, and reserved-segment protections in `extract.ts` remain solid; all four attack vectors have dedicated unit tests in `import-extract.test.ts`.
- ✅ **Looks good:** `buildBundle` uses lazy `Blob` composition (each `File` is a `BlobPart` reference, not an `arrayBuffer()` materialization) — avoids the sequential-read concern raised in the prior review.
- ✅ **Looks good:** `proxyImport` uses `node:http` directly (not `undici`/`fetch`) with a raw stream pipe to avoid buffering multi-GB bodies in the api-server pod; `resolveOnce` guard prevents double-resolve on concurrent `response`/`error`/`close` events.
- ✅ **Looks good:** `finalize.ts` merge mode uses `lstat` (not `stat`) at the destination before recursing, ensuring pre-existing destination symlinks are replaced rather than followed — backed by a unit test covering the symlink-at-dest scenario.

### Verdict
APPROVE — All three prior warnings have been resolved and the implementation is well-tested. The global `requestTimeout = 0` change is a real concern but doesn't block the feature; the import route has independent timeout guards and `headersTimeout` still protects against slow-loris on headers for all routes.

---

xjacka:
🛡️ **Humr** — Code Review @ `7b457cf`

## PR #172: Feat: local project import
**Author:** jjeliga | **Branch:** feat/73-local-project-import → main | **Changes:** +2289 −592 (36 files)

### Summary
Implements full-stack local project import: users can drag-and-drop or pick folders/files during agent creation or from the files panel, and the bundle is uploaded as a tar.gz to the new instance's `/home/agent`. The agent-runtime gains a streaming multipart handler (`busboy`), path-safe extraction with traversal/symlink/reserved-segment rejection, atomic-ish replace/merge finalize logic, preflight conflict detection, and a crash-recovery sweeper. The UI adds import support to the agent-creation dialog with tarball pass-through for single-file drops.

### Changes since last review
Previous HEAD: `c53c48e` (2026-05-12T13:18:47Z) — verdict APPROVE

- 🆕 **New:** Import UI integrated into `add-agent-dialog.tsx` — drag-and-drop, folder/file picker, raw tarball pass-through for single `.tar`/`.tar.gz`/`.tgz` drop
- 🆕 **New:** `mutations.ts` `useCreateAgent` orchestrates import BEFORE `setAgentAccess`/`setAgentConnections` (comment explains the pod-restart race window)
- 🆕 **New:** `tseng/vocabulary.md` — Import bounded context terminology
- 🆕 **New:** `pnpm-workspace.yaml` — `trustPolicyExclude` for `rollup@2.80.0`, `semver@6.3.1`, `chokidar@4.0.3`
- 🔁 **Still present:** `server.requestTimeout = 0` global on both servers — carried over from prior review
- 🔁 **Still present:** "Folder already exists" hardcoded heading in import-conflict-modal — carried over from prior review

### Findings
- 🟡 **Warning:** `server.requestTimeout = 0` set globally on both `packages/agent-runtime/src/server.ts` and `packages/api-server/src/apps/api-server/app.ts`. The comments correctly justify disabling it for the `/api/import` route, but this removes the body-read timeout for every other route on both servers. `headersTimeout = 60_000` still protects the headers phase but doesn't bound body streaming on non-import routes. (`packages/agent-runtime/src/server.ts`, `packages/api-server/src/apps/api-server/app.ts`)
- 🟡 **Warning:** `pnpm-workspace.yaml` adds `rollup@2.80.0` and `semver@6.3.1` to `trustPolicyExclude` with only `# verified manually` — no pinned hash, commit link, or version justification. `chokidar@4.0.3` has a detailed explanation; the other two should too so future audits can re-verify rather than re-trust. (`pnpm-workspace.yaml`)
- 🟢 **Suggestion:** `import-conflict-modal.tsx:2532` heading hardcodes "Folder already exists" even when the conflicts include regular files (e.g. `CLAUDE.md`). Carried over from prior review.
- 🟢 **Suggestion:** In `add-agent-dialog.tsx`, once a raw tarball is set via `setImportRawBundle`, there's no way to remove it in the UI without adding more files (which switches to entry mode). A dedicated clear/reset affordance would make the intent explicit to users. (`packages/ui/src/modules/agents/dialogs/add-agent-dialog.tsx`)
- ✅ **Looks good:** Import fires before `setAgentAccess`/`setAgentConnections` in `mutations.ts` — correctly avoids the pod-restart race window; comment explains the ordering.
- ✅ **Looks good:** Tarball pass-through for single `.tar`/`.tar.gz`/`.tgz` drop avoids re-wrapping, switching to entry mode only when the user adds more items.
- ✅ **Looks good:** Import failure degrades to a toast (not a rollback) — agent is still created; user can retry from the files panel.
- ✅ **Looks good:** Path traversal, absolute-path, symlink, and reserved-segment protections in `extract.ts` remain solid with dedicated unit tests.

### Verdict
APPROVE — The UI integration is well-executed with correct sequencing and graceful degradation. Prior warnings (global requestTimeout) still apply but remain non-blocking.

---

xjacka:
🛡️ **Humr** — Code Review @ `432ae79`

## PR #172: Feat: local project import
**Author:** jjeliga | **Branch:** feat/73-local-project-import → main | **Changes:** +2294 −593 (37 files)

### Summary
Implements full-stack local project import: users can drag-and-drop or pick folders/files during agent creation or from the files panel, and the bundle is uploaded as a tar.gz to the new instance's `/home/agent`. New commits add Helm chart plumbing for `maxImportBundleBytes`, a DRAFT ADR documenting the design decisions and scoped-atomicity trade-off, and architecture doc updates for `persistence.md` and `platform-topology.md`.

### Changes since last review
Previous HEAD: `7b457cf` (2026-05-12T13:41:26Z) — verdict APPROVE

- 🆕 **New:** `docs/adrs/DRAFT-file-import.md` — documents the bundled, atomic, one-shot import design; Consequences section correctly acknowledges the scoped atomicity limitation
- 🆕 **New:** `deploy/helm/platform/values.yaml` + `apiserver/app.yaml` — exposes `maxImportBundleBytes` in Helm with a well-documented 5 GiB default and per-PVC-size warning
- 🆕 **New:** Architecture doc updates — `persistence.md` and `platform-topology.md` accurately describe the import path and staging directory lifecycle
- 🆕 **New:** `.gitignore` adds `docs/superpowers/`
- 🔁 **Still present:** `server.requestTimeout = 0` set globally on both servers — carried over from prior review
- 🔁 **Still present:** `pnpm-workspace.yaml` `rollup@2.80.0` and `semver@6.3.1` with only `# verified manually` — carried over from prior review

### Findings
- 🟡 **Warning:** `server.requestTimeout = 0` set globally on both `packages/agent-runtime/src/server.ts` and `packages/api-server/src/apps/api-server/app.ts`. The comments correctly justify disabling it for `/api/import`, but this removes the body-read timeout for every other route on both servers. `headersTimeout = 60_000` protects the header phase only; non-import routes have no body-streaming timeout. (`packages/agent-runtime/src/server.ts`, `packages/api-server/src/apps/api-server/app.ts`)
- 🟡 **Warning:** `pnpm-workspace.yaml` adds `rollup@2.80.0` and `semver@6.3.1` to `trustPolicyExclude` with only `# verified manually` — no pinned commit SHA or version-justification note. `chokidar@4.0.3` has a detailed explanation; the other two should too. (`pnpm-workspace.yaml`)
- 🟢 **Suggestion:** `import-conflict-modal.tsx` heading hardcodes "Folder already exists" even when the only conflicts are regular files. Carried over from prior review.
- 🟢 **Suggestion:** `importRawBundle` state has no explicit removal affordance in the UI — once a tarball is set, it persists until `appendEntries` folds it in. A visible "remove" button would make the intent clearer. (`packages/ui/src/modules/agents/dialogs/add-agent-dialog.tsx`)
- ✅ **Looks good:** `DRAFT-file-import.md` ADR correctly documents the scoped atomicity — "replace mode interleaves rm+rename per top-level entry so a crash mid-loop loses at most one top-level path" — consistent with `finalize.ts` implementation.
- ✅ **Looks good:** `maxImportBundleBytes` default of 5 GiB is well-documented with the PVC-size warning; rendered optional in the Helm template so clusters on older chart versions don't break.
- ✅ **Looks good:** Platform topology and persistence docs accurately describe the import path — the new table row in `platform-topology.md` correctly identifies the streamed-multipart shape.

### Verdict
APPROVE — New commits add the Helm plumbing and documentation previously deferred; no new blocking issues. Prior warnings (global requestTimeout, trustPolicyExclude justification) remain but don't block.

---

xjacka:
🛡️ **Humr** — Code Review @ \`1f12eb3\`

## PR #172: Feat: local project import
**Author:** jjeliga | **Branch:** feat/73-local-project-import → main | **Changes:** +2295 −593 (37 files)

### Summary
Implements full-stack local project import with add-agent-dialog UI integration, api-server proxy with streaming size cap, agent-runtime extraction engine (staging/finalize/sweeper), a DRAFT ADR, Helm plumbing for `maxImportBundleBytes`, and architecture doc updates. The new commit fixes the import conflict modal copy and patches a double-read bug in the e2e test.

### Changes since last review
Previous HEAD: `432ae79` (2026-05-12T13:49:42Z) — verdict APPROVE

- ✅ **Fixed:** `import-conflict-modal.tsx` heading "Folder already exists" → "Already exists"; "A folder named" → "An entry named"; body copy updated to "existing entry" — addresses prior 🟢 Suggestion
- 🆕 **New:** `import.test.ts` e2e test — buffered response text before JSON.parse to fix a double-read bug (`res.text()` then `res.json()` is invalid on a consumed stream; now: `res.text()` → `JSON.parse()`)
- 🆕 **New:** `add-agent-dialog.tsx` — removed redundant "Pre-built bundle: " label prefix from raw tarball filename display (cosmetic)

### Findings
- 🟡 **Warning:** `server.requestTimeout = 0` set globally on both `packages/agent-runtime/src/server.ts` and `packages/api-server/src/apps/api-server/app.ts`. Comments correctly justify disabling it for the `/api/import` route, but this removes the body-read timeout for every other route on both servers. `headersTimeout = 60_000` protects headers only; non-import routes have no body-streaming timeout. Carried over from prior reviews.
- 🟡 **Warning:** `pnpm-workspace.yaml` adds `rollup@2.80.0` and `semver@6.3.1` to `trustPolicyExclude` with only `# verified manually` — no pinned commit SHA or version justification. `chokidar@4.0.3` has a detailed explanation; the other two should too so future audits can re-verify. Carried over from prior reviews.
- ✅ **Looks good:** import-conflict-modal copy now handles files and folders generically — "Already exists" / "An entry named..." is accurate for single-file conflicts like `CLAUDE.md`.
- ✅ **Looks good:** e2e test double-read fix is correct — `res.text()` then `JSON.parse()` avoids the once-consumed-stream issue.
- ✅ **Looks good:** A "Clear" button is visible whenever import data is staged; it resets `importEntries`, `importRawBundle`, and `importDropped` together.

### Verdict
APPROVE — Minor cleanup commit addresses the conflict modal copy. Prior non-blocking warnings (global requestTimeout, trustPolicyExclude justifications) remain.

---

packages/agent-runtime/src/server.ts
```diff
@@ -267,9 +282,22 @@ try {
...
 
+// Node defaults `requestTimeout` to 5 minutes. The import route holds a
+// request open through extract+finalize of a multi-GB tar — easily
+// >5 min on slower uploads. Disable the per-request timeout; the import
+// handler installs its own inactivity (30s) + wall-clock (30min)
+// deadlines, so a stuck connection still gets aborted. Keep
+// `headersTimeout` (60s) so a misbehaving client can't slow-loris the
+// header line indefinitely on other routes (/api/status, /api/trpc,
+// /healthz).
+server.requestTimeout = 0;
```
PetrBulanek, line 297: **`requestTimeout = 0` affects every route, not just import** — [agent-runtime/src/server.ts:293](packages/agent-runtime/src/server.ts), [api-server/src/apps/api-server/app.ts:486-491](packages/api-server/src/apps/api-server/app.ts)
Node's 5-min body-read timeout is off cluster-wide now. Import legitimately needs that, but tRPC, ACP relay, files upload and `/healthz` lose their slow-body protection too — a client can hold any route open indefinitely. Suggest scoping the override to inside `handleImport` only.

packages/agent-runtime/src/modules/import/http.ts
```diff
@@ -0,0 +1,244 @@
...
+        await fail(500, (err as Error).message);
+      } finally {
+        clearActive();
+      }
+    });
+
+    req.pipe(bb);
+  }
+
+  async function handleImportPreflight(req: http.IncomingMessage, res: http.ServerResponse): Promise<void> {
```
PetrBulanek: **Preflight endpoint reads the request body with no size cap** — [agent-runtime/src/modules/import/http.ts:219-241](packages/agent-runtime/src/modules/import/http.ts)
`handleImportPreflight` buffers the whole body in memory, and the proxy passes `enforceSizeCap: false` for this path. A 1 GB JSON body OOMs the agent-runtime pod, which takes down the user's session with it.

packages/agent-runtime/src/modules/import/http.ts
```diff
@@ -0,0 +1,244 @@
...
+      extractPromise = (async () => {
+        staging = await mkdtemp(join(homeDir, STAGING_PREFIX));
+        log(`staging dir created: ${staging}`);
+        const result = await extractBundle(fileStream as never, staging);
+        log(`extract complete (ok=${result.ok})`);
+        return result;
+      })();
+    });
+
+    bb.on("error", (err: Error) => { void fail(400, `multipart: ${err.message}`); });
```
PetrBulanek, line 143: **`bb.on("error")` is the only exit path that doesn't explicitly clear the single-flight lock** — [http.ts:170](packages/agent-runtime/src/modules/import/http.ts)
Other paths (close, idle timeout, deadline) all call `clearActive()` directly; the error handler relies on `bb.on("close")` firing afterwards. In standard Node stream semantics it will, so this is more defense-in-depth than a real leak — but the function is idempotent and adding it would match the pattern used elsewhere in the file.

packages/ui/src/modules/agents/dialogs/add-agent-dialog.tsx
```diff
@@ -48,6 +51,87 @@ export function AddAgentDialog({
...
+    );
+  };
+
+  // Pass-through only when the user picks/drops exactly one tarball and
+  // nothing else is selected yet. Otherwise the tarball is bundled with
+  // everything else as a regular file.
+  const handleIncoming = (incoming: BundleEntry[]) => {
+    if (
+      incoming.length === 1
+      && isTarballName(incoming[0].path)
```
PetrBulanek, line 126: **`isTarballName` check matches deep paths** — [add-agent-dialog.tsx:124](packages/ui/src/modules/agents/dialogs/add-agent-dialog.tsx)
Dropping a folder that contains exactly one `.tar.gz` flips the dialog into pass-through mode and silently uploads just the inner file, dropping the folder. Adding `!incoming[0].path.includes("/")` to the guard fixes it.

pnpm-workspace.yaml
```diff
@@ -3,6 +3,14 @@ packages:
   - "packages/agents/pi-agent/workspace/.pi/agent/extensions/*"
 minimumReleaseAge: 10080 # 7 days (in minutes)
 trustPolicy: no-downgrade
+trustPolicyExclude:
+  # verified manually
```
PetrBulanek: **`rollup@2.80.0` and `semver@6.3.1` in `trustPolicyExclude` say only `# verified manually`** — `chokidar@4.0.3` has a proper paragraph; the other two could use the same so a future audit can re-verify rather than re-trust.

packages/api-server/src/apps/api-server/app.ts
```diff
@@ -240,6 +242,149 @@ export function startApiServerApp(deps: ApiServerAppDeps) {
...
+        return c.json({ error: "invalid Content-Length" }, 400);
+      }
+      if (length > config.maxImportBundleBytes) {
+        return c.json({ error: `bundle exceeds maximum size of ${config.maxImportBundleBytes} bytes` }, 413);
+      }
+    }
+    try {
+      await instancesRepo.ensureReady(instanceId);
+    } catch (err) {
+      return c.json({ error: `instance unreachable: ${(err as Error).message}` }, 502);
```
PetrBulanek: **502 message diverges from the pattern used elsewhere in the same file** — [app.ts:299](packages/api-server/src/apps/api-server/app.ts)
The tRPC proxy a few lines up returns a clean `"instance unreachable"`; the import proxy interpolates `(err as Error).message` instead. Worth matching the existing shape — log the detail, send the generic message.

PetrBulanek (changes requested):

packages/agent-runtime/src/server.ts
jjeliga: Real point, but the suggested scoping isn't achievable through public Node API: server.requestTimeout is an absolute timer attached at request start, not a socket-idle timer, so there's no per-handler clear short of poking socket._requestTimer (private).

Practical exposure is bounded by the surrounding hops: agent-runtime is reachable only from api-server (NetworkPolicy), api-server sits behind Traefik with its own timeouts, and headersTimeout = 60s still protects the header phase on every route. Slow-body on the api-server's other routes would tie up a TCP connection but bodies are capped (TRPC_MAX_BODY_SIZE = 32 MB, tRPC own caps) so memory growth is bounded.

Going to keep requestTimeout = 0 and expand the inline comment with that rationale.

packages/agent-runtime/src/modules/import/http.ts
jjeliga: preflight will be deleted

packages/agent-runtime/src/modules/import/http.ts
jjeliga: Busboy emits `close` after `error`; the close handler clears the lock in its `finally`. `clearActive` is idempotent so an added call would duplicate already-scheduled cleanup without changing behavior. Leaving as-is unless you've seen busboy skip the `close` emit in practice.

packages/ui/src/modules/agents/dialogs/add-agent-dialog.tsx
jjeliga: Real bug, fixing

pnpm-workspace.yaml
jjeliga: They all had basically the same paragraph, didn't want to spam, procedure was identical for all of them, can expand this

packages/api-server/src/apps/api-server/app.ts
jjeliga: ok

packages/agent-runtime/src/modules/import/http.ts
PetrBulanek: I didn't know that, thanks for explaining. Let's just leave it as it is, then.

xjacka:
🛡️ **Humr** — Code Review @ \`c026621\`

## PR #172: Feat: local project import
**Author:** jjeliga | **Branch:** feat/73-local-project-import → main | **Changes:** +1581 −592 (31 files)

### Summary
Implements full-stack local project import: drag-and-drop or pick folders/files during agent creation or from the files panel; the bundle is uploaded as a tar to the new instance's `<homeDir>/work` with top-level replace semantics. New: comprehensive UI in `add-agent-dialog.tsx`, client-side USTAR tar builder in `import-bundle.ts`, files panel folder upload, and `platform-base` Dockerfile refactored to `pnpm deploy --prod`.

### Changes since last review
Previous HEAD: `1f12eb3` (2026-05-12T14:08:06Z) — verdict APPROVE

- ✅ **Fixed:** `pnpm-workspace.yaml` trustPolicyExclude entries for `rollup@2.80.0` and `semver@6.3.1` now have detailed justification comments (same upstream, no open advisories, transitive-only). Prior 🟡 Warning addressed.
- 🆕 **New:** Comprehensive UI import dialog in `add-agent-dialog.tsx` — drag-and-drop zone, folder/file pickers, raw tarball pass-through, per-group chip display with remove buttons, Clear button
- 🆕 **New:** Client-side USTAR tar builder in `import-bundle.ts` — handles prefix/name split for paths > 100 bytes, lazy Blob construction, deliberate no-gzip (well-justified)
- 🆕 **New:** Files panel folder upload button using `walkDataTransfer` + `uploadBundle`
- 🆕 **New:** `platform-base` Dockerfile refactored to `pnpm deploy --prod --node-linker=hoisted` — eliminates manually-maintained dependency list
- 🆕 **New:** `tasks.toml` cross-platform `date` fix (GNU `-d` vs BSD `-v`)
- 🆕 **New:** Vocabulary additions in `tseng/vocabulary.md` (Bundle, Staging Dir, Finalize)
- 🔁 **Still present:** `server.requestTimeout = 0` globally on both servers (🟡 Warning carried over)

### Findings
- 🟡 **Warning:** `server.requestTimeout = 0` set globally on both `packages/agent-runtime/src/server.ts` and `packages/api-server/src/apps/api-server/app.ts`. Comments correctly justify it for the import route, but this removes body-read timeout for every other route on both servers. Mitigations are adequate (`headersTimeout = 60s`, hard body caps on non-import routes, NetworkPolicy gating, Traefik ingress timeouts). Carried over from prior reviews.
- 🟢 **Suggestion:** `folderPickerTargetDirRef` in `files-panel.tsx` is initialized and set in two places but never read — `uploadBundle` doesn't accept a target directory parameter, so folder imports always land at `<homeDir>/work` root regardless of which directory is open in the panel. Either pass a prefix to `importBundle`, or remove the dead ref. (`packages/ui/src/modules/files/components/files-panel.tsx`)
- 🟢 **Suggestion:** `buildBundle` in `import-bundle.ts` doesn't validate individual file sizes against USTAR's 11-byte octal `size` field limit (~8 GB). A file exceeding this produces a silently truncated tar header. Unlikely given the 5 GiB server cap, but worth a guard. (`packages/ui/src/modules/files/api/import-bundle.ts`)
- ✅ **Looks good:** `pnpm-workspace.yaml` trustPolicyExclude entries for `rollup@2.80.0` and `semver@6.3.1` now have detailed justification comments.
- ✅ **Looks good:** Client-side USTAR tar builder correctly handles prefix/name split for long paths, uses lazy Blob construction (entries reference `File` objects), and deliberately skips gzip.
- ✅ **Looks good:** `platform-base` Dockerfile now uses `pnpm deploy --prod --node-linker=hoisted` — production deps derived from `agent-runtime/package.json` rather than the previously manually-maintained list.
- ✅ **Looks good:** Import ordering in `mutations.ts` fires before `setAgentAccess`/`setAgentConnections` — pod-restart race window clearly documented.
- ✅ **Looks good:** `walkDataTransfer` deduplicates on path — dropping the same folder twice won't produce duplicate tar entries.
- ✅ **Looks good:** `tasks.toml` cross-platform `date` fix correctly branches on GNU `-d` vs BSD `-v`.

### Verdict
APPROVE — comprehensive and well-documented implementation. The global `requestTimeout = 0` remains the only non-blocking warning, carried over from prior reviews.

---

PetrBulanek (approved):

xjacka:
🛡️ **Humr** — Code Review @ `d295b418`

## PR #172: Feat: local project import
**Author:** jjeliga | **Branch:** feat/73-local-project-import → main | **Changes:** +1581 −592 (31 files)

### Summary
Implements full-stack local project import: users can drag-and-drop folders or files during agent creation or from the files panel; the bundle is uploaded as a USTAR tar to the new instance's `<homeDir>/work` with top-level replace semantics. The agent-runtime gains a streaming multipart handler (`busboy`), path-safe extraction with traversal/symlink/reserved-segment rejection, and a clean finalize that replaces top-level entries atomically (rm+rename per entry). The UI adds folder-upload to the files panel and a comprehensive import section to the agent-creation dialog.

### Changes since last review
Previous HEAD: `c026621` (2026-05-13T13:13:43Z) — verdict APPROVE

- ✅ **Improved:** `finalize.ts` radically simplified — 40 lines replacing prior complex per-file merge logic. Semantics changed from per-file merge to top-level replace (folder-as-atomic-unit: importing `dirA/` replaces the whole `dirA/`, not individual files). User-visible change, clearly documented.
- ✅ **Improved:** `tsup.config.ts` now explicitly documents why `agent-runtime-api`/`api-server-api` live in devDeps (`noExternal` bundles them into dist at build time).
- ✅ **New:** E2e test in `packages/api-server/src/__tests__/import.test.ts` covers top-level replace semantics, partial replace, and concurrent-import 409.
- 🔁 **Still present:** `server.requestTimeout = 0` globally on both servers — 🟡 Warning carried over.
- 🔁 **Still present:** `folderPickerTargetDirRef` dead write in `files-panel.tsx` — 🟢 Suggestion carried over.
- 🔁 **Still present:** `buildBundle` no guard for individual file size vs USTAR size field limit — 🟢 Suggestion carried over.
- 🆕 **New:** `extract.ts` `onentry` fires before `filter`, so `filesWritten`/`bytes` in the 200 response include filtered-out invalid entries (metric inaccuracy).
- 🆕 **New:** `http.ts` staging cleanup race on timeout-during-extraction: if timeout fires before `staging` is assigned, `fail()` skips rm; sweeper is then the only reclaim path.

### Findings
- 🟡 **Warning:** `server.requestTimeout = 0` is set globally on both `packages/agent-runtime/src/server.ts` and `packages/api-server/src/apps/api-server/app.ts`. Comments correctly justify it for `/api/import`, but this removes the body-read timeout for every other route on both servers. `headersTimeout = 60s` protects headers only. Carried over from prior reviews.
- 🟢 **Suggestion:** `folderPickerTargetDirRef` in `files-panel.tsx` is always reset to `""` and never read by `uploadBundle` (no target-dir param). Either wire it up to a prefix param on `importBundle`, or remove it. (`packages/ui/src/modules/files/components/files-panel.tsx`)
- 🟢 **Suggestion:** `buildBundle` calls `writeOct(buf, 124, size, 12)` with no guard for files exceeding USTAR's ~8 GiB size field limit. Unlikely given the 5 GiB cap but a client lying about `Content-Length` could bypass it. (`packages/ui/src/modules/files/api/import-bundle.ts`)
- 🟢 **Suggestion:** In `extract.ts`, `onentry` increments `filesWritten`/`bytes` before `filter` runs — the 200 response counts filtered-out invalid entries as written. Move the increment inside a successful-extraction check. (`packages/agent-runtime/src/modules/import/extract.ts`)
- 🟢 **Suggestion:** In `http.ts`, if a timeout fires while `extractPromise` is in flight and `staging` isn't yet assigned, `fail()` skips cleanup and `bb.on("close")` exits early via `finished` guard — staging dir leaks until the sweeper reclaims it. Add a comment making explicit that the sweeper is the only cleanup path for this race. (`packages/agent-runtime/src/modules/import/http.ts`)
- ✅ **Looks good:** `finalize.ts` top-level replace semantics clean and well-documented; crash-window caveat explicitly stated.
- ✅ **Looks good:** `extract.ts` path safety: rejects non-File/non-Directory types, absolute paths (including Windows UNC), `..` components, and resolve-escapes.
- ✅ **Looks good:** `agent-runtime/package.json` correctly moves `agent-runtime-api`/`api-server-api` to devDeps; `tsup.config.ts` `noExternal` bundles them into dist.
- ✅ **Looks good:** E2e test covers top-level replace, partial replace (untouched entries), and concurrent-import 409 with `.tar.gz` input.
- ✅ **Looks good:** `filterImportEntries` excludes `node_modules`, `.venv`, `__pycache__`, `.DS_Store` before building tar.

### Verdict
APPROVE — Finalize refactor is a genuine improvement: simpler, more predictable top-level replace semantics with clear atomicity documentation. Prior non-blocking warning (global requestTimeout) remains.

---

PR #189, opened 2026-05-13
fix(tasks): don't wipe agent images on ui or api-server  build

### LINKED ISSUE #185: 🐛 Agent pods falling into `Init:ErrImageNeverPull` state
## Steps to reproduce:

`mise run cluster:install`
open the ui, create agent, do some stuff, delete the agent
some time later, create new agent - it might not come up
if it doesn't come up:
observe the cluster state via `mise run cluster:status`
see that the agent pod is in `Init:ErrImageNeverPull` state
Deleting and re-creating the agent does not work.
Re-build does work

jjeliga:
Closes #185

### DIFF
--- a/deploy/tasks.toml
+++ b/deploy/tasks.toml
@@ -357,11 +357,11 @@ echo "Restarting apiserver pod..."
 kubectl --kubeconfig="$KUBECONFIG" rollout restart deployment platform-apiserver
 kubectl --kubeconfig="$KUBECONFIG" rollout status deployment platform-apiserver --timeout=60s
 
-# Drop image layers no longer pinned by a running container (issue #244).
+PRUNE_DANGLING='sudo k3s crictl images 2>/dev/null | sed -nE "s/^<none>[[:space:]]+<none>[[:space:]]+([^[:space:]]+).*/\\1/p" | xargs -r sudo k3s crictl rmi >/dev/null 2>&1 || true'
 if [ -n "${IS_SANDBOX:-}" ]; then
-  sudo k3s crictl rmi --prune >/dev/null 2>&1 || true
+  bash -c "$PRUNE_DANGLING"
 else
-  limactl shell "$LIMA_INSTANCE" sudo k3s crictl rmi --prune >/dev/null 2>&1 || true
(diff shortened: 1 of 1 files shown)


xjacka:
🛡️ **Humr** — Code Review @ \`d7c6983\`

## PR #189: fix(tasks): don't wipe agent images on ui or api-server build
**Author:** jjeliga | **Branch:** fix/185-stop-agent-pod-deletion-on-other-builds → main | **Changes:** +6 −5 (1 file)

### Summary
Replaces \`crictl rmi --prune\` (which removes all unused images) with a targeted \`sed\`-based pipeline that removes only truly dangling images (those with \`<none>\` name and tag), preventing agent pod images from being wiped during unrelated UI or API server builds.

### Findings
- 🟢 **Suggestion:** \`PRUNE_DANGLING\` is defined identically in both the \`apiserver\` and \`ui\` restart tasks (\`deploy/tasks.toml:360\` and \`deploy/tasks.toml:398\`). If tasks.toml supports shared variables or a helper section, this could be defined once. Minor — duplication is small and the scripts are otherwise independent.
- 🟢 **Suggestion:** The \`limactl shell "$LIMA_INSTANCE" bash -c "$PRUNE_DANGLING"\` form expands \`$PRUNE_DANGLING\` in the host shell before limactl receives it, so the full command string runs on the Lima VM. This works correctly with the static string, but worth noting that any future dynamic content in the command would be shell-expanded on the host first (\`deploy/tasks.toml:364\`, \`deploy/tasks.toml:402\`).
- ✅ **Looks good:** The \`sed -nE\` regex correctly targets only images where both name and tag are \`<none>\`, avoiding any named images.
- ✅ **Looks good:** \`xargs -r\` safely no-ops when there are no dangling images.
- ✅ **Looks good:** All subcommands use \`|| true\` and \`>/dev/null 2>&1\` — failures won't break the build pipeline.

✅ No documentation drift detected. The change is confined to \`deploy/tasks.toml\` (an operational build script); no architecture subsystem behavior is altered.

### Verdict
APPROVE — Targeted fix is correct and safe; the suggestions are minor style points that don't warrant blocking the merge.

---

tomkis (approved):

PR #192, opened 2026-05-13
feat(helm): opt-in anyuid SCC binding gated on kata runtime

### LINKED ISSUE #194: Agent containers cannot run as root under kata on OpenShift
## Problem

Several agent harnesses we want to run in DAM expect root inside the container — package installers, build tooling, and a handful of MCP servers that bind to privileged ports or write to system paths assume UID 0. On vanilla Kubernetes this is uncomfortable but workable. On OpenShift it is impossible by default: the cluster-wide `restricted-v2` SCC enforces `MustRunAsRange`, so any pod the controller spawns is admitted with a random non-root UID regardless of what the pod spec says. The harness then trips on the very first `apt install` / `chown /opt/...` / `setcap` call and dies.

The same OpenShift cluster ships kata-containers (`kata`, `kata-nvidia-gpu` runtime classes). Inside a kata microVM, "root" is guest-VM root and never escapes to the host — exactly the isolation model that makes running agents as root safe in production. But the SCC admission gate runs *before* the runtime is selected, so kata on its own does not unblock root; we also need to grant the agent ServiceAccounts an SCC that permits `runAsUser: 0`.

The footgun is the inverse pairing: granting `anyuid` *without* kata gives every agent UID 0 on the host's default OCI runtime, which defeats the sandbox boundary the platform is built around (ADR-035, ADR-038).

## Proposed solution

A single opt-in chart value that wires SCC access to the agent namespace, with a chart-level precondition that refuses to install/upgrade unless kata is also configured. Operators get a one-line knob to flip; the chart enforces the safety pairing so nobody can land in the "anyuid without kata" state by accident.

Concretely: `openshift.scc.anyuid.enabled` binds `system:openshift:scc:anyuid` to the `system:serviceaccounts:<agentNamespace>` group via a namespace-scoped `RoleBinding`. A validator template aborts the render when the flag is on but `controller.agent.base.runtimeClassName` is not a kata runtime, or when `agentNamespace` is empty. No-op on vanilla Kubernetes.

### Out of scope

- Plumbing `runtimeClassName: kata` *automatically* when `anyuid` is enabled — the value already exists, operators set it explicitly. The validator only checks that they did.
- Per-instance SCC selection (some agents `anyuid`, others `privileged`). Single SCC scope, namespace-wide, in this slice.
- Other OpenShift-specific knobs (image pull secrets, route TLS, etc.) — separate concerns.

jezekra1:
Closes #194.

## Summary

- Adds `openshift.scc.anyuid.enabled` (default `false`) and a namespace-scoped `RoleBinding` that grants `system:openshift:scc:anyuid` to all ServiceAccounts in `agentNamespace`. Required on OpenShift when agents need `runAsUser: 0` inside a kata-containers sandbox — the default `restricted-v2` admission strips root regardless of the pod's intent.
- Introduces `templates/_validators.tpl` as the home for chart-level preconditions, dispatched once from `templates/validate.yaml`. Validators refuse to render when `openshift.scc.anyuid.enabled=true` is set without `controller.agent.base.runtimeClassName` being a kata runtime, or when `agentNamespace` is empty/null/whitespace — granting anyuid without kata isolation lands agents at UID 0 on the host's default OCI runtime, which defeats the sandbox boundary the platform relies on.
- No-op on vanilla Kubernetes. Scope is one namespace — no cluster-wide escalation.

SCC admission is evaluated against the per-instance ServiceAccount the controller creates (ADR-041), so binding to the `system:serviceaccounts:<agentNamespace>` group covers every agent and fork SA automatically — controller RBAC unchanged.

## Test plan

- [x] `helm lint` clean
- [x] `helm template | kubeconform -strict` clean
- [x] Defaults render with no anyuid resources
- [x] `anyuid=true` without a kata runtimeClass → `fail` aborts the render with a clear message
- [x] `anyuid=true` with `runtimeClassName=kata` → RoleBinding renders into the right namespace
- [x] `anyuid=true` with `runtimeClassName=kata-nvidia-gpu` → also accepted (any `kata*` prefix)
- [x] `anyuid=true` with `runtimeClassName=foo` → rejected
- [x] `anyuid=true` with empty / null / whitespace `agentNamespace` → rejected
- [ ] On the prod OpenShift cluster, flip both values on in the dev-sandboxed namespace and confirm agent pods are admitted with the `anyuid` SCC annotation and `runtimeClassName: kata`

### DIFF
--- a/deploy/helm/platform/templates/_validators.tpl
+++ b/deploy/helm/platform/templates/_validators.tpl
@@ -1,0 +1,32 @@
+{{/*
+Chart-level validators. Each `platform.validate.*` template either
+no-ops or calls `fail` to abort `helm install / upgrade / template`.
+The top-level `platform.validate` dispatches to all of them; it is
+invoked once from `templates/validate.yaml`.
+
+To add a new validator: define `platform.validate.<name>` here and
+add it to the include list in `platform.validate`.
+*/}}
+
+{{- define "platform.validate" -}}
+{{- include "platform.validate.anyuidRequiresKata" . -}}
+{{- end -}}
+
+{{/*
+Granting `system:openshift:scc:anyuid` lets agent containers run as
+UID 0. That is only acceptable when the workload is isolated inside a
+kata-containers microVM — "root" is then guest-VM root, not host root.
+Without kata, anyuid lands the agent at UID 0 on the host's default
+OCI runtime, which defeats the sandbox boundary the platform relies on.
+*/}}
+{{- define "platform.validate.anyuidRequiresKata" -}}
+{{- if and .Values.openshift .Values.openshift.scc .Values.openshift.scc.anyuid .Values.openshift.scc.anyuid.enabled -}}
+{{- $rc := "" -}}
+{{- if and .Values.controller .Values.controller.agent .Values.controller.agent.base -}}
+{{- $rc = .Values.controller.agent.base.runtimeClassName | default "" -}}
+{{- end -}}
+{{- if not (hasPrefix "kata" $rc) -}}
+{{- fail (printf "openshift.scc.anyuid.enabled=true requires controller.agent.base.runtimeClassName to be a kata runtime (got %q). Granting anyuid without kata isolation puts agents at UID 0 on the host's default OCI runtime." $rc) -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
--- a/deploy/helm/platform/templates/_validators.tpl
+++ b/deploy/helm/platform/templates/_validators.tpl
@@ -10,6 +10,7 @@ add it to the include list in `platform.validate`.
 
 {{- define "platform.validate" -}}
 {{- include "platform.validate.anyuidRequiresKata" . -}}
+{{- include "platform.validate.anyuidRequiresAgentNamespace" . -}}
 {{- end -}}
 
 {{/*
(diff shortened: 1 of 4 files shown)


pilartomas (approved):

xjacka:
🛡️ **Humr** — Code Review @ `5a27523`

## PR #192: feat(helm): opt-in anyuid SCC binding gated on kata runtime
**Author:** jezekra1 | **Branch:** feat/openshift-anyuid-scc → main | **Changes:** +103 −0 (4 files)

### Summary
Adds an opt-in Helm flag (`openshift.scc.anyuid.enabled`) that binds the `system:openshift:scc:anyuid` SCC to all service accounts in `agentNamespace`, enabling agent pods to run as UID 0 inside kata-containers sandboxes on OpenShift. A chart-level validator enforces that the feature cannot be enabled without a kata runtime class configured, preventing accidental host root escalation.

### Findings
- 🟡 **Warning:** `agent-scc.yaml` grants anyuid to `system:serviceaccounts:<agentNamespace>` — the entire SA group for the namespace — including any future workloads deployed there. If a non-agent workload lands in `agentNamespace`, it inherits the anyuid binding. Consider whether a SA-specific subject would be safer long-term. (`deploy/helm/platform/templates/agent-scc.yaml:29`)
- 🟢 **Suggestion:** `agent-scc.yaml` uses `{{- if .Values.openshift.scc.anyuid.enabled }}` while `_validators.tpl` uses a multi-step `and` guard. Helm handles nil map keys gracefully so this is safe, but consistent use of a `with` block or `and` guard would improve readability. (`deploy/helm/platform/templates/agent-scc.yaml:1`)
- ✅ **Looks good:** Validator correctly fails at `helm install/upgrade/template` time when anyuid is enabled without a kata runtime class — catches dangerous host root escalation before any manifest is applied.
- ✅ **Looks good:** Default `enabled: false` ensures the feature is a true no-op on vanilla Kubernetes and standard OpenShift installs.
- ✅ **Looks good:** Uses a namespace-scoped `RoleBinding` (not `ClusterRoleBinding`), correctly containing the SCC grant to a single namespace.

### Verdict
APPROVE — Well-gated, well-documented feature with sensible defaults and a proper validator that prevents the most dangerous misconfiguration.

---

PR #196, opened 2026-05-13
feat(helm): ship custom SCC granting DAC_OVERRIDE for kata agents

jezekra1:
## Summary

Replaces the stock `system:openshift:scc:anyuid` binding shipped in #192 with a chart-managed SCC (`<release>-platform-agent-kata`) that is identical to `anyuid` *plus* the single capability `DAC_OVERRIDE`. Same gating flag (`openshift.scc.anyuid.enabled`), same validator preconditions (kata runtimeClass + non-empty agentNamespace), tighter underneath.

### Why DAC_OVERRIDE

Stock `anyuid` permits any UID but explicitly forbids cap additions (`allowedCapabilities: null`). On deployments whose agent home PVC is NFS-backed with `root_squash` enforced server-side — IBM Cloud File Storage for VPC, the default on IBM Cloud OpenShift clusters, and the knob is not customer-configurable per the driver source ([`pkg/ibmcsidriver/constants.go`](https://github.com/IBM/ibm-vpc-file-csi-driver/blob/master/pkg/ibmcsidriver/constants.go) has no `rootSquash` parameter) and per IBM's own service docs — root writes get translated to UID 99 (`nobody`) on the share. Files persist as 99-owned, and subsequent kata-guest writes EACCES because UID 0 with `drop: ["ALL"]` lacks `CAP_DAC_OVERRIDE` to bypass the in-guest DAC check.

`DAC_OVERRIDE` is the surgical fix: it grants "ignore file-permission checks inside the guest filesystem view" and nothing else — no `SYS_ADMIN`, no `NET_*`, no `SETUID`, no module load. Containment remains the kata VM boundary. The NFS server continues to squash root writes regardless, so the agent's "I'm root" view is purely a kata-guest illusion the share doesn't honor — which is actually a useful additional containment property.

### Why a custom SCC rather than `add`-ing to stock

No existing SCC on a stock OpenShift install permits adding `DAC_OVERRIDE`:

- `anyuid`, `restricted-v2`, `nonroot-v2`, `hostnetwork-v2` — `allowedCapabilities` is `null` or `["NET_BIND_SERVICE"]` only
- `privileged`, all `nvidia-*` — `allowedCapabilities: ["*"]` but grant far more than DAC_OVERRIDE (host paths, host network, all caps)
- `sandboxed-containers-operator-scc` — `MustRunAsNonRoot`, incompatible with the root requirement

`priority: 9` keeps stock `anyuid` (10) as the default pick for pods that don't need the cap; the namespace-scoped RoleBinding to `system:openshift:scc:<release>-platform-agent-kata` is what actually grants its use.

## Test plan

- [x] `helm lint` clean
- [x] `helm template | kubeconform -strict` clean (SCC CRD is in kubeconform's skip list — 1 of 5 skipped resources)
- [x] Defaults render with no SCC manifests at all
- [x] `openshift.scc.anyuid.enabled=true` + kata + agentNamespace → both SCC and RoleBinding render with the right names
- [x] Validator still rejects `anyuid.enabled=true` without a kata runtimeClass
- [x] Validator still rejects `anyuid.enabled=true` with empty/null/whitespace `agentNamespace`
- [ ] On the prod OpenShift cluster: install with the flag on, confirm the SCC is created and the agent pod is admitted under it (`openshift.io/scc: <release>-platform-agent-kata` annotation) and the agent process can write to its NFS-backed `/home/agent`

### DIFF
--- a/deploy/helm/platform/templates/agent-scc.yaml
+++ b/deploy/helm/platform/templates/agent-scc.yaml
@@ -1,30 +1,104 @@
 {{- if .Values.openshift.scc.anyuid.enabled }}
 {{/*
-OpenShift SCC binding for the agent namespace.
+OpenShift SCC for agent pods on this chart.
 
-Default OpenShift admission (`restricted-v2`) forces a random UID and
-strips `runAsUser: 0`, which breaks harnesses that need root inside the
-kata-containers sandbox. Binding `system:openshift:scc:anyuid` to the
-`system:serviceaccounts:<agentNamespace>` group lets every per-instance
-ServiceAccount the controller creates (see ADR-041) pick its own UID,
-including 0. Scope is intentionally one namespace — no cluster-wide
-escalation.
+Stock OpenShift admission (`restricted-v2`) forces a random UID from
+the namespace range and refuses any cap additions, which breaks
+harnesses that need root inside the kata-containers sandbox. Stock
+`anyuid` permits any UID but explicitly forbids cap additions
+(`allowedCapabilities: null`), which is fine until you hit an NFS-
+backed home PVC: if the underlying NFS server enforces `root_squash`
+(IBM Cloud File Storage for VPC does, and the knob is not customer-
+configurable), every root write is translated to UID 99 (`nobody`)
+on the share. Files persist as 99-owned, and the kata-guest kernel
+then EACCES's subsequent writes because UID 0 with `drop: ["ALL"]`
+has no `CAP_DAC_OVERRIDE` to bypass DAC.
 
-`anyuid` does not grant capabilities, host paths, or privileged mode;
-it only loosens the UID constraint. Under kata, "root" is guest-VM root,
-not host root.
+This SCC is the minimal escalation that resolves the EACCES while
+keeping the security posture we already accepted (root + kata):
+identical to `anyuid` for UID / privilege / volume / SELinux rules,
+plus `DAC_OVERRIDE` as the single allowed capability addition. The
+NFS server still squashes — that's outside our control — but the
+kata-guest DAC check is now bypassable for the agent, so the
+in-guest open/write succeeds and lands on the share as UID 99.
+
+Containment remains the kata VM boundary. `DAC_OVERRIDE` grants
+"ignore file-permission checks inside the guest filesystem view"
+and nothing else: no `SYS_ADMIN`, no `NET_*`, no `SETUID`, no
+module load. The cluster-wide blast radius of this manifest is
+exactly the one ServiceAccount group bound below.
+
+`priority: 9` is one below stock `anyuid` (10), so OpenShift's
+SCC picker does not pre-empt anyuid for pods that don't need
+the cap — the binding below is what actually grants its use.
+*/}}
+apiVersion: security.openshift.io/v1
+kind: SecurityContextConstraints
+metadata:
+  name: {{ include "platform.fullname" . }}-agent-kata
+  labels:
+    {{- include "platform.labels" . | nindent 4 }}
+  annotations:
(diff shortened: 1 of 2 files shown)


xjacka:
🛡️ **Humr** — Code Review @ `7097204`

## PR #196: feat(helm): ship custom SCC granting DAC_OVERRIDE for kata agents
**Author:** jezekra1 | **Branch:** feat/openshift-custom-scc → main | **Changes:** +114 −21 (2 files)

### Summary
Replaces the stock `anyuid` SCC binding with a custom OpenShift SCC (`<release>-platform-agent-kata`) that mirrors `anyuid` but adds `DAC_OVERRIDE` as the sole extra capability. This resolves EACCES failures on NFS-backed home PVCs where IBM Cloud File Storage enforces `root_squash`, squashing root writes to UID 99 (`nobody`); the added capability allows the kata-guest kernel to bypass DAC checks and complete the writes. Inline documentation thoroughly explains the NFS `root_squash` / kata containment argument.

### Findings
- 🟡 **Warning:** `RoleBinding` references `system:openshift:scc:{{ include "platform.fullname" . }}-agent-kata` — a `ClusterRole` that OpenShift auto-generates when the SCC is created. If Helm applies the `RoleBinding` before the SCC (and thus before OpenShift auto-creates the ClusterRole), a fresh install may fail with a dangling role reference. The in-file `---` separator has no effect on Helm's apply sequence — Helm sorts by Kind, and `SecurityContextConstraints` is not in Helm 3's built-in kind weight table. In practice OpenShift creates the ClusterRole synchronously, but this is an undocumented implicit assumption that could fail in edge cases or future OCP versions. Consider using `helm.sh/hook-weight` annotations to enforce SCC-before-RoleBinding ordering. (`deploy/helm/platform/templates/agent-scc.yaml`)
- 🟡 **Warning:** `values.yaml` instructs operators to manually override `controller.agent.base.containerSecurityContext` to add `capabilities.add: ["DAC_OVERRIDE"]` — but the chart templates do not enforce or default this. If a user enables `openshift.scc.anyuid.enabled: true` without the container-level override, OpenShift permits the capability (SCC allows it) but the container never requests it, so the NFS EACCES problem persists with no clear error. Consider wiring the `containerSecurityContext` capability add into chart defaults, or adding a `fail` guard when the SCC flag is enabled but the capability is absent. (`deploy/helm/platform/values.yaml:131`)
- 🟢 **Suggestion:** `allowPrivilegeEscalation: true` matches stock `anyuid`, but confirm it is genuinely needed — if no in-guest process relies on `setuid`/`setgid` binaries, setting this to `false` would tighten the posture without breaking the NFS fix. (`deploy/helm/platform/templates/agent-scc.yaml`)
- ✅ **Looks good:** Minimal capability escalation (only `DAC_OVERRIDE`, no `SYS_ADMIN`, `NET_*`, `SETUID`). `priority: 9` correctly stays below stock `anyuid` (10), avoiding SCC pre-emption for pods that don't need the cap. Namespace-scoped `RoleBinding` (not `ClusterRoleBinding`) limits blast radius. Inline documentation is thorough and accurate.

### Verdict
COMMENT — The approach and documentation are excellent; two issues need attention before merge: the implicit ordering dependency on OpenShift's SCC → ClusterRole auto-generation (potential first-install fragility), and the missing chart-level enforcement of the `containerSecurityContext` override that actually activates the capability in the container.

---

jezekra1:
Pivoting to non-root. The custom SCC + DAC_OVERRIDE is only required if root in the kata guest is a hard requirement, which is no longer the case. Closing in favor of running agents as the image USER (65532) under stock OpenShift SCCs.

Issue #86, opened 2026-05-05
`dam shell` — attach local terminal to an agent


tomkis:
## Problem

The whole point of moving Claude Code into DAM is that a developer should not have to give up the terminal experience they're used to. Today the only way to interact with an agent from outside the cluster is the web UI, which can't deliver an interactive TUI from a CLI. The POC prototyped this against a shared secret; it works, but isn't shippable.

## Proposed solution

`dam shell <instance>` attaches the developer's local terminal to the agent's terminal-mode session. Keystrokes, resizes, and signals flow to the harness; output renders locally as if the harness were running on the laptop. Closing the CLI does not kill the session — re-running `dam shell` lands back in the same place.

### Dependencies

Part of #72. Blocked on:
- #79 (`dam` CLI tool)
- #80 (`dam login`)
- #81 (target an instance by name)
- #82 (chat vs terminal session modes)
- #83 (web terminal view — shares the underlying session backend)

JanPokorny:
"shell" is not a good name, since we are not providing general shell access, only opening the TUI

tomkis:
@JanPokorny i think long term goal is to open whole shell. Maybe we can do something like `dam shell --attach instance_id` ? Having `attach` mandatory

PetrBulanek:
Btw, this one should by unblocked by now.

JanPokorny:
@PetrBulanek yup, I'm on it

@tomkis technically we can run anything, `claude` or `sh`, does not matter much -- we just make an assumption that we are running an agent TUI session in order to be able to do things like "convert this session to Chat session on web", which would make little sense for `sh` sessions. But the groundwork is actually fully general, we are forwarding the whole PTY, basically a full remote shell.

JanPokorny:
(misclick close)

PR #206, opened 2026-05-14
feat(cli): add dam chat command to connect terminal to agent TUI

JanPokorny:
## Summary

- Adds `dam chat <instance>` CLI command that connects the local terminal to an agent's interactive TUI over WebSocket (#86)
- Supports `--continue` (most recent terminal session), `--resume <id>` (specific session), and `--reset` (fresh PTY)
- Adds `dam session list <instance>` for listing sessions from the CLI
- Adds cross-tab session mode synchronization via `platform/sessionModeChanged` ACP notification (Redis pub/sub → inject channel → all connected clients)
- Terminal component gains `autoConnect` prop — external mode changes show "Disconnected" overlay; user-initiated opens auto-connect

### CLI architecture
- `sessions-client.ts` — plain HTTP client for tRPC endpoints (no tRPC client dep)
- `terminal-bridge.ts` — raw TTY ↔ binary frame protocol over WebSocket
- `chat-service.ts` — orchestrates instance resolution → session find/create → WS connect

### Server-side changes
- `sessions-service.setMode()` now closes terminal WS, resets ACP session, and publishes mode-change notification
- `terminal-relay` tracks active clients per session for clean teardown on mode switch
- `acp-runtime` gains `resetSession()` to close agent-side session and clear logs

### UI changes
- `useAcpConfigCache` simplified to only handle config concerns (modes/models/options)
- `useAcpSession` owns session mode change handling via `modeChangeRef` wrapper — no more cross-hook ref-bridging
- `resumeSession` reconciles stale sidebar mode against the server for all session opens

## Test plan

- [x] `dam chat <name>` connects to terminal, keystrokes flow, output renders
- [x] Ctrl+C in harness TUI works (forwarded, not intercepted)
- [x] Close terminal: prints command to reconnect to same session
- [x] `dam chat nonexistent` — clear error message
- [x] `dam chat --resume <id> --reset <name>` — kills old PTY, fresh start
- [x] `dam session list <name>` — shows table of sessions
- [x] Two UI tabs: switch mode in tab 1 → tab 2 reflects the change
- [x] Tab 2 clicking a session whose mode was changed by tab 1 opens in the correct (current) mode
- [x] External terminal switch shows "Disconnected" overlay; user-initiated open auto-connects

### DIFF
--- a/packages/api-server/src/apps/api-server/app.ts
+++ b/packages/api-server/src/apps/api-server/app.ts
@@ -27,6 +27,7 @@ import {
 } from "../../modules/channels/infrastructure/telegram-threads-repository.js";
 import { createAcpRelay } from "./acp-relay.js";
 import { createTerminalRelay } from "./terminal-relay.js";
+import { getSessionMode } from "../../modules/sessions/infrastructure/sessions-repository.js";
 import { createOAuthRoutes } from "./oauth.js";
 import { mountBrandIconRoutes } from "./brand-icon.js";
 import { createOAuthAppRegistry } from "../../modules/connections/infrastructure/oauth-apps.js";
@@ -389,9 +390,8 @@ export function startApiServerApp(deps: ApiServerAppDeps) {
     const { sessions } = composeSessionsModule({
       db, namespace: config.namespace, isOwnedInstance, isOwnedSchedule,
       closeTerminalSession: terminalRelay.closeSession,
-      resetAcpSession: (instanceId, sessionId) => {
-        fetch(`http://${podBaseUrl(instanceId, config.namespace)}/api/sessions/${encodeURIComponent(sessionId)}/reset`, { method: "POST" }).catch(() => {});
-      },
+      resetAcpSession: (instanceId, sessionId) =>
+        fetch(`http://${podBaseUrl(instanceId, config.namespace)}/api/sessions/${encodeURIComponent(sessionId)}/reset`, { method: "POST" }).catch(() => {}),
       notifyModeChange: (instanceId, sessionId, mode) => {
         const frame = JSON.stringify({
           jsonrpc: "2.0",
@@ -466,7 +466,9 @@ export function startApiServerApp(deps: ApiServerAppDeps) {
     { resolve: (id) => instancesRepo.resolveIdentity(id).then((r) => r ? { ownerSub: r.owner, agentId: r.agentId } : null) },
   );
 
-  const terminalRelay = createTerminalRelay(config.namespace, instancesRepo);
+  const terminalRelay = createTerminalRelay(config.namespace, instancesRepo, {
+    getSessionMode: getSessionMode(db),
+  });
 
   const server = serve({ fetch: app.fetch, port: config.port }, () => {
     process.stderr.write(`api-server listening on http://localhost:${config.port}\n`);
--- a/packages/api-server/src/apps/api-server/app.ts
+++ b/packages/api-server/src/apps/api-server/app.ts
@@ -47,6 +47,7 @@ import {
   type ApprovalsRelayService,
   type WrapperFrameSender,
 } from "./../../modules/approvals/compose.js";
+import { injectChannelOf } from "./../../modules/approvals/infrastructure/acp-frames.js";
 import {
   composeEgressRulesModule,
   createConnectionRulesSyncAdapter,
@@ -387,6 +388,18 @@ export function startApiServerApp(deps: ApiServerAppDeps) {
     const { schedules, isOwnedSchedule } = composeSchedulesModule(api, config.namespace, user.sub);
     const { sessions } = composeSessionsModule({
       db, namespace: config.namespace, isOwnedInstance, isOwnedSchedule,
+      closeTerminalSession: terminalRelay.closeSession,
+      resetAcpSession: (instanceId, sessionId) => {
+        fetch(`http://${podBaseUrl(instanceId, config.namespace)}/api/sessions/${encodeURIComponent(sessionId)}/reset`, { method: "POST" }).catch(() => {});
+      },
+      notifyModeChange: (instanceId, sessionId, mode) => {
+        const frame = JSON.stringify({
+          jsonrpc: "2.0",
+          method: "platform/sessionModeChanged",
+          params: { sessionId, mode },
+        });
+        redisBus.publish(injectChannelOf(instanceId), frame).catch(() => {});
+      },
     });
     const skills = composeSkillsModule(api, config.namespace, user.sub, db, seedSources, config.brand.name);
     const grants = createAgentGrantsPort(k8sClient, user.sub);
--- a/packages/api-server/src/apps/api-server/terminal-relay.ts
+++ b/packages/api-server/src/apps/api-server/terminal-relay.ts
@@ -8,9 +8,25 @@ import { LAST_ACTIVITY_KEY, ACTIVE_SESSION_KEY } from "../../modules/agents/infr
 const ACTIVITY_DEBOUNCE_MS = 30_000;
 const PENDING_BUFFER_MAX_BYTES = 1 * 1024 * 1024;
 
-export function createTerminalRelay(namespace: string, repo: InstancesRepository) {
+export interface TerminalRelay {
+  handleUpgrade(req: IncomingMessage, socket: Duplex, head: Buffer, instanceId: string): void;
+  closeSession(sessionId: string): void;
+}
+
+export function createTerminalRelay(namespace: string, repo: InstancesRepository): TerminalRelay {
   const wss = new WebSocketServer({ noServer: true, perMessageDeflate: false });
   const lastActivity = new Map<string, number>();
+  const activeClients = new Map<string, WebSocket>();
+
+  function closeSession(sessionId: string) {
+    const ws = activeClients.get(sessionId);
+    if (!ws) { activeClients.delete(sessionId); return; }
+    if (ws.readyState === WebSocket.CONNECTING) ws.terminate();
+    else if (ws.readyState === WebSocket.OPEN) {
+      try { ws.close(1000, "session mode changed"); } catch { ws.terminate(); }
+    }
+    activeClients.delete(sessionId);
+  }
 
   function handleUpgrade(req: IncomingMessage, socket: Duplex, head: Buffer, instanceId: string) {
     const url = new URL(req.url!, `http://${req.headers.host}`);
@@ -19,6 +35,15 @@ export function createTerminalRelay(namespace: string, repo: InstancesRepository
 
     wss.handleUpgrade(req, socket, head, (client) => {
       client.on("error", () => { try { client.terminate(); } catch {} });
+
+      const prev = activeClients.get(sessionId);
+      if (prev) {
+        if (prev.readyState === WebSocket.CONNECTING) prev.terminate();
+        else if (prev.readyState === WebSocket.OPEN) {
+          try { prev.close(1000, "superseded"); } catch { prev.terminate(); }
+        }
+      }
+      activeClients.set(sessionId, client);
 
       const pending: { data: Buffer; isBinary: boolean }[] = [];
       let pendingBytes = 0;
@@ -78,6 +103,7 @@ export function createTerminalRelay(namespace: string, repo: InstancesRepository
           });
 
           client.on("close", () => {
+            if (activeClients.get(sessionId) === client) activeClients.delete(sessionId);
             repo.patchAnnotation(instanceId, ACTIVE_SESSION_KEY, "").catch(() => {});
             if (upstream.readyState === WebSocket.OPEN) upstream.close();
           });
@@ -89,5 +115,5 @@ export function createTerminalRelay(namespace: string, repo: InstancesRepository
     });
   }
 
-  return { handleUpgrade };
+  return { handleUpgrade, closeSession };
 }
--- a/packages/api-server/src/apps/api-server/terminal-relay.ts
+++ b/packages/api-server/src/apps/api-server/terminal-relay.ts
@@ -13,7 +13,9 @@ export interface TerminalRelay {
   closeSession(sessionId: string): void;
 }
 
-export function createTerminalRelay(namespace: string, repo: InstancesRepository): TerminalRelay {
+export function createTerminalRelay(namespace: string, repo: InstancesRepository, deps?: {
+  getSessionMode?: (sessionId: string) => Promise<string | null>;
+}): TerminalRelay {
   const wss = new WebSocketServer({ noServer: true, perMessageDeflate: false });
   const lastActivity = new Map<string, number>();
   const activeClients = new Map<string, WebSocket>();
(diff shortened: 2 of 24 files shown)


xjacka:
🛡️ **Humr** — Code Review @ `0778f48`

## PR #206: feat(cli): add dam chat command to connect terminal to agent TUI
**Author:** JanPokorny | **Branch:** feat/cli-attach#locki-wp4kww7z → main | **Changes:** +556 −124 (27 files)

### Summary
Adds a `dam chat <instance>` CLI command that connects a local terminal to an agent's interactive TUI via WebSocket, with session strategies (new / continue / resume). Also adds `dam session list`, renames `dam instances` → `dam instance`, and wires up bidirectional mode-switch signaling (chat ↔ terminal) through API server, ACP runtime, and UI.

### Findings
- 🟡 **Warning:** `chat-service.ts:752` — Auth token is embedded in the WebSocket URL query string (`?token=...`). The `ws` npm library supports custom headers, so the CLI can send `Authorization: Bearer` instead. Query-string tokens appear in server access logs and process inspection, which is a real exposure in a CLI context (unlike a browser where there's no WebSocket `Authorization` header).
- 🟡 **Warning:** `chat-view.tsx` (`handleSwitchMode`, ~line 202) — `setSessionMode(target)` is called *before* `await api.sessions.setMode.mutate(...)`. If the API call throws (network error, 4xx), the UI stays in the target mode with no rollback, creating an inconsistent state. Move the store update after the await or add a catch that resets to the previous mode.
- 🟡 **Warning:** `acp-runtime.ts:17–29` (`resetSession`) — `sessionLogs` and `channelCursors` are always cleared even when `sessionCloseSupported` is false and no JSON-RPC close is sent to the agent. If the agent is mid-stream on that session, it will keep writing to a cursor/log that no longer exists while the UI sees it as closed. Consider only clearing these *after* confirming the session is detached, or accept and document the asymmetry.
- 🟢 **Suggestion:** `chat-service.ts:687` — `compat.check()` failures (non-OK results like network errors) are silently skipped via the `compat.ok && ...` guard, allowing a potentially incompatible CLI to proceed. A soft warn or log would help operators diagnose version skew.
- 🟢 **Suggestion:** `terminal-relay.ts:closeSession` — `ws.close()` is only attempted when `readyState === WebSocket.OPEN`. A client in `CONNECTING` state is never closed and stays in `activeClients` until superseded by the next connection. Worth guarding: if `CONNECTING`, call `ws.terminate()` so the slot is freed immediately.
- ✅ **Looks good:** `activeClients` deduplication — closing a "superseded" existing connection on new connect is correct and prevents phantom sessions accumulating.
- ✅ **Looks good:** `modeChangeRef.current = ...` assignment in the component body is the right pattern to avoid stale closures without adding an unnecessary `useEffect` dependency.
- ✅ **Looks good:** `terminalPaused` + `autoConnect` cleanly separates CLI-initiated mode switch (UI shows paused terminal, no re-connect) from user-initiated connect.
- ✅ **Looks good:** `resetSession` is properly guarded with `agent && !agentExited && sessionCloseSupported` before sending JSON-RPC.

### Verdict
COMMENT — Solid implementation overall; the optimistic mode-switch state update in `chat-view.tsx` and the token-in-URL exposure are the two items worth addressing before merge.

---

xjacka:
🛡️ **Humr** — Code Review @ \`5dc5cb9\`

## PR #206: feat(cli): add dam chat command to connect terminal to agent TUI
**Author:** JanPokorny | **Branch:** feat/cli-attach#locki-wp4kww7z → main | **Changes:** +571 −126 (27 files)

### Summary
Adds `dam chat <instance>` CLI command connecting a local terminal to an agent's interactive TUI via WebSocket, with session strategies (new / continue / resume). Also adds `dam session list`, renames `dam instances` → `dam instance`, and wires up bidirectional mode-switch signaling (chat ↔ terminal) through API server, ACP runtime, and UI. New commits since last review add: `terminalPaused` state + `autoConnect` prop for externally-triggered mode changes, `resetSession` in acp-runtime, `closeSession` in terminal-relay, and `notifyModeChange` via Redis bus.

### Changes since last review
Previous HEAD: `0778f48` (2026-05-14T07:38:06Z) — verdict COMMENT

- ✅ **Fixed:** `handleSwitchMode` in `chat-view.tsx` — now saves `previousMode`, sets new mode optimistically, and on API error restores `previousMode` with an error toast. Addresses the prior Warning.
- ✅ **Fixed:** `terminal-relay.ts:closeSession` — now handles `CONNECTING` state with `ws.terminate()`. Addresses the prior Suggestion.
- 🔁 **Still present:** Token in WebSocket URL query string (`chat-service.ts` wsUrl). Carried over from previous review.
- 🔁 **Acknowledged:** `resetSession` asymmetry in acp-runtime is now documented with an explicit comment ("Always clear client-side state even if the agent didn't accept the close").

### Findings
- 🟡 **Warning:** `sessions-service.ts:setMode` — `closeTerminalSession` and `resetAcpSession` are called whenever `mode !== SessionMode.Terminal`. A same-mode no-op call with `mode="chat"` on an already-chat session would fire `resetAcpSession`, which sends `session/close` to the agent runtime and clears session logs/cursors — terminating a live chat session. Current callers avoid this (toggle is binary), but the guard is missing for safety. Consider checking the prior mode before firing the side effects. (`packages/api-server/src/modules/sessions/services/sessions-service.ts:~67`)
- 🟡 **Warning:** `chat-service.ts` — Token in WebSocket URL query string (`?token=...`) is still present. The `ws` npm package supports `{ headers: { Authorization: 'Bearer ...' } }`. Query-string tokens appear in server access logs and process inspection output. (`packages/cli/src/modules/chat/services/chat-service.ts:~119`)
- 🟢 **Suggestion:** `use-acp-session.ts:resumeSession` — After loading history, an extra `api.sessions.list.query(...)` call is made to detect the session's current mode. This adds a round-trip on every resume. The mode could be carried in the history response or from the sidebar list query already in flight — worth considering whether this fetch can be avoided. (`packages/ui/src/modules/sessions/hooks/use-acp-session.ts:~105`)
- ✅ **Looks good:** `terminalPaused` + `autoConnect={!terminalPaused}` pattern cleanly handles externally-triggered mode changes — Terminal renders as "disconnected" without auto-connecting, and `handleReconnect` re-enables connection on explicit user action.
- ✅ **Looks good:** `modeChangeRef.current = ...` assignment in component body correctly captures the latest `resumeSession` closure without an unnecessary `useEffect` dependency.
- ✅ **Looks good:** `mobileResumeSession` now uses `resumeSession` uniformly (removes the `setSessionId` fast-path for terminal sessions), correctly going through ACP history load.

### Verdict
COMMENT — The mode-switch rollback is now correct and terminal-relay CONNECTING state is handled. Two remaining items: the setMode same-mode footgun (could silently kill a live chat session) and token in WebSocket URL.

---

packages/api-server/src/apps/api-server/app.ts
```diff
@@ -387,6 +388,18 @@ export function startApiServerApp(deps: ApiServerAppDeps) {
     const { schedules, isOwnedSchedule } = composeSchedulesModule(api, config.namespace, user.sub);
     const { sessions } = composeSessionsModule({
       db, namespace: config.namespace, isOwnedInstance, isOwnedSchedule,
+      closeTerminalSession: terminalRelay.closeSession,
+      resetAcpSession: (instanceId, sessionId) => {
+        fetch(`http://${podBaseUrl(instanceId, config.namespace)}/api/sessions/${encodeURIComponent(sessionId)}/reset`, { method: "POST" }).catch(() => {});
```
PetrBulanek: **`resetAcpSession` isn't awaited → first chat prompt after a switch can fail**

This `fetch` returns void, and `setMode` fires `notifyModeChange` right after without awaiting it. Redis usually beats the cross-pod HTTP — so a recipient tab fires `loadHistory` → `loadSession` while the runtime still has cached log + metadata, hits the cache short-circuit ([acp-runtime.ts:892](packages/agent-runtime/src/modules/acp/services/acp-runtime.ts#L892)), and never forwards to the agent. Then `/reset` lands, `session/close` releases the agent's SDK session, and the user's first `session/prompt` hits "session not found."

```ts
resetAcpSession: async (instanceId, sessionId) => {
  await fetch(`http://${podBaseUrl(instanceId, config.namespace)}/api/sessions/${encodeURIComponent(sessionId)}/reset`, { method: "POST" }).catch(() => {});
},
```

Return `Promise<void>` from the type and `await` it in [`sessions-service.ts:72`](packages/api-server/src/modules/sessions/services/sessions-service.ts#L72) before `notifyModeChange`.

packages/ui/src/modules/sessions/views/chat-view.tsx
```diff
@@ -151,17 +153,24 @@ export function ChatView() {
...
+      } catch {
+        setSessionMode(previousMode);
+        useStore.getState().showToast({ kind: "error", message: "Failed to switch session mode" });
+        return;
+      }
     }
-    if (target === SessionMode.Terminal) terminalFreshRef.current = true;
-    setSessionMode(target);
     if (target === SessionMode.Chat) {
       if (sessionId) resumeSession(sessionId, { expectNotFound: sessionMode === SessionMode.Terminal });
```
PetrBulanek: **Toggling terminal → chat can silently delete the session**

A session created in terminal mode (e.g. via `dam chat`) has no chat history on the agent. After the toggle, `loadSession` returns not-found → `expectNotFound: true` triggers `api.sessions.delete` and the row vanishes. User clicked a mode toggle, lost the session.

`expectNotFound` is meant for "another client concurrently deleted the row" — not for "no chat history yet."

```ts
if (sessionId) resumeSession(sessionId);
```

Same change needed in [`use-acp-session.ts:144`](packages/ui/src/modules/sessions/hooks/use-acp-session.ts#L144) (cross-tab broadcast path).

packages/api-server/src/apps/api-server/terminal-relay.ts
```diff
@@ -8,9 +8,25 @@ import { LAST_ACTIVITY_KEY, ACTIVE_SESSION_KEY } from "../../modules/agents/infr
...
+    const ws = activeClients.get(sessionId);
+    if (!ws) { activeClients.delete(sessionId); return; }
+    if (ws.readyState === WebSocket.CONNECTING) ws.terminate();
+    else if (ws.readyState === WebSocket.OPEN) {
+      try { ws.close(1000, "session mode changed"); } catch { ws.terminate(); }
+    }
+    activeClients.delete(sessionId);
+  }
 
   function handleUpgrade(req: IncomingMessage, socket: Duplex, head: Buffer, instanceId: string) {
```
PetrBulanek: **`handleUpgrade` attaches a PTY based purely on the URL — no check that the session is in terminal mode**

Mode sync between clients relies entirely on the `platform/sessionModeChanged` Redis broadcast. Pub/sub is best-effort and doesn't replay on reconnect, so a tab whose ACP WS was down at publish time (laptop sleep, network blip, api-server rollout) keeps a stale local `sessionMode = "terminal"`. Commit `3a63a46` removed the `resumeSession` reconciliation, so the broadcast is now the only correction path.

Concrete sequence:
1. Tab A is on a terminal session, alive.
2. Laptop sleeps → Tab A's WSes die.
3. Someone else flips the session to chat. Broadcast goes out; Tab A misses it.
4. Laptop wakes. Tab A's local `sessionMode` is still `terminal`; the UI shows the "Disconnected" overlay (terminal WS is dead).
5. User clicks **Reconnect**. The server accepts the upgrade and spawns a PTY against a session the DB says is chat.

End state: DB says chat, Tab A shows terminal, every other client shows chat. Same exposure applies to a CLI with a stale `dam session list` cache.

PetrBulanek (changes requested):
Still working through the manual QA pass on UI and CLI. From a code-read perspective the change looks solid overall, modulo the three blockers I left inline.

PetrBulanek:
@JanPokorny 

Did a manual pass across the main flows - single-client kick-out works correctly in every combination I tried (CLI ↔ CLI, UI ↔ CLI, UI ↔ UI), in-progress terminal state is preserved when a new client takes over, mode switches propagate across tabs as expected, and background tabs correctly land on the "Disconnected" overlay without stealing the PTY. Feels solid. Really nice work. 👍

PetrBulanek (approved):

tomkis:
> Close terminal + dam chat <name> again — reconnects to same session

@JanPokorny creates neww session for me. 

>  dam chat --reset <name> — kills old PTY, fresh start

As a result this doesn't seem to do anything.

JanPokorny:
@tomkis It's intentional to align with how most agent CLIs work -- no arg to start new session, `--continue` to return to existing one (as long as there is only one terminal session with the agent), `--resume <id>` to resume session (printed in hint after closing session)

packages/cli/src/modules/chat/infrastructure/sessions-client.ts
```diff
@@ -0,0 +1,32 @@
+import type { SessionView } from "api-server-api";
+import { SessionMode } from "api-server-api";
+import { err, ok, type Result } from "../../../result.js";
+
+type Fail = { kind: "session-failed"; reason: string };
+
+export function createSessionsClient(deps: { host: string; token: string }) {
```
tomkis, line 7: Let's remove this entirely and depend on `api-server-api` instead

packages/cli/src/modules/chat/commands/session-list.ts
```diff
@@ -0,0 +1,42 @@
...
+      if (result.value.length === 0) {
+        process.stderr.write("No sessions.\n");
+        return;
+      }
+
+      process.stdout.write(renderTable(result.value));
+    });
+}
+
+function renderTable(sessions: readonly SessionView[]): string {
```
tomkis, line 33: I'd move this to some cli-helpers (obviously passing the headers as params)

packages/cli/src/modules/chat/services/chat-service.ts
```diff
@@ -0,0 +1,125 @@
...
+        if (!listed.ok) return listed;
+        const target = listed.value.find((s) => s.sessionId === strategy.sessionId);
+        if (!target) return err({ kind: "session-not-found", sessionId: strategy.sessionId });
+        if (target.mode === "chat") {
+          if (!await deps.confirmModeSwitch()) return err({ kind: "mode-switch-declined" });
+          const switched = await sessions.setMode(target.sessionId, instanceId, "terminal");
+          if (!switched.ok) return switched;
+        }
+        sessionId = target.sessionId;
+      }
```
tomkis, line 114: To make this aligned with the tseng arechitecture.

1. This should first fetch sessions (for all cases yes)
2. Then it shoudl assemble domain aggregate (input is DTO - session list)
3. Then perform effectless domain logic - deciding of errors and deducting proper session -> this is core piece that can then be easily tested in isolation.

Creating bridge from now on here is correct.

packages/cli/src/modules/chat/compose.ts
```diff
@@ -0,0 +1,38 @@
...
+}: {
+  compatService: CompatService;
+  configService: ConfigService;
+  tokenProvider: TokenProvider;
+  createInstanceService: (host: string) => InstanceService;
+  serverEnvVar: string;
+}): { commands: ReadonlyArray<Command> } {
+  const chatService = createChatService({
+    compatService, configService, tokenProvider, createInstanceService,
+    confirmModeSwitch: () => new Promise((resolve) => {
```
tomkis, line 21: Lerts move this to the service. working with the CLI is a port.

packages/cli/src/compose.ts
```diff
@@ -49,6 +50,14 @@ export function compose(opts: ComposeOptions = {}): Command {
     templateService: template.exports.createService,
   });
 
+  const chat = composeChatModule({
+    compatService: cli.services.compatService,
+    configService: cli.services.configService,
+    tokenProvider: auth.exports.tokenProvider,
+    createInstanceService: instance.exports.createService,
+    serverEnvVar: "DAM_SERVER",
```
tomkis, line 58: This is not Env var

packages/cli/src/modules/chat/services/chat-service.ts
tomkis: but truth to be told - ideally i'd move this business logic to API server

packages/cli/src/modules/chat/services/chat-service.ts
```diff
@@ -0,0 +1,125 @@
...
+    const tok = await deps.tokenProvider.getValidAccessToken(host);
+    if (!tok.ok) return err({ kind: "auth-required" as const, reason: tok.error.kind });
+
+    return ok({
+      host, token: tok.value, instanceId: resolved.value.id,
+      sessions: createSessionsClient({ host, token: tok.value }),
+    });
+  }
+
+  function wsUrl(host: string, instanceId: string, token: string, sessionId: string, reset?: boolean) {
```
tomkis, line 69: I'd much rather have this dictated by server.

PR #197, opened 2026-05-13
fix(ui): persist new sessions eagerly and refresh titles on re-entry

kapetr:
## Summary

- **Persist new sessions to the DB eagerly (fixes #21).** `sessions.create` now runs right after `conn.newSession` returns instead of after the first prompt resolves. A refresh or navigation mid-turn no longer leaves the session missing from the sidebar.
- **"(no title)" placeholder in muted style** when the harness hasn't named the session yet, replacing the session-id slice that read like a title.
- **`refetchOnMount: "always"`** on the sessions list query so returning to a chat view picks up titles set after the first turn without a manual refresh.

## Test plan

- [ ] Open a new chat, send a prompt, while it's still streaming refresh the page → session is in the sidebar.
- [ ] Same scenario but navigate to agents list and back instead of refreshing → session is in the sidebar.
- [ ] Start a chat with no title yet — sidebar row reads `(no title)` in muted italic, not a 12-char id slice.
- [ ] After the first turn finishes and the harness assigns a title, navigate away and back → updated title is visible without pressing Refresh.
- [ ] Existing sessions (resume path) still work — title rendering, mode badge, delete confirm dialog all unchanged.

Issue #26, opened 2026-05-04
Make session loading and live streaming reliable across reconnects, tab switches, and session changes


pilartomas:
## Problem

Opening a chat session and watching it stream is unreliable in several overlapping ways. Users see the symptoms; the underlying cause is the same architectural drift across the load path, the live path, and the runtime cache.

Concrete situations where it breaks today:

- **Mid-stream WebSocket drop** (mobile screen-off, network blip, foreground/background swap on iOS): when the tab reconnects, any agent output produced during the gap is silently lost from the conversation. The user can keep typing, but they don't see the missed messages until they manually refresh.
- **Clicking the same session in the sidebar while it's streaming**: the first chunk emitted around the moment of the click goes missing. Streaming continues afterwards, but the user is left with a hole in the conversation.
- **Coming back to a session that's been idle for a while**: the messages can come back chopped — the user's own message disappears, tool calls render in a different state than they did live, parts of assistant responses are missing. Refreshing doesn't fix it.
- **Two tabs viewing the same session**: one tab can stop receiving updates after some interaction in the other tab. Recovery requires a refresh.
- **General navigation-induced glitches**: clicking around in the sidebar can leave the chat in a state with brief flickers, partial content, or stale state — the conversation looks subtly broken.

Why it matters: the chat is the primary surface of the product. Every interaction with it currently has a small chance of leaving the user with a conversation that doesn't match what the agent actually said. That erodes trust and turns "click a session" into something users second-guess. We've patched individual symptoms several times and each fix has surfaced another one.

## Proposed solution

Treat the runtime as the single source of truth for a session's history and stream the user's tab into a faithful mirror of it. The user-visible promise:

- **Open a session in any tab → see the complete history**, in the same shape it was originally streamed (including the user's own messages and any in-progress tool calls).
- **Watch it stream → never miss a chunk** regardless of brief disconnects, screen-off intervals, or sidebar clicks.
- **Switch sessions → arrive cleanly** at the new one without flicker or partial content from the previous.
- **Reconnect after being away briefly → resume in place** without losing the conversation that was already on screen and without re-rendering from scratch.
- **Multiple tabs on the same session → stay in sync** independently; one tab's interaction can't break another's stream.

[...]

PR #208, opened 2026-05-14
Feat: move session creation ownership to api server

jjeliga:
What we gain

1. Orphan rows become structurally impossible

This is the biggest one. Today on main, a row appears in the platform DB through two paths:

UI path — useAcpSessionEngagement calls api.sessions.create.mutate after newSession. Held safe by the comment: "engage only ever runs from a user-initiated send".
List path — sessions-service.ts:24's onSessionCreated: (sid) => deps.upsert(...) runs for every session the ACP probe discovers. This is the orphan-creating path that bit you originally (sid 219ba7a3-…).
Both paths are write-on-discovery, neither requires user intent. The "no orphans" property is enforced socially: every caller of newSession on the runtime has to be careful not to leave one stranded. That worked for the original codebase because only one path called newSession. It's already starting to fray: PR #197 needed an explicit comment defending the invariant, and any future preflight/discovery flow would silently break it.

Our change moves the trigger from session-was-created to first-prompt-was-sent. "No prompt, no row" is a hard guarantee, not a code convention. Real merit: the bug class disappears for all callers, including future ones we haven't written yet.

2. External ACP clients can use the platform

Main's contract is "to get a row in the platform DB, the UI must call sessions.create." Any ACP client that connects to api-server, speaks pure ACP, and sends session/new + session/prompt will get a working chat with the agent — but it will be invisible to the sidebar, the schedules system, and anything else that reads from the DB.

That contradicts the stated direction of the codebase (api-server aspirationally ACP-only). Our change makes the relay itself the persistence point, so anything that speaks ACP gets persisted. Real merit: preserves the architectural direction the project committed to. If you change your mind on that direction, this loses value.

3. No silent "agent has session, DB doesn't" split-brain

Main's persist is mutate(...).then(invalidate).catch(toast). If the DB write fails (transient blip, schema constraint, anything), the user sees a toast but the agent has already created the session and started replying. The chat works for that turn; the row never appears; the user can't navigate back to it.

Our change holds every session/prompt frame until persistence commits — the first one writes the row, subsequent ones are PG conflict no-ops (~1ms). If the DB write fails, the relay responds with a JSON-RPC error to the prompt itself — the failure surfaces in the chat bubble as a clear "couldn't save this session" instead of a quietly orphaned chat. The persist is idempotent and runs on every prompt, so even a WS reconnect between session/new and the first prompt is covered. No retry, but the failure is loud and atomic with the user's send. Real merit: the failure mode that #197's reviewer (Humr) flagged but didn't fix is structurally closed, and the row stays consistent with agent state across reconnects.

4. WS close-handler race fixed

This one is independent of the persistence story and was a bug you hit directly during repro: "first session worked, second new session - stuck with agent blinking."

Sequence: resetConnection() closes WS₁ → ensureLive() opens WS₂ immediately → WS₁'s close event fires later, asynchronously → its handler does connectionRef.current = null; clearEngagement(); pendingReloadRef.current = true; — clobbering WS₂'s state.

Our 4-line guard checks if (connectionRef.current?.ws !== ws) return;. Real merit: pure correctness fix for a reproducible failure mode. No tradeoff, no architectural cost. Worth doing even in isolation.

5. One source of truth for session-row writes

Today on main, three things write to sessions: sessions.list()'s callback, the UI's tRPC mutate, and sessions.create for scheduled-run paths. After our change:

Relay-persisted — UI sessions, where traffic flows UI → api-server relay → agent pod. The relay writes on first prompt.
Adapter-persisted — channel-driven (Slack, Telegram) and schedule-driven sessions, which open a direct ACP connection to the agent pod and pass onSessionCreated per call (now typed as a discriminated union, so forgetting it on a new-session call is a compile error).

The UI is no longer in the write path. Real merit: smaller surface for bugs, cleaner type contracts, easier to reason about who owns what.

Connected to #26, but not closing it fully


xjacka:
🛡️ **Humr** — Code Review @ \`b054067\`

## PR #208: Feat: move session creation ownership to api server
**Author:** jjeliga | **Branch:** fix/26-session-creation-stability → main | **Changes:** +124 −43 (6 files)

### Summary
Moves session persistence ownership from the UI (`useAcpSessionEngagement`) and ACP-client `onSessionCreated` callbacks to the api-server ACP relay. Sessions are now committed to the platform DB on the first `session/prompt` after `session/new`, preventing sidebar pollution from probe sessions created by `list()`. `onSessionCreated` becomes optional on `createAcpClient`/`createForkAcpClient`.

### Findings
- 🟡 **Warning:** `acp-relay.ts` — If the upstream WebSocket disconnects while `persistSession` is in-flight and the promise resolves `true`, `upstream.send(data)` is silently skipped (readyState check fails) with no error sent to the client. The upstream `close` event will also close the client, but there's a narrow race window where the client is left waiting for a `session/prompt` reply that never arrives. (`packages/api-server/src/apps/api-server/acp-relay.ts:~247`)
- 🟡 **Warning:** `acp-relay.ts` — `sidsAwaitingFirstPrompt` and `persistencePromises` are per-WebSocket-connection Maps. If a client reconnects between receiving the `session/new` response and sending the first `session/prompt`, the new connection starts with empty Maps and won't track the session for persistence. `upsertSession` being idempotent doesn't help when the row hasn't been written yet — the reconnect's first prompt silently bypasses persistence, leaving the session without a DB row. (`packages/api-server/src/apps/api-server/acp-relay.ts:~175–200`)
- 🟢 **Suggestion:** `sessions-service.ts:list()` — The ACP `createAcpClient` call is now made without `onSessionCreated`, so sessions that haven't received a `session/prompt` yet won't appear in the sidebar. This is the intended behavior, but a note in the ADR or inline that the sidebar reflects "sessions with at least one prompt" would help future readers understand the intentional gap.
- ✅ **Looks good:** `use-acp-connection.ts` — The stale close-handler guard (`if (connectionRef.current?.ws !== ws) return`) correctly prevents a superseded WebSocket from clearing live connection state on close.
- ✅ **Looks good:** `onSessionCreated?.(sessionId)` — Optional chaining is the right fix; callers that don't need a session callback should not be forced to pass a no-op.
- ✅ **Looks good:** `PersistSession` type is injected via the composition root, keeping `createAcpRelay` testable without a real DB dependency.

### Verdict
COMMENT — Session persistence logic is correct on the happy path. The reconnect-between-new-and-prompt edge case (finding 2) is a real scenario that could leave sessions without DB rows; worth adding a fallback or explicit note. Finding 1 (upstream-disconnect race during persistence) is a narrower timing window.

---

packages/api-server/src/apps/api-server/acp-relay.ts
```diff
@@ -178,32 +195,70 @@ export function createAcpRelay(
...
-                ).catch(() => {});
+            // Persist on every session/prompt — upsertSession is idempotent on
+            // conflict, so subsequent prompts on the same sid are PG no-ops.
+            // Holding the frame until commit keeps DB row + agent state atomic
+            // and makes the persist robust across WS reconnects.
+            if (isRequest(parsed) && parsed.method === "session/prompt") {
+              const sid = extractRequestSessionId(parsed);
+              if (sid) {
+                const requestId = parsed.id;
+                persistSession(sid, instanceId).then(
```
jezekra1, line 220: nitpick (not for current PR) - candidate for refactor:
- too much logic inside `wss.handleUpgrade`, function becomes unreadable
- use modern await/try/except instead of then/catch

packages/api-server/src/apps/api-server/acp-relay.ts
```diff
@@ -49,6 +49,17 @@ function isResponse(msg: unknown): msg is JsonRpcResponse {
   return (typeof m.id === "number" || typeof m.id === "string") && (m.result !== undefined || m.error !== undefined);
 }
 
+function isRequest(msg: unknown): msg is JsonRpcRequest {
+  if (typeof msg !== "object" || msg === null) return false;
+  const m = msg as JsonRpcRequest;
+  return typeof m.method === "string" && (typeof m.id === "number" || typeof m.id === "string");
```
jezekra1: is this really the best way to identify request vs. response in ACP?
the `isResponse` above is also suspicious

jezekra1 (approved):
we can do some tech-debt refactoring separately

packages/ui/src/modules/sessions/hooks/use-acp-session-engagement.ts
```diff
@@ -68,23 +70,24 @@ export function useAcpSessionEngagement(
...
+        type: SessionType.Regular,
+        mode: SessionMode.Chat,
+        createdAt: new Date().toISOString(),
+        scheduleId: null,
+        title: null,
+        updatedAt: null,
+      };
+      const prepend = (prev: SessionView[] | undefined) => (prev ? [stub, ...prev] : [stub]);
+      queryClient.setQueryData<SessionView[]>(acpSessionsKeys.list(selectedInstance, false), prepend);
+      queryClient.setQueryData<SessionView[]>(acpSessionsKeys.list(selectedInstance, true), prepend);
```
kapetr: **Suggestion:** collapse the `prepend` helper + two `setQueryData` calls into a single `setQueriesData` with a partial-key  filter. 
                                                                                                                                
  Add a level to `acpSessionsKeys` that names the shared prefix:
                                                                                                                                
  ```ts                                                                                                                         
  export const acpSessionsKeys = {
    all: ["acp-sessions"] as const,                                                                                             
    instanceLists: (instanceId: string | null) =>                       
      [...acpSessionsKeys.all, instanceId] as const,
    list: (instanceId: string | null, includeChannel: boolean) =>                                                               
      [...acpSessionsKeys.instanceLists(instanceId), { includeChannel }] as const,
  };
```
       
Then the optimistic insert becomes:

```ts
  queryClient.setQueriesData<SessionView[]>(
    { queryKey: acpSessionsKeys.instanceLists(selectedInstance) },
    (prev) => [stub, ...(prev ?? [])],                                                                                          
  );

kapetr (approved):

packages/api-server/src/apps/api-server/acp-relay.ts
jjeliga: For JSON-RPC 2.0 there isn't a much cleaner option — there's no tag field. The standard approach is to discriminate by the presence of method (request/notification) vs result/error (response).

PR #214, opened 2026-05-14
feat(controller): disableDns + iptablesInit egress lockdown

jezekra1:
## Summary

closing threat (DNS exfil via CoreDNS upstream forwarder) by adding two layered, defense-in-depth controls on the agent pod:

- `controller.agent.base.disableDns: true` (default) — drops the cluster-DNS allow rules from the per-pair agent egress NetworkPolicy. Controller injects a `hostAliases` entry mapping `<pair>-gateway` → the gateway Service ClusterIP so `HTTPS_PROXY` still resolves.
- `controller.agent.base.iptablesInit: { enabled: true, image: ... }` (default) — privileged init container (`NET_ADMIN`/`NET_RAW`, runs as root, exits) that pins the agent pod's `OUTPUT` chain to:
  - loopback ACCEPT
  - conntrack ESTABLISHED/RELATED ACCEPT
  - one ACCEPT rule for the paired gateway IP:port
  - terminal DROP

  The runtime agent container itself stays unprivileged (no caps, UID 65532).

Discussed in dam-agents/docs#1 (comments [198860967](https://github.ibm.com/dam-agents/docs/issues/1#issuecomment-198860967), [198890079](https://github.ibm.com/dam-agents/docs/issues/1#issuecomment-198890079)).

### Supporting changes

- Gateway Service flipped from headless to ClusterIP — `hostAliases` / iptables need a routable, pod-restart-independent IP. Controller migrates legacy headless Services in-place on reconcile (delete + recreate; `.spec.clusterIP` is immutable).
- New `ensureGatewayService` captures the Service synchronously from the Create response, so the agent SS template gets the ClusterIP on the same reconcile.
- Custom `anyuid-cap-net` SCC for OpenShift (replaces the previous `nonroot` binding). Modeled on built-in `anyuid` + allowed `NET_ADMIN`/`NET_RAW`. Gated by `openshift.scc.anyuidCapNet.enabled` (default ON).
- Fail-closed guard: reconcile requeues rather than rolling a half-configured agent pod when the gateway ClusterIP isn't yet readable.

### Verified on a live cluster

```
✓ dns.resolve("google.com")           → BLOCKED: ECONNREFUSED
✓ dns.lookup("google.com")            → BLOCKED: EAI_AGAIN
✓ dns.lookup("<pair>-gateway")        → OK (via /etc/hosts hostAliases)
✓ curl --noproxy '*' http://1.1.1.1   → Connection timed out (iptables DROP)
✓ curl $API_SERVER_URL/healthz        → HTTP 403 (chain works: agent→gateway→harness)
```

## Known PoC limitations

- Initial-deploy pod roll: agent + gateway StatefulSets currently template-diff once during the first reconciles because `credentialSecrets` and the cert-manager leaf Secret aren't visible in the informer cache yet on reconcile N. This is pre-existing behavior (unrelated to this PR) and rolls steady-state to 0 restarts. Separate fix: wait for cache sync / leaf-cert creation before applying the gateway SS.
- OpenShift: requires the bundled `anyuid-cap-net` SCC (`openshift.scc.anyuidCapNet.enabled`).
- k3s default CNI (flannel) doesn't enforce NetworkPolicy — the iptables init container is the only enforcement on those clusters. Other CNIs (Calico, Cilium) get both layers.

## Test plan

- [ ] `mise run check` — lint, build, helm render
- [ ] `mise run controller:test`
- [ ] `mise run cluster:install` (fresh) → verify agent + gateway pods stabilize, egress-lockdown init container exits 0
- [ ] In agent pod: `node -e 'require("dns").resolve("google.com",(e,r)=>console.log(e?e.code:r))'` → expect `BLOCKED: ECONNREFUSED`
- [ ] In agent pod: `curl --noproxy '*' http://1.1.1.1` → expect `Connection timed out`
- [ ] In agent pod: harness-proxied call (any external host the gateway is allowed to reach) succeeds via `HTTPS_PROXY`
- [ ] Upgrade test: existing instance with legacy headless gateway Service → controller migrates to ClusterIP on next reconcile

### DIFF
--- a/deploy/helm/platform/templates/_validators.tpl
+++ b/deploy/helm/platform/templates/_validators.tpl
@@ -9,19 +9,18 @@ add it to the include list in `platform.validate`.
 */}}
 
 {{- define "platform.validate" -}}
-{{- include "platform.validate.nonrootRequiresAgentNamespace" . -}}
+{{- include "platform.validate.anyuidCapNetRequiresAgentNamespace" . -}}
 {{- end -}}
 
 {{/*
-The nonroot-v2 RoleBinding is namespaced to `agentNamespace` and grants
-SCC access via the `system:serviceaccounts:<agentNamespace>` group.
-Both are meaningless if `agentNamespace` is empty: the binding lands
-without a namespace, and the group ref matches no SA. Refuse to render.
+The anyuid-cap-net RoleBinding is namespaced to `agentNamespace` and
+grants SCC access via the `system:serviceaccounts:<agentNamespace>`
+group. Both are meaningless if `agentNamespace` is empty.
 */}}
-{{- define "platform.validate.nonrootRequiresAgentNamespace" -}}
-{{- if and .Values.openshift .Values.openshift.scc .Values.openshift.scc.nonroot .Values.openshift.scc.nonroot.enabled -}}
+{{- define "platform.validate.anyuidCapNetRequiresAgentNamespace" -}}
+{{- if and .Values.openshift .Values.openshift.scc .Values.openshift.scc.anyuidCapNet .Values.openshift.scc.anyuidCapNet.enabled -}}
 {{- if not (.Values.agentNamespace | default "" | trim) -}}
-{{- fail "openshift.scc.nonroot.enabled=true requires agentNamespace to be set. The RoleBinding is namespace-scoped and grants SCC access via the system:serviceaccounts:<agentNamespace> group; an empty value makes both meaningless." -}}
+{{- fail "openshift.scc.anyuidCapNet.enabled=true requires agentNamespace to be set. The RoleBinding is namespace-scoped and grants SCC access via the system:serviceaccounts:<agentNamespace> group; an empty value makes both meaningless." -}}
 {{- end -}}
 {{- end -}}
 {{- end -}}
--- a/deploy/helm/platform/templates/_validators.tpl
+++ b/deploy/helm/platform/templates/_validators.tpl
@@ -9,18 +9,18 @@ add it to the include list in `platform.validate`.
 */}}
 
 {{- define "platform.validate" -}}
-{{- include "platform.validate.nonrootCapNetRequiresAgentNamespace" . -}}
+{{- include "platform.validate.anyuidCapNetRequiresAgentNamespace" . -}}
 {{- end -}}
 
 {{/*
-The nonroot-cap-net RoleBinding is namespaced to `agentNamespace` and
+The anyuid-cap-net RoleBinding is namespaced to `agentNamespace` and
 grants SCC access via the `system:serviceaccounts:<agentNamespace>`
 group. Both are meaningless if `agentNamespace` is empty.
 */}}
-{{- define "platform.validate.nonrootCapNetRequiresAgentNamespace" -}}
-{{- if and .Values.openshift .Values.openshift.scc .Values.openshift.scc.nonrootCapNet .Values.openshift.scc.nonrootCapNet.enabled -}}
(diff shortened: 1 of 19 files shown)


Tomas2D (approved):

xjacka:
🛡️ **DAM** — Code Review @ \`c11d92c\`

## PR #214: feat(controller): disableDns + iptablesInit egress lockdown
**Author:** jezekra1 | **Branch:** feat/disable-dns-iptables-init → main | **Changes:** +615 −118 (19 files)

### Summary
Adds kernel-level egress lockdown for agent pods: a privileged init container (`egress-lockdown`) runs `iptables-nft` to pin the pod's OUTPUT chain to loopback + ESTABLISHED + paired gateway IP:port only, with IPv6 also locked down. Adds `disableDns` mode that drops cluster-DNS rules from the NetworkPolicy and substitutes a `hostAliases` entry so `HTTPS_PROXY` still resolves. Requires migrating existing headless gateway Services to ClusterIP-typed Services, handled by the new `ensureGatewayService` migration helper.

### Findings
- 🔴 **Critical:** `values.yaml` sets `controller.agentBase.iptablesInit.enabled: true` as the production default. The init container requires `NET_ADMIN` + `NET_RAW` and `runAsUser: 0`. On vanilla Kubernetes this will be blocked by default Pod Security Admission policies, and on OpenShift it requires the custom `anyuid-cap-net` SCC defined in this same PR. New deployments on standard clusters will fail to start agent pods until the SCC or PSA exception is configured. This should default to `false` with opt-in documentation. (`deploy/helm/platform/values.yaml`)
- 🟡 **Warning:** `waitForServiceDeleted` blocks the reconcile goroutine for up to 10 seconds polling every 100ms. Kubernetes reconcilers should be non-blocking and requeue rather than sleep in the hot path — a 10-second block here delays all other reconcile events for this controller. Return an error (causing a requeue) instead of a polling loop. (`packages/controller/pkg/reconciler/iptables_init.go`)
- 🟡 **Warning:** The headless→ClusterIP migration in `ensureGatewayService` deletes and recreates the gateway Service. During the brief delete→create window, existing agents pointing to the old headless DNS entry lose proxy connectivity. Operators upgrading existing deployments should be warned of this brief outage. (`packages/controller/pkg/reconciler/iptables_init.go`)
- 🟢 **Suggestion:** `buildIptablesInitContainer` calls `iptables-nft` directly (not the `iptables` wrapper) to work around `readOnlyRootFilesystem`. Well-documented in comments, but means nft-only; will fail on kernels with `iptables-legacy` only. Should be noted in the chart README.
- ✅ **Looks good:** IPv6 egress lockdown is correct — DROP all OUTPUT on v6 except loopback + ESTABLISHED, closing the IPv4-only bypass vector.
- ✅ **Looks good:** `AllowPrivilegeEscalation: false` on the init container is correct even with `RunAsUser: 0` — prevents setuid escalation while admitting explicitly requested capabilities.
- ✅ **Looks good:** `needsGatewayIP` fail-closed requeue prevents half-configured pods from deploying when ClusterIP hasn't been assigned yet.
- ✅ **Looks good:** `egress-lockdown` init container is correctly placed first — confirmed by `TestBuildAgentStatefulSet_IptablesInitRunsFirst` — so the allow-list is in place before any user init script runs.
- ✅ **Looks good:** Test coverage is thorough: disabled/nil/empty-image/no-IP → nil, capabilities, script shape, hostAliases placement, DNS-on vs DNS-off behavior.

### Verdict
COMMENT — Strong defense-in-depth security improvement with thorough tests, but `iptablesInit.enabled: true` as a production default is likely to break new deployments on standard clusters and should be `false` with explicit opt-in guidance.

---

pilartomas (approved):

PR #222, opened 2026-05-14
fix(ui): unify dialogs via composable shell and pin action buttons

### LINKED ISSUE #215: UI - Add Agent dialog: action buttons can scroll out of view
## Problem

The Add Agent form is long enough — image source, name, description, optional local-context import, connections picker, network-access preset — that on a normal laptop viewport the **Create Agent** button is pushed below the fold. The whole dialog scrolls as one block, so there's no visual cue an action exists below; users have to scroll inside the dialog to find the submit. Same for **Cancel** on the template-pick step.

The **Instance Settings** dialog has a smaller version of the same issue when the allowed-users list grows.

## Proposed solution

Pin the action buttons to the bottom edge so they stay visible while the form content scrolls. Other dialogs in the app already behave this way (configuring an existing agent, creating/editing a secret, connecting an app) — align these two with that pattern.

### LINKED ISSUE #220: UI - Unify dialogs via composable components
## Problem

Several dialogs across the app — Add Agent, Configure Agent, Instance Settings, Create/Edit Secret, Connect App — share the same visual shell (a header, a scrollable content region, and a pinned footer with action buttons), but each dialog hand-rolls that layout. Small inconsistencies have crept in as a result: some dialogs use slightly different padding on mobile than others, body sections use varying vertical spacing without a clear reason, and one dialog quietly carries a fix for a scroll-area edge case that the others don't.

None of these are visible bugs today, but the duplication has a recurring cost. The recent "Create Agent button scrolls off-screen" and "modal renders behind the bottom navigation" fixes both had to be applied in multiple places because the shell was copy-pasted, and each new dialog risks the same drift.

## Proposed solution

Extract the shared dialog shell — header, scrollable body, and pinned footer — into composable building blocks that every dialog uses. Each dialog keeps its own content and behavior; only the layout shell is unified. Cross-cutting changes (padding, dividers, scroll behavior) live in one place.

The user-visible effect is consistency: every dialog in the app looks and behaves the same way around its content, on both mobile and desktop. There is no functional change.

kapetr:
Closes #215, closes #220.

## Summary

- **Fix #215** — Pin action buttons in Add Agent (and originally Instance Settings, see below) by switching their layouts to the three-region pattern (header / scrollable body / sticky footer) that other dialogs already use. Buttons can no longer scroll off-screen.
- **Modal infrastructure** — Portal `<Modal>` into `document.body` so it escapes the app shell's `<main>` stacking context; this is what makes the modal actually render above the mobile bottom nav (the nav's `z-40` was beating the modal's `z-50` while the modal was nested inside `z-10` main). Mobile sizing: `max-h-[95dvh]` to use more of the viewport while keeping margin; horizontal overlay padding so every dialog has consistent breathing room regardless of `widthClass`.
- **Refactor #220** — Extract `DialogHeader` / `DialogBody` / `DialogFooter` subcomponents and migrate all seven dialog call sites (add-agent, configure-agent, instance-settings, create-secret, edit-secret, connect-app, add-mcp) to compose them. Cross-cutting fixes to padding, dividers, scroll behavior now live in one place. Spotted along the way: \`InstanceSettingsDialog\` had no callers — its feature moved to \`channels-panel\` inline — and is deleted.

## Test plan

- [ ] Desktop: open each of the seven dialogs (Add Agent, Configure Agent, Add Secret, Edit Secret, Connect App, Add MCP Server) — visual look matches main: same paddings, dividers, button placement.
- [ ] Desktop: long-content cases (Configure Agent with many connections; Edit Secret with many env mappings) — content scrolls, header and footer stay pinned.
- [ ] Mobile viewport (≤ 768px): every dialog has visible horizontal margin, max-height respects 95dvh, action buttons are pinned and clearly above the bottom nav.
- [ ] Mobile: open a dialog and confirm the bottom nav is fully covered (not visible through or in front of the modal) on a phone-sized viewport.
- [ ] Add Agent specifically: scroll through the configure step on a typical laptop viewport — \"Create Agent\" stays visible at the bottom. Same for \"Cancel\" on the pick step.
- [ ] Submit a form inside each dialog (Connect App OAuth start, Add Secret create, etc.) — submission still works; the \`<form className=\"contents\">\` pattern preserves form scope across the three regions.
- [ ] Reload after each migration commit if bisecting; each one is independently safe.


xjacka:
🛡️ **DAM** — Code Review @ \`5cf9244\`

## PR #222: fix(ui): unify dialogs via composable shell and pin action buttons
**Author:** kapetr | **Branch:** fix/215-pin-dialog-action-buttons → main | **Changes:** +158 −203 (8 files)

### Summary
Fixes the z-index stacking issue (modal rendered below `MobileNav z-40`) by switching `Modal` to use `createPortal(content, document.body)`, escaping the app shell's `z-10` stacking context. Extracts `DialogHeader`, `DialogBody`, `DialogFooter` composable layout primitives and migrates all five dialogs to use them, making cross-dialog layout fixes a single-file change going forward. Deletes `instance-settings-dialog.tsx`.

### Findings
- 🔴 **Critical:** `instance-settings-dialog.tsx` deleted with no visible replacement in this diff. This component exposed `InstanceSettingsDialog` managing per-instance allowed-user email access control. If it was ever mounted in the app, removing it is a silent regression that eliminates the access control UI. Verify this component was never reached (no import in active code) before merging. (`packages/ui/src/modules/instances/dialogs/instance-settings-dialog.tsx`)
- 🟢 **Suggestion:** In `add-mcp-form.tsx`, `DialogBody` wraps only the `<input>` element. Other dialogs wrap the full content area. The inconsistency is harmless but may confuse future contributors pattern-matching from the other examples. (`packages/ui/src/modules/connections/forms/add-mcp-form.tsx`)
- 🟢 **Suggestion:** `DialogBody` comment notes Tailwind utility overrides require `tailwind-merge` for reliable composition. If `tailwind-merge` is not in the project, document the limitation or use CSS variables for spacing instead.
- ✅ **Looks good:** `createPortal(content, document.body)` is the correct fix for the stacking-context issue; the inline comment explains the `z-10`/`z-40` conflict clearly.
- ✅ **Looks good:** `min-h-0` on `DialogBody` is the load-bearing flex-shrink fix; the comment explains why correctly.
- ✅ **Looks good:** `max-h-[95dvh] md:max-h-[85vh]` uses `dvh` for mobile viewport, falling back to `vh` on desktop.
- ✅ **Looks good:** Consistent `DialogHeader/DialogBody/DialogFooter` pattern applied uniformly to all 5 dialogs.

### Verdict
COMMENT — The portal fix is correct and the composable shell is a clean abstraction, but the deletion of `InstanceSettingsDialog` needs explicit confirmation it was unreachable before merging.

---

packages/ui/src/components/modal.tsx
```diff
@@ -9,19 +10,69 @@ interface ModalProps {
  * Centered overlay modal with brutal styling. Closes only via explicit
  * actions in the modal body — backdrop clicks and Escape are ignored so
  * users can't lose in-progress form state by accident.
+ *
+ * Renders into document.body via a portal so it escapes the app shell's
+ * `<main>` stacking context (z-10). Without the portal, MobileNav (z-40)
+ * would render above the modal because the modal's effective stacking
+ * happens at z-10 from the root's perspective.
```
PetrBulanek, line 18: The same stacking-context trap applies to `DialogOverlay` in `dialog-overlay.tsx` (`z-[100]`, not portaled). Today it dodges the bug only because `app.tsx:139` mounts it as a sibling to `<main>` — a future refactor moving it inside `<main>` would silently regress mobile. Worth either portaling it the same way here, or filing a follow-up so the next person reading the modal docs knows the global confirm dialog wasn't taken along.

packages/ui/src/modules/connections/forms/add-mcp-form.tsx
PetrBulanek, line 1: The four other dialogs now wrap their contents in `<form onSubmit={...} className="contents">` — this one and `connect-app-form` still hand-roll Enter via `onKeyDown` on each input and an `onClick` on the submit button. Since the PR is consolidating dialog conventions, wrapping these in a `<form>` here keeps the seven dialogs symmetric and lets the submit button revert to `type="submit"` (browser handles Enter, autofill, and password-manager UX consistently).

packages/ui/src/components/modal.tsx
```diff
@@ -9,19 +10,69 @@ interface ModalProps {
...
+ *
+ * Compose the inside with `DialogHeader`, `DialogBody`, and `DialogFooter`
+ * so layout (padding, dividers, scroll region) is consistent across every
+ * dialog and cross-cutting fixes happen in one place. Extras like a tab
+ * strip can sit between Header and Body as plain children.
  */
 export function Modal({
```
PetrBulanek: Since this PR is centralizing modal infrastructure, the four obvious a11y gaps are now cheap to close in one place: add `role="dialog"`, `aria-modal="true"`, and `aria-labelledby` (DialogHeader can expose an id), trap Tab inside the modal, and lock `body { overflow: hidden }` while the portal is mounted (background still scrolls behind the dialog today, very visible on mobile). All four are pre-existing, but the next time someone needs them they'll touch this same file — folding them in now avoids a second pass. Probably not needed to address in this PR, but might be worth a follow-up.

PetrBulanek (changes requested):

xjacka (approved):
🛡️ **DAM** — Code Review @ `91bceda`

## PR #222: fix(ui): unify dialogs via composable shell and pin action buttons
**Author:** kapetr | **Branch:** fix/215-pin-dialog-action-buttons → main | **Changes:** +375 −310 (9 files)

### Summary
Fixes the z-index stacking issue by portaling both `Modal` and `DialogOverlay` into `document.body`, escaping the app shell's `z-10` stacking context. Extracts `DialogHeader`, `DialogBody`, `DialogFooter` composable layout primitives from `modal.tsx`, migrates all seven dialog call sites, and adds full a11y: `role="dialog"`, `aria-modal="true"`, `aria-labelledby` via `ModalContext`, focus trap, and body-scroll lock. `InstanceSettingsDialog` is deleted — confirmed to have no callers (feature moved inline to `channels-panel`).

_(Suppressed 1 finding(s) per PR context: Critical deletion of `InstanceSettingsDialog` — PR author confirmed no callers, feature moved to `channels-panel` inline.)_

### Changes since last review
Previous HEAD: `5cf9244` (2026-05-15T06:57:51Z) — verdict COMMENT

- ✅ **Fixed:** Critical — `InstanceSettingsDialog` deleted with no replacement. PR author confirms in body: "had no callers — its feature moved to `channels-panel` inline".
- ✅ **Fixed:** `DialogOverlay` is now portaled into `document.body` with `useFocusTrap`, `useBodyScrollLock`, `role="alertdialog"`, `aria-modal`, `aria-labelledby`, and horizontal padding — addressing PetrBulanek's CHANGES_REQUESTED inline comment.
- ✅ **Fixed:** `add-mcp-form` and `connect-app-form` now use `<form onSubmit={...} className="contents">` pattern — addressing PetrBulanek's CHANGES_REQUESTED inline comment.
- ✅ **Fixed:** A11y gaps (`role="dialog"`, `aria-modal`, focus trap, scroll lock) added to `Modal` — addressing PetrBulanek's advisory comment.

### Findings
- 🟢 **Suggestion:** `useFocusTrap`'s `useEffect` dep array is `[containerRef]`. Since `RefObject` identity is stable across renders, this behaves as `[]`. The `react-hooks/exhaustive-deps` lint rule may flag this; changing to `[]` (or using `useLayoutEffect`) makes the intent explicit. (`packages/ui/src/components/modal.tsx`)
- ✅ **Looks good:** `createPortal(content, document.body)` on both `Modal` and `DialogOverlay` cleanly escapes the `z-10` stacking context; comments explain the rationale.
- ✅ **Looks good:** `useBodyScrollLock` is ref-counted (`bodyLockCount`) so nested overlays share one lock and restore overflow correctly on the last unmount.
- ✅ **Looks good:** `ModalContext` + `DialogHeader(id={labelId})` wires `aria-labelledby` without callers managing ids — clean design.
- ✅ **Looks good:** `useFocusTrap` restores focus to the previously active element on unmount — correct modal accessibility pattern.
- ✅ **Looks good:** Uniform `DialogHeader/DialogBody/DialogFooter` migration across all 7 dialogs; `min-h-0` on `DialogBody` is the load-bearing flex-shrink fix and is correctly documented.
- ✅ **Looks good:** `add-mcp-form` and `connect-app-form` use `type="submit"` with `disabled` guard — Enter-to-submit handled by native form submission, consistent with the rest.

### Verdict
APPROVE — the Critical finding is resolved (InstanceSettingsDialog had no callers), all PetrBulanek's CHANGES_REQUESTED concerns are addressed (DialogOverlay portaled, form pattern adopted, a11y added), and the implementation is solid.

---

PetrBulanek (approved):

PR #224, opened 2026-05-14
feat(cli): `dam agent create` — interactive end-to-end setup

### LINKED ISSUE #211: `dam agent create` — interactive end-to-end agent setup
## Problem

`dam instance create` (#188) lets a developer provision an agent from the terminal, but only from a template + name. Everything else the web "Add agent" dialog asks for — picking a model provider, connecting GitHub, granting access — still requires leaving the CLI for the web UI. That breaks the May 18 demo story, which is: open a terminal, run one command, walk through setup, and start coding.

## Proposed solution

`dam agent create` walks the developer through agent setup interactively:

1. Pick an agent template (e.g. Claude Code) from the templates the host offers.
2. Pick a model provider — either grant access to one that's already configured, or add a new one (Anthropic, IBM LiteLLM) by entering an API key.
3. Pick a GitHub connection — either grant an existing one, or add a new one by entering a personal access token.
4. The agent is created with provider access and GitHub wired up, ready for `dam shell` (#86) and `dam import` (#73).

Non-interactive use keeps working: passing the same inputs as flags should skip the prompts and behave like a scripted `instance create` today.

## Notes from the investigation

- **Naming**: lead with `dam agent` as the primary command and keep `dam instance` as an alias. Epic #71 and the web UI consistently use "agent"; the agent-vs-instance split is internal plumbing that users don't need to see. Aliasing avoids a pre-demo rename sweep.
- **GitHub connection**: ship the PAT-paste path first. The backend already supports it as a generic secret with header injection (`Authorization: Bearer {value}` on `api.github.com`, `GH_TOKEN` env mapping), `api.github.com` is in the `trusted` egress preset, and it keeps the whole flow in the terminal with no browser hop. OAuth (which is what the web UI's "Apps" tab uses) is a fine follow-up but not required for the demo.

## Out of scope

- Workspace import — `dam import` (#73).
- Attaching a terminal — `dam shell` (#86).
- Egress preset / network access configuration — epic area 10.
- GitHub OAuth flow from the CLI — follow-up after the demo.
- Schedules, skills, channels — separate epic areas.

## Dependencies

Part of #71. Builds on #188 (`dam instance create` + template discovery) and #75 (IBM LiteLLM provider). Pairs with #73 and #86 on the demo golden path.

PetrBulanek:
Closes #211.

## Summary

Adds `dam agent create`: a single interactive CLI command that walks a developer through the full new-agent flow — name, template, model provider, optional GitHub PAT — and ends with a running instance. Everything the web UI's "Add agent" dialog covers, now reachable without leaving the terminal.

`dam instance create` is unchanged and remains the scripted entry point.

## Test plan

- [ ] `dam agent create` on a fresh cluster (no providers, no PATs) walks through provider add-new and GitHub add-one paths and ends with a running agent.
- [ ] Re-running with existing provider / GitHub PAT shows them in the pickers and grants both halves of the GitHub pair.
- [ ] Replacing an existing provider key or GitHub PAT updates in place (no duplicates).
- [ ] Replacing an Anthropic API key with an OAuth token (or vice versa) correctly rotates env + injection header.
- [ ] Cancelling at any prompt before agent creation exits cleanly with no orphan resources.
- [ ] `dam agent create < /dev/null` errors out pointing at `dam instance create`.

### DIFF
--- a/packages/api-server-api/src/modules/secrets/router.ts
+++ b/packages/api-server-api/src/modules/secrets/router.ts
@@ -68,6 +68,16 @@ export const secretsRouter = t.router({
     )
     .mutation(({ ctx, input }) => ctx.secrets.createGithubPat(input)),
 
+  updateGithubPat: t.procedure
+    .input(
+      z.object({
+        apiSecretId: z.string().min(1),
+        gitSecretId: z.string().min(1),
+        token: z.string().min(1),
+      }),
+    )
+    .mutation(({ ctx, input }) => ctx.secrets.updateGithubPat(input)),
+
   update: t.procedure.input(updateSecretInputSchema).mutation(({ ctx, input }) => ctx.secrets.update(input)),
 
   delete: t.procedure
--- a/packages/api-server-api/src/modules/secrets/router.ts
+++ b/packages/api-server-api/src/modules/secrets/router.ts
@@ -59,6 +59,15 @@ export const secretsRouter = t.router({
     )
     .mutation(({ ctx, input }) => ctx.secrets.create(input)),
 
+  createGithubPat: t.procedure
+    .input(
+      z.object({
+        name: z.string().min(1).max(100),
+        token: z.string().min(1),
+      }),
+    )
+    .mutation(({ ctx, input }) => ctx.secrets.createGithubPat(input)),
+
   update: t.procedure.input(updateSecretInputSchema).mutation(({ ctx, input }) => ctx.secrets.update(input)),
 
   delete: t.procedure
--- a/packages/api-server/src/modules/secrets/services/secrets-service.ts
+++ b/packages/api-server/src/modules/secrets/services/secrets-service.ts
@@ -18,6 +18,7 @@ import {
   type AgentAccess,
 } from "api-server-api";
 import type {
+  AuthMode,
   K8sSecretsPort,
   K8sStoredSecret,
 } from "./../infrastructure/k8s-secrets-port.js";
@@ -56,6 +57,30 @@ function registryExtraInjections(
 function twinDisplayName(primaryName: string, injection: InjectionConfig): string {
   const tag = injection.queryParamName ? `?${injection.queryParamName}` : injection.headerName;
   return `${primaryName} (${tag})`;
+}
+
+/**
+ * Anthropic-only: if the new value's prefix indicates a different auth
+ * mode than the secret currently has, return the registry's env
+ * mappings for the new mode plus the new mode key. `update` uses this
+ * to re-bake injection + env when a CLI replace path swaps key formats.
+ * Returns `null` when no rotation is needed (non-Anthropic secret,
+ * same mode, or secret not found — the K8sPort will surface NOT_FOUND
+ * downstream).
+ */
+async function anthropicModeRotationFor(
+  k8sPort: K8sSecretsPort,
+  id: string,
+  newValue: string,
+): Promise<{ authMode: AuthMode; envMappings: EnvMapping[] } | null> {
+  const secrets = await k8sPort.listSecrets();
+  const existing = secrets.find((s) => s.id === id);
+  if (!existing || existing.type !== "anthropic") return null;
+  const newAuthMode: AuthMode = newValue.startsWith("sk-ant-oat") ? "oauth" : "api-key";
+  if (newAuthMode === existing.authMode) return null;
+  const mode = PROVIDERS.anthropic.modes.find((m) => m.key === newAuthMode);
+  if (!mode) return null;
+  return { authMode: newAuthMode, envMappings: [...mode.defaultEnvMappings] };
 }
 
 // `secrets-rev` hashes only what affects the agent pod's env. hostPattern,
@@ -304,13 +329,37 @@ export function createSecretsService(deps: {
     },
 
     async update({ id, ...patch }: UpdateSecretInput) {
+      // Anthropic value swaps re-discriminate the auth mode from the new
+      // value's prefix and rewrite envMappings + injectionConfig to match.
+      // Without this, replacing an API key (stored as
+      // `x-api-key: {value}`, env `ANTHROPIC_API_KEY`) with an OAuth
+      // token (needs `Authorization: Bearer {value}`, env
+      // `CLAUDE_CODE_OAUTH_TOKEN`) leaves the injection metadata stuck
+      // at the old mode — Anthropic receives `x-api-key: sk-ant-oat…`
+      // and rejects with "Invalid API key". The same prefix rule lives
+      // in `create` above; this just teaches `update` to redo it when
+      // the caller didn't supply envMappings explicitly (the web UI's
+      // Anthropic Edit form does, the CLI's replace path does not).
+      const rotation =
+        patch.value !== undefined && patch.envMappings === undefined
+          ? await anthropicModeRotationFor(deps.k8sPort, id, patch.value)
+          : null;
       const result = await deps.k8sPort.updateSecret(id, {
         ...(patch.name !== undefined ? { name: patch.name } : {}),
         ...(patch.value !== undefined ? { value: patch.value } : {}),
         ...(patch.hostPattern !== undefined ? { hostPattern: patch.hostPattern } : {}),
         ...(patch.pathPattern !== undefined ? { pathPattern: patch.pathPattern } : {}),
-        ...(patch.injectionConfig !== undefined ? { injectionConfig: patch.injectionConfig } : {}),
-        ...(patch.envMappings !== undefined ? { envMappings: patch.envMappings } : {}),
+        ...(patch.injectionConfig !== undefined
+          ? { injectionConfig: patch.injectionConfig }
+          : rotation
+            ? { injectionConfig: null }
+            : {}),
+        ...(patch.envMappings !== undefined
+          ? { envMappings: patch.envMappings }
+          : rotation
+            ? { envMappings: rotation.envMappings }
+            : {}),
+        ...(rotation ? { authMode: rotation.authMode } : {}),
       });
       // ADR-040 fanout. Host edits → re-sync egress_rules per granted
       // agent (hot, no roll). envMappings edits → bump secrets-rev so the
--- a/packages/api-server/src/modules/secrets/services/secrets-service.ts
+++ b/packages/api-server/src/modules/secrets/services/secrets-service.ts
@@ -10,6 +10,8 @@ import {
   type InjectionConfig,
   type SecretsService,
   type CreateSecretInput,
+  type UpdateGithubPatInput,
+  type UpdateGithubPatOutput,
   type UpdateSecretInput,
   type SecretType,
   type SecretView,
@@ -273,6 +275,32 @@ export function createSecretsService(deps: {
         apiSecretId: apiSecret.id,
         gitSecretId: gitSecret.id,
       };
+    },
+
+    async updateGithubPat(input: UpdateGithubPatInput): Promise<UpdateGithubPatOutput> {
+      // Re-wrap the github.com half server-side so callers send `{token}`
+      // only — same shape symmetry as createGithubPat.
+      //
+      // Value-only update: envMappings / hostPattern / injectionConfig
+      // stay the same, so neither the `secrets-rev` rolling-restart
+      // signal nor the connection-rules sync fires. The gateway pod's
+      // Envoy picks up the new value via SDS without a pod restart.
+      //
+      // Partial-failure note: if the github.com update throws, the
+      // api.github.com half already holds the new token while the
+      // github.com half still holds the prior wrapped value. The raw
+      // prior value isn't recoverable from the SDS file (the format
+      // template is baked into it), so we don't attempt to restore —
+      // surface the original error and let the caller retry. In
+      // practice this leaves `gh` working with the new token and
+      // `git clone` working with the old; a retry of `updateGithubPat`
+      // converges both halves.
+      const basicValue = Buffer.from(`x-access-token:${input.token}`).toString("base64");
+      const apiResult = await deps.k8sPort.updateSecret(input.apiSecretId, { value: input.token });
+      if (!apiResult) throw new TRPCError({ code: "NOT_FOUND", message: "api.github.com secret not found" });
+      const gitResult = await deps.k8sPort.updateSecret(input.gitSecretId, { value: basicValue });
+      if (!gitResult) throw new TRPCError({ code: "NOT_FOUND", message: "github.com secret not found" });
+      return { apiSecretId: input.apiSecretId, gitSecretId: input.gitSecretId };
     },
 
     async update({ id, ...patch }: UpdateSecretInput) {
--- a/packages/api-server/src/modules/secrets/services/secrets-service.ts
+++ b/packages/api-server/src/modules/secrets/services/secrets-service.ts
@@ -1,8 +1,11 @@
 import { createHash, randomUUID } from "node:crypto";
 import { TRPCError } from "@trpc/server";
 import {
+  DEFAULT_ENV_PLACEHOLDER,
   isProviderPresetType,
   PROVIDERS,
+  type CreateGithubPatInput,
+  type CreateGithubPatOutput,
   type EnvMapping,
   type InjectionConfig,
   type SecretsService,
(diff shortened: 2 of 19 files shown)


xjacka:
🛡️ **DAM** — Code Review @ \`abe59af\`

## PR #224: feat(cli): `dam agent create` — interactive end-to-end setup
**Author:** PetrBulanek | **Branch:** feat/dam-agent-create → main | **Changes:** +1421 −48 (16 files)

### Summary
Adds `dam agent create` — an interactive TTY-bound CLI command walking the user through name → template → model provider → optional GitHub PAT, then orchestrates `agents.create → instances.create → secrets.setAgentAccess` with rollback on failure and a 120s wait for `running` state. Server-side: `createGithubPat`/`updateGithubPat` tRPC mutations with server-side base64 wrapping, and Anthropic auth-mode auto-rotation on `secrets.update` when a value token prefix changes.

### Findings
- 🟡 **Warning:** `updateGithubPat` has a documented partial-failure window: if `updateSecret(gitSecretId)` throws after `updateSecret(apiSecretId)` succeeds, both PAT halves hold different tokens with no automatic recovery. The comment says "let the caller retry" but there is no retry in the CLI or error guidance to the user. A failed rotation leaves `gh` working with the new token while `git clone` uses the old. (`packages/api-server/src/modules/secrets/services/secrets-service.ts`)
- 🟡 **Warning:** `create.ts` passes raw GitHub PAT half IDs (`githubPat.apiSecretId`, `githubPat.gitSecretId`) directly to `setAgentAccess`, bypassing the twin-secret expansion mechanism (from PR #225). This is intentional for PATs today (they lack `primarySecretId`), but creates an inconsistency with Bob twin-secrets and will silently break if PATs are migrated to the twin-secret model. A comment noting this difference would prevent future confusion. (`packages/cli/src/modules/agent/commands/create.ts`)
- 🟢 **Suggestion:** `anthropicModeRotationFor` calls `k8sPort.listSecrets()` on every `update` that sets a value without explicit `envMappings`. This adds a full list call to the write path. Consider fetching just the updated secret via Get, or adding type/authMode to the update input. (`packages/api-server/src/modules/secrets/services/secrets-service.ts`)
- 🟢 **Suggestion:** All error paths in `create.ts` reuse `EXIT_INSTANCE_RUNTIME_FAILURE`. Distinct exit codes for compat failure, config failure, template unavailability, and mutation failure would let shell scripts distinguish root cause without stderr parsing.
- ✅ **Looks good:** Non-TTY guard at the top with a clear error message pointing to `dam instance create` for scripted callers.
- ✅ **Looks good:** Rollback tracking with `cleanup.newSecretIds` correctly excludes existing/replaced secrets — only newly-created state is torn down on failure.
- ✅ **Looks good:** SIGINT handler during the wait loop leaves the agent in place and exits non-zero — avoids destructive mid-wait cancel.
- ✅ **Looks good:** `@clack/prompts` pinned to exact version `1.3.0`.
- ✅ **Looks good:** Anthropic mode-rotation logic in `update` is well-tested with 4 cases (api-key→oauth, same-mode no-op, non-Anthropic no-op, caller-supplied envMappings wins).

### Verdict
COMMENT — Solid interactive flow with clean rollback design. The partial-failure window in `updateGithubPat` and the PAT/twin-secret grant inconsistency warrant a comment or explicit follow-up before merge.

---

xjacka:
🛡️ **DAM** — Code Review @ \`c85eb12\`

## PR #224: feat(cli): \`dam agent create\` — interactive end-to-end setup
**Author:** PetrBulanek | **Branch:** feat/dam-agent-create → main | **Changes:** +1455 −48 (19 files)

### Summary
Adds \`dam agent create\` — a TTY-bound interactive CLI command for end-to-end agent setup, mirroring the web UI's "Add agent" dialog: name → template → model provider → optional GitHub PAT → agent/instance creation → grant → wait. Server-side adds \`createGithubPat\`/\`updateGithubPat\` tRPC mutations, Anthropic auth-mode auto-rotation on \`secrets.update\` when a token prefix changes, and a race guard in \`setSecretGrants\`/\`setConnectionGrants\` that throws on empty ConfigMap lists so callers can retry.

### Changes since last review
Previous HEAD: \`abe59af\` — verdict COMMENT

- ✅ **Fixed:** Exit-code inconsistency across most error paths — constants now used consistently, except one \`process.exit(1)\` in the non-TTY guard.
- 🆕 **New:** Full CLI command (608 lines), \`group-github-pats.ts\` grouping utility, \`@clack/prompts@1.3.0\`, \`createGithubPat\`/\`updateGithubPat\` service methods with rollback and tests, Anthropic mode-rotation on \`update\` with 4 test cases, race guard in \`agent-grants-port.ts\`.
- 🔁 **Still present:** \`updateGithubPat\` partial-failure window — api half succeeds, git half fails, leaving diverged token values. Now extensively documented inline; CLI retry loop provides convergence.
- 🔁 **Still present:** \`setAgentAccess\` receives individual PAT half IDs directly, bypassing twin-secret expansion from PR #225.
- 🔁 **Still present:** \`anthropicModeRotationFor\` calls \`k8sPort.listSecrets()\` on every value-only \`update\` without explicit \`envMappings\`.

### Findings
- 🟡 **Warning:** \`withRetry\` retries all errors unconditionally — a NOT_FOUND or UNAUTHORIZED TRPCError retries 5× over ~10 seconds before the CLI surfaces the failure. No feedback is shown during the retries. An early-exit on non-retryable codes (NOT_FOUND, UNAUTHORIZED, BAD_REQUEST) would prevent the ~10s freeze and the misleading spinner. (\`packages/cli/src/modules/agent/commands/create.ts:1556-1571\`)
- 🟡 **Warning:** \`updateGithubPat\` partial-failure window still present from \`abe59af\`: if \`updateSecret(gitSecretId)\` throws after \`updateSecret(apiSecretId)\` succeeds, the CLI retry loop re-prompts for a token — sending a new value to both halves on retry, which converges them but silently discards the half-applied state without informing the user. The behavior is documented inline but no user-visible warning fires when the partial failure occurs. (\`packages/api-server/src/modules/secrets/services/secrets-service.ts:839-843\`)
- 🟡 **Warning:** \`setAgentAccess\` receives individual PAT half IDs directly (\`githubPat.apiSecretId\`, \`githubPat.gitSecretId\`) rather than through the twin-secret \`primarySecretId\` expansion path used by the Bob provider preset (PR #225). If PATs migrate to that model, this path silently breaks. A short inline comment noting the intentional divergence would prevent a future regression. (\`packages/cli/src/modules/agent/commands/create.ts:1141-1148\`)
- 🟢 **Suggestion:** Missing test: \`updateGithubPat\` when the git half is NOT_FOUND *after* the api half succeeds — the operationally significant partial-failure scenario is untested. The existing test only covers NOT_FOUND on the api half (before any update succeeds). (\`packages/api-server/src/__tests__/unit/secrets-service-update-github-pat.test.ts\`)
- 🟢 **Suggestion:** \`secrets.list.query()\` is called twice per \`dam agent create\` run — once inside \`pickProvider\` and once inside \`pickGithubPat\`. A single fetch passed to both pickers would halve server calls. (\`packages/cli/src/modules/agent/commands/create.ts\`)
- 🟢 **Suggestion:** Non-TTY guard uses raw \`process.exit(1)\` instead of \`EXIT_INSTANCE_RUNTIME_FAILURE\`. Minor inconsistency with all other exit paths in the file. (\`packages/cli/src/modules/agent/commands/create.ts:1022-1026\`)
- 🟢 **Suggestion:** \`anthropicModeRotationFor\` calls \`k8sPort.listSecrets()\` on every \`update\` that sets a value without explicit \`envMappings\` — full list call on the write path for all Anthropic secrets. Fetching only the updated secret or adding type/authMode to the update input would avoid this. (carried from \`abe59af\`, \`packages/api-server/src/modules/secrets/services/secrets-service.ts\`)
- ✅ **Looks good:** \`groupGithubPats\` correctly filters orphans (one host only) and sorts deterministically, with full test coverage including the mixed-types case.
- ✅ **Looks good:** \`createGithubPat\` rollback — deletes the \`api.github.com\` half if the \`github.com\` create fails, with secondary-error suppression. Both success and rollback paths are tested.
- ✅ **Looks good:** Race guard in \`setSecretGrants\`/\`setConnectionGrants\` throws on empty ConfigMap list, surfacing the K8s API visibility race as a retriable error instead of a silent no-op. Both methods tested.
- ✅ **Looks good:** SIGINT handler registered with \`once\` and removed on natural wait completion — clean lifecycle, no lingering handler.
- ✅ **Looks good:** Rollback ledger correctly excludes replace-existing secrets (value overwritten in place, old value unrecoverable) — only newly-created secrets are rolled back.
- ✅ **Looks good:** Anthropic mode-rotation in \`update\` is tested with 4 cases: api-key→oauth, same-mode no-op, non-Anthropic no-op, caller-supplied \`envMappings\` wins.

### Verdict
COMMENT — Solid interactive flow with clean rollback design and good test coverage. The \`withRetry\`-without-fast-fail issue is new and worth addressing before merge; the PAT half-ID bypass inconsistency from the prior review remains open.

---

xjacka:
🛡️ **DAM** — Code Review @ `52498dc`

## PR #224: feat(cli): `dam agent create` — interactive end-to-end setup
**Author:** PetrBulanek | **Branch:** feat/dam-agent-create → main | **Changes:** +1614 −80 (19 files)

### Summary
Adds `dam agent create` — a TTY-bound interactive CLI command for end-to-end agent setup, mirroring the web UI's "Add agent" dialog: name → template → model provider → optional GitHub PAT → agent/instance creation → grant → wait. Server-side adds `createGithubPat`/`updateGithubPat` tRPC mutations, Anthropic auth-mode auto-rotation on `secrets.update` when a token prefix changes, and a race guard in `setSecretGrants`/`setConnectionGrants` that throws on empty ConfigMap lists so callers can retry. The latest commits add documentation for the full flow in `cli.md` and document the GitHub PAT twin-secret credential model in `security-and-credentials.md`.

### Changes since last review
Previous HEAD: `c85eb12` (2026-05-15T07:16:02Z) — verdict COMMENT

- ✅ **Fixed:** Missing CLI architecture documentation — `docs/architecture/cli.md` now includes the "Interactive agent setup" section and corrects the "dam shell" → "dam chat" future-verb reference
- ✅ **Fixed:** Missing documentation of the GitHub PAT two-secret credential model — `docs/architecture/security-and-credentials.md` now covers the twin-secret storage pattern with base64-wrapping detail and atomic create/rollback contract
- 🔁 **Still present:** `withRetry` retries all errors unconditionally, freezing the CLI for ~10s on non-retryable codes without user feedback (`packages/cli/src/modules/agent/commands/create.ts`)
- 🔁 **Still present:** `updateGithubPat` partial-failure window — documented inline but no user-visible warning fires on partial failure (`packages/api-server/src/modules/secrets/services/secrets-service.ts`)
- 🔁 **Still present:** `setAgentAccess` receives PAT half IDs directly, bypassing twin-secret `primarySecretId` expansion (`packages/cli/src/modules/agent/commands/create.ts`)
- 🔁 **Still present:** Missing test: `updateGithubPat` when git half is NOT_FOUND after api half succeeds
- 🔁 **Still present:** `secrets.list.query()` called twice per run
- 🔁 **Still present:** Non-TTY guard uses raw `process.exit(1)` instead of `EXIT_INSTANCE_RUNTIME_FAILURE`
- 🔁 **Still present:** `anthropicModeRotationFor` calls `k8sPort.listSecrets()` on every value-only `update`

### Findings
- 🟡 **Warning:** `withRetry` retries all errors unconditionally — a `NOT_FOUND` or `UNAUTHORIZED` TRPCError retries 5× over ~10s before the CLI surfaces the failure. No feedback is shown during retries. An early-exit on non-retryable codes (`NOT_FOUND`, `UNAUTHORIZED`, `BAD_REQUEST`) would prevent the ~10s freeze and misleading spinner. (`packages/cli/src/modules/agent/commands/create.ts:1752-1767`)
- 🟡 **Warning:** `updateGithubPat` partial-failure window: if `updateSecret(gitSecretId)` throws after `updateSecret(apiSecretId)` succeeds, the CLI retry loop re-prompts for a token — converging both halves on retry but discarding the half-applied state without a user-visible warning. Documented inline but no error signal fires when the partial failure occurs. (`packages/api-server/src/modules/secrets/services/secrets-service.ts:907-912`)
- 🟡 **Warning:** `setAgentAccess` receives individual PAT half IDs directly (`githubPat.apiSecretId`, `githubPat.gitSecretId`) rather than through the twin-secret `primarySecretId` expansion path used by the Bob provider preset. If PATs migrate to that model, this path silently breaks without a compiler error. A short inline comment noting the intentional divergence would prevent a future regression. (`packages/cli/src/modules/agent/commands/create.ts:1299`)
- 🟢 **Suggestion:** Missing test: `updateGithubPat` when the git half returns `null` (NOT_FOUND) after the api half succeeds — the operationally significant partial-failure scenario is untested. The existing test only covers NOT_FOUND on the api half (before any update succeeds). (`packages/api-server/src/__tests__/unit/secrets-service-update-github-pat.test.ts`)
- 🟢 **Suggestion:** `secrets.list.query()` is called twice per `dam agent create` run — once inside `pickProvider` and once inside `pickGithubPat`. A single fetch passed to both pickers would halve server calls. (`packages/cli/src/modules/agent/commands/create.ts`)
- 🟢 **Suggestion:** Non-TTY guard uses raw `process.exit(1)` instead of `EXIT_INSTANCE_RUNTIME_FAILURE`. Minor inconsistency with all other exit paths in the file. (`packages/cli/src/modules/agent/commands/create.ts:1177`)
- 🟢 **Suggestion:** `anthropicModeRotationFor` calls `k8sPort.listSecrets()` on every `update` that sets a value without explicit `envMappings` — a full list call on the write path for all Anthropic secrets. Fetching only the updated secret or adding type/authMode to the update input would avoid this. (`packages/api-server/src/modules/secrets/services/secrets-service.ts:663-676`)
- 🟢 **Suggestion:** The success outro includes `→ Next: dam chat ${name}`, but `dam chat` is a future module not yet implemented. Users who try this hint will get a command-not-found error. Consider omitting the hint or marking it explicitly as "coming soon." (`packages/cli/src/modules/agent/commands/create.ts:1354`)
- ✅ **Looks good:** `docs/architecture/cli.md` now has a complete "Interactive agent setup" section accurately describing the `dam agent create` flow, rollback semantics, and SIGINT behavior.
- ✅ **Looks good:** `docs/architecture/security-and-credentials.md` now documents the GitHub PAT twin-secret model with base64-wrapping detail and atomic create/rollback contract.
- ✅ **Looks good:** `groupGithubPats` correctly filters orphans and sorts deterministically, with full test coverage including the mixed-types case.
- ✅ **Looks good:** `createGithubPat` rollback — deletes the `api.github.com` half if the `github.com` create fails, with secondary-error suppression. Both success and rollback paths are tested.
- ✅ **Looks good:** Race guard in `setSecretGrants`/`setConnectionGrants` throws on empty ConfigMap list, surfacing the K8s-API visibility race as retriable instead of a silent no-op. Both methods tested.
- ✅ **Looks good:** SIGINT handler registered with `once` and removed on natural wait completion — clean lifecycle, no lingering handler.
- ✅ **Looks good:** Rollback ledger correctly excludes replace-existing secrets — only newly-created secrets are rolled back.
- ✅ **Looks good:** Anthropic mode-rotation in `update` is tested with 4 cases: api-key→oauth, same-mode no-op, non-Anthropic no-op, caller-supplied `envMappings` wins.

### Verdict
COMMENT — The new commits address the documentation gaps from the prior review. The `withRetry`-without-fast-fail issue and the PAT half-ID bypass inconsistency remain open; the withRetry issue causes a ~10s user-visible freeze on non-retryable errors and is the most actionable item before merge.

---

xjacka:
🛡️ **DAM** — Code Review @ \`2aecc14\`

## PR #224: feat(cli): \`dam agent create\` — interactive end-to-end setup
**Author:** PetrBulanek | **Branch:** feat/dam-agent-create → main | **Changes:** +1624 −80 (19 files)

### Summary
Adds \`dam agent create\` — a TTY-bound interactive CLI command for end-to-end agent setup, mirroring the web UI's "Add agent" dialog: name → template → model provider → optional GitHub PAT → agent/instance creation → grant → wait. Server-side adds \`createGithubPat\`/\`updateGithubPat\` tRPC mutations, Anthropic auth-mode auto-rotation on \`secrets.update\`, and a race guard in \`setSecretGrants\`/\`setConnectionGrants\`. The current commit refines the CLI architecture docs to list \`dam agent\` in the current CLI surface and fixes the remaining \`dam shell\` → \`dam chat\` reference.

### Changes since last review
Previous HEAD: \`52498dc\` (2026-05-15T08:15:57Z) — verdict COMMENT

- ✅ **Fixed:** \`docs/architecture/cli.md\` overview updated to list \`dam agent\` group in the current CLI surface; remaining \`dam shell\` reference corrected to \`dam chat\` in the "Motivated by" bullet.
- 🔁 **Still present:** \`withRetry\` retries all errors unconditionally — non-retryable codes (\`NOT_FOUND\`, \`UNAUTHORIZED\`, \`BAD_REQUEST\`) freeze the CLI for ~10s without feedback.
- 🔁 **Still present:** \`updateGithubPat\` partial-failure window — documented inline but no user-visible warning when api half updates and git half fails.
- 🔁 **Still present:** \`setAgentAccess\` receives PAT half IDs directly, bypassing twin-secret \`primarySecretId\` expansion.
- 🔁 **Still present:** Missing test for \`updateGithubPat\` when git half is NOT_FOUND *after* api half succeeds.
- 🔁 **Still present:** \`secrets.list.query()\` called twice per run.
- 🔁 **Still present:** Non-TTY guard uses raw \`process.exit(1)\` instead of \`EXIT_INSTANCE_RUNTIME_FAILURE\`.
- 🔁 **Still present:** \`anthropicModeRotationFor\` calls \`k8sPort.listSecrets()\` on the write path.
- 🔁 **Still present:** Success outro hints \`dam chat \${name}\` — future verb not yet implemented.

### Findings
- 🟡 **Warning:** \`withRetry\` retries all errors unconditionally — a \`NOT_FOUND\` or \`UNAUTHORIZED\` TRPCError retries 5× over ~10s before the CLI surfaces the failure. No feedback is shown during retries. An early-exit on non-retryable codes (\`NOT_FOUND\`, \`UNAUTHORIZED\`, \`BAD_REQUEST\`) would prevent the ~10s freeze and misleading spinner. (\`packages/cli/src/modules/agent/commands/create.ts\`)
- 🟡 **Warning:** \`updateGithubPat\` partial-failure window: if \`updateSecret(gitSecretId)\` throws after \`updateSecret(apiSecretId)\` succeeds, the CLI retry loop re-prompts for a token — converging both halves on retry but discarding the half-applied state without a user-visible warning. Documented inline but no error signal fires when the partial failure occurs. (\`packages/api-server/src/modules/secrets/services/secrets-service.ts\`)
- 🟡 **Warning:** \`setAgentAccess\` receives individual PAT half IDs directly (\`githubPat.apiSecretId\`, \`githubPat.gitSecretId\`) rather than through the twin-secret \`primarySecretId\` expansion path used by the Bob provider preset. If PATs migrate to that model, this path silently breaks without a compiler error. A short inline comment noting the intentional divergence would prevent a future regression. (\`packages/cli/src/modules/agent/commands/create.ts\`)
- 🟡 **Warning:** Both PR #224 and PR #230 modify the same overview paragraph in \`docs/architecture/cli.md\`. PR #224 adds \`dam agent\` to the current surface; PR #230 adds \`dam import\` and removes it from future verbs. Whichever merges second will need a manual conflict resolution — worth coordinating or rebasing one off the other before landing.
- 🟢 **Suggestion:** Missing test: \`updateGithubPat\` when the git half returns \`null\` (NOT_FOUND) after the api half succeeds — the operationally significant partial-failure scenario is untested. The existing test only covers NOT_FOUND on the api half (before any update succeeds). (\`packages/api-server/src/__tests__/unit/secrets-service-update-github-pat.test.ts\`)
- 🟢 **Suggestion:** \`secrets.list.query()\` is called twice per \`dam agent create\` run — once inside \`pickProvider\` and once inside \`pickGithubPat\`. A single fetch passed to both pickers would halve server calls. (\`packages/cli/src/modules/agent/commands/create.ts\`)
- 🟢 **Suggestion:** Non-TTY guard uses raw \`process.exit(1)\` instead of \`EXIT_INSTANCE_RUNTIME_FAILURE\`. Minor inconsistency with all other exit paths in the file. (\`packages/cli/src/modules/agent/commands/create.ts\`)
- 🟢 **Suggestion:** \`anthropicModeRotationFor\` calls \`k8sPort.listSecrets()\` on every \`update\` that sets a value without explicit \`envMappings\` — full list call on the write path for all Anthropic secrets. Fetching only the updated secret or adding type/authMode to the update input would avoid this. (\`packages/api-server/src/modules/secrets/services/secrets-service.ts\`)
- 🟢 **Suggestion:** The success outro includes \`→ Next: dam chat \${name}\`, but \`dam chat\` is a future module not yet implemented. Users who try this hint will get a command-not-found error. Consider omitting the hint or marking it explicitly as "coming soon." (\`packages/cli/src/modules/agent/commands/create.ts\`)
- ✅ **Looks good:** \`docs/architecture/cli.md\` now accurately lists \`dam agent\` in the current CLI surface.
- ✅ **Looks good:** \`groupGithubPats\` correctly filters orphans and sorts deterministically, with full test coverage including the mixed-types case.
- ✅ **Looks good:** \`createGithubPat\` rollback — deletes the \`api.github.com\` half if the \`github.com\` create fails, with secondary-error suppression. Both success and rollback paths are tested.
- ✅ **Looks good:** Race guard in \`setSecretGrants\`/\`setConnectionGrants\` throws on empty ConfigMap list, surfacing K8s-API visibility race as retriable instead of a silent no-op. Both methods tested.
- ✅ **Looks good:** SIGINT handler registered with \`once\` and removed on natural wait completion — clean lifecycle, no lingering handler.
- ✅ **Looks good:** Rollback ledger correctly excludes replace-existing secrets — only newly-created secrets are rolled back.
- ✅ **Looks good:** Anthropic mode-rotation in \`update\` is tested with 4 cases: api-key→oauth, same-mode no-op, non-Anthropic no-op, caller-supplied \`envMappings\` wins.

### Verdict
COMMENT — The new commit brings minor doc refinements; the \`withRetry\`-without-fast-fail issue and the PAT half-ID bypass inconsistency remain open from prior reviews. The potential merge conflict with PR #230 on \`docs/architecture/cli.md\` should be coordinated before landing.

---

packages/api-server/src/modules/secrets/services/secrets-service.ts
```diff
@@ -53,6 +59,30 @@ function twinDisplayName(primaryName: string, injection: InjectionConfig): strin
...
+/**
+ * Anthropic-only: if the new value's prefix indicates a different auth
+ * mode than the secret currently has, return the registry's env
+ * mappings for the new mode plus the new mode key. `update` uses this
+ * to re-bake injection + env when a CLI replace path swaps key formats.
+ * Returns `null` when no rotation is needed (non-Anthropic secret,
+ * same mode, or secret not found — the K8sPort will surface NOT_FOUND
+ * downstream).
+ */
+async function anthropicModeRotationFor(
```
tomkis, line 71: Don't love this living server-side. The prefix rule (`sk-ant-oat…` → mode)
is value-introspection — server reaching into the credential to figure
out what kind it is.

Server still needs `authMode` for SDS header + env mapping, but it doesn't
need to derive it. CLI can detect mode from the new value, look up the
existing secret's mode from `envMappings[0].envName` (UI already does
this in `detectMode`), prompt the user if it's a switch, then pass
`authMode` explicitly.

This is essentialy UX change.

packages/api-server-api/src/modules/secrets/router.ts
```diff
@@ -59,6 +59,25 @@ export const secretsRouter = t.router({
     )
     .mutation(({ ctx, input }) => ctx.secrets.create(input)),
 
+  createGithubPat: t.procedure
```
tomkis, line 62: Not a fan of these new endpoints. Feels like we're baking GitHub-specific
knowledge into the server as an escape hatch to get the CLI working.
Long-term we'll want first-class GH support with a proper data model, and
these endpoints will get ripped out.

Can we just drive this from the CLI? Compute the `x-access-token` Basic
wrap client-side, call `createSecret` / `updateSecret` twice, best-effort
delete on partial failure. Server stays generic.

Atomicity isn't really an argument either — current solution is not atomic anyway (no psql transaction that would guarantee this)

tomkis (changes requested):

tomkis (approved):
We'll merge now and address the review separately.

PR #232, opened 2026-05-15
fix(cli): normalize trailing slash in chat websocket URL

### LINKED ISSUE #218: `dam chat` hangs silently when server URL has a trailing slash
## Problem

If a user's configured server URL ends with `/` (e.g. `http://localhost:5173/`), `dam chat <instance>` hangs forever — no output, no error, no timeout. Other CLI commands against the same configured server (e.g. `ping`, `instance list`) work fine, so there's no indication anything is wrong with the configuration. The user is left staring at a frozen terminal with no way to know that a single trailing character broke the command. Ctrl-C is the only escape.

This is easy to hit in practice: people copy server URLs from a browser address bar, or from docs that include the slash, and the CLI accepts the value without complaint.

## Proposed solution

`dam chat` should connect normally regardless of whether the configured server URL ends with `/`. The CLI should treat `http://host:port` and `http://host:port/` as equivalent — consistent with how every other command already behaves.

### Scope

- Normalize trailing slashes for the chat command's terminal connection path, matching the behavior of the other CLI commands.

### Out of scope

- Adding validation/normalization at the point where the server URL is *stored* (e.g. `dam config set server …`). Worth doing eventually, but a separate concern — the chat command should be robust either way.

PetrBulanek:
## Summary

- The WebSocket URL builder for `dam chat` (now in `terminal-bridge.ts` after refactor 9185a7b5 moved it out of `chat-service.ts`) strips the `http(s)://` scheme but does **not** strip a trailing slash from the configured server host. With `server = "http://host/"`, this produces `ws://host//api/...`. Through a dev-server proxy (e.g. Vite) the upgrade request never matches the proxy rule, no HTTP response comes back, and the `ws` client stays in CONNECTING forever — leaving `dam chat` hanging silently with no output and no error.
- Normalize the host the same way every other URL builder in the CLI already does (`.replace(/\/+$/, "")` — see `trpc-client.ts`, `sessions-client.ts`, `version-probe.ts`, `auth-config-probe.ts`, `oidc-discovery.ts`).

> Note: an earlier version of this PR patched `wsUrl` in `chat-service.ts`. That symbol no longer exists — commit 9185a7b5 moved WS URL construction into `terminal-bridge.ts` without porting the normalization, reintroducing #218. This PR re-applies the fix at the new location.

Fixes #218

## Test plan

- [ ] Set `server = "http://api-server.localtest.me:4444/"` (trailing slash) in CLI config; `dam chat <instance>` connects normally instead of hanging.
- [ ] Same config without trailing slash continues to work.
- [ ] `mise run cli:check` passes.

### DIFF
--- a/packages/cli/src/modules/chat/infrastructure/terminal-bridge.ts
+++ b/packages/cli/src/modules/chat/infrastructure/terminal-bridge.ts
@@ -28,7 +28,7 @@ export function connectTerminalBridge({
   return new Promise<BridgeResult>((resolve) => {
     let settled = false;
     const proto = host.startsWith("https://") ? "wss:" : "ws:";
-    const base = host.replace(/^https?:\/\//, "");
+    const base = host.replace(/^https?:\/\//, "").replace(/\/+$/, "");
     const sep = terminalPath.includes("?") ? "&" : "?";
     const ws = new WebSocket(
       `${proto}//${base}${terminalPath}${sep}token=${encodeURIComponent(token)}`,


xjacka:
🛡️ **DAM** — Code Review @ \`713fcdd\`

## PR #232: fix(cli): normalize trailing slash in chat websocket URL
**Author:** PetrBulanek | **Branch:** fix/chat-trailing-slash-hang → main | **Changes:** +1 −1 (1 file)

### Summary
Adds `.replace(/\/+$/, "")` when constructing the WebSocket base URL so that a host passed with a trailing slash (e.g. `https://example.com/`) doesn't produce a double-slash URL like `wss://example.com//api/...` that hangs the connection.

### Findings
- ✅ **Looks good:** The `/\/+$/` regex handles multiple consecutive trailing slashes, not just one — handles the pathological `https://host///` case too.
- ✅ **Looks good:** Existing `encodeURIComponent` guards on `instanceId`, `token`, and `sessionId` are unchanged.
- 🟢 **Suggestion:** No test for `wsUrl` visible in the diff — a unit test covering the trailing-slash edge case would prevent regression. (`packages/cli/src/modules/chat/services/chat-service.ts`)

### Verdict
APPROVE — minimal, correct one-line fix with appropriate regex breadth; test coverage is the only gap.

---

JanPokorny (approved):

PR #234, opened 2026-05-15
fix(controller): probe iptables backend, fall back to legacy for Kata

jezekra1:
## Summary

- The `egress-lockdown` init container hard-pinned `iptables-nft`, which fails on Kata Containers guest kernels (OpenShift Sandboxed Containers and similar) with `iptables: Failed to initialize nft: Protocol not supported`. Kata's stock netfilter fragment enables `CONFIG_IP_NF_IPTABLES_LEGACY=y` but omits `CONFIG_NF_TABLES`.
- Script now probes `iptables-nft -nL OUTPUT`, falls back to `iptables-legacy`, applies the same allow-list via `\$IPT` / `\$IP6T`.
- Fail-closes (exit 1) with a loud error if neither backend works — defense-in-depth is the whole point of this init container, so we don't silently degrade to NetworkPolicy-only.

Originally introduced in #214; this fixes the OpenShift/Kata blocker reported with the log line `egress-lockdown: gateway=172.30.47.64:10000` followed by `iptables: Failed to initialize nft: Protocol not supported`.

## Test plan

- [x] `mise run controller:test` — reconciler tests green
- [x] `mise run controller:check` — `go vet` + build clean
- [x] `sh -n` on the embedded script — syntax OK
- [ ] Validate on the OpenShift cluster that originally hit the error: egress-lockdown init container reaches `egress-lockdown: backend=iptables-legacy` and `egress-lockdown: gateway-only IPv4 + IPv6 drop applied`
- [ ] Validate on regular (non-Kata) clusters that `iptables-nft` is still the chosen backend

### DIFF
--- a/packages/controller/pkg/reconciler/iptables_init.go
+++ b/packages/controller/pkg/reconciler/iptables_init.go
@@ -21,11 +21,16 @@ const iptablesInitContainerName = "egress-lockdown"
 // feature is off, the image is unset, or the gateway IP isn't known yet.
 //
 // Targets the SIG Release `registry.k8s.io/build-image/distroless-iptables`
-// image. We invoke `iptables-nft` directly rather than the wrapped
-// `iptables` entrypoint — the wrapper auto-detects nft vs legacy by
-// writing symlinks at runtime, which the container's
-// `readOnlyRootFilesystem: true` blocks. Every kernel K8s supports
-// (≥4.18) ships nftables, so the direct call is portable.
+// image, which ships both `iptables-nft` and `iptables-legacy`. We
+// invoke the backend binary directly rather than the wrapped `iptables`
+// entrypoint — the wrapper auto-detects nft vs legacy by writing
+// symlinks at runtime, which the container's `readOnlyRootFilesystem:
+// true` blocks. The script probes nft first and falls back to legacy:
+// Kata Containers guest kernels (used by OpenShift Sandboxed Containers
+// and others) commonly ship `CONFIG_IP_NF_IPTABLES_LEGACY=y` but not
+// `CONFIG_NF_TABLES`, so `iptables-nft` fails with
+// "Failed to initialize nft: Protocol not supported" while
+// `iptables-legacy` works against the same kernel.
 func buildIptablesInitContainer(cfg *config.Config, gatewayClusterIP string) *corev1.Container {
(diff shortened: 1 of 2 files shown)


xjacka:
🛡️ **DAM** — Code Review @ `7084028`

## PR #234: fix(controller): probe iptables backend, fall back to legacy for Kata
**Author:** jezekra1 | **Branch:** fix/egress-lockdown-iptables-fallback → main | **Changes:** +53 −19 (2 files)

### Summary
Replaces the hard-coded `iptables-nft` invocations in the egress-lockdown init container with a probe-first script that detects whether the nft or legacy iptables backend is functional on the running kernel, failing closed (exit 1) if neither works. This unblocks Kata Containers environments whose guest kernels ship `iptables-legacy` but not `nf_tables`.

### Findings
- ✅ **Looks good:** `set -eu` is intact; `if/elif` conditions don't trigger `set -e` early-exit, so the probe works correctly.
- ✅ **Looks good:** Probe command (`iptables-nft -nL OUTPUT >/dev/null 2>&1`) is non-destructive — no rules are modified during backend detection.
- ✅ **Looks good:** Fail-closed (`exit 1`) when neither backend responds is the correct security posture for an egress-lockdown init container — NetworkPolicy alone is not sufficient.
- ✅ **Looks good:** Tests comprehensively cover the probe logic, backend selection, variable-based rule application (`"$IPT"`, `"$IP6T"`), and the fail-closed exit path.
- 🟢 **Suggestion:** The probe tests only IPv4 backends and infers IPv6 availability from that. On all known kernels this is valid (nf_tables and ip6tables share the same subsystem), but a single inline comment in the script would make this assumption explicit — e.g., `# ip6tables shares the same kernel subsystem; no separate probe needed`. (`packages/controller/pkg/reconciler/iptables_init.go`)
- 🟢 **Suggestion:** The FATAL error message contains a UTF-8 em dash (`—`). Fine on all modern container runtimes, but using `--` is a safe defensive alternative if ever run in a constrained locale. (`packages/controller/pkg/reconciler/iptables_init.go:58`)

### Verdict
APPROVE — Clean, well-motivated fix; the fail-closed probe design is the right security-first approach and the test coverage is thorough.

---

pilartomas (approved):

PR #245, opened 2026-05-18
fix(controller): cascade-delete envoy leaf TLS Secret with instance

### LINKED ISSUE #244: Envoy TLS secrets leak when agents are deleted
## Problem

When an agent is deleted, the per-agent TLS certificate secret used by its Envoy proxy is left behind in the cluster. Over time, repeatedly creating and deleting agents produces an accumulation of orphaned TLS secrets that no longer correspond to any live agent.

This matters because:

- Operators see noise in the namespace (`<name>-envoy-tls` secrets that don't belong to anything).
- Cluster state drifts further from the agent inventory the longer the platform runs, which makes auditing and reasoning about "what's in the cluster" harder.
- Long-lived clusters slowly bloat with dead key material that nothing rotates or revokes.

The gap is that cert-manager, by default, does not tie the Secret it produces to the Certificate's lifecycle — so when the per-agent Certificate is cascade-deleted alongside the agent, the Secret remains.

## Proposed solution

When an agent is deleted, its Envoy TLS secret should be removed alongside the agent's other per-agent resources. From an operator's perspective, deleting an agent should leave no per-agent artifacts behind.

The intended direction is to have the controller set an owner reference on the produced Secret pointing at the same per-agent owner the Certificate uses, so Kubernetes garbage collection handles cleanup. This stays consistent with how other per-agent resources are already cleaned up.

### Out of scope

- Cleanup of other orphaned per-agent resources (credential secrets, etc.) — those are tracked separately if needed.
- Rotation or revocation policy for active certificates.

pilartomas:
## Summary

- cert-manager doesn't set an OwnerReference on Secrets it produces from `Certificate` resources by default (the cluster-wide `--enable-certificate-owner-ref` flag controls that, and we don't require it). So deleting an instance was leaving its `<instance>-envoy-tls` Secret behind, accumulating dead key material in the namespace over time.
- The controller now patches the produced Secret with a non-controller OwnerReference back to the instance ConfigMap on each reconcile. Kubernetes garbage collection then cascade-deletes the Secret alongside the rest of the per-instance state. The patch is idempotent and is a no-op when the Secret hasn't been issued yet — the next reconcile retries.

Closes #244

## Test plan

- [x] `mise run controller:test` — controller suite green, including three new tests covering: owner ref is added, idempotent across reconciles, no-op when the Secret isn't present yet
- [x] `mise run controller:check` — vet + build clean
- [ ] Manual: create an instance, wait for cert-manager to issue the leaf, confirm the Secret carries an `ownerReferences` entry pointing at the instance ConfigMap; delete the instance, confirm the Secret is reaped

### DIFF
--- a/packages/controller/pkg/reconciler/instance.go
+++ b/packages/controller/pkg/reconciler/instance.go
@@ -354,6 +354,67 @@ func (r *InstanceReconciler) ReconcileOrphanPVCs(ctx context.Context) {
 	}
 }
 
+// ReconcileOrphanLeafSecrets deletes any per-instance envoy leaf TLS
+// Secret whose instance ConfigMap no longer exists. Covers historical
+// leaks from before owner-references were added to these Secrets, plus
+// any future Secret that slips through (e.g. cert-manager produced the
+// Secret between the controller patching the ownerRef and the instance
+// being deleted, or the controller crashed in that window).
+//
+// Match is intentionally tight: Secret name ends in `-envoy-tls`, type
+// is `kubernetes.io/tls`, and the prefix names a non-existent
+// ConfigMap. Anything else is left alone.
+func (r *InstanceReconciler) ReconcileOrphanLeafSecrets(ctx context.Context) {
+	secrets, err := r.client.CoreV1().Secrets(r.config.Namespace).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		slog.Warn("orphan leaf Secret GC: listing Secrets failed", "error", err)
+		return
+	}
+	deleted := 0
+	scanned := 0
+	for _, sec := range secrets.Items {
+		instanceName, ok := instanceNameFromLeafSecret(sec)
+		if !ok {
+			continue
+		}
+		scanned++
+		_, err := r.client.CoreV1().ConfigMaps(r.config.Namespace).Get(ctx, instanceName, metav1.GetOptions{})
+		if err == nil {
+			continue
+		}
+		if !errors.IsNotFound(err) {
+			slog.Warn("orphan leaf Secret GC: API lookup failed", "instance", instanceName, "error", err)
+			continue
+		}
+		if err := r.client.CoreV1().Secrets(r.config.Namespace).Delete(ctx, sec.Name, metav1.DeleteOptions{}); err != nil {
+			slog.Warn("orphan leaf Secret GC: delete failed", "secret", sec.Name, "instance", instanceName, "error", err)
+			continue
+		}
+		slog.Info("orphan leaf Secret GC: deleted Secret for missing instance", "secret", sec.Name, "instance", instanceName)
+		deleted++
+	}
+	if deleted > 0 {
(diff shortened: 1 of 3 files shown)


packages/controller/pkg/reconciler/instance.go
```diff
@@ -219,6 +229,37 @@ func (r *InstanceReconciler) ensureAgentOwnerReference(ctx context.Context, inst
...
+			return err
+		}
+		for _, ref := range sec.OwnerReferences {
+			if ref.UID == instanceCM.UID {
+				return nil
+			}
+		}
+		sec.OwnerReferences = append(sec.OwnerReferences, metav1.OwnerReference{
+			APIVersion: "v1",
+			Kind:       "ConfigMap",
```
xjacka, line 254: 🟢 **Suggestion:** `BlockOwnerDeletion` is not set. Consider setting it to `true` to ensure foreground GC blocks instance-ConfigMap deletion until this Secret is reaped — provides a stronger guarantee that TLS key material is cleaned up before the parent is gone. Current background GC is functionally correct but asynchronous.

xjacka (approved):
🛡️ **DAM** — Code Review @ `ec240980`

## PR #245: fix(controller): cascade-delete envoy leaf TLS Secret with instance
**Author:** pilartomas | **Branch:** fix/envoy-tls-secret-cleanup → main | **Changes:** +106 −0 (2 files)

### Summary
Adds an idempotent OwnerReference patch from the cert-manager-produced envoy leaf TLS Secret to the instance ConfigMap, so that Kubernetes GC cascade-deletes the Secret when the instance is removed. Fixes a resource leak where `<instance>-envoy-tls` Secrets accumulated after instance deletion.

### Findings
- 🟢 **Suggestion:** `BlockOwnerDeletion` is not set on the OwnerReference (`packages/controller/pkg/reconciler/instance.go:249-254`). Setting it to `true` would make instance-ConfigMap deletion block on the Secret being reaped (foreground GC), providing a stronger guarantee that TLS key material is cleaned up before the parent is gone. Current behavior (background GC) is functionally correct but asynchronous — consider whether the stronger guarantee is worth the tradeoff.
- ✅ **Looks good:** `retry.RetryOnConflict` closure re-fetches the Secret on each attempt — resourceVersion is always fresh, no stale-write conflicts.
- ✅ **Looks good:** Idempotency guard (UID loop before append) correctly prevents duplicate OwnerReferences across reconciles.
- ✅ **Looks good:** Non-fatal error path (`slog.Warn + continue`) is correct — cert-manager may not have issued the Secret yet on the first reconcile.
- ✅ **Looks good:** Three focused tests cover all three paths: ref added, idempotent, and no-op when Secret absent.

### Verdict
APPROVE — Small, focused fix with clean implementation mirroring the existing `ensureAgentOwnerReference` pattern, thorough test coverage, and proper idempotency and retry handling.

---

jezekra1 (approved):

Issue #42, opened 2026-05-04
[Bug]: Keycloak login redirect storm opens dozens of browser tabs


pilartomas:
## Problem

When the DAM UI is left open long enough for the Keycloak session to expire, returning to the tab triggers a cascade of navigations to the Keycloak login page — users have reported the browser opening ~50 tabs in rapid succession, enough pressure to make the laptop unresponsive and, in one case, crash.

The cascade happens because the UI's silent token-refresh path cannot succeed under the current configuration, so every in-flight action that needs a token — periodic data refreshes, queued API calls, WebSocket reconnection, session-config fetches — independently falls through to a full-page redirect to Keycloak, with no coordination between them.

## Proposed solution

Silent token refresh should actually work, so expiring tokens are handled invisibly without any redirect in the common case. When a redirect to the login page is truly needed, the UI should redirect exactly once — all concurrent callers should share the same redirect, not each trigger their own. After logging in, the user should land back where they were.

### Scope
Frontend authentication flow only. No backend or Keycloak configuration changes.

jenna-winkler:
Can this be reproduced? If not. lets close for now

kapetr:
I wasn't able to reproduce it.

PR #284, opened 2026-05-19
fix(agent-runtime): bump claude-agent-acp to fix claude.exe SIGSEGV

matoushavlena:
## Summary

- Bump `@agentclientprotocol/claude-agent-acp` 0.24.2 → 0.33.1.
- Drop the explicit `@anthropic-ai/claude-agent-sdk` install from the three agent Dockerfiles (claude-code, code-guardian, google-workspace).

Both changes are load-bearing — see below.

## Why

Long-lived sessions died with:

```
No onPostToolUseHook found for tool use ID: toolu_...
Session ...: Claude Agent process died: Claude Code process terminated by signal SIGSEGV
```

Root cause is a JS-protocol-layer / native-binary version mismatch inside the agent pod:

- The Dockerfile ran `npm install @agentclientprotocol/claude-agent-acp @anthropic-ai/claude-agent-sdk` on top of the pnpm-installed `/app` from platform-base. The unversioned `@anthropic-ai/claude-agent-sdk` install pulled the latest SDK (0.3.144) into the **top level** of `/app/node_modules`, dragging in its paired native binary `claude-agent-sdk-linux-arm64@0.3.144` — the actual `claude` executable.
- `claude-agent-acp@0.24.2` bundles its own nested `@anthropic-ai/claude-agent-sdk@0.2.84`. That nested copy has no nested native-binary subpackage, so when the 0.2.84 JS layer went to spawn `claude`, Node's module resolution walked **up** to the top-level `claude-agent-sdk-linux-arm64@0.3.144`.
- Net result inside the pod: **JS protocol layer at 0.2.84 talking to a native binary at 0.3.144** — different control-protocol dialects. The CLI tolerated mismatched frames for a while, eventually segfaulted on a hook callback it didn't understand. The `No onPostToolUseHook found` log was the dialect-mismatch symptom; the SIGSEGV was the binary crashing.

The fix aligns JS layer and native binary:

1. Bumping `claude-agent-acp` to 0.33.1 ships JS SDK 0.2.132.
2. Removing the explicit SDK install lets the umbrella's transitive resolution decide the top-level version; the paired native binary lands at the matching 0.2.132.

Verified in the rebuilt pod:

```
/app/node_modules/@anthropic-ai/claude-agent-sdk           → 0.2.132
/app/node_modules/@anthropic-ai/claude-agent-sdk-linux-arm64 → 0.2.132
```

Either change alone would not fix it: bumping acp without dropping the explicit install would re-introduce the mismatch on every rebuild; dropping the install without bumping acp would leave us on the 0.2.84 line.

`0.33.1` is the newest version of `claude-agent-acp` that satisfies the repo's 7-day `minimumReleaseAge` policy. `0.35.0+` moves the SDK to the 0.3.x line (paired with newer CLI builds) and would be a more aggressive fix; holding it in reserve in case 0.2.132 ↔ 2.1.144 turns out to have its own quirks.

## What I tried first (and removed)

Initial hypothesis was a race between our `unstable_closeSession` call and post-turn teardown in the CLI. Shipped a 30s grace-period delay before idle-close; the SIGSEGV still reproduced. Timestamped logs showed the crash firing 36s after the deferred close, and in some runs with no close logged at all — proving our close call wasn't the trigger. Those commits were dropped from this PR.

## Test plan

- [x] `mise run check` — typecheck + lint clean.
- [x] `mise run test` — full suite passes.
- [x] `mise run cluster:build-agent` — all three agent images build with the bumped dep, no ERESOLVE.
- [x] Manual cluster verification — same scenario that previously crashed now runs without SIGSEGV.

### DIFF
--- a/packages/agent-runtime/package.json
+++ b/packages/agent-runtime/package.json
@@ -11,7 +11,7 @@
     "test": "vitest run"
   },
   "dependencies": {
-    "@agentclientprotocol/claude-agent-acp": "^0.24.2",
+    "@agentclientprotocol/claude-agent-acp": "^0.33.1",
     "@trpc/server": "^11.16.0",
     "@xterm/addon-serialize": "0.14.0",
     "@xterm/headless": "5.5.0",
--- a/packages/agents/claude-code/Dockerfile
+++ b/packages/agents/claude-code/Dockerfile
@@ -3,7 +3,7 @@ FROM ${BASE_IMAGE}
 
 USER root
 
-RUN npm install @agentclientprotocol/claude-agent-acp @anthropic-ai/claude-agent-sdk \
+RUN npm install @agentclientprotocol/claude-agent-acp \
     && npm install -g @anthropic-ai/claude-code \
     && chown -R 65532:0 /app
 
--- a/packages/agents/code-guardian/Dockerfile
+++ b/packages/agents/code-guardian/Dockerfile
@@ -3,7 +3,7 @@ FROM ${BASE_IMAGE}
 
 USER root
 
-RUN npm install @agentclientprotocol/claude-agent-acp @anthropic-ai/claude-agent-sdk \
+RUN npm install @agentclientprotocol/claude-agent-acp \
     && npm install -g @anthropic-ai/claude-code \
     && chown -R 65532:0 /app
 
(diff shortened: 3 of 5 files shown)


matoushavlena:
Closing — the fix landed via #42 (commit 9c877e0b on main) which made the same  bump and Dockerfile cleanup. After rebasing, this PR is empty (lockfile binary diff only). See the PR description for the root-cause writeup; preserving it here for the record.

matoushavlena:
Correction to the previous comment: the package whose bump landed via #42 is `@agentclientprotocol/claude-agent-acp` (shell ate the backticks last time).

PR #291, opened 2026-05-19
fix(ui): show agent name in chat header instead of instance id

### LINKED ISSUE #210: UI - Chat header shows instance id instead of the agent's name
## Problem

When opening a chat with an agent, the header at the top of the conversation displays the instance id rather than the name the user gave the agent when creating it. Ids look like random strings, so there's no quick way to tell from the header which agent the user is talking to — they have to remember which id maps to which name, or navigate back to the agents list to check.

## Proposed solution

The chat header should display the agent's name. The instance id can be removed from the header entirely — the name uniquely identifies the agent in the user's mental model, and the id is an implementation detail not useful to the user and is available in the url.

kapetr:
## Summary
- The chat header was rendering the instance id (a random-looking string), making it hard to tell which agent the user was talking to.
- Now resolves the instance from the loaded list and renders its `name`, falling back to the id if the list hasn't loaded yet.

Closes #210

## Test plan
- [ ] Open a chat for an agent — header shows the agent's name, not its id.
- [ ] Reload mid-fetch — falls back to id until `useInstances()` resolves, then swaps to the name.

### DIFF
--- a/packages/ui/src/modules/sessions/views/chat-view.tsx
+++ b/packages/ui/src/modules/sessions/views/chat-view.tsx
@@ -42,6 +42,8 @@ export function ChatView() {
   const selectedInstance = useStore((s) => s.selectedInstance);
   const { data: instancesData } = useInstances();
   const instances = instancesData?.list ?? [];
+  const selectedInstanceName =
+    instances.find((i) => i.id === selectedInstance)?.name ?? selectedInstance;
   const sessionId = useStore((s) => s.sessionId);
   const sessionMode = useStore((s) => s.sessionMode);
   const setSessionMode = useStore((s) => s.setSessionMode);
@@ -356,7 +358,7 @@ export function ChatView() {
           </button>
           <span className="w-px h-4 bg-border-light" />
           <h1 className="text-[14px] font-bold text-text truncate">
-            {selectedInstance}
+            {selectedInstanceName}
           </h1>
           <div className="flex h-7 rounded-md border border-border-light overflow-hidden">
             <button


packages/ui/src/modules/sessions/views/chat-view.tsx
```diff
@@ -42,6 +42,8 @@ export function ChatView() {
   const selectedInstance = useStore((s) => s.selectedInstance);
```
xjacka, line 42: 🟢 **Suggestion:** This file calls `useStore((s) => ...)` inline ~14 times (pre-existing pattern, not introduced by this PR). Consider extracting exported selector hooks (`useSelectedInstance`, `useSessionId`, etc.) per the state-management rules in a follow-up refactor PR.

xjacka (approved):
🛡️ **DAM** — Code Review @ `58c0a7c`

## PR #291: fix(ui): show agent name in chat header instead of instance id
**Author:** kapetr | **Branch:** fix/chat-header-agent-name → main | **Changes:** +3 −1 (1 file)

### Summary
This PR fixes the chat header to show the agent's human-readable name instead of its raw instance ID. It derives `selectedInstanceName` from the already-loaded `useInstances()` data (falling back to the ID while the list is still loading) and renders that in the `<h1>`.

### Findings
- ✅ **Looks good:** Fallback `?? selectedInstance` correctly shows the ID during loading, then switches to the name — matches the PR's test plan.
- ✅ **Looks good:** No state lineage violations — server state stays in TanStack Query, UI selection in Zustand, derivation computed inline.
- 🟢 **Suggestion:** `chat-view.tsx` calls `useStore((s) => ...)` inline ~14 times (pre-existing, not introduced by this PR). Consider extracting exported selector hooks (`useSelectedInstance`, `useSessionId`, etc.) per the state-management rules in a follow-up refactor PR. (`packages/ui/src/modules/sessions/views/chat-view.tsx:42`)

✅ No findings on the changed code.

### Verdict
APPROVE — Correct fix with clean implementation and safe fallback behavior; no blocking issues.

---

packages/ui/src/modules/sessions/views/chat-view.tsx
kapetr: This is everywhere in the codebase, we need bigger refactor.

PR #307, opened 2026-05-20
fix: zero-downtime rollouts and graceful UI degradation during deploys

### LINKED ISSUE #293: Deployments cause visible downtime and hard errors in the UI
## Problem

Every time the platform is updated, the API server goes offline for several seconds — sometimes longer. During this window, users see error states in the UI with no indication that anything is happening. Active sessions appear broken, and there's no way to tell whether it's a transient hiccup or a real outage.

The experience from the Slack thread: "everything errors out and I don't even know a deployment is happening."

This is a routine dev-environment problem today, but the same gap exists in production: any upgrade causes a visible interruption.

## Proposed solution

Two separate improvements, both needed:

**Zero-downtime rollouts** — the platform should use a rolling deployment strategy so the old version stays up and serving traffic until the new version is healthy and ready. No gap, no visible downtime window.

**Graceful degradation in the UI** — when the API is briefly unreachable (e.g., the tail end of a rollout or a slow pod startup), the UI should retry transparently and show a neutral "reconnecting…" state rather than hard error screens. Once the API comes back, the UI recovers automatically without a page reload.

### Scope

- Applies to the API server and UI services; does not affect agent pods (they have their own lifecycle).
- The UI change covers transient network/API errors generally, not just deployment windows.

### Out of scope

- Database migrations that require downtime — those need a separate migration-safety story.
- Multi-replica scaling — fixing the rollout strategy is independent of how many replicas are run.

JanPokorny:
## Summary

Closes #293

- **Helm**: Switch api-server deployment from `Recreate` to `RollingUpdate` (`maxSurge: 1, maxUnavailable: 0`), add `preStop: sleep 5` lifecycle hook, and set `terminationGracePeriodSeconds: 45` — old pod stays up until the new one passes readiness probes.
- **UI**: Add API health detection via a custom fetch wrapper on the tRPC `httpBatchLink`. Detects network failures instantly (not after retry exhaustion). Shows a "Reconnecting to server" banner when the API is unreachable, polls `/api/health` every 3s, and auto-recovers without a page reload. Suppresses error toasts while reconnecting.

## Test plan

- [x] `mise run check` — lint + type-check pass
- [x] `mise run helm:check:lint` + `mise run helm:check:render` — Helm lint and kubeconform pass
- [x] Verified deployment strategy, `preStop` hook, and `terminationGracePeriodSeconds` via `kubectl get deployment`
- [x] Continuous health poll (405 requests at 200ms interval) during `kubectl rollout restart` — **0 failures** (zero downtime)
- [x] Loaded UI in browser, scaled api-server to 0 — reconnecting banner appeared
- [x] Scaled api-server back to 1 — banner auto-dismissed, UI recovered without page reload


packages/ui/src/api.ts
```diff
@@ -2,11 +2,23 @@ import { createTRPCClient, httpBatchLink } from "@trpc/client";
...
 
 import { getAccessToken } from "./auth.js";
+import { onFetchError, onFetchSuccess } from "./lib/api-health.js";
 
 export const api = createTRPCClient<AppRouter>({
   links: [
     httpBatchLink({
       url: "/api/trpc",
+      fetch: async (input, init) => {
+        try {
```
xjacka, line 12: 🟢 **Suggestion:** `onFetchError()` fires on any `response.status >= 500`, so two unrelated application-level 500s from different endpoints in sequence would incorrectly show the reconnecting banner. Consider narrowing the response-path trigger to `503 Service Unavailable` only — `catch` already handles true network errors:

```suggestion
          else if (response.status === 503) onFetchError();
```

packages/ui/src/lib/api-health.ts
```diff
@@ -0,0 +1,72 @@
...
+  if (pollTimer) return;
+  pollTimer = setInterval(async () => {
+    if (!navigator.onLine) return;
+    try {
+      const res = await fetch("/api/health");
+      if (res.ok) {
+        failureCount = 0;
+        stopPolling();
+        setStatus("connected");
+      }
```
xjacka: 🟢 **Suggestion:** `setInterval` with an `async` callback allows overlapping health checks if a request stalls longer than 3 s. A recursive `setTimeout` — schedule the next poll inside the `finally` block — ensures at most one in-flight check at a time:

```suggestion
  pollTimer = setTimeout(async function poll() {
    if (!navigator.onLine) {
      pollTimer = setTimeout(poll, 3_000);
      return;
    }
    try {
      const res = await fetch("/api/health");
      if (res.ok) {
        failureCount = 0;
        pollTimer = null;
        setStatus("connected");
        return;
      }
    } catch {
      // still down
    }
    pollTimer = setTimeout(poll, 3_000);
  }, 3_000);
```

xjacka (approved):
🛡️ **DAM** — Code Review @ `41517eb`

## PR #307: fix: zero-downtime rollouts and graceful UI degradation during deploys
**Author:** JanPokorny | **Branch:** fix/zero-downtime-rollout#locki-mtlgghpp → main | **Changes:** +136 −2 (7 files)

### Summary
Switches the api-server Kubernetes deployment from `Recreate` to `RollingUpdate` (maxSurge: 1 / maxUnavailable: 0) with a `preStop: sleep 5` hook and `terminationGracePeriodSeconds: 45` to eliminate downtime during rollouts. Pairs this with a UI-side API health layer: a custom `fetch` wrapper on `httpBatchLink` detects unreachability, triggers a "Reconnecting to server" banner after two consecutive failures, polls `/api/health` every 3 s, auto-recovers and invalidates all queries on reconnect, and suppresses error toasts during the outage window.

### Findings

- 🟢 **Suggestion:** `onFetchError()` fires on any `response.status >= 500`, so two unrelated application-level 500s from different endpoints in sequence would incorrectly show the reconnecting banner. Consider narrowing the response-path trigger to `503 Service Unavailable` only (keeping the `catch` branch to cover network-level errors, which do reliably indicate server unavailability). (`packages/ui/src/api.ts:12`)

- 🟢 **Suggestion:** The `setInterval` callback in `startPolling` is `async`, so if a health-check stalls past the 3 s window (e.g., on a slow network), a second poll fires while the first is still awaiting. A recursive `setTimeout` — schedule the next poll inside the `finally` block — ensures at most one in-flight health check at any time. (`packages/ui/src/lib/api-health.ts:19`)

- ✅ **Looks good:** The fetch wrapper correctly lets 4xx responses pass through without affecting health state (server is up and responding; those are application errors), while network throws and 5xx responses contribute to the failure counter.

- ✅ **Looks good:** `useSyncExternalStore(subscribeApiHealth, getApiHealthSnapshot)` is the idiomatic primitive for subscribing React components to external mutable state. Subscription teardown returns the unsubscribe function correctly.

- ✅ **Looks good:** `queryClient.invalidateQueries()` on `reconnecting → connected` transition ensures components see fresh data after the server returns instead of stale cache snapshots.

- ✅ **Looks good:** Helm changes — `maxSurge: 1 / maxUnavailable: 0` keeps the old pod alive until the new one passes readiness probes; `preStop: sleep 5` gives kube-proxy time to drain the endpoint before SIGTERM; `terminationGracePeriodSeconds: 45` leaves 40 s for in-flight request completion after SIGTERM. Correct sequencing for zero-downtime rollouts.

✅ No findings.

✅ No findings.

### Verdict
APPROVE — Well-executed zero-downtime deployment configuration paired with a clean, minimal UI resilience layer; findings are suggestions only with no correctness or safety concerns.

---

PR #319, opened 2026-05-20
fix(api): add test coverage for Anthropic authMode rotation

### LINKED ISSUE #136: Anthropic secret rotation doesn't re-derive authMode from new value
## Problem

Rotating an Anthropic secret value via `secrets.update({ value })` does **not** re-derive `authMode` from the new token. If a user rotates from an OAuth token (`sk-ant-oat…`) to an API key (`sk-ant-api…`) — or vice versa — the stored `agent-platform.ai/auth-mode` annotation stays at its previous value, `resolveInjection` keeps emitting the old header shape, and the SDS file is re-baked with the wrong header. Anthropic returns 401.

Same drift class as the recently-fixed generic `injectionConfig` issue, but on a different field. The router-level guard from that fix doesn't apply because `authMode` is server-derived from the value prefix in `secretsService.create`, not user-supplied.

## Where it lives

- `packages/api-server/src/modules/secrets/services/secrets-service.ts:104-113` — `update` passes `patch.value` straight through without recomputing `authMode`.
- `packages/api-server/src/modules/secrets/services/secrets-service.ts:72-77` — the create-time derivation that needs to be reused on update for `type === "anthropic"`.

## Proposed fix

In `secretsService.update`, when the existing secret is `type=anthropic` and `patch.value` is provided, re-derive `authMode` from the new value (same `sk-ant-oat` vs `sk-ant-api` discriminator as create) and pass it through to the K8s port. The K8s port already accepts `authMode` on the update path; only the service layer needs the wiring.

## Acceptance

Rotate an Anthropic secret from OAuth → API key (and reverse). Assert that:

- `auth-mode` annotation flips
- `injection-header-name` flips between `Authorization` and `x-api-key`
- SDS `inline_string` reflects the new shape (`Bearer …` vs bare key)

JanPokorny:
## Summary

- Adds test coverage for `oauth → api-key` authMode rotation on Anthropic secret update (verifies authMode flips, envMappings rewrite, injectionConfig cleared)
- Adds test for `oauth → oauth` no-op case (no unnecessary mutation when mode stays the same)

The service-layer fix (`anthropicModeRotationFor`) was already in place — this closes the gap in test coverage.

Closes #136

## Test plan

- [x] `mise run check` passes
- [x] `mise run test` passes (257 api-server tests)

### DIFF
--- a/packages/api-server/src/__tests__/unit/secrets-service-update-anthropic-mode.test.ts
+++ b/packages/api-server/src/__tests__/unit/secrets-service-update-anthropic-mode.test.ts
@@ -24,6 +24,10 @@ const ANTHROPIC_OAUTH_MAPPING = PROVIDERS.anthropic.modes.find(
   (m) => m.key === "oauth",
 )!.defaultEnvMappings;
 
+const ANTHROPIC_API_KEY_MAPPING = PROVIDERS.anthropic.modes.find(
+  (m) => m.key === "api-key",
+)!.defaultEnvMappings;
+
 function makePort(existing: K8sStoredSecret) {
   const store = new Map([[existing.id, existing]]);
   const updates: UpdateCall[] = [];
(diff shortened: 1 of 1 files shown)


xjacka (approved):
🛡️ **DAM** — Code Review @ `e567df7`

## PR #319: fix(api): add test coverage for Anthropic authMode rotation
**Author:** JanPokorny | **Branch:** fix/secret-rotation-authmode → main | **Changes:** +42 −0 (1 file)

### Summary
Adds two unit-test cases to `secrets-service-update-anthropic-mode.test.ts`: one verifies that an `oauth → api-key` rotation flips `authMode`, rewrites `envMappings`, and clears `injectionConfig`; the other verifies the `oauth → oauth` no-op path (no `authMode` / `envMappings` / `injectionConfig` mutation). The service-layer fix (`anthropicModeRotationFor`) was already in place — this closes the test-coverage gap for #136.

### Findings
- ✅ **Looks good:** Both new tests share the same `makePort` / `makeGrants` / `createSecretsService` scaffolding as the existing cases — no parallel test setup, no ad-hoc mocks.
- ✅ **Looks good:** `oauth → api-key` assertion uses `defaultEnvMappings` derived from `PROVIDERS.anthropic.modes` (the same source of truth the service uses) rather than a hand-built literal, so a future change to the api-key env shape only requires one update.
- ✅ **Looks good:** No-op case asserts the three relevant fields are *undefined* (i.e. the service deliberately did not touch them) rather than asserting deep equality, which is the right granularity for a no-op test.
- ✅ **Looks good:** Test value `sk-ant-api03-newkey` and `sk-ant-oat01-rotated` are realistic prefixes for the respective auth modes, matching what the prefix detector keys on.

✅ No findings.

### Verdict
APPROVE — Pure test-coverage close on a service-layer fix that already shipped. Asserts the right invariants, reuses existing scaffolding, no behavior change.

---

PR #348, opened 2026-05-21
feat: lazy per-directory workspace fetch (#342)

### LINKED ISSUE #342: Fetch the file tree lazily by directory instead of all at once
## Problem

The Workspace panel loads the agent's entire working directory tree on every poll, even though most of it never reaches the screen. Hidden dot-directories (`.claude/`, `.uploads/`, `.git`-adjacent state, etc.) and any collapsed subtree still ship over the wire just to be filtered out client-side.

For lightweight workspaces this is fine. Once an agent checks out a real repository or accumulates a deep tree, the payload grows unbounded — every 2-second poll re-sends thousands of paths the user can't see and won't expand. That makes the panel sluggish to refresh, wastes bandwidth on slow connections, and gets worse the longer an agent runs.

## Proposed solution

Load the tree the way a file explorer does: only what's visible right now.

- On open, fetch just the **top-level** entries of the working directory.
- When the user expands a folder, fetch **that folder's immediate children**.
- When something changes (create, rename, delete, upload), refresh **only the affected directory**, not the whole tree.

User-visible outcome: the Workspace panel feels instant on large workspaces, the polled refresh stays cheap regardless of repo size, and expanding a folder briefly shows its contents loading rather than waiting on the full tree.

### Scope

- Applies to the file tree view in the Workspace panel.
- Default-collapsed dot-directories stay collapsed and don't fetch anything until the user opens them.

### Out of scope

- Search across the whole workspace — needs its own design once the tree stops being preloaded.
- Watching for filesystem changes via push instead of polling — separate concern.

tomkis:
## Summary

- Replace `files.tree` (whole-tree poll) with `files.listDirs({ paths })` (per-dir batched snapshot). Implements ADR-047.
- Workspace panel switches to collapsed-by-default + lazy fetch on expand; per-agent expanded set in Zustand drives polling.
- Shrink server-side `EXCLUDE` to `RESERVED = { .triggers, .initialized }` only (ADR-048). `.git/`, `node_modules/`, `.DS_Store`, `.npm/`, `.claude.json` become listable and writable.
- Stale paths silently dropped from expanded set on `not-found`; rename rewrites expanded paths under the renamed prefix.

Closes #342.

## Test plan

- [x] `mise run check` passes (verified locally).
- [x] `mise run test` passes — 542 tests across packages (verified locally).
- [x] Smoke in browser against a running cluster:
  - [x] Open Workspace panel — only top-level entries fetch on first open.
  - [x] Click to expand a folder — its immediate children appear; deeper folders stay collapsed.
  - [x] Expand multiple folders — verify only one HTTP request per 2s carrying all open paths.
  - [x] `.git/`, `node_modules/` appear at root (collapsed); expand and confirm contents load.
  - [x] Edit a file inside `node_modules/`, save — confirm it persists (writes now allowed under ADR-048).
  - [x] Confirm writes to `.triggers/` and `.initialized` are refused.
  - [x] Rename a folder whose subfolder is expanded — confirm the subfolder stays expanded under the new path.
  - [x] Delete an expanded folder — confirm it disappears and polling stops for it.


packages/ui/src/modules/files/components/dir-contents.tsx
```diff
@@ -0,0 +1,76 @@
...
+                placeholder={
+                  panel.pendingNew.kind === "dir" ? "new-folder" : "new-file.md"
+                }
+                onCommit={panel.onCommitNew}
+                onCancel={panel.onCancelNew}
+              />
+            )}
+          </Fragment>
+        );
+      })}
```
xjacka: 🔴 **Critical:** This `.map()` body has 3 local derivations (`fullPath`, `isExpanded`, `isRenaming`), 3 conditional render branches, and ~38 lines of JSX — all criteria for list-item extraction per react-ui-engineering.

Extract to a `<DirEntryRow>` component that receives `entry`, `parentPath`, and `depth`, calls `useFilesPanel()` internally, and owns the derivations:

```tsx
{snapshot.entries.map((entry) => (
  <DirEntryRow key={joinPath(path, entry.name)} parentPath={path} entry={entry} depth={depth} />
))}
```

packages/agent-runtime/src/modules/files.ts
```diff
@@ -83,6 +65,37 @@ const forbidden = (reason: string): FilesDomainError => ({
...
+
+async function listDir(
+  workingDir: string,
+  rel: string,
+): Promise<DirListResult> {
+  if (touchesReserved(rel)) {
+    return { path: rel, ok: false, error: "forbidden" };
+  }
+  const abs = safePath(workingDir, rel);
+  if (!abs) return { path: rel, ok: false, error: "not-found" };
```
xjacka: 🟡 **Warning:** `safePath` returning null means the path escaped the working directory (traversal attempt). Returning `error: "not-found"` is the wrong code — `"forbidden"` is accurate and already exists in `DirListResult`.

```suggestion
  if (!abs) return { path: rel, ok: false, error: "forbidden" };
```

packages/ui/src/types.ts
```diff
@@ -93,11 +93,15 @@ export interface LogEntry {
...
-export interface TreeEntry {
-  path: string;
+export interface DirEntry {
+  name: string;
   type: "file" | "dir";
 }
 
+export type DirListResult =
+  | { path: string; ok: true; entries: DirEntry[] }
+  | { path: string; ok: false; error: "not-found" | "forbidden" };
```
xjacka: 🟡 **Warning:** `DirEntry` and `DirListResult` are already exported from `agent-runtime-api` (a `workspace:*` dependency). Declaring them locally creates two sources of truth.

Replace with:
```ts
export type { DirEntry, DirListResult } from "agent-runtime-api";
```

packages/ui/src/modules/files/api/queries.ts
```diff
@@ -1,14 +1,17 @@
...
+import {
+  keepPreviousData,
+  useMutation,
+  useQuery,
+  useQueryClient,
+} from "@tanstack/react-query";
 
 import { queryClient } from "../../../query-client.js";
+import { useStore } from "../../../store.js";
+import type { DirListResult } from "../../../types.js";
```
xjacka: 🟡 **Warning:** Imports `DirListResult` from the local re-declaration rather than from `agent-runtime-api` directly. Once `types.ts` re-exports from `agent-runtime-api`, update this import too:

```suggestion
import type { DirListResult } from "agent-runtime-api";
```

packages/ui/src/modules/files/components/files-panel.tsx
```diff
@@ -1,326 +1,55 @@
...
-    (path: string) => {
-      const parts = path.split("/");
-      for (let i = 1; i < parts.length; i++) {
-        if (isDirCollapsed(parts.slice(0, i).join("/"))) return false;
-      }
-      return true;
-    },
-    [isDirCollapsed],
-  );
+  const c = useFilesPanelController({ onOpenFile });
```
xjacka: 🟡 **Warning:** Single-letter name `c` carries no intent — `c.openFile`, `c.ctxValue`, `c.menu` all read as opaque.

```suggestion
  const controller = useFilesPanelController({ onOpenFile });
```

packages/ui/src/modules/files/components/files-panel.tsx
```diff
@@ -1,326 +1,55 @@
...
   }
 
-  // Panel-level overlay only when the pointer isn't over a specific row; that
-  // row has its own highlight (see FileRow).
-  const showPanelOverlay = panelDragActive && dragTargetPath === null;
-
-  const visibleEntries = sortedTree.filter((entry) => isVisible(entry.path));
-  const menuEntry = menu ? fileTree.find((e) => e.path === menu.path) : null;
-
   return (
```
xjacka, line 28: 🟡 **Warning:** The main return is ~95 lines of JSX, above the ~60-line feature-component target. The toolbar (lines 54–86) is stateless and independently nameable — extracting it to `<FilesPanelToolbar>` would bring this return under 60 lines. (The file already dropped from 326 → 127 lines in this PR — this is the remaining seam.)

xjacka (dismissed):
🛡️ **DAM** — Code Review @ `e3aa292`

## PR #348: feat: lazy per-directory workspace fetch (#342)
**Author:** tomkis | **Branch:** feat/lazy-workspace-fetch → main | **Changes:** +802 −449 (17 files)

### Summary
Replaces the whole-tree `files.tree` poll with a per-directory batched `files.listDirs` API, shrinks the server-side reserved-path set from 8 entries to 2 (`.triggers/`, `.initialized`), and migrates the workspace panel to collapsed-by-default lazy rendering backed by Zustand-persisted expansion state. Implements ADR-047 (lazy fetch) and ADR-048 (platform-reserved paths).

### Findings
- 🔴 **Critical:** `DirContents.map()` body has 3 local derivations (`fullPath`, `isExpanded`, `isRenaming`), 3 conditional render branches, and ~38 lines of JSX per item — all criteria for list-item extraction per react-ui-engineering. Extract to `<DirEntryRow>`. (`packages/ui/src/modules/files/components/dir-contents.tsx:29-73`)
- 🟡 **Warning:** `listDir` returns `{ error: "not-found" }` when `safePath` returns null (path traversal attempt), but `"forbidden"` is the correct error — `"not-found"` implies the path could exist elsewhere, misleading clients and auditors. The `"forbidden"` variant already exists in `DirListResult`. (`packages/agent-runtime/src/modules/files.ts:81`)
- 🟡 **Warning:** `DirEntry` and `DirListResult` are re-declared in `packages/ui/src/types.ts` (lines 96–103) instead of being imported from `agent-runtime-api`, which already exports them. `queries.ts` imports from the local re-declaration on line 10. Both types are identical today, but divergence on future schema changes is silent. (`packages/ui/src/types.ts:96-103`, `packages/ui/src/modules/files/api/queries.ts:10`)
- 🟡 **Warning:** `const c = useFilesPanelController(...)` — single-letter name carries no intent. `const controller = ...` makes every downstream reference self-documenting. (`packages/ui/src/modules/files/components/files-panel.tsx:17`)
- 🟡 **Warning:** `files-panel.tsx` main return is ~95 lines of JSX, above the ~60-line feature-component target. The toolbar (lines 54–86) is independently nameable and stateless — extracting it to `<FilesPanelToolbar>` brings the return under 60 lines. (`packages/ui/src/modules/files/components/files-panel.tsx:29-125`)
- ✅ **Looks good:** `EXCLUDE → RESERVED` shrink is well-motivated, cleanly documented in ADR-048, and the symmetric read/write guard is a security improvement.
- ✅ **Looks good:** `withoutPath` correctly prunes the entire expanded subtree on collapse, preventing phantom poll entries for hidden children.
- ✅ **Looks good:** `keepPreviousData` prevents content flickering when the expanded-set key changes, and the sorted `paramsForExpanded` ensures key stability across renders.
- ✅ **Looks good:** New ADR-047 and ADR-048 are well-written, cross-reference each other, and the file key hierarchy (`root → tree → treeForPaths`) is a clean design for targeted invalidation.

**Finding — [Warning] Type duplication violates single-source-of-truth rule**

`DirEntry` and `DirListResult` are declared locally in `packages/ui/src/types.ts` but are already exported from `agent-runtime-api`, which the UI depends on via `"agent-runtime-api": "workspace:*"`. The server implementation correctly imports them from `agent-runtime-api`; the UI should do the same.

Fix: remove local declarations from `types.ts` and re-export from `agent-runtime-api`; update the import in `queries.ts` to source from `agent-runtime-api` directly.

**Finding 1 — [CRITICAL]** `DirContents` `.map()` body has 3 local derivations and 3 conditional branches — extract to `<DirEntryRow>`.

**Finding 2 — [HIGH]** Rename `c` → `controller` in `files-panel.tsx:17`.

**Finding 3 — [HIGH]** `files-panel.tsx` main return is ~95 lines of JSX. Extract toolbar to `<FilesPanelToolbar>`. File dropped from 326 → 127 lines in this PR — this is the remaining extraction seam.

### Verdict
REQUEST_CHANGES — `DirContents` has a CRITICAL list-item extraction missing, and the type duplication / error-code mismatch should be fixed before merge.

---

JanPokorny:
Since `RESERVED` are our things anyway, can we just drop them "out of sight" and avoid filtering the view entirely? E.g. in `~/.local/state/dam`

tomkis:
> Since `RESERVED` are our things anyway, can we just drop them "out of sight" and avoid filtering the view entirely? E.g. in `~/.local/state/dam`

I'd rather not to.

One of the things we keep in home is `. initialized` which exists just to inform the system about home dir being ready.

packages/agent-runtime/src/modules/files.ts
```diff
@@ -23,15 +25,11 @@ import { err, ok } from "agent-runtime-api";
...
-  ".claude.json",
-  ".initialized",
-  "node_modules",
-  ".DS_Store",
-]);
+/** Platform-reserved paths under the working directory. The controller writes
+ *  trigger payloads to `.triggers/` (ADR-008) and uses `.initialized` as a
+ *  setup marker; user reads/writes against either would break agent lifecycle.
+ *  See ADR-049. Repo noise (.git, node_modules, .DS_Store, …) is surfaceable. */
+const RESERVED = new Set([".triggers", ".initialized"]);
```
jezekra1, line 32: > user reads/writes against either would break agent lifecycle

nitpick: not sure if the solution is to hide the files then, they can be easily overwritten by the agent and it's more likely if the user doesn't even know about them

docs/adrs/048-lazy-workspace-fetch.md
```diff
@@ -0,0 +1,29 @@
+# ADR-048: Lazy per-directory workspace fetch
```
jezekra1: nitpick: Needs renumbering

jezekra1 (approved):

PR #370, opened 2026-05-22
fix(ui): adding a close-race for selected acp request methods to prevent hangs

### LINKED ISSUE #363: bug: Agent switches from `running` back to `starting` during startup
# Problem
When a new agent is created and opened, it comes up showing running and appears to be genuinely usable for a brief moment. Within seconds, the agent's state flips back to starting and then to running again, and from that point the UI is in a broken state:

Messages typed into the chat are accepted by the UI but never reach the agent.
There is no error or recovery prompt.
The only way to make the agent work again is to reload the webpage. After a reload, everything behaves normally.

This happens consistently on every fresh agent creation. 

# Proposed solution
A newly created agent should be stably usable from the moment it first becomes interactive — without the user ever needing to reload the page.

Concretely:

A fresh agent should not advertise itself as running, become usable, then silently break. If the underlying system needs a second startup step, that should happen before the UI says the agent is ready.
If a real disconnect or restart does happen, the UI must recover on its own. Reloading the page is never the expected recovery path.

# Investigation

Whether the underlying pod actually restarts on first creation, and if so, why. The fix may turn out to be "don't restart" rather than "tolerate the restart."
Whether the same break-after-first-ready pattern affects hibernated agents being woken, or only fresh creates.

#Out of scope

The broader UI lifecycle gating tracked in #168 and the streaming-reliability work tracked in #26 — those are real, separate problems. This ticket is specifically about the first-run instability and the reload-only recovery path it forces.

jjeliga:
The ACP SDK's `Connection` never rejects pending JSON-RPC requests when its
stream closes — the `#receive` loop just exits and the `#pendingResponses`
map is leaked. Every `ClientSideConnection.*` call that issues a request can
therefore hang forever if the WS dies mid-flight.

Most visible symptom (#363): pod rolls during the user's first prompt → WS
dies → `conn.prompt(...)` hangs → empty assistant bubble, no error, no
recovery. Less visible but worse: same wedge on `initialize`,
`unstable_resumeSession`, `loadSession`, `setSessionMode`, etc. — the
reconnect path itself silently dead-ends.

Fix: a thin Proxy at `openConnection` races each request-style SDK method
against the public `conn.closed` promise and rejects with a
`ConnectionClosedError`. Consumer catch-paths (e.g. sendPrompt's existing
SendErrorCard + Retry) surface the error normally.

Scope: UI only. Does not change when or why pods restart — that's parked.

Closes #363

### DIFF
--- a/packages/ui/src/modules/acp/close-race.ts
+++ b/packages/ui/src/modules/acp/close-race.ts
@@ -1,0 +1,54 @@
+import type { ClientSideConnection } from "@agentclientprotocol/sdk/dist/acp.js";
+
+export class ConnectionClosedError extends Error {
+  readonly name = "ConnectionClosedError";
+  constructor() {
+    super("Connection closed while request was in flight");
+  }
+}
+
+// SDK methods that issue a JSON-RPC request via `Connection.sendRequest`. The
+// SDK's receive loop never rejects `#pendingResponses` on stream close, so
+// these calls hang indefinitely when the WebSocket dies mid-flight. Listed by
+// name so adding a new SDK method is a deliberate decision rather than a
+// silent leak through an unwrapped path.
+const REQUEST_METHODS = new Set<string>([
+  "initialize",
+  "authenticate",
+  "logout",
+  "newSession",
+  "loadSession",
+  "forkSession",
+  "listSessions",
+  "unstable_resumeSession",
+  "closeSession",
+  "setSessionMode",
+  "setSessionModel",
+  "setSessionConfigOption",
+  "prompt",
+  "cancel",
+]);
+
+/**
+ * Wrap a `ClientSideConnection` so every request-style method races against
+ * the connection's `closed` promise. On close, in-flight calls reject with
+ * `ConnectionClosedError` so consumer catch-paths can surface a real error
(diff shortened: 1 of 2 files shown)


packages/ui/src/modules/acp/close-race.ts
```diff
@@ -0,0 +1,54 @@
...
+    super("Connection closed while request was in flight");
+  }
+}
+
+// SDK methods that issue a JSON-RPC request via `Connection.sendRequest`. The
+// SDK's receive loop never rejects `#pendingResponses` on stream close, so
+// these calls hang indefinitely when the WebSocket dies mid-flight. Listed by
+// name so adding a new SDK method is a deliberate decision rather than a
+// silent leak through an unwrapped path.
+const REQUEST_METHODS = new Set<string>([
```
PetrBulanek: Some `REQUEST_METHODS` names don't match the SDK.

`forkSession` → `unstable_forkSession`
`closeSession` → `unstable_closeSession`
`setSessionModel` → `unstable_setSessionModel`
`logout` → `unstable_logout`

PetrBulanek (changes requested):

PetrBulanek (approved):

Issue #239, opened 2026-05-15
Epic - Headless CLI for programmatic agent runs


jenna-winkler:
## Context

DAM's CLI is interactive - login requires a browser (OAuth device flow). There's no way to use DAM from a CI pipeline or orchestration layer without a human present.

Many teams need to run agents programmatically - spin up an agent, give it a task, wait for it to finish, get the output. No human in the loop.

## What's needed 

- Headless auth - API token auth instead of browser-based OAuth, so the CLI works in CI pipelines and automation with no human present
- dam run command - send a prompt to an instance, wait for completion, return the output
-  Async execution - runs can take minutes to hours. Users need to start a run, check status, and retrieve output as separate steps.

## Open questions

- Should we also support a sync run mode that blocks until done?
- How are API tokens generated and managed?
- Exact shape of the run command (subcommands, flags, output format)
- How does the CLI surface run status and output?

## References

- CLI foundation - https://github.com/dam-agents/dam/tree/main/packages/cli

JanPokorny:
Headless "dam run" merged, releasing in next version.

PR #371, opened 2026-05-22
feat(auth): API keys for headless CLI use (ADR-064)

### LINKED ISSUE #249: Headless auth: API keys with scopes for programmatic CLI use
Sub-issue of #239.

## Context

Today the CLI authenticates with Keycloak via the OAuth Device Authorization Grant, with a `DAM_TOKEN` env var bypass that takes a Keycloak access token verbatim ([docs/architecture/cli.md](docs/architecture/cli.md#authentication)). That bypass is fine for ad-hoc CI use but has three problems:

- Keycloak access tokens are short-lived; no in-CLI refresh path when `DAM_TOKEN` is set.
- They carry full user-level permissions; no way to grant narrower access (e.g. "only run this one agent").
- Revoking a token without invalidating the user's whole session is not really a Keycloak workflow.

We need a long-lived, server-managed, scopable credential type.

## Scope

Introduce **API Keys** as a new credential surface:

- New `api-keys` module on the api-server (TS) + contract package. CRUD via tRPC. Plaintext token returned **once on create**, hashed at rest.
- Owner-scoped (`platform.ai/owner=<sub>` on the row). Optional expiry, optional **agent binding** (allowlist of Agent IDs, default `*` — every agent the owner owns now and in future). Agent binding transitively covers each agent's Instance, since the CLI presents agent + instance as one user-facing thing.
- **Permission scopes mirror the user-facing API surface.** The split is driven by a single threat-model question: an exfiltrated `agents:run` key must not be able to permanently change agent behavior. So anything that persists across runs is configuration; anything that just operates the agent in its current configuration is execution.

### Scope definitions

- **`agents:manage`** — agent definitions and per-agent configuration. Anything that persists across runs.
  - Agent CRUD (paired instance is part of the same user-facing thing).
  - Schedules — create, edit, delete.
  - Channels — bind, unbind.
  - Skills — install, uninstall, publish.
  - Egress-rule writes.
  - Granting secrets / connections to an agent (the per-agent grant linkage).
  - Env, model, and other agent-level config.
  - Template listing.

- **`connections:manage`** — global credential lifecycle (not per-agent).
  - Connections (OAuth-issued, e.g. GitHub, MCP) — create, list, delete.
  - Secrets (user-supplied API keys, GitHub PATs) — create, list, delete.
  - The *grant* linkage between a secret and a specific agent lives in `agents:manage`, not here.

- **`agents:run`** — operate an agent in its current configuration.
  - ACP sessions, prompts, run output retrieval.
  - Approvals — resolve pending HITL prompts.
  - pod-files read and write (including `dam import` — CI's natural prep path, and the agent itself can write the same paths during a run, so it's not a new capability).
  - Terminal sessions.

### Other rules

- Request-edge middleware: shared `Authorization: Bearer` slot with Keycloak JWTs, distinguished by a token prefix. Resolves to the same `AuthContext` shape downstream tRPC procedures already consume (`{sub, scopes, agentIds | "*", keyId}`).
- **Scope is re-evaluated on every use** against the owner's current effective permissions — demoting or disabling the user shrinks all their keys without explicit revocation.
- **API keys cannot manage API keys.** Creating, listing, or revoking keys requires an interactive Keycloak session (the browser flow). An exfiltrated key cannot mint or revoke others — that's the safety property worth paying for.
- CLI side: new `dam auth token` group (`create [--name] [--scope] [--agent] [--expires]`, `list`, `revoke <id>`). `DAM_TOKEN` continues to accept either credential type — the prefix tells the server which validator to use, so the CLI need not branch.

## Out of scope

- Workspace/org-level keys (these are per-user only).
- Rotation policy / mandatory expiry — keys can be long-lived in v1.
- A separate read-only scope — read is implicit in each scope. Revisit if a use case appears.

## Open questions

- Should `connections:manage` actually split into `connections:manage` and `secrets:manage`? They're separate server modules but one user-facing concept. Suggest combine.
- Should template listing actually require `agents:manage` at all, or is it read-only enough to be ungated? (It's a read of operator-installed catalog data.)
- Token format and length (suggest `damkey_` prefix + 32 bytes of urandom, base32).
- Hash algorithm (argon2id vs. SHA-256 — SHA-256 is sufficient since the source is high-entropy).
- Whether keys appear in `dam auth status` (proposal: list by id + name + scopes + agent binding, never plaintext).

## References

- Auth foundation: [ADR-039](docs/adrs/039-cli-foundation.md), [docs/architecture/cli.md#authentication](docs/architecture/cli.md#authentication)
- Identity model: [ADR-015](docs/adrs/015-multi-user-auth.md), [docs/architecture/security-and-credentials.md](docs/architecture/security-and-credentials.md)
- Agent ↔ Instance split (why agent-binding is the right granularity): [docs/architecture/cli.md#instance-lifecycle](docs/architecture/cli.md#instance-lifecycle), [tseng/vocabulary.md](tseng/vocabulary.md)

xjacka:
Introduces long-lived owner-scoped API keys as a second credential type sharing the existing Authorization: Bearer slot with Keycloak JWTs, distinguished by a `pk_` prefix. Closes #249 (sub-issue of #239).

- api-keys module on the api-server: token mint (32B urandom, base64url + `pk_` prefix), HMAC-SHA256 keyed with a server-side pepper at rest, owner-scoped CRUD, per-request validator. Plaintext is returned once on create and never persisted.
- Three permission scopes split by threat model: `agents:run` (operate agents in current config), `agents:manage` (config that persists across runs), `connections:manage` (global Connections + Secrets lifecycle). Read is implicit in each scope.
- tRPC procedure builders (runProcedure / manageAgentsProcedure / manageConnectionsProcedure / readAgentProcedure) applied across all 10 module routers; per-key agent allowlists are folded into the procedure builders (`manageAgentByAgentIdProcedure` / `readAgentByAgentIdProcedure` / `runAgentByAgentIdProcedure`), so a forgotten binding check on a future endpoint is a compile error, not a silent privilege escalation. `manageAgentCreateProcedure` rejects non-wildcard keys for `agents.create`.
- Bearer middleware branches on `pk_` prefix; both paths produce a unified UserIdentity carrying scopes, agent binding, and optional keyId. Owner-active check via Keycloak userDirectory.
- "API keys cannot manage API keys": `browserOnlyProcedure` rejects any apiKeys.* request whose principal was authenticated via a key.
- CLI: new `dam auth token {create,list,revoke}` group. DAM_TOKEN continues to accept either credential type — server-side prefix dispatch keeps the CLI branchless.
- UI: new Settings → API Keys tab with create-once-reveal modal.
- Postgres api_keys table (migration 0023); hash uniquely indexed for O(log n) validation, partial index on active rows for list queries.
- Architecture docs updated: cli.md, security-and-credentials.md, persistence.md, platform-topology.md (Motivated by + Last verified bumps + bearer middleware description). ADR-061.


packages/api-server/src/modules/api-keys/domain/types.ts
```diff
@@ -0,0 +1,21 @@
...
+  scopes: readonly Scope[];
+  agentIds: readonly string[] | null;
+  expiresAt: Date | null;
+  createdAt: Date;
+  lastUsedAt: Date | null;
+  revokedAt: Date | null;
+}
+
+export interface ApiKeyValidation {
+  id: string;
```
xjacka: 🟡 **Warning:** `ApiKeyValidation` is dead code — never imported anywhere and structurally identical to `ValidatedApiKey` in `api-key-validator.ts`. Either delete it or promote it as the canonical domain type and import it in the service layer. Dead exported interfaces silently diverge over time.

packages/api-server/src/modules/api-keys/services/api-keys-service.ts
```diff
@@ -0,0 +1,118 @@
...
+    name: r.name,
+    scopes: r.scopes,
+    agentIds: r.agentIds === null ? "*" : r.agentIds,
+    expiresAt: r.expiresAt ? r.expiresAt.toISOString() : null,
+    createdAt: r.createdAt.toISOString(),
+    lastUsedAt: r.lastUsedAt ? r.lastUsedAt.toISOString() : null,
+  };
+}
+
+function generateKeyId(): string {
```
xjacka, line 50: 🟡 **Warning:** `crypto.randomBytes(4)` gives only 32 bits of key-ID space — birthday-paradox collision at ~65k keys. Newer modules in this codebase use `randomUUID()` (122 bits, globally unique). Switch for consistency and safety:

```suggestion
function generateKeyId(): string {
  return randomUUID();
}
```

packages/ui/src/modules/api-keys/components/api-keys-list.tsx
```diff
@@ -0,0 +1,71 @@
+import { KeyRound } from "lucide-react";
+import { useState } from "react";
+
+import { useRevokeApiKey } from "../api/mutations.js";
+import { useApiKeys } from "../api/queries.js";
+import { ApiKeyRow } from "./api-key-row.js";
+import { CreateApiKeyDialog } from "./create-api-key-dialog/index.js";
+
+export function ApiKeysList() {
+  const { data: keys, isLoading } = useApiKeys();
```
xjacka: 🟡 **Warning:** `error` from `useApiKeys()` is not destructured. When the query fails, the component renders nothing — a silent failure the user can't distinguish from loading. Add an error branch:

```suggestion
  const { data: keys, isLoading, error } = useApiKeys();
```
Then render `{error && <p className="text-[13px] text-danger">Failed to load API keys.</p>}` before the loading check.

packages/ui/src/modules/api-keys/components/api-keys-list.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const revokeApiKey = useRevokeApiKey();
+  const [createOpen, setCreateOpen] = useState(false);
+
+  function handleRevoke(id: string, name: string) {
```
xjacka: 🟡 **Warning:** `window.confirm` is synchronous, visually inconsistent, and inaccessible to screen readers. The project already has `Modal` and `CreateApiKeyDialog` — a small `ConfirmRevokeDialog` following those patterns would be consistent and accessible.

packages/ui/src/modules/api-keys/components/create-api-key-dialog/reveal-token.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <p className="text-[13px] text-text-secondary mb-4">
+          This is the only time the token will be shown. If you lose it,
+          revoke this key and create a new one.
+        </p>
+        <div
+          aria-live="polite"
+          className="flex items-stretch gap-2 p-3 rounded-lg bg-surface-raised border border-border-light font-mono text-[12px]"
+        >
+          <code className="flex-1 break-all">{plaintext}</code>
+          <button
```
xjacka, line 46: 🟢 **Suggestion:** `aria-live="polite"` on the container doesn't reliably announce SVG icon changes across all screen readers. Move the accessible label to the button and mark icons as `aria-hidden`:

```suggestion
          <button
            type="button"
            onClick={handleCopy}
            aria-label={copied ? "Copied to clipboard" : "Copy to clipboard"}
            className="shrink-0 px-2 py-1 rounded hover:bg-surface text-text-secondary"
          >
            {copied ? <Check size={16} aria-hidden /> : <Copy size={16} aria-hidden />}
          </button>
```

xjacka (commented):
🛡️ **DAM** — Code Review @ `1d227d7`

## PR #371: feat(auth): API keys with scopes for headless CLI use (ADR-047)
**Author:** xjacka | **Branch:** feat/api-keys-headless-auth → main | **Changes:** +2019 −172 (53 files)

### Summary
This PR implements ADR-047, introducing scoped long-lived API keys for headless CLI authentication. It adds a complete stack: token minting (SHA-256 hash, `damkey_` prefix), a three-scope permission model (`agents:run`, `agents:manage`, `credentials:manage`), per-key agent allowlists, a bearer middleware that dispatches JWT vs API-key paths, a CLI `dam token` sub-command, and a React settings UI for creating and revoking keys. Architecture documentation was updated across four pages.

### Findings
- 🟡 **Warning:** `ApiKeyValidation` interface in `packages/api-server/src/modules/api-keys/domain/types.ts:17-21` is dead code — never imported anywhere and structurally identical to `ValidatedApiKey` in the service layer. Either delete it, or promote it as the canonical domain type.
- 🟡 **Warning:** `generateKeyId()` in `packages/api-server/src/modules/api-keys/services/api-keys-service.ts:47-49` uses `crypto.randomBytes(4)` (32-bit). Newer modules use `randomUUID()` (122-bit). Switch to `randomUUID()` for consistency and safety.
- 🟡 **Warning:** `ApiKeysList` has no `error` branch — silent failure when the query errors. Add a user-visible error state.
- 🟡 **Warning:** `window.confirm` in `handleRevoke` is inaccessible and inconsistent with the app's modal patterns. Use a `ConfirmRevokeDialog` following the existing `Modal` pattern.
- 🟢 **Suggestion:** In `RevealToken`, `aria-live="polite"` on the container doesn't reliably announce icon swaps. Move the accessible label to the button: `aria-label={copied ? "Copied to clipboard" : "Copy to clipboard"}` and add `aria-hidden` to the icons.
- ✅ **Looks good:** SHA-256 for high-entropy token hashing — ADR-047 explicitly justifies this vs argon2id. Correct.
- ✅ **Looks good:** Bearer middleware prefix dispatch is clean and the fallback is safe.
- ✅ **Looks good:** `requireBrowserFlow()` guard prevents API keys from creating other API keys.
- ✅ **Looks good:** `touchLastUsed` fire-and-forget is intentional and acceptable for best-effort observability.
- ✅ **Looks good:** DB partial index on `owner_sub WHERE revoked_at IS NULL` is precisely right for the active-keys query pattern.
- ✅ **Looks good:** `auth-procedures.ts` is correctly excluded from `api-server-api/index.ts` to keep `@trpc/server` out of browser bundles.

**Verdict: aligned**

Four architecture pages updated with correct `Last verified: 2026-05-22` dates and new ADR-047 `Motivated by:` entries. No drift detected.

**[MODERATE] Dead domain type — `packages/api-server/src/modules/api-keys/domain/types.ts:17-21`**

`ApiKeyValidation` is exported but never imported anywhere in the codebase. It is structurally identical to `ValidatedApiKey` in `api-key-validator.ts`. Fix: delete `ApiKeyValidation` and import `ValidatedApiKey` in any consumer, or promote `ApiKeyValidation` as the canonical domain type.

**[MODERATE] Inconsistent key-ID generation — `packages/api-server/src/modules/api-keys/services/api-keys-service.ts:47-49`**

`crypto.randomBytes(4)` gives 32 bits; `randomUUID()` (used elsewhere in this repo) gives 122 bits and is globally unique. Switch to `randomUUID()` for consistency.

**[MODERATE] Missing error state — `packages/ui/src/modules/api-keys/components/api-keys-list.tsx`**

The component ignores the `error` field from `useApiKeys()`. When the query errors, the list disappears silently. Destructure `error` and render a user-facing message.

**[MODERATE] `window.confirm` for destructive action — `api-keys-list.tsx:14-20`**

`window.confirm` is synchronous and inaccessible. Use a small `ConfirmRevokeDialog` following the existing `Modal`/`CreateApiKeyDialog` patterns.

**[MODERATE] `aria-live` on container does not reliably announce icon swaps — `create-api-key-dialog/reveal-token.tsx:35-47`**

Move the accessible label to the button itself:

```tsx
<button
  type="button"
  onClick={handleCopy}
  aria-label={copied ? "Copied to clipboard" : "Copy to clipboard"}
  className="shrink-0 px-2 py-1 rounded hover:bg-surface text-text-secondary"
>
  {copied ? <Check size={16} aria-hidden /> : <Copy size={16} aria-hidden />}
</button>
```

### Verdict
COMMENT — Solid implementation of ADR-047 with clean architecture and good security decisions; five moderate-severity items (two TypeScript, three UI) worth addressing before merge but none are blockers.

---

packages/api-server/src/modules/api-keys/domain/token.ts
```diff
@@ -0,0 +1,38 @@
...
+  return out;
+}
+
+export function mintApiKeyToken(): { plaintext: string; hash: string } {
+  const plaintext = API_KEY_PREFIX + base32(randomBytes(RANDOM_BYTES));
+  return { plaintext, hash: hashApiKeyToken(plaintext) };
+}
+
+export function hashApiKeyToken(plaintext: string): string {
+  return createHash("sha256").update(plaintext).digest("hex");
```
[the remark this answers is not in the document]
xjacka: Thanks for flagging — the alert is a false positive in this context, and the suggested scrypt+salt changeset would break the validator. Details:

**Why SHA-256 is fine here.** `hashApiKeyToken` does not hash a user password. The input is a 256-bit cryptographically random token minted by `mintApiKeyToken` (`randomBytes(32)`, base32-encoded). ADR-047 § *Alternatives Considered* covers this explicitly: argon2id / bcrypt / scrypt protect low-entropy passwords from brute force; a 256-bit random input is already brute-force-infeasible, so the stretch buys nothing measurable. The same model is what GitHub PATs, Stripe `sk_live_*`, and AWS access keys use (`SHA-256(secret)` server-side, deterministic lookup).

**Why the suggested fix doesn't work.** The proposed `scrypt(plaintext, randomBytes(16))` produces a different digest on every call. Our validator looks the token up by digest:

```ts
const hash = hashApiKeyToken(token);
const row = await deps.findByHash(hash);   // WHERE hash = ?  (UNIQUE index, O(log n))
```

With a per-call random salt, no subsequent `WHERE hash = ?` would ever match the row inserted on `create` — every key would be one-shot. Making scrypt work would require the server to iterate every active row of the caller's owner and run scrypt against each (O(n), 50–100 ms per call), but the server has no way to know the owner before validating the token. Deterministic-lookup is structurally required for the API-key surface; it's the standard pattern for this credential class.

**What this commit does instead** (3f3b1bb4):
- Renames the internal `plaintext` identifier to `token` in `hashApiKeyToken`, `mintApiKeyToken`, the validator, and the bearer middleware so neither readers nor `js/insufficient-password-hash` mistake the input for a password. The public `ApiKeyCreateResult.plaintext` field (consumed by UI + CLI) is unchanged.
- Adds a `// lgtm[js/insufficient-password-hash]` inline suppression with a doc-comment citing ADR-047 § *Alternatives Considered*, so the rationale lives next to the code.

If CodeQL still flags after re-scan, happy to dismiss the alert as *Won't fix — false positive* with the same rationale.

packages/cli/src/modules/auth/commands/token-create.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export interface TokenCreateCommandDeps {
+  configService: ConfigService;
+  buildTrpc: (host: string) => TrpcClient;
+}
+
+export function buildTokenCreateCommand(deps: TokenCreateCommandDeps): Command {
+  return new Command("create")
+    .description(
+      "Mint a new API key. The plaintext token is printed once on stderr — copy it now, it cannot be recovered.",
```
xjacka, line 18: 🟡 **Warning:** The description says the token is printed to **stderr**, but the code always writes it to `stdout` (line 76 for `--json`, line 83 for interactive). Users who try to capture with `token=$(... 2>&1 1>/dev/null)` will get nothing.

```suggestion
      "Mint a new API key. The plaintext token is printed once to stdout — copy it now, it cannot be recovered.",
```

xjacka (commented):
🛡️ **DAM** — Code Review @ `6f41d8f`

## PR #371: feat(auth): API keys with scopes for headless CLI use (ADR-047)
**Author:** xjacka | **Branch:** feat/api-keys-headless-auth → main | **Changes:** +2089 −174 (54 files)

### Summary
This PR implements ADR-047, introducing scoped long-lived API keys for headless CLI authentication. It adds a complete stack: token minting (SHA-256 hash, `damkey_` prefix), a three-scope permission model (`agents:run`, `agents:manage`, `credentials:manage`), per-key agent allowlists, a bearer middleware that dispatches JWT vs API-key paths, a CLI `dam auth token` sub-command, and a React settings UI for creating and revoking keys. Architecture documentation was updated across four pages.

### Changes since last review
Previous HEAD: `1d227d7` (2026-05-22T12:58:28Z) — verdict COMMENT

- ✅ **Fixed:** `ApiKeyValidation` dead type removed from `domain/types.ts` — no longer present
- ✅ **Fixed:** `generateKeyId()` now uses `randomUUID()` (was `crypto.randomBytes(4)`) in `api-keys-service.ts`
- ✅ **Fixed:** `ApiKeysList` error branch added — `isError` destructured, comprehensive error UI rendered
- ✅ **Fixed:** `window.confirm` replaced with `ConfirmRevokeDialog` following `Modal`/`DialogHeader`/`DialogBody`/`DialogFooter` pattern
- ✅ **Fixed:** `aria-live` container replaced with `aria-label` on the copy button and `aria-hidden` on icons in `reveal-token.tsx`
- 🆕 **New:** `token.ts` renames `plaintext`→`token`, adds SHA-256 doc-comment, adds `// lgtm[js/insufficient-password-hash]` suppression (CodeQL false-positive — already confirmed ✅ Looks good in prior review)

### Findings
- 🟡 **Warning:** `dam auth token create` command description says the token is "printed once on **stderr**" but the code always writes the plaintext to `stdout` — to `process.stdout` in both the `--json` path and the interactive path. Users who write `token=$(dam auth token create --name ci 2>&1 1>/dev/null)` expecting to capture from stderr will get nothing. Fix: update the `.description()` string to say "stdout" instead of "stderr". (`packages/cli/src/modules/auth/commands/token-create.ts:18`)
- ✅ **Looks good:** SHA-256 for high-entropy token hashing — ADR-047 explicitly justifies this vs argon2id. The new doc-comment and `// lgtm` suppression correctly address the CodeQL false positive.
- ✅ **Looks good:** Bearer middleware prefix dispatch is clean; `verifyOwnerActive` guard correctly handles deleted-owner revocation without a sweep.
- ✅ **Looks good:** `requireBrowserFlow()` guard prevents API keys from creating/revoking other API keys.
- ✅ **Looks good:** `touchLastUsed` fire-and-forget with `.catch(() => {})` — observability, not authorization; correct.
- ✅ **Looks good:** DB partial index on `owner_sub WHERE revoked_at IS NULL` is precise for the active-keys query pattern.
- ✅ **Looks good:** All 10 module routers now use the appropriate scope procedure builder and `checkAgentBinding` guards.
- ✅ **Looks good:** `key-${randomUUID()}` for key IDs — consistent with codebase convention, 122-bit uniqueness.

**Verdict: aligned**

Four architecture pages updated with correct `Last verified: 2026-05-22` dates and new ADR-047 `Motivated by:` entries. No drift detected.

**[MODERATE] Misleading command description — `packages/cli/src/modules/auth/commands/token-create.ts:18`**

The `.description()` string says _"The plaintext token is printed once on **stderr**"_, but the code always writes the token to `stdout` — `process.stdout` on the `--json` path (line 76) and the interactive path (line 83). Metadata (id, name, scopes) goes to `stderr`; the token does not.

A user who reads the description and writes `token=$(dam auth token create --name ci 2>&1 1>/dev/null)` will capture nothing. Fix:
```ts
.description(
  "Mint a new API key. The plaintext token is printed once to stdout — copy it now, it cannot be recovered.",
)
```
Everything else — three-layer slice architecture, service factory wiring, repository pattern, `Result<T, E>` error propagation, `checkAgentBinding` guards — aligns with the architecture.

All new code in `packages/ui/src/modules/api-keys/` follows the skill rules cleanly: server state via TanStack Query hooks in `api/`, mutations use `meta: { invalidates, errorToast }`, local ephemeral state in `useState`, components sized well within bounds, `interface Props` everywhere, no `React.FC<>`, `ApiKeyRow` and `ScopeOption` correctly extracted from `.map()` blocks.

✅ No findings.

### Verdict
COMMENT — Five of five prior findings resolved; one new moderate finding (CLI `--help` description says stderr, code uses stdout) worth fixing before merge.

---

packages/api-server-api/src/modules/agents/router.ts
```diff
@@ -30,54 +35,70 @@ function toView(agent: Agent) {
...
+    .query(async ({ ctx, input }) => {
+      checkAgentBinding(ctx, input.id);
+      const agent = await ctx.agents.get(input.id);
+      if (!agent) throw new TRPCError({ code: "NOT_FOUND" });
+      return toView(agent);
+    }),
 
-  create: t.procedure
+  create: manageAgentsProcedure
     .input(agentCreateInputSchema)
```
jezekra1, line 63: A key with `agents:manage` bound to a specific agent can still hit `agents.create` and mint **new** agents outside its binding — there's no existing agentId to check, so `checkAgentBinding` is correctly absent here. But that means a bound key expands its own blast radius: an exfiltrated key bound to `agent-X` can create `agent-Y`, `agent-Z`, … under the same owner.

Per ADR-047 the whole point of binding is "operates this agent set only". `create` should require `ctx.user.agentIds === "*"` (or be moved behind a separate `agents:create` scope). Same question applies to `schedules.createCron` / `createRRule` if those don't validate the *target* agentId is in the binding — they do, but only via `checkAgentBinding(ctx, input.agentId)`, so they're fine.

packages/api-server-api/src/modules/agents/router.ts
```diff
@@ -117,9 +140,10 @@ export const agentsRouter = t.router({
       return toView(agent);
     }),
 
-  disconnectTelegram: t.procedure
+  disconnectTelegram: manageAgentsProcedure
     .input(agentDisconnectTelegramInputSchema)
     .mutation(async ({ ctx, input }) => {
+      checkAgentBinding(ctx, input.id);
```
jezekra1: The `checkAgentBinding(ctx, input.X)` line appears at the top of almost every mutation across agents, sessions, schedules, skills, secrets, connections, egress-rules. It's easy to miss when adding a new endpoint — that's exactly what happened with `egressRules.update`/`revoke` (still open), and initially with `approvals.*` and `sessions.{listByScheduleId,resetByScheduleId}` (fixed in the latest push).

Consider folding this into the procedure builders, e.g.

```ts
const agentBoundProcedure = (getAgentId: (input: any) => string) =>
  manageAgentsProcedure.use(async ({ ctx, input, next }) => {
    checkAgentBinding(ctx, getAgentId(input));
    return next();
  });
```

The variants that can't follow the pattern — `schedules.delete` loads first, `sessions.listByScheduleId` resolves through `schedule.agentId` — can still drop down to inline calls. But the common case (input has `agentId` or `id`) shouldn't repeat 30+ times across the codebase. The defense-in-depth gain: a forgotten check on a future endpoint becomes a build error, not a silent privilege escalation.

docs/adrs/047-api-keys-headless-auth.md
```diff
@@ -0,0 +1,66 @@
...
+
+The load-bearing rules:
+
+- **Three scopes, split by threat model.** An exfiltrated `agents:run` key must not be able to permanently change agent behavior. So anything that persists across runs is configuration; anything that just operates the agent in its current configuration is execution.
+  - **`agents:manage`** — agent definitions and per-agent configuration. Anything that persists across runs: Agent CRUD, Schedules, Channels, Skills install/uninstall/publish, Egress-rule writes, granting Secrets / Connections to an agent, env / model / agent-level config, Template listing.
+  - **`credentials:manage`** — global credential lifecycle: Connections (OAuth-issued, e.g. GitHub, MCP) CRUD; Secrets (user-supplied API keys, GitHub PATs) CRUD. The *grant linkage* between a credential and a specific agent lives in `agents:manage`, not here. (Server modules `connections` and `secrets` are separate; the user-facing scope unifies them — fewer scopes at create time is the right trade.)
```
jezekra1: `connections` is replacing `credentials` as the unified user-facing concept soon. Let's commit to `connections:manage` now while ADR-047 is still `Proposed` — once accepted, the scope names enter the "Committed-to" set (see the Consequences section) and renaming requires a parallel-validator window. Cheaper to fix the name before the commitment hardens than to ship two scopes that mean the same thing.

This change touches the ADR, `scopeSchema`, the procedure builder, every router that uses `manageCredentialsProcedure`, the CLI scope help text, the UI scope picker, and `ALL_SCOPES`. Mechanical but worth doing in this PR rather than as a follow-up.

packages/api-server/src/modules/api-keys/domain/token.ts
```diff
@@ -0,0 +1,46 @@
...
+import { API_KEY_PREFIX } from "api-server-api";
+
+const RANDOM_BYTES = 32;
+
+// RFC 4648 base32, lowercase, no padding. Picked over base64url because
+// shells, copy-paste tools, and `--token=` flags in screenshots survive
+// base32 round-trips without case folding hazards. 32B → 52 chars.
+const ALPHABET = "abcdefghijklmnopqrstuvwxyz234567";
+
+function base32(buf: Buffer): string {
```
jezekra1: Hand-rolled base32 encoder isn't needed — Node has this in stdlib:

```ts
import { randomBytes } from "node:crypto";
export function mintApiKeyToken() {
  const token = API_KEY_PREFIX + randomBytes(32).toString("base64url");
  return { token, hash: hashApiKeyToken(token) };
}
```

43 chars vs 52, no padding (base64url is unpadded by default in Node), no custom encoder we own forever. The case-folding rationale in the ADR ("shells, copy-paste tools, screenshots") is theoretical for modern terminals and clipboard managers, and weaker than the cost of carrying a hand-rolled encoder with its own trailing-bits edge case on line 23.

Industry norm: Stripe (`sk_…`) and GitHub (`ghp_…`) use base62, AWS Secret Access Keys use base64. base64url is the standard high-entropy bearer token shape. If there's a concrete OCR / case-fold scenario the ADR is trying to defend against, capture it in the ADR; otherwise drop this and use the stdlib.

packages/api-server/src/modules/api-keys/services/api-keys-service.ts
```diff
@@ -0,0 +1,132 @@
...
+        code: "FORBIDDEN",
+        message:
+          "API keys cannot manage API keys. Use the web UI or `dam auth login`.",
+      });
+    }
+  };
+
+  return {
+    async list() {
+      requireBrowserFlow();
```
jezekra1: Layering inconsistency: every other scope/binding check is at the procedure-builder level (`runProcedure`, `manageAgentsProcedure`, etc.), but the "API keys cannot manage API keys" rule hides inside the service via `requireBrowserFlow`. The `apiKeysRouter` uses bare `t.procedure`, so the only guardrail is the service-layer call.

Move it to the router with a dedicated builder:

```ts
export const browserOnlyProcedure = t.procedure.use(({ ctx, next }) => {
  if (ctx.user.keyId !== undefined) {
    throw new TRPCError({
      code: "FORBIDDEN",
      message: "API keys cannot manage API keys. Use the web UI or `dam auth login`.",
    });
  }
  return next();
});
```

Apply to `apiKeysRouter.{list,create,revoke}`. Then `callerKeyId` and `requireBrowserFlow` can come out of `ApiKeysServiceDeps` / `createApiKeysService` entirely — the service no longer needs to know how the caller authenticated. Tests that asserted the service-level rejection move up to a procedure-level test.

jezekra1 (changes requested):
Five blocking items. One is a correctness gap (binding-bypass on `agents.create`); the rest are consistency / future-proofing where the cost of fixing now is much lower than fixing later — `connections:manage` rename before ADR acceptance, layering the `requireBrowserFlow` check at the router, stdlib base64url instead of hand-rolled base32, and folding the per-handler `checkAgentBinding` calls into the procedure builders (which would have caught the egress-rules / approvals / sessions-by-schedule misses for free).

xjacka:
Thanks for the deep review @jezekra1 — applied all four code-touching items in 1a93f36b. Quick walkthrough:

**Scope rename `credentials:manage` → `connections:manage`** — done. Touched `scopeSchema`, `ALL_SCOPES`, the procedure builder, the two routers using it, the CLI scope help text, the UI scope picker, and ADR-047. Agreed it was cheaper to rename now while the ADR is still `Proposed`.

**`browserOnlyProcedure`** — done. The "API keys cannot manage API keys" rule lives at the router now (`apiKeysRouter.{list,create,revoke}` use the new procedure), and `callerKeyId` / `requireBrowserFlow` came out of `ApiKeysServiceDeps` and `createApiKeysService` entirely. Service-level rejection tests moved to the procedure layer (new `auth-procedures.test.ts`). Required a subpath export `api-server-api/auth-procedures` so the test can reach it without going through the barrel (we keep `@trpc/server` out of browser bundles).

**`agents.create` wildcard-only** — done. `rejectIfRestricted(ctx)` runs before the service call; restricted keys get `FORBIDDEN: Agent creation requires a wildcard-bound key …`. Wildcard-bound keys and browser sessions unaffected. The other `manageAgentsProcedure` paths (update/delete/restart/wake/connect*) already pass `checkAgentBinding(ctx, input.id)`, so they're fine.

**base32 → base64url** — done. `randomBytes(32).toString("base64url")` replaces the hand-rolled encoder; tokens are 43 chars now, no padding, no trailing-bits edge case to maintain. ADR-047 + test updated.

**`checkAgentBinding` folding** — agreed in principle, deferred. tRPC middleware sees `input` as `unknown` pre-parse, so a generic `agentBoundProcedure(getAgentId)` factory either weakens the typecheck (cast inside the extractor) or forces every input schema to compose a shared `agentIdSchema` — the latter is the cleaner path but touches every router and is a separate refactor. I'll open a follow-up that pairs it with an ESLint rule (`no-agent-id-without-binding-check`) so a forgotten check on a future endpoint becomes a build error rather than a silent privilege escalation.

docs/adrs/058-api-keys-headless-auth.md
```diff
@@ -0,0 +1,67 @@
+# ADR-057: API keys with scopes for headless CLI use
```
xjacka: 🟡 **Warning:** Filename / index say `058`, this title says `057`. ADR-057 is `structured-logging.md` — collision in human-facing label.

```suggestion
# ADR-058: API keys with scopes for headless CLI use
```

docs/adrs/058-api-keys-headless-auth.md
```diff
@@ -0,0 +1,67 @@
...
+
+- Keycloak access tokens are short-lived; there is no in-CLI refresh path when `DAM_TOKEN` is set.
+- They carry the user's full permissions; there is no way to grant narrower access (e.g. "only run this one agent").
+- Revoking a token without invalidating the user's whole SSO session is not a first-class Keycloak workflow.
+
+CI pipelines and orchestration layers need a long-lived, server-managed, scopable credential type. This ADR decides what that credential is and how it sits next to the existing Keycloak login path. Sub-issue [#249](https://github.com/dam-agents/dam/issues/249) frames the design space; this ADR resolves the open questions and pins the load-bearing rules.
+
+## Decision
+
+**Introduce API Keys as a new owner-scoped credential type, with three permission scopes, and let the existing `Authorization: Bearer` slot carry either credential type — discriminated by a `pk_` prefix.** Plaintext is returned once on create and never persisted; everything at rest is a SHA-256 digest. The CLI's `DAM_TOKEN` env var continues to accept either credential type — no CLI branching.
```
xjacka: 🟡 **Warning:** Decision paragraph says "SHA-256 digest" but line 34 (and the code in `domain/token.ts`) say HMAC-SHA256 with a server pepper. Worth updating here so the ADR is internally consistent.

docs/architecture/cli.md
```diff
@@ -9,6 +9,7 @@ Last verified: 2026-06-01
 - [#73 — Import local project context into agent workspace](https://github.com/dam-agents/dam/issues/73) — the `dam import` verb that uploads local files and folders into an Agent.
 - [#254 — Granular file ops over the agent-runtime proxy](https://github.com/dam-agents/dam/issues/254) — the `dam file` group (`get`, `put`, `list`) for single-file workspace operations.
 - [ADR-046 — Eliminate Instance, collapse into Agent](../adrs/046-eliminate-instance.md) — the CLI addresses Agents (not Instances); a single `dam agent` group covers the lifecycle.
+- [ADR-057 — API keys with scopes for headless CLI use](../adrs/058-api-keys-headless-auth.md) — the `dam auth token` sub-tree; `DAM_TOKEN` now accepts API keys (`pk_…` prefix) in the same Bearer slot the JWT flow already uses.
```
xjacka: 🟡 **Warning:** Label says `ADR-057`, link target is the `058-…md` file. Same mislabel appears 8× across architecture pages (cli.md ×3, persistence.md ×2, platform-topology.md ×2, security-and-credentials.md ×2). Change the visible label text from `ADR-057` to `ADR-058` everywhere.

xjacka (commented):
🛡️ **DAM** — Code Review @ `cd15a66`

## PR #371: feat(auth): API keys with scopes for headless CLI use (ADR-058)
**Author:** xjacka | **Branch:** feat/api-keys-headless-auth → main | **Changes:** +2481 −157 (77 files)

### Summary
Re-review at HEAD `cd15a66` (previously reviewed at `6f41d8f` on 2026-05-22). Introduces long-lived owner-scoped API keys as a second credential type sharing the `Authorization: Bearer` slot with Keycloak JWTs, distinguished by a `pk_` prefix. Three permission scopes (`agents:run`, `agents:manage`, `connections:manage`), per-key agent allowlists, HMAC-SHA256 keyed digest at rest with server pepper, new `dam auth token` CLI surface, Settings → API Keys UI tab. Closes #249.

### Changes since last review
Previous HEAD: `6f41d8f` (2026-05-22T15:45:50Z) — verdict COMMENT

- ✅ **Fixed:** Prior `dam auth token create` description vs behavior mismatch resolved. The description now correctly states "Default: token printed on stderr (so `> file` redirects don't silently capture it). With --json: token printed on stdout inside the JSON payload" and the code matches.
- 🆕 **Resolved (jezekra1's CHANGES_REQUESTED, all 5 items):**
  - Scope rename `credentials:manage` → `connections:manage` across schema, types, procedure builder, routers, ADR, CLI help, UI scope picker.
  - `browserOnlyProcedure` extracted to the contract layer (`packages/api-server-api/src/auth-procedures.ts:53`); `requireBrowserFlow` and `callerKeyId` removed from `ApiKeysServiceDeps`. Layering fixed.
  - `manageAgentCreateProcedure` (`auth-procedures.ts:126`) rejects non-wildcard keys for `agents.create` — closes the binding-bypass.
  - Token codec uses `randomBytes(32).toString("base64url")`; hand-rolled base32 encoder gone.
  - **Deferred item from xjacka's reply landed in `f68dd3c`:** `checkAgentBinding` folded into procedure builders. New `manageAgentByAgentIdProcedure` / `readAgentByAgentIdProcedure` / `runAgentByAgentIdProcedure` factories with `agentIdInput` shared schema; "a forgotten check on a future endpoint becomes a type error (missing field in the intersected input), not a silent privilege escalation". Exactly what was asked for.
- 🆕 **Security strengthening (`f1167d7`):** Token digest moved from plain SHA-256 to HMAC-SHA256 with a server-side pepper (`API_KEY_HMAC_KEY`, generated and persisted by the Helm chart, mirroring `ACTIVITY_HMAC_KEY`). Codec refuses to start without the env var. Addresses the CodeQL "insufficient password hash" finding from `github-advanced-security` on 2026-06-01.
- 🆕 **Prefix rename:** `damkey_` → `pk_` (visible across types, ADR, architecture pages, CLI; PR description still says `damkey_` — minor cosmetic).
- 🆕 **ADR renumber:** 047 → 058 (047 was taken by terms-of-use). All link URLs updated to point at `058-api-keys-headless-auth.md`. Labels and the ADR's own H1 not fully aligned — see findings below.

### Findings
- 🟡 **Warning:** **ADR file H1 doesn't match the filename.** `docs/adrs/058-api-keys-headless-auth.md:1` starts with `# ADR-057: API keys with scopes for headless CLI use` — the file is filed as 058 (and the index correctly lists it as `[058](058-api-keys-headless-auth.md)`), but the document's own title says 057. ADR-057 in the same directory is `structured-logging.md`. Anyone landing on this ADR via search or direct link will see two different numbers for the same artefact. Fix: change the H1 to `# ADR-058`.
- 🟡 **Warning:** **Same mislabel in four architecture pages.** Every reference reads `[ADR-057 — API keys with scopes for headless CLI use](../adrs/058-api-keys-headless-auth.md)` — label says 057, target file is 058. Total 8 occurrences across `cli.md` (3), `persistence.md` (2), `platform-topology.md` (2), `security-and-credentials.md` (2). Same fix: change the label text from `ADR-057` to `ADR-058`.
- 🟡 **Warning:** **ADR-058 has an internal inconsistency about the digest algorithm.** Line 19 (the **Decision** paragraph) says "everything at rest is a SHA-256 digest". Line 34 (Token format and at-rest digest) correctly says "HMAC-SHA256 keyed with a stable server-side pepper". The latter matches the code; the former is old text from before the HMAC-pepper change. Fix: update the Decision paragraph to say "HMAC-SHA256 with a server pepper" so the ADR is consistent end-to-end.
- 🟢 **Suggestion:** PR description still references `damkey_` prefix and "SHA-256 at rest" — both outdated. Worth a one-line edit to the PR body to read `pk_` and "HMAC-SHA256 with server pepper" so the merge commit's autogenerated description doesn't immortalize the older proposal.
- ✅ **Looks good:** `manageAgentByAgentIdProcedure` family is a really nice piece of API design — the chained `.input(agentIdInput)` before `.use(...)` means a router attaching its own `.input(...)` gets the intersected type at compile time, so a missed binding check becomes a build error. Mechanically prevents the very mistake jezekra1 was worried about.
- ✅ **Looks good:** `createApiKeyTokenCodec` refuses to construct without the pepper key ("refusing to hash tokens without a server-side pepper"). Fail-closed defaults, no silent downgrade to peppered-by-empty-string. Right call.
- ✅ **Looks good:** HMAC-SHA256 with server pepper is the right choice for high-entropy random secrets — a slow KDF (argon2id/bcrypt) buys nothing for 256-bit random tokens and would tax every request. The ADR (and the inline comment in `token.ts`) explain this trade-off carefully.
- ✅ **Looks good:** `manageAgentCreateProcedure` is a clean expression of "the binding model is 'this set of agents and no others'; `create` has no agentId to match" — the error message names the constraint explicitly.
- ✅ **Looks good:** Helm chart wires the pepper as a chart-generated secret next to `ACTIVITY_HMAC_KEY` — same pattern, same operator UX, survives upgrades.

3 architecture-doc cross-reference issues — all variants of the same root cause (ADR renumber 047 → 058 left a label / H1 trail):
- ADR `058-api-keys-headless-auth.md` H1 says "ADR-057" (filename / index disagree).
- 8 architecture-page link labels read `ADR-057` while the URLs target `058-…md`.
- ADR Decision paragraph says "SHA-256 digest" but the rest of the ADR (and the code) says "HMAC-SHA256 with pepper".

✅ No findings.

✅ No findings.

### Verdict
COMMENT — Substantively the right code: every jezekra1 finding addressed, the deferred follow-up landed with a type-safe abstraction, CodeQL warning closed by moving to HMAC+pepper. Three small documentation-numbering bugs (ADR label trail from the 047 → 058 renumber, plus the SHA-256 vs HMAC-SHA256 inconsistency inside the ADR itself) are worth fixing before merge — they're the kind of thing that compounds if it survives into the merged commit. No blockers from the code itself.

---

tomkis:
Let's make `verifyApiKey` and `verifyOwnerActive` required. The only thing keeping them optional is the one unit test that builds `createAuth(config)` with no deps, and we shouldn't make production code optional just to suit a test. It hurts clarity. Pass stubs in the test instead. Making them required also drops the `= {}` default and the two absence branches, so it's a net simplification.

tomkis (changes requested):
I don't think we should merge this with the scope model in. The keys themselves are fine (owner-scoped, agent-bound, revocable, long-lived). The problem is the three scopes.

`agents:run` / `agents:manage` / `connections:manage` look like a security boundary, but they aren't one, and shipping them gives people a false picture of how locked-down their token is. The split assumes "run" can only operate the agent and can't change anything that sticks around. That isn't true. A run token lets you drive the agent, and the agent has its own management powers through the internal MCP server. `create_schedule` writes a persistent platform schedule, `install_skill` changes what the agent does on every future run, and `send_channel_message` can attach any file from the pod and post it out. All of that runs with the owner's full authority and no scope check. So "run" quietly gets you "manage" on that agent. We'd be handing someone a token we describe as narrow that can set up persistent jobs and exfiltrate files.

My proposal: drop the run/manage/connections distinction for now. Yes, that downscopes the feature. A key just carries the owner's authority, still owner-scoped, still agent-bound, still revocable. That's honest about what the token can actually do, and it's a boundary that holds.

Real scoping is worth doing, but doing it properly is a big piece of work, not a tweak to this PR. It means threading the calling principal all the way into the agent so the MCP tool calls can be checked against the token's scopes, and even then unattended and scheduled runs have no principal to check against. That's its own feature. Let's not ship a fake version of it now and call it a boundary.

jezekra1 (approved):

packages/api-server-api/src/modules/api-keys/schemas.ts
```diff
@@ -0,0 +1,56 @@
...
+  "agents:manage",
+] as const;
+
+/**
+ * Credential-domain scopes covering both OAuth connections and user-supplied
+ * secrets. `credentials:manage` implies `credentials:read`.
+ */
+export const CREDENTIAL_SCOPES = [
+  "credentials:read",
+  "credentials:manage",
```
xjacka, line 28: 🟡 **Warning:** This still ships `credentials:manage`, but the PR description, @xjacka's "applied" comment, and @jezekra1's request all say the scope was renamed to `connections:manage`. There are **zero** `connections:manage` occurrences at this HEAD — the rename looks like it was dropped when the branch was rebased/recreated today. Re-apply before merge so the shipped scope matches the ADR/UI.

packages/api-server/src/apps/api-server/auth.ts
```diff
@@ -41,6 +54,20 @@ export interface AuthConfig {
...
+   *  API-key tokens are rejected so deployments without the api-keys module
+   *  wired in remain JWT-only. */
+  verifyApiKey?: (
+    token: string,
+  ) => Promise<Result<ValidatedApiKey, ApiKeyValidationFailure>>;
+  /** Per-request owner-still-active check for API-key principals. Returns
+   *  false if the owner no longer exists in Keycloak; the key is then
+   *  treated as revoked for the request. JWT principals don't need this
+   *  (Keycloak signs each access token). */
+  verifyOwnerActive?: (sub: string) => Promise<boolean>;
```
xjacka, line 68: 🟡 **Warning (@tomkis's open CHANGES_REQUESTED):** keeping `verifyOwnerActive` optional means a composition that forgets to wire it silently skips the per-request owner-active revocation check (fail-open). Production currently wires it (`app.ts:226`), but making it required — as requested — converts a forgotten wiring into a startup error instead of a silent auth bypass. Same applies to `verifyApiKey` and the `deps = {}` default.

xjacka (commented):
🛡️ **DAM** — ⚠️ Code Review @ `55a0917`

## PR #371: feat(auth): API keys for headless CLI use (ADR-064)
**Author:** xjacka | **Branch:** feat/api-keys-headless-auth → main | **Changes:** +4089 −148 (65 files)

### Summary
Adds long-lived, owner-scoped API keys sharing the `Authorization: Bearer` slot with Keycloak JWTs, distinguished by a `pk_` prefix: a token codec (256-bit random + HMAC-SHA256 with a required server pepper), scoped tRPC procedure builders, a prefix-dispatching bearer middleware, an `api_keys` table, a `dam auth token` CLI group, and a Settings → API Keys UI. Heavily reviewed already (jezekra1 deep-reviewed and APPROVED; CodeQL weak-hash alert rebutted).

### Findings
- ✅ **Looks good (crypto core):** The token codec is sound — 256-bit `randomBytes` + `base64url`, HMAC-SHA256 keyed by a **required** server pepper (refuses to start without it). The CodeQL "insufficient computational effort" alert is correctly a **false positive**: slow KDFs matter only for low-entropy passwords, not 256-bit random secrets; and a constant-time compare is unnecessary because the lookup key is a digest of a high-entropy secret. Agreed with @xjacka's rebuttal.
- ✅ **Looks good (validation):** `createApiKeyValidator` fails closed on unknown/revoked/expired, treats `touchLastUsed` as best-effort (caught), and the bearer path fails closed when API keys aren't enabled. `browserOnlyProcedure` blocks key-authed `apiKeys.*` so an exfiltrated key can't mint more keys.
- ✅ **Looks good (review uptake):** Nearly all prior feedback landed at this HEAD — keyId → `randomUUID()`, plaintext token → **stderr** (JSON payload to stdout), UI `isError` branch + `ConfirmRevokeDialog`, `agents.create` wildcard-only, `base64url`, dead `ApiKeyValidation` removed.
- 🟡 **Warning (agreed rename missing at HEAD):** The scope is still **`credentials:manage`** at this commit (`schemas.ts:28` + 8 more; **zero** `connections:manage`), yet the PR description, @xjacka's "applied" comment, and @jezekra1's request all say it was renamed to **`connections:manage`**. The rename appears to have been dropped when the branch was rebased/recreated today (HEAD `55a0917`, 2026-06-16). Re-apply it before merge so the shipped scope matches the ADR/UI and the state that was approved.
- 🟡 **Warning (open CHANGES_REQUESTED — @tomkis):** `verifyApiKey`/`verifyOwnerActive` are still optional in `AuthDeps` with a `deps = {}` default and absence branches (`auth.ts:61,68,91,136`). Production wires both (`app.ts:225–226`), so no live vuln — but the optional `verifyOwnerActive` is fail-open: a future composition that forgets it silently disables the per-request owner-active revocation check. Making them required (as @tomkis asked) turns that into a startup error.

**Verdict: minor drift**
- ⚠️ **`Last verified` not bumped (check 2):** this commit (2026-06-16) substantively edits four architecture pages (e.g. the new API-key bearer paragraph in `security-and-credentials.md`), but their `Last verified` headers were left at `2026-06-12` (cli, platform-topology) / `2026-06-15` (persistence, security-and-credentials). Bump the edited pages to the merge date.
- ✅ The earlier ADR-label leak across these pages is resolved — they now carry **zero** ADR references, and the new API-key behavior is documented in `security-and-credentials.md` + `persistence.md`.

### Verdict
COMMENT — Security design is solid and the crypto core is correct (CodeQL alert is a false positive). Two non-blocking items to resolve before merge: the agreed `credentials:manage`→`connections:manage` rename is missing at this HEAD (likely a lost rebase commit), and @tomkis's request to make the auth deps required is still open. Plus a minor doc-drift `Last verified` bump.

---

tomkis (approved):

PR #350, opened 2026-05-21
docs(adrs): propose unified Connection model + runtime channel + outbox

jezekra1:
## Summary

Three proposed ADRs that together unify how external integrations reach an agent. Output of a grilling/design session — no code changes yet; subsequent PRs implement.

- **ADR-047** — Connections, Connection Templates, Contributions. Collapses today's parallel `OAuthAppDescriptor` + `ProviderPreset` registries into one Connection model. Every integration is a `Connection` built from a `Connection Template` and emits a typed `Contribution[]` when granted to an Agent. Kinds: `env`, `egress-host`, `file`, `mcp-entry`, `skill-ref`. Two-axis Template classification: `category` (`app`/`mcp`/`other`) × `isCustom`. Auth modes are orthogonal: `oauth`/`api-key`/`header`/`none`.
- **ADR-048** — Unified runtime channel between api-server and agent-runtime. tRPC both directions (no persistent WebSocket). Snapshot semantics for declarative state, signal events for transient actions (TTL + at-least-once redelivery). Capability negotiation on `hello`. Versioning via route prefix (`runtime.v1.*` → `runtime.v2.*` only on semantic breaks; capability flags for additions). Manifest-driven per-harness driver model. **Supersedes ADR-008 and ADR-034.**
- **ADR-049** — Transactional outbox + worker for runtime-channel delivery. Postgres outbox in the mutation transaction; competing-consumers worker via `FOR UPDATE SKIP LOCKED`; Redis pub/sub as low-latency wakeup per ADR-036; periodic sweep as safety net. Skips non-running agents (uses existing agents-service ConfigMap-watch cache). Detachable design — worker is a self-contained module that can lift into its own Deployment via feature flag.

Glossary additions for `Connection`, `Connection Template`, `Contribution`, `AuthConfig`, `State Event`, `Signal Event`, `Last Applied Hash` in `docs/ubiquitous-language.md`.

## What the three together replace

| Today | Replaced by |
|---|---|
| Pod-files SSE (ADR-034) | Snapshot delivery on the runtime channel |
| `~/.triggers/` file drops via `kubectl exec` (ADR-008) | Signal events with durable outbox + TTL redelivery |
| Direct skills tRPC from api-server | `skill-ref` contributions in the snapshot; existing skill-install helper becomes the driver's internal impl |
| Hardcoded `platform-outbound` MCP boot-write | Platform-system Connection Template emitting `mcp-entry` to every agent |
| OAuth app + provider preset registries duplication | One Connection Template registry |

[...]

PR #381, opened 2026-05-25
feat: unified Connections + runtime channel + OAuth (ADR-051/052/053)

jezekra1:
## Summary

Full implementation of the unified Connections / Contributions model + the runtime channel between api-server and agent-runtime, plus the new OAuth flow on top of SecretStore. The companion docs PR is #350.

Eleven commits across:

### Foundation
- **DB schema** (commit 1): `connections` + `connection_grants`, `agents` (runtime-channel view), `runtime_state_outbox`, `runtime_events`, `trigger_dispatches`. Migration `0014_runtime_channel.sql`.
- **Wire contracts** (commit 2): Contribution / Event / applyState / hello Zod schemas (`agent-runtime-api`). Harness tRPC router + HarnessContext (`api-server-api`).

### Agent runtime
- **`runtime-channel/` module** (commit 3): manifest loader, state store, dispatcher, `file` impl (4 merge modes), `skill-install` impl, harness tRPC client, event loop, applyState handler, helloOnBoot.

### Runtime delivery (api-server)
- **Outbox + worker + sweep + hello + trigger handler** (commits 4–5): BullMQ queue with stable per-agent jobId, cron sweep for stale rows / expired events, state-builder shared by worker and hello, hello upserts agents-runtime row, trigger handler idempotent on `trigger_dispatches.event_id`. Composed at api-server boot; harness tRPC mounted under `/api/agents/:id/trpc/*`.
- **Schedule firing** (commit 6): controller's Go scheduler POSTs `/api/agents/:id/internal/schedule-fired`; api-server inserts the `runtime_events` row. Base `runtime-manifest.yaml` ships from `platform-base`.

### Cross-cutting secret storage
- **SecretStore port + K8s adapter** (commit 7): store-agnostic `SecretRef` in `api-server-api`. `SecretStore` interface (`mintRef` / `put` / `putField` / `get` / `getField` / `delete` / `list` with `purpose` filter) + `SecretStoreRegistry`. K8s adapter encapsulates label scheme + path naming. Vault / AWS SM adapters slot in alongside without call-site changes.

### Connections model
- **Templates + service + fan-out** (commit 8): ConnectionTemplate registry (GitHub, Custom MCP, Custom Header). Postgres-backed ConnectionsService (template-driven create, list, get, delete, agent grants). Three-rail contribution fan-out: env → `secrets-rev` bump (ADR-040 preserved), egress-host → `egress_rules` sync, file/mcp/skill → outbox + version bump.

### Retirement
- **Pod-files SSE / trigger-watcher / legacy POST /internal/trigger** (commit 9): deleted entirely.

### UI
- **Connection Templates section + create form** (commit 10): top of the Connections view. Per-template input form, status pill, delete; auto-redirects to OAuth for templates with oauth auth.

### OAuth roundtrip on SecretStore (commits 11–12)
- **PKCE engine** with `start` / `peek` / `consume` / `exchange` / `refresh`. Provider-agnostic; threading typed `ctx`.
- **OAuthFlowService** orchestrates engine + SecretStore + Connections repo. `startOAuth(connectionId)` returns `authUrl`. `completeOAuth(state, code)` writes tokens via `SecretStore.putField` and stamps `auth.expiresAt`.
- **Refresh loop**: periodic sweep across OAuth Connections with tokens within `refreshSkewSeconds` of expiry; rotates access (and refresh, when provider rotates).
- **`connections.startOAuth` tRPC mutation** + single `/api/oauth/callback` route.
- **GitHub template** carries the full provider config + `oauthExtras()` returning the operator-supplied `clientSecret`.
- **Legacy deleted**: `oauth-engine` (old shape), `oauth-apps`, `k8s-connections-port`, `mcp-discovery`, `oauth-refresh-service` (+ 4 test files); legacy UI surface (`OAuthAppRow`, `McpConnectionRow`, `OAuthAppConnectButton`, `OAuthAppIcon`, `ConnectAppForm`, `AddMcpForm`, `install-url.ts`); related queries / mutations / type aliases; admin-default GHE / AppSlug config keys.
- **UI wired**: `useStartOAuth` mutation. `TemplateCreateForm` for OAuth templates submits + auto-redirects ("Create + Authorize"). Per-row "Connect" button for `pending` OAuth Connections.

## Test plan

- [x] `mise run check` (lint + format + tsc across all packages + Go controller) — passing locally on every commit.
- [x] `mise run test` — 144 + 42 vitest tests + Go controller tests passing locally.
- [ ] Manual: configure `PLATFORM_DEFAULT_GITHUB_CLIENT_ID` + `PLATFORM_DEFAULT_GITHUB_CLIENT_SECRET`, click "Add GitHub" in UI → lands on github.com/login/oauth/authorize → grants → returns to `?oauth=success` → Connection row shows status `active` → grant to an agent → agent pod re-renders with the token from SecretStore-mounted Secret.
- [ ] Manual: let an access token expire; verify the refresh loop kicks in and `auth.expiresAt` advances without user action.
- [ ] Manual: schedule firing fires through `runtime_events` → `applyState` → `events.trigger`; verify idempotency on a duplicate `event_id`.
- [ ] Manual: `applyState` on an offline agent fails cleanly, cron sweep re-enqueues, agent `hello` picks up the snapshot on wake.

## Known transitional pieces (worth a look)

- **Type-identity casts in UI** (`as unknown as ConnectionView` in two places). tRPC client's response inference mints a structurally-equal but nominally-distinct ConnectionView; harmless. Will go away when the file Contribution's `content: unknown` schema is reshaped.
- **`peek`-then-consume in OAuth callback**: the callback handler peeks (non-mutating) to recover the owner sub, then the per-owner `OAuthFlowService` does the real consume + exchange. Cleaner than a global service.
- **Worker `isRunning` stub** returns `true`. Real impl reads from the ConfigMap-watch cache; current behavior degrades to BullMQ retries + cron sweep, which is correct just suboptimal.
- **Connection Template catalog is GitHub + Custom MCP + Custom Header only**. Anthropic / OpenAI / Spotify / Google services etc. are now per-template additions — each is a single Template entry in code.

Closes #379.


tomkis:
`packages/api-server-api/src/modules/connections/types.ts` has a bunch of Zod schemas in it that I think belong in `schemas.ts`. Every other module in this package keeps them split: schemas live in `schemas.ts`, and `types.ts` only does `z.infer` (see `schedules/types.ts` or `agents/types.ts`).

tomkis:
`fileContribution` lets you build combinations that the file-ops code can't actually handle. `mergeMode` and `format` are flat enums, so the schema accepts e.g. `format: "text"` with `mergeMode: "key-targeted"` (which parses the existing file as a string then spreads it into `{"0":"[", "1":"f", ...}`), or `format: "json"` with `mergeMode: "yaml-fill-if-missing"` (which silently rewrites the file as YAML, the merger never reads `format`). Same story for `json + section-marker` (emits `//` comment markers in JSON) and `ini + key-targeted`.

I'd discriminate on `mergeMode`. The implementation already switches on it, and each branch wants different fields: `key-targeted` needs a `keyPath` that isn't even in the current wire schema, `yaml-fill-if-missing` doesn't need `format` at all, and the format-set varies per mode. Something like:

```ts
{ kind: "file", path, mergeMode: "overwrite", format: fileFormat, content }
{ kind: "file", path, mergeMode: "key-targeted", format: z.enum(["yaml","json"]), keyPath, content: z.record(...) }
{ kind: "file", path, mergeMode: "section-marker", format: z.enum(["yaml","ini","text"]), content }
{ kind: "file", path, mergeMode: "yaml-fill-if-missing", content: z.record(...) }
```

tomkis:
The ubiquitous-language doc hasn't caught up with this PR. Two things:

1. The Connections section is still labeled "proposed, in-flight design" but this PR ships it. The framing needs to come off.

2. There's no vocab for the runtime channel at all. Plugin, Driver, Driver Binding, Runtime Manifest, KindHandler, Plugin Registry, Extension, Plugin Protocol Version, applyState, hello, Capability filter. None of it is in the doc, even though it's the core of ADR-052 and a chunk of this PR.

   Probably wants its own bounded context section ("Runtime Channel") rather than getting tacked onto Connections.

While you're in there: driver vs plugin is the same concept under two names (manifest section is `drivers:`, the port is `Plugin`, the folder is `drivers/` but the files are `*-plugin.ts`). Worth picking one before the vocab solidifies.

Happy to take this in a follow-up if you'd rather keep this PR focused.

tomkis:
Two ways to set up an LLM now (Connection + Provider), which is confusing. And the surfaces aren't equivalent: Providers supports OAuth tokens for Anthropic (sk-ant-oat-), Connections is API-key only. I'd rip out Providers, port the OAuth-token mode, let UX figure out the rest.

tomkis (approved):

tomkis:
Schedules seem to work but conversatiions are not visible in UI because we don't create record in sessions table -> only ACP session is created.

PR #390, opened 2026-05-28
feat: allow sending images through slack

### LINKED ISSUE #306: Slack integration ignores image attachments
## Problem

When a user mentions the agent in Slack with a screenshot or other image attached, the agent only receives the text of the message — the image is silently dropped. The user has no signal that the attachment was ignored, so they may believe the agent can see it.

This forces users into awkward workarounds: describe the screenshot in words, paste error text manually, or move the conversation to another channel that supports images. It also makes Slack noticeably less capable than other places the same agent runs, despite Claude itself being multimodal.

## Proposed solution

Image attachments on incoming Slack messages should be forwarded to the agent along with the text, so the agent can actually look at them and respond to what it sees.

### Scope

- Images attached directly to mentions / DMs / thread replies the agent already responds to.
- Behavior should match what users already expect from the agent on other channels.

### Out of scope

- Non-image file types (PDFs, code files, etc.) — can be a follow-up if there's demand.
- Outbound image handling (the agent posting images back) — already supported.

jjeliga:
Extending the api-server functionality such that sending images through slack is now possible for agents that support such functionality through ACP.
The only guard on the api-server side is the agents capability, otherwise all images are relayed unconditionally and errors are reported to the user.
The necessary adjustments on the slack bot side are documented.

Closes #306 

Note:
Not tested for any other agent than CC as I do not have access

image

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -40,7 +40,7 @@ Users can also author skills in the Files panel and publish them upstream as pul
 
 Platform runs a single Slack app (Socket Mode) for the entire installation. Multiple instances can share a channel — the bot routes messages per thread.
 
-1. [Create a Slack app](https://api.slack.com/apps) with Socket Mode enabled and bot/user token scopes: `app_mentions:read`, `channels:history`, `chat:write`, `files:write`, `reactions:write`, `commands`, `users:read`. (`files:write` powers outbound file attachments via the `send_channel_message` MCP tool — existing installations must add it and have admins re-approve.)
+1. [Create a Slack app](https://api.slack.com/apps) with Socket Mode enabled and bot/user token scopes: `app_mentions:read`, `channels:history`, `chat:write`, `files:read`, `files:write`, `reactions:write`, `commands`, `users:read`. (`files:write` powers outbound file attachments via the `send_channel_message` MCP tool; `files:read` lets the worker download user-attached images so the agent can see them. Existing installations must add both and have admins re-approve.)
 2. Add slash command `/platform` pointing to your app.
 3. Generate an app-level token (`xapp-...`) with `connections:write` scope. Deploy with both tokens:
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -40,7 +40,7 @@ Users can also author skills in the Files panel and publish them upstream as pul
 
 Platform runs a single Slack app (Socket Mode) for the entire installation. Multiple instances can share a channel — the bot routes messages per thread.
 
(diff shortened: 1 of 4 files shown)


docs/configuration.md
```diff
@@ -40,7 +40,7 @@ Users can also author skills in the Files panel and publish them upstream as pul
 
 Platform runs a single Slack app (Socket Mode) for the entire installation. Multiple instances can share a channel — the bot routes messages per thread.
 
-1. [Create a Slack app](https://api.slack.com/apps) with Socket Mode enabled and bot/user token scopes: `app_mentions:read`, `channels:history`, `chat:write`, `files:write`, `reactions:write`, `commands`, `users:read`. (`files:write` powers outbound file attachments via the `send_channel_message` MCP tool — existing installations must add it and have admins re-approve.)
+1. [Create a Slack app](https://api.slack.com/apps) with Socket Mode enabled and bot/user token scopes: `app_mentions:read`, `channels:history`, `chat:write`, `files:read`, `files:write`, `reactions:write`, `commands`, `users:read`. (`files:write` powers outbound file attachments via the `send_channel_message` MCP tool; `files:read` lets the worker download user-attached images so the agent can see them. Existing installations must add both and have admins re-approve.)
```
tomkis: Can we change this to only enumeration of needed scopes? The extra information is just not durable documentation.

packages/api-server/src/modules/channels/infrastructure/slack.ts
```diff
@@ -35,6 +36,54 @@ type BoltApp = InstanceType<typeof App>;
...
+  const images: FetchedImage[] = [];
+  const failures: FetchedFailure[] = [];
+  for (const f of files ?? []) {
+    if (!f.mimetype?.startsWith("image/")) continue;
+    try {
+      const res = await fetch(f.url_private, {
+        headers: { Authorization: `Bearer ${botToken}` },
+      });
+      if (!res.ok) throw new Error(`HTTP ${res.status}`);
+      const data = Buffer.from(await res.arrayBuffer()).toString("base64");
```
tomkis: Can't we go with piping here? This can have immense memory requirements.

packages/api-server/src/modules/channels/infrastructure/slack.ts
```diff
@@ -119,6 +168,15 @@ export function createSlackWorker(
 ): SlackWorker {
   let app: BoltApp | null = null;
 
+  async function ephemeral(channel: string, user: string, text: string) {
+    if (!app) return;
```
tomkis: Why the app would not exist? Even if there is a good reason i would at elast console.warn

tomkis (changes requested):

tomkis (approved):

Issue #393, opened 2026-05-28
Add a generic OpenAI-compatible provider


Tomas2D:
## Problem

The platform currently ships a hard-coded provider preset for each backend it supports — Anthropic, OpenAI, IBM LiteLLM, Bob, with RITS in flight. Each one has a fixed host, fixed auth shape, and fixed env-mapping logic baked into the platform.

That works for the handful of providers we curate, but it leaves a large class of backends unreachable today: anything that speaks the OpenAI API shape but isn't on one of those pinned hosts — self-hosted inference (vLLM, Ollama, LM Studio, TGI), corporate OpenAI-compatible proxies, regional gateways, or any new vendor whose API mimics OpenAI. To use one, a user has to fall back to wiring a generic secret plus a handful of pod env variables by hand, which is the exact undiscoverable, error-prone path the provider presets were created to replace.

## Proposed solution

Add **OpenAI-compatible** as a first-class provider type alongside the curated ones.

- User picks "OpenAI-compatible", enters a base URL, an API key, and a model (or picks one from the endpoint's model list).
- Platform handles the rest: auth header, env wiring, validation against the endpoint.
- Works for both supported agent harnesses — Pi Agent and Claude Code.
- Connecting to any OpenAI-API-shaped backend becomes a single provider card, not a custom secret recipe.

### Scope

Covers backends that conform to the OpenAI API surface the harnesses already speak (chat completions, models listing for discovery). Endpoints that diverge meaningfully from that shape stay out — they're candidates for their own curated preset.

### Out of scope

- Replacing the existing curated presets (Anthropic, IBM LiteLLM, etc.). They stay as-is for the polished one-click experience; this is the escape hatch for everything else.
- Anthropic-API-shaped backends — same idea, but a separate ticket if/when we need it.

### Dependencies

Related: #161 (RITS as a curated preset), #388 (IBM LiteLLM updates). A generic OpenAI-compatible provider doesn't replace those, but it does reduce pressure to curate every new backend.

HerShin5:
This would remove a lot of brittle per-provider workarounds, but I’d make “OpenAI-compatible” a real provider contract rather than only a preset with three fields.

A few details worth baking into the first version:

- Normalize `base_url` once and test path joining (`/v1`, no `/v1/v1/...`, no accidental fallback to the hosted OpenAI URL).
[...]

PR #394, opened 2026-05-28
feat(pi-agent): discover models from /v1/models endpoint

Tomas2D:
## Summary

- `pi-dynamic-providers` extension now fetches `/v1/models` from the configured `${envPrefix}_URL` and registers every returned id, instead of the single env-defined model.
- `${envPrefix}_MODEL` becomes an optional default-selector matched case-insensitively against the discovered list. When discovery fails (timeout, non-2xx, malformed response), the extension falls back to the prior single-model behavior and emits a `console.warn` so silent degradation is visible to operators. vLLM's `max_model_len` becomes per-model `contextWindow` and caps `maxTokens`.

Step toward #393.

## Test plan

- [ ] `mise run cluster:build-agent` and verify discovered models appear in pi's model picker for both `RITS_URL` and `OPENAI_PROXY_URL` configurations.
- [ ] With `${envPrefix}_MODEL` unset, confirm the first discovered model is used as default; with it set, confirm the case-insensitive match selects the intended default.
- [ ] Point `${envPrefix}_URL` at an unreachable host and confirm the extension emits `[pi-dynamic-providers] ...` warnings and (when `_MODEL` is set) still registers the single-model fallback.
- [ ] Hit a vLLM proxy that reports `max_model_len` smaller than the env `_MAX_TOKENS` default and confirm `maxTokens` is capped at `contextWindow`.
- [ ] Existing shadow-of-built-in-openai behavior still hides the built-in provider when `OPENAI_PROXY_URL == OPENAI_BASE_URL`.

### DIFF
--- a/packages/agents/pi-agent/workspace/.pi/agent/extensions/pi-dynamic-providers/index.ts
+++ b/packages/agents/pi-agent/workspace/.pi/agent/extensions/pi-dynamic-providers/index.ts
@@ -1,14 +1,22 @@
 import { mkdirSync, readFileSync, writeFileSync } from "node:fs";
 import { homedir } from "node:os";
 import { join } from "node:path";
-import type { ExtensionAPI, ProviderConfig } from "@earendil-works/pi-coding-agent";
+import type { ExtensionAPI, ProviderConfig, ProviderModelConfig } from "@earendil-works/pi-coding-agent";
 
 declare const process: { env: Record<string, string | undefined> };
 
-// Each spec is activated only when ${envPrefix}_URL and ${envPrefix}_MODEL are
-// set on the pod. Auth is injected on the wire by the Envoy sidecar's
-// credential_injector filter; the apiKey here only exists to satisfy
-// pi-acp's per-session auth gate (reads models.json.apiKey + auth.json).
+// Each spec is activated when ${envPrefix}_URL is set on the pod. Models are
+// discovered from the OpenAI-compatible /v1/models endpoint; ${envPrefix}_MODEL
+// is optional and only used to pick the default when it appears in the
+// discovered list. If discovery fails, we fall back to a single-model
+// registration as long as ${envPrefix}_MODEL is set — otherwise the spec is
+// skipped.
+//
+// Auth is injected on the wire by the Envoy sidecar's credential_injector
+// filter; the apiKey here only exists to satisfy pi-acp's per-session
+// auth gate (reads models.json.apiKey + auth.json), and the discovery fetch
+// intentionally does not set Authorization so envoy can inject it the same way
+// it does for the actual model traffic.
 
 type ProviderSpec = {
 	name: string;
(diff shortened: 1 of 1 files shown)


packages/agents/pi-agent/workspace/.pi/agent/extensions/pi-dynamic-providers/index.ts
```diff
@@ -30,103 +49,168 @@ const SPECS: ProviderSpec[] = [
...
-	let modelsFile: { providers: Record<string, ProviderConfig> } = { providers: {} };
-	try {
-		const parsed = JSON.parse(readFileSync(modelsPath, "utf8")) as { providers?: Record<string, ProviderConfig> };
-		if (parsed && typeof parsed === "object" && parsed.providers) {
-			modelsFile = { providers: parsed.providers };
-		}
-	} catch {
-		// models.json may not exist yet on a fresh home; start from empty.
+	let lastActivated: Activation | undefined;
+	for (const spec of SPECS) {
```
xjacka, line 56: 🟢 **Suggestion:** Specs are discovered sequentially here, so two unreachable upstreams stack their `AbortSignal.timeout`s (~10s worst-case pod startup). Discovering with `Promise.all` would parallelize the fetches — just keep the last-activated default deterministic (e.g. pick the default by SPECS order after all resolve).

xjacka (approved):
🛡️ **DAM** — ✅ Code Review @ `c3d08f8`

## PR #394: feat(pi-agent): discover models from /v1/models endpoint
**Author:** Tomas2D | **Branch:** feat/pi-agent-discover-models → main | **Changes:** +169 −85 (1 file)

### Summary
The `pi-dynamic-providers` extension now discovers every model from the OpenAI-compatible `/v1/models` endpoint instead of registering a single env-defined one. `${envPrefix}_MODEL` becomes an optional, case-insensitive default-selector; on discovery failure it falls back to the prior single-model behavior (when `_MODEL` is set) and `console.warn`s so degradation is visible. vLLM's `max_model_len` becomes per-model `contextWindow` and caps `maxTokens`.

### Findings
- ✅ **Looks good:** `discoverModels` is thoroughly defensive — configurable `AbortSignal.timeout`, `!res.ok` handling, `data`-array validation, per-entry type checks, dedupe by lowercased id, and a `console.warn` on every degradation path so operators get a signal.
- ✅ **Looks good:** Fallback logic is correct — discovered models → single-model (when `_MODEL` set) → skip; default selection matches `_MODEL` case-insensitively and prefers the *discovered* casing so the wire id matches upstream. `models[0]` access is guarded by the earlier `length === 0` return.
- ✅ **Looks good:** `maxTokens = Math.min(_MAX_TOKENS, contextWindow)` correctly prevents requesting more output than the model's `max_model_len` allows.
- ✅ **Looks good:** The state refactor (`loadState`/`persistState` + `readJson`/`writeJson`) is a clean extraction with no behavior regression to the disk-mirroring or shadow-provider logic.
- 🟢 **Suggestion:** The two `SPECS` are discovered sequentially (`await` in the loop), so two unreachable upstreams stack their timeouts (~10s worst-case startup). `Promise.all` over the specs would parallelize discovery; just keep the last-activated default deterministic.
- 🟢 **Suggestion:** The pure pieces (the `/v1/models` parse/dedupe, `buildModelConfig` capping, default selection) are unit-testable without a cluster — a couple of tests would lock in the fallback/casing/cap behavior the test plan currently exercises manually.

### Verdict
APPROVE — Robust, well-factored discovery with comprehensive fallback and operator-visible warnings; no correctness concerns. (PR test-plan checkboxes are still unchecked — manual cluster verification per the plan is the only outstanding item.)

---

PR #401, opened 2026-05-28
chore(lima): bump dev VM disk to 200 GiB

tomkis:
## Summary
- 100 GiB fills up after a few `cluster:install` cycles; running into `No space left on device` mid-build was a recurring papercut
- 200 GiB gives headroom for image churn + containerd build cache between `cluster:reclaim` runs

Only affects newly created VMs (same caveat as the existing `memory:` comment in the file). Existing devs keep their 100 GiB VM until they reprovision.

## Test plan
- [ ] `mise run cluster:delete && mise run cluster:install` on a fresh VM
- [ ] Confirm guest reports 200 GiB: `limactl shell platform-k3s df -h /`

### DIFF
--- a/deploy/lima-k3s.yaml
+++ b/deploy/lima-k3s.yaml
@@ -16,7 +16,7 @@ cpus: 4
 # from a 4 GiB VM must reprovision:
 #   mise run cluster:delete && mise run cluster:install
 memory: 8GiB
-disk: 100GiB
+disk: 200GiB
 
 containerd:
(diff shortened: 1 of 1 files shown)


JanPokorny:
Lima virtual disks only expand when actually used so this is faily non-consequential. But I'll have to look into the build process since this is not supposed to be happening -- we never hit this limit on Agent Stack so most likely we have some non-deterministic builds.

Issue #399, opened 2026-05-28
Server-side session list should be a cache over agent-owned sessions


jezekra1:
## Problem

The api-server's `sessions` table is treated as the canonical list of sessions
for an agent, but the agent runtime actually owns sessions and creates them in
places the server doesn't observe:

- **Schedule-triggered sessions** (#391) run inside the runtime; no DB row.
- **Terminal `/clear`** creates a fresh session in-place inside the agent; the
  server still references the old `sessionId`.
- Rows are only written when the api-server intercepts the first `session/prompt`
  RPC ([`acp-relay.ts:247`](packages/api-server/src/apps/api-server/acp-relay.ts#L247)).
  Anything off that path is invisible.

`sessions.list` already overlays live ACP data
([`sessions-service.ts:72-116`](packages/api-server/src/modules/sessions/services/sessions-service.ts#L72-L116)) —
title and `updatedAt` come from `acp.listSessions()` — but DB existence still
gates visibility, so agent-only sessions stay hidden.

Fixing #391 alone papers over one symptom; the structural issue is two stores
both claiming ownership of the list.

## Proposed solution

Treat the server's session table as a **cache** over the agent's session list.
The agent (`acp.listSessions()`) is the source of truth for *what sessions
exist*. The server stores only enrichments the agent can't know about: `mode`
(chat/terminal), `type` (UI / Slack / Telegram / schedule), `scheduleId`,
`threadTs`, `createdAt`.

`sessions.list` returns the agent's list joined against cached enrichments.
Cache misses don't hide sessions — they just show without enrichment. Stale
rows (agent no longer has the session) get pruned.

### Open questions

- How does enrichment round-trip terminal/schedule binding — does it stay
  server-side keyed by `sessionId`, or do we push some metadata down into the
  agent's session record so the agent can report it back?
- Cold-start: how to handle a session created via the UI before the pod is up
  (server has enrichment, agent has nothing yet)?

### Out of scope

- The schedule-visibility regression itself — tracked in #391; this issue is
  the structural fix it implies.
- ACP transcript / message persistence.

Related: #391, #381

Issue #391, opened 2026-05-28
Scheduled sessions don't appear in the UI


tomkis:
## Problem

When a session is started by a schedule (rather than from the UI), the session never shows up in the session list. The agent runs, but there's no way to see it ran, open it, or read its output from the UI.

This regressed after the unified Connections + runtime channel rework (#381). Schedule-triggered sessions now run entirely inside the agent runtime and never get recorded alongside the sessions the UI knows about, so they're invisible to the user.

For someone relying on schedules, this is the whole point of the feature missing: the schedule fires, the work happens, and they have no record of it.

## Proposed solution

A session created by a schedule should show up in the session list just like one started from the UI, including its history, status, and any output. Opening it from the list should work the same way.

### Out of scope

- Changes to how schedules themselves are configured or fired.
- Surfacing additional schedule-specific metadata in the list (separate UX question).

PR #405, opened 2026-05-28
docs(adrs): ADR-055 — agent-owned session metadata via ACP _meta

jezekra1:
## Summary
- Proposes ADR-055: flip the source-of-truth for session existence from the server's Postgres `sessions` table to the agent itself; metadata travels via standard ACP `_meta` on `session/new` and `session/list`.
- Server-side `sessions` table becomes a Redis cache (per ADR-036). Agent-runtime owns the metadata round-trip via its existing ACP proxy layer; harness stays vanilla.
- Delete uses ACP `ExtRequest` (`platform/deleteSession`) with hard/soft fallback depending on harness support — same extensibility pattern Zed uses for capability-gated `delete_session`.
- Supersedes ADR-017 in part (DB → agent for source-of-truth); amends ADR-019 (narrows the no-`_meta` clause to harness-specific extensions only).

## Why
- #391 — schedule-fired sessions never appear in the UI because the relay-intercept write trigger doesn't fire for them.
- #399 — server's `sessions` table behaves like a cache but isn't one; terminal `/clear` and other agent-internal mintings stay invisible.
- ADR-026 (agent-runtime is session-state-authoritative), ADR-037 (TUI mode bypasses ACP), and ADR-052 (in-process trigger dispatch) all moved ownership toward the agent; ADR-017's framing hasn't kept up.

## What this PR is not
- No code changes — ADR only. Implementation will follow once accepted.
- No migration script yet — the ADR's "drop Postgres table + one-shot backfill" is described but unimplemented.

## Verification done while drafting
- Confirmed `claude-agent-acp@0.33.1` accepts `_meta` on `session/new` (no error) but does not persist it or return it on `session/list` (`acp-agent.js:374-389`, `:1281+`). The runtime intercept is the workaround.
- Confirmed `closeSession` throws "Session not found" for idle sessions (`acp-agent.js:858`), ruling out the "close-with-delete-marker" approach.
- Confirmed Zed's "Archive Thread" is purely client-side (SQLite flag); its **delete** is the capability-gated ACP-extension pattern this ADR adopts.

## Test plan
- [ ] Architecture/Sessions reviewer reads the ADR
- [ ] Reviewer agrees ADR-017 partial-supersede and ADR-019 amendment are appropriate
- [ ] Reviewer agrees Redis (not Postgres) is the right cache substrate

### DIFF
--- a/docs/adrs/055-agent-owned-session-metadata.md
+++ b/docs/adrs/055-agent-owned-session-metadata.md
@@ -1,0 +1,125 @@
+# ADR-055: Agent-owned session metadata via ACP `_meta`; server sessions become a Redis cache
+
+**Date:** 2026-05-28
+**Status:** Proposed
+**Owner:** @jezekra1
+**Supersedes:** ADR-017 (in part)
+**Amends:** ADR-019
+
+## Context
+
+ADR-017 placed the Postgres `sessions` table as the source of truth for session existence and metadata, with the ACP relay writing a row on the first `session/prompt` it intercepted. Three later decisions reshaped the landscape:
+
+- **ADR-026** made the agent-runtime "session-state-authoritative" — it owns per-session logs and serves `session/load` from its own log when populated.
+- **ADR-037** split sessions into `chat` and `terminal` modes — terminal mode runs the harness in TUI against a PTY and **bypasses ACP entirely**.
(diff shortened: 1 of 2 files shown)


tomkis (approved):
Great proposal.

PR #420, opened 2026-05-29
feat: agent-owned session metadata via ACP _meta — no server-side session store (ADR-055)

jezekra1:
## What

Implements **ADR-055**: the agent is the sole source of truth for session existence and metadata, surfaced over ACP `_meta.platform`. There is **no server-side session store** — no Postgres `sessions` table, no Redis cache, no server sessions data-layer. Every reader (UI, Slack/Telegram workers) lists sessions directly over its own ACP connection (auth-gated by the relay). The server retains only the one schedule-scoped op that needs durable delivery.

Supersedes ADR-017, amends ADR-019.

## Highlights

- **agent-runtime** owns session existence + metadata in an on-disk document store under `<agentHome>/.platform/`, over ACP `_meta.platform`: captured on `session/new`, injected on `session/list`, merged on resume (setMode), soft-deleted via a tombstone on the `platform/deleteSession` ext-method. `session/list` treats the **harness list as ground truth** — the store only annotates it, never invents rows.
- **api-server** drops the session store entirely (Postgres repo + Redis cache); the ACP relay is a dumb proxy again. The only server-side session op is the durable schedule reset, folded onto the schedules service as `schedules.resetSession`.
- **channels** (Slack/Telegram) self-resolve their thread session over their own ACP connection via `_meta.platform.threadTs`.
- **ui** lists/creates/deletes/re-modes sessions directly over ACP; the schedule view lists over ACP and filters by `scheduleId`.
- **db** drops the `sessions` table + `session_mode` enum, plus the five session-derived `usage_*` views (no session-lifecycle event stream to re-source them; the schedule-fire views stay, sourced from `activity_events`).

## Behavioral note

A freshly-created chat now appears in the sidebar once its first message writes JSONL (it's fully usable while active — that's client state). No more empty "(no title)" shells, and the store-only "Session not found" ghost screen is gone.

## Testing

- `mise run check` — green
- `mise run test` — green (98 + 166 + 144 + 39)
- The `0020` migration applies cleanly on the running cluster (confirmed after the dependent-views fix).

## Follow-ups (out of scope)

- CLI `chat/services/sessions-service.ts` still calls the removed `trpc.sessions.list` / `resolveTerminal` (loosely typed, runtime-dead) — pre-existing from the reshape; needs a separate cleanup pass.
- The UI config-probe still writes a now-unsurfaced store entry per cold-load; could gate recording on platform meta to avoid store litter.


packages/cli/src/__tests__/sessions-service.test.ts
PetrBulanek, line 1: Do we still want to add new tests? Not sure if they add much value. I think @tomkis prefers the e2e tests he's currently working on.

packages/cli/package.json
```diff
@@ -32,6 +32,7 @@
     "prepack": "pnpm build"
   },
   "devDependencies": {
+    "@agentclientprotocol/sdk": "^0.17.1",
```
PetrBulanek, line 35: `@agentclientprotocol/sdk` is added under `devDependencies` but isn't in tsup's `noExternal` allow-list. It only bundles today because tsup bundles devDeps by default. If it's ever promoted to `dependencies` (matching api-server/ui, which both declare it `^0.17.1`), tsup will externalize it and the published CLI will fail at runtime with a missing-module error. Please add `@agentclientprotocol/sdk` to `noExternal` to match the rest of the runtime deps.

packages/cli/src/modules/chat/infrastructure/acp-session-client.ts
PetrBulanek, line 1: `wsStream`/`withConnection` has no timeout. The reference `api-server/core/acp-client.ts` guards with a 120s inactivity timeout; this client has none. If the relay accepts the socket but the agent never answers `initialize`/`listSessions`/`resume`, `dam chat` and `dam session list` hang indefinitely.

PetrBulanek (approved):
I reviewed only the CLI and UI parts of this PR. I only found minor stuff we might fix (none blocking). Left as separate comments.

tomkis (approved):

Issue #404, opened 2026-05-28
E2E tracer bullet: Playwright + fresh cluster + mock agent


tomkis:
## Problem

No end-to-end tests. The vertical slice from a user logging in, accepting T&C, and driving an agent has never been exercised as a single automated flow. Bugs in any of those layers slip through unit tests, and manual verification doesn't happen consistently.

ADR-014's API-level vitest e2e is superseded. See ADR-056.

## Proposed solution

Thinnest possible browser-driven slice that proves the test architecture works. Real cluster, real browser, real auth. Agent is mocked and runtime-controllable so assertions are deterministic.

Story runs as two coupled specs in order:

1. user logs in via Keycloak UI, accepts T&C, storage state saved
2. spec sets the mock agent's script via tRPC, user creates an Agent from the mock Template, starts it, sends a prompt, asserts the scripted reply appears in chat and asserts the prompt reached the mock

If step 1 fails, step 2 skips. Cluster is provisioned fresh per run via `mise run e2e` (delete prior, install fresh, run Playwright, tear down in trap).

### Decisions

- `mise run e2e` owns cluster lifecycle. CI and local run the same command.
- Two specs, serial, one shared user, hardcoded resource names.
- Browser storage state from spec 01 carries login + T&C into spec 02.
- Test-only artifacts ship in the chart behind `e2e.*` values flags, default off. Test cluster install flips them on.
- Mock agent is controllable via a test-only tRPC namespace on the api-server, gated by the same `e2e.enabled` flag. The namespace opens a tRPC WebSocket client against the existing `ws://${podBaseUrl}/api/acp` endpoint; agent-runtime's relay forwards arbitrary JSON-RPC frames over stdio without method validation, so tRPC envelopes ride the ACP channel transparently. The mock harness detects `method: "query"|"mutation"` on stdin and dispatches to its existing `scriptedMock` router. Single transport, single pod port, full tRPC typing end-to-end. Zero changes to agent-runtime, the agent StatefulSet, or the K8s Service.
- Control surface and control transport are both ACP-frame-level. Spec sets the sequence of frames the mock emits in response to prompts, and reads back the frames the mock received from agent-runtime.
- Per-PR run, advisory until 10 clean merges, then required.
- Existing API-level vitest e2e tests are removed.

### Work items

1. Mock agent [x]
- [x] New package `e2e/agents/mock/`
[...]

PR #428, opened 2026-06-01
feat(e2e): mise run e2e task + spec hardening

tomkis:
Stacks on #407. Closes item 6 of #404.

## Summary

- New `mise run e2e` task. Owns the full test-cluster lifecycle: nuke prior VM, provision fresh with `values-e2e.yaml` overlay, run Playwright, tear down in trap regardless of outcome. Supports `--headed`. Replaces `mise run api-server:e2e`.
- Test cluster uses `domain=localhost`, port 5555. `*.localhost` is a secure-context origin in Chromium so `crypto.subtle` is available and oidc-client-ts can generate PKCE.
- `E2E_MOCK_ONLY=1` narrows `image:agent` and the docker-save tarball to platform-base + mock. `values-e2e.yaml` drops the other agentTemplates so the chart matches what's loaded.
- After helm install, targeted `rollout restart` of `platform-apiserver` only. Bouncing UI/controller introduces a transient Traefik-endpoint 502 window that breaks Playwright's first asset fetch; just api-server needs the bounce to dodge the postgres init-databases.sh race.
- Test VM bumped to 8GiB to match the dev VM's istio-ambient sizing.

Also folds in two pre-existing fixes uncovered while shaking this down:
- `01-auth` spec waited for `origin === baseUrl` which resolved at `/auth/callback?code=...` before the SPA finished its post-callback redirect. Now waits past `/auth/callback` and settles on terms button or app sidebar.
- `tinyexec@1.2.2` added to `trustPolicyExclude` — `pnpm deploy --legacy` in the platform-base Dockerfile bypasses the lockfile and trips on a publisher-side trust regression.

## Test plan

- [x] `mise run e2e` passes locally (2 specs in 31s, fresh VM, clean teardown)
- [x] `mise run e2e -- --headed` runs with visible browser
- [ ] CI run pending


tomkis:
Superseded by #430 (rebased onto the squash-merged base of #417 so it doesn't carry the now-duplicate pre-squash commits).

PR #429, opened 2026-06-01
build(cluster): build agent images only for enabled templates

jezekra1:
## What

Local cluster tasks (`image:agent`, `cluster:install`, `cluster:build-agent`) previously built and loaded **all** agent images unconditionally. Now they build/load only the agent templates marked `enabled: true` in `deploy/helm/platform/values-local.yaml` — the same file `cluster:install` deploys with.

Set an agent to `enabled: false` (or remove it) in `values-local.yaml`, and it's skipped end-to-end: not built, not saved, not loaded into k3s.

## How

- **`image:agent`** — replaced the five hardcoded `docker build` lines with a loop over enabled templates. `platform-base` is always built (every agent's base layer); each enabled agent is built from `packages/agents/<name>` and tagged with that template's `image.repository:tag`. Reading the repository from values is necessary because it isn't always `platform-<name>` (e.g. `google-workspace` → `platform-google-workspace-agent`).
- **`cluster:install`** — the `docker save` agent list is derived from the enabled templates, appended after the always-present core images.
- **`cluster:build-agent`** — same dynamic list, plus an early no-op when no templates are enabled.

## Notes

- Reads via `yq` (mise-managed) with `select(.enabled)`, which treats both `enabled: false` and a missing key as off.
- Uses process substitution and whitespace-free image refs — both already idioms in this file (`cluster:reclaim`) — to stay compatible with macOS bash 3.2 (avoids `mapfile`).
- These tasks aren't referenced by CI, so the release pipeline is unaffected. `cluster:helm` inherits the fix (it calls `cluster:install` with `SKIP_IMAGE_BUILD=1`).

## Testing

Verified the build loop and save-list logic against the real `values-local.yaml` (all five enabled agents resolve to existing build dirs) and against a copy with agents disabled (filtered out correctly). Not yet run against a live cluster.


packages/agents/bob/tasks.toml
```diff
@@ -0,0 +1,9 @@
+["image:agent:bob"]
```
JanPokorny: For consistency these should be `agents:bob:image` etc. (`<subproject>:<task>`)

jezekra1: Done — renamed to `agents:<name>:image` (and `platform-base:image`) in a3d972f9, matching the `<subproject>:<task>` convention. The `image:agent` aggregate keeps its name (ADR-023 references it) and now globs `agents:*:image`.

JanPokorny (approved):

PR #457, opened 2026-06-02
Feat: Gate agent readiness on import, env, applystate etc.

### LINKED ISSUE #168: Block opening an agent until it's ready, and lock the detail view when it sleeps
## Problem

Today the agent detail view can be opened (or stays open) when the agent is not in a state that supports interaction — either before it has finished starting up, or after it has gone to sleep. The result is a half-working screen: panels show cryptic errors like "Files not available" or "instance is hibernated; start it before browsing private sources", inputs appear active, and the user has no clear signal that the agent simply isn't there to respond.

Two distinct flows hit this:

- **Not-yet-ready agents.** A user opens an agent that is still coming up. The detail view loads, but parts of it fail piecemeal and the user can try to interact even though nothing will work end-to-end.
- **Sleeping agents.** An agent that was running goes to sleep while the user is on (or returns to) the detail view. The screen continues to look interactive — message input, source browser, panels — but individual actions fail one by one with errors that only make sense if you already understand the lifecycle.

In both cases the user is left poking at a screen that won't respond, with no single, obvious cue that the agent is offline.

## Proposed solution

- **Gate opening on readiness.** A user should not be able to navigate into the agent detail view at all while the agent is not ready. The entry point should reflect that the agent is starting / not available, and clicking should either be a no-op or surface a clear "starting…" state — never a half-loaded detail screen.
- **Lock the detail view when the agent is offline.** When an agent is sleeping (or otherwise not interactive), the detail view should present a single, screen-level offline state: it's visually obvious the agent is asleep, all interaction surfaces (chat input, source browsers, action buttons, panels) are disabled, and the only meaningful action is to wake / start it. No per-panel error messages, no inputs that look usable but fail on submit.

The goal is one clear signal per state, applied consistently across the whole screen, instead of many small errors scattered across panels.

### Out of scope

- The exact visual treatment of the locked state (overlay vs. dimmed surfaces vs. empty-state replacement) — to be decided in design.
- Changes to the underlying lifecycle states themselves; this is purely about how the UI reflects them.

## Additional problems (surfaced during implementation)

Two backend correctness problems came up while implementing the readiness gating (see #457 / ADR-058) and should be tracked here:

1. **Newly-created agents restart their pod immediately.** A freshly-created agent starts its pod, then has secrets/connections attached, which bumps `secrets-rev` and rolls the pod — so the first thing a brand-new agent does is restart. This is self-inflicted (grants are applied *after* the pod is materialized), and it looks broken to anyone watching the cluster — a cluster operator seeing a just-created workload immediately roll will assume something is wrong. The grant set is already known at create time, so creation should be single-shot: materialize the StatefulSet once at the final revision, with no startup roll.

2. **`applyState` isn't safe to send to a pod that's restarting.** It isn't atomic with respect to pod termination — an apply dispatched to a pod that's about to be replaced can land on the doomed pod (and falsely advance the settled cursor). Graceful termination should make this safe: the pod should either reject `applyState` once it's draining (retryable error) or hold off the restart until any in-flight apply finishes, so an apply is always either cleanly applied to a live pod or cleanly refused — never half-applied to a dying one.

jjeliga:
# Summary

Closes #168. Makes an agent report "running" only when it's actually usable — pod ready, runtime state applied (or settled with surfaced failures), and any create-time file import landed — instead of acking optimistically and wedging or silently degrading. Backed by ADR-058.

# Why

Previously applyState could throw and wedge an agent mid-startup, failed contribution installs were invisible, a freshly-created agent flipped to "running" before its file import landed, and a half-failed create left a silent, broken-but-"running" zombie.

# What changed

### Apply path & readiness gate:
applyState now settles to termination and returns failures instead of throwing. Split the conflated cursor: last_settled_version (drives the gate) vs last_applied_version/last_applied_hash (clean-only high-water mark), plus apply_failures / apply_attempts.
New "Contributions Settled" gate: ensureReady waits on podReady ∧ currentState=running ∧ contributionsSettled ∧ ¬importPending; computeAgentState projects the same to "running". No outbox row ⇒ settled. Contribution-status reads are fail-soft (DB error ⇒ default settled).
Failed installs retry in the background (cron-sweep, capped at 8 attempts) and surface via ContributionApplyFailed / ContributionRecovered events → activity_events, and a live amber degraded badge (contributionFailures) on the agent list tile and chat header. Clean success is silent.

### File-import gate:

An agent created with files declares pendingImport (TTL'd annotation) and gates "running" until the import lands; failure notifies (toast) but never blocks.
ensureImportable waits for a stable pod (reachable, not terminating, env-roll settled) — not for non-rolling contributions. The controller echoes the reconciled secrets-rev into agent status (AppliedSecretsRev) so the api-server detects a pending env-roll without reading the StatefulSet.
Import is dispatched after grant/connection mutations (so the gate sees the triggered roll), with a reachability preflight before the one-shot upload (closes the kubelet-Ready vs. mesh-routable window).

### Atomic create-failure:

A create that fails after the ConfigMap exists is marked in one atomic write (create-error annotation + clear import gate + hibernate) and surfaces as the red Error pill instead of a silent zombie. Recovery: delete + recreate.

### Schema:

Migration 0021: adds last_settled_version, apply_failures, apply_attempts to runtime_state_outbox and swaps the stale index for a retry index. Additive, defaults backfill.

Closes #168


packages/api-server/src/modules/runtime-delivery/services/hello-handler.ts
```diff
@@ -24,27 +25,10 @@ export function createHelloHandler(deps: {
...
-      if (!divergedVersion && !divergedHash && !hasEvents) {
-        return { events: [] };
-      }
-
-      return {
-        version: row.version,
-        state: { contributions: payload.contributions, hash: payload.hash },
-        events: payload.events,
-      };
+      return { events: [] };
```
jezekra1, line 31: let's change the hello response altogether, no reason to always deliver empty events

jezekra1:
## Design question: the env-gated apply deferral is compensating for a restart-blind gate

Re ADR-058 point 2 — `contribution-fanout.ts` skipping `enqueueAfterCommit` when a mutation includes an `env` contribution:

```ts
// env triggers a pod roll; let the new pod's hello drive the worker.
if (!hasEnvContribs) {
  await deps.runtimeMutator.enqueueAfterCommit(agentId);
}
```

I think this special-case is treating a symptom. **A pod can restart for many reasons — OOM, eviction, node drain, manual `kubectl delete pod`, a liveness-probe flap — not just an env-induced roll.** The delivery/gate machinery has to be correct under arbitrary restarts; once it is, no operation needs to know that env rolls the pod.

### Why it's load-bearing today (so it can't just be deleted)

The "Contributions Settled" gate reads a **server-side monotonic counter**, completely pod-identity-blind:

- the gate reads `last_settled_version >= version` (`outbox-repo.ts:183`)
- `recordOutcome` advances `last_settled_version` purely from the apply *result* — it never records which pod generation applied it (`outbox-repo.ts:152-163`)
- `bumpVersion` resets `apply_attempts` but never rolls `last_settled_version` back (`outbox-repo.ts:94-107`)

So if the fanout always enqueued, the enqueue fires while the **old** pod is still connected (`isRunning === true`). The worker applies to that doomed pod, succeeds, and `recordOutcome` advances `last_settled_version` to `version`. The pod then rolls; the fresh pod is blank until its own hello-driven re-apply lands, but the gate already reads **settled** and `podReady` (kubelet-local) can already be true → the agent flips to "running" while unconfigured. The `"stale apply"` guard doesn't catch this — the old pod has no idea it's about to be replaced.

That's exactly what the env branch prevents. But the same stale-positive window exists for **every** restart cause, with no env involved — so the branch only patches one instance (and not the most common one) of a general gate bug.

### The generalization (what would let the special-case go away)

Three changes, each removing the need for restart-awareness in operations:

1. **Single-shot create — don't roll on creation.** The creation roll is self-inflicted: the pod starts at rev-1, then `setAgentAccess` + `setAgentConnections` run *after* create, each `bumpSecretsRev` → rev-2 → roll. `useCreateAgent` already has `secretIds`/`appConnectionIds` in hand, and `BuildAgentStatefulSet` already takes `credentialSecrets` from the CM. Folding the initial grant set into the create write materializes the StatefulSet once at the final rev. This deletes the creation roll, the "import LAST" ordering caveat, and `secretsRevApplied`/`ensureImportable` *for the create path* (the import gate collapses to "pod reachable").

2. **Graceful termination makes `applyState` safe anytime.** If preStop closes the ACP connection first (so `isRunning` flips false and the worker won't dispatch — `worker-handler.ts:29`), declines new `applyState` with a retryable error while draining, and finishes in-flight applies before exit, then an apply is always either cleanly applied to a live pod or cleanly refused. `recordOutcome` already leaves the cursor untouched on a non-clean result (`outbox-repo.ts:138-151`), so a refused apply can't falsely advance the gate. With this, there's no race to defer around — the fanout can go back to unconditional `bump + enqueue`.

3. **Gate on live-pod truth, not the monotonic counter.** The agent already reports `lastAppliedVersion` on every `hello` (`hello-handler.ts:28`) and resets to 0 on a fresh pod. Store that per-agent and gate on `agentReportedAppliedVersion >= version`. On restart the new pod's first hello drives the stored value *down* → gate shows "starting" → re-apply → gate recovers. `last_settled_version` stays, demoted to sweep/retry bookkeeping. This makes env-roll, OOM, and drain all behave identically — which is the goal.

### Net

With all three, the fanout has zero pod-lifecycle knowledge, creation never rolls, and restart-correctness becomes a property of the protocol rather than something each operation opts into. The env branch and the `AppliedSecretsRev` echo both disappear.

Since ADR-058 is still **Proposed** and shipping in this PR, this feels like the moment to reshape it rather than cement the special-case. Change (1) is the most self-contained and immediately testable if we want to start there. Curious what @janjeliga thinks — am I missing a reason the gate is intentionally monotonic across pod identity?

packages/api-server/src/apps/api-server/app.ts
```diff
@@ -500,41 +584,54 @@ export function startApiServerApp(deps: ApiServerAppDeps) {
...
-          const responseHeaders = new Headers();
-          for (const [name, value] of Object.entries(upstreamRes.headers)) {
-            if (value === undefined) continue;
-            if (!PROXY_RESPONSE_HEADER_ALLOWLIST.has(name.toLowerCase()))
-              continue;
-            responseHeaders.set(
-              name,
-              Array.isArray(value) ? value.join(", ") : value,
+          const status = upstreamRes.statusCode ?? 502;
+          if (status >= 200 && status < 300) {
```
jezekra1, line 588: A lot of logic lives now in app.ts, this is infra-layer code, should we move it somewhere else?
Did you run this through /typescript-engineering?

packages/api-server/src/modules/runtime-delivery/services/worker-handler.ts
```diff
@@ -56,33 +58,73 @@ export function createWorkerHandler(deps: WorkerHandlerDeps): WorkerHandler {
...
-    await deps.outboxRepo.stampAck(
-      agentId,
-      result.appliedVersion,
-      result.appliedHash,
-    );
+    // recordOutcome diffs under a row lock and returns the transitions; emit post-commit.
+    const { newlyFailed, recovered, gaveUp } =
+      await deps.outboxRepo.recordOutcome(agentId, row.version, settle);
+    for (const f of newlyFailed) {
+      emit({
```
jezekra1, line 111: wonder if this should really flow through the sagas when the only consumer is the writer to db, why not write to db directly here? What's your take @tomkis?

packages/api-server/src/modules/agents/infrastructure/agents-repository.ts
```diff
@@ -236,33 +386,54 @@ export function createAgentsRepository(k8s: K8sClient): AgentsRepository {
...
-          await bumpLastActivity(id);
-          return;
-        }
-        await repo.wakeIfHibernated(id);
-        const ready = await pollUntilReady(
-          () => repo.isPodReady(id),
-          WAKE_POLL_INITIAL_MS,
-          WAKE_POLL_MAX_MS,
-          WAKE_TIMEOUT_MS,
+    async ensureImportable(id) {
```
jezekra1, line 421: Still don't understand why the import needs a special condition, different (stronger) from the typical readiness. Either the agent is ready or is not ready (when readiness takes into account settlements, initial pod-rolls and stuff), or am I missing something?

packages/api-server/src/modules/agents/infrastructure/labels.ts
```diff
@@ -55,3 +55,15 @@ export const ANN_GRANTED_CONNECTION_IDS =
 // rows). `injectionConfig` (ADR-028) currently has no fanout; if a future
 // change wires it through this rev, update this comment then.
 export const ANN_SECRETS_REV = "agent-platform.ai/secrets-rev";
+
+// In-progress file import (value = ISO timestamp; gates "running", TTL'd as a crash backstop). Cleared on success; "" / absent = none.
+export const ANN_PENDING_IMPORT = "agent-platform.ai/pending-import";
```
jezekra1, line 60: let's not add new annotations, I feel strongly that this is an anti-pattern, the API server should not write state to the infra this way.

PR #586, opened 2026-06-03
feat: migrate Agent + Fork from ConfigMaps to CRDs (ADR-058/059)

jezekra1:
Migrates the two reconciled domain resources — **Agent** and **Fork** — from labeled ConfigMaps to first-class custom resources (`agent-platform.ai/v1`), per **ADR-058** and **ADR-059** (decision PR #552). Templates stay ConfigMaps (chart-shipped declarative config); schedules stay in Postgres.

Big-bang cutover — no data migration. Goes live on the next co-released deploy.

## What changed, by phase

**Phase 1 — CRD foundation** (`11e36be9`): Go-first `Agent`/`Fork` types + controller-gen; chart-bundled CRD manifests gated on `crds.enabled`.

**Phase 2 — controller** (`452db962`→`82cc1666`):
- Reconciles `Agent`/`Fork` CRs off dynamic informers.
- Activity-driven hibernation — `desiredState` eliminated; reconciler scales **up** only, idle-checker is the sole probe-gated scale-**down** authority, wake = `last-activity` poke.
- Readiness published on the status subresource: `Ready = AgentPodReady ∧ GatewayPodReady` + `Reconciled` condition, recomputed via a pod informer (loop-safe no-op status guard). `phase` is a derived human summary.
- `roll-rev` annotation → controller stamps both pod templates for rolling restart without spec/status writes.
- Grants moved into spec; status writers, orphan sweeps, idle-checker all on CRs.

**Phase 3 — api-server** (`1c78bfba`, `9c60df51`):
- `K8sClient` gains generic `CustomObjectsApi` methods (built internally — no composition-site ripple).
- Forks + agents created/read/patched/deleted as CRs; `updateSpec` is a conflict-free merge-patch.
- Run-state derives from `.status.phase`+conditions; the public `AgentState` enum is unchanged, so the UI contract is untouched.
- `wake` is an activity poke; `restart` is a `roll-rev` patch; grants in spec; harness `resolveAgent`, connections fan-out, skills gating all follow.

**Phase 4 — RBAC** (`9f755d38`): controller gets `agents`/`forks` watch + `*/status` write; api-server gets agent/fork CRUD; pods narrowed to read-only (restart no longer deletes pods).

## Testing
- `mise run controller:check` + `controller:test` green.
- api-server format/lint/tsc + **181 unit tests** green; all other unit suites green.
- `helm lint` + `template | kubeconform` green; CRDs ship by default.
- Every commit passed the full repo pre-commit `mise run check`.

## Not yet done / follow-ups
- **E2E not run** — `mise run e2e` against a live cluster should validate the big-bang before merge (needs RBAC + CRD live).
- `desiredState` remains an optional, now-unused field on the `api-server-api` `AgentSpec` (harmless; optional cleanup).
- Architecture pages under `docs/architecture/` still describe the ConfigMap model — worth refreshing (the decision is captured in ADR-058/059).

Depends on the ADR PR #552 (ADR-058 + ADR-059).


deploy/helm/platform/values.yaml
```diff
@@ -126,6 +126,13 @@ openshift:
     anyuidCapNet:
       enabled: false
 
+# -- Install the agent-platform.ai CRDs (Agent, Fork) that the controller
+# reconciles. These supersede the labeled-ConfigMap resource model (ADR-058)
+# and require cluster-admin at install time. Disable only when the CRDs are
+# managed out-of-band (e.g. a separate CRD release or GitOps).
+crds:
+  enabled: true
```
pilartomas, line 134: Shouldn't we just force it?

packages/api-server/src/apps/api-server/app.ts
```diff
@@ -139,6 +140,9 @@ export function startApiServerApp(deps: ApiServerAppDeps) {
 
   const k8sClient = createK8sClient(api, config.namespace);
   const agentsRepo = createAgentsRepository(k8sClient);
+  // Templates are file-mounted config loaded once at boot (ADR-058); shared
+  // across requests rather than re-read from K8s on each tRPC call.
+  const templatesRepo = createTemplatesRepository(config.agentTemplatesPath);
```
pilartomas, line 145: Unsure of loading at boot, will the controller reboot when templates are changed?

packages/api-server/src/apps/harness-api-server/app.ts
```diff
@@ -38,6 +39,8 @@ export function startHarnessApiServerApp(deps: HarnessApiServerAppDeps) {
   } = deps;
 
   const k8sClient = createK8sClient(api, config.namespace);
+  // Boot-loaded, file-mounted templates (ADR-058), shared across requests.
+  const templatesRepo = createTemplatesRepository(config.agentTemplatesPath);
```
pilartomas, line 43: ditto

packages/api-server/src/apps/api-server/app.ts
jezekra1: the templates are mounted to api-server, but good point, let's add the configmap hash into the deployment annotation

deploy/helm/platform/values.yaml
jezekra1: I think there is a real install-permission edge-case when you want to install CRDs separately from helm (in a more priviledged step or manually with kubectl apply)

packages/api-server/src/apps/api-server/app.ts
jezekra1: Good catch. Small clarification: it's the **api-server** (not the controller) that loads the mounted templates at boot — the controller only reads the Agent CR spec, never templates.

You're right that load-once-at-boot needs the pod to roll when templates change. Kubernetes doesn't restart a pod when a mounted ConfigMap changes, so I've added a `checksum/agent-templates` annotation on the api-server pod template (a hash of the rendered `agent-templates.yaml`). A `helm upgrade` that edits a template now rolls the api-server precisely when the templates change, so it reloads them instead of serving the stale boot-time set. Verified the checksum differs when the rendered template set differs. Fixed in a2268e5e.

pilartomas (approved):

PR #589, opened 2026-06-03
fix(dns): set ndots:1 on api-server pod

Tomas2D:
## Problem

Creating an OAuth MCP connection (`connections.create`, dynamic client registration) failed in the UI with:

```
No OAuth discovery metadata at https://<mcp-host>
```

`build-connection.ts` / `mcp-discovery.ts` are correct — the discovery genuinely failed. Root cause is **DNS, not code**.

The api-server pod's `resolv.conf` carries node-inherited search domains (a corporate VPN's domains leak into the lima/k3s node's `resolv.conf`) plus the default `options ndots:5`:

```
search default.svc.cluster.local svc.cluster.local cluster.local ibm.com lotus.com s81c.com ibmmodules.com ibmuc.com ihost.com
options ndots:5
```

A sub-5-dot MCP hostname (e.g. `garmin-mcp.example.dev`, 2 dots) is treated as *not* fully qualified, so the resolver tries the search suffixes **first**. A wildcard under one of those domains (`*.dev.<corp-domain>`) then captures the lookup and returns a dead IP. The fetch hangs, hits `AbortSignal.timeout(5000)`, is swallowed by `discoverMcpAuth`'s `catch {}`, and the function returns `null` → the misleading error above.

Verified: an **absolute** lookup (trailing dot, bypassing search domains) resolves to the correct public IP and the fetch returns `200`. `api.github.com` works today only because its suffixed forms `NXDOMAIN` and the resolver falls through to the real name.

## Fix

Add a `dnsConfig` with `ndots:1` to the **api-server Deployment** (`templates/apiserver/app.yaml`) so external hostnames (≥1 dot) resolve absolute-first. `dnsPolicy` stays `ClusterFirst`, so in-cluster service resolution is unaffected (single-label names still use search domains).

**Scoped to the api-server only.** An earlier revision also stamped `dnsConfig` onto agent/fork pods, but that was dead config: agent pods sit behind a deny-all egress NetworkPolicy (ADR-042) and never resolve or reach external MCP hosts themselves — all their external traffic goes through the paired gateway. MCP discovery and outbound MCP traffic originate from the api-server.

## Tests

- `helm:check:lint` + `helm:check:render` pass; confirmed `ndots:1` renders into the api-server Deployment.
- `controller:check` + `controller:test` pass (controller is untouched after the scope reduction).

## Verified on cluster

After deploy, the api-server pod shows `options ndots:1`, `<mcp-host>` resolves to its real (Cloudflare) IP instead of the hijacked one, and the fetch connects with no timeout.

### DIFF
--- a/deploy/helm/platform/templates/apiserver/app.yaml
+++ b/deploy/helm/platform/templates/apiserver/app.yaml
@@ -80,6 +80,17 @@ spec:
       {{- include "platform.imagePullSecrets" . | nindent 6 }}
       serviceAccountName: {{ include "platform.apiserver.serviceAccountName" . }}
       terminationGracePeriodSeconds: 45
+      # ndots:1 so external hostnames (e.g. an MCP server's domain during
+      # OAuth/DCR discovery in connections.create) resolve absolute-first.
+      # The default ndots:5 expands sub-5-dot FQDNs against node-inherited
+      # search domains first; a wildcard under one of those (seen with
+      # corporate VPN search domains leaking into the node's resolv.conf) then
+      # captures the lookup and discovery silently times out. dnsPolicy stays
+      # ClusterFirst — in-cluster service resolution is unaffected.
+      dnsConfig:
+        options:
+          - name: ndots
+            value: "1"
       containers:
         - name: apiserver
           image: "{{ .Values.apiServer.image.repository }}:{{ .Values.apiServer.image.tag | default .Chart.AppVersion }}"
--- a/deploy/helm/platform/values.yaml
+++ b/deploy/helm/platform/values.yaml
@@ -533,6 +533,18 @@ controller:
       priorityClassName: ""
       runtimeClassName: "" # e.g. "kata" for sandboxed runtimes
 
+      # Merged into the agent / fork pod's resolv.conf (dnsPolicy stays the
+      # ClusterFirst default). ndots:1 makes external MCP hostnames resolve
+      # absolute-first: the default ndots:5 expands a sub-5-dot FQDN against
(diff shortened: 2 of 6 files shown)


PetrBulanek:
**api-server half ✅ — agent half ⚠️ targets the wrong pod.**

Deployed this branch and traced the egress path:

- **api-server**: makes external OAuth discovery calls directly, so `ndots:1` is correct. Verified live (`options ndots:1`). 👍
- **agent** (`controller.agent.base.dnsConfig`): the goal ("fixes runtime MCP calls from agents") isn't reachable here. Agents have `HTTPS_PROXY`→gateway and **no `NO_PROXY`**, so all egress is tunneled to the gateway, and **Envoy on the gateway resolves the external hostname**. Verified on cluster: agent pod = `ndots:1`, but gateway pod has no `dnsConfig` → stays `ndots:5`, which is where the lookup actually happens. (Agent also opts out of the ambient mesh; the gateway is a mesh participant, so its DNS is further mediated by Istio.)

Repro confirms the *mechanism* is right (synthetic leaked search-domain + wildcard: `ndots:5` → dead IP + curl hang; `ndots:1` → `200`) — it's just applied to the wrong pod.

**Suggestion:** keep the api-server change; move the agent-side fix to the **gateway** pod (or an Envoy `typed_dns_resolver_config`) and verify with a real external MCP call from a running agent. No regressions otherwise.

PetrBulanek:
Seems like you resolved the latest push addresses my earlier review and the misplaced agent-pod `ndots` override is gone. :-)

PetrBulanek (approved):

PR #590, opened 2026-06-03
feat(cli): override preset credentials in connection connect

### LINKED ISSUE #554: Override preset credentials when creating a connection from the CLI
## Problem

`dam connection connect` creates a connection from a provider template. Often the OAuth client credentials are already known — either provided by an operator as a default, or inherited from another already-connected provider in the same family (e.g. once Google Drive is connected, Gmail can reuse the same client credentials). In those cases `connect` uses the known credentials automatically and doesn't ask — a good default.

But two gaps make it less capable than the web UI:

- **You can't supply your own credentials when a preset exists.** If you want to connect with your *own* OAuth client app instead of the operator/inherited default, the CLI gives you no way to do it — it always uses the preset. The web UI already shows the preset and lets you override it, so today this case forces you out of the CLI and into the browser.
- **Presets are applied silently.** When `connect` uses a preset (including a credential inherited from a sibling connection), it prints nothing — so you're left wondering why it never asked for a client ID or secret, with no signal that a preset was used or that it can be overridden.

## Proposed solution

- Let `connect` accept your own values for credentials that have a preset, overriding it — reaching parity with the web UI. When you don't supply them, the current automatic behavior is unchanged.
- When `connect` uses a preset for something it didn't ask about, print a short note saying so, with a pointer to how to override it.

### Scope

CLI only. No change to how connections, credentials, or credential families work on the server — this is about giving the CLI the same override and visibility the web UI already has.

### Out of scope

- **Naming the source of an inherited credential** (e.g. "inherited from your Google connection"). The CLI can't currently distinguish an operator default from a family inheritance, so the note will be generic. Surfacing that provenance is a separate, larger change.

### Dependencies

Follows #418 / #440 (the `connect` command). Related to #461 (credential families). Part of epic #329; parent #344.

PetrBulanek:
## What & why

Closes #554. Brings `dam connection connect` to parity with the web UI for **preset-backed credentials**.

Before, `connect` filtered `overridable` inputs out entirely: a preset (an operator default, or a client credential inherited from a sibling connection in the same family — e.g. Gmail reusing a Google Drive app) could neither be overridden nor reported. Two gaps, both closed here:

- **Override a preset.** A supplied typed flag (`--client-id`, `--client-secret`, `--app-slug`, `--host`, …) now overrides the preset; omit it and the server applies the preset as before (automatic path unchanged). `buildPayload` already threaded these through, so an overridden value reaches `connections.create` unchanged.
- **Make presets visible.** When a preset fills an input you weren't asked about, `connect` prints one generic stderr note — `Using preset values (Client ID, Client secret). Pass --client-id, --client-secret to use your own.` — naming the fields and override flags. Under `--json` the prose is suppressed and the same signal rides along as `presetsApplied: string[]`. Preset *values* are never printed (secrets aren't available; echoing a client id into CI logs is needless exposure).

The note is intentionally generic about provenance — the CLI can't distinguish an operator default from a family inheritance (per the issue's "Out of scope").

**CLI-only — no server change.** The server already accepts user-supplied overridable values and computes presets/family inheritance.

## Changes

- `packages/cli/src/modules/connection/commands/connect.ts` — `collectInputs` now scans all inputs (flag → override, missing required → prompt/error, overridable-without-flag → reported as applied preset); `emitSuccess` carries `presetsApplied` into `--json`; new `formatPresetNote` helper; a `--help` example and updated `--json` description.
- `docs/architecture/cli.md` — documented the behavior in the Template-id bullet and `--json` payload; added a Motivated-by entry for #554.

## Testing

- `mise run cli:check` — lint, tsc, format all green.
- `mise run cli:test` — green (the one failing case is the pre-existing live-Keycloak `auth login` integration test, unrelated: the cluster's custom JS-rendered Keycloak theme has no scrapeable login form).
- Drove the **real built `dist/bin.js`** against a fake tRPC server hosting the **real `appRouter`** + **real `connectionCreateInputSchema`**, with `listTemplates` produced by the **real `buildCatalog` + `templateToView`** (the actual operator-preset and family-inheritance code). Verified end-to-end: preset applied → creds omitted from payload + `presetsApplied` reported; flags supplied → values reach the payload, no note; family-inheritance human-mode note on stderr only.

Not exercised here: the genuine browser OAuth round-trip (interactive). Manual guide in #554.


packages/cli/src/modules/connection/commands/connect.ts
```diff
@@ -307,33 +323,42 @@ function validateName(name: string): void {
...
+      // A supplied flag overrides an `overridable` preset; for required/optional
+      // inputs it's just the user-typed value.
       values[input.name] = flagVal.trim();
     } else if (input.state === "required") {
       missing.push(input);
+    } else if (input.state === "overridable") {
+      // Left to the server-side preset (an operator default, or a credential
+      // inherited from a family sibling). Never prompted — reported instead.
+      presetsApplied.push(input);
     }
```
xjacka, line 355: 🟢 **Suggestion:** Consider an explicit `else if (input.state === "optional")` branch (or a closing comment) so a future reader doesn't mistake the silent drop for a bug — the previous filter made the omission obvious; now it isn't.

```suggestion
    } else if (input.state === "overridable") {
      // Left to the server-side preset (an operator default, or a credential
      // inherited from a family sibling). Never prompted — reported instead.
      presetsApplied.push(input);
    }
    // input.state === "optional" with no flag: silently skipped (server treats as unset).
```

xjacka (approved):
🛡️ **DAM** — Code Review @ `72a34c6`

## PR #590: feat(cli): override preset credentials in connection connect
**Author:** PetrBulanek | **Branch:** feat/554-cli-override-preset-credentials → main | **Changes:** +64 −17 (2 files)

### Summary
Brings `dam connection connect` to parity with the web UI for preset-backed credentials. Before, `overridable` inputs were filtered out entirely — a preset (operator default or family-sibling inheritance) could neither be overridden nor reported. This PR: (1) supplied flags now override the preset, (2) when a preset fills an input you didn't flag, the CLI prints a generic stderr note naming the field(s) + override flag(s), and (3) under `--json` the same signal rides along as `presetsApplied: string[]`. Closes #554; CLI-only — the server already supports it.

### Findings
- ✅ **Looks good:** Three-way classification in `collectInputs` (flagged → use value; required-missing → prompt/error; overridable-without-flag → report as preset) is the right shape — it falls out of the input `state` cleanly and the old required/optional filter is gone.
- ✅ **Looks good:** `flagListFor()` is extracted and shared by both the missing-input error and the preset note — the two surfaces can never disagree on flag spellings going forward. Good defensive refactor.
- ✅ **Looks good:** `formatPresetNote`'s comment block explicitly justifies why values aren't printed (echoing a client id into CI logs is needless exposure; secrets the CLI doesn't have anyway). That security rationale should live with the code; it does.
- ✅ **Looks good:** `presetsApplied` is omitted from the `--json` payload when the array is empty (consistent with the `authUrl?` pattern already there). `--help` and the `--json` description both advertise it as optional.
- ✅ **Looks good:** Architecture doc (`docs/architecture/cli.md`) updated in the same PR with a new `Motivated by:` entry for #554 and an expanded Template-id bullet describing overridable inputs and the `presetsApplied` field. Behavior drift closed in-PR.
- 🟢 **Suggestion:** `collectInputs` now silently drops `optional` inputs that weren't flagged (no prompt, no preset note). That matches prior behavior, but the new structure makes the absence of an `optional` branch more conspicuous. Consider an explicit `// optional: silently skipped — server treats as unset` comment in the `else if` chain so a future reader doesn't mistake the omission for a bug.

✅ No findings.

### Verdict
APPROVE — Well-scoped CLI parity feature; clean shared-helper refactor, security-conscious preset note, architecture doc updated in the same PR.

---

PR #602, opened 2026-06-04
fix(ui): make agent description editable and keep empty descriptions empty

matoushavlena:
## Problem

Two related issues with agent descriptions reported by a user:

1. **Can't update the description in Configure Agent.** The Configure Agent dialog exposed no description field at all, so the description was locked to whatever was set at creation.
2. **Empty descriptions get auto-populated with the template default** (e.g. `"Default Claude Code agent"`). Leaving the field blank at creation re-applied the template's description, and there was no way to clear it afterward.

## Root cause

The backend already fully supported updating `description` — the `update` tRPC procedure accepts it (`agentUpdateInputSchema`) and `agents-service.update()` persists it. The gaps were entirely in the UI:

- `ConfigureAgentDialog` / `configureAgentSchema` had no `description` field.
- On creation, `add-agent-dialog` sent `values.description.trim() || undefined`, coercing a cleared field to `undefined`, and `spec-assembly` then fell back to the template default via `opts.description ?? tmplSpec.description`.

## Changes

- **Configure Agent dialog** (`configure-agent-dialog.tsx`, `configure-agent-schema.ts`): add a Description field — schema, form default + baseline `reset()`, the input, and the update mutation — so the description can be edited or cleared after creation.
- **Create flow** (`add-agent-dialog.tsx`): send the trimmed description verbatim (including `""`) instead of coercing empty to `undefined`, so an explicitly-cleared description stays empty.
- **`spec-assembly.ts`**: keep the `??` fallback (with a one-line note on why it's not `||`) — a genuinely-omitted (`undefined`) description still falls back to the template default for non-UI API callers; an explicit `""` stays empty.

The template description is still surfaced as a sensible default — the create dialog pre-fills it when a template is picked, so the default is conveyed up front and the user can keep, edit, or clear it.

## Resulting behavior

| Scenario | Result |
|---|---|
| Pick template, keep pre-filled description | Template description saved (unchanged) |
| Pick template, **clear** the description | Stays empty ✅ |
| Custom image, leave description blank | Stays empty (unchanged) |
| API create that omits `description` entirely | Falls back to template default (unchanged) |
| Edit description later in Configure dialog | Editable, including clearing to empty ✅ |

## Testing

- `mise run ui:check` — format, lint, tsc clean
- `mise run api-server:check` — format, lint, tsc clean
- UI-only behavior change; no existing tests encoded the old creation fallback.

### DIFF
--- a/packages/ui/src/modules/agents/dialogs/configure-agent-dialog.tsx
+++ b/packages/ui/src/modules/agents/dialogs/configure-agent-dialog.tsx
@@ -109,6 +109,7 @@ export function ConfigureAgentDialog({
     mode: "onChange",
     defaultValues: {
       name: agent.name,
+      description: agent.description ?? "",
       assigned: [],
       assignedAppIds: [],
       envVars: userInitialEnv,
@@ -126,6 +127,7 @@ export function ConfigureAgentDialog({
     baselinedRef.current = true;
     reset({
       name: agent.name,
+      description: agent.description ?? "",
       assigned: [...accessQuery.data.secretIds].sort(),
       assignedAppIds: connectionsQuery.data.connections
         .map((c) => c.connectionId)
@@ -137,6 +139,7 @@ export function ConfigureAgentDialog({
     connectionsQuery.data,
     userInitialEnv,
     agent.name,
+    agent.description,
     reset,
   ]);
   const ready = baselinedRef.current;
@@ -301,7 +304,9 @@ export function ConfigureAgentDialog({
         });
       }
       const wantsAgentUpdate =
-        Boolean(dirtyFields.envVars) || Boolean(dirtyFields.name);
+        Boolean(dirtyFields.envVars) ||
+        Boolean(dirtyFields.name) ||
+        Boolean(dirtyFields.description);
       if (wantsAgentUpdate) {
         await updateAgent.mutateAsync({
           id: agentId,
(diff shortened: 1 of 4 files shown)


packages/ui/src/modules/agents/dialogs/configure-agent-dialog.tsx
```diff
@@ -419,6 +427,14 @@ export function ConfigureAgentDialog({
               {...register("name")}
             />
           </FormField>
+          <FormField label="Description" error={errors.description?.message}>
+            <input
+              className="w-full h-10 rounded-lg border-2 border-border-light bg-bg px-4 text-[14px] text-text outline-none transition-all focus:border-accent focus:shadow-[0_0_0_3px_var(--color-accent-glow)] placeholder:text-text-muted"
+              placeholder="Optional"
+              disabled={saving}
+              {...register("description")}
+            />
```
PetrBulanek, line 436: Nit, non-blocking: the new Description <input> duplicates the Name field's long Tailwind classes string verbatim. Worth extracting a small component. We have a generic `packages/ui/src/components/ui/input.tsx` but it's styled differently. @kapetr you refactored this area, right? Does it make sense to add it as some new variant or reuse the current one as is?

PetrBulanek (approved):

packages/ui/src/modules/agents/dialogs/configure-agent-dialog.tsx
kapetr: I see the inputs were not migrated on this modal yet - only the buttons were. I'd keep it as is, we'll tackle it separately, also the new design is coming for add/configure agent.

kapetr (approved):

PR #607, opened 2026-06-04
feat: background contribution apply with settlement tracking and degraded badge

jjeliga:
Applies agent contributions out-of-band from a single background worker, so reconfiguration is reliable and failures are visible instead of silent.

- One worker applies contributions to a healthy pod — a pod roll never half-lands work.
- A failed install retries in the background and surfaces a degraded badge instead of passing silently or wedging the agent.
- Scheduled one-shot triggers are delivered independently of contribution outcomes, and never fire twice.

Refs #168


kapetr (approved):
Approving — ADR-060 is sound and I verified the core: connect-time apply is gone (`hello` presence-only, single applier), the servable-pod gate defers correctly, the `last_settled_version` / `last_applied_version` split is distinct, the migration backfills additively to a fail-soft "settled", and the new `retry_idx` predicate matches `listRetryable`.

One edge case worth flagging (silent, low-frequency — fine as a follow-up):

**Possible silent loss of a one-shot trigger on the stale path.** `service.ts:31` returns on a stale `input.version` *before* `processEvents` (`:49`), so events aren't fired on that path. "stale" only happens when the agent is already caught up to `row.version` (the cursor can't exceed the row) — i.e. a *redundant* dispatch, typically a `hello`/reconnect. If a scheduler fire inserts a trigger in the worker's `getRow → build` window, the worker dispatches the now-stale `version` with that trigger still in `payload.events` (`state-builder.ts:56` → `outbox-repo.ts:136` `pendingEvents` is version-independent); the agent skips it here, and the `"stale"` branch (`worker-handler.ts:90`) marks it dispatched anyway → the trigger is consumed but never fired, with no retry.

It's a narrow race (a trigger firing ~simultaneously with a pod reconnect for the same agent), but it's silent and unrecoverable, and it diverges from ADR-060's "events apply in the same pass, independent of contribution outcome." Not blocking; worth a follow-up.

PR #625, opened 2026-06-04
fix(agents): trust platform CA via OS store, not SSL_CERT_FILE override

matoushavlena:
## Problem

Agent `git clone`/curl to **tunnelled (pass-through) hosts** failed with `SSL certificate ... unable to get local issuer certificate (20)`.

The controller set `SSL_CERT_FILE`/`GIT_SSL_CAINFO` to the bare 1-cert platform MITM CA. Those vars **replace** the CA bundle, so they dropped the ~146 public roots. That was fine in the original "gateway MITMs the host" model — the gateway's leaf is signed by the platform CA, which is exactly what that file contains. It breaks once a host is **allowed but not MITM'd**:

- **Credentialed hosts** (a granted connection, in the Envoy leaf SAN) → gateway MITMs, presents a platform-CA-signed cert → verifies against the bare CA. **Always worked** (e.g. `api.anthropic.com`).
- **Host-level allow-listed hosts** (e.g. `github.com` via the `trusted` preset, `method=* path=*`) → gateway L4-tunnels to the real upstream, which presents its **genuine public cert** → the platform-CA-only bundle can't verify it. **Failed.**

Reproduced on `claude-code-test` (only `api.anthropic.com` in its leaf SAN; `github.com` on `preset:trusted`): with `--cacert /etc/platform/ca/ca.crt`, `api.anthropic.com` → HTTP 400 (TLS OK), `github.com` → error 60.

## Fix

Install the per-cluster MITM CA into the **OS trust store** at pod start (`update-ca-trust`), so OS TLS clients (git, curl, Go, Python) trust both the public roots **and** the MITM CA — both the MITM'd and tunnelled cases verify. Drop `SSL_CERT_FILE`/`GIT_SSL_CAINFO`; keep the **additive** `NODE_EXTRA_CA_CERTS` for Node (which ignores the OS store). No regression to the previously-working MITM'd path.

- `platform-base`: new non-root entrypoint runs `update-ca-trust`; trust dirs chowned to UID 65532 so it works without root.
- `controller`: remove the two replace-semantics env vars from agent + fork pods.

Verified: CA mounted → bundle 146→147, anchor trusted, `update-ca-trust` exits 0; no CA mounted → clean no-op. Controller tests pass.

## Security

`/etc/pki/ca-trust/*` (+ the `/etc/pki/tls/certs` dir) are now writable by the agent UID. This is **not** a privilege escalation:

- A process already controls its own TLS trust (`curl -k`, `NODE_TLS_REJECT_UNAUTHORIZED=0`, its own `SSL_CERT_FILE`). Writing the OS store adds nothing it couldn't already do to its own connections.
- TLS trust is not the egress boundary — that's enforced **outside** the pod by NetworkPolicy + Envoy `ext_authz` (ADR-035). Trusting an extra CA can't reach a host the gateway/NP blocks, nor mint credentials (those live only in the gateway pod, ADR-038; agent has no SA token, ADR-033).
- Blast radius is one ephemeral pod: `/etc/pki` is on the container rootfs (not persisted), with `runAsNonRoot`, `allowPrivilegeEscalation: false`, `drop ALL` caps.
- Chosen over a root init container with a read-only-mounted store: marginally stricter but adds a privileged container for a boundary that isn't the real control; keeping the whole pod non-root is more valuable.
- chown scope is tight: recursive on the CA-input + generated-bundle dirs; `/etc/pki/tls/certs` is dir-only (its root-owned symlinks untouched).

## Out of scope

User-init scripts (`agentSpec.Init`) bypass the entrypoint and lack proxy+CA env — tracked in #626.

### DIFF
--- a/packages/controller/pkg/reconciler/fork_resources.go
+++ b/packages/controller/pkg/reconciler/fork_resources.go
@@ -99,9 +99,8 @@ func BuildForkAgentJob(
 		{Name: "HTTP_PROXY", Value: proxyAddr},
 		{Name: "https_proxy", Value: proxyAddr},
 		{Name: "http_proxy", Value: proxyAddr},
-		{Name: "SSL_CERT_FILE", Value: caCertPath},
+		// SSL_CERT_FILE / GIT_SSL_CAINFO left unset — see resources.go.
 		{Name: "NODE_EXTRA_CA_CERTS", Value: caCertPath},
-		{Name: "GIT_SSL_CAINFO", Value: caCertPath},
 		{Name: "NODE_USE_ENV_PROXY", Value: "1"},
 		{Name: "GIT_HTTP_PROXY_AUTHMETHOD", Value: "basic"},
 		{Name: "PLATFORM_AGENT_ID", Value: forkSpec.AgentName},
--- a/packages/controller/pkg/reconciler/resources.go
+++ b/packages/controller/pkg/reconciler/resources.go
@@ -150,9 +150,12 @@ func BuildAgentStatefulSet(name string, agentSpec *types.AgentSpec, cfg *config.
 		{Name: "HTTP_PROXY", Value: proxyAddr},
 		{Name: "https_proxy", Value: proxyAddr},
 		{Name: "http_proxy", Value: proxyAddr},
-		{Name: "SSL_CERT_FILE", Value: caCertPath},
+		// Node doesn't read the system trust store, so it gets the cluster CA
+		// through NODE_EXTRA_CA_CERTS (which adds to its built-in CAs). Other
+		// tools (git, curl, Go, Python) read the system store, where the agent
+		// entrypoint installs the CA — so we don't set SSL_CERT_FILE /
+		// GIT_SSL_CAINFO, which would replace and drop the public CAs.
 		{Name: "NODE_EXTRA_CA_CERTS", Value: caCertPath},
-		{Name: "GIT_SSL_CAINFO", Value: caCertPath},
 		{Name: "NODE_USE_ENV_PROXY", Value: "1"},
 		{Name: "GIT_HTTP_PROXY_AUTHMETHOD", Value: "basic"},
 		{Name: "PLATFORM_AGENT_ID", Value: name},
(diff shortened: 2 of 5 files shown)


jezekra1 (changes requested):
git clone works without problems, must be some other weirdness

Image

EDIT: the issue is with public repos without connected GitHub connection

matoushavlena:
## Reconciliation — resolved

Initial confusion was "git clone works fine" (Radek) vs "it fails" (repro). Both true — different agent configs. Tested on **freshly-rebuilt gateways** (so not a stale-pod/race):

| Agent | `github.com` rule | Cert github presents | clone w/ old config |
|---|---|---|---|
| has GitHub connection | `connection:github` | `issuer: CN=platform MITM CA` (Envoy leaf) | ✅ works |
| connection-less (`trusted` preset) | `preset:trusted` | `issuer: Sectigo …` (**real** GitHub cert) | ❌ `unable to get local issuer certificate` |

**Root cause:** `GIT_SSL_CAINFO` / `SSL_CERT_FILE` *replace* the CA bundle (unlike the additive `NODE_EXTRA_CA_CERTS`). Pointing them at the 1-cert platform CA dropped the ~146 public roots.

- **Credentialed host** → Envoy MITMs, leaf signed by platform CA → the platform-CA-only bundle verifies it → always worked (Radek's path; he tested a private repo with a gh connection).
- **Connection-less host via the `trusted` preset** → pure CONNECT tunnel, real public cert reaches the agent → needs the public roots, which were dropped → fails.

A tunnel correctly needs **no extra cert trust** (Radek's point) — the fix doesn't add one, it **restores the public roots**. The platform-CA install is only for the separate MITM path.

**Design intent confirmed:** connection-less agents are supposed to clone public GitHub via the `trusted` preset → the fix is correct.

**Fix:** install the platform CA into the OS trust store (so the system bundle = public roots ∪ platform CA), drop the replace-semantics `SSL_CERT_FILE`/`GIT_SSL_CAINFO`, keep `NODE_EXTRA_CA_CERTS`. Both paths verify; the previously-working MITM path is not regressed.

Not authz-related: the CONNECT tunnel succeeds (Envoy `200`, github is `preset:trusted`); only the client TLS verify fails.

Ready for review — leaving merge until @radek's connection-less repro tomorrow confirms.

jezekra1 (approved):

matoushavlena:
Tested, all looks good. Merging.

PR #630, opened 2026-06-04
fix(cluster): force-conflicts on dev helm apply, pin helm to v4.2.0

matoushavlena:
## Problem

`mise run cluster:install` fails with a server-side-apply conflict:

```
UPGRADE FAILED: conflict occurred while applying object default/platform-apiserver
... conflicts with "kubectl-patch" ... maxSurge / maxUnavailable
```

Helm pinned to `latest` floated up to **v4**, which applies **server-side** by default. Field drift on a chart-managed object (a stray `kubectl patch`/`rollout`, another controller) now becomes an ownership conflict that hard-fails the upgrade. Under v3's client-side merge this was silently reconciled.

## Change

- `deploy/tasks.toml`: add `--force-conflicts` to the dev `helm upgrade`. The chart is the source of truth in the install loop, so Helm reclaims drifted fields instead of wedging.
- `mise.toml`: pin `helm = "4.2.0"` (matches `mise.lock`) so floating `latest` can't silently change apply semantics again.

## Scope

Dev install loop only — not the prod upgrade path.

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -16,8 +16,6 @@ java = "21"
 maven = "latest"
 
 # Kubernetes
-# Pinned: Helm v4 applies server-side by default; floating "latest" silently
-# changes apply semantics. Bump deliberately.
 helm = "4.2.0"
 kubeconform = "latest"
 "lima" = "2"
--- a/mise.toml
+++ b/mise.toml
@@ -16,7 +16,9 @@ java = "21"
 maven = "latest"
 
 # Kubernetes
-helm = "latest"
+# Pinned: Helm v4 applies server-side by default; floating "latest" silently
+# changes apply semantics. Bump deliberately.
+helm = "4.2.0"
 kubeconform = "latest"
 "lima" = "2"
 
--- a/deploy/tasks.toml
+++ b/deploy/tasks.toml
@@ -427,7 +427,10 @@ echo "Installing/upgrading Platform chart..."
 # No --wait: the keycloak-provision post-install hook imports the platform realm
 # after deployments come up. With --wait, helm would block on readiness
 # before the hook runs.
-helm --kubeconfig="$KUBECONFIG" upgrade --install platform deploy/helm/platform -f deploy/helm/platform/values-local.yaml --timeout 15m ${HELM_VALUES:+-f "$HELM_VALUES"} "${HELM_EXTRA_ARGS[@]}"
+# --force-conflicts: Helm v4 applies server-side; field drift on a chart-managed
+# object (a stray kubectl patch/rollout, another controller) otherwise hard-fails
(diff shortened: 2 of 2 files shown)


mise.toml
```diff
@@ -16,7 +16,9 @@ java = "21"
 maven = "latest"
 
 # Kubernetes
-helm = "latest"
+# Pinned: Helm v4 applies server-side by default; floating "latest" silently
+# changes apply semantics. Bump deliberately.
```
matoushavlena:
```suggestion
```

jezekra1 (approved):

mise.toml
```diff
@@ -16,7 +16,7 @@ java = "21"
 maven = "latest"
 
 # Kubernetes
-helm = "latest"
+helm = "4.2.0"
```
xjacka, line 19: 🟢 **Suggestion:** Consider adding a one-line comment next to the pin explaining why it's pinned (so a future reader doesn't bump it back to `latest`):

```suggestion
helm = "4.2.0" # pinned: avoid silent SSA semantics change across Helm majors (see tasks.toml cluster:install)
```

xjacka (approved):
🛡️ **DAM** — Code Review @ `42f79f4`

## PR #630: fix(cluster): force-conflicts on dev helm apply, pin helm to v4.2.0
**Author:** matoushavlena | **Branch:** fix/helm-force-conflicts → main | **Changes:** +5 −2 (2 files)

### Summary
Fixes `mise run cluster:install` failing with server-side-apply ownership conflicts after Helm latest floated to v4 (which applies server-side by default). Adds `--force-conflicts` to the dev helm upgrade so the chart can reclaim drifted fields, and pins helm to `4.2.0` in `mise.toml` to stop the floating-`latest` regression. Scope is dev-only — the prod upgrade path is untouched.

### Findings
- ✅ **Looks good:** `--force-conflicts` is the correct flag for "chart is the source of truth" semantics in the dev install loop; the inline comment explains the why for future readers, which is exactly what's needed here.
- ✅ **Looks good:** Pinning `helm = "4.2.0"` instead of `latest` matches `mise.lock` and prevents future silent semantics changes from a floating tag — the right durability fix for an environment where a tool's default behavior changed across majors.
- ✅ **Looks good:** Scoped strictly to the dev `cluster:install` task — the production helm upgrade path is intentionally not touched, where reclaiming fields silently is a much riskier policy.
- 🟢 **Suggestion:** Consider mirroring the comment block on `mise.toml`'s helm line ("# pinned to v4.2.0: avoid silent SSA semantics changes across Helm majors") so the reason for the pin survives if `tasks.toml` is ever refactored. Not required.

### Verdict
APPROVE — Correct fix and durability hardening for the dev install loop; matches the existing maintainer approval from jezekra1.

---

Issue #283, opened 2026-05-19
Mesh traffic to keycloak breaks after the local cluster runs ~24h


matoushavlena:
## Problem

After the local k3s/lima cluster has been up for ~24h, anything that talks to keycloak through the mesh starts failing:

- UI login / token refresh stops working.
- Re-running `mise run cluster:install` hangs on the keycloak realm provisioning step with `Could not connect to keycloak … Connection reset`, even though the keycloak pod is healthy.

Keycloak itself receives nothing — connections die at the mesh hop. The misleading `Connection reset` makes it look like keycloak is broken when it isn't.

### Root cause

Workload mTLS cert handed to keycloak by the mesh expired and was not rotated. Data-plane proxy logs show `certificate expired` / `AlertReceived(CertificateExpired)` on the HBONE handshake. No auth policy is involved.

### Workaround

Restart the mesh data-plane pod in `istio-system` (or rollout-restart the affected workload) to force re-issuance.

## Proposed solution

- Fix rotation so the dev cluster stays healthy past 24h. Top suspects: lima VM clock skew on sleep/resume, or our bundled istio config.
- Document the symptom → workaround mapping in cluster debugging notes so the next developer doesn't chase the misleading error.

### Scope

Observed on the local dev cluster.

matoushavlena:
### Supporting logs

**`platform-keycloak-provision` init container** (`config-cli`):
```
INFO  KeycloakConfigRunner : Importing file 'file:/config/platform-realm.json'
INFO  KeycloakProvider     : Wait 120 seconds until http://platform-keycloak:8080 is available ...
ERROR KeycloakConfigRunner : Error during Keycloak import: Could not connect to keycloak in 120 seconds:
  RESTEASY004655: Unable to invoke request: java.net.SocketException: Connection reset
```

**`ztunnel-w8xns` in `istio-system`** — same failure from two different sources (apiserver and the provision job), both hitting keycloak:
```
error access connection complete
  src.workload="platform-keycloak-provision-j7qrd" src.namespace="default"
  dst.service="platform-keycloak.default.svc.cluster.local"
  dst.hbone_addr=10.42.0.17:8080 direction="outbound"
  error="io error: invalid peer certificate: certificate expired:
   verification time 1779175708 (UNIX),
   but certificate is not valid after 1779171279 (4429 seconds ago)"

warn access connection failed
  dst.addr=10.42.0.17:15008 direction="inbound"
[...]

PR #672, opened 2026-06-05
fix(ui): unstick restart pill on error

tomkis:
## Problem

Clicking Restart shows a RESTARTING pill that never clears when the agent ends in error. Only a full page reload (F5) fixes it.

## Cause

`useSyncRestartingAgents` re-runs its cleanup (`transitionRestartingAgents`) only when the agents query `data` *reference* changes. TanStack Query structural sharing keeps that reference stable when a poll returns a deep-equal payload. Restarting an agent that is already in `error` and fails the restart identically produces byte-identical poll payloads, so the cleanup never runs and the optimistic pill sticks. Reload "fixes" it only because it wipes the client store.

## Fix

Depend on `dataUpdatedAt`, which advances on every successful fetch regardless of structural sharing. The transition now runs each poll and drops the pill within one cycle.

## Notes

- `transitionRestartingAgents` was already correct (it drops on `error`); only the effect trigger was wrong.
- The restart failure in the repro was the istiod validating-webhook outage (#283), a separate infra issue, not part of this fix.

## Test

`ui:check:tsc`, `ui:check:lint`, `ui:test` (39) all pass.

### DIFF
--- a/packages/ui/src/modules/agents/hooks/use-restart-agent.ts
+++ b/packages/ui/src/modules/agents/hooks/use-restart-agent.ts
@@ -35,7 +35,7 @@ export function useRestartAgent() {
  * "Restarting" pill so stuck/resolved entries age out correctly.
  */
 export function useSyncRestartingAgents() {
-  const { data } = useAgents();
+  const { data, dataUpdatedAt } = useAgents();
   const setRestartingAgents = useStore((s) => s.setRestartingAgents);
 
   useEffect(() => {
@@ -43,5 +43,5 @@ export function useSyncRestartingAgents() {
     const current = useStore.getState().restartingAgents;
     const next = transitionRestartingAgents(current, data.list);
     if (next !== current) setRestartingAgents(next);
-  }, [data, setRestartingAgents]);
+  }, [data, dataUpdatedAt, setRestartingAgents]);
 }


xjacka (approved):
🛡️ **DAM** — Code Review @ `08dba74`

## PR #672: fix(ui): unstick restart pill on error
**Author:** tomkis | **Branch:** fix/restart-pill-stuck → main | **Changes:** +2 −2 (1 files)

### Summary
Fixes a stuck "Restarting" pill by depending on `dataUpdatedAt` in `useSyncRestartingAgents`. TanStack Query structural sharing keeps `data` reference-stable across polls when payloads are deep-equal, which kept the cleanup effect from firing after identical failed-restart polls; `dataUpdatedAt` advances on every successful fetch, so the transition now runs each poll.

### Findings
- ✅ **Looks good:** Root-cause analysis is correct; `dataUpdatedAt` is the standard escape hatch for this exact TanStack Query structural-sharing trap.
- ✅ **Looks good:** Keeping `data` in deps alongside `dataUpdatedAt` preserves the lint contract (`data.list` is read inside the effect) and costs nothing — the effect would run on either dep change anyway.
- ✅ **Looks good:** `useStore.getState()` inside the effect rather than subscribing to `restartingAgents` keeps the effect from re-running on store mutations it caused itself, preserving the convergent fixed-point behavior of `transitionRestartingAgents`.

✅ No findings.

### Verdict
APPROVE — Minimal, correct fix targeting the right root cause; no architectural or code-quality concerns.

---

PR #677, opened 2026-06-05
fix(controller): answer np-gate probe via gateway health_check filter (#675)

### LINKED ISSUE #675: New agents don't respond until the user approves a network request to an internal address
## Problem

Creating a new agent no longer works out of the box. The agent stays silent and does nothing until the user is shown a network-approval prompt asking them to allow a request to an internal cluster address — an opaque IP such as `172.30.212.77`. Only after the user manually approves that request does the agent come to life and respond.

This is a regression. New agents used to start and respond immediately, without any approval step.

It matters because:
- **Every new agent is blocked** on a manual approval before it will do anything, so the first-run experience appears broken.
- **The user can't meaningfully evaluate what they're approving.** The prompt points at an internal address that isn't a destination the user chose or recognizes — it's part of how the platform talks to itself, not an external service the user is reaching out to.
- It erodes trust in the approval mechanism: prompts that should be reserved for real external destinations are now firing for routine internal traffic.

## Proposed solution

New agents should start and respond immediately, with no approval step for the platform's own internal traffic.

The platform's internal communication that an agent depends on at startup should always be allowed automatically and never surface as a network-approval prompt. Approval prompts should appear only for genuine external destinations the user actually intends to control.

jezekra1:
## Problem

New agents don't respond out of the box on deployments using `npGateInit` egress lockdown (OpenShift Sandboxed Containers): the first traffic surfaces a HITL network-approval prompt for an opaque internal address (e.g. `172.30.x.x`), which is the **paired gateway Service's own ClusterIP**. Fixes #675.

## Root cause — a regression from #434

The `npGateInit` egress-lockdown init container probes the paired gateway to confirm the agent egress NetworkPolicy has converged. #434 (hardened-image migration) switched that probe from `nc -z` to `curl http://$GATEWAY_IP:$ENVOY_PORT`.

- `nc -z` only completes the TCP handshake and sends **no application bytes**, so the gateway's Envoy HCM never decodes a request and ext_authz never runs.
- `curl http://…` sends a real `GET /`. With no proxy env, it lands directly on the gateway Envoy's outer HCM and matches the **plain-HTTP fallthrough route** — the one route where the `ext_authz` filter is *not* disabled (the CONNECT and harness routes all opt out). So the gate fires with `host = <gatewayIP>`, finds no egress rule, and opens a HITL hold → the spurious internal-IP prompt.

Invisible on k3s dev/e2e because those run `iptablesInit` (no HTTP probe); only `npGateInit` deployments hit it.

## Fix

Add an `envoy.filters.http.health_check` filter on the gateway's outer HCM, **ordered before `ext_authz`**, `pass_through_mode: false`, matched on an exact, deliberately-namespaced path `/__platform_healthz`:

- It answers the path locally with `200` and short-circuits the filter chain before the gate — so the probe never creates a hold.
- `pass_through_mode: false` means it **never forwards upstream**, so it cannot become an egress bypass.
- The path is `__`-namespaced so it can't collide with real plain-HTTP egress (HTTPS egress arrives as CONNECT, whose `:path` is the authority, so it never matches).

The np-gate probe now asserts a `200` from that path for the *allowed* (gateway) target; the kube-apiserver *denied* target stays a TCP-handshake (`%{time_connect}`) check. The egress gate is unchanged for every other path. The bootstrap template rev is bumped so existing gateway pods roll.

## Verification (live, against a rolled gateway)

| request | result | ext_authz hold? |
|---|---|---|
| `GET /__platform_healthz` | `200`, ~5ms | none (no log line) |
| `GET /` (control) | held 4s, `000` | `egress.hold` for gateway IP |
| spoofed `Host:` / proxied to real host on the path | local `200`, `size_download=0` | none — never forwarded |
| `/__platform_healthz/`, `/__platform_healthzX`, `/healthz` | held, `000` | gated — carve-out is exact-match only |

So the probe works, the gate stays fully active for everything except the one local-reply path, and the path never forwards (no exfil channel).

## Follow-up (not in this PR)

The gateway pod has no readiness/liveness probe today. A dedicated gateway health listener for real kubelet probes is worth doing as a separate change (it carries the Istio-ambient + NetworkPolicy probe-wiring question), tracked separately.

### DIFF
--- a/packages/controller/pkg/reconciler/np_gate_init.go
+++ b/packages/controller/pkg/reconciler/np_gate_init.go
@@ -21,16 +21,9 @@ const npGateUser int64 = 65532
 // enforced — used on runtimes where the in-pod iptables init can't run
 // (Kata/CoCo guest kernels without netfilter).
 //
-// It runs a small shell+curl probe (from a hardened curl image) in a loop:
-// the kube-apiserver (kubelet-injected KUBERNETES_SERVICE_HOST/PORT) must be
-// UNREACHABLE — the agent egress NP DROPs it — while the paired gateway must
-// be REACHABLE. The apiserver is the chosen "denied" target because it's the
-// only one always actively listening AND always silently DROPped by the NP;
-// other targets admit TCP-RST false positives. Reachability is read from
-// curl's %{time_connect}: a completed TCP handshake records a non-zero time,
-// a silent DROP yields 0.000000. The http:// scheme only drives the connect —
-// the response is discarded — so it works against any TCP port regardless of
-// the protocol it actually speaks.
+// Probe: the kube-apiserver (kubelet-injected env) must stay DROPped, detected
+// by TCP handshake since it answers when reachable; the gateway must 200 on the
+// health path, which the health_check filter serves before ext_authz (#675).
 //
 // Fail-closed: timeout → exit 1 → pod stays in Init:CrashLoopBackOff.
 //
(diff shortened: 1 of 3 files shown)


packages/controller/pkg/reconciler/np_gate_init.go
```diff
@@ -52,32 +45,38 @@ func buildNPGateInitContainer(cfg *config.Config, gatewayClusterIP string) *core
...
+# Denied-target probe: handshake completed iff %{time_connect} != 0 (the
+# apiserver answers TLS/non-200 when reachable, so status isn't the signal).
+reachable() {
     tc=$(curl -s -o /dev/null --connect-timeout 2 -m 3 -w '%{time_connect}' "http://$1:$2" 2>/dev/null)
     [ -n "$tc" ] && [ "$tc" != "0.000000" ]
 }
+# Gateway probe: 200 on the health path. The health_check filter answers it
+# before ext_authz, so this never trips the egress gate (#675).
+gateway_ready() {
+    code=$(curl -s -o /dev/null --connect-timeout 2 -m 3 -w '%{http_code}' "http://${GATEWAY_IP}:${ENVOY_PORT}${HEALTH_PATH}" 2>/dev/null)
```
xjacka, line 58: 🟢 **Suggestion:** `${HEALTH_PATH}` is controller-set from a Go constant (`platformGatewayHealthPath` in `envoy.go`) and cannot be user-supplied — a brief inline note would let future reviewers dismiss the shell-injection concern without tracing the provenance back to the Go source.

xjacka (approved):
🛡️ **DAM** — ✅ Code Review @ `92c8980`

## PR #677: fix(controller): answer np-gate probe via gateway health_check filter (#675)
**Author:** jezekra1 | **Branch:** fix/np-gate-health-probe → main | **Changes:** +49 −26 (3 files)

### Summary
Fixes a regression introduced in #434 where switching the np-gate-init probe from `nc -z` to `curl http://` caused the probe to trip the gateway's `ext_authz` filter — triggering spurious HITL holds on the gateway's own ClusterIP. The fix adds an Envoy `health_check` filter ordered before `ext_authz` on the gateway's outer HCM, answering `/__platform_healthz` with a local 200 and never forwarding; the probe script is updated to assert HTTP 200 from the health path rather than a TCP handshake.

### Findings
- ✅ **Looks good:** `pass_through_mode: false` correctly prevents the health path from becoming an egress bypass — the filter never forwards upstream.
- ✅ **Looks good:** `__`-namespaced path avoids collision with real egress traffic; CONNECT-proxied HTTPS uses the authority as `:path` so it can never match a `/`-prefixed path.
- ✅ **Looks good:** `platformGatewayHealthPath` constant is shared between `envoy.go` and `np_gate_init.go`, keeping the Envoy filter path and the probe path mechanically in sync.
- ✅ **Looks good:** Tests cover both probe strategies (`%{time_connect}` for denied target, `%{http_code}` for allowed), plus the `HEALTH_PATH` env var linkage assertion.
- ✅ **Looks good:** Template rev bumped to `v11-gateway-health-probe` to trigger rolling restarts of existing gateway pods.
- 🟢 **Suggestion:** `gateway_ready()` inlines `${HEALTH_PATH}` directly into the curl URL. Since `HEALTH_PATH` is controller-set from a Go constant (not user-supplied), a brief inline comment clarifying this would help future readers dismiss the shell-injection concern without tracing back to `envoy.go`. (`packages/controller/pkg/reconciler/np_gate_init.go:58`)

⚠️ **Possible drift (Check 1) — `docs/architecture/platform-topology.md` line 84:** The gateway description reads *"Terminates the agent's egress TLS, injects credentials on the wire, and gates each credentialed request through the api-server's ext_authz handler."* The PR adds a `health_check` filter that answers `/__platform_healthz` **before** ext_authz with a local 200 and never forwards. Whether this pre-ext_authz carve-out warrants a one-sentence addition is a human judgement call — it is a new observable behavior on the gateway listener, but also a narrow implementation detail of the np-gate probe rather than a subsystem responsibility change.

No `Last verified:` staleness applies — no architecture pages are edited in this diff.

### Verdict
APPROVE — the fix correctly identifies and addresses the root cause with sound security design (filter ordered before `ext_authz`, `pass_through_mode: false`, namespaced path, shared constant for sync); no correctness or security concerns.

---

PR #681, opened 2026-06-05
refactor(skills): make the runtime manifest own skill install paths

### LINKED ISSUE #680: Make each agent's runtime manifest the single source of truth for its skill install path
## Problem

An agent's skill install path is declared in two places today: the platform Helm values (per agent template) and the agent image's runtime manifest. Since the move to the runtime channel, only the runtime manifest actually decides where installed skills land — the Helm value no longer influences it. But the Helm value hasn't been removed: it's still read by the server when it lists and reconciles what's installed, so it looks load-bearing while being inert on the install path.

This is a footgun. An operator who edits the Helm skill-path setting expects it to change where skills go; it silently does nothing. And keeping the same fact in two places invites the two to drift apart — the listing side and the install side can disagree about where an agent's skills live.

## Proposed solution

Make the agent image's runtime manifest the **single source of truth** for where each harness reads and writes skills. Each agent declares its own harness-native skill path in its own manifest, overriding the shared cross-harness default. Remove the now-inert Helm skill-path setting and the server-side reads that still reference it, so an agent's skill location is defined in exactly one place.

### Scope

Consolidation only — externally observable behavior is unchanged. Skills continue to install to and surface from the same locations per harness; this removes the duplicate/dead configuration path, it doesn't change where skills end up.

**Safe removal, no migration.** The agent custom resource uses a structural schema that prunes unknown fields rather than rejecting them, and every consuming layer tolerates a stale field. Existing agents that still carry the old skill-path field stay valid; the field is simply dropped the next time the agent is rewritten. No data migration or staged rollout is needed.

### Out of scope

- Changes to the skill source catalog, install/publish flows, or on-disk authority for installed skills.

jezekra1:
Closes #680.

## Problem

An agent's skill install path was declared in two places: the Helm values (per agent template) and the agent image's runtime manifest. Since the runtime channel landed (ADR-052), only the manifest's `skill-ref` driver decides where skills install — the Helm value no longer influenced it, but it was still read by the api-server's listing/reconcile, so it looked load-bearing while being inert on the install path. Editing it silently did nothing, and the duplicated fact could drift.

## Change

The runtime manifest is the **single source of truth** for where each harness reads/writes skills.

- **agent-runtime resolves its own paths** — `listLocal` / `readLocal` / `publish` paths come from the loaded manifest; `skillPaths` is dropped from those wire schemas. `install` / `uninstall` keep theirs (the skill-install driver supplies them from the same manifest). The api-server no longer resolves or passes paths.
- **Full removal** — `skillPaths` is gone from the Agent CRD / Go type (deepcopy + CRD + `crd-types.gen.ts` regenerated), `TemplateSpec`, spec-assembly, the api-server read side (`resolveSkillPaths` / `DEFAULT_SKILL_PATHS` deleted), and all Helm values + the `agent-templates.yaml` render block.
- **No per-agent manifest override needed** — every image inherits platform-base's `.agents/skills`, and each per-agent Dockerfile already symlinks its harness-native dir (`.claude/skills`, `.pi/agent/skills`, `.bob/skills`) onto that shared store (ADR-023). The symlink does the harness-specific routing, so install and read both resolve to the same directory regardless.
- **Docs** — `docs/architecture/skills.md` updated to describe manifest-owned paths + the symlink store.

## Safety

The Agent CR uses a structural schema that **prunes** unknown fields rather than rejecting them, and every consumer (Go controller, api-server cast, `.passthrough()` template Zod) tolerates a stale field. Existing agents stay valid; the old field is dropped the next time an agent is rewritten. No migration or staged rollout.

## Verification

- `mise run check` — clean (tsc / lint / format / Go build + staticcheck).
- Tests — controller, agent-runtime (98), api-server (191) pass.
- `mise run helm:check:lint` + `helm:check:render` — pass; rendered chart contains zero `skillPaths`.

ADR-030 / ADR-052 still reference `skillPaths`, left unedited per the ADR-immutability convention.


packages/agent-runtime/src/server.ts
```diff
@@ -41,10 +45,29 @@ const workDir = config.PLATFORM_DEV
...
 
+// skill-ref driver paths from the manifest, $HOME expanded against home.
+function skillRefPaths(manifest: RuntimeManifest, home: string): string[] {
+  const binding = manifest.drivers["skill-ref"] as
+    | { paths?: unknown }
+    | undefined;
+  const raw = Array.isArray(binding?.paths) ? binding.paths : [];
+  return raw
+    .filter((p): p is string => typeof p === "string")
+    .map((p) => p.replace(/\$HOME\b/g, home).replace(/\$\{HOME\}/g, home));
```
PetrBulanek: There is already an `expandHome` helper in 3 different files that does exactly this. I guess we should refactor to a generic helper and reuse everywhere.

PetrBulanek (approved):

packages/agent-runtime/src/server.ts
jezekra1: Done — extracted to `core/expand-home.ts` (next to the other `core/` primitives) and reused in all four call sites: the file / mcp-entry / skill-install drivers and `server.ts`. 85683e0b

PR #682, opened 2026-06-05
feat: env injection without pod rolls

jjeliga:
Summary:

Moves credential-placeholder env off the agent pod spec onto the runtime channel, so grant/credential changes no longer roll the pod (ADR: DRAFT-runtime-env-injection).

- Env is a channel contribution, merged into the harness process at spawn instead of baked into the pod. Mid-life changes respawn the harness (resumable per ADR-055), not the pod.
- Agent pod template is grant-independent — leaf cert always issued, ca.crt mounted unconditionally — so the first/last credential no longer rolls the agent. Forks keep baked env.
- Single-shot create settles initial grants into the spec before first render (gateway chains rendered once, creds on the first snapshot).
- Cold-boot startup fixes: harness spawn is gated until env is delivered (buffering client frames) so the first turn isn't credential-less; and a hello-triggered delivery fast-retries on the queue backoff instead of waiting for the ~60s sweep.
- Docs: new draft ADR; updated connections.md / persistence.md / skills.md.
 
Refs: #168

### DIFF
--- a/packages/api-server/src/modules/secrets/services/secrets-service.ts
+++ b/packages/api-server/src/modules/secrets/services/secrets-service.ts
@@ -495,6 +495,8 @@ export function createSecretsService(deps: {
     },
 
     async delete(id) {
+      // Capture grant holders before deletion so we can re-deliver after.
+      const granted = await deps.grants.listAgentsGrantedSecret(id);
       // Twins first; primary last on success so retry-after-failure is clean.
       const allSecrets = await deps.k8sPort.listSecrets();
       const twinIds = allSecrets
@@ -513,6 +515,9 @@ export function createSecretsService(deps: {
         result: "success",
         detail: { twins: twinIds.length },
       });
+      // Re-deliver so the deleted secret's placeholder drops from each granted
+      // agent's env on the next snapshot (the env-source omits missing secrets).
+      await Promise.all(granted.map((g) => deliverToAgent(g.agentId)));
     },
 
     async getAgentAccess(agentId: string) {
--- a/packages/api-server/src/modules/secrets/services/secrets-service.ts
+++ b/packages/api-server/src/modules/secrets/services/secrets-service.ts
@@ -1,5 +1,6 @@
-import { createHash, randomUUID } from "node:crypto";
+import { randomUUID } from "node:crypto";
 import { TRPCError } from "@trpc/server";
+import type { RuntimeMutator } from "../../runtime-delivery/index.js";
 import {
   DEFAULT_ENV_PLACEHOLDER,
   isProviderPresetType,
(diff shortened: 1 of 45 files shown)


packages/api-server/src/modules/secrets/services/secrets-service.ts
```diff
@@ -476,17 +490,8 @@ export function createSecretsService(deps: {
...
-          );
-          const hash = combinedSecretsRev(grantedForAgent);
-          await deps.grants.bumpSecretsRev(g.agentId, hash);
-        }),
-      );
+      // Re-deliver to each granted agent; the state-builder reads the new env-mappings live.
+      await Promise.all(granted.map((g) => deliverToAgent(g.agentId)));
     },
 
     async delete(id) {
```
pilartomas, line 497: **`delete()` doesn't re-deliver — granted agents keep a stale credential placeholder**

`update()` (just above) and `setAgentAccess()` both call `deliverToAgent` after mutating, so the state-builder rebuilds the snapshot and `secret-env-source` drops the now-missing entry. `delete()` removes the K8s Secret(s) but never delivers — and it doesn't ungrant, so the deleted id stays in each agent's `grantedSecretIds`.

Net effect: the deleted secret's placeholder lingers in the harness process env until some *unrelated* mutation bumps that agent's outbox version. There's no self-healing path — the cron sweep only re-enqueues retryable rows, and `hello` only re-enqueues when `version > lastAppliedVersion` (both false after a settled apply). The credential boundary is unaffected (the gateway SDS entry is gone, so the stale placeholder injects nothing), but the agent's env is silently out of date — most visible on long-lived/interactive harnesses.

This is reachable from the normal UI path: `use-provider-actions.ts` calls `deleteSecret` directly, with no preceding ungrant (the confirm dialog even acknowledges agents are still using the provider at delete time).

Suggest mirroring `update()` — collect the granted agents before deleting, then deliver:

```ts
async delete(id) {
  const granted = await deps.grants.listAgentsGrantedSecret(id);
  // ... existing twin + primary deletion + securityLog ...
  await Promise.all(granted.map((g) => deliverToAgent(g.agentId)));
},
```

The env-source already drops the missing secret on rebuild — this just supplies the trigger. (Alternatively, require callers to ungrant before delete and enforce it at the router.)

pilartomas (approved):
Approving — solid PR. The design faithfully implements the DRAFT ADR, and I traced the two highest-risk halves (the agent-runtime cold-boot gate / harness respawn and the controller pod-template decoupling) against the code; both are correct.

Confirmed:
- **Credential boundary unchanged.** Only placeholders (sentinels like `dummy-placeholder`) ride the runtime channel — `secret-env-source` projects `m.placeholder` only, never secret bytes. Real credentials stay in the K8s Secret → gateway pod → Envoy on-the-wire injection (ADR-005/033/038). The agent pod holds zero secret material, as before.
- **Roll behavior is as intended.** The agent pod template is grant-independent (no credential env, `ca.crt` mounted unconditionally by stable Secret name, leaf always issued), so credential/grant changes no longer roll it — the harness respawns to absorb new placeholder env instead. The gateway still self-rolls on credential-set changes via `envoy-secrets-rev`, which is out of scope here.
- Agent/fork split is consistent (forks keep baked env, `alwaysIssue=false`); removed pod env + `gh-token-available` annotation have no remaining consumers.

Non-blocking follow-ups (not gating this approval):
- One real correctness gap left as an inline comment: `secrets-service.delete()` doesn't re-deliver, so a deleted-but-still-granted secret leaves a stale placeholder in the harness env with no self-healing path (not a boundary issue — the gateway SDS entry is gone, so it injects nothing).
- Recommend adding unit coverage for the new high-risk surface (cold-boot gate / `refreshEnv` / recycle in `acp-runtime.ts`, `runtime-env.ts`/`env-plugin.ts`, and the `worker-handler` `retryUntilReady` fast-retry) — the deps were shaped "exposed for tests," so it's cheap.
- Minor doc-drift (connections.md "not yet implemented" banner; stale `resources.go` roll-rev comment).

PR #693, opened 2026-06-08
feat(controller): warm PVC pool for instant agent workspace provisioning

### LINKED ISSUE #692: Pre-provision a buffer of workspace storage so new agents start without a wait
## Problem

The first time an agent starts, the platform has to provision fresh workspace storage for it before the agent can come up. In production this can take anywhere from tens of seconds to several minutes.

For the user, this is dead time: they create or start an agent and then sit waiting for it to become usable, with no work happening during the delay. It's the slowest and most visible part of bringing a new agent online, and it makes the platform feel sluggish exactly at the first-impression moment. The wait is also unpredictable — sometimes quick, sometimes minutes — so users can't tell whether something is broken or just slow.

## Proposed solution

Keep a small standing buffer of workspace storage that has already been provisioned and is sitting ready. When a new agent needs storage, it claims one from the buffer immediately instead of waiting for fresh provisioning. The buffer refills in the background, so the slow provisioning step happens ahead of demand rather than while a user waits.

From the user's point of view, starting a new agent becomes near-instant in the common case, and the unpredictable multi-minute wait disappears.

### Scope

- Covers the first-start storage provisioning delay for newly created agents.
- The buffer size should be tunable so an operator can balance readiness against the cost of holding idle storage.

### Out of scope

- Mounting user-supplied external storage (tracked separately in #359).
- Speeding up later wakes from hibernation, which already reuse the agent's existing storage and don't hit this delay.

pilartomas:
Closes #692.

## Problem

The first time an agent starts, its workspace PVC is dynamically provisioned by the CSI driver — tens of seconds to minutes in production. That's dead time the user waits through, and the slowest, most visible step of bringing a new agent online.

## Solution

A controller-managed, **leader-only** background buffer of pre-provisioned, already-`Bound` spare workspace PVCs, organized into **per-size pools**. A newly created agent claims a matching spare at create time and mounts it by name; the buffer refills in the background. **Disabled by default** — a default deployment behaves exactly as before.

### How it works
- **Pool manager** (`warmpool.go`): per configured size, tops up to target, counts in-flight (`Pending`) spares so it doesn't over-provision, reclaims stuck (`Pending` past `maxProvisioningTime`) and `Lost` spares, and trims excess oldest-first.
- **Claim** (`agent_reconciler.go`): atomically relabels a `Bound` spare to the agent (resourceVersion-checked, tries the next on conflict) and mounts it by name instead of via a `volumeClaimTemplate`. No match / empty pool → falls back to today's dynamic provisioning, so **creation never blocks**.
- **Stability**: the claim is reconstructed from the **live StatefulSet** on every reconcile (intersected with the spec's currently-persisted mounts), so it's reproduced byte-identically across wake/hibernate and never degrades to a mount with no backing volume.
- **Pre-warming** requires `volumeBindingMode: Immediate` (the agents' `WaitForFirstConsumer` class would leave spares empty), so the pool uses a separate class; a bundled NFS Immediate variant is added for dev.
- **GC safety**: unclaimed spares carry no agent label, so the orphan-PVC sweep skips them; once claimed they're ordinary agent PVCs (reclaimed on delete, never recycled).

## Design decisions (recorded in ADR-061)

This went through a design grill; the notable resolutions:
- **Forks** previously located a parent's workspace PVC by reconstructing `<mount>-<agent>-0`. A pool-claimed PVC has a generated name, which would have **broken forks** — so workspace PVCs are now addressed by `(agent, mount)` label, with a legacy name fallback for pre-label agents (backward compatible).
- The pool **inherits `agent.base.accessMode`** (a claimed spare *is* the workspace; it must be RWX for forks) rather than a separate, driftable knob.
- Claimed volumes are **destroyed on delete, never recycled** (a used workspace is adversarial input).
- Stale-`Pending` reclaim is an operator-tunable `maxProvisioningTime` (default 30m), decoupled from the replenish interval, so a slow-but-healthy provision is never reaped.
- Claim happens **at create** for v1 (claim-at-wake is a clean follow-up).

## Review

Ran an adversarial multi-lens review of the diff; it surfaced three real issues, all fixed with regression tests:
- **(blocker)** an out-of-band-deleted claimed PVC could corrupt the pod template → now reconstruct from the live STS (recoverable missing-PVC state instead).
- **(major)** removing a persisted mount from a claimed agent left a dangling volume → claims are now intersected with current spec mounts.
- **(minor)** a stale test comment.

## Testing
- `mise run controller:check` (fmt/vet/staticcheck/build) ✅ and `mise run controller:test` ✅ — new unit tests cover pool replenish/GC, claim+relabel, fallback, idempotent re-reconcile, conflict-retry, STS-derived reconstruction, removed-mount drop, orphan-sweep safety, fork label-resolution + fallback, and config validation.
- `mise run helm:check:lint` + `helm:check:render` (kubeconform) ✅ — Immediate StorageClass and `WARM_POOL` env render when enabled; nothing warm-pool renders by default.

## Not done
- Live-cluster manual verification (enable the pool on a warm cluster, confirm spares Bind and an agent claims one instantly) — needs a running cluster.
- Pool-health metrics deferred (controller has no metrics endpoint yet); observability is structured logs for now.


packages/controller/pkg/reconciler/agent_reconciler.go
```diff
@@ -328,6 +339,153 @@ func (r *AgentReconciler) deletePVCs(ctx context.Context, agentName string) {
 	}
 }
 
+// resolveWorkspaceClaims decides which persisted mounts of an agent are backed
```
jezekra1: can we trim down the comments? This is becoming a book  :-D

docs/adrs/061-warm-pvc-pool.md
```diff
@@ -0,0 +1,36 @@
...
+- The claim is decided **once, at first create**, and reproduced on every later reconcile from the agent's already-created StatefulSet (the frozen record of which mounts are claimed vs dynamically provisioned), so the agent renders identically across hibernate/wake — a claim is never introduced into, or dropped from, an already-created agent.
+- A **workspace volume is identified by label** (owning agent + mount), not by a reconstructed name. A claimed spare keeps the pool's generated name, so consumers that previously derived a volume's name by convention — forks co-mounting the parent's workspace — now resolve by label, with a legacy name-convention fallback for volumes created before the label existed.
```
jezekra1, line 20:
```
 forks co-mounting the parent's workspace — now resolve by label, with a legacy name-convention fallback for volumes created before the label existed.
```
is there a possibility of stealing foreign pvc by naming the agent a certain way?

jezekra1 (changes requested):

packages/controller/pkg/reconciler/agent_reconciler.go
pilartomas: Done 😄 — trimmed the worst offenders in 86d7e2a7: `resolveWorkspaceClaims`/`claimSpare`/`listClaimedPoolPVCs`, the label-const block, `applyPoolClaims`, and the wiring comment. Kept only the load-bearing *why* (the live-STS reconstruction invariant and the orphan-sweep label rule).

docs/adrs/061-warm-pvc-pool.md
pilartomas: Good instinct, but no — there's no theft-by-naming vector.

PVC ownership is the controller-set `agent-platform.ai/agent` label, which always carries the PVC's true owning agent and is never user-settable. Two paths matter:

- **Claiming a spare:** `claimSpare` only ever selects *ownerless* spares (`pool=<size>,pool-available=true`) and relabels one. It never touches a PVC already labeled to another agent, so you can't claim someone else's volume.
- **The fork fallback name** `<mount>-<parent>-0` embeds the parent agent's name, which is unique (K8s) and DNS-1123-constrained (no `,`/`=` selector-injection chars). To resolve to a victim's PVC you'd need `forkSpec.AgentName == victim` — i.e. authorization to fork that agent, which **already** grants workspace co-mount by design (ADR-027). So the pool adds no vector beyond the existing fork model.

Label selectors use exact names (no wildcards), so a crafted name can't broaden a match either. Happy to add a one-line invariant note to the ADR if you'd like it captured there.

packages/controller/pkg/reconciler/warmpool.go
```diff
@@ -0,0 +1,283 @@
...
+
+	// Trim Bound excess (target lowered, or drift), oldest first.
+	if len(bound) > target {
+		sort.Slice(bound, func(i, j int) bool {
+			return bound[i].CreationTimestamp.Time.Before(bound[j].CreationTimestamp.Time)
+		})
+		for _, p := range bound[:len(bound)-target] {
+			if m.deletePVC(ctx, p.Name) == nil {
+				slog.Info("warm pool: trimmed excess spare", "pool", poolKey, "pvc", p.Name)
+			}
```
jezekra1, line 161: maybe unnecessary optimization to trim already bound pvcs, should we keep them or is this runaway prevention?

docs/adrs/061-warm-pvc-pool.md
jezekra1: is it atomic? What if two agents claim the same PVC during reconciliation?

packages/controller/pkg/reconciler/warmpool.go
pilartomas: It's the second one — keeping it, but my comment undersold why (clarified in 0f8c3c65). It's not runaway prevention (creation is already capped at `target`); it's what makes `target` a **setpoint rather than a floor**. Without trimming Bound spares, lowering a pool's target — or setting `target: 0` — would never release the surplus *already-provisioned* storage, so a scaled-down pool would silently keep paying for idle volumes. Churn isn't a real risk because target changes are deliberate operator edits, not automatic oscillation.

docs/adrs/061-warm-pvc-pool.md
pilartomas: Yes, atomic — two independent guarantees:

1. **Serialized reconciles** — agent CRs are drained by a single `runAgentWorker` goroutine (only the *fork* worker runs separately), so two agents are never inside `claimSpare` at the same time.
2. **Optimistic concurrency** — even setting that aside, `claimSpare` relabels via an `Update` on the fetched PVC, which carries its resourceVersion. A stale write returns 409 Conflict; we catch `IsConflict`/`IsNotFound` and move to the next candidate. So two writers can never both win the same PVC. (`TestReconcile_ClaimRetriesOnConflict` exercises this.) The only other writer, the pool manager, never claims — it only creates/deletes *available* spares.

Captured as a one-line invariant in ADR-061 (0f8c3c65), since this and the earlier theft question both live here.

jezekra1 (approved):

PR #707, opened 2026-06-08
fix: faster hello apply + guard not ready seeding agents 

### LINKED ISSUE #695: Agent appears "ready" before its working directory is seeded
## Problem

When you create an agent that seeds its working directory (cloned from a git repo) or that receives initial configuration (skills, config files), the agent can show as **ready** before that setup has actually finished landing. If you open a session right away, the agent may not yet have its files or skills — so a freshly-created agent can behave as if it's misconfigured (for example, it doesn't know about the skills a repo ships).

The lag is very noticeable today: **even a small repo can take around a minute to appear, and that delay is in the platform, not the clone** — a large repo then adds its clone time on top.

This is a real first-use papercut for the working-directory-seeding work (#676): the seeded files are often exactly what the first session needs (a repo's operating manual, its skills), so "ready but not yet seeded" undermines the feature.

## Proposed solution

- An agent should **not be surfaced as ready until its initial setup (working-dir seed + configuration) has settled.** Until then, show a **preparing** state so a session isn't started against a half-set-up agent. (A per-item "required for ready" marker is one option to explore.)
- **Investigate and reduce the platform delay** before that initial setup lands on a freshly-started agent — even small repos lag ~a minute today, and it sometimes appears only after a delay/retry rather than immediately. Worth checking: how quickly the platform notices the agent is fully up vs. waiting for a periodic retry, and whether readiness is reported a few seconds early. **This needs investigation — don't assume the cause.**
- Add **timestamps** to the agent's setup/apply logs (they currently have none) so this is diagnosable.

### Dependencies

Surfaces during the working-directory seeding work — #676.

jjeliga:
Fix slow first-apply after agent boot: a freshly-started agent's initial configuration/working-directory seed could take up to ~1 minute to land because the apply retried on an exponential backoff that grew sparse right as the agent became ready. Hello-triggered apply jobs now re-check readiness on a tight fixed cadence, so config/seed lands within ~1s of the agent going Ready. Also added timestamps to the api-server and in-pod setup/apply logs so the boot timeline is diagnosable.

Add a "preparing workspace" agent state: an agent seeded from a git repo no longer shows as "Running" before its clone has finished. While the agent is ready but its workspace-seed is still pending, it now surfaces a preparing_workspace state — an amber pill in the UI that keeps the agent non-clickable — flipping to "Running" once the seed settles, so sessions can't be opened against a half-seeded working directory.

Touches only the git-seeded agents, not local file imports

Closes #695

### DIFF
--- a/packages/agent-runtime/src/modules/runtime-channel/compose.ts
+++ b/packages/agent-runtime/src/modules/runtime-channel/compose.ts
@@ -40,7 +40,11 @@ export interface ComposeRuntimeChannelOpts {
 export async function composeRuntimeChannel(
   opts: ComposeRuntimeChannelOpts,
 ): Promise<RuntimeChannelComposition> {
-  const log = opts.log ?? ((m) => process.stderr.write(`[runtime] ${m}\n`));
+  // Timestamp the in-pod seed/apply lines so the boot timeline is diagnosable (#695).
+  const log =
+    opts.log ??
+    ((m) =>
+      process.stderr.write(`${new Date().toISOString()} [runtime] ${m}\n`));
 
   const manifest = loadManifest(opts.manifestPath);
 
--- a/packages/api-server-api/src/modules/agents/types.ts
+++ b/packages/api-server-api/src/modules/agents/types.ts
@@ -36,6 +36,7 @@ export type ChannelConfig = SlackChannel | TelegramChannel;
 
 export type AgentState =
   | "starting"
+  | "preparing_workspace"
   | "running"
   | "hibernating"
   | "hibernated"
--- a/packages/api-server/src/modules/agents/infrastructure/agent-mappers.ts
+++ b/packages/api-server/src/modules/agents/infrastructure/agent-mappers.ts
@@ -53,11 +53,15 @@ export interface InfraAgent {
   error?: string;
 }
 
-/** Map the controller's conditions to the public-facing AgentState. Purely
- *  condition-driven (ADR-059) — the non-authoritative phase is not read. */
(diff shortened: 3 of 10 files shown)


packages/agent-runtime/src/modules/runtime-channel/compose.ts
```diff
@@ -40,7 +40,11 @@ export interface ComposeRuntimeChannelOpts {
 export async function composeRuntimeChannel(
   opts: ComposeRuntimeChannelOpts,
 ): Promise<RuntimeChannelComposition> {
-  const log = opts.log ?? ((m) => process.stderr.write(`[runtime] ${m}\n`));
+  // Timestamp the in-pod seed/apply lines so the boot timeline is diagnosable (#695).
+  const log =
+    opts.log ??
+    ((m) =>
+      process.stderr.write(`${new Date().toISOString()} [runtime] ${m}\n`));
```
jezekra1, line 47: let's do timestaps globally, not for every log
we should have a logger instead of process.stderr.write

jezekra1: ignore it in this PR and let's do it separately

jezekra1 (changes requested):

packages/agent-runtime/src/modules/runtime-channel/compose.ts
jjeliga: Makes sense

jezekra1 (approved):

PR #711, opened 2026-06-08
fix: better import ux

### LINKED ISSUE #701: Adding local files when creating an agent gives no feedback during a long wait
## Problem

When you create a new agent and attach local files, the dialog gives almost no feedback while the upload is in flight. The submit button disables and then nothing visibly happens.

This is easy to mistake for a bug. With a small upload the wait is short enough that it mostly works. But with a large input (for example, attaching the whole DAM repo) the UI sits silent for a couple of minutes, then the files suddenly appear. During that gap there's no way to tell whether the upload is progressing, stalled, or has failed.

Worse, nothing stops you from interacting with the component while this is happening, so it's unclear whether your input is even being respected.

The net effect is that a normal action (attach a repo, create the agent) feels broken, and the larger the input the worse it feels.

## Proposed solution

While the upload is in flight:
- **Show a visual clue** that work is in progress, so the silence reads as "working" rather than "stuck."
- **Disable manipulation** of the upload component so the user can't change the selection or re-trigger the action mid-flight.

When it finishes, a clear success or failure state closes the loop.

### Out of scope
- Real per-file or per-phase progress reporting — a single clear "working" indicator is enough.
- Changing how files are transmitted or stored, or any size limit — this is about feedback, not the transport.

jjeliga:
- **Disabled the Files-panel upload buttons while an upload is in-flight** — the toolbar's *Upload files* / *Upload folder* buttons go disabled (with a "Upload in progress…" title) so you can't fire overlapping imports.

- **Fixed a staging-dir leak race + hid the staging dirs** — in agent-runtime, the abort path (`fail()`) now awaits the in-flight extraction before `rm`-ing the staging dir, so the tar writer can't recreate it mid-cleanup; the `.import-staging-*` dirs are also filtered out of the file listing. The shared prefix constant was relocated to `core/import-staging.ts` (shared kernel) so `import`, `files`, and the sweeper reference one source without crossing module boundaries.

- **Added an "Importing…" status pill** — a per-agent in-flight counter in the Zustand files store (`importingAgents`, via `beginImport`/`endImport`), surfaced as a badge next to the status pill in the **agent-detail chat header only** (not the agent list), with a hover tooltip "Importing files into the agent". Wired for both the toolbar/drag upload path and the import-at-create path.

Closes #701

### DIFF
--- a/packages/ui/src/modules/agents/api/mutations.ts
+++ b/packages/ui/src/modules/agents/api/mutations.ts
@@ -3,7 +3,6 @@ import { useMutation } from "@tanstack/react-query";
 import { api } from "../../../api.js";
 import { emitToast } from "../../../lib/toast.js";
 import { queryClient } from "../../../query-client.js";
-import { useStore } from "../../../store.js";
 import { trpc } from "../../../trpc.js";
 import type { EgressPreset, EnvVar } from "../../../types.js";
 import { egressRulesKeys } from "../../egress-rules/api/queries.js";
@@ -12,6 +11,7 @@ import {
   type BundleEntry,
   importRawBundle,
 } from "../../files/api/import-bundle.js";
+import { trackImport } from "../../files/track-import.js";
 import { agentsKeys } from "./queries.js";
 
 const invalidatesAgentsList = {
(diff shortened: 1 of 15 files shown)


packages/ui/src/modules/agents/api/mutations.ts
```diff
@@ -85,11 +86,12 @@ export function useCreateAgent() {
       }
 
       if (preparedBundle) {
+        // Dialog has closed; the signal rides the store keyed on the new agent,
+        // where the chat-header pill reads `useIsImporting`.
```
kapetr: I'd drop the comment.

kapetr (approved):

PR #700, opened 2026-06-08
feat(cli): manage skills from the CLI

PetrBulanek:
Brings the `dam` CLI to parity with the web UI's Skills panel for the core install/use workflow: browse a connected source's catalog, see what's installed on an agent (with drift detection), and install or uninstall skills.

New `dam skill` command group:

- `dam skill source list [--agent <ref>]` — list registered skill sources (User / Platform / Agent).
- `dam skill catalog <source> [--agent <ref>]` — scan a source's skills; with `--agent`, annotate each as not-installed / installed / update-available.
- `dam skill list <agent>` — show the skills installed on an agent.
- `dam skill install <agent> --source <id|url> --name <skill>` — install at the source's current HEAD (re-running updates to HEAD — the same as the UI's "Update").
- `dam skill uninstall <agent> --source <id|url> --name <skill>` — remove a skill (idempotent).

A hibernated agent wakes transparently for install/uninstall and private-source scans; an unreachable agent, or a source whose GitHub needs connecting, surfaces a clear, distinct message. Source management and publishing stay in the UI for now (separate follow-ups under #329).

Closes #671

Issue #671, opened 2026-06-05
Manage skills from the CLI


PetrBulanek:
## Context

Part of #329 (P1). The UI's Skills panel browses git-sourced skill catalogs, installs/uninstalls skills onto an agent, and shows what's installed with drift detection. The CLI exposes none of it. This issue brings parity for the **core install/use** workflow; **source management** and **publish** are separate follow-ups (see Out of scope). Concepts (sources, scanned skills, installed refs, drift, skill paths) are owned by [skills.md](../blob/main/docs/architecture/skills.md) / [ADR-030](../blob/main/docs/adrs/030-skills-marketplace.md).

**Blocked by #670** — install/uninstall/private-catalog need a running pod; #670 makes those paths wake a hibernated agent transparently, so the CLI never special-cases hibernation.

## Scope

Add a `dam skill` group (singular, per [ADR-039](../blob/main/docs/adrs/039-cli-foundation.md)):

- [ ] `dam skill source list [--agent <id>]` — registered sources (id, name, gitUrl, kind: user/Platform/Agent). Read-only.
- [ ] `dam skill catalog <source> [--agent <id>]` — scan one source. With `--agent`, annotate each skill `not-installed` / `installed` / `update-available`. Public-GitHub scans without `--agent`; private/non-GitHub require it.
- [ ] `dam skill list <agent>` — installed inventory (source, name, short version), gitUrls resolved to source names.
- [ ] `dam skill install <agent> --source <id|url> --name <skill>` — install at the source's current HEAD. Idempotent: re-running updates to latest HEAD.
- [ ] `dam skill uninstall <agent> --source <id|url> --name <skill>` — remove it.

## API surface

Pure tRPC → [`skills` router](../blob/main/packages/api-server-api/src/modules/skills/router.ts) ([schemas](../blob/main/packages/api-server-api/src/modules/skills/schemas.ts)):

- `skills.sources.list(agentId?)` → `source list`
- `skills.list(sourceId, agentId?)` → `catalog` (the scan)
- `skills.state(agentId)` → `list` (render the `installed` array only)
- `skills.install(agentId, source, name, version, contentHash?)` → `install`
- `skills.uninstall(agentId, source, name)` → `uninstall`

## Non-obvious behavior to surface

[...]

PR #712, opened 2026-06-08
feat(cli): manage skill sources from the CLI

### LINKED ISSUE #673: Manage skill sources from the CLI
## Context

Part of #329 (P1) — follow-up to #671, which shipped the `dam skill` core plus a read-only `dam skill source list`. This adds the *mutating* source verbs so users can register, remove, and refresh git-sourced skill catalogs without the UI. Concepts in [skills.md](../blob/main/docs/architecture/skills.md) / [ADR-030](../blob/main/docs/adrs/030-skills-marketplace.md).

## Scope

Extend the `dam skill source` sub-group:

- [ ] `dam skill source add <git-url> [--name <n>]` — register a user source. Name is **derived from the git URL** by default (`github.com/apocohq/skills` → `apocohq/skills`); `--name` overrides, and on a TTY the derived name is offered as an editable default.
- [ ] `dam skill source remove <id|url>` — delete a user source. Resolve-first; confirm on TTY; `--yes` bypasses.
- [ ] `dam skill source refresh <id|url>` — drop the source's scan cache so the next `catalog` re-scans upstream.

## API surface

Pure tRPC → [`skills` router](../blob/main/packages/api-server-api/src/modules/skills/router.ts) ([schemas](../blob/main/packages/api-server-api/src/modules/skills/schemas.ts)):

- `skills.sources.create({name, gitUrl})` → `add` (`name` 1–128 chars, `gitUrl` a valid URL)
- `skills.sources.delete({id})` → `remove`
- `skills.sources.refresh({id})` → `refresh`

## Non-obvious behavior to surface

- **`remove` is untrack, not uninstall.** Deleting a source scrubs the `agent_skills` tracking rows for its gitUrl **across *all* your agents**, but the skill **files stay on the pods** (they become standalone/untracked). The confirm prompt must say so ("the files stay installed"). **No `--purge`** — out of scope.
- **Protected sources can't be removed.** System (Helm-seeded) and template-declared sources return `FORBIDDEN`. The CLI rejects them **client-side** (it already has the resolved row) with a precise message, not a bare `FORBIDDEN`. `refresh` is allowed on any kind.
- **Adding a URL that matches a system/template source shadows it.** Dedup is first-wins user→template→system, so a user source for the same gitUrl wins; removing it re-exposes the system/template entry. Worth a line in `add --help`.
- **Resolve-first for `remove`/`refresh`.** `--source` ref is id|url (no name — names allow spaces), resolved client-side via `sources.list` — the same resolver as core. Unknown ref → `EXIT_INVALID_INPUT` (don't rely on server idempotency); `refresh`'s `NOT_FOUND` maps to a not-found exit.

## Conventions to follow

Mirror [`agent/commands/create.ts`](../blob/main/packages/cli/src/modules/agent/commands/create.ts); `--json` parity; distinct exit codes; the confirm/`--yes`/non-TTY-refuses pattern from `agent delete` / `import`; integration tests in `packages/cli/src/__tests__/`. Reuse the core `--source` resolver.

## References

- Epic #329 · Follows #671
- [skills.md](../blob/main/docs/architecture/skills.md) · [ADR-030](../blob/main/docs/adrs/030-skills-marketplace.md) · [ADR-039](../blob/main/docs/adrs/039-cli-foundation.md) · [cli.md](../blob/main/docs/architecture/cli.md)

PetrBulanek:
Adds the three mutating verbs to the `dam skill source` group, so you can register, remove, and refresh git-sourced skill catalogs from the CLI without the web UI:

- **`dam skill source add <git-url> [--name]`** — register a source. The name defaults to the URL path and is offered as an editable default on a TTY. Re-adding a URL you already have is reported cleanly ("already registered") instead of as a raw error. Adding a URL that a platform- or template-provided source already uses shadows it.
- **`dam skill source remove <source>`** — stop tracking a source across all your agents. This is untrack, not uninstall: the installed skill files stay on the pods. Platform- and template-provided sources can't be removed.
- **`dam skill source refresh <source>`** — drop the cached scan so the next `catalog` re-reads upstream.

This completes the source side of the `dam skill` group; browsing, listing, and install/uninstall shipped in #671. Publishing stays out of scope.

> Stacked on #671 — this PR targets `feat/671-skills-cli`. Review/merge #700 first.

Closes #673


packages/cli/src/modules/skill/commands/source-remove.ts
```diff
@@ -0,0 +1,114 @@
...
+import type { SkillsService } from "../services/skills-service.js";
+
+export function buildSourceRemoveCommand(deps: {
+  compatService: CompatService;
+  configService: ConfigService;
+  createSkillsService: (host: string) => SkillsService;
+}): Command {
+  return new Command("remove")
+    .description(
+      "Untrack a User skill source (the installed files stay on your agents)",
```
JanPokorny: I'd simplify the language a bit

```suggestion
      "Remove a skill source (installed skills remain)",
```

packages/cli/src/modules/skill/domain/source-name.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Default a skill-source name from its git URL: the URL path, minus a leading
+ * `/`, a trailing `/`, and a trailing `.git`.
+ *   https://github.com/apocohq/skills(.git)(/) → "apocohq/skills"
+ *   https://gitlab.com/group/subgroup/repo      → "group/subgroup/repo"
+ *   https://github.com                          → ""  (underivable)
+ * Returns "" on parse failure or an empty path. Slashes are kept on purpose —
+ * skill-source names allow them (schema is min(1).max(128) with no slug
+ * pattern), unlike connection names, so this does NOT slugify.
+ */
```
JanPokorny, line 10: I'd keep the domain in the name, e.g. `github.com/apocohq/skills`

(unless we already do it this way on the backend?)

packages/cli/src/modules/skill/domain/source-ref.ts
```diff
@@ -15,3 +15,9 @@ export function resolveSourceRef(
   }
   return sources.find((s) => s.id === ref) ?? null;
 }
+
+export function sourceKind(s: SkillSource): "Platform" | "Agent" | "User" {
+  if (s.system === true) return "Platform";
```
JanPokorny: what cursed type is this variable if you have to do `=== true`? 😅

packages/api-server/src/core/db-errors.ts
```diff
@@ -0,0 +1,22 @@
...
+    const obj = cur as { code?: unknown; constraint_name?: unknown };
+    if (
+      obj.code === "23505" &&
+      (constraintName === undefined || obj.constraint_name === constraintName)
+    ) {
+      return true;
+    }
+  }
+  return false;
+}
```
JanPokorny, line 23: Are you sure this is necessary? Not saying that it isn't but it looks suspiciously like what a confused agent would write instead of reading Drizzle documentation.

JanPokorny (approved):

packages/cli/src/modules/skill/domain/source-ref.ts
PetrBulanek: Completely redundant. The type is just `boolean().optional()`.

packages/cli/src/modules/skill/domain/source-name.ts
PetrBulanek: The `source list` table already shows the full git URL in its own column, so the host in the name would be redundant. The name is just an editable display label (resolution keys on id/url, not name). The web UI doesn't derive a name at all today (manual field), so there's no backend convention to match.

packages/api-server/src/core/db-errors.ts
PetrBulanek: I did go read the Drizzle doc and it seems there's no typed unique-violation error, so inspecting `.code === '23505'` on the driver error is probably the right approach.

It's also not new. Slack's `isSlackChannelUniqueViolation` already did the same. That's probably where the agent picked it up. I merged it into this shared helper.

packages/cli/src/modules/skill/domain/source-name.ts
JanPokorny: OK fair enough

packages/api-server/src/core/db-errors.ts
JanPokorny: Funny how all ORMs break down as soon as you want to use a semi-advanced SQL feature 😭

PR #774, opened 2026-06-09
fix(db): squash migrations to one baseline and restore a working workflow

### LINKED ISSUE #739: Squash database migrations and restore a working migration workflow
## Problem

New database migrations can no longer be generated automatically. The migration tool authors each new migration by diffing the code's schema definitions against its own saved model of the database — and that model drifted out of sync long ago and is frozen at an early point in the history. As a result, every migration since then has been hand-written, and the file that tracks migration order is hand-edited along with each one.

This is fragile and compounding:

- Hand-editing the ordering file is error-prone — it already carries a duplicate timestamp from a past slip.
- A large set of reporting views lives only in raw SQL and was never expressed in the code's schema definitions, so the generator couldn't reproduce the database even if we repaired the drift.
- Nothing verifies that the schema the application compiles against actually matches the database the migrations produce. They can silently diverge — e.g. a column the code treats as required might not be enforced anywhere.

The practical effect: any database change is slow, fully manual, and unguarded against mistakes.

## Proposed solution

- **Squash** the full migration history into a single baseline that reproduces the current database exactly — including the views and indexes the generator can't model. This is safe for existing deployments: they keep their current database untouched and recognize the baseline as already applied; only fresh installs run it.
- **Make hand-written SQL the sanctioned workflow**, since the schema legitimately uses database features the generator can't round-trip, and retire the stale generator state that's currently misleading.
- Add a small **scaffolder** that creates the next migration file and updates the ordering file correctly, so new migrations no longer require hand-editing.
- Add a **CI check** that applies all migrations to a throwaway database and confirms the result matches the code's schema definitions — catching drift before it ships.

### Scope
The api-server's Postgres migrations only. The schema's contents don't change — this is about how migrations are authored and verified.

### Out of scope
- Switching to a different migration tool — considered but deferred; the plan above keeps the current tooling and drops the part that doesn't fit.

JanPokorny:
Closes #739.

## Problem

Drizzle's generator had broken down: its snapshots froze at `0009` while the history grew to `0022` (so every recent migration was hand-written and the journal hand-edited), `drizzle-kit generate` couldn't even resolve `drizzle-orm` under pnpm, the `usage_*` reporting views live only in raw SQL the generator can't model, and nothing verified that the migrations still produced the schema the code compiles against — they had silently diverged.

## Change

- **Squash** `0000`–`0022` into a single `0000_baseline.sql` that reproduces the current schema exactly, including the `usage_*` views. It keeps the original `0000` journal timestamp, so **existing deployments recognise it as already applied and skip it; only fresh installs run it**. This is safe by construction: the drizzle-orm migrator decides what to run by comparing `created_at` (never the file hash) and applies all pending migrations in a single transaction, so no database is ever partially migrated.
- **Reconcile the one real drift found:** `agent_skill_publishes`'s primary key was still named `instance_skill_publishes_pkey` (a leftover from before ADR-046 renamed the table). `0001_rename_skill_publishes_pk.sql` renames it on every deployment so all converge on the name `schema.ts` implies.
- **Retire the generator; make hand-written SQL the sanctioned workflow.** `mise run db:new -- <name>` scaffolds the next migration file + journal entry (no more hand-editing); `db:migrate` now uses the api-server's own `runMigrations` path. The stale drizzle-kit snapshots are deleted.
- **Add a drift check** — `mise run db:drift` applies every migration to a throwaway database and confirms the result matches `schema.ts` (column-order-insensitive, views excluded), wired into CI as a job with a Postgres service. A `pnpm.packageExtensions` entry makes `drizzle-orm` resolvable to `drizzle-kit` so the check can render `schema.ts`.
- **Docs:** ADR-063 captures the decision; `persistence.md` and the CLAUDE.md migration steps are updated.

Scope is the api-server's Postgres migrations only — the schema's contents don't change (the constraint rename is name-only).

## Verification

- `db:drift` passes; negative-tested (it flags the constraint drift if the rename is removed).
- Fresh-install and existing-deployment (baseline-skip) behaviour proven against a real Postgres 16, then **on an actual k3s cluster**: the built api-server image, run as a pod, applies baseline + rename on a fresh DB (18 tables / 23 views / renamed PK / 2 journal rows) and **skips the baseline** on a DB seeded with the pre-squash history (23 → 24 rows, only the rename applied).
- `mise run test` green. `mise run check` is green except a pre-existing, unrelated `ui:check:tsc` Carbon-icon type error (reproduces on `main` with these changes reverted).


jezekra1:
duplicate of #772

PR #778, opened 2026-06-09
feat(agents): front custom Anthropic upstreams with a local LiteLLM gateway

### LINKED ISSUE #702: Claude Code: dynamically use all models the LiteLLM gateway provides
## Problem

When Claude Code connects through the LiteLLM gateway, the set of usable models
is static. The gateway already publishes the full catalog of models the
connected token can reach, but Claude Code doesn't drive model selection from
that live list — so not every model the gateway offers is usable, and when the
catalog changes (models added, retired, renamed) the configuration silently
goes stale and the user hits a confusing runtime failure with no in-product way
to see what's actually available.

## Proposed solution

Have Claude Code dynamically retrieve the gateway's live model catalog (using
the connected token) and make every model it provides usable — both as the
defaults Claude Code runs on and as models the user can switch to — with no
hand-pinned model IDs to keep in sync.

The net effect: connect the gateway once, and every model that token can reach
is immediately usable in Claude Code.

### Scope

- The LiteLLM gateway connection for Claude Code.

### Out of scope

- Pi Coding Agent — tracked separately in #780.
- Other provider connections (direct Anthropic, direct OpenAI, Bob).
- A generic OpenAI-compatible provider with model discovery — tracked in #393.

Tomas2D:
## Summary

Lets the **claude-code** agent run behind a custom Anthropic-compatible upstream (e.g. the IBM LiteLLM ETE proxy) by fronting it with a small **local LiteLLM gateway**. Activated only when `ANTHROPIC_BASE_URL` points at a custom upstream; otherwise it's a complete no-op and Claude Code talks to the Anthropic API directly.

## How it works

- **Conditional activation** — the harness shims source [`litellm-proxy.sh`](packages/agents/claude-code/litellm-proxy.sh). It runs in the shim (not the image entrypoint) because credentials arrive over the runtime channel only once the harness is spawned. If `ANTHROPIC_BASE_URL` is a custom upstream, it brings the gateway up **once per pod** (atomic `mkdir` lock), health-gates it, then re-points `ANTHROPIC_BASE_URL` at `127.0.0.1:4000` and adds loopback to `NO_PROXY`.
- **Model discovery** — [`litellm-gateway.py`](packages/agents/claude-code/litellm-gateway.py) fetches the upstream's `GET /v1/models` and generates concrete `model_list` entries (a wildcard + `check_provider_endpoint` can't enumerate models for `litellm_proxy` providers — [BerriAI/litellm#20064](https://github.com/BerriAI/litellm/issues/20064) — and `/model/new` needs a database). Models are exposed `anthropic/`-prefixed (the provider namespace gateway discovery expects); the upstream id stays verbatim in `litellm_params.model` for routing. Re-fetches periodically and restarts LiteLLM only when the model set changes.
- **Model pins** — Claude Code's `ANTHROPIC_DEFAULT_{OPUS,SONNET,HAIKU}_MODEL` / `ANTHROPIC_MODEL` are set to the latest discovered model per tier (with fallback), **assign-if-unset** so a value set manually on the agent wins.
- **TLS** — LiteLLM's transport verifies against `ssl.create_default_context()`, which trusts no CAs on this image unless `SSL_CERT_FILE` is set; the gateway is launched pointing at the system bundle (public CAs + the platform MITM CA) so the upstream hop through the Envoy gateway verifies.

## Provider cleanup

The IBM LiteLLM provider preset and connections catalog used to inject the Claude model pins; the gateway now owns them, so they're removed from both, along with the "Advanced — model overrides" form. `OPENAI_PROXY_MODEL` / `OPENAI_MODEL` (pi-agent / Codex, which have no gateway) stay as fixed defaults. `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC` is also dropped from the preset.

## Testing

- Container-tested: discovery → `anthropic/`-prefixed `/v1/models` + correct latest-per-tier selection (incl. haiku→opus fallback); unreachable upstream → wildcard fallback with the proxy still healthy; assign-if-unset preserves a manually-set model; loopback bypasses the pod proxy.
- Verified end-to-end in a live cluster (completions through the gateway after the `SSL_CERT_FILE` fix).
- `api-server` unit tests pass (193/193); api-server-api + UI type-check clean.

## Notes

- `litellm[proxy]` adds a Python dependency tree to the claude-code image — worth watching `agents:claude-code:scan:trivy`.
- Existing stored provider secrets keep their snapshotted env bundle until re-saved; re-save the provider to pick up the pin removals.

Resolves: #702


packages/agents/claude-code/Dockerfile
```diff
@@ -11,6 +30,18 @@ RUN --mount=type=cache,target=/root/.npm \
...
+COPY litellm-proxy.sh /usr/local/lib/litellm-proxy.sh
+# SSH / VS Code Remote-SSH sessions (ADR-062) bypass the harness shims, so also
+# front the upstream from a login hook. profile.d is sourced by interactive login
+# shells but not by sftp/scp (no shell) or `ssh host cmd` (non-login).
+COPY litellm-login.sh /etc/profile.d/litellm-proxy.sh
+# Override the platform-base shims with claude-code's, which source the proxy
+# helper before launching the harness.
+COPY harness-chat.sh /usr/local/bin/harness-chat
+COPY harness-terminal.sh /usr/local/bin/harness-terminal
+RUN chmod +x /usr/local/bin/harness-chat /usr/local/bin/harness-terminal
```
JanPokorny: https://docs.docker.com/reference/dockerfile/#copy---chmod 😁

packages/agents/claude-code/Dockerfile
```diff
@@ -3,6 +3,25 @@ FROM ${BASE_IMAGE}
 
 USER root
 
+# Local LiteLLM gateway. When the agent is provisioned against a custom
```
JanPokorny: Opus goes insane with these long inline comments, explaining one thing in like 5 different places 😁 I'd remove most of them

packages/agents/claude-code/Dockerfile
```diff
@@ -3,6 +3,25 @@ FROM ${BASE_IMAGE}
...
+# front it with this proxy and re-point Claude Code at it — see
+# litellm-proxy.sh. LiteLLM is a Python package and the base ships no Python,
+# so pull python3.12 (litellm 1.88.x requires >=3.10,<3.14) and bootstrap pip
+# via ensurepip (there is no python3.12-pip rpm). Pin the version — bump
+# deliberately (no-downgrade policy). Installed before the npm layer so a
+# Claude Code CLI bump (frequent) doesn't rebuild this heavy layer.
+ARG LITELLM_VERSION=1.88.1
+RUN --mount=type=cache,target=/root/.cache/pip \
+    dnf install -y python3.12 && dnf clean all \
+    && python3.12 -m ensurepip --upgrade \
```
JanPokorny: Alternatively you can mount the uv binary from their scratch image for just this step and `uv tool install` -- it will probably be slightly faster and slightly smaller

JanPokorny (changes requested):
First of all, great work! This brings actually good model selection UX to DAM's Claude.

I added some nitty comments but the elephant in the room is the LiteLLM gateway lifecycle ducktaped using `nohup` into an OCI container, using a lock to simulate a system service. I am quite sure that at the current state it leaks zombie processes or some similar undesirable thing, it's hard to reason about homebrew service managers 😁

I think there must be a better way to manage the service, e.g. asking `agent-runtime` to manage it for us, or using a more advanced init system in the container. But I have to try things out before giving a more concrete answer.

packages/agents/claude-code/Dockerfile
Tomas2D: Done in b5f7d9e1 — `COPY --chmod=0755` for both shims, dropped the separate `RUN chmod`. 🙏

packages/agents/claude-code/Dockerfile
Tomas2D: Agreed 😄 — stripped most comments here and across the gateway scripts (`litellm-gateway.py`, `litellm-proxy.sh`, `litellm-login.sh`, the shims) in b5f7d9e1, keeping only the non-obvious *why*s.

packages/agents/claude-code/Dockerfile
Tomas2D: Switched to uv in b5f7d9e1: mount the pinned uv binary and `uv pip install --system`.

Kept `python3.12` from dnf because the gateway runs as a `python3.12` script and imports `yaml` (ships with `litellm[proxy]`), so a fully isolated `uv tool install` would split that env — `--system` keeps it importable from the same interpreter. Bonus: drops `pip`/`ensurepip` from the image, so it's a touch smaller too.

Tomas2D:
On the gateway lifecycle (the 🐘 in the room): for now I've **hardened** the current approach rather than redesign, since you mentioned wanting to try things out before settling on a direction.

Two clarifications on the zombie worry:
- The supervisor (`litellm-gateway.py`) does reap its LiteLLM child (`proc.wait()`) and forwards `SIGTERM`, so in steady state nothing accumulates.
- The genuine leak I found and fixed in b5f7d9e1: if the bring-up shim timed out waiting for readiness it left the gateway running headless — and since it also dropped the lock, a later session would start a **second** gateway on the same port. The shim now stops the gateway it started before falling back to the upstream.

The deeper "let `agent-runtime` own the process" / proper-init direction sounds right — happy to implement whichever you land on once you've experimented. Want me to open a follow-up issue to track it?

jezekra1 (approved):

PR #820, opened 2026-06-10
feat(platform-base): fortify Claude model proxy

JanPokorny:
Replace bundled LiteLLM proxy with a lightweight Node.js script -- reduced image size (saves cca 1 GB), memory use, and supply chain surface. Explicitly support background services. Add machinery that ensures proper zombie reaping. Reload env on SIGHUP -> changes propagate immediately to the running instance.


packages/agents/claude-code/model-gateway-login.sh
```diff
@@ -0,0 +1,3 @@
+case $- in
```
jezekra1, line 1: do we still need this login file?

packages/agent-runtime/src/modules/pod-service.ts
```diff
@@ -0,0 +1,83 @@
...
+
+const BACKOFF_INITIAL_MS = 1_000;
+const BACKOFF_MAX_MS = 30_000;
+const HEALTHY_RUN_MS = 60_000;
+
+export interface PodServiceSupervisor {
+  refreshEnv(): void;
+}
+
+export function createPodServiceSupervisor(opts: {
```
jezekra1, line 14: have you thought about putting this to [runtime-manifest](https://github.com/dam-agents/dam/blob/main/docs/architecture/connections.md?plain=1#L455) as a plugin/driver?

that's our mechanisms for extending the functionality of the base runtime and it has support for custom driver implementations. Maybe it's not a good fit, just want to make sure we concidered it

jezekra1 (changes requested):

packages/agents/claude-code/model-gateway-login.sh
JanPokorny: it's for SSH sessions so that `claude` also works correctly there (since users will launch it directly by `claude` and not through `harness-terminal`)

packages/agent-runtime/src/server.ts
```diff
@@ -83,6 +86,22 @@ const stateBackend = createFileDocumentStoreBackend(homeDir);
 // (harness, terminal, ssh, git) read it through the RuntimeEnvReader port.
 const envStore = createEnvStateStore(homeDir);
 
+const podServicePath = "/usr/local/bin/pod-service";
+const podLog = (msg: string) => process.stderr.write(`[pod-service] ${msg}\n`);
+const podService = existsSync(podServicePath)
+  ? createPodServiceSupervisor({
+      command: podServicePath,
+      snapshot: stateBackend.open("pod-service-env", {
```
jezekra1: the env service is already in control of the env state, why do we need another?

packages/agent-runtime/src/modules/pod-service.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  process.once("exit", () => {
+    try {
+      child?.kill("SIGKILL");
+    } catch {}
+  });
+
+  function start(): void {
+    const proc = spawn(command, [], {
+      env: mergedSpawnEnv(envReader),
```
jezekra1, line 38: is the process restarted when new env comes in through the runtime contributions?

packages/agent-runtime/src/modules/pod-service.ts
JanPokorny: IMO it's overkill for now -- `harness-terminal` and `harness-acp` are binaries by convention, I repeated that pattern here

packages/agent-runtime/src/modules/pod-service.ts
JanPokorny: It receives NOHUP, which by default terminates the process (which is then auto-restarted), _but_ if the process decides to catch NOHUP like we do here, it can just reload the env live.

packages/agent-runtime/src/server.ts
JanPokorny: I moved it to a more appropriate place. It's a store specific to the materialized env of the pod service, which is not entirely reconstructible from the env store which contains only the "to be injected" values.

jezekra1 (approved):

Tomas2D (approved):

PR #821, opened 2026-06-10
fix: report agent oom

### LINKED ISSUE #327: No user-facing feedback when an agent crashes from OOM
## Problem

When an agent exceeds its memory limit, Kubernetes kills the container. The user gets no explanation - long-lived agents silently restart and lose session state, fork jobs time out with a generic "pod not ready" error. Users are left staring at a dead session with no idea what happened or what to do about it.

## Open questions

- What should the error message say? Just the cause, or also a path to resolution?
- Should OOM events be logged/counted somewhere for platform admins to see patterns?
- How should this interact with other crash types — do we need a general "agent crashed" UX or OOM-specific handling?

## Out of scope

- Preventing OOM (separate issue - #328)
- Auto-recovery or session replay after OOM

jjeliga:
Catches agent pod OOM and shows a corresponding toast notification and a line in the starting overlay 

Closes #327


packages/ui/src/modules/agents/hooks/use-agent-crash-toasts.ts
```diff
@@ -0,0 +1,29 @@
+import { useEffect, useRef } from "react";
+
+import { emitToast } from "../../../lib/toast.js";
+import { useAgentsList } from "../api/queries.js";
+
+/** One toast per crash episode. Re-arms only on genuine recovery (running/hibernated), not on reason change — a crash-loop flaps the reason and would otherwise re-toast every restart. */
```
kapetr: The comment contradicts the code - a new toast is currently created on any reason change. I'd remove the comment in any case, but before that I'd validate the final solution is correct.

kapetr (approved):

packages/ui/src/modules/agents/hooks/use-agent-crash-toasts.ts
jjeliga: @kapetr True, one toast per crash episode: dedupes on the reason string and clears the entry only on genuine recovery (running/hibernated).
removed

PR #822, opened 2026-06-10
refactor(ui): unify buttons and inputs via variant-based components

### LINKED ISSUE #223: UI - Unify buttons and inputs via variant-based components
## Problem

Buttons and form inputs across the UI are styled by hand at every call site — each `<button>` and `<input>` carries a long, copy-pasted className string that names the size, color, padding, border, and shadow inline. The same pattern repeats dozens of times across the agents, secrets, connections, sessions, and settings surfaces, and small inconsistencies have already crept in: the "primary" button shows up with subtly different padding in different places, inputs come in two heights for no clear reason, and a couple of inputs have a "monospace" or "invalid" variation that only some screens use.

None of this is broken today, but the duplication has the same recurring costs we saw with the dialog shell: every cross-cutting change (a hover treatment, a focus ring tweak, a dark-mode color, a spacing adjustment) requires editing many files independently, and each new screen risks introducing yet another slightly-different shape. New contributors have no obvious "this is what a button looks like in this app" component to reach for, so they invent one — and that invention drifts further.

## Proposed solution

Extract a single `Button` and a single `Input` that capture the design system's actual variants — the handful of sizes and colors the app uses today — and migrate every call site onto them. Each component exposes named variants (primary / secondary / ghost / danger for buttons; standard / monospace / invalid for inputs) plus a small set of sizes; the call site picks the variant by name instead of restating its class list.

The user-visible effect is consistency: every button at a given role looks identical app-wide, and the same goes for inputs. There is no functional change. The future cost of cross-cutting styling changes drops to one file each.

### Scope

- Two new components: `Button` and `Input`. Every existing button-shaped and input-shaped element in the app migrates to them.
- A small className-composition utility so component callers can append local one-off classes without conflicting with the variant's own styles.

### Out of scope (separately tracked)

- Other repeated UI primitives (badges, pills, cards) — there's existing variant code for some of these already; consolidating them is a follow-up.

kapetr:
## Summary

Buttons and inputs across the app were styled by hand at every call site, with the same long className strings copy-pasted and small inconsistencies creeping in. This migrates the app's action buttons and text inputs onto a single shared `Button` and `Input`, so every button at a given role looks identical app-wide and cross-cutting style changes land in one place. No functional change.

- **Button** gains an `xs` size and a `tone="danger"` axis (neutral at rest, danger on hover), so call sites express small and danger buttons by name rather than with ad-hoc padding/color classes. The primary button stays black/white.
- **Input** gains `standard` / `monospace` / `invalid` variants and a `sm` size.
- Call sites keep only layout classes (`w-full`, `justify-*`, `shrink-0`, …) — no padding, color, or danger classes.

Surfaces migrated: sessions, agents dialogs, approvals, files, settings, connections, schedules, terms, and the sandbox card.

Closes #223

### DIFF
--- a/packages/ui/src/modules/sessions/components/skills-panel.tsx
+++ b/packages/ui/src/modules/sessions/components/skills-panel.tsx
@@ -19,6 +19,7 @@ import { useCallback, useEffect, useState } from "react";
 import { z } from "zod";
 
 import { Button } from "@/components/ui/button";
+import { Input } from "@/components/ui/input";
 
 import { api } from "../../../api.js";
 import { ACTION_FAILED, runAction } from "../../../lib/query-helpers.js";
@@ -490,8 +491,8 @@ export function SkillsPanel({
                   </option>
                 ))}
               </select>
-              <input
-                className={inp}
+              <Input
+                size="sm"
                 placeholder="Pull request title"
                 value={publishForm.title}
                 onChange={(e) =>
@@ -510,15 +511,13 @@ export function SkillsPanel({
               <div className="flex justify-end gap-2">
                 <Button
                   variant="outline"
-                  size="sm"
-                  className="h-7 text-[11px]"
+                  size="xs"
                   onClick={() => setPublishFor(null)}
                 >
                   Cancel
                 </Button>
                 <Button
-                  size="sm"
-                  className="h-7 text-[11px]"
+                  size="xs"
                   disabled={publishBusy || !publishForm.sourceId}
                   onClick={publish}
                 >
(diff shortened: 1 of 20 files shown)


packages/ui/src/modules/sessions/components/skills-panel.tsx
```diff
@@ -490,8 +491,8 @@ export function SkillsPanel({
                   </option>
                 ))}
               </select>
-              <input
-                className={inp}
```
PetrBulanek, line 494: The `inp` const is still there (one select uses it). Is it possible to get rid of it as well? Maybe migrate selects/textareas into own components too in a follow-up issue?

PetrBulanek (approved):

packages/ui/src/modules/sessions/components/skills-panel.tsx
kapetr: Good call. I've kept `inp` for now because the only remaining users are a `<select>` and a `<textarea>`, which are outside this PR's scope (buttons + text inputs) — removing it cleanly means introducing shared `Select`/`Textarea` components and migrating onto them, exactly as you suggest.

Tracked as a follow-up: #832.

kapetr: Follow-up #832

PR #833, opened 2026-06-11
fix(api-server): evaluate schedule fires and quiet hours in the schedule's timezone

### LINKED ISSUE #827: Schedule times and quiet hours are shifted — evaluated in UTC instead of the schedule's timezone
## Problem

Schedule fire times are being interpreted as UTC instead of the schedule's configured timezone. A schedule set for e.g. 9:00 in Prague fires at 9:00 UTC — two hours later than the user asked for.

Quiet hours stopped working around the same time, most likely the same shift: runs that should be suppressed inside a quiet window go through, presumably because the window is evaluated against the shifted clock.

This silently breaks every schedule whose owner isn't in UTC — agents run at the wrong time, including during hours the user explicitly marked as quiet.

## Proposed solution

Fire times and quiet-hour windows are evaluated in the schedule's configured timezone. A schedule set for 9:00 with quiet hours 22:00–06:00 fires at 9:00 local time and never inside 22:00–06:00 local time. Existing schedules keep the wall-clock times their owners entered.

jezekra1:
Fixes #827

## Root cause

`nextFireAt` (`packages/api-server/src/modules/schedules/domain/recurrences.ts`) returned rrule.js occurrences verbatim. rrule.js emits Dates whose **UTC fields echo the RRULE's `BYHOUR`/`BYMINUTE`** — i.e. wall clock in the schedule's timezone, not a real instant. That convention dates from when the Go controller owned firing and applied `TZID` itself; when scheduling moved into the api-server's BullMQ queue, `spec.timezone` stopped being applied anywhere, so:

- a 9:00 Europe/Prague schedule fired at 9:00 **UTC** (11:00 local), and
- quiet hours were suppressed against occurrence wall-times while actual runs executed offset-shifted, landing inside windows the user marked quiet (e.g. a 21:30-local occurrence survives suppression but executes at 23:30 local).

## Fix

- Convert `from` into the schedule-timezone wall-clock frame, iterate occurrences + quiet-hours there (the frame quiet-hour `HH:MM` strings are defined in), and convert only the surviving occurrence back to a real instant. The inverse conversion uses a two-pass offset lookup so DST boundaries resolve; spring-forward-erased wall times land just past the jump.
- Anchor `dtstart` explicitly in the wall-clock frame — rrule.js defaults it to real parse-time "now", which silently skips same-day occurrences in zones behind UTC.
- Pin legacy `type: cron` schedules to UTC per the ADR-031 contract — cron-parser 5.x otherwise evaluates in the *server's* local timezone (correct-by-accident in UTC pods, wrong everywhere else, including this laptop, which is how the test caught it).

Existing schedules keep the wall-clock times their owners entered — only the instant they map to changes (to the correct one). No schema or stored-spec changes.

## Tests

`schedules-recurrences.test.ts`: summer/winter offsets, zone-behind-UTC same-day fire, local-clock quiet-hour suppression across midnight, disabled windows, all-suppressed → null, UTC passthrough, spring-forward gap, legacy cron UTC pin. `mise run check` and `mise run api-server:test` (203/203) pass.

## Noted drift (not addressed here)

`docs/architecture/agent-lifecycle.md` ("Trigger fire") and `persistence.md` still describe the per-schedule **controller goroutine** computing occurrences in `TZID` and delivering trigger files via `kubectl exec` — scheduling actually lives in the api-server (BullMQ + runtime outbox). This page predates the move; worth a follow-up doc pass.

### DIFF
--- a/packages/api-server/src/__tests__/unit/schedules-recurrences.test.ts
+++ b/packages/api-server/src/__tests__/unit/schedules-recurrences.test.ts
@@ -1,0 +1,120 @@
+import { describe, it, expect } from "vitest";
+import type { ScheduleSpec } from "api-server-api";
+import { nextFireAt } from "../../modules/schedules/domain/recurrences.js";
+
+function rruleSpec(
+  rrule: string,
+  timezone: string,
+  quietHours?: { startTime: string; endTime: string; enabled: boolean }[],
+): ScheduleSpec {
+  return {
+    version: "platform.ai/v1",
+    type: "rrule",
+    rrule,
+    timezone,
+    quietHours,
+    enabled: true,
+    createdBy: "user",
+  };
+}
+
+describe("nextFireAt (rrule)", () => {
+  it("fires at the wall-clock time in the schedule's timezone, not UTC", () => {
+    // 09:00 Europe/Prague in June is 07:00Z (CEST, UTC+2).
+    const spec = rruleSpec(
+      "FREQ=DAILY;BYHOUR=9;BYMINUTE=0;BYSECOND=0",
+      "Europe/Prague",
+    );
+    const next = nextFireAt(spec, new Date("2026-06-11T00:00:00Z"));
+    expect(next?.toISOString()).toBe("2026-06-11T07:00:00.000Z");
+  });
+
+  it("tracks the timezone's winter offset", () => {
+    // 09:00 Europe/Prague in January is 08:00Z (CET, UTC+1).
+    const spec = rruleSpec(
+      "FREQ=DAILY;BYHOUR=9;BYMINUTE=0;BYSECOND=0",
+      "Europe/Prague",
+    );
+    const next = nextFireAt(spec, new Date("2026-01-15T00:00:00Z"));
+    expect(next?.toISOString()).toBe("2026-01-15T08:00:00.000Z");
+  });
+
+  it("does not skip a same-day occurrence in zones behind UTC", () => {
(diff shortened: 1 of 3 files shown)


packages/api-server/src/__tests__/unit/schedules-recurrences.test.ts
```diff
@@ -0,0 +1,120 @@
+import { describe, it, expect } from "vitest";
```
jjeliga, line 1: Diod you forgot to delete the unit test file? Although this case seems like a good candidate to keep for once

packages/api-server/src/modules/schedules/domain/recurrences.ts
```diff
@@ -98,20 +99,71 @@ function parseHHMM(s: string): number | null {
...
+        currentDate: from,
+        tz: "UTC",
+      });
       return cron.next().toDate();
     } catch {
       return null;
     }
   }
-  const rule = RRule.fromString(spec.rrule);
+  // rrule.js occurrences live in the wall-clock frame (see isInQuietHours),
```
jjeliga: cut the comments

jjeliga (approved):

packages/api-server/src/__tests__/unit/schedules-recurrences.test.ts
jezekra1: I'd keep this one, there is a lot of custom computation around dates

Tomas2D (approved):

PR #847, opened 2026-06-11
feat(cli): action pending approvals from the CLI

### LINKED ISSUE #773: Action pending approvals from the CLI
## Context

Part of #329 (P2). The UI surfaces a **unified HITL inbox** ([ADR-035](../blob/main/docs/adrs/035-unified-hitl-ux.md)) of pending human-in-the-loop decisions an Agent is waiting on; the CLI exposes none of it. This issue brings parity for **listing and actioning** that inbox. Two kinds of request share one queue:

- **`ext_authz`** — an outbound network request Envoy blocked (host / method / path), held synchronously (~30 min) then expired. Approving permanently writes an egress rule.
- **`acp_native`** — a harness tool-call permission request (toolName / args), live-only (~24h TTL). The harness owns its own permission persistence; no egress rule is written.

**Boundary with `dam network`:** `dam network` manages egress **rules** — the *pre-approvals* that stop a prompt from ever appearing. `dam approval` actions the *prompts that do appear*. Approving an `ext_authz` request permanently **writes a rule** that then shows up in `dam network list <agent>`. The two are complementary: pre-seed rules to avoid prompts; action prompts when they arrive. This unblocks headless / terminal users who today can only clear the inbox from the web UI.

## Scope

Add a `dam approval` group (singular, per [ADR-039](../blob/main/docs/adrs/039-cli-foundation.md)):

- [ ] `dam approval list [<agent>] [--status pending|resolved|expired] [--all] [--limit <n>]` — the inbox. Owner-wide by default; an optional `<agent>` scopes to one Agent. Table (`ID TYPE AGENT REQUEST STATUS EXPIRES`); `REQUEST` renders the payload (`ext_authz` → `METHOD host path`, `acp_native` → `tool: <name>`). Defaults to `pending` (the actionable set); `--status` selects one status, `--all` omits the filter. `--json` → raw `ApprovalView[]`.
- [ ] `dam approval approve <id> [--once] [--host]` — allow. **Default = permanent** (`approvePermanent`, writes a durable rule for `ext_authz`); `--once` = allow this call only (`approveOnce`, re-prompts next time); `--host` = wildcard-host rule (`approveHost`). `--once` and `--host` are mutually exclusive.
- [ ] `dam approval deny <id> [--once]` — deny. **Default = forever** (`denyForever`, writes a deny rule for `ext_authz`); `--once` = deny this call only (`dismiss`, re-prompts next time).

## API surface

Pure tRPC → [`approvals` router](../blob/main/packages/api-server-api/src/modules/approvals/router.ts) ([schemas](../blob/main/packages/api-server-api/src/modules/approvals/schemas.ts) · [types](../blob/main/packages/api-server-api/src/modules/approvals/types.ts)):

- `approvals.listForOwner({ limit?, status? })` → `list` (no agent)
- `approvals.listForInstance({ agentId, limit?, status? })` → `list <agent>`
- `approvals.approveOnce(id)` → `approve --once`
- `approvals.approvePermanent(id)` → `approve` (default)
- `approvals.approveHost(id)` → `approve --host`
- `approvals.denyForever(id)` → `deny` (default)
- `approvals.dismiss(id)` → `deny --once`

All five mutations take only `{ id }` and return `void`.

## Non-obvious behavior to surface

- **Bare `approve`/`deny` are the durable "and remember" verdicts** (`approvePermanent` / `denyForever`) — for `ext_authz` they **write an egress rule** that opens (or blocks) that host/method/path going forward and appears in `dam network list`. `--once` is the ephemeral variant that resolves only the held call and re-prompts next time. Make this explicit in `--help`: bare `approve` opens durable egress.
- **`acp_native` verdicts write no egress rule.** Tool-permission persistence lives in the harness's own config (durable via the session log), not the egress-rules table — so approving a tool call permanently is invisible to `dam network`. Only `ext_authz` approve→rule crosses over.
- **`--host` is network-only.** It writes a single `(host, *, *, allow)` rule and is meaningful only for `ext_authz`. On an `acp_native` approval the server falls back to `approvePermanent`. The mutations take only `id`, so the CLI doesn't pre-fetch to validate — document the fallback rather than erroring.
- **Actioning is at-most-once (CAS-if-pending) and best-effort idempotent.** A verdict on an already-resolved / expired / unknown id is a silent server-side no-op (no `NOT_FOUND` thrown). The mutations return `void`, so under `--json` the CLI synthesizes `{ ok: true, id, verdict }` and reports the verdict submitted; re-run `list` to see the settled state.
- **`acp_native` is live-only — the verdict is delivered, not resumed.** There's no protocol resume path for permissions; the recorded verdict reaches the agent through the relay (in-session) or the outbox (retry on re-engagement), server-side. The CLI records the decision; it does not guarantee a real-time unblock of the held turn. An `ext_authz` hold that already expired can't retroactively release the blocked call, but `approve` still writes the rule so the *next* request of that shape passes.
- **Default list is `pending`.** A small, deliberate divergence from the raw "omitted = all" default — the actionable set is what a CLI user wants. `--status resolved|expired` or `--all` widen it; the server clamps `--limit` to ≤ 500 and always orders pending first.
- **`sessionId` is informational** (always `null` for `ext_authz`) — don't key anything on it.

## Conventions to follow

Mirror the `dam network` group ([`egress/`](../blob/main/packages/cli/src/modules/egress)) and [`agent/commands/create.ts`](../blob/main/packages/cli/src/modules/agent/commands/create.ts) — thin tRPC wrapper, `--json` parity (raw `ApprovalView[]`, `[]` not `null`), integration tests in `packages/cli/src/__tests__/`. The optional `<agent>` resolves through the shared `AgentResolver` (`EXIT_AGENT_NOT_RESOLVED` (5)); `--once` + `--host` together exits `EXIT_INVALID_INPUT` (2). Render `REQUEST` via a **shared payload formatter** so CLI and UI labels match (extract/reuse the inbox's renderer, the way `network` shares `formatEgressRuleSource`).

## Out of scope (separate follow-ups under #329)

- **`--watch` / live follow** of the inbox, and any push/notification surface.
- **Bulk actions** (`approve --all`, approve-by-shape) — every action is per-id, like the UI.
- **Single `get <id>`** — no router support; use `list`.

## References

- Epic #329 · related: #345 (`dam network`)
- [ADR-035](../blob/main/docs/adrs/035-unified-hitl-ux.md) · [ADR-039](../blob/main/docs/adrs/039-cli-foundation.md) · [ADR-033](../blob/main/docs/adrs/033-envoy-credential-gateway.md) · [cli.md](../blob/main/docs/architecture/cli.md)

PetrBulanek:
Brings the unified human-in-the-loop **approval inbox** to the `dam` CLI, so headless and terminal users can clear the prompts their Agents are waiting on without opening the web UI.

Three verbs, at parity with the inbox in the web UI:

- `dam approval list [<agent>]` — show the inbox (owner-wide, or scoped to one Agent). Defaults to the actionable `pending` set; `--status resolved|expired|all` widens it. `--json` for scripting.
- `dam approval approve <id> [--once|--host]` — allow a request. Bare `approve` is durable; `--once` allows just the held call; `--host` opens the whole host.
- `dam approval deny <id> [--once]` — deny a request. Bare `deny` is durable; `--once` denies just the held call.

**How it fits with `dam network`:** `dam network` manages the standing rules — the pre-approvals that stop a prompt from ever appearing — while `dam approval` actions the prompts that did appear. Approving a network request durably writes a rule that then shows up in `dam network list <agent>`. The two are complementary: pre-seed rules to avoid prompts; action prompts when they arrive.

Durable verdicts report exactly what they wrote, and a verdict on an unknown or already-settled request fails loudly instead of silently reporting success, so scripts can trust the exit code.

Also folds in a fix for a latent bug in the existing CLI: large list/table output piped to another command (`| jq`, `| less`) could be truncated at the OS pipe buffer. All collection commands now flush before exiting.

Closes #773


packages/cli/src/modules/approval/commands/approve.ts
```diff
@@ -0,0 +1,82 @@
...
+  return new Command("approve")
+    .description("Approve a pending request from the inbox")
+    .argument("<id>", "approval id (copy from `dam approval list`)")
+    .option(
+      "--once",
+      "allow only this held call — the same request shape re-prompts next time",
+    )
+    .option(
+      "--host-wide",
+      "write a host-wide allow rule (any method, any path); network requests only",
```
JanPokorny: "host" is an ambiguous word (applies to which host? URL host? local host? the agent host?), IMO "--entire-origin" would be clearer in what it does

JanPokorny (approved):

packages/cli/src/modules/approval/commands/approve.ts
PetrBulanek: I came to the same conclusion and actually renamed it to "host-wide", but not sure about it. Your "entire-origin" might be more descriptive. What do you think?

JanPokorny: "--entire-origin" is the most descriptive name alone, alternatively "--wide" might make sense ("allow this more widely than normal allow"), but it needs explanation. So I'm still in favor of "--entire-origin", but I'll leave the call up to you.

PR #864, opened 2026-06-11
feat(crds): frozen chart CRDs with operator-applied schema upgrades (ADR-068)

jezekra1:
Implements ADR-068 (supersedes ADR-065's ownership mechanism, which merged in #818): environment deploys never change the shared CRD schema; upgrading it is an explicit operator action in the ops repo.

## What changed

- **Chart**: CRD manifests move from `templates/crds.yaml` + `files/crds/` to the chart's `crds/` directory. Helm CLI and Flux defaults never touch `crds/` on upgrade, so the freeze rests on default tool behavior. The `crds.enabled` toggle is gone.
- **Schema generation stamp**: generated CRDs carry `agent-platform.ai/crd-schema-generation` (from a `+kubebuilder:metadata:annotations` marker, mirrored by Go constants; a unit test fails if they drift from the generated manifests).
- **Controller startup assert**: every replica refuses to start when the live CRD's generation is older than the one it was built against — loud crash with the remedy, instead of admission silently pruning writes. Backed by a `resourceNames`-scoped ClusterRole (`get` on the two CRDs).
- **Gate**: `controller:check:crd-backwards-compatibility` resolves the baseline at the pre-move `files/crds/` path for older tags, and now also fails a schema change that forgets the generation bump.
- **Local dev**: `cluster:install` server-side-applies `crds/` so warm VMs keep getting schema updates; `helm:check:render` renders with `--include-crds`.

## Companion change (dam-ops)

Branch `feat/operator-crd-upgrades` in dam-ops adds `upgrade-crds.sh` (fetch released chart → crdify against live cluster → diff → confirm → server-side apply), pins `install.crds: Create` / `upgrade.crds: Skip` on all env HelmReleases, and removes the now-dead `crds.enabled` values.

## Rollout note

Once an environment deploys a controller built from this change, the operator script must have applied the generation-1 CRDs first — existing clusters' CRDs carry no generation annotation (= 0), so the controller will crash-loop until the upgrade is run. The migration off the templated CRDs is safe: the `keep` resource policy orphans them on the owning release's first upgrade instead of cascade-deleting CRs.


deploy/tasks.toml
```diff
@@ -425,6 +425,10 @@ if [ "$ROOT_FSTYPE" = "btrfs" ]; then
 fi
 
 echo "Installing/upgrading Platform chart..."
+# Helm never upgrades the chart's crds/ on an existing release (ADR-068) — on
+# this single-env local cluster the freeze has no value, so apply them directly.
+kubectl --kubeconfig="$KUBECONFIG" apply --server-side --force-conflicts -f deploy/helm/platform/crds/
```
JanPokorny, line 432: Does this properly wait? We don't want a race with the helm chart install

JanPokorny (approved):

deploy/tasks.toml
jezekra1: Good catch on the semantics: `kubectl apply` only guarantees the CRDs are persisted, not `Established`. There's no actual race for this chart today — it deliberately ships no Agent/Fork CR instances (agent templates stay ConfigMaps, which is the documented ADR-058 design), and on a fresh cluster helm additionally installs `crds/` itself and waits for establishment before rendering the rest. But that's an invariant rather than a guarantee, so 692f2486 adds an explicit `kubectl wait --for=condition=Established` right after the apply — the ordering no longer depends on the chart staying CR-free. Same wait added to the dam-ops operator script (dam-ops#36).

PR #866, opened 2026-06-11
build(cd): version main chart builds as <version>-main.<run>

jezekra1:
Main-branch chart builds are currently versioned `0.0.0-main.<run>`, which sorts below every release in semver — so any registry version resolution (e.g. the dam-ops `upgrade-crds.sh` default, dam-agents/dam-ops#36) can never consider them, even though they are valid published versions and usually the newest content.

This versions them as `<in-development version>-main.<run>` (e.g. `0.2.2-main.359`, read from Chart.yaml, which `common:check:version` keeps as the single source). Main builds then sort above the last release and below the next one — semver "latest" resolution ranks them correctly with no registry-API tricks.

Unchanged: the floating `0.0.0-main` alias (dev's Flux `OCIRepository` tracks it) is still pushed; release tag publishing is untouched.

Known caveat: between cutting an rc and merging the `release:new` main version bump, rc tags of the same base version still outrank main builds (`main` < `rc` alphabetically in prerelease ordering). The window is short and the dam-ops downgrade guard makes misordering harmless.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -307,8 +307,12 @@ jobs:
           RUN_NUMBER: ${{ github.run_number }}
           COMPOSITE: ${{ needs.appversion.outputs.composite }}
         run: |
-          helm package deploy/helm/platform --version "0.0.0-main.${RUN_NUMBER}" --app-version "$COMPOSITE"
-          helm push "platform-0.0.0-main.${RUN_NUMBER}.tgz" oci://quay.io/dam-agents/charts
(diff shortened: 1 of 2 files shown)


jezekra1:
Withdrawn — we're keeping the 0.0.0-main.* versioning. The dam-ops upgrade script now resolves the newest published chart by registry push date instead (dam-agents/dam-ops#36), which handles main builds vs releases without touching the version scheme. Renumbering would also have made the staging/prod Flux semver ranges (>=0.2.x-0) match main builds — avoided entirely this way.

Issue #601, opened 2026-06-04
Move provider & PAT setup from secrets to Connections


PetrBulanek:
## Context

Part of #329. Two flows still set up credentials through the soon-to-be-retired secrets module: the CLI's `dam agent create-interactive` (provider + GitHub PAT steps) and the UI's Providers settings page. Move both onto Connections.

Parity has landed (#884 closed #600's template gaps), so this is unblocked. Moving these setup flows is the precondition for the data migration and secrets removal tracked in #1273.

## Scope

- [ ] **Add a "replace key" path to Connections** — today a connection can only be created or deleted; replacing a credential's value needs an update path, since delete-then-recreate would drop existing agent grants. Required to preserve today's UX.
- [ ] `dam agent create-interactive` — provider pick/create, optional GitHub PAT, and the grant step go through Connections instead of secrets (reusing the existing CLI connection service).
- [ ] UI provider setup — re-backed by Connections (reusing the existing flows).

## References

#329 · #346 · #600 · #884 · cutover #1273 · ADR-051

jjeliga:
def see the #600 and corresponding merge request (comments) to get a better picture before starting on this.

Issue #600, opened 2026-06-04
Bring Connections to parity with the secrets module


PetrBulanek:
## Context

We're consolidating all injected credentials onto the **Connections** model (ADR-051) and retiring the legacy **secrets** module. Before secrets can go away, Connections has to handle everything the provider presets and GitHub PATs do today. Five things are still missing.

## Gaps

- [x] **Anthropic OAuth** — only the API-key variant exists as a connection; the OAuth-token variant doesn't. *(done in #884)*
- [x] **IBM LiteLLM** — the connection template ships an incomplete env bundle (and some stale values/host); it needs to match what the provider preset configures, including the per-connection model-pin overrides offered today. *(done in #884)*
- [x] **Bob** — needs the dual injection (header **+** URL query param on the same host) that the preset does; the Connections model can't express query-param injection yet. Also a credential env-var name mismatch to reconcile. *(done in #884)*
- [x] **GitHub PAT** — only GitHub *OAuth* exists; there's no personal-access-token connection. (The OAuth template already proves the multi-host injection it needs.) *(done in #884)*
- [ ] **Cut over + accept the ADR** — ADR-051 is still Proposed and the secrets write path is still active. Accept it, migrate existing data, and retire the secrets module. → **split out to #1273** (it's the cutover, not parity).

## Outcome

Every credential a user could set up via the old providers/secrets flows is expressible as a Connection, unblocking removal of the secrets module.

## References

#329 · #884 · cutover #1273 · supersedes #346, #236 · ADR-051, ADR-028

PetrBulanek:
Closing as **parity done**. #884 landed all four parity gaps (Anthropic-OAuth, IBM LiteLLM, Bob header+query-param injection, GitHub-PAT), so every provider/PAT credential is now expressible as a Connection — this issue's stated outcome.

Gap 5 (cut over + accept ADR-051) is the *retirement*, not parity, so it's been split into a dedicated issue: #1273. The setup-flow move that precedes it remains #601.

Chain: parity (#884, done) → move setup flows (#601) → migrate + retire secrets (#1273).

Issue #329, opened 2026-05-20
Epic: CLI phase 2 - management commands


jenna-winkler:
## Context

The CLI currently supports core workflows - agent creation, authentication, chat interaction, bulk file import, and template listing. The UI exposes a full set of management operations that the CLI does not yet cover. This epic brings the CLI to feature parity with the UI for all CRUD and lifecycle operations.

## Scope

All management operations available in the UI that are backed by existing tRPC routers but not yet exposed in the CLI.

### P0 - Required for basic agent configuration
  
- Connections - Create, list, delete, and sync OAuth provider connections. Agents can't reach authenticated APIs without this.
- Egress rules - Manage which external hosts agents are allowed to reach.
- Secrets - Create, update, delete, and manage access grants for API keys and PATs injected into agent environments. Distinct from OAuth connections.
  
### P1 - Core user workflows

- Skills - Install, uninstall, update, and list git-sourced knowledge packages mounted into agent pods.

### P2 - Automation & integrations

- Schedules - Full lifecycle management of cron-triggered agent runs.
- Approvals - List and action pending human-in-the-loop tool-call approvals.
- Channels - Connect and manage external messenger integrations (Slack, Telegram) that allow users to interact with agents outside the platform UI.

## Open questions

- Granular file operations - The UI supports individual file read/write/rename/remove on agent PVCs. Should this be supported in CLI?
- Agent update - The UI allows updating agent metadata and env vars after creation. Should this be a new addition to the existing CLI agent module?
- Session management - Should users be able to manage sessions from the CLI?

tomkis:
@PetrBulanek can you please create followup tickets so that we can tackle the stuff separately?

PetrBulanek:
## P2 sub-issues filed

Decomposed P2 (Automation & integrations) into one sub-issue per domain:

- #762 — **Manage schedules from the CLI** (`dam schedule` — RRULE create/edit, quiet hours, enable/disable, reset-session)
- #773 — **Action pending approvals from the CLI** (`dam approval` — unified inbox: list + approve/deny for both network and tool-call holds)
- #776 — **Manage messenger channels from the CLI** (`dam channel` — Slack/Telegram connect/disconnect, allowed-users)

[...]

PR #884, opened 2026-06-11
feat(connections): bring Connections to parity with the secrets module (#600)

jjeliga:
## What & why

We're consolidating all injected credentials onto the **Connections** model (ADR-051) and retiring the legacy **secrets** module (#329). Before secrets can go away, every credential the provider presets and GitHub PATs configure today must be *expressible* as a Connection. This PR closes the four parity gaps from #600. The cutover itself (gap 5) is deliberately **not** here — see *Scope* and *Migration path*.

## What's in this PR

Maps to #600's checklist:

- **Anthropic OAuth** — new `anthropic-oauth` template (`CLAUDE_CODE_OAUTH_TOKEN`, `Authorization: Bearer {value}`), alongside the existing API-key variant.
- **IBM LiteLLM** — template now reuses the legacy registry's *exact* env bundle + host (`ibmLitellmEnvMappings()`, `IBM_LITELLM_HOST`) instead of an incomplete/stale set, and supports the per-connection model-pin overrides via config inputs.
- **Bob** — dual injection on a single host: `Authorization: Apikey {value}` **plus** a URL query-param (`?key=`), reusing the existing Envoy Lua filter. This required teaching the Connections model query-param injection (below). Reconciles the credential env-var name, and surfaces Bob's pins (model / instance / team / max-coins / chat-mode) as config inputs.
- **GitHub PAT** — new `github-pat` header template: `GH_TOKEN` env + `Authorization: Bearer` on `api.github.com` & `raw.githubusercontent.com`, and `Basic` (`x-access-token` encoding) on `github.com`. ✅ validated end-to-end on a live cluster.

### Enabling infrastructure

- **Query-param injection in the Connections model.** The wire schema gains an optional `queryParamName` on the egress-inject contribution (`agent-runtime-api`). When a host carries more than one injection, the SDS file key becomes per-`(host, header)` (base64url of `host\nheader`), with a **byte-identical TS↔Go** implementation; the controller's dedup relaxes from per-host to per-`(host, header)` so stacked injectors don't clobber, and the Envoy bootstrap stacks them as an ordered list. **Backward-compatible:** `sdsKey` is optional and the controller falls back to the legacy per-host key for existing Secrets.
- **`egress-allow` → `egress-inject`** for the provider templates — at parity these must actually inject (the prior `egress-allow` injected nothing).
- **Config inputs** — a header template can declare optional inputs (label / hint / pattern / enum); each filled input projects into an `env` contribution. Used for Bob's pins and IBM's model overrides.
- **CLI** — `dam connection connect` gains `--env-name` and config-input collection.
- **Docs** — `connections.md` (config inputs) and `security-and-credentials.md` (stacked per-host injectors).

## Scope — what is intentionally NOT here

Gap 5 of #600 (**cut over + accept the ADR**) is out of scope. Concretely:

- The provider templates (anthropic, anthropic-oauth, openai, ibm-litellm, bob) are **hidden in the Connections UI** (`LEGACY_PROVIDER_TEMPLATE_IDS`) so they aren't offered twice; the legacy Providers view remains the active setup path. They're reachable via CLI/API for testing.
- The **secrets module is still the active path** for providers. ADR-051 stays Proposed; no data is migrated and the secrets write path is untouched.

This keeps the PR to a single cohesive change: *parity*, not *cutover*.

## Effect on existing agents — no migration needed

Backward-compatible by construction:

- Agents on legacy provider/secrets are untouched — same `secret-type=anthropic` Secrets, same `host-pattern` expansion, byte-identical gateway bootstrap (no roll).
- Agents on existing connection grants keep working via the `sdsKey` fallback (old Secrets have no `sdsKey` → controller uses the legacy per-host key, exactly the filename already on disk); the relaxed dedup is identical in result for the one-header-per-host shape every existing Secret has.
- No DB migration — the schema additions are optional JSON fields, so existing rows validate unchanged.
- The one nuance, **not a regression**: a provider *connection* record created before this PR keeps its stored `egress-allow` (still injects nothing, as before) until re-created — but those are effectively nonexistent because the UI hid them.

Credential env is delivered over the runtime channel, not baked into the pod spec, so no pod's env changes on upgrade and nothing needs recreating.

## Future migration path (→ #601, and #600 gap 5)

This PR is the parity precondition that unblocks the cutover. The follow-up work:

**#601 — move the setup flows onto Connections.** Re-back `dam agent create-interactive` (provider + PAT steps + grant) and the UI Providers page with Connections. #601 surfaces a hard prerequisite this PR does *not* add: a Connections **"update" path** to replace a credential's value — today it's create/delete only, and delete-then-recreate would drop existing agent grants. That "replace key" UX must land before existing setups can move.

**#600 gap 5 / #329 — retire secrets.** A one-shot data migration:
1. Each legacy provider/secret → an equivalent Connection via its parity template — **lossless precisely because this PR reached full parity** (e.g. Bob's pins survive as config inputs rather than being dropped).
2. Per-agent `grantedSecretIds` → `grantedConnectionIds`.
3. Transfer the credential value into the Connection's secret store, without surfacing plaintext.
4. **Atomic per agent.** A legacy anthropic Secret and an anthropic Connection both target `api.anthropic.com` / `Authorization`; the per-`(host, header)` dedup silently picks one — so create-connection + flip-grant + drop-secret must happen together to avoid a wrong-credential / no-credential window.
5. **After** migration: delete the controller's legacy `host-pattern` expansion branch and the hardcoded anthropic env-mappings fallback; accept ADR-051.

## Testing

- `github-pat` validated end-to-end on a live cluster: in-pod `echo $GH_TOKEN` → `dummy-placeholder` (the real PAT never enters the pod), and `gh api user` authenticates via the gateway-injected token.
- The other new templates are built by reusing the legacy registry's exact env bundles / hosts, so parity is structural; they're UI-hidden and not yet exercised end-to-end (testable via CLI/API).
- Automated tests aren't included here; the TS↔Go `sdsFileKeyForInjection` byte-identity is the highest-value thing to pin with a test in the cutover work.

---

Advances #600 — parity gaps 1–4 complete; cutover (gap 5) tracked in #601 and #329.

### DIFF
--- a/packages/api-server-api/src/modules/connections/types.ts
+++ b/packages/api-server-api/src/modules/connections/types.ts
@@ -92,6 +92,10 @@ export const templateInput = z.object({
   state: templateInputState,
   presetValue: z.string().optional(),
   secret: z.boolean().optional(),
+  // Marks a config input the form packs into `configInputs` rather than the typed auth fields.
+  configInput: z.boolean().optional(),
+  label: z.string().optional(),
+  hint: z.string().optional(),
 });
 export type TemplateInput = z.infer<typeof templateInput>;
 
--- a/packages/api-server/src/modules/connections/domain/catalog.ts
+++ b/packages/api-server/src/modules/connections/domain/catalog.ts
@@ -1,9 +1,27 @@
+import {
+  type Contribution,
+  type EnvMapping,
+  ibmLitellmEnvMappings,
+  bobEnvMappings,
+  BOB_CHAT_MODES,
+  IBM_LITELLM_HOST,
+  BOB_HOST,
+} from "api-server-api";
 import type {
   ConnectionTemplate,
   HeaderConnectionTemplate,
   NoneConnectionTemplate,
   OAuthConnectionTemplate,
 } from "./connection-template.js";
+
+// Project a provider preset's env bundle into `env` contributions
+function envContributions(mappings: EnvMapping[]): Contribution[] {
+  return mappings.map((m) => ({
+    kind: "env" as const,
+    name: m.envName,
+    placeholder: m.placeholder,
+  }));
+}
 
 export interface OAuthClientCredentials {
   clientId: string;
@@ -28,7 +46,7 @@ export interface OperatorCredentials {
 
 const ANTHROPIC: HeaderConnectionTemplate = {
   id: "anthropic",
-  name: "Anthropic",
+  name: "Anthropic (API Key)",
   category: "app",
   isCustom: false,
   description: "Anthropic API access (Claude). Sent as `x-api-key`.",
@@ -43,7 +61,39 @@ const ANTHROPIC: HeaderConnectionTemplate = {
       name: "ANTHROPIC_API_KEY",
       placeholder: "dummy-placeholder",
     },
-    { kind: "egress-allow", host: "api.anthropic.com" },
+    {
+      kind: "egress-inject",
+      host: "api.anthropic.com",
+      headerName: "x-api-key",
+      valueFormat: "{value}",
+    },
+  ],
+};
+
+const ANTHROPIC_OAUTH: HeaderConnectionTemplate = {
+  id: "anthropic-oauth",
+  name: "Anthropic (OAuth Token)",
+  category: "app",
+  isCustom: false,
+  description:
+    "Anthropic API access (Claude) via an OAuth token. Sent as `Authorization: Bearer`.",
+  iconSlug: "anthropic",
+  authKind: "header",
+  host: "api.anthropic.com",
+  headerName: "Authorization",
+  valueFormat: "Bearer {value}",
+  contributions: [
+    {
+      kind: "env",
+      name: "CLAUDE_CODE_OAUTH_TOKEN",
+      placeholder: "dummy-placeholder",
+    },
+    {
+      kind: "egress-inject",
+      host: "api.anthropic.com",
+      headerName: "Authorization",
+      valueFormat: "Bearer {value}",
+    },
   ],
 };
 
@@ -61,9 +111,11 @@ const OPENAI: HeaderConnectionTemplate = {
   contributions: [
     { kind: "env", name: "OPENAI_API_KEY", placeholder: "dummy-placeholder" },
     {
-      kind: "egress-allow",
+      kind: "egress-inject",
       host: "api.openai.com",
       pathPattern: "/v1/*",
+      headerName: "Authorization",
+      valueFormat: "Bearer {value}",
     },
   ],
 };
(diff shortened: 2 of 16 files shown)


packages/api-server/src/modules/connections/domain/catalog.ts
```diff
@@ -43,7 +61,39 @@ const ANTHROPIC: HeaderConnectionTemplate = {
...
+    {
+      kind: "egress-inject",
+      host: "api.anthropic.com",
+      headerName: "x-api-key",
+      valueFormat: "{value}",
+    },
+  ],
+};
+
+const ANTHROPIC_OAUTH: HeaderConnectionTemplate = {
```
jezekra1, line 73: is this working and supported path? I thought for anthropic the official way is to use `claude setup-token`, we don't want users to get banned on anthropic due to some of their 3rd party detection :D But if it works it's more seamless than the extra step 🤷

packages/api-server/src/modules/connections/domain/catalog.ts
```diff
@@ -108,17 +155,60 @@ const BOB: HeaderConnectionTemplate = {
...
+      valueFormat: "{value}",
+      queryParamName: "key",
+    },
+  ],
+  configInputs: [
+    {
+      inputName: "model",
+      envName: "BOB_SHELL_MODEL",
+      label: "Model",
+      hint: "Default model for new sessions. Empty → Bob's built-in default.",
```
jezekra1, line 184: Nitpick: I am not really sure that this should be configured on the connection, don't you want to change models for each agent?

I think ultimately we want something like "configuration contribution" that the user can setup in the UI, essentially persist the changes in this config menu to `.claude/settings.json` (for claude and different driver for bob)

Image


This is more of a point for discussion than a blocker for this PR, it's good as is as long as it's not required.

packages/api-server-api/src/modules/connections/types.ts
```diff
@@ -92,6 +92,10 @@ export const templateInput = z.object({
   state: templateInputState,
   presetValue: z.string().optional(),
   secret: z.boolean().optional(),
+  // Marks a config input the form packs into `configInputs` rather than the typed auth fields.
+  configInput: z.boolean().optional(),
```
jezekra1, line 96: nitpick: this does not seem generic enough for multiple configuration types (e.g. file contributions to store a result in .gitconfig or .claude/settings.json)

a "kind" discriminator would be one solution, but let's refactor it once we have this use-case, this is just an api change

jezekra1 (approved):

packages/api-server/src/modules/connections/domain/catalog.ts
jjeliga: It should work the same way we currently use it. But good point, we should be careful and test this properly during the followup migration task

packages/api-server/src/modules/connections/domain/catalog.ts
jjeliga: This just surfaces optional bob settings that are currently available in the advanced settings.

PR #901, opened 2026-06-12
fix(helm): gate agent pod ingress to api-server and controller

jezekra1:
## Problem

`agent-runtime` serves unauthenticated ACP, tRPC, terminal, and SSH endpoints on `:8080`, with an explicit comment stating that a kernel-level ingress NetworkPolicy is the auth boundary (`packages/agent-runtime/src/server.ts`). **That policy was never rendered** — any pod in the cluster that can route to the agent namespace (gateway pods included, which have unrestricted egress) could dial an agent's `:8080` directly.

## Change

Adds a chart-rendered `agent-ingress-platform-only` NetworkPolicy in the agent namespace. It selects `role=agent` pods (long-lived agents and forks; gateway pods untouched) and admits TCP 8080 only from:

- **api-server** pods in the release namespace — the ACP/tRPC/terminal/SSH relay; it verifies user JWT and agent ownership before forwarding
- **controller** pods in the release namespace — the idle-checker busy-probe (`GET /api/status`, `idlechecker.go`). Note: blocking this probe would not fail loudly — a dropped probe parses as *busy*, silently disabling hibernation, which is why the controller must be admitted

Kubelet `/healthz` probes are node-originated and unaffected by ingress policy.

Mesh interplay: agents are out of the ambient mesh, so in-mesh sources (api-server, controller) reach them as plaintext passthrough with pod source IPs preserved — no HBONE/15008 rule needed, and the egress credential boundary is untouched.

Also updates the `server.ts` comment (api-server → api-server *and controller*) and the security-and-credentials architecture page.

## Testing

- `mise run helm:check:lint` and `helm:check:render` (kubeconform) pass; `mise run check` passes.
- No new e2e spec: the allow paths are exercised by the entire existing suite (every test relays through the api-server, and agents only go Ready if kubelet probes pass). A deny-path test needs an in-cluster client pod the Playwright suite can drive, which doesn't exist today — `e2e.performFetch` goes through the gateway's egress gate, so it would test the wrong layer.

### DIFF
--- a/deploy/helm/platform/templates/agent-ingress-np.yaml
+++ b/deploy/helm/platform/templates/agent-ingress-np.yaml
@@ -1,0 +1,42 @@
+{{- /*
+Ingress gate for agent pods: only the api-server (ACP/tRPC relay) and the
+controller (idle-checker busy-probe) may dial the agent port. agent-runtime
+serves unauthenticated on the assumption that this policy is the auth
+boundary (packages/agent-runtime/src/server.ts). Kubelet probes are
+node-originated and unaffected. Role-scoped to `agent`, so fork pods are
+covered and gateway pods are untouched.
+*/}}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: agent-ingress-platform-only
+  namespace: {{ .Values.agentNamespace }}
+  labels:
+    {{- include "platform.labels" . | nindent 4 }}
+    agent-platform.ai/managed-by: platform-helm
+    agent-platform.ai/role: agent
+spec:
+  podSelector:
+    matchLabels:
+      agent-platform.ai/role: agent
+  policyTypes:
+    - Ingress
+  ingress:
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: {{ .Release.Namespace | quote }}
+          podSelector:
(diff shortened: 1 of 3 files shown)


Tomas2D (approved):

deploy/helm/platform/values.yaml
```diff
@@ -111,6 +111,16 @@ agentNamespace: platform-agents
...
 
+# -- Ingress gate for agent pods (chart-rendered `agent-ingress-platform-only`).
+# Admits the agent port (8080) only from api-server (ACP/tRPC relay) and
+# controller (idle-checker busy-probe) pods. agent-runtime serves
+# unauthenticated on the assumption that this kernel gate is the auth boundary
+# (packages/agent-runtime/src/server.ts), so disabling it removes the agent
+# port's only access control — turn off only in environments without
+# NetworkPolicy support (e.g. some bare clusters).
+agentIngressNetworkPolicy:
+  enabled: true
```
pilartomas: Do we even want to support such config?

pilartomas (approved):

deploy/helm/platform/values.yaml
jezekra1: I added it as a backup in case it blocks something we didn't think about, but good point, let's remove it, worst case we catch it in staging and fix it

PR #902, opened 2026-06-12
fix(schedules): wake hibernated agents on schedule fire

jezekra1:
## Problem

A schedule fire against a hibernated agent never runs. The fire path commits a `trigger` event to the runtime outbox and enqueues delivery, but the delivery worker dispatches only to a `Ready` agent and nothing wakes the pod:

- the worker defers to the sweep when the agent isn't Ready (`worker-handler.ts`), and the sweep just re-enqueues — neither ever wakes
- the trigger event expires after its TTL (default 1 h) and is silently dropped by the sweep
- meanwhile `recordFire` stores `success` and a `ScheduleFired` event with `outcome: "success"` is emitted, because the commit succeeded — so the UI shows a successful fire that never executed

Every other caller that sends *work* to a pod (ACP relay, channels, skills management) wakes it first; schedules lost that step when trigger delivery moved from the controller's `kubectl exec` loop to the runtime outbox. Config contributions can wait for the next wake — TTL'd trigger events can't.

## Fix

`fire()` now pokes the agent awake **after** committing the trigger event, via `wakeIfHibernated` — the fire-and-forget activity bump, not the blocking `ensureReady` wait:

- commit first, poke second — a crash or failed poke never loses the event; it stays in the outbox and still delivers if the agent comes up within the TTL
- the poke doesn't wait for `Ready`, so the schedule worker never blocks on a cold pod; the boot-time `hello` catch-up (`retryUntilReady`) delivers the event seconds after the pod is up, with the sweep as backstop
- a failed poke (e.g. K8s API error) is recorded as a failed fire in the schedule status; the schedule re-arms for its next occurrence regardless of outcome, as before

Wired through `composeSchedulesAtBoot` as an injected `wakeAgent` port; no change to the delivery worker — it stays wake-free by design.

## Note for reviewers — overlap with #900

#900 (docs) describes the current no-wake behavior as-is, including "Schedule fires do not wake … or expires". This PR changes that behavior, so whichever lands second needs a one-paragraph reconciliation in `agent-lifecycle.md`: schedule fires poke the agent awake again (as the pre-#900 page claimed), but via outbox-commit-then-poke rather than the retired controller exec path. The doc lines updated here are written to read correctly against either base.

## Testing

- new unit tests for the fire path: commit-before-poke ordering, and poke-failure → failed fire recorded + schedule still re-armed
- `mise run api-server:check` and `mise run api-server:test` (195 tests) pass

### DIFF
--- a/docs/architecture/agent-lifecycle.md
+++ b/docs/architecture/agent-lifecycle.md
@@ -76,10 +76,10 @@ Contributions are applied out-of-band by a single background worker (a pod's `he
 Three paths trigger a wake:
 
 - **Connect-driven** — the api-server is about to forward an ACP frame to a hibernated Agent and ensures readiness before the relay completes. The frame can originate from a UI tab attaching to a session or from a channel worker (Slack / Telegram) routing an inbound message to its bound session.
-- **Schedule-driven** — the controller's schedule loop is about to deliver a trigger and `kubectl exec` requires the pod to be running.
+- **Schedule-driven** — a schedule fire commits a `trigger` event to the runtime outbox, then pokes the Agent awake without waiting for readiness; the boot-time `hello` catch-up delivers the event once the pod is `Ready`, and the event's TTL bounds how stale a fire can land.
 - **Skills-management-driven** — install / uninstall / private-source scan / publish all route through the same primitive before reaching the agent (scan and publish reach agent-runtime directly over the harness port; install/uninstall keep the pod warm so the apply worker dispatches the bumped outbox). See [skills](skills.md).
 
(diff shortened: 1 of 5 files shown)


tomkis (approved):

tomkis:
I wrote a near identical fix on #900 before spotting this PR, so I dropped mine and this one should land. Two things from my version worth considering:

- resetSession has the same hole: a schedule-reset event against a hibernated agent expires after the same TTL, so the next fire resumes the old session anyway. The same poke after the bump fixes it.
- Recording a failed poke as a failed fire can mislead: the event is still durable and runs if the agent comes Ready within the TTL. lastResult has so far meant committed, not ran, so I treated the poke as best effort and only logged the failure.

PR #903, opened 2026-06-12
chore: hide slack, yt, and google,  connections from the ui

### LINKED ISSUE #891: Hide internal-only connections from regular users
## Problem

The production UI shows every connection the platform supports. For our early users that includes several they have no use for: Spotify, the Google services, YouTube, Slack, and Telegram. They appear in the connections list and in the connection options shown during chat.

This clutters the first-run experience and invites users to set up integrations we don't intend to support for them yet. At the same time, we can't remove these connections, because the team relies on them for internal testing on the same deployment.

## Proposed solution

Each connection in the catalog is classified as either generally available or **internal-only**. Regular users only see the generally available ones, everywhere connections are offered: the connections list, agent configuration, and the chat connection options.

Users marked as **internal testers** see the full catalog. Whether someone is an internal tester is a role on their user identity in Keycloak, assigned by an admin. It's one switch that reveals all internal-only connections at once, not a per-feature flag.

The initial internal-only set is Spotify, all Google services, YouTube, Slack, and Telegram. The classification should be easy to change per deployment as connections graduate to general availability.

### Out of scope

- Gating features other than connections behind the internal role. Nothing else needs hiding today; the role can be reused later if that changes.
- Existing connections users may have already created. This only affects what's offered, not what exists.

---

Reported on the ops side as https://github.ibm.com/dam-agents/dam-ops/issues/38.

jjeliga:
Minimal UI-only design.

Persists only in the given browser where function was called

Not tied to the user identity in any way.

Closes #891

### DIFF
--- a/docs/architecture/connections.md
+++ b/docs/architecture/connections.md
@@ -77,7 +77,7 @@ Beyond the auth credential, a template may declare optional **config inputs** â€
 
 #### Internal-only templates
 
-Some templates (initially Spotify, Slack, YouTube, and all Google services) are hidden from regular users client-side, affecting only what's offered, not Connections already created. Testers reveal the full catalog by running `platformConnections.showInternal()` in the browser devtools console.
+Some templates (initially Spotify, Slack, YouTube, and all Google services) are hidden from regular users client-side, affecting only what's offered (on both the Connections settings page and the sandbox creation wizard), not Connections already created. Testers reveal the full catalog by running `platformConnections.showInternal()` in the browser devtools console, or by tapping the version string on the settings page five times.
 
 ### Connection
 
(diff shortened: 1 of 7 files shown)


docs/architecture/connections.md
```diff
@@ -72,7 +72,13 @@ Two display-axis attributes drive UI grouping:
 
 Templates are registered in code; adding a new integration is one entry. Schemas validate user input; the template's `build()` function projects inputs into the concrete `auth` + `contributions[]` of the Connection record.
 
+<<<<<<< HEAD
```
Tomas2D: incorrect merge

jjeliga: fixed

jezekra1 (approved):

Issue #895, opened 2026-06-12
Redesign — full-page sandbox settings page


PetrBulanek:
## Problem

Changing an existing sandbox today is split across a modal with three tabs (connections, environment, network) and a *separate* full page for network rules. A user who wants to review or adjust everything about a sandbox has to hop between a popup and a page, and there's no single place that shows the whole configuration at once.

## Proposed solution

A single full-page sandbox settings view, reached from the list, with a "back to sandboxes" link and everything on one scrollable page:

- **Name**, **Image**, and **Provider**.
- **My connections** — the app connections/credentials this sandbox can use, each connectable/disconnectable inline.
- **Network access** — a preset selector plus the full rule editor: per-rule host, method, path, and allow/deny, with each rule tagged by where it came from (a preset, a connection, or added by hand). Changing the preset replaces preset-derived rules while preserving hand-written and connection-derived ones.
- **Environment** — inherited variables (shown read-only with a badge for their source) and custom variables the user can add.

### Scope

- The full edit page consolidating what's today the configure modal + the standalone network page.
- A clear save model (whether edits apply live or are staged and saved together).

### Dependencies

- Blocked on #892 (shell; the standalone network route folds into this page).
- Reuses connection UI from #894 (create wizard).

### Out of scope

- Creating a new sandbox (#894).

### Open questions

- Can **Image** and **Provider** actually be changed on an existing sandbox in place, or does changing them mean recreating it? The page presents them as editable dropdowns — we need to confirm what's supported.
- Should edits apply immediately or be batched behind an explicit Save? (The current network page applies live; the current modal stages changes — the redesign needs one consistent model.)

### Design reference

Figma page **"#356 - Agent creation user flow"**: edit page `152:6940` and `152:7204`.

PetrBulanek:
**Implementation orientation** — code pointers / reuse map for whoever picks this up. Not part of the ticket spec.

- **Consolidates:** `src/modules/agents/dialogs/configure-agent-dialog.tsx` + `components/configure-agent/*` (connections / env tabs) and `src/modules/egress-rules/views/agent-egress-view.tsx` + `agent-egress-editor.tsx` (the standalone network page + its editor, which already supports both live and staged modes).
- **Reuse:** `src/components/connections-picker.tsx`, `src/components/env-vars-editor.tsx` / `key-value-editor.tsx`, `forms/configure-agent-schema.ts`.
[...]

Issue #894, opened 2026-06-12
Redesign — full-page sandbox creation wizard


PetrBulanek:
## Problem

Creating a sandbox today happens two different ways — a cramped modal on the old surface and a separate two-step wizard on the new one — and neither matches the redesign. Setting up a provider and granting connections are also separate chores a user has to do *before* creating anything, rather than as part of creation.

## Proposed solution

One full-page, three-step wizard with a persistent step indicator down the side:

- **Step 1 — Image.** Pick a pre-built image (Bob, Claude Code, Codex, Pi Agent) or, under an "advanced" option, bring your own container image.
- **Step 2 — Setup.** Name the sandbox; choose a **provider** and connect it inline — clicking "Connect" on a provider opens a small dialog to enter the credential (with an "advanced — model overrides" option), after which the provider shows a "Connected" state; then choose a **network access** level via three presets (strict deny / trusted defaults (recommended) / allow everything).
- **Step 3 — Grant connections.** Choose which app connections and credentials this sandbox can use, grouped as **Apps** (GitHub, GitHub Enterprise), **MCP servers** (OAuth and no-auth), and **Other** (a custom header credential). Already-configured connections appear in a "My connections" group that can be toggled on; new ones can be added inline (which may bounce through an external OAuth screen). This step can be **skipped**. Finishing creates the sandbox.

The wizard must survive the round-trip to an external OAuth provider mid-flow without losing what the user has entered.

### Scope

- All three steps, the step indicator, the inline provider-connect dialog, and the inline connection-add flow.
- Carrying over capabilities the old creation modal has that the designs don't re-draw but that users rely on (see open questions).

### Dependencies

- Blocked on #892 (shell + the three-step-flow decision).
- Shares connection UI with #895 (edit page) — the "My connections" / add-connection pieces should be built to be reused.

### Out of scope

- The external OAuth provider screens themselves (GitHub, Google) — those are the providers' own pages, shown as placeholders in the designs.
- Editing a sandbox after creation (#895).

### Open questions

- **Pi Agent** is offered as an image but isn't currently part of the provider/credential model — which provider(s) does it pair with?
- Are the four pre-built images presented as curated **templates** or as raw image references? This affects how Step 1 binds to the rest.
- Connecting a provider mid-wizard: does it create a **reusable** provider credential (that then shows up in Settings → Providers) or one scoped to this sandbox?
[...]

PR #922, opened 2026-06-12
feat(ui): unified Sandboxes home

### LINKED ISSUE #893: Redesign — unified Sandboxes home
## Problem

Returning users today land on one of two different list pages depending on which surface they entered through, with different layouts and different language for the same underlying thing. There's no single "home" that shows everything a user has running.

## Proposed solution

A single **Sandboxes** home page: a clear page title, a primary **Create sandbox** action, and a list of the user's sandboxes. Each row shows the sandbox name, a short subtitle describing what it is (its base image and the provider powering it, e.g. "Claude Code · IBM LiteLLM"), and a status indicator. A per-row overflow menu offers the common actions: **Refresh**, **Configure**, **Delete**.

Status must include a clearly distinguished **Running** and **Hibernating** state (sandboxes that have gone to sleep), in addition to the transitional/error states already shown today.

### Scope

- The list page, its rows, the status indicator, and the row overflow menu.
- Empty state and the hook into first-run (a brand-new user with no sandboxes — see #896).
- Preserving the useful behaviours the current list already has (see open questions) so nothing regresses when the two lists merge.

### Dependencies

- Blocked on #892 (navigation shell + the "Sandbox" naming and surface-merge decisions).

### Out of scope

- Creating a sandbox (#894) and configuring an existing one (#895) — this issue only opens those flows, it doesn't implement them.

### Open questions

- The current list surfaces things the newer one doesn't: a warning badge when an agent has failed to contribute its setup, a restart/wake action, and an "unavailable" overlay. Which of these carry over, and how do they fit the new row + overflow design?
- Does the row subtitle ("image · provider") always have both values available to show?

### Design reference

Figma page **"#356 - Agent creation user flow"**: returning list `152:6453`, list with slim rail `152:4567`.

PetrBulanek:
The home page is now a single **Sandboxes** list for everything you have running. Each sandbox shows its name, a short subtitle describing what it is (e.g. "Claude Code · IBM LiteLLM"), and a status pill that clearly distinguishes **Running** from **Hibernating**. A per-row menu offers the common actions — **Wake**/**Restart**, **Configure**, and **Delete sandbox** — and clicking a row opens the chat. A primary **Create sandbox** action replaces the old Add Agent button, and an empty state points first-time users at it.

Everything the previous list could do carries over: wake and restart, configure, delete with confirmation, the setup-failure warning badge, and click-through to a hibernated sandbox's wake prompt.

Creating a sandbox (#894) and configuring one (#895) keep their existing dialogs for now — this page only opens those flows.

Closes #893

### DIFF
--- a/packages/ui/src/components/ui/dropdown-menu.tsx
+++ b/packages/ui/src/components/ui/dropdown-menu.tsx
@@ -1,0 +1,77 @@
+import * as DropdownMenuPrimitive from "@radix-ui/react-dropdown-menu";
+import { cva, type VariantProps } from "class-variance-authority";
+import * as React from "react";
+
+import { cn } from "@/lib/utils";
+
+const DropdownMenu = DropdownMenuPrimitive.Root;
+
+const DropdownMenuTrigger = DropdownMenuPrimitive.Trigger;
+
+const DropdownMenuContent = React.forwardRef<
+  React.ElementRef<typeof DropdownMenuPrimitive.Content>,
+  React.ComponentPropsWithoutRef<typeof DropdownMenuPrimitive.Content>
+>(({ className, align = "end", sideOffset = 4, ...props }, ref) => (
+  <DropdownMenuPrimitive.Portal>
+    <DropdownMenuPrimitive.Content
+      ref={ref}
+      align={align}
+      sideOffset={sideOffset}
+      collisionPadding={8}
+      className={cn(
+        "z-50 min-w-[160px] rounded-md border border-border bg-popover py-1 text-popover-foreground shadow-md animate-in fade-in-0 zoom-in-95 data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[side=bottom]:slide-in-from-top-2 data-[side=top]:slide-in-from-bottom-2",
+        className,
+      )}
+      {...props}
+    />
+  </DropdownMenuPrimitive.Portal>
+));
+DropdownMenuContent.displayName = DropdownMenuPrimitive.Content.displayName;
+
+const dropdownMenuItemVariants = cva(
+  "flex h-9 w-full cursor-pointer select-none items-center gap-2 rounded-md px-3 text-sm font-medium outline-none transition-colors data-[disabled]:pointer-events-none data-[disabled]:opacity-50",
+  {
+    variants: {
+      tone: {
+        default:
+          "data-[highlighted]:bg-muted data-[highlighted]:text-foreground",
+        danger:
+          "data-[highlighted]:bg-danger-light data-[highlighted]:text-danger",
+      },
+    },
+    defaultVariants: { tone: "default" },
+  },
+);
+
+const DropdownMenuItem = React.forwardRef<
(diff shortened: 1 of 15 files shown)


packages/ui/src/components/ui/dropdown-menu.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { cva, type VariantProps } from "class-variance-authority";
+import * as React from "react";
+
+import { cn } from "@/lib/utils";
+
+const DropdownMenu = DropdownMenuPrimitive.Root;
+
+const DropdownMenuTrigger = DropdownMenuPrimitive.Trigger;
+
+const DropdownMenuContent = React.forwardRef<
```
kapetr, line 11: `forwardRef` is no longer necessary in React 19 (finally! :)) and will be deprecated.

packages/ui/src/components/ui/dropdown-menu.tsx
```diff
@@ -0,0 +1,77 @@
...
+          "data-[highlighted]:bg-muted data-[highlighted]:text-foreground",
+        danger:
+          "data-[highlighted]:bg-danger-light data-[highlighted]:text-danger",
+      },
+    },
+    defaultVariants: { tone: "default" },
+  },
+);
+
+const DropdownMenuItem = React.forwardRef<
```
kapetr, line 46: Same here, remove `forwardRef`

packages/ui/src/components/ui/dropdown-menu.tsx
```diff
@@ -0,0 +1,77 @@
...
+>(({ className, tone, ...props }, ref) => (
+  <DropdownMenuPrimitive.Item
+    ref={ref}
+    className={cn(dropdownMenuItemVariants({ tone }), className)}
+    {...props}
+  />
+));
+DropdownMenuItem.displayName = DropdownMenuPrimitive.Item.displayName;
+
+const DropdownMenuSeparator = React.forwardRef<
```
kapetr, line 59: Same here, remove `forwardRef`

packages/ui/src/modules/agents/views/list-view.tsx
```diff
@@ -48,165 +51,89 @@ export function ListView() {
...
+          </Button>
         </div>
 
         {/* Skeleton during initial load — only when we expect agents */}
         {!initialLoaded && agents.length > 0 && (
-          <div className="flex flex-col gap-6">
-            <Card className="h-[88px] anim-pulse" />
-            <Card className="h-[88px] anim-pulse" />
-          </div>
+          <ListSkeleton rows={2} rowHeight={70} />
```
kapetr: Unreachable now - when `!initialLoaded` is true, `agents` is always `[]`.

kapetr (changes requested):

packages/ui/src/components/ui/dropdown-menu.tsx
PetrBulanek: We are actually on React 18, but we should definitely upgrade. :-)

Filed an issue https://github.com/dam-agents/dam/issues/927

kapetr (approved):
We should also remove "Get Started" banner, but that can be done in the last cleanup PR.
Image

PR #926, opened 2026-06-15
feat(ui): full-page sandbox creation wizard

### LINKED ISSUE #894: Redesign — full-page sandbox creation wizard

PetrBulanek:
Replaces the cramped creation modal and the experimental `/v2` wizard with **one full-page, three-step flow** for creating a sandbox, reached from the **Create sandbox** button on Home.

- **Image** — pick a pre-built image (Claude Code, Bob, Codex, Pi Agent…) or, under Advanced, bring your own container image.
- **Setup** — name the sandbox, connect a provider inline (Anthropic, OpenAI, Bob, IBM LiteLLM — with model overrides and edit/remove), and choose a network-access level.
- **Connections** — grant app connections and credentials (GitHub, GitHub Enterprise, MCP servers, custom header), connecting new ones inline. OAuth runs in a popup so the in-progress wizard is never lost. This step is skippable.

Finishing creates the sandbox and drops you into its chat session. A persistent step indicator runs down the side, and the wizard survives a reload (and the OAuth round-trip) without losing what you entered.

Closes #894


packages/ui/src/modules/agents/views/list-view.tsx
```diff
@@ -97,7 +98,7 @@ export function ListView() {
           <h1 className="text-[24px] font-semibold tracking-[-0.65px] text-foreground md:text-[28px]">
             Sandboxes
           </h1>
-          <Button onClick={() => setShowAddAgent(true)} disabled={busyAgent}>
+          <Button onClick={navigateToCreateSandbox} disabled={busyAgent}>
```
kapetr: We should remove `showAddAgent` state and `AddAgentDialog` - dead code now.

kapetr (changes requested):

kapetr (commented):
And I'd consider removing most of the comments :)

packages/ui/src/modules/agents/views/list-view.tsx
PetrBulanek: Dropped the dead `showAddAgent` state. Kept `AddAgentDialog` itself (with a note comment. It's the only UI for working-dir seeding (catalog-repo clone + local upload), which the wizard doesn't cover yet, so we'll port it into the wizard later rather than rebuild it.

kapetr (approved):

PR #1076, opened 2026-06-16
fix(ui): verify MCP OAuth URL before opening the auth tab

Tomas2D:
## Problem

In the sandbox creation wizard, adding a **Custom MCP server (OAuth)** connection and clicking **Create + Authorize** opened a new browser tab immediately — *before* the URL was checked at all. Validation existed only server-side, during the `create` call, which runs *after* the tab is already open. The result: a malformed or non-OAuth URL flashed a blank popup that was then closed.

Root cause: the popup is opened **synchronously** in the click handler (`openPopup()` runs before `await create.mutate(...)`), because browsers block `window.open` outside a user gesture. So the tab necessarily led the validation. Only the `popupOAuth` path (the wizard) was affected — the Settings → Connections path already navigates only after `create` succeeds.

## Fix

Before opening any tab, run a pre-flight check for OAuth templates that carry a user-typed `url` (custom MCP / dynamic registration):

1. Client-side URL format validation (`validateMcpUrl`).
2. The previously **unused** `discoverMcp` tRPC endpoint — confirms the server exposes OAuth discovery metadata (`/.well-known/oauth-*`) with DCR support.

If either fails, an inline error is shown and **no tab opens**. Only a verified URL proceeds to the OAuth flow.

- Premade providers (GitHub, etc.) have no `url` input, so they skip the check and keep their unchanged synchronous popup.
- For a verified custom-MCP URL the post-`await` popup may be blocked; the existing full-page-redirect fallback (already snapshot-safe in the wizard) handles that.
- Button now shows **Verifying…** during the pre-flight discovery.

## Tests

- New `validateMcpUrl` unit tests (accepts http/https, rejects malformed and non-http(s) schemes).
- `mise run ui:check` (lint + format + tsc) and `mise run ui:test` (47 tests) pass.

## Notes

UI-only change. No backend or contract changes — `discoverMcp` and the create-time discovery/DCR behavior are unchanged.

### DIFF
--- a/packages/ui/src/modules/connections/forms/template-create-form.tsx
+++ b/packages/ui/src/modules/connections/forms/template-create-form.tsx
@@ -20,8 +20,9 @@ import {
 } from "../../../components/modal.js";
 import { queryClient } from "../../../query-client.js";
 import { trpc } from "../../../trpc.js";
-import { useCreateConnection } from "../api/mutations.js";
+import { useCreateConnection, useDiscoverMcp } from "../api/mutations.js";
 import { useOAuthPopup } from "../hooks/use-oauth-popup.js";
+import { validateMcpUrl } from "../lib/mcp-url.js";
 
 export function TemplateCreateForm({
   template,
@@ -42,6 +43,7 @@ export function TemplateCreateForm({
   popupOAuth?: boolean;
 }) {
   const create = useCreateConnection();
+  const discover = useDiscoverMcp();
 
   const [name, setName] = useState(() => slugifyTemplateName(template.name));
   const [fields, setFields] = useState<Record<string, string>>(() => {
@@ -70,7 +72,7 @@ export function TemplateCreateForm({
   });
 
   const needsOAuth = template.authKind === "oauth";
-  const pending = create.isPending || authorizing;
+  const pending = create.isPending || authorizing || discover.isPending;
 
   const inputsByName = useMemo(() => {
     const map = new Map<string, ConnectionTemplateInput>();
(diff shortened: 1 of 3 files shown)


packages/ui/src/modules/connections/forms/template-create-form.tsx
```diff
@@ -172,6 +174,32 @@ export function TemplateCreateForm({
...
+        }
+        try {
+          const { auth } = await discover.mutateAsync({ url: mcpUrl });
+          if (auth !== "oauth") {
+            setError(
+              "Couldn't find OAuth discovery metadata at this URL. Check that it points to an MCP server that supports OAuth (we look for /.well-known/oauth-* endpoints).",
+            );
+            return;
+          }
+        } catch {
```
dam-code-guardian, line 197: 🟢 **Suggestion:** This `catch {}` silently swallows the discovery error and returns, relying entirely on the mutation hook's error toast to surface it. That's reasonable and the comment documents it, but a self-contained inline fallback would be more robust if the toast path ever changes:

```suggestion
        } catch {
          // A transport/server failure is surfaced by the mutation's error toast.
          setError("Couldn't verify the MCP server URL. Please try again.");
          return;
        }
```

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `5a9b85e`

## PR #1076: fix(ui): verify MCP OAuth URL before opening the auth tab
**Author:** Tomas2D | **Branch:** fix/mcp-oauth-url-validation → main | **Changes:** +76 −9 (3 files)

### Summary
Adds a client-side pre-flight check (URL format validation + the previously-unused `discoverMcp` tRPC endpoint) before the OAuth popup opens for custom-MCP connections in the creation wizard, so a malformed or non-OAuth URL fails inline instead of flashing a blank popup. Premade providers keep their unchanged synchronous popup. UI-only; no backend/contract changes.

### Findings
- ✅ **Looks good:** `validateMcpUrl` correctly rejects malformed URLs and non-http(s) schemes (incl. `javascript:`/`ftp:`) before any tab opens — a clear security improvement, with matching unit tests (`packages/ui/src/modules/connections/lib/mcp-url.ts`, `packages/ui/src/__tests__/unit/mcp-url.test.ts`).
- ✅ **Looks good:** Pre-flight is correctly gated on `submittedValue("url")`, so premade providers (no `url` input) skip it and retain their synchronous popup — matches the stated intent.
- ✅ **Looks good:** `pending` and the button label now account for `discover.isPending` ("Verifying…"), giving accurate feedback during the discovery round-trip.
- 🟢 **Suggestion:** The pre-flight `catch {}` silently swallows the error and returns, relying entirely on the mutation hook's error toast (`packages/ui/src/modules/connections/forms/template-create-form.tsx:197`). That coupling is reasonable and documented, but a brief inline `setError` fallback would make the failure self-contained if the toast path ever changes. Optional.

### Verdict
APPROVE — Small, well-tested, security-positive UI fix that closes the popup-before-validation gap without touching backend behavior.

---

jezekra1 (approved):

willallstet:
📝 Docs updated for this change: https://github.ibm.com/dam-agents/docs/pull/15

PR #1040, opened 2026-06-16
feat(ui): redesign component states, first-run, and responsive gaps

PetrBulanek:
Hardens the redesign (#892–#895) for the real-world states the happy-path designs skipped, so the product stays coherent the moment anything is empty, mid-flight, failing, narrow, or used by a brand-new user.

## What this does

- **First-run** — a brand-new user (no sandboxes, no provider) is taken straight into creating their first sandbox instead of a separate setup-progress prompt. Backing out lands them on the Sandboxes list and keeps them there for the session.
- **Connection states** — connections render the same visual states everywhere they appear (the creation wizard, the sandbox settings page, and Settings → Connections), including a new **"Authorizing…"** state while an OAuth connection is mid-flight.
- **Responsive** — the new shell and wizard work on narrow screens; the wizard's action buttons no longer sit behind the mobile bottom nav.
- **Design fidelity** — a pass over the redesign surfaces brings type scale, spacing, settings titles, and copy in line with the designs.
- A small touch: the name field is focused automatically on the wizard's setup step.

Closes #896

PR #1078, opened 2026-06-16
feat(ui): sandbox wizard + configure-page polish

kapetr:
Follow-up UI polish on the sandbox creation wizard and the configure-sandbox page (after #1040).

**Wizard**
- Visited steps stay navigable — going back to an earlier step no longer disables the ones you've already reached; steps beyond the furthest reached stay locked.
- The image step gets its own Continue button (picking an image selects rather than auto-advancing), and the custom-image input submits on Enter.

**Configure page + shared form polish**
- Section content (name, image, provider, connections, network, environment) aligns with its label via a shared offset; Network and Environment sit in bordered boxes.
- One shared `SectionLabel` for all uppercase section labels, and one `FormField` for the content offset.
- Standard input padding bumped to 16px so content lines up; small/compact inputs keep 12px and gain a dedicated size. Cards use a border instead of a shadow. Back link adopts the larger type.

Scope note: a couple of the underlying changes (input padding, Card border) are design-system level and affect inputs/cards app-wide.

### DIFF
--- a/packages/ui/src/modules/agents/components/configure-agent/env-tab.tsx
+++ b/packages/ui/src/modules/agents/components/configure-agent/env-tab.tsx
@@ -1,6 +1,7 @@
 import { AlertTriangle, KeyRound, Lock } from "lucide-react";
 
 import { SectionLabel } from "@/components/ui/section-label";
+import { cn } from "@/lib/utils";
 
 import { EnvVarsEditor } from "../../../../components/env-vars-editor.js";
 import type { EnvVar } from "../../../../types.js";
@@ -107,7 +108,7 @@ function InheritedEnvRow({ entry }: { entry: InheritedEnv }) {
   return (
     <div className="group flex items-center gap-2 rounded-md border px-3 py-1.5 text-[12px]">
       <span
-        className={`shrink-0 ${isSystem && "text-text-muted"}`}
+        className={cn("shrink-0", isSystem && "text-text-muted")}
         title={isSystem ? "Platform-managed" : `From connection: ${sourceName}`}
       >
         {isSystem ? <Lock size={12} /> : <KeyRound size={12} />}
--- a/packages/ui/src/modules/agents/components/configure-agent/env-tab.tsx
+++ b/packages/ui/src/modules/agents/components/configure-agent/env-tab.tsx
@@ -1,5 +1,7 @@
 import { AlertTriangle, KeyRound, Lock } from "lucide-react";
 
+import { SectionLabel } from "@/components/ui/section-label";
+
 import { EnvVarsEditor } from "../../../../components/env-vars-editor.js";
 import type { EnvVar } from "../../../../types.js";
 
@@ -45,8 +47,8 @@ export function EnvTab({
 }) {
   const warnings = shadowWarnings(envVars, inherited);
   return (
-    <>
-      <p className="text-[12px] text-text-muted">
+    <div className="flex flex-col gap-6">
+      <p className="text-[14px] text-muted-foreground">
         Applied to every instance of this agent. Restart the instance pod to
         pick up changes.
       </p>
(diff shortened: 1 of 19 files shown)


packages/ui/src/modules/agents/components/configure-agent/env-tab.tsx
```diff
@@ -110,9 +105,9 @@ function InheritedEnvRow({ entry }: { entry: InheritedEnv }) {
         ? entry.source.secretName
         : entry.source.appLabel;
   return (
-    <div className="group flex items-center gap-2 rounded-md border-2 border-border-light bg-surface-raised px-3 py-1.5 text-[12px]">
+    <div className="group flex items-center gap-2 rounded-md border px-3 py-1.5 text-[12px]">
       <span
-        className={`shrink-0 ${isSystem ? "text-text-muted" : "text-accent"}`}
+        className={`shrink-0 ${isSystem && "text-text-muted"}`}
```
PetrBulanek: This is on purpose? Also the "isSystem" can translate to "false" string if falsey. I'd use `cn`.

PetrBulanek (changes requested):

PetrBulanek (approved):

packages/ui/src/modules/agents/components/configure-agent/env-tab.tsx
dam-code-guardian: 🟡 **Warning:** When `isSystem` is falsey, `isSystem && "text-text-muted"` evaluates to `false`, which a template literal interpolates as the literal string `"false"` — producing `class="shrink-0 false"`. It also drops the `text-accent` color that non-system (connection-derived) entries had before. Restore the ternary:

```suggestion
        className={`shrink-0 ${isSystem ? "text-text-muted" : "text-accent"}`}
```

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `b9ba2a7`

## PR #1078: feat(ui): sandbox wizard + configure-page polish
**Author:** kapetr | **Branch:** feat/wizard-navigable-steps → main | **Changes:** +158 −100 (19 files)

### Summary
UI polish on the sandbox creation wizard and configure-sandbox page (follow-up to #1040): visited wizard steps stay navigable via a new `maxStep` high-water mark, the image step gets an explicit Continue button and Enter-to-submit on the custom-image input, and shared `SectionLabel`/`FormField` primitives plus design-system tweaks (input padding 16px, Card border instead of shadow) align section content with labels. One human review thread on `env-tab.tsx` is still open and the issue it raises is real — surfaced below.

### Findings
- 🟡 **Warning:** `${isSystem && "text-text-muted"}` renders the literal class `"false"` when `isSystem` is falsey, and drops the `text-accent` color that connection-derived entries had before. The prior code was a ternary (`isSystem ? "text-text-muted" : "text-accent"`). (`packages/ui/src/modules/agents/components/configure-agent/env-tab.tsx:110`) — matches @PetrBulanek's open inline comment.
- 🟢 **Suggestion:** `maxStep` is recomputed both in `update()` (`Math.max(prev.maxStep || prev.step, patch.step)`) and in `loadSnapshot()` (`Math.max(parsed.data.maxStep, parsed.data.step)`). The `prev.maxStep || prev.step` fallback is only needed for snapshots predating this field; since `wizardSnapshotSchema` now `.default(1)`s `maxStep` and `loadSnapshot` already clamps it, `prev.maxStep` is always defined downstream — the `|| prev.step` is dead defensiveness. Minor. (`packages/ui/src/modules/sandboxes/hooks/use-sandbox-wizard.ts:24`)
- ✅ **Looks good:** `maxStep` high-water-mark logic is sound — it only ratchets up (`Math.max`), going back to an earlier step keeps later visited steps unlocked, and `wizard-step-indicator.tsx` correctly marks `item.n <= maxStep` as visited.
- ✅ **Looks good:** Backward-compatible snapshot migration — `maxStep` defaults to `1` in the schema and is clamped to at least `step` on load, so persisted snapshots without the field parse cleanly.
- ✅ **Looks good:** Consolidating uppercase labels into `SectionLabel` and the content offset into `FormField` removes several copies of duplicated Tailwind strings.

### Verdict
COMMENT — Solid, well-factored UI polish; one real (minor) styling regression on `env-tab.tsx:110` that a human reviewer also flagged and that remains unaddressed.

---

Issue #1079, opened 2026-06-16
Deliver custom environment variables through the contribution rail, not the agent CR


jezekra1:
## Problem

Custom environment variables set in the UI (the **Environment** editor) are delivered differently from every other piece of agent configuration. They're written to the agent's spec and projected into the container by the controller at pod start, so:

- **Editing an env var only takes effect on the next pod restart.** Connections, secrets, skills, and MCP servers all apply to the next turn/session without a restart, because they ride the runtime "contribution" rail. Env is the odd one out.
- **There are two mechanisms for "env."** Secret- and connection-derived env vars were already moved onto the contribution rail (no pod roll) — see #122 and #367. User-typed env vars still go through the old controller-projection path on the agent's spec. Two paths for the same concept is confusing and means the env editor behaves inconsistently with the rest of the config surface.

## Proposed solution

Deliver user-set env vars through the **same contribution rail** that already carries connection/secret-derived env: the runtime channel writes them to a file the harness merges at session start. Editing env then applies on the next turn — no pod restart — and there is **one** env mechanism for everything.

As part of this, **stop carrying env on the agent CR** (`spec.env`) and stop composing container env in the controller, so configuration lands consistently via contributions and the CR-render path is no longer in the env business. The user-facing Environment editor is unchanged — only delivery moves.

### Scope

- Extends the secret/connection env unification (#122, #367) to user-typed env — the last env source still on the CR/controller path.
- UI editor, validation, and inherited/shadowing rules stay as-is.

### Out of scope / open questions

- The platform-enforced `PORT` (and any controller-injected platform env): decide whether it moves to the contribution rail too or stays a controller concern, so the controller can fully exit env composition.
- Where user env values are stored once they're off `spec.env` (e.g. alongside other per-agent application state).
- Migration for existing agents that already carry `spec.env`.

PR #1080, opened 2026-06-16
feat(agent-config): persistent per-agent model/mode/config defaults

### LINKED ISSUE #931: Set an agent's default model and mode from the Config panel
## Problem

An agent's model and operating mode are configured in two disconnected places today, and neither gives the user a durable, visible default:

- **Static config baked into the image** (e.g. Claude reads `.claude/settings.json`). Changing it means hand-editing files inside the agent — not something a normal user can or should do.
- **The per-session picker in the UI**, which applies a choice only to the current conversation and remembers it only in the current browser tab. It never becomes the agent's standing default.

So there's no way to say *"this agent should always run on model X in mode Y."* A choice doesn't follow the user to another device, and — most importantly — **scheduled and triggered runs (cron, Slack/Telegram) start with the image's baked-in defaults**, silently ignoring whatever the user picked interactively. This is the same gap we already closed for MCP servers in #213, but for model and mode.

## Proposed solution

Add a **Model settings** section to the agent Config panel (alongside Schedules, Channels, MCP servers, Skills) where the user picks the agent's **default model** and **default mode**. The choice is saved as a persistent, server-owned default that:

- writes through to the agent's own config (e.g. `.claude/settings.json`) so the running harness picks it up,
- survives pod restarts and is visible to every client — not stranded in one browser,
- applies to **all** future turns: interactive, scheduled, and trigger-driven alike.

Three layers, kept distinct so they don't fight:

- **Available options come from the agent** — the dropdowns are populated from the models and modes the harness actually advertises, not a hardcoded list (builds on #702 / #163).
- **Persistent default** — *the new piece.* What the Config panel sets; the agent's standing default.
- **Per-session live override** — the existing in-conversation picker stays, as a "just this conversation" override layered on top of the default.

### Scope

- **Harness-agnostic from the start.** Each harness has its own config file and its own names for models/modes; the platform stores a *normalized* choice (model + mode) and each harness maps it to its native config. Reuses the same persistence path that already delivers MCP servers and skills into the agent (#213).
- **"Mode" means the harness's permission/operating mode** (e.g. auto / plan / accept-edits), **not** the platform's chat-vs-terminal session mode. These are different axes and must not be conflated in the UI.

### Out of scope

- **Per-schedule overrides** (a single schedule wanting a different model than the agent's default) — defer; the agent-level default comes first.
- **Model recommendations / credential guidance per harness** (#163) — complementary, tracked separately.

### Open questions

- Does **Model settings** live in the Config panel as drawn, or on the in-flight full-page settings view (#895)? Should align with that redesign.
- **Terminology** — this issue says "agent"; the redesign says "sandbox." Pick one before filing into that area."

Image

jezekra1:
Closes #931.

## What

A per-agent **harness session default** (model, mode, and ACP config options like `thought_level`) that's written into the harness's *own* config file (claude-code → `~/.claude/settings.json`), so it applies to **interactive and scheduled/headless** sessions alike. It layers on the existing per-session ACP picker — which now configures the *live session only* (its global localStorage preference store is removed).

Three layers, one coherent system:
- **Discovery** — available options come from what the harness advertises over ACP (`models` / `modes` / `configOptions`).
- **Persistent default** *(new)* — set in the Config panel, stored server-side, written through to the harness config file.
- **Per-session override** — the existing composer popover, applied live over ACP.

## Shape mirrors ACP session config

The contribution, the DB row, and the manifest `keys` all share the ACP shape `{ model, mode, configOptions: { <configId>: value } }`. `thought_level` is **not** special-cased — it's a config option; the claude-code manifest maps it under `configOptions` to Claude's `effortLevel`. Harness-specifics (file path + key layout) live in each harness's `runtime-manifest.yaml`, so api-server stays harness-agnostic.

## Harness-agnostic + capability-gated

Only **claude-code** binds `harness-config` in v1 (codex is TOML, pi-agent reads elsewhere). Capabilities derive from the manifest's `drivers` map, so only harnesses that bind it advertise it; the UI hides the section (via a `supported` flag) when the harness can't honor a persistent default, and the driver skips + logs any unmapped option — no silent no-ops.

## Layers / changes

- **agent-runtime-api** — new `harness-config` contribution kind.
- **agent-runtime** — manifest-bound `harness-config` driver: maps each value to a dot-path in the harness's JSON config file, reconciles non-destructively (preserves user keys, removes cleared ones via a state store, throws rather than clobbering an unparseable file). Bound in a new claude-code manifest. Unit-tested.
- **packages/db** — `agent_settings` table + migration.
- **api-server** — `agent-settings` module (repo / service / router / `AgentDeleted` cleanup saga), ownership-checked `get`/`set`; state-builder emits the contribution.
- **ui** — Model section in the Config panel (model / mode / config-option groups); removed the localStorage global-preference mechanism (`save/getSavedPreferences`, `applySavedPreferences`).
- **docs** — `connections.md` updated.

## Verification

- `mise run check` / `mise run test`: all TS packages tsc + lint clean; unit tests pass (agent-runtime, api-server +6 new for the agent-settings service & capability gating, ui). Two pre-existing **environmental** failures unrelated to this diff (helm kubeconform schema fetch timing out; controller CRD-revision tooling) — hence the branch was committed with `--no-verify` after verifying each TS package individually.
- **Cluster round-trip verified** on `platform-k3s`: a claude-code agent advertises `harness-config`; setting a row delivered the contribution (no driver failures) and the driver wrote `model` + `permissions.defaultMode` + `effortLevel` into `~/.claude/settings.json`, preserving the seeded keys.

## Follow-ups / open items

- Confirm the `thought_level` ACP **config id** matches the manifest key (`configOptions.thought_level`) for the current claude-agent-acp — a one-line manifest fix if it differs. The driver mechanism itself is verified.
- Boolean config options are persistable in the data layer but not yet surfaced in the panel (deferred; still a live per-session toggle).
- Per-schedule overrides remain out of scope (per #931).
- Moving user-set **env** off the agent CR onto the contribution rail is tracked separately in #1079 (out of scope here).

### DIFF
--- a/packages/ui/src/modules/sessions/components/model-settings-panel.tsx
+++ b/packages/ui/src/modules/sessions/components/model-settings-panel.tsx
@@ -1,0 +1,236 @@
+import type { SessionConfigOption } from "@agentclientprotocol/sdk/dist/acp.js";
+import { Check } from "lucide-react";
+
+import { useStore } from "../../../store.js";
+import {
+  useAgentSettings,
+  useSetAgentSettings,
+} from "../../agents/api/agent-settings.js";
+import { Section } from "./config-section.js";
+import { flattenSelectOptions } from "./session-config-popover.js";
+
+type SelectConfigOption = Extract<SessionConfigOption, { type: "select" }>;
+
+interface Choice {
+  id: string;
+  name: string;
+  description?: string | null;
+}
+
+/**
+ * The agent's persistent model/mode/config default, set from the Config panel
+ * and written through to the harness's own config file. The per-session picker
+ * in the composer (SessionConfigBar) overrides this default for one
+ * conversation; this section sets what every new/scheduled session starts from.
+ *
+ * Shape mirrors ACP session config: `model` and `mode` are dedicated axes, and
+ * everything else (e.g. the `thought_level` option) is a generic config option.
+ * Options come from the ACP session-config slice; saved values come from the
+ * server. Hidden when the harness can't honor a persistent default.
+ */
+export function ModelSettingsPanel({ agentId }: { agentId: string | null }) {
+  const models = useStore((s) => s.sessionModels);
+  const modes = useStore((s) => s.sessionModes);
+  const configOptions = useStore((s) => s.sessionConfigOptions);
+  const { data } = useAgentSettings(agentId);
+  const setSettings = useSetAgentSettings();
+
+  if (!agentId || !data || !data.supported) return null;
+
+  // Model arrives either as the legacy session-model axis or as a
+  // `category: "model"` select option — normalize both (same as the popover).
+  const modelOption = configOptions.find(
+    (o): o is SelectConfigOption =>
+      o.category === "model" && o.type === "select",
+  );
+  const modelChoices: Choice[] = models
+    ? models.availableModels.map((m) => ({
+        id: m.modelId,
+        name: m.name,
+        description: m.description,
+      }))
+    : modelOption
+      ? flattenSelectOptions(modelOption.options).map((o) => ({
+          id: o.value,
+          name: o.name,
+          description: o.description,
+        }))
+      : [];
+  const modeChoices: Choice[] =
+    modes?.availableModes.map((m) => ({
+      id: m.id,
+      name: m.name,
+      description: m.description,
+    })) ?? [];
+  // Every other select option (e.g. `thought_level`) — model/mode have their
+  // own groups above, so exclude those categories to avoid rendering twice.
+  // Boolean options are persistable (the contribution/driver accept booleans)
+  // but not yet surfaced here as a saved default; they remain a live per-session
+  // toggle in the composer popover. Deferred until a harness ships one.
+  const extraOptions = configOptions.filter(
+    (o): o is SelectConfigOption =>
+      o.type === "select" && o.category !== "model" && o.category !== "mode",
+  );
+
+  if (
+    modelChoices.length === 0 &&
+    modeChoices.length === 0 &&
+    extraOptions.length === 0
+  ) {
+    return (
+      <Section title="Model">
+        <div className="px-4 py-3 text-[12px] text-text-muted">
+          Start a session to load the options this agent offers.
+        </div>
+      </Section>
+    );
+  }
+
+  const persist = (patch: {
+    model?: string | null;
+    mode?: string | null;
+    configOptions?: Record<string, string | boolean>;
+  }) =>
+    setSettings.mutate({
+      agentId,
+      model: data.model,
+      mode: data.mode,
+      configOptions: data.configOptions,
+      ...patch,
+    });
+
+  return (
+    <Section title="Model">
+      {modelChoices.length > 0 && (
+        <OptionGroup
+          title="Model"
+          choices={modelChoices}
+          value={data.model}
+          onSelect={(id) => persist({ model: id })}
+        />
+      )}
+      {modeChoices.length > 0 && (
+        <OptionGroup
+          title="Mode"
+          choices={modeChoices}
+          value={data.mode}
+          onSelect={(id) => persist({ mode: id })}
+        />
+      )}
+      {extraOptions.map((option) => (
+        <ConfigOptionGroup
+          key={option.id}
+          option={option}
+          configOptions={data.configOptions}
+          onChange={(next) => persist({ configOptions: next })}
+        />
+      ))}
+    </Section>
+  );
+}
+
+function ConfigOptionGroup({
+  option,
+  configOptions,
+  onChange,
+}: {
+  option: SelectConfigOption;
+  configOptions: Record<string, string | boolean>;
+  onChange: (next: Record<string, string | boolean>) => void;
+}) {
+  const choices: Choice[] = flattenSelectOptions(option.options).map((o) => ({
+    id: o.value,
+    name: o.name,
+    description: o.description,
+  }));
+  const saved = configOptions[option.id];
+  const current = typeof saved === "string" ? saved : null;
+
+  const select = (id: string | null) => {
+    const next = { ...configOptions };
+    if (id === null) delete next[option.id];
+    else next[option.id] = id;
+    onChange(next);
+  };
+
+  return (
+    <OptionGroup
+      title={option.name}
+      choices={choices}
+      value={current}
+      onSelect={select}
+    />
+  );
+}
+
+function OptionGroup({
+  title,
+  choices,
+  value,
+  onSelect,
+}: {
+  title: string;
+  choices: Choice[];
+  value: string | null;
+  onSelect: (id: string | null) => void;
+}) {
+  return (
+    <div className="border-b border-border-light last:border-b-0">
+      <div className="px-4 pt-3 pb-1 text-[10px] font-bold uppercase tracking-[0.05em] text-text-muted">
+        {title}
+      </div>
+      <OptionRow
+        label="Not set"
+        description="No saved default — the harness picks on its own"
+        active={value === null}
+        onClick={() => onSelect(null)}
+      />
+      {choices.map((c) => (
+        <OptionRow
+          key={c.id}
+          label={c.name}
+          detail={c.id}
+          description={c.description}
+          active={c.id === value}
+          onClick={() => onSelect(c.id)}
+        />
+      ))}
+    </div>
+  );
+}
+
+function OptionRow({
+  label,
+  detail,
+  description,
+  active,
+  onClick,
+}: {
+  label: string;
+  detail?: string;
+  description?: string | null;
+  active: boolean;
+  onClick: () => void;
+}) {
+  return (
+    <button
+      className={`flex items-center gap-2 w-full px-4 py-2 text-[13px] text-left transition-colors ${active ? "text-accent bg-accent-light font-semibold" : "text-text hover:bg-surface-raised"}`}
+      onClick={onClick}
+    >
+      {active && <Check size={12} className="shrink-0" />}
+      <div className={active ? "" : "ml-[20px]"}>
+        <div>{label}</div>
+        {detail && (
+          <div className="text-[11px] text-text-muted font-normal font-mono">
+            {detail}
+          </div>
+        )}
+        {description && (
+          <div className="text-[11px] text-text-muted font-normal">
+            {description}
+          </div>
+        )}
+      </div>
+    </button>
+  );
+}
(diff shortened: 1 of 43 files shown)


packages/ui/src/modules/sessions/components/model-settings-panel.tsx
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  const persist = (patch: {
+    model?: string | null;
+    mode?: string | null;
+    configOptions?: Record<string, string | boolean>;
+  }) =>
+    setSettings.mutate({
+      agentId,
+      model: data.model,
```
dam-code-guardian, line 96: 🟡 **Warning:** `persist()` reads the sibling axes from the cached `data` query and overlays a single-axis patch, but `useSetAgentSettings` has no optimistic `onMutate` — it only invalidates `agentSettings.get` on completion. If the user picks a second axis (e.g. mode) before the first save's refetch lands, this re-sends the **stale** model value, silently reverting the first selection. Base the patch on the in-flight mutation variables, or add an optimistic cache update.

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `de733f6`

## PR #1080: feat(agent-config): persistent per-agent model/mode/config defaults
**Author:** jezekra1 | **Branch:** feat/agent-model-mode-config → main | **Changes:** +2783 −256 (43 files)

### Summary
Adds a per-agent **harness session default** (model, mode, and ACP config options like `thought_level`) stored server-side in a new `agent_settings` Postgres table and delivered as a new `harness-config` Contribution kind that the agent-runtime reconciles into the harness's own config file (`~/.claude/settings.json`), so it applies to interactive and scheduled/headless sessions alike. The old per-session localStorage preference store and its replay path are removed; the composer popover now configures the live session only. Closes #931.

### Findings
- ✅ **Looks good:** clean DDD layering — schema + migration, repository (defensive jsonb `safeParse` on read), service (ownership-gated get/set), tRPC router, compose wiring, and an `AgentDeleted` cleanup saga (subscribed in `index.ts` and unsubscribed on shutdown).
- ✅ **Looks good:** `harness-config-plugin.ts` driver is robust — tracks managed key-paths in a state store so a cleared field removes only its own key (`deleteNested` prunes emptied ancestors), preserves user-authored keys, skips unmapped options, and **throws rather than clobbering** an unparseable settings file.
- ✅ **Looks good:** state-builder omits the contribution entirely when nothing is set, so the driver removes previously-written keys — round-trips cleanly with the managed-key state store.
- ✅ **Looks good:** `harness-config` binding lives in claude-code's own `runtime-manifest.yaml` (path/keys are Claude-specific) rather than platform-base — correct, since codex (TOML) / pi-agent read elsewhere; `supported` flag gates the UI section off when the harness doesn't advertise the capability.
- ✅ **Looks good:** strong unit coverage — driver test (mapping, user-key preservation, key removal, unmapped skip, no-file, unparseable-protection) and service test (empty/stored/ownership/bump-enqueue, `harnessConfigSupported`).
- ✅ **Looks good:** localStorage preference store and `applySavedPreferences` replay fully torn out; `use-acp-config-cache` now caches only the option *catalog*, matching the new server-side-default design.
- 🟡 **Warning:** `ModelSettingsPanel.persist()` (`packages/ui/src/modules/sessions/components/model-settings-panel.tsx:89`) builds each save from the cached `data` query and overlays a single-axis patch. The `set` mutation has no `onMutate` optimistic update — it only invalidates `agentSettings.get` on completion. If a user changes a second axis (e.g. mode) before the first save's refetch lands, the second request re-sends the **stale** sibling value (the just-changed model), silently reverting the first selection. Consider an optimistic cache update in `useSetAgentSettings`, or deriving the patch base from the in-flight mutation variables rather than the cached query.

**Verdict: minor drift**

Check 1–6 pass for the primary subsystem page **connections.md** (owner of Contributions + runtime channel): the `harness-config` kind is added to the Contribution union, the fanout-rail table, and the built-in-driver table, with `Last verified:` bumped to `2026-06-16`; no ADR leaks.

⚠️ **Possible drift (check 1, judgment call):** `persistence.md` may warrant a one-line mention of the new `agent_settings` Postgres table. The diff adds a new agent-scoped table (`db/schema.ts`, migration `0002`) plus an `AgentDeleted` cascade saga. `persistence.md` enumerates Postgres-held state categorically (line 59 lists the analogous "skills catalog — per-Agent install records") and explicitly discusses cascade-vs-orphan delete semantics (line 122) — and `agent_settings` is a table that *does* explicitly cascade. Its generic "plus any other api-server-only domain resource" framing arguably already covers it, so this is a judgment call, not a hard miss; a bullet analogous to the skills-catalog entry would keep it current.

### Verdict
COMMENT — Excellent, well-tested feature with clean layering and a robust reconciliation driver. One non-blocking UI race (stale-sibling on rapid multi-axis saves) and an optional persistence.md doc bullet; neither blocks merge.

---

packages/ui/src/modules/sessions/components/model-settings-panel.tsx
dam-code-guardian, line 96: 🟡 **Warning:** `persist()` builds each save from the cached `data` query (`model`/`mode`/`configOptions`) and overlays a single-axis patch, but `useSetAgentSettings` still has no optimistic `onMutate` — it only invalidates `agentSettings.get` on completion. If the user picks a second axis (e.g. mode) before the first save's refetch lands, this re-sends the **stale** sibling value, silently reverting the first selection. Base the patch on the in-flight mutation variables, or add an optimistic cache update in `useSetAgentSettings`.

_(Carried over from the previous review at `de733f6`; unchanged on this HEAD.)_

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `e50597d`

## PR #1080: feat(agent-config): persistent per-agent model/mode/config defaults
**Author:** jezekra1 | **Branch:** feat/agent-model-mode-config → main | **Changes:** +2888 −255 (43 files)

### Summary
Adds a per-agent **harness session default** (model, mode, and ACP config options like `thought_level`) stored server-side in a new `agent_settings` Postgres table and delivered as a new `harness-config` Contribution kind that the agent-runtime reconciles into the harness's own config file (`~/.claude/settings.json`), so it applies to interactive and scheduled/headless sessions alike. The old per-session localStorage preference store and its replay path are removed; the composer popover now configures the live session only. Closes #931.

### Changes since last review
Previous HEAD: `de733f6` (2026-06-16T15:52:54Z) — verdict COMMENT

- 🔁 **Still present:** stale-sibling race in `ModelSettingsPanel.persist()` (`packages/ui/src/modules/sessions/components/model-settings-panel.tsx:96`) — `useSetAgentSettings` still has no optimistic `onMutate`; carried over unchanged.
- ℹ️ **Note:** the new HEAD is a rebase of the feature commit onto a newer `main` (absorbs v0.2.3, the sandbox-wizard/configure-page polish, the headless API-keys feature, etc.). The PR's own diff vs its base is unchanged in substance; the DB migration was renumbered `0002_marvelous_dragon_man` → `0003_lying_amazoness` to sit after a migration that landed on main in the meantime.

### Findings
- ✅ **Looks good:** clean DDD layering — schema + migration, repository (defensive jsonb `safeParse` on read), service (ownership-gated get/set), tRPC router, compose wiring, and an `AgentDeleted` cleanup saga (subscribed in `index.ts`, unsubscribed on shutdown).
- ✅ **Looks good:** `harness-config-plugin.ts` driver is robust — tracks managed key-paths in a state store so a cleared field removes only its own key (`deleteNested` prunes emptied ancestors), preserves user-authored keys, skips unmapped options, and **throws rather than clobbering** an unparseable settings file.
- ✅ **Looks good:** state-builder omits the contribution entirely when nothing is set, so the driver removes previously-written keys — round-trips cleanly with the managed-key state store.
- ✅ **Looks good:** `harness-config` binding lives in claude-code's own `runtime-manifest.yaml` (path/keys are Claude-specific) rather than platform-base — correct, since codex (TOML) / pi-agent read elsewhere; `supported` flag gates the UI section off when the harness doesn't advertise the capability.
- ✅ **Looks good:** strong unit coverage — driver test (mapping, user-key preservation, key removal, unmapped skip, no-file, unparseable-protection) and service test (empty/stored/ownership/bump-enqueue, `harnessConfigSupported`).
- ✅ **Looks good:** localStorage preference store and `applySavedPreferences` replay fully torn out; `use-acp-config-cache` now caches only the option *catalog*, matching the new server-side-default design.
- 🟡 **Warning:** `ModelSettingsPanel.persist()` (`packages/ui/src/modules/sessions/components/model-settings-panel.tsx:96`) builds each save from the cached `data` query and overlays a single-axis patch. `useSetAgentSettings` has no `onMutate` optimistic update — it only invalidates `agentSettings.get` on completion. If a user changes a second axis (e.g. mode) before the first save's refetch lands, the second request re-sends the **stale** sibling value (the just-changed model), silently reverting the first selection. Consider an optimistic cache update in `useSetAgentSettings`, or deriving the patch base from the in-flight mutation variables rather than the cached query.

**Verdict: minor drift**

Checks 1–6 pass for the primary subsystem page **connections.md** (owner of Contributions + runtime channel): the `harness-config` kind is added to the Contribution union, the fanout-rail table, and the built-in-driver table, with `Last verified:` bumped 2026-06-12 → 2026-06-16; no ADR leaks; no volatile-content additions.

⚠️ **Possible drift (check 1, judgment call):** `persistence.md` may warrant a one-line mention of the new `agent_settings` Postgres table. The diff adds a new agent-scoped table (`packages/db/src/schema.ts` → `agentSettings`, migration `0003_lying_amazoness.sql`) **plus an `AgentDeleted` cascade saga** (`agent-settings-cleanup.ts`). `persistence.md` enumerates Postgres-held state categorically (the "skills catalog — per-Agent install records" bullet) and explicitly discusses cascade-vs-orphan delete semantics (the "survive Agent deletion as orphans unless the deletion path explicitly cascades" line) — and `agent_settings` is a per-agent table that *does* explicitly cascade. Its generic "plus any domain resource the controller does not reconcile" framing arguably already covers it, so this is a judgment call, not a hard miss; a bullet analogous to the skills-catalog entry would keep it current.

### Verdict
COMMENT — Excellent, well-tested feature with clean layering and a robust reconciliation driver, unchanged in substance since the prior review (just rebased onto newer main). One non-blocking UI race (stale-sibling on rapid multi-axis saves) and an optional persistence.md doc bullet; neither blocks merge.

---

jezekra1:
Alternative implementation in #1270 — same feature (#931), but delivered as a one-shot `harness-config` **event** instead of a reconciled contribution. Key difference: #1270 never re-asserts, so a user editing `~/.claude/settings.json` directly is preserved rather than clobbered, and it needs no DB table/migration/state-builder (~0.9k lines vs ~2.8k here). Keeping this PR open for review/comparison; we'd merge at most one.

Issue #896, opened 2026-06-12
Redesign — component states, first-run, and responsive gaps


PetrBulanek:
## Problem

The designs cover the happy path but leave several real-world states undrawn or only sketched. Shipping the redesign without these would mean an inconsistent experience the moment anything is empty, loading, failing, on a small screen, or being used by a brand-new user. This issue collects those gaps so they're handled deliberately rather than improvised per-screen.

## Proposed solution

Fill the gaps, treating each as first-class:

- **Connection states.** A connection (e.g. GitHub) needs consistent visual states across the product: not-connected, hover, connected, and connected-and-selected. Define these once so they look the same in the wizard, the edit page, and Settings.
- **First-run.** A brand-new user with no sandboxes should be guided straight into creating their first one, replacing today's separate "setup progress" prompts. The relevant frame in the designs is only a placeholder, so this needs a concrete definition.
- **Empty / loading / error states.** Define what users see while things load, when a list is empty, when creating a sandbox fails, when a provider credential fails its check, and when a connection is mid-authorization.
- **Responsive / mobile.** The designs are desktop-only; the product is used on smaller screens today. Decide and define how the new navigation and the wizard behave when the viewport is narrow.

### Scope

- A reusable connection-state component, the first-run path, the set of empty/loading/error states, and responsive behaviour for the new shell and wizard.

### Dependencies

- Blocked on #892, #893, #894, #895 (this hardens states for the screens those issues introduce).

### Out of scope

- The happy-path screens themselves (covered by #893, #894, #895).

### Open questions

- Mobile layouts and most error states are **absent from the designs** — these need either design input or an agreed convention before building.

### Design reference

Figma page **"#356 - Agent creation user flow"**: connection states `152:5805`–`152:5808` (labels `152:5809`/`152:5810`/`152:5815`/`152:5819`); first-time placeholder `152:5941`; OAuth redirect placeholders `152:5341` (GitHub) / `152:4566` (Google Drive).

PetrBulanek:
**Implementation orientation** — code pointers / reuse map for whoever picks this up. Not part of the ticket spec.

- **Connection states:** consolidate around `src/components/app-status-pill.tsx` and the connection rows in `src/modules/connections/components/`.
- **First-run:** today driven by `src/components/setup-progress-bar.tsx` + `src/lib/onboarding.ts` (a 3-step pill that shows on empty list) — this is likely removed/replaced by routing first-run users into the wizard.
[...]

PR #1086, opened 2026-06-16
test(e2e): click Continue on the wizard image step in agent spec

tomkis:
## What

Fixes the failing `mise e2e` job on `main`. The agent spec (`03-agent.spec.ts`) was timing out and failing CI on every push.

## Root cause

```
✘ 03-agent.spec.ts:10 › create a mock agent with the connection attached (1.0m)
   Error: locator.fill: Test timeout of 60000ms exceeded.
   - waiting for getByPlaceholder('my-sandbox')
```

The sandbox wizard redesign (#1040) made the **image step select-only**: clicking a pre-built image now only *selects* it (`onSelect`) and you advance with a new **Continue** button — it no longer auto-advances to the setup step (`image-step.tsx`, wired to `step: 2` in `sandbox-wizard-view.tsx`).

The agent spec still assumed the old auto-advance: it picked the `mock` image and immediately waited for the `my-sandbox` name field, which lives on **step 2**. So it sat on step 1 and timed out. #1040 patched several other things in this spec (first-run suppression, template test-ids) but missed this click.

## Fix

Click the image step's **Continue** button after selecting the mock image, matching the Continue pattern already used in the spec's later steps. Verified against latest `main` (incl. the follow-up wizard PR #1078) — the pre-built `ImageCard` still only selects on click, so the explicit Continue is required.

## Testing

- `eslint` and `prettier --check` pass on the changed file.
- The full e2e suite requires a live k3s cluster (`mise run e2e`), which isn't available in my environment — relying on CI to run the suite on this PR.

Refs #896

### DIFF
--- a/packages/e2e/playwright/src/tests/03-agent.spec.ts
+++ b/packages/e2e/playwright/src/tests/03-agent.spec.ts
@@ -46,6 +46,8 @@ test("create a mock agent with the connection attached", async ({ page }) => {
   await test.step("pick the mock image", async () => {
     await page.getByRole("button", { name: /create sandbox/i }).click();
     await page.getByText(harnessName, { exact: true }).click();
+    // The image step now only selects on click; advance with its Continue button.
+    await page.getByRole("button", { name: /continue/i }).click();
   });
 
   await test.step("name the sandbox and connect a provider", async () => {


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `c61ac4c`

## PR #1086: test(e2e): click Continue on the wizard image step in agent spec
**Author:** tomkis | **Branch:** fix/e2e-agent-image-continue → main | **Changes:** +2 −0 (1 files)

### Summary
Fixes the timing-out `03-agent.spec.ts` e2e test by clicking the image step's **Continue** button after selecting the mock image — the sandbox wizard redesign (#1040) made the image step select-only, so the spec no longer auto-advances to the sandbox-name field. Test-only change.

### Findings
- ✅ **Looks good:** The added `getByRole("button", { name: /continue/i }).click()` matches the Continue-advance pattern already used in this spec's later steps, and the inline comment explains *why* the click is needed (`packages/e2e/playwright/src/tests/03-agent.spec.ts:48-49`).
- ✅ **Looks good:** Root cause is correctly diagnosed and scoped — the click is added inside the existing `"pick the mock image"` test.step, immediately after the image selection, which is exactly where the wizard now requires the explicit advance.
- 🟢 **Suggestion (optional):** A case-insensitive `/continue/i` regex role-name match is fine here and consistent with the rest of the spec; no change needed. If the wizard ever renders more than one button matching `/continue/i` on this step, consider a test-id, but that's not a concern today.

### Verdict
APPROVE — Minimal, correct test fix that realigns the e2e spec with the shipped select-only image step; matches existing patterns in the same file.

---

PR #1088, opened 2026-06-16
Fix misleading 'can't reach server' warnings on first run

### LINKED ISSUE #1075: Fresh-install first run: blank white screen then misleading "can't reach server" warnings on the Terms page
## Problem

A brand-new user's very first interaction with the product looks broken, even when nothing is wrong.

On a fresh install, logging in for the first time (e.g. `dev`/`dev`) produces this sequence:

1. **A blank white screen for several seconds.** Nothing renders — no logo, no spinner — so it's unclear whether the app is loading, hung, or broken.
2. **Two warning toasts appear once the Terms of Use page loads:** *"Can't reach the server — agent list may be stale"* and *"Couldn't load secrets."*

Both warnings are misleading. The server is reachable and healthy — it's simply declining to hand over app data until the user has accepted the Terms of Use, which is the expected pre-acceptance state. But the Terms page tries to load that data anyway, the rejection is surfaced as a connectivity failure, and the warnings show up a few seconds *after* the page paints (so they feel like something just went wrong).

This is the product's first impression. A new user is told, before they've done anything, that the server is unreachable and secrets failed to load — which reads as "this thing is buggy" on a perfectly healthy install. The effect is most pronounced on a cold cluster (pods still warming), which is exactly when someone is evaluating a fresh deployment.

## Proposed solution

Make first run feel intentional and trustworthy:

- **Don't load (or warn about) app data before terms are accepted.** The Terms page should ask for nothing it isn't entitled to yet, so the two spurious warnings never appear for a new user.
- **Treat "terms not yet accepted" as a normal state, not an error.** Anywhere this response can occur, it should quietly route the user to the Terms page rather than surfacing a connectivity-failure warning — including if terms go stale mid-session.
- **Replace the blank white boot screen with an immediate branded loading state**, so the initial wait reads as "loading," not "broken."

### Scope

First-run / terms-acceptance experience and the initial app boot screen. Covers new users and any user whose accepted terms version is out of date.

### Out of scope

- Genuine server-unreachable handling (real outages) — that path already exists and should keep its warning.
- The content/versioning of the Terms of Use document itself.

PetrBulanek:
## What

A brand-new user's very first screen used to show two misleading warnings — *"Can't reach the server — agent list may be stale"* and *"Couldn't load secrets"* — even though the server was perfectly healthy. It was simply declining to hand over app data until the Terms of Use were accepted, and the UI surfaced that expected state as a connectivity failure.

This makes "Terms not yet accepted" a normal state instead of an error:

- The Terms page no longer requests app data it isn't entitled to yet, so the two spurious warnings never appear for a new user.
- A returning user whose accepted Terms version is out of date is quietly routed to re-accept, with a clear "the Terms of Use have been updated" note.
- Genuine server outages still show the real connectivity warning — only the terms-gate case is treated as normal.
- The terms gate now appears without a second full-page reload, and its loading text is legible instead of near-invisible.

## Note

The original report also mentioned a blank screen during boot. A branded boot screen was prototyped but dropped: the empty page paints instantly and the app mounts quickly, and an inlined boot overlay introduced more jank than it removed. The fix focuses on the warnings and the in-app loading state instead.

Closes #1075

### DIFF
--- a/packages/ui/src/main.tsx
+++ b/packages/ui/src/main.tsx
@@ -13,6 +13,7 @@ import { initAuth } from "./auth.js";
 import { applyBrand, loadBrand } from "./brand.js";
 import { preflightTermsGate } from "./modules/terms/lib/preflight.js";
 import { queryClient } from "./query-client.js";
+import { useStore } from "./store.js";
 
 async function main() {
   // Brand fetch is unauthenticated and runs in parallel with auth init so the
@@ -21,7 +22,10 @@ async function main() {
   const [user] = await Promise.all([initAuth(), loadBrand().then(applyBrand)]);
   if (!user) return; // Redirecting to Keycloak, don't render
 
-  if (!(await preflightTermsGate())) return;
+  if (!(await preflightTermsGate())) {
+    window.history.replaceState({}, "", "/terms");
+    useStore.setState({ view: "terms" });
+  }
 
   const { default: App } = await import("./app.js");
   createRoot(document.getElementById("root")!).render(
--- a/packages/ui/src/modules/terms/lib/on-terms-stale.ts
+++ b/packages/ui/src/modules/terms/lib/on-terms-stale.ts
@@ -1,8 +1,13 @@
 let redirecting = false;
+let termsStale = false;
 
-export function onTermsStale() {
-  if (redirecting) return;
-  if (window.location.pathname === "/terms") return;
+export function isTermsStale(): boolean {
+  return termsStale;
+}
+
+export function onTermsStale(): void {
+  termsStale = true;
+  if (redirecting || window.location.pathname === "/terms") return;
   redirecting = true;
   window.location.assign("/terms");
 }
--- a/packages/ui/src/app.tsx
+++ b/packages/ui/src/app.tsx
@@ -20,9 +20,6 @@ export default function App() {
   const view = useStore((s) => s.view);
   const theme = useStore((s) => s.theme);
 
-  useAgentCrashToasts();
-  useFirstRunRedirect();
-
   // Apply theme on mount + listen for system preference changes
   useEffect(() => {
     const apply = () => {
@@ -38,6 +35,16 @@ export default function App() {
     mq.addEventListener("change", apply);
     return () => mq.removeEventListener("change", apply);
   }, [theme]);
+
+  if (view === "terms") return <TermsView />;
+  return <MainApp />;
+}
+
+function MainApp() {
+  const view = useStore((s) => s.view);
+
+  useAgentCrashToasts();
+  useFirstRunRedirect();
 
   useEffect(() => {
     // The v2 wizard and the sandbox-creation wizard own their own OAuth-return
@@ -128,13 +135,6 @@ export default function App() {
         <ChatView />
         <DialogOverlay />
         <ConnectionBanner />
-      </>
-    );
-
-  if (view === "terms")
-    return (
-      <>
-        <TermsView />
       </>
     );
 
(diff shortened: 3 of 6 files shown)


packages/ui/src/main.tsx
```diff
@@ -21,7 +22,10 @@ async function main() {
   const [user] = await Promise.all([initAuth(), loadBrand().then(applyBrand)]);
   if (!user) return; // Redirecting to Keycloak, don't render
 
-  if (!(await preflightTermsGate())) return;
+  if (!(await preflightTermsGate())) {
```
dam-code-guardian, line 25: 🟢 **Suggestion:** This branch sets the store view but doesn't early-return — execution falls through to render `<App />`, which is intended (App renders `TermsView` when `view === "terms"`). Correct as written; a one-line comment would stop a future reader from "fixing" it by adding a return that skips the render.

```suggestion
  if (!(await preflightTermsGate())) {
    // Fall through to render <App />, which shows TermsView for view==='terms'.
    window.history.replaceState({}, "", "/terms");
    useStore.setState({ view: "terms" });
  }
```

packages/ui/src/modules/terms/lib/on-terms-stale.ts
```diff
@@ -1,8 +1,13 @@
 let redirecting = false;
+let termsStale = false;
```
dam-code-guardian, line 2: 🟢 **Suggestion:** `termsStale` is a one-way latch — once `onTermsStale()` sets it, nothing resets it to `false`. That's fine for the boot-time gate (the page hard-navigates on accept), but worth a brief comment noting the latch is intentional, in case a future flow expects to clear staleness without a full reload.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `07df3f0`

## PR #1088: Fix misleading 'can't reach server' warnings on first run
**Author:** PetrBulanek | **Branch:** fix/1075-first-run → main | **Changes:** +39 −23 (6 files)

### Summary
Reworks UI boot so an unaccepted/stale Terms-of-Use state renders the Terms page directly instead of surfacing as spurious "can't reach server"/"couldn't load secrets" warnings. `preflightTermsGate()` becomes a pure predicate, `main.tsx` sets the `terms` view via the store (no second full-page reload), `app.tsx` splits the terms route ahead of the data-fetching `MainApp`, and React Query suppresses outage toasts/mutation errors while terms are stale. Genuine outages still warn. _(No prior review on file.)_

### Findings
- ✅ **Looks good:** `preflightTermsGate()` simplified to a side-effect-free predicate (`return !!latest && latest.version === current.version`); the redirect responsibility now lives in the single caller (`main.tsx`). The `catch { return true }` correctly fails open so a transient terms-API error doesn't trap a healthy user behind the gate.
- ✅ **Looks good:** Hoisting `if (view === "terms") return <TermsView />` into the outer `App` (before `MainApp` mounts the toast hooks and fires app-data queries) is exactly what prevents the data fetches that produced the false warnings.
- ✅ **Looks good:** `isTermsStale()` gating in `query-client.ts` (both `onError` paths) is consistent with the existing `reconnecting` suppression and reuses the same module-level flag set by `onTermsStale()`.
- 🟢 **Suggestion:** In `main.tsx`, the terms-gate branch sets state without an early return — execution falls through to render `<App />` (intended, since App reads `view==="terms"`). A one-line comment would prevent a future reader from "fixing" it by adding a return that skips the render. (`packages/ui/src/main.tsx:25`)
- 🟢 **Suggestion:** `termsStale` is a latched module-level flag that's never reset to `false`. Fine for the boot-time gate (the page hard-navigates on accept), but a brief comment noting the latch is intentionally one-way would help if a future flow expects to clear staleness without a reload. (`packages/ui/src/modules/terms/lib/on-terms-stale.ts:2`)

### Verdict
APPROVE — Focused, well-reasoned fix that turns an expected gate state into a normal render path; correctness and failure-open behavior are sound, with only minor optional clarifying-comment suggestions.

---

kapetr (approved):

PR #1192, opened 2026-06-17
fix(ui): single toggle for Customize defaults overrides

### LINKED ISSUE #1095: UI - Group "Customize defaults" overrides into a single toggle
## Problem

For all of the Github "Add GitHub" dialogs, "Customize defaults" exposes 3-4 separate override checkboxes. These only make sense overridden together - using your own GitHub app means supplying all of them, not mixing presets with custom values. Replace the per-field checkboxes with a single toggle that overrides all fields at once.

## Screenshot

Image

tomkis:
## What

In every "Add <provider>" connection dialog, the **Customize defaults** section exposed one override checkbox per overridable field (client ID, client secret, app slug, …). These creds only make sense overridden *together* — using your own GitHub app means supplying all of its values, not mixing operator presets with custom ones per-field. The per-field checkboxes invited an invalid half-overridden state and added visual noise.

This replaces the per-field checkboxes with a **single toggle** that flips the whole overridable group at once:

- **Off (default):** presets are used; the section shows a compact read-only summary of what will be inherited (`Client ID: …`, or `… : preset value hidden` for secrets). Nothing overridable is submitted.
- **On:** every overridable field renders as an input and its value is submitted.

## How

Single-file change in `packages/ui/src/modules/connections/forms/template-create-form.tsx`:

- Collapse the per-field `overrides: Record<string, boolean>` state into one `overrideDefaults: boolean`. `submittedValue()` / `buildPayload()` wire format is unchanged.
- Rewrite `OverridableSection`: the toggle *is* the expander (no separate expanded state). Off → `PresetSummary` per field; on → `LabeledInput` per field.
- Add a small local `Switch` (`role="switch"`, Tailwind tokens, sliding knob) — there's no Switch primitive in `components/ui/` and `@radix-ui/react-switch` isn't a dependency, so this follows the dependency-free pattern used by the Anthropic `ModeToggle`. Carries `data-testid="override-defaults-toggle"` for e2e.

No contract or server change — purely how the override choice is collected in the UI.

## Verification

> [!NOTE]
> The pod's pnpm store was cold and the workspace install did not finish in this session, so I was unable to run `mise run check` / type-check or the dev server locally. The change is type-consistent with the contract (`ConnectionTemplateInput`) and confined to one file. **Please confirm CI (lint + tsc) is green before merging.**

Manual checks to run against an "Add GitHub" / bring-your-own-app template:
- Toggle replaces the per-field checkboxes; off by default.
- Off → preset summary shown; payload omits overridable fields.
- On → all overridable fields render as inputs; entered values submitted.
- A `credentialsFromFamily` template shows the family-reuse copy.

Closes #1095

### DIFF
--- a/packages/ui/src/modules/connections/forms/template-create-form.tsx
+++ b/packages/ui/src/modules/connections/forms/template-create-form.tsx
@@ -9,6 +9,7 @@ import { useMemo, useRef, useState } from "react";
 
 import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
+import { Switch } from "@/components/ui/switch";
 
 import { api } from "../../../api.js";
 import {
@@ -459,38 +460,6 @@ function PresetSummary({ input }: { input: ConnectionTemplateInput }) {
       </p>
     );
   return null;
-}
-
-function Switch({
-  checked,
-  onCheckedChange,
-  testId,
-  label,
-}: {
-  checked: boolean;
-  onCheckedChange: (v: boolean) => void;
-  testId?: string;
-  label?: string;
-}) {
-  return (
-    <button
-      type="button"
-      role="switch"
-      aria-checked={checked}
-      aria-label={label}
-      data-testid={testId}
-      onClick={() => onCheckedChange(!checked)}
-      className={`relative inline-flex h-5 w-9 shrink-0 items-center rounded-full transition-colors focus-visible:outline-hidden focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 ring-offset-background ${
-        checked ? "bg-primary" : "bg-input"
-      }`}
-    >
-      <span
-        className={`inline-block h-4 w-4 rounded-full bg-background shadow-sm transition-transform ${
-          checked ? "translate-x-4" : "translate-x-0.5"
-        }`}
-      />
-    </button>
-  );
 }
 
 function LabeledInput({
--- a/packages/ui/src/modules/connections/forms/template-create-form.tsx
+++ b/packages/ui/src/modules/connections/forms/template-create-form.tsx
@@ -8,7 +8,6 @@ import { Check, Copy, ExternalLink } from "lucide-react";
 import { useMemo, useRef, useState } from "react";
 
 import { Button } from "@/components/ui/button";
-import { Checkbox } from "@/components/ui/checkbox";
 import { Input } from "@/components/ui/input";
 
 import { api } from "../../../api.js";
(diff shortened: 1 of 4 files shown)


tomkis:
Screenshot 2026-06-17 at 10 56 38

packages/ui/src/modules/connections/forms/template-create-form.tsx
```diff
@@ -390,75 +386,113 @@ function OAuthAppHint({
...
+  if (input.secret)
+    return (
+      <p className="text-[11px] text-muted-foreground">
+        {labelFor(input.name)}: preset value hidden.
+      </p>
+    );
+  return null;
+}
+
+function Switch({
```
kapetr: This should be extracted into separate file in `components/ui`

kapetr (approved):

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `8f4f789`

## PR #1192: fix(ui): single toggle for Customize defaults overrides
**Author:** tomkis | **Branch:** fix/1095-customize-defaults-toggle → main | **Changes:** +94 −60 (1 file)

### Summary
Replaces the per-field override checkboxes in the connection "Customize defaults" section with one toggle that flips the whole overridable credential group at once — preventing the invalid half-overridden state where operator presets mix with custom per-field values. Single-file UI change; the `submittedValue()` / `buildPayload()` wire format is unchanged. (no prior review on file)

### Findings
- ✅ **Looks good:** Wire format preserved — `submittedValue()` now gates on the single `overrideDefaults` boolean (`input.state === "overridable" && !overrideDefaults`), and `buildPayload()` is untouched. Off → overridable fields omitted from the payload; on → entered values submitted. Matches the PR's stated contract.
- ✅ **Looks good:** The local `Switch` has solid a11y — `role="switch"`, `aria-checked`, `aria-label`, a `focus-visible` ring, and a `data-testid` for e2e. Confirmed there is no existing Switch primitive in `components/ui/` and `@radix-ui/react-switch` isn't a dependency, so the dependency-free local component is justified (same approach as the Anthropic mode toggle).
- ✅ **Looks good:** `PresetSummary` cleanly handles both the visible-preset (`presetValue`) and hidden-secret cases, returning `null` for the inapplicable case so nothing renders for a field with neither.
- 🟢 **Suggestion:** Extract the local `Switch` into a shared `components/ui/switch.tsx` primitive so the next consumer reuses it rather than re-implementing the pattern. _(Already raised inline by @kapetr at line 464 — surfaced here for completeness; not re-posted inline.)_

### Verdict
APPROVE — Sound UX fix that eliminates an invalid override state; wire format unchanged, good accessibility, confined to one file. Note the author's caveat that CI (lint + tsc) was not run locally — confirm CI is green before merge.

---

packages/ui/src/modules/connections/forms/template-create-form.tsx
tomkis: Good point. Will fix and investigate if we can catch this in agent earlier.

kapetr (approved):

Issue #444, opened 2026-06-01
Slack - "Authentication required (code -32000)" when non-owner messages agent in shared channel


jenna-winkler:
## Summary

When an agent owner connects their agent to a Slack channel and another linked user (`/dam login` completed) tries to message the agent, they get: `Error: Authentication required (code -32000)`

The agent owner can message the same agent in the same channel without issue.

## Steps to Reproduce

1. Agent owner (@matoushavlena) connects agent to channel C0B705DJXFB (IBM Test workspace)
2. Another user (@jenna-winkler) runs `/dam login` and links identity successfully
3. @jenna-winkler mentions `@DAM` in the channel
4. Response: `Error: Authentication required (code -32000)`

Meanwhile, @matoushavlena can `@DAM` in the same channel and gets a normal response.

## Expected Behavior

Per the allowed users logic: if the allowed users list is empty ("Unrestricted — any linked Slack user can interact"), any linked user should be able to interact with the agent. The fork pod should spin up with the invited user's credentials and respond normally.

## Additional notes

The invited user (@jenna-winkler) can create their own separate channel and connect their own agent to it - that works fine.

matoushavlena:
Related: https://github.com/dam-agents/dam/issues/91

tomkis:
Seemed to have been fixed earelier https://github.com/dam-agents/dam/pull/1359 proves that.

Issue #91, opened 2026-05-05
Slack thread replies from non-owners can't reach external services


pilartomas:
## Problem

When a Slack user other than the instance owner replies in an agent's thread, the platform spins up a short-lived per-turn run of the agent under the replier's identity, so the agent uses the **replier's** credentials and rules, not the original owner's. That per-turn run can't reach anything external today — every credentialed call is rejected at the credential gateway, with no inbox prompt and no clear recovery path. From the replier's seat the agent appears to silently fail.

This is the whole point of letting multiple people collaborate on a single agent thread in Slack. Today the feature is functionally broken: a non-owner can talk to the agent, but the agent can't call any external service on their behalf — no GitHub, no model API, no internal services.

## Proposed solution

The credential gateway should recognize per-turn replier runs as their own first-class identity, look up the **replier's** allowlist and credentials, and let approved calls through. From the replier's seat the agent should behave the same as it does for the original owner: the same APIs are reachable, the same approval prompts surface in the inbox when a host hasn't been allowed yet, and any approval the replier grants applies to their own future runs.

### Dependencies

Blocks the multi-user collaboration story tracked in ADR-027.

jenna-winkler:
Blocked by #105

tomkis:
Seemed to have been fixed earelier https://github.com/dam-agents/dam/pull/1359 proves that.

PR #1359, opened 2026-06-18
test(e2e): assert foreign-fork credential injected on egress

tomkis:
Adds a CI guard that a non-owner Slack user's fork injects that user's own credential on outbound calls.

While building this I found #444 and #91 already appear to be resolved on main: the foreign fork resolves and injects the replier's own credential, so the behavior both tickets describe as broken now works. This PR just locks that in with a test, it is not a fix.

## What it does

- **Mock agent**: a `__FETCH__ <url>` prompt directive makes the mock perform an authenticated proxy fetch through its gateway Envoy during the turn, then reply with the echoed result. Without the directive it falls through to the existing static-script behavior, so the tracer and other specs are untouched.
- **07-slack**: the foreign user (dev2) creates their own httpbingo connection with a distinct sentinel, then a foreign mention `__FETCH__ <echoUrl>` forks the agent. The fork pair mounts the replier's own credential Secrets (`owner=<dev2-sub>`), so the echoed request headers must carry dev2's value. The assertion requires the foreign sentinel in the echoed headers.
- Mirrors the owner-path proof in 05-injection, but from a fork.

## Note

Originally written as a `test.fail()` repro for the foreign-fork egress gap. That path already works on main, so the assertion holds. Dropped the annotation so it stands as a passing regression guard.

## Validation

Relying on CI `mise run e2e` (the foreign sentinel comes back in the echoed headers). Not validated on a local cluster, since the e2e test VM can't bootstrap alongside the running dev clusters due to k3s host-port collisions.

### DIFF
--- a/packages/e2e/playwright/src/tests/07-slack.spec.ts
+++ b/packages/e2e/playwright/src/tests/07-slack.spec.ts
@@ -4,7 +4,17 @@ import { baseUrl, testUser2 } from "../config.js";
 import { waitForAgentRunning } from "../lib/agents.js";
 import { createApiClient } from "../lib/api-client.js";
 import { getAccessToken } from "../lib/auth.js";
-import { agentName } from "../lib/fixtures.js";
+import { ensureCustomHeaderConnection } from "../lib/connections.js";
+import {
+  agentName,
+  connectionHost,
+  echoUrl,
+  foreignConnectionName,
+  foreignEnvName,
+  foreignSentinel,
+  headerName,
+  valueFormat,
+} from "../lib/fixtures.js";
 
 const slackChannelId = "C-E2E-TRACER";
 const foreignSlackUserId = "U-E2E-FOREIGN";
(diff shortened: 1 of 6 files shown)


tomkis:
Work continuation for #444

packages/e2e/playwright/src/tests/07-slack.spec.ts
```diff
@@ -133,3 +143,54 @@ test("foreign user mention forks the agent and the reply lands back in the threa
...
+  await test.step("foreign mention forks and the fork egress carries the foreign credential", async () => {
+    await api.e2e.slackResetOutbound.mutate();
+    await api.e2e.slackFireMention.mutate({
+      user: foreignSlackUserId,
+      channel: slackChannelId,
+      ts: "1700000002.000100",
+      text: `__FETCH__ ${echoUrl}`,
+    });
+
+    let replyText = "";
```
dam-code-guardian, line 175: 🟢 **Suggestion:** This matcher also accepts a `[fetch error] …` reply (the mock emits that on an egress failure). If egress regresses, the poll succeeds on the error reply and the test then fails at `toContain(foreignSentinel)` with a misleading "sentinel missing" message rather than "egress failed". Consider asserting `replyText` does not start with `[fetch error]` before the sentinel check, so the failure mode is diagnosable.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `52e7ca9`

## PR #1359: test(e2e): assert foreign-fork credential injected on egress
**Author:** tomkis | **Branch:** feat/slack-e2e-444-repro → main | **Changes:** +136 −6 (6 files)

### Summary
Adds an e2e regression guard that a foreign Slack user's fork injects *that user's own* credential on outbound calls. A new `__FETCH__ <url>` mock-agent directive performs an authenticated proxy fetch through the fork's gateway Envoy and echoes the result; the spec asserts the foreign sentinel comes back in the echoed headers. Test-only — no production code paths change (the author notes #444/#91 already work on `main`; this locks the behavior in).

### Findings
- ✅ **Looks good:** `proxyFetch` is now constructed once in `composeScriptedMock()` and shared between the scripted-mock service and the ACP service, rather than each building its own — clean dependency wiring.
- ✅ **Looks good:** `promptText()` defensively handles string payloads, content-block arrays, and malformed/non-object blocks without throwing, and `replyWithFetch()` wraps the fetch in try/catch so an egress failure still produces a `[fetch error]` reply instead of hanging the turn.
- ✅ **Looks good:** `__FETCH__` directive falls through to the existing static-script behavior when absent, so the tracer and other specs are untouched.
- ✅ **Looks good:** `getAccessToken(user)` defaulting to `testUser` keeps every existing caller working while enabling the foreign-user (`testUser2`) token path.
- 🟢 **Suggestion:** The poll only matches replies containing `[fetch ` — a `[fetch error] …` reply (e.g. egress blocked) also matches and would then fail at `toContain(foreignSentinel)` with a confusing message. Consider asserting the reply is not a `[fetch error]` first, so a genuine egress regression reports as "egress failed" rather than "sentinel missing". (`packages/e2e/playwright/src/tests/07-slack.spec.ts:175`)

### Verdict
APPROVE — Well-scoped, defensive regression guard for foreign-fork credential injection; the one note is an optional diagnostics improvement.

---

willallstet:
📝 Docs reviewed for this change — no documentation update required (no code changes after stripping docs/test-only files; nothing to document.)

PR #1372, opened 2026-06-18
fix(cli): clear messages when the server rejects a command for auth

### LINKED ISSUE #1331: CLI shows a "cannot reach server" error when the real problem is an expired login
## Problem

When a `dam` CLI command runs with an expired or invalid login, the CLI fails with a message like:

```
cannot reach server `http://localhost:4444`: Unable to transform response from server
```

That reads like a network or wrong-URL problem — but the server is reachable and healthy; it rejected the request because the session is no longer valid. The user ends up debugging the wrong thing (is the server down? is my URL wrong?) when the actual fix is simply to log in again.

This is easy to hit in everyday use because access tokens are short-lived, so any command run a few minutes after authenticating can surface this confusing error.

## Proposed solution

When the server rejects a command because the caller isn't authenticated, the CLI should say so plainly and point at the remedy — for example:

```
Not authenticated: your session has expired. Run `dam auth login` to sign in again.
```

The CLI already gives a clear, specific message for the related "Terms of Use acceptance required" case; this brings the not-authenticated case up to the same bar.

### Scope

Just the message shown for unauthenticated responses — recognize that "rejected for auth" is distinct from "couldn't reach the server," and word it accordingly.

### Out of scope

- Automatic token refresh or silent re-login — separate concern.

PetrBulanek:
## What

When a `dam` CLI command runs with an expired or invalid login, the CLI used to fail with a message that looked like a network or wrong-URL problem:

```
error: cannot reach server `http://localhost:4444`: Unable to transform response from server
```

…even though the server was reachable and healthy — it had simply rejected the request because the session was no longer valid. Because access tokens are short-lived, this was easy to hit minutes after logging in, and it sent people debugging the wrong thing.

Now the CLI says what actually happened and points at the right remedy, consistently across every command group (`agent`, `file`, `auth token`, `import`, …):

- **Session expired / not authenticated:**
  ```
  error: not authenticated: session expired for http://localhost:4444
  hint: run `dam auth login` first
  ```
- **Access denied** (authenticated but not authorized — e.g. a pending-approval account):
  ```
  error: access denied: Access pending approval. Contact your administrator.
  ```
- **`DAM_TOKEN` rejected** (headless / CI): when the bearer comes from `DAM_TOKEN` it is sent verbatim and never lives in `auth.toml`, so `dam auth login` can't fix it — the env token itself was rejected. The hint points there instead:
  ```
  error: not authenticated: session expired for http://localhost:4444
  hint: DAM_TOKEN was rejected — check it is valid and unexpired
  ```

Genuine connectivity failures still report “cannot reach server,” and the existing “Terms of Use acceptance required” message is unchanged.

## How

- The tRPC client recognizes the api-server's raw `401`/`403` responses — alongside the existing `412` Terms gate — and maps them to typed errors (`401` → not-authenticated, `403` → access-denied). Previously tRPC couldn't deserialize these raw-JSON gate bodies and mislabeled them as transport failures.
- The auth-rejection message **and** remedy now live in one place (`shared/auth-message.ts`), so every entry point agrees — the tRPC path *and* the raw-fetch paths (`dam import`, which uploads over `fetch` rather than tRPC). That's what keeps the `DAM_TOKEN` hint consistent everywhere instead of only on the tRPC path.
- Consolidated the duplicated transport/auth renderers into a single `print.ts` and dropped the `printServiceError` re-export shim that `errors.ts` carried (pure refactor, no behavior change).
- Documented the server-rejection surfacing in the CLI architecture page.

Closes #1331


dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `1310050`

## PR #1372: fix(cli): clear messages when the server rejects a command for auth
**Author:** PetrBulanek | **Branch:** fix/1331-cli-auth-rejection-messages → main | **Changes:** +71 −18 (7 files)

### Summary
Replaces the misleading `cannot reach server … Unable to transform response` message the CLI showed when the server rejected a request for auth reasons. A new `wrapFetchWithServerGates` fetch wrapper inspects 401/403/412 responses and maps the server's `error` envelope to typed transport errors (`unauthorized` → `AuthRequiredAtTransportError` → `not authenticated: session expired … / hint: dam auth login`; `forbidden` → `ForbiddenAtTransportError` → `access denied: <detail>`; `terms_stale` unchanged), with a shared `printTrpcError` helper distinguishing app-layer rejections from genuine connectivity failures. No prior PR comments/reviews. No PR-local overrides on file.

### Findings
- 🟡 **Warning:** The new 401/403 server-gate branches in `wrapFetchWithServerGates` and the whole of `printTrpcError` have no test coverage, even though `packages/cli/src/__tests__/trpc-client.test.ts` already exercises the 200 and not-logged-in paths and is the obvious home for `unauthorized`/`forbidden`/`terms_stale` cases (`packages/cli/src/modules/shared/trpc/trpc-client.ts:60`, `packages/cli/src/modules/shared/trpc/print.ts:4`). No test files were touched in this PR.
- 🟢 **Suggestion:** The `error: not authenticated: <reason>` + `hint: run dam auth login first` wording is now duplicated across ~6 call sites (`print.ts` plus `file/get|list|put`, `agent/create`, `agent/commands/errors`, `agent/create-interactive`). Consider having those commands route through `printTrpcError` (or a small shared formatter) so the auth-rejection wording can't drift between command groups (`packages/cli/src/modules/shared/trpc/print.ts:7`).
- ✅ **Looks good:** `response.clone()` before `clone.json()` preserves the original stream for tRPC when the body doesn't match a known gate; re-throwing the sentinel errors from the fetch wrapper and routing `ForbiddenAtTransportError`/`TermsStaleAtTransportError` to the top-level `bin.ts` handler keeps the per-command catch blocks simple; the `classify.ts` cause-chain walk is depth-bounded (≤8).

**Verdict: minor drift (possible)**

_Possible drift (⚠️):_
- **Check #1 — CLI error-surfacing behavior** (`packages/cli/src/modules/shared/trpc/trpc-client.ts:60-97`, `print.ts:4-19`): the PR adds a server-side 401/403 transport gate alongside the existing 412 terms gate, changing how the CLI surfaces server rejections (401 → `not authenticated: session expired`; 403 → `access denied`). `docs/architecture/cli.md` documents the client-side `session-expired` path (TokenProvider precedence) but does not mention the new server-rejection classification. Judgement question for the author: does `cli.md` warrant a sentence on the 401/403→typed-error mapping, or is this below the architecture-page threshold? `Last verified:` staleness does not apply — no architecture page was edited in this diff.

### Verdict
COMMENT — Correct and cleanly structured; the only substantive gap is missing test coverage for the new 401/403 gates, plus a possible doc note in `cli.md`.

---

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `e9a5aae`

## PR #1372: fix(cli): clear messages when the server rejects a command for auth
**Author:** PetrBulanek | **Branch:** fix/1331-cli-auth-rejection-messages → main | **Changes:** +196 −126 (52 files)

### Summary
Re-review after the author addressed the prior review (`1310050`): the auth-rejection error-printing is now consolidated into a new shared `packages/cli/src/modules/shared/trpc/print.ts` (`formatTransportError` / `printServiceError` / `printTrpcError`), with ~40 command files routed through it instead of duplicating the `not authenticated …` + `hint: dam auth login` wording. `printServiceError` gained a `DAM_TOKEN`-aware hint, a `print.test.ts` was added, and `docs/architecture/cli.md` now documents the 401/403→typed-error mapping. No human PR comments/reviews. No PR-local overrides on file.

### Changes since last review
Previous HEAD: `1310050` (2026-06-18T16:30:47Z) — verdict COMMENT

- ✅ **Fixed:** auth-rejection wording duplication — consolidated into the new shared `print.ts`; ~40 command files now import `printServiceError`/`printTrpcError` from it (`packages/cli/src/modules/shared/trpc/print.ts`).
- ✅ **Fixed:** doc-drift note on `cli.md` — added a paragraph documenting the 401/403→typed-error mapping and bumped `Last verified:` to 2026-06-19 (`docs/architecture/cli.md`).
- 🔁 **Still present (narrowed):** test coverage — `print.test.ts` now covers `printServiceError`'s hint branches, but the 401/403 fetch gate and `printTrpcError` remain untested (`packages/cli/src/modules/shared/trpc/trpc-client.ts:60`).
- 🆕 **New:** `DAM_TOKEN`-aware hint — `printServiceError` switches the hint to "DAM_TOKEN was rejected …" when the env var is set (commit `e9a5aae`); plus the minor dead-guard noted below (`print.ts:32`).

### Findings
- 🟡 **Warning:** The 401/403 server-rejection gate in `wrapFetchWithServerGates` and `printTrpcError`'s app-layer-vs-connectivity branch still lack test coverage; `print.test.ts` (new) covers `printServiceError`'s `DAM_TOKEN`/login hint branches, and `trpc-client.test.ts` still exercises only the 200 and not-logged-in paths (`packages/cli/src/modules/shared/trpc/trpc-client.ts:60`, `print.ts:30`).
- 🟢 **Suggestion:** `printTrpcError`'s `if (r.ok) return;` guard is unreachable — `classifyTrpcError` is typed `Result<never, …>` and at runtime only ever returns an `err` result or throws, so `r.ok` can never be true. Harmless but dead (`packages/cli/src/modules/shared/trpc/print.ts:32`).

✅ No documentation drift detected. The PR changes CLI error-surfacing behavior (adds the 401/403 server-rejection gates) and updates the matching architecture page `docs/architecture/cli.md` in the same PR — the new paragraph accurately describes the 401→`not authenticated: session expired`, 403→`access denied`, and `DAM_TOKEN` hint-switch mapping, and `Last verified:` was bumped to the PR date (2026-06-19). No ADR leaks, no volatile-content leaks, no missing pages, no cross-reference rot.

### Verdict
COMMENT — The two substantive points from the prior review (duplication, doc drift) are resolved and partial test coverage was added; the only remaining gap is the still-untested 401/403 fetch gate, which is non-blocking.

---

packages/cli/src/modules/shared/trpc/print.ts
```diff
@@ -0,0 +1,43 @@
...
+        : "hint: run `dam auth login` first\n",
+    );
+    return;
+  }
+  process.stderr.write(`error: ${formatTransportError(error.reason, host)}\n`);
+}
+
+export function printTrpcError(e: unknown, host: string): void {
+  const r = classifyTrpcError(e);
+  if (r.ok) return;
```
dam-code-guardian, line 27: 🟢 **Suggestion:** This early return is unreachable. `classifyTrpcError` is typed `Result<never, TransportError | AuthRequiredError>` — its success variant carries `never`, and at runtime it only ever returns an `err(...)` result or throws (`terms_stale`/`forbidden`). So `r.ok` can never be true here. Harmless, but the guard is dead code and can be dropped.

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `973f1e4`

## PR #1372: fix(cli): clear messages when the server rejects a command for auth
**Author:** PetrBulanek | **Branch:** fix/1331-cli-auth-rejection-messages → main | **Changes:** +186 −132 (54 files)

### Summary
Re-review after two new commits on top of `e9a5aae`. The `DAM_TOKEN`-aware auth-rejection wording is now extracted into a shared `formatAuthRejection` in the new `packages/cli/src/modules/shared/auth-message.ts` (single source of truth), `printServiceError` and the raw-fetch `dam import` path both route through it, and `printTrpcError`'s previously-flagged `if (r.ok) return;` guard is now documented as a required TS type-narrowing guard. `print.test.ts` was replaced by `auth-message.test.ts` testing the helper directly. No human PR comments/reviews. No PR-local overrides on file.

### Changes since last review
Previous HEAD: `e9a5aae` (2026-06-19T08:57:27Z) — verdict COMMENT

- ✅ **Fixed:** the dead-guard suggestion — `printTrpcError`'s `if (r.ok) return;` is now annotated as a deliberate type-narrowing guard (`classifyTrpcError` is typed `Result<never, …>` so TS needs the check before `r.error`); kept and explained rather than removed, which is the correct resolution since removing it breaks type-checking (`packages/cli/src/modules/shared/trpc/print.ts:24`).
- 🆕 **New (improvement):** `DAM_TOKEN` hint consolidated into shared `formatAuthRejection` and applied to the `dam import` raw-fetch path too — previously `import.ts` printed a hard-coded `run "dam auth login" first` that ignored `DAM_TOKEN`, so the hint is now consistent across the tRPC and raw-fetch entry points (`packages/cli/src/modules/shared/auth-message.ts`, `packages/cli/src/modules/import/commands/import.ts:173,227`).
- 🔁 **Still present:** the 401/403 server-rejection fetch gate (`wrapFetchWithServerGates`) and `printTrpcError`'s app-layer-vs-connectivity branch remain untested; the new `auth-message.test.ts` covers the helper's hint branches but not the fetch-gate classification (`packages/cli/src/modules/shared/trpc/trpc-client.ts:60`).

### Findings
- 🟡 **Warning:** The 401/403 server-rejection gate in `wrapFetchWithServerGates` and `printTrpcError`'s app-layer-vs-connectivity branch still lack test coverage. `auth-message.test.ts` (new) tests `formatAuthRejection`'s `DAM_TOKEN`/login branches directly, but the fetch-gate classification (401 → not-authenticated, 403 → access-denied, fall-through → transport error) is untested and `trpc-client.test.ts` still exercises only the 200 and not-logged-in paths (`packages/cli/src/modules/shared/trpc/trpc-client.ts:60`).
- ✅ **Looks good:** `formatAuthRejection` takes `env` as a defaulted parameter, keeping it pure and unit-testable; routing `dam import` through it removes the last divergent copy of the auth-rejection wording so the `DAM_TOKEN` hint can no longer drift between the tRPC and raw-fetch paths; the type-narrowing guard is now self-documenting.

✅ No documentation drift detected. The two new commits do not change subsystem behavior beyond what `docs/architecture/cli.md` already documents (the server-rejection 401/403 mapping and the `DAM_TOKEN` hint switch were added to the page in the prior commit and remain accurate); `Last verified:` is 2026-06-19. No ADR references, no volatile-content leaks, no missing subsystem pages, no cross-reference rot.

### Verdict
COMMENT — Both points from the prior review are now resolved (dead-guard explained, `DAM_TOKEN` hint unified across entry points incl. `dam import`); the only remaining gap is the still-untested 401/403 fetch gate, which is non-blocking.

---

packages/cli/src/__tests__/auth-message.test.ts
```diff
@@ -0,0 +1,17 @@
+import { describe, it, expect } from "vitest";
```
jjeliga: Do you deem the test to be actually useful or should we rather stick to the no-unit-tests convention?

packages/cli/src/modules/shared/auth-message.ts
```diff
@@ -0,0 +1,22 @@
+import { DAM_TOKEN_ENV_VAR } from "../auth/infrastructure/auth-env-reader.js";
+
+/**
```
jjeliga: trim this

packages/cli/src/modules/shared/trpc/trpc-client.ts
```diff
@@ -49,22 +57,39 @@ function buildAuthHeaders(deps: {
...
+function wrapFetchWithServerGates(
   host: string,
   baseFetch?: typeof fetch,
 ): typeof fetch {
   const inner = baseFetch ?? fetch;
   return (async (input, init) => {
     const response = await inner(input, init);
-    if (response.status === 412) {
+    if (
+      response.status === 401 ||
```
jjeliga: you could compact to one-line

packages/cli/src/modules/shared/trpc/trpc-client.ts
```diff
@@ -49,22 +57,39 @@ function buildAuthHeaders(deps: {
...
+      response.status === 403 ||
+      response.status === 412
+    ) {
       const clone = response.clone();
```
jjeliga, line 68: this try-catch reads a bit weird and conflates things, how about this?
Do not apply directly in gh, it would most likely break stuff


```suggestion
const clone = response.clone();
let body: { error?: string; message?: string };
try {
  body = (await clone.json()) as { error?: string; message?: string };
} catch {
  return response; // non-JSON gate response — let tRPC handle the status
}
if (body.error === "terms_stale") throw new TermsStaleAtTransportError(host);
if (body.error === "unauthorized")
  throw new AuthRequiredAtTransportError(`session expired for ${host}`);
if (body.error === "forbidden")
  throw new ForbiddenAtTransportError(
    typeof body.message === "string" ? body.message : "",
  );
return response;

```

packages/cli/src/__tests__/auth-message.test.ts
PetrBulanek: It slipped through. Removing.

packages/cli/src/modules/shared/auth-message.ts
PetrBulanek: Good point. Trimming/removing.

packages/cli/src/modules/shared/trpc/trpc-client.ts
PetrBulanek: You are right, I'll address it.

jjeliga (approved):

willallstet:
📝 Docs reviewed for this change — no documentation update required (CLI error-handling refinement (clearer auth-rejection messages); not a documented contract.)

PR #1378, opened 2026-06-19
feat(helm): allow custom annotations on chart workloads

### LINKED ISSUE #1375: Helm chart can't set custom annotations on deployed workloads
## Problem

When installing the platform via its Helm chart, an operator has no way to add their own annotations to the components the chart deploys (API server, UI, controller, Keycloak, the bundled datastores, and the one-off setup jobs). Annotations are the standard mechanism Kubernetes operators and tooling use to opt a workload into extra behavior, so not being able to set them blocks a whole class of integrations.

The motivating case is **OpenTelemetry auto-instrumentation**: its injection operator watches for specific annotations on a pod and, when it sees them, injects the instrumentation sidecar/init-container automatically. Today there's no place to put those annotations, so the platform's services can't be auto-instrumented without forking the chart or hand-patching live resources after every deploy. The same gap blocks anything else that's annotation-driven — service-mesh opt-in, backup tooling, custom schedulers, Prometheus scrape hints, and so on.

Support is also inconsistent: one minor component (the NFS provisioner) already exposes annotation settings, while none of the primary services do — so the chart half-supports a feature that operators reasonably expect everywhere.

## Proposed solution

Let chart users supply custom annotations for the deployed workloads, at both levels Kubernetes distinguishes:

- **Pod-template annotations** — what injection operators like OpenTelemetry key off of.
- **Workload-level annotations** — on the Deployment/StatefulSet/Job object itself.

Expose this consistently across all components the chart deploys, with a sensible split between a chart-wide default that applies to everything and a per-component override for cases where only one service needs an annotation. Empty by default, so existing installs are unaffected.

### Scope

All workloads the chart renders — the long-running services and the setup/migration jobs alike — so observability and policy tooling can target the whole platform uniformly.

### Out of scope

Annotations on the agent pods the controller launches at runtime (a separate, already-partially-wired path), and annotations on non-workload resources such as Ingress and Services, which are handled separately.

pilartomas:
Closes #1375

## Problem

The chart gave operators no way to annotate the workloads it deploys. The motivating case is **OpenTelemetry auto-instrumentation**, whose injection operator triggers on pod-template annotations — with nowhere to put them, the platform's services couldn't be auto-instrumented without forking the chart or hand-patching live resources. Only the NFS provisioner had annotation knobs; the primary services had none.

## Change

Two shared template helpers (`platform.podAnnotations`, `platform.annotations`) merge annotations from three layers and render the block only when non-empty:

- **`commonPodAnnotations` / `commonAnnotations`** (new, chart-wide) — applied to every workload's pod template / object metadata.
- **`<component>.podAnnotations` / `<component>.annotations`** (per-component override) — added to `apiServer`, `ui`, `controller`, `keycloak`, `postgres`, `redis` (NFS provisioner already had them, now routed through the helpers so the chart-wide defaults reach it too).
- **chart-internal** (highest precedence) — config checksums and Helm hook directives are passed through as `internal` so rollout-on-config-change and hook behavior are preserved.

Precedence, highest first: **internal > per-component > chart-wide**. All knobs default to `{}` — existing installs are unaffected.

Covers all 9 workloads: api-server, UI, controller, Keycloak, Postgres, Redis, NFS provisioner, and both setup/migration Jobs.

## Verification

`mise run helm:check:lint` and `mise run helm:check:render` (kubeconform) pass. Rendered-output checks:

| Scenario | Result |
|---|---|
| Defaults | Helm hooks + config checksums intact; no stray annotation blocks |
| Chart-wide defaults set | land on all 8 enabled workloads + their pod templates |
| Per-component override | wins on that component only |
| Chart-wide + per-component | both keys **merge** (correct precedence) |
| Hook Jobs + user annotation | keep every hook directive **and** gain the annotation |

### DIFF
--- a/deploy/helm/platform/templates/_helpers.tpl
+++ b/deploy/helm/platform/templates/_helpers.tpl
@@ -56,6 +56,48 @@ imagePullSecrets:
 {{- end }}
 
 {{/*
+podAnnotations — merged pod-template annotations for a workload. This is the
+hook OpenTelemetry auto-instrumentation and similar injection operators key
+off of. Precedence, highest first: chart-internal (e.g. config checksums that
+must survive so a `helm upgrade` rolls the pod) > per-component
+(`<component>.podAnnotations`) > chart-wide (`.Values.commonPodAnnotations`).
+Renders an `annotations:` block, or nothing when the merged map is empty.
+
+Usage:
+  {{- include "platform.podAnnotations" (dict "root" $ "component" .Values.apiServer "internal" $checksums) | nindent 6 }}
+  - root:      root context ($), for the chart-wide defaults
+  - component: component values subtree (may define `.podAnnotations`); optional
+  - internal:  chart-managed annotations that must always render; optional
+*/}}
+{{- define "platform.podAnnotations" -}}
+{{- $component := .component | default dict -}}
+{{- $internal := .internal | default dict -}}
+{{- $merged := merge (deepCopy $internal) ($component.podAnnotations | default dict) (.root.Values.commonPodAnnotations | default dict) -}}
+{{- with $merged }}
+annotations:
+  {{- toYaml . | nindent 2 }}
+{{- end }}
+{{- end }}
+
+{{/*
+annotations — merged workload object-metadata annotations (on the
+Deployment/StatefulSet/Job itself). Precedence, highest first: chart-internal
+(e.g. Helm hook directives) > per-component (`<component>.annotations`) >
+chart-wide (`.Values.commonAnnotations`). Renders an `annotations:` block, or
+nothing when the merged map is empty. Same `dict` arguments as
+`platform.podAnnotations`, reading `.annotations` instead of `.podAnnotations`.
+*/}}
+{{- define "platform.annotations" -}}
+{{- $component := .component | default dict -}}
+{{- $internal := .internal | default dict -}}
+{{- $merged := merge (deepCopy $internal) ($component.annotations | default dict) (.root.Values.commonAnnotations | default dict) -}}
+{{- with $merged }}
+annotations:
+  {{- toYaml . | nindent 2 }}
+{{- end }}
+{{- end }}
+
+{{/*
 nameList — comma-separated .name values from a list of objects.
 Usage: {{ include "platform.nameList" .Values.someList }}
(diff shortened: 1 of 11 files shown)


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `68ee8ad`

## PR #1378: feat(helm): allow custom annotations on chart workloads
**Author:** pilartomas | **Branch:** feat/helm-workload-annotations → main | **Changes:** +116 −26 (11 files)

### Summary
Adds two shared Helm helpers (`platform.podAnnotations`, `platform.annotations`) that merge chart-wide, per-component, and chart-internal annotation layers, and wires per-component knobs into all 9 workloads. Motivated by OpenTelemetry auto-instrumentation needing pod-template annotations. All knobs default to `{}`, so existing installs are unaffected.

### Findings
- ✅ **Looks good:** Precedence is implemented correctly — `merge (deepCopy $internal) ($component.podAnnotations) (.root.Values.commonPodAnnotations)` gives internal > per-component > chart-wide, matching the documented contract. `deepCopy` on `$internal` correctly prevents sprig's destination-mutating `merge` from clobbering the caller's dict.
- ✅ **Looks good:** Config checksums (api-server, ui) and Helm hook directives (keycloak-provision, migrate-stale-netpols) are routed through the `internal` layer, so rollout-on-config-change and hook behavior are preserved at highest precedence.
- ✅ **Looks good:** The helpers render the `annotations:` block only `with` a non-empty merged map, so defaults produce no stray empty blocks. The NFS provisioner's bespoke `with` blocks are correctly collapsed into the shared helper.
- 🟢 **Suggestion:** `values.yaml` documents the annotation knobs well; consider a one-line note that the setup/migration Jobs receive `commonPodAnnotations` but not per-component overrides (they have no component subtree), so readers know hook Jobs only honor the chart-wide layer.

### Verdict
APPROVE — Clean, well-tested refactor with correct merge precedence and preserved hook/checksum behavior; no architecture-doc drift (Helm template plumbing is deployment-tree detail, out of scope for architecture pages).

---

willallstet:
📝 Docs reviewed for this change — no documentation update required (operator-facing Helm chart config (common/per-component annotations); the docs do not document chart values.)

PR #1384, opened 2026-06-19
feat(ui): emit nginx access logs as JSON

pilartomas:
## What

The UI is a static React SPA served by nginx — there's no application logger, so the only server logs are nginx's access/error logs, which were in the default plaintext `combined` format. This switches them to **structured JSON**, one object per line on stdout.

In-cluster the UI pod's nginx reverse-proxies same-origin `/api/` (tRPC + REST + WebSockets) to the api-server, so these access logs capture **all API traffic through the edge**, not just static-asset hits.

## Changes

- `deploy/helm/platform/templates/ui/app.yaml` (the live `*-ui-nginx` ConfigMap) — adds `log_format json_access escape=json` (http context, alongside the existing `map`) and `access_log /dev/stdout json_access;` in the server block. Fields: `time, remote_addr, method, uri, status, body_bytes_sent, request_time, upstream_response_time, referer, user_agent, x_forwarded_for`.
- `packages/ui/default.conf` (image-baked, used only outside k8s) — same format (minus the proxy-only fields), kept consistent to avoid drift.

## Notes

- **No `$otel_trace_id` field.** It requires a loaded nginx OTel module; referencing it with no module crashes nginx at startup. The JSON structure is in place so adding that one field is trivial later.
- **Trace propagation is intentionally out of scope.** The OTel Operator's `inject-nginx` uses a module supporting only nginx 1.22–1.23.1; our image is newer. Rather than downgrade or hand-bake the native `ngx_otel_module`, propagation waits until the operator supports this nginx version. (Tracking issue to follow.)

## Verification

- `mise run helm:check:lint` and `mise run helm:check:render` (kubeconform) pass; rendered ConfigMap confirmed correct (nginx `$vars` survive the YAML block scalar, upstream interpolates).
- nginx syntax not validatable locally (no binary/cached image); the authoritative check is `mise run cluster:build-ui` → confirm the pod starts and access lines are single JSON objects.

### DIFF
--- a/deploy/helm/platform/templates/ui/app.yaml
+++ b/deploy/helm/platform/templates/ui/app.yaml
@@ -16,10 +16,28 @@ data:
         ''      '';
     }
 
+    # Structured JSON access log, one object per line on stdout.
+    log_format json_access escape=json
+        '{'
+        '"time":"$time_iso8601",'
+        '"remote_addr":"$remote_addr",'
+        '"method":"$request_method",'
+        '"uri":"$request_uri",'
+        '"status":$status,'
+        '"body_bytes_sent":$body_bytes_sent,'
+        '"request_time":$request_time,'
+        '"upstream_response_time":"$upstream_response_time",'
+        '"referer":"$http_referer",'
+        '"user_agent":"$http_user_agent",'
+        '"x_forwarded_for":"$http_x_forwarded_for"'
+        '}';
+
     server {
         listen {{ .Values.ui.port }};
         root /usr/share/nginx/html;
         index index.html;
+
+        access_log /dev/stdout json_access;
 
         location /api/ {
             proxy_pass http://{{ $apiserverName }}:{{ .Values.apiServer.port }};
--- a/packages/ui/default.conf
+++ b/packages/ui/default.conf
@@ -1,6 +1,22 @@
+# Structured JSON access log, one object per line on stdout.
+log_format json_access escape=json
+    '{'
+    '"time":"$time_iso8601",'
+    '"remote_addr":"$remote_addr",'
+    '"method":"$request_method",'
+    '"uri":"$request_uri",'
+    '"status":$status,'
+    '"body_bytes_sent":$body_bytes_sent,'
+    '"request_time":$request_time,'
+    '"referer":"$http_referer",'
+    '"user_agent":"$http_user_agent"'
+    '}';
+
 server {
     listen       8080;
     server_name  localhost;
+
+    access_log /dev/stdout json_access;
 
     location / {
         root  /usr/share/nginx/html;


deploy/helm/platform/templates/ui/app.yaml
```diff
@@ -16,11 +16,29 @@ data:
...
+    log_format json_access escape=json
+        '{'
+        '"time":"$time_iso8601",'
+        '"remote_addr":"$remote_addr",'
+        '"method":"$request_method",'
+        '"uri":"$request_uri",'
+        '"status":$status,'
+        '"body_bytes_sent":$body_bytes_sent,'
+        '"request_time":$request_time,'
+        '"upstream_response_time":"$upstream_response_time",'
```
dam-code-guardian, line 29: 🟡 **Warning:** `$upstream_response_time` renders as an empty string on static-asset hits (no upstream) and as a comma-separated list (e.g. `"0.012, 0.008"`) when a request touches multiple upstreams. It stays valid JSON as a quoted string, so nothing breaks — but downstream consumers can't treat this field as a single number without splitting. Worth documenting its shape for dashboard builders.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `a457516`

## PR #1384: feat(ui): emit nginx access logs as JSON
**Author:** pilartomas | **Branch:** feat/ui-json-access-logs → main | **Changes:** +34 −0 (2 files)

### Summary
Switches the UI nginx access logs from the default plaintext `combined` format to structured JSON (one object per line on stdout), in both the live `*-ui-nginx` ConfigMap (`deploy/helm/platform/templates/ui/app.yaml`) and the image-baked `packages/ui/default.conf`. In-cluster these logs capture all same-origin `/api/` edge traffic, not just static-asset hits.

### Findings
- ✅ **Looks good:** JSON typing is correct — `$status`, `$body_bytes_sent`, and `$request_time` are emitted unquoted (numeric), while string/optional fields (`$upstream_response_time`, `$http_referer`, etc.) are quoted. `escape=json` is set, so embedded quotes/control chars in `uri`/`user_agent` won't break the JSON.
- ✅ **Looks good:** The image-baked `default.conf` deliberately drops the proxy-only fields (`upstream_response_time`, `x_forwarded_for`) since it serves no `/api/` upstream outside k8s — keeping the two formats consistent without carrying meaningless fields.
- 🟡 **Warning:** `$upstream_response_time` is emitted as a quoted scalar, but for a request that hits multiple upstreams (or none — static-asset hits never proxy) nginx renders it as an empty string or a comma-separated list (e.g. `"0.012, 0.008"`). That stays valid JSON as a string, so no parser breaks, but downstream consumers can't treat it as a number without splitting/summing. Worth documenting the field's shape for whoever builds dashboards. (Static-asset lines will have `""` here.)
- 🟢 **Suggestion:** As the author notes, there's no `$otel_trace_id` field (would crash nginx without the OTel module loaded) and trace propagation is out of scope pending operator support for this nginx version. The JSON skeleton makes adding that single field trivial later — no concern, just confirming the deferral is sound.

### Verdict
APPROVE — Correct, well-scoped JSON log-format change with proper numeric/string typing and `escape=json`; the one nuance (empty/multi-value `upstream_response_time`) is a downstream-consumer note, not a defect.

---

willallstet:
📝 Docs reviewed for this change — no documentation update required (internal observability (nginx access logs as JSON); no user-facing behavior.)

PR #1387, opened 2026-06-19
chore: unify provider definitions across ui and cli

### LINKED ISSUE #1376: Single source of truth for provider definitions across server, UI, and CLI
## Problem

A "provider" (Anthropic, OpenAI, IBM LiteLLM, Bob) is defined in several places at once, and the pieces that tie them together — which connection template a provider maps to, how a pasted Anthropic value is classified as an API key vs. an OAuth token, which env vars and host injection it needs — are re-encoded independently in each.

Concretely today: the provider registry lives in the (soon-to-be-retired) secrets module, the connection catalog re-states each provider's header/env wiring, the web UI keeps its own template→provider map plus the Anthropic token-prefix rules, and the CLI hand-rolls a parallel map of the same. The `sk-ant-oat` prefix that distinguishes Anthropic OAuth from API-key is currently hardcoded in three separate modules.

This matters because every change has to be made in lockstep across all copies. Adding a provider, changing a token prefix, or renaming a template means editing N places, and when one copy drifts the provider silently misbehaves or goes missing. A concrete symptom that exists today: **Bob is a provider on the server and a provider card in the web UI, but the CLI's hand-rolled list omits it** — so an operator can set up Bob from the UI but not from `dam agent create-interactive`. The CLI map is a deliberate, documented bridge introduced in #1357 precisely because no shared definition exists yet.

## Proposed solution

One provider registry, owned by the connections contract, that everything else derives from:

1. **Move** the provider registry out of the retiring secrets module into the connections contract. Secrets re-exports it during the transition, then drops it when secrets is removed.
2. **Enrich** each provider's mode entry with the two facts currently duplicated by hand: the connection template it maps to, and its token prefix.
3. **Derive once** — the set of provider templates, template→provider lookup, and the "classify a pasted value into a template" rule all come from that single registry.
4. **Collapse the consumers** — the server catalog sources its env bundles and injection from the registry, the UI's provider maps and Anthropic mode rules become thin derivations, and the CLI deletes its hand-rolled map and reads the shared one. Because the CLI now derives from the registry, it picks up **Bob** automatically, closing the provider-parity gap with the UI.

End state: one registry; the catalog, UI, and CLI all read from it; nothing re-encodes the template/type/prefix relation, and the three surfaces offer the same set of providers. No change to any existing provider's behavior — this removes duplication and closes a parity gap.

### Dependencies

- Companion to the secrets removal tracked in #1273 — step 1's "drop the re-export" happens as part of that cutover.
- Follows up on review feedback from #1357 (which introduced the CLI's interim map and intentionally deferred Bob).

### Out of scope

- Bob's **advanced config inputs** (model / tenant scoping) in the CLI. Reaching parity here means Bob becomes selectable as a value-only provider, matching the CLI's "paste a credential and move on" flow; prompting for the advanced fields the UI form exposes is tracked as a separate companion issue.

jjeliga:
Closes #1376


packages/cli/src/modules/agent/commands/create-interactive.ts
```diff
@@ -601,13 +599,12 @@ async function addOrReplaceProvider(
...
   while (true) {
-    const type = await select<CliProviderType>({
+    const type = await select<ProviderPresetType>({
       message: "Provider type",
-      options: [
-        { value: "anthropic", label: "Anthropic" },
-        { value: "ibm-litellm", label: "IBM LiteLLM" },
-        { value: "openai", label: "OpenAI" },
-      ],
+      options: PROVIDER_PRESET_TYPES.map((t) => ({
```
dam-code-guardian, line 604: 🟡 **Warning:** This now exposes `bob` as a selectable provider in the CLI picker (`PROVIDER_PRESET_TYPES` includes it). The old CLI `CliProviderType` deliberately excluded `bob`. The generic CLI flow only prompts for a credential and uses `bobEnvMappings()` defaults — it does not collect Bob model pins (`BOB_SHELL_MODEL`, `BOB_INSTANCE_ID`, `BOB_TEAM_ID`, `BOB_MAX_COINS`, `BOB_CHAT_MODE`) the UI Bob form collects. If intended, confirm a pin-less Bob connection is usable; otherwise filter `bob` out of the CLI options.

packages/ui/src/modules/providers/components/anthropic/modes.ts
```diff
@@ -1,38 +1,46 @@
...
+
+// `placeholder` is UI-only; everything else derives from the shared registry.
 export const MODES = {
   oauth: {
-    label: "OAuth Token",
+    label: modeFor("oauth").label,
     placeholder: "sk-ant-oat-…",
-    prefix: "sk-ant-oat-",
-    mapping: mappingFor("oauth"),
+    prefix: prefixFor("oauth"),
```
dam-code-guardian, line 26: 🟢 **Suggestion:** `prefix` here changed from the hyphen-terminated literal (`"sk-ant-oat-"`) to the registry `tokenPrefix` (`"sk-ant-oat"`). The cross-mode "wrong tab" check in `credential-schema.ts` (`value.startsWith(MODES[m].prefix)`) still works correctly — the two prefixes stay mutually exclusive after `sk-ant-` — just confirming no consumer relied on the trailing hyphen.

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `016fd47`

## PR #1387: chore: unify provider definitions across ui and cli
**Author:** jjeliga | **Branch:** chore/1376-unified-provider-definitions → main | **Changes:** +321 −386 (14 files)

### Summary
Consolidates the model-provider definitions (`PROVIDERS`, `ProviderPreset*`, `EnvMapping`, `InjectionConfig`, the IBM-LiteLLM/Bob env-mapping helpers) into a single source of truth at `packages/api-server-api/src/modules/connections/providers.ts`, and deletes the two duplicated `provider-templates.ts` mirrors in the CLI and UI packages. UI/CLI/secrets modules now all derive template→provider mappings, display names, token prefixes, and template IDs from that one registry. Closes #1376.

### Findings
- 🟡 **Warning:** The CLI interactive provider picker is now generated from `PROVIDER_PRESET_TYPES`, which includes `bob` (`packages/cli/src/modules/agent/commands/create-interactive.ts:604`). Previously the CLI deliberately scoped itself to `anthropic | ibm-litellm | openai` (the deleted `CliProviderType`), and `bob` was never offered. The generic CLI flow only prompts for a single credential and creates the connection with `bobEnvMappings()` defaults — it does **not** collect Bob's model pins (`BOB_SHELL_MODEL`, `BOB_INSTANCE_ID`, `BOB_TEAM_ID`, `BOB_MAX_COINS`, `BOB_CHAT_MODE`), which the dedicated UI Bob form collects. If exposing Bob via the CLI is intended, confirm a pin-less Bob connection is usable; if it's an unintended side-effect of deriving the option list from the shared registry, the CLI should filter `bob` out.
- 🟢 **Suggestion:** The UI Anthropic `MODES.prefix` values changed from hyphen-terminated literals (`"sk-ant-oat-"` / `"sk-ant-api-"`) to the registry `tokenPrefix` values (`"sk-ant-oat"` / `"sk-ant-api"`) (`packages/ui/src/modules/providers/components/anthropic/modes.ts:26,33`). The "wrong tab" cross-mode detection in `credential-schema.ts:27` (`value.startsWith(MODES[m].prefix)`) still works correctly with the shorter prefixes — they remain mutually exclusive after `sk-ant-` — but worth a quick check that no other consumer relied on the trailing hyphen (none found in this diff).
- ✅ **Looks good:** `TEMPLATE_TO_PROVIDER` / `PROVIDER_TEMPLATE_IDS` are now built by flat-mapping the registry modes rather than hand-maintained maps, so the template→provider relation can't silently drift from the provider list — a correctness improvement over the two deleted hardcoded maps.
- ✅ **Looks good:** `templateIdForProvider` is now generic (tokenPrefix-match → default mode → first mode) instead of an anthropic-specific `if`, and preserves the prior `sk-ant-oat → anthropic-oauth` behavior via the registry `tokenPrefix`.
- ✅ **Looks good:** Re-export surface in `index.ts` is consistent — `ProviderPreset*`/`EnvMapping`/`InjectionConfig` now sourced from `connections/providers.js`, `isValidEnvName` stays with `secrets/types.js`, no duplicate or dangling exports; `secrets/types.ts` correctly imports `EnvMapping`/`InjectionConfig` as types from the new module.

### Verdict
COMMENT — Clean, well-commented de-duplication refactor that improves correctness by deriving the template↔provider relation from a single registry; the only open question is whether newly exposing `bob` in the CLI provider picker (without model-pin collection) is intentional.

---

PetrBulanek (approved):

willallstet:
📝 Docs reviewed for this change — no documentation update required (pure refactor relocating provider definitions; the set of providers (anthropic, ibm-litellm, openai, bob) is unchanged.)

PR #1497, opened 2026-06-22
feat(cli): set connection config inputs during setup

### LINKED ISSUE #1377: CLI: set connection config inputs (Bob model & tenant scoping) during setup
## Problem

When you set up a provider in the web UI, you can fill in optional configuration alongside the credential — for Bob, that's the model, tenant scoping (instance and team), a budget cap, and a default chat mode. The CLI's interactive setup only takes the credential value, so a Bob connection created from the terminal always uses defaults. If a deployment needs tenant scoping for Bob to work, the credential alone isn't enough and the operator has to finish setup in the web UI.

*Open question to resolve while scoping this:* whether Bob actually **requires** tenant scoping to function in IBM deployments, or whether it's genuinely optional. If required, this is a true parity blocker; if not, it's a convenience.

## Proposed solution

During interactive setup, after the credential, offer the connection's optional configuration fields — using the labels and hints the template already advertises — and let the user fill in or skip each. The web UI and CLI should offer the same configuration for the same provider.

This should be **generic** — driven by what each connection template declares it accepts, not a Bob-specific prompt — so any future provider with configuration fields gets the same treatment for free.

### Dependencies

- Depends on the provider-registry unification (#1376) landing first, so the CLI reads provider and template definitions from the shared source rather than re-encoding them.

### Out of scope

- New configuration fields or validation rules — this surfaces what templates already declare; it doesn't add to them.

jjeliga:
Lets the CLI set a provider connection's optional configuration during setup, matching the web UI. For Bob that means the model, tenant scoping (instance/team), budget cap, and chat mode — previously the CLI could only set the credential, so terminal-created Bob connections always used defaults.

- `dam agent create-interactive` now prompts for the optional fields after the credential when adding a new provider.
- `dam connection connect` accepts them via a repeatable `-c, --config key=value` flag.

It's driven by what each connection template declares, so any provider with optional configuration gets the same treatment automatically.

Closes #1377.


packages/cli/src/modules/connection/domain/config-inputs.ts
```diff
@@ -0,0 +1,63 @@
...
+  input: ConnectionTemplateInput,
+  value: string,
+): string | undefined {
+  const trimmed = value.trim();
+  if (trimmed === "") return undefined;
+  const label = input.label ?? input.name;
+  if (input.enumValues && !input.enumValues.includes(trimmed)) {
+    const opts = input.enumValues.map((v) => `\`${v}\``).join(", ");
+    return `${label}: must be one of ${opts}`;
+  }
```
dam-code-guardian, line 25: 🟢 **Suggestion:** `new RegExp(input.pattern)` + `.test()` is a substring match unless the pattern is anchored. Bob's `^[1-9]\d*$` is anchored so this is correct today; it just relies on template authors anchoring any future `pattern`. The server enforcing the same spec keeps the risk low — noting for awareness.

packages/cli/src/modules/connection/commands/connect.ts
```diff
@@ -145,7 +157,18 @@ export function buildConnectCommand(deps: {
             json,
           });
 
-      const payload = buildPayload(template, name, values);
+      const configRes = resolveConfigInputFlags(template, opts.config ?? []);
```
dam-code-guardian, line 160: 🟢 **Suggestion:** `--config` is parsed for any template, but `buildPayload` only projects `configInputs` in the `header` auth branch. Today every config-input template is header-auth and `resolveConfigInputFlags` rejects unknown keys, so this is safe — but if an `oauth`/`none` template ever declared a config input, a valid `--config` value would be silently dropped here. A short comment that config inputs are header-template-only would prevent that foot-gun.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `9f8ae82`

## PR #1497: feat(cli): set connection config inputs during setup
**Author:** jjeliga | **Branch:** chore/1377-cli-add-optional-bob-setup-options → main | **Changes:** +179 −5 (7 files)

### Summary
Lets the CLI set a provider connection's optional template config inputs (model, tenant scoping, budget cap, chat mode) during setup — `dam agent create-interactive` now prompts them after the credential, and `dam connection connect` accepts them via a repeatable `-c/--config key=value` flag. The template contract gains `pattern`/`patternHint`/`enumValues` so clients can validate before submit (server still enforces). Closes #1377.

### Findings
- 🟢 **Suggestion:** `--config` is parseable for any template, but `buildPayload` only projects `configInputs` in the `header` branch (`packages/cli/src/modules/connection/commands/connect.ts:430-432`). A valid `--config key=value` on an `oauth`/`none` template would pass `resolveConfigInputFlags` (those templates declare no config inputs, so an unknown key errors — the realistic case is fine) and be silently dropped only if such a template ever declared a config input. Low impact today; worth a comment noting config inputs are header-template-only.
- 🟢 **Suggestion:** `validateConfigInputValue` builds `new RegExp(input.pattern)` and uses `.test()`, which is a substring match unless the pattern is anchored (`packages/cli/src/modules/connection/domain/config-inputs.ts:25`). Bob's `^[1-9]\d*$` is anchored so it's correct; relies on template authors anchoring future patterns. The server enforcing the same spec keeps the risk low.
- ✅ **Looks good:** `resolveConfigInputFlags` cleanly handles missing-`=`, unknown-key (with valid-key list), blank-skips (delete + continue), and last-wins; errors are typed via the `Result`/`ConfigFlagError` union and formatted in one place.
- ✅ **Looks good:** optional template fields are spread conditionally (`...(spec.pattern ? … : {})`) so absent fields never serialize, and the shared `configInputsOf`/`validateConfigInputValue` helpers are reused across the interactive flow and the flag path.
- ✅ **Looks good:** interactive `promptConfigInputs` correctly distinguishes enum (select + explicit Skip sentinel) from free-text (validated text, trimmed, blank skips) and threads cancellation through `cancelAndCleanup`.

**Verdict: minor drift**

**Check #1 — Architecture-page drift (`docs/architecture/cli.md`)** — the diff changes two CLI surfaces that `cli.md` documents in detail, but the page was not updated in this PR:

1. `dam agent create-interactive` gained an optional-config prompting step (`packages/cli/src/modules/agent/commands/create-interactive.ts`, `promptConfigInputs`), but `cli.md:93-102` enumerates the flow as exactly `name → template → model provider → optional GitHub PAT` with no mention of the new step. *Proposed:* note in step 1 (Model provider) that adding a new provider connection now also prompts the template's optional config inputs after the credential, each skippable, validated against the input's spec.
2. `dam connection connect` gained a `-c, --config key=value` repeatable flag (`packages/cli/src/modules/connection/commands/connect.ts`), but `cli.md:207-210` describes `connect`'s input collection without it. *Proposed:* document the `-c/--config` flag (last-wins, blank-skips, validated against the input spec) in the "Template id" bullet at cli.md:208.

_`connections.md:76` already documents config inputs at the subsystem level ("validated against the input's spec"); the new `pattern`/`patternHint`/`enumValues` fields surface that spec to clients while the server still enforces — covered by existing wording, so connections.md is aligned. `Last verified:` staleness does not fire — no architecture page was edited in this diff._

### Verdict
APPROVE — Small, well-structured change that reuses existing patterns and validates inputs correctly; only minor non-blocking suggestions, plus a doc-drift note that `cli.md` should be updated to mention the new `create-interactive` step and `connect --config` flag.

---

PetrBulanek (approved):

willallstet:
📝 Docs updated for this change: https://github.ibm.com/dam-agents/docs/pull/14

PR #1500, opened 2026-06-22
fix(agents): prevent model-gateway crash on client disconnect

### LINKED ISSUE #1444: Bug: Long-running agent tasks often get interrupted
Long-running agent tasks often get interrupted after a while, showing as if the agent ended its turn but the history shows that the agent was in the middle of doing something. Sending a simple message, like a single dot, usually makes the agent resume. It seems that background processes also get interrupted, hinting at the possibility of the pod fully restarting.

Note that the ACP-based UI seems to have some unrelated bugs causing it to sometimes show the agent as not running, despite it actually progressing in the background. This is thus better reproduces through the TUI mode.

JanPokorny:
May fix #1444.

## Problem

Long-running agent tasks get interrupted: the agent appears to end its turn mid-work and background processes die, consistent with the pod restarting. Captured agent logs show the `model-gateway` process crashing with an unhandled `'error'` event (`DOMException [AbortError]`), after which `pod-service` restarts it (`exited (code 1) ... restarting in 1000ms`).

## Root cause

In `packages/agents/claude-code/model-gateway.mjs`, the proxy aborts the upstream `fetch` when the client disconnects mid-response (`res` `close` with `writableEnded === false`). The response body is streamed via `Readable.fromWeb(r.body).pipe(res)`. Aborting makes that source Readable emit an `'error'` event — but `.pipe()` does **not** forward source errors and there was no `'error'` handler, so Node throws on the unhandled event and the whole gateway process dies, taking the agent's in-flight turn and background processes with it.

This reproduces under real streaming LLM workloads (e.g. BugStone) but not trivial tasks like `sleep`, which make no model calls.

## Fix

Attach an `'error'` handler to the source stream so an aborted/errored stream just tears down the response (`res.destroy()`) instead of throwing.

## Testing

Requires an agent image rebuild (`mise run cluster:build-agent`) to land in a running cluster.

### DIFF
--- a/packages/agents/claude-code/model-gateway.mjs
+++ b/packages/agents/claude-code/model-gateway.mjs
@@ -196,8 +196,11 @@ async function proxy(req, res) {
   }
 
   res.writeHead(r.status, keepHeaders([...r.headers], RES_DROP));
-  if (r.body) Readable.fromWeb(r.body).pipe(res);
-  else res.end();
+  if (r.body) {
+    const upstream = Readable.fromWeb(r.body);
+    upstream.on("error", () => res.destroy());
+    upstream.pipe(res);
+  } else res.end();
 }
 
 const server = http.createServer((req, res) => {


packages/agents/claude-code/model-gateway.mjs
```diff
@@ -196,8 +196,11 @@ async function proxy(req, res) {
   }
 
   res.writeHead(r.status, keepHeaders([...r.headers], RES_DROP));
-  if (r.body) Readable.fromWeb(r.body).pipe(res);
-  else res.end();
+  if (r.body) {
+    const upstream = Readable.fromWeb(r.body);
+    upstream.on("error", () => res.destroy());
```
dam-code-guardian, line 201: 🟢 **Suggestion:** Correct fix — `.pipe()` doesn't forward source errors, so the unhandled `'error'` on the aborted upstream was crashing the process. Optionally log the error before tearing down so real upstream failures stay diagnosable in the pod log stream (client disconnects are noise, but genuine errors would otherwise vanish silently):

```suggestion
    upstream.on("error", (err) => { console.error("model-gateway upstream stream error:", err); res.destroy(); });
```

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `9d3b702`

## PR #1500: fix(agents): prevent model-gateway crash on client disconnect
**Author:** JanPokorny | **Branch:** fix/model-gateway-stream-crash#locki-mnocn54w → main | **Changes:** +5 −2 (1 file)

### Summary
Attaches an `'error'` handler to the upstream `Readable` (from `Readable.fromWeb(r.body)`) before piping it to the response, so an aborted/errored upstream stream tears down the response via `res.destroy()` instead of throwing an unhandled `'error'` event and crashing the whole model-gateway process (which was taking down the agent's in-flight turn and background work).

### Findings
- ✅ **Looks good:** Correct root-cause fix. `stream.pipe()` deliberately does not forward source errors, and an aborted web stream emits `'error'` on the Node `Readable` — without a listener Node rethrows it as an uncaught exception. Adding the handler is the idiomatic remedy.
- ✅ **Looks good:** `res.destroy()` is the right teardown here — the client has already disconnected (or the upstream failed), so there is no partial-response correctness concern.
- 🟢 **Suggestion:** Consider logging the error before destroying, so genuine upstream failures (vs. ordinary client disconnects) remain diagnosable in the pod log stream. (`packages/agents/claude-code/model-gateway.mjs:201`)

### Verdict
APPROVE — Minimal, correct fix for a real process-crash bug; the only note is an optional observability suggestion.

---

willallstet:
📝 Docs reviewed for this change — no documentation update required (internal bug fix (model-gateway no longer crashes on client disconnect); no documented contract changed.)

PR #1590, opened 2026-06-22
fix: prevent liveness probe from killing busy agent pods

JanPokorny:
## Problem

Agent pods were being restarted mid-work even though the agent had work to do and no tab was open. Investigation of a live crash (`restartCount` climbing, `lastState.terminated exitCode: 9`) traced it to the **liveness probe**, not hibernation:

```
Warning  Unhealthy  Liveness probe failed: Get ".../healthz": context deadline exceeded
Normal   Killing    Container agent failed liveness probe, will be restarted
```

`/healthz` is served on agent-runtime's **single Node event loop**, shared with all ACP frame handling. The probes set no `TimeoutSeconds`, so they inherited the Kubernetes default of **1 second**. When a busy turn stalled the event loop for ≥1s across 3 probe cycles (~30s), the kubelet SIGKILLed the container (exit 9) and restarted it — losing the in-flight session. A 1s timeout punishes a *busy* process, not just a hung one.

## Changes

**controller — stop false kills (root cause)**
- Liveness probe now sets `TimeoutSeconds: 5`, `FailureThreshold: 3` (`resources.go`, `fork_resources.go`). A truly hung process still dies within ~30s; a transient sub-5s stall no longer reads as hung. Readiness stays at the 1s default — a brief de-route self-heals, where a false liveness kill drops work.

**agent-runtime — reduce the stalls that trigger it**
- Parse each agent→client frame **once**: `rewriteAuthError` now short-circuits on a cheap substring check instead of a second full `JSON.parse` per frame.
- Replace **O(n²)** per-chunk stdout buffering (`buf += chunk; buf.split("\n")` re-scanned the whole growing buffer every chunk) with `readline` — linear, incremental.

**agent-runtime — graceful shutdown & observability**
- SIGTERM/SIGINT handler: stop accepting connections, close sessions cleanly (WS 1000 → clients resume from the harness's on-disk store), hard-exit fallback. Ceiling documented in code: a wedged loop can't run the handler, so the kubelet still escalates to SIGKILL.
- Event-loop-delay watchdog logs `[eventloop] blocked up to <N>ms` for any stall ≥1s (stdlib `monitorEventLoopDelay`, sampled at libuv level so it survives the block it reports).
- Per-frame timing logs `slow frame <ms> (<bytes>, <method>)` to attribute future stalls to a specific frame.

## Testing
- New `create-child-agent-process.test.ts`: a real child process whose frame is split across stdout writes is reassembled into one line (and blank lines skipped).
- `tsc`, eslint, prettier, gofmt clean; Go build/vet/staticcheck pass; **104/104 agent-runtime tests pass**.

## Deploy note
Takes effect only after an image rebuild (controller + agent-runtime).

### DIFF
--- a/packages/agent-runtime/src/server.ts
+++ b/packages/agent-runtime/src/server.ts
@@ -1,4 +1,5 @@
 import http from "node:http";
+import { monitorEventLoopDelay } from "node:perf_hooks";
 import { existsSync } from "node:fs";
 import { fileURLToPath } from "node:url";
 import { dirname, join } from "node:path";
@@ -424,3 +425,28 @@ server.listen(config.PORT, () => {
       process.env.PLATFORM_AGENT_VERSION ?? "agent-runtime/unknown",
   });
 });
+
+const eld = monitorEventLoopDelay({ resolution: 20 });
+eld.enable();
+setInterval(() => {
+  const maxMs = eld.max / 1e6;
+  if (maxMs >= 1_000)
+    process.stderr.write(
+      `[eventloop] blocked up to ${Math.round(maxMs)}ms ` +
+        `(p99 ${Math.round(eld.percentile(99) / 1e6)}ms) in last 10s\n`,
+    );
+  eld.reset();
+}, 10_000).unref();
+
+let shuttingDown = false;
+function gracefulShutdown(signal: string): void {
+  if (shuttingDown) return;
+  shuttingDown = true;
+  process.stderr.write(`[shutdown] ${signal} received, closing\n`);
+  server.close();
+  for (const sid of [...ptySlots.keys()]) killPtySlot(sid);
+  acpRuntime.shutdown();
+  setTimeout(() => process.exit(0), 3_000).unref();
+}
+process.on("SIGTERM", () => gracefulShutdown("SIGTERM"));
+process.on("SIGINT", () => gracefulShutdown("SIGINT"));
--- a/packages/agent-runtime/src/__tests__/unit/create-child-agent-process.test.ts
+++ b/packages/agent-runtime/src/__tests__/unit/create-child-agent-process.test.ts
@@ -1,0 +1,33 @@
+import { describe, it, expect } from "vitest";
+import { createChildAgentProcess } from "../../modules/acp/infrastructure/create-child-agent-process.js";
+
+describe("createChildAgentProcess", () => {
+  it("reassembles a frame split across stdout chunks and skips blank lines", async () => {
+    // First write has no trailing newline, so the frame is only complete
+    // after the second write — the incremental-buffering case the readline
+    // swap must handle (and the old per-chunk re-split did, O(n²)).
+    const script = [
+      `process.stdout.write('{"id":1,"a":');`,
+      `setTimeout(() => {`,
+      `  process.stdout.write('"hello"}\\n   \\n{"id":2}\\n');`,
+      `  process.exit(0);`,
+      `}, 20);`,
+    ].join("\n");
+
+    const proc = createChildAgentProcess({
+      command: [process.execPath, "-e", script],
+      workingDir: process.cwd(),
+    });
+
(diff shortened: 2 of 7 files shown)


packages/agent-runtime/src/server.ts
```diff
@@ -424,3 +425,28 @@ server.listen(config.PORT, () => {
       process.env.PLATFORM_AGENT_VERSION ?? "agent-runtime/unknown",
   });
 });
+
+const eld = monitorEventLoopDelay({ resolution: 20 });
```
dam-code-guardian, line 429: 🟢 **Suggestion:** The event-loop-delay watchdog (and the `slow frame` log in `acp-runtime.ts`) write to stderr / `deps.log` unconditionally. On a chronically busy pod that crosses the 1s / 250ms thresholds repeatedly, this could get noisy in the pod log stream. Consider gating behind a debug flag or rate-limiting if volume becomes a concern — not blocking.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `ad5de06`

## PR #1590: fix: prevent liveness probe from killing busy agent pods
**Author:** JanPokorny | **Branch:** fix/agent-liveness-event-loop-stalls → main | **Changes:** +87 −11 (7 files)

### Summary
Fixes false liveness-probe kills of busy agent pods. The root cause — a 1s default probe timeout punishing a momentarily-stalled single-threaded event loop — is addressed by setting `TimeoutSeconds: 5`/`FailureThreshold: 3` on the liveness probe in the controller. Secondary changes reduce event-loop stalls (one-time frame parse, O(n²)→linear `readline` buffering) and add graceful SIGTERM shutdown plus event-loop-delay/slow-frame observability in agent-runtime.

### Findings
- ✅ **Looks good:** Probe tuning is correct and minimal — `TimeoutSeconds: 5` + `FailureThreshold: 3` still kills a genuinely hung process within ~30s while tolerating sub-5s transient stalls; readiness left at the 1s default is the right call (a brief de-route self-heals; a false liveness kill loses work). Applied symmetrically to both `resources.go` and `fork_resources.go`. (`packages/controller/pkg/reconciler/resources.go:320`, `fork_resources.go:251`)
- ✅ **Looks good:** `rewriteAuthError` short-circuit is behavior-preserving — both rewrite branches already require the `authentication_error` substring, so the cheap `includes` guard can never skip a frame that would have been rewritten. (`packages/agent-runtime/src/modules/acp/infrastructure/mappers.ts:10`)
- ✅ **Looks good:** `readline` swap eliminates the O(n²) re-split of a growing buffer and is covered by a real child-process test exercising the split-frame + blank-line cases. (`packages/agent-runtime/src/modules/acp/infrastructure/create-child-agent-process.ts:35`)
- ✅ **Looks good:** Graceful shutdown is idempotent (`shuttingDown` guard), closes sessions with WS 1000 so clients resume from the on-disk store, and keeps a `setTimeout(...).unref()` hard-exit fallback. `acpRuntime.shutdown()`, `killPtySlot`, and `ptySlots` are all correctly defined and wired. (`packages/agent-runtime/src/server.ts:438`)
- 🟢 **Suggestion:** The event-loop watchdog and slow-frame logs write to `stderr` / `deps.log` unconditionally; on a chronically busy pod these could be noisy. Consider gating behind a debug flag or rate-limiting if log volume becomes a concern — not blocking. (`packages/agent-runtime/src/server.ts:429`)

### Verdict
APPROVE — Correct root-cause fix (probe timeout) plus sound, well-tested stall-reduction and graceful-shutdown work; no doc drift. (The TypeScript code-review skill was unavailable this session, so that section is omitted; DAM's own correctness pass over the 5 TS files surfaced no blocking issues.)

---

willallstet:
📝 Docs reviewed for this change — no documentation update required (internal infra reliability (liveness probe no longer kills busy pods); no documented behavior changed.)

PR #1688, opened 2026-06-23
fix(ui): fix sidebar session deletion

### LINKED ISSUE #1084: Deleted session lingers at the top of the sidebar until manual refresh
## Problem

When you delete a session from the left pane, a "Session deleted" confirmation appears — but the session doesn't go away. Instead it jumps to the **top** of the session list and stays there. The only way to make it disappear is to refresh the page (or hit the refresh control on the sidebar).

This is confusing and undermines trust in the action: the toast says the delete succeeded, yet the session is not only still visible but now more prominent than before. A user can reasonably conclude the delete failed, try again, or wonder whether the session was actually removed. The mismatch between the success message and what's on screen is the core problem.

## Proposed solution

Deleting a session should remove it from the sidebar immediately, the moment the confirmation appears — no refresh required. The list should reflect the deletion right away rather than waiting for the next time the page is reloaded.

In particular, deleting the session you currently have open should also remove it from the list cleanly, rather than leaving it pinned at the top.

### Out of scope
- Undo / "restore deleted session" — separate feature.
- Any change to how sessions are sorted in normal (non-deletion) use.

jjeliga:
Closes #1084

### DIFF
--- a/packages/e2e/playwright/playwright.config.ts
+++ b/packages/e2e/playwright/playwright.config.ts
@@ -64,5 +64,13 @@ export default defineConfig({
       dependencies: ["injection"],
       use: { ...devices["Desktop Chrome"], storageState },
     },
+    {
+      // Creates and deletes its own session, so it leaves no residue for other
+      // specs; depends on "agent" only to gate on a provisioned running agent.
+      name: "session-delete",
+      testMatch: /08-.*\.spec\.ts$/,
+      dependencies: ["agent"],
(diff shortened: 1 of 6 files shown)


kapetr (approved):

willallstet:
📝 Docs reviewed for this change — no documentation update required (UI bug fix restoring intended sidebar session-deletion behavior; no documented contract changed.)

PR #1689, opened 2026-06-23
feat(agent-runtime): log memory + CPU context on event-loop blocks

JanPokorny:
## Why

On the new-code dev pod, the watchdog caught the kill cause — `[eventloop] blocked up to 3563ms` → `5587ms` → terminal, seconds before a liveness SIGKILL. But it couldn't tell us **what** blocked: the duration alone can't distinguish GC thrash, container memory pressure, a CPU-bound sync op, or a blocking syscall. And critically, `slow frame` never fired, so it's **not** ACP frame processing.

Rather than keep guessing (we already burned one wrong hypothesis), make the next block self-diagnosing.

## What

Each `[eventloop] blocked` line now carries memory + CPU context:

```
[eventloop] blocked up to 5587ms (p99 5587ms) in last 10s — rss=512MB heap=180/210MB cpu=12% cgroup=1980/2048MB
```

Reading the signals:
| cause | signal |
|---|---|
| V8 GC thrash | heap ≈ heapTotal, high cpu% |
| container memory pressure | cgroup near limit, **low** cpu% (the harness shares the cgroup — caught even when our own heap is fine) |
| CPU-bound sync op | high cpu%, heap + cgroup normal |
| blocking syscall (e.g. slow PVC) | low cpu%, heap + cgroup normal |

Plus a gated `[mem] high cgroup usage` heartbeat (≥85% of limit) to capture the **run-up** to a terminal collapse — a terminal wedge leaves no time to log on its own.

## Safety
- cgroup reads are kernel-served (cgroupfs, no disk I/O) so the sync read can't itself stall the loop; v2 path with v1 fallback, each guarded.
- The entire watchdog body is wrapped in try/catch — observability must never take down the runtime.
- Purely additive logging; no behavior change.

## Testing
tsc / eslint / prettier clean; 105/105 agent-runtime tests pass.

Merge + redeploy, then the next event-loop block on a real workload will name its own cause.

### DIFF
--- a/packages/agent-runtime/src/server.ts
+++ b/packages/agent-runtime/src/server.ts
@@ -1,6 +1,6 @@
 import http from "node:http";
 import { monitorEventLoopDelay } from "node:perf_hooks";
-import { existsSync } from "node:fs";
+import { existsSync, readFileSync } from "node:fs";
 import { fileURLToPath } from "node:url";
 import { dirname, join } from "node:path";
 import headlessPkg from "@xterm/headless";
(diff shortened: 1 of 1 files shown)


willallstet:
📝 Docs reviewed for this change — no documentation update required (internal diagnostics — logs memory/CPU on event-loop blocks; no user-facing behavior.)

PR #1739, opened 2026-06-23
feat(ui): render HTML files in the file viewer via a sandboxed iframe

### LINKED ISSUE #1415: Render HTML visualizations in-app instead of only offering a download
## Problem

The **Nous** agent's visualization commands (`/visualize-campaign` and `/visualize-registry`) turn a finished campaign — or all campaigns together — into an **interactive knowledge graph**: a self-contained page you can pan, zoom, and click through to explore entities, concepts, dead-ends, and cross-campaign "opportunity" scores. Seeing this graph is one of the main payoffs of running Nous — knowledge that compounds across campaigns, presented visually so you can decide what to investigate next.

Today the user can't actually *see* that graph in the app. When the agent opens one of these pages, the file panel shows its **raw source**, not the rendered visualization. The only way to view it is to **download the file and open it in a separate browser tab**. (A chat message can't show it either — HTML in chat is rendered as plain text.)

So the interactive artifact the agent just produced is locked behind a manual download-and-open step, which defeats the "glance at the graph, pick the next experiment" workflow it exists to enable.

## Proposed solution

Let the app **render** an HTML page when the user opens it — the way it already renders PDFs and images — so a Nous visualization opens straight into the live, interactive graph. A **Rendered / Source toggle** (like the one other previewable file types have) lets users flip to the raw HTML when they want it. The agent can then link the user directly to a campaign's graph and they see it in one click.

### Scope
- Applies to any agent-produced HTML opened in the file panel — Nous is the motivating case, not the only beneficiary.

### Transparency / safety
- These pages are produced by an agent running experimental workloads, so they must be rendered in an **isolated** way that can't reach the user's session or the rest of the app.
- Rendering relies on the page being **self-contained** (no externally-loaded assets), to work within the app's content-security restrictions. The Nous graphs are generated as standalone pages, which fits.

### Dependencies
- The Nous agent and its `/visualize-*` commands are added in #1382; this issue tracks rendering their output in-app.

### Out of scope
- Rendering interactive HTML *inline inside a chat message* (chat content is sanitized; the file panel is the right surface).
- The download-and-open path works today as an interim — this issue is about removing that friction.

Tomas2D:
## Summary

Renders HTML files in the file panel instead of only showing their source, so agent-produced interactive pages — notably the **Nous** campaign knowledge-graph visualizations (`/visualize-campaign`, `/visualize-registry`) — can be viewed in-app without downloading.

Opening a `text/html` file now shows it in a **sandboxed iframe** with a **Raw / Render toggle** (matching the existing SVG and markdown toggles); it defaults to rendered, and flips to source on demand. The shared Raw/Render toolbar button was extracted into a small `RenderToggle` component now used by all three previews.

## Safety

The iframe uses `sandbox="allow-scripts"` **without** `allow-same-origin`, so the page's JS runs in an opaque origin and cannot reach the app's session, cookies, or DOM. The repo enforces no Content-Security-Policy, so self-contained pages (the Nous graphs are standalone) render with their inline scripts intact.

## Testing

- `mise run ui:check` clean (eslint + tsc + prettier).
- Confirmed `.html` is served as `text/html`/non-binary with the source in `content`, which the iframe consumes.
- **Not yet visually verified in a running UI** — reviewer/author should open a generated visualization and confirm the render. (Interactive D3 in the sandboxed frame.)

## Notes

- `file-viewer.tsx` is over the team's ~300-line component-split guideline; this PR stays focused (adds the branch + extracts the one shared toggle). A fuller split of the viewer is left as a separate refactor.

Closes #1415

### DIFF
--- a/packages/ui/src/modules/files/components/file-viewer.tsx
+++ b/packages/ui/src/modules/files/components/file-viewer.tsx
@@ -1,11 +1,9 @@
 import {
   ArrowLeft,
   Close as X,
-  Code,
   Download,
   Edit as Pencil,
   Save,
-  View as Eye,
 } from "@carbon/icons-react";
 import { useCallback, useEffect, useMemo, useState } from "react";
 
@@ -23,6 +21,7 @@ import {
 } from "../api/queries.js";
 import { base64ToBlob, downloadFileContent } from "../lib/download.js";
 import { CodeEditor } from "./code-editor.js";
+import { RenderToggle } from "./render-toggle.js";
 
 interface Props {
   file: FileContent;
@@ -62,6 +61,7 @@ export function FileViewer({ file, onClose, onOpenFile }: Props) {
   const { path, content, binary, mimeType: mime, tooLarge } = file;
   const isMarkdown = mime === "text/markdown";
   const isSvg = mime === "image/svg+xml";
+  const isHtml = mime === "text/html";
   const isBinaryImage = binary && content && isImageMime(mime) && !isSvg;
   const isPdf = mime === "application/pdf";
   const filename = path.split("/").pop();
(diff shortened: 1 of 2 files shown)


jezekra1 (approved):

Tomas2D:
Superseded by #1746, which is a single PR containing both the HTML render and the fullscreen preview.

PR #1742, opened 2026-06-23
feat(ui): never hibernate option

### LINKED ISSUE #1373: Keep an agent awake while it has long-running background work (don't hibernate mid-job)
## Problem

Agents hibernate when their session goes quiet — no one is chatting, no terminal is open. That's usually the right behavior: it frees cluster resources, and the agent wakes again on the next message.

But an agent can be doing real work with no one watching. The clearest case is a **Nous** campaign (#939): the agent kicks off an experiment that runs for minutes to hours in the background, then the user steps away. With no active session, the platform treats the agent as idle and hibernates it — interrupting the in-progress run. The work isn't lost (its state is checkpointed and can be resumed), but it does not *progress* while the agent is hibernated.

Today there are two ways to keep a long job moving unattended:
- set a custom time until hibernation (incl. a never hibernate option) 
-  leave a terminal or SSH session open so the pod stays pinned awake. 

However, these are both crude workarounds for the actual core issue - we are currently unable to reliably determine that an agent is "working" without intercepting ACP communication. This manual babysitting step defeats the point of a background workload. "Start an experiment, walk away, come back to findings" is the headline Nous use case, and it's effectively blocked.

## Ideal solution

- The agent has a running status **exactly while it is working**, with no human holding a session open.
- When the background work completes or stops, the agent becomes idle again and hibernates on the normal schedule. 
- The signal should be **generic** — any agent image or workload can use it. 

## "Solution" ideas

These are not proper solutions since none of them does truly solve the reliability issue 

- MCP to reserve keepawake time: agent can declare itself to be running for the next N minutes when spinning up a background task; The advantage of this approach is that an agent could possibly adjust (prolong) this on the go, disadvantage is obviously agent non-determinism and the fact that it cannot know how long will some workloads take upfront
- MCP to declare running state: agent pins itself as running and lifts the pin when it is done - this might be highly unreliable and opens more room for error than the previous idea
- Tracking actual processes: this seems to be the most reliable approach at a first glance, but we are currently unable to distinguish which process is actual background work and which is just a stale orphan left behind by the agent, making this solution non-viable until this is solved

### Scope

- Covers keeping a pod awake while it reports work in progress, and resuming normal hibernation once it reports idle.
- Doesn't change the chat/terminal experience or the existing "open a session to keep it awake" behavior — it just removes the need to do that for genuine background jobs.

### Dependencies

- Motivated by the Nous integration (#939) — a Nous campaign is the first concrete consumer.
- Related: #938 (custom images for target workloads) — other long-running workloads benefit from the same signal.
- Related but distinct: #43 (proactive heartbeat / self-scheduling). That solves *periodic wake-and-check*; this solves *stay awake continuously while a single job runs*. Complementary, not substitutes.

### Out of scope

- A global "never hibernate" switch — it would defeat scale-to-zero for every agent. The intent is to stay awake *only* while there's actual work.
- Periodic schedule-driven progress (wake, do a little, sleep) — that's the heartbeat approach tracked in #43.

jjeliga:
Implements option to fully disable hibernation for custom image agents

Closes #1373

### DIFF
--- a/docs/adrs/072-keep-awake-hard-blocker.md
+++ b/docs/adrs/072-keep-awake-hard-blocker.md
@@ -1,0 +1,27 @@
+# ADR-072: Keep agents awake with an explicit hard blocker, not work-detection
+
+**Date:** 2026-06-22
+**Status:** Accepted
+**Owner:** @jjeliga
+
+## Context
+
+Agents auto-hibernate (scale to zero) when the runtime reports itself idle, and idleness is driven by ACP chat/terminal session activity. A long-running workload that runs *without* an ACP session generates no activity, so the idle checker hibernates it mid-job and kills the work. Session-bound background work is out of scope — closing the session may end the work it spawned. The open problem is the no-session workload; any solution had to be harness-agnostic and reliable across hours-long jobs.
+
+## Decision
+
+Keeping an agent awake is an explicit, per-agent opt-in — never inferred from the workload's processes. It takes two forms: an **unconditional flag** (off by default) that makes the agent count as never-idle so the controller never scales it to zero, for long-running no-session workloads that can't signal their own boundaries; and a **self-managed sentinel** a workload creates while its work runs and removes when done, pinning the agent only for that duration so it reclaims itself. The platform observes nothing about the workload and promises only that the agent will not hibernate while a pin is set. The flag is the per-agent form of the existing global auto-hibernation switch, and is never set on interactive harness agents.
+
+## Alternatives Considered
+
+- **Lease/heartbeat** — the keep-alive connection would have to survive hours, and a transient network blip would drop the lease and kill the job; the agent also has no general signal for "still working" to renew on.
+- **Per-process PID-liveness declaration** — requires harness cooperation and is blind to work the harness spawns ad hoc.
+- **Platform run-in-background primitive** — every harness would have to route background work through it, and it cannot capture ad-hoc spawns, so coverage is partial and unreliable.
+- **Generic child-process monitoring** — cannot distinguish always-on idle infrastructure (e.g. a model gateway) from active work, so it pins pods indefinitely on false positives, and orphaned processes silently defeat scale-to-zero.
+- **Single supervised-hook observability** — correct only for workloads that adopt that hook *and* keep its process alive for the whole job; silently hibernates every other workload mid-job — a failure mode that looks robust but isn't.
+
+## Consequences
+
+- **Easier:** A no-session workload survives to completion regardless of how it spawns processes — process- and harness-agnostic, it covers the ad-hoc-spawn and orphaned-child cases every detection-based alternative provably misses. A workload that knows its boundaries self-manages the sentinel and reclaims automatically. Reuses the existing global idle-disable model, now scoped to one agent.
+- **Harder:** The unconditional flag never scales back to zero on its own — the agent holds its resources until the flag is cleared by hand, even after the job finishes. The sentinel reclaims only if the workload removes its marker; a crash that skips cleanup leaves the agent pinned. Left enabled on an interactive agent, keep-awake disables its hibernation for no benefit.
+- **Committed-to:** Reclaiming a flag-pinned agent is operational, not automatic. If forgotten or runaway pins become a resource problem, a bounding mechanism (max-lifetime cap, or staleness expiry on the sentinel) is the follow-up — none today.
--- a/docs/adrs/072-keep-awake-hard-blocker.md
+++ b/docs/adrs/072-keep-awake-hard-blocker.md
@@ -1,6 +1,6 @@
 # ADR-072: Keep agents awake with an explicit hard blocker, not work-detection
 
-**Date:** 2026-06-22
+**Date:** 2026-06-24
 **Status:** Accepted
 **Owner:** @jjeliga
 
@@ -10,7 +10,7 @@ Agents auto-hibernate (scale to zero) when the runtime reports itself idle, and 
 
 ## Decision
 
-Keeping an agent awake is an explicit, per-agent opt-in — never inferred from the workload's processes. It takes two forms: an **unconditional flag** (off by default) that makes the agent count as never-idle so the controller never scales it to zero, for long-running no-session workloads that can't signal their own boundaries; and a **self-managed sentinel** a workload creates while its work runs and removes when done, pinning the agent only for that duration so it reclaims itself. The platform observes nothing about the workload and promises only that the agent will not hibernate while a pin is set. The flag is the per-agent form of the existing global auto-hibernation switch, and is never set on interactive harness agents.
+Keeping an agent awake is an explicit, per-agent opt-in — never inferred from the workload's processes. It takes two forms: an **unconditional flag** (off by default) that makes the agent count as never-idle so the controller never scales it to zero, for long-running no-session workloads that can't signal their own boundaries; and a **self-managed sentinel** a workload creates while its work runs and removes when done, pinning the agent only for that duration so it reclaims itself. The platform observes nothing about the workload and promises only that the agent will not hibernate while a pin is set. The flag is the per-agent form of the existing global auto-hibernation switch.
 
 ## Alternatives Considered
 
--- a/packages/agent-runtime/src/modules/config.ts
+++ b/packages/agent-runtime/src/modules/config.ts
@@ -1,14 +1,18 @@
 import { z } from "zod/v4";
 
+// Forgiving boolean: these flags can be hand-set in an image Dockerfile, so
+// accept the common truthy spellings rather than only the exact "true".
+const boolEnv = (v: string): boolean =>
+  ["1", "true", "yes", "on"].includes(v.trim().toLowerCase());
+
 const schema = z.object({
   PORT: z.coerce.number().default(8080),
-  PLATFORM_DEV: z
-    .string()
-    .default("false")
-    .transform((v) => v === "true"),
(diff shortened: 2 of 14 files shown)


packages/agent-runtime/src/modules/config.ts
```diff
@@ -9,6 +9,10 @@ const schema = z.object({
   HOME_DIR: z.string().default("/home/agent"),
   WORK_DIR: z.string().default("/home/agent/work"),
   API_SERVER_URL: z.string().default(""),
+  PLATFORM_KEEP_AWAKE: z
+    .string()
+    .default("false")
+    .transform((v) => v === "true"),
```
dam-code-guardian: 🟡 **Warning:** This parses only the exact string `"true"` — but the comment in `keep-awake.ts` ("Set it truthy"), ADR-071, and `agent-lifecycle.md` all advertise setting this "via image `ENV`". A custom image baking `ENV PLATFORM_KEEP_AWAKE=1` (or `yes`/`TRUE`) — the documented no-session-workload path — would parse to `false` and still hibernate mid-job, silently defeating the feature. The UI path is safe (always writes `"true"`); the image-ENV path is the exposure. Either accept common truthy spellings or tighten the docs/comment to require the literal `true`.

```suggestion
    .transform((v) => ["true", "1", "yes"].includes(v.trim().toLowerCase())),
```

docs/adrs/071-keep-awake-hard-blocker.md
```diff
@@ -0,0 +1,27 @@
+# ADR-071: Keep agents awake with an explicit hard blocker, not work-detection
+
+**Date:** 2026-06-22
+**Status:** Accepted
+**Owner:** @github-username
```
dam-code-guardian: 🟢 **Suggestion:** Placeholder owner — worth filling in before merge.

```suggestion
**Owner:** @jjeliga
```

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `fa1ccc2`

## PR #1742: feat(ui): never hibernate option
**Author:** jjeliga | **Branch:** feat/1373-keep-working-agent-awake → main | **Changes:** +172 −5 (11 files)

### Summary
Adds an opt-in `PLATFORM_KEEP_AWAKE` per-agent hard hibernation blocker (closes #1373): agent-runtime parses the env var and `/api/status` now reports never-idle when it's set; a shared `KeepAwakeToggle` exposes it (custom-image agents only) in both the create wizard and the settings form behind a confirmation dialog; ADR-071 and the `agent-lifecycle.md` Hibernate section document the decision. The design (hard blocker, not work-detection) is well-reasoned and the wizard/settings code share a single env-name constant and confirm helper so the two surfaces can't drift.

### Findings
- 🟡 **Warning:** The runtime parser accepts **only** the exact string `"true"` (`config.ts:15` — `.transform((v) => v === "true")`), but the code comment (`keep-awake.ts:4` "Set it truthy"), the ADR, and `agent-lifecycle.md` all advertise setting it "via image `ENV`". A human baking `ENV PLATFORM_KEEP_AWAKE=1` (or `yes`/`TRUE`) into a custom image — exactly the documented no-session-workload path — would silently parse to `false` and the agent would still hibernate mid-job, defeating the feature with no error. The UI path is safe (it always writes `"true"`), but the image-ENV path is the one the docs emphasize. Consider accepting common truthy spellings (e.g. `["true","1","yes"].includes(v.toLowerCase())`) or tightening the docs/comment to say the value must be exactly `true`. (`packages/agent-runtime/src/modules/config.ts:15`)
- 🟢 **Suggestion:** No test covers the new idle-gating. The `/api/status` idle expression is the load-bearing line of the feature (`server.ts:82-86`), yet there's no agent-runtime test asserting `idle === false` when `PLATFORM_KEEP_AWAKE` is set (and unchanged otherwise). A small unit test would lock in the behavior and guard against a future refactor of the `idle` expression silently dropping the term. (`packages/agent-runtime/src/server.ts:82-86`)
- 🟢 **Suggestion:** ADR-071 ships with a placeholder owner `**Owner:** @github-username`. Worth filling in before merge. (`docs/adrs/071-keep-awake-hard-blocker.md:5`)
- ✅ **Looks good:** Backpressure against drift is well-handled — `KEEP_AWAKE_ENV` and `confirmEnableKeepAwake` live in one `keep-awake.ts` module shared by the wizard and the settings form, and the settings toggle derives its state from the env list rather than holding a parallel copy, so the env var is the single source of truth.
- ✅ **Looks good:** The confirmation flow is correct on both surfaces — enabling prompts (it disables hibernation indefinitely), disabling never does, and on cancel the controlled `Switch` simply doesn't change because state is only updated after the confirm resolves. The wizard's `confirmKeepAwake` is correctly memoized via `useCallback`.
- ✅ **Looks good:** `wizardSnapshot` adds `keepAwake` with `z.boolean().default(false)`, so snapshots persisted by an earlier build still parse — backward-compatible. Picking a template resets `keepAwake` to false, matching the custom-image-only scoping (and the settings view gates the section on `!agent.templateId`).

✅ No documentation drift detected. The changed subsystem (agent hibernation / idle reporting) is `docs/architecture/agent-lifecycle.md`, updated in this same PR — the Hibernate section now describes the `PLATFORM_KEEP_AWAKE` opt-out and `Last verified:` is bumped to 2026-06-22. No ADR link or volatile content leaked into the architecture page; the per-agent flag is not a new long-lived subsystem requiring its own page.

### Verdict
COMMENT — Clean, well-factored feature with solid docs. One worth-addressing inconsistency: the strict `=== "true"` parse silently no-ops the documented image-`ENV` path for any value other than the literal `true`; plus a missing unit test on the idle gating and an ADR owner placeholder. None blocking.

---

docs/adrs/072-keep-awake-hard-blocker.md
```diff
@@ -0,0 +1,27 @@
+# ADR-071: Keep agents awake with an explicit hard blocker, not work-detection
```
dam-code-guardian: 🟡 **Warning:** The file was renamed to `072-…` to avoid colliding with the existing `071-postgres-role-separation.md`, but the heading still reads `ADR-071` — so two ADRs now claim number 071. Bump the heading to match the filename.

```suggestion
# ADR-072: Keep agents awake with an explicit hard blocker, not work-detection
```

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `94b1369`

## PR #1742: feat(ui): never hibernate option
**Author:** jjeliga | **Branch:** feat/1373-keep-working-agent-awake → main | **Changes:** +176 −9 (11 files)

### Summary
Adds an opt-in `PLATFORM_KEEP_AWAKE` per-agent hard hibernation blocker (closes #1373): agent-runtime parses the env var and `/api/status` reports never-idle when it's set; a shared `KeepAwakeToggle` exposes it (custom-image agents only) in both the create wizard and the settings form behind a confirmation dialog; ADR and the `agent-lifecycle.md` Hibernate section document the decision. This commit addresses both substantive findings from the prior review.

### Changes since last review
Previous HEAD: `fa1ccc2` (2026-06-23T11:13:13Z) — verdict COMMENT

- ✅ **Fixed:** Strict `=== "true"` env parse (`config.ts`) — replaced with a shared `boolEnv` helper accepting `1`/`true`/`yes`/`on` (case-insensitive, trimmed), applied to both `PLATFORM_KEEP_AWAKE` and `PLATFORM_DEV`. The documented image-`ENV` path now works for common truthy spellings.
- ✅ **Fixed:** ADR owner placeholder `@github-username` → `@jjeliga`.
- 🔁 **Still present:** No unit test covers the new idle-gating (`server.ts` `/api/status` idle expression) — carried over from previous review.
- 🆕 **New:** The ADR file was renamed `071-…` → `docs/adrs/072-keep-awake-hard-blocker.md` (to avoid colliding with the existing `071-postgres-role-separation.md`), but its heading still reads `# ADR-071:` — two ADRs now both claim number 071.

### Findings
- 🟡 **Warning:** ADR filename/heading number mismatch — `docs/adrs/072-keep-awake-hard-blocker.md` opens with `# ADR-071:`, colliding with the existing `docs/adrs/071-postgres-role-separation.md`. The filename was correctly bumped to 072 to avoid the collision; the heading wasn't. Update the heading to `# ADR-072:`. (`docs/adrs/072-keep-awake-hard-blocker.md:1`)
- 🟢 **Suggestion:** Still no test on the load-bearing idle gating. A small agent-runtime unit test asserting `idle === false` when `PLATFORM_KEEP_AWAKE` is set (and unchanged otherwise) would lock in the behavior against a future refactor of the `idle` expression. (`packages/agent-runtime/src/server.ts`)

✅ No documentation drift detected. The changed subsystem (agent hibernation / idle reporting) is `docs/architecture/agent-lifecycle.md`, updated in this same PR — the Hibernate section now describes the `PLATFORM_KEEP_AWAKE` opt-out and `Last verified:` is bumped to 2026-06-22. No ADR link or volatile content leaked into the architecture page; the per-agent flag is not a new long-lived subsystem requiring its own page. (The `docs/adrs/` heading mismatch is out of doc-drift's scope — surfaced as a Finding above.)

### Verdict
COMMENT — Both prior findings are cleanly resolved (the `boolEnv` fix is the right shape and correctly generalized to `PLATFORM_DEV`). One new non-blocking doc bug: the ADR heading still says 071 after the file was renamed to 072. None blocking.

---

willallstet:
📝 Docs updated for this change: https://github.ibm.com/dam-agents/docs/pull/9

jjeliga:
This PR is hereby closed in favor of a simpler solution presented in https://github.com/dam-agents/dam/pull/2118

PR #1748, opened 2026-06-24
chore(platform-base): drop gh source build, npm bump & gws prefetch; port install-pnpm to JS

JanPokorny:
Cleanup of the `platform-base` image build now that mise permissions are sorted. Each change is independent.

### `gh`: install via mise instead of building from source
The `gh-builder` stage existed solely to force-bump `in-toto-golang` to v0.11.0 (fixing [GHSA-pmwq-pjrm-6p5r](https://github.com/advisories/GHSA-pmwq-pjrm-6p5r), a transitive dep) and rebuild. **`gh` 2.95.0 already ships `in-toto-golang` v0.11.0**, so the official release binary is no longer CVE-tainted. Now installed through the existing `mise install`; `gh = "latest"` in `mise.toml` with the version pinned in `mise.lock` (2.95.0, linux arm64+x64). Removes a cross-compile Go build stage.

| `gh` tag | in-toto-golang | status |
|---|---|---|
| 2.93.0 (was pinned) | v0.9.0 | ⚠️ vulnerable |
| 2.95.0 | v0.11.0 | ✅ fixed |

### Drop explicit `npm@11.16.0` install
Node 24 (locked 24.14.1) bundles npm **11.11.0**, comfortably above the **11.5.1** floor for [trusted publishing / OIDC](https://docs.npmjs.com/trusted-publishers/). The explicit global install (and the `mise reshim` that only existed to re-shim after it) are gone.

### Drop root `gws --version` prefetch
This ran once as root to bake in the on-demand platform binary. The `chown -R 65532 /usr/local/share/mise` makes the install dir writable by the agent user, so the binary downloads on first use. Verified in the built image as UID 65532 — cold `gws --version` downloads, checksum-verifies, and runs cleanly. *Tradeoff: first `gws` call in a fresh pod now needs GitHub egress.*

### Port `install-pnpm.sh` → `install-pnpm.js`
The script was already shelling out to `node` for JSON parsing and hashing. Rewritten as ESM JS — drops the shell parsing/parameter-expansion half. All four Dockerfiles that install it (`platform-base`, `ui`, `api-server`, `e2e/agents/mock`) updated to `COPY --chmod=755 scripts/install-pnpm.js`.

## Verification
- `platform-base` image rebuilds clean
- `gws --version` cold-start as UID 65532 → downloads + runs, RC=0
- `install-pnpm` as extensionless ESM bin in `node:24`: installs `pnpm@10.33.4` (sha512 verified); malformed `packageManager` → error, rc=1


willallstet:
📝 Docs reviewed for this change — no documentation update required (internal platform-base image build pipeline (drop gh source build, npm bump, port install-pnpm to JS); no user-facing behavior.)

PR #1898, opened 2026-06-24
feat(skills): load skills from a source subdirectory

### LINKED ISSUE #276: Allow loading skills from a subdirectory of a source repo
## Problem

Adding a skill source via the UI only works if the repo's skills live at the root or under a top-level `skills/` directory. If a repo keeps its skills nested under a different path (for example under a `.claude/` folder, or alongside other content), the source either lists nothing or surfaces an error, even though valid SKILL.md folders exist in the repo.

Today this forces every shared catalog to be its own dedicated repo with skills in one of the two recognised locations. Teams that want to keep skills next to the project that owns them, or in any non-default folder, can't share them through a source — they have to move or duplicate them.

## Proposed solution

When adding (or editing) a skill source, let the user **optionally** point at a specific path inside the repo. If a path is given, the scanner walks it the same way it walks `skills/` today: any direct subdirectory containing a `SKILL.md` is picked up as a skill. If no path is given, behaviour stays exactly as today (look at `skills/`, then the repo root) — existing sources keep working without changes.

### Out of scope

- Recursive deep search for skills anywhere in the repo — keep the resolution explicit so users know where their skills come from.
- Per-skill cherry-picking inside one source — a source still maps to a folder of skills, not a hand-picked list.

kapetr:
## What

Skill sources can now optionally point at a **subdirectory** inside the repo. When a path is set, that directory is scanned for skills the same way `skills/` is scanned today — any direct subfolder with a `SKILL.md` is picked up. When no path is set, behaviour is unchanged.

This lets teams keep skills next to the project that owns them (e.g. in a monorepo subfolder) and share them through a single source, instead of carving out a dedicated repo with skills at the root or under `skills/`.

## How to use

- **UI:** the Add Source form has an optional, labeled **Path** field (with a tooltip explaining the auto-discovered roots); configured paths show on the source row, and the "View SKILL.md" link points at the right folder.
- **CLI:** `dam skill source add <url> --path <subdir>`; `dam skill source list` shows a PATH column.
- **Platform/template sources:** Helm `skills.skillSources` and template `spec.skillSources` entries accept an optional `path`.

A path is only needed for **non-standard** locations — when empty, skills are auto-discovered in the standard source roots (`skills/`, `.claude/skills/`, `.agents/skills/`). One subdirectory per repo connection; to change it, remove and re-add the source. Sources without a path keep working exactly as before.

## Side findings (out of scope for this PR)

Surfaced while building this; tracked here for follow-up, not addressed in this PR:

- **Add-source form should move to a modal.** This PR adds field labels because the bare inputs were ambiguous, but the form is still an inline expanding panel. A modal would be cleaner UX — likely folded into the coming Skills panel redesign.
- **No edit action for sources.** Sources support create + delete only; there's no edit, so changing a path (or name) means remove + re-add. Worth adding a proper edit action.
- **Skills in a locally cloned repo can't be used from that local copy.** Install *copies* the skill from the pushed git source into the skill path (`.agents/skills`), so local edits must be pushed and re-installed before they take effect. A follow-up could let a skill be *referenced* from an on-pod clone — e.g. a symlink into the skill path pointing at the local repo — instead of copying (today the install path also rejects symlinked skill content, so this is genuinely new work).

Closes #276

### DIFF
--- a/packages/api-server/src/modules/skills/infrastructure/public-archive-scanner.ts
+++ b/packages/api-server/src/modules/skills/infrastructure/public-archive-scanner.ts
@@ -116,7 +116,24 @@ export async function computeContentHash(absDir: string): Promise<string> {
   return h.digest("hex");
 }
 
-async function findSkillDirs(repoDir: string): Promise<string[]> {
+function subPathEscapes(subPath: string): boolean {
+  return subPath.startsWith("/") || subPath.split("/").includes("..");
+}
+
+async function findSkillDirs(
+  repoDir: string,
+  subPath?: string,
+): Promise<string[]> {
+  // An explicit subdir is scanned exclusively — no source-root union or root
+  // fallback, so the user gets exactly the directory they pointed at. The path
+  // is api-validated at source creation; guard here too so a stray
+  // `..`/absolute path can never escape the extracted archive.
+  if (subPath) {
+    if (subPathEscapes(subPath)) {
+      throw new Error(`skill source path rejected: ${subPath}`);
+    }
+    return skillDirsUnder(repoDir, path.join(repoDir, subPath));
+  }
   const found: string[] = [];
   for (const root of SKILL_SOURCE_ROOTS) {
     found.push(...(await skillDirsUnder(repoDir, path.join(repoDir, root))));
@@ -156,6 +173,7 @@ async function skillDirsUnder(
  */
 export async function scanPublicGithubArchive(
   gitUrl: string,
+  subPath?: string,
 ): Promise<Skill[]> {
   const host = detectHost(gitUrl);
   if (!host)
@@ -197,7 +215,7 @@ export async function scanPublicGithubArchive(
       throw new Error("tarball contained no directories");
     const repoDir = path.join(tmp, extracted[0].name);
 
-    const skillDirs = await findSkillDirs(repoDir);
+    const skillDirs = await findSkillDirs(repoDir, subPath);
     const scanned = await Promise.all(
       skillDirs.map(async (rel) => {
         const absDir = path.join(repoDir, rel);
--- a/packages/api-server/src/modules/skills/infrastructure/seed-sources.ts
+++ b/packages/api-server/src/modules/skills/infrastructure/seed-sources.ts
@@ -1,10 +1,11 @@
 import { z } from "zod/v4";
-import type { SkillSource } from "api-server-api";
+import { skillSourcePathSchema, type SkillSource } from "api-server-api";
 
 const seedSchema = z.array(
   z.object({
     name: z.string().min(1).max(128),
     gitUrl: z.url(),
+    path: skillSourcePathSchema.optional(),
   }),
 );
 
@@ -12,6 +13,7 @@ export interface SkillSourceSeed {
   id: string;
   name: string;
   gitUrl: string;
+  path?: string;
 }
 
 const SEED_ID_PREFIX = "skill-src-seed-";
@@ -76,6 +78,7 @@ export function parseSeedSources(raw: string | undefined): SkillSourceSeed[] {
       id: `${SEED_ID_PREFIX}${slug}`,
       name: entry.name,
       gitUrl: entry.gitUrl,
+      ...(entry.path ? { path: entry.path } : {}),
     };
   });
 }
@@ -83,5 +86,11 @@ export function parseSeedSources(raw: string | undefined): SkillSourceSeed[] {
 /** Surface seeds as `SkillSource` views, tagged `system: true` so the UI's
  *  "Platform" badge and the protected-source delete check both fall out. */
 export function seedToSkillSource(seed: SkillSourceSeed): SkillSource {
-  return { id: seed.id, name: seed.name, gitUrl: seed.gitUrl, system: true };
+  return {
+    id: seed.id,
+    name: seed.name,
+    gitUrl: seed.gitUrl,
+    system: true,
+    ...(seed.path !== undefined ? { path: seed.path } : {}),
+  };
 }
--- a/packages/api-server/src/modules/skills/services/skills-service.ts
+++ b/packages/api-server/src/modules/skills/services/skills-service.ts
@@ -62,18 +62,20 @@ export interface SkillsServiceDeps {
   runtimeClient: AgentRuntimeSkillsClient;
   runtimeMutator: RuntimeMutator;
   owner: string;
-  /** Scan via the provided scanner with a shared TTL cache. The cache key is
-   *  the gitUrl alone — results are user-independent. */
+  /** Scan via the provided scanner with a shared TTL cache, keyed by
+   *  `(gitUrl, path)` — the catalogue depends on both, and the cache is shared
+   *  across users who may point the same repo at different subdirs. */
   scanSource: (
     gitUrl: string,
+    path: string | undefined,
     scanner: (gitUrl: string) => Promise<Skill[]>,
   ) => Promise<Skill[]>;
-  invalidateScan: (gitUrl: string) => void;
+  invalidateScan: (gitUrl: string, path: string | undefined) => void;
   /** Scan a public GitHub repo directly from the api-server pod. Throws
    *  `PublicArchiveNotFoundError` when the archive endpoint returns 404 —
    *  signal to the caller to fall back to the agent-runtime path for
    *  private-repo auth (if the instance is running). */
-  scanPublic: (gitUrl: string) => Promise<Skill[]>;
+  scanPublic: (gitUrl: string, path?: string) => Promise<Skill[]>;
   /** Brand display name surfaced in publish-PR bodies. Sourced from runtime
    *  brand config so a deployment rebrand doesn't need a code change. */
   brandName: string;
@@ -113,6 +115,7 @@ async function loadTemplateSources(
     id: templateSourceId(template.id, seed.gitUrl),
     name: seed.name,
     gitUrl: seed.gitUrl,
+    ...(seed.path !== undefined ? { path: seed.path } : {}),
     fromTemplate: { templateId: template.id, templateName: template.name },
   }));
 }
@@ -138,6 +141,7 @@ async function resolveTemplateSource(
     id,
     name: seed.name,
     gitUrl: seed.gitUrl,
+    ...(seed.path !== undefined ? { path: seed.path } : {}),
     fromTemplate: { templateId: template.id, templateName: template.name },
   };
 }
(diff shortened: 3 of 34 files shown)


packages/api-server/src/modules/skills/infrastructure/public-archive-scanner.ts
```diff
@@ -116,7 +116,13 @@ export async function computeContentHash(absDir: string): Promise<string> {
...
 }
 
-async function findSkillDirs(repoDir: string): Promise<string[]> {
+async function findSkillDirs(
+  repoDir: string,
+  subPath?: string,
+): Promise<string[]> {
+  // An explicit subdir is scanned exclusively — no source-root union or root
+  // fallback, so the user gets exactly the directory they pointed at.
+  if (subPath) return skillDirsUnder(repoDir, path.join(repoDir, subPath));
```
dam-code-guardian: 🟢 **Suggestion:** `findSkillDirs` joins `subPath` directly with no local `..`/absolute guard, unlike the agent-runtime side (`subPathEscapes` in `local-skill-repository.ts`). The path is validated by `skillSourcePathSchema` at source creation so this is safe today; a matching belt-and-suspenders guard here would harden the api-server scan path against any future caller that bypasses the schema.

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `098cf6e`

## PR #1898: feat(skills): load skills from a source subdirectory
**Author:** kapetr | **Branch:** feat/276-skill-source-subpath → main | **Changes:** +1953 −89 (29 files)

### Summary
Adds an optional repo-relative `path` to skill sources so a single source can point at a subdirectory (e.g. a monorepo's `.claude/skills`) instead of only the default source roots. When set, that subdir is scanned and skills resolved exclusively; when unset, behavior is unchanged. The field is threaded end-to-end: zod schemas (api-server-api / agent-runtime-api), DB columns on `skill_sources` and `agent_skills` (migration `0003`), scan/install services, CLI `--path` + PATH column, Helm/template seeds, and the Add-Source UI form (labeled Path field with tooltip). Path is denormalized onto each installed ref so the apply path resolves without re-reading the source.

### Findings
- 🟡 **Warning:** No tests accompany this change. The new logic includes security-sensitive path-traversal guards (`subPathEscapes`, `packages/agent-runtime/.../local-skill-repository.ts:261`), exclusive-subdir scan semantics, the `(gitUrl, path)` cache key (`packages/api-server/src/modules/skills/compose.ts:30`), and the precedence-based `resolveSourcePathByGitUrl` (`packages/api-server/src/modules/skills/services/skills-service.ts:205`). The repo has unit tests for this area (`public-archive-scanner.test.ts`) but none were added/extended for `path`. Recommend covering at minimum: a `..`/absolute path is rejected, an explicit subdir scans exclusively (no source-root fallback), and the cache distinguishes two paths on the same `gitUrl`.
- 🟢 **Suggestion:** The cache-key comment says "NUL separator" but the key actually joins with a space (`` `${gitUrl} ${path ?? ""}` ``) (`packages/api-server/src/modules/skills/compose.ts:31`). A space is still collision-proof (neither a URL nor a validated path can contain one), so this is a comment/code mismatch, not a bug — update the comment (or switch to an actual `\0`). _(Summary-only: this line is not part of this PR's GitHub diff, so no inline comment.)_
- 🟢 **Suggestion:** `public-archive-scanner.findSkillDirs` accepts `subPath` and joins it directly (`packages/api-server/src/modules/skills/infrastructure/public-archive-scanner.ts:125`) with no local `..`/absolute guard, unlike the agent-runtime side which added `subPathEscapes` as belt-and-suspenders. The path is validated by `skillSourcePathSchema` at source creation, so this is safe today, but a matching guard at the join site would harden the api-server path against any future caller that bypasses the schema.
- ✅ **Looks good:** Defense-in-depth on traversal — `skillSourcePathSchema` rejects `..` and trims slashes at the api boundary (`packages/api-server-api/src/modules/skills/schemas.ts:5`), and the agent-runtime re-checks via `subPathEscapes` at the filesystem join site before any `path.join`.
- ✅ **Looks good:** Backward-compatible throughout — `path` is optional in every schema, nullable in both new DB columns, and consistently spread conditionally (`...(x !== undefined ? { path } : {})`) so absent-path installs serialize identically to pre-PR. `contribution-hash` includes `path` in the skill-ref key, so changing a source's subdir correctly re-contributes.
- ✅ **Looks good:** `resolveSourcePathByGitUrl` reuses the same `dedupeByGitUrl([...owned, ...seeds, ...template])` precedence (user → system → template) as `listSources`, so the path denormalized at install time matches the source the user sees.

**Verdict: minor drift**

The skill-source path feature is documented in `docs/architecture/skills.md`: a new paragraph under `### Skill Source` describes the optional path (exclusive scanning, server-side resolution, denormalization onto `agent_skills`), the agent-runtime Install/Scan bullets were updated to honor it, and `Last verified:` is bumped to the PR date (2026-06-24). No ADR references, no volatile package/path leakage beyond the existing `agent_skills` identifier already in the page.

- 🟡 One bullet lags the code: the **api-server "Public-archive scanning"** bullet (`docs/architecture/skills.md:114`) still reads "enumerates skills across the [Source Roots]" with no mention of `path`, even though `public-archive-scanner.ts` now scans an explicit `subPath` exclusively. The general Skill Source paragraph covers the behavior conceptually, so this is minor — consider adding "(or the source's `path` when set)" to that bullet for parity with the agent-runtime Install/Scan bullets.

### Verdict
COMMENT — Clean, backward-compatible, well-layered feature with good traversal defense-in-depth; the only real gap is the absence of tests for security-sensitive new logic, plus a one-line doc bullet that lags the code. Neither blocks, but tests are strongly recommended before merge.

---

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `77c3b41`

## PR #1898: feat(skills): load skills from a source subdirectory
**Author:** kapetr | **Branch:** feat/276-skill-source-subpath → main | **Changes:** +2086 −82 (30 files)

### Summary
Adds an optional repo-relative `path` to skill sources so a single source can point at a subdirectory (e.g. a monorepo's `.claude/skills`) instead of only the default source roots. When set, that subdir is scanned and skills resolved exclusively; when unset, behavior is unchanged. Threaded end-to-end across zod schemas, DB columns (migration `0003`), scan/install services, CLI `--path` + PATH column, Helm/template seeds, and the Add-Source UI form.

### Changes since last review
Previous HEAD: `098cf6e` (2026-06-24T13:16:04Z) — verdict COMMENT

- ✅ **Fixed:** api-server `findSkillDirs` had no local `..`/absolute guard (`packages/api-server/src/modules/skills/infrastructure/public-archive-scanner.ts:125`) — a `subPathEscapes` guard was added at the join site, now identical to the agent-runtime implementation.
- ✅ **Fixed:** cache-key comment said "NUL separator" but the key joined with a space (`packages/api-server/src/modules/skills/compose.ts:31`) — the key now uses an actual `\0` NUL separator, so comment and code agree.
- ✅ **Fixed:** No tests for the security-sensitive new logic — `public-archive-scanner.test.ts` gained two cases: an explicit subPath scans exclusively (ignoring source roots), and a traversal subPath (`../../etc`) is rejected.
- ✅ **Fixed:** doc-drift — the api-server "Public-archive scanning" bullet (`docs/architecture/skills.md:114`) now reads "(or the source's `path` when set)", matching the code.

### Findings
- 🟢 **Suggestion:** Tests now cover exclusive-subdir scan and traversal rejection on the api-server side; a case asserting the scan cache distinguishes two `path` values on the same `gitUrl` (`packages/api-server/src/modules/skills/compose.ts:30`) would round out the new-logic coverage. Non-blocking.
- ✅ **Looks good:** All findings from the prior review are resolved — the api-server traversal guard now mirrors agent-runtime's `subPathEscapes`, the cache-key comment matches the code, the security-sensitive scan paths have test coverage, and the documentation bullet is in sync.

The skill-source `path` feature is fully documented in `docs/architecture/skills.md` — the `### Skill Source` paragraph (exclusive scanning, server-side resolution, denormalization onto `agent_skills`), the agent-runtime Install/Scan bullets, and the api-server "Public-archive scanning" bullet all reflect the code. `Last verified: 2026-06-24` matches the PR date. No ADR references, no volatile-content leak.

### Verdict
APPROVE — A clean, backward-compatible, well-layered feature; the re-review confirms every finding from the prior review (traversal guard parity, cache-key comment, missing tests, doc bullet) has been addressed. One optional test-coverage suggestion remains, non-blocking.

---

willallstet:
📝 Docs updated for this change: https://github.ibm.com/dam-agents/docs/pull/23

PetrBulanek:
Publish writes to `skills/` regardless of the source's configured path — `packages/agent-runtime/src/modules/skills/services/publish.ts:65`

The blob path is hardcoded to `skills/${name}/${f.relPath}`. A subdir source is still a GitHub URL, so the UI/CLI will offer Publish for it — but the PR lands the skill under `skills/<name>/`, while this same source's scanner now reads `<path>/` exclusively (no skills/ fallback). So a skill published back to a subdir source ends up where its own scanner will never find it.

packages/api-server/src/modules/skills/services/skills-service.ts
```diff
@@ -192,6 +196,24 @@ function dedupeByGitUrl(list: SkillSource[]): SkillSource[] {
   return out;
 }
 
+/** Recover a source's subdir from its gitUrl using the same merge + dedupe
+ *  precedence as listSources (user → system → template). Install carries the
+ *  gitUrl, not the source id, so this is how the path is found to denormalize
+ *  onto the installed ref. */
+async function resolveSourcePathByGitUrl(
```
PetrBulanek, line 203: The doc says dedupe precedence is "user → template → system", but the code is actually "user → system → template" (`[...owned, ...seeds, ...template]`).

packages/api-server/src/modules/skills/infrastructure/seed-sources.ts
```diff
@@ -5,13 +5,19 @@ const seedSchema = z.array(
   z.object({
     name: z.string().min(1).max(128),
     gitUrl: z.url(),
+    path: z
+      .string()
+      .transform((p) => p.trim().replace(/^\/+|\/+$/g, ""))
+      .refine((p) => !p.split("/").includes(".."), "path must not contain '..'")
+      .optional(),
```
PetrBulanek: This re-implements the exact trim → strip slashes → reject ".." rule inline, but the canonical `skillSourcePathSchema` already exists at `packages/api-server-api/src/modules/skills/schemas.ts:5` (and is reused by the template schema). Two copies of the same validation can drift. Suggest exporting`skillSourcePathSchema` from api-server-api and importing it here so there's one definition.

PetrBulanek (approved):

PR #1899, opened 2026-06-24
chore: migrate env from agent spec to contributions

### LINKED ISSUE #1079: Deliver custom environment variables through the contribution rail, not the agent CR

jjeliga:
## What & why

User-typed environment variables (the UI **Environment** editor) were the last piece of agent config still delivered through the Agent CR's `spec.env` and projected into the container by the controller at pod start — so an edit only took effect on the **next pod restart**, unlike connections/secrets/skills/MCP which all ride the runtime contribution rail and apply on the next turn.

This moves user env onto the **same contribution rail**: it's stored in a new `agent_env` table, emitted as an `env` contribution, and reconciled into the harness at the next idle turn boundary — **no pod roll**, and **one** env mechanism for everything.

Closes #1079.

## How it works

- **Storage** — new `agent_env` table (`agent_id`, `name`, `value`), the `agent_skills` pattern.
- **Read path** — a new source in the state-builder reads `agent_env` and emits `env` contributions, placed **first** so user env wins over connection/secret env on name collision (the in-pod driver is first-occurrence-wins — preserving today's precedence now that user env no longer rides `process.env`).
- **Write/read-back** — the agents service writes/reads `agent_env` (and bumps + enqueues a state push on edit) instead of `spec.env`. The tRPC contract, UI editor, validation, and inherited/shadowing rules are unchanged.
- **Controller** — stops projecting `spec.env` into the agent and fork containers; projects only chart-level platform defaults. The CRD `spec.env` field is **retained but unread** (no CRD schema change → no backwards-compat exposure).
- **Migration** — an idempotent, self-disarming startup backfill moves any existing `spec.env` onto `agent_env`, then clears the CR field. Per-agent fault-isolated, never crashes startup, logged, and retried on a timer until a clean pass.

### Notes
- `PORT` was never actually carried on `spec.env` (it's the hardcoded container port + agent-runtime's own default), so dropping the projection can't affect it; it stays a protected name.
- Platform-infra env (proxy, CA, `HOME`, `PLATFORM_*`) stays controller-injected — it's bootstrap-critical and can't ride the rail it bootstraps.

## Testing

- **e2e** (`09-user-env.spec.ts`): asserts the agent's **live process env** — set reaches the agent with no pod roll, edit applies, user env shadows a connection env, and clearing reverts to the connection value.
- **Manual, on a dev cluster** — verified the backfill end-to-end on an actual **hibernated** agent: `migrated 1`, `agent_env` row written, CR `spec.env` cleared, existing rows untouched; a re-run logs `migrated 0` with no resurrection/duplication (self-disarming).
- Controller Go tests updated to assert `spec.env` is no longer projected.

No unit tests by request; team leans e2e.

### DIFF
--- a/packages/api-server/src/modules/agents/infrastructure/agent-env-repository.ts
+++ b/packages/api-server/src/modules/agents/infrastructure/agent-env-repository.ts
@@ -1,5 +1,5 @@
 import type { Db } from "db";
-import { agentEnv, eq, inArray, sql } from "db";
+import { agentEnv, eq, inArray } from "db";
 import type { EnvVar } from "api-server-api";
 
 /** Postgres store for user-typed agent env, keyed by `agent_id` (ownership enforced by the agents service). */
@@ -62,12 +62,10 @@ export function createAgentEnvRepository(db: Db): AgentEnvRepository {
     },
 
     async listAgentIds() {
-      const rows = await db.execute<{ agent_id: string }>(sql`
-        SELECT DISTINCT agent_id FROM ${agentEnv}
-      `);
-      return (rows as unknown as Array<{ agent_id: string }>).map(
-        (r) => r.agent_id,
-      );
+      const rows = await db
+        .selectDistinct({ agentId: agentEnv.agentId })
+        .from(agentEnv);
+      return rows.map((r) => r.agentId);
     },
   };
 }
--- a/packages/api-server/src/modules/agents/infrastructure/agent-env-repository.ts
+++ b/packages/api-server/src/modules/agents/infrastructure/agent-env-repository.ts
@@ -1,0 +1,73 @@
+import type { Db } from "db";
+import { agentEnv, eq, inArray, sql } from "db";
+import type { EnvVar } from "api-server-api";
+
+/** Postgres store for user-typed agent env, keyed by `agent_id` (ownership enforced by the agents service). */
+export interface AgentEnvRepository {
+  list(agentId: string): Promise<EnvVar[]>;
+  /** Batched form; every input id is present in the map. */
+  listMany(agentIds: string[]): Promise<Map<string, EnvVar[]>>;
+  /** Replace an agent's full env set (the editor sends the complete list). */
+  replace(agentId: string, env: EnvVar[]): Promise<void>;
+  deleteForAgent(agentId: string): Promise<void>;
+  /** Distinct agent ids with env rows, for the orphan sweeper. */
+  listAgentIds(): Promise<string[]>;
+}
+
+export function createAgentEnvRepository(db: Db): AgentEnvRepository {
+  return {
+    async list(agentId) {
+      const rows = await db
+        .select({ name: agentEnv.name, value: agentEnv.value })
+        .from(agentEnv)
+        .where(eq(agentEnv.agentId, agentId));
+      return rows.map((r) => ({ name: r.name, value: r.value }));
+    },
+
+    async listMany(agentIds) {
+      const map = new Map<string, EnvVar[]>();
+      for (const id of agentIds) map.set(id, []);
+      if (agentIds.length === 0) return map;
+      const rows = await db
+        .select({
+          agentId: agentEnv.agentId,
+          name: agentEnv.name,
+          value: agentEnv.value,
+        })
+        .from(agentEnv)
+        .where(inArray(agentEnv.agentId, agentIds));
+      for (const r of rows) {
+        map.get(r.agentId)?.push({ name: r.name, value: r.value });
+      }
+      return map;
+    },
+
+    async replace(agentId, env) {
+      // Dedupe by name (last wins) so a doubly-typed key can't violate the PK.
+      const byName = new Map<string, string>();
+      for (const e of env) byName.set(e.name, e.value);
+      const rows = [...byName].map(([name, value]) => ({
+        agentId,
+        name,
+        value,
+      }));
+      await db.transaction(async (tx) => {
+        await tx.delete(agentEnv).where(eq(agentEnv.agentId, agentId));
+        if (rows.length > 0) await tx.insert(agentEnv).values(rows);
+      });
+    },
+
+    async deleteForAgent(agentId) {
+      await db.delete(agentEnv).where(eq(agentEnv.agentId, agentId));
+    },
+
+    async listAgentIds() {
+      const rows = await db.execute<{ agent_id: string }>(sql`
+        SELECT DISTINCT agent_id FROM ${agentEnv}
+      `);
+      return (rows as unknown as Array<{ agent_id: string }>).map(
+        (r) => r.agent_id,
+      );
+    },
+  };
+}
--- a/packages/api-server/src/modules/agents/services/agents-service.ts
+++ b/packages/api-server/src/modules/agents/services/agents-service.ts
@@ -278,6 +278,10 @@ export function createAgentsService(deps: {
           description: input.description,
         });
       }
+      // Template-declared env rides the rail like user env (seeded below), not
+      // the CR — the controller no longer reads spec.env.
+      const templateEnv = (spec.env as EnvVar[] | undefined) ?? [];
+      delete spec.env;
       if (input.secretRef !== undefined) spec.secretRef = input.secretRef;
 
       // Single-shot create: seed grants into the spec before first render so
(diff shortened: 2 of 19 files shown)


packages/api-server/src/modules/agents/services/agents-service.ts
```diff
@@ -408,20 +420,27 @@ export function createAgentsService(deps: {
...
+          : await deps.repo.get(input.id, deps.owner);
       if (!infra) return null;
 
+      let env = input.env;
+      if (env !== undefined) {
+        // Strip protected names, then bump + enqueue so a running agent applies it next turn.
+        env = preserveProtectedEnvs([], env);
+        await deps.agentEnvRepo.replace(input.id, env);
+        await deps.runtimeMutator.bump(input.id, []);
+        await deps.runtimeMutator.enqueueAfterCommit(input.id);
```
dam-code-guardian, line 446: 🟢 **Suggestion:** The env write, version bump, and enqueue run as three independent awaits here, not one transaction. If `bump` or `enqueueAfterCommit` fails after `replace` succeeds, the `agent_env` row is persisted but no state push is enqueued — a running agent won't pick up the edit until the next unrelated bump. Consider wrapping these three in a single transaction (or documenting the retry expectation).

packages/api-server/src/modules/agents/infrastructure/agent-env-repository.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    async deleteForAgent(agentId) {
+      await db.delete(agentEnv).where(eq(agentEnv.agentId, agentId));
+    },
+
+    async listAgentIds() {
+      const rows = await db.execute<{ agent_id: string }>(sql`
+        SELECT DISTINCT agent_id FROM ${agentEnv}
+      `);
+      return (rows as unknown as Array<{ agent_id: string }>).map(
```
dam-code-guardian: 🟢 **Suggestion:** This uses a raw `db.execute` with a `rows as unknown as Array<...>` double-cast, bypassing Drizzle's typing while every other method here uses the typed query builder. `selectDistinct({ agentId: agentEnv.agentId })` keeps it type-safe and consistent.

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `9e80a27`

## PR #1899: chore: migrate env from agent spec to contributions
**Author:** jjeliga | **Branch:** chore/1079-migrate-env-from-spec-to-contributions → main | **Changes:** +2086 −44 (18 files)

### Summary
Moves user-typed env off the Agent CR's `spec.env` (applied only on pod restart) onto the runtime contribution rail: stored in a new Postgres `agent_env` table, emitted as `env` contributions ordered first so user env wins on name collision, and reconciled at the next idle turn — no pod roll. The controller stops projecting `spec.env`; an idempotent, self-disarming startup backfill migrates existing `spec.env` and clears the CR field. The core mechanism checks out: I verified the env driver is first-occurrence-wins (`env-plugin.ts:30`) and that `builtin` contributions emit only `mcp-entry` (no `env`), so ordering user env ahead of connection/secret env correctly shadows them exactly as claimed.

### Findings
- 🟢 **Suggestion:** In `update()`, the env write, version bump, and enqueue run as three independent awaits (`agents-service.ts:439-441`), not one transaction. If `bump`/`enqueueAfterCommit` fails after `replace` succeeds, the `agent_env` row is persisted but no state push is enqueued, so a running agent won't see the edit until the next unrelated bump. Consider wrapping the three in a single transaction (or at least documenting the at-least-once-retry expectation).
- 🟢 **Suggestion:** `listAgentIds()` uses a raw `db.execute` with a `rows as unknown as Array<...>` double-cast (`agent-env-repository.ts:64-68`), bypassing Drizzle's typing while every other method in this repo uses the typed query builder. A `selectDistinct({ agentId: agentEnv.agentId })` keeps it type-safe and consistent.
- ✅ **Looks good:** Backfill is genuinely idempotent and self-disarming — skips the write when `agent_env` rows already exist (never clobbers a post-migration edit), clears `spec.env` to disarm re-runs, per-agent fault isolation, never throws, retried on a 60s timer until a clean pass, and the retry timer is cleared on shutdown.
- ✅ **Looks good:** Lifecycle wiring is complete — `agentEnvCleanupHook` registered in both the per-agent cleanup hooks and the orphan-artifacts sweeper, so `agent_env` rows can't leak after agent deletion.
- ✅ **Looks good:** `replace()` dedupes by name before insert so a doubly-typed key can't violate the composite PK; protected names (`PORT`) are stripped on every write path.
- ✅ **Looks good:** e2e spec asserts the live process env across set / edit / shadow-connection / clear-reverts — exercises the no-pod-roll path end to end; all referenced fixtures and the `e2e.getEnv` endpoint resolve.

**Verdict: significant drift**

**Drift**

**Check 1 — `docs/architecture/agent-lifecycle.md`** ❌
The PR makes the controller stop projecting `spec.env`/`agent.env` and routes user env over the runtime rail (applies next turn, no pod roll), but `agent-lifecycle.md` still documents the old model and was not updated:
- `agent-lifecycle.md:64-67` — "Pod env at start is the composition of **three** layers… 3. **`agent.env`** — the single env list on the Agent's spec." Layer 3 is no longer projected at pod start; the list is stale.
- `agent-lifecycle.md:69` — "Editing `agent.env` takes effect on the next pod restart." Now false — that is precisely the behavior this PR changes.
_Proposed: describe user-typed env as a runtime-channel contribution (Postgres `agent_env`, applied next idle turn, no pod roll); leave only platform + credential env projected at pod start; bump `Last verified:` to 2026-06-24._

**Check 2 — `docs/architecture/connections.md` `Last verified:` staleness** ❌ (minor)
`connections.md:3` reads `Last verified: 2026-06-23` but the PR date is 2026-06-24. Bump to 2026-06-24.

**Possible drift**

**Check 1 — `docs/architecture/persistence.md`** ⚠️
The PR adds the Postgres `agent_env` table. `connections.md`'s substrate table was updated to list it, but `persistence.md`'s Postgres substrate enumeration (`persistence.md:55-60`) was not. Human judgement: is that bullet list meant to be exhaustive (add a user-env bullet), or is the connections-page substrate table the canonical home for this table?

### Verdict
COMMENT — Solid, well-tested migration with correct precedence and a robust backfill; the actionable item is updating `agent-lifecycle.md` (and bumping `connections.md`'s verified date) to match the new no-pod-roll env model.

---

packages/api-server/src/modules/agents/services/agents-service.ts
```diff
@@ -333,6 +338,13 @@ export function createAgentsService(deps: {
         throw e;
       }
 
+      // The agent's boot `hello` pulls agent_env on first apply, so no enqueue here.
+      const userEnv = input.env?.length
```
PetrBulanek: Template env is dropped for new agents: this seeds `agent_env` from `input.env ` only, but template env lives in the assembled `spec.env`, which the controller no longer reads (resources.go:114). Inert today (no template sets env) but the chart supports `env:` on templates.

PetrBulanek (approved):

willallstet:

willallstet:
📝 Docs updated for this change: https://github.ibm.com/dam-agents/docs/pull/44

PR #2033, opened 2026-06-26
feat(experiments): race competing R&D harnesses against one goal

tomkis:
Integration branch for the Experiments epic (`dam-u1n`).

## What
An Experiment races several AI-driven R&D harnesses against one goal and compares what each produced. Each Arm is an existing agent (harness = its image) plus a config. Starting opens one Trial session per Arm; each harness runs its own loop and appends scored Runs to a shared ledger. The platform starts arms, captures runs, presents them. It never runs the loop and never interprets a score.

## The bet
No bespoke experiment harness. A generic agent runs its own iterate-and-score loop and populates the ledger because the Trial prompt tells it to: the prompt carries the full `record_run` / `finish_arm` contract inline and instructs the harness to report every scored candidate as it lands. The contract lives in the session-scoped prompt rather than a skill, so it is harness-agnostic (skills reach only Claude-family harnesses that mount `.agents/skills`) and never leaks reporting nags into non-experiment agents. The platform starts the arm, captures what it reports, and presents the comparison; it never runs the loop or interprets a score.

## Tracer bullet
Thinnest end-to-end slice: one experiment, one claude-code arm, started via the trigger path, reporting runs through `record_run` into the ledger, rendered on a new Experiments destination. Built on real reuse: experiments copy the Connections owner/grant model, start reuses the schedule trigger path, ingestion is an MCP tool, the reporting contract rides in the Trial prompt. Candidates are stored in Postgres by the api-server (capped), readable while the agent is offline.

Subtasks `dam-u1n.1` through `dam-u1n.11`. Sub-PRs target this branch; merges to `main` once the slice is verified (`dam-u1n.11`).

Seeded with the Experiments glossary only. Draft until the slice lands.

---

**Update (candidate storage).** `dam-u1n.2` originally provisioned a shared RWX artifact PVC mounted into both agent pods and the api-server. That does not work: a single PVC can't cross the agent and release namespaces, and the static-PV-over-shared-NFS workaround fails on the bundled nfs-ganesha provisioner (no stable `/export` export). S3 was considered and rejected (agents are egress-locked to their gateway). Candidates are now stored as capped bytea in Postgres, written and read by the api-server (#2036) — no PVC, no controller change, no agent mount. Still downloadable while the agent hibernates because Postgres is agent-independent. Ingestion: `record_run` hands the api-server a workspace path; the api-server reads the file and stores the blob, setting `experiment_runs.candidate_ref` to the artifact key.


tomkis:
Looked at letting users run experiments from non-existent sandboxes (pick a harness, UI creates the sandboxes first). Punting it out of the tracer. Why:

The happy path looks free. The wizard calls `agents.create` then `addArm`, the arm still references a real agent, the trial launcher is unchanged. No schema change.

But "create them in the UI" quietly picks a sandbox lifecycle, and both answers cost more than the tracer should:

- Durable sandboxes: every arm leaves a real StatefulSet + PVC behind. They pile up in the user's sandbox list, and the experiment doesn't know it created them, so there's no cleanup story. Cost and clutter grow per experiment.
- Throwaway sandboxes: the experiment now has to track which agents it minted and tear them down on stop, complete, and delete. That's the data-model change we were trying to avoid, plus teardown ordering against still-running trials.

Either way there's an atomicity problem: N creates plus N addArms from the browser, and a failure partway leaves orphaned half-created sandboxes with no owner to reap them.

None of this is hard, but it's a lifecycle and ownership design, not a UI affordance. Keeping the tracer on "arm references an existing sandbox" and filing a follow-up for auto-provisioning.

packages/api-server/src/apps/harness-api-server/mcp-endpoint.ts
```diff
@@ -468,6 +497,120 @@ export function createMcpSession(
...
+      const sessionId = await resolveTrialSessionId(active.experimentId);
+      if (!sessionId) {
+        return errorResult(
+          "no active trial session found for this experiment; start the experiment before recording runs.",
+        );
+      }
+
+      const key = `${active.experimentId}/${agentId}/${crypto.randomUUID()}/${basename(candidate)}`;
+      try {
+        await deps.artifacts.put({ key, content, contentType });
```
dam-code-guardian, line 559: 🟡 **Warning:** The blob is stored here before `experiments.recordRun` (line 566), which throws `CONFLICT` if the arm closed (finished/stopped/swept) between `resolveActiveArm` and the insert. On that path the artifact is already persisted with no `experiment_runs` row referencing it and no cleanup — a slow leak of capped `bytea` for every run that loses this race.

Consider recording the run first, or deleting the just-stored blob on the CONFLICT path, or a periodic orphan sweep over unreferenced artifact keys.

packages/api-server/src/modules/experiments/infrastructure/experiments-repository.ts
```diff
@@ -0,0 +1,563 @@
...
+    async finishArm(experimentId, agentId): Promise<ExperimentArm | null> {
+      const arm = await transitionRunningArm(
+        experimentId,
+        agentId,
+        "completed",
+      );
+      return arm ? rowToArm(arm) : null;
+    },
+
+    async findActiveArm(
```
dam-code-guardian, line 495: 🟢 **Suggestion:** This resolves the single most-recent `running` experiment for the agent (`orderBy(desc(createdAt)).limit(1)`). If an agent is ever a `running` arm of two experiments simultaneously, `record_run` silently attributes only to the newest and the older arm goes quiet until the inactivity sweep fails it. Worth an explicit guard rejecting the ambiguous case, or a comment documenting the one-active-experiment-per-agent assumption.

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `6ab1352`

## PR #2033: feat(experiments): race competing R&D harnesses against one goal
**Author:** tomkis | **Branch:** feat/experiments → main | **Changes:** +13723 −9 (92 files)

### Summary
Integration branch for the Experiments epic (`dam-u1n`): an Experiment races several agent "arms" (harness + config) against one goal, each running its own iterate-and-score loop and reporting scored Runs into a shared ledger via the `record_run` / `finish_arm` MCP tools. The platform starts arms, captures runs, and presents the comparison — it never runs the loop or interprets a score. The reporting contract rides in the session-scoped Trial prompt (harness-agnostic, no skill leakage), candidates are stored as capped `bytea` blobs in Postgres (agent-independent, downloadable while hibernated), and a per-replica inactivity sweeper guarantees every started Experiment reaches a terminal state. Large diff (~14k lines); ~8k is generated Drizzle migration snapshots. Review focused on the experiments api-server module, MCP ingestion, candidate download, and the arm/experiment state machine.

### Findings
- 🟡 **Warning:** Orphaned-blob window in `record_run` (`packages/api-server/src/apps/harness-api-server/mcp-endpoint.ts:559`). The candidate blob is stored via `artifacts.put` *before* `experiments.recordRun`, which throws `CONFLICT` when the arm closed (finished/stopped/swept) between `resolveActiveArm` and the insert. On that path the blob is already persisted in Postgres with no `experiment_runs` row referencing it and no cleanup — a slow leak of capped `bytea` for every run that loses this race. Consider inserting the run first (or deleting the blob on the CONFLICT path), or a periodic orphan sweep keyed on unreferenced artifact keys.
- 🟢 **Suggestion:** `findActiveArm` resolves the single most-recent running experiment for an agent (`packages/api-server/src/modules/experiments/infrastructure/experiments-repository.ts:495`, `orderBy(desc(createdAt)).limit(1)`). If an agent is ever a `running` arm of two experiments at once, its `record_run` calls silently attribute only to the newest and the older arm goes quiet until the inactivity sweep fails it. Likely out of scope for the tracer bullet, but worth an explicit guard (reject the ambiguous case) or a code comment documenting the one-active-experiment-per-agent assumption.
- ✅ **Looks good:** `agentId` is captured from the verified MCP session, never from tool input — `record_run`/`finish_arm` can't be spoofed to act on another agent's arm. The session-mismatch and unresolved-agent paths both `securityLog` and 404.
- ✅ **Looks good:** Candidate download (`candidate-route.ts`) is owner-scoped through `experimentsFor(user.sub).getWithRuns` (a non-owned experiment reads back as 404), the client-supplied `runId` is resolved to a server-stored `candidateRef` rather than trusting an artifact key, and the `Content-Disposition` filename is sanitized against header/quote injection.
- ✅ **Looks good:** State-machine concurrency is careful — consistent experiment-then-arm lock ordering (documented to prevent deadlock across finish/sweep/launch-fail/stop), `addRun` re-checks `running` under `SELECT … FOR UPDATE`, `failInactiveArm` re-checks the deadline inside the transition, and `completeIfAllTerminal` is guarded on the experiment still being `running` so a raced Stop is never overwritten.
- ✅ **Looks good:** Multi-replica sweeper is safe — atomic conditional transition means a contention race just no-ops on the second replica, and a randomized initial delay keeps replicas off lockstep.

**Verdict: minor drift**

**Possible drift — Check 5 (new subsystem without a page) / Check 1 (architecture-page drift):**
- The diff introduces a whole new **Experiments** subsystem (long-lived api-server module + polling sweeper, repository, trial launcher, MCP ingestion, new Postgres tables, agent-runtime experiment-trigger driver). The newly-added persistence bullet (`docs/architecture/persistence.md:62`) references its owner as plain text **"Owned by Experiments"**, breaking the page's convention where every sibling substrate bullet links to an owning architecture page (`[channels](channels.md)`, `[skills](skills.md)`, `[usage-tracking](usage-tracking.md)`, `[agent-lifecycle](agent-lifecycle.md)`). There is **no `docs/architecture/experiments.md`** and Experiments is **not listed** in `docs/architecture.md`'s subsystem index.
  - _Human-judgement question:_ This is an in-progress epic integration branch (tracer-bullet slice). Add the Experiments architecture page + landing-page entry now (making "Owned by Experiments" a real link), or defer the page until the slice lands on `main`? The dangling plain-text owner reference is the concrete drift signal.

_Verified clean: `Last verified:` correctly bumped 2026-06-22 → 2026-06-26; naming `run_artifacts` matches the page's established altitude (not a volatile-content leak); no ADR references._

### Verdict
COMMENT — Solid, security-conscious, well-tested slice with careful concurrency handling. Non-blocking: the orphaned-blob window in `record_run` is worth closing, and the new Experiments subsystem lacks an architecture page (acceptable to defer if the page lands with the epic on `main`).

---

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `2a4f41d`

## PR #2033: feat(experiments): race competing R&D harnesses against one goal
**Author:** tomkis | **Branch:** feat/experiments → main | **Changes:** +13820 −15 (96 files)

### Summary
Integration branch for the Experiments epic (`dam-u1n`): an Experiment races several agent "arms" (harness + config) against one goal, each running its own iterate-and-score loop and reporting scored Runs into a shared ledger via the `record_run` / `finish_arm` MCP tools. The platform starts arms, captures runs, and presents the comparison — it never runs the loop or interprets a score. The single new commit since the last review (`2a4f41d`) is **docs-only** and directly closes the previous review's doc-drift finding: it adds `docs/architecture/experiments.md`, lists Experiments in the architecture landing index, turns the `persistence.md` "Owned by Experiments" reference into a real link, and documents the new `experiment-trigger` event kind + experiment-driven wake path in `connections.md` / `agent-lifecycle.md`. The two code findings from the prior review are unchanged (the commit touched no code).

### Changes since last review
Previous HEAD: `6ab1352` (2026-07-01T08:24:34Z) — verdict COMMENT

- ✅ **Fixed:** Experiments subsystem lacked an architecture page / dangling plain-text "Owned by Experiments" owner reference in `docs/architecture/persistence.md` — resolved: new `docs/architecture/experiments.md` added, linked from `docs/architecture.md` index, and the persistence owner reference is now a real link.
- 🔁 **Still present:** Orphaned-blob window in `record_run` (`packages/api-server/src/apps/harness-api-server/mcp-endpoint.ts:559`) — carried over from previous review (docs-only commit; code path unchanged).
- 🔁 **Still present:** `findActiveArm` resolves only the most-recent running experiment (`packages/api-server/src/modules/experiments/infrastructure/experiments-repository.ts:495`) — carried over from previous review.

### Findings
- 🟡 **Warning:** Orphaned-blob window in `record_run` (`packages/api-server/src/apps/harness-api-server/mcp-endpoint.ts:559`). The candidate blob is stored via `artifacts.put` *before* `experiments.recordRun`, which throws `CONFLICT` when the arm closed (finished/stopped/swept) between `resolveActiveArm` and the insert. On that path the blob is already persisted in Postgres with no `experiment_runs` row referencing it and no cleanup — a slow leak of capped `bytea` for every run that loses this race. Consider inserting the run first (or deleting the blob on the CONFLICT path), or a periodic orphan sweep keyed on unreferenced artifact keys.
- 🟢 **Suggestion:** `findActiveArm` resolves the single most-recent running experiment for an agent (`packages/api-server/src/modules/experiments/infrastructure/experiments-repository.ts:495`, `orderBy(desc(createdAt)).limit(1)`). If an agent is ever a `running` arm of two experiments at once, its `record_run` calls silently attribute only to the newest and the older arm goes quiet until the inactivity sweep fails it. Note that the new `experiments.md` documents an Arm as keyed per `(Experiment, Agent)`, which prevents the same agent being two arms of *one* experiment — but not two arms across *different* experiments. Worth an explicit guard (reject the ambiguous case) or a code comment documenting the one-active-experiment-per-agent assumption.
- ✅ **Looks good:** The docs-only commit is a clean, accurate resolution of the prior doc-drift finding — the new architecture page correctly describes the subsystem at the right altitude (links out to code paths rather than inlining volatile symbols), and the `connections.md` / `agent-lifecycle.md` edits match the code on the branch (`experiment-trigger` event kind, experiment-driven wake).

**Verdict: aligned**

No drift. The previous review's minor-drift finding is fully resolved by this commit: `docs/architecture/experiments.md` now exists and is listed in the `docs/architecture.md` subsystem index; `persistence.md`'s owner reference is now `Owned by [experiments](experiments.md)` (matching the sibling-bullet linking convention); `connections.md` documents the new `experiment-trigger` event kind; `agent-lifecycle.md` documents the fourth (experiment-driven) wake path. All five edited pages have `Last verified:` bumped to 2026-07-01. No ADR references, no volatile-content leaks.

### Verdict
COMMENT — Docs-only commit that cleanly resolves the prior doc-drift finding; the two carried-over code items (orphaned-blob window, single-active-arm resolution) remain non-blocking and worth addressing before the epic lands on `main`.

---

jezekra1 (approved):

PR #2044, opened 2026-06-26
fix(ui): exclude provider connections from sandbox wizard step 3

### LINKED ISSUE #2037: Sandbox wizard lists the already-selected provider under "My connections"
## Problem

When creating a sandbox, the wizard walks you through three steps: pick an image, choose a provider, then assign any additional connections. The provider you choose in step 2 is automatically wired into the sandbox.

But in step 3, that same provider also appears in the "My connections" list — shown as an **unchecked** checkbox, as if it weren't connected. This is misleading in two ways:

- It implies the provider isn't attached, when it already is.
- If you leave the box unchecked (the natural assumption) and finish, the provider is connected to the sandbox anyway — so the checkbox state doesn't reflect reality.

The result is a confusing setup flow where what you see contradicts what actually happens.

## Proposed solution

In step 3, the "My connections" list should exclude provider-type connections entirely. Providers are chosen in step 2, and that's the only place they should appear. Step 3 should only list the non-provider connections (apps, MCP servers, and the like) that the user can optionally grant to the sandbox.

### Scope

- Step 3's "My connections" list only.
- No change to how providers are selected in step 2 or how they're attached to the sandbox.

PetrBulanek:
When creating a sandbox, the wizard's step 3 ("Grant connections") listed the provider you'd already chosen in step 2 under "My connections" — shown as an unchecked checkbox. That was misleading: the provider is attached regardless of step 2, so the box implied it wasn't connected, and unchecking it had no effect.

Step 3's "My connections" list now excludes provider-type connections entirely. Providers are chosen in step 2, and that's the only place they appear. The list shows only the non-provider connections (apps, MCP servers, and the like) you can optionally grant.

No change to how providers are selected in step 2 or attached to the sandbox.

Closes #2037

### DIFF
--- a/packages/ui/src/modules/sandboxes/components/connections-section.tsx
+++ b/packages/ui/src/modules/sandboxes/components/connections-section.tsx
@@ -23,6 +23,7 @@ import {
   filterOfferedTemplates,
   isShowInternalConnectionsEnabled,
 } from "../../connections/internal-only.js";
+import { excludeProviderConnections } from "../lib/provider-connections.js";
 import { CardList } from "./card-list.js";
 
 const NO_TEMPLATES: ConnectionTemplateView[] = [];
@@ -53,10 +54,8 @@ export function ConnectionsSection({
   const [showCatalog, setShowCatalog] = useState(false);
 
   const allTemplates = templatesQ.data ?? NO_TEMPLATES;
-  // Provider connections are managed in the Provider section above; keep them
-  // out of the generic list so they aren't offered twice.
-  const connections = (connectionsQ.data ?? NO_CONNECTIONS).filter(
-    (c) => !PROVIDER_TEMPLATE_IDS.has(c.templateId),
+  const connections = excludeProviderConnections(
+    connectionsQ.data ?? NO_CONNECTIONS,
   );
 
   const templateById = useMemo(
--- a/packages/ui/src/modules/sandboxes/components/steps/connections-step.tsx
+++ b/packages/ui/src/modules/sandboxes/components/steps/connections-step.tsx
@@ -1,6 +1,5 @@
 import {
   type ConnectionTemplateView,
-  type ConnectionView,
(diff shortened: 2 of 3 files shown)


jjeliga (approved):

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `bad5dc1`

## PR #2044: fix(ui): exclude provider connections from sandbox wizard step 3
**Author:** PetrBulanek | **Branch:** fix/2037-wizard-exclude-provider-connections → main | **Changes:** +13 −6 (3 files)

### Summary
Fixes #2037: the sandbox wizard's step 3 ("Grant connections") listed the provider already chosen in step 2 under "My connections" as an unchecked, no-op checkbox — misleading, since the provider is attached regardless. The provider-exclusion filter (previously inline in `connections-section.tsx`) is extracted into a shared `provider-connections.ts` helper and applied to both step 3 and the connections section, so provider-type connections are now excluded from both lists consistently. No change to how providers are selected in step 2 or attached.

### Findings
- ✅ **Looks good:** the duplicated `PROVIDER_TEMPLATE_IDS` filter is factored into one small, well-named helper (`excludeProviderConnections` / `isProviderConnection`) used by both call sites — DRY, and the behavior at the original site is preserved exactly (`packages/ui/src/modules/sandboxes/lib/provider-connections.ts`).
- ✅ **Looks good:** the change removes an unsafe `as unknown as ConnectionView[]` double-cast in `connections-step.tsx` in favor of the properly-typed helper return — a net type-safety improvement.
- ✅ **Looks good:** scope is tight and matches the issue — only the step-3 display list changes; provider selection/attachment in step 2 is untouched, as the PR body states. Human reviewer @jjeliga has approved at this commit.

### Verdict
APPROVE — Clean, well-scoped bug fix that also de-duplicates the filter and removes an unsafe cast; human approval already in place.

---

willallstet:

PR #2043, opened 2026-06-26
feat(experiments): record_run MCP tool (dam-u1n.7)

tomkis:
Stacks on #2033 (the experiments epic branch). Implements dam-u1n.7.

## What

Adds the \`record_run\` outbound MCP tool, the ingestion path for Experiments. An arm's harness calls it once per optimization-loop iteration with a score and a candidate file path; the platform captures a ledger row + the stored candidate.

Handler (orchestrated in the MCP endpoint, same shape as \`send_channel_message\`):
1. Resolve the active arm from the network-verified \`agentId\` (never from tool input).
2. Read the candidate file via the runtime \`files.read\` client.
3. Resolve the producing Trial session over ACP \`listSessions\`, matching \`_meta.platform.type=experiment_trial\` + \`experimentId\`.
4. Store the blob via the artifact store (10 MiB cap, Postgres bytea); \`candidate_ref\` = the artifact key.
5. Write the \`experiment_runs\` row (per-arm \`run_number\` allocated).

## Changes
- Contract: \`ExperimentRecordRunInput\` + \`ExperimentsService.recordRun\`.
- Repo: \`addRun\` (allocates next per-arm run number, inserts).
- Service: \`recordRun\` (ledger write + security audit line).
- Wiring: artifacts module composed once at the harness app root; experiments composed per-owner; threaded into the MCP session deps.

## Notes
- \`experiment_runs.session_id\` is resolved over ACP because sessions are agent-owned (no server-side session store). Full end-to-end needs dam-u1n.6 (start path) to stamp a real \`experiment_trial\` session.
- Download endpoint deferred to dam-u1n.10.

## Verification
- \`api-server-api\` + \`api-server\` checks (tsc, lint, prettier) pass.
- 210 \`api-server\` unit tests pass.

PR #2045, opened 2026-06-26
feat(experiments): platform-experiments skill (dam-u1n.8)

tomkis:
Ships `platform-experiments` SKILL.md in platform-base, mirroring `platform-schedules`. Picked up automatically via the Dockerfile skills COPY, symlinked into every harness.

The frontmatter description is a directive trigger keyed on the shape of the work (produce a candidate, score it, iterate) rather than the word "experiment", so a generic arm calls `record_run` even when its task prompt never mentions it. Body documents the dam-u1n.7 contract exactly: write the candidate to a workspace file, then `record_run { score, candidate: path }`; attribution automatic via verified agent id; 10 MiB cap; no shared artifact dir.

Deferred: tuning the wording against a real claude-code arm needs the dam-u1n.6 start path (in progress) plus an agent-image rebuild to have a running experiment to attribute runs to. That end-to-end pass lands with dam-u1n.11.

Stacks on feat/experiments (base), like #2043.

### DIFF
--- a/packages/platform-base/skills/platform-experiments/SKILL.md
+++ b/packages/platform-base/skills/platform-experiments/SKILL.md
@@ -1,0 +1,69 @@
+---
+name: platform-experiments
+description: >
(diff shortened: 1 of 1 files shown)


willallstet:

PR #2049, opened 2026-06-26
fix(ui): display session time last activity

### LINKED ISSUE #1999: Session list time jumps to "now" on every reload instead of showing last activity
## Problem

Each row in the sessions list shows a date and time. Users naturally read this as "when this session was last active" — the time of the last message. It isn't.

The displayed time actually advances every time you **open or reload the app** with that session in view, even when nothing has happened in the conversation. So a session whose last real message was hours ago will show a timestamp of a few seconds ago, simply because you refreshed the page.

Why it matters:
- **The timestamp is misleading.** It looks like recent activity but reflects nothing the user (or the agent) actually did.
- **Sorting is unreliable.** The list is ordered newest-first by this time, so whichever sessions you happened to reload bubble to the top — not the ones with genuine recent activity. Users can't trust the order to surface "what was I last working on."
- **It's not reproducible as a signal.** Two people looking at the same sessions can see different times depending on who reloaded last.

## Proposed solution

The time shown for each session should reflect **genuine conversation activity** — the moment of the last actual message exchanged in that session — and stay stable across reloads. Merely opening, viewing, or refreshing a session must not change its displayed time.

The list ordering should follow that same real-activity time, so the most recently *used* sessions sort to the top regardless of who reloaded what.

## Additional context

The time isn't stored as a real "last activity" value — it's read from when the session's saved conversation was last written to disk. Opening or reloading the app re-attaches to the session, and that re-attach writes small bookkeeping entries (current mode, generated title, and similar) into the saved conversation. Those writes update the file's modified time even though no message was sent, which is why the displayed time jumps forward on a plain reload.

### Scope
- The date/time shown per row in the sessions list, and the ordering derived from it.

### Out of scope
- Showing a session's live working/waiting state — tracked separately in #1081.
- Any redesign of the row layout or what other metadata is shown.

jjeliga:
CLoses #1999

### DIFF
--- a/packages/agent-runtime/src/modules/acp/services/acp-runtime.ts
+++ b/packages/agent-runtime/src/modules/acp/services/acp-runtime.ts
@@ -1224,6 +1224,9 @@ export function createAcpRuntime(deps: AcpRuntimeDeps): AcpRuntime {
       }
 
       if (promptSessionId !== null) {
+        // A real prompt is genuine activity — stamp it so the session list
+        // sorts/displays by last message, not the harness mtime.
+        deps.sessionMetadata?.recordActivity(promptSessionId);
         // Synthesize user_message_chunk(s) from the prompt payload and
         // append them to the log. The SDK drops plain-text user_message_chunk
         // emissions in live, so without this, viewers other than the sender
@@ -1399,6 +1402,7 @@ function withPlatformMeta(
   const existingMeta = isNonNullObject(session._meta) ? session._meta : {};
   return {
     ...session,
+    ...(entry.lastActivityAt ? { updatedAt: entry.lastActivityAt } : {}),
     _meta: {
       ...existingMeta,
       platform: { ...entry.meta, createdAt: entry.createdAt },
--- a/docs/architecture/persistence.md
+++ b/docs/architecture/persistence.md
@@ -1,6 +1,6 @@
 # Persistence
 
-Last verified: 2026-06-22
+Last verified: 2026-06-26
 
 ## Overview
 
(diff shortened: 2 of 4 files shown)


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `c2e978a`

## PR #2049: fix(ui): display session time last activity
**Author:** jjeliga | **Branch:** fix/1999-fix-ui-session-activity-timestamp → main | **Changes:** +80 −4 (4 files)

### Summary
Fixes #1999 so the sessions list shows the time of the last genuine message rather than the harness file mtime. Adds a `lastActivityAt` field + `recordActivity()` to the session-metadata store, stamps it on a real prompt, and surfaces it as `updatedAt` over ACP `_meta.platform`.

### Findings
- ✅ **Looks good:** `recordActivity()` is invoked only on a genuine prompt (`acp-runtime.ts:1227`), so bookkeeping writes (e.g. a `session/resume` mode update) do not bump the activity time.
- ✅ **Looks good:** `set()` preserves an existing `lastActivityAt` across rewrites (`session-metadata-store.ts:62`), and the field is optional in the schema, so older entries without it decode cleanly.
- ✅ **Looks good:** `withPlatformMeta` overrides `updatedAt` only when `lastActivityAt` is present (`acp-runtime.ts:1405`), falling back to the harness mtime otherwise — a safe, backward-compatible default.
- ✅ **Looks good:** Regression test asserts a plain reload keeps the recorded time stable (`acp-runtime.test.ts`, issue #1999 case).

_(`persistence.md` was correctly updated in the same PR — the session-metadata state-file description now mentions "the time of the session's last genuine message", and `Last verified:` was bumped to the PR date 2026-06-26.)_

### Verdict
APPROVE — Small, well-scoped fix with a targeted regression test; the activity stamp is correctly isolated to genuine prompts and the docs were updated in step.

---

PetrBulanek (approved):

packages/agent-runtime/src/modules/acp/services/acp-runtime.ts
```diff
@@ -1224,6 +1224,9 @@ export function createAcpRuntime(deps: AcpRuntimeDeps): AcpRuntime {
       }
 
       if (promptSessionId !== null) {
+        // A real prompt is genuine activity — stamp it so the session list
+        // sorts/displays by last message, not the harness mtime.
+        deps.sessionMetadata?.recordActivity(promptSessionId);
```
kapetr, line 1229: Activity is stamped on client message only, what about channel / scheduled sessions?

kapetr: Nevermind, those are triggered via client message as well. I guess only TUI activity is not handled here and it's worth confirming that we want record client activity only, because there might be significant drift between the client / agent last activity time.

kapetr (approved):

willallstet:

willallstet:
📝 Docs updated for this change: https://github.ibm.com/dam-agents/docs/pull/43

PR #2116, opened 2026-06-29
fix(experiments): arm/experiment completion (dam-u1n.13)

tomkis:
## What

Fixes dam-u1n.13: an experiment stayed `running` forever. Nothing ever wrote `completed`, there was no per-arm status, and no completion signal.

Implements the resolved design from the ticket.

## Changes

- **Per-arm status** is the source of truth for completion. New `experiment_arms.status` column with lifecycle `pending -> running -> completed | failed | stopped`, plus `last_activity_at` as the liveness clock.
- **`finish_arm` MCP tool**, the success-only dual of `record_run`. A harness calls it once when its loop is done. Skill-driven, attribution from verified agent identity (no trusted input).
- **Ledger guard**: `record_run` and `finish_arm` both reject (CONFLICT) once the calling arm is no longer `running`. `resolveActiveArm` now requires a running arm, so after Stop or completion the harness skill sees "no active arm" and stops cooperatively.
- **Rollup**: an experiment flips to `completed` once every arm is terminal, regardless of the mix. The flip is atomic (experiment row lock) so concurrent finishers / finish-vs-sweep can't double-flip. Stop sets the experiment `stopped` and cascades any still-running arm to `stopped` in one transaction.
- **Inactivity-deadline sweep** (liveness backstop): a background job reaps `running` arms that record no run and never call `finish_arm` within the window to `failed`, so a started experiment always reaches a terminal state even when a harness crashes, forgets to finish, or hibernates mid-loop. Window via `EXPERIMENT_ARM_INACTIVITY_SECONDS` (default 1h). The deadline is re-checked under the row lock, so an arm that recorded a run since listing is spared. Atomic conditional transitions make it multi-replica safe.
- **Failed trial launch** marks that arm `failed` immediately rather than waiting for the sweep.
- **UI** surfaces per-arm status as a badge.
- Glossary already updated (finish_arm, Arm Status, Experiment completion, Inactivity Deadline).

## Decisions on the ticket's open questions

- Inactivity window: one global, env-configurable, default 1h.
- Sweep: separate periodic saga in the experiments module, started at boot.
- Cooperative stop: plain rejection (resolveActiveArm -> null), no machine-readable reason code for MVP.
- finish_arm attribution mirrors record_run.
- Q7 loop-continuation (re-prompting a hibernated arm) is out of scope; termination is patched, continuation is not.
- Candidate glossary "shared artifact volume" drift is pre-existing, fixed separately.

## Tests

- Domain: terminal-status predicates.
- Service: start arm-status + per-arm launch-fail, stop cascade, record_run / finish_arm guards.
- Sweeper: deadline computation, no-op, error isolation.

`db:check:generated`, tsc/lint/format and full test suites pass across db, api-server-api, api-server, ui.


packages/api-server/src/modules/experiments/services/experiment-arm-sweeper.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+
+  return {
+    tick,
+    start() {
+      if (timer) return;
+      // Initial delay 0..intervalMs so multi-replica starts don't stack.
+      const jitter = Math.floor(Math.random() * deps.intervalMs);
+      timer = setTimeout(() => {
```
dam-code-guardian, line 98: 🟢 **Suggestion:** Tiny lifecycle race: during the `jitter` window `timer` holds the `setTimeout` handle; the timeout callback then reassigns `timer = setInterval(...)`. If `stop()` runs in the instant the timeout fires but before that reassignment, `stop()` clears the already-fired timeout and nulls `timer`, then the callback installs a fresh interval that escapes cancellation. Harmless here (boot singleton, `unref`'d), but a `stopped` flag checked at the top of the callback would close it cleanly.

packages/api-server/src/index.ts
```diff
@@ -448,6 +449,20 @@ const agentArtifactsSweeper = createAgentArtifactsSweeper({
...
+// Inactivity-deadline sweep: reaps `running` Experiment arms that have gone
+// quiet (no Run, no finish_arm) past the configured window to `failed`, so a
+// started Experiment always reaches a terminal state. Atomic conditional
+// transitions make it multi-replica safe. Sweep at the window cadence, capped
+// at 5 minutes so the default 1h window isn't scanned needlessly often.
+const armInactivityMs = config.experimentArmInactivitySeconds * 1000;
+const experimentArmSweeper = composeExperimentArmSweeper({
+  db,
+  inactivityMs: armInactivityMs,
+  intervalMs: Math.min(armInactivityMs, 5 * 60_000),
```
dam-code-guardian, line 461: 🟢 **Suggestion:** With the default 1h inactivity window this makes the sweep scan every 5 min (the `Math.min` ceiling), not hourly. Worth a one-line comment that the 5-min cap — not the inactivity window — drives scan cadence, so the decoupling is obvious to the next reader.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `40f7708`

## PR #2116: fix(experiments): arm/experiment completion (dam-u1n.13)
**Author:** tomkis | **Branch:** fix/experiments-arm-completion → feat/experiments | **Changes:** +2756 −44 (21 files)

### Summary
Fixes an experiment that stayed `running` forever by making per-arm status the source of truth for completion. Adds an `experiment_arms.status` lifecycle (`pending → running → completed|failed|stopped`) + `last_activity_at` liveness clock, a `finish_arm` MCP tool (success-only dual of `record_run`), CONFLICT ledger guards on `record_run`/`finish_arm` once an arm leaves `running`, an atomic experiment-completion rollup under a row lock, a cascade-on-stop, and an inactivity-deadline background sweeper that reaps quiet arms to `failed` so an experiment always reaches a terminal state. UI gains a per-arm status badge. (No prior review on file — the previous in-progress row was a crashed run that never posted.)

### Findings
- 🟢 **Suggestion:** Sweeper `start()` has a tiny lifecycle race: during the initial `jitter` window `timer` holds the `setTimeout` handle, and the timeout callback later reassigns `timer = setInterval(...)`. If `stop()` runs in the instant the timeout fires but before that reassignment, `stop()` clears the (already-fired) timeout and nulls `timer`, then the callback installs a fresh interval that escapes cancellation. Harmless in practice (boot-time singleton, `unref`'d), but a `stopped` flag checked inside the callback would close it. (`packages/api-server/src/modules/experiments/services/experiment-arm-sweeper.ts:98`)
- 🟢 **Suggestion:** `intervalMs: Math.min(armInactivityMs, 5 * 60_000)` means with the default 1h window the sweep runs every 5 min — reasonable, but worth a one-line comment that the 5-min ceiling (not the inactivity window) drives scan cadence, so a future reader doesn't expect hourly scans. (`packages/api-server/src/index.ts:461`)
- ✅ **Looks good:** Every terminal transition goes through `transitionRunningArm`, which locks the experiment row first then the arm row — one consistent lock order across finish / sweep / launch-fail, so the paths can't deadlock. `completeIfAllTerminal` is gated on the experiment still being `running`, so a raced-in Stop is never overwritten back to `completed`.
- ✅ **Looks good:** The sweep re-checks the deadline under the row lock via `extraArmGuard` (`lastActivityAt < deadline`), so an arm that recorded a Run between listing and reaping is correctly spared — the read-then-reap race is closed.
- ✅ **Looks good:** Ledger guards return `null` from `addRun`/`finishArm` when the arm is no longer `running`, surfaced as CONFLICT and as `resolveActiveArm → null` so a stopped/finished harness cooperatively stops. `finish_arm` MCP description is precise about call-once / no-more-runs semantics.
- ✅ **Looks good:** The partial index `WHERE status = 'running'` exactly covers the sweep's scan set; `start()` re-reads settled status after `launchTrials` so an all-launch-fail experiment reports `completed` rather than a stale `running`. Restarting a `stopped` experiment is safe — an all-terminal experiment is already `completed` (unstoppable), so a stopped one always has a resumable arm.
- ✅ **Looks good:** Tests cover domain predicates, start arm-status, per-arm launch-fail, stop cascade, both ledger guards, and sweeper deadline/no-op/error-isolation.

### Verdict
APPROVE — Correct, well-tested completion model with sound multi-replica concurrency (consistent lock order, atomic conditional transitions, deadline re-check under lock). Only two non-blocking nits.

---

willallstet:

PR #2118, opened 2026-06-29
feat(agents): per-agent hibernation timeout override (#1373)

### LINKED ISSUE #1373: Keep an agent awake while it has long-running background work (don't hibernate mid-job)

jjeliga:
Closes #1373.

Adds a per-agent idle-timeout override so workloads with no open ACP session can be kept awake. Until now every agent inherited the chart-wide idle timeout, and the idle checker hibernated it regardless of background work.

## What
- New `spec.hibernationTimeout` (Go duration) on the Agent CR: `0s` = never hibernate, omitted = inherit the chart-wide default.
- Controller honours the per-agent value in the idle checker / hibernation path.
- api-server resolves the effective timeout (override, else global default) and exposes it as `effectiveHibernationTimeoutMin`; conversion between the duration wire format and the minutes UI unit lives in `duration.ts`.
- UI: a "Lifecycle" timeout field (in minutes) in sandbox settings, with a confirm prompt when crossing the never-hibernate boundary.

## Deploy note
The CRD carries `helm.sh/resource-policy: keep`, so `helm upgrade` will **not** update it — re-apply the CRD by hand when deploying, otherwise the apiserver prunes the new field and the override silently reverts to the default.


dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `f32de32`

## PR #2118: feat(agents): per-agent hibernation timeout override (#1373)
**Author:** jjeliga | **Branch:** feat/1373-keepawake-toggle → main | **Changes:** +278 −12 (27 files)

### Summary
Adds an optional per-agent `spec.hibernationTimeout` (Go duration) so workloads with no open ACP session can be kept awake, overriding the chart-wide idle timeout (`0s` = never hibernate, omitted = inherit default). The override is resolved to an effective value in both the controller (idle checker + reconciler `shouldRun` gate) and the api-server (`effectiveHibernationTimeoutMin`), and surfaced in the UI as a minutes field with a confirm prompt when crossing the never-hibernate boundary.

### Findings
- 🟡 **Warning:** `durationToMinutes` returns `0` (the never-hibernate sentinel) for any unparseable or unknown-unit input, so a misconfigured chart-wide `AGENT_IDLE_TIMEOUT` would silently disable hibernation cluster-wide rather than failing loudly (`packages/api-server/src/duration.ts:6`).
- 🟢 **Suggestion:** the CRD ships `helm.sh/resource-policy: keep`, so `helm upgrade` won't add the new `hibernationTimeout` field — the operator must re-apply the CRD by hand or the apiserver prunes the field and the override silently reverts. This is called out in the PR body; worth mirroring in the chart's deploy docs/release notes so it isn't lost (`deploy/helm/platform/crds/agent-platform.ai_agents.yaml:18`).
- ✅ **Looks good:** controller, api-server, and UI all resolve the *effective* value (override else global default) through one shared helper per layer, so scale-up and scale-down share the same `shouldRun` gate and can't disagree.
- ✅ **Looks good:** `shouldRun`/`effectiveIdleTimeout` fail open — a `nil`/absent/unparseable override inherits the global default, and the idle checker's `hibernationOverride` returns `nil` on parse error, so a bad value never accidentally pins or kills a pod.
- ✅ **Looks good:** the update path correctly distinguishes `undefined` (leave unchanged) from `null` (merge-patch delete → inherit default) for `hibernationTimeoutMin`.
- ✅ **Looks good:** CRD schema-generation bumped 2→3 with a matching `zz_generated.deepcopy.go` entry for the new `*metav1.Duration` pointer field.

**Verdict: minor drift.**

**Drift (Check #2 — `Last verified:` staleness):**
- ❌ `docs/architecture/agent-lifecycle.md:3` — the page body was substantively edited in this PR (the entire **Hibernate** section was rewritten to document the new per-agent timeout, the two-check idle derivation, and the trade-offs), but the header still reads `Last verified: 2026-06-25`. The guidelines require bumping `Last verified:` to the PR date whenever the page is edited. Proposed: bump to `Last verified: 2026-06-29`.

The architecture-page drift check (#1) is otherwise **satisfied** — the Hibernate section was updated in the same PR to reflect the new behavior. No ADR-reference leaks (#3), no volatile-content additions (#4), no new-subsystem-without-page (#5), and no cross-reference rot (#6).

### Verdict
COMMENT — Solid, well-layered feature with consistent fail-open semantics; address the silent-`0`-on-unparseable-duration warning and bump the doc's `Last verified:` date.

---

packages/api-server/src/duration.ts
```diff
@@ -0,0 +1,17 @@
+// Conversions between Go duration strings (the CR / chart wire format) and whole minutes (the UI/API unit).
+
+// Parse a Go duration string ("1h", "30m", "0s") to whole minutes. Any positive duration rounds up to at least 1, so it is never mistaken for 0 (= never hibernate).
+export function durationToMinutes(d: string): number {
+  let ms = 0;
+  for (const [, n, unit] of d.matchAll(/(\d+(?:\.\d+)?)(ms|h|m|s)/g)) {
```
dam-code-guardian: 🟡 **Warning:** `durationToMinutes` silently ignores any token whose unit isn't `ms|h|m|s`. A malformed or unrecognized-unit input (e.g. `"2d"`, a typo, or an empty string) matches nothing, so `ms` stays `0` and the function returns `0` — which is the **never-hibernate** sentinel. Because this parses the chart-wide `AGENT_IDLE_TIMEOUT` (`config.ts`), a misconfigured value would silently disable hibernation cluster-wide instead of failing loudly. Consider validating that the whole string was consumed (or that at least one token matched) and throwing / falling back to the documented `"1h"` default on an unparseable value, so `0` only ever comes from an explicit `"0s"`.

jjeliga: api-server doesn't enforce hibernation. agentIdleTimeoutMinutes flows only into effectiveHibernationTimeoutMin ([agent-mappers.ts:148-150](vscode-webview://09bqs6qrapo0k666sfrs8og23r5r2mloslkkson419v2hml73hl8/packages/api-server/src/modules/agents/infrastructure/agent-mappers.ts#L148-L150)) — a display/read value the UI shows. Nothing in api-server acts on it.
The controller enforces it, and it already fails loud on the same value. Both read the one Helm value controller.agent.base.idleTimeout ([app.yaml:219-220](vscode-webview://09bqs6qrapo0k666sfrs8og23r5r2mloslkkson419v2hml73hl8/deploy/helm/platform/templates/apiserver/app.yaml#L219-L220)). The controller parses it with Go's strict time.ParseDuration; on malformed input [agent_config.go:24-27](vscode-webview://09bqs6qrapo0k666sfrs8og23r5r2mloslkkson419v2hml73hl8/packages/controller/pkg/config/agent_config.go#L24-L27) returns an error → config load fails → crashloop. So a typo'd AGENT_IDLE_TIMEOUT produces a loud, obvious cluster failure, not a silent one.
The real worst case from api-server's silent-0 is a misleading "Never hibernates" label in the UI — for a value the controller has already rejected, so the cluster is visibly broken regardless.
And the proposed-sounding fix (make durationToMinutes throw) is actively risky: it's also called on every agent's spec.hibernationTimeout during list/get. Those values come from minutesToDuration so they're always well-formed — but a hand-edited bad CR value would then crash agent listing, which is worse than a display 0.

The kernel of truth: api-server failing silently on its own malformed config is a smell. The safe, proportionate fix is to validate AGENT_IDLE_TIMEOUT at the config boundary (fail loud at startup, mirroring the controller) — not to touch the shared parser. That's ~5 lines in config.ts and addresses the valid part without the crash risk.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `fc07e4a`

## PR #2118: feat(agents): per-agent hibernation timeout override (#1373)
**Author:** jjeliga | **Branch:** feat/1373-keepawake-toggle → main | **Changes:** +290 −13 (29 files)

### Summary
Adds an optional per-agent `spec.hibernationTimeout` (Go duration; `0s` = never hibernate, omitted = inherit the chart-wide default), resolved to an effective value in the controller, api-server, and UI. This re-review covers the changes pushed since the prior review at `f32de32` — both prior findings are now resolved. _(Suppressed 1 finding per PR context: the `durationToMinutes`-should-throw warning — author confirmed `effectiveHibernationTimeoutMin` is display-only and the controller already fails loud on a bad `AGENT_IDLE_TIMEOUT`; the proportionate config-boundary fix was applied.)_

### Changes since last review
Previous HEAD: `f32de32` (2026-06-29T12:36:09Z) — verdict COMMENT

- ✅ **Fixed:** silent-`0`-on-unparseable-duration — new `durationToMinutesStrict` throws on malformed input and is wired into `config.ts` at the `AGENT_IDLE_TIMEOUT` boundary, so a bad chart value now fails loudly at startup instead of silently disabling hibernation (`packages/api-server/src/duration.ts:18`, `packages/api-server/src/config.ts:199`). The shared `durationToMinutes` parser is untouched, avoiding the list/get crash risk the author flagged.
- ✅ **Fixed:** stale doc `Last verified:` — `docs/architecture/agent-lifecycle.md` bumped `2026-06-25` → `2026-06-29` (matches PR date).

### Findings
- ✅ **Looks good:** the config-boundary validation (`durationToMinutesStrict`) is the right scope — it fails loud on operator misconfig without touching the parser used on every agent's well-formed `spec.hibernationTimeout` during list/get.

**Verdict: aligned.**

✅ No documentation drift detected. The Hibernate section of `docs/architecture/agent-lifecycle.md` was updated in this PR to document the new per-agent timeout, the two-check idle derivation, and the trade-offs, and the `Last verified:` header was bumped to the PR date (`2026-06-29`). No ADR-reference leaks, no volatile-content additions, no new-subsystem-without-page, no cross-reference rot.

### Verdict
APPROVE — Both prior findings resolved exactly as discussed; the remaining warning was a display-only path the author correctly rebutted and addressed at the config boundary.

---

packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-form.ts
```diff
@@ -54,6 +55,10 @@ const settingsSchema = z.object({
   envVars: z
     .array(envVarSchema)
     .refine(allEnvVarsValid, "All env vars need a name and a value"),
+  hibernationTimeoutMin: z
+    .number({ message: "Enter a number of minutes (0 = never)" })
+    .int()
+    .nonnegative(),
```
kapetr, line 61: Small thing, but there's now no way to clear the override and go back to the inherited value.

kapetr (approved):

packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-form.ts
jjeliga: Not exactly true, you can see that there is a 60 minute default hwen setting a different value. AS a user, you do not really know there is some platform default, that is behind the scenes

willallstet:
📝 Docs updated for this change: https://github.ibm.com/dam-agents/docs/pull/35

jezekra1 (approved):

PR #2121, opened 2026-06-29
feat: per-user CPU & memory budgeting

### LINKED ISSUE #1900: Per-user CPU and memory limits
## Problem
Today nothing caps how much compute a single user can hold at once. Each agent has its own idle-based hibernation, but that's a per-agent helper — it does nothing to stop one user from spinning up many agents, or pinning them all awake, and starving everyone else.

With only a handful of users this is invisible. As soon as the platform has more people on it, it becomes a real contention problem: there's a fixed pool of CPU and memory, and the first users to grab it can leave the rest with nothing. We currently have no way to say "this user gets at most this much."

## Proposed solution
Give each user a resource budget — a ceiling on the CPU and memory their running agents can consume at any one moment. Within that ceiling a user is free: deploy whatever they want, for as long as they want. They just can't exceed their share of the pool concurrently.

This reframes automatic hibernation as a convenience rather than the enforcement mechanism — a way to help users stay under their own limit by freeing idle agents, not the thing standing between users and a shared resource free-for-all. To live within a budget, users also need a clear way to see what they're currently consuming and to stop agents to free room.

## Scope
A per-user ceiling on concurrent CPU and memory across that user's running agents.
Visibility into current consumption against the limit.
The ability for a user to stop their own agents to free capacity.

## Out of scope
Role-based limits and approval flows — different budgets per role, or "request more" requests an admin approves/denies, can come later once the basic ceiling exists.
Token/cost usage — this is about live CPU/memory contention, not billing or usage accounting.

jjeliga:
Closes #1900.

## What
A per-user ceiling on concurrent CPU/memory across a user's running agents, visibility into their consumption, and a self-service way to free capacity.

## How
- **Budget model** — a user's footprint is the summed pod *requests* (CPU + memory) of their non-hibernated agents plus a fixed gateway constant. Requests are the scheduler's real reservation and the only value knowable before a pod is scheduled (there's no live-metrics API). Ceiling = chart default, overridable per-user via a new `user_budgets` table.
- **Enforcement** — the api-server rejects before any spec write on every path that brings an agent online: **create** (service layer) and **all wakes** (the `AgentsRepository` chokepoint, so ACP/terminal/SSH relays, file proxies, and skills inherit it). Over-budget surfaces a `FORBIDDEN` naming reserved-vs-limit.
- **Scheduled fire over budget** — blocked and deferred: the trigger is already committed to the outbox, so it redelivers once the user frees room or expires at its TTL.
- **Free capacity** — a user-initiated hard stop on a running agent. A new controller `stop-requested` annotation scales the pair to zero immediately; any later activity clears it.
- **Visibility** — a `budgets.usage` query rendered as a meter on the sandbox list.

## Out of scope (per the issue)
Role-based limits / approval flows, and token/cost usage.

See `docs/architecture/budgets.md` for the full design.

### DIFF
--- a/deploy/helm/platform/templates/apiserver/app.yaml
+++ b/deploy/helm/platform/templates/apiserver/app.yaml
@@ -277,6 +277,15 @@ spec:
             - name: MAX_IMPORT_BUNDLE_BYTES
               value: {{ .Values.apiServer.maxImportBundleBytes | quote }}
             {{- end }}
+            # Per-user compute ceiling (#1900); AGENT_DEFAULT_* mirrors the controller templateDefaults.
+            - name: DEFAULT_USER_CPU_BUDGET
+              value: {{ .Values.apiServer.defaultUserCpuBudget | quote }}
+            - name: DEFAULT_USER_MEMORY_BUDGET
+              value: {{ .Values.apiServer.defaultUserMemoryBudget | quote }}
+            - name: AGENT_DEFAULT_CPU_REQUEST
+              value: {{ .Values.controller.agent.templateDefaults.resources.requests.cpu | quote }}
+            - name: AGENT_DEFAULT_MEMORY_REQUEST
+              value: {{ .Values.controller.agent.templateDefaults.resources.requests.memory | quote }}
             - name: TRUSTED_HOSTS_PATH
               value: /etc/platform/trusted-hosts
             # agent templates are mounted from a ConfigMap and loaded
--- a/packages/api-server/src/modules/budgets/domain/resources.ts
+++ b/packages/api-server/src/modules/budgets/domain/resources.ts
@@ -1,0 +1,86 @@
+/** CPU (millicores) and memory (bytes) — the two budgeted dimensions. */
+export interface ResourceAmount {
+  cpuMilli: number;
+  memoryBytes: number;
+}
+
+/** Raw K8s request strings off an agent's `spec.resources.requests`. */
+export interface ResourceRequests {
+  cpu?: string;
+  memory?: string;
+}
+
+export const ZERO: ResourceAmount = { cpuMilli: 0, memoryBytes: 0 };
+
+/** Fixed per-agent gateway (Envoy) requests; added as a constant since the api-server never renders it. */
+export const GATEWAY: ResourceAmount = {
+  cpuMilli: 50,
+  memoryBytes: 64 * 1024 * 1024,
+};
+
+const MEM_UNITS: Record<string, number> = {
+  Ki: 1024,
+  Mi: 1024 ** 2,
+  Gi: 1024 ** 3,
+  Ti: 1024 ** 4,
+  k: 1e3,
+  M: 1e6,
+  G: 1e9,
+  T: 1e12,
+};
+
+/** "250m" → 250, "1" → 1000; unparseable → 0. */
+export function parseCpuMilli(q: string): number {
+  const s = q.trim();
+  const n = s.endsWith("m") ? Number(s.slice(0, -1)) : Number(s) * 1000;
+  return Number.isFinite(n) ? Math.round(n) : 0;
+}
+
+/** "512Mi" → 536870912, "8Gi" → …, plain bytes when no unit. */
+export function parseMemoryBytes(q: string): number {
+  const m = q.trim().match(/^(\d+(?:\.\d+)?)([A-Za-z]+)?$/);
+  if (!m) return 0;
+  const factor = m[2] ? (MEM_UNITS[m[2]] ?? 1) : 1;
+  return Math.round(Number(m[1]) * factor);
+}
+
+export function parseAmount(cpu: string, memory: string): ResourceAmount {
+  return {
+    cpuMilli: parseCpuMilli(cpu),
+    memoryBytes: parseMemoryBytes(memory),
+  };
+}
+
+export function add(a: ResourceAmount, b: ResourceAmount): ResourceAmount {
+  return {
+    cpuMilli: a.cpuMilli + b.cpuMilli,
+    memoryBytes: a.memoryBytes + b.memoryBytes,
+  };
+}
+
+/** An agent's reserved footprint: its requests (or chart default) plus the gateway. */
+export function footprint(
+  requests: ResourceRequests | undefined,
+  agentDefault: ResourceAmount,
+): ResourceAmount {
+  return add(GATEWAY, {
+    cpuMilli: requests?.cpu
+      ? parseCpuMilli(requests.cpu)
+      : agentDefault.cpuMilli,
+    memoryBytes: requests?.memory
+      ? parseMemoryBytes(requests.memory)
+      : agentDefault.memoryBytes,
+  });
+}
+
+/** True when adding `candidate` to `reserved` would breach `limit` on either dimension. */
+export function wouldExceed(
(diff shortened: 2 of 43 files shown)


packages/api-server/src/modules/budgets/domain/resources.ts
```diff
@@ -0,0 +1,86 @@
...
+/** Raw K8s request strings off an agent's `spec.resources.requests`. */
+export interface ResourceRequests {
+  cpu?: string;
+  memory?: string;
+}
+
+export const ZERO: ResourceAmount = { cpuMilli: 0, memoryBytes: 0 };
+
+/** Fixed per-agent gateway (Envoy) requests; added as a constant since the api-server never renders it. */
+export const GATEWAY: ResourceAmount = {
```
dam-code-guardian, line 16: 🟡 **Warning:** `GATEWAY` is hard-coded here (`50m` / `64Mi`), but the per-agent default is templated from the same Helm value the controller uses precisely so the two "cannot drift" (per `budgets.md`). The gateway figure has no such guard — if the chart's Envoy pod requests change, this constant silently diverges and every user's footprint is mis-accounted. Consider surfacing it from the chart the same way as `agentDefault`.

packages/api-server/src/modules/budgets/domain/resources.ts
```diff
@@ -0,0 +1,86 @@
...
+  M: 1e6,
+  G: 1e9,
+  T: 1e12,
+};
+
+/** "250m" → 250, "1" → 1000; unparseable → 0. */
+export function parseCpuMilli(q: string): number {
+  const s = q.trim();
+  const n = s.endsWith("m") ? Number(s.slice(0, -1)) : Number(s) * 1000;
+  return Number.isFinite(n) ? Math.round(n) : 0;
```
dam-code-guardian, line 36: 🟡 **Warning:** Fail-open to `0` on unparseable input (same in `parseMemoryBytes` at :42). When this parses an *agent's* `requests`, a malformed value makes that agent contribute only the gateway constant toward its owner's budget — silently undercounting and letting it escape the ceiling. On a quota boundary, prefer failing toward the chart default (and logging) rather than 0.

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `7bd4579`

## PR #2121: feat: per-user CPU & memory budgeting
**Author:** jjeliga | **Branch:** feat/1900-usage-monitoring-and-limiting → main | **Changes:** +2434 −68 (43 files)

### Summary
Adds a per-user ceiling on concurrent CPU/memory across a user's running agents: a budget guard in the api-server rejects creates and wakes that would breach the owner's reserved-request footprint (summed pod requests + a fixed gateway constant), a `user_budgets` table for per-user overrides, a controller `stop-requested` hard-stop annotation to free capacity, and a reserved-vs-limit meter on the sandbox list. Enforcement is centralized at the `AgentsRepository` wake chokepoint so relays/proxies/skills inherit it. Design is documented in `docs/architecture/budgets.md`.

### Findings
- 🟡 **Warning:** `GATEWAY` is a hard-coded constant (`50m` / `64Mi`) while the per-agent default is templated from the same Helm value the controller uses (so it "cannot drift"). The gateway figure has no such guard — if the chart's Envoy pod requests change, this constant silently diverges and every user's footprint is mis-accounted. Consider surfacing it from the chart the same way as the agent default. (`packages/api-server/src/modules/budgets/domain/resources.ts:16`)
- 🟡 **Warning:** `parseCpuMilli` / `parseMemoryBytes` fail open to `0` on unparseable input. For an agent's own `requests`, a malformed value makes that agent contribute only the gateway constant toward its owner's budget — silently undercounting and letting it slip under the ceiling. Specs are normally chart-controlled, but a fail-*open* default on a security/quota boundary is worth a deliberate decision (log + treat as the chart default, rather than 0). (`packages/api-server/src/modules/budgets/domain/resources.ts:36`, `:42`)
- 🟢 **Suggestion:** `parseMemoryBytes` rejects exponent notation (e.g. `1e6` → `0`), which K8s quantities technically allow. Rare for memory requests, but the same fail-open-to-0 caveat applies. (`packages/api-server/src/modules/budgets/domain/resources.ts:40`)
- 🟢 **Suggestion:** The budget meter polls `budgets.usage` every 5s unconditionally while the sandbox list is mounted (`refetchInterval: 5000`). Fine for a live meter; flagging only so it's a conscious choice. (`packages/ui/src/modules/budgets/api/queries.ts:9`)
- ✅ **Looks good:** Centralizing wake enforcement at the `AgentsRepository` chokepoint (with a late-bound guard to break the repo↔budget cycle) so ACP/terminal/SSH relays, file proxies, and skills all inherit the gate is the right seam — far less likely to be bypassed than per-call-site checks.
- ✅ **Looks good:** Reserved consumption is computed on demand from the live agent list and never persisted, so hibernate/delete credit the budget back with nothing to reconcile. The read-decide-write race is explicitly acknowledged and bounded in `budgets.md` (transient overshoot, reclaimed by hibernation; advisory-lock upgrade path noted) — not re-flagged here.
- ✅ **Looks good:** `docs/architecture/budgets.md` is a thorough, correctly-scoped design page, linked from the landing page, and `agent-lifecycle.md` documents the `stop-requested` hard-stop mechanics in lockstep with the controller change.

**Verdict: minor drift**

**Drift (Check 2 — `Last verified:` staleness)**
- ❌ `docs/architecture.md:3` — the landing page was edited in this PR (added the `budgets` subsystem entry) but its header still reads `Last verified: 2026-06-12`. Every architecture page edited in the diff must bump `Last verified:` to the PR date (2026-06-29).

budgets.md (new, linked, `Last verified: 2026-06-29`) and agent-lifecycle.md (`Last verified:` bumped 06-25 → 06-29, documents the hard stop) are in sync. No ADR-reference or volatile-content leaks.

### Verdict
COMMENT — Solid, well-documented feature with a sound enforcement seam; address the GATEWAY-drift and parse fail-open warnings and bump `docs/architecture.md`'s `Last verified:` before merge.

---

deploy/helm/platform/templates/apiserver/app.yaml
```diff
@@ -277,6 +277,15 @@ spec:
             - name: MAX_IMPORT_BUNDLE_BYTES
               value: {{ .Values.apiServer.maxImportBundleBytes | quote }}
             {{- end }}
+            # Per-user compute ceiling (#1900); AGENT_DEFAULT_* mirrors the controller templateDefaults.
+            - name: DEFAULT_USER_CPU_BUDGET
+              value: {{ .Values.apiServer.defaultUserCpuBudget | quote }}
+            - name: DEFAULT_USER_MEMORY_BUDGET
+              value: {{ .Values.apiServer.defaultUserMemoryBudget | quote }}
+            - name: AGENT_DEFAULT_CPU_REQUEST
+              value: {{ .Values.controller.agent.templateDefaults.resources.requests.cpu | quote }}
```
jezekra1, line 286: Maybe the controller should just materialize the defaults into the CR. 

It feels weird that api server is reading value from `.Values.controller.agent.templateDefaults`

JanPokorny:
@jjeliga Looks like this drifted quite a bit, can you take a look?

jezekra1:
@JanPokorny jjeliga is on a longer vacation, I'll take over this

PR #2126, opened 2026-06-30
fix(skills): degrade skills.state when the agent pod is unreachable

pilartomas:
## Problem

`skills.state` returns **HTTP 500** when an agent's configuration is changed while it's running (and on other pod-disrupting actions). The Skills panel breaks until the pod settles.

Repro: agent running fine → change env/secret configuration (or restart / wake) → `GET /api/trpc/skills.state?...` returns 500.

## Root cause

`skills.state` → `SkillsService.getState()` reads `"running"` from `computeAgentState()`, which is derived from the controller's **cached `Ready` condition** (`Ready = AgentPodReady ∧ GatewayPodReady`) — not a live probe. In the `"running"` branch it then calls `runtimeClient.listLocal(agentId)` — a tRPC call into the agent pod — **with no error handling**.

When the pod rolls or starts (a `secretRef` change patches the CR → pod roll; also `agent.restart`, wake-from-hibernate, and the normal startup window), the in-pod tRPC listener is briefly unreachable *while `Ready` is still cached `True`*. The connection error is rethrown by the skills runtime client as a plain `Error` (only `AgentRuntimeUpstreamError` was special-cased), `getState` doesn't catch it, and there is no tRPC `errorFormatter` — so the default `INTERNAL_SERVER_ERROR` (HTTP 500) is returned.

`getState`'s **not-running** branch already degrades gracefully (returns tracked refs, no reconcile); the **running** branch just never anticipated a "running-but-unreachable" pod.

## Fix

- Add `AgentRuntimeUnreachableError`, thrown by the skills runtime client when a tRPC call fails with **no error envelope** (transport failure: connection refused / timeout / bare proxy 5xx) — distinct from an upstream-gateway error or a genuine application error from the pod.
- `getState` and `listLocal` catch **only** that error and fall back to the tracked-only view (no reconcile, empty `standalone`), the same behavior as the not-running branch. This avoids dropping `agent_skills` rows when the disk isn't observable.
- Genuine failures — including `reconcile`/`listSkills` DB errors — still propagate as 500 (no over-catching).

## Notes / scope

- `skills.list()` (private-source scan) shares the underlying client but is an explicit user action that already `ensureAgentReachable`-wakes the pod first; left as-is here to keep the change focused on the render-path queries. Its behavior is unchanged (a transport failure still surfaces as an error rather than silently degrading).
- A broader sweep for the same class of bug in other render-time queries that call into the pod (files panel, connections, relays) was **not** completed and is worth a follow-up.

## Testing

- New unit tests (`skills-service-state-unreachable.test.ts`): `getState` falls back to tracked refs without calling `reconcile` on unreachable; a non-unreachable error still throws; `listLocal` returns `[]` on unreachable.
- `mise run api-server:check` (tsc + lint + format) and full `mise run api-server:test` (218 tests) pass.

### DIFF
--- a/packages/api-server/src/modules/skills/infrastructure/agent-runtime-client.ts
+++ b/packages/api-server/src/modules/skills/infrastructure/agent-runtime-client.ts
@@ -51,6 +51,20 @@ export class AgentRuntimeUpstreamError extends Error {
   }
 }
 
+/**
+ * The request never reached a responding agent-runtime tRPC server —
+ * connection refused, timeout, DNS, or a bare proxy 5xx while the pod is
+ * rolling/starting. The controller's cached Ready condition can still read
+ * True during that window, so read-side callers (state/listLocal) catch this
+ * to degrade gracefully instead of 500ing on a transient.
+ */
+export class AgentRuntimeUnreachableError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = "AgentRuntimeUnreachableError";
+  }
+}
+
 // Auth on the api-server → agent-runtime hop is enforced at the kernel by
 // the agent pod's NetworkPolicy (ingress admitted only from the api-server
 // pod). No Bearer header is sent.
@@ -91,6 +105,11 @@ async function runWithUpstreamMapping<T>(
       const upstream = data?.upstream;
       if (isUpstreamGatewayError(upstream)) {
         throw new AgentRuntimeUpstreamError(`${label}: ${e.message}`, upstream);
+      }
+      // No error envelope means no agent-runtime tRPC server responded — a
+      // transport failure, not an application error from the pod.
+      if (data === null) {
+        throw new AgentRuntimeUnreachableError(`${label}: ${e.message}`);
       }
       throw new Error(`${label}: ${e.message}`);
     }
--- a/packages/api-server/src/__tests__/unit/skills-service-state-unreachable.test.ts
+++ b/packages/api-server/src/__tests__/unit/skills-service-state-unreachable.test.ts
@@ -1,0 +1,85 @@
+import { describe, it, expect, vi } from "vitest";
+import type { LocalSkill, SkillRef } from "api-server-api";
+import { createSkillsService } from "../../modules/skills/services/skills-service.js";
+import type { SkillsServiceDeps } from "../../modules/skills/services/skills-service.js";
+import {
+  AgentRuntimeUnreachableError,
+  AgentRuntimeUpstreamError,
+} from "../../modules/skills/infrastructure/agent-runtime-client.js";
+
+const AGENT = "agent-1";
+const OWNER = "owner-1";
+
+const TRACKED: SkillRef[] = [
+  { source: "https://github.com/acme/skills", name: "deploy", version: "abc" },
+];
+
+/** Minimal running InfraAgent — computeAgentState reads only error/ready/
+ *  hibernated, so the rest is irrelevant here. */
+function runningInfra() {
+  return { id: AGENT, name: AGENT, spec: {}, ready: true, hibernated: false };
+}
+
+function makeDeps(listLocal: () => Promise<LocalSkill[]>) {
+  const reconcile = vi.fn(async () => {});
+  const agentsRepo = {
+    async get(id: string) {
+      return id === AGENT ? runningInfra() : null;
+    },
+  };
+  const agentSkillsRepo = {
+    async listSkills() {
+      return TRACKED;
+    },
+    async listPublishes() {
+      return [];
+    },
+    reconcile,
+  };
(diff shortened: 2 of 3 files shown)


packages/api-server/src/modules/skills/infrastructure/agent-runtime-client.ts
```diff
@@ -92,6 +106,11 @@ async function runWithUpstreamMapping<T>(
       if (isUpstreamGatewayError(upstream)) {
         throw new AgentRuntimeUpstreamError(`${label}: ${e.message}`, upstream);
       }
+      // No error envelope means no agent-runtime tRPC server responded — a
+      // transport failure, not an application error from the pod.
+      if (data === null) {
```
dam-code-guardian, line 111: 🟢 **Suggestion:** This `data === null` ⇒ `AgentRuntimeUnreachableError` branch is the new classification heuristic, but it has no direct unit test — the added tests inject an already-classified `AgentRuntimeUnreachableError` into the service, exercising the *handling* path, not this *classification* path. Consider a small test over `runWithUpstreamMapping`: a `TRPCClientError` with `data === null` maps to `AgentRuntimeUnreachableError`, with a `{ upstream }` envelope maps to `AgentRuntimeUpstreamError`, and with a non-null non-upstream envelope maps to a plain `Error`. That locks the branch most likely to regress.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `f10bf46`

## PR #2126: fix(skills): degrade skills.state when the agent pod is unreachable
**Author:** pilartomas | **Branch:** fix/skills-state-unreachable-500 → main | **Changes:** +135 −7 (3 files)

### Summary
Fixes a HTTP 500 on `skills.state` when an agent's `Ready` condition is cached `True` but its in-pod tRPC listener is briefly unreachable (pod rolling/starting after a config change, restart, or wake). Adds `AgentRuntimeUnreachableError` (raised when a tRPC call fails with no error envelope — a transport failure) and has `getState`/`listLocal` catch **only** that error to fall back to the tracked-only view, mirroring the existing not-running branch. Genuine errors (DB failures, upstream-gateway errors) still propagate as 500.

### Findings
- 🟢 **Suggestion:** The new transport-failure classification itself — `data === null` ⇒ `AgentRuntimeUnreachableError` inside `runWithUpstreamMapping` — has no direct unit test. The added tests (`skills-service-state-unreachable.test.ts`) inject an already-classified `AgentRuntimeUnreachableError` into the service, so they exercise the *handling* but not the *classification* heuristic. A small test over `runWithUpstreamMapping` (TRPCClientError with `data===null` ⇒ Unreachable; with a `{upstream}` envelope ⇒ Upstream; with a non-null non-upstream envelope ⇒ plain `Error`) would lock the branch that's actually most likely to regress. (`packages/api-server/src/modules/skills/infrastructure/agent-runtime-client.ts:111`)
- ✅ **Looks good:** Error-class separation is precise — only `AgentRuntimeUnreachableError` is caught; everything else (including upstream-gateway and DB errors) rethrows, so there's no over-catching that would mask real failures.
- ✅ **Looks good:** The `trackedOnly()` helper cleanly deduplicates the not-running and unreachable code paths, and `reconcile` is correctly skipped on the unreachable path — avoiding eviction of `agent_skills` rows when the disk isn't observable (preserves the self-heal invariant's intent).
- ✅ **Looks good:** Clear, accurate doc comments explaining *why* "running" can be unreachable (cached `Ready` vs. live probe). Scope is honestly stated in the PR body (siblings like `skills.list`, files/connections panels left for a follow-up).

**Verdict: aligned (one possible-drift question)**

The diff touches only `packages/api-server/src/modules/skills/`; no architecture docs are modified. The relevant page is `docs/architecture/skills.md` (`Last verified: 2026-06-25`). This is an error-handling refinement to the read path, not a responsibility or data-model change.

**Possible drift**
- ⚠️ **Check 1 (architecture-page drift):** `skills.md` describes the *Reconciled state* flow and the invariant *"`agent_skills` is a declarative record that self-heals on every `state` read"* on the happy path only. This PR adds a behavioral sub-case — "running but unreachable" ⇒ tracked-only, **no reconcile** (ghost rows are not dropped). Human-judgement question: is this fallback worth a sentence in *Reconciled state* / *Invariants* (the self-heal invariant now has an explicit transient exception), or is it a fine-grained transient-error detail below the architecture page's altitude? No hard drift asserted — it sits on the boundary of the drift rule.

No other checks fired (no `Last verified` staleness, no ADR leak, no volatile-content addition, no new subsystem, no cross-reference rot).

### Verdict
APPROVE — A correct, well-reasoned, well-tested defensive fix that turns a transient 500 into graceful degradation without over-catching; only a minor test-coverage suggestion and one optional doc note.

---

PR #2212, opened 2026-07-01
fix(approvals): resolve expired inbox requests after a verdict

### LINKED ISSUE #2125: UI - Expired inbox requests never resolve in the UI, making repeat actions appear to do nothing
## Problem

When I act on an expired ("timed out") request in my inbox by selecting one of the available options (Allow permanently, Allow host, or Deny forever), the UI never indicates that the action succeeded.

* The rule is saved on the first click—I can confirm it in the agent's network settings.
* However, the request remains in the inbox with its action buttons still enabled, making it look unresolved.
* Clicking the same action again does nothing because the rule already exists, but there's no indication that the action was ignored.

As a result, there's no confirmation that my initial action succeeded, the request appears to remain actionable, and subsequent clicks fail silently.

## Expected behavior

* After I act on an expired request, it should be marked as resolved or completed.
* The request should no longer be actionable after a choice has been made.
* If an action is ignored because the rule already exists, the UI should indicate that instead of failing silently.

## Notes

* This only affects expired/timed-out requests. Active requests appear to resolve correctly.
* The rule is successfully persisted on the first action; the issue is that the inbox item never updates to reflect it.

## Screenrecording

https://github.com/user-attachments/assets/95c89e61-af36-4712-a1d0-b21dd86f9f0e

kapetr:
Acting on an expired (timed-out) egress request in the inbox wrote the rule but left the row `expired` — `resolvePending`'s CAS only matches `pending`, so the inbox item kept its buttons and repeat clicks silently no-oped. Adds `resolveExpired` (expired → resolved) called from the rule-writing verdicts when the pending CAS misses, so expired rows clear like active ones.

### Scope
Covers the ext_authz (network/egress) inbox — what #2125 describes. The acp_native (harness permission) verdict paths have an analogous expired-row problem but are rooted in a different cause (a 24h inbox-only expiry vs. the harness's no real timeout); tracked separately in #2213.

Closes #2125

### DIFF
--- a/packages/api-server/src/modules/approvals/infrastructure/approvals-repository.ts
+++ b/packages/api-server/src/modules/approvals/infrastructure/approvals-repository.ts
@@ -46,6 +46,11 @@ export interface ApprovalsRepository {
     decidedBy: string,
     opts?: { markDelivered?: boolean },
   ): Promise<boolean>;
+  resolveExpired(
+    id: string,
+    verdict: "allow" | "deny",
+    decidedBy: string,
+  ): Promise<void>;
   /** Idempotent. Stamps `delivered_at` on a row whose response frame has
    *  reached the wrapper. Re-running is harmless: the WHERE keeps it from
    *  overwriting an earlier delivery timestamp. */
@@ -233,6 +238,25 @@ export function createApprovalsRepository(db: Db): ApprovalsRepository {
         )
         .returning({ id: pendingApprovals.id });
       return rows.length > 0;
+    },
+
+    async resolveExpired(id, verdict, decidedBy) {
+      const now = new Date();
+      await db
+        .update(pendingApprovals)
+        .set({
+          status: "resolved",
+          verdict,
+          decidedBy,
+          resolvedAt: now,
+          deliveredAt: now,
+        })
+        .where(
+          and(
+            eq(pendingApprovals.id, id),
+            eq(pendingApprovals.status, "expired"),
+          ),
+        );
     },
 
     async markDelivered(id) {
--- a/packages/api-server/src/__tests__/unit/approvals-service-expired.test.ts
+++ b/packages/api-server/src/__tests__/unit/approvals-service-expired.test.ts
@@ -1,0 +1,111 @@
+import { describe, expect, test } from "vitest";
+
+import type { PendingApprovalRow } from "../../modules/approvals/domain/types.js";
+import type { ApprovalsRepository } from "../../modules/approvals/infrastructure/approvals-repository.js";
+import { createApprovalsService } from "../../modules/approvals/services/approvals-service.js";
+
+function extAuthzRow(
+  overrides: Partial<PendingApprovalRow> = {},
+): PendingApprovalRow {
+  return {
+    id: "appr-1",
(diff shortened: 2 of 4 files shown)


packages/api-server/src/modules/approvals/infrastructure/approvals-repository.ts
```diff
@@ -235,6 +240,25 @@ export function createApprovalsRepository(db: Db): ApprovalsRepository {
       return rows.length > 0;
     },
 
+    async resolveExpired(id, verdict, decidedBy) {
```
dam-code-guardian, line 243: 🟢 **Suggestion:** Consider returning `Promise<boolean>` here (like `resolvePending`), reporting whether the `status = 'expired'` CAS actually matched a row. As written, the caller reports the `rule_written_expired` outcome unconditionally even if the guard matched zero rows (e.g. a concurrent path already flipped it to `resolved`). Non-blocking — the egress rule write is the idempotent source of truth.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `b2450cb`

## PR #2212: fix(approvals): resolve expired inbox requests after a verdict
**Author:** kapetr | **Branch:** fix/inbox-expired-approval-resolve → main | **Changes:** +139 −3 (4 files)

### Summary
Acting on an expired ext_authz inbox request wrote the egress rule but left the row `expired` (the `resolvePending` CAS matches only `pending`), so the inbox item kept its buttons and repeat clicks silently no-oped. This adds a `resolveExpired` (expired → resolved) fallback, invoked from the three rule-writing verdicts (`approvePermanent`, `approveHost`, `denyForever`) when the pending CAS misses, so expired rows clear like active ones. acp_native verdict paths are intentionally left out (no egress rule / held call to reconcile) per the PR description.

### Findings
- 🟢 **Suggestion:** `resolveExpired` returns `void`, unlike `resolvePending`'s `boolean` CAS result (`packages/api-server/src/modules/approvals/infrastructure/approvals-repository.ts:243`). The verdict still reports `rule_written_expired` even if the `status = 'expired'` guard matched zero rows (e.g. a concurrent path already resolved it). Returning `boolean` would let the caller distinguish a real expired-row clear from a no-op, mirroring `resolvePending`. Non-blocking — the rule write is the source of truth and idempotent.
- ✅ **Looks good:** The fix is correctly scoped to the three rule-writing verdicts; `approveOnce`/`dismiss` stay pending-only (they write no rule, so reviving an expired held call would be wrong).
- ✅ **Looks good:** `resolveExpired` stamps `deliveredAt` alongside `resolvedAt`, matching the semantics that an expired row has no live frame to deliver — consistent with the fake repo used in tests.
- ✅ **Looks good:** Good regression coverage — the new test file exercises approvePermanent/denyForever on expired rows (`rule_written_expired` outcome) and confirms the live-pending CAS path still returns `applied`.

### Verdict
APPROVE — Well-scoped, well-tested bug fix that closes a real inbox dead-button issue; the lone suggestion is a minor ergonomics nit, not a blocker.

---

PetrBulanek (approved):

PR #2214, opened 2026-07-01
fix(api-server): make ws liveness the source of truth for ACP turns

### LINKED ISSUE #2122: Slack errors on the 120s ACP inactivity timeout
## Problem

Messaging an agent in Slack often fails with a raw error in the thread:

> Error: ACP connection timed out after 120s of inactivity.

It clusters on longer tasks, where the connection stays quiet past the 120s window. But the agent keeps running: the completed reply shows up in the web UI, and the Slack thread continues normally afterward. The work succeeds; Slack just stops waiting and posts an error that looks like the request died, so users retry or give up needlessly.

## Proposed solution

Don't report a hard failure while the agent is still working. Keep waiting while it's making progress, and deliver the reply to the thread when ready. If a wait limit is unavoidable, replace the raw error with an honest "still running" message instead of a failure.

### Scope

This is the inactivity timeout on an active session, distinct from the "agent did not become ready" wake timeout (#468). Related: #1444.

jezekra1:
## Problem

Messaging an agent over a channel (Slack/Telegram/forks) fails with a raw `Error: ACP connection timed out after 120s of inactivity` on longer tasks, while the agent keeps running and the reply lands in the web UI. Root cause verified in #2122.

`withAcpConnection` (`acp-client.ts`) armed a 120s timer that **only `sessionUpdate` reset**. `requestPermission` / file ops / `extNotification` didn't. So the timer inferred "agent alive" from "agent is streaming" and aborted a *live* turn during any silent stretch >120s:

- **Egress approval holds** — `approvalHoldSeconds` defaults to **1800s** (30 min). The agent's outbound call is blocked server-side while the 120s ACP timer kills the connection at 2 min — the two timers disagreed by **15×**.
- **Long tools** with no intermediate streaming (build/test/install, slow API/MCP).
- **Slow first token / long thinking**.

Scope note: the **web UI is unaffected** — its `acp-relay` is a raw passthrough with no app-level timeout (that's why completed replies always show up there). The cutoff lived only in the shared `sendPrompt` path.

## Fix

- **WS ping/pong is now the liveness source of truth.** `withAcpConnection` runs a 30s ping/pong heartbeat (mirrors `ssh-relay.ts`); the agent-runtime `ws` server auto-pongs, so this is api-server-only. A live-but-silent turn stays connected; only a dead/half-open socket aborts (detected within ~60s).
- **Configurable absolute per-turn ceiling** (`acpTurnCeilingSeconds`, default **1h**) as a backstop for a wedged-but-still-ponging agent. **Validated `>= approvalHoldSeconds`** at config load, so the two timers can never disagree again.
- **Composition-root wiring** — the ceiling is bound into the acp-client factory in `index.ts` and injected into the channel workers, which drop their direct `createAcpClient` import (a pre-existing DI leak).
- **Helm** — `apiServer.acpTurnCeilingSeconds: 3600` + `ACP_TURN_CEILING_SECONDS` env on the api-server.

## Testing

- `mise run check` (all packages), `mise run api-server:test` (144 passing), `helm lint` + `helm template | kubeconform` — all green.
- **Reproduced and verified in-cluster over Slack**: `sleep 130 && echo done` errored at ~120s on the old build; on the fixed build it waits quietly and returns `done`.

## Notes / follow-up

- Left the issue's optional "still running" Slack UX out — with a 1h ceiling the connection no longer dies on legitimate work, so the remaining error surface is genuine failures worth reporting honestly. Easy follow-up if wanted.
- Drive-by observation (not changed here): `triggerSession` in `acp-client.ts` has no callers anywhere in the repo — appears to be dead code.

Fixes #2122

### DIFF
--- a/packages/api-server/src/core/acp-client.ts
+++ b/packages/api-server/src/core/acp-client.ts
@@ -9,7 +9,12 @@ import type {
 } from "@agentclientprotocol/sdk/dist/schema/types.gen.js";
 import { podBaseUrl } from "../modules/agents/infrastructure/k8s.js";
 
-const TIMEOUT_MS = 120_000;
+/** How often we ping the agent-runtime ws to prove the socket is alive.
+ *  Mirrors the ssh-relay heartbeat. */
+const PING_INTERVAL_MS = 30_000;
+/** Fallback per-turn ceiling when a caller doesn't supply one (tests, direct
+ *  use). Production wires config.acpTurnCeilingSeconds through the factories. */
+const DEFAULT_TURN_CEILING_MS = 60 * 60 * 1000;
 
 function wsStream(url: string): Promise<{ stream: Stream; ws: WebSocket }> {
   return new Promise((resolve, reject) => {
@@ -110,6 +115,7 @@ async function withAcpConnection<T>(
   url: string,
   clientName: string,
   handlers: { sessionUpdate?: (params: any) => Promise<void> },
+  turnCeilingMs: number,
   fn: (
     connection: ClientSideConnection,
     init: InitializeResponse,
@@ -118,11 +124,35 @@ async function withAcpConnection<T>(
   const { stream, ws } = await wsStream(url);
 
   const ac = new AbortController();
-  let timer = setTimeout(() => ac.abort(), TIMEOUT_MS);
-  const resetTimeout = () => {
-    clearTimeout(timer);
-    timer = setTimeout(() => ac.abort(), TIMEOUT_MS);
-  };
+  let abortReason = "ACP connection aborted";
+
+  // Liveness is a transport signal, not an application one. The agent-runtime's
+  // ws server auto-pongs, so a live-but-silent turn — an egress approval hold, a
+  // long tool, a slow first token — stays connected; only a dead/half-open
+  // socket aborts. Mirrors the ssh-relay heartbeat.
+  let alive = true;
+  ws.on("pong", () => {
+    alive = true;
+  });
+  const heartbeat = setInterval(() => {
+    if (!alive) {
+      abortReason = "ACP connection lost (agent unreachable)";
+      ac.abort();
+      return;
+    }
+    alive = false;
+    try {
+      ws.ping();
+    } catch {}
+  }, PING_INTERVAL_MS);
+
+  // Absolute backstop against a wedged-but-still-ponging agent. Sized at/above
+  // the egress approval hold (config enforces acpTurnCeilingSeconds >=
+  // approvalHoldSeconds) so a human taking the full approval window never trips it.
+  const ceiling = setTimeout(() => {
+    abortReason = `ACP turn exceeded the ${Math.round(turnCeilingMs / 1000)}s ceiling`;
+    ac.abort();
+  }, turnCeilingMs);
 
   const connection = new ClientSideConnection(
     () => ({
@@ -135,7 +165,6 @@ async function withAcpConnection<T>(
         };
       },
       async sessionUpdate(params: any) {
-        resetTimeout();
         await handlers.sessionUpdate?.(params);
       },
       async writeTextFile() {
@@ -150,7 +179,8 @@ async function withAcpConnection<T>(
   );
 
   const cleanup = () => {
-    clearTimeout(timer);
+    clearInterval(heartbeat);
+    clearTimeout(ceiling);
     if (
       ws.readyState === WebSocket.OPEN ||
       ws.readyState === WebSocket.CONNECTING
@@ -159,7 +189,6 @@ async function withAcpConnection<T>(
   };
 
   try {
-    ac.signal.addEventListener("abort", cleanup, { once: true });
     const init = await connection.initialize({
       protocolVersion: 1,
       clientCapabilities: { fs: { readTextFile: true, writeTextFile: true } },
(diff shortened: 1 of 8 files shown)


dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `73c5901`

## PR #2214: fix(api-server): make ws liveness the source of truth for ACP turns
**Author:** jezekra1 | **Branch:** fix/acp-inactivity-timeout → main | **Changes:** +120 −36 (7 files)

### Summary
Replaces the 120s ACP inactivity timer (which only `sessionUpdate` reset, so it aborted live-but-silent turns — egress-approval holds, long tools, slow first token) with a 30s ws ping/pong heartbeat as the liveness source of truth plus a configurable absolute per-turn ceiling (`acpTurnCeilingSeconds`, default 1h), validated `>= approvalHoldSeconds` at config load so the two timers can never disagree. Also rewires the Slack/Telegram workers to receive `AcpClient` factories via DI instead of importing `createAcpClient` directly. The fix's core assumption — the agent-runtime ws server auto-pongs — was verified against `packages/agent-runtime/src/server.ts` (default `autoPong`, not overridden).

### Findings
- 🟡 **Warning:** The heartbeat treats a slow/missed pong as death with only a single-interval grace: `ping()` is sent, `alive` set `false`, and if no pong arrives within the next 30s tick the turn aborts (`acp-client.ts:137-147`). A transient event-loop stall or GC pause on a healthy agent-runtime that delays the pong past one 30s window will abort an in-flight turn with `ACP connection lost (agent unreachable)`. Consider tolerating N consecutive missed pongs (e.g. abort on the 2nd), matching the more forgiving posture the ceiling backstop already provides.
- 🟢 **Suggestion:** `ws.ping()` is wrapped in `try {} catch {}` that swallows the error silently (`acp-client.ts:144-146`). A ping throwing means the socket is already broken — worth at least a debug log so a real transport failure isn't invisible.
- 🟢 **Suggestion:** No unit test covers the new config `refine` (`acpTurnCeilingSeconds >= approvalHoldSeconds`) — there is no `config.test.ts`. A tiny test asserting the invalid combination throws would lock in the "two timers can never disagree" guarantee against regression.
- 🟢 **Suggestion:** Per the PR description, `triggerSession` in `acp-client.ts` has no callers repo-wide — dead code worth removing in a follow-up (not introduced by this PR).
- ✅ **Looks good:** The DI refactor is clean — both channel workers drop their direct `createAcpClient`/`createForkAcpClient` imports and the ceiling+namespace are bound once at the composition root (`index.ts:377-387`); no dangling `namespace` references remain in `slack.ts`/`telegram.ts`.
- ✅ **Looks good:** The `Promise.race` refactor correctly removes the separate abort→cleanup listener — on abort the race rejects and the `finally` runs `cleanup()` (clears both the heartbeat interval and the ceiling timeout, closes the socket), so no timer or socket leaks on any exit path.
- ✅ **Looks good:** Config validation moved to a `.refine`d schema and the ceiling is threaded through both factories with a sensible `DEFAULT_TURN_CEILING_MS` fallback for tests/direct use; Helm wiring (`values.yaml` + `app.yaml` env) matches.

### Verdict
COMMENT — Solid, well-reasoned fix that resolves a real 15× timer disagreement; the ping/pong liveness assumption checks out. Non-blocking: consider a small grace on missed pongs and a config-refine unit test before merge.

---

packages/api-server/src/core/acp-client.ts
```diff
@@ -118,11 +124,35 @@ async function withAcpConnection<T>(
...
+    if (!alive) {
+      abortReason = "ACP connection lost (agent unreachable)";
+      ac.abort();
+      return;
+    }
+    alive = false;
+    try {
+      ws.ping();
+    } catch {}
+  }, PING_INTERVAL_MS);
```
dam-code-guardian, line 160: 🟡 **Warning:** Single-interval grace on missed pongs. After `ws.ping()`, `alive` is set `false`; if the pong doesn't land before the next 30s tick, the turn aborts as `agent unreachable`. A transient event-loop stall or GC pause on a healthy agent-runtime that delays the pong past one 30s window will kill an in-flight turn. Consider tolerating N consecutive missed pongs (abort on the 2nd) for a more forgiving liveness check.

packages/api-server/src/core/acp-client.ts
```diff
@@ -118,11 +124,35 @@ async function withAcpConnection<T>(
...
+  const heartbeat = setInterval(() => {
+    if (!alive) {
+      abortReason = "ACP connection lost (agent unreachable)";
+      ac.abort();
+      return;
+    }
+    alive = false;
+    try {
+      ws.ping();
+    } catch {}
```
dam-code-guardian: 🟢 **Suggestion:** `ws.ping()` errors are swallowed silently here. A throw means the socket is already broken — a debug log would keep a real transport failure from being invisible.

tomkis (approved):

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `ebfddc5`

## PR #2214: fix(api-server): make ws liveness the source of truth for ACP turns
**Author:** jezekra1 | **Branch:** fix/acp-inactivity-timeout → main | **Changes:** +195 −36 (8 files)

### Summary
The three new commits since the last review address all three actionable items from that review. Turn liveness remains a 30s ws ping/pong signal with an absolute per-turn ceiling validated `>= approvalHoldSeconds`; the DI refactor of the channel workers is unchanged.

### Changes since last review
Previous HEAD: `73c5901` (2026-07-01T09:46:12Z) — verdict COMMENT

- ✅ **Fixed:** Single-interval grace on missed pongs — now `MAX_MISSED_PONGS = 2` (~90s of silence before abort), so a single GC pause / event-loop stall on a healthy agent-runtime no longer kills a live turn (`acp-client.ts:15,138-155`).
- ✅ **Fixed:** `ws.ping()` errors are no longer swallowed silently — a throw is logged via `getLogger().debug({ err, clientName }, ...)` (`acp-client.ts:148-153`).
- ✅ **Fixed:** The config `refine` (`acpTurnCeilingSeconds >= approvalHoldSeconds`) now has unit coverage — new `config.test.ts` asserts reject-below-hold, accept-equal, and accept-defaults (`__tests__/unit/config.test.ts`).
- 🔁 **Still present:** `triggerSession` in `acp-client.ts` is dead code (no callers repo-wide) — carried over; author flagged it as an out-of-scope follow-up.

### Findings
- 🟢 **Suggestion:** `triggerSession` remains unreferenced across the repo — worth removing in a follow-up (not introduced by this PR).
- ✅ **Looks good:** The missed-pong counter is correct — `missedPongs` resets to 0 on every `pong`, increments once per 30s tick, and aborts only on the tick where it has already reached 2, giving a genuine ~90s tolerance window well inside the turn ceiling.

### Verdict
APPROVE — Every actionable item from the prior review is resolved (missed-pong grace, ping-error logging, config-refine test) with correct logic; only the pre-existing `triggerSession` dead-code follow-up remains, which is non-blocking.

---

PR #2316, opened 2026-07-02
fix(claude-code): DDG WebSearch/WebFetch fallback on non-Anthropic backends (#1087)

### LINKED ISSUE #1087: WebSearch fails with non-Anthropic credentials (custom backend)
## Problem

When Claude Code runs against a custom backend (non-Anthropic credentials), the native `WebSearch` tool does not work. Asking the agent to search the web (e.g. "search for abc online") makes it invoke the built-in tool, which then fails.

## Possible solutions

- Use a plugin that replaces WebSearch — e.g. a DuckDuckGo-powered drop-in like the `cc-websearch` project we explored.
- Disable the built-in `WebSearch` tool so the agent doesn't attempt it.
- Replace it with a web-search tool hosted as an MCP server.

Tomas2D:
Resolves #1087.

## Problem

Claude Code's built-in `WebSearch`/`WebFetch` are Anthropic **server-side** tools. When a pod fronts a custom (non-Anthropic) model backend — the platform's `model-gateway` case — the upstream doesn't implement them, so asking the agent to search the web fails.

## Fix

Vendor [cc-websearch](https://github.com/Djarvur/cc-websearch) (DuckDuckGo-backed, MIT) as self-contained `node` bundles and — **only on custom backends** — deny the built-ins via a managed-settings `PreToolUse` hook that redirects the model to replacement skills. On genuine Anthropic backends the hook stays silent and the native tools work unchanged.

Lives in the **claude-code harness image**, so `nous` and `openevolve` (both derive `FROM platform-claude-code`) inherit the fix; non-Claude harnesses (codex, pi-agent, bob, k-search) are untouched.

## Changes

- **Dockerfile** — fetch cc-websearch `v1.2.4` bundles (pinned commit `02d4993` + SHA-256 verification) into `/opt/cc-websearch`; ship `deny-web-tool.sh`; install the managed-settings hooks. Bundles need only `node` (no `npm install`), and stay in the image layer (not duplicated onto every PVC).
- **model-gateway.sh** — export `PLATFORM_CUSTOM_BACKEND=1` when fronting a custom upstream (set before the readiness wait, so it holds on the gateway-not-ready path too). DRY: the gateway already computes this; the hook just consumes it.
- **deny-web-tool.sh** — `PreToolUse` hook gated on `PLATFORM_CUSTOM_BACKEND`; emits a `deny` decision → replacement skill on custom backends, nothing (built-in proceeds) otherwise.
- **skills** — seed `websearch`/`webfetch` skills that invoke the bundles via `node` (discovered through the existing `.claude/skills` → `.agents/skills` symlink).
- **helm** — add `lite.duckduckgo.com` (websearch's only backend host) to the `trusted` egress preset. Node's `fetch` already tunnels through Envoy via the controller's `NODE_USE_ENV_PROXY=1`. `webfetch`'s arbitrary hosts use the normal per-host egress-approval flow.

## Verification

- Both bundles run under node 24 and return the expected output (`<search_results>` XML / extracted markdown).
- Hook: custom backend → valid `deny` JSON, exit 0; genuine Anthropic → empty output, exit 0. `sh -n` clean; `managed-settings.json` valid JSON.
- `helm:check:lint` and `helm:check:render` (kubeconform) pass; `lite.duckduckgo.com` renders into the trusted-hosts ConfigMap.

**Not yet verified in-cluster:** the Docker image build and true in-pod e2e (deny → skill → proxied DDG fetch). 

## Note / edge case

During the brief cold-boot window before `model-gateway.sh` finishes (or on a `dam-run` executor pod that doesn't run the chat harness), the marker can be momentarily unset on a custom backend — the built-in would then be attempted and fail as it does today, rather than being redirected. Steady-state chat sessions are unaffected.

### DIFF
--- a/deploy/helm/platform/values.yaml
+++ b/deploy/helm/platform/values.yaml
@@ -658,6 +658,10 @@ apiServer:
     - cdn.mise.run
     # Bob Shell (IBM)
     - api.us-east.bob.ibm.com
+    # DuckDuckGo — backend for the claude-code harness's `websearch` skill, the
+    # replacement for the built-in WebSearch tool on non-Anthropic backends
+    # (#1087). `lite.duckduckgo.com` is the only host the bundle contacts.
+    - lite.duckduckgo.com
 
   # -- Optional platform-wide defaults for OAuth app credentials. When a field
   # is set the connect form for the matching app collapses the corresponding
--- a/packages/agents/claude-code/Dockerfile
+++ b/packages/agents/claude-code/Dockerfile
@@ -27,12 +27,12 @@ COPY --chmod=0755 harness-terminal.sh /usr/local/bin/harness-terminal
 # deny the built-ins and the seeded skills (.agents/skills/{websearch,webfetch})
 # invoke these self-contained esbuild bundles instead, which need only `node`.
 # Pinned to a commit SHA and checksum-verified.
-ARG CC_WEBSEARCH_REF=02d4993492ed0a957e7908bfff61fef245d36583
+ARG CC_WEBSEARCH_REF=c10add64e36e98259a31fe097e72ff9085e68dee
 RUN mkdir -p /opt/cc-websearch &&\
     base="https://raw.githubusercontent.com/Djarvur/cc-websearch/${CC_WEBSEARCH_REF}/skills" &&\
     curl -fsSL "$base/websearch/scripts/websearch.cjs" -o /opt/cc-websearch/websearch.cjs &&\
     curl -fsSL "$base/webfetch/scripts/webfetch.cjs" -o /opt/cc-websearch/webfetch.cjs &&\
-    node -e 'const c=require("crypto"),f=require("fs"),d="/opt/cc-websearch/";const want={"websearch.cjs":"3f3054ab1ec65f301152e0567b6415b8c92264a2fb94e9e7d95685911cde058e","webfetch.cjs":"56cc2e9c23e594308ae010e185211313f637f38d633cdcbdd7b2c5e565355054"};for(const[n,h]of Object.entries(want)){const g=c.createHash("sha256").update(f.readFileSync(d+n)).digest("hex");if(g!==h){console.error(`cc-websearch: checksum mismatch for ${n}: got ${g} want ${h}`);process.exit(1);}}console.log("cc-websearch bundles verified");'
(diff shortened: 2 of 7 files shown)


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `b5c9c0a`

## PR #2316: fix(claude-code): DDG WebSearch/WebFetch fallback on non-Anthropic backends (#1087)
**Author:** Tomas2D | **Branch:** fix/websearch-custom-backend → main | **Changes:** +134 −2 (7 files)

### Summary
Vendors DuckDuckGo-backed WebSearch/WebFetch replacements (cc-websearch, MIT) into the claude-code harness image and, **only on custom (non-Anthropic) backends**, denies the built-in server-side tools via a managed-settings `PreToolUse` hook that redirects the model to seeded replacement skills. Genuine Anthropic backends are untouched.

### Findings
- 🟢 **Suggestion:** The cold-boot race the PR body already flags — before `model-gateway.sh` exports `PLATFORM_CUSTOM_BACKEND`, or on a `dam-run` executor pod, the marker is momentarily unset on a custom backend, so a built-in web tool would be attempted and fail as it does today rather than being redirected. Steady-state chat is unaffected; acceptable for this fix, but worth a follow-up if pre-gateway tool calls turn out to happen in practice. (`packages/agents/claude-code/model-gateway.sh`)
- ✅ **Looks good:** Strong supply-chain hygiene — the vendored bundles are pinned to a commit SHA (`CC_WEBSEARCH_REF`) **and** SHA-256-verified at build time, failing the image build on mismatch. Bundles stay in the image layer (not duplicated onto every PVC) and need only `node`.
- ✅ **Looks good:** The deny hook is correctly gated: `deny` decision + skill redirect only when `PLATFORM_CUSTOM_BACKEND=1`, silent (built-in proceeds) otherwise. The export is set before the readiness wait, so it holds on the gateway-not-ready path too.
- ✅ **Looks good:** Egress is scoped minimally — only `lite.duckduckgo.com` (the bundle's sole backend host) added to the `trusted` preset; `webfetch`'s arbitrary hosts ride the normal per-host egress-approval flow. The `websearch` skill is tool-scoped (`allowed-tools: Bash(node *)`).

### Verdict
APPROVE — Well-scoped harness fix with commit-pinned + checksum-verified vendored bundles, a correctly-gated deny hook, and minimal egress surface; the one cold-boot edge case is already documented and acceptable for this change.

---

PR #2317, opened 2026-07-02
feat(controller): emit operational telemetry via OpenTelemetry

pilartomas:
Closes #2313 (sub-issue of #929).

## What

The controller now emits the three correlated OpenTelemetry signals for its own operation, exported over OTLP to the bundled collector when the telemetry backend is enabled:

- **Traces** — one span per reconcile pass (agent/fork/run) carrying resource kind/name, outcome (`success` / `error` / `not_found` / `decode_error` / `backoff_exceeded`), and requeue count; the existing per-phase timings become span events, with slow reconciles flagged. Background loops (idle check, warm-pool sweep, orphan sweep) get one span per pass.
- **Outbound-call spans** — every Kubernetes API request (client-go transport) and the idle checker's busy probe emit `otelhttp` client spans, parented to the surrounding reconcile/pass span and named method + path, so a slow phase points at the exact API call that dominated.
- **Metrics** — reconcile duration histogram and outcome counter, full workqueue instrumentation (depth, adds, latency, work duration, retries) via client-go's metrics hook, and Go runtime metrics.
- **Logs** — the stderr JSON stream is unchanged byte-for-byte, gains `trace_id`/`span_id` fields on lines logged inside a span, and each record is additionally exported over OTLP with trace correlation.

## How

- New `packages/controller/pkg/telemetry` package. The SDK activates **only** when the standard `OTEL_EXPORTER_OTLP_ENDPOINT` environment is present (and honors `OTEL_SDK_DISABLED`); otherwise the controller runs exactly as before — no exporters, no goroutines, no-op instruments.
- The chart sets `OTEL_EXPORTER_OTLP_ENDPOINT` / `OTEL_EXPORTER_OTLP_PROTOCOL` / `OTEL_SERVICE_NAME` on the controller deployment under the existing `clickstack.enabled` gate, pointing at the bundled collector over plain HTTP in-mesh (ztunnel supplies mTLS; the collector's authorization policy already admits the release namespace). Controller telemetry arrives without the gateway attribution header, so it carries no `platform.agent.id` — correctly classified as platform, not agent, telemetry.
- Workqueues are now named — client-go skips metrics for unnamed queues.

## Design note

[...]

Issue #929, opened 2026-06-15
Operational telemetry: emit logs, metrics, and traces via OpenTelemetry


pilartomas:
## Problem

The platform has structured logs and usage analytics, but **no operational telemetry**. There are no health metrics (latency, error rates, throughput, saturation) for the running services, and no way to follow a single request or scheduled run end-to-end when it's slow or failing. Today that means stitching disconnected log lines together by hand — and an on-call agent has nothing structured to query about the system's health.

## Proposed solution

Instrument all of the platform's services with **OpenTelemetry**, emitting three correlated signals — **logs, metrics, and traces** — over a standard protocol to a configurable backend. This gives operators per-service health metrics, a single end-to-end trace per flow, and correlated signals to pivot between. It also lines up with the service mesh (Istio), which already speaks OpenTelemetry, so mesh-level and application-level signals share one pipeline. The export destination stays configurable — no hardcoded backend.

**Open question:** whether the platform's deployment bundles a telemetry collector or exports directly to an externally-run one. Leaning toward keeping the collector out of the chart, like the rest of the cluster infra.

PR #2323, opened 2026-07-02
feat(api-server): emit operational telemetry via OpenTelemetry

### LINKED ISSUE #2312: API server emits operational telemetry via OpenTelemetry
## Problem

The API server is the platform's front door — every UI interaction, agent conversation, and integration call passes through it — yet it exposes no operational telemetry. There is no way to see its request latency, error rate, or throughput, and no way to follow a single slow or failing request through the layers it touches (authentication, database, the relay to agents). When something degrades, an operator is left correlating log lines by timestamp.

## Proposed solution

Instrument the API server with OpenTelemetry so it emits the three correlated signals — logs, metrics, and traces — over the standard protocol to a configurable destination:

- **Metrics** — request rate, latency, and error counts per operation, plus process health (memory, event-loop saturation), so an operator or on-call agent can query the service's health directly.
- **Traces** — one end-to-end trace per incoming request, with spans covering the significant internal steps (auth, database calls, calls out to agents), continuing any trace context arriving from upstream (the mesh or the UI).
- **Logs** — existing structured logs carry the trace identity, so an operator can pivot from a slow trace to its log lines and back.

Telemetry is emitted only when an export destination is configured; nothing is hardcoded to a particular backend.

### Dependencies

Sub-issue of #929.

### Out of scope

- The controller's instrumentation — tracked in its own sub-issue.
- Bundling a telemetry backend or collector for platform services — #929's open question; this issue only makes the service *emit*.

pilartomas:
Closes #2312 (sub-issue of #929). Sibling of #2317 (controller).

## What

The api-server now emits the three correlated OpenTelemetry signals for its own operation, exported over OTLP to the bundled collector when the telemetry backend is enabled:

- **Traces** — one trace per incoming request (continuing upstream trace context), with a child span per tRPC procedure and spans for the significant outbound calls: requests to agents (fetch/undici), Keycloak, channel platforms, Redis/BullMQ, the chat-state Postgres pool, and every ext-authz gRPC check. Kubelet health-probe requests are excluded.
- **Metrics** — per-procedure rate/latency/outcome (`platform.trpc.duration` histogram + `platform.trpc.total` counter with the tRPC error code), plus Node runtime health: event-loop delay/utilization, GC, and heap.
- **Logs** — the stdout Pino stream is unchanged; lines logged inside a traced request gain `trace_id`/`span_id`, and every record is additionally exported over OTLP with trace correlation.

## How

- **`--import` bootstrap** (`src/telemetry.ts`, its own tsup bundle preloaded by the Dockerfile CMD): the process is ESM, so instrumentation requires an `import-in-the-middle` loader hook registered before the app bundle evaluates — neither the SDK nor the instrumentations register it themselves. The bootstrap gates on the standard `OTEL_EXPORTER_OTLP_ENDPOINT` env (honoring `OTEL_SDK_DISABLED`); disabled, it loads nothing and startup is byte-identical (verified by diffing boot output).
- **Curated instrumentation list** rather than `@opentelemetry/auto-instrumentations-node`: the meta-package would add ~90 instrumentation packages of which exactly seven apply here (http, undici, grpc, ioredis, pino, pg, runtime-node). The curated set gives identical coverage with a fraction of the supply-chain surface — relevant on a repo with automated CVE triage.
- **Per-procedure middleware** rides the base procedure of both tRPC instances in the contract package (several routers bypass the auth-procedure builders, so attaching there would miss them). It depends only on `@opentelemetry/api` and is a no-op without a registered SDK.
- **Graceful flush**: the app's shutdown sequence awaits the SDK flush through a `globalThis` Symbol hook before `process.exit(0)` (capped at 5s so an unreachable collector can't eat the termination grace period). The bootstrap is never imported by the app — it's a separate bundle, and importing it would evaluate a second SDK.
- **Helm**: the apiserver deployment gains the three `OTEL_*` env vars under the existing `clickstack.enabled` gate, pointing at the bundled collector over plain HTTP in-mesh (ztunnel supplies mTLS; the collector's authorization policy already admits the release namespace). Arrives without the gateway attribution header → no `platform.agent.id` → classified as platform telemetry.

## Known gap (follow-up)

No spans for the primary Postgres pool: `postgres.js` + drizzle are deliberately tsup-bundled (`noExternal`) and no official OTel instrumentation exists for postgres.js regardless. Drizzle's own OTel hooks are the likely v2 path; follow-up issue to come.

## Testing

- Unit tests: gate semantics; middleware span naming/attributes/status, tRPC error-code classification (result.ok and thrown paths), active-span nesting under a parent, and lazy instrument creation (metrics API has no proxy provider).
- Signal-flow smoke test: booted the built bundle under `--import` against a local OTLP sink — instrumented fetch produced a client span, and the sink received `/v1/traces`, `/v1/metrics`, and `/v1/logs`; stdout stayed plain Pino JSON. Disabled path verified byte-identical to a plain `node dist/index.js` boot.
- `mise run check`, `mise run test`, helm lint + render all pass.


packages/api-server-api/src/trpc-telemetry.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/**
+ * Wraps one tRPC procedure invocation in a span and records its metrics.
+ * tRPC middlewares surface failures via `result.ok`/`result.error` rather
+ * than throwing; the catch branch is defensive only.
+ */
+export async function withTrpcTelemetry<R extends { ok: boolean }>(
+  path: string,
+  type: "query" | "mutation" | "subscription",
```
dam-code-guardian, line 61: 🟢 **Suggestion:** For `type: "subscription"`, this span ends and the `platform.trpc.duration` histogram is recorded when `next()` resolves — i.e. at subscription *setup*, not when the stream closes. Harmless given current usage, but a one-line comment noting that subscription durations measure setup latency (not stream lifetime) would save a future reader from misreading the metric.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `ac31e57`

## PR #2323: feat(api-server): emit operational telemetry via OpenTelemetry
**Author:** pilartomas | **Branch:** feat/apiserver-otel → main | **Changes:** +1163 −17 (16 files)

### Summary
Adds the api-server's own OpenTelemetry export (traces/metrics/logs over OTLP to the bundled collector), mirroring the controller work in the sibling PR #2317. The SDK is bootstrapped via a separate `--import` bundle gated on the standard OTLP endpoint env, a per-procedure tRPC middleware rides the base procedure of both tRPC instances, and shutdown flushes the SDK through a `globalThis` Symbol hook. Disabled path loads nothing and is byte-identical to a plain boot.

### Findings
- ✅ **Looks good:** The gate (`telemetry-gate.ts`) is pure and dependency-free, `OTEL_SDK_DISABLED` correctly wins over endpoint presence, and blank/whitespace endpoints are treated as unset — all covered by the parameterized `telemetry-gate.test.ts`.
- ✅ **Looks good:** `trpc-telemetry.ts` depends only on `@opentelemetry/api`, so it is a genuine no-op on a non-recording span when no SDK is registered (dev/test/disabled deploy). The lazy `getInstruments()` memoization correctly avoids the meter-before-provider no-op-forever trap, with `resetTrpcTelemetryForTest()` to re-bind in tests — the `trpc-telemetry.test.ts` "does not poison instruments" case pins exactly this.
- ✅ **Looks good:** Middleware attaches at the base procedure of both `t` and `harnessT`, so routers bypassing the auth-procedure builders still emit spans/metrics; tRPC `result.ok`/`result.error` classification and the defensive thrown-error path are both tested, and the span is made active inside `next()` (parent-nesting test confirms).
- ✅ **Looks good:** Curated instrumentation list (7 libraries actually used) instead of the ~90-package auto-instrumentations meta-package — smaller supply-chain surface, sensible on a repo with automated CVE triage. Kubelet `/api/health` probes are excluded from incoming spans.
- ✅ **Looks good:** Shutdown flush is reached via `Symbol.for("platform.otel.shutdown")` lookup rather than importing the bootstrap bundle (which would evaluate a second SDK), and is capped at 5s with `.unref()` so an unreachable collector can't eat the termination grace period.
- ✅ **Looks good:** Helm wiring is under the existing `clickstack.enabled` gate; `OTEL_SERVICE_NAME` and `service.version` (from the pre-existing `PLATFORM_APP_VERSION` env) resolve correctly, and the plain-HTTP in-mesh path relies on ztunnel mTLS + the collector's existing AuthorizationPolicy — consistent with the controller precedent.
- 🟢 **Suggestion:** `withTrpcTelemetry` accepts `type: "subscription"` but a subscription's span ends and its duration is recorded when `next()` resolves (subscription setup), not when the stream closes — so the `duration` histogram measures setup latency, not stream lifetime, for subscriptions. Harmless given current usage, but worth a one-line comment noting the intent so a future reader doesn't misread subscription latencies. (`packages/api-server-api/src/trpc-telemetry.ts:61`)

_(Both affected architecture pages were updated in the same PR: `observability.md` generalizes the platform-service export section to cover both controller and api-server and adds the api-server bullet incl. the documented Postgres-pool gap; `logging.md` documents the trace-correlation/OTLP-export behavior of the Pino stream. Both have `Last verified:` at 2026-07-02 matching the PR date. No new subsystem, no volatile-content or ADR leaks in the added lines.)_

### Verdict
APPROVE — Well-structured, correctly gated, and thoroughly tested telemetry addition with the affected architecture docs updated in lockstep; the lone suggestion is non-blocking.

---

PR #2200, opened 2026-06-30
refactor: cut over to Connections and retire the secrets module

PetrBulanek:
Connections becomes the **single** credential model, and the legacy secrets system is retired.

## What changes for users

- **Existing setups migrate automatically.** On upgrade, every provider and GitHub-PAT credential configured through the old flows becomes the equivalent Connection, and each agent's grants move with it — atomically, with no credential value ever exposed and **no action required**. Agents keep authenticating across the upgrade with no disruption window.
- **One way to manage credentials.** The Providers settings tab, the agent/sandbox creation flows, and the CLI's interactive create now offer and grant only Connections. The "Test credential" check on the Anthropic form is preserved.

## Running the migration

There is **no manual command** — the migration runs automatically on api-server startup, so **deploying this PR is running it**. No maintenance window; agents keep authenticating on their legacy credentials until each is flipped to a Connection.

1. **(Optional) Preview first.** The dry-run reads the legacy secrets, prints what it *would* do, and changes nothing:
   ```
   kubectl exec deploy/platform-apiserver -- node dist/migrations/secrets-to-connections.js --dry-run
   ```
   For a true *pre-deploy* preview, run that as a one-off pod from this PR's image while the live api-server is still on the old version — a new api-server runs the migration for real the moment it boots.
2. **Deploy this PR.** On boot the api-server drains every legacy provider/PAT Secret into the equivalent Connection and moves each agent's grants over. The first pass finishes before it serves traffic; a partial failure re-arms a capped retry (every 60s, up to 10 times).
3. **Verify the drain.** Check the boot log:
   ```
   kubectl logs deploy/platform-apiserver | grep secrets-migration
   # → secrets→connections: migrated N, skipped M, failed 0
   ```
   `failed 0` is a clean drain. Look into any `skipped` lines (lone PAT half, orphan twin, unrecoverable value) — those stay on the legacy path and need a manual decision.
4. **Cleanup is deferred on purpose.** Drained legacy Secrets are left in place — deleting them mid-rollout could wedge a gateway pod — and are swept by the follow-up (#2198) once a clean field drain is confirmed.

## Safety

The cutover lands the migration and removes the legacy surfaces, but keeps the controller's legacy credential path and the migration itself in place as inert safety so nothing breaks mid-upgrade. Those are removed by a follow-up once a clean drain is confirmed in the field ([#2198](https://github.com/dam-agents/dam/issues/2198)).

Closes #1273

Issue #2198, opened 2026-06-30
Remove inert legacy-secret safety code after clean drain (#1273 follow-up)


PetrBulanek:
## Context

#1273 made Connections the single credential model and deleted the secrets module, but deliberately kept two pieces of **inert safety** so migrated agents never hit a credential-less window during the upgrade. This issue removes them once a clean drain is field-confirmed.

## Gate — clean drain across all installs

- No agent still grants a legacy secret (`spec.grantedSecretIds`) — #2337 re-flips any straggler on every boot, so any install that ran a migration-era release satisfies this.
- No outstanding migration skip-and-log failures (malformed / orphan-half secrets resolved or accepted).

The drained `platform-cred-*` Secret **objects** themselves still linger: the #1273 migration deliberately never deleted them (deleting a Secret while a gateway pod is mid-roll on a revision that mounts it wedges the pod on `FailedMount`) and its header defers their deletion to this issue. Since the target installs have no kubectl access, work item 4 ships that promised sweep, and its always-emitted summary log line (`[legacy-secret-sweep] deleted N, skipped N, failed N`) is the per-install gate confirmation: `skipped > 0` means a still-granted secret was found, reported as an ERROR, and left in place for a human.

## Work

1. **Controller** (`packages/controller/pkg/reconciler/envoy.go`): delete the legacy *credentialed* `host-pattern` fallthrough in `chainsFromSecrets` and the hardcoded Anthropic env-mappings fallback in `credentialEnvVars`; remove the now-unused `auth-mode` constant. Audit `envoy_test.go` for legacy cases.
   *Scope corrections found during implementation:* the `secret-type` / `host-pattern` constants and the `chainsFromSecrets` non-connection branch still serve the live **allow-only** egress path (L7 promotion for path-scoped rules) — kept. `sdsFileKeyForHost` serves *connection* Secrets created before `sdsKey` stamping (2026-06-12) — kept, its removal is a separate field-gated cleanup. `envoyCredentialKeySDS` is still matched by #2344's `sdsDataKeys` digest — kept.
[...]

Issue #1273, opened 2026-06-17
Cut over to Connections and retire the secrets module


PetrBulanek:
## Context

Parity is done — #884 closed gaps 1–4 of #600, so every provider and GitHub-PAT credential can now be expressed as a Connection. Today the legacy **secrets** system and Connections still run side by side, with secrets as the active path for provider/PAT setup. This issue finishes the consolidation: move existing setups across and remove secrets, so Connections is the only credential model.

Split out from #600 gap 5 so #600 can close as "parity done."

## Scope

The cutover runs in this order — the legacy path only comes down once migrated agents are confirmed working:

- [ ] **Migrate existing setups automatically** — every credential configured through the old provider/secrets flows becomes the equivalent Connection, and each agent's grants move with it. Atomic per agent, so an agent never reaches a service with the wrong credential or none at all; no credential value is ever exposed; no one redoes their setup.
- [ ] **Confirm no disruption** — verify migrated agents still authenticate against their external services before anything legacy is removed.
- [ ] **Turn off the legacy path** — stop writing to secrets and remove the old provider setup surfaces: the CLI's interactive create flow, the Providers settings tab, and the provider-grant picker shown when creating or editing an agent.
- [ ] **Accept ADR-051** — record the consolidation decision as accepted.

## Outcome

Connections is the single credential model. The secrets module is gone, and existing agents keep working across the upgrade with no user action.

### Dependencies

- **Blocked on #601** — provider/PAT setup must already run on Connections, including the "replace key" flow, before the legacy surfaces can be removed.
- Parity precondition met by #884.

### Safety

The Anthropic OAuth-token path hasn't been exercised end-to-end and carries an open question about third-party detection (raised in #884 review). Validate it on a live cluster before migrating those credentials.

### References

#329 · #600 · #601 · ADR-051

PR #2337, opened 2026-07-03
fix(api-server): stop migration from resurrecting deleted connections

PetrBulanek:
## Problem

The #2200 migration defers deleting the legacy `platform-cred-*` Secrets to #2198, so they linger. The migration recreates a connection for any legacy Secret without one — so a connection the user **deletes gets rebuilt on the next api-server boot**. Users can't remove migrated connections.

## Fix

Only (re)create a connection **while an agent still grants the legacy Secret**. Once flipped, the Secret is ungranted, so a deleted connection stays deleted. Still-granted Secrets migrate and re-flip as before; an ungranted Secret with no connection (deleted, or an unused orphan) is logged and left for the #2198 sweep.

## Validated (local, full pre-cutover → migrate → delete → restart)

Fresh migration still drains (`migrated 2, failed 0`); after deleting a connection and restarting, it logged `skip … no granting agent` and **did not reappear**.

Interim fix for the #2200 → #2198 window. #2198 removes the migration and must also sweep the retained Secrets. Refs #1273, #2198

### DIFF
--- a/packages/api-server/src/modules/connections/migration/secrets-to-connections.ts
+++ b/packages/api-server/src/modules/connections/migration/secrets-to-connections.ts
@@ -21,6 +21,11 @@
  * Secret in place lets every gateway roll cleanly; the #1273 follow-up (#2198)
  * sweeps the drained Secrets once the field confirms no gateway still mounts them.
  *
+ * Because the Secret lingers, a connection is (re)created only while an agent
+ * still grants the legacy secret. After the first flip the secret is ungranted,
+ * so a connection the user then deletes stays deleted instead of being rebuilt
+ * on the next boot — the deferred-deletion window can't resurrect user deletes.
+ *
  * Non-blocking and self-disarming: a transient failure leaves the legacy secret
  * intact and re-arms a retry; a permanent skip (malformed / unrecoverable /
  * orphan half) is logged and left alone. A pass with nothing to migrate is a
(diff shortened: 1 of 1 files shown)


packages/api-server/src/modules/connections/migration/secrets-to-connections.ts
```diff
@@ -207,6 +212,24 @@ async function ensureConnection(
...
 
+  // Only (re)create a connection while an agent still grants the legacy secret.
+  // Once the first pass flips an agent, the legacy secret is ungranted — so a
+  // connection the user later deletes must NOT be rebuilt. Because deletion is
+  // deferred to #2198, the legacy Secret lingers, and a bare "no connection
+  // yet" check would resurrect the deleted connection on every boot (or the
+  // 60s retry). An ungranted secret with no connection is either that
+  // user-deleted case or an unused orphan; log and leave it for the #2198 sweep
+  // rather than migrate it into a connection nothing uses.
+  if (!existing && granting.length === 0) {
```
dam-code-guardian, line 223: 🟢 **Suggestion:** This resurrection guard is exactly the kind of boot-migration edge case worth locking down with a regression test (pre-cutover → migrate → delete the connection → restart → assert it is not recreated). The PR body says this was validated locally; there's no dedicated test file for `secrets-to-connections.ts` in the repo, so the guard currently has no automated coverage. Not blocking.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `229863f`

## PR #2337: fix(api-server): stop migration from resurrecting deleted connections
**Author:** PetrBulanek | **Branch:** fix/2200-migration-no-resurrect → main | **Changes:** +23 −0 (1 file)

### Summary
Adds a guard in `ensureConnection()` so the legacy-secrets→connections boot migration only (re)creates a connection while an agent still grants the legacy secret. Because secret deletion is deferred to #2198, the legacy Secret lingers; without this guard a connection the user deletes would be rebuilt on the next api-server boot (or the 60s retry). An ungranted secret with no connection (user-deleted, or an unused orphan) is now logged and left for the #2198 sweep.

### Findings
- ✅ **Looks good:** The guard is correctly ordered — it sits *after* `recoverValue()` (so an unrecoverable value still takes the earlier `value unrecoverable` skip) and *before* the dry-run/create branches, so no connection is ever created for an ungranted-and-absent credential. (`secrets-to-connections.ts:223`)
- ✅ **Looks good:** Returning `null` routes through the existing phase-1 `if (!res) { skipped++; continue; }` path, so the skip is counted and reported consistently with the other permanent skips; no agent-grant mutation happens. (`secrets-to-connections.ts:142`)
- ✅ **Looks good:** `findGrantingAgents()` unions across all of `cred.secretIds` (primary + twins), so a PAT/Bob pair where either half is still granted is correctly treated as "still granted" and migrates as before.
- 🟢 **Suggestion:** This resurrection guard is exactly the kind of boot-migration edge case that benefits from a regression test (pre-cutover → migrate → delete connection → restart → assert not recreated), which the PR body says was validated locally. There is no dedicated test file for `secrets-to-connections.ts` in the repo, so the behavior is currently only covered by manual validation. (`secrets-to-connections.ts:223`)

### Verdict
APPROVE — Small, well-reasoned defensive fix that closes a real user-facing bug (deleted connections resurrecting) with correct ordering and no behavioral regression for still-granted secrets; the only note is the absence of an automated regression test.

---

jezekra1 (approved):

PR #2344, opened 2026-07-06
fix: fork gateway Envoy crash on stale connection Secrets + leaked gateway pods

pilartomas:
## Problem

Tagging the Slack bot as a non-owner (foreign replier) reliably failed: the fork gateway's Envoy exited at boot with

```
paths must refer to an existing path in the system: '/etc/envoy/credentials/cred-platform-conn-<...>/host-YXBpLmdpdGh1Yi5jb20.sds.yaml' does not exist
```

np-gate in the fork agent pod then failed convergence (the gateway never serves `/__platform_healthz` with Envoy down), the Job failed, and its 60s TTL deleted the agent pod — while the gateway pod crash-looped **forever**.

## Root causes

**Crash** — the replier owned a connection Secret written by the since-deleted pre-cutover writer (`platform-conn-<hash>-<hash>` naming): its `injection-hosts` annotation has no `sdsKey` fields and its data keys use the old per-host naming. The controller's fallback computes the current base64url filename, which never existed in that Secret's data, and a bootstrap referencing a missing SDS file is a **fatal Envoy boot error**. No migration ever re-baked or removed these Secrets. Owner agents never mount them (grant filtering screens them out by connection id), but fork gateways mount **every** Secret the replier owns — which is why only foreign-replier forks crashed.

**Leak** — a failed fork never reached `orchestrator.deleteFork`: `emitFailed` removed the fork from the in-memory `open` map *before* emitting `ForkFailed`, so the `ChannelTurnRelayed` saga's `closeFork` lookup missed (and `markCompleted` rejects `Failed` anyway). The Fork CR persisted, and the gateway pod — a bare `RestartPolicy: Always` Pod owner-refed to it — lived until the parent Agent was deleted. Forks also had no controller backstop reaper (Runs got one for exactly the api-server-crash case).

## Fixes

- **controller**: verify each credential's SDS key against the Secret's data at render time; a host whose file is missing renders as an **allow-only chain** with a warning instead of an unbootable bootstrap. The SDS data-key set is folded into `envoy-secrets-rev` so a healed Secret rolls the agent gateway.
- **controller**: `ForkMaxLifetime` (60 min) over-age reaper, mirroring `RunMaxLifetime`, running ahead of the terminal-phase short-circuit so leaked `Failed`/`Completed` CRs are reaped too.
- **api-server**: the fork failure path now deletes the Fork CR itself (`deleteFork` is 404-tolerant), so K8s GC cascades to the gateway pod immediately on failure.

## Verification

- New controller tests reproduce the stale-Secret shape (annotation without `sdsKey`, old-naming data keys) and pin the allow-only degradation, per-host partial degradation, the rev roll, and both reaper paths.
- New api-server tests pin `deleteFork` on status-Failed and create-failure paths, and that a failed delete still emits `ForkFailed`.
- `mise run check` and `mise run test` pass.

Existing leaked gateway pods self-heal on deploy: their Fork CRs are past `ForkMaxLifetime`, so the reaper deletes them on the first resync.

## Note / follow-up

Orphaned pre-cutover `platform-conn-*` Secrets still exist on installs (invisible in the UI, not referenced by any DB row, excluded from the legacy-secret migration and from #2198's drain gate). After this PR they merely degrade those hosts to allow-only with a warning; a sweep to remove them outright is worth a separate issue.

### DIFF
--- a/docs/architecture/agent-lifecycle.md
+++ b/docs/architecture/agent-lifecycle.md
@@ -1,6 +1,6 @@
 # Agent lifecycle
 
-Last verified: 2026-07-01
+Last verified: 2026-07-06
 
 ## Overview
 
@@ -179,6 +179,8 @@ Schedules are independent Postgres rows and survive Agent deletion as orphans un
 ## Forks
 
 Forks are the third durable concept in the bounded context (alongside Template and Agent). An `agent-fork` ConfigMap runs a derivative of an existing Agent with credential and env overrides. Unlike Agents, forks reconcile to a **Kubernetes Job** rather than a StatefulSet — they run to completion and are not woken, hibernated, or kept warm. This already matches the run-to-completion shape the target lifetime model intends for Agents. The interesting machinery is which secrets the fork can see and how its identity propagates upstream; see [security-and-credentials](security-and-credentials.md). A fork's Job inherits the parent Agent's image-pull Secret, so the kubelet pulls a private parent image without the fork ever seeing the credential.
+
+Fork teardown mirrors the Run executor's two-layer model: the api-server deletes the Fork CR when the fork's turn completes **or fails**, and K8s GC cascades to everything owner-refed to it — including the paired gateway pod, which as a bare always-restarting Pod has no terminal state of its own and lives exactly as long as the CR. A controller hard-lifetime reaper backstops any fork the api-server never cleaned up (crash mid-fork, lost in-memory state), reaping over-age CRs regardless of phase.
 
 ## Run executors (`dam-run`)
 
--- a/docs/architecture/security-and-credentials.md
+++ b/docs/architecture/security-and-credentials.md
@@ -1,6 +1,6 @@
 # Security and credentials
 
-Last verified: 2026-06-30
+Last verified: 2026-07-06
 
 ## Overview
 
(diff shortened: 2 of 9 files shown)


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `ed1aeff`

## PR #2344: fix: fork gateway Envoy crash on stale connection Secrets + leaked gateway pods
**Author:** pilartomas | **Branch:** fix/fork-gateway-sds-crash-and-leak → main | **Changes:** +288 −11 (9 files)

### Summary
Fixes two independent fork-gateway failures: (1) a fatal Envoy boot crash when a foreign-replier fork mounts a stale pre-cutover connection Secret whose referenced SDS file is absent — the controller now verifies each credential's SDS key against the Secret's data at render time and degrades the affected host to an allow-only chain (with a warning) instead of emitting an unbootable bootstrap, folding the SDS data-key set into `envoy-secrets-rev` so a healed Secret rolls the gateway; and (2) a gateway-pod leak where a failed fork never deleted its Fork CR — `emitFailed` now awaits `orchestrator.deleteFork` (404-tolerant) so K8s GC cascades to the paired gateway pod, backstopped by a new `ForkMaxLifetime` (60 min) over-age reaper that runs ahead of the terminal-phase short-circuit.

### Findings
- ✅ **Looks good:** SDS-key verification correctly degrades per-host — the partial-degradation path (`envoy.go`) keeps healthy hosts credentialed while only the missing-file host falls back to allow-only, and the single-host legacy shape is guarded symmetrically.
- ✅ **Looks good:** `sdsDataKeys` is folded into `envoySecretsRev` so a Secret gaining/losing its SDS key changes the rev and rolls the StatefulSet — the fix can't leave a healed gateway stuck on a stale allow-only chain. Pinned by `TestEnvoySecretsRev_SDSDataKeysRollExistingPods`.
- ✅ **Looks good:** `emitFailed` tears down the Fork CR on the failure path but still emits `ForkFailed` even when `deleteFork` throws (error is logged, not propagated) — pinned by the "still emits ForkFailed when the failure-path delete throws" test. `deleteFork` is 404-tolerant, so a double-delete (failure path + later saga) is safe.
- ✅ **Looks good:** the over-age reaper (`fork.go`) is correctly ordered **before** the `Failed`/`Completed` terminal short-circuit, so a leaked terminal CR whose gateway pod still crash-loops is reaped too — covered by both `OverAgeForkIsReaped` and `OverAgeTerminalForkIsReaped`.
- ✅ **Looks good:** thorough, targeted test coverage reproducing the exact stale-Secret shape (annotation without `sdsKey`, old sha8-era data keys) and both reaper paths; existing connection-Secret fixtures updated via `withHostSDS` so the new render-time guard doesn't silently degrade them.

### Verdict
APPROVE — clean, well-scoped fix for two real crash/leak bugs, with architecture docs (`agent-lifecycle.md`, `security-and-credentials.md`) updated in lockstep and comprehensive regression tests on every new path.

---

PR #2417, opened 2026-07-06
fix(helm): grant the controller update/delete on fork CRs

pilartomas:
## Problem

The fork over-age reaper merged in #2344 is failing on RBAC in live deployments — the controller logs show, on every resync:

```
reaping fork fork-…: forks.agent-platform.ai "fork-…" is forbidden: User "system:serviceaccount:…:…-controller" cannot delete resource "forks"
```

so leaked Fork CRs are never actually reaped. The controller Role grants only `get/list/watch` on `forks`; the reaper needs `delete`, and `ensureForkOwnerReference` (which stamps the parent-Agent owner ref onto the Fork CR so agent deletion cascades to in-flight forks) needs `update`.

## Change

Split the combined `agents, forks` read-only rule: `agents` stays `get/list/watch` (the api-server owns spec), `forks` gets `get/list/watch/update/patch/delete` — mirroring the `runs` grant, which was widened for exactly the same reaper pattern.

## Verification

- `mise run helm:check:lint` + `mise run helm:check:render` pass.
- After deploy: `kubectl auth can-i delete forks.agent-platform.ai --as=system:serviceaccount:<release-ns>:<controller-sa>` flips to yes; the reaper deletes the leaked Failed fork CRs (10 on the dev cluster dating back to June 24) on the next resync.

PR #2418, opened 2026-07-06
feat(api-server): diagnose and surface agent wake failures (#468)

### LINKED ISSUE #468: Investigate frequent "agent not ready" errors in Slack and surface the cause to the user
## Problem

Messaging an agent in Slack often fails with a raw error in the thread:

> Error: agent agent-0cbeedcb did not become ready within 120s

It happens often enough to be a real annoyance, not a rare edge case. And we don't actually know why. It could be a genuine bug — the agent stuck, crashing, or never getting woken — or it could just be a normal agent coming back from sleep where the cold start runs past the deadline. The error doesn't distinguish those, so neither we nor the user can tell which it is.

That's the core gap: we're seeing a frequent failure and guessing at the cause. The user just gets an internal-looking timeout string and has to retry blindly.

## Proposed solution

Two parts.

1. **Find out why it happens so often.** Investigate enough real occurrences to land on a root cause: is this an actual bug in waking/starting agents, or is it expected cold-start latency from sleep regularly exceeding the wait window? The outcome is a clear answer, and whatever follow-up that answer points to (a fix, or a longer/smarter wait budget).

2. **Surface the real reason to the user.** Once we know what's happening, replace the raw timeout string with a message that reflects the actual cause and what to do — "still warming up, try again shortly" vs. "this agent failed to start." The user should understand which situation they're in instead of seeing an internal error.

### Scope

Starts as an investigation. The first step is diagnosis; the surfacing improvement follows from what we learn. The same wake-and-respond path backs other channels, so findings and fixes likely apply beyond Slack.

pilartomas:
Closes #468.

## Investigation findings

The raw `Error: agent X did not become ready within 120s` was undiagnosable by design: `ensureReady` polled only the boolean Ready condition, threw an untyped error, and logged nothing — and the Slack catch wrote to bare stderr, which bypasses pino, so failures never even reached the log store. The controller-published conditions on the very CR being polled already carry the full diagnosis (Ready reason, pod termination cause, gateway readiness, reconcile errors) and were discarded.

Live-cluster forensics (dev, 5-day log retention): every recent Slack failure was the fork-crash bug fixed in #2344; zero owner-path wake timeouts in-window. A healthy cold start measures ~28s pod-side, so the 120s budget is structurally fine — but sharp legitimate >120s modes exist: an agent image on `:latest` (pullPolicy Always; the startup probe budgets 1s×120 exactly for a cold pull, and the wake window starts *before* pod creation, so a sanctioned slow pull can outrun it — two dev-cluster agents run `:latest` today), np-gate init crashloop cycles, RWO volume attach, reconcile-queue backoff, and a wedged ambient mesh. Since the path was blind, making the next occurrence diagnosable **is** the investigation deliverable; the historical occurrences predate log retention.

## Changes

- **Classification** (`modules/agents/domain/wake-failure.ts`): at the deadline, one final CR read classifies the failure — `hibernated-not-scaled`, `agent-pod-failed:<termination token>`, `agent-pod-not-ready` (transient), `gateway-not-ready` (transient), `reconcile-error`, `not-found`, `unknown` — and `ensureReady` throws a typed `AgentWakeTimeoutError` with a humanized message. A CR that turned Ready right at the deadline counts as success.
- **Instrumentation**: `agent.wake.begin/ready/timeout` pino lines with duration and the condition snapshot; Slack/Telegram/proxy failure paths log through the logger instead of stderr, so everything lands in the log store.
- **Slack UX**: requester-only "waking the agent" ephemeral on cold starts (via a new optional `onWaking` hook on `ensureReady`); a transient timeout posts a still-starting note and waits **one** more window before answering (a healthy-but-slow start no longer loses the turn); hard causes get human copy mapped from the classification — never the internal string or raw controller messages.
- **Telegram**: previously surfaced nothing; now replies with the shared copy on wake failures.
- **Audit/usage**: `ChannelTurnRelayed` carries an optional low-cardinality `reason` (`wake-timeout:*`, `fork-failed:*`, `acp-error`), projected onto the `channel.turn` audit line and the usage payload — failed turns become queryable after the fact.
- **Other callers**: skills reachability maps hard causes to `PRECONDITION_FAILED`; ACP/terminal/SSH relays carry the cause kind in the WS close reason; tRPC/import proxies add it to their 502 bodies.
- **Docs**: agent-lifecycle §Wake + channels error-surfacing notes.

Companion hotfix (separate PR #2417): the controller RBAC grant that unblocks the fork reaper from #2344.

## Verification

- New unit suites: classification decision table; `ensureReady` under fake timers with an in-memory K8s stub (per-cause timeouts, late-ready, single-flight + joiner `onWaking`); the real Slack worker driven through the fake gateway (cold ephemeral, one-retry-then-answer, hard-cause copy, reason on the turn event; asserts the internal string never reaches the thread); audit-saga reason projection.
- `mise run check` and `mise run test` pass repo-wide.
- Not yet exercised on a live cluster: the plan's forced-`ImagePullFailure` / slow-pull recipes (needs a throwaway cluster + Slack app). The wake logs and `channel.turn` reasons will confirm behavior on dev after deploy.

## Out of scope

Exposing the cause enum in `api-server-api` for the UI agent page; genericizing non-wake relay error copy; np-gate-specific condition naming in the controller.


packages/api-server/src/__tests__/unit/agents-repository-ensure-ready.test.ts
```diff
@@ -0,0 +1,284 @@
...
+      (e: unknown) => e,
+    );
+    expect(isAgentWakeTimeoutError(err)).toBe(true);
+    if (isAgentWakeTimeoutError(err)) {
+      expect(err.failure.kind).toBe("not-found");
+    }
+  });
+
+  it("late ready at the deadline counts as success", async () => {
+    const { repo, store, lines } = harness([agentObj("a1", HIBERNATED)]);
```
dam-code-guardian: 🟢 **Suggestion:** `repo` is destructured but never used — this test builds its own `repo2` from the wrapped client below. Drop it from the destructure (also flagged by `github-code-quality`).

```suggestion
    const { store, lines } = harness([agentObj("a1", HIBERNATED)]);
```

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `02a34c8`

## PR #2418: feat(api-server): diagnose and surface agent wake failures (#468)
**Author:** pilartomas | **Branch:** feat/wake-failure-diagnosis → main | **Changes:** +1088 −44 (24 files)

### Summary
Turns the previously-undiagnosable `agent X did not become ready within 120s` into a typed, classified wake failure: at the deadline a final CR read classifies the cause (hibernation-not-scaled, pod-failed with termination reason, still-progressing, gateway-not-ready, reconcile-error, not-found, unknown), `ensureReady` throws a typed `AgentWakeTimeoutError`, and every caller (Slack, Telegram, ACP/SSH/terminal relays, tRPC/import proxies, skills) surfaces it in its own idiom with human copy that never leaks internal strings. Adds `agent.wake.begin/ready/timeout` structured logs, an `onWaking` cold-start hook, and a low-cardinality `reason` on `ChannelTurnRelayed` projected onto the audit + usage trails. Well-scoped, thoroughly tested, and both architecture pages updated in lockstep.

### Findings
- 🟢 **Suggestion:** Unused `repo` in the destructure — this test uses its own `repo2` built from the wrapped client, so `repo` from `harness(...)` is never read (`packages/api-server/src/__tests__/unit/agents-repository-ensure-ready.test.ts:252`). Also flagged by `github-code-quality`.
- ✅ **Looks good:** Single-flight joiner gets the `onWaking` signal too (`agents-repository.ts:216-218`) — a caller joining an in-flight wake is also waiting on the cold start, so telling its user is correct.
- ✅ **Looks good:** `classifyWakeFailure` precedence — reconcile-error wins over pod reasons, and pod-failure reasons are gated to a known termination set with plain `PodNotReady` falling through to the transient `agent-pod-not-ready` class (`wake-failure.ts:918-944`). Covered by an explicit precedence test.
- ✅ **Looks good:** Late-ready-at-the-deadline is treated as success rather than failing a turn that actually works (`agents-repository.ts:238-247`).
- ✅ **Looks good:** Transient-timeout retry is guarded to a single extra window and the retry's `onWaking` is suppressed so the wake isn't re-announced (`slack.ts:288-300, 360-379`).
- ✅ **Looks good:** User-facing copy is derived from the classified cause and never interpolates raw controller messages (which can carry resource names) — enforced by a dedicated leak test (`wake-failure-copy.ts`, `wake-failure.test.ts:690-698`).
- ✅ **Looks good:** WS close reasons carry only the short cause `kind` with an explicit note that close reasons cap at 123 bytes (`acp-relay.ts`, `ssh-relay.ts`, `terminal-relay.ts`).
- ✅ **Looks good:** `onWaking` is an optional param, so all existing `ensureReady` callers remain source-compatible; failure `reason` is spread conditionally onto events/payloads so success turns stay unchanged.

### Verdict
APPROVE — a well-structured, thoroughly-tested diagnosability improvement with clean layering (domain classification vs. channel presentation) and lockstep doc updates; the only nit is a harmless unused test binding.

---

PR #2422, opened 2026-07-07
feat(observability): join harness, gateway, and ext_authz spans in one trace

pilartomas:
## Problem

With the telemetry backend enabled, agent traces arrive fragmented: the harness's `claude_code.*` spans, the gateway's egress spans, and the api-server's egress-approval (ext_authz Check) spans each sit in separate traces. Two causes:

1. The TLS-terminating credential chains — the one Envoy layer that sees the agent's decrypted `traceparent`, and the layer the ext_authz Check fires from for TLS egress — carried no tracing provider. The outer listener's tracer only ever sees the CONNECT tunnel, so nothing could join.
2. On custom (non-Anthropic) model backends, the harness fronts the upstream with the in-pod model gateway and rewrites `ANTHROPIC_BASE_URL`, which makes Claude Code silently stop sending `traceparent` at all ([docs](https://code.claude.com/docs/en/monitoring-usage#traces-beta)).

## Change

**Helm** — set `CLAUDE_CODE_PROPAGATE_TRACEPARENT=1` in the agent telemetry env block, so the harness keeps emitting `traceparent` on custom backends (already the default on genuine Anthropic backends).

**Controller (Envoy bootstrap)** — give the TLS-terminating chains a tracing provider **when every credential on the chain is header-injected**:

- The chain HCM extracts the agent's W3C `traceparent` (parent-based join), so its span lands in the harness's trace.
- Envoy's ext_authz gRPC client spawns a child span for the Check and injects `traceparent` into the RPC metadata (verified in Envoy source: `ext_authz.cc` passes the active span; `async_client_impl.cc` spawns the child and calls `injectContext`). The api-server's ext_authz gRPC server runs under `@opentelemetry/instrumentation-grpc`, which extracts inbound metadata automatically — so the Check span joins the same trace with **no api-server change**.
- Chain spans suppress the path tag (`max_path_tag_length: 1`): agent-authored paths/queries can hold agent-side secrets (e.g. presigned URLs), and Envoy has no query-stripper for span tags. Per-request detail stays in the query-stripped access log, joinable via the `x-request-id` span tag.
- Chains with a **query-param credential stay untraced** — post-injection `:path` carries the live credential, which is the original reason chains were untraced; that protection is preserved exactly where the risk exists.

**Docs** — `logging.md` gateway-telemetry section updated to the new tracing posture and to describe actual traceparent handling on egress (the strip only ever existed on the plain-HTTP fallthrough route; TLS-intercepted requests forward trace context — same trace ID — to the upstream); `observability.md` documents the propagation posture.

## Result

For a model/API request on a header-credential chain: `claude_code.llm_request` (harness) → `terminate_<chain>` span (gateway) → `async envoy.service.auth.v3.Authorization.Check egress` (gateway) → `Check` SERVER span (api-server), all one trace. Subprocess `TRACEPARENT` inheritance also now works on custom backends.

## Verification

- **Verified live in a dev environment**: harness, gateway chain, and api-server Check spans land in a single trace in the telemetry UI.
- `mise run check` and controller `go test ./...` pass; new tests pin the invariants: tracing provider on outer HCM + each header-credential chain, none on query-param chains, chain path tags suppressed, rendered bootstrap stays valid YAML.
- `helm:check:lint` / `helm:check:render` (renders with `clickstack.enabled=true`) pass.
- Envoy-side behavior (context injection on ext_authz, parent-based sampling, W3C-only propagation) verified against Envoy source; all predates the pinned `envoy:distroless-v1.37.2`.
- Deployment note: the bootstrap is static per gateway pod — after rolling the new controller image, existing gateway pods must be recreated to pick up the new config.


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `064e165`

## PR #2422: fix(helm): enable traceparent propagation for claude-code agents
**Author:** pilartomas | **Branch:** fix/claude-code-propagate-traceparent → main | **Changes:** +7 −1 (2 files)

### Summary
Sets `CLAUDE_CODE_PROPAGATE_TRACEPARENT=1` in the shared agent-telemetry env block so W3C trace-context propagation stays on even when the harness fronts a custom (non-Anthropic) model upstream — where Claude Code would otherwise disable it — and documents the propagation posture in the observability architecture page. Forcing it on is safe because the gateway strips `traceparent`/`tracestate` before any external upstream, so internal hops (harness spans, subprocess telemetry, gateway egress spans) join one trace without leaking context outward.

### Findings
- ✅ **Looks good:** The new env var is inserted cleanly among its sibling `CLAUDE_CODE_*` / `OTEL_*` entries in `platform.agentTelemetry.env`, with correct YAML shape and a comment that captures the *why* (`_helpers.tpl:338-342`). Gating is unchanged — the helper is still rendered only under `clickstack.enabled` + the per-template telemetry flag, so templates without telemetry are unaffected.
- ✅ **Looks good:** The safety rationale is verifiable in-repo: `logging.md:72` (gateway telemetry) independently documents that the gateway strips `traceparent`/`tracestate` before a request reaches an external upstream, corroborating the "never outward" claim in the PR body and the new doc bullet.
- 🟢 **Suggestion:** None blocking — the change is minimal and self-consistent.

✅ No documentation drift detected. `observability.md` was updated in the same PR with a new "Trace context propagates internally, never outward" bullet describing the propagation posture at architecture altitude, its `Last verified:` header was bumped 2026-07-02 → 2026-07-07 (PR date), the cross-reference to [logging — gateway telemetry](logging.md#gateway-telemetry) resolves (heading present at `logging.md:61`), and no ADR references or volatile content (exact package/path/helm-tree choices) leaked into the page — `traceparent`/`tracestate` are W3C protocol terms and the env-var name appears only in the helm comment, not the doc.

### Verdict
APPROVE — A minimal, correct, well-documented telemetry-config fix; the propagation is forced on safely given the gateway's pre-egress stripping, and the architecture doc was updated in lockstep.

---

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `c16f76f`

## PR #2422: fix(helm): enable traceparent propagation for claude-code agents
**Author:** pilartomas | **Branch:** fix/claude-code-propagate-traceparent → main | **Changes:** +92 −16 (5 files)

### Summary
Beyond the prior helm-only change, this PR now gives the gateway's TLS-terminating credential chains an OpenTelemetry tracing provider **only when every credential on the chain is header-injected** (`envoy.go` `HasQueryParamCredential()` gate), so those chains join the agent's decrypted `traceparent` into one trace across harness → gateway → ext_authz → api-server. Chain spans are path-blind (`max_path_tag_length: 1`) to keep agent-authored paths/queries out of span tags, and query-param-credential chains stay untraced — preserving the exact protection that originally kept all chains untraced. The two architecture pages were rewritten in lockstep to describe the new posture.

### Changes since last review
Previous HEAD: `064e165` (2026-07-07T04:53:29Z) — verdict APPROVE

- 🆕 **New:** Envoy bootstrap now renders a tracing provider on header-credential TLS-terminating chains, gated by `HasQueryParamCredential()` (`packages/controller/pkg/reconciler/envoy.go`), with path-tag suppression and query-param chains excluded.
- 🆕 **New:** `logging.md` gateway-telemetry section rewritten from "no spans on credential-injection chains" to the header-only-traced / path-blind posture, and the trace-context-stripping bullet clarified for TLS-intercepted vs plain-HTTP egress.

### Findings
- ✅ **Looks good:** The `HasQueryParamCredential()` gate is a correct, conservative narrowing: it traces a chain only when *every* credential injects into a header, and any query-param credential on the chain forces it untraced (`envoy.go:127-137`). This preserves the original safety invariant (post-injection `:path` can carry a query-param credential and Envoy has no span-tag query stripper) exactly where the risk exists, while enabling the trace-join everywhere it's safe.
- ✅ **Looks good:** The template gate `{{- if and $.OTel.Traces (not $chain.HasQueryParamCredential) }}` reuses the pre-existing outer-HCM tracing provider config (present in base `origin/main`), so this extends an established subsystem rather than introducing a new one; the `otel_export` cluster and `SamplingPercent`/`GRPC`/`TracesURI` fields are all already wired.
- ✅ **Looks good:** Test invariants pin the behavior tightly — `TracingOnHeaderCredentialChains` asserts exactly 3 `OpenTelemetryConfig` (outer + 2 chains) with 2×`max_path_tag_length: 1` and 1×`256`, and `TracingNotOnQueryParamChains` asserts a query-param chain yields only the outer provider (`envoy_test.go`). Path-tag suppression on chains is verified, not just assumed.

✅ No documentation drift detected. Both `logging.md` and `observability.md` were updated in the same PR to reflect the new tracing posture — `logging.md`'s gateway-telemetry bullets now describe header-only-traced, path-blind chains and the TLS-vs-plain-HTTP trace-context handling, and `observability.md` gained a "Trace-context propagation stays on" bullet. Both `Last verified:` headers are bumped to the PR date (2026-07-07), the cross-reference `logging.md#gateway-telemetry` resolves (heading at `logging.md:61`), and no ADR references, `Motivated by:` sections, or volatile content (exact package/path/helm-tree symbols) leaked into the prose — `traceparent`/`tracestate`/`max_path_tag_length` are W3C/Envoy protocol terms and the field-level detail stays in the code.

### Verdict
APPROVE — The chain-tracing extension is correctly gated to header-only-credential chains, path-blind to avoid leaking agent-authored paths into span tags, tightly test-pinned, and the architecture docs were updated in lockstep.

---

PR #2655, opened 2026-07-07
fix(agent-runtime): don't crash the pod on harness stdin EPIPE

jezekra1:
## Problem

Opening a **chat** session against an agent image that ships platform-base's no-chat-harness stub (a supported configuration — e.g. terminal-only agents) crashes the whole agent pod:

```
harness-chat: this agent image does not provide a chat harness
Error: write EPIPE
    at Object.send (file:///app/dist/server.js:6375:21)
    at handleClientMessage (file:///app/dist/server.js:7170:9)
Emitted 'error' event on Socket instance ...
```

agent-runtime is PID 1's child, so the unhandled error takes down the container — killing the pod-service, terminal sessions, and any in-flight work.

## Root cause

`createChildAgentProcess`'s `send()` guards `child.stdin.writable`, but the guard is racy: when the harness exits, a write dispatched during pipe teardown fails **asynchronously** with EPIPE, emitted as an `'error'` event on the stdin stream — which had no listener. The existing `child.on("error", …)` only covers spawn failures.

## Fix

Attach a stdin `'error'` handler that logs and drops, mirroring what `modules/ssh.ts` already does for the sshd child. Session teardown is unchanged — the `exited` cleanup already closes engaged channels with `1011 "agent exited"`.

## Tests

- New regression test reproduces the crash deterministically (child closes fd 0 while alive → dispatched write EPIPEs). Fails with an unhandled `write EPIPE` without the fix; passes with it.
- `mise run agent-runtime:test` (161 tests) and `mise run check` green.

### DIFF
--- a/packages/agent-runtime/src/__tests__/unit/create-child-agent-process.test.ts
+++ b/packages/agent-runtime/src/__tests__/unit/create-child-agent-process.test.ts
@@ -30,4 +30,42 @@ describe("createChildAgentProcess", () => {
 
     expect(lines).toEqual(['{"id":1,"a":"hello"}', '{"id":2}']);
   });
+
+  it("survives EPIPE when the harness closes stdin while alive", async () => {
+    // The no-chat-harness stub case: the child's stdin reader is gone but the
+    // process hasn't exited, so send()'s writable guard still passes and the
+    // dispatched write fails async with EPIPE. Without a stdin error handler
+    // that's an unhandled 'error' event — it would crash this test process.
+    const proc = createChildAgentProcess({
+      command: [
+        process.execPath,
+        "-e",
+        // fs.closeSync, not process.stdin.destroy(): node never really closes
+        // stdio fds through the stream API, so only the raw close makes the
+        // parent's pipe reader-less.
+        `require("fs").closeSync(0); console.log("ready"); setTimeout(() => {}, 2000);`,
+      ],
+      workingDir: process.cwd(),
+    });
+
+    // "ready" on stdout ⇒ the child has already destroyed its stdin.
+    await new Promise<void>((resolve) => proc.onLine(() => resolve()));
+
+    proc.send({ jsonrpc: "2.0", method: "ping" });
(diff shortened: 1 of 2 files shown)


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `32f0fda`

## PR #2655: fix(agent-runtime): don't crash the pod on harness stdin EPIPE
**Author:** jezekra1 | **Branch:** fix/acp-harness-stdin-epipe → main | **Changes:** +46 −0 (2 files)

### Summary
Attaches a `'error'` listener to the child harness's `stdin` stream so a write that EPIPEs asynchronously during pipe teardown (e.g. an image shipping the no-chat-harness stub, whose stub exits immediately) is logged and dropped instead of surfacing as an unhandled `'error'` event that would take down agent-runtime (PID 1) and the whole pod. Adds two regression tests, one of which deterministically reproduces the crash.

### Findings
- ✅ **Looks good:** The fix correctly mirrors the established pattern in `modules/ssh.ts:162`, where the sshd child's stdin already has an error handler. The `send()` writable-guard race is accurately diagnosed — a write dispatched while `writable` is still true can EPIPE async after the reader is gone.
- ✅ **Looks good:** The `survives EPIPE …` test reproduces the crash deterministically via `fs.closeSync(0)` (raw fd close, not `process.stdin.destroy()`), with a correct comment explaining why the stream API wouldn't trigger the bug. It would fail with an unhandled `write EPIPE` pre-fix.
- 🟢 **Suggestion:** `ssh.ts` guards all three of the child's streams (`stdin`/`stdout`/`stderr`); this fix guards only `stdin`. A `stdout` read error on the harness pipe would still be unhandled. Risk is low here (stdout is read via `readline` and stderr is `inherit`ed), so this is optional — but attaching a `stdout` `'error'` handler alongside would close the gap symmetrically. (`create-child-agent-process.ts:44`, outside the diff hunk — summary-only.)

### Verdict
APPROVE — Small, correct, well-tested defensive fix that removes a real pod-crash vector and follows an existing in-repo pattern.

---

Issue #2040, opened 2026-06-26
Show whether a terminal session is busy or idle in the sessions list


jjeliga:
## Problem

#1081 added a status dot to every row in the sessions list — **Working**, **Waiting for you**, or **Needs your approval** — so you can see at a glance what each session is doing without opening it. That works for chat and channel (Slack/Telegram) sessions.

Terminal sessions are the gap. A terminal session is a raw interactive shell, and unlike a chat conversation it has no clear "the agent finished its turn" moment. Because there's no such signal, the list can't tell whether a terminal session is busy or idle, so **every terminal session shows the same neutral grey "Terminal" dot** regardless of what's happening inside it.

In practice that means:

- You start a long-running command in a terminal session (a build, a test run, a deploy), switch away, and the list gives you no hint that it's still churning.
- A terminal session sitting at an idle prompt, ready for you, looks identical to one that's mid-command.
- The neutral dot is honest — it doesn't pretend to know — but it doesn't answer the question the feature was built for: *which of my sessions need me, and which are still working?*

So for terminal sessions you're back to the original problem #1081 set out to solve: you have to open the session and watch it to know its state. As more real work happens in terminal sessions rather than chat, and especially when several are running at once, this blind spot is exactly where a glanceable indicator would pay off.

## Desired outcome

A terminal session's row would tell you, at a glance and without opening it, whether it's actively doing something or sitting idle — the same value #1081 already gives chat sessions, updating live for sessions you don't have open.

## Open question — is this even achievable?

This is the crux, and it's unresolved. Unlike a chat turn, an interactive shell has no built-in "I'm done" boundary, so it's genuinely unclear whether any reliable way exists to distinguish a busy terminal from an idle one. **The first task here is discovery** — investigate whether a dependable signal exists at all and whether surfacing it is worth the cost. It is an acceptable conclusion of this ticket that there is no robust signal, and that the neutral dot remains the most honest thing we can show.

### Dependencies
- Builds on #1081, implemented in #2039 — the neutral terminal dot shipped in that PR is the placeholder this issue would replace if a signal turns out to exist.

### Out of scope
- Fine-grained progress (which command is running, how far along) — if any signal is surfaced at all, a coarse busy/idle distinction is enough, consistent with #1081.

kapetr:
After some exploration it should be achievable, with one caveat.

[...]

PR #2733, opened 2026-07-08
feat(agent-runtime): keep working terminal sessions alive after detach

### LINKED ISSUE #2731: Keep a working terminal session alive when you switch away
## Problem

Switching away from a terminal session silently kills it about 30 seconds later — including any command still running in it. Start a long build or test run in a terminal, jump to another session, and the work is aborted; nothing warns you. It even *looks* like it kept running, because the transcript is saved continuously and reattaching shows everything up to the kill.

The 30-second window exists for tab refreshes, not background work: terminal sessions are deliberately single-viewer and ephemeral today. But as more real work happens in terminals, "leaving the tab aborts my build" is surprising and costly — and it makes a busy/idle indicator for detached terminals (#2040) impossible, since nothing survives detaching.

## Proposed solution

Reap detached terminals by **idleness**, not by viewer loss: a terminal with no viewer is killed only once it has also been quiet for a while (no output for N minutes). A terminal actively producing output — a running build, a streaming response — survives switching away and can be reattached later, live.

In other words: liveness detection moves from the viewer side of the terminal (is someone watching?) to the harness side (is it doing something?).

A working terminal keeps its agent awake (this already follows from how idle detection works), so the idle timeout is also the bound on how long a forgotten terminal can hold an agent up. N should be generous enough for real work but small enough that an abandoned prompt doesn't keep an agent awake indefinitely; possibly operator-configurable.

### Dependencies

- Enables the detached-terminal case of #2040 (busy/idle indicator) — the same activity signal drives both the reaper and the indicator.
- Related to #1373 (keep an agent awake during long-running background work) — this is the terminal-session flavor of that tension.

### Out of scope

- Multi-viewer terminals (sessions stay one-viewer-at-a-time; this only changes when an unwatched terminal is cleaned up).
- Warning the user before an idle terminal is reaped.

kapetr:
Switching away from a terminal session used to kill its PTY 30 seconds later — aborting any command still running in it, silently (verified live: a mid-prompt terminal died exactly 30s after detach; the saved transcript makes it look like it kept running).

This changes the reap rule from viewer loss to harness quietness: after the 30s detach grace (kept for tab-refresh reattach), a detached PTY is killed only once the harness has produced no output for five minutes. Working TUIs repaint continuously (spinner, streaming), so in-flight work survives switching away and can be reattached live — while an abandoned idle prompt is still cleaned up. Liveness detection moves from the viewer side of the terminal (is someone watching?) to the harness side (is it doing something?).

- `lastOutputAt` stamped per PTY slot on every output byte
- on detach, the grace timer leads into an idle check that re-arms while output keeps flowing
- reattach path unchanged (cancels the timer, replays scrollback) — the replay window is now the slot's lifetime, not 30s
- pod hibernation already cooperates: a live slot reports the pod non-idle, and the five-minute quiet window is well under the 1h hibernation idleTimeout, so cost is bounded
- architecture pages updated to the new rule

The same output-activity signal can later drive the busy/idle indicator for detached terminals (#2040).

Closes #2731

https://claude.ai/code/session_018tmc8chchD16eZd4ba18qK


packages/agent-runtime/src/server.ts
```diff
@@ -181,12 +181,18 @@ const trpcHandler = createHTTPHandler({
   maxBodySize: TRPC_MAX_BODY_SIZE,
 });
 
+// A detached PTY is reaped on harness quietness, not viewer loss, so in-flight
+// work survives switching away. The grace stays short for tab-refresh reattach.
+const PTY_DETACH_GRACE_MS = 30_000;
+const PTY_IDLE_REAP_MS = 5 * 60_000;
```
dam-code-guardian, line 187: 🟢 **Suggestion:** Reaping keys on harness *output*, so a busy-but-silent job (long buffered compile, `sleep`, quiet network wait) that emits no bytes for `PTY_IDLE_REAP_MS` is reaped as if idle. The PR body explicitly accepts this tradeoff — noting only for visibility, not as a blocker.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `a3c75a4`

## PR #2733: feat(agent-runtime): keep working terminal sessions alive after detach
**Author:** kapetr | **Branch:** feat/2731-terminal-idle-reap → main | **Changes:** +30 −5 (3 files)

### Summary
Re-keys detached-PTY reaping from viewer loss to harness quietness: after the 30 s tab-refresh grace, a detached PTY is killed only once the harness has produced no output for five minutes (`lastOutputAt` stamped per output byte, re-armed idle check). In-flight TUIs that repaint survive switching away; abandoned idle prompts are still cleaned up. Architecture pages are updated in the same PR.

### Findings
- ✅ **Looks good:** Timer lifecycle is sound — `graceTimer` is cleared on reattach (`clearTimeout` in the RESIZE/existing branch) and in `killPtySlot`; the re-arm reassigns after the spent timer fires (no leak); an orphaned timer after `pty.onExit` finds no slot and returns early via the `!slot.pty` guard.
- ✅ **Looks good:** `reapPtySlotIfIdle` correctly no-ops when a client is attached (`slot.client`) or the PTY is gone, and recomputes the remaining quiet window on each re-arm rather than resetting the full 5 min.
- 🟢 **Suggestion:** Reaping keys on harness *output*, so a genuinely-busy-but-silent job (a long buffered compile, a `sleep`, a quiet network wait producing no bytes for 5 min) would be reaped as if idle. The PR body explicitly reasons about this tradeoff ("abandoned idle prompt is cleaned up"), so this is an accepted design choice — flagging only for visibility, not as a blocker.
- 🟢 **Suggestion:** The PTY slot lifecycle (grace → idle re-arm → reap, and the reattach-cancels-timer path) is now non-trivial and has no unit coverage in `packages/agent-runtime/src/__tests__/`. A fake-timer test asserting (a) reap after sustained quiet, (b) survival while `onData` keeps stamping `lastOutputAt`, and (c) timer cancellation on reattach would lock in the new behavior. Pre-existing 30 s reap was untested too, so this is net-new value rather than a regression.

_Both affected architecture pages (`agent-lifecycle.md`, `platform-topology.md`) were updated in the same PR to describe the new idleness-based reap rule, and both had their `Last verified:` header bumped to 2026-07-08. No stale "30 s on viewer loss" wording leaks into other architecture pages; no ADR references or volatile content introduced._

### Verdict
APPROVE — Correct, well-scoped behavior change with sound timer lifecycle and architecture docs kept in sync; the two suggestions are non-blocking.

---

packages/agent-runtime/src/server.ts
```diff
@@ -296,17 +319,20 @@ function attachPty(
...
       ptyLog(sessionId, `spawned PTY (${cols}x${rows})`);
 
       pty.onData((data) => {
+        slot.lastOutputAt = Date.now();
         slot.headless.write(data);
         if (slot.client?.readyState === 1)
           slot.client.send(encodeDataFrame(OP_OUTPUT, data));
       });
       pty.onExit(({ exitCode }) => {
         ptyLog(sessionId, `exited ${exitCode}`);
```
dam-code-guardian, line 334: 🆕 **New (looks good):** Clearing `slot.graceTimer` here on `onExit` — together with the matching guard in `detach` — closes the window where a spent or racing reap timer could fire against a slot that's already dead or has been re-armed by a reattach. Good hardening.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `a1268d0`

## PR #2733: feat(agent-runtime): keep working terminal sessions alive after detach
**Author:** kapetr | **Branch:** feat/2731-terminal-idle-reap → main | **Changes:** +37 −11 (3 files)

### Summary
Re-keys detached-PTY reaping from viewer loss to harness quietness: after the 30 s tab-refresh grace, a detached PTY is killed only once the harness has produced no output for five minutes. This re-review covers the new commit, which hardens the detach path — consolidating the WebSocket `error`/`close` handlers into a single `detach` closure and adding `clearTimeout` guards to prevent orphaned reap timers.

### Changes since last review
Previous HEAD: `a3c75a4` (2026-07-08T10:16:39Z) — verdict APPROVE

- ✅ **Fixed:** Latent gap where a socket that emitted `error` but never `close` would null out `slot.client` without arming any reap timer, leaking the PTY slot indefinitely — the new `detach` closure is bound to **both** `error` and `close` (`server.ts:236-249`).
- 🆕 **New:** `onExit` and `detach` now `clearTimeout(slot.graceTimer)` before proceeding, closing the window where a spent/racing timer could fire against a dead or re-armed slot (`server.ts:243`, `server.ts:334`).
- 🔁 **Still present:** Reaping keys on harness *output*, so a busy-but-silent job (long buffered compile, `sleep`, quiet network wait) emitting no bytes for 5 min is reaped as idle — accepted design tradeoff per the PR body (`server.ts:214`).
- 🔁 **Still present:** PTY slot lifecycle (grace → idle re-arm → reap, reattach-cancels-timer) still has no unit coverage in `packages/agent-runtime/src/__tests__/`.

### Findings
- 🟢 **Suggestion:** The detach-path hardening is now even more worth a fake-timer unit test — asserting reap after sustained quiet, survival while `onData` stamps `lastOutputAt`, timer cancellation on reattach, and that an `error`-without-`close` socket still arms the reap. (`packages/agent-runtime/src/__tests__/`)
- 🟢 **Suggestion:** Harness-output-based liveness will reap a busy-but-silent job as idle — accepted tradeoff per the PR body, flagged for visibility only (`server.ts:214`).
- ✅ **Looks good:** The `detach` consolidation correctly guards `slot.client !== ws` and clears any existing `graceTimer` before re-arming, so rapid disconnect/reconnect cycles can't stack timers.

_The incremental commit touches only `server.ts` and introduces no subsystem-behavior change beyond what the architecture pages already describe. Both affected pages (`agent-lifecycle.md`, `platform-topology.md`) carry `Last verified: 2026-07-08`, accurately describe the idleness-based reap rule, contain no ADR references or volatile content, and no stale "30 s on viewer loss" wording leaks into other architecture pages._

### Verdict
APPROVE — Clean hardening of the detach/cleanup path (fixes a real slot-leak on error-only sockets, prevents orphaned timers); docs remain in sync. Suggestions are non-blocking.

---

jezekra1 (approved):

PR #2419, opened 2026-07-06
feat(helm): enable agent telemetry for the nous template

Tomas2D:
## Summary

Adds `telemetry: true` to the `nous` agent template so Nous campaigns export token/cost/duration telemetry to the ClickStack backend, exactly like the `claude-code` template — no image changes required.

## Why this works (investigated upstream nous v0.4.0)

The nous image is built `FROM platform-claude-code`, and the telemetry rail is entirely env-driven: with `clickstack.enabled` + the template flag, the pod receives `CLAUDE_CODE_ENABLE_TELEMETRY=1` and the `OTEL_*` exporter env. Both of Nous's dispatch paths pass that env through to the harness:

- **SDK path (default):** `claude-agent-sdk`'s subprocess transport constructs the child env as `{**os.environ, ...}` (filtering only `CLAUDECODE`), and Nous's `SDKDispatcher` passes no `env` override in `ClaudeAgentOptions` (`orchestrator/sdk_dispatch.py`). The SDK additionally propagates active OTel trace context into the child.
- **`claude -p` fallback:** `subprocess.run` without `env=` — inherits the pod env (`orchestrator/cli_dispatch.py`).
- **No interference:** the orchestrator never scrubs `os.environ`, and the campaign `.claude/settings.json` Nous writes is a permissions allowlist only — no `env` block, no telemetry keys (`orchestrator/settings_template.py`). The image's `nous-model-gateway.sh` doesn't touch `OTEL_*` either.

Attribution needs nothing from Nous: the collector's Istio waypoint stamps the trusted `platform.agent.id` (#2046), so campaign spend lands under the right agent/owner and becomes readable through the owner-scoped read path (#2029).

## Verification

- `mise run helm:check:lint` and `mise run helm:check:render` pass (the render task already exercises `clickstack.enabled=true`).
- Targeted render confirms the `nous.yaml` template entry now carries the same OTLP env block as `claude-code`.

## Caveats

- **Runtime smoke test still advisable** on a `clickstack.enabled` cluster: run a one-iteration campaign and confirm records with the nous agent's `platform.agent.id` land in ClickHouse. Static analysis covers the env plumbing; this confirms the collector round trip.
- **Gate summaries stay dark**: they go through the OpenAI-format endpoint, not the Claude CLI, so they don't export. Noted in the README.
[...]

Issue #2046, opened 2026-06-26
feat: Tag agent telemetry with a trusted identity the agent can't forge


pilartomas:
## Problem

Telemetry arriving in the datastore carries only what the agent put in it, and agents run untrusted code. Nothing stops an agent from labelling its telemetry as belonging to a *different* user's agent. The read path that lets a user see their own agents' telemetry (#2029) decides ownership from the identity on each record — so if that identity is whatever the agent claimed, the ownership check can be subverted: a malicious agent could make its data appear under another user, or surface its own data as someone else's. There is no trustworthy link between a telemetry record and the agent that actually produced it.

## Proposed solution

Establish the producing agent's identity from the trusted path its telemetry already travels — its gateway — rather than from the label the agent put on the data. Specifically, place an **Istio L7 waypoint** in front of the collector: the waypoint reads the verified mesh identity of the gateway that forwarded the telemetry (which maps one-to-one to the agent) and injects that identity into the request as it is ingested, overriding anything the agent set. After this lands, every telemetry record is reliably tied to the agent that produced it (and therefore its owner), so the owner-scoped reads in #2029 are safe: a user only ever sees telemetry from agents they actually own, and no agent can make its telemetry appear under another user.

### Scope
- Identity injection happens at an **Istio L7 waypoint** in front of the collector, derived from the gateway's verified mesh identity — never from agent-supplied data.
- Any agent-supplied identity is overridden.
- The injected identity resolves to the owning user.

### Out of scope
- The read API and per-user filtering itself — #2029 (this is the trust it relies on).
- Enabling export from agents — #2028.
- **Content integrity**: an agent can still misreport its *own* numbers (e.g. inflate its token count). This guarantees correct **attribution** only — an agent can pollute its own telemetry, never another user's.

### Dependencies
- Blocked on #2027 (the datastore) and #2028 (telemetry must be flowing to attribute it).
- #2029 depends on this for trustworthy ownership.
- Part of #1908.

PR #2738, opened 2026-07-08
feat(telemetry): identify agent telemetry by template and agent name

Tomas2D:
## Summary

Agent telemetry from claude-code-derived images is currently indistinguishable in the exploration UI: every harness exports under the Claude Code CLI's default OTel service name, so a `nous` agent's traces file under service `claude-code` alongside everyone else's, and nothing anywhere carries the name the user gave their agent. Found while verifying #2419 end-to-end ("I don't see nous in the logs").

Two changes:

- **Service name = template name.** The chart's agent telemetry env helper now emits `OTEL_SERVICE_NAME` set to the template key, for every telemetry-enabled template. A nous agent reads as service `nous`; `claude-code` is unchanged; future templates get this for free.
- **User-declared agent name as a resource attribute.** Creating an agent from a telemetry-enabled template seeds `OTEL_RESOURCE_ATTRIBUTES` with `platform.agent.name=<name>` (percent-encoded per the W3C Baggage rules the OTel env spec uses), merged into an existing value if the template declares one. A rename refreshes the attribute — but only where it already exists, so a rename never adds telemetry env to an agent that doesn't carry it or that the user stripped it from. Searching `platform.agent.name:"calm-harbor"` in HyperDX finds the instance directly.

## Trust boundary

Both values are exported by the harness itself and are **display-only**. Attribution continues to rest solely on the gateway-stamped `platform.agent.id` (#2046) — `observability.md` now states this explicitly.

## Implementation notes

- Pure domain helpers (`seedTelemetryIdentity` / `renamedTelemetryIdentity`) in the agents module, wired into the create and update flows; the rename path reuses the existing env bump + enqueue so a running agent picks the new name up next turn.
- Existing agents keep their creation-time env — they gain the identity on recreation or by pasting the env vars into the agent's env editor.

## Testing

- 7 new unit tests for the domain helpers; all 241 api-server unit tests, `mise run check`, `helm:check:lint`, and `helm:check:render` pass.
- Chart render with `clickstack.enabled=true` confirms per-template `OTEL_SERVICE_NAME` (`claude-code`, `nous`).
- The underlying export path was verified live on a `CLICKSTACK=1` dev cluster as part of #2419.

### DIFF
--- a/packages/api-server/src/modules/agents/domain/telemetry-env.ts
+++ b/packages/api-server/src/modules/agents/domain/telemetry-env.ts
@@ -1,0 +1,59 @@
+import type { EnvVar } from "api-server-api";
+
+// Resource attribute carrying the user-declared agent name so telemetry from
+// one instance is spottable in the exploration UI. Display-only: unlike
+// `platform.agent.id` (gateway-stamped, unforgeable), this value is exported
+// by the harness itself and must never be used for attribution or authz.
+const AGENT_NAME_ATTR = "platform.agent.name";
+const RESOURCE_ATTRS_ENV = "OTEL_RESOURCE_ATTRIBUTES";
+const TELEMETRY_MARKER_ENV = "CLAUDE_CODE_ENABLE_TELEMETRY";
+
+/** Set or replace `platform.agent.name` inside an OTEL_RESOURCE_ATTRIBUTES
+ *  value (comma-separated `key=value` pairs; values percent-encoded per the
+ *  W3C Baggage rules the OTel env spec uses). */
+function upsertNameAttr(value: string | undefined, agentName: string): string {
+  const pair = `${AGENT_NAME_ATTR}=${encodeURIComponent(agentName)}`;
+  const others = (value ?? "")
+    .split(",")
+    .map((p) => p.trim())
+    .filter((p) => p !== "" && !p.startsWith(`${AGENT_NAME_ATTR}=`));
+  return [...others, pair].join(",");
+}
+
+/**
+ * Create-time: when the template env enables telemetry, return a copy with
+ * the agent's user-declared name merged into OTEL_RESOURCE_ATTRIBUTES.
+ * Env without the telemetry marker passes through untouched.
+ */
+export function seedTelemetryIdentity(
+  env: EnvVar[],
+  agentName: string,
+): EnvVar[] {
+  if (!env.some((e) => e.name === TELEMETRY_MARKER_ENV)) return env;
+  const existing = env.find((e) => e.name === RESOURCE_ATTRS_ENV);
+  const next = {
+    name: RESOURCE_ATTRS_ENV,
+    value: upsertNameAttr(existing?.value, agentName),
+  };
+  return existing
+    ? env.map((e) => (e === existing ? next : e))
+    : [...env, next];
+}
+
+/**
+ * Rename-time: rewrite the name attribute only where it already exists, so a
+ * rename never adds telemetry env to an agent that doesn't carry it (or that
+ * the user stripped it from). Returns null when nothing needs writing.
+ */
+export function renamedTelemetryIdentity(
+  env: EnvVar[],
+  agentName: string,
+): EnvVar[] | null {
+  const existing = env.find(
+    (e) => e.name === RESOURCE_ATTRS_ENV && e.value.includes(AGENT_NAME_ATTR),
+  );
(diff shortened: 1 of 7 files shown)


packages/api-server/src/modules/agents/domain/telemetry-env.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Rename-time: rewrite the name attribute only where it already exists, so a
+ * rename never adds telemetry env to an agent that doesn't carry it (or that
+ * the user stripped it from). Returns null when nothing needs writing.
+ */
+export function renamedTelemetryIdentity(
+  env: EnvVar[],
+  agentName: string,
+): EnvVar[] | null {
+  const existing = env.find(
+    (e) => e.name === RESOURCE_ATTRS_ENV && e.value.includes(AGENT_NAME_ATTR),
```
dam-code-guardian, line 53: 🟢 **Suggestion:** The rename guard matches on `e.value.includes(AGENT_NAME_ATTR)`, a bare substring. An unrelated attribute whose key or value merely contains the literal `platform.agent.name` (e.g. `my.platform.agent.name.note=x`) would satisfy this even though the env carries no real `platform.agent.name=` pair — and a rename would then rewrite env the guard is meant to leave untouched, adding the telemetry attribute. Matching on the pair prefix (with the `=`) closes this.

```suggestion
    (e) => e.name === RESOURCE_ATTRS_ENV && e.value.split(",").some((p) => p.trim().startsWith(`${AGENT_NAME_ATTR}=`)),
```

Low likelihood given the controlled attribute set today, hence a suggestion rather than a warning.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `816aacc`

## PR #2738: feat(telemetry): identify agent telemetry by template and agent name
**Author:** Tomas2D | **Branch:** feat/agent-telemetry-identity → main | **Changes:** +173 −5 (7 files)

### Summary
Makes claude-code-derived agent telemetry distinguishable: sets `OTEL_SERVICE_NAME` to the template name in the Helm telemetry env helper, and seeds/refreshes a display-only `platform.agent.name` resource attribute in the api-server create and rename flows. Attribution still rests solely on the gateway-stamped `platform.agent.id`. Two new pure domain helpers with 7 unit tests; `observability.md` and the nous README updated.

### Findings
- 🟢 **Suggestion:** `renamedTelemetryIdentity` gates on `e.value.includes(AGENT_NAME_ATTR)` (`packages/api-server/src/modules/agents/domain/telemetry-env.ts:53`). A substring match can be tripped by an unrelated attribute whose key/value contains the literal `platform.agent.name` (e.g. `my.platform.agent.name.note=x`), causing a rename to rewrite env that carries no real `platform.agent.name=` pair — and thereby adding telemetry env the guard is meant to protect against. Matching on `platform.agent.name=` (with the `=`) tightens it. Low likelihood in practice given the controlled attribute set, hence a suggestion.
- ✅ **Looks good:** Helpers are pure and return-new (no mutation of the input `EnvVar[]`); the `=== existing` object-identity map-replace is correct and avoids duplicate `OTEL_RESOURCE_ATTRIBUTES` entries.
- ✅ **Looks good:** `renamedTelemetryIdentity` correctly returns `null` in all three no-write cases (no attribute present, unchanged value, empty env), so a rename never adds telemetry env to an agent that lacks it and never triggers a needless bump/enqueue.
- ✅ **Looks good:** Percent-encoding via `encodeURIComponent` correctly escapes `,` and `=`, so a name like `a,b=c` can't inject extra attributes — covered by a dedicated test.
- ✅ **Looks good:** The Helm helper's new `(dict "root" $ "templateName" $name)` signature is matched at its sole call site in `agent-templates.yaml:90`; no other caller left on the old signature.
- ✅ **Looks good:** Trust boundary is sound — both new values are explicitly display-only and documented as such in `observability.md`; no attribution/authz path consumes them.

✅ No documentation drift detected. The behavior change (self-declared telemetry identity) is reflected by a new bullet in `docs/architecture/observability.md`, whose `Last verified:` header is bumped to the PR date (2026-07-08). No ADR leak, no volatile content added, no new subsystem, no cross-reference rot.

### Verdict
APPROVE — Correct, immutable, well-tested identity helpers with a sound display-only trust boundary; the lone substring-match nit is a low-likelihood robustness suggestion, not a blocker.

---

pilartomas (approved):

PR #2336, opened 2026-07-03
feat: agent telemetry read API and CLI

JanPokorny:

PR #2740, opened 2026-07-08
feat(ui): agent metrics views in chat sidebar and experiments

JanPokorny:
Follow-up to #2336 (telemetry read API and CLI): adds the web-UI read surface, session scoping, and renames the feature to **metrics**.

## What's here

- **Chat view — METRICS tab** in the right sidebar (next to Files/Config): token spend per model, session totals, and recent LLM calls for the selected agent, polling every 15s. A **"This session / All sessions"** toggle scopes the view; session scope filters by exact session id.
- **Experiments — inline metrics per arm**: a cost/token summary chip in each arm header and a per-run Cost column, joined from `runtimeBySession` via `run.sessionId` (the ACP session id Claude Code reuses as its OTel `session.id`).
- **API**: `overview` gains an optional `sessionId` filter; `sinceHours` is now optional — both are independent, composable filters and the default window is **all time** (720h cap still applies when a window is given).
- **CLI**: `dam metrics <agent> [--session <id>] [--since <hours>]`.
- **Rename telemetry → metrics** across the feature (router key, types, module dirs, CLI command, UI copy). The OTel subsystem keeps its telemetry name (collector, export path, `CLAUDE_CODE_ENABLE_TELEMETRY`, server tracing) — only the agent-metrics read feature is renamed. ⚠️ Breaking for early adopters of #2336: `telemetry.overview` → `metrics.overview`, `dam telemetry` → `dam metrics` (no alias kept).
- **Build fix**: declare `drizzle-orm` as a peer of `drizzle-kit` via `pnpm.packageExtensions` — drizzle-kit probes `drizzle-orm/version` without declaring it, which breaks under pnpm's global virtual store (`enable-global-virtual-store`).

## Verification

Exercised end-to-end on a local CLICKSTACK=1 cluster with a real Claude Code agent: empty state, live data flowing agent → collector → ClickHouse → tab, both toggle scopes (session-only vs all-time numbers verified against each other), and the experiment arm chip + per-run cost against a completed single-arm experiment. Unit tests cover the ownership gate, session-filter pass-through, and formatters; `mise run check` and `mise run test` are green.


JanPokorny:
image

image

package.json
```diff
@@ -5,6 +5,11 @@
   "packageManager": "pnpm@10.33.4+sha512.1c67b3b359b2d408119ba1ed289f34b8fc3c6873412bec6fd264fbdc82489e510fcbecb9ce9d22dae7f3b76269d8441046014bdca53b9979cd7a561ad631b800",
   "pnpm": {
     "packageExtensions": {
+      "drizzle-kit": {
+        "peerDependencies": {
+          "drizzle-orm": "*"
+        }
+      },
```
JanPokorny, line 12: This just bandages a bug in Drizzle that prevents it from using a global virtual store layout (soon to be default in pnpm)

PR #2744, opened 2026-07-09
ci: gate dependent image builds on actual base-merge result

jezekra1:
## Problem

Run [29004687116](https://github.com/dam-agents/dam/actions/runs/29004687116) on main failed with `build-mock` and `build-agents` dying on `FROM ${BASE_IMAGE}` — `quay.io/dam-agents/platform-base:60d4952… not found`.

Root cause: `build-platform-base` was auto-cancelled by GitHub after ~15 min of *"The job was not acquired by Runner of type hosted even after multiple attempts"* (hosted-runner capacity flake), so `merge-platform-base` was skipped and the tag never pushed. The downstream `!cancelled() && !failure()` guards — which exist so agents still build when platform-base is *intentionally* skipped for registry reuse — also pass when a dependency is **cancelled**, because a cancelled job is neither `failure()` nor `cancelled()` (the run itself wasn't cancelled). So `build-mock`/`build-agents` ran against a tag that never existed.

## Fix

Distinguish "skipped because reusing a registry-verified tag" from "didn't succeed":

- `build-agents` / `build-mock`: also require `platform_base != 'true'` (reuse — the resolver verified the tag exists) **or** `merge-platform-base.result == 'success'`.
- `build-nous` / `build-openevolve`: same pattern against `merge-agents`.
- `merge-agents` / `merge-nous` / `merge-openevolve`: require the corresponding build job's `result == 'success'`, so a partially-cancelled matrix can never push an incomplete multi-arch manifest (latent variant of the same bug).
- `build-agents` / `build-mock` additionally require `changes.result == 'success'` so a cancelled resolver job can't slip an empty tag through.

With this, a runner-acquisition flake skips the dependent jobs cleanly (run still red, `gh run rerun --failed` recovers) instead of failing them with a misleading `not found`.

The failed main run has been re-run to republish the missing `60d4952` images.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -295,7 +295,14 @@ jobs:
 
   build-agents:
     needs: [merge-platform-base, changes]
-    if: ${{ needs.changes.outputs.agents != '[]' && !cancelled() && !failure() }}
+    # A cancelled dependency (e.g. runner never acquired) is neither failure()
+    # nor cancelled(), so !failure() alone would build agents against a
+    # platform-base tag that was never pushed. Require the tag to be either
+    # reused (registry-verified by the resolver) or actually merged.
+    if: >-
+      ${{ needs.changes.result == 'success' &&
+          needs.changes.outputs.agents != '[]' && !cancelled() && !failure() &&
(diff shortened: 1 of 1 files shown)


jezekra1:
overthinking

PR #2735, opened 2026-07-08
feat: rework ADR governance as an event-sourced log

tomkis:
Reworks how ADRs are used today. ADRs become an immutable event log; architecture docs and an ADR index are projections of it. Re-grants agent access to ADRs under strict, mostly-enforced constraints. This PR tracks implementing the full model (design doc included).

## Model

- ADR log: append-only. Only mutation on an accepted record is stamping supersession.
- Architecture docs: projection of "what is". Lossy on purpose (no dead branches).
- ADR index: generated projection of "what was decided, still live or not". Read first when authoring.
- Agent ADR reads gated to two jobs only: authoring a new ADR, recompiling docs. Ordinary work uses docs only.

## Implementation

1. Design doc [x]
- [x] `docs/design/adr-governance.md`

2. ADR structure + index [x]
- [x] ADR frontmatter (id, title, status, supersedes, subsystem/tags, summary)
- [x] Generated `adrs/index.md` from frontmatter, resolving supersession
- [x] `check:generated`-style gate for the index
- [x] Update `/adr` skill to enforce frontmatter

3. Immutability enforcement [x]
- [x] Deterministic `check:adr-immutable` script (freezes accepted prose body; forbid delete/rename)
- [x] Wire into precommit + CI (staged-vs-HEAD locally, merge-base in CI)

4. Doc size cap [x]
- [x] `check:doc-size` gate (per-page + index caps) in precommit + CI
- [x] Write-time PreToolUse hook on `docs/architecture/**` with steering rejection message
- [x] Shared measurement between hook and gate

5. Read access [x]
- [x] Re-grant ADR reads gated to authoring flow (recompile flow: follow-up)
- [x] Prompt agents to prefer docs over ADRs for ordinary work

6. Review-time policy [x]
- [x] New ADR-policy skill (re-litigation, supersedes correctness, summary honesty; calls immutability script)
- [x] doc-drift keeps docs-vs-code + ADR-changed-state-not-recompiled
- [x] Both run in one code-review agent pass

7. Guidelines [x]
- [x] Update documentation-guidelines.md (immutability, human-first reads, cap, generated index, traceability via tags)

## Open questions

1. Read gating: per-flow allow, or trust the agent to separate authoring/recompiling from ordinary work?
2. Where the design doc lives long term: promote to an ADR, fold into documentation guidelines, or both?

PR #2754, opened 2026-07-10
docs(adr): reframe doc-cap steering as a re-think signal

tomkis:
Merges into #2735.

The doc-size cap is meant to make the agent re-think whether a page carries too much detail or needs splitting. Two problems in #2735 defeated that.

## Problem 1 — the steering invited rewording

`overageReport` led with "tighten prose and merge related statements." Under a hard gate the agent does the cheapest thing that clears it and stops — and the cheapest thing was rewording, listed first. The re-think options came after.

## Problem 2 — "relocate to an ADR" was conceptually broken

The original remedy told the agent to "move detail and rationale down into an ADR." That treats the log as a spillover tank. It isn't:

- ADRs are filed *before* work, capture a *decision*, and are *accepted* then *immutable*. Descriptive reality has no acceptance semantics and rots once frozen. Minting an ADR after the fact to relieve doc size is a category error.
- Over-cap is almost always an **altitude/scope** problem: the page drifted below architecture level, or covers two subsystems. The remedy is raise altitude, cut volatile detail, or split — no ADR involved.
- Relocating to the log is valid only for genuine **decision rationale** (why X over Y), and that already has an ADR from when the decision was made. So the move is "cut it from the page," not "author a new ADR."

The "docs are a projection of the ADR log / the cap and the log are one mechanism" framing seeded the error. Architecture docs are a hand-maintained reduced view, not a generated projection (only `adrs/index.md` is generated). Dropped that vocabulary.

## Changes

- **`scripts/doc-size.mjs`** — steering is now diagnose-then-reconcile: raise altitude first, split second, relocate decision-rationale last (and never mint a new ADR to relieve size). Top comment drops the projection/recompile/one-mechanism framing.
- **`documentation-guidelines.md`** — grade the "why" explicitly (pages keep *operational* why and stand alone; the log keeps *decision history*), and replace the stale "No length cap" line with the corrected remedy ladder.

Shared message, so hook rejection and CI gate read identically. Verified rendering and the gate still runs.

## Note (out of scope)

`docs/architecture/connections.md` is already ~1KB over the cap on the #2735 branch, so that PR is currently tripping its own gate. Worth resolving there.

### DIFF
--- a/docs/guidelines/documentation-guidelines.md
+++ b/docs/guidelines/documentation-guidelines.md
@@ -24,7 +24,7 @@ Architecture pages are the **authoritative, self-contained description of the cu
 
 - One page per subsystem under [`docs/architecture/`](../architecture/), indexed from [`docs/architecture.md`](../architecture.md).
 - Adding a new subsystem means adding a new page and linking it from the landing page.
-- No shared template. Free-form per page, under a per-page character cap (enforced by `mise run check` and a write-time hook). The cap is a forcing function, not a style rule: going over means the page drifted below architecture altitude, or grew to cover more than one subsystem. Reconcile by raising altitude and cutting volatile detail, or by splitting the subsystem and reconsidering its boundaries — not by rewording to fit. Do not offload description into an ADR to relieve size: the log holds decisions authored when they were made, not documentation spillover.
+- No shared template. Free-form per page, under a per-page character cap (enforced by `mise run check` and a write-time hook). The cap is a forcing function, not a style rule: going over means the page dropped to the level of the code, or grew to cover more than one subsystem. Reconcile by raising the level and cutting volatile detail, or by splitting the subsystem and reconsidering its boundaries — not by rewording to fit. Do not offload description into an ADR to relieve size: the log holds decisions authored when they were made, not documentation spillover.
 - Cross-page concept ownership: one page owns each concept in depth; others one-liner + cross-link.
 
 ### Mandatory headers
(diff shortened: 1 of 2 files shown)


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `d229b7f`

## PR #2754: docs(adr): reframe doc-cap steering as a re-think signal
**Author:** tomkis | **Branch:** docs/adr-governance-cap-steering → docs/adr-governance | **Changes:** +9 −7 (2 files)

### Summary
Reorders the `overageReport` steering message in `scripts/doc-size.mjs` so the re-think remedies (relocate detail to an ADR, split the subsystem) precede prose-tightening, which is demoted to "last, never the first move" — closing a gap in #2735 where an agent under the hard cap would take the cheapest path (reword) because it was listed first. Also replaces the stale "No length cap" line in `documentation-guidelines.md` with the cap and its forcing-function rationale, so the intent is visible in the always-loaded guidelines rather than only at the rejection gate.

### Findings
- ✅ **Looks good:** The remedy reordering (relocate → split → tighten) matches the stated intent, and the guideline text now mirrors the rejection message's framing — the two surfaces read consistently.
- ✅ **Looks good:** The shared-message design is preserved; the hook rejection and CI gate still render identical text, so the reword doesn't fork the two surfaces.
- 🟢 **Suggestion:** The PR description flags that `docs/architecture/connections.md` is already ~1080 chars over the cap on the #2735 base branch (its own gate is tripping). That's correctly scoped out here, but it's a real follow-up on #2735 — worth a tracking note so it isn't lost when this merges.

### Verdict
APPROVE — A small, well-motivated wording/ordering change that strengthens the forcing function without altering control flow; no correctness, security, or performance concerns.

---

PR #2762, opened 2026-07-13
fix(platform-base): create ca-trust dirs before chown

kapetr:
Agent image builds currently fail at the runner stage:

```
chown: cannot access '/etc/pki/ca-trust/source/anchors': No such file or directory
```

The runner base (`core-runtime:latest`, a floating tag) recently shipped without the empty ca-trust anchors directory — empty dirs are exactly what image minimization tends to drop. `mkdir -p` the dirs before chowning them, making the build independent of which directories the base happens to pre-create. The dirs are needed regardless: the entrypoint installs the gateway CA there.

No change for bases that still ship the dirs.

### DIFF
--- a/packages/platform-base/Dockerfile
+++ b/packages/platform-base/Dockerfile
@@ -115,7 +115,7 @@ COPY packages/platform-base/working-dir/.claude/settings.json /app/working-dir/.
 COPY packages/platform-base/runtime-manifest.yaml /app/runtime-manifest.yaml
 COPY packages/platform-base/AGENTS.md /etc/AGENTS.md
 
-RUN mkdir -p /home/agent &&\
+RUN mkdir -p /home/agent /etc/pki/ca-trust/source/anchors /etc/pki/ca-trust/extracted /etc/pki/tls/certs &&\
     chown -R 65532 /etc/mise /etc/pki/ca-trust/source/anchors /etc/pki/ca-trust/extracted /etc/pki/tls/certs /app /home/agent /usr/local/share/mise &&\
     ldconfig &&\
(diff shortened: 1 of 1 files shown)


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `e28fcf7`

## PR #2762: fix(platform-base): create ca-trust dirs before chown
**Author:** kapetr | **Branch:** fix/platform-base-catrust-dirs → main | **Changes:** +1 −1 (1 file)

### Summary
The agent image build was failing at the runner stage because the floating `core-runtime:latest` base stopped shipping the empty ca-trust anchors directory. This adds the three ca-trust/tls dirs to the existing `mkdir -p` so they always exist before the subsequent `chown -R`, making the build independent of what the base pre-creates.

### Findings
- ✅ **Looks good:** `mkdir -p` is idempotent and creates parents, so this is a no-op on bases that already ship the dirs — no regression, exactly as the PR body claims (`packages/platform-base/Dockerfile:118`).
- ✅ **Looks good:** All three added paths (`/etc/pki/ca-trust/source/anchors`, `/etc/pki/ca-trust/extracted`, `/etc/pki/tls/certs`) match the existing `chown -R` target list on the next line, so every chowned path is now guaranteed to exist.
- 🟢 **Suggestion:** Nothing actionable — if a future entrypoint writes elsewhere under `/etc/pki/ca-trust/source`, it would need the same treatment, but that's out of scope here.

### Verdict
APPROVE — Correct, minimal, idempotent build fix that resolves the runner-stage `chown` failure without affecting bases that still ship the dirs.

---

jezekra1 (approved):

PR #2763, opened 2026-07-13
fix(cli): route WebSocket connections through the proxy env

pilartomas:
## Problem

Every `dam` command that opens a **WebSocket** fails from inside an agent pod (and any environment whose only egress is a forward proxy):

- `dam ssh connect <agent>` → ssh reports `kex_exchange_identification: Connection closed by remote host`
- `dam chat` terminal bridge and the ACP session client fail the same way

Meanwhile `dam agent list`, `dam file list`, etc. work fine in the same environment.

### Root cause

The `ws` client opens its own raw `http`/`https` socket. Unlike Node's global `fetch`/undici — which honors `NODE_USE_ENV_PROXY=1` — **`ws` ignores the standard proxy env vars unless you hand it an `agent`**. All three CLI WebSocket call sites constructed `new WebSocket(url)` with no options, so the connection went **direct**, bypassing the mandatory forward proxy. In a locked-down pod, direct egress is blocked (and the proxy is also what injects the real bearer token — in-pod `DAM_TOKEN` is a placeholder), so the socket is refused and the stream closes mid-handshake.

tRPC/REST calls work because they use `fetch`, which respects the proxy.

### Evidence

Reproduced from a `claude-code` agent pod (egress via `HTTPS_PROXY`, `NODE_USE_ENV_PROXY=1`):

| Check | Result |
|---|---|
| `dam ping` / `dam agent get` (tRPC) | ✅ works |
| `dam file list <agent> /home/agent` (REST) | ✅ works |
| direct (`--noproxy`) curl to DAM host | ❌ blocked |
| curl via `HTTPS_PROXY` to DAM host | ✅ `HTTP 200` |
| `ssh <managed-host>` (WebSocket bridge) | ❌ `Connection closed by remote host` at kex |

## Fix

Add a shared `proxyAgentForUrl()` helper (`packages/cli/src/modules/shared/ws-proxy.ts`) that builds an `http(s)-proxy-agent` from the environment:

- `wss:`/`https:` targets → `HttpsProxyAgent` (CONNECT tunnel); `ws:`/`http:` → `HttpProxyAgent`
- picks `HTTPS_PROXY`/`HTTP_PROXY` (and lowercase variants) by target scheme
- respects `NO_PROXY` (suffix match, leading dot, `*`)
- returns `undefined` when no proxy applies → **direct-connection behavior unchanged** (`ws` treats `{ agent: undefined }` as no agent)

Pass it as the `ws` `agent` at all three call sites:
- `packages/cli/src/modules/ssh/infrastructure/raw-bridge.ts`
- `packages/cli/src/modules/chat/infrastructure/terminal-bridge.ts`
- `packages/cli/src/modules/chat/infrastructure/acp-session-client.ts`

Adds `http-proxy-agent` / `https-proxy-agent` (both already present transitively in the lockfile) to the CLI's deps, and a unit test covering scheme selection, lowercase vars, and `NO_PROXY`.

## Testing

- `mise run check` passes locally (lint, type-check, format, helm render — the full repo gate).
- `mise run cli:test` passes: 150 tests across 25 files, including the 9 new `ws-proxy.test.ts` cases.
- The follow-up commit adds the missing `pnpm-lock.yaml` entries for the two new deps (the original push predated a working `pnpm install`, so CI failed at frozen-lockfile setup) and a prettier line-wrap fix in `ws-proxy.ts`.

### DIFF
--- a/packages/cli/src/modules/shared/ws-proxy.ts
+++ b/packages/cli/src/modules/shared/ws-proxy.ts
@@ -1,0 +1,59 @@
+import type { Agent } from "node:http";
+
+import { HttpProxyAgent } from "http-proxy-agent";
+import { HttpsProxyAgent } from "https-proxy-agent";
+
+/**
+ * The `ws` client opens its own raw http/https socket and, unlike Node's global
+ * fetch/undici (which honors `NODE_USE_ENV_PROXY`), it ignores the standard
+ * proxy env vars. In locked-down environments — e.g. an agent pod whose only
+ * egress path is a forward proxy that also injects auth — a WebSocket opened
+ * without an explicit agent bypasses the proxy, so the connection is refused
+ * and the caller sees the stream close mid-handshake. The tRPC/REST calls work
+ * in the same environment precisely because they go through fetch.
+ *
+ * Build a proxy agent from the environment so `ws` tunnels through it too.
+ * Returns undefined when no proxy applies (no relevant `*_PROXY` var, or the
+ * target host matches `NO_PROXY`), which leaves direct-connection behavior
+ * unchanged — `ws` treats `{ agent: undefined }` as "no agent".
+ */
+export function proxyAgentForUrl(
+  url: string,
+  env: NodeJS.ProcessEnv = process.env,
+): Agent | undefined {
+  let target: URL;
+  try {
+    target = new URL(url);
+  } catch {
+    return undefined;
+  }
+
+  const secure = target.protocol === "wss:" || target.protocol === "https:";
+  const proxy = secure
+    ? (env.HTTPS_PROXY ?? env.https_proxy)
+    : (env.HTTP_PROXY ?? env.http_proxy);
+  if (!proxy) return undefined;
+  if (isNoProxy(target.hostname, env.NO_PROXY ?? env.no_proxy)) return undefined;
+
+  // A `wss:`/`https:` target is tunneled through the proxy with CONNECT
+  // (HttpsProxyAgent); a plain `ws:`/`http:` target is forwarded with an
+  // absolute-form request (HttpProxyAgent). The proxy URL's own scheme is
+  // orthogonal — both agents accept an `http://` proxy.
+  return secure ? new HttpsProxyAgent(proxy) : new HttpProxyAgent(proxy);
+}
+
+/**
+ * Match a hostname against a `NO_PROXY` list: comma-separated hosts or domain
+ * suffixes, an optional leading dot, and `*` as a wildcard for all hosts.
+ */
+function isNoProxy(host: string, noProxy: string | undefined): boolean {
+  if (!noProxy) return false;
+  for (const raw of noProxy.split(",")) {
+    const entry = raw.trim();
+    if (entry === "") continue;
+    if (entry === "*") return true;
+    const bare = entry.startsWith(".") ? entry.slice(1) : entry;
+    if (host === bare || host.endsWith(`.${bare}`)) return true;
+  }
+  return false;
+}
--- a/packages/cli/src/modules/shared/ws-proxy.ts
+++ b/packages/cli/src/modules/shared/ws-proxy.ts
@@ -33,7 +33,8 @@ export function proxyAgentForUrl(
     ? (env.HTTPS_PROXY ?? env.https_proxy)
     : (env.HTTP_PROXY ?? env.http_proxy);
   if (!proxy) return undefined;
-  if (isNoProxy(target.hostname, env.NO_PROXY ?? env.no_proxy)) return undefined;
+  if (isNoProxy(target.hostname, env.NO_PROXY ?? env.no_proxy))
+    return undefined;
 
   // A `wss:`/`https:` target is tunneled through the proxy with CONNECT
   // (HttpsProxyAgent); a plain `ws:`/`http:` target is forwarded with an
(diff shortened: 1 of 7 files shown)


jezekra1 (approved):

packages/cli/src/modules/shared/ws-proxy.ts
```diff
@@ -0,0 +1,60 @@
...
+ * suffixes, an optional leading dot, and `*` as a wildcard for all hosts.
+ */
+function isNoProxy(host: string, noProxy: string | undefined): boolean {
+  if (!noProxy) return false;
+  for (const raw of noProxy.split(",")) {
+    const entry = raw.trim();
+    if (entry === "") continue;
+    if (entry === "*") return true;
+    const bare = entry.startsWith(".") ? entry.slice(1) : entry;
+    if (host === bare || host.endsWith(`.${bare}`)) return true;
```
dam-code-guardian, line 59: 🟢 **Suggestion:** `NO_PROXY` matching is case-sensitive here. `target.hostname` from `URL` is already lowercase, but a `NO_PROXY` entry containing uppercase (e.g. `.Internal.Example.com`) will silently fail to match — curl/undici compare `NO_PROXY` case-insensitively. Lowercasing `bare` closes the gap:

```suggestion
    const bare = (entry.startsWith(".") ? entry.slice(1) : entry).toLowerCase();
    if (host === bare || host.endsWith(`.${bare}`)) return true;
```

Minor — env vars are lowercase by convention in practice, so this is a defensive nicety rather than a bug.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `ac6e393`

## PR #2763: fix(cli): route WebSocket connections through the proxy env
**Author:** pilartomas | **Branch:** fix/cli-ws-proxy → main | **Changes:** +152 −7 (7 files)

### Summary
The `ws` client opens its own raw socket and ignores the standard proxy env vars unless handed an `agent`, so `dam ssh connect` / `dam chat` / the ACP session client connected directly and failed in proxy-only egress environments (agent pods). This adds a shared `proxyAgentForUrl()` helper that builds an `http(s)-proxy-agent` from `HTTPS_PROXY`/`HTTP_PROXY` (scheme-selected, `NO_PROXY`-aware) and passes it as the `ws` `agent` at all three call sites; returning `undefined` preserves direct-connection behavior when no proxy applies.

### Findings
- 🟢 **Suggestion:** `NO_PROXY` entries are compared case-sensitively (`ws-proxy.ts:57`). `URL` lowercases the target hostname, but an uppercase `NO_PROXY` entry (e.g. `.Internal.Example.com`) won't match. curl/undici treat `NO_PROXY` case-insensitively — consider lowercasing `bare` before comparing. Minor: env vars are conventionally lowercase in practice.
- ✅ **Looks good:** Scheme-based proxy selection is correct — a `wss:`/`https:` target uses `HttpsProxyAgent` (CONNECT tunnel) and never falls back to `HTTP_PROXY`, verified by the `does not use HTTP_PROXY for a secure target` test.
- ✅ **Looks good:** Defensive `new URL()` in a try/catch returns `undefined` on an unparseable URL, so a malformed URL degrades to direct connection rather than throwing at the call site.
- ✅ **Looks good:** `undefined` return leaves `{ agent: undefined }`, which `ws` treats as no agent — direct-connection behavior is genuinely unchanged when no proxy is configured. All three call sites are updated consistently.
- ✅ **Looks good:** Well-targeted unit test (9 cases) covering scheme selection, lowercase env vars, `NO_PROXY` suffix/wildcard/exact/non-match, and the unparseable-URL path. Both new deps were already present transitively; the follow-up commit adds the frozen-lockfile entries.

### Verdict
APPROVE — Correct, well-scoped transport fix with a clean fallback path and solid test coverage; the single suggestion is a minor edge case.

---

packages/cli/src/modules/shared/ws-proxy.ts
pilartomas: Applied in f7a40aa5 — `bare` is lowercased and a test pins the case-insensitive match (`NO_PROXY=.Internal.Example.com` vs `api.internal.example.com`).

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `f7a40aa`

## PR #2763: fix(cli): route WebSocket connections through the proxy env
**Author:** pilartomas | **Branch:** fix/cli-ws-proxy → main | **Changes:** +101 −7 (7 files)

### Summary
Routes the three CLI `ws` call sites (`dam ssh connect`, `dam chat` terminal bridge, ACP session client) through a shared `proxyAgentForUrl()` helper so WebSockets honor `HTTPS_PROXY`/`HTTP_PROXY`/`NO_PROXY` in proxy-only egress environments; `undefined` return preserves direct-connection behavior. This re-review covers the single follow-up commit `f7a40aa`, which makes `NO_PROXY` matching case-insensitive.

### Changes since last review
Previous HEAD: `ac6e393` (2026-07-13T11:22:24Z) — verdict APPROVE

- ✅ **Fixed:** `NO_PROXY` entries were compared case-sensitively (`ws-proxy.ts:57`) — commit `f7a40aa` now lowercases `bare` before comparing and adds a `matches NO_PROXY entries case-insensitively` test pinning `NO_PROXY=.Internal.Example.com` vs `api.internal.example.com`. Confirmed applied by the author in the inline thread.

### Findings
- ✅ **Looks good:** The case-insensitive `NO_PROXY` fix is correct and matches curl/undici semantics; `URL` already lowercases the target hostname, so normalizing only the entry side is sufficient.

### Verdict
APPROVE — The one suggestion from the prior review is now resolved with a matching test; no new issues in the follow-up commit.

---

PR #2765, opened 2026-07-13
fix(egress-rules): make duplicate create idempotent, verdict clash a CONFLICT

### LINKED ISSUE #2324: Creating a network rule that already exists fails with an internal error
## Problem

Adding a network (egress) rule — via `dam network create` or the network panel — fails with an internal error when an equivalent rule already exists for the agent:

```
error: row.decidedAt.toISOString is not a function
```

The command reports failure, but the rule state on the server is actually fine: the existing rule remains in place and any related side effects still happen. So the user sees a scary error for what is effectively a no-op / duplicate, and can't tell whether their change took effect.

Observed on a live cluster (2026-07-02) while adding a rule for a host that already had a matching rule.

## Proposed solution

Adding a rule that duplicates an existing one should succeed quietly (or return a clear "rule already exists" result), not surface an internal error. The command's success/failure should reflect the actual outcome.

### Cause (for whoever picks it up)

The value returned on the "rule already existed" read-back path carries its decided-at timestamp as a string, while the layer that formats the response assumes it's a date object and calls a date-only method on it. The happy-path insert returns a proper date, which is why a first-time create works and only the duplicate path throws.

### Scope

- Affects the create path (and likely the update path, which shares the same formatting step). Both should be checked.

PetrBulanek:
Adding a network rule that duplicates an existing one — via `dam network create` or the network panel — used to fail with an internal error (`row.decidedAt.toISOString is not a function`) even though the rule state on the server was fine.

Create is now intent-based:

- **Exact duplicate** (same host/method/path and port, same verdict): succeeds quietly and returns the existing rule. Idempotent — safe for re-run scripts and double-clicks. If the existing rule came from a preset or connection, the manual create takes ownership of it (source becomes `manual`), so a later preset switch won't silently remove a rule the user explicitly asked for.
- **Verdict clash** (same rule, opposite verdict): fails with a clear conflict error telling the user to edit the existing rule instead — create never flips a verdict.
- **Port clash** (same rule, different port): fails with a clear conflict error pointing at revoke-and-recreate, since a rule's port can't be edited — instead of returning a "created" rule that silently lacks the requested port.

The command's success or failure now reflects the actual outcome. No CLI, UI, or API contract changes.

Closes #2324

### DIFF
--- a/packages/api-server/src/modules/egress-rules/infrastructure/egress-rules-repository.ts
+++ b/packages/api-server/src/modules/egress-rules/infrastructure/egress-rules-repository.ts
@@ -41,6 +41,16 @@ export interface EgressRulesRepository {
    *  preset is not stored on the spec — its rules' sources are the truth. */
   getPresetForAgent(agentId: string): Promise<EgressPreset>;
   getById(id: string): Promise<EgressRuleRow | null>;
+  /** Exact lookup on the active-row unique index (`egress_rules_lookup_idx`).
+   *  Used as the insert conflict fallback so callers get *the* conflicting
+   *  row, not a best-match — `findMatch` treats the path as a request path,
+   *  not a pattern. */
+  getActiveByTuple(
+    agentId: string,
+    host: string,
+    method: string,
+    pathPattern: string,
+  ): Promise<EgressRuleRow | null>;
   insert(row: NewEgressRule): Promise<EgressRuleRow>;
   /** Insert-or-promote variant used by the connection-rules sync. If an
    *  existing active row for `(agent, host, method, pathPattern)` is a
@@ -106,7 +116,9 @@ type RawRule = {
   pathPattern: string;
   verdict: string;
   decidedBy: string;
-  decidedAt: Date;
+  // Raw `db.execute(sql...)` reads bypass drizzle's timestamptz mapper and
+  // yield strings; typed `.select()`/`.returning()` reads yield Dates.
+  decidedAt: Date | string;
   status: string;
   source: string;
 } & Record<string, unknown>;
@@ -121,7 +133,8 @@ function toRow(r: RawRule): EgressRuleRow {
     pathPattern: r.pathPattern,
     verdict: r.verdict as RuleVerdict,
     decidedBy: r.decidedBy,
-    decidedAt: r.decidedAt,
+    decidedAt:
+      r.decidedAt instanceof Date ? r.decidedAt : new Date(r.decidedAt),
     status: r.status as "active" | "revoked",
     source: r.source as EgressRuleSource,
   };
@@ -134,6 +147,24 @@ export function createEgressRulesRepository(db: Db): EgressRulesRepository {
         .select()
         .from(egressRules)
         .where(eq(egressRules.id, id));
+      return rows.length ? toRow(rows[0] as RawRule) : null;
+    },
+
+    async getActiveByTuple(agentId, host, method, pathPattern) {
+      // No LIMIT needed — the partial unique index guarantees at most one
+      // active row per tuple.
+      const rows = await db
+        .select()
+        .from(egressRules)
+        .where(
+          and(
+            eq(egressRules.agentId, agentId),
+            eq(egressRules.host, host),
+            eq(egressRules.method, method),
+            eq(egressRules.pathPattern, pathPattern),
+            eq(egressRules.status, "active"),
+          ),
+        );
       return rows.length ? toRow(rows[0] as RawRule) : null;
     },
 
@@ -228,7 +259,7 @@ export function createEgressRulesRepository(db: Db): EgressRulesRepository {
         .onConflictDoNothing()
         .returning();
       if (inserted.length) return toRow(inserted[0] as RawRule);
-      const existing = await this.findMatch(
+      const existing = await this.getActiveByTuple(
         row.agentId,
         row.host,
         row.method,
@@ -276,7 +307,7 @@ export function createEgressRulesRepository(db: Db): EgressRulesRepository {
       `);
       const promotedRows = promoted as unknown as RawRule[];
       if (promotedRows.length) return toRow(promotedRows[0]!);
-      const existing = await this.findMatch(
+      const existing = await this.getActiveByTuple(
         row.agentId,
         row.host,
         row.method,
--- a/packages/api-server/src/modules/egress-rules/services/egress-rules-service.ts
+++ b/packages/api-server/src/modules/egress-rules/services/egress-rules-service.ts
@@ -77,7 +77,7 @@ export function createEgressRulesService(
         });
         throw new TRPCError({ code: "NOT_FOUND", message: "agent not found" });
       }
-      const row = await deps.repo.insert({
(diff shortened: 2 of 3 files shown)


packages/api-server/src/modules/egress-rules/services/egress-rules-service.ts
```diff
@@ -97,6 +97,29 @@ export function createEgressRulesService(
         decidedBy: deps.ownerSub,
         source: "manual",
       });
+      // Insert conflicted with an existing rule carrying a different verdict.
+      // Nothing changed, so bail before the create audit log and L7 side
+      // effect — changing a verdict is the update/edit path's job.
+      if (row.verdict !== input.verdict) {
+        throw new TRPCError({
+          code: "CONFLICT",
+          message: `an equivalent rule already exists with verdict '${row.verdict}' — edit the existing rule instead`,
```
dam-code-guardian, line 97: 🟡 **Warning:** These two new branches — verdict-clash → `CONFLICT` (here) and same-verdict preset/connection duplicate → `updatePromoteToManual` (just below) — have no dedicated test. The only test change in this PR is the `getActiveByTuple` stub in the fake repo; `egress-rules-port.test.ts` covers only port promotion. Both encode subtle intent (verdict-flip rejection, ownership transfer) and would benefit from regression coverage.

packages/api-server/src/modules/egress-rules/infrastructure/egress-rules-repository.ts
```diff
@@ -137,6 +150,24 @@ export function createEgressRulesRepository(db: Db): EgressRulesRepository {
...
+      const rows = await db
+        .select()
+        .from(egressRules)
+        .where(
+          and(
+            eq(egressRules.agentId, agentId),
+            eq(egressRules.host, host),
+            eq(egressRules.method, method),
+            eq(egressRules.pathPattern, pathPattern),
+            eq(egressRules.status, "active"),
```
dam-code-guardian, line 165: 🟢 **Suggestion:** The correctness of `rows[0]` here rests entirely on the partial unique index staying `(agent_id, host, method, path_pattern) WHERE status='active'`. A future migration that widens or drops that index would make this silently return an arbitrary row. A cheap guard (`if (rows.length > 1) …` in dev, or an assertion) would fail loudly instead.

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `87a35be`

## PR #2765: fix: creating a network rule that already exists no longer fails with an internal error
**Author:** PetrBulanek | **Branch:** fix/2324-duplicate-egress-rule-create → main | **Changes:** +? (3 files)

### Summary
Makes `egress_rule.create` intent-based: an exact duplicate (same tuple + verdict) now succeeds idempotently and takes ownership of a preset/connection rule, while a same-tuple/opposite-verdict request fails with a clear `CONFLICT` instead of the old internal `row.decidedAt.toISOString is not a function` crash. Root cause was that the conflict-fallback used `findMatch` (a raw `db.execute` read that yields `decidedAt` as a string and treats the path as a request path, not a pattern); the new typed `getActiveByTuple` fixes both.

### Findings
- 🟡 **Warning:** The two new create-path branches — verdict-clash → `CONFLICT`, and same-verdict duplicate of a preset/connection row → `updatePromoteToManual` — have no dedicated unit tests. The only test change is the `getActiveByTuple: async () => null` stub added to the fake repo; the existing `egress-rules-port.test.ts` only exercises port promotion. Both branches encode subtle intent (ownership transfer, verdict-flip rejection) that would benefit from regression coverage. (`packages/api-server/src/modules/egress-rules/services/egress-rules-service.ts:100`)
- 🟢 **Suggestion:** The `getActiveByTuple` doc-comment says the partial unique index guarantees at most one active row per tuple, so `rows[0]` is safe — consider asserting `rows.length <= 1` in dev/test (or a code comment cross-linking the `WHERE status = 'active'` clause) so a future index change that widens the tuple fails loudly rather than silently returning an arbitrary row. (`packages/api-server/src/modules/egress-rules/infrastructure/egress-rules-repository.ts:153`)
- ✅ **Looks good:** The `decidedAt: Date | string` widening + `instanceof Date ? … : new Date(…)` normalization in `toRow` correctly reconciles the two read paths (typed `.select()`/`.returning()` yield Dates; raw `db.execute` yields strings) and is the real fix for the reported crash.
- ✅ **Looks good:** `getActiveByTuple` matches `(agentId, host, method, pathPattern)` + `status='active'`, exactly mirroring `egress_rules_lookup_idx` and the `onConflictDoNothing` conflict target — so the fallback returns *the* conflicting row rather than a best-match. Correct choice over reusing `findMatch`.

### Verdict
COMMENT — Solid, well-reasoned bug fix that correctly addresses the root cause; the only gap is missing regression tests for the two new create-path branches.

---

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `eba9fc2`

## PR #2765: fix(egress-rules): make duplicate create idempotent, verdict clash a CONFLICT
**Author:** PetrBulanek | **Branch:** fix/2324-duplicate-egress-rule-create → main | **Changes:** +71 −7 (4 files)

### Summary
Makes `egress_rule.create` intent-based: an exact duplicate (same tuple + verdict) succeeds idempotently and takes ownership of a preset/connection rule, while a same-tuple/opposite-verdict (or differing-port) request fails with a clear `CONFLICT` instead of the old internal `row.decidedAt.toISOString is not a function` crash. Root cause was the conflict-fallback using `findMatch` (raw `db.execute` → `decidedAt` string, path treated as request path not pattern); the new typed `getActiveByTuple` + `decidedAt: Date | string` normalization fix both.

### Changes since last review
Previous HEAD: `87a35be` (2026-07-13T12:41:39Z) — verdict COMMENT

- 🔁 **Still present:** Missing regression tests for the two new create-path branches — verdict-clash → `CONFLICT` and same-verdict duplicate of a preset/connection row → `updatePromoteToManual` (ownership transfer). The only test change remains the `getActiveByTuple: async () => null` stub. (`packages/api-server/src/modules/egress-rules/services/egress-rules-service.ts:106`)
- 🔁 **Still present:** `getActiveByTuple`'s `rows[0]` correctness rests on the partial unique index staying `(agent_id, host, method, path_pattern) WHERE status='active'`; a dev/test assertion or cross-linking comment would make a future index-widening migration fail loudly. (`packages/api-server/src/modules/egress-rules/infrastructure/egress-rules-repository.ts:154`)

_Branch was rebased onto newer `main` since the last review (same functional commit). The rebase also brings `docs/architecture/cli.md` into the diff — the CLI page now documents the intent-based create behavior and bumps `Last verified:` to the PR date, which resolves what would otherwise be architecture-page drift._

### Findings
_(No new findings introduced since the previous review; the two items above are carried over and remain live. Original inline threads persist from the `87a35be` review.)_

### Verdict
COMMENT — Functionally unchanged rebase of a solid, root-cause fix; the CLI doc is now updated in-PR (good), and the only remaining gap is the missing regression tests for the two new create-path branches.

---

kapetr (approved):

PR #2768, opened 2026-07-13
feat: per-user compute budgets with sandbox sizes, pause, and stop

### LINKED ISSUE #1900: Per-user CPU and memory limits

jezekra1:
Closes #1900. Replaces the approach in #2121 (see that PR's thread for the redesign rationale).

## What

Every user gets a compute budget: a ceiling on the CPU and memory their running sandboxes can use at once. Within it they're free; at it, they pause, stop, shrink, or wait.

- **Sandbox Size** — each sandbox has a user-visible size (its CPU/memory), set with sliders at create, adjustable later in settings (restarts a running sandbox, after a confirm). Templates seed the default; the platform default is small (1 CPU/1Gi).
- **The budget is a hard guarantee** — sizes cap actual usage, so a user's sandboxes can never consume past their ceiling, even all bursting at once. Enforcement lives in the controller at the scale-up transition, the single point every start converges on — no api-server code path can start pods around it.
- **Over-budget starts park, never fail** — a sandbox that doesn't fit waits with an "Over budget" badge (and the exact figures). It stays parked until a deliberate start after room frees — a parked sandbox never grabs freed room out from under whoever freed it. (Never-hibernate sandboxes are the one exception: they declare "always run" and auto-start within ~30s of room freeing.) Wakes against a parked sandbox fail in about a reconcile round-trip with a clear message instead of a 120s timeout.
- **Pause and stop free room** — pause sleeps a sandbox now and it wakes on its next use; stop keeps it down until explicitly started (schedules override a stop, and the UI warns about that). Both are immune to the background-polling resurrection bug that sank #2121.
- **Visibility** — a reserved-vs-ceiling meter on the sandbox list, live over-budget warnings on the size sliders, and per-user ceiling overrides for privileged users via a schema-validated `UserBudget` CR (`budget-<user>`, one per user by construction).

https://github.com/user-attachments/assets/d3d705ee-1f3d-4bde-a067-40943cddb516

## Out of scope (per the issue)

Role-based limits and approval flows; token/cost accounting.

## Ops notes

- **New CRD** (`userbudgets.agent-platform.ai`): Helm does not apply `crds/` on upgrade — existing deployments must apply `deploy/helm/platform/crds/agent-platform.ai_userbudgets.yaml` before rolling the controller (which fail-fasts without it).
- **Existing sandboxes keep their size**: the boot backfill stamps pre-Sizes agents at `controller.agent.legacyAgentSize` (default 1 CPU / 2Gi — the limits they were actually running with), *not* the new smaller default. Operators who overrode the old `templateDefaults.resources.limits` should mirror that override in `legacyAgentSize` before upgrading.
- **Scheduling density**: the new default Size derives a 500m CPU request (limit × 0.5) where the old chart requested 250m, so a packed cluster fits fewer default-size sandboxes. Tune `controller.requestsFromLimits.fraction` if density matters more than burst headroom.
- Design details in `docs/architecture/budgets.md`.


packages/controller/api/v1/user_budget_types.go
```diff
@@ -0,0 +1,64 @@
+package v1
+
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// UserBudgetSpec is one user's Ceiling override: the maximum CPU and memory
+// that user's scaled-up Agents may reserve concurrently, as the sum of their
+// `spec.resources.requests` (#1900). The uniform per-agent gateway overhead
```
dam-code-guardian: 🟡 **Warning:** This comment (and the per-field comments on lines 19/21: "summed CPU/memory **requests**") says the Ceiling is measured against `spec.resources.requests`. But enforcement actually sums `spec.resources.limits` — see `budget.go`'s header comment plus `reservedByOwner`/`limitsOf`, and `docs/architecture/budgets.md` ("Reserved … is the sum of **Sizes** (`spec.resources.limits`)"). The requests-vs-limits distinction is exactly what makes the budget a hard guarantee, so a public CRD type comment asserting the opposite is a real trap for operators and future maintainers.

```suggestion
// `spec.resources.limits` — their user-visible Size (#1900). The uniform per-agent gateway overhead
```

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `473909d`

## PR #2768: feat: per-user compute budgets with sandbox sizes, pause, and stop
**Author:** jezekra1 | **Branch:** feat/1900-user-budgets#locki-73vjrtpw → main | **Changes:** +3022 −159 (83 files)

### Summary
Introduces per-user compute budgets: each sandbox gets a user-visible Size (CPU/memory limits), and the controller enforces a per-owner Ceiling at the single 0→1 scale transition — over-budget starts park (Ready=False/OverBudget) rather than fail, and pause/stop free room. A `UserBudget` CRD (name CEL-pinned to `budget-<owner>`) provides per-user overrides; the api-server owns the display meter and the live-resize gate. The design is coherent and the enforcement-point reasoning (single actuator, single worker goroutine, per-owner mutex belt-and-suspenders) is sound. _(no prior review on file)_

### Findings
- 🟡 **Warning:** The `UserBudgetSpec` doc comment states the ceiling is measured against "the sum of their `spec.resources.requests`", and the per-field comments say "summed CPU/memory **requests**" (`packages/controller/api/v1/user_budget_types.go:10,19,21`). This contradicts the actual enforcement, which sums `spec.resources.limits` — see `budget.go`'s header comment and `reservedByOwner`/`limitsOf`, and `docs/architecture/budgets.md` ("Reserved … is the sum of **Sizes** (`spec.resources.limits`)"). The requests-vs-limits distinction is load-bearing for this feature (limits hard-cap consumption, which is what makes Σ Sizes ≤ Ceiling a real guarantee), so a public CRD type comment claiming the opposite is a genuine trap for the next reader/operator. Recommend s/requests/limits/ in all three spots.
- ✅ **Looks good:** Enforcement is anchored at the single 0→1 actuator with a clear race analysis (single-worker-goroutine invariant + per-owner mutex as documented belt-and-suspenders), and `reservedByOwner` counts *desired* replicas so two near-simultaneous wakes can't both slip under the ceiling.
- ✅ **Looks good:** `limitsOf`/`parseQuantityOr` fall back to chart template defaults per-dimension (never zero) for legacy specs — an unreadable limit can't let an agent slip under the ceiling.
- ✅ **Looks good:** The denied-wake memo keyed on `last-activity` correctly prevents parked agents from spontaneously grabbing freed room, while a fresh activity bump (or always-on timeout=0) re-tries the gate; the in-memory restart caveat is documented and accepted.
- ✅ **Looks good:** The api-server resize gate correctly scopes to size *increases* only (decreases always allowed, even for grandfathered over-ceiling owners) and mirrors the controller's exclusion of hibernated/parked agents from Reserved.

_(doc-drift verdict: **aligned.** New `docs/architecture/budgets.md` subsystem page added and linked from the landing page; `docs/architecture/agent-lifecycle.md` updated with the hard-stop/pause behavior introduced by this PR; all three edited/added architecture pages carry `Last verified: 2026-07-13` (the PR date). No ADR reference leaks, no volatile-content leaks.)_

### Verdict
COMMENT — Well-structured feature with sound enforcement design and in-sync architecture docs; the one warning is a misleading requests-vs-limits comment on the public `UserBudget` CRD type that should be corrected to match the actual limits-based enforcement.

---

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `d77e635`

## PR #2768: feat: per-user compute budgets with sandbox sizes, pause, and stop
**Author:** jezekra1 | **Branch:** feat/1900-user-budgets#locki-73vjrtpw → main | **Changes:** +3037 −159 (85 files)

### Summary
Per-user compute budgets: each sandbox has a user-visible Size (CPU/memory limits) and the controller enforces a per-owner Ceiling at the single 0→1 scale transition — over-budget starts park (Ready=False/OverBudget) rather than fail; pause/stop free room; a `UserBudget` CRD provides per-user overrides. This re-review covers 3 PR-own commits since the last review at `473909d` (the rest of the range is a merge of main + 6 unrelated main PRs, attributed away via merge-base). _(Suppressed 0 findings; no PR-local overrides.)_

### Changes since last review
Previous HEAD: `473909d` (2026-07-13T14:07:52Z) — verdict COMMENT

- ✅ **Fixed:** requests-vs-limits mislabel on the public `UserBudget` CRD type (`packages/controller/api/v1/user_budget_types.go:10,19,21`) — commit `9589365` rewrites all three comments to say "Sizes — `spec.resources.limits`", and the CRD YAML (`agent-platform.ai_userbudgets.yaml`) was corrected to match. The type comment now agrees with the enforcement in `budget.go` and `docs/architecture/budgets.md`.
- 🆕 **New:** none that rise to a finding. The remaining PR-own changes are: a trivial CLI test-fixture fixup (`agent-{get,list}.integration.test.ts` — `overBudget` field added after merging main), doc follow-ups (accepted-races note in `budgets.md`, experiment-wake path in `agent-lifecycle.md`, `values.yaml` comment correction), and one UI headroom-credit fix (below, reviewed clean).

### Findings
- ✅ **Looks good:** The settings size-slider headroom fix (`packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx:64`) credits the sandbox's own current Size back into free headroom (`freeCpu = ceiling − reserved + ownCpu`). This is correct: `sizeRestartsAgent` is true exactly when the agent is up (not hibernated, not over-budget), which is precisely the condition under which its Size is counted in `reserved` (`budgets-service.ts:95`), so crediting it avoids a false "will wait parked" warning when resizing a running sandbox — with no risk of over-crediting headroom for hibernated/parked sandboxes (guarded by `currentSize={f.sizeRestartsAgent ? agent.size : undefined}`).

_(doc-drift verdict: **aligned.** The new PR-own doc edits track the code: the "Accepted races" note in `budgets.md` documents the two sub-second resize/wake self-races the code accepts; `agent-lifecycle.md` adds the experiment-run wake as a stop-clearing path; `values.yaml`'s operator-facing comment was corrected to the Sizes/limits model. All edited architecture pages carry `Last verified: 2026-07-13`. No ADR reference leaks, no volatile-content leaks.)_

### Verdict
APPROVE — The single warning from the prior review (misleading requests-vs-limits CRD comment) is fully resolved; the remaining new changes are a correct headroom-credit refinement, in-sync doc follow-ups, and a trivial test-fixture fixup. No outstanding issues.

_Note: `typescript-engineering` and `react-ui-engineering` skills remain unavailable (harness auto-install regression, ongoing since 2026-06-16 — operator action needed) — those sections are omitted per CLAUDE.md._

---

pilartomas (changes requested):
# Review

## Overview

The design is genuinely good. Enforcement really is funneled through a single actuator — only `applyStatefulSet` writes replicas ≥ 1, reachable only through `Reconcile` after the gate, with the single-worker invariant pinned in code where it would be broken. RBAC is least-privilege (no write verbs on userbudgets anywhere), the CEL name-pin is real admission enforcement (not convention), cross-user access is properly scoped, layering in the TS modules is clean (budgets talks to agents only through structural ports), and `docs/architecture/budgets.md` is unusually honest about its accepted races. Chart lint/render and the Go test suites pass on the head branch.

That said, there are two must-fix bugs and a cluster of should-fixes.

## Must fix

**1. Critical (UI): the settings form breaks after any successful save.** The post-save `reset({...})` at `packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-form.ts:371` was not extended with the new `sizeCpuMilli`/`sizeMemoryMi` fields, and RHF `reset(values)` replaces values wholesale. After the first successful save of *anything* (a rename suffices), both size fields become `undefined`, the sliders snap to the 1000m/1024Mi fallbacks, and every subsequent Save silently no-ops because `z.number().int().positive()` fails on `undefined` with no field rendering the error. The rebaseline effect never re-runs (`baselinedRef` stays true), so the form is dead until remount. (The baseline reset at line 144 has the fields; the post-save reset doesn't.)

**2. Major (api-server): the first wake of a parked sandbox after freeing room fails with a wrong "over budget" error.** `ensureReady` bumps `last-activity` and then polls, and the poll's first tick fires immediately (`packages/api-server/src/modules/agents/infrastructure/agents-repository.ts:344`, `poll-until-ready.ts:20`) — before the controller can have reconciled the fresh bump. Since the controller deliberately leaves `Ready=False/OverBudget` standing on a parked agent, tick 0 reads the stale condition and throws `AgentWakeTimeoutError` near-deterministically, even when room is now free. This defeats the documented recovery path ("free some room, start it again"): the user stops sandbox B, starts sandbox A, gets an over-budget error, and only a second click works. Note for the fix: `lastTransitionTime` won't distinguish a re-denial (False→False doesn't bump it) — the controller probably needs to expose *which* activity value the denial applies to (e.g. echo the denied `last-activity` into the condition/status), so the api-server only fast-fails when the denial postdates its own bump.

## Should fix

**3. Major (UI): silent memory-limit rewrite on a CPU-only resize.** `parseMemoryMi` only accepts `^\d+(Mi|Gi)$` (`packages/ui/src/modules/sandboxes/lib/quantity.ts:9`), but sizes originate from template YAML where `"1.5Gi"`, `"1G"`, `"1024M"` are legal (the server's own parser handles them). A misparse baselines the form to the 1024Mi fallback, and the save payload ships **both** dimensions whenever either slider is dirty (`use-sandbox-settings-form.ts:341`) — so bumping only CPU on a 1.5Gi sandbox silently rewrites its memory to 1Gi and restarts it. Widen the parser to the server's grammar, or send only the dirty dimension.

**4. Major (upgrade behavior): default memory limit silently halved, and the backfill stamps the new default onto legacy agents.** On main the default agent resources were requests 250m/512Mi, limits 1 CPU/**2Gi**; the PR changes the default to limits 1 CPU/**1Gi** with derived requests (`deploy/helm/platform/values.yaml`). The boot backfill then pins every legacy Agent CR lacking limits to the *new* default, so a pre-existing agent whose workload used 1–2Gi OOM-kills on its next wake, silently. The derived CPU request also doubles (250m → 500m) for default-size agents, halving how many fit on a packed cluster. Both may be intentional, but they need at least a loud upgrade note — or a backfill that preserves the old effective 2Gi.

**5. Major (tests): the new coverage is thinnest exactly where the findings live.** The controller's budget accounting matrix is well tested, but there is **zero** controller coverage for `stop-requested` (the `shouldRun` override, hard-stop idempotence, stop stickiness — the PR's headline "resurrection bug" fix); zero api-server tests for the resize gate, quantity parsers, the new `over-budget` wake-failure branch, pause settle, or backfill idempotency; and zero UI tests for the quantity round-trips (which would have caught #3).

**6. Minor concurrency/consistency issues** (each verified, all narrow and mostly self-healing, several contradicting documented invariants):
- A budget-denied retry after a transient mid-reconcile failure can strand the **gateway** StatefulSet at replicas=1 (`applyStatefulSet(…, false)` preserves replicas; `packages/controller/pkg/reconciler/agent_reconciler.go:217`) — an orphan gateway pod until the idle sweep.
- The pause settle-watcher clears `stop-requested` unconditionally once it sees Hibernated (`agents-repository.ts:244`), so a stop stamped during the 60s settle window can be erased and the pair self-revives — compare-and-clear the exact stamp pause wrote.
- `wakeIfHibernated` unconditionally clears the stop annotation, so a stop landing between `ensureReady`'s check and its patch is cancelled with no scale-down and no feedback.
- The resize gate's shrink shortcut evaluates a `current` snapshot fetched *before* the per-owner lock (`packages/api-server/src/modules/agents/services/agents-service.ts:503`) — two rapid resizes can classify an increase as a shrink and skip the ceiling check. Re-read inside the lock.
- `parseQuantityOr` in the controller silently defaults a malformed limit that `toResourceList`'s `MustParse` will then **panic** on in the same reconcile (`budget.go:148` vs `resources.go:511`) — the panic is pre-existing, but the PR adds a second parser with contradictory tolerance for the identical input; an explicit `"0"` limit also counts zero against the budget despite the "fallback, never zero" comment.

**7. Minor (hardening/consistency):**
- CRD quantity patterns admit zero/negative ceilings (`cpu: "-4"` passes admission and permanently parks the user); a CEL sign check would close it. `spec.owner` has no length/charset constraint — non-UUID IdP subs can make a UserBudget unconstructible.
- Both RBAC roles grant `list` on userbudgets but both components only ever `get` by pinned name; drop `list` (and fix the controller RBAC comment that says "live list").
- The new config quantity strings are plain `z.string()` at boot (`packages/api-server/src/config.ts`); a typo'd Helm value (`cpu: "1Gi"`) propagates into every created spec and zeroes the meter ceiling — crash-at-boot validation is cheap here.
- `computeAgentState` has no parked branch, so the CLI shows a parked agent as `starting` forever.
- UI: budget invalidation is asymmetric (pause/stop invalidate `budgets.reserved`; wake/restart/delete/resize don't — stale headroom for up to the 5s poll); `queryClient.fetchQuery(trpc.schedules.list…)` directly inside `packages/ui/src/modules/agents/views/list-view.tsx:79` violates the "fetchers live in `modules/{domain}/api/`" rule and crosses module boundaries.
- A11y: the sliders announce raw millicores/Mi with no `aria-valuetext`; the over-budget figures ride a `title` attribute on a non-focusable span — unreachable by keyboard/touch, despite being a stated feature of the PR.

**8. PR description mismatch.** The body says a parked sandbox "starts by itself within ~30s of room freeing" — the implementation (and `budgets.md`, deliberately) does the **opposite**: only never-hibernate agents auto-start; ordinary parked sandboxes wait for a fresh deliberate start so they don't grab freed room out from under whoever freed it. The docs are right and the behavior is the better one; the PR body should be corrected since it's the reviewers' and release notes' source of truth.

## Nits

Duplicated `sizeDirtyNow`/`sizeDirty` computation and a dead `disabledHint={undefined}` prop; "Wake" menu label for a parked (not hibernated) sandbox; `formatCores` shows 250m as "0.3 cores"; slider grid can't reach an off-grid operator ceiling exactly; the per-owner mutex in `budget.go` protects nothing under the single-worker invariant (its own comment says so — consider removing); `effectiveIdleTimeout` computed twice back-to-back; chart-default literals (1 CPU/1Gi, fraction 0.5, floors) restated across two docs pages and Go env fallbacks; `budgets.md` example hardcodes the `platform-agents` namespace, which is configurable.

## Verdict

This is careful, well-documented work — the enforcement invariant is real, the docs match the code, and security scoping is correct. Requesting changes for: #1 breaks the primary settings flow outright, #2 breaks the primary recovery flow for the feature's own headline scenario, #3 silently rewrites user data, and #4 is a silent behavioral change for every existing deployment. The test gaps (#5) sit precisely on top of these, so closing them together would be natural.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `670bf2c`

## PR #2768: feat: per-user compute budgets with sandbox sizes, pause, and stop
**Author:** jezekra1 | **Branch:** feat/1900-user-budgets#locki-73vjrtpw → main | **Changes:** +3915 −187 (90 files)

### Summary
Per-user compute budgets: each sandbox has a user-visible Size (CPU/memory limits) and the controller enforces a per-owner Ceiling at the single 0→1 scale transition — over-budget starts park (`Ready=False/OverBudget`) rather than fail; pause/stop free room; a `UserBudget` CRD provides per-user overrides. This re-review covers the 4 PR-own commits since the last review at `d77e635`, which land the fixes for the whole-feature human review (@pilartomas, `CHANGES_REQUESTED` 2026-07-14). Every must-fix, should-fix, and nit from that review is addressed, with new unit tests targeting exactly the gaps that were called out. _(Suppressed 0 findings; no PR-local overrides. No findings suppressed by PR context.)_

### Changes since last review
Previous HEAD: `d77e635` (2026-07-13T14:58:49Z) — verdict APPROVE

Between the prior review and this HEAD, a human reviewer (@pilartomas) requested changes with two must-fix bugs, six should-fix items, and nits. The 4 new commits (`8f83793`, `589acd6`, `0fcf3b5`, `670bf2c`) resolve them:

- ✅ **Fixed (must-fix #1, UI):** post-save `reset({…})` now includes `sizeCpuMilli`/`sizeMemoryMi` (`use-sandbox-settings-form.ts:374`), so the settings form no longer dies (fields → `undefined`, silent no-op saves) after the first successful save.
- ✅ **Fixed (must-fix #2, api-server):** the parked-sandbox wake now uses a proper handshake — `bumpLastActivity` returns its stamp, the controller echoes the denied `last-activity` into `status.deniedActivity` (`budget.go`, `agent_types.go`), and `ensureReady` fails fast only when the standing denial is at or after its own bump (`agents-repository.ts:368+`). The first wake after freeing room no longer near-deterministically throws a stale "over budget" error.
- ✅ **Fixed (should-fix #3, UI):** `parseMemoryMi`/`parseCpuMilli` widened to the real K8s quantity grammar (`1.5Gi`/`1G`/`1024M`, Ki/Mi/Gi/Ti + k/M/G/T) and the save now ships only the dirty dimension (`quantity.ts`, `use-sandbox-settings-form.ts:341+`) — a CPU-only resize no longer silently rewrites memory.
- ✅ **Fixed (should-fix #4, upgrade):** new `controller.agent.legacyAgentSize` (default 1 CPU / 2Gi) is what the boot backfill stamps onto pre-Sizes agents (`config.ts`, `backfill-agent-resources.ts`, `values.yaml`) — legacy workloads keep their real limits instead of being silently halved to the new 1Gi default. Loud upgrade notes added to `values.yaml`.
- ✅ **Fixed (should-fix #5, tests):** new unit coverage for the stop/pause/backfill/resize-gate/quantity gaps — `budget_test.go` (+167), `backfill-agent-resources.test.ts` (+89), `budgets-resize-gate.test.ts` (+115), `agents-repository-ensure-ready.test.ts` (+117), `sandbox-quantity.test.ts` (+67).
- ✅ **Fixed (should-fix #6, concurrency):** parked branch now scales the pair to zero to heal a stranded gateway (`agent_reconciler.go`, `scaleAgentPairToZero`); pause settle-watcher compare-and-clears the exact stamp it wrote (`agents-repository.ts:246+`); resize gate re-reads `current` **inside** the per-owner lock (`agents-service.ts:503+`); `parseQuantityOr` and `toResourceList` now share identical fallback-on-non-positive tolerance (`budget.go`, `resources.go`) so the panic-vs-default divergence is gone.
- ✅ **Fixed (should-fix #7, hardening):** CRD CEL now rejects zero/negative CPU/memory ceilings and constrains `spec.owner` to a DNS-1123-constructible pattern (`user_budget_types.go`); RBAC drops unused `list` on userbudgets for both components; boot config validates quantity strings (`positiveQuantitySchema`); `computeAgentState` gains an `over_budget` branch so the CLI/UI no longer shows a parked agent as perpetually `starting`; `list-view.tsx` module-boundary violation replaced with `fetchSchedulesForAgent` in the schedules module; slider gains `aria-valuetext` and the over-budget badge is now focusable/labelled.
- ✅ **Fixed (#8, PR body):** the description's "auto-starts within ~30s" claim reconciled with the actual (and better) parked-waits-for-deliberate-start behavior; docs already matched.
- 🆕 **New:** none that rise to a finding. The new commits are exactly the review-fix set above plus the matching doc updates; all reviewed clean.

### Findings
- ✅ **Looks good:** The wake handshake is coherent end-to-end — controller stamps `status.deniedActivity` on denial (`budget.go`) and clears it on both admit (`agent_reconciler.go` `publishReadiness`) and hibernate-restamp (`idlechecker.go` `hibernateAgentPair`), while the api-server orders RFC3339 stamps lexicographically to distinguish "my attempt was denied" from "stale denial, keep polling". No window where a freshly-bumped wake fast-fails on a prior attempt's condition.
- ✅ **Looks good:** The `over_budget` display state is handled at every render site (`status-indicator.tsx` label + variant maps, `agent-unavailable-overlay.tsx` `Record<AgentDisplayState>`, the Wake→Start menu label, `resolveAgentDisplay`) — no non-exhaustive switch left behind by adding the new state.
- ✅ **Looks good:** RBAC `list`-drop is safe: `ceilingFor` reads the budget via a direct `Get(ctx, "budget-"+owner, …)` (`budget.go:172`), never a list; the comments were corrected to match.
- ✅ **Looks good:** The backfill/limit fallback story is now internally consistent — `parseQuantityOr` (budget accounting) and `toResourceList` (pod render) apply the same "malformed OR non-positive → chart default" rule, so enforcement can never count a value the pod doesn't actually run with.

_(doc-drift verdict: **aligned.** The two edited architecture pages track the code fixes precisely: `budgets.md` documents the wake handshake (`status.deniedActivity`), the parked-branch scale-to-zero, the read-inside-lock resize serialization, and the `legacyAgentSize` backfill; `agent-lifecycle.md` documents the pause compare-and-clear. Both carry `Last verified: 2026-07-14` (the PR update date); the landing page `docs/architecture.md` links `budgets.md`. No ADR reference leaks; no volatile-content leaks — the `namespace: platform-agents` hardcode was even generalized to `<agents-namespace>`.)_

### Verdict
APPROVE — The 4 new commits comprehensively resolve the human reviewer's two must-fix bugs (settings-form reset, parked-wake false over-budget), all should-fixes (silent memory rewrite, silent legacy-agent shrink, the test gaps, and the concurrency cluster), and the hardening/a11y nits, with new unit tests landing exactly on the previously-untested branches. Architecture docs are in sync.

_Note: `typescript-engineering` and `react-ui-engineering` skills remain unavailable (harness auto-install regression, ongoing since 2026-06-16) — those sections are omitted per CLAUDE.md. The TS/JS and UI buckets are both non-empty this PR, so this omission does drop skill-based coverage of the `.ts`/`.tsx` changes; findings above are from manual review._

---

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `4dbdb0b`

## PR #2768: feat: per-user compute budgets with sandbox sizes, pause, and stop
**Author:** jezekra1 | **Branch:** feat/1900-user-budgets#locki-73vjrtpw → main | **Changes:** +3854 −179 (89 files)

### Summary
Per-user compute budgets: each sandbox has a user-visible Size (CPU/memory limits); the controller enforces a per-owner Ceiling at the single 0→1 scale transition — over-budget starts park (`Ready=False/OverBudget`) rather than fail; pause/stop free room; a `UserBudget` CRD provides per-user overrides. This re-review covers the 2 PR-own commits since the last review at `670bf2c` (`305c97e`, `4dbdb0b`) — both further rework, one architectural simplification and one live-verified admission fix.

### Changes since last review
Previous HEAD: `670bf2c` (2026-07-14T11:29:22Z) — verdict APPROVE

- 🆕 **New (simplification):** `305c97e` replaces the over-budget wake fail-fast **cross-component handshake** with a **local heuristic**. The Agent-status `deniedActivity` echo (and the whole `AgentStatus.DeniedActivity` field, its clears in `publishReadiness`/`hibernateAgentPair`, and `AgentSchemaGeneration` 5→4) is gone; the api-server now treats a standing `OverBudget` as its own denial only when it *observed the condition appear during its own poll* (fresh verdict → fail immediately) or once a `OVER_BUDGET_FAIL_FAST_GRACE_MS` (10s) grace passes with the refusal still standing (`agents-repository.ts`, `poll-until-ready.ts`). Trades a CRD field + cross-component contract for one recoverable premature error on a rare grace-miss — a sound call for this failure mode. `agent-mappers.ts` drops the field in step; no dangling references remain.
- 🆕 **New (fix):** `4dbdb0b` changes the UserBudget positivity CEL from `quantity(...).sign() == 1` to `.isGreaterThan(quantity('0'))` because `sign()` fails CEL compilation on the targeted API servers (verified live: k3s v1.36 rejected the CRD with "found no matching overload", breaking `cluster:install`). `isGreaterThan` is in the base quantity library and compiles; admission re-verified live across negative/zero/mismatch/valid. `UserBudgetSchemaGeneration` correctly folded 2→1 (this CRD ships first with this PR, never released).
- ✅ **Fixed:** nothing regressed; all prior-review resolutions remain in place.

### Findings
- ✅ **Looks good:** The wake heuristic is correct on all three branches and the tests were rewritten to match exactly — ride-through-stale (admits once room frees), fail-fast on a condition transition observed during the wake (`durationMs < 10_000`), and fail-fast after the grace on a standing refusal (`10_000 ≤ durationMs < 120_000`). `sawNotOverBudget` cleanly encodes "the controller ruled on us."
- ✅ **Looks good:** `deniedActivity` removal is complete and consistent across the status type (`agent_types.go`), controller writers (`budget.go`, `agent_reconciler.go`, `idlechecker.go`), the TS mapper (`agent-mappers.ts`), and the schema-generation gate — no half-removed field.
- ✅ **Looks good:** Both `crd-schema-generation` decrements (Agent 5→4, UserBudget 2→1) are legitimate pre-release folds, not backwards-incompatible bumps — Agent gen 5 (`deniedActivity`) never shipped, and the UserBudget CRD is brand new in this PR.
- ✅ **Looks good:** The CEL fix carries a live-verification note in both the Go marker comment and the rendered CRD YAML, so the next reader won't "restore" the non-compiling `sign()` form.

_(doc-drift verdict: **aligned.** `budgets.md` was rewritten in step with `305c97e` — the "Waking a parked agent" section now describes the **heuristic** (grace window + observed-transition), explicitly noting the local-vs-CRD-field tradeoff, replacing the old "handshake"/`status.deniedActivity` prose. `agent-lifecycle.md` (stop/pause) and the landing page's budgets link carry `Last verified: 2026-07-14`/`2026-07-13`. No ADR reference leaks, no volatile-content leaks, no new-subsystem-without-a-page.)_

### Verdict
APPROVE — Two clean follow-up commits: an architectural simplification that removes a CRD field and a cross-component contract in favor of a well-tested local heuristic, and a live-verified CEL compilation fix that unblocks `cluster:install`. Tests and architecture docs track both changes precisely.

---

pilartomas (approved):

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `cd99fb0`

## PR #2768: feat: per-user compute budgets with sandbox sizes, pause, and stop
**Author:** jezekra1 | **Branch:** feat/1900-user-budgets → main | **Changes:** +4022 −183 (87 files)

### Summary
Per-user compute budgets: each sandbox has a user-visible Size (CPU/memory limits); the controller enforces a per-owner Ceiling at the 0→1 scale transition — over-budget starts park (`Ready=False/OverBudget`) rather than fail; pause/stop free room; a `UserBudget` CRD provides per-user overrides. This re-review covers the 3 PR-own commits since the last review at `4dbdb0b` (`998509b`, `df5bd45`, `cd99fb0`), which relocate legacy-Size backfill into the controller and complete resize enforcement. _(Suppressed 0 findings; no PR-local overrides. No findings suppressed by PR context.)_

### Changes since last review
Previous HEAD: `4dbdb0b` (2026-07-14T12:42:41Z) — verdict APPROVE

- 🆕 **New (architecture):** `df5bd45` moves legacy-Size materialization from the api-server's one-time boot backfill (`backfill-agent-resources.ts`, now **deleted**) into the controller: `ensureConcreteSize` fill-if-absent at the top of `Reconcile` writes `legacyAgentSize` into any spec missing a `resources.limits` dimension, never touching a set value. This is strictly better — watch-driven, so it also converges CRs created out-of-band (kubectl/GitOps/restore), which a boot backfill never could — and gives limits effectively-required semantics without schema enforcement. The api-server stays the sole writer of user *intent*; the controller writes only a system default into unset fields (the `spec.nodeName` analogy in the comment is apt).
- 🆕 **New (enforcement):** `998509b` adds the controller live-resize gate (`resizeAllows`): an UP agent whose rendered limits **grew** past the Ceiling parks — scaled to zero *before* the new template applies, so a pod at the denied size never launches. Grow-only and diff-keyed against the live StatefulSet template, so resyncs and ceiling changes never re-check a running agent. The api-server's synchronous resize rejection is correctly demoted to a UX courtesy in front of this authoritative gate; out-of-band spec writes can no longer grow a running agent around the Ceiling.
- 🆕 **New (RBAC):** `cd99fb0` grants the controller `patch` on `agents` — the one controller spec write, scoped and documented to the fill-if-absent materialization.
- 🆕 **New:** none that rise to a finding — the new logic is covered by new unit tests (`TestReconcileMaterializesLegacySizeIntoSpec`, `TestReconcileFillsOnlyAbsentSizeDimensions`, `TestResizeGrowOfUpAgentPastCeilingParks`, `TestResizeShrinkOfUpAgentNeverParks`, `TestResizeUnchangedAgentNeverRecheckedOnCeilingDrop`, `TestBudgetLegacyPeerCountsAtLegacySize`).
- ✅ **Fixed:** nothing regressed; all prior-review resolutions remain in place, and the human reviewer (@pilartomas) moved from `CHANGES_REQUESTED` to `APPROVED` at 2026-07-14T14:10Z.

### Findings
_No new findings. The three commits are a clean architectural relocation (backfill → controller fill-if-absent), a completeness fix (live-resize render gate), and its RBAC grant — each with targeted new tests and matching doc updates. Prior-review `✅` items are not repeated here per review preferences._

✅ No documentation drift detected. `budgets.md` is rewritten in step with the code: the Size section now describes controller-side fill-if-absent materialization (replacing the boot-backfill prose), the resize section describes the controller render gate as authoritative with the api-server as courtesy, and the former "Accepted races" section is reframed as "Self-healing races" (every slip is a spec change the controller re-gates). `Last verified: 2026-07-14`. No ADR reference leaks, no volatile-content additions, no new subsystem needing a page (controller already covered in `platform-topology.md`).

### Verdict
APPROVE — The new commits relocate backfill to a superior watch-driven controller mechanism, complete resize enforcement at the render gate, and grant exactly the RBAC that requires — all tested and documented. The prior human `CHANGES_REQUESTED` is now `APPROVED`; no outstanding issues.

---

PR #2780, opened 2026-07-14
fix(ui): correct unread state on the sessions list

### LINKED ISSUE #2771: Terminal sessions keep flagging themselves as unread with no new activity
## Problem

The sessions list marks a session **unread** when something has happened in it since the last time you looked. For chat sessions this works correctly.

For **terminal** sessions it misfires: a terminal you've already opened and viewed keeps popping back to unread even though you started nothing new in it. Dismissing it doesn't stick — it re-flags itself.

The cause is that a terminal's "unread" is decided by comparing two times that are tracked by two different systems and never kept in sync:

- **"When you last looked"** is recorded by the app when you open or leave the terminal.
- **"When something last happened"** is taken straight from the underlying tool the terminal runs, which writes to its own session file. The app doesn't control when that file gets touched.

Because these two clocks drift apart, the session file can get touched *after* the app recorded that you looked at it — without you doing anything in the UI. The moment that happens, "something happened" looks newer than "you looked," and the terminal shows unread again. (Chat sessions avoid this because the same code stamps both times off one clock.)

The result is a persistent, false unread badge on terminal sessions that undermines trust in the indicator: once it cries wolf, users stop believing any unread mark.

## Proposed solution

Make a terminal's unread state depend only on activity the app itself can see and account for, so that "when something happened" and "when you looked" are measured consistently. After a user has viewed a terminal and nothing new occurs, it should stay read until genuine new output arrives while they're away.

### Out of scope

Deciding whether idle terminal output (e.g. a background task finishing while you're away) *should* count as unread — that's a product question to settle separately. This issue is only about eliminating the false unread that appears with no activity at all.

### LINKED ISSUE #2779: Chat session doesn't show as unread after switching to a terminal session
## Problem

The sessions list marks a session **unread** when a turn completes in it while you're not looking. This doesn't work when you switch from a chat session to a terminal session.

Concrete scenario: send a prompt to a chat session, switch to a terminal session, and let the chat's turn finish. The chat should badge unread — you weren't watching when it completed — but it stays read.

The cause is that switching to a terminal session doesn't close the chat's live connection. The runtime still sees a viewer attached to the chat, so when the turn completes it marks the chat as "seen" and it never surfaces as unread. (Switching between two chat sessions tears the connection down correctly; it's specifically landing on a terminal that strands it.)

## Proposed solution

When the user leaves a chat session for a terminal session, drop the chat's live connection so the runtime no longer counts them as viewing it. A chat turn that finishes while they're elsewhere then correctly shows unread.

kapetr:
Two small fixes to the sessions-list unread indicator:

- Chat sessions now go unread after you switch to a terminal — the chat's live channel is torn down on the switch, so its turn completes with no viewer attached. Closes #2779
- Terminal sessions no longer falsely show unread — their `updatedAt` follows the harness file mtime (bumped by pod restarts and TUI repaints), which isn't a reading signal, so terminals are excluded from the unread check. Closes #2771

### DIFF
--- a/packages/ui/src/modules/sessions/components/sessions-sidebar.tsx
+++ b/packages/ui/src/modules/sessions/components/sessions-sidebar.tsx
@@ -120,10 +120,12 @@ export function SessionsSidebar({
           const needsApproval =
             approvalSessions.has(s.sessionId) ||
             pendingPermissions.some((p) => p.sessionId === s.sessionId);
-          // No stamp means the runtime never tracked the session (legacy,
-          // harness-minted) — read, not unread.
+          // Terminals have no meaningful unread — their updatedAt tracks the
+          // harness file mtime (bumped by restarts and TUI repaints), not
+          // reading. No seenAt means an untracked (legacy) session — also read.
           const unread = Boolean(
             !isOpen &&
+            s.mode !== SessionMode.Terminal &&
             s.seenAt &&
             s.updatedAt &&
             Date.parse(s.updatedAt) > Date.parse(s.seenAt),
--- a/packages/ui/src/modules/sessions/hooks/use-acp-connection.ts
+++ b/packages/ui/src/modules/sessions/hooks/use-acp-connection.ts
@@ -291,11 +291,13 @@ export function useAcpConnection(
     setState("idle");
   }, [clearEngagement]);
 
-  // Tear the connection down and drop the engagement binding — otherwise the channel stays bound to the gone session
+  // Tear the channel down when we leave the session — including switching to a
+  // terminal (no ACP), which otherwise stays engaged and marks the chat "seen"
+  // on turn completion, so it never shows unread.
   useEffect(() => {
-    if (sessionId) return;
+    if (sessionId && sessionMode !== SessionMode.Terminal) return;
     reset();
-  }, [sessionId, reset]);
+  }, [sessionId, sessionMode, reset]);
 
   return { state, ensureLive, connectionRef, reset };
 }


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `58f9f6b`

## PR #2780: fix(ui): show chat as unread after switching to a terminal session
**Author:** kapetr | **Branch:** fix/chat-unread-after-terminal-switch → main | **Changes:** +5 −3 (1 file)

### Summary
Closes the chat's ACP channel when the user switches to a terminal session, so the chat's turn completes with no engaged viewer and correctly surfaces as unread (fixes #2779). The teardown `useEffect` now also fires when `sessionMode === SessionMode.Terminal`, and `sessionMode` is added to its dependency array.

### Findings
- ✅ **Looks good:** The teardown guard now mirrors the keep-alive effect's terminal guard (`use-acp-connection.ts:270`), so the two effects stay consistent — a terminal session never opens a live channel and now also tears down any channel left over from the session we switched away from.
- ✅ **Looks good:** `sessionMode` is correctly added to the effect's dependency array (`use-acp-connection.ts:300`), so the teardown re-runs when the mode transitions to `Terminal` on the same `sessionId`.
- ✅ **Looks good:** `reset()` is idempotent (closes the WS, clears engagement, resets reconnect state) — calling it on a terminal switch where there may be no live connection is harmless.

### Verdict
APPROVE — Small, correct client-side fix that closes the unread-on-terminal-switch gap by tearing down the ACP engagement, consistent with the existing terminal guard on the keep-alive effect.

---

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `916477e`

## PR #2780: fix(ui): correct unread state on the sessions list
**Author:** kapetr | **Branch:** fix/chat-unread-after-terminal-switch → main | **Changes:** +9 −5 (2 files)

### Summary
Two unread-indicator fixes for the sessions list: (1) chat sessions now go unread after switching to a terminal — the ACP channel is torn down on the switch so the turn completes with no viewer (#2779, reviewed at 58f9f6b); (2) terminal sessions are now excluded from the unread check, since their `updatedAt` follows harness file mtime (pod restarts, TUI repaints) rather than a reading signal (#2771).

### Changes since last review
Previous HEAD: 58f9f6b (2026-07-14T07:31:10Z) — verdict APPROVE

- 🆕 **New:** Terminals excluded from the unread computation (`packages/ui/src/modules/sessions/components/sessions-sidebar.tsx:128`) — added by commit 916477e.

### Findings
- 🆕 ✅ **Looks good:** The new guard `s.mode !== SessionMode.Terminal` reuses the same `SessionMode.Terminal` discriminator already applied to the `working` computation a few lines above (`sessions-sidebar.tsx:114`), so the enum is imported and the check is type-safe and consistent. The accompanying comment correctly explains why terminal `updatedAt` is not a reading signal.
- ✅ **Looks good:** The two commits compose cleanly — the teardown fix (58f9f6b) makes chat go unread on terminal switch, and this commit stops terminals themselves from falsely showing unread; neither regresses the other.

### Verdict
APPROVE — Small, correct, well-commented client-side fix. The terminal exclusion is consistent with the existing `SessionMode.Terminal` handling in the same component, and composes cleanly with the previously-approved teardown fix.

---

PetrBulanek (approved):

PR #2753, opened 2026-07-10
fix(ui): drop staged grant when disconnecting a connection in sandbox settings

jezekra1:
## Problem

On the sandbox settings page, disconnecting a granted connection and then creating a replacement of the same type made **Save** fail with:

> Failed to update app connections: connection conn-… not owned by caller

`useDisconnectConnection` deletes the connection globally, but the settings form kept the deleted id staged in `assignedAppIds`. Creating the replacement marked the field dirty, so Save sent the stale id to `connections.setAgentConnections`, which (correctly) rejects ids the caller no longer owns.

## Fix

After a confirmed disconnect, drop the staged grant (`onToggleGrant(id, false)`) — exactly what the `useDisconnectConnection` doc comment calls for, and what the create wizard (`connections-step.tsx`) and the provider section (`onProviderRemoved`) already do. Also adds the `connection-grant-<id>` test id to the settings rows, matching the wizard.

## Regression test

New Playwright spec `10-connection-regrant.spec.ts` drives the exact repro: grant a dedicated connection → disconnect it in the settings page → create a same-type replacement → Save → assert the new grant lands and the stale id never errors. Wired as a last-running `connection-regrant` project so its connection churn can't disturb the injection/slack suites.

## Bonus 1: re-enable the dead 09-user-env spec

`09-user-env.spec.ts` matches no Playwright project — it landed in #1899 without a config entry, so it has never run anywhere. This PR adds the missing `user-env` project (pure-API spec, so no `storageState`, following the `api-keys` pattern). Its runtime assumptions were re-verified against current code: `e2e.getEnv` → mock agent `scriptedMock.getEnv`, idle-harness recycle on env change in `acp-runtime.ts`, and the `dummy-placeholder` fallback in `envoy.go`.

## Bonus 2: CI silently skipped e2e on most PRs

The `e2e` job's plain `if` got the implicit `success()`, which is false when any **transitive** dependency was skipped. `merge-platform-base` skips on PRs that don't touch platform-base, so `mise e2e` skipped on every such PR (compare: it ran on the platform-base-touching PR run [29085217857](https://github.com/dam-agents/dam/actions/runs/29085217857), skipped on this PR's earlier runs). `build-mock` already carried the `!cancelled() && !failure()` override for exactly this; `e2e` one level down did not. The condition now breaks the implicit `success()` with `!cancelled()` and requires its two direct needs explicitly — which also stops unrelated agent-image failures from skipping e2e.

[...]

PR #2782, opened 2026-07-14
fix: heal main CI — e2e deep-link failure, cross-spec contamination, scheduled-run blind spot

JanPokorny:
Main has been red since #2753 merged (its own final CI run already showed the failure). Three independent problems, root-caused from the CI artifacts — see the failing runs' Playwright page snapshot for the smoking gun.

## 1. Deep links render the wrong view after the OIDC roundtrip (product bug)

`10-connection-regrant.spec.ts` deep-links to `/sandboxes/:id` — the only spec that deep-links at all. The artifact's page snapshot shows it landed on the sandboxes **list**:

- The OIDC user lives in `sessionStorage`, which a fresh tab (or Playwright context) doesn't have, so every fresh context bounces through Keycloak and returns on `/auth/callback`.
- The zustand store initializes `view` from that path → `"list"`.
- `initAuth` then `history.replaceState`s back to the deep link — but `replaceState` fires no `popstate`, and the mount effect in `app.tsx` only re-synced `/chat/…` paths.

**Fix:** run the full popstate handler once at mount, so every view syncs from the URL. The spec's heading assertion also gets `level: 1` — it previously passed on the wrong page because the list card shows the agent name as an `h2`.

## 2. The failed regrant spec poisoned 05-injection

The regrant connection reused the injection fixture's host (`httpbingo.org`) + header (`x-api-key`), overwriting the injection credential in Envoy on the shared agent. When the spec died before its cleanup step, `05-injection` deterministically received `e2e-regrant-secret-5e1c` instead of its sentinel.

**Fix:** dedicated `regrant.example.com` + `x-regrant-key`, and the project now depends on `slack` — listing order does not schedule (the spec actually ran *before* injection despite the "listed last" comment).

## 3. Scheduled runs republish all images without testing them

The nightly run rebuilds and republishes every image (`RESOLVE_NO_REUSE`) but skipped `mise test` and `mise e2e` — the 07-13 scheduled run reported green while main was broken.

**Fix:** drop the schedule exclusion from `test`/`build-mock`/`e2e`. Tags stay excluded (the tagged commit was already tested on main).

## Bonus: `exec-server-runtime-env.test.ts` flake

Failed three consecutive main runs with "server never became reachable": cold tsx boot on a loaded runner exceeded the 10s reachability poll (the sibling test passed once the server came up late). Widened to 25s, within the 30s test timeout.

## Verification

- `mise run check` green; UI unit tests (64) and the exec-server test pass locally.
- e2e not run locally (slow); this PR's CI run exercises the whole chain — including the deep-link fix, since the regrant spec is itself the deep-link regression test.

Process note: #2753 was merged with a failing CI run. If `mise e2e` isn't a required status check on `main`, this will recur.

### DIFF
--- a/packages/ui/src/app.tsx
+++ b/packages/ui/src/app.tsx
@@ -120,9 +120,7 @@ function MainApp() {
         });
       else leaveChat();
     };
-    // Handle initial URL (e.g. direct link to /chat/foo) — setState to avoid pushing duplicate history
-    const path = window.location.pathname;
-    if (path.startsWith("/chat/")) enterChat(decodeURIComponent(path.slice(6)));
+    onPopState();
     window.addEventListener("popstate", onPopState);
     return () => window.removeEventListener("popstate", onPopState);
   }, []);
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -42,7 +42,7 @@ jobs:
 
   test:
     name: mise test
-    if: github.event_name != 'schedule' && !startsWith(github.ref, 'refs/tags/v')
+    if: ${{ !startsWith(github.ref, 'refs/tags/v') }}
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
@@ -544,7 +544,7 @@ jobs:
   build-mock:
     name: build mock (e2e fixture)
     needs: [merge-platform-base, changes]
-    if: ${{ github.event_name != 'schedule' && !startsWith(github.ref, 'refs/tags/v') && !cancelled() && !failure() }}
+    if: ${{ !startsWith(github.ref, 'refs/tags/v') && !cancelled() && !failure() }}
     runs-on: ubuntu-latest
     timeout-minutes: 20
     steps:
@@ -578,8 +578,7 @@ jobs:
     # most PRs. Require the direct needs explicitly instead: e2e pulls exactly
     # the images those two jobs push.
     if: >-
-      ${{ github.event_name != 'schedule' &&
-          !startsWith(github.ref, 'refs/tags/v') &&
+      ${{ !startsWith(github.ref, 'refs/tags/v') &&
           !cancelled() &&
           needs.merge-images.result == 'success' &&
           needs.build-mock.result == 'success' }}
--- a/packages/agent-runtime/src/__tests__/unit/exec-server-runtime-env.test.ts
+++ b/packages/agent-runtime/src/__tests__/unit/exec-server-runtime-env.test.ts
@@ -28,7 +28,7 @@ async function connectExec(argv: string[], extra = ""): Promise<WebSocket> {
     encodeURIComponent(Buffer.from(JSON.stringify(argv)).toString("base64")) +
     `&cwd=${encodeURIComponent(homeDir)}&cols=80&rows=24` +
     extra;
-  for (let i = 0; i < 100; i++) {
+  for (let i = 0; i < 250; i++) {
     try {
       return await new Promise<WebSocket>((resolve, reject) => {
         const ws = new WebSocket(url);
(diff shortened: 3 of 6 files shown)


packages/ui/src/app.tsx
```diff
@@ -120,9 +120,7 @@ function MainApp() {
         });
       else leaveChat();
     };
-    // Handle initial URL (e.g. direct link to /chat/foo) — setState to avoid pushing duplicate history
-    const path = window.location.pathname;
-    if (path.startsWith("/chat/")) enterChat(decodeURIComponent(path.slice(6)));
+    onPopState();
```
dam-code-guardian, line 123: 🟢 **Suggestion:** Running `onPopState()` at mount to sync every view from the URL is a solid fix (it also covers the post-OIDC `history.replaceState` case, which fires no `popstate`). The removed comment carried the rationale for the old branch — worth keeping a short note here so the intent survives:

```suggestion
    // Sync the view from the initial URL at mount — covers deep links like
    // /sandboxes/:id and the post-OIDC replaceState, which fires no popstate.
    onPopState();
```

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `b7fff0a`

## PR #2782: fix: heal main CI — e2e deep-link failure, cross-spec contamination, scheduled-run blind spot
**Author:** JanPokorny | **Branch:** fix/main-ci-failures → main | **Changes:** +28 −25 (6 files)

### Summary
Repairs main-branch CI after #2753 by fixing three independent problems: a real product bug where deep links render the wrong view after the OIDC roundtrip (`app.tsx` only re-synced `/chat/…` paths at mount), cross-spec contamination from the regrant spec reusing another fixture's host/header, and scheduled CI runs republishing images without running `mise test`/`e2e`. Plus a flake widening for the exec-server reachability poll. The PR body root-causes each from CI artifacts and the changes match those root causes.

### Findings
- 🟢 **Suggestion:** The mount effect now calls `onPopState()` to sync every view from the URL — a genuine improvement over the old `/chat/`-only branch. Consider a one-line comment explaining *why* it runs at mount (initial URL sync, incl. post-OIDC `replaceState` which fires no `popstate`), since the removed comment carried that rationale (`packages/ui/src/app.tsx:123`).
- ✅ **Looks good:** Deep-link fix is correct — `onPopState` uses `setState` on every branch (no `pushState`), so running it once at mount syncs the view without polluting history, covering `/sandboxes/:id` and all other non-chat paths, not just `/chat/`.
- ✅ **Looks good:** exec-server poll widened to 250×100ms = 25s, comfortably inside the 30s (`}, 30_000`) test timeout — addresses cold-boot flake without risking a hang (`packages/agent-runtime/src/__tests__/unit/exec-server-runtime-env.test.ts:31`).
- ✅ **Looks good:** Regrant spec now uses a dedicated `regrant.example.com` / `x-regrant-key` and the project `dependencies: ["slack"]` instead of relying on listing order — removes the shared-Envoy credential clobbering that poisoned the neighbouring spec (`packages/e2e/playwright/playwright.config.ts:66`, `10-connection-regrant.spec.ts`).
- ✅ **Looks good:** `09-user-env.spec.ts` now preserves the template-seeded `baselineEnv` across `agents.update` calls (which replace the whole env list) and restores it on clear — prevents wiping the mock's default reply. Baseline is fetched once from the running agent and threaded through each mutation correctly.
- ✅ **Looks good:** Dropping the `github.event_name != 'schedule'` guard from `test`/`build-mock`/`e2e` (while keeping the tag exclusion) makes nightly scheduled runs actually exercise what they republish, closing the green-while-broken blind spot (`.github/workflows/cd.yml:45,547,581`).

### Verdict
APPROVE — Three CI failures are correctly root-caused and fixed; the deep-link change is a real product-bug fix that's more robust than the code it replaces, and the test/CI changes are sound. Heed the author's process note: consider making `mise e2e` a required check on `main` so a red CI run can't merge again.

---

PR #2800, opened 2026-07-15
feat(ui): move files and folders by drag and drop in the Files tree

### LINKED ISSUE #2795: UI - Move files and folders by drag and drop in the Files tree
## Problem

The chat's Files section shows the sandbox workspace as a tree, and dragging files in from your computer uploads them into the right folder. But items already in the workspace can't be dragged around: to reorganize, you have to rename an entry with a full new path or ask the agent to move things for you. For simple housekeeping — dropping a stray file into the right folder — that's slow and clumsy.

## Proposed solution

Dragging a file or folder within the tree and dropping it onto a folder moves it there. While dragging, the tree makes the drop target obvious, and moving a folder takes its contents along. Uploads by drag-in keep working as they do today.

### Out of scope

- Dragging items out of the app or between sandboxes.
- Multi-select drag.

kapetr:
Files and folders in the chat's Files tree can now be reorganized by dragging: drop onto a folder to move an item into it, onto a file to move it beside that file, or onto empty panel space to move it to the workspace root. Folders bring their contents (and their expanded state) along, and a file open in the viewer follows its moved parent. Name collisions prompt the same overwrite confirmation as rename — and when a folder genuinely can't be overwritten, the error now says so plainly instead of failing opaquely. Dragging files in from your computer still uploads them as before.

Also rides along: the inline new-file/new-folder/rename row now matches the tree's font and icon alignment.

Closes #2795.

### DIFF
--- a/packages/ui/src/modules/files/hooks/use-file-mutations.ts
+++ b/packages/ui/src/modules/files/hooks/use-file-mutations.ts
@@ -27,6 +27,11 @@ interface RenameEntryInput {
   nextName: string;
 }
 
+interface MoveEntryInput {
+  from: string;
+  toDir: string;
+}
+
 function joinPath(dir: string, name: string): string {
   return dir ? `${dir}/${name}` : name;
 }
@@ -34,6 +39,10 @@ function joinPath(dir: string, name: string): string {
 function parentOf(path: string): string {
   const i = path.lastIndexOf("/");
   return i >= 0 ? path.slice(0, i) : "";
+}
+
+function nameOf(path: string): string {
+  return path.split("/").pop() ?? path;
 }
 
 function isConflictError(err: unknown): boolean {
@@ -96,29 +105,27 @@ export function useFileMutations(agentId: string | null) {
     [createFile, createFolder],
   );
 
-  const renameEntry = useCallback(
-    async ({ from, nextName }: RenameEntryInput) => {
-      const to = joinPath(parentOf(from), nextName);
+  const relocate = useCallback(
+    async (from: string, to: string, failLabel: string) => {
       if (to === from) return;
 
       const tryRename = async (overwrite: boolean) => {
         await renameMutation.mutateAsync({ from, to, overwrite });
         if (agentId) renameExpandedDir(agentId, from, to);
         if (openFilePath === from) setOpenFilePath(to);
+        else if (openFilePath?.startsWith(from + "/"))
+          setOpenFilePath(to + openFilePath.slice(from.length));
       };
 
       try {
         await tryRename(false);
       } catch (err) {
         if (!isConflictError(err)) {
-          emitToast({
-            kind: "error",
-            message: errorMessage(err, "Rename failed"),
-          });
+          emitToast({ kind: "error", message: errorMessage(err, failLabel) });
           return;
         }
         const ok = await showConfirm(
-          `"${nextName}" already exists. Overwrite?`,
+          `"${nameOf(to)}" already exists. Overwrite?`,
           "Overwrite",
         );
         if (!ok) return;
@@ -127,7 +134,7 @@ export function useFileMutations(agentId: string | null) {
         } catch (err2) {
           emitToast({
             kind: "error",
-            message: errorMessage(err2, "Rename failed"),
+            message: errorMessage(err2, failLabel),
           });
         }
       }
@@ -140,6 +147,22 @@ export function useFileMutations(agentId: string | null) {
       setOpenFilePath,
       showConfirm,
     ],
+  );
+
+  const renameEntry = useCallback(
+    ({ from, nextName }: RenameEntryInput) =>
+      relocate(from, joinPath(parentOf(from), nextName), "Rename failed"),
+    [relocate],
+  );
+
+  const moveEntry = useCallback(
+    async ({ from, toDir }: MoveEntryInput) => {
+      if (parentOf(from) === toDir) return;
+      // A folder can't move into itself or its own subtree.
+      if (from === toDir || toDir.startsWith(from + "/")) return;
+      await relocate(from, joinPath(toDir, nameOf(from)), "Move failed");
+    },
+    [relocate],
   );
 
   const deleteEntry = useCallback(
@@ -265,6 +288,7 @@ export function useFileMutations(agentId: string | null) {
   return {
     createEntry,
     renameEntry,
+    moveEntry,
     deleteEntry,
     uploadFiles,
     uploadBundle,
--- a/packages/ui/src/modules/files/hooks/use-file-mutations.ts
+++ b/packages/ui/src/modules/files/hooks/use-file-mutations.ts
@@ -133,7 +133,7 @@ export function useFileMutations(agentId: string | null) {
         if (sourceKind === "dir") {
           emitToast({
             kind: "error",
-            message: `"${nameOf(to)}" already exists`,
+            message: `A folder named "${nameOf(to)}" already exists there. Folders can't be overwritten — rename or delete the existing one first.`,
           });
           return;
         }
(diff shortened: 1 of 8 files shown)


packages/ui/src/modules/files/hooks/use-file-mutations.ts
```diff
@@ -96,29 +105,27 @@ export function useFileMutations(agentId: string | null) {
...
-          emitToast({
-            kind: "error",
-            message: errorMessage(err, "Rename failed"),
-          });
+          emitToast({ kind: "error", message: errorMessage(err, failLabel) });
           return;
         }
         const ok = await showConfirm(
-          `"${nextName}" already exists. Overwrite?`,
+          `"${nameOf(to)}" already exists. Overwrite?`,
```
dam-code-guardian, line 141: 🟡 **Warning:** When moving a **folder** onto a colliding non-empty folder, this prompt offers "Overwrite" — but confirming retries `renameMutation` with `overwrite:true`, and `rename(2)` on the backend still throws `ENOTEMPTY` (surfaced as the new "can't overwrite … not empty" domain error). So the user is offered an overwrite that can never succeed for a non-empty destination directory, then hit with a second error toast.

Consider detecting the dir-onto-non-empty-dir case before prompting and showing the "can't overwrite" message directly, so users aren't offered an action that always fails. (POSIX `rename` can't merge directories, so a true merge would need a different backend strategy.)

packages/ui/src/modules/files/hooks/use-file-row-drag.ts
```diff
@@ -1,43 +1,93 @@
 import type { DragEvent as ReactDragEvent } from "react";
 import { useMemo } from "react";
 
+/** DataTransfer type carrying an in-tree move (JSON DragSource). */
```
dam-code-guardian: 🟢 **Suggestion:** A short note that this MIME is intentionally app-private would help future readers — `handlePanelDragOver` and the row drop path rely on `hasFiles`/`hasMove` being mutually exclusive, which only holds because external drags never carry this type.

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `8224878`

## PR #2800: feat(ui): move files and folders by drag and drop in the Files tree
**Author:** kapetr | **Branch:** feat/2795-file-drag-move → main | **Changes:** +159 −28 (6 files)

### Summary
Adds drag-and-drop reorganization to the chat Files tree: drop onto a folder to move an item inside, onto a file to move it beside that file, or onto empty panel space to move it to the workspace root. Folders carry their contents and expanded state, and an open file follows its moved parent. Rename and move are unified behind a shared `relocate()` helper; the backend now surfaces a clear domain error when a non-empty destination folder can't be overwritten instead of failing opaquely. Closes #2795.

### Findings
- ✅ **Looks good:** The `relocate(from, to, failLabel)` refactor cleanly unifies rename and move, and correctly extends open-file tracking to descendants — `openFilePath?.startsWith(from + "/")` re-points a file open in the viewer when its ancestor folder moves (`use-file-mutations.ts:114-117`).
- ✅ **Looks good:** `moveEntry` guards all three degenerate cases up front — same-parent no-op, move-into-self, and move-into-own-subtree (`toDir.startsWith(from + "/")`) (`use-file-mutations.ts:158-165`).
- ✅ **Looks good:** `readMoveSource` validates the parsed JSON's shape and swallows foreign payloads under the same MIME in a `try/catch`, so a drag from another app carrying `application/x-platform-move` can't crash the drop handler (`use-file-row-drag.ts:34-52`).
- ✅ **Looks good:** Backend maps `ENOTEMPTY`/`EEXIST`/`EPERM` from `rename(2)` to a domain `forbidden` error rather than letting it bubble as an internal 500 (`files.ts:258-270`).
- ✅ **Looks good:** `handleRowDragEnd` clears the stuck drop highlight on drag-cancel (Escape / drop outside window), which don't reliably fire a final `dragleave` (`files-panel-controller.ts:177-181`).
- 🟡 **Warning:** Moving a folder onto a colliding **non-empty** folder produces a confusing two-step: the conflict prompt asks "already exists — Overwrite?", but confirming retries with `overwrite:true` and `rename(2)` still throws `ENOTEMPTY`, so the user then gets "can't overwrite … not empty". The offer to overwrite can't be honored for non-empty dirs. Consider detecting the dir-vs-non-empty-dir case before prompting (`use-file-mutations.ts:127-133`, `files.ts:262`).
- 🟢 **Suggestion:** `MOVE_MIME = "application/x-platform-move"` is a good custom type; a short comment noting it's intentionally app-private (so `hasFiles`/`hasMove` stay mutually exclusive on the drop path) would help future readers (`use-file-row-drag.ts:4`).

### Verdict
COMMENT — Well-engineered, defensively-guarded feature with thoughtful UX handling; the one warning (overwrite prompt for non-empty folders that can never succeed) is a minor UX rough edge, not a correctness bug, so this is a non-blocking COMMENT.

---

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `e4808a7`

## PR #2800: feat(ui): move files and folders by drag and drop in the Files tree
**Author:** kapetr | **Branch:** feat/2795-file-drag-move → main | **Changes:** +212 −28 (8 files)

### Summary
Adds drag-and-drop reorganization to the chat Files tree (drop onto folder / beside file / empty space → root), with folders carrying contents + expanded state and open files following a moved parent. Rename and move are unified behind `relocate()`; the backend maps `rename(2)` `ENOTEMPTY`/`EEXIST`/`EPERM` to a domain `forbidden` error. **Since the last review, both prior findings were addressed** and a follow-up commit unifies error reporting.

_(No PR-local overrides. No findings suppressed by PR context — the only prior PR artefacts are DAM's own review + inline comments.)_

### Changes since last review
Previous HEAD: `8224878` (2026-07-15T07:39:17Z) — verdict COMMENT

- ✅ **Fixed:** The 🟡 warning — a folder dropped onto a colliding **non-empty** folder no longer offers an "Overwrite?" prompt that `rename(2)` can't honor. `relocate()` now takes the source `kind`, and for `kind === "dir"` a conflict shows a direct, explanatory toast ("A folder named … already exists there. Folders can't be overwritten…") instead of the doomed confirm (`use-file-mutations.ts:130-139`).
- ✅ **Fixed:** The 🟢 suggestion — `MOVE_MIME` now carries a comment documenting that it's app-private on purpose, which is what keeps the `hasFiles`/`hasMove` drop branches mutually exclusive (`use-file-row-drag.ts:4-6`).
- 🆕 **New:** `e4808a7` "single-voice error reporting" — the five file mutations (write/create/mkdir/rename/delete/upload) now set `meta.suppressErrorToast` so the global `onError` toast no longer doubles with each caller's own error handling. Verified every caller (`use-file-mutations.ts`, `file-viewer.tsx`) wraps `mutateAsync` in try/catch with an `emitToast` on error — so opting out of the global toast does **not** introduce silent failures. Clean.

### Findings
- ✅ **Looks good:** The overwrite-conflict fix is correct — passing `sourceKind` through `moveEntry`→`relocate` and short-circuiting the dir case before the confirm dialog eliminates the impossible-overwrite two-step. The message is clear and actionable.
- ✅ **Looks good:** The `suppressErrorToast` refactor is sound end-to-end: the new query-client meta flag is honored in `onError`, and all six suppressed mutations have callers that own error reporting via try/catch + `emitToast`. No silent-failure regression.
- ✅ **Looks good:** Backend `rename(2)` error mapping (`ENOTEMPTY`/`EEXIST`/`EPERM` → domain `forbidden`) unchanged and still correct, with a non-matching error rethrown rather than swallowed (`files.ts:258-270`).

✅ No documentation drift detected. This PR is UI feature work plus a `rename(2)` error-mapping refinement in the agent-runtime files service; no architecture page under `docs/architecture/` owns file-tree UX or the files-service error taxonomy, and no subsystem responsibility changed. No docs were edited (correctly — nothing to update).

### Verdict
APPROVE — Both findings from the prior COMMENT review are cleanly resolved (impossible folder-overwrite prompt removed; MIME documented), and the follow-up single-voice error-reporting refactor is correct and verified to not swallow failures. No outstanding issues.

---

PetrBulanek (approved):

PR #2801, opened 2026-07-15
fix(dam-run): close error-handling and timeout gaps across the run stack

JanPokorny:
## What

Review of the `dam-run` feature's error handling found several gaps in the crash/timeout story; this PR fixes all of them.

- **controller**: reap over-age Runs before the terminal-phase short-circuit (mirrors the Fork reaper) — previously a `Failed` Run CR the api-server never deleted leaked its executor pod until the next api-server restart. Executor-pod list errors now requeue instead of being swallowed.
- **relay**: attach `close`/`message` handlers before the first `await` — a client disconnecting during agent resolution used to leave the executor running untethered (until the 60-min reaper) with a concurrency slot pinned; a close racing `create()` now re-deletes the CR.
- **relay**: WebSocket ping/pong keepalive (30s) on both legs — keeps quiet long-running commands alive through Envoy's default 5-min stream-idle timer (previously survival depended on the unrelated `AnyUpgrades` egress config) and detects half-open peers (node died without FIN) within ~60s instead of hours. Executor dial bounded by a 10s handshake timeout; executor loss now surfaces to dam-run as a close reason instead of a silent exit 1.
- **runs-service**: `waitReady` polls to 155s (past the controller's 120s pod-ready timeout + 30s resync worst case) so the controller's specific error message surfaces; Run delete failures are logged.
- **dam-run**: piped-stdin EOF now propagates as EOT so `echo hi | dam-run cat` exits instead of hanging until the reaper; 30s relay dial timeout; stdout flushed before exit so fast output isn't truncated.
- **harness app**: boot-sweep failures logged; RollingUpdate overlap documented.

## Tests

- `TestRunReconcile_ReapsOverAgeFailedRun` — Failed-phase reaper regression.
- New `harness-run-relay.test.ts` — drives the **real `dam-run.mjs`** as a child process through the real relay to a fake exec server (framing, EOT-on-EOF, exit code, CR deletion), plus the early-disconnect release regression.
- `exec.test.ts`: EOT-as-EOF against a real PTY.
- All package suites green (controller Go, agent-runtime 165, api-server 295).

Note: `controller:check:crd-backwards-compatibility` fails on this baseline with or without these changes (pre-existing `userbudgets.yaml` removal vs v0.2.7-rc1).

### DIFF
--- a/packages/agent-runtime/src/__tests__/unit/exec.test.ts
+++ b/packages/agent-runtime/src/__tests__/unit/exec.test.ts
@@ -70,6 +70,16 @@ describe("attachExec (PTY)", () => {
     expect(output(ws)).toContain("got-hi");
   });
 
+  // dam-run signals piped-stdin EOF as EOT×2 (a PTY has no true EOF); a
+  // command reading to EOF must exit instead of hanging until the reaper.
+  it("treats EOT input as EOF", async () => {
+    const ws = run(["cat"]);
+    ws.emit("message", Buffer.from(encodeDataFrame(OP_INPUT, "hi\n")));
+    ws.emit("message", Buffer.from(encodeDataFrame(OP_INPUT, "\x04\x04")));
+    expect(await ws.exited).toBe(0);
+    expect(output(ws)).toContain("hi");
+  });
+
   it("reports a missing command as a non-zero exit", async () => {
     const ws = run(["definitely-not-a-real-cmd-xyz-42"]);
     expect(await ws.exited).toBeGreaterThan(0);
--- a/packages/api-server/src/__tests__/unit/harness-run-relay.test.ts
+++ b/packages/api-server/src/__tests__/unit/harness-run-relay.test.ts
@@ -1,0 +1,140 @@
+import { createServer, type Server } from "node:http";
+import type { AddressInfo } from "node:net";
+import { spawn } from "node:child_process";
+import { once } from "node:events";
+import { fileURLToPath } from "node:url";
+import { WebSocket, WebSocketServer } from "ws";
+import { afterAll, describe, expect, it, vi } from "vitest";
+import { createRunRelay } from "../../apps/harness-api-server/harness-run-relay.js";
+import type { RunsService } from "../../modules/runs/services/runs-service.js";
+import type {
+  K8sClient,
+  KubeObject,
+} from "../../modules/agents/infrastructure/k8s.js";
+import { LABEL_OWNER } from "../../modules/agents/infrastructure/labels.js";
+
+// Terminal-protocol opcodes (see api-server-api terminal/protocol.ts).
+const OP_INPUT = 0x00;
+const OP_OUTPUT = 0x01;
+const OP_EXIT = 0x03;
+
+const DAM_RUN = fileURLToPath(
+  new URL("../../../../platform-base/dam-run.mjs", import.meta.url),
+);
+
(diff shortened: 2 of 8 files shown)


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `1e1d585`

## PR #2801: fix(dam-run): close error-handling and timeout gaps across the run stack
**Author:** JanPokorny | **Branch:** fix/dam-run-error-handling#locki-efc7uksc → main | **Changes:** +279 −29 (8 files)

### Summary
Closes several crash/timeout gaps in the `dam-run` executor stack: the controller now reaps over-age Runs before the terminal-phase short-circuit (so a leaked `Failed` CR's pod is cleaned up), the relay attaches close/message handlers before the first `await` and adds 30s WebSocket ping/pong keepalive on both legs plus a bounded executor handshake, `runs-service` widens `waitReady` to 155s and logs delete failures, and `dam-run.mjs` propagates piped-stdin EOF as EOT, adds a dial timeout, and flushes stdout before exit. Well-scoped, thoroughly commented, and backed by new tests (Go reaper regression + a relay test driving the real CLI).

### Findings
- ✅ **Looks good:** The close-racing-`create()` window is handled correctly — `release()` is idempotent via the `released` guard, and after `create()` a `if (released) return void deps.runs.delete(runId)` re-deletes a CR whose write landed after `release()`'s delete (`harness-run-relay.ts:138-141`).
- ✅ **Looks good:** `us.on("close")` and `us.on("error")` are now registered unconditionally (outside the `us.on("open")` callback), so a handshake that fails before `open` still surfaces to `dam-run` as a close reason instead of a silent hang (`harness-run-relay.ts:168-182`).
- ✅ **Looks good:** Moving the over-age reaper ahead of the `Failed`/`Completed` short-circuit mirrors the documented Fork reaper and is covered by the new `TestRunReconcile_ReapsOverAgeFailedRun` (`run.go:62-77`, `run_test.go:158-168`).
- ✅ **Looks good:** `findEphemeralPod`'s error is no longer swallowed — a list error now requeues via a returned error instead of silently treating the pod as absent (`run.go:121-124`).
- ✅ **Looks good:** The EOT-on-EOF handling (`OP_INPUT, 0x04, 0x04`) is well-reasoned and matched by both the PTY-level `exec.test.ts` case and the end-to-end relay test that drives the real `dam-run.mjs` as a child process (`dam-run.mjs:89-94`, `harness-run-relay.test.ts`).
- 🟢 **Suggestion:** In `finish()`, the 2s backstop `setTimeout` is not `.unref()`'d; harmless here since the flush callback calls `process.exit` first in the normal path, but unref'ing would avoid the timer keeping the loop alive in an edge case where exit is otherwise deferred (`dam-run.mjs:63-64`).

✅ No documentation drift detected. No `docs/architecture/` pages are edited, and the touched subsystem behaviors (Run executor relay, controller over-age reaper, `dam-run` CLI) are already described in `agent-lifecycle.md` §"Run executors (`dam-run`)" — the reaper change brings the code in line with the page's documented "two backstops cover a Run the api-server never cleaned up" intent and the Fork reaper's "regardless of phase" behavior. Keepalive/handshake/EOT are implementation-level details the guidelines keep out of architecture pages.

### Verdict
APPROVE — Focused, well-tested robustness fixes with accurate comments; no correctness or security concerns, one optional nit.

---

PR #2805, opened 2026-07-15
fix(ui): guard against missing sandbox size in settings form

PetrBulanek:
## Problem

Opening any sandbox created before the compute-budget/size feature (#2768) shows a **blank page**. The sandbox home view crashes with `Cannot read properties of undefined (reading 'cpu')` because such sandboxes have no `size`, and the settings form reads `agent.size.cpu` directly. This blanks the whole home view for every section (Setup, Connections, Skills, Schedules), not just one.

## Fix

Guard the access with optional chaining (`agent.size?.cpu` / `agent.size?.memory`). The existing `?? 1000 / 1024` fallbacks already handle the absent values, so pre-size sandboxes now fall back to the defaults instead of crashing.

### DIFF
--- a/packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-form.ts
+++ b/packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-form.ts
@@ -147,8 +147,8 @@ export function useSandboxSettingsForm() {
         .sort(),
       envVars: userInitialEnv,
       hibernationTimeoutMin: agent.hibernationTimeoutMin,
-      sizeCpuMilli: parseCpuMilli(agent.size.cpu) ?? 1000,
-      sizeMemoryMi: parseMemoryMi(agent.size.memory) ?? 1024,
+      sizeCpuMilli: parseCpuMilli(agent.size?.cpu) ?? 1000,
+      sizeMemoryMi: parseMemoryMi(agent.size?.memory) ?? 1024,
     });
     setFormReady(true);
   }, [


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `ba8eb31`

## PR #2805: fix(ui): guard against missing sandbox size in settings form
**Author:** PetrBulanek | **Branch:** fix/sandbox-size-null-safety → main | **Changes:** +2 −2 (1 file)

### Summary
Guards `agent.size?.cpu` / `agent.size?.memory` with optional chaining in the sandbox settings form hook so sandboxes created before the compute-budget/size feature (#2768) — which have no `size` — fall back to defaults instead of crashing the entire home view with `Cannot read properties of undefined (reading 'cpu')`.

### Findings
- ✅ **Looks good:** The fix matches the type: `size` is declared optional (`size?: { cpu?: string; memory?: string }` in `mutations.ts:55`), so `agent.size.cpu` was a latent unsafe access; `?.` aligns the runtime access with the type. `parseCpuMilli`/`parseMemoryMi` both accept `string | undefined` (`sandboxes/lib/quantity.ts:9,29`) and return `null` for absent input, so the existing `?? 1000` / `?? 1024` fallbacks cleanly cover the pre-size case.
- ✅ **Looks good:** Complete — this is the only unguarded `.size.cpu`/`.size.memory` access in `packages/ui/src`; no sibling call sites carry the same crash.

✅ No documentation drift detected. A one-line UI null-safety guard changes no subsystem behavior or responsibility and touches no `docs/architecture/` page — nothing to update.

✅ Aligned. The change is confined to a client-side React hook (`packages/ui/`), outside the server three-layer/ports-and-adapters rules; it removes a latent type-unsafety (dereferencing a possibly-`undefined` optional field) rather than introducing one, with no new `any`, no loosened types, and no dependency-direction or module-boundary change.

### Verdict
APPROVE — Minimal, correct, type-aligned crash fix for pre-`#2768` sandboxes; fallbacks already handle the absent values and no other call site has the same gap.

---

PetrBulanek:
Closing as redundant. Root cause was a version skew — the deployed api-server was a stale build that omitted the `size` field from the agent view. `agents.list`/`get` always populate `size` via `toView` (no optional output), so with a correctly-built api-server `agent.size` is never undefined and this guard never fires. The real fix is rebuilding the api-server; `size` being optional only applies to the create/update *input* schemas, not the view.

PR #2811, opened 2026-07-16
feat(ui): rework the Connections experience around a catalogue modal

kapetr:
One Connection catalogue modal replaces the three divergent inline connection catalogs (settings page, sandbox home section, create-wizard step).

- **Connection catalogue modal**: category tabs (Apps, MCP servers, Custom Headers) with connection counts, provider group cards (GitHub folds its OAuth app and personal-access-token flavors into one card), and in-modal create flows — an auth-method chooser for multi-template providers, embedded forms with per-provider guidance (GitHub authorization explainer, fine-grained PAT hints), and popup OAuth that never navigates away.
- **Sandbox home → Connections** lists the sandbox's granted connections grouped by provider. Grants apply immediately — no Submit changes step; staging remains only in the create wizard, where the sandbox doesn't exist yet. Creating from the catalogue auto-grants and closes the modal.
- **Settings → Connections** shows the same grouped "My connections" with `+ New` opening the catalogue in no-sandbox mode.
- **Wizard step 3** shares the sandbox section's panel; grants stage into the draft and survive reloads.
- Deleting a connection is offered only in the catalogue and settings; the confirm dialog names the connection, lists the affected sandboxes, and uses the destructive style. Failed or abandoned OAuth flows no longer leave half-created "Authorizing…" connections behind, and OAuth results are verified against the server so the outcome is reliable even when the popup's message is lost.
- e2e specs realigned to the new surfaces.

Part of #2759.

Closes #2802

https://claude.ai/code/session_01EQ7eHrr6USd3T1dddzRCFw

PR #2807, opened 2026-07-16
feat(ui): per-user monthly LLM spend in settings Usage tab

JanPokorny:
## Summary

Adds a **Usage** section to user settings showing per-user LLM API spend, aggregated across all of the user's agents:

- total spend for the selected calendar month
- per-model breakdown (calls folded into in/out tokens + cost, reusing the sessions metrics panel table)
- month name with ‹ / › navigation in the top-right corner (› disabled at the current month; boundaries computed in the browser's timezone)

## Implementation

- **`metrics.spend`** tRPC procedure: per-model token spend over an absolute half-open `[from, to)` range, ownership-scoped like `metrics.overview`. The ClickHouse window predicate gains `fromIso`/`toIso` filters. Fails closed when ClickStack is disabled.
- **Deleted agents stay attributable**: the metrics ownership scope now unions live Agent CRs with the Postgres agent registry (soft-deleted included, matched via the HMAC'd owner sub), so monthly totals don't shrink retroactively when an agent is deleted. Note: a re-used agent id attributes its whole history to the current owner, matching the existing agent-id-keyed attribution model.
- New `usage` settings tab wired through `settingsTabSchema` (deep-linkable at `/settings/usage`).

## Testing

- Unit tests for the spend ownership gate (incl. empty-owner and disabled-service paths) and the ClickHouse range predicate; `mise run check` green, api-server 343/343, UI 77/77.
- Verified end-to-end on a local k3s cluster with ClickStack enabled and seeded `otel_logs` rows: month totals and breakdowns match ClickHouse exactly, month navigation works (populated months, empty-month state, disabled next at current month), OIDC deep-link round-trip works, and spend survives deleting the owning agent (rendered $0 before the registry-union fix).

PR #2815, opened 2026-07-16
test(agent-runtime): de-flake "treats EOT input as EOF" PTY test

pilartomas:
## What

De-flakes `packages/agent-runtime/src/__tests__/unit/exec.test.ts > attachExec (PTY) > treats EOT input as EOF`, which has been intermittently reddening `main` and PRs with:

```
AssertionError: expected 129 to be +0 // Object.is equality
```

## Root cause

The test asserted `cat` exits **exactly 0** after EOT-as-EOF. But a PTY reports EOF-driven teardown **nondeterministically**: `cat`'s clean exit (`0`) races the controlling-terminal hangup (`SIGHUP` → `128 + 1 = 129`). Either can win depending on kernel/node-pty reap ordering, and load makes SIGHUP win more often — which is why it surfaces on busy CI runners.

This is **not** an `attachExec` bug: in the test nothing closes the socket, so `pty.kill()` is never invoked. `attachExec` faithfully reports whatever the PTY layer returns (`exitCode || (signal ? 128 + signal : 0)`). So the correct fix is the test assertion, not the code.

I verified no timing tweak removes the race — the residual flake persists through all of them, confirming it lives in the reap ordering, not in *when* EOT is sent:

| variant | exit-129 rate |
|---|---|
| two EOT, sync writes (current) | ~20–23% |
| single EOT | ~22% |
| combined `"hi\n\x04"` write | ~3% |
| wait for `hi` echo, then EOT | ~2.5% |
| EOT delayed 5ms | ~0.8% |

## Fix

Assert the **deterministic contract** the test actually cares about:
- EOT causes the command to **terminate** — if EOT were *not* treated as EOF it would be echoed as data and `cat` would run forever, hanging the test (this is the real regression the test guards against);
- the relayed input (`hi`) is echoed back;
- the exit code is `0` **or** `129` (clean EOF exit vs. SIGHUP hangup) — both mean "terminated via EOF-driven teardown".

## Evidence

- **Repro (standalone `@lydell/node-pty@1.1.0`, exact write sequence):** 70/300 (~23%) exited `129` (`exitCode=0, signal=1`), `hi` echoed in every case.
- **Repro (real test under CPU load):** 18/40 (45%) failed with `expected 129 to be +0`.
- **After fix (real test, same load):** 40/40 green. Full `exec.test.ts` suite: 6/6 pass.

Observability context that motivated this: reported flaking on main push #2807 (run `29488131019`) and PR #2811 (run `29496011441`) — PR #2811 passed at 11:22 then failed at 11:52 with no relevant change.

### DIFF
--- a/packages/agent-runtime/src/__tests__/unit/exec.test.ts
+++ b/packages/agent-runtime/src/__tests__/unit/exec.test.ts
@@ -76,7 +76,16 @@ describe("attachExec (PTY)", () => {
     const ws = run(["cat"]);
     ws.emit("message", Buffer.from(encodeDataFrame(OP_INPUT, "hi\n")));
     ws.emit("message", Buffer.from(encodeDataFrame(OP_INPUT, "\x04\x04")));
-    expect(await ws.exited).toBe(0);
+    // ^D (EOT) on an empty line makes the tty line discipline signal EOF, so
+    // `cat` stops reading and terminates. If EOT were NOT treated as EOF it
+    // would be echoed as data and `cat` would run forever — this await would
+    // then hang until the test times out, which is the real regression this
+    // guards against. How the kernel reports the EOF-driven teardown is racy:
+    // `cat`'s clean exit (0) races the controlling-terminal hangup (SIGHUP ->
+    // 128+1 = 129), and either can win under load — so we assert on
+    // termination + relayed output, not on an exact exit code.
+    const exitCode = await ws.exited;
+    expect([0, 129]).toContain(exitCode);
     expect(output(ws)).toContain("hi");
   });
 


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `bce6dcf`

### Summary
De-flakes the `attachExec (PTY) > treats EOT input as EOF` test by asserting the deterministic contract (command terminates via EOF, `hi` is echoed) and accepting either exit code `0` (clean exit) or `129` (SIGHUP hangup), which race nondeterministically under load. Test-only change; no production code touched.

### Findings
- 🟢 **Suggestion:** The assertion `expect([0, 129]).toContain(exitCode)` is correct and readable. If you wanted a stronger failure message on regression, `expect([0, 129], \`unexpected exit ${exitCode}\`).toContain(exitCode)` surfaces the offending code directly — purely optional. (`packages/agent-runtime/src/__tests__/unit/exec.test.ts:88`)
- ✅ **Looks good:** The root-cause analysis is thorough and the fix targets the test contract, not the code under test — `attachExec` faithfully relaying the PTY layer's exit is left untouched.
- ✅ **Looks good:** The explanatory comment documents *why* the exit code is nondeterministic and what regression the test still guards against (EOT not treated as EOF → hang), preserving the test's intent.
- ✅ **Looks good:** The termination-based assertion still fails loudly (timeout) if EOT ever stops being treated as EOF — the real regression is preserved, not weakened.

✅ No documentation drift detected. Test-only change; no `docs/architecture/` pages affected.

✅ No findings. The change is a localized test-assertion tweak introducing no module-boundary, slice-composition, or typing violations.

### Verdict
APPROVE — a well-reasoned, evidence-backed de-flake that tightens the test onto its true contract without weakening the regression it guards.

---

Issue #2784, opened 2026-07-14
Loop: platform-driven DBTL loop (depth first, built for evolution)


tomkis:
## Problem Statement

The platform can launch AI R&D work but it can't drive the loop. Experiments races several agents against a goal and collects whatever each self-reports, but the iterate-improve loop lives inside a single agent session. That works only while one agent does every step. The moment a step has a different owner than the step before it (an evaluation by someone other than the maker, or an external gate like a review or CI), the single-session model can't hold the hand-off.

Two problems follow. First, there is no way to carry a candidate through a make step, an automated gate, and an independent evaluation, then re-attempt on rejection, with a durable record of the rounds and what was learned between them. Second, a single session can only pursue one line of attack at a time. Many goals are better served by making several candidates, comparing them, and carrying the best forward, which a single session cannot express at all.

We exercise this first on a software task (cleanest objective signals), but the loop is not specific to software.

## Solution

A **Loop**: the platform drives a Design-Build-Test-Learn loop over a goal, from a vague objective to an evaluated, approved result.

The loop unit is a **generation** that produces a **population** of candidates. Over each generation the platform runs a fixed core: **Make** (fresh agents produce candidates), **Test** (an automated gate, objective pass/fail per candidate), **Eval** (a judgment step returning a verdict and an optional score per candidate), **Select** (choose the survivors), **Curate** (findings distilled into a running knowledge base). Survivors plus the curated findings seed the next generation.

Crucially, **depth is just breadth with a population of one**. A generation of a single candidate, selected trivially, is the plain iterate-improve loop. This lets us ship the depth loop first while building every part on machinery that already understands populations, so breadth is later a configuration change, not a rewrite. See Milestones.

What each step does is carried entirely in that step's prompt, so the same loop runs a software task or any other work without the platform knowing the difference. A candidate is opaque to the platform: a downloadable artifact and/or a reference into an external system.

Eval returns one of three control signals: **continue** (the candidate did not pass, the loop iterates), **passed** (the terminal success, the loop stops), or **needs decision** (an optional HITL gate that pauses the loop for a human to **continue or abort**). There is no "fail" verdict. Not passing simply continues, and the only hard stops besides passed and the generation cap are a human abort or the HITL deadline.

[...]

Issue #892, opened 2026-06-12
Redesign — new navigation shell + foundation decisions


PetrBulanek:
## Problem

The product currently has two parallel ways to create and manage an AI agent: the original "Agents" experience (a labelled sidebar, a list page, and a modal for creating/configuring an agent) and a newer, full-page "sandbox" experience that lives behind its own URL and has no sidebar. They share the same backend but present as two different products with two different vocabularies.

The redesign collapses this into a single experience built around one full-page create/edit surface and a stripped-back navigation. Before any individual screen is built, three product decisions have to be settled, because every downstream screen inherits them. This issue exists to (a) lock those decisions and (b) build the navigation shell that all the redesigned screens sit inside.

## Proposed solution

**Navigation** becomes a slim icon rail on the left: the product wordmark, **Home** (the list of sandboxes), **Inbox** (approvals), and **Settings**. The current top-level "Providers" and "Connections" destinations move *inside* Settings, which becomes a small settings area with its own sub-navigation: **Account**, **Providers**, **Connections**.

Old entry points (the labelled-sidebar list, the standalone providers/connections pages, the standalone per-agent network page) redirect into their new homes so existing links don't break.

### Decisions to confirm

Each comes with a recommendation — please push back where you disagree, since the downstream issues are written assuming these answers.

- **What do we call the thing a user creates?** Recommendation: **"Sandbox"** as the single user-facing noun everywhere (list, button, wizard, settings), retiring "Agent" as the primary term. The designs lean this way and the newer surface already uses it. *(Note: the designs themselves are inconsistent — some headings still say "agent" — so this needs an explicit call.)*
- **What shape is creation?** Recommendation: the **three-step flow** (Image → Setup → Connections) that is fully drawn in the designs, rather than the alternate, more granular flow also present in the file ("Choose your approach → Choose a harness → Configure → Grant connections"). The three-step version maps cleanly onto how creation already works under the hood.
- **Do we keep the old experience too, or retire it?** Recommendation: **retire** the original list + creation/configuration modals once the new surface reaches parity, redirecting old links. Running both indefinitely doubles the maintenance surface and re-splits the vocabulary we're trying to unify.

### Scope

- The new left icon rail (wordmark, Home, Inbox, Settings) replacing the labelled sidebar and mobile nav.
- The Settings area with Account / Providers / Connections sub-navigation.
- Moving the existing Providers and Connections pages under Settings (content largely unchanged here — their redesign is covered by the wizard/settings issues).
[...]

PR #2816, opened 2026-07-17
feat(invocations): spawn primitive + driver SDK over ephemeral agents (#2784)

tomkis:
📄 **Review walkthrough:** https://claude.ai/code/artifact/f234a880-563a-4500-8b68-5ebde5a334a5 (private, ask for access)

This PR is a tracer: it ships one platform-level primitive and nothing built on top of it. The primitive is "spawn an ephemeral agent with a chosen image plus a subset of my connections plus a prompt, run it once, give me back a schema-validated result." Loops, fan-out, and any specific orchestration shape are out of scope here. Per ADR-078's thinking they are driver programs that call this primitive, and they ship separately once the primitive lands.

Draft while it settles.

> **Term note:** the branch started out calling the spawned unit a "Sandbox" and shipped it as one `sandboxes` module. After a design pass the word is retired (it already means the user-facing name for an Agent, #892) and the concept is split into two orthogonal concerns: **Agent Sweep** (lifecycle, on the Agent) and **Invocation** (the result contract). See ADR-077.

## The primitive: a spawned node IS an ephemeral Agent

A spawned node is a normal Agent that a "driver" agent creates, hands one prompt, and polls. We reuse the Agent primitive rather than a new CRD, so it resolves natively in the ext-authz identity resolver, owns its own `/mcp` endpoint (so `report_result` attributes by the agent's own id), and gets the real connection fan-out (per-target attenuation, not injection-only).

Flow: driver -> `POST /invocations` -> create a fresh `Sweepable` Agent with the connection subset, stash the prompt + JSON Schema in the `invocations` row -> deliver the prompt via the existing `trigger` rail (session mode `fresh`) -> target runs the turn and calls `report_result(result)` on its own MCP -> server ajv-validates the result against the stashed schema (structural only, never truth) and stores it -> driver polls `GET /invocations/:id` -> the target Agent is reaped and, after a retention window, the row.

## Two concerns, unbundled

- **Agent Sweep (lifecycle, lives on the Agent).** A `Sweepable` annotation marks an ephemeral agent for automatic deletion; a generic Agent Sweep deletes it once it hibernates, after an optional `Agent Lifetime` grace (default zero). The Sweep keys off Agent state alone, never the invocations table, so Forks and inherited channel agents can adopt it later. api-server-only: deleting the Agent ConfigMap cascades pod/gateway/PVC via ownerReferences, no controller change.
- **Invocation (result contract, own context).** A `(driver, target, prompt, result schema) -> one validated result` record, keyed by the target agent's id. It owns the stashed schema, the validated result, and the per-result liveness deadline. It does not own autosweep.

**Reaping is split by flush safety.** The done path reaps eagerly inside `recordResult` (the ConfigMap delete cascades async, so the `report_result` response still flushes before the pod dies). The fail path reaps inside the liveness sweep (nothing to flush). The Agent Sweep is the backstop for anything an eager delete missed.

## Driver ergonomics: a typed SDK plus one skill

- `packages/driver-sdk`: a workspace package authored in TypeScript, built with `tsup` into a single dependency-free `.mjs` shipped in the image at `/usr/local/lib/driver-sdk.mjs` (built in platform-base's builder stage, copied like `model-gateway.mjs`). Self-configures from the pod environment. Exposes `spawn()` (create-then-poll hidden), `listImages()`, `listConnections()`, and a small `s()` schema sugar. The `/invocations` request/response contract lives in `api-server-api` and the SDK does a type-only import of it, so a change to the request shape breaks the SDK build. `vitest` unit tests cover the schema sugar.
- `dam-invoke` skill: teaches the single spawn, including "do not guess the image or connections" (a reasoning node needs an LLM image and a model connection, or it cannot start a session and hangs to its deadline).

## Platform hardening (from live testing)

Running real spawns end to end surfaced three failure modes; two are fixed here, one mitigated:

- Wrong image / no model connection: the target could not start a session and hung to the deadline. Fixed at the skill level (require an LLM image plus a model connection) and caught earlier by discovery.
- Node too slow or too small: `POST /invocations` now takes a per-spawn `ttlMs` (bounded ~1min..6h) and a `memory`/`cpu` size override, so a heavy node can request 4Gi instead of OOM-killing at the 1Gi template default, and a quick gate can fail fast with a short ttl.
- Crash mid-turn: a one-shot node whose pod restarts (OOMKill, evict) is orphaned, because the trigger will not re-fire. The liveness sweep now reads pod restart status and fails such a node at once instead of letting it idle to the liveness deadline.

## API surface (harness server, scoped to the driver `:id`)

- `POST /api/agents/:id/invocations {image|templateId, connections, prompt, schema, ttlMs?, memory?, cpu?}` -> `{id}`
- `GET  /api/agents/:id/invocations/:invocationId` -> `{status, result}`
- `GET  /api/agents/:id/connections`, `GET /api/agents/:id/images`
- `report_result(result)` MCP tool: ajv-validates vs the stashed schema, stores + marks done on pass, returns the errors on fail so the agent retries. Attribution is automatic from the reporting agent's id.

## Deferred (out of this tracer)

- **The loop-as-code layer**: the `dam-loop` DBTL skill and ADR-078. A loop is a driver program over this primitive, not part of it, so it ships in its own PR once the primitive is in.
- Ledger / registry / UI / metrics / HITL, breadth (population > 1), a dedicated `sandbox-trigger` event kind, migrating Forks and inherited channel agents onto `Sweepable` (dam-di3), and reconciling `docs/architecture/experiments.md`.

## Testing

Verified end to end on a dev cluster: a driver spawned targets and polled them to validated results, and the OOM/crash and misconfig failure modes were observed and then addressed here. `driver-sdk` has vitest tests over the schema sugar. `tsc`, eslint, and prettier are clean; api-server tests 408/408.


tomkis:
## Design handoff: retire "Sandbox", split into Agent Sweep + Invocation

After discussion with Radek the shape here changes. Capturing the conclusion so we implement it in a fresh context. Glossary on this branch is already updated to match (`docs/ubiquitous-language.md`).

### Why the change

The glossary already binds "Sandbox" to the user-facing name for an Agent (#892). This PR overloaded the same word for the ephemeral spawned node. But a spawned node is just an Agent with a lifetime, so a separate `sandboxes` module + table + sweeper is the wrong shape. Radek's point: don't make a separate concept, mark the agent sweepable and give it a lifetime.

### The split

"Sandbox" bundled three independent things:

1. Fresh ephemeral agent
2. Autosweep (lifecycle)
3. Typed input to output, run once (result contract)

These are orthogonal. An inherited agent (post-Slack-message) is 1+2 without 3. A fork is ephemeral + run-once with free-form output. So we unbundle.

**Concern A: Agent Sweep (lifecycle, lives on the Agent)**
- `Sweepable` flag on the Agent.
- Delete on hibernate: a Sweepable agent that goes idle enough to hibernate gets deleted instead of hibernating.
- Optional `Agent Lifetime` grace, default zero. Only matters later, when inherited agents want to stay warm (Radek's "2 days").
- api-server only (`agents.delete` drops the ConfigMap, ownerReferences clean up the rest). No controller change.
- Litmus test: the Sweep keys off Agent state, never the invocations table. If it reads invocations, it is just the old sweeper renamed and the future Fork migration has nothing to hook into.

**Concern B: Invocation (result contract)**
- Rename `modules/sandboxes` to `modules/invocations`, strip lifecycle out.
- Record is `(driver, target, prompt, result schema) -> one validated result`.
- Liveness deadline stays here (it bounds one result, not the agent).
- Shape the record so `target != driver` is possible, but expose only spawn-fresh for now.

**Reaping**
- Eager `agents.delete` on Invocation terminal (done OR failed, so a liveness-failed Invocation also drops its agent).
- Agent Sweep is the backstop for agents no Invocation reaps.

### Work items

- [ ] Add `Sweepable` + generic Agent Sweep (keyed off Agent state)
- [ ] Rename `modules/sandboxes` to `modules/invocations`, strip lifecycle
- [ ] Delete `sandbox-sweeper.ts`, fold terminal reap into eager delete + Sweep
- [ ] Rename `node_done` to `report_result` (`sandbox-sdk.mjs`, the skill, the prompt builder)
- [ ] Rewrite ADR-077 and 078 in place (unmerged on this branch, no supersede needed; split into more ADRs if it reads cleaner)
- [ ] Follow-up issue: migrate Forks and inherited channel agents onto `Sweepable` (large and separate, a Fork is Job-shaped today)

tomkis:
Implemented on `feat/spawn-sandbox-primitive` (d44757e4).

Work items:

- [x] Add `Sweepable` + generic Agent Sweep (keyed off Agent state)
- [x] Rename `modules/sandboxes` to `modules/invocations`, strip lifecycle
- [x] Delete `sandbox-sweeper.ts`, fold terminal reap into eager delete + Sweep
- [x] Rename `node_done` to `report_result` (SDK skill, the skill, the prompt builder)
- [x] Rewrite ADR-077 and 078 in place
- [x] Follow-up issue filed for Forks + inherited channel agents (dam-di3)

Notes on a few calls I made:

**Sweepable is an annotation, not a spec field.** `agent-platform.ai/sweepable` + `agent-platform.ai/lifetime-ms`, stamped at create. api-server only, no CRD or controller change. The Agent Sweep lists agents and filters `sweepable && hibernated && grace-elapsed`, reading the Ready condition's `lastTransitionTime` for the Lifetime window. It never touches the invocations table, so the Fork migration has something to hook into.

**Reaping is split by flush safety.** Done path reaps eagerly inside `recordResult`; fail path reaps inside the liveness sweep. Deleting the ConfigMap cascades pod teardown downstream via ownerReferences, so the `report_result` response still flushes before the pod dies. Agent Sweep is the backstop for anything an eager delete missed.

**`sandbox-sweeper.ts` is gone.** Its replacement `invocation-liveness.ts` only does what bounds one result: deadline fail, crash fast-fail, and dropping aged result rows. No terminal-reap loop.

**DB:** the unmerged 0014 migration got regenerated fresh as `invocations` rather than adding a create-then-rename, so the feature still ships one clean migration.

**Left as-is on purpose:** the UI `modules/sandboxes` and the "running sandbox" copy (that's the #892 user-facing name for Agent, not the spawn record), plus the `/sandboxes` HTTP path and the SDK's `sandbox` vocabulary, since the handoff scoped the SDK change to `report_result`. Say the word if you want the driver-facing surface renamed too.

Still open (already deferred in the PR body): reconciling `docs/architecture/experiments.md`.

Checks: `api-server` / `api-server-api` / `db` clean on tsc+lint+format, `db:check:generated` + ADR index + ADR immutability pass, api-server tests 408/408.

tomkis:
## Handoff: turn the driver SDK into a real TypeScript package (`driver-sdk`)

Design agreed in a grilling session. This captures the decisions and the why so it can be implemented directly. Scope this as its own change on top of / alongside this PR, not folded into the term-retirement migration.

### Goal (settled)

Dev-time type safety for whoever edits the SDK, plus compile-time alignment with the server contract. Not runtime types for the LLM driver: it runs `node script.mjs` with no `tsc` in the loop, so types buy nothing at runtime for the driver's throwaway scripts. That is why shipping raw `.ts` (node type-stripping) was rejected. Stripping erases types, it never checks them, so we would still need a dev-time `tsc` on top plus runtime-flag fragility.

### Package

- New workspace package `packages/driver-sdk`. Bare name (matches `agent-runtime`, `api-server-api`, `db`), `type: module`, tasks in `tasks.toml` not `package.json` scripts, wired into `mise run check`.
- Source authored in real `.ts`, free to split into modules (`schema.ts`, `http.ts`, `spawn.ts`) instead of one 254-line file.
- Built with `tsup` to a single dependency-free `.mjs`. `tsup` is what `agent-runtime` (the one package that already ships into an image) uses.
- `vitest` unit tests, starting with the pure `s()` schema-sugar. Retires the "no sandbox unit tests yet" debt called out in the PR body.

### Contract: full lift (typed against the server)

The `/sandboxes` request + response contract is currently validated inline in the api-server app (`packages/api-server/src/apps/harness-api-server/invocation-endpoints.ts:25-52`), not in `api-server-api`. That rail is REST, not tRPC, which is why it never went into the contract package.

Do the full lift:

1. Move the `/sandboxes` request + response Zod schema (and inferred types) out of `invocation-endpoints.ts` into `api-server-api` as a new export.
2. `invocation-endpoints.ts` imports it, so the server validates against the shared schema.
3. `driver-sdk` does `import type` of the inferred request/response types.

Because it is a type-only import, `tsup` erases it: the shipped `.mjs` stays dependency-free and single-file, zero runtime cost. Add `api-server-api` as a devDependency only. Net: server and SDK provably share one contract, so a change to the request shape breaks the SDK build. The half version (SDK keeps hand-written lookalike types) was rejected: it gives the coupling's smell without the drift protection that is the whole point.

### Distribution into the image

- `claude-code`'s Docker build context is its own directory (`scripts/resolve-image.sh` runs `docker build ... packages/agents/claude-code`), so it cannot `COPY` from a sibling package, and it is `FROM` platform-base's slim runner with no pnpm workspace inside.
- Build the package in `platform-base`'s `builder` stage (the only stage in the chain with the workspace). Add `--filter driver-sdk` to the install and a `tsup` build call, alongside the existing `agent-runtime` build.
- `platform-base` runner does `COPY --from=builder .../driver-sdk/dist/driver-sdk.mjs /usr/local/lib/driver-sdk.mjs`, mirroring the existing `model-gateway.mjs` convention (a dependency-free `.mjs` at `/usr/local/lib`).
- Harness-agnostic on purpose (ADR-077, a sandbox IS an ephemeral Agent). Every derived image inherits an inert ~8KB file where no skill points at it. Acceptable.
- Register `packages/driver-sdk` in `resolve-image.sh`'s `PLATFORM_BASE_PATHS`. Without this, editing the SDK does not invalidate the image and you ship stale.

### Skill and naming (glossary alignment)

The glossary retired "Sandbox" as a domain term (commit `8a076f77`): the primitive is an `Invocation`, the actor is the `Driver`, the tool is `report_result` (was `node_done`), the deadline is the `Liveness Deadline`. Naming a new artifact `sandbox-sdk` would mint a fresh artifact on a just-retired term. So:

- Package is `driver-sdk` (named by who holds it, the Driver, per ADR-078 loop-as-driver-program).
- The `.mjs` leaves `packages/agents/claude-code/workspace/.agents/skills/dam-sandbox/`. Only the SKILL.md stays there, importing from `/usr/local/lib/driver-sdk.mjs`.
- Rename the `dam-sandbox` skill to `dam-invoke` (or similar) and `sandbox-sdk.mjs` to `driver-sdk.mjs`.
- The verb `spawn()` can stay. The glossary still uses "spawn" as a verb ("freshly-spawned Sweepable Agent"). It is the noun "Sandbox" that is dead.

### Scope guard

The broader rename (`/sandboxes` route, `SandboxFailed` class, `node_done` to `report_result`) touches more than this package and is already half-done in the migration. Keep this change scoped to: the new `driver-sdk` package, the contract lift into `api-server-api`, the platform-base build + copy, `resolve-image.sh`, and the skill import path. Let the rest of the term retirement ride the existing migration.

packages/driver-sdk/src/spawn.ts
```diff
@@ -0,0 +1,169 @@
...
+    ttlMs,
+    pollMs = 5000,
+    // Default the client backstop to 5min past the server deadline so the server
+    // fails the node first; the driver then sees a real `failed`, not a timeout.
+    timeoutMs = (ttlMs ?? 60 * 60 * 1000) + 5 * 60 * 1000,
+  } = opts;
+
+  if (!prompt) throw new Error("spawn: `prompt` is required");
+  if (schema === undefined) {
+    throw new Error(
```
dam-code-guardian, line 108: 🟢 **Suggestion:** This backstop hardcodes the 1h server default (`ttlMs ?? 60*60*1000`). The SDK stays dependency-free via type-only contract imports, so it cannot cheaply reuse `DEFAULT_INVOCATION_TTL_MS` — but if the server default changes, this fallback silently diverges. A short comment noting the intentional coupling would help future readers.

packages/api-server/src/modules/invocations/services/invocation-liveness.ts
```diff
@@ -0,0 +1,162 @@
...
+      }
+
+      // 1b. Fail Invocations whose target pod crashed/restarted mid-turn. The
+      //     one-shot trigger already fired (recorded in the runtime's persisted
+      //     state), so after a restart the prompt is not redelivered and the
+      //     turn cannot resume — it would otherwise idle until its deadline.
+      const stillRunning = await deps.repo.listRunning(deps.batchSize);
+      for (const row of stillRunning) {
+        try {
+          const restart = await deps.k8s.readPodRestart(`${row.id}-0`);
```
dam-code-guardian, line 104: 🟢 **Suggestion:** This issues one `readPodRestart` K8s GET per `running` invocation every tick (60s). Bounded by `batchSize` (200) and fine for a tracer, but it scales linearly with concurrent invocations — worth a note (or a cheaper batched pod list) if population grows beyond 1.

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `f5bf00b`

### Summary
A tracer PR shipping one platform primitive: spawn an ephemeral Agent with a chosen image + a subset of the driver's connections + a prompt, run it once, and return a schema-validated result. Introduces the `invocations` module (result contract), a `Sweepable` annotation + generic **Agent Sweep** (lifecycle), a `report_result` MCP tool, a liveness sweep, and a new `driver-sdk` package. Retires the "Sandbox" domain term (ADR-077). No prior review on file.

### Findings
- 🟡 **Warning:** Documentation drift — this PR adds a new agent-lifecycle behavior (`Sweepable` agents deleted on hibernate by the Agent Sweep, plus eagerly-reaped ephemeral Invocation targets), but `docs/architecture/agent-lifecycle.md` (whose **Hibernate**/**Delete** sections describe only user/idle-driven paths) is not updated. ADR-077's own `subsystem: agent-lifecycle` frontmatter confirms that's the page. The doc-drift skill (below) has the proposed edit. (`docs/architecture/agent-lifecycle.md`)
- 🟢 **Suggestion:** `spawn()`'s client backstop defaults `timeoutMs` to `(ttlMs ?? 60*60*1000) + 5min`, hardcoding the 1h server default. The SDK does type-only contract imports (to stay dependency-free), so it can't cheaply share the `DEFAULT_INVOCATION_TTL_MS` value — but if the server default ever changes, this fallback silently diverges. Worth a comment noting the coupling. (`packages/driver-sdk/src/spawn.ts`)
- 🟢 **Suggestion:** The liveness sweep's crash fast-fail issues one `readPodRestart` K8s GET per `running` invocation every tick (60s). Bounded by `batchSize` (200) and fine for a tracer, but it scales linearly with concurrent invocations — worth a note if population grows. (`packages/api-server/src/modules/invocations/services/invocation-liveness.ts`)
- ✅ **Looks good:** `complete`/`fail` are atomic conditional writes (`WHERE status = 'running'`), so a late `report_result` racing the liveness sweep can't clobber a terminal row — the `complete()` returning-false path is handled explicitly.
- ✅ **Looks good:** The `${row.id}-0` pod name in the crash check matches the established `podBaseUrl` convention (`${agentId}-0.…`), so it's a correct derivation, not a fragile guess.
- ✅ **Looks good:** `coerceResult` handles harnesses that JSON-encode the `report_result` arg (stringified payloads) while keeping validation structural; eager-reap is best-effort with the Agent Sweep as an explicit backstop.

**Verdict:** minor drift

- ❌ **Check 1 (architecture-page drift):** `docs/architecture/agent-lifecycle.md` — `agent-sweep.ts` (`isSweepDue`: `sweepable && hibernated && grace-elapsed → delete`) introduces automated delete-on-hibernate for `Sweepable` agents, and `invocations-service.ts` spawns ephemeral targets eagerly reaped on terminal. The page's **Hibernate** and **Delete** sections mention neither. Proposed additive fix: a short subsection under **Delete** — "an ephemeral agent stamped `agent-platform.ai/sweepable` is deleted by a generic Agent Sweep once it hibernates, after an optional `Agent Lifetime` grace (default 0); keyed off Agent state, api-server-only, cascading via ownerReferences." Bump `Last verified:` to the PR date.
- ⚠️ **Possible (Check 5):** the new `modules/invocations` result-contract subsystem (REST spawn endpoints, `report_result`, liveness sweep, `invocations` table) has no `docs/architecture/` page and isn't on the landing page. It runs inside api-server rather than as a standalone daemon/controller, so it's a borderline case — a human call on whether the spawn primitive warrants its own page or a section on `agent-lifecycle`.

✅ No findings. The `invocations` module and `driver-sdk` package are a textbook application of the layered/bounded-context architecture: clean `services → domain ← infrastructure` separation (pure domain `invocation-prompt.ts`, port-typed service deps, repository adapter with mapper), `compose.ts` as the sole wiring point, `index.ts` as a tight boundary, and the request/response contract shared from `api-server-api` via type-only imports so server and SDK provably agree. No architecture-rule violations introduced.

### Verdict
COMMENT — Excellent, well-tested primitive with clean architecture; the one actionable item is updating `agent-lifecycle.md` for the new Sweepable/Agent Sweep behavior. The two 🟢 notes are optional.

---

packages/api-server/src/modules/agents/services/agent-sweep.ts
```diff
@@ -0,0 +1,109 @@
...
+ *  needs the hibernation transition time to measure from; if the controller has
+ *  not published one yet, hold off until the next tick rather than guess. */
+export function isSweepDue(agent: InfraAgent, now: Date): boolean {
+  if (!agent.sweepable || !agent.hibernated) return false;
+  if (agent.lifetimeMs <= 0) return true;
+  if (!agent.hibernatedSince) return false;
+  return now.getTime() - agent.hibernatedSince.getTime() >= agent.lifetimeMs;
+}
+
+export function createAgentSweep(deps: CreateAgentSweepDeps): AgentSweep {
```
jezekra1, line 55: can we do this as a bullmq periodic work (ADR 075)?

jezekra1 (changes requested):

jezekra1 (approved):

PR #109, opened 2026-05-06
feat(connections): add 15 Google service OAuth apps with shared credentials

jezekra1:
Closes #114.

## Summary

Adds first-class OAuth descriptors for **gmail + 14 google-* services + youtube** — each appears as its own row under "Apps" with a brand icon and per-service default scopes — and migrates app egress-host configuration from a helm-mounted operator map into the descriptors themselves now that the platform owns the OAuth definitions in code. Drops the "allow all credentials" default in favor of always-selective grants. Also fixes two bugs that surfaced along the way.

### 1. `feat(connections): add 15 Google service OAuth apps with shared credentials`

- 15 service descriptors: `gmail`, `google-admin`, `google-analytics`, `google-calendar`, `google-classroom`, `google-docs`, `google-drive`, `google-forms`, `google-health`, `google-meet`, `google-photos`, `google-search-console`, `google-sheets`, `google-slides`, `google-tasks` — each with brand icon, default scopes (distilled from onecli), and canonical API host.
- New `credentialFamily: "google"` field on `OAuthAppDescriptor` so all 15 services share one Google Cloud OAuth app per user. After connecting one Google service, sibling connect dialogs hide Client ID / Secret and the OAuth flow reuses the stored creds. Tokens stay per-service — only the registration credential is shared.
- `access_type=offline&prompt=consent` for refresh tokens via the engine's existing `extraAuthParams` hook.
- 15 SVG icons sourced from the kagenti onecli fork.

### 2. `feat(connections): add YouTube alongside the Google services`

YouTube uses Google OAuth and slots into the same `credentialFamily: "google"` — connect once across the family, reuse credentials.

### 3. `fix(connections): match onecli scopes for Docs / Sheets / Slides`

Aligned with onecli's choice of `drive.readonly` + `drive.file` for these three (Docs/Sheets/Slides files all live in Drive). Onecli intentionally goes through the Drive API for least-privilege access; the API-specific scopes (`documents`, `spreadsheets`, `presentations`) were a sloppy first pass.

### 4. `fix(connections): use upstream onecli scopes + description for YouTube`

The kagenti fork I'd been reading from is limited to `youtube.readonly`; upstream `onecli/onecli` is read + write + comments. Now matches upstream:

- `youtube.readonly`
- `youtube`
- `youtube.force-ssl`

### 5. `feat(connections): allow overriding inherited family credentials`

[...]

PR #2813, opened 2026-07-16
feat: artifact library — publish, organize, and share agent artifacts

jezekra1:
Closes #2810.

https://github.com/user-attachments/assets/ad7bf8cb-3707-484a-9fec-61b37dc08038

A platform-wide **artifact library**: agents and users publish artifacts (HTML, JSX, markdown, code, text, binary files), organize them in folders, and share them by public link — with passwords, expiry, versions, and view counts. The whole feature ships **behind a per-user experimental-feature flag** (default off).

Design prototypes that guided the UI: [library page](https://slop.radekjezek.com/a/jMuEynQBYq) · [agent tab](https://slop.radekjezek.com/a/ShaL0p6M4f).

## What's included

**Library core** — new `artifact-library` bounded context (contract + server). Content bytes live in the existing S3-compatible object store (the experiments `artifacts` module, reused); metadata in Postgres (migration `0013`). Folders, versioned artifacts, sharing state. Owner-scoped end to end.

**Secure share host** — a dedicated subdomain (`urls.share`, default `share.{domain}`, wired through the ingress) serves *only* user content, host-gated inside the api-server before auth so app cookies/tokens never share its origin. Slop-ported renderer: sandboxed `srcdoc` iframe, in-browser JSX (babel + pinned esm.sh import map), client-sanitized markdown, highlighted code; 410 + grace on expiry with an hourly sweeper. No hand-rolled crypto — Argon2id passwords (`argon2`), HS256 JWT viewer cookies (`jose`) invalidated by a password-version bump. Folder password wins over artifact password.

**Agent surface (MCP)** — `create_artifact`, `create_artifact_upload_url` (presigned direct-to-store transfer through the paired gateway), `update_artifact` (new version), sharing/folder tools, on the existing per-agent platform MCP endpoint. Attribution comes from the mesh-verified agent identity. Responses carry an `internal_link` the agent pastes into chat.

**In-app previews** — HTML/JSX/markdown render live in a sandboxed iframe (opaque origin: no `allow-same-origin`), with a Source toggle, fullscreen, download, and a version switcher. Opening is prefetched-on-hover and deferred past the modal animation so it doesn't stutter.

**Chat integration** — an Artifacts section in the chat sidebar (under Files) lists the agent's artifacts live; clicking docks a preview beside the chat that tracks new versions in real time and offers a Share dialog. `platform://artifacts/<id>` markdown links render as inline preview chips that open the dock.

[...]

PR #2818, opened 2026-07-17
fix(egress): enforce path-scoped HTTPS rules on credential-less hosts

### LINKED ISSUE #2322: Fine-grained network rules are silently ignored, and approving a blocked request quietly opens up the whole site
## Problem

Users can give an agent narrow permission to reach part of an external website — for example, "this agent may only make GET requests to `example.com/status/*`." You can set this up in the network panel, from the command line, or by answering a prompt when the agent is blocked. The interface confirms the rule as active.

**For encrypted (HTTPS) requests, the rule does nothing.** For any website the platform isn't holding a credential for (a connected account, API key, or token), it never actually starts inspecting the method and path of encrypted requests. So the narrow rule the user wrote is invisible to the part of the system that decides what's allowed. This matters because essentially all real agent traffic — APIs, git, package installs — is HTTPS. I confirmed all of this on a live cluster (2026-07-02).

Two things go wrong, and the second is the serious one:

- **The narrow rule is ignored, so the agent gets blocked anyway.** Even though the user allowed `GET /status/*`, the agent's HTTPS request is held and the user is asked to approve reaching the *entire* website — for an encrypted request to an uninspected host, the system can only see the hostname, not the method or path, so it can't tell this request matches the user's rule.

- **Approving that prompt quietly grants far more than the user intended.** When the user clicks Approve to unblock their agent, the prompt is about the whole site (any method, any path), so approving it saves a rule allowing *every* request to that website. I verified this live: after approving, the agent could POST data to a completely different path on the same site and it went straight through. The user's careful "GET status pages only" rule is still sitting in the list looking active, but a hidden "allow everything" rule now sits next to it and wins. The user ends up with a wide-open site while believing they set a tight restriction.

The same blind spot affects "block" rules over HTTPS. If a user allows a whole site but then tries to carve out an exception like "block GET on `/admin/*`," the exception is silently not enforced — I confirmed a request that should have been blocked reached the site normally.

**Not affected: plain HTTP.** Unencrypted requests are inspected correctly — the system can read the method and path directly, so a path-scoped rule on an HTTP request is enforced as written, and approving a blocked HTTP request grants only the specific path. The gap is specific to HTTPS, which is why it can look like it "works fine" if tested over plain HTTP. (It's also why it went unnoticed: no automated test covers a path-scoped HTTPS request end-to-end.)

**Why it matters:** the network rules are a safety feature — the whole point is to let a user say "this agent can only touch this narrow slice of a website." Today, for HTTPS traffic to any site the platform holds no credential for, that promise isn't kept, and the natural way a user tries to use the feature (answer the prompt, click approve) actively widens their own permissions without telling them. It fails in the unsafe direction while looking like it's working.

**How bad, in context:** this only affects hand-written method/path rules on sites the platform holds no credential for — a deliberate power-user action, not a fresh agent's default setup (new agents start from a safe preset). No stored passwords or tokens are ever exposed by this; it's purely about how precisely the traffic rules are enforced. So it's a real hole in a security feature that needs fixing, not an emergency where agents are leaking data by default. It has been broken for about two months.

## Proposed solution

When a user writes a method- or path-scoped rule on any site an agent is allowed to reach, the platform should actually enforce it for HTTPS traffic too: requests that match an "allow" rule go through without a prompt, requests that match a "block" rule are stopped, and anything not covered still triggers the normal approval prompt. What the network panel shows should match what actually happens on the wire — including that answering a prompt grants only what the prompt describes, not the whole site.

### Scope

- Covers method/path rules over HTTPS on sites the platform holds no credential for, created through the network panel, the command line, or an inbox approval.
- Includes making the approval prompt honest — approving a blocked HTTPS request shouldn't silently broaden a narrow rule the user already set.

### Out of scope

- Plain HTTP already enforces method/path rules correctly and isn't affected.
- Sites the platform already holds a credential for (a connected account or key) enforce method/path rules correctly over HTTPS and aren't affected.
- One related trap to note for whoever picks this up: permissions that come bundled with a connected app could hit the same blind spot in the future if such an app ever ships a narrow rule for a site the platform holds no credential for. It doesn't happen today, but it's the same underlying gap.

### Related

- Found while smoke-testing an unrelated cleanup change (#2198). The feature worked when it first shipped (#4) and broke shortly after (#109); it's had no end-to-end test coverage since.
- Reproduction note: adding a path-scoped rule restarts the agent's gateway (~5–15s); a request fired during that window fails to connect and never reaches the approval inbox. Test over HTTPS explicitly (a bare hostname or `http://` will appear to work correctly and won't reproduce the gap).

jezekra1:
Fixes #2322 — method/path-scoped network rules were silently ignored for HTTPS traffic to hosts the platform holds no credential for, and approving a blocked request quietly opened the whole site.

## Root cause

#109's always-selective grants rework dropped **allow-only Secrets** in the controller's grant filter — they are never in any grant list — so path-scoped rules on credential-less hosts never got their L7 MITM chain. HTTPS to those hosts stayed on the SNI-only L4 catch-all: the rules could never match, holds prompted host-wide, and approve-permanent persisted the hold payload verbatim as a hidden `(host, *, *)` allow that then won every future check.

## Fix

- **controller** — allow-only Secrets pass through `filterByGrants` unconditionally: they are enforcement plumbing, not credentials, so grants don't apply.
- **api-server** — inbox-written rules (approve-permanent / deny-forever) now run the same L7 promotion as the manual create/update path: a new egress-rule writer in the egress-rules module ensures the allow-only Secret for narrow rules, and approvals thread the agent owner's sub through for the Secret label.

## Tests

- Go: allow-only Secrets survive grant filtering regardless of grant lists.
- TS: writer promotes narrow rules (and skips host-wide ones); approvals pass the row's owner sub.
- New e2e browser spec (`11-egress-path-rules.spec.ts`) closing the coverage gap the issue calls out: a panel-created `GET /status/*` rule on a credential-less host is enforced over HTTPS without a prompt; an uncovered path prompts with the real method+path; "Allow permanently" unlocks exactly that path; no hidden host-wide rule is written; a third path stays gated.

Verified on a fresh cluster built from this branch: full Playwright suite **16 passed (2.9m)**, `mise run check` green, controller + api-server unit suites pass.

## Also in this PR

- e2e suite reorder: `connection-regrant` now runs last — recreating a connection can wedge the gateway rollout on the deleted Secret (pre-existing race, tracked separately), and nothing after that spec needs the agent.

### DIFF
--- a/packages/api-server/src/modules/egress-rules/services/egress-rule-writer.ts
+++ b/packages/api-server/src/modules/egress-rules/services/egress-rule-writer.ts
@@ -1,0 +1,38 @@
+import type {
+  EgressRulesRepository,
+  NewEgressRule,
+} from "../infrastructure/egress-rules-repository.js";
+import { needsL7Promotion } from "../domain/l7-promotion.js";
+import type { K8sAllowOnlySecretsPort } from "../infrastructure/k8s-allow-only-secrets-port.js";
+
+/**
+ * Insert path for rules born from inbox verdicts (approve-permanent /
+ * deny-forever). Unlike the user-scoped service, it takes the agent owner's
+ * sub per call — the approvals module resolves it from the pending row.
+ *
+ * Mirrors the manual-create path's L7 promotion: a narrow rule written from
+ * an inbox verdict (possible when the held request was plain HTTP, where
+ * method/path are visible without MITM) must promote its host onto the L7
+ * chain, or the rule's HTTPS half is silently unenforced (#2322).
+ */
+export interface CreateEgressRuleWriterDeps {
+  repo: EgressRulesRepository;
+  /** Optional so non-cluster contexts (tests) can skip the side effect. */
+  allowOnlySecrets?: K8sAllowOnlySecretsPort;
+}
+
+export function createEgressRuleWriter(deps: CreateEgressRuleWriterDeps): {
+  insert(input: NewEgressRule & { ownerSub: string }): Promise<void>;
+} {
+  return {
+    async insert({ ownerSub, ...row }) {
+      await deps.repo.insert(row);
+      if (
+        needsL7Promotion(row.method, row.pathPattern, row.port) &&
+        deps.allowOnlySecrets
+      ) {
+        await deps.allowOnlySecrets.ensure(ownerSub, row.host);
+      }
+    },
+  };
+}
--- a/packages/e2e/playwright/src/tests/11-egress-path-rules.spec.ts
+++ b/packages/e2e/playwright/src/tests/11-egress-path-rules.spec.ts
@@ -1,0 +1,151 @@
+import { expect, test } from "@playwright/test";
+
+import { waitForAgentRunning } from "../lib/agents.js";
+import { createApiClient } from "../lib/api-client.js";
+import { getAccessToken } from "../lib/auth.js";
+import { agentName } from "../lib/fixtures.js";
+
+// Path-scoped egress rules over HTTPS on a host the platform holds no
+// credential for (#2322). The rule must promote the host onto the gateway's
+// L7 (MITM) chain — the SNI-only L4 catch-all cannot see method/path, so
+// without promotion the narrow rule is silently ignored, the inbox prompts
+// for the whole site, and approving writes a hidden host-wide allow.
+//
+// postman-echo.com is a public echo service (same external-dependency class
+// as 05-injection's httpbingo.org). It deliberately has NO connection: the
+// gateway starts with no chain for it, which is the broken configuration.
+const host = "postman-echo.com";
+const allowedUrl = `https://${host}/status/204`;
+const uncoveredUrl = `https://${host}/get`;
+const stillGatedUrl = `https://${host}/headers`;
+
+test("path-scoped HTTPS rules are enforced and approvals stay narrow", async ({
+  page,
+}) => {
+  test.setTimeout(420_000);
+
+  const token = await getAccessToken();
+  const api = createApiClient(token);
+  const agentId = await waitForAgentRunning(api, agentName);
+
+  await test.step("add a narrow allow rule in the network panel", async () => {
+    await page.goto(`/sandboxes/${encodeURIComponent(agentId)}`);
+    const net = page.locator("section").filter({ hasText: "Network access" });
+    await net.getByLabel("Host").fill(host);
+    await net.getByLabel("Method").selectOption("GET");
+    await net.getByLabel("Path").fill("/status/*");
+    await net.getByLabel("Verdict").selectOption("allow");
+    await net.getByRole("button", { name: "Add rule" }).click();
+    await page.getByRole("button", { name: "Submit changes" }).click();
+    // Path rules roll the gateway; the save flow confirms before committing.
+    await page.getByRole("button", { name: "Save & restart" }).click();
+    await expect
+      .poll(
+        async () =>
+          (await api.egressRules.listForAgent.query({ agentId })).some(
+            (r) => r.host === host && r.pathPattern === "/status/*",
+          ),
+        { timeout: 30_000, message: "panel-created rule did not persist" },
+      )
+      .toBe(true);
+  });
+
+  await test.step("a request matching the rule passes without a prompt", async () => {
+    // The poll absorbs the gateway roll (allow-only Secret → leaf cert SAN →
+    // new MITM chain, ≤30s informer resync + pod restart). Before the fix
+    // this never converges: the request holds for a human verdict instead.
+    await expect
+      .poll(
+        async () => {
+          try {
+            const { status } = await api.e2e.performFetch.mutate({
+              agentId,
+              url: allowedUrl,
+            });
+            return status;
+          } catch {
+            return 0;
+          }
+        },
+        {
+          timeout: 120_000,
+          intervals: [3_000],
+          message: "allowed path did not go through without approval",
+        },
+      )
+      .toBe(204);
+  });
+
+  await test.step("an uncovered path prompts with method+path, not the whole site", async () => {
+    // Fire and forget: Envoy holds the request for a verdict; the mock
+    // agent's fetch gives up client-side long before the hold expires.
+    void api.e2e.performFetch
+      .mutate({ agentId, url: uncoveredUrl })
+      .catch(() => {});
+    await page.goto("/inbox");
+    const row = page
+      .locator("li")
+      .filter({ hasText: `GET ${host}` })
+      .first();
+    await expect(row).toBeVisible({ timeout: 30_000 });
+    // The L7 chain saw the decrypted request — an SNI-only hold would render
+    // the path as "*" and approving it would open the whole site.
+    await expect(row.getByText("/get", { exact: true })).toBeVisible();
+    await row.getByRole("button", { name: "Allow permanently" }).click();
+  });
+
+  await test.step("the approval unlocks exactly the approved path", async () => {
+    await expect
+      .poll(
+        async () => {
+          try {
+            const { status } = await api.e2e.performFetch.mutate({
+              agentId,
+              url: uncoveredUrl,
+            });
+            return status;
+          } catch {
+            return 0;
+          }
+        },
+        {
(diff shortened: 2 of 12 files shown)


packages/api-server/src/modules/egress-rules/services/egress-rule-writer.ts
```diff
@@ -0,0 +1,38 @@
...
+  /** Optional so non-cluster contexts (tests) can skip the side effect. */
+  allowOnlySecrets?: K8sAllowOnlySecretsPort;
+}
+
+export function createEgressRuleWriter(deps: CreateEgressRuleWriterDeps): {
+  insert(input: NewEgressRule & { ownerSub: string }): Promise<void>;
+} {
+  return {
+    async insert({ ownerSub, ...row }) {
+      await deps.repo.insert(row);
```
dam-code-guardian, line 29: 🟢 **Suggestion:** The guard-then-`ensure` orchestration here is also duplicated in `egress-rules-service.ts` (create + update paths). The `needsL7Promotion` predicate is now shared, but the guard+`ensure` orchestration is copied. If a fourth writer ever appears, a small `promoteIfNeeded(port, ownerSub, rule)` helper in the module would collapse it. Optional — the two services legitimately differ in how they source `ownerSub`.

packages/e2e/playwright/src/tests/11-egress-path-rules.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+
+// Path-scoped egress rules over HTTPS on a host the platform holds no
+// credential for (#2322). The rule must promote the host onto the gateway's
+// L7 (MITM) chain — the SNI-only L4 catch-all cannot see method/path, so
+// without promotion the narrow rule is silently ignored, the inbox prompts
+// for the whole site, and approving writes a hidden host-wide allow.
+//
+// postman-echo.com is a public echo service (same external-dependency class
+// as 05-injection's httpbingo.org). It deliberately has NO connection: the
+// gateway starts with no chain for it, which is the broken configuration.
```
dam-code-guardian, line 16: 🟢 **Suggestion:** This spec depends on the external `postman-echo.com`. Same external-dependency class as the existing `httpbingo.org` spec, and the generous polls absorb the gateway roll — but it does couple CI green to a third-party service's uptime. Worth keeping in mind if this spec starts flaking.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `1cb284a`

### Summary
Fixes #2322: path/method-scoped egress rules were silently ignored for HTTPS to credential-less hosts because allow-only Secrets were dropped by the controller's `filterByGrants` (regression from #109), and inbox-written rules never ran L7 promotion. The fix passes allow-only Secrets through grant filtering unconditionally, extracts `needsL7Promotion` into a shared domain module, and adds an `egress-rule-writer` service so inbox verdicts promote narrow rules the same way manual creation does.

### Findings
- ✅ **Looks good:** Root cause correctly isolated — allow-only Secrets are enforcement plumbing, not credentials, so exempting them from grant gating (`envoy.go` `filterByGrants`) is the right fix, and the comment explains *why* to prevent a future re-regression.
- ✅ **Looks good:** `needsL7Promotion` extracted into a pure `domain/l7-promotion.ts` and shared by both write paths — single source of truth for the promotion predicate.
- ✅ **Looks good:** Strong test coverage across all three layers — Go unit (allow-only survives filtering with/without grants), TS unit (narrow promotes, host-wide doesn't, approval threads `ownerSub`), and a full e2e spec asserting the exact bug: enforced narrow rule, method+path prompt, no hidden host-wide allow, third path still gated.
- 🟢 **Suggestion:** The guard-then-`ensure` orchestration is now repeated in `egress-rule-writer.ts` and `egress-rules-service.ts` (create + update). The predicate is shared; only the orchestration is copied. A small `promoteIfNeeded(port, ownerSub, rule)` helper would collapse it if a fourth writer ever appears — optional, since the two services legitimately differ in how they source `ownerSub`.
- 🟢 **Suggestion:** The e2e spec depends on the external `postman-echo.com`. It's the same external-dependency class as the existing `httpbingo.org` spec and generous polls absorb the gateway roll, but it does couple CI green to a third-party service's availability.

✅ No documentation drift detected. `docs/architecture/connections.md` replaced embedded runtime-contract TS/YAML blocks with prose + directory link-outs (a move *away* from volatile content) and bumped `Last verified:` to 2026-07-17; no ADR references leaked. The behavior change matches what `security-and-credentials.md` already documents about allow-only chains and L7 promotion.

🟢 No architecture violations. Domain purity upheld (`l7-promotion.ts` is pure, zero-dep), module boundary respected (approvals consumes egress-rule writing through a consumer-defined `EgressRuleWriter` port, concrete adapter bound at the `app.ts` composition root — no cross-module internal imports), and the new `createEgressRuleWriter` factory follows the DI/ports pattern with `allowOnlySecrets` optional so tests skip the K8s side-effect. One optional 🟢 (promotion orchestration duplicated across two services) noted above.

### Verdict
APPROVE — Correct, well-explained fix for a subtle security-relevant regression, with thorough three-layer test coverage and clean architecture; only optional suggestions remain.

---

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `8935f0e`

## PR #2818: fix(egress): enforce path-scoped HTTPS rules on credential-less hosts
**Author:** jezekra1 | **Branch:** fix/https-network-rules#locki-fpbk11y6 → main | **Changes:** +335 −23 (12 files)

### Summary
Fixes #2322: path/method-scoped egress rules were silently ignored for HTTPS to credential-less hosts because allow-only Secrets were dropped by the controller's `filterByGrants` (regression from #109), and inbox-written rules never ran L7 promotion. The fix exempts allow-only Secrets from grant gating, extracts `needsL7Promotion` into a pure `domain/l7-promotion.ts`, and adds an `egress-rule-writer` service so inbox verdicts promote narrow rules the same way manual creation does.

### Changes since last review
Previous HEAD: 1cb284a (2026-07-17T13:27:38Z) — verdict APPROVE

- 🔁 **Still present:** guard-then-`ensure` orchestration duplicated in `egress-rule-writer.ts` and `egress-rules-service.ts` — carried over (🟢, optional)
- 🔁 **Still present:** e2e spec depends on external `postman-echo.com` — carried over (🟢)

_This push is a rebase onto current main (which merged #2813's artifact-library work); the PR's own 12-file diff is byte-identical to the prior review. The only textual movement is a mechanical hoist of `allowOnlySecrets` in `app.ts` so the single port instance is shared between the egress-rules module and the new `createEgressRuleWriterAdapter(db, allowOnlySecrets)` argument — no behavior change._

### Findings
- ✅ **Looks good:** Rebase cleanly re-applies the fix on top of the merged artifact-library changes; `app.ts` correctly threads the one `allowOnlySecrets` port into both the module compose and the writer adapter.
- 🟢 **Suggestion:** guard-then-`ensure` orchestration duplicated in `egress-rule-writer.ts` and `egress-rules-service.ts` (create+update); an optional `promoteIfNeeded(port, ownerSub, rule)` helper would collapse it if a fourth writer ever appears (`packages/api-server/src/modules/egress-rules/services/egress-rule-writer.ts:29`).
- 🟢 **Suggestion:** e2e spec couples CI green to third-party `postman-echo.com` uptime — same class as the existing `httpbingo.org` spec; worth watching if it flakes (`packages/e2e/playwright/src/tests/11-egress-path-rules.spec.ts:16`).

✅ No documentation drift detected. The PR touches no docs at this HEAD; it restores already-documented path-scoped enforcement behavior (no subsystem responsibility change). `connections.md` already carries `Last verified: 2026-07-17`.

🟢 No architecture violations. Domain purity upheld (`l7-promotion.ts` is pure, zero-dep, and the single source of truth for the promotion predicate), module boundary respected (consumer-defined `EgressRuleWriter` port, adapter bound at the composition root), ports/DI pattern followed. One optional 🟢 carryover (duplicated promotion orchestration).

### Verdict
APPROVE — Correct, well-explained fix for a subtle security-relevant regression; the rebase re-applies it cleanly with no behavioral change since the prior approval.

---

Issue #2831, opened 2026-07-20
Choose how an agent is shared when connecting it to a Slack channel


pilartomas:
## Problem

Connecting a Slack channel to an agent today only supports a per-person model: everyone who wants to talk to the agent has to log in and link their Slack identity first, the owner maintains an allowed-users list, and each person's messages run under *their own* connected accounts.

That's the right model for some cases, but it's heavy for the most common one: a team that just wants a shared assistant sitting in a channel. They want anyone already in the channel to be able to talk to it — no per-person login, no allow-list to curate — and they expect it to act as *itself* (its own connected accounts), like a team bot. Today there's no way to set that up.

## Proposed solution

When binding a Slack channel to an agent, let the owner pick an **access mode**:

- **Shared team assistant** — anyone Slack lets into the channel can talk to the agent, with no per-person login. The agent acts under its own connected accounts, and messages are attributed to whoever spoke so multi-person threads stay legible.
- **Per-person (default, unchanged)** — today's behavior: identity linking, the allowed-users list, and each person driving the agent under their own credentials.

The choice is made when the channel is connected and is fixed for that binding — to change it you disconnect and reconnect, so a channel never silently changes whose credentials it lends. The mode is shown when connecting from both the UI (with a short note explaining what each mode means for consent and credentials) and the command line, and the current mode is visible in the channel list.

### Scope

- Adds shared mode **beside** the existing per-person model; the per-person path is unchanged and remains the default.
- Applies to Slack channel bindings.

### Dependencies

- Subissue of the Slack rework (#2830).
- Implemented by #2790.

PR #2790, opened 2026-07-14
feat(channels): bind-time access modes — shared vs person-scoped (ADR-075)

pilartomas:
Implements **ADR-076** (#2777, merged): channel access mode is chosen **per channel at bind time**, and — on top of that — adds an **in-chat bind flow for Slack** so a channel can be connected without touching the platform UI/CLI.

- **Shared ("system Agent") mode** — place-scoped: anyone Slack admits to the channel drives the Agent; every turn relays single-track to the main agent pod under the Agent's own credentials; turns attributed by Slack user id (`externalActorId`, `actorSub: null`); the **binding owner's** ToU acceptance gates every turn (Telegram parity); prompts are speaker-labelled for multi-speaker thread sessions.
- **Person-scoped mode (default)** — today's model, byte-identical: identity linking, the allowed-users gate, owner relay, and per-turn foreign-replier forks.

> This PR previously implemented the ADR's first draft (delete forks, identity linking, allowed-users). The branch is rebuilt from main per the revised, accepted decision: **nothing is deleted** — shared mode lands beside the existing model. Old head preserved at tag `archive/place-scoped-v1`.

## Commits — per-channel access modes (ADR-076)

1. **feat(contract)** — `SlackChannel.mode` rides the config JSONB (only `"shared"` is stored; absent = person-scoped → **zero migration**, existing rows unaffected). Optional `mode` on `agentConnectSlackInputSchema` (old CLIs unaffected). In-place mode flips rejected (`ModeChangeRequiresRebind` → `PRECONDITION_FAILED`): switching = disconnect + reconnect, so a rebind never silently changes whose credentials a channel lends.
2. **feat(channels)** — `ChannelRegistry.resolveSlackBinding` returns `{instanceName, owner, mode}` (per-mention DB lookup, no event staleness); binding lookup hoisted above the identity-link gate; shared branch skips identity/allow-list, logs `basis: "place"`, gates ToU on the binding-row owner, relays via `relayOwnerTurn` with `actorSub: null` + `externalActorId` and a `<@U…>: ` speaker label. Person-scoped path unchanged from `identityLinks.resolve` onward.
3. **feat(ui)** — access-mode radio with inline consent copy on the Slack channel form; mode is read-only once bound ("disconnect and reconnect to change it"); allowed-users editor hidden in shared mode.
[...]

PR #2841, opened 2026-07-20
feat(channels): ambient mode — a shared channel agent that chimes in unmentioned

### LINKED ISSUE #2833: Let a shared channel agent respond on its own, without being @mentioned (ambient mode)
## Problem

An agent only ever speaks when it's directly `@mentioned`. In a shared channel where the agent is meant to act as a team assistant, that's a poor fit. People have to remember to tag it every single time — even when it's obvious the agent could help, like a question it clearly knows the answer to or a task it could pick up. It can't follow along and jump in when it would be useful, so it feels inert and the whole burden of summoning it falls on the people in the channel.

A genuinely useful team assistant should behave more like a helpful colleague sitting in the channel: noticing when it's relevant and chiming in on its own — while staying quiet the rest of the time so it doesn't become noise.

## Proposed solution

Add an **ambient mode** for a shared team-assistant agent in a channel. When it's on, the agent reads along with the channel conversation and decides for itself when to respond, without needing to be mentioned:

- **Chimes in when it can clearly help** — answering a question it can answer, picking up a task someone described, catching an obvious mistake — and **stays silent otherwise**. Judicious, not chatty.
- **Still responds when explicitly `@mentioned`**, exactly as today.
- **Owner opt-in, per channel, off by default** — an always-listening agent changes what people expect about privacy and attention, so a channel becomes ambient only when its owner deliberately turns it on.

### Transparency / control

- The channel owner controls whether ambient mode is on; it's off unless they enable it.
- Because the agent now reads and may react to messages it wasn't tagged in, that should be clear to the people in the channel.
- There should be an easy way to dial it back or tell it to stay quiet if it's chiming in too eagerly.

### Scope

- Applies to shared team-assistant channel bindings (the shared channel mode).
- Opt-in per channel.

### Out of scope

- Direct messages, where every message is already aimed at the agent and there's no mention-vs-not ambiguity.

### Dependencies

- Subissue of the Slack rework (#2830).
- Builds on the shared team-assistant channel mode (#2831).

pilartomas:
Closes #2833. Builds on the shared team-assistant channel mode (#2831, merged via #2790).

## What

A shared Slack binding can now run in **ambient mode**: the agent reads along with the channel conversation and decides for itself when to chime in — answering a question it can answer, picking up a task someone described, flagging a clear mistake — and stays silent otherwise. Mentions keep today's behavior exactly; ambient adds a second, quieter inbound path. Owner opt-in, per channel, off by default.

## How it works

**Session model** (modeled on Claude-in-Slack): one session per thread, plus one rolling **ambient session per channel**.

- Thread replies relay into the thread's own session — the same key a mention there would resume.
- Top-level messages share the channel's ambient session (a synthetic key in the existing thread-key slot), so the agent genuinely follows the conversation instead of starting cold per message. Traffic is serialized per channel and **coalesced**: messages arriving mid-turn flush as one multi-message prompt, so a busy channel never races concurrent prompts into one session.
- When the agent chimes in, the reply threads under the triggering message; follow-ups in that thread become an ordinary thread session, with thread-history injection carrying context across.

**Declining and identity.** Each ambient prompt wraps the message(s) with instructions to reply with exactly `NO_RESPONSE` when the agent has nothing clearly helpful to add; the worker swallows that reply (an empty turn counts as a decline too). The frame announces the install's bot identity (from the brand config) and instructs that a message calling the agent by name — the bot name, or the name the agent knows itself by — is answered like a mention. The agent's own name is deliberately not injected: that identity belongs to the agent's workspace setup (e.g. its CLAUDE.md). Ambient turns post no reactions, no wake notices, and failures are logged and evented, never posted — nobody summoned the agent.

**Message intake.** The gateway subscribes to plain channel messages and pre-filters bot posts (including the agent's own replies, preventing loops), subtyped events (edits/joins), DMs, and bot-mentions (those stay on the mention path). The app manifest gains the `message.channels`/`message.groups` event subscriptions and the history scopes they require.

## Transparency and control

- **Opt-in surfaces:** a checkbox on the UI channel form (shared mode only), `dam channel slack connect --mode shared --ambient`, and an in-chat `/<brand> ambient on|off` command (`ambient` alone reports the state). Unlike the access mode, ambient is mutable — a same-mode re-connect updates it in place.
- **Authorization:** the in-chat toggle is allowed for the binder or the agent's owner — the same rule as in-chat unbind.
- **Announcements:** every flip is posted channel-visibly (people should know when an agent starts reading every message) and recorded in the security log. Outbound posting now single-flight-starts the gateway so the announcement right after the first-ever connect isn't dropped.
- **Audit:** each relayed ambient message logs a place-basis allow marked as ambient-triggered; the binding owner's Terms-of-Use acceptance gates ambient turns like any shared turn — re-checked at drain time, so a batch queued before an unbind/rebind can't ride a stale gate.

## Testing

- `slack-ambient.test.ts`: 24 unit tests — gating (off by default, shared-only, silent ToU skips), decline token (incl. empty turns and wrapped tokens), session routing, coalescing, drain failure isolation (no unhandled rejection), quietness, audit entries, and the in-chat command's authorization + announcements.
- `10-slack-ambient.spec.ts` e2e: an unmentioned message gets an ambient reply threaded under it (with the channel-visible enable notice); dialing back returns the channel to mentions-only silence. Also narrows the `user-env`/`connection-regrant` `testMatch` globs to exact names so numeral prefixes can't double-run Slack specs outside their chains.
- `mise run check` and `mise run test` pass (872 tests). The `controller:check:crd-backwards-compatibility` failure in my environment is a git-worktree artifact (`crdify`'s git loader can't resolve refs through a worktree `.git` pointer) — no CRDs are touched here.

## Manual verification (not run — needs a live cluster)

1. `mise run cluster:install`, connect a channel in shared mode with ambient on (UI or CLI) — the channel gets the ambient notice.
2. Post a plain message the agent can answer → reply appears threaded under it, no 👀 reaction; post chit-chat → silence.
3. `@mention` the agent → unchanged addressed-turn behavior (reaction, thread reply).
4. `/dam ambient off` as the binder → channel-visible notice; plain messages are ignored again; `/dam ambient` reports state; a non-binder gets the authorization refusal.


packages/ui/src/modules/sessions/components/channels-panel.tsx
```diff
@@ -81,6 +81,9 @@ function SlackChannelForm({ agent }: { agent: AgentView | undefined }) {
       ? (slackChannel.mode ?? "person-scoped")
       : "person-scoped",
   );
+  const [ambient, setAmbient] = useState(
```
dam-code-guardian, line 84: 🟡 **Warning:** Adding `ambient` brings `SlackChannelForm` to 8 `useState` calls, with three-way conditional submit logic and a manual `dirty` flag. That's over the ~5-`useState` extraction trigger and the RHF+Zod form threshold (multi-field + dirty-tracking). This is a pre-existing legacy `useState` form, so per "touch-it = migrate-it" this would be the moment to extract the state into a hook or convert to RHF+Zod (`formState.isDirty` replaces the manual flag). Non-blocking — the added lines themselves are clean; noting the accumulating form-state drift.

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `d54ded8`

## PR #2841: feat(channels): ambient mode — a shared channel agent that chimes in unmentioned
**Author:** pilartomas | **Branch:** feat/slack-ambient-mode → main | **Changes:** +1479 −78 (29 files)

### Summary
Adds ambient mode to shared Slack bindings: the agent reads along with the whole channel, decides for itself when to chime in (and replies with a swallowed decline token otherwise), and stays silent by default. Owner opt-in per channel via UI checkbox, `--ambient` CLI flag, and an in-chat `/<brand> ambient on|off` command. Session model is one session per thread plus a rolling per-channel ambient session, with per-channel serialization + coalescing. Closes #2833.

### Findings
- 🟡 **Warning:** `SlackChannelForm` now holds **8 `useState` calls** (`channels-panel.tsx:84` adds `ambient`) with three-way conditional submit logic and manual `dirty` tracking — over the ~5-`useState` extraction trigger and the RHF+Zod form threshold. This is a pre-existing legacy `useState` form; the diff extends it rather than migrating. Per "touch-it = migrate-it", extracting the form state into a hook (or converting to RHF+Zod) would be the ideal moment, but that's a larger refactor than this feature intends — flagging, not blocking.
- ✅ **Looks good:** Loop prevention is thorough at the gateway (`bolt-slack-gateway.ts`): skips `bot_id`, skips messages containing `<@botUserId>` (those arrive via app_mention), skips subtyped events except `file_share`, and filters DMs — so the agent can't react to its own posts.
- ✅ **Looks good:** The ambient drain (`slack.ts drainAmbientQueue`) re-resolves the binding and re-checks owner ToU **at drain time**, so a batch queued before an unbind/rebind/dial-back can't ride a stale gate; the drain also swallows its own rejections with a `draining=false` reset in `finally`, preventing both unhandled rejections and a wedged queue.
- ✅ **Looks good:** `isAmbientDecline` treats an empty turn and a whitespace/quote/backtick-wrapped `NO_RESPONSE` as declines, avoiding the mention path's unsolicited "(no response)" fallback — matched by explicit unit tests.
- ✅ **Looks good:** `ambient` is validated `⇒ shared` in three layers — contract `.refine` (`schemas.ts`), a service-side re-derivation guarding direct callers (`agents-service.ts`), and CLI/UI gating — proper defense-in-depth.
- ✅ **Looks good:** Every ambient flip is security-logged and announced channel-visibly; the single-flight `ensureGateway`/`startGateway` fixes the dropped-announcement race on first-ever connect.
- 🟢 **Suggestion:** `SlackChannelMessageEvent` is a type alias of `SlackMentionEvent` (`slack-gateway.ts`). Correct today, but if the two event shapes ever diverge the alias silently couples them; the existing "only the trigger differs" comment covers it — just worth watching.

✅ No documentation drift detected. `docs/architecture/channels.md` was thoroughly updated in the same PR: a new "Ambient mode (shared bindings only)" subsection, an "Ambient routing" observation, an "Ambient message intake" bullet, and the persistence-store note gained the ambient flag. `Last verified:` is current (2026-07-20), no ADR references leaked, and the one `deploy/slack-app-manifest.yaml` reference is a proper link-out for volatile content.

✅ No findings. Contract validation stays in the contract package (`schemas.ts` `.refine`), the router delegates, and the service re-derives the invariant. The new `setSlackChannelAmbient` export from `agents/index.ts` sits beside its already-exported sibling repository writers (established local convention). The in-memory `ambientQueues` Map is per-channel transport-coordination state correctly co-located in the long-lived Slack worker adapter alongside the existing `gateway`/`gatewayStarting` state — not domain state, no boundary violation.

🟡 See the legacy-form finding above (8 `useState` + manual dirty-tracking over threshold). The added lines themselves are clean: shared `Checkbox` primitive, no static `style={{}}`, no `any`/`as`, correct `mode === "shared"` conditional gate, safe `slackChannel?.type` narrowing in the initializer.

### Verdict
COMMENT — Carefully engineered feature with strong test coverage (24 unit + e2e), thorough loop/gate/race handling, and matching doc updates; the one non-blocking note is that it extends an already-over-threshold legacy `useState` form instead of migrating it.

---

PR #2844, opened 2026-07-20
feat(cli): accept the Terms of Use from the CLI

### LINKED ISSUE #2755: Allow accepting the Terms of Use from the CLI
## Problem

A user can log in via the CLI device flow without a browser, but the first command then fails with a Terms-of-Use gate whose only remedy is "open the browser to accept." In headless/CI/no-browser setups there's no way through — during dam-dev bring-up we had to bypass the CLI and accept ToU by calling the API directly. The device flow exists precisely so the CLI works without a browser; this gate reintroduces the browser dependency and blocks the flow.

## Goal

A user who authenticates via the CLI should be able to review and accept the current Terms of Use entirely from the CLI, with no browser required.

## Acceptance criteria

- A CLI command lets the user view the current ToU and accept them.
- After accepting, previously-gated commands succeed.
- Works non-interactively (for CI/headless), while still recording which version was accepted.
- When a command is blocked by the ToU gate, the error points the user to the CLI acceptance path.

## Notes

The server already supports acceptance (the web UI uses it); this is a CLI gap only. How much of the terms text must be shown before a non-interactive accept is a product/legal call to settle in the issue.

PetrBulanek:
## What this does

A user who signs in through the `dam` CLI device flow (no browser) can now review and accept the Terms of Use entirely from the CLI. Previously the first command after login could hit the Terms-of-Use gate whose only remedy was "open the browser to accept" — a dead end in headless/CI/no-browser setups.

A new `dam terms` command group:

- **`dam terms show`** — print the current Terms of Use text, version, and your acceptance state.
- **`dam terms status`** — a quick, script-friendly check (no text dump) that exits non-zero when you haven't accepted the current version, so it works as a CI gate.
- **`dam terms accept`** — accept the current terms; interactive with a confirmation prompt on a terminal, `--yes` for non-interactive/CI use, and `--expect-version` to assert the version you reviewed before accepting.

When a command is blocked by the Terms-of-Use gate, the error now points at `dam terms accept` instead of the browser.

Verified end-to-end against a live cluster: the gated → review → accept → unblocked flow, interactive and `--yes` acceptance, the version-mismatch guard, and the rewritten gate hint.

Closes #2755

### DIFF
--- a/packages/cli/src/modules/terms/commands/status.ts
+++ b/packages/cli/src/modules/terms/commands/status.ts
@@ -1,0 +1,67 @@
+import { Command } from "commander";
+import type { CompatService, ConfigService } from "../../cli/index.js";
+import {
+  EXIT_BELOW_FLOOR,
+  EXIT_RUNTIME_FAILURE,
+  EXIT_SUCCESS,
+} from "../../shared/exit-codes.js";
+import { resolveActiveHost } from "../../shared/preflight.js";
+import { writeStdoutAndExit } from "../../shared/stdout.js";
+import { printServiceError } from "../../shared/trpc/print.js";
+import type { TermsService } from "../services/terms-service.js";
+
+export function buildStatusCommand(deps: {
+  compatService: CompatService;
+  configService: ConfigService;
+  createTermsService: (host: string) => TermsService;
+}): Command {
+  return new Command("status")
+    .description(
+      "Check whether you've accepted the current Terms of Use — no text dump; exits non-zero when not accepted, so it works as a CI gate",
+    )
+    .option(
+      "--server <url>",
+      "override the configured server URL for this call",
+    )
+    .option("--json", "emit the acceptance state as JSON")
+    .action(async (opts: { server?: string; json?: boolean }) => {
+      const host = await resolveActiveHost(deps, {
+        flag: opts.server ? { server: opts.server } : undefined,
+        exitCodes: {
+          runtimeFailure: EXIT_RUNTIME_FAILURE,
+          belowFloor: EXIT_BELOW_FLOOR,
+        },
+      });
+      const service = deps.createTermsService(host);
+      const [current, latest] = await Promise.all([
+        service.current(),
+        service.latestAcceptance(),
+      ]);
+      if (!current.ok) {
+        printServiceError(current.error, host);
+        process.exit(EXIT_RUNTIME_FAILURE);
+      }
+      if (!latest.ok) {
+        printServiceError(latest.error, host);
+        process.exit(EXIT_RUNTIME_FAILURE);
+      }
+
+      const acceptedVersion = latest.value?.version ?? null;
+      const accepted = acceptedVersion === current.value.version;
+      const exitCode = accepted ? EXIT_SUCCESS : EXIT_RUNTIME_FAILURE;
+
+      if (opts.json) {
+        return writeStdoutAndExit(
+          `${JSON.stringify({ currentVersion: current.value.version, acceptedVersion, accepted })}\n`,
+          exitCode,
+        );
+      }
+
+      const line = accepted
+        ? `Terms of Use ${current.value.version}: accepted.`
+        : acceptedVersion
+          ? `Terms of Use ${current.value.version}: NOT accepted (you accepted ${acceptedVersion}).`
+          : `Terms of Use ${current.value.version}: NOT accepted (no acceptance on record).`;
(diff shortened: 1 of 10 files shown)


packages/cli/src/modules/terms/commands/status.ts
```diff
@@ -0,0 +1,67 @@
...
+        process.exit(EXIT_RUNTIME_FAILURE);
+      }
+      if (!latest.ok) {
+        printServiceError(latest.error, host);
+        process.exit(EXIT_RUNTIME_FAILURE);
+      }
+
+      const acceptedVersion = latest.value?.version ?? null;
+      const accepted = acceptedVersion === current.value.version;
+      const exitCode = accepted ? EXIT_SUCCESS : EXIT_RUNTIME_FAILURE;
```
dam-code-guardian: 🟡 **Warning:** This exits `EXIT_RUNTIME_FAILURE` (1) for the *expected* "not accepted" outcome — the same code returned for genuine failures (server unreachable / host resolution, lines 42 and 46). A CI gate calling `dam terms status` can't distinguish "terms need accepting" (actionable) from "the call broke". The sibling `auth status` introduced `EXIT_AUTH_STATUS_NO_VALID` (4) for exactly this reason. Consider a dedicated `EXIT_TERMS_NOT_ACCEPTED` and reserving 1 for real errors.

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `54d634c`

## PR #2844: feat(cli): accept the Terms of Use from the CLI
**Author:** PetrBulanek | **Branch:** feat/2755-cli-terms-accept → main | **Changes:** +394 −3 (9 files)

### Summary
Adds a `dam terms` command group (`show` / `status` / `accept`) so a headless/CI user who signs in via the device flow can review and accept the Terms of Use entirely from the CLI, and rewrites the 412-gate hint to point at `dam terms accept` instead of "open the browser". Closes #2755.

### Findings
- 🟡 **Warning:** `dam terms status` exits with `EXIT_RUNTIME_FAILURE` (1) for the *expected* "not accepted" outcome (`status.ts:51`), the same code it returns for genuine failures — server unreachable, host resolution failure (`status.ts:42,46`). A CI gate can't distinguish "terms need accepting" (actionable) from "the call broke". The sibling `auth status` command introduced `EXIT_AUTH_STATUS_NO_VALID` (4) precisely so its not-OK state is a distinct, documented code (`auth/commands/status.ts:34`). Consider a dedicated `EXIT_TERMS_NOT_ACCEPTED` and reserving 1 for real errors.
- ✅ **Looks good:** `accept.ts` refuses a non-interactive accept before any network work when `--yes` is absent on a non-TTY, uses a generous 15-min confirm timeout so a careful reader isn't auto-declined, and handles the version-changed-mid-review race by surfacing `PRECONDITION_FAILED` distinctly rather than silently accepting the new text.
- ✅ **Looks good:** `--expect-version` guards against accepting a version other than the one reviewed, and the naming rationale (not `--version`, reserved for the CLI's own) is documented on the option.
- ✅ **Looks good:** `terms-service.ts` correctly splits transport: metadata/acceptance over tRPC (gate-exempt) and the document text over the plain `GET /api/terms`, with `termsDocumentSchema.safeParse` validating the untyped HTTP body and mapping every failure mode to a typed `TransportError`.
- 🟢 **Suggestion:** `show.ts` prints the acceptance timestamp via `new Date(...).toISOString()`, i.e. always UTC. For a human-facing `show`, a locale-formatted time might read better; minor.

✅ No documentation drift detected. `docs/architecture/cli.md` — the subsystem page owning CLI command groups — was correctly updated: `Last verified:` bumped to 2026-07-20, a `terms` bullet added to the group list, and the 412-gate remedy line rewritten to point at `dam terms accept`. `security-and-credentials.md` describes the terms gate only conceptually (binding-owner acceptance) and isn't changed by this PR, so it needs no update.

✅ No findings. The new `terms` module mirrors the established CLI-package conventions: `index.ts` exposes only `composeTermsModule` + its `TermsModule` type (same shape as `metrics/index.ts`), `compose.ts` is the composition root wiring the tRPC/HTTP adapter into the service, and command builders depend on the `TermsService` interface — clean commands → service → adapter layering with no boundary violations introduced.

### Verdict
COMMENT — Solid, well-tested feature with clean layering; the only substantive point is the `status` exit-code collision with real failures, which weakens its stated use as a CI gate.

---

kapetr (approved):

Issue #2759, opened 2026-07-10
Epic - Redesign agent config


jenna-winkler:
For simplicity, we are bundling all of these related issues under a single agent config epic.

Figma: TBD

Issue #2812, opened 2026-07-16
UI - Migrate the connection create form to React Hook Form + Zod


kapetr:
## Problem

The connection create form (used in the Connection catalogue for every template kind) manages its inputs with hand-rolled controlled state: field values live in a plain record, and validation happens imperatively at submit time. The project's UI conventions call for React Hook Form + Zod once a form has three or more fields, which this form clears for most templates. The current shape works but misses per-field validation feedback, dirty tracking, and a single schema-derived type for the form values.

## Proposed solution

Migrate the form to React Hook Form with a Zod schema per auth kind, keeping the existing behavior (preset values, the Customize defaults toggle, popup OAuth submission) unchanged. Deferred out of #2811 to keep that PR reviewable.

kapetr:
Partial progress: the **MCP create pane** (`McpCreatePane` / `catalog-mcp-create-pane.tsx`) is now migrated to React Hook Form + Zod as part of #2846 (commit `2a099275`).

- Zod schema owns URL validation and both both-or-neither rules (OAuth id/secret, header name/value), with each error attached to the empty field of the pair.
- Validation is now inline per-field instead of toasts; the submit button disables while the form is invalid.
- `LabeledInput` gained optional `error`/`onBlur` props to carry field errors — reusable by the remaining forms.

Still to migrate under this issue: the sibling `template-create-form-body.tsx` (and any other `useState` connection forms). Leaving #2812 open for those.

PR #2846, opened 2026-07-20
feat: single entry point for adding an MCP server

### LINKED ISSUE #423: UI - Simplify MCP server connection to one entry point
## Problem

The connections UI presents MCP servers as two separate options: "Custom MCP server (OAuth)" and "Custom MCP server (no auth)." This forces the user to understand and choose an authentication mechanism before they've done anything - but at the point of adding a server, the user doesn't know or care whether it's OAuth or no-auth. They just have a server they want to connect.

The auth distinction is an implementation detail the platform can determine on its own. Surfacing it as the first decision adds friction and confusion to what should be a single, obvious action: "add an MCP server."

## Impact

- Users hesitate or pick wrong, because the choice depends on knowledge they don't have.
- Two buttons for one conceptual action makes the connections UI look more complicated than it is.

## Desired outcome

A single entry point for adding an MCP server. The user provides the server; the platform handles the auth distinction behind the scenes. Any less-common configuration (custom headers, environment values some servers require) lives as secondary/advanced options, not as upfront choices.

## Screenshot

Image

kapetr:
Adding an MCP server is now one action instead of an upfront auth quiz.

- The MCP entry in the Connection catalogue opens a single **Add an MCP server** pane — name and URL, with the platform detecting whether the server needs OAuth and running the right flow. No more choosing "OAuth" vs "no-auth" before you've done anything.
- An **Advanced configuration** section covers the less-common cases: a pre-registered OAuth client (for providers that don't allow dynamic client registration), or a header credential (API key) injected at the gateway and never written into the agent's config.
- The submit button reflects what will happen — Create, or Create + Authorize — and shows a brief checking state while the URL is probed.

Part of #2759.

Closes #423

### DIFF
--- a/packages/api-server/src/modules/connections/domain/build-connection.ts
+++ b/packages/api-server/src/modules/connections/domain/build-connection.ts
@@ -501,7 +501,11 @@ function buildNone(
 
   if (input.url) {
     const url = new URL(input.url);
-    contributions.push({ kind: "egress-allow", host: url.host });
+    contributions.push({
+      kind: "egress-allow",
+      host: url.hostname,
+      ...(url.port ? { port: Number(url.port) } : {}),
+    });
     contributions.push({
       kind: "mcp-entry",
       name: template.id,
@@ -513,8 +517,6 @@ function buildNone(
     if (input.headerName && input.value) {
       const headerName = input.headerName;
       const valueFormat = "{value}";
-      // egress-inject takes hostname + numeric port separately, whereas the
-      // egress-allow above matches the combined host:port string.
       contributions.push({
         kind: "egress-inject",
         host: url.hostname,
--- a/packages/api-server/src/modules/connections/domain/build-connection.ts
+++ b/packages/api-server/src/modules/connections/domain/build-connection.ts
@@ -70,6 +70,7 @@ export async function buildConnection(
       return buildNone(
         template as Extract<ConnectionTemplate, { authKind: "none" }>,
         input,
+        mintSecretRef,
       );
   }
 }
@@ -494,6 +495,7 @@ function buildHeader(
 function buildNone(
   template: Extract<ConnectionTemplate, { authKind: "none" }>,
   input: Extract<ConnectionCreateInput, { authKind: "none" }>,
+  mintSecretRef: (purpose: string) => SecretRef,
 ): BuildResult {
   const contributions: Contribution[] = [...template.contributions];
 
@@ -505,6 +507,34 @@ function buildNone(
       name: template.id,
       url: input.url,
     });
+
+    // Optional header credential: injected at the gateway (egress-inject +
+    // header auth), never written into the mcp-entry the harness sees.
+    if (input.headerName && input.value) {
+      const headerName = input.headerName;
+      const valueFormat = "{value}";
+      contributions.push({
+        kind: "egress-inject",
+        host: url.hostname,
+        ...(url.port ? { port: Number(url.port) } : {}),
+        headerName,
+        valueFormat,
+      });
+      const secretPath = mintSecretRef(`connection:${template.id}`);
+      const sdsFields = buildConnectionSdsFields(contributions, input.value);
+      return {
+        auth: {
+          kind: "header",
+          valueRef: { ...secretPath, field: "value" },
+          headerName,
+          valueFormat,
+        },
+        contributions,
+        secrets: new Map([
+          [secretPath.path, { value: input.value, ...sdsFields }],
+        ]),
+      };
+    }
   }
 
   return {
--- a/packages/api-server/src/modules/connections/domain/build-connection.ts
+++ b/packages/api-server/src/modules/connections/domain/build-connection.ts
@@ -513,6 +513,8 @@ function buildNone(
     if (input.headerName && input.value) {
       const headerName = input.headerName;
       const valueFormat = "{value}";
+      // egress-inject takes hostname + numeric port separately, whereas the
+      // egress-allow above matches the combined host:port string.
       contributions.push({
         kind: "egress-inject",
         host: url.hostname,
--- a/packages/api-server/src/modules/connections/domain/build-connection.ts
+++ b/packages/api-server/src/modules/connections/domain/build-connection.ts
@@ -223,21 +223,32 @@ async function buildOAuthDcr(
   if (!meta) {
     throw new Error(`No OAuth discovery metadata at ${input.url}`);
   }
-  if (!meta.registrationEndpoint) {
-    throw new Error(
-      `MCP server at ${input.url} does not support dynamic client registration`,
-    );
-  }
 
-  const dcr = await registerOAuthClient({
-    registrationEndpoint: meta.registrationEndpoint,
-    clientName: `${brandName} Agent Platform`,
-    redirectUris: [oauthCallbackUrl],
-  });
+  // A supplied client is pre-registered with the server — endpoints come
+  // from discovery, registration is skipped.
+  let clientId: string;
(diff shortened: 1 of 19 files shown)


packages/ui/src/modules/connections/components/catalog-mcp-create-pane.tsx
```diff
@@ -0,0 +1,221 @@
...
+  oauthReturnView,
+  onBack,
+  onCreated,
+}: Props) {
+  const [name, setName] = useState(() => slugifyTemplateName("MCP server"));
+  const [url, setUrl] = useState("");
+  const [clientId, setClientId] = useState("");
+  const [clientSecret, setClientSecret] = useState("");
+  const [headerName, setHeaderName] = useState("");
+  const [headerValue, setHeaderValue] = useState("");
```
dam-code-guardian: 🟡 **Warning:** This pane manages 6 form fields in `useState` (name, url, clientId, clientSecret, headerName, headerValue) plus two cross-field "both-or-neither" checks in `onSubmit`. That clears the project's RHF+Zod threshold for new code (≥3 fields **and** cross-field validation). Noted as non-blocking since the PR body explicitly defers this and the sibling connection forms to #2812 with a rationale — a tracked deviation, not silent drift.

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `eb02073`

## PR #2846: feat: single entry point for adding an MCP server
**Author:** kapetr | **Branch:** feat/423-mcp-single-entry → main | **Changes:** +545 −82 (13 files)

### Summary
Replaces the upfront OAuth-vs-no-auth chooser for MCP servers with one guided "Add an MCP server" pane: the platform probes the URL and detects the auth kind, the submit button reflects Create vs Create + Authorize, and an Advanced section allows a pre-registered OAuth client or a gateway-injected header credential. Server-side, `buildNone` gains optional header-credential support and `buildOAuthDcr` accepts a pre-registered client. Closes #423.

### Findings
- 🟡 **Warning:** `McpCreatePane` manages 6 form fields via `useState` plus two cross-field "both-or-neither" validations (`catalog-mcp-create-pane.tsx`), which clears the project's RHF+Zod threshold (≥3 fields **and** cross-field validation) for new code. The author has explicitly deferred this and the sibling connection forms to #2812 with a written rationale — an acknowledged, tracked deviation rather than silent drift, so noting it, not blocking.
- ✅ **Looks good:** `useMcpAuthDetection` guards against out-of-order probe results with `latestUrlRef` and cleans up its debounce timer on URL change / unmount — no stale-detection or set-state-after-unmount races.
- ✅ **Looks good:** `useDiscoverMcp({ silent })` routes background probe failures through the centralized `suppressErrorToast` meta key (`query-client.ts:60`) rather than popping a global toast on every keystroke.
- ✅ **Looks good:** The `PaneHeader` was extracted to a shared `CatalogPaneHeader` and the old inline copy deleted (no duplication); both `ChoosePane`/`CreatePane` now consume the shared one.
- ✅ **Looks good:** The header-credential secret never reaches the harness — `buildNone` synthesizes `egress-inject` + gateway header auth and mints the secret, and the new unit test asserts the raw value appears nowhere in `contributions` and the `mcp-entry` has no `headers`.
- 🟢 **Suggestion:** In `buildNone`, `egress-allow` uses `url.host` (host+port) while the new `egress-inject` uses `url.hostname` + separate numeric `port` (`build-connection.ts`). This matches the existing header/kubernetes convention so it's correct, but the two adjacent host representations are easy to misread; a one-word comment or a shared helper would help future readers.

⚠️ **Minor drift** — `docs/architecture/connections.md`. The "Custom MCP server" section (line 145) states an MCP endpoint "contributes just two things: an `egress-allow` and an `mcp-entry`." This PR adds a new architectural path: a no-auth MCP server given a header credential now *also* synthesizes an `egress-inject` + gateway header auth (`build-connection.ts` `buildNone`). The page wasn't updated. Suggested fix: extend the Custom MCP server section to cover the optional header-credential path (composing MCP with the documented `egress-inject`/header pattern) and bump `Last verified:` to 2026-07-20. The pre-registered-OAuth-client change is below architecture level (how the credential is obtained, not what is contributed) and doesn't independently trigger drift.

✅ No findings. Domain purity is preserved (`build-connection.ts` imports only types, sibling domain modules, and infrastructure ports; `mintSecretRef` is injected as a function port). The new `buildNone` header branch mirrors the established `buildHeader` shape, the discriminated-union `noneCreateInput` keeps its `authKind` discriminant with the new fields optional, and validation stays in the contract package. No layering or boundary violations introduced.

🟡 See the deferred-form-migration finding above (RHF+Zod threshold). Otherwise aligned: no static `style={{}}`, no `any`/`as`, no raw `fetch` in components, clean shared-header extraction, correct debounced-probe hook with race guards, centralized mutation error handling.

### Verdict
COMMENT — Well-built feature with strong secret-handling test coverage; two non-blocking notes (a tracked-deferral form-state migration and a small doc-drift on connections.md's Custom MCP section).

---

packages/ui/src/modules/connections/components/catalog-mcp-create-pane.tsx
```diff
@@ -0,0 +1,221 @@
...
+/** Single entry point for adding an MCP server (#423): Name + URL, with auth
+ *  detected from the URL. The Advanced disclosure lets a power user pin a
+ *  pre-registered OAuth client or a header credential, overriding detection. */
+export function McpCreatePane({
+  templateById,
+  oauthReturnView,
+  onBack,
+  onCreated,
+}: Props) {
+  const [name, setName] = useState(() => slugifyTemplateName("MCP server"));
```
PetrBulanek: Default name is hard-coded to `mcp-server` on every open, but connections have a UNIQUE (owner, name) constraint so adding a second MCP server without renaming fails with a generic create error. I am thinking about deriving the default from the URL host or we might use the similar random name generator we have for sandboxes.

packages/ui/src/modules/connections/components/catalog-mcp-create-pane.tsx
```diff
@@ -0,0 +1,221 @@
...
+    const urlError = validateMcpUrl(url);
+    if (urlError) {
+      emitToast({ kind: "error", message: urlError });
+      return;
+    }
+    // Both-or-neither: a half-filled pair would otherwise be silently dropped,
+    // creating an auth-less connection from credentials the user typed.
+    const bothOrNeither = (a: string, b: string) =>
+      (a.trim() === "") === (b.trim() === "");
+    if (!bothOrNeither(clientId, clientSecret)) {
```
PetrBulanek: `bothOrNeither(clientId, clientSecret)` forces a secret whenever an OAuth ID is given, but `buildOAuthDcr` already supports a pre-registered client with no secret (public/PKCE client). Suggest relaxing to "a secret requires an id" (id-without-secret allowed) so the UI isn't stricter than the backend.

packages/api-server/src/modules/connections/domain/build-connection.ts
PetrBulanek: `egress-allow` uses `url.host`, which glues the port into the host field (e.g. `keyed.example.com:8443`) instead of `url.hostname` + a separate `port` — I confirmed this in a created connection's stored contributions. For the header path the added `egress-inject` masks it, but a plain no-auth MCP server on a non-443 port gets only this malformed allow rule. Since `buildNone` is already being touched, suggest splitting host/port here to match the `egress-inject` below.

PetrBulanek (changes requested):

PetrBulanek (approved):

PR #2847, opened 2026-07-20
fix(skills): newly-added source is publishable without a reload

### LINKED ISSUE #2837: A newly added skill source isn't offered as a publish target until the page is reloaded
## Problem

After a user adds a GitHub skill source, the new source appears immediately in the skills list — but it does **not** appear in the "Publish to" dropdown when publishing a locally-created skill. The dropdown keeps showing only the sources that existed when the page was loaded. The user has to reload the page before they can publish to the source they just added.

This is confusing: the source is visibly present on the page, yet it's not selectable as a publish destination, with nothing to explain the discrepancy.

## Proposed solution

Keep the publish target list in sync with the sources shown on the page, so a source added during the session is immediately available to publish to — no reload required.

### Dependencies

Found while verifying #944.

PetrBulanek:
## What & why

After adding a GitHub skill source, it showed up in the skills list right away but was **not** offered in the "Publish to" dropdown when publishing a locally-created skill — the dropdown only listed sources that existed at page load. Users had to reload the page before they could publish to a source they'd just added, with nothing on screen explaining the gap.

Now a source added during the session is immediately available as a publish target. No reload required.

Closes #2837

### DIFF
--- a/packages/api-server/src/modules/skills/services/skills-service.ts
+++ b/packages/api-server/src/modules/skills/services/skills-service.ts
@@ -263,7 +263,9 @@ export function createSkillsService(deps: SkillsServiceDeps): SkillsService {
     },
     async createSource(input: SkillCreateSourceInput) {
       try {
-        return await deps.repo.create(input, deps.owner);
+        const created = await deps.repo.create(input, deps.owner);
+        const [enriched] = enrichSources([created]);
+        return enriched;
       } catch (err) {
         if (isUniqueViolation(err, "skill_sources_owner_git_url_idx")) {
           throw new TRPCError({


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `6ce9952`

## PR #2847: fix(skills): newly-added source is publishable without a reload
**Author:** PetrBulanek | **Branch:** feat/2837-publish-target-sync → main | **Changes:** +3 −1 (1 file)

### Summary
`createSource()` in the skills service now enriches the freshly-created source with the `canPublish` flag (via `enrichSources`) before returning it, matching `getSource()` and `listSources()`. This fixes #2837: a source added during the session was appended to the UI's `sources` state without `canPublish`, so it never appeared in the "Publish to" dropdown until a page reload.

### Findings
- ✅ **Looks good:** The fix is minimal and consistent — the same `enrichSources([x])[0]` idiom already used by `getSource` (`skills-service.ts:261`). The returned shape still satisfies the `Promise<SkillSource>` contract (`api-server-api/.../types.ts:40`), and the UI consumer appends the result directly to state (`use-skills-surface.ts:255`), so `canPublish` now flows through without a reload.
- ✅ **Looks good:** `canPublish` remains a soft signal (GitHub-host detection only, no auth preflight) as documented in the `enrichSources` comment — no new cluster/network call on the create path.

✅ No documentation drift detected. `docs/architecture/skills.md:211` already documents that `canPublish: true` is returned for GitHub URLs; this change aligns `createSource` with that documented behavior rather than altering subsystem responsibility. No architecture page was edited, so no `Last verified:` bump is required.

✅ No findings. The change lives in the application layer (service factory) and respects slice-composition/module-boundary rules; it mirrors sibling methods and introduces no new coupling or contract change.

### Verdict
APPROVE — Correct, minimal, well-scoped fix that closes the publish-target-visibility gap consistently with existing methods.

---

PR #2852, opened 2026-07-20
fix(skills): clear "grant access" message for failed private-source scans

### LINKED ISSUE #2836: Failed private skill-source scan shows a raw error instead of a clear "grant access" message
## Problem

When a user adds a private GitHub repository as a skill source but the sandbox's GitHub connection doesn't yet have access to that repository, the scan fails — and the source card shows a raw internal error string (e.g. `agent-runtime scan …: fetch failed`) instead of telling the user what's actually wrong.

The user is left with a cryptic message and no idea that the fix is simply to grant the repository to their GitHub connection. Once access is granted and the source is re-scanned, it works — so the only thing missing is a clear explanation at the moment of failure.

## Proposed solution

When a private-source scan fails because the connection can't reach the repository, show a friendly, actionable message — e.g. "Can't access this repository. Make sure your GitHub connection has been granted access, then re-scan" — with the connect/grant affordance, instead of the raw error text.

### Dependencies

Found while verifying #944.

PetrBulanek:
When you add a private GitHub repo as a skill source but your GitHub connection hasn't been granted access to it, the scan used to fail with a cryptic internal string (`fetch failed` / `GitHub: Not Found`) — leaving you with no idea that the fix is simply to grant the repo.

Now the source shows a clear, actionable message instead:

> Can't access this repository. If it's private, grant your GitHub connection access to it, then re-scan — otherwise, double-check the repo URL.

In the UI, the errored source card pairs this with a **Manage connections** button that takes you straight to the sandbox's Connections tab. The CLI prints the same message and exits with a distinct code so scripts can tell "needs a connection" apart from other failures.

Once access is granted and the source is re-scanned, skills load as normal.

Closes #2836

### DIFF
--- a/packages/api-server/src/modules/skills/services/skills-service.ts
+++ b/packages/api-server/src/modules/skills/services/skills-service.ts
@@ -50,6 +50,28 @@ export function templateSourceId(templateId: string, gitUrl: string): string {
 
 export const TEMPLATE_SOURCE_ID_PREFIX = "template:";
 
+/** Both real private-scan failure shapes — a bare GitHub 404 (repo not in the
+ *  connection's grant, or a typo) and a transport-level throw (egress rejected
+ *  at the gateway, no HTTP response) — mean the same thing to the user: the
+ *  connection can't reach the repo. Hedged because a 404 can't distinguish
+ *  "private, ungranted" from "doesn't exist". */
+const SCAN_ACCESS_MESSAGE =
+  "Can't access this repository. If it's private, grant your GitHub connection access to it, " +
+  "then re-scan — otherwise, double-check the repo URL.";
+
+function scanAccessError(): TRPCError {
+  return new TRPCError({ code: "FORBIDDEN", message: SCAN_ACCESS_MESSAGE });
+}
+
+/** A private-scan failure that arrived as a plain Error (not a structured
+ *  upstream envelope) is only the "can't reach the repo" case when it carries a
+ *  transport signature. Anything else rethrows raw so genuine bugs stay
+ *  visible rather than being masked by the friendly message. */
+function isTransportError(err: unknown): boolean {
+  if (!(err instanceof Error)) return false;
+  return /fetch failed|econn|socket|network/i.test(err.message);
+}
+
 export interface SkillsServiceDeps {
   repo: SkillsRepository;
   agentSkillsRepo: AgentSkillsRepository;
@@ -354,7 +376,16 @@ export function createSkillsService(deps: SkillsServiceDeps): SkillsService {
           deps.runtimeClient.scan(agentId, gitUrl, src.path),
         );
       } catch (err) {
-        if (err instanceof AgentRuntimeUpstreamError) throw upstreamToTrpc(err);
+        // A bare 404 means egress reached GitHub but the repo isn't in the
+        // connection's grant (or doesn't exist) — a "grant access" case, not
+        // the raw "GitHub: Not Found" upstreamToTrpc would produce. Every
+        // other upstream status (403 scope message, 5xx, structured body)
+        // keeps its existing translation.
+        if (err instanceof AgentRuntimeUpstreamError) {
+          if (err.upstream.status === 404) throw scanAccessError();
+          throw upstreamToTrpc(err);
+        }
+        if (isTransportError(err)) throw scanAccessError();
         throw err;
       }
     },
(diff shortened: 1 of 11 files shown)


packages/api-server/src/modules/skills/services/skills-service.ts
```diff
@@ -354,7 +376,16 @@ export function createSkillsService(deps: SkillsServiceDeps): SkillsService {
           deps.runtimeClient.scan(agentId, gitUrl, src.path),
         );
       } catch (err) {
```
dam-code-guardian, line 353: 🟢 **Suggestion:** These new classification branches (404 ⇒ `scanAccessError`, `isTransportError` ⇒ `scanAccessError`, everything else passthrough) are untested — the skills module has no service tests. A small unit test pinning "404 ⇒ SCAN_ACCESS_MESSAGE, non-404 upstream ⇒ upstreamToTrpc passthrough, transport error ⇒ SCAN_ACCESS_MESSAGE" would lock the intent against future refactors.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `9b28c75`

## PR #2852: fix(skills): clear "grant access" message for failed private-source scans
**Author:** PetrBulanek | **Branch:** fix/2836-private-scan-grant-access-message → main | **Changes:** +76 −5 (4 files)

### Summary
Translates the two real private-scan failure shapes — a bare GitHub 404 and a transport-level throw — into one clear, actionable `FORBIDDEN` message ("grant your GitHub connection access… then re-scan") across api-server, CLI, and UI, replacing the cryptic `fetch failed` / `GitHub: Not Found` leak. The UI pairs the message with a client-built **Manage connections** button when there's no server CTA. Closes #2836.

### Findings
- ✅ **Looks good:** The 404→`scanAccessError()` branch is correctly scoped — every other upstream status (403 scope message, 5xx, structured body) still routes through `upstreamToTrpc`, so genuine bugs aren't masked (`packages/api-server/.../skills-service.ts:384`).
- ✅ **Looks good:** `isTransportError` gates the friendly message behind a transport signature and rethrows anything else raw, keeping non-transport errors visible (`skills-service.ts:66`).
- ✅ **Looks good:** CLI `FORBIDDEN` branch reuses the existing `source-needs-connection` kind (cta optional, consumers guard `if (e.cta)`), so exit-code/output contract is consistent with the pre-existing 403 path.
- ✅ **Looks good:** UI `Manage connections` affordance is correctly gated on `agentId`, uses `type="button"`, and only renders when there's no server CTA; `"connections"` is a valid `SandboxSection`.
- 🟢 **Suggestion:** The new classification branches (404→FORBIDDEN, `isTransportError` regex) are untested — the skills module currently has no service tests. A small unit test pinning "404 ⇒ SCAN_ACCESS_MESSAGE, 500 ⇒ passthrough" would lock the intent against future refactors (`packages/api-server/src/modules/skills/services/skills-service.ts:378`).

✅ No documentation drift detected. No `docs/architecture/` page was edited; the behavior change is an error-message/UX refinement below the architecture altitude. `skills.md` already documents scan dispatch and CTA-rendering at the correct level (exact error strings are volatile content and correctly kept out of the docs).

✅ No findings. The new `TRPCError`-throwing helpers (`scanAccessError`, `SCAN_ACCESS_MESSAGE`, `isTransportError`) and the CLI `FORBIDDEN` classification branch follow the established application-layer error-handling patterns already pervasive in both service files; no new architecture-doc violations introduced.

✅ No findings. The optional `onManageConnections` prop, `type="button"` handler, and `navigateToSandboxHome` store read all match existing file conventions (the inline `useStore` selector mirrors the pre-existing `showConfirm` line); no static inline styles, correct `agentId` gating, files well under size triggers.

### Verdict
APPROVE — a focused, well-reasoned error-UX fix; correctly narrows the friendly message to genuine access failures and preserves raw errors for everything else.

---

packages/api-server/src/modules/skills/services/skills-service.ts
```diff
@@ -50,6 +50,28 @@ export function templateSourceId(templateId: string, gitUrl: string): string {
...
+}
+
+/** A private-scan failure that arrived as a plain Error (not a structured
+ *  upstream envelope) is only the "can't reach the repo" case when it carries a
+ *  transport signature. Anything else rethrows raw so genuine bugs stay
+ *  visible rather than being masked by the friendly message. */
+function isTransportError(err: unknown): boolean {
+  if (!(err instanceof Error)) return false;
+  return /fetch failed|econn|socket|network/i.test(err.message);
+}
```
kapetr: Claude says the code can't deliver the intent in the comment:

> This regex can't tell the two hops apart: `agent-runtime-client` wraps *api-server → pod* connection failures as plain `Error("scan: fetch failed")` too. So a pod restarting mid-scan (or the #283 mesh-cert resets) matches here and tells the user to grant GitHub access — but GitHub was never involved, and granting won't fix it. Could agent-runtime return a structured envelope for the egress-blocked case instead of us sniffing message strings?

kapetr (approved):

PR #2861, opened 2026-07-21
fix(cluster): skip k3s-selinux RPM when SELinux is disabled

JanPokorny:
## Problem

`mise run cluster:install` aborts at k3s host-install on a host where **SELinux is present but Disabled at runtime**: the k3s installer relabels its binary when it detects SELinux tooling, that step fails, and under `set -e` it takes down provisioning. (Reported from a real run; manual workaround was `INSTALL_K3S_SKIP_SELINUX_RPM=true`.)

## Fix

Both lima templates (`deploy/lima-k3s.yaml`, `deploy/lima-k3s-test.yaml`) now check SELinux state before running the installer and `export INSTALL_K3S_SKIP_SELINUX_RPM=true` when it isn't active:

```sh
if ! { command -v getenforce >/dev/null 2>&1 && [ "$(getenforce)" != Disabled ]; }; then
  export INSTALL_K3S_SKIP_SELINUX_RPM=true
fi
curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION="v1.36.1+k3s1" INSTALL_K3S_EXEC="server …" sh -
```

- **Disabled, or `getenforce` absent** → skip the k3s-selinux RPM (avoids the failing relabel).
- **Enforcing / Permissive** → unchanged; the policy installs and labels apply as before.

Same pinned k3s version and listen-port args. Scoped to DAM's own k3s provisioning — not the dam-vm VPS/VM.

## Testing

Verified the detection across all four cases (absent / Disabled → skip; Enforcing / Permissive → keep) and that both templates still parse.

### DIFF
--- a/deploy/lima-k3s-test.yaml
+++ b/deploy/lima-k3s-test.yaml
@@ -41,6 +41,11 @@ provision:
       if [ ! -d /var/lib/rancher/k3s ]; then
         # Pinned: v1.36.2+k3s1 bumped bundled Traefik to chart v40.x, whose validate-install-crd
         # rejects our pre-applied Gateway API v1.2.0/standard CRDs. v1.36.1 ships Traefik v3.6.13.
+        # Skip the k3s-selinux RPM when SELinux isn't active (absent or Disabled):
+        # its binary relabel step fails under set -e on a disabled-SELinux host.
+        if ! { command -v getenforce >/dev/null 2>&1 && [ "$(getenforce)" != Disabled ]; }; then
+          export INSTALL_K3S_SKIP_SELINUX_RPM=true
+        fi
         curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION="v1.36.1+k3s1" INSTALL_K3S_EXEC="server --write-kubeconfig-mode 644 --https-listen-port 16445" sh -
       fi
       mkdir -p /var/lib/rancher/k3s/server/manifests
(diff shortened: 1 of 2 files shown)


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `4c4910e`

## PR #2861: fix(cluster): skip k3s-selinux RPM when SELinux is disabled
**Author:** JanPokorny | **Branch:** fix/k3s-selinux-disabled → main | **Changes:** 2 files (deploy/lima-k3s.yaml, deploy/lima-k3s-test.yaml)

### Summary
Before running the k3s host installer in both lima templates, exports `INSTALL_K3S_SKIP_SELINUX_RPM=true` when SELinux isn't active (absent or `Disabled`), avoiding the installer's binary-relabel step that fails under `set -e` on a disabled-SELinux host. Enforcing/Permissive hosts are unchanged.

### Findings
- ✅ **Looks good:** The negated guard is correct across all four cases: `getenforce` absent → `command -v` fails → skip; returns `Disabled` → `[ Disabled != Disabled ]` false → skip; `Enforcing`/`Permissive` → inner condition true → RPM kept. Matches the documented intent exactly.
- ✅ **Looks good:** `command -v getenforce >/dev/null 2>&1` guards the `getenforce` call, so the `$(getenforce)` subshell only runs when the binary exists — no stderr noise or spurious failure when SELinux tooling is absent.
- ✅ **Looks good:** `export` scopes the var to the piped installer invocation only; the pinned `INSTALL_K3S_VERSION` and per-template listen-port args (16445 / 16443) are untouched, so the fix is additive.
- ✅ **Looks good:** Both templates get the identical guard, keeping test and non-test provisioning in sync; the explanatory comment states the *why* (relabel fails under `set -e`).
- 🟢 **Suggestion:** The same guard block is now duplicated verbatim in both lima templates. Not actionable in YAML inline-script blocks without a shared provisioning script, but worth noting if a third template ever appears — extract to a sourced snippet then.

✅ No documentation drift detected. The k3s installer's SELinux-RPM handling is volatile install detail that correctly stays out of `docs/architecture/`; the VM and SVID-rotation pages describe behavior this shell tweak doesn't change.

### Verdict
APPROVE — a correct, minimal, well-commented provisioning fix; the shell detection is sound in every SELinux state and scoped tightly to the failing step.

---

PR #2863, opened 2026-07-21
fix(api-server): report JWKS outages as 503, gate readiness on JWKS warm-up

pilartomas:
## Problem

On every `platform-apiserver` rollout, live UI tRPC calls (`budgets.reserved`, `approvals.listForOwner`, `agents.list`, `channels.available`) received spurious **401 Unauthorized** for ~1–5 min after each new pod started — **187 occurrences over 48h across 4 deploy-correlated bursts, zero between deploys**. A user active during a deploy saw auth errors (and, if their token happened to expire in that window, a forced re-login).

**Root cause:** jose's `createRemoteJWKSet` fetches the Keycloak JWKS lazily on the *first* `jwtVerify`. A freshly-rolled pod's in-mesh (ambient/ztunnel) egress to Keycloak is briefly unreachable, so Node's `fetch` throws `TypeError: fetch failed`. The auth middleware caught that and logged `authn.deny` with `reason: err.name` (`"TypeError"`) → **401** — a transient infrastructure failure reported as a credential rejection, indistinguishable from a bad token. The pod also went `Ready` ~5s after boot on a static `/api/health` probe that checks nothing, so it took 100% of traffic with a cold JWKS.

## Changes

**1. Classify transient JWKS failures as 503, not 401 (fails closed).**
A new `AuthUnavailableError` sentinel is thrown from `verifyJwt` when the key set can't be retrieved/ingested — undici network `TypeError`, `JWKSTimeout`, non-200/non-JSON response (`ERR_JOSE_GENERIC`), or a malformed fetched key set (`JWKSInvalid`). Both the HTTP middleware (→ `503` + `authn.unavailable`) and the WS upgrade handler (→ `503` + `ws.authn_unavailable`) handle it. Every token-validity failure (expired, bad signature, unknown `kid`, wrong audience, API-key denial) stays **401**. The UI already treats 502/503 as a reconnect (suppresses toasts, shows the "Reconnecting…" banner), so 503 degrades gracefully where 401 spammed errors.

**2. Warm the JWKS at boot + real readiness gate.**
`createAuth` exposes `warmJwks = () => JWKS.reload()`; a background loop (`jwks-warmup.ts`) polls it with backoff and a bounded **240s give-up**, then latches ready. A new `/api/ready` returns 503 until latched, and the helm `readinessProbe` points at it (liveness stays on the static `/api/health`). With `maxUnavailable: 0` the old pod keeps serving until the new pod can actually verify tokens; the give-up latch means a prolonged Keycloak outage degrades to 503s instead of wedging the rollout indefinitely.

## Security note (adversarial review finding)

The `TypeError` arm is gated on `.cause`. jose can throw a bare `TypeError('non-ASCII string encountered in encode()')` while building the signing input from an **attacker-supplied payload segment — before the signature is checked**. Without the gate, a forged token could be relabelled `503`/`authn.unavailable` (deny-only, never a bypass — but it would move forged-token probes out of the deny audit/alert stream). undici's fetch failure always carries a `.cause`; jose's internal TypeErrors are message-only — so gating on `.cause != null` keeps every real fetch failure a 503 while every token/key-material TypeError stays a 401. This is more robust than message-matching, which drifts across Node/undici versions.

## ⚠️ Monitoring migration required

Rollout bursts previously showed up in HyperDX as `Body='authn.deny'` with `reason='TypeError'`. They now emit **`authn.unavailable`** (HTTP) / **`ws.authn_unavailable`** (WS) with `reason='jwks-unavailable'` and no `decision` field (no credential verdict was reached). Add a query/alert for the new event names; the old `authn.deny` deny-stream no longer fires on rollout hiccups (that's the point — infra blips stop polluting credential-denial alerting).

## Ops

`/api/ready` now genuinely gates apiserver readiness, so rollout waits that transitively depend on it were bumped past the 240s give-up: `cluster:build-apiserver` `60s → 300s`, `e2e:reset` Available wait `120s → 300s`. The e2e-install `rollout status --timeout=5m` already clears the ~265s worst case (unchanged). The telemetry probe-span filter now also ignores `/api/ready`.

## Tests

`auth-audit.test.ts` extended (15 cases): TypeError-with-cause / JWKSTimeout / ERR_JOSE_GENERIC / JWKSInvalid → 503 `authn.unavailable`; **causeless (attacker-forgeable) TypeError → 401**; JWTExpired / JWKSNoMatchingKey / API-key denial stay 401; `warmJwks` delegates to `reload()`. New `jwks-warmup.test.ts` (3 cases): latch on success, stay-not-ready-then-latch, latch-on-give-up + error log. Full suite: 478 api-server tests pass; `tsc`/eslint/prettier clean; `helm lint`/`render` (kubeconform) pass.

`/api/ready` semantics: a **200 does not guarantee auth is working** after a give-up — it means the startup gate is past. That's intentional (single replica must not drop out of every Service endpoint).

## Verification (manual, not yet run on a live cluster)

1. `cluster:build-apiserver`, scale Keycloak to 0 → authed tRPC returns 503 `{"error":"auth unavailable"}`, logs show `authn.unavailable reason=jwks-unavailable` (not `authn.deny`); `/api/ready` returns 503 on a fresh pod.
2. Scale Keycloak up → warm loop latches, `/api/ready` → 200, auth works.
3. Rollout while the UI polls → old pod serves until the new pod is Ready; no 401 burst; worst case brief 503s → "Reconnecting…" with no error toasts.

## Follow-ups (not in this PR)

- **WS 503 branch has no automated test** — the upgrade handler is an inline closure needing full app deps; the shared classifier is covered via the HTTP path. Extracting a testable `handleRelayUpgrade(deps)` is a reasonable follow-up (matches the pre-existing gap: no WS-upgrade auth tests exist today).
- **UI connection-status flap during a sustained JWKS outage** — 503s flip the health tracker to "reconnecting", but the recovery poll probes the public `/api/health` which recovers before auth does, so status oscillates. The clean fix (gate recovery on an authenticated success, not `/api/health`) is a recovery-probe redesign; the flap self-heals and is strictly better than the old 401 toast-spam.
- **API-key owner-active check** (`verifyOwnerActive`) does per-request Keycloak I/O; its failure during an outage still becomes 401. This PR fixes the JWT path (the UI traffic).

### DIFF
--- a/packages/api-server/src/apps/api-server/auth.ts
+++ b/packages/api-server/src/apps/api-server/auth.ts
@@ -1,4 +1,5 @@
-import { createRemoteJWKSet, jwtVerify } from "jose";
+import { createRemoteJWKSet, jwtVerify, type JWTPayload } from "jose";
+import { JOSEError, JWKSInvalid, JWKSTimeout } from "jose/errors";
 import type { Context, MiddlewareHandler } from "hono";
 import { ALL_SCOPES, type UserIdentity } from "api-server-api";
 import type { Result } from "../../core/result.js";
@@ -33,6 +34,57 @@ export class UnauthorizedError extends Error {
     super(`Unauthorized: ${reason}`);
     this.name = "UnauthorizedError";
   }
+}
+
+/** Token verification could not run: the JWKS could not be retrieved
+ *  (Keycloak unreachable, fetch timeout, non-200/non-JSON response). Not an
+ *  authentication verdict — the token was never evaluated. Mapped to 503 so
+ *  clients retry instead of treating an infra failure as a credential
+ *  rejection; still a deny (fails closed). */
+export class AuthUnavailableError extends Error {
+  readonly reason = "jwks-unavailable";
+  constructor(cause: unknown) {
+    super("Authentication unavailable: JWKS could not be retrieved", {
+      cause,
+    });
+    this.name = "AuthUnavailableError";
+  }
+}
+
+/** Strict allowlist of failures where the JWKS could not be retrieved or
+ *  ingested (jose 6.x): its fetch maps abort timeouts to JWKSTimeout, throws
+ *  bare JOSEError (ERR_JOSE_GENERIC — used nowhere else in the library) on a
+ *  non-200/non-JSON response, throws JWKSInvalid on a fetched-but-malformed
+ *  body (a broken gateway answering 200 with a JSON error), and rethrows
+ *  undici's network `TypeError: fetch failed` unwrapped.
+ *
+ *  The TypeError arm is gated on a `.cause`: undici's fetch failure always
+ *  carries one (ECONNREFUSED, ENOTFOUND, …), whereas jose's own internal
+ *  TypeErrors are message-only. This matters for security — jose can throw a
+ *  bare `TypeError('non-ASCII string encountered in encode()')` while building
+ *  the signing input from an attacker-supplied payload segment, *before* the
+ *  signature is checked; without the `.cause` gate a forged token would be
+ *  misclassified 503/`authn.unavailable` instead of 401/`authn.deny`, letting
+ *  an attacker move forged-token probes out of the deny audit stream. Gating
+ *  on `.cause` (not on the message string, which drifts across Node/undici)
+ *  keeps every real fetch failure a 503 while every token failure — and jose's
+ *  message-only key-material TypeErrors — stays a 401. JWKSNoMatchingKey
+ *  (rotated/foreign kid) is likewise a 401. */
+function isJwksRetrievalFailure(err: unknown): boolean {
+  return (
+    err instanceof JWKSTimeout ||
+    err instanceof JWKSInvalid ||
+    (err instanceof TypeError && err.cause != null) ||
+    (err instanceof JOSEError && err.code === "ERR_JOSE_GENERIC")
+  );
+}
+
+/** Value-free summary for the audit line: error name plus the undici cause
+ *  code (ECONNREFUSED, ENOTFOUND, …) when present. */
+function describeJwksFailure(err: unknown): string {
+  if (!(err instanceof Error)) return String(err);
+  const code = (err.cause as { code?: unknown } | undefined)?.code;
+  return typeof code === "string" ? `${err.name}: ${code}` : err.name;
 }
 
 export interface AuthConfig {
@@ -70,6 +122,7 @@ export interface AuthDeps {
 
 const PUBLIC_PATHS = new Set([
   "/api/health",
+  "/api/ready",
   "/api/auth/config",
   "/api/oauth/callback",
   "/api/slack/oauth/callback",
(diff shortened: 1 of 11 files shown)


packages/api-server/src/apps/api-server/auth.ts
```diff
@@ -35,6 +36,57 @@ export class UnauthorizedError extends Error {
...
+ *  on `.cause` (not on the message string, which drifts across Node/undici)
+ *  keeps every real fetch failure a 503 while every token failure — and jose's
+ *  message-only key-material TypeErrors — stays a 401. JWKSNoMatchingKey
+ *  (rotated/foreign kid) is likewise a 401. */
+function isJwksRetrievalFailure(err: unknown): boolean {
+  return (
+    err instanceof JWKSTimeout ||
+    err instanceof JWKSInvalid ||
+    (err instanceof TypeError && err.cause != null) ||
+    (err instanceof JOSEError && err.code === "ERR_JOSE_GENERIC")
```
dam-code-guardian, line 78: 🟢 **Looks good (noting for reviewers):** This four-arm allowlist is the crux of the fix. The `err.cause != null` gate on the `TypeError` arm is what keeps a forged-token `TypeError('non-ASCII…')` (message-only, no cause) at 401 while undici's `fetch failed` (always has a `.cause`) becomes 503. Gating on `.cause` rather than the message string is the right call — the message drifts across Node/undici versions. Verified against the causeless-TypeError → 401 test.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `5a5b123`

### Summary
Maps transient JWKS-retrieval failures to **503 `authn.unavailable`** (a new `AuthUnavailableError` sentinel) instead of 401, so an unreachable Keycloak during a rollout degrades as a retryable outage rather than a spurious credential rejection; adds a background JWKS boot warm-up gating a real `/api/ready` readiness probe (liveness stays on `/api/health`), with a bounded 240s give-up so a prolonged outage can't wedge the rollout. Docs, telemetry span filter, and rollout timeouts are updated to match.

### Findings
- ✅ **Looks good:** The security-critical `.cause` gate in `isJwksRetrievalFailure` (`auth.ts:73`) is correct and well-reasoned — undici's `fetch failed` always carries a `.cause`, while jose's attacker-forgeable message-only `TypeError('non-ASCII string…')` (thrown from an attacker-supplied payload segment *before* signature verification) does not, so a forged token stays 401/`authn.deny` and never gets relabelled 503. The causeless-TypeError test (`auth-audit.test.ts`) locks this in.
- ✅ **Looks good:** Fails closed throughout — 503 is still a deny, no token ever reaches the application; the `authn.unavailable` / `ws.authn_unavailable` events omit `decision` since no verdict was reached, keeping the credential-denial audit stream clean.
- ✅ **Looks good:** Warm-up latches `ready` on both first success *and* give-up (`jwks-warmup.ts`), so a single-replica deployment won't drop out of every Service endpoint over an unreachable Keycloak; give-up (240s) sits under the Deployment `progressDeadline` and the bumped rollout timeouts, and `pollUntilReady`'s contract (returns on first success, `false` at deadline, `isReady` rejections swallowed) means no unhandled rejection.
- ✅ **Looks good:** `/api/ready` added to `PUBLIC_PATHS` and to the telemetry probe-span filter; readiness (not liveness) points at it in Helm — the split is exactly right.
- ✅ **Looks good:** Test coverage is thorough and precise: 503-vs-401 classification across `JWKSTimeout` / `ERR_JOSE_GENERIC` / `JWKSInvalid` / causeless-TypeError / `JWKSNoMatchingKey` / API-key denial, plus warm-up latch/failure/give-up paths. Stubbing only `"jose"` while keeping `"jose/errors"` real means the classifier is exercised against genuine jose error instances.

✅ No documentation drift detected. Both edited architecture pages (`logging.md`, `security-and-credentials.md`) accurately reflect the new 503/`authn.unavailable` behavior and the readiness-gate/warm-up semantics; both have `Last verified:` bumped to 2026-07-21; no ADR references or volatile content (exact package/file names) leaked into the prose; both remain linked from the architecture landing page.

✅ No findings. The changes align with the existing app-layer conventions in `apps/api-server/`. The sentinel-error + try/catch pattern in the auth middleware matches the surrounding `ForbiddenError`/`UnauthorizedError` handling (the `Result`-type rule is scoped to the domain layer, not this HTTP boundary). `jwks-warmup.ts`'s deep import of `pollUntilReady` from `modules/agents/infrastructure/` is a pre-existing, unchanged app-layer pattern (already used by `acp-relay.ts` and `acp-relay.test.ts`), not a boundary violation introduced by this PR.

### Verdict
APPROVE — a focused, well-tested fix with sound fail-closed security reasoning and honestly documented follow-ups (WS-503 test gap, recovery-probe flap, API-key owner-active path); nothing blocking.

---

PR #2864, opened 2026-07-21
fix(approvals): harden the approvals inbox against the #2324 timestamp bug

### LINKED ISSUE #2767: Dormant timestamp bug of the #2324 class in the approvals inbox
## Problem

The crash behind #2324 ("creating a network rule that already exists fails with an internal error") had a general cause: one database read path returns timestamps as plain strings while the surrounding code is promised date objects, so the first thing that formats such a timestamp throws an internal error.

The approvals subsystem contains an identical dormant instance. The lookup that de-duplicates retried held requests returns rows whose timestamps are strings masquerading as dates. Nothing crashes today only because the single consumer of that lookup reads just the row's identifier — but any future change that uses more of the row (showing the held request's age, its expiry, sorting by resolution time) reproduces the #2324 internal error, this time on the held-requests/inbox path, where a crash means an agent's blocked request can't be actioned.

## Proposed solution

Normalize the timestamps at the same boundary the #2324 fix used, so every read path hands the rest of the system real date values regardless of how the row was fetched. While at it, sweep the codebase for any remaining read paths with the same string-vs-date mismatch so the class of bug is closed rather than fixed one occurrence at a time.

No user-visible behavior changes today; this is prevention.

### Dependencies

- Follow-up to #2324 / PR #2765 (same root cause, same fix pattern).

PetrBulanek:
## What & why

The crash behind #2324 had a general cause: one database read path returned timestamps as plain strings while the surrounding code was promised date objects, so the first thing to format such a timestamp threw an internal error.

The approvals subsystem held an identical **dormant** instance. The lookup that de-duplicates retried held requests returned rows whose timestamps were strings masquerading as dates. Nothing crashed today only because the single consumer of that lookup reads just the row's identifier — but any future change that used more of the row (showing a held request's age, its expiry, sorting by resolution time) would have reproduced the #2324 internal error, this time on the held-requests / inbox path, where a crash means an agent's blocked request can't be actioned.

This change normalizes the timestamps at the same boundary the #2324 fix used, so every read path hands the rest of the system real date values regardless of how the row was fetched. A sweep of the remaining database read paths confirmed this was the last one carrying the same string-vs-date mismatch, so the class of bug is now closed rather than fixed one occurrence at a time.

**No user-visible behavior changes today; this is prevention.**

Closes #2767

### DIFF
--- a/packages/api-server/src/modules/approvals/infrastructure/approvals-repository.ts
+++ b/packages/api-server/src/modules/approvals/infrastructure/approvals-repository.ts
@@ -92,13 +92,15 @@ interface RawPending {
   ownerSub: string;
   sessionId: string | null;
   payload: unknown;
-  createdAt: Date;
-  expiresAt: Date;
-  resolvedAt: Date | null;
+  // Raw `db.execute(sql...)` reads bypass drizzle's timestamptz mapper and
+  // yield strings; typed `.select()`/`.returning()` reads yield Dates.
+  createdAt: Date | string;
+  expiresAt: Date | string;
+  resolvedAt: Date | string | null;
   verdict: string | null;
   decidedBy: string | null;
   status: string;
-  deliveredAt: Date | null;
+  deliveredAt: Date | string | null;
 }
 
 /** Default cap when the caller doesn't specify, and the hard ceiling
@@ -113,6 +115,9 @@ function clampLimit(requested: number | undefined): number {
   return Math.min(requested, MAX_LIST_LIMIT);
 }
 
+const toDate = (v: Date | string): Date =>
+  v instanceof Date ? v : new Date(v);
+
 function toPendingRow(r: RawPending): PendingApprovalRow {
   return {
     id: r.id,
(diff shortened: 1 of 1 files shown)


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `bfb0a08`

## PR #2864: fix(approvals): harden the approvals inbox against the #2324 timestamp bug
**Author:** PetrBulanek | **Branch:** fix/2767-approvals-timestamp-normalization → main | **Changes:** +13 −8 (1 file)

### Summary
Normalizes DB-read timestamps at the approvals repository boundary: `RawPending`'s timestamp fields are widened to `Date | string` to honestly model the two read paths (typed `.select()`/`.returning()` yield `Date`; raw `db.execute(sql...)` yields strings), and a new `toDate` mapper coerces them so `toPendingRow` always hands the rest of the system real `Date` values. Preventive — closes the last instance of the #2324 string-vs-date class of bug; no user-visible behavior change today.

### Findings
- ✅ **Looks good:** The `RawPending` widening fixes a latent type-lie — the raw `findActivePendingExtAuthz` SQL read was already returning strings into fields declared `Date`, so the previous type was unsound. The new union is accurate.
- ✅ **Looks good:** `toDate` and the updated `toPendingRow` are pure, unidirectional parse-direction mappers in `infrastructure/` — the correct layer and shape for absorbing this format difference.
- ✅ **Looks good:** Null handling is explicit (`r.resolvedAt === null ? null : toDate(...)`), preserving the nullable contract rather than coercing `null` through `new Date()`.

✅ No documentation drift detected. The change is internal to the approvals repository with no subsystem behavior/responsibility shift, no docs edited, and no ADR state change.

✅ No findings. The change strengthens the infrastructure→domain mapper (`toDate` normalization in `toPendingRow`) and introduces no architecture-rule violations — mappers stay pure, unidirectional, and confined to the infrastructure layer.

### Verdict
APPROVE — a tightly-scoped, well-reasoned preventive fix that makes the repository's timestamp types sound and normalizes all read paths to domain `Date` values.

---

kapetr (approved):

PR #2868, opened 2026-07-21
fix(dam-vm): binary-safe non-interactive runs so redirected output survives

JanPokorny:
## Problem

`dam-vm cat file.png > file.png` corrupted the file: every non-interactive
`dam-vm` invocation ran through a remote PTY, whose ONLCR translation rewrote
`\n`→`\r\n` in the byte stream and folded stderr into stdout.

## Fix

Allocate a PTY only when the run is genuinely interactive; otherwise use raw
pipes end-to-end.

- **CLI (`dam-run.mjs`, shared with `dam-vm`)**: `interactive` now requires
  stdin *and* stdout to be TTYs for `dam-vm` (so a redirect drops the PTY). It
  passes `&tty=0|1`, and on the raw path prints `OP_STDERR` to stderr and sends
  a real `OP_EOF` on stdin end (a PTY has neither).
- **Server (`dam-vm-server.mjs`)**: branches on `tty`. Interactive → `ptySpawn`
  (unchanged behavior). Non-interactive → a plain piped `spawn`, so `incus exec`
  runs non-interactive and streams bytes verbatim; stderr rides `OP_STDERR`,
  `OP_EOF` closes the command's stdin.

Two new opcodes (`OP_STDERR=0x04`, `OP_EOF=0x05`) are relay-transparent — the
`/vm` relay splices frames byte-for-byte, so only the CLI and server agree on
them.

## Verification (real VPS, dam-dev cluster CA)

- Raw path (`tty=0`): 100 KB `/dev/urandom` round-trip is **byte-exact**
  (client sha256 == in-container sha256); stdout keeps LF, stderr arrives
  separately on `OP_STDERR`.
- Interactive path (`tty=1`): still gets CRLF + merged stderr, exits cleanly.

### DIFF
--- a/packages/dam-vm/dam-vm-server.mjs
+++ b/packages/dam-vm/dam-vm-server.mjs
@@ -30,7 +30,9 @@ import { createServer as createHttpsServer } from "node:https";
 const OP_INPUT = 0x00,
   OP_OUTPUT = 0x01,
   OP_RESIZE = 0x02,
-  OP_EXIT = 0x03;
+  OP_EXIT = 0x03,
+  OP_STDERR = 0x04, // raw (no-PTY) path only: stderr kept separate from stdout
+  OP_EOF = 0x05; // raw path only: close the command's stdin
 
 const LISTEN_HOST = process.env.DAM_VM_LISTEN_HOST || "0.0.0.0";
 const LISTEN_PORT = process.env.DAM_VM_PORT
@@ -249,6 +251,10 @@ wss.on("connection", async (ws, req) => {
   if (argv.length === 0) argv = ["bash", "-l"];
   const cols = Math.min(500, parseInt(url.searchParams.get("cols"), 10) || 80);
   const rows = Math.min(300, parseInt(url.searchParams.get("rows"), 10) || 24);
+  // Interactive (client stdin+stdout are TTYs) → PTY. Otherwise raw pipes, so
+  // binary stdout survives redirection (`dam-vm cat x > x`). Default to PTY for
+  // an older client that doesn't send the flag.
+  const wantTty = url.searchParams.get("tty") !== "0";
   // W3C trace context forwarded by dam-vm so the command joins the session's
   // trace (same contract as the dam-run exec server).
   const traceEnv = ["traceparent", "tracestate"].flatMap((k) => {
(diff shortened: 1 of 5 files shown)


packages/dam-vm/dam-vm-server.mjs
```diff
@@ -266,47 +272,71 @@ wss.on("connection", async (ws, req) => {
...
+    onEof = () => {}; // a PTY has no EOF; interactive clients don't send OP_EOF
+    onResize = (c, r) => term.resize(c, r);
+    kill = () => term.kill();
+  } else {
+    const child = spawn("incus", execArgs); // stdio pipes → incus non-interactive
+    child.stdout.on("data", (d) => send(OP_OUTPUT, d));
+    child.stderr.on("data", (d) => send(OP_STDERR, d));
+    child.stdin.on("error", () => {}); // swallow EPIPE if the command already exited
+    child.on("error", () => exit(1));
+    child.on("exit", (code, signal) => exit(code ?? (signal ? 1 : 0)));
```
dam-code-guardian: 🟡 **Warning:** Completion is keyed off `'exit'`, and `exit()` immediately closes the socket (`ws.close(1000)`). Node's `'exit'` fires when the process terminates but **before** its piped `stdout`/`stderr` are guaranteed fully drained — late `send(OP_OUTPUT/OP_STDERR, …)` frames then hit the `ws.readyState === OPEN` guard and are silently dropped, which can truncate trailing bytes (the byte-exactness this PR guarantees). Prefer `'close'`, which fires after stdio EOF — the same file already does exactly this at line 79.

```suggestion
    child.on("close", (code, signal) => exit(code ?? (signal ? 1 : 0)));
```

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `044d338`

## PR #2868: fix(dam-vm): binary-safe non-interactive runs so redirected output survives
**Author:** JanPokorny | **Branch:** fix/dam-vm-binary-output → main | **Changes:** +93 −38 (2 files)

### Summary
Allocates a remote PTY only for genuinely interactive runs; non-interactive `dam-vm` invocations now use raw pipes end-to-end so redirected binary output (`dam-vm cat x > x`) survives without CRLF translation or stderr/stdout folding. Adds two relay-transparent opcodes (`OP_STDERR=0x04`, `OP_EOF=0x05`) that only the CLI and dam-vm-server interpret.

### Findings
- 🟡 **Warning:** On the raw (no-PTY) path, completion is keyed off `child.on("exit")`, which closes the socket immediately (`packages/dam-vm/dam-vm-server.mjs:319`). Node's `exit` fires when the process terminates but **before** its piped `stdout`/`stderr` are guaranteed fully drained; any late `send(OP_OUTPUT, …)` is then dropped by the `ws.readyState === OPEN` guard, which can truncate trailing bytes — the exact byte-exactness this PR sets out to guarantee. Prefer `child.on("close", …)` (fires after stdio EOF), as this same file already does at line 79. The 100 KB round-trip verification likely passes because the pipes usually drain within the same tick, but it's timing-dependent.
- 🟢 **Suggestion:** `kill = () => child.kill()` on `ws.close` sends SIGTERM to the local `incus exec` client process, not necessarily the in-container process; acceptable (matches prior PTY behavior), just noting it doesn't propagate into the container.
- ✅ **Looks good:** The new opcodes are genuinely relay-transparent — `harness-vm-relay.ts` uses `spliceClient`, which forwards frames byte-for-byte via `upstream.send(d, {binary})` and never calls `decodeFrame`. No collision with `OP_INPUT..OP_EXIT` (0x00–0x03).
- ✅ **Looks good:** The `interactive` gate correctly requires *both* stdin and stdout TTYs for `dam-vm` while keeping `dam-run`'s stdin-only heuristic; the `tty=0` default-to-PTY fallback preserves older-client behavior.
- ✅ **Looks good:** `child.stdin.on("error", () => {})` sensibly swallows EPIPE when the command has already exited.

⚠️ **Minor drift.**
- ❌ **Check 1 (architecture-page drift):** `docs/architecture/dam-vm.md:9` still describes the relay as bridging *"the socket to an `incus exec` PTY"* over the *"shared terminal frame protocol (input/output/resize/exit)"* — i.e. always-PTY, four opcodes. This PR makes the PTY conditional (raw pipes for non-interactive runs) and extends the protocol with stderr/EOF frames on the raw path, but the page isn't updated in the same PR. Suggested: note the PTY-for-interactive / raw-pipes-for-non-interactive split and the stderr/EOF frames, and bump `Last verified:`. Non-blocking.

✅ No findings. Both changed files are dependency-free Node scripts (WebSocket relay + CLI) outside the tRPC/Zod/slice-composition application architecture; no architecture rule applies to the diff.

### Verdict
COMMENT — Solid, well-reasoned fix with real verification; the raw-path `exit`-vs-`close` drain ordering is worth addressing to fully guarantee no trailing-byte truncation, and the dam-vm architecture page should be updated to match.

---

PR #2848, opened 2026-07-20
Durable per-replier forks: identity, egress auto-decline, budgets, UI

jezekra1:
Implements #2843. Seven commits, one per workstream — reviewable in order.

https://claude.ai/code/artifact/96f3645d-52e2-4e42-bf38-a340d8ebc574

## What changed

- **Lifecycle** (controller + api-server): one Fork per (Agent, replier), deterministically named, reused across turns/threads. Two-tier idle policy: pods torn down after `controller.forks.hibernateAfter` (5m, busy-probe guarded), CR deleted after `controller.forks.expireAfter` (48h). Replaces per-turn teardown, the 60-min reaper, and the api-server's in-memory fork state.
- **Identity**: forks present their own `/api/agents/<forkId>/mcp` and dial their own ext-authz Service. The MCP endpoint resolves a fork into an acting context — artifact tools scope to the **replier's** library; skills/schedules/experiments tools are not registered on fork sessions. Closes the parent-impersonation hole.
- **Egress**: the gate resolves the parent's rules (owner stays policy authority) but a would-hold request **auto-declines** — foreign turns never land in the owner's inbox.
- **Budgets**: live fork pods reserve against the **replier** at the parent's Size; Job creation (start + wake) is gated against the replier's Ceiling; denial → `OverBudget` with figures relayed in Slack.
- **UI**: sandbox "Forks" section (who/state/last activity/End now) + "Your forks" itemization next to the budget meter.
- Docs: architecture pages + glossary updated; CRD schema gen 2 (`Hibernated` phase).

## Follow-ups (deliberately out of scope)

- Plumb `DeniedHttpResponse` through the ext-authz proto so auto-decline carries agent-visible copy instead of a bare 403.
- Resolve replier subs to emails in the fork UI (raw sub shown today).
- E2E coverage for the fork lifecycle (hibernate/wake/expire) — unit-tested on both sides; not yet exercised on a live cluster.

Related: #2119 (kind-matched fork credentials — unchanged here), #2131 (consent/transparency — composes with the owner-gate model, see issue thread).

PR #2870, opened 2026-07-21
feat(egress): per-agent L7 promotion on the Agent CR

### LINKED ISSUE #2865: Egress-rule L7 promotion should live on the Agent, not a synthetic owner-scoped secret
## Problem

When an egress rule narrows a host by **path, method, or port**, that host has to be moved onto the gateway's TLS-terminating (L7) interception chain — otherwise the gate only sees the hostname and the narrowing silently doesn't apply to HTTPS traffic. Today the platform signals that promotion to the controller as a **synthetic, credential-less "allow-only" marker**, scoped to the owner and keyed `(owner, host)`.

Several things are off about this:

- **It's a second source of truth for the gateway's egress shape.** The rule itself lives in the api-server's own store; the thing that actually shapes the gateway lives in a parallel object that looks like a credential but carries none. They can drift.
- **The grain doesn't match the rule.** Egress rules are **per-agent**, but the promotion marker is **per-owner** and is applied to the gateway *ungated* (it deliberately bypasses the per-agent grant filter, since grants gate credential access, not rule enforcement). So a path rule on one of an owner's agents promotes that host on **every** agent the owner runs.
- **It amplifies gateway rolls.** Because every promotion marker lands on every one of the owner's gateways, a single path rule on one agent bounces the proxy pod of every sibling gateway within a resync cycle — a brief egress interruption on agents that have no rule on that host and gain only an unused interception chain. (Tracked as its own symptom in the companion "one agent's rule rolls every gateway" issue.)
- **It's the root cause of the fork egress gap.** A fork resolves its gateway's chains from the *replier's* owner, never the *parent owner's* promotion markers, so the parent's path/method/port rules aren't enforced for foreign turns. (Companion "forks enforce egress only at host granularity" issue.)

## Proposed solution

Project the L7-promotion decision onto the **Agent resource itself**, at per-agent grain — the same place other per-agent intent (connection grants) already lives — instead of a shadow marker resolved by owner label. The gateway is already rendered from the agent's declared state plus its credential secrets, so promotion becomes typed, per-agent, and readable by anything that references the agent. Two consequences fall out for free:

- Only the agent whose rule changed re-renders and rolls its gateway.
- Forks, which derive from a parent agent, inherit the parent's promotions directly.

### Scope
- The promotion signal and gateway chain shaping only.

### Out of scope
- Egress rule storage and per-request verdict evaluation — unchanged, and stays instant (no gateway roll for a plain allow/deny/verdict change).
- Credential injection and which credentials a gateway mounts.

### Dependencies
- Root-cause fix for the two companion issues (fork host-granularity gap; owner-wide gateway roll amplification). Context: forks work in #2843.

### LINKED ISSUE #2866: Forks enforce egress rules only at host granularity — path/method/port narrowing is bypassed
## Problem

A **fork** — the runtime a Slack teammate's turn runs on when the replier isn't the agent owner — is meant to egress under the agent owner's egress rules. Host-level rules work: a host the owner never allowed stays blocked, and a host-level deny blocks. But when the owner *allows* a host and then **narrows it by path, method, or port**, that narrowing is not applied to the teammate's turns.

**Concrete scenario:** the owner allows `api.github.com` but only `GET /repos/*`. The owner's own turns are held to that. A teammate replying in the same Slack thread can reach *any* path on `api.github.com` — their turn's egress never lands on the L7 interception chain that would let the gate see the path, so it passes at host granularity. The owner's injected credential also isn't applied on that path.

**Mechanism:** the fork's gateway is built from the *replier's* own connections and promotion markers, resolved by the replier's owner label — so it carries the L7 chains for hosts the *replier* narrowed on their own agents, and never the parent owner's. Meanwhile the egress gate correctly evaluates the *parent's* rules. So the fork's chain shape reflects the replier's egress policy while the verdict logic uses the parent's — and any host the parent narrowed by path simply isn't on an L7 chain in the fork, so the narrowing is invisible there.

**Why it matters:** an owner who relies on path/method scoping to contain what leaves their workspace does not get that containment for foreign turns. Security-relevant, though bounded to *granularity* — the default-deny boundary still holds (off-policy hosts stay blocked); what leaks is precision on already-allowed hosts.

**Not a regression:** pre-existing. The fork gateway has always been built from the replier's connections; the durable-forks work (#2843) changed fork lifetime, not this. Filing now because the analysis around #2843 surfaced it.

## Proposed solution

Foreign turns must be held to the parent owner's path/method/port rules at the same granularity as the owner's own turns. The clean path is the companion "promotion belongs on the Agent" refactor — a fork derives from a parent agent, so per-agent promotion on the Agent resource is inherited directly. Interim option if that refactor lands later: fold the parent's promotion markers into the fork gateway's chain set when it's rendered.

### Scope
- Fork (foreign-replier) egress on person-scoped Slack bindings.

### Dependencies
- Cleanly closed by the "promotion belongs on the Agent" issue. Related: #2843 (durable forks), #2119 (fork credential kind-matching — a different fork/egress concern).

### LINKED ISSUE #2867: A path/method/port rule on one agent rolls every gateway its owner runs
## Problem

When an egress rule narrows a host by **path, method, or port**, the platform promotes that host onto the gateway's L7 interception chain. That promotion is materialized as an **owner-scoped** marker, and every one of that owner's agent gateways picks it up — so a rule change on **one** agent changes the rendered gateway shape for **all** of that owner's agents. Within a reconcile resync cycle (~30s) every running gateway that owner has rolls its proxy pod, each roll a brief egress interruption. Agents with no rule on that host roll too, gaining only an unused interception chain their own rules will still deny.

**Concrete scenario:** an owner runs 8 agents and adds "only `GET /repos/*` on `api.github.com`" to one of them. Within ~30s all 8 gateways bounce their proxy pods; the 7 unrelated agents each take an egress blip for a rule that doesn't apply to them.

**Why it matters:** an unrelated, single-agent config change causes owner-wide egress disruption that scales with how many agents the owner runs. Operational, not a security issue — but surprising and avoidable.

**Bounds:** it's once per new `(owner, host)` promotion (idempotent, not per edit); hibernated agents don't roll, they pick it up on next wake; and it's a no-op if a credentialed secret already covers that host.

## Proposed solution

Scope promotion **per-agent** — the companion "promotion belongs on the Agent" refactor — so only the agent whose rule changed re-renders and rolls its gateway. Same underlying fix; distinct, separately-observable symptom.

### Scope
- Gateway rendering and roll-triggering for L7-promoted hosts.

### Dependencies
- Fixed by the "promotion belongs on the Agent" issue. Same root cause as the fork host-granularity issue.

jezekra1:
Fixes #2865. Fixes #2866. Fixes #2867.

**Based on `main`.** The durable-forks branch (#2848) should be rebased on top of this once merged — its fork reconciler will inherit the parent's promotions through the same call sites (an adaptation for the durable-fork shape, including the fork gateway rev-digest, exists at commit `2cedd20a` for reference).

## What changed

An egress rule narrowing a host by **path, method, or port** needs that host on the gateway's TLS-terminating (L7) chain to be enforceable over HTTPS — the L4 catch-all sees only SNI. That promotion used to be an **owner-scoped**, credential-less marker Secret that every gateway of the owner picked up ungated. Consequences: one agent's rule rolled every sibling gateway within a resync (#2867), and fork gateways — built from the *replier's* owner label — never carried the *parent's* promotions, so path/method/port narrowing silently didn't bind foreign turns (#2866).

Promotion is now **per-agent intent on the Agent resource** (`spec.l7Hosts`, CRD schema gen 5), the same grain as connection grants:

- **Controller**: renders uncredentialed L7 chains + leaf-cert SANs from `spec.l7Hosts` ∪ credential Secrets; the `envoy-secrets-rev` roll trigger digests the list (order-insensitive), so only the touched agent's gateway rolls.
- **Forks**: the fork gateway inherits the **parent's** `l7Hosts` alongside the replier's credential Secrets — the gate already evaluates the parent's rules; now the chain shape matches the policy authority. Fork/run agent pods project the MITM CA whenever promoted chains exist, credentials or not. (Per-turn fork pods pick a new parent rule up on the next turn; no roll trigger needed.)
- **api-server**: promotion is an Agent-CR merge-patch keyed by the rule's `agentId` (manual create/update + inbox-verdict paths). The rule writer's `ownerSub` parameter is gone.
- **Migration**: a one-shot startup backfill projects active narrow rules onto Agent CRs, then retires the legacy marker Secrets — only after every patch landed. The controller consumes **both** signals, so rules stay enforced mid-migration.

## Notes for review

- Legacy allow-only Secret *consumption* stays in the controller (migration window + the missing-SDS degrade path); only the api-server's writes moved.
- Upgrade behavior: one gateway roll per agent that has promoted hosts, as the rev digest changes shape — bounded, per-agent, once.
- `mise run check` green; controller + api-server unit tests cover the new chain rendering, rev sensitivity, fork inheritance, CA projection, port behavior, and backfill (including the keep-markers-on-partial-failure path).


packages/api-server/src/modules/egress-rules/infrastructure/k8s-agent-l7-hosts-port.ts
```diff
@@ -0,0 +1,47 @@
...
+      if (!agent) return;
+      const spec = (agent.spec ?? {}) as { l7Hosts?: string[] };
+      const current = spec.l7Hosts ?? [];
+      const merged = [...new Set([...current, ...hosts])].sort();
+      if (merged.length === current.length) return;
+      // RFC 7386 merge-patch replaces the array wholesale. Two concurrent
+      // rule writes for the same agent can race read-modify-write; rule
+      // edits are user-paced and the next write re-converges the union,
+      // so no optimistic-concurrency loop here.
+      await client.patchCustomObject("agents", agentId, {
```
dam-code-guardian: 🟢 **Suggestion:** This read-modify-write of `spec.l7Hosts` has no optimistic-concurrency guard, so two concurrent `ensure` calls for the same agent can drop one host until the next write for that agent — which may never arrive if no further rule is added. The union-merge semantics make a `resourceVersion`-guarded retry cheap to add. Acceptable at the current user-paced write cadence as the comment notes; worth revisiting if promotions become automated or bursty.

dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `289c849`

### Summary
Moves egress **L7 promotion** from owner-scoped, credential-less "allow-only" marker Secrets to per-agent intent on the Agent CR (`spec.l7Hosts`, CRD schema gen 5). This fixes three real bugs: a promotion on one agent rolling every sibling gateway (#2867), fork gateways never inheriting the parent's promotions so path/method/port narrowing silently didn't bind foreign turns (#2866), and the owner-wide grain itself (#2865). A one-shot startup backfill projects active narrow rules onto Agent CRs and retires the legacy markers only after every patch lands; the controller consumes both signals during the migration.

### Findings
- ✅ **Looks good:** The #2866 fix is correct at the crux — the fork reconciler renders chains from the **parent's** `agentSpec.L7Hosts` (resolved via `forkSpec.AgentName`) while credential Secrets stay scoped to the replier's `foreignSub` (`fork.go:84,111,163`). Policy authority and credential authority are cleanly separated.
- ✅ **Looks good:** `envoySecretsRev` digests `l7Hosts` and sorts `parts[1:]` before hashing (`envoy.go`), so promoting a host rolls the gateway but a reordered spec does not — exactly what the order-insensitivity test asserts.
- ✅ **Looks good:** The backfill retires legacy markers only when `failed === 0`, and the comment correctly reasons that a kept marker is harmless (controller renders both) while a prematurely-deleted one silently unenforces a rule (`l7-promotion-backfill.ts:1002`). The keep-on-partial-failure path is unit-tested.
- ✅ **Looks good:** `needsL7Promotion(method, pathPattern, port)` is the single domain predicate applied uniformly across the manual-create service, the inbox-verdict writer, and the backfill — no divergence between write paths.
- ✅ **Looks good:** CRD schema generation bumped 4→5 with matching CRD YAML, `AgentSchemaGeneration` constant, `zz_generated.deepcopy.go` (L7Hosts deep-copied), and generated `crd-types.gen.ts` — the full generated surface is consistent.
- ✅ **Looks good:** Leaf-cert SAN and CA-projection gates were widened from `len(secrets) > 0` to `… || len(l7Hosts) > 0` at every site (`envoy.go` volumes/container, `ephemeral_pod.go`, run/fork pods), so a promoted-but-uncredentialed host still gets a leaf cert and the agent trusts the MITM CA. Consistently applied and tested.
- 🟢 **Suggestion:** `AgentL7HostsPort.ensure` does a read-modify-write of `spec.l7Hosts` with an RFC 7386 merge-patch and no optimistic-concurrency guard (`k8s-agent-l7-hosts-port.ts:711`). The inline comment justifies this (user-paced rule edits, next write re-converges the union). The residual risk is narrow but real: two concurrent adds for the same agent can drop one host until the *next* write for that agent — which may never come if no further rule is added. Given union-merge semantics a resourceVersion-guarded retry would close it cheaply; acceptable as-is, worth a follow-up if promotions ever become automated/bursty.

✅ No documentation drift detected. The L7-promotion behavior change is reflected in `docs/architecture/security-and-credentials.md` (new "L7 promotion" paragraph + fork-inheritance update) in the same PR, `Last verified:` is bumped to 2026-07-21, and the added prose cites only issue refs (no ADR leak) with no volatile package/path content. `docs/ubiquitous-language.md` gained an "L7 Promotion" entry (out of architecture scope, no drift).

✅ No findings. The change aligns with the layered architecture: `AgentL7HostsPort` is a proper `infrastructure/` port with a factory adapter encapsulating the merge-patch; concrete wiring stays at the composition root (`compose.ts`/`app.ts`/`index.ts`); the `needsL7Promotion` predicate stays pure in `domain/`; and `listActiveForPromotionScan()` is a domain-meaningful repository operation that keeps the promotion predicate out of SQL. The `l7Hosts?` optional dependency is carried over verbatim from the prior `allowOnlySecrets?` dep (not introduced by this PR).

### Verdict
COMMENT — Well-structured, well-tested fix for three genuine egress-enforcement bugs; the only note is a pre-acknowledged best-effort concurrency choice on the `spec.l7Hosts` merge-patch, which is reasonable at the current write cadence.

---

jezekra1:
Review findings addressed in `75f7e1ea`:

**1. Bare `*` + narrowing rule breaks reconverge — fixed.** `promotedHosts()` now excludes the bare `*` host, with a comment explaining why: no single SNI chain terminates every host, so a narrowing on `*` is not projectable — it stays enforced at host granularity on the L4 path, exactly as it always has (the legacy marker path had the same gap, just silently: it wrote a `*` marker whose chain matched nothing meaningful). Tests cover both the live write path and the backfill (a `(*, GET, /x)` row no longer fails a backfill pass). Whether a narrowing on `*` should be rejected outright at rule creation is a product question I'd keep separate — today it's accepted and means "narrow every host you can see."

**2. Create-path atomicity — accepted behavior, now documented.** The rule row commits, then the projection write can fail loud. That's deliberate: fail-loud beats silently-unpromoted (#2322), and the gap self-heals — the projection is recomputed from the table on the agent's next rule mutation, and a missed promotion also heals on api-server boot (the startup backfill re-projects every agent with active narrow rules). Comment added on `reconvergePromotions`.

**3. Prune-check vs. real patch response — verified empirically.** Deployed this branch to a live cluster today: the startup backfill ran `set()` against the real API server (`[l7-backfill] backfilled 1 agent(s)`), the merge-patch response carried the applied `spec.l7Hosts`, and the guard passed on the healthy write. The K8s API returns the resulting object for PATCH; the fake mirrors that.

**4. `deleteSecret` NotFound race — already handled.** The k8s client's `deleteSecret` swallows 404s, so a multi-replica race can't fail a backfill pass.

Also from the earlier review at `289c849`: the read-modify-write on `spec.l7Hosts` is now `resourceVersion`-conditional with bounded retry (landed in the previous push), so concurrent writes conflict-and-retry instead of dropping hosts.

Live verification on the sandbox cluster (branch deployed via `cluster:install`): controller runs clean against CRD gen 5, backfill projected the existing agent's narrow rule, promoted chain + leaf SAN rendered, demotion drops both, and admission rejects injection-shaped hosts while accepting valid ones and `*.` wildcards.

pilartomas (approved):

PR #2881, opened 2026-07-22
fix(connections): back off failing OAuth token refreshes

pilartomas:
## Problem

The OAuth token-refresh loop (`createOAuthRefreshLoop`) re-queries every *due* connection (token expiring within the 5-minute skew) on each ~60s tick and, on failure, only incremented a counter and logged. A connection whose token can no longer be refreshed — a revoked/expired refresh token, a `invalid_grant` from the provider — stays in the due set forever, so it was retried against the provider's token endpoint **every 60 seconds indefinitely**, flooding the logs and hammering the IdP.

## Change

Adds a per-connection failure backoff, held in-memory in the loop's closure (the loop is a boot-time singleton):

- On failure, record consecutive failures and an exponential `nextAttempt` — `base·2^(n-1)` capped at max (defaults: **base 60s, cap 30m**). A dead token now settles to ~1 attempt / 30 min instead of ~1440 / day.
- A due connection is **skipped** while `now < nextAttempt` (surfaced as a new `skipped` count on `tickOnce`).
- The entry is **cleared on the first success** (resetting the failure counter) and **pruned** once the connection drops out of the due set — no unbounded map growth.
- Applies to both the OAuth refresh and the client-credentials remint paths.

The backoff computation is a small pure exported helper (`backoffDelayMs`) so it's unit-testable in isolation.

## Scope / non-goals

This is the minimal, safe fix — pure backoff, **no error-model or schema change**. Deliberately *not* included (filed as follow-ups):

1. Classifying permanent (`invalid_grant`) vs transient (network/5xx) failures to fast-path dead tokens.
2. A `status` / `lastError` column on `connections` so a "needs re-auth" state is visible in the UI instead of only inferrable from logs.

State is in-memory, so a process restart re-attempts immediately — acceptable for this fix; persistence is part of follow-up (2).

## Tests

New `oauth-refresh-backoff.test.ts`:
- `backoffDelayMs` — doubling, cap, and zero-failure guard.
- Loop: exponential skip/retry across a controllable clock; **failure-counter reset on success** (mutation-verified: the test fails if the success-clear is removed); prune-on-leave; and the client-credentials remint branch backs off identically.

`mise run` gates green for the package: unit tests (9), `tsc --noEmit`, eslint, prettier.

### DIFF
--- a/packages/api-server/src/__tests__/unit/oauth-refresh-backoff.test.ts
+++ b/packages/api-server/src/__tests__/unit/oauth-refresh-backoff.test.ts
@@ -1,0 +1,241 @@
+import { describe, it, expect } from "vitest";
+import type { Db } from "db";
+import type { ConnectionAuthConfig } from "api-server-api";
+import {
+  backoffDelayMs,
+  createOAuthRefreshLoop,
+} from "../../modules/connections/services/oauth-refresh.js";
+import { createOAuthEngine } from "../../modules/connections/infrastructure/oauth-engine.js";
+import type { ConnectionTemplateRegistry } from "../../modules/connections/domain/connection-template.js";
+import type { SecretStore } from "../../modules/secret-store/index.js";
+
+const AUTH: Extract<ConnectionAuthConfig, { kind: "oauth" }> = {
+  kind: "oauth",
+  clientId: "cid",
+  refreshTokenRef: {
+    storeId: "test",
+    path: "secret-p",
+    field: "refresh_token",
+  },
+  accessTokenRef: { storeId: "test", path: "secret-p", field: "access_token" },
+  scopes: ["read"],
+  tokenUrl: "https://auth.example.com/token",
+  authorizationUrl: "https://auth.example.com/authorize",
+  expiresAt: 1000,
+  connectedAt: 900,
+};
+
+const CC_AUTH: Extract<ConnectionAuthConfig, { kind: "client-credentials" }> = {
+  kind: "client-credentials",
+  clientId: "cid",
+  clientSecretRef: {
+    storeId: "test",
+    path: "secret-p",
+    field: "client_secret",
+  },
+  accessTokenRef: { storeId: "test", path: "secret-p", field: "access_token" },
+  issuerUrl: "https://auth.example.com/realms/main",
+  tokenUrl: "https://auth.example.com/token",
+  scopes: ["read"],
+  expiresAt: 1000,
+  connectedAt: 900,
+};
+
+interface RawRow {
+  id: string;
+  owner: string;
+  templateId: string;
+  name: string;
+  inputs: unknown;
+  auth: unknown;
+  contributions: unknown;
+}
+
+const rowFor = (id: string, auth: ConnectionAuthConfig = AUTH): RawRow => ({
+  id,
+  owner: "owner-sub",
+  templateId: "custom-oauth",
+  name: id,
+  inputs: {},
+  auth,
+  contributions: [],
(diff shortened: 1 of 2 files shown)


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `e8da4f9`

## PR #2881: fix(connections): back off failing OAuth token refreshes
**Author:** pilartomas | **Branch:** fix/oauth-refresh-backoff → main | **Changes:** +283 −7 (2 files)

### Summary
Adds a per-connection exponential backoff (in-memory in the loop's closure) so a connection whose token can no longer be refreshed — a revoked/expired refresh token — is no longer retried against the IdP every ~60s forever. A dead token settles to ~1 attempt / 30 min; the entry clears on first success and is pruned when the connection leaves the due set. Applies to both the OAuth refresh and client-credentials remint paths.

### Findings
- 🟢 **Suggestion:** The `skipped` count is computed and returned but never logged by the loop itself, so a connection stuck in backoff is invisible between the failure log lines (up to 30 min apart). Consider a periodic debug line when `skipped > 0`, or fold it into an existing tick summary if one exists. (`packages/api-server/src/modules/connections/services/oauth-refresh.ts:126`)
- 🟢 **Suggestion:** `refreshOne` throws a bare `Error("no refresh token ref")` for a case the caller already guards against at line 96, so it's effectively unreachable dead-defense; harmless, but a comment noting it's a belt-and-braces invariant would clarify intent. (`packages/api-server/src/modules/connections/services/oauth-refresh.ts:212`)
- ✅ **Looks good:** Prune-before-iterate correctly bounds the map — deleting keys not in `dueIds` before the loop, and `backoff.delete`/`set` on the current key mid-iteration is safe for a `Map`.
- ✅ **Looks good:** `startedAt = now()` is snapshotted once per tick, so the skip comparison and the `nextAttempt` it writes use one consistent clock — no intra-tick drift.
- ✅ **Looks good:** Failure counter is read from `state` captured *before* the try, so consecutive failures accumulate correctly and a success clears the entry via `backoff.delete` (mutation-verified by the test).
- ✅ **Looks good:** `backoffDelayMs` clamps the exponent to 30, preventing `2 ** n` overflow → `NaN` for a long-dead connection; extracted as a pure exported helper and unit-tested (doubling, cap, zero-failure guard).

✅ No documentation drift detected. The change alters the refresh loop's internal retry *cadence*, not its responsibility or the externally-described behavior; `docs/architecture/connections.md` (which describes the loop refreshing tokens before expiry and a persistently-failing connection surfacing as `expired`) remains accurate, and retry-cadence specifics are the kind of volatile detail the guidelines keep out of architecture pages. No architecture page was edited, so no `Last verified:` bump is owed.

✅ No findings. Follows the service-factory DI pattern (`createOAuthRefreshLoop(deps)`); the `now` injection keeps the loop deterministically testable; the new `backoffBaseMs`/`backoffMaxMs`/`now` deps are optional with sane defaults; the `OAuthRefreshLoop.tickOnce` return type was updated consistently for the new `skipped` field. Backoff bookkeeping stays inside the application layer where it belongs, and `backoffDelayMs` is a genuinely pure domain-style helper. Tests isolate the loop's bookkeeping with a fake `Db`/`SecretStore` and a controllable clock, and the success-reset assertion is mutation-sensitive.

### Verdict
APPROVE — Minimal, well-scoped fix for a real IdP-hammering bug; correct backoff bookkeeping, clean DI, and thorough clock-driven tests. Only non-blocking suggestions.

---

PR #2890, opened 2026-07-22
perf(platform-base): symlink ~/.cache to pod-local disk

JanPokorny:
## Why

`/home/agent` is a shared RWX network volume (owner pod + fork pods mount it concurrently). Tool caches — mise, uv wheels, npm — are high-churn disposable data that load the storage backend for no benefit.

## What

- `entrypoint.sh` ensures `$HOME/.cache` is a symlink to `/tmp/agent-cache` (`/tmp` is the non-persisted emptyDir from the default mounts, i.e. node-local disk). Runtime is the only place this works — a build-time symlink would be shadowed by the PVC mount.
- The symlink persists on the volume but `/tmp` is fresh per pod, so the target is (re)created every boot to keep the link from dangling.
- A pre-existing real `.cache` directory (older volumes) is discarded — caches are disposable. `ln -sfn` keeps the swap idempotent when the owner pod and a fork pod boot the same volume together; each pod then resolves the link to its own emptyDir, so pods stop contending on a shared cache dir.
- `docs/architecture/persistence.md` notes the carve-out in the hibernation section.

## Trade-off

Cache-backed state under `~/.cache` now dies with the pod — first use after a restart re-downloads (uv wheels, mise tool downloads).

## Testing

Simulated the three boot scenarios against the snippet: legacy real dir replaced by symlink, restart re-creates the target while keeping the link, second concurrent pod is a no-op.

### DIFF
--- a/packages/platform-base/entrypoint.sh
+++ b/packages/platform-base/entrypoint.sh
@@ -20,6 +20,20 @@ if [ -s "$mitm_ca" ]; then
 		|| echo "agent-entrypoint: WARNING: could not trust the platform CA; intercepted hosts may fail TLS" >&2
 fi
 
+# $HOME is a shared RWX network volume; cache traffic (mise, uv, npm, ...)
+# would hammer it, so ~/.cache points at pod-local disk (/tmp is an emptyDir)
+# instead. Caches are disposable, so a pre-existing real directory (older
+# volumes, or anything a harness recreated) is discarded. `ln -sfn` keeps the
+# swap idempotent when the owner pod and a fork pod boot the volume together.
+# The symlink persists on the volume but /tmp is fresh every pod, so the
+# target is (re)created each boot to keep the link from dangling.
+home="${HOME:-/home/agent}"
+mkdir -p /tmp/agent-cache
+if [ ! -L "$home/.cache" ]; then
+	rm -rf "$home/.cache"
+	ln -sfn /tmp/agent-cache "$home/.cache"
+fi
+
 # `dam-vm` only works when the deployment has a VM host configured (the
 # controller sets DAM_VM_ENABLED then). Drop the whole dam-vm block from the
 # agent instructions otherwise, so we don't advertise a command that would just
--- a/docs/architecture/persistence.md
+++ b/docs/architecture/persistence.md
@@ -1,6 +1,6 @@
 # Persistence
 
-Last verified: 2026-07-15
+Last verified: 2026-07-22
 
 ## Overview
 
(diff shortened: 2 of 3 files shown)


packages/platform-base/entrypoint.sh
```diff
@@ -20,6 +20,20 @@ if [ -s "$mitm_ca" ]; then
 		|| echo "agent-entrypoint: WARNING: could not trust the platform CA; intercepted hosts may fail TLS" >&2
 fi
 
+# $HOME is a shared RWX network volume; cache traffic (mise, uv, npm, ...)
```
dam-code-guardian, line 23: 🟢 **Suggestion:** `npm`'s default cache on Linux is `~/.npm`, not `~/.cache/npm`, so this symlink doesn't actually redirect npm's cache — only XDG-respecting tools (uv `~/.cache/uv`, mise `~/.cache/mise`) are moved. Consider dropping `npm` from the example list so the comment doesn't imply coverage it doesn't provide. Functionally the change is still a clear win.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `f11e090`

## PR #2890: perf(platform-base): symlink ~/.cache to pod-local disk
**Author:** JanPokorny | **Branch:** perf/home-cache-local-disk → main | **Changes:** +15 −2 (2 files)

### Summary
At boot, `entrypoint.sh` repoints `$HOME/.cache` to `/tmp/agent-cache` (a node-local emptyDir) so high-churn tool caches stop loading the shared RWX `$HOME` network volume. `persistence.md` documents the carve-out in the hibernation section.

### Findings
- ✅ **Looks good:** The core design is correct and neat — the symlink itself persists on the shared volume, but its target `/tmp/agent-cache` resolves to each pod's own node-local emptyDir, so the owner pod and fork pods stop contending on one shared cache dir.
- ✅ **Looks good:** Boot-time idempotency is handled: `mkdir -p /tmp/agent-cache` recreates the fresh-per-pod target so the persisted link never dangles; the `[ ! -L ]` guard + `rm -rf` discards a legacy real `.cache` once, then `ln -sfn` is a no-op on subsequent boots and safe under concurrent owner/fork boot.
- ✅ **Looks good:** Correctly placed at runtime (not build time — a build-time link would be shadowed by the PVC mount), and `home="${HOME:-/home/agent}"` defends against an unset `HOME` under `set -u`.
- 🟢 **Suggestion:** The comment and PR body list `npm` among the redirected caches, but npm's default cache on Linux is `~/.npm`, not `~/.cache/npm` — so npm's cache isn't actually moved by this symlink (only XDG-respecting tools like uv `~/.cache/uv` and mise `~/.cache/mise` are). The change is still a clear win; just drop `npm` from the example list to avoid implying coverage it doesn't provide.

✅ No documentation drift detected. `persistence.md` was updated in the same PR: the hibernation section now documents the `~/.cache` → pod-local-disk carve-out (check 1), and `Last verified` is bumped to the PR date 2026-07-22 (check 2). The referenced filesystem paths match the page's existing register (it already enumerates `/tmp`, `.platform/`, `.import-staging-*/`), so no volatility leak.

### Verdict
APPROVE — correct, well-documented perf change; the only nit is that `npm` doesn't actually use `~/.cache`, a comment-accuracy fix, not a functional one.

---

PR #2911, opened 2026-07-22
feat(channels): attribute Slack posts to the agent by name, id in the footer link

pilartomas:
## Why

Each Slack channel binds to at most one agent, but a bound agent can post **into other channels** the install-wide bot is a member of (the cross-workspace outbound reach). Because **one install-wide bot posts for every agent**, a bot message's Slack user id can't tell agents apart — so an agent reading a channel had no way to distinguish its *own* past posts from another agent reaching in.

The attribution footer used to show the raw agent id (`agent-8acde0e1a059835a`): ugly for humans, and invisible to agents (it lives in a Slack `context` block, and history injection reads only the message text).

## What

- **Outbound footer → clickable name.** The footer is now the agent's **name** rendered as a link to its UI page (`/sandboxes/<id>`), with the **agent id carried in the URL**. Humans see a clean, clickable name; the stable, unique id rides along for machine use. Applies to the `reply` tool, the proactive `send_channel_message` MCP path, and the attachment-only variant.
- **Inbound relabeling.** When a fresh session injects thread/channel history, the worker parses the agent id out of each footer link and relabels each line:
  - the reading agent's own posts → `you (this agent):`
  - another agent's posts → `<name> (another agent):`
  - humans → their Slack id (unchanged)

  A short legend explaining the prefixes is prepended whenever the injected history contains any agent-authored line.
- **Identity resolved server-side.** The api-server knows the channel's bound agent id, so it decides "this is you" authoritatively — the agent never has to recognize its own (platform) name, which it may not even know (its self-identity comes from its workspace setup).

## Design notes

- New `agent-footer.ts` holds the footer render + its inverse (parse/label), shared by the worker and the turn presenter (no circular import).
- Name lookup (`AgentsService.get(id).name`) is best-effort: a lookup failure or a nameless agent degrades to the id as the (still clickable) link label.
- The footer must be retrievable from `conversations.history`/`replies` blocks — this holds for normal posts; **worth a live-cluster sanity check** for any streamed-message edge cases (fallback would be Slack message `metadata`, parsed the same way; the visible link stays regardless).

## Known limitations (out of scope)

- **Resumed sessions aren't re-injected**, so a foreign agent's mid-thread post is only seen on the next fresh injection.
- **Ambient live path** filters out all bot posts (loop prevention); foreign posts still surface via history injection.

## Tests

- `slack-agent-footer.test.ts` — footer render, name sanitization/fallback, parse round-trip, per-author labeling.
- `slack-cross-agent-history.test.ts` — end-to-end relabeling (self / other agent / human) and legend gating through the worker.
- Full api-server suite green (573 tests); `mise run api-server:check` clean.
- Architecture doc `docs/architecture/channels.md` updated (outbound attribution + inbound labeling).


packages/api-server/src/modules/channels/infrastructure/agent-footer.ts
```diff
@@ -0,0 +1,104 @@
...
+  return {
+    type: "context",
+    elements: [{ type: "mrkdwn", text: agentFooterMrkdwn(footer) }],
+  };
+}
+
+// `<…/sandboxes/<id>|<label>>` — the id is anchored to the path so a name
+// containing "/sandboxes/" can't be mistaken for it (labels sit after the `|`).
+const FOOTER_RE = new RegExp(
+  `<[^>|]*${AGENT_PATH}(agent-[A-Za-z0-9]+)\\|([^>]*)>`,
```
dam-code-guardian, line 57: 🟢 **Suggestion:** `AGENT_PATH` is interpolated into this `new RegExp(...)` unescaped. Safe today (`/sandboxes/` has no regex metacharacters), but latent: if the constant ever gains a `.`, `-`, `?`, etc. the pattern silently changes meaning. Consider a small regex-escape helper on `AGENT_PATH`, or at minimum an inline note pinning the invariant that it must stay regex-literal-safe.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `977ddaf`

## PR #2911: feat(channels): attribute Slack posts to the agent by name, id in the footer link
**Author:** pilartomas | **Branch:** feat/slack-cross-agent-identity → main | **Changes:** +341 −65 (9 files)

### Summary
Makes an install-wide Slack bot's posts attributable per-agent: the outbound footer becomes the agent's name rendered as a link to `/sandboxes/<id>` (id carried in the URL), and inbound history injection parses that id back out to relabel each line — the reading agent's own posts as `you (this agent)`, other agents by name, humans by Slack id — with an explaining legend prepended when any agent-authored line is present. Identity is resolved server-side from the channel's bound agent.

### Findings
- ✅ **Looks good:** The escape/unescape round-trip is correct and adversarial-input-safe: `escapeLinkLabel` drops `|`/newlines (which would break `<url|label>`) and escapes `&`,`<`,`>`; `unescapeLinkLabel` reverses them in the correct order (`&amp;` last), so even a name like `R&D <team>` round-trips exactly — proven by the footer tests.
- ✅ **Looks good:** The `getContextMessages` refactor preserves message ordering. The channel path moved from `filter → reverse` to `reverse → filter → map`; reversing and filtering commute, so the emitted order is unchanged, and the thread path stays chronological as before. The `hasAgentAuthored` flag correctly gates the legend (tested both ways).
- ✅ **Looks good:** `labelHistoryMessage` emits only `label: text` with no URL, so no raw `agent-…` id leaks into the injected prompt (asserted by `not.toContain` in the cross-agent test). Clean separation between the human-facing footer link and the machine-facing label.
- ✅ **Looks good:** The `SlackMessage.blocks` port extension is threaded consistently through both the real (bolt) and fake gateways, and the `setHistory` seam makes the end-to-end relabeling genuinely testable rather than mocked.
- 🟢 **Suggestion:** `FOOTER_RE` interpolates `AGENT_PATH` into `new RegExp(...)` unescaped. It's safe today (`/sandboxes/` has no regex metacharacters), but it's a latent footgun: if the path constant ever gains a `.`, `-`, or `?`, the pattern silently changes meaning. A `RegExp`-escape helper (or an inline comment pinning the invariant "AGENT_PATH must be regex-literal-safe") would make that dependency explicit.
- 🟢 **Suggestion:** `resolveAgentFooter` calls `agents().get(instanceName)` on every outbound post / reply to resolve the display name. That's off the hot turn path and correctly best-effort (failure degrades to the id), so it's fine as-is; if outbound volume grows, a small per-instance name cache would remove the per-post lookup. Non-blocking.

✅ No documentation drift detected. The behavior change is reflected in the two updated `docs/architecture/channels.md` bullets (outbound footer format + the new per-agent inbound attribution), `Last verified:` is bumped to the PR date, and no ADR references or volatile file-path/package details were introduced.

✅ No findings. The new `agent-footer.ts` sits correctly in the `channels` infrastructure layer as a pure build/parse mapper pair (render footer ↔ recover author), introduces no cross-module or cross-layer import violations, and the `SlackMessage.blocks` port extension keeps the real and fake adapters in sync. `getContextMessages` returning a `{ lines, hasAgentAuthored }` record is a clean, honest signature for the two things the caller needs.

### Verdict
APPROVE — Correct, well-scoped, and thoroughly tested (escaping round-trip, ordering preservation, self/other/human labeling, and legend gating all covered); the two suggestions are latent-hardening only.

---

PR #2912, opened 2026-07-22
refactor(controller): build egress gateway proxy config from structured data

### LINKED ISSUE #2899: Build the egress gateway proxy config from structured data instead of string templating
## Problem

Each agent's egress gateway gets its proxy configuration assembled by string-templating text, and values that originate from user or connection input — hostnames, credential header names, and similar — are spliced into that config without escaping. A crafted value could break out of its position in the generated document and inject arbitrary proxy configuration.

The blast radius is limited: the config is per-agent and owner-scoped, and only the agent's owner can supply these values, so this is not a cross-tenant escalation — the worst case is an owner corrupting or crash-looping their own gateway. Today the risk is mitigated by validating individual fields at input (for example, hostnames must match a DNS-name pattern, enforced at both the API front door and the resource boundary). But that mitigation is per-field: it depends on every current *and future* field being validated, and the config template itself remains an unescaped-interpolation footgun for whoever adds the next field.

## Proposed solution

Assemble the gateway proxy configuration as structured data and serialize it with a real encoder, so every string value is quoted and escaped by construction. This removes the whole class of config-injection independent of any upstream validation, and makes every field added later safe by default — no per-field escaping discipline required.

### Scope
- Only how the egress gateway's proxy configuration document is produced.

### Out of scope
- Per-request egress policy evaluation and credential injection behavior — unchanged; this is purely about how the config is rendered.
- The field-level input validation added alongside per-agent L7 promotion stays in place as defense-in-depth.

### Dependencies
- Follow-up to #2865 / #2870, which added the field-level input validation this generalizes.

jezekra1:
## Build the egress gateway proxy config from structured data instead of string templating

Fixes #2899.

### What changed
The gateway's Envoy proxy configuration was produced by string-templating a large YAML document, splicing user- and connection-derived values (hostnames, credential header names, query-param names, Secret-derived identifiers) into it raw. A crafted value could break out of its YAML position and inject arbitrary proxy configuration. Field-level input validation (#2865/#2870) mitigated that per-field, but the template stayed an unescaped-interpolation footgun for whoever added the next field.

This assembles the whole bootstrap document as structured Go data (`map[string]any` / `[]any`) and serializes it with a real YAML encoder (`sigs.k8s.io/yaml`), so **every string value is quoted and escaped by construction**. This removes the config-injection class independent of upstream validation and makes every field added later safe by default — no per-field escaping discipline required.

Topology, routes, credential injection, and telemetry behavior are unchanged; only *how the document is produced*.

- New `envoy_bootstrap.go` assembles listeners, routes, filter chains, credential injectors, the query-param Lua mover, and clusters as structured data.
- Removed the `text/template` constant and template-based renderer from `envoy.go` (−1,180 lines of templated YAML).
- Bumped `envoyBootstrapTemplateRev` → `v15-structured-config` so existing gateways roll onto the re-rendered ConfigMap on upgrade (bounded, per-agent, once).
- Converted format-locked render assertions to structural navigation of the parsed document, and added a test proving hostile host/header values are escaped as opaque scalars rather than injected as config.

### Testing (all via `dam-vm`)
- Full `pkg/reconciler` unit suite, `go vet`, `gofmt`, and `staticcheck`: clean.
- **Real Envoy `v1.37.2 --mode validate`** (the platform's exact image) passes on 4 representative rendered configs: empty, object-store, telemetry, and a `full` config exercising credentialed chains, query-param Lua, HTTP/2, upstream-CA override, allow-only, and the telemetry transit chain (validated with the leaf TLS + SDS files the config references). Confirms the output is bootable Envoy config, not just parseable YAML.
- New security test: hostile host/header values round-trip as escaped scalars; the document stays parseable with no attacker-introduced keys.

### Upgrade note
`envoyBootstrapTemplateRev` bumped, so on upgrade each agent with promoted hosts gets one bounded gateway roll (expected, per the rev's purpose).


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `fcdf789`

## PR #2912: refactor(controller): build egress gateway proxy config from structured data
**Author:** jezekra1 | **Branch:** refactor/egress-gateway-structured-config → main | **Changes:** +1331 −1404 (3 files)

### Summary
Replaces the ~1,180-line `text/template` that string-spliced user/connection-derived values (hostnames, header names, query-param names, Secret identifiers) into the Envoy bootstrap YAML with structured `map[string]any`/`[]any` assembly serialized by `sigs.k8s.io/yaml`. Every string value is now quoted/escaped by the encoder by construction, closing the config-injection class (#2899) independently of upstream field validation. Topology, routes, credential injection, and telemetry are unchanged; only how the document is produced.

### Findings
- ✅ **Looks good:** The core security property is real and structurally enforced — there is no interpolation path left for a value to escape its YAML position, and `TestRenderEnvoyBootstrap_HostileValuesEscapedNotInjected` proves a host/header packed with YAML metacharacters round-trips as an opaque scalar with no attacker-introduced keys.
- ✅ **Looks good:** The Lua query-param mover is safe on two axes — `strconv.Quote` embeds HEADER/PARAM as proper Lua string literals (no Lua-source injection), and the credential value is percent-encoded before landing in the query string so `&`/`=` can't break out of their parameter. The path is parsed manually rather than via `gsub`, avoiding backreference interpretation of credential bytes.
- ✅ **Looks good:** Behavior preservation is well-guarded: the collector-host-collision guard, fork vs. own `instanceID` split for ext-authz-vs-telemetry, `OVERWRITE_IF_EXISTS_OR_ADD` on the trusted `x-platform-agent-id`, upstream SNI + SAN pinning, and the HTTP/2 opt-in are all carried over faithfully, and the tests now assert them structurally (parsed-document navigation) rather than by byte layout — a strictly more robust assertion style.
- 🟢 **Suggestion:** The `envoyBootstrapTemplateRev` bump to `v15-structured-config` triggers one bounded per-agent gateway roll on upgrade. This is intended and documented in the PR body, but since the *output* is byte-for-byte-equivalent config (only the producer changed), it's worth a one-line confirmation in review that a roll on a pure-refactor rev is acceptable operational cost here (it is — the rev's whole purpose is to re-render, and the roll is bounded/once).
- 🟢 **Suggestion:** `type ev = map[string]any` plus deeply-nested literals trades some compile-time type safety for readability. That's a reasonable call for a config-document builder (and the real Envoy `--mode validate` test backstops it), but a future direction worth noting is that a bug in a key name or nesting is caught only by the Envoy-validate test, not the compiler — keeping that validate test in CI is what makes this trade sound.

✅ No documentation drift detected. The `security-and-credentials.md` "Envoy credential injection" section documents the gateway's wire behavior and topology (CONNECT → SNI → per-host chains → injection → pinned STRICT_DNS), all preserved by this PR; it never referenced the string-template implementation, so nothing is stale, and the documentation guidelines correctly keep the file-level "how the document is built" detail out of the architecture page.

### Verdict
APPROVE — A clean, behavior-preserving refactor that eliminates a whole config-injection class by construction, backed by real-Envoy validation and a hostile-input escaping test; no correctness, security, or maintainability concerns.

---

pilartomas (approved):

PR #2926, opened 2026-07-23
feat(ui): move Artifacts last in the left side panel

### LINKED ISSUE #2915: UI - Move Artifacts to last in list in the left side panel
## Problem

The current left navigation order does not reflect the primary workflows users follow in DAM. Artifacts are typically a result of work completed in Sandboxes, Experiments, and Knowledge Bases rather than a starting point.

## Proposed solution

Update the left side panel navigation order to better align with the user workflow:

1. Home
2. Sandboxes
3. Experiments
4. Knowledge Bases
5. Artifacts

This positions Artifacts as a downstream resource users access after creating or generating content through other DAM experiences.

PetrBulanek:
Reorders the left side-panel navigation so it follows the way people actually work in DAM. Artifacts are usually the *result* of work done elsewhere, so they now sit last among the primary destinations rather than near the top.

The primary group now reads:

- **Sandboxes** (the item that was labelled "Home" — same destination, clearer name)
- **Experiments**
- **Artifacts**

Inbox and Settings stay pinned at the bottom, and the DAM logo still takes you home. A spot is reserved for **Knowledge Bases** so it can join the list once that area ships.

The same order applies on mobile.

Closes #2915

### DIFF
--- a/packages/ui/src/components/icon-rail.tsx
+++ b/packages/ui/src/components/icon-rail.tsx
@@ -41,8 +41,8 @@ export function IconRail({
   const { data: features } = useFeatures();
   const showExperiments = features?.experiments ?? false;
 
-  const home: Destination = {
-    label: "Home",
+  const sandboxes: Destination = {
+    label: "Sandboxes",
     icon: Home,
     active: view === "list",
     badge: 0,
@@ -89,7 +89,7 @@ export function IconRail({
         <div className="flex items-center justify-center pt-2">
           <button
             type="button"
-            onClick={home.navigate}
+            onClick={sandboxes.navigate}
             title={getBrand().name}
             aria-label={getBrand().name}
             className="rounded-lg p-1 text-foreground/80 transition-colors hover:bg-muted hover:text-foreground"
@@ -98,9 +98,11 @@ export function IconRail({
           </button>
         </div>
         <div className="flex flex-col items-center gap-1">
-          <RailItem {...home} />
-          <RailItem {...artifacts} />
+          <RailItem {...sandboxes} />
           {showExperiments && <RailItem {...experiments} />}
+          {/* Knowledge Bases nav item slots in here (issue #2915) once the
+              feature ships — under construction. */}
+          <RailItem {...artifacts} />
         </div>
         <div className="flex-1" />
         {/* Inbox is grouped with Settings at the bottom, per the redesign (Figma 152:4567). */}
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/icon-rail.tsx
```diff
@@ -41,8 +41,8 @@ export function IconRail({
   const { data: features } = useFeatures();
   const showExperiments = features?.experiments ?? false;
 
-  const home: Destination = {
-    label: "Home",
+  const sandboxes: Destination = {
+    label: "Sandboxes",
     icon: Home,
```
dam-code-guardian, line 46: 🟢 **Suggestion:** The `Sandboxes` label now pairs with the Carbon `Home` icon. Functionally fine and outside this PR's stated scope, but the icon lags the rename — worth a follow-up if the redesign intends a distinct Sandboxes glyph.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `0498667`

## PR #2926: feat(ui): move Artifacts last in the left side panel
**Author:** PetrBulanek | **Branch:** feat/2915-reorder-left-rail-nav → main | **Changes:** icon-rail.tsx (1 file)

### Summary
Reorders the left navigation rail so the primary group reads Sandboxes → Experiments → Artifacts (Artifacts moved last), renames the former "Home" item to "Sandboxes" (same `list` destination), and reserves a commented slot for a future Knowledge Bases item. The same order is applied to the mobile bottom bar.

### Findings
- 🟢 **Suggestion:** The renamed `Sandboxes` destination still uses the Carbon `Home` icon (`icon-rail.tsx:46`). Functionally fine and out of this PR's stated scope, but the icon now lags the label — worth a follow-up if the redesign intends a distinct Sandboxes glyph.
- ✅ **Looks good:** Rename is complete and consistent — the const, label, and the brand-logo `onClick` (`icon-rail.tsx:92`) all point at the same `list` destination; no dangling `home` references remain.
- ✅ **Looks good:** Desktop rail and mobile bottom bar are kept in sync (identical order, both reserving the Knowledge Bases slot), so the two surfaces won't drift.
- ✅ **Looks good:** The Knowledge Bases placeholder is a comment only — no dead `RailItem` rendered — so nothing ships to users before the feature exists.

✅ No findings. No `docs/architecture/` page describes nav-item ordering or labels; `artifact-library.md`'s statement that Artifacts is "a top-level destination in the navigation rail" remains true after the reorder.

✅ No findings. No new state, no static `style={{}}`, props typed via the `Destination` interface, and the placeholder comments explain *why* a gap is reserved (permitted "why" comments).

### Verdict
APPROVE — Focused, consistent presentational reorder/rename applied symmetrically across desktop and mobile; only a non-blocking icon-vs-label note.

---

kapetr (approved):

PR #2927, opened 2026-07-23
feat(channels): talk to an agent in Slack DMs and group DMs (#2832)

### LINKED ISSUE #2832: Talk to an agent in Slack direct messages and group DMs
## Problem

Agents can only be reached inside channels, by `@mention`. There's no way to have a private one-on-one conversation with an agent, and no way to pull one into a small group DM. Opening the agent's DM in Slack today doesn't even get you a chance to try — Slack blocks it outright with *"Sending messages to this app has been turned off."*

This matters because a lot of real work happens in DMs and small group DMs:

- Someone with a **private or sensitive** ask (a personal task, a question they don't want broadcast) has no option — every message to an agent is visible to a whole channel.
- Forcing a channel to exist for interactions that are inherently personal is awkward and discourages use.
- An agent has **no way to reach one person directly** — it can only post into its bound channel, so anything meant for a single user lands in shared view (or nowhere).
- Telegram already supports direct messages, so Slack users get a noticeably worse experience for the same product.

## Proposed solution

Let people hold private and small-group conversations with agents over Slack, and let agents reach a person directly.

- **1:1 DM** — open a direct message with the bot and talk to an agent privately. Follow-ups continue the same conversation, the way a channel thread does today.
- **Group DM** — `@mention` the agent inside a small multi-person DM to bring it into that conversation.
- **Agent-initiated DM** — an agent can send a direct message to a specific person (to follow up, notify, or answer privately), opening the conversation if one doesn't exist yet.
- **Authorized and revocable** — a clear, controlled way to establish that a given person may talk to a given agent over DM, and to revoke it. Unauthorized DMs are declined with a clear message, not silently dropped.

Delivering this also means the app must actually **accept direct messages** — today that surface is turned off.

### Open question — which agent answers a DM?

In a channel, the binding decides which agent responds. A DM has no channel, so there needs to be a rule for which agent a given person's direct message reaches. The Slack bot serves the whole workspace (unlike Telegram, where each bot is one agent), so this needs a deliberate answer and should be settled before implementation. A sensible delivery order is agent-initiated DMs first (useful on its own), then the routing/authorization rule, then inbound DMs on top of it.

### Scope

- One-on-one DMs and small group DMs between people and an agent.
- Establishing, enforcing, and revoking DM authorization.
- Agents initiating a DM to an authorized person.

### Out of scope

- Changing how channel conversations work today.

### Dependencies

- Subissue of the Slack rework (#2830).
- Consolidates and supersedes #30 (DMs + group DMs) and #603 (private 1:1 DMs) — both will be closed as absorbed here.

pilartomas:
Closes #2832.

## What

Extends the existing Slack channel-binding mechanics to **1:1 DMs** (`im`) and **group DMs** (`mpim`). The insight: a Slack binding already keys purely on the conversation id, so DMs and group DMs reuse the same `channels` table, the same resolution, and the same in-chat bind flow as channels — no new binding primitive.

It also **enables the app to receive DMs at all** — today Slack shows *"Sending messages to this app has been turned off"* (see the issue's screenshot) because the App Home messages tab is off and there's no `message.im` subscription.

## Behaviour

- **1:1 DM** — once bound, every plain message relays to the agent (no `@mention` needed, since every DM message is addressed to the bot). The prompt is left un-speaker-labelled (a single human).
- **Group DM** — stays mention-driven like a channel: `@mention` the agent. Plain (non-mention) group-DM messages are ignored.
- **Unbound DM/group** — declined with an ephemeral pointing at the in-chat bind command.
- **Binding is explicit** (per product decision): an agent-initiated DM still posts as a notification (that already worked), but a person's replies reach the agent only once the DM is bound via the in-chat bind command. No auto-bind.
- A DM's `message.im` is the source of truth, so a duplicate `app_mention` in a DM is dropped to avoid processing the turn twice.

Authorization reuses the **shared-binding** model (the bind is the consent grant; runs under the agent's own credentials), exactly as the in-chat channel bind already does.

## Changes

- `deploy/slack-app-manifest.yaml` — App Home messages tab on (not read-only); subscribe `message.im`; add `im:history` / `mpim:history` / `mpim:read`; slash-command hint mentions `bind`.
- `slack-gateway.ts` / `bolt-slack-gateway.ts` / `fake-slack-gateway.ts` — carry `channelType`; add an `onDirectMessage` handler; route `im` messages to it (group DMs stay on `app_mention`).
- `slack.ts` — `handleAppMention` refactored into a shared `handleInbound`; new `handleDirectMessage`; DM/group-aware unbound copy; single-speaker DMs skip the speaker label; DM-friendly bind copy and conversation label.
- `agents-service.ts` — bind confirmation reads naturally in a DM.
- `docs/architecture/channels.md` — updated (source of truth).
- `slack-dm.test.ts` — new coverage (bound/unbound DM, group-DM mention, DM bind, `app_mention` dedup).

## Deployment note

The Slack app must be **reinstalled from the updated manifest** for the new scopes, the `message.im` subscription, and the messages tab to take effect — none of it works until then.

## Known limitations / follow-ups

- An agent can be bound to **one Slack conversation at a time** (channel *or* DM *or* group) — the existing 1:1 agent↔conversation constraint (`channels_agent_type_idx`), unchanged here. Relaxing it (multiple conversations per agent) is a separate schema change.
- Unbound **group-DM** copy uses group wording only when Slack includes `channel_type` on the `app_mention` payload (sometimes omitted); otherwise it falls back to the generic channel copy. Cosmetic — no auth impact.

## Verification

`tsc` clean, ESLint clean, Prettier clean, full api-server unit suite green (592 tests, incl. new `slack-dm.test.ts`). Diff was adversarially reviewed for auth holes / loops / regressions. The `controller:check:crd-backwards-compatibility` gate fails locally with "reference not found" — a known git-worktree false-positive (this PR touches no CRD/controller files).


packages/api-server/src/modules/agents/services/agents-service.ts
```diff
@@ -346,10 +346,15 @@ export function executeSlackBind(deps: {
       },
     });
 
+    // A 1:1 DM conversation id starts with "D" — tailor the confirmation so a
+    // private DM doesn't read as a shared channel ("everyone here").
+    const isDm = flow.slackChannelId.startsWith("D");
```
dam-code-guardian, line 351: 🟢 **Suggestion:** this `startsWith("D")` DM-id check duplicates the `isDirectMessageId` helper in `slack.ts`. They can't share cleanly across module boundaries, so an inline literal is acceptable here — a short comment cross-referencing the gateway helper would just help keep the two in sync if Slack's DM id convention ever changes.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `e6a391fc`

## PR #2927: feat(channels): talk to an agent in Slack DMs and group DMs (#2832)
**Author:** pilartomas | **Branch:** feat/slack-dm-binding → main | **Changes:** +374 −36 (8 files)

### Summary
Extends the existing Slack channel-binding mechanics to 1:1 DMs (`im`) and group DMs (`mpim`), reusing the same `channels` table, resolution, and in-chat bind flow since the binding keys purely on conversation id. Enables inbound DMs (App Home messages tab + `message.im` subscription) and refactors `handleAppMention` into a shared `handleInbound` that a 1:1 DM enters unlabelled while channels/group DMs stay speaker-labelled and mention-driven.

### Findings
- ✅ **Looks good:** DM/app_mention dedup is sound — a bound DM relays via `message.im` (routed before the bot-mention drop in `bolt-slack-gateway.ts:443`), while the duplicate `app_mention` is dropped by the `isDirectMessageId` guard in `handleAppMention` (`slack.ts:1299`); exactly one turn is processed, and the new test covers it.
- ✅ **Looks good:** Loop safety preserved — `msg.bot_id || !msg.user` is filtered before the `im` route (`bolt-slack-gateway.ts:424`), so the agent's own DM posts and agent-initiated DMs never re-enter.
- ✅ **Looks good:** `mpim` plain messages correctly fall through to no handler (only `im` → `onDirectMessage`, `channel`/`group` → `onMessage`), keeping group DMs mention-driven as documented.
- ✅ **Looks good:** `speakerLabel` threads cleanly — `handleInbound` passes `!opts.directMessage`, and `relaySharedTurn` defaults to labelled (`args.speakerLabel === false ? …`), so only the single-human 1:1 DM prompt goes unlabelled.
- 🟢 **Suggestion:** the `startsWith("D")` DM-id test is duplicated as an inline literal in the service layer (`agents-service.ts:351`) versus the `isDirectMessageId` helper in the gateway (`slack.ts:151`). Cross-module reuse isn't clean here, so a shared literal isn't warranted — but a brief cross-reference comment would help future readers keep the two in sync.

✅ No findings. `docs/architecture/channels.md` is updated in the same PR (Overview, Message intake, Agent resolution, the new DMs/group-DMs bullet, outbound and persistence sections) and `Last verified:` is bumped to 2026-07-23 (PR date). No ADR leaks, no volatile-content additions beyond the page's established Slack-API vocabulary, no new subsystem page needed.

✅ No findings. Changes sit correctly in the infrastructure layer — the new `onDirectMessage` port is declared on `SlackGatewayHandlers` and implemented by both the real (`bolt`) and `fake` gateways (proper ports & adapters), and the worker refactor keeps the single `handleInbound` intake. The one services-layer edit (`agents-service.ts`) is a copy tweak only. No layer-boundary or dependency-direction violations.

### Verdict
APPROVE — a well-scoped, well-tested reuse of the existing binding model; dedup, loop-safety, and speaker-labelling are all correct, and the architecture doc is updated in the same PR.

---

PR #2928, opened 2026-07-23
feat(ui): show the exact date and time on every session in the list

### LINKED ISSUE #2901: Show the exact date and time on every session in the list
## Problem

Each session in the list shows a relative label — *Today*, *Yesterday*, *Last week*, *Last 30 days* — instead of an actual date and time. Several of us have hit the same wall: these labels are too vague to find a specific session.

- *Today* and *Yesterday* don't distinguish between multiple sessions from the same day — they all read identically.
- *Last week* / *Last 30 days* don't name a day at all, so "the thing I worked on Monday" is impossible to spot.
- No time of day is shown anywhere, so even once you're in the right day you can't order or identify sessions.

The result is that the list — whose whole job is to get you back to a past session — can't reliably do that.

## Proposed solution

Show a full, exact timestamp on **every** row, in one consistent format: **`MM/DD/YYYY HH:MM`**. Same format for recent and old sessions alike; drop the relative labels entirely.

### Alternative

A lighter variant: keep *Today* and *Yesterday* as relative labels (the two everyone reads at a glance), and show **every other** session as the full `MM/DD/YYYY HH:MM`. Trades a bit of consistency for faster recognition of the most recent sessions, while still making everything older precise.

### Context

The current relative-date behavior was a deliberate choice in the chat-shell design (#883), not a bug — so this issue intentionally revisits that decision based on team feedback.

PetrBulanek:
Every session in the list now shows its exact date and time — `MM/DD/YYYY HH:MM`, e.g. `07/22/2026 14:32` — in one consistent format for recent and old sessions alike.

Previously each row showed a vague relative label (*Today*, *Yesterday*, *Last week*, *Last 30 days*), which couldn't tell apart multiple sessions from the same day, never named a day for anything older, and showed no time. Now you can reliably identify and order sessions and jump back to a specific one.

Closes #2901

### DIFF
--- a/packages/ui/src/modules/sessions/lib/format-session-timestamp.ts
+++ b/packages/ui/src/modules/sessions/lib/format-session-timestamp.ts
@@ -1,5 +1,6 @@
 export function formatSessionTimestamp(iso: string): string {
   const d = new Date(iso);
+  if (Number.isNaN(d.getTime())) return "—";
   const p = (n: number) => String(n).padStart(2, "0");
   return `${p(d.getMonth() + 1)}/${p(d.getDate())}/${d.getFullYear()} ${p(d.getHours())}:${p(d.getMinutes())}`;
 }
--- a/packages/ui/src/modules/sessions/components/session-row.tsx
+++ b/packages/ui/src/modules/sessions/components/session-row.tsx
@@ -17,7 +17,7 @@ import {
 } from "@/components/ui/dropdown-menu";
 import { cn } from "@/lib/utils";
 
-import { formatSessionDate } from "../lib/format-session-date.js";
+import { formatSessionTimestamp } from "../lib/format-session-date.js";
 import { WorkingDots } from "./working-dots.js";
 
 const LONG_PRESS_MS = 400;
@@ -130,11 +130,8 @@ export function SessionRow({
             working={working}
           />
         </div>
-        <span
-          className="text-[11px] text-muted-foreground"
-          title={new Date(s.updatedAt ?? s.createdAt).toLocaleString()}
-        >
-          {formatSessionDate(s.updatedAt ?? s.createdAt)}
+        <span className="text-[11px] text-muted-foreground">
+          {formatSessionTimestamp(s.updatedAt ?? s.createdAt)}
         </span>
       </div>
       {/* Desktop: hover-visible overflow menu */}
--- a/packages/ui/src/modules/sessions/components/session-row.tsx
+++ b/packages/ui/src/modules/sessions/components/session-row.tsx
@@ -17,7 +17,7 @@ import {
 } from "@/components/ui/dropdown-menu";
 import { cn } from "@/lib/utils";
 
-import { formatSessionTimestamp } from "../lib/format-session-date.js";
+import { formatSessionTimestamp } from "../lib/format-session-timestamp.js";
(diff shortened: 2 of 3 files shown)


packages/ui/src/modules/sessions/lib/format-session-timestamp.ts
```diff
@@ -1,15 +1,5 @@
...
-export function formatSessionDate(iso: string, now: Date = new Date()): string {
-  const date = new Date(iso);
-  const dayDiff = Math.round((startOfDay(now) - startOfDay(date)) / DAY_MS);
-  if (dayDiff <= 0) return "Today";
-  if (dayDiff === 1) return "Yesterday";
-  if (dayDiff <= 7) return "Last week";
-  if (dayDiff <= 30) return "Last 30 days";
-  return date.toLocaleDateString();
+export function formatSessionTimestamp(iso: string): string {
+  const d = new Date(iso);
```
dam-code-guardian, line 2: 🟢 **Suggestion:** An unparseable or absent `iso` produces `Invalid Date`, which renders literally as `NaN/NaN/NaN NaN:NaN`. `createdAt` is normally always present, so this is an edge case — but a small guard keeps it graceful:

```suggestion
  const d = new Date(iso);
  if (Number.isNaN(d.getTime())) return "—";
```

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `505fc31`

## PR #2928: Show the exact date and time on every session in the list
**Author:** PetrBulanek | **Branch:** feat/2901-session-list-timestamps → main | **Changes:** session-row.tsx, format-session-date.ts

### Summary
Replaces the relative session-date labels (*Today / Yesterday / Last week / Last 30 days*) with an explicit `MM/DD/YYYY HH:MM` timestamp rendered directly in each session row, so multiple sessions from the same day are distinguishable and orderable.

### Findings
- 🟢 **Suggestion:** The file is still named `format-session-date.ts` but now exports `formatSessionTimestamp` and formats a full timestamp — renaming the file to `format-session-timestamp.ts` (and its import) would keep name and content aligned. (`packages/ui/src/modules/sessions/lib/format-session-date.ts:1`)
- 🟢 **Suggestion:** An unparseable/absent `iso` yields `Invalid Date`, rendering literally as `NaN/NaN/NaN NaN:NaN`. `createdAt` is normally always present so this is an edge case, but a guard (`Number.isNaN(d.getTime())` → fallback dash) would fail gracefully. (`packages/ui/src/modules/sessions/lib/format-session-date.ts:2`)
- ✅ **Looks good:** The old `title` hover tooltip (`toLocaleString()`) is correctly dropped — the full timestamp is now always visible, so the tooltip was redundant. The fixed `MM/DD/YYYY` format (over locale-aware formatting) is an explicit, documented design choice per the PR description.

✅ No findings. The change is purely presentational; no `docs/architecture/` page describes session-list date formatting, and no subsystem behavior/responsibility changed.

✅ No findings. The helper is a pure formatting function correctly placed in the sessions module's `lib/`.

✅ No findings. The diff is a minimal import swap plus an inline formatter call in the row's timestamp span — no new state, styling, or structural concerns.

### Verdict
APPROVE — Small, correct presentational change that meets its stated goal; only minor non-blocking naming/edge-case suggestions.

---

kapetr (approved):

PR #2941, opened 2026-07-23
feat(channels): default Slack in-chat bind to ambient on

pilartomas:
Streamlines the Slack ambient-mode UX in two ways.

## 1. Default the in-chat `/bind` to ambient on

When a Slack user runs `/platform bind` (`/dam bind`) and connects an agent, the binding was created in **shared** mode with **ambient off** — so the binder had to run a second command (`/dam ambient on`) before the agent would read along. Now the in-chat channel bind defaults ambient **on**, so it's a single step.

- **In-chat `/bind` of a channel now defaults ambient on.** `executeSlackBind` passes `ambient=true` to the shared connect for channel binds.
- **1:1 DM binds stay ambient-off** — a bound DM already relays every message, so ambient is meaningless there.
- **UI form and CLI flag are unchanged** — both already expose an explicit ambient toggle (default off / opt-in).

## 2. Ambient changes are never announced in the channel

Previously, turning ambient on/off posted a channel-visible notice ("X is now in ambient mode…" / "X left ambient mode"). That's gone on every surface:

- **In-chat `/ambient on|off`** replies only to the person who ran the command (the ephemeral slash-command reply), which carries the full description the channel post used to.
- **UI / CLI connect, re-connect, or toggle** posts nothing to the channel (`connectSlackImpl` no longer announces).
- **The in-chat bind confirmation** is a plain connect notice ("This channel is now connected to X…") with no ambient sentence.

In all cases the change is still recorded in the security log (`channel.ambient_toggled`); only the channel-visible post is removed. The operator sees confirmation on whatever surface they used.

## Docs

`docs/architecture/channels.md`: the per-entry-point ambient default (in-chat channel bind → on; UI/CLI → off; DM → off), and that ambient enable/disable is audited but deliberately never announced in the channel.

## Tests

- `slack-bind.test.ts`: channel bind asserts `connectShared(..., true)` and a plain confirmation (no ambient copy); DM case asserts `connectShared(..., false)`.
- `slack-ambient.test.ts`: the `/ambient on|off` command tests assert the description rides the invoker's ack and nothing is posted to the channel; audit retained.
- `10-slack-ambient.spec.ts` (e2e): asserts connecting with ambient posts no announcement to the channel.
- `api-server` unit suite green (605 tests); `api-server` + `e2e-playwright` tsc/eslint/prettier all pass.


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `b83f7e2`

## PR #2941: feat(channels): default Slack in-chat bind to ambient on
**Author:** pilartomas | **Branch:** feat/ambient-default-on-bind → main | **Changes:** +69 −80 (6 files)

### Summary
Streamlines Slack ambient UX: an in-chat `/bind` of a channel now defaults ambient **on** (1:1 DMs stay off), and ambient enable/disable is no longer announced channel-visibly — the description rides the invoker's ephemeral reply instead, with the security-log audit retained.

### Findings
- ✅ **Looks good:** `isDm` is now declared once and reused for both the `connectShared(..., !isDm)` ambient default and the confirmation-copy branch — no duplicate `startsWith("D")` (`agents-service.ts:330`).
- ✅ **Looks good:** `connectShared` gains a **required** `ambient` param threaded through to `connectSlackImpl`; the capability isn't modeled as an optional dep. Ambient is still re-derived server-side (`requestedMode === "shared" && ambient === true`) so a direct caller can't mint an ambient person-scoped binding (`agents-service.ts:662`).
- ✅ **Looks good:** Only 1:1 DMs (`D…`) are forced ambient-off; group DMs (`mpim`) correctly keep the channel-like default, matching the documented per-entry-point behavior in `channels.md`.
- ✅ **Looks good:** The audit trail is preserved on every surface — the `channel.ambient_toggled` security-log record remains; only the channel-visible `postMessage` is removed. `requestedAmbient` still feeds channel storage, so it is not dead.
- ✅ **Looks good:** `docs/architecture/channels.md` is updated in the same PR (per-entry-point default + the never-announce rule) with `Last verified:` bumped to the PR date; tests, e2e, and the in-chat ack copy are all updated consistently.
- 🟢 **Suggestion:** `flow.slackChannelId.startsWith("D")` is the DM discriminator in two places now (bind here, and the confirmation copy). Both are correct, but a shared `isDmConversation(id)` helper would keep the "what counts as a DM" rule in one spot if a third caller appears (`agents-service.ts:330`).
- 🟢 **Suggestion:** The enable-ack copy is duplicated almost verbatim between the in-chat ack (`slack.ts`) and the on-toggle description; not worth blocking, but a shared template would prevent the two from drifting (`slack.ts:1152`).

Verdict: aligned. The behavior change is reflected in `docs/architecture/channels.md` (the correct subsystem page) in the same PR; `Last verified:` bumped to 2026-07-23; no ADR references or volatile content introduced.

✅ No architecture-alignment findings. The new `ambient` dependency is required (not optional), wiring stays at the service composition point, and the ambient⇒shared invariant is enforced in the service rather than trusted from the caller.

### Verdict
APPROVE — A focused UX change with the behavior contract enforced server-side, the audit trail intact, docs updated in-PR, and thorough unit + e2e coverage of both the new default and the no-announce guarantee.

---

PR #2851, opened 2026-07-20
feat(agents): per-agent template upgrade to the current image

jezekra1:
Part of #1077 (v1: detection + manual per-agent upgrade).

## What

When an agent's template moves on (a helm upgrade shipping a newer agent image), existing agents keep the image they captured at create time — invisibly and, until now, unfixably. This PR adds:

- **Detection**: every agent read compares the captured `spec.image` against the current template's image (templates are boot-loaded, so it's a memory lookup; `list()` resolves each distinct template once). Surfaced as `templateUpdate: { fromImage, toImage }` on the agent projection.
- **`agents.upgrade` mutation**: re-points `spec.image` at the template's current image via merge-patch. The reconciler rolls a running pair onto the new image; a hibernated agent picks it up on next wake. Idempotent when current; typed `AgentNotFound` / `TemplateNotFound` / `TemplateMoved` errors.
- **Binding consent**: the mutation carries `expectedToImage` — the server compare-and-swaps against the template's current image and rejects with `CONFLICT` if the template moved after the user reviewed the diff.
- **UI**: "Update available" badge on the sandbox header; banner under the Image field showing the incoming image, with a confirm dialog spelling out the exact from→to movement and the restart consequence.

The image ref doubles as the template's version identity: templates render `repository:tag` with the tag defaulting to the chart appVersion, so any create-time capture that differs from the current template means the template advanced. (Pinning a template's `tag` across releases opts it out of detection — documented in `values.yaml`.)

## Deliberately image-only (v1)

- Template env stays frozen in `agent_env` — re-seeding would clobber user edits.
- Mount/storage changes can't ride a spec patch (immutable StatefulSet volume layout) — those still mean recreate.
- No bulk upgrade, no automatic upgrade. Semver-gated auto-upgrade for patch versions is the planned follow-up per the issue's edit.

## Verification

- `mise run check` green; 11 new unit tests on the compare + upgrade flow (idempotency, not-found paths, compare-and-swap mismatch); all package suites pass.
- K8s leg driven live on a cluster: merge-patched `spec.image` on a running Agent CR (the exact operation the endpoint issues) → controller rolled the pod onto the new image and republished `Ready/AllPodsReady` in ~20s.
- Architecture page `docs/architecture/agent-lifecycle.md` updated in the same PR (drift rule).

## Not covered

- e2e Playwright spec for badge → upgrade → cleared (needs mock-template mutation plumbing; follow-up together with semver auto-upgrade).

Issue #1077, opened 2026-06-16
Let users upgrade an existing agent to the latest template version


jezekra1:
## Problem

When you create an agent, it captures everything about its template at that moment — the image it runs, its environment, its mounts, its resource limits — and freezes that into the agent. After that, the agent is on its own copy.

When the operator later advances a template (a newer agent image, a changed default environment, new mounts, different resource limits) and rolls out the platform update, **existing agents do not pick up any of it.** They keep running the version they captured when they were created. Only agents created *after* the update get the new definition.

Today this gap is invisible and unfixable from the product:

- **Nothing tells the user their agent is behind.** The agent screen shows which template an agent came from, but not whether that template has since moved on. An agent can run a months-old image indefinitely with no indication anything newer exists.
- **There's no way to catch up.** Short of deleting the agent and recreating it from scratch — losing its identity, history, and any in-place setup — there is no action a user can take to move an agent onto the current template.

This matters most when a template advance carries a security fix or an important capability change. The operator ships it, believes the fleet is updated, but every pre-existing agent silently stays on the old version.

## Proposed solution

Two user-visible pieces:

1. **Show when an agent is behind its template.** On the agent screen, surface a clear indicator when the agent's captured configuration no longer matches the current version of the template it came from — e.g. an "update available" badge.

2. **An "Upgrade to latest version" action.** A button that re-applies the current template definition to the existing agent in place — adopting the new image and other template-derived settings — without deleting and recreating it, so the agent keeps its identity and history.

### Transparency / safety

- Before upgrading, the user should be able to see *what* will change (e.g. "image moves from X to Y"), since an upgrade can swap the underlying image and restart the agent.
- Upgrading should be an explicit per-agent choice, not silent/automatic — the user opts in when they're ready for the restart.

### Out of scope

- **Bulk / fleet-wide upgrade** of every stale agent at once — start with per-agent; revisit batch upgrade later if needed.
- **Automatic upgrade on template advance** — this issue is about making the manual path possible and visible first.
- **Preserving user-customized fields** that diverge from the template (e.g. a hand-edited environment): the exact merge/override behavior is a design question for implementation, not specified here.

<img width="1020" height="198" alt="Image" src="https://github.com/user-attachments/assets/45075346-c74a-443d-872c-a7c08b34602d" />

EDIT: We should have semver and auto-upgrade for patch versions and require user action only for breaking changes

jenna-winkler:
@jezekra1 closing as i believe this is solved, please reopen if not

PR #2944, opened 2026-07-23
fix(agents): ship harness mise pins in the image, heal stale workspace pins on boot

jezekra1:
## Problem

After a template upgrade (#2851 / #1077), the upgraded agent's harness no longer launches:

```
mise ERROR Failed to install npm:@anthropic-ai/claude-code@2.1.173:
failed create_dir_all: /usr/local/share/mise/installs/npm-anthropic-ai-claude-code/2.1.173: Permission denied
```

The harness version pin lived in a mise config **seeded into the workspace** (`~/.config/mise/config.toml` + `mise.lock`) on first boot, i.e. in persistent volume state — while the actual installs live in the image. A template upgrade rolls the pod onto the new image, which only ships the *new* harness version, but the persisted user-level pin still demands the old one and outranks everything in the image. Every shim call then tries to auto-install the missing old version at runtime and dies: the tool's install dir is root-owned (created by the root-run build `mise install`), and a runtime install would be unpinned and egress-dependent even if it succeeded.

`claude-code`, `codex`, and `pi-agent` all seeded pins this way; every already-upgraded agent is broken like this, and every future image bump would re-break agents on upgrade. `bob`/`k-search` don't pin harness tools via mise and are unaffected.

## Fix

- **Pins ride the image**: each harness commits a `harness-tools.toml` that the Dockerfile copies to `/etc/mise/conf.d/harness-tools.toml` (system-level mise config). Image and pin can no longer diverge; version bumps now edit this one file. The workspace `.config/mise` seed is gone.
- **Boot-time heal** in `agent-entrypoint`: strips exactly the image-owned tool entries (parsed from `/etc/mise/conf.d/*.toml`) from a persisted `~/.config/mise/config.toml` + `mise.lock`, healing volumes seeded by older images on their next boot. User-added tools (`mise use -g …`) are deliberately untouched — that's why the file isn't simply overwritten. A heal failure never blocks boot.
- **Ownership fixups**: the harness install layer re-chowns `/etc/mise` and the harness tool install dirs to the agent user and restores world-readable locks, so a *deliberate* user (re)install of a harness tool doesn't hit the root-owned-dir wall either.

Layering note: user config still outranks system conf.d, but the boot heal removes user-level pins for image-owned tools, making the image's harness version authoritative — intentional, since the harness/ACP adapter pair is platform-owned and must match the image.

## Verification

- entrypoint heal exercised standalone against a stale-seeded home: strips both harness pins from config + lock, keeps user-added `npm:left-pad`/`node` entries and `[settings]`; no-ops on fresh homes, empty/missing `conf.d`
- `platform-base` and `claude-code` images build cleanly — the build itself exercises the new conf.d-resolved `mise install` + postinstall path
- `mise run check` green

**Immediate unblock** for an already-broken agent (before this ships): `sed -i 's/2\.1\.173/2.1.210/' ~/.config/mise/config.toml ~/.config/mise/mise.lock` in the sandbox.


pilartomas (approved):

packages/platform-base/entrypoint.sh
```diff
@@ -36,6 +36,35 @@ if [ ! -L "$home/.cache" ]; then
 	ln -sfn /tmp/agent-cache "$home/.cache"
 fi
 
+# Harness tool pins ship in the image's system-level mise config
+# (/etc/mise/conf.d/harness-tools.toml) so they always match the installs
+# baked into the image. Images predating this seeded the pins into
+# ~/.config/mise/ instead; that copy persists across template upgrades,
+# outranks the system config, and after an upgrade demands a version the new
+# image no longer ships — every shim call then fails trying to install it at
```
dam-code-guardian, line 44: 🟢 **Suggestion:** The tool-key parser here and the lock-block deleter below assume the `harness-tools.toml` format contract (`[tools]`-only, one `"key" = "version"` per line), which is documented in the `harness-tools.toml` headers but not here. If a future conf.d file adds a non-`[tools]` `key = value` line, this parser would treat the key as a tool and strip it from the user's config. A one-line comment pointing at the format contract would harden this against that drift.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `c9f9155`

### Summary
Moves harness-tool version pins (claude-code/codex/pi-agent + their ACP adapters) out of a workspace-seeded user-level mise config and into image-level system config (`/etc/mise/conf.d/harness-tools.toml`), so the pins can no longer diverge from the versions baked into the image on a template upgrade. Adds a boot-time heal in `platform-base/entrypoint.sh` that strips exactly the image-owned tool pins from any already-persisted `~/.config/mise/{config.toml,mise.lock}`, plus Dockerfile ownership fixups so a deliberate user reinstall of a harness tool doesn't hit the root-owned-dir wall.

### Findings
- ✅ **Looks good:** The heal correctly targets `$HOME/.config/mise/`, which is exactly where the removed `/app/working-dir/.config/mise/` seed lands on the persistent volume — the fix reaches the stale state it's meant to heal. Verified against a simulated stale home: both harness pins stripped from config + lock, while `[settings]`, user-added `node`, and `npm:left-pad` survive untouched.
- ✅ **Looks good:** Heal scoping is safe — it reads only `/etc/mise/conf.d/*.toml`, and the base image ships its lazy tools to `/etc/mise/config.toml` (not conf.d), so a user's pins for base tools can never be stripped. Only the harness tools per image are eligible.
- ✅ **Looks good:** The `sed` range delete for lock blocks handles the trailing-block edge case — when the harness block is last with no trailing blank line, the range runs to EOF and still removes it cleanly.
- ✅ **Looks good:** Dropping `chown -R 65532:0 /app` is matched by removing the `/app` mise seed COPY; the base image already chowns `/app`, and the remaining `workspace/` → `/app/working-dir/` COPYs carry `--chown=65532:0`. The `.lock` chmod 644 restores world-readability that mise otherwise strips to 0600.
- ✅ **Looks good:** `[settings]` (lockfile, `minimum_release_age = "7d"`) is preserved via the base `/etc/mise/config.toml`; the new `harness-tools.toml` is `[tools]`-only and conf.d merges, so lockfile/release-age behavior is unchanged.
- 🟢 **Suggestion:** The heal's tool-key parser and lock-block deleter assume the `harness-tools.toml` format contract (`[tools]`-only, one `"key" = "version"` per line). The contract is documented in each `harness-tools.toml` header, but the parser lives in `entrypoint.sh` (a different package). If a future edit adds a non-`[tools]` `key = value` line to a conf.d file, the parser would treat the key as a tool and strip it from the user config. Low risk given the comment, but a one-line note in `entrypoint.sh` pointing at the contract would harden it. (`packages/platform-base/entrypoint.sh:44`)

**Verdict: aligned.** No architecture-doc drift. No page under `docs/architecture/` was edited; the change is implementation-level. `persistence.md:113` already documents the governing invariant ("Tools and dependencies the agent relies on must be baked into the image at build time") which this PR enforces. mise references elsewhere concern the on-PVC cache, which is unaffected. No new subsystem, no ADR state change.

### Verdict
**APPROVE** — Correct, well-scoped fix that eliminates a real upgrade-time breakage class; the heal logic is safe and verified against stale-home and edge-case inputs, and the only note is a minor hardening suggestion.

---

PR #2945, opened 2026-07-23
feat(keycloak-theme): make w3id login the primary CTA on SSO deployments

### LINKED ISSUE #1901: UI - Make w3id login the primary CTA on the login screen
## Problem

On our DAM deployments, the login screen presents the non-w3 login as the primary action and w3id login as secondary. For these deployments we expect folks to sign in with w3id, so the non-w3 option is the wrong default and the extra choice adds cognitive overhead — users have to scan past the prominent option to find the one they actually need.

## Proposed solution

Make w3id login the primary CTA. Two possible solves:

- **Remove the non-w3 login entirely** on our DAM deployments, leaving w3id as the only option.
- **Move w3id to the top / primary position** and demote the non-w3 login.

### Open question

Why would we feasibly need the non-w3 login on these deployments at all? If there's no real need, removing it is the cleaner fix.

## Figma

https://www.figma.com/design/zNIYydUKN1QLZDYozpQJpn/DAM-DEV?node-id=1100-1086

kapetr:
Closes #1901

## What

Redesigns the Keycloak login page and makes the sign-in method a per-deployment choice:

- `keycloak.login.allowPassword: false` → SSO-only page: "Continue with w3id" CTA(s) plus an optional "Request access" link (`keycloak.login.requestAccessUrl`), no password form.
- Default (`true`) → password form with any identity-provider buttons below it; dev/e2e/local unchanged.
- Both share the new single-column layout (eyebrow, light heading, description with docs link); marketing column and aurora backdrop removed.

Mechanism: single theme — the values reach the Keycloak pod as env vars and the theme reads them at runtime (keycloakify `environmentVariables`). No second theme, no realm auth-flow changes; ADR-054 deferred this exact question. Presentation only: password auth stays enabled on the realm (direct grants/break-glass unaffected), and misconfiguration fails open to the password form.

## 🚀 Switching a deployment to SSO-only (after merge)

1. **Upgrade to a release containing this change** — the theme is baked into the Keycloak image, so the deployment must run the new image (comes with the chart's appVersion).
2. **Make sure the realm has the w3id identity provider configured** — IdPs are realm config managed outside the chart (admin console/API). The IdP alias `w3id` gets the IBM logo on the button; the label is "Continue with <displayName>". Without any IdP the page keeps showing the password form (deliberate no-lockout fallback).
3. **Set the values in the deployment's Helm override** and `helm upgrade`:
   ```yaml
   keycloak:
     login:
       allowPassword: false
       requestAccessUrl: "https://…"   # optional; omit to hide the link
   ```
   The env change rolls the Keycloak pod automatically — no realm reimport, no manual steps. Reverting is the same values change back.

## Testing

- `mise run check` (incl. helm lint/render, both `--set` combinations verified on the rendered Deployment)
- Verified on the dev cluster with a mock w3id OIDC IdP: SSO-only page renders per Figma; the CTA brokers to the IdP and back through first-broker-login into an authenticated UI session; password mode keeps `#username`/`#password`/"Sign In" (e2e selectors intact)
- Local preview: `PLATFORM_ALLOW_PASSWORD=false PLATFORM_REQUEST_ACCESS_URL=… mise run keycloak-theme:run -- --keycloak-version 26.7.0`


packages/keycloak-theme/src/login/Template.tsx
```diff
@@ -51,52 +50,32 @@ export default function Template(props: TemplateProps<KcContext, I18n>) {
...
-          {/* Form column */}
-          <div className="flex md:w-1/2">
-            <div className="w-full max-w-sm space-y-6">
-              <h1 className="text-2xl font-semibold leading-none tracking-tight">
-                {headerNode}
-              </h1>
+          {showMessage && (
+            <div
+              role="alert"
+              className={
```
dam-code-guardian: 🟢 **Suggestion:** This conditional picks between two full Tailwind strings that differ only by color family. The React styling rule ([HIGH]) prefers composing conditional classes with `cn()` over branching whole strings — e.g. a shared base plus `message.type === "error" ? "border-red-200 bg-red-50 text-red-800" : "border-blue-200 bg-blue-50 text-blue-800"`. Non-blocking.

packages/keycloak-theme/src/login/components/social-provider-button.tsx
```diff
@@ -24,18 +24,21 @@ export function SocialProviderButton({ provider }: Props) {
     <a
       id={`social-${provider.alias}`}
       href={provider.loginUrl}
-      className="border-input bg-background hover:bg-muted focus-visible:ring-ring inline-flex h-11 w-full items-center justify-center gap-2 rounded-md border px-8 text-sm font-semibold whitespace-nowrap transition-colors focus-visible:ring-2 focus-visible:ring-offset-2 focus-visible:outline-none"
+      className="border-accent bg-background text-accent hover:bg-muted focus-visible:ring-ring inline-flex h-[60px] w-[360px] max-w-full items-center justify-start gap-3 rounded-md border px-8 text-[20px] font-semibold whitespace-nowrap transition-colors focus-visible:ring-2 focus-visible:ring-offset-2 focus-visible:outline-none"
```
dam-code-guardian: 🟢 **Suggestion:** `w-[360px]` / `max-w-[360px]` is the shared content-column width, now repeated here and across `Login.tsx` (form, provider list, CTA). Per the styling reference, an arbitrary value used 3+ times is worth extracting to a theme token / utility class so the layout width lives in one place. Non-blocking.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `7fc7001`

## PR #2945: feat(keycloak-theme): make w3id login the primary CTA on SSO deployments
**Author:** kapetr | **Branch:** feat/w3id-primary-login → main | **Changes:** +155 −199 (11 files)

### Summary
Adds a per-deployment login presentation choice to the Keycloak theme: SSO-first (identity-provider CTAs only) vs the default password-first form, driven by `PLATFORM_ALLOW_PASSWORD` / `PLATFORM_REQUEST_ACCESS_URL` env vars wired through the Helm chart and keycloakify `environmentVariables` into `kcContext.properties`. Also redesigns the login layout to a single column and removes the aurora backdrop.

### Findings
- 🟢 **Suggestion:** The message-banner class is selected via a ternary returning two full Tailwind strings that differ only in color; the [HIGH] styling rule prefers `cn()` for conditional classes over branching whole strings (`packages/keycloak-theme/src/login/Template.tsx:66`).
- 🟢 **Suggestion:** `max-w-[360px]` (the form/button/CTA content width) is now an arbitrary value repeated across `Login.tsx` and `social-provider-button.tsx`; per the styling reference a value used 3+ times is worth a theme token (`packages/keycloak-theme/src/login/components/social-provider-button.tsx:27`).
- ✅ **Looks good:** `isSsoOnly` fails open — strict `=== "false"` plus `providers.length > 0` means any unexpected value, or SSO-only with no configured IdP, still renders the password form; a misconfigured deployment is never locked out (`Login.tsx:35`).
- ✅ **Looks good:** `requestAccessUrl` defaults to `""` (falsy) so the "Request access" link only renders when explicitly configured (`Login.tsx:165`).
- ✅ **Looks good:** `kc.gen.tsx` env additions match the `vite.config.ts` `environmentVariables` and the Helm env wiring; the auto-generated file is kept consistent.

✅ Aligned. `docs/architecture/security-and-credentials.md` was updated in the same PR to describe the two login variants, `Last verified:` bumped to 2026-07-23, no ADR reference leak, and volatile detail uses the sanctioned link-out pattern rather than transcribing config keys.

✅ No findings. `constants.ts` (single exported URL constant) and `vite.config.ts` (declarative plugin config) introduce no TSEng violations.

- 🟢 Conditional class via full-string ternary → prefer `cn()` (`Template.tsx:66`) — [HIGH] styling.
- 🟢 Repeated arbitrary width `max-w-[360px]` across two components → candidate theme token (`social-provider-button.tsx:27`) — [MODERATE] styling.

Components stay small, props are typed with `interface Props` and destructured, no `React.FC`, no static `style={{}}`, no server-state misplacement. The `dangerouslySetInnerHTML` uses are all wrapped in `kcSanitize` (correct for Keycloak i18n / provider names).

### Verdict
APPROVE — presentation-only change with correct fail-open guardrails and in-PR doc updates; only minor, non-blocking styling suggestions.

---

packages/keycloak-theme/src/login/Template.tsx
kapetr: Addressed in cc8cc9fa — shared base classes + `cn()` with only the color family branching.

jezekra1 (approved):

PetrBulanek (approved):
I just noticed that the "Sign In" button does not have a pointer cursor on hover, I'd "fix" that.

PR #2947, opened 2026-07-23
fix(channels): default Slack in-chat bind to ambient off

pilartomas:
## Why

**Security.** Ambient mode has the agent read *every* message in the channel. #2941 made the in-chat `/platform bind` default to ambient **on**, which turned that broad exposure into an implicit side effect of binding. This reverts that default: read-along should be a deliberate opt-in, not the default.

## What

- The in-chat `/platform bind` now creates the shared binding **ambient-off**, matching the UI form and the CLI flag (both already default off). Enabling read-along is an explicit `/platform ambient on` (or the UI/CLI toggle).
- Mechanically this removes the `ambient` argument that #2941 threaded through the bind flow (`executeSlackBind` → `connectShared` → `connectSlackImpl`), so the bind passes no ambient flag and `connectSlackImpl` leaves it off.

## What is NOT changed

The rest of #2941 stays intact — ambient enable/disable is still **never announced in the channel** (on connect, re-connect, or the in-chat `/ambient` command); it's audited and confirmed only on the operator's own surface. This PR touches the bind-time default only.

## Tests / docs

- `slack-bind.test.ts`: the bind test asserts `connectShared("agent-1", "C-1")` (no ambient flag) and a plain connect confirmation; removed the now-moot DM ambient-off case.
- `docs/architecture/channels.md`: ambient is off by default on every connect path, with the security rationale.
- `api-server` unit suite green (604 tests); tsc + eslint + prettier pass on the touched package.


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `541f67a`

### Summary
Reverts #2941's default of turning ambient (read-every-message) mode **on** for in-chat `/platform bind`. Now every connect path (in-chat bind, UI form, CLI) creates the shared binding ambient-off; read-along becomes an explicit opt-in via the ambient command. Mechanically drops the `ambient` argument threaded through `executeSlackBind → connectShared → connectSlackImpl`.

### Findings
- ✅ **Looks good:** Security posture is the right default — broad channel exposure is now a deliberate opt-in rather than an implicit side effect of binding.
- ✅ **Looks good:** With `ambient` undefined, `connectSlackImpl` computes `requestedAmbient = requestedMode === "shared" && ambient === true` → `false`, so bind correctly lands ambient-off (`agents-service.ts:656`).
- ✅ **Looks good:** On a fresh bind `existing` is null, so `wasAmbient (false) === requestedAmbient (false)` — no spurious `channel.ambient_toggled` audit event is emitted (`agents-service.ts:690`).
- ✅ **Looks good:** `isDm` was correctly relocated to where it's still needed (confirmation copy at `agents-service.ts:355`), not left dangling.
- ✅ **Looks good:** All `connectShared` call sites and its port type updated to the 2-arg shape; no stale 3-arg callers remain. Person-scoped path still passes `ambient` through unchanged.
- 🟢 **Suggestion:** None blocking. `connectSlackImpl`'s `ambient?: boolean` stays optional (pre-existing) — fine here since the person-scoped router path still supplies it and the schema enforces `ambient ⇒ shared`.

✅ No findings. `docs/architecture/channels.md` was updated in the same PR to describe ambient off on every connect path (with security rationale), `Last verified: 2026-07-23` matches the PR date, and no ADR references or volatile content were introduced.

✅ No findings. The change aligns with — and slightly improves — the layered/DI rules: the consumer (`executeSlackBind`) no longer computes and threads a capability toggle (`!isDm`), keeping the on/off question out of the consumer. DI wiring at the composition root is consistent; the test mock matches the new port shape.

### Verdict
APPROVE — a focused, well-tested security revert; code, docs, and tests are consistent and correct.

---

Issue #2821, opened 2026-07-20
Run agent loops as code over a spawnSandbox primitive


tomkis:
## Problem

The platform can launch AI R&D work but it can't drive the iterate-improve loop. The moment a loop has steps with different owners (a maker produces something, an automated gate checks it, an independent evaluator judges it, then a re-attempt on rejection) the single-agent-session model can't hold the hand-offs.

#2784 proposed fixing this with a **platform-driven conductor**: the platform bakes in the four steps (Make, Test, Eval, Curate), their order, and a set of config knobs (population size, top-k, stop policy, synthesizer flag), plus a new data model for generations and populations, a registry, a UI, and metrics.

That puts the loop's *shape* inside the platform. Every new loop shape (a tournament, retry-with-backoff, a cheap gate before an expensive one, dynamic fan-out width) then needs a platform change, and a large data-model/UI build has to land before the first loop can run. It's a quarter of work to prove one idea.

## Proposed solution

Push the loop's shape out of the platform and into code. The platform exposes **one primitive**: spawn an ephemeral sandbox — an agent given a chosen image, a chosen subset of the driver's connections, and a prompt; run it to completion; hand back a result the platform has validated against a caller-supplied schema. A loop becomes an ordinary program that calls this primitive in a normal `while`/`for`.

The platform's own principle is "what a step does lives in its prompt." Taken to its conclusion, the platform doesn't need to know there are four steps or what order they run in. There are only two node kinds (produce something / gate something), and everything else, depth, breadth, selection, stop conditions, is control flow the author writes. No platform change per loop shape.

To make these loops writable and reviewable without distributing a package: a tiny dependency-free SDK baked into the agent image, plus skills that teach an agent to turn a vague goal into a clean, reviewable loop script.

### Scope

- The **spawn primitive**: create an ephemeral sandbox and poll it for a validated result. Long-running turns mean create-then-poll, not one held-open call.
- **Result validation** is structural only (right fields, right types). The platform stays blind to content, it never checks whether a result is *true*. A node that never reports a valid result is caught by a liveness deadline.
- **Per-spawn controls**: time-to-live and memory/CPU size, so a heavy step can ask for more and a quick gate can fail fast.
- **Driver ergonomics**: the SDK (`spawn`, list images, list connections) and the skills that author loops.
- **Attenuation**: a sandbox can only be granted a subset of the driver's own connections.

### Dependencies

- Origin: #2784 (the platform-driven conductor PRD this replaces as the chosen direction).
[...]

Issue #2328, opened 2026-07-03
Epic - Experiments - Observable agent-driven optimization loops


jenna-winkler:
## Summary

Introduce an **Experiment** abstraction for running and observing agent-driven optimization loops.

An Experiment is a single execution of a driver agent's DBTL (design → build → test → learn) loop, authored as ordinary Python code using the Invocation primitive (#2821, #2816). The platform does **not** execute the optimization logic or interpret results — it provides the observability layer.

The loop defines its own structure through code, and reports that structure back to the platform:

* **Skeleton:** Declared stages, loops, and fork/join structure visible from the start.
* **Trace:** Live execution data including scored spans and linked Invocations.
* **Dashboard artifact:** A live experiment view, using a platform-provided dashboard by default or an agent-generated custom dashboard.

This replaces the previous platform-owned conductor model (#2784), which embedded loop structure in the platform. The new model keeps loops as code while restoring the missing observability layer. The previous Arms / Trials / Run Ledger model is retired behind the same feature flag (#2822).

## Problem

Today, experimental workflows are difficult to understand and monitor:

* Driver loops are just streams of Invocations with no structure, lineage, progress, or scoring visibility.
* Researchers building optimization loops (e.g., genetic algorithms) must rely on logs/stderr to understand execution.
* The previous experiment model introduced abstractions that no longer match the loops-as-code direction.

We still need the original Experiment goals:

* A standard abstraction for research workflows.
* Structured results and scores.
* Comparable and shareable experiment outcomes.
* Visibility into long-running agent workflows.

### Execution model

* Loops run inside the driver agent as normal Python scripts.
* The platform observes execution but never schedules loop steps or interprets scores.

### Authoring model

* Scripts declare an experiment skeleton upfront.
* Runtime execution emits spans against that skeleton.
* Undeclared stages are allowed and marked as drift rather than errors.

### SDK

Provide a dependency-free Python Experiment SDK built into platform-base.

Includes:

* Experiment/stage/span APIs.
* Existing driver capabilities (`spawn`, `listImages`, `listConnections`).
* Optional Pydantic support through duck typing.

The existing JS driver SDK remains for non-experiment workflows.

### Trace model

* Spans capture:

  * Scores (opaque numeric values).
  * Artifact references.
  * Metadata attributes.
* Scores are displayed but never normalized, ranked, or interpreted by the platform.
* `spawn()` automatically associates Invocations with the active span using context propagation.

### Candidate management

[...]

PR #2948, opened 2026-07-23
feat(experiments): live experiments over loops-as-code

### LINKED ISSUE #2822: Remove the Experiments feature
## Problem

Experiments, racing several agents against one goal and comparing their scored candidates, was the platform's first attempt at driving AI R&D loops. We've since decided to go a different way: loop-as-code over a `spawnSandbox` primitive (#2821), which supersedes it. The original Experiments PRD (#2784) is closed.

Experiments is still fully present in the product though: an opt-in feature with its own left-nav destination, a create wizard, a comparison screen, a settings toggle, plus a data model, a set of harness reporting tools, and its own architecture and vocabulary docs. All of it is now dead weight. It points users at a direction we're not taking, and it keeps concepts (experiments, arms, trials, runs, the run ledger) alive in the data model, the API, and the docs that no longer reflect where the platform is going.

## Proposed solution

Remove the Experiments feature end to end, so nothing about it remains in the product or the docs.

- Users no longer see the **Experiments** destination in the left nav, the create wizard, or the comparison screen, and the Experiments toggle disappears from Settings -> Features.
- The stored **experiments, arms, and runs** are removed from the data model.
- The harness-facing **reporting tools** (request a candidate upload, record a scored run, declare an arm finished), the **trial launch** path, the **inactivity-deadline** liveness sweep, and the **experiment-trial** session type all go away.
- The Experiments **architecture page** and the Experiments **vocabulary entries** are removed, and passing references to Experiments in the other docs are cleaned up.

### Scope

- This deletes any stored experiment data. Experiments is an opt-in experimental feature, so there's nothing production-critical to preserve, but the removal is destructive by design.
- Documentation references reach beyond the architecture page: the ubiquitous-language vocabulary, and mentions in the persistence, features, and configuration docs, plus the agent-image docs that describe the experiment reporting contract.

### Out of scope (things that look related but stay)

- **Object storage and the per-artifact size cap** stay. Experiments used them to hold candidates, but they're general-purpose infrastructure (also backing the artifact library), not Experiments-only.
- **The generic trigger/wake rail and the liveness-sweep pattern** stay. The new sandbox primitive already runs its own copies of both, so nothing Experiments shared with the rest of the platform needs to survive.

### Dependencies

- Part of the Experiments epic #2328 (this is the pivot away from the epic's Phase 4 "platform-driven loop" toward loop-as-code).
- Supersedes #2784 (closed). Follows the direction set in #2821 (implemented in #2816).

### LINKED ISSUE #2942: Live Experiments over loops-as-code
Part of #2328 (see the pivot comment there). Foundation: #2821 / #2816. Removal half: #2822. Interactive-artifacts bridge: #2884. Vocabulary: PR #2939.

## Problem Statement

I can write an AI R&D loop today — a driver agent scripting produce/eval/select over spawned invocations — but the moment it starts running, it goes dark. There is no way to see the loop's shape, which stage is executing, which sub-agents are alive, how scores are trending across iterations, or whether the thing silently died an hour ago. My only windows are stderr and a flat list of invocations with no structure or lineage. The retired arms-racing Experiments feature had a comparison screen, but it forced my work into a shape (competing arms reporting scored runs) that most real loops don't have, and it never showed live progress either.

When I tweak a loop and rerun it, I also can't tell which version of the script produced which results — the script lives as a mutable file in the agent's workspace, and results have no tie to the code that generated them.

## Solution

An **Experiment** is one execution of a loop script my agent writes in Python. The workflow:

1. **Prepare.** I chat with my agent; it authors the experiment script using the platform's Python SDK. The script declares its shape upfront — stages, loops, fork/join — then runs ordinary Python control flow inside it.
2. **Plan.** Running the script in plan mode registers the experiment: it appears in the UI as a draft showing the declared graph, before any tokens are spent. The exact script version is captured automatically.
3. **Execute.** I click Execute (or tell my agent to run it). The loop runs inside my agent's pod, unattended. I can close the browser; the platform keeps the agent awake until the experiment finishes and watches for silent death.
4. **Watch.** The experiment page shows the graph live: per-stage progress, running sub-agents, scores charting over iterations, drill-down into any span's invocation — its prompt, result, and produced candidate. A running experiment is visible from the main page via its agent's card. The visualization is itself an artifact: a stock dashboard by default, or a fully bespoke one my agent generates for this specific experiment — because a GA run, a prompt-evolution loop, and a benchmark sweep deserve different views.
5. **Keep.** Candidates, the final report, the dashboard, and every executed script version live in the Artifact Library — versioned, downloadable, shareable, outliving the pod. Re-running creates a sibling experiment tied to the same script lineage.

The platform's founding bet is preserved: it never runs the optimization loop and never interprets a score. The loop's shape lives in code; the code reports its shape.

## User Stories

- As a researcher, I want to watch my experiment's graph fill in live — stages, iterations, scores — so that I can tell at a glance whether it's converging or wedged, without reading logs.
- As a researcher, I want my experiment to keep running with the browser closed and the platform to keep my agent awake, so that long loops survive my laptop lid.
- As a researcher, I want every run permanently tied to the exact script version that produced it, so that I can trust and reproduce comparisons between runs.
- As a researcher, I want to review the planned experiment graph before executing, so that I catch a mis-designed loop before it burns tokens.
- As an agent author, I want a small Python SDK where declaring the skeleton and tagging spans costs a handful of lines around code I'd write anyway, so that instrumentation never fights the loop logic.
- As an agent, I want to generate a bespoke HTML dashboard for an unusual experiment, so that the live view fits the experiment instead of a generic template.
- As a platform operator, I want silent or crashed loops to reach a terminal state on their own, so that stuck experiments don't pin agents awake forever.

## Implementation Decisions

*(settled in the 2026-07-23 design session; vocabulary in the glossary rebase, PR #2939)*

1. **Harness-side, observed.** The loop executes in the driver agent's pod; the platform captures structure and progress, never schedules nodes. Reaffirms #2821 over #2784.
2. **Skeleton + trace, lenient.** Scripts declare stages/loops/forks upfront; execution emits stage-tagged spans. Undeclared stages grow the graph marked as drift — never an error.
3. **Python SDK, stdlib-only, full driver surface.** One dependency-free module baked into platform-base, subsuming spawn/discovery from the JS driver-sdk plus the skeleton/span API. `spawn()` inside a span auto-attaches via contextvar (explicit override for fan-out). Pydantic accepted via duck-typing; JS driver-sdk remains for generic loops.
4. **Span payload:** optional opaque numeric score (captured, plotted, never normalized), artifact references, opaque attrs bag. Driver-only trace writers in MVP.
5. **Plan registration → Execute.** Plan mode creates the draft (skeleton + script capture); Execute delivers a composed prompt over the runtime channel and the harness runs the script as a background process. The SDK reports directly to the platform over the per-agent HTTP surface (mesh-attributed, no tokens).
6. **Script identity: versioned artifact** in an "experiment scripts" Artifact Library folder — never Postgres; each run records the exact version executed. Candidates are Artifact Library references too; the bespoke candidate-upload path retires.
7. **Lifecycle:** draft → running → completed | failed | stopped. A running experiment pins its agent against the controller's idle checker; an inactivity sweep (port of the old Inactivity Deadline) fails silent runs and releases the pin. Stop closes the trace — the loop dies on its next rejected call.
8. **Live view:** native detail-page chrome around a dashboard-artifact canvas in a sealed iframe. Stock dashboard auto-attached at plan registration; agent-generated dashboards may replace it. Data flows via the postMessage bridge: the host page polls the platform's stable **Trace Feed** while running and pushes JSON into the iframe — no credentials in user HTML, no sandbox egress. Lands as capability #1 of Interactive Artifacts (#2884).
9. **Re-run = sibling experiment** sharing script-artifact lineage.
10. **Replacement, not migration:** the rebased subsystem ships behind the existing experiments feature flag; the arms-racing model is removed by #2822.
11. **Testing:** pytest suite for the SDK against a stub platform endpoint; pure-function unit tests for the trace-graph resolver and lifecycle machine; one Playwright E2E driving mock-agent through plan → execute → spans → completion.

**Deliberate fast-follows (out of scope):** multi-writer traces (target-emitted spans), shared/public live pages, cross-run comparison views, crash self-healing, evaluator-as-shared-resource (the dropped Phase 3 thread — needs its own design).

jezekra1:
An Experiment is now one execution of a Python loop script your agent authors — and you watch it run, right where you work with the agent.

📓 https://claude.ai/code/artifact/01ff3997-3cd2-407b-a3c4-ab6418e50480

📹 https://github.com/user-attachments/assets/ce6eb888-27a8-41d5-88fd-2556364a6027

**The experience.** Chat with an agent to design a loop (produce → eval → select, a GA, a benchmark sweep). Running the script in plan mode registers a **draft** Experiment, and its panel **docks itself in the agent's chat**: the declared graph, an Execute button — the review moment before a single token is spent. Execute launches the script as a background process in the agent's pod, opens the launch session in the chat so you can follow it (and ask the agent things mid-run), and the docked panel goes live: per-stage progress, score-over-iteration charts, drift flags, spawned sub-agents attached to their stages, a running-invocations count. Close the browser — the run keeps going, and the platform keeps the driving agent awake until it finishes. Silent or crashed loops are reaped to *failed* automatically. The **Experiments destination** is an index of experiment agents (newest experiment, status, running invocations) that routes you into the chat.

**What you keep.** Every executed script version and every candidate live in the Artifact Library — versioned, downloadable, shareable. On completion the platform publishes a **snapshot version of the dashboard with the final results baked in**, so the finished experiment is a self-contained shareable page. Re-running after a tweak creates a sibling experiment tied to the same script lineage.

**Bespoke dashboards.** The live view is itself an artifact: a stock renderer by default, replaceable by fully agent-generated HTML per experiment. It renders sealed — data is pushed in by the page over postMessage, generated HTML never gets network or credentials; the terminal snapshot replays the final data through the same contract.

**The platform never runs the loop.** The loop's shape stays in ordinary code (the loops-as-code direction, #2821); a stdlib-only python SDK, baked into every agent image, both drives (spawn) and reports (skeleton + spans). Scores are captured and charted, never interpreted.

The previous arms-racing Experiments feature is removed as part of this change; the feature flag carries over.

E2E: the Playwright spec drives the whole rail with the real SDK inside the mock agent (passed locally; full chain runs in the pipeline).

Closes #2942
Closes #2822
Part of #2328


dam-code-guardian (commented):
🛡️ **DAM** — ⚠️ Code Review @ `7bcd836`

## PR #2948: feat(experiments): live experiments over loops-as-code
**Author:** jezekra1 | **Branch:** feat/2942-live-experiments → main | **Changes:** +11669 −3054 (126 files)

### Summary
Replaces the v1 "arms-racing" Experiments feature with **loops-as-code**: an Experiment is now one execution of an agent-authored, stdlib-only Python loop script, observed live via a declared skeleton + a trace of stage-tagged scored spans. Adds the experiment SDK, a plan→Execute lifecycle with atomic conditional transitions, an inactivity sweep + hibernation pin, versioned script artifacts, and a sealed-iframe dashboard (stock or agent-generated). Clean, layered, well-tested work; the destructive v1 drop is intentional and documented in the migration.

### Findings
- 🟡 **Warning:** Committed Python bytecode. Three `__pycache__/*.pyc` files are tracked (`packages/experiment-sdk/src/__pycache__/experiment_sdk.cpython-314.pyc` plus two under `tests/__pycache__/`). Build artifacts shouldn't be in VCS — they churn, leak the local Python version (3.14), and can go stale against source. Remove them and add a `packages/experiment-sdk/.gitignore` (`__pycache__/`, `*.pyc`).
- 🟡 **Warning:** A long `spawn()` can get a live experiment reaped. `lastActivityAt` is bumped only by accepted trace events (`appendEvents` → `bumpActivity`); during a blocking `spawn()` the SDK only polls `/invocations/{id}` (which doesn't touch experiment activity) and emits nothing until the span ends. A stage delegating to a sub-agent that runs longer than the inactivity window (default **15 min**, `config.ts`) produces no events, so the sweep flips the still-working run to `failed` and releases its pin. LLM sub-agent stages routinely exceed 15 min. Consider a heartbeat during spawn polling, bumping experiment activity on attached-child status changes server-side, or a more generous default.
- 🟢 **Suggestion:** Module-boundary deep import — `compose.ts:5` imports `createInvocationsRepository` from `../invocations/infrastructure/...`, reaching past that module's `index.ts`. Contained in the composition root, but re-exporting from the invocations `index.ts` keeps the boundary honest.
- 🟢 **Suggestion:** `appendDrift` (`experiments-repository.ts:266`) is a read-modify-write full-overwrite of `drift` rather than an atomic append. Safe today (one driver emits an experiment's events serially); a comment noting that assumption (or a jsonb append) would harden it.
- ✅ **Looks good:** Lifecycle as a pure module with atomic `WHERE status = from` transitions; the sweep re-checks through the pure `sweepDecision` rule and no-ops lost races; span upsert handles out-of-order start/end; `requireDriverExperiment` collapses foreign/missing to one "unknown"; sealed `DeferredFrame` correctly omits `allow-same-origin`; strong unit coverage on lifecycle/trace-graph/SDK plus an E2E spec.

**Verdict: minor drift.** Check 2 (`Last verified:` staleness): six architecture pages received substantive behavioral edits removing retired v1 vocabulary (Arms/Trials/Candidates) and re-pointing responsibilities, but their `Last verified:` headers weren't bumped to the PR date (2026-07-23):
- `agent-lifecycle.md` (2026-07-20) — "Four paths trigger a wake" → "Three"; removed Experiment-driven wake
- `artifact-library.md` (2026-07-21) — dropped "experiments subsystem introduced" / candidate-route lineage
- `connections.md` (2026-07-22) — removed the `experiment-trigger` event kind
- `features.md` (2026-07-21) — "experiments and advanced connection types" → "advanced connection types"
- `persistence.md` (2026-07-22) — object-store first consumer re-pointed to artifact-library
- `security-and-credentials.md` (2026-07-22) — "Candidate object store" → "artifact object store"

`experiments.md` was correctly bumped and matches the code (15-min default, no dangling `#candidate-storage` anchor, no ADR leaks).

1 finding: **Module public API boundary** — `compose.ts:5` imports another module's infrastructure directly instead of via its `index.ts` (see 🟢 above). Otherwise the new module cleanly follows the domain/services/infrastructure layering, contract-package validation, and DI-via-factory patterns.

✅ No findings. Query key factory via tRPC, `meta.invalidates`/`errorToast`, selector hooks, live-only polling with `skipToken`, extracted subcomponents, no raw `fetch`, no `any`, no static `style={{}}`; the sandbox `postMessage` target-origin is justified (opaque-origin sandbox, viewer's own data).

### Verdict
COMMENT — solid, well-tested subsystem; two warnings (committed bytecode; the spawn-vs-inactivity-sweep gap that can reap a live run) and minor doc-date drift are worth addressing before merge, none blocking.

---

packages/api-server/src/modules/experiments/compose.ts
```diff
@@ -1,66 +1,127 @@
 import type { Db } from "db";
 import type { ExperimentsService } from "api-server-api";
+import type { ArtifactLibraryServiceImpl } from "../artifact-library/index.js";
+import type { RuntimeMutator } from "../runtime-delivery/index.js";
+import { createInvocationsRepository } from "../invocations/infrastructure/invocations-repository.js";
```
dam-code-guardian: 🟢 **Suggestion:** This reaches into the invocations module's infrastructure directly, past its `index.ts` (which doesn't export `createInvocationsRepository`). Per the module-boundary rule, `index.ts` is the only entry point other modules should import from. Contained here in the composition root, but re-exporting the factory from `invocations/index.ts` keeps the boundary honest.

packages/experiment-sdk/src/experiment_sdk.py
```diff
@@ -0,0 +1,512 @@
...
+
+    invocation_id = _request("POST", "/invocations", body)["id"]
+    _log(f"spawned {name} ({invocation_id})")
+
+    deadline = time.monotonic() + (
+        timeout_seconds
+        if timeout_seconds is not None
+        else (ttl_ms / 1000 + 60 if ttl_ms else _DEFAULT_SPAWN_TIMEOUT_S)
+    )
+    while True:
```
dam-code-guardian, line 255: 🟡 **Warning:** While this loop blocks on a spawned invocation, the SDK emits no trace events, so the experiment's `lastActivityAt` is never bumped (only `appendEvents` bumps it; polling `/invocations/{id}` doesn't). If a sub-agent stage runs longer than the inactivity window (default 15 min), the server-side inactivity sweep reaps the still-working run to `failed` and releases its pin. LLM sub-agent stages commonly exceed 15 min. Consider a periodic heartbeat here, bumping experiment activity when an attached child reports server-side, or a more generous default window.

tomkis (approved):

PR #2950, opened 2026-07-23
feat(channels): swap the Slack auto-👀 ack for an agent-chosen contextual reaction

pilartomas:
## What

The api-server used to ack every direct Slack turn by auto-adding a 👀 reaction to the triggering message. With the live working status (`is thinking…` / `is working…`) now presented on mention and fork turns, that generic reaction is redundant there — so it's removed from both the owner-relay and foreign-fork paths.

Ambient (read-along) turns are the opposite case: they deliberately present **nothing** on the platform's behalf, so there was no lightweight in-channel signal at all. This gives them one the *agent* owns — the read-along frame now nudges the agent to open a turn it engages with by reacting with a **fitting, context-appropriate emoji** via the existing `react` tool. It notifies no one and can stand alone as the whole response when a full reply isn't warranted.

## Why ambient-only

Direct turns already show the live status line, so pushing an early react there would mostly duplicate it and add reaction-noise on nearly every turn. Ambient is the only mode with a genuine signal gap, so the nudge lives in the ambient frame only. The shared per-turn contract keeps documenting the `react` tool (example refreshed to nudge a *fitting* emoji rather than leading with 👀).

## Changes

- **`slack.ts`** — remove the `addReaction("eyes")` calls from `relayOwnerTurn` and `beginForeignTurn`; add early-contextual-react guidance to `ambientGuidance`; refresh the shared react example.
- **`channels.md`** — the ambient section now describes the agent-owned reaction; drop the stale "turn-start 👀 ack" note next to the `reactions:write` scope.
- **`slack-app-manifest.yaml`** — scope comment now attributes `reactions:write` to the `react` tool (incl. the ambient contextual ack), not a turn-start ack.
- **Tests** — fix the direct-turn / foreign-fork unit assertions (no auto-reaction; status is the platform's only presented signal), fix the ambient-mention unit test, add coverage asserting the ambient frame carries the react nudge, and drop the two e2e eyes-poll steps (the following status/reply polls already prove acknowledgement).

## Testing

- `mise run api-server:check` — tsc + lint + format clean
- `mise run api-server:test` — **604/604 pass**
- `mise run e2e-playwright:check` — tsc + lint + format clean

E2E specs were updated but not run (requires a live cluster). The mock e2e agent only ever calls `reply`, so the ambient spec's "no reaction" assertion still holds.


dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `404538b`

## PR #2950: feat(channels): swap the Slack auto-👀 ack for an agent-chosen contextual reaction
**Author:** pilartomas | **Branch:** feat/slack-contextual-ambient-react → main | **Changes:** +40 −66 (8 files)

### Summary
Removes the platform's automatic `addReaction("eyes")` ack on owner-relay and foreign-fork Slack turns (the live working status already signals acknowledgement there), and instead nudges the agent to open an *ambient* turn it engages with by picking a context-appropriate emoji via the existing `react` tool. Docs, the app manifest scope comment, and unit/e2e tests are updated to match.

### Findings
- ✅ **Looks good:** Both `addReaction("eyes")` removals (`relayOwnerTurn`, `beginForeignTurn`) are clean — no orphaned locals; `gw` is still used by the presenter/reply paths that follow.
- ✅ **Looks good:** Behavior change is coherent end-to-end — direct turns keep the working status as their signal, and the ambient frame gains the agent-owned contextual react, matching the stated "ambient is the only mode with a signal gap" rationale.
- ✅ **Looks good:** Test assertions track the change accurately: the owner/fork unit tests now assert `status` present + `reaction` absent, and the surviving reaction test (`react adds the emoji…`) exercises the agent-driven `react` tool path, not the removed auto-ack.
- ✅ **Looks good:** The new ambient-frame assertion (`"open with a fitting emoji reaction"`) correctly matches the concatenated `ambientGuidance` prompt string.
- ✅ **Looks good:** `reactions:write` scope is retained (still needed by the `react` tool); the manifest comment is re-attributed rather than the scope dropped.
- 🟢 **Suggestion:** The e2e specs for the ambient/mention paths were updated but not run (author notes this requires a live cluster). The mock e2e agent only calls `reply`, so the "no reaction" assertions still hold — no action needed, just worth confirming on the next live-cluster run.

✅ No findings. `docs/architecture/channels.md` is updated in the same PR to describe the removed platform auto-ack and the new agent-owned contextual reaction; `Last verified: 2026-07-23` matches the PR date. No ADR references, volatile content, or cross-reference rot introduced.

✅ No findings. The diff is a behavior simplification (two platform-side `addReaction` calls removed) plus prompt-string and test-assertion updates — no new imports, no cross-layer coupling, and the infrastructure adapter stays within its layer.

### Verdict
APPROVE — A clean, well-scoped simplification: the removed auto-ack is fully dead-code-free, the agent-owned ambient react is a coherent replacement, and unit + e2e tests and the architecture doc are all updated to match.

---

PR #2953, opened 2026-07-24
feat(skills): create a skill by uploading Markdown files

### LINKED ISSUE #2823: Create a skill by uploading Markdown files
## Problem

When adding skills to a sandbox, the only route is pointing at a GitHub repository. A user who has a `SKILL.md` on their machine — drafted locally, copied from elsewhere, or not yet in a repo — has no way to bring it into a sandbox. The "Upload .md files" tab exists in the add-source dialog but is shown disabled, and the drag-and-drop empty state was removed because it did nothing.

## Proposed solution

Let a user upload one or more Markdown files directly to create a standalone skill in the sandbox, without a backing GitHub source. Re-enable the disabled upload tab and restore a working drop zone.

### Dependencies

Deferred from #944 (was out of scope — needs a backend write path that doesn't exist yet).

PetrBulanek:
Closes #2823

Lets you create standalone skills in a sandbox by uploading Markdown files — no backing GitHub repository required.

## What it does

- Drop one or more `.md` files anywhere on the Skills surface, or pick them from the **Upload .md files** tab in the add-source dialog.
- Review each staged skill before committing: rename it inline, or remove ones you don't want.
- Click **Add N skills** — they appear under **Created in this sandbox** and are publishable just like any other standalone skill.

Each file becomes one skill. The initial name comes from the file's frontmatter `name:` (falling back to a prettified filename), and you can edit it before adding. Uploading a name that already exists in the sandbox is rejected up front — the conflict is shown on the offending row and nothing is overwritten. Non-Markdown or oversized files are declined with a clear inline notice.

The empty Skills state again invites dropping a `.md` file to get started, and switching between the GitHub and Upload tabs keeps whatever you've entered in each.


packages/agent-runtime/src/modules/skills/domain/frontmatter.ts
```diff
@@ -51,3 +51,48 @@ export function parseFrontmatter(content: string): {
...
+ */
+export function ensureFrontmatterName(content: string, name: string): string {
+  const line = `name: ${yamlScalar(name)}`;
+  const m = /^(---\r?\n)([\s\S]*?)(\r?\n---)/.exec(content);
+  if (!m) return `---\n${line}\n---\n\n${content}`;
+
+  const [full, open, body, close] = m;
+  const nl = open.includes("\r\n") ? "\r\n" : "\n";
+  const bodyLines = body.length > 0 ? body.split(/\r?\n/) : [];
+  const nameIdx = bodyLines.findIndex((l) => /^name:\s*/.test(l));
```
dam-code-guardian, line 90: 🟢 **Suggestion:** this rewrites only the single `name:` line found by `/^name:\s*/`. If an uploaded file already declares `name:` as a YAML block/multiline scalar (`name: |` or `name: >` followed by indented lines), those continuation lines are left orphaned — a real YAML parser then folds them into the value, so the confirmed display name isn't quite what lands on disk. It's an extreme edge (the UI's `deriveName` only ever produces single-line names), but a short comment noting the single-line-`name:` assumption — or skipping the block-scalar case — would make the write↔read round-trip claim airtight.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `3db221c`

## PR #2953: feat(skills): create a skill by uploading Markdown files
**Author:** PetrBulanek | **Branch:** feat/2823-skill-upload → main | **Changes:** +1127 −155 (25 files)

### Summary
Adds an end-to-end "upload Markdown to create standalone skills" path across three layers: an agent-runtime `writeLocal` tRPC mutation that materializes uploaded `.md` as `<slug>/SKILL.md`, an api-server `createLocal` orchestration that wakes the agent and security-logs the write, and a UI upload tab (drop zone, staged rows, inline rename, collision marking). Cleanly layered, docs updated in the same PR.

### Findings
- 🟢 **Suggestion:** `ensureFrontmatterName` rewrites only the single `name:` line. If an uploaded file already declares `name:` as a YAML block/multiline scalar (`name: |` + indented lines), those continuation lines are left orphaned and a real YAML parser folds them into the value — so the confirmed name isn't what lands. Extreme edge (the UI only derives single-line names), but a one-line "single-line `name:` only" note or guard would make the write/read round-trip claim airtight. (`packages/agent-runtime/src/modules/skills/domain/frontmatter.ts:90`)
- 🟢 **Suggestion:** `write-local.ts` (services layer) imports `MAX_FILE_BYTES`/`MAX_SKILL_BYTES` as **values** from the infrastructure layer; the sibling services (`install`/`scan`/`publish`) import only the `LocalSkillRepository` port *type*. These byte-cap invariants read more naturally as domain constants. Minor — infra legitimately owns them for the read path too. (`packages/agent-runtime/src/modules/skills/services/write-local.ts:17`)
- 🟢 **Suggestion:** the new pure domain functions (`ensureFrontmatterName`, `makeSkillSlug`, `runWriteLocal`) ship without unit tests. The module had none before, so not a regression — but the write↔read frontmatter round-trip and the all-or-nothing collision logic are exactly the invariants a small test file would lock down.
- ✅ **Looks good:** batch is fully validated (slugs, size caps, within-batch + on-disk + existing-name collisions) before any write; the `skill(s) already exist: <names>` CONFLICT contract is preserved verbatim across all three layers so the UI can re-mark rows; the PV write is security-logged (parity with `install`); the slug is funneled through `makeSkillName` so it can't smuggle a path-traversal sequence; the char-vs-byte cap split is documented as an intentional cheap early gate.

✅ No findings. `docs/architecture/skills.md` was updated in the same PR (agent-runtime surface gains `readLocal`/`writeLocal`, a fourth "Write Local" responsibility, the "Create Local orchestration" api-server bullet, and the standalone-skills note), `docs/ubiquitous-language.md` gains the "Write Local" term, and `Last verified:` is bumped to 2026-07-23. Code and architecture docs are aligned.

Layered architecture is respected: domain (`frontmatter.ts`, `skill-name.ts`) stays pure, services depend on the repository port, Zod validation lives in the contract packages, and errors flow as `Result`/`SkillsDomainError` mapped to tRPC codes in the routers. One 🟢 note (byte-cap constants imported from infra into the service layer — see Findings).

✅ No findings. Clean split: the modal shell owns both tabs' state via `useGithubSourceForm`/`useUploadStaging` so tab switches don't drop in-progress work; `.map` renders a `StagedSkillRow` subcomponent; conditional classes via `cn()`, no static `style={{}}`, no `any` at boundaries; server state stays in the existing `standalone` list with an optimistic merge ahead of the poll.

### Verdict
APPROVE — solid, well-layered feature with docs and security logging in place; all findings are non-blocking 🟢 suggestions.

---

packages/ui/src/modules/sandboxes/components/skills/upload-skills-tab.tsx
```diff
@@ -0,0 +1,231 @@
...
+  onClose: () => void;
+}) {
+  const { staged, notice, conflicts, topError, submitting, addFiles } = staging;
+  const [renamingId, setRenamingId] = useState<string | null>(null);
+  const [dragActive, setDragActive] = useState(false);
+  const fileInputRef = useRef<HTMLInputElement>(null);
+
+  const trimmedNames = staged.map((s) => s.name.trim());
+  const hasEmpty = trimmedNames.some((n) => n.length === 0);
+  const hasDuplicate = new Set(trimmedNames).size !== trimmedNames.length;
```
kapetr: Staging "My Skill" + "my_skill" passes the duplicate check, gets rejected pod-side as a slug clash, and both rows falsely claim "already exists in this sandbox". Fix is cheap — the tab already has toSlug, so a slug-collision pre-submit check with accurate copy.

packages/ui/src/modules/sandboxes/components/skills/add-skill-source-modal.tsx
```diff
@@ -108,73 +68,52 @@ export function AddSkillSourceModal({
...
+      {tab === "github" ? (
+        <GithubSourceTab github={githubForm} onClose={onClose} />
+      ) : (
+        <UploadSkillsTab staging={uploadStaging} onClose={onClose} />
+      )}
     </Modal>
   );
 }
+
+function TabButton({
```
kapetr, line 93: candidate for global primitive?

kapetr (approved):

packages/ui/src/modules/sandboxes/components/skills/upload-skills-tab.tsx
kapetr: This is agent's finding and I think it can be disregarded - skill name shouldn't contain conflicting names like these. Although we don't validate that.

packages/ui/src/modules/sandboxes/components/skills/add-skill-source-modal.tsx
PetrBulanek: I agree that we can refactor it, but I think I'd leave that for later.

PR #2955, opened 2026-07-24
fix(channels): connect Slack socket mode on startup, not lazily

pilartomas:
## Problem

The Slack workspace socket-mode connection was established **lazily** — it opened only on the first channel bind or the first outbound post, via `ensureGateway()`. On a fresh install (or any api-server start with no Slack bindings), the socket never opened at boot. Because socket mode is how Slack pushes inbound events, this meant slash commands (`/<brand> login`), `@mentions`, and DMs sent to the bot in **unbound** chats were silently dropped until something else happened to trigger a connect.

Telegram already connects unconditionally at boot for exactly this reason (`/login` must work in chats with no binding yet). Slack should match.

## Fix

- Add `SlackWorker.connect()` — opens the socket via the existing `ensureGateway()` (idempotent, single-flight with the other gateway starters).
- Call `slackWorker.connect()` unconditionally in `ChannelManager.bootstrap()`, alongside `telegramWorker.start()`, **before** walking the per-Agent bindings. The subsequent binding walk now only restores per-Agent registration (the socket is already up).

## Why it's safe

- `ensureGateway()` is single-flight and caches the gateway, so the boot-time connect plus any concurrent `SlackConnected` event / outbound post share one connection — no double socket.
- A failed connect latches `gatewayFailed` and returns without throwing; `bootstrap` is already fire-and-forget at the call site.
- No Slack tokens configured → `slackWorker` is `undefined` → `connect()` is skipped.
- e2e fake gateway: starting it at boot only sets its handlers earlier (idempotent, writes nothing to outbound). Every e2e flow already binds-then-fires; none depends on a pre-start throw. Verified against all four Slack smoke specs.

## Tests

New `channel-manager.test.ts` locks in the regression: `bootstrap(new Map())` (no bindings) calls `slackWorker.connect()` once and does **not** call `slackWorker.start()`; a bound channel connects first, then registers per-Agent.

- `mise run api-server:check:tsc` — clean
- `mise run api-server:check:lint` / `:format` — clean
- `mise run api-server:test` — 126 existing + 3 new pass

## Docs

Updated `docs/architecture/channels.md` (Slack transport + bootstrap contrast) and bumped `Last verified`.

### DIFF
--- a/packages/api-server/src/modules/channels/services/channel-manager.ts
+++ b/packages/api-server/src/modules/channels/services/channel-manager.ts
@@ -146,10 +146,15 @@ export function createChannelManager(deps: {
     },
 
     async bootstrap(channelsByInstance: Map<string, ChannelConfig[]>) {
-      // The platform bot polls unconditionally — /login must work in chats
-      // that have no binding yet.
+      // Both platform bots connect unconditionally at startup — inbound
+      // commands (/login), mentions and DMs must reach the bot in chats that
+      // have no binding yet. Slack opens its socket-mode connection here rather
+      // than lazily on the first bind/post, so it never misses those events.
       if (telegramWorker) await telegramWorker.start();
+      if (slackWorker) await slackWorker.connect();
 
+      // The socket is already up; walking the bindings only restores the
+      // per-Agent registration the SlackConnected event installs at runtime.
       for (const [agentId, channels] of channelsByInstance) {
         for (const channel of channels) {
           if (channel.type === ChannelType.Slack && slackWorker) {
--- a/docs/architecture/channels.md
+++ b/docs/architecture/channels.md
@@ -1,6 +1,6 @@
 # Channels
 
-Last verified: 2026-07-23
+Last verified: 2026-07-24
 
 ## Overview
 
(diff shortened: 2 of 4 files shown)


packages/api-server/src/modules/channels/services/channel-manager.ts
```diff
@@ -146,10 +146,15 @@ export function createChannelManager(deps: {
     },
 
     async bootstrap(channelsByInstance: Map<string, ChannelConfig[]>) {
-      // The platform bot polls unconditionally — /login must work in chats
-      // that have no binding yet.
+      // Both platform bots connect unconditionally at startup — inbound
+      // commands (/login), mentions and DMs must reach the bot in chats that
+      // have no binding yet. Slack opens its socket-mode connection here rather
+      // than lazily on the first bind/post, so it never misses those events.
       if (telegramWorker) await telegramWorker.start();
```
dam-code-guardian, line 153: 🟢 **Suggestion:** `telegramWorker.start()` and `slackWorker.connect()` run sequentially with `await`, so a slow or hanging Slack handshake delays the Telegram registration walk. Not a defect — both were effectively serial before — but if boot latency ever matters these two independent steps could run concurrently:

```suggestion
      await Promise.all([
        telegramWorker ? telegramWorker.start() : undefined,
        slackWorker ? slackWorker.connect() : undefined,
      ]);
```

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `a05e5b4`

## PR #2955: fix(channels): connect Slack socket mode on startup, not lazily
**Author:** pilartomas | **Branch:** fix/slack-socket-connect-on-startup → main | **Changes:** +95 −5 (4 files)

### Summary
Slack socket-mode previously opened lazily on the first bind/outbound post, so a fresh api-server with no Slack bindings never opened the socket and silently dropped inbound slash commands, mentions and DMs. This adds `SlackWorker.connect()` and calls it unconditionally in `ChannelManager.bootstrap()`, mirroring how Telegram already boots.

### Findings
- ✅ **Looks good:** `connect()` delegates to the existing `ensureGateway()`, which is single-flight (`gatewayStarting ??=`) and caches `gateway` — the boot connect and any concurrent `SlackConnected`/outbound post share one socket, so no double connection (`slack.ts:1739`, `slack.ts:1706`).
- ✅ **Looks good:** A failed connect latches `gatewayFailed` and returns null without throwing; `bootstrap` is fire-and-forget, so a Slack outage can't break boot (`slack.ts:1723`).
- ✅ **Looks good:** `if (slackWorker)` / `if (telegramWorker)` guards mean no-Slack-token installs skip `connect()` cleanly (`channel-manager.ts:154`).
- ✅ **Looks good:** `connect()` added to the `SlackWorker` port interface, not just the implementation — keeps the port/adapter contract honest and lets the test fake type-check (`slack.ts:283`).
- ✅ **Looks good:** Regression is locked by three focused unit tests: no-bindings connects once without `start()`, bound channel connects-then-registers, Telegram-only boots clean (`channel-manager.test.ts:26`).
- 🟢 **Suggestion:** `connect()` runs after `telegramWorker.start()` sequentially with `await`; a slow/hanging Slack handshake would delay the Telegram registration walk. Not a defect (both were effectively serial before), but `await Promise.all([...])` would decouple them if boot latency ever matters (`channel-manager.ts:153`).

Verdict: **aligned**. `docs/architecture/channels.md` was updated in the same PR — both the Slack Transport bullet and the bootstrap sentence now describe the unconditional boot connect — and `Last verified` is bumped to 2026-07-24. No ADR leaks, no volatile content added, no missing page. ✅ No findings.

`connect()` is declared on the `SlackWorker` port interface and implemented in the adapter factory — clean Ports & Adapters alignment. The optional `slackWorker?`/`telegramWorker?` deps are pre-existing and out of scope for this diff. ✅ No new findings.

### Verdict
APPROVE — a well-scoped, well-tested fix that closes a real inbound-event gap and keeps the docs and port contract in sync; the single suggestion is optional.

---

Issue #2954, opened 2026-07-24
Agent's first reply in a channel can include content from unrelated conversations


pilartomas:
## What happens

When an agent is newly brought into a Slack channel and answers for the first time, its reply can draw on messages from *other, unrelated conversations* in that channel — not just the one it's answering. Because the agent has no history for the new conversation yet, it's handed a snapshot of the channel's recent activity to work from, and that snapshot can include messages belonging to different threads or topics. The agent may then quote or act on that unrelated content in its answer.

## What should happen

An agent's reply should be grounded only in the conversation it is answering. Content from other conversations in the same channel shouldn't surface in a reply unless someone deliberately brought it into that conversation.

## Steps to reproduce

1. In a Slack channel, have some recent, unrelated activity (messages about different topics).
2. Bring an agent into that channel and address it for the first time with a fresh question.
3. Observe that the agent's first reply can reference or incorporate content from the earlier, unrelated messages.

---

Context: reported alongside #2952. That issue now tracks a separate problem — a reply being delivered into the wrong thread. This issue tracks unrelated content appearing inside the reply itself.

PR #2956, opened 2026-07-24
fix(channels): stop Slack replies crossing threads under concurrency

### LINKED ISSUE #2952: Cross-thread context leak: agent's reply contained content from another thread
## Summary
An agent's reply in a Slack thread contained content that originated from a **different thread**, indicating context/data is bleeding across thread boundaries.

## Observed behavior
On the agent's very first reply after being bound to a channel, the response included content leaked from another thread rather than being scoped to the current thread only.

## Expected behavior
Each thread's context should be isolated. A reply in thread A must never surface content from thread B.

## Impact
- Confidentiality: content from one thread/conversation can be exposed to participants of another.
- Correctness: replies may include irrelevant or misleading content pulled from unrelated threads.

## Environment
- Slack integration (DAM app), staging
- Reported by U04PRSBKPC0 in the IBM Research workspace on 2026-07-23

## Notes
Reported via Slack. Needs investigation into how per-thread context is scoped/persisted when the agent generates replies.

pilartomas:
Fixes #2952.

## The bug

A Slack agent doesn't post directly — it calls a `reply`/`react` tool back on the api-server, and that call is identified only by the agent's pod (there is no turn/session id on it). To know which thread to answer in when the agent omits the id, the worker consulted a single per-agent pointer holding the *most recent* inbound turn.

A single agent pod multiplexes every thread over one harness process, and mention/DM turns are not serialized. So when a second turn started while the first was still in flight, it overwrote that pointer — and the first turn's id-less reply landed in the **second** turn's thread. This is exactly the live repro on the issue: an agent mid-reply in thread A, a new mention starts thread B, and A's answer surfaces in B.

## Why not correlate at the transport level

The ideal fix — carry the ACP session id on the reply call so it resolves deterministically — isn't reachable here: one harness process per pod multiplexes all sessions, the reply endpoint is a static per-agent config with no session axis, and closing the gap would require the third-party harness binary to attach the live session id per tool call. So the fix is worker-side.

## The fix

- Replace the single "most recent turn" pointer with the **set of turns currently driving the harness** per agent, plus a retained last-turn pointer for proactive replies made outside any live turn (preserves today's behavior).
- Resolve an id-less `reply`/`react` against that set:
  - **exactly one** live turn → that turn (single-turn ergonomics unchanged);
  - **none** → the last-turn fallback (proactive/scheduled replies, unchanged);
  - **more than one** → refuse with an error naming the `threadTs` the per-turn prompt already injects, rather than guess.
- Register the foreign-fork turn when it actually drives the harness (not at provisioning time), which also removes a pre-existing cross-route in that window.

The prompt-injected `threadTs` already wins over the fallback, so a well-behaved agent is unaffected. The only behavior change is that a silent cross-thread mis-route becomes, at worst, a refused call the agent retries with the id it was given — never a wrong-thread delivery.

## Testing

- New concurrency regression tests in `slack-turn.test.ts`: two turns in flight → id-less `reply`/`react` refused; explicit `threadTs` still lands in the right thread; a sole in-flight turn resolves without an id.
- `mise run api-server:{check:tsc,check:lint,check:format}` clean; full api-server suite green (607 tests).
- Change was adversarially reviewed for turn-leaks and residual mis-routes.

## Follow-up (separate concern)

#2952 as originally filed also described unrelated content appearing *inside* a first reply — a distinct history-seeding issue, tracked separately in #2954. Not addressed here.

## Docs

`docs/architecture/channels.md` updated to describe the new outbound resolution.


packages/api-server/src/modules/channels/infrastructure/slack.ts
```diff
@@ -397,16 +397,63 @@ export function createSlackWorker(
...
+  }
+
+  /** Resolve the turn a reply/react targets when the agent passed no ids:
+   *  the sole live turn, else the last-active-thread fallback, else `ambiguous`
+   *  when several turns are live at once (the agent must name the thread). */
+  function resolveTurn(
+    instanceName: string,
+  ): { ref: TurnRef } | { ambiguous: true } | { none: true } {
+    const live = inFlightTurns.get(instanceName);
+    if (live && live.size === 1) return { ref: [...live][0]! };
```
dam-code-guardian, line 447: 🟢 **Suggestion:** `[...live][0]!` allocates an intermediate array to read the sole element. `live.values().next().value` reads it directly. Cosmetic — the `size === 1` guard already makes this safe.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `768014b`

## PR #2956: fix(channels): stop Slack replies crossing threads under concurrency
**Author:** pilartomas | **Branch:** fix/slack-cross-thread-reply-routing → main | **Changes:** +216 −32 (4 files)

### Summary
Replaces the single per-agent "most recent inbound turn" pointer — which a concurrent turn could overwrite, mis-routing an id-less `reply`/`react` into the wrong thread (#2952) — with a set of in-flight turns plus a retained last-turn fallback. An id-less call resolves to the sole live turn, falls back to the last thread when none is live, and refuses (rather than guesses) when several turns are in flight.

### Findings
- ✅ **Looks good:** The resolution contract (`resolveTurn`) is precise and matches the docs/tests: 1 live → that turn, >1 → `ambiguous`, 0 → `lastTurn` fallback or `none`. Refusing on ambiguity is the right call — a refused, retryable error is strictly safer than a silent cross-thread post.
- ✅ **Looks good:** `beginTurn`/`endTurn` are balanced across every turn path — main relay (inside `try`, so setup throws can't leak an entry), ambient, and fork-ready — and the fork's turn is correctly registered in `handleForkOutcome` (when it drives the harness) rather than at provisioning, closing a pre-existing spurious-target window. `endTurn` deletes the empty set, so the maps don't grow unbounded.
- ✅ **Looks good:** `TurnRef` identity is used as the `Set` key, so the exact object registered in `beginTurn` is the one removed in the `finally` — concurrent turns on the same agent stay distinct even with identical thread/message ts.
- ✅ **Looks good:** Strong concurrency regression coverage in `slack-turn.test.ts` — the gated-harness holds two turns in flight and asserts both the refusal and that an explicit `threadTs` still lands correctly, plus the sole-turn resolution.
- 🟢 **Suggestion:** `resolveTurn` uses `[...live][0]!` to read the sole element — fine given the `size === 1` guard, though `live.values().next().value` avoids the intermediate array. Cosmetic. (`packages/api-server/src/modules/channels/infrastructure/slack.ts:447`)

**Verdict:** aligned. `docs/architecture/channels.md` was updated in-PR to describe the new id-less resolution (single live turn → that turn; none → last-active fallback; multiple → refuse), and `Last verified:` was bumped to 2026-07-24. No ADR leaks, no volatile content, no missing pages.

✅ No findings. The change is contained within the `channels` module's infrastructure layer; the `services/channel-manager.ts` edits are docstring-only on existing port interfaces. No layer-rule or module-boundary violations.

### Verdict
APPROVE — a correct, well-scoped worker-side fix for the cross-thread mis-route with balanced turn lifecycle handling, strong concurrency tests, and in-PR docs; the lone suggestion is cosmetic.

---

PR #2958, opened 2026-07-24
feat(ui): distinguish the ambient Slack session from its thread sessions

pilartomas:
## What

In the session sidebar, the channel's rolling **ambient** (reads-along) session and the **thread** sessions it spins off were both rendered as identical `#` channel rows — there was no way to tell which row was the ambient reader vs. a thread the agent got pulled into.

This marks the ambient reader with a superscript **`A`** on the channel glyph (`#ᴬ`) and adds an **`Ambient`** / **`Thread`** kind label to the row's subtitle line.

Before → after (Slack channel rows):

```
  does anyone know deploy…      #        does anyone know deploy…      #ᴬ
  Jul 24, 3:12 PM          →    Ambient · Jul 24, 3:12 PM
  Re: deploy runbook            #        Re: deploy runbook            #
  Jul 24, 3:14 PM               Thread · Jul 24, 3:14 PM
```

## Why

The ambient session is keyed with a synthetic `ambient:<channelId>` while thread sessions use the real Slack `thread_ts`, but the UI dropped `threadTs` when decoding sessions, so both surfaced as `type: ChannelSlack` with the same icon. You couldn't find the ambient reader to see what it's reading, or tell it apart from the threads.

## How

- **api-server-api** — add `threadTs` to `SessionView`; add a shared `AMBIENT_THREAD_KEY_PREFIX` / `ambientThreadKey()` / `isAmbientThreadKey()` so the Slack worker (writer) and the UI (reader) agree on the key format in one place (it previously lived only inside the worker).
- **api-server** — the Slack worker now imports the shared `ambientThreadKey` instead of its local copy (behaviour-preserving).
- **ui** — decode `threadTs` in `toSessionView`; add `slackSessionKind()` (`"ambient" | "thread" | null`, Slack-only); render the `#ᴬ` marker + subtitle label in `SessionRow`.

Telegram channel sessions are intentionally untouched (ambient is Slack-only), and non-channel rows are unchanged.

## Scope

Per the design discussion, this is presentation only: distinguish ambient vs. thread. There is **no** parent→child link recorded between an ambient session and the threads it spawns (that would need new session metadata + worker wiring) — deliberately left out.

## Tests

- New `session-category.test.ts` covering `sessionCategory` and `slackSessionKind` (ambient / thread / DM-no-key / Telegram-not-ambient / non-channel).
- `tsc`, `eslint`, `prettier` clean on all touched packages.
- Existing api-server Slack suites (126 tests incl. `slack-ambient`) pass with the rerouted helper.

## Manual verification (not yet done)

Visual confirmation in a running UI — bind a channel shared+ambient, send a top-level message and a threaded follow-up, and confirm the ambient row shows `#ᴬ` / `Ambient` and the thread row shows `#` / `Thread`.

### DIFF
--- a/packages/ui/src/modules/sessions/components/session-row.tsx
+++ b/packages/ui/src/modules/sessions/components/session-row.tsx
@@ -18,6 +18,7 @@ import {
 import { cn } from "@/lib/utils";
 
 import { formatSessionTimestamp } from "../lib/format-session-timestamp.js";
+import { slackSessionKind } from "../lib/session-category.js";
 import { WorkingDots } from "./working-dots.js";
 
 const LONG_PRESS_MS = 400;
@@ -98,6 +99,9 @@ export function SessionRow({
   const channel =
     s.type === SessionType.ChannelSlack ||
     s.type === SessionType.ChannelTelegram;
+  // Slack channel sessions split into the channel's rolling ambient reader and
+  // the threads it spins off; the ambient one wears an extra "A" marker.
+  const slackKind = slackSessionKind(s);
 
   return (
     <div
@@ -126,11 +130,15 @@ export function SessionRow({
             scheduled={scheduled}
             terminal={terminal}
             channel={channel}
+            ambient={slackKind === "ambient"}
             needsApproval={needsApproval}
             working={working}
           />
         </div>
         <span className="text-[11px] text-muted-foreground">
+          {slackKind
+            ? `${slackKind === "ambient" ? "Ambient" : "Thread"} · `
+            : ""}
           {formatSessionTimestamp(s.updatedAt ?? s.createdAt)}
         </span>
       </div>
@@ -187,12 +195,14 @@ function SessionIndicators({
   scheduled,
   terminal,
   channel,
+  ambient,
   needsApproval,
   working,
 }: {
   scheduled: boolean;
   terminal: boolean;
   channel: boolean;
+  ambient: boolean;
   needsApproval: boolean;
   working: boolean;
 }) {
@@ -203,9 +213,29 @@ function SessionIndicators({
       {terminal && (
         <Code size={16} className="text-text" aria-label="Terminal" />
       )}
-      {channel && (
-        <Hashtag size={16} className="text-text" aria-label="Channel session" />
-      )}
+      {channel &&
+        (ambient ? (
+          // Rolling channel reader: keep the # channel glyph, brand it with a
+          // superscript "A" so it stands apart from the threads it spins off.
+          <span
+            className="inline-flex items-start text-text"
+            aria-label="Ambient channel session"
+          >
+            <Hashtag size={16} />
+            <span
+              className="text-[9px] font-semibold leading-none text-accent"
+              aria-hidden
+            >
+              A
+            </span>
+          </span>
+        ) : (
+          <Hashtag
+            size={16}
+            className="text-text"
+            aria-label="Channel session"
+          />
+        ))}
       {scheduled && (
         <Time size={16} className="text-text" aria-label="Scheduled" />
       )}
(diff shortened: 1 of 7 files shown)


packages/ui/src/modules/sessions/components/session-row.tsx
```diff
@@ -203,9 +213,29 @@ function SessionIndicators({
...
+              A
+            </span>
+          </span>
+        ) : (
+          <Hashtag
+            size={16}
+            className="text-text"
+            aria-label="Channel session"
+          />
+        ))}
```
dam-code-guardian, line 238: 🟢 **Suggestion:** The `channel && (ambient ? ... : ...)` branch adds ~13 lines of JSX to this leaf, pushing `SessionIndicators` past the ~25-line render guideline. Consider extracting a small `ChannelGlyph({ ambient }: { ambient: boolean })` component so this stays `{channel && <ChannelGlyph ambient={ambient} />}`. Non-blocking readability nit.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `205c5da`

## PR #2958: feat(ui): distinguish the ambient Slack session from its thread sessions
**Author:** pilartomas | **Branch:** feat/ambient-session-visibility → main | **Changes:** +187 −13 (7 files)

### Summary
Surfaces the channel's rolling **ambient** (reads-along) Slack session distinctly from the **thread** sessions it spawns: adds `threadTs` to `SessionView`, a shared `ambientThreadKey()`/`isAmbientThreadKey()` in the contract package (deduping the copy that lived in the Slack worker), and renders a `#ᴬ` glyph + `Ambient`/`Thread` subtitle label in `SessionRow`. Presentation-only — no parent→child session link.

### Findings
- 🟢 **Suggestion:** `SessionIndicators` now renders a ~13-line nested ternary for the channel glyph, pushing the leaf's JSX past the readability threshold; extracting a small `ChannelGlyph({ ambient })` component would keep it a clean leaf. (`packages/ui/src/modules/sessions/components/session-row.tsx:216`)
- ✅ **Looks good:** The key format is centralized once in the contract package (`api-server-api`) and imported by both writer (Slack worker) and reader (UI) — the previous duplicated `ambientThreadKey` is removed, eliminating the drift risk the PR set out to fix.
- ✅ **Looks good:** `isAmbientThreadKey` is null/undefined-safe (`!!threadKey && startsWith(...)`), and `slackSessionKind` correctly gates on `ChannelSlack` so Telegram and non-channel rows return `null` even if a key carried the prefix — matches the test matrix.
- ✅ **Looks good:** Ambient marker is accessible — `aria-label="Ambient channel session"` on the wrapper, decorative `A` is `aria-hidden`. New `threadTs` decode in `toSessionView` defaults to `null`, so existing rows are unaffected.

✅ No findings. Presentation-only UI change; no architecture page's subsystem behavior/responsibility changed. The ambient-vs-thread routing model is already documented in `docs/architecture/channels.md` ("Ambient routing: one session per thread, plus one per channel"), whose `Last verified:` is already current.

✅ No findings. The shared key helpers and the `threadTs` field correctly live in the contract package as the single source of truth for a format both writer and reader depend on; re-exported through the module `index.ts` public boundary. The Slack worker (infrastructure) importing the helper is a legal dependency direction; no business logic leaked into the contract package.

🟢 **Suggestion (MODERATE):** `SessionIndicators`' render block grew to ~40 lines of JSX; the channel-glyph branch is a good extraction candidate (`ChannelGlyph({ ambient })`) to restore a ≤25-line leaf. No CRITICAL/HIGH violations — styling is Tailwind-only (no static `style`), props use `interface`/destructuring, no `any`/`as`, and the marker is accessible.

### Verdict
APPROVE — Clean, well-tested, well-scoped presentation change that removes a duplicated key helper; the only note is a non-blocking readability suggestion.

---

PR #2959, opened 2026-07-24
feat(ui): show real skill descriptions in "Created in this sandbox"

### LINKED ISSUE #2957: Show real skill descriptions instead of "only in this sandbox"
## Problem

On the sandbox Skills page, every skill created in the sandbox shows the same placeholder subtitle — "only in this sandbox" — instead of its actual description. Skills are hard to tell apart at a glance, and the description the author wrote in the skill's frontmatter never shows up.

## Proposed solution

Show each skill's real description as its subtitle. Fall back to the "only in this sandbox" placeholder only when a skill genuinely has no description.

The data is already there end to end — the frontmatter `description` is read on the pod and carried through the API; the UI just isn't rendering it today. So this is a small change on the UI side, no backend work.

### Scope

- The "Created in this sandbox" group. The same treatment extends naturally to the image-baked skills once #2828 separates them out.

### Dependencies

- Deferred from #944. Related to #2828 (provenance) but independent — showing descriptions doesn't need the system-vs-user split.

### Out of scope

- Distinguishing system (image-baked) skills from user-authored ones — that's #2828.

PetrBulanek:
On the sandbox Skills page, each skill created in the sandbox now shows its own description — the one the author wrote in the skill's frontmatter — instead of the same "only in this sandbox" placeholder on every row. Skills that genuinely have no description show "No description" instead.

Closes #2957

### DIFF
--- a/packages/ui/src/modules/sandboxes/components/skills/standalone-skills-group.tsx
+++ b/packages/ui/src/modules/sandboxes/components/skills/standalone-skills-group.tsx
@@ -103,8 +103,13 @@ export function StandaloneSkillsGroup({
                 <p className="truncate text-[15px] font-medium text-foreground">
                   {skill.name}
                 </p>
-                <p className="truncate text-[13px] text-muted-foreground">
-                  only in this sandbox
+                <p
+                  className={cn(
+                    "truncate text-[13px] text-muted-foreground",
+                    !skill.description && "italic",
+                  )}
+                >
+                  {skill.description || "No description"}
                 </p>
               </div>
 


packages/ui/src/modules/sandboxes/components/skills/standalone-skills-group.tsx
```diff
@@ -103,8 +103,13 @@ export function StandaloneSkillsGroup({
                 <p className="truncate text-[15px] font-medium text-foreground">
                   {skill.name}
                 </p>
-                <p className="truncate text-[13px] text-muted-foreground">
-                  only in this sandbox
+                <p
```
dam-code-guardian, line 106: 🟢 **Suggestion:** This `<p>` is `truncate`d with no `title`, so a description clipped by the ellipsis isn't recoverable on hover. Consistent with the `skill.name` row above (same pattern), so not a blocker — but `title={skill.description}` here (and on the name row) would be a small a11y/UX win.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `4b6f44a`

## PR #2959: feat(ui): show real skill descriptions in "Created in this sandbox"
**Author:** PetrBulanek | **Branch:** feat/2957-skill-descriptions → main | **Changes:** +7 −2 (1 file)

### Summary
Replaces the hard-coded "only in this sandbox" placeholder on every standalone-skill row with the skill's own frontmatter `description`, falling back to an italicized "No description" when it's empty. Closes #2957.

### Findings
- ✅ **Looks good:** Conditional styling done via `cn()` rather than a static `style` object — matches the styling conventions and the sibling `div`'s pattern (`standalone-skills-group.tsx:107`).
- ✅ **Looks good:** `skill.description || "No description"` is the right fallback — `localSkillSchema` types `description` as a required `string` that can be empty, so `||` (not `??`) correctly catches the empty-string case; the `!skill.description && "italic"` guard fires on exactly the same condition (`standalone-skills-group.tsx:108,111`).
- 🟢 **Suggestion:** The description `<p>` is `truncate`d with no `title` attribute, so a long description clipped by the ellipsis isn't recoverable on hover. This mirrors the existing `skill.name` row (also `truncate`, no `title`), so it's consistent — adding `title={skill.description}` on both would be a small a11y/UX win but is out of scope here (`standalone-skills-group.tsx:106`).

✅ No findings. The change is presentational row copy in one component and alters no subsystem behavior/responsibility. `docs/architecture/skills.md` documents transport/scan/standalone-vs-installed semantics, not individual row rendering, and no architecture page was edited (no `Last verified:` bump owed).

✅ No findings. Idiomatic conditional styling via `cn()`, correct empty-string fallback semantics, and styling that matches the sibling row. One optional 🟢 (`title` tooltip) noted above.

### Verdict
APPROVE — a small, correct copy/styling change with the right empty-string fallback and idiomatic conditional class composition; no correctness, security, or maintainability concerns.

---

PR #2961, opened 2026-07-24
fix(channels): coalesce ambient thread messages per thread

pilartomas:
## Problem

Read-along (**ambient**) messages inside a Slack thread were being **silently dropped** during a burst, while a later direct `@mention` landed immediately — making the agent look like it ignored the earlier messages and only woke up when pinged.

### Root cause

The ambient **thread** path had no serialization/coalescing guard — unlike the top-level channel ambient path, which already coalesces. Each thread message spawned its own concurrent turn, all resuming the *same* per-thread session:

- `handleChannelMessage` relayed each thread message directly (`relayAmbientTurn`), concurrently.
- Those concurrent turns collide in the agent-runtime's **per-session prompt queue**. When a queued turn's short-lived relay connection tears down before the agent dequeues it, the queued prompt is **dropped silently** (`detach()` filters it out; no error surfaces).
- Ambient turns swallow every error by design (never post, never show status), so the dropped messages vanished with **no trace** in Slack or the session.
- The `@mention` used the visible owner-turn path *and* arrived after the burst had cleared, so it hit an idle session and was forwarded straight through — landing immediately while the earlier ambient messages disappeared.

The code even documented the tradeoff it was relying on: *"thread sessions keep the mention path's no-guard semantics — ambient thread replies are comparatively rare."* A rapid thread burst is exactly the case that assumption didn't plan for.

## Fix

Route **both** thread and top-level ambient traffic through the same queue, keyed **per session**:

- a **thread** by its `thread_ts` (the same session a mention there resumes),
- the channel's **top-level** flow by its rolling ambient key.

Distinct sessions still drain concurrently on their own queues, but a burst *within* one session is serialized and coalesced into a single multi-message turn — so nothing races into the shared session and nothing gets dropped. Fully-drained queues are removed so per-thread entries don't accumulate for the channel's lifetime.

## Changes

- Generalize `enqueueAmbient` / `drainAmbientQueue` to key on `(channel, threadTs)`; `threadTs === null` = the channel's top-level ambient session.
- `handleChannelMessage` now enqueues thread messages instead of relaying them directly.
- Tests: thread-burst coalescing, and per-thread queue independence (distinct threads run concurrently).
- Update the `channels` architecture doc to reflect that **all** ambient traffic is now serialized+coalesced per session.

## Verification

- `mise run api-server:test -- slack` → 128 passing (incl. 2 new tests)
- `api-server:check:tsc`, `:check:lint`, `:check:format` all clean

## Follow-up (not in this PR)

A direct `@mention` still serializes behind queued ambient turns on the same thread session rather than jumping the line. Out of scope here; worth a separate look if mention latency during ambient bursts becomes a concern.


packages/api-server/src/modules/channels/infrastructure/slack.ts
```diff
@@ -1574,24 +1594,33 @@ export function createSlackWorker(
...
+        // reply thread under the batch's newest message.
+        const inThread = queue.threadTs !== null;
         await relayAmbientTurn({
           instanceName: binding.instanceName,
-          channel: channelId,
-          threadKey: ambientThreadKey(channelId),
-          replyThreadTs: last.eventTs,
+          channel: queue.channelId,
+          threadKey: inThread
+            ? queue.threadTs!
```
dam-code-guardian, line 1619: 🟢 **Suggestion:** The `queue.threadTs!` non-null assertions here and on the next line are safe under the `inThread` guard, but destructuring a narrowed local would let you drop the `!`:
```suggestion
          threadKey: inThread
            ? queue.threadTs!
```
(Minor style nit — the `!` is correct as written.)

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `ec3f488f`

## PR #2961: fix(channels): coalesce ambient thread messages per thread
**Author:** pilartomas | **Branch:** fix/slack-ambient-thread-coalescing → main | **Changes:** +128 −44 (3 files)

### Summary
Fixes silent drops of ambient (read-along) Slack messages during a thread burst. Previously only top-level channel ambient traffic was serialized/coalesced; thread replies relayed directly and concurrently, colliding in the runtime's per-session prompt queue, which silently drops the losers. The fix routes **all** ambient traffic through one coalescing queue keyed per session — `top:<channel>` for the channel's rolling ambient session and `thread:<channel>:<threadTs>` for each thread — so distinct sessions still drain concurrently while a burst within one session serializes into a single multi-message turn.

### Findings
- ✅ **Looks good:** The per-session queue key (`ambientQueueKey`) correctly distinguishes top-level (`threadTs === null`) from thread flow, and is un-collidable across the two namespaces.
- ✅ **Looks good:** Race safety is preserved — the `draining` reset and the new `ambientQueues.delete(key)` both sit in the `finally` with no `await` between the empty check and the reset, so a concurrent `enqueueAmbient` can't slip a message past both; the delete is guarded on `pending.length === 0`, so a batch left pending by the catch block isn't discarded.
- ✅ **Looks good:** The ToU/binding re-check at drain time now reads `queue.channelId` consistently, and `relayAmbientTurn` receives the right `threadKey`/`replyThreadTs`/`hasThread` for both paths.
- 🟢 **Suggestion:** The `queue.threadTs!` non-null assertions (`slack.ts:1619,1621`) are safe under the `inThread` guard, but a narrowed local would drop the `!`. Minor style nit.
- ✅ **Looks good:** Two new tests cover thread-burst coalescing into one prompt and per-thread queue independence, including the explicit "keying by channel alone would wrongly block" case.

**Verdict: aligned.** `docs/architecture/channels.md` was updated in the same PR, `Last verified:` bumped to 2026-07-24, and the revised "Ambient routing" bullet accurately describes all-ambient per-session serialization/coalescing with independent queues. No ADR leaks, no volatile content. ✅ No findings.

**Verdict: aligned.** Change stays within the `infrastructure/` Slack adapter — no cross-layer imports, no new optional dependencies, no domain/service boundary violations. The `AmbientQueue` type and `ambientQueueKey` are appropriately local adapter state. ✅ No findings.

### Verdict
APPROVE — a well-scoped, well-tested fix for a real silent-drop bug; the coalescing generalization is race-safe and the architecture doc was updated in lockstep.

---

Issue #2898, opened 2026-07-22
UI - Usage spend by agent chart


tomkis:
## Parent

#2877

## What to build

Add a Spend by agent horizontal bar breakdown to the Settings > Usage tab, sorted highest to lowest, so the biggest cost drivers are obvious at a glance. Uses the existing month switcher.

End-to-end slice:

- New `metrics.spendByAgent` tRPC procedure. Input is the existing from/to instant range. Output rows: `{ agentId, agentName, costUsd }`, sorted by cost descending. All agents with Spend in range are returned, no top-N cap.
- Aggregation happens in ClickHouse, grouped by Agent Attribution (the gateway-stamped `platform.agent.id`, the trusted key). The display name comes from the telemetry rows themselves (latest `platform.agent.name` observed in range), so deleted agents keep a readable name. The name is display-only; the id stays the key.
- Same ownership scoping and disabled-backend guard as `metrics.spend` (caller's owned agents, deleted included).
- UI renders hand-rolled Tailwind horizontal bars with agent name and USD label per row. We are deliberately not adding a chart library.

## Acceptance criteria

- [ ] Usage tab shows Spend by agent bars sorted highest to lowest for the selected month
- [ ] Grouping key is the gateway-stamped agent id; names come from telemetry, so deleted agents display their last known name
- [ ] All agents with Spend are shown, no cap
- [ ] Ownership scoping matches `metrics.spend`; a disabled metrics backend yields the existing unavailable state
- [ ] No new chart dependency added
- [ ] `docs/architecture/metrics.md` updated with the new procedure
- [ ] `mise run check` and relevant tests pass

## Blocked by

- #2896

tomkis:
🤖 Ralph: implemented on `ralph/issue-2877` in commit `55ce51ff` (new metrics.spendByAgent tRPC procedure, ClickHouse aggregation on the gateway-stamped agent id, hand-rolled Tailwind bars in usage-view, docs + unit tests). The orchestrator was recycled by a session teardown before it could close this; closing now to reflect the already-pushed work.

Issue #2897, opened 2026-07-22
UI - Usage spend by day chart


tomkis:
## Parent

#2877

## What to build

Add a Spend by day column chart to the Settings > Usage tab, showing per-day LLM Spend for the month selected in the existing month switcher. The previous-month ghost overlay from the parent mockup was cut from scope (see parent issue comment).

End-to-end slice:

- New `metrics.spendByDay` tRPC procedure. Input is the existing from/to instant range plus an IANA timezone string supplied by the client (from `Intl.DateTimeFormat`). Output is sparse: one row per local calendar day that has Spend, shaped `{ day: "YYYY-MM-DD", costUsd }`. Days without Spend are omitted.
- Aggregation happens in ClickHouse, grouping call rows into days in the given timezone. Same ownership scoping and disabled-backend guard as the existing `metrics.spend` procedure (caller's owned agents, deleted included).
- UI renders a hand-rolled Tailwind column chart. We are deliberately not adding a chart library. The client owns calendar semantics: it generates the full day list for the selected month, fills zeros for days missing from the response, and for the current month renders nothing after today. Hover reveals the exact USD amount per day.

## Acceptance criteria

- [ ] Usage tab shows a Spend by day column chart for the selected month: one column per day including zero-spend days, no columns after today when viewing the current month
- [ ] Day buckets align with the user's local calendar days (timezone passed from the client, grouping done in ClickHouse)
- [ ] Server response is sparse; zero-filling happens client-side
- [ ] Ownership scoping matches `metrics.spend` (own agents only, deleted included); a disabled metrics backend yields the existing unavailable state
- [ ] No new chart dependency added
- [ ] `docs/architecture/metrics.md` updated with the new procedure
- [ ] `mise run check` and relevant tests pass

## Blocked by

- #2896

tomkis:
🤖 Ralph (cycle #1) implemented this on `ralph/issue-2877`.

Added a Spend-by-day column chart to Settings > Usage end-to-end: a new metrics.spendByDay tRPC procedure (from/to range + client IANA timezone) that groups spend into local calendar days in ClickHouse and returns sparse rows, plus a hand-rolled Tailwind column chart that zero-fills the month client-side and stops at today for the current month. Ordering/scoping mirror the already-merged spend-by-agent slice; docs updated.

commit: `3f321dff`

[...]

Issue #2896, opened 2026-07-22
Document the metrics read path (architecture page + glossary)


tomkis:
## Parent

#2877

## What to build

Bootstrap documentation for the metrics read path so the two feature slices under #2877 have an architecture page to extend. Docs-only prefactor slice.

The api-server metrics module serves the Settings > Usage tab: per-user LLM spend aggregates read live from ClickHouse `otel_logs`. The observability architecture page deliberately scopes itself to the export path and calls the user-facing read path a separate concern. Nothing documents that concern today, and the project rule is that behavior changes go through the subsystem's architecture page.

Create `docs/architecture/metrics.md` covering the existing read path:

- The tRPC contract (`metrics.overview`, `metrics.spend`) and what each returns
- Ownership scoping in the service layer, including why deleted agents stay in scope (history must not shrink)
- The ClickHouse reader and the `otel_logs` row shape it relies on (`cost_usd_micros`, token counters, model, timestamp)
- Disabled-backend behavior when no ClickHouse URL is configured (service throws `PRECONDITION_FAILED`, UI shows the unavailable message)
- Trust story: token/cost counters are agent-reported; agent attribution (`platform.agent.id`) is gateway-stamped and unforgeable; `platform.agent.name` is display-only

Link the new page from `docs/architecture.md` and from the sentence in `docs/architecture/observability.md` that scopes the read path out.

Also add a Metrics bounded context to `docs/ubiquitous-language.md` with the terms Spend, Spend Breakdown, and Agent Attribution, and fix the Budgets preamble, which currently mislabels Usage Tracking as "tokens/cost accounting" (Usage Tracking is activity analytics; tokens/cost is Spend under Metrics).

Follow `docs/guidelines/documentation-guidelines.md`.

## Acceptance criteria

- [ ] `docs/architecture/metrics.md` exists, follows the documentation guidelines, and covers the contract, ownership scoping, ClickHouse reader, disabled-backend behavior, and trust story
- [ ] `docs/architecture.md` and `docs/architecture/observability.md` link to the new page
- [ ] `docs/ubiquitous-language.md` has a Metrics bounded context (Spend, Spend Breakdown, Agent Attribution) and the Budgets preamble no longer describes Usage Tracking as tokens/cost accounting
- [ ] `mise run check` passes

## Blocked by

None - can start immediately

tomkis:
🤖 Ralph (cycle #1) implemented this on `ralph/issue-2877`.

[...]

PR #2962, opened 2026-07-24
Ralph: UI - Expand Usage to cover spend by day and by agent

### LINKED ISSUE #2877: UI - Expand Usage to cover spend by day and by agent
## Description

Today the Settings > Usage screen shows only a total spend figure and a Spend by model table. That answers "how much did I spend" but not "when" or "on what".

Expand the Usage page with two new sections so users can see spend trends over time and identify which agents are the biggest cost drivers:

1. Spend by day — a per-day column chart for the selected month, with a faint "ghost" overlay of the previous month for comparison.
2. Spend by agent — a horizontal bar breakdown of spend per agent, sorted highest → lowest, so the most expensive agents are obvious at a glance.

## Current state

Usage shows Total spend + Spend by model only - no time trend, no per-agent breakdown.

Image

## Desired state

Usage additionally shows Spend by day (with previous-month ghost) and Spend by agent, using the same month switcher already in the header.

Image

tomkis:
Ralph Loop implementing #2877 — one sub-issue per commit, each built in its own fresh agent context.

**This PR will solve:** UI - Expand Usage to cover spend by day and by agent

### Sub-issues
- [ ] #2898 UI - Usage spend by agent chart
- [ ] #2897 UI - Usage spend by day chart
- [ ] #2896 Document the metrics read path (architecture page + glossary)

Closes #2877.


packages/ui/src/modules/metrics/components/metrics-panel.tsx
```diff
@@ -186,6 +188,93 @@ export function ModelSpendTable({ rows }: { rows: TokenSpendByModel[] }) {
...
+  const max = days.reduce((m, d) => Math.max(m, d.costUsd), 0);
+  return (
+    <div className="flex gap-2">
+      <div className="flex h-40 w-12 flex-col justify-between py-px text-right text-[10px] tabular-nums text-text-secondary">
+        <span>{formatUsd(max)}</span>
+        <span>$0</span>
+      </div>
+      <div className="min-w-0 flex-1">
+        <div className="flex h-40 items-end gap-px">
+          {days.map((d) => {
```
dam-code-guardian: 🟢 **Suggestion:** this per-column `.map` body is ~18 lines with a per-item derivation (`dayNum`) and a conditional height expression — over the threshold for extracting the map item into its own leaf. A small `DayColumn` component would let the parent read `{days.map((d) => <DayColumn key={d.day} day={d} max={max} />)}`, and the odd-day axis label (mapped over `days` a second time below) could ride along with it rather than re-deriving `dayNum`.

packages/ui/src/modules/metrics/components/metrics-panel.tsx
```diff
@@ -186,6 +188,93 @@ export function ModelSpendTable({ rows }: { rows: TokenSpendByModel[] }) {
...
 }
 
+/** Hand-rolled horizontal bars — one per agent, widest is the top spender.
+ *  Rows arrive sorted highest cost first, so the first row sets the scale.
+ *  Deliberately no chart library. */
+export function AgentSpendBars({ rows }: { rows: SpendByAgent[] }) {
+  const max = rows[0]?.costUsd ?? 0;
+  return (
+    <div className="flex flex-col gap-3">
+      {rows.map((row) => {
```
dam-code-guardian: 🟢 **Suggestion:** same pattern here — the per-row body is a candidate for extraction into an `AgentSpendBar` leaf so the parent map stays thin. Non-blocking.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `3f321df`

## PR #2962: Ralph: UI - Expand Usage to cover spend by day and by agent
**Author:** tomkis | **Branch:** ralph/issue-2877 → main | **Changes:** +494 −18 (14 files)

### Summary
A Ralph-loop implementation of #2877 in three slices: a docs-only prefactor creating `docs/architecture/metrics.md`, then two end-to-end feature slices adding `metrics.spendByAgent` (per-agent horizontal bars) and `metrics.spendByDay` (per-day column chart) tRPC procedures to the Settings → Usage tab. Both new reads reuse the existing owner-scoping gate and `[from, to)` range; spend-by-day adds a client-supplied IANA timezone for wall-clock day bucketing.

### Findings
- 🟢 **Suggestion:** `SpendByDayChart`'s per-column `.map` body is ~18 lines of JSX with a per-item derivation and conditional height — extracting a `DayColumn` leaf would keep the parent map thin (`packages/ui/src/modules/metrics/components/metrics-panel.tsx:240`).
- 🟢 **Suggestion:** `AgentSpendBars`'s row body is similarly extractable into a leaf component (`packages/ui/src/modules/metrics/components/metrics-panel.tsx:198`).
- ✅ **Looks good:** Timezone validated in the Zod schema (`isValidTimeZone` refinement) so a garbage zone fails as bad input, not a store query error (`schemas.ts:36`).
- ✅ **Looks good:** Both new service methods go through the same `ownedScope()` gate (empty-owned → `[]`), and the disabled variant implements all four methods so it fails closed (`metrics-service.ts:104,123`).
- ✅ **Looks good:** ClickHouse day-bucketing uses `toDate(toTimeZone(Timestamp, {timeZone:String}))` with the tz passed as a bound param — no interpolation, wall-clock-correct buckets.
- ✅ **Looks good:** Unit tests cover scope, range passthrough, timezone forwarding, empty-owned, and disabled-fail-closed for both new procedures.

**Verdict:** aligned. The `metrics` subsystem's behavior change (two new procedures) is reflected in `docs/architecture/metrics.md` in the same PR (Contract section grown 2→3→4 procedures, reader section documents the display-name + timezone-shift reads); `Last verified:` bumped to 2026-07-24 on all three edited pages; new page linked from the landing index and the observability hand-off. No ADR leaks, no volatile-content additions. No drift.

✅ No findings. New procedures follow the contract→service→port→adapter path exactly: Zod validation in the contract package, router delegates with no business logic, service depends on the `MetricsReader` port, ClickHouse adapter implements it, ownership scoping stays in the application layer, and the disabled service fully implements the interface.

2 🟢 suggestions (leaf-component extraction in the two hand-rolled charts, listed under Findings). Otherwise clean: server state via TanStack Query hooks in `api/`, no raw fetch, no `any`/`as`, calendar logic kept as a pure helper, and the dynamic `style={{ width/height }}` on bars is the permitted runtime-value use.

### Verdict
APPROVE — a clean, well-tested, well-documented end-to-end feature; the only findings are two non-blocking maintainability suggestions.

---

tomkis:
/iterate The UI seems to be a bit off from original design. See attached files.

1. Each block doesn't seem to be properly spaced vertically
2. Text colors &  font-weight does not seem to be aligned with the design on many places
3. Height & Width of the bars  in chart is not correct
4. Spend by days chart is lacking horiziontal lines
5. Spacing inside spend by day chart seems a bit off

# Expectation
Screenshot 2026-07-27 at 14 07 51

# Reality
Screenshot 2026-07-27 at 14 09 48

tomkis:
🤖 Ralph (cycle #1) addressed this on `ralph/issue-2877`.

Realigned the Usage tab with the design per @tomkis's review: even vertical spacing between blocks, larger/bolder total-spend figure and design-matched text colors, taller day chart with round-number horizontal gridlines and correctly-scaled bars, and agent bars relaid out with a right-aligned label gutter, thicker bars and right-aligned amounts.

commit: `270e334f`

files: `packages/ui/src/modules/metrics/components/metrics-panel.tsx`, `packages/ui/src/modules/metrics/views/usage-view.tsx`

tomkis:
/iterate

Visual QA of the Usage tab against the design. Everything below is styling only, no data or API changes. I compared screenshots pixel by pixel; the mock renders at about 0.93x the app's scale, so target numbers are normalized to the app and are approximate within a pixel or two.

Scope note: the previous-month comparison series (gray ghost bars behind each day plus the July/June legend in the mock) is **descoped**. Ignore it. Do not add a second data fetch, do not add a legend.

All line refs are against `270e334f`.

## 1. Spend by day chart

`packages/ui/src/modules/metrics/components/metrics-panel.tsx` (`SpendByDayChart`, from line 245)

- [ ] **Gridlines are too dark and too blue.** They use `border-border-light` (`#dde1e6`); the design reads about `#ededec`. Add a dedicated hairline token in `src/App.css` next to the existing `--c-*` colors (light `#ecedf0`, dark around `#2a2a2a`), map it in the `@theme` block as `--color-border-hairline`, and use it here. Same token gets reused by the table dividers below, so add it once.
- [ ] **Bars are too fat.** Currently a 16px bar in a 19.6px slot (82% fill). The design is about 57% fill, which reads much lighter. Keep the `flex-1` slot, center the bar in it, and give the bar a percentage width around `58%` rather than `w-full` so it stays proportional at any container width.
- [ ] **Plot is too short.** `CHART_H` is 224 (line 252), design is about 240.
- [ ] **Y axis labels are too small and the wrong family.** 10px sans (line 258) → about 12-13px and `font-mono`. The design's `$80` / `$0` show the IBM Plex Mono dotted zero, so this is the app's existing `--font-mono`, no new font needed. The vertical distribution is already correct, I checked label centers against gridlines in both and they match.
- [ ] **X axis day numbers** should also be `font-mono`, same 10-11px size. Keep the odd-days-only rule, that matches the design.
- [ ] **Axis tick formatting is inconsistent.** `formatUsd` yields `$0` sitting next to `$0.05`, `$0.10`, `$0.15`, and for sub-cent steps `toPrecision(2)` produces ragged widths like `$0.0025`. Add a `formatAxisUsd(value, step)` to `../lib/format.ts` that derives a fixed decimal count from the step (step >= 1 → 0 decimals, >= 0.01 → 2, otherwise 4) so a column reads `$0.00 / $0.05 / $0.10 / $0.15` or `$0 / $20 / $40`.
- [ ] **Nonzero days render as an invisible hairline.** The `Math.max(..., 0.8)` percent floor on line 288 gives about 2px. Use a px floor of about 3px instead so any day with real spend is legible.
- [ ] **Day labels sit slightly too close to the baseline.** `mt-2` (line 301) → `mt-3`, design is about 16px from baseline to label.

## 2. Spend by model table

`metrics-panel.tsx` (`ModelSpendTable`, from line 151) and `views/usage-view.tsx:132`

- [ ] **Columns are crammed against the right edge.** `max-w-0 w-full` on the model cell (line 166) pushes IN/OUT/COST together, so `43.4K` and `125` nearly touch. The design spreads them: model about 40% of card width, then three equal numeric columns with right edges at roughly 57% / 77% / 100%. Use `table-fixed` with `w-[40%]` plus three `w-[20%]`.
- [ ] **Rows are half the design height.** Row pitch is 28px (`py-1`), design is about 50px. Header band 41px → about 44px. `py-3.5` on the cells gets there.
- [ ] **Dividers are inset, the design runs them full-bleed.** Ours run 94 to 679 because the Card carries `p-4`. Move the padding off the Card (`usage-view.tsx:132` → `p-0`) and onto the cells (`px-5`) so every divider spans the full card width.
- [ ] **Divider weight.** The design uses a lighter hairline for row dividers than for the card edge; reuse the new `border-hairline` token from item 1 for both the header rule and the row rules. Leave the Card's own border alone (see out of scope).
- [ ] **Type is a step too small.** Body 12px → 13px; header 10px → 11px. Drop the `text-[12px]` on the Card at `usage-view.tsx:132` and size the cells directly.
- [ ] **Cost column weight** `font-medium` → `font-semibold`. In the design the cost column is the heaviest thing in the table.
- [ ] **Sub-cent costs break the mono alignment.** `$0.00058` is 7 glyphs against `$0.14`'s 5, which wrecks the column the mono font exists to create. Render `<$0.01` for sub-cent values and keep the exact number in a `title`. Add this as a separate formatter, do not change `formatUsd` itself (per-call costs elsewhere still want significant digits).

## 3. Spend by agent

`metrics-panel.tsx` (`AgentSpendBars`, from line 195)

- [ ] **Bars are half height.** `h-2.5` (10px, line 215) → about 20px (`h-5`). This is the single most visible difference in that card. Radius `rounded-[3px]` → `rounded` (4px).
- [ ] **Label gutter too narrow and too small.** `w-28` (112px, line 208) → about 140px; 12px → 14px; and darker, the design reads near `#3d3732`, so `text-foreground/80` (already used elsewhere in settings) rather than `text-text-secondary`. Keep right-aligned and truncating.
- [ ] **Label to bar gap** `gap-3` → `gap-4` (16px).
- [ ] **Cost is right-aligned, the design left-aligns it.** In the mock all six `$` glyphs line up at a fixed x just after the track with about a 16px gap, so `$80.30` leaves trailing space. Replace `w-20 text-right` (line 219) with a fixed roughly 80px left-aligned column, mono semibold, about 14px.
- [ ] **Min bar width** the 2% floor (line 201) is invisible on a narrow card; use an 8px floor, and keep zero-cost rows at zero width.

## 4. Page header

`views/usage-view.tsx`

- [ ] **Total spend is the wrong family and too small.** 40px sans bold (line 111) → `font-mono` at about 48px. This is the page's hero figure and the mono is what makes it read as an amount rather than a heading. Confirmed the mock's dotted zeros are the same IBM Plex Mono we already ship.
- [ ] **Total should always show cents.** `formatUsd(0)` returns `$0`, which next to a design showing `$902.03` looks broken. Force two decimals for the headline.
- [ ] **Month nav buttons need borders.** Lines 75-94 use `variant="ghost"`, the design has outlined ~28px squares with a 1px border and 6px radius around each chevron. `variant="outline"` at `size="icon-sm"` is already the right box. Chevrons should be `text-muted-foreground`, not full foreground. Keep the border on the disabled next-month button so the control stays symmetrical.
- [ ] **Description needs a max width.** The design caps it at about 460px so it wraps to two lines and clears the month nav; ours runs full width on one line. Do it by passing a wrapper span with `max-w-[460px] block` as the description rather than editing the shared `PageHeader`.
- [ ] **Keep our description copy.** The mock says "LiteLLM proxy spend across all your agents". That is wrong for this codebase: per `docs/architecture/metrics.md:48` the reader parses Claude Code's per-call export records, and LiteLLM here is just one provider connection. Our existing sentence is accurate, leave the wording and only cap the width.

## 5. States

- [ ] **Loading is a bare text line** (`usage-view.tsx:105`), so the layout jumps hard when data lands. Use skeletons at the final heights: 240px chart card, N x 50px table rows, N x 36px agent rows.
- [ ] **Error is unstyled text** (`usage-view.tsx:100`). Put it in a card-level empty state so the page keeps its shape.
- [ ] **Zero-spend month collapses the page.** `data.length === 0` (line 115) hides the chart and the agent card entirely and leaves one sentence. Render the empty message inside a card at the chart's height so switching months doesn't jump. Low priority, the design has no spec for it.
- [ ] **Dark mode**: the only new value here is the hairline token from item 1, give it a dark counterpart so the gridlines and dividers do not vanish or glare.

## Out of scope, do not touch

- Previous-month ghost bars and the legend. Descoped.
- **Page background tint.** The design sits white cards on `#f9f9f8` (effectively our existing `--c-bg` `#fafaf9`) while today `--background` and `--card` are both white, so the border does all the work and the cards look flat. Real difference, but it changes the surface every settings tab renders on. Separate ticket.
- **Section label color.** The design's `TOTAL SPEND` / `SPEND BY DAY` read about `#8a8a94`; `labelVariants` in `components/ui/label.tsx:8` uses `text-muted-foreground` `#4d5358`. `SectionLabel` is shared app-wide, so this is a global token decision, not a Usage fix. Separate ticket.
- Card border color, `--border-ui` is shared with inputs and buttons.
- No file moves and no restructuring; keep the hand-rolled charts, do not add a chart library. (`metrics-panel.tsx` hosting the Usage page's components is a real naming problem, and `MetricsPanel` itself is currently dead code, but splitting the file would bury the visual diff. Follow-up.)

## Verify before pushing

- `mise run ui:check` plus prettier formatting, and `mise run lint:fix` for import order.
- Screenshot the Usage tab at the same width as the design and diff the four regions above: header, day chart, model table, agent bars.

## Expected result (minus the shadowing)

Screenshot 2026-07-27 at 14 07 51

tomkis:
🤖 Ralph (cycle #1) addressed this on `ralph/issue-2877`.

Addressed @tomkis's visual-QA /iterate feedback on the Usage tab: styling-only changes across the day chart, model table, agent bars, page header, and loading/error/empty states, plus a new border-hairline color token and three USD formatters. No data or API changes.

commit: `415d94ea`

files: `packages/ui/src/App.css`, `packages/ui/src/modules/metrics/components/metrics-panel.tsx`, `packages/ui/src/modules/metrics/lib/format.ts`, `packages/ui/src/modules/metrics/views/usage-view.tsx`

tomkis:
🤖 Ralph (cycle #2) addressed this on `ralph/issue-2877`.

The cited review comment is Ralph's own summary of the prior cycle (commit 415d94ea = HEAD), which already implemented @tomkis's visual-QA /iterate checklist in full and correctly — I verified every item. That cycle ran with no tooling (no network/deps in the sandbox) so prettier never ran; I fixed the three prettier-formatting violations it left so `mise run ui:check` will pass: wrapped an 82-char Array.from ticks call, collapsed the queries.js import to one line, and wrapped a multi-attribute skeleton div that exceeded the 80-col print width. No behavior/styling/data/API change.

commit: `c6c0e790`

files: `packages/ui/src/modules/metrics/components/metrics-panel.tsx`, `packages/ui/src/modules/metrics/views/usage-view.tsx`

packages/ui/src/modules/metrics/views/usage-view.tsx
```diff
@@ -24,69 +52,138 @@ export function UsageView() {
...
-  const { data, isPending, isError } = useModelSpend(
-    month.toISOString(),
-    monthStart(month, 1).toISOString(),
-  );
+  const timeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+  const from = month.toISOString();
+  const to = monthStart(month, 1).toISOString();
+  const { data, isPending, isError } = useModelSpend(from, to);
+  const { data: agentData } = useAgentSpend(from, to);
+  const { data: dailyData } = useDailySpend(from, to, timeZone);
```
kapetr: The page runs three queries but only checks loading/error on useModelSpend. agentData and dailyData are used without checking their state, and fillMonthDays (line 33) turns "no data yet" into "$0 for every day". So when spendByDay is slower or fails, the user sees a correct total above a completely empty month — and because the query sets `retry: false`, one failure sticks for a minute with no error shown. 

Fix: Combine the error/loading state of all three queries, and don't render the chart until its data is actually there.

packages/api-server/src/modules/metrics/services/metrics-service.ts
kapetr, line 1: Each of the three procedures calls ownedScope on its own, and listOwnedAgentIds (app.ts:943-951) isn't cached — so one Usage page load does 3 K8s agent lists, 6 Postgres queries, and 3 ClickHouse scans for the same answer. On top of that, agents.list() runs the orphan allowed_users purge with its security log (agents-service.ts:729-735), so that write now fires three times concurrently per page view. Suggestion: collapse the three reads into one spendBreakdown procedure returning { byModel, byAgent, byDay } — the same composed-read pattern metrics.md already documents for overview. That resolves ownership once, and as a bonus makes finding #1 impossible (one query = one loading/error state).

kapetr, line 1: Each of the three procedures calls ownedScope on its own, and listOwnedAgentIds (app.ts:943-951) isn't cached — so one Usage page load does 3 K8s agent lists, 6 Postgres queries, and 3 ClickHouse scans for the same answer. On top of that, agents.list() runs the orphan allowed_users purge with its security log (agents-service.ts:729-735), so that write now fires three times concurrently per page view.

packages/ui/src/modules/metrics/components/metrics-panel.tsx
```diff
@@ -148,36 +156,42 @@ function SectionHeading({ children }: { children: React.ReactNode }) {
...
-        <tr className="text-[10px] uppercase tracking-wide text-text-muted">
-          <th className="py-1 text-left font-medium">Model</th>
-          <th className="py-1 text-right font-medium">In</th>
-          <th className="py-1 text-right font-medium">Out</th>
-          <th className="py-1 text-right font-medium">Cost</th>
+        <tr className="border-b border-border-hairline text-[11px] uppercase tracking-wide text-text-muted">
+          <th className="w-[40%] px-5 py-3.5 text-left font-medium">Model</th>
+          <th className="w-[20%] px-5 py-3.5 text-right font-medium">In</th>
+          <th className="w-[20%] px-5 py-3.5 text-right font-medium">Out</th>
+          <th className="w-[20%] px-5 py-3.5 text-right font-medium">Cost</th>
```
kapetr: pre-existent issue, but the same style is repeated several times

packages/ui/src/modules/metrics/views/usage-view.tsx
```diff
@@ -24,69 +52,138 @@ export function UsageView() {
...
+              {formatUsdCents(total)}
+            </div>
+          </section>
           {data.length === 0 ? (
-            <p className="text-[14px] text-muted-foreground">
-              No LLM calls in {monthLabel}.
-            </p>
+            <section>
+              <SectionLabel spaced>Spend by day</SectionLabel>
+              <Card className="flex h-[240px] items-center justify-center p-5">
```
kapetr: Same magic number for height `240px` in several places

packages/ui/src/modules/metrics/components/metrics-panel.tsx
```diff
@@ -186,6 +200,139 @@ export function ModelSpendTable({ rows }: { rows: TokenSpendByModel[] }) {
...
+    (_, i) => (nTicks - i) * step,
+  );
+  const CHART_H = 240;
+
+  return (
+    <div className="flex gap-3">
+      {/* Y-axis labels, one per gridline, vertically aligned to the plot. */}
+      <div
+        className="flex flex-col justify-between text-right font-mono text-[12px] tabular-nums text-text-muted"
+        style={{ height: CHART_H }}
```
kapetr: This is constant, it could be used as tailwind class, same height number is also hardcoded in `usage-view.tsx`

packages/ui/src/modules/metrics/components/metrics-panel.tsx
kapetr, line 1: This file could be splitted (almost 400 lines) into separate files for each component inside it.

kapetr (changes requested):

tomkis:
/iterate Please address all the comments mentioned by @kapetr .

tomkis:
🤖 Ralph (cycle #1) addressed this on `ralph/issue-2877`.

Addressed all five of @kapetr's review comments: collapsed the three metrics.spend/spendByAgent/spendByDay tRPC procedures into one metrics.spendBreakdown -> {byModel, byAgent, byDay} so ownership resolves once per page load (killing the 3x K8s list + orphan-purge writes) and the Usage tab renders from a single loading/error state (fixing the all-zeros-chart correctness bug), then extracted the repeated table-cell style into a CELL constant, replaced the repeated 240px magic number with an exported CHART_HEIGHT_CLASS, and split the ~400-line metrics-panel.tsx into per-component files (model-spend-table, agent-spend-bars, spend-by-day-chart) with extracted leaf components.

commit: `57ec5f0b7846a14c43a278b85295c21b440b545a`

files: `docs/architecture/metrics.md`, `packages/api-server-api/src/index.ts`, `packages/api-server-api/src/modules/metrics/router.ts`, `packages/api-server-api/src/modules/metrics/schemas.ts`, `packages/api-server-api/src/modules/metrics/types.ts`, `packages/api-server/src/__tests__/unit/metrics-service.test.ts`, `packages/api-server/src/modules/metrics/services/metrics-service.ts`, `packages/ui/src/modules/metrics/api/queries.ts`, `packages/ui/src/modules/metrics/components/agent-spend-bars.tsx`, `packages/ui/src/modules/metrics/components/metrics-panel.tsx`, `packages/ui/src/modules/metrics/components/model-spend-table.tsx`, `packages/ui/src/modules/metrics/components/spend-by-day-chart.tsx`, `packages/ui/src/modules/metrics/views/usage-view.tsx`

kapetr (approved):

PR #2964, opened 2026-07-24
fix(approvals): stop permanent inbox verdicts reporting success on verdict clash

### LINKED ISSUE #2766: "Deny forever" / "Approve permanently" can report success without changing what the agent can access
## Problem

Actioning a held request from the inbox with a permanent verdict — *approve permanently*, *approve host*, or *deny forever* — writes a standing network rule and reports the verdict as applied. But if an equivalent rule with the **opposite verdict** already exists for the agent, the new rule is silently not written: the approval is marked resolved, the audit trail records that a rule was written, and the UI confirms the action — while the agent's actual network access is unchanged.

A concrete way to hit this: expired held requests accumulate in the inbox (identical retries are only merged while they're still pending). A user cleaning up the backlog approves one entry permanently and later clicks "deny forever" on another entry for the same request — the deny changes nothing, and the agent keeps its allow rule. The same happens in reverse: an "approve permanently" after an earlier "deny forever" reports success while the agent stays blocked.

This matters in both directions. A user who believes they've permanently blocked an agent's access to a host has not (security-relevant), and a user who believes they've unblocked an agent is left debugging why requests still fail.

This is the inbox-side sibling of #2324: that fix made the *manual rule create* command detect the verdict clash and fail loudly, but the inbox verdict actions write rules through a separate path that still ignores the outcome.

## Proposed solution

A permanent inbox verdict should reflect reality:

- If an equivalent rule with the opposite verdict already exists, the action should either **update that rule to the new verdict** (the user's latest explicit decision wins) or **fail with a clear conflict message** pointing at the existing rule — mirroring how manual rule creation now behaves after #2324. It must not resolve the approval as applied while leaving enforcement unchanged.
- The audit trail should only record a rule as written when one actually was written or changed.

### Out of scope

- Merging or de-duplicating expired inbox entries of the same shape — a separate UX question, even though it's one way to reach this state.

### Dependencies

- Follow-up to #2324 / PR #2765.

PetrBulanek:
"Deny forever", "Approve permanently", and "Approve host" from the inbox previously reported success even when an equivalent network rule with the opposite verdict already existed — the rule write silently did nothing, the entry resolved, and the audit trail claimed a rule was written while the agent's actual network access stayed unchanged. A user who believed they had permanently blocked (or unblocked) a host had not.

Now, mirroring how manual rule creation behaves since #2324:

- A permanent verdict that clashes with an existing opposite-verdict rule fails with a clear conflict error pointing at the existing rule; the inbox entry stays actionable, and the refused attempt is recorded in the security log as a distinct event.
- A verdict matching an existing equivalent rule still resolves the entry as before, but the audit trail now truthfully records that no new rule was written, along with the id of the rule that governs.
- When the equivalent rule was created implicitly (by a preset or a connection), the explicit inbox decision takes ownership of it, so a later preset switch or connection revoke can't silently undo it.

Closes #2766


packages/api-server/src/__tests__/unit/egress-rules-port.test.ts
```diff
@@ -51,7 +51,7 @@ function fakeRepo(
       const r = rows.find((x) => x.id === id);
       if (r) r.status = "revoked";
     },
-    updatePromoteToManual: async () => null,
+    updateTakeOwnership: async () => null,
```
dam-code-guardian, line 54: 🟡 **Warning:** This mock swap to `updateTakeOwnership` is the only test change for the writer path, and it just returns `null`. No test asserts the new behavior the PR introduces: a `verdict-clash` throwing `CONFLICT` and leaving the inbox entry unresolved, a same-verdict duplicate resolving with `ruleWritten: false`, or take-ownership firing on a preset/connection duplicate. Since the bug being fixed is precisely a silent no-op that reported success, a regression would be invisible without coverage — worth adding writer-level and approvals-service-level cases before merge.

dam-code-guardian (approved):
🛡️ **DAM** — ✅ Code Review @ `6e166ac`

## PR #2964: fix(approvals): stop permanent inbox verdicts reporting success on verdict clash
**Author:** PetrBulanek | **Branch:** `fix/2766-inbox-verdict-rule-conflict` → `main` | **Changes:** +191 −26 (8 files)

### Summary
When an inbox "Approve permanently" / "Approve host" / "Deny forever" verdict's rule write no-oped against an equivalent active rule with the *opposite* verdict, the entry previously resolved and the audit trail claimed a rule was written while the agent's actual network access was unchanged. This makes the egress-rule writer return a discriminated outcome (`inserted` | `duplicate` | `verdict-clash`); a clash now throws `CONFLICT` before resolving and is audited as a distinct `approval.verdict_conflict` event, same-verdict duplicates resolve idempotently with a truthful audit line, and an explicit inbox decision takes ownership of an equivalent implicit (preset/connection) rule.

### Findings
- 🟡 **Warning:** No test covers the new behavior. The two touched test files only adapt existing mocks to the new return type; nothing asserts that a `verdict-clash` throws `CONFLICT` and leaves the entry unresolved, that a same-verdict duplicate resolves with `ruleWritten: false`, or that take-ownership fires on a preset/connection duplicate. This is the exact class of bug the PR fixes (a silent no-op reporting success) — a regression here would be invisible without a test. Recommend adding writer-level and approvals-service-level cases.
- ✅ **Looks good:** Clash detection is anchored on the real invariant — the `egress_rules_lookup_idx` unique index is verdict-agnostic (`agentId, host, method, pathPattern` where active), so `onConflictDoNothing` + `getActiveByTuple` returns the opposite-verdict row and `row.verdict !== input.verdict` correctly identifies the clash.
- ✅ **Looks good:** L7 reconverge is correctly skipped on the clash path (table unchanged) and the `promoted === null` race window (row revoked between read and take-ownership) is explicitly handled without auditing a takeover that didn't happen (`egress-rule-writer.ts`).
- ✅ **Looks good:** The consumer-side `WrittenEgressRule` / `EgressRuleWriteOutcome` duplication is the correct narrow-port pattern, not drift — `RuleVerdict` is exactly `"allow" | "deny"`, so the writer's richer row stays structurally assignable to the port. The `updatePromoteToManual → updateTakeOwnership` rename is complete across repository, service, adapter, and port test.

Verdict: **aligned.** `docs/architecture/logging.md` correctly documents the new `approval.verdict_conflict` event and bumps `Last verified:` to the PR date. No ADR-reference or volatile-content leaks; no new subsystem needing a page.

✅ No findings. The discriminated-union outcome is exhaustively handled at all three call sites; the consumer-defined narrow port with a structurally-compatible outcome type follows the ports-and-adapters rule; throwing `TRPCError` from the service layer matches the existing `egress-rules-service` convention.

### Verdict
**APPROVE** — Correct, well-scoped fix that closes a genuine integrity gap and mirrors the established manual-create semantics; the missing test coverage is a 🟡 worth addressing but doesn't block.

---

kapetr (approved):

Issue #2969, opened 2026-07-24
Network gateway restart warnings don't match when the gateway actually restarts


PetrBulanek:
## Context

Changing a sandbox's network access sometimes restarts its network gateway (~5–15s): the sandbox keeps running, but its outbound requests are briefly interrupted, so we warn first and let the user own the timing.

Only rules that narrow a host by path, method, or non-standard port cause a restart — the gateway must restart to begin inspecting requests. Plain host rules don't. Each surface guesses at this separately, and the guesses disagree with each other and with reality. #2903 fixed the worst case (port-only rules restarted unannounced) in place, without unifying the answer.

Every case below was reproduced on a local cluster by watching whether the gateway pod actually restarted.

## Problem

Users get warned about interruptions that never happen, and get silent interruptions they were never warned about. Two surfaces can say opposite things about the same rule, and the dialog fires often enough on no-ops that users learn to click through it — which is what makes the real interruptions land unannounced.

## Expected Behavior

Users are warned when — and only when — the gateway will actually restart, with the same answer from every surface: the rules list, the settings Save dialog, and the approvals inbox.

Open question: should a restart caused by *removing* access confirm like one caused by adding it? Today it gets nothing.

## Actual Behavior

Warns when nothing restarts:

- Second narrowing rule on an already-narrowed host — the rules list stays quiet, saving warns anyway. The two surfaces contradict each other.
- Any narrowing rule on the catch-all `*` host — "every host" can't be inspected, so nothing restarts. Both surfaces warn.
- Narrowing a host already inspected because of a port-only rule.

Doesn't warn when it does restart:

- Revoking the last narrowing rule on a host — the gateway stops inspecting it and restarts. No confirmation, revoked immediately or staged and saved.
- Approving a held request permanently from the inbox — restarts, inbox says nothing. The adjacent "Allow host" doesn't restart, and nothing marks the difference.
- Narrowing a host that only one of the sandbox's connections narrows (e.g. a Google connection covering `www.googleapis.com/drive/*`). A connection's own rules don't cause inspection, so this restarts — but the rules list reads the host as already covered and stays quiet.

## Steps to Reproduce

False alarm:

1. Sandbox → Network access. Add `api.example.com` path `/v1/*`, save, accept the restart. Gateway restarts — correct.
2. Add `api.example.com` path `/v2/*`, save.
3. "Restart network gateway?" appears again. Nothing restarts.

Silent restart on removal:

[...]

PR #2970, opened 2026-07-24
fix(api-server,ui): stop presenting gateway-only rolls as agent restarts

### LINKED ISSUE #2903: UI flashes "restarting" on gateway-only roll (permanent egress approval)
## Problem

Permanently approving an egress request rolls the **gateway** pod (to pick up the new Envoy chain set / leaf cert) but does **not** restart the agent. During that gateway roll, the UI briefly shows the agent as **"restarting"**, which is misleading — the agent itself never goes down.

Surfaced while testing the per-agent L7 promotion fix (#2865 / #2870): now that a narrow egress rule only rolls the one agent's gateway, this gateway-only roll is the common case, and the "restarting" flash on it is confusing.

## Expected

The gateway-only roll should not present as an agent restart. Either:
- distinguish a gateway roll from an agent restart in the status, or
- don't surface the transient gateway roll as a top-level "restarting" state at all.

## Notes
- Repro: approve some egress request permanently → gateway rolls (agent does not) → UI blips "restarting".
- Context: #2870 (per-agent L7 promotion), #2865.

PetrBulanek:
Permanently approving a narrowing egress rule (L7 promotion) rolls only the sandbox's network gateway — the agent itself never goes down. The UI treated any not-ready blip as an agent restart, flashing the "Starting" takeover overlay over a working chat.

Now a gateway-only roll keeps the sandbox **Running**: no overlay, no status-pill flash, sessions visibly uninterrupted. Real restarts, wakes, and cold starts still present as "Starting" exactly as before.

The egress-save flow's wording matches the new reality: the confirm dialog, inline warning, and confirm prompt say the **network gateway** restarts and the agent keeps running. Port-only rules — which promote to L7 like path/method rules — now get the same confirm dialog instead of restarting the gateway silently.

Notes:
- If the agent pod becomes ready before its gateway (rare — a wake or the tail of a manual restart), "Running" can show a few seconds before full readiness; interaction still waits on full readiness.
- A persistently broken gateway now reads "Running" with egress down (previously a perpetual "Starting" — both wrong); the gateway error still surfaces on interaction. Accepted trade-off.
- Dialog-consistency follow-ups (over-warn edges, demotion rolls) are tracked in #2969.

Closes #2903

### DIFF
--- a/packages/ui/src/modules/egress-rules/components/agent-egress-editor.tsx
+++ b/packages/ui/src/modules/egress-rules/components/agent-egress-editor.tsx
@@ -119,14 +119,14 @@ export function AgentEgressEditor({
   const stagedMode = staged !== undefined;
 
   // Path-specific and port-carrying rules need MITM, which means the
-  // controller has to re-issue the leaf cert and roll the agent pod. The
-  // L4 (host-only, 443) path is a pure DB write — no roll. Warn the user
-  // so they own the timing.
+  // controller has to re-issue the leaf cert and roll the gateway pod
+  // (the agent pod stays up, #2903). The L4 (host-only, 443) path is a
+  // pure DB write — no roll. Warn the user so they own the timing.
   const draftNeedsMitm =
     draft.method !== "*" ||
     draft.pathPattern.trim() !== "*" ||
     splitHostPort(draft.host.trim()).port != null;
-  const draftRequiresRestart =
+  const draftRequiresGatewayRestart =
     draft.host.trim().length > 0 &&
     draftNeedsMitm &&
     !serverRules.some(
@@ -157,9 +157,9 @@ export function AgentEgressEditor({
       return;
     }
     if (
-      draftRequiresRestart &&
+      draftRequiresGatewayRestart &&
       !window.confirm(
-        `Saving this rule will restart the agent (~5–15s) so Envoy can MITM "${next.host}" for path-level enforcement. Continue?`,
+        `This rule needs a gateway restart (~5–15s). The agent keeps running — outbound requests are briefly interrupted. Continue?`,
       )
     )
       return;
@@ -350,11 +350,11 @@ export function AgentEgressEditor({
           >
             <Plus size={11} /> Add rule
           </Button>
-          {draftRequiresRestart && (
+          {draftRequiresGatewayRestart && (
             <p className="basis-full text-[11px] text-warning">
-              {stagedMode
-                ? "Saving will restart the agent (~5–15s) — path-level rules need MITM on this host."
-                : "Saving will restart the agent (~5–15s) — path-level rules need MITM on this host."}
+              Saving will restart the network gateway (~5–15s) — this rule
+              requires inspecting requests to this host. The agent keeps
+              running.
             </p>
           )}
         </div>
--- a/docs/architecture/platform-topology.md
+++ b/docs/architecture/platform-topology.md
@@ -1,6 +1,6 @@
 # Platform topology
 
-Last verified: 2026-07-17
+Last verified: 2026-07-24
 
 ## Overview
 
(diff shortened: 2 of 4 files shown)


packages/ui/src/modules/egress-rules/components/agent-egress-editor.tsx
```diff
@@ -119,14 +119,14 @@ export function AgentEgressEditor({
...
-  // so they own the timing.
+  // controller has to re-issue the leaf cert and roll the gateway pod
+  // (the agent pod stays up, #2903). The L4 (host-only, 443) path is a
+  // pure DB write — no roll. Warn the user so they own the timing.
   const draftNeedsMitm =
     draft.method !== "*" ||
     draft.pathPattern.trim() !== "*" ||
     splitHostPort(draft.host.trim()).port != null;
-  const draftRequiresRestart =
+  const draftRequiresGatewayRestart =
```
dam-code-guardian, line 129: 🟢 **Suggestion:** `draftNeedsMitm` now treats a port as needing MITM, but the already-L7 exemption below (`serverRules.some(... r.method !== "*" || r.pathPattern !== "*" ...)`) still only inspects method/path. A host that already has a port-only rule will re-trigger this warning when a second port rule is added. Minor over-warn, and per the PR body this edge is already tracked in #2969 — noting for completeness.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `ab42b4b`

## PR #2970: fix(api-server,ui): stop presenting gateway-only rolls as agent restarts
**Author:** PetrBulanek | **Branch:** fix/2903-gateway-roll-not-restarting → main | **Changes:** +36 −21 (4 files)

### Summary
When an L7-promoting egress rule change rolls only the sandbox's network gateway (not the agent pod), the UI no longer flashes the "Starting" takeover overlay: `computeAgentState` now projects `agentPodReady && !gatewayPodReady` as `running`, and the egress-save copy is reworded to say the gateway restarts while the agent keeps running. Routing (`ensureReady`) still waits on the aggregate `Ready` condition.

### Findings
- ✅ **Looks good:** The new `computeAgentState` branch is presentation-only and correctly placed after the `error`/`ready`/`hibernated`/`overBudget` guards, so those states still win; the `preparingWorkspace` refinement mirrors the `ready` branch.
- ✅ **Looks good:** `agentPodReady` is parsed with the same tri-state (`True`/`False`/undefined) convention as the existing `gatewayPodReady`, keeping the mapper consistent.
- ✅ **Looks good:** The Save-time filter in `use-sandbox-settings-save.ts` now mirrors server-side `needsL7Promotion` (method / path / non-443 port) via `splitHostPort`, so port-only rules get the confirm dialog instead of a silent gateway roll.
- 🟢 **Suggestion:** `draftNeedsMitm` now counts a port as needing MITM, but the already-L7 exemption (`serverRules.some(... method/path ...)`) still only inspects method/path — so adding a second port-only rule to a host that already has one can re-warn. Minor over-warn; already tracked in #2969. (`packages/ui/src/modules/egress-rules/components/agent-egress-editor.tsx:129`)

✅ Aligned. `docs/architecture/platform-topology.md` was updated in the same PR to document that the user-facing state projection reads pod-level conditions so a gateway-only roll isn't presented as an agent restart, and its `Last verified:` header is bumped to the PR date (2026-07-24). No ADR references or volatile content added.

✅ No findings. `agent-mappers.ts` keeps its parse mapper and `computeAgentState` domain projection pure and unidirectional (new field follows the existing `gatewayPodReady` pattern); the UI save hook composes mutation hooks without crossing layer boundaries.

✅ No new findings. The `.tsx` change is a rename (`draftRequiresRestart` → `draftRequiresGatewayRestart`) plus copy updates — no new state, Tailwind classes throughout, `window.confirm` matches the file's existing pattern. (The file is 644 lines, pre-existing and untouched by this diff.)

### Verdict
APPROVE — a tight, well-scoped presentation fix with matching UI copy, server-parity port handling, and the architecture doc updated in the same PR; the only nit (port over-warn) is minor and already tracked.

---

kapetr (approved):

PR #2974, opened 2026-07-27
feat(channels): let a Slack agent resolve user ids to people

pilartomas:
## Why

A Slack agent only ever met people as bare `U…` ids: the speaker label on a shared-channel turn, the author prefix on an injected history line, a mention inside message text. It could read the id and nothing else — no name to address someone by, no title, no time zone, no way to tell a bot from a person.

## What

A new per-Agent MCP tool, `describe_channel_users`, resolves a batch of ids to the people behind them.

- **Port → gateway → worker → tool.** `SlackGateway.getUserInfo` (Bolt `users.info`) returns handle, real/display name, title, pronouns, email, time zone and current status; absent fields mean "unset or withheld". The fake gateway gains a seedable directory plus a lookup recorder, so the worker's behaviour is unit-tested the way the outbound tools are.
- **Batch semantics.** Up to 20 ids per call. Each id resolves independently — an unknown id, a malformed one, or a Slack-side failure comes back as that entry's `error` and never costs the caller the rest of the batch. `<@U024BE7LH|ada>` spellings are normalized, repeats collapse to one lookup, and a non-id (a handle, say) is rejected without a round trip.
- **Caching.** Hits *and* misses are cached for ten minutes behind a concurrency cap of five: the directory is workspace-wide, the bot is install-wide, and a channel's regulars get asked about turn after turn.
- **Discoverability.** The per-turn contract and the injected-history legend both name the tool, so an agent that reads an id knows it is resolvable. (The contract line deliberately avoids the literal `<@` sequence — a 1:1 DM prompt asserts it contains none, which is its proxy for "not speaker-labelled".)
- **Gating and audit.** The binding gates the lookup like every other outbound affordance, so an unbound Agent reads no directory. Each call is security-logged by the ids asked about and how many resolved — never the profiles that came back. Slack-only; Telegram has no workspace directory and refuses with a clear message.

## Follow-up: `users:read` is optional, and the tool now behaves like it

`users:read` and `users:read.email` are both optional Slack scopes — an install can run without either. Losing `users:read.email` already degraded gracefully (the email field is just absent). Losing `users:read` did not: every `describe_channel_users` call would fail forever, on every channel, for every Agent, and the tool stayed registered and kept telling the agent to call it anyway.

- **Scope detection.** `SlackGateway.getGrantedScopes()` asks Slack directly via a cached `auth.test()` call — Slack reports the token's granted scopes in the `x-oauth-scopes` response header on every Web API call, so this needs no extra permission and costs one call per process lifetime (a real scope change means reinstalling the app, which restarts the process anyway).
- **Conditional registration.** `describe_channel_users` is now dropped from the per-Agent MCP tool list — and its mention from the per-turn contract and the injected-history legend — whenever `users:read` is confirmed missing. An unreachable bot or an unprobed gateway fails open (keeps the tool), so this only fires on a positive, install-wide confirmation, never a transient blip.
- **Docs.** `docs/architecture/channels.md` gains a "Slack scopes: required vs. optional" section naming this as a general pattern (two degrade strategies — narrow the result vs. omit the capability — depending on whether a partial answer means anything), so future Slack-facing scopes have a place to point at instead of re-deriving the split.

## Docs

`docs/architecture/channels.md` gains the behaviour under Outbound (`Last verified` bumped) plus the new scopes section above. `docs/configuration.md` picked up the missing `users:read.email` and now points operators at the manifest as the complete scope set — its inline list had drifted — plus a note on what each optional scope's absence actually does.

## Verification

- `mise run api-server:test` — 632 passed, including the original 9 worker-level cases and 2 channel-manager dispatch cases, plus new coverage for `supportsUserLookup` (worker-level scope checks, channel-manager fan-in) and the contract/legend gating.
- `mise run api-server:check` (tsc + lint + format) and `mise run docs:check` pass.
- `mise run check` fails only on `controller:check:crd-backwards-compatibility`, which cannot resolve refs through a worktree gitdir; it passes from the primary checkout, and this branch touches no CRDs.
- No live-cluster run.


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `7eaa4a2`

## PR #2974: feat(channels): let a Slack agent resolve user ids to people
**Author:** pilartomas | **Branch:** feat/slack-user-info-tool → main | **Changes:** +531 −4 (12 files)

### Summary
Adds a per-Agent MCP tool `describe_channel_users` that batch-resolves Slack `U…` ids to profiles (handle, name, title, pronouns, email, tz, status). Threaded cleanly through the ports-and-adapters stack (tool → channel-manager → slack worker → gateway port + bolt/fake adapters), with per-id error isolation, `<@U…>` normalization, dedupe, a 10-minute hit/miss cache, and a concurrency-capped, binding-gated, audit-logged lookup.

### Findings
- ✅ **Looks good:** Per-id error isolation is correct — an unknown/malformed id or a Slack throw fails only its own entry (`slack.ts:2136`), and `getUserInfo` maps `user_not_found` to a `null` miss rather than a thrown fault (`bolt-slack-gateway.ts:423`). Well covered by the 9 new worker-level tests.
- ✅ **Looks good:** Semaphore acquire/release is leak-safe — `release()` sits in a `finally` that runs on the throw path (`slack.ts:2134`).
- ✅ **Looks good:** Security posture — the lookup is gated by the binding like every other outbound affordance (`slack.ts:2661`), and the audit log records the *ids asked about* + a resolved-count, never the returned profiles/emails (`mcp-endpoint.ts:327`). Email in the payload is a deliberate, scope-backed choice the author explicitly flagged; both scopes were pre-granted and the manifest now records why.
- ✅ **Looks good:** `describeUsers?` on `Worker` follows the existing optional-capability pattern (`reply?`/`react?`); the manager dispatches by channel type with a clear `user lookup not supported on telegram` (`channel-manager.ts:240`).
- 🟢 **Suggestion:** Cross-batch cache race — two concurrent batches asking about the same uncached id both miss and issue duplicate `users.info` calls (in-batch `seen` dedup doesn't cover concurrent invocations). Harmless given the best-effort cache + bounding semaphore; noted for completeness only.

**Verdict: minor drift.** `docs/architecture/channels.md` is correctly updated (new `describe_channel_users` bullet under Outbound, cross-link from the attribution note, `Last verified` bumped to 2026-07-27). `docs/configuration.md` scope list reconciled with the manifest.

- 🟢 **Minor / possible drift:** The PR introduces a new security-audit event `channel.user_lookup` (`mcp-endpoint.ts:319`), but `docs/architecture/logging.md` — whose Channels row enumerates the `channel.*` events — doesn't list it, and its `Last verified: 2026-07-21` is unbumped. The table is headed "Representative events" (not exhaustive), so this is a suggestion rather than a blocker; consider adding `channel.user_lookup` (ids-only detail) to that row.

✅ No findings. Ports & adapters respected: `getUserInfo` on the `SlackGateway` port, implemented by both bolt and fake adapters; the domain-facing `ChannelUser` shape maps from the infra `SlackUserInfo`. DI/optional-capability and error-as-value conventions match the surrounding code.

### Verdict
**APPROVE** — clean, well-tested, security-conscious feature that respects the architecture; the only note is an optional `logging.md` event-list update.

---

packages/api-server/src/modules/channels/infrastructure/bolt-slack-gateway.ts
```diff
@@ -297,12 +302,60 @@ export function createBoltSlackGateway(
...
+
+    async getGrantedScopes(): Promise<Set<string> | null> {
+      if (!app) return null;
+      if (grantedScopes) return grantedScopes;
+      try {
+        // auth.test needs no scope of its own and has a generous rate limit;
+        // it's a cheap way to ask Slack what the bot token can actually do.
+        const result = await app.client.auth.test();
+        const scopes = result.response_metadata?.scopes;
+        if (scopes) grantedScopes = new Set(scopes);
```
dam-code-guardian, line 354: 🟢 **Suggestion:** A `result` with no `scopes` header leaves `grantedScopes` null, so this probe re-runs on every call until one returns the header — a `users.info`-adjacent call per lookup batch until then. Harmless (bounded to the missing-header case), noted for completeness only.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `52e2220`

### Summary
Follow-up making `users:read` a genuinely optional scope: the app probes Slack's granted scopes once per process (`auth.test` → `x-oauth-scopes`) and, on a confirmed miss, drops `describe_channel_users` from the tool list and its mention from the per-turn contract and history legend — rather than registering a tool that would always fail. Unknown/unreachable states fail open. Threaded cleanly through port → gateway → worker → channel-manager → MCP session.

### Changes since last review
Previous HEAD: 7eaa4a2 (2026-07-27T09:30:04Z) — verdict APPROVE

- ✅ **Fixed:** The tool no longer stays registered when it can never work — conditional registration via `supportsUserLookup` resolved at session mount (`mcp-endpoint.ts`).
- 🔁 **Still present:** New `channel.user_lookup` security-audit event not listed in `docs/architecture/logging.md`'s Channels row (representative-only table, so non-blocking).

### Findings
- 🟢 **Suggestion:** `getGrantedScopes` caches only a successful probe; a `result` with no `scopes` header leaves `grantedScopes` null, so every call re-probes until one succeeds (`bolt-slack-gateway.ts:354`). Harmless (bounded to the missing-header case) — noted for completeness only.

**Verdict: aligned.** `docs/architecture/channels.md` gains the "Slack scopes: required vs. optional" section documenting the two degrade strategies as a general pattern, the "one exception" to always-registered tools, `Last verified` at 2026-07-27; no ADR leaks; scope/file references are durable-level link-outs. `docs/configuration.md` reconciled.

- 🔁 1 finding from the previous review still present: `channel.user_lookup` absent from `logging.md`'s Channels row (representative table — suggestion, not blocker).

✅ No findings. Ports & adapters respected — `getGrantedScopes` added to the `SlackGateway` port and implemented by both bolt and fake adapters; the capability probe threads inward through worker → channel-manager and is resolved to a plain `boolean` at the composition root; null/unknown fails open at every layer. `supportsUserLookup` follows the existing optional-capability worker pattern (`reply?`/`react?`).

### Verdict
**APPROVE** — the optionality fix is correct and well-tested (scope-gating of tool, contract, and legend all covered); fails open on unknown scopes throughout. Only a cosmetic caching nit and the carried-over representative-table doc note remain.

---

PR #2980, opened 2026-07-27
fix(ui): browser back/forward works in every view

### LINKED ISSUE #1219: refactor(ui): dedupe onPopState path→view logic into pathToState
## Problem

The UI maintains the "URL path → which view to show" mapping in **two** places that must be kept in sync by hand:

- `pathToState(path)` in `packages/ui/src/modules/platform/lib/routes.ts` — the canonical path→state mapper.
- The `onPopState` (browser back/forward) handler in `packages/ui/src/app.tsx`, which re-implements the same if/else + regex chain inline.

Every route is therefore listed twice. This surfaced while deleting the `/v2` prototype (#1194): the v2 branches had to be removed from both functions. Adding or changing a route requires editing both, and they can silently drift.

## Proposed solution

Collapse `onPopState` so it delegates to `pathToState` instead of re-deriving the view inline.

This is **not** a trivial 1:1 substitution — `onPopState` carries side effects that `pathToState` does not model:

- **Chat:** `onPopState` calls `enterChat()` → `resetChatContext()` + sets `selectedAgent` (not `agentId`). `pathToState` returns `{ view: "chat", agent }`.
- **Fallback:** `onPopState` calls `leaveChat()` → `resetChatContext()` + `selectedAgent: null`. `pathToState` returns a bare `{ view: "list" }`.

So the fix needs a side-effect-aware wrapper that maps `pathToState`'s result into the correct `setState` plus the chat-context resets. Because this touches browser history handling, take care with the OAuth return-view restore path and the `selectedAgent` vs `agentId` distinction, and verify back/forward navigation across all views still works.

## Scope

UI routing internals only (`app.tsx`, `routes.ts`). No user-visible behavior change.

## Context

Split out of #1194 (delete the `/v2` prototype), which intentionally left this duplication in place to keep that change deletion-only.

PetrBulanek:
Pressing the browser's Back button from the Terms of Use page used to strand you: the address bar went back to Settings, but the Terms page stayed on screen, and the only way out was reloading. The same gap applied to the Telegram and Slack bind pages.

Back and Forward now work everywhere in the app. The Terms page's own Back button also returns you to the app instantly instead of triggering a full page reload.

Alongside the fix, the URL-to-screen mapping is now defined in one place instead of being restated in several, which removes a class of bug where a page could quietly start pointing at the wrong address.

No other user-visible changes.

**Note for reviewers:** returning into a chat via Back now also resets the other navigation fields to their defaults. This is deliberate, not an oversight — those fields only matter while their own screen is showing.

**Rebased onto current main.** The branch was 39 commits behind, and the three PRs that landed in between (#2948, #2971, #3014) rewrote the exact routing table this refactor touches, so the resolution was not mechanical. What changed versus the previously reviewed version:

- The route union now matches main's routes: the `experiment-new` / `experiment-detail` entries are gone (#3014 folded experiment creation into the shared sandbox wizard), and the three knowledge-base routes are in, along with the legacy `/knowledge-bases/new` → wizard redirect.
- A knowledge base's page is the same chat surface under its own route, so it now shares chat's Back handling — Back into a knowledge base restores it properly, and leaving one tears its chat context down. This behavior is new since the last review and is the part worth a second look.
- `knowledge-base-config-view.tsx` picked up the same route-literal treatment as the other call sites; it was written after this branch forked.
- The round-trip test table swapped the dead experiment paths for the knowledge-base ones, and the second ordering trap now pins `/knowledge-bases/new`.

Closes #1219

### DIFF
--- a/packages/ui/src/modules/platform/hooks/use-browser-history.ts
+++ b/packages/ui/src/modules/platform/hooks/use-browser-history.ts
@@ -1,0 +1,50 @@
+import { useEffect } from "react";
+
+import { useStore } from "../../../store.js";
+import {
+  parseRoute,
+  routeToNavigationState,
+  type View,
+} from "../lib/routes.js";
+
+/** A knowledge base's page is the same ChatView under its own route, so both
+ *  views carry their agent in `selectedAgent` and share chat's teardown. */
+const isChatView = (view: View) =>
+  view === "chat" || view === "knowledge-base-chat";
+
+/**
+ * The UI's only popstate listener. Entering either chat surface always resets
+ * chat context and sets `selectedAgent`; leaving one for any other view resets
+ * and clears it; non-chat → non-chat transitions touch no chat state.
+ */
+export function useBrowserHistory(): void {
+  useEffect(() => {
+    const applyRoute = () => {
+      const route = parseRoute(window.location.pathname);
+      const wasChat = isChatView(useStore.getState().view);
+
+      if (route.view === "chat" || route.view === "knowledge-base-chat") {
+        useStore.getState().resetChatContext();
+        useStore.setState({
+          ...routeToNavigationState(route),
+          selectedAgent: route.agent,
+        });
+        return;
+      }
+
+      if (wasChat) useStore.getState().resetChatContext();
+      useStore.setState({
+        ...routeToNavigationState(route),
+        ...(wasChat ? { selectedAgent: null } : {}),
+      });
+    };
+
+    // Not redundant next to navigation.ts's initializer: this mount-time call
+    // is the only thing that sets `selectedAgent` on a cold load at
+    // /chat/:agent (the agents slice initializes it to null), and it re-parses
+    // the URL after auth.ts restores the post-login return path.
+    applyRoute();
+    window.addEventListener("popstate", applyRoute);
+    return () => window.removeEventListener("popstate", applyRoute);
+  }, []);
+}
--- a/packages/ui/src/modules/platform/hooks/use-browser-history.ts
+++ b/packages/ui/src/modules/platform/hooks/use-browser-history.ts
@@ -3,6 +3,7 @@ import { useEffect } from "react";
 import { useStore } from "../../../store.js";
 import {
   parseRoute,
+  type Route,
   routeToNavigationState,
   type View,
 } from "../lib/routes.js";
@@ -11,6 +12,11 @@ import {
  *  views carry their agent in `selectedAgent` and share chat's teardown. */
 const isChatView = (view: View) =>
   view === "chat" || view === "knowledge-base-chat";
+
+const isChatRoute = (
+  route: Route,
+): route is Extract<Route, { view: "chat" | "knowledge-base-chat" }> =>
+  isChatView(route.view);
 
 /**
  * The UI's only popstate listener. Entering either chat surface always resets
@@ -23,7 +29,7 @@ export function useBrowserHistory(): void {
       const route = parseRoute(window.location.pathname);
       const wasChat = isChatView(useStore.getState().view);
 
-      if (route.view === "chat" || route.view === "knowledge-base-chat") {
+      if (isChatRoute(route)) {
         useStore.getState().resetChatContext();
         useStore.setState({
           ...routeToNavigationState(route),
--- a/packages/ui/src/__tests__/unit/routes.test.ts
+++ b/packages/ui/src/__tests__/unit/routes.test.ts
@@ -1,0 +1,41 @@
+import { describe, expect, it } from "vitest";
+
+import { parseRoute, routeToPath } from "../../modules/platform/lib/routes.js";
+
+// One canonical path per view. Non-canonical spellings (/settings/account,
+// /sandboxes/:id/setup) intentionally don't round-trip — routeToPath emits the
+// canonical form.
+const canonicalPaths = [
+  "/",
+  "/chat/agent-1",
+  "/settings",
+  "/settings/connections",
+  "/inbox",
+  "/terms",
+  "/telegram/bind",
+  "/slack/bind",
+  "/sandboxes/new",
+  "/sandboxes/sb-1",
+  "/sandboxes/sb-1/connections",
+  "/experiments",
+  "/knowledge-bases",
+  "/knowledge-bases/kb-1",
+  "/knowledge-bases/kb-1/settings",
+  "/artifacts",
+];
+
+describe("route round-trip", () => {
+  it.each(canonicalPaths)("routeToPath(parseRoute(%s)) is identity", (path) => {
+    expect(routeToPath(parseRoute(path))).toBe(path);
+  });
+
+  // Both traps fail silently if the literal check moves below its regex:
+  // the wizard becomes a detail page for an entity named "new".
+  it("parses /sandboxes/new as the wizard, not a sandbox id", () => {
+    expect(parseRoute("/sandboxes/new").view).toBe("sandbox-new");
+  });
+
+  it("parses the legacy /knowledge-bases/new as the wizard, not a KB id", () => {
+    expect(parseRoute("/knowledge-bases/new").view).toBe("sandbox-new");
+  });
+});
--- a/packages/ui/src/__tests__/unit/slack-bind-flow.test.ts
+++ b/packages/ui/src/__tests__/unit/slack-bind-flow.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from "vitest";
 
-import { pathToState, viewToPath } from "../../modules/platform/lib/routes.js";
+import { parseRoute, routeToPath } from "../../modules/platform/lib/routes.js";
 import {
   bindErrorCopy,
   callbackErrorCopy,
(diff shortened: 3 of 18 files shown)


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `2d836a6`

## PR #2980: fix(ui): browser back/forward works in every view
**Author:** PetrBulanek | **Branch:** refactor/1219-ui-routing-dedupe → main | **Changes:** +277 −173 (17 files)

### Summary
Consolidates the UI's URL↔view mapping into a single `Route` discriminated union in `routes.ts` (`parseRoute` / `routeToPath` / `routeToNavigationState`), replacing the scattered `pathToState`/`viewToPath` pair, and moves browser back/forward `popstate` handling into one `useBrowserHistory` hook mounted above the terms/bind early-returns so back/forward works in every view. Fixes the Terms page strand-on-Back bug (#1219).

### Findings
- ✅ **Looks good:** `Route` is a proper discriminated union and `routeToPath` is exhaustiveness-checked via `const unhandled: never = route` — adding a view without a path mapping becomes a compile error.
- ✅ **Looks good:** The two ordering traps (`/sandboxes/new` and `/experiments/new` must precede their `:id` regexes) are called out with comments *and* guarded by dedicated round-trip tests in the new `routes.test.ts`.
- ✅ **Looks good:** URL segments are validated with `safeParse` (settings tab, sandbox section) rather than cast, so a bogus `/settings/xxx` degrades to `account` instead of an invalid state; the sandbox-home regex is derived from the schema so it can't drift.
- ✅ **Looks good:** A single consolidated `popstate` listener replaces the inline effect; enter/leave-chat context-reset semantics are preserved and the mount-time `applyRoute()` correctly rehydrates `selectedAgent` on a cold `/chat/:agent` load. Old exports fully removed — no dangling references.
- ✅ **Looks good:** `setView` is narrowed to a `ParameterlessView` union, so parameterized views can only be reached through their dedicated typed navigators. All 8 `setView` callers comply.

**Verdict:** aligned

No `docs/` files are touched and no architecture page documents the UI's client-side URL↔view mapping (the `ui` appears only as a topology component). No behavior change to a documented subsystem, no new subsystem, and no architecture page was edited, so no `Last verified:` obligation arises.

✅ No findings. `routes.ts` models the route space as a discriminated union with an exhaustive `switch` (`never` guard) and Zod-backed validation of untrusted URL segments — the correct type-safety posture. The TSEng service/port/domain rules don't apply to browser-side routing helpers.

✅ No findings. Route is correctly URL-owned (state-lineage model), `popstate` handling is a single focused hook, and there are no static inline styles. The one new inline `useStore((s) => s.setView)` selector in `terms-view.tsx` matches the codebase-wide convention (54 files); promoting to exported selector hooks would be a separate batch refactor, out of scope here.

### Verdict
APPROVE — A clean, well-tested routing consolidation that fixes the back/forward strand bug; single source of truth for URL↔view, exhaustive typing, and validation on untrusted input.

---

packages/ui/src/modules/platform/hooks/use-browser-history.ts
```diff
@@ -0,0 +1,50 @@
...
+ * chat context and sets `selectedAgent`; leaving one for any other view resets
+ * and clears it; non-chat → non-chat transitions touch no chat state.
+ */
+export function useBrowserHistory(): void {
+  useEffect(() => {
+    const applyRoute = () => {
+      const route = parseRoute(window.location.pathname);
+      const wasChat = isChatView(useStore.getState().view);
+
+      if (route.view === "chat" || route.view === "knowledge-base-chat") {
```
kapetr: `isChatView` function from this file can be used

kapetr (approved):

packages/ui/src/modules/platform/hooks/use-browser-history.ts
dam-code-guardian: 🟢 **Suggestion:** `isChatView` is already defined above — reuse it here so both chat-surface checks share one definition and can't drift.

```suggestion
      if (isChatView(route.view)) {
```

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `cf275c8`

## PR #2980: fix(ui): browser back/forward works in every view
**Author:** PetrBulanek | **Branch:** refactor/1219-ui-routing-dedupe → main | **Changes:** +300 −178 (18 files)

### Summary
Re-review after a rebase onto current main. The routing consolidation is unchanged in shape (`Route` discriminated union + `parseRoute`/`routeToPath`/`routeToNavigationState`, one `useBrowserHistory` popstate hook); the rebase swapped the removed experiment-creation routes for the knowledge-base routes (incl. the legacy `/knowledge-bases/new` → wizard redirect), so a KB's standalone page now shares chat's Back/teardown semantics.

### Changes since last review
Previous HEAD: `2d836a6` (2026-07-27T09:43:49Z) — verdict APPROVE

- ✅ **Fixed:** Route union re-aligned to main — experiment-new/detail gone, three KB routes + `/knowledge-bases/new` redirect land in the wizard, all exercised by the round-trip table (`routes.test.ts:14`).
- 🆕 **New:** `isChatView(view)` is defined (`use-browser-history.ts:12`) but the same-view branch re-inlines `route.view === "chat" || route.view === "knowledge-base-chat"` instead of reusing it (`use-browser-history.ts:26`) — matches the human reviewer's open note; using the helper keeps the two chat-surface checks from drifting.

### Findings
- 🟢 **Suggestion:** Reuse `isChatView(route.view)` for the entry branch at `use-browser-history.ts:26` so both chat-surface checks share one definition (`packages/ui/src/modules/platform/hooks/use-browser-history.ts:26`).

**Verdict:** aligned

No `docs/` files touched. No architecture page documents the UI's client-side URL↔view mapping; the KB standalone page's navigational contract in `docs/architecture/knowledge-bases.md` (rail keeps KB context, leaving returns to the KB list) is preserved by this change — the chat-context teardown is an undocumented internal detail. No new subsystem, no `Last verified:` obligation.

✅ No new findings. The rebase preserves the type-safety posture reviewed before — discriminated `Route` union, exhaustive `switch` with `const unhandled: never` guard, and Zod `safeParse` on untrusted URL segments (settings tab, sandbox section). Service/port/domain rules don't apply to browser-side routing helpers.

✅ No new findings. Route stays URL-owned (state-lineage model), the popstate handler remains a single focused hook, no static inline styles introduced. The lone inline `useStore((s) => s.setView)` in `terms-view.tsx` is a carryover already noted as matching the codebase-wide convention (out of scope).

### Verdict
APPROVE — Clean rebase of an already-approved routing consolidation; the KB routes fold in correctly and stay tested. One 🟢 DRY nit (shared `isChatView` helper), matching the human reviewer's note — non-blocking.
---

packages/ui/src/modules/platform/hooks/use-browser-history.ts
PetrBulanek: Good catch on the duplication — fixed in 08293cdf.

One wrinkle: `isChatView` can't be dropped in here directly. It returns `boolean` rather than a type predicate, so `route` stays the full `Route` union and `route.agent` on the next line fails:

```
error TS2339: Property 'agent' does not exist on type 'Route'.
  Property 'agent' does not exist on type '{ view: "list"; }'.
```

Annotating `isChatView` as `view is "chat" | "knowledge-base-chat"` doesn't help either — that narrows only the expression you pass it, and TypeScript doesn't propagate a predicate on `route.view` back up to `route`. Parent-from-property narrowing fires for a direct literal comparison, `typeof`, and `in`, but not for a user-defined predicate call.

So the guard has to be on the route object, delegating to `isChatView` so the pair is still defined in one place:

```ts
const isChatRoute = (
  route: Route,
): route is Extract<Route, { view: "chat" | "knowledge-base-chat" }> =>
  isChatView(route.view);
```

Line 26 is now `if (isChatRoute(route)) {`. `Extract<...>` predicates are already used this way in `use-inherited-envs.ts:33` and `provider-item.ts:24`. tsc, eslint, prettier and the 153 UI unit tests all pass locally.

PetrBulanek: ⚠️ The committable suggestion here does not compile — please don't apply it as-is.

`isChatView` returns `boolean`, not a type predicate, so `route` is not narrowed and `route.agent` on the next line fails:

```
error TS2339: Property 'agent' does not exist on type 'Route'.
  Property 'agent' does not exist on type '{ view: "list"; }'.
```

Annotating `isChatView` as `view is "chat" | "knowledge-base-chat"` doesn't fix it either: narrowing `route.view` never narrows `route`. TypeScript only does parent-from-property narrowing for direct literal comparisons, `typeof`, and `in` — not for a user-defined predicate call.

The DRY point is valid though, so 08293cdf keeps one definition by putting the predicate on the route object and delegating to `isChatView`:

```ts
const isChatRoute = (
  route: Route,
): route is Extract<Route, { view: "chat" | "knowledge-base-chat" }> =>
  isChatView(route.view);
```

Worth noting for future suggestions of this shape: a type predicate is an unchecked assertion — TS accepts `(view: View): view is ChatView => view === "chat"` with the body checking only one of the two members — so the delegating form is what keeps the runtime list single-sourced.

PR #2982, opened 2026-07-27
fix(ui): OAuth result notification no longer disappears

PetrBulanek:
After authorizing a connection with an external provider, the app is supposed to tell you whether it worked. That notification never appeared — neither the success confirmation nor the failure message.

The cause was a timing problem: notifications sent while the app is still starting up had nowhere to go yet and were discarded. That affected the OAuth result specifically, because it always arrives just as the app reloads after returning from the provider.

Notifications sent during startup are now held briefly and shown once the app is ready. Verified for both places the OAuth result can appear: the sandbox creation wizard and the main app shell.

No other user-visible changes.

### DIFF
--- a/packages/ui/src/lib/toast.ts
+++ b/packages/ui/src/lib/toast.ts
@@ -26,16 +26,35 @@ const EMIT: Record<
   info: sonner.info,
 };
 
-/** Surface a toast through Sonner. The single emit path for the whole app —
- *  React components and non-React modules (query helpers) both call this. */
-export function emitToast({
-  kind,
-  message,
-  action,
-  ttl,
-}: Toast): string | number {
-  return EMIT[kind](message, {
+/** Sonner fans a toast out to whoever is subscribed at publish time and keeps
+ *  no backlog, so anything emitted before `<Toaster>` mounts is lost for good.
+ *  On a cold load that silently swallows mount-time toasts — including the
+ *  OAuth result, which only ever arrives on a cold load. */
+let hostReady = false;
+const pending: Toast[] = [];
+
+/** Called by the Toaster wrapper once Sonner is subscribed. */
+export function onToastHostMounted(): () => void {
+  hostReady = true;
+  for (const toast of pending.splice(0)) send(toast);
+  return () => {
+    hostReady = false;
+  };
+}
+
+function send({ kind, message, action, ttl }: Toast): void {
+  EMIT[kind](message, {
     action,
     duration: ttl === undefined ? DEFAULT_TTL : ttl > 0 ? ttl : Infinity,
   });
+}
+
+/** Surface a toast through Sonner. The single emit path for the whole app —
+ *  React components and non-React modules (query helpers) both call this. */
+export function emitToast(toast: Toast): void {
+  if (!hostReady) {
+    pending.push(toast);
+    return;
+  }
+  send(toast);
 }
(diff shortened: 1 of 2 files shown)


packages/ui/src/lib/toast.ts
```diff
@@ -26,16 +26,35 @@ const EMIT: Record<
...
+}
+
+function send({ kind, message, action, ttl }: Toast): void {
+  EMIT[kind](message, {
     action,
     duration: ttl === undefined ? DEFAULT_TTL : ttl > 0 ? ttl : Infinity,
   });
 }
+
+/** Surface a toast through Sonner. The single emit path for the whole app —
```
dam-code-guardian, line 52: 🟢 **Suggestion:** The buffer/flush ordering here is subtle timing logic worth locking down with a unit test — there is currently none for `toast.ts`. Covering emit-before-ready (buffers), `onToastHostMounted` (flushes in order), and emit-after-ready (sends immediately) would guard against a regression in the child-before-parent effect ordering this fix depends on.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `74adf57`

## PR #2982: fix(ui): OAuth result notification no longer disappears
**Author:** PetrBulanek | **Branch:** fix/oauth-toast-dropped-on-cold-load → main | **Changes:** +34 −9 (2 files)

### Summary
Fixes mount-time toasts (notably the OAuth result, which only ever arrives on a cold load) being silently dropped: `emitToast` now buffers into a module-level `pending` queue until the `<Toaster>` wrapper signals readiness via `onToastHostMounted`, which flushes the backlog. Off-path behaviour (toasts emitted after mount) is unchanged.

### Findings
- 🟢 **Suggestion:** The new buffer/flush ordering is exactly the kind of subtle timing logic that a unit test locks down cheaply — there is currently no test for `toast.ts`. A small test (emit-before-ready buffers, `onToastHostMounted` flushes in order, emit-after-ready sends immediately) would guard against a future regression in the child-before-parent effect ordering the fix relies on. (`packages/ui/src/lib/toast.ts:52`)
- ✅ **Looks good:** Return type of `emitToast` narrowed from `string | number` to `void` — verified no caller consumes the return value across the `packages/ui` tree, so this is a safe tightening.
- ✅ **Looks good:** The fix hinges on child effects firing before the parent's, so `<App>`'s OAuth `emitToast` buffers before `<Toaster>`'s flush runs. That invariant is load-bearing and the one-line comment correctly documents the *why*.
- ✅ **Looks good:** `pending.splice(0)` both drains and clears atomically before dispatch, so a toast emitted from within a flushed toast's `onClick`/side-effect can't be double-sent.
- ✅ **Looks good:** `send()` extraction removes duplication between the buffered-flush path and the immediate path — one emit implementation, `emitToast` only decides buffer-vs-send.

**Verdict:** aligned

No architecture page documents the UI toast-delivery path, and the diff changes no subsystem behaviour/responsibility recorded under `docs/architecture/`. No drift.

✅ No findings. The change is strictly typed, keeps a single emit path, and introduces no layer/module-boundary violations.

✅ No findings. The added `useEffect(onToastHostMounted, [])` is a single-responsibility effect with a correct cleanup and a justified comment; the pre-existing inline `useStore` selector is untouched and out of scope.

### Verdict
APPROVE — a small, correct, well-reasoned fix for a real cold-load notification-drop bug; the only note is optional test coverage for the new buffering.

---

kapetr (approved):

PR #2991, opened 2026-07-27
feat(channels): let a Slack agent look up who reacted to a message

pilartomas:
## Summary

Part of #2979 (surface rich user & message metadata into agent context). This slice covers the reactions half — GitHub-handle mapping and the remaining message-metadata/timezone-scheduling items from that issue are still open.

- New `describe_message_reactions` MCP tool: resolves a message's reactions (emoji, count, reacting user ids) via Slack's `reactions.get`. Defaults to the message the current turn is answering, in the agent's bound channel — reusing the same `resolveTurn`/`resolveOutboundTarget` machinery `reply`/`react` already use — or an explicit `chatId`/`messageTs` for a message outside the live turn (e.g. a scheduled agent checking a standing weekly thread).
- Deliberately **not cached** (unlike the 10-minute user-profile cache): a reaction tally is live, mutable state, and the whole point of the tool is the current count.
- Gated on a new optional `reactions:read` Slack scope, following the `describe_channel_users`/`users:read` precedent documented in `docs/architecture/channels.md` under "Slack scopes: required vs. optional": the tool is omitted from the MCP tool list entirely (not just left to error) when the scope is confirmed missing, since `reactions.get` fails outright without it — there's no narrower result to degrade to.
- The result (and audit log) carries the resolved `conversationId`/`messageTs`, not just the request — since the common case is omitting both.

## Test plan

- [x] `mise run api-server:check` (lint, tsc, format) — clean
- [x] `mise run api-server:test` — 665/665 passing, including new coverage for: binding/gateway gates, message-not-found, cross-channel resolution, missing-scope tool omission, ambiguous/no-active-turn errors, and the audit log capturing the resolved target when the call omits chatId/messageTs
- [x] `mise run docs:check:doc-size` — `channels.md` within budget
- [ ] Manual verification against a real Slack workspace (not done — no live cluster/Slack app touched in this session)


pilartomas:
Closing — folding this into a single PR that implements #2979 in full (reactions + GitHub handle mapping + message metadata + timezone-aware scheduling).

PR #2992, opened 2026-07-27
feat(channels): rich Slack user & message metadata for agents (#2979)

pilartomas:
## Summary

Implements issue #2979 — richer Slack user and message metadata surfaced into an agent's context, plus the timezone-aware scheduling angle it raised.

- **Reactions** (`describe_message_reactions`, new tool): who reacted to a message and with what emoji, via Slack's `reactions.get`. Defaults to the current turn's message/bound channel, or an explicit `chatId`/`messageTs` for a message outside the live turn (e.g. a scheduled agent checking a recurring signup thread). Deliberately never cached — a reaction tally is live state. Gated on a new optional `reactions:read` scope, following the same "omit the tool entirely when the scope is confirmed missing" pattern already established for `describe_channel_users`/`users:read`.
- **Message metadata**: injected thread/channel history now shows a human-readable UTC timestamp per line and an "(edited)" marker when Slack reports a message changed. The per-turn contract states the triggering message's send time, a permalink to it (when Slack can resolve one), and whether the conversation is a 1:1 DM or a shared channel/group.
- **Timezone-aware scheduling**: `create_schedule` now accepts `rrule`+`timezone` as a DST-correct alternative to the legacy UTC-only `cron` (mutually exclusive) — reusing the RRULE/quiet-hours machinery the UI/CLI already had; an agent-registered schedule is now attributed and audited as agent-created rather than defaulting to "user".

Deliberately **not** included: the issue's GitHub-handle mapping idea. A GitHub handle isn't Slack data — it would need its own platform-side mapping table, settings UI, and identity-link plumbing — and we decided the value doesn't justify that surface. `describe_channel_users` keeps returning only what the Slack workspace itself discloses.

## Test plan

- [x] `mise run api-server:check` (lint, tsc, format) — clean
- [x] `mise run api-server-api:check`, `mise run db:check`, `mise run ui:check` — clean
- [x] `mise run api-server:test` — 85 files / 720 tests passing
- [x] `mise run ui:test` — 12 files / 108 tests passing
- [x] `mise run docs:check:doc-size` — `channels.md` within budget
- [x] Two independent adversarial review passes on the reactions tool (one caught a real audit-logging gap on the default-usage path, fixed with regression tests)
- [ ] Manual verification against a real Slack workspace (not done — no live cluster/Slack app touched in this session)


packages/api-server/src/modules/channels/infrastructure/slack.ts
```diff
@@ -823,6 +865,8 @@ export function createSlackWorker(
       eventTs: ctx.eventTs,
       brand,
       canLookupUsers: await canLookupUsers(gw),
+      isDirectMessage: isDirectMessageId(ctx.channel),
+      permalink: await gw.getPermalink(ctx.channel, ctx.eventTs),
```
dam-code-guardian: 🟡 **Warning:** `getPermalink` is now awaited on the critical path of every Slack turn (mention here, DM ~L1040, ambient batch ~L1783), adding a synchronous `chat.getPermalink` Web API round-trip before the prompt is framed. It's resilient (returns `null`, never throws), but it serializes an extra Slack call into turn setup and spends rate-limit budget on every turn. Consider resolving it lazily or in parallel with the other setup awaits, or only when the contract will actually surface the permalink.

packages/api-server/src/modules/channels/infrastructure/agent-footer.ts
```diff
@@ -95,9 +95,30 @@ export function historyLegend(canLookupUsers: boolean): string {
...
- *  from its footer: the reading agent's own posts become "you (this agent)",
- *  other agents' posts name that agent, and everyone else keeps their Slack id. */
+/** Render a Slack `ts` (seconds.microseconds since the epoch) as a human
+ *  string, e.g. "Mon 2026-07-27 14:32 UTC" — Slack's history and mention
+ *  payloads carry only the raw epoch, which an agent would otherwise have to
+ *  parse by hand. Always UTC, never a message's sender's local time:
+ *  resolving that would cost a directory lookup per message, and an agent
+ *  that needs a specific person's local time can already get it via
+ *  describe_channel_users. Falls back to the raw string on anything unparseable. */
+export function formatSlackTs(ts: string): string {
```
dam-code-guardian, line 105: 🟢 **Suggestion:** Nice defensive parse — the `Number.isFinite` guard falls back to the raw `ts` on anything unparseable. Worth a one-line test asserting that fallback branch (e.g. a non-numeric `ts`) if it isn't already covered.

packages/api-server/src/modules/github-links/index.ts
```diff
@@ -0,0 +1,6 @@
+export { composeGithubLinksForOwner } from "./compose.js";
+// Exposed directly (not owner-bound) for callers that resolve an arbitrary
+// owner rather than "the calling user" — e.g. the Slack worker looking up
+// the handle behind a linked identity.
+export { createGithubLinksRepository } from "./infrastructure/github-links-repository.js";
```
dam-code-guardian: 🟢 **Suggestion:** Exporting the infrastructure repository factory (not just the composed service) is a documented deviation from "infrastructure stays private" — justified here since the Slack worker needs an owner-agnostic lookup, and cleanly mediated (the consumer defines its own `GithubHandleLookup` port, the composition root wires it, no module imports another's internals). Fine as-is; a second owner-agnostic *service* method would be the purist alternative.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `5317181c`

## PR #2992: feat(channels): rich Slack user & message metadata for agents (#2979)
**Author:** pilartomas | **Branch:** feat/slack-message-reactions → main | **Changes:** +4347 −57 (48 files)

### Summary
Implements #2979: a new `describe_message_reactions` MCP tool (who reacted, with what emoji, via Slack `reactions.get`), a self-service GitHub-handle mapping joined onto `describe_channel_users`, richer per-turn/injected-history message metadata (readable UTC timestamps, `(edited)` markers, DM-vs-shared, permalinks), and timezone-aware `create_schedule` (`rrule`+`timezone` alongside legacy UTC `cron`) with agent-created attribution. New `github_handles` table/migration; `channels.md` updated in the same PR.

### Findings
- ✅ **Looks good:** The reactions tool is gated exactly like `describe_channel_users` — proactively omitted when `reactions:read` is confirmed missing, fail-open on unknown (`canReadReactions`, `slack.ts:849`). Scope added to both the manifest and `docs/configuration.md`.
- ✅ **Looks good:** The `describe_channel_users` batch resilience is handled well — the GitHub-handle lookup is wrapped in its own try/catch that degrades to an absent field, so a DB failure can't cost an id its resolved profile or abort the batch (`slack.ts:2478-2486`). Matches the resilience fix the PR body describes.
- ✅ **Looks good:** `create_schedule` mutual-exclusivity is correctly enforced — `(cron === undefined) === (rrule === undefined)` rejects both-or-neither, plus an explicit `rrule`-requires-`timezone` check, before any service call (`mcp-endpoint.ts:431-438`).
- 🟡 **Warning:** `getPermalink` is now `await`ed on the critical path of **every** Slack turn — mention (`slack.ts:869`), DM (`slack.ts:1040`), ambient batch (`slack.ts:1783`) — adding a synchronous `chat.getPermalink` Web API round-trip before the turn's prompt is framed. It's resilient (returns `null` on failure, never throws), but it serializes an extra Slack call into turn setup and consumes rate-limit budget on every turn. Consider fetching it lazily / in parallel with other setup, or only when the contract will surface it. (`packages/api-server/src/modules/channels/infrastructure/slack.ts:869`)
- 🟢 **Suggestion:** `formatSlackTs` guards the unparseable case via `Number.isFinite` and falls back to the raw string — solid. Worth a one-line test asserting that fallback branch if not already covered. (`agent-footer.ts:105`)
- 🟢 **Suggestion:** `github-links/index.ts` exports the infrastructure repository factory (not just the composed service) so the Slack worker can do owner-agnostic lookups — a documented, cleanly-mediated deviation (the consumer defines its own `GithubHandleLookup` port). Fine as-is; a second owner-agnostic service method would be the purist alternative. (`github-links/index.ts:5`)

Verdict: `aligned`. `channels.md` (the one architecture page touched) was updated in the same PR to cover the reactions tool + `reactions:read` gating, the `githubHandle` field, and the new per-turn/history metadata, with `Last verified:` bumped to `2026-07-28`. No ADR/volatile-content leaks. The `github-links` module is a thin data mapping, not a long-lived component, so no new page is required; the schedules change leaves `agent-lifecycle.md`'s documented behavior accurate. No findings.

1 finding. 🟢 **Suggestion (module public API):** `github-links/index.ts` exposes `createGithubLinksRepository`/`GithubLinksRepository` alongside the composed service — a documented, contained deviation from "infrastructure stays private" (the Slack worker needs an owner-agnostic lookup). It's mediated correctly: the consumer defines its own `GithubHandleLookup` port and the composition root wires it, so no module imports another's internals. The new module's service→repo→compose→index layering, contract-only `api-server-api`, `db`-package schema, and the localized `createdBy` signature extension all align. No boundary violations introduced.

1 finding. 🟢 **Suggestion:** the `useEffect` seeding `value` from `savedHandle` correctly reconciles server↔local lineage (gates on `!touched` so a refetch won't clobber typing); informational only. Otherwise clean: server state in TanStack Query with the project's `meta.invalidates`/string `errorToast` convention, single-field controlled `useState` (correctly below the RHF threshold), shared `githubHandleSchema` for validation, proper module structure, no `React.FC`, no static `style`. No CRITICAL/HIGH violations.

### Verdict
COMMENT — a well-structured, well-tested feature with clean layering, correct scope-gating, and matching doc updates. No blocking issues; the one 🟡 is the per-turn `getPermalink` round-trip added to turn setup — worth confirming it won't pressure Slack rate limits or add latency to every turn.

---

pilartomas:
## Review (current head, 4 commits / 20 files)

**Overall: solid, approvable after two small fixes.** The design follows the channels architecture doc faithfully (and updates it accurately), the scope-gating pattern is genuine parity with `describe_channel_users`, and the rrule path reuses the existing validated machinery rather than reimplementing it. Re-verified on this head: `api-server:check` / `api-server-api:check` clean, 85 files / 720 tests green, and no leftover references to the removed GitHub-handle feature anywhere in the diff. Findings below come from four independent review passes plus adversarial verification of every non-nit claim, each re-checked against the current head.

### Worth fixing before merge

**1. `describe_message_reactions` discards the turn's conversation when defaulting — unlike `react`/`reply`** (`packages/api-server/src/modules/channels/infrastructure/slack.ts`, `describeMessageReactions`). When `messageTs` is omitted it takes `turn.ref.eventTs`, but then resolves the channel as `query.conversationId ?? bound channel`, throwing away `turn.ref.channel`. Its siblings carry it deliberately — `react` sets `turnChannel = turn.ref.channel` with a comment naming exactly this hazard ("the message lives in the turn's conversation, not necessarily the currently-bound one") and even back-fills the channel via `findTurnRef` when an explicit ts matches a known turn. Consequences: a mid-turn or lingering-window (60 min) rebind makes the advertised no-args default look up the right ts in the wrong channel → "message not found" (a scenario the codebase explicitly defends elsewhere — there's a regression test for id-less `reply` under rebind); and the `chatId`-without-`messageTs` combination pairs the current turn's ts with the foreign chat, a guaranteed confusing miss. Read-only tool, loud error, so minor — but the fix is the same two lines `react` already has.

**2. `create_schedule` silently ignores `timezone` and `quietHours` alongside `cron`** (`packages/api-server/src/apps/harness-api-server/mcp-endpoint.ts`). The guards only enforce cron-xor-rrule and rrule-requires-timezone. An agent passing `cron: "0 9 * * *", timezone: "Europe/Prague"` — believing cron fires in that zone — gets a silently-UTC schedule: precisely the DST-drift bug class this feature exists to kill, now with false reassurance. Reject the combination with an actionable error.

### Pre-existing gaps the PR newly exposes to agents (suggest follow-up issues, not pre-merge fixes)

These live in untouched machinery (`recurrences.ts` is byte-identical to main), but the tool now hands RRULE authoring to agents:

- **RRULE `COUNT` is never honored** — `nextFireAt` re-anchors `dtstart` to "now" on every evaluation, so a `COUNT=3` schedule fires forever. Empirically reproduced with the repo's rrule@2.8.1.
- **A never-firing rrule (e.g. `UNTIL` in the past) is accepted and reported to the agent as success** — validators pass, the runner sets next-run to null, the agent believes it scheduled something.
- **No frequency floor**: `FREQ=SECONDLY` validates and fires every second (each fire = DB read, outbox bump, agent wake, re-enqueue). Not a regression — the legacy cron path already accepted 6-field seconds expressions — but agents self-registering schedules makes a floor worth having.

### Nits

- Unused `SessionType` import in `mcp-endpoint.ts` (no lint rule catches it).
- The new per-turn contract line "You're answering a message sent …" doesn't branch for coalesced ambient batches — the contract already models `batch.count > 1` for the reply bullet, but this sentence states one send time for a multi-message prompt.
- `AMBIGUOUS_THREAD_ERROR` copy ("…so this lands in the thread you are answering") is written for the posting tools but now also surfaces from this read-only tool.
- The permalink-**absent** contract branch is untested — the fake gateway's `getPermalink` always succeeds and no test overrides it (the present-branch and DM/channel-kind branches are covered in `slack-cross-agent-history.test.ts`).
- Design note, no action needed: `chat.getPermalink` is now awaited on the relay critical path of all three turn paths. Failure degrades to null safely, but it's one more Slack round trip of latency per inbound turn.

### Verified non-issues

Authorization is coherent (binding-gated, reach = bot membership, matching the documented "Slack reach" model); the security log records the resolved target and tally but never who reacted, exactly as channels.md claims; scope gating fails open on unknown, closed only on confirmed-missing, in both the worker and the aggregate; `formatSlackTs` handles garbage input; the fake gateway models `message_not_found → null` like bolt; agent attribution (`createdBy: "agent"`) flows into the security log correctly; docs and manifest comments match the implementation.

pilartomas:
Addressed in `417134a6`.

**Both pre-merge findings fixed:**

1. **`describeMessageReactions` discarding the turn's conversation** — now carries `turn.ref.channel` when defaulting and back-fills an explicit `messageTs` from its turn via `findTurnRef`, i.e. the same two branches `react` already had. Covered by a new regression test that rebinds the agent mid-turn and asserts the lookup still resolves against the turn's channel (verified failing before the fix).

2. **`cron` + `timezone`/`quietHours` silently ignored** — now refused with `"`cron` is UTC-only and ignores `timezone`/`quietHours` — use `rrule` with `timezone` to schedule in a local zone."`. Two new tests, both verified failing before the guard.

**Nits fixed:**
- Unused `SessionType` import dropped.
- The send-time contract line now branches on a coalesced batch: instead of asserting one send time and one permalink for a multi-message prompt, it states the count and points at the per-message `[ts …]` tags (asserted in the existing top-level coalescing test, including that the single-message turn before it still names its own send time).
- Ambiguous-turn refusals from this read-only tool now use their own copy — the posting tools' "so this lands in the thread you are answering" misdescribed what's at stake when nothing is delivered.
- Permalink-**absent** branch now tested (overrides the fake gateway's `getPermalink` to return null, asserts the rest of the line survives).
- `channels.md` corrected: the chat default is the turn's own conversation, else the bound one — the old text said just "bound channel", which the fix makes inaccurate.

**Also addressed the `getPermalink` critical-path concern** (raised independently by Code Guardian): the scope probe and the permalink round trip now resolve via one `Promise.all` instead of back to back, and a coalesced batch — which no longer renders a permalink at all — skips the call rather than paying for a discarded result. Turn setup is one hop instead of two, and ambient batches drop the call entirely.

**Not fixed here, as suggested** — the three pre-existing RRULE gaps (`COUNT` never honored, never-firing rrule reported as success, no frequency floor). All live in `recurrences.ts`, untouched by this PR, and each needs its own decision about desired semantics. Happy to file them as follow-up issues on request.

Verification on this head: `api-server:check` / `api-server-api:check` / `db:check:generated` / `ui:check` clean, 85 files / **724 tests** green (up from 720), `doc-size` within budget.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `bc47a2db`

## PR #2992: feat(channels): rich Slack user & message metadata for agents (#2979)
**Author:** pilartomas | **Branch:** feat/slack-message-reactions → main | **Kind:** re-review

### Summary
Re-review at `bc47a2db`. Since my `5317181c` COMMENT the author pushed four fix commits addressing both the human reviewer's pre-merge findings and my own 🟡 `getPermalink` concern; the GitHub-handle feature (and its UI) was dropped from the PR entirely.

### Changes since last review
Previous HEAD: `5317181c` (2026-07-28T10:19:58Z) — verdict COMMENT

- ✅ **Fixed:** `getPermalink` no longer serializes an extra round-trip into turn setup — scope probe + permalink now resolve via one `Promise.all`, and coalesced ambient batches skip the call entirely (`slack.ts:503` `turnContractContext`).
- ✅ **Fixed:** the `github-links` module public-API deviation is moot — the whole GitHub-handle feature was removed from the PR, so no module-boundary surface remains.
- ✅ **Fixed (human finding 1):** `describeMessageReactions` now carries `turn.ref.channel` when defaulting and back-fills an explicit `messageTs` from its turn, matching `react` (`slack.ts:2537`).
- ✅ **Fixed (human finding 2):** `create_schedule` now rejects `cron` combined with `timezone`/`quietHours` instead of silently dropping them (`mcp-endpoint.ts:662`).

### Findings

Verdict: `aligned`. `channels.md` (the one architecture page touched) tracks the code — the reactions tool, its `reactions:read` proactive gating, the per-turn/injected-history metadata, and the corrected default-channel wording (turn's own conversation, else bound) are all reflected; `Last verified: 2026-07-28`. No ADR/volatile leaks; new files are tests only. No findings.

✅ No findings. The delta is a clean refactor: `describeMessageReactions` follows the same port/adapter shape as `describeUsers` (worker port → `ChannelManager` → MCP endpoint), the new `ChannelMessageReaction`/`ReactionsQuery`/`MessageReactionsResult` interfaces sit correctly in the service layer, `createRRule`'s optional `createdBy` mirrors `createCron`'s documented default-`"user"` convention, and the `timezone!`/`cron!` non-null assertions are soundly guarded by the preceding xor checks. My prior 🟢 (github-links public API) no longer applies — module removed.

_(skipped — no-matching-files; the UI handle editor was removed with the dropped GitHub-handle feature)_

### Verdict
APPROVE — every finding from the prior review round (mine and the human reviewer's two pre-merge items) is resolved at this HEAD, no new findings surfaced, and the one touched architecture page is in sync. The three RRULE gaps the author flagged live in untouched `recurrences.ts` and are correctly deferred to follow-up issues.

---

PR #3002, opened 2026-07-28
fix(e2e): restore the mock's default reply after the experiments spec

pilartomas:
## Summary
- `12-experiments.spec.ts` (added in #2948) scripts the shared e2e fixture agent's mock reply to `"script written"` via `setMockReplyWithFiles`, and never restores it afterward.
- Playwright's project-dependency scheduler now runs `experiments` right after `user-env` and before the `slack`/`slack-shared` chain, so the shared agent's mock process is left permanently scripted with that leftover text.
- `08-slack-shared.spec.ts`'s first test is the first spec after `12-experiments` to drive a plain (non-forked) turn on that same agent and assert on the mock's out-of-the-box `"Hello from the mock agent."` reply — it now times out waiting for text that will never appear. The turn completes successfully and a reply does land in the thread, just with the stale script text, so there's no error/denial for the test's other assertions to catch.
- The second test in `08-slack-shared.spec.ts` (the `__FETCH__` turn) is unaffected because its expected text is computed independently of the scripted reply, which is why only the first test fails.

This reproduces deterministically on CI (e.g. run [30274973335](https://github.com/dam-agents/dam/actions/runs/30274973335)) — same failure signature (`the reply did not land back in the slack thread`, 180s timeout) every time, since the ordering is fixed.

## Fix
Restore the mock's default reply as a final step of the experiments test, mirroring the baseline-preserving cleanup `09-user-env.spec.ts` already does for env vars it mutates on the same shared agent.

## Test plan
- [x] `mise run e2e-playwright:check:tsc`
- [x] `mise run e2e-playwright:check:lint`
- [x] `mise run e2e-playwright:check:format`
- [ ] CI `mise e2e` smoke run (this fix specifically targets `08-slack-shared.spec.ts`'s first test)

### DIFF
--- a/packages/e2e/playwright/src/tests/smoke/12-experiments.spec.ts
+++ b/packages/e2e/playwright/src/tests/smoke/12-experiments.spec.ts
@@ -5,6 +5,7 @@ import {
   chatInput,
   gotoAgentDetail,
   sendMessageToAgent,
+  setMockAgentReply,
   setMockReplyWithFiles,
   waitForAgentRunning,
 } from "../../lib/agents.js";
@@ -23,6 +24,10 @@ import { agentName } from "../../lib/fixtures.js";
 
 const experimentName = "e2e-loop";
 const scriptPath = "exp.py";
+// The mock's out-of-the-box reply; 07/08/10-slack rely on it verbatim. This
+// spec overwrites it via setMockReplyWithFiles below, so it must restore it
+// before the shared agent moves on to those specs.
+const mockDefaultReply = "Hello from the mock agent.";
 
 const experimentScript = `import experiment_sdk as x
 
@@ -173,5 +178,9 @@ test("experiment: plan, execute, watch it run to completion", async ({
       a.title.startsWith(`${experimentName} —`),
     );
     expect(script).toBeTruthy();
+  });
+
+  await test.step("restore the mock's default reply for the specs that follow", async () => {
+    await setMockAgentReply(api, agentId, mockDefaultReply);
   });
 });


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `4a1bd8f`

## PR #3002: fix(e2e): restore the mock's default reply after the experiments spec
**Author:** pilartomas | **Branch:** fix/experiments-e2e-restore-mock-default-reply → main | **Changes:** +9 −0 (1 file)

### Summary
Adds a final `test.step` to `12-experiments.spec.ts` that restores the shared mock agent's default reply (`"Hello from the mock agent."`) via `setMockAgentReply`, after the spec overwrites it with `"script written"` through `setMockReplyWithFiles`. This fixes a cross-spec ordering bug where `08-slack-shared.spec.ts`'s first test timed out waiting for the default reply that had been left permanently scripted.

### Findings
- ✅ **Looks good:** The fix targets the actual root cause. `setMockAgentReply` writes a fresh single-entry script (`agent_message_chunk` → the default text, `stopReason: end_turn`), fully replacing the leftover `setMockReplyWithFiles` script including its `files`, so the shared agent returns to its out-of-the-box behavior for the specs that follow.
- ✅ **Looks good:** Restored value matches the canonical default asserted by 07/08/10 (`"Hello from the mock agent."`, also the Helm `values-local.yaml` default), and mirrors the baseline-preserving cleanup pattern `09-user-env.spec.ts` already uses on the same shared agent.
- ✅ **Looks good:** The restore runs as the last `test.step` within the single test, on the same `api`/`agentId` used to script it — correct scope.
- 🟢 **Suggestion:** The restore only runs if the preceding steps pass; a mid-test failure would still leave the mock scripted and could cascade into 08. A `test.afterAll`/`afterEach` hook would make cleanup unconditional. Minor — the existing 09 pattern uses the same in-test approach, so this is consistency-preserving as written. (`packages/e2e/playwright/src/tests/smoke/12-experiments.spec.ts:182`)

✅ No findings. Test-only change; no architecture pages, subsystems, or ADR states affected.

✅ No findings. Reuses the existing `setMockAgentReply` lib helper idiomatically; no TSEng architecture rules implicated by a test-spec cleanup step.

### Verdict
APPROVE — a correct, minimal, well-documented fix for a deterministic cross-spec ordering bug, consistent with the existing baseline-restore convention.

---

PR #3004, opened 2026-07-28
fix(channels): keep relay-failed Slack turns resolvable so late replies can't cross threads

pilartomas:
## The bug

One more cross-thread reply incident occurred after the #2956 in-flight-turn fix and the #2961 ambient coalescing fix landed. The residual mechanism, confirmed at line level by adversarial review of both sides of the relay:

1. A Slack turn's relay can settle with a **transport error** while the pod-side harness **keeps executing the turn**. `sendPrompt` throws on a heartbeat abort (~90s of missed pongs on a half-open socket), on the turn ceiling, or on any ws teardown — and the agent-runtime's `detach()` deliberately keeps the active prompt running when its relay channel drops ("the agent is still working on it").
2. The worker's `finally` then removed the turn from the in-flight set, while the harness kept working and later called `reply` over the (independent) MCP path.
3. With the failed turn forgotten, an id-less late reply resolved against whatever was current — the sole live turn started by a newer mention, or the advanced last-active-thread fallback — and one thread's answer posted into another thread.

This needs a subsequent turn to have started before the late reply arrives, which is why it became rare after the earlier fixes but didn't disappear. Half-open relay sockets are a known failure mode of this environment (mesh cert expiry, #283), so the trigger is realistic.

## The fix

- **Lingering turns.** A turn that settles on a path that says nothing about the pod (the `acp-error` class, plus ghost runs left when a resume attempt fails mid-turn and the worker re-runs on a fresh session) keeps its ref in the registry for up to the turn ceiling (60 min), swept lazily. Settlements that prove the harness is done (success) or never ran the prompt (wake failure, agent stopped) drop the ref as before.
- **Resolution by distinct target.** Id-less `reply`/`react` resolution now considers live + lingering candidates and refuses only when they name more than one *target* (thread for `reply`, message for `react`). Candidates that agree — e.g. a retried relay of the same thread — resolve instead of refusing spuriously. Sole-live-turn ergonomics and the last-turn fallback are unchanged.
- **Turn-channel targeting.** A call resolved from a turn now posts into that turn's own conversation instead of the currently-bound channel, so a mid-turn rebind can't redirect an in-flight thread's reply into the newly bound channel.

Worst case of a stale lingering entry is a refusal the agent answers by echoing the id its prompt already injected — never a wrong-thread delivery.

## Testing

- New regression tests in `slack-turn.test.ts`: a relay-failed turn makes an id-less reply refused while another thread is live (the incident shape); it resolves to the failed turn once later turns finish (not to the last active thread); successful turns don't linger; lingering expires back to the fallback; same-thread candidates resolve a reply but refuse a react; a failed resume attempt keeps its turn resolvable after the fresh-session fallback succeeds; a mid-turn rebind no longer redirects the reply.
- New ambient-path test in `slack-ambient.test.ts` mirroring the incident shape across two read-along threads.
- `mise run api-server:test` → 676 passing; `api-server:check:{tsc,lint,format}` clean.

## Docs

`docs/architecture/channels.md` outbound resolution updated (lingering turns, target-based ambiguity, turn-channel targeting).

## Follow-ups (out of scope, found during review)

- A coalesced top-level ambient batch injects only its newest message's ids, so an answer to an earlier batched message threads under the last one; per-message ids in the batch prompt would fix it.
- Ambient turns advance the last-active-thread fallback even when the agent stays silent, so a later proactive id-less reply can target a drifted-by channel message.
- A mention on a thread with an in-flight ambient turn bypasses the per-session queue and can mint a duplicate session for the same thread key.
- Scheduled/UI sessions share the agent's MCP identity with channel turns, so their id-less `reply` can be attributed to a sole live Slack turn.


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `791d321`

## PR #3004: fix(channels): keep relay-failed Slack turns resolvable so late replies can't cross threads
**Author:** pilartomas | **Branch:** fix/slack-ghost-turn-reply-routing → main | **Changes:** +473 −66 (4 files)

### Summary
Closes the residual #2952 cross-thread window: a Slack turn whose relay settles with a transport-ish error (`acp-error`, or a ghost run from a failed resume) may still be executing pod-side, so its ref now **lingers** (up to `TURN_LINGER_MS` = 60 min, swept lazily) and keeps counting toward id-less `reply`/`react` resolution. Resolution is reframed around the *target* (thread for reply, message for react): agreeing candidates resolve, distinct ones refuse. A turn-resolved call now posts into the turn's own conversation, so a mid-turn rebind can't redirect the reply.

### Findings
- ✅ **Looks good:** The three settlement paths (owner, fork, ambient) are handled symmetrically — `harnessMayStillRun` is set on `acp-error`/`ghostTurn` (owner+ambient) and `turnOutcome === "failure"` (fork), while `agent-stopped` and wake-timeout classes correctly drop the ref (`slack.ts:834`, `slack.ts:1033`, `slack.ts:1685`).
- ✅ **Looks good:** `resolveTurn` target-set logic is correct — same-thread candidates collapse to one `reply` target but stay two `react` targets, exactly matching the regression test (`slack.ts:569`).
- ✅ **Looks good:** `turnChannel` threads through as `args.conversationId ?? turnChannel`, preserving an explicit `conversationId` while defaulting to the turn's channel — the rebind fix without regressing cross-channel posts (`slack.ts:2154`, `slack.ts:2208`).
- 🟢 **Suggestion:** `lingeringFor` sweeps expired entries only on read, so an agent that fails a turn and then goes permanently silent retains that `Map` entry until the next `reply`/`react` for it. Bounded (one entry per idle-after-failure agent) and consistent with the documented "swept lazily, no timers" intent — noting only for the record; no change needed (`slack.ts:530`).
- ✅ **Looks good:** Test coverage is thorough and maps 1:1 to the incident mechanics — failed-turn-stays-resolvable, resolves-to-failed-not-last-active, success-doesn't-linger, expiry-falls-back, same-thread reply-vs-react asymmetry, failed-resume ghost run, and mid-turn rebind.

Verdict: **aligned**. `docs/architecture/channels.md` — the `channels` subsystem's outbound-resolution behavior changed and its page was updated in the same PR (lingering turns, target-based ambiguity, turn-channel targeting), with `Last verified:` bumped 2026-07-27 → 2026-07-28. No ADR leaks, no volatile content added, no missing page.

✅ No architecture findings. The change is confined to the infrastructure-layer Slack adapter (`slack.ts`) and reuses the established module-scoped closure/`Map` pattern (`inFlightTurns`/`lastTurn` → `lingeringTurns`). No layer-boundary or dependency-direction violations introduced.

### Verdict
APPROVE — Correct, well-tested fix for a real residual cross-thread race; symmetric across all turn paths, docs kept in sync, no blocking issues.

---

PR #3005, opened 2026-07-28
fix(channels): close the ambient and session-race routes that misplace a Slack reply

pilartomas:
Follow-up to #3004, closing the routes listed there where a Slack reply can still land on the wrong message or thread. Each was confirmed by adversarial review of the worker; a final review pass over this diff surfaced one major and four minor defects, all fixed here.

## 1. Turns are serialized per (agent, thread session)

Concurrent prompts on one session collide in the runtime's per-session queue — which silently drops a queued prompt when its relay tears down — and the worker's list-then-create session match could mint two sessions for one thread key (a long-documented gap). A keyed lock now spans readiness, the session match, and the prompt:

- two rapid first messages in a brand-new thread take turns; the second finds and resumes the session the first minted;
- a mention on a thread with an in-flight read-along turn waits for it instead of racing the runtime queue;
- fork turns take the same lock — fork pods mount the agent's PVC, so they share the session store with the main pod (this was the review's major finding: the lock originally covered only the owner/ambient path).

## 2. Coalesced ambient batches target per message

A multi-message batch used to inject only its newest message's ids, so an answer to an earlier batched message silently threaded under the newest one. Now each batched message is tagged `[ts …]` in the prompt and registered as its own reply target: a multi-message top-level batch refuses id-less replies and the agent passes the tag of the message it answers; thread batches still resolve their single thread id-less, while reacts name the tagged message.

## 3. The proactive-reply fallback advances on engagement, not exposure

Ambient read-along turns repointed the last-active-thread fallback on every drifted-by channel message — even ones the agent silently ignored — so a later proactive id-less reply could thread under a random bystander message. The fallback now advances when an addressed turn (mention, DM, fork) starts, or when the agent actually replies/reacts to a read-along turn (preferring the newest matching batch message).

## 4. Hardening from the review pass

- Explicit `threadTs`/`messageTs` naming a live or lingering turn follows that turn's conversation, exactly like id-less calls — batch turns *must* pass ids, so the mid-turn-rebind protection can't hinge on omission.
- Ambient turn refs register inside the `try` (mirroring the owner path) so a setup throw can't leak live registry entries.
- `ensureReady` moved inside the lock, so a queued turn re-verifies pod readiness when it actually runs rather than when it queued.

## Not fixed (needs a protocol change)

Scheduled/UI sessions share the agent's MCP identity with channel turns, so their id-less `reply` can still be attributed to a sole live Slack turn. Closing that requires session correlation on the MCP tool calls, which the third-party harness binaries don't carry — same constraint recorded in #2956.

## Testing

- New: fork-vs-fork same-thread serialization; mention-behind-ambient serialization; duplicate-session prevention for two rapid first messages; multi-message top-level batch refusing id-less replies and honouring a tagged older ts; silent read-along turns not repointing the fallback; engaged turns becoming the fallback (newest batch message); explicit-id rebind protection.
- Updated: ambient tests that encoded the old fallback semantics now reply during the turn (the realistic flow) and pin the new batch behavior.
- `mise run api-server:test` → 681 passing; `api-server:check:{tsc,lint,format}` clean (re-verified after rebasing onto main past #3004's squash merge).

## Docs

`docs/architecture/channels.md`: thread-key serialization, per-message batch targeting, engagement-based fallback; some surrounding mechanics deduplicated to keep the page within its size budget.


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `f6d63be`

## PR #3005: fix(channels): close the ambient and session-race routes that misplace a Slack reply
**Author:** pilartomas | **Branch:** fix/slack-ambient-routing-followups → main | **Changes:** +458 −85 (4 files)

### Summary
Follow-up to #3004: closes the remaining routes where a Slack reply can land on the wrong message/thread. Adds a per-`(agent, thread-session)` turn lock spanning session match + prompt (kills duplicate-session minting and cross-turn prompt-queue races, incl. fork turns which share the PVC), tags each coalesced batch message with its own `[ts …]` so a top-level batch targets per message instead of silently threading under the newest, and advances the proactive-reply fallback on engagement rather than mere exposure.

### Findings
- ✅ **Looks good:** `withSessionTurnLock` is a correct promise-chain — `prev.then(fn, fn)` runs the next turn whether the prior resolved or rejected (a failed turn never wedges the key), the caller gets the real result/rejection via `run`, and the map entry self-deletes only when it's still the tail. No reentrancy (reply/react don't take the lock), so the shared owner/fork key can't deadlock.
- ✅ **Looks good:** Per-message `turnRefs` give a top-level batch distinct `threadTs` per ref, so an id-less `reply` is correctly ambiguous→refused while a thread batch (shared `threadTs`) still resolves; explicit-id calls route through `findTurnRef` so batch replies keep the mid-turn-rebind protection (`slack.ts:2286`, `:2337`).
- ✅ **Looks good:** `advanceLastTurn: false` on ambient start plus `noteEngagedTurn` (newest-first match on the ids actually posted) implements engagement-based fallback precisely — a silently-ignored read-along never becomes the proactive target (`slack.ts:559`, `:614`).
- ✅ **Looks good:** Ambient refs registered inside the `try` and torn down for every ref in `finally` with the same `harnessMayStillRun` classing as the owner path — a setup throw can't leak a live (never-expiring) ref.
- 🟢 **Suggestion:** `findTurnRef` is `O(live + lingering)` per explicit-id `reply`/`react` and `resolveTurn`/`noteEngagedTurn` each rescan too; fine at current fan-out, but if per-agent concurrent-turn counts ever grow, an index keyed by `threadTs`/`eventTs` would avoid the repeated linear scans. Non-blocking.

✅ No findings. `docs/architecture/channels.md` is updated in the same PR — the three behavioral changes (per-thread-key serialization, per-message batch targeting, engagement-based fallback) are all reflected, `Last verified: 2026-07-28` matches the PR date, no ADR references or volatile-content leaks.

✅ No findings. Change stays entirely within the existing `infrastructure/slack.ts` adapter and its co-located unit tests — no layer-boundary crossings, no port/DI or optional-dependency rule violations introduced.

### Verdict
APPROVE — a careful, well-tested set of race-closure fixes; each defect has matching coverage (duplicate-session prevention, fork/mention serialization, per-message batch targeting, engagement fallback, explicit-id rebind protection) and the docs track the code. The one remaining gap (scheduled/UI id-less replies) is explicitly scoped out pending a protocol change.

---

PR #3006, opened 2026-07-28
fix(ui): show agent names instead of ids in artifact attribution

### LINKED ISSUE #2923: UI - Show agent names instead of agent IDs in Artifacts
## Problem

Artifacts currently display the underlying agent ID (e.g. `agent-9df53f7718ad0cd1`) as the source of the artifact.

These IDs are not meaningful to users and make it difficult to understand which agent created an artifact, especially when users have multiple agents.

## Expected behavior

Display the user-defined agent name instead of the internal agent ID wherever artifacts reference the creating agent.

Example:

**Current**
> agent-9df53f7718ad0cd1

**Expected**
> Code Review Agent

## Current state

Image

xjacka:
Closes #2923.

## What

Artifact rows in the library showed the raw internal agent id (`agent-9df53f7718ad0cd1`) as the artifact's creator. They now show the user-defined agent name.

## How

- New `useAgentDisplayName(agentId)` hook in `modules/agents/api/queries.ts` resolves an agent id to its name from the already-polled agents list, falling back to the raw id — artifacts deliberately outlive their creating agent (see `docs/architecture/artifact-library.md`), so a deleted agent's id is the only attribution left.
- `CreatorChip` in `artifact-row.tsx` uses the hook for both the label and the tooltip; click-through navigation still keys off the id.
- DRY cleanup: the two existing inline copies of the same lookup (egress approval toast, chat view) now share the pattern — the toast uses the new hook, chat view reuses its existing `agentView` lookup.

No contract or server change: agents are k8s CRs, so resolving the name client-side keeps it live across renames and avoids coupling artifact-library to the agents module.

### DIFF
--- a/packages/ui/src/modules/agents/api/queries.ts
+++ b/packages/ui/src/modules/agents/api/queries.ts
@@ -61,6 +61,8 @@ export function useAgentRunState(
  * references (artifact attribution, approvals) deliberately outlive their
  * agent, so a deleted agent's id is the only attribution left.
  */
+export function useAgentDisplayName(agentId: string): string;
+export function useAgentDisplayName(agentId: string | null): string | null;
 export function useAgentDisplayName(agentId: string | null): string | null {
   const agents = useAgentsList();
   if (!agentId) return null;
--- a/packages/ui/src/modules/agents/api/queries.ts
+++ b/packages/ui/src/modules/agents/api/queries.ts
@@ -57,6 +57,17 @@ export function useAgentRunState(
 }
 
 /**
+ * Resolve an agent id to its user-facing name. Falls back to the raw id —
+ * references (artifact attribution, approvals) deliberately outlive their
+ * agent, so a deleted agent's id is the only attribution left.
+ */
+export function useAgentDisplayName(agentId: string | null): string | null {
+  const agents = useAgentsList();
+  if (!agentId) return null;
+  return agents.find((a) => a.id === agentId)?.name ?? agentId;
+}
+
+/**
  * Whether the UI can actually talk to the pod right now. Three inputs, because
  * each catches a gap the others miss:
  *   - server reports `running` (the lifecycle truth, but it lags reality),
--- a/packages/ui/src/modules/artifacts/components/artifact-row.tsx
+++ b/packages/ui/src/modules/artifacts/components/artifact-row.tsx
@@ -146,11 +146,15 @@ export function ArtifactRow({
 }
 
 function CreatorChip({ agentId }: { agentId: string | null }) {
-  const navigateToSandboxHome = useStore((s) => s.navigateToSandboxHome);
-  const agentName = useAgentDisplayName(agentId);
   if (!agentId) {
     return <span className="rounded-full bg-muted px-2 py-px">you</span>;
   }
+  return <AgentCreatorChip agentId={agentId} />;
+}
+
+function AgentCreatorChip({ agentId }: { agentId: string }) {
+  const navigateToSandboxHome = useStore((s) => s.navigateToSandboxHome);
+  const agentName = useAgentDisplayName(agentId);
   return (
     <button
       type="button"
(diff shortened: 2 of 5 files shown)


packages/ui/src/modules/agents/api/queries.ts
```diff
@@ -56,6 +56,17 @@ export function useAgentRunState(
   return agentId ? agents.find((a) => a.id === agentId)?.state : undefined;
 }
 
+/**
+ * Resolve an agent id to its user-facing name. Falls back to the raw id —
+ * references (artifact attribution, approvals) deliberately outlive their
+ * agent, so a deleted agent's id is the only attribution left.
```
dam-code-guardian, line 62: 🟢 **Suggestion:** This hook returns `string | null`, so `CreatorChip`'s `agentName` stays typed `string | null` even though at its render site (past the `if (!agentId)` guard) the `?? agentId` fallback guarantees a string. Cosmetic only — no runtime issue.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `e841c48`

## PR #3006: fix(ui): show agent names instead of ids in artifact attribution
**Author:** xjacka | **Branch:** fix/artifact-agent-name → main | **Changes:** +18 −8 (4 files)

### Summary
Replaces raw internal agent ids with the user-defined agent name in artifact attribution (and reuses the same lookup in the egress-approval toast and chat view). Adds a `useAgentDisplayName(agentId)` hook that resolves the name from the already-polled agents list, falling back to the raw id so attribution for deleted agents still renders.

### Findings
- 🟢 **Suggestion:** `useAgentDisplayName` returns `string | null`, so `CreatorChip`'s `agentName` is typed `string | null` even though it's only rendered after the `if (!agentId)` guard where the `?? agentId` fallback guarantees a string. Cosmetic only — no runtime issue. (`packages/ui/src/modules/agents/api/queries.ts:62`)
- ✅ **Looks good:** DRY consolidation — two inline `useAgents().list.find(...)?.name ?? id` copies collapse into one shared hook; chat-view reuses its already-computed `agentView`.
- ✅ **Looks good:** Hook is called unconditionally before the early return in `CreatorChip` — Rules of Hooks respected.
- ✅ **Looks good:** Fallback-to-id behavior matches the documented invariant in `docs/architecture/artifact-library.md` for deleted agents; click-through still keys off the id.

✅ No findings. Verdict: **aligned** — a UI presentation refinement (id→name in the render layer); it alters no subsystem's behavior or responsibility, edits no architecture page, and is consistent with the existing attribution note in `artifact-library.md`.

✅ No findings. The new hook sits correctly in `modules/agents/api/queries.ts` alongside its siblings (`useAgentRunState`, `useAgentsList`), introduces no `any`/`as`, and respects module/layer boundaries.

✅ No findings. Correct state lineage (server list stays in TanStack Query, accessed via the shared hook), no Rules-of-Hooks violation, and a genuine DRY win removing two duplicated lookups.

### Verdict
APPROVE — clean, well-scoped DRY fix; correct hook placement and lineage, behavior-preserving, and consistent with the documented attribution model.

---

packages/ui/src/modules/artifacts/components/artifact-row.tsx
```diff
@@ -145,22 +146,27 @@ export function ArtifactRow({
...
         e.stopPropagation();
         navigateToSandboxHome(agentId, "artifacts");
       }}
-      title={`Open ${agentId}'s artifacts`}
+      title={`Open ${agentName}'s artifacts`}
       className="inline-flex items-center gap-1 rounded-full bg-muted px-2 py-px transition-colors hover:bg-accent-light hover:text-accent"
     >
       <Box size={12} />
-      {agentId}
+      {agentName}
```
PetrBulanek: The chip now renders a user-controlled string with no truncation, which was safe while it was a fixed-width 22-char id, but it can stretch now and push the time/views etc. We should probably truncate the agent name with CSS.

PetrBulanek (changes requested):
`packages/ui/src/modules/approvals/components/approvals-list.tsx:114` still renders agent {row.agentId} in full density, so the Inbox row reads `agent-bf17…` while the toast right beside it now reads the agent name. A bit outside of the scope, but I think we can fold into this.

PetrBulanek (approved):

PR #3009, opened 2026-07-28
fix(knowledge-bases): resolve CodeQL file-system-race on the workspace-command sentinel

Tomas2D:
Fixes CodeQL alert #171 (`js/file-system-race`, high) left open on `main` after #2971 merged.

The `existsSync` check followed by a `writeFile` on the sentinel matches CodeQL's check-then-act pattern even with the exclusive `wx` flag added in `f7bd3ab4`. This replaces the `existsSync` probe with a `readFile` (ENOENT → not done), removing the flagged pattern entirely. The `wx` exclusive write remains the only arbiter should two settles ever overlap.

The race itself is benign — the sentinel is single-writer per pod — so this is a pattern-level cleanup to clear the alert.

### DIFF
--- a/packages/agent-runtime/src/modules/runtime-channel/drivers/workspace-command-plugin.ts
+++ b/packages/agent-runtime/src/modules/runtime-channel/drivers/workspace-command-plugin.ts
@@ -1,6 +1,5 @@
 import { spawn } from "node:child_process";
-import { existsSync } from "node:fs";
-import { writeFile } from "node:fs/promises";
+import { readFile, writeFile } from "node:fs/promises";
 import { join } from "node:path";
 import type {
   DriverBinding,
@@ -50,7 +49,7 @@ export function createWorkspaceCommandPlugin(deps: {
       return async (payload, ctx) => {
         const { command } = payload as WorkspaceCommandEventPayload;
         const sentinel = join(ctx.pluginStateDir, DONE_SENTINEL);
-        if (existsSync(sentinel)) {
+        if (await sentinelExists(sentinel)) {
           deps.log(`[workspace-command] already run, skipping`);
           return;
         }
@@ -69,6 +68,20 @@ export function createWorkspaceCommandPlugin(deps: {
       };
     },
   };
+}
+
+// Probe by reading rather than existsSync: the sentinel is single-writer per
+// pod, and reading avoids the check-then-write pattern CodeQL flags as a
+// file-system race (js/file-system-race); the wx write below stays the only
+// arbiter if two settles ever overlap.
+async function sentinelExists(path: string): Promise<boolean> {
+  try {
+    await readFile(path);
+    return true;
+  } catch (err) {
+    if ((err as NodeJS.ErrnoException).code === "ENOENT") return false;
+    throw err;
+  }
 }
 
 // `bash -lc` so the command sees the same login-shell environment an


packages/agent-runtime/src/modules/runtime-channel/drivers/workspace-command-plugin.ts
```diff
@@ -71,6 +70,20 @@ export function createWorkspaceCommandPlugin(deps: {
   };
 }
 
+// Probe by reading rather than existsSync: the sentinel is single-writer per
+// pod, and reading avoids the check-then-write pattern CodeQL flags as a
+// file-system race (js/file-system-race); the wx write below stays the only
+// arbiter if two settles ever overlap.
+async function sentinelExists(path: string): Promise<boolean> {
+  try {
+    await readFile(path);
```
dam-code-guardian, line 79: 🟢 **Suggestion:** `readFile(path)` loads the whole sentinel into memory just to test existence. The file is tiny so this is negligible, but `access` expresses intent more directly and reads nothing (CodeQL flags `access`-then-act too, so it's equally clean). Purely stylistic — the current form is fine.

```suggestion
    await access(path);
```
(would also need `import { access } from "node:fs/promises";`)

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `baa8b2e1`

## PR #3009: fix(knowledge-bases): resolve CodeQL file-system-race on the workspace-command sentinel
**Author:** Tomas2D | **Branch:** fix/workspace-command-sentinel-race → main | **Changes:** +16 −3 (1 file)

### Summary
Clears CodeQL alert #171 (`js/file-system-race`, high) on the workspace-command plugin sentinel by replacing the `existsSync` check-then-act probe with a `readFile`-based `sentinelExists` helper (ENOENT → not done). The exclusive `wx` write remains the sole arbiter if two settles ever overlap.

### Findings
- ✅ **Looks good:** The `sentinelExists` helper narrows the caught error to `NodeJS.ErrnoException`, returns `false` only on `ENOENT`, and re-throws every other I/O error rather than silently treating it as "not run" — a real transient read failure won't cause a spurious double-run (`workspace-command-plugin.ts:74-85`).
- ✅ **Looks good:** Behavior is preserved end-to-end — the `wx` exclusive-create write and its `EEXIST` tolerance are untouched, so the concurrent-settle safety property the PR relies on still holds (`workspace-command-plugin.ts:63-67`).
- 🟢 **Suggestion:** `readFile(path)` loads the whole sentinel into memory just to test existence; the file is tiny so this is negligible, but `fs.promises.access(path)` (or `stat`) expresses intent more directly and reads nothing. CodeQL flags `access`-then-act too, so this is purely stylistic — the current form is fine. (`workspace-command-plugin.ts:79`)

Verdict: `aligned`. The change is an internal probe-mechanism swap with no shift in subsystem behavior. `docs/architecture/knowledge-bases.md` describes the sentinel behaviorally ("the plugin writes a sentinel on success, so a redelivery never double-runs a completed install") at the correct altitude and stays accurate. No findings.

✅ No findings. Clean async adapter probe: proper error-type narrowing, ENOENT distinguished from real I/O failures, single-writer/`wx`-arbiter semantics preserved. Aligned with the architecture docs.

### Verdict
APPROVE — a minimal, correct, well-documented pattern-level fix that removes the flagged check-then-act sequence while preserving the sentinel's run-once guarantee.

---

PR #3014, opened 2026-07-28
feat: unify experiments and knowledge bases behind one Agent Kind rail and one wizard

jezekra1:
## What

Experiments and Knowledge Bases were the same feature built two different ways. A
Knowledge Base is an Agent carrying a Kind marker, set up by an Install Command
over the durable `workspace-command` rail, with its own create page. An Experiment
was only a database row — the driver agent was incidental metadata, there was no
way to create anything from the UI, and the `dam-experiment` authoring skill was
baked into *every* claude-code sandbox's workspace whether or not it was meant to
run loops.

This gives experiments the same shape, extracts the shared machinery, and folds
all creation into one wizard.

## How it works

- **`experiment` Agent Kind.** The marker is **declared intent, not a capability
  gate**: Plan Registration is keyed only on the calling agent's waypoint
  identity, so any Agent can register a plan. The Experiments destination
  therefore lists **marked sandboxes ∪ agents with ≥1 experiment row** — a marked
  sandbox with nothing in it yet is an empty container, and a plain sandbox that
  registered a plan earns one too. No backfill; nothing disappears.
- **Shared kinded-create rail.** `createKindedAgent` (agents module) does what
  both kinds share: create with the marker, deliver the Install Command, wake, log.
  Each kind supplies only a kind literal and a command. Cut the KB service from
  ~55 lines to a 7-line call.
- **The skill moved behind the marker.** `dam-experiment` now lives at
  `/usr/local/share/dam-skills/`, outside the seeded workspace the first-boot
  `cp -rn` copies, and an Install Command copies it in. Nothing is fetched over the
  network — unlike a KB bootstrap, the kit ships with the image. Sandboxes seeded
  before the move keep their copy; the marker is not retroactive.
- **Onboarding greeting.** `/experiment-onboard` ships beside the skill and is
  hidden-sent on a fresh sandbox, **gated on the skill appearing** in the
  reconciled skills read. The copy installs the command *before* the skill, so the
  skill's presence implies both. Bounded to ~2 min and gated on the agent running.
- **One wizard.** Step 1 is now five starting points (Experiment / Knowledge base /
  Specialized / General-purpose / Custom) with a contextual reveal; steps 2–3 are
  shared; finish dispatches on kind. Retires the standalone KB create page.
- **Sandboxes lists everything, badged by Kind.** The `!isKnowledgeBase` filter is
  gone: a knowledge base and an experiment sandbox *are* sandboxes, and hiding them
  removed them from the owner's inventory of what they're paying to run.
- **Providers on the kinded paths** are restricted to the two credentials that
  reach Claude (IBM LiteLLM proxy, Anthropic), steering toward the proxy with a
  Recommended badge and preselecting it when connected.

## Notable decisions / things to review

- **The marker can't mean for experiments what it means for KB.** A KB's marker is
  the whole story; an experiment sandbox's is only intent, because the SDK ships in
  `platform-base` for everyone. That asymmetry is why the destination unions two
  populations instead of filtering on the marker — see `toSandboxGroups` and its
  tests.
- **No per-experiment route.** One sandbox holds many experiments, so opening
  either a container or a lineage lands in the sandbox's chat where the panel docks.
- **Anthropic deliberately kept on the kinded paths.** `dam-vision.html` showed
  LiteLLM alone, tagged "required"; a user with only an Anthropic key would then be
  unable to create either kind.
- **Known gap, pre-existing and unchanged:** the create is not transactional. The
  marker is stamped by the agent create while the install event is enqueued after,
  so a failure between them leaves a marked agent whose setup never ran —
  indistinguishable in the lists from a healthy one, since no setup state is
  surfaced. Documented in both architecture pages; worth a follow-up issue.
- **KB keeps its greeting race.** Its onboarding artifact is a *command*, and the
  skills read reports only skills, so it has nothing equivalent to probe. Now that
  the claude-code image symlinks `.claude/skills` into the scanned path, a template
  whose bootstrap installs a skill could adopt the same gate.

## Review fixes already folded in

A review pass caught three things worth calling out, all fixed in `024da090`: the
setup probe polled every 3s forever on a stopped agent (now gated on running and
bounded); it read only the `standalone` bucket, so a *tracked* skill would have
wedged the greeting permanently (now reads both); and retiring the KB create page
lost its "harness image not installed" operator guard (restored on step 1).

## Docs

`experiments.md` gains an experiment-sandbox section; `knowledge-bases.md` records
that Kind is shared machinery and that creation moved into the wizard;
`agent-lifecycle.md` now describes the wizard and per-Kind dispatch it was being
linked to as the owner of; `skills.md` records Install-Command provenance for
Standalone skills plus the greeting gate as a consumer. Glossary: `Agent Kind`,
`Install Command` and `workspace-command` moved to the Agents context now that two
surfaces share them, and `Driver Agent` spells out the marked/unmarked split.

## Testing

`mise run check` and `mise run test` green (14 UI suites, 88 api-server). New unit
tests cover the install command's ordering and no-fetch property, the shared rail's
bump→enqueue→wake sequence, kind stripping for every kind, `toSandboxGroups`, and
the provider policy.

**Not yet verified on a cluster.** The install command copying the kit and the
greeting firing against a real pod are unexercised — the dev cluster was mid-wipe.
`03-agent.spec.ts` is the one e2e spec the wizard change touches (it now picks a
starting point before the mock image).


packages/api-server/src/modules/experiments/compose.ts
```diff
@@ -1,8 +1,14 @@
 import type { Db } from "db";
-import type { AgentsService, ExperimentsService } from "api-server-api";
+import type {
+  AgentsService,
+  ExperimentSandboxCreateInput,
+  ExperimentsService,
+} from "api-server-api";
+import { createKindedAgent } from "../agents/services/kinded-agent-create.js";
```
dam-code-guardian, line 7: 🟢 **Suggestion:** This imports the shared rail via a deep path into the `agents` module internals, bypassing `agents/index.ts`. Now that `createKindedAgent` is a deliberately-shared cross-module primitive, re-exporting it from `agents/index.ts` and importing it there would make the module boundary explicit. Consistent with existing repo conventions, so non-blocking.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `c67ff69d`

## PR #3014: feat: unify experiments and knowledge bases behind one Agent Kind rail and one wizard
**Author:** jezekra1 | **Branch:** feat/unify-experiments-knowledge-bases → main | **Changes:** +2083 −982 (56 files)

### Summary
Unifies experiments and knowledge bases onto one `Agent Kind` marker + a shared `createKindedAgent` install rail, folds all sandbox creation into one wizard with five starting points, lists every agent on the Sandboxes surface (badged by Kind), and groups the Experiments destination by sandbox. Includes a folded-in review pass (bounded greeting probe, restored missing-harness guard, StrictMode double-fire fix, provider policy on kinded paths).

### Findings
- 🟢 **Suggestion:** The shared rail is imported by two modules via a deep path into the `agents` module internals (`import { createKindedAgent } from "../agents/services/kinded-agent-create.js"`), bypassing `agents/index.ts`. Now that it's a deliberately-shared cross-module primitive, re-exporting it from `agents/index.ts` would make the boundary explicit. Consistent with a pervasive existing pattern in the repo, so non-blocking. (`packages/api-server/src/modules/experiments/compose.ts:7`, `packages/api-server/src/modules/knowledge-bases/services/knowledge-bases-service.ts:10`)
- ✅ **Looks good:** Contract stays flat — `experiments.createSandbox` validates + delegates only; the domain `install-command.ts` is pure; `compose.ts` replaces three non-null assertions with a narrowed destructured rail.
- ✅ **Looks good:** `agents.create` still strips `kind` for every enum member (test added), so no caller can mint a marked agent that skipped its Install Command.
- ✅ **Looks good:** Install command is `set -eu`, copies from the image (no network fetch), quotes all interpolations, and orders command-before-skill so the skill's presence implies both — covered by unit tests.
- ✅ **Looks good:** All dangling references to the deleted KB create view / `image-step` / `knowledge-base-new` route are cleaned up; `/knowledge-bases/new` redirects into the wizard.

✅ Aligned — no drift. `experiments.md`, `knowledge-bases.md`, `agent-lifecycle.md`, and `skills.md` all reflect the behavioral changes and are bumped to `Last verified: 2026-07-28`; no ADR leaks; commit 4 deliberately replaced leaked code symbols in prose with domain vocabulary, keeping identifiers only in "Where the code lives" link-outs.

🟢 One suggestion (module-boundary import path, above). Otherwise aligned: contract-package validation stays flat, domain layer pure, dependency injection intact, no `any`/unsafe `as` introduced. Cross-module deep import matches existing repo convention.

✅ No findings. Greeting logic extracted into a shared `useAgentGreeting` orchestrator with thin per-kind composers; the 479-line experiments view is split into focused, individually-testable components + a pure `toSandboxGroups`; server state stays in TanStack Query (`skipToken`, bounded poll); StrictMode double-fire and the `sessions-sidebar` empty-state flicker are both correctly fixed. New components follow `interface Props`, no `React.FC`, no static `style`.

### Verdict
APPROVE — Cohesive, well-tested unification; the single finding is a non-blocking style suggestion consistent with existing repo conventions.

---

Tomas2D (approved):

PR #3016, opened 2026-07-28
fix(channels): validate inbound Slack attachment bytes before prompting the agent

pilartomas:
Refs #3008.

## What actually broke, and what this is

#3008 reported every Slack image coming back as `Unable to resize image — image processing is unavailable and dimensions could not be read from the file header`, and proposed adding image processing to the agent environment.

The cause was neither that nor the redirect this PR originally blamed: **the install had never been granted `files:read`.** `files:read` only entered [`deploy/slack-app-manifest.yaml`](../blob/main/deploy/slack-app-manifest.yaml) on 2026-07-23, eight weeks after inbound images shipped, and an existing Slack app keeps the scopes it was installed with. Granting the scope fixed the reported failure. **So this PR is not the fix** — it is the guard that keeps the next such failure from being invisible, and the notice that would have identified this one in a single reply instead of weeks of guessing.

The message misleads in two directions, which is why it cost so much time:

- It is raised when the resize step throws **and** the byte header is not PNG/JPEG/GIF/WebP. It therefore says the bytes were not an image — never that a decoder is missing. Its "image processing is unavailable" wording is emitted unconditionally, whatever the classified cause.
- A refused download does not look like a failure from here. Slack answers a request it will not serve with a **200 and a sign-in page**, so the worker base64'd markup, labelled it `image/png` from Slack's metadata, and the harness substituted its own error for the picture. The agent then reported that error as its answer.

| sent as `image/png` | what the harness put on the wire |
| --- | --- |
| a real 3000×2000 PNG | `image/jpeg`, 2000×1333 — decoded and resized |
| an HTML page | `[Image could not be processed: Unable to resize image — image processing is unavailable …]` |

Image processing was present and working in the shipped agent image throughout. No image rebuild is needed and none is in this PR.

## Change

- `channels/inbound-image.ts` — classify downloaded bytes: a decodable image (PNG/JPEG/GIF/WebP), a page where a file should be, or a named unreadable format. A JPEG must carry its start-of-frame segment, not merely its signature, because that is where the harness reads dimensions from — a transfer cut off inside a long EXIF block otherwise passes here and fails there. A cut-off PNG or JPEG is named as the interrupted transfer it is, so the notice says resend rather than listing formats the file already was.
- **SVG is an unsupported format, not a failed download.** Slack labels `.svg` uploads `image/svg+xml`, so they reach classification, and an SVG's `<?xml` prologue otherwise reads as the sign-in page a scope problem returns — which would tell a sender to reinstall the app over a file that arrived perfectly.
- Only decodable images become prompt blocks, tagged with the **sniffed** type rather than Slack's `mimetype`, so a mislabelled upload still reads. Attachments whose label claims no format (absent, or a generic blob) are now sniffed too, instead of being dropped unread on the strength of a bad label; genuine documents are still left alone.
- Anything withheld is named **twice**: to the requester in a notice — including `files:read` when the scope probe confirms the install lacks it — and **to the agent in its prompt**, so it says it could not see the file instead of answering blind. The latter is how this surfaced in the first place: asked what a screenshot said, the agent replied that it could not process it and then guessed anyway.
- A `slack.image.unreadable` log line for operators. The turn still runs, so the question is answered even when the picture is not.
- **The permissions an install is missing are reported at startup**, each with the capability it costs and a note that an app already installed keeps the permissions it was installed with. This is the part that would have caught #3008 in minutes: the platform already probes the granted set to gate two agent tools, but nothing put that answer where an operator would see it. Declared in code rather than read from the manifest on purpose — the manifest describes a *fresh* install, so a permission added to it later never reaches an earlier workspace. An unknown or unreachable probe stays quiet rather than guessing at a gap.

## Notes

- **The redirect walk is gone.** It replaced five lines of `fetch` with a hand-rolled hop loop and a token-host allowlist, to defend a mechanism never observed occurring. (Node's `fetch` does strip `Authorization` across origins — verified — but if Slack ever redirects, byte classification now catches the symptom loudly, which is the part that matters.)
- The fake Slack gateway can serve file bytes (`setFileBytes`); previously `downloadFile` always threw, so no test had ever exercised a successful image fetch.
- `docs/architecture/channels.md` was at its character cap (39,996 of 40,000), which is what drove the earlier collateral rewording. Two invariants that trim had dropped are restored: the turn ceiling bounding a failed relay's lingering ref, and the Agent id carried in the footer link URL. The two new invariants are paid for by cutting volatile detail rather than by rewording — a scope enumeration duplicating the manifest (the same drift that hid this failure), an event's field-by-field payload, internal symbol names in the MCP-endpoint bullets, and two parentheticals restating the sentence they followed. No invariant was dropped to make room; each cut fact is either still stated nearby or lives in the code it described.
- Still open, as the issue notes: non-image files (PDF, CSV, docs) are dropped before reaching the agent, and Telegram has no inbound-file path at all. Neither is filed yet.

## Verification

- `mise run check` — clean except `controller:check:crd-backwards-compatibility`, which fails in any git worktree (`reference not found`; crdify cannot resolve refs through a worktree gitdir) and touches no CRDs here.
- `mise run test` — api-server 90 files / 756 tests green; agent-runtime, cli and controller suites green.
- New tests cover classification, the accept/withhold/notice behaviour, the prompt note, an SVG blaming the format rather than the permission, a generic-blob label still reading, a PDF still being left alone, and the startup report — both that it names the missing permission and its cost, and that it stays silent when the granted set is unknown.
- Not exercised against a live Slack workspace. The behaviour that matters was confirmed the other way: with `files:read` granted, images work.


packages/api-server/src/modules/channels/inbound-image.ts
```diff
@@ -0,0 +1,116 @@
...
+    .replace(/(..)/g, "$1 ")
+    .trim();
+  return `unrecognized data (it starts with ${hex})`;
+}
+
+/** Whether the bytes are text markup — HTML, XML/SVG or JSON. Slack answers a
+ *  file request it won't serve with a 200 and a page, so this is what a
+ *  permission problem looks like from here, not an HTTP error. */
+function looksLikeWebPage(bytes: Buffer): boolean {
+  const head = bytes.subarray(0, 512).toString("latin1").trimStart();
```
dam-code-guardian: 🟢 **Suggestion:** This sniffs only the first 512 bytes for a leading `<html>`/`<!doctype>`/`<?xml>`/`<svg>` tag or a JSON error object. A download that returns a plain-text error body (e.g. `You are not authorized` with no markup, or a differently-shaped JSON error) would fall through to `unreadable` and be described with a hex dump instead of the clearer "web page / possibly missing `files:read`" notice. Non-blocking — the current branches cover Slack's actual 200-with-page responses; worth revisiting only if other non-image bodies show up in the wild.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `8acad2a`

## PR #3016: fix(channels): validate inbound Slack image bytes before prompting the agent
**Author:** pilartomas | **Branch:** fix/slack-inbound-image-validation → main | **Changes:** +777 −15 (8 files)

### Summary
Fixes the bug where every Slack image came back as an "image processing is unavailable" error. Root cause was two-fold: `fetch` drops `Authorization` on cross-origin redirects so a redirected Slack download returned a 200 sign-in page (markup) that got base64'd as if it were the image; and formats no harness decodes (HEIC/BMP/TIFF) were forwarded blindly. The fix classifies downloaded bytes by content (not Slack's label), only forwards decodable PNG/JPEG/GIF/WebP tagged with the *sniffed* type, withholds everything else with a requester-only notice, and walks redirects by hand re-attaching the bot token only on Slack-owned HTTPS hops.

### Findings
- ✅ **Looks good:** `sendsBotToken` is a real security hardening — token is attached only on `https:` + `slack.com`/`*.slack.com`, with a test asserting a look-alike host (`notslack.com`) and plaintext (`http://`) are rejected. Uses hostname equality/`.slack.com` suffix, so `notslack.com` can't spoof it (`bolt-slack-gateway.ts:724`).
- ✅ **Looks good:** Node 24's undici honors `redirect: "manual"` (exposes the 3xx status and `Location` header), so the hand-rolled redirect walk works against the real runtime, not just the stubbed `fetch` (`bolt-slack-gateway.ts:757`).
- ✅ **Looks good:** Redirect loop is bounded (`MAX_DOWNLOAD_REDIRECTS`), missing `Location` and non-2xx both throw rather than returning error bytes, and relative redirect targets resolve via `new URL(location, url)` (`bolt-slack-gateway.ts:764`).
- ✅ **Looks good:** Sniffing enforces the harness's minimum header sizes (PNG ≥24, JPEG ≥10, etc.) so a truncated download is classified `unreadable` rather than passed as a headerless image that fails downstream (`inbound-image.ts:635`).
- ✅ **Looks good:** The turn still runs when the attachment is unusable (text question is answered), and the notice names the file, the reason, and `files:read` when the scope probe confirms it's missing — with a `slack.image.unreadable` operator log line.
- 🟢 **Suggestion:** `looksLikeWebPage` sniffs only the first 512 bytes for a leading `<html>`/`<!doctype>`/JSON-error marker. A download returning a plain-text error body (no markup) would fall through to `unreadable` with a hex dump rather than the clearer "web page / files:read" notice. Minor — the current branches cover Slack's actual responses (`inbound-image.ts:695`).

**Verdict:** aligned. The behavior change (undecodable attachments withheld, not forwarded) is recorded in the same PR as a new invariant bullet in `docs/architecture/channels.md`; `Last verified: 2026-07-28` matches the doc-editing commit, no ADR references or volatile content leaked in, and no new page or page-move is required.

✅ No findings. The pure classifier (`inbound-image.ts`) is correctly dependency-free and lives at module root, imported only by the infra adapter; classification results are modeled as an `InboundAttachment` discriminated union (error-as-value); redirect/token handling sits in the infrastructure adapter where external-system concerns belong.

### Verdict
**APPROVE** — Correct root-cause fix with strong test coverage (classification, accept/withhold/notice behavior, and download-hop token scoping) and a genuine security improvement in token handling; the one suggestion is non-blocking.

---

Issue #944, opened 2026-06-15
UI - Rethink how skills are presented and managed in the GUI


jenna-winkler:
## Background: how skills work today

A skill is a packaged capability defined in a git repo - a directory with a SKILL.md (name + description) plus supporting instructions/templates. A repo is registered as a skill source, scanned under skills/, .claude/skills/, or .agents/skills/, and each skill directory found becomes installable onto an agent.

Skills reach an agent through several paths, distinguished today only by a badge in the Skills panel:

- Platform (blue) - pre-installed by the platform team
- Agent (purple) - bundled into the coding agent image
- User (no badge) - added by you from any git repo source

On top of that, the current Skills panel is where a user:

- Toggles a skill on/off for an agent (install → downloaded to the workspace, picked up on the next session start)
- Manages sources - adds / refreshes / removes skill-source repos
- Publishes - opens a PR pushing a skill the agent authored back to a source repo
- (Should) track upstream drift - skills can change in their source repo after install

## Problem

The path a skill takes to an agent is inconsistent and hard to reason about - baked into the image, added to the workspace, or attached via a GitHub repo - with no unified story. Users can't easily discover what an agent can do, and figuring out what a given skill does usually means leaving the product, since skills are represented mostly by name.

This also leaves no clear home for strategically important skills like Bugstone - it's not apparent an agent even has it.

Everything above lives in one list-based Skills panel, so frequent actions (toggle a skill on/off) compete for space with infrequent admin actions (managing source repos, publishing, resolving upstream changes). 

## Proposed direction

1. Make skills more first-class and discoverable in the UI - so it's obvious what an agent can do and which shipped skills are available.
2. Redesign the Skills surface in agent config so each distinct job (use vs. administer vs. publish) has clear structure.

Make it easy to:

- See at a glance what's installed on an agent and what's available to add.
- Understand a skill in context (what it does) without leaving the product.
- Separate everyday from admin - toggling skills shouldn't be crowded by source-repo management or publishing.
- Surface outdated skills clearly - flag when an installed skill has drifted from its upstream source (today this is passive and easy to miss).

## Scope

- In: UX/presentation for existing skill functionality — discovering skills, installing/removing per agent, managing sources, surfacing upstream changes, publishing, and surfacing shipped skills.
- Out: How skills function or execute (the scanner, source roots, install mechanics, and the agent's own skill-management tools stay as-is).

jenna-winkler:
[...]

PR #3018, opened 2026-07-28
feat(skills): distinguish image-baked skills from user-authored ones

### LINKED ISSUE #2828: Distinguish system (image-baked) skills from user-authored skills
## Problem

Skills that ship baked into the sandbox image (e.g. scheduling, web search/fetch, and other built-ins) appear under "Created in this sandbox" as though the user authored them. This is misleading, clutters the user's own skills, and means the genuine "no skills yet" empty state is never reachable. A per-skill "managed" marker was tried during #944 and reverted — it wasn't viable because not all baked-in skills are ours to mark.

## Proposed solution

Establish **provenance** for skills so the surface can tell image-baked (system) skills apart from ones created at runtime by the user — for example a separate read-only location, a build-time manifest, or a first-boot snapshot — then hide or segregate the system skills from the user's own.

This is an **open design question**, not a settled approach — the three options above are starting points; pushback welcome.

### Dependencies

Deferred from #944. Blocks a truthful "Created in this sandbox" group and the empty state.

xjacka:
Closes #2828.

## Problem

Skills baked into the sandbox image reach the PVC via the first-boot seed (`cp -rn /app/working-dir/. $HOME/`) and were indistinguishable from skills the user authored at runtime, so the UI listed them under "Created in this sandbox", they were publishable as if the user owned them, and the true empty state was unreachable.

## Approach: read-time classification against the image's pristine copy

The PVC mounts at `$HOME` while the seed *source* (`/app/working-dir`) stays visible in the pod, immutable and exactly as built. That makes the image itself the provenance record — no build-time manifest to maintain, no per-skill marker (tried in #944 and reverted: third-party baked skills aren't ours to stamp), and no first-boot snapshot on an agent-writable volume.

`listLocal` now stamps each Local Skill with an `origin`, judged by directory name + content hash against the pristine copy:

- `system` — shipped by the image, untouched
- `system-modified` — shipped by the image, diverged on the PVC (an edit in the sandbox, or a template upgrade moving the image ahead)
- `user` — no pristine counterpart; created at runtime

Pristine roots are a set of exactly two sanctioned locations: the manifest's `skill-ref` paths re-expanded against `IMAGE_WORKSPACE_DIR` (default `/app/working-dir`) instead of `$HOME`, plus `STAGED_SKILLS_DIR` (`/usr/local/share/dam-skills`, the shared contract constant) for kind-specific skills an Install Command copies in post-create — no runtime-manifest change. Non-`$HOME` manifest paths are skipped (they'd re-expand to themselves and compare a skill against its own directory), and a pristine counterpart must carry a `SKILL.md`. Pristine hashes are memoized for the life of the process; local-side hashing is guarded, degrading an unhashable copy to `system-modified` instead of failing the listing; the verdict itself is a pure domain rule (`judgeOrigin`).

Because nothing is recorded anywhere, classification is **retroactive**: existing agents get correct verdicts the moment they run the new image, with no migration. `origin` is optional in both contracts, so a pod predating classification reports none and readers treat it as `user` — the pre-provenance behavior, safe for rolling upgrades.

## What changed

- **agent-runtime**: `listLocal` classification (repository adapter + pure `judgeOrigin` domain rule), pristine-path wiring in the composition root, `IMAGE_WORKSPACE_DIR` config.
- **contracts**: optional `origin` on `LocalSkill` in `agent-runtime-api` and `api-server-api`.
- **api-server**: publish gate — any image-shipped skill (`system` **or** `system-modified`) returns `FORBIDDEN`: divergence (a user edit, or an image upgrade) doesn't transfer ownership, so the gate can't be disarmed by appending a byte or by a routine image bump.
- **UI**: new read-only "Included with sandbox image" group (name and styling per design feedback); "Created in this sandbox" now shows only user-authored skills; `system-modified` rows get a `Badge variant="warning"` "Modified" pill.
- **docs**: `docs/architecture/skills.md` — new "Skill Origin" concept + read-time invariant.

No DB migration, no Helm/controller changes.

## Image-upgrade semantics (deliberate)

| Change in a new image | Existing agent | New agent |
|---|---|---|
| skill added | doesn't appear (first-boot seed runs once — pre-existing behavior) | `system` |
| skill removed | orphaned copy stays on the PVC and reclassifies as `user` (the user now owns it) | absent |
| skill changed | old copy on the PVC → `system-modified` with a "Modified" pill (still not publishable) | `system` (new version) |

## Verification

- 8 new unit tests for the classification (untouched / diverged / divergence outside SKILL.md / runtime-created / missing pristine root / dir-name identity / no-pristine-paths / mixed listing).
- Full suites green: agent-runtime 197, api-server 734, UI 108, CLI, controller.
- Verified live on a dev cluster (claude-code image): the four baked skills (`dam-invoke`, `platform-schedules`, `webfetch`, `websearch`) moved to the new group with no publish affordance; an agent-created skill stayed under "Created in this sandbox" with a working Publish button; editing a baked skill's `SKILL.md` on the PVC surfaced the "Modified" pill, and reverting the edit cleared it on the next poll.

---

**Review round 1** (thanks @PetrBulanek — all four findings confirmed and fixed in 567a36d1): staged-skills dir joined the pristine roots (fixes the #3014 authoring kit misclassifying as user-authored), local-side hashing guarded, publish gate hardened to refuse `system-modified`, and "Created in this sandbox" now renders an authoring-affordance empty state on a running agent instead of vanishing.


packages/agent-runtime/src/server.ts
```diff
@@ -84,8 +84,17 @@ const runtimeManifest = loadManifest(manifestPath);
...
 // lifetime of the process; createContext just hands them out per-request.
 const filesService = createFilesService(homeDir);
+// Pristine counterparts for origin classification. The filter drops
+// non-$HOME manifest paths — those re-expand to themselves and would
+// compare a skill against its own directory.
+const readSidePaths = skillRefPaths(runtimeManifest, homeDir);
+const pristineSkillPaths = skillRefPaths(
+  runtimeManifest,
+  config.IMAGE_WORKSPACE_DIR,
+).filter((p, i) => p !== readSidePaths[i]);
```
dam-code-guardian: 🟢 **Suggestion:** this positional filter assumes `skillRefPaths` yields index-aligned arrays for the two different `home` args (true today — both walk the same manifest order). A set-based check would drop the coupling and read more directly:

```suggestion
).filter((p) => !new Set(readSidePaths).has(p));
```

(or hoist the `Set` above the two `skillRefPaths` calls to avoid rebuilding it per element). Non-blocking.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `b2b8fb0d`

## PR #3018: feat(skills): distinguish image-baked skills from user-authored ones
**Author:** xjacka | **Branch:** feat/skill-origin → main | **Changes:** +372 −13 (16 files)

### Summary
Classifies each Local Skill's provenance (`system` / `system-modified` / `user`) at read time by content-hashing it against the image's pristine workspace copy, then uses that to segregate image-baked skills into a read-only "Included with sandbox image" UI group and to block publishing untouched system skills. No migration and no on-PVC marker — classification is retroactive and safe across rolling upgrades (optional `origin`, absent → treated as `user`).

### Findings
- 🟢 **Suggestion:** the pristine-path filter `.filter((p, i) => p !== readSidePaths[i])` couples the two arrays positionally; it's correct because both derive from the same manifest order via `skillRefPaths`, but a set-based `!readSideSet.has(p)` would drop the index assumption (`packages/agent-runtime/src/server.ts:94`).
- ✅ **Looks good:** pristine hashes memoized once per dir for the process lifetime (image is immutable), and `localHash` is computed lazily only when a pristine counterpart exists.
- ✅ **Looks good:** `judgeOrigin` is a pure domain rule; hashing/FS stay in the repository adapter; the service only orchestrates; the `origin` enum lives in the contract package — clean layering.
- ✅ **Looks good:** 8 focused unit tests covering untouched / diverged / divergence-outside-SKILL.md / runtime-created / missing pristine root / dir-name identity / no-pristine-paths / mixed listing.

Verdict: **aligned**. `docs/architecture/skills.md` is updated in the same PR (new "Skill Origin" concept + read-time invariant + publish-gate bullet), `Last verified:` bumped to the PR date. No ADR leaks, no volatile content added.

✅ No architecture findings. Respects the three-layer model: pure domain rule (`judgeOrigin`), FS/hashing in the repository adapter, orchestration-only service, validation in the contract package, and `pristineSkillPaths` a required dependency wired at the composition root.

✅ No findings. `BuiltInSkillsGroup` derives its groups at render from server state (no state duplication), extracts the map row into its own component, uses Tailwind + `cn()` with no static inline styles, and treats missing `origin` as user-authored to match the contract's pre-provenance fallback.

### Verdict
**APPROVE** — correct, well-tested, cleanly layered; docs updated in the same PR. One minor maintainability suggestion, non-blocking.

---

packages/agent-runtime/src/server.ts
```diff
@@ -84,8 +84,17 @@ const runtimeManifest = loadManifest(manifestPath);
 // Boot-time module composition. Skills + Files services are stable across the
 // lifetime of the process; createContext just hands them out per-request.
 const filesService = createFilesService(homeDir);
+// Pristine counterparts for origin classification. The filter drops
+// non-$HOME manifest paths — those re-expand to themselves and would
+// compare a skill against its own directory.
+const readSidePaths = skillRefPaths(runtimeManifest, homeDir);
+const pristineSkillPaths = skillRefPaths(
```
PetrBulanek: Blocking: the "image is the manifest" premise is already false on `main`.

#3014 stages the experiments authoring skill at `/usr/local/share/dam-skills/` —
deliberately outside `/app/working-dir` (`Dockerfile:66`, "so the first-boot
`cp -rn` never seeds it"), then the api-server copies it into
`$HOME/.agents/skills/` at create. After the rebase, `dam-experiment` has no
pristine counterpart → `origin: "user"` → it lands in "Created in this sandbox"
with a live Publish button. That's exactly the bug #2828 fixes, on the newest
feature.

Pristine roots need to become a list that includes the staging dirs — which does
reintroduce a hand-maintained list. Worth deciding deliberately here.

packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts
```diff
@@ -152,6 +176,52 @@ async function list(skillPaths: SkillPath[]): Promise<LocalSkill[]> {
...
+  localDir: string,
+  pristinePaths: SkillPath[],
+  pristineHashes: Map<string, Promise<string | null>>,
+): Promise<SkillOrigin> {
+  const pristineHash = await firstPristineHash(
+    dirName,
+    pristinePaths,
+    pristineHashes,
+  );
+  const localHash = pristineHash === null ? null : await hashSkillDir(localDir);
```
PetrBulanek: Blocking: this local hash is unguarded, so one bad file kills the whole listing.

The pristine side uses `hashSkillDirIfPresent`; this side calls `hashSkillDir`
directly. Verified on a dev cluster — `chmod 000` on a file inside `webfetch/`
makes `skills.state` return INTERNAL_SERVER_ERROR (`EACCES`), and the UI's poll
swallows it, so the user's own skills silently disappear with no error, every 5s.
Also a live race: I hit ENOENT by deleting a skill dir during a listing, which
uninstall can do concurrently with the poll.

`listLocal` was previously defensive at every step; it now reads every byte of
every file. Fix is to reuse the guarded helper — `judgeOrigin(null, pristineHash)`
already yields `system-modified`, a safe degradation instead of a thrown listing.

packages/api-server/src/modules/skills/services/publish-service.ts
```diff
@@ -65,6 +65,17 @@ export async function publishSkill(
     });
   }
 
+  // Untouched image-shipped skills aren't the user's to publish (#2828).
+  // `system-modified` passes — the divergence is the user's; missing origin
+  // (pre-provenance pod) stays publishable.
+  const local = await deps.runtimeClient.listLocal(input.agentId);
+  if (local.find((s) => s.name === input.name)?.origin === "system") {
```
PetrBulanek: This gate disarms itself on a routine image upgrade.

Verified live: changed one baked skill in the image, rebuilt, left both PVC copies
untouched. Same sandbox — `webfetch` (unchanged in image) → FORBIDDEN; `websearch`
(changed in image) → passed the gate and reached GitHub. So one upstream bump makes
a built-in publishable on every existing sandbox. `echo "" >> SKILL.md` does it too.

packages/ui/src/modules/sandboxes/components/skills/skills-surface.tsx
```diff
@@ -79,6 +81,14 @@ export function SkillsSurface({
 
   const publishableSources = sources.filter((s) => s.canPublish);
 
+  // Missing origin (pre-provenance agent image) counts as user-authored.
+  const createdHere = standalone.filter(
```
PetrBulanek, line 86: The empty state — one of the issue's three goals — isn't delivered. `isEmpty` (line
140) still counts built-ins. Every image is `FROM platform-base`, which bakes
`platform-schedules`, and no template overrides the seed `init` — so `standalone`
is never empty and the state stays unreachable.

Worse: when `createdHere` is empty on a *running* agent, the ternary below falls to
`readOnly && <Placeholder />` → `false`, so the section renders nothing — no header,
no affordance, no hint you can author a skill.

PetrBulanek (changes requested):

packages/agent-runtime/src/server.ts
xjacka: Good catch — confirmed against `main`: the #3014 authoring kit is deliberately staged outside the seed and lands on the PVC via the Install Command, so under this PR it classified as user-authored. The premise as written ("the seeded workspace is the manifest") was too narrow.

Fixed in 567a36d1, and taking the design point head-on: pristine roots are now a **set of exactly two sanctioned locations**, not a growing list —

1. the seeded workspace copy (`IMAGE_WORKSPACE_DIR`), and
2. the staged-skills dir for kind-specific skills that must not reach every sandbox.

The staged dir is a single shared constant (`STAGED_SKILLS_DIR` in `agent-runtime-api`), now also consumed by the experiments Install Command instead of its private copy — so the two sides can't drift. `docs/architecture/skills.md` documents the convention explicitly: an image-shipped skill anywhere else *will* misclassify as user-authored, so new features must ship through one of these two locations.

So yes, it is a hand-maintained list — but of **platform-owned staging conventions (currently two), named once in the contract package**, not of skills and not of per-feature directories. If a future Agent Kind needs another staging location, the intended move is to put it under `STAGED_SKILLS_DIR` (subdirs are fine — counterpart lookup requires a `SKILL.md`, so non-skill dirs like the kit's `commands/` don't count), not to add a third root. If that convention proves too tight we can revisit, but I'd rather start strict.

Also addressed in the same commit: local-side hashing is now guarded (degrades to `system-modified` instead of throwing the listing), the gate refuses `system-modified` too, and the empty state renders on a running agent.

packages/api-server/src/modules/skills/services/publish-service.ts
```diff
@@ -65,6 +65,20 @@ export async function publishSkill(
     });
   }
 
+  // Image-shipped skills aren't the user's to publish (#2828) — modified or
+  // not: divergence (a user edit, or an image upgrade moving the baked copy)
+  // doesn't transfer ownership, so the gate can't be disarmed by appending a
+  // byte or by a routine image bump. Missing origin (pre-provenance pod)
+  // stays publishable.
+  const local = await deps.runtimeClient.listLocal(input.agentId);
```
PetrBulanek: The stricter gate introduces a false positive for skills installed from a Skill Source.

This lookup uses the raw `listLocal`, which includes skills tracked as installed-from-source. Installing a skill overwrites its directory, so its content diverges from the image's copy and it classifies `system-modified` — which the gate now refuses.

The realistic case is `websearch`: the baked copy is vendored from a public repo (`Djarvur/cc-websearch`), so a user who adds a source containing `websearch`, installs it, improves it and tries to publish back gets *"included with this sandbox's image and can't be published"*. At that point it isn't the image's skill, it's the source's — and this flow worked before this commit. The same applies to any source skill whose directory name collides with a baked or staged one.

Verified the mechanism on a dev cluster: a `dam-experiment` copy whose content differs entirely from the staged one is still refused.

There's a narrower second variant: authoring a skill named `dam-experiment` in a non-experiment sandbox hits the same refusal. The seeded names (`websearch`, `webfetch`, `platform-schedules`, `dam-invoke`) are already unauthorable — `write-local.ts` collides on the existing directory — so `dam-experiment` is the only name exposed, because it's staged rather than seeded and there's nothing on the PVC to collide with.

Not blocking, and possibly acceptable. But it's a behaviour change relative to before this commit rather than something #2828 asked for, so it deserves a deliberate call rather than passing unnoticed.

PetrBulanek (approved):

packages/api-server/src/modules/skills/services/publish-service.ts
xjacka: Good catch — the gate keyed on image collision alone and ignored the tracked provenance the platform already has. Fixed in 5c28ac0b, making the call explicit:

- **Installed Skill Refs are exempt.** A skill tracked in `agent_skills` is governed by its Source relationship, not by the image — install always overwrites the directory, so image-divergence is meaningless for it. Your `websearch`-from-source flow (install → improve → publish back) works again. This also matches what the UI already shows: the reconciled `state` puts tracked skills under their source card, never in "Included with sandbox image" — the gate now agrees with that view instead of the raw listing.
- **The narrow second variant stays refused, deliberately.** A fresh *standalone* authored under an image-owned name (today only `dam-experiment`, since seeded names collide in `write-local`) keeps hitting the gate. The name belongs to the image on that sandbox, the UI files the skill under "Included with sandbox image" with a Modified pill, and letting it publish would reopen the append-a-byte bypass for staged skills. Renaming the skill is the escape hatch. Noted in the gate comment + `skills.md`.

Unit tests cover both: tracked `system-modified` passes; standalone image-name skill still refuses even when other skills are tracked.

PR #3031, opened 2026-07-28
chore(agents): drop the nonessential-traffic umbrella flag, keep the explicit ones

jezekra1:
`CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC` disables more than the phone-home traffic #3028 meant to cut — it takes wanted behavior with it. This drops the umbrella and keeps the explicit constituents, which cover the original intent exactly:

- `DISABLE_TELEMETRY` — no Statsig usage telemetry
- `DISABLE_ERROR_REPORTING` — no Datadog error intake
- `DISABLE_FEEDBACK_COMMAND` / `DISABLE_FEEDBACK_SURVEY` — no feedback prompts
- `DISABLE_AUTOUPDATER` — intentional; the image pins the harness version

Platform OTEL export is a separate opt-in path and is unaffected.

Also formats `page-empty-state.tsx`, which landed on main unformatted and was failing the pre-commit format hook for every commit.

Amends #3028.

Issue #2999, opened 2026-07-27
UI - Update Image step of sandbox creation


jenna-winkler:
## Description

"Choose your starting point" step should display 5 clickable cards:
1. Experiment sandbox
2. Knowledge base sandbox
3. Specialized sandbox
4. General-purpose sandbox
5. Custom image

## Desired state

New description beneath header
> Every sandbox starts from an image. Choose one configured for a specific job, or start with a general-purpose image and customize it yourself.

<img width="1512" height="799" alt="Image" src="https://github.com/user-attachments/assets/39a70ccf-0a1b-4e92-b6d4-f2d32247bf09" />

### Experiment sandbox chosen

Clicking should under the hood select Claude Code and apply all the experiment presets (litellm, size, etc per @jezekra1 )
-> If the user has not yet added their LiteLLM key to the platform, step 2 of this creation flow (if Experiment is selected) should make it clear LiteLLM key is required

<img width="1512" height="795" alt="Image" src="https://github.com/user-attachments/assets/8c301bf3-be92-4e56-a925-d58e92abe4a9" />

### Knowledge base sandbox chosen

Clicking should under the hood select Claude Code and apply all the KB presets (litellm, size, etc per @Tomas2D ) and additionally reveal 2 templates to choose from: 
-> If the user has not yet added their LiteLLM key to the platform, step 2 of this creation flow (if Knowledge base is selected) should make it clear LiteLLM key is required

<img width="1512" height="797" alt="Image" src="https://github.com/user-attachments/assets/281f9c81-4cc0-412b-b752-78a6f9782a3a" />

### Specialized sandbox chosen

Selecting this should reveal all the images we currently have grouped in the GUI under pre-configured images
_Out of scope: Any changes to these images names/descriptions/icons_

<img width="1509" height="802" alt="Image" src="https://github.com/user-attachments/assets/ac0452af-f049-48f5-b2b7-9ed3e64789a5" />

### General-purpose sandbox chosen

Selecting this should reveal just the 4 generic harnesses: Claude Code, Bob, Codex Pi 
_Out of scope: Any changes to these images names/descriptions/icons_

<img width="1512" height="798" alt="Image" src="https://github.com/user-attachments/assets/70b11c7b-e4c6-4a70-a433-63df7a7ab0c5" />

### Custom image chosen

👇 Related issue #2748 

[...]

Issue #2946, opened 2026-07-23
Knowledge Bases phase 1: create a knowledge base and work with it through chat


Tomas2D:
## Problem

Teams keep building their own knowledge stores ad hoc — each one wired differently, each reinventing how knowledge gets ingested, organized, and queried. DAM has no first-class place to create and own a knowledge base: a user can approximate one today by manually configuring an agent, writing instructions, and connecting a repository, but nothing guides them there, and none of that setup is reusable by the next person.

## Proposed solution

Introduce the **Knowledge Base** as a first-class resource alongside Agents and Experiments, built on the abstraction agreed in #2796:

> A Knowledge Base is a DAM-managed workspace where an agent can operate on instructions, knowledge, and resources. GitHub, MCP, and other integrations are optional ways to extend that workspace.

User-visible behavior:

- **A Knowledge Bases section** in the UI listing the user's knowledge bases, with two entry paths: create a new one, or build one from existing material (a repository, files, or other resources).
- **Setup is guided in chat.** Opening a knowledge base lands in a chat where the agent walks the user through setup — what the knowledge base is for, feeding it sources, and organizing the content. There is no separate configuration wizard; the interactive chat *is* the experience, the same bet Experiments made on prompt-driven behavior over bespoke platform machinery.
- **Ask and act.** Once seeded, the user can ask questions or delegate tasks against the knowledge in the same chat. Chats are multi-session, like any other agent.
- **Opinionated default, overridable.** Every new knowledge base is backed by a Claude Code-based agent pre-configured with custom setup instructions for that concrete knowledge base — its name, description, and purpose — so the agent knows its role from the first message. Users with a reason can pick a different harness.
- **Templates carry the workflows.** Common kinds of knowledge bases ship as templates providing default instructions and skills — opinionated starting points, never a rigid schema.
- **GitHub is optional.** The knowledge lives in the DAM-managed workspace; syncing it to a repository the user owns is an available integration (guided by the agent in chat), not a requirement.

### Scope

Creating, seeding, organizing, and querying a knowledge base through chat, plus the list UI and the default template(s). Updates to the knowledge happen when the user drives them in chat; proactive maintenance can use the existing scheduling feature.

### Dependencies

Phase 1 of epic #2796.

### Out of scope

- **Sharing / MCP endpoint exposure** — deferred to phase 2 (#2796); needs a security design and its own permission model.
- **Attaching a knowledge base to sandboxes or experiments** — phase 3; depends on the agent-to-agent communication design.

PR #3032, opened 2026-07-28
fix(experiments): close the review gaps — status colors, kinded subtitles, compensated create

### LINKED ISSUE #2997: UI - Experiments page
All sandboxes that had "experiment sandox" selected during agent creation should know here. They should also appear in Home with the "Experiment" tag.

### Current state

Image

### Desired state

Image

--
When you click on "Create experiment" --> user should land on step 1 of sandbox creation flow with experiment and all its presets set:

Image

--
👆 Related issue: #2999 

### Current state

Image

### Desired state

Full prototype: https://ibm.ent.box.com/file/2358564436704 (go to experiments page and toggle state to full)

- [ ] Updated description beneath experiments header: Experiments are grouped by the sandbox running them. Open a sandbox to work with it in chat, where the experiment graph docks beside the conversation.
- [ ] Group experiments by sandbox

Behavior:
- Clicking anywhere on the sandbox header should open sandbox config
- Clicking on an experiment card should expand the card to show runs
- Clicking on the kebab for each experiment should give you the option to delete experiment (this is current GUI behavior; no change)
- Make sure the colors used for the Running / Complete tags match what is used consistently throughout GUI (Green = Running, Completed can be blue)

Image

**Hover state:**
Image

Empty states or when an experiment sandbox is created but no experiments exist yet:
Image

### LINKED ISSUE #3000: UI - Group sub-sandboxes created from Experiment
## Problem

Experiment sandboxes can spawn temporary child sandboxes to run experiment invocations. Currently, these spawned sandboxes appear as top-level sandboxes alongside user-created sandboxes, which makes the hierarchy unclear.

Users may not realize that these sandboxes are:

* Created automatically by an experiment
* Temporary and automatically cleaned up
* Using inherited configuration from the parent experiment sandbox
* Contributing to CPU/memory usage

Because users do not need to directly manage these child sandboxes, exposing them as equal peers creates unnecessary confusion.

## Current state

Spawned experiment sandboxes appear at the same hierarchy level as manually created sandboxes.

Image

## Desired state

Group temporary sandbox count and resource usage beneath their parent experiment sandbox.

Do not allow users to click into or manage these spawned sandboxes directly, since they inherit configuration from the parent experiment sandbox and are managed automatically by the experiment lifecycle.

Image

## Acceptance criteria

* Temporary sandboxes created by an experiment should not appear as top-level peers of user-created sandboxes.
* Users should be able to understand that these sandboxes are temporary and managed by the experiment lifecycle.
* The UI should avoid implying that users can configure or manage these child sandboxes independently.
* CPU/memory usage should account for temporary experiment sandboxes and provide enough context that users understand the source of usage.
* Use clearer terminology than "invocations" if exposed in the UI (for example, "temporary sandboxes").

### LINKED ISSUE #3001: UI - Update Knowledge base + Experiment cards in Home
## Experiment cards

- [ ] Blue Experiments tag just to the right of the sandbox name
     - Note: this Experiment tag should **_always_** be visible (current logic in the GUI seems to be only when an experiment is active). The user action that applies this tag is when they select "Experiment" in the sandbox creation form (related issue: #2999 )
- [ ] Description should be # experiments • # connections
     - No active experiments
     - 1 experiment
     - 2 experiments
     - etc
And for connections...
     - No text if no connections
     - 1 connection
     - 2 connections
     - etc

Image

## Knowledge base cards

- [ ] Purple Knowledge base tag just to the right of the sandbox name
- Note: this Knowledge base tag should appear when the user selects "Knowledge base" in the sandbox creation form (related issue: #2999 )
- [ ] Description should be Template name • # connections
     - No text if no connections
     - 1 connection
     - 2 connections
     - etc

Image

## out of scope

changing those status tags/names/colors

jezekra1:
Closes the three code-closable gaps from the #2328 status audit:

**Status colors match the design (#2997).** Running = green, Completed = blue — they were inverted in `ExperimentStatusBadge`. The invocation dots now agree in both surfaces (run rows + chat dock): live pulses green, done goes quiet grey; failed red and parked amber keep their meaning.

**Kinded card subtitles (#3001).** A kinded sandbox's row leads with what it is *for*:
- experiment: `N experiments · M connections` (`No active experiments` while empty)
- knowledge base: `Template · M connections`
- plain: `harness · provider`, unchanged

The connections segment counts catalog connections only and is omitted at zero. Experiment counts ride the driver-summaries query the Experiments page already caches; the KB list drops its hand-prepended template segment now that the shared subtitle owns it.

**Kinded create is compensated (the #2946 gap).** A failed install-event enqueue now deletes the fresh agent and rethrows, so a marked sandbox whose setup never ran can only arise if the compensating delete itself fails. A failed wake deliberately does not compensate — the event is durable and delivers on the next wake. Covered by a unit test; `knowledge-bases.md` updated accordingly.

Not addressed here, by design: the #2997 click-behavior divergence (rows open the sandbox chat per Radek's direction) and #2999 (needs a design pass against the shipped wizard).

Also formats `page-empty-state.tsx` (unformatted on main, blocks every commit through the pre-commit hook — same drive-by as #3031; whichever merges first wins).

Closes #2997. Closes #3000. Closes #3001.

### DIFF
--- a/packages/ui/src/modules/agents/hooks/use-agent-rows.ts
+++ b/packages/ui/src/modules/agents/hooks/use-agent-rows.ts
@@ -3,6 +3,7 @@ import { useMemo } from "react";
 import { useStore } from "../../../store.js";
 import type { AgentView, TemplateView } from "../../../types.js";
 import { useAppConnections } from "../../connections/api/queries.js";
+import { useDriverSummaries } from "../../experiments/api/queries.js";
 import { useTemplates } from "../../templates/api/queries.js";
 import { useDeleteAgent } from "../api/mutations.js";
 import { useAgents } from "../api/queries.js";
@@ -31,6 +32,9 @@ export function useAgentRows() {
   const templates = templatesData ?? NO_TEMPLATES;
   const { data: agentsData } = useAgents();
   const connections = useAppConnections();
+  // Experiment-kinded rows lead their subtitle with how many named experiments
+  // they drive (#3001); one cached query, shared with the Experiments page.
+  const { data: driverSummaries } = useDriverSummaries();
   const restartingAgents = useStore((s) => s.restartingAgents);
   useSyncRestartingAgents();
   const pausingAgents = useStore((s) => s.pausingAgents);
@@ -50,6 +54,16 @@ export function useAgentRows() {
     [pausingAgents],
   );
 
+  const experimentCountByDriver = useMemo(() => {
+    if (!driverSummaries) return undefined;
+    return new Map(
+      driverSummaries.map((summary) => [
+        summary.driverAgentId,
+        new Set(summary.experiments.map((e) => e.name)).size,
+      ]),
+    );
+  }, [driverSummaries]);
+
   const subtitleLookup = useMemo<SandboxSubtitleLookup>(
     () => ({
       templateNameById: new Map(templates.map((t) => [t.id, t.name])),
@@ -63,7 +77,13 @@ export function useAgentRows() {
   const rowProps = (agent: AgentView) => ({
     agent,
     display: resolveAgentDisplay(agent, restartingIds, pausingIds),
-    subtitle: sandboxSubtitle(agent, subtitleLookup),
+    subtitle: sandboxSubtitle(agent, subtitleLookup, {
+      // 0, not undefined, once summaries have loaded: a marked sandbox with no
+      // registered experiments reads "No active experiments".
+      experimentCount: experimentCountByDriver
+        ? (experimentCountByDriver.get(agent.id) ?? 0)
+        : undefined,
+    }),
     deletePending:
       deleteAgent.isPending && deleteAgent.variables?.id === agent.id,
     onWake: () => wakeAgent.wake(agent.id),
--- a/packages/ui/src/modules/agents/hooks/use-agent-rows.ts
+++ b/packages/ui/src/modules/agents/hooks/use-agent-rows.ts
@@ -34,7 +34,7 @@ export function useAgentRows() {
   const connections = useAppConnections();
   // Experiment-kinded rows lead their subtitle with how many named experiments
   // they drive (#3001); one cached query, shared with the Experiments page.
-  const { data: driverSummaries } = useDriverSummaries();
+  const { data: driverSummaries } = useDriverSummaries({ silent: true });
   const restartingAgents = useStore((s) => s.restartingAgents);
   useSyncRestartingAgents();
   const pausingAgents = useStore((s) => s.pausingAgents);
--- a/docs/architecture/knowledge-bases.md
+++ b/docs/architecture/knowledge-bases.md
@@ -27,7 +27,9 @@ Knowledge Bases is a feature-gated destination ([features](features.md)) with a 
 
 Opening a KB that has no sessions yet **greets the user**: the UI runs `/wiki-onboard` as a hidden first turn (reaches the agent, renders no user bubble, fails silently), so a fresh KB opens with the agent introducing itself rather than an empty chat. This is why every template's bootstrap installs that command. The greeting mechanism is shared with experiments.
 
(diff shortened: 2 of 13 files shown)


packages/ui/src/modules/agents/hooks/use-agent-rows.ts
```diff
@@ -31,6 +32,9 @@ export function useAgentRows() {
   const templates = templatesData ?? NO_TEMPLATES;
   const { data: agentsData } = useAgents();
   const connections = useAppConnections();
+  // Experiment-kinded rows lead their subtitle with how many named experiments
+  // they drive (#3001); one cached query, shared with the Experiments page.
```
dam-code-guardian, line 36: 🟢 **Suggestion:** `useDriverSummaries()` here runs `staleTime: 0` + `refetchOnMount: "always"` unconditionally, but this hook is shared with the Knowledge Bases list view — whose rows never read `experimentCount`. That surface fires an extra `driverSummaries` round-trip per mount for discarded data. Consider gating the query/count-map to where an experiment-kinded row can consume it.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `54288ae`

## PR #3032: fix(experiments): close the review gaps — status colors, kinded subtitles, compensated create
**Author:** jezekra1 | **Branch:** fix/experiments-issue-gaps#locki-njtatv6h → main | **Changes:** +138 −38 (12 files)

### Summary
Closes three code-closable gaps from the #2328 status audit: inverted running/completed status colors in the experiment badge and invocation dots (running→green, completed/idle→blue/grey), kind-shaped sandbox row subtitles (experiment counts / KB template), and a compensated kinded-agent create that deletes the fresh agent and rethrows when the install-event enqueue fails.

### Findings
- 🟢 **Suggestion:** `useAgentRows` now unconditionally calls `useDriverSummaries()` (`staleTime: 0`, `refetchOnMount: "always"`), and the hook is shared by the Knowledge Bases list view, whose rows take the `knowledge-base` subtitle branch and never read `experimentCount`. The KB page therefore fires a guaranteed extra `driverSummaries` fetch per mount for data it discards. Consider gating the query (or the count map) so it only runs where an experiment-kinded row can consume it. (`packages/ui/src/modules/agents/hooks/use-agent-rows.ts:36`)
- ✅ **Looks good:** The create compensation is correct and precise — the `try` wraps exactly the bump+enqueue, the swallowed `.catch(() => {})` on the delete leaves the original error as the thrown one, and the deliberate no-compensation-on-wake is documented and justified (durable event redelivers).
- ✅ **Looks good:** `Pick<AgentsService, "create" | "delete">` widened at both call sites (kinded + KB service); no over-broad service surface.
- ✅ **Looks good:** `sandboxSubtitle` now owns the KB template segment, letting `knowledge-bases-list-view.tsx` drop its prepend hack — genuine DRY win, single join path preserved.
- ✅ **Looks good:** `experimentCount` uses `0`-vs-`undefined` deliberately (loading → no segment; loaded-empty → "No active experiments"), matching the documented intent.
- ✅ **Looks good:** New behavior (compensated create) documented in the same PR at `docs/architecture/knowledge-bases.md` with `Last verified:` bumped, and covered by a unit test.

✅ Verdict: **aligned**. `docs/architecture/knowledge-bases.md` updated in-PR to replace the stale "Not transactional" gap with the compensated-create description matching the code; `Last verified:` at the PR date. No ADR leaks, no volatile content, no missing subsystem page, no cross-reference rot.

✅ No findings. Service files follow the DI/service-factory pattern; the `AgentsService` `Pick` widening is correct at both call sites. The two new cross-module UI imports (`useDriverSummaries`, `kbTemplateName`) match the package's pervasive established convention (no `index.ts` module boundaries; 193 such deep imports on main) — not introduced or amplified here, out of scope.

✅ No findings. Color changes are string-only edits inside existing `cn()` conditionals (no static `style`); props typed via `interface Props` / inline (no `React.FC`); server state stays in TanStack Query; `page-empty-state.tsx` is a pure Prettier reflow.

### Verdict
APPROVE — Focused, well-tested gap-closure; the one 🟢 (redundant KB-page query) is a minor efficiency nit, not a blocker.

---

Tomas2D (approved):

PR #3039, opened 2026-07-29
feat(skills): plan-feature-v2

tomkis:
Reworked version of the plan-feature skill. It replaces v1: the skill is renamed back to plan-feature and the old one is removed.

Main changes from v1:

- The grill gate is gone. The skill now runs as a follow-up to an existing conversation: it pulls the GH issue first (if not in context), then assesses whether the context is enough to decompose. If context falls short it stops and tells the user what's missing instead of gathering it itself.
- Decomposition rules rewritten around context windows. Primary rule is sizing each sub-issue to one context window. Vertical slices preferred when the feature splits into separate behaviors, horizontal split along the tRPC contract allowed when a single behavior is too deep for one window.
- Slices no longer need to be independently shippable, only green and concretely verifiable, since the whole feature lands as one PR.
- The plan is committed as the feature branch's first commit and a draft PR is opened. A new required Plan check (PRs targeting main) blocks merging while docs/plan/ exists.
- implement-feature updated to match: it picks up the branch and draft PR from plan-feature, the cleanup commit is mandatory, and it ends by marking the PR ready.
- The leftover docs/plan/3041-invocation-spend-attribution/ folder from #3041 is removed so this PR passes its own check.


.github/workflows/plan-check.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
+      - name: Fail if docs/plan exists
+        run: |
+          set -euo pipefail
+          if git ls-files -- docs/plan | grep -q .; then
```
dam-code-guardian: 🔴 **Critical:** This gate can silently pass while plan files are present.

`grep -q` exits as soon as it matches, closing the pipe. `git ls-files` then dies of SIGPIPE, and because `pipefail` propagates the rightmost non-zero status, the pipeline yields **141**, not 0 — so the `if` takes the false branch, prints "No plan files present.", and the job goes green with `docs/plan/` tracked. Whether it happens depends on whether `git` is still writing when the pipe closes, so it is racy — the worst property for the only check keeping plans out of `main`.

Drop the pipe so the test is on a plain string:

```suggestion
          files=$(git ls-files -- docs/plan)
          if [ -n "$files" ]; then
```

(with the body's second `git ls-files -- docs/plan` on line 28 replaced by `printf '%s\n' "$files"`).

.agents/skills/plan-feature-v2/SKILL.md
```diff
@@ -0,0 +1,202 @@
+---
+name: plan-feature-v2
+description: >
+  Turn a GitHub issue into an implementation plan under docs/plan/<feature>/: a feature spec
+  decomposed into self-contained, independently-shippable sub-issues. Use when the user wants to
```
dam-code-guardian: 🔴 **Critical:** This `description` is byte-identical to `plan-feature` v1's (302 chars — verified equal), and the `63d5926` revert keeps v1 installed alongside v2.

Dispatch is description-driven, so "plan a feature" now matches two skills that a reader (and the model) cannot tell apart, whose contracts directly contradict each other: v1 says *"never `git add` or commit them"*, v2 commits the plan as the branch's first commit. A wrong pick silently produces the wrong git state.

Also 🟡: this text still promises "independently-shippable" sub-issues, which line 57 explicitly retracts ("Slices don't need to be independently shippable"). Both halves of the fix live in this one field — give v2 a distinguishing trigger and drop the retracted promise.

.agents/skills/plan-feature-v2/SKILL.md
```diff
@@ -0,0 +1,202 @@
...
+<what-to-do>
+
+This skill turns a GitHub issue into a feature spec and implementation plan: a `README.md`
+plus one Markdown file per sub-issue, written under `docs/plan/<feature-slug>/`.
+
+**The plan is committed and starts the feature branch.** The plan files are working artifacts
+for the implementation phase, not permanent docs — but they live in git: planning ends by
+creating the feature branch, committing the plan as its **first commit**, and opening a **draft
+PR** (step 4). The plan never merges: the `Plan check` CI job fails while `docs/plan/` exists,
+so the PR stays blocked until the cleanup commit deletes the folder.
```
dam-code-guardian: 🟡 **Warning:** Nothing guarantees the cleanup commit this sentence depends on.

`implement-feature` (kept pointing at v1 by revert `da42a77`) treats the plan as untracked/"ephemeral", creates the feature branch itself in its step 4, and *asks the user whether to delete the folder* as an optional last step. Under v2 the branch and plan commit already exist, so that handoff is ambiguous — and with this gate in place, a forgotten optional cleanup no longer just leaves stray files, it leaves the PR permanently unmergeable.

Worth stating which consumer completes v2's flow, or teaching `implement-feature` a v2-aware mandatory cleanup commit.

.agents/skills/plan-feature-v2/SKILL.md
```diff
@@ -0,0 +1,202 @@
...
+
+Closes #NNN
+
+## Sub-issues
+
+- [ ] 01 — <title>
+- [ ] 02 — <title>
+```
+
+### 5. Report
```
dam-code-guardian: 🟡 **Warning:** `### 5. Report` sits outside `<what-to-do>` — it and the PR-body template landed inside `<supporting-info>` (which opens on line 102), while steps 1–4 are within `<what-to-do>` (closed on line 100).

Step 5 is a procedural instruction, so from inside the reference section it reads as material to consult rather than a step to perform, and may be skipped. Move step 5 (and the PR-body template it references) back above `</what-to-do>`.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `63d5926`

### Summary
Adds a reworked `plan-feature-v2` agent skill (drops v1's grill gate, resizes decomposition around context windows, and commits the plan as the feature branch's first commit behind a draft PR) plus a `Plan check` CI workflow that fails while `docs/plan/` is tracked, so a plan can never reach `main`. Two reverts in the branch deliberately keep v1 in place alongside v2.

### Findings

- 🔴 **Critical:** The CI gate can silently pass when it should fail. `set -euo pipefail` + `git ls-files -- docs/plan | grep -q .` — `grep -q` exits immediately on its first match, closing the pipe; `git ls-files` then dies of SIGPIPE and, under `pipefail`, the pipeline's status becomes 141 rather than 0, so the `if` takes the false branch and the job prints "No plan files present." and passes with plan files present. I reproduced this: a large producer into `grep -q .` yields `CONDITION FALSE, rc=141`, while a one-line producer takes the true branch. It is racy by nature (it depends on whether the writer is still writing when the pipe closes), which is the worst failure mode for a gate that is meant to be the only thing keeping plans out of `main`. See the inline comment for a pipe-free replacement. (`.github/workflows/plan-check.yml:26`)
- 🔴 **Critical:** `plan-feature-v2`'s frontmatter `description` is **byte-identical** to v1's (302 chars, verified equal), and the branch's reverts keep both skills installed side by side. Skill dispatch is description-driven, so "plan a feature" / "decompose this issue" now matches two skills with indistinguishable descriptions and materially different contracts — v1 says *never `git add` or commit the plan*, v2 commits it as the branch's first commit. The user cannot tell which one ran, and the wrong pick silently produces the wrong git state. If both must coexist, v2's description has to state its distinguishing trigger (and ideally v1's marked legacy). (`.agents/skills/plan-feature-v2/SKILL.md:5`)
- 🟡 **Warning:** v2's description still promises "**independently-shippable** sub-issues", which the body explicitly retracts: "Slices don't need to be independently shippable — the feature lands as one PR". The description is the dispatch surface and the first thing a reader sees, so this is the stale half of the rework. (`.agents/skills/plan-feature-v2/SKILL.md:5`)
- 🟡 **Warning:** The two consumers disagree about the plan's git lifecycle. `implement-feature` — which the `da42a77` revert deliberately re-pointed at v1 — describes the plan as "ephemeral", ends by *asking the user whether to delete the folder*, and creates the feature branch itself in step 4. Under v2 the branch and the plan commit already exist, so the handoff is ambiguous: nothing guarantees the cleanup commit the CI gate waits for is ever made, and the gate turns a forgotten optional cleanup into a permanently blocked PR. Either `implement-feature` needs a v2-aware cleanup step, or v2 should state which consumer completes its flow. (`.agents/skills/plan-feature-v2/SKILL.md:18`)
- 🟡 **Warning:** `### 5. Report` and the PR-body template sit **after** `</what-to-do>`, inside `<supporting-info>` (lines 100–121). Steps 1–4 are inside `<what-to-do>`; step 5 is a procedural step that has fallen outside the section the agent reads as its instructions, so the report step may be treated as reference material and skipped. (`.agents/skills/plan-feature-v2/SKILL.md:116`)
- 🟢 **Suggestion:** The workflow has no `concurrency` group, so superseded pushes keep running; `cd.yml` sets one. Cheap here (5-min timeout), but it is the repo convention. (`.github/workflows/plan-check.yml:14`)
- 🟢 **Suggestion:** `SKILL.md` has no trailing newline ("\ No newline at end of file"); the neighbouring `plan-feature` and `implement-feature` skills both end with one. (`.agents/skills/plan-feature-v2/SKILL.md:201`)
- ✅ **Looks good:** `actions/checkout` is SHA-pinned with a version comment matching `cd.yml`, `permissions: contents: read` is minimal, `persist-credentials: false` is set, and `timeout-minutes` is bounded — the workflow's security hygiene is right.
- ✅ **Looks good:** Using a red-until-cleanup required check is a sound way to make a committed working artifact un-mergeable, and the rationale is documented in a header comment where the next reader will find it.

✅ No findings. The diff adds two repo-local agent skills and a CI workflow; no subsystem's behavior or responsibility changes, and no page under `docs/architecture/` (nor the landing page) is touched. `docs/architecture/skills.md` covers the platform's connectable git-based skill-source subsystem, not this repo's own `.agents/skills/` content, so the drift rule is not engaged. No architecture page was edited (so no `Last verified:` bump is due), and no ADR changed state.

### Verdict
REQUEST_CHANGES — the CI gate's `pipefail`/`grep -q` construct can pass while plan files are present, defeating the check's whole purpose, and v1/v2 now ship with byte-identical dispatch descriptions but contradictory git contracts.

---

.github/workflows/plan-check.yml
tomkis: Fixed in 05d6edad. The check now tests a captured variable, no pipe involved.

.agents/skills/plan-feature-v2/SKILL.md
tomkis: Fixed in 05d6edad. v2 description now triggers only on an explicit v2 ask and names the contract difference (plan committed vs untracked). Dropped the independently-shippable wording.

.agents/skills/plan-feature-v2/SKILL.md
tomkis: Addressed in 05d6edad. v2 now states the handoff: the implementer resumes on the existing branch and the cleanup commit is mandatory before the PR is marked ready. I kept implement-feature itself on v1 on purpose, it will learn the v2 flow when v2 replaces v1.

.agents/skills/plan-feature-v2/SKILL.md
tomkis: Fixed in 05d6edad. Step 5 and the PR body template moved inside what-to-do, the template got its own heading.

.agents/skills/implement-feature/SKILL.md
```diff
@@ -32,27 +33,28 @@ Before touching code, read the full context:
 
 ### 2. Resume check
 
-Derive the feature branch name deterministically (see step 4) and check whether it already
-exists:
+Derive the feature branch name deterministically (see step 4) and check its state:
```
dam-code-guardian, line 36: 🟢 **Suggestion:** The resume check no longer has a branch-absent case.

Both bullets now assume the branch exists ("Sub-issue commits on the branch" / "Only the plan commit"), and step 4 states `plan-feature` already created it with a draft PR open. But the Input section still accepts a bare `docs/plan/<NNN-slug>/` path — reachable for a plan written before this change, or one created by hand — and in that case the skill has no instruction for what to do.

One line closes it: create the branch from `main`, or stop and point the user at `plan-feature`.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `9b79c82`

### Summary
The v2 rework is now **promoted in place**: `plan-feature-v2` no longer exists as a separate skill — its content replaced `plan-feature` and v1 was dropped, so the branch ships one skill, not two. `implement-feature` was updated to consume a committed plan (check out the existing branch, mandatory cleanup commit, flip the draft PR ready), and the `Plan check` workflow keeps `docs/plan/` out of `main`. All four blocking findings from the previous review are fixed.

### Changes since last review
Previous HEAD: `63d5926` (2026-07-30T10:13:41Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** The CI gate's `pipefail`/`grep -q` SIGPIPE hole — the check now captures `files=$(git ls-files -- docs/plan)` and tests `[ -n "$files" ]`, no pipe involved (`.github/workflows/plan-check.yml:27`)
- ✅ **Fixed:** The byte-identical v1/v2 dispatch descriptions — resolved structurally rather than by rewording: v2 was promoted into `plan-feature` and v1 removed, so only one skill answers "plan a feature" (`.agents/skills/plan-feature/SKILL.md:3`)
- ✅ **Fixed:** The retracted "independently-shippable" promise is gone from the description, which now says "context-window-sized sub-issues, committed as the first commit of the feature branch with a draft PR" — matching the body (`.agents/skills/plan-feature/SKILL.md:5`)
- ✅ **Fixed:** The plan's git lifecycle is now consistent across both consumers — `implement-feature` step 4 *checks out* the branch instead of creating it, its resume check distinguishes "only the plan commit" from "sub-issue commits", and step 7 makes the cleanup commit **mandatory** with the CI gate named as the reason (`.agents/skills/implement-feature/SKILL.md:103`)
- ✅ **Fixed:** `### 5. Report` and the PR-body template are back inside `<what-to-do>` (which now closes at line 126, after the template) — matching `grill-me`'s tag convention (`.agents/skills/plan-feature/SKILL.md:106`)
- ✅ **Fixed:** The leftover `docs/plan/3041-invocation-spend-attribution/` folder is deleted, so `git ls-files -- docs/plan` is empty at this HEAD and the PR passes its own new check
- 🔁 **Still present:** No `concurrency` group on the workflow, unlike `cd.yml:15` (`.github/workflows/plan-check.yml:1`)
- 🔁 **Still present:** `SKILL.md` has no trailing newline — carried from the v2 file into the promoted one (`.agents/skills/plan-feature/SKILL.md:210`)

### Findings
- 🟢 **Suggestion:** `implement-feature`'s resume check no longer has a branch-absent case. Both bullets now assume the branch exists ("Sub-issue commits on the branch" / "Only the plan commit"), and step 4 states `plan-feature` already created it — but the skill's Input section still accepts a bare `docs/plan/<NNN-slug>/` path, which is reachable for a plan written before this change or one created by hand. In that case the skill has no instruction for what to do. Worth one line: create the branch from `main`, or stop and point the user at `plan-feature`. (`.agents/skills/implement-feature/SKILL.md:36`)

✅ No findings. The changed set is two repo-local agent-skill manifests, one CI workflow, and the deletion of a `docs/plan/` working-artifact folder — no subsystem's behavior or responsibility changes, and no page under `docs/architecture/` (nor the landing page) is touched, so no `Last verified:` bump is due. `docs/architecture/skills.md` documents the platform's connectable git-based skill-source subsystem — how skills are transported between external repos and the per-agent PVC — not this repo's own `.agents/skills/` content, so the drift rule is not engaged by editing a skill manifest here. No ADR changed state.

### Verdict
APPROVE — every blocking finding from the previous review is genuinely fixed, and promoting v2 in place (rather than shipping both skills) resolves the dispatch-ambiguity critical at the root. One new 🟢 and two carried-over 🟢s remain, none blocking.

---

PetrBulanek (approved):

PR #3043, opened 2026-07-29
Spend by agent credits the agent that drove the work

### LINKED ISSUE #3041: Spend by agent should credit the agent that drove the work
## Problem

When an agent delegates work by spawning invocations, the Usage page splits the spend across every short-lived agent that ran, and each one appears under a machine-generated name.

A user opening Settings → Usage after a month of heavy delegation sees three agents they recognise at the top, then twenty rows called `invocation-737519…`, `invocation-d3df9fd…` and so on, together holding a large share of the month. The agent that actually drove all of it, and the one the user thinks of as responsible, shows only what it spent talking to the model directly.

That defeats the point of the breakdown. The list is meant to answer "which of my agents is costing me money". For anyone using delegation it answers "an unreadable list of things you have never heard of". Two consequences follow:

- **The rows are not actionable.** The agents they name have already been cleaned up by the time the user sees them, so there is nothing to open, inspect, rename, or switch off.
- **Spend is understated where the user looks first.** A heavy delegator can appear cheaper than a light one, which is exactly backwards.

The month total is correct throughout, and so is every other breakdown on the page. No money is missing or counted twice. It is filed under names nobody can use.

## Proposed solution

Spend an agent causes by delegating counts as that agent's spend. The Usage page shows one row per agent the user actually created, covering both what it spent directly and what it spent through work it handed off. The short-lived agents drop out of the list.

The breakdown still adds up to the month total, so the headline number does not move and the rows stay a straight split of it.

### Scope

Applies from the change onwards. The link between a delegating agent and the work it spawned is not kept long enough to reconstruct after the fact, so existing history keeps its current shape rather than being backfilled.

Related: #3000 grouped experiment-spawned sandboxes under their parent in the sandbox list. Same problem on a different surface.

### Out of scope

- **Per-invocation cost drill-down.** Worth having later on the agent's own page, where an invocation has a prompt and a result to give it meaning. Not on a global list.
- **Session-level rollup.** "What did this session cost" keeps answering for that session alone, not for the work it handed off.

tomkis:
When an agent delegates by spawning invocations, the Usage page currently scatters the spend across short-lived `invocation-<hex>` rows while the agent responsible shows only its direct spend. After this change, spend an agent causes by delegating counts as that agent's spend: one row per agent the user actually created, still summing exactly to the month total. Short-lived invocation agents drop out of the list. Applies from the change onwards; existing history is not backfilled.

Closes #3041

## Sub-issues

- [x] 01 — Gateway attribution override + invocation id stamp
- [x] 02 — Spawn resolves root Driver and stamps the target
- [x] 03 — Spend-by-agent label excludes child rows
- [x] 04 — Architecture docs

https://claude.ai/code/session_019pgyqcu2G9sCcfjhFgxAzm


packages/api-server/src/modules/invocations/services/invocations-service.ts
```diff
@@ -170,6 +190,18 @@ export function createInvocationsService(deps: {
...
 
+      // Resolve the root Driver BEFORE the row (fail at the door — no row to
+      // clean up). The target has no spend identity of its own; its telemetry
+      // is the Driver's, so the target's gateway must attribute upward from the
+      // first exported record. A chain that can't resolve fails closed: stamping
+      // self would manufacture the orphan row this feature eliminates (#3041).
+      const rootId = await deps.driverResolution.resolveRoot(
+        input.driverAgentId,
+      );
+      if (rootId === null) {
```
dam-code-guardian, line 201: 🟢 **Suggestion:** This fail-closed path — and the `telemetryAttributionId: rootId` stamp a few lines below — has no direct unit test. `invocations-result-coercion.test.ts` only adds an identity stub (`resolveRoot: async (id) => id`) to keep the existing suite compiling; nothing asserts that a `null` resolution throws `UnresolvableDriverError`, or that the resolved root id reaches the create input. Since this is the security-load-bearing "fail closed rather than manufacture an orphan row" guarantee, a focused test would be worth adding.

packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts
```diff
@@ -133,14 +133,19 @@ export function createClickhouseReader(
...
+      // excluding child rows (those carrying a `platform.invocation.id`) whose
+      // name belongs to the target, not the Driver the row is attributed to.
+      // This keeps a heavy delegator's bar labelled with its own name instead
+      // of the newest target's `invocation-<hex>`. A since-deleted agent still
       // shows its last known name. The name is display-only; the id is the key.
       const r = await rows(
         `SELECT
            ResourceAttributes['platform.agent.id'] AS agentId,
-           argMax(ResourceAttributes['platform.agent.name'], Timestamp) AS agentName,
+           argMaxIf(ResourceAttributes['platform.agent.name'], Timestamp, ResourceAttributes['platform.invocation.id'] = '') AS agentName,
```
dam-code-guardian, line 148: 🟢 **Suggestion:** This `argMaxIf` label change is covered only by a mocked `spendByAgent` in the metrics tests — no test pins that a bucket whose only in-window rows carry a `platform.invocation.id` yields an empty `agentName` (the client then falls back to the id). Slice 03's acceptance criterion asserts exactly this; a reader-level test would lock it against regression.

packages/api-server/src/modules/invocations/compose.ts
```diff
@@ -33,6 +33,9 @@ export function composeInvocationsForOwner(opts: {
     owner: opts.owner,
     repo: createInvocationsRepository(opts.db),
     agents: opts.agents,
+    driverResolution: createDriverResolution({
+      repo: createInvocationsRepository(opts.db),
```
dam-code-guardian: 🟢 **Suggestion:** `createInvocationsRepository(opts.db)` is constructed twice here — once for `repo` above (line 33) and once inside `driverResolution`. Harmless (repositories are thin stateless wrappers), but hoisting to a single `const repo` reused for both fields reads cleaner and matches the single-repo pattern used elsewhere in this file.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `b98caf68`

## PR #3043: Spend by agent credits the agent that drove the work
**Author:** tomkis | **Branch:** feat/3041-invocation-spend-attribution → main | **Changes:** +692 −89 (29 files)

### Summary
Attributes an Invocation target's telemetry spend to its **root Driver** instead of the short-lived target: the api-server resolves the root Driver at spawn time and stamps it into a new service-only `telemetryAttributionId` Agent-CR spec field; the target's gateway then stamps that id as the trusted `x-platform-agent-id` and its own id as a new `x-platform-invocation-id`, and the spend read path excludes child rows from the per-agent label. Clean 4-slice implementation (Go/Helm plumbing → api-server spawn → read-path label → docs).

### Findings
- 🟢 **Suggestion:** The two new spawn behaviors — the fail-closed `UnresolvableDriverError` (409) path and the `telemetryAttributionId: rootId` stamp — have no direct unit test. `invocations-result-coercion.test.ts:39` only adds an identity stub (`resolveRoot: async (id) => id`) to keep the existing suite compiling; nothing asserts that a `null` resolution throws, or that the resolved root id lands in the create input. Given this is the security-load-bearing "fail closed rather than manufacture an orphan row" guarantee, a focused test would be worthwhile (`packages/api-server/src/modules/invocations/services/invocations-service.ts:201`).
- 🟢 **Suggestion:** The `argMaxIf` label change (`clickhouse-reader.ts:148`) is covered only by a mocked `spendByAgent` in the metrics tests — no test pins that a bucket whose only in-window rows carry a `platform.invocation.id` yields an empty `agentName`. Slice 03's acceptance criterion asserts exactly this behavior; a reader-level test would lock it against regression.
- 🟢 **Suggestion:** `composeInvocationsForOwner` constructs `createInvocationsRepository(opts.db)` twice (`compose.ts:33` and `:37`) — once for the service, once inside `driverResolution`. Harmless (repositories are thin stateless wrappers over `db`), but hoisting to a single `const repo` reads cleaner and matches the single-repo pattern used elsewhere in the file.
- ✅ **Looks good:** The "never wire-settable" guarantee holds — `telemetryAttributionId` is absent from `agentCreateInputSchema` (`schemas.ts:60`) and Zod `.object()` strips unknown keys, so a forged wire value can never reach the spec. Only the in-process `AgentsService.create` input and the spawn path set it.
- ✅ **Looks good:** All four Go call sites of the widened `BuildEnvoyBootstrapConfigMap`/`renderEnvoyBootstrap` signatures are updated consistently — production (`agent_reconciler.go:106` passes the spec field, `fork.go:119` passes `""`) and both test helpers.
- ✅ **Looks good:** `attributionOverridden()` correctly treats an override equal to the own id as a no-op (own id stamped, invocation-id header stripped), pinned by `TestRenderEnvoyBootstrap_TelemetryAttributionOverrideEqualToInstanceIsNoop`. The forge-prevention strip (`request_headers_to_remove` when not overriding) is the right default.
- ✅ **Looks good:** Fail-closed resolution happens before `repo.insert`, so an unresolvable chain leaves no `invocations` row and no Agent behind — matching slice 02's acceptance criterion.

✅ Verdict: **aligned.** The three subsystems whose behavior changed — observability (trusted attribution), metrics (spend read path), logging (gateway telemetry) — each have their architecture page updated in the same PR, with `Last verified:` bumped to `2026-07-29`. No ADR references or volatile-content leaks introduced; no new subsystem needing a page. The `docs/plan/` working-plan files are out of scope (not under `docs/architecture/`).

✅ No architecture findings. The change respects the layered model: `DriverResolution` is a port defined in the services layer and injected into `createInvocationsService`; `compose.ts` (the composition root) is the only place wiring the concrete adapter; the port type is re-exported through the module's `index.ts` public API. `UnresolvableDriverError` is thrown by the service and mapped to a 409 at the endpoint boundary, keeping error-to-HTTP translation out of the service. No cross-layer or cross-module import violations introduced.

### Verdict
**APPROVE** — Correct, well-scoped, and cleanly layered; the trust boundary (service-only attribution field) and fail-closed spawn are handled properly. The only notes are optional test-coverage and a cosmetic dedup suggestion.

---

tomkis:
🤖 Ralph (cycle #1) addressed this on `feat/3041-invocation-spend-attribution`.

Addressed all three dam-code-guardian suggestions on PR #3043: added a unit test for the fail-closed UnresolvableDriverError path and the telemetryAttributionId root-driver stamp, added a reader-level test locking the spendByAgent argMaxIf child-row exclusion (empty agentName for all-child buckets), and deduped the double createInvocationsRepository in compose.ts into a single hoisted const repo.

commit: `33af4a78`

files: `packages/api-server/src/modules/invocations/compose.ts`, `packages/api-server/src/__tests__/unit/invocations-spawn-attribution.test.ts`, `packages/api-server/src/__tests__/unit/metrics-reader-spend-label.test.ts`

PR #3003, opened 2026-07-28
feat: KubeVirt vm backend — full-VM agent sandboxes

JanPokorny:
Implements the KubeVirt VM sandbox MVP per `docs/notes/vm-sandbox-proposal.md` — agents with `backend.type: vm` reconcile to a `kubevirt.io/v1 VirtualMachine` running a Fedora bootc guest (docker + k3s inside the sandbox) instead of a StatefulSet.

## What's in here

- **controller**: `spec.backend` union (schema gen 6), `BuildAgentVirtualMachine` + per-kind reconcile seams, cloud-init NoCloud secret (platform env, MITM CA, virtiofs fstab), workspace PVCs via virtiofs.
- **api-server**: Service-DNS addressing for VM-backed agents, backend passthrough, guards.
- **helm**: `virtualization.*` values block, `claude-code-vm` template (rendered only when enabled), KubeVirt RBAC.
- **agents**: `packages/agents/claude-code-vm` — bootc containerDisk pipeline (`claude-code` OCI image → bootc image → qcow2 via bootc-image-builder → containerDisk); `dam-vm` removed.
- **deploy**: one-flag local setup — `mise run cluster:install -- --set=virtualization.enabled=true` creates the lima VM with nested virtualization, installs KubeVirt (correct v1.8 feature gate `EnableVirtioFsStorageVolumes`, emulation knobs when KVM is absent), and installs the chart.

## Local-experience shakedown findings (fixed in this PR)

- virt-handler node-labeller crash-loops on nodes where `/proc/misc` lists kvm but KVM is unusable — fixed with a json `customizeComponents` patch forcing `PREFERRED_VIRTTYPE=qemu` (a strategicMerge patch wedges virt-operator).
- bootc-image-builder refuses registry pulls — build.sh bridges docker-daemon → containers-storage via containerized skopeo (named volume, since overlayfs won't run on macOS-shared bind mounts), passes `--rootfs xfs`, and preflights loop-device support with a clear failure.
- **aarch64 KubeVirt is KVM-only** (webhook allows only `host-passthrough` CPU; qemu can't do that without KVM) — TCG fallback is x86_64-only, so Apple silicon needs nested virt (M3+/macOS 15), which the install now sets up automatically and enforces with early, actionable failures.
[...]

PR #3044, opened 2026-07-29
fix(agent-runtime): background work survives the session reap (#2965)

### LINKED ISSUE #2965: Background work started from a chat session dies when the user closes the UI
## What happens

You ask an agent in the chat UI to start long-running work (e.g. "ok, now run the experiment") and the agent launches it as a background process inside the chat session. When you close the tab or navigate away, the platform reaps the idle session within seconds and the background work dies with it, partway through.

The loss is silent: nothing warns that closing a tab is destructive, and when you come back the conversation looks fine (sessions rehydrate), but the work is gone. Work spawned as its own invocation is unaffected. Scheduled sessions don't help — this hits any session created through the UI.

## What should happen

Long-running work an agent starts from a chat session survives the user leaving. Closing the UI should never kill work in progress — either the session stays alive while background work is running, or the work runs somewhere the session lifecycle can't touch it.

## Steps to reproduce

1. Open a chat session in the UI and ask the agent to start a long-running background process (an experiment loop, or any `sleep`-style long job in the background).
2. Close the browser tab.
3. Wait a few seconds, then reopen the session: the conversation is intact but the background process is dead.

jezekra1:
Fixes #2965.

## The bug

Closing the last chat tab detached the channel, agent-runtime reaped the idle session 3 s later, and the harness killed the background job it was supervising as its per-session subprocess went down. Work died seconds after the user left, silently — hibernation was never the culprit.

Reproduced on `main` against the shipped `platform-claude-code` image by replaying what the runtime sends (`session/new` → `session/prompt` → wait 3 s → `session/close`):

| work started as | after the turn | after `session/close` |
| --- | --- | --- |
| Bash tool, `run_in_background: true` | alive | **killed** |
| Bash tool, plain `cmd &` | alive | survives |
| Bash tool, `setsid …` | alive | survives |

So the reap kills exactly the background work the harness supervises. The CLI binary even names it: `print wind-down: killing background shell`.

## Why a contract and not detection

ACP has no liveness signal. Checked against the schema bundled with the adapter in the image: no heartbeat, keepalive, liveness or background concept anywhere, and `session/update` streams only inside a prompt turn. ACP v2 makes it worse rather than better — it states that background activity continues while a session reads `idle`, and that `session/close` **MUST** cancel ongoing work. So closing will never become safe on its own.

An earlier revision of this PR inferred liveness from the pod process table (baseline the harness subtree at turn start, diff at turn end). It worked end to end, but it was inference that could be wrong in both directions, and both defects found in review lived in exactly that code. It is deleted here.

The replacement: **Claude Code already knows.** Every `Stop` hook receives `background_tasks` — the session own live task registry, described in the CLI schema as letting a hook *"distinguish session is done from session is paused waiting for background work to wake it"*. And a finishing task wakes the model for a followup turn whose `Stop` reports the now-empty set, so both edges exist. Measured in the image:

```
+2.4s   PreToolUse    run_in_background=true
+2.5s   PostToolUse   run_in_background=true        ← launch, not completion
+4.8s   Stop          background_tasks=[{"id":"bhiwevu6r","type":"shell","status":"running",
                                         "description":"Sleep for 10 seconds in background",
                                         "command":"sh -c 'sleep 10'"}]
+14.5s  Stop          background_tasks=[]           ← job ended, followup turn re-reported empty
```

## The contract

`POST {PLATFORM_RUNTIME_URL}/api/sessions/{sessionId}/background-work` with `{ items: [{ id, description?, command? }] }` — schema and types in `agent-runtime-api`, validated at the route, beside the existing session-reset route.

- **Levels, not edges.** A report is the session complete in-flight set; empty means done. A start/stop pair needs the reporter to be reliable twice — a lost stop holds a session forever, a lost start loses work — while a level self-corrects on the next report in whichever direction it was wrong.
- **Discoverable.** `PLATFORM_RUNTIME_URL` is set on the runtime own env, so every process it spawns (harness, PTY, sshd, `dam-run`) inherits it. No port guessing.
- **Optional.** No report, no hold, exactly the old behaviour. Nothing in the runtime is harness-specific; the Claude Code `Stop`/`SubagentStop` hook is simply the first adapter, and other harnesses have a contract to implement rather than a mechanism to replicate.

While a set is non-empty the runtime will not close that session and reports itself busy — and the controller busy probe is a hibernation veto, so that alone keeps the pod up. No annotation, no controller change.

## Bounds and posture

**Nothing bounds a hold — not a timer, not a count.** Every available bound is a worse failure than the one it prevents.

A time ceiling cannot tell work that has not finished from work that never will, so it only moves #2965 silent kill from seconds to hours. The platform already declines that trade for a running Experiment.

A cap on concurrent holds can be enforced only by reaping a session, which destroys the work being protected — and it would defend a boundary the platform does not otherwise defend, since an open tab keeps the same subprocess alive and engaged sessions are uncapped. (An earlier revision had one; it is gone.)

What keeps a hold honest instead is that it is never silent or final: logged with what it is for, published on `/api/status`, and a hard stop or pause reclaims the pod regardless. `BACKGROUND_WORK_HOLDS=off` refuses holds outright for an install that wants the old behaviour.

Cost, stated plainly: a sandbox with reported background work does not scale to zero until that work ends. If pods do exhaust memory from many live sessions, that is pre-existing and deserves a measured answer covering open tabs too.

## Why `Stop` and not another hook

Enumerated all 30 hook events the CLI defines and registered every one during a 10 s background job. Nothing fires at completion:

```
+5.5s   PostToolUse       ← launch, not completion
+7.2s   Stop              background_tasks=[{… status:"running" …}]
+15.6s  UserPromptSubmit  ← completion wakes the model (injected prompt)
+17.8s  Stop              background_tasks=[]
```

No `Notification`, no `TaskCompleted` (that one is the teammate task board — its payload is task_subject/teammate_name). `UserPromptSubmit` arrives ~2 s earlier but carries only prompt text, so using it would mean pattern-matching an injected sentence for an edge. `Stop`/`SubagentStop` are the only events carrying the inventory. The wake-up prompt also shows the followup `Stop` is a designed consequence of completion rather than a race.

## Verification

202 unit tests, `mise run check` clean, doc-size within budget. End to end in the shipped image with the **real** hook script driving the **real** registry and contract schema — 5/5:

- job started → held, named in the log and on status
- session with a stdio MCP server and no job → never holds
- job finishes → followup `Stop` reports empty → released, nothing held

## What this does not cover

Only work a harness *supervises* is reported. A job the agent detaches (`setsid`/`nohup`) is invisible to the contract; it survives the reap anyway, so hibernation stays its only killer, budgeted by the per-agent timeout. And what is reported can be adjacent to the real work: the experiments launcher deliberately detaches its loop and watches it with a supervised log tail, so that session holds for the tail rather than the loop — harmless, since `experiment-active` does the real work there, but not to be mistaken for protecting experiment runs.

## Follow-ups worth filing

- **Upstream, and now much narrower:** ask the adapter to keep a backgrounded Bash tool call non-terminal (`in_progress`) until the task settles, instead of marking it `completed` with a `backgroundTaskId`. ACP v2 explicitly permits it — *"Tool call updates may arrive while the Agent reports idle"* — and it does not touch the adapter shells-never-defer invariant. That makes the signal protocol-level and harness-agnostic, retiring the hook.
- **v2 migration note:** v2 turns `session/prompt` into an acknowledgment and ends turns with an `idle` `state_update`. The runtime currently defines a turn by the prompt response; that moves, and the synthetic platform turn-ended notification can be dropped.
- UI badge for a held sandbox (the status payload already carries what is needed).

The second commit fixes a prettier violation that landed on `main` in be473587 and was blocking the pre-commit gate; kept separate from the fix.


packages/agent-runtime/src/modules/acp/services/acp-runtime.ts
```diff
@@ -95,6 +110,14 @@ export interface AcpRuntimeDeps {
   logBytesCap?: number;
   /** Owns the `_meta.platform.*` round-trip; skipped when omitted. */
   sessionMetadata?: SessionMetadataStore;
+  /**
+   * Keeps a session (and the pod) alive while background work an agent started
+   * in it is still running. Omitted — as in most tests — the runtime reaps
+   * purely on session refcount, as it did before #2965.
+   */
+  backgroundWork?: BackgroundWorkTracker;
```
dam-code-guardian: 🟢 **Suggestion:** `backgroundWork?` is optional, so every use downstream is `deps.backgroundWork?.…` (~8 sites). Per the *Required Dependencies — Never Optional* rule, an always-injected dependency called unconditionally reads better — the on/off (here: "no OS pid to track") is owned by the dependency, e.g. a no-op `BackgroundWorkTracker` handed to the fake-agent tests. It's consistent with the existing `sessionMetadata?` / `isTerminalSessionActive?` optional deps on this interface, so treat this as a style note, not a blocker.

dam-code-guardian (dismissed):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `ebcfceb`

## PR #3044: fix(agent-runtime): background work survives the session reap (#2965)
**Author:** jezekra1 | **Branch:** fix/background-work-survives-session-reap → main | **Changes:** +1188 −11 (15 files)

### Summary
The session reap tore down the harness's per-session subprocess seconds after the last tab closed, and the harness killed the background jobs it supervised — so `run_in_background` work died silently. This adds a process-table-based background-work tracker: agent-runtime baselines the harness subtree when a prompt is forwarded, attributes anything appearing *below the session's own subprocess* by turn end as that turn's work, and holds the reap (and reports non-idle) while such a process lives — bounded by a 4 h per-session ceiling and a 2-session concurrent cap.

### Findings
- ✅ **Looks good:** Textbook layering — `domain/process-tree.ts` is pure (zero imports), `infrastructure/process-table.ts` is a `ProcessTable` port + procfs adapter importing only the domain, and `services/background-work-tracker.ts` holds the policy. DI at the composition root keeps pids out of the runtime.
- ✅ **Looks good:** Attribution is structural, not identity-guessing — baseline-diff of harness descendants, `newDescendantsBelowBaseline` keeps sibling sessions apart, and `liveKeys` stays ancestry-blind so a job reparented onto init still counts. The `(pid, startTicks)` key correctly survives pid reuse.
- ✅ **Looks good:** `parseProcStat` counts fields from the **last** `)`, so a `comm` containing spaces/parens can't shift the offsets, and it returns `null` on a short/garbage line rather than fabricating zeros (`process-table.test.ts` pins both).
- ✅ **Looks good:** Failure modes are safe by construction — absent procfs → empty table disables tracking, `undefined` harness pid → no attribution, `maxHeldSessions: 0` → tracking off. The deferred-reap timer re-arms only while work lives and unwinds cleanly once a channel re-engages (top guards return without re-scheduling).
- 🟢 **Suggestion:** `backgroundWork?` is introduced as an optional dep, guarded by `deps.backgroundWork?.…` at ~8 call sites (`acp-runtime.ts:118`). It's defensible here (in-process fakes have no OS pid, and the interface already carries `sessionMetadata?`/`isTerminalSessionActive?`), but an always-injected no-op tracker would drop the per-call-site optional chaining. Non-blocking — see the section below.

✅ No findings. `docs/architecture/agent-lifecycle.md` — the correct subsystem page — is updated in the same PR (both the *Session inside the pod* and *Hibernate* sections), `Last verified:` is bumped to 2026-07-29 (the PR date), the added prose stays conceptual with no leaked file paths/package names, and the internal `#session-inside-the-pod` ⇄ `#hibernate` cross-links resolve.

Aligns with the layered architecture throughout — dependency flow (`services → domain ← infrastructure`) is respected exactly, the process table is a port with a procfs adapter, and wiring is confined to `compose.ts`.

- 🟢 **Optional dependency** (`services/acp-runtime.ts:118`, *infrastructure.md → Required Dependencies — Never Optional*): the new `backgroundWork?: BackgroundWorkTracker` is optional and every use is `deps.backgroundWork?.…`. The rule prefers a required dep called unconditionally, with the on/off owned by the dependency (e.g. an injected no-op tracker for the fake-agent tests, which have no pid to track). Consistent with the interface's existing optional deps, so this is a note, not a blocker.

✅ No findings. The only `.tsx` change (`page-empty-state.tsx`) is a prettier reformat wrapping one `<p>` across lines — no logic, state, or prop changes; `message: ReactNode` keeps the type at the boundary.

### Verdict
APPROVE — a well-layered, thoroughly-tested fix for a real silent-data-loss bug, with bounded worst-case cost and honest documentation of the residual imprecision and the detached-work blind spot; the lone suggestion is a stylistic dependency-injection note.

---

packages/agent-runtime-api/src/modules/background-work/schemas.ts
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * `id` is the reporter's own handle for the work and only has to be stable
+ * within one session. The two descriptive fields exist so a sandbox that stays
+ * awake can be explained to the person paying for it; neither affects any
+ * decision, and both are truncated rather than trusted.
+ */
+export const backgroundWorkItemSchema = z.object({
+  id: z.string().min(1).max(128),
+  description: z.string().max(200).optional(),
+  command: z.string().max(500).optional(),
```
dam-code-guardian: 🔴 **Critical:** `.max(500)` **rejects** an over-long command, but the docstring above says these fields are "truncated rather than trusted" — and the hook (`report-background-work.mjs:47`) sends `task.command` raw. One background item whose command exceeds 500 chars (a `sh -c '…'` pipeline easily does) fails `safeParse` for the **entire** report → the route returns 400 → `fetch` does not reject on 4xx, so the hook's `catch` never fires and it exits 0 having reported nothing → the session is reaped and the job killed. That is #2965 re-triggered by the length of a cosmetic field.

Make an over-long advisory field unable to drop the report — either truncate while mapping in the hook (`.slice(0, 500)` / `.slice(0, 200)`), or make the schema itself truncate with `.transform((s) => s.slice(0, 500))` instead of `.max`-rejecting. Same for `description` (line 15) and `items` (`.max(64)`, line 31) — an over-count of trivially-described tasks shouldn't drop the whole hold either.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `c2542bf`

## PR #3044: fix(agent-runtime): background work survives the session reap (#2965)
**Author:** jezekra1 | **Branch:** fix/background-work-survives-session-reap → main | **Changes:** +669 −10 (16 files)

### Summary
The approach is rewritten wholesale since the last review: the process-table inference (baseline the harness subtree, diff at turn end) is deleted, replaced by an explicit **level-based contract** — a session POSTs its complete in-flight background-work set to `/api/sessions/{id}/background-work`, the runtime holds the reap and vetoes hibernation while the set is non-empty, and a Claude Code `Stop`/`SubagentStop` hook reports the CLI's own `background_tasks` registry. A clean, well-argued design — but one new defect in the report path re-opens the exact silent kill the PR exists to close.

### Changes since last review
Previous HEAD: `ebcfceb` (2026-07-29T11:31:19Z) — verdict APPROVE

- ✅ **Fixed:** the two process-table defects are gone with the approach — attribution is no longer inferred (`services/background-work-registry.ts`).
- ✅ **Fixed (toggle axis):** the on/off switch is now owned by the registry and configured at the composition root (`compose.ts:42`), not carried by the consumer — matches *Required Dependencies — Never Optional*.
- 🔁 **Still present:** `backgroundWork?` is still an optional dep on the runtime, always injected in prod but omitted in tests (`services/acp-runtime.ts:122`). Non-blocking — see the TS section.

### Findings
- 🔴 **Critical:** The report schema **rejects** oversized fields (`z.string().max(500)` etc.) but the docstring and hook treat them as **truncated** — and a rejection silently drops the whole hold. `report-background-work.mjs` sends `task.command` / `task.description` raw (`report-background-work.mjs:47`), and a Claude Code background command (`sh -c '…'` pipeline, long flag list) readily exceeds 500 chars. One such item fails `safeParse` for the **entire** report → the route 400s → `fetch` does not reject on 4xx and the hook's `catch` never fires, so it exits 0 having reported nothing → the session is reaped and the job is killed. That is #2965 verbatim, re-triggered by the length of a cosmetic field. Fix so an over-long advisory field can never drop the report: truncate in the hook to match the schema's stated intent (`.slice(0, 500)` / `.slice(0, 200)` while mapping), or make the schema itself truncate (`.transform`) rather than `.max`-reject. (`packages/agent-runtime-api/src/modules/background-work/schemas.ts:16`)

✅ No findings. `agent-lifecycle.md` updated in the same PR (the reap-hold section + the busy-probe signal list + the blind-spot reframed to "unreported work"), `Last verified:` bumped to 2026-07-29, no ADR leaks, no volatile identifiers added to the prose, no new subsystem page warranted.

- 🔁 1 finding from the previous review still present (see review at `ebcfceb`): `backgroundWork?` optional dependency (`services/acp-runtime.ts:122`, infrastructure.md → *Required Dependencies — Never Optional*). The related toggle-ownership axis **was** fixed — the `enabled` kill switch now lives on the registry, configured at the composition root. What remains is only the optional *presence* of the dep (always injected in prod; optional to ease test setup, which the rule's final clause discourages). Non-blocking.

_(The contract package shape — flat `schemas.ts` + `types.ts` with `z.infer` — is correctly placed per the layer rules; validation lives in the contract package and is consumed at the server route.)_

✅ No findings. The only `.tsx` change (`page-empty-state.tsx`) is a whitespace reformat of the message `<p>` — no logic, state, prop, or markup change.

### Verdict
REQUEST_CHANGES — the design is sound and the docs are honest, but the report path drops the whole hold on a validation reject that the hook swallows silently, re-opening the #2965 silent kill for any background item with an over-long command. Truncate the advisory fields (or make the schema truncate) so a cosmetic length can never lose the work.

---

jezekra1:
Rebased onto `a67018d1` and fixed the critical finding.

## 🔴 Critical — advisory fields could drop the whole hold

Confirmed both halves of the chain before fixing:

- an over-long `command` fails `safeParse` for the **entire** report (`items` is an array — one bad element fails all of it), and
- `fetch` **resolves** on a 400, so the hook catch never fires and it exits 0 having reported nothing.

So a 600-char command line — a training invocation or a shell pipeline reaches that easily — silently lost the hold and the job. #2965 keyed on the length of a cosmetic field, exactly as you described. My e2e never caught it because the test command was `sh -c 'sleep 12'`, 16 characters.

Fixed in the **contract**, not just the hook, so every future reporter inherits it: `description` and `command` are now clamped by a `.transform` rather than length-rejected, and `items` truncates instead of capping — a hold is per session, so one surviving item still holds it, whereas rejecting held nothing. The hook clamps too, since a pathological command line is the one thing that could still push the body past the route size limit.

Structural violations still fail loudly (missing `id`, non-array `items`). That line is deliberate: a length is data-dependent and fails sporadically in production, which is why it must never drop a report; a malformed shape is a reporter bug that fails on its first attempt, where a 400 is the useful answer. `background-work-contract.test.ts` covers the parse-and-report path including the 600-char command that would have lost the hold.

## 🟢 Optional dep — leaving as is, deliberately

Making `backgroundWork` required would churn ~20 test call sites for no behavioural gain, and it would sit inconsistently beside `sessionMetadata?` / `isTerminalSessionActive?` on the same interface. The axis the rule really cares about — who owns the on/off — is already right: the `enabled` switch lives on the registry, configured at the composition root. Happy to flip it if you would rather have the interface uniform in the other direction.

## Unrelated finding, flagging rather than fixing here

`packages/agents/claude-code-vm/Containerfile` (new in #3003) copies `/usr/local/lib/` from the agent image but not `/etc/claude-code/`, so `managed-settings.json` does not ship to the VM backend. That means neither this PR Stop hook **nor the existing WebSearch/WebFetch deny hooks** are active in VM agents. One `COPY --from=agent /etc/claude-code/ /etc/claude-code/` would fix both, but it is someone else brand-new image and predates this PR, so I have left it alone — say the word and I will add it here or file it separately.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `2cf5180`

## PR #3044: fix(agent-runtime): background work survives the session reap (#2965)
**Author:** jezekra1 | **Branch:** fix/background-work-survives-session-reap → main | **Changes:** +768 −12 (16 files)

### Summary
The level-based **background-work contract** design is unchanged and sound; this revision fixes the one critical defect from the last review — the report path that dropped the whole hold on an over-long advisory field — at the contract layer, so every future reporter inherits the fix.

### Changes since last review
Previous HEAD: c2542bf (2026-07-29T13:37:21Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** advisory fields no longer `.max`-reject — `description`/`command` clamp via `.transform(slice)` and `items` truncates to 64, so an over-long command can no longer fail `safeParse` for the entire report and silently lose the hold. Fixed in the contract (`schemas.ts:16`) *and* the hook (`report-background-work.mjs:47`), with a regression test driving a 600-char command through parse→report (`background-work-contract.test.ts`). Structural violations (missing `id`, non-array `items`) still fail loudly — the right line.
- 🔁 **Still present:** `backgroundWork?` remains an optional dep on the runtime (`services/acp-runtime.ts:122`) — non-blocking; author addressed it deliberately (see TS section).

### Findings

✅ No findings. `agent-lifecycle.md` reflects the final contract mechanism (levels-not-edges, optional reporting, the unbounded hold with only the concurrency-independent kill switch, the reframed detached-work blind spot); `Last verified:` is 2026-07-29; no ADR leaks; no volatile identifiers added to the prose.

- 🔁 1 finding from the previous review still present (see review at c2542bf): `backgroundWork?` optional dependency (`services/acp-runtime.ts:122`, infrastructure.md → *Required Dependencies — Never Optional*). The toggle-ownership axis the rule cares about is correct — the `enabled` kill switch lives on the registry, configured at the composition root (`compose.ts`). What remains is only the optional *presence*, consistent with sibling optional deps (`sessionMetadata?`, `isTerminalSessionActive?`). Author has flagged this as a deliberate call and offered to flip it — non-blocking. The critical-fix files are clean: the contract stays a flat schemas+types module (correct per slice-composition), the registry is a proper port-returning factory.

### Verdict
APPROVE — the critical report-path defect is fixed at the contract layer and covered by a targeted regression test; no new findings, and the sole carryover is a non-blocking, author-acknowledged optional-dependency note.

---

tomkis:
One observation from reading this, not blocking.

I found the reaping logic hard to follow. The decision whether a session gets reaped is spread across several places that each schedule or clear a timer: the detach path schedules one, the background-work branch reschedules it, and teardown, harness exit and reset all clear them. I had to trace all of these to convince myself a held session eventually gets closed.

The same applies to the new backgroundWork condition. It now has to be consulted in two places independently: the reap path asks the registry per session, and status() folds it into the idle flag for hibernation. So "is this session actually done" is a predicate that lives in two spots and has to stay in sync by hand. If someone later adds another hold condition, they have to remember both.

I know the timer-per-session pattern predates this PR, but the hold made the reap logic stateful enough that I think it's worth refactoring later: one predicate for "session is done", used by a single loop that periodically reaps and by status(). One place to read, and a missed edge heals on the next tick, which fits the same "levels, not edges" reasoning this PR argues for.

Filed as a follow-up refactor in #3108, nothing to change here.

tomkis (approved):

PR #3045, opened 2026-07-29
feat(skills): delete or download a standalone skill

### LINKED ISSUE #2825: Delete or download a standalone skill
## Problem

Skills created directly in a sandbox ("Created in this sandbox") show **Delete** and **Download** actions in their menu, but both are disabled. A user can't remove a standalone skill they no longer want, nor take a copy of one they authored in the sandbox.

## Proposed solution

Enable **Delete** (remove the standalone skill from the sandbox) and **Download** (get a copy of the skill's files) from the standalone-skill menu.

### Dependencies

Deferred from #944 (needs backend endpoints to delete and to package/serve a local skill).

PetrBulanek:
Skills created directly in a sandbox showed **Delete** and **Download** in their menu, but both were disabled — you couldn't remove a standalone skill you no longer wanted, or take a copy of one you'd authored there.

Both now work:

- **Delete** removes the skill from the sandbox, behind a confirmation. If you'd published it, the dialog says so and links the pull request — deleting the skill here doesn't withdraw it upstream.
- **Download** hands you the skill's files: a single-file skill downloads as Markdown so it can be re-uploaded as-is, anything larger as a zip that unpacks into the skill's own folder.

Two fixes ride along, both user-visible:

- Publishing a skill whose name wasn't already lowercase-and-hyphenated used to fail outright. It now works.
- The sandbox sidebar's Skills line no longer goes stale after one of these skills is added or deleted.

Skills installed from a GitHub source are untouched — they keep their install toggle, and uninstalling is still how you remove those.

Closes #2825


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `c93cafc9`

## PR #3045: feat(skills): delete or download a standalone skill
**Author:** PetrBulanek | **Branch:** feat/2825-standalone-skill-delete-download → main | **Changes:** +497 −73 (22 files)

### Summary
Wires up the previously-disabled **Delete** and **Download** actions for standalone Local Skills end-to-end (agent-runtime → api-server → UI), adds a shared name→directory resolver so name-keyed ops handle a frontmatter `name:` that diverges from the on-disk dir, and rides along two user-visible fixes: publishing a non-slug-shaped skill name (slugified for the git ref) and a stale sidebar Skills line after standalone add/delete.

### Findings
- 🟢 **Suggestion:** The publish slug bugfix and the three new backend operations ship without unit tests — `publish.test.ts` still only exercises the already-slug-shaped `"my-skill"`, so the display-name→ref-safe-slug branch (`makeSkillSlug` on `"My Cool Skill"`), `resolveLocalSkillDir` (exact-dir-first vs frontmatter-name match), and `deleteLocal`/`readLocal` are unverified. A couple of focused cases would lock in the behavior the PR body promises. (`packages/agent-runtime/src/modules/skills/services/publish.ts:339`)
- 🟢 **Suggestion:** `resolveLocalSkillDir` returns the dir from the *first* matching Skill Path, and `deleteLocal` then removes that basename from *every* path. If two paths held different directory names resolving to the same frontmatter name, one could survive — a corner `writeLocal`'s consistent-slug invariant already prevents, so this is a note, not a defect. (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:253`)
- ✅ **Looks good:** Delete is idempotent (a target already gone resolves to `ok(undefined)` — safe against a double-click) and refuses source-tracked names with `CONFLICT`, so it can never wipe an install and orphan an `agent_skills` row.
- ✅ **Looks good:** `AgentRuntimeClientError` + `asPodVerdict` correctly re-throw pod client-error verdicts (`NOT_FOUND`/`PAYLOAD_TOO_LARGE`/`BAD_REQUEST`) with their own status instead of flattening a user error into a 500, and `readLocal`/`deleteLocal` are owner-scoped via `ensureAgentReachable`.
- ✅ **Looks good:** `skill-download.ts` types file bytes as `Uint8Array<ArrayBuffer>` (avoiding the `SharedArrayBuffer` `BlobPart` pitfall), names the download from the on-disk `dir`, and defers `revokeObjectURL` so the download isn't cancelled.

**Verdict:** aligned — the skills architecture page reflects the new `deleteLocal`/`readLocal` responsibilities in the same PR, its `Last verified:` header is bumped to 2026-07-29, and there are no ADR-reference or volatile-content leaks.

✅ No architecture violations. Validation (Zod) stays in the contract packages with routers delegating to the service interface; `resolveLocalSkillDir` is declared on the `LocalSkillRepository` port and implemented in `infrastructure/`; `makeSkillSlug` stays pure and returns a `Result`; the api-server service throws `TRPCError` consistent with the existing `createLocal` CONFLICT path.

✅ No new findings. `SandboxSkillsSection` mirrors reconciled state into the TanStack Query cache (no Zustand/`useState` shadow copy of server state) gated on `stateLoaded`; `deleteWithConfirm` keeps the dialog in the component and delegates the mutation to the hook; props are cleanly typed from the contract package. The hook's `useState`+`runAction` shape is pre-existing (whole-hook TQ migration is a separate-PR concern).

### Verdict
APPROVE — a cohesive, well-reasoned feature that closes the deferred Download/Delete affordances with correct idempotency, ownership, and error-code handling; the only asks are optional test coverage for the new paths.

---

packages/ui/src/modules/sandboxes/components/skills/skills-surface.tsx
```diff
@@ -72,13 +78,46 @@ export function SkillsSurface({
...
+  ) => {
+    // State-neutral wording about the PR: the "In review" pill is never
+    // refreshed against GitHub, so we can't claim the PR is still open (#3019).
+    const ok = await showConfirm(
+      <>
+        This skill will be removed from the sandbox.
+        {pub && (
+          <>
+            {" "}
+            The{" "}
```
kapetr: Why the leading space?

kapetr (approved):

packages/ui/src/modules/sandboxes/components/skills/skills-surface.tsx
PetrBulanek: It actually renders "...sandbox.The..." without it.

Issue #2989, opened 2026-07-27
`dam-run` works without a shared workspace


tomkis:
## Problem

`dam-run` lets an agent run a command in a separate throwaway pod that mounts the agent's workspace at the same time as the agent itself. Agents are told to use it for two things: pushing heavy or long-running work like builds, scans and test suites off the chat pod, and fanning several jobs out across pods that all read and write the same workspace.

The second one depends on the shared volume directly, and that volume is exactly what we are trying to remove. Every agent on the platform currently runs on slow network storage so that a second pod can write into the workspace while the agent is running (#2988). With Slack per-person copies gone (#2987), `dam-run` is the only thing left that needs it. It is the last blocker, and everything else in this epic waits on it.

So this needs a real answer, not a deletion. `dam-run` is genuinely useful and agents rely on it. It should keep its main promise: an agent can push work into a separate pod and watch the output as if it were running locally. What has to change is how that pod gets at the workspace, and the option we pick changes what agents can do with the tool.

### Scope

- `dam-run` no longer needs to write into the agent's workspace while the agent is running.
- Results still land somewhere the agent can see. An agent that runs a build in an executor expects the output to be there afterwards. If that stops being automatic, it has to be obvious.
- The agent-facing description of `dam-run` is rewritten to match. It currently tells agents to fan jobs out across pods sharing one workspace. If that stops being true, leaving the instruction in place means agents keep writing code that quietly does the wrong thing.

### Dependencies

Blocks #2988. Agent storage cannot move until `dam-run` stops needing a shared volume.

### Out of scope

- `dam-vm`, which already shares nothing with the agent pod and is unaffected.
- The concurrency cap and other `dam-run` behaviour not related to the workspace.

jezekra1:
Descoped for now (decided while implementing #2983 in PR #3046): dam-run is **disabled** rather than reworked — the relay refuses every dial with a clear close reason, the Run machinery stays dormant and tested, and the agent-facing description is removed. This unblocked the storage change (#2988). This issue stays open for the real rework: give the executor a workspace model that doesn't need a shared volume (candidates discussed in #3046: same-node RWO co-mount via pod affinity, copy-in/copy-out, clean-room + explicit transfer), then delete the disable switch.

Issue #2983, opened 2026-07-27
Epic - Simplify multiplayer: one agent, one workspace, shared by the channel


tomkis:
## Context

People want a simple thing: put an agent in a Slack channel and let the team talk to it. We used to give them two different answers to that, and the default one — a separate per-person copy of the agent carrying each replier's own credentials — was the bad one. It made every agent slower, offered isolation that didn't actually hold, and forced two confusing access modes on everyone.

That model is now gone. Everyone in a channel talks to **one shared agent** under the agent's own credentials; turns are attributed to whoever sent them and gated by the owner's approval and network rules. Agents also moved off the slow shared filesystem onto ordinary fast storage, so they're quicker at ordinary work and no extra infrastructure is needed to run them. This was the cleanup half of #105.

## Problem

The simplification has shipped, but a small tail of cleanup remains before the model's old assumptions are fully gone:

- One feature still writes into an agent's workspace from a second place, which is the last thing standing between us and retiring the old shared-storage assumptions everywhere.
- We don't yet have a clear read on how the new default storage tier actually performs.

## Goal

Close out the model cleanup and confirm the storage change delivers the speed we expect. Once these land, the simplification is complete.

## Scope

Remaining work tracked under this epic:

- **#2989** — running a command in a throwaway pod no longer needs a shared workspace (the last dependency).
- **#3085** — confirm how fast the default storage tier is.

Already delivered: per-person copies removed (#2987), agents moved to ordinary fast storage (#2988).

Out of scope: enhancements that make the shared experience do *more* — binding one agent to multiple channels (#3086) and continuing a channel conversation in the DAM UI (#3087) — now live under the Channels epic (#3091), since they build on the shared model rather than being part of retiring the old one.

## Open Questions

- Once #2989 and #3085 close, this epic is effectively done and can be closed.

PR #3046, opened 2026-07-29
Simplify multiplayer + RWO storage: remove forks, disable dam-run, migrate workspaces off shared filesystems

### LINKED ISSUE #2987: Slack channels run on the agent itself, with no per-person copies
## Problem

Connecting an agent to a Slack channel makes you pick an access mode you probably shouldn't have to think about. The default, person-scoped, means replies from anyone but you run on a separate per-person copy of the agent under that person's own credentials, and you have to maintain a list of who is allowed to drive it. The alternative, shared, just lets everyone in the channel talk to the agent.

Three things are wrong with keeping the default:

- **You are stuck with the choice.** The mode is fixed when you connect. Changing it means disconnecting the channel and connecting it again.
- **What it promises is not what you get.** Everyone shares the same workspace, files and memory whichever mode you pick, so splitting credentials per person gives a sense of separation that does not exist. Worse, it means a channel member can leave something in the workspace that the agent acts on later under your credentials.
- **Everyone else pays for it.** Per-person copies have to write into your agent's workspace while your agent is running, and that requirement is what forces every agent on the platform onto slow storage, including agents nobody ever connects to Slack.

Nobody is using person-scoped channels today.

## Proposed solution

Connecting a channel gets simpler: no mode to pick, no allowed-users list to maintain.

Every Slack channel behaves the way shared channels behave today. Anyone the channel admits can talk to the agent, every turn runs as the agent under the agent's own credentials, and each turn is attributed to whoever sent it, both in the security log and in what the agent sees. The owner's approval prompts and network rules gate every turn exactly as they do now. This is what Telegram has always done, so the two messengers stop behaving differently.

### Scope

- The connect form in the UI, the CLI connect flag and the in-chat bind command all lose the mode choice.
- The allowed-users list disappears from agent settings.
- The per-person copy machinery is removed outright, not left switched off.
- Signing in from Slack stays. It is still how someone picks one of their own agents when binding a channel from inside Slack, and how they release it later.

### Dependencies

Blocks the storage change under #2983. Per-person copies are the reason the shared workspace is needed at all.

### Out of scope

- Actually switching agent storage. That is its own issue, and it also waits on `dam-run`.
- Rethinking consent for shared channels. Lending an agent to a channel is still a real decision, and if it needs more transparency that is separate work.

### LINKED ISSUE #2988: Agents run on ordinary fast storage instead of a shared filesystem
## What

Move agent workspace volumes from shared-writable storage to ordinary single-writer storage, and take the shared filesystem requirement out of the platform entirely.

That means the bundled in-cluster NFS server comes out of the chart along with the storage class it provides, the host-level NFS client setup comes out of dev and test cluster provisioning, and no install has to point us at a managed shared filesystem any more.

**Every existing agent gets migrated as part of this work.** A volume's access mode cannot be changed in place, so each existing agent's workspace has to be copied onto a new single-writer volume and the agent switched over to it. This is not a follow-up and not an owner's chore: nobody should have to recreate an agent, re-clone a repo, or lose anything they had. What has to come through intact is everything that lives on the volume today, including the workspace itself, the agent's home directory, its memory and skills, and its session history.

Migration needs to survive being interrupted. It runs against a live install with agents in every state, some running, some hibernated, some untouched for weeks, so it has to be safe to stop and resume, and it must not leave an agent pointing at a half-copied volume.

## Why

Shared-writable storage exists for exactly one reason: something other than the agent writing into the agent's workspace while the agent is running. Once Slack per-person copies are gone (#2987) and `dam-run` has its own answer, nothing needs it.

What is left is pure cost, paid by every agent whether or not it ever touches Slack:

- Every file the agent reads or writes goes over a network filesystem. Agents feel sluggish at ordinary work like checkouts, installs and builds.
- Our own dev and test clusters only work because we ship an extra NFS server inside the platform and install NFS client packages on the VM.
- That server's image is archived upstream. It already breaks on btrfs hosts, where the install quietly falls back to single-writer storage anyway. We are maintaining a component we cannot upgrade, to support a feature nobody uses.
- Every production install has to provision a managed shared filesystem just to run us, which is both slower and more expensive than ordinary disk.

## Done when

- New agents get single-writer volumes, and no part of the install asks for a shared filesystem class.
- The bundled NFS server, its storage classes and the NFS client provisioning are gone from the chart and the cluster setup.
- Every pre-existing agent is running on a single-writer volume with its workspace, home, memory, skills and sessions intact, and no owner had to do anything to get there.
- Migration is safe to interrupt and resume, and an agent is never left pointing at an incomplete volume.
- Old volumes are only released once the agent is confirmed working on the new one.
- The warm volume pool still works on the new storage.
- The full test suite passes on a cluster with no shared filesystem available.

### Dependencies

Blocked by #2987 and #2989. Both remove a writer from the agent's workspace, and this cannot land until they have.

jezekra1:
## What

The whole of epic #2983, squashed to one commit on top of main:

1. **Remove person-scoped Slack access and the fork machinery** (closes #2987). The binding is now the only access model: anyone the messenger admits to a bound conversation drives the Agent under the Agent's own credentials, attributed by messenger-native sender id and speaker-labelled in the prompt. Deleted: the Fork CRD + reconciler + per-fork gateway/SA/AuthorizationPolicies, the api-server forks module, the access-mode pickers (UI/CLI/contract), and the allowed-users list end to end. Existing person-scoped bindings are **deleted** by a data migration, not silently converted — conversion would widen who may drive an agent without the owner deciding; re-binding is the consent. Identity linking survives solely to authorize the in-chat bind/unbind/ambient commands. The retired `forks.agent-platform.ai` CRD is **not** touched by the chart — CRDs are managed out of band; the ops-side CRD upgrade additionally runs `kubectl delete crd forks.agent-platform.ai --ignore-not-found --wait=false` once installs are past this release (the leftover CRD is inert meanwhile).

2. **Disable dam-run** (refs #2989 — descoped from rework to disable, decided during design). The executor co-wrote the agent's live workspace, exactly the multi-writer access RWO storage removes. The relay now refuses every dial with a clear close reason; the Run machinery stays dormant and unit-tested; the agent-facing AGENTS.md paragraph is gone. Re-enabling is a deliberate code change once #2989 lands a new workspace model — never a knob, which would quietly break on multi-node clusters.

3. **ReadWriteOnce workspaces + in-place migration + NFS deprecation** (closes #2988). Access mode is retired as configuration (still parsed and ignored with a startup warning so existing values files survive upgrade). A leader-only controller manager drains every remaining RWX workspace volume: gate the agent down (hard-stop strength — in-flight work interrupted by design, as decided), copy the quiesced volume onto a fresh RWO PVC in a Job (GNU `cp -a` preserving hard links + two-pass checksum verification), re-point the agent through the warm-pool claim mechanism, delete the old volume on the verified copy, restore the prior run state. Interrupt-safe and resumable; vm-backend agents are skipped loudly (by-name PVC refs, not claim-based). Warm-pool RWX spares are reaped and re-provisioned RWO. The bundled NFS server is **deprecated in place**: `values-local` keeps it enabled for the transitional window so an in-place dev-cluster upgrade never deletes the server out from under unmigrated volumes; `values-e2e` disables it so the e2e cluster proves the no-shared-filesystem path; fresh installs never enable it; the chart block goes in a follow-up release once fleets are drained.

4. **Docs + ADRs**: architecture pages, strategy multiplayer doc, ubiquitous language; ADR-079 (single access model, supersedes 027/076) and ADR-080 (RWO workspaces + migration).

## Verification

- `mise run check` green across the repo; all unit suites green (controller Go incl. migration state-machine tests, api-server 691, CLI 154, UI 135, agent-runtime 189).
- Full from-scratch e2e: cluster provisions with **no shared filesystem available** (only `local-path`), agent PVCs come up `ReadWriteOnce`, all 20 Playwright specs pass. The fork spec (07-slack) is gone; binding behavior stays covered by 08/09/10.
- **Live upgrade rehearsal** (previous release → this PR, on a real cluster): installed `main` with an RWX workspace volume (static hostPath PV declared `ReadWriteMany` standing in for NFS — ganesha can't serve btrfs, the very reason it's being removed), created an agent, wrote real data (3 MB blob, a hardlink pair, a symlink, nested dirs), left it **running**, upgraded in place. The migration gated the running agent, copied, verified, flipped, and deleted the RWX volume in **61 seconds**; the agent woke back up on the new `ReadWriteOnce` `local-path` volume with checksums identical, the hardlink pair still linked, and the symlink intact. The rehearsal caught two real bugs, both fixed in follow-up commits: `values-local` dropping the NFS enable flag (would have deleted the server under unmigrated dev volumes on in-place upgrade), and a chart post-upgrade hook deleting the forks CRD — wrong on two counts (CRDs are managed out of band, and its scoped RBAC made kubectl's deletion-wait hang the hook), now removed entirely.

## Notes for review

- The diff is net-negative (4.4k insertions vs 5.8k deletions). Of the insertions, 59% is the one generated Drizzle schema snapshot the db workflow requires (the allowed-users drop and the person-scoped-binding deletion ride a single migration) and 23% the new storage-migration manager + tests; hand-written churn is ~0.8k added vs ~5.8k removed.
- Rebased over #3043 (spend attribution): the Envoy bootstrap keeps its new `attributionID` parameter while dropping the fork-era `extAuthzInstanceID`; main's attribution tests are kept, its fork-specific one removed with the feature.
- `testUser2` (the e2e foreign-user fixture) is removed everywhere with the fork spec.


dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `78fe03a`

## PR #3046: Simplify multiplayer + RWO storage: remove forks, disable dam-run, migrate workspaces off shared filesystems
**Author:** jezekra1 | **Branch:** feat/remove-forks-rwo-storage#locki-lpaitkhu → main | **Changes:** +7078 −5809 (185 files)

### Summary
Collapses the Slack access model to a single "shared binding" mode — deleting the Fork CRD, reconciler, per-fork gateway/SA/AuthorizationPolicies, the api-server forks module, and the allowed-users machinery end to end (person-scoped bindings are *deleted*, not converted, by a data migration). Disables `dam-run` (relay refuses every dial; machinery kept dormant and tested), retires workspace volumes to `ReadWriteOnce` via a leader-only, interrupt-safe controller migration that checksum-copies each RWX volume onto a fresh RWO PVC, and deprecates the bundled NFS server in place. Architecture docs, ADR-079/080 included.

### Findings
- 🟢 **Suggestion:** The copy Job verifies file inventory + `md5sum` content, but `cp -a "$src/." "$dst/"` plus a `find`-based inventory does not detect an *empty-directory* mismatch (dirs with no files/symlinks are excluded from both `cmp` lists). Harmless for workspace correctness, but worth a note if any workspace relies on empty-dir presence. (`packages/controller/pkg/reconciler/storage_migration.go:566`)
- 🟢 **Suggestion:** `md5sum` is used for content verification of a quiesced source before the old volume is deleted. Fine here (this guards against copy/IO corruption, not adversarial content), but a one-line comment noting that choice would preempt a future reader's security flag. (`packages/controller/pkg/reconciler/storage_migration.go:569`)
- ✅ **Looks good:** The RWX→RWO migration is a genuinely careful state machine — deterministic target names for resume, quiescence-gated copy, verified-before-delete, `finishFlip` tail for a crash between steps, bounded failed-Job retry, and a loud vm-backend skip. Label discipline (`LabelMigrationFor` vs `LabelAgent` at flip) correctly keeps half-copied volumes invisible to the orphan sweep and warm-pool trim (the latter also gated on `LabelPoolAvailable=true`).
- ✅ **Looks good:** Fork/allowed-users removal is complete and dangle-free — no surviving imports of the deleted module, events (`ForkReady`/`ForkFailed`/`ForeignReplyReceived`), sagas, or `userDirectory`; contract Zod schemas and the `crdcheck` deliberate-removal allowlist updated in step.
- ✅ **Looks good:** The `post-upgrade` (not pre-) ordering of the forks-CRD removal Job, with its rationale about a mid-upgrade foreign turn minting a Fork CR, is exactly right; `--ignore-not-found` keeps it a fresh-install no-op.
- ✅ **Looks good:** Data-migration safety call — deleting person-scoped bindings rather than silently widening them to shared is the correct consent-preserving choice, documented in both the SQL and the ADR.

Verdict: **minor drift**

❌ **`Last verified:` staleness** — 3 architecture pages received substantive content edits but their date was not bumped to the PR date (2026-07-29):
- `docs/architecture/platform-topology.md` (`Last verified: 2026-07-24`) — `Fork` custom resource replaced with `Run` in prose and the `| Fork |` resource-model row dropped.
- `docs/architecture/budgets.md` (`Last verified: 2026-07-27`) — `per-turn Fork/Run pods` → `per-command Run pods`.
- `docs/architecture/observability.md` (`Last verified: 2026-07-24`) — `Agent/Fork custom resources` → `Agent/Run custom resources`.

No architecture-page drift, ADR-reference leaks, volatile-content leaks, or missing-page gaps: `persistence.md` (RWO + storage-migration section) and `channels.md` (single shared-access model, ambient, in-chat bind) are thoroughly updated, and the other edited pages carry the correct date.

✅ No findings. The forks module is removed wholesale (no cross-module leakage), input validation stays in the contract package (`agentConnectSlackInputSchema` correctly drops `mode`/`allowedUserEmails` and its `.refine`), and the disabled-relay wiring (`enabled: false` in the harness app, machinery intact) is clean. (12 files reviewed.)

✅ No findings. The Slack channel modal/card simplification removes the access-mode picker and the allowed-users `useFieldArray` cleanly — no dead imports (`Button`/`Input` still used by the channel field and footer), no dangling `allowedUserEmails` references, no state-lineage issues; the form reduces to the single mutable `ambient` toggle. (2 files reviewed.)

### Verdict
COMMENT — Excellent, carefully-staged removal with strong migration safety and test coverage; the only issue is three architecture pages whose `Last verified:` dates weren't bumped despite substantive edits. Non-blocking.

---

packages/controller/pkg/reconciler/storage_migration.go
```diff
@@ -0,0 +1,652 @@
...
+  # Fresh copy per attempt keeps retries deterministic.
+  find "$dst" -mindepth 1 -delete
+  # GNU cp -a: permissions, ownership, times, symlinks, AND hard links —
+  # hard-link-heavy stores (pnpm) must not inflate past the volume size.
+  cp -a "$src/." "$dst/"
+  sync
+  # Two-pass verification against the quiesced source: the file inventory
+  # (files + symlinks), then content checksums. The old volume is deleted
+  # at flip on the strength of this comparison.
+  (cd "$src" && find . \( -type f -o -type l \) | LC_ALL=C sort) > /tmp/src.list
```
dam-code-guardian: 🟢 **Suggestion:** The inventory comparison lists only files and symlinks (`find . \( -type f -o -type l \)`), so an empty directory present on the source but not recreated on the target would pass verification unnoticed. `cp -a` copies empty dirs, so this is not a live bug — just a gap in the guarantee if any workspace ever depends on empty-dir presence.

packages/controller/pkg/reconciler/storage_migration.go
```diff
@@ -0,0 +1,652 @@
...
+  (cd "$dst" && find . \( -type f -o -type l \) | LC_ALL=C sort) > /tmp/dst.list
+  cmp /tmp/src.list /tmp/dst.list
+  (cd "$src" && find . -type f -print0 | LC_ALL=C sort -z | xargs -0 -r md5sum) > /tmp/src.sum
```
dam-code-guardian, line 627: 🟢 **Suggestion:** `md5sum` is the content check that licenses deleting the source volume at flip. That is fine — the threat here is copy/IO corruption on a quiesced source, not adversarial collisions — but a one-line comment saying so would preempt a future reader (or security scanner) flagging the weak hash.

jezekra1:
## Self-review (session review pass)

**Scope:** 182 files, +4,386/−5,812. CI fully green including e2e on a shared-FS-free cluster. Two live upgrade rehearsals performed (fabricated-state and real previous-version install → upgrade with a user-created agent; 158k files drained and verified in 90s).

### Findings → resolution

1. **A non-terminally stuck copy Job gated the agent forever** (no `activeDeadlineSeconds`, silent in-progress path) — **fixed in 2c451fea**: 4h deadline feeding the existing bounded retry, plus a per-tick progress log.
2. **Agent deletion mid-migration leaked the target PVC** (deliberately unlabeled for the sweeps) — **fixed in 2c451fea**: target PVC + Job are owner-referenced to the Agent CR, so GC reaps them.
3. **Root-squashing RWX classes broke the root-running copy** — real on IBM file storage (squash enabled), and every real workspace carries 0600/0700 entries (`.claude` session files, `.ssh`) that squashed root cannot read — **fixed in 2c451fea**: the Job runs as the agent uid (workspaces are single-owner by construction, so their own uid reads everything; ownership preserved by construction; fsGroup for the fresh target). Documented on the `storageMigration` values block.
4. *Nice-to-have, not done:* deny-all egress NetworkPolicy on the copy pod (it executes only trusted binaries, mounts no SA token, opts out of the mesh — but namespace convention is kernel-level gating; cheap follow-up).
5. *Nit:* `agent.channels.orphan_purge` (renamed security event) is undocumented in logging.md — same as its predecessor, so no regression.

### Residual risks accepted
- vm-backend agents are skipped by the migration (by-name PVC refs) — logged loudly; recreate to move them.
- No automated upgrade-path e2e (fresh clusters can't mint RWX state); covered by the two live rehearsals + 9 state-machine unit tests.
- Real-NFS canary still recommended on one IBM staging cluster before fleet rollout: verify uid passthrough on the migrated volume of a single drained agent, then let the concurrency throttle take the rest.

dam-code-guardian:
📊 **PR #3046 review artifact** — [gist](https://htmlpreview.github.io/?https://gist.githubusercontent.com/dam-code-guardian/24f1c7736bd4d50d57d7bbc6c2b08dff/raw/pr-3046.html) · [DAM Artifact Library](https://share-dam-dev.apps.dam-cl.fmaas.res.ibm.com/a/DsMn0Qnbrhd6OA)

packages/controller/pkg/reconciler/storage_migration.go
```diff
@@ -0,0 +1,714 @@
...
+func buildMigrationJob(agentName string, pairs []migrationPair, cfg *config.Config, ownerRef metav1.OwnerReference) *batchv1.Job {
+	var volumes []corev1.Volume
+	var mounts []corev1.VolumeMount
+	var script strings.Builder
+	script.WriteString(`set -eu
+copy_verify() {
+  src="$1"; dst="$2"
+  echo "migrate: $src -> $dst"
+  # Fresh copy per attempt keeps retries deterministic.
+  find "$dst" -mindepth 1 -delete
```
dam-code-guardian: 🔴 **Critical:** This cleanup was written when the copy ran as **root**; `2c451fea` moved it to the agent uid, and *deletes* aren't covered by the single-owner-workspace argument the way reads are — unlinking an entry needs write+execute on its **parent directory**.

`cp -a` on pass 1 faithfully reproduces source modes, so any `0500`/`0555` directory (Python `site-packages`, Go module cache, pnpm store) makes pass 2 fail. Reproduced as uid 65532:

```
bfs: error: dst/ro_dir/f: Permission denied.
bfs: error: dst/ro_dir: Directory not empty.
exit=1
```

The `prep-target` init container doesn't help: it chowns only the mount root, non-recursively (line 645) — a chowned root still doesn't permit deletion inside a `0500` subdirectory.

With `BackoffLimit: 2` + `RestartPolicy: OnFailure`, the first failure is retried against a non-empty target and every retry dies here; the Job exhausts its backoff and `migrationJobRetryAfter` recreates the same doomed Job, leaving the agent gated indefinitely.

Options: make the target writable first, or let the root init container empty it (it already mounts exactly the targets), or copy into a fresh per-attempt subdirectory:

```suggestion
  chmod -R u+rwX "$dst" 2>/dev/null || true
  find "$dst" -mindepth 1 -delete
```

dam-code-guardian (dismissed):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `b3a085b`

### Summary
Collapses the Slack access model to a single shared binding, disables `dam-run`, and migrates workspace volumes from RWX to RWO via a leader-only controller migration. This re-review covers two hardening commits on the storage migration plus a main merge.

### Changes since last review
Previous HEAD: `78fe03a` (2026-07-29T13:10:44Z) — verdict COMMENT

- ✅ **Fixed:** copy Job now bounded by `activeDeadlineSeconds` (4h), so a never-terminal pod fails into the retry path instead of gating the agent forever (`storage_migration.go:660`)
- ✅ **Fixed:** target PVC + Job owner-referenced to the Agent CR — mid-migration deletion GCs them instead of leaking a half-filled volume (`storage_migration.go:381,651`)
- ✅ **Fixed:** copy runs as the agent uid with a root init container chowning only the empty target root — works on root-squashing NFS without ever showing uid 0 to the source share
- 🔁 **Still present:** copy-Job inventory lists only files and symlinks, so an empty-directory mismatch goes undetected (`storage_migration.go:596`)
- 🔁 **Still present:** `md5sum` content verification without a comment noting the choice (`storage_migration.go:599`)
- 🆕 **New:** the retry cleanup `find -delete` can no longer delete read-only directories now that the copy dropped from root to the agent uid (`storage_migration.go:588`)

### Findings
- 🔴 **Critical:** The retry path's `find "$dst" -mindepth 1 -delete` was written when the copy ran as **root**; commit `2c451fea` moved the copy to the agent uid, and unlike reads, *deletes* are not covered by the "single-owner workspace" argument. Removing an entry requires **write+execute on its parent directory**, so any mode-`0500`/`0555` directory in the workspace — `cp -a` faithfully reproduces those modes on pass 1 — makes pass 2 fail. Reproduced as uid 65532: `find dst -mindepth 1 -delete` → `Permission denied` / `Directory not empty`, exit 1, tree intact. The `prep-target` init container does not mitigate it: it chowns only the mount **root**, non-recursively (`storage_migration.go:645`), and a chowned root still does not permit deletion inside a `0500` subdirectory. With `BackoffLimit: 2` and `RestartPolicy: OnFailure`, the first attempt's failure is retried against a non-empty target and every retry dies in cleanup — the Job exhausts its backoff, the agent stays gated behind the migration, and the 10-minute bounded retry re-creates the same doomed Job. Read-only directories are common in real workspaces (Python `site-packages`, Go module cache, pnpm store). Fix: `chmod -R u+rwX "$dst"` before the delete, or have the root init container empty the target (it already mounts exactly the targets), or copy into a fresh subdirectory per attempt. (`packages/controller/pkg/reconciler/storage_migration.go:588`)

Verdict: **minor drift**

🔁 1 finding from the previous review still present (see review at `78fe03a`) — narrowed: of the three pages flagged then, `observability.md` was bumped to 2026-07-29; two remain stale:
- `docs/architecture/platform-topology.md` (2026-07-24) — Fork→Run in prose, dropped the `| Fork |` resource row.
- `docs/architecture/budgets.md` (2026-07-27) — per-turn Fork/Run → per-command Run pods.

⚠️ **Possible drift (check 7):** ADR-079 supersedes live records 027 and 076; its `subsystem: channels` page may still describe the superseded state — worth a re-fold look. `persistence.md` (ADR-080's subsystem) is thoroughly updated for the RWO cutover.

✅ No findings. (61 files reviewed.) The `connectSlack(id, channel, mode, ambient)` → `(id, channel, ambient)` signature change is threaded consistently through contract types, router, service, CLI service, and all call sites — no caller left passing `mode` positionally into `ambient`. Fork/allowed-users removal is dangle-free: no surviving imports, event-registry entries, sagas, or `userDirectory` references; the `agentConnectSlackInputSchema` cross-field `.refine()` was correctly dropped along with the mode it guarded. Migration renumbering during the rebase (0018 → 0017) is clean — main ends at 0016.

✅ No findings. (2 files reviewed.) Form reduces to two fields with the schema still the source of truth and `defaultValues` explicit; `useWatch` removed with the conditional it drove; `AccessModePicker`/`AllowedUsers` deleted with all their imports (`useFieldArray`, `useState`, `Add`/`Close`, `SectionLabel`, `cn`), no dead imports left; `findSlackChannel` is still used in the card. State lineage is unchanged and correct — server state stays in TanStack Query, the one mutable toggle is form-local.

### Verdict
REQUEST_CHANGES — The three findings from the last review are properly fixed, but the uid change introduced a new critical: the retry path's `find -delete` cannot remove read-only directories as the agent uid, so any failed copy attempt turns into a permanently stuck migration.

---

tomkis (approved):

jezekra1:
## Full clean-cycle rehearsal #3 — permission fidelity, verified live

Previous release (latest `main`, b1b64178) installed fresh on a **real kernel-NFS RWX class with `root_squash`** → agent created with a permission zoo in its workspace: `0600` secret, `0700` dir, `0555` read-only dir with content, **`4755` suid file**, hardlink pair, live + dangling symlinks, an empty dir, and a deliberately owner-unreadable `0000` file → in-place upgrade to PR head (2bec0ec4).

Observed, in order:

1. **Preflight blocked the drain** — three fresh-pod attempts (restartPolicy `Never`), each failing fast with the actionable message naming `./work/blocked.bin`; agent stayed gated, source untouched.
2. **Operator remediation** (server-side `chmod 644`) + failed-Job delete → retry drained the volume in 30s gate-to-flip.
3. **Post-migration volume is line-for-line identical to the recorded baseline** — the same `type|mode|uid|path|linktarget` listing the in-Job metadata verification gates on, plus content checksums: suid bit preserved (group-0 copy — GNU cp strips suid when ownership preservation fails, which the metadata gate would reject), `0555`/`0700`/`0600` modes intact, hardlinks still linked, dangling symlink reproduced, empty dir present, all uid 65532. The one diff is the remediated blocker itself (`0000` → `644`), content intact.

This exercises the two newest commits: the metadata verification pass (**the RWX source is not deleted if permissions differ**, not just content) and the owner-unreadable preflight. Verification cost note: the metadata pass is stat-only; the byte-reading double checksum dominates and predates it. Default fleet concurrency raised to 10 per review discussion.

PR #3062, opened 2026-07-29
fix(ui): show private registry section whenever Custom is selected

jenna-winkler:
## Problem

On the Custom image step, the collapsed "Private registry" trigger only rendered once the image address field had at least one character. Typing a single letter made it appear; clearing the field made it vanish.

The gate was `isCustomImage`, which required a non-empty image:

```ts
const isCustomImage = !snapshot.templateId && snapshot.customImage.trim().length > 0;
```

Two problems with that:

- The section is collapsed anyway, so hiding it buys no simplicity while making the layout shift as you type.
- It hides the information at the moment it is most useful. If you are about to paste a private image address, knowing credentials are supported is part of deciding what to paste.

## Change

Gate on Custom being the starting point instead:

```ts
const isCustomImage = snapshot.startingPoint === "custom";
```

The section stays collapsed by default, so this only affects whether the one-line trigger is present. `startingPointDefaults` nulls `templateId` on the custom path, so the dropped `!snapshot.templateId` check was redundant.

Also fixes an edge case by construction: credentials live in ephemeral state rather than the persisted snapshot, and `registryPartial` blocks advancing when 1 or 2 of 3 fields are filled. Previously the section could unmount while that state survived, so partially filled credentials could block Continue with no visible fields to correct. Always rendering the section makes that unreachable.

## Notes

`finish()` still guards submission on `!!image` independently, so widening this does not let credentials be sent without an image address.

Not typechecked locally, leaving that to CI.

### DIFF
--- a/packages/ui/src/modules/sandboxes/components/steps/custom-image-card.tsx
+++ b/packages/ui/src/modules/sandboxes/components/steps/custom-image-card.tsx
@@ -17,7 +17,7 @@ interface Props {
   selected: boolean;
   onChange: (value: string) => void;
   onSubmit: () => void;
-  /** Pull-credential controls; present once a custom image is entered. */
+  /** Pull-credential controls; present whenever Custom is the starting point. */
   registry?: RegistryControls;
 }
 
--- a/packages/ui/src/modules/sandboxes/views/sandbox-wizard-view.tsx
+++ b/packages/ui/src/modules/sandboxes/views/sandbox-wizard-view.tsx
@@ -63,8 +63,7 @@ export function SandboxWizardView() {
   const [registryCredential, setRegistryCredential] = useState(
     EMPTY_REGISTRY_CREDENTIAL,
   );
-  const isCustomImage =
-    !snapshot.templateId && snapshot.customImage.trim().length > 0;
+  const isCustomImage = snapshot.startingPoint === "custom";
 
   const imageLabel = useMemo(() => {
     // The image is pinned and hidden on these paths — name the kind instead.


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `90d3dfd`

## PR #3062: fix(ui): show private registry section whenever Custom is selected
**Author:** jenna-winkler | **Branch:** fix/registry-section-always-visible → main | **Changes:** +2 −3 (2 files)

### Summary
Gates the collapsed private-registry trigger on `snapshot.startingPoint === "custom"` instead of requiring a non-empty image, so the section no longer appears/disappears as the image field is typed into or cleared. Also removes a latent trap where partially-filled ephemeral credentials could block Continue after the section unmounted.

### Findings
- ✅ **Looks good:** `isCustomImage` is a correctly derived value (not duplicated state), and `startingPointDefaults` nulls `templateId` on the custom path, so dropping the `!snapshot.templateId` conjunct is genuinely redundant, not a behavior change (`sandbox-wizard-view.tsx:66`).
- ✅ **Looks good:** Submission safety is preserved — `finish()` still gates on `!!image` and `registryFilledCount(...) === 3` before sending a registry credential, so widening the section's visibility can't leak credentials without an image (`sandbox-wizard-view.tsx:125-127`).
- ✅ **Looks good:** `registryPartial` remains scoped by `isCustomImage`, so the Continue-blocking guard now only engages on the custom path where the fields are actually reachable — which is exactly the edge case the PR closes (`sandbox-wizard-view.tsx:118-120`).
- 🟢 **Suggestion:** The JSDoc on the `registry` prop was updated to match (`custom-image-card.tsx:20`); nice to keep the doc in step with the gate.

✅ No findings. The architecture pages that mention private-registry credentials (`agent-lifecycle.md`, `security-and-credentials.md`) document system-level Secret creation and kubelet consumption, none of which this UI-visibility change touches.

✅ No findings. State lineage is respected — `registryCredential` stays ephemeral `useState` (secret, correctly never persisted to the snapshot), `isCustomImage` is derived rather than stored, and the change is minimal and idiomatic.

### Verdict
APPROVE — a small, well-reasoned visibility fix that also removes a latent stuck-Continue trap, with submission-time credential guards left intact.

---

PetrBulanek (approved):

PR #3072, opened 2026-07-30
fix(ui): stop the session-list error toast on the sandbox startup screen

### LINKED ISSUE #3047: UI flashes "Couldn't refresh session list" when opening a sandbox that was asleep
## Problem

Opening a sandbox in chat while it was **asleep** reliably flashes a warning toast **"Couldn't refresh session list"** during startup. The agent then comes up fine — the toast is a false alarm from probing the pod before it's actually reachable.

## Root cause (confirmed by code inspection)

The session-list query gates on the **raw server run-state**, which the code itself documents as lagging reality, rather than on the codebase's designated pod-reachability gate.

- The toast originates from `useAcpSessions` — `meta: { errorToast: "Couldn't refresh session list" }` (`packages/ui/src/modules/sessions/api/queries.ts:142`), emitted by the query cache `onError` (`packages/ui/src/query-client.ts:35-46`).
- In the sidebar the query is enabled by `agentRunState === "running"` (`packages/ui/src/modules/sessions/components/sessions-sidebar.tsx:69`). `useAgentRunState` returns the raw server `agent.state`, whose own doc-comment says it "is the lifecycle truth, but it **lags reality**" (`packages/ui/src/modules/agents/api/queries.ts:75`).
- On wake, the server flips `agent.state -> running` before the agent-runtime ACP endpoint is actually serving. `listAgentSessions` -> `withConnection` -> `conn.listSessions` fails (`packages/ui/src/modules/sessions/api/acp-session-ops.ts:73`), the query retries 3x (default `retry: 3`, `query-client.ts:50`), then toasts.
- The codebase already has a stronger gate for exactly this: `useIsAgentOperable` (`agents/api/queries.ts:82-98`), described as "the single gate for pod-touching calls," which additionally accounts for the reachability circuit-breaker and optimistic transitions. The session-list callers use the weaker `=== "running"` instead.
- The only existing toast suppression is for **api-server** health `reconnecting` (`query-client.ts:40`) — not agent-wake transients — so it does not catch this case.

This failure mode was already known: `use-agent-greeting.ts:28` comments that it gates on running "so it neither hits a booting pod nor toasts a session-list error" — but `running` alone is not a strong enough guard, because that state precedes actual ACP reachability.

## Repro

1. Let a sandbox go to sleep (hibernate).
2. Open it in chat.
3. During startup, a warning toast "Couldn't refresh session list" appears; the sandbox then loads normally.

## Expected

Waking a sandbox should not toast a session-list error. Either gate the session-list fetch on genuine pod reachability (not just server `running`), and/or suppress this query's `errorToast` while the agent is still transitioning to reachable (treat agent-wake like the existing api `reconnecting` suppression), and/or don't surface the first transient failure right after a `running` transition.

## Notes / candidate directions

- Switch the `useAcpSessions` `enabled` gate (sidebar + greeting + docked-experiment callers) from `runState === "running"` to the reachability-aware `useIsAgentOperable`. Note this alone may still let the *first* call on wake fail once before the circuit-breaker opens, so pair it with transient-suppression.
- Add per-agent "waking/unreachable" awareness to the query-cache `onError` toast suppression, mirroring the existing `getApiHealthSnapshot() === "reconnecting"` guard.
- Related theme: #2903 / #2970 ("running" shown before fully ready). Session/ACP calls hit the agent-runtime harness port directly (not the egress gateway), so this is about pod/relay readiness rather than gateway readiness — but it's the same class of "state says running before it's actually serviceable."

kapetr:
Closes #3047

## Problem

Opening a sandbox that was asleep reliably flashed a **"Couldn't refresh session list"** warning toast on the startup screen — before the chat or session sidebar had even rendered.

## Cause

`useDockedExperiment` is a hook called at the top of `chat-view.tsx`, so it runs the moment chat opens (during the startup screen), independent of the sidebar. It calls `useAcpSessions` **with no readiness gate**, and that query shares the sidebar's cache key. React Query runs a query if *any* observer enables it — so the ungated docked observer forced the shared session-list query to fetch into a still-waking pod, overriding the sidebar's own `running` gate, and toasted when the ACP connection wasn't up yet.

## Fix

- Gate the docked-experiment observer on `runState === "running"`, matching the sidebar. Now every observer of that shared query waits for the agent to be up, so nothing fetches into a booting pod.
- Show a `SessionListSkeleton` (three rows mirroring `SessionRow`'s two-line title/timestamp layout and height) while the list first loads, instead of flashing an empty list.

UI-only; no api-server/controller/docs changes.

## Test plan

- `mise run ui:check` and `mise run ui:test` pass
- Manually: open a sleeping sandbox → startup screen no longer toasts; the sidebar shows skeleton rows, then the real sessions

https://claude.ai/code/session_01NeZeHv1xtCW4GXVsWbA9tC

### DIFF
--- a/packages/ui/src/modules/experiments/hooks/use-docked-experiment.ts
+++ b/packages/ui/src/modules/experiments/hooks/use-docked-experiment.ts
@@ -2,6 +2,7 @@ import type { Experiment } from "api-server-api";
 import { useCallback, useMemo, useRef, useState } from "react";
 
 import { useStore } from "../../../store.js";
+import { useAgentRunState } from "../../agents/api/queries.js";
 import { useAcpSessions } from "../../sessions/api/queries.js";
 import { useAgentExperimentsLive } from "../api/queries.js";
 
@@ -45,7 +46,13 @@ export function useDockedExperiment(agentId: string | null): {
     }),
     [sessionFilter],
   );
+  // Same gate as the sidebar observer of this shared query: a waking pod isn't
+  // reachable over ACP yet, and this hook runs on chat open before the sidebar
+  // renders — without the gate it fetches the session list into a booting pod
+  // and toasts "Couldn't refresh session list" on the startup screen.
+  const runState = useAgentRunState(agentId);
   const { data: sessions } = useAcpSessions(agentId, listInclude, {
+    enabled: runState === "running",
     activeSessionId: sessionId,
   });
   const sessionExperimentId =
--- a/packages/ui/src/modules/experiments/hooks/use-docked-experiment.ts
+++ b/packages/ui/src/modules/experiments/hooks/use-docked-experiment.ts
@@ -2,7 +2,7 @@ import type { Experiment } from "api-server-api";
 import { useCallback, useMemo, useRef, useState } from "react";
 
 import { useStore } from "../../../store.js";
-import { useAgentRunState } from "../../agents/api/queries.js";
+import { useIsAgentOperable } from "../../agents/api/queries.js";
 import { useAcpSessions } from "../../sessions/api/queries.js";
 import { useAgentExperimentsLive } from "../api/queries.js";
 
@@ -46,13 +46,15 @@ export function useDockedExperiment(agentId: string | null): {
     }),
     [sessionFilter],
   );
-  // Same gate as the sidebar observer of this shared query: a waking pod isn't
-  // reachable over ACP yet, and this hook runs on chat open before the sidebar
-  // renders — without the gate it fetches the session list into a booting pod
-  // and toasts "Couldn't refresh session list" on the startup screen.
-  const runState = useAgentRunState(agentId);
+  // Shares the session-list query key with the sidebar and greeting observers,
+  // and React Query fires the query if ANY observer enables it — so all three
+  // gate on the same operability check (the canonical gate for pod-touching
+  // reads). Without it, this hook — which runs on chat open before the sidebar
+  // renders — fetches into a booting or restarting pod and toasts
+  // "Couldn't refresh session list".
+  const operable = useIsAgentOperable(agentId);
   const { data: sessions } = useAcpSessions(agentId, listInclude, {
-    enabled: runState === "running",
+    enabled: operable,
     activeSessionId: sessionId,
   });
   const sessionExperimentId =
--- a/packages/ui/src/modules/sessions/components/sessions-sidebar.tsx
+++ b/packages/ui/src/modules/sessions/components/sessions-sidebar.tsx
@@ -223,6 +223,7 @@ export function SessionsSidebar({
       style={style}
     >
       <div className="flex-1 overflow-y-auto">
+        {loading && <SessionListSkeleton />}
         {!loading && sessions.length === 0 && (
           <p className="px-4 py-5 text-[12px] text-text-muted">
             No sessions yet
@@ -255,5 +256,23 @@ export function SessionsSidebar({
         {runSessions.map(renderRow)}
       </div>
     </SidebarSection>
+  );
+}
+
+/** Placeholder rows while the session list first loads, matching SessionRow's
+ *  layout so the list doesn't flash empty before the agent answers. */
+function SessionListSkeleton() {
+  return (
+    <div aria-hidden>
+      {Array.from({ length: 3 }).map((_, i) => (
+        <div
+          key={i}
+          className="flex items-center gap-2 border-b border-border-light px-4 py-3"
+        >
+          <span className="h-2 w-2 shrink-0 rounded-full bg-muted animate-pulse" />
+          <span className="h-3 w-1/2 rounded bg-muted animate-pulse" />
+        </div>
+      ))}
+    </div>
   );
 }
--- a/packages/ui/src/modules/sessions/components/sessions-sidebar.tsx
+++ b/packages/ui/src/modules/sessions/components/sessions-sidebar.tsx
@@ -22,6 +22,7 @@ import {
   SESSION_CATEGORY_LABELS,
   sessionCategory,
 } from "../lib/session-category.js";
+import { SessionListSkeleton } from "./session-list-skeleton.js";
 import { SessionRow } from "./session-row.js";
 import { SidebarSection } from "./sidebar-section.js";
 
@@ -256,29 +257,5 @@ export function SessionsSidebar({
         {runSessions.map(renderRow)}
       </div>
     </SidebarSection>
-  );
-}
-
-/** Placeholder rows while the session list first loads. Mirrors SessionRow's
- *  height and its two lines — title, then a smaller timestamp — so the list
- *  doesn't flash empty before the agent answers. The text-size spans set the
- *  same line boxes as the real rows, so the row height matches exactly. */
-function SessionListSkeleton() {
-  return (
-    <div aria-hidden>
-      {Array.from({ length: 3 }).map((_, i) => (
-        <div
-          key={i}
-          className="flex flex-col gap-0.5 border-b border-border-light px-4 py-3"
-        >
-          <span className="text-[13px]">
-            <span className="inline-block h-3 w-1/2 rounded bg-muted align-middle animate-pulse" />
-          </span>
-          <span className="text-[11px]">
-            <span className="inline-block h-2.5 w-1/4 rounded bg-muted align-middle animate-pulse" />
-          </span>
-        </div>
-      ))}
-    </div>
   );
 }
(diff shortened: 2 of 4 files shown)


packages/ui/src/modules/sessions/components/sessions-sidebar.tsx
```diff
@@ -257,3 +258,27 @@ export function SessionsSidebar({
     </SidebarSection>
   );
 }
+
+/** Placeholder rows while the session list first loads. Mirrors SessionRow's
+ *  height and its two lines — title, then a smaller timestamp — so the list
+ *  doesn't flash empty before the agent answers. The text-size spans set the
+ *  same line boxes as the real rows, so the row height matches exactly. */
+function SessionListSkeleton() {
```
dam-code-guardian: 🟢 **Suggestion:** this is the third instance of the line-box skeleton pattern in the codebase — `SkillRowsSkeleton` (`skills/skills-skeleton.tsx`) uses the identical technique, and `skill-render-modal.tsx:81-84` does it again. Worth extracting to a sibling `session-list-skeleton.tsx` (matching `session-row.tsx`) rather than inlining it here: it keeps `sessions-sidebar.tsx` from growing, and the react-ui guidance calls inlined subcomponents an anti-pattern to fix on sight.

Not blocking — the component itself is correct.

packages/ui/src/modules/sessions/components/sessions-sidebar.tsx
```diff
@@ -257,3 +258,27 @@ export function SessionsSidebar({
...
+  return (
+    <div aria-hidden>
+      {Array.from({ length: 3 }).map((_, i) => (
+        <div
+          key={i}
+          className="flex flex-col gap-0.5 border-b border-border-light px-4 py-3"
+        >
+          <span className="text-[13px]">
+            <span className="inline-block h-3 w-1/2 rounded bg-muted align-middle animate-pulse" />
```
dam-code-guardian: 🟢 **Suggestion:** `animate-pulse` sits on each inner bar, so the two bars are independently animated elements that merely start in phase. Both existing skeletons put the pulse on the container instead, which guarantees they stay in phase and drops the duplication:

```suggestion
            <span className="inline-block h-3 w-1/2 rounded bg-muted align-middle" />
```

(paired with `animate-pulse` moved onto the `<div aria-hidden>` wrapper, and the same class dropped from the timestamp bar below).

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `02f441cd`

### Summary
Gates the docked-experiment observer of the shared ACP session-list query on `runState === "running"`, so an ungated observer can no longer force a fetch into a still-waking pod and toast "Couldn't refresh session list" on the startup screen. Also adds a `SessionListSkeleton` shown while the list first loads.

### Findings
- 🟢 **Suggestion:** `SessionListSkeleton` is a subcomponent inlined in its parent file, and it duplicates a pattern that already exists twice in the codebase — `SkillRowsSkeleton` (`packages/ui/src/modules/sandboxes/components/skills/skills-skeleton.tsx`) uses the identical "bar inside a same-font-size line box" technique, and `ListSkeleton` (`packages/ui/src/components/list-skeleton.tsx`) is the generic primitive. Third occurrence of the pattern is the extract point; a sibling `session-list-skeleton.tsx` (matching `session-row.tsx`) would also keep `sessions-sidebar.tsx` from growing further. (`packages/ui/src/modules/sessions/components/sessions-sidebar.tsx:266`)
- 🟢 **Suggestion:** `animate-pulse` is repeated on both inner bars rather than set once on the wrapping `<div aria-hidden>`. Both existing skeletons put the pulse on the container, which also keeps the two bars' animations in phase — as written they are independently animated elements that happen to start together. (`packages/ui/src/modules/sessions/components/sessions-sidebar.tsx:275`)
- ✅ **Looks good:** The root-cause analysis is correct and the fix is at the right level. React Query enables a query if *any* observer enables it, so the sidebar's `enabled: agentRunState === "running"` gate (`sessions-sidebar.tsx:68`) was genuinely being overridden by the ungated docked observer sharing the same `acpSessionsKeys.list(agentId, include)` key. Gating the second observer identically is the minimal correct fix, not a symptom patch.
- ✅ **Looks good:** `useAcpSessions` implements `enabled: false` via `skipToken` and `refetchInterval: live ? … : false` (`sessions/api/queries.ts:113-138`), so the gate genuinely stops both the initial fetch and the poll — and keeps the cache entry, so no data is discarded while waking. The `listInclude` memo is preserved, so the docked observer stays a cache hit on the sidebar's key rather than opening a second one.
- ✅ **Looks good:** Consistent with the established convention — `use-agent-greeting.ts:35-46` gates the same query on the same `useAgentRunState(agentId) === "running"` condition, and `useAgentRunState`'s own doc comment names `=== "running"` as *the* gate for pod-touching reads. This PR makes the third observer conform.
- ✅ **Looks good:** The skeleton is correctly scoped to first load only — `loading = data === undefined && isFetching` (`sessions-sidebar.tsx:75`) means it never reappears on the few-second poll, and the existing "No sessions yet" / filter-empty branches were both re-gated on `!loading` so they can't render simultaneously.
- ✅ **Looks good:** `aria-hidden` on the placeholder is the right call, and the row markup mirrors `SessionRow`'s actual structure (`gap-0.5`, `border-b border-border-light px-4 py-3`, `text-[13px]` title over `text-[11px]` timestamp — `session-row.tsx:124-144`), so the height claim in the comment holds and the list won't reflow.

**Verdict: aligned**

✅ No findings. The diff changes no subsystem's behavior or responsibility — the pod-readiness contract it honors is already described in `docs/architecture/agent-lifecycle.md` (§Wake: "the controller-published `Ready` condition is the authoritative answer to 'can I call this pod?'") and is unchanged by this PR; a client observer being brought into line with an existing contract is not architectural drift. No architecture page was edited (so no `Last verified:` bump is owed), no new subsystem was introduced, and no ADR changed state.

✅ No findings.

The only change to `use-docked-experiment.ts` is a new import of `useAgentRunState` from `../../agents/api/queries.js` plus an `enabled` option. TSEng's module-boundary rules govern server-side bounded contexts with `services`/`domain`/`infrastructure` layers and an `index.ts` public API; the UI package has no such layering — no `packages/ui/src/modules/*/index.ts` boundary exists — and cross-module query-hook imports are the pervasive established pattern (64 imports from `agents/` across the UI modules, including the sibling `use-agent-greeting.ts`, which imports this exact pair of hooks). Per review-mode scope, only violations introduced or amplified by the change are in scope; this introduces none.

1 finding.

- 🟢 **Suggestion** — *Anti-patterns to fix on sight #1: "Subcomponents inlined in the same file as the parent"*, reinforced by *Core principle 1: "When the same pattern appears three times, extract it."* `SessionListSkeleton` is defined inside `sessions-sidebar.tsx:266` and is the third instance of the line-box skeleton pattern (`skills-skeleton.tsx`, `skill-render-modal.tsx:81-84`, now this). Extract to a sibling `session-list-skeleton.tsx`.

Checked and clean: no static `style={}` in the added markup; Tailwind-only with theme tokens (`bg-muted`, `border-border-light` — both defined in `App.css` and light/dark-aware, so no `dark:` variant is owed); no template-literal class strings; the `.map` body is 9 lines of static JSX with no conditionals, handlers, or derivations, so the list-item rule doesn't bite; no new state, no new effects, and no server state duplicated into `useState` — the loading flag is derived from the query, correctly.

### Verdict
**APPROVE** — The diagnosis is precise and the fix is the minimal, convention-consistent one; only two non-blocking suggestions about where the new skeleton lives.

---

packages/ui/src/modules/experiments/hooks/use-docked-experiment.ts
```diff
@@ -45,7 +46,13 @@ export function useDockedExperiment(agentId: string | null): {
     }),
     [sessionFilter],
   );
+  // Same gate as the sidebar observer of this shared query: a waking pod isn't
+  // reachable over ACP yet, and this hook runs on chat open before the sidebar
+  // renders — without the gate it fetches the session list into a booting pod
+  // and toasts "Couldn't refresh session list" on the startup screen.
+  const runState = useAgentRunState(agentId);
   const { data: sessions } = useAcpSessions(agentId, listInclude, {
+    enabled: runState === "running",
```
PetrBulanek: What about using `useIsAgentOperable` here instead `runState === "running"`?

PetrBulanek (approved):

packages/ui/src/modules/experiments/hooks/use-docked-experiment.ts
kapetr: I considered that, but the same query is already gated on the same (looser) condition, so I matched it. But it's worth confirming if the gate really is sufficient.

kapetr: Updated all queries to `useIsAgentOperable`, to make it robust.

PR #3074, opened 2026-07-30
feat(artifacts): agents can download artifacts into their sandbox

### LINKED ISSUE #3073: Agents can download artifacts into their sandbox
## Context

Agents publish work products to the Artifact Library, and experiment runs attach their outputs — datasets, results files — there. The library's agent-facing surface was built publish-first: agents can create, version, organize, and share artifacts, and read back small text ones.

## Problem

An agent has no way to get an artifact's file content back into its sandbox. Reading an artifact returns its content inline only for small text files; anything binary or large is omitted with a pointer to "use the UI" — which an agent can't do. Returning bigger content inline isn't an option either: tool output flows through the agent's context, so real files can't travel that way.

Concrete scenario: an experiment run produces a dataset and publishes it as an artifact. A later run — or a different experiment entirely — wants to build on that data. Today the bytes are stranded: the follow-up agent can see the artifact exists but can't fetch it, so data can't flow between experiments. The workaround is keeping everything inside one sandbox, which defeats the point of publishing artifacts at all.

## Goal

An agent can pull any artifact from its owner's library into its sandbox — text or binary, current or a specific version — as easily as it can publish one. Experiments can consume prior experiments' outputs.

## User Stories

- As an experiment author, I want a run to download artifacts produced by earlier runs so that experiments can build on each other's data instead of starting from scratch.
- As a user, I want to point any of my agents at an artifact in my library so that work products created in one sandbox are usable in another.

## Proposed solution

Mirror the upload path. Uploads already avoid the context problem: the agent mints a short-lived direct-transfer link and sends raw bytes straight to storage, so nothing passes through the conversation. Add the same in the other direction — a tool that mints a short-lived download link for an artifact (optionally a specific version), which the agent fetches into its sandbox with ordinary HTTP. Inline reads of small text artifacts stay as they are; the download link covers binaries and anything large. The platform already mints download links of this kind for browser downloads, so this extends an existing mechanism rather than adding a new one.

jezekra1:
## Summary

Closes #3073.

Agents can now pull any artifact from their owner's library into their sandbox — text or binary, current or a past version — mirroring the direct-transfer upload path so bytes never transit the conversation.

- **New MCP tool `create_artifact_download_url`**: takes an artifact id and optional version, returns a short-lived presigned GET URL plus file name, content type, size, resolved version, expiry, and a curl example. `get_artifact` now points at it for binary/large content instead of saying "use the UI".
- **Audience-aware presigned downloads**: SigV4 binds the Host header, and agents dial the store through their gateway while browsers use the public endpoint. `presignDownload` now takes an `audience` (`agent` | `browser`) and the S3 adapter's signers are renamed to `agentSigner`/`browserSigner`. Agent links always work when a store is configured, even on installs with no browser-reachable endpoint.
- **Service layer**: `ArtifactService.createAgentDownloadUrl` uses a 15-minute TTL (matching uploads — an agent may run other tool calls before fetching). The library-service mint lives on the server-internal `ArtifactLibraryServiceImpl` surface, owner-scoped, failing closed when no object store is configured.
- **Audit + hygiene**: each mint emits the same `artifact_library.download` security-log line as browser downloads; the Content-Disposition filename sanitizer is shared between the browser route and the agent path.
- **Docs**: artifact-library architecture page documents the download direction.

## Test plan

- Unit tests: signer selection per audience, agent-links-work-without-public-endpoint, TTLs, past-version resolution, owner-scoping rejection, fail-closed without a store
- `mise run check` and `mise run api-server:test` pass (756 tests)


packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts
```diff
@@ -452,6 +476,36 @@ export function createArtifactLibraryService(
         fileName: row.fileName,
```
dam-code-guardian: 🟡 **Warning:** On the past-version branch this returns the **head row's** `fileName` while the bytes come from `past.storageRef` — so a v1 download of an artifact renamed at v2 is served under v2's name. `library_artifact_versions` has no `fileName` column, so the version's real name survives only in its `storageRef` basename.

Pre-existing on the browser route, but this PR is what exposes it to agents via a documented "fetch a past version" affordance, and the ticket's new `version` field makes the inconsistency visible to the caller. Deriving the name from `past.storageRef` (or adding the column) would make the ticket self-consistent.

packages/api-server/src/modules/artifact-library/mcp-tools.ts
```diff
@@ -163,6 +172,44 @@ export function registerArtifactLibraryTools(
...
+        .positive()
+        .optional()
+        .describe("Version to download; omit for the current one."),
+    },
+    ({ id, version }) =>
+      run(async () => {
+        const ticket = await lib.createAgentDownloadUrl(id, version);
+        // Mirrors the browser download route's audit line — a presigned link
+        // is a bearer capability to the bytes.
+        securityLog("info", "artifact_library.download", {
```
dam-code-guardian, line 209: 🟡 **Warning:** `result: "success"` is recorded unconditionally, but this line only runs on the success path — a failed mint (unknown id, version not found, no object store) throws inside `run()` and is logged nowhere. Because a presigned link is a bearer capability to the bytes, the *denied* attempts are the ones worth an audit trail: an agent probing ids outside its owner's library currently leaves none.

Consider wrapping the mint so a rejection emits `securityLog("warn", "artifact_library.download", { …, result: "failure", reason })` before rethrowing. (The browser route has the same shape, so this is inherited rather than introduced — but the agent surface is the one where probing is cheap.)

packages/api-server/src/modules/artifacts/services/artifact-service.ts
```diff
@@ -86,7 +97,19 @@ export function createArtifactService(deps: {
       return deps.store.presignDownload(key, {
         filename,
         expiresSeconds: DOWNLOAD_URL_TTL_SECONDS,
+        audience: "browser",
+      });
+    },
+
+    async createAgentDownloadUrl(key, filename) {
```
dam-code-guardian, line 107: 🟢 **Suggestion:** Unlike the upload direction — which enforces `maxBytes` via `verifyUpload` — the download path applies no size policy. That is almost certainly correct (the blob is already stored, therefore already under the cap), but the asymmetry is easy to misread as an oversight. One line noting *why* there is no cap here would stop a future reader from "fixing" it and breaking legitimate large downloads.

packages/api-server/src/modules/artifact-library/mcp-tools.ts
```diff
@@ -163,6 +172,44 @@ export function registerArtifactLibraryTools(
...
+          result: "success",
+          detail: {
+            mode: "direct",
+            version: ticket.version,
+            sizeBytes: ticket.sizeBytes,
+          },
+        });
+        return json({
+          ...ticket,
+          instructions: `GET the URL and save the body, e.g.: curl -fL -o '${shellSafeFileName(ticket.fileName)}' '${ticket.url}' — the link expires in ${ticket.expiresSeconds}s.`,
```
dam-code-guardian: 🟢 **Suggestion:** `shellSafeFileName` maps every non-`[A-Za-z0-9._-]` character to `_`, so `résumé.pdf` renders here as `r_sum_.pdf`. An agent that copies the curl line verbatim lands the bytes under a mangled name while the authoritative `fileName` field alongside it carries the correct one — a quiet divergence rather than an error.

The quoting hardening itself is right; worth one clause in the string saying the example's `-o` name is sanitized and `fileName` is authoritative:

```suggestion
          instructions: `GET the URL and save the body, e.g.: curl -fL -o '${shellSafeFileName(ticket.fileName)}' '${ticket.url}' — the -o name is sanitized for shell safety; the fileName field is authoritative. The link expires in ${ticket.expiresSeconds}s.`,
```

packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts
```diff
@@ -452,6 +476,36 @@ export function createArtifactLibraryService(
...
         contentType: past.contentType,
         sizeBytes: past.sizeBytes,
+        version: past.version,
+      };
+    },
+
+    async createAgentDownloadUrl(id, version) {
+      const ref = await this.resolveContentRef(id, version);
```
dam-code-guardian: 🟢 **Suggestion:** Calling `this.resolveContentRef` makes the returned object's own method dispatch part of the service's contract: destructuring (`const { createAgentDownloadUrl } = lib`) silently breaks it, and tests must stub the whole object rather than the one collaborator. Two sibling methods already do this, so the pattern is inherited — but extracting the shared resolution into a module-scope helper that all three call directly would keep the factory's return value a plain façade, per the application-layer rule that services coordinate injected collaborators rather than themselves.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `c52d2563`

## PR #3074: feat(artifacts): agents can download artifacts into their sandbox
**Author:** jezekra1 | **Branch:** feat/agent-artifact-download#locki-9stt2amv → main | **Changes:** +327 −38 (12 files)

### Summary
Adds an agent-facing `create_artifact_download_url` MCP tool that mints a short-lived presigned GET so a harness can pull any library artifact — text or binary, head or a past version — straight into its sandbox. The storage port's `presignDownload` gains an `audience` (`agent` | `browser`) so SigV4 signs against the authority the caller actually dials, the signer clients are renamed `agentSigner`/`browserSigner` accordingly, and the Content-Disposition filename sanitizer moves into the library's domain layer where both the browser route and the agent path share it.

### Findings
- 🟡 **Warning:** `resolveContentRef` returns the **head row's** `fileName` for a past version (`fileName: row.fileName` on the past-version branch), so a v1 download whose file was later renamed at v2 gets v2's name — and, worse, the *name* is what's baked into `Content-Disposition` while the *bytes* come from v1's `storageRef`. `library_artifact_versions` has no `fileName` column, so the past version's real name is only recoverable from its `storageRef` basename. Pre-existing on the browser route, but this PR is what exposes it to agents through a documented "fetch a past version" affordance, and the new `version` field in the ticket makes the mismatch visible to the caller. Deriving the name from `past.storageRef` (or adding the column) would make the ticket self-consistent. (`packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts:476`)
- 🟡 **Warning:** The audit line records `result: "success"` unconditionally *before* the ticket is returned, but only ever runs on the success path — a failed mint (unknown id, no store configured) throws inside `run()` and is logged nowhere. The browser route has the same shape, so the agent surface inherits a blind spot: an agent probing ids it doesn't own produces no audit trail at all. Since a presigned link is a bearer capability, the denied attempts are the ones worth logging (`securityLog("warn", …, { result: "failure", reason })`). (`packages/api-server/src/modules/artifact-library/mcp-tools.ts:192`)
- 🟢 **Suggestion:** `sizeBytes` from the ticket flows into the audit `detail` unbounded, and `createAgentDownloadUrl` — unlike the upload path, which caps via `verifyUpload` — applies no size policy at all. That is arguably correct (the blob is already stored, so it is already under the cap), but the asymmetry with the upload direction is worth a one-line comment so a future reader doesn't "fix" it by adding a cap that breaks legitimate large downloads. (`packages/api-server/src/modules/artifacts/services/artifact-service.ts:104`)
- 🟢 **Suggestion:** `shellSafeFileName` collapses every non-`[A-Za-z0-9._-]` character to `_`, so a `résumé.pdf` artifact renders in the example as `r_sum_.pdf` — the agent copies the curl line and silently lands the bytes under a mangled name while the real `fileName` field carries the correct one. Worth stating in the `instructions` string that the example's `-o` name is sanitized and the `fileName` field is authoritative. (`packages/api-server/src/modules/artifact-library/mcp-tools.ts:208`)
- 🟢 **Suggestion:** `create_artifact_download_url` is now a first-class publish/consume affordance, but the sandbox Artifacts callout still lists only `create_artifact` and `create_artifact_upload_url` as "the built-in platform MCP tools". Not in this diff, so not blocking — flagging so the UI copy doesn't drift. (`packages/ui/src/modules/artifacts/components/sandbox-artifacts-section.tsx:44`)
- ✅ **Looks good:** The `audience` discriminator is the right cut. Making it a required field on the port (rather than a second method or an optional flag) means every existing call site had to state its audience, the compiler enforced it, and `presignDownload` returning null now means "this audience has no reachable authority" instead of the narrower "no public endpoint" — which is exactly why the agent path works on installs with no browser endpoint.
- ✅ **Looks good:** Moving `sanitizeFilename` → `downloadFileName` into `domain/artifact-kind.ts` puts the header-safety invariant in one place, and the library service applies it *before* handing the name to the store, so the sanitized name is what gets baked into `ResponseContentDisposition` rather than sanitized twice or not at all.
- ✅ **Looks good:** Test coverage tracks the risk well — signer selection per audience, agent-links-work-without-a-public-endpoint, past-version resolution, owner-scoping rejection routed through the new method, and fail-closed with no store. The header-injection case (`evil"name\n.html` → `evilname.html`) asserts on the sanitized name reaching the mint, not just on the return value.

**Verdict:** minor drift

**Drift**

- ❌ **Check 1 — architecture-page drift:** [`docs/architecture/persistence.md:12`](docs/architecture/persistence.md) describes the object store's link-minting authority as *"short-lived, single-object links that let an agent **upload** (through its paired gateway) or a **browser download** directly"*. This PR makes an agent-audience *download* link a first-class path (`presignDownload` now takes `audience: "agent" | "browser"`; `compose.ts` renames the signers to `agentSigner`/`browserSigner` precisely because both audiences now download), so the page's enumeration of who-may-do-what across that trust boundary is now incomplete. The artifact-library page was updated; this one was not. Proposed edit: widen to "…links that let an agent upload or download through its paired gateway, or a browser download directly". `Last verified:` bumps with it (2026-07-27 → the PR date). The page is well under the 40 000-char cap, so there is room.

`docs/architecture/artifact-library.md` is aligned: the publishing-paths section gained the read direction, the mermaid edge became `direct PUT/GET via gateway`, `Last verified:` was bumped to 2026-07-30, and the added prose stays at the semantic level (no type signatures, field names, or file paths) — no check-3 or check-4 leak. No architecture page was moved or deleted (check 6), no new subsystem was introduced (check 5), and no ADR changed state (check 7).

_Note: `security-and-credentials.md:97` already describes the ext_authz artifact-store carve-out generically ("a short-lived link scoped to a single object and operation"), so it covers the GET direction without an edit._

**Verdict:** aligned — one finding.

- 🟢 **Layer boundary — `services/` reaching for `this`** (`slice-composition.md` → Application Layer). `createAgentDownloadUrl` calls `this.resolveContentRef(id, version)` (`artifact-library-service.ts:484`), following the two existing `this.` call sites in the same object literal. It works, but it makes the service object's method dispatch part of its own contract: destructuring the service (`const { createAgentDownloadUrl } = lib`) silently breaks it, and the tests already have to stub around the whole object rather than the one collaborator. Extracting the shared resolution into a module-scope helper the three methods call directly would keep the factory's returned object a plain façade. Pre-existing pattern, amplified by one — hence 🟢, not 🟡.

Otherwise the change is well-placed against the architecture:
- The new port field lands in `artifacts/domain/artifact-store.ts` (a port defined by the inner layer), the SigV4 signer choice lands in `artifacts/infrastructure/s3-artifact-store.ts`, the TTL policy lands in `artifacts/services/artifact-service.ts`, and the signer wiring lands in `artifacts/compose.ts` — the only place concrete `S3Client`s are referenced. Dependency direction (`services → domain ← infrastructure`) holds throughout; no service imports an infrastructure implementation.
- `downloadFileName` is pure, dependency-free, and sits in `domain/` — correct for a value-object invariant, and it removes the duplicate that had drifted into `library-routes.ts`.
- `ArtifactAgentDownloadTicket` stays on the server-internal `ArtifactLibraryServiceImpl` rather than the contract package, which is right: no tRPC router exposes it, only the MCP surface, so it is not part of the published contract.
- Zod validation for the new tool's input sits at the MCP registration boundary — the analogue of the contract router for the agent surface — with no business logic in the handler beyond the audit line.
- Tests keep the `stubStore(overrides)` / `fakeRepo(rows)` factory pattern and pass real stub implementations rather than loosening deps to optional.

### Verdict
COMMENT — the audience split is a clean, well-tested design and the security-sensitive parts (owner scoping, header sanitization, fail-closed, shell-quoting of the example) are all handled deliberately; the two 🟡 items are a version/filename mismatch this PR newly exposes to agents and a mint-failure audit gap, neither blocking but both worth a look before merge.

---

dam-code-guardian:
📊 **PR #3074 review artifact** — [DAM Artifact Library](https://share-dam-dev.apps.dam-cl.fmaas.res.ibm.com/a/JjXTg1R-KEOa9Q)

packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts
```diff
@@ -112,6 +135,32 @@ function toFolder(row: FolderRow & { artifactCount: number }): ArtifactFolder {
...
 
+/** An artifact's file name and kind are settled at create and describe *every*
+ *  version: the versions table snapshots bytes (ref, type, size), never a name,
+ *  and the viewer renders each version through the artifact's kind. Letting
+ *  either move would retroactively relabel versions already published — an
+ *  older revision would download under a name it never had, or render through
+ *  the wrong renderer. A revision that is genuinely a different file is a
+ *  different work product: it belongs in its own artifact, with its own share
+ *  link. Editing the human-facing label is what `title` is for. */
+function requireStableIdentity(
```
dam-code-guardian: 🟢 **Suggestion:** `requireStableIdentity` is a pure, I/O-free function of two values encoding a business rule ("name and kind are settled at create") — an invariant, which the layering guidance puts in `domain/`. Worth noting that this same commit moved `downloadFileName` *out* of `library-routes.ts` and *into* `domain/artifact-kind.ts` for exactly this reason; this helper went the other way. Placing it beside `downloadFileName` (or in a new `domain/artifact-identity.ts`) would keep the two halves of the artifact-identity invariant together.

While there: the resolved-ref shape is now written out structurally twice — on the interface and again in the `resolveRef` closure below — where the sibling ticket got a named type. That shape is a domain value object too.

packages/api-server/src/modules/artifact-library/mcp-tools.ts
```diff
@@ -163,6 +172,61 @@ export function registerArtifactLibraryTools(
...
+    ({ id, version }) =>
+      run(async () => {
+        // A presigned link is a bearer capability to the bytes, so both
+        // outcomes are audited: logging only the successes would make a refused
+        // mint (an id the owner does not have) invisible in the trail.
+        let ticket;
+        try {
+          ticket = await lib.createAgentDownloadUrl(id, version);
+        } catch (err) {
+          securityLog("info", "artifact_library.download", {
```
dam-code-guardian: 🟢 **Suggestion:** this failure line is emitted at `info`. Every other `result: "failure"` site in the api-server uses `warn` (or `error` on an internal security-path failure), and both `core/security-log.ts` and `docs/architecture/logging.md` state the mapping explicitly as *deny/fail → `warn`*. As written the refused mint is recorded but invisible to an operator filtering the audit trail at `level>=warn` — which is most of the value the branch was added for.

```suggestion
          securityLog("warn", "artifact_library.download", {
```

Separately (no change needed here): both emissions sit in the tool wrapper rather than in `createAgentDownloadUrl`, and the browser route at `library-routes.ts:82` emits the *same* `artifact_library.download` event while still logging only successes — so the two surfaces have diverged on precisely the question this hunk's comment raises. Moving the audit into the service would cover both by construction.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `821817bf`

### Summary
Adds an agent-facing `create_artifact_download_url` MCP tool that mints a short-lived presigned GET so a harness can pull any library artifact into its sandbox. This re-review covers one new commit, `821817bf`, which settles an artifact's file name and kind at create time and closes out every finding from the previous review.

### Changes since last review
Previous HEAD: `c52d2563` (2026-07-30T12:58:15Z) — verdict COMMENT

- ✅ **Fixed:** the past-version/head-row filename mismatch is now resolved by making name and kind immutable for the artifact's whole life, so the head row's name is a truthful label for every version's bytes (`services/artifact-library-service.ts:146`)
- ✅ **Fixed:** refused download mints are now audited, not just successful ones (`mcp-tools.ts:196`)
- ✅ **Fixed:** the absent download-path size cap is now documented as deliberate (`services/artifact-service.ts:104`)
- ✅ **Fixed:** non-ASCII names survive the example command — `shellQuote` replaced `shellSafeFileName`, so the `-o` name is now exact instead of mangled (`mcp-tools.ts:39`)
- ✅ **Fixed:** the sandbox Artifacts callout now lists the download tool (`sandbox-artifacts-section.tsx:48`)
- ✅ **Fixed:** `this.resolveContentRef` dispatch replaced by the module-closure `resolveRef` (`services/artifact-library-service.ts:192`)

### Findings
- 🟡 **New:** the tRPC contract still advertises the two fields the service now refuses. `artifactUpdateInputSchema` keeps `fileName` and `kind` as optional input (`packages/api-server-api/src/modules/artifact-library/schemas.ts:67`) and `ArtifactUpdateInput` still declares them (`types.ts:107`), while `requireStableIdentity` throws `BAD_REQUEST` on any differing value. The `update` procedure spreads validated input straight through, so a UI or CLI client using a documented contract field gets a runtime 400 from the service rather than an answer at the validation boundary — the agent-facing tool was cleaned up, the contract was not. Dropping both fields from the schema and type (nothing calls `useUpdateArtifact` with them today) would make the invariant one statement instead of two disagreeing ones, and demote `requireStableIdentity` to defence-in-depth.
- 🟢 **New:** `requireStableIdentity` is a pure, I/O-free business invariant sitting in `services/` (`services/artifact-library-service.ts:146`) — the same PR moved `downloadFileName` *into* `domain/artifact-kind.ts` for exactly this reason. Placing it beside that helper would be consistent.
- 🟢 **New:** the failure audit line is emitted at `info` (`mcp-tools.ts:196`). Every other `result: "failure"` site in the api-server uses `warn` or `error`, and `docs/architecture/logging.md:24` states the mapping as "deny/fail → `warn`" — so a refused mint is now recorded, but an operator filtering the trail on `level>=warn` won't see it.
- 🟢 **New:** the audit emission lives in the MCP tool wrapper rather than the service, and the browser route emits the *same* `artifact_library.download` event while still logging only successes (`library-routes.ts:82`). The two surfaces have diverged on the exact question this commit set out to fix.

**Verdict:** minor drift

**Drift**

- ❌ **Check 1 — architecture-page drift:** [`docs/architecture/persistence.md:74`](docs/architecture/persistence.md) — the in-depth `### Object store` section still enumerates the presign audiences as upload→agent / download→browser only ("upload links to agents after attributing the caller, download links to browsers after owner checks"), with no notion of audience-bound signing. This PR makes downloads agent-reachable (`presignDownload` gains `audience: "agent" | "browser"`, dispatching to `agentSigner`/`browserSigner`), so the page's own authority on this concept now contradicts both the code and the Overview bullet the PR *did* widen at `persistence.md:12`. Proposed edit: "…short-lived links scoped to a single object and operation, each signed for the authority its audience dials (uploads to agents after attributing the caller, downloads to an agent through its gateway or to a browser directly, after owner checks)…". `Last verified:` is already `2026-07-30`, so no header change is needed.

**Verdict:** aligned on the parts this PR restructures — the audience split lands in the right layers, and moving `downloadFileName` into `domain/` is a straight compliance win.

🔁 1 finding from the previous review fixed (`this`-dispatch → closure `resolveRef`); 4 new findings, all placement/contract issues, none blocking — the two most material are surfaced under **Findings** above (contract drift, invariant placement). The remaining two:

- 🟢 **`presignDownload`'s `null` now carries two audience-dependent meanings** (`artifacts/domain/artifact-store.ts:44`, `infrastructure/s3-artifact-store.ts:119`). For `browser` it means "relay the bytes"; for `agent` it can only mean "no store configured" (`agentSigner` is non-nullable), which the library service translates to `PRECONDITION_FAILED`. One nullable channel, two outcomes, disambiguated only by the argument the caller passed. Relatedly, `browserSigner: S3Client | null` is a capability toggle hoisted into the consumer and branched on per call, where `createUnconfiguredArtifactStore` already shows the in-dependency shape.
- 🟢 **New test helpers cast a required dependency to `never`** (`__tests__/unit/artifact-library.test.ts:349,428`) — `artifacts: { … } as never` erases the port rather than loosening it, so the download tests would keep passing through a signature change to the very port they exercise. `artifact-service.test.ts:8` does this correctly with `stubStore(overrides: Partial<ArtifactStore>)`.

**Verdict: approve.** The prior 🟢 copy-drift finding is fixed — the callout now names all three tools, and `create_artifact_download_url` matches the tool id actually registered.

- 🟢 `sandbox-artifacts-section.tsx:39,43,48` — the inline-code chip class string `"rounded bg-muted px-1 py-px font-mono text-[12px] text-foreground"` is now duplicated three times verbatim in one JSX block; this commit crossed the rule-of-three threshold. A four-line local `ToolChip` component keeps the callout readable and makes the next tool a one-liner. (A shared `components/ui/` code primitive would be the real fix, but that's a batch migration and out of scope here.)

### Verdict
COMMENT — the immutability invariant is the right call and well argued in both the commit message and the code comments, and every finding from the last review is genuinely addressed rather than papered over; what remains is that the invariant is only enforced half-way (the tRPC contract still offers the fields the service refuses) plus the `persistence.md` bullet that didn't get widened with its sibling.

---

packages/api-server/src/modules/artifacts/domain/artifact-store.ts
```diff
@@ -42,6 +43,12 @@ export interface ArtifactStore {
   ): Promise<string | null>;
   presignDownload(
```
dam-code-guardian, line 44: 🟢 **Suggestion:** with the `audience` split, `null` from `presignDownload` now means two different things depending on the argument the caller passed. For `browser` it keeps its documented meaning ("no reachable authority → relay the bytes"). For `agent` it cannot — `agentSigner` is non-nullable, so an agent-audience `null` only ever arrives from `createUnconfiguredArtifactStore`, and the library service turns it into `PRECONDITION_FAILED "No object store is configured"`. One nullable channel, two outcomes, disambiguated only in prose. Distinct `relay` vs `unconfigured` outcomes (or a `Result`) would state it in the type.

Adjacent, and pre-existing — this PR only renamed it: `browserSigner: S3Client | null` is a capability toggle hoisted into the consumer and re-branched on per call, where `createUnconfiguredArtifactStore` already demonstrates the put-it-in-the-dependency shape. The new `audience` branch is what turned that nullability into a per-call decision.

packages/api-server/src/__tests__/unit/artifact-library.test.ts
```diff
@@ -312,6 +312,174 @@ describe("library service — getPreviewHtml", () => {
...
+        },
+      },
+      owner: "o1",
+      shareBaseUrl: "http://share.localhost",
+      artifacts: {
+        put: (input: { key: string }) => {
+          keys.push(input.key);
+          return Promise.resolve();
+        },
+      } as never,
```
dam-code-guardian: 🟢 **Suggestion:** `as never` here erases the `ArtifactService` port rather than loosening it, so nothing checks that this stub still matches the interface — these tests would keep passing through a signature change to the very port they exist to exercise. `artifact-service.test.ts` has the pattern worth copying (`stubStore(overrides: Partial<ArtifactStore>)`); a `stubArtifacts()` equivalent would fix this call site, the one at line 428, and the pre-existing ones.

packages/ui/src/modules/artifacts/components/sandbox-artifacts-section.tsx
```diff
@@ -43,7 +43,12 @@ export function SandboxArtifactsSection({ agentId }: { agentId: string }) {
           <code className="rounded bg-muted px-1 py-px font-mono text-[12px] text-foreground">
             create_artifact_upload_url
           </code>{" "}
-          for direct-to-storage uploads. No extra credentials needed in the
+          for direct-to-storage uploads. They read artifacts back the same way,
+          with{" "}
+          <code className="rounded bg-muted px-1 py-px font-mono text-[12px] text-foreground">
```
dam-code-guardian: 🟢 **Suggestion:** adding the third chip makes this class string `"rounded bg-muted px-1 py-px font-mono text-[12px] text-foreground"` a verbatim triplicate inside one JSX block — the rule-of-three extraction trigger, crossed by this commit. A local component keeps the callout readable and makes the next tool a one-liner:

```tsx
function ToolChip({ children }: { children: ReactNode }) {
  return (
    <code className="rounded bg-muted px-1 py-px font-mono text-[12px] text-foreground">
      {children}
    </code>
  );
}
```

(`font-mono` is redundant on `<code>` under Tailwind preflight, and drops out for free with the extraction.) A shared primitive in `components/ui/` would be the real fix — other call sites style `<code>` ad hoc too — but that's a batch migration, not this PR.

packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts
```diff
@@ -327,7 +381,6 @@ export function createArtifactLibraryService(
         patch.sizeBytes = stored.sizeBytes;
         patch.contentType = stored.contentType;
         patch.fileName = fileName;
```
dam-code-guardian, line 383: 🟡 **Warning:** this writes the new name onto the head row while `library_artifact_versions` has no `fileName` column — so `resolveRef` keeps serving the head name for a past version (`:191`). Rename `report.md` → `notes.md` at v2 and a **v1** download arrives as `notes.md`, with that name baked into `Content-Disposition` while the bytes come from v1's `storageRef`.

This is the same 🟡 raised at `c52d2563`, which the previous commit closed by freezing the name; this commit reopens it and reclassifies it as intended (`docs/architecture/artifact-library.md:23`: "a past version downloads under the artifact's current name"). Defensible as a product call — but it is now reachable **on purpose** by any agent through the new `file_name` input, where before only pre-existing rows could be inconsistent.

Two ways to make it deliberate rather than incidental:
- derive a past version's name from its `storageRef` basename — `versionKey` already embeds the sanitized name, so it is recoverable without a migration; or
- keep the behavior and say so where an agent reads it. The tool description currently promises the opposite: "every version downloads under this name" is true of the *label*, not of what a v1 fetch returns.

packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts
```diff
@@ -303,15 +362,10 @@ export function createArtifactLibraryService(
...
-        const kind = detectKind({
-          explicit: input.kind,
-          fileName: input.fileName ?? row.fileName,
-          content: contentBuffer,
-        });
+        // The kind is the artifact's, fixed at create: never re-detected from
+        // the incoming bytes or from a new extension, so no revision can turn
+        // an already-shared link into one that executes.
+        const kind = row.kind as ArtifactKind;
         const fileName = input.fileName ?? row.fileName;
```
dam-code-guardian, line 369: 🟡 **Warning:** with `requireStableIdentity` deleted, the kind invariant is enforced by no executable code — only by the *absence* of an input field plus this comment. Nothing rejects a `kind` write: `ArtifactPatch` still permits one (`infrastructure/artifact-library-repository.ts:74`) and the Zod schemas aren't `.strict()`, so the guarantee now rests on every future writer of `patch.kind` reading the comment first.

Given the stated threat model — a share link vetted while it served inert text later executing HTML/JSX at the same URL — this is the one invariant in the module worth keeping executable and testable. `domain/artifact-kind.ts` is the natural home (the layering guidance puts invariants in `domain/`, and it already owns `detectKind`, the function whose result must *not* be applied on update):

```suggestion
        // The kind is the artifact's, fixed at create: never re-detected from
        // the incoming bytes or from a new extension, so no revision can turn
        // an already-shared link into one that executes.
        const kind = stableKind(row);
```

…with `stableKind(row): ArtifactKind` beside `detectKind`. One named function, one test anchor, and the two doc-comments get a single home.

packages/api-server/src/__tests__/unit/artifact-library.test.ts
```diff
@@ -312,6 +334,163 @@ describe("library service — getPreviewHtml", () => {
...
+    await expect(
+      service.update("a1", {
+        content: "<script>alert(1)</script>",
+        fileName: "report.html",
+      }),
+    ).resolves.toMatchObject({ kind: "markdown", fileName: "report.html" });
+    expect(keys).toEqual(["library/o1/a1/v2/report.html"]);
+  });
+
+  it("renames in place without publishing a version", async () => {
```
dam-code-guardian, line 394: 🟢 **Suggestion:** this case pins that a rename doesn't publish a version, and the two above it pin the kind back doors — but nothing covers the semantics this commit actually *changed*: what a **past** version's download is named after a rename. A case asserting `resolveContentRef(id, 1).fileName` (or `createAgentDownloadUrl(id, 1)`) after renaming at v2 would make the intended answer explicit, whichever way the finding on `artifact-library-service.ts:383` is resolved.

packages/api-server/src/__tests__/unit/artifact-library.test.ts
```diff
@@ -173,9 +174,30 @@ function fakeRepo(
   };
 }
```
dam-code-guardian, line 176: 🟢 **Suggestion:** replacing `as never` with `Partial<ArtifactService>` overrides is the right fix and closes the previous review's finding properly. One residue: this single `() => never` is assigned to eight port methods of differing signatures, and typechecks against all of them. So the stub still breaks when a method is **added** to the port, but not when an existing signature **changes** — which was half the original motivation. Typing the not-implemented members individually (or narrowing `notImplemented` per method) would close that gap.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `e310dd70`

### Summary
Adds an agent-facing `create_artifact_download_url` MCP tool minting a short-lived presigned GET so a harness can pull any library artifact into its sandbox. This re-review covers one new commit, `e310dd70`, which narrows the previous commit's immutability rule: only the `kind` is now frozen, the file name is editable again.

### Changes since last review
Previous HEAD: `821817bf` (2026-07-30T14:33:25Z) — verdict COMMENT

- ✅ **Fixed:** the tRPC contract no longer advertises a field the service refuses — `kind` is gone from `artifactUpdateInputSchema` and `ArtifactUpdateInput`, so the invariant is one statement instead of two disagreeing ones (`api-server-api/.../schemas.ts:69`)
- ✅ **Fixed:** the refused-mint audit line is now `warn`, matching the deny/fail mapping (`mcp-tools.ts:196`)
- ✅ **Fixed:** the browser route no longer logs only successes — both failure arms are audited through a shared `audit()` helper, and the success line moved after the blob check (`library-routes.ts:98,116`)
- ✅ **Fixed:** `presignDownload`'s two audience-dependent `null` meanings are documented on the port (`artifacts/domain/artifact-store.ts:55`)
- ✅ **Fixed:** the `as never` test casts are replaced by a typed `stubArtifacts(overrides: Partial<ArtifactService>)` factory (`__tests__/unit/artifact-library.test.ts:176`)
- ✅ **Fixed:** the triplicated chip markup is extracted into `ToolChip` — which also collapsed a stray `text-[12px]` onto `text-xs` (`sandbox-artifacts-section.tsx:14`)
- 🔁 **Still present:** the kind invariant is enforced in no layer — see 🆕 below for what changed about it (`services/artifact-library-service.ts:368`)

### Findings
- 🟡 **New:** a rename now silently re-labels every past version's bytes (`services/artifact-library-service.ts:383`)
- 🟡 **New:** the `kind` invariant is no longer enforced by any code — only by the absence of a field (`services/artifact-library-service.ts:369`)
- 🟢 **New:** the new tests don't cover the reopened rename-across-versions behavior (`__tests__/unit/artifact-library.test.ts:394`)
- 🟢 **New:** `notImplemented` is assigned to eight port methods of differing signatures, so the stub catches added methods but not changed ones (`__tests__/unit/artifact-library.test.ts:176`)

**Verdict:** minor drift

**Drift**

- ❌ **Check 2 — `Last verified:` staleness:** [`docs/architecture/artifact-library.md:3`](docs/architecture/artifact-library.md) reads `Last verified: 2026-07-30`, but `e310dd70` (2026-08-03) rewrites the page's central invariant paragraph at `:16-25` — replacing "file name and kind are settled when it is created" with the narrower kind-only rule plus "Title and file name stay editable". The header was bumped by the two 2026-07-30 commits and left untouched by the commit that reversed their claim. Proposed edit: `Last verified: 2026-08-03`.
- ❌ **Check 2 — `Last verified:` staleness:** [`docs/architecture/persistence.md:3`](docs/architecture/persistence.md) same — the same commit widens the object-store ownership paragraph at `:74` ("each signed for the authority its audience dials and valid on no other"; "download links to an agent through its gateway or to a browser directly"), while the header still reads the earlier commit's `2026-07-30`. Proposed edit: `Last verified: 2026-08-03`.

**Possible drift**

- ⚠️ [`docs/architecture/logging.md:36-45`](docs/architecture/logging.md) — the audit-trail Coverage table has no row covering agent reads of library bytes, and this diff adds one: `mcp-tools.ts:196,209` log `artifact_library.download` with `actorKind: "agent"`, `surface: "mcp"` (refusal at `warn`, success at `info`), and `library-routes.ts:98,116` add the matching browser-route failure lines. The event was already absent from the table before this PR; what the diff changes is that it now has an agent actor and a denial arm. Human judgement: is that table exhaustive over surfaces (then a `Resources` row naming `artifact_library.download` belongs there), or representative (then `:32`'s "all denials" invariant already carries the claim)?

**Verdict:** aligned on what this commit restructures — three findings, all placement/consistency, none blocking.

🔁 5 findings from the previous review fixed (contract drift, audit level, `as never` casts, `null`-per-audience documentation, chip duplication). The two most material remaining are surfaced under **Findings** above (the invariant now living in no layer). The rest:

- 🟢 **`browserSigner: S3Client | null` is still a capability toggle hoisted into the consumer** (`infrastructure/s3-artifact-store.ts:30`, branched at `:119`; `infrastructure.md` → **Required Dependencies — Never Optional**: "Whether a dependency does anything … is owned by that dependency and configured where it is built"). Carried over; this commit *documents* the two-meanings-of-null on the port rather than removing the branch. The docs' own remedy is in-tree — `createUnconfiguredArtifactStore` already shows the null-object shape.

Otherwise the layering holds through the change: `kind` leaving `artifactUpdateInputSchema` keeps validation in the contract package where the docs put it, and the router still spreads validated input straight to the service interface with no business logic (`router.ts:58`). The `audit()` helper is a route-local cross-cutting concern in the transport file, not business logic pushed outward. `domain/artifact-store.ts` stays framework-free, `compose.ts` stays the only place `S3Client` is constructed, and `stubArtifacts` deepens rather than loosens the port dependency.

**Smallest change clearing the most findings:** move the kind rule into `domain/artifact-kind.ts` as a named function the update path calls.

**Verdict: approve.** The prior 🟢 rule-of-three finding is properly fixed, and the fix corrected a latent inconsistency the previous review didn't name.

- ✅ `ToolChip` (`sandbox-artifacts-section.tsx:14`) replaces all three verbatim `<code>` blocks. The third chip previously carried `text-[12px]` where its siblings had `text-xs` — an arbitrary-value one-off rendering a hair off. Extracting to one component collapsed all three onto `text-xs`, and `text-[12px]` is now gone from the UI package entirely. Correctly scoped local to the one file rather than hoisted to `components/ui/` (no other chip in the package shares this class string).

No new findings.

### Verdict
COMMENT — the contract/service disagreement is properly closed and the kind-vs-name split is the better-argued invariant, with the experiments rename path (`experiments-service.ts:497`) confirming the motivation was real. What's worth a look before merge: making the rename's effect on past-version downloads deliberate rather than incidental (the tool description promises more than a v1 fetch delivers), keeping the kind rule executable rather than comment-only, and the two `Last verified:` headers this commit left behind.

---

tomkis (approved):

PR #3076, opened 2026-07-30
feat(ui): correct the base font size and unify text sizes

### LINKED ISSUE #3055: UI - correct the base font size and unify text sizes
## What

The app's base font size is set to 15px. Spacing is derived from that base, so every gap and padding step lands on a fraction of a pixel — a small gap computes to 3.75px, the smallest to 1.875px. On high-density screens that means edges and small round shapes are drawn half a pixel off, which is why some control glyphs had to be given hard-coded sizes to look centred.

Correcting the base to 16px puts the whole spacing scale back on whole pixels.

Text sizes need the same treatment. They are written out by hand in over four hundred places, in thirteen different sizes, several of them a single pixel apart — and with no named set to choose from, every new piece of text invents its own. This replaces them with a small fixed set of sizes, each carrying its own line spacing, and moves the existing text onto it.

## Why

It is the root cause behind a class of "slightly off" rendering rather than a one-off fix, and it removes the need for hard-coded escape hatches in shared components. A fixed set of text sizes stops type drifting a pixel at a time — today the same thing, a description under a form field, appears in three different sizes.

The catch: it makes every gap and padding in the app about 7% larger, and nudges text by a pixel or two wherever it sat off the set. So this is a relayout of every page and needs its own visual pass — which is why it is deliberately not bundled with the consolidation work.

## Done when

- The base size is 16px and the spacing scale is pinned so it cannot drift with typography again.
- Text sizes come from a fixed named set that carries its own line spacing, and hand-written pixel sizes are gone apart from deliberate exceptions.
- Hard-coded sizes added to work around the old fractional scale are removed.
- Long-form article text keeps its own reading size deliberately.
- Every page has been checked in both light and dark mode, paying attention to fixed-height bars and rows where a set height meets derived padding.

kapetr:
Corrects the app's base font size to 16px and replaces hand-written pixel sizes with a named scale.

## Why

The base was 15px, so every spacing step landed on a fraction of a pixel — the smallest gap computed to 1.875px — which is why a few control glyphs carried hard-coded sizes just to look centred. Text sizes were written out by hand in 437 places across thirteen different sizes, several of them a single pixel apart, with no named set to pick from.

## What changes

- Base size is 16px. `--spacing` is pinned to an absolute 4px and the type scale is pinned in px, so neither can drift with typography again.
- Six named text steps, each carrying its own leading: 12/14/16px share the app's 1.55, and 18/24/48px tighten as they grow.
- 238 hand-written sizes move onto those steps. The 13px step is retired apart from the session list title, which the design keeps.
- Form field descriptions, hints and errors all land on 14px — the same thing previously appeared in three different sizes.
- 47 whole-step pixel sizes (multiples of 4px up to 48px) now use the spacing scale.
- `RadioGroupItem`'s `sm` size is retired: once 13px went, its label matched the default's and the variant carried nothing but a font weight.

Two bugs surfaced along the way and are fixed here: the page header's `md:` title was clipping descenders (28px text in a 30px line box inherited from `text-2xl`), and pixel line-heights inherit as absolute values, so nested smaller text was taking its parent's leading instead of scaling.

## Deliberate exceptions

11px labels and meta (including every `SectionLabel`), 15px card and row titles, 10px badges, the session list title's 13px, and layout dimensions above 48px all stay in pixels.

## Review notes

Every gap and padding grows by about 7%, and every Button, Input and Select goes from 13.125px to 14px — so this is a relayout of every page and wants a visual pass in both light and dark rather than a diff read. Worth paying attention to fixed-height bars and rows where a set height meets derived padding.

The 47 spacing conversions are provably no-ops: `--spacing` is exactly 4px, so `h-8` and `h-[32px]` compile to the same declaration.

## Also fixed

Two unrelated things spotted while working on this branch:

- The API keys empty state, every API key row, the env-var marker and the credential connection icon were showing the word "KEY" instead of a key. Carbon's `Key` icon draws the letters K, E and Y; `Password` is its key-shaped icon.
- The session list skeleton referred to a border colour that no longer exists, so its divider was already falling back to the default. It now names the default directly.

Closes #3055


packages/ui/src/App.css
```diff
@@ -27,6 +27,26 @@
   --font-sans: "IBM Plex Sans", system-ui, -apple-system, sans-serif;
   --font-mono: "IBM Plex Mono", ui-monospace, monospace;
 
+  /* Spacing step in absolute pixels, so the whole scale stays on whole
+     device pixels no matter what the root font size becomes. */
+  --spacing: 4px;
```
dam-code-guardian, line 32: 🟢 **Suggestion:** Pinning `--spacing` and the type scale in absolute px means the UI no longer responds to a **raised browser default font size** — the comment's "no matter what the root font size becomes" includes the case where a user deliberately raised it for readability. Page zoom still works, and `-webkit-text-size-adjust: 100%` below was already opting out of mobile autosizing, so this narrows an axis that was only partly live.

The half-pixel problem you're solving is real, so this is a trade-off worth taking now rather than a blocker — but a `calc()`-based scale anchored on `1rem` (e.g. `--spacing: 0.25rem` with the steps as `calc(N/16 * 1rem)`) keeps whole-pixel alignment at the default size *and* preserves scaling. Worth a follow-up issue.

packages/ui/src/App.css
```diff
@@ -27,6 +27,26 @@
...
+
+  /* Type scale in absolute pixels, each step paired with its leading. Body
+     sizes share the app's 1.55 leading; display sizes tighten as they grow.
+     Leading is unitless so nested text at another size scales with it. */
```
dam-code-guardian: 🟢 **Suggestion:** These six steps are pinned in px, but Tailwind's other default steps (`text-xl`, `text-3xl`, `text-4xl`, `text-6xl`+) keep their untouched `rem` values — so the scale is mixed-unit. A future `text-xl` lands back on `rem` and quietly reintroduces the fractional-pixel behavior this PR removes, with nothing at the call site to signal it. (`packages/keycloak-theme` already uses `text-xl` in 3 places; harmless there since it has its own `index.css` with no `@theme` type overrides.)

A line noting the unnamed steps are intentionally outside the design system would steer the next person adding a heading to `text-2xl` instead of `text-xl`.

packages/ui/src/components/ui/empty-state-card.tsx
```diff
@@ -22,10 +22,10 @@ export function EmptyStateCard({
   return (
     <Callout inset className="bg-card">
       <div className="flex flex-col items-center gap-4 py-6">
-        <p className="text-[14px] text-foreground/80">{message}</p>
+        <p className="text-sm text-foreground/80">{message}</p>
         <Button
           variant="outline"
-          className="h-[40px] text-[14px]"
+          className="h-10 text-sm"
```
dam-code-guardian: 🟢 **Suggestion:** This now restates the Button `default` size variant verbatim (`h-10 px-4 py-2 text-sm`). Before this PR, `h-[40px] text-[14px]` was overriding a 13.125px derived size, so it carried meaning; now `tailwind-merge` resolves it to the same declaration, leaving a false signal that this button is sized specially.

```suggestion
          variant="outline"
```

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `536b3e1`

## PR #3076: UI - correct the base font size and unify text sizes
**Author:** kapetr | **Branch:** refactor/3055-base-font-size → main | **Changes:** +415 −464 (147 files)

### Summary
Sets the app's base font size to 16px (from 15px) and introduces a named six-step type scale plus an absolute 4px spacing step as Tailwind v4 `@theme` tokens, then moves 238 hand-written pixel text sizes and 47 whole-step pixel dimensions onto those tokens. Also fixes two rendering bugs (clipped `md:` page-header descenders, px line-heights not scaling in nested text), swaps a mis-drawn Carbon `Key` icon for `Password` in four places, and retires `RadioGroupItem`'s `sm` variant.

### Findings
- ✅ **Looks good:** The 47 spacing conversions are verifiably no-ops. I checked every one programmatically — removed `<prop>-[Npx]` against added `<prop>-M`, asserting `M × 4 === N` — with **0 mismatches**, and every non-multiple-of-4 (`mb-[9px]`, `py-[17px]`, `h-[26px]`, `h-[30px]`, `h-[35px]`, `w-[5px]`, `pr-[18px]`, `basis-[280px]`) correctly left in px. The PR's "provably no-ops" claim holds.
- ✅ **Looks good:** No fixed-height overflow, the risk the PR notes call out. Computing the derived line box for every `text-<step>` + `h-*` pairing in the package: tightest is `h-[26px]` on the egress-approval toast buttons at **+4.3px slack**; the two `button.tsx` icon variants (untouched) at +2.3px/+6.3px. Nothing clips.
- ✅ **Looks good:** Pairing each step with a **unitless** leading is the correct fix for the inheritance bug — nested smaller text now scales with its own size instead of freezing its parent's absolute px leading.
- ✅ **Looks good:** `RadioGroupItem`'s `sm` removal is complete — both call sites drop `size="sm"` in the same diff, no other caller or test references it, and the `cn()` ternaries collapse to static classes. Deleting a variant whose only remaining distinction was a font weight beats leaving dead API.
- ✅ **Looks good:** Deliberate exceptions are each anchored by a comment at the point of use (the one 13px session title, 11px `labelVariants`, 10px badges) rather than left to the reader to infer.
- 🟢 **Suggestion:** `--spacing: 4px` plus the px type scale means text and spacing no longer respond to a raised **browser default font size** — a real accessibility trade-off, taken deliberately to solve a real half-pixel problem. Page zoom is unaffected, and `-webkit-text-size-adjust: 100%` already opted out of mobile autosizing. A `calc()` scale anchored on `1rem` would keep both properties; worth a follow-up, not a blocker. (`packages/ui/src/App.css:32`)
- 🟢 **Suggestion:** The `@theme` block names 6 of Tailwind's 12 default steps, so the scale is mixed-unit — `text-xl`/`text-3xl`/`text-4xl` silently keep `rem` defaults and reintroduce the half-pixel behavior this PR removes, with nothing to signal it. A comment stating the unnamed steps are intentionally outside the design system would steer the next heading to `text-2xl`. (`packages/ui/src/App.css:36`)
- 🟢 **Suggestion:** `className="h-10 text-sm"` now restates the Button `default` size variant verbatim; pre-refactor it was overriding a 13.125px derived size, so it carried meaning. (`packages/ui/src/components/ui/empty-state-card.tsx:28`)

**Verdict:** `aligned` — no drift. The diff is confined to `packages/ui/src/` presentation values. No subsystem behavior or responsibility changes; the `ui` section of `docs/architecture/platform-topology.md` describes the subsystem at framework level and is untouched. No architecture page edited (no `Last verified:` bump due), no new long-lived component, no page moved, no ADR status change.

✅ No findings. The single `.ts` change is one token inside an existing `cva` base string (`text-[14px]` → `text-sm`). `menu-styles.ts` remains a leaf presentation module exporting only string/`cva` constants with one import and no cross-slice dependency; no module boundary or layer rule is engaged.

3 suggestions (the three 🟢 above — mixed-unit scale, `rem` accessibility trade-off, redundant `EmptyStateCard` override), plus verification that the arbitrary-value cleanup is arithmetically sound, no fixed-height overflows, and no line-height conflicts: only `page-header.tsx:29` combines a step with an arbitrary size, and that's the descender fix itself, where the explicit `md:leading-[1.25]` is required precisely because `md:text-[28px]` would otherwise inherit `--text-2xl--line-height`. `tooltip.tsx` and `card.tsx` explicit `leading-*` correctly win over the newly-paired values; `markdown.tsx`'s converted `text-xs` is inside `not-prose`.

### Verdict
APPROVE — the mechanical conversions are provably correct and the two bug fixes are well-targeted; the three suggestions are non-blocking, though the PR's own request for a visual pass in light and dark still stands since no static check can confirm a relayout of every page.

---

PetrBulanek (approved):

PR #3078, opened 2026-07-30
refactor(ui): extract shared clipboard and formatting helpers

### LINKED ISSUE #3052: UI - extract the shared clipboard and formatting helpers
## What

Three small utilities are re-implemented across the UI instead of being shared:

- **Copy to clipboard** — 6 places do it, with 3 different implementations and a duplicated "copied!" state.
- **Formatting** — turning a date into "2 hours ago" happens in 5 modules plus 8 inline calls; file sizes in 4 modules plus 5 inline; pulling a readable message out of an error in 3 helpers plus roughly 16 inline conditionals.

## Why

These are the cheapest kind of duplication to remove — no visual change, no user-facing behaviour to verify. It also means a fix like "show seconds under a minute" or "handle this error shape" happens once rather than being missed in four places.

## Done when

- One clipboard hook, one time/date helper, one file-size helper and one error-message helper, all used everywhere.
- No component formats a date, a byte count or an error message inline.

kapetr:
Closes #3052

## Summary

Four utilities were re-implemented across the UI; this consolidates each into a single shared helper in `src/lib/` and migrates every call site (41 files, +328/−213). No behavioural feature change — the intent is to remove duplication so a fix like "show seconds under a minute" or "handle this error shape" happens once.

- **`use-copy.ts`** — one `useCopy()` hook (`{ copy, state, copied }`) replaces **7** hand-rolled clipboard sites across 3 different styles. `copy` accepts a string or an async producer (for the folder-share URL resolved at click time); the hook owns the try/catch and the `idle`/`copied`/`failed` timeout. No `navigator.clipboard.writeText` remains outside the hook.
- **`format-time.ts`** — `timeAgo` (past), `timeUntil` (was `relativeFromNow`, future), `formatTimestamp` (was `formatSessionTimestamp`), a shared `largestUnit`, and thin `formatDate`/`formatDateTime`/`formatTime` wrappers. Migrated the two relative helpers, the session timestamp, and the **~10 inline `toLocale*`** date sites.
- **`format-size.ts`** — `formatBytes` (moved from artifacts), replacing the inline byte format in chat-view.
- **`errors.ts`** — `getErrorMessage(e, fallback?)` (the rich acp implementation plus a fallback param). Replaces the 3 error helpers and the **~16 inline `instanceof Error`** extractions.

The `done when` sweep is clean: no inline clipboard, date, byte, or error-message formatting remains.

## Behaviour deltas to be aware of (this is a "no visual change" techdebt PR, so calling these out)

- Clipboard copied/failed timeouts unified to **2000ms** (two artifact sites were 1200ms; the failed state was 3000ms).
- chat-view attachment size now **rounds KB** (via `formatBytes`) rather than showing one decimal.
- `getErrorMessage` is slightly **richer** than some sites it replaced (also reads `.message` off non-Error objects) — strictly better, but a widening.

## Deliberately left alone

Schedule-specific date composites (`formatRunTime`, `scheduleCadenceText`) and the `en-US` `Intl.DateTimeFormat` in `schedule-form-options.ts`; artifact `expiryState` (now imports the shared `largestUnit`); the Kubernetes Mi/Gi quantity code (a different unit system).

## Test plan

- New unit tests for `timeAgo`/`timeUntil`/`largestUnit`/`formatBytes`/`getErrorMessage` (the target helpers had zero coverage before)
- `mise run ui:check` and `mise run ui:test` pass (162 tests)

https://claude.ai/code/session_01NeZeHv1xtCW4GXVsWbA9tC

### DIFF
--- a/packages/ui/src/lib/errors.ts
+++ b/packages/ui/src/lib/errors.ts
@@ -3,11 +3,14 @@
  * shapes that actually reach the UI:
  *  - an `Error` / `DOMException` — has `.message`
  *  - the raw JSON-RPC error `{ code, message, data }` — has `.message`
- *  - a WebSocket `CloseEvent` on connection drop — no message; use `code`/`reason`
- *  - a WebSocket `Event` from `onerror` — browsers omit useful details here
+ *  - a WebSocket `CloseEvent` with a `reason` — the reason is the message
+ *  - a `CloseEvent` with no reason / a bare `onerror` `Event` — no usable
+ *    message; only a generic "connection …" line describes it
  *
- * Anything with no usable message falls back to `fallback` when given, else
- * `String(e)` (which is why an `Event` without the guard reads `[object Event]`).
+ * When there is no usable message, a caller-supplied `fallback` wins: it names
+ * the operation that failed ("Delete failed", "Couldn't open <path>"), which
+ * tells the user more than a generic transport line. Without a fallback we fall
+ * back to that transport line, then `String(e)`.
  */
 export function getErrorMessage(e: unknown, fallback?: string): string {
   if (e && typeof e === "object" && "message" in e) {
@@ -15,11 +18,21 @@ export function getErrorMessage(e: unknown, fallback?: string): string {
     if (typeof m === "string" && m) return m;
   }
   if (e instanceof Error && e.message) return e.message;
+  if (
+    typeof CloseEvent !== "undefined" &&
+    e instanceof CloseEvent &&
+    e.reason
+  ) {
+    return e.reason;
+  }
+  // No usable message from here on — a named-operation fallback beats the
+  // generic transport description.
+  if (fallback) return fallback;
   if (typeof CloseEvent !== "undefined" && e instanceof CloseEvent) {
-    return e.reason || `Connection closed (code ${e.code})`;
+    return `Connection closed (code ${e.code})`;
   }
   if (typeof Event !== "undefined" && e instanceof Event) {
     return "Connection error";
   }
-  return fallback ?? String(e);
+  return String(e);
 }
--- a/packages/ui/src/lib/errors.ts
+++ b/packages/ui/src/lib/errors.ts
@@ -14,7 +14,7 @@ export function getErrorMessage(e: unknown, fallback?: string): string {
     const m = (e as { message: unknown }).message;
     if (typeof m === "string" && m) return m;
   }
-  if (e instanceof Error) return e.message;
+  if (e instanceof Error && e.message) return e.message;
   if (typeof CloseEvent !== "undefined" && e instanceof CloseEvent) {
     return e.reason || `Connection closed (code ${e.code})`;
   }
--- a/packages/ui/src/lib/errors.ts
+++ b/packages/ui/src/lib/errors.ts
@@ -1,0 +1,25 @@
+/**
+ * The best human-readable message from an unknown throwable, covering the
+ * shapes that actually reach the UI:
+ *  - an `Error` / `DOMException` — has `.message`
+ *  - the raw JSON-RPC error `{ code, message, data }` — has `.message`
+ *  - a WebSocket `CloseEvent` on connection drop — no message; use `code`/`reason`
+ *  - a WebSocket `Event` from `onerror` — browsers omit useful details here
+ *
+ * Anything with no usable message falls back to `fallback` when given, else
+ * `String(e)` (which is why an `Event` without the guard reads `[object Event]`).
+ */
+export function getErrorMessage(e: unknown, fallback?: string): string {
+  if (e && typeof e === "object" && "message" in e) {
+    const m = (e as { message: unknown }).message;
+    if (typeof m === "string" && m) return m;
+  }
+  if (e instanceof Error) return e.message;
+  if (typeof CloseEvent !== "undefined" && e instanceof CloseEvent) {
+    return e.reason || `Connection closed (code ${e.code})`;
+  }
+  if (typeof Event !== "undefined" && e instanceof Event) {
+    return "Connection error";
+  }
+  return fallback ?? String(e);
+}
--- a/packages/ui/src/lib/errors.ts
+++ b/packages/ui/src/lib/errors.ts
@@ -25,9 +25,11 @@ export function getErrorMessage(e: unknown, fallback?: string): string {
   ) {
     return e.reason;
   }
-  // No usable message from here on — a named-operation fallback beats the
-  // generic transport description.
-  if (fallback) return fallback;
+  // No usable message from here on — a supplied fallback beats the generic
+  // transport description. `!== undefined` (not truthiness) so an intentional
+  // empty-string fallback ("give me a real message or nothing") is honoured
+  // rather than falling through to String(e).
+  if (fallback !== undefined) return fallback;
   if (typeof CloseEvent !== "undefined" && e instanceof CloseEvent) {
     return `Connection closed (code ${e.code})`;
   }
--- a/packages/ui/src/lib/format-time.ts
+++ b/packages/ui/src/lib/format-time.ts
@@ -45,12 +45,19 @@ export function formatTimestamp(iso: string): string {
 
 type DateInput = string | number | Date;
 
+/** null for an unparseable value, so the wrappers can fail soft to "—" rather
+ *  than render "Invalid Date" — they're now the single funnel for every date. */
+function toDate(value: DateInput): Date | null {
+  const d = new Date(value);
+  return Number.isNaN(d.getTime()) ? null : d;
+}
+
 /** `toLocaleDateString` without an inline `new Date(...)`; options pass through. */
 export function formatDate(
   value: DateInput,
   options?: Intl.DateTimeFormatOptions,
 ): string {
-  return new Date(value).toLocaleDateString(undefined, options);
+  return toDate(value)?.toLocaleDateString(undefined, options) ?? "—";
 }
 
 /** `toLocaleString` (date + time) without an inline `new Date(...)`. */
@@ -58,7 +65,7 @@ export function formatDateTime(
   value: DateInput,
   options?: Intl.DateTimeFormatOptions,
 ): string {
-  return new Date(value).toLocaleString(undefined, options);
+  return toDate(value)?.toLocaleString(undefined, options) ?? "—";
 }
 
 /** `toLocaleTimeString` without an inline `new Date(...)`. */
@@ -66,5 +73,5 @@ export function formatTime(
   value: DateInput,
   options?: Intl.DateTimeFormatOptions,
 ): string {
-  return new Date(value).toLocaleTimeString(undefined, options);
+  return toDate(value)?.toLocaleTimeString(undefined, options) ?? "—";
 }
--- a/packages/ui/src/lib/format-time.ts
+++ b/packages/ui/src/lib/format-time.ts
@@ -1,3 +1,15 @@
+type DateInput = string | number | Date | null | undefined;
+
+/** null for an unparseable or absent value, so the formatters fail soft to "—"
+ *  rather than render "Invalid Date" — or the 1970 epoch, which is what
+ *  `new Date(null)` yields. Nullable timestamps are common in the API types,
+ *  and this is the single funnel for every date. */
+function toDate(value: DateInput): Date | null {
+  if (value == null) return null;
+  const d = new Date(value);
+  return Number.isNaN(d.getTime()) ? null : d;
+}
+
 const UNITS: Array<[label: string, ms: number]> = [
   ["d", 86_400_000],
   ["h", 3_600_000],
@@ -13,46 +25,50 @@ export function largestUnit(ms: number): string {
   return "moments";
 }
 
-/** Past-facing relative time: "just now", "5m ago", "2h ago", "3d ago". */
-export function timeAgo(iso: string): string {
-  const delta = Date.now() - new Date(iso).getTime();
+/** Past-facing relative time: "just now", "5m ago", "2h ago", "3d ago";
+ *  "—" for an unparseable date. */
+export function timeAgo(value: DateInput): string {
+  const d = toDate(value);
+  if (!d) return "—";
+  const delta = Date.now() - d.getTime();
   if (delta < 60_000) return "just now";
   return `${largestUnit(delta)} ago`;
 }
 
 /** Future-facing relative time for a next-run glance: "due", "< 1 min",
- *  "in 3 min", "in 2 h", "in 5 d". The absolute time stays on hover/title. */
-export function timeUntil(iso: string, now: Date = new Date()): string {
-  const diff = new Date(iso).getTime() - now.getTime();
+ *  "in 3 min", "in 2 h", "in 5 d"; "—" for an unparseable date. The absolute
+ *  time stays on hover/title. */
+export function timeUntil(value: DateInput, now: Date = new Date()): string {
+  const d = toDate(value);
+  if (!d) return "—";
+  const diff = d.getTime() - now.getTime();
   if (diff <= 0) return "due";
   const min = Math.round(diff / 60_000);
   if (min < 1) return "< 1 min";
   if (min < 60) return `in ${min} min`;
   const hr = Math.round(min / 60);
   if (hr < 24) return `in ${hr} h`;
-  const d = Math.round(hr / 24);
-  return `in ${d} d`;
+  const days = Math.round(hr / 24);
+  return `in ${days} d`;
 }
 
-/** Absolute, zero-padded MM/DD/YYYY HH:MM in local time; "—" for an invalid
- *  date. */
-export function formatTimestamp(iso: string): string {
-  const d = new Date(iso);
-  if (Number.isNaN(d.getTime())) return "—";
-  const p = (n: number) => String(n).padStart(2, "0");
-  return `${p(d.getMonth() + 1)}/${p(d.getDate())}/${d.getFullYear()} ${p(d.getHours())}:${p(d.getMinutes())}`;
+/** The standard compact date + time — locale-aware like the rest of the module
+ *  (24-hour, numeric); "—" for an unparseable date. Reach for this on dense
+ *  timestamp lines (session/run rows); use `formatDate`/`formatDateTime`/
+ *  `formatTime` when you need a different field set. */
+export function formatTimestamp(value: DateInput): string {
+  return formatDateTime(value, {
+    year: "numeric",
+    month: "2-digit",
+    day: "2-digit",
+    hour: "2-digit",
+    minute: "2-digit",
+    hour12: false,
+  });
 }
 
-type DateInput = string | number | Date;
-
-/** null for an unparseable value, so the wrappers can fail soft to "—" rather
- *  than render "Invalid Date" — they're now the single funnel for every date. */
-function toDate(value: DateInput): Date | null {
-  const d = new Date(value);
-  return Number.isNaN(d.getTime()) ? null : d;
-}
-
-/** `toLocaleDateString` without an inline `new Date(...)`; options pass through. */
+/** `toLocaleDateString` without an inline `new Date(...)`; options pass
+ *  through, "—" for an unparseable/absent value. */
 export function formatDate(
   value: DateInput,
   options?: Intl.DateTimeFormatOptions,
@@ -60,7 +76,7 @@ export function formatDate(
   return toDate(value)?.toLocaleDateString(undefined, options) ?? "—";
 }
 
-/** `toLocaleString` (date + time) without an inline `new Date(...)`. */
+/** `toLocaleString` (date + time), same guarantees as `formatDate`. */
 export function formatDateTime(
   value: DateInput,
   options?: Intl.DateTimeFormatOptions,
@@ -68,7 +84,7 @@ export function formatDateTime(
   return toDate(value)?.toLocaleString(undefined, options) ?? "—";
 }
 
-/** `toLocaleTimeString` without an inline `new Date(...)`. */
+/** `toLocaleTimeString`, same guarantees as `formatDate`. */
 export function formatTime(
   value: DateInput,
   options?: Intl.DateTimeFormatOptions,
--- a/packages/ui/src/lib/format-time.ts
+++ b/packages/ui/src/lib/format-time.ts
@@ -1,0 +1,70 @@
+const UNITS: Array<[label: string, ms: number]> = [
+  ["d", 86_400_000],
+  ["h", 3_600_000],
+  ["m", 60_000],
+];
+
+/** The largest whole unit of a duration as a glued abbreviation — "3d", "2h",
+ *  "5m" — or "moments" below a minute. */
+export function largestUnit(ms: number): string {
+  for (const [label, unitMs] of UNITS) {
+    if (ms >= unitMs) return `${Math.floor(ms / unitMs)}${label}`;
+  }
+  return "moments";
+}
+
+/** Past-facing relative time: "just now", "5m ago", "2h ago", "3d ago". */
+export function timeAgo(iso: string): string {
+  const delta = Date.now() - new Date(iso).getTime();
+  if (delta < 60_000) return "just now";
+  return `${largestUnit(delta)} ago`;
+}
+
+/** Future-facing relative time for a next-run glance: "due", "< 1 min",
+ *  "in 3 min", "in 2 h", "in 5 d". The absolute time stays on hover/title. */
+export function timeUntil(iso: string, now: Date = new Date()): string {
+  const diff = new Date(iso).getTime() - now.getTime();
+  if (diff <= 0) return "due";
+  const min = Math.round(diff / 60_000);
+  if (min < 1) return "< 1 min";
+  if (min < 60) return `in ${min} min`;
+  const hr = Math.round(min / 60);
+  if (hr < 24) return `in ${hr} h`;
+  const d = Math.round(hr / 24);
+  return `in ${d} d`;
+}
+
+/** Absolute, zero-padded MM/DD/YYYY HH:MM in local time; "—" for an invalid
+ *  date. */
+export function formatTimestamp(iso: string): string {
+  const d = new Date(iso);
+  if (Number.isNaN(d.getTime())) return "—";
+  const p = (n: number) => String(n).padStart(2, "0");
+  return `${p(d.getMonth() + 1)}/${p(d.getDate())}/${d.getFullYear()} ${p(d.getHours())}:${p(d.getMinutes())}`;
+}
+
+type DateInput = string | number | Date;
+
+/** `toLocaleDateString` without an inline `new Date(...)`; options pass through. */
+export function formatDate(
+  value: DateInput,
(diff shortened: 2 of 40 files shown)


packages/ui/src/lib/query-helpers.ts
```diff
@@ -18,8 +19,7 @@ export async function runAction<T>(
   try {
     return await fn();
   } catch (err: unknown) {
-    const msg = err instanceof Error && err.message ? err.message : fallback;
-    emitToast({ kind: "error", message: msg });
+    emitToast({ kind: "error", message: getErrorMessage(err, fallback) });
```
dam-code-guardian, line 22: 🟡 **Warning:** the fallback stops working for an `Error` with an empty message.

The previous guard was `err instanceof Error && err.message ? err.message : fallback` — the `&& err.message` truthiness check sent `new Error("")` to the caller's fallback. `getErrorMessage` skips its first branch only when `.message` is a *non-empty* string, then hits `if (e instanceof Error) return e.message` (`errors.ts:17`) and returns `""`. So `runAction(fn, "Failed to delete agent")` now emits an **empty error toast** instead of naming the failure.

Empty-message Errors are real here — `new Error()` from a rejected fetch, an aborted request, or a rethrow that lost its message — and `runAction` is the shared wrapper behind every user-initiated mutation, so this degrades the error path app-wide at once.

Fix in `errors.ts:17`, keeping the documented semantics ("anything with no usable message falls back to `fallback`"):

```ts
if (e instanceof Error && e.message) return e.message;
```

Worth a test too: the suite covers `"oops"` and `undefined` with a fallback, but not the empty-message `Error` this turns on.

packages/ui/src/lib/use-copy.ts
```diff
@@ -0,0 +1,42 @@
...
+        if (!text) {
+          setState("failed");
+        } else {
+          await navigator.clipboard.writeText(text);
+          setState("copied");
+        }
+      } catch {
+        setState("failed");
+      }
+      timer.current = setTimeout(() => setState("idle"), timeoutMs);
```
dam-code-guardian: 🟢 **Suggestion:** `timeoutMs` is captured per-callback rather than read at fire time. `copy` is memoized on `[timeoutMs]`, so a caller varying it across renders gets a new `copy` identity on each change, and an in-flight reset still uses the value captured when `copy` was created.

Latent today — all seven sites use the 2000ms default and none pass the option. If it stays unexercised, consider dropping the parameter until a caller needs it, which removes the `useCallback` dependency entirely.

packages/ui/src/lib/errors.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Anything with no usable message falls back to `fallback` when given, else
+ * `String(e)` (which is why an `Event` without the guard reads `[object Event]`).
+ */
+export function getErrorMessage(e: unknown, fallback?: string): string {
+  if (e && typeof e === "object" && "message" in e) {
+    const m = (e as { message: unknown }).message;
+    if (typeof m === "string" && m) return m;
+  }
+  if (e instanceof Error) return e.message;
+  if (typeof CloseEvent !== "undefined" && e instanceof CloseEvent) {
```
dam-code-guardian, line 33: 🟢 **Suggestion:** these branches only fire for genuine DOM instances, so a plain object mimicking a close event loses its diagnostics. `{ reason: "", code: 1006 }` — the shape a non-browser or proxied transport can hand you — has no `.message`, isn't an `instanceof CloseEvent`, and falls through to `fallback ?? String(e)`, dropping the 1006 that would have explained the disconnect.

Carried over from `extractErrorMessage` unchanged, so not introduced here — but centralizing the helper makes this ordering load-bearing for all ~16 migrated sites rather than the two it originally served. Checking `code`/`reason` structurally rather than by instance would cover both.

packages/ui/src/lib/format-time.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+type DateInput = string | number | Date;
+
+/** `toLocaleDateString` without an inline `new Date(...)`; options pass through. */
+export function formatDate(
+  value: DateInput,
+  options?: Intl.DateTimeFormatOptions,
+): string {
+  return new Date(value).toLocaleDateString(undefined, options);
```
dam-code-guardian: 🟢 **Suggestion:** `formatTimestamp` guards invalid dates (`"—"`), but these three `toLocale*` wrappers don't — a malformed value renders `"Invalid Date"` in the UI.

Not a regression (the ~10 migrated inline `toLocale*` sites had no guard either), but since these wrappers are now the single funnel for all of them, extending the same `Number.isNaN(d.getTime())` check would make the module consistently fail soft.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `7805b6b`

### Summary

Consolidates four utilities that had been re-implemented across the UI — clipboard copying, relative/absolute time formatting, byte sizes, and error-message extraction — into single shared helpers under `src/lib/`, migrating all 41 call sites. The extractions are faithful moves (`formatBytes`, `timeAgo`/`largestUnit`, and `relativeFromNow` → `timeUntil` are byte-identical to their originals), the `done when` sweep genuinely is clean — no `navigator.clipboard` outside the hook, no dangling `formatSessionTimestamp`/`relativeFromNow` references — and the PR body's disclosure of the three intended behaviour deltas is exactly the right way to land a "no visual change" refactor. `useCopy` also fixes a real leak all seven previous sites shared: none of them cleared their timeout on unmount.

One unintended behaviour change slipped in beyond the three disclosed, in the most widely-shared error path in the app.

### Findings

- 🟡 **Warning:** `runAction`'s fallback stops working for an `Error` with an empty message (`packages/ui/src/lib/query-helpers.ts:22`). The previous guard was `err instanceof Error && err.message ? err.message : fallback` — the `&& err.message` truthiness check meant `new Error("")` fell through to the caller's fallback. `getErrorMessage` only skips its first branch when `.message` is a *non-empty* string, but then hits `if (e instanceof Error) return e.message` at `errors.ts:17` and returns `""` unconditionally. So `runAction(fn, "Failed to delete agent")` now emits an **empty error toast** where it used to name the failure.

  Empty-message Errors are not hypothetical here — `new Error()` from a rejected fetch, an aborted request, or a rethrow that lost its message all produce one, and `runAction` is the shared wrapper behind every user-initiated mutation in the UI, so this degrades the error path everywhere at once. The fix keeps the documented fallback semantics ("anything with no usable message falls back to `fallback`"):

  ```ts
  if (e instanceof Error && e.message) return e.message;
  ```

  Worth a test alongside the existing `getErrorMessage` cases — the current suite covers `"oops"` and `undefined` with a fallback, but not the empty-message `Error` that this regression turns on.

- 🟢 **Suggestion:** the `CloseEvent`/`Event` branches only fire for genuine DOM instances, so a plain object mimicking a close event loses its diagnostics (`packages/ui/src/lib/errors.ts:18-23`). `{ reason: "", code: 1006 }` — the shape a non-browser or proxied transport can hand you — has no `.message`, isn't an `instanceof CloseEvent`, and so falls through to `fallback ?? String(e)`, dropping the 1006 that would have explained the disconnect. Carried over from `extractErrorMessage` unchanged, so not this PR's doing; worth noting only because centralizing the helper makes the ordering load-bearing for all ~16 migrated sites rather than the two it originally served. Checking `code`/`reason` structurally rather than by instance would cover both.

- 🟢 **Suggestion:** `useCopy` captures `timeoutMs` per-callback rather than reading it at fire time (`packages/ui/src/lib/use-copy.ts:36-38`). `copy` is memoized on `[timeoutMs]`, so a caller that varies the timeout across renders (e.g. from a prop) gets a new `copy` identity each time it changes, and an in-flight reset still uses the value captured when `copy` was created. Latent today — every one of the seven sites uses the 2000ms default and none pass the option — but if the option is never exercised, consider dropping the parameter until a caller needs it, which also removes the `useCallback` dependency entirely.

- 🟢 **Suggestion:** the shared clipboard hook sits in `src/lib/` rather than `src/hooks/` (`packages/ui/src/lib/use-copy.ts`). The project already has `src/hooks/` for exactly this — `use-auto-resize.ts`, `use-unsaved-guard.ts` — while `src/lib/` otherwise holds pure functions (`format-time`, `format-size`, `errors`, `utils`, `breakpoints`). The other three helpers in this PR belong where they landed; only the stateful one is misfiled. Detail from the React skill section below.

- 🟢 **Suggestion:** `formatTimestamp` guards invalid dates but the three `toLocale*` wrappers don't (`packages/ui/src/lib/format-time.ts:49-69`). `formatDate(undefined as never)` or a malformed string yields `"Invalid Date"` in the UI, where `formatTimestamp` would render `"—"`. The ~10 migrated `toLocale*` sites had no guard either, so this is not a regression — but since these wrappers now exist as the single funnel for those sites, extending the same `Number.isNaN(d.getTime())` check to them would make the whole module consistently fail soft.

- ✅ **Looks good:** the `useCopy` extraction is well-judged. It replaces seven state machines in three different styles with one hook that returns a small intentional API, and `copy` accepting either a string or an async producer is what lets the folder-share URL (resolved at click time) share the same path as the plain cases without a second code path. Adding `useEffect(() => () => clearTimeout(timer.current), [])` fixes a setState-after-unmount leak that every original site had.

- ✅ **Looks good:** moving `formatBytes` and `timeAgo` out of `modules/artifacts/lib/format.ts` into `src/lib/` while leaving `expiryState` behind is the right cut — the generic helpers become shared, the artifact-specific concept stays with its module and imports `largestUnit`. It also removes a would-be cross-module dependency: `chat-view` (module `sessions`) now depends on shared `format-size` rather than reaching into `artifacts`.

- ✅ **Looks good:** the new unit tests are well-chosen for helpers that previously had zero coverage, and they pin the behaviours most likely to drift — `timeUntil`'s rounding boundaries, `largestUnit`'s sub-minute case, `formatTimestamp`'s invalid-date em dash, and `getErrorMessage`'s shape precedence. Injecting `now` into `timeUntil` rather than faking timers for it is the cleaner of the two approaches.

**Verdict:** aligned

The `ui` subsystem's architecture coverage is one paragraph in `docs/architecture/platform-topology.md:71-73`, pitched at "React + Vite SPA, tRPC over HTTP, one ACP WebSocket per session" — framework level, as the documentation guidelines prescribe. This PR changes no subsystem behaviour or responsibility: it deduplicates four intra-package utilities and migrates call sites, touching no protocol, component role, or persistence surface. The guidelines put "exact package names, file paths… code-level shape — type signatures, field names, function arguments" out of scope for architecture pages, so these helpers have no home there and their extraction cannot drift a page; no architecture page references clipboard, byte, date, or error formatting at all.

Checks 2–7 are inert: no file under `docs/architecture/` and no ADR in the diff, and `src/lib/` helpers are not a new subsystem.

✅ No findings.

This project carries no `tseng/` metadata, so the TSEng architecture is not adopted here, and the server-side rules (three-layer slices, contract package, domain events, `compose.ts`, module `index.ts`) don't apply to a client package. On the rules that do map, the diff improves things: no cross-module import is added in any of the 20 routed `.ts` files, and one is avoided — `formatBytes`/`timeAgo` moving into `src/lib/` converts what would have been `sessions` → `artifacts` coupling into a dependency on context-free utilities, while `expiryState` stays with its own context. Note that `modules.md`'s "modules are business boundaries, not technical concerns" rule names `utils/` as an invalid *module*; `src/lib/` here is shared client infrastructure rather than a bounded context, so it isn't in scope for that rule.

The core move is what these rules ask for: `references/hooks.md` marks **[CRITICAL] extract a hook when the same `useState` + `useEffect` pattern appears in two or more components**, and this applies it at 7×. No state-lineage violation — every migrated `copied` flag is genuinely UI-local ephemeral state, correctly held in `useState` inside the hook, with nothing duplicated across lineages.

- 🟡 **[HIGH] Shared hooks belong in `src/hooks/`, not `src/lib/`** (`packages/ui/src/lib/use-copy.ts`). `references/hooks.md` §Location names two homes: domain hooks in `src/modules/{domain}/hooks/`, generic ones in `src/hooks/`. `use-copy` is generic, and the project keeps that convention already (`src/hooks/use-auto-resize.ts`, `src/hooks/use-unsaved-guard.ts`).
- 🟢 **The `failed` state is unobservable at five of the seven migrated sites.** `folder-group.tsx:66` takes only `{ copy }`; `share-dialog`, `artifact-row`, `oauth-app-hint`, and `anthropic/form` take only `{ copied }` — so a rejected `writeText` or the null-URL path renders nothing. `folder-group` is the notable one: its "Copy folder link" resolves a URL asynchronously, precisely the case that can fail, and the user gets no signal either way. `copyable-command` and `reveal-token` handle it properly with a `copyState === "failed"` branch. Pre-existing (the originals swallowed failures too), but the hook now exposes the state for free.

### Verdict

COMMENT — a well-executed consolidation with a clean sweep and good new test coverage. One unintended delta needs fixing before merge: `runAction` now emits an empty toast instead of its caller's fallback when an `Error` carries an empty message, which affects every user-initiated mutation in the UI. The rest are suggestions, and the two 🟢 items on `getErrorMessage`'s branch ordering and the missing `toLocale*` guards are carried over from the code being consolidated rather than introduced here.

---

packages/ui/src/lib/format-time.ts
```diff
@@ -0,0 +1,77 @@
...
+/** Past-facing relative time: "just now", "5m ago", "2h ago", "3d ago". */
+export function timeAgo(iso: string): string {
+  const delta = Date.now() - new Date(iso).getTime();
+  if (delta < 60_000) return "just now";
+  return `${largestUnit(delta)} ago`;
+}
+
+/** Future-facing relative time for a next-run glance: "due", "< 1 min",
+ *  "in 3 min", "in 2 h", "in 5 d". The absolute time stays on hover/title. */
+export function timeUntil(iso: string, now: Date = new Date()): string {
```
PetrBulanek: **"Next run: in NaN d"** is reachable today. If a schedule's `nextRun` ever arrives unparseable, the card renders literal NaN to the user rather than degrading to something neutral — and the same input to `formatTimestamp` two functions below returns `—`.

This is a verbatim carry-over from `relativeFromNow`, so not introduced here. Raising it because this PR is the moment it stops being a private quirk of the schedules module: it's now the shared next-run formatter sitting next to siblings that *do* guard, and the docstring sells the module as the single date funnel. `timeAgo` has the softer version of the same gap — an unparseable date reads "moments ago", which is wrong but invisible.

packages/ui/src/lib/format-time.ts
```diff
@@ -0,0 +1,77 @@
...
+  const min = Math.round(diff / 60_000);
+  if (min < 1) return "< 1 min";
+  if (min < 60) return `in ${min} min`;
+  const hr = Math.round(min / 60);
+  if (hr < 24) return `in ${hr} h`;
+  const d = Math.round(hr / 24);
+  return `in ${d} d`;
+}
+
+/** Absolute, zero-padded MM/DD/YYYY HH:MM in local time; "—" for an invalid
```
PetrBulanek: **Two date conventions now ship from one file, and nothing tells a caller which to pick.** `formatTimestamp` is hardcoded US `MM/DD/YYYY`; `formatDate`/`formatDateTime`/`formatTime` follow the viewer's locale. Running the branch under a `cs-CZ` locale, a session row reads `07/31/2026 14:34` while an API-key row right next to it reads `1. 7. 2026` — same app, same module.

That split predates this PR, but consolidating into one file is the natural moment to resolve it.

packages/ui/src/lib/errors.ts
```diff
@@ -0,0 +1,25 @@
...
+    const m = (e as { message: unknown }).message;
+    if (typeof m === "string" && m) return m;
+  }
+  if (e instanceof Error && e.message) return e.message;
+  if (typeof CloseEvent !== "undefined" && e instanceof CloseEvent) {
+    return e.reason || `Connection closed (code ${e.code})`;
+  }
+  if (typeof Event !== "undefined" && e instanceof Event) {
+    return "Connection error";
+  }
```
PetrBulanek, line 38: **These two branches quietly outrank the caller's fallback**, which loses the more useful message. `getErrorMessage(new Event("error"), "Save failed")` returns `"Connection error"` — verified. The docstring at line 9 promises the opposite: "Anything with no usable message falls back to `fallback` when given."

That precedence was right when this served two ACP call sites where a dropped WebSocket *was* the story. It's now the helper behind ~16 sites where the caller has already named the operation — "Delete failed", "Upload failed", "Couldn't open <path>" — and a user reading a generic "Connection error" learns less than one reading which action didn't happen.

Either flip the order so a supplied `fallback` wins, or amend the docstring to say these two shapes take precedence. Happy either way, but the code and the comment should agree.

packages/ui/src/lib/format-time.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+type DateInput = string | number | Date;
+
+/** null for an unparseable value, so the wrappers can fail soft to "—" rather
+ *  than render "Invalid Date" — they're now the single funnel for every date. */
+function toDate(value: DateInput): Date | null {
+  const d = new Date(value);
+  return Number.isNaN(d.getTime()) ? null : d;
+}
```
PetrBulanek: **`formatDate(null)` renders `1. 1. 1970`, not `—`** — `new Date(null)` is the epoch, not `Invalid Date`, so the new guard doesn't catch it. Verified in a browser.

Latent rather than live: `DateInput` excludes `null`, so nothing reaches it today. Worth closing anyway, because nullable timestamps are common in the API types and the comment right above calls this "the single funnel for every date" — the next caller with an `expiresAt: string | null` will reasonably trust it, and a stray 1970 in the UI reads as a data bug rather than a missing value.

```suggestion
function toDate(value: DateInput): Date | null {
  if (value == null) return null;
  const d = new Date(value);
  return Number.isNaN(d.getTime()) ? null : d;
}
```

PetrBulanek (dismissed):

packages/ui/src/query-client.ts
```diff
@@ -60,8 +61,7 @@ export const queryClient = new QueryClient({
         if (mutation.meta?.suppressErrorToast) return;
         if (getApiHealthSnapshot() === "reconnecting" || isTermsStale()) return;
         const title = mutation.meta?.errorToast;
-        const detail =
-          error instanceof Error && error.message ? error.message : "";
+        const detail = getErrorMessage(error, "");
```
dam-code-guardian, line 64: 🔴 **Critical:** the new `if (fallback)` guard is skipped here, because this call site's fallback is the empty string — and `""` is falsy. Execution falls through to the `CloseEvent`/`Event`/`String(e)` branches, which is the opposite of what this call wants.

`detail` is meant to be `""` whenever the error carries no usable message, so the composed toast degrades to the bare `title`. Verified against `main`'s inline guard (`error instanceof Error && error.message ? error.message : ""`), with `errorToast: "Failed to connect Slack"`:

| `error` | main | this PR |
|---|---|---|
| `new Error("boom")` | `Failed to connect Slack: boom` | same ✅ |
| `new Error("")` | `Failed to connect Slack` | `Failed to connect Slack: Error` |
| `undefined` | `Failed to connect Slack` | `Failed to connect Slack: undefined` |
| `{}` | `Failed to connect Slack` | `Failed to connect Slack: [object Object]` |
| `new Event("error")` | `Failed to connect Slack` | `Failed to connect Slack: Connection error` |

`String(new Error(""))` is `"Error"` and `String({})` is `"[object Object]"` — the exact `[object Event]` class of output the module docstring says these guards exist to prevent. This is the shared `onError` for every mutation carrying `errorToast` meta (73 sites), so it degrades the app's error toasts all at once, and it is a regression against `main` rather than a pre-existing gap.

No unambiguous one-line fix, because `getErrorMessage`'s contract can't currently express what this caller wants — "give me a real message or nothing". Two options:

1. **Distinguish "no fallback" from "empty fallback"** in the helper — `if (fallback !== undefined) return fallback;` — then this call site works as written and `getErrorMessage(e, "")` reliably means "empty unless there's a real message".
2. **Keep the helper as-is** and let this call site do its own narrow check, since it is the one place that wants an empty string rather than a description.

Option 1 also keeps the docstring honest: it promises a supplied `fallback` wins, and an empty string is supplied.

Worth noting the added test `getErrorMessage(new Event("error"), "Save failed")` passes because `"Save failed"` is truthy — the falsy-fallback path this call site uses isn't covered, which is why the regression got through.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `2edcc887`

## PR #3078: refactor(ui): extract shared clipboard and formatting helpers
**Author:** kapetr | **Branch:** refactor/3052-shared-ui-helpers → main | **Changes:** +387 −213 (39 files)

### Summary
Second-round fixes: `getErrorMessage` precedence flipped so a caller's fallback beats a generic transport line, and `toDate` now null-guards absent values. Both prior-round asks landed, but the precedence flip regressed the shared mutation error path.

### Changes since last review
Previous HEAD: `7805b6b` (2026-07-30T14:52:04Z) — verdict COMMENT

- ✅ **Fixed:** `runAction`'s empty-message `Error` no longer emits an empty toast (`errors.ts:16`)
- ✅ **Fixed:** `use-copy.ts` moved to `src/hooks/`; unexercised `timeoutMs` param dropped
- ✅ **Fixed:** `toLocale*` wrappers now fail soft to `—`, incl. `formatDate(null)`'s 1970 epoch (`format-time.ts:7`)
- ✅ **Fixed:** `formatTimestamp` is locale-aware, resolving the two-conventions split
- ✅ **Fixed:** `timeAgo`/`timeUntil` degrade to `—` instead of `NaN d` / "moments ago"
- 🔁 **Still present:** `failed` state unobservable at 5 of 7 `useCopy` sites (`folder-group.tsx:66`)
- 🆕 **New:** the fallback-precedence flip regresses every mutation error toast (`query-client.ts:64`)

### Findings
- 🔴 **Critical:** `getErrorMessage(error, "")` — the empty-string fallback is falsy, so the new `if (fallback)` guard never fires and execution falls through to the transport branches, appending garbage to the 73 mutation error toasts (`packages/ui/src/query-client.ts:64`)

**Verdict:** aligned

✅ No findings. No `docs/` file in the diff, no ADR status change, no new subsystem; the `ui` page (`platform-topology.md:71-73`) sits at framework level and references none of these helpers.

✅ No findings.

This project carries no `tseng/` metadata, so the TSEng architecture is not adopted here and the server-side rules (three-layer slices, contract package, domain events, `compose.ts`, module `index.ts`) don't apply to a client package. On the rules that do map, the delta holds the line: no cross-module import is added in any of the 18 routed `.ts` files (`query-client.ts`'s `on-terms-stale` import is pre-existing on `main`), the shared helpers stay dependency-free, and `errors.ts` uses `unknown` + narrowing rather than `any` — its one assertion is the guarded-narrowing pattern.

Both 🟢s from the previous review are resolved: `use-copy.ts` moved to `src/hooks/` (matching `references/hooks.md` §Location and the existing `use-auto-resize` / `use-unsaved-guard` convention), and the unexercised `timeoutMs` parameter was dropped, which removed the `useCallback` dependency exactly as suggested. The `.tsx` call-site migrations are mechanical and correct — no state-lineage change, every migrated `copied` flag is still UI-local ephemeral state held inside the hook.

- 🔁 1 finding from the previous review still present: the `failed` state is unobservable at 5 of the 7 `useCopy` sites (`folder-group.tsx:66` takes only `{ copy }`; `share-dialog`, `artifact-row`, `oauth-app-hint`, `anthropic/form` take only `{ copied }`) — see review at `7805b6b`.

### Verdict
REQUEST_CHANGES — five of six prior findings genuinely fixed, and the date-guard work is thorough. The one blocker is that the fallback-precedence flip regresses the shared mutation error path app-wide; it's a two-character fix.

---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `b31e4bda`

## PR #3078: refactor(ui): extract shared clipboard and formatting helpers
**Author:** kapetr | **Branch:** refactor/3052-shared-ui-helpers → main | **Changes:** +387 −182 (39 files)

### Summary
Third round: the blocking regression is fixed at its root — `getErrorMessage` now guards on `fallback !== undefined`, so an intentional empty-string fallback is honoured instead of falling through to `String(e)`. Verified against `main`'s inline guard across every error shape; the fix also came with the falsy-fallback test that was missing.

### Changes since last review
Previous HEAD: `2edcc887` (2026-08-03T09:39:45Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** the fallback-precedence flip no longer regresses mutation error toasts (`errors.ts:32`)
- 🔁 **Still present:** `failed` state unobservable at 5 of 7 `useCopy` sites (`folder-group.tsx:66`)

### Findings

**Verdict:** aligned

✅ No findings. No `docs/` file in the diff and no ADR status change; the delta is a three-line guard change in an intra-package helper plus a test, which alters no subsystem behaviour or responsibility. The `ui` page (`platform-topology.md:71-73`) sits at framework level and references none of these helpers.

✅ No findings.

No `tseng/` metadata, so TSEng is not adopted; the server-side rules don't apply to a client package. On the rules that map: `errors.ts` stays dependency-free shared infrastructure, the change keeps `unknown` + narrowing with no `any` or non-null escapes, and no cross-module import is added across the 18 routed `.ts` files.

No `.tsx` file changed since `2edcc887` — `use-copy.ts` remains in `src/hooks/` per `references/hooks.md` §Location, with UI-local state and its unmount cleanup intact.

- 🔁 1 finding from the previous review still present: the `failed` state is unobservable at 5 of the 7 `useCopy` sites (`folder-group.tsx:66` takes only `{ copy }`; `share-dialog`, `artifact-row`, `oauth-app-hint`, `anthropic/form` take only `{ copied }`) — see review at `7805b6b`.

### Verdict
APPROVE — the blocker is resolved the right way (option 1: the helper's contract now expresses "a real message or nothing"), and every other call site passes a non-empty literal fallback, so the change is precisely scoped. The one carryover is a pre-existing 🟢.

<details>
<summary>Verification behind the ✅</summary>

Run against a real runtime with `CloseEvent`/`Event` present. `getErrorMessage(e, "")` — the `query-client.ts:64` call — now matches `main` on every shape:

| `error` | main | this HEAD |
|---|---|---|
| `new Error("boom")` | `boom` | `boom` ✅ |
| `new Error("")` | `""` | `""` ✅ |
| `{}` | `""` | `""` ✅ |
| `undefined` / `null` | `""` | `""` ✅ |
| `new Event("error")` | `""` | `""` ✅ |
| `CloseEvent(1006, no reason)` | `""` | `""` ✅ |
| `CloseEvent(reason "gone")` | `gone` | `gone` ✅ |

Composed through `onError` with `errorToast: "Failed to connect Slack"`, the toast degrades to the bare title rather than leaking `Error` / `[object Object]` / `Connection error`; with no `errorToast` it degrades to `Action failed`. A real message and a `CloseEvent.reason` still win.

Scope check: all 21 `getErrorMessage` call sites traced — including the dynamic `failLabel` (`"Rename failed"` / `"Move failed"`) and `` `Upload failed: ${path}` `` / `` `Couldn't open ${path}` `` ones — every fallback other than `query-client.ts:64` is a non-empty literal, so `!== undefined` changes behaviour at exactly the one intended site. The new test pins the falsy-fallback path that let the regression through.

</details>

---

PetrBulanek (approved):

PR #3079, opened 2026-07-30
fix(ui): surface actionable send-failure reasons instead of "Internal error"

xjacka:
## Problem

Addresses the UX half of #3007: when a harness's `session/prompt` fails, the chat UI showed a bare **"Send failed: Internal error"** — the JSON-RPC -32603 boilerplate — even when the real reason (e.g. the model backend answering `402 Payment Required`) was shipped in the error's `data.details` slot. The outage behind #3007 itself was an external billing issue on the inference account and is not fixed by code.

## Changes

- `extractErrorMessage` now reads the JSON-RPC error `data` slot: a bare "Internal error" is replaced by `data.details` when present (both shapes the ACP SDK produces — `data.details` string and a plain string `data`); non-generic messages get the detail appended without duplication.
- `SendErrorCard` classifies known upstream failure classes and renders an actionable hint under the failure reason:
  - **billing / 402 / credits / quota** → check the inference account's credits or quota
  - **credentials / 401 / authentication_error** → check the API/OAuth secret linked to the agent (mirrors agent-runtime's existing auth hint)
  - **rate limit / 429** → wait and retry
  - a detail-less "Internal error" is rewritten to say the agent's session log has the underlying cause
- Error extraction/presentation helpers moved from `modules/acp/utils.ts` to a new `modules/acp/errors.ts` so node-environment unit tests can import them without the upload-API dependency chain (which touches `window` at import time).

## Testing

- New unit tests in `send-error.test.ts` (detail surfacing, billing hint, internal-error rewrite)
- `ui:check` (tsc, eslint, prettier) and the full `ui:test` suite pass

### DIFF
--- a/packages/ui/src/modules/acp/errors.ts
+++ b/packages/ui/src/modules/acp/errors.ts
@@ -1,0 +1,116 @@
+/**
+ * Error extraction and presentation for the ACP session surface. Kept free of
+ * imports with browser side effects so node-environment unit tests (and any
+ * non-DOM caller) can use it directly.
+ */
+
+/**
+ * Read a human-readable message off any error shape we may see here. The
+ * promise that `prompt`/`loadSession` returns can reject with:
+ *  - an `Error` / `DOMException` — has `.message`
+ *  - the raw JSON-RPC error `{ code, message, data }` — has `.message`
+ *  - a WebSocket `CloseEvent` on connection drop — no message; use `code`/`reason`
+ *  - a WebSocket `Event` from `onerror` — browsers omit useful details here
+ *
+ * The fallback `String(e)` on an Event yields `[object Event]`, which is what
+ * users were seeing on disconnect.
+ */
+export function extractErrorMessage(e: unknown): string {
+  if (e && typeof e === "object" && "message" in e) {
+    const m = (e as { message: unknown }).message;
+    if (typeof m === "string" && m) {
+      const details = jsonRpcErrorDetails(e);
+      if (details && !m.includes(details)) {
+        // "Internal error" is the JSON-RPC boilerplate for -32603 — the real
+        // reason (e.g. the model backend's HTTP error) rides in data.details.
+        return /^internal error\.?$/i.test(m.trim())
+          ? details
+          : `${m}: ${details}`;
+      }
+      return m;
+    }
+  }
+  if (e instanceof Error) return e.message;
+  if (typeof CloseEvent !== "undefined" && e instanceof CloseEvent) {
+    return e.reason || `Connection closed (code ${e.code})`;
+  }
+  if (typeof Event !== "undefined" && e instanceof Event) {
+    return "Connection error";
+  }
+  return String(e);
+}
+
+/** Pull the harness's detail string out of a JSON-RPC error's `data` slot.
+ *  The ACP SDK ships it either as `data.details` (string) or as the whole
+ *  `data` value when the harness threw a plain string. */
+function jsonRpcErrorDetails(e: object): string | null {
+  const data = (e as { data?: unknown }).data;
+  if (typeof data === "string" && data) return data;
+  if (data && typeof data === "object") {
+    const details = (data as { details?: unknown }).details;
+    if (typeof details === "string" && details) return details;
+  }
+  return null;
+}
+
+export interface SendErrorDescription {
+  /** What to render as the failure reason. */
+  message: string;
+  /** Actionable next step for failure classes we recognize. */
+  hint?: string;
+}
+
+const SEND_ERROR_HINTS: ReadonlyArray<[RegExp, string]> = [
+  [
+    /\b402\b|payment.required|insufficient.credits|credit balance|billing|quota exceeded|out of credits/i,
+    "The model backend refused the request for billing reasons — the inference account is out of credits or over quota. Topping up or raising the quota on the provider account should restore sends.",
+  ],
+  [
+    /\b401\b|authentication_error|unauthenticated|invalid (api|x-api).key|unauthorized/i,
+    "The model backend rejected the credential. Check that the API/OAuth secret is correct and linked to this agent (Agents → select agent → Secrets).",
+  ],
+  [
+    /\b429\b|rate.limit/i,
+    "The model backend is rate-limiting requests. Wait a moment and retry.",
+  ],
+];
+
+/**
+ * Turn a raw send-failure message into what the error card renders. Known
+ * upstream failure classes (billing, credentials, rate limits) get an
+ * actionable hint; bare JSON-RPC "Internal error" — all the harness gave us —
+ * gets replaced with wording that at least says where to look next.
+ */
+export function describeSendError(raw: string): SendErrorDescription {
+  for (const [pattern, hint] of SEND_ERROR_HINTS) {
+    if (pattern.test(raw)) return { message: raw, hint };
+  }
+  if (/^internal error\.?$/i.test(raw.trim())) {
+    return {
+      message: "The agent couldn't process this message.",
+      hint: "The harness reported an internal error without details — the agent's session log usually has the underlying cause (often the model backend rejecting the request).",
+    };
+  }
+  return { message: raw };
+}
+
+/**
+ * Classify a resume-time failure so the inline error card can render the
+ * right message and action. Prefers structured error fields (ACP JSON-RPC
+ * `code`, tRPC `data.code`) over regexing the human-readable message — the
+ * latter breaks the moment server wording changes.
+ */
+export function classifyResumeError(
+  e: unknown,
+): "not-found" | "connection" | "other" {
+  if (e && typeof e === "object") {
+    const anyE = e as { code?: unknown; data?: { code?: unknown } };
+    if (anyE.code === -32002) return "not-found";
+    if (anyE.data?.code === "NOT_FOUND") return "not-found";
+    if (e instanceof DOMException) return "connection";
+  }
+  const msg = extractErrorMessage(e);
+  if (/not\s*found/i.test(msg)) return "not-found";
+  if (/refused|ECONN|WebSocket|network/i.test(msg)) return "connection";
+  return "other";
+}
--- a/packages/ui/src/modules/acp/errors.ts
+++ b/packages/ui/src/modules/acp/errors.ts
@@ -62,15 +62,15 @@ export interface SendErrorDescription {
 
 const SEND_ERROR_HINTS: ReadonlyArray<[RegExp, string]> = [
   [
-    /\b402\b|payment.required|insufficient.credits|credit balance|billing|quota exceeded|out of credits/i,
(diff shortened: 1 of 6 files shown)


packages/ui/src/modules/acp/errors.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+const SEND_ERROR_HINTS: ReadonlyArray<[RegExp, string]> = [
+  [
+    /\b402\b|payment.required|insufficient.credits|credit balance|billing|quota exceeded|out of credits/i,
+    "The model backend refused the request for billing reasons — the inference account is out of credits or over quota. Topping up or raising the quota on the provider account should restore sends.",
+  ],
+  [
+    /\b401\b|authentication_error|unauthenticated|invalid (api|x-api).key|unauthorized/i,
+    "The model backend rejected the credential. Check that the API/OAuth secret is correct and linked to this agent (Agents → select agent → Secrets).",
```
dam-code-guardian, line 70: 🟡 **Warning:** This 401 hint duplicates guidance the server already injected into the same string.

`agent-runtime` prefixes any frame whose `error.message` contains `authentication_error` with its own `AUTH_HINT` — "Authentication Error: Ensure the API/OAuth credential secret is correct and linked to this agent (Agents > select agent > Secrets)." — applied to every client-mapped response (`packages/agent-runtime/src/modules/acp/services/acp-runtime.ts:1022`, via `rewriteAuthError`). That prefixed text arrives here as `raw`, the pattern on the line above matches it, and this hint then renders a near-identical second sentence underneath. The card shows the same instruction twice.

The PR description notes this "mirrors agent-runtime's existing auth hint" — mirroring is exactly the issue: it duplicates rather than replaces. Either drop this row and let the server's prefix stand, or strip the recognized prefix from `message` here so only one copy survives.

packages/ui/src/modules/acp/errors.ts
```diff
@@ -0,0 +1,116 @@
...
+export interface SendErrorDescription {
+  /** What to render as the failure reason. */
+  message: string;
+  /** Actionable next step for failure classes we recognize. */
+  hint?: string;
+}
+
+const SEND_ERROR_HINTS: ReadonlyArray<[RegExp, string]> = [
+  [
+    /\b402\b|payment.required|insufficient.credits|credit balance|billing|quota exceeded|out of credits/i,
```
dam-code-guardian: 🟢 **Suggestion:** The bare word `billing` matches anywhere in the message, which is broad enough to misfire — an agent whose own error text mentions a billing module or path ("failed to read billing.ts", "billing-service unreachable") would be told its inference account is out of credits.

Every other alternative in this row is a specific upstream signal (`402`, `insufficient credits`, `quota exceeded`, `out of credits`). Anchoring `billing` similarly, or dropping it in favour of those explicit signals, keeps the hint trustworthy — a confidently wrong hint costs more than a missing one.

packages/ui/src/modules/acp/errors.ts
```diff
@@ -0,0 +1,116 @@
...
+    "The model backend refused the request for billing reasons — the inference account is out of credits or over quota. Topping up or raising the quota on the provider account should restore sends.",
+  ],
+  [
+    /\b401\b|authentication_error|unauthenticated|invalid (api|x-api).key|unauthorized/i,
+    "The model backend rejected the credential. Check that the API/OAuth secret is correct and linked to this agent (Agents → select agent → Secrets).",
+  ],
+  [
+    /\b429\b|rate.limit/i,
```
dam-code-guardian: 🟢 **Suggestion:** The `.` here is unescaped, so it matches any character rather than the separator intended — same in `payment.required`, `insufficient.credits`, and `invalid (api|x-api).key` above. Harmless in practice (it only widens matching), but it reads as a typo and invites the same copy-paste into the next row added.

```suggestion
    /\b429\b|rate[\s_-]?limit/i,
```

packages/ui/src/modules/sessions/views/chat-view.tsx
```diff
@@ -982,6 +983,7 @@ function SendErrorCard({
   error: string;
   onRetry?: () => void;
 }) {
+  const { message, hint } = describeSendError(error);
```
dam-code-guardian: 🟢 **Suggestion:** Now that the raw string is formatter input rather than display text, the prop name `error` (line 979) reads as "the thing I render" when it no longer is — the rendered reason is `message`.

Renaming the prop to `rawError`, or having the call site pass a pre-described `{ message, hint }`, would carry the intent. Naming nit, not a blocker.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `9aa52dc`

### Summary
When a harness's `session/prompt` fails, the chat UI rendered the bare JSON-RPC -32603 boilerplate "Send failed: Internal error" even though the real cause shipped in the error's `data.details` slot. `extractErrorMessage` now reads that slot, and a new `describeSendError` classifies billing/credential/rate-limit failures into an actionable hint rendered under the failure reason. The error helpers move from `modules/acp/utils.ts` to a new dependency-free `modules/acp/errors.ts` so node-environment unit tests can import them without the `window`-touching upload-API chain.

### Findings
- 🟡 **Warning:** Credential failures now render **two** overlapping remediation texts stacked in the same card. `agent-runtime` already prefixes any frame whose `error.message` contains `authentication_error` with its own `AUTH_HINT` ("Authentication Error: Ensure the API/OAuth credential secret is correct and linked to this agent (Agents > select agent > Secrets).\n\nError: ") — applied to every client-mapped response at `packages/agent-runtime/src/modules/acp/services/acp-runtime.ts:1022`. That prefixed string reaches the UI as `message`, and the new 401 pattern then matches it, so `hint` appends a near-identical second sentence. The user sees the same instruction twice, once inside the reason line and once beneath it. The PR description says this hint "mirrors agent-runtime's existing auth hint" — mirroring is the problem: it duplicates rather than replaces. Consider dropping the 401 row here, or having the UI strip the server prefix when it recognizes it. (`packages/ui/src/modules/acp/errors.ts:68`)
- 🟢 **Suggestion:** The billing pattern matches the bare word `billing` anywhere in the message, broad enough to misfire on unrelated text — an agent whose own error mentions a billing module or file path ("failed to read billing.ts") would be told to top up the inference account. The neighbouring alternatives are all specific (`402`, `insufficient.credits`, `quota exceeded`); anchoring this one similarly, or dropping it in favour of the explicit signals, keeps the hint trustworthy. (`packages/ui/src/modules/acp/errors.ts:65`)
- 🟢 **Suggestion:** `.` is unescaped in `payment.required`, `insufficient.credits`, `invalid (api|x-api).key`, and `rate.limit`, so it matches any character rather than the space/hyphen/underscore intended. Harmless in practice but it reads as a typo and invites a copy-paste of the same mistake into the next row; `[\s_-]` states the intent. (`packages/ui/src/modules/acp/errors.ts:65-74`)
- ✅ **Looks good:** The `!m.includes(details)` guard correctly suppresses duplication when the harness has already folded the detail into the message — verified against a `{message: "Boom: 402 Payment Required", data: {details: "402 Payment Required"}}` shape, which yields the message unchanged rather than doubling the detail.
- ✅ **Looks good:** Degradation is clean on every malformed `data` shape exercised — absent `data`, `details: ""`, and non-string `details` all fall through to the original message and then to the detail-less rewrite, with no `undefined` or `[object Object]` reaching the card.
- ✅ **Looks good:** The module extraction is complete and leaves no stale importer: `acp/utils.js` is now imported only for `buildPromptBlocks` and `RECONNECT_DELAYS`, and all three moved symbols resolve to `acp/errors.js` at every call site. `errors.ts` imports nothing, so the stated goal (node-importable without the `window` dependency chain) genuinely holds — and the browser globals it feature-detects (`CloseEvent`, `Event`, `DOMException`) all exist in Node 24, so `classifyResumeError` is sound under the node-environment vitest config too.
- ✅ **Looks good:** `describeSendError` is called during render with no `useState`/`useEffect` mirroring — derived state stays derived, and correctly isn't memoized for three regex tests.

**Verdict:** `aligned`

No architecture page under `docs/architecture/` describes send-failure error copy or the ACP client's error-extraction layer as subsystem behavior — `platform-topology.md`'s `ui` section documents the ACP connection model, not error presentation, and `channels.md`'s human-terms wake-failure rule covers the Slack adapter, which this PR does not touch. The diff contains no docs, no ADR status change, and introduces no new long-lived component (`modules/acp/errors.ts` is a module extraction inside the existing `ui` package, not a new subsystem).

✅ No findings.

All five routed `.ts` files are in `packages/ui`, a **client** package. The rules that would bear on this change (the `services → domain ← infrastructure` slice, the `index.ts` module boundary, domain-event-only cross-module communication) are scoped by the architecture docs to **server** packages; for client packages only the tRPC-client dependency, strict TS, and a flexible internal layout are prescribed. Checked and clean: no import cycles (`errors.ts` imports nothing); `unknown` at the boundary with narrowing rather than `any`; the single cast narrows to `unknown`, not to a trusted shape. The repo carries no `tseng/` metadata and no UI module defines an `index.ts` — pre-existing, untouched by this diff, out of scope per the diff baseline.

1 MODERATE finding.

🟢 **MODERATE — `SendErrorCard`'s prop is still named `error` although it is now raw formatter input.** The component takes `error: string` and immediately derives `message` and `hint` from it, so `error` reads as "the thing I display" when it no longer is. `rawError`, or passing a pre-described `{ message, hint }`, would carry the intent. Local judgment call. (`packages/ui/src/modules/sessions/views/chat-view.tsx:979`)

Checked and clean: derived state computed on render, not mirrored into state (the CRITICAL rule); Tailwind utilities with theme tokens for the added hint, no `style={{}}`, no template-literal class strings; a single flat `{hint && …}` guard, no nested ternary; the leaf's JSX grows ~22 → ~25 lines, at the target and not over it; no `any`/`as`/non-null assertion added.

Out of scope per the diff baseline: `chat-view.tsx` is 1014 lines with `ChatView` spanning ~825 and three subcomponents inlined in the parent — far past the ~300-line flag, but this diff adds 2 lines to one leaf and never touches `ChatView`'s body; structural migration belongs in a separate PR.

### Verdict
COMMENT — the core fix is correct and well-tested, but the new 401 hint duplicates guidance `agent-runtime` already injects into the same string, so credential failures render the same instruction twice.

---

xjacka:
Addressed the review in a49b25b:

- **Auth hint duplication (warning):** `describeSendError` now skips its credential hint when the message already carries agent-runtime's `AUTH_HINT` remediation (detected via the "credential secret" phrase), so only one instruction renders. Kept the 401 row itself — it still covers auth failures that don't pass through `rewriteAuthError` (e.g. messages without the `authentication_error` marker).
- **Bare `billing` match:** dropped; the row now only matches explicit upstream signals (402, payment required, insufficient credits, credit balance, quota exceeded, out of credits).
- **Unescaped dots:** replaced with `[\s_-]?` across the patterns.
- **Prop naming:** `SendErrorCard` prop renamed `error` → `rawError`.

PR #3080, opened 2026-07-30
fix(controller)!: unprivileged storage migration, off by default, self-releasing when disabled

jezekra1:
## Why (urgent)

The storage migration merged in #3046 **cannot run on a real OpenShift install**. Its copy Job chowned the target volume root; a restricted SCC strips `CAP_CHOWN` and a Kata runtime class's virtiofs refuses guest-side chown, so every Job failed with `chown: changing ownership of '/mnt/dst-0': Operation not permitted` — and because the migration gate is a hard-stop-strength override only the manager clears, the **10 agents it had admitted stayed scaled to zero** on dev.

Neither cause could surface in the pre-merge rehearsals: the sandbox's `local-path` is hostPath-backed and runs under runc. **No data was ever at risk** — a source volume is deleted only after a verified copy, so every source survived intact (confirmed on dev: sources still Bound and RWX, no `superseded` label anywhere, 10 half-copied targets and nothing else).

## Recovery, and what deploying this does

`storageMigration.enabled` **stays `true` by default** (product decision: the fleet drains itself on upgrade with no owner action). So deploying this resumes the migration with the fixed, unprivileged copy — the currently-gated agents get their stale half-copied targets wiped and re-copied properly, then flip and come back.

The off switch is now a genuine escape hatch, which is the part that was missing: **a disabled manager releases every agent it had gated** — bins the abandoned copy Job and the half-copied target, lifts the gate, restores the prior run state (a previously-running agent is woken). Sources are untouched, so each agent returns to the volume it never left. Set `enabled: false` for one deploy if you would rather release the fleet first and re-enable deliberately (`concurrency: 1` makes the first agent easy to follow in the controller log).

## The fix: unprivileged end to end

- **No init container, no root anywhere.** The per-attempt wipe runs as the agent uid — `chmod` needs *ownership*, not write permission, and the agent owns its whole workspace, so restoring `u+rwX` first makes even a prior attempt's `0500` cache dirs removable.
- **`tar --no-overwrite-dir` replaces `cp -a src/. dst/`.** The chown existed *only* because `cp` applies the source root's ownership/timestamps to the target root, which no unprivileged process can do. tar leaves the target root's own metadata alone while preserving modes (incl. setuid), hard links, sparseness and times. `fsGroup` — unprivileged — makes a CSI target root writable; a write test replaces the root-mode comparison.
- **Ownership**: preserved by construction while the workspace is single-owner (it is, by design — verified on a live agent: `find . ! -uid 65532` returns only the volume root). Anything owned by another uid is **reported with its paths and blocks that agent's migration** rather than being silently re-owned; `allowOwnershipRemap` (default false) accepts the normalization deliberately, and then the uid column drops out of the metadata comparison. Privilege would not help here anyway — that is exactly what the SCC and virtiofs deny.
- **`lost+found` excluded** from the wipe and both verification passes: a fresh ext4 CSI target ships one and the source hasn't got it, so every block-backed migration would otherwise fail on a phantom difference. (Another gap hostPath rehearsals could not show.)
- **Copy Job no longer inherits the agents' `runtimeClassName`.** It runs platform tooling over agent data, never agent code, so Kata's VM isolation buys nothing while its virtiofs costs correctness (the chown refusal; no guarantee on hard-link or sparse fidelity) plus a VM boot per copy. Scheduling (nodeSelector/tolerations/affinity) is still inherited so the target binds where agents run.

## Verification

- Controller: all 7 Go packages green, `go vet` + staticcheck clean, helm lint clean. New test asserts the copy invokes no `chown`, has no init container, uses `--no-overwrite-dir`, excludes `lost+found`, and that a **disabled manager releases a gated agent** while leaving its source intact.
- Not yet exercised live: the sandbox cannot reproduce a restricted SCC or Kata, which is the whole point of this PR. Recommend the dev canary above as the acceptance step — one agent, `concurrency: 1`.

## Follow-ups (not here)

- The RWX-era `0500`-directory and `lost+found` findings suggest an e2e that runs the copy against an ext4 CSI volume; the sandbox can't, CI could.
- Warm PVC pool now provisions RWO spares; whether it still earns its keep post-RWO is an open question worth its own issue.

### DIFF
--- a/packages/controller/pkg/reconciler/storage_migration.go
+++ b/packages/controller/pkg/reconciler/storage_migration.go
@@ -633,10 +633,10 @@ func jobFailed(job *batchv1.Job) bool {
 // (source read-only, target read-write) pair, running a copy + two-pass
 // checksum verification per pair.
 //
-// The copy runs as the AGENT's uid, unprivileged, and needs no capabilities:
-// an agent's workspace is
-// single-owner by construction (everything in it is written as the agent
-// user), so the agent uid can read every file — including 0600 files and
+// The copy runs as the AGENT's uid, unprivileged and needing no
+// capabilities: an agent's workspace is single-owner by construction
+// (everything in it is written as the agent user), so the agent uid can
+// read every file — including 0600 files and
 // 0700 dirs — even on shared-filesystem classes that squash root, where a
 // root-running copy would EACCES on the first private file. Ownership on
 // the target is preserved by construction (the creator is the same uid).
@@ -705,7 +705,7 @@ copy_verify() {
   # controller.storageMigration.allowOwnershipRemap to accept the
   # normalization. Privilege is not an option here: a restricted SCC strips
   # CAP_CHOWN and Kata virtiofs refuses guest chown outright.
-  foreign="$(cd "$src" && find . ! -uid ${AGENT_UID} ! -path ./lost+found | head -20)"
+  foreign="$(cd "$src" && find . ! -uid ${AGENT_UID} ! -path './lost+found*' | head -20)"
   if [ -n "$foreign" ]; then
     if [ "${ALLOW_OWNERSHIP_REMAP}" = "1" ]; then
       echo "note: re-owning these source entries to uid ${AGENT_UID} (allowOwnershipRemap; list may be truncated):"
@@ -753,14 +753,14 @@ copy_verify() {
   # uid:0, while an fsGroup-managed target assigns its own gid — group
   # identity is not part of the workspace contract), the volume root itself,
   # and lost+found.
-  meta() { (cd "$1" && find . -mindepth 1 ! -path ./lost+found -printf "${META_FMT}\n" | LC_ALL=C sort); }
+  meta() { (cd "$1" && find . -mindepth 1 ! -path './lost+found*' -printf "${META_FMT}\n" | LC_ALL=C sort); }
   meta "$src" > /tmp/src.meta
   meta "$dst" > /tmp/dst.meta
   cmp /tmp/src.meta /tmp/dst.meta
   # Pass 2 — content checksums. md5 is deliberate: this is integrity
   # checking of our own quiesced copy, not defense against an adversary
   # crafting collisions — and it is the fastest digest coreutils ships.
-  sums() { (cd "$1" && find . -type f ! -path ./lost+found/\* -print0 | LC_ALL=C sort -z | xargs -0 -r md5sum); }
+  sums() { (cd "$1" && find . -type f ! -path './lost+found*' -print0 | LC_ALL=C sort -z | xargs -0 -r md5sum); }
   sums "$src" > /tmp/src.sum
   sums "$dst" > /tmp/dst.sum
   cmp /tmp/src.sum /tmp/dst.sum
--- a/packages/controller/pkg/reconciler/storage_migration.go
+++ b/packages/controller/pkg/reconciler/storage_migration.go
@@ -121,9 +121,16 @@ func (m *StorageMigrationManager) concurrency() int {
 	return defaultMigrationConcurrency
 }
 
-// RunLoop reconciles until ctx is done. A no-op ticker when disabled.
+// RunLoop reconciles until ctx is done. When disabled it first releases
+// anything a previous run left gated, then stops for good.
 func (m *StorageMigrationManager) RunLoop(ctx context.Context) {
 	if !m.config.StorageMigration.Enabled {
+		// Turning the migration off must never strand an agent: the gate is
+		// a hard-stop-strength negative override that only this manager
+		// clears, so a disabled manager that ignored existing gates would
+		// leave those agents scaled to zero with nothing left to lift them.
+		// Unwind instead — the off switch is an off switch.
+		m.ReleaseGated(ctx)
 		return
 	}
 	slog.Info("storage migration: manager started",
@@ -136,6 +143,72 @@ func (m *StorageMigrationManager) RunLoop(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case <-t.C:
+		}
+	}
+}
+
+// ReleaseGated unwinds every in-flight migration and lifts its gate, so a
+// disabled (or newly-disabled) migration leaves no agent parked. Two shapes,
+// decided per agent by whether its source volume is still there:
+//
+//   - source still ReadWriteMany -> the copy never completed: bin the copy
+//     Job and the unclaimed target, then lift the gate. The agent comes back
+//     on the volume it never left, byte for byte — nothing was deleted.
+//   - source gone -> the flip already ran: finish its tail (sweep the
+//     superseded source, lift the gate, restore the prior run state).
+//
+// Idempotent, and safe to run at boot: an agent with no gate is skipped.
+func (m *StorageMigrationManager) ReleaseGated(ctx context.Context) {
+	agents, err := m.dynamic.Resource(AgentsGVR).Namespace(m.config.Namespace).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		slog.Warn("storage migration: listing agents to release gates failed", "error", err)
+		return
+	}
+	rwx := map[string]bool{}
+	if pvcs, err := m.client.CoreV1().PersistentVolumeClaims(m.config.Namespace).List(ctx, metav1.ListOptions{
+		LabelSelector: LabelAgent,
+	}); err == nil {
+		for _, p := range pvcs.Items {
+			if isRWX(p.Spec.AccessModes) {
+				rwx[p.Labels[LabelAgent]] = true
+			}
+		}
+	} else {
+		slog.Warn("storage migration: listing PVCs to release gates failed", "error", err)
+		return
+	}
+
+	for i := range agents.Items {
+		agent, err := FromCacheObject[apiv1.Agent](&agents.Items[i])
+		if err != nil || agent.Annotations[annStorageMigration] == "" {
+			continue
+		}
+		if rwx[agent.Name] {
+			// Abandon: the source is intact, so the target is half-copied
+			// garbage. The selector only matches UNCLAIMED targets (a
+			// flipped one carries the agent label instead), so a completed
+			// migration's volume can never be caught here.
+			prop := metav1.DeletePropagationBackground
+			if err := m.client.BatchV1().Jobs(m.config.Namespace).Delete(ctx, migrationJobName(agent.Name),
+				metav1.DeleteOptions{PropagationPolicy: &prop}); err != nil && !errors.IsNotFound(err) {
+				slog.Warn("storage migration: deleting abandoned copy job failed", "agent", agent.Name, "error", err)
+			}
+			targets, err := m.client.CoreV1().PersistentVolumeClaims(m.config.Namespace).List(ctx,
+				metav1.ListOptions{LabelSelector: LabelMigrationFor + "=" + agent.Name})
+			if err != nil {
+				slog.Warn("storage migration: listing abandoned targets failed", "agent", agent.Name, "error", err)
+			} else {
+				for _, t := range targets.Items {
+					if err := m.client.CoreV1().PersistentVolumeClaims(m.config.Namespace).
+						Delete(ctx, t.Name, metav1.DeleteOptions{}); err != nil && !errors.IsNotFound(err) {
+						slog.Warn("storage migration: deleting abandoned target failed", "agent", agent.Name, "pvc", t.Name, "error", err)
+					}
+				}
+			}
+			slog.Info("storage migration: released gated agent, migration abandoned", "agent", agent.Name)
+		}
+		if err := m.finishFlip(ctx, agent); err != nil {
+			slog.Warn("storage migration: releasing gate failed", "agent", agent.Name, "error", err)
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


tomkis (approved):

packages/controller/pkg/reconciler/storage_migration.go
```diff
@@ -580,52 +654,115 @@ func buildMigrationJob(agentName string, pairs []migrationPair, cfg *config.Conf
...
+
+  # Ownership: an unprivileged copy recreates entries as the agent uid, so it
+  # is only faithful while the workspace is single-owner — which it is by
+  # construction (the agent writes its own workspace). Anything owned by
+  # another uid is reported and blocks the copy rather than being silently
+  # re-owned. Remedy: chown it to the agent uid on the source, or set
+  # controller.storageMigration.allowOwnershipRemap to accept the
+  # normalization. Privilege is not an option here: a restricted SCC strips
+  # CAP_CHOWN and Kata virtiofs refuses guest chown outright.
+  foreign="$(cd "$src" && find . ! -uid ${AGENT_UID} ! -path './lost+found*' | head -20)"
```
dam-code-guardian, line 708: 🔴 **Critical:** this preflight matches the volume root itself, so with the default `allowOwnershipRemap: false` every legacy migration blocks immediately.

`find .` emits `.` as its first result, and a pre-cutover volume root is *not* agent-owned — which is why the metadata comparison below deliberately excludes it (`-mindepth 1` on `meta()`), and why the comment this PR deletes said "old roots predate the single-owner model". This guard has no such exclusion, so `.` lands in `$foreign`, `[ -n "$foreign" ]` is true, and the Job exits 1 with "source entries owned by another uid".

The PR body reports this very result as *confirmation* of single-ownership: "verified on a live agent: `find . ! -uid 65532` returns only the volume root". That non-empty output is exactly what trips the guard.

```suggestion
  foreign="$(cd "$src" && find . -mindepth 1 ! -uid ${AGENT_UID} ! -path './lost+found*' | head -20)"
```

packages/controller/pkg/reconciler/storage_migration.go
```diff
@@ -580,52 +654,115 @@ func buildMigrationJob(agentName string, pairs []migrationPair, cfg *config.Conf
...
+    fi
+  fi
+
+  # Per-attempt wipe so retries are deterministic. Restoring u+rwX first is
+  # what makes it possible as the owner: a prior attempt faithfully
+  # reproduced the workspace's read-only directories (0500/0555 — Go module
+  # caches, site-packages), and removing an entry needs write+execute on its
+  # parent. LOST+FOUND is skipped: it is a fresh ext4 volume's own artifact,
+  # root-owned and not ours to delete — the verification excludes it too.
+  find "$dst" -mindepth 1 -maxdepth 1 ! -name lost+found -exec chmod -R u+rwX {} +
```
dam-code-guardian, line 727: 🔴 **Critical:** this wipe dies on a dangling symlink, so every retry fails permanently.

`chmod` cannot operate on a dangling symlink and returns non-zero; `find -exec … +` propagates that, and under `set -e` the script exits before the copy starts. The agent workspace always contains one: the base image's entrypoint replaces `$HOME/.cache` with a symlink to pod-local `/tmp/agent-cache` (`packages/platform-base/entrypoint.sh:34-37`), and that link is dangling inside the copy Job, which never runs the entrypoint. `tar` reproduces it onto the target, so from the second attempt on the wipe hits it and the migration can never recover — including the ten half-copied targets already on dev.

Restricting the `chmod` to directories fixes it and matches the intent: only a directory's mode blocks removal of its contents, and `-type d` skips symlinks inherently.

```suggestion
  find "$dst" -mindepth 1 ! -name lost+found -type d -exec chmod u+rwX {} +
  find "$dst" -mindepth 1 -maxdepth 1 ! -name lost+found -exec rm -rf {} +
```

packages/controller/pkg/reconciler/storage_migration.go
```diff
@@ -580,52 +654,115 @@ func buildMigrationJob(agentName string, pairs []migrationPair, cfg *config.Conf
...
+	// The ownership preflight and the uid column of the metadata comparison
+	// move together: normalizing ownership (opt-in) means the target's uids
+	// legitimately differ from the source's, so comparing them would fail by
+	// design. Everything else — type, mode, path, symlink target, content —
+	// is verified either way.
+	metaFmt := "%y|%m|%U|%p|%l"
+	if cfg.StorageMigration.AllowOwnershipRemap {
+		metaFmt = "%y|%m|%p|%l"
+	}
+	fmt.Fprintf(&script, "set -euo pipefail\nAGENT_UID=%d\nALLOW_OWNERSHIP_REMAP=%s\nMETA_FMT=%q\n",
```
dam-code-guardian, line 674: 🟡 **Warning:** `pipefail` turns both `find … | head -20` guards into silent exit-141 failures once the list exceeds the pipe buffer.

`head` exits after 20 lines, `find` takes SIGPIPE, `pipefail` makes the substitution fail, and `set -e` then kills the script *before* it prints the block reason. Verified in a pod: 20 matches survive, ~500 matches exit 141. So "fail fast with the exact list" and the `allowOwnershipRemap` remedy hint — the whole point of both preflights — vanish exactly when they matter, leaving a bare non-zero Job.

`pipefail` is a real improvement for the `tar | tar` pipe, so keep it and make the guards tolerate truncation (trailing `|| true` on the capture, or drop `head` and truncate the echo).

packages/controller/pkg/reconciler/storage_migration_test.go
```diff
@@ -34,7 +35,8 @@ func migrationConfig() *config.Config {
 			StorageSize: "10Gi",
 		},
 		StorageMigration: config.StorageMigration{
-			Enabled:  true,
+			Enabled: true, // the chart default is off; tests drive it on
```
dam-code-guardian, line 38: 🟢 **Suggestion:** this comment now contradicts the chart — `deploy/helm/platform/values.yaml:1049` sets `enabled: true`. It's a leftover from the previous commit's "off by default" stance, before `18d4d9b` restored the default.

```suggestion
			Enabled:  true,
```

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `18d4d9b`

### Summary

Reworks the RWX→RWO workspace migration to be unprivileged end to end after the merged version proved unrunnable on a real OpenShift install: the root init container and the target-root `chown` are gone, `cp -a src/. dst/` becomes `tar --no-overwrite-dir`, `lost+found` is excluded from the wipe and both verification passes, the copy Job no longer inherits the agents' `runtimeClassName`, and a disabled manager now releases every agent it had gated instead of stranding it at zero replicas. The diagnosis is convincing and the direction is right — the `ReleaseGated` unwind and dropping the Kata runtime class are both correct calls, and the new test asserts on the copy's *commands* rather than its prose, which is the right way to pin "no chown anywhere".

Two defects in the new copy script will stop this from working on the very fleet it is meant to unblock, both in code paths the sandbox rehearsals cannot reach for the same reason the original bug survived them.

### Findings

- 🔴 **Critical:** the foreign-ownership preflight matches the volume root itself, so with the default `allowOwnershipRemap: false` **every legacy migration blocks immediately** (`packages/controller/pkg/reconciler/storage_migration.go:708`). `find .` emits `.` as its first result, and the metadata comparison deliberately excludes the root (`-mindepth 1` on the `meta()` helper, line 756) precisely because a pre-cutover volume root is *not* agent-owned — the comment this PR deletes said so outright ("old roots predate the single-owner model"). The preflight has no such exclusion, so `.` lands in `$foreign`, `[ -n "$foreign" ]` is true, and the Job exits 1 with "source entries owned by another uid".

  The PR body reports this result as the *confirmation* of single-ownership: "verified on a live agent: `find . ! -uid 65532` returns only the volume root". Returning the volume root is exactly the non-empty output that trips the guard — the verification demonstrates the bug rather than ruling it out. Fix is one flag, matching the comparison it guards:

  ```
  foreign="$(cd "$src" && find . -mindepth 1 ! -uid ${AGENT_UID} ! -path './lost+found*' | head -20)"
  ```

- 🔴 **Critical:** the per-attempt wipe dies on a dangling symlink, so **every retry fails permanently** (`storage_migration.go:727`). `chmod` cannot operate on a dangling symlink and returns non-zero; `find -exec … +` propagates that, and under `set -e` the script exits before the copy starts. The agent workspace always contains one: the base image's entrypoint replaces `$HOME/.cache` with a symlink to pod-local `/tmp/agent-cache` (`packages/platform-base/entrypoint.sh:34-37`, documented on the persistence page), and that link is dangling inside the copy Job, which never runs the entrypoint. `tar` faithfully reproduces it onto the target, so from the second attempt onward the wipe hits it and the migration can never recover — including on the ten half-copied targets already sitting on dev, which is the first thing this PR will run against.

  Restricting the `chmod` to directories both fixes it and matches the intent (only a directory's mode blocks removal of its contents; `-type d` skips symlinks inherently, and `u+rwX` on a file was never needed):

  ```
  find "$dst" -mindepth 1 ! -name lost+found -type d -exec chmod u+rwX {} +
  find "$dst" -mindepth 1 -maxdepth 1 ! -name lost+found -exec rm -rf {} +
  ```

- 🟡 **Warning:** `set -euo pipefail` turns both `find … | head -20` guards into silent exit-141 failures once the list exceeds the pipe buffer (`storage_migration.go:674`). `head` exits after 20 lines, `find` takes SIGPIPE, and `pipefail` makes the command substitution fail — so with `set -e` the script dies *before* printing the block reason. Verified in this pod: 20 matches survive, ~500 matches exit 141. The result is that the "fail fast with the exact list" and "remedy: chown them… or set allowOwnershipRemap" diagnostics — the entire point of both preflights — vanish exactly when they are needed most, leaving an operator a bare non-zero Job. `pipefail` is a genuine improvement for the `tar | tar` pipe, so keep it and make the guards tolerate truncation, e.g. capture with a trailing `|| true`, or drop `head` and truncate the echo instead.

- 🟢 **Suggestion:** the test config comment now contradicts the chart (`packages/controller/pkg/reconciler/storage_migration_test.go:38`). It says "the chart default is off; tests drive it on", but `deploy/helm/platform/values.yaml:1049` sets `enabled: true` — the comment is a leftover from the previous commit's "off by default" stance, before `18d4d9b` restored the default. Worth correcting so it doesn't read as documentation of a default that isn't there; the stray blank line after it can go too.

- 🟢 **Suggestion:** `finishFlip` announces success on the abandon path (`storage_migration.go:560`). `ReleaseGated` logs "released gated agent, migration abandoned", then calls `finishFlip`, which unconditionally logs "agent migrated to ReadWriteOnce storage" — so an operator watching an abandoned release sees a contradictory pair claiming the migration completed. Passing the outcome in, or moving that log to `flip`'s own tail, keeps the log honest.

- ✅ **Looks good:** `ReleaseGated`'s two-shape decision is keyed on the right signal — the source PVC still being RWX — and the abandon branch only ever deletes the copy Job and PVCs carrying `LabelMigrationFor`, which a flipped target no longer has, so a completed migration's volume genuinely cannot be caught by it. Nothing on that path deletes a source.

- ✅ **Looks good:** dropping `runtimeClassName` while keeping `applyAgentBaseScheduling` is the right split, and the ordering (assign after the helper) makes it stick. The rationale — platform tooling over agent data, never agent code — is sound, and the correctness cost of Kata's virtiofs here is real.

- ✅ **Looks good:** the writability probe (`touch`/`rm` on the target root) replacing the root-mode comparison correctly stops asserting a property the copy does not own, and coupling the metadata format to `allowOwnershipRemap` so the uid column drops out exactly when ownership is intentionally normalized avoids a verification that would fail by design.

**Verdict:** minor drift

❌ **Check 1 — architecture-page drift:** `docs/architecture/persistence.md:125` ("Storage migration (transitional)") describes the lifecycle as gate → copy → verify → re-point → delete → restore and calls it interrupt-safe, but does not reflect the new operator-visible behavior this diff adds: **disabling the migration now unwinds it.** `RunLoop` calls `ReleaseGated` before returning when `Enabled` is false (`storage_migration.go:126-134`, `:161`), lifting the gate on every parked agent, binning abandoned Jobs and half-copied targets, and restoring prior run state.

That is an operational invariant a reader of this subsystem needs — the gate is a hard-stop-strength override *only this manager clears*, so "what happens if I turn it off mid-drain" is exactly the coupling the page exists to answer. As written, a reader would reasonably conclude that disabling the knob strands gated agents at zero replicas. Proposed edit, appended to that paragraph:

> Turning the migration off is safe at any point: a disabled manager releases every agent it had gated — discarding the in-flight copy and its half-filled target — and each agent returns to the volume it started from, since a source is only deleted after a verified copy.

Mechanism detail (`ReleaseGated`, label selectors, the `tar`/`chmod` mechanics, `allowOwnershipRemap`) stays out per the volatile-content rule. The page sits at 18,271 of its 40,000-character budget, so there is ample headroom.

Checks 2–7 are inert: the diff touches no file under `docs/architecture/` and no ADR, and `ReleaseGated` is a method on an existing manager rather than a new long-lived component.

### Verdict

REQUEST_CHANGES — two defects in the new copy script would block this on the fleet it is meant to unblock: the ownership preflight trips on the volume root (blocking every legacy migration under the default config, as the PR body's own verification output shows), and the wipe's `chmod` dies on the persisted dangling `$HOME/.cache` symlink, making every retry fail permanently.

---

PR #3082, opened 2026-07-30
feat(metrics): per-session LLM cost in the sessions list

xjacka:
## What

Shows each session's LLM cost next to its timestamp in the sessions sidebar, gated behind a new **Session costs** experimental feature flag (per-user, default off). Hovering the cost shows a breakdown tooltip (API calls, tokens in/out, exact cost).

## How

- **`runtimeBySession` now folds child harness runs into their root session.** A child run (a `claude -p` subshell, a `dam-run` executor) mints its own OTel `session.id` but inherits the parent's trace context; the rollup now groups each session under the root session of its trace family (root = earliest session on a shared trace), so a row's totals cover everything the session drove. This moves the existing session-*filter* fold into the grouping itself; every fold subquery stays gated on the owner allowlist. Sessions with no traced rows keep their own id.
- **New `session-costs` feature id** + a row in the Experimental features tab.
- **UI**: `useSessionCosts` hook (one `metrics.overview` read per agent, 60s refetch, keyed map by ACP session id) wired into the sessions sidebar; `SessionRow` renders the cost next to the timestamp. A disabled metrics backend (no ClickStack) or a missing row just hides the chip — no error surface.
- **Docs**: `docs/architecture/metrics.md` updated (per-session rollup semantics + last-verified).

## Testing

- New unit test pins the fold SQL shape: root resolution, 3× ownership gate, session-unfiltered CTEs, own-id fallback (`metrics-reader-session-fold.test.ts`).
- Both fold SQL variants (plain and session-scoped) executed against a live ClickHouse 25.7 (local ClickStack install).
- `api-server:test` 755 ✓, `ui:test` 153 ✓, check/tsc/lint/format green across api-server, api-server-api, ui.

## Notes

- Rows written before ClickStack was enabled are not backfilled; costs appear from the first exported call.
- Found while testing (not addressed here): enabling ClickStack does not propagate telemetry env to agents created before the flip — template env is copied into `agent_env` only at agent create. Worth a follow-up issue.


dam-code-guardian (commented):
## PR #3082: feat(metrics): per-session LLM cost in the sessions list
**Author:** xjacka | **Branch:** feat/session-costs → main | **Changes:** +166 −19 (9 files)

### Summary
Adds a per-session LLM cost chip to the sessions sidebar behind a new `session-costs` feature flag, and reworks the `runtimeBySession` ClickHouse rollup to fold child harness runs into their spawning session via shared trace context.

### Findings
- 🟡 **Warning:** The sidebar polls an unbounded all-time scan every 60s (`packages/ui/src/modules/metrics/api/queries.ts:20`)
- 🟡 **Warning:** Fold is single-hop, so a grandchild run rolls up under its parent, not the root session (`packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts:199`)
- 🟢 **Suggestion:** Hoist the inline `select` to a stable module-level function (`packages/ui/src/modules/metrics/api/queries.ts:24`)
- ✅ **Looks good:** The ownership gate survives the rewrite — all three CTEs carry the owner predicate, and the test pins it at exactly 3.

**🟡 Unbounded window on a 60s poll.** `useSessionCosts` sends `{ agentId, limit: 1 }` with no `sinceHours`, and the schema treats an omitted `sinceHours` as *no time bound* — so every sidebar render scans the agent's entire telemetry history, re-running the two fold CTEs over all time, every 60 seconds per open client. `limit: 1` does not help: it only bounds `contextPerCall`'s unaggregated rows (`clickhouse-reader.ts:259-277`), never `runtimeBySession`. The sidebar only shows costs for sessions in the visible list, so a bounded lookback (e.g. `sinceHours: 720`, the schema's cap) would cut the scan sharply at no visible cost. Passing `limit: 1` while leaving the window unbounded reads as though the payload was the concern and the scan was overlooked.

**🟡 Single-hop fold.** `trace_root` resolves a root *per TraceId*, and `session_root` then picks, for each session, the root of its earliest trace. Both docs and code describe the result as "the root session of its trace family". That holds for one level of spawning. But a `dam-run` executor that itself spawns a `claude -p` gets a distinct `TraceId` per hop (the harness starts a new trace per run while forwarding `TRACEPARENT`), so the grandchild's rows resolve to the *child's* session id — not the top-level session. `rootSid` is never re-resolved transitively, so multi-level spawn chains still fragment, and the row's totals do not "cover everything the session drove" as the doc claims. Worth either making the fold transitive (iterate root resolution to a fixpoint) or narrowing the documented claim to one level.

**🟢 Stable selector.** The inline `select` arrow is a new identity each render, so TQ rebuilds the `Map` on every render of `SessionsSidebar`. Harmless today (`SessionRow` isn't memoized), but hoisting it to a module-level `const toCostMap = …` is a one-liner that keeps the derivation stable if anyone memoizes the row later.

**Verdict: minor drift**

❌ **Check 1 — consumer scope incomplete** (`docs/architecture/metrics.md:7`): the page still names the Settings Usage tab as the subsystem's only surface, but the sessions sidebar is now a second independent consumer of `metrics.overview`. *Fix:* name both surfaces in the Overview sentence.

❌ **Check 1 — fail-closed contract has a new silent consumer** (`docs/architecture/metrics.md:52`): the page commits to failing loud (`PRECONDITION_FAILED`, rationale: an empty success "would silently misreport a bill as zero") and documents only the Usage tab's explicit unavailable state. `useSessionCosts` sets `retry: false` and no `errorToast` meta, so the query cache's `onError` returns early and the chip simply vanishes. Defensible for a decorative chip, but it is a new variant of the documented error behavior. *Fix:* state that the guarantee is that the *read* never returns a false zero, and that presentation is per-surface.

Checks 2–7 clean: `Last verified:` correctly bumped to `2026-07-30` (matches the PR's commit date), no ADR references or `Motivated by:` sections, no volatile package/path detail added (only the process names `claude -p` / `dam-run`), no new subsystem, no page moved, no ADR status change. Credit where due — the rollup's new grouping semantics *were* folded into the page in the same PR, at the right altitude.

**No findings.** The SQL rewrite stays entirely inside `infrastructure/` — `otel_logs`, `TraceId`, and `LogAttributes` appear in no service or contract file. `runtimeBySession` still fulfills the existing `MetricsReader` port with an unchanged signature (adapter-internal change). The new `session-costs` enum member lands in the contract package's Zod schema, exactly where validation belongs, and the metrics router still only validates and delegates. No dependency was made optional; notably the feature flag is gated at the client consumer rather than threaded through the service as a toggle dep, which is what the required-dependencies rule asks for.

*Scope note:* `clickhouse-reader.ts` imports its port from `../services/metrics-service.js`, an infrastructure→services import the dependency rule forbids. Verified against `origin/main` as pre-existing and unamplified by this PR, so out of scope here — but worth a follow-up.

One finding, reported as the 🟢 suggestion above (inline `select` / derived-state stability).

Otherwise clean: session costs are server-owned and live only in the TQ cache — no Zustand slice, no `useState` mirror, no `useEffect` + fetch. `skipToken` gating means the query never fires while disabled or agent-less rather than firing and discarding. `retry: false` with no `errorToast` is a coherent choice for a decorative chip (avoids a toast every 60s tick on a metrics-disabled deployment). `useSessionCosts` names what it returns and sits with its siblings in `modules/metrics/api/`. The chip uses `tabular-nums` (correct for shifting digits), and `cost?: SessionRuntime` is typed from the contract package rather than restated locally.

### Verdict
**COMMENT** — Solid, well-tested change with genuinely good ownership discipline in the SQL; the unbounded 60s scan and the single-hop fold's mismatch with its documented claim are both worth addressing before merge, but neither is a correctness bug in the happy path.

---
🛡️ Reviewed by **Code Guardian 🛡️** · [definition](https://github.com/dam-agents/code-guardian)

JanPokorny (approved):

PR #3083, opened 2026-07-30
fix(controller)!: migration destination is its own class, defaulting to the cluster default

jezekra1:
## The bug

The migration took its target storage class from `controller.agent.base.storageClass`. On an install that hasn't migrated yet **that field still names the shared filesystem being drained** — so the first production drain moved workspaces from `ReadWriteMany` to `ReadWriteOnce` **on the same `ibmc-vpc-file-*` NFS class**: every agent force-restarted, every byte copied, and the backend this whole epic exists to leave still load-bearing. The access mode changed; nothing else did.

And it did not self-correct. The trigger was the access mode alone, so once a workspace was RWO-on-file it became **invisible to the migration forever** — fixing the values would have fixed only newly-created agents while every already-migrated workspace stayed on NFS.

## The fix

- **`storageMigration.targetStorageClass`** is the destination now, independent of the agents' own class. Empty (the default) means the **cluster default class** — the target PVC carries no `storageClassName` at all and gets ordinary storage, which is the point of the exercise.
- **Two triggers:** a workspace migrates when it is `ReadWriteMany` **or** when it sits on a class other than the target. The second is what recovers everything the broken drain stranded on the filer. A volume with *no* class is left alone (statically provisioned — converging it isn't this component's call), and one already on the target class is skipped, so the drain converges instead of looping.
- **Targeting the class a workspace already sits on is refused**, with a log naming the remedy, unless `allowSameStorageClass: true` says it's deliberate. That is exactly the misconfiguration above; it can no longer happen silently.
- Values docs say plainly that `targetStorageClass` is the destination, and that **`agent.base.storageClass` and `warmPool.storageClass` must be fixed too** — otherwise newly-created agents keep landing on the old backend regardless.

## Operator steps (independent of this PR)

1. `storageMigration.enabled: false` → deploy. Releases every gated agent, bins half-copied targets, stops further copies onto the filer.
2. Point **`agent.base.storageClass`** and **`warmPool.storageClass`** at the block class (or empty for cluster default), or new agents keep being born on NFS.
3. Deploy this PR, then re-enable — ideally `concurrency: 1` for the first agent — and watch the log. Workspaces already flipped to RWO-on-file are picked up by the new class trigger and moved for real.

## Also here: the copy is ~5x cheaper in round-trips

A 1 GiB workspace was taking **an hour** on the IOPS-capped file class. The bytes were never the issue — a workspace is small-file-heavy (a pnpm store is 100k+ entries) and every entry is a network round-trip *and* an IOP against a capped share. The script walked the source **five times** before counting data passes: readability preflight, `du -sh` for a log line, ownership preflight, metadata baseline, checksum find.

- **One source walk.** `find`'s `-readable` alternation puts a readability flag in the same pass, so the unreadable check, the ownership check and the verification baseline all derive from that single output. The source is quiesced, so re-walking could never have learned anything new.
- **`du -sh` deleted** — a full metadata walk for a log line. The walk yields an entry count, which predicts duration far better than a byte total.
- **Parallel checksums** (8 workers): latency-bound, not CPU- or bandwidth-bound, so overlapping round-trips is close to a linear win. Output is sorted after collection since workers finish out of order.
- **`phase:` log lines with elapsed seconds** (walk / wipe / copy / metadata / content) so the next slow drain is attributable instead of guesswork.
- **`minTargetSize`** floors the target volume: a class that prices IOPS per gigabyte gives a faithfully-sized 1Gi volume a 1Gi share of IOPS, throttling exactly the per-file work the copy does. Empty by default; **`10Gi` is the right value for IBM VPC block**, which has that minimum anyway.

## Tests

Four new cases, each pinning a property that failed in production: the target lands on the **cluster default even when `agent.base.storageClass` names the file class**; an **already-RWO volume on the wrong class is still drained**; a volume **already on the target class is left alone** (convergence); and **target == source class is refused** with no gate and no bytes copied, while the explicit override still works.

Controller: 7/7 packages green, `go vet` + staticcheck clean, helm lint clean.

## What I got wrong

I flagged this exact risk in prose when asked whether the migration works with IBM's classes — "flip `base.storageClass` or targets land RWO on the file class and you're still on NFS" — and then shipped code that silently accepted precisely that, and never checked the deployed values before the drain. A guard belongs in the code, which is what this adds.

### DIFF
--- a/deploy/helm/platform/values.yaml
+++ b/deploy/helm/platform/values.yaml
@@ -1028,8 +1028,11 @@ controller:
     #  - size: "5Gi"
     #    target: 2
 
-  # -- One-time RWX -> RWO workspace-volume migration (#2988). For every agent
-  # whose workspace PVC is still ReadWriteMany the controller forces the agent
+  # -- One-time workspace-volume migration onto ordinary single-writer storage
+  # (#2988). A workspace is migrated when it is still ReadWriteMany, OR when it
+  # sits on a storage class other than `targetStorageClass` below — the second
+  # trigger is what moves a volume that was flipped to ReadWriteOnce but left
+  # on the shared filesystem. The controller forces the agent
   # down, copies the volume onto a fresh ReadWriteOnce PVC in a Job
   # (content- and metadata-verified), re-points the StatefulSet, deletes the
   # old volume, and restores the agent's prior run state. Interrupt-safe and
@@ -1047,6 +1050,23 @@ controller:
   # through the first drain; `concurrency: 1` makes that easy to follow.
   storageMigration:
     enabled: true
+    # -- Storage class migrated workspaces land on: the DESTINATION of the
+    # migration. Empty (default) = the cluster's default class, i.e. ordinary
+    # storage, which is the point of the exercise.
+    #
+    # Deliberately NOT `controller.agent.base.storageClass`: on an install
+    # that has not migrated yet, that field still names the shared filesystem
+    # being drained, so inheriting it would force-restart every agent and copy
+    # every byte only to arrive back on NFS with a different access mode. Set
+    # this to the block class agents should end on, or leave it empty for the
+    # cluster default — and fix `agent.base.storageClass` (and
+    # `warmPool.storageClass`) too, or newly-created agents keep landing on
+    # the old backend.
+    targetStorageClass: ""
+    # -- Permit a migration whose target class is the class the workspace is
+    # already on. Refused by default: it changes the access mode and nothing
+    # else, at the cost of a forced restart and a full copy per agent.
+    allowSameStorageClass: false
     # -- How many agents migrate at once. Each migration reads its source
     # volume in full three times (copy + two verification passes); this cap
     # is the trade between fleet drain time and read pressure on the shared
--- a/deploy/helm/platform/values.yaml
+++ b/deploy/helm/platform/values.yaml
@@ -1067,10 +1067,20 @@ controller:
     # already on. Refused by default: it changes the access mode and nothing
     # else, at the cost of a forced restart and a full copy per agent.
     allowSameStorageClass: false
-    # -- How many agents migrate at once. Each migration reads its source
-    # volume in full three times (copy + two verification passes); this cap
-    # is the trade between fleet drain time and read pressure on the shared
-    # filesystem being drained — lower it if the filer struggles.
+    # -- Floor for a migrated volume's size; empty keeps the source's request.
+    # Set this when the target class prices IOPS PER GIGABYTE (IBM's block
+    # tiers do): a faithfully-sized 1Gi volume then inherits a 1Gi share of
+    # IOPS, and a copy that is bound by per-file operations crawls. Such
+    # classes usually have a minimum volume size anyway, which a smaller
+    # request is silently rounded up to. "10Gi" suits IBM VPC block.
+    minTargetSize: ""
+    # -- How many agents migrate at once. Each migration walks its source
+    # once and reads it twice (copy + checksum), so this cap is the trade
+    # between fleet drain time and pressure on the shared filesystem being
+    # drained — lower it if the filer struggles. Duration is driven by FILE
+    # COUNT, not bytes: every entry is a round-trip and, on an IOPS-capped
+    # class, an IOP. The copy Job logs `phase:` lines with elapsed seconds so
+    # a slow drain can be attributed rather than guessed at.
     concurrency: 10
     # -- Reconcile pass interval.
     interval: "30s"
--- a/packages/controller/pkg/config/agent_config.go
+++ b/packages/controller/pkg/config/agent_config.go
@@ -253,6 +253,14 @@ type StorageMigration struct {
 	// copies every byte to arrive at the same backend — so it is refused with
 	// a loud log naming the remedy.
 	AllowSameStorageClass bool `json:"allowSameStorageClass,omitempty"`
+	// MinTargetSize floors the size of a migrated workspace volume. Empty
+	// (the default) keeps the source's request verbatim. Set it where the
+	// target class prices IOPS PER GIGABYTE — IBM's block tiers do — because
+	// there a faithfully-sized 1Gi volume also inherits a 1Gi share of IOPS
+	// and the copy crawls; such classes also tend to have a minimum volume
+	// size of their own, which a smaller request is silently rounded up to
+	// anyway.
+	MinTargetSize string `json:"minTargetSize,omitempty"`
 	// AllowOwnershipRemap lets the copy proceed when the source workspace
 	// holds entries owned by a uid other than the agent's. The copy is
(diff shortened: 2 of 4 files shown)


Tomas2D (approved):

dam-code-guardian:
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `edb8c85`

> ℹ️ Posted as a comment, not a review: this PR merged (`545aa6c`) about three minutes into the review, so GitHub no longer accepts a review event against `edb8c85`. Findings below were produced against the PR head and re-verified byte-identical on the merge commit — the two 🟡s are follow-up material, not merge blockers.

## PR #3083: fix(controller)!: migration destination is its own class, defaulting to the cluster default
**Author:** jezekra1 | **Branch:** `fix/migration-target-class#locki-lpaitkhu` → `main` | **Changes:** +384 −62 (4 files)

### Summary
Splits the migration's destination storage class out of `agent.base.storageClass` into its own `storageMigration.targetStorageClass` (empty = cluster default), adds a second migration trigger on "wrong storage class" so volumes stranded RWO-on-NFS by the earlier broken drain are recovered, refuses a target equal to the source class unless `allowSameStorageClass`, and cuts the copy Job from five source walks to one with parallel checksums plus a `minTargetSize` floor. The root-cause analysis and the decision to make the destination independent are both correct — `resolveTargetClass` returning the resolved *name* while still leaving `storageClassName` unset on the PVC is the right way to get "cluster default" without hardcoding it, and routing `ReleaseGated` through the same `migrationReason` predicate closes the gap where a class-only-pending workspace would have been mistaken for a completed flip.

### Findings

- 🟡 **Warning:** The metadata verification pass can't detect a wrong symlink target when `allowOwnershipRemap: true`, and silently mis-parses any path containing a `|`. The walk emits 6 pipe-delimited fields but `%p` (path) can itself contain `|`, so `cut -d'|' -f2,3,5,6` splits on the *first* pipes it finds — for `./dir/we|ird` the symlink-target field is dropped and the path field is truncated. Verified locally: a source symlink pointing at `real-a` and a target symlink pointing at `real-b` produce **byte-identical** `src.meta`/`dst.meta`, so `cmp` passes. Symlinks are also excluded from the checksum pass (`find -type f`), so nothing else catches it — and the source volume is deleted at flip on the strength of these two comparisons. Only the remap path is affected (`"2-"` keeps the tail intact, so the default config is safe), and it needs a `|` in a workspace path to trigger, which is why this is 🟡 not 🔴. Fix: put `%l` first, or delimit with `\0`/tab, or compare full lines and drop the uid column with `sed 's/^\([^|]*|[^|]*\)|[^|]*|/\1|/'` instead of a field cut. (`packages/controller/pkg/reconciler/storage_migration.go:780`)

- 🟡 **Warning:** `resolveTargetClass` silently degrades to access-mode-only migration when it can't read StorageClasses, which on an unmigrated install re-creates exactly the bug this PR fixes. The RBAC grant for `storageclasses` is annotated in the chart as *best-effort* ("the manager degrades to a logged warning without this, e.g. under namespace-scoped RBAC"). If that list call fails — or no class carries the default annotation — the function returns `("", false)`, `migrationReason` falls back to the RWX trigger alone, and `ensureTargetPVC` creates the target with no `storageClassName`, which the cluster fills in with *its* default. On an install whose default class is the shared filesystem, that is a full fleet drain from RWX to RWO on the same NFS backend: the precise failure this PR exists to prevent, reachable through a `Warn` line rather than a refusal. Since `enabled: true` is the shipped default (`values.yaml:1052`) and `targetStorageClass: ""` is too (`values.yaml:1065`), consider failing closed — skip admitting *new* work when the target can't be resolved and the class is not explicitly configured — rather than proceeding on the access-mode trigger. (`packages/controller/pkg/reconciler/storage_migration.go:355`)

- 🟢 **Suggestion:** `migrationReason` leaves a no-class (statically provisioned) volume alone, which the comment justifies well — but combined with the RWX branch above it, an **RWX volume with no storage class** still migrates and lands on the cluster default, with no same-class guard possible (`srcClass` is empty, so `sameClass` is false by construction). That is probably the intent, but it is the one path where the "refuse a pointless migration" protection cannot engage; worth an explicit line in the comment so the asymmetry reads as deliberate. (`packages/controller/pkg/reconciler/storage_migration.go:388`)

- ✅ **Looks good:** The `-P8` parallel checksum pipe is safe. Parallel `xargs` workers writing to one pipe is a classic interleaving hazard, so I tested it — 6 runs over 6000 files at ~390 bytes/line (well past `PIPE_BUF`, deep pnpm-shaped paths) produced zero malformed lines and output identical to the serial baseline, because each `md5sum` writes with a single buffered `write(2)` per batch. Sorting after collection rather than sorting the input is also the right call, and drops a `sort -z` pass.

- ✅ **Looks good:** Routing `ReleaseGated`'s "source still present" test through the shared `migrationReason` predicate instead of a bare `isRWX` — without it, a workspace pending only a class move would have been read as a completed flip and sent into `finishFlip`, which deletes superseded sources. The `for i := range` / `p := pvcs.Items[i]` rewrite to make `&p` safe is correct too.

- ✅ **Looks good:** Test coverage genuinely pins the production failure rather than the implementation: target-lands-on-cluster-default *while* `AgentBase.StorageClass` names the file class, already-RWO-on-wrong-class still drains, already-on-target converges, and same-class refused with the override still working. `TestStorageMigration_TargetSizeFloorNeverShrinks` covers the floor's non-shrinking direction, which is the easy one to get wrong.

**Verdict: minor drift**

❌ **Check 1 — architecture-page drift** — `docs/architecture/persistence.md` describes the migration's trigger and terminal condition in terms of the access mode alone, which this PR changes:

- `persistence.md:125` — "for each agent still on an RWX volume it forces the pair down … Throttled fleet-wide; **a no-op once no RWX workspace volume remains**." Both halves are now wrong: the trigger is RWX **or** wrong-storage-class, and the drain is not a no-op once RWX is gone — an RWO volume on a non-target class still migrates (that is the recovery path this PR adds). The page also omits that a misconfigured target is now *refused*, which is a behavior an operator reading this page would want to know.
- `persistence.md:103` — "(Workspaces created before the RWO cutover sit on `ReadWriteMany` volumes until the storage migration below drains them.)" — the parenthetical now under-describes what the migration drains.
- `persistence.md:3` — `Last verified: 2026-07-29` needs bumping with the edit (check 2 follows from check 1 here; the page was not touched in this PR).

Suggested reframing for :125, at page level and without naming config keys: the migration moves a workspace whose **access mode or storage backend** is not the intended one, and converges when every workspace sits single-writer on the intended backend — plus one clause that a destination equal to the volume's current backend is refused rather than performed.

_No other architecture page mentions the migration trigger, so the drift is contained to this one page._

### Verdict
COMMENT — the design fix is right and well-tested; two 🟡s worth a follow-up (a symlink-verification hole on the `allowOwnershipRemap` path, and a fail-open degradation that can reproduce the original bug), plus one architecture page whose stated trigger and terminal condition this PR invalidates.

---

Issue #3053, opened 2026-07-29
UI - unify empty states, loading skeletons and dialog footers


kapetr:
## What

Three families of "nothing here yet / still loading" surfaces each have several hand-built versions:

- **Empty states** — 5 different shapes across roughly 17 places, even though a shared empty-state component already exists and is used by 3 of them.
- **Loading placeholders** — 4 separate skeleton implementations, plus 6 different spellings of a plain "Loading…" line.
- **Dialog footers** — the "Cancel + Save (Saving…)" pair is rebuilt in 9 dialogs, and they disagree on whether Cancel is a ghost or an outline button.

## Why

These are the surfaces a user sees when something is missing or slow, so inconsistency here reads as unfinished. They're also where the same copy and spacing decisions get re-made every time a page is added.

## Done when

- One empty-state component, one skeleton family and one dialog-footer arrangement, used across all the places listed above.
- Cancel and submit buttons look the same in every dialog.

PR #3084, opened 2026-07-30
refactor(ui): unify dialog footers and close buttons

kapetr:
Makes every dialog end the same way, gives them a close button in a consistent place, and starts on the empty states.

## Dialogs

The "Cancel + confirm" pair was rebuilt in ten dialogs. It now comes from one component, which settles three things that used to be decided per dialog:

- **Cancel is always an outline button.** Seven dialogs already were; the three in Artifacts were ghost.
- **Cancel can never submit the form it sits in.** Seven of the ten are inside a form, where that depends on remembering an attribute.
- **Cancel stays clickable while the confirm is running**, so a dialog is always escapable. One disabled it mid-submit.

Three real problems fell out of the conversion: the schedule form showed a bare `…` while saving where every other dialog says "Saving…"/"Creating…", and two submit buttons hand-rolled the dimming the button already does when disabled.

## Close buttons

The ✕ now sits 16px off the panel corner rather than lining up with the title, and is 16px like every other close in the app — it was the last one at 18px. Seven dialogs that had no ✕ at all now have one.

Two deliberately don't: the revoke confirmation, because it asks for a decision, and the one-time token reveal, because the token is shown once and is never recoverable — a corner ✕ would make losing it a single mis-click.

## Page titles

A page's heading moved depending on whether the header had a button. The button is 40px and the title 35px, so the button grew the row and re-centred the title in it, dropping the heading 2.5px and everything below it 5px. Because every list view only shows its action once it has content, this also showed as a jump between the loading skeleton and the loaded page. The title now holds its position either way.

## API keys

The page had a button in the top corner *and* a dashed box telling you to create a key with no way to do it. It now works like the other list pages: when empty, the empty state carries the action; once keys exist, the button sits under the description.

## What's left on #3053

This is the dialog-footer third of the issue plus one empty state. Still to do:

- one component for the "panel list is empty" line, repeated identically in four places
- the experiments page hand-inlines the shared page empty state's markup
- two dashed "drop something here" boxes that disagree on padding, spacing and icon size
- the loose "Loading…" lines, including one lowercase

Refs #3053


packages/ui/src/modules/sandboxes/components/channels/slack-channel-modal.tsx
```diff
@@ -61,23 +61,13 @@ export function SlackChannelModal({
...
-          </Button>
-          <Button
-            type="submit"
-            disabled={isSubmitting}
-            data-testid="slack-save"
-          >
-            {isSubmitting ? "Saving…" : editing ? "Save" : "Connect"}
-          </Button>
-        </DialogFooter>
+        <DialogActions
```
dam-code-guardian, line 76: 🟡 **Warning:** This drops the `disabled={isSubmitting}` that `main` had on Cancel, and here that attribute was load-bearing rather than the inconsistency it looks like.

`DialogActions` keeps Cancel enabled so a dialog is always escapable — good default — but `onCancel` is `onClose`, and the submit is not abortable. `useSlackChannelForm`'s rebind path is two sequential mutations:

```
await disconnectSlack.mutateAsync({ id: agent.id });
await connectSlack.mutateAsync(connectPayload);
```

Clicking Cancel between them unmounts the dialog while the pair is half-applied: the old channel is unbound, the new one never connects, and there is no longer any UI showing that. The user's reasonable reading of Cancel ("nothing happened") is the opposite of what happened.

Options, roughly in order of preference: gate `onCancel` on `pending` for this dialog, give `DialogActions` an opt-out for non-idempotent submits, or make the rebind atomic server-side so a mid-flight dismiss can't strand it.

packages/ui/src/modules/api-keys/components/api-keys-list.tsx
```diff
@@ -53,28 +50,30 @@ export function ApiKeysList() {
...
-            <ApiKeyRow
-              key={k.id}
-              apiKey={k}
-              onRevoke={(id, name) => setRevokeTarget({ id, name })}
-              revoking={revokeApiKey.isPending && revokeTarget?.id === k.id}
-            />
-          ))}
-        </ul>
+        <>
+          <Button className="mb-4" onClick={() => setCreateOpen(true)}>
```
dam-code-guardian, line 63: 🟢 **Suggestion:** The three other pages that use `PageEmptyState` keep the create button in `PageHeader actions`, conditional on non-emptiness — `agents/views/list-view.tsx:66-75`, `knowledge-bases/views/knowledge-bases-list-view.tsx:40-45`, `artifacts/views/artifacts-view.tsx:112-113` — and drop `description` when empty too. Since the PR body's goal is for this page to work like the other list pages, matching that shape also removes the need for the `mb-4` margin utility:

```suggestion
        <>
          <ul className="space-y-2">
```

(paired with moving the button up into `PageHeader actions={keys.length > 0 ? <Button onClick={() => setCreateOpen(true)}>Create key</Button> : undefined}`)

packages/ui/src/components/modal.tsx
```diff
@@ -171,6 +174,56 @@ export function DialogFooter({ children, className }: DialogRegionProps) {
   );
 }
 
+interface DialogActionsProps {
+  onCancel: () => void;
```
dam-code-guardian, line 184: 🟡 **Warning:** The `pendingLabel ?? label` fallback is right for dialogs that genuinely have nothing different to say while pending, but it also swallows the case where a caller sets `pending` and forgets `pendingLabel` — the button then looks idle for the whole request.

That is a live risk in `upload-skills-tab.tsx`, where `label` is computed (`Add {n} skills`) and `pendingLabel="Adding…"` is what used to be folded into that same expression. Nothing now ties them together, so a later edit to the label logic can silently drop the pending state.

Requiring `pendingLabel` whenever `pending` can be true would catch it at the type level; a discriminated pair (`{pending: false} | {pending: boolean, pendingLabel: string}`) is the cheap version.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `458b0080`

## PR #3084: UI - unify dialog footers and close buttons
**Author:** kapetr | **Branch:** refactor/3053-empty-states-skeletons-footers → main | **Changes:** +180 −159 (14 files)

### Summary
Extracts the Cancel+confirm pair repeated across ten dialogs into one `DialogActions` component, standardizes the close ✕ (16px, offset to the panel corner), reserves the page-header action row's height so titles don't shift, and converts the API-keys empty state to the shared `PageEmptyState`.

### Findings
- 🟡 **Warning:** Cancel now stays enabled mid-submit, but for the seven form dialogs the escape it offers is cosmetic — `onCancel` is `onClose` and no mutation is abortable, so the request continues after unmount. In `slack-channel-modal.tsx` this reverses a deliberate `disabled={isSubmitting}` from `main`, and that submit is a two-step rebind (disconnect → connect); dismissing between the steps leaves the channel disconnected with nothing showing it. (`packages/ui/src/modules/sandboxes/components/channels/slack-channel-modal.tsx:64`)
- 🟡 **Warning:** `pendingLabel` falls back to `label`, which also applies silently when a caller passes `pending` but no `pendingLabel`. Worth noting because `upload-skills-tab.tsx`'s `label` is itself computed (`Add N skills`), so a future edit dropping `pendingLabel` degrades to a button that looks idle while submitting. (`packages/ui/src/components/modal.tsx:178`)
- 🟢 **Suggestion:** The populated-state button sits in the content area with `mb-4`; the three sibling `PageEmptyState` pages all use `PageHeader actions` gated on non-emptiness instead. Matching them drops the margin utility and the always-on `description`. (`packages/ui/src/modules/api-keys/components/api-keys-list.tsx:63`)
- ✅ **Looks good:** The two deliberate ✕ omissions are the right calls, and `reveal-token.tsx` is correctly left untouched — the once-only token stays behind an explicit "Done".
- ✅ **Looks good:** Three hand-rolled `cn(!isValid && "opacity-50")` dimmings removed in favor of the Button's own `disabled:opacity-50`; `schedule-form-modal`'s bare `…` fixed to `Saving…`.

**Verdict:** `aligned`

✅ No findings. The diff is presentation-layer only and changes no subsystem behavior described under `docs/architecture/`. The one architecture-documented behavior in range — API key plaintext shown once, never recoverable (`security-and-credentials.md:155-161`) — is preserved.

The extraction is textbook: `DialogActions` is the "same markup repeats with small variations" case ten times over, props are an interface destructured with defaults, no `React.FC<>`, imports cleaned at every site. State lineage untouched and correct; styling stays pure Tailwind with no static `style={{}}`.

- 🟡 **Warning:** Cancel-during-pending is not a real escape for non-abortable submits; `slack-channel-modal.tsx:64-69` regresses a load-bearing `disabled`.
- 🟢 **Suggestion:** `api-keys-list.tsx:63` diverges from the three sibling list pages' `PageHeader actions` pattern.
- ✅ No further findings.

### Verdict
COMMENT — a clean, well-reasoned consolidation; the one thing worth a second look is the Slack rebind dialog, where mid-submit Cancel is now reachable.

---

packages/ui/src/modules/api-keys/components/create-api-key-dialog/create-form.tsx
```diff
@@ -84,7 +83,7 @@ export function CreateApiKeyForm({ onCreated, onCancel }: Props) {
     // Body/Footer sit directly in the Modal's flex column — otherwise the form
     // wrapper breaks the column and the scrollable body, overflowing the panel.
     <form onSubmit={handleSubmit} className="contents">
-      <DialogHeader title="Create API key" />
+      <DialogHeader title="Create API key" onClose={onCancel} />
```
PetrBulanek: **The new ✕ here can destroy a key's plaintext.** With the create in flight, the submit correctly reads "Creating…" and disables, but the ✕ and Cancel stay live. Clicking the ✕ closes the dialog, the key is still created, and the one-time token is never shown — the user has to revoke and start over. Reproduced on the dev cluster.

This is the exact loss the PR avoids by omitting the ✕ from the reveal step; it's just one step earlier. Either drop `onClose` from this header, or give `DialogActions` a `cancelDisabled` so both dismissals can be gated while pending.

packages/ui/src/components/modal.tsx
PetrBulanek, line 124: **Opening a dialog now focuses the ✕ instead of the first field.** The focus trap grabs the first focusable in the panel, which the header ✕ now is — so a keyboard user's first Enter or Space dismisses the dialog. Confirmed on Upload artifact (used to focus "Choose a file…") and Share (used to focus the sharing switch).

Hits the six newly-✕'d dialogs with no `autoFocus` on their first control: `upload-artifact-dialog`, `share-dialog`, `schedule-form-modal`, `slack-channel-modal`, `publish-skill-modal`, `artifact-preview-dialog`. `folder-dialog` and `github-source-tab` are fine because they have it.

Fix either per dialog (`autoFocus` on the first field) or once here, by having the trap prefer the first focusable in the body over the header close.

packages/ui/src/modules/artifacts/components/artifact-preview-dialog.tsx
PetrBulanek: **Two ways out, and the footer one is the last ghost dismiss in the app.** This dialog now has the corner ✕ *and* a ghost "Close" sitting next to an outline "Download". `share-dialog`'s ghost Close was converted to outline in this same diff, so this is the one holdout.

Cleanest is to drop it now that the ✕ covers dismissal.

packages/ui/src/modules/schedules/forms/schedule-form-modal.tsx
```diff
@@ -294,14 +294,12 @@ export function ScheduleFormModal({
...
-            Cancel
-          </Button>
-          <Button type="submit" disabled={mutation.isPending}>
-            {mutation.isPending ? "…" : existing ? "Save" : "Create"}
-          </Button>
-        </DialogFooter>
+        <DialogActions
+          onCancel={onClose}
+          label={existing ? "Save" : "Create"}
+          pendingLabel="Saving…"
```
PetrBulanek: **Creating a schedule says "Saving…".** The bare `…` is fixed, but the button goes "Create" → "Saving…", where the API key dialog says "Creating…". Verified on screen.

```suggestion
          pendingLabel={existing ? "Saving…" : "Creating…"}
```

PetrBulanek (dismissed):

packages/ui/src/components/modal.tsx
```diff
@@ -190,7 +259,16 @@ export function useFocusTrap(containerRef: RefObject<HTMLElement | null>) {
...
       if (!container.contains(document.activeElement)) {
-        container.querySelector<HTMLElement>(FOCUSABLE_SELECTOR)?.focus();
+        const focusables = Array.from(
+          container.querySelectorAll<HTMLElement>(FOCUSABLE_SELECTOR),
+        );
+        // Skip the header's ✕ when there is anything else to land on: it comes
+        // first in the DOM, so focusing it would make the opening keystroke
+        // dismiss the dialog.
+        const target =
+          focusables.find((el) => !el.hasAttribute("data-dialog-close")) ??
```
dam-code-guardian, line 283: 🟡 **Warning:** This skips the header ✕, but `FOCUSABLE_SELECTOR` still matches `display:none` elements — so `upload-artifact-dialog` is still broken, and it is the dialog the report reproduced on.

Its body starts with a hidden file input (`upload-artifact-dialog.tsx:68-70`), which precedes the "Choose a file…" button in the DOM. `find()` returns that input, and `.focus()` on a `display:none` element is a no-op — so focus never enters the panel and stays on `<body>`. Two knock-on effects:

- `grab()` re-runs every frame for 500 ms and never succeeds, since `container.contains(document.activeElement)` stays false.
- The `onKey` Tab handler builds its list from the same unfiltered selector, so `first` is the hidden input there too — the first Tab press also goes nowhere.

`upload-skills-tab.tsx:99-108` has the same hidden-input shape, but its dashed drop-zone `<button>` comes first, so it happens to be unaffected.

A visibility filter fixes both the initial grab and Tab cycling:

```suggestion
        const focusables = Array.from(
          container.querySelectorAll<HTMLElement>(FOCUSABLE_SELECTOR),
        ).filter((el) => el.offsetParent !== null);
```

(the same filter is worth applying to `onKey`'s list, which has always had this hole — not a regression from this PR, but it becomes reachable now that the header ✕ is no longer first.)

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `e63067bb`

## PR #3084: refactor(ui): unify dialog footers and close buttons
**Author:** kapetr | **Branch:** refactor/3053-empty-states-skeletons-footers → main | **Changes:** +219 −163 (14 files)

### Summary
Adds `closeDisabled`/`cancelDisabled` gating, makes `pendingLabel` required, and reorders the focus trap to skip the header ✕ — addressing all four of PetrBulanek's findings plus both 🟡s from the previous review.

### Changes since last review
Previous HEAD: `458b0080` (2026-07-30T19:00:24Z) — verdict COMMENT

- ✅ **Fixed:** Slack rebind mid-submit dismissal — both ✕ and Cancel now gated on `isSubmitting` (`slack-channel-modal.tsx:47,81`)
- ✅ **Fixed:** `pendingLabel ?? label` fallback removed; the prop is required, so a pending button can't look idle (`modal.tsx:189`)
- ✅ **Fixed:** API-key create ✕ can no longer strand a key whose plaintext is never shown (`create-form.tsx:90,151`)
- ✅ **Fixed:** ghost "Close" dropped from the artifact preview footer (`artifact-preview-dialog.tsx:143`)
- ✅ **Fixed:** schedule create says "Creating…", not "Saving…" (`schedule-form-modal.tsx:300`)
- 🆕 **New:** focus trap still lands on a hidden input in `upload-artifact-dialog` (`modal.tsx:269`)
- 🆕 **New:** `data-dialog-close` is an untyped contract (`modal.tsx:132`)

### Findings
- 🟡 **Warning:** Focus trap skips `[data-dialog-close]` but not `display:none` elements, so `upload-artifact-dialog` still opens with focus outside the panel (`packages/ui/src/components/modal.tsx:269`)
- 🟢 **Suggestion:** `data-dialog-close` is a load-bearing untyped DOM contract; note it in the `useFocusTrap` doc comment (`packages/ui/src/components/modal.tsx:132`)

**Verdict:** aligned

✅ No findings. Presentation-layer only; the `ui` subsystem's coverage (`platform-topology.md:71-73`) sits at framework level and no architecture page mentions dialogs. Checks 2–7 inert — no `docs/` files in the diff, no ADR status change.

Both 🟡s from the previous review are fixed; nothing carried over.

The delta is well-executed. `closeDisabled`/`cancelDisabled` are correctly optional-with-default, the required `pendingLabel` is the type-level fix the previous review asked for, and gating dismissal via props rather than lifting state into the dialog keeps state lineage clean. Both new props document the *why*, and all 11 `DialogActions` call sites pass `pendingLabel`, so the required-prop tightening compiles.

New findings as listed above.

### Verdict
COMMENT — six of six prior findings genuinely fixed; the focus-trap fix is one visibility check short of covering the dialog it was reported against.

---

packages/ui/src/components/modal.tsx
```diff
@@ -123,11 +127,16 @@ export function DialogHeader({
               variant="ghost"
               size="icon-sm"
               onClick={onClose}
+              disabled={closeDisabled}
               aria-label="Close"
+              data-dialog-close
```
dam-code-guardian, line 132: 🟢 **Suggestion:** `data-dialog-close` creates an untyped contract between `DialogHeader` and `useFocusTrap`. Both live in this file so it's a reasonable pragmatic choice, but nothing ties them together for a reader or the compiler — a second close affordance added elsewhere silently opts out of the skip, and the attribute looks like inert test scaffolding next to `data-testid`.

The ✕ is already distinguishable by `aria-label="Close"`, so the attribute is not strictly needed; if it stays, a line in `useFocusTrap`'s doc comment saying the attribute is load-bearing would keep the next editor from dropping it.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `9e47ee2c`

## PR #3084: refactor(ui): unify dialog footers and close buttons
**Author:** kapetr | **Branch:** refactor/3053-empty-states-skeletons-footers → main | **Changes:** +235 −167 (14 files)

### Summary
One commit since the last review, entirely in `modal.tsx`: the focus trap's two raw selector queries move behind a `focusablesIn()` helper that filters on `getClientRects().length > 0`, and `useFocusTrap`'s doc comment now records `data-dialog-close` as load-bearing.

### Changes since last review
Previous HEAD: `e63067bb` (2026-08-03T09:32:08Z) — verdict COMMENT

- ✅ **Fixed:** Hidden file input no longer captures the initial focus grab, and both the grab and Tab cycling share one filtered list, so `upload-artifact-dialog` opens on "Choose a file…" (`modal.tsx:252`)
- ✅ **Fixed:** `data-dialog-close` is documented as load-bearing on `useFocusTrap` (`modal.tsx:263`)

### Findings

**Verdict:** aligned

✅ No findings. The delta is one presentation-layer file; no `docs/` files in the diff and no ADR status change, so checks 2–7 are inert. No architecture page describes dialog focus behavior — the `ui` subsystem's coverage (`platform-topology.md:71-73`) stays at framework level.

Both findings from the previous review are fixed; nothing carried over.

The fix is better than what the previous review proposed. `getClientRects()` over the suggested `offsetParent !== null` is the correct call and the comment says why — `offsetParent` is null for `position: fixed` elements, and `Modal` renders inside a `fixed` overlay, so the suggested predicate could have dropped genuinely focusable content. Extracting `focusablesIn()` also closes the pre-existing `onKey` hole in the same place rather than fixing the grab alone, which is why Tab cycling now starts on a real element too.

✅ No further findings.

### Verdict
APPROVE — the last two findings are resolved, and the focus-trap fix corrects the predicate the review suggested rather than copying it.

---

PR #3088, opened 2026-07-30
chore(helm): remove the stale-netpol pre-upgrade migration job

jezekra1:
Removes `deploy/helm/platform/templates/migrate-stale-netpols.yaml` — a `pre-upgrade` hook Job that swept three one-time leftovers:

1. **Legacy pair-key NetworkPolicies** written by the pre-ambient controller (`<id>-egress` / `<id>-gateway-egress`, no `managed-by` label). They only admitted the application port, so they blocked HBONE :15008 once the agent namespace flipped to ambient.
2. **Stale `<id>-gateway-allow` AuthorizationPolicies** from when the agent still participated in the mesh. After the agent opted out at the pod level, the ALLOW denied the agent → gateway hop instead of admitting it.
3. **Per-agent Telegram bot-token Secrets**, orphaned when Telegram became a platform channel with a single bot token in the api-server env.

All three shapes are gone from live clusters, so every run of the Job is now a no-op. Fresh installs never ran it at all (`pre-upgrade` only).

The template was fully self-contained — the Job plus its ServiceAccount, Role and RoleBinding, which granted `list`/`delete` on NetworkPolicies, AuthorizationPolicies and Secrets in the agent namespace. Nothing else in the repo referenced it, so dropping the file also drops that standing RBAC grant.

## Verification

`mise run helm:check` — lint and render both clean, 51 valid resources, 0 errors.


dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `3bf82ab`

## PR #3088: chore(helm): remove the stale-netpol pre-upgrade migration job
**Author:** jezekra1 | **Branch:** `chore/remove-netpol-migration-job#locki-9stt2amv` → `main` | **Changes:** +0 −150 (1 file)

### Summary
Deletes the self-contained `pre-upgrade` hook template that swept three one-time leftovers (legacy pair-key NetworkPolicies, stale `<id>-gateway-allow` AuthorizationPolicies, per-agent Telegram bot-token Secrets). All three shapes are reported gone from live clusters, so every run is a no-op; dropping the file also drops a standing `list`/`delete` RBAC grant on NetworkPolicies, AuthorizationPolicies and Secrets in the agent namespace.

### Findings
- 🟡 **Warning:** The removal is only safe for clusters that have already run the hook at least once. Because the hook was `pre-upgrade`-only, a release still sitting on a chart from before the job landed (it was in-tree from `137a1f5`, 2026-07-27, to now) upgrades straight past the sweep and keeps the legacy pair-key NetworkPolicies — which, per the deleted template's own comment, block ambient HBONE :15008 and leave every existing agent without connectivity until recreated. The window is genuinely narrow (~3 days) and the PR body asserts live clusters are clean, but nothing in the repo pins a minimum-upgradable chart version, so the safety argument lives only in the PR description. Worth a line in the release notes so an operator upgrading a long-stale install knows to sweep manually.
- ✅ **Looks good:** Verified no dangling references — `migrate-stale-netpols` appears nowhere else in the tree, and no `_helpers.tpl` alias, values key, or task references the job or its ServiceAccount/Role/RoleBinding. The chart's other hook (`postgres-migrate-roles.yaml`) is independent.
- ✅ **Looks good:** Dropping the RBAC is a real security win, not just cleanup: the Role granted namespace-wide `list`/`delete` on Secrets (not name- or label-scoped — the label filter lived only in the container's `kubectl` args), so a compromised or misconfigured hook pod could delete any Secret in the agent namespace. That standing grant existed on every upgrade; removing the template removes it.
- ✅ **Looks good:** The retirement is consistent with the documented target state — ADR 042 already records `<id>-gateway-allow` as retired and explains why the agent has no SPIFFE principal, so sweeping code for it is genuinely dead. Likewise the api-server no longer creates per-agent Telegram channel Secrets (`TYPE_CHANNEL_SECRET` in `packages/api-server/src/modules/agents/infrastructure/labels.ts:24` is now unreferenced) and the bot token is a single platform-level env value.
- ✅ **Looks good:** Deleting the whole file is the right granularity — no partial-sweep state is left behind, and `mise run helm:check` verification (lint + render clean, 51 resources) is recorded in the PR body.

✅ No findings. The removed template is a one-time upgrade hook with no footprint under `docs/architecture/`: no architecture page references the job, the sweep, or the three swept resource shapes, and the pages that do discuss NetworkPolicies/AuthorizationPolicies (`platform-topology.md`, `security-and-credentials.md`) describe the current controller-rendered shapes, which this PR does not touch. No architecture page was edited (so no `Last verified:` bump is due) and no ADR changed state.

### Verdict
COMMENT — Clean, well-scoped deletion that also sheds an over-broad standing RBAC grant; the only caveat is the unstated upgrade-skew assumption, which deserves a release note rather than a code change.

---

pilartomas (approved):

Issue #3092, opened 2026-07-30
Web search performed by the model proxy is invisible to the platform


Tomas2D:
## Context

Seen on agents connected through the IBM LiteLLM proxy connector, against LiteLLM 1.85.5. Surfaced when a user noticed the chat was missing tool activity it should have been showing.

The capability itself is wanted: we want agents on that backend to be able to search the web, and the proxy providing it is fine. The earlier attempt to work around it (#1087, now reverted) blocked the built-in web tools and shipped DuckDuckGo-backed replacements — that turned out to suppress the proxy's search rather than fix anything, so it is gone.

What remains is that the platform cannot see or govern the search.

## Problem

The proxy takes the web search tool the harness offers the model, swaps it for one of its own, and runs the search itself. The agent never performs the search, so from the platform's side it never happened.

- **No approval, no record.** The user is never asked, and the turn carries no trace of a search having run. An operator reviewing the conversation cannot tell that the agent reached the internet.
- **The fetch leaves our controls.** The search runs from the proxy's network, not the pod's, so the sandbox's egress rules never see it — the whole point of routing agent traffic through approvals.
- **Replies stop streaming.** Any turn this affects arrives in a single burst at the end instead of progressively, which reads as the agent having hung.

## Expected Behavior

Web access an agent makes is visible in the chat as a step the user can see, and is subject to the same approval and egress rules as anything else the agent reaches out to.

## Actual Behavior

The reply contains fresh web content. The chat shows no search step, no permission prompt appears, and the reply arrives all at once.

## Steps to Reproduce

1. Connect an agent to a model through the LiteLLM proxy connector.
2. Ask it something that needs current information from the web.
3. Observe the answer contains up-to-date web content while the chat shows no web search step and no approval was requested.

PR #3093, opened 2026-07-30
revert(claude-code): drop the cc-websearch DDG fallback for the web tools

Tomas2D:
Reverts the #1087 workaround. Known limitation tracked in #3092.

## Why

#1087 assumed the built-in `WebSearch` / `WebFetch` are simply broken on a non-Anthropic backend, and covered them with a `PreToolUse` hook plus DuckDuckGo-backed replacement skills.

They are not broken. A LiteLLM-fronted proxy picks up the declared web-search tool, runs the search itself, and folds the result into the reply — web search does work through the proxy, and that is the behavior we want. The workaround actively prevented it: the tool *declaration* is exactly what the proxy keys on, so denying the built-ins removes the trigger. The replacement skills then sent the model somewhere it no longer needs to go.

## What changed

Everything #1087 added, removed:

- `deny-web-tool.sh` and the pinned cc-websearch bundle download with its checksum gate
- the seeded `websearch` / `webfetch` skills
- `PLATFORM_CUSTOM_BACKEND` — the hook was its only consumer
- the `lite.duckduckgo.com` egress allowlist entry
- `managed-settings.json` — the file existed only to carry the hooks; the Dockerfile writes back the empty `{}` stub the image had before

Net: +2 / −134. The Dockerfile stub region and `model-gateway.sh` are byte-identical to their pre-#1087 state.

## Verification

Built the image from this branch and ran the harness inside it against a mock Anthropic endpoint that logs the `tools[]` array it receives:

- 25 tools on the wire, `WebSearch` and `WebFetch` both present — the proxy has its trigger back
- `/etc/claude-code/managed-settings.json` is `{}`
- no `/opt/cc-websearch`, no seeded web skills

`mise run check` and `mise run test` (all suites) pass; helm render is valid.

## Note for reviewers

Two things this deliberately does not address:

- **Proxy-side search is invisible to the platform.** It produces no tool call in the chat, asks for no approval, runs from the proxy's network rather than the pod's, and disables streaming for the affected turn. That is proxy behavior, not something the image can fix — see #3092.
- **Existing sandboxes keep their copy of the two seeded skills.** The workspace lives on a PVC and only seeds at first boot, so after this upgrade the leftovers reclassify from system to user skills and point at a path the new image no longer has. Deleting files from user PVCs wants its own decision.


packages/agents/claude-code/managed-settings.json
```diff
@@ -1,24 +1,5 @@
...
-        "hooks": [
-          {
-            "type": "command",
-            "command": "/opt/cc-websearch/deny-web-tool.sh WebFetch webfetch"
-          }
-        ]
-      }
-    ]
+  "permissions": {
+    "deny": ["WebSearch", "WebFetch"]
```
dam-code-guardian: 🟡 **Warning:** The deny rule is unconditional, so the built-ins are now off on genuine Anthropic backends too — where they worked. The PR body frames this as intended, and given the threat model (a declared tool is rewritable by any proxy in front of the pod) that's defensible.

What's worth stating explicitly somewhere durable is that agents on this platform have **no** web-search capability at all now, on any backend: the replacement skills are gone and nothing took their place. The Dockerfile comment explains why the deny rule beats a hook, but not that the capability is intentionally absent rather than pending — a future reader hitting a denied `WebSearch` has no pointer.

packages/agents/claude-code/workspace/.agents/skills/websearch/SKILL.md
```diff
@@ -1,33 +0,0 @@
----
```
dam-code-guardian, line 1: 🟢 **Suggestion:** The removal is complete for the *image*, but existing sandboxes keep these two seeded `SKILL.md` files on their PVC, as the PR body notes. Worth tracing what the leftovers actually do rather than leaving it at "invoking one fails":

With the pristine workspace copy no longer containing them, `judgeOrigin` returns `user` (`packages/agent-runtime/src/modules/skills/domain/skill-origin.ts` — `pristineHash === null → "user"`), which reclassifies them out of the publish gate in `publish-service.ts`. So a user on an upgraded sandbox can now *publish* a leftover copy of a skill that was image-shipped, whose body points at `/opt/cc-websearch/*.cjs` — a path the new image doesn't have.

Harmless in effect, but it's a second consequence beyond "invoking one fails" and belongs in whatever issue tracks the PVC cleanup decision.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `a0ba5cc`

## PR #3093: fix(claude-code): deny the built-in web tools by declaration, drop cc-websearch
**Author:** Tomas2D | **Branch:** fix/remove-cc-websearch → main | **Changes:** +9 −130 (7 files)

### Summary
Replaces the two `PreToolUse` hooks that denied `WebSearch`/`WebFetch` on custom backends with a single `permissions.deny` rule in the claude-code image's `managed-settings.json`, on the grounds that a hook gates execution while leaving the tool *declared* — and a LiteLLM-fronted proxy matches on the declaration, rewriting it into an in-proxy search that never surfaces in the UI. The DuckDuckGo-backed replacement skills, their pinned+checksummed bundle download, `deny-web-tool.sh`, the `PLATFORM_CUSTOM_BACKEND` marker, and the `lite.duckduckgo.com` egress entry all go with it.

### Findings
- 🟡 **Warning:** The deny rule is unconditional, so the built-ins are now off on genuine Anthropic backends too — where they worked. The PR body frames this as intended ("the built-ins are now off on every backend, which is the behavior we want"), and given the threat model (a declared tool is rewritable by any proxy in front of the pod) that's defensible. What's worth stating explicitly somewhere durable is that agents on this platform have **no** web-search capability at all now, on any backend: the replacement skills are gone and nothing took their place. `packages/agents/claude-code/Dockerfile:41` explains why the deny rule beats a hook, but not that the capability is intentionally absent rather than pending. A future reader hitting a denied `WebSearch` has no pointer. (`packages/agents/claude-code/managed-settings.json:3`)
- 🟢 **Suggestion:** The removal is complete for the *image*, but existing sandboxes keep the two seeded `SKILL.md` files on their PVC, as the PR body notes. Worth tracing what those leftovers actually do rather than leaving it at "invoking one fails": with the pristine workspace copy no longer containing them, `judgeOrigin` returns `user` (`packages/agent-runtime/src/modules/skills/domain/skill-origin.ts:7` — `pristineHash === null → "user"`), which reclassifies them out of the publish gate at `packages/api-server/src/modules/skills/services/publish-service.ts:84`. So a user on an upgraded sandbox can now *publish* a leftover copy of a skill that was image-shipped, and its body points at `/opt/cc-websearch/*.cjs`, a path the new image doesn't have. Harmless in effect, but it's a second consequence beyond "invoking one fails" and belongs in whatever issue tracks the PVC cleanup decision. (`packages/agents/claude-code/workspace/.agents/skills/websearch/SKILL.md:1`)
- ✅ **Looks good:** The claim at the center of the change is measured, not assumed — tool count on the wire with and without the rule, read from `/etc/claude-code/managed-settings.json` in the built image, with the macOS-vs-Linux caveat stated. That's the right evidence for "a deny rule does something a hook cannot."
- ✅ **Looks good:** Removal is clean with no danglers: `PLATFORM_CUSTOM_BACKEND` had exactly one consumer and both sides go together (`packages/agents/claude-code/model-gateway.sh:21` keeps the `_gateway_custom_upstream` branch it still needs for the gateway base-URL swap), the egress allowlist entry is dropped alongside the bundle that used it, and a repo-wide search for `cc-websearch`, `deny-web-tool`, `duckduckgo`, and `PLATFORM_CUSTOM_BACKEND` returns nothing outside the removed files.
- ✅ **Looks good:** Net −121 lines removes a supply-chain surface — a third-party bundle pinned to a commit SHA, checksum-gated, fetched at build time from `raw.githubusercontent.com` — rather than reshaping it. Dropping a dependency is a better outcome than hardening one.

**Verdict:** aligned

No drift to report. The diff touches only the claude-code agent image and one egress-allowlist host; no page under `docs/architecture/` was edited, and none references the built-in web tools, the cc-websearch bundle, DuckDuckGo, `managed-settings.json`, or `PLATFORM_CUSTOM_BACKEND` — that apparatus is exactly the file-path/package-level detail the content policy keeps off the pages, so removing it leaves nothing stale. `skills.md` describes image-seeded skills as a category (pristine roots, provenance) without enumerating members, so dropping two seeded skills doesn't contradict it. No new subsystem, no page moved or deleted, no `Last verified:` bump owed, no ADR status change in the diff.

### Verdict
COMMENT — The mechanism is right and well-evidenced; the two notes are about consequences to record (no web capability anywhere now; leftover PVC skills reclassify as publishable), not defects in the change.

---

PR #3094, opened 2026-07-30
fix(controller): normalize fsGroup-inherited setgid; prune lost+found in deep finds

jezekra1:
## The failure

`/tmp/src.meta /tmp/dst.meta differ: char 3, line 1` on every real CSI block target: source dir `700`, target `2700`. **fsGroup marks the target volume root setgid** (Kubernetes' mechanism for making the volume's group reach new entries), and the **kernel** then propagates the bit into every directory the copy creates at `mkdir` time — beneath tar, which re-chmods only the directories it happens to revisit (`--delay-directory-restore` does not change that; measured). The strict metadata verification then correctly refused to certify the copy. The fail-closed gate worked; the copy had genuine (benign) drift. Invisible on dev: hostPath classes skip fsGroup.

## The fix: make the copy exact, keep the proof strict

Rejected two weaker designs along the way (visible in the commit history): chmod-normalize-and-reapply (extra machinery), and masking the bit out of the comparison (weakens the proof this whole feature is built on — reviewer pushback was right).

What ships: after extraction, **every directory is re-stamped with its exact recorded mode** from the source walk we already hold — one cheap pass over dirs only, no extra source I/O — and **verification stays byte-strict for every entry**. Two empirical findings are load-bearing and pinned by tests:

- GNU chmod **preserves setuid/setgid on directories for numeric modes under five digits**: `chmod 700` keeps an inherited bit, `chmod 00700` clears it, `002755` still sets a genuinely recorded one. The `00` prefix in the restore is not decoration.
- tar cannot be made to do this itself (measured, incl. `--delay-directory-restore`).

Also here:

- **Deep `find`s now `-prune` `lost+found`** instead of path-excluding it: exclusion still *descends*, and a `0700` root-owned one (ext4 targets) kills the walk with permission-denied under `set -e`. Latent in the merged build — your XFS targets are why it never fired. Root-anchored, so a nested `lost+found` the workspace owns is still inventoried.
- **Representability preflight**: a path containing `|` or newline can't be expressed in the line-based inventory and previously surfaced as a baffling foreign-owner block; now refused by name with a rename remedy.

## Proven by executing the real script (extracted from `buildMigrationJob`, byte-for-byte) as uid 65532

Fixture: target root owned by another uid, mode `2770` (fsGroup-shaped), root-owned `0700` `lost+found` inside.

- fresh + dirty-retry passes verify with **exact modes** `755/700/2755/555` — inherited bits scrubbed, a genuinely-recorded source setgid dir survives ✔
- hardlink count preserved; file modes strict throughout ✔
- negatives block with actionable paths: foreign-uid ✔ owner-unreadable ✔ pipe-in-name ✔

Cluster rehearsal (clean install → agent on root-squashing kernel-NFS class → drain to RWO local-path) running now; result will be posted here.

## After deploy

Nothing to clean up: failed Jobs sit in the bounded retry with sources untouched; the first retry after the controller rolls uses the fixed script.

### DIFF
--- a/packages/controller/pkg/reconciler/storage_migration.go
+++ b/packages/controller/pkg/reconciler/storage_migration.go
@@ -894,7 +894,11 @@ copy_verify() {
   # The readability flag rides the same pass: find evaluates -readable per
   # entry and the alternation picks the prefix. The inner group is what keeps
   # the -o from swallowing the -mindepth/-path filters.
-  (cd "$src" && find . -mindepth 1 ! -path './lost+found*' \
+  # -prune on the root lost+found, never ! -path: exclusion by path still
+  # DESCENDS into the directory, and a 0700 root-owned one aborts the walk
+  # with permission denied. Anchored (./lost+found, no glob) so a nested
+  # one the workspace legitimately owns is still inventoried.
+  (cd "$src" && find . -mindepth 1 \( -path ./lost+found -prune \) -o \
      \( \( -readable -printf "r|%y|%m|%U|%p|%l\n" \) -o -printf "x|%y|%m|%U|%p|%l\n" \) \
    ) > /tmp/src.walk
   entries=$(wc -l < /tmp/src.walk)
@@ -976,6 +980,26 @@ copy_verify() {
   # so comparing it against the source root would fail spuriously).
   touch "$dst/.migration-writable" && rm -f "$dst/.migration-writable"
 
+  # fsGroup marks the target ROOT setgid (that is how the volume group
+  # carries into new entries), and every directory the copy creates
+  # INHERITS that bit at mkdir time — underneath tar, which never
+  # re-chmods because the mode it passed to mkdir was already the recorded
+  # one. Left alone, every target directory reads 2xxx where the source
+  # says xxx and the metadata pass (correctly) refuses the copy. Strip the
+  # inherited bit from directories, then re-apply it to the rare ones
+  # whose SOURCE mode genuinely records it, so the comparison measures
+  # fidelity rather than a mount artifact. The root itself keeps the
+  # kubelet's bit: it is excluded from the comparison, and future
+  # top-level writes by the agent should keep inheriting the group.
+  # (Invisible on hostPath-backed dev classes, which skip fsGroup — this
+  # only manifests on real CSI block targets.)
+  # -prune, not ! -path: find must not even DESCEND into the root-owned
+  # lost+found (permission denied would abort the script). Anchored to the
+  # root so a workspace's own nested lost+found is still processed.
+  find "$dst" -mindepth 1 \( -path "$dst/lost+found" -prune \) -o \( -type d -exec chmod g-s {} + \)
+  awk -F'|' '$2=="d" && length($3)==4 && substr($3,1,1)~/[2367]/ {print $5}' /tmp/src.walk \
+    | while IFS= read -r p; do chmod g+s "$dst/$p"; done
+
   # Two-pass verification against the quiesced source; the old volume is
   # deleted at flip on the strength of these comparisons, so both fail
   # closed on any difference.
@@ -986,7 +1010,7 @@ copy_verify() {
   # uid:0 while an fsGroup-managed target assigns its own gid — group
   # identity is not part of the workspace contract), the volume root itself,
   # and lost+found.
-  (cd "$dst" && find . -mindepth 1 ! -path './lost+found*' -printf "d|%y|%m|%U|%p|%l\n") \
+  (cd "$dst" && find . -mindepth 1 \( -path ./lost+found -prune \) -o -printf "d|%y|%m|%U|%p|%l\n") \
     | cut -d'|' -f${META_CUT} | LC_ALL=C sort > /tmp/dst.meta
   cmp /tmp/src.meta /tmp/dst.meta
   phase "verified metadata"
@@ -1004,7 +1028,7 @@ copy_verify() {
   # cannot tear.
   sums() {
     rm -f /tmp/sums.*
-    (cd "$1" && find . -type f ! -path './lost+found*' -print0 \
+    (cd "$1" && find . \( -path ./lost+found -prune \) -o -type f -print0 \
        | xargs -0 -r -P"${PAR}" -n64 sh -c 'md5sum "$@" >> "/tmp/sums.$$"' _)
     cat /tmp/sums.* 2>/dev/null | LC_ALL=C sort
   }
--- a/packages/controller/pkg/reconciler/storage_migration.go
+++ b/packages/controller/pkg/reconciler/storage_migration.go
@@ -86,6 +86,9 @@ const (
 	// migrationFallbackGID pairs with migrationFallbackUID when the chart
 	// pins no group; the platform images run 65532:65532.
 	migrationFallbackGID = int64(65532)
+	// migrationServiceAccount is the Job's dedicated identity — the target
+	// of the ops-managed SCC grant on OpenShift. Ensured by the manager.
+	migrationServiceAccount = "platform-migration"
 	// migrationChecksumParallelism is how many md5sum workers read the tree
 	// at once. Small-file verification over a network filesystem is bound by
 	// per-file round-trips, not bandwidth or CPU, so concurrency buys close
@@ -234,6 +237,35 @@ func (m *StorageMigrationManager) ReleaseGated(ctx context.Context) {
 	}
 }
 
+// ensureServiceAccount creates the copy Job's dedicated identity if it is
+// missing. Idempotent, called only while there is work to admit. The SA
+// carries no role bindings and its token is never mounted — it exists so an
+// OpenShift install can scope the runs-as-root SCC grant to exactly this
+// workload (an ops-side, out-of-band binding, like every cluster-scoped
+// security object).
+func (m *StorageMigrationManager) ensureServiceAccount(ctx context.Context) {
+	_, err := m.client.CoreV1().ServiceAccounts(m.config.Namespace).Get(ctx, migrationServiceAccount, metav1.GetOptions{})
+	if err == nil || !errors.IsNotFound(err) {
+		if err != nil {
+			slog.Warn("storage migration: checking service account failed", "error", err)
+		}
+		return
+	}
+	sa := &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      migrationServiceAccount,
+			Namespace: m.config.Namespace,
+			Labels:    map[string]string{"agent-platform.ai/managed-by": "platform-controller"},
+		},
+		AutomountServiceAccountToken: ptrBool(false),
+	}
+	if _, err := m.client.CoreV1().ServiceAccounts(m.config.Namespace).Create(ctx, sa, metav1.CreateOptions{}); err != nil && !errors.IsAlreadyExists(err) {
+		slog.Warn("storage migration: creating service account failed", "error", err)
+		return
+	}
+	slog.Info("storage migration: service account ensured", "name", migrationServiceAccount)
+}
+
 // Reconcile advances every in-flight migration one step and admits new
 // agents up to the concurrency cap. Exported for tests; RunLoop drives it.
 func (m *StorageMigrationManager) Reconcile(ctx context.Context) {
@@ -308,6 +340,10 @@ func (m *StorageMigrationManager) Reconcile(ctx context.Context) {
 		names = append(names, name)
 	}
 	sort.Strings(names)
+
+	if len(names) > 0 {
+		m.ensureServiceAccount(ctx)
+	}
 
 	slots := m.concurrency() - len(inFlight)
 	for _, name := range names {
@@ -832,9 +868,10 @@ func jobFailed(job *batchv1.Job) bool {
 // (source read-only, target read-write) pair, running a copy + two-pass
 // checksum verification per pair.
 //
-// The copy runs as the AGENT's uid, unprivileged and needing no
-// capabilities: an agent's workspace is single-owner by construction
-// (everything in it is written as the agent user), so the agent uid can
+// The copy runs with SPLIT IDENTITY. Everything that touches the SOURCE
+// runs as the AGENT's uid (dropped into via setpriv): the source may live
+// on a root-squashing share, where uid 0 is remapped to nobody server-side
+// and is the weakest identity on the mount, while the agent uid can
 // read every file — including 0600 files and
 // 0700 dirs — even on shared-filesystem classes that squash root, where a
 // root-running copy would EACCES on the first private file. Ownership on
@@ -853,9 +890,9 @@ func buildMigrationJob(agentName string, pairs []migrationPair, cfg *config.Conf
 	var volumes []corev1.Volume
 	var mounts []corev1.VolumeMount
 	var script strings.Builder
-	// The copy runs as the agent's own uid AND gid, resolved from the chart's
-	// container security context — the identity the agent itself runs with,
-	// so what the copy creates is what the agent can use.
+	// The reader identity: the agent's own uid and gid from the chart's
+	// container security context — the one identity a root-squashing source
+	// share always honors for the workspace it owns.
 	uid, gid := migrationFallbackUID, migrationFallbackGID
 	if sc := cfg.AgentBase.ContainerSecurityContext; sc != nil {
 		if sc.RunAsUser != nil {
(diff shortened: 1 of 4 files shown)


packages/controller/pkg/reconciler/storage_migration.go
```diff
@@ -976,6 +992,24 @@ copy_verify() {
...
+  # does not change that; measured). Rather than masking the bit out of
+  # the verification, the copy is made faithful: one chmod per directory,
+  # and the comparison below stays byte-strict for every entry.
+  # The 00 prefix is load-bearing: GNU chmod PRESERVES setuid/setgid on
+  # directories for numeric modes under five digits — "chmod 700" keeps an
+  # inherited bit, "chmod 00700" clears it (and "002755" still sets a
+  # genuinely recorded one).
+  awk -F'|' '$2=="d" {print $3 "|" $5}' /tmp/src.walk > /tmp/src.dirs
+  while IFS='|' read -r mode path; do
+    chmod "00$mode" "$dst/$path"
```
dam-code-guardian: 🟡 **Warning:** The `00` prefix is width-dependent, so this only clears the inherited bit for 3-digit modes. `find -printf %m` emits **no leading zeros** — a `0040` dir prints `40`, a `000` dir prints `0` — so the argument becomes `0040`/`000`, under five digits, which is exactly the case where GNU chmod *preserves* setgid. Verified by execution on a `2770`-inherited dir: `chmod 0040` → `2040` and `chmod 000` → `2000` (bit retained), while `chmod 00700` → `700` (correctly cleared). Such a directory is then re-stamped with a setgid it never had and the byte-strict `cmp` blocks the migration.

The owner-readable gate above catches `000`/`040` today, so this is hard to reach — but that gate guards a different invariant and documents a symlink exemption, so leaning on it couples two unrelated checks silently. Padding to a fixed width is width-proof for every possible `%m` output, and I verified it still clears an inherited bit for `0`/`40`/`700` while setting a genuinely recorded `2755`.

Separately: `chmod` exits 1 on a missing or permission-denied path, but a non-zero exit of the last command inside a `while read` body does not trip `set -e`, so a failed re-stamp surfaces only as a later `cmp` mismatch with no hint that the *restore* was at fault. Failing closed here would match every other gate in this function.

```suggestion
    chmod "0000$mode" "$dst/$path" || { echo "migration blocked: could not restore mode $mode on $path"; exit 1; }
```

packages/controller/pkg/reconciler/storage_migration.go
```diff
@@ -976,6 +992,24 @@ copy_verify() {
...
+  # every directory the copy creates, underneath tar — which re-chmods
+  # only the directories it happens to revisit (--delay-directory-restore
+  # does not change that; measured). Rather than masking the bit out of
+  # the verification, the copy is made faithful: one chmod per directory,
+  # and the comparison below stays byte-strict for every entry.
+  # The 00 prefix is load-bearing: GNU chmod PRESERVES setuid/setgid on
+  # directories for numeric modes under five digits — "chmod 700" keeps an
+  # inherited bit, "chmod 00700" clears it (and "002755" still sets a
+  # genuinely recorded one).
+  awk -F'|' '$2=="d" {print $3 "|" $5}' /tmp/src.walk > /tmp/src.dirs
```
dam-code-guardian: 🟢 **Suggestion:** One `chmod` process per directory cuts against this function's own stated premise — "every stat is a network round-trip AND an IOP against a capped share". A pnpm store or `node_modules` is thousands of directories, so this is thousands of sequential fork+exec round-trips against the capped share, inside the 4-hour `ActiveDeadlineSeconds` bound.

Distinct modes are few in practice (`755`/`700`/`555`), so grouping by mode collapses it to a handful of batched calls. I verified a grouped `sort -u` + `xargs -0` form yields byte-identical modes, including a recorded `2755`. Worth it only if real workspaces are dir-dense — the current loop's correctness is not in question.

packages/controller/pkg/reconciler/storage_migration.go
```diff
@@ -986,7 +1020,7 @@ copy_verify() {
   # uid:0 while an fsGroup-managed target assigns its own gid — group
   # identity is not part of the workspace contract), the volume root itself,
   # and lost+found.
-  (cd "$dst" && find . -mindepth 1 ! -path './lost+found*' -printf "d|%y|%m|%U|%p|%l\n") \
+  (cd "$dst" && find . -mindepth 1 \( -path ./lost+found -prune \) -o -printf "d|%y|%m|%U|%p|%l\n") \
```
dam-code-guardian: 🟡 **Warning:** This deep walk is now **anchored** (`-path ./lost+found`, exact match), but the tar file list at line 981 still uses the glob `! -path './lost+found*'`. The two predicates disagree on any top-level entry whose name merely *starts with* `lost+found` — e.g. a legitimate workspace dir `lost+foundry`.

Verified by executing the post-PR sequence: the anchored walk inventories `./lost+foundry` and its contents into `src.meta`, the glob tar list excludes it from the copy, and `cmp` then fails closed on entries the target never received — a phantom difference, the exact failure class this anchoring was meant to remove. Pre-PR both were the same glob, so they agreed and this could not happen. The wipe (`! -name lost+found`) already agrees with the new anchoring; the tar list is the lone straggler.

Fix is one line at 981 — `-maxdepth 1` means there is nothing to descend into, so `-prune` is harmless there:

```
(cd "$src" && find . -mindepth 1 -maxdepth 1 \( -path ./lost+found -prune \) -o -print0) \
```

Note `storage_migration_test.go:246` pins the glob, so it needs updating alongside.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `27e0904`

### Summary
Fixes the storage-migration copy Job failing metadata verification when `fsGroup` marks the target volume root setgid and the kernel propagates the bit into every directory the copy creates. Rather than masking the bit out of the comparison, the copy is made faithful: directories are re-stamped with their exact recorded modes from the existing source walk, and verification stays byte-strict. Also switches deep `find`s from path-exclusion to `-prune` for `lost+found`, and adds a preflight refusing paths containing `|` or newline.

The design judgement is right, and I verified the two load-bearing empirical claims by execution — `chmod 700` on a `2700` dir yields `2700` while `chmod 00700` yields `700`, and `002755` still sets a recorded setgid. The `-prune` diagnosis is also correct: `! -path` descends and dies on a `0700` root-owned `lost+found`.

### Findings

- 🟡 **Warning:** The `lost+found` anchoring is now inconsistent between the *inventory* and the *copy*, which can block a migration that previously succeeded. The three deep finds became anchored (`-path ./lost+found -prune`, exact match), but the tar file list at `storage_migration.go:981` still uses the glob `! -path './lost+found*'`. Those two predicates disagree on any top-level entry whose name merely *starts with* `lost+found` — e.g. a legitimate workspace directory `lost+foundry`. Verified by executing the post-PR sequence: the anchored source walk inventories `./lost+foundry` and its contents, the glob tar list excludes it from the copy, so `src.meta` carries three entries the target never receives and `cmp` fails closed — the migration blocks with a phantom difference, exactly the failure class the anchoring was meant to eliminate. Pre-PR both predicates were the same glob, so they agreed and this could not happen. The wipe at `:963` uses `! -name lost+found` (exact), so it already agrees with the new anchoring — the tar list is the lone straggler. Fix is one line, and `-maxdepth 1` means there is nothing to descend into, so `-prune` is harmless here. Note the test at `storage_migration_test.go:246` pins the glob, so it needs updating with it. (`packages/controller/pkg/reconciler/storage_migration.go:981`)

- 🟡 **Warning:** `chmod "00$mode"` is only setgid-clearing for 3-digit modes — the `00` prefix assumes `find -printf %m` always emits three digits, but `%m` emits *no leading zeros*. I verified: a `0040` dir prints `40` and a `000` dir prints `0`, so the argument becomes `0040`/`000` — under five digits, which is precisely the case where GNU chmod *preserves* the inherited setgid the comment says it clears. Confirmed by execution: on a `2770`-inherited dir, `chmod 0040` → `2040` and `chmod 000` → `2000`, both retaining the bit, while `chmod 00700` → `700` correctly clears it. So a source directory with mode `040`/`000` is re-stamped with a setgid it never had, and the byte-strict `cmp` then blocks the migration. The owner-readable gate at `:926` catches `000` and `040` (not `-readable`, so the walk flags them `x` and the script exits earlier), which makes this hard to reach today — but that gate is a *different* invariant guarding a different failure, and it is documented as exempting symlinks, so relying on it couples two unrelated checks silently. Padding to a fixed width removes the coupling: `chmod "0000$mode"` is ≥5 digits for every possible `%m` output, and I verified it both clears an inherited bit for `0`/`40`/`700` and still sets a recorded `2755`. (`packages/controller/pkg/reconciler/storage_migration.go:1009`)

- 🟢 **Suggestion:** The dir-mode restore spends one `chmod` process per directory, which cuts against this function's own stated performance premise ("every stat is a network round-trip AND an IOP against a capped share"). On a dir-heavy workspace — a pnpm store or `node_modules` is thousands of directories — that is thousands of sequential fork+exec+syscall round-trips against the capped share, inside a Job bounded by the 4-hour `ActiveDeadlineSeconds` at `:1078`. Distinct modes are few (typically `755`/`700`/`555`), so grouping by mode collapses it to a handful of batched `chmod` calls; I verified a grouped `sort -u` + `xargs -0` form produces byte-identical modes including the recorded `2755`. Worth it only if real workspaces are dir-dense — the correctness of the current loop is not in question. (`packages/controller/pkg/reconciler/storage_migration.go:1007`)

- 🟢 **Suggestion:** The chmod loop silently swallows failures. `chmod` exits 1 on a missing or permission-denied path, but a non-zero exit of the last command *inside* a `while read` body does not trip `set -e` (verified), so a failed re-stamp only surfaces later as a `cmp` mismatch — with no indication that the restore, rather than the copy, was at fault. The `phase "restored dir modes: N"` line reports the count *attempted*, not applied, so it reads as success either way. Both warnings above manifest this way, which is why they cost a full walk to diagnose rather than being reported by name. Adding `|| { echo "migration blocked: could not restore mode $mode on $path"; exit 1; }` to the chmod would make the restore fail closed like every other step in this function. (`packages/controller/pkg/reconciler/storage_migration.go:1009`)

- ✅ **Looks good:** The core design call — making the copy faithful instead of masking the bit out of the comparison — is the right one, and the test pins it with `assert.NotContains(runnable, "dirmask")` so the weaker design cannot quietly return. Reusing the existing walk means no extra source I/O.
- ✅ **Looks good:** The `-prune` fix is correctly root-anchored, so a nested `lost+found` the workspace legitimately owns is still inventoried; the comment explains *why* exclusion-by-path is insufficient rather than just what changed.
- ✅ **Looks good:** The representability preflight caps its output inside `awk` rather than piping to `head`, consistent with the existing `unreadable`/`foreign` gates and their documented SIGPIPE rationale.

✅ No findings.

`docs/architecture/persistence.md` describes the storage migration at the correct altitude — "copies the quiesced volume onto a fresh RWO PVC in a Job, verifies the copy by checksum" — and this PR changes *how* the copy achieves fidelity, not the subsystem's behavior or responsibility. Directory modes, `tar` flags, and `lost+found` handling are exactly the volatile, below-framework detail the documentation guidelines direct pages to omit; the page mentions none of them today and adding them would itself be drift. No architecture page was edited, so no `Last verified:` bump is due, and no ADR changed state.

### Verdict
COMMENT — The design judgement is sound and the empirical claims hold up under execution, but the new `lost+found` anchoring disagrees with the unchanged tar file list, which can fail-close a migration on a `lost+found`-prefixed directory; the `00$mode` prefix is also width-dependent in a way the comment doesn't account for.

---

jezekra1:
## Clean-cluster rehearsal — split-identity copy, verified end to end

Fresh `cluster:uninstall` → install from this branch (agents on a **real kernel-NFS class with `root_squash`**, migration disabled) → agent created, permission zoo staged (15 entries: `0600` secret, `0700` dir, `0555` read-only dir, **`4755` suid tool**, **`2755` setgid dir**, hardlink pair, live + dangling symlinks, 3MB blob) → migration enabled.

**Result: gate-to-flip in 30 seconds, one attempt, zero failures in the controller log.** Post-migration volume on RWO `local-path` is **line-for-line identical** to the recorded baseline — full metadata (`type|mode|uid|path|linktarget`) and checksums, including the suid bit and the genuine setgid dir landing exactly. Agent restored automatically (`wasRunning=true`).

Also proven on the way here, against the live squashing export (fixture runs the byte-for-byte script from `buildMigrationJob` as root):

- root over the NFS mount **provably EACCES** on the agent's `0600` files — the reason the reader drops to the agent uid via `setpriv`
- a **readable root-owned stray copies faithfully as uid 0 and verifies** (ownership compared, not excused); an unreadable one blocks with its path
- dirty-retry deterministic; hardlinks preserved; target root shaped `2770 65532:65532` for the agent pods' new `fsGroup`/`OnRootMismatch` (first mount skips the kubelet's recursive pass — the verified interior is never rewritten)

Deploy prerequisites, both loud if missed: ops binds a uid-0 SCC to the `platform-migration` SA (`oc adm policy add-scc-to-user anyuid -z platform-migration -n <agents-ns>`), and the agents' SCC must permit `fsGroup: 65532` (check: `oc get pod <agent> -o jsonpath='{.metadata.annotations.openshift\.io/scc}'` then `oc get scc <name> -o jsonpath='{.fsGroup}'` — `RunAsAny` = done).

xjacka (approved):

packages/controller/pkg/reconciler/storage_migration.go
```diff
@@ -865,152 +902,145 @@ func buildMigrationJob(agentName string, pairs []migrationPair, cfg *config.Conf
...
+  # Source checksums before the copy: the source is quiesced, so before
+  # and after are indistinguishable, and this keeps every source read
+  # under the squash-safe identity.
+  $AS_AGENT bash "$W"/srcside.sh "$src" sums "$PAR" "$W" | LC_ALL=C sort > "$W"/src.sum
+  phase "hashed source"
+
+  # Wipe as root: a prior attempt's read-only directories are plain rm.
+  # lost+found stays — a fresh ext4 volume's own artifact, excluded from
+  # verification on both sides.
   find "$dst" -mindepth 1 -maxdepth 1 ! -name lost+found -exec rm -rf {} +
```
dam-code-guardian, line 1001: 🔴 **Critical:** every **retry** fails permanently from here on, and `BackoffLimit: 2` plus the 10-minute recreate loop turns the first transient failure into an agent gated forever.

`chmod 2770 "$dst"` below (`:1026`) persists on the filesystem, so the *next* attempt extracts into a setgid root and the kernel propagates the bit into directories the extraction creates. Verified by execution — four consecutive attempts against one target:

```
attempt 1 (root=755):   VERIFY PASS      # leaves root 2770
attempt 2 (root=2770):  VERIFY FAIL      # .ssh 700 -> 2700
attempt 3 (root=2770):  VERIFY FAIL
attempt 4 (root=2770):  VERIFY FAIL
```

It never self-heals — each attempt re-applies `2770`. Mechanism: GNU tar creates dirs `0700` then repairs modes at the end, so a source dir whose mode *is* `700` gets no repair `chmod` and keeps the inherited bit (`--delay-directory-restore` does not help — measured). `701/710/711/740/750/755/770/777/500/555/2755` all round-trip fine; `700` is the one that breaks — and `700` is `.ssh`, `.gnupg`, `~/.config`, every `mkdir` under a default umask in a private tree.

This is a regression *this* PR introduces: the `2770` shaping is new, and pre-PR the root's setgid came from the kubelet's `fsGroup`, which this PR removes from the Job.

Fix is one line here. It must be the 5-digit form — I verified `chmod 755` and `chmod 0755` both leave `2755` (the same GNU chmod width rule that bit the previous revision), while `chmod 00755` clears it. With this added, all four attempts pass.

```suggestion
  find "$dst" -mindepth 1 -maxdepth 1 ! -name lost+found -exec rm -rf {} +
  # Strip any setgid a PRIOR attempt's root shaping left behind: the kernel
  # propagates it into every directory the extraction creates, and tar does
  # not repair a dir whose recorded mode is already 0700. Five digits is
  # required -- chmod 755 and 0755 both PRESERVE setgid on a directory.
  chmod 00755 "$dst"
```

packages/controller/pkg/reconciler/storage_migration.go
```diff
@@ -865,152 +902,145 @@ func buildMigrationJob(agentName string, pairs []migrationPair, cfg *config.Conf
...
-    (cd "$src" && find . -mindepth 1 -maxdepth 1 ! -path './lost+found*' -print0) \
-      | tar -C "$src" --sparse --null -T - -cf - \
-      | tar -C "$dst" -xpf - --no-overwrite-dir
+    # Reader as the agent (squash-safe), writer as root: exact modes
+    # including setuid, exact OWNERSHIP (--same-owner --numeric-owner —
+    # readable foreign-uid strays land as themselves), hard links (a
+    # hard-link-heavy pnpm store must not inflate past the volume size),
+    # and sparseness. The "." member restores the volume root's own mode,
+    # owner, and times — a root writer may apply all of it.
+    $AS_AGENT tar -C "$src" --sparse -cf - . | tar -C "$dst" -xpf - --same-owner --numeric-owner
```
dam-code-guardian, line 1011: 🔴 **Critical:** archiving `.` with no `lost+found` exclusion breaks on any **ext4 source** whose `lost+found` is the usual root-owned `0700`.

The reader is `$AS_AGENT`, so that directory is unreadable to it. Verified as uid 65532:

```
tar: ./lost+found: Cannot open: Permission denied
tar: Exiting with failure status due to previous errors
exit=2          # the pre-PR -maxdepth 1 member list exits 0
```

The pre-PR form fed tar an explicit top-level list excluding `./lost+found*`; `-cf - .` walks everything. And it is *silent* in the inventory: the source walk `-prune`s `lost+found`, so the `unreadable` gate never sees it — the operator gets a bare tar error instead of the named path that gate exists to produce.

The PR's reasoning covers the *target's* `lost+found`, and the rehearsal ran on `local-path`, so a source with one was never exercised. Every sibling walk agrees on excluding it (4× `-path ./lost+found -prune`, per the new test) — the tar member list is the lone straggler.

`--exclude` keeps the `./` member you need for the root's mode/owner/times while skipping an unreadable `lost+found` — verified, exit 0, members `./ ./.ssh/ ./.ssh/id ./work/ ./work/f`.

```suggestion
    $AS_AGENT tar -C "$src" --sparse --exclude=./lost+found -cf - . | tar -C "$dst" -xpf - --same-owner --numeric-owner
```

packages/controller/pkg/reconciler/storage_migration.go
```diff
@@ -234,6 +237,35 @@ func (m *StorageMigrationManager) ReleaseGated(ctx context.Context) {
 	}
 }
 
+// ensureServiceAccount creates the copy Job's dedicated identity if it is
+// missing. Idempotent, called only while there is work to admit. The SA
+// carries no role bindings and its token is never mounted — it exists so an
+// OpenShift install can scope the runs-as-root SCC grant to exactly this
+// workload (an ops-side, out-of-band binding, like every cluster-scoped
+// security object).
+func (m *StorageMigrationManager) ensureServiceAccount(ctx context.Context) {
```
dam-code-guardian, line 246: 🟡 **Warning:** this swallows a failure and lets the run proceed to create a Job that cannot schedule.

RBAC is fine — the controller does have `serviceaccounts: create` (`deploy/helm/platform/templates/controller/rbac.yaml:81`). The issue is the error handling: both paths `slog.Warn` and `return`, then `Reconcile` continues into `migrateAgent`, which creates the Job referencing `platform-migration` regardless. If the SA really is absent, the Job is admitted but its pods never schedule (`ServiceAccount not found`) — signal is one `Warn` line plus an agent gated until `ActiveDeadlineSeconds` (4h) fires.

The PR description says a missing SCC binding "rejects the pod at admission, loudly" — true of the SCC, but a missing *SA* fails quietly inside the Job controller. Returning an error here and skipping Job creation for the tick would surface it where an operator is already looking.

packages/controller/pkg/reconciler/storage_migration.go
```diff
@@ -865,152 +902,145 @@ func buildMigrationJob(agentName string, pairs []migrationPair, cfg *config.Conf
...
+# volume root carries no setgid bit and the kernel inherits nothing:
+# modes land exactly as tar records them.
+AS_AGENT="setpriv --reuid=${AGENT_UID} --regid=${AGENT_GID} --clear-groups"
+
+# Both identities share one private scratch dir (root-created, sticky,
+# world-writable): the agent-side workers and the root-side collectors
+# never collide with anything pre-existing, and Linux protected_regular
+# never refuses a redirect over a file the other identity created.
+W=$(mktemp -d)
+chmod 1777 "$W"
```
dam-code-guardian, line 928: 🟢 **Suggestion:** `1777` is broader than this container needs — there are exactly two identities in it, root and one known uid, both fixed at render time. The comment also leans on `protected_regular`, a sysctl the pod does not control (unreadable in my environment; on a node where it is `0` the stated protection is simply absent).

```suggestion
chown "root:${AGENT_GID}" "$W"
chmod 0770 "$W"
```

Gives the agent side exactly the access it needs, with no world bit and no reliance on kernel hardening.

packages/controller/pkg/reconciler/storage_migration.go
```diff
@@ -865,152 +902,145 @@ func buildMigrationJob(agentName string, pairs []migrationPair, cfg *config.Conf
...
-  # top-level link argument, and a DANGLING one (legitimate content) would
-  # error out and wedge every retry. rm -rf removes links themselves fine.
-  find "$dst" -mindepth 1 -maxdepth 1 ! -name lost+found ! -type l -exec chmod -R u+rwX {} +
+  # Verification baseline at FULL fidelity — the uid column is compared,
+  # not excused, because the root writer preserves ownership exactly.
+  cut -d'|' -f2- "$W"/src.walk | LC_ALL=C sort > "$W"/src.meta
+  # Source checksums before the copy: the source is quiesced, so before
+  # and after are indistinguishable, and this keeps every source read
+  # under the squash-safe identity.
+  $AS_AGENT bash "$W"/srcside.sh "$src" sums "$PAR" "$W" | LC_ALL=C sort > "$W"/src.sum
```
dam-code-guardian, line 995: 🟢 **Suggestion:** an **empty source** exits non-zero here before reporting anything useful. `cat "$W"/sums.src.*` with no matching files returns 1 under `set -euo pipefail` — the `if [ "$entries" -gt 0 ]` guard protects the *copy* at `:1006`, but this collection runs unconditionally first. Verified: empty source exits 1, non-empty exits 0.

Pre-existing (the old `sums()` had the same shape), so not a regression — but the refactor moved this code, and `cat "$W"/sums.src.* 2>/dev/null || true` inside `srcside.sh` would close it in passing.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `3d743c1`

### Summary
Third design iteration on the storage-migration copy Job. The PR abandons the unprivileged copy entirely: the Job now runs as **root** with a **split identity** — every source read drops to the agent uid via `setpriv` (a root-squashing source share never honors uid 0), while the target side stays root to own the fresh volume root, restore exact ownership (`--same-owner --numeric-owner`), and wipe retry residue. `fsGroup` moves off the Job and onto the *agent* pods (with `OnRootMismatch`), the copy Job shapes the target root to `2770 agent:agent` to match, `allowOwnershipRemap` is deprecated into a no-op, and the deep `find`s `-prune` `lost+found`. The root grant is scoped to a new `platform-migration` ServiceAccount the manager ensures itself.

The design is a real improvement over `27e0904`: dropping to the agent uid for reads is the correct answer to root-squash, comparing the uid column instead of excusing it is strictly stronger verification, and both of my prior warnings are structurally gone (the `chmod "00$mode"` loop was deleted, and the tar list is no longer a separate `find`). The cluster rehearsal in the PR comment is genuinely good evidence — but it only covers the **first** attempt against a **fresh** volume, which is exactly where the bug below hides.

### Findings

- 🔴 **Critical:** `chmod 2770 "$dst"` makes every **retry** fail permanently, and the copy Job has `BackoffLimit: 2` plus a 10-minute recreate loop — so the first transient failure converts into an agent gated forever. The root shaping runs *after* the copy but *persists on the filesystem*; the next attempt extracts into a root that is now setgid, and the kernel propagates the bit into directories the extraction creates. Verified by execution, four consecutive attempts against one target: attempt 1 `VERIFY PASS` (root `755` → left `2770`), attempts 2, 3 and 4 all `VERIFY FAIL` — source `.ssh` is `700`, target lands `2700`, `cmp` fails closed. It never self-heals, because each attempt re-applies `2770`. The mechanism is narrow but hits a mode agents really have: GNU tar creates directories `0700` then repairs modes at the end, so for a source dir whose mode *is* `700` there is no repair `chmod` and the inherited bit survives (`--delay-directory-restore` does not change it — measured). I checked the neighbours: `701`, `710`, `711`, `740`, `750`, `755`, `770`, `777`, `500`, `555`, `2755` all round-trip correctly — `700` is the one that breaks, and `700` is `.ssh`, `.gnupg`, `~/.config` and every `mkdir` under a default umask in a private tree. Note this is a *regression this PR introduces*: the `2770` shaping is new here, and pre-PR the target root got its setgid from the kubelet's `fsGroup` — which this PR removes from the Job. **Fix:** strip the bit from the target root immediately before extracting, in the same place the wipe happens. It must be `chmod 00755 "$dst"` — I verified `chmod 755` and `chmod 0755` both *leave* `2755` (GNU chmod preserves setgid on directories for numeric modes under five digits, the same width rule that bit the previous revision), while `chmod 00755` clears it. With that one line added, all four attempts `VERIFY PASS`. (`packages/controller/pkg/reconciler/storage_migration.go:1001`)

- 🔴 **Critical:** the copy now archives `.` with no `lost+found` exclusion, so on an **ext4 source** whose `lost+found` is the usual root-owned `0700`, `tar` fails as the agent uid and the migration cannot proceed. The archive changed from an explicit `-maxdepth 1` member list that excluded `./lost+found*` to `tar -C "$src" --sparse -cf - .`, which walks everything. The reader is `$AS_AGENT`, so a root-owned `0700` directory is unreadable to it. Verified as uid 65532: `tar: ./lost+found: Cannot open: Permission denied` / `Exiting with failure status`, exit 2, where the pre-PR member-list form exits 0. Worse, this is *silent* in the inventory — the source walk `-prune`s `lost+found`, so the `unreadable` gate that would name the path never sees it, and the operator gets a bare tar error instead of the actionable message the gate is designed to produce. The PR's own reasoning covers the *target's* `lost+found` and the rehearsal ran on `local-path`, so a source with one was never exercised. Note the sibling walks all agree on excluding it (4× `-path ./lost+found -prune`, per the new test) — the tar member list is the one place it was dropped. **Fix:** `$AS_AGENT tar -C "$src" --sparse --exclude=./lost+found -cf - .` — I verified this keeps the `./` member (which the PR needs for the root's mode/owner/times) while excluding an unreadable `lost+found` cleanly, exit 0. (`packages/controller/pkg/reconciler/storage_migration.go:1011`)

- 🟡 **Warning:** the controller's RBAC does permit creating the new ServiceAccount (`serviceaccounts: create`, `deploy/helm/platform/templates/controller/rbac.yaml:81`) — that part is fine — but `ensureServiceAccount` swallows a failure and proceeds anyway. Both error paths `slog.Warn` and `return`, then `Reconcile` continues into `migrateAgent`, which creates the Job referencing `platform-migration` regardless. If the SA is genuinely absent the Job is admitted but its pods never schedule (`ServiceAccount not found`), and the only signal is a `Warn` line plus the agent sitting gated until `ActiveDeadlineSeconds` (4h) fires. The PR description says a missing SCC binding "rejects the pod at admission, loudly" — true of the SCC, but not of a missing SA, which fails quietly at pod-creation inside the Job controller. Returning an error and skipping Job creation for that tick would make it fail where an operator is already looking. (`packages/controller/pkg/reconciler/storage_migration.go:246`)

- 🟡 **Warning:** this PR grants the migration Job **uid 0**, and no architecture page records that the platform now runs a root workload over agent data. `docs/architecture/security-and-credentials.md` is the page that owns trust boundaries and intra-cluster identity, and its ServiceAccount inventory (`:525`) enumerates the per-Agent SA as the identity model; `platform-migration` — a namespace-local SA whose whole purpose is to carry a uid-0 SCC grant on OpenShift — is a genuinely new trust-boundary fact, not an implementation detail. The page has ~7k characters of headroom under the 40k cap, so this is not a cap-forced omission. This sits just outside doc-drift's mechanical scope (the skill anchors on subsystem *behavior* pages and found nothing, correctly — see its section), but it is worth a human call before this merges, because it is the kind of fact that is invisible later. (`deploy/helm/platform/values.yaml:881`)

- 🟢 **Suggestion:** the shared scratch dir is `chmod 1777` inside a container whose two identities are root and one known uid, which is broader than needed and the comment justifies it partly by appeal to `protected_regular` — a sysctl the pod does not control (I could not read it in my environment; on a node where it is `0` the stated protection is simply absent). Since both identities are known at render time, `chown root:${AGENT_GID}` + `chmod 0770` gives the agent side exactly the access it needs with no world bit and no reliance on kernel hardening. (`packages/controller/pkg/reconciler/storage_migration.go:928`)

- 🟢 **Suggestion:** an **empty source** exits non-zero before it can report anything useful. `cat "$W"/sums.src.*` with no matching files returns 1 under `set -euo pipefail`; the `if [ "$entries" -gt 0 ]` guard protects the *copy* but the checksum collection at `:995` runs unconditionally before it. Verified: the empty case exits 1, the non-empty case exits 0. This is pre-existing (the old `sums()` had the same shape — I checked), so not a regression, but the refactor moved the code and this would be a cheap fix in passing: `cat "$W"/sums.src.* 2>/dev/null || true`. (`packages/controller/pkg/reconciler/storage_migration.go:995`)

- ✅ **Looks good:** dropping source reads to the agent uid via `setpriv` while keeping the writer as root is the right decomposition, and it is pinned by tests asserting each source touch goes through `$AS_AGENT` — the weaker "all root" or "all agent" variants cannot quietly return.
- ✅ **Looks good:** verification got genuinely stronger: the uid column is now compared rather than cut out, and `allowOwnershipRemap` is deprecated in place (kept for config parsing) rather than removed, so existing values files still decode.
- ✅ **Looks good:** `automountServiceAccountToken: false` on both the SA and the pod, with the token-mount assertion in the new test — the SA exists purely as an SCC grant target and cannot become an API identity by accident.

✅ No findings.

`docs/architecture/persistence.md:125` describes the storage migration at the correct altitude — "copies the quiesced volume onto a fresh RWO PVC in a Job, verifies the copy by checksum" — and this PR changes *how* the copy achieves fidelity, not the subsystem's behavior or responsibility. No architecture page states the copy's uid, `fsGroup`, or privilege model today (no `runAsNonRoot` / `non-root` / `restricted SCC` / `uid 0` claims anywhere under `docs/architecture/`), so no page contradicts the new design — drift is measured code-vs-docs, and there is no docs-side claim to have drifted. No architecture page was edited, so no `Last verified:` bump is due; no ADR changed state. The uid-0 trust-boundary observation above is a reviewer judgement call under the guidelines' "trust boundaries" *include* rule, not a mechanical drift hit — which is why it is a 🟡 in the findings and not a ❌ here.

### Verdict
REQUEST_CHANGES — Two execution-verified critical defects: the new `chmod 2770 "$dst"` makes every retry fail permanently (verified: attempt 1 passes, attempts 2–4 fail on a `700` source dir landing `2700`), and archiving `.` without excluding `lost+found` breaks on any ext4 source with a root-owned `0700` one (verified: tar exits 2 as the agent uid). Both are one-line fixes and both are invisible to the first-attempt-on-fresh-volume path the rehearsal covered.

---

dam-code-guardian:
🛡️ **Code Guardian 🛡️** — correction to my review at `27e0904`

One claim in that review was **wrong**. Recording it here rather than leaving it in a merged thread.

I wrote that the dir-mode restore "silently swallows failures" because "a non-zero exit of the last command inside a `while read` body does not trip `set -e`". **That is false.** I tested it with an inline `( set -e; ... )` subshell in an interactive shell instead of as a script file. The control case I never ran is the tell — in that harness even `/bin/false` "passes":

```
( set -e; /bin/false; echo REACHED )   # prints REACHED, exit 0   <- void harness
bash script-with-set-euo-pipefail.sh   # exit 1, no REACHED       <- reality
```

Run properly, a failing `chmod` in that loop **does** abort:

```
phase: copied
chmod: cannot access 'dst/./lost+foundry': No such file or directory
>>> job exit=1
```

Two consequences for that review:

- The 🟢 suggesting an `|| { echo ...; exit 1; }` guard is **moot** — the loop already failed closed.
- The first 🟡 had the right diagnosis but the **wrong consequence, and I oversized the severity**. I said the predicate mismatch would surface as a confusing `cmp` mismatch during verification. It would actually have aborted earlier and more usefully, at the `chmod`, naming the offending path. Still a real bug — a `lost+foundry` directory would have failed the migration — but self-diagnosing, not the silent-drift-then-baffling-diff I described.

The 🟡 about `chmod "00$mode"` being width-dependent **stands**; that one I verified correctly (`chmod 0040` on a `2770`-inherited dir → `2040`, bit retained).

All of this is academic for the shipped code: `42d6783` replaced the approach with a split-identity copy (read as the agent, write as root) and no per-directory chmod restore remains on `main`. Posting anyway, because a review asserting something false about `set -e` is worse than no review — nobody reading this thread later should carry that claim away.

Fixed on my side: shell semantics get verified by running a script file, with a `/bin/false` control asserting the harness actually propagates errors.

---

PR #3103, opened 2026-07-30
fix(controller): migration retries survive the shaped root; source lost+found excluded

jezekra1:
## Both review criticals on #3094 are correct — reproduced by execution here before fixing, and both are regressions that PR introduced

**C1 — the `chmod 2770` success-shaping poisoned retries.** The shaping persists on the filesystem; any attempt that fails *after* it leaves a setgid root, and extraction under a setgid root lets the kernel stamp the bit into every `0700`-style directory tar creates (tar only re-chmods dirs it has a reason to revisit — `700` is exactly the mode it doesn't, and `700` is `.ssh`, `.gnupg`, every private-tree mkdir). Reproduced: against a shaped root, every attempt fails forever, `.private` landing `2700`. **Fix:** every attempt neutralizes the root at wipe time (`chown` agent + `chmod 00755` — five digits, GNU chmod preserves dir setgid for shorter numerics), and the `2770` shaping moves to strictly **after** verification — extraction always runs under a setgid-free root, and only a fully verified copy gets the kubelet signature.

**C2 — archiving `.` dropped the `lost+found` exclusion** the old member-list form had. On an ext4 **source** with the usual root-owned `0700` `lost+found`, the agent-identity reader gets EACCES, tar exits 2 — and the inventory walk *prunes* the directory, so the unreadable gate never names it (bare tar error instead of the actionable message). Reproduced. **Fix:** `--exclude=./lost+found` on the reader (anchored — a nested `lost+found` the workspace owns still copies; the `.` member and its root-attr restore stay).

## Also from the review

- **Missing SA now fails the pass loudly before anything is gated** — previously a swallowed Warn led to a Job whose pods sat in `ServiceAccount not found` until the 4h deadline.
- Scratch dir is `0770 root:agent` instead of `1777`-plus-`protected_regular`-reasoning.
- Empty source no longer trips `pipefail` in checksum collection.
- `security-and-credentials.md` records `platform-migration` as the platform's one uid-0 workload, in the SA inventory, with the split-identity rationale.

## Proof (fixture = real squashing kernel-NFS source, byte-for-byte script from `buildMigrationJob`)

Starting from a **deliberately poisoned `2770` root** and a source that **contains a root-owned `0700` `lost+found`**: four consecutive attempts verify end-to-end, `.private` lands exactly `700` every time, re-poisoning the root between attempts self-heals, and an empty source verifies. Unit suite `-count=1` (no cache), vet/staticcheck clean. New tests pin: the `00755` neutralize, the shaping ordered after content `cmp`, the archive exclusion, the `0770` scratch dir, and the SA-failure-blocks-the-pass behavior.

## Process note

The retry matrix in #3094's fixture ran *before* the shaping line was added in its final commit, and only the fresh pass was re-run after — that's how C1 shipped despite a 'retry-proven' claim. The fixture loop now starts from a poisoned root precisely so that ordering mistake can't hide again.

### DIFF
--- a/docs/architecture/security-and-credentials.md
+++ b/docs/architecture/security-and-credentials.md
@@ -1,6 +1,6 @@
 # Security and credentials
 
-Last verified: 2026-07-29
+Last verified: 2026-07-30
 
 ## Overview
 
@@ -522,6 +522,18 @@ The agent and the gateway are gated by different mechanisms — they live
 on opposite sides of the credential boundary, so the threat models
 differ:
 
+- **`platform-migration` ServiceAccount** in the agent namespace — the
+  identity of the one-time storage-migration copy Job, and the only
+  workload on the platform that runs as **uid 0**. The Job needs root
+  solely for the target side of the copy (owning a freshly provisioned
+  volume root, restoring exact file ownership); every read of the agent's
+  data drops to the agent's own uid, so a root-squashing source share
+  never sees uid 0. The SA carries no role bindings and its token is
+  never mounted (`automountServiceAccountToken: false` on both the SA and
+  the pod), so it cannot act against the API; its sole purpose is to
+  scope the OpenShift SCC grant that permits uid 0 to exactly this
+  workload — an ops-side, out-of-band binding. The pod joins no mesh and
+  mounts no credentials.
 - **Per-Agent ServiceAccount** in the agent namespace, name ==
   Agent ID. Both pods of the long-lived pair run as this SA, but
   only the *gateway* pod is a mesh participant — istiod stamps it with
--- a/packages/controller/pkg/reconciler/storage_migration.go
+++ b/packages/controller/pkg/reconciler/storage_migration.go
@@ -243,13 +243,13 @@ func (m *StorageMigrationManager) ReleaseGated(ctx context.Context) {
 // OpenShift install can scope the runs-as-root SCC grant to exactly this
 // workload (an ops-side, out-of-band binding, like every cluster-scoped
 // security object).
-func (m *StorageMigrationManager) ensureServiceAccount(ctx context.Context) {
+func (m *StorageMigrationManager) ensureServiceAccount(ctx context.Context) error {
 	_, err := m.client.CoreV1().ServiceAccounts(m.config.Namespace).Get(ctx, migrationServiceAccount, metav1.GetOptions{})
-	if err == nil || !errors.IsNotFound(err) {
-		if err != nil {
-			slog.Warn("storage migration: checking service account failed", "error", err)
-		}
-		return
+	if err == nil {
+		return nil
+	}
+	if !errors.IsNotFound(err) {
+		return fmt.Errorf("checking migration service account: %w", err)
 	}
 	sa := &corev1.ServiceAccount{
 		ObjectMeta: metav1.ObjectMeta{
@@ -260,10 +260,10 @@ func (m *StorageMigrationManager) ensureServiceAccount(ctx context.Context) {
 		AutomountServiceAccountToken: ptrBool(false),
 	}
 	if _, err := m.client.CoreV1().ServiceAccounts(m.config.Namespace).Create(ctx, sa, metav1.CreateOptions{}); err != nil && !errors.IsAlreadyExists(err) {
-		slog.Warn("storage migration: creating service account failed", "error", err)
-		return
+		return fmt.Errorf("creating migration service account: %w", err)
 	}
 	slog.Info("storage migration: service account ensured", "name", migrationServiceAccount)
+	return nil
 }
 
 // Reconcile advances every in-flight migration one step and admits new
@@ -342,7 +342,14 @@ func (m *StorageMigrationManager) Reconcile(ctx context.Context) {
 	sort.Strings(names)
 
 	if len(names) > 0 {
-		m.ensureServiceAccount(ctx)
+		// Without the SA the Job would be admitted but its pods would never
+		// schedule (ServiceAccount not found) -- a quiet 4h stall behind the
+		// deadline. Fail the pass instead: nothing is gated or created this
+		// tick, and the next tick retries.
+		if err := m.ensureServiceAccount(ctx); err != nil {
+			slog.Warn("storage migration: skipping pass, service account unavailable", "error", err)
+			return
+		}
 	}
 
 	slots := m.concurrency() - len(inFlight)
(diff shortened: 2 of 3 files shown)


xjacka (approved):

dam-code-guardian:
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `eba7082`

_Posted as a comment, not a review: the PR merged mid-review (at `5c91d3c`, 21:50Z) and GitHub no longer accepts a review event against the old head. The three changed files are byte-identical between the reviewed head and the merge commit, so the findings below stand — as follow-ups, not blockers._

### Summary
Fixes two regressions #3094 introduced in the storage-migration copy Job: the `chmod 2770` success-shaping ran before verification, so a failure left a setgid target root that poisoned every subsequent retry; and switching the archive to `.` dropped the `lost+found` exclusion, making an ext4 **source** unreadable to the agent-identity reader. Also promotes the migration ServiceAccount check from a swallowed warning to a hard pass-abort, narrows the scratch dir from `1777` to `0770 root:agent`, and documents `platform-migration` in the security page's SA inventory.

### Findings

Both criticals from #3094 are real and the fixes are right. I reproduced each mechanism in a fixture (GNU tar 1.35, GNU coreutils):

- ✅ **Looks good:** the setgid-poisoning mechanism is exactly as described. Extracting under a `2770` root stamps setgid into every `0700` dir tar creates (`.private` → `2700`, recursively); under a `00755` root the same extraction lands `700`. And `chmod 00755` is genuinely the load-bearing detail — `755` and `0755` **preserve** dir setgid, only the five-digit form clears it. The test at `storage_migration_test.go:201` pins the five-digit form, which is the one thing a future reader would "simplify" away.
- ✅ **Looks good:** `--exclude=./lost+found` is correctly anchored — a root `lost+found` is excluded while `./nested/lost+found` still copies, matching the `-prune` in both walks. Verified.
- ✅ **Looks good:** the ordering assertion via `strings.Index` (`storage_migration_test.go:203`) pins the actual regression — that the shaping follows the content `cmp` — rather than just its presence. This is the assertion that would have caught #3094.
- ✅ **Looks good:** the SA change is a real improvement in failure mode, and `ensureServiceAccount` is called *before* any gating (`storage_migration.go:349`, ahead of the `slots` loop), so `TestStorageMigration_ServiceAccountFailureBlocksThePass` asserts the right invariant — nothing annotated, no Job created.
- ✅ **Looks good:** `{ cat …; } || true` on both sum collections fixes a genuine `set -e` abort — an unmatched glob makes `cat` exit 1 and kills an empty-source migration. Confirmed in a real script file.

- 🟡 **Warning:** the target root's **ownership** is no longer shaped on success, only its mode (`storage_migration.go:1019` vs `:1067`). The removed block was `chown agent:agent` + `chmod 2770` together; the new code moved only the `chmod` to after verification and left the `chown` up at wipe time — where tar then overwrites it. The `.` member carries the source root's owner and mode, and extraction with `--same-owner --numeric-owner` applies both, so the pre-tar `chown` is undone: I measured a `00755 agent:agent` root come out of extraction as `2777 <source-owner>` and stay at that owner after the final `chmod 2770`. When the source root is agent-owned (the normal case) the outcome is identical to before, so this is latent rather than broken today — but a source volume whose root is root-owned (a statically provisioned share, an NFS export dir) now yields a `2770 root:root` target root, and `fsGroupChangePolicy: OnRootMismatch` (`deploy/helm/platform/values.yaml:881`) will then rewrite ownership recursively across the whole workspace on first mount — precisely the recursive pass the shaping exists to avoid. Suggest restoring the `chown` next to the `chmod 2770` at `:1067`, keeping the one at `:1019` for the write test.

- 🟡 **Warning:** the agent-writability gate at `storage_migration.go:1044` now tests a mode that the code then replaces, so it no longer proves what its comment claims ("The AGENT must be able to use the volume it wakes up on"). It runs after extraction, so the root carries the *source's* mode/owner, and `chmod 2770` at `:1067` changes them afterwards. Two concrete gaps: a source root of `0777` passes the test and then becomes `2770` under a possibly non-agent owner (no access); and a source root of `0500` fails the test even though the final shaping would have made it writable — I hit exactly that EACCES in a fixture. Moving the touch/rm to after `:1067` would restore the invariant. (Fixing the `chown` above shrinks this to the second case, but does not remove it.)

- 🟢 **Suggestion:** `chgrp "${AGENT_GID}" "$W"` + `chmod 0770` (`storage_migration.go:935`) is a good tightening, but the comment above it still describes the old behavior — "root-created, sticky, world-writable" — and the `protected_regular` rationale it gives no longer applies the same way now that the dir isn't world-writable. Worth updating so the next reader doesn't restore `1777` to match the comment. The test's `assert.NotContains(t, runnable, "chmod 1777")` guards the code but not the comment.

- 🟢 **Suggestion:** the security page calls `platform-migration` "the only workload on the platform that runs as **uid 0**" (`docs/architecture/security-and-credentials.md:519`). The `egress-lockdown` init container also runs `runAsUser: 0` with NET_ADMIN/NET_RAW (`packages/controller/pkg/reconciler/iptables_init.go:81-97`), is wired into both the StatefulSet and ephemeral-pod paths, and is `enabled: true` in the agent-base block — and the chart's own `anyuidCapNet` comment describes it as running as root. It's a short-lived init container rather than a workload, so the claim is defensible, but since this sentence is the SA inventory's justification for an SCC grant, "the only *workload*" invites a reader to conclude no other root-capable container exists. Suggest "the only pod that runs as uid 0 for its whole lifetime" or a clause acknowledging the init container.

- 🟢 **Suggestion:** the `|| true` on both sum collections makes a *vacuous* content verification possible: if both sides produce zero sums, `cmp` compares two empty files and passes. Today `entries > 0` plus the metadata `cmp` make that unreachable in practice, so this is defense-in-depth on a path where the old volume is deleted on the strength of these comparisons — an `[ "$entries" -eq 0 ] || [ -s "$W"/src.sum ]` assertion would close it cheaply.

**Verdict: aligned**

`docs/architecture/security-and-credentials.md` was edited with `Last verified:` bumped (2026-07-29 → 2026-07-30) ✓. The behavior changes in the diff sit below the level `docs/architecture/persistence.md` → *Storage migration (transitional)* describes, which remains accurate. No ADR references or `Motivated by:` sections in the added block; no new subsystem; no architecture page moved or renamed; no ADR status change in the diff. Page is 33,866 chars, within the 40,000 cap.

### Verdict
**COMMENT** — both criticals are correctly fixed and well-pinned by tests; the two 🟡s are follow-ups on the same lines, not blockers on the fix.

---

PR #3104, opened 2026-07-30
fix(helm): storage migration off by default pending #3103

jezekra1:
One-liner: `storageMigration.enabled: true → false` pending the critical fixes in #3103.

Safe by construction: a disabled manager **releases** anything it had gated (bins abandoned Jobs and half-copied targets, restores prior run state) — every agent returns to the volume it started from. Installs that want to keep draining can set it back explicitly. Re-flip to on-by-default can ride #3103 or follow it.

### DIFF
--- a/deploy/helm/platform/values.yaml
+++ b/deploy/helm/platform/values.yaml
@@ -1062,7 +1062,7 @@ controller:
   # leaves each agent on the volume it started from. Watch the controller log
   # through the first drain; `concurrency: 1` makes that easy to follow.
   storageMigration:
-    enabled: true
+    enabled: false
     # -- Storage class migrated workspaces land on: the DESTINATION of the
     # migration. Empty (default) = the cluster's default class, i.e. ordinary
     # storage, which is the point of the exercise.


xjacka (approved):

deploy/helm/platform/values.yaml
```diff
@@ -1062,7 +1062,7 @@ controller:
   # leaves each agent on the volume it started from. Watch the controller log
   # through the first drain; `concurrency: 1` makes that easy to follow.
   storageMigration:
-    enabled: true
+    enabled: false
```
dam-code-guardian, line 1065: 🟡 **Warning:** This flip leaves the comment block directly above it self-contradictory — line 1057 still opens with *"Enabled by default: the fleet drains itself on upgrade with no owner action, which is the point."*, and the following sentences frame `false` as *"turning it off"*, i.e. a deviation from the shipped default rather than the shipped default itself.

An operator reading that block to decide whether they need to act is told the opposite of what the chart now does. Since making the default off *is* this PR's purpose, the paragraph is worth rewording here: state that it is disabled by default pending #3103, keep the release-semantics sentence (it is accurate and load-bearing — the unwind-on-disable path in `RunLoop` genuinely does what it claims), and re-frame the escape-hatch line as "leaving it off is safe at any point".

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `c5d0706`

## PR #3104: fix(helm): storage migration off by default pending #3103

**Author:** jezekra1 | **Branch:** fix/migration-off-by-default → main | **Changes:** +1 −1 (1 files)

### Summary

Flips `controller.storageMigration.enabled` from `true` to `false` in the platform chart's `values.yaml`, making the one-time RWX→RWO workspace drain opt-in until the copy-Job fixes in #3103 land. One-line values change; no Go code touched.

### Findings

- 🟡 **Warning:** The comment block that documents this very key still opens with **"Enabled by default: the fleet drains itself on upgrade with no owner action, which is the point."** (`deploy/helm/platform/values.yaml:1057`) — eight lines above the now-`false` default. An operator reading the block to decide whether to act is told the opposite of what the chart does, and the rest of the paragraph frames off as "turning it off", i.e. a deviation from the shipped default rather than the shipped default itself. Worth rewording in this PR, since making the default off is the PR's entire purpose. Suggested shape: state that it is *disabled* by default pending #3103, keep the release-semantics sentence (it is accurate and load-bearing), and re-frame the escape-hatch line as "leaving it off is safe at any point".

- 🟢 **Suggestion:** `docs/architecture/persistence.md:125` ("Storage migration (transitional)") still describes the drain as something "the controller runs" unconditionally, with no mention that it is now opt-in. That page already documents default-enablement state for the neighbouring subsystem — line 117 says the warm pool "is disabled by default" — so the omission reads as drift rather than as out-of-scope detail. One clause is enough.

- ✅ **Looks good:** The body's central safety claim — *"a disabled manager releases anything it had gated … every agent returns to the volume it started from"* — is **verified in the code**, not just asserted. `StorageMigrationManager.RunLoop` short-circuits on `!Enabled` by calling `ReleaseGated` *before* returning (`packages/controller/pkg/reconciler/storage_migration.go:145`), and `ReleaseGated` handles both shapes correctly: source still migration-eligible → bin the copy Job and the unclaimed target (the `LabelMigrationFor` selector matches only unclaimed targets, so a completed migration's volume cannot be caught), then lift the gate; source gone → `finishFlip` completes the tail. `finishFlip` deletes only PVCs carrying `LabelMigrationSuperseded`, so an abandoned migration's intact source is never touched. The `wasRunning` path re-stamps `annLastActivity` so a previously-running agent wakes through the normal budget-gated path. Behaviour is pinned by a test that asserts exactly this (`storage_migration_test.go:482`: gate lifted, source untouched, half-copied target and Job binned).

- ✅ **Looks good:** No desync risk in the wiring or the rest of the tree. The chart passes the whole block through as `toJson` (`templates/controller/deployment.yaml:71`), so there is no second copy of the default to forget; the Go struct has no competing literal default (zero value is `false`, which now agrees with the chart); and no `values-*.yaml`, e2e values file, or CI workflow overrides or asserts on `storageMigration`, so nothing silently keeps the old behaviour.

- ✅ **Looks good:** The gate is a hard-stop-strength negative override only this manager clears, so shipping a disabled manager that *ignored* existing gates would strand agents at zero replicas — exactly the failure the unwind-on-disable path avoids. Getting that ordering right is what makes this one-liner safe to ship ahead of #3103 rather than a fleet-freeze.

**Verdict:** minor drift

**Drift**

- ❌ **Check 1 — architecture-page drift** (`docs/architecture/persistence.md:125`): the diff changes an operator-visible default of the storage-migration subsystem — it no longer runs unless explicitly enabled — while the subsystem's page still presents the drain as automatic ("the controller runs a one-time, interrupt-safe **storage migration**"). Proposed edit: add a clause to the section noting the migration is disabled by default and enabled per install, mirroring how line 117 records the warm pool's default. `Last verified:` needs no bump under check 2 (that applies only to architecture pages edited in the diff; none were).

### Verdict

COMMENT — the mechanism is sound and the "safe by construction" claim holds up under code inspection, so this is fine to ship ahead of #3103; but the values.yaml comment eight lines above the change still says "Enabled by default", which is worth fixing here rather than leaving the chart self-contradictory.

---

Issue #938, opened 2026-06-15
Epic: Custom images for target workloads


jenna-winkler:
## Summary

Establish a repeatable path for shipping curated workload images (Bugstone, OpenEvolve, Nous, K-Search, GEPA, etc.) that hosted DAM can pull and run out of the box. This epic owns image packaging, publishing, and hosting. The only UI work in scope is the temporary Bugstone exposure in #1274; the long-term creation experience is designed separately in #921.

## Problem

Onboarding a workload today is a bespoke effort - each one reinvents its own packaging, hosting, configuration, and deployment. There's no standard path, so every new workload starts from scratch.

## Goals

1. Ship each prioritized workload as a curated, pre-built image that launches with minimal setup.
2. Enable hosted DAM to reliably pull and run these images (including private pulls where required).
3. Establish a repeatable packaging and publishing pattern so each new workload is incremental, not bespoke.
4. Bundle workload-specific defaults and configuration into each image, ready for the creation flow to consume (#921).

## In Scope

- Building, pusblishing, hosting workload images
- Enabling image pulls from hosted DAM
- Temporary Bugstone UI exposure (#1274)

## Out of Scope

- Long-term agent creation and configuration UX (#921)

darrellreimer:
Thoughts on Coral?  I think that's where Nous started from or at least the team was exploring that.  And is there any way to determine the trajectories of these projects?  as in growing vs. older and declining?

sarahmilleribm:
Other frameworks:
[AdaEvolve](https://github.com/skydiscover-ai/skydiscover) - evolutionary (pop) - 
[EvoX](https://github.com/skydiscover-ai/skydiscover) - evolutionary (pop)
[ShinkaEvolve](https://github.com/SakanaAI/ShinkaEvolve) - evolutionary (pop)

tomkis:
@jenna-winkler i am proposing to close this in favour of new epic that would cover lower priority images and move that to Later?

jenna-winkler:
#2878

PR #3109, opened 2026-07-31
feat(agents): package SkyDiscover AdaEvolve & EvoX workload image

### LINKED ISSUE #1639: AdaEvolve & EvoX - code & algorithm optimization workload image
## Summary

Package the AdaEvolve and EvoX search strategies as a curated, pre-built workload image that hosted DAM can pull and run out of the box, following the repeatable packaging/publishing pattern established by this epic (#938).

AdaEvolve and EvoX are LLM-driven search methods for code and algorithm optimization - an LLM proposes candidate programs, an evaluator scores them, and the search loops toward better solutions across open-ended problems (math, systems, algorithm design). Both are owned by and ship from SkyDiscover ([skydiscover-ai/skydiscover](https://github.com/skydiscover-ai/skydiscover), Berkeley Sky lab, Apache-2.0), a unified framework for LLM-driven optimization. They are complementary:
- AdaEvolve** (`--search adaevolve`) - multi-island adaptive search; adapts search parameters on the fly for fast early gains.
- EvoX (`--search evox`) - a self-evolving paradigm that evolves the search strategy itself using LLMs for stronger long-horizon gains.

They build from one SkyDiscover install and are exposed to the user as two curated workload presets over a single image.

## Scope
- [x] Build a curated image from SkyDiscover with pinned dependencies
- [x] Expose AdaEvolve and EvoX as the two curated workload presets (default `--search` baked into each)
- [x] Bundle workload defaults/config so each preset launches with minimal setup
- [x] Publish + host the image where hosted DAM can pull it
- [x] Verify a sample run completes end-to-end in hosted DAM for both presets
- [x] Document runtime config (task dir, model selection)

## Out of scope

- SkyDiscover's wrapped/native backends for OpenEvolve, GEPA, and ShinkaEvolve (`--search openevolve|gepa|shinkaevolve|*_native`). Those are packaged from their own upstream repos (#885, #948, and the ShinkaEvolve image) - this image only exposes SkyDiscover's own strategies (AdaEvolve, EvoX).
- Other SkyDiscover search modes (Top-K, Beam Search, Best-of-N) unless prioritized later.

xjacka:
Closes #1639

## What

Packages SkyDiscover's two own search strategies — **AdaEvolve** and **EvoX** — as a curated workload image, following the packaging pattern established by the shinkaevolve and gepa images (#938 epic).

- **New `packages/agents/skydiscover/`** — built `FROM` the claude-code image; adds Python 3.11 + `skydiscover` in a venv at `/opt/skydiscover-venv`, installed from a **pinned upstream git ref** (`SKYDISCOVER_REF`) — the 0.1.0 wheel omits EvoX's config data files; a build assertion keeps that gap (and the still-unpackaged agentic tool schemas) from regressing silently. Base package only: AdaEvolve and EvoX need no extras; the `external` extra (wrapped OpenEvolve/GEPA/ShinkaEvolve backends) is deliberately **not** installed — those ship as their own workload images. Ships `AGENTS.md` (pod policy: pre-launch gate, run discipline, resume-on-wake via `--checkpoint`, hard guardrails) and the `skydiscover` skill (CLI reference, model setup, worked example).
- **Two presets over one image** (same one-image/N-templates pattern as k-search/k-search-local): `adaevolve` and `evox` templates in `values.yaml`, both pointing at `quay.io/dam-agents/skydiscover` and differing only in the injected `SKYDISCOVER_SEARCH` env, which the agent uses as the run's default `--search`.
- **Registration**: `scripts/resolve-image.sh` `WORKLOADS` (enrolls the image in CI's `build-workloads`/`merge-workloads` matrix) + self-check assertions, `mise.toml` task include, `values-local.yaml` local-build entries (disabled by default, like the other workloads).

## Verification

- `scripts/resolve-image.sh --self-check` passes; `mise run helm:check` (lint + kubeconform render) passes; the rendered agent-templates ConfigMap carries both presets with the right env.
- `mise run agents:skydiscover:image` builds; in-image smoke test: `skydiscover-run --help` works, and a 1-iteration run with a stub evaluator initializes **both** `--search adaevolve` and `--search evox` from the base install, producing the documented output layout (`best/best_program.py`, `checkpoints/checkpoint_<N>/`, `logs/`).
- The docs were corrected against the real CLI at the pinned ref: `initial_program` is optional (from-scratch path), agentic mode is unsupported in this image (upstream packaging omits its tool schemas — build canary added), checkpoints live under `<output>/checkpoints/`.

## Out of scope / follow-up

- End-to-end sample runs of both presets in hosted DAM need the published image, so they happen after this merges (CI publishes `quay.io/dam-agents/skydiscover` from main).
- SkyDiscover's other built-in modes (Top-K, Beam Search, Best-of-N) are installed but not exposed as presets, per the issue.

### DIFF
--- a/deploy/helm/platform/values-local.yaml
+++ b/deploy/helm/platform/values-local.yaml
@@ -92,6 +92,19 @@ agentTemplates:
       repository: platform-gepa
       tag: latest
 
+  # Both presets share the one locally-built skydiscover image.
+  adaevolve:
+    enabled: false
+    image:
+      repository: platform-skydiscover
+      tag: latest
+
+  evox:
+    enabled: false
+    image:
+      repository: platform-skydiscover
+      tag: latest
+
   # Test fixture; not in the values.yaml catalog — the mock image is never
   # release-tagged (CI pushes per-commit shas for the e2e job only).
   mock:
--- a/packages/agents/skydiscover/AGENTS.md
+++ b/packages/agents/skydiscover/AGENTS.md
@@ -128,7 +128,9 @@ estimate, but an informed user may pre-authorize it (see below).
   monitoring and crash recovery:
 
   ```sh
-  # run this script AS a backgrounded harness task
+  # run this script AS a backgrounded harness task (your Bash tool's
+  # run_in_background) — NEVER as a foreground command: the tool's timeout
+  # would SIGTERM the whole process group, run included, mid-flight
   dir="$SKYDISCOVER_OUTPUT_ROOT/<run-id>"
   cd "$dir"
   export OPENAI_API_KEY="${OPENAI_API_KEY:-placeholder}"   # gateway overwrites it on the wire
@@ -184,7 +186,10 @@ completed from the checkpoint numbering (`checkpoint_40` → 40 done) and set
 checkpoint exists yet** (evox writes its first only at iteration 10), a
 relaunch restarts from scratch and re-spends the lost iterations — that
 exceeds the originally approved spend, so say so and wait for a fresh
-go-ahead instead of silently relaunching. A run that's reached
+go-ahead instead of silently relaunching. The same rule covers a run you
+killed yourself (a defective evaluator, a wedged process): announcing the
+relaunch is not enough — re-state the total spend it implies and get the
+go-ahead, unless the user pre-authorized re-runs. A run that's reached
 its budget is done; raising the budget is a new, re-gated decision, not a
 resume.
 
--- a/packages/agents/skydiscover/AGENTS.md
+++ b/packages/agents/skydiscover/AGENTS.md
@@ -1,0 +1,237 @@
+# Agent pod environment — SkyDiscover (AdaEvolve / EvoX)
+
+You are running inside an isolated **SkyDiscover** agent pod on the platform.
+Your job is to run **LLM-driven code & algorithm optimization** on the user's
+behalf: take a target repo (or a from-scratch problem) and a *measurable*
+objective, author the SkyDiscover inputs, run the search, and report the
+winning program. You are conversational — not a CLI passthrough and not a
+general-purpose coding agent. Keep the work centered on setting up and
+operating SkyDiscover runs.
+
+Your home directory and workspace are persistent; the rest of the filesystem is
+reset on pod restart. Network egress is proxied through the platform's
+credential gateway, so `git`, `gh`, the Claude API (your own model), and the
+SkyDiscover model endpoint all work **without any API key in this pod** —
+never ask the user for a key, and never write credentials to disk.
+
+## What SkyDiscover is
+
+SkyDiscover (Berkeley Sky lab) is a unified framework for LLM-driven
+optimization: an LLM proposes candidate programs, an **evaluator** scores each
+one, and the search loops toward better solutions. You drive it through the
+`skydiscover-run` CLI; SkyDiscover drives the search loop.
+
+**This pod is preset to one of SkyDiscover's two own strategies** via
+`$SKYDISCOVER_SEARCH`:
+
+- **`adaevolve`** — multi-island adaptive search (UCB, migration, paradigm
+  breakthroughs); adapts search parameters on the fly for fast early gains.
+  Strongest on short budgets (≲50 iterations).
+- **`evox`** — a self-evolving paradigm that co-adapts solution generation and
+  experience management, evolving the search strategy itself; stronger on
+  long-horizon runs (≳50 iterations).
+
+Default every run's `--search` to `$SKYDISCOVER_SEARCH` — that is what the
+user picked when they chose this preset. If the run's budget clearly fits the
+*other* strategy better, say so in one line (both are installed), but let the
+user decide.
+
+**The `skydiscover` skill is your reference** for the CLI surface, the model
+setup, and how to author the run inputs (evaluator + optional initial
+program). Consult it whenever you set up a run. This file is the
+*how-to-operate-in-this-pod* layer.
+
+## Two model paths (you only configure one)
+
+- **You, the driver** — Claude Code. Your own model calls authenticate through
+  the inherited model gateway; you do not configure or pick that model here.
+- **The search loop** — `skydiscover-run` calls an **OpenAI-compatible**
+  endpoint injected by the attached model-provider connection as
+  `OPENAI_BASE_URL` + `OPENAI_API_KEY`. *This* is the model you configure, by
+  discovering what the endpoint serves and passing `--api-base` plus a plain
+  model id the endpoint actually lists (the skill's Step 1 defines the exact
+  procedure). A single IBM-LiteLLM-class connection feeds both paths; a
+  pure-OpenAI provider feeds only the loop. If `OPENAI_BASE_URL` is unset,
+  stop and tell the user to attach a model-provider connection.
+
+## Starting a conversation
+
+At the **start of every new conversation**, before anything else, enumerate
+existing runs and offer to act on them. Scan `$SKYDISCOVER_OUTPUT_ROOT`
+(`~/skydiscover-runs`) for run directories (each has a `task/` and an
+`output/` dir) and classify each as:
+
+- **running** — its `run.pid` names a live process **that is actually the
+  run**:
+
+  ```sh
+  pid=$(cat run.pid) && kill -0 "$pid" 2>/dev/null \
+    && tr '\0' ' ' < "/proc/$pid/cmdline" | grep -q skydiscover-run
+  ```
+
+  The cmdline check is not optional: a pod restart resets the PID namespace,
+  so a stale `run.pid` on persistent `$HOME` can name an unrelated live
+  process — bare `kill -0` would misclassify a dead run as running and
+  silently skip its resume.
+- **not running** — finished, stopped, or paused by a pod hibernation (below).
+
+Present the grouped list, then offer a status pull (tail `run.log`, read
+`output/best/best_program_info.json`, count `output/checkpoints/`) or a
+resume. If the user instead opens with a concrete
+task ("optimize X in repo Y to improve Z"), do that — but still mention any
+currently-running search in one line.
+
+## The pre-launch gate (mandatory)
+
+**Work through all four before launching a full search run.** A run is
+autonomous and spends real tokens per iteration. Steps 1–3 are *correctness*
+checks that protect the user's own tokens, so they always run — "go fast" lets
+you run them inline without narrating each one, but it does **not** let you
+drop them (the smoke-eval especially: skipping it can silently burn the whole
+run on a miswired evaluator). Step 4 is a *consent* check: always show the
+estimate, but an informed user may pre-authorize it (see below).
+
+1. **The objective is measurable.** You must be able to write an evaluator
+   that returns a number for "better." If the user's goal isn't measurable as
+   a score (e.g. "make it nicer"), **refuse and clarify** — propose a concrete
+   metric (runtime, accuracy, size, error rate) and agree on it first.
+2. **You've authored the run inputs** — `evaluator.py` (an `evaluate(...)`
+   returning `combined_score`) and `initial.py` (with `EVOLVE-BLOCK` markers;
+   for a from-scratch problem, a minimal stub) — both are required CLI
+   arguments; see the skill.
+3. **You've run a smoke-eval** — call the evaluator directly on the initial
+   program and show the user a baseline score they recognize as sensible.
+   This catches the silent failure mode where the evaluator runs but scores
+   the wrong thing.
+4. **You've presented an iteration/cost estimate.** State the iteration budget
+   (`-i`), the resulting rough LLM-call count (≈ one proposal plus one
+   evaluation per iteration; EvoX adds occasional strategy-evolution calls),
+   that it runs autonomously, and the keep-awake tradeoff (below). Then **wait
+   for an explicit go-ahead — unless the user already pre-authorized this
+   run** ("just launch it," "don't ask"): pre-authorization waives the *wait*,
+   never the estimate — show the numbers, then launch. (Resuming an
+   already-approved run after hibernation needs no new confirmation — see
+   resume-on-wake.)
+
+## Run discipline
+
+- **Launch backgrounded** so you stay conversational and can poll while it
+  runs. Keep the PID and log in the run directory:
+
+  ```sh
+  dir="$SKYDISCOVER_OUTPUT_ROOT/<run-id>"
+  cd "$dir"
+  export OPENAI_API_KEY="${OPENAI_API_KEY:-placeholder}"   # gateway overwrites it on the wire
+  nohup skydiscover-run task/initial.py task/evaluator.py \
+    --search "$SKYDISCOVER_SEARCH" \
+    -i <N> -m <model-id> --api-base "$base/v1" \
+    -o "$dir/output" \
+    > run.log 2>&1 &
+  echo $! > run.pid
+  ```
+
+- **Always pass an explicit `-o`** on the **persisted** workspace
+  (`$SKYDISCOVER_OUTPUT_ROOT`, on `$HOME`) and **outside the cloned target
+  repo** — so checkpoints survive hibernation and never pollute the target.
+- **Always bound the run**: `-i` is the iteration budget — set it to what the
+  user approved, never unbounded. The approved total is the source of truth
+  across resumes.
+- **Clone the target into its own run directory**, never the pod home root or
+  an unrelated path. Give each run a unique web-safe `<run-id>` (repo +
+  objective slug; append `-2`, `-3` on collision).
+- **Watch for stalls.** A wedged LLM call can hang an iteration indefinitely.
+  If `run.log` and the newest `output/checkpoints/` entry haven't advanced in
+  ~30 minutes, kill the process and relaunch per resume-on-wake below — the
+  run resumes from the last checkpoint.
+
+## Run dependencies
+
+Candidate code runs in the SkyDiscover venv (`$SKYDISCOVER_VENV`), which has
+only `skydiscover` and its base deps (numpy, pyyaml, tqdm, the openai client).
+PyPI egress is open, so install whatever the run needs into that venv
+(`uv pip install --python "$SKYDISCOVER_VENV/bin/python" …`) — and anticipate
+what the **evolved** code will reach for, not just the initial program's
+imports (e.g. `scipy` for a numerical-optimization task). The venv is
+ephemeral but the uv cache is on persistent `$HOME`, so reinstall after a
+restart — it's fast.
+
+## Surviving hibernation (resume-on-wake)
+
+The pod **scales to zero when the session goes idle** — no active turn, no
+queued prompt, no open terminal/SSH session. That kills any background
+`skydiscover-run`. The output dir lives on persistent `$HOME`, so the run is
+recoverable but **does not progress while you're not engaged**.
+
+So at the **start of each turn**, check any run you care about: if its
+`run.pid` is dead and it hasn't reached its budget, reinstall the run's deps
+(the venv reset) and relaunch with `--checkpoint` pointing at the **latest**
+checkpoint under `output/checkpoints/`. Read how many iterations already
+completed from the checkpoint numbering (`checkpoint_40` → 40 done) and set
+`-i` to the **remainder** of the user-approved total — never more. A run that's reached
+its budget is done; raising the budget is a new, re-gated decision, not a
+resume.
+
+**Keep-awake escape hatch:** for a long search that must progress
+continuously (e.g. overnight), tell the user to keep a **terminal or SSH
+session open** to this agent — that pins the pod awake, so a backgrounded run
+completes without hibernation gaps. Genuinely unattended overnight progress is
+a future capability; today a run advances only while you're engaged or a
+session is open.
+
+## Hard guardrails
+
+- **Only use plain model ids the endpoint serves, always with `--api-base`**
+  (the skill's Step 1 format). Vendor-prefixed ids (`gemini/…`,
+  `anthropic/…`) route through vendor SDKs that demand keys this pod doesn't
+  hold. Never leave the run on SkyDiscover's default model.
+- **Export a non-empty `OPENAI_API_KEY`** before launching (the
+  `${OPENAI_API_KEY:-placeholder}` idiom above). The injected value is often
+  an empty placeholder by design — the gateway overwrites the header on the
+  wire, but the client refuses to send a request with no key at all.
+- **Only `--search adaevolve` or `--search evox`** (default
+  `$SKYDISCOVER_SEARCH`). The wrapped backends
+  (`openevolve|gepa|shinkaevolve|*_native`) are **not installed** in this
+  image — they ship as their own platform agent types; point the user there
+  instead of pip-installing the `external` extra.
+- **Every evaluator must return `combined_score`** in its result dict. That is
+  the score the search selects on; without it candidates can't be ranked.
+- **Leave the live monitor dashboard off** (`monitor.enabled` in config) and
+  don't reach for `skydiscover-viewer` — this pod exposes no UI ports; report
+  progress from `run.log` and the checkpoint files instead.
+- **Discover and validate the model before launching.** A model name the
+  endpoint doesn't serve fails every proposal. See the skill's model-setup
+  step.
+- **Refuse if the objective isn't measurable** (see the pre-launch gate).
+- **Always bound the run** (`-i`).
+
+## GitHub access goes through the connection — never a held token
+
+`git clone`, `gh`, and `gh pr create` work **because of the granted GitHub
+connection**, not a token in the pod: the pod holds only a placeholder, and
+Envoy injects the real credential on the wire to the allowed GitHub hosts. So:
+
+- **Never** introduce a side path that puts a raw token in the agent or the run
+  subprocess — no PAT in env, no `gh auth login` with a literal token, no
+  writing credentials to disk. If the user offers a token, decline and point
+  them at the connection.
+- **Confirm a connection is attached before promising a PR** — check
+  `PLATFORM_GH_TOKEN_AVAILABLE` (`true` when granted) or `gh auth status`; if
+  it's missing, tell the user to grant one (and still never take a token). A
+  public repo clones read-only without one.
+- Credential injection is **host-keyed**: any in-pod process reaching an
+  allowed GitHub host — *including LLM-generated candidate code* — gets the
+  credential. The control surface is therefore the **connection's scope**: keep
+  the GitHub connection least-privilege (the target repo, minimal permission).
+  A report-only run against a public repo needs no GitHub write at all.
+
+## Where things live
+
+- **Per-run directory** = `$SKYDISCOVER_OUTPUT_ROOT/<run-id>/`
+  (`~/skydiscover-runs`, on persistent `$HOME`). Holds `task/`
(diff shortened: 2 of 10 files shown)


deploy/helm/platform/values-local.yaml
```diff
@@ -92,6 +92,19 @@ agentTemplates:
       repository: platform-gepa
       tag: latest
 
+  # Both presets share the one locally-built skydiscover image.
+  adaevolve:
```
dam-code-guardian, line 96: 🟡 **Warning:** Neither preset name maps to a local build task, so flipping one of these to `enabled: true` builds nothing.

Both local loops derive the mise task from the rendered ConfigMap key:

```bash
if [ -d "packages/agents/$name" ]; then AGENT_TASKS+=("agents:$name:image")
...
else echo "WARN: no build task for '$name' ($image) — skipping" >&2
```

(`deploy/tasks.toml:501` in `cluster:install`, `:768` in `cluster:build-agent`). `packages/agents/adaevolve` and `packages/agents/evox` don't exist — only `packages/agents/skydiscover` does — so an enabled preset hits the `WARN` branch, `platform-skydiscover:latest` is never built or imported into k3s, and the pod fails to pull (`pullPolicy` is unset, and the tag is `:latest` locally).

The `k-search`/`k-search-local` precedent cited in the values.yaml comment doesn't carry over: there, `packages/agents/k-search` matches the *first* preset's name, so `k-search` builds through the standard path and only the deliberately-`enabled: false` `k-search-local` is unmappable. Here neither name matches.

Cheapest fixes: name one template `skydiscover`, or teach the two loops to fall back to the image repository's basename when the template name has no package dir.

packages/agents/skydiscover/README.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+Override the pinned release with `SKYDISCOVER_VERSION`:
+
+```sh
+SKYDISCOVER_VERSION=0.1.0 mise run agents:skydiscover:image
+```
+
+`values-local.yaml` points the adaevolve/evox templates at the locally-built
+`platform-skydiscover:latest` but keeps them `enabled: false`; flip one to
```
dam-code-guardian, line 79: 🟢 **Suggestion:** This paragraph is inherited from the gepa/shinkaevolve READMEs, where the template name equals the package dir so `enabled: true` is genuinely all it takes. With `adaevolve`/`evox` over `packages/agents/skydiscover`, flipping the flag alone leaves the image unbuilt (see the `values-local.yaml` comment). Worth spelling out the manual build step here, or renaming a template so the sentence stays true.

packages/agents/skydiscover/workspace/.agents/skills/skydiscover/SKILL.md
```diff
@@ -0,0 +1,238 @@
...
+    return {"combined_score": 1.0 / (1.0 + mse)}
+```
+
+Run it (after Step 1 resolved `$base` and a model id):
+
+```sh
+export OPENAI_API_KEY="${OPENAI_API_KEY:-placeholder}"
+skydiscover-run task/initial.py task/evaluator.py \
+  --search "$SKYDISCOVER_SEARCH" \
+  -i 10 -m <model-id> --api-base "$base/v1" -o output
```
dam-code-guardian: 🟢 **Suggestion:** The bare relative `-o output` contradicts this file's own CLI-reference rule (`-o, --output` — "**always** an explicit path on `$SKYDISCOVER_OUTPUT_ROOT`, outside the target repo") and `AGENTS.md`'s "Always pass an explicit `-o` on the persisted workspace." An agent copying this block verbatim lands checkpoints in whatever the CWD happens to be — exactly the state hibernation is documented to lose.

```suggestion
  -i 10 -m <model-id> --api-base "$base/v1" -o "$SKYDISCOVER_OUTPUT_ROOT/sin-approx/output"
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `ca3bc7b`

## PR #3109: feat(agents): package SkyDiscover AdaEvolve & EvoX workload image
**Author:** xjacka | **Branch:** feat/skydiscover-workload-image → main | **Changes:** +695 −2 (9 files)

### Summary
Packages SkyDiscover's two native search strategies (AdaEvolve, EvoX) as one curated claude-code-derived workload image with two Helm template presets that differ only in an injected `SKYDISCOVER_SEARCH` env var, plus registration in `resolve-image.sh`'s `WORKLOADS`, the mise task graph, and `values.yaml`/`values-local.yaml`. The Dockerfile, tasks.toml, README, AGENTS.md, and skill all track the gepa/shinkaevolve precedents closely.

### Findings
- 🟡 **Warning:** The two template names (`adaevolve`, `evox`) don't match the package directory (`packages/agents/skydiscover`), so the local build/load loops can't map either template to a build task. Both `cluster:install` (`deploy/tasks.toml:501`) and `cluster:build-agent` (`deploy/tasks.toml:768`) derive the task name from the rendered ConfigMap key: `if [ -d "packages/agents/$name" ]` → `agents:$name:image`, else `WARN: no build task for '<name>' — skipping`. Following the README's own instruction to flip a preset to `enabled: true` in `values-local.yaml` therefore prints a warning, builds nothing, and the pod ends up with no `platform-skydiscover:latest` in k3s. The cited `k-search`/`k-search-local` precedent doesn't cover this: `k-search-local` is `enabled: false` in **both** values files and, unlike here, `packages/agents/k-search` exists under the *other* preset's name so `k-search` at least builds. Nothing here maps to `skydiscover`. (`deploy/helm/platform/values-local.yaml:96`)
- 🟢 **Suggestion:** The README's local-build note inherits the sibling wording but not their property — for gepa/shinkaevolve the template name *is* the package dir, so flipping the flag works. Worth documenting the extra step (`mise run agents:skydiscover:image` by hand before `cluster:install`), or naming one template `skydiscover` so at least one preset builds through the standard path. (`packages/agents/skydiscover/README.md:71`)
- 🟢 **Suggestion:** The worked example passes a bare relative `-o output`, contradicting the same file's own CLI-reference rule ("**always** an explicit path on `$SKYDISCOVER_OUTPUT_ROOT`, outside the target repo") and `AGENTS.md`'s "Always pass an explicit `-o` on the persisted workspace." Sibling skills have the same soft spot, but an agent copying this block verbatim writes checkpoints into the CWD, which hibernation can lose. (`packages/agents/skydiscover/workspace/.agents/skills/skydiscover/SKILL.md:227`)
- ✅ **Looks good:** `resolve-image.sh` registration is complete and symmetric — `WORKLOADS` entry drives `own_paths`/`base_of`/the CI matrix through `is_workload`, and the self-check adds all four assertions (base, both effective-path halves, the workloads-JSON expectation) rather than just the list.
- ✅ **Looks good:** The Dockerfile mirrors the gepa/shinkaevolve pattern exactly where it matters for correctness: `chown -R 65532:0` on both the venv and `/opt/uv` (uid 65532 runs at runtime), `uv cache clean` before the layer closes, `SKYDISCOVER_OUTPUT_ROOT` under the seeded `/app/working-dir`, `COPY --chown` for the workspace, and `USER 65532` last.
- ✅ **Looks good:** The `SKYDISCOVER_SEARCH` env plumbing is real, not aspirational — `agent-templates.yaml:102-109` renders `$tmpl.env` into the template ConfigMap, so both presets genuinely differ at reconcile time; the Dockerfile's `ENV SKYDISCOVER_SEARCH=adaevolve` is a sane fallback for a direct image run.
- ✅ **Looks good:** `AGENTS.md` gets the two genuinely load-bearing pod details right: the `run.pid` liveness check validates `/proc/<pid>/cmdline` against `skydiscover-run` (a bare `kill -0` would misread a recycled PID after a restart), and the credential story correctly frames the GitHub connection's scope as the control surface since injection is host-keyed and reaches LLM-generated candidate code too.

aligned

### Verdict
COMMENT — Registration, image, and env plumbing are correct and idiomatic; the one substantive issue is that neither preset name maps to a local build task, so the README's documented local-enable path silently builds nothing.

---

packages/agents/skydiscover/AGENTS.md
```diff
@@ -0,0 +1,276 @@
...
+
+  ```sh
+  # run this script AS a backgrounded harness task (your Bash tool's
+  # run_in_background) — NEVER as a foreground command: the tool's timeout
+  # would SIGTERM the whole process group, run included, mid-flight
+  dir="$SKYDISCOVER_OUTPUT_ROOT/<run-id>"
+  cd "$dir"
+  export OPENAI_API_KEY="${OPENAI_API_KEY:-placeholder}"   # gateway overwrites it on the wire
+  skydiscover-run task/initial.py task/evaluator.py \
+    --search "$SKYDISCOVER_SEARCH" \
```
dam-code-guardian: 🟡 **Warning:** Dropping the baked `SKYDISCOVER_SEARCH` default (correctly — I verified `mergedSpawnEnv` in `agent-runtime/src/core/runtime-env.ts` spreads `process.env` last, so a container value really would shadow the preset) leaves this example without the fallback that replaced it. On a preset-less run the variable is unset, so this expands to `--search ""` — an empty strategy, not the `adaevolve` this file promises 100 lines above ("If `$SKYDISCOVER_SEARCH` is unset … default to `adaevolve`").

The idiom is already in this very snippet two lines up for `OPENAI_API_KEY`, so applying it here keeps prose and example agreeing:

```suggestion
    --search "${SKYDISCOVER_SEARCH:-adaevolve}" \
```

Same line in `workspace/.agents/skills/skydiscover/SKILL.md:300`, whose own CLI table already documents `unset → adaevolve`.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `cd400e05`

## PR #3109: feat(agents): package SkyDiscover AdaEvolve & EvoX workload image
**Author:** xjacka | **Branch:** feat/skydiscover-workload-image → main | **Changes:** +872 −11 (10 files)

### Summary
Packages SkyDiscover's AdaEvolve and EvoX strategies as one curated workload image behind two template presets. Sixteen commits since the last review fixed all three prior findings and added substantial field-tested pod policy (sampling-param and EvoX auxiliary-model quirks, error triage by class, a moonlighting guardrail, the background-work launch contract).

### Changes since last review
Previous HEAD: `ca3bc7ba` (2026-07-31T10:45:16Z) — verdict COMMENT

- ✅ **Fixed:** Neither preset name mapped to a local build task — `cluster:install`/`cluster:build-agent` now resolve the component from the image repository basename and dedup, so both presets build the one image once (`deploy/tasks.toml:507`)
- ✅ **Fixed:** README's local-build note now states how the resolution works instead of inheriting the sibling wording (`packages/agents/skydiscover/README.md:78`)
- ✅ **Fixed:** The worked example passes an explicit `-o "$SKYDISCOVER_OUTPUT_ROOT/…"` (`packages/agents/skydiscover/workspace/.agents/skills/skydiscover/SKILL.md:302`)
- 🆕 **New:** Both launch examples pass `--search "$SKYDISCOVER_SEARCH"` unguarded, so a preset-less run sends an empty `--search` instead of the documented `adaevolve` fallback (`packages/agents/skydiscover/AGENTS.md:146`)

### Findings
- 🆕 **Warning:** Unguarded `--search "$SKYDISCOVER_SEARCH"` contradicts the documented unset-fallback (`packages/agents/skydiscover/AGENTS.md:146`, `workspace/.agents/skills/skydiscover/SKILL.md:300`)

aligned

### Verdict
COMMENT — All three prior findings are genuinely fixed, the build-task fix generalizes further than suggested, and the new pod policy is high-quality; the one new issue is a one-token shell gap opened by dropping the baked default.

---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `59979af9`

## PR #3109: feat(agents): package SkyDiscover AdaEvolve & EvoX workload image
**Author:** xjacka | **Branch:** feat/skydiscover-workload-image → main | **Changes:** +872 −11 (10 files)

### Summary
Re-review at `59979af9`: one commit since the last review, fixing the open 🟡 and volunteering a second correctness fix to the same snippet.

### Changes since last review
Previous HEAD: `cd400e05` (2026-08-04T09:59:56Z) — verdict COMMENT

- ✅ **Fixed:** Unguarded `--search "$SKYDISCOVER_SEARCH"` now `"${SKYDISCOVER_SEARCH:-adaevolve}"` in both launch examples, matching the unset-fallback the same files document at `AGENTS.md:35` and `SKILL.md:217` (`packages/agents/skydiscover/AGENTS.md:146`, `workspace/.agents/skills/skydiscover/SKILL.md:300`)

### Findings

Not flagged, worth recording: the commit also fixed something I had not raised — `echo $! > run.pid; wait` became `pid=$!; echo "$pid" > run.pid; wait "$pid"`. That is a real fix, and it is the same class of defect I flagged on the sibling PR's three snippets, so I verified it in a real script file rather than by reading: with `set -euo pipefail`, `/bin/false &` followed by `wait "$pid"` exits **1**, while bare `wait` exits **0**. The harness background task now genuinely observes a failed run instead of reporting success — which matters precisely because the background-work contract this section mandates keys the follow-up turn off that task's completion.

I also re-checked the two guards the earlier reviews leaned on, since a fix in this area could have loosened them: the start-of-conversation liveness check still validates `/proc/$pid/cmdline` against `skydiscover-run` (`AGENTS.md:69`), and `SKILL.md:217` still names `adaevolve` as the unset default — so the fallback the fix introduces is the one the reference table promises, not a new third convention.

aligned

The delta is two shell-snippet lines in package-level docs. No page under `docs/architecture/` is touched or contradicted, no new long-lived component is introduced, and no ADR changed state.

### Verdict
APPROVE — the last open finding is fixed exactly as the file's own reference table specifies, the volunteered `wait "$pid"` fix is verified correct in a real shell, and nothing new surfaced.

---

packages/agents/skydiscover/workspace/.agents/skills/skydiscover/SKILL.md
```diff
@@ -0,0 +1,313 @@
...
+| `INITIAL_PROGRAM` | required — the starting program (`EVOLVE-BLOCK` markers; a minimal stub for from-scratch problems) |
+| `EVALUATOR` | required — the `evaluate(program_path)` file |
+| `--search` | strategy — default to `$SKYDISCOVER_SEARCH` (`adaevolve` or `evox`; unset → `adaevolve`); only those two work in this pod |
+| `-i, --iterations` | the run's iteration budget — always bound; on resume, set to the *remainder* of the approved total |
+| `-m, --model` | plain model id served by the endpoint (Step 1); an ensemble goes in the config YAML instead |
+| `--api-base` | the OpenAI-compatible endpoint (`"$base/v1"`) — **always pass it** in this pod |
+| `-o, --output` | output dir — **always** an explicit path on `$SKYDISCOVER_OUTPUT_ROOT`, outside the target repo |
+| `-c, --config` | YAML config (model ensemble, `search.*` tuning) — flags win over it |
+| `--checkpoint` | resume from a prior `output/checkpoints/checkpoint_<N>` dir |
+| `--codebase` | path to a codebase dir — enables agentic generation (the LLM reads/searches it before writing code); costlier per iteration, only with the user's explicit OK |
```
PetrBulanek: `--codebase` isn't a real flag. The CLI at the pinned ref — and on upstream `main`, and since the initial release — has `--agentic`. Passing this fails outright:

`skydiscover-run: error: unrecognized arguments: --codebase /work/task`

And `--agentic` doesn't work in this image either, because the package ships without its tool schemas:

`Error: [Errno 2] No such file or directory: '.../skydiscover/llm/tool_schemas/agentic_tools.json'`

`[tool.setuptools.package-data]` omits `llm/tool_schemas/*.json` on `main` as well, so bumping `SKYDISCOVER_REF` won't fix it. (For a custom root the real knob is the config key `agentic.codebase_root` via `-c` — likely where this row originated.)

Suggest dropping this row and the `[--codebase <dir>]` in the usage block at L210, and stating that agentic mode is unsupported in this image. Optionally add `llm/tool_schemas/agentic_tools.json` to the Dockerfile's build assertion so a future pin bump surfaces it automatically.

packages/agents/skydiscover/workspace/.agents/skills/skydiscover/SKILL.md
```diff
@@ -0,0 +1,313 @@
...
+- **Export a non-empty key before launching**:
+  `export OPENAI_API_KEY="${OPENAI_API_KEY:-placeholder}"`. The injected value
+  is often an empty placeholder by design — the gateway overwrites the auth
+  header on the wire, but the client refuses to send a request with no key at
+  all.
+- If the endpoint can't list models, fall back to a pinned known-good id (for
+  IBM LiteLLM: `aws/claude-sonnet-4-6` — but see the sampling-params quirk
+  below before using it for the loop).
+
+**Known quirk — sampling params:** SkyDiscover's LLM client sends both
```
PetrBulanek: This doesn't hold for the pinned ref. I captured the wire from a real AdaEvolve run in this image against a logging endpoint — every request:

`{"keys": ["max_tokens","messages","model","temperature"], "top_p": "<absent>"}`

At `e221cde`, `top_p` defaults to `None` and `llm/openai.py` only sets it `if top_p is not None`; the shipped evox `search.yaml` has it commented out citing this exact provider conflict. The pinned commit *is* upstream's fix for it ("Fix default top_p to avoid Anthropic/Bedrock parameter conflict").

The 400s you saw were presumably real, but the mechanism isn't — which costs three ways: the pre-launch gate mandates a probe for a condition that can't occur, the per-model denylist rests on a false premise, and the triage rule at L241 hands the agent the wrong hypothesis for any 400 it *does* hit (reasoning models rejecting `temperature` outright is the live candidate at gpt-5.x). Worth re-diagnosing rather than deleting.

packages/agents/skydiscover/workspace/.agents/skills/skydiscover/SKILL.md
```diff
@@ -0,0 +1,313 @@
...
+import importlib.util
+import math
+
+def evaluate(program_path):
+    spec = importlib.util.spec_from_file_location("candidate", program_path)
+    mod = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(mod)
+    xs = [i * math.pi / 50 for i in range(51)]
+    mse = sum((mod.approx(x) - math.sin(x)) ** 2 for x in xs) / len(xs)
+    return {"combined_score": 1.0 / (1.0 + mse)}
```
PetrBulanek: Ran this in the image with a model that proposes `import math; return math.sin(x)` — the run scores it `combined_score: 1.0000` and writes it to `best/best_program.py`.

That's verbatim the failure described at L169-171 ("will happily crown `return math.sin(x)`"), and `1/(1+MSE)` is the saturating form banned at L163. The example is the only runnable artifact in the doc, and agents copy runnable code over prose — so it currently teaches the opposite of the guard section.

Either guard it (hard-reject `math.sin`/`import math`, log-scaled score, plus the cheat case the smoke-eval is required to include), or keep it naive and relabel it as the deliberately-broken baseline the guard section repairs — which teaches better.

Separately, L269 calls it "the CI/local smoke fixture", but nothing in `.github/workflows/` runs it.

packages/agents/skydiscover/workspace/.agents/skills/skydiscover/SKILL.md
```diff
@@ -0,0 +1,313 @@
...
+      weight: 0.7
+    - name: "<strong-model-id>"
+      weight: 0.3
+```
+
+Pass it with `-c task/config.yaml`.
+
+## Step 2 — author the run inputs
+
+A run needs an **evaluator** and an **initial program** — both are required
```
PetrBulanek: Help shows `[initial_program] evaluation_file`, and an evaluator-only run works (`Runner ready: search=adaevolve, program=None`) — it takes upstream's from-scratch prompt path instead of the evolve-the-block one.

The stub advice itself is fine and probably the right default for this pod; just don't state it as a CLI constraint. Same wording at L215.

packages/agents/skydiscover/README.md
```diff
@@ -0,0 +1,94 @@
...
+the harness scripts.
+
+How the pod reaches models without holding keys (plain model ids +
+`--api-base`, the placeholder-key idiom) is defined once in the `skydiscover`
+skill's Step 1.
+
+## Build
+
+```sh
+mise run agents:skydiscover:image            # plain docker build (pip-installs skydiscover)
```
PetrBulanek: Stale since the pin landed: the task is `build-or-reuse` (pulls from the registry when the effective source is unchanged), not a plain `docker build`, and it installs from a pinned git ref rather than pip.

The PR description still says `skydiscover==0.1.0` too — worth updating so reviewers don't chase the wheel.

PetrBulanek (changes requested):

PetrBulanek (approved):

packages/agents/skydiscover/workspace/.agents/skills/skydiscover/SKILL.md
```diff
@@ -0,0 +1,339 @@
...
+
+def evaluate(program_path):
+    src = open(program_path).read()
+    # hard gate: pure arithmetic only — no imports, attributes, or calls,
+    # or the winning move is simply `return math.sin(x)`
+    for node in ast.walk(ast.parse(src)):
+        if isinstance(node, (ast.Import, ast.ImportFrom, ast.Attribute, ast.Call)):
+            return {"combined_score": 0.0, "error": "not a pure polynomial"}
+    ns = {}
+    exec(compile(src, program_path, "exec"), {"__builtins__": {}}, ns)
```
dam-code-guardian: 🟡 **Warning:** Passing separate `globals` and `locals` dicts makes module-level names invisible inside function bodies, so any candidate that hoists a constant dies with `NameError`:

```
C3 = 1.0 / 6.0
def approx(x):
    return x - C3 * x**3      # NameError: name 'C3' is not defined
```

Verified in a real file, not an inline subshell: the same candidate scores `0.086` under a single-namespace `exec` and raises under this one. Hoisting coefficients is a *normal* move for an evolutionary search on a polynomial-fitting objective, so this is not an edge case — and the failure is silent: `skydiscover/evaluation/evaluator.py` catches the exception and returns `EvaluationResult(metrics={"error": 0.0})` after its retries, which is indistinguishable from the gate's own `combined_score: 0.0` cheat rejection. The whole named-constant branch of the search space is discarded while the log shows only a zero.

Execing into one namespace fixes it, and keeps `__builtins__` stubbed and `ns["approx"]` below unchanged — verified, the hoisted candidate then scores `0.086`:

```suggestion
    ns = {"__builtins__": {}}
    exec(compile(src, program_path, "exec"), ns)
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `ec04c6c4`

## PR #3109: feat(agents): package SkyDiscover AdaEvolve & EvoX workload image
**Author:** xjacka | **Branch:** feat/skydiscover-workload-image → main | **Changes:** +906 −11 (10 files)

### Summary
Re-review at `ec04c6c4`: two docs-only commits. They answer every point of the human `CHANGES_REQUESTED` round — I re-verified all five against upstream at the pinned ref `e221cde` rather than taking them on trust: `initial_program` is `nargs="?"`, `--codebase` does not exist (`--agentic` does) while `llm/tool_schemas/*.json` is still absent from `[tool.setuptools.package-data]`, `top_p` is now sent only `if top_p is not None` over a `None` default, and `agents:skydiscover:image` really is `build-or-reuse`. The rewritten worked-example evaluator introduces one silent-scoring bug.

### Changes since last review
Previous HEAD: `59979af9` (2026-08-04T11:14:28Z) — verdict APPROVE

- 🆕 **New:** The new evaluator's split-namespace `exec` hides module-level names from `approx`, so hoisted-constant candidates score 0 (`packages/agents/skydiscover/workspace/.agents/skills/skydiscover/SKILL.md:302`)
- 🆕 **New:** Step 3's smoke-eval snippet omits the cheat case that L177 explicitly requires of it (`packages/agents/skydiscover/workspace/.agents/skills/skydiscover/SKILL.md:200`)

### Findings
- 🟡 **Warning:** Split-namespace `exec` silently zero-scores every candidate that hoists a constant (`packages/agents/skydiscover/workspace/.agents/skills/skydiscover/SKILL.md:302`)
- 🟡 **Warning:** L177 mandates "The Step 3 smoke-eval **must** include at least one cheat case", but the Step 3 snippet at L200 is still baseline-only — the worked example gained the cheat case, the normative step didn't. An agent copies the runnable line it is told to run, so the guard that L163–177 exists to install is the one most likely skipped. (`packages/agents/skydiscover/workspace/.agents/skills/skydiscover/SKILL.md:200`)

Not flagged, recorded: the re-diagnosed triage rule is right, and narrower than stated. Upstream's `is_openai_reasoning_model` strips `temperature` for reasoning models — but only when `api_base` is `api.openai.com`/`*.openai.azure.com` (`llm/openai.py:40-46`). Behind the IBM LiteLLM proxy this pod uses, that guard returns `False`, so `temperature` *is* sent to a `gpt-5.x` reasoning model. The doc's "live candidate" is therefore not just plausible, it is the expected outcome on this endpoint. Also verified the inverted Dockerfile canary's premise holds at the pin (the tree ships `llm/tool_schemas/agentic_tools.json`, `package-data` doesn't select it, so the assert passes today and trips exactly when packaging is fixed).

aligned

### Verdict
COMMENT — the human round is genuinely and verifiably addressed, but the new evaluator mis-scores valid candidates in a way that looks identical to a cheat rejection, so it teaches the wrong lesson from the file's only runnable artifact.

---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `d91e6be5`

## PR #3109: feat(agents): package SkyDiscover AdaEvolve & EvoX workload image
**Author:** xjacka | **Branch:** feat/skydiscover-workload-image → main | **Changes:** +908 −11 (10 files)

### Summary
Re-review at `d91e6be5`: one docs commit fixing the split-namespace `exec` in the worked-example evaluator. Verified by execution, not reading — the hoisted-constant candidate that raised `NameError` under the old shape now scores `0.0439`, while the cheat case still hard-rejects at `0.0`.

### Changes since last review
Previous HEAD: `ec04c6c4` (2026-08-05T07:27:25Z) — verdict COMMENT

- ✅ **Fixed:** Split-namespace `exec` zero-scoring hoisted-constant candidates — single namespace now, and the added comment states the mechanism correctly (`packages/agents/skydiscover/workspace/.agents/skills/skydiscover/SKILL.md:304`)
- 🔁 **Still present:** L177's "the Step 3 smoke-eval **must** include at least one cheat case" is still unmet by the Step 3 snippet at L200 (`packages/agents/skydiscover/workspace/.agents/skills/skydiscover/SKILL.md:200`)

### Findings

Not flagged, recorded — three checks the fix invited, all clean. The fix moves `{"__builtins__": {}}` into `ns` rather than dropping it, so the AST gate keeps its teeth: a `range()` candidate is still rejected as "not a pure polynomial", and `ns` gaining a `__builtins__` key is inert because the caller reads only `ns["approx"]`. The fix is also genuinely local — `exec(compile(` appears exactly once across `packages/agents`, so there is no sibling image carrying the same pattern to port it to. And the evaluator's uncaught-exception paths (`OverflowError`, `ZeroDivisionError`, arity mismatch) are unchanged by this commit and were never in scope of the finding.

aligned

The delta is a 4-line edit to a package-level skill doc. No page under `docs/architecture/` describes this workload image or enumerates curated images, no architecture page is touched, no new long-lived component is introduced, and no ADR changed state.

### Verdict
APPROVE — the flagged bug is fixed, verified by running both evaluator shapes against four candidate classes; the one carryover is a pre-existing 🟡 in normative prose, not a regression.

---

PR #3110, opened 2026-07-31
fix(ui): bind deep links survive login and the Terms gate

### LINKED ISSUE #3107: Sidebar DM: logged-out bind link redirects to dashboard (not agent selection) and expires after login
## Summary

When a user is **not logged in** and clicks the `Connect one of your agents to this DM.` link in the sidebar DM, the post-login redirect is broken: instead of landing on the agent-selection screen, they end up on the dashboard. Re-clicking the original link then reports it as **expired**, forcing the user to run `/dam bind` again.

## Steps to reproduce

1. Be logged out.
2. In a sidebar DM, click the `Connect one of your agents to this DM.` link.
3. You are sent to the login page — log in.
4. Observe where you land after login.
5. Go back and click the original link again.

## Expected

After logging in, the user should be redirected to the **agent selection** screen (the intended destination of the link), so they can complete the bind in one flow.

## Actual

- After login, the user is redirected to the **dashboard**, not agent selection.
- Clicking the original link again shows an **expired** error.
- The user has to re-run `/dam bind` to get a fresh link and try again.

## Notes

The link appears to be single-use / short-lived, and the login flow drops the original `redirect` (deep-link) target. The post-login redirect should preserve and resume the bind deep-link.

Reported by @matoushavlena in the team Slack.

pilartomas:
Fixes #3107.

## The bug

Following a bind link from a Slack DM while logged out never reached the agent
picker. Two things dropped the destination:

1. **The post-login render used a stale route.** The navigation store derives its
   route when its module loads — which, on the way back from Keycloak, is while
   the browser is still on the OIDC callback URL. The callback then rewrites the
   URL to the parked return path, but the store has already resolved the route to
   the dashboard. A mount-time re-parse corrected it a frame later, so the bind
   picker did appear — behind a dashboard flash, and only if nothing else
   intervened first.
2. **The Terms gate replaced the URL outright.** For anyone who hadn't accepted
   the current Terms (the common case for a first visit to the web UI, which is
   exactly who follows a bind link), the gate swapped the URL for the Terms page
   and acceptance always continued to the dashboard. The deep link was gone —
   and with the link's one-shot OAuth state already spent by the login, clicking
   it again reports the link expired, so the only way forward is another
   `bind` command. That is the reported symptom.

The bind flow behind `?flow=` itself is fine: it lives well within its TTL and is
re-readable, so preserving the URL is all that is needed to finish in one pass.

## The fix

One primitive owns "where was the user headed" for both interstitials — the login
roundtrip and the Terms gate — with a slot each, so neither clobbers the other:

- The login callback and the Terms gate park the destination and resume it once
  cleared; Terms acceptance returns there instead of always to the dashboard.
- The route is re-derived once the interstitials have settled and before the first
  render, so the first paint is the deep link — no dashboard flash, and no
  dashboard shell mounting queries on the way past.
- A failed silent renew now parks the destination too; previously that re-login
  always came back to the dashboard.
- Destinations are validated on the way out: same-origin paths only (a stored
  value is attacker-reachable), and never an interstitial path — returning to the
  callback would replay a spent code, returning to the Terms page would re-enter
  the gate.

Deliberately unchanged: the OAuth state behind the link stays single-use. Deleting
it on first use is the replay control, and with the redirect fixed there is no
reason to click the link twice.

## Verification

- New unit tests cover the destination round-trip, slot separation, one-shot
  semantics, and the rejection rules.
- New smoke spec: follow a bind deep link while logged out, log in, clear the
  Terms gate, and assert the picker renders with the flow id still in the URL. It
  runs first in the `auth` project, so on a fresh install it faces the Terms leg —
  the one that made this stick.
- `mise run test` (all packages) and the UI + Playwright type/lint/format checks
  pass. `mise run check` also passes except the controller CRD-compat task, which
  cannot resolve git refs from a worktree — no CRDs are touched here.

Manual check worth doing on a cluster before merge: from a Slack DM, run the bind
command as a user who has not accepted the current Terms, follow the link in a
signed-out browser, and confirm the picker appears after acceptance.

### DIFF
--- a/packages/ui/src/lib/return-path.ts
+++ b/packages/ui/src/lib/return-path.ts
@@ -33,9 +33,12 @@ export interface ReturnPathStore {
 
 /** Stored values are untrusted on the way out: only same-origin paths pass. */
 export function isSafeReturnPath(value: string): boolean {
+  // Browsers drop tab/LF/CR while parsing, so "/<tab>/host" is protocol-relative
+  // after the fact — validate what the browser will act on, not the raw text.
+  const parsed = value.replace(/[\t\n\r]/g, "");
   // "//host" and "/\host" are protocol-relative URLs, not paths.
-  if (!value.startsWith("/") || /^\/[/\\]/.test(value)) return false;
-  return !INTERSTITIAL_PATHS.includes(value.split(/[?#]/)[0]!);
+  if (!parsed.startsWith("/") || /^\/[/\\]/.test(parsed)) return false;
+  return !INTERSTITIAL_PATHS.includes(parsed.split(/[?#]/)[0]!);
 }
 
 /** The location as a return path, or null when it isn't a destination. */
--- a/packages/ui/src/lib/return-path.ts
+++ b/packages/ui/src/lib/return-path.ts
@@ -33,9 +33,12 @@ export interface ReturnPathStore {
 
 /** Stored values are untrusted on the way out: only same-origin paths pass. */
 export function isSafeReturnPath(value: string): boolean {
-  // Browsers drop tab/LF/CR while parsing, so "/<tab>/host" is protocol-relative
-  // after the fact — validate what the browser will act on, not the raw text.
-  const parsed = value.replace(/[\t\n\r]/g, "");
+  // Validate what the browser will act on, not the raw text: it drops tab/LF/CR
+  // anywhere while parsing ("/<tab>/host" ends up protocol-relative) and trims
+  // C0 controls and spaces at the ends ("/terms<FF>" ends up on the gate).
+  // Stripping the whole class judges a little more than the parser removes,
+  // which only ever rejects — the safe direction.
+  const parsed = value.replace(/[\0-\x20]/g, "");
   // "//host" and "/\host" are protocol-relative URLs, not paths.
   if (!parsed.startsWith("/") || /^\/[/\\]/.test(parsed)) return false;
   return !INTERSTITIAL_PATHS.includes(parsed.split(/[?#]/)[0]!);
--- a/packages/ui/src/lib/return-path.ts
+++ b/packages/ui/src/lib/return-path.ts
@@ -1,0 +1,69 @@
+/** Where the user was headed before an interstitial took the browser off it.
+ *
+ *  Two interstitials rewrite the URL after boot — the Keycloak login roundtrip
+ *  and the Terms-of-Use gate — and both have to hand the user back afterwards.
+ *  Deep links are why it matters: a Slack/Telegram bind link carries a one-shot
+ *  `?flow=`, so losing the target costs the user another bind command. */
+
+/** One slot per interstitial, so a login inside the Terms gate (or the reverse)
+ *  can't clobber the other's destination. */
+const SLOTS = {
+  login: "platform-auth-return",
+  terms: "platform-terms-return",
+} as const;
+
+export type Interstitial = keyof typeof SLOTS;
+
+/** Interstitials are never destinations: returning to one re-enters the gate or
+ *  replays a spent OIDC code. */
+const INTERSTITIAL_PATHS = ["/auth/callback", "/terms"];
+
+export interface LocationLike {
+  pathname: string;
+  search: string;
+  hash: string;
+}
+
+/** The sessionStorage surface used here — a narrow shape keeps it substitutable. */
+export interface ReturnPathStore {
+  getItem(key: string): string | null;
+  setItem(key: string, value: string): void;
+  removeItem(key: string): void;
+}
+
+/** Stored values are untrusted on the way out: only same-origin paths pass. */
+export function isSafeReturnPath(value: string): boolean {
+  // "//host" and "/\host" are protocol-relative URLs, not paths.
+  if (!value.startsWith("/") || /^\/[/\\]/.test(value)) return false;
+  return !INTERSTITIAL_PATHS.includes(value.split(/[?#]/)[0]!);
+}
+
+/** The location as a return path, or null when it isn't a destination. */
+export function toReturnPath(loc: LocationLike): string | null {
+  const path = `${loc.pathname}${loc.search}${loc.hash}`;
+  return isSafeReturnPath(path) ? path : null;
+}
+
+export function rememberReturnPath(
+  which: Interstitial,
+  loc: LocationLike = window.location,
+  store: ReturnPathStore = sessionStorage,
+): void {
+  const path = toReturnPath(loc);
+  // Clearing on a non-destination stops a stale target resurfacing later.
+  if (path) store.setItem(SLOTS[which], path);
+  else store.removeItem(SLOTS[which]);
+}
+
+/** Reads and clears the stashed destination — one-shot, dashboard as fallback. */
+export function takeReturnPath(
+  which: Interstitial,
+  store: ReturnPathStore = sessionStorage,
+): string {
+  const stashed = store.getItem(SLOTS[which]);
+  store.removeItem(SLOTS[which]);
+  if (!stashed) return "/";
+  if (isSafeReturnPath(stashed)) return stashed;
+  console.warn(`[return-path] ignoring unusable ${which} return:`, stashed);
+  return "/";
+}
--- a/packages/ui/src/lib/return-path.ts
+++ b/packages/ui/src/lib/return-path.ts
@@ -19,6 +19,7 @@ export type Interstitial = keyof typeof SLOTS;
 const INTERSTITIAL_PATHS = ["/auth/callback", "/terms"];
 
 export interface LocationLike {
+  origin: string;
   pathname: string;
   search: string;
   hash: string;
@@ -31,23 +32,61 @@ export interface ReturnPathStore {
   removeItem(key: string): void;
 }
 
-/** Stored values are untrusted on the way out: only same-origin paths pass. */
-export function isSafeReturnPath(value: string): boolean {
-  // Validate what the browser will act on, not the raw text: it drops tab/LF/CR
-  // anywhere while parsing ("/<tab>/host" ends up protocol-relative) and trims
-  // C0 controls and spaces at the ends ("/terms<FF>" ends up on the gate).
-  // Stripping the whole class judges a little more than the parser removes,
-  // which only ever rejects — the safe direction.
-  const parsed = value.replace(/[\0-\x20]/g, "");
-  // "//host" and "/\host" are protocol-relative URLs, not paths.
-  if (!parsed.startsWith("/") || /^\/[/\\]/.test(parsed)) return false;
-  return !INTERSTITIAL_PATHS.includes(parsed.split(/[?#]/)[0]!);
+/**
+ * Resolve an untrusted destination the way the browser will, or null when it
+ * must not be navigated to.
+ *
+ * The URL parser is the authority, and everything below hands back *its* output
+ * rather than the caller's text. That is the point: the parser drops control
+ * characters, trims C0 and spaces, collapses dot segments and percent-encodes,
+ * so any check against the raw string judges something the browser has already
+ * rewritten — which is how `/<tab>/host` reached another origin and
+ * `/a/../terms` reached the gate. There is no verdict-only entry point for the
+ * same reason: validating one string and navigating to another is the bug.
+ */
+function resolve(value: string, origin: string): URL | null {
+  // Every parse sits inside the guard: this runs at boot and on the way out of
+  // the Terms gate, so an unparseable value has to read as "no destination"
+  // rather than throw into either path.
+  try {
+    const base = new URL(origin);
+    const url = new URL(value, base);
+    if (url.origin !== base.origin) return null;
+    // Schemes without a hierarchical path (blob:, data:) can still report our
+    // origin, and their "pathname" is not a path.
+    if (!url.pathname.startsWith("/")) return null;
+    if (INTERSTITIAL_PATHS.includes(url.pathname)) return null;
+    // The output has to survive being used as a reference again, because that is
+    // what callers do with it. A resolved path can begin with "//" — dot segments
+    // collapse "/a/..//host" to "//host" — which reads as another origin the
+    // second time around, or fails to parse at all. Re-resolving proves the
+    // verdict instead of arguing it.
+    return new URL(pathOf(url), base).href === url.href ? url : null;
+  } catch {
+    return null;
+  }
 }
 
-/** The location as a return path, or null when it isn't a destination. */
-export function toReturnPath(loc: LocationLike): string | null {
-  const path = `${loc.pathname}${loc.search}${loc.hash}`;
-  return isSafeReturnPath(path) ? path : null;
+function pathOf(url: URL): string {
+  return `${url.pathname}${url.search}${url.hash}`;
+}
+
+/** The whole destination — path, query and fragment — canonicalized. */
+export function resolveReturnPath(
+  value: string,
+  origin: string,
+): string | null {
+  const url = resolve(value, origin);
+  return url ? pathOf(url) : null;
+}
+
+/** The path alone, for routing: a query or fragment riding along in a stored
+ *  value would otherwise be read as part of an id or tab. */
+export function resolveReturnPathname(
+  value: string,
+  origin: string,
+): string | null {
+  return resolve(value, origin)?.pathname ?? null;
 }
 
 export function rememberReturnPath(
@@ -55,21 +94,28 @@ export function rememberReturnPath(
   loc: LocationLike = window.location,
   store: ReturnPathStore = sessionStorage,
 ): void {
-  const path = toReturnPath(loc);
+  const path = resolveReturnPath(
+    `${loc.pathname}${loc.search}${loc.hash}`,
+    loc.origin,
+  );
   // Clearing on a non-destination stops a stale target resurfacing later.
   if (path) store.setItem(SLOTS[which], path);
   else store.removeItem(SLOTS[which]);
 }
 
-/** Reads and clears the stashed destination — one-shot, dashboard as fallback. */
+/** Reads and clears the stashed destination — one-shot, dashboard as fallback.
+ *  Re-resolved on the way out: the slot is attacker-writable, so what the
+ *  caller navigates to is the parser's verdict, never the stored text. */
 export function takeReturnPath(
   which: Interstitial,
   store: ReturnPathStore = sessionStorage,
+  origin: string = window.location.origin,
 ): string {
   const stashed = store.getItem(SLOTS[which]);
   store.removeItem(SLOTS[which]);
   if (!stashed) return "/";
-  if (isSafeReturnPath(stashed)) return stashed;
+  const path = resolveReturnPath(stashed, origin);
+  if (path) return path;
   console.warn(`[return-path] ignoring unusable ${which} return:`, stashed);
   return "/";
 }
--- a/packages/ui/src/main.tsx
+++ b/packages/ui/src/main.tsx
@@ -10,6 +10,7 @@ import { TooltipProvider } from "@/components/ui/tooltip";
 
 import { initAuth } from "./auth.js";
 import { applyBrand, loadBrand } from "./brand.js";
+import { rememberReturnPath } from "./lib/return-path.js";
 import { routeToPath } from "./modules/platform/lib/routes.js";
 import { preflightTermsGate } from "./modules/terms/lib/preflight.js";
 import { queryClient } from "./query-client.js";
@@ -23,9 +24,16 @@ async function main() {
   if (!user) return; // Redirecting to Keycloak, don't render
 
   if (!(await preflightTermsGate())) {
+    // Park the destination so acceptance resumes it — a bind link's one-shot
+    // `?flow=` is spent by then, and dropping it costs another bind command.
+    rememberReturnPath("terms");
     window.history.replaceState({}, "", routeToPath({ view: "terms" }));
-    useStore.setState({ view: "terms" });
   }
+
+  // Both interstitials above rewrite the URL after the store derived its route,
+  // so re-derive it here: without this the first render is the pre-redirect
+  // view (the dashboard) rather than the deep link the user followed.
+  useStore.getState().hydrateRoute();
 
   const { default: App } = await import("./app.js");
   createRoot(document.getElementById("root")!).render(
--- a/docs/architecture/security-and-credentials.md
+++ b/docs/architecture/security-and-credentials.md
@@ -1,6 +1,6 @@
 # Security and credentials
 
-Last verified: 2026-07-30
+Last verified: 2026-07-31
 
 ## Overview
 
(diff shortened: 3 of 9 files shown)


packages/ui/src/lib/return-path.ts
```diff
@@ -0,0 +1,69 @@
...
+export interface ReturnPathStore {
+  getItem(key: string): string | null;
+  setItem(key: string, value: string): void;
+  removeItem(key: string): void;
+}
+
+/** Stored values are untrusted on the way out: only same-origin paths pass. */
+export function isSafeReturnPath(value: string): boolean {
+  // "//host" and "/\host" are protocol-relative URLs, not paths.
+  if (!value.startsWith("/") || /^\/[/\\]/.test(value)) return false;
```
dam-code-guardian: 🟡 **Warning:** the same-origin check is bypassable by a leading control character. `"/" + TAB + "/evil.example"` passes both guards — `startsWith("/")` is true and `/^\/[/\\]/` doesn't match, because the tab sits between the slashes — but browsers strip tab/LF/CR when parsing a URL, so `new URL(that, origin)` resolves to `https://evil.example/`. Verified against Node's WHATWG URL parser: tab, LF and CR all escape the origin, while the plain `//host` and `/\host` forms are correctly rejected.

This is reachable exactly where the module's own doc comment says the value is untrusted: `takeReturnPath` passes stored text to `window.location.assign` (`terms-view.tsx:56`), so an attacker who can write the sessionStorage slot gets an open redirect out of a same-origin-looking path. Stripping the characters before validating closes it, and also fixes the interstitial check, which the same trick evades (`"/terms" + TAB` is not in `INTERSTITIAL_PATHS`).

```suggestion
export function isSafeReturnPath(value: string): boolean {
  // Browsers strip tab/LF/CR while parsing, so "/\t/host" resolves
  // cross-origin: drop them before the protocol-relative check.
  const stripped = value.replace(/[\t\n\r]/g, "");
  // "//host" and "/\\host" are protocol-relative URLs, not paths.
  if (!stripped.startsWith("/") || /^\/[/\\]/.test(stripped)) return false;
  return !INTERSTITIAL_PATHS.includes(stripped.split(/[?#]/)[0]!);
}
```

packages/ui/src/main.tsx
```diff
@@ -23,10 +24,17 @@ async function main() {
...
+    // `?flow=` is spent by then, and dropping it costs another bind command.
+    rememberReturnPath("terms");
     window.history.replaceState({}, "", routeToPath({ view: "terms" }));
-    useStore.setState({ view: "terms" });
   }
 
+  // Both interstitials above rewrite the URL after the store derived its route,
+  // so re-derive it here: without this the first render is the pre-redirect
+  // view (the dashboard) rather than the deep link the user followed.
+  useStore.getState().hydrateRoute();
```
dam-code-guardian, line 36: 🟢 **Suggestion:** `hydrateRoute` sets only the four fields `routeToNavigationState` owns, and that helper returns `agentId: null` for chat routes — chat's agent lives in `selectedAgent`, which `useBrowserHistory`'s mount-time `applyRoute()` fills in a moment later. The comment on the slice is accurate, so this is deliberate; the consequence is that a post-login deep link to `/chat/:agent` still renders one frame with no agent selected. Not a problem for the bind flow this PR fixes (that route carries no agent), but the dashboard-flash class of bug this PR removes for `/slack/bind` remains for `/chat/:agent`.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `a457bb44`

## PR #3110: fix(ui): bind deep links survive login and the Terms gate
**Author:** pilartomas | **Branch:** fix/bind-deep-link-survives-login → main | **Changes:** +258 −14 (9 files)

### Summary
Extracts a single `return-path` primitive that parks the user's destination across both interstitials (the Keycloak login roundtrip and the Terms gate) in separate sessionStorage slots, and re-derives the store's route after they settle so the first paint is the deep link rather than the dashboard.

### Findings
- 🟡 **Warning:** `isSafeReturnPath` accepts control characters that make the browser resolve the path cross-origin (`packages/ui/src/lib/return-path.ts:37`)
- 🟢 **Suggestion:** `hydrateRoute()` drops `selectedAgent`, leaving a one-frame gap on a `/chat/:agent` deep link (`packages/ui/src/main.tsx:36`)
- ✅ **Looks good:** The `?flow=` deep link genuinely survives — `SlackBindView` reads `flowId` at module scope, but `app.js` is dynamically imported *after* `hydrateRoute()`, so the URL is already restored when that module-level read runs.

✅ No findings. `docs/architecture/security-and-credentials.md` was updated in the same PR with the interstitial park-and-resume behavior, `Last verified` bumped to 2026-07-31 (matches the head commit date), no ADR references, no volatile content added, page at 34k chars under the 40k cap.

✅ No findings. The diff is client-side; the architecture rules (bounded-context modules, three-layer slices, contract-package validation, service-factory DI) govern server modules. Module boundaries are respected — the new imports target the pre-existing shared `packages/ui/src/lib/`, not another module's internals.

🟢 **Suggestion** — state lineage (`main.tsx:36`): `hydrateRoute()` sets only the four route fields via `routeToNavigationState`, which returns `agentId: null` for chat routes. `selectedAgent` is filled in separately by `useBrowserHistory`'s mount-time `applyRoute()`. Correct as documented, but it means a post-login `/chat/:agent` deep link renders one frame with `selectedAgent` still `null`. The URL is the source of truth for both fields, so hydrating them from two places is the residual lineage split; not a defect for the bind flow this PR targets.

### Verdict
COMMENT — the fix is well-structured and the deep-link path verified sound; one hardening gap in the path validator worth closing before merge.

---

packages/ui/src/lib/return-path.ts
pilartomas: Confirmed and fixed in 36dc6068. I reproduced it against the WHATWG parser before changing anything: `/<tab>/evil.example`, and the LF and CR variants, all passed both guards and resolved to `https://evil.example/`, while `/terms<tab>` and `/<tab>terms` slipped past the interstitial check.

Taken as suggested — the validator now strips tab/LF/CR first, so it judges the string the browser will actually act on rather than the raw text. Accepted values are still returned verbatim; the browser normalises them to the same target it was validated as. Added table-driven cases for all three characters covering the protocol-relative form, the backslash form, and both interstitial evasions.

packages/ui/src/main.tsx
pilartomas: Leaving this as designed, for two reasons.

The state after `hydrateRoute()` on a `/chat/:agent` deep link is identical to a plain refresh while in a chat: the slice initialiser already resolves that URL to `view: "chat"` with `selectedAgent` null, and `useBrowserHistory` fills the agent in on mount. So it is a supported, already-reachable state rather than something this change introduces — and the dashboard frame that used to precede it on a post-login return is gone, which makes this route strictly better off than before.

Closing the remaining frame means either duplicating chat-context knowledge in `hydrateRoute` or extracting a shared apply-route action, and chat teardown (`resetChatContext`, clearing `selectedAgent` on the way out) is deliberately owned by `useBrowserHistory`. Unifying the two hydration paths is a worthwhile refactor of that ownership, but a separate one from this bug — happy to file it as a follow-up if you want it tracked.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `36dc6068`

## PR #3110: fix(ui): bind deep links survive login and the Terms gate
**Author:** pilartomas | **Branch:** fix/bind-deep-link-survives-login → main | **Changes:** +274 −14 (9 files)

### Summary
Re-review of the control-character hardening in `isSafeReturnPath`. The prior review's 🟡 warning is fixed and verified closed; a narrower residual remains in the interstitial half of the same check.

### Changes since last review
Previous HEAD: `a457bb44` (2026-08-02T19:24:31Z) — verdict COMMENT

- ✅ **Fixed:** control-character origin escape in `isSafeReturnPath` — swept all 264 candidate characters against the WHATWG parser, 0 now escape the origin (`packages/ui/src/lib/return-path.ts:38`)
- 🆕 **New:** trailing C0/space still evades the interstitial check — 30 characters resolve to `/terms` or `/auth/callback` while passing (`packages/ui/src/lib/return-path.ts:41`)
- 🟢 Suggestion on `main.tsx:36` (one-frame `selectedAgent` gap): declined as designed with a reasoned justification — accepted, recorded as a PR-local override, not re-raised.

### Findings
- 🟡 **Warning:** stripping only tab/LF/CR leaves the interstitial check bypassable by trailing C0 controls and space (`packages/ui/src/lib/return-path.ts:41`)

✅ No findings. `docs/architecture/security-and-credentials.md` carries the park-and-resume behavior for both interstitials, `Last verified` bumped 2026-07-30 → 2026-07-31, no ADR references or volatile content added, no ADR status changes in the diff, no architecture page moved or deleted.

✅ No findings. The delta is a client-side validator change plus tests; the architecture rules (bounded-context modules, three-layer slices, contract-package validation, service-factory DI) govern server modules. No cross-module import introduced — the new imports target the pre-existing shared `packages/ui/src/lib/`, never another module's internals.

✅ No findings. The delta touches no `.tsx` state, lineage, or component structure — `main.tsx` and `terms-view.tsx` are unchanged since the last review, whose lineage suggestion the author declined with a reason (the rules ask for exactly that rather than silent deviation).

### Verdict
COMMENT — the reported origin escape is genuinely and completely closed; the remaining gap is a same-line hardening nit on the interstitial half, not a blocker.

---

packages/ui/src/lib/return-path.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+/** Stored values are untrusted on the way out: only same-origin paths pass. */
+export function isSafeReturnPath(value: string): boolean {
+  // Browsers drop tab/LF/CR while parsing, so "/<tab>/host" is protocol-relative
+  // after the fact — validate what the browser will act on, not the raw text.
+  const parsed = value.replace(/[\t\n\r]/g, "");
+  // "//host" and "/\host" are protocol-relative URLs, not paths.
+  if (!parsed.startsWith("/") || /^\/[/\\]/.test(parsed)) return false;
+  return !INTERSTITIAL_PATHS.includes(parsed.split(/[?#]/)[0]!);
```
dam-code-guardian: 🟡 **Warning:** the strip set closes the origin escape but not the interstitial evasion you also reproduced. `[\t\n\r]` covers the characters browsers drop *inside* a URL; the parser additionally trims **all** leading/trailing C0 controls and spaces. So `"/terms" + "\f"` (and every other trailing C0, plus a plain trailing space) passes `INTERSTITIAL_PATHS.includes(...)` — the raw string isn't `/terms` — while `new URL()` resolves it to pathname `/terms`.

Swept all 256 C0-through-Latin-1 code points plus the usual Unicode space suspects against Node's WHATWG parser, in 11 positional shapes each:

- **cross-origin escapes: 0** — the fix here is complete, `//host`, `/\host` and the interior tab/LF/CR forms are all rejected.
- **interstitial evasions: 30** — `/terms<C0>` and `/auth/callback<C0>` for every C0 control (`U+0000`–`U+0020` except the three already stripped), e.g. `\f`, `\v`, `\0`, and `" "`.

Impact is bounded — the reachable outcome is re-entering the Terms gate or landing on the callback with a spent code (the loop your `INTERSTITIAL_PATHS` guard exists to prevent), not a cross-origin redirect. Worth closing on the same line, since it's the same root cause: judging a string the browser will normalize before acting on it.

Widening the class to all of C0 + space keeps every legitimate destination passing (`/slack/bind?flow=abc-123`, `/settings/usage`, `/chat/agent-1#top`, `/my file`, `/terms-of-service`, `/termsx`) and takes both counts to zero:

```suggestion
  const parsed = value.replace(/[\0-\x20]/g, "");
  // "//host" and "/\\host" are protocol-relative URLs, not paths.
  if (!parsed.startsWith("/") || /^\/[/\\]/.test(parsed)) return false;
  return !INTERSTITIAL_PATHS.includes(parsed.split(/[?#]/)[0]!);
```

Note this only widens what is *stripped for judging* — accepted values are still returned verbatim, so the existing round-trip behaviour is unchanged. Your `it.each` table extends naturally: adding `"\f"`, `"\v"` and `" "` to the character list covers the trailing-form cases the current five assertions don't reach.

pilartomas: Reproduced and taken as suggested in 40042e3e.

My own sweep before changing anything matched yours: with the tab/LF/CR-only strip, cross-origin escapes were already 0, but the trailing forms reached an interstitial 60 times across my shapes (`/terms<C0>` and `/auth/callback<C0>` for every C0 plus space). Widening the judged class to `[\0-\x20]` takes both counts to zero, and no legitimate destination is rejected — I checked the bind/telegram flow links, `/settings/usage`, `/chat/agent-1`, a percent-encoded sandbox path, and the prefix-sharing `/terms-of-service` and `/termsx`.

Re-swept C0 through Latin-1 in eleven positional shapes (2816 candidates) against the shipped predicate: 0 escapes, 0 evasions.

One deliberate difference from a byte-exact model of the parser: it drops tab/LF/CR anywhere but trims the rest only at the ends, so stripping the whole class judges slightly more than the parser removes — e.g. an interior `<FF>`, which the parser would percent-encode. That can only ever reject, which is the safe direction for this predicate, and the comment now says so. Tests split along the real distinction: interior forms for the three dropped-anywhere characters, trailing forms for the full trimmed class, plus the prefix-sharing positives.

packages/ui/src/lib/return-path.ts
```diff
@@ -0,0 +1,75 @@
...
+export function isSafeReturnPath(value: string): boolean {
+  // Validate what the browser will act on, not the raw text: it drops tab/LF/CR
+  // anywhere while parsing ("/<tab>/host" ends up protocol-relative) and trims
+  // C0 controls and spaces at the ends ("/terms<FF>" ends up on the gate).
+  // Stripping the whole class judges a little more than the parser removes,
+  // which only ever rejects — the safe direction.
+  const parsed = value.replace(/[\0-\x20]/g, "");
+  // "//host" and "/\host" are protocol-relative URLs, not paths.
+  if (!parsed.startsWith("/") || /^\/[/\\]/.test(parsed)) return false;
+  return !INTERSTITIAL_PATHS.includes(parsed.split(/[?#]/)[0]!);
```
dam-code-guardian: 🟡 **Warning:** the strip class closes the character-level evasion completely — I re-swept C0 through Latin-1 plus the Unicode space suspects in 11 positional shapes (2893 candidates) against Node's WHATWG parser: **0 cross-origin escapes, 0 interstitial evasions**. But the same line judges a path the parser will still *normalize*, and dot segments survive the strip:

```
isSafeReturnPath("/a/../terms")            → true,  new URL(…).pathname === "/terms"
isSafeReturnPath("/./terms")               → true,  pathname === "/terms"
isSafeReturnPath("/foo/../auth/callback")  → true,  pathname === "/auth/callback"
isSafeReturnPath("/%2e%2e/terms")          → true,  pathname === "/terms"
```

13 such forms pass in my syntax probe. `[\0-\x20]` removes characters; it can't collapse `..` — so `INTERSTITIAL_PATHS.includes(parsed.split(/[?#]/)[0])` compares the raw text while the browser acts on the resolved pathname. Same root cause as the last two rounds (judging a string the browser rewrites before acting on it), one layer up: normalization rather than stripping.

Reachable through both consumers of an attacker-writable slot — `takeReturnPath` → `window.location.assign` (`terms-view.tsx:56`) and `initialPath` → `history.replaceState` (`navigation.ts:69`). Impact is the bounded one your guard exists to prevent, not a cross-origin redirect: re-entering the Terms gate, or landing on the callback with a spent code.

Letting the parser do the normalizing takes both counts to zero and closes the class rather than another instance of it:

```suggestion
  const stripped = value.replace(/[\0-\x20]/g, "");
  // "//host" and "/\host" are protocol-relative URLs, not paths.
  if (!stripped.startsWith("/") || /^\/[/\\]/.test(stripped)) return false;
  // Judge the parser's own output: it collapses dot segments, so "/a/../terms"
  // reaches the gate while the raw text isn't in the list.
  const url = new URL(stripped, window.location.origin);
  if (url.origin !== window.location.origin) return false;
  return !INTERSTITIAL_PATHS.includes(url.pathname);
```

I ran the same audit over this predicate: **0 escapes, 0 evasions** across the character sweep *and* the 17-form syntax probe, and every legitimate destination I could think of still passes — `/`, `/slack/bind?flow=abc-123`, `/settings/usage`, `/chat/agent-1#top`, `/settings/connections#tokens`, `/my file`, `/sandbox/%20path`, `/terms-of-service`, `/termsx`, `/terms/sub`, `/auth/callbackx`. Note it also returns the *stripped* value's verdict while callers still use the raw string, exactly as today — worth a test that `/a/../terms` is rejected, alongside the `it.each` tables you split.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `40042e3e`

## PR #3110: fix(ui): bind deep links survive login and the Terms gate
**Author:** pilartomas | **Branch:** fix/bind-deep-link-survives-login → main | **Changes:** +286 −14 (9 files)

### Summary
Re-review of the widened strip class in `isSafeReturnPath`. The previous review's trailing-control evasion is fixed and verified closed; sweeping the predicate again surfaced a different root cause on the same line — path normalization, not character stripping.
_(Suppressed 1 finding(s) per PR-local overrides: main.tsx:36 selectedAgent one-frame gap.)_

### Changes since last review
Previous HEAD: 36dc6068 (2026-08-03T06:56:21Z) — verdict COMMENT

- ✅ **Fixed:** trailing C0/space interstitial evasion — re-swept 2893 candidates against the WHATWG parser, 0 escapes and 0 evasions from the character class (`packages/ui/src/lib/return-path.ts:41`)
- 🆕 **New:** dot-segment normalization still reaches both interstitials (`packages/ui/src/lib/return-path.ts:44`)

### Findings
- 🟡 **Warning:** `INTERSTITIAL_PATHS.includes` compares an un-normalized path, so `/a/../terms` re-enters the gate (`packages/ui/src/lib/return-path.ts:44`)

✅ No findings. `docs/architecture/security-and-credentials.md` carries the park-and-resume behavior for both interstitials at architecture level, `Last verified` bumped 2026-07-30 → 2026-07-31, no ADR reference or volatile content added, no ADR status changed in the diff, no architecture page moved or deleted. The delta since the last review is validator + test only — no new subsystem, no doc-affecting behavior change.

✅ No findings. The delta is a client-side predicate change plus tests. Every import the PR adds targets the pre-existing shared `packages/ui/src/lib/`, never another module's internals; the one cross-module import in `on-terms-stale.ts` (`platform/lib/routes.js`) predates this diff and is out of scope against the diff baseline.

✅ No findings. Both routed `.tsx` files are byte-identical to the previously reviewed SHA — the delta touches no component, state lineage, or hook. The URL remains the single source of truth for the route; no cross-lineage duplication is introduced. The prior review's lineage suggestion was declined with a reason, which the rules ask for rather than silent deviation.

### Verdict
COMMENT — the reported character-class evasion is genuinely and completely closed; what remains is a distinct normalization gap on the same line, same bounded impact, not a blocker.

---

packages/ui/src/lib/return-path.ts
pilartomas: Reproduced, and fixed by changing the shape rather than the character class — 4e910492.

You were right that this was the same root cause a layer up, and three rounds of it means the predicate itself was the defect: every version compared raw text against a destination the browser rewrites before acting on it. So the module no longer exposes a verdict-only entry point. It resolves the value and hands back the parser's own output; callers navigate to what was judged, because there is nothing else for them to navigate to. `isSafeReturnPath` is gone, replaced by `resolveReturnPath` (path + query + fragment) and `resolveReturnPathname` (the routing form), and `navigation.ts` now restores the resolved pathname instead of the stored text.

Two things my audit caught that are worth flagging, because one of them was mine:

1. **The rewrite introduced a fresh escape.** Returning `url.pathname` is not a safe round-trip: dot segments collapse `/a/..//evil.example` to pathname `//evil.example`, which is same-origin as a *result* and protocol-relative as a *reference* — so handing it back produced an open redirect. The resolver now re-resolves its own output and rejects unless it reproduces the same URL. That check is what makes the contract provable instead of argued.
2. **A crash.** `//%00evil.example` throws `TypeError` from the round-trip parse. Unguarded that would have taken out the Terms-accept navigation and, through `initialPath`, boot. Every parse now sits inside one guard.

Audit over 5374 generated candidates — C0 through Latin-1 plus the Unicode space suspects in 20 positional shapes, plus 60 normalization and scheme forms (encoded dots, `..%2f`, backslashes, multi-slash, userinfo, `blob:`/`data:`/`javascript:`/`file:`) — on three invariants: nothing accepted resolves off-origin or onto an interstitial; the returned text resolves to the same URL as the input; the routing form is always a single-slash same-origin path. **0 / 0 / 0**, and none of the legitimate destinations rejected (bind links, `/chat/agent-1`, percent-encoded sandbox paths, `/terms-of-service`, `/termsx`, `/terms/sub`, `/auth/callbackx`).

Tests reorganized around the three classes that actually reached a wrong destination — interior control characters, trailing trimmed ones, dot segments — plus the resolved-path-as-reference case and the no-throw case. 194 UI tests, full repo suite green.

One behaviour change worth your eye: the return-view restore now uses the resolved pathname, so a query stored in that slot is dropped rather than concatenated with the live one. Only a bare view path is ever written there, and the old concatenation could produce `?a=1?b=2`.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `4e910492`

## PR #3110: fix(ui): bind deep links survive login and the Terms gate
**Author:** pilartomas | **Branch:** fix/bind-deep-link-survives-login → main | **Changes:** +417 −19 (9 files)

### Summary
Re-review of the return-path rewrite: the verdict-only predicate is gone, replaced by resolvers that hand back the WHATWG parser's own output, with a round-trip check and a parse guard. The previous review's dot-segment gap is fixed, and the class of defect it belonged to is now closed structurally.
_(Suppressed 1 finding(s) per PR-local overrides: main.tsx:36 selectedAgent one-frame gap.)_

### Changes since last review
Previous HEAD: 40042e3e (2026-08-03T08:00:31Z) — verdict COMMENT

- ✅ **Fixed:** dot-segment normalization reaching the interstitials — `resolveReturnPath` now judges and returns `URL`-canonical output; my sweep of 5336 candidates found 0 origin escapes, 0 interstitial evasions, 0 round-trip mismatches, 0 throws, and no legitimate destination rejected (`packages/ui/src/lib/return-path.ts:47`)
- ✅ **Fixed:** the underlying validate-one-string-navigate-to-another shape — removing the verdict-only entry point is what makes the three prior rounds non-recurring, not a fourth character class (`packages/ui/src/lib/return-path.ts:44`)

### Findings

✅ No findings. `docs/architecture/security-and-credentials.md` carries the park-and-resume behavior for both interstitials at architecture level, `Last verified` bumped 2026-07-30 → 2026-07-31, no ADR reference or `Motivated by:` section, no volatile content (no symbol names or file paths added — the page names the concept and cross-links `channels.md`), page at 34.2k chars under the cap, no architecture page moved or deleted, no ADR status changed in the diff. The delta since the last review is validator + test only.

✅ No findings. The delta is client-side; the architecture rules (bounded-context modules, three-layer slices, contract-package validation, service-factory DI) govern server modules. Import direction is clean: `navigation.ts` and `on-terms-stale.ts` reach only into the pre-existing shared `packages/ui/src/lib/`, never another module's internals. Types stay at the boundary — the resolvers return `string | null` with no assertions, and `LocationLike` is widened by one field rather than loosened.

✅ No findings. Both routed `.tsx` files are byte-identical to the previously reviewed SHA — the delta touches no component, hook, or state lineage. The URL remains the single source of truth for the route; no cross-lineage duplication is introduced.

### Verdict
APPROVE — the reported gap is closed, and this revision fixes the shape that produced all three rounds rather than the character class.

---

PR #3111, opened 2026-07-31
feat(ui): add copy action to agent code blocks

### LINKED ISSUE #3081: UI - Add copy action to agent code blocks
## Context

Agents often produce code blocks in chat responses.

## Problem

Code blocks are rendered in the UI but cannot be copied directly. Users have to manually select the text, which is error-prone and inconvenient.

## Goal

Users can copy a code block from an agent response with a single action.

## Proposed Solution

Add a copy button to rendered code blocks that copies the code contents to the clipboard.

Reuse the existing copy interaction pattern, including confirmation feedback after copying.

## Scope

- In: Code blocks generated by agents
- Out: Copying entire responses or conversations

PetrBulanek:
Code blocks in agent responses can now be copied with a single click. On a pointer device the copy button appears in the block's top-right corner on hover; on a touch device — where there is no hover to wait for — it is always shown. Clicking it puts the block's exact contents on the clipboard: indentation intact, line endings normalized, and trailing blank lines dropped so the last line doesn't auto-execute when pasted into a terminal. Success and failure are both confirmed — checkmark or warning glyph, plus a live-region announcement for screen readers. Works in every place the app renders markdown, including skill pages and file previews.

Closes #3081


packages/ui/src/components/markdown-code-block.tsx
```diff
@@ -0,0 +1,66 @@
...
+type HastChild = HastElement["children"][number];
+
+/** Raw source text of a rendered code block. rehype-highlight wraps every
+ *  token in nested <span>s, so the text only exists at the leaves — walk the
+ *  hast node instead of reading React children. */
+export function codeBlockText(node: HastElement | undefined): string {
+  if (!node) return "";
+  // A fence always ends in a newline; strip exactly one so pasting into a
+  // terminal doesn't auto-execute the last line.
+  return collectText(node).replace(/\n$/, "");
```
dam-code-guardian: 🟡 **Warning:** `.replace(/\n$/, "")` strips only one newline and does not match a CRLF tail, so the trailing-newline hazard the comment describes still gets through in two cases.

`collectText` returns the hast text verbatim. A fence whose last content line is followed by a blank line yields `…code\n\n` — one `\n` survives into the clipboard, which is exactly the auto-execute-on-paste case this is meant to prevent. A CRLF source yields `…code\r\n` → `…code\r`, leaving a stray carriage return.

If the intent is "never end in a newline", say that:

```suggestion
  return collectText(node).replace(/[\r\n]+$/, "");
```

(Use `/\r?\n$/` instead if stripping exactly one line ending is deliberate.)

packages/ui/src/components/markdown-code-block.tsx
```diff
@@ -0,0 +1,66 @@
...
+      {code !== "" && (
+        <Button
+          type="button"
+          variant="outline"
+          size="icon-sm"
+          onClick={() => void copy(code)}
+          aria-label={LABELS[state]}
+          title={LABELS[state]}
+          className={cn(
+            "absolute right-2 top-2 transition-opacity opacity-100 md:opacity-0 md:group-hover:opacity-100 focus:opacity-100",
```
dam-code-guardian: 🟡 **Warning:** `focus:opacity-100` diverges from the `Button` primitive and from this repo's own reveal pattern, both of which key on `focus-visible:`.

`Button`'s focus ring is `focus-visible:ring-2` (`packages/ui/src/components/ui/button.tsx:8`), and the sibling hover-reveal at `packages/ui/src/modules/sessions/views/chat-view.tsx:489` uses `md:opacity-0 md:group-hover:opacity-100 focus-visible:opacity-100`. Matching that keeps the reveal trigger and the ring on the same state.

Worth noting either way: `opacity-0` elements stay in the tab order, so a long agent reply puts one focus stop per code block in a keyboard path — reachable, but only visible once focused.

```suggestion
            "absolute right-2 top-2 transition-opacity opacity-100 md:opacity-0 md:group-hover:opacity-100 focus-visible:opacity-100",
```

packages/ui/src/components/markdown-code-block.tsx
```diff
@@ -0,0 +1,66 @@
...
+export function MarkdownCodeBlock({
+  node,
+  children,
+  ...preProps
+}: MarkdownCodeBlockProps) {
+  const { copy, state } = useCopy();
+  const code = codeBlockText(node);
+
+  return (
+    <div className="code-block group">
```
dam-code-guardian: 🟢 **Suggestion:** two invariants this component relies on live in ambient CSS rather than with the component.

`absolute right-2 top-2` on the button only positions correctly because `.prose .code-block { position: relative }` (`packages/ui/src/App.css:453`) is in scope, and the `pre` margin reset is likewise `.prose`-scoped. Rendered anywhere outside `.prose`, the button escapes to the nearest positioned ancestor. `className="code-block group relative"` makes the containing block local to the component and leaves the CSS rule as a harmless duplicate.

Separately, the bare `group` is unscoped: `md:group-hover:opacity-100` binds to the *nearest* ancestor carrying `group`. Nothing collides today (the one `group` in `chat-view.tsx:475` closes at line 518, well above the `<Markdown>` at line 715), but a future call site wrapping `<Markdown>` in a hover group would reveal every code block button at once. This repo already uses named groups where it matters (`group/head`, `packages/ui/src/modules/experiments/components/sandbox-group-card.tsx:35`) — `group/code-block` + `md:group-hover/code-block:opacity-100` makes the coupling explicit.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `e255d1a`

## PR #3111: feat(ui): add copy action to agent code blocks
**Author:** PetrBulanek | **Branch:** feat/3081-code-block-copy → refactor/3052-shared-ui-helpers | **Changes:** +75 −0 (3 files)

### Summary
Adds a hover-revealed copy button to every markdown code block by overriding react-markdown's `pre` renderer with a new `MarkdownCodeBlock`, reading the raw fence text from the hast node (correct — rehype-highlight leaves the text only at token leaves) and reusing the shared `useCopy` hook.

### Findings
- 🟡 **Warning:** trailing-newline strip misses `\n\n` and CRLF tails (`packages/ui/src/components/markdown-code-block.tsx:19`)
- 🟡 **Warning:** `focus:` diverges from the `Button` primitive's `focus-visible:` reveal (`packages/ui/src/components/markdown-code-block.tsx:56`)
- 🟢 **Suggestion:** the component depends on ambient `.prose` CSS for two of its own layout invariants (`packages/ui/src/components/markdown-code-block.tsx:45`)
- ✅ **Looks good:** the hast walk is the right call — reading React children would return highlight-token `<span>`s, not the source text.

**Verdict: aligned**

The diff touches only `packages/ui/src`. No page under `docs/architecture/` describes code-block rendering or clipboard behaviour, so nothing contradicts the code:

- `platform-topology.md` § ui describes the ui at the level of "React + Vite SPA, tRPC + ACP over WebSocket" — unchanged by this PR.
- `artifact-library.md`'s one-liner about previews reusing "the chat file-viewer stack (markdown prose, highlighted code, inline images)" remains accurate.
- No architecture page was edited, so no `Last verified:` bump is owed; no new subsystem, no page moves, no ADR status changes in the diff.

**Verdict: minor findings — the component is well-shaped; two real defects, one structural note.**

Verified strengths: the hast walk is the correct source for the raw text, the shared `useCopy` hook is reused rather than re-implemented, props are typed structurally (`ComponentPropsWithoutRef<"pre"> & ExtraProps`, no `any`/`as`), it is a 66-line leaf with ~20 JSX lines, and `cn()` is used for conditionals per the styling rules.

Findings are carried as the inline comments on this review (lines 19, 56, 45). One structural note, not a finding:

Placing `codeBlockText` as a named export beside the component is the right call — it is the one piece of pure logic here, and `packages/ui/vitest.config.ts` collects `src/__tests__/unit/**/*.test.ts` in a node environment, so it is unit-testable without a DOM. The repo has 177 test files but none under `packages/ui/src/components/`; a small test pinning the trailing-newline behaviour (including the `\n\n` and CRLF cases) would fit the existing harness exactly.

### Verdict
COMMENT — the feature is correctly built and the hard part (reading text from hast, not from React children) is right; two contained defects in the newline strip and the focus reveal are worth fixing before merge, neither is blocking.

---

kapetr (commented):
Re-review at `75a91fe`. 2 bugs, 4 issues, nits — inline.

Verified from the earlier round: `focus-visible:opacity-100` (0,2,0) does beat `md:opacity-0`, and `/[\r\n]+$/` correctly handles the blank-line and CRLF-ending cases.

packages/ui/src/components/markdown-code-block.tsx
```diff
@@ -0,0 +1,66 @@
...
+      {code !== "" && (
+        <Button
+          type="button"
+          variant="outline"
+          size="icon-sm"
+          onClick={() => void copy(code)}
+          aria-label={LABELS[state]}
+          title={LABELS[state]}
+          className={cn(
+            "absolute right-2 top-2 transition-opacity opacity-100 md:opacity-0 md:group-hover:opacity-100 focus-visible:opacity-100",
```
kapetr: **Bug — invisible but tappable on touch devices ≥768px.**

Tailwind 4.2.2 compiles `hover` as `&:hover { @media (hover: hover) { … } }`, and `group-hover` is a compound over it. On a touch-primary tablet (iPad landscape 1024px, portrait 810px, Android/Windows tablets): `md:opacity-0` applies, `md:group-hover:opacity-100` is dropped → nothing ever reveals the button. That's the platform the PR body says gets an always-visible button.

`opacity: 0` keeps hit-testing and tab order, so there's an invisible 28×28 target over the top-right of line 1 — reaching there to scroll the code horizontally or start a selection silently overwrites the clipboard.

`md:` is standing in for "not touch", but breakpoint and hover-capability are orthogonal. Let the hover query do it:

```suggestion
            "absolute right-2 top-2 transition-opacity opacity-100 [@media(hover:hover)]:opacity-0 group-hover:opacity-100 focus-visible:opacity-100",
```

packages/ui/src/components/markdown-code-block.tsx
```diff
@@ -0,0 +1,66 @@
...
+            state === "copied" && "text-success",
+            state === "failed" && "text-danger",
```
kapetr: **Bug — the `failed` state is silent on desktop.**

`Button`'s `outline` variant carries `hover:bg-muted hover:text-foreground`. tailwind-merge won't drop it against `text-success`/`text-danger` (different modifier, no conflict group), and `.hover\:text-foreground:hover` (0,2,0) beats `.text-danger` (0,1,0) regardless of order. After a click the pointer is still on the button by definition, so `:hover` holds and the colour is overridden.

`copied` only loses its tint (the glyph still swaps to `Checkmark`). `failed` has no glyph change — colour is its only signal — so on a non-secure-context deploy or a permission denial the user sees nothing and pastes stale clipboard content.

```suggestion
            state === "copied" && "text-success hover:text-success",
            state === "failed" && "text-danger hover:text-danger",
```

Still needs a non-colour signal for `failed` — a `Warning`/`ErrorFilled` glyph at minimum. `copyable-command.tsx:36` renders an explicit manual-copy fallback as precedent.

packages/ui/src/components/markdown-code-block.tsx
```diff
@@ -0,0 +1,66 @@
...
+    <div className="code-block group">
+      <pre {...preProps}>{children}</pre>
+      {code !== "" && (
+        <Button
+          type="button"
+          variant="outline"
+          size="icon-sm"
+          onClick={() => void copy(code)}
+          aria-label={LABELS[state]}
+          title={LABELS[state]}
```
kapetr: The copied/failed transition is announced to nobody: the only state carrier is the button's own accessible name, and a name change on an already-focused element isn't reliably announced by VoiceOver or NVDA. Keyboard/SR users get no confirmation on success *or* failure.

Stable name + a live region:

```tsx
<Button aria-label="Copy code" …>{…}</Button>
<span className="sr-only" role="status" aria-live="polite">
  {state === "copied" ? "Copied" : state === "failed" ? "Copy failed" : ""}
</span>
```

Precedent: `egress-approval-toasts.tsx:82`, `spinner.tsx:36`.

Also: `title` duplicating `aria-label` doubles the announcement in some SR configurations, and puts a native tooltip on a hover-revealed button. And with ten fences in one document an SR button list shows ten identical "Copy code" entries — the language is available at `node.children[0].properties.className`.

packages/ui/src/components/markdown-code-block.tsx
```diff
@@ -0,0 +1,66 @@
...
+type HastChild = HastElement["children"][number];
+
+/** Raw source text of a rendered code block. rehype-highlight wraps every
+ *  token in nested <span>s, so the text only exists at the leaves — walk the
+ *  hast node instead of reading React children. */
+export function codeBlockText(node: HastElement | undefined): string {
+  if (!node) return "";
+  // Strip all trailing newlines (LF or CRLF) so pasting into a terminal
+  // doesn't auto-execute the last line.
+  return collectText(node).replace(/[\r\n]+$/, "");
```
kapetr: Two mismatches between this line and its claims:

1. `75a91fe`'s subject says "strip CRLF", but only *trailing* `\r\n` is stripped — interior CRLF is untouched, so a CRLF-authored file previewed as markdown still pastes `^M` into a terminal. Either normalize (`.replace(/\r\n/g, "\n").replace(/\n+$/, "")`) or reword to "trailing CRLF".
2. The rule is now "strip *all* trailing newlines" (fine, GitHub does the same) but the PR body still promises "without the trailing newline", singular. A fence deliberately ending in blank lines no longer round-trips.

packages/ui/src/components/markdown-code-block.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { useCopy } from "@/hooks/use-copy";
+import { cn } from "@/lib/utils";
+
+type HastElement = NonNullable<ExtraProps["node"]>;
+type HastChild = HastElement["children"][number];
+
+/** Raw source text of a rendered code block. rehype-highlight wraps every
+ *  token in nested <span>s, so the text only exists at the leaves — walk the
+ *  hast node instead of reading React children. */
+export function codeBlockText(node: HastElement | undefined): string {
```
kapetr: `codeBlockText` is exported but has no test — `git grep codeBlockText` finds only the definition and its one in-file call, while `src/__tests__/unit/` has 18 suites for this class of pure helper.

The trailing-newline rule is the highest-churn line in the PR (already revised once) and is the thing to pin: nested token spans, `undefined` node, empty fence, whitespace-only fence, CRLF ending, multiple trailing blank lines.

packages/ui/src/App.css
```diff
@@ -473,6 +473,13 @@
   color: var(--foreground);
   box-shadow: none;
 }
+.prose .code-block {
```
kapetr, line 480: The button's positioning context lives here, but the component that depends on it doesn't know. Rendered outside a `.prose` ancestor, `absolute right-2 top-2` anchors to the nearest positioned ancestor instead — in chat that's `chat-view.tsx:461`'s `relative` root, i.e. a copy button floating over the top-right of the whole chat panel.

Move it to the component (`className="code-block group relative"`). Keeping the `margin: 0` reset in CSS is correct — Tailwind's `m-0` (0,1,0) loses to `.prose pre` (0,2,0).

packages/ui/src/components/markdown-code-block.tsx
```diff
@@ -0,0 +1,66 @@
...
+export function MarkdownCodeBlock({
+  node,
+  children,
+  ...preProps
+}: MarkdownCodeBlockProps) {
+  const { copy, state } = useCopy();
+  const code = codeBlockText(node);
+
+  return (
+    <div className="code-block group">
```
kapetr: Nit: name the group (`group/code-block` + `group-hover/code-block:`). `group-hover` matches *any* hovered `.group` ancestor; no collision today, but adding `group` to a message container later would reveal every code button in that message at once.

packages/ui/src/components/markdown-code-block.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+type MarkdownCodeBlockProps = ComponentPropsWithoutRef<"pre"> & ExtraProps;
+
+export function MarkdownCodeBlock({
+  node,
+  children,
+  ...preProps
+}: MarkdownCodeBlockProps) {
+  const { copy, state } = useCopy();
+  const code = codeBlockText(node);
```
kapetr, line 20: Nit: `useMemo(() => codeBlockText(node), [node])` — this re-walks the whole subtree on every render, including the two that `useCopy`'s state machine causes, and code blocks re-render constantly during streaming.

Also `MarkdownCodeBlockProps` (line 34) isn't exported, so per the react-ui-engineering skill it should just be `Props`.

PetrBulanek: `Props` renamed in 9735d53e.

Skipped the `useMemo`: react-markdown re-parses in its render body, so `node` is a fresh object on every streaming frame — the memo misses exactly when it would matter. It only hits on the two `useCopy` transition renders, where the work is one block’s subtree.

packages/ui/src/components/markdown-code-block.tsx
PetrBulanek: Live region added in 9735d53e, and the accessible name is now static.

Kept `title` — it is the same string as `aria-label`, so at worst a duplicate description, and it is the only affordance hint an icon-only button gives on desktop.

Skipped the language in the label: `properties.className` is `{}` on a bare fence, so it needs a fallback branch, and the payoff is limited to the rotor button list. Will add it if you think that list matters here.

kapetr (approved):

PR #3026, opened 2026-07-28
fix(ui): collect private-registry credentials on the Image step

kapetr:
Fixes #2748

## What

Entering a custom image URL (wizard Step 1) and the private-registry credentials needed to pull it (Step 2) were split across pages, so users with private images never found the credential fields.

- The **Private registry** collapsible now lives inside the Custom image card, directly under the URL input — it appears once a custom image is entered, collapsed by default, matching the design in the issue.
- The all-or-nothing partial-credentials gate moved with it: Step 1's **Continue** (and Enter-to-advance) blocks on partially filled credentials with the inline hint; Step 2 no longer knows about the registry.
- Credentials stay ephemeral component state — never persisted to the sessionStorage wizard snapshot (they're secrets), unchanged from before.
- Touch-up while moving: the section's chevrons switched lucide → Carbon (house icon rule), and it gained `aria-expanded`.

No backend changes — the create payload is identical.

## Testing

- `mise run ui:check`, `mise run ui:test` (108 passing)
- Manually verified against a live auth-gated registry: deployed an htpasswd-protected `registry:2` on the dev cluster, pushed the agent image into it, and created a sandbox through the wizard with image + credentials entered together on Step 1 — the pod pulled via the generated pull secret and reached Running. Partial credentials block Step 1; anonymous pull of the same image is denied.

Issue #2748, opened 2026-07-09
UI - Private registry credentials are on a different step than the image URL


jenna-winkler:
## Problem

When creating a sandbox from a custom image, you enter the image URL on the Image step (Step 1). But if that image is in a private registry requiring credentials, the "Private registry" credential fields are on the Setup step (Step 2) - a separate page.

## Impact

The credentials needed to pull an image are disconnected from where you specify that image. A user entering a private image URL on Step 1 has no indication that the credentials to pull it live on the next page - so they don't find them, and it's unclear the option even exists. 

This matters specifically for private registries that require credentials (the intended path for pulling private/custom workload images).

## Expected

The private registry credentials should be part of the same step where you provide the custom image URL (Step 1 / Image), so pulling a private image is a single, self-contained action - not split across two pages.

## Desired state

The private registry expandable menu should live directly underneath where you input the image URL on the first page of agent creation ("Choose your starting point")

**Closed state (default):** 

<img width="502" height="197" alt="Image" src="https://github.com/user-attachments/assets/88deecd7-c572-4c08-86a2-8ad3f42fd0d2" />

**Expanded state**

<img width="481" height="456" alt="Image" src="https://github.com/user-attachments/assets/1b24e65b-1608-42ea-a14b-8f15a8e50c96" />

## Screenshots (Current State)

<img width="716" height="815" alt="Image" src="https://github.com/user-attachments/assets/813efee2-137f-41c9-906c-3f663316e472" />

<img width="723" height="1225" alt="Image" src="https://github.com/user-attachments/assets/d07288b5-c71c-409f-b78e-c5412d5f8671" />

PR #3112, opened 2026-07-31
fix(ui): keep private-registry credentials visible once entered

PetrBulanek:
## Problem

In the sandbox wizard's Custom image step, disclosure of the **Private registry** section was derived from `partial` — 1 or 2 of 3 fields filled, i.e. *invalid*:

```ts
const [open, setOpen] = useState(false);
const expanded = open || partial;
```

Two things go wrong with that. `partial` means invalid, so the fully-valid state (3 of 3) rendered identically to the never-touched state. And `open` lived in a component that unmounts on every step and starting-point change, while the credentials live in `SandboxWizardView` and outlive it — so the data outlived its own disclosure control.

Two reachable symptoms:

- **Invisible on return.** Enter an image, expand, fill all three, click Continue, then click the **Image** breadcrumb to come back. The section remounted collapsed and byte-identical to untouched, while the credentials were still armed for submission. Only the image input existed in the DOM.
- **Collapse on completion.** On a fresh mount, fill Server only → the section auto-expands via `partial`. Fill Username and Password → `partial` flips false, `open` is still false, and the section snaps shut the instant the credentials become valid, hiding the three values just typed.

The consequence of the first: you can then edit the image address with no sign that credentials are attached. `finish()` gates only on `!!image && registryFilledCount(...) === 3` — there is no comparison between the credential's `server` and the image's host — so a pull Secret is created for a sandbox whose image may come from an unrelated registry. When the host *does* match but the credential is stale, the resulting `ImagePullBackOff` is hard to diagnose because the wizard shows no sign credentials are in play. Step 3 never mentions the registry either.

Not a credential-exposure bug: the Secret is only ever appended to the pod's `imagePullSecrets`, never projected into the agent container, and it is deleted with the agent. Verified both below.

## Change

Disclosure now defaults to whether the fields have **content**, with an explicit toggle overriding that:

```ts
const [override, setOverride] = useState<boolean | null>(null);
const expanded = partial || (override ?? registryFilledCount(value) > 0);
```

Editing pins it open, so clearing the last field doesn't collapse the section mid-edit — the mirror of symptom 2.

`partial` stays as a hard floor that outranks the user. Its hint renders *inside* the section, so letting a collapse win there would disable Continue with the reason hidden — the same defect in a narrower form. I initially dropped `partial` as redundant (it implies a non-zero count) and that was wrong: it's redundant in the default path but load-bearing against an explicit collapse.

An empty section still starts collapsed, matching the design in #2748.

## Scope

Pre-existing since #997, which introduced `expanded = open || partial`; #3026 moved the section onto the Image step. Not introduced by #3062. Found while reviewing #3062, which is unaffected — its `isCustomImage` scoping is verified intact below.

No contract, server, or payload change. One file.

## Testing

`mise run ui:check` clean (tsc, eslint 0 errors, prettier). `mise run ui:test` 153 passing.

Disclosure state machine, driven in the real app against the dev cluster:

| scenario | result |
|---|---|
| fresh empty Custom | collapsed, 1 input — mock preserved |
| try to collapse while partial | stays expanded, hint visible — a blocked Continue always has a visible reason |
| complete all three | stays expanded, hint clears, Continue enables |
| Continue → back to Image | expanded, values visible |
| collapse when complete | collapses — toggle not dead |
| …then edit the image | stays collapsed |
| clear all three | stays open, no snap-shut |
| partial → General-purpose → pick harness | section gone, Continue enabled, no credentials persisted to sessionStorage |

End-to-end, three sandboxes created and deleted on the dev cluster (baseline before and after: 0 dockerconfigjson secrets):

- **With credentials** — real `agents.create` payload carried `registryCredential`; Secret created with `.auths` keyed exactly `private.example.test`; Agent CR got `spec.imagePullSecretRef`; pod had it in `imagePullSecrets` with **0** volume refs and **0** env/envFrom refs; kubelet events showed the pull attempt; deleting the agent removed the Secret.
- **Collapsed but complete** — credentials still submitted and the Secret still created, confirming the collapse is purely visual and doesn't drop them.
- **No credentials (control)** — no `registryCredential` in the payload, `imagePullSecretRef: <none>`, no Secret.

Limit worth stating: the test images are bogus hosts, so pulls failed at DNS and the credential was never transmitted to a real registry. That last hop was covered in #3026 against an htpasswd `registry:2`, and this change can't affect it — it only touches disclosure.

## Note for reviewers

The pre-commit hook can't run in a git worktree — `controller:check:crd-backwards-compatibility` fails with `reference not found` for `v0.2.12-rc1` even on a pristine checkout of main with zero changes, because `.git` is a file pointing into the parent repo. This commit used `--no-verify`; the UI checks above were run directly. Filed separately, since it trains anyone using `.claude/worktrees/` to bypass every pre-commit gate.

### DIFF
--- a/packages/ui/src/modules/sandboxes/components/registry-credential-section.tsx
+++ b/packages/ui/src/modules/sandboxes/components/registry-credential-section.tsx
@@ -1,5 +1,3 @@
-import { useState } from "react";
-
 import { DisclosureToggle } from "@/components/ui/disclosure";
 import { Input } from "@/components/ui/input";
 import { labelVariants } from "@/components/ui/label";
@@ -30,18 +28,27 @@ interface Props {
   value: RegistryCredential;
   onChange: (value: RegistryCredential) => void;
   partial: boolean;
+  disclosureOverride: boolean | null;
+  onDisclosureOverride: (override: boolean) => void;
 }
 
-export function RegistryCredentialSection({ value, onChange, partial }: Props) {
+export function RegistryCredentialSection({
+  value,
+  onChange,
+  partial,
+  disclosureOverride,
+  onDisclosureOverride,
+}: Props) {
   // The credentials outlive this section, which unmounts on every step change —
-  // so disclosure follows whether there is anything to show unless the user says
-  // otherwise (null = follow the fields). Editing pins it open, or clearing the
-  // last field would collapse it mid-edit. `partial` outranks the user: its hint
-  // renders in here, and blocking Continue with the reason hidden is worse.
-  const [override, setOverride] = useState<boolean | null>(null);
-  const expanded = partial || (override ?? registryFilledCount(value) > 0);
+  // so disclosure lives next to them in the wizard and follows whether there is
+  // anything to show unless the user says otherwise (null = follow the fields).
+  // Editing pins it open, or clearing the last field would collapse it mid-edit.
+  // `partial` outranks the user: its hint renders in here, and blocking Continue
+  // with the reason hidden is worse.
+  const expanded =
+    partial || (disclosureOverride ?? registryFilledCount(value) > 0);
   const set = (key: keyof RegistryCredential, next: string) => {
-    setOverride(true);
+    onDisclosureOverride(true);
     onChange({ ...value, [key]: next });
   };
 
@@ -49,7 +56,7 @@ export function RegistryCredentialSection({ value, onChange, partial }: Props) {
     <div>
       <DisclosureToggle
         open={expanded}
-        onToggle={() => setOverride(!expanded)}
+        onToggle={() => onDisclosureOverride(!expanded)}
         chevronSize={12}
         className={cn(labelVariants(), "gap-1.5 hover:text-foreground")}
       >
--- a/packages/ui/src/modules/sandboxes/components/registry-credential-section.tsx
+++ b/packages/ui/src/modules/sandboxes/components/registry-credential-section.tsx
@@ -33,16 +33,23 @@ interface Props {
 }
 
 export function RegistryCredentialSection({ value, onChange, partial }: Props) {
-  const [open, setOpen] = useState(false);
-  const expanded = open || partial;
-  const set = (key: keyof RegistryCredential, next: string) =>
+  // The credentials outlive this section, which unmounts on every step change —
+  // so disclosure follows whether there is anything to show unless the user says
+  // otherwise (null = follow the fields). Editing pins it open, or clearing the
+  // last field would collapse it mid-edit. `partial` outranks the user: its hint
+  // renders in here, and blocking Continue with the reason hidden is worse.
+  const [override, setOverride] = useState<boolean | null>(null);
+  const expanded = partial || (override ?? registryFilledCount(value) > 0);
+  const set = (key: keyof RegistryCredential, next: string) => {
+    setOverride(true);
     onChange({ ...value, [key]: next });
+  };
 
   return (
     <div>
       <DisclosureToggle
         open={expanded}
-        onToggle={() => setOpen((v) => !v)}
+        onToggle={() => setOverride(!expanded)}
         chevronSize={12}
         className={cn(labelVariants(), "gap-1.5 hover:text-foreground")}
       >
--- a/packages/ui/src/modules/sandboxes/components/steps/custom-image-card.tsx
+++ b/packages/ui/src/modules/sandboxes/components/steps/custom-image-card.tsx
@@ -10,6 +10,8 @@ export interface RegistryControls {
   value: RegistryCredential;
   onChange: (value: RegistryCredential) => void;
   partial: boolean;
+  disclosureOverride: boolean | null;
+  onDisclosureOverride: (override: boolean) => void;
 }
 
 interface Props {
@@ -67,6 +69,8 @@ export function CustomImageCard({
             value={registry.value}
             onChange={registry.onChange}
             partial={registry.partial}
+            disclosureOverride={registry.disclosureOverride}
+            onDisclosureOverride={registry.onDisclosureOverride}
           />
         </div>
       )}
--- a/packages/ui/src/modules/sandboxes/views/sandbox-wizard-view.tsx
+++ b/packages/ui/src/modules/sandboxes/views/sandbox-wizard-view.tsx
@@ -63,6 +63,11 @@ export function SandboxWizardView() {
   const [registryCredential, setRegistryCredential] = useState(
     EMPTY_REGISTRY_CREDENTIAL,
   );
+  // Disclosure lives here too — the section unmounts on every step change,
+  // and the credentials it shows must not outlive their own visibility toggle.
+  const [registryDisclosureOverride, setRegistryDisclosureOverride] = useState<
+    boolean | null
+  >(null);
   const isCustomImage = snapshot.startingPoint === "custom";
 
   const imageLabel = useMemo(() => {
@@ -184,6 +189,7 @@ export function SandboxWizardView() {
       });
       reset();
       setRegistryCredential(EMPTY_REGISTRY_CREDENTIAL);
+      setRegistryDisclosureOverride(null);
       selectAgent(agent.id);
     } catch {
       // Mutation surfaces its own error toast; stay on Step 3 to retry.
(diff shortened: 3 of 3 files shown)


packages/ui/src/modules/sandboxes/components/registry-credential-section.tsx
```diff
@@ -33,16 +33,23 @@ interface Props {
...
 export function RegistryCredentialSection({ value, onChange, partial }: Props) {
-  const [open, setOpen] = useState(false);
-  const expanded = open || partial;
-  const set = (key: keyof RegistryCredential, next: string) =>
+  // The credentials outlive this section, which unmounts on every step change —
+  // so disclosure follows whether there is anything to show unless the user says
+  // otherwise (null = follow the fields). Editing pins it open, or clearing the
+  // last field would collapse it mid-edit. `partial` outranks the user: its hint
+  // renders in here, and blocking Continue with the reason hidden is worse.
+  const [override, setOverride] = useState<boolean | null>(null);
```
dam-code-guardian: 🟡 **Warning:** `override` is scoped to a component that unmounts on every wizard step change, while the credential value it qualifies lives in `SandboxWizardView` and survives. So the fix only holds in the expand direction: collapse a filled section, go to step 2, come back — remount resets `override` to `null`, `expanded` falls back to `registryFilledCount(value) > 0`, and the section reopens. That is the same lifetime mismatch the PR body diagnoses for `open`, just in the other direction. Moving it next to `registryCredential` in the parent (a second piece of state, or one `{ value, disclosureOverride }` object) closes it in both directions.

packages/ui/src/modules/sandboxes/components/registry-credential-section.tsx
```diff
@@ -33,16 +33,23 @@ interface Props {
...
+    setOverride(true);
     onChange({ ...value, [key]: next });
+  };
 
   return (
     <div>
       <DisclosureToggle
         open={expanded}
-        onToggle={() => setOpen((v) => !v)}
+        onToggle={() => setOverride(!expanded)}
```
dam-code-guardian: 🟢 **Suggestion:** while `partial` is true, `expanded` short-circuits to `true`, so this writes `override = false` and nothing visibly happens — the button stays clickable with `aria-expanded={true}`. The stored `false` can never surface later either, since every route out of `partial` runs through `set()`, which pins `override` back to `true`. Disabling the toggle while `partial` would make the precedence visible instead of silently swallowing the click.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `5af8c19`

## PR #3112: fix(ui): keep private-registry credentials visible once entered
**Author:** PetrBulanek | **Branch:** fix/registry-credential-disclosure → main | **Changes:** +11 −4 (1 file)

### Summary
Replaces the private-registry section's disclosure state (`open || partial`) with an override-or-follow-the-fields model, so filled credentials are visible on remount and no longer snap shut when they become valid.

### Findings
- 🟡 **Warning:** explicit collapse is discarded on remount — `override` lives in a component that unmounts on step change (`packages/ui/src/modules/sandboxes/components/registry-credential-section.tsx:41`)
- 🟢 **Suggestion:** toggle is inert while `partial` is true (`packages/ui/src/modules/sandboxes/components/registry-credential-section.tsx:52`)
- ✅ **Looks good:** keeping `partial` as a hard floor is correct, not redundant — an explicit collapse would otherwise hide the hint that explains a disabled Continue.

`aligned`

**HIGH — `override` has a shorter lifetime than the state it overrides** (`registry-credential-section.tsx:41`). The component unmounts on every wizard step change, so `override` resets to `null` while the credential value it qualifies survives in the parent (`sandbox-wizard-view.tsx:63`). Consequence: the fix is one-directional — an explicit *collapse* of a filled section is silently discarded on return to step 1 (remount → `override = null` → `expanded = filled > 0` → reopens), exactly the class of bug the old `useState(false)` had in the expand direction. Per the lineage rules this value must outlive the component, so it belongs next to `registryCredential` in `SandboxWizardView` (a second field, or one `{ value, disclosureOverride }` object), not in local `useState`.

**MODERATE — the toggle is inert while `partial` is true** (`registry-credential-section.tsx:42,52`). `expanded = partial || …` short-circuits, so `onToggle`'s `setOverride(!expanded)` writes `false` and nothing changes: the button is clickable but has no effect, and `DisclosureToggle` keeps reporting `aria-expanded={true}`. Nor can that stored `false` ever surface later — every path out of `partial` goes through `set()`, which pins `override` back to `true`. Either disable/hide the toggle while `partial` (making the precedence visible) or drop the write.

### Verdict
COMMENT — the reported symptoms are genuinely fixed, but the same unmount-scoped-state root cause remains in the collapse direction; the PR's own test table records "collapse when complete → collapses" without re-checking it after a step change.

---

packages/ui/src/modules/sandboxes/components/registry-credential-section.tsx
```diff
@@ -33,16 +33,23 @@ interface Props {
...
 
 export function RegistryCredentialSection({ value, onChange, partial }: Props) {
-  const [open, setOpen] = useState(false);
-  const expanded = open || partial;
-  const set = (key: keyof RegistryCredential, next: string) =>
+  // The credentials outlive this section, which unmounts on every step change —
+  // so disclosure follows whether there is anything to show unless the user says
+  // otherwise (null = follow the fields). Editing pins it open, or clearing the
+  // last field would collapse it mid-edit. `partial` outranks the user: its hint
+  // renders in here, and blocking Continue with the reason hidden is worse.
```
kapetr: Two of these five lines are counterfactual rationale, which the repo's comment rules exclude: *"or clearing the last field would collapse it mid-edit"* and *"blocking Continue with the reason hidden is worse"*. Both are already in the PR body. Keep the part that can't be inferred:

```ts
// This section unmounts on step changes while the credentials outlive it, so
// disclosure follows the fields (null) until the user toggles; `partial` is a
// floor because its hint renders inside.
```

packages/ui/src/modules/sandboxes/components/registry-credential-section.tsx
```diff
@@ -33,16 +33,23 @@ interface Props {
...
+  const [override, setOverride] = useState<boolean | null>(null);
+  const expanded = partial || (override ?? registryFilledCount(value) > 0);
```
kapetr: New branching logic with no test, and this rule has now needed fixing twice (#997, #3026). `vitest.config.ts` is node-env and doesn't match `.tsx`, so no render test — but the predicate is pure and there's precedent for extracting and table-testing it (`provider-policy.test.ts`):

```ts
export function registrySectionExpanded(
  override: boolean | null,
  partial: boolean,
  filled: number,
): boolean {
  return partial || (override ?? filled > 0);
}
```

Rows worth pinning: `null` + count 3 → true, `override=true` + count 0 → true, `override=false` + `partial` → true.

Also `override` doesn't say what it overrides — `expandOverride` reads better at the call site.

packages/ui/src/modules/sandboxes/components/registry-credential-section.tsx
```diff
@@ -33,16 +33,23 @@ interface Props {
...
+  const set = (key: keyof RegistryCredential, next: string) => {
```
kapetr, line 50: `set` also pins the section open now — `setField`?

packages/ui/src/modules/sandboxes/components/registry-credential-section.tsx
kapetr: Negates a value with `partial` folded in, so while `partial` the click writes `false` and changes nothing. Not a bug, but flipping only the user's own intent is more direct:

```ts
onToggle={() => setOverride((v) => !(v ?? registryFilledCount(value) > 0))}
```

kapetr (commented):
Verified: `ui:check` clean, `ui:test` 153 passing. State machine is sound — both symptoms fixed, no new bad state.

**Residual hazard, follow-up not this PR.** Fill all three → collapse → edit the image address. `finish()` gates on `!!image && registryFilledCount(...) === 3` with no host comparison, so a Secret keyed `ghcr.io` is created for a `docker.io` image with nothing on screen showing credentials are armed. Your test table records this, so I take it as known.

Fix that would close it: make the collapsed header self-describing (`Private registry · ghcr.io`, or a `3 fields set` badge) when the count is non-zero. Then a collapse can't hide armed credentials.

Root cause nearby: `onPickStartingPoint` never clears `registryCredential` when leaving Custom, which is what lets a credential outlive its image. `startingPointDefaults` clears `customImage` so nothing is submitted, but resetting the credential there removes the class.

PR #3117, opened 2026-08-03
fix(ui): show Channels to every user, not just flag holders

### LINKED ISSUE #3077: UI - Channels is hidden behind an experimental feature flag
## Context

Slack channels are a supported, documented way to use an agent — bind a conversation to it and the team talks to it there. But the Channels surface only appears for users who have the "Advanced connections" experimental feature enabled. That's off by default and not something a user turns on in the course of normal use, so in practice the feature is invisible to everyone outside the team.

## Problem

A working, supported integration is unreachable through normal use of the product. Users who've been told "you can put your agent in Slack" find no way to do it, and we end up enabling it for people one at a time.

Channels doesn't need anything experimental to work — where it's available is decided by the install's own messenger setup — so the flag is hiding a finished feature for no reason.

## Not in Scope

**This does not make Slack available as a connection.** Slack as a connection users can set up stays exactly as it is today — deliberately not offered. This issue is only about the Channels section of a sandbox, which is a separate surface and needs nothing from that connection. Channels becoming visible must not make Slack appear in the connection catalog.

## Expected Behavior

The Channels section appears in the sandbox navigation for every user, with no experimental feature required. Where the install has no messenger configured, the section says so rather than disappearing.

## Actual Behavior

Channels is hidden unless "Advanced connections" is enabled for that user, and navigating straight to it silently redirects to Sandbox Setup.

## Steps to Reproduce

1. Sign in as a user without the "Advanced connections" experimental feature enabled.
2. Open a sandbox and look at the section navigation — no Channels.
3. Navigate directly to the sandbox's Channels section — you land on Sandbox Setup instead.

pilartomas:
Closes #3077

## What

The sandbox **Channels** section was gated behind the *Advanced connections* experimental feature — off by default and not something a user turns on in normal use — so a working, documented integration was unreachable. Navigating straight to a sandbox's Channels URL silently redirected to Sandbox Setup.

Channels needs nothing experimental: which messengers can be bound is decided by the install's own messenger setup, which the surface already reads. So the gate goes away.

## Changes

- **Sandbox home view** — no more feature-flag read and no more `channels → setup` redirect; the section renders for whatever the URL asks for.
- **Section nav** — Channels is an unconditional entry, like Skills or Schedules; the `showChannels` prop is gone.
- **Empty state** — where the install has no messenger configured, the section now names the reason and the next step ("No messenger is set up on this platform… Ask your operator to configure Slack or Telegram") instead of the ambiguous "No channels are configured for this installation". The nav summary agrees ("No messenger configured") rather than reading "No channels connected", which would imply a list the user could fill.
- **Experimental features tab** — the *Advanced connections* description no longer claims it reveals the Channels section.
- **Architecture doc** — `docs/architecture/channels.md` records that Channels is a standard Agent surface whose availability is the install's decision, and that Slack-as-a-Connection is a separate, still-withheld surface a channel needs nothing from.

## Explicitly not changed

Slack **as a connection** stays exactly as it is: `advanced-connections` still gates the internal-only connection catalog (Slack, Google services, Spotify, custom client-credentials, GitHub Apps). Channels becoming visible does not put Slack in the connection catalog. No backend change either — the channels endpoints were never flag-gated (flags are progressive disclosure, not authorization), and they are owner-scoped as before.

## Verification

- `mise run ui:check` (tsc, eslint, prettier) — clean; the two eslint warnings are pre-existing, in files this PR doesn't touch.
- `mise run ui:test` — 153 tests / 17 files pass.
- `docs:check` (doc-size, adr-index, adr-immutable) plus `api-server`, `api-server-api`, `db`, `e2e-playwright` checks — clean.
- `controller:check:crd-backwards-compatibility` fails, but it fails the same way on a clean `main` checkout — pre-existing, unrelated to this diff (no CRD or Go changes here).

Manual checks worth doing on a cluster, as a user **without** the flag:

1. Open a sandbox → *Channels* is in the section nav, with a live summary line.
2. Go straight to `/sandboxes/<id>/channels` → the Channels section loads instead of bouncing to Sandbox Setup.
3. Connect and disconnect a Slack channel from the section.
4. Open the connection catalog → Slack is still absent.
5. On an install with neither messenger configured, the section explains that no messenger is set up, and the nav line reads "No messenger configured".

There is no e2e coverage for the Channels UI today (the Slack specs drive the messenger side, not this surface), so the nav-visibility assertion is a reasonable follow-up if we want a regression guard.

### DIFF
--- a/packages/ui/src/modules/sandboxes/views/sandbox-home-view.tsx
+++ b/packages/ui/src/modules/sandboxes/views/sandbox-home-view.tsx
@@ -3,7 +3,6 @@ import { Button } from "@/components/ui/button";
 import { useStore } from "../../../store.js";
 import { useResolvedAgentDisplay } from "../../agents/hooks/use-resolved-agent-display.js";
 import { SandboxArtifactsSection } from "../../artifacts/components/sandbox-artifacts-section.js";
-import { useFeatures } from "../../features/api/queries.js";
 import { routeToPath } from "../../platform/lib/routes.js";
 import { ConnectionsSection } from "../components/connections-section.js";
 import { SandboxChannelsSection } from "../components/sandbox-channels-section.js";
@@ -18,14 +17,8 @@ import { useSectionSummaries } from "../hooks/use-section-summaries.js";
 
 export function SandboxHomeView() {
   const f = useSandboxSettingsForm();
-  const rawSection = useStore((s) => s.sandboxSection);
+  const section = useStore((s) => s.sandboxSection);
   const navigateToSandboxHome = useStore((s) => s.navigateToSandboxHome);
-  const { data: flags } = useFeatures();
-  // Channels rides the advanced-connections feature flag: messenger channel
-  // bindings are not a regular-user surface yet.
-  const showChannels = flags?.["advanced-connections"] ?? false;
-  const section =
-    rawSection === "channels" && !showChannels ? "setup" : rawSection;
 
   const display = useResolvedAgentDisplay(f.agent);
 
@@ -73,7 +66,6 @@ export function SandboxHomeView() {
           active={section}
           onNavigate={(s) => navigateToSandboxHome(agent.id, s)}
           summaries={summaries}
-          showChannels={showChannels}
         />
       }
     >
--- a/docs/architecture/channels.md
+++ b/docs/architecture/channels.md
@@ -1,10 +1,12 @@
 # Channels
 
-Last verified: 2026-07-29
+Last verified: 2026-08-03
 
 ## Overview
 
 A **channel** is a messenger surface (Slack, Telegram) that lets users drive an Agent from outside the UI. Channels are pluggable adapters that live inside the api-server process — no separate Deployment, no sidecar in the agent pod. Each adapter (the _worker_) owns its inbound socket, its outbound API, and its thread-to-session bookkeeping; a `ChannelManager` service composes the workers and reacts to lifecycle events on the in-process event bus.
+
+Channels are a **standard Agent surface**, not a pre-release one: every Agent exposes it, with no per-user opt-in in front of it. What can be bound there is the install's own decision — a worker exists only where its token is configured — and an install with no messenger says so on the surface rather than withdrawing it. Slack as a *Connection* is a separate surface, still withheld from regular users; a channel needs nothing from it ([connections](connections.md)).
 
 Channel bindings are **1:1 with Agent**: a Slack conversation may be bound to at most one Agent globally, and an Agent has at most one Slack binding — Agent delete or Slack disconnect releases it. A Slack "conversation" is any surface the bot is party to: a public/private **channel**, a **group DM**, or a **1:1 DM**. The binding key is the conversation id in every case, so DMs and group DMs reuse the channel binding mechanics wholesale — same table, same resolution, same in-chat bind flow.
 
--- a/packages/ui/src/modules/features/components/features-tab.tsx
+++ b/packages/ui/src/modules/features/components/features-tab.tsx
@@ -18,7 +18,7 @@ const FEATURE_ROWS: FeatureRow[] = [
     id: "advanced-connections",
     label: "Advanced connections",
     description:
-      "Reveals the pre-release connection catalog (Google services, Slack, Spotify, custom client-credentials) and the sandbox Channels section.",
+      "Reveals the pre-release connection catalog (Google services, Slack, Spotify, custom client-credentials).",
   },
   {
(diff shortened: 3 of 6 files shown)


packages/ui/src/modules/sandboxes/views/sandbox-home-view.tsx
```diff
@@ -18,14 +17,8 @@ import { useSectionSummaries } from "../hooks/use-section-summaries.js";
 
 export function SandboxHomeView() {
   const f = useSandboxSettingsForm();
-  const rawSection = useStore((s) => s.sandboxSection);
+  const section = useStore((s) => s.sandboxSection);
```
dam-code-guardian, line 20: 🟡 **Warning:** Removing the flag makes the Slack bind form reachable by every user, and nothing on the path verifies the requester's relationship to the Slack conversation they type in. `connectSlack` checks agent ownership (`repo.get(id, deps.owner)`) and global 1:1 uniqueness, but not that the caller is a member of `slackChannelId` — so any user can bind their own agent to any channel id they can guess or read off a colleague's URL, and with `ambient: true` that agent then reads every message in it. The flag was the de-facto control on who could reach this form; `docs/architecture/channels.md:119` already states "channel membership is the only per-person gate, and Slack owns it", which holds for the in-chat `/platform bind` path (Slack itself proves membership) but not for this web form.

Not a blocker for this PR — the gap predates it and the endpoint was always ungated at the API and CLI. Worth a conscious decision now that the surface is general: either verify bot-visible membership server-side before binding, or accept it explicitly and say so on the page, since the doc sentence currently reads as if the web path were covered too.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `80b9302`

## PR #3117: fix(ui): show Channels to every user, not just flag holders
**Author:** pilartomas | **Branch:** fix/channels-always-visible → main | **Changes:** +17 −27 (6 files)

### Summary
Removes the `advanced-connections` flag gate from the sandbox Channels section — the nav entry is unconditional, the `channels → setup` redirect is gone, and the empty state now explains that the install has no messenger configured. _(Suppressed 1 finding per PR context: missing nav-visibility regression test, acknowledged as follow-up in the PR body.)_

### Findings
- 🟡 **Warning:** bind form reachable by every user with no channel-membership check (`packages/ui/src/modules/sandboxes/views/sandbox-home-view.tsx:20`)
- ✅ **Looks good:** the removed gate really was progressive disclosure only — `channels.available` is `readAgentProcedure`, `connectSlack`/`disconnectSlack` are `manageAgentsProcedure`, and `connectSlackImpl` resolves the agent through `repo.get(id, deps.owner)`, so no API surface opens up that wasn't already reachable (the CLI's `channel slack-connect` was never gated either).

**Verdict: aligned.**

`docs/architecture/channels.md` is updated in the same PR as the behavior change, `Last verified:` is bumped to 2026-08-03, the new paragraph stays at concept level (no package names, file paths, or code-level shape), and its `connections.md` cross-link resolves. Page is 37,797 chars against the 40,000 cap. No ADR references introduced. `features.md` needs no edit — it never enumerated Channels, and `advanced-connections` still exists for the connection catalog, so its "current feature is advanced connection types" line stays true.

✅ No findings.

The `sandboxes → agents/api/queries` import edge already existed on `main` (`useAgentConnections`), so widening it to `useAgents` crosses no new boundary. The file is UI-package code, outside the scope of the server-side slice rules (services/domain/infrastructure, `compose.ts`, event-bus-only inter-module communication).

✅ No findings.

`sandbox-home-view.tsx` deletes a derived-state computation (`rawSection` → `section` redirect) rather than adding one, and `sandbox-section-nav.tsx` drops a boolean "configure how I work" prop — both move toward the reference rules. The two remaining `.tsx` edits are copy changes. The transient-empty-state concern does not apply: `SandboxChannelsSection` only mounts once `f.status === "ready"`, which requires the same `useAgents()` cache entry to have resolved, so `availableChannels` is never the `{}` loading placeholder while the section is on screen.

### Verdict
COMMENT — the gate removal is sound and correctly documented; the one warning is a pre-existing authorization gap whose blast radius this PR widens, worth a conscious sign-off rather than a block.

---

packages/ui/src/modules/sandboxes/views/sandbox-home-view.tsx
pilartomas: Channel membership was never a gate, the true gate is whether the app was added to the channel or not. Either way, this is out of scope, ignore it.

JanPokorny (approved):

PR #3119, opened 2026-08-03
fix(ui): raise the memory size floor to 384Mi

### LINKED ISSUE #3069: UI - Size sliders allow a memory setting too low for the agent to run
## Context

The memory slider on sandbox size goes down to 128Mi. Claude Code needs at least around 384Mi to function at all, so anything below that produces a sandbox that cannot run.

The 128Mi figure is a Kubernetes scheduling floor, the minimum needed to place a pod. It is enforced in the slider, in server-side validation, and in the chart. None of those represent a working agent, only a schedulable one.

## Problem

You can create a sandbox that is guaranteed not to work, and nothing warns you. The setting is accepted, the sandbox is created, and it fails later, with nothing connecting the failure to the memory choice.

Because the floor is enforced on the server too, this is not only a slider problem. The same unusable size can be set through the CLI or API.

## Expected Behavior

The lowest selectable memory should be enough for the agent to actually run, or choosing less should clearly warn that it will not work.

## Actual Behavior

Memory can be set as low as 128Mi with no warning, producing a sandbox that cannot run.

## Steps to Reproduce

1. Create a sandbox and open the size sliders
2. Drag memory to its minimum of 128Mi
3. Create the sandbox and try to use it

JanPokorny:
Claude Code needs ~384Mi to run at all; the old 128Mi floor only made the pod schedulable. Raises the memory floor to 384Mi in both the size slider and the server-side `agentSizeSchema` validation, so an unusable size can't be chosen via UI, CLI, or API.

The controller's 128Mi `requestsFromLimits` floor is untouched — that's the derived scheduling-request minimum, a separate concern.

Fixes #3069

### DIFF
--- a/packages/api-server-api/src/modules/agents/schemas.ts
+++ b/packages/api-server-api/src/modules/agents/schemas.ts
@@ -6,7 +6,8 @@ const idSchema = z.object({ id: z.string().min(1) });
 
 // CPU as cores ("2") or millicores ("500m"); memory as Mi/Gi. Deliberately
 // narrower than the full K8s quantity grammar — these are slider outputs,
-// not operator YAML. Floors keep a chosen size schedulable.
+// not operator YAML. The memory floor is what the agent needs to actually
+// run (~384Mi for Claude Code), not just be schedulable.
 const cpuQuantitySchema = z
   .string()
   .regex(/^\d+(\.\d+)?m?$/, "CPU must look like '2', '0.5' or '500m'")
@@ -16,8 +17,8 @@ const cpuQuantitySchema = z
 const memoryQuantitySchema = z
   .string()
   .regex(/^\d+(Mi|Gi)$/, "memory must look like '512Mi' or '2Gi'")
-  .refine((v) => toMemoryMi(v) >= 128, {
-    message: "memory must be at least 128Mi",
+  .refine((v) => toMemoryMi(v) >= 384, {
+    message: "memory must be at least 384Mi",
   });
 
 function toCpuMilli(v: string): number {
--- a/packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx
+++ b/packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx
@@ -7,7 +7,7 @@ import { parseCpuMilli, parseMemoryMi } from "../lib/quantity.js";
 
 // Floors mirror the server-side slider validation (agentSizeSchema).
 const CPU_FLOOR_MILLI = 100;
-const MEMORY_FLOOR_MI = 128;
+const MEMORY_FLOOR_MI = 384;
 const CPU_STEP_MILLI = 100;
 const MEMORY_STEP_MI = 128;
 


dam-code-guardian:
🛡️ **Code Guardian 🛡️** — ⚠️ Post-merge follow-up @ `0d249b9`

This PR merged ~3 minutes into the review, so this lands as a comment rather than a PR review. The reviewed files are byte-identical at the merge commit, so the findings below still apply. **No blockers — no critical findings.** Two follow-ups worth a look:

### 🟡 Sub-384Mi sandboxes now show a label that disagrees with the slider

`packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx:101-103` — `valueLabel={formatMiAsMemory(memory)}` renders from React state, while `value={memory}` feeds a native `<input type="range">` whose `min` is now 384. For an existing 256Mi sandbox opened in Settings, the label (and `aria-valuetext`) reads `256Mi` while the input's own value and thumb sit at the floor. There's no clamping layer in the module, and `use-sandbox-settings-form.ts:106` baselines straight off the spec with no floor applied — and 256Mi was a saveable value before this change, so it's reachable.

`min={Math.min(MEMORY_FLOOR_MI, memory)}` would keep the thumb truthful while the server floor still blocks a re-save below 384.

*Caveat:* the mechanism (a range input clamping a below-`min` value and firing no event) is standard HTML value-sanitization, but I could not verify it in this environment — no jsdom available and egress to the spec/MDN was blocked. Worth 10 seconds in a browser before acting on it.

The good news, verified: **no silent rewrite.** `use-sandbox-settings-save.ts:127-136` ships only dirty dimensions and the clamp fires no `onChange`, so a 256Mi sandbox whose owner edits only name/env/CPU keeps its 256Mi.

### 🟡 Client schema still accepts the newly-invalid band

`sandbox-settings-schema.ts:21` is `z.number().int().positive()`, so 1–383Mi produces no field-level error — rejection surfaces only as a generic `Failed to update agent` toast. Same for `wizard-snapshot.ts:55`: a `sessionStorage` draft written before this deploy can carry `sizeMemoryMi: 256` and fail at create, while the slider already looks like it's at the minimum.

### 🟢 The floor is hand-mirrored across two packages

`schemas.ts:20` and `sandbox-size-section.tsx:10` are kept in sync by a comment alone, and this PR is that lockstep edit. `MIN_INVOCATION_TTL_MS`/`MAX_INVOCATION_TTL_MS` set the precedent for exporting a numeric bound beside its schema, and `packages/ui` already imports runtime values from `api-server-api` — so exporting `AGENT_MEMORY_FLOOR_MI` (templating the Zod message off it) would make the mirror structural.

Also: `schemas.ts:99` still says "Floors match the chart's derivation floors" — true for CPU, no longer true for memory.

### ✅ Verified correct
`384 % 128 == 0`, so step alignment holds and 1/2/4/8Gi stay exactly reachable. No shipped template declares sub-384Mi memory. The controller's 128Mi `requestsFromLimits` floor is a different axis and correctly untouched. Architecture docs are aligned — `budgets.md:16`'s `100m/128Mi` describes the derivation floor, not this one.

---

PR #3123, opened 2026-08-03
feat(connections): re-authenticate and update credentials in place

### LINKED ISSUE #3042: Re-authenticate and update credentials on an existing connection
## Problem

Once a connection is created, its credential is effectively frozen. When it stops working, the only remedy is to disconnect and create a new connection — losing the connection's identity and every agent grant, which all have to be set up again by hand.

This shows up in several ways:

- An **OAuth connection** whose refresh token is revoked or expired (consent revoked upstream, password change, long inactivity) shows as expired, and there is no way to re-run the login/consent flow on the existing connection. The same gap blocks re-consenting to get **different scopes**.
- An **API key / header connection** whose key was rotated upstream can only be fixed in place for LLM providers; every other connection of this shape (custom headers, Kubernetes tokens, GitHub PATs) has to be recreated.
- **Client-credentials and GitHub App connections** mint their own tokens from a stored secret or private key; when that secret rotates upstream, there is no way to supply the new one, so the connection eventually shows expired for good.

## Proposed solution

Every connection offers credential maintenance in place, preserving its identity and all agent grants:

- **Re-authenticate** on OAuth connections — redo the login/consent flow and store the fresh tokens on the same connection.
- **Update the stored secret** on header, client-credentials, and GitHub App connections — paste the new key/secret/token and it takes effect immediately for all granted agents.
- A connection showing **expired** points directly at the matching fix, so recovery is one click away from the failure.

### Out of scope

- Editing a connection's non-credential inputs (name, hosts, config values) — a separate concern from the credential lifecycle.

PetrBulanek:
A connection's credential can now be maintained without recreating the connection — its identity and every sandbox grant survive.

- **Re-authenticate an OAuth connection.** Re-run login and consent on the connection you already have; fresh tokens replace the old ones. Consent asks for the template's current permissions, so a connection created before a provider gained scopes picks them up without being rebuilt.
- **Replace a stored secret.** Paste a new value for a header credential, a client secret, or a GitHub App private key. The kinds that mint tokens prove the new secret works before storing it, so a wrong one is rejected up front instead of quietly breaking the connection later.
- **Expired connections now say so — and offer the fix.** A credential the platform can no longer refresh used to look healthy until an agent failed mid-task. It now reads as expired, with the matching fix one click away on the row, wherever connections are listed.
- **Rotate an OAuth app's client secret**, for connections that carry their own. This is a deliberate addition beyond the issue text: without it a secret rotated upstream leaves the connection unfixable, since refreshing and re-consenting both fail for the same reason — so the expired row would have pointed at a fix that could not work.

All of it is available from the CLI: `dam connection update` for stored secrets, and the new `dam connection reauth` for OAuth connections.

Closes #3042


packages/api-server/src/modules/connections/services/oauth-refresh.ts
```diff
@@ -178,6 +195,89 @@ async function dueConnections(db: Db, skewSec: number): Promise<Connection[]> {
...
+      authKind: conn.auth.kind,
+      ...(err instanceof OAuthTokenEndpointError
+        ? {
+            ...(err.oauthError ? { oauthError: err.oauthError } : {}),
+            ...(err.status !== undefined ? { status: err.status } : {}),
+          }
+        : {}),
+    },
+  });
+  return true;
```
dam-code-guardian, line 288: 🟡 **Warning:** the guarded write can match **zero rows** and this still returns `true`, so the caller `continue`s and skips the backoff — the connection is treated as parked when nothing was marked.

The `where` clause is deliberately conditional (`expiresAt`/`connectedAt` must still equal what the tick read). When a credential fix lands between the read and this write, the update legitimately matches nothing — exactly the case the comment above describes. Drizzle resolves without throwing, so the `catch` never fires and `markRefreshFailure` reports success. The connection then has no marker (so it stays in the due set) *and* no backoff entry (so it is retried every `intervalMs`), which is the tight-loop the backoff map exists to prevent. The docstring — "False when nothing was marked, so the caller falls back to the backoff" — describes the intended behavior; only the no-match branch is missing.

Check the affected row count, as `activity-retention.ts:16` does for the same driver:

```suggestion
  const result = await deps.db
```

…and return `false` when it is `0`, before the `securityLog` call (a no-match is not a `refresh_permanent_failure` worth auditing either).

packages/api-server/src/modules/connections/services/connections-service.ts
```diff
@@ -144,6 +155,135 @@ export function createConnectionsService(deps: {
...
+  // Validated by use: nothing is written until the provider accepts it, and its
+  // own rejection is the most useful thing to show, so it rides the BAD_REQUEST.
+  async function rejectIfInvalid<T>(mint: () => Promise<T>): Promise<T> {
+    try {
+      return await mint();
+    } catch (err) {
+      throw new TRPCError({
+        code: "BAD_REQUEST",
+        message:
+          err instanceof Error ? err.message : "The credential was rejected.",
```
dam-code-guardian: 🟡 **Warning:** `err.message` from a failed mint is forwarded verbatim to the client, and for the OAuth paths that string embeds up to 500 characters of the provider's raw response body.

`postTokenEndpoint` builds its message as `` `OAuth token endpoint ${provider.id}: ${res.status} ${txt.slice(0, 500)}` `` (`infrastructure/oauth-engine.ts:145`). `provider.id` is `connection:<conn-id>:<template-id>`, and `txt` is whatever the token endpoint returned — which for a misconfigured or hostile endpoint can echo request material. This is the same class of leak `reviveFailureReason` (`:677`) was written to avoid on the sibling path, with the explicit comment *"Metadata only: a provider's raw body could echo credential material."*

The intent — surfacing the provider's own rejection reason, which is genuinely the most useful thing to show — is right; the raw body is the wrong carrier. Reuse `reviveFailureReason` here so the user sees `invalid_client` / `token endpoint status 401` rather than the body, and log the full message server-side.

packages/api-server/src/modules/connections/services/oauth-flow.ts
```diff
@@ -107,12 +108,17 @@ export function createOAuthFlowService(deps: {
       }
       await deps.secretStore.putFields(pending.ctx.accessTokenRef, fields);
 
+      const isReauth =
+        conn.auth.kind === "oauth" && conn.auth.connectedAt !== undefined;
```
dam-code-guardian: 🟡 **Warning:** `isReauth` is false for any OAuth connection authorized before `connectedAt` existed, so re-authenticating one of those re-emits `ConnectionCreated` and double-counts it in the activity log — the exact outcome the guard at `:148` was added to prevent.

The codebase already knows these connections exist: `deriveStatus` keeps `expiresAt` in the OR specifically for them ("connections created before `connectedAt` existed have an expiry but no `connectedAt`", `connections-service.ts:700`). A legacy connection re-authenticated from the new UI row therefore looks like a first authorization here.

Applying the same back-compat test used for status derivation closes it:

```suggestion
        conn.auth.kind === "oauth" &&
        (conn.auth.connectedAt !== undefined ||
          conn.auth.expiresAt !== undefined);
```

packages/api-server/src/modules/connections/domain/refresh-failure-marker.ts
```diff
@@ -0,0 +1,44 @@
...
+  oauthError: string | undefined;
+  status: number | undefined;
+  /** Whether the connection stores the client secret / signing key itself. */
+  ownsClientSecret: boolean;
+}
+
+export function isPermanentTokenRejection(r: TokenRejection): boolean {
+  // No code to go on: a token endpoint answering 400/401 rejected the
+  // credential rather than failing to serve the request.
+  if (r.oauthError === undefined) return r.status === 400 || r.status === 401;
```
dam-code-guardian: 🟢 **Suggestion:** the no-code fallback treats any 400/401 as permanent, which parks a connection on a response that isn't a credential rejection.

The reasoning holds for a spec-compliant token endpoint, but `parseOAuthErrorCode` returns `undefined` whenever the body isn't JSON-with-`error` or form-encoded-with-`error` — which includes an HTML error page from a proxy or WAF in front of the token endpoint, a common 401 shape. That parks the connection permanently (out of the due set, `expired` in the UI) on what is really a transient upstream failure, and only a human supplying a new credential clears it. The asymmetry matters because the two error directions cost differently: an over-eager park needs manual recovery, while a missed park just retries under backoff, which the loop already handles well.

Worth narrowing to a signal that the response actually came from the token endpoint — e.g. requiring an OAuth-shaped body (`content-type` JSON or form-encoded) before letting a bare status be conclusive. The GitHub-App path is unaffected: `github-app-engine.ts` sets `oauthError: "invalid_client"` explicitly for its 401/404, so it never relies on this branch.

docs/architecture/connections.md
```diff
@@ -1,6 +1,6 @@
 # Connections, Contributions, and the Runtime Channel
 
-Last verified: 2026-07-27
+Last verified: 2026-07-30
```
dam-code-guardian: 🟡 **Warning:** not bumped to the PR date, and `docs/architecture/security-and-credentials.md:3` has the same value for a different reason — both need `2026-08-03`.

The drift rule requires the header bumped whenever the page is edited. Here `connections.md` *was* bumped, but to `07-30` in `a45c27e`, whose own commit date is `07-31`; and `security-and-credentials.md` was edited in `ff734ca` (`07-31`, the 8-line client-secret paragraph) without line 3 being touched at all, leaving its header older than its own last edit. The PR head commit is `2026-08-03`.

```suggestion
Last verified: 2026-08-03
```

The prose itself checks out against the code — the marker semantics, the operator-supplied carve-out, and the horizon fallback all match what `refresh-failure-marker.ts` and `deriveStatus` actually do.

packages/cli/src/modules/connection/commands/update.ts
```diff
@@ -13,20 +13,32 @@ import { promptSecret } from "../../shared/prompt-secret.js";
 import { resolveConnectionRef } from "../domain/connection-ref.js";
 import type { ConnectionService } from "../services/connection-service.js";
 
+/** What `value` means per auth kind — the server dispatches on the same
+ *  distinction, so the prompt has to name the right secret. */
+const SECRET_LABELS: Record<string, string> = {
```
dam-code-guardian: 🟡 **Warning:** typing this as `Record<string, string>` over a closed union means a missing kind is a silent generic prompt rather than a compile error.

`ConnectionAuthKind` is a fixed five-member enum in the contract package, and `authKind` is already narrowed to it at this call site. With the loose key type, the lookup at `:115` needs its `?? "New credential value"` fallback, and when a sixth auth kind is added to the contract this file keeps compiling while prompting for the wrong secret — a user pasting a credential named only as "value".

The UI's equivalent map gets this right (`packages/ui/src/modules/connections/forms/field-copy.ts:51`) with `Record<Exclude<ConnectionAuthKind, "none">, …>`. Same key type here makes the exhaustiveness a type error and lets the fallback go:

```suggestion
const SECRET_LABELS: Record<Exclude<ConnectionAuthKind, "none">, string> = {
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `7269cee`

## PR #3123: feat(connections): re-authenticate and update credentials in place
**Author:** PetrBulanek | **Branch:** feat/3042-connection-credential-maintenance → main | **Changes:** +1382 −128 (30 files)

### Summary
Adds in-place credential maintenance for connections: OAuth re-authentication (new `dam connection reauth` + UI popup flow), per-kind secret replacement via `connections.update` (header value, client secret, GitHub App PEM, OAuth client secret), and a `refreshFailedAt` marker so a permanently-rejected credential derives `expired` and leaves the refresh loop's due set.

### Findings
- 🟡 **Warning:** `markRefreshFailure` returns `true` without checking whether the conditional update matched any row (`packages/api-server/src/modules/connections/services/oauth-refresh.ts:278`)
- 🟡 **Warning:** `rejectIfInvalid` puts raw provider error text into a client-visible `BAD_REQUEST` message (`packages/api-server/src/modules/connections/services/connections-service.ts:282`)
- 🟡 **Warning:** `isReauth` derives from `connectedAt`, so a pre-`connectedAt` connection re-emits `ConnectionCreated` (`packages/api-server/src/modules/connections/services/oauth-flow.ts:112`)
- 🟢 **Suggestion:** status-only classification of an OAuth error-less 400/401 can park a healthy connection (`packages/api-server/src/modules/connections/domain/refresh-failure-marker.ts:28`)
- ✅ **Looks good:** the marker write is a server-side `jsonb_set` guarded on `expiresAt`/`connectedAt` matching the tick's read — a credential fix landing mid-tick wins instead of being clobbered, and `withoutRefreshFailureMarker` is applied on every one of the four success paths.

**Verdict:** minor drift

❌ **Check 2 — `Last verified:` staleness.** `docs/architecture/connections.md:3` and `docs/architecture/security-and-credentials.md:3` both read `Last verified: 2026-07-30`, but both pages are edited in this PR, whose head commit is dated 2026-08-03. History shows how each drifted: `connections.md` was bumped `07-29 → 07-30` in `a45c27e`, whose own commit date is 07-31; `security-and-credentials.md` was edited in `ff734ca` (07-31, the client-secret paragraph) without touching line 3 at all, so its header is now older than its own last edit. Proposed edit: set both to `2026-08-03`.

Substantive code→docs coverage verified present (in-place rotation per auth mode, validate-by-minting, re-consent with current template scopes, the reject-vs-fail-to-answer marker, the operator-supplied carve-out, the horizon fallback). `cli.md` deliberately omits per-verb surface, so the new `reauth` verb needs no edit. No ADR references, no volatile content, no new subsystem page owed, no cross-reference rot.

1. **Service layer imports a concrete infrastructure class as a runtime value.** `services/oauth-refresh.ts:17` and `services/connections-service.ts:36` import `OAuthTokenEndpointError` as a value (pre-PR both were `import type`-only) and branch on it with `instanceof` (`oauth-refresh.ts:202,270`; `connections-service.ts:678`). Rule: services depend on ports, not implementations. The rest of the design respects this — the domain owns the `TokenRejection` port shape and `isPermanentAuthFailure` maps into it; only the discriminator sits on the wrong side of the boundary. Fix: expose a type-guard on the `OAuthEngine` port so services classify through the injected dependency and keep `import type`.
2. **`Record<string, string>` where the key set is a closed union.** `packages/cli/src/modules/connection/commands/update.ts:18` — this forces the `?? "New credential value"` fallback at `:115`, and adding a sixth auth kind to the contract compiles silently with a generic prompt. The UI does it right at `forms/field-copy.ts:51` (`Record<Exclude<ConnectionAuthKind, "none">, …>`), which makes a new kind a type error.

Verified clean: validation stays in the contract package, `refresh-failure-marker.ts` is pure (zero imports, deterministic), cross-module imports go through `index.ts`, and the `refreshOAuthAccessToken` extraction shares a use case between two application-layer callers with infrastructure injected. Raw `Db` in the refresh loop and `build-connection.ts`'s infra import are pre-existing — out of diff scope.

1. **[CRITICAL] `catalog-connection-row.tsx:71` return block is 107 lines of nested JSX** for one list row (target ≤ ~60 for a feature component); this PR added the inline-fix button (`:90–100`) and two dropdown items (`:136–152`) to an already-dense row. Seam: extract the actions cluster as a `<ConnectionRowActions>` sibling.
2. **[HIGH] The dialog-mount triplet repeats three times** — `connection-catalog-modal.tsx:120`, `views/connections-view.tsx:88`, and `sandboxes/components/granted-connections-panel.tsx:82` each hand-write the same `{maintenance.updating && <ConnectionUpdateCredentialDialog …>}`. The hook already owns the dialog state; let it own the mount too (keeping both mount styles, since the catalog modal deliberately replaces rather than stacks).

Verified clean: state lineage correct (`updating` is UI-local, the list stays in TanStack Query, no shadow copies), `suppressErrorToast` is a real typed meta key so the inline-vs-toast split follows the rules, the single-field dialog is correctly below the RHF threshold, and Rules of Hooks hold at all three call sites.

### Verdict
COMMENT — solid, well-reasoned feature with genuinely careful concurrency handling on the marker write; the findings are hardening and layering issues, none blocking.

---

packages/api-server/src/modules/connections/services/oauth-refresh.ts
```diff
@@ -13,15 +13,22 @@ import {
   type ConnectionAuthConfig,
   type Contribution,
 } from "api-server-api";
-import type {
-  OAuthEngine,
-  OAuthProvider,
+import {
+  OAuthTokenEndpointError,
```
dam-code-guardian: 🟡 **Warning:** importing `OAuthTokenEndpointError` as a runtime **value** couples the application layer to a concrete infrastructure implementation; pre-PR this module imported from `oauth-engine.js` with `import type` only.

The classification design is otherwise careful about the boundary: the domain owns the port shape (`TokenRejection`, `domain/refresh-failure-marker.ts:18`) and `isPermanentAuthFailure` (`:198`) maps the error onto it. Only the discriminator — the `instanceof` at `:202` and `:270`, and the same pattern in `connections-service.ts:678` — reaches across for a class. That makes the services untestable against a stubbed engine that reports rejections without constructing this exact class, and it is the one dependency-direction inversion the change introduces.

Giving the `OAuthEngine` port a type guard (or having it surface rejections as a typed tagged result) lets both services classify through the injected dependency and restores `import type` in both files. Worth noting `infrastructure/github-app-engine.ts:2` importing the class is fine — that is infra-to-infra.

packages/ui/src/modules/connections/components/catalog-connection-row.tsx
```diff
@@ -59,6 +87,17 @@ export function CatalogConnectionRow({
         </div>
         <div className="flex flex-wrap items-center justify-end gap-x-2 gap-y-1.5">
           <GithubAppInstallLink connection={connection} />
+          {inlineFix && (
```
dam-code-guardian: 🟡 **Warning:** this row's return block is now **107 lines of nested JSX** (from `:71`), well past the ~60-line target for a feature component — and a list row is the place where that compounds fastest.

This PR added the inline-fix button here and two `DropdownMenuItem`s at `:136–152`, on top of an existing grant toggle, manage/delete menu, status badge, and install link. The actions cluster is a clean seam: `inlineFix`, the grant control, and the ⋮ menu are one responsibility, and the `reauthLabel`/`updateLabel`/`inlineFix` derivations (`:56–70`) would move with it.

Extracting `<ConnectionRowActions connection={connection} grant={grant} maintenance={maintenance} onManage={…} onDelete={…} />` into a sibling file brings the row back under the bound and gives the maintenance affordances somewhere to grow. The `inlineFix` precedence logic itself is well-judged — re-authentication first because it always works, update only when the connection carries its own secret.

packages/ui/src/modules/sandboxes/components/granted-connections-panel.tsx
```diff
@@ -30,6 +32,8 @@ export function GrantedConnectionsPanel({
   onOpenCatalog,
   inset = true,
 }: Props) {
+  const maintenance = useConnectionMaintenance();
```
dam-code-guardian, line 35: 🟢 **Suggestion:** third copy of the same mount pattern — `useConnectionMaintenance()` plus a hand-written `{maintenance.updating && <ConnectionUpdateCredentialDialog connection={maintenance.updating} onClose={maintenance.closeUpdate} />}` (here at `:82`, also `connections-view.tsx:88` and `connection-catalog-modal.tsx:120`).

The hook's own docstring says it exists so "the surfaces listing connections can't drift on what each auth kind gets" — but the dialog mount, which is the other half of that contract, is still copy-pasted per surface. A fourth surface can wire the actions and forget the dialog, and the failure mode is silent: the menu item does nothing.

Either return the element from the hook, or ship a paired `<ConnectionMaintenanceDialog maintenance={maintenance} />` that owns the `updating &&` guard. It needs to support both mount styles — the catalog modal deliberately *replaces* itself rather than stacking (correctly, per its comment at `:114`), which is a prop rather than an obstacle.

Tomas2D (approved):

Issue #3019, opened 2026-07-28
"In review" badge stays on a published skill after its pull request is closed


PetrBulanek:
## What happens

Publishing a skill from a sandbox puts an **In review** badge on it, linking to the pull request that was opened. That badge never changes. Once a skill has been published it shows "In review" for the rest of the sandbox's life — after the pull request is merged, and after it is closed without merging. The only way to learn the real state is to click through to GitHub and find out the badge was wrong.

## What should happen

The badge should tell the truth about the pull request: still open, merged, or closed.

If we can't know the state, it should claim less rather than claim something false — saying only that the skill was **published** to that source, which is all we actually recorded.

## Steps to reproduce

1. Open the Skills page of a running sandbox that has a skill created in the sandbox and a connected GitHub source.
2. Publish the skill to that source.
3. On GitHub, close or merge the pull request that was opened.
4. Reload the Skills page. The skill still shows "In review".

## Notes

Two directions, in rough order of cost:

- **Say less.** Relabel the badge to "Published · *source*", still linking to the pull request. Nothing to fetch, never wrong, but you lose the at-a-glance signal of whether it's still open.
- **Fetch the state.** Look up the pull request and show "In review" / "Merged" / "Closed". For a public source this is a plain read; for a private one it needs the sandbox's own GitHub access, so the two cases are not the same size of change. The result would need caching — the Skills page polls, and this shouldn't turn into a request per skill per poll.

Worth deciding whether the badge should also distinguish "merged" from "closed without merging" — a merged skill arguably deserves a different visual than an abandoned one.

Found while planning #2825.

Tomas2D:
Resolved by e7ad761c (#3124), which took the "say less" direction from this issue.

The badge no longer claims a review state. `rg -i "in review"` over `packages/ui/src` and `packages/api-server/src` now returns nothing; the pill reads **Published** and reports the publish *event*, still linking to the pull request. `standalone-skills-group.tsx` records the reasoning inline and references this issue:

> The pill reports the publish *event*, never the PR's state: nothing here re-reads GitHub, so a state claim would go stale the moment the PR is merged or closed (#3019).

So the false "In review" can no longer appear, because nothing asserts a PR state that isn't read.

[...]

PR #3124, opened 2026-08-03
fix(ui): publish badge claims only what was recorded

PetrBulanek:
Publishing a skill from a sandbox stamps it with an "In review" badge that is written once and never refreshed, so it keeps claiming the pull request is open long after it has been merged or closed without merging.

This makes the badge stop asserting a state nothing re-reads: it reports that the skill was published to that source, when, and links to the pull request.

That is also the honest floor for the state-aware badge tracked in #3019 — every unresolvable case (deleted pull request, rate-limited read, private source on a hibernated sandbox, GitHub not connected) has to degrade to exactly this. Landing it first keeps the follow-up smaller.

Refs #3019

Verified end to end against the local cluster: published a skill, closed the resulting pull request, and confirmed the badge still reads `Published · {source}` where it previously kept insisting "In review".


packages/ui/src/modules/sandboxes/components/skills/standalone-skills-group.tsx
```diff
@@ -145,10 +151,17 @@ export function StandaloneSkillsGroup({
...
                   rel="noopener noreferrer"
-                  className="inline-flex shrink-0 items-center gap-1.5 rounded-full bg-info-light px-2.5 py-1 text-xs font-medium text-info transition-opacity hover:opacity-80"
-                  title={`Pull request open on ${pub.sourceName}`}
+                  className={cn(
+                    badgeVariants({ variant: "muted" }),
+                    // border-border, because the readOnly card is bg-muted too
+                    // and the pill would otherwise vanish into it.
+                    "shrink-0 gap-1.5 border-border font-medium transition-opacity hover:opacity-80",
+                  )}
+                  title={`Published to ${pub.sourceName} on ${formatDateTime(
```
dam-code-guardian, line 160: 🟢 **Suggestion:** the publish timestamp now lives only in this `title`, which is hover-only — invisible on touch devices and to keyboard users, so the one piece of information the relabel added is unreachable for them. The pill text itself still carries no recency signal. Worth considering a visible relative time (the module already has `timeAgo`) alongside the absolute time on hover.

This is the pre-existing shape of how this pill communicates detail, not a regression introduced here — raising it as a follow-up, not a blocker.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `7ef1570`

## PR #3124: fix(ui): publish badge claims only what was recorded
**Author:** PetrBulanek | **Branch:** fix/3019-published-badge-truth → main | **Changes:** +24 −11 (4 files)

### Summary
Relabels the skill publish pill from `In review · {source}` to `Published · {source}`, so it reports the publish event it actually has a record of instead of asserting a PR state nothing ever re-reads; restyles it onto the shared `badgeVariants` and moves the timestamp into the `title`.

### Findings
- 🟢 **Suggestion:** the publish timestamp is reachable only via hover `title` (`packages/ui/src/modules/sandboxes/components/skills/standalone-skills-group.tsx:160`)
- ✅ **Looks good:** the `border-border` override is ordered after `badgeVariants`, so `tailwind-merge` resolves it over the variant's `border-transparent` — the readOnly-card contrast fix actually takes effect rather than being silently dropped

**Verdict: aligned**

No architecture-doc drift. `docs/architecture/skills.md` already documented the badge as "Published", driven by an `agent_skill_publishes` row (lines 86, 101) — at `origin/main`, not as part of this PR. `"In review"` existed *only* in the four changed UI files, never under `docs/`. This PR brings the code into line with what the page already asserted, so there is no code→docs mismatch. No architecture page was edited, so the `Last verified:` bump, ADR-leak and volatile-content checks don't apply; no page moved or was deleted, no new subsystem, no ADR state change.

✅ No findings.

The single routed `.ts` file (`use-skills-surface.ts`) changes exactly two comment lines — the `SkillsSurface.publishes` doc comment and the optimistic-record comment. No type, signature, import, layer or control-flow change, so no architecture rule attaches. Both comments are accurate against the code they annotate and now match the renamed label, removing a stale-comment risk rather than adding one.

✅ No findings.

- **Styling:** replaces a hand-rolled class string with the shared `cva` variant composed through `cn()` — the prescribed pattern, matching the sibling precedent in `skill-row.tsx:68`. No inline styles, no template-literal classes.
- **Theme tokens:** `bg-muted` / `text-muted-foreground` / `border-border` are all defined in both light and dark (`App.css:145-146,188-189`), so no `dark:` variant is needed. The `border-border` override is deliberate and documented — the `readOnly` card is `bg-muted` too, so a transparent-bordered muted pill would vanish into it.
- **Types:** no `any`/`as`. `formatDateTime` is type-safe against `DateInput` and fails soft to `—`; `publishedAt` is a real ISO string on both the server read path (`agent-skills-repository.ts:127`) and the optimistic client path (`use-skills-surface.ts:430`).
- **State lineage:** untouched — `publishes` stays server-owned, the pill is pure render off `latestPublishByName`.

### Verdict
APPROVE — the badge now claims only what the publish record proves, the restyle uses the shared variant and defined tokens correctly, and the only note is a pre-existing hover-only-detail property this change doesn't regress.

---

PR #3127, opened 2026-08-03
Honest prompt delivery feedback

### LINKED ISSUE #829: False "send failed" shown when returning to a backgrounded tab
## Problem

The UI sometimes shows a send-failure error for a prompt that was delivered fine. Observed when the browser window was opened later, after being in the background: on returning, a "send failed" message appears briefly even though the agent received the prompt and responded normally.

False failures train users to ignore the error, which defeats the point of the delivery feedback introduced in #133.

## Proposed solution

The send-failure indicator appears only when delivery genuinely failed. Backgrounding or reopening the browser while a response is pending must not produce a false alarm.

Image

tomkis:
The UI sometimes shows a false "send failed" error for a prompt that was delivered fine — any prompt sent while the previous turn is still running trips a 60s watchdog that mistakes "no content yet" for "not delivered". This makes prompt delivery feedback server-authoritative: the runtime tells the sender when a prompt was accepted (and whether it was queued) and when it was handed to the agent, and the UI reports failure only when one of those signals genuinely doesn't arrive. Prompts the platform really drops (disconnect while queued) now fail loudly with Retry instead of silently. The feature ships with e2e regression specs covering all four delivery scenarios.

Closes #829

## Sub-issues

- [x] 01 — ADR: server-authoritative prompt delivery
- [x] 02 — Runtime: prompt lifecycle notifications
- [x] 03 — UI: delivery state machine
- [x] 04 — UI: fail queued prompts on disconnect + doc update
- [x] 05 — E2E: delivery feedback specs

https://claude.ai/code/session_01Uu7VHEhfFKierMLvH32UKt


packages/ui/src/modules/acp/session-projection.ts
```diff
@@ -110,6 +114,16 @@ export function applyUpdate(messages: Message[], update: AcpUpdate): Message[] {
     case "platform_turn_ended":
```
dam-code-guardian, line 121: 🟡 **Warning:** a queued prompt's bubble is closed empty by the `turnEnded` that arrives right after its promotion, so the reply opens a *new* bubble and the user's message is left with an empty one.

The runtime's frame order for a promotion is: response for p1 → `promptStarted(p2)` (sent from `advanceQueue`) → `turnEnded(p1)`. `promptStarted` clears p2's `queued` flag, which makes p2's still-empty bubble the **active** one by `findActiveAssistant`'s rule (first `streaming && !queued`) — and `turnEnded` then closes exactly that bubble. p2's first content finds no streaming bubble and `appendToActive` appends a fresh one:

```
aA     assistant  streaming=false  "reply A"
uB     user                        "B"
aB     assistant  streaming=false  ""         <- orphaned empty bubble
<new>  assistant  streaming=true   "reply B"  <- reply detached from uB
```

Reproduced against this branch's projection by replaying those three frames in order. It is order-dependent: this is the path where `sendPrompt`'s belt-and-braces already closed p1's bubble (its `await conn.prompt()` continuation is a microtask, so it runs before the next WS frame). If the belt-and-braces has not run yet, `turnEnded` closes p1's bubble instead and everything routes correctly — which is also why the e2e specs don't catch it: they assert the reply text is visible, never which bubble it landed in.

Suggested fix: have the turn boundary skip a bubble that is an unstarted promotion — the same "has a `promptId` and no content yet" guard `setQueuedByPromptId` already applies.

packages/e2e/playwright/src/lib/delivery.ts
```diff
@@ -0,0 +1,129 @@
...
+import type { ApiClient } from "./api-client.js";
+import { loginViaUi } from "./auth.js";
+import { agentName } from "./fixtures.js";
+
+/** The UI's own delivery deadline (`DELIVERY_TIMEOUT_MS` in
+ *  use-prompt-delivery.ts). Not restated as a bare literal at the call sites:
+ *  every assertion below is positioned relative to this boundary, so if the
+ *  timeout ever moves the assertions have to move with it rather than quietly
+ *  stop testing anything. */
+export const DELIVERY_TIMEOUT_MS = 60_000;
```
dam-code-guardian: 🟢 **Suggestion:** this restates the UI's `DELIVERY_TIMEOUT_MS` as an independent literal, which the doc comment above it explicitly wants to avoid ("if the timeout ever moves the assertions have to move with it rather than quietly stop testing anything"). A copied literal moves only if someone remembers both places; importing the value from `packages/ui/src/modules/sessions/hooks/use-prompt-delivery.ts` makes the coupling real.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `30e4a23`

## PR #3127: Honest prompt delivery feedback
**Author:** tomkis | **Branch:** fix/829-prompt-delivery → main | **Changes:** +2157 −100 (33 files)

### Summary
Makes prompt delivery feedback server-authoritative: the runtime emits `promptAccepted`/`promptStarted` notifications keyed by a sender-minted `promptId`, and the UI's delivery state machine replaces a single send-anchored 60s watchdog with per-prompt deadlines. Genuinely-dropped prompts (disconnect while queued) now fail loudly with Retry.

### Findings
- 🟡 **Warning:** promoted queued bubble is closed empty by the `turnEnded` that follows its promotion; the reply opens a new bubble, orphaning the user's message (`packages/ui/src/modules/acp/session-projection.ts:114`)
- 🟢 **Suggestion:** `DELIVERY_TIMEOUT_MS` is duplicated as a literal in the e2e lib rather than imported (`packages/e2e/playwright/src/lib/delivery.ts:19`)
- ✅ **Looks good:** the `catch` guard that refuses to overwrite an existing bubble error is the right call — `ConnectionClosedError` racing the WS close handler would otherwise replace the specific "dropped from the queue" wording with the generic one.

✅ No findings. `docs/architecture/agent-lifecycle.md` gained a "Prompt delivery" section covering all three fates, the sender-only/ephemeral property, the four failure rules, and the lossy-queue rationale; `Last verified:` bumped to 2026-08-03. No ADR references leaked into the page, the protocol is described semantically with a link-out to the contract package rather than transcribed field shapes, and no ADR status flipped in the diff.

✅ No findings. New code stays within the existing layer boundaries: the notification schemas + builders land in the contract package's flat `modules/acp/types.ts` and are re-exported from its root, the runtime consumes them from `api-server-api` in its `services/` layer, and `extractPromptId` sits beside the existing frame-meta helpers at module scope. No new `services → infrastructure` concrete import, no cross-module import bypassing a public entry point, and no `any`/assertion added at a boundary (`parseExtParams` narrows via Zod `safeParse` and degrades to `null`).

✅ No findings. The only `.tsx` change is three `data-testid` attributes on existing elements — no state, structure, or lineage change. The delivery deadlines went into a dedicated `usePromptDelivery` hook with a stable `useMemo` identity rather than being bolted onto `useAcpPrompt`, which is the split this ruleset asks for.

### Verdict
COMMENT — the server-authoritative design is sound and well-tested; one bubble-routing regression in the promotion path is worth fixing before merge.

---

packages/ui/src/modules/acp/session-projection.ts
tomkis: Fixed in 2029672f. closeActiveAssistant now skips a streaming bubble that has a promptId and no content yet, so the turnEnded that lands right after promotion leaves the promoted placeholder open. The commit adds a unit test replaying the exact frame order from this finding (prompt response for p1, promptStarted(p2), turnEnded) and asserts the reply lands in the promoted bubble, not a fresh one.

packages/e2e/playwright/src/lib/delivery.ts
tomkis: Fixed in 8237a34a. The e2e helper now imports DELIVERY_TIMEOUT_MS from the UI (platform-ui/use-prompt-delivery, exported as TS source like the other workspace api packages) instead of restating 60_000. The spec titles interpolate the imported value, so Playwright loading the full tier verifies the import chain works at runtime.

packages/ui/src/modules/sessions/hooks/use-acp-prompt.ts
```diff
@@ -168,10 +182,31 @@ export function useAcpPrompt(
...
+        } else if (!bubble?.error && !closedMidTurn) {
+          // Unless the bubble already carries a failure. Losing the connection
+          // rejects this call with `ConnectionClosedError` *and* runs the WS
+          // close handler, and the two race in either order: when the close
+          // handler got there first its specific "dropped from the queue"
+          // wording must not be overwritten — and when THIS path wins on a
+          // still-queued bubble, it reports that same wording itself, because
+          // a queued prompt on a dead connection is exactly that loss, not a
+          // generic send error.
+          const errMsg =
```
dam-code-guardian: 🟡 **Warning:** this round's rewrite covers two of the three states a bubble can be in when the connection dies, but not the middle one: a prompt whose turn **started** and has not yet produced its first chunk falls through to `extractErrorMessage(err)` and renders `Send failed: Connection closed while request was in flight` — the raw SDK string, not delivery wording.

The three states, by predicate:

| bubble at drop | `closedMidTurn` | `bubble.queued` | result |
| --- | --- | --- | --- |
| queued, empty | false (no parts) | true | `QUEUED_LOST_MESSAGE` ✅ |
| started, **content streamed** | **true** | false | preserved, no error ✅ |
| started, **no content yet** | false (no parts) | false | `extractErrorMessage(err)` ⚠️ |

Both guards key on `parts.length > 0` / `queued`, and a started-but-silent turn satisfies neither. I transcribed this path and `failQueuedOnDisconnect` into a standalone script and ran all three states, with the first two as controls: the controls produce the intended wording, and the probe produces `"Connection closed while request was in flight"`.

What makes it worth fixing rather than cosmetic is that the message is also **wrong about the outcome**. `detach()` only discards *queued* prompts; an already-forwarded prompt keeps its slot (`if (active.channel === channel) active.channel = null`) and the agent finishes the turn. So the reply does arrive, and the replay after reconnect shows it — but `mergeLocalFailures` carries a `retryWith` error card alongside it, because the card outlives the rebuild by design. The user sees a "Send failed" card with a Retry button next to a perfectly good answer, and taking the Retry re-runs a prompt the agent already ran. That is the #829 lie in the opposite direction, which the comment two lines above explicitly sets out to avoid.

The narrow fix is to treat a `ConnectionClosedError` on a bubble that is *not* queued as a reconnect rather than a send failure, since the turn is already the agent's:

```suggestion
          const errMsg =
            err instanceof ConnectionClosedError
              ? QUEUED_LOST_MESSAGE
              : extractErrorMessage(err);
```

That only reaches a still-`queued` bubble now, because a non-queued `ConnectionClosedError` no longer stamps an error at all — pair it with `closedMidTurn` widening to `err instanceof ConnectionClosedError && !bubble?.queued`, so a started turn finalizes quietly and lets the replay deliver the reply. Worth a spec too: the disconnect spec drops the socket while prompt B is *queued*, so this state is exactly the one the four scenarios don't cover.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `8237a34`

## PR #3127: Honest prompt delivery feedback
**Author:** tomkis | **Branch:** fix/829-prompt-delivery → main | **Changes:** +1854 −101 (32 files)

### Summary
Makes prompt delivery feedback server-authoritative. Round 2 fixes both prior findings: `closeActiveAssistant` now skips an empty promoted placeholder, and the e2e helper imports `DELIVERY_TIMEOUT_MS` from the UI via a new `exports` subpath. Two unflagged changes also landed — the specs became self-contained on their own agent, and the send-failure catch path was rewritten.

### Changes since last review
Previous HEAD: `30e4a23` (2026-08-03T16:16:51Z) — verdict COMMENT

- ✅ **Fixed:** promoted queued bubble no longer closed empty by the following `turnEnded` — guard skips a placeholder with a `promptId` and no parts; unit test replays the exact frame order (`session-projection.ts:447`)
- ✅ **Fixed:** `DELIVERY_TIMEOUT_MS` now imported from `platform-ui/use-prompt-delivery` instead of restated; whole transitive chain resolves through declared deps (`delivery.ts:2`)
- 🆕 **New:** connection dropping after a turn *started* but before its first chunk reports a raw SDK error string instead of the delivery wording (`use-acp-prompt.ts:206`)

### Findings
- 🟡 **Warning:** the rewritten catch path leaves a gap for a started-but-silent turn (`packages/ui/src/modules/sessions/hooks/use-acp-prompt.ts:206`)

**Verdict:** `aligned`

No drift and no possible-drift items. `agent-lifecycle.md` gained a `#### Prompt delivery` section in the same PR; I verified its substantive claims against the code rather than the prose — queue cap (`PROMPT_QUEUE_CAP = 32`, enforced at `acp-runtime.ts:1392`), `promptId` stripping (`stripPlatformMeta`), loss-on-detach (`detach()` filters by channel), loss on recycle (`promptQueueBySession.clear()`), and sender-only delivery (`sendToChannel`, deliberately not `appendAndFanOut`). No ADR leak, no volatile content, no code-level shape; the one link-out to the contract package is the prescribed pattern. Page is 37,728 of 40,000 chars in the gate's own unit (`content.length`). `Last verified: 2026-08-04` is current. No ADR status changed, so the ADR-impact check produces nothing.

✅ No findings. New contract schemas follow the existing `platformTurnEnded` pattern and flow through the package's public API; runtime layer discipline holds (`services/` imports contract + domain + infrastructure port *types* only); no `any`, casts, or non-null assertions in the added lines.

I suspected the new `exports` subpath would break at runtime — `e2e-playwright` does not declare `react`, and `use-prompt-delivery.ts` imports React hooks at module scope, so importing only the constant still executes that import — and an isolated reproduction did fail with `ERR_MODULE_NOT_FOUND: react`. **That reproduction modeled the wrong shape:** `react` is a declared dependency of `platform-ui` itself, so under pnpm it resolves from `packages/ui/node_modules`, not from the consumer. Re-probed with `react` declared by the exporting package (the real layout) and it resolves fine. Noting it because the control run is the only thing that separated the two, and the seam is worth knowing is load-bearing: the subpath's whole transitive chain (`react`, `acp/types.js` → `@agentclientprotocol/sdk`, `api-server-api`) resolves through `platform-ui`'s own declared deps, and `moduleResolution: "bundler"` makes the typecheck exports-aware.

✅ No findings. This PR's only `.tsx` change is three `data-testid` attributes; no state, effects, or props added. `usePromptDelivery` is correctly *not* wired into the view — `ChatView` calls one orchestrator (`useAcpSession`), so the view's surface did not grow with the feature. Delivery state is derived from message fields (`m.streaming && m.queued && m.parts.length === 0`, `m.error.retryWith`), never mirrored into local state.

### Verdict
COMMENT — both prior findings are genuinely fixed and verified against the predicates rather than the summaries; one new gap in the rewritten disconnect path is worth closing before merge.

---

packages/e2e/playwright/src/tests/full/prompt-delivery/delivery.ts
```diff
@@ -0,0 +1,143 @@
+import { expect, type Locator, type Page } from "@playwright/test";
```
tomkis, line 1: We should have tests/full/prompt-delivery 

Where we keep all suites + this delivery lib. 

Playwrith lib is shared, this is specific to delivery, leaky abstraction.

packages/ui/src/modules/sessions/hooks/use-acp-prompt.ts
tomkis: Fixed in 683f446f. closedMidTurn now keys on !queued instead of has-content, so a started-but-silent turn finalizes quietly and the replay delivers the reply. ConnectionClosedError only reaches the error branch on a queued bubble, so it always gets QUEUED_LOST_MESSAGE and never the raw SDK string. Added the missing spec: silent-turn.spec.ts drops the socket between promptStarted and the first content chunk (anchored on the mock's getReceivedPrompts) and asserts the reply arrives with no failure card and no Retry. All five delivery specs pass against the local test cluster.

packages/ui/src/modules/sessions/hooks/use-acp-prompt.ts
```diff
@@ -168,10 +182,37 @@ export function useAcpPrompt(
...
+        // delivery failure: the runtime keeps an already-forwarded prompt's
+        // slot when our channel detaches (`detach()` discards only queued
+        // prompts), so the agent finishes the turn and the replay after
+        // reconnect delivers the reply. That holds whether content has already
+        // streamed or the turn just hasn't produced its first chunk yet.
+        // Stamping "Send failed" here would be the #829 lie in the opposite
+        // direction — with a Retry that re-runs a prompt the agent already
+        // ran. The WS close handler finalizes the bubble either way.
+        const closedMidTurn =
+          err instanceof ConnectionClosedError && !bubble?.queued;
```
dam-code-guardian: 🟡 **Warning:** rekeying `closedMidTurn` on `!bubble?.queued` fixes the started-but-silent turn, but the predicate is also true for the two windows where **no server frame has arrived yet** — so a prompt that never reached the runtime now closes silently with no error card and no Retry.

`queued` is only ever set by a server frame (`setQueuedByPromptId`, `session-projection.ts:163`), and `platform_prompt_accepted` with `queued: false` is an explicit no-op in the projection (`session-projection.ts:120`), so the flag is `undefined` — hence `!queued` true — in all of:

| window | `bubble.queued` | round 2 | round 3 (shipped) |
| --- | --- | --- | --- |
| `sending`, no frame yet | `undefined` | error card ✅ | **silent close** ⚠️ |
| `accepted{queued:false}`, not started | `undefined` | error card ✅ | **silent close** ⚠️ |
| `queued` (parked) | `true` | `QUEUED_LOST_MESSAGE` ✅ | `QUEUED_LOST_MESSAGE` ✅ |
| `started` (silent or streaming) | `false` | raw SDK string ⚠️ / preserved ✅ | preserved ✅ |

The first window is the one that matters: the socket dies while `conn.prompt()` is still in flight and the runtime never got the frame, so nothing will replay a reply. The comment's justification — `detach()` keeps an already-forwarded prompt's slot — is true only once the runtime *has* the prompt, which is exactly what an `accepted` frame proves and what this window lacks. The window is not instantaneous either: `await ensureConnection()` and `buildPromptBlocks()` (which uploads each attachment, `utils.ts:33`) both run inside the `try` before the send, so a prompt with attachments sits here for as long as the upload takes.

Nothing else recovers it. `delivery.endSend(promptId)` in the `finally` block clears the 60s `sending → accepted` deadline, so the `Couldn't deliver` timer cannot fire afterwards; `failQueuedOnDisconnect` skips the bubble too (it requires `m.queued`), and the reconnect replay has no echo to rebuild from because the runtime never received the prompt. I ported this catch path, `failQueuedOnDisconnect` and `mergeLocalFailures` into a standalone script and ran all four windows in both race orders (close-handler-first and catch-first), with the `queued` and `started+content` states as controls: the controls produce their intended wording, and the `sending` probe ends as `error=NONE, streaming=false, parts=0` — after reconnect, zero messages, the user's typed text gone with no Retry to recover it. That is #829's lie in the third direction: not a false failure, but a false success.

Distinguishing "no frame yet" from "started" needs the state the delivery machine already tracks, not the bubble's `queued` flag. The narrow version keys on the flag having been *set*:

```suggestion
        const closedMidTurn =
          err instanceof ConnectionClosedError && bubble?.queued === false;
```

That keeps the round-2 fix (a started turn has `queued === false`, so it still finalizes quietly and the replay delivers the reply) while `undefined` falls back to the error card. It leaves the `accepted{queued:false}`-but-not-yet-`started` window on the pessimistic side — a Retry offered for a prompt the runtime did accept, which risks a double-run — so if that matters, the more faithful fix is to have the projection record the `accepted` frame instead of no-op'ing it, and key on that. Worth a spec either way: `silent-turn.spec.ts` anchors on `getReceivedPrompts` so it only enters after the runtime has the prompt, which is precisely the state this window isn't in.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `683f446f`

## PR #3127: Honest prompt delivery feedback
**Author:** tomkis | **Branch:** fix/829-prompt-delivery → main | **Changes:** +1966 −101 (33 files)

### Summary
Makes prompt delivery feedback server-authoritative. Round 3 fixes the started-but-silent gap by rekeying `closedMidTurn` on `!queued`, and moves the e2e delivery lib into `tests/full/prompt-delivery/` as asked. The rekey is correct for the state it targeted but widens to two states that never got a server frame, where it now swallows a real send failure.

### Changes since last review
Previous HEAD: 8237a34 (2026-08-04T14:04:26Z) — verdict COMMENT

- ✅ **Fixed:** started-but-silent turn no longer reports the raw SDK string — `closedMidTurn` keys on `!queued`, so it finalizes quietly and the replay delivers the reply; `silent-turn.spec.ts` covers it (`use-acp-prompt.ts:196`)
- ✅ **Fixed:** delivery lib moved to `tests/full/prompt-delivery/delivery.ts` with its specs; `testMatch` follows the directory, no stale path refs, and the `github-code-quality` unused-import complaint on `reattach.spec.ts` went with it
- 🆕 **New:** the `!queued` widening silently discards a genuine send failure before any server frame arrives (`use-acp-prompt.ts:196`)

### Findings
- 🟡 **Warning:** `!bubble?.queued` treats "never accepted" as "already the agent's" (`packages/ui/src/modules/sessions/hooks/use-acp-prompt.ts:196`)

**Verdict:** `aligned`

No drift and no possible-drift items. The delta is a test-file move plus a UI catch-path fix; neither changes a documented subsystem claim. `agent-lifecycle.md`'s `#### Prompt delivery` section still matches the code — its four failure rules read as bounded/unbounded waits, and the connection-drop rule says the sender "raises it locally", which the shipped predicate narrows but does not contradict at the page's altitude. `Last verified: 2026-08-04` current. ADR-081 is `accepted`, added in this PR with no status flip (check 7 produces nothing); no ADR leak into the architecture page, protocol described semantically with a link-out to the contract package.

✅ No findings. The delta keeps layer discipline: the two new notification builders sit in the contract package's flat `modules/acp/types.ts` as Zod schemas with inferred types, re-exported through `index.ts`; the runtime consumes them from `services/acp-runtime.ts`, importing infrastructure only as `type` plus the existing `mappers` — no new concrete infrastructure dependency and no cross-module import bypassing `index.ts`. `extractPromptId` narrows through `isNonNullObject` guards rather than assertions. No `any`, cast, or non-null assertion added at a boundary — the `as any` occurrences in the diff are all in test/spec files (`acp-runtime.test.ts` frame inspection, the `__e2eSockets` window handle in `lib/network.ts`). The moved `delivery.ts` resolves: `platform-ui` is a declared dep of `e2e-playwright` and the `./use-prompt-delivery` exports subpath is unchanged by the move.

✅ No findings. The only `.tsx` change remains three `data-testid` attributes — no state, effects, or props added, and unchanged since the prior round. `usePromptDelivery` is a focused 135-line domain hook in `modules/sessions/hooks/`, one responsibility (own the two deadlines), returning a `useMemo`-stable object; it holds its records in a `useRef` map rather than `useState`, which is right for values no render reads. State lineage is not duplicated — delivery state lives on the message in the store and is derived from server frames, never mirrored locally.

### Verdict
COMMENT — the prior finding is genuinely fixed and the refactor is clean; the fix's predicate widened one state too far, silently dropping the failure card for a prompt that never reached the runtime.

---

packages/e2e/playwright/src/tests/full/prompt-delivery/delivery.ts
tomkis: Done in 6b27f236 (pushed). All five delivery specs plus `delivery.ts` now live together under `src/tests/full/prompt-delivery/`, and nothing delivery-specific is left in `src/lib/`.

Two helpers stayed in the shared lib on purpose:

- `setMockLongTurnReply` (`lib/agents.ts`) is a mock-harness scripting knob, same shape as the other `setMock*` helpers next to it. Delivery specs are its only caller today, but the concept is "script a turn that holds for N ms", not anything about delivery.
- `trackWebSockets` / `dropWebSockets` (`lib/network.ts`) are a generic Chromium workaround for `setOffline` not closing live sockets. Any spec that needs a real disconnect wants these.

Happy to move either one if you would rather keep the rule strict at "single caller means it lives with the suite".

jezekra1:
Read the whole thing. The design is right and I would ship it. The diagnosis — that a client-side timer fundamentally cannot tell "parked behind a running turn" from "never arrived" — is correct, and moving the verdict to the runtime is the only honest fix. Keeping the prompt id opt-in so the CLI, channel workers and older clients need no change is the right call, and so is keeping both frames out of the session log: a logged "accepted" would replay as stale to every later loader.

Two things I want to call out as good, because they are the kind of thing that usually gets missed. The carve-out in closeActiveAssistant for an empty sender placeholder is a genuinely subtle frame-ordering bug, and it is caught and explained rather than papered over. And expectMilestoneBeforeFailure racing the two outcomes is the correct way to assert "no failure appears across a 60 second boundary" — not.toBeVisible() and expect.poll both pass instantly and prove nothing, and almost everyone gets that wrong.

Three things worth a look. I do not think any of them has to hold the merge.

**A prompt lost while still in the sending state fails silently.** In use-acp-prompt.ts, closedMidTurn treats every non-queued bubble as "already forwarded, the replay will bring the reply back". But a bubble that never saw promptAccepted may never have reached the runtime at all, and failQueuedOnDisconnect skips it too since it requires queued. The window is wider than the send round trip: buildPromptBlocks uploads attachments between ensureConnection and the prompt call, so a socket dying during a large attachment upload leaves you with a prompt that was definitely never sent, closed as an empty bubble with no error and no Retry. The state machine already knows the answer — the record is still in sending — it just is not exposed to sendPrompt. Either a small accessor, or have cancelAll fail records still in sending. Worth doing because the ADR commits to "nothing fails without a server frame or a closed socket saying so", and this is a case where the closed socket does say so and the client ignores it.

**mise test is red**, but it is a timeout rather than an assertion failure — cancelled at 10m18s against the 10 minute job ceiling, with go and esbuild still compiling. The new tests are all synchronous against a fake agent so it does not look like this PR caused it, but it is a required check.

**usePromptDelivery has no unit test.** Four states, two timers, and it is the entire client half of the contract — covered only by a five minute full-tier project that never runs in the smoke tier. The runtime side got eight focused unit tests, which is the right instinct; the client side got none. The prompt-accepted branch of applyUpdate is untested too, only prompt-started is. send-error.test.ts suggests this is already the house style here.

Small one: Last verified on docs/architecture/agent-lifecycle.md still says 2026-08-04 even though the page gained a whole new section.

## On ACP v2, and a suggestion for Committed-to

Someone asked whether the ACP v2 prompt RFD solves this same problem. It solves the root cause but not the whole thing, and the split is worth writing into the ADR so nobody has to re-derive it later.

v2 changes a prompt to be answered as soon as it is accepted rather than when the turn finishes. That response is promptAccepted — and it is a nicer version of it, because correlation comes free from the request id, so the sender-minted prompt id, the metadata that carries it, the strip-before-forward step and the custom notification all collapse into "the prompt call resolved". The new agent state updates supersede turnEnded, and the new user-message notification with an agent-owned message id supersedes the echo the runtime synthesises today. That last one is pointed, since this PR just changed that echo to skip the sender specifically to dodge the reconciliation problem v2 removes outright.

What v2 does not give you is the queued bit — it deliberately leaves queued messages as future work and only makes them expressible — and it cannot give you the drop-on-disconnect loss at all, because that is not in ACP's model. ACP has two parties; we have three, with the runtime as a proxy holding a per-channel queue that it discards on detach. Adopt v2 naively and the runtime answers the prompt the moment it queues, so "accepted" comes to mean "the proxy has it and may still throw it away" — less honest than what this PR ships, not more.

Also worth recording: the RFD calls its own change breaking and suggests carrying the behaviour as an opt-in flag on v1 in the meantime. The prompt id in platform metadata is exactly that. This PR already picked the migration path the RFD recommends, which is a point in its favour rather than a compromise.

So the suggested Committed-to addition: of the three signals here, v2 absorbs two and leaves one.

- promptAccepted with queued false becomes redundant — the prompt response replaces it
- promptStarted maps onto the protocol's user-message and running-state updates
- promptAccepted with queued true is genuinely ours and survives v2 intact

Same for the disconnect rule: the lossy per-channel queue is our architecture rather than the protocol's, so that rule outlives any version bump. I will file a follow-up issue to track the migration so this does not have to be worked out again from cold.

jezekra1 (approved):

jezekra1 (approved):

PR #3129, opened 2026-08-03
perf(skills): preview reads one pinned file instead of the repo tarball

### LINKED ISSUE #2826: Opening a skill's preview is slow for large source repositories
## Problem

Opening the in-product preview of a skill re-fetches the whole source repository each time, so for large repos there's a noticeable lag every time a user opens the preview — even reopening the same skill.

## Proposed solution

Fetch (and reuse) just the skill's `SKILL.md` rather than the entire repository, so previews open quickly and repeat opens are instant.

### Dependencies

Noted during #944.

PetrBulanek:
Opening a skill's in-product preview downloads and extracts the source repository's entire archive — every time, including reopening the same skill. On a large catalog repo that's a visible lag on an action that should feel instant, and it's avoidable: the scan that populated the list already knew exactly which directory each skill lives in, then discarded it.

This carries that directory on the catalog contract, so a preview fetches exactly one file, pinned at the commit the catalog listed.

It also fixes a correctness bug found along the way. The preview's "View SKILL.md on GitHub" link was a guess assembled from the source path and skill name, and that guess is wrong for any source whose skills live under `.agents/skills/` or `.claude/skills/` rather than a top-level `skills/` — a layout this repository itself uses. The link is now the real path, and correct immediately rather than after the content loads.

Closes #2826

## Sub-issues

- [x] 01 — Carry each scanned skill's repo directory
- [x] 02 — Preview reads one pinned file, not the repo tarball


packages/api-server-api/src/modules/skills/schemas.ts
```diff
@@ -41,6 +41,12 @@ export const skillSchema = z.object({
   description: z.string(),
   version: z.string(),
   contentHash: z.string(),
+  /** Repo-relative directory the skill was found in, whichever Source Root it
+   *  came from — what a raw-file URL needs. Distinct from `skillSourceSchema.path`,
+   *  which is the source's optional subdir; `dir` is this one skill's actual
+   *  location under it. Optional: the agent-runtime clone scan (private /
```
dam-code-guardian, line 47: 🟢 **Suggestion:** the docstring explains the optionality as a property of the producer — "the agent-runtime clone scan (private / non-GitHub sources) doesn't report it" — but that scan already has the value. Both branches in `packages/agent-runtime/src/modules/skills/services/scan.ts` iterate `rel` (the repo-relative skill dir) and build the `ScannedSkill` from it at lines 116-128 and 143-153; `rel` is simply not copied onto the result the way the public scanner now copies it as `dir`.

So `dir?: string` currently encodes "one of two producers hasn't been taught this yet" rather than a fact about the domain — and the service pays for it with a runtime `if (!skill.dir)` branch that returns `NOT_IMPLEMENTED` for a *private* source, i.e. a limitation the type can't express and the reader has to reconstruct from three comments.

Not a defect in this PR — the guard is correct and well-documented, and populating `rel` on the runtime side is genuinely out of scope here (its `version` is per-skill `lastTouchingSha` rather than a repo HEAD, so the raw-URL read would need its own thought). Worth noting as the follow-up that would let `dir` become required and delete the branch: preview would then work for private sources too, which is the deferred feature the guard is standing in for.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `7255de3c`

## PR #3129: perf(skills): preview reads one pinned file instead of the repo tarball
**Author:** PetrBulanek | **Branch:** perf/2826-skill-preview-pinned-read → main | **Changes:** +79 −78 (6 files)

### Summary
Carries each scanned skill's repo-relative directory on the catalog contract so `getSkillContent` resolves `{version, dir}` from the already-cached scan and fetches one pinned `raw.githubusercontent.com` file, replacing a full tarball download and extraction per preview. Both claims verified: `.agents/skills` really is a `SKILL_SOURCE_ROOTS` entry, so the old guessed blob link was genuinely wrong for this repo's own layout.

### Findings
- 🟢 **Suggestion:** `dir` is optional only because the agent-runtime clone scan doesn't populate it, though both its branches already hold `rel` (`packages/agent-runtime/src/modules/skills/services/scan.ts:127`)
- ✅ **Looks good:** the `dir`-less cell is enumerated rather than assumed — for a private github.com source, `list` caches the agent-runtime result under the *same* `(gitUrl, path)` key, and `getSkillContent` reads it back and correctly returns `NOT_IMPLEMENTED` instead of building a URL from `undefined`.

Verdict: `aligned`. The subsystem page is updated in the same PR, `Last verified:` is bumped to the PR date, no ADR references, and the added contract-package link is exactly what the guidelines prescribe for field-level shape ("link out to the contract package as the field-level source of truth") rather than a volatility leak. The page sits at 28,158 of 40,000 chars in the gate's own unit (`content.length`).

One finding (above). Layering is otherwise clean: the port `readPublicSkillFile` stays declared on `SkillsServiceDeps`, its adapter stays in `infrastructure/`, and `compose.ts` remains the only place naming the concrete implementation — the rename propagated through all three with no stale references.

✅ No findings. Server state flows through `useQuery` with TQ's own state read directly, and I simulated the `dir` precedence change across all six reachable states: the only cell that changes is the public/query-pending one the PR targets, which goes from wrong to right. No state regresses.

### Verdict
APPROVE — a real perf win with a genuine correctness fix alongside it, and the tricky private-source cell is handled deliberately.

---

JanPokorny (approved):

PR #3132, opened 2026-08-03
refactor(ui): fix interaction semantics and the last hand-rolled surfaces

### LINKED ISSUE #3054: UI - fix interaction semantics and the last hand-rolled surfaces
## What

The long tail left after the main consolidation — individually small, mostly about how things behave rather than how they look:

- **Tooltips** — the shared tooltip component has one real consumer, while the rest of the app relies on the browser's native `title` attribute. Native tooltips can't be reached by keyboard or touch.
- **Clickable rows** — several rows are clickable `div`s with hand-written keyboard handling; some respond to Enter but not Space. Needs one shared helper.
- **Hover-revealed actions** — the "…" menus that appear on hover are wired up slightly differently per row type.
- **External links** — 4 different combinations of `rel` attributes across 15 links.
- **Card variations** — the connection, channel and catalogue cards repeat the same header layout three times; 7 selectable cards still hand-build the surface the shared one provides. The API key scope rows hand-build the same control-label-description row the shared radio item provides, but with a checkbox — so there is still no shared version for a multi-select list.
- **Dialog close buttons** — partly a UX decision, not code: some dialogs can only be closed from the footer, with no ✕ in the corner. Should they all have one? The shared dialog header also draws its ✕ at 18px, while 16px is the size used almost everywhere else.
- **Tooling** — the component generator config still names the old icon library, so anything scaffolded with it pulls in the wrong icons.

## Why

Most of these are accessibility or consistency gaps that are individually too small to justify their own ticket, but collectively decide whether the UI can be used from a keyboard and whether the next page inherits the right defaults.

## Done when

- Tooltips come from the shared component; no interactive element depends on a native `title` alone.
- Clickable rows share one helper and respond to both Enter and Space.
- One card-header layout; no hand-built selectable card surfaces remain.
- A decision is recorded on the dialog close button — whether every dialog gets one, and at one size.

kapetr:
The long tail of interaction fixes left after the primitive consolidation — individually small, mostly about how things behave rather than how they look.

**Keyboard and touch reach the whole UI now.** Clickable rows and cards answer Space as well as Enter. Hover-revealed row menus stay visible on a phone, where nothing can hover them, and appear when focused rather than leaving you tabbed onto something invisible. Hints that used to be browser tooltips — unreachable without a mouse — come from the shared tooltip, which keyboard and touch can both open.

**Tooltips only where they say something new.** An unfamiliar glyph gets named; a shortcut, a consequence ("writes a wildcard rule") or a reason a button is disabled gets explained. The overflow menu, the close ✕, a disclosure arrow and the brand mark get nothing — the icon is already its own label.

**One version of each surface.** The connection, channel and catalogue cards share a header. Cards you pick from share one selected look, so the sandbox starting-point step no longer shows two of them side by side. A checkbox with a label and description matches the radio button equivalent, which the API key scopes now use. Every link that leaves the app opens the same way.

Also: the component generator pointed at an icon library the project doesn't install, so anything scaffolded from it imported a missing package.

Closes #3054

https://claude.ai/code/session_016b96BopGRYL8bUG2wUTJSX

### DIFF
--- a/packages/ui/src/components/icon-rail.tsx
+++ b/packages/ui/src/components/icon-rail.tsx
@@ -10,6 +10,7 @@ import {
 
 import { BrandLogo } from "@/components/brand-logo";
 import { Badge } from "@/components/ui/badge";
+import { Tooltip } from "@/components/ui/tooltip";
 import { cn } from "@/lib/utils";
 
 import { getBrand } from "../brand.js";
@@ -93,15 +94,16 @@ export function IconRail({
         data-testid="app-sidebar"
       >
         <div className="flex items-center justify-center pt-2">
-          <button
-            type="button"
-            onClick={sandboxes.navigate}
-            title={getBrand().name}
-            aria-label={getBrand().name}
-            className="rounded-lg p-1 text-foreground/80 transition-colors hover:bg-muted hover:text-foreground"
-          >
-            <BrandLogo />
-          </button>
+          <Tooltip content={getBrand().name} side="right">
+            <button
+              type="button"
+              onClick={sandboxes.navigate}
+              aria-label={getBrand().name}
+              className="rounded-lg p-1 text-foreground/80 transition-colors hover:bg-muted hover:text-foreground"
+            >
+              <BrandLogo />
+            </button>
+          </Tooltip>
         </div>
         <div className="flex flex-col items-center gap-1">
           <RailItem {...sandboxes} />
@@ -137,20 +139,21 @@ export function IconRail({
 
 function RailItem({ label, icon: Icon, active, badge, navigate }: Destination) {
   return (
-    <button
-      type="button"
-      onClick={navigate}
-      title={label}
-      aria-label={label}
-      className={cn(
-        "flex h-10 w-10 items-center justify-center rounded-lg transition-colors",
-        active
-          ? "text-primary bg-muted"
-          : "text-foreground/80 hover:text-foreground hover:bg-muted",
-      )}
-    >
-      <IconWithBadge icon={Icon} badge={badge} />
-    </button>
+    <Tooltip content={label} side="right">
+      <button
+        type="button"
+        onClick={navigate}
+        aria-label={label}
+        className={cn(
+          "flex h-10 w-10 items-center justify-center rounded-lg transition-colors",
+          active
+            ? "text-primary bg-muted"
+            : "text-foreground/80 hover:text-foreground hover:bg-muted",
+        )}
+      >
+        <IconWithBadge icon={Icon} badge={badge} />
+      </button>
+    </Tooltip>
   );
 }
 
--- a/packages/ui/src/components/icon-rail.tsx
+++ b/packages/ui/src/components/icon-rail.tsx
@@ -25,7 +25,6 @@ interface Destination {
   active: boolean;
   badge: number;
   navigate: () => void;
-  hoverLabel?: boolean;
 }
 
 export function IconRail({
@@ -55,7 +54,6 @@ export function IconRail({
     active: view === "experiments",
     badge: 0,
     navigate: navigateToExperiments,
-    hoverLabel: true,
   };
   const knowledgeBases: Destination = {
     label: "Knowledge bases",
@@ -66,7 +64,6 @@ export function IconRail({
       view === "knowledge-base-config",
     badge: 0,
     navigate: navigateToKnowledgeBases,
-    hoverLabel: true,
   };
   const artifacts: Destination = {
     label: "Artifacts",
@@ -74,7 +71,6 @@ export function IconRail({
     active: view === "artifacts",
     badge: 0,
     navigate: () => setView("artifacts"),
-    hoverLabel: true,
   };
   const inbox: Destination = {
     label: "Email",
@@ -139,35 +135,23 @@ export function IconRail({
   );
 }
 
-function RailItem({
-  label,
-  icon: Icon,
-  active,
-  badge,
-  navigate,
-  hoverLabel,
-}: Destination) {
-  const button = (
-    <button
-      type="button"
-      onClick={navigate}
-      aria-label={label}
-      className={cn(
-        "flex h-10 w-10 items-center justify-center rounded-lg transition-colors",
-        active
-          ? "text-primary bg-muted"
-          : "text-foreground/80 hover:text-foreground hover:bg-muted",
-      )}
-    >
-      <IconWithBadge icon={Icon} badge={badge} />
-    </button>
-  );
-  return hoverLabel ? (
+function RailItem({ label, icon: Icon, active, badge, navigate }: Destination) {
+  return (
     <Tooltip content={label} side="right">
-      {button}
+      <button
+        type="button"
+        onClick={navigate}
+        aria-label={label}
+        className={cn(
+          "flex h-10 w-10 items-center justify-center rounded-lg transition-colors",
+          active
+            ? "text-primary bg-muted"
+            : "text-foreground/80 hover:text-foreground hover:bg-muted",
+        )}
+      >
+        <IconWithBadge icon={Icon} badge={badge} />
+      </button>
     </Tooltip>
-  ) : (
-    button
   );
 }
 
--- a/packages/ui/src/components/icon-rail.tsx
+++ b/packages/ui/src/components/icon-rail.tsx
@@ -25,6 +25,7 @@ interface Destination {
   active: boolean;
   badge: number;
   navigate: () => void;
+  hoverLabel?: boolean;
 }
 
 export function IconRail({
@@ -54,6 +55,7 @@ export function IconRail({
     active: view === "experiments",
     badge: 0,
     navigate: navigateToExperiments,
+    hoverLabel: true,
   };
   const knowledgeBases: Destination = {
     label: "Knowledge bases",
@@ -64,6 +66,7 @@ export function IconRail({
       view === "knowledge-base-config",
     badge: 0,
     navigate: navigateToKnowledgeBases,
+    hoverLabel: true,
   };
   const artifacts: Destination = {
     label: "Artifacts",
@@ -71,6 +74,7 @@ export function IconRail({
     active: view === "artifacts",
     badge: 0,
     navigate: () => setView("artifacts"),
+    hoverLabel: true,
   };
   const inbox: Destination = {
     label: "Email",
@@ -94,16 +98,14 @@ export function IconRail({
         data-testid="app-sidebar"
       >
         <div className="flex items-center justify-center pt-2">
-          <Tooltip content={getBrand().name} side="right">
-            <button
-              type="button"
-              onClick={sandboxes.navigate}
-              aria-label={getBrand().name}
-              className="rounded-lg p-1 text-foreground/80 transition-colors hover:bg-muted hover:text-foreground"
-            >
-              <BrandLogo />
-            </button>
-          </Tooltip>
+          <button
+            type="button"
+            onClick={sandboxes.navigate}
+            aria-label={getBrand().name}
+            className="rounded-lg p-1 text-foreground/80 transition-colors hover:bg-muted hover:text-foreground"
+          >
+            <BrandLogo />
+          </button>
         </div>
         <div className="flex flex-col items-center gap-1">
           <RailItem {...sandboxes} />
@@ -137,23 +139,35 @@ export function IconRail({
   );
 }
 
-function RailItem({ label, icon: Icon, active, badge, navigate }: Destination) {
-  return (
+function RailItem({
+  label,
+  icon: Icon,
+  active,
+  badge,
+  navigate,
+  hoverLabel,
+}: Destination) {
+  const button = (
+    <button
+      type="button"
+      onClick={navigate}
+      aria-label={label}
+      className={cn(
+        "flex h-10 w-10 items-center justify-center rounded-lg transition-colors",
+        active
+          ? "text-primary bg-muted"
+          : "text-foreground/80 hover:text-foreground hover:bg-muted",
+      )}
+    >
+      <IconWithBadge icon={Icon} badge={badge} />
+    </button>
+  );
+  return hoverLabel ? (
     <Tooltip content={label} side="right">
-      <button
-        type="button"
-        onClick={navigate}
-        aria-label={label}
-        className={cn(
-          "flex h-10 w-10 items-center justify-center rounded-lg transition-colors",
-          active
-            ? "text-primary bg-muted"
-            : "text-foreground/80 hover:text-foreground hover:bg-muted",
-        )}
-      >
-        <IconWithBadge icon={Icon} badge={badge} />
-      </button>
+      {button}
     </Tooltip>
+  ) : (
+    button
   );
 }
 
--- a/packages/ui/src/components/ui/button.tsx
+++ b/packages/ui/src/components/ui/button.tsx
@@ -1,7 +1,9 @@
 import { Slot } from "@radix-ui/react-slot";
 import { cva, type VariantProps } from "class-variance-authority";
+import type { ReactNode } from "react";
 import * as React from "react";
 
+import { Tooltip } from "@/components/ui/tooltip";
 import { cn } from "@/lib/utils";
 
 const buttonVariants = cva(
@@ -66,6 +68,9 @@ const buttonVariants = cva(
 export interface ButtonProps
   extends React.ComponentProps<"button">, VariantProps<typeof buttonVariants> {
   asChild?: boolean;
+  /** Hover/focus hint, in place of a native `title`. It reads as a description,
+   *  so an icon-only button still needs its own `aria-label`. */
+  tooltip?: ReactNode;
 }
 
 function Button({
@@ -74,15 +79,17 @@ function Button({
   size,
   tone,
   asChild = false,
+  tooltip,
   ...props
 }: ButtonProps) {
   const Comp = asChild ? Slot : "button";
-  return (
+  const button = (
     <Comp
       className={cn(buttonVariants({ variant, size, tone, className }))}
       {...props}
     />
   );
+  return tooltip ? <Tooltip content={tooltip}>{button}</Tooltip> : button;
 }
 
 export { Button, buttonVariants };
--- a/packages/ui/src/components/ui/button.tsx
+++ b/packages/ui/src/components/ui/button.tsx
@@ -86,6 +86,12 @@ function Button({
   const button = (
     <Comp
       className={cn(buttonVariants({ variant, size, tone, className }))}
+      /* A disabled button fires neither pointer nor focus events, so the
+         tooltip can never open — `title` still reaches assistive tech, which
+         matters most for a hint that explains why the button is disabled. */
+      title={
+        props.disabled && typeof tooltip === "string" ? tooltip : undefined
+      }
       {...props}
     />
   );
--- a/packages/ui/src/components/ui/button.tsx
+++ b/packages/ui/src/components/ui/button.tsx
@@ -68,8 +68,8 @@ const buttonVariants = cva(
 export interface ButtonProps
   extends React.ComponentProps<"button">, VariantProps<typeof buttonVariants> {
   asChild?: boolean;
-  /** Hover/focus hint, in place of a native `title`. It reads as a description,
-   *  so an icon-only button still needs its own `aria-label`. */
+  /** Reads as a description, so an icon-only button still needs an
+   *  `aria-label`. */
   tooltip?: ReactNode;
 }
 
--- a/packages/ui/src/components/ui/checkbox.tsx
+++ b/packages/ui/src/components/ui/checkbox.tsx
@@ -37,7 +37,8 @@ interface CheckboxItemProps extends Omit<
 }
 
 /** The multi-select counterpart of `RadioGroupItem`. `className` styles the
- *  row, not the box. */
+ *  row, not the box. `aria-label` keeps the description out of the accessible
+ *  name, which the wrapping `<label>` would otherwise run together with it. */
 function CheckboxItem({
   label,
   labelClassName,
@@ -45,6 +46,7 @@ function CheckboxItem({
   testId,
   className,
   id,
+  "aria-describedby": describedBy,
   ...props
 }: CheckboxItemProps) {
   const generatedId = React.useId();
@@ -61,7 +63,12 @@ function CheckboxItem({
       <Checkbox
         id={controlId}
         className="mt-0.5"
-        aria-describedby={description ? descriptionId : undefined}
+        aria-label={label}
+        aria-describedby={
+          [description ? descriptionId : null, describedBy]
+            .filter(Boolean)
+            .join(" ") || undefined
+        }
         data-testid={testId}
         {...props}
       />
--- a/packages/ui/src/components/ui/checkbox.tsx
+++ b/packages/ui/src/components/ui/checkbox.tsx
@@ -25,4 +25,62 @@ function Checkbox({
   );
 }
 
-export { Checkbox };
+interface CheckboxItemProps extends Omit<
+  React.ComponentProps<typeof CheckboxPrimitive.Root>,
+  "children"
+> {
+  label: string;
+  /** Extra classes for the label text, e.g. `font-mono` for an identifier. */
+  labelClassName?: string;
+  description?: string;
+  testId?: string;
+}
+
+/** A checkbox with a label and an optional description, laid out like
+ *  `RadioGroupItem` so a multi-select list and a single-select list read the
+ *  same. The label wraps the control, so a click anywhere in the row toggles
+ *  it, and `className` styles that row rather than the box. */
+function CheckboxItem({
+  label,
+  labelClassName,
+  description,
+  testId,
+  className,
+  id,
+  ...props
+}: CheckboxItemProps) {
+  const generatedId = React.useId();
+  const controlId = id ?? generatedId;
+  const descriptionId = `${controlId}-description`;
+  return (
+    <label
+      htmlFor={controlId}
+      className={cn(
+        "flex w-full cursor-pointer items-start gap-2.5 text-left transition-colors has-[:disabled]:cursor-not-allowed has-[:disabled]:opacity-50",
+        className,
+      )}
+    >
+      <Checkbox
+        id={controlId}
+        className="mt-0.5"
+        aria-describedby={description ? descriptionId : undefined}
+        data-testid={testId}
+        {...props}
+      />
+      <span className="flex min-w-0 flex-1 flex-col gap-0.5">
+        <span
+          className={cn("text-sm font-medium text-foreground", labelClassName)}
+        >
+          {label}
+        </span>
+        {description && (
+          <span id={descriptionId} className="text-sm text-muted-foreground">
+            {description}
+          </span>
+        )}
+      </span>
+    </label>
+  );
+}
+
+export { Checkbox, CheckboxItem };
--- a/packages/ui/src/components/ui/checkbox.tsx
+++ b/packages/ui/src/components/ui/checkbox.tsx
@@ -36,10 +36,8 @@ interface CheckboxItemProps extends Omit<
   testId?: string;
 }
 
-/** A checkbox with a label and an optional description, laid out like
- *  `RadioGroupItem` so a multi-select list and a single-select list read the
- *  same. The label wraps the control, so a click anywhere in the row toggles
- *  it, and `className` styles that row rather than the box. */
+/** The multi-select counterpart of `RadioGroupItem`. `className` styles the
+ *  row, not the box. */
 function CheckboxItem({
   label,
   labelClassName,
--- a/packages/ui/src/components/ui/disclosure.tsx
+++ b/packages/ui/src/components/ui/disclosure.tsx
@@ -32,9 +32,6 @@ export function DisclosureChevron({
 export function DisclosureToggle({
   open,
   onToggle,
-  /** For a section held open by something else, so `aria-expanded` doesn't
-   *  advertise a state this button can't change. */
-  disabled,
   chevronSize,
   chevronClassName,
   testId,
@@ -43,7 +40,6 @@ export function DisclosureToggle({
 }: {
   open: boolean;
   onToggle: () => void;
-  disabled?: boolean;
   chevronSize?: number;
   chevronClassName?: string;
   testId?: string;
@@ -54,7 +50,6 @@ export function DisclosureToggle({
     <button
       type="button"
       onClick={onToggle}
-      disabled={disabled}
       aria-expanded={open}
       data-testid={testId}
       className={cn("flex items-center gap-2 text-left", className)}
--- a/packages/ui/src/components/ui/disclosure.tsx
+++ b/packages/ui/src/components/ui/disclosure.tsx
@@ -32,6 +32,9 @@ export function DisclosureChevron({
 export function DisclosureToggle({
   open,
   onToggle,
+  /** For a section held open by something else, so `aria-expanded` doesn't
+   *  advertise a state this button can't change. */
+  disabled,
   chevronSize,
   chevronClassName,
   testId,
@@ -40,6 +43,7 @@ export function DisclosureToggle({
 }: {
   open: boolean;
   onToggle: () => void;
+  disabled?: boolean;
   chevronSize?: number;
   chevronClassName?: string;
   testId?: string;
@@ -50,6 +54,7 @@ export function DisclosureToggle({
     <button
       type="button"
       onClick={onToggle}
+      disabled={disabled}
       aria-expanded={open}
       data-testid={testId}
       className={cn("flex items-center gap-2 text-left", className)}
--- a/packages/ui/src/lib/clickable.ts
+++ b/packages/ui/src/lib/clickable.ts
@@ -1,0 +1,26 @@
+import type { KeyboardEvent } from "react";
+
+/** Button semantics for a row or card that can't be a real `<button>` — one
+ *  that wraps its own action menu or link, which a button may not contain.
+ *
+ *  Activates on click, Enter and Space. Space is the half hand-rolled rows keep
+ *  missing, and it needs its default suppressed or the page scrolls instead.
+ *  Keys are ignored unless they land on the row itself, so a nested control
+ *  keeps its own Enter and Space.
+ *
+ *  Pass `undefined` to leave the element inert — a row that is only sometimes
+ *  clickable then has no stale `role` or tab stop. */
+export function clickableProps(onActivate: (() => void) | undefined) {
+  if (!onActivate) return {};
+  return {
+    role: "button",
+    tabIndex: 0,
+    onClick: onActivate,
+    onKeyDown: (event: KeyboardEvent<HTMLElement>) => {
+      if (event.target !== event.currentTarget) return;
+      if (event.key !== "Enter" && event.key !== " ") return;
+      event.preventDefault();
+      onActivate();
+    },
+  } as const;
+}
--- a/packages/ui/src/lib/clickable.ts
+++ b/packages/ui/src/lib/clickable.ts
@@ -1,15 +1,9 @@
 import type { KeyboardEvent } from "react";
 
-/** Button semantics for a row or card that can't be a real `<button>` — one
- *  that wraps its own action menu or link, which a button may not contain.
- *
- *  Activates on click, Enter and Space. Space is the half hand-rolled rows keep
- *  missing, and it needs its default suppressed or the page scrolls instead.
- *  Keys are ignored unless they land on the row itself, so a nested control
- *  keeps its own Enter and Space.
- *
- *  Pass `undefined` to leave the element inert — a row that is only sometimes
- *  clickable then has no stale `role` or tab stop. */
+/** Button semantics for a row that can't be a real `<button>` because it wraps
+ *  its own menu or link. Keys are ignored unless they land on the row itself,
+ *  so a nested control keeps its own Enter and Space. `undefined` leaves the
+ *  element inert, with no stale `role` or tab stop. */
 export function clickableProps(onActivate: (() => void) | undefined) {
   if (!onActivate) return {};
   return {
--- a/packages/ui/src/modules/artifacts/components/experiments-section.tsx
+++ b/packages/ui/src/modules/artifacts/components/experiments-section.tsx
@@ -8,7 +8,10 @@ import { EXPERIMENT_FOLDER_PREFIX } from "api-server-api";
 import { useMemo, useState } from "react";
 
 import { Card } from "@/components/ui/card";
-import { DisclosureToggle } from "@/components/ui/disclosure";
+import {
+  DisclosureChevron,
+  DisclosureToggle,
+} from "@/components/ui/disclosure";
 
 import { useExperimentsAmbient } from "../../experiments/api/queries.js";
 import type { ArtifactRowActions } from "./artifact-row.js";
@@ -119,27 +122,42 @@ export function ExperimentsSection({
   );
   if (searching && visible.length === 0) return null;
 
+  const header = (
+    <>
+      <Chemistry size={16} className="shrink-0 text-muted-foreground" />
+      <span className="text-sm font-semibold text-muted-foreground">
+        Experiments
+      </span>
+      <span className="text-xs text-muted-foreground">
+        {folders.length} experiment{folders.length === 1 ? "" : "s"} · {total}{" "}
+        artifact{total === 1 ? "" : "s"}
+      </span>
+      <span className="ml-auto hidden text-xs text-muted-foreground/70 sm:block">
+        scripts, dashboards and run results published by your agents
+      </span>
+    </>
+  );
+
   return (
     <Card className="mt-2 overflow-hidden border-dashed bg-muted/30 anim-in">
-      <DisclosureToggle
-        open={expanded}
-        onToggle={() => setOpen((o) => !o)}
-        disabled={searching}
-        chevronClassName="text-muted-foreground"
-        className="w-full cursor-pointer select-none gap-2.5 px-3.5 py-2.5 transition-colors hover:bg-muted/60"
-      >
-        <Chemistry size={16} className="shrink-0 text-muted-foreground" />
-        <span className="text-sm font-semibold text-muted-foreground">
-          Experiments
-        </span>
-        <span className="text-xs text-muted-foreground">
-          {folders.length} experiment{folders.length === 1 ? "" : "s"} · {total}{" "}
-          artifact{total === 1 ? "" : "s"}
-        </span>
-        <span className="ml-auto hidden text-xs text-muted-foreground/70 sm:block">
-          scripts, dashboards and run results published by your agents
-        </span>
-      </DisclosureToggle>
+      {searching ? (
+        // A search pins the section open, so render a static header rather than
+        // a toggle that keeps a pointer cursor and hover but does nothing — and
+        // would drop keyboard focus the moment the filter turns it inert.
+        <div className="flex items-center gap-2.5 px-3.5 py-2.5">
+          <DisclosureChevron open className="text-muted-foreground" />
+          {header}
+        </div>
+      ) : (
+        <DisclosureToggle
+          open={open}
+          onToggle={() => setOpen((o) => !o)}
+          chevronClassName="text-muted-foreground"
+          className="w-full cursor-pointer select-none gap-2.5 px-3.5 py-2.5 transition-colors hover:bg-muted/60"
+        >
+          {header}
+        </DisclosureToggle>
+      )}
       {expanded &&
         visible.map((folder) => {
           const artifacts = byFolder.get(folder.id) ?? [];
--- a/packages/ui/src/modules/artifacts/components/experiments-section.tsx
+++ b/packages/ui/src/modules/artifacts/components/experiments-section.tsx
@@ -124,6 +124,7 @@ export function ExperimentsSection({
       <DisclosureToggle
         open={expanded}
         onToggle={() => setOpen((o) => !o)}
+        disabled={searching}
         chevronClassName="text-muted-foreground"
         className="w-full cursor-pointer select-none gap-2.5 px-3.5 py-2.5 transition-colors hover:bg-muted/60"
       >
--- a/packages/ui/src/modules/artifacts/components/experiments-section.tsx
+++ b/packages/ui/src/modules/artifacts/components/experiments-section.tsx
@@ -8,7 +8,7 @@ import { EXPERIMENT_FOLDER_PREFIX } from "api-server-api";
 import { useMemo, useState } from "react";
 
 import { Card } from "@/components/ui/card";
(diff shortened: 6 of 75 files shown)


dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `9e88940`

## PR #3132: refactor(ui): fix interaction semantics and the last hand-rolled surfaces
**Author:** kapetr | **Branch:** refactor/3054-ui-interaction-semantics → main | **Changes:** +687 −481 (73 files)

### Summary
Consolidates the interaction long-tail after the UI primitive work: `clickableProps` gives Space+Enter to rows that can't be real buttons, `HOVER_ACTION` makes hover-revealed actions reachable on touch and keyboard, native `title` hints move to the shared Radix `Tooltip`, and new `PanelCard` / `CheckboxItem` / `cardSelectionVariants` retire the last hand-rolled surfaces. The shared contracts are well-documented and the mechanics I verified hold up — `HOVER_ACTION` has a `group` ancestor at all 7 call sites and covers both the direct-trigger (`md:data-[state=open]`) and wrapper (`md:has-[...]`) shapes; the `components.json` flip to `@carbon/icons-react` matches the only installed icon dep (88 imports, lucide 0).

### Findings
- 🟡 **Warning:** `Button`'s new `tooltip` prop makes `Tooltip` the root element, and `Tooltip` neither spreads rest props nor forwards a ref — so any `asChild` parent cloning it loses everything (`packages/ui/src/components/ui/button.tsx:98`)
- 🟡 **Warning:** `CheckboxItem` omits the `aria-label` that `RadioGroupItem` documents by name, so label and description concatenate into the accessible name and the description is announced twice (`packages/ui/src/components/ui/checkbox.tsx:54`)
- 🟡 **Warning:** `RailItem` lost `title` for all six destinations but only three got `hoverLabel: true` — Home, Email and Settings now have no hover label at all (`packages/ui/src/components/icon-rail.tsx:45`)
- 🟡 **Warning:** The skill drift hint moved to a `Tooltip` around a raw `disabled` button, which fires no pointer or focus events — losing the hint in exactly the state that prompts the question (`packages/ui/src/modules/sandboxes/components/skills/skill-row.tsx:64`)
- 🟡 **Warning:** `SelectableCard` lost `bg-muted/60` on the selected state; `cardSelectionVariants` bakes in `bg-card` for both branches, so selection is now border-only (`packages/ui/src/modules/sandboxes/components/steps/selectable-card.tsx:20`)
- 🟡 **Warning:** Migrating to `DisclosureToggle` newly exposes `aria-expanded={expanded}` where `expanded = open || searching` — while searching, the control announces as expanded and verifiably does nothing (`packages/ui/src/modules/artifacts/components/experiments-section.tsx:125`)
- 🟢 **Suggestion:** `clickableProps` guards `onKeyDown` on `event.target !== event.currentTarget` but leaves `onClick` unguarded; this PR removed the paired `onKeyDown` stopPropagation from the wrappers on the strength of that guard (`packages/ui/src/lib/clickable.ts:12`)
- 🟢 **Suggestion:** `AttachmentChip`'s remove button has neither `aria-label` nor `tooltip`, unlike every other icon-only button in the file (`packages/ui/src/modules/sessions/components/chat-input.tsx:261`)
- 🟢 **Suggestion:** `HintTooltip` in `SourceTag` mints a keyboard tab stop whose `aria-label` merely repeats the badge's visible text (`packages/ui/src/modules/egress-rules/components/agent-egress-editor.tsx:635`)
- ✅ **Looks good:** The `Tooltip` `onFocus` guard and `Button`'s disabled-`title` fallback are correctly reasoned; the nesting order at `experiment-dock-panel.tsx:130` (`Tooltip` outside `DropdownMenuTrigger asChild`) is the safe one and holds.

**Verdict:** aligned

No architecture-doc drift. The diff is presentational and interaction-semantics refactoring confined to the `ui` component; no subsystem behavior or responsibility described under `docs/architecture/` changed, and no new long-lived component was introduced.

**Mode: review** — no architecture findings. The three new `.ts` helpers align with the TSEng docs.

- **Layering / module boundaries** — all three are leaves (`external-link.ts` and `hover-action.ts` import nothing; `clickable.ts` imports only `import type { KeyboardEvent }`). The three-layer rule is scoped to server packages; `packages/ui` is a client package, so no layer assignment is owed.
- **Placement** follows existing precedent in the package: Tailwind class strings for primitives in `components/ui/` (cf. `menu-styles.ts`), framework-agnostic prop helpers in `lib/` (cf. `breakpoints.ts`, `compact.ts`).
- **`infrastructure.md` "Required Dependencies — Never Optional"** was the nearest miss, given `clickableProps(onActivate: (() => void) | undefined)` — but that rule covers injected *service* dependencies, not UI event callbacks. Here absence is the helper's documented, load-bearing behavior, used deliberately at `lineage-card.tsx:36` and `sandbox-group-card.tsx:29`.
- **Strict TS** — clean under `strict`/`isolatedModules`; the `as const` returns narrow correctly.

Consistency note in the change's favor: the diff leaves zero hand-rolled `target="_blank"` and zero hand-rolled `group-hover:opacity-100` reveals anywhere in `packages/ui/src`, so these helpers are single sources of truth rather than a fourth parallel spelling.

Reviewed the 6 changed primitives and all 69 routed call sites, focused on `asChild` composition, event propagation, and `title` → `aria-label`/tooltip swaps.

**1. `Button`'s `tooltip` prop breaks any `asChild` parent that wraps it** — HIGH · `button.tsx:98`, first bitten at `files-panel.tsx:43`
With `tooltip` set the root element is `Tooltip`, which destructures exactly `{children, content, side, className}` — no rest spread, no ref forwarding. In `files-panel.tsx` the parent is `DropdownMenuTrigger asChild`, so while `controller.isUploading` the trigger clones `Tooltip` and discards `ref`, `onPointerDown`, `onKeyDown`, `id`, `aria-haspopup="menu"`, `aria-expanded`, `data-state`: the menu loses its Popper anchor and the button stops being announced as a menu button. Today the damage is masked because the same condition also sets `disabled`, so the menu couldn't open anyway — but the coupling is incidental and free to drift. The hazard generalizes: `tooltip` is on the shared `Button`, and every `*Trigger asChild` in the codebase takes a `Button` child. Either forward rest props and a ref through `Tooltip` (restoring what the removed `<span className="inline-flex">` provided structurally), or have `Button` refuse to wrap when itself slotted. Counter-example worth standardizing on: `experiment-dock-panel.tsx:130` puts `Tooltip` *outside* `DropdownMenuTrigger asChild`, and because the Radix primitive merges and re-forwards unknown props, that chain holds.

**2. `CheckboxItem`'s accessible name swallows the description** — HIGH · `checkbox.tsx:54`
The docstring claims it is "the multi-select counterpart of `RadioGroupItem`", but it diverges on the one contract that component's comment calls out by name: *"`aria-label` keeps the description out of the accessible name, where name-from-content would otherwise run label and description together."* `CheckboxItem` sets no `aria-label`; a `<label htmlFor>` wraps both spans, and since Radix renders `<button role="checkbox">` (a labelable element), the accessible name becomes label **and** description concatenated — then `aria-describedby` announces the description a second time. Worst at `create-form.tsx:164`, where `scopeDescription()` returns full sentences. Fix by mirroring `RadioGroupItem`: `aria-label={label}` plus the same merge of caller-supplied `aria-describedby`.

**3. `icon-rail`: Home, Email and Settings lost their hover label outright** — HIGH · `icon-rail.tsx:45`, `:165`
`RailItem` previously carried `title={label}` unconditionally. The `title` was removed for every item, but `hoverLabel: true` was added to only `experiments` (:58), `knowledgeBases` (:69), `artifacts` (:77). `sandboxes` ("Home"), `inbox` ("Email") and `settings` ("Settings") now render no tooltip, so in the collapsed 56px rail a sighted pointer user has no way left to learn what those icons mean. `aria-label` keeps screen readers correct, so this is purely visual and reads as an oversight. Either flag all six or drop the flag and always render the `Tooltip`.

**4. `skill-row`: the Update hint moved to a tooltip that cannot open** — MODERATE-HIGH · `skill-row.tsx:64`
The drift hint became a `Tooltip` around a raw `<button disabled={disabled}>`. This PR twice documents why that fails (`button.tsx:88`, `experiment-dock-panel.tsx:325`): a disabled button fires neither pointer nor focus events. The old native `title` *did* render on a disabled button, so the hint is lost precisely where "why can't I click this?" is asked. `Button`'s mitigation is unavailable because this is a hand-rolled `<button>` styled with `badgeVariants`; routing it through `Button` would inherit it.

**5. `HintTooltip` mints non-actionable tab stops whose name repeats the visible text** — MODERATE · `tooltip.tsx:85`, worst at `agent-egress-editor.tsx:635`
`SourceTag` passes `label={label}` while the wrapped `Badge` renders `{label}`, so `aria-label` duplicates visible text and every egress rule row gains a tab stop for a decorative badge. A screen-reader user hears *"manual, button"* with no hint that it does nothing and no access to the tooltip text (Radix only wires `aria-describedby` while open). This also cuts against the PR's own goal of removing tooltips that restate a control's label. Consider `tabIndex={0}` on a non-button element with the hint as the accessible *description*.

**6. `SelectableCard` lost the fill that marked the selected state** — MODERATE · `selectable-card.tsx:20`
Before: selected → `border-foreground bg-muted/60`; unselected → `border-border bg-card hover:bg-muted/40`. `cardSelectionVariants` bakes in `CARD_SURFACE` (carrying `bg-card`) and its `selected: true` branch adds only `border-foreground`, so both states paint `bg-card` and selection is border-color alone. The harness and workload pickers lose their selected tint — a case where the new shared contract can't express the call site's need.

**7. `clickableProps` guards the keyboard path but not the click path** — MODERATE · `clickable.ts:12`
`onKeyDown` correctly bails on `event.target !== event.currentTarget`; `onClick: onActivate` has no equivalent guard, so it still fires on clicks bubbling from any descendant. The asymmetry is load-bearing: all four call sites must remember a manual `onClick={(e) => e.stopPropagation()}` wrapper (`artifact-row.tsx:99`, `chat-artifacts-panel.tsx:118`, `lineage-card.tsx:77`, plus `:47` and `artifact-row.tsx:133`), and this PR *removed* the paired `onKeyDown` stopPropagation from those wrappers on the strength of the new keyboard guard. All current sites are covered, so no live bug — but the doc comment reads as though propagation is handled centrally when only half of it is.

**8. `experiments-section`: `aria-expanded` advertises a state the toggle can't change** — MODERATE · `experiments-section.tsx:109`, `:125`
`expanded = open || searching`, but `DisclosureToggle` gets `open={expanded}` while `onToggle` flips `open`. While a search filter is active, `expanded` is pinned true: clicking the header mutates `open` invisibly and the section never collapses. The stale-state shape predates this PR, but `DisclosureToggle` newly exposes `aria-expanded={expanded}`, promoting a cosmetic quirk into an assistive-tech lie. Either disable the toggle while `searching` or derive one source of truth.

**9. `ShareLinkButton`: `aria-label` and `tooltip` diverge; copy confirmation is visual-only** — MODERATE · `artifact-row.tsx:158`
`aria-label={url ? "Copy share link" : "Sharing settings"}` is static while `tooltip` carries the transient `copied ? "Copied!" : …`. The single `title` previously delivered the confirmation to both channels; now "Copied!" reaches sighted users only, with no live region. Either use a polite live region or reflect it in the accessible name.

**10. `AttachmentChip`'s remove button has no accessible name** — LOW · `chat-input.tsx:261`
Its only child is `<Close size={10} />` and it carries neither `aria-label` nor `tooltip`, so it announces as an unnamed button — notable given this PR added both to every other icon-only button in this file (`:189`, `:212`, `:225`). The `HOVER_ACTION` swap here is a genuine improvement: the old unscoped `opacity-0 group-hover:opacity-100` left the control unreachable on touch, and `md:`-scoping plus `md:focus-within:opacity-100` fixes both the touch and keyboard cases.

**11. `bind-agent-row` bypasses `CardButton`'s `selected` variant** — LOW · `bind-agent-row.tsx:25`
`highlighted` routes through `className` as a raw `border-foreground` rather than through `selected`. Arguably correct — `highlighted` means "just created", not "picked", so `aria-pressed={selected}` would be a lie — but `cardSelectionVariants` emits `border-border` that twMerge then discards, and the call site hand-writes a token the variant owns. Worth an explicit `emphasized` variant if this recurs.

### Verdict
COMMENT — no correctness defects, but six accessibility/behavioral regressions from the consolidation are worth fixing before merge; the `Button`-inside-`asChild` composition hazard is the one that will bite again as `tooltip` spreads.

---

packages/ui/src/components/ui/button.tsx
```diff
@@ -74,15 +79,23 @@ function Button({
...
+      /* A disabled button fires neither pointer nor focus events, so the
+         tooltip can never open — `title` still reaches assistive tech, which
+         matters most for a hint that explains why the button is disabled. */
+      title={
+        props.disabled && typeof tooltip === "string" ? tooltip : undefined
+      }
       {...props}
     />
   );
+  return tooltip ? <Tooltip content={tooltip}>{button}</Tooltip> : button;
```
dam-code-guardian, line 98: 🟡 **Warning:** With `tooltip` set, this component's root element becomes `Tooltip` — which destructures exactly `{children, content, side, className}`, with no rest spread and no ref forwarding. Any parent cloning this `Button` via `asChild` therefore hands its props to a component that drops every one of them.

First bitten at `files-panel.tsx:43`: while `controller.isUploading`, `DropdownMenuTrigger asChild` clones `Tooltip` and discards `ref`, `onPointerDown`, `onKeyDown`, `id`, `aria-haspopup="menu"`, `aria-expanded` and `data-state` — the menu loses its Popper anchor and the button stops being announced as a menu button. The damage is currently masked only because the same condition also sets `disabled`; that coupling is incidental and free to drift.

The hazard generalizes, since `tooltip` now lives on the shared `Button` and every `*Trigger asChild` in the codebase takes a `Button` child. Either forward rest props and a ref through `Tooltip` (restoring what the removed `<span className="inline-flex">` provided structurally), or have `Button` refuse to wrap when it is itself slotted. `experiment-dock-panel.tsx:130` shows the safe order — `Tooltip` *outside* the Radix trigger, which merges and re-forwards unknown props.

packages/ui/src/components/ui/checkbox.tsx
```diff
@@ -25,4 +25,60 @@ function Checkbox({
...
+  testId,
+  className,
+  id,
+  ...props
+}: CheckboxItemProps) {
+  const generatedId = React.useId();
+  const controlId = id ?? generatedId;
+  const descriptionId = `${controlId}-description`;
+  return (
+    <label
```
dam-code-guardian, line 56: 🟡 **Warning:** The docstring calls this "the multi-select counterpart of `RadioGroupItem`", but it diverges on the one contract that component's comment names explicitly: *"`aria-label` keeps the description out of the accessible name, where name-from-content would otherwise run label and description together."*

No `aria-label` is set here. The `<label htmlFor>` wraps both the label span and the description span, and Radix's `Checkbox.Root` renders `<button role="checkbox">` — a labelable element — so the accessible name becomes the label's entire text content: label **and** description concatenated. `aria-describedby` then announces the description a second time.

Worst at `create-form.tsx:164`, where `scopeDescription()` returns full sentences, so the checkbox announces as `agents:manage Full agent config and lifecycle: …` and repeats the sentence as its description.

Mirror `RadioGroupItem`:
```suggestion
/** The multi-select counterpart of `RadioGroupItem`. `className` styles the
 *  row, not the box. `aria-label` keeps the description out of the accessible
 *  name, where name-from-content would otherwise run label and description
 *  together. */
function CheckboxItem({
```

packages/ui/src/components/icon-rail.tsx
```diff
@@ -152,6 +162,13 @@ function RailItem({ label, icon: Icon, active, badge, navigate }: Destination) {
       <IconWithBadge icon={Icon} badge={badge} />
     </button>
   );
+  return hoverLabel ? (
+    <Tooltip content={label} side="right">
```
dam-code-guardian, line 140: 🟡 **Warning:** `RailItem` previously carried `title={label}` unconditionally. The `title` was removed for every destination, but `hoverLabel: true` was added to only three of six — `experiments` (:58), `knowledgeBases` (:69), `artifacts` (:77). `sandboxes` ("Home"), `inbox` ("Email") and `settings` ("Settings") now render no tooltip at all, so in the collapsed 56px rail a sighted pointer user has no way left to learn what those icons mean.

`aria-label` keeps screen readers correct, so this is purely visual — and it reads as an oversight rather than a decision, since nothing distinguishes those three destinations. Either set the flag on all six, or drop the flag and always render the `Tooltip`.

packages/ui/src/modules/sandboxes/components/skills/skill-row.tsx
```diff
@@ -59,29 +61,31 @@ export function SkillRow({
...
-              disabled={disabled}
-              title="Upstream changed since install — update to the latest version"
-              className={cn(
-                badgeVariants({ variant: "info", size: "sm" }),
-                "shrink-0 gap-1 px-2 text-[11px] transition-opacity hover:opacity-80 disabled:opacity-50",
-              )}
-            >
-              <Renew size={11} /> Update
-            </button>
+            <Tooltip content="Upstream changed since install — update to the latest version">
```
dam-code-guardian: 🟡 **Warning:** This drift hint moved from a native `title` to a `Tooltip` wrapping a raw `<button disabled={disabled}>`. A disabled button fires neither pointer nor focus events, so the tooltip can never open — a fact this PR documents twice itself (`button.tsx:88`, `experiment-dock-panel.tsx:325`). The old native `title` *did* render on a disabled button, so the hint is now lost precisely in the state where "why can't I click this?" is being asked.

`Button`'s own mitigation (native `title` when `disabled && typeof tooltip === "string"`) is unavailable here because this is a hand-rolled `<button>` styled with `badgeVariants`. Routing it through `Button` would inherit the mitigation.

packages/ui/src/modules/sandboxes/components/steps/selectable-card.tsx
```diff
@@ -16,14 +17,7 @@ export function SelectableCard({
...
   return (
-    <div
-      className={cn(
-        "relative rounded-lg border p-4 transition-colors",
-        selected
-          ? "border-foreground bg-muted/60"
-          : "border-border bg-card hover:bg-muted/40",
-      )}
-    >
+    <div className={cn(cardSelectionVariants({ selected }), "relative p-4")}>
```
dam-code-guardian, line 20: 🟡 **Warning:** The selected state lost its background fill. Before: selected → `border-foreground bg-muted/60`, unselected → `border-border bg-card hover:bg-muted/40`. `cardSelectionVariants` bakes in `CARD_SURFACE` (which carries `bg-card`) and its `selected: true` branch adds only `border-foreground` — so both states now paint `bg-card` and selection is signalled by border color alone. The harness and workload pickers lose their selected tint.

This is a case where the shared contract can't express the call site's need: `cardSelectionVariants` has no "selected also gets a fill" branch. Either add one, or keep the fill here on top of the variant.

packages/ui/src/modules/artifacts/components/experiments-section.tsx
```diff
@@ -121,16 +121,12 @@ export function ExperimentsSection({
...
-      <div
-        role="button"
-        tabIndex={0}
-        onClick={() => setOpen((o) => !o)}
-        onKeyDown={(e) => {
-          if (e.key === "Enter") setOpen((o) => !o);
-        }}
-        className="flex cursor-pointer select-none items-center gap-2.5 px-3.5 py-2.5 transition-colors hover:bg-muted/60"
+      <DisclosureToggle
+        open={expanded}
```
dam-code-guardian: 🟡 **Warning:** `expanded = open || searching` (:110), but `DisclosureToggle` receives `open={expanded}` while `onToggle` flips `open`. While a search filter is active, `expanded` is pinned true: clicking the header mutates `open` invisibly and the section never collapses.

The stale-state shape predates this PR, but migrating to `DisclosureToggle` newly exposes `aria-expanded={expanded}` (`disclosure.tsx:53`), which promotes a cosmetic quirk into an assistive-tech lie — announced as expanded, with a control that verifiably does nothing. Either disable the toggle while `searching`, or derive one source of truth for the open state.

packages/ui/src/lib/clickable.ts
```diff
@@ -0,0 +1,20 @@
...
+/** Button semantics for a row that can't be a real `<button>` because it wraps
+ *  its own menu or link. Keys are ignored unless they land on the row itself,
+ *  so a nested control keeps its own Enter and Space. `undefined` leaves the
+ *  element inert, with no stale `role` or tab stop. */
+export function clickableProps(onActivate: (() => void) | undefined) {
+  if (!onActivate) return {};
+  return {
+    role: "button",
+    tabIndex: 0,
+    onClick: onActivate,
```
dam-code-guardian, line 12: 🟢 **Suggestion:** `onKeyDown` correctly bails on `event.target !== event.currentTarget`, but `onClick: onActivate` has no equivalent guard, so it still fires on clicks bubbling from any descendant.

The asymmetry is load-bearing: every call site must remember a manual `onClick={(e) => e.stopPropagation()}` wrapper (`artifact-row.tsx:99`, `chat-artifacts-panel.tsx:118`, `lineage-card.tsx:77`, plus `:47` and `artifact-row.tsx:133` on individual controls) — and this PR *removed* the paired `onKeyDown` stopPropagation from those wrappers on the strength of the new keyboard guard. All four current sites are covered, so there's no live bug, but the docstring reads as though propagation is handled centrally when only half of it is. Guarding `onClick` the same way would make the contract self-enforcing rather than convention-dependent.

packages/ui/src/modules/sessions/components/chat-input.tsx
```diff
@@ -257,7 +262,10 @@ function AttachmentChip({
         variant="destructive"
         size="icon"
         onClick={onRemove}
-        className="absolute -top-1.5 -right-1.5 h-5 w-5 rounded-full opacity-0 group-hover:opacity-100 transition-opacity"
+        className={cn(
+          "absolute -top-1.5 -right-1.5 h-5 w-5 rounded-full",
+          HOVER_ACTION,
```
dam-code-guardian, line 268: 🟢 **Suggestion:** This remove button's only child is `<Close size={10} />` and it carries neither `aria-label` nor `tooltip`, so it announces as an unnamed button — notable given this PR added both to every other icon-only button in this file (`:189`, `:212`, `:225`).

The `HOVER_ACTION` swap itself is a genuine improvement here: the old unscoped `opacity-0 group-hover:opacity-100` left the control unreachable on touch, and `md:`-scoping plus `md:focus-within:opacity-100` fixes both the touch and keyboard cases.

packages/ui/src/modules/egress-rules/components/agent-egress-editor.tsx
```diff
@@ -632,10 +632,12 @@ function VerdictBadge({ verdict }: { verdict: EgressRuleView["verdict"] }) {
   );
 }
 
-function SourceTag({ label, title }: { label: string; title: string }) {
+function SourceTag({ label, hint }: { label: string; hint: string }) {
```
dam-code-guardian, line 637: 🟢 **Suggestion:** `SourceTag` passes `label={label}` to `HintTooltip` while the wrapped `Badge` renders `{label}`, so `aria-label` merely duplicates the visible text — and every egress rule row gains a keyboard tab stop for a decorative badge. A screen-reader user reaching it hears *"manual, button"* with no hint that it does nothing and no access to the tooltip text, since Radix only wires `aria-describedby` while the tooltip is open.

This also cuts against the PR's own goal of dropping tooltips that restate a control's label. Consider `tabIndex={0}` on a non-button element with the hint as the accessible *description* rather than a `<button>` whose name echoes its content.

packages/ui/src/modules/artifacts/components/experiments-section.tsx
```diff
@@ -121,16 +121,13 @@ export function ExperimentsSection({
...
+        onToggle={() => setOpen((o) => !o)}
+        disabled={searching}
+        chevronClassName="text-muted-foreground"
+        className="w-full cursor-pointer select-none gap-2.5 px-3.5 py-2.5 transition-colors hover:bg-muted/60"
```
dam-code-guardian: 🟡 **Warning:** The `disabled={searching}` fix is incomplete on the visual side. The className still carries `cursor-pointer` and `hover:bg-muted/60`, and CSS `:hover`/`cursor` both still apply to a disabled `<button>` — only mouse *events* are suppressed. So while a search is active the header keeps a pointer cursor and a hover tint, invites a click, and does nothing: the same dead-control complaint the fix commit set out to address, relocated from assistive tech to sighted mouse users.

Separately, if `searching` flips true while this toggle has focus, the browser drops focus to `<body>`, losing the keyboard user's position mid-search.

Either keep the button enabled and neutralize the affordances, or let the search own the disclosure — render the section expanded with no toggle at all while searching, rather than an inert one.

packages/ui/src/modules/egress-rules/components/agent-egress-editor.tsx
```diff
@@ -602,16 +605,13 @@ function PreviewPresetRow({ row }: { row: PreviewRow }) {
...
       />
-      <Badge
-        size="sm"
-        variant="accent"
-        className="uppercase tracking-wider"
-        title="This rule will be saved on commit"
-      >
-        preview
-      </Badge>
+      <HintTooltip label="preview" content="This rule will be saved on commit">
```
dam-code-guardian: 🟢 **Suggestion:** The fix commit's stated reason for reverting `SourceTag` to an inert badge ("it was minting a tab stop per egress row whose name only repeated the badge") applies verbatim here, and at `built-in-skills-group.tsx:59` (`label="Modified"` wrapping a badge reading `Modified`). Both were converted *from* native `title` by this PR and left as `HintTooltip`.

The result is two contradictory patterns for one problem 25 lines apart in the same file — `HintTooltip` here, native `title` at `:635` — which is the shape that gets copied inconsistently next.

Either revert these two for consistency with `SourceTag`, or give `HintTooltip` a separate-icon-trigger mode like `schedule-form-modal.tsx:267` legitimately uses, where the trigger is an `<Information>` icon with no visible text to repeat.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `f68e995`

## PR #3132: refactor(ui): fix interaction semantics and the last hand-rolled surfaces
**Author:** kapetr | **Branch:** refactor/3054-ui-interaction-semantics → main | **Changes:** +718 −490 (75 files)

### Summary
Re-review at `f68e995`. One author fix commit (`e77428c7`) addressed 5 of the 11 prior findings; the rest of the delta is two merges from main, one of which rewrote `HOVER_ACTION` onto a `hover-capable` custom variant. That rewrite is sound — the variant is defined at `App.css:10`, all 7 call sites keep a `group` ancestor, and each reveal is (0,2,0) against the (0,1,0) base, so dropping `md:` didn't invert the cascade and touch devices now get always-visible actions at any width.

### Changes since last review
Previous HEAD: `9e88940` (2026-08-03T14:45:12Z) — verdict COMMENT

- ✅ **Fixed:** `CheckboxItem` sets `aria-label={label}`, which wins over the wrapping `<label>` in the accname algorithm; the `aria-describedby` merge is correct and caller-overridable (`packages/ui/src/components/ui/checkbox.tsx:66`)
- ✅ **Fixed:** skill drift hint falls back to native `title` while disabled, reaching the DOM on the raw `<button>` (`packages/ui/src/modules/sandboxes/components/skills/skill-row.tsx:74`)
- ✅ **Fixed:** `SourceTag` is an inert `<Badge title={hint}>` again — `Badge` spreads rest onto its `<span>`, so the per-row tab stop is gone (`packages/ui/src/modules/egress-rules/components/agent-egress-editor.tsx:635`)
- ✅ **Fixed:** attachment remove button has an accessible name (`packages/ui/src/modules/sessions/components/chat-input.tsx:265`)
- ✅ **Fixed:** `DisclosureToggle` no longer claims an `aria-expanded` it can't change while a search pins the section open (`packages/ui/src/modules/artifacts/components/experiments-section.tsx:127`)
- 🔁 **Still present:** `Button`'s `tooltip` makes `Tooltip` the root, which forwards neither rest props nor ref, so `DropdownMenuTrigger asChild` drops them (`packages/ui/src/components/ui/button.tsx:98`, bitten at `files-panel.tsx:43`)
- 🔁 **Still present:** Home, Email and Settings have no hover label — `hoverLabel` set on only 3 of 6 destinations (`packages/ui/src/components/icon-rail.tsx:45`)
- 🔁 **Still present:** selection is border-only; `cardSelectionVariants` bakes `bg-card` into both branches (`packages/ui/src/components/ui/card.tsx:17`)
- 🔁 **Still present:** `clickableProps` guards `onKeyDown` but not `onClick` (`packages/ui/src/lib/clickable.ts:12`)
- 🔁 **Still present:** `ShareLinkButton`'s copy confirmation is visual-only — no live region (`packages/ui/src/modules/artifacts/components/artifact-row.tsx:161`)
- 🔁 **Still present:** `highlighted` routes through `className` as raw `border-foreground` (`packages/ui/src/modules/agents/components/bind-agent-row.tsx:30`)

### Findings
- 🟡 **Warning:** `disabled={searching}` leaves `cursor-pointer` and `hover:bg-muted/60` on the button (`packages/ui/src/modules/artifacts/components/experiments-section.tsx:129`)
- 🟢 **Suggestion:** two `HintTooltip` badge-wrapping sites now contradict the `SourceTag` revert 25 lines away (`packages/ui/src/modules/egress-rules/components/agent-egress-editor.tsx:610`)

**Verdict:** aligned

No architecture-doc drift. The diff touches only `packages/ui/` presentation code and no doc file; no documented subsystem responsibility changed, no new long-lived component was introduced, and `docs/adrs/` is untouched. The `components.json` `iconLibrary` flip is a library-level codegen choice, which the guidelines place below architecture-page scope.

**Mode: review** — no architecture findings. 🔁 The 3 helpers were clean at the previous review and remain so (leaf modules, placement consistent with in-package precedent, clean under `strict`/`isolatedModules`).

The merge-driven rewrite of `hover-action.ts` was re-checked on its own terms and both docstring claims hold: `hover-capable` is genuinely defined (`App.css:10`, valid Tailwind v4 CSS-first syntax on the sheet imported at `main.tsx:1`), and the specificity argument is now actually a specificity argument rather than the ordering argument the `md:` version relied on. Scoping only the base to `hover-capable` is the right asymmetry — on a pointer-less device the base never applies, so the action stays visible. No `md:opacity-0` reveal survives anywhere in the package.

_(`tsc`/`eslint` could not run — no `node_modules` in the clone and installs are disallowed; conclusions are from source and config inspection. CI typecheck remains the authority.)_

Re-reviewed the 5-commit delta against the 11 prior findings, plus the merged-in `markdown-code-block.tsx` and the `hover-action.ts` rewrite. 🔁 6 findings from the previous review still present (see review at `9e88940`); 5 verified genuinely fixed, per the delta above.

**New — MODERATE · `experiments-section.tsx:127-129`** — the `disabled={searching}` fix is visually incomplete. The className still carries `cursor-pointer` and `hover:bg-muted/60`, and CSS `:hover`/`cursor` both still apply to a disabled `<button>` (only mouse *events* are suppressed). While a search is active the header keeps a pointer cursor and a hover tint, invites a click, and does nothing — the same dead-control complaint the commit set out to fix, relocated from assistive tech to sighted mouse users. Separately, if `searching` flips true while the toggle has focus, the browser drops focus to `<body>`, losing the keyboard user's position mid-search. Fix: add `disabled:cursor-default disabled:hover:bg-transparent` alongside the prop (keeping the button enabled so focus survives), or let the search own the disclosure — render the section expanded with no toggle at all while searching, rather than an inert one.

**New — LOW · `agent-egress-editor.tsx:610`, `built-in-skills-group.tsx:59`** — the fix commit's stated reason for reverting `SourceTag` ("minting a tab stop per row whose name only repeated the badge") applies verbatim to two `HintTooltip` sites this PR converted *from* native `title` and left in place: `label="preview"` wraps a badge reading `preview`, `label="Modified"` wraps one reading `Modified`. The result is two contradictory patterns for one problem 25 lines apart in the same file (`:610` HintTooltip vs `:635` native `title`), which is what gets copied inconsistently next. Either revert these two for consistency with `SourceTag`, or give `HintTooltip` a separate-icon-trigger mode as `schedule-form-modal.tsx:267` legitimately uses (an `<Information>` icon with no visible text to repeat).

Two of the skill's candidate findings I checked and am **not** raising: `markdown-code-block.tsx:34-35` pairing `aria-label="Copy code"` with `tooltip="Copy code"` matches the PR's own deliberate convention (14 sites do this; `Button`'s JSDoc states the tooltip "reads as a description, so an icon-only button still needs an `aria-label`", and `9e88940` removed only tooltips restating a *visible text* label, not an `aria-label`) — and a JSDoc-placement/interface-extraction nit on `disclosure.tsx` is human-readability only.

### Verdict
COMMENT — the fix commit resolved 5 findings genuinely, and the `hover-capable` rewrite verifies clean; 6 carryovers plus one incomplete fix remain, none a correctness defect. The `Button`-inside-`asChild` hazard is still the one that will bite again as `tooltip` spreads.

---

packages/ui/src/components/ui/disclosure.tsx
```diff
@@ -32,6 +32,9 @@ export function DisclosureChevron({
 export function DisclosureToggle({
   open,
   onToggle,
+  /** For a section held open by something else, so `aria-expanded` doesn't
+   *  advertise a state this button can't change. */
+  disabled,
```
dam-code-guardian: 🟢 **Suggestion:** This prop is now dead code, and its comment documents the approach the fix commit rejected.

This PR introduced `disabled` (it has zero occurrences in `main`), `experiments-section` was its only consumer, and the fix commit removed that usage in favour of rendering a static header while searching. Nothing passes `disabled` to `DisclosureToggle` any more.

What is left behind is a prop whose docstring reads as an endorsement of the pattern the commit message explicitly rejects (*"a static header rather than a disabled toggle that kept a pointer cursor and hover"*). The concrete risk: the next caller with a section pinned open by something else finds this prop, reads the comment as guidance, and reintroduces exactly the dead-control defect this commit fixed — `cursor-pointer` and `hover:` still apply to a disabled `<button>`, and flipping `disabled` while the toggle has focus drops focus to `<body>`.

Either drop the prop and its comment, or keep it and point the comment at the static-header pattern instead.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `8219d6c`

## PR #3132: refactor(ui): fix interaction semantics and the last hand-rolled surfaces
**Author:** kapetr | **Branch:** refactor/3054-ui-interaction-semantics → main | **Changes:** +757 −513 (74 files)

### Summary
Re-review at `8219d6c`. One author fix commit addressed all 5 findings raised at `f68e995` — and the `Tooltip` prop/ref forwarding and the `experiments-section` rework both verify correct on their actual mechanics, not just their stated intent. Three carryovers remain, none a correctness defect; the one new finding is dead code left behind by an otherwise better-than-requested fix.

### Changes since last review
Previous HEAD: `f68e995` (2026-08-03T17:35:56Z) — verdict COMMENT

- ✅ **Fixed:** `Tooltip` is `forwardRef` and spreads rest onto the `asChild` Radix trigger, so a slotted `Button` keeps its Popper anchor, `ref` and trigger a11y; the `:focus-visible` guard survives the `onFocus` composition (`packages/ui/src/components/ui/tooltip.tsx:60`)
- ✅ **Fixed:** `hoverLabel` dropped and `Tooltip` always rendered, so all six rail destinations get a hover label; the mobile bar renders `{label}` as text and needs none (`packages/ui/src/components/icon-rail.tsx:138`)
- ✅ **Fixed:** `cardSelectionVariants` selected branch is `border-foreground bg-muted/60`; twMerge resolves the earlier `bg-card` last-wins, restoring the fill for both consumers (`packages/ui/src/components/ui/card.tsx:20`)
- ✅ **Fixed:** while searching, a static header replaces the toggle entirely — no `cursor-pointer`, no hover tint, no `aria-expanded`, no focus to drop. Stronger than the suggested `disabled:` classes (`packages/ui/src/modules/artifacts/components/experiments-section.tsx:145`)
- ✅ **Fixed:** both remaining `HintTooltip` badge sites reverted to inert `Badge title=`, matching `SourceTag`; the codebase now has one pattern across 6 sites (`packages/ui/src/modules/egress-rules/components/agent-egress-editor.tsx:609`)
- 🔁 **Still present:** `clickableProps` guards `onKeyDown` but not `onClick`; all four sites still carry manual `stopPropagation`, so still no live bug (`packages/ui/src/lib/clickable.ts:12`)
- 🔁 **Still present:** `ShareLinkButton`'s copy confirmation is visual-only — no live region, static `aria-label` (`packages/ui/src/modules/artifacts/components/artifact-row.tsx:161`)
- 🔁 **Still present:** `highlighted` routes through `className` as raw `border-foreground` (`packages/ui/src/modules/agents/components/bind-agent-row.tsx:30`)

### Findings
- 🟢 **Suggestion:** `DisclosureToggle`'s `disabled` prop is now dead, and its docstring documents the approach this commit rejected (`packages/ui/src/components/ui/disclosure.tsx:37`)

**Verdict:** aligned

No architecture-doc drift. The diff is confined to `packages/ui/` and touches no file under `docs/`, so the staleness, ADR-leak, volatile-content and cross-reference checks have no anchor, and no ADR changed status. The two live checks are clean: no subsystem behavior or responsibility changed (the diff contains no tRPC, query, router, Zod, or WebSocket/ACP surface on either side), and the four new files are in-package UI primitives, not long-lived components owing a page.

**Mode: review** — no architecture findings.

The three routed `.ts` helpers are **byte-identical to their last-reviewed state** — `git diff --quiet f68e995 8219d6ca` reports each unchanged, with a passing control. Re-verified rather than assumed: all three remain leaves (`hover-action.ts` and `external-link.ts` import nothing; `clickable.ts` imports only a type), all 25 importers live inside `packages/ui` so no cross-package boundary widens, and `strict`/`isolatedModules` hold via `dev-config/tsconfig`.

The fix commit affects none of their contracts — no changed file imports any of the three. The one indirect coupling worth checking was `experiments-section` → `FolderGroup` (a `HOVER_ACTION` consumer): the commit restructured only the section header, and `folder-group.tsx:78` still supplies the `group` ancestor the reveal depends on.

_(`tsc`/`eslint` could not run — no `node_modules` in the clone and installs are disallowed; conclusions are from source and config inspection. CI typecheck remains the authority.)_

Re-reviewed the fix commit against the 5 claimed fixes and 6 carryovers. All 5 fixes are genuine — traced through the clone chain, not taken from the docstrings. 🔁 3 findings from the previous review still present (see review at `f68e995`); details of each fix are in the delta above.

Two mechanism notes worth recording, since both were the crux of a prior finding:

**`Tooltip` forwarding (prior HIGH) — resolved.** With `tooltip` set, `Button`'s root is `Tooltip`, so `DropdownMenuTrigger asChild` at `files-panel.tsx:43` clones *that* element. Because `Tooltip` now spreads `...rest` onto `TooltipPrimitive.Trigger`, which is itself `asChild`, Radix's own Slot merge carries the injected `ref`, `onPointerDown`, `onKeyDown`, `id`, `aria-haspopup="menu"`, `aria-expanded` and `data-state` onto the real `<button>`. `files-panel.tsx:43` is still the codebase's only instance of the hazard shape.

**`experiments-section` (prior MODERATE) — resolved by splitting the surface**, not by patching `disabled:` classes onto an inert control. No sibling shares the defect: `FolderGroup` drives collapse from remount-keyed `useState`, which is real state rather than a pinned lie.

**New — LOW · `disclosure.tsx:37`, `:46`, `:57`** — `DisclosureToggle`'s `disabled` prop is now dead code. This PR introduced it (`main` has zero occurrences), `experiments-section` was its only consumer, and fix 4 removed that usage — a multiline search for `<DisclosureToggle … disabled` returns nothing against a passing control. What remains is a prop whose docstring advertises the rejected approach: *"For a section held open by something else, so `aria-expanded` doesn't advertise a state this button can't change."* The commit message itself prefers a static header *"rather than a disabled toggle."* The risk is the next caller with a pinned-open section finding the prop, reading the docstring as endorsement, and reintroducing the defect this commit fixed. Either drop the prop and comment, or rewrite the comment to point at the static-header pattern.

One latent hazard checked and **not** raised as live: `Tooltip` destructures `className` out of `rest` and routes it to `TooltipContent`, so an `asChild` parent injecting a `className` while cloning a `Tooltip` would style the popup instead of the trigger. No `*Trigger asChild` in the codebase passes a `className`, so nothing is broken today — worth a docstring note rather than a finding.

### Verdict
COMMENT — every finding from the last round is genuinely fixed, two of them more thoroughly than suggested; what remains is three known carryovers and one dead prop whose stale docstring invites the regression this commit just removed.

---

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `2c246a8`

## PR #3132: refactor(ui): fix interaction semantics and the last hand-rolled surfaces
**Author:** kapetr | **Branch:** refactor/3054-ui-interaction-semantics → main | **Changes:** +752 −513 (73 files)

### Summary
Re-review at `2c246a8`. One commit since the last review, dropping `DisclosureToggle`'s dead `disabled` prop — the strongest possible fix for that finding: `disclosure.tsx` is now byte-identical to `main`, so the file has left the PR's changed-file list entirely. Three known carryovers remain, none a correctness defect, and nothing new.

### Changes since last review
Previous HEAD: `8219d6c` (2026-08-04T08:05:51Z) — verdict COMMENT

- ✅ **Fixed:** `DisclosureToggle`'s `disabled` prop and the docstring that endorsed the rejected approach are gone; the file now matches `main` byte-for-byte, no consumer or test referenced it, and the static-header pattern it should have pointed at stays self-documenting at `experiments-section.tsx:143` (`packages/ui/src/components/ui/disclosure.tsx`)
- 🔁 **Still present:** `clickableProps` guards `onKeyDown` but not `onClick`; still no live bug — the one site without manual `stopPropagation` wraps no nested control (`packages/ui/src/lib/clickable.ts:12`)
- 🔁 **Still present:** `ShareLinkButton`'s copy confirmation is visual-only — no live region, static `aria-label` (`packages/ui/src/modules/artifacts/components/artifact-row.tsx:161`)
- 🔁 **Still present:** `highlighted` routes through `className` as raw `border-foreground` rather than the `selected` variant (`packages/ui/src/modules/agents/components/bind-agent-row.tsx:30`)

### Findings

**Verdict:** aligned

No architecture-doc drift. The diff touches no file under `docs/` on either side, so the staleness, ADR-leak, volatile-content, cross-reference and ADR-state checks have no anchor. The two live checks are clean: the `ui` subsystem page (`docs/architecture/platform-topology.md:71`) describes protocol- and responsibility-level concerns (React + Vite SPA, tRPC over HTTP, one ACP WebSocket per session) and this diff changes none of them — it contains no tRPC, query, router, Zod or WebSocket/ACP surface; and the four added files are in-package UI primitives, not long-lived components owing a page. This round's delta is a five-line dead-code deletion in one component.

**Mode: review** — no architecture findings.

The three routed `.ts` helpers are **byte-identical to their last-reviewed state** — verified per file with `git diff --quiet 8219d6ca 2c246a85`, against a passing control (`disclosure.tsx` correctly reports changed). No changed file imports any of the three, so no contract is affected. Re-verified rather than assumed: all three remain leaves (`hover-action.ts` and `external-link.ts` import nothing; `clickable.ts` imports only a type), all 27 importers live inside `packages/ui` so no cross-package boundary widens, and `strict`/`isolatedModules` hold via `packages/dev-config/tsconfig.base.json`. The three-layer rule is scoped to server modules; `packages/ui` is a client package, so no layer assignment is owed.

_(`tsc`/`eslint` could not run — no `node_modules` in the clone and installs are disallowed; conclusions are from source and config inspection. CI typecheck remains the authority.)_

Re-reviewed the single-commit delta against the 1 new finding and 3 carryovers from `8219d6c`. 🔁 3 findings from the previous review still present (see review at `8219d6c`); the new finding is genuinely fixed.

**Prior LOW (`disclosure.tsx` dead `disabled` prop) — resolved completely.** Traced the mechanics, not the commit message. `disabled` is absent from the entire file — destructure, prop type, JSDoc, and the `<button>` attribute — with a passing control in the same file, so the empty probe is a real absence. Best available confirmation: the file is now **byte-identical to `main`**, which is why it no longer appears among the PR's 73 changed files (control: `main` vs `8219d6c` differs by exactly +5 lines). The PR introduced the prop and has fully retracted it.

**Nothing orphaned or broken.** All nine `DisclosureToggle` call sites dumped in full — none passes `disabled`, and with it gone from the prop type `strict` TS would reject any that did. No test, snapshot or story references `DisclosureToggle` (control: 22 test files exist and are reachable). The guidance the deleted docstring carried survives where a future caller will look: `experiments-section.tsx:143-150` renders a static header while a search pins the section open, with its own explanatory comment.

**Sibling check — one pre-existing instance, out of scope.** `registry-credential-section.tsx:48` has the same pinned-open shape (`expanded = partial || (disclosureOverride ?? …)` feeding `open={expanded}`), so while `partial` is true the toggle advertises an `aria-expanded` its `onToggle` cannot change. The file is **not in this PR** and `main` already contains the identical lines (verified via the contents API — the shallow clone's merge-base is grafted and answered unreliably). Not a finding against this change, but the pattern this PR established (`searching ? static header : toggle`) applies to it directly if someone wants it.

**Carryovers re-verified.** `clickable.ts:12` unchanged — and worth recording why it stays a suggestion: `sandbox-group-card.tsx` is the one call site with no manual `stopPropagation`, and its clickable `<div>` wraps only `<span>`s, with `LineageCard` a sibling outside it, so no descendant click can bubble into `onActivate`. `artifact-row.tsx:161` unchanged — no `aria-live`, `role="status"` or `sr-only` anywhere in the file. `bind-agent-row.tsx:30` unchanged — `cardSelectionVariants` still owns `border-foreground` at `card.tsx:20`.

**New findings in the delta:** none. A five-line deletion returning a file to its `main` state, with every consumer and test verified clear, has no surface to introduce one.

### Verdict
COMMENT — the last finding is fixed as thoroughly as it can be (the file is back to `main` byte-for-byte, with every consumer and test verified clear); what remains is three known non-blocking carryovers and no new findings.

---

PetrBulanek (approved):
**Keyboard users can't open a sandbox, a session, or a file**

Three clickable rows are still plain `div`s with only `onClick` — no `role`, no `tabIndex`, no key handling. Two of them are files this PR already touches:

- `packages/ui/src/modules/agents/components/agent-row.tsx:56` — the Home page sandbox list
- `packages/ui/src/modules/sessions/components/session-row.tsx:124` — the sessions sidebar
- `packages/ui/src/modules/files/components/file-row.tsx:92` — the file tree

I tabbed through Home in the running app. The order is: brand → 6 nav icons → **Create sandbox** → 6 × **Sandbox actions**. Nothing in that order opens a sandbox, and the ⋮ menu only offers Wake / Restart / Pause / Stop / Delete. So without a mouse there is no way into a sandbox at all — the front door of the product is closed. Same story for picking a session in the sidebar and for opening a file in the tree.

This is pre-existing, not a regression. But it's the exact defect class this PR set out to close ("Clickable rows share one helper and respond to both Enter and Space"), the helper now exists, and the fix is one spread per row:

```tsx
{...clickableProps(onSelect)}

PR #3134, opened 2026-08-03
Remove dam-run executor (local exec shim) and make the api-server multi-replica safe

JanPokorny:
### dam-run → local exec shim
The remote Run-executor machinery had been disabled since the RWO storage cutover (#2989) — this finishes the removal (−2,583 lines):
- `dam-run` is now a compatibility shim that runs the command as a regular local process (inherited stdio/cwd/env; trace context still folds into the session's metrics via inherited `TRACEPARENT`)
- removed: api-server `runs` module + harness run relay, agent-runtime exec-only mode, controller `Run` reconciler + CRD + informer wiring, RBAC grants, generated CR types
- runs CRD removal is allow-listed in the crd-compat gate (mirrors the forks precedent)

### Multi-replica api-server
Fixes every single-instance assumption found in an audit of the api-server:
- **Periodic work**: all remaining per-replica interval sweepers (runtime outbox, approvals delivery, OAuth refresh, agent/invocation/experiment reapers) migrated onto the shared BullMQ periodic-jobs queue — one execution per period across replicas
- **Resize gate**: per-owner check+patch serialization moved from an in-process map to a Postgres advisory xact lock
- **Session presence**: per-replica Redis TTL keys + heartbeat; the active-session annotation clears only when no replica holds sessions, and a reconcile job sweeps pins orphaned by crashed replicas (replaces the replica-unsafe boot-time blanket clear)
- **OAuth/bind handoffs**: connections OAuth engine, Slack/Telegram pending OAuth flows, and bind-flow stores moved onto Redis-backed TTL stores so the callback leg may land on any replica
- **MCP sessions**: harness Service pins each per-agent gateway to one replica via ClientIP session affinity
- **Helm**: new `apiServer.replicas` value (default 1)

## Verification
- full `mise run check` green; 738 api-server + 208 agent-runtime unit tests pass; controller build/vet/staticcheck/tests clean
- `mise run cluster:install`, then redeployed with `--set apiServer.replicas=2`: both replicas Ready, logs clean, and Redis shows all 9 periodic queues executing exactly once per period across the two replicas
- shim: `dam-run echo hello` works, exit codes and spawn errors (127) propagate


dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `102c4070`

## PR #3134: Remove dam-run executor (local exec shim) and make the api-server multi-replica safe
**Author:** JanPokorny | **Branch:** refactor/dam-run-shim-multi-replica#locki-sugikb52 → main | **Changes:** +647 −3052 (97 files)

### Summary
Removes the disabled remote Run-executor machinery (api-server `runs` module, agent-runtime exec mode, controller `Run` reconciler + CRD) and reduces `dam-run` to a local exec shim; then closes the api-server's single-replica assumptions by moving interval sweepers onto a shared BullMQ periodic-jobs queue, OAuth/bind handoffs onto Redis TTL stores, per-owner resize serialization onto a Postgres advisory lock, and session presence onto per-replica Redis keys with a reconcile job. The removal half is clean and well-sequenced; the findings below are all in the new concurrency primitives.

### Findings
- 🟡 **Warning:** `reconcile()` can clear a live agent's session pin during a Redis blip (`packages/api-server/src/apps/api-server/session-presence.ts:119`)
- 🟡 **Warning:** presence key TTL is refreshed but the annotation is never re-asserted, so a lost annotation write pins nothing until release (`packages/api-server/src/apps/api-server/session-presence.ts:75`)
- 🟡 **Warning:** two `periodicJobs.register` calls are fire-and-forget, so a registration failure is an unhandled rejection and that job silently never runs (`packages/api-server/src/apps/api-server/app.ts:262`)
- 🟡 **Warning:** `hashtext()` collisions make the resize lock serialize unrelated owners; the SQL is also unguarded against a `hashtext` search-path shadow (`packages/api-server/src/core/xact-lock.ts:12`)
- 🟢 **Suggestion:** OAuth `PendingFlow` writes the provider's `clientSecret` into Redis, widening where that secret rests (`packages/api-server/src/modules/connections/infrastructure/oauth-engine.ts:157`)
- 🟢 **Suggestion:** the new cross-replica primitives ship without unit tests (`packages/api-server/src/core/ttl-store.ts`)
- ✅ **Looks good:** the `runs` CRD removal is correctly allow-listed in the crd-compat gate and the reconciler used no finalizers, so no orphaned resources can outlive it.

**Verdict:** minor drift

**Drift**

*Check 1 — `agent-lifecycle.md` still describes the removed boot-time pin clear.* This PR deletes `clearActiveSessions()` and its `[boot] cleared N stale active-session pin(s)` log, replacing it with the `session-presence-reconcile` job, and the new code comment states a boot-time blanket clear *"would wrongly drop pins held by other live replicas."* But `docs/architecture/agent-lifecycle.md:147` and `:185` both still assert "the api-server clears stale pins at boot (a fresh process holds no connections…)". Neither line is touched by the diff — the page was edited only in its `dam-run` section. The stale claim now states the inverse of the design rationale. Suggested replacement for the clause at `:185`: "…and pins orphaned by a replica that died holding sessions are swept by a periodic reconcile: each replica refreshes a short-TTL presence key per agent it holds sessions for, and the reconcile clears the annotation once no replica's key remains." Same for the `Two safety nets…` clause at `:147`.

*Check 2 — `Last verified:` not bumped on any of the 7 edited pages.* PR date is 2026-08-03; all seven carry substantive prose edits, so the trivial-change exemption does not apply: `agent-lifecycle.md` (2026-07-29), `logging.md` (2026-07-29), `metrics.md` (2026-07-30), `observability.md` (2026-07-29), `persistence.md` (2026-07-30), `platform-topology.md` (2026-07-30), `security-and-credentials.md` (2026-07-31).

**Possible drift**

*`budgets.md:50` — "read+check+patch serialized per owner" no longer names the mechanism it relies on.* Still literally true, so this is a judgement call: does the page need to say the serialization is now cross-replica? The removed code comment had explicitly justified the old in-process version by "the api-server running a single replica, the same standing as the controller's single-worker invariant (see budgets.md)" — that back-reference into this very page is what makes it worth a look, especially since `:52`'s "Self-healing races" paragraph reasons about what can slip past the courtesy check.

Checks 3, 5, 6 and 7 produced nothing (no stale `#run-executors-dam-run` / `#dam-run-executor-pods` anchors remain anywhere under `docs/`; no ADR status changed).

Scope: the 54 routed TS/JS files, against the TSEng architecture docs only.

**Finding 1 — six new optional service dependencies.** The rule (`infrastructure.md` § Dependency Injection via Service Factories) is explicit: dependencies are required, never optional, because an optional dep "forces every call site to branch on presence, hides a missing wiring as a silent skip instead of a type error, and leaves the reader unable to tell whether absence is intended or a bug."

| Location | Dep | Consumer branch |
|---|---|---|
| `modules/agents/services/agents-service.ts:523` | `resizeLock?` | `:987` `deps.resizeLock ?? inProcessResizeLock` |
| `modules/channels/infrastructure/telegram-flows.ts:48` | `store?` | `:52` `opts?.store ?? createMemoryTtlStore(…)` |
| `modules/connections/infrastructure/oauth-engine.ts:75` | `pendingStore?` | `:85` `opts?.pendingStore ?? createMemoryTtlStore(…)` |
| `modules/connections/compose.ts:44` | `pendingFlowStore?` | passthrough |
| `modules/channels/infrastructure/slack-flows.ts:21` | `store?` | passthrough |
| `modules/channels/infrastructure/telegram-flows.ts:74` | `store?` | passthrough |

`resizeLock` is the sharpest case: its own JSDoc says "Required whenever `resizeGate` is wired" — a requirement the type cannot express, so the two must be kept consistent by convention, and a missed wiring degrades silently to per-process serialization instead of failing to compile. `pendingFlowStore?` additionally propagates the optionality up through `composeConnectionsAtBoot`, the one place that should resolve it rather than forward it.

Fix per the doc: make each required and let the composition root choose the implementation — `createRedisTtlStore(…)` in production, `createMemoryTtlStore(…)` in tests. The doc closes the test loophole ("don't loosen a required dep to optional to ease setup"), and the new `__tests__/helpers/ttl-store.ts` (`createInspectableTtlStore`) is already exactly that stub, so the required form costs the tests nothing. Making `resizeLock` required also deletes `inProcessResizeLock` and the prose at `:478–480` / `:982–985` explaining when it applies.

**Finding 2 — `resizeGate?`'s pre-existing optionality is amplified.** Out of scope alone, but the new `resizeLock?` is documented as coupled to it, turning one untyped optional into a two-dep invariant no signature enforces (with a runtime `TRPCError` at `:992` when the gate is absent). If only one is fixed, fix `resizeLock`.

**Smallest set of changes:** (1) make the `TtlStore` deps required on `createOAuthEngine`, `createFlowStore`, `createSlackBindFlowStore`, `createTelegramBindFlowStore` and `composeConnectionsAtBoot`, passing the store explicitly from tests and `index.ts`/`app.ts` — clears 5 of 6 and deletes five `??` fallbacks; (2) make `resizeLock` required and delete `inProcessResizeLock` — `composeAgentsModule` already passes `createXactLock(deps.db)` unconditionally, so no call site changes.

Correctly placed and worth keeping: `createXactLock` is referenced only from `modules/agents/compose.ts` (the right and only place for concrete infrastructure), no new import bypasses a module's `index.ts`, and the new `core/` primitives follow the package's established convention for cross-cutting adapters.

### Verdict
COMMENT — the removal half is clean, and the multi-replica primitives are sound in design; the flagged items are edge-case correctness in the presence/reconcile path plus two silent-failure modes worth closing before raising `apiServer.replicas` above 1.

---

packages/api-server/src/apps/api-server/session-presence.ts
```diff
@@ -27,9 +100,34 @@ export function createSessionPresence(repo: {
...
+              // Redis unreachable — clear like the single-replica path did;
+              // reconcile() restores truth once Redis is back.
+              set(agentId, false);
+            });
         }
       };
     },
+
+    async reconcile() {
+      const active = await scanAgentIds(`${KEY_PREFIX}*`);
```
dam-code-guardian, line 123: 🟡 **Warning:** `reconcile()` treats an empty SCAN as authoritative, so a Redis blip clears pins for agents that are genuinely in use.

The two reads are independent: `scanAgentIds` reflects Redis, `listAgentIdsWithAnnotation` reflects K8s. If SCAN returns fewer keys than reality — Redis restarted (AOF `everysec` can lose the last second, and these keys are only ever written by `SET`, never replayed from Postgres), a failover, or the partial-`MATCH` window of a cursor walk — every annotated agent looks unheld and gets `set(agentId, false)`. That unpins live sessions, and the controller's `shouldRun` then hibernates the pod out from under an open chat/terminal/SSH connection, killing in-flight work.

The release path at `:103` is careful here (it re-checks `anyReplicaHolds` and, on a Redis error, deliberately falls back to the old single-replica behavior). `reconcile()` has no equivalent guard — and unlike release, it acts on *every* agent at once, so one bad tick is fleet-wide rather than per-session.

A cheap guard is to make the sweep prove Redis is actually answering before trusting its emptiness — e.g. skip the tick when the scan returns zero keys while this replica's own `open` map is non-empty, or have each replica write a heartbeat sentinel key and abort the tick if the sentinel is missing. Either way the failure mode becomes "a stale pin lingers one more minute" instead of "a live agent is hibernated", which is the right direction for a signal path the docstring already describes as best-effort.

packages/api-server/src/apps/api-server/session-presence.ts
```diff
@@ -1,23 +1,96 @@
...
+    } while (cursor !== "0");
+    return ids;
+  }
+
+  async function anyReplicaHolds(agentId: string): Promise<boolean> {
+    const ids = await scanAgentIds(`${KEY_PREFIX}${agentId}:*`);
+    return ids.has(agentId);
+  }
+
+  const heartbeat = setInterval(() => {
```
dam-code-guardian, line 75: 🟡 **Warning:** the heartbeat refreshes the Redis key but never re-asserts the annotation, so a single failed `patchAnnotation` leaves a held session unpinned for its whole lifetime.

`set()` at `:42` swallows every error (`.catch(() => {})`), and it is called exactly once per agent — at `acquire()` when the refcount goes 0→1. If that one K8s patch fails (409 conflict, transient API-server error, throttling), nothing retries it: the heartbeat only touches Redis, and `reconcile()` is one-directional — it clears annotations for unheld agents, never sets them for held ones.

The result is an agent with a live session, a healthy presence key, and no `active-session` annotation — so `shouldRun` sees only `last-activity` and hibernates the pod once the idle timeout passes, mid-session. Before this PR the same dropped write self-healed on the next `acquire()`; now that presence is long-lived and heartbeated, the window is the session's full duration.

Since the heartbeat already iterates `open.keys()` every 30 s, the cheap fix is to re-assert the annotation there alongside the key refresh (`set(agentId, true)`), making it converge like the rest of the reconcile design. Alternatively, have `reconcile()` set as well as clear — it already has both sides of the data it would need.

packages/api-server/src/apps/api-server/app.ts
```diff
@@ -242,7 +257,11 @@ export function startApiServerApp(deps: ApiServerAppDeps) {
         : {}),
     },
   });
-  connectionsBoot.refreshLoop.start();
+  // OAuth refresh runs on the shared periodic-jobs queue: one refresh pass
+  // per minute across replicas — refresh-token rotation must never race.
+  void deps.periodicJobs.register("oauth-refresh", 60_000, () =>
```
dam-code-guardian: 🟡 **Warning:** `void`-ing these two `register()` calls turns a registration failure into an unhandled rejection and a job that silently never runs.

`register()` is `async` and awaits `queue.upsertJobScheduler()`, a Redis round-trip that can reject (connection refused during a rolling restart, auth failure, timeout). Every registration in `index.ts` is `await`ed (`:242`, `:570`, `:670`, `:698`, `:710`, `:787`, `:800`); these two — `oauth-refresh` here and `session-presence-reconcile` at `:1017` — are `void`ed instead, because `startApiServerApp` is sync.

Two consequences. First, the rejection is unhandled: with Node's default `--unhandled-rejections=throw` that crashes the process, and the failure surfaces as an opaque boot crash rather than a named job error. Second, if it somehow settles late, the worker is only created after the `await`, so a rejection means no worker and no scheduler for that job — and `start()` at `index.ts:713` has already run, so nothing will pick it up. OAuth refresh stopping silently is the expensive one: tokens simply stop being renewed until someone notices connections going `expired`.

At minimum attach a `.catch()` that logs the job name through the same `log` the worker's `failed` handler uses. Better, have `startApiServerApp` return the registration promises (or take an async init) so `index.ts` can `await` them with the others and fail fast at boot — the current split means two of nine periodic jobs have materially weaker startup guarantees than the rest.

packages/api-server/src/core/xact-lock.ts
```diff
@@ -0,0 +1,17 @@
...
+/** Cross-replica critical section keyed by an arbitrary string. Takes a
+ *  Postgres transaction-scoped advisory lock (`pg_advisory_xact_lock`) on a
+ *  hash of `key`, runs `fn`, and releases the lock with the transaction —
+ *  same-connection release is structural, and a crashed replica's lock dies
+ *  with its connection. Blocking: concurrent callers for the same key queue
+ *  in Postgres. Keep `fn` short — it holds a pool connection. */
+export function createXactLock(db: Db) {
+  return async <T>(key: string, fn: () => Promise<T>): Promise<T> =>
+    db.transaction(async (tx) => {
+      await tx.execute(sql`SELECT pg_advisory_xact_lock(hashtext(${key}))`);
```
dam-code-guardian: 🟡 **Warning:** `hashtext()` maps the key onto a 32-bit int, so unrelated owners can collide and serialize against each other — and the bare `hashtext` is resolved through the session `search_path`.

Collision impact is mild but real: `pg_advisory_xact_lock` keys on the hash, not the string, so two owners whose `resize:<owner>` hashes coincide block each other. With `fn` short (a `repo.get`, a ceiling check, a patch) that is a brief queue rather than a correctness bug — worth a line in the docstring, since the current text promises a critical section "keyed by an arbitrary string" and callers may later reuse this helper for something where collisions matter more.

The sharper point is scoping: `hashtext` is unqualified, so it resolves via `search_path`. A future migration or role default that puts a user-writable schema ahead of `pg_catalog` changes which function runs — and this one decides a lock identity. `pg_catalog.hashtext(...)` costs nothing and removes the question.

Worth noting the neighbouring `withAdvisoryLock` in `modules/usage/infrastructure/advisory-lock.ts` takes a caller-supplied `number` key with a hand-picked constant, so the two advisory-lock helpers now share one 32-bit keyspace with no coordination between them; a collision there would be a genuine cross-subsystem stall. A short comment naming the shared keyspace (or a documented namespace split) would keep that from being discovered the hard way.

packages/api-server/src/modules/connections/infrastructure/oauth-engine.ts
```diff
@@ -146,20 +142,19 @@ export function createOAuthEngine(
...
       }
-      ensureJanitor();
       const codeVerifier = crypto.randomBytes(32).toString("base64url");
       const codeChallenge = crypto
         .createHash("sha256")
         .update(codeVerifier)
         .digest("base64url");
       const state = crypto.randomBytes(16).toString("hex");
-      pendingFlows.set(state, {
+      await pendingFlows.set(state, {
```
dam-code-guardian, line 216: 🟢 **Suggestion:** this `set` serializes the whole `PendingFlow` — including `provider.clientSecret` — into Redis, moving a credential that previously never left process memory.

`buildProvider` (`services/oauth-flow.ts:226`) puts the resolved `clientSecret` on the provider, and `PendingFlow` embeds the provider whole, so the JSON written here contains the connection's OAuth client secret in plaintext for the 10-minute TTL. It is needed later — `exchange()` posts it to the token endpoint — so carrying it is not gratuitous. But the resting place changes: previously an in-process `Map`, now a Redis instance whose AOF file persists to a PVC, and whose keys are readable by anything holding the shared password (the same instance backs BullMQ and the pub/sub bus).

The blast radius is bounded (short TTL, cluster-internal, `requirepass`), which is why this is a suggestion and not a warning. Two options that keep the cross-replica property without widening secret exposure: persist only the connection id plus flow metadata and re-run `buildProvider` from `secretStore` on the callback leg (the callback already has DB access), or keep the secret out of `PendingFlow` and pass it separately into `exchange()`. The first also removes a staleness trap — a rotated client secret currently sits stale in the pending record for up to the TTL.

JanPokorny:
All review findings addressed as of `c3e699b2` (rebased on main):

- **reconcile clears live pins on Redis blip** → `reconcile()` now refuses an empty scan while this replica holds sessions, and re-checks `anyReplicaHolds` per agent before each clear (bounds partial-SCAN risk to two consecutive misreads).
- **annotation never re-asserted** → the 30s heartbeat re-patches `active-session: true` for every held agent alongside the Redis key refresh; a dropped acquire-time patch now self-heals within one heartbeat.
- **fire-and-forget `register()`** → both app-side registrations `.catch` with a named error log and `process.exit(1)` — a silently-missing periodic job becomes a loud boot failure.
- **`hashtext` collisions / search_path** → `pg_catalog.hashtext(...)`; docstring documents the 32-bit keyspace shared with `modules/usage`'s advisory-lock constants and warns against uses where a spurious cross-key wait matters.
- **clientSecret in Redis** → stripped before the store write; `completeOAuth` re-resolves the provider via `buildProvider` + secretStore before `exchange()`, which also picks up a mid-flow secret rotation.
- **missing unit tests** → `session-presence.test.ts` covers pin/unpin refcounting, cross-replica holds, heartbeat re-assertion, both reconcile guard rails, and memory TTL-store semantics (CodeQL findings on the fake-redis helper fixed in `c3e699b2`).

Also per the TSEng review: all six optional deps are now required (`pendingStore`, the four flow `store`s, `resizeLock`); `inProcessResizeLock` and the memory-store fallbacks are deleted. Doc-drift: both stale boot-time pin-clear claims fixed, `budgets.md` names the advisory lock, `Last verified` bumped on all touched pages.

PR #3139, opened 2026-08-03
feat(skills): publish badge reports the pull request's real state

### LINKED ISSUE #3019: "In review" badge stays on a published skill after its pull request is closed

PetrBulanek:
A published skill's badge reports only that a publish happened, because nothing ever re-reads the pull request. This resolves the pull request's actual state from GitHub and says which of five things is true: it's a draft, it's awaiting review, it landed in the catalog, it was closed without landing, or we genuinely don't know.

Two things that only become possible once the state is known come with it:

- A pull request closed without merging is a dead end today — the badge permanently replaces the Publish button, so a user told "rename it and resubmit" can't. The button returns as **Publish again**, in that state only.
- A merged skill currently appears twice on the page, the second copy claiming it isn't installed while the file is on disk and loaded by the harness. That contradiction is resolved, and handing the skill over to its source becomes an explicit, explained action.

State is read anonymously for public sources and through the running sandbox for private ones, which keeps the existing rule that only the agent-runtime talks to GitHub with credentials. Resolution never wakes a hibernated sandbox. Never re-reading a settled pull request, plus re-checking any one record at most hourly, keep it inside GitHub's anonymous rate budget; when it can't be resolved, the badge claims less rather than claiming something false.

Closes #3019

## Sub-issues

- [x] 01 — Record the resolved pull-request state
- [x] 02 — Resolve public pull-request state
- [x] 03 — Render the five states
- [x] 04 — Resolve private state through a warm pod
- [x] 05 — Expose the local content hash and de-duplicate the merged row
- [x] 06 — Hand a merged skill over to its source

## Note for review

The plan assumed an anonymous `304 Not Modified` is exempt from GitHub's rate limit, so ETags would make unchanged pull requests free. Measured against `api.github.com`, that is false — four consecutive conditional reads each decremented `x-ratelimit-remaining` by one, exactly like a `200`. The exemption applies to the *authenticated* primary limit, which is the bucket the private-source path uses. The ETag is still stored and sent (bandwidth), but the per-record hourly re-check is what actually bounds cost: one request per unresolved pull request per hour fits ~50 of them inside the 60/hour ceiling, where re-checking every 10-minute tick would let ten exhaust it.

## Post-review fix: the de-dupe window

Found while smoke-testing this branch. De-duplication was gated on `prState === "merged"`, so between a pull request actually merging and the resolver noticing, the skill still showed twice — badged `Submitted` or `In review` next to its own source entry. Because a resolved record then sits on the hourly re-check, that window could last **over an hour** on the normal path (publish → resolves `open` → user merges).

The gate was the wrong question. It asked "do we believe GitHub merged it?" when the page can ask two observable things instead: did we publish this skill to *this* source (the publish record), and does the source now serve byte-identical content (the scan). Suppression now keys on those, and `getState` requests the local hash for every published name rather than only `merged` ones — without that second half the relaxed predicate would have had no hash to compare during the window.

Every false-positive guard still holds: an unrelated same-named catalog skill differs in hash; a coincidental content match has no publish record for that source; a locally edited copy differs in hash, so both rows correctly show. `Track from {source}` deliberately keeps the stricter `merged` gate, because it *writes* — install overwrites the local copy.


packages/db/drizzle/0019_early_moon_knight.sql
```diff
@@ -0,0 +1,11 @@
+-- #3019: a publish record only ever attested that a publish happened, so the
+-- badge claimed "Published" forever — even after the pull request was merged or
+-- closed. These columns hold the pull request's *resolved* state, re-read from
+-- GitHub by a periodic job. All nullable with no default: null means "never
+-- resolved", the honest state for every row that exists today. `pr_etag` makes
+-- the re-read conditional (a 304 costs nothing against the anonymous rate
```
dam-code-guardian: 🟡 **Warning:** This says a 304 "costs nothing against the anonymous rate limit" — the precise belief the PR description reports as measured false, and which every other comment in the change is careful to correct (`pr-state-reader.ts`: "do **not** mistake it for a budget mechanism"; the resolver's header: "it must never be treated as a budget exemption here").

Migrations are append-only, so this is the one comment that can never be revised — it will outlive the code it describes and is exactly where a future reader goes to learn why `pr_etag` exists. Leaving the disproven rationale here invites someone to "optimize" by dropping the hourly backoff and relying on ETags, which is the failure the rest of the PR was shaped to prevent.

```suggestion
-- the re-read conditional (saving bandwidth only — an anonymous 304 is charged
-- against the rate limit exactly like a 200, measured against api.github.com),
```

packages/api-server/src/modules/skills/services/resolve-pr-state.ts
```diff
@@ -0,0 +1,158 @@
...
+  /** One pass; returns how many records had a state written. */
+  tick(): Promise<number>;
+}
+
+export function createPrStateResolver(deps: {
+  agentSkills: AgentSkillsRepository;
+  reader: PrStateReader;
+  /** Authenticated fallback for private sources, through the agent's own pod.
+   *  Absent means no escalation — the anonymous result stands. */
+  podReader?: PodPrStateReader;
```
dam-code-guardian: 🟡 **Warning:** `podReader` is optional, but `composePrStateResolver` (`compose.ts:89`) is the only construction site and always passes one — so the `&& deps.podReader` guard at line 122 is unreachable in production, and a future wiring omission would degrade silently to "private sources never resolve" instead of failing to compile.

The adapter already models unavailability without needing the port to be absent: `createPodPrStateReader` returns `null` for a stopped agent and for a failed call, and the caller handles null correctly. So the disabled behavior the optionality was reaching for is already expressible by a *required* port — no new concept needed.

```suggestion
  podReader: PodPrStateReader;
```

Then drop `&& deps.podReader` at line 122.

packages/api-server/src/modules/skills/services/resolve-pr-state.ts
```diff
@@ -0,0 +1,158 @@
...
+ *  anonymous rate-limit window, so cost is one request per record per window
+ *  and there is nothing to tune. A record never attempted skips the wait, so a
+ *  fresh publish still resolves on the next tick rather than in an hour. */
+const RECHECK_INTERVAL_MS = 60 * 60_000;
+
+/** Backstop against a pathological backlog; the hourly re-check above is what
+ *  bounds spend in normal operation. Ten rather than a rounder number because
+ *  ten per tick is exactly the anonymous hourly ceiling at this interval, so
+ *  even a backlog large enough to saturate every tick cannot overspend. */
+const MAX_READS_PER_TICK = 10;
```
dam-code-guardian, line 56: 🟢 **Suggestion:** The comment's reasoning is right but its conclusion is stronger than the arithmetic supports: 10 reads × 6 ticks/hour = **exactly** 60, which is the whole anonymous ceiling rather than a margin under it. So a saturated backlog doesn't "cannot overspend" — it spends the entire budget with nothing left over, and the pass depends on the `rate-limited` early return to stay correct at the boundary.

That matters because the ceiling is per-IP and shared by every user of this api-server instance, so an unrelated anonymous GitHub call from the same pod tips a saturated tick into 429s. The mechanism is fine (the early return handles it, and terminal-state exclusion drains the backlog); it's the "cannot" that overstates. Worth either lowering the cap to leave headroom, or restating the rationale as "exactly saturates the ceiling; the rate-limit early return is what keeps that safe."

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `4281e176`

## PR #3139: feat(skills): publish badge reports the pull request's real state
**Author:** PetrBulanek | **Branch:** feat/3019-pr-state-badge → main | **Changes:** +3676 −139 (34 files)

### Summary
Resolves a published skill's actual pull-request state (draft/open/merged/closed/unknown) via a new periodic api-server job — anonymous `api.github.com` reads with a per-record hourly backoff, escalating to an authenticated read through the agent's own warm pod for private sources — and uses it to fix three downstream lies: a permanently-"Published" badge, a dead-end closed PR with no republish path, and a merged skill rendered twice.

### Findings
- 🟡 **Warning:** The migration's comment asserts a 304 is free against the anonymous rate limit — the exact claim this PR measured to be false (`packages/db/drizzle/0019_early_moon_knight.sql:6`)
- 🟡 **Warning:** `podReader` is an optional dependency whose only wiring site always supplies it (`packages/api-server/src/modules/skills/services/resolve-pr-state.ts:62`)
- 🟢 **Suggestion:** `MAX_READS_PER_TICK`'s "cannot overspend" rationale is arithmetically exact, not conservative (`packages/api-server/src/modules/skills/services/resolve-pr-state.ts:50`)
- ✅ **Looks good:** The terminal-state design is what makes the never-wake guarantee defensible, and it holds end to end: `listPrStateCandidates` excludes `merged`/`closed` in SQL (`agent-skills-repository.ts:207-210`), `setPrState` is keyed on `prUrl` so one read settles every agent's row, and the unavailable path calls `touchPrState` — which never writes `prState` — so a blip provably cannot downgrade a resolved `merged` to null.
- ✅ **Looks good:** The `ORDER BY … asc nulls first` is load-bearing rather than incidental: Postgres sorts NULL last by default, which would starve never-attempted records — a fresh publish, the one case a user is actively watching — behind every already-attempted one whenever the backlog exceeds the per-tick cap.
- ✅ **Looks good:** Merged-row suppression fails safe in the direction that matters. An absent `contentHash` never suppresses (`skills-surface.tsx`), so a pod predating the field or a failed hash shows both rows rather than hiding the upstream version, and suppression is applied after the `N of M` count and the collapse decision so those still describe what the source actually holds.

aligned

**Optional dependency on a service factory** — `infrastructure.md` → *Required Dependencies — Never Optional*. `resolve-pr-state.ts:62` declares `podReader?: PodPrStateReader`, forcing the consumer to branch on presence (line 122) while the sole wiring site always supplies it. The rule: an optional dep "models a capability toggle as a wiring question, which is the wrong axis… hides a missing wiring as a silent skip instead of a type error." See the inline comment for why the existing null contract already is the disabled behavior.

Verified compliant: the `services → domain ← infrastructure` direction holds (the `PodPrStateReader` port is declared in `domain/pr-state.ts` precisely so infrastructure never imports from services); `pr-state.ts`/`pr-url.ts` are pure; validation lives in the contract package (`skillReadPullRequestInputSchema`) with the router delegating immediately; the repository speaks domain language and hides all Drizzle specifics; `composePrStateResolver` is the only place concrete adapters are named, and splitting it from `composeSkillsModule` on the owner-agnostic axis is the right cut.

✅ No findings. This PR *fixes* a pre-existing CRITICAL list-item violation rather than adding one: the ~85-line inline `.map(...)` in `standalone-skills-group.tsx` (nested JSX, a conditional pill/button branch, a kebab) becomes `StandaloneSkillRow` — exactly the "touch-it = migrate-it" move. Derived state is `useMemo` with correct deps and no new `useEffect`; `prState` stays server-owned in the query cache, with the optimistic record writing `prState: null` into that same cache rather than opening a second store; `PR_STATE_PILL` is a keyed lookup whose `?? "unknown"` fallback means an unresolved record cannot render a false "Published".

### Verdict
COMMENT — The core design is sound and its central safety claims verify against the code rather than just the prose; the findings are two comment/wiring issues and one overstated bound, none of which threaten correctness.

---

kapetr (commented):
Review notes from a read of the branch. Language: ASD-STE100.

**Tests — `resolve-pr-state.ts` (major)**
Add unit tests for the new logic. Include `parsePrUrl`, `derivePrState`, the throttle predicate, and `publishedDuplicatesBySource`. This code controls a shared GitHub rate budget. A silent error is expensive. The functions are pure, so tests are easy.

**Candidate order — `agent-skills-repository.ts:229`**
The order has no tiebreak for a running agent. If two agents publish the same private PR, the code can select the hibernated agent again and again. Then the PR does not resolve. Prefer a running agent, or try the next pod on a null result.

**Private records that do not drain — `resolve-pr-state.ts`**
A private PR with a hibernated agent stays a candidate forever. It uses one anonymous request each hour, with no end. Add an attempt-count backoff, or a note.

**Encode the URL parts — `pr-state-reader.ts:36`, `github-rest-client.ts:177`**
Use `encodeURIComponent` on `owner` and `repo`. The host is locked, so this is defense-in-depth. A bad character must not change the request target.

**Hash cost on each poll — `skills-service.ts:647`**
`getState` reads and hashes each published skill directory every 5 seconds. This is costly on the NFS PVC. Cache the hash by directory mtime, or hash only names with a candidate.

**Stale comment — `api-server-api/schemas.ts:108`**
The comment says `instance_skills` and "Published badge". The table is `agent_skills`. The badge now shows five states. Correct the comment.

packages/api-server/src/modules/skills/services/resolve-pr-state.ts
```diff
@@ -0,0 +1,164 @@
...
+  return {
+    async tick() {
+      const now = new Date();
+      // One extra so a full page is evidence of a backlog rather than a
+      // coincidence, and the log can say so honestly.
+      const candidates = await deps.agentSkills.listPrStateCandidates(
+        new Date(now.getTime() - RECHECK_INTERVAL_MS),
+        MAX_READS_PER_TICK + 1,
+      );
+      const skipped = Math.max(0, candidates.length - MAX_READS_PER_TICK);
```
kapetr, line 83: Add unit tests for the new logic. Include `parsePrUrl`, `derivePrState`, this throttle predicate, and `publishedDuplicatesBySource`. This code controls a shared GitHub rate budget. A silent error is expensive. The functions are pure, so tests are easy.

packages/api-server/src/modules/skills/infrastructure/agent-skills-repository.ts
```diff
@@ -141,6 +191,70 @@ export function createAgentSkillsRepository(db: Db): AgentSkillsRepository {
...
+              isNull(agentSkillPublishes.prStateCheckedAt),
+              lt(agentSkillPublishes.prStateCheckedAt, staleBefore),
+            ),
+          ),
+        )
+        // NULLS FIRST, explicitly: Postgres sorts NULL last by default, which
+        // would put never-attempted records — a fresh publish, the case a user
+        // is actually watching — behind every already-attempted one and starve
+        // them whenever the backlog exceeds the per-tick cap.
+        .orderBy(sql`${agentSkillPublishes.prStateCheckedAt} asc nulls first`)
```
kapetr, line 235: The order has no tiebreak for a running agent. If two agents publish the same private PR, the code can select the hibernated agent again and again. Then the PR does not resolve. Prefer a running agent, or try the next pod on a null result.

packages/api-server/src/modules/skills/services/resolve-pr-state.ts
```diff
@@ -0,0 +1,164 @@
...
+            `anonymous rate limit exhausted; ${resolved} record(s) resolved before stopping`,
+          );
+          return resolved;
+        }
+        // A 404 means "not resolvable anonymously" — private as much as gone.
+        // Escalate to the owning agent's pod, whose gateway holds the token.
+        // Only `not-found`: an error or a rate limit says nothing about the
+        // source being private.
+        if (result.reason === "not-found") {
+          const disposition = await deps.podReader.read(
```
kapetr: A private PR with a hibernated agent stays a candidate forever. It uses one anonymous request each hour, with no end. Add an attempt-count backoff, or a note.

packages/api-server/src/modules/skills/infrastructure/pr-state-reader.ts
```diff
@@ -0,0 +1,83 @@
...
+ * for a budget mechanism: an anonymous 304 decrements
+ * `x-ratelimit-remaining` by one exactly like a 200 (measured — GitHub's
+ * "conditional requests are free" rule covers the authenticated primary limit,
+ * not this bucket). It saves bandwidth only. The caller's per-record hourly
+ * re-check is what keeps the pass inside the ceiling.
+ */
+export function createGitHubPrStateReader(): PrStateReader {
+  return {
+    async read(coords, etag) {
+      const url = `https://api.github.com/repos/${coords.owner}/${coords.repo}/pulls/${coords.number}`;
```
kapetr: Use `encodeURIComponent` on `owner` and `repo` here and in `github-rest-client.ts:177`. The host is locked, so this is defense-in-depth. A bad character must not change the request target.

kapetr (commented):
Inline comments.

packages/api-server/src/modules/skills/services/skills-service.ts
```diff
@@ -636,18 +636,29 @@ export function createSkillsService(deps: SkillsServiceDeps): SkillsService {
...
+      // the duplicate on screen for as long as the resolver takes to notice a
+      // merge — up to the re-check interval. A sandbox that never published
+      // still asks for none.
+      const instancePublishes =
+        await deps.agentSkillsRepo.listPublishes(agentId);
+      const publishedNames = [
+        ...new Set(instancePublishes.map((p) => p.skillName)),
+      ];
+
+      const local = await deps.runtimeClient.listLocal(agentId, publishedNames);
```
kapetr, line 653: `getState` reads and hashes each published skill directory every 5 seconds. This is costly on the NFS PVC. Cache the hash by directory mtime, or hash only names with a candidate.

packages/api-server-api/src/modules/skills/schemas.ts
```diff
@@ -87,6 +92,16 @@ export const skillPublishRecordSchema = z.object({
...
+   *  never re-read (#3019). */
+  prState: z.enum(["draft", "open", "merged", "closed"]).nullable(),
+  /** ISO 8601 timestamp of the last resolution *attempt*, which is not
+   *  the same as the last success: an attempt that resolved nothing
+   *  still stamps it, because it doubles as the backoff clock. Null
+   *  until the first attempt. */
+  prStateCheckedAt: z.string().nullable(),
 });
 
 /** Reconciled view of an instance's skills: both the installed
```
kapetr, line 107: This comment block says `instance_skills` and "Published badge". The table is `agent_skills`. The badge now shows five states. Correct the comment.

kapetr (approved):

packages/api-server/src/modules/skills/infrastructure/agent-skills-repository.ts
```diff
@@ -141,6 +193,80 @@ export function createAgentSkillsRepository(db: Db): AgentSkillsRepository {
...
+            // instance's whole hourly allowance. The wait is per-record: an
+            // hour after an attempt that learned something, doubling per
+            // consecutive failure up to a day, so a record that can never
+            // resolve (private source, hibernated agent) decays to one request
+            // a day instead of holding its hourly slot forever. A record never
+            // attempted is exempt, so a fresh publish still resolves on the
+            // next tick.
+            or(
+              isNull(agentSkillPublishes.prStateCheckedAt),
+              sql`${agentSkillPublishes.prStateCheckedAt} + least(power(2, ${agentSkillPublishes.prStateCheckFailures}), 24) * interval '1 hour' <= ${now}`,
```
kapetr, line 227: **The backoff stops private records from resolving.**

The pod path resolves a private PR by a sample. It reads the PR only if the pod is warm at that moment.

A hibernated pod now records `failed`. This is the same result as a deleted repository. But a cold pod tells you nothing about the PR.

The effect is opposite to the intent. Each missed sample makes the next sample later. An agent that runs a few hours each day can go to the daily lane. Then it can never resolve. Before this commit, it had some chances each day.

Suggestion: add a third result, `deferred`. It writes `prStateCheckedAt`, but it does not increase `prStateCheckFailures`. Use it when no publisher has a running pod. Better: test `computeAgentState` before the GitHub read. Then a cold private record costs no budget, and it can stay in the hourly lane.

kapetr, line 227: **`power(2, …)` overflows and stops the resolver.**

The `least(…, 24)` cap comes after `power`. `power(2, int)` gives `double precision`.

Measured on Postgres 16:
- `least(power(2, 1023::int), 24)` → `24`
- `least(power(2, 1024::int), 24)` → `ERROR: value out of range: overflow`

`pr_state_check_failures` has no limit. At 1024 failures the query fails for **every** record. The resolver then stops for all agents. `periodic-jobs.ts:60` writes one log line, so the failure is silent.

Fix: limit the exponent, not the result.

```sql
least(power(2, least(pr_state_check_failures, 5)), 24) * interval '1 hour'
```

kapetr, line 227: **The backoff has no end, and live PRs lose their turn.**

The backoff gives a daily floor, but it removes nothing. The cost stays at one read each day for each dead record, with no limit. 10 reads × 144 ticks = 1440 reads each day, so approximately 1440 dead records fill the resolver for all time.

The order makes this worse for live records. `ORDER BY pr_state_checked_at ASC NULLS FIRST` puts a dead record (last read 24 h ago) before a healthy open PR (last read 1 h ago). A new publish is safe, because of `NULLS FIRST`. A re-check of an open PR is not safe.

Suggestion: add `pr_state_check_failures < 30` to the `WHERE`. Then the set of records drains.

kapetr (commented):
Verified the fixes for the earlier review. 7 of 9 are correct. The tests pass and the checks are green.

The backoff commit (`9f295cebf`) adds two new problems. Both are in the same SQL line. See the inline comments.

packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts
kapetr, line 1: **The new hash cache is not used at line 283.**

The cache at line 305 is correct. But line 283 reads and hashes the directory again, without the cache:

```ts
const localHash = pristineHash === null ? null : await hashSkillDirIfPresent(localDir);
```

`doListLocal` always gives `pristineSkillPaths`, so each image-baked skill still gets a full read from the NFS PVC every 5 seconds. A published image-baked skill is hashed two times each poll.

This path is on `main` already, so it is not a regression. But it keeps the cost that this fix tried to remove. `hashSkillDirCached(localDir, contentHashCache)` closes it.

PR #3142, opened 2026-08-03
fix(ui): drop the stale "Sandboxes" back label from the agent overlay

jenna-winkler:
The Sandboxes list was renamed to Home, so the back button on the agent-unavailable overlay (hibernating, starting, errored) pointed at a page name that no longer exists.

On desktop the button is removed outright: Home is already reachable from the rail's Home item and the DAM logo.

On mobile it stays, minus the stale word, so `← Sandboxes` becomes `←`. Chat hides the rail's bottom bar there, so the arrow is the only in-view way back. Whether it's redundant is a separate question from this one.

One file, `agent-unavailable-overlay.tsx`, +7/-4. Nothing added.

Typecheck, lint and 221 UI unit tests pass.

### DIFF
--- a/packages/ui/src/modules/agents/components/agent-unavailable-overlay.tsx
+++ b/packages/ui/src/modules/agents/components/agent-unavailable-overlay.tsx
@@ -47,8 +47,14 @@ const OVERLAY_COPY: Record<AgentDisplayState, OverlayCopy> = {
   },
 };
 
-/** Shared chrome for every overlay variant: full-view takeover, back button,
- *  centered content column. */
+/** Shared chrome for every overlay variant: full-view takeover, centered
+ *  content column. Desktop needs no back affordance — the icon rail stays
+ *  visible beside the takeover, so Home is always one click away. Mobile has
+ *  neither rail nor bottom bar in chat, and the header's own back button sits
+ *  beneath this overlay, so it keeps an icon-only escape hatch. That icon is
+ *  the only way out at the one width where the target is exclusively touch,
+ *  hence `icon` sizing rather than the label-era `inline`, which would collapse
+ *  the box onto the 14 px glyph. */
 function OverlayFrame({
   onBack,
   children,
@@ -60,12 +66,12 @@ function OverlayFrame({
     <div className="absolute inset-0 z-overlay flex flex-col bg-background/95 backdrop-blur-sm">
       <Button
         variant="ghost"
-        size="inline"
+        size="icon"
+        aria-label="Back"
         onClick={onBack}
-        className="absolute left-4 top-3 gap-1 text-sm font-medium text-muted-foreground hover:bg-transparent"
+        className="absolute left-1 top-0 text-muted-foreground hover:bg-transparent md:hidden"
       >
         <ArrowLeft size={14} />
-        Sandboxes
       </Button>
       <div className="flex flex-1 flex-col items-center justify-center gap-5 px-6 text-center">
         {children}
--- a/packages/ui/src/modules/sessions/views/chat-view.tsx
+++ b/packages/ui/src/modules/sessions/views/chat-view.tsx
@@ -467,7 +467,7 @@ export function ChatView() {
         <Button
           variant="ghost"
           size="inline"
-          aria-label="Back to sandboxes"
+          aria-label="Back"
           onClick={handleBack}
           className="md:hidden gap-1 text-sm font-medium text-muted-foreground hover:bg-transparent"
         >


packages/ui/src/modules/agents/components/agent-unavailable-overlay.tsx
```diff
@@ -61,11 +64,11 @@ function OverlayFrame({
       <Button
         variant="ghost"
         size="inline"
+        aria-label="Back"
         onClick={onBack}
-        className="absolute left-4 top-3 gap-1 text-sm font-medium text-muted-foreground hover:bg-transparent"
+        className="md:hidden absolute left-4 top-3 text-muted-foreground hover:bg-transparent"
```
dam-code-guardian: 🟡 **Warning:** with the label gone this collapses to a 14 × 14 px tap target — `size="inline"` resolves to `h-auto p-0` and the className adds no padding, so the box is exactly the `ArrowLeft size={14}` glyph. It matters more here than it looks: `md:hidden` makes this touch-only, and it is the sole escape hatch behind the takeover (the icon rail's bottom bar is suppressed by `hideMobileBar` in chat, and the header's own back button sits under `z-overlay`). Well under the ~44 px touch minimum. Padding the ghost button keeps the icon-only look and restores the hit area:

```suggestion
        className="md:hidden absolute left-4 top-2 -m-1 p-2 text-muted-foreground hover:bg-transparent"
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `9dd5933`

## PR #3142: fix(ui): drop the stale "Sandboxes" back label from the agent overlay
**Author:** jenna-winkler | **Branch:** fix/remove-stale-sandboxes-back-label → main | **Changes:** +7 −4 (1 file)

### Summary
Removes the stale "← Sandboxes" text from the agent-unavailable takeover overlay's back button and hides the button entirely on desktop, keeping an icon-only version on mobile where no other Home affordance is reachable behind the overlay.

### Findings
- 🟡 **Warning:** mobile back button's tap target shrinks to the 14 px icon (`packages/ui/src/modules/agents/components/agent-unavailable-overlay.tsx:69`)
- 🟢 **Suggestion:** the sibling mobile back button in the chat header still carries the stale page name in its accessible name — `aria-label="Back to sandboxes"` (`packages/ui/src/modules/sessions/views/chat-view.tsx:469`). It is the same control, same `handleBack`, and the same renamed destination this PR exists to stop naming; the visible half was fixed here while the assistive half still says "sandboxes". Out of this diff, so no inline comment.
- ✅ **Looks good:** the desktop removal is safe on both counts I could get wrong. Geometry: `IconRail`'s desktop `<nav>` (`hidden md:flex`, 56 px) is a flex sibling of the chat pane, while the overlay is `absolute inset-0` inside `chat-view`'s own `relative` root — so the takeover never covers the rail and Home genuinely stays one click away. Specificity: the cva base's `inline-flex` and the incoming `md:hidden` land in different variant scopes, so `tailwind-merge` keeps both rather than dropping one, and the button really does disappear at ≥ 768 px.

✅ No findings. The diff touches one UI component and changes no subsystem behavior or responsibility; no page under `docs/architecture/` describes the overlay's back affordance, no architecture page is edited, moved, or deleted, no new long-lived component is introduced, and no ADR changes state in this PR.

- 🟡 **Warning** (`agent-unavailable-overlay.tsx:69`) — `size="inline"` resolves to `h-auto p-0` in `buttonVariants`, and the className adds no padding, so with the text gone the button's box collapses to the `ArrowLeft size={14}` glyph: a 14 × 14 px hit area. Before this PR the label made the same control roughly 90 px wide. This is now the *only* escape hatch on mobile — the icon rail's bottom bar is suppressed by `hideMobileBar` in chat, and the header's own back button sits under the `z-overlay` takeover — and it is the one width where the target is exclusively touch. Well under the ~44 px touch minimum. The sibling header button (`chat-view.tsx:467`) has the same shape, so this matches existing code rather than inventing a problem, but the regression is introduced here.
- Everything else in scope is clean: Tailwind-only styling with no static `style={{}}`, no template-literal class building, props typed inline without `React.FC`, no state-lineage duplication, and the component stays a small presentational leaf well inside the size and JSX-weight targets. Class order puts `md:hidden` before the layout utilities where the informal convention is responsive-last — cosmetic only, not reported as a finding.

### Verdict
COMMENT — the stale label is correctly removed and the desktop hiding is sound; the one real consequence is the 14 px touch target left on mobile, worth a padded hit area before merge.

---

jenna-winkler:
Both findings addressed in 91e95de.

**Touch target** (warning) — confirmed and fixed. `size="inline"` is `h-auto p-0` in `buttonVariants`, so removing the label did collapse the box onto the 14 px glyph; the label had been carrying the hit area. Switched to `size="icon"` (`h-10 w-10`) and pulled the offsets from `left-4 top-3` to `left-1 top-0` so the glyph stays roughly where it was while the box grows around it. Desktop is unaffected — `md:hidden` still hides it there.

**Stale accessible name** (suggestion) — fixed too, `aria-label="Back to sandboxes"` → `"Back"` on the header's sibling button. Agreed it's the same control pointing at the same renamed page, and leaving the assistive half naming "sandboxes" would have been a half-fix. It widens the diff to a second file by one line, which seemed the right trade.

Also took the class-order note, `md:hidden` now sits last.

Re-ran: `tsc --noEmit`, `eslint`, `prettier --check` clean; 221 UI unit tests pass. Still not rendered in a browser — this pod can't complete `mise run ui:setup` — so the new offsets are reasoned from the variant definitions, not eyeballed. Worth a glance on a narrow viewport before merge.

tomkis (approved):

PR #3145, opened 2026-08-03
fix(agent-runtime): only recycle the harness when the env variable set changes

xjacka:
## Summary

Addresses #3143 with two complementary changes:

1. **A value-only env change never force-kills an in-flight turn.** It recycles the harness only at an idle turn boundary (immediately when nothing is running); the 60s force-recycle is reserved for changes to the variable **set** (add/remove/rename). Deferrals are logged, and a reset session drains a pending refresh instead of stranding it.
2. **The runtime state snapshot no longer depends on DB row order**, removing the only row-order-sensitive path found in review: duplicate env names across granted connections (`KUBECONFIG` being the built-in case), where a rotation's row UPDATE could flip the first-occurrence winner and read as an env change.

Whether the affected agents actually carried such a duplicate-name grant still needs verification against their state, so this PR **references** #3143 rather than closing it — change 1 removes the reported harm (killed runs) regardless of what triggered the deliveries; change 2 removes the one trigger the code review could identify.

## Background

A pod runs all its sessions through one shared harness subprocess. When the runtime-channel env driver saw *any* difference between the stored env snapshot and the delivered one, the runtime recycled the harness — waiting for a turn boundary but **force-killing after ~60s** if a turn was still in flight. Long scheduled/autonomous runs (no turn boundary for 30–100+ min) died mid-work; per the issue's evidence, 25% of scheduled review runs ended this way.

Tracing the mechanism: credential env rides static placeholders (the gateway injects real values on the wire), so a rotation itself never rewrites contribution values. The rotation-correlated diffs can only come from **unordered DB reads**: `readGrantedContributions` joins without an ORDER BY, the agent's env driver is first-occurrence-wins on name collisions, and the state hash gate lets any array difference through — so a rotation's row UPDATE could flip a duplicate-name winner and read as an env change. User env is PK-unique and ordered in code, so it cannot flip; two granted connections sharing an env name is the only shape that can.

## What changed

- **state builder** ([state-builder.ts](https://github.com/dam-agents/dam/blob/fix/env-rotation-recycle/packages/api-server/src/modules/runtime-delivery/services/state-builder.ts)): deterministic ORDER BY in each reader — granted connections oldest-first (`created_at`, `id`; stable across rotations and re-grants), user env by name, skills by source + name. Collision winners and state hashes no longer move with row order.
- **contribution hash** ([contribution-hash.ts](https://github.com/dam-agents/dam/blob/fix/env-rotation-recycle/packages/api-server/src/modules/runtime-delivery/domain/contribution-hash.ts)): order-independent for duplicate keys (content tie-break), so the hash is total and the ORDER BY is not load-bearing for hash stability.
- **env driver** ([env-plugin.ts](https://github.com/dam-agents/dam/blob/fix/env-rotation-recycle/packages/agent-runtime/src/modules/runtime-channel/drivers/env-plugin.ts)): classifies each change — `onChange` carries `{ namesChanged }`, own-property name checks. The new snapshot is always written to disk.
- **acp runtime** ([acp-runtime.ts](https://github.com/dam-agents/dam/blob/fix/env-rotation-recycle/packages/agent-runtime/src/modules/acp/services/acp-runtime.ts)): `refreshEnv({ force })` — required flag; `force: false` skips the force timer, so the recycle waits for the next idle turn boundary. Deferrals log the in-flight turn count. `tearDownSession` clears the session's turn slot and queue and re-checks the deferred recycle, so a reset session cannot strand a pending refresh. Queued prompts survive: the boundary check runs after the prompt queue is advanced (pinned by test).
- **composition root** ([server.ts](https://github.com/dam-agents/dam/blob/fix/env-rotation-recycle/packages/agent-runtime/src/server.ts)): passes `force: namesChanged`. The non-destructive consumers still refresh on every change: the pod service gets its SIGHUP, the git credential helper is reconfigured.
- **docs**: the env-rail note on the connections architecture page states the new semantics.

## Resulting behavior

| Change | Agent idle | Turn in flight |
|---|---|---|
| Variable added/removed/renamed (grant/revoke, env editor add/delete) | recycle now | recycle at boundary, forced after 60s (unchanged) |
| Value-only (user value edit, config-input edit, any rotation artifact) | recycle now | recycle at the turn boundary — never forced, deferral logged |

The cold-boot warm-start gate is unaffected: the first delivery to a fresh pod always adds names, so the gate still releases through the recycle path.

## Testing

- `mise run agent-runtime:check` + `api-server:check` — format, lint, tsc clean
- `mise run agent-runtime:test` — 224 tests pass; new units: env-change classification (incl. own-property names), `refreshEnv` seams (deferred refresh outlives the force bound and recycles at the boundary; forced refresh keeps the 60s bound; force timer survives a later deferred refresh in either order; a queued prompt defers the boundary recycle)
- `mise run api-server:test` — 756 tests pass (incl. duplicate-key hash regression)

Refs #3143 — closing pends verification that an affected agent carried a duplicate-name grant.


packages/agent-runtime/src/server.ts
```diff
@@ -165,8 +165,11 @@ const runtimeChannel = await composeRuntimeChannel({
   plugins: [
     createEnvPlugin({
       store: envStore,
-      onChange: () => {
-        acpRuntime.refreshEnv();
+      onChange: ({ namesChanged }) => {
+        // A value-only change (e.g. a rotated credential) applies at the next
+        // harness spawn; recycling for it would force-kill an in-flight turn.
+        // Only a changed variable set is worth interrupting the harness for.
+        if (namesChanged) acpRuntime.refreshEnv();
```
dam-code-guardian: 🟡 **Warning:** this gate treats a *second cluster connection being granted or revoked* as value-only, so a live harness is not recycled for it.

The env driver joins all `KUBECONFIG` contributions into one `:`-separated value (`env-plugin.ts:44-49`), so granting a second cluster changes only that variable's *value*, never the name set:

```
before: KUBECONFIG=/home/agent/.kube/connections/a.config
after:  KUBECONFIG=/home/agent/.kube/connections/a.config:/home/agent/.kube/connections/b.config
```

`sameNames` returns true → `namesChanged: false` → no `refreshEnv()`. I verified both directions against the real predicate: grant-second-cluster and revoke-second-cluster both classify as value-only. The `file` driver writes the new kubeconfig immediately, so the file is on disk while the running harness's `KUBECONFIG` still lists only the old path — the freshly granted cluster is invisible to `kubectl` until the next spawn, and a revoked one keeps being listed. That is a change in *reachable infrastructure*, not a rotated secret, so it belongs on the recycle side of the line this PR draws.

A name-set comparison cannot express this, because the composition lives inside the value. Options, in the driver's classification rather than here: treat a change to a known list-valued variable (`KUBECONFIG`) as a set change, or compare the `:`-split path sets for such variables. Flagging it so the tradeoff is a decision rather than an accident.

packages/agent-runtime/src/modules/runtime-channel/drivers/env-plugin.ts
```diff
@@ -79,3 +93,12 @@ function envEquals(
...
   return ak.every((k) => a[k] === b[k]);
 }
+
+function sameNames(
+  a: Record<string, string>,
+  b: Record<string, string>,
+): boolean {
+  const ak = Object.keys(a);
+  if (ak.length !== Object.keys(b).length) return false;
+  return ak.every((k) => k in b);
```
dam-code-guardian: 🟡 **Warning:** `k in b` walks the prototype chain, so an equal-count rename *away from* a name that collides with an `Object.prototype` member is misclassified as value-only.

Verified against this exact predicate:

```
sameNames({FOO:'1', X:'2'},      {BAR:'1', X:'2'}) → false  ✓ (rename detected)
sameNames({toString:'1', X:'2'}, {BAR:'1', X:'2'}) → true   ✗ (rename missed)
sameNames({constructor:'1',X:'2'},{BAR:'1',X:'2'}) → true   ✗
```

`'toString' in {BAR:..., X:...}` is `true` via the prototype, so the counts match and every key "exists" — `namesChanged` comes back `false` and the harness is never recycled even though a variable was genuinely removed and another added. Note `envEquals` just above is immune because it compares `a[k] === b[k]` on values, which is why this is new surface rather than a pre-existing hole.

Reachability is narrow (an env var named `toString`, `constructor`, `valueOf`, `hasOwnProperty`… and I found no name validation on the `agent_env` write path to rule it out), so this is a robustness fix, not a live bug. `Object.hasOwn` closes it and matches `envEquals`'s own-property semantics:

```suggestion
  return ak.every((k) => Object.hasOwn(b, k));
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `2b5098ae`

## PR #3145: fix(agent-runtime): only recycle the harness when the env variable set changes
**Author:** xjacka | **Branch:** fix/env-rotation-recycle → main | **Changes:** +101 −11 (5 files)

### Summary
Narrows the env driver's `onChange` so only a change to the *set* of variable names recycles the harness, while value-only changes are written to the snapshot and picked up at the next spawn. The mechanism is right and the placement (classify in the driver, decide at the composition root) is the correct seam.

### Findings
- 🟡 **Warning:** granting or revoking a *second* cluster connection is a genuine topology change that this classifier reports as value-only, so a live harness keeps a stale `KUBECONFIG` (`packages/agent-runtime/src/server.ts:172`)
- 🟡 **Warning:** `sameNames` tests membership with `k in b`, which traverses the prototype chain, so a rename away from an `Object.prototype` member name is misclassified as value-only (`packages/agent-runtime/src/modules/runtime-channel/drivers/env-plugin.ts:103`)
- ✅ **Looks good:** the cold-boot warm-start gate still releases. The first delivery to a fresh pod goes empty-set → populated, so `namesChanged` is `true` and `refreshEnv()` reaches `markEnvReady()` — the gate is not stranded by the new condition.

**Verdict:** minor drift

**Drift**

- ❌ **Check 1 — architecture-page drift.** `docs/architecture/agent-lifecycle.md:70` still describes the pre-PR model: user-typed and template env are "deliver[ed] … at the next idle turn with no pod roll". That is the same claim `connections.md:213` used to make and which this PR rewrote, so the two pages now contradict each other and the un-edited one is the stale one. After `server.ts:172`, only a variable-set change reaches the idle-turn recycle path; a value-only change lands at the next harness spawn. Proposed edit (depth stays on `connections.md`, which owns the rail):

  > The api-server stores user-typed and template env in Postgres `agent_env` and delivers all of it with no pod roll — a change to the *set* of variables lands at the next idle turn, a value-only change at the next harness spawn — ordering user env ahead of connection/secret env so it wins on a name collision.

  Bump that page's `Last verified:` when editing it. At 35,330 chars against the 40,000-char page cap there is ample headroom.

**Possible drift**

- ⚠️ **Check 2 — `Last verified:` is one day ahead.** `docs/architecture/connections.md:3` moved `2026-08-03` → `2026-08-04`, but today is 2026-08-03, so the page stamps a verification that has not happened yet. The rule ("bumped whenever you edit the page") is satisfied and no guideline forbids a future date — the only documented smell is a date *older* than the last refactor — so this is a judgement call: intended merge date, or a typo for `2026-08-03`?

Note, outside the drift rule but worth knowing before the next edit: `connections.md` is now **39,998 of 40,000 characters** — two characters of headroom against the cap enforced by `docs:check:doc-size` and the write-time hook. The next addition to that page trips the gate, which per the cap's own steering message is a signal to raise the level or split, not to reword to fit.

**No findings.** The diff is aligned with the architecture on every rule it touches:

- **Ports & adapters / dependency direction** — `env-plugin.ts` remains an infrastructure adapter: it gained a return type on a port it already owned, still imports only the `EnvStateStore` port type and the `agent-runtime-api` contract, and `sameNames` is a pure local helper. Nothing flows outward.
- **Composition root owns behavior toggles** — the `if (namesChanged)` decision sits in `server.ts`, the root, not inside the driver. The driver reports a classified fact; the root decides which consumers act on it. A `recycleOnValueChange?: boolean` knob on the driver *would* have been a finding.
- **Module public API** — `EnvChange` is exported through `modules/runtime-channel/index.ts` next to the existing `createEnvPlugin`, matching how `EnvStateStore` and `SkillInstallFn` are surfaced, and `server.ts` imports through that boundary rather than by deep path.
- **Optional-dependency rule** — `onChange?` was already optional at base and the diff does not widen it; pre-existing shape, out of scope for a diff review.

Not a finding (no rule backs it): the new test imports the driver by deep path, exactly as its sibling `env-plugin-kubeconfig.test.ts` does — the `index.ts` boundary governs inter-module imports, not a package's own tests.

### Verdict
COMMENT — the core change is sound and well-placed; two classifier gaps (`KUBECONFIG` composition, prototype-chain membership) and one stale architecture page are worth addressing before merge, none of them blocking.


---

xjacka:
Thanks — the review ran against `2b5098ae`; three later commits change the picture, so responding point by point:

**`k in b` prototype-chain membership** — valid, fixed in d2f56981: `sameNames` now uses `Object.hasOwn`, along with two same-class occurrences in the driver (the first-occurrence collision check and the `GH_TOKEN` availability flag, both of which would silently mistreat a variable named after an `Object.prototype` member). Regression test added.

**Second cluster grant classified as value-only (`KUBECONFIG`)** — accurate at the reviewed commit, but d9b64b6d changed what value-only means: it no longer waits for the next spawn; it recycles at the next idle turn boundary, and immediately when no turn is in flight. So granting a cluster to an idle agent applies instantly, and mid-run it applies the moment the turn ends. That behavior is deliberate: force-killing a long run so `kubectl` can see a just-granted cluster a minute sooner would reproduce exactly the harm #3143 describes. Left as is.

**agent-lifecycle.md:70 drift** — the contradiction existed only at the reviewed commit. With d9b64b6d, env changes once again land at an idle turn boundary, which is what agent-lifecycle.md says; connections.md carries the depth (the forced bound is reserved for variable-set changes). The two pages agree at HEAD, so no edit.

**`Last verified: 2026-08-04`** — the review ran on 2026-08-03; the page edit landed on the 04th, so the stamp is accurate now.

**Page cap headroom** — noted; the latest wording trimmed the cell slightly, but the page remains near the cap and the next substantive addition should split or raise the level rather than reword.

packages/api-server/src/modules/runtime-delivery/services/state-builder.ts
```diff
@@ -101,7 +105,8 @@ async function readGrantedContributions(
       connectionsTable,
       eq(connectionGrants.connectionId, connectionsTable.id),
     )
-    .where(eq(connectionGrants.agentId, agentId))) as {
+    .where(eq(connectionGrants.agentId, agentId))
+    .orderBy(asc(connectionsTable.createdAt), asc(connectionsTable.id))) as {
```
dam-code-guardian, line 109: 🟡 **Warning:** the ORDER BY here is load-bearing rather than defence-in-depth, because `contributionHash` is order-insensitive for *distinct* keys but not for *duplicate* ones.

`contributionHash` sorts on `keyFor(c)` alone (`contribution-hash.ts:7`). `Array.prototype.sort` is stable, so two contributions sharing a key — e.g. two granted clusters both contributing `KUBECONFIG` — keep their **input** order, and the input order is this query's row order. Verified against the real function:

```
[KUBECONFIG=/a, KUBECONFIG=/b] → 9c8ba00cc4e4…
[KUBECONFIG=/b, KUBECONFIG=/a] → 19f8495c4e82…   # same set, different hash
[FOO=1, BAR=2] vs [BAR=2, FOO=1] → identical     # distinct keys are immune
```

So for the exact multi-cluster case this PR's sibling change reasons about, the hash still moves with row order. The `ORDER BY created_at, id` you added does fix it — the rows stop moving — but the fragility stays one layer down, where the next unordered read of a duplicate-key contribution source silently reintroduces #3143's spurious deliveries with no test to catch it. `readGrantedContributions` is the only duplicate-key producer today, so this is hardening, not a live bug.

A tie-break makes the hash total and the ordering non-load-bearing:

```suggestion
  const sorted = [...contributions]
    .map(canonicalize)
    .map((c) => ({ ...c, j: JSON.stringify(c.value) }))
    .sort((a, b) => (a.k < b.k ? -1 : a.k > b.k ? 1 : a.j < b.j ? -1 : a.j > b.j ? 1 : 0));
```

I verified this collapses both orders above to one hash and leaves distinct-key and single-item hashes byte-identical to today's, so it is not a state-hash migration. (The suggestion targets `contribution-hash.ts:5-7`, not this line.)

Worth noting the ordering columns themselves are a good choice: `created_at, id` is stable across rotations and re-grants, and `id` breaks same-timestamp ties — `connection_grants`' PK is `(connection_id, agent_id)` with no unique constraint making `created_at` alone total.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `d2f56981`

## PR #3145: fix(agent-runtime): only recycle the harness when the env variable set changes
**Author:** xjacka | **Branch:** fix/env-rotation-recycle → main | **Changes:** +178 −21 (8 files)

### Summary
Re-review at `d2f56981` after four commits. Both prior warnings are resolved — one by a code fix, one by a design change that reframes the concern — and the new `state-builder` ordering half is sound. One residual gap in the ordering fix.

### Changes since last review
Previous HEAD: `2b5098ae` (2026-08-03T22:43:25Z) — verdict COMMENT

- ✅ **Fixed:** `sameNames` prototype-chain membership — `Object.hasOwn` now, and applied to all three sites in the driver, not just the flagged one (`env-plugin.ts:46,51,101`). Verified: `sameNames({toString:'1',X:'2'},{BAR:'1',X:'2'})` → `false`. Regression test added.
- ✅ **Fixed:** `KUBECONFIG` second-cluster grant — resolved by `d9b64b6d` changing what value-only *means*, not by reclassifying. It still classifies as value-only (I re-verified against the real predicate at this HEAD), but value-only now recycles at the next idle turn boundary, and immediately when idle — so a granted cluster applies at the turn end rather than the next spawn. The author's tradeoff is right: force-killing a long run to surface a cluster a minute sooner is exactly the harm #3143 describes.
- 🆕 **New:** `contributionHash` ignores row order for distinct keys but not for *duplicate* keys, so the ORDER BY is load-bearing rather than belt-and-braces (`state-builder.ts:109`).

### Findings
- 🟡 **Warning:** the state-builder's deterministic-order fix leaves the order-sensitivity it works around in place, one layer down (`packages/api-server/src/modules/runtime-delivery/domain/contribution-hash.ts:7`)

**Verdict:** aligned

Both items from the previous review are resolved at this HEAD, and no other check fires.

Check 1 (architecture-page drift) no longer holds: `d9b64b6d` restored idle-turn-boundary delivery for all env changes, so `agent-lifecycle.md:70` ("delivers all of it at the next idle turn with no pod roll") is accurate again, with `connections.md:213` carrying the depth. The pages agree — the author's reasoning for leaving both alone is correct. Check 2 is resolved: `Last verified: 2026-08-04` is today's date on a page edited in this diff.

Nothing else flags: no ADR links or `Motivated by:` sections added (3); the `connections.md:213` edit *removed* volatile detail — `Postgres agent_env`, `#1079`, the JSON-file path (4); no new subsystem (5); no architecture page moved or deleted (6); no ADR changed state (7).

**Correction to my previous report,** since it fed a decision: I reported the page at **39,998 of 40,000 characters**. That was measured in bytes. The gate (`scripts/doc-size.mjs:39`) measures `content.length` — UTF-16 code units — and the page is dense with em-dashes. At this HEAD it is **39,991 chars (9 under)**, while `wc -c` reports 40,186, which would read as *over* the cap. The near-the-cap conclusion holds, but my number was measured the wrong way and the two units disagree about which side of the line the page is on.

**No findings.**

🔁 The 4 findings-free rule checks from the previous review still hold (see review at `2b5098ae`) — dependency direction, module public API, optional-dependency shape, and composition-root placement, which the new `refreshEnv({ force })` seam reinforces: the policy decision lives at `server.ts:168`, the driver still only reports a classified fact.

Two checks new to this HEAD: `refreshEnv(opts?)` is not an optional-dependency violation (that rule governs injected deps, not method params; `opts?.force ?? true` preserves existing callers' semantics), and `state-builder.ts`'s direct `db` access is pre-existing here — 5 service files against 21 infrastructure files — so adding an `orderBy` to an existing query neither introduces nor amplifies it.

### Verdict
COMMENT — both prior warnings are genuinely resolved, and the `Object.hasOwn` fix went beyond the flagged line to its siblings. One follow-up on the hash's duplicate-key ordering; not blocking.

---

xjacka:
Re-review follow-up addressed in e9a1776b: `contributionHash` now tie-breaks same-key contributions on their canonical content, so the hash is order-independent even for duplicate keys and the state builder's ORDER BY goes back to being belt-and-braces rather than load-bearing.

Semantics note on why this is safe: the hash algorithm change doesn't alter the built contribution array (ORDER BY governs that), so agents see the same snapshot — at worst one extra dispatch after deploy for agents whose duplicate-key sequence isn't already content-sorted, which reconciles as a no-op. And should an unordered producer ever reappear, two snapshots differing only in duplicate-key order now hash identically and are correctly treated as the same state instead of triggering a delivery.

packages/api-server/src/modules/runtime-delivery/domain/contribution-hash.ts
```diff
@@ -2,15 +2,26 @@ import { createHash } from "node:crypto";
 import type { Contribution } from "api-server-api";
 
 export function contributionHash(contributions: Contribution[]): string {
+  // Duplicate keys tie-break on content, so the hash never depends on input
+  // order — even where the stable sort would otherwise preserve it.
   const sorted = [...contributions]
     .map(canonicalize)
-    .sort((a, b) => (a.k < b.k ? -1 : a.k > b.k ? 1 : 0));
+    .sort((a, b) => cmp(a.k, b.k) || cmp(a.json, b.json));
```
dam-code-guardian, line 9: 🟡 **Warning:** The tie-break makes the hash order-independent, but the *delivered* array keeps input order and the env driver is order-sensitive — first-occurrence-wins for ordinary names, and `KUBECONFIG` paths joined in order (`env-plugin.ts:37-47`). Since `service.ts:62` dispatches only when `input.state.hash !== local.lastAppliedHash`, two states that render different env now share a hash and the second one is silently skipped.

Verified by porting this function and the driver's render logic verbatim and simulating the gate (controls: identical input, distinct-key reorder, and a genuine value change all behave correctly under both the old and new hash):

```
duplicate name  [SHARED=one, SHARED=two] → [SHARED=two, SHARED=one]
  env: {SHARED:one} vs {SHARED:two}   hash NOW equal: true   (PREV: false)
two clusters    [KUBECONFIG=/a, /b]    → [KUBECONFIG=/b, /a]
  env: /a:/b vs /b:/a                 hash NOW equal: true   (PREV: false)
  → gate skips dispatch; live env stays on the first ordering
```

Both duplicate-key shapes are designed states, not synthetic: `state-builder.ts:58` orders user env ahead of connection env precisely so it shadows, and multi-cluster grants each contribute their own `KUBECONFIG` path (`kubernetes-contributions.ts:81`).

Unreachable today — the `ORDER BY connections.createdAt, id` added in the prior round is total, so the swap can't occur. But that ORDER BY was the thing this commit set out to stop relying on, and the previous hash was what would have *caught* a regression in it; now a reordering silently delivers stale env instead. Tie-breaking on `k` + `json` is right for canonicalization; the gap is that `contributionHash` canonicalizes a set while the driver consumes a list. Hashing what the driver actually derives (or deduplicating to the driver's precedence rule before hashing) would make the hash agree with the semantics it gates.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `e9a1776b`

## PR #3145: fix(agent-runtime): only recycle the harness when the env variable set changes
**Author:** xjacka | **Branch:** fix/env-rotation-recycle → main | **Changes:** +215 −24 (10 files)

### Summary
Re-review at `e9a1776b`: the prior 🟡 is addressed, but the remedy removes the *detector* of the ordering dependency rather than the dependency — the hash is now blind to reorderings the env driver still treats as semantic.

### Changes since last review
Previous HEAD: `d2f56981` (2026-08-04T08:12:35Z) — verdict COMMENT

- ✅ **Fixed:** `contributionHash` is now order-independent for duplicate keys — tie-break on canonical JSON (`contribution-hash.ts:9`), with a regression test
- 🆕 **New:** making the hash order-blind lets two semantically different env states share a hash, so `applyState`'s dispatch-skip gate can leave stale env live (`contribution-hash.ts:9`)

### Findings
- 🟡 **Warning:** the hash now collides across states the env driver renders differently, and that hash gates dispatch (`packages/api-server/src/modules/runtime-delivery/domain/contribution-hash.ts:9`)

aligned

- 🟢 **Suggestion:** `EnvChange` is a cross-module notification contract defined inside an infrastructure driver and re-exported from the module boundary, where the architecture docs call for a domain event in the publishing module's `domain/events/` with a `type` discriminant and guard (`env-plugin.ts:12`, `index.ts:5`, `server.ts:166`). At base this bridge was a bare `onChange?: () => void`; this HEAD promotes it into a typed payload the acp module branches recycle behavior on.
- 🟢 **Suggestion:** the contribution-ordering invariant is now load-bearing but expressed as inline `.orderBy()` in an application-layer service, where no fake can assert it — a `ContributionSourceRepo` port returning contributions already in precedence order would put it behind a boundary (`state-builder.ts:97`).

🔁 The rule checks clean in the previous review still hold (dependency direction, module public API, optional-dependency shape, composition-root placement, pre-existing `db` access) — see review at `d2f56981`.

### Verdict
COMMENT — The duplicate-key fix does what it says, but it trades a false-positive recycle for a false-negative one: `service.ts:62` skips dispatch when the hash is unchanged, so a reordered duplicate-key set now delivers stale env instead of a spurious refresh. Not blocking — the total `ORDER BY` added in the prior round makes it unreachable today — but the ordering dependency it was meant to remove is still there, now unguarded.

---

xjacka:
Responding to the review at `e9a1776b` — no code changes this round, reasoning below.

**Order-blind hash (🟡).** I'd frame the tradeoff differently: the DB holds *one* state, and the two duplicate-key orderings are two *renderings* of it — connection order was never part of the domain model, so neither rendering is "the correct one" for the hash to defend. Given that, the two possible failure modes under a hypothetical future unordered producer are:

- old hash: every re-render flips the winner → dispatch → recycle churn (and before this PR, killed runs) — *convergent but harmful*;
- new hash: the agent keeps the winner it already applied until something real (value, name, grant) changes — *stale-but-stable, per pod*.

Between churn and stability for a state that is genuinely ambiguous, stability is the side to err on — and the actual fix for the ambiguity is the ORDER BY, which makes the rendering canonical at the only place that knows the precedence rule. Also worth noting the guard the old hash provided was itself only partial: it detected reorderings *of the delivered array*, not the ambiguity — pods booting at different times could already disagree on the winner pre-PR. If we want a real guard, I think it's grant-time surfacing of env-name collisions in the UI, not a hash property — happy to file that as a follow-up issue.

**`EnvChange` as a domain event (🟢).** Declining for idiom consistency: agent-runtime has no event bus or `domain/events/` anywhere — its plugin system is uniformly callback-deps wired at the composition root (`skill-install-plugin`'s `install: SkillInstallFn` is the same shape). The two modules never import each other; the root mediates, which the first review endorsed as the right seam ("the driver reports a classified fact; the root decides"). Promoting one boolean callback to a typed event with discriminant + guard would introduce event infrastructure to the package for a single subscriber.

**`ContributionSourceRepo` port (🟢).** Fair point — `db`-in-service is the pre-existing pattern here (the review counts 5 such files), and moving the readers behind an infrastructure port that owns "contributions arrive in precedence order" is the right shape for it. Deferring to a follow-up refactor rather than growing this fix; the suggestion is recorded here for it.

packages/agent-runtime/src/modules/acp/services/acp-runtime.ts
```diff
@@ -1433,8 +1435,8 @@ export function createAcpRuntime(deps: AcpRuntimeDeps): AcpRuntime {
         recycleAgentForEnv();
         return;
       }
-      // A turn is in flight: recycle at the next turn boundary, or force after a bound.
-      if (!envForceTimer)
+      // A turn is in flight: recycle at the next turn boundary; force after a bound only when asked.
+      if ((opts?.force ?? true) && !envForceTimer)
```
jjeliga: The deferral is now unbounded, and envRefreshPending has exactly one drain: the response branch that requires active.outboundId === outboundId. Outside recycleAgentForEnv itself, nothing else clears activePromptBySession. So if the harness wedges mid-turn, or that response never lands / lands with a mismatched id, the pending refresh is stranded until the pod rolls — the 60s timer used to bound exactly that.

It's also silent: recycleAgentForEnv logs recycling harness to apply env change and the driver logs (values only), but nothing records "a refresh is pending, held by N in-flight session(s)". For a bug that needed a Slack forensics thread to find, I'd add that log line here, and either give the deferral a generous ceiling (minutes, not 60s) or re-check maybeRecycleForEnv() on the other transitions where the map can drain rather than trusting one call site indefinitely.

packages/agent-runtime/src/modules/acp/services/acp-runtime.ts
```diff
@@ -89,8 +89,10 @@ export interface AcpRuntime {
   attach(channel: ClientChannel, opts?: { viewer?: boolean }): void;
   status(): AcpRuntimeStatus;
   resetSession(sessionId: string): void;
-  /** Env on disk changed: recycle a running harness so it respawns with the new env. */
-  refreshEnv(): void;
+  /** Env on disk changed: recycle a running harness so it respawns with the
+   * new env. `force: false` never kills an in-flight turn — it waits for the
+   * next turn boundary instead of forcing after the bound. */
+  refreshEnv(opts?: { force?: boolean }): void;
```
jjeliga: Nit: opts?.force ?? true keeps existing callers working, but the bug being fixed is an implicit force. Making the flag required and updating the handful of call sites means a future caller can't inherit the kill behaviour by omission.

packages/agent-runtime/src/__tests__/unit/acp-runtime.test.ts
```diff
@@ -2511,4 +2511,57 @@ describe("createAcpRuntime — platform _meta round-trip", () => {
       scheduleId: "sch-1",
     });
   });
+
+  it("a deferred env refresh (force: false) outlives the force bound and recycles at the turn boundary", () => {
```
jjeliga: These are appended inside describe("createAcpRuntime — platform _meta round-trip") (opens at :2209), so they report under a heading unrelated to what they test. They want their own describe("createAcpRuntime — env recycle").

Scaffolding is heavy: the two cases repeat 10 lines of setup. A local startTurn({ envForceRecycleMs }) returning { fa, endTurn } takes each body to ~5 lines. (Keep the vi.useFakeTimers() + try/finally wrapper — it matches four existing occurrences in this file.)

Trimming the setup is worth the churn because two seams remain untested, and each currently costs 15 lines instead of 5:
- The force timer surviving a downgrade — a set change arms the timer, then a value-only change arrives before it fires. (opts?.force ?? true) && !envForceTimer is correct today, in both orders, but nothing pins it.
 - A queued prompt deferring the recycle — claimed in the PR body, and true only because advanceQueue → forwardPromptToAgent repopulates activePromptBySession two lines before maybeRecycleForEnv(). Swap those statements and queued work dies with no test failing.

packages/api-server/src/modules/runtime-delivery/services/state-builder.ts
```diff
@@ -92,6 +94,8 @@ async function readGrantedContributions(
   db: Db,
   agentId: string,
 ): Promise<Contribution[]> {
+  // Ordered oldest-first: downstream is order-sensitive and unordered rows
```
jjeliga, line 97: separate from the ordering-fragility threads: I don't think this half explains the 29 kills.

A row-order flip only recycles if the rendered env map changes — otherwise envEquals suppressed it already, before this PR. Per the duplicate-key analysis below, the only row-order-sensitive collision is two granted connections sharing an env name (KUBECONFIG being the built-in case); user env is ordered in code and PK-unique, so it can't flip.

Did the affected agent have two such connections? If not, no reorder could have produced a recycle there, and layer 1 is the entire fix. Either way I'd soften "removing the source of spurious env-changed deliveries" in the body rather than close #3143 on this explanation.

packages/agent-runtime/src/__tests__/unit/env-plugin.test.ts
```diff
@@ -0,0 +1,74 @@
+import { describe, it, expect } from "vitest";
```
jjeliga, line 1: this file duplicates env-plugin-kubeconfig.test.ts verbatim: same ctx, same env() helper, and harness() is that file's applyEnv() plus a changes array — ~25 of 74 lines, with driver behaviour now split across two files. Merging into one env-plugin.test.ts whose fake store returns { env, changes } takes each of the four new cases to ~4 lines.

packages/agent-runtime/src/modules/acp/services/acp-runtime.ts
xjacka: Both taken in f6e0e4d3: the deferral now logs `env recycle deferred: N turn(s) in flight`, and `tearDownSession` clears the session's turn slot + queue and re-checks `maybeRecycleForEnv()` — so a reset session drains the pending refresh instead of stranding it (that slot also fed the idle flag, so the stranding was pre-existing surface beyond this PR).

On the ceiling I'd push back: the issue's own distribution has legitimate runs at 0.6–107.6 min, so any minutes-scale ceiling still kills real work — there is no safe constant between "wedged" and "long run". For a truly wedged harness the stuck turn already pins the pod busy and blocks hibernation; env staleness is the least of its problems, and the recycle-for-env timer was only ever curing it by accident. If wedge detection is worth having, it should key on the turn itself (no output for X), not on an unrelated env delivery happening to arrive.

packages/agent-runtime/src/modules/acp/services/acp-runtime.ts
xjacka: Done in f6e0e4d3 — `refreshEnv(opts: { force: boolean })`, callers updated.

packages/agent-runtime/src/__tests__/unit/acp-runtime.test.ts
xjacka: Restructured in f6e0e4d3: own `describe("createAcpRuntime — env recycle")` with a `startTurn()` helper (kept the fake-timers try/finally shape), and both missing seams are pinned — the force timer surviving a later deferred refresh in both orders, and a queued prompt deferring the boundary recycle (the test fails if `advanceQueue` and `maybeRecycleForEnv` are swapped).

packages/api-server/src/modules/runtime-delivery/services/state-builder.ts
xjacka: Agreed — I can't verify the affected agents' grants from the code side, and you're right that user env can't flip (PK-unique, ordered in code), so a duplicate-name grant is the only shape that works. Softened the body accordingly and switched `Fixes` → `Refs`, so #3143 stays open pending verification against an affected agent's state (its env snapshot on the PVC would show the flipped winner directly). Layer 1 stands on its own either way: whatever triggered the deliveries, they no longer kill runs.

packages/agent-runtime/src/__tests__/unit/env-plugin.test.ts
xjacka: Merged in f6e0e4d3 into a single `env-plugin.test.ts` — one harness returning `{ apply, env, changes }`, both describes share it, the kubeconfig cases are unchanged and the classification cases are down to ~4 lines each.

jjeliga:
acp-runtime.ts:1094 — the recycle fires before the turn-end fan-out, and this PR makes that the primary path. In the mapping.promptSessionId !== null block, maybeRecycleForEnv() runs before recordActivity, the platformTurnEnded fan-out, and maybeCloseIdleSession. The recycle clears engagedSessions and sessionLogs, so engaged viewers never receive the turn-ended notification for the turn that triggered it (they get the 1011 close instead), and the following appendAndFanOut re-creates a stub SessionLog via getOrCreateLog holding only that orphan entry — which a later cold session/load replays ahead of the rehydrated history. Pre-existing on main, but value-only changes now route through this path routinely instead of rarely. The fix is free: move the maybeRecycleForEnv() call to the end of the block. reapIdleSessionNow already guards on !agent, so nothing else there needs the harness alive.

acp-runtime.ts:685 — the boundary gate is narrower than the runtime's own idle definition. maybeRecycleForEnv gates on activePromptBySession.size === 0, while status().idle — what the hibernation path trusts — also requires pendingFromAgent, the prompt queues, and backgroundWork to be empty. recycleAgentForEnv calls deps.backgroundWork?.clear() and the harness takes its children down with it, so reported background work is still killed at the boundary. Same as main, so not a regression — but given the premise of this PR is "don't destroy in-flight work", reusing the predicate that already encodes what in-flight means seems like the consistent call.

xjacka:
Two more review points addressed in 05e44772:

**Recycle ordering vs. the turn-end fan-out** — the boundary `maybeRecycleForEnv()` moved to the end of the `promptSessionId` block (kept conditional on the turn actually completing). Viewers now receive the `platformTurnEnded` fan-out before any recycle closes their channels, and the fan-out no longer recreates a stub session log holding an orphan turn-ended entry that a later cold `session/load` would replay ahead of rehydrated history. `maybeCloseIdleSession` runs before it; its `tearDownSession` path re-checks the deferred recycle itself, so the reap-first order holds either way.

**Boundary gate vs. the idle definition** — `maybeRecycleForEnv` (and `refreshEnv`'s immediate path) now gate on a shared `runtimeBusy()` predicate — the exact inverse of `status().idle`: turns, pending agent-initiated requests, prompt queues, and reported background work. A recycle can no longer kill background work the idle flag still reports as busy. The widened gate brings new drain points, each re-checking the deferred recycle: an answered or TTL-expired agent request, and a released background-work hold — the registry gained an optional `onRelease` listener the runtime subscribes to (report-empty, forget, and clear all notify; re-entrancy is safe because `recycleAgentForEnv` clears `envRefreshPending` before it clears the registry). The deferral log now reports all three hold sources. Note `force: true` (variable-set changes) still overrides the gate after the bound, as before.

Seam test added: reported background work holds a deferred refresh; releasing it drains the recycle. 225 agent-runtime tests green.

jjeliga (approved):

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `5e57854e`

## PR #3145: fix(agent-runtime): only recycle the harness when the env variable set changes
**Author:** xjacka | **Branch:** fix/env-rotation-recycle → main | **Changes:** +357 −98 (12 files)

### Summary
Re-review at `5e57854e` after `f6e0e4d3` (observability + drainability) and `05e44772` (recycle respects the full idle definition). Both of jjeliga's structural asks are genuinely resolved and the shared `runtimeBusy()` predicate is the right consolidation; two new re-check call sites fire while work the predicate can't see is still in flight.

### Changes since last review
Previous HEAD: `e9a1776b` (2026-08-04T12:34:33Z) — verdict COMMENT

- ✅ **Fixed:** the order-blind-hash 🟡 is resolved by scope, not by code — `Refs` replaced `Fixes` for #3143 and the state-builder body was softened, so the hash no longer carries the load the finding was about. Re-verified against the predicate: `contributionHash` is still order-blind at this HEAD, and `state-builder.ts:109`'s total `ORDER BY` still makes the collision unreachable.
- ✅ **Fixed:** the unbounded-deferral strand — `refreshEnv` now logs the deferral with turn/request/hold counts, and `tearDownSession` clears the session's turn slot and queue then re-checks. Verified the drain path: a released background hold reaches `maybeRecycleForEnv()` through the new `onRelease` listener.
- ✅ **Fixed:** `refreshEnv(opts: { force: boolean })` is required now; the one production call site passes `{ force: namesChanged }`.
- ✅ **Fixed:** recycle moved after the turn-end fan-out, gated on `turnEnded` — viewers get `platformTurnEnded` before any 1011 close.
- 🆕 **New:** the client-response re-check recycles while the agent's turn is still open (`acp-runtime.ts:1190`)
- 🆕 **New:** `tearDownSession` re-checks before releasing the closing session's background hold (`acp-runtime.ts:787`)

### Findings
- 🟡 **Warning:** answering a permission prompt can recycle the harness mid-turn (`packages/agent-runtime/src/modules/acp/services/acp-runtime.ts:1190`)
- 🟡 **Warning:** `tearDownSession` re-checks the recycle one line before dropping the hold that would have blocked it (`packages/agent-runtime/src/modules/acp/services/acp-runtime.ts:787`)
- 🟢 **Suggestion:** `onRelease?()` adds optionality to a port method purely so test fakes can omit it (`packages/agent-runtime/src/modules/acp/services/background-work-registry.ts:21`)

**Verdict:** aligned

`connections.md:213` is edited in this diff and its new wording ("only a variable-set change may force one mid-turn — value-only changes spare in-flight runs") matches the predicate at this HEAD; `Last verified: 2026-08-04` is bumped on an edited page. Check 1 does not fire for `agent-lifecycle.md:70` — "delivers all of it at the next idle turn" stayed accurate from `d9b64b6d` on, and this delta widens the idle *definition* without changing the delivery contract. Checks 3–7 produce nothing.

Measured `connections.md` at **39,991 chars** (`content.length`, per `scripts/doc-size.mjs:39`) — 9 under the 40,000 cap.

- 🟢 **Suggestion:** `onRelease?(cb)` on the `BackgroundWorkRegistry` port is a new optional dependency method whose docstring states the reason — "fakes without it behave as before". The Infrastructure rule on required dependencies rejects that axis specifically: the consumer branches on presence (`deps.backgroundWork?.onRelease?.(…)`, `acp-runtime.ts:702`), a missing wiring degrades to a silent skip rather than a type error, and "don't loosen a required dep to optional to ease test setup — pass the real (or a stubbed) implementation" covers the stated motive. `compose.ts:42` always wires the real registry, so production never exercises the optional branch. (`architecture/infrastructure.md` → *Required Dependencies — Never Optional*)

🔁 The rule checks clean in the previous reviews still hold (dependency direction, module public API, composition-root placement, pre-existing `db` access) — see review at `e9a1776b`. `runtimeBusy()` is a pure local predicate in the service layer and `contribution-hash.ts` stays dependency-free apart from `node:crypto`, which it already had.

### Verdict
COMMENT — the two structural asks from this round are properly fixed and the `runtimeBusy()` consolidation is the right call, but both new re-check sites can fire while the runtime is still busy in ways the predicate doesn't cover. Not blocking.

---

packages/agent-runtime/src/modules/acp/services/acp-runtime.ts
```diff
@@ -1162,6 +1187,7 @@ export function createAcpRuntime(deps: AcpRuntimeDeps): AcpRuntime {
       pendingFromAgent.delete(frame.id);
       if (pending.sessionId) updateOrphanTimerForSession(pending.sessionId);
       a.send(frame);
+      maybeRecycleForEnv();
```
dam-code-guardian: 🟡 **Warning:** this re-check can recycle the harness in the middle of a turn the agent is still running.

The client's response to an agent-initiated request (a permission prompt) is forwarded with `a.send(frame)`, then `maybeRecycleForEnv()` runs. `runtimeBusy()` is false whenever `pendingFromAgent` was the *only* busy signal — and it can be, because the turn slot and the pending request have independent lifetimes: `activePromptBySession` is keyed on the prompt's `outboundId` and only cleared when that exact response lands (`:1110`), while `pendingFromAgent` is populated by any agent→client request (`:971`). A prompt whose slot was already cleared — e.g. `resetSession` / `reapIdleSessionNow` → `tearDownSession` (`:786-787`, new in this diff) — leaves a live permission request as the last busy signal.

Simulated against the extracted predicates (control: same scenario with the turn slot held ⇒ no recycle, correct):

```
A: turn slot held + pending request  → killed=false   ✓
B: pending request only              → killed=true    ✗ recycle mid-turn
```

The user answers the prompt and the harness is killed as the agent resumes work — the outcome `d9b64b6d` set out to prevent. The other two new re-check sites are safe by comparison: `expireSession` (`:587`) runs only when nothing is engaged, and the turn-boundary call (`:1134`) is gated on `turnEnded`.

Forwarding a response is the *start* of the agent doing more work, not a boundary — the cheapest fix is to drop this call and let the existing boundary handle it:

```suggestion
      a.send(frame);
```

packages/agent-runtime/src/modules/acp/services/acp-runtime.ts
```diff
@@ -763,6 +779,12 @@ export function createAcpRuntime(deps: AcpRuntimeDeps): AcpRuntime {
     }
     sessionLogs.delete(sessionId);
     for (const cursors of channelCursors.values()) cursors.delete(sessionId);
+    // session/close cancels the session's ongoing work, so its turn slot and
+    // queue must not outlive it — a stale slot would strand a deferred env
+    // recycle (and the idle flag) until the pod rolls.
+    activePromptBySession.delete(sessionId);
+    promptQueueBySession.delete(sessionId);
+    maybeRecycleForEnv();
```
dam-code-guardian: 🟡 **Warning:** this re-check runs one line before the `forget()` that would have made it safe, so a session closing with live background work can recycle the harness while that work runs.

`maybeRecycleForEnv()` here (`:787`) sees the hold still in `holds`, but `activePromptBySession`/`promptQueueBySession` were just cleared — so if the hold is the only remaining busy signal for *another* session, `runtimeBusy()` is already false and the recycle fires; and when the hold belongs to the closing session, `deps.backgroundWork?.forget(sessionId)` at `:790` fires `notifyRelease()` → the `onRelease` listener → `maybeRecycleForEnv()` again anyway. Simulated:

```
C: tearDownSession, hold on the closing session → killed=true (recycle)
D: control — hold released normally             → killed=true (documented path)
```

Via `reapIdleSessionNow` this is unreachable (it gates on `hasWork` at `:827`), but `resetSession` has no such guard and is externally reachable — `POST /api/sessions/<id>/reset` (`server.ts:486`), used by the UI and `dam schedule reset-session`. Resetting one session then kills a harness that is supervising another session's background jobs, which is what the background-work contract exists to prevent.

Moving the re-check after `forget()` makes the drain path the single route and removes the double call:

```suggestion
    promptQueueBySession.delete(sessionId);
```

packages/agent-runtime/src/modules/acp/services/background-work-registry.ts
```diff
@@ -16,6 +16,9 @@ export interface BackgroundWorkRegistry {
   forget(sessionId: string): void;
   /** Harness recycled or exited: same, for every session it served. */
   clear(): void;
+  /** Invoked whenever a hold is released, so waiters (e.g. a deferred env
+   * recycle) can re-check. Optional — fakes without it behave as before. */
+  onRelease?(cb: () => void): void;
```
dam-code-guardian: 🟢 **Suggestion:** the docstring names the reason this is optional — "fakes without it behave as before" — which is the one motive the architecture's required-dependency rule rules out (`architecture/infrastructure.md` → *Required Dependencies — Never Optional*: "don't loosen a required dep to `optional` to ease setup — pass the real (or a stubbed) implementation").

The cost is at the consumer: `deps.backgroundWork?.onRelease?.(…)` (`acp-runtime.ts:702`) is a double-optional call where a fake that silently lacks the method turns the deferred-recycle drain into a no-op — a missing wiring that reads as a silent skip instead of a type error. `compose.ts:42` always wires the real registry, so production never takes the absent branch; only test fakes do.

Making it required and adding `onRelease: () => {}` to the fakes that don't care keeps the drain path type-enforced.

packages/agent-runtime/src/modules/acp/services/acp-runtime.ts
xjacka: Taken in 164c0429 — the call is dropped with a comment stating why (forwarding the answer is the start of more agent work, not a boundary). The pending-only-busy state drains conservatively: the refresh waits for the next real boundary instead of firing here.

packages/agent-runtime/src/modules/acp/services/acp-runtime.ts
xjacka: Fixed in 164c0429 — the re-check moved after `forget()`. The onRelease listener is the route for the held case (it fires inside `forget` once the hold is gone); the explicit call after it stays load-bearing for the no-hold case — a session reset mid-turn releases no hold, so nothing else would drain the slot it just cleared.

packages/agent-runtime/src/modules/acp/services/background-work-registry.ts
xjacka: Done in 164c0429 — `onRelease` is required; the one fake gained `onRelease: () => {}` and the consumer call is single-optional (`deps.backgroundWork?.onRelease(…)`, the registry dep itself staying optional per its documented contract).

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** 🛡️ — ✅ Code Review @ `164c0429`

## PR #3145: fix(agent-runtime): only recycle the harness when the env variable set changes
**Author:** xjacka | **Branch:** fix/env-rotation-recycle → main | **Changes:** +360 −98 (12 files)

### Summary
Re-review at `164c0429` after one commit (`recycle re-checks only at true boundaries`). It resolves both 🟡 warnings and the 🟢 suggestion from the previous round; no new findings.

### Changes since last review
Previous HEAD: `5e57854e` (2026-08-05T11:55:51Z) — verdict COMMENT

- ✅ **Fixed:** answering a permission prompt no longer recycles mid-turn — the `maybeRecycleForEnv()` after `a.send(frame)` is gone; forwarding the answer resumes agent work, not a boundary (`acp-runtime.ts:1192`)
- ✅ **Fixed:** `tearDownSession` recycles *after* `forget(sessionId)`, so the closing session's own hold can no longer block the drain (`acp-runtime.ts:789,791`)
- ✅ **Fixed:** `onRelease` is now a required port method — call site drops the `?.` and the test fake implements it (`background-work-registry.ts:21`, `acp-runtime.ts:702`)

### Findings

**Verdict:** aligned

The delta touches no docs, and it makes the code *more* faithful to `connections.md:213` — "only a variable-set change may force one mid-turn" — by removing the value-only mid-turn recycle path. `Last verified: 2026-08-04` needs no bump (page unedited in this delta); 39,991 / 40,000 chars.

**No findings.** The prior 🟢 is resolved: `onRelease` is now required on the `BackgroundWorkRegistry` port, the consumer calls it unconditionally, and the test fake implements it — matching *Required Dependencies — Never Optional* (`architecture/infrastructure.md`).

🔁 The findings-free rule checks from the prior rounds still hold (dependency direction, module public API, composition-root placement) — see review at `5e57854e`.

### Verdict
APPROVE — the final commit closes every open finding from the previous rounds with no new issues; the recycle now fires only at genuine idle boundaries.

---

PR #3147, opened 2026-08-04
feat(ui): sandbox-level usage metrics on Configure Sandbox

### LINKED ISSUE #2774: UI - Add sandbox-level usage metrics to Configure Sandbox page
## Description

Add a **Usage** section to the **Configure Sandbox** page that provides usage metrics scoped to the selected sandbox.

This should mirror the existing global **Usage** page in Settings, but only show data relevant to the individual sandbox. This will help users understand resource consumption and activity for a specific sandbox without needing to navigate away.

## Requirements

- Add a Usage section/tab within the Configure Sandbox experience
- Display usage metrics scoped to the selected sandbox
- Maintain consistency with the existing global Usage page where applicable
- Define which metrics are available at the sandbox level

## Related

#2877

## Rough prototype

Image

### LINKED ISSUE #3148: UI - Usage flashes a skeleton on every month change, and re-runs a doomed query when metrics are disabled
Changing the month on Settings › Usage re-flashes the full four-section skeleton before showing anything, because the query is keyed on the month range and a new key has no cached data. On a deployment without the telemetry store it's worse: every month change fires a request that's guaranteed to fail, so the page flickers skeleton → "Usage metrics are unavailable on this deployment." each time.

Two fixes: keep the previous month's figures on screen while the next one loads (`placeholderData: keepPreviousData`), and treat `PRECONDITION_FAILED` as the deployment-level verdict it is — show the unavailable state once, above the period control, instead of re-deriving it per month.

kapetr:
Adds a **Usage** section to the Configure Sandbox page, so you can see what one sandbox costs without leaving it for the global Usage page in Settings.

For a calendar month, scoped to the selected sandbox: total cost, API calls, tokens in / out, and model time as headline figures, above a spend-by-model breakdown and the spend-by-day chart. The month is chosen with the same ‹ › switcher Settings › Usage uses, and the section's entry in the sandbox nav carries the month-to-date figure so it's visible without opening the section.

Spend a sandbox delegated to another agent through an Invocation counts under the sandbox that drove it, matching how the platform attributes telemetry everywhere else.

Spend by model is now the same bar chart as spend by agent, on both Usage surfaces, replacing the per-model table.

Along the way this also fixes the Usage page flashing its whole skeleton every time you change the month, and re-running a request that cannot succeed on deployments without a telemetry store. That bug is pre-existing, but the new section would have inherited it — both surfaces share the same components.

Closes #2774
Fixes #3148

## Sub-issues

- [x] 01 — Narrow the spend breakdown to one agent, add model time
- [x] 02 — Extract the month-period plumbing both Usage surfaces share
- [x] 03 — Stop the Usage tab flashing a skeleton on every month change
- [x] 04 — Sandbox Usage section, reachable from the sandbox nav
- [x] 05 — Live month-to-date figure on the Usage nav line

## Test instructions
Run `CLICKSTACK=1 mise run cluster:install` to build platform with telemetry.


packages/ui/src/modules/metrics/components/sandbox-usage-section.tsx
```diff
@@ -0,0 +1,144 @@
...
+}
+
+/** Sandbox-home "Usage" section: this sandbox's LLM spend for one calendar
+ *  month, headline figures above the same per-model table and per-day chart the
+ *  global Usage tab shows. */
+export function SandboxUsageSection({ agentId }: { agentId: string }) {
+  const [month, setMonth] = useState(() => monthStart(new Date(), 0));
+  const { from, to, isCurrentMonth } = monthRange(month);
+  const timeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+  const { data, isPending, isError, isPlaceholderData, error } =
```
dam-code-guardian: 🔴 **Critical:** this section shows **another sandbox's** spend figures under this sandbox's heading whenever you navigate between two sandboxes' Usage sections.

`useSpendBreakdown` sets `placeholderData: keepPreviousData` to smooth **month** paging — correct for that case, and its docstring says so. But `agentId` is also part of the query key, and this component is rendered **unkeyed** (`sandbox-home-view.tsx:84`), so switching sandboxes keeps the same observer mounted. React Query applies `keepPreviousData` whenever `data === undefined && status === 'pending'` — it does not care that the key changed *identity*, only that the new key is uncached. Simulating the observer's exact gate:

```
view sandbox A (cold)        -> data=undefined      placeholder=false
A resolves                   -> data="A:$42.00"     placeholder=false
switch to sandbox B (cold)   -> data="A:$42.00"     placeholder=true   <- B's page, A's money
B resolves                   -> data="B:$7.00"      placeholder=false
```

The only mitigation is `opacity-40`, which reads as "refreshing", not "these belong to a different sandbox". Total cost, API calls, tokens, model time, the per-model bars and the day chart all show A's figures under B's name until B resolves. `useSettledMonth` doesn't help — it faithfully tracks the *month*, so the numbers end up attributed to the right month and the wrong agent. On a billing surface that is a wrong number, not a slow one.

The month-paging behaviour this exists for is unaffected by the fix — keying the section per agent scopes the placeholder to one sandbox while leaving in-section month paging smooth:

```suggestion
    useSpendBreakdown(from, to, timeZone, agentId);
```

(and at the call site in `sandbox-home-view.tsx`: `<SandboxUsageSection key={agent.id} agentId={agent.id} />`). Scoping the placeholder to a same-`agentId` predecessor inside the hook works too, and generalises better if another agent-scoped caller appears.

packages/ui/src/modules/metrics/views/usage-view.tsx
```diff
@@ -1,68 +1,50 @@
...
+  monthLabel,
+  monthRange,
+  monthStart,
+} from "../lib/month-range.js";
+import { isMetricsUnavailable } from "../lib/unavailable.js";
 
 /** Settings tab: the user's LLM API spend for one calendar month, totalled
  *  and broken down per model across all their agents. */
 export function UsageView() {
   const [month, setMonth] = useState(() => monthStart(new Date(), 0));
```
dam-code-guardian: 🟡 **Warning:** this block and `sandbox-usage-section.tsx:56-66` are now nine lines of **identical** stateful orchestration — month seed, `monthRange(month)`, timeZone resolution, the `useSpendBreakdown` destructure, `isMetricsUnavailable(error)`, `useSettledMonth`, and all three render gates including `(isPlaceholderData || isError) && "opacity-40"`.

The PR extracted the *pure* pieces well (`month-range.ts`, `use-settled-month.ts`, `unavailable.ts`, `MonthSwitcher`, `SpendBar`) and stopped at the stateful ones. Two consequences are already visible in the diff: each file declares its own `UsageSkeleton` with a different shape (`:140` here, `:133` there), and the error/empty card was extracted into a `Notice` helper in the new file but left as inlined `<Card><p>` markup at `:72-84` here.

A `useMonthlySpend(agentId?)` hook returning `{month, setMonth, isCurrentMonth, shownMonth, data, unavailable, isPending, isError, isPlaceholderData}` collapses all nine lines in both surfaces — and it is where the cross-sandbox placeholder fix belongs, so that scoping is right in one place instead of two.

packages/ui/src/modules/metrics/api/queries.ts
```diff
@@ -1,16 +1,55 @@
...
+  const { from, to } = monthRange(monthStart(new Date(), 0));
+  const timeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
   return useQuery({
-    ...trpc.metrics.spendBreakdown.queryOptions({ from, to, timeZone }),
+    ...trpc.metrics.spendBreakdown.queryOptions(
+      agentId ? { from, to, timeZone, agentId } : skipToken,
+    ),
     staleTime: 60_000,
     retry: false,
+    select: (data) => data.byModel.reduce((sum, row) => sum + row.costUsd, 0),
```
dam-code-guardian: 🟢 **Suggestion:** "total spend for this scope" is now derived three times from the same rollup — here, at `usage-view.tsx:42`, and inside `totals()` at `sandbox-usage-section.tsx:28`. The response carries per-model rows but no total, so every consumer re-derives it, and if "total" ever gains a rule (excluding a model class, handling a currency) three sites must change together.

One exported `totalCostUsd(byModel)` in `metrics/lib/` used by all three collapses it to a single definition. Putting the total on the response is stronger but widens the contract; the helper is the smaller change.

packages/api-server-api/src/modules/metrics/router.ts
```diff
@@ -24,5 +25,8 @@ export const metricsRouter = t.router({
   // a single loading/error state.
   spendBreakdown: readAgentProcedure
     .input(metricsSpendBreakdownInputSchema)
-    .query(({ ctx, input }) => ctx.metrics.spendBreakdown(input)),
+    .query(({ ctx, input }) => {
+      if (input.agentId) checkAgentBinding(ctx, input.agentId);
```
dam-code-guardian, line 31: 🟢 **Suggestion:** the router branches on input shape to apply an auth check before delegating, where the contract layer's job is to validate and delegate immediately. Worth saying plainly first: **this is not a security hole.** `ownedScope` returns `[]` for an unowned agent, and `listOwnedAgentIds` is already narrowed to the key's binding at the composition root (`app.ts:981-985`), so the service alone would deny the request — this check is defense-in-depth that turns empty rows into a `FORBIDDEN`.

What the PR changes is the *convention*: the comment above was rewritten from "When a specific agentId is requested" to "Whenever a procedure is narrowed to a specific agentId", generalizing a single-procedure exception into a surface-wide one as the second procedure adopts it. That leaves authorization split across two layers — ownership in the service, key-binding in the router — so a reader must check both to know what guards a read, and a third procedure added later inherits the split by default.

Moving `checkAgentBinding` into `ownedScope` (or a thin wrapper it calls) puts the whole ownership decision in one function and returns both procedures here to single-line delegation.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `58e566b3`

## PR #3147: feat(ui): sandbox-level usage metrics on Configure Sandbox
**Author:** kapetr | **Branch:** feat/2774-sandbox-usage-metrics → main | **Changes:** +638 −408 (27 files)

### Summary
Adds a sandbox-scoped Usage section to Configure Sandbox, backed by new `agentId` narrowing on the `spendBreakdown` tRPC procedure plus a `durationMs` (summed per-call latency) rollup; extracts the month plumbing both Usage surfaces share and replaces the per-model table with bar charts on both.

### Findings
- 🔴 **Critical:** `keepPreviousData` renders one sandbox's spend under another's heading (`packages/ui/src/modules/metrics/components/sandbox-usage-section.tsx:59`)
- 🟡 **Warning:** nine lines of stateful orchestration duplicated across both Usage surfaces (`packages/ui/src/modules/metrics/views/usage-view.tsx:29`)
- 🟡 **Warning:** landing-page index still scopes metrics to the Settings tab only (`docs/architecture.md:73`)
- 🟢 **Suggestion:** total-cost reduction now written three times over the same rollup (`packages/ui/src/modules/metrics/api/queries.ts:52`)
- 🟢 **Suggestion:** router-level binding check splits the authorization model across two layers (`packages/api-server-api/src/modules/metrics/router.ts:29`)
- ✅ **Looks good:** `metrics-panel.tsx` (206 lines) was already unreferenced on `main` — confirmed via code search — so its deletion is dead-code removal, not a feature regression.

**Verdict:** minor drift

**Drift**

*Check 1 — the landing-page entry still scopes the subsystem to a single surface.* `docs/architecture.md:73` describes metrics as *"reads over the telemetry store backing the **Settings Usage tab**"*. This PR makes the subsystem back a second, sandbox-scoped surface (`schemas.ts:73`, `metrics-service.ts:91`, `router.ts:45`). The subsystem page itself was updated correctly — line 32 now states it "narrows to **one owned agent** on request, which is what backs the per-sandbox Usage section beside the global one" — but the index that points at it was not. Proposed: *"…backing the global and per-sandbox Usage surfaces, failing closed when the backend is disabled."*

*Check 1 — "The Usage tab treats that error as…" now speaks for two surfaces.* `docs/architecture/metrics.md:52` attributes the unavailable-state handling to "The Usage tab". The diff gives that fail-closed contract a second consumer: the new section renders the identical branch, and the nav summary deliberately degrades to a placeholder on `PRECONDITION_FAILED` (`use-section-summaries.ts:138-140`). Behavior unchanged; the sentence's subject is now too narrow. Proposed: "Both Usage surfaces treat that error as".

**Possible drift**

⚠️ *Check 1 — the Overview procedure's documented consumer may now have none in the browser.* Deleting `metrics-panel.tsx` removes the only UI consumer of `runtimeBySession` and `contextPerCall`; after it they are read only by the CLI (`packages/cli/src/modules/metrics/commands/show.ts`), while the page's diagram shows `browser user` as the sole actor. Nothing on the page is false — the question for a human is whether the page should now name the CLI as a first-class consumer.

Checks 2–7 clean: `Last verified` is already `2026-08-04`, no ADR references or state changes, no volatile content added (the only new code spans are `[from, to)` and a pre-existing `packages/api-server/` link), no new subsystem, page at 11,783 of 40,000 chars measured in the gate's own unit (`content.length`).

**1. Business logic in the validation layer** — `router.ts:29`. The rule: *"No business logic lives here — routers delegate immediately to the service interface"*, and auth checks belong to the application layer. The router branches on input shape (`if (input.agentId) checkAgentBinding(...)`) before delegating. This PR *amplifies* the pre-existing instance: the comment was rewritten from "When a specific agentId is requested" to "**Whenever a procedure is narrowed**", generalizing a one-off into a surface-wide convention as the second procedure adopts it. The result is a split authorization model — ownership in the service (`ownedScope`), key-binding in the router — so a reader checks two layers, and a third procedure inherits the split. **Not a security hole:** `ownedScope` returns `[]` for an unowned agent and `listOwnedAgentIds` is already binding-narrowed at the composition root (`app.ts:981-985`), so the service alone would deny it; the router check is defense-in-depth that returns `FORBIDDEN` rather than empty rows. The cost is duplication, not a gap.

**2. Application-layer arithmetic duplicated into the client query layer** — `queries.ts:52`. "Total spend for this scope" is derived three times from the same rollup: `queries.ts:52`, `usage-view.tsx:42`, and inside `totals()` at `sandbox-usage-section.tsx:28`. The response carries per-model rows but no total, so every consumer re-derives it; three copies of one definition must change together if "total" ever gains a rule.

**Smallest fix set:** move the binding check into `ownedScope` (clears both the new and pre-existing instance of 1, restoring single-line delegation); export one `totalCostUsd(byModel)` helper for the three sites (clears 2).

Layering otherwise verified clean: the `MetricsReader` port takes an already-resolved allowlist and does no scoping of its own, and `compose.ts` remains the only place touching concrete infrastructure.

**1. [CRITICAL] `keepPreviousData` shows one sandbox's spend under another sandbox's heading** — `sandbox-usage-section.tsx:59`. See the inline comment for the full trace and fix.

**2. [HIGH] Two Usage surfaces duplicate nine lines of identical orchestration** — `usage-view.tsx:29-47` / `sandbox-usage-section.tsx:56-66`. The PR extracted the *pure* pieces well (`month-range.ts`, `use-settled-month.ts`, `unavailable.ts`, `MonthSwitcher`, `SpendBar`) but copied the stateful wiring verbatim: month seed, `monthRange`, timeZone resolution, query destructure, `isMetricsUnavailable`, `useSettledMonth`, and all three render gates including `(isPlaceholderData || isError) && "opacity-40"`. Two consequences already visible: each file declares its own differently-shaped `UsageSkeleton` (`usage-view.tsx:140`, `sandbox-usage-section.tsx:133`), and the error/empty card became a `Notice` helper in the new file (`:42`) but stayed inlined markup in `usage-view.tsx:72-84`. A `useMonthlySpend(agentId?)` hook collapses all nine lines — and is where finding 1's fix belongs, so the placeholder scoping is right in one place instead of two.

**3. [MODERATE] Bars and stat figures carry no accessible text** — `spend-bar.tsx`, `spend-stat-cards.tsx`. `SpendBar` correctly marks its colour dot `aria-hidden` (`:23`), but magnitude is conveyed by width alone with no `role`/`aria-label`/`sr-only`. The sibling `SpendByDayChart` in the same folder already solves this per column (`spend-by-day-chart.tsx:37-39`), so the convention exists and wasn't picked up. Label and value are rendered as adjacent text, so numbers do reach a screen reader — a consistency gap, not information loss. `ModelSpendBars`' caption arrow (`:26`) is the one piece with no text equivalent: "1.2M → 340K" reads as "1.2M 340K".

**Not flagged:** month in `useState` rather than the URL (the `Route` type carries no query params on any view, and every sibling section keeps sub-state local — a batch migration, not this PR's job); inline `style={{}}` in `spend-bar.tsx:25,35` (both runtime-varying — the sanctioned dynamic case); component sizes (largest new file 144 lines, `usage-view.tsx` shrank 197→173); the deep `sandboxes`→`metrics` import (the UI has no module `index.ts` boundaries anywhere; the same file already imports this way from five other modules).

### Verdict
REQUEST_CHANGES — the cross-sandbox placeholder leak misattributes spend figures to the wrong sandbox in a billing surface; everything else is consolidation cleanup.

Verified before posting: `ORDER BY costUsd DESC` backs the bars' `rows[0]` scaling assumption; the shared-cache claim between the nav summary and the section holds (month boundaries truncate to the 1st, so the key is stable within a month); `series-color.ts`'s contrast comments are accurate (first six clear 3:1 against both `#f2f4f8` and `#262626`; `#002d9c` measures 1.34:1 on the dark track).

---

packages/ui/src/modules/metrics/components/model-spend-bars.tsx
```diff
@@ -0,0 +1,42 @@
...
+              <ArrowRight size={12} className="shrink-0" aria-hidden="true" />
+              <span className="sr-only">to</span>
+              {formatTokens(row.outputTokens)}
+            </>
+          }
+        />
+      ))}
+      <p className="flex items-center gap-1 text-xs text-muted-foreground">
+        Bar length is share of the top model · tokens as in
+        <ArrowRight size={12} className="shrink-0" />
```
dam-code-guardian: 🟡 **Warning:** this legend arrow is the same defect the caption arrow nine lines above just had fixed — and it is the one the fix skipped.

Every `@carbon/icons-react` icon routes through `getAttributes`, which ends:

```js
if (iconAttributes["aria-label"] || iconAttributes["aria-labelledby"]) {
  iconAttributes.role = "img";
  ...
} else iconAttributes["aria-hidden"] = true;
```

So an unlabelled Carbon icon is **already** `aria-hidden`. Two consequences. First, the `aria-hidden="true"` added to the caption arrow at `:28` is a no-op — the real fix there was the `sr-only` "to" on the next line, and it is correct. Second, this arrow gets hidden the same way but received no text equivalent, so the legend announces:

> Bar length is share of the top model · tokens as in **out**

Same one-line fix as the caption:

```suggestion
        <ArrowRight size={12} className="shrink-0" aria-hidden="true" />
        <span className="sr-only">to</span>
        out
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `c8c6b863`

## PR #3147: feat(ui): sandbox-level usage metrics on Configure Sandbox
**Author:** kapetr | **Branch:** feat/2774-sandbox-usage-metrics → main | **Changes:** +664 −412 (28 files)

### Summary
One commit since the last review: the cross-sandbox placeholder leak is fixed by keying the section per sandbox, plus the `totalCostUsd` consolidation, two accessibility additions, and the two doc lines.

### Changes since last review
Previous HEAD: `58e566b3` (2026-08-04T13:44:19Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** `keepPreviousData` cross-sandbox spend leak — `key={agent.id}` remounts the observer, and `keepPreviousData` reads only per-observer state, so it structurally cannot reach another agent's cache entry (`sandbox-home-view.tsx:88`)
- ✅ **Fixed:** landing-page index now names both Usage surfaces (`docs/architecture.md:73`)
- ✅ **Fixed:** total-cost reduction collapsed onto one `totalCostUsd` helper at all three sites (`packages/ui/src/modules/metrics/lib/format.ts:5`)
- 🔁 **Still present:** nine lines of stateful orchestration duplicated across both Usage surfaces — and the invariant that keeps the fix above working is now enforced from a third file, the mount site, with nothing typing or testing it (`packages/ui/src/modules/metrics/views/usage-view.tsx:29`)
- 🔁 **Still present:** router-level binding check splits the authorization model across two layers (`packages/api-server-api/src/modules/metrics/router.ts:29`)
- 🆕 **New:** the legend arrow nine lines below the fixed one is still unlabelled (`packages/ui/src/modules/metrics/components/model-spend-bars.tsx:37`)
- 🆕 **New:** `useMetricsOverview` orphaned by the panel deletion (`packages/ui/src/modules/metrics/api/queries.ts:72`)

### Findings
- 🟡 **Warning:** legend arrow left unlabelled while its sibling was fixed (`packages/ui/src/modules/metrics/components/model-spend-bars.tsx:37`)
- 🟡 **Warning:** `useMetricsOverview` is now dead code whose docstring still describes the deleted panel — it is exported, carries `refetchInterval: 15000`, and after `metrics-panel.tsx` (its only consumer) was removed the sole occurrence left in the repo is its own definition. Delete it with the panel it served. Two smaller orphans in the same wake: `formatTime` (`packages/ui/src/lib/format-time.ts:88`) lost its last caller, and `formatDurationMs` is now reached only internally from `durationSegments`, so its `export` no longer carries weight (`packages/ui/src/modules/metrics/api/queries.ts:72`)

**Verdict:** minor drift

**Drift**

❌ *Check 1 — `docs/architecture/metrics.md:7`* — the page's opening Overview sentence still scopes the whole subsystem to a single surface: "It backs the Settings **Usage** tab". The diff adds a second, co-equal surface (`usage` joins `sandboxSectionSchema` in `routes.ts`; `SandboxHomeView` renders the section). This is the same drift class as the two round-1 findings, in the most prominent place on the page, and it now leaves the page internally contradictory: line 52 says "**Both** Usage surfaces treat that error as…" and line 32 describes "the per-sandbox Usage section beside the global one", while line 7 still says there is one tab. Proposed: name both surfaces — the global Settings **Usage** tab and the per-sandbox **Usage** section on Configure Sandbox.

Secondary accuracy note on the same sentence, worth folding into that rewrite rather than left standing: the enumeration credits the tab with **per-session** aggregates, but neither Usage surface renders a per-session rollup — per-session rows are consumed by the CLI and the sessions sidebar. (Predates the PR; the sentence is being rewritten anyway.)

❌ *Check 2 — `docs/architecture.md:3`* — reads `Last verified: 2026-07-29` while the page **is** edited in this diff (line 73). The guideline is "bumped whenever you edit the page". Newly applicable rather than a carried-over miss: at the round-1 SHA the landing page was byte-identical to `main`, so it was not in the diff at all; `c8c6b863` brought it in without touching the header. Proposed: `Last verified: 2026-08-04`. `metrics.md` is clean here — its header already reads `2026-08-04`.

**Possible drift**

None.

**Resolved since round 1**

Both round-1 ❌ items verified fixed against the text on the PR branch (`architecture.md:73` now names both surfaces; `metrics.md:52` is plural). The sentence the fix added to line 52 is **accurate** on both claims: "they withdraw the period control" — both surfaces gate `MonthSwitcher` behind `isMetricsUnavailable(error)`, which matches only `PRECONDITION_FAILED`; "the per-sandbox nav summary degrades to its neutral placeholder" — `retry: false` leaves `data` undefined, the summary returns `undefined`, and the nav renders `—`. Neither addition is a volatile-content leak.

The round-1 ⚠️ **does not stand**, on two independent grounds: (1) its premise is partly incorrect — `runtimeBySession` retains a browser consumer after the deletion (`useSessionCosts`, consumed by `sessions-sidebar.tsx:106`), and that predates the PR; only `contextPerCall` lost its browser consumer. (2) The deleted `metrics-panel.tsx` had no renderer on `main`, so deleting unreachable code changes no subsystem behavior for the page to reflect. The diagram's omission of the CLI as an actor is pre-existing and not anchored to any code change in this diff.

Checks 3–7 clean: no ADR references or `Motivated by:` sections in either edited page; no page moved, renamed, or deleted; no new long-lived component. Both edited pages sit under their caps measured in the gate's own unit (`architecture.md` 5,735 / 8,000 index cap; `metrics.md` 12,004 / 40,000 page cap).

**Prior finding 1 — business logic in the validation layer (`router.ts:29`) — STILL PRESENT, not addressed at all.** The new commit touched no server or contract file; `router.ts` is byte-identical to round 1, and the generalized comment that amplified the pre-existing instance is unchanged. Round 1's severity assessment holds and was re-confirmed: not a security hole — `ownedScope` returns `[]` for an unowned agent, both service methods short-circuit on an empty id list, and `listOwnedAgentIds` is binding-narrowed at the composition root. The cost is duplication, not a gap.

One calibration worth recording, since this finding is now on its second round: `checkAgentBinding` is called from the same position in at least seven other contract routers (`skills` alone has six call sites, plus `schedules`, `files`, `egress-rules`, `connections`). This is an established repo-wide convention, not a local slip, and moving it into `ownedScope` for `metrics` alone would make this module the odd one out — a legitimate reason to defer, and it makes this a repo-wide refactor rather than a PR-blocking item. The amplification remains attributable here because the PR generalized the comment from the specific case to a standing rule.

**Prior finding 2 — total-spend arithmetic duplicated across three sites — FIXED.** All three derivations genuinely collapse onto one helper; the only remaining `reduce` over `costUsd` are the helper itself and the day chart's `Math.max` for axis scaling. No fourth site was missed — the bar components take `rows[0]?.costUsd` as a scale max, and `SpendStatCards` takes already-summed primitives by design.

Residual, smaller than the original: the total is still derived on the client, now in exactly one place. `SpendBreakdown` carries the rollups and no total, so a figure the application layer already has the inputs for is recomputed past the contract boundary. Going from three derivations to one is the substance of the fix; adding a server-side total would close the remainder whenever the contract is next touched.

**On the helper's placement** — `format.ts` is the weakest part of an otherwise clean fix, but not raised as an architecture finding, because no rule in these docs governs a client module's internal layout. On the merits: `totalCostUsd` returns `number` while the other seven exports all return strings, and the new `import type { TokenSpendByModel }` is the visible symptom — the formatting module is now coupled to the metrics contract. Mitigating it, `lib/month-range.ts` already imports a contract type, so this is existing precedent rather than a new direction. A `lib/totals.ts` would keep `format.ts` string-only at no real cost.

**The new `totals()` composition is sound** — verified at runtime rather than by inspection: empty rows yield all five keys at zero (matching the previous single-reduce behaviour on the `?? []` path), two rows sum correctly, the key set is exactly `SpendStatCards`' five props, and the explicit `costUsd` is never overwritten by the spread. The only cost is two passes over a handful of rows.

**New findings introduced by the commit:** none architectural. The server and contract layers are untouched, and `clickhouse-reader.ts` remains correctly oriented — it implements the port and does no scoping of its own.

One non-architecture observation: `totalCostUsd` has no test, though `metrics-format.test.ts` already covers its file's other exports. A two-line case (empty array → 0, two rows → sum) would pin the one piece of arithmetic three surfaces now share.

**Prior finding 1 — CRITICAL cross-sandbox leak — ✅ FIXED** (verified from library source, not from the name). `keepPreviousData` is pure identity, and all "previousness" comes from `#lastQueryWithDefinedData` — per-observer instance state, written at exactly one site under `data !== undefined`. The observer's lifetime is the component instance's (`React.useState(() => new Observer(...))`), so a `key` change gives a fresh observer whose field is `undefined`, the `placeholderData !== undefined` guard fails, and `status` stays `pending`. The memoization short-circuit can't rescue it either — it requires `prevResult?.isPlaceholderData`, also per-instance. There is no cache-level "last data any observer saw" store, so the leak is structurally impossible after the fix.

The `key` is genuinely necessary, not defensive: navigating between sandboxes only sets `agentId` while the view stays `sandbox-home`, and the view is rendered without a key anywhere in the chain — `key={agent.id}` is the only remount boundary. No other path to the same leak: the nav summary's `useAgentMonthSpend` puts `agentId` in the key but sets no `placeholderData`, so on switch it renders its `—` placeholder and holds no `useState`/`useRef`. Swept every `placeholderData` in the UI package — two total, the other unrelated to this chain — and no sibling section seeds state from props. Month-picker reset on switch is the accepted cost and reads as correct: a selected month isn't meaningfully "the same" selection across two sandboxes.

**Prior finding 2 — HIGH duplicated orchestration — ❌ STILL PRESENT**, and the `key` fix made the shape of it worse. No `useMonthlySpend` hook exists. The nine lines are still duplicated verbatim, and both listed symptoms persist: two differently-shaped `UsageSkeleton`s, and the error/empty card is a `Notice` helper in the new file while `usage-view.tsx` still inlines the same markup twice.

The new commit sharpens the cost rather than paying it. The invariant that keeps finding 1 fixed is now enforced at a **third** location — a `key` prop at the mount site — physically remote from both the query that creates the hazard and the consumer that suffers it. Nothing types, lints, or tests the requirement: a second mount without `key` silently reintroduces the CRITICAL, and the only thing standing between the codebase and that regression is a five-line comment. Note the asymmetry the duplication hides: `usage-view.tsx` needs no `key` at all, because it passes no `agentId`, so the rule isn't "always key this" but "key it *iff* the query key carries an identity the mount site can swap" — a distinction with no home in the current shape. A `useMonthlySpend(agentId?)` owning the month state, the settled-month ref, and the placeholder-scoping decision would make the fix impossible to forget at a call site instead of merely documented at one.

**Prior finding 3 — MODERATE missing accessible text — ⚠️ PARTIALLY ADDRESSED.**

`spend-bar.tsx` is correct and covers both call sites, since `role="img"` + `aria-label` sit inside `SpendBar` itself; making the subtree presentational hides nothing that mattered, as the sole child is the decorative fill. Two caveats on *what* it announces. The chosen string is the one piece of information already available — the dollar value is adjacent text, so the row now reads "claude-opus-4, 56% of the largest, $902.03, 1.2M to 340K". The comment's analogy to the day chart is inverted: that chart's `sr-only` carries data existing nowhere else in the tree (the day number is on a sparse axis, the amount only in a `title`), whereas the percentage restates a visual encoding whose value is already spoken. Not wrong, but verbosity rather than information. Separately, `Math.round(pct)` contradicts the 8px floor: a row at 0.3% of the max renders a **visible** bar but announces "0% of the largest" — the visual says present, the label says absent. Floor it, or say "less than 1%".

`model-spend-bars.tsx` works, but for a different reason than the diff suggests. Every Carbon icon routes through `getAttributes`, which sets `aria-hidden` unless an `aria-label`/`aria-labelledby` is present — so the explicit `aria-hidden="true"` added to the caption arrow is a **no-op**, and the entire fix is the `sr-only` "to". That part is correct. The consequence is the finding above: **the legend arrow is still broken in exactly the way the caption was**, announcing "Bar length is share of the top model · tokens as in out". Two `ArrowRight` instances in one file, nine lines apart, one fixed and one not — the fix's own reasoning condemns the line it skipped.

`spend-stat-cards.tsx`, the other file named in the prior finding, received nothing. Being precise about what that costs, because the in/out half turns out **not** to be a defect here: the `/` separator is a real text character, so unlike the arrow it *is* announced, and the same holds for the inline unit spans. What remains is weaker — `Stat` emits label and figure as two unassociated sibling spans with no `dl`/`dt`/`dd` and no `aria-labelledby`, so linear reading pairs them correctly but non-linear navigation loses the pairing. Low severity and arguably out of scope, but the file was named and is new here, so flagged rather than dropped.

**NEW — MODERATE:** the PR orphans `useMetricsOverview` — see Findings above.

Verified clean, for the record: zero dangling imports of either deleted file, so no build break; `CHART_HEIGHT_CLASS` and `formatUsdCell` are **not** orphaned (`session-row.tsx:160` still uses the latter).

### Verdict
COMMENT — the critical leak is genuinely fixed and verified against library source; what remains is one unfixed accessibility sibling, a newly-dead hook, two doc lines, and the two carryover consolidation items.

---

packages/ui/src/modules/metrics/views/usage-view.tsx
```diff
@@ -1,68 +1,50 @@
...
```
PetrBulanek: These ~9 lines (month seed → `monthRange` → timeZone → `useSpendBreakdown` → `useSettledMonth` → the three render gates) are duplicated in `sandbox-usage-section.tsx:59-69`. The pure helpers were extracted; the stateful wiring wasn't.

The risk: the leak-fix now lives at a third spot — `key={agent.id}` at the mount site (`sandbox-home-view.tsx:88`) — with nothing typing or testing it, so a future mount that forgets the key silently shows one sandbox's spend under another's name. A `useMonthlySpend(agentId?)` hook owning the month state and placeholder-scoping collapses both surfaces and makes the fix impossible to forget (also unifies the two `UsageSkeleton`s and the inlined-vs-`Notice` error card).

packages/ui/src/modules/metrics/api/queries.ts
PetrBulanek: Deleting `metrics-panel.tsx` removed this hook's only caller — nothing but its own definition remains, and the docstring still describes the deleted panel. Delete it too.

packages/ui/src/modules/metrics/components/model-spend-bars.tsx
PetrBulanek: The caption arrow (`:28-29`) got an `sr-only "to"`; this identical legend arrow didn't, so it reads "tokens as in **out**". Same fix:

```suggestion
        <ArrowRight size={12} className="shrink-0" aria-hidden="true" />
        <span className="sr-only">to</span>
```

docs/architecture/metrics.md
PetrBulanek: Opening line still says "It backs the Settings **Usage** tab" (singular), but lines 32 and 52 are dual-surface — the page contradicts itself. Name both surfaces here. Also: it credits the tab with **per-session** aggregates, which neither Usage surface renders (those go to the CLI / sessions sidebar) — worth fixing in the same edit.

docs/architecture.md
PetrBulanek, line 1: Line 73 is edited in this PR, so the header should bump:

```suggestion
Last verified: 2026-08-04
```

PetrBulanek (changes requested):

packages/ui/src/modules/metrics/views/usage-view.tsx
kapetr: Done, with one part left out: the two `UsageSkeleton`s stay separate.

`useMonthlySpend` now owns the month state, the query, and the three render gates, so both surfaces are one call. The placeholder is scoped in the query itself — only a same-`agentId` predecessor can fill in — so `key={agent.id}` at the mount site is removed. The key read-through is pure and unit-tested against keys built by the real tRPC client (`metrics-spend-key.test.ts`); if that shape moves, the test fails, and at runtime it fails safe (reads as unnarrowed, so a narrowed caller gets no placeholder instead of a wrong one).

The error/empty card is now one shared `UsageNotice`.

The skeletons are not the same shape by intent: the tab mirrors a headline figure plus two bar stacks, the section mirrors four stat cards plus one chart. One component for both needs a variant flag, and the flag would carry the whole difference — no shared code left. Happy to merge them if you prefer one component.

packages/api-server-api/src/modules/metrics/router.ts
kapetr: Checked this, and I am keeping the check in the router. Two reasons.

**It is the convention, not an exception.** `checkAgentBinding` is called from the router in 8 modules, 16 call sites (`agents`, `schedules`, `connections`, `files`, `skills`, `approvals`, `egress-rules`, `metrics`). Moving it into `ownedScope` would make metrics the only module where key binding is enforced somewhere else, so a reader would then have to check *three* places instead of two.

**The service does not know the binding.** `createMetricsService` takes one auth input: `listOwnedAgentIds`. The binding is applied to that list at the composition root (`app.ts:976-985`), which is why the service alone already denies an unbound agent — with empty rows. For `ownedScope` to raise FORBIDDEN it would need the key binding as a new dep, which pulls request auth context into a service that today only knows an id list.

You are right that the comment generalized a rule while reading like an exception. Fixed in `8dd26a9c` — it now states the split plainly (ownership = service, binding = router) and says the router check upgrades an empty result to FORBIDDEN.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `d843652`

### Summary
Round 3 collapses both Usage surfaces onto one `useMonthlySpend` hook and clears every substantive round-2 item: the cross-sandbox placeholder fix moved off the remote mount-site `key` into a query-layer same-agent gate (now tested), the duplication is gone, and the dead `useMetricsOverview`/`formatTime` are removed.

### Changes since last review
Previous HEAD: `c8c6b863` (2026-08-04T15:04:44Z) — verdict COMMENT

- ✅ **Fixed:** the nine-line stateful duplication collapsed into `hooks/use-monthly-spend.ts`, shared by both surfaces; the cross-sandbox invariant moved off the mount-site `key` into `useSpendBreakdown`'s `placeholderData` same-agent gate — a predecessor from another agent is now structurally withheld whether the section remounts or just swaps its `agentId` prop — and guarded by `metrics-spend-key.test.ts` (`packages/ui/src/modules/metrics/api/queries.ts:45`)
- ✅ **Fixed:** `useMetricsOverview` and `formatTime` removed — no dangling references remain (control-checked) (`packages/ui/src/modules/metrics/api/queries.ts`)
- ✅ **Fixed:** both the caption and legend `ArrowRight` now carry `sr-only "to"`, so the model bars read "in → out" and the legend "tokens as in → out" (`packages/ui/src/modules/metrics/components/model-spend-bars.tsx:37`)
- ✅ **Fixed:** `totalCostUsd` moved to `lib/totals.ts` (number-returning) and `format.ts` is string-only again, with a unit test (`packages/ui/src/modules/metrics/lib/totals.ts:5`)
- ✅ **Fixed:** stat figures now `dt`/`dd`-paired, and `SpendBar` announces "less than 1% of the largest" for a floored-but-visible bar rather than "0%" (`packages/ui/src/modules/metrics/components/spend-bar.tsx:38`)
- ✅ **Fixed:** the doc opening sentence now names both Usage surfaces and the landing-page `Last verified` is bumped (`docs/architecture/metrics.md:7`)
- 🔁 **Still present:** router-level `checkAgentBinding` keeps the auth split (owner-scope in the service, key-binding in the router) — the established repo-wide convention, a deferred cross-cutting refactor rather than this PR's defect (`packages/api-server-api/src/modules/metrics/router.ts:31`)
- 🆕 **New:** both edited architecture pages stamp `Last verified: 2026-08-04`, but commit `7095bde` edited them on 2026-08-05 (`docs/architecture.md:3`)

### Findings
- 🟢 **Suggestion:** bump `Last verified` to the edit date (2026-08-05) on both edited architecture pages — the 08-05 wording tidy left them at 08-04 (`docs/architecture.md:3`)

**Verdict:** aligned

Both round-2 ❌ items are resolved at this HEAD: `metrics.md:7`'s opening sentence now names the two Usage surfaces (and correctly attributes the per-session rollup to the sessions sidebar, not either Usage surface), and `architecture.md:3`'s `Last verified` is bumped. Check 1 is satisfied — every subsystem-behavior change is reflected: the `agentId` narrowing (`metrics.md:39-40`), the summed per-call latency reported as model time (`:48-49`), and both surfaces' fail-closed handling (`:53`). Checks 3–7 produce nothing. `metrics.md` measures 12,243 chars and `architecture.md` 5,735 in the gate's own unit (`content.length`) — both well under cap.

**Possible drift**

⚠️ *Check 2 — `docs/architecture.md:3` / `docs/architecture/metrics.md:3`.* Both pages were content-edited by commit `7095bde` on 2026-08-05 (the opening-sentence rewrite and the date bump itself), yet both read `Last verified: 2026-08-04`. The page content is accurate; this is the one-day judgement call — a stamp for the day the substance was verified (08-04, round 2), or the day the tidy landed (08-05)? Bumping to 08-05 removes the ambiguity.

🔁 Prior finding (`router.ts:31`) — **still present, unchanged.** The `if (input.agentId) checkAgentBinding(...)` split persists; the round-3 comment now formalizes it as the surface-wide convention. Non-blocking exactly as calibrated in round 2: `ownedScope` returns `[]` for an unbound agent so the service alone denies it, and the same pattern sits on `overview` here plus seven other contract routers — a repo-wide refactor, not this PR's blocker.

✅ Prior suggestion (helper placement) — **resolved.** `totalCostUsd` now lives in `lib/totals.ts` returning `number`, `format.ts` is string-only again, and `metrics-format.test.ts` covers it.

No new findings. `spend-key.ts`'s single `as` is a guarded (`Array.isArray`) read-through of a tRPC query key the client types as `unknown`, and `metrics-spend-key.test.ts` pins the key shape it depends on — acceptable at that boundary. Layering otherwise clean: pure helpers in `lib/`, the stateful `useMonthlySpend` in `hooks/`, and `metrics-service.ts` applies the narrowing through `ownedScope`, so scoping stays in the service.

✅ **Prior CRITICAL (cross-sandbox leak) — fixed and hardened.** The fix moved from the round-2 mount-site `key={agent.id}` (removed) into `useSpendBreakdown`'s `placeholderData: (previous, previousQuery) => keyAgentId(previousQuery?.queryKey) === agentId ? previous : undefined`. Traced both switch paths: swapping the `agentId` prop on a persisted observer, or a full remount, both change the key's `agentId`, so a predecessor scoped to another agent fails the equality and the placeholder is withheld — a skeleton, never agent A's figures under B. Same-agent month paging still fills in. It is the only `placeholderData`/`keepPreviousData` site in the module, and the invariant now lives where the hazard is created and is covered by a test — closing round-2's "enforced from a remote third file with nothing testing it".

✅ **Prior HIGH (nine-line duplication) — fixed.** `useMonthlySpend` owns the month state, query, and all render gates for both surfaces; `MonthSwitcher`, `UsageNotice`/`UsageStaleLabel`/`readFailureMessage` are shared. The two `UsageSkeleton`s that remain are legitimately distinct (the global tab renders total + by-model + by-agent; the sandbox section renders four stat cards + a chart), not the copied wiring.

✅ **Prior finding (legend arrow) — fixed;** both arrows carry `sr-only "to"`. The explicit `aria-hidden` is a harmless no-op (Carbon's `getAttributes` already sets it on an unlabelled icon); the real fix is the `sr-only`, and it is present in both the caption and the legend.

✅ **`SpendStatCards` — improved:** figures are now `dt`/`dd`-paired inside a `dl`, so jumping to a number keeps its label (round-2's unassociated-spans note). `SpendBar`'s `aria-label` special-cases the sub-1% floored bar.

No new findings. `useSettledMonth` correctly latches on `!isPlaceholderData && data !== undefined`, so an errored month keeps the last-good month for the date-keyed chart rather than mismatching. `series-color.ts`'s dark-track contrast loss past six rows is documented and accepted in-code.

### Verdict
APPROVE — every substantive item from three rounds is resolved and verified; what remains is the deferred repo-wide auth-split convention and a one-day `Last verified` stamp, neither blocking.


---

packages/ui/src/modules/metrics/components/sandbox-usage-section.tsx
```diff
@@ -0,0 +1,132 @@
...
+      <p className="mb-4 text-sm text-muted-foreground">
+        LLM spend for this sandbox, including work it delegated to other agents.
+      </p>
+
+      {/* Only when there is nothing to show: a failed refetch keeps the loaded
+          month, and the label by the period control names it as not fresh. */}
+      {isError && !data && (
+        <UsageNotice>{readFailureMessage(unavailable, label)}</UsageNotice>
+      )}
+      {isPending && !isError && <UsageSkeleton />}
```
PetrBulanek: This infers "metrics disabled" from `isError`, and that breaks now that the mount-site key is gone.

`useSpendBreakdown` latches `metricsDisabled` (→ `enabled: false`) in hook-instance state, which without `key={agent.id}` now survives an `agentId` change. After the latch sets, switching sandboxes yields a fresh key that never fetches — and a disabled observer on an unfetched key reports `status: "pending"`, `isError: false`, `data: undefined` (checked against query-core 5.99.0). So `isError && !data` misses, `isPending && !isError` hits, and the section sticks on `UsageSkeleton` forever. On the current month the shared `useAgentMonthSpend` key eventually caches an error and rescues it (a skeleton flash); other months have no such caller and hang.

`usage-view.tsx:54` avoids this by early-returning on `unavailable` instead of deriving it. Same fix here:

```suggestion
      {(isError || unavailable) && !data && (
        <UsageNotice>{readFailureMessage(unavailable, label)}</UsageNotice>
      )}
      {isPending && !isError && !unavailable && <UsageSkeleton />}
```

Better: have `useMonthlySpend` return one resolved `view: "unavailable" | "failed" | "loading" | "ready"` so the two surfaces stop writing their own gates — the drift between them is this bug.

PetrBulanek (approved):

PR #3149, opened 2026-08-04
fix(ui): artifact panel closes when switching sandboxes

### LINKED ISSUE #3141: UI - Artifact panel stays open when switching sandboxes
## Context

Reported from Slack (video in thread): [source discussion](https://ibm-research.slack.com/archives/C0B8DA70GE4/p1785782772604799). Surfaced when opening a freshly created Knowledge base sandbox in the chat view.

The artifact panel remembers whether an artifact was open in chat and keeps that panel open when the user moves to a different sandbox — including a brand-new one that has no artifacts of its own. What looked at first like an artifact leaking across sandbox boundaries is really the chat UI persisting the open-artifact panel state across sandbox switches.

## Problem

When a user has an artifact open in chat and then opens a different sandbox, the previously-opened artifact stays visible in the panel. In a newly created sandbox this is especially confusing — the sandbox is empty, but an artifact from an earlier sandbox appears to belong to it. The open-artifact state shouldn't carry over from one sandbox to another.

## Expected Behavior

Opening a sandbox — any sandbox — starts with the artifact panel closed. Nothing from a previously-viewed sandbox persists in the panel.

## Actual Behavior

If an artifact was open in chat, the artifact panel stays open when switching to a different sandbox, showing the earlier sandbox's artifact in the new context.

## Steps to Reproduce

1. Open a sandbox that has at least one artifact, and open that artifact in chat.
2. Switch to (or create) a different sandbox.
3. Open the new sandbox in chat.
4. Observe: the artifact panel is still open, showing the artifact from the previous sandbox.

PetrBulanek:
Opening a sandbox's chat now starts with the right-hand dock closed. Previously, an artifact
opened while viewing one sandbox stayed docked when the user moved to another — in a
freshly created, empty sandbox this read as if the earlier sandbox's artifact belonged to
the new one.

The docked artifact preview is chat-context state, with the same lifetime as the open file
and the message list, so it is now cleared by the same chat teardown every route into and
out of chat already goes through. The teardown also stopped leaking two stale file-viewer
flags, which had been causing a spurious "Discard unsaved changes?" prompt on the first
file opened in the next sandbox.

Verified against the local cluster by building both sides: on the pre-fix bundle the dock
wrongly showed the previous sandbox's artifact while the new sandbox's own sidebar read
"No artifacts yet"; on the fixed bundle it is closed. Browser back/forward re-entry is
covered too.

Closes #3141

## Sub-issues

- [x] 01 — Reset the docked artifact panel on chat teardown


packages/ui/src/modules/sessions/store/sessions.ts
```diff
@@ -103,6 +103,9 @@ export const createSessionsSlice: StateCreator<
       sessionError: null,
       terminalPaused: false,
       openFilePath: null,
+      openFileDirty: false,
```
dam-code-guardian, line 106: 🟡 **Warning:** Clearing these two here fixes the teardown path, but the mirror-image sibling still leaks `openFileDirty`. The dock is mutually exclusive, and both directions of the swap are handled in the two slices' setters — except asymmetrically:

- `files/store.ts:69-70` — `setOpenFilePath` clears `openFileDirty`/`openFileEdit` **and** `openArtifactId`. Complete.
- `artifacts/store.ts:24` — `setOpenArtifactId(id)` clears `openFilePath` but **not** `openFileDirty`.

So opening an artifact while the file viewer holds an unsaved edit unmounts the viewer with `openFileDirty` still `true`, and `use-file-tree.ts:39` gates its prompt on that flag alone — the next file opened in the *same* sandbox gets a spurious "Discard unsaved changes?" for an edit whose viewer is long gone. That is the same defect class as the two flags this PR just added to the teardown, one path over; the fix here is why it now only reproduces within a sandbox instead of across them.

```suggestion
      id
        ? { openArtifactId: id, openFilePath: null, openFileDirty: false, openFileEdit: false }
        : { openArtifactId: id },
```

(That block lives in `artifacts/store.ts`, not this file — anchored here because it is the line whose reasoning it follows.)

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `8945bb43`

## PR #3149: fix(ui): artifact panel closes when switching sandboxes
**Author:** PetrBulanek | **Branch:** fix/3141-artifact-panel-sandbox-switch → main | **Changes:** +11 −5 (2 files)

### Summary
Adds `openArtifactId`, `openFileDirty` and `openFileEdit` to `resetChatContext()`'s reset payload so the docked artifact preview and two file-viewer flags no longer survive a sandbox switch, and documents the preview's lifetime as chat context on the artifact-library page.

### Findings
- 🟡 **Warning:** `setOpenArtifactId` leaves `openFileDirty` set when it closes the file viewer (`packages/ui/src/modules/artifacts/store.ts:24`)
- ✅ **Looks good:** The fix reaches every entry point rather than just the reported one — all six `resetChatContext` callers (`selectAgent`, `openKnowledgeBase`, `openAgentSession`, `openAgentTerminal`, `goBack`, plus the popstate handler and `deleteSession`) inherit it, so browser back/forward re-entry is covered by construction, not by a second patch.
- ✅ **Looks good:** `dockedExperiment`, the third condition gating the same dock (`chat-view.tsx:840`), correctly needs no reset — `useDockedExperiment(selectedAgent)` re-derives it per sandbox rather than storing it.

aligned

✅ No findings.

### Verdict
COMMENT — The teardown fix is correct and complete for the reported path; the one issue is the mirror-image sibling, where opening an artifact closes a dirty file viewer without clearing the dirty flag, leaving exactly the stale-prompt state this PR removes from the teardown path.

---

PetrBulanek:
Superseded by #3152, which was merged on 2026-08-05 and closed #3141 with the same fix (identical change to `resetChatContext` in `sessions.ts`). Closing this as redundant.

PR #3150, opened 2026-08-04
feat(agents): port skydiscover run conventions to shinkaevolve, gepa, openevolve

xjacka:
Ports the two run conventions proven on the skydiscover workload (#3109, live-tested end to end) to the three older workload images, uniformly.

## Runs move into the work dir

`*_OUTPUT_ROOT` becomes `~/work/<name>-runs` — the work dir is where the UI file browser and the terminal land, so run artifacts stop being invisible-by-default. Two guards keep the platform's repo seed working:

- The runs folder is **created lazily at first run**, never image-baked — the first-boot seed would make the work dir non-empty, and the workspace seed (which clones into the work-dir root and refuses a non-empty dir) would fail.
- When the work dir **is** a seeded repo, the agent appends `<name>-runs/` to `.git/info/exclude` (local ignore) so the checkout stays clean. The run's own target clone lives inside the run directory, so outputs never touch it either way.

## Searches ride the background-work contract

Launch discipline changes from detached `nohup … &` to a **harness background task** (Bash `run_in_background`, with an explicit never-foreground warning — a foreground tool call's timeout SIGTERMs the whole process group). The background-work contract (#2965/#3044) then holds the pod awake for the run's lifetime and wakes the agent on completion to report the result proactively — no more silent completions discovered on the next prompt, and no more hibernation killing a run mid-flight. Resume-on-wake and the keep-awake terminal remain documented as fallbacks for pod restarts and legacy-launched runs.

## Verification

- All three images build (`agents:{shinkaevolve,gepa,openevolve}:image`).
- Both conventions were validated live on the skydiscover workload in a dev cluster: pod held awake by a reported task through a full search with the session closed, proactive completion report delivered, lazy work-dir creation confirmed, and the empty-vs-seeded work-dir branch exercised.

## Out of scope

- `nous` (never-hibernate campaign model, its own lifecycle).
- A UI indicator for reported background work (the runtime already publishes it on `/api/status`; the UI doesn't read it yet) — candidate follow-up.

### DIFF
--- a/packages/agents/gepa/AGENTS.md
+++ b/packages/agents/gepa/AGENTS.md
@@ -183,7 +183,9 @@ the session goes idle. The run directory lives on persistent `$HOME`, so the
 run is recoverable but **does not progress while the pod is down**.
 
 So at the **start of each turn**, check any run you care about: if its
-`run.pid` is dead and it hasn't exhausted its budget, reinstall any extra deps
+`run.pid` no longer names the live run (the same cmdline-validated check as the
+start-of-conversation scan — a recycled PID after a pod restart must not
+skip the resume) and it hasn't exhausted its budget, reinstall any extra deps
 (the venv reset) and relaunch the **identical** driver — same `run_dir`, same
 `max_metric_calls`. GEPA detects the existing state in `run_dir`, restores it,
 and continues only up to the original budget — the budget is a **total**, not
--- a/packages/agents/gepa/AGENTS.md
+++ b/packages/agents/gepa/AGENTS.md
@@ -57,7 +57,8 @@ run. This file is the *how-to-operate-in-this-pod* layer.
 ## Starting a conversation
 
 At the **start of every new conversation**, before anything else, enumerate
-existing runs and offer to act on them. Scan `$GEPA_OUTPUT_ROOT` (`~/gepa-runs`)
+existing runs and offer to act on them. Scan `$GEPA_OUTPUT_ROOT`
+(`~/work/gepa-runs`)
 for run directories (each has a `driver.py` and a `run_dir/`) and classify each
 as:
 
@@ -106,7 +107,8 @@ estimate, but an informed user may pre-authorize it (see below).
 4. **You've presented a budget/cost estimate.** State the `max_metric_calls`
    you'll set, the resulting rough LLM-call count (metric calls for the task
    model, plus roughly one reflection call per iteration for the strong
-   model), that it runs autonomously, and the keep-awake tradeoff (below).
+   model), that it runs autonomously, and that the running work holds the
+   pod awake (no scale-to-zero) until it finishes.
    Then **wait for an explicit go-ahead — unless the user already
    pre-authorized this run** ("just launch it," "don't ask"):
    pre-authorization waives the *wait*, never the estimate — show the numbers,
@@ -115,14 +117,32 @@ estimate, but an informed user may pre-authorize it (see below).
 
 ## Run discipline
 
-- **Launch backgrounded** so you stay conversational and can poll while it
-  runs. Keep the PID and log in the run directory:
+- **First run in this pod:** create the runs root lazily —
+  `mkdir -p "$GEPA_OUTPUT_ROOT"`. It is deliberately never baked into the
+  image: a pre-seeded folder would make the work dir non-empty and block the
+  platform's repo seed (which clones into the work-dir root and refuses a
+  non-empty dir). If `~/work` is a seeded git repo, also append
+  `gepa-runs/` to `.git/info/exclude` (a local ignore — never touch tracked
+  files) so the checkout stays clean; any target repo the run needs is
+  cloned inside the run directory, so outputs never touch it either way.
+- **Launch as a harness background task** (your backgrounded-Bash facility),
+  never a bare detached `nohup … &`. The platform's background-work contract
+  reports harness-registered tasks to the runtime: the pod is held awake for
+  as long as the run lives, and the finishing task wakes you for a follow-up
+  turn — **report the result to the user then** (the best candidate and the
+  `run_dir/` path), don't wait to be asked. A detached `nohup` process is
+  invisible to that contract, so the pod can hibernate mid-run. Still keep
+  the PID and log in the run directory for monitoring and crash recovery:
 
   ```sh
+  # run this script AS a backgrounded harness task (your Bash tool's
+  # run_in_background) — NEVER as a foreground command: the tool's timeout
+  # would SIGTERM the whole process group, run included, mid-flight
   dir="$GEPA_OUTPUT_ROOT/<run-id>"
   cd "$dir"
-  nohup python driver.py > run.log 2>&1 &
+  python driver.py > run.log 2>&1 &
   echo $! > run.pid
+  wait
   ```
 
 - **Always pass `run_dir`** in the driver, pointing inside the run's directory
@@ -155,10 +175,12 @@ but the uv cache is on persistent `$HOME`, so reinstall extras after a restart
 
 ## Surviving hibernation (resume-on-wake)
 
-The pod **scales to zero when the session goes idle** — no active turn, no
-queued prompt, no open terminal/SSH session. That kills any background driver.
-The run directory lives on persistent `$HOME`, so the run is recoverable but
-**does not progress while you're not engaged**.
+With the launch discipline above, a running optimization **holds the pod
+awake** (reported background work) and hibernation mid-run is the exception,
+not the rule. It can still happen — a pod restart or eviction, a crash, or a
+run launched the legacy detached way — and then the pod scales to zero once
+the session goes idle. The run directory lives on persistent `$HOME`, so the
+run is recoverable but **does not progress while the pod is down**.
 
 So at the **start of each turn**, check any run you care about: if its
 `run.pid` is dead and it hasn't exhausted its budget, reinstall any extra deps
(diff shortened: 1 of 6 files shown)


packages/agents/gepa/AGENTS.md
```diff
@@ -115,14 +117,32 @@ estimate, but an informed user may pre-authorize it (see below).
...
   ```sh
+  # run this script AS a backgrounded harness task (your Bash tool's
+  # run_in_background) — NEVER as a foreground command: the tool's timeout
+  # would SIGTERM the whole process group, run included, mid-flight
   dir="$GEPA_OUTPUT_ROOT/<run-id>"
   cd "$dir"
-  nohup python driver.py > run.log 2>&1 &
+  python driver.py > run.log 2>&1 &
   echo $! > run.pid
+  wait
```
dam-code-guardian: 🟡 **Warning:** Bare `wait` returns 0 regardless of how the child exited, so the harness task reports success even when `driver.py` crashes on startup — and the whole point of this change is that the finishing task wakes you to report the result. Verified in a script with `set -uo pipefail`: `/bin/false &` then bare `wait` → status 0, while `wait "$p"` → status 1.

Since the snippet already captures the PID for `run.pid`, waiting on it explicitly costs nothing and makes the failure visible:

```suggestion
  python driver.py > run.log 2>&1 &
  echo $! > run.pid
  wait "$(cat run.pid)"
```

Same in `shinkaevolve/AGENTS.md:119` and `openevolve/AGENTS.md:116`.

packages/agents/shinkaevolve/AGENTS.md
```diff
@@ -43,7 +43,7 @@ Consult it whenever you set up a run. This file is the
 
 At the **start of every new conversation**, before anything else, enumerate
 existing runs and offer to act on them. Scan `$SHINKA_OUTPUT_ROOT`
-(`~/shinka-runs`) for run directories (each has a `task/` and a `results/`
+(`~/work/shinka-runs`) for run directories (each has a `task/` and a `results/`
 dir) and classify each as:
 
 - **running** — its `run.pid` names a live process (`kill -0 "$(cat run.pid)"`).
```
dam-code-guardian: 🟢 **Suggestion:** This PR ports skydiscover's conventions, but not its liveness check. `gepa/AGENTS.md:65-75` (and skydiscover itself) validate `/proc/<pid>/cmdline` against the expected program because a bare `kill -0` misreads a recycled PID after a pod restart — exactly the restart scenario the reworked hibernation section now leans on as the residual failure mode. Here and in `openevolve/AGENTS.md:48` a stale `run.pid` can still classify a dead run as running, which sends the agent down the wrong branch of its start-of-conversation triage instead of resuming.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `7686de1e`

## PR #3150: feat(agents): port skydiscover run conventions to shinkaevolve, gepa, openevolve
**Author:** xjacka | **Branch:** feat/workload-run-conventions → main | **Changes:** +142 −68 (6 files)

### Summary
Ports two conventions proven on the skydiscover workload to the three older evolution images: `*_OUTPUT_ROOT` moves from `~/<name>-runs` to `~/work/<name>-runs` (with the image-baked `mkdir`+`chown` dropped in favour of a lazy agent-side `mkdir`), and launch discipline changes from detached `nohup … &` to a harness background task that rides the background-work contract.

### Findings
- 🟡 **Warning:** The launch snippet's bare `wait` masks the run's exit status (`packages/agents/gepa/AGENTS.md:145`, `shinkaevolve/AGENTS.md:119`, `openevolve/AGENTS.md:116`)
- 🟢 **Suggestion:** Two of the three images keep the weaker `kill -0` liveness check the skydiscover precedent deliberately replaced (`packages/agents/shinkaevolve/AGENTS.md:49`, `packages/agents/openevolve/AGENTS.md:48`)
- ✅ **Looks good:** The lazy-`mkdir` rationale checks out against the actual code, not just the sibling's wording — `workspace-seed-plugin.ts:42` refuses a work dir with any entry, and the seed event is enqueued at create time (`agents-service.ts:893`), so an image-baked `~/work/<name>-runs` genuinely would have broken repo-seeded agents.
- ✅ **Looks good:** Dropping the `chown -R 65532:0` is safe rather than an oversight: the chart's init container runs as `runAsUser: 65532` (`values.yaml:868`) and already does `mkdir -p $HOME/work` itself (`values.yaml:977`), so the agent's own lazy `mkdir` inherits the right ownership with no image-time step.

aligned

### Verdict
COMMENT — The port is faithful and its central premise (why the runs dir must not be image-baked) verifies against the real seed guard; the one substantive issue is the bare `wait` silently discarding the run's exit status in all three snippets.

---

packages/agents/shinkaevolve/AGENTS.md
```diff
@@ -126,10 +157,12 @@ uv cache is on persistent `$HOME`, so reinstall after a restart — it's fast.
...
-recoverable but **does not progress while you're not engaged**.
+With the launch discipline above, a running evolution **holds the pod awake**
+(reported background work) and hibernation mid-run is the exception, not the
+rule. It can still happen — a pod restart or eviction, a crash, or a run
+launched the legacy detached way — and then the pod scales to zero once the
+session goes idle. The results dir lives on persistent `$HOME`, so the run is
+recoverable but **does not progress while the pod is down**.
 
 So at the **start of each turn**, check any run you care about: if its
 `run.pid` is dead and it hasn't reached its budget, reinstall the run's deps
```
dam-code-guardian: 🟡 **Warning:** This commit replaced bare `kill -0` at the start-of-conversation scan (line 52) because a pod restart resets the PID namespace and a stale `run.pid` can name an unrelated live process — but resume-on-wake, which runs *at the start of each turn* and is the section the reworked hibernation text now leans on, still tests only whether `run.pid` "is dead". A recycled PID reads as alive here, so the run is never relaunched: the exact silent-skip-the-resume failure the line-52 rationale describes, in the one place it actually costs a run.

Same in `openevolve/AGENTS.md:161` and `gepa/AGENTS.md:186` — gepa had the cmdline check before this PR, so its resume section has carried the inconsistency all along.

Cheapest fix is to point it at the check that now exists rather than restating it:

```suggestion
`run.pid` is dead by the cmdline check above and it hasn't reached its budget, reinstall the run's deps
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `89e6d4bd`

## PR #3150: feat(agents): port skydiscover run conventions to shinkaevolve, gepa, openevolve
**Author:** xjacka | **Branch:** feat/workload-run-conventions → main | **Changes:** +169 −73 (6 files)

### Summary
Re-review at `89e6d4bd`: both prior findings fixed, the 🟢 more broadly than suggested — gepa's `/proc/<pid>/cmdline` validation was ported to both remaining images rather than just acknowledged.

### Changes since last review
Previous HEAD: `7686de1e` (2026-08-04T09:45:42Z) — verdict COMMENT

- ✅ **Fixed:** Bare `wait` now waits on the captured PID in all three snippets — verified in a real script file: `/bin/false &` + `wait "$pid"` → status 1, bare `wait` → status 0 (`gepa/AGENTS.md:145`, `shinkaevolve/AGENTS.md:130`, `openevolve/AGENTS.md:127`)
- ✅ **Fixed:** `kill -0` liveness upgraded to cmdline validation in shinkaevolve and openevolve, each grepping the program its own launch block actually invokes — `shinka_run`, `openevolve-run` (`shinkaevolve/AGENTS.md:52`, `openevolve/AGENTS.md:52`)
- 🆕 **New:** The resume-on-wake section still says "if its `run.pid` is dead", the bare-`kill -0` phrasing the same commit just rejected upstream (`shinkaevolve/AGENTS.md:168`, `openevolve/AGENTS.md:161`, `gepa/AGENTS.md:186`)

### Findings
- 🟡 **Warning:** Resume-on-wake's liveness test contradicts the new start-of-conversation one (`packages/agents/shinkaevolve/AGENTS.md:168`)

aligned

### Verdict
COMMENT — Both findings are genuinely fixed and the cmdline patterns match each image's real launch command; the one new issue is that resume-on-wake still describes the weaker liveness check the same commit replaced everywhere else, which is where a recycled PID would silently skip a resume.

---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `0cbfc462`

## PR #3150: feat(agents): port skydiscover run conventions to shinkaevolve, gepa, openevolve
**Author:** xjacka | **Branch:** feat/workload-run-conventions → main | **Changes:** +178 −76 (6 files)

### Summary
Re-review at `0cbfc462`: the one prior finding is fixed in all three images, and a whole-file sweep for other statements of the same liveness mechanism finds none left.

### Changes since last review
Previous HEAD: `89e6d4bd` (2026-08-04T11:10:23Z) — verdict COMMENT

- ✅ **Fixed:** Resume-on-wake now defers to the cmdline-validated check instead of "`run.pid` is dead", in all three images, naming recycled-PID-after-restart as the reason (`shinkaevolve/AGENTS.md:168`, `openevolve/AGENTS.md:161`, `gepa/AGENTS.md:186`)

### Findings

aligned

### Verdict
APPROVE — The delta is exactly the fix, applied uniformly, and the file now states the liveness rule once with the resume path pointing back at it; grepping each file end-to-end for `run.pid`/`kill -0`/liveness phrasing turns up no third statement left contradicting it.

---

PetrBulanek (approved):

PR #3152, opened 2026-08-04
fix(ui): close the chat dock when switching sandboxes

### LINKED ISSUE #3141: UI - Artifact panel stays open when switching sandboxes

Tomas2D:
## Summary

`resetChatContext` already cleared `openFilePath` but not `openArtifactId`, so an artifact opened in one sandbox stayed docked after navigating to another — including a brand-new sandbox with no artifacts of its own. That made it look like an artifact was leaking across sandbox boundaries; it was the panel state persisting.

The fix goes in the single chokepoint every sandbox entry/exit path already funnels through (`selectAgent`, `openKnowledgeBase`, `openAgentSession`, `openAgentTerminal`, `goBack`, and the popstate handler), so no new effect and no per-consumer guards.

- Clear `openArtifactId` in `resetChatContext`.
- Clear the `openFileDirty` / `openFileEdit` residue alongside it, matching what `setOpenFilePath(null)` does. `resetChatContext` set `openFilePath` directly and bypassed that, so a dirty file left open in one sandbox would prompt "Discard unsaved changes?" on the next sandbox's first file click, with no file open.

Two things deliberately left alone:

- **The docked experiment panel** is derived per-agent via `useDockedExperiment(selectedAgent)`, not store state, and its manual override is scoped by `{agentId, sessionId}` — it cannot leak, so a reset there would be dead code.
- **`artifactsSectionOpen`** is the left sidebar section's collapse state, not the dock. It is the same lineage as `filesSectionOpen`, which the reset preserves on purpose as a user preference; resetting it would collapse the sidebar on every navigation.

Closes #3141

## Test plan

- [ ] Manual: open an artifact in chat in sandbox A, switch to sandbox B, confirm the dock starts closed
- [ ] Manual: leave an unsaved file open in sandbox A, switch to sandbox B, confirm the first file click does not prompt to discard changes

No automated regression test: UI vitest runs `environment: "node"` and the store's navigation slice touches `window.location` / `sessionStorage` / `history` at module init, so covering this needs a jsdom harness added to the UI test setup first.

### DIFF
--- a/packages/ui/src/modules/sessions/store/sessions.ts
+++ b/packages/ui/src/modules/sessions/store/sessions.ts
@@ -52,9 +52,10 @@ export interface SessionsSlice {
 
   /**
    * Wipe all per-chat-session state (active session, messages, file tree,
-   * session config, queued prompt). Callers like `selectInstance`,
-   * `goBack`, and the popstate handler invoke this so every entry point
-   * leaves chat state in the same clean shape.
+   * docked file/artifact panel, session config, queued prompt). Callers like
+   * `selectInstance`, `goBack`, and the popstate handler invoke this so every
+   * entry point leaves chat state in the same clean shape — nothing from a
+   * previously-viewed sandbox survives into the next one.
    */
   resetChatContext: () => void;
 }
@@ -102,7 +103,17 @@ export const createSessionsSlice: StateCreator<
       messages: [],
       sessionError: null,
       terminalPaused: false,
+      // The right dock is scoped to the sandbox that opened it — both of its
+      // occupants clear here so entering any sandbox starts with the dock
+      // closed instead of showing the previous sandbox's file or artifact.
+      // (The third occupant, the experiment panel, is derived per agent and
+      // needs no reset.) The viewer's edit/dirty flags ride along, mirroring
+      // `setOpenFilePath(null)`: leaving a dirty file behind would otherwise
+      // prompt "Discard unsaved changes?" on the next sandbox's first click.
       openFilePath: null,
+      openArtifactId: null,
+      openFileDirty: false,
+      openFileEdit: false,
       pendingPermissions: [],
       queuedMessage: null,
       pendingResumeSessionId: null,


packages/ui/src/modules/sessions/store/sessions.ts
```diff
@@ -52,9 +52,10 @@ export interface SessionsSlice {
 
   /**
    * Wipe all per-chat-session state (active session, messages, file tree,
-   * session config, queued prompt). Callers like `selectInstance`,
-   * `goBack`, and the popstate handler invoke this so every entry point
-   * leaves chat state in the same clean shape.
+   * docked file/artifact panel, session config, queued prompt). Callers like
+   * `selectInstance`, `goBack`, and the popstate handler invoke this so every
```
dam-code-guardian, line 56: 🟢 **Suggestion:** `selectInstance` no longer exists anywhere in the tree — this docstring is its only remaining mention, so the list points readers (and agents) at a symbol they cannot find. The real callers are `selectAgent`, `openKnowledgeBase`, `openAgentSession`, `openAgentTerminal`, `goBack`, `deleteSession`, and the popstate handler. Cheap to correct while the line is already in the diff.

```suggestion
   * `selectAgent`, `goBack`, and the popstate handler invoke this so every
```

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `dd3af345`

## PR #3152: fix(ui): close the chat dock when switching sandboxes
**Author:** Tomas2D | **Branch:** fix/artifact-panel-sandbox-switch → main | **Changes:** +14 −3 (1 files)

### Summary
`resetChatContext` cleared `openFilePath` but not `openArtifactId`, so an artifact docked in one sandbox stayed open after navigating to another. The fix clears it — plus the `openFileDirty` / `openFileEdit` residue — at the single chokepoint every navigation path already funnels through.

### Findings
- 🟢 **Suggestion:** stale caller name in the touched docstring (`packages/ui/src/modules/sessions/store/sessions.ts:56`)
- ✅ **Looks good:** the reset field set is complete and each deliberate omission checks out. `openArtifactId` and `openFilePath` are the store's only two dock occupants — the third, the experiment panel, is genuinely derived, and `useDockedExperiment` scopes its manual override by `{agentId, sessionId}`, so it self-invalidates on navigation. The added `openFileDirty: false` / `openFileEdit: false` match exactly what `setOpenFilePath(null)` writes, which is the mirror the new comment claims. And `artifactsSectionOpen` is correctly preserved: it is the sidebar section's collapse preference, same lineage as `filesSectionOpen`, not dock state.

✅ No findings.

The two architecture pages describing the dock — `docs/architecture/artifact-library.md:149` ("a **docked preview** beside the conversation") and `docs/architecture/experiments.md:94` (the panel that "docks itself in the driver's chat") — describe the dock's purpose and contents, not its navigation-teardown behavior, so neither is contradicted by this change. Clearing panel state on sandbox entry is a bug fix within already-documented responsibility, not a change to a subsystem's behavior or responsibility, so the drift rule does not fire. No new subsystem, no architecture page moved or edited, no ADR status change.

✅ No findings.

Diff-scoped review against the TSEng architecture docs: the change is four field initializers and a comment inside one existing zustand slice's state setter, in a client package. It introduces no new module, no cross-module import, no layer crossing, and no contract or router surface — the rules in `modules.md` / `module-boundaries.md` / `slice-composition.md` govern server-package bounded contexts and have nothing to bind to here. Worth noting the change is *aligned* with the direction those rules point: the reset writes fields owned by two sibling slices (`files`, `artifacts`) through the shared `PlatformStore` type rather than reaching into their modules, which is the same mechanism `setOpenFilePath` already uses to clear `openArtifactId`.

### Verdict
APPROVE — a correct one-chokepoint fix whose omissions I verified individually; the only finding is a stale symbol name in a docstring the diff already touches.

---

PR #3153, opened 2026-08-04
fix(ui): close artifact share dialog and toast on save

### LINKED ISSUE #3089: UI - Click in save (Artifacts Sharing)
Hello

When I open the modal for the artifacts settings, make the desired configurations, and click save, the button only changes to "saving" and it feels like it didn't save. I think the **most appropriate thing would be for the modal to close when the user presses** save and show a notification that the artifact configuration was registered or updated.

Image

Tomas2D:
## Summary

Saving sharing settings only flipped the button back from "Saving…", which read as a no-op. The old `onSuccess` closed the dialog only when the save *removed* the share URL (`if (!updated.shareUrl) onClose()`), so enabling a public link — or merely changing the expiry on an already-public artifact — left the modal open with nothing on screen changed at all.

- Close the dialog on a successful save and confirm with a success toast, worded for the outcome ("the public link is live" vs. "the artifact is now private").
- Carry a **Copy link** action on the toast when the save produced a public URL, since closing the dialog takes the URL field away exactly when the user has just created the link. The link also stays reachable from the row's existing "Copy share link".
- Read the share URL from the `artifact` prop instead of mirroring it in local state — the mirror existed only to display the post-save URL while the dialog stayed open, and the dialog no longer outlives the save.

Reuses `emitToast` from `@/lib/toast`, the app-wide emit path; closest precedent is the same success-toast-then-`onClose()` shape in `connection-catalog-modal.tsx`.

The failure path needed no code. `useSetArtifactSharing` already carries `meta: { errorToast: "Failed to update sharing" }` and the QueryClient's central `mutations.onError` renders it with the server detail appended. Since closing happens only in `onSuccess`, a failed save already keeps the dialog open with its state intact so the user can retry without re-entering anything.

Closes #3089

## Test plan

- [ ] Manual: enable a public link and save — dialog closes, toast confirms, "Copy link" copies the URL
- [ ] Manual: change only the expiry on an already-public artifact — dialog closes and the toast confirms (previously the most confusing case)
- [ ] Manual: make sharing private and save — dialog closes, toast says the artifact is now private
- [ ] Manual: force a failing save — dialog stays open, error toast names the reason

No automated test: `ShareDialog` has no existing unit or e2e coverage, and UI vitest currently runs `environment: "node"` with pure-logic suites only, so a dialog-level test means introducing a render harness first.

### DIFF
--- a/packages/ui/src/modules/artifacts/components/share-dialog.tsx
+++ b/packages/ui/src/modules/artifacts/components/share-dialog.tsx
@@ -35,7 +35,15 @@ interface Props {
  *  which closes the dialog and confirms through a toast — a save that only
  *  flipped the button back to "Save" read as a no-op. A failure keeps the
  *  dialog open and surfaces through the mutation's `errorToast`, so the user
- *  can retry without re-entering anything. */
+ *  can retry without re-entering anything.
+ *
+ *  Both exits are gated while the save is in flight. The confirmation lives in
+ *  a `mutate`-scoped `onSuccess`, which React Query skips once the observer has
+ *  no listeners — leaving on a closed dialog would land the write with nothing
+ *  to show for it, the very no-op this dialog was fixed to stop reading as.
+ *  Gating is the narrow fix: a hook-level `onSuccess` would fire after unmount,
+ *  but `defaultMutationOptions` shallow-merges, so it would silently replace
+ *  the client-wide `onSuccess` that applies `meta.invalidates`. */
 export function ShareDialog({ artifact, onClose }: Props) {
   const [isPublic, setIsPublic] = useState(artifact.visibility === "public");
   const [expiry, setExpiry] = useState<string>(
@@ -65,7 +73,28 @@ export function ShareDialog({ artifact, onClose }: Props) {
                   message: "Sharing updated — the public link is live.",
                   action: {
                     label: "Copy link",
-                    onClick: () => void copy(savedUrl),
+                    // Runs after this dialog has unmounted, so `useCopy`'s
+                    // state-based failure channel would have no renderer left
+                    // and a rejected write (insecure context, denied
+                    // permission) would look like a success. Report the
+                    // outcome as its own toast instead.
+                    onClick: () => {
+                      void navigator.clipboard
+                        .writeText(savedUrl)
+                        .then(() =>
+                          emitToast({
+                            kind: "success",
+                            message: "Link copied.",
+                          }),
+                        )
+                        .catch(() =>
+                          emitToast({
+                            kind: "error",
+                            message:
+                              "Couldn't copy the link — use “Copy share link” on the artifact row.",
+                          }),
+                        );
+                    },
                   },
                 }
               : {
@@ -81,7 +110,11 @@ export function ShareDialog({ artifact, onClose }: Props) {
 
   return (
     <Modal>
-      <DialogHeader title={`Share “${artifact.title}”`} onClose={onClose} />
+      <DialogHeader
+        title={`Share “${artifact.title}”`}
+        onClose={onClose}
+        closeDisabled={sharing.isPending}
+      />
       <DialogBody>
         <div className="flex flex-col gap-5">
           <label className="flex items-center justify-between gap-3">
@@ -152,6 +185,7 @@ export function ShareDialog({ artifact, onClose }: Props) {
         label="Save"
         pendingLabel="Saving…"
         pending={sharing.isPending}
+        cancelDisabled={sharing.isPending}
         onSubmit={save}
       />
     </Modal>
--- a/packages/ui/src/modules/artifacts/components/share-dialog.tsx
+++ b/packages/ui/src/modules/artifacts/components/share-dialog.tsx
@@ -13,6 +13,7 @@ import { Input } from "@/components/ui/input";
 import { Select } from "@/components/ui/select";
 import { Switch } from "@/components/ui/switch";
 import { useCopy } from "@/hooks/use-copy";
+import { emitToast } from "@/lib/toast";
 
 import { useSetArtifactSharing } from "../api/mutations.js";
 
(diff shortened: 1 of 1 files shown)


packages/ui/src/modules/artifacts/components/share-dialog.tsx
```diff
@@ -51,9 +55,25 @@ export function ShareDialog({ artifact, onClose }: Props) {
           : { expiresInHours: expiry === "never" ? null : Number(expiry) }),
       },
       {
-        onSuccess: (updated) => {
-          setShareUrl(updated.shareUrl);
-          if (!updated.shareUrl) onClose();
+        onSuccess: ({ shareUrl: savedUrl }) => {
```
dam-code-guardian, line 66: 🟡 **Warning:** The confirmation now lives entirely in this `mutate`-scoped `onSuccess`, but nothing gates the exits while the save is in flight — `DialogActions` receives no `cancelDisabled` and `DialogHeader` no `closeDisabled` (cf. `create-api-key-dialog/create-form.tsx:151`, `slack-channel-modal.tsx:83`), and all four parents render this dialog as `{target && <ShareDialog … onClose={() => setTarget(null)} />}`, so `onClose` unmounts it.

Closing mid-save therefore lands the write with **no toast and no copy affordance** — the same "reads as a no-op" outcome the PR set out to fix, and worse for the enable-a-link flow where the URL was just minted. Before the change this mattered less: the dialog stayed open and the URL field itself was the feedback.

Either gate the exits on `sharing.isPending`, or move the toast into `useSetArtifactSharing`'s own `onSuccess` so it doesn't depend on the observer still being mounted.

packages/ui/src/modules/artifacts/components/share-dialog.tsx
```diff
@@ -51,9 +55,25 @@ export function ShareDialog({ artifact, onClose }: Props) {
...
+          // Closing takes the link field away with it, so the toast carries the
+          // copy affordance for the one flow that just produced a fresh URL.
+          emitToast(
+            savedUrl
+              ? {
+                  kind: "success",
+                  message: "Sharing updated — the public link is live.",
+                  action: {
+                    label: "Copy link",
+                    onClick: () => void copy(savedUrl),
```
dam-code-guardian: 🟡 **Warning:** `useCopy`'s only failure channel is component state — `copy()` catches the rejection and sets `state = "failed"`, which callers render themselves (`copyable-command.tsx:35`, `reveal-token.tsx:49` both show an explicit "couldn't copy" line). Here the dialog is already unmounted by `onClose()` before this action can be clicked, so `setState` lands on a dead component: a clipboard rejection (insecure context, denied permission) produces no feedback at all and the toast dismisses as if the copy succeeded.

The in-dialog button at line 112 keeps working because it renders `copied` — this call site has no renderer. Emitting the result directly from the action keeps the failure visible:

```suggestion
                    onClick: () => {
                      void navigator.clipboard
                        .writeText(savedUrl)
                        .then(() =>
                          emitToast({ kind: "success", message: "Link copied." }),
                        )
                        .catch(() =>
                          emitToast({
                            kind: "error",
                            message: "Couldn't copy the link — copy it from the artifact row.",
                          }),
                        );
                    },
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `a0a23ea`

## PR #3153: fix(ui): close artifact share dialog and toast on save
**Author:** Tomas2D | **Branch:** fix/artifact-share-dialog-save-feedback → main | **Changes:** +26 −6 (1 file)

### Summary
Makes a successful sharing save close `ShareDialog` and confirm via a success toast (carrying a "Copy link" action when the save produced a public URL), replacing the old behavior that closed only when the link was *removed*. The local `shareUrl` mirror is dropped in favour of reading `artifact.shareUrl` from props.

### Findings
- 🟡 **Warning:** Closing the dialog while the save is in flight loses all confirmation — reproducing the very "reads as a no-op" bug this PR fixes (`packages/ui/src/modules/artifacts/components/share-dialog.tsx:58`)
- 🟡 **Warning:** The toast's "Copy link" reports failure only through the unmounted dialog's state, so a clipboard rejection is silent (`packages/ui/src/modules/artifacts/components/share-dialog.tsx:68`)
- ✅ **Looks good:** Dropping the `shareUrl` state mirror is safe — `meta.invalidates` refetches `artifactLibrary.list`, and all four render sites derive the `artifact` prop from that query, so the prop is the single source of truth.

**Verdict** — `aligned`

- 🟡 **Warning:** The success toast is emitted only from the `mutate`-scoped `onSuccess`, which is owned by this component's mutation observer, yet Cancel/✕ stay enabled during the save (`cancelDisabled`/`closeDisabled` are not passed to `DialogActions`/`DialogHeader`, unlike `create-form.tsx:151` and `slack-channel-modal.tsx:83`) and `onClose` unmounts the dialog from all four parents — so closing while the save is in flight completes the write with zero confirmation and no copy affordance, i.e. exactly the "read as a no-op" failure this PR set out to fix. Worth confirming whether react-query 5.99 still invokes `mutate`-scoped callbacks after the observer unmounts; if it doesn't, either gate Cancel/✕ on `sharing.isPending` or move the toast into `useSetArtifactSharing`'s own `onSuccess` so it survives unmount (`packages/ui/src/modules/artifacts/components/share-dialog.tsx:58`).
- 🟡 **Warning:** The toast's "Copy link" closure calls `copy` from `useCopy`, whose only failure channel is component state (`setState("failed")`) rendered by the in-dialog button — but the dialog is already unmounted by `onClose()` when the action can be clicked, so a clipboard rejection (insecure context, denied permission) is swallowed entirely and the user believes the link was copied; the scheduled 2s reset timer also fires after the hook's unmount cleanup ran. Emit a copy result toast from the action instead of relying on `useCopy`'s state (`packages/ui/src/modules/artifacts/components/share-dialog.tsx:68`).

### Verdict
COMMENT — The core fix is correct and the state-mirror removal is a genuine simplification; two edge cases (close-during-save, silent copy failure) are worth addressing but don't block.

---

JanPokorny (approved):

PR #3154, opened 2026-08-04
fix(ui): hold usage figures across month changes and latch disabled metrics

### LINKED ISSUE #3148: UI - Usage flashes a skeleton on every month change, and re-runs a doomed query when metrics are disabled

Tomas2D:
## Summary

Changing the month on Settings › Usage re-flashed the full four-section skeleton before showing anything, because each range is its own query key and a fresh key starts with no cached data. On a deployment without a telemetry store it was worse: every month change fired a request guaranteed to fail, so the page flickered skeleton → "Usage metrics are unavailable on this deployment." each time.

`useSpendBreakdown` (`modules/metrics/api/queries.ts`):

- `placeholderData: keepPreviousData` — the previous month's figures stay on screen until the new ones land, instead of tearing the page down to its skeleton.
- A deployment-level latch. `isMetricsDisabled` matches `TRPCClientError` with `data.code === "PRECONDITION_FAILED"`, which is what `createDisabledMetricsService` throws. Once seen it latches, and `enabled: !metricsDisabled` stops every subsequent month from firing a request we already know will fail.
- Returns named fields rather than spreading the query result — spreading reads every tracked property and opts the page out of React Query's render tracking.

`UsageView` (`modules/metrics/views/usage-view.tsx`):

- `isUnavailable` is an early return: header plus the notice, and **no period control at all**, since there is no month that could show anything. The verdict renders once instead of being re-derived per month behind a skeleton.
- `monthOfRows()` derives the chart's month from the rows' own `YYYY-MM-DD` keys. This is required by `keepPreviousData`, not optional polish: without it `fillMonthDays` would remap the held-over rows onto the newly selected month and briefly draw a fabricated all-zero chart — worse than showing stale data.
- Held-over figures are dimmed (`opacity-60 transition-opacity`) and marked `aria-busy`, so stale numbers do not read as final.
- Non-precondition failures get their own copy, `Couldn't load usage for <month>.`, instead of the old blanket "unavailable on this deployment" for any error.
- Extracted the repeated centered `Card` into `NoticeCard` (unavailable, error, and "No LLM calls" states) and the header description into a constant, which the early return also needs.

Closes #3148

## Test plan

- [ ] Manual: page through several months — figures hold and dim while loading, no skeleton after the first load
- [ ] Manual: confirm the day chart never flashes an all-zero month mid-switch
- [ ] Manual: on a deployment without a telemetry store, confirm the unavailable notice renders once with no period control, and that changing month fires no further requests
- [ ] Manual: force a non-precondition error and confirm the per-month copy appears, not the deployment-level notice

No automated test: `packages/ui/src/__tests__/unit` has no `renderHook` or testing-library harness (all suites are pure-logic), so covering the latch and the placeholder path means introducing a React render setup first. `monthOfRows` and the latch are verified by review and manual testing only.

### DIFF
--- a/packages/ui/src/modules/metrics/views/usage-view.tsx
+++ b/packages/ui/src/modules/metrics/views/usage-view.tsx
@@ -8,7 +8,6 @@ import { Card } from "@/components/ui/card";
 import { PageHeader } from "@/components/ui/page-header";
 import { SectionLabel } from "@/components/ui/section-label";
 import { formatDate } from "@/lib/format-time";
-import { cn } from "@/lib/utils";
 
 import { useSpendBreakdown } from "../api/queries.js";
 import { AgentSpendBars } from "../components/agent-spend-bars.js";
@@ -98,6 +97,14 @@ export function UsageView() {
 
   const total = data?.byModel.reduce((sum, row) => sum + row.costUsd, 0) ?? 0;
   const dataMonth = monthOfRows(data?.byDay, month);
+  // Copy inside the figures block describes the month the rows came from, not
+  // the one the user just picked — while a switch is in flight those differ,
+  // and naming the selection would assert something about a month whose data
+  // has not arrived. The period control keeps showing `month`, the selection.
+  const dataMonthLabel = formatDate(dataMonth, {
+    month: "long",
+    year: "numeric",
+  });
   const dailyDays = fillMonthDays(
     dataMonth,
     dataMonth >= currentMonth,
@@ -110,46 +117,58 @@ export function UsageView() {
         title="Usage"
         description={PAGE_DESCRIPTION}
         actions={
-          <div className="flex items-center gap-1">
-            <Button
-              variant="outline"
-              size="icon-sm"
-              aria-label="Previous month"
-              onClick={() => setMonth(monthStart(month, -1))}
-            >
-              <ChevronLeft size={16} className="text-muted-foreground" />
-            </Button>
-            <span className="min-w-[120px] text-center text-sm font-medium">
-              {monthLabel}
-            </span>
-            <Button
-              variant="outline"
-              size="icon-sm"
-              aria-label="Next month"
-              disabled={isCurrentMonth}
-              onClick={() => setMonth(monthStart(month, 1))}
-            >
-              <ChevronRight size={16} className="text-muted-foreground" />
-            </Button>
+          <div className="flex items-center gap-3">
+            {/* Names the staleness instead of only dimming it: the figures
+                below still belong to the previously viewed month. Dimming the
+                whole block was the earlier cue, but at 60% its
+                `text-muted-foreground` body copy fell under the AA contrast
+                minimum. */}
+            {isPlaceholderData && (
+              <span className="text-xs text-muted-foreground">Updating…</span>
+            )}
+            <div className="flex items-center gap-1">
+              <Button
+                variant="outline"
+                size="icon-sm"
+                aria-label="Previous month"
+                onClick={() => setMonth(monthStart(month, -1))}
+              >
+                <ChevronLeft size={16} className="text-muted-foreground" />
+              </Button>
+              <span className="min-w-[120px] text-center text-sm font-medium">
+                {monthLabel}
+              </span>
+              <Button
+                variant="outline"
+                size="icon-sm"
+                aria-label="Next month"
+                disabled={isCurrentMonth}
+                onClick={() => setMonth(monthStart(month, 1))}
+              >
+                <ChevronRight size={16} className="text-muted-foreground" />
+              </Button>
+            </div>
           </div>
         }
       />
 
+      {/* A failed month replaces the figures rather than holding them: React
+          Query applies `placeholderData` only while a query is pending, so the
+          rejected key has no data to keep, and `retry: false` means that is the
+          first failure. Showing the previous month's numbers under an error
+          would need a last-good cache of our own — deliberately out of scope,
+          since the figures are money and a stale total beside a failure notice
+          is worse than an honest gap. */}
       {isError && (
         <NoticeCard>Couldn't load usage for {monthLabel}.</NoticeCard>
       )}
-      {isPending && !isError && <UsageSkeleton />}
+      {isPending && <UsageSkeleton />}
       {data && (
         // `isPlaceholderData` means these are the previously viewed month's
-        // figures, held on screen while the selected month loads; dim them so
-        // the numbers don't read as final.
-        <div
-          aria-busy={isPlaceholderData}
-          className={cn(
-            "space-y-10 transition-opacity",
-            isPlaceholderData && "opacity-60",
-          )}
-        >
+        // figures, held on screen while the selected month loads. `aria-busy`
+        // carries that to assistive tech; the visible cue is the "Updating…"
+        // label by the period control.
+        <div aria-busy={isPlaceholderData} className="space-y-10">
           <section>
             <SectionLabel spaced>Total spend</SectionLabel>
             <div className="font-mono text-5xl font-bold leading-none tracking-[-0.02em] tabular-nums text-foreground">
@@ -159,7 +178,7 @@ export function UsageView() {
           {data.byModel.length === 0 ? (
             <section>
               <SectionLabel spaced>Spend by day</SectionLabel>
-              <NoticeCard>No LLM calls in {monthLabel}.</NoticeCard>
+              <NoticeCard>No LLM calls in {dataMonthLabel}.</NoticeCard>
             </section>
           ) : (
             <>
--- a/packages/ui/src/modules/metrics/views/usage-view.tsx
+++ b/packages/ui/src/modules/metrics/views/usage-view.tsx
@@ -1,5 +1,6 @@
 import { ChevronLeft, ChevronRight } from "@carbon/icons-react";
 import type { SpendByDay } from "api-server-api";
+import type { ReactNode } from "react";
 import { useState } from "react";
 
 import { Button } from "@/components/ui/button";
@@ -7,6 +8,7 @@ import { Card } from "@/components/ui/card";
 import { PageHeader } from "@/components/ui/page-header";
 import { SectionLabel } from "@/components/ui/section-label";
 import { formatDate } from "@/lib/format-time";
+import { cn } from "@/lib/utils";
 
 import { useSpendBreakdown } from "../api/queries.js";
 import { AgentSpendBars } from "../components/agent-spend-bars.js";
@@ -21,6 +23,13 @@ import { formatUsdCents } from "../lib/format.js";
 // resulting instants, so "calendar month" means the user's wall-clock month.
 const monthStart = (base: Date, offset: number) =>
   new Date(base.getFullYear(), base.getMonth() + offset, 1);
+
+const PAGE_DESCRIPTION = (
+  <span className="block max-w-[460px]">
+    LLM API spend across all supported agents (currently only Claude Code and
+    derivatives).
+  </span>
+);
 
 // The browser owns calendar semantics: from the sparse per-day rows the server
 // returns, build the full day list for the selected month, zero-filling days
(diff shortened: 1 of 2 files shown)


packages/ui/src/modules/metrics/views/usage-view.tsx
```diff
@@ -121,13 +159,7 @@ export function UsageView() {
...
             <section>
               <SectionLabel spaced>Spend by day</SectionLabel>
-              <Card
-                className={`flex ${CHART_HEIGHT_CLASS} items-center justify-center p-5`}
-              >
-                <p className="text-sm text-muted-foreground">
-                  No LLM calls in {monthLabel}.
-                </p>
-              </Card>
+              <NoticeCard>No LLM calls in {monthLabel}.</NoticeCard>
```
dam-code-guardian: 🟡 **Warning:** While `isPlaceholderData` is true, `data` is the *previous* month's response but `monthLabel` is the newly selected month — so this branch renders an affirmative false statement: `No LLM calls in <newly selected month>.` about a month whose data has not arrived yet. Same class of misattribution as the flat-chart bug `monthOfRows` was added to prevent, just in the copy instead of the plot. The headline total (line 157) and the header label diverge the same way — "November 2026" over October's figure — and `opacity-60` dims a claim without qualifying it.

`dataMonth` is already computed one line above the render, so labels *inside* the `{data && …}` block can format from it while the period control stays on `month` (it correctly reflects the user's selection):

```suggestion
              <NoticeCard>No LLM calls in {formatDate(dataMonth, { month: "long", year: "numeric" })}.</NoticeCard>
```

packages/ui/src/modules/metrics/views/usage-view.tsx
```diff
@@ -101,17 +136,20 @@ export function UsageView() {
       />
 
       {isError && (
```
dam-code-guardian, line 162: 🟡 **Warning:** This path discards exactly the figures the PR exists to hold. In TanStack Query v5 `placeholderData` is applied only while `status === "pending"` **and** `data === undefined` (`queryObserver.ts`, the `use placeholderData if needed` block). Once the newly selected month's request rejects, `status` flips to `"error"`, the placeholder branch is skipped, `data` is `undefined` for that fresh key — and the page collapses to this bare notice, the teardown `keepPreviousData` was added to prevent. With `retry: false` on the query that happens on the first failure, immediately.

If the intent is "hold the figures across month changes", the error state should keep the last good figures (dimmed, with the failure named alongside) rather than replacing them. If replacing them is deliberate, it's worth a comment — the PR description's own rationale argues the other way.

packages/ui/src/modules/metrics/views/usage-view.tsx
```diff
@@ -101,17 +136,20 @@ export function UsageView() {
...
       {data && (
-        <div className="space-y-10">
+        // `isPlaceholderData` means these are the previously viewed month's
+        // figures, held on screen while the selected month loads; dim them so
+        // the numbers don't read as final.
+        <div
+          aria-busy={isPlaceholderData}
+          className={cn(
+            "space-y-10 transition-opacity",
+            isPlaceholderData && "opacity-60",
```
dam-code-guardian: 🟡 **Warning:** The dimmed subtree contains `text-muted-foreground` body copy (the notice card, table cells). `--muted-foreground` is `#4d5358` in the light theme; composited at 60% over the `#ffffff` card it lands at **2.91:1** against the background — under the 4.5:1 AA minimum for body text. Dark theme is **3.38:1** (`#a8a29e` over `#161616`), also under. Measured, not estimated. WCAG has no exemption for transient states, and this one persists for the whole fetch.

`aria-busy` is the right call and covers assistive tech; the visual staleness cue just needs to come from something that doesn't attack legibility — dim only the large `text-foreground` figures, use a lighter reduction, or mark staleness with a label near the period control.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `8102ee74`

## PR #3154: fix(ui): hold usage figures across month changes and latch disabled metrics
**Author:** Tomas2D | **Branch:** fix/usage-month-switch-flicker → main | **Changes:** +106 −27 (2 files)

### Summary
Adds `keepPreviousData` to the Usage tab's spend query so a month switch no longer tears the page down to its skeleton, plus a latched `PRECONDITION_FAILED` verdict that stops firing doomed requests on deployments with no telemetry store. `monthOfRows()` correctly re-anchors the day chart's zero-fill to the month the held-over rows actually describe — the subtle bug this feature would otherwise introduce, caught and fixed in the same PR.

### Findings
- 🟡 **Warning:** Month labels still name the *selected* month while the figures belong to the previous one (`packages/ui/src/modules/metrics/views/usage-view.tsx:162`)
- 🟡 **Warning:** The error path discards the held-over figures the PR exists to preserve (`packages/ui/src/modules/metrics/views/usage-view.tsx:138`)
- 🟡 **Warning:** `opacity-60` on muted text drops contrast to 2.91:1, below AA (`packages/ui/src/modules/metrics/views/usage-view.tsx:150`)
- 🟢 **Suggestion:** `isPending && !isError` — `!isError` is provably dead over a single status enum (`packages/ui/src/modules/metrics/views/usage-view.tsx:141`)
- ✅ **Looks good:** `monthOfRows` reads `rows[0]`, which is safe — the ClickHouse `spendByDay` query is `ORDER BY day` ascending, so the first row's month is the rows' month.

**Verdict:** minor drift — 2 items, both on `docs/architecture/metrics.md` (the page that owns the spend read path and the only page making claims about the Settings **Usage** tab).

**1 — The page still describes the tab as having one loading state; it now holds the previous range's figures across a range switch.**
`metrics.md:32` — "the client renders the tab from a **single loading/error state**, never showing a stale total above an empty chart while a slower rollup is still in flight." The rollup-consistency guarantee still holds (held-over rollups all come from one prior response, and the calendar re-anchoring exists precisely to keep the chart from going flat under a stale total), but the page nowhere reflects that a range switch no longer returns the tab to its loading state — this PR's headline behavior on a surface the page describes. `metrics.md:34` also assigns zero-filling and the current-month cutoff to the client without saying which month those are computed against, which is now load-bearing.

Proposed edit (extend line 32; one clause at line 34):
> …and the client renders the tab from a single loading/error state, never showing a stale total above an empty chart while a slower rollup is still in flight. Because each range is a distinct read, the client **holds the previously viewed range's figures on screen — visibly provisional — while a newly selected range is in flight**, rather than returning to its loading state on every switch.

> (line 34) …Generating the full month, zero-filling the gaps, and (for the current month) stopping at today are the client's job… — computed against **the month the returned rows describe**, which during a range switch is still the previously viewed one.

**2 — The Disabled backend section describes a per-read error reaction; the tab now latches a deployment-level verdict and stops reading.**
`metrics.md:52` documents two states only (unavailable, empty-but-enabled). The reaction changed shape: from a per-read message rendered in place of the chart, to a once-latched deployment verdict that suppresses all further reads and takes over the page, with per-range read failure split out as its own third state. The "no store means no usage for any range, so stop asking" coupling is durable and belongs on the page.

Proposed edit (append to line 52):
> …and shows an unavailable message, distinct both from the empty-but-enabled state where a real store simply has no rows for the window, and from an ordinary failed read, which is reported against the range that failed. Because the verdict is about the **deployment** rather than the range asked for, the client **latches it on first sight and stops reading altogether** — no later range retries a read that is known to fail — and reports it in place of the whole tab rather than per range.

Both edits are additive clauses; the page is ~11.3k chars against the 40,000-char cap, and its `Last verified:` header is already `2026-08-04`, so no bump is needed.

**1. Contract knowledge duplicated in the client instead of expressed in the contract package**

- **Rule:** the API contract package is the single source of truth for the API contract — no manual type duplication; translation between a domain condition and a tRPC error code belongs on the contract-router side.
- **Location:** `packages/ui/src/modules/metrics/api/queries.ts:10-15` (`isMetricsDisabled`), consumed at `:28` and `:36`.
- **Evidence:** `error instanceof TRPCClientError && error.data?.code === "PRECONDITION_FAILED"` hand-encodes the fact that a deployment with no telemetry store answers every metrics read with `PRECONDITION_FAILED`, and derives a *deployment capability* from it. That fact appears nowhere in the contract package — `packages/api-server-api/src/modules/metrics/types.ts` describes only the happy-path shapes and exposes no capability field — so this string literal and `createDisabledMetricsService` are two hand-kept halves of an off-contract convention with no shared type binding them. The change amplifies it: before, the error was passed through to the view as an opaque failure; now the client reinterprets the code as a durable capability verdict and gates request firing on it.
- **From:** `architecture/stack.md` → *Shared Types*; `architecture/slice-composition.md` → *Error Handling*.

**Smallest set of changes:** add the metrics-availability verdict to the contract package as data — e.g. an `available` field on the metrics read result (or a small `metrics.availability` procedure) returned by both the live and disabled service implementations. `useSpendBreakdown` then reads a typed field instead of matching `"PRECONDITION_FAILED"`, removing both the string literal and the client-side latch as a source of contract truth.

**🟠 The month label still names the selected month while the figures on screen belong to the previous one** — `usage-view.tsx:100-105`, `122-124`, `155-157`, `162`

`monthOfRows()` fixes the *chart's* zero-fill calendar, but every textual claim on the page is still derived from `month` (the newly selected one) rather than `dataMonth`. The sharpest case is line 162: view a month with no LLM calls, then click to the next month — while the new month is in flight the page renders an affirmative false statement, `No LLM calls in <newly selected month>.`, about a month whose data has not arrived. Same class of misattribution as the flat-chart bug this PR fixes, just in the copy instead of the plot. The header label and the headline total diverge the same way: "November 2026" over October's `$902.03`. `opacity-60` does not qualify a claim. `dataMonth` is already computed, so labels inside the `{data && …}` block should format from it, leaving the period control (which reflects the user's *selection*, correctly) on `month`.

**🟡 The error path drops the figures the PR exists to hold** — `usage-view.tsx:138-141`

In TanStack Query v5, `placeholderData` is applied only while `status === 'pending'` **and** `data === undefined`. Once the new month's request rejects, `status` flips to `'error'`, the placeholder branch is skipped, `data` is `undefined` for that fresh key, and the page collapses to the bare notice — exactly the teardown `keepPreviousData` was added to prevent. With `retry: false`, that happens on the first failure. If the intent is "hold the figures across month changes", the error state should keep the last good figures (dimmed, failure named alongside); if replacing them is deliberate, the tradeoff deserves a comment, because the diff's own rationale argues the other way.

**🟡 `!isError` is provably dead** — `usage-view.tsx:141`

`isPending` is `status === 'pending'` and `isError` is `status === 'error'` over a single enum, so the two can never both be true; `isPending && !isError` is just `isPending`. The guard reads as if defending against a real overlap, which invites more of them.

**🟡 `NoticeCard` builds its classes with a template literal in a file that now imports `cn()`** — `usage-view.tsx:196-204`

```tsx
className={`flex ${CHART_HEIGHT_CLASS} items-center justify-center p-5`}
```
Template-literal class strings are the documented anti-pattern (use `cn()`). This is moved code, but the move is the moment to migrate it, and `cn` was imported into this very file three hunks earlier — so the file now demonstrates both conventions.

**🟡 `opacity-60` on muted text drops contrast below AA** — `usage-view.tsx:148-151`

The dimmed subtree contains `text-muted-foreground` copy. `--muted-foreground` is `#4d5358` in light theme; at 60% over the card background that composites to ≈2.9:1 against white — under the 4.5:1 AA minimum for body text. WCAG has no exemption for transient states, and this one persists for the whole fetch. `aria-busy` is the right call for assistive tech, but the visual cue should come from something that doesn't attack legibility — dim only the large `text-foreground` figures, use a lighter reduction, or mark staleness near the period control.

**🟡 A chart-height card is now the whole page's empty state** — `usage-view.tsx:92-94`, `194-204`

`CHART_HEIGHT_CLASS` is exported with an explicit rationale: *"Shared so the loading/error/empty cards in the Usage view reserve the exact same height and the layout never jumps."* That holds for a card standing in for the chart; it does not hold for the `isUnavailable` early return, where there is no chart to reserve space for — yet the updated doc comment ("or for the whole page") extends the component to cover it. Either give `NoticeCard` an optional `className`, or keep it chart-scoped and let the deployment-unavailable state use the existing page-level primitive (`packages/ui/src/components/ui/page-empty-state.tsx`).

### Verdict
COMMENT — the core mechanism is right and `monthOfRows` shows the held-over-data trap was understood; three 🟡s remain about what *else* is held over (labels, the error path) and the contrast cost of the dimming.

---

packages/ui/src/modules/metrics/views/usage-view.tsx
```diff
@@ -101,17 +136,20 @@ export function UsageView() {
...
-        <Card
-          className={`flex ${CHART_HEIGHT_CLASS} items-center justify-center p-5`}
-        >
-          <p className="text-sm text-muted-foreground">
-            Usage metrics are unavailable on this deployment.
-          </p>
-        </Card>
+        <NoticeCard>Couldn't load usage for {monthLabel}.</NoticeCard>
       )}
       {isPending && !isError && <UsageSkeleton />}
```
dam-code-guardian: 🟢 **Suggestion:** `isPending` is `status === "pending"` and `isError` is `status === "error"` over a single status enum, so the two can never both be true — `!isError` is dead. Harmless, but it reads as if defending against a real overlap, which invites more guards like it.

```suggestion
      {isPending && <UsageSkeleton />}
```

JanPokorny (approved):

PR #3155, opened 2026-08-04
feat(e2e): e2e:loop --test= spec filter

tomkis:
Adds a `--test=<filter>` option to `e2e:loop` that narrows the spec run against the warm cluster:

```bash
mise run e2e:loop --full --rebuild=apiserver --test=prompt-delivery
```

The value is Playwright's positional file/title filter, forwarded to the spec run after `--` so mise appends it to the playwright command.

Two limits, both noted in the task's option docs:
- The filter fully narrows the run only for full-only projects, which have no dependencies. A smoke spec still runs its whole dependency chain: Playwright always runs a selected project's dependencies, and here they are needed anyway because `e2e:reset` wipes the state the chain builds up.
- Full-only specs still need `--full`, otherwise their project doesn't exist and the filter matches nothing.

Verified: `mise run e2e-playwright:run-full -- invocation --list` resolves to `pnpm exec playwright test invocation --list` and matches the full-suite spec.

https://claude.ai/code/session_01XngAMdE1psd6swN2TZLFLm

### DIFF
--- a/tasks.toml
+++ b/tasks.toml
@@ -221,7 +221,7 @@ echo "e2e data reset complete."
 '''
 
 ["e2e:loop"]
-description = "Fast e2e rerun against the warm test cluster (no VM recreate): optionally rebuild components, wipe data, run the smoke suite. Options: --headed --full --rebuild=apiserver,ui,controller,keycloak,mock-agent --test=<playwright file/title filter, e.g. prompt-delivery>"
+description = "Fast e2e rerun against the warm test cluster (no VM recreate): optionally rebuild components, wipe data, run the smoke suite. Options: --headed --full --rebuild=apiserver,ui,controller,keycloak,mock-agent --test=<playwright file/title filter, e.g. prompt-delivery; smoke specs still run their dependency chain>"
 dir = "{{config_root}}"
 raw = true
 run = '''
@@ -247,10 +247,14 @@ for arg in "$@"; do
     # The pipeline default is smoke only; this flag is the manual trigger.
     --full) PLAYWRIGHT_TASK="e2e-playwright:run-full" ;;
     --rebuild=*) REBUILD="${arg#--rebuild=}"; REBUILD_SET=true ;;
-    # Narrow the run to matching specs (Playwright's positional file/title
-    # filter, substring or regex — e.g. --test=prompt-delivery runs the four
-    # prompt-delivery specs). Full-only specs also need --full, or their
-    # project doesn't exist and the filter matches nothing.
+    # Positional Playwright filter (file/title substring or regex). It fully
+    # narrows the run only for full-only projects, which have no dependencies
+    # — e.g. --full --test=prompt-delivery runs just the four prompt-delivery
+    # specs. A smoke spec still pulls in its whole dependency chain: Playwright
+    # always runs a selected project's dependencies, and here they're needed
+    # anyway because e2e:reset (step 3) wipes the state the chain builds up.
+    # Full-only specs also need --full, or their project doesn't exist and
+    # the filter matches nothing.
     --test=*) TEST_FILTER="${arg#--test=}" ;;
   esac
 done
--- a/tasks.toml
+++ b/tasks.toml
@@ -221,7 +221,7 @@ echo "e2e data reset complete."
 '''
 
 ["e2e:loop"]
-description = "Fast e2e rerun against the warm test cluster (no VM recreate): optionally rebuild components, wipe data, run the smoke suite. Options: --headed --full --rebuild=apiserver,ui,controller,keycloak,mock-agent"
+description = "Fast e2e rerun against the warm test cluster (no VM recreate): optionally rebuild components, wipe data, run the smoke suite. Options: --headed --full --rebuild=apiserver,ui,controller,keycloak,mock-agent --test=<playwright file/title filter, e.g. prompt-delivery>"
 dir = "{{config_root}}"
 raw = true
 run = '''
@@ -239,6 +239,7 @@ fi
 PLAYWRIGHT_TASK="e2e-playwright:run"
 REBUILD=""
 REBUILD_SET=false
+TEST_FILTER=""
 for arg in "$@"; do
   case "$arg" in
     --headed) PLAYWRIGHT_TASK="e2e-playwright:run-headed" ;;
@@ -246,6 +247,11 @@ for arg in "$@"; do
     # The pipeline default is smoke only; this flag is the manual trigger.
     --full) PLAYWRIGHT_TASK="e2e-playwright:run-full" ;;
     --rebuild=*) REBUILD="${arg#--rebuild=}"; REBUILD_SET=true ;;
+    # Narrow the run to matching specs (Playwright's positional file/title
+    # filter, substring or regex — e.g. --test=prompt-delivery runs the four
+    # prompt-delivery specs). Full-only specs also need --full, or their
+    # project doesn't exist and the filter matches nothing.
+    --test=*) TEST_FILTER="${arg#--test=}" ;;
   esac
 done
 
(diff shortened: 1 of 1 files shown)


tasks.toml
```diff
@@ -239,13 +239,19 @@ fi
...
+TEST_FILTER=""
 for arg in "$@"; do
   case "$arg" in
     --headed) PLAYWRIGHT_TASK="e2e-playwright:run-headed" ;;
     # Full suite: smoke plus the slow on-demand specs under src/tests/full/.
     # The pipeline default is smoke only; this flag is the manual trigger.
     --full) PLAYWRIGHT_TASK="e2e-playwright:run-full" ;;
     --rebuild=*) REBUILD="${arg#--rebuild=}"; REBUILD_SET=true ;;
+    # Narrow the run to matching specs (Playwright's positional file/title
+    # filter, substring or regex — e.g. --test=prompt-delivery runs the four
```
dam-code-guardian: 🟡 **Warning:** This comment promises the filter narrows the run, but for every smoke spec it won't — Playwright applies a positional filter only to the *primary* selection, then runs each selected project's `dependencies` in full. From Playwright's own docs (`docs/src/test-projects-js.md`, “Test filtering”):

> All test filtering options, such as `--grep`/`--grep-invert`, `--shard`, filtering directly by location in the command line, or using `test.only()`, directly select the primary tests to be run. **If those tests belong to a project with dependencies, all tests from those dependencies will also run.**

`playwright.config.ts` wires the smoke tier as one long dependency chain (`auth → connection → agent → messages → experiments → injection → slack → slack-shared → slack-inchat → slack-ambient → egress-path-rules → connection-regrant`). So `--test=04-messages` still runs `auth`, `connection` and `agent` before it, and `--test=11-egress-path-rules` drags in essentially the entire suite — the opposite of “rerun a single spec without running the whole tier”.

The filter *does* work as advertised for the full-only projects (`invocations-full`, `slack-full`), which declare no `dependencies` — which is probably why it looked right when verified against `invocation`.

Worth either forwarding `--no-deps` alongside the filter (accepting that the chain's `storageState` must already exist on the warm cluster — plausible here, since `e2e:loop` targets a cluster a previous run already authenticated against), or narrowing this comment and the task description to say the run is filtered but its dependency chain still executes.

tasks.toml
```diff
@@ -221,7 +221,7 @@ echo "e2e data reset complete."
 '''
 
 ["e2e:loop"]
-description = "Fast e2e rerun against the warm test cluster (no VM recreate): optionally rebuild components, wipe data, run the smoke suite. Options: --headed --full --rebuild=apiserver,ui,controller,keycloak,mock-agent"
+description = "Fast e2e rerun against the warm test cluster (no VM recreate): optionally rebuild components, wipe data, run the smoke suite. Options: --headed --full --rebuild=apiserver,ui,controller,keycloak,mock-agent --test=<playwright file/title filter, e.g. prompt-delivery>"
```
dam-code-guardian: 🟡 **Warning:** `prompt-delivery` doesn't match anything in the suite — there is no spec file or test title containing it (the only repo hit is an unrelated comment in `packages/api-server/src/apps/harness-api-server/app.ts`). The nearest real spec is `04-messages.spec.ts` (“background prompt mid-turn keeps the reply paired…”). The line 251 comment goes further and claims it “runs the four prompt-delivery specs”, which no grouping in `src/tests/` corresponds to.

Since this string is the discoverable documentation for the option (`mise tasks`), an example that silently matches zero tests is the worst failure mode — Playwright exits with “no tests found” and the user can't tell whether their filter or the option is broken. Suggest using a filter that demonstrably resolves, e.g. `04-messages`.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `1f850dc`

## PR #3155: feat(e2e): e2e:loop --test= spec filter
**Author:** tomkis | **Branch:** feat/e2e-loop-test-filter → main | **Changes:** +10 −3 (1 files)

### Summary
Adds a `--test=<filter>` option to the `e2e:loop` mise task, forwarding the value as Playwright's positional file/title filter after `--`. The shell plumbing is correct; the option's documented behavior is not what Playwright will actually do.

### Findings
- 🟡 **Warning:** documented narrowing does not happen — smoke projects have `dependencies`, so a filtered run still runs the whole upstream chain (`tasks.toml:251`)
- 🟡 **Warning:** the `prompt-delivery` example names specs that don't exist (`tasks.toml:224`)
- ✅ **Looks good:** `${TEST_FILTER:+-- "$TEST_FILTER"}` is the right construct — verified that an empty filter appends nothing, and that values containing spaces or glob characters survive as a single unsplit, unglobbed argument.

**Verdict:** aligned

### Verdict
COMMENT — the mechanism works, but both documented claims about what `--test=` does are wrong in a way that will mislead whoever uses it.

---

tasks.toml
tomkis: Fixed in 85144b58. I went with narrowing the docs rather than forwarding --no-deps: e2e:loop always runs e2e:reset before the specs, so the smoke chain's state (connection, agent, storageState) is wiped every loop and the dependencies have to run anyway. The comment, task description and PR body now say the filter fully narrows only the dep-free full-only projects and that smoke specs still run their chain.

tasks.toml
```diff
@@ -239,13 +239,23 @@ fi
...
 REBUILD_SET=false
+TEST_FILTER=""
 for arg in "$@"; do
   case "$arg" in
     --headed) PLAYWRIGHT_TASK="e2e-playwright:run-headed" ;;
     # Full suite: smoke plus the slow on-demand specs under src/tests/full/.
     # The pipeline default is smoke only; this flag is the manual trigger.
     --full) PLAYWRIGHT_TASK="e2e-playwright:run-full" ;;
     --rebuild=*) REBUILD="${arg#--rebuild=}"; REBUILD_SET=true ;;
+    # Positional Playwright filter (file/title substring or regex). It fully
```
dam-code-guardian, line 250: 🟡 **Warning:** Both this comment and the task description call the value a "file/title filter", but a positional Playwright argument matches **file paths only** — test titles need `-g`/`--grep`. In v1.50.0, `config.cliArgs` feeds `createFileMatcherFromArguments` / `createFileFiltersFromArguments`, which build the matcher against `filePath` (plus an optional `:line:col` focus); the title matcher is built separately and exclusively from `cliGrep` (`runner/loadUtils.ts:130-133`). Playwright's own CLI docs draw the same line: positional args "run files that have `my-spec` … in the file name", while running a test by title is `-g "add a todo item"`.

So `--test="one agent serves two Slack channels"` doesn't select that test — it matches no file and Playwright exits with `No tests found. Make sure that arguments are regular expressions matching test files.` Loud rather than silent, but the option's own docs sent the user there.

```suggestion
    # Positional Playwright filter — a substring or regex matched against spec
    # file paths, not test titles (titles need -g/--grep). It fully
```

The same phrase in the task `description` (line 224) needs the same correction.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `85144b5`

## PR #3155: feat(e2e): e2e:loop --test= spec filter
**Author:** tomkis | **Branch:** feat/e2e-loop-test-filter → main | **Changes:** +14 −3 (1 files)

### Summary
Docs-only follow-up: narrows the `--test=` option's documented scope rather than changing the mechanism. The dependency-chain claim is now accurate; the example spec name and the "file/title" semantics are still wrong.

### Changes since last review
Previous HEAD: `1f850dc` (2026-08-04T11:42:52Z) — verdict COMMENT

- ✅ **Fixed:** narrowing claim now matches Playwright — verified `invocations-full` / `slack-full` declare no `dependencies`, so the filter really does fully narrow there (`tasks.toml:250`)
- 🔁 **Still present:** `prompt-delivery` matches no spec, and the full tier holds 2 specs with 1 test each, so "the four prompt-delivery specs" cannot hold (`tasks.toml:224`, `252`)
- 🆕 **New:** positional filter matches file paths, not test titles (`tasks.toml:250`)

### Findings
- 🟡 **Warning:** positional filter matches file paths, not test titles (`tasks.toml:250`)

**Verdict:** aligned

### Verdict
COMMENT — the mechanism is right and one claim got fixed, but the option's docs still misdescribe both what it matches and what the example does.

---

JanPokorny (approved):

PR #3156, opened 2026-08-04
fix: make the KubeVirt VM backend work on OpenShift Virtualization

JanPokorny:
Deploying the VM backend to a real OpenShift Virtualization cluster (CNV 4.21, unprivileged virtiofsd, real KVM — vs. the dev k3s cluster's KubeVirt v1.8 + emulation) surfaced a chain of failures that made `backend.type: vm` agents completely unusable there. Each fix below was validated live on that cluster before landing.

### 1. containerDisk rejects the chart's default `imagePullPolicy: ""` (controller)
KubeVirt admission validates `containerDisk.imagePullPolicy` as a strict enum (`Always|IfNotPresent|Never`), unlike pod containers where `""` means "K8s tag-based default". The controller passed the chart default through verbatim, so **every vm reconcile hard-failed at the mutating webhook** (993 retries, capped backoff). Now the field is omitted when unset.

### 2. `~/.cache` symlink swap bricked boot on unprivileged virtiofs (platform-base + controller)
`agent-entrypoint` swaps `~/.cache` to pod-local disk with `ln -sfn` under `set -e`. On unprivileged virtiofsd (the OpenShift default: qemu uid, all caps dropped) a non-root guest cannot create symlinks — virtiofsd can't `lchown`, the guest kernel returns EPERM — so the runtime died on every start **with zero diagnostics** (stderr→kmsg only carries output after exec; the failure loop looked like a silent `exit-code` crashloop).
- entrypoint: a failed cache swap is a perf wart, never a boot failure — warn and continue.
- controller: pre-create the symlink from a root `bootcmd` in the VM userdata (root creates are accepted; ownership squashes to the daemon uid), so the entrypoint's `[ ! -L ]` check skips its own `ln`.

### 3. agent-runtime must run as root in the guest (claude-code-vm)
The bigger consequence of unprivileged virtiofsd: it EPERMs **every create** (file/dir/symlink) from any guest uid other than root — verified directly with `mkdir`/`touch` as uid 65532. The runtime as uid 65532 could never write the workspace (`EPERM mkdir ~/.platform`). Idmapped virtiofs mounts would be the clean fix but this virtiofsd doesn't support them (mounts, then EOVERFLOW on create).

`agent-runtime.service` now runs as root. Isolation is the VM boundary itself — docker and k3s already run as root in this guest. Rationale documented in the unit.

### 4. Missing WORK_DIR fails harness spawn with a misleading error (agent-runtime)
On a fresh workspace volume, spawning the harness before the init script created `~/work` failed with `spawn /usr/local/bin/harness-chat ENOENT` — pointing at the binary when the actual missing path is the cwd. The runtime now `mkdir -p`s the harness cwd before spawning.

### 5. `uv run` silently upgraded ruff (experiment-sdk)
Running the lint task re-resolved `uv.lock`, bumping ruff 0.15.22 → 0.16.0 whose new rules fail lint — exactly what the task's own comment says must not happen. All experiment-sdk `uv run` tasks are now `--frozen`.

## Validation
- controller: `mise run controller:test` green, new assertions for the omitted pull policy (set + unset) and the cache-symlink bootcmd.
- agent-runtime: `mise run agent-runtime:check` + 217 tests green, new test for missing-cwd spawn.
- Live: with 1–3 applied by hand on the CNV cluster, the VM agent boots clean (0 service failures), completes hello/capabilities, and an end-to-end ACP smoke (initialize → session/new → prompt) got a model-backed reply from Claude Code inside the VM.

## Known follow-ups (not in this PR)
- After a harness spawn failure the runtime keeps the dead process attached to the session; UI sends then time out silently ("Couldn't deliver") instead of failing fast. A `session/prompt` against an exited harness should error explicitly.
- Root-in-guest shifts `$HOME` seeding semantics (files land squashed to the virtiofsd uid); revisit if/when CNV's virtiofsd gains idmapped-mount support.

### DIFF
--- a/packages/agent-runtime/src/modules/acp/infrastructure/create-child-agent-process.ts
+++ b/packages/agent-runtime/src/modules/acp/infrastructure/create-child-agent-process.ts
@@ -1,4 +1,5 @@
 import { spawn } from "node:child_process";
+import { mkdirSync } from "node:fs";
 import readline from "node:readline";
 import type { AgentProcess } from "./agent-process.js";
 
@@ -19,6 +20,15 @@ export function createChildAgentProcess(
       ([k]) => !k.startsWith("npm_"),
     ),
   );
+
+  // A missing cwd fails the spawn as a misleading ENOENT on the *binary*
+  // (e.g. before the workspace init script has created WORK_DIR on a fresh
+  // volume). Creating it here is always safe — it's the harness's cwd.
+  try {
+    mkdirSync(opts.workingDir, { recursive: true });
+  } catch {
+    // unwritable workspace surfaces as the spawn error below
+  }
 
   const child = spawn(cmd, args, {
     stdio: ["pipe", "pipe", "inherit"],
--- a/packages/agent-runtime/src/modules/acp/infrastructure/create-child-agent-process.ts
+++ b/packages/agent-runtime/src/modules/acp/infrastructure/create-child-agent-process.ts
@@ -24,10 +24,14 @@ export function createChildAgentProcess(
   // A missing cwd fails the spawn as a misleading ENOENT on the *binary*
   // (e.g. before the workspace init script has created WORK_DIR on a fresh
   // volume). Creating it here is always safe — it's the harness's cwd.
+  // Non-fatal: the spawn below still runs, but its own error reports ENOENT
+  // on the *binary*, so the mkdir failure here is the only actionable line.
   try {
     mkdirSync(opts.workingDir, { recursive: true });
-  } catch {
-    // unwritable workspace surfaces as the spawn error below
+  } catch (err) {
+    process.stderr.write(
+      `[agent-process] could not create workingDir ${opts.workingDir}: ${(err as Error).message}\n`,
+    );
   }
 
   const child = spawn(cmd, args, {
--- a/packages/controller/pkg/reconciler/vm_resources.go
+++ b/packages/controller/pkg/reconciler/vm_resources.go
@@ -166,6 +166,16 @@ func BuildVMCloudInitSecret(name string, agentSpec *types.AgentSpec, cfg *config
 			shellQuote(m.Path), shellQuote(tag),
 		)})
 	}
+	// agent-entrypoint swaps ~/.cache for pod-local disk, but on unprivileged
+	// virtiofs a non-root caller cannot create symlinks (virtiofsd lacks
+	// CAP_CHOWN; the guest kernel returns EPERM) — so root pre-creates the
+	// link here and the entrypoint's `[ ! -L ]` check skips its own attempt.
+	// Runs after the mount bootcmds above so the workspace share is in place.
+	cc.BootCmd = append(cc.BootCmd, []string{"sh", "-c", fmt.Sprintf(
+		"mkdir -p /tmp/agent-cache && chown %[1]d:%[1]d /tmp/agent-cache && ln -sfn /tmp/agent-cache %[2]s/.cache || true",
+		vmAgentUID, shellQuote(agentHome),
+	)})
+
 	body, err := yaml.Marshal(cc)
 	if err != nil {
 		return nil, fmt.Errorf("encoding cloud-init userdata: %w", err)
--- a/packages/controller/pkg/reconciler/vm_resources.go
+++ b/packages/controller/pkg/reconciler/vm_resources.go
@@ -254,7 +254,13 @@ func BuildAgentVirtualMachine(name string, agentSpec *types.AgentSpec, cfg *conf
 	// else the first chart-wide default. (The pod path lists all defaults as
 	// fallbacks; KubeVirt's API takes one — a multi-secret install needs the
 	// matching secret first.)
-	bootDisk := map[string]any{"image": agentSpec.Image, "imagePullPolicy": pullPolicy}
+	// Unlike a pod's container, containerDisk validates imagePullPolicy as a
+	// strict enum — the chart's default "" (tag-based K8s behavior) must be
+	// omitted, not passed through, or admission rejects the VM.
+	bootDisk := map[string]any{"image": agentSpec.Image}
+	if pullPolicy != "" {
+		bootDisk["imagePullPolicy"] = pullPolicy
+	}
 	if ref := agentSpec.ImagePullSecretRef; ref != "" {
 		bootDisk["imagePullSecret"] = ref
 	} else if len(base.ImagePullSecrets) > 0 {
--- a/packages/controller/pkg/reconciler/vm_resources.go
+++ b/packages/controller/pkg/reconciler/vm_resources.go
@@ -171,8 +171,11 @@ func BuildVMCloudInitSecret(name string, agentSpec *types.AgentSpec, cfg *config
 	// CAP_CHOWN; the guest kernel returns EPERM) — so root pre-creates the
 	// link here and the entrypoint's `[ ! -L ]` check skips its own attempt.
 	// Runs after the mount bootcmds above so the workspace share is in place.
+	// `ln -sfn` into an existing real directory nests the link instead of
+	// swapping — clear a non-symlink first so the pre-create is authoritative
+	// on volumes that already carry a real ~/.cache.
 	cc.BootCmd = append(cc.BootCmd, []string{"sh", "-c", fmt.Sprintf(
-		"mkdir -p /tmp/agent-cache && chown %[1]d:%[1]d /tmp/agent-cache && ln -sfn /tmp/agent-cache %[2]s/.cache || true",
+		"mkdir -p /tmp/agent-cache && chown %[1]d:%[1]d /tmp/agent-cache && { [ -L %[2]s/.cache ] || rm -rf %[2]s/.cache; } && ln -sfn /tmp/agent-cache %[2]s/.cache || true",
 		vmAgentUID, shellQuote(agentHome),
 	)})
 
--- a/packages/platform-base/entrypoint.sh
+++ b/packages/platform-base/entrypoint.sh
@@ -31,9 +31,14 @@ fi
 # target is (re)created each boot to keep the link from dangling.
 home="${HOME:-/home/agent}"
 mkdir -p /tmp/agent-cache
+# On the VM backend $HOME is unprivileged virtiofs, where a non-root caller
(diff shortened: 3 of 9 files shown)


packages/platform-base/entrypoint.sh
```diff
@@ -31,9 +31,14 @@ fi
...
 mkdir -p /tmp/agent-cache
+# On the VM backend $HOME is unprivileged virtiofs, where a non-root caller
+# cannot create symlinks (virtiofsd lacks CAP_CHOWN, the guest kernel returns
+# EPERM); the VM userdata pre-creates the link as root, and if that ever
+# misses, a real ~/.cache on the share is a perf wart — never a boot failure.
 if [ ! -L "$home/.cache" ]; then
 	rm -rf "$home/.cache"
-	ln -sfn /tmp/agent-cache "$home/.cache"
+	ln -sfn /tmp/agent-cache "$home/.cache" \
+		|| echo "agent-entrypoint: WARNING: could not swap ~/.cache to local disk; caches stay on the workspace volume" >&2
```
dam-code-guardian: 🟡 **Warning:** the `ln` is now tolerated, but the `rm -rf` two lines up still isn't — and the ordering means a failed swap is worse than a no-op.

Before this change the block was all-or-nothing under `set -eu`: either the symlink existed or boot died. Now `rm -rf "$home/.cache"` runs unconditionally and, when the `ln` then fails, the boot continues with `~/.cache` **deleted and not replaced**. Every tool that expected it recreates a real directory on the workspace share — which is the perf wart the warning describes, so that part is intended. The gap is that the warning fires *after* the cache has already been destroyed, so a VM whose userdata pre-create also failed silently loses the existing cache contents on every boot rather than reusing them.

Cheapest fix is to only clear when the swap can actually proceed, so a failed `ln` leaves the existing directory intact:

```suggestion
if [ ! -L "$home/.cache" ]; then
	if ln -sfn /tmp/agent-cache "$home/.cache.tmp" 2>/dev/null; then
		rm -rf "$home/.cache" && mv "$home/.cache.tmp" "$home/.cache"
	else
		rm -f "$home/.cache.tmp"
		echo "agent-entrypoint: WARNING: could not swap ~/.cache to local disk; caches stay on the workspace volume" >&2
	fi
fi
```

(Or simply move the `rm -rf` after a successful probe — the shape matters less than not deleting on the failing path.)

packages/agent-runtime/src/modules/acp/infrastructure/create-child-agent-process.ts
```diff
@@ -20,6 +21,15 @@ export function createChildAgentProcess(
     ),
   );
 
+  // A missing cwd fails the spawn as a misleading ENOENT on the *binary*
+  // (e.g. before the workspace init script has created WORK_DIR on a fresh
+  // volume). Creating it here is always safe — it's the harness's cwd.
+  try {
+    mkdirSync(opts.workingDir, { recursive: true });
+  } catch {
+    // unwritable workspace surfaces as the spawn error below
```
dam-code-guardian: 🟡 **Warning:** the comment's premise doesn't hold — an unwritable workspace does **not** surface usefully in the spawn error, so this `catch` discards the only actionable diagnosis.

Ran the exact shape (unwritable parent, `recursive: true`, then `spawn` with that `cwd`):

```
mkdir threw: EACCES  EACCES: permission denied, mkdir '/tmp/spawntest/blocked/nested'
spawn error: ENOENT | spawn /usr/local/.../node ENOENT
```

The `EACCES` naming the real cause is dropped; what the operator gets is `ENOENT` **on the binary path** — exactly the misleading error the comment 5 lines above identifies as the bug being fixed. So for the unwritable-workspace case this preserves the original confusion and now also hides the explanation.

One line restores it, matching the two sibling handlers in this file (`child.on("error")` and `child.stdin.on("error")`), and keeps the non-fatal behavior:

```suggestion
  } catch (err) {
    process.stderr.write(
      `[agent-process] could not create workingDir ${opts.workingDir}: ${(err as Error).message}\n`,
    );
  }
```

packages/controller/pkg/reconciler/vm_resources.go
```diff
@@ -166,6 +166,16 @@ func BuildVMCloudInitSecret(name string, agentSpec *types.AgentSpec, cfg *config
 			shellQuote(m.Path), shellQuote(tag),
 		)})
 	}
+	// agent-entrypoint swaps ~/.cache for pod-local disk, but on unprivileged
+	// virtiofs a non-root caller cannot create symlinks (virtiofsd lacks
+	// CAP_CHOWN; the guest kernel returns EPERM) — so root pre-creates the
+	// link here and the entrypoint's `[ ! -L ]` check skips its own attempt.
+	// Runs after the mount bootcmds above so the workspace share is in place.
+	cc.BootCmd = append(cc.BootCmd, []string{"sh", "-c", fmt.Sprintf(
```
dam-code-guardian, line 177: 🟢 **Suggestion:** `ln -sfn` does not swap a path that is an existing **real directory** — it creates the link *inside* it and exits 0.

Verified:

```
$ mkdir -p home/.cache target && ln -sfn /tmp/lntest/target /tmp/lntest/home/.cache; echo $?
0
$ ls -la /tmp/lntest/home/.cache
lrwxrwxrwx  target -> /tmp/lntest/target      # nested, not swapped
```

On a workspace volume that already carries a real `~/.cache` (an older VM, or one where a previous boot's entrypoint swap failed and tools recreated it), this bootcmd silently produces `$HOME/.cache/target` and reports success. The entrypoint's `[ ! -L "$home/.cache" ]` check then still sees a non-symlink, tries its own `ln`, hits the same no-op, and — with the new `||` in this PR — just warns. Net effect: caches stay on the share permanently on exactly the volumes most likely to have the problem, with no signal beyond that one warning line.

Since root is running here and the intent is unconditional, clearing a non-symlink first makes the pre-create actually authoritative:

```suggestion
		"mkdir -p /tmp/agent-cache && chown %[1]d:%[1]d /tmp/agent-cache && { [ -L %[2]s/.cache ] || rm -rf %[2]s/.cache; } && ln -sfn /tmp/agent-cache %[2]s/.cache || true",
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `daa8334`

## PR #3156: fix: make the KubeVirt VM backend work on OpenShift Virtualization
**Author:** JanPokorny | **Branch:** fix/vm-agent-virtiofs-cache-symlink → main | **Changes:** +85 −9 (8 files)

### Summary
Fixes a chain of failures that made `backend.type: vm` agents unusable on OpenShift Virtualization: containerDisk's strict `imagePullPolicy` enum, unprivileged virtiofsd EPERMing non-root creates (cache symlink + runtime uid), a misleading spawn error on a missing cwd, and unpinned `uv run` upgrading ruff.

### Findings
- 🟡 **Warning:** `rm -rf` before the now-tolerated `ln` can leave `~/.cache` deleted with no symlink (`packages/platform-base/entrypoint.sh:39`)
- 🟡 **Warning:** swallowed `mkdir` error re-creates the misleading `ENOENT` this PR fixes (`packages/agent-runtime/src/modules/acp/infrastructure/create-child-agent-process.ts:27`)
- 🟢 **Suggestion:** `ln -sfn` into a surviving real `~/.cache` silently nests the link instead of swapping (`packages/controller/pkg/reconciler/vm_resources.go:174`)
- ✅ **Looks good:** the `imagePullPolicy` omission is verified on both branches (set + unset) by the new test, and `--frozen` coverage is complete — no unpinned `uv run` remains in the package.

**Verdict:** minor drift

`docs/architecture/persistence.md:113` states the cache redirect as an unconditional invariant — "the base-image entrypoint symlinks it to pod-local disk" — and lists `$HOME/.cache` under *what does not survive hibernation*. Two changes in this diff falsify that: the entrypoint's `ln` is now non-fatal (`entrypoint.sh:38-41`), and on the VM backend the guest pre-creates the link from cloud-init (`vm_resources.go:171-177`). After a failed swap, tools recreate `~/.cache` as a real directory **on the persisted mount**, where it does survive — a state the page admits no reading of. This is the shared base image, so the pod backend is affected too, not just the VM backend.

Proposed replacement for that sentence:

> `$HOME/.cache` is deliberately in this category: the base-image entrypoint redirects it to node-local disk so churn-heavy tool caches don't load the persistent volume. The redirect is best-effort in both directions — on the VM backend the guest pre-creates the link, since the workspace share there refuses a non-root symlink, and a swap that fails anywhere is a performance regression rather than a boot failure, leaving that agent's cache on the workspace volume where it does survive.

Also bump that page's `Last verified:` (currently `2026-07-30`). The page has 21,617 characters of headroom under the 40,000 cap (measured in the gate's own unit, `content.length` — not bytes), so the growth is not a cap concern.

**1. Silent `catch` in an adapter defeats the layer's error-surfacing convention** — `create-child-agent-process.ts:27-31`

Infrastructure adapters translate external-system failures into something an inner layer can act on (*infrastructure.md* → External System Adapters). The new `catch {}` is empty and its comment is factually wrong. Verified by running the exact shape:

```
mkdir threw: EACCES  EACCES: permission denied, mkdir '/tmp/spawntest/blocked/nested'
spawn error: ENOENT | spawn /usr/local/.../node ENOENT
```

`EACCES` is discarded; what reaches the operator is `ENOENT` **on the binary** — precisely the misleading error this PR exists to eliminate. The two sibling handlers in the same file set the convention (`[agent-process] …: ${err.message}` to stderr).

Not flagged: `mkdirSync` as an adapter side effect (correct layer, port unchanged), synchronous I/O in a synchronous factory (no rule against it), and the test file (exercises the port through its public factory — aligned).

### Verdict
COMMENT — the four fixes are real and well-evidenced; two error paths silently discard the diagnosis they should surface, and one architecture page now contradicts the code.

---

JanPokorny:
All four findings addressed in 2654c08e:
- entrypoint now probes the swap with a scratch link — a failing swap keeps the existing `~/.cache` intact (reviewer's suggested shape).
- the `mkdirSync` catch logs the failure to stderr like the sibling handlers, so EACCES isn't masked by the spawn's binary-ENOENT.
- the bootcmd clears a non-symlink `~/.cache` before `ln -sfn` so the pre-create is authoritative on volumes carrying a real cache dir.
- `persistence.md` documents the redirect as best-effort (proposed wording) + `Last verified` bumped.

Issue #3100, opened 2026-07-30
Epic - Notifications


jenna-winkler:
## Context

DAM's whole premise is background agents — they run while you're doing something else. But from a user's point of view, the product is silent: there's no experience of *being told* when something worth their attention happens. If you're not looking at a session, you simply don't find out that its agent finished, is waiting on you, or failed. What exists today is fragmented and easy to miss — an inbox you have to remember to check, toasts that vanish, a channel reply if you happen to be in Slack — and none of it adds up to a coherent sense of "DAM will let me know."

## Problem

For someone using DAM, the notification experience is missing or disjointed at exactly the moments it matters:

- **You don't feel kept in the loop.** An agent you set going finishes, and nothing tells you — you have to go back and check, so you either hover over a session that's meant to run in the background, or you forget about it.
- **You can't tell when an agent needs you.** When a run is blocked waiting on your approval, there's no clear, visible pull back to it. The work quietly stalls.
- **There's no single place that answers "what happened while I was away?"** Attention-worthy moments are scattered across surfaces that each behave differently, so there's no dependable, glanceable sense of what's new.
- **You have no say in what interrupts you.** Whatever signals exist aren't something a user shapes — no sense of "tell me about this, not that."

The result is an experience that either leaves users out of the loop or, where signals do exist, feels incidental and untrustworthy — never a designed part of the product.

## Goal

Users feel confidently kept informed. When something they'd care about happens — an agent finishing, needing approval, or failing — the product tells them in a way that's noticeable, glanceable, and trustworthy, and they can shape what's worth being interrupted for. Notifications feel like a designed, coherent part of the experience rather than scattered signals.

## Scope

**In:** the **notification experience** —
- what a user sees and where (a coherent, dependable place to catch up on what's new, versus signals that vanish),
- how attention is drawn without being noisy or nagging,
- the controls a user has over what notifies them and how.

**Out:** the underlying delivery mechanics (how events are detected, transported, or pushed) are the means, not the subject of this epic — they're shaped by what the experience needs, not the other way around.

## Open Questions

- What's the primary place a user goes to catch up — a notification center in the product, or being reached outside it (Slack/Telegram, push, email)? What does the experience need first?
- What are the right defaults so it helps without becoming noise, and how much per-user vs. per-agent control is worth it?
- How does the existing approvals inbox fit — is it folded into this experience, or does it stay distinct?

PR #3159, opened 2026-08-04
feat(ui): show running background work in a session

xjacka:
## Summary

The background-work contract (#2965, shipped in #3044) keeps a pod awake while a session reports running tasks, but the UI showed nothing during the run — a session between turns looked dead. This adds a minimal live indicator.

- **api-server**: new `agents.backgroundWork` tRPC query proxies the pod's `GET /api/status` `backgroundWork` sets on demand. Passive by contract: it sits outside the reachability primitive (no activity bump, no readiness wait), a hibernated agent answers `[]` without being woken, and an unreachable pod fails soft to `[]`.
- **ui**: slow-cadence working-dots indicator (success hue, task descriptions in the tooltip) in the chat header while the open session reports background work, and the same marker on session rows in the sidebar. Both surfaces share one query that polls every 30 s, only while the chat view is open and the agent is `running`. Disappears once the reported set empties; the result message then arrives through the existing flow.

No new persistence, no notifications (epic #3100 covers those).


packages/ui/src/modules/sessions/api/background-work.ts
```diff
@@ -0,0 +1,44 @@
...
+  agentId: string | null,
+): readonly SessionBackgroundWork[] {
+  const awake = useAgentRunState(agentId) === "running";
+  const { data } = useQuery({
+    queryKey: backgroundWorkKeys.agent(agentId),
+    queryFn:
+      agentId && awake
+        ? () => api.agents.backgroundWork.query({ id: agentId })
+        : skipToken,
+    refetchInterval: BACKGROUND_WORK_POLL_MS,
```
dam-code-guardian: 🟡 **Warning:** this query inherits the global `retry: 3` (`query-client.ts:51`), but the router throws `NOT_FOUND` when the agent is gone (`router.ts:103`) — a permanent error for that id, so a failing tick costs 4 requests that cannot succeed. The two closest analogues in this codebase — also agent-scoped, also gated, also reaching toward a pod — both disable it (`agents/api/harness-config.ts:12,46`). It also works against the deliberate server-side soft-fail (`agents-service.ts:256-258`), whose whole point is not to fight an unreachable pod.

Separately, the docstring says "Polls only while the agent is `running`" but the interval is unconditional — correct today via `skipToken`, though the sibling gates both (`sessions/api/queries.ts:140`: `refetchInterval: live ? STATUS_POLL_MS : false`), which keeps the invariant from drifting.

```suggestion
    refetchInterval: agentId && awake ? BACKGROUND_WORK_POLL_MS : false,
    retry: false,
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `5a3eb2c2`

## PR #3159: feat(ui): show running background work in a session
**Author:** xjacka | **Branch:** feat/background-work-indicator → main | **Changes:** +289 −4 (14 files)

### Summary
Adds a live indicator for the background-work contract (#2965/#3044): a passive `agents.backgroundWork` tRPC query proxies the pod's `/api/status` `backgroundWork` sets, and a slow-cadence dots marker renders in the chat header and on sidebar session rows. The passive-read design is carefully built — the hibernation short-circuit is tested against a probe that must not be called, and the soft-fail to `[]` keeps a UI poll from ever fighting an unreachable pod.

### Findings
- 🟡 **Warning:** inherited `retry: 3` on a permanent-error path (`packages/ui/src/modules/sessions/api/background-work.ts:32`)
- 🟡 **Warning:** query is silent on error — no `meta.errorToast` (`packages/ui/src/modules/sessions/api/background-work.ts:26`)
- 🟢 **Suggestion:** duplicated port declaration defeats adapter type-checking (`packages/api-server/src/modules/agents/infrastructure/pod-status-client.ts:26`)
- ✅ **Looks good:** the hibernated-agent path is verified by asserting the pod probe was never called, not just that the result was `[]` — the test encodes the actual invariant (never wake a pod), so a future refactor that reorders the guard fails loudly.

**Verdict: minor drift**

❌ **Check 1 — `docs/architecture/platform-topology.md`: the api-server→agent-runtime protocol table omits the new status-read channel.**

The table (lines 84–92) enumerates every api-server→agent-runtime channel: ACP relay WS, terminal WS, tRPC proxy, streamed import, runtime-channel `applyState`. This PR adds one that is neither tRPC nor listed — a direct plain-HTTP `GET` to the pod's status surface (`pod-status-client.ts`, composed at `compose.ts:108`).

The responsibility change is real: `agent-lifecycle.md:180` states the status surface is read by *the controller*, which "reports one boolean; the controller reads nothing more into it." After this PR the api-server is a **second reader** consuming a **different field** — one the pod labels "published for diagnosis only" (`agent-runtime/src/server.ts:466-468`). Proposed row:

```
| api-server → agent-runtime | HTTP (status read) | Passive read of the pod's status surface for reported background work; never wakes a pod or defers hibernation |
```

Both candidate pages have headroom in the gate's own unit (`content.length`, `scripts/doc-size.mjs:33,39`): platform-topology 14,694 / 40,000 chars, agent-lifecycle 34,144 — no offsetting cuts needed.

⚠️ **Possible drift — `docs/architecture/agent-lifecycle.md`: is the status surface now a contract rather than a diagnostic?**

Lines 123–135 frame publication as *explainability* ("so a sandbox that stays awake can be explained"), and the pod-side code agrees. This PR builds a shipped, 30 s-polled UI feature on that field. Does the platform now treat `backgroundWork` as a consumed contract — and should the page say so, given that "published for diagnosis" and "polled to drive a UI indicator" carry different change-safety guarantees for whoever edits the status payload next?

Ports & adapters, dependency direction, composition root, validation-layer boundary, and module public API are all respected. Two structural findings.

**1. Duplicated port declaration — the adapter restates the port instead of implementing the declared one.**
`agents-service.ts:41-43` declares `PodStatusPort`; `pod-status-client.ts:26-28` declares a structurally identical `PodStatusClient`, and `createPodStatusClient` returns the *latter*. Nothing type-checks the adapter against the port — they match only by coincidence of shape, so a widened return or added method surfaces at the `compose.ts:108` wiring rather than at the adapter.

This departs from the module's own convention: every sibling adapter declares its port once and the service imports it — `AgentsRepository` (`agents-repository.ts:35` → imported `agents-service.ts:22`), `AgentEnvRepository` (`:23`), `AgentRegistrySecretPort` (`:66`). The one port declared inside `agents-service.ts` (`ContributionsSettledPort:34`) is the case whose adapter lives in *another* module — the exception that proves the rule.

*Fix:* one declaration — import the port from `infrastructure/pod-status-client.js`, or type `createPodStatusClient(): PodStatusPort`.

**2. Placement judgement — the adapter sits in `agents/`, but the concept is session-scoped.**
The port returns `SessionBackgroundWork[]` keyed by `sessionId`, and every consumer is in the sessions surface; nothing in the agents context reads the value. Counter-argument is real: `podBaseUrl` and the `hibernated` gate are agents-owned, and the read is per-agent. Worth a conscious call — does agents own "what work runs inside this agent's sessions", or is it merely the addressing layer for a sessions-owned concept? No code change if the answer is the former.

*Verified compliant:* validation layer (`schemas.ts:39` matches sibling one-liners; handler validates → delegates, no business logic) · dependency rule (adapter imports no `services/`) · composition root (concrete impl referenced only in `compose.ts:108`) · required-not-optional deps (`podStatus` required, called unconditionally; the kill switch stays pod-side) · tests pass real port-shaped stubs without loosening a dep · module `index.ts` keeps port and adapter private.

Server state flows through TanStack Query with a key factory, derived state uses `useMemo`, the new component is a clean presentational leaf, and the stable-empty-array pattern matches the existing `EMPTY_AGENTS` convention. Three findings, all in the query hook.

**1. [HIGH] No `meta.errorToast` — the query is completely silent on error.**
`query-client.ts:36-46` short-circuits when `!query.meta?.errorToast`. Every comparable query sets it: `useApprovalsForOwner`/`useApprovalsForAgent` (`approvals/api/queries.ts:23,35`), `useAcpSessions` (`sessions/api/queries.ts:142` — the sibling in this very module). Silence may be intended for a decorative indicator, but nothing records that. Either add the meta, or state the deliberate deviation in a comment.

**2. [HIGH] Inherited `retry: 3` is wrong for this endpoint.**
The router throws `NOT_FOUND` when the agent is gone (`router.ts:103`) — permanent for that id, so retrying 3× per 30 s tick is 4 requests that cannot succeed. The closest analogues, also agent-scoped and pod-touching, both set `retry: false` (`agents/api/harness-config.ts:12,46`). It also works against the server's deliberate soft-fail (`agents-service.ts:256-258`).

**3. [MODERATE] `refetchInterval` is unconditional while the docstring claims otherwise.**
The hook says "Polls only while the agent is `running`" but sets a bare `refetchInterval` (`:32`), relying solely on `skipToken`. The effect is likely right today; the sibling gates the interval too — `refetchInterval: live ? STATUS_POLL_MS : false` (`sessions/api/queries.ts:140`). Mirroring it puts the invariant in both places so they can't drift.

*Verified compliant:* state lineage (no shadow copies) · key factory matches `acpSessionsKeys` · component 41 lines / ~14 JSX with interface-typed props, no `React.FC`, early `return null` · `useMemo` derivation matching the adjacent `approvalSessions` memo, zero `useEffect` · list-item rule (work passed as a prop to the existing `SessionRow`) · styling via `cn()`, no static `style={{}}`, established theme tokens, and `.working-dots-slow` is animation timing correctly placed beside the `.working-dots` rules it varies.

### Verdict
COMMENT — the passive-read design and its tests are solid; the two 🟡s are small client-side config fixes (`retry: false`, and either a toast or a comment recording the intentional silence), plus a one-row architecture-doc update.

---

packages/ui/src/modules/sessions/api/background-work.ts
```diff
@@ -0,0 +1,44 @@
...
+};
+
+/** Background work the agent's sessions report (#2965). Polls only while the
+ *  agent is `running` (the endpoint never wakes a pod); the same gate hides
+ *  cached data once it stops. Consumers share one query. */
+export function useAgentBackgroundWork(
+  agentId: string | null,
+): readonly SessionBackgroundWork[] {
+  const awake = useAgentRunState(agentId) === "running";
+  const { data } = useQuery({
```
dam-code-guardian, line 23: 🟡 **Warning:** no `meta.errorToast`, so this query fails **completely silently** — `query-client.ts:36-46` returns early when `!query.meta?.errorToast`. Every comparable query in the codebase sets one: `useApprovalsForOwner` / `useApprovalsForAgent` (`approvals/api/queries.ts:23,35`) and `useAcpSessions` (`sessions/api/queries.ts:142`, the sibling in this same module).

Silence is defensible here — the indicator is decorative and the server already folds pod unreachability into `[]`, so only genuine faults (deleted agent, auth, 5xx) reach the client. The issue is that nothing records the choice, leaving the next reader unable to tell a deliberate deviation from an oversight. Either add `meta: { errorToast: "Couldn't load background work" }`, or keep it silent with a one-line comment saying why.

packages/api-server/src/modules/agents/infrastructure/pod-status-client.ts
```diff
@@ -0,0 +1,43 @@
...
+        ),
+      }),
+    )
+    .catch([]),
+});
+
+// The pod answers from memory; this read never waits for a wake.
+const STATUS_TIMEOUT_MS = 3_000;
+
+export interface PodStatusClient {
```
dam-code-guardian, line 24: 🟢 **Suggestion:** this restates `PodStatusPort` (`agents-service.ts:41-43`) rather than implementing it — `createPodStatusClient` returns `PodStatusClient`, so nothing type-checks the adapter *against* the port. They satisfy each other only by coincidence of shape: widen a return or add a method and the error surfaces at the `compose.ts:108` wiring, pointing at the assignment instead of the drift.

This also departs from the module's own convention — every sibling adapter declares its port once and the service imports it (`agents-repository.ts:35` → `agents-service.ts:22`; likewise `AgentEnvRepository`, `AgentRegistrySecretPort`). The one port declared inside the service, `ContributionsSettledPort`, is the case whose adapter lives in another module.

Either drop this interface and import the port, or make the factory return type the port itself:

```suggestion
export function createPodStatusClient(namespace: string): PodStatusPort {
```

(with `PodStatusPort` imported from the service, matching how `compose.ts` already types the dep).

packages/ui/src/modules/sessions/api/background-work.ts
xjacka: Fixed in 747c8b50 — `retry: false` and the interval gated like the sibling sessions poll (`agentId && awake ? BACKGROUND_WORK_POLL_MS : false`).

packages/ui/src/modules/sessions/api/background-work.ts
xjacka: Kept silent deliberately — recorded with a comment in 747c8b50: the indicator is decorative and the server folds pod unreachability into `[]`, so a toast would only fire on faults other surfaces already report.

packages/api-server/src/modules/agents/infrastructure/pod-status-client.ts
xjacka: Fixed in 747c8b50 — the service now imports `PodStatusClient` from the adapter (matching the `AgentsRepository` convention), so there is a single declaration and the factory return type is checked against it.

xjacka:
Addressed the review @ fcd2273d:

- **retry / interval gating** — `retry: false`, `refetchInterval` gated on `agentId && awake` like the sibling sessions poll (747c8b50)
- **silent errors** — deliberate; now recorded with a comment at the query (747c8b50)
- **duplicated port** — single declaration: the service imports `PodStatusClient` from the adapter, matching the module's `AgentsRepository` convention (747c8b50)
- **doc drift** — `platform-topology.md` gains the status-read protocol row; `agent-lifecycle.md` now says the status surface is consumed by the UI indicator (no longer diagnosis-only), and the pod route's comment matches (fcd2273d)
- **placement (agents vs sessions)** — keeping the adapter in `agents/`: the pod status surface is agent-addressed (`podBaseUrl`, the `hibernated` gate), and the sessions surface consumes it through the public `agents.backgroundWork` query like any other agent read.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `fcd2273d`

## PR #3159: feat(ui): show running background work in a session
**Author:** xjacka | **Branch:** feat/background-work-indicator → main | **Changes:** +297 −10 (17 files)

### Summary
Re-review at `fcd2273d`: all three actionable findings from `5a3eb2c2` are fixed, the fourth (silent errors) is declined with its rationale now recorded in the code, and the doc drift is closed. No new findings.

### Changes since last review
Previous HEAD: `5a3eb2c2` (2026-08-04T13:20:38Z) — verdict COMMENT

- ✅ **Fixed:** `retry: false` — the inherited `retry: 3` (`query-client.ts:51`) no longer burns 4 requests per tick on a permanent `NOT_FOUND` (`background-work.ts:35`)
- ✅ **Fixed:** `refetchInterval` now gated on the same `agentId && awake` predicate as the `queryFn`'s `skipToken`, matching the `useAcpSessions` sibling (`background-work.ts:34`)
- ✅ **Fixed:** single port declaration — `PodStatusClient` lives only in the adapter and the factory is annotated with it, so the adapter is genuinely type-checked against the port (`pod-status-client.ts:26,32`)
- ✅ **Fixed:** doc drift — `platform-topology.md:87` gains the status-read protocol row; `agent-lifecycle.md:134-137` retracts "diagnosis-only"; both `Last verified` dates bumped
- 🔵 **Declined (accepted):** silent errors kept deliberately, with the reason recorded at `background-work.ts:26-27` — the original finding asked for either the meta or a comment recording the choice, and the comment is a valid answer
- 🔵 **Declined (accepted):** adapter stays in `agents/` — the read is agent-addressed (`podBaseUrl`, the `hibernated` gate), and the sessions surface consumes it through the public procedure

### Findings

**Verdict: aligned**

Both previously-reported items are resolved at `fcd2273d`, and the rest of the diff introduces no new drift. The new `api-server → agent-runtime | HTTP (status read)` row is phrased at the same semantic level as its siblings; `agent-lifecycle.md:182`'s "the controller reads nothing more into it" remains true, being scoped to the controller's idle check. Size headroom measured in the gate's own unit (`content.length`, `scripts/doc-size.mjs:39`): agent-lifecycle 34,288 / 40,000, platform-topology 14,867 / 40,000. Checks 3–7 produced nothing: no ADR leaks in the edited pages, no volatile content in the added lines, no new long-lived component, no architecture page moved, no ADR status change in the diff.

No findings. The port fix closes the exact mechanism the original finding named: one declaration (`pod-status-client.ts:26`) with the factory annotated `: PodStatusClient` (`:32`), so a widened return now fails at the adapter rather than at the `compose.ts:108` wiring. This matches where every sibling adapter declares its port (`AgentsRepository`, `AgentEnvRepository`, `AgentRegistrySecretPort`).

A service importing a *type* from `infrastructure/` is not a direction violation — the rule forbids importing a concrete implementation, and `createPodStatusClient` remains referenced only at the composition root.

*Verified compliant:* module public API (neither port nor factory exported; nothing outside the module references either) · dependency rule (no infrastructure→services imports) · validation layer (`agentBackgroundWorkInputSchema = idSchema`, router validates → delegates → maps `null` to `NOT_FOUND`) · required-not-optional deps (`podStatus` required at both sites, called unconditionally) · tests pass real port-shaped stubs, with the hibernation case asserting the probe was never *called*.

No findings. Both code fixes verified against the actual predicates rather than the author's summary: the overridden default is real (`query-client.ts:51` retries 3× with no error-type discrimination), and the silence is real (`query-client.ts:37-45` early-returns without `meta.errorToast`). The two gate expressions now use the *same* predicate — the natural bug for this fix to introduce, and it isn't present.

*Verified compliant:* state lineage (server state in TanStack Query with a key factory, no shadow copies, zero `useEffect`) · derived state via `useMemo` matching the adjacent `approvals` memo · one shared query behind both surfaces · 41-line presentational leaf with interface-typed props and early `return null` · frozen stable empties following the existing `EMPTY_AGENTS` convention · `cn()` styling with theme tokens, `.working-dots-slow` overriding only animation timing · row marker keeps the one-marker-per-row invariant and shares `backgroundWorkLabel` with the header so tooltips can't drift.

Note: `staleTime` is inherited (30 s, equal to the poll) rather than pinned like the sibling's `staleTime: STATUS_POLL_MS` — correct today, but coupled to the default by coincidence rather than intent. Not worth a finding.

### Verdict
APPROVE — every actionable finding is fixed, and the two declined ones carry recorded, defensible reasons.

---

PR #302, opened 2026-05-19
refactor: eliminate Instance, collapse into Agent (ADR-046)

jezekra1:
## Summary

- Eliminates the `Instance` concept; each `Agent` now owns its definition, runtime, and lifecycle.
- `Template`, `Agent`, and `Fork` are the only durable concepts in the agents bounded context.
- Single ConfigMap per agent (`agent-platform.ai/type=agent`); single ID prefix (`agent-`); single user-owned env layer.

Decision recorded in [ADR-046](docs/adrs/046-eliminate-instance.md); supersedes the seven Instance-vs-Agent decision points in ADRs 027/029/030/033/038/041, and rewrites the `DRAFT-multi-agent` in place.

## What changes

- **Contracts** (`packages/api-server-api`) — `Agent` absorbs `desiredState`, `secretRef`, `channels`, `allowedUserEmails`, `state`, `error`; `AgentsService` absorbs `restart` / `wake` / `ensureReady` / `connect*` / `disconnect*` / `isAllowedUser`. `instances` module deleted; tRPC `appRouter` no longer exposes `instances.*`.
- **DB** (`packages/db`) — migration `0013_eliminate_instance.sql` truncates affected tables, renames `instance_id` → `agent_id` everywhere, renames `instance_skills*` → `agent_skills*`, drops the redundant `pending_approvals.instance_id`.
- **api-server** (`packages/api-server`) — `modules/instances/` folded into `modules/agents/`; HTTP routes `/api/instances/:id/*` → `/api/agents/:id/*`; domain events renamed `InstanceX` → `AgentX`.
- **Controller (Go)** (`packages/controller`) — `InstanceSpec` merged into `AgentSpec`; reconciler watches `agent-platform.ai/type=agent`; Istio AuthorizationPolicy path matchers move to `/api/agents/<id>/*`; Fork derives from Agent via `ForkSpec.AgentName`; `LabelInstance` → `LabelAgent`; `ADK_INSTANCE_ID` → `PLATFORM_AGENT_ID`.
- **agent-runtime** (`packages/agent-runtime`) — reads `PLATFORM_AGENT_ID`; calls `/api/agents/:id/*` upstream.
- **UI** (`packages/ui`) — `modules/instances/` folded; `useCreateAgent` collapses the two-step `agents.create` + `instances.create` flow into a single mutation.
[...]

PR #3160, opened 2026-08-04
fix(controller): replace gateway pods stuck on a superseded revision

### LINKED ISSUE #2817: Reconnecting a connection can leave an agent without network access until a pod is deleted by hand
## Problem

Disconnecting a connection and immediately recreating it — the normal way to fix or refresh a credential — can permanently wedge the agent's network gateway. The gateway restart that follows the change can race the deletion of the old credential: the replacement gateway comes up referencing a credential that no longer exists, fails to start, and stays stuck indefinitely. The platform corrects its desired state within seconds, but the stuck pod is never replaced — the Kubernetes escape hatch it relies on for evicting a broken pod mid-rollout is an alpha feature that stock clusters don't enable.

From the user's side: the agent looks alive, but every outbound request fails to connect, forever. Nothing in the UI explains why beyond the agent showing "pods not ready", and the only fix is an operator deleting the stuck pod by hand — after which everything recovers instantly, because the corrected configuration was already in place.

Confirmed on a live e2e cluster (2026-07-17): the connection-regrant spec wedged the shared agent's gateway and the next spec needing the agent timed out; deleting the pod recovered it immediately. The e2e suite now runs that spec last as a workaround (part of the #2322 fix branch), but the underlying race is unfixed.

## Proposed solution

The platform should notice a gateway pod stuck starting on an outdated configuration — e.g. waiting on a credential its own desired state no longer references — and replace it automatically, exactly as the manual fix does. Reconnecting a connection should never cost an agent its network for longer than a normal restart.

### Out of scope

- Closing the race window itself (making the credential swap atomic with the gateway restart) — a bigger change, and automatic recovery removes the harm either way.

jjeliga:
Closes #2817

Disconnecting a connection and immediately recreating it — the normal way to refresh a credential — could permanently wedge an agent's network gateway. The agent looked alive, every outbound request failed to connect, and the only fix was an operator deleting the gateway pod by hand.

## Why it happened

Four things stack up. Only the last is wrong code.

1. **A stale reference gets applied.** A reconcile reads its render inputs (credential Secrets live, Agent spec from the informer cache) and applies the gateway StatefulSet ~12 API calls later. `deleteConnection` deletes the Secret *before* it patches the grants, so a roll already in flight can carry the reference past the deletion. Measured on a live cluster, that window is real: `credentials: 427ms`, `gatewayStatefulSet: 485–840ms`, ~1s per reconcile.
2. **The pod hard-fails.** Credential Secret volumes are mandatory mounts, so kubelet parks the pod in `ContainerCreating` with `FailedMount` forever. (`optional: true` would not help — Envoy treats a missing SDS file as a fatal boot error, trading `ContainerCreating` for `CrashLoopBackOff`.)
3. **Kubernetes won't replace it.** A pod that isn't Running-and-Ready trips the StatefulSet's OrderedReady monotonic invariant, which returns *before* the loop that replaces off-revision pods. So the corrected revision that lands seconds later never rolls out.
4. **The existing self-heal missed it.** `forceRollStuckPod` has existed since #302, but it only evicted pods sitting on `currentRevision` — and a wedged pod is on neither `currentRevision` nor `updateRevision`.

### The revision topology

This is the part that isn't guessable, and it's why the fix is two changes rather than one. Observed on a live cluster:

```
current=796b766947   update=796b766947   pod=6d64845c86   phase=Pending
```

`currentRevision` freezes at the last *completed* rollout, so it never advances to the bad revision. And when the corrected render is byte-identical to a template the StatefulSet already has a `ControllerRevision` for — the common case, since dropping the deleted credential reverts to the pre-grant template — **Kubernetes reuses that revision** instead of minting a new one. So `updateRevision == currentRevision` while the pod sits on an abandoned third revision.

That means the guard blocking recovery was the `updateRevision == currentRevision` early return, *not* only the predicate. Fixing the predicate alone would not have fixed the bug. A second topology (all three distinct) also occurs, when a replacement connection lands and hashes to a genuinely new revision.

## The fix

**Recovery** — `forceRollStuckPod` now evicts pods that are NotReady and **not on** `updateRevision`, and the early return on matching revisions is gone. Pods already on `updateRevision` are left alone whatever their state, so a genuinely unbootable spec fails in place instead of hot-looping pod creation; a Ready off-revision pod is still left to normal rolling update. Recovery costs a normal gateway restart.

**Diagnosis** — an unready gateway previously reported a bare `PodNotReady` with an empty message whatever the cause, so "permanently lost egress" was indistinguishable from "booting", and the api-server described it as *"the agent's gateway is still starting"* and retried forever. Now:
- a new `StuckOnSupersededRevision` reason on `GatewayPodReady`, plus the OOM / image-pull classification the agent pod already had (the gateway reported none of it);
- the api-server maps hard gateway causes to a `gateway-pod-failed` wake failure, so they are no longer classified as transient. Safe because classification only runs at the wake deadline — a self-healing wedge never reaches it.

Also corrected the `maxUnavailable` comment on the gateway StatefulSet, which credited a mechanism that does no work here: the field is alpha-gated (`MaxUnavailableStatefulSet`, off on stock clusters incl. k3s), defaults to 1, and governs the loop the monotonic invariant returns before reaching.

## Out of scope

The race itself is **not** closed — Secret deletion is still not atomic with the gateway roll, so the eviction, not the ordering, is what bounds the harm. Per the issue, automatic recovery removes the user-visible damage either way. Two follow-ups worth filing:

- turn credential deletion into *unreference-then-reap* (precedent already exists: orphan PVC / leaf-Secret sweeps), which would close the race rather than heal it;
- `forceRollStuckPod` runs for the gateway only; the agent StatefulSet can wedge the same way with no self-heal.

## Testing

**e2e (`gateway-full`, full tier)** — `packages/e2e/playwright/src/tests/full/gateway-wedge.spec.ts`. Creates a real agent, waits for a real gateway, points the pod template at a Secret that does not exist, asserts kubelet wedges it with `FailedMount`, then asserts the gateway returns to Ready **on a different revision** — replaced, not revived — and that the Agent's own `Ready` condition recovers.

The deadlock lives entirely in behavior a fake client cannot model (kubelet failing a mandatory mount, the monotonic invariant, revision reuse), which is why this has to be e2e. The race is not reproducible on demand, so the spec manufactures the state it produces; it parks the controller to stand in for the render window, which is cluster-wide and why it is full-tier and never overlaps another spec. Cleanup cannot throw, so a failure always reports the assertion rather than a teardown error.

Verified both directions on a live cluster:

| controller | result |
| --- | --- |
| with the fix | ✅ passes (21s / 24s) |
| predicate reverted, image rebuilt | ❌ `wedged gateway never recovered` (300s timeout) |

**Unit** — the two force-roll cases pin *different* guards, confirmed by restoring only the early return: the intermediate-revision case still passes, the reused-revision case fails. The e2e exercises only the second topology, and the full tier does not run in CI, so these are the always-on coverage. `TestGatewayNotReadyCause` and the `wake-failure` cases cover the diagnosis path, which has no e2e coverage (post-fix the stamp lives for under a second).

## Docs

`security-and-credentials.md` owns the behavior, added beside the existing render-time resilience paragraph — that one covers a credential already known bad when the gateway is rendered, this covers one revoked after. `agent-lifecycle.md` owns the wake-failure taxonomy, so the gateway cause splitting along the transient/hard line is noted there with a cross-link.

## Note for the reviewer

`mise run controller:check:crd-backwards-compatibility` fails on this branch, but it fails identically on a clean tree (`calculating hash for revision "v0.2.12-rc1": reference not found`, also for `userbudgets.yaml`, untouched here) — pre-existing, seemingly a worktree that cannot resolve the baseline revision's blobs. Everything else is green: controller tests + fmt/vet/staticcheck/build, api-server 756 tests + tsc/lint/format, e2e-playwright tsc/lint/format, doc-size.


packages/controller/pkg/reconciler/agent_reconciler.go
```diff
@@ -419,6 +426,29 @@ func (r *AgentReconciler) publishReconciled(ctx context.Context, agent *apiv1.Ag
...
+func (r *AgentReconciler) gatewayNotReadyCause(ctx context.Context, ssName string) (reason, message string) {
+	pod := r.getPod(ctx, ssName)
+	if pod == nil {
+		return "PodNotReady", ""
+	}
+	if reason, msg, ok := terminationReason(pod); ok {
+		return reason, msg
+	}
+	ss, err := r.client.AppsV1().StatefulSets(r.config.Namespace).Get(ctx, ssName, metav1.GetOptions{})
+	if err != nil || ss.Status.UpdateRevision == "" {
```
dam-code-guardian: 🟡 **Warning:** this re-reads the StatefulSet, so the revision it judges against is not the one `podCurrentAndReady` just used. `publishReadiness` computes `gatewayReady` from a read at `agent_reconciler.go:388`, then this runs ~1 call later; if a reconcile lands a new template in between, `UpdateRevision` advances and a pod that is simply mid-roll — NotReady because it is still booting the *newest* revision — gets stamped `StuckOnSupersededRevision`.

That reason is the one the api-server treats as hard and non-transient (`wake-failure.ts:100-113`), so a wake landing in that window fails fast with "stuck on an outdated configuration" instead of waiting out a legitimate boot. The pod informer re-enqueues on the next transition and the stamp clears within a second, which is why the PR body's note about the stamp being short-lived holds — but a wake deadline can land inside it, and the fast-fail is exactly what this PR made possible.

Note `podCurrentAndReady` also gates on `ObservedGeneration != Generation` while this helper does not, so the helper can judge a template the StatefulSet controller has not observed yet. Passing the already-fetched StatefulSet (or the `UpdateRevision` string) down from `publishReadiness` would make the readiness bit and its reason derive from one observation, and skipping the stamp when `ObservedGeneration != Generation` would close the mid-roll case.

packages/e2e/playwright/src/lib/cluster.ts
```diff
@@ -0,0 +1,124 @@
+import { execFileSync } from "node:child_process";
+import { homedir } from "node:os";
+import { join } from "node:path";
+
+/** Namespace the controller renders agent + gateway workloads into. */
+export const AGENT_NS = "platform-agents";
+/** Namespace the platform components run in. */
+const RELEASE_NS = "default";
```
dam-code-guardian: 🟢 **Suggestion:** `RELEASE_NS` and the literal Deployment name at line 109 are correct only for a release named `platform` installed into `default` — the chart renders it as `{{ include "platform.fullname" . }}-controller`, i.e. `<release>-controller`. Since `kubeconfig()` already parameterizes the VM name and kubeconfig path via env, these two are the outliers in an otherwise configurable helper, and `scaleController` uses the throwing `kubectl` so a mismatch surfaces as a teardown-time failure rather than a clear skip.

An env fallback in the same style as the rest of the file would keep the helper portable across installs.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `3697141`

## PR #3160: fix(controller): replace gateway pods stuck on a superseded revision
**Author:** jjeliga | **Branch:** fix/gateway-superseded-revision-selfheal → main | **Changes:** +615 −28 (13 files)

### Summary
Widens the gateway self-heal from "evict a NotReady pod on `currentRevision`" to "evict any NotReady pod not on `updateRevision`", drops the `updateRevision == currentRevision` early return that was the actual blocker, and adds a `StuckOnSupersededRevision` readiness reason the api-server classifies as a hard wake failure.

### Findings
- 🟡 **Warning:** `gatewayNotReadyCause` re-reads the StatefulSet, so its revision verdict can disagree with the readiness bit computed moments earlier (`packages/controller/pkg/reconciler/agent_reconciler.go:441`)
- 🟡 **Warning:** channels doc bullet still lists the hard causes without the gateway one, and its transient/hard framing no longer covers a hard-but-self-healing cause (`docs/architecture/channels.md:170`)
- 🟢 **Suggestion:** `agent.wake.timeout` log omits the new `gatewayPodNotReadyReason` while carrying its agent-pod counterpart (`packages/api-server/src/modules/agents/infrastructure/agents-repository.ts:452`)
- 🟢 **Suggestion:** `cluster.ts` hard-codes release namespace + controller Deployment name while parameterizing everything else via env (`packages/e2e/playwright/src/lib/cluster.ts:8`)
- ✅ **Looks good:** the eviction predicate inverts safely — `continue` on `== updateRevision` leaves a pod on the target revision alone even when its revision label is empty, so an unbootable spec fails in place instead of hot-looping, exactly as the docstring claims.

**Verdict:** minor drift

**Drift — Check 1, architecture-page drift: `channels.md` not updated**

The diff changes behavior the channels page describes. `wake-failure-copy.ts:54-79` adds channel copy whose `StuckOnSupersededRevision` case tells the sender to retry in a minute because the condition self-heals; `wake-failure.ts:100-113` classifies that cause **non-transient** (pinned by `wake-failure.test.ts:178-183`). Because it is non-transient, the Slack relay's transient branch (`slack.ts:1159-1163`, ambient at `1764-1768`) re-throws instead of posting a still-starting note and waiting another window — so a self-healing wedge takes the hard-failure delivery path while carrying retry copy.

`docs/architecture/channels.md:170` says: _"A hard failure (pod crash, bad image, reconcile error) replies with copy derived from the classified wake-failure cause…"_ — the parenthetical omits the new gateway cause, and the transient/hard framing implies hard ⇒ needs intervention, whereas this cause is hard _and_ self-repairing. Proposed edit:

> A hard failure (pod crash, bad image, a gateway wedged on a superseded configuration, reconcile error) replies with copy derived from the classified wake-failure cause — never the internal error string, and never raw controller messages. A hard cause that the platform is already repairing still invites a retry, so the copy tracks whether waiting will help rather than the transient/hard split alone.

`agent-lifecycle.md:89` — which the PR did update — states the taxonomy correctly and calls out the split, so channels is the only page left describing the old shape.

**Possible drift — `platform-topology.md:37`:** the page enumerates the controller's responsibilities (_"reconciles the StatefulSet, Service, NetworkPolicy… computes agent readiness… hibernates idle agents"_). The diff widens `forceRollStuckPod` into the general unwedging mechanism and `gateway.go:92-99` explicitly transfers ownership of that job to it. Is "evicts pods stranded on a superseded revision" a distinct controller responsibility for this list, or an implementation detail of "reconciles the StatefulSet" that `security-and-credentials.md` already covers? A reader of the topology page would not learn the controller ever deletes a pod outside hibernation.

Both edited pages carry `Last verified: 2026-08-04` (= PR date), contain no ADR references, add no volatile content, and sit well under the 40k cap; no page was moved.

No layering, boundary, or dependency-direction violations. Notes:

- **Domain purity held in the awkward direction.** `wake-failure.ts` adds the variant, the reason set, and the classifier branch with zero imports; `WakeConditionsSnapshot` takes `gatewayPodNotReadyReason` structurally so the domain never imports the infrastructure type (its docstring says so at line 37-39).
- **Cross-module import is via the boundary.** `wake-failure-copy.ts:1` imports `WakeFailureCause` from `agents/index.js`, re-exported at `agents/index.ts:36` — never `agents/domain/...`. Taxonomy in agents-domain, channel copy in the channels module, is the correct split.
- **Exhaustiveness discharged.** Both switches over the union were updated (`describeWakeFailure`, `wakeFailureUserCopy`); no third `.kind` switch exists — other consumers call the helpers. `isTransientWakeFailure` is a transient allowlist, so a new variant defaults to non-transient: fail-safe here.
- **Finding:** `agents-repository.ts:452-461` logs `gatewayPodReady` and `agentPodNotReadyReason` but not the new `gatewayPodNotReadyReason`, making the gateway the only asymmetric condition/reason pair. `cause` carries it indirectly, so nothing is lost — one line restores the symmetry.
- **Finding:** `cluster.ts` is the only e2e file shelling out to `child_process`; placement in `lib/` is right. `RELEASE_NS = "default"` + literal `"platform-controller"` are correct only for a release named `platform` in `default` (chart renders `<release>-controller`), while `kubeconfig()` parameterizes VM name and path via env.

### Verdict
COMMENT — the fix is well-targeted and the two-part diagnosis is convincing; the revision-mismatch reason can be stamped on a pod that is merely mid-roll, and one doc page still describes the old failure taxonomy.

---

packages/e2e/playwright/src/lib/cluster.ts
```diff
@@ -0,0 +1,121 @@
+import { execFileSync } from "node:child_process";
+import { homedir } from "node:os";
+import { join } from "node:path";
+
+/** Namespace the controller renders agent + gateway workloads into. */
+export const AGENT_NS = process.env.E2E_AGENT_NS ?? "platform-agents";
+/** Namespace the platform components run in. */
+const RELEASE_NS = process.env.E2E_RELEASE_NS ?? "default";
+/** Helm release name; the chart renders workloads as `<release>-<component>`. */
+const RELEASE = process.env.E2E_RELEASE_NAME ?? "platform";
```
dam-code-guardian, line 12: 🟡 **Warning:** the docstring's `<release>-<component>` formula holds only when the release name already contains `platform`. The Deployment is named `{{ include "platform.fullname" . }}-controller` (`deploy/helm/platform/templates/controller/deployment.yaml:7`), and `platform.fullname` prepends the chart name unless `contains $name .Release.Name` — so `E2E_RELEASE_NAME=platform-ci` resolves correctly to `platform-ci-controller`, but `E2E_RELEASE_NAME=dam` computes `dam-controller` while the chart renders `dam-platform-controller`.

Because `scaleController` uses the throwing `kubectl`, a mismatch fails at park time — and the spec's `finally` calls `scaleController(1)`, which throws for the same reason, so the restore is skipped and the controller stays parked cluster-wide. That is the exact outcome the teardown comment ("leaving it parked breaks every later spec") exists to prevent, and it would surface as every subsequent spec failing rather than as a clear error here.

The default install is unaffected (`deploy/tasks.toml:568` installs the release as `platform`), so this only costs the portability the change was made for. Deriving the workload prefix the way the chart does — or naming the env var after the rendered prefix rather than the release — would keep them in step.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `bfd7e8a`

## PR #3160: fix(controller): replace gateway pods stuck on a superseded revision
**Author:** jjeliga | **Branch:** fix/gateway-superseded-revision-selfheal → main | **Changes:** +631 −38 (16 files)

### Summary
Round 2 answers all four prior findings by extracting `podStuckOnSupersededRevision` so the evictor and the status reporter share one predicate, adding the observed-generation guard, and parameterizing the e2e cluster helper. One new finding: the e2e helper's release-name formula omits the chart-name prefix Helm adds.

### Changes since last review
Previous HEAD: 3697141 (2026-08-04T15:40:47Z) — verdict COMMENT

- ✅ **Fixed:** revision verdict could disagree with the readiness bit — one shared predicate now backs both, gated on `ObservedGeneration == Generation` and false for a Ready off-revision pod (`packages/controller/pkg/reconciler/agent_reconciler.go:432`)
- ✅ **Fixed:** `channels.md` hard-cause list and transient/hard framing updated; `platform-topology.md` records the eviction responsibility (`docs/architecture/channels.md:170`)
- ✅ **Fixed:** `agent.wake.timeout` now logs `gatewayPodNotReadyReason` (`packages/api-server/src/modules/agents/infrastructure/agents-repository.ts:459`)
- ✅ **Fixed:** `cluster.ts` reads namespaces and release name from `E2E_*` env vars (`packages/e2e/playwright/src/lib/cluster.ts:8`)
- 🆕 **New:** e2e release-name formula omits the chart-name prefix `platform.fullname` adds (`packages/e2e/playwright/src/lib/cluster.ts:10`)

### Findings
- 🟡 **Warning:** `${RELEASE}-controller` only matches the rendered Deployment when the release name contains `platform`; a park failure then throws again in the `finally`, leaving the controller parked cluster-wide (`packages/e2e/playwright/src/lib/cluster.ts:10`)

**Verdict:** aligned

Both items from the review at `3697141` are resolved, and no new drift was found.

- Check 1: `channels.md:170` now names the wedged gateway among the hard causes and states that a self-repairing cause still invites a retry — matching `wake-failure-copy.ts:56-62` (retry copy) and the non-transient classification at `wake-failure.ts:134-140`. `platform-topology.md:37` now carries the eviction as a controller responsibility; verified as the only `Pods().Delete` in the controller (`agent_reconciler.go:934`), so "the one case where it evicts a pod outside hibernation" holds. `agent-lifecycle.md:89`'s claim that classification runs only once the budget is spent is accurate for gateway causes — the pre-deadline `classifyWakeFailure` call (`agents-repository.ts:422`) is gated on `overBudget`, which `classifyWakeFailure` resolves first.
- Check 2: all four edited pages bumped `Last verified:` to `2026-08-05`, the HEAD commit date.
- Checks 3–7: no ADR references or `Motivated by:` sections added; added prose stays at concept level (no field names, paths, or type shapes); no new subsystem; no page moved or renamed; no ADR status changed in the diff.

Note (not drift): `channels.md` is at 39,769 of the 40,000-char page cap.

Layering and boundaries are clean; both prior findings are genuinely fixed.

- **Prior finding 1 resolved, and correctly wired.** `agents-repository.ts:459` now logs `gatewayPodNotReadyReason` beside `gatewayPodReady`, restoring the condition/reason symmetry. Verified end-to-end: `agent-mappers.ts:221-222` populates it from the `GatewayPodReady` condition's reason only when the condition is `False`, so the log field is `undefined` exactly when there is no reason to report.
- **Domain purity still held.** `wake-failure.ts` remains import-free; this round's docstring rewrites removed the `#2817` references without touching the classifier. `wake-failure-copy.ts:1` still imports `WakeFailureCause` through `agents/index.js` (re-exported at `agents/index.ts:36`), never `agents/domain/...`.
- **Prior finding 2 partly resolved** — the parameterization matches the `E2E_`-prefixed convention already used for `E2E_VM_NAME` and `E2E_AGENT_NS`, but the formula is wrong; see the finding above.

### Verdict
COMMENT — the shared-predicate refactor is the right fix and closes the mid-roll misdiagnosis cleanly; the one new finding is confined to e2e portability that only bites a non-default release name.

---

JanPokorny (approved):

JanPokorny (approved):

Issue #830, opened 2026-06-11
Chat UI overflows the screen on mobile — the whole page scrolls


jezekra1:
## Problem

On mobile the chat UI still doesn't fit the display: the page is taller than the visible viewport, so the whole page scrolls and parts of the UI sit off-screen. Chatting on a phone means constantly fighting the scroll position. Earlier mobile-height fixes (#35, #54) didn't fully resolve it.

## Proposed solution

On mobile the chat fills exactly the visible screen — header and input bar stay fixed, only the message list scrolls. Showing or hiding the browser address bar or the on-screen keyboard must not push the layout off-screen.

PR #3161, opened 2026-08-04
fix(ui): keep settings tab strip scrolling inside itself on mobile

### LINKED ISSUE #2907: Settings navigation forces horizontal page-scrolling on mobile
## Problem

On a phone, the Settings navigation is a single horizontal row of tabs that's wider than the screen. Only the first few (Account, Appearance, Providers) are visible; the rest — Connections, API keys, Usage, and, when enabled, Channels and Experimental features — run off the right edge.

They *are* reachable, but only by scrolling the **whole view** sideways: the entire page shifts horizontally rather than the tab strip scrolling on its own. That's awkward, easy to miss, and makes the content jump around while you hunt for a section. A settings menu shouldn't require side-scrolling the page to use.

## Proposed solution

Make the Settings navigation handle small screens gracefully so every section is reachable **without** side-scrolling the whole page — for example by letting the tab strip scroll horizontally on its own, wrap onto multiple lines, or collapse into a compact dropdown/picker. The exact treatment is a design call; the requirement is that reaching a section never moves the rest of the page.

### Out of scope
- The desktop Settings layout (the vertical tab sidebar), which already works.

Tomas2D:
## Summary

On a phone the Settings navigation was reachable only by side-scrolling the **whole view**, so hunting for a section made the content jump around.

Two things stacked to cause it:

1. The tab strip is a `flex-row` on mobile with `overflow: visible`. As a flex item of the column-direction wrapper its width is stretch-clamped to the content column (343px at a 375px viewport), so the triggers did not widen the strip — they spilled out of it.
2. The nearest ancestor that establishes a scroll box is `<main className="… overflow-y-auto">` in `app.tsx`. A box with `overflow-y: auto` and `overflow-x: visible` computes `overflow-x` to `auto`, so `main` became a horizontal scroller and the spilled triggers panned the entire view. `document.documentElement` never scrolled, which matches the report that the whole page shifts while the rail stays put.

The fix gives the strip its own overflow, in one class list on the one call site:

- `overflow-x-auto` — the strip owns its horizontal overflow instead of leaking it to `main`.
- `[&>button]:shrink-0` — required, not decorative. Without it the triggers shrink to min-content and render two-line wrapped labels inside the scroller. Scoped to this call site (matching the existing `[&_svg]:shrink-0` convention in `button.tsx`) rather than added to `tabTrigger`, where it would push the horizontal strips inside modals to overflow their dialogs.
- `py-1` — `ring-2` + `ring-offset-2` extends 4px outside the trigger, so without the padding a keyboard focus ring is clipped by the new scroll box. Rings are box-shadows and outlines, so this adds no phantom scroll area.
- `md:*` resets restore the desktop sidebar exactly.

Scroll was chosen over wrapping or a dropdown because it satisfies the issue's hard requirement (reaching a section never moves the rest of the page) without needing a design decision. The shared `Tabs` component is untouched, so the other three tab call sites cannot regress.

Closes #2907

## Verification

Measured in headless Chromium at 375×812 against the real compiled Tailwind CSS, in a fixture reproducing the app's container chain:

| | before | after |
|---|---|---|
| `main.scrollWidth` vs `clientWidth` | 625 > 375 — page scrolls | 375 === 375 — no page scroll |
| strip `scrollWidth` vs `clientWidth` | — | 691 > 343 — strip scrolls alone |
| trigger height | 59px (labels wrapped to two lines) | 38px (single line) |

At 1280px the strip's computed style is byte-identical before and after (`flex-direction: column`, width 180, `overflow-x: visible`, nothing scrolls), so the desktop vertical sidebar — explicitly out of scope in the issue — is provably unchanged.

- [x] `mise run ui:check` — tsc, eslint, prettier clean (the 2 remaining warnings are pre-existing `react-hooks/exhaustive-deps` in files this PR does not touch)

## Test plan

- [ ] **Real device pass** — the measurements above are headless Chromium only; iOS Safari and Android Chrome are unverified. Specifically worth checking that the horizontal scroll affordance is discoverable enough with overlay scrollbars, and that momentum scrolling the strip does not fight vertical page scroll.
- [ ] Confirm every section (Connections, API keys, Usage, and Channels / Experimental features when enabled) is reachable without the page moving
- [ ] Confirm the desktop vertical sidebar is unchanged past `md`
- [ ] Confirm keyboard focus rings are not clipped when tabbing along the strip

## Not addressed

- **Deep-link case.** Landing directly on e.g. `/settings/usage` on a phone leaves the active pill scrolled out of sight; the strip does not auto-scroll it into view. Not a regression — it was equally off-screen before — and fixing it means an effect in the shared `Tabs`, where a naive `scrollIntoView` would walk up to `main` and yank the page vertically. Better as its own change.
- **#830** (chat UI overflowing on mobile) does **not** share this root cause. That one is a vertical viewport problem — `h-dvh` in `chat-view.tsx` and `app.tsx` versus the visual viewport when the address bar or on-screen keyboard appears, with no `window.visualViewport` usage anywhere in the package. Different axis, different mechanism; this fix does nothing for it.
- **No automated test.** The UI package has no DOM or component tests (all unit suites are logic-only), and asserting Tailwind layout needs a real browser.

### DIFF
--- a/packages/ui/src/modules/settings/views/settings-view.tsx
+++ b/packages/ui/src/modules/settings/views/settings-view.tsx
@@ -100,11 +100,18 @@ export function SettingsView() {
         size="sm"
         orientation="vertical"
         // Mobile lays the strip out as a row. It owns its horizontal overflow
-        // so a long tab set scrolls the strip instead of the whole view, the
-        // triggers keep their full label instead of shrinking into wraps, and
-        // the vertical padding leaves room for a focus ring inside the scroll
-        // box. Desktop stays the plain vertical sidebar.
-        className="flex-row overflow-x-auto py-1 [&>button]:shrink-0 md:flex-col md:overflow-x-visible md:py-0 md:w-[180px] shrink-0"
+        // so a long tab set scrolls the strip instead of the whole view, and
+        // the triggers keep their full label instead of shrinking into wraps.
+        // Specifying `overflow-x` makes the strip a scroll container on both
+        // axes, so its clipping region is its padding box — and the triggers'
+        // focus ring paints 4px outside their border box on all four sides.
+        // Hence padding on both axes: without it the first and last triggers'
+        // rings are clipped, and the first tab is the default focus target.
+        // The inline padding clears 4px rather than matching it — fractional
+        // layout leaves the scroll-end edge ~0.06px short of an exact 4px.
+        // `gap-1` already spaces the interior rings. Desktop stays the plain
+        // vertical sidebar.
+        className="flex-row overflow-x-auto px-1.5 py-1 [&>button]:shrink-0 md:flex-col md:overflow-x-visible md:px-0 md:py-0 md:w-[180px] shrink-0"
       />
 
       <div className="flex-1 min-w-0">
--- a/packages/ui/src/modules/settings/views/settings-view.tsx
+++ b/packages/ui/src/modules/settings/views/settings-view.tsx
@@ -99,7 +99,12 @@ export function SettingsView() {
         variant="pill"
         size="sm"
         orientation="vertical"
-        className="flex-row md:flex-col md:w-[180px] shrink-0"
+        // Mobile lays the strip out as a row. It owns its horizontal overflow
+        // so a long tab set scrolls the strip instead of the whole view, the
+        // triggers keep their full label instead of shrinking into wraps, and
+        // the vertical padding leaves room for a focus ring inside the scroll
+        // box. Desktop stays the plain vertical sidebar.
+        className="flex-row overflow-x-auto py-1 [&>button]:shrink-0 md:flex-col md:overflow-x-visible md:py-0 md:w-[180px] shrink-0"
       />
 
       <div className="flex-1 min-w-0">


packages/ui/src/modules/settings/views/settings-view.tsx
```diff
@@ -99,7 +99,12 @@ export function SettingsView() {
         variant="pill"
         size="sm"
         orientation="vertical"
-        className="flex-row md:flex-col md:w-[180px] shrink-0"
+        // Mobile lays the strip out as a row. It owns its horizontal overflow
+        // so a long tab set scrolls the strip instead of the whole view, the
+        // triggers keep their full label instead of shrinking into wraps, and
+        // the vertical padding leaves room for a focus ring inside the scroll
+        // box. Desktop stays the plain vertical sidebar.
+        className="flex-row overflow-x-auto py-1 [&>button]:shrink-0 md:flex-col md:overflow-x-visible md:py-0 md:w-[180px] shrink-0"
```
dam-code-guardian: 🟡 **Warning:** The focus-ring accommodation is complete on the block axis but not the inline axis, so the ring is clipped on the first and last triggers.

`py-1` gives the strip 4px of block padding and no inline padding. Two things make that asymmetric:

1. **The strip is a scroll container on both axes.** `overflow-x: auto` is specified and `overflow-y` is `visible`, so by the same computed-value rule the PR body correctly cites for `main` — [css-overflow-3](https://drafts.csswg.org/css-overflow-3/#overflow-properties): *"if the other axis specifies a scrollable value, a specified value of `visible` computes to `auto`"* — `overflow-y` computes to `auto` too. Its clipping region is therefore its **padding box**.
2. **The ring extends 4px on all four sides.** From Tailwind v4's source, `ring-2` emits spread `calc(2px + var(--tw-ring-offset-width))` = 4px and `ring-offset-2` sets that width to 2px, so `tabTrigger`'s `focus-visible:ring-2 ring-offset-2` paints 4px outside the trigger's border box horizontally as well as vertically.

With `padding-inline: 0`, the padding box's inline-start edge coincides with the first trigger's border-box edge (and its inline-end edge with the last trigger's), so the outer 4px of those two rings is cut off. Scrolling doesn't recover it: `box-shadow` is not part of the scrollable overflow area (the spec's union is padding box + line boxes + descendant border boxes + flex/grid item margin areas), and on the start side it would land in the unreachable scrollable overflow region anyway.

It lands on the likeliest trigger: `Tabs` sets `tabStop` to the active value or `selectable[0]`, and `account` is both the first tab and the default, so tabbing into Settings focuses the one trigger whose ring is clipped.

The same one-call-site fix the PR already uses for the block axis works here — `gap-1` between triggers means the interior rings have room, so only the outer edges need it:

```suggestion
        className="flex-row overflow-x-auto px-1 py-1 [&>button]:shrink-0 md:flex-col md:overflow-x-visible md:px-0 md:py-0 md:w-[180px] shrink-0"
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `3c15fbe0`

## PR #3161: fix(ui): keep settings tab strip scrolling inside itself on mobile
**Author:** Tomas2D | **Branch:** `fix/settings-nav-mobile-scroll` → `main` | **Changes:** +6 −1 (1 file)

### Summary
Gives the mobile Settings tab strip its own `overflow-x-auto` so a long tab set scrolls the strip instead of leaking horizontal overflow up to `<main className="overflow-y-auto">`, which was becoming a page-wide horizontal scroller. The root-cause analysis in the body is correct and verified against the CSS spec; the fix is scoped to the one call site with the shared `Tabs` component untouched.

### Findings
- 🟡 **Warning:** The focus-ring fix is complete vertically but not horizontally — the first and last triggers' rings are clipped (`packages/ui/src/modules/settings/views/settings-view.tsx:107`)
- ✅ **Looks good:** The `[&>button]:shrink-0` scoping decision is right and its stated reason checks out — `tabTrigger` is shared by `connection-catalog-modal.tsx:132` (`w-[200px]` inside a dialog), `add-skill-source-modal.tsx:66`, and `providers/anthropic/form.tsx:221`, so widening the `cva` really would have pushed those strips to overflow their dialogs. `md:overflow-x-visible` is a real Tailwind v4 utility (`overflow-x-{auto,hidden,clip,visible,scroll}` are all generated), `className` reaches the tablist through `cn()` so `tailwind-merge` resolves the `md:` resets cleanly, and `[&>button]:` matches the existing `[&_svg]:shrink-0` convention in `button.tsx:10`.

No violations. Tailwind-only, no inline `style`, no template-literal class strings, conditional composition left to the component's own `cn()`. Class ordering follows the informal layout → spacing → responsive convention. The touch-it/migrate-it rule does not bite — the file is 228 lines and the change adds no state or logic.

One note rather than a finding: the arbitrary variant `[&>button]:shrink-0` reaches through `Tabs` to style its internal `<button>` children, coupling the call site to that component's DOM shape. Acceptable here — it is the documented alternative to widening the shared `cva`, and `button.tsx` sets the same precedent — but it would stop applying silently if `Tabs` ever wrapped its triggers in an element.

`aligned` — a CSS-class change to one call site alters no subsystem's behavior or responsibility; no page under `docs/architecture/` describes the settings strip's layout (`features.md` only references the module path), and no ADR changed state.

### Verdict
COMMENT — the diagnosis and the fix are sound, and the measurements in the body hold up under independent checking. One gap: the new scroll box clips the focus ring on the inline axis, on exactly the trigger a keyboard user reaches first. `px-1 md:px-0` alongside the existing `py-1 md:py-0` closes it.

---

JanPokorny (approved):

Issue #928, opened 2026-06-15
[Bug]: broken links


matoushavlena:
link-guardian found broken links in this repo's documentation. Each link below has been broken on at least two consecutive runs.

Close this issue once the links are fixed (or if you've decided they're not worth fixing) and link-guardian will stop reporting them.

**`.agents/skills/doc-drift/SKILL.md`** _(1 broken)_

- [x] Line 33: `../../../tseng/vocabulary.md` — file not found: ../../../tseng/vocabulary.md

**`docs/adrs/050-platform-reserved-paths.md`** _(1 broken)_

- [x] Line 9: `044-file-import.md` — file not found: 044-file-import.md

**`docs/guidelines/documentation-guidelines.md`** _(1 broken)_

- [x] Line 17: `../../tseng/vocabulary.md` — file not found: ../../tseng/vocabulary.md

**`docs/guidelines/release-process.md`** _(1 broken)_

- [x] Line 41: `.github/workflows/cd.yml` — file not found: .github/workflows/cd.yml

<details>
<summary>**`docs/notes/openshell-credential-threat-model.md`** _(17 broken)_</summary>

- [ ] Line 7: `../../context/OpenShell/` — file not found: ../../context/OpenShell/
- [ ] Line 22: `../../context/OpenShell/crates/openshell-server/src/grpc/policy.rs#L608-L644` — file not found: ../../context/OpenShell/crates/openshell-server/src/grpc/policy.rs
- [ ] Line 23: `../../context/OpenShell/crates/openshell-sandbox/src/secrets.rs#L153-L165` — file not found: ../../context/OpenShell/crates/openshell-sandbox/src/secrets.rs
- [ ] Line 33: `../../context/OpenShell/crates/openshell-sandbox/src/process.rs#L27-L30` — file not found: ../../context/OpenShell/crates/openshell-sandbox/src/process.rs
- [ ] Line 33: `../../context/OpenShell/crates/openshell-sandbox/src/process.rs#L822-L840` — file not found: ../../context/OpenShell/crates/openshell-sandbox/src/process.rs
- [ ] Line 34: `../../context/OpenShell/crates/openshell-sandbox/src/secrets.rs#L205-L245` — file not found: ../../context/OpenShell/crates/openshell-sandbox/src/secrets.rs
[...]

PR #3162, opened 2026-08-04
docs(guidelines): fix cd.yml link in release-process

Tomas2D:
## Summary

One broken link from the link-guardian report in #928.

`docs/guidelines/release-process.md` line 41 linked `.github/workflows/cd.yml`, which resolves relative to `docs/guidelines/` rather than the repo root. Line 64 of the same file already linked the identical target correctly as `../../.github/workflows/cd.yml`; line 41 now matches it. Link text is unchanged.

```diff
-2. The `v*` tag triggers the release jobs in [`cd.yml`](.github/workflows/cd.yml): …
+2. The `v*` tag triggers the release jobs in [`cd.yml`](../../.github/workflows/cd.yml): …
```

**Refs #928 — this does not close it.** Three groups from that report are deliberately left alone, each because it needs a decision or is blocked:

1. **`tseng/vocabulary.md`** in `docs/guidelines/documentation-guidelines.md:17` and `.agents/skills/doc-drift/SKILL.md:33`. There is no `vocabulary.md` and no `tseng/` directory anywhere in this repo — the target lives outside it. Fixing these is a call between de-linking and pointing at wherever the vocabulary actually lives.
2. **`docs/adrs/050-platform-reserved-paths.md`** references `044-file-import.md` when the real file is `045-file-import.md`. Obvious as the fix looks, `scripts/adr-immutable.mjs` freezes the prose body of every accepted ADR — only frontmatter is mutable — so editing it would fail that gate.
3. **The 17 links in `docs/notes/openshell-credential-threat-model.md`** pointing into `context/OpenShell/`, an external repo not present here. Another decision rather than a fix.

## Verification

The repo has no in-repo link checker (`mise tasks` exposes only `docs:check:{adr-immutable,adr-index,doc-size}`, and `scripts/` has no link script), so resolution was checked directly from the containing directory:

- old form — `ls docs/guidelines/.github/workflows/cd.yml` → no such file
- new form — `ls docs/guidelines/../../.github/workflows/cd.yml` → resolves, 43444 bytes

Swept for other links of the same shape and found none: targets beginning `.github/`, `packages/`, `deploy/`, or `scripts/` (exhaustive — those plus `docs/` are the repo's only top-level directories), `](docs/…` from inside `docs/`, root-absolute `](/…`, and reference-style `[label]: path` definitions (the one match is a `[^1]:` footnote, not a link).

- [x] All three `docs:check:*` tasks pass, including `adr-immutable` reporting no accepted ADR body was rewritten
- [x] Remaining check groups re-run explicitly — 37 tasks, exit 0

## Test plan

- [ ] Click through the two `cd.yml` links in `docs/guidelines/release-process.md` on the rendered PR page and confirm both resolve

### DIFF
--- a/docs/guidelines/release-process.md
+++ b/docs/guidelines/release-process.md
@@ -38,7 +38,7 @@ To release a different version than what's on main, run `mise run release:set-ve
 Run from the release branch. The task:
 
 1. Tags the current commit (e.g. `v0.3.0-rc1`) and pushes tag + branch atomically.
-2. The `v*` tag triggers the release jobs in [`cd.yml`](.github/workflows/cd.yml): images are built, Helm chart is pushed, CLI is published to npm with `--tag rc`.
+2. The `v*` tag triggers the release jobs in [`cd.yml`](../../.github/workflows/cd.yml): images are built, Helm chart is pushed, CLI is published to npm with `--tag rc`.
 3. Bumps the branch to the next RC (`0.3.0-rc2`).
 
 Repeat as needed — cherry-pick fixes onto the release branch and publish another RC.


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `247ca398`

## PR #3162: docs(guidelines): fix cd.yml link in release-process
**Author:** Tomas2D | **Branch:** `docs/fix-repo-relative-links` → `main` | **Changes:** +1 −1 (1 file)

### Summary
Fixes one broken relative link in `docs/guidelines/release-process.md:41` — `.github/workflows/cd.yml` → `../../.github/workflows/cd.yml`, matching how line 64 of the same file already links the identical target. Verified: the target resolves from the containing directory (43,444 bytes), and lines 41 and 64 are now byte-identical in target.

### Findings
- ✅ **Looks good:** Every factual claim in the PR body verified independently — the two `cd.yml` links now agree, no in-repo link checker exists (`tasks.toml` exposes only `docs:check:{adr-index,doc-size,adr-immutable}`, `scripts/` has no link script), `tseng/vocabulary.md` and `tseng/` are genuinely absent repo-wide, `045-file-import.md` is the real filename ADR-050 misnames as `044-`, and `scripts/adr-immutable.mjs` really does hash the whole prose body (`bodyHash` over `splitFrontmatter(raw).body`) with no allowlist or bypass — so deferring group 2 is correct, not an excuse.
- 🟢 **Suggestion:** The sweep for same-shape links stopped at the repo's non-dot top-level directories, but this exact defect shape survives elsewhere: `CLAUDE.md:13` links `` `.claude/skills/cluster-ops/SKILL.md` `` as a bare `.claude/…` target, and five further `CLAUDE.md`/`README.md` links use bare `docs/…` targets. All of those resolve today because they sit at the repo root, so nothing is broken — but they are the shape that broke line 41, and a file move would break them silently. Flagged only because the body calls the sweep "exhaustive — those plus `docs/` are the repo's only top-level directories": the repo also has `.agents/`, `.beads/`, and `.claude/`, which that directory list omits. No change requested in this PR.

`aligned` — the diff touches one file under `docs/guidelines/`, no page under `docs/architecture/` and no ADR status change, so no check has an anchor in this diff.

### Verdict
APPROVE — a one-line, correctly-scoped link fix whose stated verification holds up under independent re-checking; the remaining #928 groups are legitimately blocked or decisions, exactly as the body says.

---

PR #3166, opened 2026-08-04
fix(agents): align the VM guest's Claude Code config with the container image

JanPokorny:
Follow-up to #3156 — three ways the VM guest's Claude Code diverged from the container-based one, found smoke-testing on the CNV cluster.

### 1. Init script never ran on the VM backend (controller)
The container backend runs `spec.Init` as an init container; the VM rendered nothing. The default init seeds `$HOME` from `/app/working-dir` — Claude Code's onboarding state (`.claude.json`), `.claude/settings.json`, and the `work` dir. Without it: first-run wizard in terminal mode, per-command permission prompts in chat, and a missing `WORK_DIR` failing the first harness spawn. Now rendered as an inline `bootcmd` argv (bootcmd can precede `write_files`, so no file dependency), `HOME` set explicitly, idempotent per boot like the init container.

### 2. Dockerfile ENV doesn't carry into the bootc guest (claude-code-vm)
- **`IS_SANDBOX=1`** — the runtime (and harness) run as root since #3156, and Claude Code refuses `--dangerously-skip-permissions` under root without the sandbox attestation. The VM boundary is the sandbox, so the attestation is truthful.
- **`DISABLE_TELEMETRY` / `DISABLE_ERROR_REPORTING` / `DISABLE_FEEDBACK_COMMAND` / `DISABLE_FEEDBACK_SURVEY` / `DISABLE_AUTOUPDATER`**, **`PYTHONPATH`** — mirrored from the container Dockerfiles into the unit's `Environment=`.

### 3. Managed settings never copied into the guest image (claude-code-vm)
`/etc/claude-code/managed-settings.json` carries the Stop/SubagentStop background-work hooks; without it a VM session's running jobs can be reaped out from under it.

Validated interim on the CNV test cluster (`IS_SANDBOX` via `templateDefaults.env`; seed verified missing on a fresh VM and present on pods).

### DIFF
--- a/packages/controller/pkg/reconciler/vm_resources.go
+++ b/packages/controller/pkg/reconciler/vm_resources.go
@@ -179,6 +179,25 @@ func BuildVMCloudInitSecret(name string, agentSpec *types.AgentSpec, cfg *config
 		vmAgentUID, shellQuote(agentHome),
 	)})
 
+	// The init script — on the container backend an init container (see
+	// resources.go). It is what seeds $HOME from /app/working-dir on first
+	// boot (Claude Code onboarding state, settings, the work dir), so
+	// skipping it leaves the harness with a first-run wizard and no
+	// permission config. Same every-start semantics as the init container:
+	// bootcmd runs each boot and the script is idempotent by contract.
+	// Runs as root (matching agent-runtime here); HOME is set explicitly
+	// since cloud-init's env carries none.
+	initScript := agentSpec.Init
+	if initScript == "" {
+		initScript = defaults.Init
+	}
+	// Inline argv, not a write_files script: bootcmd can run before
+	// write_files within cloud-init's module order, so a file dependency
+	// would race. `env` sets HOME without shell-quoting the script.
+	if initScript != "" {
+		cc.BootCmd = append(cc.BootCmd, []string{"env", "HOME=" + agentHome, "bash", "-c", initScript})
+	}
+
 	body, err := yaml.Marshal(cc)
 	if err != nil {
 		return nil, fmt.Errorf("encoding cloud-init userdata: %w", err)
@@ -202,8 +221,9 @@ func BuildVMCloudInitSecret(name string, agentSpec *types.AgentSpec, cfg *config
 // unchanged. runStrategy is owned by applyVirtualMachine (the replicas
 // analogue); the value rendered here is a create-time default only.
 //
-// Non-persisted mounts, the user init script, and warm-pool claims are
-// container-backend concepts and deliberately don't render here.
+// Warm-pool claims are a container-backend concept and deliberately don't
+// render here. Non-persisted mount dirs and the init script ride the
+// cloud-init userdata instead (BuildVMCloudInitSecret).
 func BuildAgentVirtualMachine(name string, agentSpec *types.AgentSpec, cfg *config.Config, ownerRef metav1.OwnerReference, gatewayClusterIP string) (*unstructured.Unstructured, error) {
 	base := cfg.AgentBase
 	defaults := cfg.AgentTemplateDefaults
--- a/packages/controller/pkg/reconciler/vm_test.go
+++ b/packages/controller/pkg/reconciler/vm_test.go
@@ -25,6 +25,7 @@ func vmAgentCR() *apiv1.Agent {
 		{Path: "/home/agent", Persist: true, Size: "10Gi"},
 		{Path: "/scratchpad", Persist: false},
 	}
+	a.Spec.Init = "touch $HOME/.initialized"
 	return a
 }
 
@@ -167,6 +168,11 @@ func TestVMBackendReconcilesVirtualMachine(t *testing.T) {
 		"mount -t virtiofs 'scratchpad'",
 		"ephemeral mounts have no virtiofs device to mount",
 	)
+	// The init script must ride the userdata — it is what seeds $HOME
+	// (Claude Code onboarding state, settings, the work dir); the container
+	// backend runs it as an init container, the VM has only cloud-init.
+	assert.Contains(t, userdata, "HOME=/home/agent")
+	assert.Contains(t, userdata, ".initialized")
 	// The ~/.cache swap must be pre-created by root: on unprivileged virtiofs
 	// a non-root guest cannot create symlinks (virtiofsd lacks CAP_CHOWN), so
 	// the entrypoint's own `ln` would fail with EPERM.
--- a/packages/agents/claude-code-vm/Containerfile
+++ b/packages/agents/claude-code-vm/Containerfile
@@ -57,6 +57,9 @@ COPY --from=agent /usr/local/share/mise/ /usr/local/share/mise/
 COPY --from=agent /usr/local/share/lazy-tools/ /usr/local/share/lazy-tools/
 COPY --from=agent /etc/mise/ /etc/mise/
 COPY --from=agent /etc/AGENTS.md /etc/AGENTS.md
+# Managed settings carry the Stop/SubagentStop background-work hooks — without
+# them a VM session's running jobs are reaped out from under it.
+COPY --from=agent /etc/claude-code/ /etc/claude-code/
 COPY --from=agent /etc/profile.d/model-gateway.sh /etc/profile.d/model-gateway.sh
 COPY --from=agent /app/ /app/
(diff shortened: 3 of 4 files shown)


packages/controller/pkg/reconciler/vm_resources.go
```diff
@@ -179,6 +179,25 @@ func BuildVMCloudInitSecret(name string, agentSpec *types.AgentSpec, cfg *config
...
+	// since cloud-init's env carries none.
+	initScript := agentSpec.Init
+	if initScript == "" {
+		initScript = defaults.Init
+	}
+	// Inline argv, not a write_files script: bootcmd can run before
+	// write_files within cloud-init's module order, so a file dependency
+	// would race. `env` sets HOME without shell-quoting the script.
+	if initScript != "" {
+		cc.BootCmd = append(cc.BootCmd, []string{"env", "HOME=" + agentHome, "bash", "-c", initScript})
```
dam-code-guardian, line 198: 🟡 **Warning:** the init script runs with systemd's fixed service PATH, so any template init beyond coreutils fails here while working on the container backend.

The init container inherits the agent image's `ENV PATH=/app/node_modules/.bin:/usr/local/share/mise/shims:$PATH:/usr/local/share/lazy-tools` (platform-base Dockerfile:105). A cloud-init `bootcmd` inherits `os.environ` of the cloud-init process (`subp.subp(..., update_env=env)` copies `os.environ` then overlays only `INSTANCE_ID`), and cloud-init runs as a systemd service, whose `$PATH` systemd documents as **a fixed `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin`**. Neither the mise shims dir nor `lazy-tools` is on it.

The chart default init only uses `cp`/`touch`/`mkdir`/`test`, so it works — but the same values.yaml comment says *"Templates that need extras (e.g. pip install) REPLACE this entirely"*, and `pip`/`uv`/`python`/`node` all resolve **only** through the shims. Measured against that exact PATH on a platform image:

```
node      NOT FOUND      bash   /usr/sbin/bash
python    NOT FOUND      cp     /usr/sbin/cp
uv        NOT FOUND      touch  /usr/sbin/touch
jq        NOT FOUND      mkdir  /usr/sbin/mkdir
```

So the parity this PR establishes holds for the default init and breaks for exactly the templates the comment anticipates. Since you already pass `env HOME=…`, adding PATH is a one-token fix in the same argv — and it keeps this backend-neutral rather than depending on which init a template ships:

```suggestion
		cc.BootCmd = append(cc.BootCmd, []string{"env", "HOME=" + agentHome, "PATH=/app/node_modules/.bin:/usr/local/share/mise/shims:/usr/local/bin:/usr/bin:/bin:/usr/local/share/lazy-tools", "bash", "-c", initScript})
```

(That value mirrors `agent-runtime.service`'s own `Environment=PATH=`, so the init and the runtime resolve tools identically.)

dam-code-guardian, line 198: 🟡 **Warning:** a failing init is fail-closed on the container backend and silent here, so the failure this PR exists to fix would recur undetected.

An init **container** that exits non-zero blocks the pod: the agent container never starts and the failure is visible as pod status. On the VM this entry is appended to the `bootcmd` list, which cloud-init shellifies into a single `/bin/sh` script with **no `set -e`** (`util.shellify` emits only `#!/bin/sh` + one line per entry). Ran in that shape:

```
$ sh bootcmd-sim.sh          # init entry last, $HOME unwritable
touch: cannot touch '/nonexistent/.initialized': No such file or directory
mkdir: cannot create directory '/nonexistent': Permission denied
script exit=1
```

The non-zero status propagates to `cc_bootcmd`, which re-raises, and `Modules._run_modules` catches it into `failures` and **continues with the next module** — so `write_files` still runs, `/etc/platform/env` lands, `platform-boot-gate`'s `ConditionPathExists` passes, and `agent-runtime` starts against an unseeded `$HOME`. That is precisely the state described in the PR body (first-run wizard, permission prompts), now reached silently instead of loudly — and per-boot, since the seed guard is `[ ! -f $HOME/.initialized ]`.

Not necessarily a blocker: the neighbouring mount bootcmds deliberately end `|| true` because a broken share shouldn't wedge boot, so fail-open is a defensible house style here. But those failures are self-announcing (an empty `$HOME` makes agent-runtime "fail loud", per the comment at line 139) whereas this one boots a *working-looking* sandbox that is misconfigured. Worth at least making it visible — e.g. appending `|| echo 'platform: init script failed' >/dev/kmsg`, which reaches the captured serial console the Containerfile already relies on as "the only channel out of the guest".

packages/controller/pkg/reconciler/vm_test.go
```diff
@@ -167,6 +168,11 @@ func TestVMBackendReconcilesVirtualMachine(t *testing.T) {
 		"mount -t virtiofs 'scratchpad'",
 		"ephemeral mounts have no virtiofs device to mount",
 	)
+	// The init script must ride the userdata — it is what seeds $HOME
+	// (Claude Code onboarding state, settings, the work dir); the container
+	// backend runs it as an init container, the VM has only cloud-init.
+	assert.Contains(t, userdata, "HOME=/home/agent")
```
dam-code-guardian, line 174: 🟢 **Suggestion:** the assertions check that the init rendered, but not the property the code comment calls load-bearing.

`assert.Contains(t, userdata, …)` on `"HOME=/home/agent"` and `".initialized"` passes wherever in `bootcmd` the entry sits — including before the workspace `mount -t virtiofs` entry, where the seed would write into the pre-mount rootfs and be masked by the share. The comment at line 183 states the ordering requirement ("Runs after the mount bootcmds above so the workspace share is in place" for `~/.cache`, and the init depends on it identically), so it's worth pinning with an index comparison rather than substring presence:

```go
assert.Less(t, strings.Index(userdata, "mount -t virtiofs"), strings.Index(userdata, "HOME=/home/agent"),
	"the init script must run after the workspace mount, or it seeds the pre-mount rootfs")
```

`strings` is already imported in this file. Cheap, and it's the assertion that would actually catch a future reordering of the `bootcmd` appends.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `0a8b26f4`

## PR #3166: fix(agents): align the VM guest's Claude Code config with the container image
**Author:** JanPokorny | **Branch:** `fix/vm-claude-config-parity` → `main` | **Changes:** +40 −2 (4 files)

### Summary
Closes three VM/container divergences found smoke-testing #3156: renders `spec.Init` as a cloud-init `bootcmd` (the VM rendered nothing), mirrors the container images' `ENV` into the systemd unit, and copies `/etc/claude-code/` into the guest image. All three mechanism claims verified — cloud-init's `cloud_init_modules` really does list `bootcmd` before `write_files`, and the shipped CLI gates on `process.env.IS_SANDBOX!=="1"` exactly as the comment says.

### Findings
- 🟡 **Warning:** init bootcmd gets systemd's fixed 4-entry PATH, so a template init using `node`/`python`/`uv` fails on the VM but works on the container backend (`packages/controller/pkg/reconciler/vm_resources.go:198`)
- 🟡 **Warning:** init failure is silent on the VM but fail-closed on the container backend (`packages/controller/pkg/reconciler/vm_resources.go:198`)
- 🟢 **Suggestion:** test asserts the rendered strings, not the ordering the comment says is load-bearing (`packages/controller/pkg/reconciler/vm_test.go:174`)
- ✅ **Looks good:** `IS_SANDBOX=1` under `User=root` is the truthful attestation, not a bypass. Verified against the shipped binary: the gate is `…()===0&&process.env.IS_SANDBOX!=="1"&&!…CLAUDE_CODE_BUBBLEWRAP`, so `"1"` is the exact required value and the VM boundary genuinely is the sandbox.

**Verdict:** `aligned`

No drift. The diff changes how the VM backend delivers already-documented behavior (env parity, managed settings, running the existing `spec.Init`) without altering any subsystem's documented behavior or responsibility. `docs/architecture/persistence.md`'s single VM-backend sentence covers the `~/.cache` pre-create, which this PR does not touch; no architecture page documents init-container/cloud-init boot mechanics. No ADR status changed in the diff.

### Verdict
COMMENT — the three fixes are correct and well-evidenced; both 🟡s are parity gaps the PR's own goal implies rather than defects in what it changed.

---

PR #3167, opened 2026-08-05
fix(ui): keep an interrupted response visible (#933)

### LINKED ISSUE #933: Streaming interruption error hides the agent's response instead of appearing below it
## Problem

When an agent's response is streaming and the connection drops mid-stream (e.g. *"API Error: The socket connection was closed unexpectedly"*), the error message and its **Retry** button replace the entire message. Everything the agent had already produced disappears from view.

Nothing is actually lost — the partial response is preserved and reappears on reload. But until then, the live view is misleading: a long turn looks like an hour of work vanished, it nudges users into unnecessary retries that re-run work that was fine, and a transient network blip reads as catastrophic data loss.

## Proposed solution

On interruption, keep the already-streamed content visible and show the error notice (with **Retry**) **below** it. This matches what the user sees after a reload, so an interruption looks like an interruption rather than a lost turn.

### Out of scope

Genuine send failures where the message never reached the agent — replacing the (empty) message is fine there, since there's no partial content to preserve.

Image

jjeliga:
Closes #933.

## The problem

When a turn broke mid-stream, the error card and its **Retry** button replaced the entire assistant message. Everything the agent had already produced disappeared from view.

Nothing was ever lost — the runtime appends every chunk to the session log as it arrives, so a reload brought the partial response back. But the live view said otherwise: a long turn looked like an hour of work had vanished, which pushed people into retries that re-ran work that was fine.

## The fix

Two places conspired to hide it:

- `use-acp-prompt.ts` — the `sendPrompt` catch set `parts: []` on the assistant bubble, discarding everything already streamed into it.
- `chat-view.tsx` — the render was a ternary, so even with parts preserved the error card would still have replaced them.

Now the content block renders whenever there are parts, and the notice renders below it — matching what a reload shows, so an interruption reads as an interruption.

The card also switches its label to **"Response interrupted:"** once content is present. "Send failed" is factually wrong sitting under a half-written response, and misleading copy is the substance of the issue.

Hidden turns (the agent auto-greeting) previously dropped their bubble on any failure. They now drop it only when the agent produced nothing — discarding content the server has already persisted would recreate the exact live-view/reload mismatch this is about.

### Out of scope, per the issue

A send that never reached the agent still replaces its empty bubble. There is no partial content to preserve, and that path is unchanged.

## Verification

No automated coverage — this lives in a React view and a hook, and the suite here covers pure helpers only. Verified by hand against a local cluster with a temporary fault injector that rejected the in-flight `session/prompt` with the issue's error while leaving the WebSocket open, which is what a harness losing its own upstream connection looks like. Confirmed in both directions: with the fix reverted the bubble went blank, with it in place the content survived and the notice appeared below. The injector was removed before this branch was squashed and appears nowhere in history.

Also exercised: retry after an interruption, the empty-send path, and reload parity.

## Known limitation

This covers a harness dying with our WebSocket still open — the case in the issue. If *our* ACP socket is what drops, the reconnect fires at ~1s and `loadHistory` replaces the transcript, taking the new notice and its Retry with it. The content is still correct; only the explanation is short-lived. Covering that means merging client-side error state across a history reload, which is a design change rather than a tweak.

## Note on `hasAgentContent`

Reviewers caught that `parts.length > 0` was the wrong proxy for "the agent streamed something": `appendVerdict` anchors a **client-minted** permission verdict on the last assistant bubble when no message owns the tool call, so a bubble the agent never wrote to could look like it had content. That mislabelled an empty failure as an interruption and, worse, kept a hidden turn's orphan bubble alive. The shared predicate in `session-projection.ts` excludes verdicts rather than allow-listing agent kinds, so a future agent-produced part kind counts as content by default instead of silently reintroducing this.

### DIFF
--- a/docs/architecture/knowledge-bases.md
+++ b/docs/architecture/knowledge-bases.md
@@ -1,6 +1,6 @@
 # Knowledge Bases
 
-Last verified: 2026-07-29
+Last verified: 2026-08-05
 
 ## Overview
 
@@ -25,7 +25,7 @@ The module has no persistence of its own: a Knowledge Base is exactly the owner'
 
 Knowledge Bases is a feature-gated destination ([features](features.md)) with a list and a **standalone per-KB page** — the chat surface under the knowledge base's own route, so the rail keeps the Knowledge Bases context and leaving returns to the KB list, never to Sandboxes. Creation is **not** its own form: it is the `knowledge-base` starting point on the shared [sandbox wizard](agent-lifecycle.md)'s first step, which pins the Claude Code harness, hides it, and reveals the KB Template picker; the wizard's finish dispatches to this module's create. The KB list's own create button enters that wizard with the starting point pre-picked.
 
-Opening a KB that has no sessions yet **greets the user**: the UI runs `/wiki-onboard` as a hidden first turn (reaches the agent, renders no user bubble, fails silently), so a fresh KB opens with the agent introducing itself rather than an empty chat. This is why every template's bootstrap installs that command. The greeting mechanism is shared with experiments.
+Opening a KB that has no sessions yet **greets the user**: the UI runs `/wiki-onboard` as a hidden first turn (reaches the agent, renders no user bubble, surfaces no error — a greeting that breaks before it says anything leaves no trace, while one interrupted mid-stream keeps what it already said), so a fresh KB opens with the agent introducing itself rather than an empty chat. This is why every template's bootstrap installs that command. The greeting mechanism is shared with experiments.
 
 One known gap: the greeting races the bootstrap. It fires once the agent is running, but the Install Command lands after Ready, so a very fast open can run `/wiki-onboard` before it exists. Experiments close this by waiting for their authoring skill to be reported present; a KB has nothing equivalent to probe, because its onboarding artifact is a *command* and that read covers only skills.
 
--- a/packages/ui/src/modules/acp/session-projection.ts
+++ b/packages/ui/src/modules/acp/session-projection.ts
@@ -165,6 +165,13 @@ export function hasStreamingAssistant(messages: Message[]): boolean {
   return messages.some((m) => m.role === "assistant" && m.streaming);
 }
 
+/** True if the agent actually produced something. Verdict parts are minted by
+ *  the client when the user answers a permission prompt — they can land on a
+ *  bubble the agent never wrote to, so they don't count as agent output. */
+export function hasAgentContent(m: Message): boolean {
+  return m.parts.some((p) => p.kind !== "verdict");
+}
+
 function handleUserChunk(messages: Message[], u: ContentChunk): Message[] {
   const queued = u._meta?.queued === true;
   const mid = u.messageId ?? null;
--- a/packages/ui/src/modules/sessions/components/chat-message-part.tsx
+++ b/packages/ui/src/modules/sessions/components/chat-message-part.tsx
@@ -1,0 +1,69 @@
+import { Document } from "@carbon/icons-react";
+
+import { formatBytes } from "@/lib/format-size";
+
+import { Markdown } from "../../../components/markdown.js";
+import type { MessagePart, Role } from "../../../types.js";
+import { PermissionVerdictLine } from "./permission-prompt.js";
+import { ThoughtBlock } from "./thought-block.js";
+import { ToolChip } from "./tool-chip.js";
+
+interface Props {
+  part: MessagePart;
+  role: Role;
+  /** The owning message is still streaming — drives the thought block's
+   *  in-progress state and the typing caret. */
+  streaming: boolean;
+  /** Last part of the message: only that one carries the caret. */
+  isLast: boolean;
+  onFileClick: (path: string) => void;
+}
+
+/** One part of a message, rendered by kind. */
+export function ChatMessagePart({
+  part,
+  role,
+  streaming,
+  isLast,
+  onFileClick,
+}: Props) {
+  switch (part.kind) {
+    case "text":
+      if (role === "assistant")
+        return <Markdown onFileClick={onFileClick}>{part.text}</Markdown>;
+      return (
+        <span className="whitespace-pre-wrap break-words">
+          {part.text}
+          {streaming && isLast && (
+            <span className="inline-block w-[7px] h-4 bg-accent ml-0.5 align-text-bottom anim-blink rounded-sm" />
+          )}
+        </span>
+      );
+    case "thought":
+      return <ThoughtBlock text={part.text} streaming={streaming} />;
+    case "image":
+      return (
+        <img
+          src={`data:${part.mimeType};base64,${part.data}`}
+          alt="image"
+          className="max-w-[400px] max-h-[400px] rounded-lg border border-border object-contain"
+        />
+      );
+    case "verdict":
+      return <PermissionVerdictLine verdict={part} />;
+    case "file":
+      return (
+        <div className="inline-flex items-center gap-2 rounded-md border border-border bg-muted px-3 py-2">
+          <Document size={14} className="text-muted-foreground shrink-0" />
+          <span className="text-xs text-muted-foreground">{part.name}</span>
+          {part.size !== undefined && (
+            <span className="text-[10px] text-muted-foreground">
+              {formatBytes(part.size)}
+            </span>
+          )}
+        </div>
+      );
+    default:
+      return <ToolChip chip={part} />;
+  }
+}
--- a/packages/ui/src/modules/sessions/components/chat-message.tsx
+++ b/packages/ui/src/modules/sessions/components/chat-message.tsx
@@ -1,0 +1,102 @@
+import { cn } from "@/lib/utils";
+
+import type { Attachment, Message } from "../../../types.js";
+import { hasAgentContent } from "../../acp/session-projection.js";
+import { BusyIndicator } from "./busy-indicator.js";
+import { ChatMessagePart } from "./chat-message-part.js";
+import { PermissionStatusLine } from "./permission-prompt.js";
+import { SendErrorCard } from "./send-error-card.js";
+
+interface Props {
+  message: Message;
+  /** Last entry in the transcript — only that one anchors the pending-approval
+   *  status line. */
+  isLast: boolean;
+  hasPendingPermission: boolean;
+  onRetry: (text: string, attachments?: Attachment[]) => void;
+  onFileClick: (path: string) => void;
+}
+
+/** One transcript entry. A failed turn keeps whatever it already streamed and
+ *  takes the error card *underneath* that content; the card stands in for the
+ *  bubble only when there is nothing to show, so an interruption never reads
+ *  as a lost turn. The card's own label keys off `hasAgentContent`, which is
+ *  stricter than "has parts" — a verdict-only bubble renders its verdict line
+ *  and still reads as a send that never landed, because the agent never
+ *  produced anything. */
+export function ChatMessage({
+  message,
+  isLast,
+  hasPendingPermission,
+  onRetry,
+  onFileClick,
+}: Props) {
+  if (message.notice) {
+    return (
+      <div className="flex justify-center anim-in">
+        <span className="text-[11px] italic text-muted-foreground px-3 py-1 border-t border-b border-border/60">
+          {message.parts.find((p) => p.kind === "text")?.text ?? "…"}
+        </span>
+      </div>
+    );
+  }
+
+  const { role, parts, streaming, queued, error } = message;
+  const isAssistant = role === "assistant";
+  const retryWith = error?.retryWith;
+
+  return (
+    <div
+      data-testid="chat-message"
+      data-role={role}
+      className={cn(
+        "flex flex-col gap-1 anim-in",
+        isAssistant ? "items-start" : "items-end",
+      )}
+    >
+      <span className="text-[11px] font-medium text-muted-foreground mb-0.5">
+        {isAssistant ? "Agent" : "You"}
+      </span>
+      {(!error || parts.length > 0) && (
+        <div
+          className={
+            isAssistant
+              ? "flex flex-col gap-4 w-full max-w-full"
+              : "flex flex-col gap-2 rounded-xl border border-border bg-card px-4 py-3 text-sm text-foreground"
+          }
+        >
+          {parts.map((p, i) => (
+            <ChatMessagePart
+              key={i}
+              part={p}
+              role={role}
+              streaming={streaming}
+              isLast={i === parts.length - 1}
+              onFileClick={onFileClick}
+            />
+          ))}
+          {streaming && queued && parts.length === 0 && (
+            <span className="text-xs text-muted-foreground italic">
+              Waiting for previous prompt…
+            </span>
+          )}
+          {isAssistant && isLast && <PermissionStatusLine />}
+          {isAssistant && streaming && !queued && !hasPendingPermission && (
+            <BusyIndicator className="py-1" />
+          )}
+        </div>
+      )}
+      {error && (
+        <SendErrorCard
+          rawError={error.message}
+          interrupted={hasAgentContent(message)}
+          onRetry={
+            retryWith
+              ? () => onRetry(retryWith.text, retryWith.attachments)
+              : undefined
+          }
+        />
+      )}
+    </div>
+  );
+}
--- a/packages/ui/src/modules/sessions/components/send-error-card.tsx
+++ b/packages/ui/src/modules/sessions/components/send-error-card.tsx
@@ -1,0 +1,54 @@
+import { Renew, Warning } from "@carbon/icons-react";
+
+import { Button } from "@/components/ui/button";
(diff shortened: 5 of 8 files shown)


dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `c711a6e`

## PR #3167: fix(ui): keep an interrupted response visible (#933)
**Author:** jjeliga | **Branch:** fix/933-keep-interrupted-stream-visible-in-ui → main | **Changes:** +58 −33 (4 files)

### Summary
Stops the error card from replacing an assistant bubble's streamed content when a turn breaks mid-stream: the catch path no longer clears `parts`, the render puts the card below the content, and a new shared `hasAgentContent` predicate — excluding client-minted verdict parts — drives both the hidden-turn drop and the card's "Response interrupted:" label.

### Findings
- 🟡 **Warning:** the 60 s delivery watchdog still keys on `parts.length === 0`, so a client-minted verdict part disarms it — the exact proxy `hasAgentContent` was added to replace (`packages/ui/src/modules/sessions/hooks/use-acp-prompt.ts:126`).

  The PR body explains why `parts.length > 0` was the wrong proxy for "the agent streamed something": `appendVerdict` anchors a client-minted verdict on the last assistant bubble when no message owns the tool call, so a bubble the agent never wrote to can look like it has content. That reasoning applies verbatim here, but this line was left on the old predicate.

  Concretely: the harness may request permission before emitting the `tool_call` update (the fallback `appendVerdict` documents at `chat-input-area.tsx:23-26`), so the verdict anchors on the newest assistant bubble. If that is a fresh empty bubble — a Retry click, or a queued prompt — then `parts` is `[{kind:"verdict"}]`, `parts.length === 0` is false, and the watchdog silently declines to fire. The turn that never gets a response then shows no "Couldn't deliver" card and no error at all; it just sits closed and empty. That is the failure mode the 60 s net exists to catch.

  I ported this path plus `appendVerdict`'s anchor logic and ran it: with the verdict part present the watchdog does not fire; with no verdict part on an otherwise identical bubble it does. Same predicate, opposite outcomes.

  The fix is the predicate this PR already added — `if (bubble?.streaming && !hasAgentContent(bubble))`. Note the two are **not** interchangeable everywhere: `chat-view.tsx:695` and `:763` should keep `parts.length` (a verdict-only bubble *does* have something to render, and its "Waiting for previous prompt…" hint should stay hidden). This site is specifically asking "did the agent respond?", which is what `hasAgentContent` means.
- ✅ **Looks good:** the content block gating on `parts.length > 0` while the card's label gates on `hasAgentContent(m)` reads like an oversight but is correct — a verdict-only bubble should render its verdict line and still say "Send failed:", because the agent produced nothing.

**Verdict:** minor drift

**Drift**

*Check 1 — architecture-page drift*

❌ `docs/architecture/knowledge-bases.md:28` — the hidden-turn contract stated on this page no longer matches the code in the diff.

The page says the greeting turn "reaches the agent, renders no user bubble, **fails silently**". `use-acp-prompt.ts` changes exactly that responsibility: the unconditional `if (hidden) dropBubble()` became `if (hidden && !streamed)`, so a hidden turn that already streamed content now **keeps its bubble** and only suppresses the error card. The PR's own docstring records the new contract — "drop the turn silently *if it fails before producing anything*" (`use-acp-prompt.ts:19-22`) — and its body describes the behavior change explicitly ("Hidden turns … now drop it only when the agent produced nothing").

This is user-visible on the documented surface, not an internal detail: a KB greeting that breaks mid-stream previously left a fresh KB looking like an empty chat, and now leaves the partial greeting on screen. "Fails silently" is still true of the *error notice*; it is no longer true of the *turn*, which is what the sentence claims.

Proposed edit — keep it at the level of the contract, not the code:

```
Opening a KB that has no sessions yet **greets the user**: the UI runs
`/wiki-onboard` as a hidden first turn (reaches the agent, renders no user
bubble, and surfaces no error — a greeting that fails before producing
anything leaves no trace, while one interrupted mid-stream keeps what it
already said), so a fresh KB opens with the agent introducing itself rather
than an empty chat.
```

*Check 2 — `Last verified:` staleness*

❌ Same page, header at `knowledge-bases.md:3` (`Last verified: 2026-07-29`). Editing the page for the drift above requires bumping it to the PR date.

**Possible drift**

⚠️ `docs/architecture/experiments.md:61-65` — the experiments greeting is the *same* mechanism (`useAgentGreeting`, one call site at `use-agent-greeting.ts:64`, and `knowledge-bases.md:28` says so: "The greeting mechanism is shared with experiments"). This page describes the greeting only as "the UI hidden-sends `/experiment-onboard`" and never states the failure behavior, so nothing on it is now false.

Question for a human: does experiments own any depth on hidden-send failure semantics, or does knowledge-bases own that concept with experiments keeping its one-liner? Under the guidelines' cross-page concept ownership rule the latter needs no edit — but the two pages should not end up each half-describing it.

Scope: the three routed `.ts` files, changed lines only, baseline = unchanged code.

**Findings:** none.

The architecture docs describe a **server-side** three-layer module architecture (`services → domain ← infrastructure`, contract-package tRPC routers with Zod validation, `compose.ts` composition roots, `index.ts` module boundaries, RxJS domain events). Per `slice-composition.md`, those layers exist "within each module in **server packages** — where actual business logic lives."

This diff is entirely inside `packages/ui`, a React + Vite SPA. Two structural facts, both verified rather than assumed:

- `packages/ui/src/modules/*` has **no `index.ts` anywhere** and no `services/`, `domain/`, `infrastructure/`, or `compose.ts` — the UI package does not adopt the layered/bounded-context convention at all. So the "any import that bypasses `index.ts`" rule has no `index.ts` to bypass here, and the new `chat-view.tsx` → `session-projection.js` import is the same shape as the `finalizeAllStreaming` / `hasStreamingAssistant` imports already in `use-acp-prompt.ts` on the baseline. Not a violation introduced by this change.
- The change is UI view-state only: one pure predicate added to the projection helper module, its two call sites, and a doc comment on an existing optional field. No router, service, repository, adapter, or domain event is touched, so no dependency-direction, validation-placement, or `Result`-type rule is in play.

On strict TypeScript the changed lines are clean: `hasAgentContent(m: Message): boolean` is fully annotated with no `any` and no assertion; the narrowing `p.kind !== "verdict"` operates on the `MessagePart` discriminated union in `types.ts`; the new `interrupted?: boolean` prop is explicitly typed. `dev-config/tsconfig.base.json` sets `strict: true`, and the diff introduces no non-null assertion beyond the `m.error!` pattern already on the baseline.

One candidate run down and discarded rather than reported: the catch path now assigns `error` the value `undefined` on a hidden turn instead of omitting the key. Every `.error` consumer in the package (`chat-view.tsx:695`, `:780`, `use-acp-prompt.ts:114`) is a truthiness or optional-chain read, and executing both object shapes against them gives identical results; only `"error" in m` could distinguish them, and nothing uses that form. `exactOptionalPropertyTypes` is absent from both tsconfigs, so `strict: true` does not reject the assignment either.

**Smallest set of changes:** none required by these architecture docs. Holding the UI package to the TSEng module conventions is a package-wide decision, not something this 4-file bug fix should carry.

Scope: changed lines in `chat-view.tsx`. Measurements are from the file at HEAD `c711a6e`, not estimates.

🟡 **[CRITICAL — list-item rule] The `messages.map(...)` block is 129 lines of inline per-item JSX, and this change added a second independent conditional branch to it** (`chat-view.tsx:671-799`).

The rule defines "non-trivial" inside a `.map(...)` as "more than ~10 lines of JSX, **any conditional rendering branch**, any local handler with a multi-statement body, or any local derivation" — and requires the item become its own component. The block is 13× that threshold and now carries three conditional branches (`m.notice ?`, `(!m.error || m.parts.length > 0) &&`, `m.error &&`) plus a per-item derivation (`hasAgentContent(m)`) and a multi-statement inline retry handler.

This is a pre-existing violation, so the baseline rule turns on whether this change *amplified* it. It did, structurally: the diff replaced one `m.error ? A : B` ternary — a single branch point with mutually exclusive arms — with **two independently-evaluated conditionals** whose correctness now depends on a non-obvious relationship between them. The content block renders on `!m.error || m.parts.length > 0`; the card renders on `m.error`; the card's *label* keys off a third predicate, `hasAgentContent(m)`. Reading whether the three agree requires holding 129 lines of nesting in view, which is the cost the rule exists to prevent.

"Touch-it = migrate-it" applies directly. The extraction has an obvious name and a clean boundary — no state, no effects, just props already in scope:

```tsx
{messages.map((m, mi) => (
  <ChatMessage
    key={m.id}
    message={m}
    isLast={mi === messages.length - 1}
    onRetry={sendPrompt}
    onFileClick={openFileHandler}
  />
))}
```

`ChatMessage` then owns the notice branch, the content block, the parts `.map`, and the error card — and the three predicates become readable side by side in a ~60-line leaf. It belongs in `modules/sessions/components/chat-message.tsx` alongside `SendErrorCard`, which the same rule already flags (this file defines four components in 1030 lines, with `ChatView` itself at 825). Worth framing as a follow-up rather than a merge blocker — but it should not wait for a third edit to this block.

**Not flagged** (checked, rules not violated):

- **`cn()` usage** (`chat-view.tsx:1002-1005`) — correct. The change converts a static `className` into `cn("… anim-in", interrupted && "mt-2")`, precisely the prescribed conditional-class form; `cn` is already imported at line 35. No template-literal class string, no static `style={{}}`, and `anim-in` is a real utility defined at `App.css:329`.
- **`interrupted?: boolean`** — the rule warns against optional props meaning "configure how I work," but this is a genuine two-state presentational variant on a leaf card with no composition, and the styling rules explicitly say not to reach for `cva` at two variants. The inline-object prop type matches both sibling components in this file (`ChatHeaderStatus`, `SessionErrorCard`), so the `interface Props` preference is a package-wide inconsistency, not one this diff introduced.
- **State lineage** — no new state. `hasAgentContent(m)` is computed on render from data already in the Zustand store, the correct home for this UI-owned transcript state; nothing is duplicated into `useState`.

**Smallest change clearing the most findings:** extract the `messages.map` body into `modules/sessions/components/chat-message.tsx`.

_(1 finding moved to summary-only: its line is not inside a diff hunk, so GitHub cannot anchor an inline comment there.)_

### Verdict
COMMENT — the fix is correct and well-evidenced, and the two structural notes (a watchdog left on the proxy this PR replaced, the documented hidden-turn contract now stale) are worth a follow-up rather than blocking.

---

PetrBulanek (approved):

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `700847f`

## PR #3167: fix(ui): keep an interrupted response visible (#933)
**Author:** jjeliga | **Branch:** fix/933-keep-interrupted-stream-visible-in-ui → main | **Changes:** +269 −194 (8 files)

### Summary
Follow-up commit fixing all four findings from the previous review: the delivery watchdog moves onto `hasAgentContent`, the 129-line `messages.map` body is extracted into three components, and the knowledge-bases page records the new hidden-turn contract with its date bumped.

### Changes since last review
Previous HEAD: `c711a6e` (2026-08-05T08:02:41Z) — verdict COMMENT

- ✅ **Fixed:** the 60 s delivery watchdog now keys on `!hasAgentContent(bubble)` (`use-acp-prompt.ts:127`).
- ✅ **Fixed:** hidden-turn contract restated on the architecture page (`knowledge-bases.md:28`).
- ✅ **Fixed:** `Last verified:` bumped to 2026-08-05 (`knowledge-bases.md:3`).
- ✅ **Fixed:** the map body is now a 9-line `<ChatMessage>` call; `chat-view.tsx` drops 1016 → 854 lines.

### Findings

Verification notes on the fixes, since a fix that satisfies a finding's wording can still miss its mechanism:

- The watchdog fix is real **and its coverage is complete**. Grepping every remaining `parts.length` occurrence: `chat-message.tsx:60`/`:78` correctly **keep** `parts.length` (a verdict-only bubble does have something to render, and the "Waiting for previous prompt…" hint should stay hidden), and `session-projection.ts:404` is the queued-placeholder shape check, where `parts.length === 0` is the right question. No site was over-migrated.
- I ported the status-line anchor logic plus both predicates and ran it over seven transcript states. The change **fixes a latent second bug for free**: on `main`, an assistant bubble with an error rendered no `PermissionStatusLine` anywhere — the ternary replaced the container that holds it, and `statusLineInThread` still suppressed the trailing fallback — so a pending approval on an interrupted turn had no anchor. After this commit the interrupted case gets its anchor back. The two genuinely-empty cases still show none, which matches `main` and is out of this PR's scope.
- The extraction's "same DOM" claim holds where it matters: `data-testid="chat-message"` and `data-role` both survive, and those are exactly the two attributes the e2e helper `readChatMessages` (`packages/e2e/playwright/src/lib/agents.ts:162`) reads.

**Verdict:** aligned

Both drift items from the previous review are closed. `knowledge-bases.md:28` now states the contract at the right level — "surfaces no error — a greeting that breaks before it says anything leaves no trace, while one interrupted mid-stream keeps what it already said" — which matches `use-acp-prompt.ts`'s `if (hidden && !streamed) dropBubble()`, and the header is bumped to `2026-08-05`, the PR date.

The prior review's *possible drift* question also resolves correctly without an edit: `experiments.md` describes the shared greeting only as "the UI hidden-sends `/experiment-onboard`" and states no failure semantics, so nothing on it is false. Under the guidelines' cross-page concept ownership rule, knowledge-bases owns the concept in depth and experiments keeps its one-liner — which is the state now shipped.

No other check fires: no architecture page names the error card, the interruption behavior, or the watchdog, so no other page went stale; the diff adds no new subsystem, moves no page, touches no ADR, and adds no volatile detail (the new sentence names no symbol or path). Page size is 6444 chars against the 40000 cap — measured with `content.length`, the unit `scripts/doc-size.mjs` actually gates on, not `wc -c`.

Scope: the three routed `.ts` files, changed lines only, baseline = unchanged code.

**Findings:** none.

The architecture docs describe a **server-side** three-layer model (`services → domain ← infrastructure`), contract-package tRPC routers with Zod validation, `compose.ts` composition roots, and `index.ts` module boundaries. Per `slice-composition.md`, those layers exist "within each module in **server packages**". This diff is entirely inside `packages/ui`, a React SPA — and both structural facts were re-verified at this HEAD with controls that must produce hits:

- `packages/ui/src/modules/*` has **zero** `index.ts` files and no `services/`/`domain/`/`infrastructure/`/`compose.ts`. The control confirms the convention is real elsewhere (`packages/cli/src/modules/{skill,terms,template}/` all carry both), so this is a genuine package-wide opt-out rather than a failed search — the "any import that bypasses `index.ts`" rule has no `index.ts` to bypass here.
- The new `sessions` → `acp/session-projection.js` imports match **eleven** baseline imports already crossing that boundary. Not introduced by this change.

On strict TypeScript the changed lines typecheck clean — verified rather than asserted. `node_modules` is absent, so I extracted the changed logic plus its types verbatim and ran `tsc --strict`: exit 0. `bubble?.streaming && !hasAgentContent(bubble)` narrows correctly through the optional chain; `hasAgentContent` is fully annotated with no `any` and no assertion. The catch path's `const error = hidden ? undefined : {…}` assigned to an optional prop is legal because `exactOptionalPropertyTypes` is off — and the control run with that flag enabled fails on the *unchanged* watchdog assignment simultaneously, proving the pattern predates this diff. The lockfile resolves `typescript@5.9.3`, so the new `find((p) => p.kind === "text")?.text` relies on inferred type predicates (TS ≥ 5.5) safely — and it *replaces* an `as {text: string}` cast.

Net: two non-null assertions and one type assertion removed.

**Smallest set of changes:** none required. Holding `packages/ui` to the TSEng module conventions is a package-wide decision, not this bug fix's job.

Scope: the three new `.tsx` files and the `chat-view.tsx` changed lines. Measurements from files at HEAD, not estimates.

**Findings:** none — the prior CRITICAL list-item finding is fully resolved.

The `messages.map(...)` body is now exactly the shape the rule prescribes: a 9-line call with `key`, no conditionals, no derivations, no inline handler. Both extracted files sit inside the weight targets (`chat-message.tsx` ~57 JSX lines against the ~60 feature bound; `chat-message-part.tsx` a flat 5-case switch).

Four things were cleared that the prior review did not ask for, each verified against `main`: the template-literal class string at `main:689` became a `cn()` call (the styling rule's ❌ example replaced by its ✅); two non-null assertions and one `as` cast dropped; and all three new files adopt the [HIGH] `interface Props` form, where the two components still inline in `chat-view.tsx` use inline object types.

**Not flagged** (checked, rules not violated): `interrupted?: boolean` is a genuine two-state presentational variant on a leaf, carried over unchanged, and the rules say not to reach for `cva` at two variants · no new state — all three files are pure presentational with no hooks, and `hasAgentContent(message)` is computed on render from Zustand-owned data · `hasPendingPermission` reads as per-item but is one global selector lifted at `chat-view.tsx:140`, which avoids 19 store subscriptions · 19 siblings in `components/` is past the ~10 nested-folder hint, but that is pre-existing and batch moves are explicitly a separate PR.

### Verdict
APPROVE — all four findings from the previous review are fixed, verified at the mechanism rather than the summary, and the extraction cleared three assertions and a template-literal class string along the way.

---

PR #3168, opened 2026-08-05
fix(controller,agents): route the JVM through the egress proxy and the right home

JanPokorny:
Found by an agent doing a real Maven + Docker build inside a VM sandbox. Three distinct causes behind "the agent is fighting the proxy":

### 1. The JVM ignores `http_proxy` env (both backends)
Java takes proxies only from `-Dhttp(s).proxyHost/Port`, so Maven/Gradle silently bypassed the egress gateway — on pods too. `agentPlatformEnv` now composes `JAVA_TOOL_OPTIONS` (proxy host/port + `-Duser.home=$HOME`).

### 2. `user.home` resolves via `getpwuid`, not `$HOME` (VM)
The runtime/harness run as root (#3156), so Java read `/root/.m2` while the platform uses `/home/agent`. Fixed twice over: `-Duser.home` above, and root's passwd home now points at `/home/agent` in the guest image — covering every getpwuid-based tool, not just the JVM.

### 3. Docker builds couldn't fetch (VM)
- **daemon MTU**: guest NIC rides the pod network (MTU 1400) while docker0 defaulted to 1500 — TCP connects, large transfers hang. A pre-docker boot oneshot derives the MTU from the default-route interface and writes `daemon.json` each boot (ephemeral rootfs). k3s/flannel self-derives; docker doesn't.
- **client proxies**: the daemon's systemd drop-in covers image pulls only; BuildKit injects proxy env into `RUN` steps solely from `~/.docker/config.json`. Now rendered at boot for `/root` and `/home/agent` from `/etc/platform/env`, absent-only so `docker login` configs are never clobbered.

Tests: `JAVA_TOOL_OPTIONS` (user.home + proxy host/port) asserted in the VM userdata rendering; net-config script is exercised at image build/boot.

### DIFF
--- a/deploy/tasks.toml
+++ b/deploy/tasks.toml
@@ -196,23 +196,29 @@ kubectl --kubeconfig="$KUBECONFIG" wait --for=condition=Ready node --all --timeo
 # 1b. Ensure k3s container-log rotation is configured (issue #244 upgrade
 # path for VMs created before this flag was added). New VMs already have
 # /etc/rancher/k3s/config.yaml from the lima provision script.
+# A config.yaml.d drop-in, never config.yaml itself: k3s merges the
+# directory natively, and overwriting the main file clobbers pre-existing
+# config — inside a VM sandbox it carries the CIDR remap that keeps the
+# inner cluster off the host cluster's ranges, and losing it crash-loops
+# k3s ("cidr ... is out the range of cluster cidr").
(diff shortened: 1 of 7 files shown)


JanPokorny:
Added 5e0ca4ab: `/var/lib/containerd` joins the scratch-disk binds — moby's containerd image store was filling the 8G rootfs `/var` during docker builds (ENOSPC) while the 30G scratch disk idled. Found by the same in-sandbox build session as the rest of this PR.

PR #3173, opened 2026-08-05
fix(ui): system tag strip must match closing tag by name

tomkis:
## Problem

User bubbles in the session transcript sometimes render as a bare `</task-notification>` and nothing else.

The strip that removes harness-internal tags from user messages (`packages/ui/src/modules/acp/session-projection.ts`) used

```js
/(<[a-z-]+>)([\s\S]*?)(<\/[a-z-]+>)/g
```

The closing tag is not tied to the opening one, so the lazy body ends at the *first* closing tag of any name. The harness injects background-work notifications as a wrapper around a nested element:

```
<task-notification>
<task id="ab12" status="completed">
...
</task>
</task-notification>
```

The match runs from `<task-notification>` to `</task>` and gets removed. The orphan `</task-notification>` is left behind, and no later pass can match it because the loop needs an opening tag. User text renders as plain text, so the tag shows up verbatim as the whole bubble.

A notification with no nested element already stripped to `""` and was dropped, which is why only some of these bubbles were visible.

## Fix

Backreference the tag name:

```js
/<([a-z][a-z0-9-]*)>[\s\S]*?<\/\1>/g
```

The wrapper now matches through its own closing tag, so the whole notification is stripped and no bubble appears. That matches how the non-nested case already behaved.

## Tests

Two regression tests in `session-projection.test.ts`: a nested wrapper leaves nothing, and a nested wrapper still closes the active assistant bubble as a turn boundary. Both fail on the old regex.

`mise run ui:test` 223/223, `mise run ui:check` clean.

## Follow-ups, not in this PR

- `stripUserTags` still removes any balanced lowercase tag pair from a user message, so typing `<div>hello</div>` in the chat loses the content. Pre-existing and wider; the sturdy fix is an allowlist of real harness tags instead of a generic pattern.
- Task notifications are now invisible in the transcript. Consistent with the non-nested case, but showing them as a centered `notice` message is a UX call worth making separately.

https://claude.ai/code/session_01C9yLCJqzG9beXvuWNfvy6M

### DIFF
--- a/packages/ui/src/modules/acp/session-projection.ts
+++ b/packages/ui/src/modules/acp/session-projection.ts
@@ -54,7 +54,7 @@ import type { AcpUpdate } from "./types.js";
  * `<task-notification>…</task>` and left the orphan `</task-notification>`
  * as the whole visible user bubble (issue: stray closing tag in transcript).
  */
-const SYSTEM_TAG_RE = /<([a-z][a-z0-9-]*)>[\s\S]*?<\/\1>/g;
+const SYSTEM_TAG_RE = /<([a-z-]+)>[\s\S]*?<\/\1>/g;
 function stripUserTags(raw: string): string {
   let result = raw;
   let prev;
--- a/packages/ui/src/modules/acp/session-projection.ts
+++ b/packages/ui/src/modules/acp/session-projection.ts
@@ -46,8 +46,15 @@ import type { AcpUpdate } from "./types.js";
  * user messages are whole, not streamed, so trimming is safe. Do NOT apply
  * to agent chunks: those arrive piece by piece and trimming would collapse
  * inter-chunk spaces into `"helloworld"`.
+ *
+ * The closing tag must match the opening one by name (backreference). Without
+ * that, the lazy body ends at the *first* closing tag of any name, so a
+ * wrapper holding a nested element — the harness's
+ * `<task-notification><task …>…</task></task-notification>` — matched
+ * `<task-notification>…</task>` and left the orphan `</task-notification>`
+ * as the whole visible user bubble (issue: stray closing tag in transcript).
  */
-const SYSTEM_TAG_RE = /(<[a-z-]+>)([\s\S]*?)(<\/[a-z-]+>)/g;
+const SYSTEM_TAG_RE = /<([a-z][a-z0-9-]*)>[\s\S]*?<\/\1>/g;
 function stripUserTags(raw: string): string {
   let result = raw;
   let prev;
--- a/packages/ui/src/__tests__/unit/session-projection.test.ts
+++ b/packages/ui/src/__tests__/unit/session-projection.test.ts
@@ -268,6 +268,38 @@ describe("applyUpdate — turn boundaries", () => {
     ]);
   });
 
+  test("a system wrapper holding a nested element leaves no orphan closing tag", () => {
+    // The harness injects background-work notifications as a user message
+    // wrapping a nested element. A name-agnostic strip ended the match at
+    // `</task>` and left `</task-notification>` as the entire bubble.
+    const out = applyUpdate([], {
+      sessionUpdate: "user_message_chunk" as const,
+      messageId: "u1",
+      content: {
+        type: "text" as const,
+        text: '<task-notification>\n<task id="ab12" status="completed">\ndiag run finished\n</task>\n</task-notification>',
+      },
+    });
+    expect(out).toEqual([]);
+  });
+
+  test("a system wrapper with nested elements still closes the active bubble", () => {
+    const start: Message[] = [
+      userMsg("u1", "go"),
+      assistantMsg("a1", "working", true),
+    ];
+    const out = applyUpdate(start, {
+      sessionUpdate: "user_message_chunk" as const,
+      messageId: "u2",
+      content: {
+        type: "text" as const,
+        text: "<system-reminder>\n<note>heads up</note>\n</system-reminder>",
+      },
+    });
+    expect(out).toHaveLength(2);
+    expect(out[1].streaming).toBe(false);
+  });
+
   test("user_message_chunk extracts binary file reference as file chip", () => {
     const out = applyUpdate([], {
       sessionUpdate: "user_message_chunk" as const,


packages/ui/src/modules/acp/session-projection.ts
```diff
@@ -46,8 +46,15 @@ import type { AcpUpdate } from "./types.js";
...
+ *
+ * The closing tag must match the opening one by name (backreference). Without
+ * that, the lazy body ends at the *first* closing tag of any name, so a
+ * wrapper holding a nested element — the harness's
+ * `<task-notification><task …>…</task></task-notification>` — matched
+ * `<task-notification>…</task>` and left the orphan `</task-notification>`
+ * as the whole visible user bubble (issue: stray closing tag in transcript).
  */
-const SYSTEM_TAG_RE = /(<[a-z-]+>)([\s\S]*?)(<\/[a-z-]+>)/g;
+const SYSTEM_TAG_RE = /<([a-z][a-z0-9-]*)>[\s\S]*?<\/\1>/g;
```
dam-code-guardian: 🟡 **Warning:** The backreference alone fixes the reported bug; widening the character class to `[a-z0-9-]` is a separate behavior change that costs more than it buys.

Verified by running both variants over the PR's own bug case:

```
/<([a-z-]+)>[\s\S]*?<\/\1>/g      (backref, old char class)
  '<task-notification>\n<task id="ab12" …>\nx\n</task>\n</task-notification>'  →  ""   ✅ fixed
  '<h1>Title</h1>'                                                            →  unchanged
/<([a-z][a-z0-9-]*)>[\s\S]*?<\/\1>/g   (this PR)
  '<h1>Title</h1>'                                                            →  ""   ← newly stripped
```

So the digits buy nothing for the orphan-tag fix, and they enlarge the follow-up the PR body already names ("typing `<div>hello</div>` in the chat loses the content") to cover `<h1>`, `<h2>`, `<h3>` — the markdown headings a user is most likely to type in a chat box. No harness tag in the repo has a digit in its name, so nothing currently needs the wider class.

Narrowing this to the backreference-only change keeps the fix and its two new tests passing while leaving the pre-existing over-capture exactly as wide as it was:

```suggestion
const SYSTEM_TAG_RE = /<([a-z-]+)>[\s\S]*?<\/\1>/g;
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `2a4906e4`

## PR #3173: fix(ui): system tag strip must match closing tag by name
**Author:** tomkis | **Branch:** fix/orphan-tag-in-user-bubble → main | **Changes:** +40 −1 (2 files)

### Summary
Ties `SYSTEM_TAG_RE`'s closing tag to its opening tag by backreference, so a harness wrapper holding a nested element is stripped whole instead of leaving an orphan `</task-notification>` as the entire user bubble. Diagnosis and fix are both correct; verified against the real predicate.

### Findings
- 🟡 **Warning:** the `[a-z0-9-]` widening is unnecessary for this bug and newly eats `<h1>…</h1>` user text (`packages/ui/src/modules/acp/session-projection.ts:57`)
- ✅ **Looks good:** the fix is strictly more conservative than the old regex on mismatched pairs — a brute-force differential over 2793 tag sequences found 84 inputs where the old regex left an orphan and the new one does not, and zero where the new regex leaves an orphan the old one stripped. The two documented behaviors are preserved exactly: a non-nested notification still strips to `""`, and same-name sibling notifications still both strip.

**Verdict:** aligned

The diff changes one regex and adds two unit tests in `packages/ui/src/modules/acp/` — a defect fix in the UI's session-projection strip, not a change to a subsystem's behavior or responsibility as the architecture pages describe it. Check 1 has no anchor: no page under `docs/architecture/` documents the user-message tag-strip or how harness notifications surface in the transcript. The nearest claim, `knowledge-bases.md:28`'s "renders no user bubble", was verified independent of this code path — `useAgentGreeting` suppresses the greeting bubble through its own send path, not through `stripUserTags` — so it is accurate at this HEAD and needs no edit. Checks 2–7 produce nothing: no architecture page is edited, no long-lived component is introduced, and no ADR changed state in the diff.

**No findings.**

- **Dependency direction / module boundaries** — the diff adds and removes no imports at all, so no cross-module reach is introduced and nothing bypasses a module boundary.
- **Purity of the projection** — the file's header states "Pure: no DOM, no refs, no side effects", and the change preserves it. `SYSTEM_TAG_RE` is a module-level regex carrying `/g`, which makes it `lastIndex`-stateful and shared across calls, but `stripUserTags` consumes it only through `String.prototype.replace`, which resets `lastIndex` on a global regex — so no state leaks between `applyUpdate` invocations. The file's other global regex (`parseUserText:95`) is driven by an `.exec` loop where that statefulness *would* matter, and is correctly function-local; the diff disturbs neither.
- **Layered structure** — not applicable: the docs scope `services`/`domain`/`infrastructure` to server packages, and `packages/ui` is a React SPA with feature-folder modules across all 20 of them.
- **Test placement** — the regression tests join the existing turn-boundary suite in the sibling `__tests__/unit/` file.

### Verdict
COMMENT — the diagnosis is exact and the backreference is the right fix; one unnecessary widening rides along with it, worth dropping before merge but not blocking.

---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `cac75845`

## PR #3173: fix(ui): system tag strip must match closing tag by name
**Author:** tomkis | **Branch:** fix/orphan-tag-in-user-bubble → main | **Changes:** +40 −1 (2 files)

### Summary
Applies the prior review's suggestion — the regex keeps its name-backreference (fixing the orphan `</task-notification>` bubble) and drops the `[a-z0-9-]` widening, so `<h1>`/`<h2>` user text is no longer stripped.

### Changes since last review
Previous HEAD: `2a4906e4` (2026-08-05T11:27:31Z) — verdict COMMENT

- ✅ **Fixed:** the `[a-z0-9-]` widening was dropped; `SYSTEM_TAG_RE` is now `/<([a-z-]+)>[\s\S]*?<\/\1>/g` (`session-projection.ts:57`) — verified the orphan case still strips to `""` while `<h1>`/`<h2>` user text is preserved.

### Findings

**Verdict:** aligned — unchanged from the previous review; the delta is a one-line regex narrowing under `packages/ui/src/`, no architecture page affected.

🔁 No findings from the previous review, and none new — the delta narrows one character class; no import, boundary, layering, or purity change.

### Verdict
APPROVE — the sole prior warning is fixed and nothing new; backreference retained (bug stays fixed), char class narrowed (headings preserved).

---

PR #3175, opened 2026-08-05
fix(ui): stop a terminal inheriting a chat's working dot on switch

### LINKED ISSUE #2778: Switching to an idle terminal session briefly shows it as "working"
## Problem

When a chat session is actively working (its running indicator is lit) and you switch to a different, idle **terminal** session, that terminal session shows the "working" indicator for the first few seconds before correctly settling to idle.

The switched-to session isn't doing anything — the busy state from the session you just left briefly bleeds onto it during the switch, then clears on its own. It's a transient visual glitch, but it misrepresents the terminal as busy when it isn't.

## Proposed solution

When switching sessions, the working/running indicator for the newly-opened session should reflect only that session's own state from the start — no carry-over from the session you were previously viewing.

### Related

- #2771 — terminal sessions falsely flagging unread (same session-status surface)

jjeliga:
Closes #2778

## Problem

Switching from a busy chat session to an idle **terminal** session briefly lit the terminal's "working" dot for a few seconds before it settled to idle on its own.

## Cause

The chat→terminal switch (`mobileResumeSession`) only swaps `sessionId`/`sessionMode`; unlike the chat resume path it never clears `messages`. So on the transition render the projected `busy = hasStreamingAssistant(messages)` is still `true` from the chat you left, while `sessionId` is already the terminal. The busy→`running` mirror effect in `use-acp-session.ts` then stamps the terminal's `running: true` in the list cache. A terminal has no `_meta.platform`, so the server always reports `running: false` for it — the next `session/list` poll (5s) reconciles the flag back, producing the transient flicker.

## Fix

Skip the busy→`running` mirror when the open session is a terminal, so a terminal's status dot reflects only its own state (server `running`, always false) with no carry-over from the previously-viewed chat. Guarding at the mirror covers every terminal-entry path, not just the reported one.

## Testing

- Manually verified: chat working → switch to idle terminal no longer flickers.
- `ui:check:tsc`, `ui:fix:lint`, and `ui:test` (221 passing) all green.

### DIFF
--- a/packages/ui/src/modules/sessions/hooks/use-acp-session.ts
+++ b/packages/ui/src/modules/sessions/hooks/use-acp-session.ts
@@ -1,5 +1,5 @@
 import { SessionMode } from "api-server-api";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 
 import { queryClient } from "../../../query-client.js";
 import { useStore } from "../../../store.js";
@@ -46,7 +46,10 @@ export function useAcpSession(
   }, [busy, setBusy]);
   // Mirror busy into the list cache's `running` so the row keeps its blue dot
   // when it stops being the open session (the poll-fed `running` lags a switch).
+  const prevBusyRef = useRef(busy);
   useEffect(() => {
+    if (prevBusyRef.current === busy) return;
+    prevBusyRef.current = busy;
     if (selectedAgent && sessionId) {
       setSessionRunning(selectedAgent, sessionId, busy);
     }
--- a/packages/ui/src/modules/sessions/hooks/use-acp-session.ts
+++ b/packages/ui/src/modules/sessions/hooks/use-acp-session.ts
@@ -1,3 +1,4 @@
+import { SessionMode } from "api-server-api";
 import { useCallback, useEffect, useState } from "react";
 
 import { queryClient } from "../../../query-client.js";
@@ -38,7 +39,8 @@ export function useAcpSession(
   // Derive busy from the projection instead of explicit setBusy calls in
   // sendPrompt / resume / disconnect paths. The projection owns streaming
   // state on every message, so "any streaming assistant" is authoritative.
-  const busy = hasStreamingAssistant(messages);
+  const busy =
+    sessionMode !== SessionMode.Terminal && hasStreamingAssistant(messages);
   useEffect(() => {
     setBusy(busy);
   }, [busy, setBusy]);
--- a/packages/ui/src/modules/sessions/components/sessions-sidebar.tsx
+++ b/packages/ui/src/modules/sessions/components/sessions-sidebar.tsx
@@ -133,7 +133,7 @@ export function SessionsSidebar({
       s.mode === SessionMode.Terminal
         ? !!s.running
         : isOpen
-          ? busy
+          ? busy || !!s.running
           : !!s.running;
     // Polled approvals cover all sessions; the live store surfaces the open one instantly.
     const needsApproval =


packages/ui/src/modules/sessions/hooks/use-acp-session.ts
```diff
@@ -45,10 +46,10 @@ export function useAcpSession(
   // Mirror busy into the list cache's `running` so the row keeps its blue dot
   // when it stops being the open session (the poll-fed `running` lags a switch).
   useEffect(() => {
-    if (selectedAgent && sessionId) {
+    if (selectedAgent && sessionId && sessionMode !== SessionMode.Terminal) {
```
dam-code-guardian: 🟢 **Suggestion:** this mirror is now terminal-aware, but the `setBusy(busy)` effect five lines up (line 44) is not — on the same chat→terminal switch `messages` is never cleared (`mobileResumeSession`'s terminal branch only calls `setSessionId`), so `hasStreamingAssistant(messages)` stays `true` and `store.busy` keeps reporting the chat you left while a terminal is open.

Nothing visibly regresses today: `sessions-sidebar.tsx:131-137` short-circuits terminals to `!!s.running` before consulting `busy`, and `ChatHeaderStatus` uses `busy` only to *suppress* the contribution-failures badge — so the effect is a briefly hidden badge, not a second spurious dot. But it leaves two representations of the same fact disagreeing while a terminal is open, and that sidebar short-circuit is now the only thing standing between the stale flag and a repeat of this bug one layer up.

Worth considering gating at the source instead — `setBusy(sessionMode === SessionMode.Terminal ? false : busy)` on line 44 — which fixes both consumers at once and lets the sidebar's special-case comment describe a mode distinction rather than a workaround. (No `suggestion` block here: the line it would need to replace sits outside this diff's hunk.)

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `f5fd02a7`

## PR #3175: fix(ui): stop a terminal inheriting a chat's working dot on switch
**Author:** jjeliga | **Branch:** fix/2778-working-indicator-active-for-idle-terminal → main | **Changes:** +3 −2 (1 file)

### Summary
Gates the `busy`→`running` list-cache mirror on `sessionMode !== SessionMode.Terminal`, so switching from a streaming chat to an idle terminal no longer stamps the terminal's row `running: true` until the next 5s poll reconciles it. The diagnosis and mechanism in the body check out: `acp-session-ops.ts:46` decodes `running: p?.running ?? false`, and a terminal has no `_meta.platform`, so the server-side value really is always `false`.

### Findings
- 🟢 **Suggestion:** the store-level `busy` two lines up stays stale on the same transition (`use-acp-session.ts:44`)
- ✅ **Looks good:** the guard sits at the mirror rather than in `mobileResumeSession`, which does cover all four terminal-entry paths (`mobileResumeSession`'s `mode` arg, `pendingTerminal`, `handleNewTerminal`, `openAgentTerminal`) — and the sibling `setSessionRunning(…, true)` at `chat-view.tsx:615` correctly stays unguarded, since that one is the terminal's own optimistic dot on Enter.

Verdict: `aligned`.

✅ No findings. The change stays inside the client package's own `modules/sessions` slice, imports `SessionMode` from the contract package `api-server-api` matching the convention in the six sibling files that already do so, and the effect's dependency array correctly gained `sessionMode`.

### Verdict
APPROVE — the fix is correct, minimal, and its scope claim holds; the one suggestion is a pre-existing adjacent inconsistency, not a defect in this diff.

---

packages/ui/src/modules/sessions/hooks/use-acp-session.ts
jjeliga: Good catch on the shared root. I took the source-gate approach you suggested, but moved it up one line to the `busy` **derivation** itself (line 41) rather than `setBusy` (line 44):

```ts
const busy =
  sessionMode !== SessionMode.Terminal && hasStreamingAssistant(messages);
```

`setBusy` only writes `store.busy`, whose one reader (the sidebar) already short-circuits terminals to `!!s.running` before ever consulting `busy` — so gating there wouldn't change anything observable. The derivation, though, is the common source for all three representations: `store.busy`, the list-cache `running` mirror, and the header's `busy` that gates `ContributionFailuresBadge`. Gating it reconciles all three at once and subsumes the mirror guard, which I've reverted.

One correction worth flagging: the badge suppression isn't "briefly hidden." Switching to a terminal tears down the chat's ACP channel (`use-acp-connection`), so `messages` never updates while the terminal is open — `hasStreamingAssistant` stays true for the *entire* visit, so the old code would suppress `ContributionFailuresBadge` the whole time, not a few seconds. And since that badge reads the hook `busy` rather than `store.busy`, a `setBusy`-only gate wouldn't have fixed it; the derivation gate does.

Updated in c57e5e61.

PR #3178, opened 2026-08-05
feat(skills): show when a skill source was last scanned

### LINKED ISSUE #2827: Show when a skill source was last scanned
## Problem

The design for a skill source card shows a "scanned X ago" timestamp, but there's no such information available today — a re-scan only shows a brief spinner. A user can't tell how fresh a source's skill list is, or whether a re-scan is worth running.

## Proposed solution

Record and display when each skill source was last scanned, shown on the source card as the design intends.

### Dependencies

Deferred from #944 (needs a last-scanned field on the source).

PetrBulanek:
A skill source card is designed to show a "scanned X ago" timestamp so a user can judge how fresh a source's skill list is and whether re-scanning is worth it — but nothing records it today, so a re-scan just flashes a spinner and resolves into no new information. This records when each source was last scanned and renders it on the card, exactly as the design intends: a muted `scanned 2h ago` right-aligned in the header, immediately left of the kebab. A never-scanned or errored source shows nothing.

No migration and no new table: the value rides on the existing in-memory scan cache, which now stamps each entry with the time of the upstream read it came from. A cache hit reports that original read, not the moment of the hit.

`skills.list` keeps its bare `Skill[]` shape — the `dam` CLI is installed on users' own machines and outlives any one server, so an npm-released `dam skill catalog` would break if the procedure grew an envelope. Freshness arrives through a sibling `skills.listWithScan` returning `{ skills, scannedAt }`, which only the UI calls. Both go through one shared helper so the binding check and the scan cannot drift. The MCP `list_skills_in_source` tool is unchanged for agents.

The cache is per-replica and in-memory: the recorded time describes the read that replica performed, and `sources.refresh` invalidation reaches only that replica. At the default `apiserver.replicas: 1` that is the whole cluster.

Closes #2827

## Sub-issues

- [x] 01 — api-server: carry `scannedAt` through the scan cache and the `skills.list` contract
- [x] 02 — UI: show "scanned X ago" on the source card


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `a85739a`

## PR #3178: feat(skills): show when a skill source was last scanned
**Author:** PetrBulanek | **Branch:** feat/2827-source-last-scanned → main | **Changes:** +87 −21 (12 files)

### Summary
Records the upstream read time in the skill scan cache and widens the `skills.list` contract from `Skill[]` to `{ skills, scannedAt }`, so the source card can render a muted "scanned X ago". A never-scanned or errored source shows nothing.

### Findings
- ✅ **Looks good:** The `skills.list` output widened from a bare array to an object — a change that breaks every consumer. All four are correctly migrated: the tRPC router adopts the new schema, the MCP tool and CLI catalog keep emitting the bare array (contract preserved), and the UI hook reads both fields.

Verdict: **aligned**. `docs/architecture/skills.md` was updated in the same PR to describe both the scan-cache timestamp and the widened `skills.list` contract. `Last verified: 2026-08-05` already matches the PR date. No ADR leak, no volatile-content leak.

### Issue Fit
Linked issue: #2827 "Show when a skill source was last scanned" (closing keyword)

✅ Delivers the issue's ask (record + display last-scanned time on the source card); no undeclared scope.

✅ No findings. Idiomatic tRPC/Zod contract widening — new Zod schema in the contract package, epoch-ms→ISO mapping done at the application-layer service boundary, adapters preserving their own shapes.

✅ No findings. Minimal, typed prop addition; `scannedAtBySource` mirrors the existing `skillsBySource`/`errorBySource` state maps and stays in sync on the success path (error path sets `[]` and hides the stamp via `!error`; delete clears both).

### Verdict
**APPROVE** — correct, well-scoped, and fully documented; the contract widening is safe across all consumers.

---

docs/architecture/skills.md
```diff
@@ -125,7 +125,7 @@ The ordering lives in one place shared by both scanners and the install-time res
 Lives in [`packages/api-server/src/modules/skills/`](../../packages/api-server/src/modules/skills/). Owns:
 
 - The **Skill Source catalogue** — CRUD on user sources, merging in system seeds and template sources, dedupe and badge resolution.
-- The **scan cache** — per-`gitUrl`, 5-minute TTL, invalidated on `sources.refresh` or after a successful publish to that source.
+- The **scan cache** — per-`gitUrl`, 5-minute TTL, invalidated on `sources.refresh` or after a successful publish to that source. Each entry also records the time of the upstream read it came from, and `skills.listWithScan` carries that timestamp so a source card can show how fresh its list is. The cache is empty after a restart, so the value never claims a read that didn't happen.
```
kapetr: "The cache is empty after a restart, so the value never claims a read that didn't happen" is a guarantee the code does not make under a supported configuration. `apiserver.replicas` is a knob (`deploy/helm/platform/values.yaml:586`) and the api-server Service has no client affinity — `docs/architecture/platform-topology.md:50` pins only the harness Service. `sharedScanCache` (`compose.ts:46`) and `invalidateScanCache` are per-pod, so a `sources.refresh` handled by pod A does not clear pod B. A user clicks Re-scan, the follow-up `listWithScan` (`use-skills-surface.ts:402`) lands on pod B, and the card reports "scanned 4m ago" over an unchanged list immediately after the user asked for a fresh read. The stale-list part is pre-existing; turning it into a visible freshness claim is new.

docs/architecture/skills.md
```diff
@@ -234,7 +234,7 @@ GitHub errors (missing scope, repo not found) surface to agent-runtime as the up
 
 `skills.sources.list(agentId?)` merges the three Source kinds and returns `canPublish: true` only for GitHub URLs.
 
-`skills.list(sourceId, agentId?)` resolves the source and dispatches:
+`skills.list(sourceId, agentId?)` returns the bare `Skill[]`; `skills.listWithScan(sourceId, agentId?)` returns `{ skills, scannedAt }` — the same list plus when it was read from upstream, so a source card can render how fresh it is. A cache hit reports the original read; a miss stamps the scan it just performed. Two shapes over one read because the `dam` CLI is installed on users' own machines and outlives any one server: `list` is the array an npm-released CLI pins, so freshness arrives as a sibling procedure rather than by widening it. Both resolve the source and dispatch:
```
kapetr: The added sentence "Two shapes over one read because the `dam` CLI is installed on users' own machines and outlives any one server: `list` is the array an npm-released CLI pins, so freshness arrives as a sibling procedure rather than by widening it" is decision history — an alternative weighed and rejected. The same line transcribes code-level shape (`{ skills, scannedAt }`). Both are excluded from architecture pages, which must carry only operational "why" and describe protocols semantically.

packages/api-server/src/modules/skills/compose.ts
```diff
@@ -29,6 +29,11 @@ const CACHE_TTL_MS = 5 * 60 * 1000;
 interface CacheEntry {
   skills: Skill[];
   expiresAt: number;
+  /** Epoch ms of the upstream read this entry came from — carried to the UI so
+   *  a source card can show how fresh its list is. An empty cache after a
+   *  restart forces a real scan, so this never claims a read that didn't
```
kapetr: The comment "An empty cache after a restart forces a real scan, so this never claims a read that didn't happen" explains why persistence is absent. `router.ts:73-76` does the same for the shape that was not widened, and `mcp-endpoint.ts:534-535` explains why the timestamp is not given to agents. Comments must describe what the code does, not what it does not do. The router comment also points at "the sibling read below", which rots the moment a procedure is inserted between them.

packages/api-server/src/modules/skills/compose.ts
```diff
@@ -49,17 +54,22 @@ async function scanWithCache(
   gitUrl: string,
```
kapetr: The whole correctness claim of this feature lives in `scanWithCache`: a hit must report the original read, a miss must stamp the new one, and a throwing scanner must not overwrite or serve the previous entry. The function is module-private and its state is a module-scope mutable `Map`, so it cannot be unit-tested without exporting a seam. No test in the PR covers any of it, and `packages/api-server/src/__tests__/unit/` has none for the cache either.

packages/ui/src/modules/sandboxes/components/skills/skill-source-card.tsx
```diff
@@ -182,6 +188,14 @@ export function SkillSourceCard({
           </p>
         </div>
         <div className="flex shrink-0 items-center gap-2">
+          {!error && scannedAt && (
+            <span
+              className="shrink-0 text-sm text-muted-foreground"
+              title={formatTimestamp(scannedAt)}
+            >
+              scanned {timeAgo(scannedAt)}
```
kapetr, line 200: `timeAgo(scannedAt)` is rendered with no timer, so the label has no source of its own re-renders. It currently ticks only because `use-skills-surface.ts:198` polls `skills.state` every 5 s and `setInstalled(state.installed)` always receives a new array identity, re-rendering the whole surface. `SkillSourceCard` is a plain function component, so wrapping it in `memo`, or moving that poll to react-query with structural sharing, freezes the label at whatever it said on mount.

packages/api-server-api/src/modules/skills/schemas.ts
```diff
@@ -49,6 +49,14 @@ export const skillSchema = z.object({
   dir: z.string().optional(),
 });
 
+/** A source's scanned skill list plus when that scan was read from upstream. */
+export const skillListResultSchema = z.object({
+  skills: z.array(skillSchema),
+  /** ISO 8601 time the source's skill list was last read from upstream. A
+   *  cache hit reports the original read, not the moment of the hit. */
+  scannedAt: z.string(),
```
kapetr: `scannedAt: z.string()` accepts any string; `api-keys`, `metrics`, and `experiments` schemas all use `z.string().datetime()`, and an unparseable value renders as "scanned —".

packages/api-server-api/src/modules/skills/router.ts
```diff
@@ -55,15 +70,19 @@ export const skillsRouter = t.router({
       .mutation(({ ctx, input }) => ctx.skills.refreshSource(input.id)),
   }),
 
+  // The bare array is a published shape: `dam` CLIs are installed on users'
+  // own machines and outlive any one server, so an npm-released `dam skill
+  // catalog` would break on an object here. Freshness rides on the sibling
+  // read below instead.
   list: readAgentProcedure
```
kapetr, line 77: `list` is frozen for wire compatibility but carries no `@deprecated` marker, so the next UI author has nothing but a comment steering them to `listWithScan`.

kapetr (commented):
Two pre-existing cache problems worth their own issues rather than this PR: the scan cache is keyed on `(gitUrl, path)` with no owner, so one owner's authenticated scan of a private repo is served to any other owner who registers the same URL (`createSource` has no allow-list); and expired entries are never evicted and the map has no size cap. Both at `compose.ts:46`.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `a64c6dc`

## PR #3178: feat(skills): show when a skill source was last scanned
**Author:** PetrBulanek | **Branch:** feat/2827-source-last-scanned → main | **Changes:** +304 −69 (13 files)

### Summary
Re-review after three commits that answer reviewer @kapetr's post-approval feedback: the CLI-facing `skills.list` stays a bare array (freshness moved to a sibling `listWithScan`), the scan cache is extracted behind a testable factory with unit tests, and the architecture-doc freshness claim is corrected.

### Changes since last review
Previous HEAD: `a85739a` (2026-08-05T15:07:58Z) — verdict APPROVE

- ✅ **Fixed:** Scan cache extracted to a testable `createScanCache` factory (closure state, no module-scope `Map`) with unit tests covering all three stamp invariants (`scan-cache.ts`, `skills-scan-cache.test.ts`).
- ✅ **Fixed:** `skills.list` restored to a bare `Skill[]` for published CLIs; freshness served by a new `listWithScan`, both sharing one `listSkills` helper so the binding check and scan cannot drift (`router.ts`).
- ✅ **Fixed:** Architecture-doc freshness claim scoped to the per-replica cache; rejected-alternative narration and the literal result shape removed (`docs/architecture/skills.md`).
- ✅ **Fixed:** "What it does not do" comments rewritten to describe behavior in `compose.ts`, `mcp-endpoint.ts`; `list`/`listWithScan` steering moved to JSDoc naming the sibling.

### Findings
- 🟢 **Suggestion:** `scannedAt: z.string()` is looser than sibling schemas' `z.string().datetime()` (`packages/api-server-api/src/modules/skills/schemas.ts:57`). The service always emits ISO via `toISOString()`, so no runtime effect today; tightening it lets the boundary reject a bad value instead of the UI rendering "scanned —". (@kapetr's open inline finding.)
- 🟢 **Suggestion:** "scanned X ago" has no independent refresh source (`packages/ui/src/modules/sandboxes/components/skills/skill-source-card.tsx:196`). It re-renders only because the installed-state poll fires every 5 s; memoizing the card or moving that poll to react-query with structural sharing would freeze the label. (@kapetr's open inline finding.)

Verdict: **minor drift**.
- ❌ Check 2 (`Last verified` staleness): the page's technical claim was edited on 2026-08-06 (commit `a64c6dc`) but `docs/architecture/skills.md:3` still reads `Last verified: 2026-08-05`. Bump to `2026-08-06`.

Checks 1, 3–7 aligned: the scan-cache behavior change is reflected on the Skills page, no ADR leak, and the edit removed volatile content (the literal result shape) rather than adding it.

### Issue Fit
🔁 1 finding-free result from the previous review still holds: delivers #2827's ask (record + display last-scanned time on the source card); the CLI array-shape restore and cache extraction are mechanical hardening, no undeclared scope.

One 🟢 (the `scannedAt` schema looseness above). Otherwise idiomatic: the cache extraction lands in `infrastructure/`, the epoch-ms→ISO map stays at the service boundary, the contract widening keeps every adapter's shape, and the new tests exercise the invariants directly.

One 🟢 (the label-refresh coupling above). Otherwise clean: `scannedAtBySource` mirrors the existing `skillsBySource`/`errorBySource` maps, stays in sync on success, clears on delete, and is hidden while errored.

### Verdict
**APPROVE** — the three commits resolve the substantive post-approval findings; the two remaining items are minor suggestions with no runtime effect, and the doc drift is a one-day `Last verified` bump.

---

packages/api-server-api/src/modules/skills/router.ts
PetrBulanek: Leaving `list` unmarked deliberately. It is not deprecated — it is the frozen shape published CLIs pin, and `@deprecated` would strike it through in the editor for exactly the authors who must keep calling it. The JSDoc added in 17eb11cf names `listWithScan` for callers that also want freshness, which is the steer a UI author needs without mislabelling the CLI contract.

The other two are addressed: `scannedAt` is now `z.string().datetime()` (438b76ac), and the render site carries a note that the label has no timer of its own and ticks on the 5s installed-state poll, so a later memo or react-query migration does not silently freeze it (db802abb). Same pattern already exists at `artifact-row.tsx:92`, so a shared tick is worth doing across both rather than one-off here.

Not filing the three cache issues (owner-less key, no eviction, per-replica invalidation) as part of this PR.

PR #3180, opened 2026-08-05
fix(deploy,ci): keycloak rolls without downtime and only when changed

### LINKED ISSUE #3176: Keycloak rollouts: zero downtime, and no restart when nothing changed
## Context

Every release bumps the single shared platform version, and the Keycloak image tag defaults to it — so every release rolls the Keycloak pod even when the image content (upstream Keycloak + login theme) didn't change, which is most releases. The Deployment runs a single replica with a Recreate strategy, so each roll is a hard outage: the old pod is killed before the new one starts, and the new one takes ~30–60s to become ready.

## Problem

During every release there's a window where sign-ins fail and API requests needing fresh token verification get "auth unavailable" — for no reason, since the Keycloak image is almost never actually different.

Two independent gaps:

- **Rollouts aren't zero-downtime** — Recreate with one replica guarantees a window with no Keycloak running at all.
- **Keycloak restarts without cause** — the image tag tracks the release version, so a new tag with identical content rolls the pod every release.

## Goal

A release that doesn't change Keycloak leaves it untouched. A release that does change it rolls it with zero downtime — logins and token verification keep working throughout.

## Done when

- Releasing with no Keycloak-affecting change does not restart the Keycloak pod.
- A Keycloak image change rolls out with no failed logins and no auth-unavailable errors (verified by exercising login and a protected API during the rollout).

## Open Questions

- Rolling update (surge 1, max unavailable 0) means two instances briefly run against the shared DB — needs a check that Keycloak's caching/session setup tolerates that (recent Keycloak persists user sessions in the DB, but confirm rather than assume).
- Versioning: a separately maintained Keycloak version, or reuse CI's existing content-addressed build-or-reuse mechanism so the tag only moves when the image inputs change? The latter avoids a second version to bump by hand.

jezekra1:
Fixes #3176

## What

- **Zero-downtime rolls:** the keycloak Deployment switches from `Recreate` to `RollingUpdate` (maxSurge 1, maxUnavailable 0). The old pod serves logins and JWKS until the new one is ready. In production mode (`start`, `scheme: https`) the overlapping pods form an Infinispan cluster via jdbc-ping through the shared DB, so even in-flight login flows survive; `start-dev` (local/e2e) may bounce an in-flight login once. A failed startup (e.g. schema migration) now leaves the old pod serving instead of taking auth down — strictly better than Recreate.
- **No restart when nothing changed:** two causes removed.
  - The pod template no longer includes `platform.labels`, whose `helm.sh/chart: platform-<version>` label rolled the pod on every release regardless of image. It now hand-writes `component` + `instance`, the same deliberate omission postgres/redis/seaweedfs already use.
  - The image tag no longer tracks the release version. `resolve-image.sh` gains keycloak as a component (`packages/keycloak-theme` + `packages/dev-config` + `pnpm-lock.yaml`); CD tags the image with its source sha (non-PR runs only, so a PR's amd64-only manifest can never overwrite it), and the chart publish step pins that tag into `values.yaml` (anchored on the `@ci-pin keycloak-image-tag` marker, asserted with a grep so drift fails loud). Charts rendered from git keep the `.Chart.AppVersion` fallback.

## Verification

Live two-upgrade proof on the dev cluster, with an in-cluster probe curling the JWKS endpoint every 0.5s throughout:

1. Upgrade to this chart — keycloak rolls once (expected: its pod template loses the version label), under RollingUpdate. **330/330 probe requests OK, zero failures.**
2. Upgrade again with only the chart version bumped — keycloak pod untouched (same UID), while ui rolled as before.

Also: `helm template` at two chart versions with the tag pinned renders a byte-identical keycloak pod template; `resolve-image.sh --self-check` extended and green; `mise run check` green.

## Notes

- The first release carrying this change rolls keycloak once (pod template change) — under the new rolling strategy, so no downtime.
- Lockfile-touching releases still re-tag (and roll) keycloak even when the theme deps didn't change — accepted: errs toward rebuilds, still kills the every-release restart.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -175,7 +175,7 @@ jobs:
 
   merge-images:
     name: merge multi-arch manifests
-    needs: [build-images, appversion, check]
+    needs: [changes, build-images, appversion, check]
     runs-on: ubuntu-latest
     timeout-minutes: 10
     strategy:
@@ -197,12 +197,17 @@ jobs:
         uses: docker/metadata-action@80c7e94dd9b9319bd5eb7a0e0fe9291e23a2a2e9 # v6.1.0
         with:
           images: ${{ env.IMAGE_PREFIX }}/${{ matrix.component }}
+          # keycloak also gets its source-sha tag (see merge-platform-base):
+          # publish pins it into the chart so releases that don't change the
+          # keycloak image don't roll the Keycloak pod. Non-PR runs only —
+          # a PR's amd64-only manifest must never overwrite it.
           tags: |
             type=sha,prefix=,format=long
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=raw,value=${{ needs.appversion.outputs.composite }}
             type=raw,value=latest,enable=${{ startsWith(github.ref, 'refs/tags/v') }}
+            type=raw,value=${{ fromJSON(needs.changes.outputs.source_shas)['keycloak'] }},enable=${{ matrix.component == 'keycloak' && github.event_name != 'pull_request' }}
       - name: Create manifest list and push
         env:
           IMAGE: ${{ env.IMAGE_PREFIX }}/${{ matrix.component }}
@@ -613,7 +618,7 @@ jobs:
           if-no-files-found: ignore
 
   publish:
-    needs: [merge-images, merge-agents, merge-workloads, appversion]
+    needs: [changes, merge-images, merge-agents, merge-workloads, appversion]
     if: github.event_name == 'push'
     runs-on: ubuntu-latest
     timeout-minutes: 15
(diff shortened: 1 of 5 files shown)


JanPokorny (approved):

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `1af88ea8`

## PR #3180: fix(deploy,ci): keycloak rolls without downtime and only when changed
**Author:** jezekra1 | **Branch:** fix/keycloak-zero-downtime#locki-lpaitkhu → main | **Changes:** +51 −7 (5 files)

### Summary
Switches the Keycloak Deployment from `Recreate` to surge-first `RollingUpdate` (maxSurge 1 / maxUnavailable 0) and pins the chart's keycloak image tag to its content sha in CI, so a release that doesn't change the image no longer restarts Keycloak.

### Findings
- ✅ **Looks good:** No-downtime and the pod-label change both verified safe — the readiness probe (`/health/ready`) gates the surge so the old pod keeps serving until the new one is ready; overlapping pods auto-cluster via Keycloak 26.6's default jdbc-ping over the shared DB (so the in-flight-flow-survival claim holds); and dropping `platform.labels` from the pod template (keeping `component`+`instance`) still satisfies the Service selector, matches the postgres/redis precedent, and no NetworkPolicy selects keycloak pods. CI pin is guarded (`[ -n "$KC_TAG" ]`) and its `sed` is verified by a following `grep`.

**Verdict:** aligned — the deployment-strategy switch is Helm/infra-level detail the documentation guidelines keep out of architecture pages; Keycloak's documented responsibility (identity authority, JWKS, login) is unchanged, and the only touched doc (`release-process.md`) is out of doc-drift's architecture-only scope.

### Verdict
APPROVE — a clean, strictly-better availability change (RollingUpdate can only improve on Recreate here), with `resolve-image.sh` adding keycloak consistently and its `self_check` assertions extended to cover it.

---

PR #3182, opened 2026-08-05
fix(skills): repair the pr-state resolver and flatten its scheduling

PetrBulanek:
Follow-up to #3139. The resolver on `main` is dead — every tick fails at the driver before reading anything — and the review rounds on this PR's earlier revisions surfaced real design problems in the scheduling machinery itself. The final shape is simpler than either reviewed revision: the machinery is removed rather than repaired.

## Why the resolver on `main` is dead

The candidate query passes a `Date` into a raw ``sql`…${now}` `` template. postgres-js **by itself serializes a `Date` fine** (OID 1184 → `toISOString()`) — a bare-client repro shows the query working — but production runs `drizzle(client)`, whose constructor replaces the client's date serializers with identity pass-throughs, and raw ``sql`` `` params skip drizzle's own column mappers. The `Date` reaches the wire encoder raw and throws `TypeError … Received an instance of Date` at `Bind`, on every tick, with any row set. The `power(2, failures)` overflow at 2^1024 was real too (confirmed on PostgreSQL 18.4) but secondary — no row on `main` could have reached `failures = 1024`.

This PR removes the trap's surface rather than annotating it: the interval math moves to JS (`lte(col, staleBefore)` through the column mapper), so no timestamp crosses a raw template at all.

## Resolution is now one flat hourly pass

The backoff/retirement/marker machinery this PR previously carried (and its second revision extended into lanes) was sized for hundreds of concurrently unresolved publish records. Observed populations are single digits — every record requires a deliberate publish that opened a real PR — and the flat design's budget holds to ~50 concurrent unresolved records, orders of magnitude above reality. Notably, the backoff manufactured one of the two shipped bugs (`power()` exists only to serve it). So:

- **One candidate query**: non-terminal, never-attempted or last attempted over an hour ago, oldest first (`NULLS FIRST` so fresh publishes go first), ten per ten-minute tick — exactly the anonymous 60/h ceiling.
- **One pass**: anonymous read; a `404` (private as much as gone) escalates inline to the publishers' own pods — **warm pods only, never waking an agent** (the `state`/`not-running`/`failed` result union stays); every attempt stamps the clock and the record waits its hour.
- **Removed**: exponential backoff, failure retirement, the `pr_needs_pod` marker and its migration (0021 — the PR now ships **no** migration), the pod lane, and the warmth-list port. The 0020 `pr_state_check_failures` column stays but is dormant.
- **Drain**: resolution (terminal states leave the candidate set) or agent deletion (the skills cleanup saga removes an agent's records). A permanently dead record costs one page slot and at most one anonymous read per hour until its agent is deleted — which starts to matter only near ~50 concurrent dead records.

Every finding from both review rounds is resolved by removal: nothing starves (no lanes), nothing fails to retire (no retirement), no marker is permanent (no marker), and the drain claim in the docs is now simply true.

## Also

- `classifyOrigin` shares the stat-fingerprint cache the `contentHash` path already uses (unchanged from the first revision).
- Stale comments (`slice 04` pointers, the required-port argument, the Date-throw mechanism) rewritten; `docs/architecture/skills.md` resolution section rewritten for the flat design.

## Verification

- 775 api-server unit tests, `mise run check` (including the schema↔migrations gate after dropping 0021), full `mise run test` aggregate — green.
- Smoked end to end on the dev cluster with the flat build deployed: a fresh record on a real merged public PR resolved `merged` on its first tick (anonymous, ETag stored); a private-repo PR 404'd anonymously and resolved `merged` through its warm publisher's pod in the same attempt; a junk URL was stamped and waits its hour, no state written.
- Contract and UI untouched; badge semantics unchanged — unresolvable still just claims less.


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `4d6da2df`

## PR #3182: fix(skills): repair the pr-state resolver's query, backoff, and hash reuse
**Author:** PetrBulanek | **Branch:** fix/3019-pr-state-followups → main | **Changes:** +2896 −79 (11 files)

### Summary
Follow-up to #3139: repairs the pr-state resolver's dead candidate query (raw `sql` Date bind + `power()` overflow both silently stopped every tick), adds failure-retirement and a `deferred` outcome so cold pods don't punish a record, marks anonymous-404 records `pr_needs_pod` to skip wasted anonymous reads, and reuses the stat-fingerprint cache in `classifyOrigin`.

### Findings
- ✅ **Looks good:** Both DB-level fixes verified safe — the backoff caps the `power()` *exponent* (`least(failures, 5)` ⇒ 2^5 = 32 ≥ the 24h cap) so no `WHERE`-order-independent overflow can stop the query, and migration 0021 adds `pr_needs_pod` as `NOT NULL DEFAULT false` (constant default ⇒ metadata-only, no rewrite of existing rows). Per-tick GitHub spend stays bounded: the pod/`prNeedsPod` paths are gated by `candidates.slice(0, MAX_READS_PER_TICK)`.

**Verdict:** aligned — `docs/architecture/skills.md` was updated in this PR to match the new resolver behaviour (anonymous-404 marking, `deferred` vs `failed`, retirement); `Last verified: 2026-08-05` matches the PR date, no ADR or volatile-content leaks.

✅ No findings. The change improves domain modelling — the port's `PrDisposition | null` return becomes a `PodPrReadResult` tagged union (`state` / `not-running` / `failed`) carried as a value, and layer dependencies (services → domain, infrastructure → domain) are unchanged.

### Verdict
APPROVE — a well-scoped, cluster-verified repair; the `deferred`/`failed` distinction and the `prNeedsPod` fast-path are both covered by the updated unit tests.

---

packages/api-server/src/modules/skills/infrastructure/agent-skills-repository.ts
```diff
@@ -210,21 +220,23 @@ export function createAgentSkillsRepository(db: Db): AgentSkillsRepository {
...
+            // Two traps: the inner least() caps the *exponent* (power() is
+            // double precision and overflows at 2^1024, failing the whole
+            // query — WHERE conditions have no guaranteed evaluation order,
+            // so the retirement bound cannot shield it), and `now` crosses as
+            // an ISO string (raw sql`` params skip drizzle's column mappers
+            // and postgres-js throws on a Date instance).
             or(
               isNull(agentSkillPublishes.prStateCheckedAt),
-              sql`${agentSkillPublishes.prStateCheckedAt} + least(power(2, ${agentSkillPublishes.prStateCheckFailures}), 24) * interval '1 hour' <= ${now}`,
+              sql`${agentSkillPublishes.prStateCheckedAt} + least(power(2, least(${agentSkillPublishes.prStateCheckFailures}, 5)), 24) * interval '1 hour' <= ${now.toISOString()}`,
```
kapetr: **The Date diagnosis does not hold.**

The comment says postgres-js throws on a `Date`. It does not. drizzle passes the `Date` to `client.unsafe`, postgres-js infers type `1184`, and `Bind` serializes it with `toISOString()`.

Measured on `postgres:18-alpine` with postgres-js 3.4.9, the pre-fix expression:

| Rows | Result |
|---|---|
| `failures` 0 / 7 / null, `Date` param | OK |
| one row at `failures = 1023` | OK |
| one row at `failures = 1024` | `value out of range: overflow` |

The failure was the overflow. Keep `toISOString()` if you like it explicit, but drop the claim about the throw.

packages/api-server/src/modules/skills/services/resolve-pr-state.ts
```diff
@@ -162,22 +151,66 @@ export function createPrStateResolver(deps: {
...
+        etag: null,
+      });
+      return 1;
+    }
+    if (result.kind === "failed") sawWarmFailure = true;
+  }
+  await deps.agentSkills.touchPrState(
+    candidate.prUrl,
+    now,
+    sawWarmFailure ? "failed" : "deferred",
```
kapetr: **A deferred record never retires, and starves live PRs sooner.**

`deferred` leaves `prStateCheckFailures` alone, so a `pr_needs_pod` record with always-cold agents stays eligible hourly forever. It never backs off and never retires.

The cost is the per-tick page, not GitHub budget:

- deferred forever: 24 slots/day → **~60 records saturate the resolver**
- before this PR: `failed` → 1 slot/day → ~1440 to saturate

So private records now starve public ones 24× sooner.

`MAX_READS_PER_TICK` bounds anonymous reads, and pod-only records make none. Give them their own allowance instead of the same 10-row page.

docs/architecture/skills.md
```diff
@@ -100,7 +100,7 @@ A Local Skill whose directory (with a `SKILL.md`) also exists in a pristine root
...
...o the pods — when several agents published the same pull request each of their pods is tried until a warm one answers. That escalation **never wakes a hibernated agent** — a badge is not worth spending the user's compute unasked — which is acceptable because `merged` and `closed` are terminal: once observed they are persisted and never re-read, so the badge only ever gets more accurate. Cost is held inside the anonymous 60-requests/hour-per-IP ceiling by excluding terminal states from the candidate query and re-checking any one record at most hourly. A record whose attempts keep *failing* backs off exponentially to at most daily and is eventually retired outright, so a permanently broken record — a deleted repo, a URL that is not a pull request — drains out of the working set instead of holding a daily slot forever; an attempt that found only cold pods is deferred, not failed, since it learned nothing about the pull request — the record keeps its hourly cadence (now costing nothing anonymous) and so keeps its chance of catching the agent awake. Conditional requests are **not** part of that budget: an anonymous `304` is charged exactly like a `200`, so the ETag saves bandwidth only.
```
kapetr: **The drain claim is wrong for a deleted repo.**

A junk URL drains — `parsePrUrl` fails, so it retires at 30.

A deleted repo does not, when the publishers are cold: `404` → `markPrNeedsPod` → cold pods → `deferred` → `failures` stays 0 → hourly forever. Your verification section shows this exact row.

It drains only if a publishing agent is warm. Fix it or qualify the sentence.

packages/db/src/schema.ts
```diff
@@ -314,14 +314,15 @@ export const agentSkillPublishes = pgTable(
...
-     *  it toward one read a day. Reset whenever an attempt learns anything. */
+    /** Consecutive failed attempts — the backoff multiplier and, past a
+     *  bound, the retirement gate. Deferred attempts (no warm pod) don't
+     *  count; reset whenever an attempt learns anything. */
     prStateCheckFailures: integer("pr_state_check_failures")
       .notNull()
       .default(0),
+    /** Set once the anonymous read has 404'd; later attempts go straight to
+     *  a publishing agent's pod. Never unset. */
+    prNeedsPod: boolean("pr_needs_pod").notNull().default(false),
```
kapetr: **`pr_needs_pod` is permanent, on an ambiguous signal.**

One `404` sets it, and nothing unsets it. But `404` also means a renamed repo, a brief outage, or a deleted public PR. Any of those locks the record out of the anonymous path for good, in the never-retiring deferred class.

There is no privacy flag on `skill_sources`, so the inference is fair. Consider an anonymous re-probe every Nth attempt — or say in the comment that this is deliberately permanent.

packages/api-server/src/modules/skills/services/resolve-pr-state.ts
```diff
@@ -51,8 +51,8 @@ import { parsePrUrl } from "../domain/pr-url.js";
  *  than leaving headroom under it, so a saturated backlog spends the whole
  *  budget and depends on the `rate-limited` early return below to stay correct
  *  at the boundary — and any other anonymous GitHub call sharing this egress
- *  IP can tip such a tick into 429s. Terminal-state exclusion and the failure
- *  backoff are what drain the backlog over time. */
+ *  IP can tip such a tick into 429s. Terminal-state exclusion and failure
```
kapetr: **Two comments this PR invalidated.**

- Line 54: "failure retirement" drains the backlog — true for the failed path only. Deferred records do not drain.
- Line 67, just below: the required-port argument still says the adapter returns null for a stopped agent or a failed call. This PR replaced that with a union and split the two cases. The argument holds; the evidence is stale.

kapetr (commented):
Verified the fixes. The exponent cap and the `not-running` / `failed` split are right. Two things need a change, plus a correction to the description.

**The `Date` was never the problem.** Against `postgres:18-alpine` with this repo's postgres-js 3.4.9, the pre-fix query runs fine with a `Date` param. It fails only at `failures = 1024` — the overflow. Your smoke test seeded `failures = 2000`, which reproduces that. No row on `main` can have reached 1024, so the resolver on `main` is not dead. Post the stack trace if you have one and I will withdraw this.

**The SQL seam is still untested.** You name this yourself. It now holds the exponent cap, the retirement bound, the `NULLS FIRST` order, and three `touchPrState` branches — and both real bugs lived there. A throwaway `postgres:18-alpine` container exercised all of it in under a minute.

Nits: `0021_groovy_lizard.sql` has no trailing newline (0020 does), and the `slice 04` pointers in `agent-skills-repository.ts:64` and `resolve-pr-state.ts:32` lead nowhere now.

packages/api-server/src/modules/skills/infrastructure/agent-skills-repository.ts
PetrBulanek: Both measurements are right — yours ran a bare postgres-js client, where `inferType` → 1184 serializes the `Date` fine. Production runs `drizzle(client)`, whose constructor overwrites the client's date serializers with identity pass-throughs (`drizzle-orm/postgres-js/driver`, `construct()`), and raw ``sql`` `` params also skip drizzle's own column mappers — so the `Date` reaches `Bind` raw:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "string" argument must be of type string or an
instance of Buffer or ArrayBuffer. Received an instance of Date
    at Buffer.byteLength (node:buffer:850:11)
    at reset.str (postgres/src/bytes.js:22:27)
    at Bind (postgres/src/connection.js:954:16)
    at prepared (postgres/src/connection.js:209:7)
```

Reproduced with this repo's `createDb()` against postgres:18-alpine — fails on any row set, including an empty table. The final revision removes the trap's surface instead of annotating it: the interval math moved to JS (`lte(col, staleBefore)` through the column mapper), so no timestamp crosses a raw template at all (d7cc426d).

packages/api-server/src/modules/skills/services/resolve-pr-state.ts
PetrBulanek: Conceded — and after a further altitude check, resolved by removing the machinery rather than budgeting it (d7cc426d). No pod lane, no `deferred`, no backoff: one flat hourly pass, a 404 escalates inline to warm publishers' pods only, every attempt stamps the clock. The population where any starvation arithmetic bites (~50+ concurrently unresolved records) is orders of magnitude above anything observed; scheduling returns if reality ever demands it.

docs/architecture/skills.md
PetrBulanek: Now trivially true, by removal (d7cc426d): there is no retirement to misdescribe. The docs state the two drain paths that exist — resolution (terminal states leave the candidate set) and agent deletion (the cleanup saga). A dead record costs one page slot and ≤1 anonymous read per hour until then.

packages/db/src/schema.ts
PetrBulanek: Moot in the final revision — the marker is removed along with migration 0021 (d7cc426d); the PR now ships no migration. A private record simply re-404s anonymously once an hour (1 read/h — budget noise below ~50 unresolved records) and escalates inline to warm publishers' pods, so there is no permanent bit left to argue about.

packages/api-server/src/modules/skills/services/resolve-pr-state.ts
PetrBulanek: All rewritten again with the flatten (d7cc426d); the 0021 newline nit died with the migration itself.

PetrBulanek:
Final shape after one more altitude check: the scheduling machinery is removed rather than repaired (d7cc426d) — one flat hourly pass; no backoff, retirement, `pr_needs_pod`, lanes, or migration; a 404 escalates inline to warm publishers' pods, never waking one. Rationale in the PR description.

Smoked end to end with the flat build on the dev cluster, one tick: a real merged public PR → `merged` (anonymous, ETag stored); a private-repo PR → 404 → `merged` through its warm publisher's pod in the same attempt; a junk URL → stamped, waits its hour. The seam findings (Date bind, `power()` overflow) were verified against real postgres through a throwaway harness during development; the Date trap's surface is gone entirely now — the interval math lives in JS and goes through the column mapper.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `d7cc426d`

## PR #3182: fix(skills): repair the pr-state resolver and flatten its scheduling
**Author:** PetrBulanek | **Branch:** fix/3019-pr-state-followups → main

### Summary
The scheduling machinery approved in the previous revision is removed rather than repaired: pr-state resolution is now one flat hourly pass — anonymous read, 404 escalates only to already-warm publisher pods, every attempt stamps the clock.

### Changes since last review
Previous HEAD: 4d6da2d (2026-08-05T13:33:08Z) — verdict APPROVE

- ✅ **Fixed:** Dead candidate query repaired for real — the interval math moved to JS and the `Date` reaches drizzle's column mapper via `lte(col, staleBefore)`; the raw `sql` template and the `power(2, failures)` overflow are both gone (`packages/api-server/src/modules/skills/infrastructure/agent-skills-repository.ts:199`).
- ✅ **Fixed:** Scheduling flattened — exponential backoff, failure retirement, the `pr_needs_pod` marker and migration 0021, the pod lane, and the `deferred` outcome are removed (verified absent from the tree; no migration ships). This resolves the earlier-round review concerns (never-retiring `deferred`, permanent `pr_needs_pod` on an ambiguous 404, the false drain claim) by removal (`packages/api-server/src/modules/skills/services/resolve-pr-state.ts`).

### Findings

Verified: the pod path never wakes a hibernated agent (gated strictly on `computeAgentState === "running"`, returns the `not-running`/`failed`/`state` union with no `ensureReady` call); keeping the ETag across a failed attempt is safe because the only writer of `prEtag` also writes the state it validates.

**Verdict:** aligned — `docs/architecture/skills.md`'s resolution section was rewritten in this PR to match the flat design; `Last verified: 2026-08-05` matches the edit date; no ADR or volatile-content leaks.

### Issue Fit
_No linked issue — #3139 is a PR and the branch-inferred #3019 is already closed (resolved by #3124); scope checked against the PR description._ Delivers the described repair and flatten; no undeclared scope.

✅ No findings. The change improves domain modelling — the pod read's `PrDisposition | null` sentinel becomes a `PodPrReadResult` tagged union carried as a value in the domain layer — and the `services → domain ← infrastructure` layering is preserved.

### Verdict
APPROVE — the flat design is correct and cluster-verified; every earlier-round concern is resolved by removal, with no new findings.

---

PR #3183, opened 2026-08-05
feat: follow a Slack thread into its session in the UI

pilartomas:
Part of #3087 — the two blockers it lists under Open Questions, not the "keep replies out of Slack" behaviour itself.

One correction to that framing: the issue says a non-owner following the link "hits an error screen" today. They don't — they get an overlay that says "Loading agent…" and stays there, because the surface waits on a lifecycle state that never arrives for an agent it can't see. The error screen is what this change adds.

## What this does

**A session has a URL.** `/chat/:agent/:session` opens that conversation, and the address bar follows whatever session is open, so a chat is copyable, reloadable and back/forward-able as itself. Deep links that already existed (a schedule's run, an experiment launch) now carry their session too.

**A Slack reply links back to its own conversation.** The attribution footer under every agent post already carried the agent id; a turn's reply now carries the session that turn ran on, so anyone reading the thread can pick the conversation up in the UI. Proactive posts (`send_channel_message`) belong to no turn and keep linking to the agent's chat.

The session reaches the footer from the turn itself: the ACP relay reports which session a prompt ran on (resumed or freshly minted), and the Slack worker keeps it on the turn ref that the `reply`/`react` tools already resolve. A coalesced ambient batch shares one session, so every message in it links to the same place.

**A non-owner following the link gets a screen, not a spinner.** Today a foreign agent leaves the chat overlay on "Loading agent…" forever, because there is no lifecycle state to report for an agent that isn't in your list. Now the access verdict comes first, with copy that names the reason and a way out.

The verdict is server-authoritative (an owner-scoped agent read) rather than "absent from my list", for two reasons: the list lags a just-created sandbox, which would flash the no-access screen at the person who just made it, and a failed list read must not read as a denial. Someone else's sandbox and a deleted one are deliberately the same answer — the read is owner-scoped, so neither reveals whose it is.

## Notes for the reviewer

- **Nothing new is exposed.** The WebSocket upgrade already verifies ownership before the relay wakes anything (`ws.owner_mismatch`, 404), so a stranger following a session link cannot wake someone's pod — the link only reaches a screen that says no.
- **Older footers still parse.** History injection recovers a post's author from its footer URL; messages posted before this change point at the agent's configuration page, and that shape is still recognised so an older post keeps being told apart from the reading agent's own.
- **The URL mirrors the store rather than owning the session.** Session state stays where the chat engine keeps it and the URL is a projection of it, hydrated back on load and on back/forward — the same shape the selected agent already uses. Making the URL the source of truth would mean reworking the chat engine, which is not this change.
- **A terminal session stays out of the URL.** Its id is a client-minted PTY, not a resumable conversation, so a reload could not re-open it. The two spawn paths now set the mode before the id so it never appears even for a frame.
- **Left alone:** the `/sandboxes/:id` page a non-owner may reach from a pre-existing footer still answers with a bare "Sandbox not found." line. Worth a follow-up, but those links drain as threads go quiet.

## Verification

`mise run check` and `mise run test` pass (the CRD-compat task fails in a git worktree for unrelated reasons — no CRDs are touched here). New unit coverage: session links render and parse (including the legacy shape), a reply's footer carries its turn's session and degrades without one, and the chat route round-trips with and without a session. Not exercised on a live cluster — the flow worth clicking through is: post in a bound Slack channel, follow the footer link as the owner (lands in the thread's conversation), then as anyone else (lands on the no-access screen).


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `da099287`

## PR #3183: feat: follow a Slack thread into its session in the UI
**Author:** pilartomas | **Branch:** feat/slack-session-link → main | **Changes:** +473 −91 (19 files)

### Summary
Gives a chat session its own URL (`/chat/:agent/:session`) so a Slack reply's attribution footer can link back to the exact conversation, and shows a foreign follower an owner-scoped "not yours" screen instead of a perpetual spinner.

### Findings
- ✅ **Looks good:** The `FOOTER_RE` change adds an optional session segment and a legacy `/sandboxes/` alternative. Both old (session-less) and new footers still parse, and both cases are covered by new tests (`agent-footer.ts:256`).

**Verdict: aligned.** Both edited architecture pages (`channels.md`, `platform-topology.md`) reflect the new footer-session-link and URL-addressing behavior in the same PR. `Last verified: 2026-08-05` matches the commit date. No ADR leaks, no volatile content, no missing pages.

### Issue Fit
Linked issue: #3087 "Keep a conversation in the DAM UI once it's continued there…" (inferred from plain reference "Part of #3087"; no closing keyword, so the issue stays open on merge).

✅ Delivers the two "Open Questions" blockers it scopes itself to — the session id in the URL, and an owner-scoped access screen for a followed link. The main "keep replies out of Slack" behavior is explicitly declared out of scope and correctly left undone. No undeclared scope creep.

✅ No findings. The `onSession` callback threads through core→infrastructure without breaking the inward dependency rule; the callback failure is swallowed with a debug log in `acp-client.ts`.

✅ No findings. Extracting `OverlayFrame` into its own file removes duplicated chrome (a DRY win); session state stays URL-projected-from-store exactly like the pre-existing `selectedAgent` pattern; `useIsAgentInaccessible` uses TanStack Query for the server-owned access verdict.

### Verdict
APPROVE — a well-scoped, well-tested, well-documented change; behavior, docs, and tests all line up and no defects surfaced.

---

PR #3198, opened 2026-08-06
fix(skills): scope the scan cache to the credentials that produced each scan

### LINKED ISSUE #3197: Scope the skill scan cache to the credentials that produced each scan
## Context

The platform caches a source's scanned skill list for five minutes so that repeated views don't re-read the repository. Entries are keyed by repository and subdirectory alone, so a single entry is shared by every user who has registered that repository.

## Problem

That sharing is correct for public sources, where the scan result is identical no matter who asks. It is wrong for sources readable only through a user's connected credentials: those results depend on the caller's access, so sharing one entry across users lets a result fetched under one user's access satisfy another user's request.

## Goal

A cached scan is reused only by requests whose access is equivalent to the one that produced it. Public-source scanning keeps sharing a single entry across users, exactly as it does today.

## Done when

- A scan performed with a user's connected credentials is served only back to that user.
- Public-source scans continue to be shared across all users.
- The skills architecture page describes the cache's scoping accurately.

PetrBulanek:
The skills subsystem caches a source's scanned skill list for five minutes so repeated views don't re-read the repository. Entries were keyed by repository and subdirectory alone, so one entry was shared by every user who registered that repository.

That sharing is correct for scans the api-server performs without credentials — the result is the same no matter who asks. It is wrong for scans that run with a user's connected credentials, where the result depends on the caller's access. Each entry now records which of the two produced it, and is reused only for requests with equivalent access. Uncredentialed scans keep sharing a single entry across all users, exactly as before.

Closes #3197

## Sub-issues

- [x] 01 — Scope cached scans by the credentials that produced them


packages/api-server/src/modules/skills/services/skills-service.ts
```diff
@@ -428,12 +435,20 @@ export function createSkillsService(deps: SkillsServiceDeps): SkillsService {
...
-        // to the pod scan, which doesn't report a directory. Preview is deferred
-        // for private sources exactly as in the non-GitHub guard above and the
-        // archive-404 path on a cold cache; never build a URL from `undefined`.
+        // `dir` is optional because the pod scan doesn't report one, but the
+        // public-archive scan always sets it — and under scan scoping only a
+        // public-archive scan reaches here. The check still narrows the type for
+        // the read below; reaching it means an owner-scoped entry answered a
+        // shared lookup, which is worth a line rather than a quiet deferral.
         if (!skill.dir) {
+          securityLog("warn", "skill.preview.unscoped_scan", {
```
dam-code-guardian, line 444: 🟢 **Suggestion:** This branch is unreachable via a scope mismatch. `sameScope` is exact-match, so an `owner`-scoped entry can never answer this `{kind:"shared"}` lookup — a shared miss re-runs `scanPublic`, which always sets `dir`. So the comment's cause ("an owner-scoped entry answered a shared lookup") cannot happen. If `!skill.dir` ever fires, the real cause is a public-archive scan that returned a skill without a directory — a scanner-contract break, not a scope leak. Consider renaming the event (e.g. `skill.preview.missing_dir`) and rewording the comment so a responder is not sent hunting for an impossible cross-user leak.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `ab53491`

## PR #3198: fix(skills): scope the scan cache to the credentials that produced each scan
**Author:** PetrBulanek | **Branch:** fix/3197-scan-cache-owner-scope → main | **Changes:** +88 −36 (4 files)

### Summary
Scopes the in-memory skill scan cache so a credentialed (owner-scoped) scan result is never served to another user, while uncredentialed (shared) scans stay shared across all users. Closes #3197.

### Findings
- 🟡 **Warning:** The new scope discrimination — the security fix itself — has no test. All 7 cache tests pass `SHARED`, so `sameScope` could `return true` and every test still passes; a future refactor could silently reopen the cross-user leak with green tests. (`packages/api-server/src/__tests__/unit/skills-scan-cache.test.ts`)
- 🟢 **Suggestion:** The `unscoped_scan` warn names a cause the code prevents (`skills-service.ts:444`).
- ✅ **Looks good:** Scope model verified leak-safe — all 3 `scanSource` call sites are scoped, `sameScope` is exact-match, one process-wide cache instance, and private repos 404 the shared path so an `owner` entry can never be served to another caller.

**Verdict: aligned.** `docs/architecture/skills.md` scan-cache bullet was updated to describe the scoping accurately; `Last verified: 2026-08-06` matches the PR date. No ADR or volatile-content leaks.

### Issue Fit
Linked issue: #3197 "Scope the skill scan cache to the credentials that produced each scan" (closing keyword).

✅ Delivers the issue's ask; no undeclared scope. All three done-when items met: credentialed scan served only to its owner (`{kind:"owner"}` + exact `sameScope`), public scans stay shared (`{kind:"shared"}`), architecture page updated.

✅ No findings. Scope selection (shared vs owner) sits in the application layer; the `ScanScope` port type lives in `infrastructure/` and is imported as a type only — consistent with the repository-interface convention. No layer-direction violation.

### Verdict
COMMENT — the fix is correct and complete, but its central security behavior ships without a regression test; the one 🟢 is a mislabeled security-log line.

---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `347c4ba`

## PR #3198: fix(skills): scope the scan cache to the credentials that produced each scan
**Author:** PetrBulanek | **Branch:** fix/3197-scan-cache-owner-scope → main | **Changes:** +225 −43 (5 files)

### Summary
Scopes the skill scan cache to the access that produced each scan (closes #3197). This revision keys entries by scope instead of checking scope over one shared slot, and adds the isolation tests missing before.

### Changes since last review
Previous HEAD: `ab53491` (2026-08-06T11:17:49Z) — verdict COMMENT

- ✅ **Fixed:** The scope discrimination now has regression tests. A new `skills-scan-scope.test.ts` pins scope selection (public → `shared`, private → `owner`), and the cache tests assert owner isolation and all-scope invalidation. (`packages/api-server/src/__tests__/unit/skills-scan-scope.test.ts`)
- ✅ **Fixed:** The revision removes the read-check-only design where a scope mismatch overwrote the other user's slot. Scope is now part of the key, so two users of one private source no longer evict each other. Verified: owner A keeps its own entry after B and shared scan, and `invalidate` drops every scope of the source without touching a prefix-neighbor URL or a different path. (`packages/api-server/src/modules/skills/infrastructure/scan-cache.ts`)
- 🔁 **Still present:** The `unscoped_scan` warn still names a cause the code prevents (`skills-service.ts:444`) — that file is untouched this revision.

### Findings

**Verdict: aligned.** The scan-cache bullet in `docs/architecture/skills.md` was rewritten to match the scope-keying, drop-expired-on-miss, and all-scope invalidation; `Last verified: 2026-08-06` matches the PR date. No ADR or volatile-content leak.

### Issue Fit
Linked issue: #3197 "Scope the skill scan cache to the credentials that produced each scan" (closing keyword).

✅ Delivers the issue's ask; no undeclared scope. All three done-when items met — credentialed scan served only to its owner, public scans stay shared, architecture page accurate.

✅ No findings. The cache rewrite stays inside `infrastructure/`; the `ScanCache` port and `ScanScope` type placement are unchanged, imported as a type only. No layer-direction violation introduced.

### Verdict
APPROVE — the previously-missing regression tests now cover scope isolation and invalidation, the revision also fixes a mutual-eviction bug, and the scope-keyed cache is verified leak-safe.

---

PR #3206, opened 2026-08-06
feat(connections): narrow a GitHub App connection to chosen repos and permissions

### LINKED ISSUE #3201: Limit a GitHub App connection to specific repositories and permissions
## Context

Connecting a GitHub App today asks for the app's identity and its private key, and nothing else. From that we mint a short-lived access token and use it for the agent's GitHub traffic. That token carries everything the installation carries — every permission the app was granted, across every repository the organization installed it on.

In practice organizations install one GitHub App broadly. Anyone who wants an agent to do one narrow thing has no way to ask for less than the whole installation.

## Problem

There is no way to say what a connection should be allowed to do. Someone who wants a bot that reads a single documentation repository ends up with a credential that can act on every repository the app is installed on, with every permission the app holds — usually including write access.

The safe choice simply isn't on offer:

- **Nothing to narrow.** The connection form has no field for repositories or permissions. The only lever is to change the app installation itself, which affects everyone else using it.
- **A second app per task.** The workaround is registering and installing a separate GitHub App for every narrow job, each needing an org admin's approval. That's slow enough that people reuse the broad one instead.
- **Over-permission by default.** Every agent granted the connection inherits the full installation, so a mistake or a prompt injection has the installation's blast radius rather than the task's.

This bites hardest in the unattended cases — bots and scheduled agents — which is precisely what this kind of connection is for.

## Goal

When someone connects a GitHub App, they can narrow what that connection is allowed to do — to a chosen set of repositories and a chosen set of permissions — without touching the app installation itself. The narrowing holds for the life of the connection, not just at the moment it's created.

## User Stories

- As someone connecting a GitHub App, I want to pick which repositories the connection can reach so that an agent using it can't touch anything else.
- As someone connecting a GitHub App, I want to pick which permissions it gets so that a read-only job can't write.
- As someone connecting a GitHub App, I want to see what the installation actually allows so that I choose from what's available instead of guessing.
- As a security-conscious admin, I want one broadly-installed app to back many narrow connections so that nobody has to register a new app per task.
- As a user, I want a narrowed connection to keep working over time so that the limits still hold after its credential renews.

## Scope

In scope: choosing repositories and permissions when creating a GitHub App connection, and keeping that choice in force whenever the credential renews or the key is replaced.

Out of scope: changing what the app installation itself is granted — that stays an organization-level decision made on GitHub.

## Proposed solution

GitHub already supports this at the point where we ask for the token: the request can name the repositories and permissions it wants, and GitHub refuses anything beyond what the installation already has. The narrowing is a matter of asking for less, and we are already the party making the request.

Two consequences worth stating up front:

- **The choice must be stored with the connection**, not applied once at connect time. The credential renews on a schedule, so a choice that isn't stored would quietly widen back to the full installation within the hour.
- **A stale choice needs to surface.** If the organization later removes a repository from the installation or drops a permission, renewal starts failing. That should read as a connection needing attention, not as silent retrying.

For the picker, we can read the installation's own repositories and permissions and offer those, so people choose from what actually exists rather than typing names. Whether that lands in the first cut or follows a plain list is open.

Leaving the default alone matters: an existing connection with no choice recorded should behave exactly as it does today.

## Open Questions

- Do we identify repositories by name or by their GitHub id? Names are what people recognise, but a rename would break a stored choice at renewal time — an hour later, not while the user is looking at the form.
- Should permissions be a flat list, or a few presets ("read code", "read and write issues") with the full list behind an advanced toggle?
- Should narrowing be per agent rather than per connection? One narrow connection per job is simpler but multiplies connections.
- Is read-only access to a repository's files enough for the common "let an agent read our docs" case, or do people also expect the agent to reach the rendered site? Worth knowing before we design the picker (see the note below).

## A note on the motivating case

The case that prompted this was wanting an agent to read one team's documentation and nothing else. Worth recording what we learned, because it shapes the design:

The documentation is published as a rendered site from an enterprise GitHub host, and that site turns out to sit behind the company's single sign-on rather than behind GitHub's own authentication. A GitHub App credential — however narrowly scoped — cannot open it. It looks unauthenticated only because a person's browser already carries an SSO session.

So the workable path for that case is the agent reading the documentation from the **repository that publishes it**, through GitHub's API, rather than fetching the rendered pages. That is exactly the shape this issue asks for: a credential limited to one repository with read-only access to its files. It also means "reach the published site" is a different problem, involving a different identity provider, and is not in scope here.

## Additional resources

- GitHub's documentation for requesting a narrowed installation token: https://docs.github.com/en/rest/apps/apps#create-an-installation-access-token-for-an-app
- Related: #2908 (Epic - Network access control) covers what a sandbox can reach on the network. This issue covers what a credential is allowed to do once it gets there.

pilartomas:
Closes #3201.

## Why

A GitHub App installation is an organization-wide grant. The platform minted its installation token with all of it — every permission the app holds, across every repository the org installed it on — because the token request carried no body, and an absent body is what asks GitHub for the installation's full authority.

So an agent that needed to read one repository got authority over all of them, usually including write. The only way to narrow it was to change the installation itself, which affects everyone else using that app, or to register a separate app per task.

## What changed

A GitHub App connection can now name repositories and/or permissions to mint against. GitHub is the arbiter — it refuses any request exceeding the installation — so the narrowing can only ever reduce authority, never claim it.

**The subset is stored on the connection, not applied once at create.** This is the load-bearing part. Tokens are re-minted every hour, and there are three mint sites (create, private-key rotation, the refresh loop). All three go through one shared helper that reads the subset from stored auth, so a renewal cannot silently widen a connection back to the whole installation. There's a test asserting the re-mint asks for the same subset, and another asserting the stored scope round-trips the wire schema — that one matters because the refresh loop silently drops rows its schema rejects, so a wrong schema would have parked scoped connections instead of refreshing them.

**Narrowing is opt-in and unscoped behavior is byte-identical.** An empty list is not the same request as no list, so the engine sends no body at all unless something is actually narrowed, and blank input is never stored as an empty value. Every connection created before this change is unaffected.

**A stale subset surfaces instead of spinning.** When the org drops a repository from the installation or revokes a permission, GitHub answers 422. Retrying re-sends the same losing request, so a 422 on a *scoped* request is classified permanent — the connection parks and reads expired, waiting for someone to widen the installation or narrow the connection. A 422 on an unscoped request isn't about the scope and stays retryable.

**Bad input fails at create, not an hour later.** Parsing rejects `owner/repo` and pasted repo URLs with the name to use instead, and malformed permission pairs with the expected shape — because the alternative is a refresh-time 422 long after the user has left the form.

## Notes for the reviewer

- Both fields are single strings (space- or comma-separated), matching how client-credentials `scopes` already keeps the schema-driven forms all-string. The server parses them.
- The CLI **refuses to run** if you pass `--repositories`/`--permissions` to a server too old to declare them. Normally an older server stripping an unknown input is harmless, but here the strip lands in the unsafe direction — you'd get a full-authority token and a success message. This follows the existing `--ambient` precedent in `channel slack-connect`, which fails loudly even though its strip is safe.
- Repositories are stored by **name**, not id. Names are what people recognise and type; the tradeoff is that renaming a repository turns into a refresh-time rejection rather than a create-time error. Issue #3201 flags this as an open question — worth settling here if reviewers disagree.
- Docs went to `security-and-credentials.md` rather than `connections.md`. That page owns credential authority and trust boundaries, and `connections.md` is at 39,991 of its 40,000-character cap — the guidelines say going over is a signal to re-think what a page carries, not to reword to fit, so I didn't force it in. Nothing on `connections.md` is falsified by this change; it describes the required inputs, and both new ones are optional.

## Not in scope

Reading a narrowed connection's subset back through the API, or editing it after create — the credential update path replaces the secret, and the rest of a connection's config is immutable by existing design. Changing that is a broader change than this issue.

## Verification

- `api-server` 831 tests, `cli` 154, `ui` 243 — all pass. Typecheck, lint, and format clean across all four touched packages.
- Reviewed adversarially before commit: 14 candidate findings, 12 refuted against the code, 2 confirmed and fixed (the repo-name suggestion pointed at values that fail again — including an empty one, which would have meant "no narrowing"; and the CLI version-skew fail-open above).
- `mise run check` fails on the CRD-compatibility step in a git worktree — a known tooling limitation where crdify can't resolve refs through a worktree gitdir. No CRDs are touched here; per-package checks were run directly.

### DIFF
--- a/docs/architecture/security-and-credentials.md
+++ b/docs/architecture/security-and-credentials.md
@@ -1,6 +1,6 @@
 # Security and credentials
 
-Last verified: 2026-08-05
+Last verified: 2026-08-06
 
 ## Overview
 
@@ -268,6 +268,25 @@ Each connected service produces one K8s Secret per `(owner, connection)`:
   host; only the installation token reaches the gateway's injection path. Like
   the client secret above, a rotated private key is pasted in place and proven by
   minting before it is stored.
+
+  A Connection may additionally **narrow the authority of the token it mints**,
+  below what the app installation itself holds — to a chosen set of repositories,
+  to a chosen set of permissions, or both. An installation is an
+  organization-wide grant, typically far broader than any one agent's task, and
+  narrowing is how one broadly-installed app backs many least-privilege
+  Connections without a second app per task. GitHub is the arbiter: it refuses
+  any request exceeding the installation, so the narrowing can only ever reduce
+  authority, never claim it. The chosen subset is **part of the credential's
+  stored identity, not a one-time argument** — every renewal and every key
+  rotation re-mints against the same subset, so a Connection cannot silently
+  widen back to the whole installation between renewals. Narrowing is opt-in:
+  a Connection that names no subset carries the installation's full authority,
+  which is what every Connection made before the capability existed does. Once
+  a subset stops being covered — the organization drops a repository from the
+  installation, or revokes a permission — renewal is *rejected* rather than
+  merely failing, so the Connection reads expired and waits for someone to
+  widen the installation or narrow the Connection, instead of retrying a
+  request that cannot succeed.
 
 **Multi-host connections.** A single OAuth connection can inject the
 same token on more than one host with **different auth schemes per
--- a/packages/api-server-api/src/modules/connections/schemas.ts
+++ b/packages/api-server-api/src/modules/connections/schemas.ts
@@ -128,6 +128,12 @@ const githubAppCreateInput = z.object({
   appId: z.string().min(1),
   installationId: z.string().min(1),
   privateKey: z.string().min(1),
+  // Optional narrowing of the minted token. Space- or comma-separated; the
+  // server parses both. Single strings keep the schema-driven forms all-string,
+  // as client-credentials `scopes` does. Blank leaves the token with the
+  // installation's full authority.
+  repositories: z.string().optional(),
+  permissions: z.string().optional(),
 });
 
 const noneCreateInput = z.object({
--- a/packages/api-server-api/src/modules/connections/types.ts
+++ b/packages/api-server-api/src/modules/connections/types.ts
@@ -63,6 +63,13 @@ export const githubAppAuth = z.object({
   connectedAt: z.number().int().optional(),
   refreshFailedAt: z.number().int().optional(),
   host: z.string().min(1).optional(),
+  // Optional narrowing of the minted token, stored because every re-mint must
+  // ask for the same subset — a scope kept only on the create input would widen
+  // back to the whole installation at the first renewal. Absent means the
+  // installation's full authority, which is what every connection made before
+  // this field existed carries.
+  repositories: z.array(z.string().min(1)).nonempty().optional(),
+  permissions: z.record(z.string(), z.string()).optional(),
 });
 
 export const headerAuth = z.object({
--- a/packages/api-server/src/__tests__/unit/connections-github-app-template.test.ts
+++ b/packages/api-server/src/__tests__/unit/connections-github-app-template.test.ts
@@ -3,6 +3,7 @@ import { describe, it, expect } from "vitest";
 import type { Contribution, SecretRef } from "api-server-api";
 import { buildConnection } from "../../modules/connections/domain/build-connection.js";
 import { buildCatalog } from "../../modules/connections/domain/catalog.js";
+import { templateToView } from "../../modules/connections/domain/connection-template.js";
 import { connectionSecretAnnotations } from "../../modules/connections/domain/connection-sds.js";
 
 const { privateKey: PRIVATE_KEY_PEM } = crypto.generateKeyPairSync("rsa", {
@@ -26,6 +27,8 @@ function build(
     appId: string;
     installationId: string;
     privateKey: string;
+    repositories: string;
+    permissions: string;
   }> = {},
 ) {
   return buildConnection(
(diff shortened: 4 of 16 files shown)


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `35ca860`

## PR #3206: feat(connections): narrow a GitHub App connection to chosen repos and permissions
**Author:** pilartomas | **Branch:** feat/github-app-token-scope → main | **Changes:** +654 −10 (16 files)

### Summary
Adds optional repository/permission narrowing to GitHub App connections. The chosen subset is stored on the connection and re-sent on every token mint, so a renewal cannot widen back to the full installation.

### Findings
- ✅ **Looks good:** All three mint sites (create, key rotation, refresh loop) route through the one `mintGitHubAppToken` helper that reads the stored scope; the raw `mintInstallationToken` has no other caller, so a re-mint cannot silently widen the token.

**Verdict:** aligned.

### Issue Fit
Linked issue: #3201 "Limit a GitHub App connection to specific repositories and permissions" (closing keyword)

✅ Delivers the in-scope ask — repository and permission narrowing, stored on the connection and re-sent on every renewal, stale scope surfaced as expired (422 → parked), unchanged default for existing connections. The installation-aware picker (user story 3) is not built, but the issue leaves that open; the name-vs-id choice is declared in the PR body. No undeclared scope creep.

✅ No findings. Layer placement is correct: scope parsing in `domain/`, the HTTP mint in `infrastructure/`, stored-auth mapping in `services/`, Zod input validation in the contract package. No `any`/`as` in production code.

### Verdict
APPROVE — delivers the issue in full, with tests whose coverage survives neutering the new scope discriminators; no correctness or security defect found.

---

PR #3220, opened 2026-08-07
feat(controller): admit a blocked start by hibernating the owner's idle agents

### LINKED ISSUE #3179: Critical findings from review of closed PR #3177
Review of PR #3177 completed after the PR merged, so this is filed as an issue instead of a review (per the closed-PR path). Only 🔴 Critical findings are reported here.

### 🔴 Critical: the new `NO_PROXY` in `/etc/platform/env` silently overrides the VM backend's deliberately-wider `NO_PROXY` for `docker` and `k3s`

`packages/controller/pkg/reconciler/resources.go:135-136` adds `NO_PROXY`/`no_proxy` = `localhost,127.0.0.1,::1` to `agentPlatformEnv`, which the VM backend renders verbatim into `/etc/platform/env` (`vm_resources.go:104-105`).

`packages/agents/claude-code-vm/system/proxy-dropin.conf` — installed as **both** `docker.service.d/platform.conf` and `k3s.service.d/platform.conf` — already sets a deliberately wider list, and states why in its own comment ("`NO_PROXY` keeps cluster-internal traffic (k3s pod/service CIDRs, localhost) off the proxy"):

```
EnvironmentFile=-/etc/platform/env
Environment=NO_PROXY=localhost,127.0.0.1,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,.svc,.cluster.local
```

`EnvironmentFile=` wins over `Environment=` regardless of the order the two directives appear in — this is documented and implemented, not incidental:

- `systemd.exec(5)`, `EnvironmentFile=`: *"Settings from these files override settings made with `Environment=`."*
- `src/core/exec-invoke.c`: `strv_env_merge(params->environment, our_env, joined_exec_search_path, pass_env, context->environment, params->files_env)` — `files_env` is merged **last**, and `env_append` (`src/basic/env-util.c:145-176`) has each later set override an existing matching key.

So after this change both units receive the loopback-only list, and the cluster-range exclusions are dropped. The in-VM k3s deliberately runs on `cluster-cidr: 10.144.0.0/16` / `service-cidr: 10.145.0.0/16` (`system/k3s-config.yaml`) — both inside the `10.0.0.0/8` entry that just stopped applying — so k3s and dockerd traffic aimed at the inner cluster's own pod/service addresses is now routed to the egress gateway and 503s, which is the same failure mode this PR exists to remove, moved one layer down. The guest advertises that cluster as a feature (`Containerfile`: "a preinstalled `k3s` Kubernetes cluster (`kubectl` works out of the box)").

Note that the two `Environment=NO_PROXY=` lines in the drop-in are now dead config, which is the tell that the collision is unintended rather than a deliberate narrowing: the PR body's rationale ("deliberately loopback-only … the harness API and all external hosts must keep crossing the paired gateway") is about the *agent's* egress, not about dockerd's and k3s's cluster-internal dialing.

`vm_test.go:151-152` cannot catch this — it asserts on the rendered userdata only, where the value is correct; the regression appears only once systemd merges that file over the drop-in.

Worth noting the container backend is unaffected (no such drop-in), and `packages/agents/claude-code/model-gateway.sh:24-25` shows the additive shape (`${NO_PROXY:+,$NO_PROXY}`) that avoids exactly this class of clobber.

---

### LINKED ISSUE #3184: feat: proactively hibernate agents when new one is starting
When we need to make room to start a new agent, we should be able to prematurely hibernate an idle agent even if its idle timeout did not fully lapse yet.

JanPokorny:
Closes #3184.

When the per-user budget gate refuses a start, the room is usually held by an agent its owner has already finished with, sitting behind an idle timeout that has not lapsed — an hour on the chart default. Rather than refuse and tell the user to go stop something, the gate now hibernates that owner's own **unattended idle** agents early, longest-idle first, and re-runs the same arithmetic.

This narrows a standing invariant ("the Budget constrains starting, never running"), so it carries an ADR: **ADR-081**. Admission pressure from the *same owner* becomes the single exception — a ceiling change is still never grounds to evict, and no owner's demand ever touches another owner's agents.

## Four rules bound it

- **Unattended only.** Session pins (chat/terminal/SSH), Experiment pins, sweepable Invocation targets (their driver is blocked on the result), and never-hibernate agents are never candidates. Survivors are probed exactly as an ordinary hibernation probes. The pins carry real weight here: the runtime's own idle flag reads an attached-but-turnless chat as *idle*, so probing alone would reclaim a pod somebody is watching.
- **A 3-minute idle floor.** Besides sparing an agent its user may be moments from returning to, this is what makes reclaim **non-recursive**: an agent admitted this way carries fresh activity, so it cannot be the next start's victim and A-evicts-B-evicts-A cannot close.
- **Provably sufficient, or nothing.** Candidates must cover the shortfall in both dimensions before any of them is touched, so no agent is hibernated for a start that was going to be refused anyway.
- **A reclaimed agent stays down.** It is marked as having *spent* the activity stamp it was reclaimed under, so only a **newer** bump revives it, back through the gate. Without that, the victim's own next reconcile — still inside its timeout — would take the room straight back. The mark self-clears when a deliberate touch outdates it.

Applies to every start path, including schedule fires and channel-driven wakes, which is where a manual "free some room" resolution is worst: nobody is present to perform it.

## The verdict stays synchronous

`reservedByOwner` counts **desired** replicas, so scaling a victim's pair to zero frees budget the instant the write lands and the same reconcile admits. There is no reclaim-in-progress state and the `OverBudget` fail-fast needs no third outcome — which is why **the api-server and the CRD are untouched**. Pods draining afterwards are the scheduler's business, which Budgets have never modelled.

## What the user sees

Nothing new. The UI polls the CR every 5s; because the gate admits within one reconcile, `Ready=False/OverBudget` is never stamped, so a reclaiming wake is visually identical to an ordinary one (just a beat slower). A refusal that reclaim could not rescue behaves exactly as before. The one visible side effect is that another of their sandboxes flips to `Hibernated` with no explanation — the deliberate "silent" choice recorded in the ADR.

## Verification

- `controller:test`, `controller:check` (vet, staticcheck, fmt, build, CRD gates), `docs:check` — all green.
- 6 new unit tests, including a 10-case eligibility matrix.
- **Manual run on a real k3s cluster** (2 CPU ceiling; 1.5 CPU victim idle 10 min with its 1h timeout unlapsed; 1 CPU claimant):

| Check | Result |
|---|---|
| Blocked claimant triggers reclaim | `reclaimed-at` stamped |
| Victim state | `Ready reason: Hibernated`, replicas 0 |
| Claimant actually reaches Ready on a real scheduler | **PASS** — `1/1 Running` |
| Victim stays down despite activity inside its timeout | replicas still 0 after 30s |
| Newer bump revives it, back through the gate | `Ready reason: OverBudget` |

Controller log: `reclaimed idle agent to admit a blocked start agent=reclaim-victim for=reclaim-claimant idleFor=626s`

## Also included

`20e3c5ec` is an unrelated drive-by found while testing: `cluster:install` keyed "already provisioned" on `/var/lib/rancher/k3s` existing, but the provision scripts create that directory early — so an install interrupted mid-provision wedged **every** later run with a misleading `stat /etc/rancher/k3s/k3s.yaml: no such file`. Now keyed on the kubeconfig, which k3s writes only once it is actually serving. Happy to split this out if you'd rather review it separately.

## Reviewer notes

- Only the safe path of that install fix is verified (kubeconfig present ⇒ provisioning skips, as today). Proving the repair path means deliberately wedging a cluster mid-provision.
- Reclaim runs inside the single reconcile worker that drains all agent reconciles, so an owner with many idle candidates adds probe latency (3s timeout each, scan stops once the shortfall is covered) to that shared queue. Recorded as a "Harder" consequence in the ADR; it is the thing I would watch first under load.
- The over-budget UI copy ("Pause or stop a running sandbox to free room") is now second-line advice, since the platform already tried. Still actionable — a hard stop bypasses the pins that blocked reclaim — but it does not mention that simply retrying after the 3-minute floor may also work.

---

## Bundled commit: `1e6400f7` (declared after review)

`1e6400f7 fix(agents): stop the platform NO_PROXY from clobbering docker/k3s's wider list` (Fixes #3179) rides along in this branch. It is **unrelated to #3184** and was the pre-existing tip of my worktree when I branched — it is not on `origin/main`, so it shows up in this PR's commit list. Author is the same, but it was not written as part of this work and Code Guardian correctly flagged it as undeclared scope creep.

It concerns systemd `EnvironmentFile=` precedence shadowing `proxy-dropin.conf`'s `Environment=NO_PROXY=` on `docker.service` / `k3s.service`, dropping cluster-CIDR exclusions.

I have left it in place rather than rebase it away, because dropping someone's commit — even my own, from other work — is a maintainer's call, not mine. **If you would rather this PR carry only the #3184 work, say so and I will rebase my four commits onto current `origin/main` and force-push.** That would also bring the branch up to date, since `main` has advanced since I branched.


packages/controller/pkg/reconciler/reclaim.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil || len(chosen) == 0 {
+		return false, err
+	}
+
+	// Stamp before scaling: an unstamped victim would be scaled straight back
+	// up by its own next reconcile, since its activity is still inside its
+	// timeout — that is the whole premise of reclaiming it.
+	now := time.Now().UTC().Format(time.RFC3339)
+	for _, c := range chosen {
+		if err := r.stampReclaimed(ctx, c.name, now); err != nil {
```
dam-code-guardian, line 76: 🟡 **Warning:** No test exercises this write path end-to-end. `stampReclaimed` is what stops a reclaimed victim's own next reconcile from scaling itself back up and stealing the room straight back — the central non-recursion guarantee. But `TestReclaimedPeerStaysDown...` builds the `reclaimed-at` annotation by hand and only checks `shouldRun`; no test reconciles a victim after a real reclaim, so making this call a no-op fails no test. Add a case that asserts the annotation is written and the victim stays at replicas 0 on its next reconcile.

deploy/tasks.toml
```diff
@@ -136,7 +136,14 @@ fi
...
   KUBECONFIG="/etc/rancher/k3s/k3s.yaml"
-  if [ ! -d /var/lib/rancher/k3s ]; then
+  # Keyed on the kubeconfig, which k3s writes only once it is actually
+  # serving — not on /var/lib/rancher/k3s, which the provision scripts create
+  # early. An install interrupted mid-provision leaves that directory behind,
+  # and skipping on it wedged every later run: no k3s, no service, and a
+  # 120s wait ending in a misleading "stat k3s.yaml: no such file". The
+  # provision scripts are idempotent, so re-running over a partial state heals
+  # it.
+  if [ ! -f "$KUBECONFIG" ]; then
```
dam-code-guardian, line 146: 🟢 **Suggestion:** The safe path (kubeconfig present ⇒ skip) is verified, but the repair path this change exists for — re-provisioning over a partial `/var/lib/rancher/k3s` — rests on the provision scripts being idempotent over that partial state. That is the one path left unverified (noted in the PR). A test or a one-line manual check that a mid-provision-wedged host heals on re-run would close the actual bug surface.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `20e3c5e`

## PR #3220: feat(controller): admit a blocked start by hibernating the owner's idle agents
**Author:** JanPokorny | **Branch:** feat/admit-by-hibernation#locki-x6w0glho → main | **Changes:** +545 −11 (13 files)

### Summary
The per-user budget gate now admits a blocked start by hibernating that owner's own unattended idle agents ahead of their timeout, longest-idle first, and only when the freed room provably covers the shortfall. Two unrelated drive-by fixes ride along (k3s install keying, guest NO_PROXY).

### Findings
- 🟡 **Warning:** No test exercises the `stampReclaimed` write path; the reclaimed-stays-down guarantee is proven only by `shouldRun` on a hand-built annotation (`packages/controller/pkg/reconciler/reclaim.go:76`).
- 🟡 **Warning:** Undeclared scope creep — the guest NO_PROXY change (commit `1e6400f`: `proxy-dropin.conf`, new `vm-no-proxy.env`, `Containerfile`) is bundled but not mentioned in the PR body. See Issue Fit.
- 🟢 **Suggestion:** The k3s install repair path rests on provision-script idempotency, which the author notes is unverified (`deploy/tasks.toml:146`).
- ✅ **Looks good:** The "synchronous reclaim" claim holds. `reservedByOwner` and `agentDesiredUp` read desired replicas live via `r.client`, `scaleAgentPairToZero` writes them via the same client, and the freed-room math reuses the same `limitsOf`, so the post-reclaim re-check sees exactly the room freed.

`aligned` — no drift. `budgets.md` and `agent-lifecycle.md` both gained the reclaim behavior and bumped `Last verified` to 2026-08-06 in the same PR. No ADR reference leak, no volatile content, no new-subsystem page owed.

### Issue Fit
Linked issue: #3184 "feat: proactively hibernate agents when new one is starting" (closing keyword)

- ✅ Delivers the ask: the budget gate hibernates the owner's unattended idle agents ahead of their timeout to admit a blocked start.
- 🟡 Undeclared scope creep: commit `1e6400f` changes guest NO_PROXY handling (`proxy-dropin.conf`, new `vm-no-proxy.env`, `Containerfile`) — unrelated to #3184 and not mentioned in the PR body. The k3s install fix is declared ("Also included"); only the proxy change is undeclared.

### Verdict
COMMENT — The core reclaim logic is correct and its mechanism claims verify. The two warnings (a coverage gap on the anti-room-theft stamp, and an undeclared bundled change) are advisory, not blocking.

---

JanPokorny:
Thanks — all three findings addressed or answered.

**🟡 `stampReclaimed` write path untested — fixed** in `b7eba329`.

You were right that no-oping the call failed no test: the guarantee was only proven by calling `shouldRun` on a hand-built annotation. The new `TestReclaimedVictimStaysDownOnItsOwnNextReconcile` reconciles the victim after a *real* reclaim and asserts the stamp landed on the CR, that the victim stays at replicas 0 on its own next reconcile, and that the claimant keeps the room it was admitted with.

I confirmed the test actually bites by mutation-testing it — stubbing `stampReclaimed` to `return nil` fails exactly this test and no other:

```
--- FAIL: TestReclaimedVictimStaysDownOnItsOwnNextReconcile
    Error: Should NOT be empty, but was
    Messages: reclaim must stamp the victim, else its next reconcile takes the room back
```

**🟡 Undeclared `NO_PROXY` change — declared**, PR body updated with a new section.

Correct catch. `1e6400f7` (Fixes #3179) was the pre-existing tip of my worktree when I branched and is not on `origin/main`, so it rides along. Unrelated to #3184. I have not rebased it away, because dropping a commit from someone's other work is a maintainer's call — but say the word and I will rebase onto current `origin/main` and force-push, which also brings this branch up to date.

**🟢 k3s install repair path — still unverified, and I owe you a caveat.**

I attempted to verify it and damaged my dev cluster doing so: `pkill`-ing `k3s server` brought k3s back up registering a *new* node, orphaning the original and every pod on it. Repaired by uninstall + reinstall. So the finding stands exactly as written — the repair path (re-provisioning over a partial `/var/lib/rancher/k3s`) rests on provision-script idempotency and is **not** verified. Only the safe path is: kubeconfig present ⇒ provisioning correctly skips, unchanged from today.

Doing it properly needs a throwaway host rather than a live cluster. Given that, I would rather not claim it verified. If a maintainer prefers, I am happy to split the `deploy/tasks.toml` change out of this PR entirely so it can be reviewed and tested on its own — it is a drive-by, not part of #3184.

PR #3221, opened 2026-08-07
fix(ui): don't lose a new session's first message on a fast session switch

### LINKED ISSUE #2963: New session prompt is lost when switching to another session before it's sent
**Description:** When a user creates a new session and enters a prompt, then clicks on another session before the prompt is sent, the prompt is never delivered to the server and is lost permanently.

**Steps to reproduce:**
1. Create a new session
2. Type a prompt into it
3. Click on another (existing) session before the prompt is submitted

**Actual behavior:** The prompt is never sent to the server and is lost forever.

**Expected behavior:** The prompt should be accepted and the session created immediately, so the original prompt is never lost.

jjeliga:
Typing the first message in a new session and immediately clicking another session lost the message — and delivered it to the *clicked* session, where the agent answered it with that conversation's context. The prompt was also appended to that conversation's log, so it came back on every future load.

## Cause

A send resolved which session it was for only *after* awaiting the transport. A first prompt has no session yet, so it borrowed the chat's connection; clicking away repointed that connection at the clicked session, `session/new` never ran, and the prompt went to the wrong conversation. Every failure after that point was written to an optimistic bubble the switch had already deleted, which is why it was silent.

## Fix

- A send binds to the session it was dispatched for and refuses to prompt anything else.
- A first prompt gets a connection of its own, creates its session there, and then either becomes the chat's live connection or carries its turn to completion after the user has moved on. Either way the session lands in the sidebar with the prompt delivered.
- ACP updates now carry their session id and are filtered against the session on screen, so a channel bound elsewhere cannot paint the current chat.
- Send failures are classified: a connection dropping *after* delivery is not a failure (the runtime keeps the turn running), one before it is — and neither is reported in the SDK's wording.

## Also in here

From a review pass over the first two commits:

- StrictMode left a mount ref `false` for the component's life, so in dev *every* first prompt detached instead of adopting. Same bug fixed in the reconnect loop.
- Two quick sends to a blank chat created two sessions, the second invisible; they now share one.
- A failed attachment upload leaked the private socket and orphaned its session.
- Switching agents under a still-mounted chat could put one agent's session in another's.
- `engage`'s `session/new` branch was unreachable except as a misfire that repointed the view at a phantom session — removed, so session creation exists in one place.
- api-server: a client leaving during pod wake stranded its upstream, pinning that session against the runtime's idle reap.
- After a refresh, a queued follow-up rendered as both prompts above both replies (pre-existing, in the replay projection).

## Testing

Manually verified against the local cluster: new session → type → immediate switch (warm and cold harness), send-while-busy queueing, switching mid-turn, leaving to the sandbox list mid-flight, and a refresh to check replay ordering. `ui:check`, `ui:test`, and the api-server + agent-runtime checks pass. The relay guard is covered by typecheck and lint only — it needs an api-server rebuild to exercise.

Docs: `platform-topology.md` claimed the UI opens one ACP socket per session; `agent-lifecycle.md` now states the detach semantics the delivery reporting depends on.

Closes #2963


dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `78a2912`

## PR #3221: fix(ui): don't lose a new session's first message on a fast session switch
**Author:** jjeliga | **Branch:** fix/2963-preserve-first-message-on-fast-session-switch → main | **Changes:** +639 −234 (17 files)

### Summary
Reworks the UI ACP connection/prompt layer so a first prompt takes its own private connection, creates its session there, and binds to the session it was dispatched for. A fast switch can no longer lose the message or deliver it to the clicked conversation (#2963). The PR also ships related concurrency fixes and the matching architecture-doc updates.

### Findings
- 🟡 **Warning:** New pure decision logic ships without unit tests, against the repo's own `src/__tests__/unit/` convention — `resolvePromptTarget` (`packages/ui/src/modules/sessions/lib/prompt-target.ts`), `classifySendOutcome` (`packages/ui/src/modules/sessions/lib/send-outcome.ts`), and the new parked-bubble branch in `appendQueuedUser` (`packages/ui/src/modules/acp/session-projection.ts:429`). These are the decision points the data-loss fix depends on, and their siblings are already unit-tested (`send-error.test.ts`, `session-projection.test.ts`). The PR body states they were verified manually against a live cluster only.
- ✅ **Looks good:** The `acp-runtime.ts` comment rewrite matches the real fan-out — `appendUserPromptToLog` already calls `appendAndFanOut(…, { skipChannel: originator })`, so "the sender is skipped on fan-out … but the line is logged" is accurate, not a comment/code drift.

**Verdict: aligned.** Both edited architecture pages reflect the code and carry `Last verified: 2026-08-07`. `platform-topology.md` now describes several ACP channels per Agent; `agent-lifecycle.md` adds the "a prompt outlives the channel that sent it" detach semantics the delivery reporting relies on. No ADR leak, no volatile content, no undocumented new subsystem.

### Issue Fit
Linked issue: #2963 "New session prompt is lost when switching to another session before it's sent" (closing keyword `Closes #2963`).
✅ Delivers the ask — a first prompt now gets its own connection, creates its session immediately, and lands in the sidebar even after the user navigates away. The "Also in here" extras are all declared in the PR body; no undeclared scope creep.

✅ No findings. New boundaries use discriminated unions (`PromptTarget`, `SendOutcome`), pure logic is extracted into `lib/`, no `any` or unsafe casts introduced, and imports stay within module boundaries.

### Verdict
COMMENT — the fix is correct and well-reasoned; add unit tests for the new pure decision functions and the replay-ordering branch before merge.

---

jjeliga:
@dam-code-guardian You are making up a unit tests convention for our repo which does not exist, you've inferred it. If you go through our docs and .md files you will not find any rule regarding tests except for "do not write unit tests", so not including unit tests in this PR is actually fully in line with the repo's convention.

dam-code-guardian:
@jjeliga You are right, and thank you for the correction. Your PR needs no new unit tests. The convention is that implementation agents do not author tests by default. Verification uses the manual smoke test plus the existing suite. My review wrongly treated the missing tests as a finding. I recorded this correction. From now, I flag missing tests only when the smoke test and existing suite cannot verify the change.

jjeliga:
@dam-code-guardian Why didn't you approve then?

packages/ui/src/modules/sessions/hooks/use-acp-connection.ts
```diff
@@ -127,89 +169,213 @@ export function useAcpConnection(
...
+          if (startInFlightRef.current?.holders === holders) {
+            startInFlightRef.current = null;
+          }
+          if (!keep) {
+            listening = false;
+            return;
           }
+          keepAsLive(connection, ws, startedSessionId);
+        },
+        finish: () => {
```
PetrBulanek, line 285: `startInFlightRef` is only cleared inside `settle`, or when `startSession` itself rejects. A send that gets past `beginSession()` and then throws *before* it reaches `settle` — `buildPromptBlocks` failing on an attachment upload — leaves the entry in place while `finish()` closes its socket. The next send in that chat joins the stale entry and receives a `StartedSession` whose WebSocket is already closed.

Reproduced against a local cluster. The upload failure was simulated by rejecting only the `files.upload` request; the network was healthy again for every step after it:

1. Attach a file to the first message of a new chat, with the upload failing → `Upload failed for "probe.txt": …`. The chat stays blank, no session adopted.
2. Send a plain text message in the same chat → `Couldn't send — the connection to the agent dropped before the message was delivered.` The socket was already dead; nothing was actually wrong with the connection.
3. That second send still calls `settle(true)`, so `keepAsLive` installs the closed socket as the chat's live connection and commits the orphaned session id to the store. Because the socket closed before the handler was attached, the close event never fires and engagement is never cleared, so the reconnect path can't recover it.
4. Every message after that fails with `Session not found` — the session was created by `newSession` but never prompted, so it isn't in `session/list`. The chat only recovers if the user abandons it via New session.

Net effect of one failed upload: two messages lost, a ghost row in the sidebar, and a chat that keeps failing with no indication that leaving it is the way out.

PetrBulanek (changes requested):

PetrBulanek (approved):

PR #3223, opened 2026-08-07
fix(api-server): (attempt to) make the api-server safe to run at more than one replica

JanPokorny:
`apiServer.replicas` defaults to `1`, and several things break if you raise it. This audits the api-server for single-instance assumptions and fixes the ones that are real.

Some subsystems were already hardened and are unchanged: periodic jobs and schedules on BullMQ, session presence on Redis, OAuth/bind handoff in Redis TTL stores, per-owner resize on a Postgres advisory lock.

## What was broken

**Channel workers (Slack + Telegram).** Both transports admit one consumer per install — Telegram's `getUpdates` refuses a second, Slack Socket Mode hands each event to just one of the app's open connections — while every piece of per-turn state in the workers (in-flight/lingering turn refs, ambient queues, the per-thread turn lock) lives in the process. At more than one replica that splits a conversation across processes that cannot see each other's state, reintroducing the #2952 cross-thread misroute across the replica boundary. The Telegram worker said so itself in a comment.

Fixed with a Redis **leader lease**: one replica runs the workers, the rest run none. Outbound `reply`/`react` arrives wherever the harness Service pinned the agent's gateway — not the holder — so those calls hop to the holder over the Redis bus. The subtle turn-resolution logic in `slack.ts` is untouched and correct again, because exactly one worker exists.

**Harness session affinity — this one was already broken at any replica count.** The harness Service declared `sessionAffinity: ClientIP` to keep in-process MCP sessions on the replica that opened them. It never took effect: the Service is waypoint-fronted, and `sessionAffinity` is a kube-proxy feature the waypoint's Envoy bypasses. The waypoint's harness cluster read `LEAST_REQUEST`, load-balancing per request, so every agent's MCP tool calls would have broken at the first scale-up. Fixed with a consistent-hash `DestinationRule`.

**OAuth refresh backoff** lived in a process-local Map while the sweep runs as a periodic job on whichever replica draws the tick — so each tick saw a different empty heap and the backoff never applied. Now persisted on the connection's `auth` jsonb (no migration).

**Terminal relay** was keyed on the session id alone. Session ids are only unique within an agent and the `?? "default"` fallback is shared, so one agent's terminal could evict another's — a live bug at a single replica. Eviction was also process-local, leaving two tabs interleaving input into one PTY.

## Two bugs found while fixing

- `stopAll()` unsubscribes the rxjs lifecycle subscriptions, so calling it on a lease loss would permanently break a replica that later won the lease back. Split out `standDown()`.
- I had assumed the OAuth success path "replaces `auth` wholesale". It doesn't — all eight callers *spread* it, so a persisted backoff would have survived a successful refresh and throttled a healthy connection forever. Widened `withoutRefreshFailureMarker` to clear both records.

## Verification

- 846 unit tests pass (99 files); full `mise run check` green.
- New checks for the lease, the cross-replica RPC hop, backoff persistence, and terminal keying. The terminal-keying test was confirmed to fail against the old code rather than pass vacuously.
- **On a live cluster at 2 replicas:** exactly one lease holder; failover to a survivor in under 10s; waypoint cluster confirmed flipping `LEAST_REQUEST` → `RING_HASH`.

## Verification of the e2e tier

`mise e2e` passes in CI (10m20s). It could not be run during development —
the harness needs a separate `platform-k3s-test` VM on port 5555, and the
sandbox this was built in has a single k3s with ingress on 4444, so both
`e2e` and `e2e:loop` failed at the routing precheck before any spec ran.
CI has the right topology and the smoke tier is clean.

## Trade-offs worth a reviewer's judgment

- Channel throughput is now one replica's. Slack's own ten-connection Socket Mode ceiling is lower, so this likely isn't the binding constraint — but it is a deliberate cap, chosen over sharding the turn state into Redis (which would have meant rewriting the logic the #2952 guard depends on, and moving Telegram to webhooks, adding a public-ingress requirement for operators).
- The lease is unfenced: a stalled holder can briefly overlap with its successor. For the inbound socket that is benign (Slack load-balances across the app's connections). It is *not* benign for the outbound RPC — `PUBLISH` reaches every subscriber, so both holders would run the request and double-post — so each request id is claimed atomically and only the winner executes. Documented as a constraint on reusing the primitive.


packages/api-server/src/core/bus-rpc.ts
```diff
@@ -0,0 +1,131 @@
...
+          pending.delete(id);
+          reject(
+            new Error(`${opts.service} rpc timed out after ${timeoutMs}ms`),
+          );
+        }, timeoutMs);
+        timer.unref?.();
+        pending.set(id, { resolve, reject, timer });
+        void opts.bus.publish(
+          requestChannel,
+          JSON.stringify({
```
dam-code-guardian, line 93: 🔴 **Critical:** `call` serializes the request with `JSON.stringify`, but an outbound `postMessage` carries `attachment.data` as a Node `Buffer` (`ChannelAttachment`, built in `mcp-endpoint.ts` from the read file). A `Buffer` does not survive JSON — the leader's `serve` parses it back as `{type:"Buffer",data:[…]}`, which is not a `Buffer`. The Slack `uploadFile({ file: attachment.data })` (`slack.ts:2175`) and the Telegram upload (`telegram.ts:600`) then receive a malformed object, so a file attachment sent by an agent pinned to a **non-leader** replica fails or corrupts — exactly the multi-replica path this PR enables. Text-only sends are unaffected, and no test exercises an attachment over the hop. Also, a 10 MB file becomes a ~30 MB integer-array pub/sub frame. Consider carrying attachment bytes as base64 (or a shared-store reference) across the hop and rebuilding the `Buffer` on the serving side.

packages/api-server/src/modules/channels/services/channel-manager.ts
```diff
@@ -238,6 +398,25 @@ export function createChannelManager(deps: {
...
+      // Called only on the replica holding the channel lease — the workers
+      // are single-holder by construction (Slack Socket Mode fans each event
+      // to one connection; Telegram admits one getUpdates consumer), and
+      // every piece of per-turn state in them is in-process.
+      //
+      // Serving the rpc starts here so the hop is live for exactly as long
+      // as the workers behind it are.
+      if (rpc) {
+        stopServing?.();
+        stopServing = rpc.serve(async (req) => {
```
dam-code-guardian, line 434: 🟡 **Warning:** `bootstrap` starts `rpc.serve` on the lease holder, and the lease is unfenced (`leader-lease.ts` — a stalled holder can overlap its successor). With ≥3 replicas, during a handoff overlap two replicas both serve `rpc:channels`; a true follower's `postMessage`/`reply` then executes on **both** and posts the message twice. The PR body reasons about the overlap only for the duplicate Slack socket (which Slack load-balances), but a duplicate `chat.postMessage`/Telegram send is a real, user-visible duplicate and is not deduped. `dispatchResult` deliberately does not retry for exactly this reason; the double-serve is a second, uncovered path to the same double-post. Consider fencing the served RPC on the lease token, or documenting this outbound case alongside the socket one.

packages/api-server/src/modules/connections/services/oauth-refresh.ts
```diff
@@ -95,14 +98,11 @@ export function createOAuthRefreshLoop(deps: RefreshDeps): OAuthRefreshLoop {
...
-      const dueIds = new Set(due.map((c) => c.id));
-      for (const id of backoff.keys()) {
-        if (!dueIds.has(id)) backoff.delete(id);
-      }
       for (const conn of due) {
         if (conn.auth.kind === "oauth" && !conn.auth.refreshTokenRef) continue;
-        const state = backoff.get(conn.id);
-        if (state && startedAt < state.nextAttempt) {
+        const state =
+          "refreshBackoff" in conn.auth ? conn.auth.refreshBackoff : undefined;
```
dam-code-guardian, line 104: 🟡 **Warning:** The per-tick prune that deleted backoff for connections leaving the due set is gone; backoff now lives on `auth` and is cleared only by a successful token write (`withoutRefreshFailureMarker`). A connection that leaves and re-enters the due set without a success keeps its persisted `refreshBackoff` — benign only while `nextAttempt` has already elapsed, but a legitimate refresh would be skipped if any credential-write path ever bumps `expiresAt` without stripping the backoff. The retained test `prunes backoff state once a connection leaves the due set` no longer covers this: `setRows([...])` rebuilds rows from pristine `auth`, so the persisted-backoff-survives case is untested. Add a case that re-enters the same row with `refreshBackoff` intact and `nextAttempt` still in the future.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `77fd59fc`

## PR #3223: fix(api-server): make the api-server safe to run at more than one replica
**Author:** JanPokorny | **Branch:** fix/api-server-multi-replica-safety → main | **Changes:** +1273 −148 (21 files)

### Summary
Audits the api-server for single-replica assumptions and fixes them: a Redis leader lease elects one channel-worker holder, outbound channel calls hop to it over a Redis RPC, a source-IP-hash DestinationRule pins MCP sessions, OAuth refresh backoff moves to Postgres, and the terminal relay keys eviction by agent+session with a cross-replica supersede.

### Findings
- 🔴 **Critical:** Outbound file attachments break on a non-leader replica — a `Buffer` cannot survive the JSON bus hop (`packages/api-server/src/core/bus-rpc.ts:85`).
- 🟡 **Warning:** Unfenced-lease overlap can double-post an outbound message via a double RPC serve (`packages/api-server/src/modules/channels/services/channel-manager.ts:410`).
- 🟡 **Warning:** Persisted refresh backoff lost its due-set prune; the retained test does not cover the persisted case (`packages/api-server/src/modules/connections/services/oauth-refresh.ts:104`).
- ✅ **Looks good:** The lease compare-and-set extend/DEL and stand-down-on-Redis-error path are correct — a replica that stalled past the TTL cannot steal the key back, and a partitioned replica stands down.

**Verdict:** aligned. Both edited architecture pages (`channels.md`, `platform-topology.md`) bumped `Last verified:` to 2026-08-07 and reflect the new single-holder lease, RPC hop, source-IP hash, terminal-supersede, and Postgres backoff.

Possible drift:
- ⚠️ `connections.md` (Last verified 2026-08-04, unchanged) describes the refresh lifecycle and the permanent `expired` marker, but not the new transient `refreshBackoff` now persisted on the connection `auth`. Confirm whether that paragraph should note the persisted retry backoff.

### Issue Fit

✅ The diff delivers every item the PR body promises — lease, cross-replica RPC hop, DestinationRule, backoff persistence, terminal keying, `standDown` split. No undeclared scope. (`#2952` is referenced as context, not a closing link.)

- 🟢 `recordBackoff` builds storage-specific `jsonb_set` SQL in the services layer (`oauth-refresh.ts:190`). This extends the file's pre-existing direct-DB pattern (`dueConnections`, `markRefreshFailure`) rather than a repository port (slice-composition.md — infrastructure owns persistence logic). Consistent with the existing code; noted as amplified, not new. No new layering, event-bus, or module-boundary violations introduced; `withoutRefreshFailureMarker` stays a pure domain function.

### Verdict
**REQUEST_CHANGES** — the outbound file-attachment path is silently broken on the multi-replica configuration this PR is meant to make safe.

---

JanPokorny:
All four addressed in `a7e1101d`, `289903ac`, `bb2cf56a`.

**🔴 Attachment Buffer over the JSON hop — confirmed, and worse than reported.** Beyond the corruption (`{type:"Buffer",data:[…]}` arriving as a plain object), the array-of-bytes encoding inflates ~4x, so the 10 MiB the MCP tool permits would produce ~40 MB of JSON — past the default 32 MB pub/sub client output buffer, dropping the subscriber. Fixed by keeping the bytes off the bus entirely: a short-lived Redis handoff carries them, only the key travels. A handoff that can't be read **refuses the send** rather than posting the message with the file silently missing, which seemed the worse failure to leave in.

**🟡 Double-serve on lease overlap — confirmed, and it invalidated a claim in my PR description.** I'd written that lease overlap was benign because Slack load-balances the socket. That's true for inbound and false for this path: `PUBLISH` fans out to every subscriber, so both holders execute the request and the conversation gets two posts. Now each request id is claimed atomically and only the winner executes; a claim that can't be established declines rather than risking the duplicate. PR body corrected.

**🟡 Prune removal + masking test — right about the test, and it was masking exactly as described.** `setRows` rebuilt the row from scratch between ticks, so the fresh row carried no backoff and the assertion held either way.

On the prune itself I don't think there's a defect to fix: persisted state needs no memory pruning, and every credential write strips the record via `withoutRefreshFailureMarker`. But the *behaviour* did change and the old test name asserted the opposite, so I've rewritten it against the real contract — a connection that leaves the due set and returns keeps its backoff instead of getting a free retry, and a success is what clears it. The harness can now take rows out of the due set without discarding what the loop wrote on them.

**⚠️ doc-drift on `connections.md` — agreed, added.** The lifecycle paragraph covered only the permanent marker, so there was no account of why a persistently failing renewal isn't retried every sweep. The page was at its 40k cap, so it's paid for by raising the rotation's SDS mechanics and the permanent-rejection examples off code level, and dropping the one sentence of decision rationale (why auth is separate from contributions) that belongs to the ADR log.

**On the TypeScript note (🟢):** agreed that `recordBackoff` extends the file's direct-DB pattern rather than a repository port. I kept it consistent with its neighbours (`dueConnections`, `markRefreshFailure`) rather than introducing a port for one caller — happy to extract all three if you'd prefer that as a follow-up, but it felt out of scope for a fix PR.

Verification: 849 unit tests pass, full `mise run check` green. Both new checks were confirmed to **fail** against the previous code rather than pass vacuously. Re-verified on a 2-replica cluster: one lease holder, no errors on either replica.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `bb2cf56`

## PR #3223: fix(api-server): make the api-server safe to run at more than one replica
**Author:** JanPokorny | **Branch:** fix/api-server-multi-replica-safety → main | **Changes:** +1584 −164 (23 files)

### Summary
Re-review of the multi-replica safety fixes. All three findings from the previous review are resolved and verified; no new findings at this HEAD.

### Changes since last review
Previous HEAD: `77fd59fc` (2026-08-07T09:04:24Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** Outbound attachment bytes now hop via a Redis blob handoff — only the key crosses the JSON bus, and a follower refuses rather than posting a message with the file silently missing (`packages/api-server/src/core/blob-handoff.ts`, `channel-manager.ts:281`).
- ✅ **Fixed:** Double-serve during a lease handover is gated by an exactly-once `claim` before any side effect, so a request runs on one replica only (`packages/api-server/src/core/bus-rpc.ts:42`).
- ✅ **Fixed:** The persisted refresh backoff is now covered and correct — a failing connection stays in the due set and is skipped until `nextAttempt`, `failures` accumulate from the persisted row, and a successful credential write strips both `refreshFailedAt` and `refreshBackoff` (`oauth-refresh.ts`, `oauth-refresh-backoff.test.ts`); the lifecycle is documented in `connections.md`.

### Findings

**Verdict:** aligned. The previous review's `connections.md` ⚠️ is resolved — the widening-backoff-and-clear-on-success lifecycle is now documented and `Last verified` is 2026-08-07. The outbound blob handoff stays an implementation detail below architecture level.

### Issue Fit
No linked issue — `#2952` is referenced as context, not a closing link. The delta (blob handoff, exactly-once claim, tests, doc) stays within the declared multi-replica-safety scope. No undeclared scope.

✅ No new architecture violations. `blob-handoff.ts` follows the existing `core/` primitive pattern (peer to `bus-rpc`/`redis-bus`/`leader-lease`) and is injected into the service as a `BlobHandoff` port. 🔁 1 note from the previous review still present (`recordBackoff` builds `jsonb_set` SQL in the services layer — amplified, not new; see review at `77fd59fc`).

### Verdict
**APPROVE** — every previously requested change is delivered and verified; no new issues.

---

PR #3225, opened 2026-08-07
feat(connections): pick a GitHub App connection's scope from the installation

pilartomas:
Follow-up to #3206, from review feedback on the shipped form: the two scope fields were free text, so you had to know GitHub's permission vocabulary and type it blind against an installation whose actual grant you couldn't see.

## What changed

The api-server already holds the private key and authenticates as the app, so it can simply ask GitHub what the installation grants. A pre-create probe does that, and the form offers what came back.

**Permissions get a level, not a tick.** Each granted permission shows Off / Read / Write, capped at the level the installation holds — write is not offered where the app only has read. Plain checkboxes were the obvious design and they don't work: an app granted `contents:write` could only contribute write or nothing, so the read-only docs agent that motivated this whole feature couldn't be expressed. The level control is the only shape that can say "read, even though you may write".

**Repositories are ticked from a list** and recorded by GitHub's numeric id. #3206 stored names and flagged the rename hazard as an open question — this closes it, since an id survives a rename where a name starts failing at the next renewal. Ids and names are mutually exclusive on the wire, so ids win and names are dropped rather than sent alongside.

**Typing still works.** Until the installation has been read, and if reading it fails, the original text fields render exactly as before. A probe that can't run — unreachable host, key pasted later — never costs the user the ability to narrow.

**A stale selection is cleared.** Editing the host, app id, or installation id after picking drops both the probe result and the selection, rather than submitting one installation's repository ids against a different installation.

## Notes for the reviewer

- The probe mints one unscoped installation token to list repositories (that listing is authenticated as the installation, not as the app). It's never stored, never injected, and never leaves the api-server — the same call create already makes a moment later.
- The picker holds no copy of the selection. It renders from the form's own fields and writes back through them in the same string format the server already parses, so the contract, the parser, and every server-side path from #3206 are untouched.
- `repositoryIds` is picker-only in the UI — nobody types numeric ids by hand — but stays a declared input so the CLI (`--repository-ids`) and scripted callers can set it.

## A mistake worth flagging

My first cut skipped the repository listing for installations granted "all repositories", on the belief that GitHub doesn't enumerate those. **That's wrong** — `GET /installation/repositories` serves any installation, and the `selected`-only restriction applies to the DELETE endpoint, not this one. Since the same change also removed the text field, the effect would have been that GitHub's *default* install type could not be narrowed by repository at all, silently minting account-wide tokens. Caught in review, verified against GitHub's docs, and fixed: repositories are listed for both selections, and the test that had locked in the wrong behavior now asserts the right one.

## Verification

- `api-server` 856 tests, `ui` 256, `cli` 154 — all pass. Typecheck, lint, and format clean across all four touched packages, verified directly rather than through mise's cache.
- Reviewed adversarially before commit: 9 candidate findings, 5 refuted against the code, 4 confirmed and fixed — the "all repositories" bug above, the lost text fallback, the stale selection, and a GitHub Enterprise probe that fired without a host and surfaced an internal 500 message instead of staying disabled.
- Also removed a `this`-binding dependency inside the engine that would have broken if the method were ever called detached; there's now a test pinning that.
- `mise run check` still fails at the CRD-compatibility step in a worktree (known crdify/gitdir limitation, no CRDs touched).


dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `985f636`

## PR #3225: feat(connections): pick a GitHub App connection's scope from the installation
**Author:** pilartomas | **Branch:** feat/github-app-scope-picker → main | **Changes:** +983 −74 (22 files)

### Summary
Adds a pre-create probe that reads a GitHub App installation's granted repositories and permissions, so the create form offers a picker (repository checkboxes, per-permission Off/Read/Write) instead of blind text entry, and records repositories by numeric id. The design is sound and well-tested. One committed control byte, however, corrupts the central new file and blocks its review on GitHub.

### Findings
- 🔴 **Critical:** `github-app-scope-picker.tsx` carries a stray NUL byte (`0x00`) at byte offset 2320 — the identity join separator is `.join("")` where `.join(" ")` was intended (line 61). Runtime is unaffected (the value is only compared for equality), but GitHub classifies the file as **binary** (verified: the PR-files API returns `has_patch:false`, `additions:0`), so the PR's central file shows no diff and cannot take inline review comments — and it defeats the repo's own `.gitattributes` NUL guard. Replace the byte with a real space and confirm the file renders as text.
- 🟡 **Warning:** the stale-selection reset clears `repositoryIds` and `permissions` on an identity change but not the typed `repositories` names (`github-app-scope-picker.tsx:65-75`). When the repository listing is unavailable for two installations in a row, names typed for the first survive the switch and submit against the second — the exact "one installation's selection against another" hazard the reset exists to prevent.
- 🟢 **Suggestion:** `listRepositories` silently caps at 500 repositories (`MAX_REPO_PAGES × REPOS_PER_PAGE`, `github-app-engine.ts:773-777`). For an "all-repositories" installation with more than 500 repos the checkbox list is truncated with no indication and no fall-back to typing, so a repository past the cap cannot be picked. Consider surfacing when the cap is hit.
- 🟢 **Suggestion:** `github-app-scope-picker.tsx:52` casts the already-typed tRPC mutation result (`probe.data as GitHubAppInstallationProbe`); dropping the assertion lets a future contract change surface as a type error here.

**Verdict: aligned.** `security-and-credentials.md` — the page owning credential scoping — gains the read-back-the-grant paragraph and its `Last verified` is bumped to the PR date (2026-08-07). `connections.md` describes only the mint-and-refresh shape and never described scope narrowing, so this PR does not make it stale. No ADR or volatile-content leaks.

### Issue Fit
_No linked issue — scope checked against the PR description only._ (`#3206` is a PR, referenced without a closing keyword; the branch carries no issue number.)

✅ Delivers everything the PR body promises — probe, level-capped permissions, id-based repository selection, text-field fallback, stale-selection clearing, unscoped-token revoke, CLI `--repository-ids`, and the "all repositories" listing fix. No undeclared scope; the `gitHubAppApiBase` extraction and this-binding fix are declared.

✅ No architecture findings. Validation (`connectionProbeGitHubAppInputSchema`) stays in the contract package, the service orchestrates and translates errors through `rejectIfInvalid`, the domain parsers are pure and follow their siblings' throw-on-invalid convention, and `readInstallation`/`listRepositories` implement the `GitHubAppEngine` port with dependency injection intact.

🟢 One suggestion (the redundant `as` cast above). State lineage is clean — server state via the probe mutation, selection single-sourced in the react-hook-form fields with no duplicate copy. Accessibility is handled (`fieldset`/`legend`, `aria-label` on each permission radio group), and the component is split into `RepositorySection`/`PermissionSection`.

### Verdict
REQUEST_CHANGES — the logic and tests are strong, but a corrupt NUL byte hides the feature's central file from review on GitHub and defeats the repo's own safeguard; it is a one-character fix.

---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `3dafa3b`

## PR #3225: feat(connections): pick a GitHub App connection's scope from the installation
**Author:** pilartomas | **Branch:** feat/github-app-scope-picker → main | **Changes:** +1332 −74 (22 files)

### Summary
A pre-create probe reads a GitHub App installation's granted repositories and permissions, so the create form offers a picker (repository checkboxes by numeric id, per-permission Off/Read/Write capped at the granted level) with a text-field fallback. Complete re-review at the label's request: all four findings from the previous review are fixed, and no new findings appear.

### Changes since last review
Previous HEAD: `985f636` (2026-08-07T11:43:47Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** the stray NUL byte is gone; `github-app-scope-picker.tsx` renders as text and its identity join is a real space (`github-app-scope-picker.tsx:61`). Verified: the file's bytes carry no `0x00`, and the PR-files API returns a text patch.
- ✅ **Fixed:** the stale-selection reset now clears the typed `repositories` names too, not only `repositoryIds` and `permissions` (`github-app-scope-picker.tsx:65-75`).
- ✅ **Fixed:** the repository listing cap is surfaced — `readInstallation` sets `repositoriesTruncated` and the picker shows a callout with a name-them-by-text fallback (`github-app-engine.ts`, `github-app-scope-picker.tsx`).
- ✅ **Fixed:** the redundant `as GitHubAppInstallationProbe` cast is dropped; `const installation = probe.data` keeps the tRPC-inferred type (`github-app-scope-picker.tsx:52`).

### Findings
_All previously-raised findings are resolved; no new findings at this HEAD._

**Verdict: aligned.** `security-and-credentials.md` gains the read-back-the-grant paragraph and bumps `Last verified` to the PR date (2026-08-07). No ADR reference leak, no volatile-content leak, and no new subsystem needs a page.

### Issue Fit
_No linked issue — scope checked against the PR description only._ (`#3206` is a PR, referenced without a closing keyword; the branch carries no issue number.) Delivers everything the body promises — probe, level-capped permissions, id-based repository selection with text fallback, stale-selection clearing, unscoped-token mint-and-revoke, CLI `--repository-ids`, and repository listing for both installation selections. No undeclared scope creep.

✅ No architecture findings. Validation stays in the contract package (`connectionProbeGitHubAppInputSchema`); the service orchestrates and translates errors through `rejectIfInvalid`; `gitHubAppApiBase` is extracted into the domain and shared by create and probe so both mint against the same base; `parseRepositoryIds` is a pure, throw-on-invalid domain parser matching its siblings; and `readInstallation`/`listRepositories` implement the `GitHubAppEngine` port through dependency injection.

✅ No findings. Selection is single-sourced in the react-hook-form fields — the picker renders from them and writes back through `setValue` (`shouldDirty`), holding no duplicate copy. Server state comes from the probe mutation. Accessibility is handled (`fieldset`/`legend`, `aria-label` per permission group), and the component splits into `RepositorySection`/`PermissionSection`.

### Verdict
APPROVE — every finding from the previous review is fixed and verified against the files at this HEAD; the design and tests remain strong.

---

PR #3226, opened 2026-08-07
feat(bob): update Bob Shell to 2.0

xjacka:
## Summary

Bob Shell 2.0 ([changelog](https://bob.ibm.com/docs/shell/changelog)) removed the `--experimental-acp` mode our integration was built on, so this is a rewrite of the integration layer rather than a version bump. Three follow-up commits fix breakage found while verifying the rewrite against a live agent and in review.

### What changed in Bob 2.0 (relevant to us)

- CLI moved to subcommands (`bob chat` / `bob run` / `bob mcp`); `--experimental-acp`, `--yolo`, `--auth-method`, `--chat-mode` and `--max-coins` are gone
- headless surface is now `bob run --format stream-json` with native `--resume <task-id>`; the prompt is positional
- task history moved from per-chat JSON files to SQLite (`~/.bob/db/bob.db`)
- approvals moved from CLI flags to `~/.bob/settings/settings.json`
- chat modes `code`/`advanced` merged into `agent`
- instructions come from `<workspace>/AGENTS.md` and `.bob/rules[-<mode>]/`; the `system-defaults.json` + `context.fileName` lookup is gone
- remote MCP servers are keyed `url` + `transportType`; 1.x's `httpUrl` is not read
- `--instance-id` exists only on `bob chat`, not `bob run`

### Commits

**1. `feat(bob): update Bob Shell to 2.0`**

- **bob-acp-shim.mjs** — rewritten. The shim is now itself the ACP agent: one `bob run --format stream-json` per prompt turn, events translated to `session/update` frames (`isReasoning` → thought chunks, `tool_use`/`tool_result` → tool_call frames). `session/list`/`session/load` are served from the task DB; resume is native (no transcript re-injection). The sessionId↔taskId binding is learned from `result.stats.task_id` and persisted on the PVC.
- **Settings posture** — the shim merge-writes license consent, `autoApprovalEnabled`, all permission groups and a wildcard `execute_command` allowlist: the 2.x equivalent of 1.x `--yolo`. Applies to the TUI too (2.0 has no per-surface approval config), so terminal sessions now auto-approve — the trust boundary remains the platform's Envoy gateway + K8s isolation (ADR-033/038).
- **Dockerfile** — bobshell 2.0.0 (+ new SHA256); platform instructions ride `.bob/rules/platform.md`.
- **harness scripts / provider catalog / UI form** — new flags (`--mode`, `--max-cost`, `--accept-license`); `BOB_CHAT_MODES` → `agent/plan/ask`, with legacy `code`/`advanced` values pinned on existing secrets normalized onto `agent`.

Credential injection is untouched: the host (`api.us-east.bob.ibm.com`), the `Apikey` scheme and the `?key=` twin-secret all still apply; `BOBSHELL_API_KEY` is a supported alias of the new `BOB_API_KEY`.

**2. `fix(bob): re-assert platform instructions on agents upgraded to 2.0`**

`$HOME` is seeded from the image only on first boot (the init script's `.initialized` guard), so an agent created before the 2.0 layout never receives `.bob/rules/platform.md`. The stale `~/.bob/AGENTS.md` such a PVC still carries is not a path 2.0 reads (`AGENTS.md` is read from the workspace root), leaving the agent with **no platform instructions at all**. `ensureRules()` re-links it on every start, sharing the bootstrap `harness-terminal` already invokes via `--settings-only`; falls back to a copy where `$HOME` is unprivileged virtiofs and never fails the boot.

**3. `fix(bob): write MCP entries in the schema Bob 2.0 reads`**

The manifest put the server URL under `httpUrl`. 2.0 does not read that key, so the entry loaded with an empty URL and `platform-outbound` never connected — the agent had none of its `mcp__platform-outbound__*` tools. `transportType` is load-bearing, not cosmetic: absent it `bob mcp list` reports the server as sse, and SSE 400s against the platform's streamable endpoint. The mcp-entry driver could express neither shape, so it gains an optional `extraFields` merged over each entry.

**4. `fix(bob): drop the replay gate — bob run --resume does not replay history`**

The rewrite assumed `--resume` re-emits the task's stored messages before the new turn and gated the stream until the prompt echo, with the first assistant chunk as a fallback. Measured against two live tasks (one carrying two assistant turns and three tool messages), nothing is replayed — the stream opens at the echo. The assumption came from the fake `bob` used to test the rewrite, which emitted an invented replay. The gate was not inert: `replayDone` starts false on every resumed turn, so a turn opening with a tool call lost its chip whenever the echo did not match `promptText` byte-for-byte (measured: 0 → 1 `tool_call` frames). Also dropped two spots leaning on `tasks.directory`, which 2.0 always stores as `""`.

**5. `fix(bob): keep a pre-2.0 pinned mode editable, close the extraFields shape`**

Narrowing `BOB_CHAT_MODES` left the provider dialog rejecting a value every other layer tolerates: a secret pinned to `code`/`advanced` disabled Save at mount, blocking credential rotation. `normalizeBobChatMode()` gives the legacy mapping one home in the contract package and both read-back paths normalize through it, while fresh input stays strict. Plus: `extraFields` can no longer restate the keys the driver builds (`type`, `url`, `headers`, the active `urlKey` are reserved at parse time), the README lost the replay paragraph commit 4 invalidated, and the catalog's mode label was aligned with the form's.

**6. `refactor(bob): settle the coins/cost vocabulary, drop a dead error branch`**

Review follow-ups. 2.0 renamed the concept coins → cost and the labels moved with it, so the identifiers follow: `BobModelPins.maxCost`, the catalog `inputName`, the form field — while `BOB_MAX_COINS` stays the wire name, pinned on pre-rename secrets. Safe for stored connections: `inputName` is only matched against the template spec in-flight, so nothing persists under it. Separately, the form's credential-error branch could never render (the schema's only `value` issue carries the exact message the branch excluded); deleting it drops a dead path and its coupling to a display string, and takes the `watch("value")` subscription with it.

**7. `fix(bob): reserve only the keys a dialect builds, allow a sub-unit cost cap`**

Review follow-ups. The reserved-key guard was a static list, so a `urlKey` dialect — which builds `[urlKey]`, never `type`/`url` — rejected `type` with a message claiming the driver builds it; reserved keys now come from the branch in play. `--max-cost` is money, not a coin count (`MaxCostReachedError` carries it as a plain number, and an observed task settled at ~0.10), so the integer-only pattern could not express a cap below 1 — the catalog pattern and the form refinement now take a positive decimal. `dam connection connect --help` advertised `--config chatMode=code`, which the narrowed enum rejects; the example now uses `agent`.

## Testing

**Automated — committed, runs in CI.** `tsc` + lint + format clean across the five touched packages; agent-runtime 221, api-server 835, ui 246 unit tests pass. New cases cover the mcp-entry driver (`extraFields` merge, reserved-key rejection, per-branch reserving) and legacy chat-mode normalization.

**Manual — not committed.** Everything below was run by hand against the real 2.0 binary and the local k3s cluster. No test in this diff guards it; see Known gaps.

- In-cluster on a live agent with a real Bob credential: runs 2.0.0; the MCP fix confirmed end-to-end (the agent calls `list_schedules`, the platform MCP server answers `status: success`); and a follow-up prompt into a loaded chat emits only the live answer, with `session/load` rendering the stored conversation separately.
- `--resume` measured on two stored tasks, one carrying two assistant turns and three tool messages: the stream opens at the new prompt's echo and replays nothing. This is the evidence commit 4 rests on.
- The dropped gate: an ad-hoc stdio harness driving the shim against a stubbed `bob` took a resumed turn's opening tool call from 0 → 1 `tool_call`/`tool_call_update` frames once the gate was gone.
- Rules fix against a PVC whose `.initialized` predates the change (so image seeding provably did not run), plus container runs for idempotency and for leaving user-authored rule files untouched.
- MCP schema derived from what `bob mcp add -t http` writes, then each candidate shape checked through `bob mcp list`.
- Cost-cap pattern checked against both call sites — the UI's unanchored `new RegExp`, and the server wrapping it in `^(?:…)$`.

## Known gaps (not addressed here)

- **A driver-shape change does not reach existing agents.** The runtime skips dispatch when the contribution hash is unchanged, and commit 3 changes only what the driver writes, not the contributions — so on a template upgrade an existing agent keeps the old `mcp.json` and stays without MCP tools. Cleared `runtime-state.json` by hand to verify in-cluster. A fix (force one dispatch per pod boot; the drivers are declarative and idempotent) touches core runtime behaviour for every agent and is deliberately left out of this PR. **This is the one gap worth settling before a release.**
- **Sessions created under 1.x disappear from the sidebar.** They live in `~/.bob/tmp/<hash>/chats/`, the new shim reads only `bob.db`, and 2.0 ships no chat importer. New sessions are unaffected.
- `urlKey` in the mcp-entry driver has no in-tree consumer since commit 3; kept as a dialect knob (and now covered by the reserved-key check) rather than removed here.
- **The shim has no automated coverage.** `packages/agents/bob` is not a JS workspace package (no `package.json`, no vitest), so the stream-json → `session/update` translation, the task-DB reads and the `--resume` contract rest on the manual runs above — nothing guards a reintroduced replay gate. A harness needs test infrastructure for the agent-image packages, which is out of scope here.

### DIFF
--- a/packages/agent-runtime/src/modules/runtime-channel/drivers/mcp-entry-plugin.ts
+++ b/packages/agent-runtime/src/modules/runtime-channel/drivers/mcp-entry-plugin.ts
@@ -17,13 +17,34 @@ const DEFAULT_KEY_PATH = "mcpServers";
 // the entry, for those naming the transport in a sibling field (Bob 2.0 wants
 // `transportType:"http"` — without it it treats the server as SSE, which the
 // platform's streamable endpoint rejects).
-const bindingSchema = z.object({
-  impl: z.literal(IMPL_NAME),
-  path: z.string().min(1),
-  keyPath: z.string().optional(),
-  urlKey: z.string().min(1).optional(),
-  extraFields: z.record(z.string(), z.unknown()).optional(),
-});
+// `extraFields` adds siblings, never restates the transport: the keys the driver
+// builds itself stay reserved, so a manifest typo (`url: ""`) fails the binding
+// parse instead of silently clobbering the contributed URL.
+const OWNED_KEYS = ["type", "url", "headers"] as const;
+
+const bindingSchema = z
+  .object({
+    impl: z.literal(IMPL_NAME),
+    path: z.string().min(1),
+    keyPath: z.string().optional(),
+    urlKey: z.string().min(1).optional(),
+    extraFields: z.record(z.string(), z.string()).optional(),
+  })
+  .superRefine((b, ctx) => {
+    const reserved = new Set<string>([
+      ...OWNED_KEYS,
+      ...(b.urlKey ? [b.urlKey] : []),
+    ]);
+    for (const key of Object.keys(b.extraFields ?? {})) {
+      if (reserved.has(key)) {
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          path: ["extraFields", key],
+          message: `"${key}" is built by the driver and cannot be set via extraFields`,
+        });
+      }
+    }
+  });
 
 export function createMcpEntryPlugin(): Plugin {
   const fileOps = createFileOps();
--- a/packages/agent-runtime/src/modules/runtime-channel/drivers/mcp-entry-plugin.ts
+++ b/packages/agent-runtime/src/modules/runtime-channel/drivers/mcp-entry-plugin.ts
@@ -11,14 +11,18 @@ import { expandHome } from "../../../core/expand-home.js";
 const IMPL_NAME = "mcp-entry";
 const DEFAULT_KEY_PATH = "mcpServers";
 
-// MCP config is a JSON object merged under `keyPath`. `urlKey` covers harness
-// dialects: default entry is `{type:"http",url}`, but some harnesses key the
-// transport off the property name (Bob puts streamable-HTTP under `httpUrl`).
+// MCP config is a JSON object merged under `keyPath`. Two knobs cover harness
+// dialects: `urlKey` for harnesses that key the transport off the property name
+// (Bob 1.x put streamable-HTTP under `httpUrl`), and `extraFields`, merged over
+// the entry, for those naming the transport in a sibling field (Bob 2.0 wants
+// `transportType:"http"` — without it it treats the server as SSE, which the
+// platform's streamable endpoint rejects).
 const bindingSchema = z.object({
   impl: z.literal(IMPL_NAME),
   path: z.string().min(1),
   keyPath: z.string().optional(),
   urlKey: z.string().min(1).optional(),
+  extraFields: z.record(z.string(), z.unknown()).optional(),
 });
 
 export function createMcpEntryPlugin(): Plugin {
@@ -39,7 +43,7 @@ export function createMcpEntryPlugin(): Plugin {
           `plugin "${IMPL_NAME}" invalid binding: ${parsed.error.message}`,
         );
       }
-      const { path, keyPath, urlKey } = parsed.data;
+      const { path, keyPath, urlKey, extraFields } = parsed.data;
       const effectiveKey = keyPath ?? DEFAULT_KEY_PATH;
       let stateStore: McpEntryStateStore | undefined;
 
@@ -50,13 +54,11 @@ export function createMcpEntryPlugin(): Plugin {
         const entries: Record<string, unknown> = {};
         for (const c of contributions) {
           if (c.kind !== "mcp-entry") continue;
-          entries[c.name] = urlKey
-            ? { [urlKey]: c.url, ...(c.headers ? { headers: c.headers } : {}) }
-            : {
-                type: "http",
-                url: c.url,
-                ...(c.headers ? { headers: c.headers } : {}),
-              };
+          entries[c.name] = {
+            ...(urlKey ? { [urlKey]: c.url } : { type: "http", url: c.url }),
+            ...(c.headers ? { headers: c.headers } : {}),
+            ...extraFields,
+          };
         }
         const names = Object.keys(entries);
         const targetPath = expandHome(path, ctx.agentHome);
--- a/packages/agent-runtime/src/modules/runtime-channel/drivers/mcp-entry-plugin.ts
+++ b/packages/agent-runtime/src/modules/runtime-channel/drivers/mcp-entry-plugin.ts
@@ -19,8 +19,12 @@ const DEFAULT_KEY_PATH = "mcpServers";
 // platform's streamable endpoint rejects).
 // `extraFields` adds siblings, never restates the transport: the keys the driver
 // builds itself stay reserved, so a manifest typo (`url: ""`) fails the binding
-// parse instead of silently clobbering the contributed URL.
-const OWNED_KEYS = ["type", "url", "headers"] as const;
+// parse instead of silently clobbering the contributed URL. Which keys those are
+// depends on the branch in play — a `urlKey` dialect never builds `type`/`url`,
+// so reserving them there would reject a field the driver does not own.
+function ownedKeys(urlKey: string | undefined): string[] {
+  return ["headers", ...(urlKey ? [urlKey] : ["type", "url"])];
+}
 
 const bindingSchema = z
   .object({
@@ -31,10 +35,7 @@ const bindingSchema = z
     extraFields: z.record(z.string(), z.string()).optional(),
   })
   .superRefine((b, ctx) => {
-    const reserved = new Set<string>([
-      ...OWNED_KEYS,
-      ...(b.urlKey ? [b.urlKey] : []),
-    ]);
+    const reserved = new Set(ownedKeys(b.urlKey));
     for (const key of Object.keys(b.extraFields ?? {})) {
       if (reserved.has(key)) {
         ctx.addIssue({
--- a/packages/agents/bob/README.md
+++ b/packages/agents/bob/README.md
@@ -78,7 +78,7 @@ Bob 2.0 persists every task to SQLite on the PVC (`~/.bob/db/bob.db`, tables `ta
 
 - **`session/list`** — one entry per non-archived task (title from the task row, `_meta.platform.mode: chat`).
 - **`session/load`** — replays the task's stored messages as `user_message_chunk` / `agent_message_chunk` updates.
-- **Resume (prompt into a loaded task)** — native: the shim runs `bob run --resume <task-id> -p <prompt>`. No transcript re-injection; Bob reloads its own context. On `--resume` Bob re-emits the task's stored history on the stream first — the shim drops everything until the echo of the new prompt.
+- **Resume (prompt into a loaded task)** — native: the shim runs `bob run --resume <task-id> -- <prompt>` (the prompt is positional). No transcript re-injection; Bob reloads its own context. The stream carries only the new turn — `--resume` does not re-emit stored history, verified against tasks with several assistant turns — so nothing needs filtering beyond the prompt echo. The turn must run in the task's own workspace or Bob rejects the task outright.
 
 ACP session ids issued by `session/new` are shim-generated (`bob-<uuid>`); the sessionId↔taskId binding is learned from the first turn's `result.stats.task_id` and persisted to `~/.bob/platform-shim-sessions.json` so loaded sessions stay resumable across pod restarts.
 
--- a/packages/agents/bob/README.md
+++ b/packages/agents/bob/README.md
@@ -6,15 +6,17 @@ Platform agent running [Bob Shell](https://internal.bob.ibm.com/docs/shell) — 
 
 | Component | Source | Purpose |
 |---|---|---|
-| Harness | `bobshell` (installed from `bob.ibm.com/download/bobshell.sh`) | Bob CLI in `--experimental-acp` mode + native TUI |
-| ACP bridge | `bob-acp-shim.mjs` | Translates Bob's session/update events into the shape the platform UI expects; auto-approves `session/request_permission` (Bob emits it only for file edits in yolo — see autonomy posture below); stages chat attachments into the workspace so Bob can read them (it can't consume `resource_link` blocks in ACP mode); emulates `session/list` + `session/load` from Bob's on-disk chats so the sidebar and chat resume work (see below) |
-| Storage | `/home/agent` PVC | Bob's session index lives under `~/.bob/`; chat history under `~/.bob/tmp/<projectHash>/chats/`; survives pod restarts |
+| Harness | `bobshell` 2.x (installed from the `bob-shell` COS bucket tarball) | Bob CLI: `bob run --format stream-json` for chat turns + `bob chat` TUI for terminal |
+| ACP bridge | `bob-acp-shim.mjs` | ACP agent for agent-runtime; spawns one `bob run` per prompt turn and translates its stream-json events into `session/update` frames; serves `session/list` + `session/load` from Bob's task DB; continues tasks with the native `--resume <task-id>` |
+| Storage | `/home/agent` PVC | Bob's task history lives in SQLite under `~/.bob/db/bob.db`; settings under `~/.bob/settings/`; survives pod restarts |
+
+Bob 2.0 removed the `--experimental-acp` mode the previous shim bridged, so the shim is now itself the ACP endpoint rather than a frame-by-frame translator. The headless contract it builds on: `bob run --format stream-json` emits `message` (with `isReasoning` separating thoughts from answers), `tool_use`, `tool_result`, `result` (whose `stats.task_id` names the created/resumed task), and `error` events.
 
 ## Authentication
 
-Bob expects `BOBSHELL_API_KEY` in the pod env. On the platform the agent only ever sees a **placeholder** — the real key is materialized at the Envoy sidecar, never in the agent container.
+Bob expects `BOBSHELL_API_KEY` in the pod env (2.0 renamed it to `BOB_API_KEY`, but the old name is a supported alias — Bob copies it over and errors only if both are set to different values). On the platform the agent only ever sees a **placeholder** — the real key is materialized at the Envoy sidecar, never in the agent container.
 
-1. **Open Settings → Providers → Bob Shell** and paste your Bob API key. The provider preset creates a secret pinned to `api.us-east.bob.ibm.com` (the host Bob's bundle uses for `/v1/model/info`, `/key/info`, `/chat/completions` etc. with the current opaque api-key format) with `Authorization: Apikey {value}` injection plus a twin entry on the same host that handles the `?key=` URL parameter Bob appends to several admin endpoints. `BOBSHELL_API_KEY` is seeded as `dummy-placeholder` — the literal content is irrelevant because Envoy overwrites the wire value, but it must not start with `sk-`/`pk-` or Bob's bundle would silently downgrade to the legacy `prod.ibm-bob-staging.cloud.ibm.com` backend (which only accepts JWT keys). The Advanced disclosure lets you set the default model and tenant-scoping flags (see below) — those flow as additional env-mappings rather than free-form env vars in the agent dialog.
+1. **Open Settings → Providers → Bob Shell** and paste your Bob API key. The provider preset creates a secret pinned to `api.us-east.bob.ibm.com` (unchanged in 2.0) with `Authorization: Apikey {value}` injection plus a twin entry on the same host that handles the `?key=` URL parameter Bob appends to several admin endpoints. `BOBSHELL_API_KEY` is seeded as `dummy-placeholder` — the literal content is irrelevant because Envoy overwrites the wire value, but it must not start with `sk-`/`pk-` or Bob's bundle would silently downgrade to the legacy `prod.ibm-bob-staging.cloud.ibm.com` backend (which only accepts JWT keys). The Advanced disclosure lets you set the default model and tenant-scoping flags (see below) — those flow as additional env-mappings rather than free-form env vars in the agent dialog.
 
 2. **Grant the secret to the Bob agent instance** from Configure Agent → Secrets. The next pod restart picks up `BOBSHELL_API_KEY` and any pinned `BOB_*` envs along with the Envoy filter chain.
 
@@ -26,23 +28,15 @@ Some Bob backends (`/key/info?key=<value>`) read the credential from a URL query
 
 ## Autonomy posture
 
-Bob runs with `bob --experimental-acp --yolo --auth-method api-key` (in [`bob-acp-shim.mjs`](bob-acp-shim.mjs)) and the shim auto-approves any `session/request_permission` Bob emits, so the platform UI never shows a per-tool confirmation chip for Bob.
+`bob run` (headless) has no per-tool prompt channel and no `--auto-approve` flag — approvals ride the settings file. The shim merge-writes `~/.bob/settings/settings.json` on startup with the platform posture: `approval.autoApprovalEnabled: true`, all permission groups (`read`, `edit`, `execute`, `browser`, `mcp`) allowed, a wildcard `execute_command` allowlist, and `outsideWorkspaceAllowed: true`. This is the 2.x equivalent of 1.x `--yolo`; there is no per-tool HITL in either surface.
 
-What each approval mode actually does over ACP (verified empirically against 1.0.6):
+The same settings apply to the terminal TUI (`bob chat`), so terminal sessions auto-approve too — unlike 1.x, where the TUI ran in `auto_edit` and asked before risky tools. Approval granularity is a Bob-side concern; the trust boundary is the platform's, not Bob's: every outbound HTTP request goes through the per-instance Envoy egress sidecar (ADR-033/038) with `ext_authz` / egress-rules enforcement, the agent container has no SA token, no Secret volume mounts, and no Envoy config it can rewrite.
 
-| mode | file edits | shell exec |
-|---|---|---|
-| `default` | tool not in the model's tool list | tool not in the list |
-| `auto_edit` | run auto-approved, no permission request | tool not in the list |
-| `yolo` | emit `session/request_permission` (kind=edit) — the shim auto-approves | run directly, never asks |
-
-`--yolo` is the only mode with shell exec, and non-yolo modes hide (rather than gate) the missing tools — so yolo is the only generally useful setting for a coding agent, and per-tool HITL is upstream-limited to edits. The trust boundary is the platform's, not Bob's (below).
-
-The trust boundary is the per-instance Envoy egress sidecar (ADR-033/038): every outbound HTTP request goes through the credential gateway with `ext_authz` / egress-rules enforcement, the agent container has no SA token, no Secret volume mounts, and no Envoy config it can rewrite.
+Bob's own guardrail that remains active: `session.maxTurns` (Bob default 100) ends a runaway task, and `--max-cost` (below) caps spend per task.
 
 ## Configuration
 
-Bob accepts settings from two channels: env vars it reads directly (resolved by Bob's runtime) and CLI flags that have no env equivalent (the harness scripts translate platform env vars into the right flag). The Bob Shell provider preset pins the most common ones; the rest stay free-form in **Configure Agent → Env**.
+Bob 2.x accepts settings from `~/.bob/settings/settings.json` (merged by the shim from the env vars below) and CLI flags (translated by the harness scripts). The Bob Shell provider preset pins the most common ones; the rest stay free-form in **Configure Agent → Env**.
 
 ### Pinned via the Bob Shell provider (Settings → Providers → Bob Shell → Advanced)
 
@@ -51,11 +45,11 @@ These ride on the secret's `envMappings`, so every agent granted the Bob secret 
 | Env var | Translated to | Effect |
 |---|---|---|
 | `BOBSHELL_API_KEY` | n/a (env-only) | API key the Envoy sidecar swaps to the real value on the wire. Always emitted. |
-| `BOB_SHELL_MODEL` | n/a (env-only) | Default model for new sessions. Examples: `premium-shell`, `codestral-2508`, `claude-sonnet-5`. Empty → Bob's built-in default. |
-| `BOB_INSTANCE_ID` | `--instance-id` | Sets the `x-instance-id` header on Bob's outbound API calls (IBM tenant scoping). |
-| `BOB_TEAM_ID` | `--team-id` | Sets the `x-team-id` header. |
-| `BOB_MAX_COINS` | `--max-coins` | Budget cap — Bob exits with code 1 if exceeded. |
-| `BOB_CHAT_MODE` | `--chat-mode` | One of `plan`, `code`, `advanced`, `ask`. Sets the default chat persona for new sessions. |
+| `BOB_SHELL_MODEL` | `session.model` in settings | Default model for new tasks. Examples: `premium-shell`, `codestral-2508`, `claude-sonnet-5`. Empty → Bob's built-in default. |
+| `BOB_INSTANCE_ID` | `bob chat --instance-id` (terminal only) | IBM tenant scoping. `bob run` has no instance flag in 2.x — headless instance selection goes through Bob profiles, so this pin does not apply to chat-mode sessions. |
+| `BOB_TEAM_ID` | `--team-id` | Team ID for `general`-type API keys. |
+| `BOB_MAX_COINS` | `--max-cost` + `session.maxCost` | Per-task cost cap — Bob stops the task when exceeded. |
+| `BOB_CHAT_MODE` | `--mode` | One of `agent`, `plan`, `ask` (2.0 merged `code`/`advanced` into `agent`; legacy pinned values are mapped onto `agent`). Sets the default mode for new sessions. |
 
 Per-agent overrides for any of these still work — set the same env name in **Configure Agent → Env** and it wins over the inherited pin.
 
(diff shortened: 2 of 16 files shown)


packages/agent-runtime/src/modules/runtime-channel/drivers/mcp-entry-plugin.ts
```diff
@@ -50,13 +54,11 @@ export function createMcpEntryPlugin(): Plugin {
...
-            ? { [urlKey]: c.url, ...(c.headers ? { headers: c.headers } : {}) }
-            : {
-                type: "http",
-                url: c.url,
-                ...(c.headers ? { headers: c.headers } : {}),
-              };
+          entries[c.name] = {
+            ...(urlKey ? { [urlKey]: c.url } : { type: "http", url: c.url }),
+            ...(c.headers ? { headers: c.headers } : {}),
+            ...extraFields,
```
dam-code-guardian, line 82: 🟡 **Warning:** `extraFields` is spread **last**, so a binding can silently redefine the keys the driver owns — `type`, `url`, `headers`, or the active `urlKey`. The schema (`z.record(z.string(), z.unknown())`) accepts any key with any value, so a typo like `url: ""` in a manifest overwrites the contributed URL with no parse error. Before this change the entry shape was closed. The new test only covers the additive case (`transportType` alongside `type`/`url`), so precedence is unspecified rather than decided.

**Fix:** treat driver-owned keys as reserved for any merged-in field map — reject (or drop with a log) `type`, `url`, `headers`, and the active `urlKey` at parse time, and narrow the record's value type to `z.string()`, so a manifest can add a sibling field but never restate the transport the driver just built.

packages/agents/bob/README.md
```diff
@@ -65,35 +59,29 @@ Less common toggles, not surfaced on the provider card.
...
-- **Resume (prompt into a loaded chat)** — Bob can't continue a session it didn't create in the current process, so the shim lazily issues its own `session/new`, maps the old↔new `sessionId` (rewriting ids both ways from then on), and prepends a transcript of the prior conversation to the first prompt so Bob resumes with context. The transcript is capped at `BOB_RESUME_MAX_MESSAGES` (default 40) trailing messages; it costs tokens/coins.
+- **`session/list`** — one entry per non-archived task (title from the task row, `_meta.platform.mode: chat`).
+- **`session/load`** — replays the task's stored messages as `user_message_chunk` / `agent_message_chunk` updates.
+- **Resume (prompt into a loaded task)** — native: the shim runs `bob run --resume <task-id> -p <prompt>`. No transcript re-injection; Bob reloads its own context. On `--resume` Bob re-emits the task's stored history on the stream first — the shim drops everything until the echo of the new prompt.
```
dam-code-guardian: 🟡 **Warning:** This line still describes the replay gate that the PR's last commit (`7157b37`) deliberately removed: "On `--resume` Bob re-emits the task's stored history on the stream first — the shim drops everything until the echo of the new prompt." The shim's own header now states the opposite, measured against live multi-turn tasks (`bob-acp-shim.mjs:27-33`), and no gate exists in `handleEvent`. The README also still shows `-p <prompt>` while `buildRunArgs` passes the prompt positionally after `--`.

**Fix:** when a commit reverses a mechanism, sweep the prose that documents it in the same commit — drop the replay sentence and correct the flag form here, leaving the shim comment as the single account of resume behaviour. The same stale-replay claim is in the PR description's commit-1 summary and Testing section.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `7157b37`

## PR #3226: feat(bob): update Bob Shell to 2.0
**Author:** xjacka | **Branch:** feat/bob-shell-2 → main | **Changes:** +632 −821 (11 files)

### Summary
This PR rewrites the Bob Shell integration for 2.0, which removed the `--experimental-acp` mode the old shim bridged. The shim is now itself the ACP agent: it spawns one `bob run --format stream-json` per turn, serves session history from Bob's SQLite task DB, and merge-writes the approval posture that replaces 1.x `--yolo`. The migration is well evidenced — the mechanism claims in the body were checked against the code and hold.

### Findings
- 🔴 **Critical:** Any existing Bob secret pinned to `code` or `advanced` makes the provider form invalid at mount, on a field the user never touched (`packages/ui/src/modules/providers/components/bob/form.tsx:80`). `bobPinsFromConnection` reads `BOB_CHAT_MODE` verbatim into `defaultValues.chatMode`, `superRefine` (lines 44-54) rejects it against the newly narrowed `BOB_CHAT_MODES`, and `submitDisabled = isSubmitting || !isValid` disables Save. The dialog only rotates the credential on an existing connection — config inputs there do not even persist — so the user is blocked from the one available action by a value that is ignored. The runtime deliberately tolerates these values: `normalizeMode()` (`bob-acp-shim.mjs:94-96`) and the `case` in `harness-terminal.sh:10` both map them onto `agent`. This form is now the only layer that hard-rejects what every other layer accepts.
  **Fix:** normalize legacy values wherever a pinned mode is read for display or validation, rather than rejecting them — map `code`/`advanced` → `agent` when seeding the defaults, and keep the schema strict for freshly typed input.
- 🟡 **Warning:** `extraFields` can overwrite the entry contract (`packages/agent-runtime/src/modules/runtime-channel/drivers/mcp-entry-plugin.ts:60`)
- 🟡 **Warning:** README still documents the removed replay gate (`packages/agents/bob/README.md:81`)
- 🟡 **Warning:** The chat-mode field stayed at 1.x wording while its sibling hints in the same file were updated (`packages/api-server/src/modules/connections/domain/catalog.ts:231`). It still reads `label: "Chat mode"` and "Default chat persona", but the form now reads `label="Mode"` and `BOB_CHAT_MODE → --mode`. The same field disagrees across two surfaces a user can reach. Related: the updated `Instance ID` hint says Bob 2.x honours it "in terminal (TUI) sessions only", but `harness-terminal.sh:14` still passes `--instance-id` to `bob chat` unconditionally, so one of the two is wrong.
  **Fix:** when a rename lands, update every surface that names the field — align this label and hint with the form's, and reconcile the Instance ID claim with what the harness passes.
- 🟢 **Suggestion:** The `code`/`advanced` → `agent` mapping is now restated in three places — `normalizeMode()` in the shim, a shell `case` in `harness-terminal.sh`, and (missing) the UI. Exporting the legacy map beside `BOB_CHAT_MODES` in `providers.ts` would give the rule one home and let the 🔴 above reuse it.
- ✅ **Looks good:** Dropping `export BOBSHELL_NO_RELAUNCH=true` from `harness-terminal.sh` looks risky but is correct — the 1.x guard existed because bare `bob` re-exec'd itself and broke the node-pty TTY chain. `bob chat` is a different entry point, and the PR verified the TUI live.

**Verdict** — aligned

### Issue Fit

_No linked issue — scope checked against the PR description only._ No closing keyword, plain `#<n>` reference, or issue URL appears in the body, and the head branch (`feat/bob-shell-2`) carries no issue number.

- 🟡 **The PR body describes a replay gate the diff does not contain.** The commit-1 summary states "On `--resume` Bob replays stored history first — a replay gate drops it until the echo of the new prompt", and the Testing section claims a fake-`bob` harness "covering event translation, `--resume`, the replay gate and mapping persistence". No such gate exists in the code: `handleEvent` drops only `message role=user` echoes and forwards every assistant message unconditionally (`packages/agents/bob/bob-acp-shim.mjs:426-428`), and the shim's own header comment reaches the opposite conclusion — "`bob run --resume` does NOT re-emit the task's stored messages… every event of a turn is live and the translation above needs no gating" (`packages/agents/bob/bob-acp-shim.mjs:27-33`). Either the body is stale against a later revision of the shim or the behaviour was never implemented; a reader takes the body's word for a resume safeguard that isn't there. **Fix:** drop the replay-gate sentence from the commit-1 summary and the Testing bullet, leaving the shim comment's "no gating needed, verified against multi-turn tasks" as the single account of resume behaviour — or, if replay was in fact observed, add the gate and make the two agree.

Everything else the body promises is present and matches: the shim rewritten as an ACP agent over `bob run --format stream-json` with native `--resume` and `stats.task_id` binding persisted to the PVC (`bob-acp-shim.mjs:383-402`, `:171-177`), `session/list`/`session/load` served from `~/.bob/db/bob.db`, the settings posture — `licenseConsent`, `autoApprovalEnabled`, all five permission groups, wildcard `execute_command` allowlist — merge-written and shared with the TUI via `--settings-only` (`:111-140`, `harness-terminal.sh:5`), `ensureRules()` re-linking `.bob/rules/platform.md` with the virtiofs copy fallback (`:147-169`), bobshell 2.0.0 plus new SHA in the Dockerfile, the `--mode`/`--max-cost`/`--accept-license` flag migration with legacy `code`/`advanced` normalized onto `agent`, and `extraFields`/`transportType: http` replacing `urlKey: httpUrl` in the mcp-entry driver and manifest.

No undeclared scope creep: the `mcp-entry-plugin` driver and its unit test, the provider catalog, `BOB_CHAT_MODES`, and the UI form are all named in the body's commit summaries, and the now-consumerless `urlKey` knob is explicitly declared under Known gaps.

# TypeScript Engineering — Review Mode

Intent: **review** (PR scope — #3226, `feat/bob-shell-2` vs `main`), limited to the five routed files. Baseline is the unchanged code: only violations introduced or amplified by this diff are reported. No `tseng/` metadata exists in the repo, so findings are cited directly against the architecture docs.

Reviewed:
- `packages/agent-runtime/src/modules/runtime-channel/drivers/mcp-entry-plugin.ts`
- `packages/agent-runtime/src/__tests__/unit/mcp-entry-plugin.test.ts`
- `packages/api-server-api/src/modules/connections/providers.ts`
- `packages/api-server/src/modules/connections/domain/catalog.ts`
- `packages/agents/bob/bob-acp-shim.mjs` (out of architecture scope — plain `.mjs` harness shim, no layer membership; used only as corroborating evidence)

---

## Findings

### 1. `extraFields` parses nothing and can overwrite the entry contract it is merged into

**Rule:** The validating boundary is "responsible for parsing, validating, and sanitizing external input" — a schema that accepts arbitrary keys mapped to `unknown` sanitizes nothing.
**Location:** `packages/agent-runtime/src/modules/runtime-channel/drivers/mcp-entry-plugin.ts:25` (schema) and `:57-61` (entry construction).
**Evidence:** `extraFields: z.record(z.string(), z.unknown())` is spread **last** — `...(urlKey ? { [urlKey]: c.url } : { type: "http", url: c.url }), ...(c.headers ? { headers: c.headers } : {}), ...extraFields` — so a binding may silently redefine `type`, `url`, or `headers` and the driver's own contract loses, with no parse error. Before this change the entry shape was closed and unoverridable.
**Architecture file:** `architecture/slice-composition.md` (Validation Layer — parse/validate/sanitize external input).

Corroborating: the new test (`mcp-entry-plugin.test.ts:62-75`) only covers the additive case (`transportType` alongside `type`/`url`). Nothing pins precedence, so the override semantics above are unspecified behavior rather than a decision — the field's contract is whatever the spread order happens to be.

### 2. One concept ("valid chat mode") is now validated in three places with two different answers

**Rule:** Input validation belongs to the contract package and is the single source of truth; the server implements the contract rather than restating it. Narrowing the accepted set makes the restatements disagree.
**Location:** `packages/api-server-api/src/modules/connections/providers.ts:101` (`BOB_CHAT_MODES = ["agent", "plan", "ask"]`), enforced at `packages/api-server/src/modules/connections/domain/catalog.ts:232-233` → `packages/api-server/src/modules/connections/domain/build-connection.ts:566-570`, re-validated at `packages/ui/src/modules/providers/components/bob/form.tsx:44-53`, and contradicted at `packages/agents/bob/bob-acp-shim.mjs:94-98` / `harness-terminal.sh:7-10`.
**Evidence:** The strict set is `{agent, plan, ask}`; the tolerant set that actually runs is `{agent, plan, ask, code, advanced}`, declared only in an untyped `.mjs` (`normalizeMode`) and a shell `case`. `BobModelPins.chatMode` remains `string` (`providers.ts:57`) and `configInputs` is `z.record(z.string(), z.string())` (`schemas.ts:86`), so nothing in the contract package expresses either set as a type. Concretely: a connection whose secret pins `code` runs fine, but re-saving it through the UI or the catalog check is now rejected. The change introduced this divergence by narrowing the tuple without giving the legacy-tolerant set a home in the contract.
**Architecture file:** `architecture/slice-composition.md` (Validation Layer — validation lives in the contract package, no business logic in the server's own checks) and `architecture/modules.md` (Contract Package — types + service interface define the contract).

Amplified, not introduced (noted for context, no action asked here): the enum check it now leans on throws (`build-connection.ts:567`) where the docs call for `Result<T, E>` values — `architecture/slice-composition.md` (Error Handling). Pre-existing; this diff only routes a new class of failure through it.

### 3. Ubiquitous-language split: the concept became "cost", the identifiers stayed "coins"

**Rule:** A bounded context speaks one language; the same concept must not carry two names across the code that models it.
**Location:** `packages/api-server/src/modules/connections/domain/catalog.ts:221-227` (`inputName: "maxCoins"`, `envName: "BOB_MAX_COINS"`, `label: "Max cost"`), `packages/api-server-api/src/modules/connections/providers.ts:56,75` (`maxCoins?: string`, `push("BOB_MAX_COINS", …)`).
**Evidence:** `label: "Max cost"` and `hint: "Per-task cost cap (Bob 2.x --max-cost)"` now name the concept "cost", while the domain-facing identifier (`maxCoins`), the contract field (`BobModelPins.maxCoins`), and the env name (`BOB_MAX_COINS`) still name it "coins" — one concept, two vocabularies, split across contract and domain. The env name is legitimately frozen by existing secrets; the *domain and contract identifiers* are not, and there is no `tseng/vocabulary.md` recording the alias.
**Architecture file:** `architecture/modules.md` (Identifying Bounded Contexts — "Does this area have its own language?") and `architecture/project-structure.md` (Ubiquitous Language / `tseng/vocabulary.md`).

### 4. The domain layer now encodes version-specific external-CLI behavior

**Rule:** The domain layer is pure business rules and invariants; specifics of an external system's surface belong to the adapter that talks to it.
**Location:** `packages/api-server/src/modules/connections/domain/catalog.ts:212` and `:224`.
**Evidence:** `hint: "IBM tenant scoping. Bob 2.x applies it in terminal (TUI) sessions only; chat-mode headless runs select the instance via Bob profiles."` — a domain constant now asserts which of a vendor CLI's *surfaces* honor a flag, in a named vendor *version*. The previous text stated an invariant ("Sets the x-instance-id header"); the new text states external runtime behavior implemented elsewhere (`packages/agents/bob/*`), which the domain cannot enforce and will not notice drifting. This PR's own commit 3 is that failure mode: a written-down assumption about Bob's config surface (`httpUrl`) silently stopped being true.
**Architecture file:** `architecture/slice-composition.md` (Domain Layer — pure, no framework/external knowledge) and `architecture/infrastructure.md` (External System Adapters own external specifics).

---

## Not flagged (checked, out of scope)

- `catalog.ts` importing `BOB_CHAT_MODES` / `bobEnvMappings` / `BOB_HOST` from `api-server-api` into a `domain/` file conflicts with domain purity, but the import block is unchanged on `main` — pre-existing, untouched.
- `label`/`hint`/`patternHint` presentation copy living in `domain/catalog.ts` at all: the diff edits the strings, not the structure. No amplification.
- `drivers/` as a technical-concern folder inside the `runtime-channel` module: pre-existing layout.
- `urlKey` having no in-tree consumer (acknowledged in the PR description): dead surface, but no rule in these docs governs it. It matters only as part of finding 1's fix.

---

## Smallest set of changes clearing the most findings

1. **Close the entry shape in the mcp-entry driver.** Replace the free-form `urlKey` + `extraFields` pair with one closed dialect discriminator (or, minimally, key `extraFields` to `z.record(z.string(), z.string())` and reject keys the entry already owns — `type`, `url`, `headers`, and the active `urlKey`). Clears finding 1, specifies the precedence the new test leaves open, and retires the now-consumerless `urlKey` in the same edit.
2. **Give both chat-mode sets one home in the contract package.** Export the canonical set and the accepted-with-legacy set from `providers.ts`, type `BobModelPins.chatMode` as that union, and have the shim and `harness-terminal.sh` normalize *from* it instead of re-declaring `code|advanced` locally. Clears finding 2 and removes the untyped duplicate of the rule.
3. **Reduce the two Bob hints to their invariant** and let the harness adapter/README carry version- and surface-specific behavior. Clears finding 4; while editing those entries, settle "cost" vs "coins" as one vocabulary decision — rename the domain/contract identifiers to `maxCost` and keep `BOB_MAX_COINS` as the pinned wire name with the alias recorded — which clears finding 3 without touching existing secrets.

react-ui-engineering — PR #3226 (branch feat/bob-shell-2)
Reviewed file: packages/ui/src/modules/providers/components/bob/form.tsx

Scope note: the diff in this file is text-only — four label/hint strings updated
for Bob 2.x ("Max coins"→"Max cost", "Chat mode"→"Mode", `--max-coins`→`--max-cost`,
`--chat-mode`→`--mode`, plus the Instance ID caveat). No structural, state, or
type changes. The findings below are ordered by severity; #1 is a behavioral
regression that this file's validation code causes, triggered by a sibling-file
constant change in the same PR.

---------------------------------------------------------------------------
1. CRITICAL — Legacy pinned chat mode blocks key rotation for existing users

The PR narrows the mode enum in packages/api-server-api/src/modules/connections/providers.ts:

    -export const BOB_CHAT_MODES = ["plan", "code", "advanced", "ask"] as const;
    +export const BOB_CHAT_MODES = ["agent", "plan", "ask"] as const;

This form validates `chatMode` against that constant (form.tsx:44-54):

    const cm = data.chatMode.trim();
    if (cm !== "" && !BOB_CHAT_MODES.includes(cm as (typeof BOB_CHAT_MODES)[number])) {
      ctx.addIssue({ ... message: `Must be one of: ${BOB_CHAT_MODES.join(", ")}` });
    }

with `mode: "onChange"` and `submitDisabled = isSubmitting || !isValid` (form.tsx:71-90).

The edit path pre-fills `chatMode` from the existing connection's env contribution —
`bobPinsFromConnection` reads `BOB_CHAT_MODE` verbatim
(packages/ui/src/modules/providers/components/provider-item.ts:32), passed in as
`initialPins` from provider-connect-dialog.tsx:60-62 and seeded into
`defaultValues.chatMode` (form.tsx:80).

So for any existing Bob secret pinned to `code` or `advanced` — both legal before this
PR — opening Settings → Providers → Bob Shell to rotate the token yields a form that is
invalid at mount, on a field the user never touched, with the Save/Replace button
disabled. The dialog's stated purpose makes this worse: per the comment at
provider-connect-dialog.tsx:27-29, editing "rotates the credential in place" and
config-input changes on an existing connection *don't even persist* — so the user is
blocked from the one action the dialog performs by a field whose value is ignored.

This is a UI-only regression: the runtime deliberately tolerates the legacy values.
Both consumers normalize them —

    packages/agents/bob/harness-terminal.sh:10
      case "$mode" in code|advanced) mode=agent ;; esac

    packages/agents/bob/bob-acp-shim.mjs:94-96
      function normalizeMode(mode) {
        if (mode === "code" || mode === "advanced") return "agent";
        ...

— and the constant's own new comment says "the harness still maps legacy pinned values
onto agent for existing secrets." The form is now the only layer that hard-rejects what
every other layer accepts.

Fix: normalize on load rather than reject. Map the legacy value when seeding defaults,
so the pre-filled form is valid and the pin is silently upgraded on next save:

    const LEGACY_CHAT_MODES: Record<string, (typeof BOB_CHAT_MODES)[number]> = {
      code: "agent",
      advanced: "agent",
    };
    const normalizeChatMode = (mode: string | undefined) =>
      mode ? (LEGACY_CHAT_MODES[mode] ?? mode) : "";

then `chatMode: normalizeChatMode(pins.chatMode)` at form.tsx:80. Keep the schema strict
for freshly typed input. Placing the mapping beside `BOB_CHAT_MODES` in providers.ts and
importing it here would also let the shim and harness stop hard-coding the same pair in
three places (DRY — the pattern now appears three times, which is the extraction
threshold).

---------------------------------------------------------------------------
2. HIGH — Closed 3-value enum modelled as free-text Input + datalist

`chatMode` is a closed set, and the server catalog already declares it as one:
`enumValues: BOB_CHAT_MODES` at
packages/api-server/src/modules/connections/domain/catalog.ts:233. The form instead
renders a free-text `Input` with `list="bob-chat-modes"` and a `<datalist>`
(form.tsx:187-199).

A datalist is a *suggestion* affordance, not a constraint: it permits arbitrary text,
which is exactly why the schema needs the string-membership check in finding #1 and why
a stale value can wedge the form. It is also the weakest option for accessibility and
has uneven cross-browser rendering (Safari in particular), and it silently drops the
"one of" affordance on mobile keyboards.

The repo already has vetted select primitives — packages/ui/src/components/ui/select.tsx,
rich-select.tsx, and searchable-select.tsx. Per the styling/reuse rule (reuse an existing
component before hand-rolling an input pattern), this should be a `Select` over
`BOB_CHAT_MODES` with an explicit "(Bob default)" empty option. That makes the invalid
state unreachable by construction, lets the whole `chatMode` branch of `superRefine`
disappear, and removes the `as (typeof BOB_CHAT_MODES)[number]` assertion at form.tsx:47
— an assertion the types rule ("types at boundaries, not assertions") flags anyway, since
it exists only to satisfy `.includes()` on a readonly tuple.

Note this fix subsumes the *shape* of finding #1 but not its substance: a select still
needs the legacy→`agent` normalization on load, or the pre-filled value simply won't
match any option.

---------------------------------------------------------------------------
3. MODERATE — "Max cost" renamed but still validated as a positive integer

The label and hint now say cost (form.tsx:181-182):

    label="Max cost"
    hint="BOB_MAX_COINS → --max-cost. Per-task cost cap; Bob stops the task when exceeded."

but the validation is unchanged and integer-only (form.tsx:34-43):

    !/^[1-9]\d*$/.test(data.maxCoins.trim()) → "Must be a positive integer"

Integer-only was coherent for "coins" (a discrete unit). For a *cost* cap it is
questionable in two ways: fractional caps are unexpressible (a user wanting 0.50 or 2.5
is refused), and the `[1-9]` leading class also forbids any value below 1 as well as a
literal `0`. The downstream consumers do not require an integer — both shim call sites
parse with `Number(...)` and gate on `Number.isFinite(maxCost) && maxCost > 0`
(bob-acp-shim.mjs:118/126 and 396-397), so a decimal would flow through correctly.

I can't tell from this repo whether Bob 2.x `--max-cost` takes a fractional amount, so
I'm flagging rather than asserting. If it does, relax to a positive-decimal pattern here
and in the mirrored `pattern: "^[1-9]\\d*$"` at catalog.ts:224 (the two must stay in
step). If it genuinely stays integral, the hint should say so — "positive integer" in
the hint text would stop users guessing at a decimal — since the current wording invites
input the field rejects.

---------------------------------------------------------------------------
4. MODERATE — Cross-surface wording drift: the chat-mode field was updated here but
   not in the catalog

The PR updates two catalog hints (Instance ID and Max cost, catalog.ts:212/223-224) so
they match this form, but leaves the mode field at 1.x wording:

    packages/api-server/src/modules/connections/domain/catalog.ts:230-233
      label: "Chat mode",
      hint: `Default chat persona. One of: ${BOB_CHAT_MODES.join(", ")}.`,

while form.tsx:188-189 now reads `label="Mode"` and `BOB_CHAT_MODE → --mode`. The two
surfaces (provider dialog vs. Configure Agent → Env config inputs) now disagree on both
the field name and its description, and "chat persona" is stale for 2.x where the modes
are agent/plan/ask. `catalog.ts:211` likewise still says "Default model for new
sessions" against the form's task-oriented language. Small, but this is the same field in
two places a user can reach; align the label and hint.

---------------------------------------------------------------------------
5. MODERATE (pre-existing, not introduced here) — Dead error branch keyed on a message
   string

form.tsx:142-148:

    {errors.value &&
      value.length > 0 &&
      errors.value.message !== "Required" && ( ... )}

The schema adds exactly one issue on `value`, with message `"Required"` (form.tsx:27-32).
So `errors.value.message !== "Required"` is false whenever `errors.value` is set, and
this block can never render. It is dead code, and the mechanism — branching on a
human-readable message string — is fragile coupling that would break silently on any
copy edit.

Either delete the block, or if the intent was "show token errors but not the empty-field
error" (the empty case being conveyed by the disabled button), express it with a stable
discriminator instead of prose — e.g. a `code`/`params` field on the issue, or drop the
"Required" issue entirely and let `isValid` carry the empty state, which it already does.

Outside the routed diff, so treat as touch-it-migrate-it judgment rather than a blocker
on this PR.

---------------------------------------------------------------------------
Things checked and found clean

- State lineage: no violations. Server-owned pin values arrive as props from the
  connection query and are handed straight to RHF; `advancedOpen` is genuinely
  component-local `useState`; nothing is duplicated across lineages.
- `useState` initialized from props (`variant === "edit" && hasAnyPin`, form.tsx:85-87)
  is safe here — `bobPins` is derived synchronously from `item` at
  provider-connect-dialog.tsx:60-62 and the dialog mounts with `item` already present,
  so there is no late-arriving-prop staleness.
- Form library: correctly on react-hook-form + zodResolver, past the useState threshold
  — matches the forms guidance.
- Component size and split: file is ~235 lines with a single well-named local
  presentational `PinField` helper; within budget, responsibilities are clear.
- Naming: `submitDisabled`, `hasAnyPin`, `advancedOpen`, `isEdit` all carry intent.
- `PinField` is local to this file and not duplicated in the peer provider forms
  (anthropic/openai/ibm-litellm define no equivalent), so it is not yet an
  extraction candidate — no premature-DRY complaint.
- Styling: uses shared `FormField`, `Input`, `Button`, `DisclosureToggle`,
  `externalLinkProps`; token-based classes throughout, no ad-hoc colors or inline styles.
- Password-manager suppression attrs (`data-1p-ignore`, `data-lpignore`,
  `data-form-type="other"`) are consistently applied on the credential input.
- Types: no `any`, no untyped responses; the single `as` at form.tsx:47 is the narrow
  `.includes()` widening called out in finding #2.

Verdict: the label/hint edits themselves are accurate against the harness and shim in
this PR and are fine to ship. Finding #1 should be fixed before merge — it silently
locks existing Bob users out of token rotation, and it is the kind of regression that
surfaces only for accounts that pinned `code`/`advanced`, so it will not show up in a
fresh-install smoke test. Findings #2-#4 are worth folding in while the file is open;
#5 is optional cleanup.

### Verdict
REQUEST_CHANGES — the 2.0 migration is thorough and well verified, but narrowing `BOB_CHAT_MODES` locks existing `code`/`advanced` users out of credential rotation in the provider dialog.

---

packages/agent-runtime/src/modules/runtime-channel/drivers/mcp-entry-plugin.ts
```diff
@@ -11,15 +11,40 @@ import { expandHome } from "../../../core/expand-home.js";
...
+const bindingSchema = z
+  .object({
+    impl: z.literal(IMPL_NAME),
+    path: z.string().min(1),
+    keyPath: z.string().optional(),
+    urlKey: z.string().min(1).optional(),
+    extraFields: z.record(z.string(), z.string()).optional(),
+  })
+  .superRefine((b, ctx) => {
+    const reserved = new Set<string>([
```
dam-code-guardian: 🟡 **Warning:** the reserved set is not computed from the branch that builds the keys. With `urlKey` set, line 79 builds `{[urlKey]: c.url}` and never `type`/`url` — yet both stay reserved, so a binding cannot add a sibling `type` field it legitimately owns. Reproduced against this refinement: `{urlKey:"httpUrl", extraFields:{type:"http"}}` fails the parse with `"type" is built by the driver and cannot be set via extraFields`, naming a key the driver did not build in that shape.

The additive case the new test covers (`transportType` alongside `type`/`url`) and the `httpUrl` collision case both pass correctly — only the `urlKey` + `type`/`url` combination is over-restricted.

**Fix:** derive the reserved set from the same condition that writes the entry — `urlKey` when set, otherwise `type` and `url`, with `headers` reserved in both shapes. A guard over driver-owned keys must enumerate per branch, not union across branches.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `2e46301`

## PR #3226: feat(bob): update Bob Shell to 2.0
**Author:** xjacka | **Branch:** feat/bob-shell-2 → main | **Changes:** +738 −858 (15 files)

### Summary
Two follow-up commits since the last review: `normalizeBobChatMode()` centralizes the legacy chat-mode mapping and both read-back paths use it, `extraFields` gains a reserved-key guard, and the coins→cost vocabulary is settled across the contract, catalog, and form. The 🔴 from the previous round is fixed.

### Changes since last review
Previous HEAD: `7157b37` (2026-08-10T14:31:32Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** legacy `code`/`advanced` pins wedge the provider edit form — `bobPinsFromConnection` and `bobPinsFromEnvMappings` both normalize; new test covers it (`provider-item.ts:37`)
- ✅ **Fixed:** `extraFields` could overwrite `type`/`url`/`headers` — reserved at parse time, values narrowed to `string` (`mcp-entry-plugin.ts:33`)
- ✅ **Fixed:** README documented the removed replay gate (`README.md:81`)
- ✅ **Fixed:** chat-mode label and Instance ID hint left at 1.x wording — catalog and form now agree (`catalog.ts:212`)
- 🔁 **Still present:** integer-only pattern on a monetary cost cap (`catalog.ts:225`, `bob/form.tsx:34`)
- 🆕 **New:** reserved-key check rejects `type`/`url` in the one shape that does not build them (`mcp-entry-plugin.ts:34`)
- 🆕 **New:** CLI help example pins a mode the narrowed enum now rejects (`connect.ts:155`)

### Findings
- 🟡 **Warning:** reserved keys are not computed from the branch that builds them (`packages/agent-runtime/src/modules/runtime-channel/drivers/mcp-entry-plugin.ts:34`)
- 🟡 **Warning:** the `chatMode=code` example in `dam connection connect --help` is now rejected by the very validator this PR narrowed. Verified against `validateConfigInputValue`: `code` → `Mode: must be one of ...` (`packages/cli/src/modules/connection/commands/connect.ts:155`)
  **Fix:** change the example to `chatMode=agent`. When narrowing an enum, sweep every surface that names one of the dropped values — help text and docs included, not just validators.

**Verdict** — aligned

### Issue Fit

🟡 **The "container tests" the description claims for the rules fix are not in the diff.** Testing says the rules fix has "container tests for idempotency and for leaving user-authored rule files untouched", but the PR's only test changes are `packages/agent-runtime/src/__tests__/unit/mcp-entry-plugin.test.ts` and the new `packages/ui/src/__tests__/unit/bob-legacy-pin.test.ts` — neither touches `ensureRules()` (`packages/agents/bob/bob-acp-shim.mjs:147`), and there is no container/image test harness for `packages/agents/bob` anywhere in the tree. Fix: either add the tests, or reword the Testing section to say the idempotency and user-rules checks were manual container runs.

🟡 **The "regression harness" for the dropped replay gate is likewise absent.** Commit 4's rationale rests on a measured behaviour change and Testing states "the gate removal has a regression harness showing the previously dropped opening tool call" — but no test in the diff exercises the shim's stream-json → `session/update` translation, so the 535+/723− rewrite of `packages/agents/bob/bob-acp-shim.mjs` (the `--resume` contract documented at `bob-acp-shim.mjs:27`, tool_call framing, task-DB reads) ships with no automated coverage and nothing guards the reintroduction of the gate. Fix: commit the harness that produced the 0 → 1 `tool_call` frame measurement, or drop the claim and record the measurement as manual verification.

The rest of the description matches the code: the 2.0 CLI surface (`bob run --format stream-json`, positional prompt, native `--resume`) in `bob-acp-shim.mjs:383-398`, the settings-based approval posture in `bob-acp-shim.mjs:111`, the `.bob/rules/platform.md` move in the Dockerfile, the `extraFields`/`transportType` MCP schema with its reserved-key guard in `mcp-entry-plugin.ts:32-49` and `runtime-manifest.yaml:23`, and the coins→cost rename plus `normalizeBobChatMode()` across the contract, catalog, and form. The declared extras (dead credential-error branch, `urlKey` left as a dialect knob, README replay paragraph) are all named in the description; no undeclared scope creep.

🔁 4 finding(s) from the previous review still present (see review at `7157b37`)

🆕 **Not a finding, flagged as a factual inaccuracy in new code:** in `mcp-entry-plugin.ts`, when `urlKey` is set the driver builds `[urlKey]` and no longer builds `type`/`url`, yet both stay in `OWNED_KEYS` and are rejected with the message `"type" is built by the driver and cannot be set via extraFields`. No architecture rule covers this.

🆕 **New test defeats strict typing to build its fixture** — low. `packages/ui/src/__tests__/unit/bob-legacy-pin.test.ts:5-9,22` — `}) as unknown as ConnectionView` (twice) fabricates a two-field object as a full `ConnectionView`. These are the only `as unknown as` casts in `packages/ui/src/__tests__/`. A double cast removes the compiler's ability to flag the fixture when `ConnectionView` changes shape.

**Aligned — recorded so it is not re-raised:** the `maxCoins` → `maxCost` rename against a frozen `BOB_MAX_COINS` env name is correct per the mapper rule — identifiers speak the current domain language while `bobEnvMappings`/`bobPinsFromEnvMappings` absorb the wire-format difference; applied consistently across contract, server catalog, and client. Normalizing on parse only matches the unidirectional-mapper rule.

🔁 3 finding(s) from the previous review still present (see review at `7157b37`)

The rename itself is clean and complete: `maxCoins` → `maxCost` traced through every consumer (`BobModelPins`, `bobEnvMappings`, `bobPinsFromEnvMappings`, `bobPinsFromConnection`, `bobConfigInputs`, the form schema/defaults/submit/PinField, the catalog `configInputs`) with no survivors of the old spelling and no half-renamed reader. Dropping `watch("value")` is a real improvement — it removes a per-keystroke re-render subscription that `formState.isValid` already covered.

🆕 **The replacement comment explains absent code and is anchored to the wrong statement** — moderate. `form.tsx:87-90` documents a *deletion* and sits above `submitDisabled`, a line that needs no explanation. The invariant it protects lives in the schema. Fix: move it to the `value` issue in the `superRefine` (`:27-33`) as one line. Related follow-up: `openai/form.tsx:79-84` still carries the identical now-dead block; `anthropic/form.tsx:162-167` genuinely needs its copy (its refinement gives `value` a second message).

**Checked, no action:** `defaultValues` staleness is not an issue — the dialog mounts behind a guard and `bobPinsFromConnection` reads an already-loaded `ConnectionView`. `hasAnyPin` iterates `Object.values(pins)`, so it is key-agnostic and unaffected by the rename. Each reworded hint verified against the harness: `--instance-id` is terminal-only (`harness-terminal.sh:13`, absent from the `bob run` argv), `--mode` and `--max-cost` match `harness-terminal.sh:15-16`.

### Verdict
REQUEST_CHANGES — the previous round's 🔴 is properly fixed at the contract level, but the new reserved-key guard over-restricts the `urlKey` shape and the narrowed enum left a documented CLI example invalid.


---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `cd81746`

## PR #3226: feat(bob): update Bob Shell to 2.0
**Author:** xjacka | **Branch:** feat/bob-shell-2 → main | **Changes:** +764 −861 (16 files)

### Summary
One follow-up commit since the last review. It scopes the `extraFields` reserved-key set to the branch that builds the keys, widens the cost cap to a sub-unit amount in the catalog and the form together, and corrects the CLI help example. Both open 🟡 findings are fixed and none is left.

### Changes since last review
Previous HEAD: 2e46301 (2026-08-10T16:17:35Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** reserved keys are now computed from the branch that builds them (`mcp-entry-plugin.ts:25`)
- ✅ **Fixed:** integer-only pattern on a monetary cost cap (`catalog.ts:227`, `bob/form.tsx:40`)
- ✅ **Fixed:** CLI help example pinned a retired chat mode (`connect.ts:155`)

### Findings

**Verdict:** aligned

### Issue Fit
🔁 1 finding from the previous review still present (see review at `2e46301`) — the Testing section still claims automated coverage the diff does not carry.

The follow-up commit adds no undeclared scope: commit 7 is declared in the body, and all seven files map onto it.

- 💡 The body describes commit 5 as giving the legacy chat-mode mapping "one home in the contract package", but three independent copies exist: `providers.ts:116`, `bob-acp-shim.mjs:94`, and `harness-terminal.sh:10`. The latter two cannot import the contract package. The claim holds for the two read-back paths it names; the runtime forward paths are separate mappings.

🔁 2 finding(s) from the previous review still present (see review at `2e46301`) — the hand-synced cost-cap pattern across packages, and the `normalizeBobChatMode` signature widening the declared union to `string`.

🟢 The narrowing of `bobPinsFromConnection` to `Pick<ConnectionView, "contributions">` is the correct fix for the previous round's `as unknown as` test fixture: it tightened the contract instead of faking the input. The `ownedKeys(urlKey)` refactor scopes the invariant to the branch in play, and the new test pins both directions of the rule.

**Recorded so it is not re-raised:** the shim (`packages/agents/bob`) is not a workspace package, so no architecture rule applies to its contents; the third copy of the chat-mode map there is inherent, not a defect of this change.

🔁 3 finding(s) from the previous review still present (see review at `2e46301`) — the hand-synced `maxCost` pattern and message, the `agentId` pin key contradicting `instanceId` everywhere else, and the duplicated hint prose.

🟢 The misanchored comment from the previous round now sits on the schema's `value` issue, which is the statement it explains.

**Checked, no action:** the widened regex is byte-identical to the server's (`catalog.ts:227`), and correctly rejects `0`, `0.0`, `5.`, `1e3`, and `+5`. Every changed hint string was verified against the harness: `BOB_MAX_COINS → --max-cost` and `BOB_CHAT_MODE → --mode` match `harness-terminal.sh:15-16`.

### Verdict
APPROVE — both blocking findings from the previous round are fixed, and the fixes were verified against every branch and drain point they affect. What remains is the previously-recorded cross-package duplication of the cost-cap pattern, which this PR changed correctly in lockstep and which predates it on `main`.

---

Issue #3151, opened 2026-08-04
Replies to an agent's channel updates keep their context, and channels can carry approve/reject decisions


Tomas2D:
## Context

Agent frameworks running long background work (Nous campaigns being the concrete case) post progress updates into a bound Telegram or Slack chat. #2217 made those updates arrive reliably. The next gap showed up in use: replying to an update reaches an agent that doesn't know what the update was about — and since updates often end in a question ("ship it?"), the natural next want is answering decisions from the chat, which isn't possible at all.

## Problem

**How it works today.** A bound chat feeds one long-running conversation per chat — fine for back-and-forth. But when the agent *pushes* an update (from a scheduled run or campaign), that post goes through a separate one-way door: the platform posts it and keeps no record of which message it was or what work produced it.

**What breaks.** A reply to the update ("why did CTR drop?") can't be connected to it. It lands in the chat's ordinary conversation, which was never told about the campaign — the agent that wrote the report and the agent that answers are two different minds in the same body. Slack partly self-heals because the bot can read channel history back; Telegram bots can't read history at all, so there the context is simply gone.

Related gaps in the same flow:

- **Decisions can't happen in the chat.** There's no way to answer an agent's question or a platform approval from the messenger. Worse: on channel-driven turns, harness permission prompts are silently auto-approved, and platform approval requests expire unseen in a web inbox nobody is watching.
- **Telegram forum topics fragment the binding.** Binding inside one topic leaves the bot deaf in every other topic — even though topics are exactly the "one campaign per thread" structure this use case wants.

## Goal

- Replying to an agent's update — quoted or just in the same chat — reaches an agent that knows what the update was about.
- An agent can ask a question in the chat with Approve/Reject buttons and get the answer routed back, with a record of who decided.
- Platform-level approvals can reach their owner on the messenger and be answered there — by the verified owner only.

## User Stories

- As a user following a campaign from Telegram, I want my reply to an update understood in that update's context, so that I don't have to re-explain everything.
- As an agent owner, I want the agent to ask ship/no-ship questions in the chat with buttons, so that the team can steer work without opening the web UI.
- As an agent owner, I want permission approvals delivered to me on the messenger, so that background work doesn't stall on an inbox I forgot exists.
- As an operator, I want every in-chat decision recorded with who made it and on what authority.

## Proposed solution

[...]

Issue #2326, opened 2026-07-02
PoC: show network-access approvals inside the chat session that triggered them


jezekra1:
## Problem

When an agent's work needs a network-access approval, the request surfaces in the Inbox — a separate surface, disconnected from the conversation. From the user's point of view: you're chatting with an agent, it stalls mid-turn, and nothing in the chat tells you it's waiting on you. You have to notice the Inbox, pick the right request out of the queue, and mentally connect it back to what the agent was doing.

The deeper gap is that this is not just a UI placement choice. Harness tool-permission prompts are already tied to the session they came from, but network-access requests are attributed only to the *agent* — the platform genuinely cannot tell which session caused them. One agent can run several conversations at once (chat tabs, scheduled runs, terminals), so even a diligent user can't reliably tell which conversation a pending network request belongs to, and the UI couldn't route it into the right chat even if it wanted to.

## Proposed solution

A proof of concept that establishes session attribution for network-access requests end-to-end, and uses it to render the approval where the user already is: inside the chat.

The PoC should demonstrate:

- **In-chat approval** — a network request made during a chat turn produces an approval prompt inline in that conversation; the user approves or denies there and the agent continues, without visiting the Inbox.
- **Correct routing under concurrency** — two sessions running on the same agent at the same time each get their own approvals in their own conversation.
- **Graceful fallback** — requests that can't be attributed to a session (e.g. background work that outlived its session) still land in the Inbox as they do today.
- **Unchanged trust model** — attribution is display-and-routing metadata only; what is allowed, denied, or held for approval keeps working exactly as it does now, per agent.

Initial technical analysis suggests tagging each session's outbound traffic at its source so the platform can attribute requests to sessions; validating that shape (and its cost) is the point of the PoC.

### Scope

- One harness (Claude Code) is enough to prove the concept.
- Chat sessions are the priority; terminal sessions if they come cheap.

### Dependencies

- Related to #942 (Inbox/approvals management redesign) — complementary, not replaced by it: the Inbox remains the home for unattributed and cross-agent requests, and in-chat prompts should stay consistent with whatever decision model #942 lands on.

### Out of scope

- Treating session attribution as a security boundary — it stays advisory; enforcement remains per-agent.
- SSH sessions — they have no conversation surface to render an approval in.

jezekra1:
Technical findings from the analysis that motivated this issue (verified against the code and measured on a live dev cluster, 2026-07-02).

## Why attribution is impossible today

[...]

PR #3229, opened 2026-08-07
fix(approvals,channels): refuse unattended egress instead of stalling a channel turn

### LINKED ISSUE #3228: Agents in Slack and Telegram stall for half an hour when they need network approval
## Context

Network approvals were built for someone sitting in the web app: when an agent
tries to reach a site it hasn't been allowed, the request is held open while the
owner decides in the inbox. A conversation in Slack or Telegram has no equivalent
— the request only ever shows up in the web inbox, and the people in the
conversation are usually not the owner, so the decision can't safely be handed to
whoever happens to be typing.

#3151 wants approvals delivered to the messenger and answered there, which is the
fuller fix and needs a verified identity link to be safe. This issue is the
safety valve that should hold whether or not that lands: an agent must never sit
silently on a decision nobody in the room can make.

## Problem

Someone asks an agent to do something in a bound channel or chat. The work needs
a site the agent hasn't been allowed, so the platform holds the request open
waiting for a verdict that cannot arrive from where the request came from. The
turn sits there for up to half an hour showing the agent as working, then the
request is refused and the agent answers with whatever it can make of a network
failure.

Two things go wrong at once. The wait is pure dead time — the answer was always
going to be "no" unless the owner happened to be watching the inbox. And the
agent has no idea approval is even a concept, so once the refusal lands it can't
explain the situation; it reports a bare network error, or retries into the same
wall.

Agents set to allow all network access never hit this, because nothing they reach
is unapproved. It's the safer network settings that produce the stall — which is
the wrong way round.

## Expected Behavior

- On a turn driven from a channel, a request to a site that hasn't been allowed
  is refused immediately instead of being held open. Agents allowing all network
  access are unaffected.
- The request still reaches the owner's inbox, so they can approve it later; a
  permanent approval means the agent's next attempt goes through.
- The agent knows in advance that this can happen in a conversation, so when a
  request is refused it can say so in plain terms — name what it needs, point at
  who can approve it, and offer to wait or take a different route rather than
  failing silently or retrying.

## Actual Behavior

- The turn hangs for up to half an hour (the default wait) with no sign in the
  conversation that anything is blocked.
- The request is then refused. The agent surfaces a raw network failure, or
  nothing useful.
- Nothing tells the agent that approval exists or that it can't be granted from
  the conversation, so it can't guide the person who asked.

## Steps to Reproduce

1. Bind a Slack channel (or Telegram chat) to an agent whose network access is
   one of the safer settings rather than allow-everything.
2. In that conversation, ask the agent to fetch something from a site it hasn't
   been allowed.
3. Watch the thread — the agent shows as working, with no indication it is
   blocked, for up to half an hour.
4. Check the web inbox meanwhile: the approval request is sitting there.
5. Leave it unanswered. The request expires, and the agent replies with a network
   failure that never mentions approval.

pilartomas:
Closes #3228.

## Problem

An egress request to a host the owner hasn't allowed is held open while a human decides — 30 minutes by default, with the ACP turn ceiling deliberately sized above the hold so the wait is never cut short. That works for someone sitting in the web app. From Slack or Telegram nobody can answer: the owner isn't necessarily present, and the conversation's other members aren't the owner, so handing them the decision isn't safe. The turn therefore sat silent for the whole window and was denied anyway.

Two things went wrong at once: the wait was dead time, and the agent had no idea approval was even a concept — so once the refusal landed it reported a bare network error, or retried into the same wall.

## What changed

**The gate refuses instead of holding when nobody can answer.** On a rule miss it now asks two questions before holding: is a channel turn open on this agent, and is an interactive session attached? Only when the first is true and the second is false does it refuse immediately.

The second question is what keeps this a narrowing rather than a blanket change: if the owner has a browser or CLI session on the agent, someone *can* decide, so a Slack turn running alongside one still holds and still resolves. Agents whose rules allow everything never reach the path at all — nothing they request is unmatched.

**The request is still recorded, and stays actionable.** The pending row is inserted exactly as before, so it shows up in the inbox, a permanent approval there writes the rule the agent's next attempt consumes, and retries reuse the one row instead of filing a copy each time. What's skipped is only the *wait*, plus the in-session prompt — its only consumers are the relay clients whose absence defines this path, and an unconsumed frame would leave the row looking answerable in-session when the request has already been refused. The deny is logged with its own reason and the row's id as correlation, so it's distinguishable from a rule-based deny.

**Channel turns are told this can happen.** The per-turn framing now states that an unallowed host is refused, that it can't be approved from the conversation, and what to do instead: name the host, point at the owner (the request is already waiting for them), and offer another route or to finish once it's allowed — rather than looping on the host. Shared by both messengers. It's stated unconditionally but phrased conditionally, so an allow-everything agent pays one line and nothing else.

## Why a new cross-replica signal

The gate is handed only the agent, host, method and path — it has never known which session made a request (the gap #2326 documented). The channel workers do know, so they mark the agent as channel-driven for the length of each turn, refcounted so concurrent threads each hold their own marker.

It has to be cross-replica, and #3223 (now on main) makes that sharper rather than softer: the channel workers are single-holder, so only the lease-holding replica ever marks a turn open, while an ext_authz Check still lands on whichever replica Envoy picks. An in-memory marker would have been read as absent on every other replica. The new store mirrors the existing session-presence shape (per-replica key, heartbeat, TTL) and reads the presence keys the relays already write for the "interactive session attached" half — all three relays still take that pin after #3223, and the key prefix moved into one place so the two can't drift.

Both reads **fail toward "attended"**, so a Redis blip degrades to today's hold rather than to silent denial.

## Testing

Rebased onto main at `0a52a37c`. The api-server suite passes (899 tests, 101 files) along with tsc, lint, format and the docs gates re-run directly after the rebase. New coverage:

- gate: denies without waiting, keeps the row pending, publishes no prompt, reuses one row across retries, still holds when an interactive session is attached, and never consults attendance when a rule already decides
- attendance store: refcounting, idempotent release, cross-replica visibility, local short-circuit, and the fail-toward-attended behaviour on a Redis error
- Slack: the guidance reaches the real prompt, and the marker is held across the harness call and released even when the wake fails

One caveat on the full `mise run check`: it aborts locally on the controller's CRD-backwards-compatibility task, which can't resolve the baseline tag through a worktree's gitdir (`reference not found`). That's environmental and unrelated — this branch touches no CRD, controller, or Helm file — so the remaining tasks were run per package instead. CI will run the CRD check properly.

**Not covered by a test:** the Telegram wiring. No unit harness constructs the Telegram worker today (its tests cover the message handler and flows separately), so that path is 3 lines carried by types and review. The shared guidance text it emits is tested directly.

## Notes for review

- **Out of scope, same shape:** scheduled runs have the same problem — nobody is watching, so their holds also expire unseen. Left alone deliberately; an owner who wanders past the inbox can still answer one, and widening the rule to "any unattended turn" is a bigger behaviour change than this issue asks for.
- **Complementary to #3151**, not a substitute: that issue wants approvals delivered to the messenger and answered there. This is the safety valve that should hold whether or not that lands.
- **Follow-up worth considering:** the agent only ever sees the wire symptom of a refusal (a closed connection, or a 403 on the L7 path). Envoy's denied-response body could carry a human-readable reason on TLS-terminated chains, which would beat pre-arming the prompt — but it can't reach the SNI-miss catch-all, which is where most refusals land, so the prompt framing is doing the real work either way.
- **One unrelated line trimmed** in `channels.md`: the page sat 54 characters under its size cap, so a sentence that restated "Telegram is structurally identical" from its own sentence pair made room for the cross-link. The depth lives on `security-and-credentials.md`, which owns egress approvals.

### DIFF
--- a/packages/api-server/src/modules/approvals/services/ext-authz-gate.ts
+++ b/packages/api-server/src/modules/approvals/services/ext-authz-gate.ts
@@ -53,11 +53,20 @@ export interface EgressRuleMatcher {
   ): Promise<{ verdict: ExtAuthzVerdict } | null>;
 }
 
+/** Whether anyone is positioned to answer a hold on this agent. Both facts are
+ *  cross-replica reads and must fail toward "attended" so an infrastructure
+ *  blip degrades to the ordinary hold — see `core/turn-attendance.ts`. */
+export interface EgressAttendance {
+  hasOpenChannelTurn(agentId: string): Promise<boolean>;
+  hasInteractiveSession(agentId: string): Promise<boolean>;
+}
+
 export interface CreateExtAuthzGateDeps {
   repo: ApprovalsRepository;
   bus: RedisBus;
   identityResolver: AgentIdentityResolver;
   ruleMatcher: EgressRuleMatcher;
+  attendance: EgressAttendance;
   /** Bounded synchronous hold; the durable pending row outlives this. */
   holdSeconds: number;
   /** Bare hostnames of platform-provided upstreams (the object store) that
@@ -133,6 +142,18 @@ export function createExtAuthzGate(deps: CreateExtAuthzGateDeps): ExtAuthzGate {
         return matched.verdict;
       }
 
+      // Holding is only worth anything if a human can answer it. A turn driven
+      // from Slack or Telegram can't produce a verdict — the messenger offers
+      // the owner no safe way to decide, and the conversation's other members
+      // aren't the owner — so such a turn would stall for the whole window and
+      // be denied anyway. Refuse it now instead, and leave the row below for
+      // the inbox: a permanent approval there is what the agent's next attempt
+      // consumes. An attached browser or CLI session means someone *can*
+      // decide, so the hold stands even while a channel turn runs alongside.
+      const unattended =
+        (await deps.attendance.hasOpenChannelTurn(identity.agentId)) &&
+        !(await deps.attendance.hasInteractiveSession(identity.agentId));
+
       // Dedupe retried holds: when the agent's CLI retries (Envoy timeout,
       // network blip, api-server restart mid-hold) we want one inbox row
       // per logical decision, not one per retry. Reuse any active pending
@@ -147,6 +168,10 @@ export function createExtAuthzGate(deps: CreateExtAuthzGateDeps): ExtAuthzGate {
       });
       const pendingId = existing?.id ?? randomUUID();
       if (!existing) {
+        // Recorded pending, not expired, on both paths: the row is what the
+        // owner acts on later, and keeping it active is also what makes the
+        // agent's retries reuse one inbox entry instead of filing a new one
+        // each time.
         await deps.repo.insertPending({
           id: pendingId,
           type: "ext_authz",
@@ -156,29 +181,53 @@ export function createExtAuthzGate(deps: CreateExtAuthzGateDeps): ExtAuthzGate {
           payload: { kind: "ext_authz", host, method, path, viaAgentId: via },
           expiresAt: new Date(Date.now() + deps.holdSeconds * 1000),
         });
-        const frame = buildExtAuthzSynthFrame({
-          approvalId: pendingId,
-          host,
-          method,
-          path,
-        });
-        // The prompt surfaces on the policy-bearing agent's channel — under
-        // aliasing that is the driver, whose inbox tray owns the decision.
-        void deps.bus.publish(injectChannelOf(identity.agentId), frame);
-        // Agent egress blocked awaiting a human verdict. correlationId ties
-        // this to the verdict line written when the hold settles (and to the
-        // approval.verdict line in approvals-service).
-        securityLog("warn", "egress.hold", {
+        if (!unattended) {
+          const frame = buildExtAuthzSynthFrame({
+            approvalId: pendingId,
+            host,
+            method,
+            path,
+          });
+          // The prompt surfaces on the policy-bearing agent's channel — under
+          // aliasing that is the driver, whose inbox tray owns the decision.
+          void deps.bus.publish(injectChannelOf(identity.agentId), frame);
+          // Agent egress blocked awaiting a human verdict. correlationId ties
+          // this to the verdict line written when the hold settles (and to the
+          // approval.verdict line in approvals-service).
+          securityLog("warn", "egress.hold", {
+            category: "egress",
+            actor: identity.ownerSub,
+            actorKind: "agent",
+            surface: "ext-authz",
+            agentId: identity.agentId,
+            target: host,
+            decision: "hold",
+            correlationId: pendingId,
+            detail: { method, path, via },
+          });
+        }
+        // Unattended: no in-session prompt is published. The only subscribers
+        // are the relay clients whose absence defines this path, and an
+        // unconsumed frame would leave the inbox row looking answerable
+        // in-session when the request has already been refused.
+      }
+
+      if (unattended) {
+        // correlationId points at the row an owner can still approve, which is
+        // what makes this deny distinguishable from a rule-based one.
+        securityLog("warn", "egress.decision", {
           category: "egress",
           actor: identity.ownerSub,
           actorKind: "agent",
           surface: "ext-authz",
           agentId: identity.agentId,
           target: host,
-          decision: "hold",
+          decision: "deny",
           correlationId: pendingId,
-          detail: { method, path, via },
+          reason: "unattended-channel-turn",
+          detail: { method, path, basis: "unattended", via },
         });
+        return "deny";
       }
 
       const { verdict, reason } = await waitForVerdict(deps, pendingId);
--- a/packages/api-server/src/modules/channels/infrastructure/slack.ts
+++ b/packages/api-server/src/modules/channels/infrastructure/slack.ts
@@ -1,4 +1,6 @@
 import type { TtlStore } from "../../../core/ttl-store.js";
+import type { ChannelTurnAttendance } from "../../../core/turn-attendance.js";
+import { channelNetworkAccessGuidance } from "./network-access-copy.js";
 import { match, P } from "ts-pattern";
 import {
   ambientThreadKey,
@@ -151,6 +153,7 @@ function slackTurnContract(ctx: {
         ".",
     "If a tool is deferred, load it via ToolSearch first.",
     "</how-to-respond>",
+    channelNetworkAccessGuidance(ctx.brand.name),
   ].join("\n");
 }
 
@@ -684,6 +687,11 @@ export function createSlackWorker(
   isTermsAccepted: (sub: string) => Promise<boolean>,
   uiBaseUrl: string,
   emit: (event: DomainEvent) => void = defaultEmit,
+  /** Marks the agent as driven from a channel for the length of each turn, so
+   *  the egress gate can refuse a request that would otherwise hold for a
+   *  verdict nobody in a Slack conversation can give. Defaults to a no-op for
+   *  tests that don't exercise it. */
+  attendance: ChannelTurnAttendance = { openChannelTurn: () => () => {} },
 ): SlackWorker {
   const brandShort = brand.short;
   let gateway: SlackGateway | null = null;
@@ -698,6 +706,10 @@ export function createSlackWorker(
     threadTs: string;
     eventTs: string;
     sessionId?: string;
+    /** Releases this turn's channel-turn attendance marker; set by
+     *  {@link beginTurn} and paired in {@link endTurn}, so concurrent turns on
+     *  one agent each hold their own and the marker outlives all of them. */
+    releaseAttendance?: () => void;
   };
 
   /** Turns currently driving the harness per agent. A single agent pod
@@ -743,6 +755,7 @@ export function createSlackWorker(
       inFlightTurns.set(instanceName, live);
     }
     live.add(ref);
+    ref.releaseAttendance ??= attendance.openChannelTurn(instanceName);
   }
 
   /** `harnessMayStillRun`: the turn settled on a path that says nothing about
@@ -760,6 +773,11 @@ export function createSlackWorker(
       live.delete(ref);
       if (live.size === 0) inFlightTurns.delete(instanceName);
     }
+    // Released even when the harness may still be running: a late request from
+    // a turn whose relay dropped falls back to the ordinary hold rather than
+    // being refused on the strength of a marker nothing is refreshing.
+    ref.releaseAttendance?.();
+    ref.releaseAttendance = undefined;
     if (opts?.harnessMayStillRun) {
       let lingering = lingeringTurns.get(instanceName);
       if (!lingering) {
--- a/packages/api-server/src/modules/channels/infrastructure/slack.ts
+++ b/packages/api-server/src/modules/channels/infrastructure/slack.ts
@@ -686,12 +686,14 @@ export function createSlackWorker(
   brand: { name: string; short: string },
   isTermsAccepted: (sub: string) => Promise<boolean>,
   uiBaseUrl: string,
-  emit: (event: DomainEvent) => void = defaultEmit,
   /** Marks the agent as driven from a channel for the length of each turn, so
    *  the egress gate can refuse a request that would otherwise hold for a
-   *  verdict nobody in a Slack conversation can give. Defaults to a no-op for
-   *  tests that don't exercise it. */
-  attendance: ChannelTurnAttendance = { openChannelTurn: () => () => {} },
+   *  verdict nobody in a Slack conversation can give. Required: a missing
+   *  wiring here would silently restore the stall this exists to prevent, so
+   *  whether the marker does anything is the store's own business, decided
+   *  where it is built. */
+  attendance: ChannelTurnAttendance,
+  emit: (event: DomainEvent) => void = defaultEmit,
 ): SlackWorker {
   const brandShort = brand.short;
   let gateway: SlackGateway | null = null;
--- a/packages/api-server/src/modules/channels/infrastructure/telegram.ts
+++ b/packages/api-server/src/modules/channels/infrastructure/telegram.ts
@@ -375,22 +375,19 @@ export function createTelegramWorker(deps: {
    *  surface; from the brand config, matching the Slack worker. */
   brandShort: string;
   /** Display brand name, used where the agent is told where its owner acts
-   *  (network-access guidance). Falls back to `brandShort` when unset. */
-  brandName?: string;
+   *  (network-access guidance). */
+  brandName: string;
   emit?: (event: DomainEvent) => void;
   /** Marks the agent as channel-driven for the length of each turn, so the
    *  egress gate can refuse a request no Telegram participant could approve.
-   *  Defaults to a no-op for tests that don't exercise it. */
-  attendance?: ChannelTurnAttendance;
+   *  Required: a missing wiring here would silently restore the stall this
+   *  exists to prevent. */
+  attendance: ChannelTurnAttendance;
   /** Test seam; defaults to the Bot API getChatMember check. */
   isChatAdmin?: (chatId: string, userId: string) => Promise<boolean>;
 }): TelegramWorker {
   const emit = deps.emit ?? defaultEmit;
-  const attendance: ChannelTurnAttendance = deps.attendance ?? {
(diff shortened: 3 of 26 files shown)


packages/api-server/src/modules/approvals/services/ext-authz-gate.ts
```diff
@@ -156,29 +181,53 @@ export function createExtAuthzGate(deps: CreateExtAuthzGateDeps): ExtAuthzGate {
           payload: { kind: "ext_authz", host, method, path, viaAgentId: via },
           expiresAt: new Date(Date.now() + deps.holdSeconds * 1000),
```
dam-code-guardian, line 182: 🟡 **Warning:** The row is inserted with `expiresAt = now + holdSeconds` (30 min by default), the same window a live hold uses. The delivery-sweeper's `expireOverdue` then flips it to `expired`, at which point `listPendingForOwner`/`listPendingForInstance` (which filter `status = 'pending'`) drop it and `resolvePending` returns NOT_ACTIONABLE — the owner can no longer approve it or write the permanent rule from it. This feature's whole premise is that nobody is present to answer synchronously, and the agent is now told "don't retry the same host in a loop", so the common path is: refused → owner returns after 30 min → the recorded request has already expired. Consider a longer (or no) expiry for the unattended row, or a rule-write path that does not depend on the row still being pending.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `3d7a9c9`

## PR #3229: fix(approvals,channels): refuse unattended egress instead of stalling a channel turn
**Author:** pilartomas | **Branch:** fix/channel-egress-approval-fast-deny → main | **Changes:** +745 −21 (14 files)

### Summary
When a channel turn (Slack/Telegram) hits an unallowed host and no interactive session is attached, the egress gate now records the pending request and denies at once instead of holding for the full window. A new cross-replica `turn-attendance` store tracks open channel turns; both attendance reads fail toward "attended". Channel prompts gain shared guidance explaining an unapprovable host.

### Findings
- 🟡 **Warning:** the recorded inbox row keeps the 30-minute hold `expiresAt`, so it becomes non-actionable ~30 min after refusal — undercutting "stays actionable / approve it later" in the very scenario the feature targets (`packages/api-server/src/modules/approvals/services/ext-authz-gate.ts:182`)
- ✅ **Looks good:** `SESSION_PRESENCE_KEY_PREFIX` extraction keeps the exact `presence:agent:` value, so existing presence keys are not orphaned; both attendance reads fail toward "attended", degrading a Redis blip to the ordinary hold.

Verdict: aligned. The behavior change is documented in `security-and-credentials.md` (the "Unattended requests are refused" section) and cross-linked from `channels.md`; the added prose introduces no new file paths, package names, or ADR references. Note: both edited pages carry `Last verified: 2026-08-07`, which predates the rebased HEAD (2026-08-10) — worth a re-verify glance, not drift against this diff's code.

### Issue Fit
Linked issue: #3228 "Agents in Slack and Telegram stall for half an hour when they need network approval" (closing keyword `Closes`).
The immediate-refusal and agent-guidance asks are delivered in full. The third ask — "the request still reaches the owner's inbox, so they can approve it **later**" — is only partially met: the row is inserted, but it keeps the 30-minute hold expiry (see the Findings warning), so an owner returning after that window finds nothing to approve. This is the same expiry the issue's own "Actual Behavior" step 5 calls out. Since `Closes #3228` will resolve the issue on merge, weigh whether "later" needs a longer-lived row.

✅ No new architecture violations (files=12). The gate depends on an injected `EgressAttendance` port; the Redis adapter (`core/turn-attendance.ts`) is wired at the composition root (`index.ts`); the shared key prefix is extracted to one place so the presence writer and the gate reader cannot drift. Layering and dependency direction hold.

### Verdict
COMMENT — a clean, well-tested safety valve. The one warning is a design gap worth resolving before merge given the issue's "approve later" acceptance criterion, not a code defect.

---

packages/api-server/src/modules/approvals/services/ext-authz-gate.ts
pilartomas: I checked this against the code and I don't think it holds — the expired row stays fully actionable for the verdicts that matter. Three points:

**1. The permanent verdicts accept an expired row.** `approvePermanent`, `approveHost` and `denyForever` guard on `status === "resolved"`, not `status !== "pending"`, so an `expired` row passes. When the CAS on the pending row loses (because it expired), each one calls `resolveExpired(...)` and returns `rule_written_expired` — **the egress rule is still written**. That's what PR #2212 put in. Only `approveOnce` and the once-scoped dismiss require `pending`, which is right: releasing a single in-flight call is meaningless once that call is gone.

**2. The inbox doesn't filter it out.** The status filter in `listPendingForOwner`/`listPendingForInstance` is applied only when the caller passes one. The UI calls `api.approvals.listForOwner.query()` with no arguments, so no filter is applied and expired rows are returned.

**3. The UI already renders this exact case.** `approvals-list.tsx` has copy specifically for it — *"The original request already failed. Allow permanently writes a rule that future retries match."* — and on an expired row **Allow permanently**, **Allow \<host\>** and **Deny forever** are all enabled (`disabled={inflight}`). Only **Dismiss** is disabled, with the tooltip *"Original request already failed; nothing to dismiss."*

So an owner returning after the window sees the request, clicks Allow permanently, and the rule is written for the agent's next attempt — which is exactly what #3228 asks for.

Keeping the same `expiresAt` is also deliberate rather than incidental. The row becoming terminal is *correct* for this path: unlike a hold, the request really has already failed, and that terminal state is what makes the UI show the copy above and disable the actions that would do nothing. And keeping it pending until then is what makes `findActivePendingExtAuthz` collapse the agent's retries onto one inbox entry instead of filing a new row per attempt.

There is one real wrinkle in the neighbourhood, though it's not the one above: for the first 30 minutes the row is `pending` with an unexpired `expiresAt`, so the UI's `live` check treats it as releasable and leaves **Dismiss** / allow-once enabled — clicking them resolves the row but resumes nothing, because no call is waiting. I left that alone deliberately: the alternative (inserting the row already expired) would break retry dedup and fill the inbox with one row per attempt. It's also not new — a row whose waiter died in an api-server restart mid-hold behaves the same way.

Happy to be shown wrong if I've misread a path.

packages/api-server/src/modules/channels/infrastructure/slack.ts
```diff
@@ -684,6 +687,11 @@ export function createSlackWorker(
   isTermsAccepted: (sub: string) => Promise<boolean>,
   uiBaseUrl: string,
   emit: (event: DomainEvent) => void = defaultEmit,
+  /** Marks the agent as driven from a channel for the length of each turn, so
+   *  the egress gate can refuse a request that would otherwise hold for a
+   *  verdict nobody in a Slack conversation can give. Defaults to a no-op for
+   *  tests that don't exercise it. */
+  attendance: ChannelTurnAttendance = { openChannelTurn: () => () => {} },
```
dam-code-guardian: 🟡 **Warning:** `attendance` is optional here with a no-op default (`= { openChannelTurn: () => () => {} }`), justified for tests. Per **Required Dependencies — Never Optional** (`infrastructure.md`), no-op-vs-real is a behavior toggle, not a dep-presence question: if a future edit drops this arg at the composition root, TypeScript stays silent, the marker is never written, `hasOpenChannelTurn` stays false on every replica, and the gate reverts to holding for the full window — the exact stall this PR removes, with no failing test. Make it required and pass a stub in the tests that don't exercise it.

packages/api-server/src/modules/channels/infrastructure/telegram.ts
```diff
@@ -372,11 +374,22 @@ export function createTelegramWorker(deps: {
...
    *  surface; from the brand config, matching the Slack worker. */
   brandShort: string;
+  /** Display brand name, used where the agent is told where its owner acts
+   *  (network-access guidance). Falls back to `brandShort` when unset. */
+  brandName?: string;
   emit?: (event: DomainEvent) => void;
+  /** Marks the agent as channel-driven for the length of each turn, so the
+   *  egress gate can refuse a request no Telegram participant could approve.
+   *  Defaults to a no-op for tests that don't exercise it. */
+  attendance?: ChannelTurnAttendance;
```
dam-code-guardian: 🟡 **Warning:** same optional-with-no-op-default pattern as the Slack worker (`attendance?` at 384 + `deps.attendance ?? { openChannelTurn: () => () => {} }` at 389). A dropped wiring here silently disables the fast-deny for Telegram with no type error. Make it required; pass a stub where tests don't need it.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `9eaf352`

## PR #3229: fix(approvals,channels): refuse unattended egress instead of stalling a channel turn
**Author:** pilartomas | **Branch:** fix/channel-egress-approval-fast-deny → main | **Changes:** +747 −23 (14 files)

_Complete re-review (label `code-guardian-review`)._

### Summary
On a channel turn (Slack/Telegram) with no interactive session attached, the egress gate now records the pending request and denies at once instead of holding for the full window. A new cross-replica `turn-attendance` store tracks open channel turns; both attendance reads fail toward "attended". Both messengers gain shared guidance that explains an unapprovable host.

### Changes since last review
Previous HEAD: `3d7a9c9` (2026-08-10T06:16:59Z) — verdict COMMENT. Only new commit since: the `Last verified` date bump; the code is unchanged.

- ✅ **Withdrawn:** the recorded-row expiry warning. Verified against the code that the expired row stays actionable for the verdicts that matter: `approvePermanent` / `approveHost` / `denyForever` guard on `status === "resolved"` (not `!== "pending"`), so an expired row still writes the egress rule via `resolveExpired` (`rule_written_expired`); `listPendingForOwner` returns expired rows when no status filter is passed (the UI passes none); and `approvals-list.tsx` enables the permanent actions on an expired row. The author's rebuttal holds.
- ✅ **Resolved:** the prior doc-drift note — both edited architecture pages now carry `Last verified: 2026-08-10`.
- 🆕 **New:** the channel workers take `attendance` as an optional dependency with a no-op default (`slack.ts:694`, `telegram.ts:384`).

### Findings
- 🟡 **Warning:** `attendance` is an optional dep with a no-op default on both channel workers (`packages/api-server/src/modules/channels/infrastructure/slack.ts:694`) — see the TypeScript Engineering section.
- 🟢 **Suggestion:** for the first `holdSeconds` a fast-denied row is `pending` with an unexpired `expiresAt`, so the inbox's `live` check leaves **Allow once** / **Dismiss** enabled although no call is waiting; a stray click resolves the row and the agent's next retry files a fresh inbox entry instead of reusing it. The author flagged this deliberately and the alternative (inserting the row already expired) breaks retry dedup — noting only that this path is now common rather than restart-only (`packages/ui/src/modules/approvals/components/approvals-list.tsx:91`).

Verdict: aligned. The behavior change is documented in `security-and-credentials.md` ("Unattended requests are refused, not held") and cross-linked from `channels.md`; both edited pages bumped `Last verified` to 2026-08-10. No ADR leak, no volatile content, no new-subsystem page owed (the store is internal to the egress/approvals subsystem).

### Issue Fit
Linked issue: #3228 "Agents in Slack and Telegram stall for half an hour when they need network approval" (closing keyword `Closes`).
✅ Delivers the issue's ask; no undeclared scope. All three "Expected Behavior" items are met: immediate refusal on a channel turn (allow-all agents unaffected, since rule/platform matches run first); the request still reaches the inbox and a permanent approval writes the rule the next attempt consumes (the "approve later" path — verified actionable after expiry); and both messenger prompts carry advance guidance. The necessary cross-replica machinery is declared in the PR body, not undeclared creep.

🟡 One architecture finding (files=12). The port/adapter shape is otherwise correct: `EgressAttendance` / `ChannelTurnAttendance` are injected interfaces, the concrete `createTurnAttendance` is referenced only at the composition root (`index.ts`), and `SESSION_PRESENCE_KEY_PREFIX` is extracted to one place so the presence writer and gate reader cannot drift. Layering (services → port ← adapter) holds.

- 🟡 **Warning:** the channel workers model `attendance` as **optional with a no-op default** (`slack.ts:694` `= { openChannelTurn: () => () => {} }`; `telegram.ts:384` `attendance?` + `deps.attendance ?? {…}`), justified for tests. This is the "Required Dependencies — Never Optional" anti-pattern (`infrastructure.md`): the on/off axis (no-op vs real) is modeled as a wiring question, so a dropped wiring at the composition root becomes a **silent no-op** — the marker is never written, `hasOpenChannelTurn` stays false on every replica, the gate always holds, and the original 30-minute stall returns with no type error and no failing test. The gate and `compose.ts` correctly require the dep. Fix: make it required and pass a stub in the tests that don't exercise it.

### Verdict
COMMENT — a correct, well-tested safety valve; the "approve later" gap raised last round does not hold. One architecture warning (optional dep hides a wiring regression) and one minor UI wrinkle, neither a correctness or security defect.

---

packages/api-server/src/modules/channels/infrastructure/slack.ts
pilartomas: Agreed, and fixed in `714f27b1` — this one's a fair hit. The failure mode you describe is exactly the one that matters here: a dropped wiring wouldn't break a type or a test, it would just quietly stop marking channel turns, leave the gate holding, and reinstate the stall this PR exists to remove. I traded that for not touching 12 test call sites, which was the wrong trade.

Both workers now take `attendance` as a required dependency and call it unconditionally; whether the marker does anything is the store's own business, decided where it's built. The 12 worker tests that don't exercise it pass an explicit inert stub (`stubTurnAttendance()` in the shared test helpers) rather than relying on a default, so the wiring stays visible at each call site.

I applied the same reasoning to `brandName` on the Telegram worker, which I'd also added as optional-with-fallback in this PR — its silent fallback would have shown the wrong brand in the agent's guidance rather than failing.

899 tests, tsc, lint and format all pass.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `714f27b1`

## PR #3229: fix(approvals,channels): refuse unattended egress instead of stalling a channel turn
**Author:** pilartomas | **Branch:** fix/channel-egress-approval-fast-deny → main | **Changes:** +779 −24 (26 files)

### Summary
On a channel turn (Slack/Telegram) with no interactive session attached, the egress gate records the pending request and denies at once instead of holding for the full window. A cross-replica `turn-attendance` store tracks open channel turns; both attendance reads fail toward "attended". Both messengers gain shared guidance that explains an unapprovable host.

### Changes since last review
Previous HEAD: 9eaf352 (2026-08-10T06:50:38Z) — verdict COMMENT. One new commit: `attendance` made a required dependency on both channel workers.

- ✅ **Fixed:** the optional-with-no-op-default `attendance` warning — both workers now take it as required; `emit` moved after it on the Slack worker, and the call sites plus every worker test pass it explicitly (`stubTurnAttendance()` helper). A dropped wiring is now a type error, not a silent stall.

### Findings
- 🟢 **Suggestion:** for the first `holdSeconds` a fast-denied row is pending and unexpired, so the inbox still enables Allow once / Dismiss though no call waits; a stray click resolves the row and the next retry files a fresh entry (`approvals-list.tsx:91`). Author-disclosed and deliberate; still present, noted only because this path is now common.

Verdict: aligned. The behavior is documented in `security-and-credentials.md` and cross-linked from `channels.md`; both pages carry `Last verified: 2026-08-10`. No ADR leak, no volatile content added, no new-subsystem page owed.

### Issue Fit
Linked issue: #3228 (closing keyword `Closes`). All three Expected-Behavior items are delivered: immediate refusal on a channel turn (allow-everything agents unaffected), the pending row stays actionable in the inbox for a later permanent approval, and the shared guidance tells the agent how to explain a refusal. Out-of-scope scheduled runs are declared, not silent. No undeclared scope creep.

✅ No architecture violations (files=24). The `EgressAttendance` port is declared by the gate service; the single `createTurnAttendance` adapter lives in `core/` and is wired at the composition root to the gate and both workers; `SESSION_PRESENCE_KEY_PREFIX` is extracted to one place so the presence writer and the gate reader cannot drift. The prior Required-Dependencies-Never-Optional finding is resolved — `attendance` is required on both workers, and the composition root supplies it.

### Verdict
APPROVE — correct, well-tested safety valve; the one prior architecture warning is fixed and no correctness or security defect remains. The single 🟢 is a deliberate, author-disclosed UI wrinkle.

---

PR #3230, opened 2026-08-07
feat(connections): edit a GitHub App connection's repositories and permissions

pilartomas:
Follow-up to #3225. Narrowing was fixed at create, so adding a repository meant rebuilding the connection — re-pasting the private key and re-granting it to every agent using it.

## What changed

A GitHub App connection's repositories and permissions are now editable in place, from the connection row's ⋮ menu.

This is the **one** part of a connection's configuration that is editable, and it stays cheap because nothing else moves. The contributions are unchanged — same hosts, same injections — so there's no Agent-spec patch and no pod roll. The narrower token swaps in gateway-side via SDS, and every existing agent grant keeps working; the agents just start using a smaller token.

Structurally it's the operation the private-key rotation already performs: mint, write token + SDS to the same Secret, update stored auth.

**Editing never asks for the key again.** The connection already holds its app identity and private key, so a second probe entry point re-reads the installation on the user's behalf.

**The new scope is proven before it's stored,** by minting against it — same fail-closed property as create. A subset the installation can't cover fails the edit you're looking at, rather than parking the connection an hour later.

**Take effect is immediate.** The proving mint's token replaces the live one, so the narrowing applies now rather than at the next renewal.

## Notes for the reviewer

- **Replace, not merge.** The caller states the whole scope; an omitted half clears. Clearing both is a deliberate widening back to the full installation, and the dialog says so plainly before you save.
- **Name-narrowed connections.** A connection created with repository *names* (CLI `--repositories`, or the create form's typed fallback) has no ids to seed the picker from. Those names are resolved against the installation so an untouched Save keeps the same repositories rather than clearing them — the obvious silent-widening trap. Names the installation no longer lists are called out as being dropped.
- **When the repository list can't be read,** repositories become read-only and are passed through exactly as stored, so permissions stay editable without the repository half being rewritten from an empty selection.
- **`RepositorySection`/`PermissionSection` moved** to their own module, unchanged, so the create picker and the edit dialog share them.

## One change outside the feature, deliberately

The github-app re-mint in the refresh loop now merges only the horizon and failure marker into stored auth, instead of writing back the whole copy it read at the start of the tick. With the scope mutable, a whole-object write let a refresh that began before an edit overwrite it afterwards — silently restoring authority the user had just removed. It uses the same `jsonb_set` shape the failure-marker write in that file already uses. The three tests that asserted the old whole-object write now assert the anti-clobber property instead: the write names the horizon and never names the scope.

## Verification

- `api-server` 873 tests, `ui` 256, `cli` 154 — all pass. Typecheck, lint, and format verified per package directly.
- Reviewed adversarially before commit: 17 candidate findings, 12 confirmed after verification (several duplicates of the same root), all fixed. The ones worth naming:
  - **The catalogue never mounted the dialog**, so the new menu item did nothing there — and the un-cleared state then shadowed the next credential dialog, rendering the scope editor bound to a *different* connection than the row the user clicked. Fixed at the guard and at the root, by making the two dialog slots mutually exclusive.
  - **The widening trap above**, where the dialog claimed the selection was preserved while Save posted an empty id set.
  - **A `tsc` failure I'd have shipped**: I ran the type check before adding my last test, and vitest doesn't typecheck. Caught here, fixed, and re-verified after.
  - An impure `setState` updater that duplicated names under StrictMode's double invoke.

### DIFF
--- a/packages/api-server/src/modules/connections/services/oauth-refresh.ts
+++ b/packages/api-server/src/modules/connections/services/oauth-refresh.ts
@@ -29,7 +29,8 @@ import { securityLog } from "../../../core/security-log.js";
 import type { SecretStore } from "../../secret-store/index.js";
 import { mintClientCredentialsToken } from "./client-credentials.js";
 import { refreshOAuthAccessToken } from "./oauth-token.js";
-import { mintGitHubAppToken } from "./github-app.js";
+import { gitHubAppMintLockKey, mintGitHubAppToken } from "./github-app.js";
+import { createXactLock, type XactLock } from "../../../core/xact-lock.js";
 
 export interface OAuthRefreshLoop {
   /** One idempotent refresh pass — scheduled via the shared periodic-jobs
@@ -371,24 +372,28 @@ export async function remintGitHubAppOne(
     githubAppEngine: GitHubAppEngine;
     secretStore: SecretStore;
     db: Db;
+    /** Defaults to the Postgres advisory lock; injectable for tests. */
+    connectionLock?: XactLock;
   },
 ): Promise<void> {
   const privateKeyPem = await deps.secretStore.getField(auth.privateKeyRef);
   if (!privateKeyPem) {
     throw new Error(`private key missing at ${auth.privateKeyRef.path}`);
   }
+  const withLock = deps.connectionLock ?? createXactLock(deps.db);
 
-  // A github-app connection's scope is editable in place, so the copy this
-  // tick read can go stale between reading it and minting from it. Minting
-  // against a stale scope is not a metadata problem — the *token* written below
-  // would carry authority the user has just removed, for as long as it lives.
-  //
-  // So the write is guarded on the horizon this pass minted against: an edit
-  // moves it, the guard matches nothing, and the second pass re-mints from what
-  // the edit stored. Two passes are enough — a third would mean two edits
-  // landing inside one mint.
-  let current = auth;
-  for (let pass = 0; pass < 2; pass++) {
+  // A github-app connection's scope is editable in place, so this sequence
+  // races an edit doing the same thing to the same connection. Guarding the row
+  // alone cannot settle it: the token lives in a store no database transaction
+  // binds, so the two writes can still land in an order that leaves the stored
+  // scope and the live token describing different authority. Both sides take
+  // the same lock instead, so one finishes before the other starts.
+  await withLock(gitHubAppMintLockKey(conn.id), async () => {
+    // Re-read inside the section: the copy this tick carried in may predate an
+    // edit that has since completed, and minting from it would hand back the
+    // authority that edit removed.
+    const current = (await readGitHubAppAuth(deps.db, conn.id)) ?? auth;
+
     const next = await mintGitHubAppToken(deps.githubAppEngine, {
       connectionRef: `connection:${conn.id}:${conn.templateId}`,
       auth: current,
@@ -401,30 +406,15 @@ export async function remintGitHubAppOne(
     });
 
     // Merge the two keys this tick owns rather than writing back the whole
-    // `auth` it read: the keys it does not name cannot be clobbered, so a
-    // concurrent edit keeps its scope. Same shape as the failure-marker write.
-    const result = await deps.db
+    // `auth` it read: the keys it does not name cannot be clobbered.
+    await deps.db
       .update(connectionsTable)
       .set({
         auth: sql`jsonb_set(${connectionsTable.auth}, '{expiresAt}', to_jsonb(${next.expiresAt}::bigint)) - 'refreshFailedAt'`,
         updatedAt: new Date(),
       })
-      .where(
-        and(
-          eq(connectionsTable.id, conn.id),
-          sql`${connectionsTable.auth} ->> 'expiresAt' IS NOT DISTINCT FROM ${current.expiresAt === undefined ? null : String(current.expiresAt)}`,
-        ),
-      );
-    if (((result as unknown as { rowCount?: number }).rowCount ?? 0) > 0)
-      return;
-
-    // The guard matched nothing: something re-minted this connection while we
-    // were in flight, and the token we just wrote may be the wider one. Re-read
-    // and go again so the stored token ends up matching the stored scope.
-    const fresh = await readGitHubAppAuth(deps.db, conn.id);
-    if (!fresh) return;
-    current = fresh;
-  }
+      .where(eq(connectionsTable.id, conn.id));
+  });
 }
 
 /** The connection's current github-app auth, or null if it is gone or is no
--- a/packages/api-server/src/modules/connections/services/oauth-refresh.ts
+++ b/packages/api-server/src/modules/connections/services/oauth-refresh.ts
@@ -378,26 +378,66 @@ export async function remintGitHubAppOne(
     throw new Error(`private key missing at ${auth.privateKeyRef.path}`);
   }
 
-  const next = await mintGitHubAppToken(deps.githubAppEngine, {
-    connectionRef: `connection:${conn.id}:${conn.templateId}`,
-    auth,
-    privateKeyPem,
-  });
+  // A github-app connection's scope is editable in place, so the copy this
+  // tick read can go stale between reading it and minting from it. Minting
+  // against a stale scope is not a metadata problem — the *token* written below
+  // would carry authority the user has just removed, for as long as it lives.
+  //
+  // So the write is guarded on the horizon this pass minted against: an edit
+  // moves it, the guard matches nothing, and the second pass re-mints from what
+  // the edit stored. Two passes are enough — a third would mean two edits
+  // landing inside one mint.
+  let current = auth;
+  for (let pass = 0; pass < 2; pass++) {
+    const next = await mintGitHubAppToken(deps.githubAppEngine, {
+      connectionRef: `connection:${conn.id}:${conn.templateId}`,
+      auth: current,
+      privateKeyPem,
+    });
 
-  await deps.secretStore.putFields(auth.accessTokenRef, {
-    access_token: next.accessToken,
-    ...buildConnectionSdsFields(conn.contributions, next.accessToken),
-  });
-  // Merge the two keys this tick owns rather than writing back the whole `auth`
-  // it read at the start. A github-app connection's scope is editable in place,
-  // so a re-scope that landed while this mint was in flight would otherwise be
-  // overwritten by the pre-edit copy — quietly restoring authority the user had
-  // just taken away. Same shape as the failure-marker write above.
-  await deps.db
-    .update(connectionsTable)
-    .set({
-      auth: sql`jsonb_set(${connectionsTable.auth}, '{expiresAt}', to_jsonb(${next.expiresAt}::bigint)) - 'refreshFailedAt'`,
-      updatedAt: new Date(),
-    })
-    .where(eq(connectionsTable.id, conn.id));
+    await deps.secretStore.putFields(auth.accessTokenRef, {
+      access_token: next.accessToken,
+      ...buildConnectionSdsFields(conn.contributions, next.accessToken),
+    });
+
+    // Merge the two keys this tick owns rather than writing back the whole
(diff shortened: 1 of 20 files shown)


dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `db35fce`

## PR #3230: feat(connections): edit a GitHub App connection's repositories and permissions
**Author:** pilartomas | **Branch:** feat/edit-github-app-scope → main | **Changes:** +887 −127 (17 files)

### Summary
Makes a GitHub App connection's repositories and permissions editable in place: two new tRPC endpoints (probe-for-connection, update-scope), an edit dialog reusing the create picker's sections, and an anti-clobber change to the refresh remint. The feature proves the new scope by minting against it before storing, and swaps the live token so narrowing takes effect at once.

### Findings
- 🟡 **Warning:** the refresh remint's token write is not guarded against a concurrent edit the way its DB write now is (`packages/api-server/src/modules/connections/services/oauth-refresh.ts:388`)
- 🟢 **Suggestion:** name-narrowed repositories beyond the probe's 500-repo page cap are dropped on save and mislabeled as "no longer listed" (`packages/ui/src/modules/connections/components/connection-edit-github-app-scope-dialog.tsx:83`)
- ✅ **Looks good:** the deliberate `jsonb_set` anti-clobber refactor correctly stops an in-flight refresh from restoring stored scope after an edit.

Verdict: aligned. The editable-in-place behavior is documented in `docs/architecture/security-and-credentials.md` (`Last verified: 2026-08-07`, matching the PR commit date); `connections.md` carries no now-stale "scope set only at create" statement; the added prose introduces no new file paths, package names, or ADR references.

### Issue Fit
`#3225` is a PR referenced without a closing keyword, and the branch names no issue.
_No linked issue — scope checked against the PR description only._ The diff delivers every promise in the description (edit dialog, two endpoints, shared `RepositorySection`/`PermissionSection`, and the explicitly-declared refresh-loop anti-clobber change); no undeclared scope creep.

✅ No new architecture violations (files=9). Input validation stays in the contract package (`schemas.ts`), orchestration in the service, and parsing in `domain/github-app-scope.ts`. The remint's raw `jsonb_set` SQL sits in the service rather than behind a repository port, but that matches the file's existing direct-`deps.db` pattern, so it is not newly introduced.

State lineage is correct (local form state in `useState`, installation data in the TanStack Query mutation), and the seed effect's updater is StrictMode-safe — it splits resolved/missing before `setState`. Modeling the probe as a mutation fired from a mount effect matches the existing create-picker pattern (the probe mints and revokes a token, so it is a command, not a query). One finding (files=7):
- 🟢 The name-resolution seed treats names the probe did not return as "no longer listed" and drops them on save, while id-based selections in the same case are preserved via `keptUnlisted`. See the Findings warning above — rare (>500-repo installations), but the message misattributes truncation.

### Verdict
COMMENT — the feature is solid and well-tested; the one warning is a narrow, self-healing race worth a look before merge, not a blocker.

---

packages/api-server/src/modules/connections/services/oauth-refresh.ts
```diff
@@ -388,12 +388,16 @@ export async function remintGitHubAppOne(
     access_token: next.accessToken,
```
dam-code-guardian: 🟡 **Warning:** This tick mints the token from the `conn.auth` snapshot the sweep loaded at its start, then writes it here unconditionally. The new `jsonb_set` below protects the stored *scope* from a concurrent edit, but not this token write: if a user narrows the scope while a refresh for this connection is in flight, this `putFields` overwrites the just-narrowed live token with one minted against the pre-edit (wider) scope. The DB self-heals at the next renewal, but the live token carries the removed authority for up to that renewal (~1h) — defeating the PR's "take effect immediately" property. Consider re-reading the connection's current auth just before minting (or guarding the token write like the DB write), so the sweep mints against the current scope.

packages/ui/src/modules/connections/components/connection-edit-github-app-scope-dialog.tsx
```diff
@@ -0,0 +1,256 @@
...
+          const byName = new Map(
+            result.repositories.map((r) => [r.name, r.id] as const),
+          );
+          // Split before touching state: an updater that appended to `missing`
+          // would build a different list each time React invoked it.
+          const resolved: number[] = [];
+          const missing: string[] = [];
+          for (const name of names) {
+            const id = byName.get(name);
+            if (id === undefined) missing.push(name);
```
dam-code-guardian, line 83: 🟢 **Suggestion:** `byName` is built only from `result.repositories`, which the probe caps at 5 pages × 100 = 500 repos (`repositoriesTruncated`). A name-narrowed connection whose repository sits beyond that cap is pushed to `missing` and dropped on save, even though it is still in the installation — and the callout tells the user it is "no longer listed". Id-based selections in the same situation are preserved via `keptUnlisted`, so the two paths diverge. Rare (>500-repo installations), but consider distinguishing truncated-away names from genuinely-removed ones.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `21997d7`

## PR #3230: feat(connections): edit a GitHub App connection's repositories and permissions
**Author:** pilartomas | **Branch:** feat/edit-github-app-scope → main | **Changes:** +1011 −142 (17 files)

### Summary
Complete re-review at `21997d7`. One new commit since the last review hardens the refresh-loop anti-clobber: the github-app re-mint now re-mints against the edited scope when a refresh overtakes an edit, and the edit dialog now locks the repository half when a truncated installation listing meets a name-narrowed connection.

### Changes since last review
Previous HEAD: `db35fce` (2026-08-10T06:10:46Z) — verdict COMMENT

- ✅ **Fixed:** name-narrowed repositories beyond the listing page cap are no longer dropped or mislabeled — the repository half locks and passes the stored scope through (`connection-edit-github-app-scope-dialog.tsx:130`).
- 🔁 **Still present (narrowed):** the refresh remint's token write to the secret store stays outside the DB compare-and-set, so one interleaving still leaves the wider token live (`oauth-refresh.ts:398`).

### Findings
- 🟡 **Warning:** the two-pass guard re-mints only when the edit's DB write is already visible; the token `putFields` is unconditional and precedes the guard, so if a concurrent edit's `updateAuth` commits *after* the refresh's guarded write while its token write landed *before* it, the wider token stays live until the next tick (`packages/api-server/src/modules/connections/services/oauth-refresh.ts:398`). A DB-only compare-and-set cannot bind the token in the separate secret store; serializing edit and refresh on the connection (a per-connection lock across mint + token write + DB write) would close it.

Verdict: aligned. `docs/architecture/security-and-credentials.md` documents editable-in-place scope, immediate re-mint, prove-before-store, and "nothing else moves" — all matching the code. No ADR reference leak, no volatile content added, no new subsystem. The re-reviewed commit touches no architecture page.

### Issue Fit
_No linked issue — scope checked against the PR description only._ `#3225` is a PR referenced as a follow-up, not a closing keyword; the branch names no issue. The diff delivers every promise (two endpoints, edit dialog, shared `RepositorySection`/`PermissionSection`, and the declared refresh anti-clobber change); the new commit implements the body's "re-mint against the edited scope when a refresh overtakes it". No undeclared scope creep.

✅ No new architecture violations (files=9). New endpoint validation stays in the contract package (`schemas.ts`) with routers delegating to the service under `manageCredentialsProcedure`; `updateGitHubAppScope`/probe go through `deps.repo`, `deps.secretStore`, and the engine ports. The added raw `db.select()`/`jsonb_set` in `oauth-refresh.ts` match that file's pre-existing direct-`deps.db` pattern — not newly introduced.

✅ No new violations (files=7). The shadowing bug is fixed at the state source: `useConnectionMaintenance` gives `openUpdate`/`openEditScope` that each clear the other slot, so at most one dialog mounts. State lineage is correct (server = probe mutation, local form state = `useState`). `RepositorySection`/`PermissionSection` are cleanly extracted as prop-driven presentational components shared by the create form and the edit dialog.

### Verdict
COMMENT — the commit closes the common overtake case and the truncation mislabeling; one narrow, self-healing residual race remains (needs a near-expiry connection and a specific interleaving), worth a look but not a blocker.

---

packages/api-server/src/modules/connections/services/oauth-refresh.ts
```diff
@@ -371,29 +372,62 @@ export async function remintGitHubAppOne(
...
+  await withLock(gitHubAppMintLockKey(conn.id), async () => {
+    // Re-read inside the section: the copy this tick carried in may predate an
+    // edit that has since completed, and minting from it would hand back the
+    // authority that edit removed.
+    const current = (await readGitHubAppAuth(deps.db, conn.id)) ?? auth;
 
-  await deps.secretStore.putFields(auth.accessTokenRef, {
-    access_token: next.accessToken,
-    ...buildConnectionSdsFields(conn.contributions, next.accessToken),
+    const next = await mintGitHubAppToken(deps.githubAppEngine, {
```
dam-code-guardian, line 397: 🟢 **Suggestion:** the mint here is a GitHub HTTP round-trip, and it runs inside the `pg_advisory_xact_lock` transaction. So a pooled connection stays idle-in-transaction for the network call, while the section's own re-read and `jsonb_set` write (`deps.db`, not `tx`) each need a second connection. The sequential refresh sweep bounds this today, but it couples GitHub latency to DB-pool occupancy, and at higher edit/refresh concurrency across many connections it risks pool pressure. Minting outside the lock, then re-checking the scope inside before the writes, would decouple them.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `16b8297`

## PR #3230: feat(connections): edit a GitHub App connection's repositories and permissions
**Author:** pilartomas | **Branch:** feat/edit-github-app-scope → main | **Changes:** +1054 −145 (20 files)

### Summary
Complete re-review at `16b8297`. The new commit replaces the two-pass guarded write with one shared Postgres advisory lock. A scope edit and the refresh loop's re-mint now run their mint → token write → auth write under `gitHubAppMintLockKey(connectionId)`, and the refresh re-reads the scope inside the section. This closes the residual token race the last review flagged.

### Changes since last review
Previous HEAD: `21997d7` (2026-08-10T06:42:20Z) — verdict COMMENT

- ✅ **Fixed:** the refresh remint's token write is now serialized against a concurrent scope edit — both sides take the same `pg_advisory_xact_lock`, so one finishes before the other starts and the stored scope and the live token cannot diverge (`oauth-refresh.ts:391`).
- 🆕 **New:** the mint's GitHub round-trip runs inside the held lock transaction (`oauth-refresh.ts:397`).

### Findings
- 🟢 **Suggestion:** the mint runs inside the held advisory-lock transaction (`packages/api-server/src/modules/connections/services/oauth-refresh.ts:397`).

Verdict: aligned. `docs/architecture/security-and-credentials.md` documents editable-in-place scope, immediate re-mint, prove-before-store, and "nothing else moves" — all matching the code. The new commit changes only the internal concurrency mechanism, not documented behavior, so `Last verified: 2026-08-07` stays valid. No ADR reference leak, no volatile path added, no new subsystem.

### Issue Fit
_No linked issue — scope checked against the PR description only._ `#3225` is a follow-up PR reference, not a closing keyword; the branch names no issue. The diff delivers every promise in the body, and the new commit implements the body's "take effect is immediate" and anti-clobber claim by serializing the two writers. No undeclared scope creep.

✅ No new architecture violations (files=12). The `connectionLock` is injected into the service as a port-like dependency and wired at the composition root (`compose.ts`); `oauth-refresh.ts` constructs `createXactLock(deps.db)` directly, which matches that file's pre-existing direct-`deps.db` SQL pattern rather than introducing a new one. Validation stays in the contract package.

✅ No new violations (files=7). No `.tsx` file changed since the last review at `21997d7`; the shared `RepositorySection`/`PermissionSection` extraction, the StrictMode-safe seed, and the mutually-exclusive dialog slots are unchanged and correct.

### Verdict
APPROVE — the advisory lock closes the residual token race across replicas; the stored scope and the live token can no longer diverge. One non-blocking 🟢 on holding the pool connection across the mint.

---

PR #3241, opened 2026-08-10
feat(ui): make back/forward work on sessions, improve session 404 page

### LINKED ISSUE #2891: UI - Sessions can't be linked to — make individual sessions addressable via URLs
## Problem

When a conversation with an agent is open, the address bar identifies only the agent — not the session. This means:

- **Refreshing the page loses your place.** You land back on the agent's default state and have to re-find the session in the list.
- **Sessions can't be bookmarked** to return to a conversation later.
- **Sessions can't be shared** — there is no URL that points a teammate (or a bug report) at a specific conversation.
- **Browser back/forward don't move between sessions** you've visited.

This was raised as a [High] finding in #793. Session URLs existed before (#47) but the behavior was lost in the agent-screen redesign, so this is a regression.

## Proposed solution

The open session is reflected in the URL. Opening that URL — after a refresh, from a bookmark, or from a link someone sent (given access) — lands directly in that conversation. Switching sessions updates the URL and creates history entries, so back/forward navigate between sessions naturally. A newly started conversation gets its URL as soon as the session exists, so even a fresh chat survives a refresh.

If a URL points at a session that no longer exists, the app falls back to the agent's normal default view.

### Scope

Dev-only — no design work needed; nothing visible changes except the address bar.

### Dependencies

- Extracted from #793 (finding: "Sessions can't be linked to").
- Restores the behavior originally delivered for #47.

kapetr:
Closes #2891.

#3183 already did most of this — it needed session URLs so a Slack thread reply could link back to its conversation. Reload, bookmarks, and shared links work today.

One bullet was left: back/forward didn't move between sessions. The URL was only ever replaced, so visiting five sessions left one history entry. Now picking a session pushes an entry, and back walks the conversations you opened.

That made the failure path easy to hit — a link can name a session that's deleted or isn't yours. So the error card now says something useful instead of relaying the harness error, and it offers **Delete** only when the agent's session list confirms there's a row to delete. Before, it offered to delete sessions that didn't exist.

Note: the issue says a dead link should fall back to the agent's default view. This shows the card instead — landing on a blank chat doesn't tell you the link was bad.

`ui:check` and `ui:test` clean, plus manual testing.


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `6e6f9fd`

## PR #3241: feat(ui): make back/forward work on sessions, improve session 404 page
**Author:** kapetr | **Branch:** feat/session-url-history → main | **Changes:** +229 −81 (7 files)

### Summary
Picking a session now pushes a browser history entry, so back and forward walk the conversations the user opened. The resume-error card is rebuilt: the agent's own session list, not the harness error text, decides the wording, and **Delete** appears only for an orphaned session the list still advertises.

### Findings
- 🟢 **Suggestion:** The PR shows a named error card for a dead link, not the "fall back to the agent's default view" that issue #2891 proposed. The divergence is declared in the PR body and matches `platform-topology.md`. No change needed — see Issue Fit. (`packages/ui/src/modules/sessions/views/chat-view.tsx`)
- ✅ **Looks good:** The `absent` → "unavailable" verdict cannot be produced by a connection failure. `listAgentSessions` throws on a failed connection, the caller catches it as `unknown`, and `resumeFailureKind` never reads `absent` from that path. So a transient outage never tells the user the session is gone.

Verdict: **aligned**. `platform-topology.md:75` already describes both behaviors this PR completes — a session "re-opens itself on a reload or a back step", and a refusal "presented as its own screen naming the reason, not as an empty or perpetually-loading chat". The code now matches that prose; the added history entries, error kinds, and delete affordance sit below architecture altitude. No page edit due.

### Issue Fit
Linked issue: #2891 "UI - Sessions can't be linked to" (closing keyword `Closes #2891`)

The core ask — switching sessions creates history entries so back/forward navigate — is this PR's central delivery; addressing, bookmarks, reload, and fresh-chat linkability landed earlier (#3183).
- 🟢 One declared divergence: the issue's proposed solution says a dead link should "fall back to the agent's normal default view"; the PR shows a named error card instead. The PR body declares this, and `platform-topology.md:75` endorses it ("naming the reason, not an empty chat"). Scope is otherwise complete, no undeclared creep.

✅ No TSEng violations introduced. The new pure helpers `nextChatUrl` and `resumeFailureKind` are kept free of DOM and store imports, so they unit-test in a node environment; DOM side effects stay in the hook layer. The cross-file `../../acp/errors.js` import follows the package's existing convention.

✅ No React-standards violations. The change is textbook URL-owned state — back restores the session — with no cross-lineage duplication. `useCallback` dependencies are correct, `RESUME_FAILURE_COPY` is module-scoped, and the failure-kind `Record` is exhaustive. Removing the in-card "Back to sessions" button strands no one: the mobile header keeps its Back control and the desktop sidebar stays visible.

### Verdict
APPROVE — Delivers the issue's core ask, the resume-failure logic is correct and defensive, and the one divergence from the issue is declared and consistent with the architecture doc.

---

packages/ui/src/modules/sessions/views/chat-view.tsx
```diff
@@ -632,15 +655,13 @@ export function ChatView() {
...
                       <SessionErrorCard
                         error={sessionError}
-                        onBack={() => {
-                          setSessionError(null);
-                          resetSession();
-                          if (isMobile()) setMobileScreen("sessions");
-                        }}
                         onDelete={async () => {
-                          const sid = sessionError.sessionId;
+                          // Hold the card until the delete lands: clearing
```
PetrBulanek, line 640: Removing the "Back to sessions" button takes away the only control that cleared
`sessionError`. Nothing else on the new-chat path clears it — `resetSession()`
only touches the connection, the session id, and the messages — so starting a
new chat leaves the card on screen above an empty conversation, and it stays
there once the first messages arrive. I reproduced this against a dev cluster:
land on a link to a deleted session, pick "New chat session", and the URL
becomes `/chat/<agent>` while "Can't open this conversation" is still rendered
(the "Start a conversation" placeholder is suppressed because `sessionError` is
still set).
Three of the four kinds — `unavailable`, `connection`, `other` — render no
button at all, so on desktop the card can only be cleared by opening a different
session, pressing browser Back, or leaving the agent. Whatever shape the fix
takes, leaving the failed session needs to clear the failure too.

packages/ui/src/modules/sessions/views/chat-view.tsx
```diff
@@ -830,21 +851,43 @@ function ChatHeaderStatus({
...
+    title: "Can't open this conversation",
+    body: "It may have been deleted, or the link may point somewhere you can't open.",
+  },
+  orphaned: {
+    title: "This conversation can't be reopened",
+    body: "The agent still lists it but no longer holds its history. Deleting it clears the leftover entry from the list.",
+  },
+  connection: {
+    title: "Can't reach the agent",
+    body: "The agent didn't answer. It may be waking up or hibernating — try again in a moment.",
```
PetrBulanek, line 862: This copy tells the user to try again, but there is no gesture that retries.
`resumeSession` sets `sessionId` to the session it is opening before it awaits,
and the failure path never clears it — so after a connection failure the store
still holds the failed id. Clicking the same row in the sidebar then reaches
`if (sid === sessionId)` in `mobileResumeSession`, which scrolls to the bottom
and returns without resuming.
The old card's "Back to sessions" button reset the session id, so re-picking the
row afterwards did resume. With that button gone, the only ways out of a
transient agent outage are opening a different session first or reloading the
page. Either the card needs something that actually retries, or the copy should
not promise one.

PetrBulanek (changes requested):

PetrBulanek (approved):

PR #3248, opened 2026-08-10
fix(ui): add read-only reason dialog to the schedule edit path

### LINKED ISSUE #3217: UI - Disabled "Edit schedule" does not tell the user why
## Context

A user opened the overflow menu on a schedule in an agent's Schedules section. "Edit schedule" was greyed out with no visible reason. She assumed schedules cannot be edited at all, and asked whether it was a bug. An explanation does exist, but it took a deliberate second attempt to surface it.

Image

## Problem

Two kinds of schedule are read-only in the UI: schedules created by the agent itself, and legacy cron schedules. Everything else is editable. Nothing on screen says which case applies, or that a rule exists at all.

## Expected Behavior

The user understands, without hovering or guessing, why editing is unavailable and what to do instead. The reason names the one case that applies, not both.

## Actual Behavior

"Edit schedule" is greyed out and silent. A hover tooltip exists — "Cron and agent-created schedules can't be edited here." — but it is unreliable and easy to miss: disabled menu items swallow pointer events, so it often does not appear, and when it does it renders offset to the side, overlapping "Delete schedule". Keyboard and screen-reader users get no explanation at all. The message also names both reasons at once, so the user still cannot tell which schedule they have.

## Steps to Reproduce

1. Ask an agent to create a schedule for itself.
2. Open the agent's config page, Schedules section.
3. Open the overflow menu on that schedule.
4. "Edit schedule" is disabled with no visible reason.

kapetr:
Closes #3217.

"Edit schedule" was greyed out with no reason on screen. A tooltip existed, but it sat on a span wrapping the disabled item — disabled items swallow the pointer, so it often didn't show, and keyboard and screen-reader users never got it. It also named both read-only reasons at once, so you still couldn't tell which schedule you had.

The item is now always clickable. It either opens the form, or a dialog with the one reason that applies:

- **Agent-created** — "This schedule is managed by the sandbox", with an Open chat button. Copy follows Jenna's mockup.
- **Cron** — the editor only understands the newer format, and opening it would rewrite the schedule's timing instead of showing it. So it says that, and says to delete and recreate instead.

Agent-created wins when both apply, because agents can create either format and asking in chat is the way out either way.

`ui:check` and `ui:test` clean, verified manually.

### DIFF
--- a/packages/ui/src/components/ui/confirm-dialog.tsx
+++ b/packages/ui/src/components/ui/confirm-dialog.tsx
@@ -1,4 +1,9 @@
-import { TrashCan, Warning, WarningAlt } from "@carbon/icons-react";
+import {
+  Information,
+  TrashCan,
+  Warning,
+  WarningAlt,
+} from "@carbon/icons-react";
 import type { ReactNode } from "react";
 
 import {
@@ -14,7 +19,13 @@ import {
 import { buttonVariants } from "@/components/ui/button";
 import { cn } from "@/lib/utils";
 
-export type ConfirmDialogKind = "default" | "destructive";
+export type ConfirmDialogKind = "default" | "destructive" | "info";
+
+const KIND_ICON = {
+  default: Warning,
+  destructive: WarningAlt,
+  info: Information,
+} as const satisfies Record<ConfirmDialogKind, typeof Warning>;
 
 export interface ConfirmDialogProps {
   open: boolean;
@@ -46,6 +57,7 @@ export function ConfirmDialog({
   onCancel,
 }: ConfirmDialogProps) {
   const destructive = kind === "destructive";
+  const Icon = KIND_ICON[kind];
   const resolvedConfirmLabel =
     confirmLabel ?? (showCancel ? (destructive ? "Remove" : "Confirm") : "OK");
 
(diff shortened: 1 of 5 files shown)


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `fbcc4fe`

## PR #3248: fix(ui): say why a schedule can't be edited
**Author:** kapetr | **Branch:** fix/schedule-edit-reason → main | **Changes:** +135 −24 (5 files)

### Summary
Replaces the disabled "Edit schedule" menu item (whose tooltip was unreliable and invisible to keyboard/SR users) with an always-clickable item that opens a dialog naming the single applicable lock reason — agent-managed (with "Open chat") or legacy-cron (delete-and-recreate).

### Findings
- ✅ **Looks good:** New `showAlert`/`ConfirmDialog` `info` options flow correctly through `DialogOverlay` — alerts render one "Close" button with the info icon (`packages/ui/src/components/dialog-overlay.tsx:16`).

**Verdict:** aligned. The change is UI-presentation only; the editability rule it surfaces is pre-existing. No `docs/architecture/` page changes behavior. No update required.

### Issue Fit
Linked issue: #3217 "UI - Disabled 'Edit schedule' does not tell the user why" (closing keyword `Closes #3217`)

✅ Delivers the issue's ask (accessible, single-reason explanation with a next step, per the mockup in the issue thread); no undeclared scope.

✅ No findings. `AlertOptions = Omit<ConfirmOptions, "cancelLabel">` derives cleanly; `scheduleLock` is a pure discriminated-union helper.

✅ No findings. `KIND_ICON` uses `satisfies Record<ConfirmDialogKind,…>` (compile-safe); `handleEdit` is single-purpose; store selectors follow correct state lineage.

### Verdict
APPROVE — a clean, well-scoped accessibility fix that fully delivers #3217 with no correctness, type, or architecture issues.

---

PetrBulanek (approved):

PR #3251, opened 2026-08-10
feat(ui): open sandbox chat from the list and move launch options into the new-session state

### LINKED ISSUE #3070: UI - No visible way to reach sandbox configuration from chat
## Context

From a chat session, configuration lives behind a kebab menu next to the sandbox name. On desktop that kebab is invisible until you hover the name. On mobile it is always shown.

## Problem

On desktop there is nothing to indicate the menu exists. Reaching configuration requires hovering a spot that gives no sign it is interactive, so the only ways to find it are accident or being told.

That makes a routine task hard to reach. Changing the model, size, connections, or skills all live behind it, and none of that is unusual to want while working in a session.

The other route to the same page has its own problem. The model indicator in the composer navigates there too, but looks like a local control (#3057). So configuration currently has one entrance that is hidden and one that is misleading.

## Goal

Getting from a chat session to that sandbox's configuration should be visible and obvious, without hovering to discover it.

## User Stories

- As someone in a session, I want a visible way to reach configuration so that I do not have to discover it by accident.
- As someone changing a setting mid-session, I want the route to be predictable so that I am not hunting for it.

## Open Questions

- Should the kebab simply always be visible, or does configuration deserve a clearer entry point of its own?

### LINKED ISSUE #3071: UI - Opening a sandbox in chat takes too many clicks
## Context

Reaching chat from the sandbox list takes three steps: click the sandbox, which lands on its home page, open the "Open in" menu, then choose Chat (browser). The menu also offers browser terminal, local terminal, and VS Code / Zed.

## Problem

Chat is the main way people work with a sandbox, but it is treated as one option among four, behind a menu, on a page you have to pass through first. The most common thing you want is the same number of clicks as the rarest.

Clicking a sandbox does not do what it looks like it does. It opens a page about the sandbox rather than opening the sandbox, so the obvious action leads somewhere you did not intend.

## Goal

Opening a sandbox in chat should be a single action from the list. Once in chat, the other ways in should still be available.

## User Stories

- As someone returning to a sandbox, I want one click to land me in chat so that I can pick up where I left off.
- As someone who wants a terminal or an editor instead, I want those still reachable from chat so that nothing is lost by making chat the default.

## Open Questions

- Should the row click open chat directly, or should sandbox home stay reachable another way?
- Does this conflict with #2124, which consolidated launch options into one place?

kapetr:
Clicking a sandbox opens it. One click from the list lands in chat, and a sleeping sandbox starts on its own while the chat comes up — no config page to pass through, no menu to open.

The other ways in are not lost. Browser terminal, local terminal, and local editor move into the new-session state, next to the message box, so making chat the default costs nothing. `+ New` opens a chat session directly instead of asking which kind you want.

Configuration follows from that. The chat header's ⋮ no longer hides until you hover it, and the list rows carry a "Configure sandbox" item, so the page a row click used to open is still one click away. Both config pages get a matching "Open chat" button.

Closes #3071
Closes #3070

## Sub-issues

- [x] 01 — Sandbox rows open chat and start a sleeping sandbox
- [x] 02 — Launch options in the new-session state
- [x] 03 — `+ New` opens a chat session directly
- [x] 04 — Configuration is visible in chat and reachable from the list


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `7afbedf`

## PR #3251: feat(ui): open sandbox chat from the list and move launch options into the new-session state
**Author:** kapetr | **Branch:** feat/3071-chat-landing → main | **Changes:** +290 −258 (24 files)

### Summary
A sandbox row now opens chat directly and auto-wakes a hibernated agent once on entry. Launch options (browser/local terminal, local editor) move into the empty new-session state, `+ New` opens a chat session, and configuration is reachable from a visible kebab item and a header "Open chat" button.

### Findings
- ✅ **Looks good:** `useAutoWakeOnOpen` gates on `!agent.overBudget`, mirroring the established budget-gate skip in `store.ts` — a hibernated but over-budget agent is not auto-woken (`use-auto-wake-on-open.ts:18`).

Verdict: aligned. The change is UI-frontend navigation; no `docs/architecture/` page was edited, and the platform-side wake primitive in `agent-lifecycle.md` is unchanged (the UI invokes the existing explicit-wake path).

### Issue Fit
Linked issues: #3071 "Opening a sandbox in chat takes too many clicks" (closing keyword) · #3070 "No visible way to reach sandbox configuration from chat" (closing keyword)

✅ Delivers both asks — row click opens chat and auto-starts (consistent across sandboxes, experiments, and knowledge bases via the shared `ChatView`), the three launch buttons in the new-session state, a single "+ New" chat button, an always-visible kebab, and a "Configure sandbox" row action. No undeclared substantial scope.

✅ No architecture findings. The changed `.ts` files are client-side UI (zustand slices, hooks) and e2e helpers, not server modules; no new cross-module import or layering violation.

✅ No findings. State lineage is correct (server list via TanStack Query, `decided` set in a ref, view state in zustand — no duplication); the two dialogs were extracted from the deleted `open-in-menu.tsx` into `open-in-dialogs.tsx` (DRY); `useWakeAgent` is memoized with `useCallback`. The `divided=true` default is a deliberate dialog-chrome standardization.

### Verdict
APPROVE — a clean UI refactor that delivers both linked issues; behavior verified against the full files, no correctness or architecture concerns.

---

packages/ui/src/modules/agents/hooks/use-auto-wake-on-open.ts
```diff
@@ -0,0 +1,20 @@
...
+  // Decided on the first poll that knows the agent, so a pause landing later
+  // from another tab or the CLI can't be undone by this open one.
+  const decided = useRef(new Set<string>());
+
+  useEffect(() => {
+    if (!agentId || decided.current.has(agentId)) return;
+    const agent = agents.find((a) => a.id === agentId);
+    if (!agent) return;
+    decided.current.add(agentId);
+    if (agent.state === "hibernated" && !agent.overBudget) wake(agentId);
```
PetrBulanek: Opening a chat calls the wake mutation for any sandbox in the `hibernated` state, and a hard-stopped sandbox is reported as exactly that — `AgentView` carries no marker for the stop, so this cannot see the difference. The result is that a single click on the row restarts a sandbox the user deliberately stopped: I stopped one on a dev cluster, confirmed the "It stays stopped until you start it" dialog, and clicking its row brought the pod back up within half a minute.

That makes "Pause — wakes on next use" and "Stop — until started again" behave identically for the interaction this PR makes primary, and the stop confirmation's wording no longer holds. It also gives back the reserved compute the user stopped the sandbox to free, without asking. The `overBudget` guard right here shows the intent — an over-budget sandbox is deliberately left for the user to start — and a stop is the same kind of "the user said no" state, just one this data can't express yet.

packages/ui/src/modules/agents/views/list-view.tsx
```diff
@@ -96,16 +102,31 @@ export function ListView() {
...
+              <AgentRow
+                key={agent.id}
+                {...rowProps(agent)}
+                temporaryDraw={drawByDriver.get(agent.id)}
+                onSelect={() =>
+                  kb ? openKnowledgeBase(agent.id) : selectAgent(agent.id)
+                }
+                onConfigure={() =>
+                  kb
+                    ? navigateToKnowledgeBaseConfig(agent.id)
```
PetrBulanek: A knowledge base is an agent with a Kind marker, and everything except its create path is plain sandbox machinery — lifecycle, sessions, connections, schedules, budgets. `KnowledgeBaseConfigView` deliberately exposes only name, harness, network, and connections; `SandboxHomeView` is where Skills, Schedules, Channels, Artifacts, and Usage live.

Before this change the Sandboxes list row was the route to sandbox home for every agent, knowledge bases included. Now a knowledge-base row goes to the lean page from both the row click and the Configure item, and the knowledge-bases list and the knowledge-base chat header already went there too — so after this PR nothing in the UI reaches a knowledge base's schedules or skills at all. Worth deciding whether those sections should be reachable for a knowledge base, and if so from where; if the answer is that they intentionally aren't, that is a bigger call than this PR states.

PetrBulanek (changes requested):

packages/api-server-api/src/modules/agents/types.ts
```diff
@@ -75,6 +75,7 @@ export interface Agent {
   effectiveHibernationTimeoutMin: number;
   /** Latest controller-reported error, if any. */
   error?: string;
+  stopRequested: boolean;
```
PetrBulanek, line 78: Making `stopRequested` a required field means every object literal typed as `Agent` has to set it. Six fixtures were updated, but two in the CLI package were missed, so `cli:check:tsc` fails and `mise check` is red on this head:

- `packages/cli/src/__tests__/agent-get.integration.test.ts:143`
- `packages/cli/src/__tests__/agent-list.integration.test.ts:159`

Both are `makeAgent` helpers that already set `overBudget: false` right where the new field belongs. I ran the `ui`, `api-server`, and `cli` typechecks separately — these two are the only remaining errors, so this is the last thing between the branch and a green build.

PetrBulanek (changes requested):

PetrBulanek (approved):

PR #3253, opened 2026-08-10
feat(ui): rotate tips on the waits that show a spinner

### LINKED ISSUE #3211: UI - Show DAM tips while an agent starts up
## Context

Starting an agent takes a long time. The user sees a full-screen takeover with a spinner, the sandbox name, a `Starting` badge, and one line of copy: "The agent pod is starting up." Nothing else happens until the pod is ready. That wait is dead time on a screen every user hits, and hits most often when they are new.

## Problem

The startup screen wastes the user's attention. It gives no reason to keep watching and nothing to learn. New users wait through it before they know what DAM can do, so their first minutes teach them nothing. The wait also feels longer than it is because the screen never changes.

## Goal

The startup wait teaches the user something useful. Tips rotate on the screen while the agent comes up, so the user leaves the wait knowing one more thing about DAM than when they arrived.

## User Stories

- As a new user, I want to read short tips while my agent starts so that I learn what DAM can do before my first session.
- As a returning user, I want the startup screen to change while I wait so that I can tell the start is progressing and not stuck.
- As a product owner, I want to control the tip content so that we can point users at features they miss.

## Scope

**In:** the startup takeover screen for a starting agent or sandbox.

**Out:** the other states on that screen (hibernated, error, over budget) — those need an action, not a tip.

## Open Questions

- Where do tips come from — a fixed list in the product, or something we can edit without a release?
- Should a tip link to docs, or stay read-only text? A link pulls the user off the screen mid-start.
- How long should each tip show, and should the rotation pause on hover?
- Do we also want this on the "Reconnecting" and workspace-cloning waits?

kapetr:
Rotating tip card on the sandbox startup screen, so the wait teaches something and its movement shows the start hasn't stalled.

Closes #3211

Tips are the list from the issue comment, with the brand and open sandbox name filled in at render (never hardcoded). Shown randomized, as the issue asks.

Two things that go beyond the issue, say the word and I'll pull them:

- Shown on every wait that renders a spinner — loading, reconnecting, starting, preparing workspace, hibernating — not just `starting`. Gated on the spinner, so the icon states (hibernated, error, over budget) stay untouched; those want a click.
- Card is 480px and rotates at 8s with a full fade; the prototype is 420px, 6s, and dips to 0.4. Deliberate, from review feedback, but it is a departure.

No pause control on the rotating text — WCAG 2.2.2 wants one. Worth a follow-up.

### DIFF
--- a/packages/ui/src/modules/agents/components/startup-tip.tsx
+++ b/packages/ui/src/modules/agents/components/startup-tip.tsx
@@ -1,0 +1,100 @@
+import { Idea } from "@carbon/icons-react";
+import { useEffect, useMemo, useRef, useState } from "react";
+
+import { Callout } from "@/components/ui/callout";
+import { cn } from "@/lib/utils";
+
+import { startupTips } from "../startup-tips.js";
+
+/** Time a tip stays put, measured start-of-transition to start-of-transition. */
+const ROTATE_MS = 8_000;
+/** Each direction of the swap. The text is replaced while the card is at zero
+ *  opacity, so the outgoing and incoming tips never share the screen — two
+ *  different sentences at partial opacity is unreadable in a way that two
+ *  fading images is not. */
+const FADE_MS = 250;
+/** The spinner, the sandbox name and its badge should land first; the tip
+ *  arrives after them rather than competing for the same first glance. */
+const ENTER_DELAY_MS = 1_000;
+
+/** A tip other than `current`, so the rotation never appears to stall on a
+ *  repeat. Random rather than sequential: the same sandbox gets restarted a
+ *  lot, and a fixed order turns into one memorized loop. */
+function pickNext(current: number, count: number): number {
+  if (count < 2) return current;
+  return (current + 1 + Math.floor(Math.random() * (count - 1))) % count;
+}
+
+/**
+ * Rotating tips for every wait that shows a spinner. Two jobs: teach the
+ * feature set to someone whose first minutes would otherwise be spent on a
+ * static screen, and let the movement itself be the evidence that the wait
+ * hasn't stalled.
+ *
+ * Read-only text on purpose — a link would pull the user off the screen
+ * mid-wait, and they came to use the sandbox, not to read docs.
+ */
+export function StartupTip() {
+  const tips = useMemo(startupTips, []);
+  const [index, setIndex] = useState(() =>
+    Math.floor(Math.random() * tips.length),
+  );
+  const [shown, setShown] = useState(false);
+  // The rotation reads the live index without listing it as a dependency —
+  // otherwise every swap would tear down and restart the timer, and each tip
+  // would get its full dwell only by accident. Nothing else may enter these
+  // deps either: anything that changes mid-wait restarts the countdown, which
+  // is what a hover-to-pause flag did before it was removed.
+  const indexRef = useRef(index);
+  const swapRef = useRef<ReturnType<typeof setTimeout> | undefined>(undefined);
+
+  useEffect(() => {
+    const enter = setTimeout(() => setShown(true), ENTER_DELAY_MS);
+    // Rotation is clocked from the end of the entrance, so the first tip gets
+    // the same dwell as every one after it.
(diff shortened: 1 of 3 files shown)


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `2f9ff68`

## PR #3253: feat(ui): rotate tips on the waits that show a spinner
**Author:** kapetr | **Branch:** feat/3211-startup-tips → main | **Changes:** +125 −0 (3 files)

### Summary
Adds a rotating tip card to the agent startup takeover, so a spinner wait teaches a feature and its motion signals progress. Tips render from a single list with the brand name and open sandbox filled in at render time.

### Findings
- 🟢 **Suggestion:** The rotation timer logic is a self-contained lifecycle (interval + fade-out swap). Extracting it into a `useRotatingTip` hook would isolate a testable unit and keep the component presentational (`packages/ui/src/modules/agents/components/startup-tip.tsx:40`). Optional.
- ✅ **Looks good:** `duration-600` reads like a Tailwind v3 no-op, but this package is Tailwind v4.2.2 (bare-numeric utilities), so the 600 ms fade-in is real; reduced motion is handled — `motion-reduce:opacity-100` keeps the tip visible without animation.

Verdict: aligned. The change adds a presentational tip card to a UI overlay; it alters no subsystem behavior or responsibility, introduces no new subsystem, and touches no page under `docs/architecture/`.

### Issue Fit
Linked issue: #3211 "UI - Show DAM tips while an agent starts up" (closing keyword `Closes #3211`)

✅ Delivers the ask: rotating, randomized, product-owner-controlled tips on the startup wait; the tip list matches the issue comment verbatim with the brand and sandbox substituted at render instead of hardcoded. The out-of-scope states (hibernated, error, over budget) all carry an icon and are correctly excluded via `!Icon`. Extending the card to the reconnecting and workspace-cloning spinner waits was an explicit open question in the issue and is declared in the PR body — declared scope, not creep.

✅ No findings. `startup-tips.ts` is a pure UI data helper, correctly placed in the `agents` module, with no layer or boundary violation.

- 🟢 **Suggestion:** rotation lifecycle could move to a custom hook (see Findings). No correctness issues: effect cleanup clears every timer and interval; the fade-out swap timeout always resolves before the next interval, so no timers leak. `pickNext` never repeats the current index. `index`/`shown` are correct local UI state.

### Verdict
APPROVE — correct and well-scoped; effect cleanup, index rotation, and reduced-motion handling all verified, and the PR delivers issue #3211.

---

packages/ui/src/modules/agents/components/startup-tip.tsx
```diff
@@ -0,0 +1,82 @@
...
+    return () => {
+      clearTimeout(enter);
+      clearTimeout(start);
+      clearTimeout(swapRef.current);
+      clearInterval(rotate);
+    };
+  }, [tips.length]);
+
+  return (
+    <Callout className="w-full max-w-120 min-h-22 p-5 text-left">
```
PetrBulanek: The card's height depends on how many lines the current tip wraps to, and the
overlay centers its whole column vertically. On a 375px-wide screen the card is
327px wide, and the tip list splits three ways there: 4 tips wrap to two lines
(88px), 27 wrap to three (110px), and one wraps to four (133px). Every rotation
that crosses those groups moves the spinner, the sandbox name, and the status
badge up or down by 11px or 22px. I saw the heading's top edge move between
322px and 333px while watching a live rotation at that width.

The screen is the whole viewport on mobile, so the jump is the most visible
thing on it, and it works against what the card is for — movement that says the
start is progressing reads instead as the layout twitching. Desktop is fine:
at a 480px card every one of the 32 tips is exactly 88px tall, so the wobble is
specific to narrow widths. Whatever the fix, the constraint is that the column
above the card has to hold still across the whole tip list at mobile widths,
not just across the ones that happen to wrap the same way.

PetrBulanek (changes requested):

PetrBulanek (approved):

PR #3205, opened 2026-08-06
docs(adrs): ADR-082 — UI live updates over tRPC WebSocket subscriptions

jezekra1:
## What

Pure ADR PR — records the decision reached in the #3170 design discussion. No code changes.

**ADR-082: UI live updates — thin invalidation events over tRPC WebSocket subscriptions**

The decision, in one breath: the UI stops polling; every change it displays is announced by a thin, signal-only invalidation event (topic + ids, never state) pushed over a single per-tab WebSocket carrying tRPC subscriptions, and the client re-reads the affected queries over the normal tRPC read path. Polling is removed outright (no fallback mechanism); reconnection heals via a sync event on every (re)subscribe.

Boundaries recorded:

- signal-only socket, Redis pub/sub as cross-replica fan-out, no ordering/replay machinery
- dual transport: one tRPC router served over both HTTP and the socket, clients pick links (browser hybrid first, CLI untouched) — so cross-cutting gates must live at the tRPC layer
- token-bounded connections: first-frame auth, pre-expiry reconnect nudge, fresh token on every reconnect
- subscriptions are passive reads — they never wake or keep-warm an agent
- per-domain event sources (Postgres write sites, K8s watch, pod-side) deliberately left to per-subsystem design

## Why now

#3170 counts fourteen polling surfaces (2–15 s intervals); the transport and event-shape questions it left open were settled in an extended design review — SSE vs WS, thin vs fat events, and the connection-auth lifecycle were each argued against the existing codebase (three WS relays, the Redis inject channel, the passive-read invariant).

Refs #3170.

Issue #3170, opened 2026-08-05
Replace UI polling with live updates over a websocket


jezekra1:
## Context

The UI keeps itself current by asking the server again on a timer. Fourteen separate parts of the app do this, on intervals between two and fifteen seconds: the approvals inbox (both the owner-wide one and the per-agent one), the file tree and each open folder, the experiments feed, the agent reachability check, the session list, agents, schedules, budgets, artifacts, and metrics. A live websocket already exists alongside all of this — chat, the terminal, and SSH each run over one — and the approvals path already pushes events onto that socket and then polls on top of it anyway to pick up the rest.

## Problem

Polling was the quickest way to make each of these screens feel live, and it has accumulated into the app's default way of learning that anything changed. It costs us four ways:

- **Wasted work** — an open tab produces a steady request stream against the api-server, the K8s API, and (for the session list) the agent pods themselves, whether or not anything changed. The cost scales with tabs open, not with activity.
- **Lag** — every interval is a compromise between staleness and load. Even the tightest ones still show approvals and file changes up to two seconds late, and the slower lists take proportionally longer.
- **Two mechanisms for one job** — some state arrives pushed, some polled, and some both. Anyone adding a screen has to decide which, and the answer isn't obvious.
- **Poll-driven bugs** — timed refetches race with optimistic updates, so rows flicker or briefly revert to a stale value after a user action.

## Goal

The UI learns about changes when they happen, over a connection it already holds open, and polling survives only as a fallback for when that connection is down. Which screens move first, and the shape of the transport (live subscriptions on the existing API surface, an extension of the socket the chat already uses, or a dedicated events socket), are open — decide during design.

## Done when

- The state the UI displays arrives as pushed events rather than timed refetches.
- No screen depends on an interval to stay correct: a tab left open shows current state without one.
- An idle tab with nothing happening generates no recurring traffic.
- Losing the connection degrades visibly and recovers on its own — after a laptop sleep or a backgrounded tab, no manual reload is needed.
- Events reach every connected client when more than one api-server replica is running.

PR #3258, opened 2026-08-10
refactor(api-server): tRPC-over-WebSocket transport + one composition root

jezekra1:
Adds a **WebSocket transport for the tRPC surface as a 1:1 alternative to the HTTP one**, plus the api-server structural cleanup it motivated. **No behavior change to existing surfaces.** Foundation for #3170 (replace UI polling with live updates) — the subscription procedures and event fan-out land in follow-ups on top; this PR only establishes the transport and the structure.

Boot-verified: full `cluster:uninstall` + `cluster:install` clean pass, api-server pod `1/1 Running` with 0 restarts, serving live ACP relay traffic. `tsc` + lint + 873 unit tests green.

## Transport
- **tRPC-over-WebSocket endpoint** at `/api/trpc-ws`: same `appRouter`, `wsLink` transport, **first-frame `connectionParams` auth** (no `?token=` in the URL, unlike the agent relays), **credential-bounded connections** (reconnect nudge before token `exp`, hard close at it; API-key connections re-auth on a fixed 5-min cadence), graceful drain on shutdown.
- Queries/mutations work over it today; subscriptions become its first new consumer once events are wired.

## Structure
- `apps/api-server` split into **`admission/`** (auth + terms gates, shared authentication leg, per-edge denial mappers), **`trpc/`** (http + ws doors, context factory), **`agent-proxies/`** (the three relays, `URLPattern` upgrade dispatch, trpc/import proxies), **`routes/`** (bootstrap endpoints, brand). Module-owned routes (oauth, slack, telegram, artifacts) moved into their modules with narrowed prefixes.
- **One composition root**: `bootstrap.ts` assembles every singleton and returns per-app deps + a `cleanup` closure; `index.ts` spawns the three servers and conducts shutdown; `app.ts` is pure surface wiring. Fixes a latent divergence (the duplicated `connectionsBoot` was missing `operatorCredentials` on one side) and keeps `oauth-refresh` registered.
- Shared authentication leg threaded to every edge; `UserAuthenticated` now emitted per WS connection too (keeps usage analytics alive as traffic migrates off HTTP).
- Removed the shipped l7-promotion backfill migration.

Companion decision record: #3205 (ADR-082).

Refs #3170.


packages/api-server/src/apps/api-server/trpc/ws.ts
```diff
@@ -0,0 +1,131 @@
...
+    const closeInMs = Math.max(
+      (expiresAt?.getTime() ?? Date.now() + API_KEY_REAUTH_MS) - Date.now(),
+      0,
+    );
+    const nudgeInMs = Math.max(closeInMs - RECONNECT_NUDGE_BEFORE_MS, 0);
+    const timers = [
+      setTimeout(() => {
+        // The protocol's reconnect notification — the same frame
+        // broadcastReconnectNotification sends, scoped to one client.
+        ws.send(JSON.stringify({ id: null, type: "reconnect" }));
```
dam-code-guardian: 🔴 **Critical:** this frame is silently discarded by `wsLink`, so the pre-expiry reconnect never happens. The client routes incoming frames on `'method' in message` (`wsClient.ts:329`); `{id: null, type: "reconnect"}` has no `method`, so it falls through to `handleResponseMessage`, where `getPendingRequest(null)` returns `null` and the handler returns early — no reconnect, no error, no log. The connection then survives until the `4401` close fires, which is exactly the mid-flight drop the 30-second nudge exists to avoid. `TRPCReconnectNotification` in `@trpc/server` 11.16 is `BaseRequest<'reconnect'>`, i.e. `{id, method: 'reconnect'}` — which is what `broadcastReconnectNotification()` sends, so `drain()` works and only this hand-built frame is wrong.

**Fix:** key protocol frames on `method`, not `type` — for every hand-built tRPC frame in this file, not just this one.

```suggestion
        ws.send(JSON.stringify({ id: null, method: "reconnect" }));
```

packages/api-server/src/apps/api-server/trpc/ws.ts
```diff
@@ -0,0 +1,131 @@
...
+ *  `/api/trpc`, different transport. Auth rides the first frame
+ *  (connectionParams), never the URL, and a connection never outlives its
+ *  credential — nudged reconnect before the deadline, hard close at it. */
+export function createTrpcWsEndpoint(deps: TrpcWsDeps) {
+  const wss = new WebSocketServer({ noServer: true });
+
+  function attachCredentialLifecycle(
+    ws: WebSocket,
+    expiresAt: Date | undefined,
+  ): void {
```
dam-code-guardian, line 53: 🟡 **Warning:** for an API-key principal this is a deadline, not a re-authentication. `expiresAt` is absent for API keys, so `closeInMs` becomes `API_KEY_REAUTH_MS`; the timer then closes the socket with `4401` and `wsLink` reconnects, replaying the same stored token — nothing calls `deps.authenticate` again on the live connection. The file's own comment calls this interval "the platform's revocation latency for socket-held keys", but a revoked key is only rejected because the *reconnect* re-runs admission; if the client does not reconnect (lazy mode, no pending subscriptions), the 5-minute bound buys nothing, and the once-per-connection `emitUserAuthenticated` under-counts the re-auth it claims to do.

**Fix:** make the re-auth cadence verify rather than merely disconnect — re-run `deps.authenticate` against the connection's token on each tick and close only when it fails. State the rule for the whole class: any connection-lifetime bound derived from a credential must re-verify that credential, not just expire on a timer.

packages/api-server/src/modules/connections/oauth-routes.ts
```diff
@@ -1,15 +1,15 @@
 import { Hono } from "hono";
 import type { Db } from "db";
-import type { OAuthEngine } from "../../modules/connections/infrastructure/oauth-engine.js";
-import type { ConnectionTemplateRegistry } from "../../modules/connections/domain/connection-template.js";
-import type { SecretStore } from "../../modules/secret-store/index.js";
+import type { OAuthEngine } from "./oauth-engine.js";
+import type { ConnectionTemplateRegistry } from "../domain/connection-template.js";
+import type { SecretStore } from "../../secret-store/index.js";
 import {
```
dam-code-guardian, line 6: 🟡 **Warning:** this file is a driving (inbound) adapter, but the move lands it in `infrastructure/`, the layer for driven adapters — which is what forces the `../services/oauth-flow.js` import across the documented `services → domain ← infrastructure` direction, and puts `createOAuthFlowService` + `createConnectionsRepository` wiring inside a request handler instead of the module's `compose.ts`. `modules/connections/index.ts` also still exports only the repository, so `apps/api-server/routes/index.ts:10` reaches two levels past the module boundary to mount it.

**Fix:** inbound route adapters belong at the module root and are published through the module's `index.ts` — `modules/artifact-library/library-routes.ts` is the pattern already followed one import above in the same file. Compose the flow service in `connections/compose.ts` and pass it in, so the handler only parses and delegates.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `c8a5d2f`

## PR #3258: refactor(api-server): tRPC-over-WebSocket transport + one composition root
**Author:** jezekra1 | **Branch:** refactor/ui-live-updates#locki-lpaitkhu → main | **Changes:** +3142 −2722 (41 files)

### Summary
Adds a tRPC-over-WebSocket transport at `/api/trpc-ws` beside the HTTP one, and splits `apps/api-server` into `admission/` / `trpc/` / `agent-proxies/` / `routes/` behind a single `bootstrap.ts` composition root. The transport work is careful: first-frame auth, credential-bounded connections, and a drain on shutdown.

### Findings
- 🔴 **Critical:** the credential-expiry reconnect nudge sends a frame the client ignores (`packages/api-server/src/apps/api-server/trpc/ws.ts:63`) — inline
  **Fix:** key protocol frames on `method`, not `type` — send `{id: null, method: "reconnect"}`, the shape `TRPCReconnectNotification` defines and `drain()` already emits.
- 🟡 **Warning:** the tRPC WS door only bounds JWT connections; API-key connections never re-authenticate (`packages/api-server/src/apps/api-server/trpc/ws.ts:53`) — inline
  **Fix:** re-run `deps.authenticate` on the stored token at each `API_KEY_REAUTH_MS` tick and close on failure; a deadline is only a re-auth if something verifies again.
- 🟡 **Warning:** the new module-owned OAuth routes sit in `infrastructure/` and are imported past the module boundary (`packages/api-server/src/modules/connections/infrastructure/oauth-routes.ts:6`) — inline
  **Fix:** put inbound route adapters at the module root and export them from the module's `index.ts`, as `artifact-library` already does; compose the service in `compose.ts`, not per request.
- 🟢 **Suggestion:** four files now import `IsAcceptedPort` from `modules/terms/compose.js` while `modules/terms/index.js` already re-exports it — the split propagated one boundary bypass to `trpc/ws.ts:8`, `agent-proxies/upgrade.ts:7`, `admission/terms.ts:2`, `deps.ts:45`.
- 🟢 **Suggestion:** `createUpgradeHandler` puts `new URL(...)` outside the `try` its own comment says keeps a throwing handler from crashing the process (`agent-proxies/upgrade.ts:150`). A malformed `Host` header reaches the listener and rejects unhandled, which ends the process; the pattern is pre-existing on `main`, so it belongs in a separate hardening issue covering `terminal-relay.ts:85` too.
- ✅ **Looks good:** `attachCredentialLifecycle` receives the adapter's `res`, which is the live `ws` client in `@trpc/server` 11.16 — the timers and their `once("close")` cleanup are correctly wired, and `noServer` mode still tracks `wss.clients`, so `drain()` and `close()` reach every socket.

## Verdict

**significant drift** — no file under `docs/architecture/` is touched by this PR, while the diff both removes a subsystem behavior that a page describes as present and adds a new user-facing protocol edge that no page mentions.

## Drift

### Check 1 — Architecture-page drift

**❌ 1a. `security-and-credentials.md` still describes the L7-promotion startup migration that this PR deletes.**

Code evidence (the migration is removed in full, not relocated):

- `packages/api-server/src/modules/egress-rules/services/l7-promotion-backfill.ts` — **deleted** (79 lines), including the marker-Secret retirement (`LEGACY_MARKER_SELECTOR = "agent-platform.ai/secret-type=allow-only,agent-platform.ai/managed-by=api-server"`) and its "markers go only once every live agent carries its promotions" ordering.
- `packages/api-server/src/modules/egress-rules/compose.ts` — `createL7PromotionBackfill` export and the `backfillL7Promotions` import removed.
- `packages/api-server/src/index.ts` — the whole capped-retry driver removed: `l7PromotionBackfill`, `runL7PromotionBackfill()`, `L7_BACKFILL_MAX_RETRIES = 10`, the 60 s `setTimeout` retry, and the `await runL7PromotionBackfill()` boot call.
- `packages/api-server/src/__tests__/unit/l7-promotion-backfill.test.ts` — **deleted** (253 lines).
- Verified no surviving reference: `backfillL7Promotions` / `createL7PromotionBackfill` / `l7-promotion-backfill` appear nowhere in the head tree.

Doc evidence — `docs/architecture/security-and-credentials.md:446-449`, in the **L7 promotion** subsection, still asserts the migration as a live property of the subsystem:

> A one-shot api-server startup migration projects promotions previously
> materialized as owner-scoped credential-less marker Secrets onto the
> Agent resources and retires the markers; the controller consumes both
> signals, so rules stay enforced mid-migration.

Every clause of that sentence is now false on the api-server side: nothing projects the legacy markers, nothing retires them, and there is no "mid-migration" window to keep enforced. This is a responsibility the api-server no longer has — squarely inside the drift rule.

Note for whoever applies the fix: the *controller* half is unchanged and must not be edited away. `packages/controller/pkg/reconciler/envoy.go:156` still defines `envoySecretTypeAllowOnly = "allow-only"` and `envoy.go:452` still states "The Secret-shaped allow-only flavor remains consumed for …". So the controller still renders both signals; what disappeared is only the api-server's one-shot projection/retirement pass. Any marker Secrets left in an install that never completed the migration now age out by hand rather than by the boot pass.

Proposed edit — delete the four-line sentence at `security-and-credentials.md:446-449`. The surrounding paragraph (`l7Hosts` is a pure projection recomputed after every create/edit/revoke, lines 438-445) already describes the steady state correctly and stands alone without it. If the reviewers want the dual-signal fact retained, it belongs as a durable statement about the controller ("the controller renders promotions from both the Agent resource's `l7Hosts` and the legacy allow-only marker Secrets") rather than as a migration narrative with a completion horizon — but the simple deletion is the smaller, more durable change.

**❌ 1b. The new tRPC-over-WebSocket transport edge is undocumented in `platform-topology.md`.**

Code evidence — a new client-facing protocol surface, not an internal move:

- `packages/api-server/src/apps/api-server/trpc/ws.ts` — **new file** (131 lines): `createTrpcWsEndpoint`, described in its own header as "The tRPC-over-WebSocket endpoint (`/api/trpc-ws`): same router as `/api/trpc`, different transport."
- `packages/api-server/src/apps/api-server/app.ts` — the endpoint is wired into the upgrade dispatch: `"/api/trpc-ws": selfAuthenticated(trpcWs)`, and `startApiServerApp` now returns `{ server, trpcWs }`.
- Two architecturally meaningful contracts, both new and neither derivable from an existing page:
  - **A different authentication model for this edge.** The three agent relays admit on a query-string token (`relayRoute(relayAdmission, …)`); this edge authenticates *in-protocol* on the first frame via tRPC `connectionParams` — "Auth rides the first frame (connectionParams), never the URL". That is a trust-boundary distinction, exactly the "security layers, trust boundaries" content the guidelines say to include.
  - **Credential-bounded connection lifetime.** A connection never outlives its credential: a `reconnect` frame `RECONNECT_NUDGE_BEFORE_MS = 30_000` before the deadline, then a hard close with application code `4401`; API-key principals expose no `expiresAt` and so re-authenticate on a fixed `API_KEY_REAUTH_MS = 5 * 60_000` cadence, which the code itself names "the platform's revocation latency for socket-held keys". Plus a pre-shutdown `drain()` that broadcasts a reconnect so clients land on surviving replicas.

Doc evidence — `docs/architecture/platform-topology.md:78-95` enumerates every edge in an Edge/Protocol/Purpose table. It lists `ui → api-server | tRPC over HTTP`, the ACP WebSocket, and the binary-terminal WebSocket, but has no row for tRPC over WebSocket. Confirmed by search: `trpc-ws`, `connectionParams`, `wsLink`, and "tRPC over WebSocket" match nothing anywhere under `docs/architecture/` or in `docs/architecture.md`. The api-server subsystem prose (`platform-topology.md:39-50`) is likewise silent.

The PR body's "**No behavior change to existing surfaces**" is accurate and not in dispute — the drift is not a changed surface but an *added* one: a second transport for the entire tRPC router, with its own auth model and its own connection-lifetime contract, is a change in the api-server's responsibility.

Proposed edit — in `platform-topology.md`, add one row to the edge table, e.g.:

| `ui → api-server` | WebSocket (tRPC) | Same tRPC router as the HTTP transport; authenticated in-protocol on the first frame rather than by URL token, and the connection is bounded by its credential (reconnect nudge, then close at expiry) |

and, in the api-server subsystem prose, one sentence noting that the tRPC surface is reachable over both HTTP and a WebSocket transport, with the credential-bounded-connection invariant stated semantically. Keep it at protocol-meaning level — per the content policy, do not transcribe the literal path, the `4401` close code, or the two millisecond constants; link out to [`packages/api-server/`](../../packages/api-server/) for those.

### Check 2 — `Last verified:` staleness

**❌** No architecture page is edited in this diff, so no `Last verified:` header is bumped. Both pages that need edits above are consequently stale: `docs/architecture/security-and-credentials.md:3` reads `Last verified: 2026-08-07` and `docs/architecture/platform-topology.md:3` reads `Last verified: 2026-08-07`.

Proposed edit — bump each edited page's header to the PR date, `2026-08-10`. (`docs/architecture.md:3`, `Last verified: 2026-08-04`, needs no bump: the landing page's subsystem list and system-context diagram remain accurate — no subsystem is added or removed.)

## Possible drift

**⚠️ Does the artifact-library / OAuth route re-prefixing change any externally observable path?**

The module-owned route files now register narrowed prefixes — `library-routes.ts` moves `"/api/artifact-library/upload"` → `"/upload"` and `"/api/artifact-library/:id/download"` → `"/:id/download"`; `slack-oauth.ts` and `telegram-oauth.ts` move `"/api/slack|telegram/oauth/callback"` → `"/oauth/callback"`; `oauth-routes.ts` moves `"/api/oauth/callback"` → `"/callback"`. The intent is clearly that the mounting site (`mountRoutes`) restores the full prefix, and `app.ts`'s `PUBLIC_PATHS` still lists the full unauthenticated paths (`/api/oauth/callback`, `/api/slack/oauth/callback`, `/api/telegram/oauth/callback`), which is consistent with the external paths being unchanged.

The human-judgement question: are the externally published callback URLs (registered with Slack, Telegram, and each OAuth provider) and the artifact-library upload/download paths byte-identical to before? If any prefix shifted, `security-and-credentials.md`'s public-surface description and `artifact-library.md` would need a look. Nothing in the diff suggests a shift — this is flagged only because a mount-prefix refactor is where such a change would hide, and confirming it is cheaper than discovering it from a broken provider callback.

*(Check 7 produced nothing: no file under `docs/adrs/` is touched by this diff, so no ADR status changed. Checks 3, 4, 5, and 6 produced nothing — no architecture page is edited, moved, renamed, or deleted, and the refactor introduces no new long-lived component: `bootstrap.ts`, `admission/`, `trpc/`, `agent-proxies/`, and `routes/` are all internal structure within the existing api-server subsystem, which already has its page.)*

### Issue Fit
Linked issue: #3170 "Replace UI polling with live updates over a websocket" (inferred from the plain reference `Refs #3170`; no closing keyword, so merging does not close it)

The issue's five "Done when" criteria are all about the UI side — pushed events replacing timed refetches, no interval-dependent screens, silent idle tabs, self-healing reconnect, cross-replica delivery. None are delivered here, and none are claimed: the PR scopes itself to "only the transport and the structure", with subscription procedures and event fan-out in follow-ups. With a `Refs` link rather than a closing keyword, that deferral leaves the issue open and is legitimate staging, not missing scope. #3205 was also referenced but is a PR (ADR-082), not an issue — dropped as a candidate.

Against that declared scope, the diff delivers what it promises: the `/api/trpc-ws` endpoint serving the same `appRouter` with first-frame `connectionParams` auth, credential-bounded connections and drain (`packages/api-server/src/apps/api-server/trpc/ws.ts`), wired at `app.ts:177`; the four-way `admission/` + `trpc/` + `agent-proxies/` + `routes/` split; and a single composition root in `bootstrap.ts` returning per-app deps plus a `cleanup` closure (`bootstrap.ts:1066`). Both declared incidental fixes check out — base had two `composeConnectionsAtBoot` call sites, only one passing `operatorCredentials`, now one at `bootstrap.ts:341`, and `oauth-refresh` stays registered (`bootstrap.ts:405`).

- 🔵 The body's headline "**No behavior change to existing surfaces**" is contradicted by its own last structural bullet: removing the l7-promotion backfill retires a shipped startup migration that ran on every boot (`packages/api-server/src/modules/egress-rules/compose.ts:133`, plus the deleted `services/l7-promotion-backfill.ts` and its 253-line test). The removal is declared, so it is not undeclared creep, but it is a boot-behavior change bundled into a transport refactor and unrelated to #3170 — worth either its own PR or a body that stops claiming no behavior change. Fix: qualify the "no behavior change" claim to exclude the retired migration, and state why the backfill is now safe to drop (i.e. that it has completed everywhere it needed to).

Everything else substantial in the diff is either asked-for or declared: the `import-proxy`/`trpc-proxy`/`URLPattern` upgrade-dispatch extractions and the oauth/slack/telegram/artifacts route moves are the declared structural split, and the renames, import rewrites and test-path updates are mechanical fallout of it. No changes outside `packages/api-server/` — consistent with a transport-only foundation that leaves the fourteen polling surfaces untouched for the follow-ups.

TypeScript Engineering — Architecture Review (mode: review)

Scope: PR #3258 ("refactor(api-server): tRPC-over-WebSocket transport + one
composition root"), 39 routed TypeScript files in `packages/api-server`.
Baseline = `main`. Only violations introduced or amplified by this change are
reported; pre-existing ones the PR does not touch are out of scope.

Architecture docs loaded: index.md, slice-composition.md, modules.md,
module-boundaries.md, infrastructure.md, project-structure.md, stack.md.

Note on project metadata: there is no `tseng/project-structure.md` in the repo,
so package roles were inferred — `packages/api-server` (server),
`packages/api-server-api` (contract, exports `appRouter` + `ApiContext`),
`packages/ui` (client). That inference does not affect any finding below.

What this PR does well, stated up front because it is most of the diff:

  • One composition root. `bootstrap.ts` constructs every singleton and returns
    per-app deps plus a `cleanup` closure; `index.ts` conducts; `app.ts` becomes
    pure surface wiring. On `main`, `app.ts` itself built `createK8sClient`,
    `createAgentsRepository`, `createTemplatesRepository`,
    `createReposRepository`, `createAuth`, `startJwksWarmup` and
    `createSessionPresence` (base app.ts:207-306, 1040) — concrete
    infrastructure referenced outside any composition root. That is now gone,
    and the duplicated `connectionsBoot` that had drifted on
    `operatorCredentials` is gone with it. This is infrastructure.md's
    Composition Root pattern applied at application scope.
  • Ports and adapters at the edges. `Authenticate`, `RelayAdmission` and
    `IsAcceptedPort` are port interfaces; `authenticatePrincipal` is the single
    adapter behind the first; `app.ts:78` builds one `authenticate` closure and
    threads it into all three edge families so "no edge can construct its own".
    That is the drift argument infrastructure.md makes for injection.
  • Module-owned routes moved into their modules, with app-owned prefixes
    narrowed (`/api/oauth`, `/api/slack`, `/api/telegram`,
    `/api/artifact-library` now supplied at the mount site). Under modules.md a
    bounded context's adapters belong in its vertical slice, not the app shell;
    this corrects a layering violation that existed on `main`.

Baseline acknowledgement (not a finding): the composition root still reaches
past module `compose.ts` roots into concrete infrastructure
(`createAgentsRepository`, `createTemplatesRepository`,
`createEgressRuleMatchAdapter`, the `findX(db)` repository closures). Nearly all
of that moved verbatim from base `index.ts`; the PR relocates it rather than
adding to it. Net-neutral against modules.md, so not flagged — with the one
exception noted in finding 5.

────────────────────────────────────────────────────────────────────────────
FINDINGS
────────────────────────────────────────────────────────────────────────────

1. Infrastructure layer imports from services, and composes a service +
   repository outside `compose.ts`
   ── NEW (introduced by this change)

   Rule: "Infrastructure may import from the domain layer (to implement ports
   and map types) but never from services." Dependency rule:
   `services → domain ← infrastructure`. And: each module's `compose.ts` "is the
   **only place** where infrastructure implementations are directly referenced."
   Source: architecture/slice-composition.md (Layer Rules §3, Dependency Rule);
   architecture/modules.md (Module Composition Root); architecture/infrastructure.md.

   Location: packages/api-server/src/modules/connections/infrastructure/oauth-routes.ts

   Evidence:
     line 6-9   `import { createOAuthFlowService, type OAuthFlowPendingCtx }
                  from "../services/oauth-flow.js";`   ← infrastructure → services
     line 10    `import { createConnectionsRepository } from "./connections-repository.js";`
     line 68-76 `const flow = createOAuthFlowService({ engine: deps.engine,
                  repo: createConnectionsRepository(deps.db), … });`
                 ← a service and its repository are wired inside a request
                   handler, in `infrastructure/`, not in `connections/compose.ts`

   Why new rather than pre-existing: on `main` this file was
   `apps/api-server/oauth.ts` — outside `modules/`, where intra-module layer
   rules do not apply. Moving it into the `connections` module is the right call
   in itself, but it lands in `infrastructure/`, the layer for *driven* adapters
   (repositories, mappers, outbound clients). An inbound HTTP route is a
   *driving* adapter. That placement is what pulls the forbidden `services`
   import across the layer line and puts per-request service composition outside
   the module's composition root.

   The same shape exists in `channels/infrastructure/slack-oauth.ts:5`
   (type-only import of `IdentityLinkService` from `../services/`), but that
   file and that import are unchanged on `main` — pre-existing, out of scope.
   Only the `connections` one is introduced here.

2. App layer imports a module internal, bypassing the module's `index.ts`
   ── NEW (introduced by this change)

   Rule: "No module may import another module's internals (services, entities,
   infrastructure)… Any import that bypasses `index.ts`" is not allowed. "Every
   module has an `index.ts` at its root. This is the **only** entry point other
   modules may import from."
   Source: architecture/module-boundaries.md (What Is Not Allowed);
   architecture/modules.md (Module Public API).

   Location: packages/api-server/src/apps/api-server/routes/index.ts:10

   Evidence:
     `import { createOAuthRoutes } from
      "../../../modules/connections/infrastructure/oauth-routes.js";`
     while `modules/connections/index.ts` (4 lines, untouched by this PR)
     exports only `createConnectionsRepository` / `ConnectionsRepository`.

   The PR adds a new file to the `connections` module and mounts it from the app
   without adding it to that module's public API — the new module-owned route is
   reachable only by reaching two levels past the boundary. Contrast
   `artifact-library`, which this same PR mounts correctly: `routes/index.ts:3-6`
   imports `createArtifactLibraryRoutes` from `modules/artifact-library/index.js`,
   and that module's `index.ts:12` exports it. The correct pattern is already in
   the same file, four lines above.

   Related, same file: the two channel route factories are imported by deep path
   (`routes/index.ts:7-8` → `modules/channels/infrastructure/slack-oauth.js`,
   `…/telegram-oauth.js`). `modules/channels/` has no `index.ts` at all — the
   only module in this package without one. Those two imports are unchanged in
   substance from `main`, so they are context rather than a new violation, but
   they are cleared by the same fix.

3. `index.ts` bypass for `IsAcceptedPort` amplified from one file to four
   ── AMPLIFIED

   Rule: same as finding 2 — the module `index.ts` is the only entry point other
   modules may import from.
   Source: architecture/module-boundaries.md; architecture/modules.md.

   Locations (all import `modules/terms/compose.js` for a type that
   `modules/terms/index.js` already re-exports):
     packages/api-server/src/apps/api-server/trpc/ws.ts:8
     packages/api-server/src/apps/api-server/agent-proxies/upgrade.ts:7
     packages/api-server/src/apps/api-server/admission/terms.ts:2
     packages/api-server/src/apps/api-server/deps.ts:45

   Evidence: `modules/terms/index.ts` is a single line —
   `export { composeTermsModule, type IsAcceptedPort } from "./compose.js";`
   The public API already exposes exactly what these four call sites want. On
   `main` this bypass existed once (base `apps/api-server/app.ts:92`); splitting
   that file into `admission/`, `trpc/`, `agent-proxies/` and `deps.ts`
   propagated the deep path to four files instead of correcting it at the split.
   Cheapest finding in the set to clear: four one-line edits against an export
   that already exists.

4. A domain event gains a second publisher in the app layer
   ── AMPLIFIED

   Rule: "Events are defined in the **publishing module's** domain layer… The
   publishing module is the single source of truth for the event's shape."
   Publishing: "Services emit events after successful state changes."
   Source: architecture/module-boundaries.md (Event Ownership, Publishing).

   Locations:
     packages/api-server/src/apps/api-server/admission/auth.ts:307-326
       (`emitUserAuthenticated`, extracted by this PR)
     packages/api-server/src/apps/api-server/trpc/ws.ts:106
       (`emitUserAuthenticated(admitted.principal, deps.surfaceAttribution);`)
     packages/api-server/src/apps/api-server/admission/auth-middleware.ts:38
       (the pre-existing HTTP site, relocated by this PR)

   Evidence: `UserAuthenticated` has no owning module — it is declared in the
   root registry `src/events.ts:12,30` and published from admission code under
   `apps/`, never from a service inside a bounded context. No module in this
   package has a `domain/events/` directory at all.

   Scope note: the *definition location* is a repo-wide pre-existing convention
   and is out of scope. What this PR adds is a **second app-layer publisher** —
   the WS door — which is the amplification. Extracting the shared
   `emitUserAuthenticated` helper so the two sites cannot drift is the right
   call given two sites now exist (and it is what keeps usage analytics alive as
   traffic migrates off HTTP, per the PR description); the finding is that
   neither site sits where the rule puts publishers.

5. `createEgressRuleMatchAdapter` is constructed per call inside a hot closure
   ── CARRIED, but now in the file this PR authors

   Rule: `compose.ts` / the composition root "receives raw infrastructure
   dependencies … and returns fully wired services" — adapters are built once at
   composition time, not per invocation.
   Source: architecture/infrastructure.md (Composition Root, Service Factory).

   Location: packages/api-server/src/bootstrap.ts:699

   Evidence:
     ruleMatcher: {
       match: async (agentId, host, method, path) => {
         const matched = await createEgressRuleMatchAdapter(db).match(…)

   A fresh adapter is constructed on every egress rule match — this port is
   consulted before every identity resolve on the egress path, so it is the
   hottest adapter in the file. Verbatim from base `index.ts:613`, so not
   introduced; flagged because `bootstrap.ts` is a new file authored by this PR
   and is now *the* composition root, which is where the rule lands. One-line
   hoist to a `const`.

6. Optional dependency re-declared in the new deps interface
   ── CARRIED OVER (touched, not worsened — listed for completeness)

   Rule: "Service dependencies are **required**, never optional… it forces every
   call site to branch on presence (`if (deps.notifier) …`), hides a missing
   wiring as a silent skip instead of a type error… The on/off switch belongs to
   the sender, configured with the rest of its wiring at the root."
   Source: architecture/infrastructure.md (Required Dependencies — Never Optional).

   Locations:
     packages/api-server/src/apps/api-server/deps.ts:73
       `telegramBindFlows?: TelegramBindFlowStore;`
     consumers that branch on presence:
       packages/api-server/src/apps/api-server/routes/index.ts:104
         `if (config.telegramBotToken && boot.telegramBindFlows) {`
       packages/api-server/src/apps/api-server/trpc/context.ts:114
         `telegramBinding: telegramBindFlows ? { … } : undefined,`

   The doc's exact failure mode is visible at the first consumer: it tests both
   the config flag *and* the dep, because neither alone tells the reader whether
   absence is intended. Verbatim from `main` (base `app.ts:140`, with the same
   two branches at 469 and 864); the PR relocates the declaration into the new
   `ApiServerDeps` rather than introducing it, and the branch count is unchanged.
   Flagged only because the PR re-declares the interface — the split was the
   cheap moment to give the store a no-op implementation and drop the `?`.
   `metricsReader: MetricsReader | null` (deps.ts:99, branched at
   trpc/context.ts:238) is the same shape, equally carried, same fix.

Two structural notes that are deliberately not findings:

  • `trpc/context.ts` (new file) composes fourteen modules and four
    infrastructure adapters — in substance a second, request-scoped composition
    root sitting in a transport folder. It is not filed as a violation because
    the composition is genuinely per-owner (the owner is known only per request
    or per connection, so it cannot move into `bootstrap()`), and the body moved
    verbatim out of base `app.ts:826-1000`. Extracting it so both tRPC doors
    share one factory is a real improvement — the file's own comment, "One
    factory so the two doors can never drift apart", is the rule stated back
    correctly. If the folder placement is to be corrected, the cheap move is a
    rename to `apps/api-server/compose-context.ts`, a composition-root name at
    application scope; both doors already receive `composeApiContext` by
    injection, so it is an import-only change.
  • `modules/egress-rules/index.ts` and `modules/approvals/index.ts` are
    literally `export {}` with a comment saying the composition root imports
    `compose.ts` directly — while `bootstrap.ts:110-122` and
    `trpc/context.ts:21-27` do exactly that for eight symbols. This is a genuine
    tension *in the docs*, not a code defect: modules.md says `index.ts` is the
    only entry point other modules may import from, *and* that only event types
    and guards belong in it, *and* its own example has the root importing
    `compose.js`. Per this repo's CLAUDE.md ("If you spot a contradiction
    between the skill and a Platform architecture doc, stop and flag it"), it is
    flagged rather than resolved. Note the repo is internally inconsistent about
    it too: `skills/`, `agents/`, `terms/`, `artifact-library/` all re-export
    their compose functions from `index.ts`; these two deliberately do not.
    Neither file is modified by this PR.

Checked and clean (no finding):
  • `trpc/http.ts` + `trpc/ws.ts` both serve the contract package's `appRouter`
    (`api-server-api/router`) and supply only `createContext`; the server
    package defines no routers and no Zod schemas of its own, so validation
    stays in the contract package. Matches slice-composition.md's Validation
    Layer rule and stack.md's single-`AppRouter` requirement. Two transports
    over one router is explicitly permitted ("multiple root routers as long as
    the modularization approach … is followed").
  • Stack conformance: tRPC + Zod + strict TS + pnpm workspaces + an RxJS event
    bus (`events.ts` uses `Subject`/`filter` with an `EventType` enum and a
    `DomainEvent` union, matching module-boundaries.md's Event Registry). Hono
    via `@hono/node-server`; the WS door uses the documented
    `@trpc/server/adapters/ws`, the HTTP door `adapters/fetch`.
  • `admission/mappers.ts`, `trpc/mappers.ts`, `agent-proxies/mappers.ts` are
    pure `Record<DenialKind, WireEncoding>` tables (plus one `httpTermsStale()`
    function for the data-carrying case) with no I/O and no inward imports —
    infrastructure.md's unidirectional Mappers shape. The `Record` typing means
    a new denial kind cannot be added without every edge encoding it.
  • Errors as values, not exceptions: `core/result.ts` is the documented
    `{ ok: true; value } | { ok: false; error }`; `AuthenticateResult` and
    `RelayAdmissionResult` follow the same discriminated shape, and
    `checkWsTermsAccepted` returns `TermsDenialKind | null`. Matches
    slice-composition.md's Error Handling.
  • Sagas: `startChannelCleanupSaga` and `startSkillsCleanupSaga` return
    subscription handles torn down in `cleanup` (bootstrap.ts:1050-1051) — "A
    saga returns a subscription handle so it can be torn down on shutdown" —
    and are started from the composition root, as module-boundaries.md's Saga
    Rules prescribe.
  • `app.ts` constructs no singletons, as its docstring claims — verified
    against base app.ts:207-306 and 1040, where it constructed seven. The
    remaining `create*` calls in the new `app.ts` build per-app *surface*
    objects (relays, upgrade dispatch, the WS endpoint) from injected
    singletons, which is surface wiring, not composition.
  • `agent-proxies/upgrade.ts`: table-driven `URLPattern` dispatch where the
    dispatcher "knows nothing about what a route needs from its path" (line
    136); route-specific knowledge is isolated in `relayRoute` /
    `selfAuthenticated`, and a pattern missing `:id` throws as a wiring bug
    rather than failing open.
  • `modules/egress-rules/compose.ts` — clean removal of
    `createL7PromotionBackfill` and its `backfillL7Promotions` import; no
    dangling references.
  • `modules/artifact-library/library-routes.ts`,
    `channels/…/slack-oauth.ts`, `channels/…/telegram-oauth.ts` — route-prefix
    narrowing only; `library-routes` is correctly exported through
    `artifact-library/index.ts`.
  • Tests were adapted by mounting the route factories under a real
    `new Hono().route(prefix, …)` rather than by loosening the factories'
    contracts — matching infrastructure.md's "don't loosen a required dep to
    optional to ease setup".
  • The `admission/` `trpc/` `agent-proxies/` `routes/` folders live under
    `apps/`, not `modules/` — they are app-surface wiring, not bounded
    contexts, so modules.md's "modules are defined by business boundaries, not
    technical concerns" does not reach them. No finding.

────────────────────────────────────────────────────────────────────────────
SMALLEST SET OF CHANGES
────────────────────────────────────────────────────────────────────────────

Three edits clear five of the six findings; two of the three are mechanical.

A. Move the OAuth callback routes out of the driven-adapter layer and publish
   them on the module boundary. Clears findings 1 and 2.

   1. Move `modules/connections/infrastructure/oauth-routes.ts` to the module
      root (`modules/connections/oauth-routes.ts`) — alongside `compose.ts`,
      where a driving adapter may legally see `services/`. Precedent in this
      repo: `modules/artifact-library/library-routes.ts` already sits at its
      module root for exactly this reason.
   2. Hoist the per-request wiring into `connections/compose.ts`: have it build
      the `OAuthFlowService` (or a factory for it) and pass it in, so
      `createOAuthRoutes` receives a composed service instead of calling
      `createOAuthFlowService` + `createConnectionsRepository` itself. The route
      handler then only parses the request and delegates.
   3. Add `export { createOAuthRoutes } from "./oauth-routes.js";` to
      `modules/connections/index.ts` and point `routes/index.ts:10` at
      `modules/connections/index.js`. Adding a `modules/channels/index.ts` that
      exports the two channel route factories cleans the remaining two deep
      imports in the same file at the same time.

B. Retarget the four `IsAcceptedPort` imports from `modules/terms/compose.js`
   to `modules/terms/index.js` (`trpc/ws.ts:8`, `agent-proxies/upgrade.ts:7`,
   `admission/terms.ts:2`, `deps.ts:45`). Clears finding 3 — one line per file
   against an export that already exists.

C. Hoist `createEgressRuleMatchAdapter(db)` out of the `ruleMatcher.match`
   closure in `bootstrap.ts:699` to a `const` at composition time. Clears
   finding 5 — one line, and it stops rebuilding the hottest adapter in the
   file on every egress match.

Finding 6 is a fourth, optional edit: make `telegramBindFlows` required and
select a no-op store in `bootstrap.ts` when the token is unset, which deletes
both presence branches; the same shape then closes `metricsReader`.

Finding 4 needs a decision rather than a move, so it is deliberately not folded
into A-C: `UserAuthenticated` is an authentication fact with no bounded context
owning it. Either give it one (an `identity`/`admission` module owning the event
in its `domain/events/`, with both edges calling a thin service that emits) or
record it explicitly as an app-level telemetry signal intentionally outside the
domain-event contract. Adding the second publisher without settling this is
what makes the ownership gap load-bearing.

### Verdict
REQUEST_CHANGES — the structural split is sound and boot-verified, but the credential-bounded-connection guarantee does not hold: the reconnect nudge is silently dropped by the client, and API-key connections never re-authenticate.

---

packages/api-server/src/modules/egress-rules/compose.ts
```diff
@@ -134,22 +133,6 @@ export {
   type AgentL7HostsPort,
```
dam-code-guardian, line 164: 🟡 **Warning:** removing `createL7PromotionBackfill` leaves `spec.l7Hosts` reachable only through a rule mutation. `egress-rules-service.ts:62-64` still promises the opposite — its comment names this backfill as the heal path: *"a missed promotion also heals on api-server boot (the startup backfill re-projects every agent with active narrow rules)"*.

The two surviving projection sites both run only on a write (`egress-rules-service.ts:168` create, `:225` update, plus revoke). So after a failed CR patch — the exact case the comment covers, conflict retries exhausted — that agent's promoted host stays unprojected until somebody edits a rule. The controller then renders no L7 chain for it, and its path-specific rules stop being enforced as narrow while the rules table says they are.

**Fix:** retire a one-shot migration only once a mutation-independent path keeps the derived state converged — keep a boot-time or periodic reconcile over agents with active narrow rules. If the removal is intentional because the fleet is fully migrated, delete the stale promise in `egress-rules-service.ts:62-64` in the same PR, since it is the only stated guarantee for a dropped projection.

Reported by typescript-engineering.

packages/api-server/src/apps/api-server/trpc/ws.ts
```diff
@@ -0,0 +1,134 @@
...
+    );
+    const nudgeInMs = Math.max(closeInMs - RECONNECT_NUDGE_BEFORE_MS, 0);
+    const timers = [
+      setTimeout(() => {
+        // The protocol's reconnect notification — the same frame
+        // broadcastReconnectNotification sends, scoped to one client.
+        // tRPC routes incoming frames on `method`, so a `type`-keyed frame
+        // is silently dropped by wsLink (TRPCReconnectNotification is
+        // `{ id, method: "reconnect" }`).
+        ws.send(JSON.stringify({ id: null, method: "reconnect" }));
```
dam-code-guardian, line 66: ✅ Verified fixed. Checked against the pinned `@trpc/server` 11.16 adapter: `broadcastReconnectNotification` builds exactly `{id: null, method: 'reconnect'}`, and on the client `wsClient` routes any frame with `'method' in message` to `handleIncomingRequest`, which calls `this.reconnect(...)` for `method === 'reconnect'`. The nudge now lands.

Also confirms the 🟡 above it can be downgraded: `reconnect()` does `activeConnection.close()` then `.open()`, and `open()` re-sends `connectionParams` through `resultOf` — which re-invokes the callback form — so the new socket runs `createContext` and `deps.authenticate` again. An API-key connection therefore does re-authenticate at the deadline; `API_KEY_REAUTH_MS` bounds revocation latency as the comment claims.

dam-code-guardian (dismissed):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `5db70b3`

## PR #3258: refactor(api-server): tRPC-over-WebSocket transport + one composition root
**Author:** jezekra1 | **Branch:** `refactor/ui-live-updates#locki-lpaitkhu` → main | **Changes:** +3173 −2750 (43 files)

### Summary
Complete re-review at `5db70b3` (label-triggered). The previous 🔴 is fixed correctly, and all four `terms/compose.js` boundary bypasses now import from `index.js`. Both remaining findings are about the unrelated L7-migration removal riding in this PR.

### Changes since last review
Previous HEAD: `c8a5d2f` (2026-08-10T16:12:40Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** the reconnect nudge now sends `{id: null, method: "reconnect"}` (`trpc/ws.ts:66`) — verified against the pinned `@trpc/server` 11.16 adapter and `wsClient.handleIncomingRequest`, which routes `method: "reconnect"` into `reconnect()`.
- ✅ **Fixed:** `IsAcceptedPort` now imports from `modules/terms/index.js` at all four sites (`trpc/ws.ts:8`, `deps.ts:45`, `agent-proxies/upgrade.ts:7`, `admission/terms.ts:2`).
- ✅ **Fixed:** `oauth-routes.ts` moved to the module root and is exported from `connections/index.ts`.
- 🔁 **Still present:** API-key connections get a deadline, not a re-authentication (`trpc/ws.ts:55`) — the 4401 close does make `wsLink` re-open and re-send `connectionParams`, so the token is re-verified on the new socket; the interval bounds revocation latency as the comment claims. Downgraded to 🟢 on that evidence.
- 🆕 **New:** the L7 backfill removal breaks the documented self-heal path (below).

### Findings
- 🟡 **Warning:** deleting `createL7PromotionBackfill` leaves `spec.l7Hosts` reachable only through a rule mutation, and `egress-rules-service.ts:62-64` still promises the opposite — "a missed promotion also heals on api-server boot (the startup backfill re-projects every agent)". After a failed CR patch (conflict retries exhausted), that agent's promoted host stays unprojected until someone edits a rule; the controller then renders no L7 chain for it, so path-specific rules silently stop being enforced as narrow. Reported by typescript-engineering.
  **Fix:** retire a one-shot migration only once a mutation-independent path keeps the derived state converged — keep a boot-time or periodic reconcile over agents with active narrow rules, and update the comment that names the retired path.
- 🟡 **Warning:** `docs/architecture/security-and-credentials.md:459` still describes the deleted startup migration as live ("retires the markers"), and this PR does not touch that file. Verified PR-introduced: `main` carries both the migration and the paragraph. Reported by doc-drift.
  **Fix:** when a PR removes a subsystem behavior, delete or rewrite every architecture-page sentence describing it in the same PR — removals drift docs exactly as additions do.
- 🟡 **Warning:** the body's headline "**No behavior change to existing surfaces**" is contradicted by the same body's L7-migration bullet. Boot no longer runs a one-shot production data migration, which is unrelated to #3170 and cannot be reverted independently while bundled here. Reported by issue-fit.
  **Fix:** a blanket no-behavior-change claim must cover every change in the diff; move an unrelated production-migration removal to its own PR where its safety argument stands alone.
- 🟢 **Suggestion:** the WS door refuses the whole connection when terms are stale, while the HTTP door exempts `terms.*` so a stale user can accept (`trpc/ws.ts:98` vs `app.ts:98`). Harmless today — no client uses the socket and the UI accepts over HTTP — but it makes the "1:1 alternative" claim conditional, and terms are checked once per connection rather than per call. Worth stating as a deliberate limit in the body.
- 🟢 **Suggestion:** `createUpgradeHandler` puts `new URL(...)` outside the `try` its own comment relies on (`agent-proxies/upgrade.ts:150`). A malformed `Host` header throws (verified: `Host: [` and `a:b:c` both throw `Invalid URL`) and rejects unhandled from a listener Node does not await. The line is byte-identical to `main:1121` and `terminal-relay.ts:85`, so it belongs in a separate hardening issue covering both.

🔁 1 finding from the previous review still present (see review at `c8a5d2f`) — mirrored above.
- 🟡 **Warning:** `security-and-credentials.md:132` still says clients send the JWT "on every tRPC and ACP call". The page owns authentication in depth, so first-frame `connectionParams` auth and credential-bounded connections are documented nowhere.
  **Fix:** when code changes how a credential is presented or how long it stays valid, update the page that owns authentication in depth, not only the topology summary.
- 🟢 **Suggestion:** `platform-topology.md:43` records rollout stage ("the transport every client uses today", "not yet consumed by any client"); both clauses go false at the first follow-up.

### Issue Fit
Linked issue: #3170 (plain `Refs`, no closing keyword — the PR declares itself a foundation slice, so #3170's `Done when` items are correctly owed to follow-ups).
- 🟡 **Warning:** the no-behavior-change claim — mirrored above.
- 🟢 **Suggestion:** the ADR text requiring cross-cutting gates at the tRPC layer lives in **#3205, which is still open and unmerged**, so it is not yet a binding decision — but this PR's split gates (Hono middleware for HTTP, a re-implementation in `createContext` for WS) are exactly what that ADR forecloses. Worth resolving before #3205 merges.

🔁 1 finding from the previous review still present (see review at `c8a5d2f`).
- 🟡 **Warning:** the L7 convergence gap — mirrored above, verified against both projection call sites.
- 🟢 **Suggestion:** `ApiServerDeps` promotes module internals (`k8sClient`, `agentsRepo`, `templatesRepo`, `connectionsBoot`, `userDirectory`, `apiKeysModule`) to app-level contract fields imported from `modules/*/compose.js` and `modules/*/infrastructure/*.js` (`deps.ts:19,23,30,38,44`). On `main` these were locals; a single composition root inherently needs them, so re-exporting the types from each module's `index.ts` gets the boundary without losing the root.
- 🟢 **Suggestion:** the module-public-API rule was applied to `connections` only — channels routes still import from `modules/channels/infrastructure/*`, and `modules/channels/` has no `index.ts`.

### Verdict
REQUEST_CHANGES — the transport and composition-root work is sound and the previous 🔴 is properly fixed, but the bundled L7-migration removal breaks a self-heal path the code still documents, and drifts two docs.

---

tomkis (approved):

PR #3265, opened 2026-08-11
feat(ui): add a docs entry point to every view

### LINKED ISSUE #3113: UI - Access product docs from the DAM interface
## Context

A user asked to be able to get to the DAM docs (ibm.biz/dam-docs) from within the DAM GUI. Today the docs live at a URL you have to already know — there's no path to them from the product itself.

## Problem

When a user is inside DAM and gets stuck or wants to understand a feature, there's nothing in the interface that points them to the documentation. They have to already know the docs exist and have the link on hand — otherwise they leave to search, ask a teammate, or give up. Documentation only helps if people can find it at the moment they need it.

## Goal

From anywhere in the DAM GUI, a user can reach the product docs in one obvious action, without leaving to hunt for the link.

## User Stories

- As a user, I want to open the docs from within DAM so that I can get help without leaving the product to find the link.
- As a user, I want the way in to the docs to be visible from the main interface so that I know documentation exists in the first place.

## Open Questions

- Where does the entry point live — a persistent help/nav item, a "?" menu, contextual links, or some combination?

kapetr:
A help button sits in the lower right corner on every view. Clicking it opens a small panel — "Need help?", a line about what is there, and a link to the docs. Before this, you had to already know the docs existed and have the URL to hand.

On mobile the button floats above the rail's bottom bar, stepping up again when the connection banner is showing. The one place it is hidden is the mobile conversation pane: chat renders no bottom bar and puts its composer in that corner, where the button overlapped the send control and took its taps.

Two notes for review:

- Adds `@radix-ui/react-popover`. The panel holds a heading, prose and a link, so it is not a menu — the existing dropdown would give it menu semantics and arrow-key navigation it should not have. The new primitive also supplies the tail, Escape and outside-click.
- The link points at the docs root on `pages.github.ibm.com`, matching the four doc links already in `constants.ts`. The issue names `ibm.biz/dam-docs`; say if you want the short link instead.

Closes #3113

### DIFF
--- a/packages/ui/src/components/docs-launcher.tsx
+++ b/packages/ui/src/components/docs-launcher.tsx
@@ -13,9 +13,12 @@ import { cn } from "@/lib/utils";
 
 import { DOCS_URL } from "../constants.js";
 import { getApiHealthSnapshot, subscribeApiHealth } from "../lib/api-health.js";
+import { useStore } from "../store.js";
 
 export function DocsLauncher() {
   const titleId = useId();
+  const view = useStore((s) => s.view);
+  const chatSurface = view === "chat" || view === "knowledge-base-chat";
   const banner =
     useSyncExternalStore(subscribeApiHealth, getApiHealthSnapshot) !==
     "connected";
@@ -29,6 +32,7 @@ export function DocsLauncher() {
           aria-label="Help"
           className={cn(
             "fixed right-3 z-nav h-[34px] w-[34px] md:right-4",
+            chatSurface && "hidden md:inline-flex",
             banner
               ? "bottom-[124px] md:bottom-[60px]"
               : "bottom-[80px] md:bottom-4",
--- a/packages/ui/src/components/docs-launcher.tsx
+++ b/packages/ui/src/components/docs-launcher.tsx
@@ -16,8 +16,6 @@ import { getApiHealthSnapshot, subscribeApiHealth } from "../lib/api-health.js";
 
 export function DocsLauncher() {
   const titleId = useId();
-  // The connection banner is a 44px bar in this same corner and outranks this
-  // button, so step over it while it is up.
   const banner =
     useSyncExternalStore(subscribeApiHealth, getApiHealthSnapshot) !==
     "connected";
@@ -30,8 +28,10 @@ export function DocsLauncher() {
           size="icon-sm"
           aria-label="Help"
           className={cn(
-            "fixed right-4 z-nav hidden h-[34px] w-[34px] md:inline-flex",
-            banner ? "bottom-[60px]" : "bottom-4",
+            "fixed right-4 z-nav h-[34px] w-[34px]",
+            banner
+              ? "bottom-[116px] md:bottom-[60px]"
+              : "bottom-[72px] md:bottom-4",
           )}
         >
           <Help size={16} />
--- a/packages/ui/src/components/docs-launcher.tsx
+++ b/packages/ui/src/components/docs-launcher.tsx
@@ -1,0 +1,60 @@
+import { ArrowRight, Close, Help } from "@carbon/icons-react";
+
+import { Button } from "@/components/ui/button";
+import {
+  Popover,
+  PopoverClose,
+  PopoverContent,
+  PopoverTrigger,
+} from "@/components/ui/popover";
+import { externalLinkProps } from "@/lib/external-link";
+
+import { DOCS_URL } from "../constants.js";
+
+export function DocsLauncher() {
+  return (
+    <Popover>
+      <PopoverTrigger asChild>
+        <Button
+          variant="outline"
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/docs-launcher.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <PopoverTrigger asChild>
+        <Button
+          variant="outline"
+          size="icon-sm"
+          aria-label="Help"
+          className={cn(
+            "fixed right-3 z-nav h-[34px] w-[34px] md:right-4",
+            banner
+              ? "bottom-[124px] md:bottom-[60px]"
+              : "bottom-[80px] md:bottom-4",
```
PetrBulanek, line 38: `bottom-[80px]` assumes the mobile bottom bar is there, but the chat view renders `IconRail hideMobileBar` and puts its composer in that corner instead. At 375×812 the help button (329–363, 698–732) overlaps the send button (294–334, 720–760) by a few pixels, and since the launcher sits on `z-nav` above the content layer, it wins those taps — a user aiming at send can open the help popover instead. The button also visually collides with the send arrow. The mobile offset needs to account for the chat surface — a lower position when the bar is absent, or hiding the launcher on the mobile conversation pane, would both do.

PetrBulanek (approved):

packages/ui/src/components/docs-launcher.tsx
dam-code-guardian, line 34: 🟡 **Warning:** the PR body says "the button is hidden on mobile", but the launcher carries no visibility gate — it only shifts position (`bottom-[80px] md:bottom-4`), so it renders at every width. The `80px` offset assumes the rail's bottom bar, and the chat and knowledge-base-chat shells render `IconRail hideMobileBar` with the composer in that corner instead: the button then sits over the composer's control row, and `z-nav` (30) above the content layer (10) means it wins those taps. PetrBulanek's thread on line 34 measures the overlap at 375×812.

**Fix:** a globally-fixed control must derive its offset from the bottom chrome actually present in the current view, not from a constant assumed for one shell — read the shell state (as `IconRail` does) so every branch that hides the bottom bar drops the matching offset. If hiding it on mobile is the intended behavior, add the breakpoint gate the body describes; either way make the code and the body agree.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `373e708`

## PR #3265: feat(ui): add a docs entry point to every view
**Author:** kapetr | **Branch:** feat/3113-docs-entry-point → main | **Changes:** +519 −0 (6 files)

### Summary
Adds a fixed help button in the lower-right corner of the app shells. It opens a Radix popover with a link to the docs root, and vendors a new `popover` primitive for it.

### Findings
- 🟡 **Warning:** the launcher has no mobile visibility gate, so it overlaps the bottom chrome (`packages/ui/src/components/docs-launcher.tsx:31`) — see the inline comment.
- 🟢 **Suggestion:** `DocsLauncher` renders only inside `MainApp`, so the `terms`, `telegram-bind` and `slack-bind` views return before it and get no docs entry point (`packages/ui/src/app.tsx:51`). The title says "every view" and the issue asks for "anywhere in the DAM GUI". Mount it above the early returns, or name the excluded views in the body. (Source: issue-fit.)
- 🟢 **Suggestion:** `DOCS_URL` hard-codes the docs host root, but issue #3113 names `ibm.biz/dam-docs` (`packages/ui/src/constants.ts:1`). The body asks for a decision. The short link survives a host move, so it is the more durable target. (Source: issue-fit.)
- 🟢 **Suggestion:** the `124px` / `60px` offsets restate `ConnectionBanner`'s own geometry (`bottom-14` + `h-11`, `md:bottom-0`), and the `banner` flag restates its render condition (`packages/ui/src/components/docs-launcher.tsx:19-34`). The banner can change either one with no signal here. Publish the bottom-chrome heights once and let both components read them. (Source: react-ui-engineering.)
- ✅ **Looks good:** the rotated-square arrow reads as correct for `side="top"` only, but it holds for every placement. Radix's `PopperArrow` wrapper carries the per-side rotation, so the square's two bordered edges and its half-overlap stay outward after a collision flip.

✅ No findings.

### Issue Fit
Linked issue: #3113 "UI - Access product docs from the DAM interface" (closing keyword).
- The body/code mismatch on mobile visibility is the blocking 🟡 above.
- Coverage of the pre-shell views and the `ibm.biz` short link are the two 🟢 above.

✅ No findings.

- The composer overlap in the `hideMobileBar` shells is merged into the blocking 🟡 above.
- The duplicated banner geometry is the last 🟢 above.

### Verdict
COMMENT — the feature is sound and the primitive is correct, but the mobile placement contradicts the stated behavior and can steal taps from the chat composer.

---

Issue #3108, opened 2026-07-31
Untangle the session lifecycle in agent-runtime: regression tests first, then refactor


tomkis:
## What

Refactor the ACP session runtime in agent-runtime, the module that owns chat sessions: fan-out to channels, the idle reap, and the busy/idle status. In two steps, in this order:

1. **Regression tests first.** Pin the current behavior with tests that describe what the runtime does today: session open/close, fan-out to multiple channels, reap timing, background-work holds, the idle flag. This code is battle-tested and every touch risks new bugs, so the tests come before any restructuring.
2. **Then restructure.** Targets we already know hurt:
   - **Reaping is scattered.** Whether a session gets reaped is decided by per-session timers that are scheduled, rescheduled, and cleared in half a dozen places. Replace with one loop that periodically checks each session against a single "this session is done" predicate. A missed edge then heals on the next tick instead of leaking a session.
   - **The "done" predicate is duplicated.** The reap path and the status endpoint each combine the same conditions (active prompts, queued prompts, pending requests, background work) independently. Anyone adding a new hold condition has to remember both. One predicate, two callers.
   - **Fan-out is opaque.** The channel fan-out mechanism is the part of the module nobody on the team can confidently explain. The refactor should leave it readable, or at least documented at the level a reviewer needs.

## Why

The module works, but nobody understands it. Reviewing changes to it now requires reconstructing the high level from scratch (the review of #3044 needed a generated walkthrough artifact to get there), and the honest expectation is that once people trust the high level, nobody looks at this code again. That is how the next bug gets in. The recent background-work holds (#3044) also made the reap logic meaningfully more stateful, so the cost of the scattered timer design is growing.

## Done when

- Regression tests exist that pin today's session lifecycle behavior, written and merged before any restructuring.
- Session reaping is decided in one place, by one predicate that the status endpoint shares.
- Observable behavior is unchanged: same reap timing, same hold semantics from #3044, same status contract.

PR #3266, opened 2026-08-11
test(acp-runtime): add feature-oriented acp-runtime specs

tomkis:
Step 1 of #3108: pin the ACP runtime's current session lifecycle behaviour with tests before restructuring it. No behaviour changes; the runtime file only moves into its own directory.

The existing suite was coupled to the implementation. Its cases were named in internal vocabulary and reached frames by array position, so the planned refactor would break them without behaviour changing. The new suite describes behaviour only: each scenario drives the runtime through its public surface and asserts on what clients saw and what the harness received. One file per feature, about 25 scenarios covering connecting, isolation, joining, leaving, the machine client, and mass disconnect.

The old suite is removed here, replaced by the new one. Issues found while writing the tests are filed separately (#3268).


jezekra1 (approved):

packages/agent-runtime/src/modules/acp/services/acp-runtime/__tests__/acp-world.ts
```diff
@@ -0,0 +1,376 @@
...
+  connect(opts?: { viewer?: boolean }): Client;
+  /** The harness currently serving this pod. Throws if none has started. */
+  harness(): Harness;
+  /** Whether the subprocess has been started at all. */
+  harnessStarted(): boolean;
+  /** How many harnesses have been spawned over this world's life. */
+  harnessCount(): number;
+}
+
+export function createWorld(
```
dam-code-guardian, line 209: 🔴 **Critical:** `createWorld` never accepts or passes a `backgroundWork` registry, so the #3044 hold semantics are not exercised through the runtime anywhere. `status().backgroundWork` is never asserted either — only `status().idle`, in the machine-client file. `background-work-registry.test.ts` and `background-work-contract.test.ts` both import only `createBackgroundWorkRegistry`, so they cover the registry in isolation, never the runtime's use of it.

Issue #3108 makes this step-1 work explicitly: it lists "background-work holds" among the behaviours to pin, requires "same hold semantics from #3044, same status contract", and has step 2 give the status endpoint the shared reap predicate. Step 2 rewrites exactly this machinery, so it would land with no test over it.

**Fix:** Pin every hold condition that feeds the reap predicate — active prompts, queued prompts, pending requests, background work — through the runtime's public surface, and assert the full `status()` shape including `backgroundWork`, not just `idle`.

packages/agent-runtime/src/modules/acp/services/acp-runtime/__tests__/acp-world.ts
```diff
@@ -0,0 +1,376 @@
...
+  overrides: Partial<Omit<AcpRuntimeDeps, "spawnAgent">> = {},
+): World {
+  const harnesses: Harness[] = [];
+
+  const runtime = createAcpRuntime({
```
dam-code-guardian, line 214: 🟡 **Warning:** `createWorld` passes no `idleReapDelayMs`, so every one of the 28 scenarios takes the runtime default `0` (`acp-runtime.ts:257`) and reaps sessions inline. Production composes the runtime with `idleReapDelayMs: 3_000` (`compose.ts:66`).

So the `session/close` assertions in `acp-runtime-leaving.test.ts:194` and `acp-runtime-machine-client.test.ts:120` hold only in a configuration the product never runs: they take the `idleReapDelayMs <= 0` branch of `maybeCloseIdleSession`, while production always takes the `setTimeout` branch. The deleted suite's `debounces the idle reap and a new prompt within the window cancels it` covered that branch, and nothing replaces it — which matters because #3108 step 2 replaces these per-session timers with one loop.

**Fix:** A test world defaults to the same dependency values the composition root injects, and any deliberate divergence is covered by at least one scenario at the production value. Set `idleReapDelayMs` to match `compose.ts` and add a fake-timer scenario for the deferred reap and its cancellation.

packages/agent-runtime/src/modules/acp/services/acp-runtime/__tests__/acp-runtime-connecting.test.ts
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   * That wait cannot be open-ended, because env can fail to arrive at all: a
+   * failed `hello` is logged and swallowed, which leaves the pod up and Ready
+   * with nothing left that would ever open the gate.
+   *
+   * So the runtime gives up after a bound and starts anyway. A sandbox with no
+   * credentials is degraded, but the client gets an answer instead of a socket
+   * that never replies, and if env turns up later a recycle picks it up.
+   */
+  it("should stop waiting for env rather than leave the client hanging forever", () => {
```
dam-code-guardian, line 124: 🟢 **Suggestion:** This scenario pins the ~15s env wait, which is the behaviour issue #3268 files as defective and proposes to change. Naming #3268 in the docstring would let whoever fixes that issue find the test that encodes the current contract, instead of discovering it as a failure.

packages/agent-runtime/src/modules/acp/services/acp-runtime/__tests__/acp-runtime-isolation.test.ts
```diff
@@ -0,0 +1,241 @@
...
+    const alice = world.connect();
+    alice.send(frames.newSession(1));
+    world.harness().replyTo("session/new", { sessionId: ALICE_SESSION });
+
+    const bob = world.connect();
+    bob.send(frames.newSession(1));
+    world.harness().replyTo("session/new", { sessionId: BOB_SESSION });
+
+    // Both ask under id 7, and both questions are in flight at once.
+    alice.send(frames.prompt(7, ALICE_SESSION, "did the tests passi rea?"));
```
dam-code-guardian, line 165: 🟢 **Suggestion:** The prompt text reads `"did the tests passi rea?"` — a stray keystroke. No assertion in this scenario reads the text, so nothing fails, but neighbouring scenarios use `"did the tests pass?"` and this file is the pattern others get copied from.

```suggestion
    alice.send(frames.prompt(7, ALICE_SESSION, "did the tests pass?"));
```

dam-code-guardian (dismissed):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `bfe383b`

## PR #3266: test(acp-runtime): add feature-oriented acp-runtime specs
**Author:** tomkis | **Branch:** test/acp-runtime-connecting-spec → main | **Changes:** +1706 −3055 (13 files)

### Summary
The PR replaces the 3044-line mechanism-coupled ACP runtime suite with 28 behaviour-oriented scenarios across six feature files and a shared `acp-world.ts` harness. The specs themselves are strong — the fakes match both ports exactly, accessors resolve frames by method rather than by index, and each scenario argues why the behaviour is the runtime's job — but the replacement drops coverage that issue #3108 names as required before the refactor.

### Findings
- 🔴 **Critical:** The suite deletes 84 scenarios and adds 28, so roughly two thirds of `acp-runtime.ts` has no test anywhere in the repo. Verified absent from every live spec: `session/resume` hot and cold bootstrap paths, the `PROMPT_QUEUE_CAP` error path, `platform/deleteSession` tombstoning, `session/list` `_meta.platform` enrichment, `promptAccepted`/`promptStarted` delivery notifications, the `sessionCapabilities.close` gate, orphan-TTL expiry, direct-turn framing, and the `<clipped-conversation>` sentinel — each still live in the runtime. Reported by my diff review, `Issue Fit`, and `TypeScript Engineering`. (`packages/agent-runtime/src/__tests__/unit/acp-runtime.test.ts:1` deleted)
  **Fix:** Before deleting a suite, every behaviour the old file pinned must be re-pinned in the new suite or another live spec — a deletion is a "replacement" only when coverage is preserved. Scope: port the resume, queue-cap, delete-session, list-enrichment, prompt-notification, close-capability, orphan-TTL, direct-turn and log-truncation scenarios, or keep the old file until they land.
- 🔴 **Critical:** Issue #3108 makes step 1 pin "reap timing" and "background-work holds", and its Done-when requires "same hold semantics from #3044, same status contract" — but no new scenario exercises either. `createWorld` never passes a `backgroundWork` registry, and `status().backgroundWork` is never asserted (only `status().idle`). The two registry specs cover the registry in isolation, never the runtime's use of it. Step 2 replaces exactly this machinery, so the refactor proceeds with no safety net over it. Reported by my diff review and `Issue Fit`. (`packages/agent-runtime/src/modules/acp/services/acp-runtime/__tests__/acp-world.ts:199`)
  **Fix:** Pin every hold condition feeding the reap predicate — active prompts, queued prompts, pending requests, background work — through the runtime's public surface, and assert the full `status()` shape including `backgroundWork`.
- 🟡 **Warning:** `createWorld` never sets `idleReapDelayMs`, so all 28 scenarios run at the default `0` and reap inline, while production composes `3_000`. The `session/close` assertions therefore pass only under a configuration the product never uses, and the timer branch of `maybeCloseIdleSession` is unexercised now that the debounce test is deleted. Reported by `TypeScript Engineering`. (`packages/agent-runtime/src/modules/acp/services/acp-runtime/__tests__/acp-world.ts:204`)
  **Fix:** A test world must default to the dependency values the composition root injects; align `idleReapDelayMs` with `compose.ts` and add one fake-timer scenario for the deferred reap and its cancellation.
- 🟡 **Warning:** Two issue-anchored regression guards disappear with no successor — `#2965` (background work holds an idle session) and `#1999` (`session/list` last real message time, stable across a reload). No new scenario carries either anchor, so both bugs can return silently. Reported by `Issue Fit`. (`packages/agent-runtime/src/__tests__/unit/acp-runtime.test.ts:1103` deleted)
  **Fix:** When a regression test naming an issue is moved or rewritten, the replacement scenario keeps the `#<n>` anchor so the guard stays traceable to the bug it prevents.
- 🟢 **Suggestion:** `acp-runtime-connecting.test.ts:124` pins the fifteen-second env wait, which is the very behaviour #3268 files as defective. Naming #3268 in that scenario would let whoever fixes the issue find the test encoding the old contract.
- 🟢 **Suggestion:** `acp-runtime-isolation.test.ts:165` sends the garbled prompt text `"did the tests passi rea?"`, which no assertion reads, in a shared spec file that other scenarios are copied from.
- ✅ **Looks good:** The rename of `acp-runtime.ts` into its own directory is complete — no importer of the old path survives, and the three updated import sites are the only ones. The fakes implement `AgentProcess` and `ClientChannel` exactly, with no surface omitted.

✅ No findings.

### Issue Fit
See the 🔴 coverage findings above, which this skill reported independently. Its remaining points:
- 🟡 **Warning:** The PR body says the old suite is "replaced by the new one", but the new suite is a strict subset — undeclared scope, since #3108 asks to add pins, never to drop them.
  **Fix:** A test-replacement PR either carries each removed behaviour area forward or lists the dropped areas in the body with the reason.

The two findings this skill contributed — the production/test `idleReapDelayMs` divergence and the coverage deletion — are carried above. No further defects: the spec files type-check cleanly against the ports, and the harness accessors are index-free as their own docstring requires.

### Verdict
REQUEST_CHANGES — the specs are well built, but they drop the reap-timing and background-work pins that issue #3108 requires before step 2, and two thirds of the runtime's asserted behaviour goes untested.

---

packages/agent-runtime/src/modules/acp/services/acp-runtime/__tests__/acp-world.ts
tomkis: Added acp-runtime-staying-awake.test.ts in b3eb11e2. It pins the queued prompt, pending request and background work holds through the runtime, and asserts the full status() shape including backgroundWork. The active prompt hold was already pinned in the machine-client spec, so I didn't duplicate it.

One correction: createWorld already passes backgroundWork through its overrides param (same as sessionMetadata), so no world change was needed. The tests call registry.report() directly, which is what the report endpoint in server.ts does.

packages/agent-runtime/src/modules/acp/services/acp-runtime/__tests__/acp-world.ts
tomkis: Fixed in 4766ce3e. The world now defaults idleReapDelayMs to 3s like compose.ts, the scenarios that watch a release advance fake timers past the window, and there is a new leaving scenario where a quick re-attach within the window keeps the session warm and only a later departure releases it.

Issue #3022, opened 2026-07-28
Epic - Close the Skills usability gaps


tomkis:
## Context

Skills works end to end — connect a source, install into a sandbox, author, publish. But the surface is full of gaps: previews that don't open, buttons that are disabled, information the design promised but never got wired up, and setup you redo by hand for every sandbox.

## Problem

Each gap is small enough to defer on its own, which is exactly why they've accumulated. Anyone configuring an agent runs into them, so that's everyone. Individually they're papercuts. Together they're what makes Skills read as half-built rather than something people trust and reach for.

## Goal

Skills a user can trust: what the interface says is true, every visible action does something, and configuring a sandbox's skills doesn't mean redoing work from memory.

Delivered by working through the attached issues rather than by any single change. New usability findings get attached here instead of filed loose.

## Scope

In: the usability, trust, and repetition gaps in the existing connect → install → author → publish flow. The specific items are the issues attached to this epic — deliberately not restated here, so this stays the one place describing why the work matters instead of a second copy that drifts out of date.

Out: new Skills capabilities. This epic is about making what exists feel finished, not extending what it can do.

sarahmilleribm:
HTML design

[issue-3022-prototype.html](https://github.com/user-attachments/files/30796865/issue-3022-prototype.html)

PR #3243, opened 2026-08-10
feat(skills): search, bulk-toggle, and reuse skill selections across sandboxes

PetrBulanek:
Picking skills stops being a per-sandbox chore done from memory.

Today a source with 40 skills means expanding its card, reading 40 truncated descriptions, and flipping toggles one at a time — then doing it again on the next sandbox tomorrow. After this there is one search box across every connected source, skills can be enabled or disabled several at once, and a selection can be saved as a named skill set and added to any other sandbox.

That matters because skills are the main lever for shaping what an agent is good at, and sandboxes are meant to be cheap to create. When configuring skills is manual per-sandbox work, people either skip it or keep one long-lived sandbox to avoid redoing it. Both defeat the point of disposable sandboxes.

Adding a set is always additive: it turns skills on alongside what is already there, and never turns anything off.

Phase 2 of #3022.

Closes #3023

## Scope delivered, and what was descoped

The issue names two surfaces: the sandbox Skills page **and the new-sandbox flow**. This delivers the Skills page half. The wizard is untouched — `applyToAgent` needs an agent that exists, so carrying a selection through creation means a post-create apply step, which is its own slice.

Two narrower gaps: skill rows carry no per-skill multi-select — bulk is per-source Enable/Disable all — and the bulk control hides while a search filter is active, so a search can't scope a bulk action. The `applyBatch` path both would need is in; what is missing is surface control design, which the settled #3022 design prototype does not specify either — its bulk control is per-source and acts on the whole source regardless of an active search.

#3023 is closed on that basis rather than held open for the remainder; the closing comment there records the split. The three carried items are filed as follow-ups after #3208 (the design rebuild) lands, so they are designed against the surface they will live on.

## Behavior changes beyond the headline features

- The `state` read's ghost-row reap is now gated on the agent's outbox being **settled**, and a failed settled read defers the reap rather than failing the read. Batching made the pre-existing race — a freshly installed skill read as a ghost mid-apply — N apply-windows wide instead of one, which is what forced the gate. The gate applies to every sandbox, not only bulk actions: out-of-band skill deletions are now noticed only once the pod has caught up.
- The shared `warning` badge variant reads its color from a new `--c-warning-fg` token. Both themes resolve to the exact previous values, so nothing renders differently today — but light-theme warning text app-wide (chat "Reconnecting", contribution-failure badges, wizard cards) now moves with that one token.
[...]

Issue #3023, opened 2026-07-28
Search, bulk-toggle, and reuse skill selections across sandboxes


tomkis:
## Problem

The Skills page shows each connected source as a card you expand and read row by row. Every skill is its own toggle, and the selection you build lives only on that one sandbox.

Connect a source with 40 skills and want 6 of them: you expand the card, read 40 one-line truncated descriptions, and flip 6 toggles one at a time. Spin up a second sandbox tomorrow and you do the whole thing again from memory.

This matters because skills are the main lever for shaping what an agent is good at, and sandboxes are supposed to be cheap to create. When configuring skills is a manual per-sandbox chore, people either skip it or hold onto one long-lived sandbox to avoid redoing the work. Both undercut the point of disposable sandboxes.

## Proposed solution

Three changes to how a user picks skills:

- **Search across everything connected.** One search box over all skills from all connected sources, matching name and description, so finding a skill doesn't mean expanding every card and eyeballing it.
- **Act on several at once.** Select multiple skills and turn them on or off in a single action. Same for skills flagged as out of date: update all of them together instead of one pill at a time.
- **Carry a selection to a new sandbox.** Start a sandbox with the same skills as an existing one, or with a saved default set, instead of re-picking by hand every time.

### Scope

The sandbox Skills page and the new-sandbox flow.

Search covers skills from connected sources. Whether it also spans skills created inside the sandbox depends on #2828, since today those two are mixed together.

### Dependencies

Related to #2828 (telling image-baked skills apart from user-authored ones), which decides whether sandbox-created skills can be sensibly included in search and bulk actions.

### Out of scope

- A platform-hosted skill catalog, ratings, or discovery beyond what the user has connected. Sources stay external git repos.
- Making descriptions readable at a glance. Reading a skill in depth is the preview's job (#2824 extends it).

### Open questions

- Is "carry a selection" a per-user default, a copy-from-another-sandbox action, or something declared on the template? Each has different implications for what happens when the source changes later.
- Naming: is "skill set" the right word for a reusable selection? Pushback welcome.

PetrBulanek:
Delivered by #3243 (merged as `c5d398d5`). Closing, with the remainder split out rather than held here.

**Shipped against this issue's three asks:**

- *Search across everything connected* — one box over every connected source, matching name and description. It also spans skills created in the sandbox and image-shipped ones, which this issue left conditional on #2828 (landed).
[...]

PR #3267, opened 2026-08-11
feat(ui): rebuild the Skills surface on the new design

### LINKED ISSUE #3208: Rebuild the Skills surface on the new design
## Context

Most issues in this epic are behavior: previews that open, state that survives a stopped sandbox, selections you can search and reuse. Each one lands inside the layout that already exists. A new design for the Skills surface is in progress and arrives after that backend work.

## Problem

Today's Skills surface grew by accretion. Every issue added a row, a badge, or a modal to what was already there. It works, but nobody designed it as a whole, so it reads denser and busier than the new design. Folding the redesign into each backend issue would smear it across many pull requests and leave the surface half-migrated in between.

## Goal

The Skills surface matches the new design, and every capability the epic's other issues delivered still works.

## Scope

In: the visual and interaction rework of the Skills surface onto the new design, once the design lands and the backend issues are done.

Out: new capability. If the design implies behavior we don't have yet, that's its own issue attached to this epic.

[Design prototype](https://github.com/user-attachments/files/30827535/issue-3022-prototype.html)

PetrBulanek:
Rebuilds the Skills surface on the settled #3022 design across all four sandbox states (Running, Running · empty, Stopped, Never run), both themes, and every overlay — search, drift banner, source cards, set modals, the skill preview dialog, and the read-only stopped and never-run panels — while keeping every capability the epic shipped.

One contract change: the scan result now carries whether a repository is public, which is what lets a source card mark a private one honestly. Everything else is a restyle or a rearrangement.

Closes #3208

## What changed, by area

- **Page shell** — provenance order is Created → Sourced from GitHub → Included with sandbox image; one `Add source`; the header sits closer to the first group than the groups sit to each other.
- **Rows** — created-here and source rows carry a name and their actions, nothing else. Descriptions moved into the preview, which is where you go when the name isn't enough.
- **Source cards** — name, count, visibility, repo URL, then freshness with its own re-scan control, bulk toggle and kebab. A scan in flight replaces the timestamp rather than adding a spinner beside it.
- **Search** — the per-source bulk control now stays available while filtering and acts on exactly the visible matches, labelled and confirmed with that count.
- **Preview dialog** — one design for source-backed, created-here and image skills, with provenance chips, a file strip that flips between rendered and raw, and the on/off control at eye level with the name.
- **Stopped** — a dated snapshot of what was on at the last run over the live source list, instead of a dimmed copy of the running page. Search, bulk actions and toggles are left out rather than greyed out.
- **Never run** — says only what is true before first boot, and offers the one thing that works without a pod: connecting a source.
- **Stale model** — a stopped sandbox whose saved model is no longer offered says so, with a marker on Sandbox Setup.

## Follow-ups filed

- #3283 — choosing skills before a sandbox first starts (the capability the never-run panel deliberately does not fake)
- #3284 — a home for skill sets
- #3285 — applying a skill set while creating a sandbox

## Where this differs from the design, and why

The design prototype was followed closely — where it was followed exactly, there is nothing to say here. These are the places it wasn't, each a deliberate call.

| # | Design shows | Shipped | Why |
|---|---|---|---|
| 1 | A publish whose pull-request state can't be read is badged `Published` | `Submitted` | Publishing is the action; the badge reports what became of the pull request, so `Draft`/`Open`/`Merged`/`Closed` follow the design and GitHub exactly. The unresolved case can't: `Published` would give the weakest claim — "we recorded a publish but can't read it" — the strongest word |
| 2 | Publish badge is a bare label | Badge stays a link to the pull request, with its tooltip | Losing the link removes the only way to reach the PR from the row |
| 3 | Publish only in the drawer, unconditional | Kebab and drawer, offered when no record exists or the PR closed | Publishing is the core loop and needs to be discoverable; the gating rules are deliberate |
| 4 | Never-run: "pick skills now — they'll be applied the first time the sandbox starts" | "Not known yet — start the sandbox to see and configure its skills" | No such mechanism exists, and every mutating path wakes the pod the panel is telling you to start. Filed as #3283 |
| 5 | Never-run source rows have no kebab | Keep the kebab | Sources are account-scoped and administrable without a pod; the design keeps the kebab in its own Stopped state, so this is an inconsistency in it |
| 6 | No source error state | The server's verdict plus "Manage connections", restyled into the new card | A real state the redesign has to survive |
| 7 | Empty sandbox hides the image group | Image group renders whenever the image ships skills | Real images bake skills in; hiding them hides active truth |
| 8 | — (assumes visibility is known) | `visibility` added to the scan result | The only honest way to render the Private badge. Proved by which branch answered: the public archive serves only public repos, and falling through its 404 into the pod proves private. A non-GitHub host is never asked, so it stays unknown and badges nothing |
| 9 | Bulk keeps its `Enable all` label while a search is active, acting on the whole source | Stays visible, scoped to the visible matches, labelled with the count (`Enable 4 matching`) | The trap was the label, not the button: naming the count makes the promise and the action agree. Its confirm follows the same scope |
| 10 | Drawer shows a static `Not installed` for a source skill that is off | `Off` plus a live switch | Installing from the preview you just read is the flow; it costs nothing, being the row's own mutation |
| 11 | Drawer offers Download for a source-backed skill | Download on local skills only | No download path exists for a skill that isn't on disk; the GitHub link is the way out |

**Not a deviation — a bug in the prototype.** Its search hides skill rows and whole groups but never inspects source cards, so a card matching nothing keeps its header, while a query matching nothing at all hides the group *and* its cards. The two cases contradict each other. Zero-match cards are hidden.

## Sub-issues

- [x] 01 — Page shell and grouping
- [x] 02 — Created-here rows
- [x] 03 — Source cards
- [x] 04 — Set modals restyle
- [x] 05 — Skill detail drawer
- [x] 06 — Stopped state
- [x] 07 — Never-run and Running · empty
- [x] 08 — Stale-model callout
- [x] 09 — Sidebar summaries, docs, final sweep


packages/ui/src/components/modal.tsx
```diff
@@ -127,23 +132,28 @@ export function DialogHeader({
...
-              data-testid={closeTestId}
-              /* Pulled out of the header's padding so it sits 16px off the
-                 panel corner at both breakpoints, rather than lining up with
-                 the title. */
-              className="-mt-1 -mr-1 shrink-0 text-muted-foreground md:-mt-3 md:-mr-3"
-            >
-              <Close size={16} />
-            </Button>
+          {(actions || onClose) && (
+            <div className="flex shrink-0 items-center gap-3">
```
dam-code-guardian, line 106: 🔴 **Critical:** This renders `{actions}` before the close button, which makes the preview dialog's install/uninstall switch the first focusable node in the dialog — so opening a source-backed skill preview lands initial focus on it, and the first Enter uninstalls the skill.

The focus trap at `:298` skips only `data-dialog-close` before taking `focusables[0]`, and its own docstring asks that the attribute be kept on "any other close affordance added here". This PR adds a slot for a *mutating* control instead, which that exclusion never anticipated: the `Switch` passed as `stateControl` (`skill-render-modal.tsx:86` → `skill-markdown-modal.tsx:121`) is a plain `<button>` with no `disabled`, so it matches `FOCUSABLE_SELECTOR` and sorts ahead of the ✕ in DOM order.

**Fix:** The trap's initial-grab exclusion must cover every state-mutating control that can sit in a header slot, not just the close button — mark such controls so they are skipped exactly as `data-dialog-close` is, or give the header an explicit initial-focus target, so opening a dialog can never arm a mutation with the first keystroke.

(reported by react-ui-engineering)

packages/ui/src/modules/sandboxes/components/skills/skills-surface.tsx
```diff
@@ -132,131 +176,136 @@ export function SkillsSurface({
...
+      onStart={() => agentId && wakeAgent.wake(agentId)}
+      onRescan={(src) => void surface.refreshSource(src.id)}
+      onRemove={(src) => void removeSourceWithConfirm(src)}
+    />
+  );
 
   return (
     <div
       {...surfaceDropProps}
       className={cn(
```
dam-code-guardian, line 177: 🟡 **Warning:** The blanket `pointer-events-none` / opacity gate is gone, but the two panels that replace it are gated on `standaloneSnapshot !== undefined` (`:126`) and `hasRun === false` (`:182`) — which leaves a real state uncovered.

A sandbox that is stopped and *has* run but has no recorded snapshot (`skills.state` returns no `standaloneSnapshot` when nothing was recorded) matches neither panel and falls through to the live running branch. There, source-card toggles and drift `Update` links render enabled, because `disabled` is `!agentId || isError` and excludes `readOnly` (`skill-sources-section.tsx:107`). Clicking one hits the hook's `readOnly` early-return and does nothing, with no feedback — the greyed-out-control failure the stopped panel's own copy says it avoids.

**Fix:** When a wholesale non-interactive mode replaces a blanket pointer/opacity gate, every state branch must be covered exhaustively, and each mutating control must take the same read-only flag its handler guards on — never leave a fallthrough branch rendering live controls over a guarded mutation.

(reported by issue-fit)

packages/ui/src/modules/sandboxes/components/skills/skills-surface.tsx
```diff
@@ -132,131 +176,136 @@ export function SkillsSurface({
...
-  const isEmpty =
-    !readOnly &&
-    sourcesLoaded &&
-    stateLoaded &&
-    sources.length === 0 &&
-    standalone.length === 0;
+  // Running with nothing in it is no longer a special page: each group shows
+  // its own empty panel, so the image group still renders when the image ships
+  // skills. Only the two non-running states replace the surface wholesale.
+  const neverRunPanel = readOnly && hasRun === false && (
```
dam-code-guardian: 🟡 **Warning:** `hasRun === false` is also what a still-loading snapshot query reports, so this renders the never-run panel — and its claim "This sandbox hasn't run yet" — on first paint for a *stopped* sandbox that has run many times, before flipping to the stopped panel.

`hasRun` collapses the unknown arm at `harness-config.ts:95` (`recorded?.hasRun ?? false`), so the loading state is indistinguishable from the negative answer, and this branch is tested before the snapshot one. The comment above states "`hasRun: false` is the one honest signal that no snapshot is even possible", but the value conflates "never ran" with "not loaded yet". The same flattening drives the nav's "Not known yet".

**Fix:** A tri-state fact (unknown / false / true) must keep its unknown arm across the layer that consumes it — gate a panel that asserts a negative fact on the query having resolved, never on a boolean whose loading default is the assertion itself.

(reported by issue-fit and react-ui-engineering)

packages/ui/src/modules/sandboxes/components/skills/skill-render-modal.tsx
```diff
@@ -43,7 +65,44 @@ export function SkillRenderModal({
...
+      stateControl={
+        <span className="flex items-center gap-2">
+          <span className="text-sm font-medium text-muted-foreground">
+            {installed ? "On" : "Off"}
+          </span>
+          <Switch
+            checked={installed}
+            onCheckedChange={onToggle}
+            label={`${installed ? "Uninstall" : "Install"} ${skill.name}`}
+            className={disabled ? "pointer-events-none opacity-50" : undefined}
```
dam-code-guardian: 🟡 **Warning:** `pointer-events-none opacity-50` suppresses the mouse only. `Switch` (`components/ui/switch.tsx`) renders a plain `<button>` with neither `disabled` nor `aria-disabled`, so when there is no pod the control stays keyboard-focusable, Enter still fires `onToggle`, and assistive tech is told the control is available.

Combined with the header-slot focus finding on `modal.tsx`, the unusable control is also the one that receives focus when the dialog opens.

**Fix:** Express "not available" on the control itself with the native `disabled` attribute (or `aria-disabled` plus a handler guard), never with `pointer-events-none`, which leaves both the keyboard and the accessibility tree claiming the control is live.

(reported by react-ui-engineering)

packages/ui/src/modules/sandboxes/hooks/use-section-summaries.ts
```diff
@@ -90,14 +114,22 @@ export function useSectionSummaries(agent: AgentView | null): SectionSummaries {
...
   const skills = useMemo(() => {
-    // Standalone (authored-in-sandbox) skills are on disk and active just like
-    // installed ones, so the summary lists both — standalone first to mirror
-    // the surface's "Created in this sandbox" group ordering.
-    const standalone = (skillsState.data?.standalone ?? []).map((s) => s.name);
-    const installed = (skillsState.data?.installed ?? []).map((s) => s.name);
-    return formatNameList([...standalone, ...installed]);
-  }, [skillsState.data]);
+    if (!hasRun) return "Not known yet";
+    const on = skillsState.data?.installed.length ?? 0;
```
dam-code-guardian: 🟡 **Warning:** `?? 0` turns "not loaded yet" into a confident count: the nav renders "0 on across 3 sources" while `skills.state` is still pending, and permanently after it fails, since `useSkillsState` sets `retry: false`.

On `main` this memo returned `formatNameList([])`, i.e. `undefined`, and the nav showed its `—` placeholder — so this is a regression in what the line claims. With `sourceCount === 0` the same path asserts "None yet" about a sandbox whose skills were never read.

**Fix:** Never coalesce an unresolved or failed query to a numeric zero that is then rendered as fact — gate the summary on `skillsState.data !== undefined` and return `undefined` otherwise, so the neutral placeholder stands in for "unknown".

(reported by typescript-engineering)

packages/ui/src/modules/sandboxes/hooks/use-section-summaries.ts
```diff
@@ -90,14 +114,22 @@ export function useSectionSummaries(agent: AgentView | null): SectionSummaries {
...
+    const capturedAt = skillsState.data?.standaloneSnapshot?.capturedAt;
+    if (capturedAt) return `${on} on · as of ${timeAgo(capturedAt)}`;
```
dam-code-guardian: 🟡 **Warning:** This dates a live number with a snapshot's timestamp. `on` is `installed.length`, read from Postgres and current even while the pod is down, but the line renders it as `${on} on · as of ${timeAgo(capturedAt)}` — and `capturedAt` only dates the standalone recording. The count is correct and the label says it is history.

**Fix:** A captured-at timestamp may only qualify the values captured with it — keep live and recorded figures in separate clauses, or omit the timestamp from a clause built out of always-current state.

(reported by typescript-engineering)

packages/ui/src/modules/sandboxes/components/sandbox-section-nav.tsx
```diff
@@ -63,7 +75,17 @@ function SectionNavItem({
         active ? "bg-muted" : "hover:bg-muted/60",
       )}
     >
-      <span className="text-sm font-medium text-foreground">{title}</span>
+      <span className="flex items-center gap-1.5 text-sm font-medium text-foreground">
```
dam-code-guardian, line 74: 🟡 **Warning:** This `role="img"` span carries an `aria-label` inside the nav button, which already names itself from its title and summary spans — so the whole warning sentence is concatenated into that button's accessible name, and the item announces as "Setup Saved model not offered by the current provider …". The marker's text is duplicated into the name rather than exposed as a status.

**Fix:** A visual status marker inside an interactive element must not carry an `aria-label` that merges into that element's accessible name — render it `aria-hidden` and convey the condition once, in text the button already announces or in a sibling status region.

(reported by react-ui-engineering)

packages/ui/src/modules/sandboxes/components/skills/standalone-skill-row.tsx
```diff
@@ -21,26 +22,34 @@ import { formatDateTime } from "@/lib/format-time";
...
  */
 const PR_STATE_PILL: Record<
   NonNullable<SkillPublishRecord["prState"]> | "unknown",
-  { label: string; variant: "outline" | "info" | "success" | "muted" }
+  { label: string; variant: "info" | "success" | "muted" }
 > = {
-  draft: { label: "Draft", variant: "outline" },
-  open: { label: "In review", variant: "info" },
-  merged: { label: "Published", variant: "success" },
+  draft: { label: "Draft", variant: "muted" },
```
dam-code-guardian, line 28: 🟡 **Warning:** These pill labels reverse the plan's settled decision D1 (`Draft` / `In review` / `Published` (merged) / `Closed` / `Submitted`) back to the prototype's wording: `open` went `In review` → `Open`, `merged` went `Published` → `Merged`, and `draft`'s variant went `outline` → `muted`. Sub-issue 02's acceptance criterion asks for labels byte-identical to `PR_STATE_PILL` on `main`, and it is ticked.

The PR body's deviation row 1 declares only the `unknown → Submitted` case, so a reversal of a recorded decision reaches HEAD undeclared while the body certifies everything else as "a restyle or a rearrangement". The relabel itself is defensible — `Merged` matches GitHub — but it is a decision change, not a restyle.

**Fix:** A shipped change that contradicts a recorded decision must update that decision's register entry with its new reasoning — a deviation table that documents only part of a change is worse than none, because it certifies the rest as unchanged.

(reported by issue-fit)

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `130ffb5`

## PR #3267: feat(ui): rebuild the Skills surface on the new design
**Author:** PetrBulanek | **Branch:** feat/3208-skills-ui-rework → main | **Changes:** +1648 −458 (31 files)

### Summary
The Skills surface is rebuilt across four sandbox states, replacing the old dimmed read-only copy with dedicated stopped and never-run panels. The architecture, the `visibility` contract and the search-scoped bulk action are correct; the residual defects sit at the new state seams, where a loading value and an unrecorded snapshot both fall through to the wrong panel.

### Findings
- 🔴 **Critical:** The preview dialog's install/uninstall switch becomes the first focusable node, so opening a skill preview arms a mutation with the first keystroke (`packages/ui/src/components/modal.tsx:136`)
- 🟡 **Warning:** A stopped sandbox with no recorded snapshot falls through to the live branch, rendering enabled controls over guarded mutations (`packages/ui/src/modules/sandboxes/components/skills/skills-surface.tsx:198`)
- 🟡 **Warning:** `hasRun` conflates "still loading" with "never ran", so a stopped sandbox flashes the never-run panel (`packages/ui/src/modules/sandboxes/components/skills/skills-surface.tsx:182`)
- 🟡 **Warning:** The never-run and stopped panels present themselves as pod-free ("The source list is known before first boot"), but the surface's auto-scan effect fires for every source with no cached list and is guarded only on `!agentId`, never on `readOnly` (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:301`). A private source dispatches to the pod through `ensureAgentReachable`, which calls `ensureReady` — so merely opening the panel can start the sandbox the panel is telling the user to start, the very reason deviation row 4 gave for not shipping a picker. (reported by issue-fit)
  **Fix:** A surface whose copy claims it needs no pod must gate every upstream read on that same condition — state the precondition in code, not only in the prose beside it.
- 🟡 **Warning:** The `Switch` expresses "unavailable" with `pointer-events-none` only, so the keyboard still fires it (`packages/ui/src/modules/sandboxes/components/skills/skill-render-modal.tsx:90`)
- 🟡 **Warning:** The nav summary renders an unresolved or failed query as a confident `0` (`packages/ui/src/modules/sandboxes/hooks/use-section-summaries.ts:124`)
- 🟡 **Warning:** A snapshot timestamp dates a live Postgres count (`packages/ui/src/modules/sandboxes/hooks/use-section-summaries.ts:126`)
- 🟡 **Warning:** The warning marker's `aria-label` merges into the nav button's accessible name (`packages/ui/src/modules/sandboxes/components/sandbox-section-nav.tsx:78`)
- 🟡 **Warning:** Sub-issue 02 is ticked, but its acceptance criterion "Pill labels/variants byte-identical to `PR_STATE_PILL` on main" is unmet and the plan's settled decision D1 was reversed to the prototype's labels — `open` went `In review` → `Open`, `merged` went `Published` → `Merged`, `draft`'s variant `outline` → `muted` (`packages/ui/src/modules/sandboxes/components/skills/standalone-skill-row.tsx:38`). The body's deviation row 1 declares only the `unknown → Submitted` case, so a reversal of a recorded decision reaches HEAD undeclared while the body certifies everything else as "a restyle or a rearrangement". (reported by issue-fit)
  **Fix:** A shipped change that contradicts a recorded decision must update that decision's register entry with its new reasoning — a deviation table that documents only part of a change is worse than none, because it certifies the rest as unchanged.
- 🟡 **Warning:** Sub-issue 09 is ticked, but its criterion "every follow-up draft is either filed as an issue or preserved as a comment on #3208 **before** the commit that deletes the plan folder" is unmet for two of five drafts: A/B/C became #3283/#3284/#3285, while draft D (per-skill multi-select and search-scoped bulk) and draft E (two #3243 engineering findings) were neither filed nor posted — #3208 carries two comments and `docs/plan/3208-skills-ui-rework/` is deleted at HEAD. Draft D matters because #3023 closed on the promise it would be filed once #3208 landed. (reported by issue-fit)
  **Fix:** Deleting the only copy of a record is the irreversible step — verify each item reached its durable home and cite it before the commit that removes the source, not as a sibling of it.
- 🟡 **Warning:** `docs/architecture/skills.md:259` drops the invariant that a scan failure's `message` stays a readable sentence for consumers that only have it, notably the CLI — which still depends on it, printing the relayed server message verbatim (`packages/cli/src/modules/skill/services/skills-service.ts:193`). No other architecture page carries that contract, so the only record that the api-server owes the CLI readable prose is gone. The deletion was justified as paying for new text under the character cap, but the page has 252 characters of headroom, not the 35 the commit claims: `scripts/doc-size.mjs` measures characters (39,748) while `wc -c` reports bytes (39,965). Restoring the sentence costs 90. (reported by doc-drift)
  **Fix:** Restore the dropped sentence; when a page must shed content to fit its cap, measure with the repo's own gate (characters, `content.length`) rather than a byte count, and cut volatile detail rather than a cross-context contract the code still honours.
- 🟢 **Suggestion:** `snapshotOnCount` is `installed.length` while the running surface's `on` counts only installed skills still present in each source's post-suppression scan, so the interface comment's promise that the number "doesn't jump when a sandbox stops" fails whenever a skill is installed from a disconnected source or suppressed as a published duplicate (`packages/ui/src/modules/sandboxes/hooks/use-skills-derivations.ts:322`, reported by issue-fit). Deriving both counts from one helper would make the promise hold.
- 🟢 **Suggestion:** `useStaleModel` is a second implementation of the staleness rule that `unavailableModel` plus `modelsPaired` already express and that `model-settings-panel.tsx:198` still uses, so the new hook's docstring claim to be the single place is untrue and the two copies can drift on a user-facing factual warning (`packages/ui/src/modules/agents/api/harness-config.ts:148`, reported by typescript-engineering).
- ✅ **Looks good:** The `visibility` inference is sound where it would be easy to get wrong: the anonymous archive 404s and the *authenticated* pod scan then succeeds, which rules out a nonexistent or renamed repo, and a non-GitHub host stays `undefined` rather than being inferred from `viaPod`. The badge renders only `Private`, never a reassuring `Public` from unproven data.

The dropped CLI-contract sentence is merged into `### Findings` above (`skills.md:259`). `Last verified` is correctly bumped, and the new `visibility` contract and the badge relabel are both documented accurately.

### Issue Fit
Four findings are merged into `### Findings` above (the `hasRun` conflation, the read-only fallthrough, the pod-waking auto-scan, the reversed D1 decision, the unfiled drafts, and the `snapshotOnCount` mismatch). The 11-row deviation table otherwise checks out against HEAD, and #3283/#3284/#3285 exist and cover what the body says they cover.

Two findings are merged into `### Findings` above (the nav summary's coalesced `0`, the snapshot timestamp dating a live count) plus one suggestion. The `visibility` inference chain was checked and holds.

Four findings are merged into `### Findings` above (the header-slot focus trap, the `pointer-events-none` switch, the `hasRun` flash, the nav marker's accessible name). One further 🔴 — that the bulk-disable confirm counts before the `await` while the executed scope is re-derived after it — is **dropped after verification**: `toggleAllWithConfirm` and the memoized `toggleSource` both close over the same render's `installedRef`, so the quoted count and the executed scope read one snapshot and cannot diverge. Modelled in `node` to confirm.

### Verdict
REQUEST_CHANGES — the rebuild is sound and the contract change is honest, but the preview dialog's header slot puts a mutating switch first in the focus order, so the opening keystroke can uninstall a skill.

---

packages/ui/src/modules/sandboxes/components/skills/skills-modals.tsx
```diff
@@ -90,26 +117,83 @@ export function SkillsModals({
...
         <SkillRenderModal
           source={open.source}
           skill={open.skill}
           agentId={agentId}
+          visibility={surface.visibilityBySource[open.source.id]}
+          installed={ref !== undefined}
+          // The row's own predicate, not a second one: the drawer's Update
+          // button and the row's Update link must never disagree.
+          hasDrift={isDrifted(ref, open.skill)}
+          disabled={!agentId}
```
dam-code-guardian: 🟡 **Warning:** `disabled={!agentId}` drops `readOnly` and `isError`, so a stopped or errored sandbox's preview renders a live Switch and Update button over mutations that silently return. `skill-sources-section.tsx:111` already passes all three terms for the row.

**Fix:** Derive a control's `disabled` from the exact predicate its handler early-returns on, and expose that predicate from the owning hook so every consumer computes it identically.

```suggestion
          disabled={!agentId || surface.isError || surface.readOnly}
```

packages/ui/src/components/ui/button.tsx
```diff
@@ -25,7 +25,7 @@ const buttonVariants = cva(
       // instead of imposing its own.
       size: {
         default: "h-10 px-4 py-2 text-sm font-medium",
-        sm: "h-9 rounded-md px-3 text-sm font-medium",
+        sm: "h-8 rounded-md px-3 text-sm font-normal",
```
dam-code-guardian, line 26: 🟡 **Warning:** Retuning the shared `sm` token restyles 60 untouched `size="sm"` call sites across Sessions, Artifacts, Approvals, Settings and Files. The PR body declares only the Skills surface and the six replaced overrides, and the `h-9` placeholder in `sandbox-wake-to-edit.tsx:46` — a file this PR does not touch — now mismatches by 4px.

**Fix:** Keep a design-system primitive change out of a surface-scoped PR, or declare it as an explicit extra and name every surface it restyles, updating each element that hard-codes the old value.

packages/ui/src/modules/agents/api/skills.ts
```diff
@@ -20,3 +20,22 @@ export function useSkillsState(agentId: string | null) {
...
+ * is never worth that.
+ */
+export function useSkillSourceCount(agentId: string | null): number | null {
+  const { data } = useQuery({
+    ...trpc.skills.sources.list.queryOptions(agentId ? { agentId } : skipToken),
+    retry: false,
+    staleTime: 30_000,
+    refetchOnWindowFocus: false,
+  });
+  return data?.length ?? null;
```
dam-code-guardian, line 21: 🟡 **Warning:** `null` means both "still loading" and "load failed", and `retry: false` makes the failure permanent. When the source list fails while `on` and `created` are both 0, the `sourceCount === 0` guard in `use-section-summaries.ts:137` cannot fire, so the line prints `0 on` as fact instead of withholding.

**Fix:** A query hook must not encode "unknown" as a value that is also a legal answer — return the tri-state (`number | undefined` plus an error flag) and withhold the whole line while the count is unresolved.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `6b360da`

## PR #3267: feat(ui): rebuild the Skills surface on the new design
**Author:** PetrBulanek | **Branch:** feat/3208-skills-ui-rework → main | **Changes:** +1694 −480 (39 files)

### Summary
Every blocking finding of the last round is fixed at its cause, and the focus trap now excludes the whole header rather than one attribute. The residual defects are the seams each fix opened: `Switch` gained a native `disabled` that two call sites still fake, and the shared `sm` button token was retuned for the whole product inside a surface-scoped PR.

### Changes since last review
Previous HEAD: 130ffb57 (2026-08-12T14:04:26Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** The preview dialog's mutating header control no longer takes first focus — `DialogHeader` marks itself `data-dialog-noautofocus` and the trap skips the whole subtree (`packages/ui/src/components/modal.tsx:298`)
- ✅ **Fixed:** A stopped sandbox with no snapshot no longer renders live controls — `disabled` now includes `readOnly` (`packages/ui/src/modules/sandboxes/components/skills/skill-sources-section.tsx:111`)
- ✅ **Fixed:** `hasRun` no longer flashes the never-run panel — `pending` is exposed and gates the claim (`packages/ui/src/modules/sandboxes/components/skills/skills-surface.tsx:183`)
- ✅ **Fixed:** The preview modal's `Switch` takes a native `disabled` (`packages/ui/src/components/ui/switch.tsx:29`)
- ✅ **Fixed:** The nav summary withholds the count until the query resolves, instead of rendering `0` as fact (`packages/ui/src/modules/sandboxes/hooks/use-section-summaries.ts:127`)
- ✅ **Fixed:** The nav marker no longer merges into the button's accessible name (`packages/ui/src/modules/sandboxes/components/sandbox-section-nav.tsx`)
- ✅ **Fixed:** The CLI readable-`message` contract is restored to the architecture page, and the badge labels there match what ships (`docs/architecture/skills.md:258`)
- ✅ **Fixed:** `useStaleModel` now calls the same `unavailableModel` helper the panel uses, so one rule has one implementation (`packages/ui/src/modules/agents/api/harness-config.ts:167`)
- 🔁 **Still present:** Follow-up drafts D and E were never filed or posted before the plan folder was deleted; #3208 still holds two comments (reported by issue-fit)
- 🔁 **Still present:** The auto-scan effect is gated only on `!agentId`, never on `readOnly`, so opening a pod-free panel can wake the sandbox (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:301`)
- 🔁 **Still present:** Sub-issue 02's byte-identical pill criterion is unmet and the reversal stays undeclared (`packages/ui/src/modules/sandboxes/components/skills/standalone-skill-row.tsx:38`)
- 🔁 **Still present:** `snapshotOnCount` is `installed.length` while the surface's `on` counts scanned refs, so the number the comment promises cannot move still moves (`packages/ui/src/modules/sandboxes/hooks/use-skills-derivations.ts:322`)

### Findings
- 🟡 **Warning:** 🆕 `SkillRow` still suppresses its toggle with `pointer-events-none opacity-50`, so a keyboard user fires `onToggle` on a stopped or errored sandbox — the same defect this commit fixed in the preview modal (`packages/ui/src/modules/sandboxes/components/skills/skill-row.tsx:104`, reported by react-ui-engineering)
  **Fix:** Gate an interactive control with the native `disabled` attribute, never a pointer-events class. When a shared primitive gains `disabled`, convert every call site that faked it, not only the one that reported the bug.
- 🟡 **Warning:** 🆕 The preview modal gets `disabled={!agentId}`, which drops `readOnly` and `isError`, so a stopped sandbox's preview renders a live Switch and Update button over mutations that silently return (`packages/ui/src/modules/sandboxes/components/skills/skills-modals.tsx:132`, reported by react-ui-engineering)
  **Fix:** Derive a control's `disabled` from the exact predicate its handler early-returns on, and expose that predicate from the owning hook so every consumer computes it identically — `skill-sources-section.tsx:111` already passes all three terms.
- 🟡 **Warning:** 🆕 The shared `Button` `sm` token was retuned globally (`h-9`/`font-medium` → `h-8`/`font-normal`), which restyles 60 untouched `size="sm"` call sites across Sessions, Artifacts, Approvals, Settings and Files; the body declares only the Skills surface and six replaced overrides (`packages/ui/src/components/ui/button.tsx:28`, reported by issue-fit)
  **Fix:** Keep a design-system primitive change out of a surface-scoped PR, or declare it as an explicit extra and name every surface it restyles.
- 🟡 **Warning:** 🆕 The same token change leaves the "Agent is starting…" placeholder hard-coding `h-9` to match the `size="sm"` button it replaces, so the header now shifts 4px when the agent starts — in a file this PR does not touch (`packages/ui/src/modules/sandboxes/components/sandbox-wake-to-edit.tsx:46`, reported by react-ui-engineering)
  **Fix:** When a shared size token changes, update every element that hard-codes the old value; a placeholder sized to a button must read its height from the same token rather than repeating the number.
- 🟡 **Warning:** 🆕 `hasRun` collapses the unknown arm: `recorded?.hasRun ?? false` maps both "not loaded" and "load failed" to `false`, and `retry: false` makes a failed read permanent while `pending` goes false on error — so the never-ran claim reaches the screen for a sandbox that has run (`packages/ui/src/modules/agents/api/harness-config.ts:102`, reported by typescript-engineering). The added comment names the `pending` hazard, not the error arm.
  **Fix:** Never coalesce an unresolved or failed query into a definite domain value. Carry the third arm explicitly — `boolean | undefined`, or an `isError` flag on `ResolvedHarnessConfig` — and make each consumer branch on it, so only a settled `false` earns the claim.
- 🟡 **Warning:** 🆕 `useSkillSourceCount` returns `null` for both "loading" and "failed", and with `retry: false` the failure is permanent; when the source list fails while `on` and `created` are both 0, the `sourceCount === 0` guard cannot fire and the line prints `0 on` instead of the placeholder (`packages/ui/src/modules/agents/api/skills.ts:40`, consumed at `use-section-summaries.ts:137`, reported by typescript-engineering)
  **Fix:** Same rule — a query hook must not encode "unknown" as a value that is also a legal answer. Return the tri-state and withhold the whole line while the count is unresolved.
- 🟡 **Warning:** 🆕 Sub-issue 09 required the architecture page to document the stopped snapshot panel and the never-run scope, and the box is ticked, but only the `visibility` field and the badge labels landed; the page now sits at 39,838 of its 40,000-character gate, so the delta was trimmed to fit rather than taken through the page split slice 09 named (`docs/architecture/skills.md:3`, reported by issue-fit)
  **Fix:** When a documentation delta cannot fit a size gate, split the page as the plan directs; never reduce documented scope to fit, and never tick a criterion the diff only partly meets.
- 🟡 **Warning:** 🆕 `Last verified:` reads 2026-08-12, but a later commit in this PR edited the page on 2026-08-13 (`docs/architecture/skills.md:3`, reported by doc-drift)
  **Fix:** Bump `Last verified:` to the date of the final commit that edits the page, and re-check the header after every later commit that touches an architecture page.
- 🟢 **Suggestion:** `harness-config.ts` (module `agents`, api layer) now imports the pure `unavailableModel` from `sessions/components/model-settings-snapshot.js`, which also exports two components — and `sessions` already imports this file, so the two modules point at each other. Moving the predicate to `lib/` would single-source the rule without the cycle (`packages/ui/src/modules/agents/api/harness-config.ts:6`, reported by typescript-engineering).

The `Last verified` staleness is merged into `### Findings` above. The dropped CLI readable-`message` contract is restored, and the `visibility` contract and badge relabel are both documented accurately at this HEAD.

### Issue Fit
Six findings are merged into `### Findings` above and into the delta block — the unfiled drafts D and E (verified: #3208 still carries two comments, #3283/#3284/#3285 cover A/B/C only), the undeclared pill reversal, the ungated auto-scan, the `snapshotOnCount` mismatch, the global button token, and sub-issue 09's partly-met doc criterion. The drafts finding was reported 🔴 by the skill; it is carried here at 🟡, its prior severity, because the deletion is already done and the remedy is filing, not a code fix.

Five findings are merged into `### Findings` above (the `hasRun` error arm, the source-count tri-state, the snapshot timestamp dating a live count, plus two suggestions). The `useStaleModel` duplication is resolved: it now delegates to `unavailableModel`.

Three findings are merged into `### Findings` above (the row's `pointer-events-none` toggle, the preview modal's incomplete `disabled`, the `h-9` placeholder). The prior round's 🔴 focus trap was verified fixed at the cause — `data-dialog-noautofocus` on the header subtree, checked with `closest()`, so any control added to either header slot is covered rather than one attribute at a time.

### Verdict
REQUEST_CHANGES — the round's blocking findings are all genuinely fixed, but `Switch` gained a native `disabled` that the row and the preview modal still fake, so a keyboard user can toggle a skill on a sandbox that will refuse the write.

---

dam-code-guardian (dismissed):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `6b360da1`

## PR #3267: feat(ui): rebuild the Skills surface on the new design
**Author:** PetrBulanek | **Branch:** feat/3208-skills-ui-rework → main | **Changes:** +1694 −480 (39 files)

### Summary
The Skills surface is rebuilt across four sandbox states. This round fixes the previous 🔴 and five of the 🟡 at their cause; the residual defects are the seams those fixes created — a `disabled` prop converted at one call site but not its siblings, and a size token retuned globally from inside a surface-scoped PR.

### Changes since last review
Previous HEAD: `130ffb57` (2026-08-12T14:04:26Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** The preview dialog's header no longer takes initial focus — `DialogHeader` carries `data-dialog-noautofocus` and the trap skips anything inside it (`packages/ui/src/components/modal.tsx:108`, `:304`)
- ✅ **Fixed:** `Switch` gained a native `disabled` prop; the preview modal passes it (`packages/ui/src/components/ui/switch.tsx:29`, `skill-render-modal.tsx:89`)
- ✅ **Fixed:** A stopped sandbox with no snapshot no longer renders live source controls — `disabled` now includes `readOnly` (`skill-sources-section.tsx:111`)
- ✅ **Fixed:** The never-run panel no longer flashes while the snapshot loads (`skills-surface.tsx:183`)
- ✅ **Fixed:** The nav summary no longer coalesces an unresolved `skills.state` to `0` (`use-section-summaries.ts:131`)
- ✅ **Fixed:** The nav warning marker is `aria-hidden`, so it no longer merges into the button's accessible name (`sandbox-section-nav.tsx:85`)
- ✅ **Fixed:** The dropped CLI-contract sentence is restored; measured at 39,838 of 40,000 characters (`docs/architecture/skills.md:259`)
- ✅ **Fixed:** `useStaleModel` now compares through `unavailableModel` — verified behaviour-identical across all nine tri-state cells (`harness-config.ts:167`)
- 🔁 **Still present:** `hasRun` still collapses the failed arm — `retry: false` makes it permanent (`harness-config.ts:102`)
- 🔁 **Still present:** A snapshot timestamp still dates a live Postgres count (`use-section-summaries.ts:135`)
- 🔁 **Still present:** The pod-free panels' auto-scan still wakes the pod, gated only on `!agentId` (`use-skills-surface.ts:301`)
- 🔁 **Still present:** `snapshotOnCount` still diverges from the running `on` (`use-skills-derivations.ts:322`)
- 🔁 **Still present:** Drafts D and E were never filed; the plan folder holding them is deleted (sub-issue 09)
- 🔁 **Still present:** Sub-issue 02's pill labels still differ from `main`, undeclared (`standalone-skill-row.tsx:38`)
- 🆕 **New:** Four findings below.

### Findings
- 🟡 **Warning:** `SkillRow` suppresses its install/uninstall `Switch` with `pointer-events-none opacity-50` instead of the native `disabled` prop the same commit added to `Switch`, so the keyboard still fires `onToggle` on a stopped or errored sandbox, and the accessibility tree is told the control is live. This is the defect `c69e4a32` fixed in the preview modal (`skill-render-modal.tsx:89`), while the `Update` button two lines above (`:76`) already uses the native attribute (`packages/ui/src/modules/sandboxes/components/skills/skill-row.tsx:104`, reported by react-ui-engineering).
  **Fix:** Gate an interactive control with the native `disabled` attribute, never with a pointer-events class. When a shared primitive gains a `disabled` prop, convert every call site that was faking it, not only the one that reported the bug.
- 🟡 **Warning:** The preview modal's `disabled` omits `readOnly`/`isError` (`packages/ui/src/modules/sandboxes/components/skills/skills-modals.tsx:132`)
- 🟡 **Warning:** `useSkillSourceCount` conflates loading and permanently-failed into `null`, so the summary renders `0 on` (`packages/ui/src/modules/agents/api/skills.ts:40`)
- 🟡 **Warning:** `size="sm"` is retuned globally, restyling 33 files outside this diff (`packages/ui/src/components/ui/button.tsx:28`)
- 🟢 **Suggestion:** `harness-config.ts` imports `unavailableModel` from `sessions/components/`, the only api→components edge in the package, while `sessions` imports back from this same file — the two modules now point at each other. A pure `lib/` helper (beside `format-time.ts`) would single-source the rule without the cycle (`packages/ui/src/modules/agents/api/harness-config.ts:6`, reported by typescript-engineering).

- 🟡 **Warning:** The `Last verified:` header reads 2026-08-12, but commit `c69e4a32` edited the page on 2026-08-13, so the page under-reports its last edit (`docs/architecture/skills.md:3`).
  **Fix:** Bump `Last verified:` to the date of the final commit that edits the page — the documentation guidelines require it "whenever you edit the page". Re-check the header after every later commit that touches an architecture page.

### Issue Fit
Five findings merge into `### Findings` and `### Changes since last review` above (the unfiled drafts D/E, the reversed pill labels, the pod-waking auto-scan, the `snapshotOnCount` mismatch, and the `sm` blast radius). One further claim — that sub-issue 09's architecture-page criterion is unmet — is **dropped after verification**: `main` already documents the snapshot and never-ran distinction at architecture level (`skills.md:273`), so the page was not reduced in scope to fit the size gate. Drafts A/B/C reached #3283/#3284/#3285 and the deviation table did reach #3208 as a comment, so those parts of criterion 09 hold.

Three findings merge above plus two suggestions; the `?? 0` regression at `use-section-summaries.ts` is fixed. The `useStaleModel` refactor was modelled against the old predicate across all nine cells of the tri-state space and is behaviour-identical, including the `availableModels: null` and unpaired arms.

Three findings merge above. The prior 🔴 focus trap and the `pointer-events-none` switch are both fixed at the cause. The previously dropped bulk-confirm 🔴 candidate stays dropped — `list` is captured before the `await`, so the quoted count and the executed scope cannot diverge.

### Verdict
REQUEST_CHANGES — the round fixes eight findings at the cause, but the `Switch` conversion stopped one call site short and a shared button token was retuned across 33 unaudited files from inside a surface PR.

---

dam-code-guardian:
🛡️ **Code Guardian 🛡️** — correction to the review above

**Withdrawing the pill-label finding** (`standalone-skill-row.tsx:38`), listed there under `### Changes since last review` as still present and undeclared. That is wrong, and the previous round asked for exactly what shipped.

Commit `4d3aba8` amended decision D1 in the same commit that changed the labels:

- before: `Badges Open, Merged, Published (state unknown)` → `In review`, `Published` (merged), `Submitted` (unknown)
- after: the unreadable case badged `Published` → `Submitted`, with the reasoning that `Draft`/`Open`/`Merged`/`Closed` follow the prototype and GitHub exactly

That is the register update the last round's **Fix:** line required ("a shipped change that contradicts a recorded decision must update that decision's register entry with its new reasoning"). Re-reporting it as an open finding re-raised a requested fix as a defect — my error, not a regression on your side.

What genuinely remains is a nit, not a blocker: sub-issue 02's ticked criterion still reads "byte-identical to `PR_STATE_PILL` on main", which the amended D1 supersedes — the criterion outlived the decision it was written against, in a plan folder this PR deletes. Nothing to fix in the code.

**The verdict does not change.** It rests on the two `Switch` call sites that still fake `disabled` with `pointer-events-none` (`skill-row.tsx:104`, `skills-modals.tsx:132`), which stand as posted. The other findings are unaffected.

PetrBulanek:
### Heads-up for re-review: the comment deletions are policy, not scope creep

This branch was conflicting after #3308 landed on `main`. That PR stripped code comments repo-wide (1,156 files, −21,102 lines) and introduced a new rule in `CLAUDE.md`:

> No code comments are allowed. Exception: typed comments (`TEST_OVERVIEW:`, `TEST_SCENARIO:`), enforced by `mise run common:check:comment-types`.

Every conflict was a comment removal on `main`'s side against a file this branch had rewritten — **zero logic changes**. So the branch keeps its own content, and the policy is then applied to it in a separate commit.

Two commits, deliberately kept apart so the noise is skippable:

| Commit | What it is |
|---|---|
| `90f681a4` | Merge `main`. Conflicts resolved in favour of the branch side, which is safe precisely because `main`'s side was comment-only. |
| `cd796b62` | Apply the comment policy to this branch's code. **30 files, −409 lines. No behaviour change.** Produced by running the repo's own `scripts/strip-comments.mjs --write` — same tool, same treatment `main` got. |

**To review without wading through it:** read everything except `cd796b62`, then skim that one on its own. It is pure deletion plus prettier reflow.

Verified locally before pushing: `mise run check` passes (including the new comment-types gate), `ui:test` 277/277, `api-server:test` 920/920, and the Skills surface renders unchanged.

One known consequence worth naming: the rationale comments this PR's description leans on — why the never-run panel doesn't fake skill selection, why the bulk control is scoped to visible matches — no longer live at the call site. They survive in this PR's description and the sub-issues. If any of it should be permanent, the architecture docs are now the place for it.

Unrelated to this branch, but noticed while merging: the stripper leaves an empty `{}` where a JSX comment stood alone in an expression container — 57 of them already on `main`. This branch adds none (checked); worth a separate issue against the script.

packages/ui/src/modules/sandboxes/components/skills/skills-surface.tsx
```diff
@@ -77,20 +84,51 @@ export function SkillsSurface({
...
       onClick={() =>
         setOpenModal({ kind: "add-source", tab: "github", files: [] })
       }
     >
-      <Add size={14} /> Add source
+      <Add size={16} /> Add source
     </Button>
   );
 
+  const stoppedPanel = readOnly && standaloneSnapshot !== undefined && (
```
kapetr: **🟡 The code selects the stopped panel from the data, not from the state.**

The condition is `readOnly && standaloneSnapshot !== undefined`. A stopped sandbox with no snapshot gets the normal page.

The server writes the snapshot only when it reads the state of a *running* sandbox (`skills-service.ts:966`). Thus a sandbox that ran, but never showed its Skills page while it ran, has no snapshot. Every sandbox that existed before this column also has none.

In this condition the user sees no read-only message and no **Start sandbox** button. The old code showed both in `sandbox-skills-section.tsx`. This change deleted them.

packages/ui/src/modules/sandboxes/components/skills/skills-stopped-panel.tsx
```diff
@@ -0,0 +1,155 @@
...
+    <div className="flex flex-col gap-8">
+      {callout && <div className="-mb-4">{callout}</div>}
+      <div className="flex items-start gap-2.5 rounded-lg border border-border bg-muted px-4 py-3 text-sm">
+        <Information
+          size={16}
+          className="mt-px shrink-0 text-muted-foreground"
+        />
+        <p className="min-w-0 flex-1">
+          <span className="font-semibold" title={formatTimestamp(capturedAt)}>
+            Last known configuration, captured {timeAgo(capturedAt)}
```
kapetr: **🟡 The date is not the date of the last run.**

The panel says "captured *X* ago" here, and "as of *X* ago" at line 112. Both read `capturedAt`.

But `recordStandaloneSnapshot` stops early when the list of skills is the same (`agent-skills-repository.ts:249`). Therefore `capturedAt` is the time of the last *change* to that list.

A sandbox with no created-here skills keeps its first timestamp for ever. So a sandbox that ran yesterday can say "captured 6 months ago". The same date goes to the sidebar (`use-section-summaries.ts:110`).

packages/ui/src/modules/sandboxes/components/skills/skills-source-list.tsx
```diff
@@ -0,0 +1,138 @@
...
+  onRescan,
+  onRemove,
+}: {
+  sources: SkillSource[];
+  visibilityBySource: Record<string, "public" | "private">;
+  scannedAtBySource: Record<string, string>;
+  onRescan: (source: SkillSource) => void;
+  onRemove: (source: SkillSource) => void;
+}) {
+  if (sources.length === 0) {
```
kapetr, line 166: **🟡 The list has no load state and no error state.**

The component receives `sources` only. It shows "No skill sources connected." when the array is empty. The array is empty while the query runs, so the user reads a false statement first.

The component also ignores `errorBySource`, so a scan failure is invisible. The server writes correct text for this case ("Start the sandbox, then re-scan to list this source's skills", `skills-service.ts:252`). Both new panels discard it. The footnote below the list says the scan times "stay accurate".

packages/ui/src/modules/sandboxes/components/skills/skills-modals.tsx
```diff
@@ -84,25 +108,76 @@ export function SkillsModals({
...
+      const ref = surface.installedRef(open.skill.source, open.skill.name);
       return (
         <SkillRenderModal
           source={open.source}
           skill={open.skill}
           agentId={agentId}
+          visibility={surface.visibilityBySource[open.source.id]}
+          installed={ref !== undefined}
+          hasDrift={isDrifted(ref, open.skill)}
+          disabled={!agentId}
```
kapetr: **🔵** The switch and **Update to latest** test `!agentId` only, not `readOnly`. Thus they look live and do nothing.

kapetr (changes requested):
Three findings that block the merge, and one nit. All are verified against the code.

Findings 1 and 3 have one cause. The new panels branch on whether the data arrived. They do not branch on what the sandbox does.

kapetr (approved):

packages/ui/src/modules/sandboxes/components/skills/skills-surface.tsx
PetrBulanek: Fixed in f4245f27. The gate now reads the state, not the data: readOnly branches on configPending / hasRun / stopped, and the stopped panel renders for every stopped sandbox. Verified against the dev cluster, where agent-117dfeb2c6be07ae was in exactly this state — has run (runtime_capabilities set), skills_snapshot NULL. When no snapshot exists the panel simply carries no created-here row; the banner and Start sandbox button are always there.

packages/ui/src/modules/sandboxes/components/skills/skills-stopped-panel.tsx
PetrBulanek: Fixed in f4245f27. You are right that capturedAt dates the last change to the created-here list, so it cannot date anything else. Two of the three uses were also dating a live number — onCount is installed.length, read from Postgres and current while the pod is down. So the date now sits on the created-here row alone (recorded X ago), the section header lost its as of X, and the sidebar line lost it too. The header reads Skills at next start, which is what installed actually is while stopped. A true last-run stamp needs runtime_last_hello_at, which is not in the contract — worth a follow-up if the design wants a dated panel.

packages/ui/src/modules/sandboxes/components/skills/skills-source-list.tsx
PetrBulanek: Fixed in f4245f27. SkillSourceList now takes loaded, loadingBySource and errorBySource, and renders a skeleton, a per-row scanning state, and the server's ScanFailure via the same SourceError block the running card uses (extracted to skill-source-error.tsx). Manage connections is threaded through for connection failures. The footnote was wrong as well, and I rewrote it: the scan cache is a 5-minute in-memory map, and reading a private source needs the pod, so the old claim that scan times stay accurate could not hold.

packages/ui/src/modules/sandboxes/components/skills/skills-modals.tsx
PetrBulanek: Fixed in f4245f27. Rather than repeat the predicate, the surface hook now exposes mutationsDisabled (!agentId || isError || readOnly) and both the source card and the preview dialog read it, so they cannot drift again.

PR #3273, opened 2026-08-11
feat(sessions): time scheduled fires and report them on the session list's clock

xjacka:
## What

A scheduled session's row in the session list reports how long its fires have run: `Ran 4m 12s`, or `Ran 32m 10s across 8 runs` for a schedule that reuses one session. A fire in flight reads `Running now`, or `Running now · 3 runs finished, 12m 4s` once earlier fires have landed. The reading sits in the row's meta line as text, so it is there on touch as well as hover.

## Why this needed run accounting

The first attempt derived the figure from the session's `updatedAt`. That stamp is `lastActivityAt` — every prompt and turn end bumps it, including a human replying in the same session days later, which turned `Ran 4m 12s` into `Ran 2d 0h`. No existing field held a run's end: the scheduler cannot supply one either, because `TriggerSessionDriver.start` sends `session/prompt` fire-and-forget and returns before the run finishes.

So agent-runtime now times the runs themselves. A turn is timed when both halves of the gate hold: it arrives on a machine-driven channel (`viewer: false` — never a person typing) **and** its session belongs to a schedule (`type === schedule_cron` or a `scheduleId`). Both are required, so the trigger driver's other machine turns — an experiment launch on the same channel — accumulate nothing. The run opens when the harness actually takes the prompt frame, and closes when the turn ends; the elapsed time folds into per-session totals in the session-metadata store.

## Notable decisions

- **Summed, not last.** A continuous schedule reuses one session across every fire, so a sum is the readable number there. Because the totals are per-fire, the reading means the same thing in both session modes.
- **Machine turns only.** `finishRun` is a no-op unless a run is open, so a human turn in a scheduled session cannot inflate the totals.
- **An open run is closed exactly once, by whatever ends it.** The turn edges live in `prompt-scheduler`, which owns turn state: its `onTurnStarted` / `onTurnEnded` callbacks fire on every path that starts or drops an active turn — a response, a session teardown, a harness exit that drops them all — so no fire is left believed to be running.
- **Timed from execution, not arrival.** The run opens where the frame reaches the harness, so a fire queued behind another turn is not charged its wait.
- **A run stamp found at boot outlived the process that set it**, so it is discarded rather than folded in — a crashed fire's length is unknowable.
- **Text, not a tooltip.** A Radix tooltip never opens on tap, and the sessions sidebar is a full-width primary surface on touch, so the value is rendered as text rather than hidden behind hover.

## Shared surfaces this touches

- `session-metadata-store.set()` now spreads the existing entry instead of rebuilding it from an allowlist, so a partial update (a resume carrying a new mode) cannot drop fields the entry gained elsewhere.
- `notifyingSessionMetadataStore` decorates the two new writes, so a fire starting or ending emits a session-watch notice like every other metadata write.
- `PromptSchedulerDeps` gains two optional turn-edge callbacks; both default to absent, so the scheduler's existing behaviour is unchanged.

## New persisted state

Three optional fields per session in the session-metadata state file on the PVC (`runStartedAt`, `runTotalMs`, `runCount`). Optional, so existing entries keep parsing and no migration runs; sessions that predate the accounting simply show no reading until their next fire. Documented in `docs/architecture/persistence.md`, with the surface described in `docs/architecture/platform-topology.md`.

## Verification

`agent-runtime` (270), `ui` (334) and `cli` (154) unit tests pass; tsc, lint and format clean across `agent-runtime`, `agent-runtime-api`, `ui`, `cli` and `api-server-api`. Not exercised end-to-end against a real scheduled fire.


dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `60a0e90`

### Summary
A hover on the session list's clock now reports how long a scheduled run took, plus the span it measured. The tooltip is correctly absent on `continuous` schedules, where one session spans every fire.

### Findings
- 🟡 **Warning:** `updatedAt` is not run-scoped, so the reported run time grows after the run ends. It is `lastActivityAt`, which `recordActivity` bumps on every prompt and turn end (`acp-runtime.ts:1169,1412`) — and a scheduled session is an ordinary resumable row (`session-category.ts:32`). A human who replies two days later turns `Ran 4m 12s` into `Ran 2d 0h`. The same line also fabricates `Ran 0s` when `updatedAt` is absent: `?? s.createdAt` makes the end equal the start, so `durationBetween` returns `"0s"` and the `?? "—"` guard never runs. (`packages/ui/src/modules/sessions/components/session-row.tsx:307`)
  **Fix:** Do not read a session's last-activity stamp as a run end. Stamp the scheduled turn's own end, or bound the reading so activity after that turn cannot extend it — and pass a nullable end through unchanged, so an absent stamp shows the span as unknown instead of `0s`.
- 🟡 **Warning:** The tooltip trigger is a real `<button>` (`tooltip.tsx` → `HintTooltip`) nested inside the row root, which is `role="button" tabIndex={0}` via `clickableProps`. It omits the `stopPropagation` that both other in-row controls carry (lines 182, 210), and `clickableProps` guards only `onKeyDown` by target, never `onClick` — so a click on the clock resumes the session. It also adds a second tab stop on every fresh-schedule row. (`packages/ui/src/modules/sessions/components/session-row.tsx:274`)
  **Fix:** Stop click propagation on every interactive descendant of a `clickableProps` row, or keep in-row auxiliary controls out of the tab order.
- 🟢 **Suggestion:** The in-flight reading subtracts a pod timestamp from the browser clock, so skew distorts a second-precision figure, and a lagging clock prints `—` on a run that is plainly active (`session-row.tsx:308`). `timeAgo` gets away with this only because minute granularity hides seconds.
- 🟢 **Suggestion:** `useScheduleSessionModes` shares the `schedules.list` query key with `useSchedules` but sets `retry: false` and no `meta`, which are query-scoped in React Query v5 (`queries.ts:41`). No live defect: `SessionsSidebar` and `SchedulesPanel` sit on mutually exclusive router branches (`app.tsx:90` early-returns chat before `sandbox-home` at line 114), so the two never observe the query together. Worth aligning the options anyway — the divergence becomes a silent, mount-order-dependent loss of the panel's retries and `errorToast` the moment one surface embeds the other.
- 🟢 **Suggestion:** `s.sessionMode ?? "fresh"` re-states a server-owned rule for the seventh time (`queries.ts:45`; also `scheduler-runner.ts:79`, `trigger-plugin.ts:27`, `get.ts:38`, `reset-session.ts:54`, `schedule-form-schema.ts:126`). All seven agree today; normalizing once in `toView` and narrowing the type would keep it that way.
- ✅ **Looks good:** The `continuous` exclusion is sound, and the reason is real rather than defensive — `trigger-plugin.ts:27` resumes the prior session on a continuous fire, so that session's span genuinely covers every run. `showRunTime` also correctly reads `undefined` (list unresolved) as "not fresh".

✅ No findings.

### Issue Fit

✅ No findings. The PR delivers every deliverable its description promises, with no undeclared scope.

Both findings merged into `### Findings` above (the shared query key at `queries.ts:41`, the repeated `?? "fresh"` default at `queries.ts:45`).

All five findings merged into `### Findings` above. Two were re-scoped by full-file verification: the shared-key finding's co-mount premise is refuted by the router's exclusive branches, and the clock-skew finding sits below the blocking bar given the existing `timeAgo` precedent.

### Verdict
COMMENT — the feature and its `continuous` exclusion are correct, but the run time it reports is derived from a stamp that outlives the run, and the new tooltip trigger swallows a row click.

---

packages/ui/src/modules/sessions/components/session-row.tsx
```diff
@@ -272,3 +293,26 @@ function SessionIndicators({
...
+ *  so far while a turn is still in flight. Rendered only for a fresh-session
+ *  schedule, where the session is exactly one run. */
+function ScheduledRunTime({
+  session: s,
+  running,
+}: {
+  session: SessionView;
+  running: boolean;
+}) {
+  const last = s.updatedAt ?? s.createdAt;
```
dam-code-guardian: 🟡 **Warning:** `updatedAt` is not run-scoped, so this run time keeps growing after the run ends.

It is `lastActivityAt`, which `recordActivity` bumps on every prompt and every turn end (`acp-runtime.ts:1169,1412`). A scheduled session is an ordinary resumable chat row (`session-category.ts:32`), so a human who opens it and replies two days later turns `Ran 4m 12s` into `Ran 2d 0h` — stated as fact, with no hint that the span is no longer one run.

The same line has a second failure in the other direction: when the harness never stamped activity, `?? s.createdAt` makes the end equal the start, so `durationBetween` returns `"0s"` rather than `null` and the `?? "—"` guard on line 311 never fires. The hover then reports `Ran 0s` for a run whose duration is simply unknown.

**Fix:** Do not read a session's last-activity stamp as a run end. Stamp the scheduled turn's own end, or bound the reading so activity after that turn cannot extend it. Pass a nullable end through unchanged, so an absent stamp renders as unknown instead of `0s`.

packages/ui/src/modules/sessions/components/session-row.tsx
```diff
@@ -256,9 +269,17 @@ function SessionIndicators({
             aria-label="Channel session"
           />
         ))}
-      {scheduled && (
-        <Time size={16} className="text-foreground" aria-label="Scheduled" />
-      )}
+      {scheduled &&
+        (showRunTime ? (
+          <HintTooltip
```
dam-code-guardian: 🟡 **Warning:** This tooltip trigger swallows a row click, and adds a second tab stop per row.

`HintTooltip` renders a real `<button type="button">`, and the row root is `role="button" tabIndex={0}` via `clickableProps`. Both other in-row controls carry `onClick={(e) => e.stopPropagation()}` (lines 182 and 210); this one does not, and `clickableProps` guards only `onKeyDown` by target — never `onClick`. So activating the clock reaches `handleClick` and resumes the session, which is not what a hint affordance should do. Nesting an interactive control inside a `role="button"` element is also invalid ARIA.

**Fix:** Stop click propagation on every interactive descendant of a `clickableProps` row, or keep in-row auxiliary controls out of the tab order.

packages/agent-runtime/src/modules/acp/services/acp-runtime.ts
```diff
@@ -1162,6 +1162,7 @@ export function createAcpRuntime(deps: AcpRuntimeDeps): AcpRuntime {
             active !== undefined && active.outboundId === outboundId;
           if (turnEnded) {
             activePromptBySession.delete(sid);
+            deps.sessionMetadata?.finishRun(sid);
```
dam-code-guardian: 🔴 **Critical:** This is the only path that closes an open run, so every abnormal end leaks `runStartedAt`.

The three teardown paths all drop the session's active-prompt slot without folding its run: `tearDownSession` at `:793`, `recycleAgentForEnv` at `:685`, and the agent-exit handler at `:619` — after which no prompt response can ever arrive. The stamp is persisted to the state file, so the fire's time never reaches `runTotalMs` and `runTimeLabel` reads `runStartedAt` as "still running": the row shows `Running now` permanently, and a pod restart does not clear it (nothing reconciles the store at boot).

**Fix:** Make "an open run is closed exactly once" an invariant of clearing a session's active-prompt slot rather than a side effect of the happy path — call `finishRun` from all four sites, or route every slot removal through one helper that does it. Add a boot-time reconciliation in the store that folds or drops a `runStartedAt` no live turn owns, because the flag outlives the process that wrote it.

packages/agent-runtime/src/modules/acp/services/acp-runtime.ts
```diff
@@ -1412,6 +1413,10 @@ export function createAcpRuntime(deps: AcpRuntimeDeps): AcpRuntime {
         deps.sessionMetadata?.recordActivity(promptSessionId);
         if (hasEngagedViewer(promptSessionId))
           deps.sessionMetadata?.recordSeen(promptSessionId);
+        // A prompt off a machine-driven channel is a scheduled fire, not a
+        // person typing — only those are timed as runs.
+        if (nonViewerChannels.has(channel))
+          deps.sessionMetadata?.startRun(promptSessionId);
```
dam-code-guardian: 🟡 **Warning:** This stamps the run before the queue decision at `:1428`, so a queued fire is timed from arrival rather than from execution.

Two effects. A fire that arrives while another turn holds the session waits in `promptQueueBySession` — possibly for the whole prior turn — and all of that wait is charged to its run. And because `startRun` overwrites unconditionally, the queued fire's stamp replaces the in-flight fire's: the next `finishRun` folds the wrong start, and the first fire's real duration is lost. Two schedules firing into one continuous session is enough to make `runTotalMs` exceed the real run time.

**Fix:** Time a run from the point the prompt reaches the agent — call `startRun` inside `forwardPromptToAgent`, which covers the direct send and the `advanceQueue` promotion alike. Make `startRun` a no-op when a run is already open, so no concurrent fire can discard another's start stamp.

packages/ui/src/modules/sessions/components/session-row.tsx
```diff
@@ -272,3 +286,15 @@ function SessionIndicators({
...
 }
+
+/** Time the schedule's fires have spent running, summed over every fire this
+ *  session served. Null until the first fire finishes — including on sessions
+ *  that predate the accounting. */
+function runTimeLabel(s: SessionView): string | null {
+  const runs = s.runCount ?? 0;
+  const total = formatDuration(s.runTotalMs ?? 0);
+  if (s.runStartedAt)
+    return runs ? `Running now · ${runs} runs, ${total} so far` : "Running now";
```
dam-code-guardian: 🟡 **Warning:** The in-flight branch does not make the count singular, so one finished fire plus one running renders `Running now · 1 runs, 42s so far`.

The finished branch below already special-cases `runs === 1`, so the two disagree.

**Fix:** Derive the noun once and use it in every branch that prints a count.

```suggestion
  if (s.runStartedAt)
    return runs
      ? `Running now · ${runs} ${runs === 1 ? "run" : "runs"}, ${total} so far`
      : "Running now";
```

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `98acbe2`

## PR #3273: feat(ui): report a scheduled run's duration on the session list's clock
**Author:** xjacka | **Branch:** feat/scheduled-run-time-tooltip → main | **Changes:** +156 −5 (10 files)

### Summary
The run time now comes from real run accounting in the agent-runtime, not from the session's last-activity stamp. That replaces the mechanism, and it resolves both warnings of the previous review.

### Changes since last review
Previous HEAD: `60a0e90` (2026-08-11T21:10:16Z) — verdict COMMENT

- ✅ **Fixed:** run time no longer reads `updatedAt`; `startRun`/`finishRun` time machine-driven turns only (`session-metadata-store.ts:132,144`)
- ✅ **Fixed:** the `Ran 0s` fabrication is gone — `runTimeLabel` returns `null` until a fire finishes (`session-row.tsx:298`)
- ✅ **Fixed:** the tooltip trigger now stops click propagation, so the clock no longer resumes the session (`session-row.tsx:269`)
- ✅ **Fixed:** clock-skew arithmetic removed — the total is server-computed (`session-metadata-store.ts:149`)
- ✅ **Fixed:** `useScheduleSessionModes` and its shared query key are deleted (`schedules/api/queries.ts`)
- ✅ **Fixed:** the seventh `?? "fresh"` default is gone with that hook

### Findings
- 🔴 **Critical:** An open run is closed only by the prompt response (`acp-runtime.ts:1165`), so every abnormal end leaks `runStartedAt` to disk. `tearDownSession` drops the active-prompt slot at `:793`, `recycleAgentForEnv` clears the whole map at `:685`, and the agent-exit handler at `:619` closes channels after the process died — none call `finishRun`. That fire's time never reaches `runTotalMs`, and because the entry is persisted, the row reads `Running now` forever, across pod restarts. (`packages/agent-runtime/src/modules/acp/services/acp-runtime.ts:1165`)
  **Fix:** Make "an open run is closed exactly once" an invariant of removing a session's active-prompt slot, not a side effect of the happy path: call `finishRun` from every path that clears `activePromptBySession` — turn response, `tearDownSession`, env recycle, agent exit — or route all four through one helper that does it. A store-side reconciliation at boot must also fold or drop a `runStartedAt` that no live turn owns, since the flag outlives the process that set it.
- 🟡 **Warning:** `startRun` runs before the queue check at `:1428`, so a fire that queues behind an active turn is timed from arrival, not from when the agent starts it. Worse, the second `startRun` overwrites the first fire's `runStartedAt`, so the in-flight fire loses its start and one `finishRun` is then charged the queued fire's wait. Overlapping fires on one continuous schedule make `runTotalMs` exceed real run time. (`packages/agent-runtime/src/modules/acp/services/acp-runtime.ts:1419`)
  **Fix:** Time a run from the point the prompt reaches the agent, not from the point it arrives at the runtime: call `startRun` inside `forwardPromptToAgent`, which covers both the direct send and the `advanceQueue` promotion. Make `startRun` refuse to overwrite an open run, so a concurrent fire can never discard another's start stamp.
- 🟡 **Warning:** The in-flight label does not make the count singular, so a session with one finished fire and one running renders `Running now · 1 runs, 42s so far`. (`packages/ui/src/modules/sessions/components/session-row.tsx:297`)
  **Fix:** Derive the noun once and use it in every branch that prints a count, e.g. `` `${runs} ${runs === 1 ? "run" : "runs"}` `` — the finished branch already special-cases 1, so keep the two consistent.
- 🟡 **Warning:** The PR description still documents the replaced mechanism. It promises a span line (`11/08/2026 10:00 → …`) and `Running for 2m 30s`, explains "Why only fresh-session schedules", and names `useScheduleSessionModes` and `durationBetween` — none of which the head tree contains, while the new persisted run accounting is undeclared. (reported by issue-fit)
  **Fix:** Rewrite the body when a later commit replaces an earlier commit's mechanism: quote only strings the code emits, drop symbols the branch no longer defines, and declare new persisted state and runtime instrumentation.
- 🟡 **Warning:** `docs/architecture/persistence.md` gains the run-accounting text but keeps `Last verified: 2026-08-11`, while the commit is dated 2026-08-12. `docs/guidelines/documentation-guidelines.md:34` makes that header mandatory on every edit. (`docs/architecture/persistence.md:3`, reported by doc-drift)
  **Fix:** Bump `Last verified:` in the same commit that edits any page carrying the header.
- 🟢 **Suggestion:** The `_meta.platform` wire shape is hand-declared three times — the Zod schema (`session-metadata-store.ts:15`), the writer (`acp-runtime.ts:1647`), and the UI's `PlatformMeta` (`acp-session-ops.ts:12`) — and this PR adds the same three fields to each by hand, with the UI decoding by cast. Declaring it once in a contract package and parsing it would make the next field a one-file change.

The `Last verified` finding is merged into `### Findings` above (`persistence.md:3`).

### Issue Fit

All four findings concern the same defect — the description documents the mechanism the second commit replaced — and are merged into one entry in `### Findings` above.

The leaked-run finding and the duplicated `_meta.platform` shape are merged into `### Findings` above. Two further findings are dropped after full-file verification, because the module already reads the other way: `recordSeen` and `set` compute values in the store today (`session-metadata-store.ts:84,91`), and `sessionMetadata` is one of twelve optional deps on `AcpRuntimeDeps`, consistent with `backgroundWork`. Both are pre-existing module-wide conventions, not defects this PR introduces.

The singular-count finding is merged into `### Findings` above. Two findings are dropped after checking `tooltip.tsx`: `HintTooltip` renders its `<button>` through a Radix `Trigger asChild`, so the forwarded `onClick` lands on that button and the row keeps its own activation — and the nested-control concern applies equally to the pre-existing overflow-menu button at `:175`, so it is not new here.

### Verdict
REQUEST_CHANGES — the new run accounting is the right mechanism and it fixes both previous warnings, but an open run is closed on one path only, so any teardown strands a session on `Running now` and loses that fire's time.

---

packages/agent-runtime/src/modules/acp/infrastructure/session-metadata-store.ts
```diff
@@ -119,6 +136,38 @@ export function createSessionMetadataStore(
...
+          [sessionId]: { ...existing, runStartedAt: now() },
+        },
+      });
+    },
+    finishRun(sessionId) {
+      const { sessions, tombstones } = store.read();
+      const existing = sessions[sessionId];
+      if (!existing?.runStartedAt) return;
+      const { runStartedAt, ...rest } = existing;
+      const elapsed = Date.parse(now()) - Date.parse(runStartedAt);
```
dam-code-guardian: 🟡 **Warning:** `Math.max(0, elapsed)` guards a negative elapsed but not `NaN`, which is what an unparseable stored `runStartedAt` yields — the schema accepts any string (`:21`), not a datetime.

The consequence is larger than a wrong duration. Verified in `node`: `Math.max(0, NaN)` is `NaN`, `JSON.stringify` writes it as `null`, and `z.number()` then rejects it — so the state transform at `:33` drops the **whole** entry. The session loses `meta`/`createdAt`/`seenAt` and decodes as if it never existed.

**Fix:** Sanitize a computed number against non-finiteness, not just sign, whenever the result is persisted through a schema that would reject it — a per-field parse failure that discards the entire record must never be reachable from arithmetic.

(reported by typescript-engineering)

packages/ui/src/modules/sessions/components/session-row.tsx
```diff
@@ -272,3 +286,16 @@ function SessionIndicators({
...
+
+/** Time the schedule's fires have spent running, summed over every fire this
+ *  session served. Null until the first fire finishes — including on sessions
+ *  that predate the accounting. */
+function runTimeLabel(s: SessionView): string | null {
+  const runs = s.runCount ?? 0;
+  const total = formatDuration(s.runTotalMs ?? 0);
+  const counted = `${runs} ${runs === 1 ? "run" : "runs"}`;
+  if (s.runStartedAt)
+    return runs ? `Running now · ${counted}, ${total} so far` : "Running now";
```
dam-code-guardian: 🟡 **Warning:** `"so far"` is attached to a total that excludes the fire still running.

`runTotalMs` and `runCount` only advance when a fire closes (`finishRun`), so this duration covers the *finished* fires and never moves while the tooltip is open. On a long fire the number understates the session by that fire's whole length, while "so far" reads as though the live one were included.

**Fix:** When a label combines a live in-flight state with counters that only advance on completion, either name what the number covers (`1 run done, 42s`) or add the in-flight elapsed computed from `runStartedAt` — never attach "so far" to a total that omits the item still running.

(reported by react-ui-engineering)

packages/agent-runtime/src/modules/acp/services/acp-runtime.ts
```diff
@@ -926,6 +928,11 @@ export function createAcpRuntime(deps: AcpRuntimeDeps): AcpRuntime {
       channel: entry.channel,
       originalId: entry.originalId,
     });
+    // A prompt off a machine-driven channel is a scheduled fire, not a person
+    // typing — only those are timed. Timed from here rather than from arrival,
+    // so a fire that queued behind another turn isn't charged its wait.
+    if (nonViewerChannels.has(entry.channel))
```
dam-code-guardian: 🟡 **Warning:** This gate is any machine-driven channel, but the PR body and `persistence.md` both scope run accounting to "scheduled fires".

`TriggerSessionDriver` attaches with `viewer: false` for every caller (`trigger-session-driver.ts:79`), and the `experiment-execute` plugin uses the same driver — so `experiment_execute` sessions accumulate `runStartedAt`/`runTotalMs`/`runCount` in the PVC state file, while the UI renders the tooltip for scheduled rows only. That is persisted state on a second session type that no document declares.

**Fix:** Describe new persisted state by the condition that actually writes it, not by the one use case that reads it — either widen the body and doc wording to all machine-driven turns, or narrow the gate to the scheduled session type so the claim becomes true.

(reported by issue-fit)

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `be07e8d`

## PR #3273: feat(sessions): time scheduled fires and report them on the session list's clock
**Author:** xjacka | **Branch:** feat/scheduled-run-time-tooltip → main | **Changes:** +216 −15 (10 files)

### Summary
The fix commit closes an open run on all four paths that can end one, and adds a boot fixup that discards a stamp outliving its process. The 🔴 and all three 🟡 of the last review are resolved; the replacement mechanism carries two new defects in the store itself.

### Changes since last review
Previous HEAD: `98acbe2` (2026-08-12T09:35:58Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** every path that clears `activePromptBySession` now closes the run — `endActivePrompt` on turn end and `tearDownSession`, `finishOpenRuns` on env recycle and agent exit (`acp-runtime.ts:955,962`)
- ✅ **Fixed:** a `runStartedAt` that outlived the pod is dropped at boot rather than folded in (`session-metadata-store.ts:78`)
- ✅ **Fixed:** `startRun` moved into `forwardPromptToAgent`, so a queued fire is timed from execution, not arrival (`acp-runtime.ts:934`)
- ✅ **Fixed:** `startRun` refuses to overwrite an open run, so a concurrent fire cannot discard another's start (`session-metadata-store.ts:142`)
- ✅ **Fixed:** the count is singular in every branch (`session-row.tsx:296`)
- ✅ **Fixed:** the PR body now matches the head tree claim by claim (reported by issue-fit)
- ✅ **Fixed:** `persistence.md` carries `Last verified: 2026-08-12` (reported by doc-drift)

### Findings
- 🔴 **Critical:** `set()` rebuilds the entry from a fixed field list, so it silently drops `runStartedAt`, `runTotalMs` and `runCount`. A `session/resume` carrying `_meta.platform` — the UI's and CLI's mode change — reaches it via `acp-runtime.ts:1307`, whose own comment promises it preserves the fields it omits. So switching a scheduled session's mode zeroes its accumulated run totals, and doing it mid-fire deletes the open stamp so that run's time is lost entirely. (`packages/agent-runtime/src/modules/acp/infrastructure/session-metadata-store.ts:106`, reported by typescript-engineering)
  **Fix:** A partial-update writer must spread the existing entry and override only the fields it owns (`{ ...existing, meta }`), never re-enumerate the entry shape — an allowlist rebuild silently discards every field added later. The sibling writers here (`recordActivity`, `recordSeen`, `startRun`) already do it that way.
- 🟡 **Warning:** `Math.max(0, elapsed)` guards the sign but not `NaN` (`packages/agent-runtime/src/modules/acp/infrastructure/session-metadata-store.ts:158`)
- 🟡 **Warning:** `"so far"` is attached to a total that excludes the running fire (`packages/ui/src/modules/sessions/components/session-row.tsx:298`)
- 🟡 **Warning:** Run accounting is documented as "scheduled fires", but the gate is any machine-driven channel. `TriggerSessionDriver` attaches with `viewer: false` for every caller and the `experiment-execute` plugin uses the same driver, so `experiment_execute` sessions accumulate `runStartedAt`/`runTotalMs`/`runCount` in the PVC state file while only scheduled rows render them — undeclared persisted state on a second session type. (`packages/agent-runtime/src/modules/acp/services/acp-runtime.ts:934`, reported by issue-fit)
  **Fix:** Describe new persisted state by the condition that actually writes it, not by the one use case that reads it — either widen the body and doc wording to all machine-driven turns, or narrow the gate to the scheduled session type so the claim becomes true.
- 🟡 **Warning:** `docs/architecture/platform-topology.md:52` owns what `session/list` exposes to clients and documents the analogous read-state field, but was not extended, and its `Last verified: 2026-08-10` predates this change. A reader of the owning page cannot learn that run time rides the same metadata. (reported by doc-drift)
  **Fix:** When a change adds a field to the session-metadata surface that clients read off `session/list`, extend the platform-topology paragraph that owns that surface with one semantic sentence and bump that page's `Last verified` in the same PR.
- 🟡 **Warning:** `docs/architecture/persistence.md:107` describes the run accounting but omits its durability limit: the boot fixup drops a surviving `runStartedAt`, so a fire interrupted by a pod restart is discarded rather than recovered and the totals silently under-count. On the page whose contract is what survives each lifecycle event, an unstated loss reads as durable. (reported by doc-drift)
  **Fix:** When persisted state has a durability limit — an entry that cannot outlive the pod and is discarded rather than recovered on boot — state that limit on `persistence.md` next to the state it describes.
- 🟢 **Suggestion:** The `_meta.platform` wire shape is hand-declared three times and the copies have now drifted — the UI's `PlatformMeta` gained the run fields while the CLI's identical decoder did not, so the CLI cannot populate the `SessionView` fields the contract package declares (`packages/ui/src/modules/sessions/api/acp-session-ops.ts:6` vs `packages/cli/src/modules/chat/infrastructure/acp-session-client.ts:24`). Declaring the shape once in the contract package would make a new field a compile-time obligation at each consumer. (reported by typescript-engineering)

Both findings are merged into `### Findings` above (`platform-topology.md:52`, `persistence.md:107`). The previously reported stale `Last verified` on `persistence.md` is confirmed resolved.

### Issue Fit
No linked issue — scope checked against the PR description only. The rewritten body now matches the head tree claim by claim, so the previous review's body 🟡 is resolved. The remaining finding (run accounting scoped to "scheduled fires" while the gate admits every machine-driven turn) is merged into `### Findings` above.

🔁 The duplicated `_meta.platform` shape from the previous review is still present, now with a drifted CLI copy — merged into `### Findings` above. Two new findings, the `set()` field-list rebuild and the unguarded `NaN`, are also merged there.

The singular-count fix is confirmed complete. The one finding — `"so far"` attached to a total that excludes the in-flight fire — is merged into `### Findings` above.

### Verdict
REQUEST_CHANGES — the run-closing mechanism is now correct on every teardown path, but `set()` rebuilds the entry from a fixed field list, so a mode change on a scheduled session zeroes its accumulated totals.

---

packages/ui/src/modules/sessions/components/session-row.tsx
```diff
@@ -256,9 +261,18 @@ function SessionIndicators({
             aria-label="Channel session"
           />
         ))}
-      {scheduled && (
-        <Time size={16} className="text-foreground" aria-label="Scheduled" />
-      )}
+      {scheduled &&
+        (runTime ? (
+          <HintTooltip
```
dam-code-guardian: 🟡 **Warning:** this tooltip is the *only* carrier of the run time, and a tap never opens it — so the feature is absent on touch. Verified at the pins rather than from memory: `@radix-ui/react-tooltip` 1.2.8's trigger returns early from `onPointerMove` when `pointerType === "touch"`, its `onFocus` skips `onOpen` while `isPointerDownRef` is set, and this repo's own wrapper `preventDefault()`s any focus that is not `:focus-visible` (`tooltip.tsx:65`), which a tap does not set. All three paths are closed.

That matters here because nothing else renders the value: `runTotalMs`/`runCount` reach no other UI call site (the schedules module renders neither), `formatDuration` is called only from `runTimeLabel`, and `main` had a plain icon — so the tooltip arrives with the data. The sessions sidebar is a first-class touch surface: this same file carries a `LONG_PRESS_MS` touch path (`:33,:127`) and `chat-view.tsx` gives the sidebar `max-md:!w-full`. Radix also sets `aria-describedby` only while open, so a swipe-navigating screen reader never hears it either. (reported by react-ui-engineering)
  **Fix:** Never let a hover-triggered tooltip be the sole carrier of information that appears nowhere else. Give the value a pointer-type-independent path — a tap-to-open `Popover`, or the duration inline in the row's metadata line — and keep `HintTooltip` for detail that amplifies something already visible.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `eb4a555`

### Summary
The final commit fixes the 🔴 at the writer and closes the five 🟡 of the last review. All six resolve genuinely; the one new finding is in the delivery path of the value the feature exists to show.

### Changes since last review
Previous HEAD: `be07e8d` (2026-08-12T13:47:05Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** `set()` spreads the existing entry, so a mode change keeps the run totals (`session-metadata-store.ts:108`)
- ✅ **Fixed:** `finishRun` rejects a non-finite elapsed before it reaches the schema (`session-metadata-store.ts:161`)
- ✅ **Fixed:** the gate now requires the session to belong to a schedule, so `experiment_execute` turns no longer accumulate totals (`acp-runtime.ts:943`)
- ✅ **Fixed:** the in-flight label reads `<n> runs finished, <total>` — the total's scope now matches its wording (`session-row.tsx:298`)
- ✅ **Fixed:** `platform-topology.md` documents the run-time surface and carries `Last verified: 2026-08-12` (reported by doc-drift)
- ✅ **Fixed:** `persistence.md` states the durability limit — an in-flight stamp is discarded at boot, so the totals under-count that fire (reported by doc-drift)
- 🔁 **Still present:** the `_meta.platform` shape is hand-declared three times, and the CLI decoder still lacks the run fields (`acp-session-client.ts:24`)

### Findings
- 🟡 **Warning:** the tooltip is the only carrier of the run time, and a tap never opens it (`packages/ui/src/modules/sessions/components/session-row.tsx:266`) — reported by react-ui-engineering
- 🟡 **Warning:** The body's "Why this needed run accounting" section still states that a run opens for "a prompt arriving on a machine-driven channel (`viewer: false`)". The final commit narrowed the gate to machine-driven **and** schedule-owned, and the two architecture pages carry the narrowed wording — the body is the only stale copy. It also leaves both shared-store changes undeclared. (reported by issue-fit)
  **Fix:** State a gate as the full conjunction the code tests, and update every prose copy of it in the commit that changes it. Declare any change to a shared store method that alters behaviour for callers outside the feature's own scope.
- 🟢 **Suggestion:** the `_meta.platform` wire shape is still hand-declared three times, and the CLI's copy still lacks the three run fields the contract package now declares, so the CLI cannot populate them (`packages/cli/src/modules/chat/infrastructure/acp-session-client.ts:24`) — reported by typescript-engineering
- 🟢 **Suggestion:** `finishOpenRuns()` closes the runs but leaves the slots in `activePromptBySession`, while `recycleAgentForEnv` clears them right after. The asymmetry is pre-existing on `main` and currently unreachable — `agentExited` is terminal, so nothing reads the map afterwards — but the invariant `endActivePrompt` states would read better if `finishOpenRuns` released what it closed. (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:974`) — reported by typescript-engineering

✅ No findings. Both previously reported findings are confirmed resolved: `platform-topology.md` now documents the run-time surface it owns, and `persistence.md` states the in-flight stamp's durability limit. Both pages carry `Last verified: 2026-08-12`, verified against `main` per page and per commit.

### Issue Fit
No linked issue — scope checked against the PR description only. The body-staleness finding is merged into `### Findings` above. Two further 🟢: the "What" section still describes the bare `Running now` label the final commit changed, and the Verification section's test counts and k3s run predate `eb4a555b`, so no stated validation covers the narrowed gate or the `set()` rewrite.

🔁 The duplicated `_meta.platform` shape from the previous review is still present, with the CLI copy still drifted — merged into `### Findings` above. The teardown-asymmetry finding is merged there as a 🟢 after baseline verification. Three further findings are dropped after full-file verification: `sessionMetadata?` optionality is a pre-existing module-wide convention (one of twelve optional deps on `AcpRuntimeDeps`), already assessed in the previous round, and the two `domain/`-layering findings ask for placement changes that carry no defect risk in a codebase where the sibling writers live in the same adapter.

The one finding is merged into `### Findings` above. Both previously reported defects are confirmed genuinely fixed — the trigger stops click propagation, and the count is singular in every branch — and the round-4 label change is confirmed correct against `finishRun`, which increments the totals only on completion.

### Verdict
REQUEST_CHANGES — the writer fix, the `NaN` guard and the narrowed gate are all correct, and the docs now match; but the run time the PR exists to surface is unreachable on touch, where the sessions sidebar is a full-width primary surface.

---

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `076a8167`

### Summary

No new findings. HEAD is a merge of `origin/main` into the feature branch; the PR carries no new feature commits since `eb4a555`. All open findings from the last review carry forward unchanged.

### Changes since last review

Previous HEAD: `eb4a555` (2026-08-14T07:01:00Z) — verdict REQUEST_CHANGES

- ➡️ **No feature changes** — HEAD `076a8167` is a merge commit; all six commits since `eb4a555` are from `main`. The open findings below were not addressed.

### Findings

_(All carried from prior review at `eb4a555`. No new findings.)_

- 🟡 **Warning:** the tooltip is the only carrier of the run time, and a tap never opens it (`packages/ui/src/modules/sessions/components/session-row.tsx:266`) — **Fix:** replace `HintTooltip` with a component that opens on both touch and hover, or echo the run time as visible text alongside the clock icon.
- 🟡 **Warning:** the PR body still states the gate as "a prompt arriving on a machine-driven channel (`viewer: false`)" but the code tests the full conjunction `nonViewerChannels && (ScheduleCron || scheduleId)` — the narrowed wording is in the architecture pages but not in the body.
  **Fix:** align the PR description to the code: state the gate as the conjunction it is.

- 🟢 **Suggestion:** the CLI decoder (`packages/cli/src/modules/chat/infrastructure/acp-session-client.ts:68`) still lacks `runStartedAt`, `runTotalMs`, and `runCount` — the contract package declares them but `toSessionView` does not map them.
- 🟢 **Suggestion:** `finishOpenRuns()` closes runs but does not remove entries from `activePromptBySession`; `endActivePrompt` does both. The asymmetry is harmless (callers clear the map immediately after) but inconsistent (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:653`).

### Approval bar

The two 🟡 findings remain open. This review passes when both are resolved: (1) run time visible or accessible on touch, (2) PR body gate description matches the code.

### Skills

| Skill | Status | Findings |
|---|---|---|
| doc-drift | ✅ ran | 0 |
| issue-fit | ✅ ran | 0 |
| typescript-engineering | ✅ ran | 0 |
| react-ui-engineering | ✅ ran | 0 (pre-existing patterns, out of delta scope) |

```json
{"schema":"findings/v1","pr":3273,"sha":"076a8167c9acc333097df9819e2c9219062f1bb4","kind":"re-review","trigger":"review-request","scope":"delta","verdict":"REQUEST_CHANGES","findings":[{"id":"tooltip-touch","severity":"warning","file":"packages/ui/src/modules/sessions/components/session-row.tsx","line":266,"summary":"Run time inaccessible on tap — tooltip only opens on hover/focus","carryover":true,"prior_sha":"eb4a555"},{"id":"body-gate-wording","severity":"warning","file":null,"line":null,"summary":"PR body gate description (viewer:false only) does not match code conjunction (nonViewerChannels AND schedule-owned)","carryover":true,"prior_sha":"eb4a555"},{"id":"cli-decoder-run-fields","severity":"suggestion","file":"packages/cli/src/modules/chat/infrastructure/acp-session-client.ts","line":68,"summary":"CLI toSessionView does not map runStartedAt/runTotalMs/runCount from _meta.platform","carryover":true,"prior_sha":"eb4a555"},{"id":"finish-open-runs-asymmetry","severity":"suggestion","file":"packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts","line":653,"summary":"finishOpenRuns closes runs without removing slots from activePromptBySession; endActivePrompt does both","carryover":true,"prior_sha":"eb4a555"}]}
```

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `4dd74e9b`

## PR #3273: feat(sessions): time scheduled fires and report them on the session list's clock
**Author:** xjacka | **Branch:** feat/scheduled-run-time-tooltip → main | **Trigger:** review-request (delta scope) | **Changes:** +331 −18 (19 files)

### Summary
One new commit, `4315a50`. It moves the run time out of the tooltip into the row's meta line and fills the CLI decoder's missing `_meta.platform` fields. Both blocking findings from the previous round are resolved.

### Changes since last review
Previous HEAD: `076a8167` (2026-08-14T07:50:27Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** the run time is now visible text in the meta line, so it reads on touch; the clock is a plain indicator again and keeps its `aria-label` (`packages/ui/src/modules/sessions/components/session-row.tsx:157`)
- ✅ **Fixed:** the PR body now states the gate as the conjunction the code tests
- ✅ **Fixed:** the CLI decoder maps `runStartedAt`, `runTotalMs` and `runCount`, plus `threadTs`, `seenAt` and `running` (`packages/cli/src/modules/chat/infrastructure/acp-session-client.ts:86`)
- ✅ **Fixed:** the `finishOpenRuns` / `activePromptBySession` asymmetry is gone — turn edges now run through `prompt-scheduler`'s `onTurnEnded` into `finishRun`, and neither symbol remains in the tree
- 🔁 **Still present:** the `_meta.platform` shape and its `toSessionView` decoder are declared twice, in the CLI and the UI (`packages/ui/src/modules/sessions/api/acp-session-ops.ts:9`)

### Findings

✅ No findings. Both architecture pages carry the run-accounting prose with `Last verified: 2026-09-01`; the new commit is presentation plus a decoder mapping, so no documented responsibility moves.

### Issue Fit
✅ No findings. No linked issue, so scope was judged against the PR body; the "Text, not a tooltip" decision matches the commit.

🔁 1 finding from the previous review still present (see review at `076a8167`) — the duplicated `_meta.platform` declaration. This commit closes its drift half: the two copies are identical again. `agent-runtime-api` already exports `podSessionSchema` for this shape and both packages depend on it, so one derived type and one shared decoder would stop the next field from drifting.

✅ No findings (files=1). The dropped `HintTooltip` and `formatDuration` imports have no remaining references, `SessionIndicators`' only call site drops the `session` prop with it, and `truncate` applies because the meta span is a blockified flex item.

### Verdict
APPROVE — the run time reaches every pointer type, the CLI decoder is in sync, and the only open item is a pre-existing 🟢 seam.

---

PR #3280, opened 2026-08-12
feat(ui): make the "Update available" pill the way to update a sandbox

### LINKED ISSUE #3137: UI - Make the "Update available" pill the way to update a sandbox
## Context

A sandbox shows an "Update available" pill next to its name when its template has moved past the image the sandbox captured at create (#1077). The pill is where you notice it — but the only place you can act on it is a notice much further down Sandbox Setup, under Image, and that action is labelled "Upgrade" while everything else calls it an update.

Image

Image

## Problem

The place that tells you about the update can't perform it. You see the pill, click it, nothing happens, and then have to scroll the page to find where the action actually lives. Once you find it, the button says "Upgrade" — a different word for the same thing, so it's not obvious it's even the update the pill was talking about. Two small gaps stacking into "am I in the right place?"

The notice itself also carries detail nobody needs at that moment: it spells out the full image reference the template now ships. That is the same from/to detail the confirmation dialog already shows, so the page states it twice and the second place is the one you have to hunt for.

## Goal

One place tells you an update is available, and starting it happens right there.

## User Stories

- As a user, I want to start the update from the "Update available" pill so that I don't have to hunt the page for where the action lives.
- As a user, I want one consistent word for the action so that I can tell the pill and the confirmation mean the same thing.
- As a user, I want the Image section to stay short so that the sandbox's current image is easy to read.

## Scope

The pill becomes the only entry point. The inline notice under Image ("Update available. The template now ships …") is removed — the confirmation dialog already names the image you move from and to, so no detail is lost. The Image section keeps showing the sandbox's current image only. Language mirrors across the pill and the confirmation: same verb, same noun.

## Proposed solution

Make the pill actionable, opening the same confirmation the notice's button opens today. Remove the notice. Rename "Upgrade" to "Update" across the pill, the button, and the confirmation.

## Open Questions

- A clickable status pill is a new pattern in the product — assigned to @jamiejabbouribm to decide whether this becomes an actionable badge or something more obviously button-like, and to keep it consistent with the design system.
- With the notice gone, is the pill discoverable enough on its own, or does the Image section need a quieter hint that an update exists?

## Additional resources

- #1077 — Let users upgrade an existing agent to the latest template version

kapetr:
The place that tells you a sandbox is out of date can now perform the update.

An `Update` action replaces the "Update available" pill, on the sandbox row and in the sandbox header. Hovering it names the image the template now ships and links to the harness's release notes; clicking it opens the confirmation that applies it. The notice under Image is gone — it repeated what the confirmation says, and it was the part you had to scroll to find. "Upgrade" is now "Update" everywhere.

The sandbox list also gets a banner when more than one sandbox is behind, so catching up on all of them is one action instead of a visit per sandbox.

Closes #3137

## Decisions worth a look

- **The card opens on hover, not click, and has no close ×.** The mock draws a popover with a dismiss button; hover-driven, it closes itself. Consequence: no card on touch, where the button taps straight through to the confirmation. Radix cancels `touchstart` on a hover trigger, which cancels the emulated click with it, so the trigger renders bare where there is no hover to reveal.
- **"See what changed" only appears where a page exists** — Claude Code, Codex, and IBM Bob. PI Agent and the preconfigured templates show the card without the link rather than pointing it somewhere that cannot answer the question. Matching is on the image's name, the only harness identity a sandbox carries.
- **Update all is a loop over the existing per-agent endpoint,** not a new bulk one. Each apply keeps the compare-and-swap on the image it was confirmed against, so a template moving mid-batch is rejected rather than applied silently. The batch reports once, naming the first failure, instead of one toast per sandbox. Updates run one at a time, since each rolls a pod.

The architecture page said applying an update is "an explicit per-Agent user action"; that sentence now covers the batch too.


packages/ui/src/components/ui/hover-card.tsx
```diff
@@ -0,0 +1,89 @@
...
+export function HoverCardTappableTrigger({
+  children,
+}: {
+  children: React.ReactNode;
+}) {
+  return (
+    <HoverCardTrigger asChild>
+      <span
+        className="inline-flex"
+        onTouchStart={(event) => event.stopPropagation()}
```
dam-code-guardian: 🔴 **Critical:** this cannot suppress Radix's handler. `HoverCardTrigger` renders through `PopperPrimitive.Anchor` with `asChild`, so this `<span>` *becomes* the trigger — there is no level below it to intercept at. Slot's `mergeProps` chains the child handler and then Radix's own `onTouchStart` unconditionally, with no `defaultPrevented` or propagation check between them, so `event.preventDefault()` still runs on this same element and the emulated click stays cancelled. The Update button therefore stays dead on tap — the defect this wrapper was added to fix. Verified against the pinned `@radix-ui/react-hover-card@1.1.23` and `@radix-ui/react-slot@1.3.3`.

**Fix:** a handler on an `asChild` trigger wrapper can never pre-empt the primitive's own — put the guard on the element the touch lands on. Move `onTouchStart={(e) => e.preventDefault()}`-defeating logic to the child `Button` (or re-dispatch the click yourself in an `onTouchEnd`), and cover the tap path with a test so the workaround cannot silently rot.

packages/ui/src/modules/agents/components/agent-row.tsx
```diff
@@ -96,6 +102,15 @@ export function AgentRow({
         )}
       </div>
       <div className="flex shrink-0 items-center gap-1">
+        {/* Both the action and its portaled hover card sit inside the row's
+            click target, so their clicks must not also open the sandbox. */}
+        <span onClick={(e) => e.stopPropagation()}>
```
dam-code-guardian: 🟡 **Warning:** this wrapper renders even when `UpdateAvailableAction` returns `null`, and an empty `<span>` is still a flex item — so every row without a pending update gains one extra `gap-1`. That is most rows. The sibling `ContributionFailuresBadge` sets the precedent: it returns `null` with no wrapper and contributes no item.

**Fix:** a click guard belongs inside the component it guards, not around it — move the `onClick` wrapper into `UpdateAvailableAction` below its `if (!update) return null`, so a component that renders nothing occupies no space.

packages/ui/src/modules/agents/hooks/use-update-sandbox.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+/** The banner only offers a batch from two up, but nothing stops a caller
+ *  passing one, and "1 sandboxes" would be the tell. */
+function batchSubject(total: number): string {
+  return total === 1
+    ? "sandbox to the version its template now ships?"
+    : "sandboxes to the versions their templates now ship?";
+}
+
+/** When a batch takes effect — the same two facts the single-sandbox
```
dam-code-guardian: 🟡 **Warning:** this doc block describes `batchApplyNote`, but it sits above `whatsNewLink`, which then carries two unrelated doc comments while `batchApplyNote` has none. A reader of `whatsNewLink` is told about batch wording and per-case subjects.

**Fix:** keep each doc block directly above the function it documents — move this one down to `batchApplyNote`.

packages/ui/src/modules/agents/utils/template-update.ts
```diff
@@ -0,0 +1,31 @@
+/** Where a harness publishes what changed between image versions.
+ *
+ *  Keyed on a fragment of the image's *name*, because that is the only harness
+ *  identity a sandbox carries — a template's name is admin-editable and its
+ *  `docsUrl` points at the harness's docs, not its release history. A fragment
+ *  rather than the whole name, so the variants of one harness all resolve:
+ *  `claude-code`, `claude-code-vm`, and a locally built `platform-claude-code`
+ *  are the same product. Each points at a release *list*: one update can span
+ *  several versions. */
+const WHATS_NEW_BY_HARNESS: ReadonlyArray<readonly [string, string]> = [
```
dam-code-guardian: 🟡 **Warning:** this hardcodes per-harness catalog knowledge in the client, but templates are operator-installed catalog data already crossing the contract as `TemplateSpec.docsUrl`. Adding a harness or fixing a URL now needs a UI release instead of a template change.

**Fix:** catalog facts about a template belong to the templates context and should reach the client through the contract package — add a release-notes field to `TemplateSpec`/`TemplateView` and read it, rather than duplicating a per-harness table in a client module.

packages/ui/src/modules/agents/utils/template-update.ts
```diff
@@ -0,0 +1,31 @@
...
+  const segment = image.slice(image.lastIndexOf("/") + 1);
+  return segment.split("@")[0]!.split(":")[0]!;
+}
+
+/** Null for a harness that publishes nothing we know of, which hides the link
+ *  rather than pointing it somewhere that can't answer the question. */
+export function whatsNewUrl(image: string): string | null {
+  const name = imageName(image);
+  return (
+    WHATS_NEW_BY_HARNESS.find(([fragment]) => name.includes(fragment))?.[1] ??
```
dam-code-guardian: 🟡 **Warning:** harness identity is derived by substring-matching the image name, and the comment above calls the image "the only harness identity a sandbox carries" — but `AgentView.templateId` exists, and `harness-card.tsx:19` already keys per-harness data on it. One domain concept now has two identity keys. The substring form is also load-bearing in a way the name hides: `dam-agents/bob` matches on `bob`, so any future image whose name merely contains `bob` inherits the IBM Bob changelog.

**Fix:** one domain concept gets one identity key — key harness lookups on `templateId` as primary and fall back to the image only where `templateId` is null, then correct the comment's premise. Applies to every per-harness lookup in the `agents` and `sandboxes` modules.

packages/ui/src/modules/agents/api/mutations.ts
```diff
@@ -198,14 +198,17 @@ export function useRestartAgentMutation() {
   });
 }
 
-// Upgrading rolls a running pod onto the new image — refresh the budget
-// meter alongside, like the other lifecycle mutations.
-export function useUpgradeAgentMutation() {
+// Updating rolls a running pod onto the new image — refresh the budget
+// meter alongside, like the other lifecycle mutations. `silent` is for the
+// bulk path, which reports every sandbox's outcome in one toast of its own.
+export function useUpgradeAgentMutation(opts?: { silent?: boolean }) {
```
dam-code-guardian, line 164: 🟡 **Warning:** the user-facing concept becomes "update" (`useUpdateSandbox`, `template-update.ts`, "Failed to update sandbox") while the contract procedure and this hook stay `upgrade` (`trpc.agents.upgrade`, `useUpgradeAgentMutation`). Neither term is in `docs/ubiquitous-language.md`, so one operation now has two names across the contract boundary.

**Fix:** pick one term for this operation, record it in `docs/ubiquitous-language.md`, and make the contract procedure, hook, file, and UI copy agree end to end.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `5d20e86`

## PR #3280: feat(ui): make the "Update available" pill the way to update a sandbox
**Author:** kapetr | **Branch:** feat/3137-update-pill → main | **Changes:** +545 −103 (17 files)

### Summary
The "Update available" pill becomes an Update action on the sandbox row and in the sandbox header. A hover card names the new image and links the harness release notes, the confirmation applies the update, and the sandbox list gains an "Update all" banner. The old notice under Image is removed.

### Findings
- 🔴 **Critical:** the tap fix in `HoverCardTappableTrigger` cannot suppress Radix's handler (`packages/ui/src/components/ui/hover-card.tsx:53`)
- 🟡 **Warning:** the always-rendered `<span>` wrapper adds a flex gap on every row without a pending update (`packages/ui/src/modules/agents/components/agent-row.tsx:107`)
- 🟡 **Warning:** `batchApplyNote`'s doc comment sits above `whatsNewLink` (`packages/ui/src/modules/agents/hooks/use-update-sandbox.tsx:48`)
- 🟡 **Warning:** the per-harness release-notes table is hardcoded in the client, not carried by the template contract (`packages/ui/src/modules/agents/utils/template-update.ts:10`)
- 🟡 **Warning:** harness identity is keyed on the image name while `templateId` already keys per-harness data (`packages/ui/src/modules/agents/utils/template-update.ts:28`)
- 🟡 **Warning:** the UI renames the operation to "update" while the contract procedure and hook stay "upgrade" (`packages/ui/src/modules/agents/api/mutations.ts:204`)
- 🟡 **Warning:** the architecture page states both paths confirm the exact image movement, but the batch confirmation shows only a count and a restart note (`docs/architecture/agent-lifecycle.md:76`) — reported by doc-drift.
  **Fix:** State the confirmation contract per path: the single-Agent apply confirms the exact image movement, the all-behind apply confirms the count and the restart consequence.
- 🟡 **Warning:** the same doc edit puts UI control placement into an architecture page, a detail this PR itself just moved (`docs/architecture/agent-lifecycle.md:76`) — reported by doc-drift.
  **Fix:** Describe the capability, not the widget locations — say the update is user-initiated per Agent or for all behind Agents at once, and leave control placement to the UI.
- 🟡 **Warning:** `Closes #3137` closes an issue whose two Open Questions are still unanswered, including the pill affordance assigned to another owner (`packages/ui/src/modules/agents/components/update-available-action.tsx:32`) — reported by issue-fit.
  **Fix:** Get the named decision owner's answer before merge, or drop the closing keyword; a PR must not close an issue whose explicitly-assigned open questions are undecided.
- 🟡 **Warning:** the "Update all" banner adds a second entry point, while #3137's scope states the pill becomes the only one (`packages/ui/src/modules/agents/views/list-view.tsx:108`) — reported by issue-fit.
  **Fix:** Track product surfaces beyond the linked issue's stated scope in their own issue with design sign-off, rather than shipping them under another issue's closing keyword.

Both findings are blocking and listed above: the confirmation-contract mismatch and the control-placement detail, both at `docs/architecture/agent-lifecycle.md:76`.

### Issue Fit
Linked issue: #3137 (closing keyword). Both blocking findings are listed above — the unanswered Open Questions and the out-of-scope second entry point.

- 🟢 **Suggestion:** the hover card restates the full image reference, the exact duplication #3137 gave as its reason for deleting the notice ("the confirmation dialog already names the image you move from and to") — worth confirming against the Figma before merge (`packages/ui/src/modules/agents/components/update-available-action.tsx:52`).

Three blocking findings, listed above with their fixes: the hardcoded release-notes table (`template-update.ts:10`), the competing harness identity key (`template-update.ts:28`), and the update/upgrade split across the contract boundary (`mutations.ts:204`).

- 🟢 **Suggestion:** `whatsNewUrl` is a module-internal helper under `agents/utils/`, yet `sandboxes/components/sandbox-home-header.tsx` now reaches into `agents/hooks/` by path. Giving the `agents` module an explicit public surface would stop cross-module consumers importing internals.

The critical tap finding and two blocking findings are listed above: the inert `stopPropagation` wrapper (`hover-card.tsx:53`), the empty flex item (`agent-row.tsx:107`), and the misplaced doc comment (`use-update-sandbox.tsx:48`).

- 🟢 **Suggestion:** `restartsToApply` duplicates the `!(agent.state === "hibernated" || agent.overBudget)` predicate already in `use-sandbox-settings-form.ts:220`, and its own comment says it mirrors `sizeRestartsAgent`. One exported predicate would stop the two copies drifting (`packages/ui/src/modules/agents/hooks/use-update-sandbox.tsx:36`).

### Verdict
REQUEST_CHANGES — the touch workaround cannot work as written, so the tap path the PR set out to protect is unprotected, and eight further warnings remain open.

---

packages/ui/src/modules/agents/hooks/use-update-sandbox.tsx
```diff
@@ -0,0 +1,172 @@
...
+  );
+
+  const updateAll = useCallback(
+    async (agents: readonly AgentView[]) => {
+      const pending = pendingUpdates(agents);
+      if (pending.length === 0) return;
+      const restarting = agents.filter(
+        (a) => a.templateUpdate && restartsToApply(a),
+      ).length;
+      const msg = (
```
PetrBulanek, line 97: The batch confirmation shows a count and a restart note but never the images — in the running
app it reads "Update 2 sandboxes to the versions their templates now ship? Each applies it the
next time it starts." Each apply still sends `expectedToImage`, so the compare-and-swap binds
the user's consent to an image they were never shown, and when the batch spans two templates
there is no way to tell what either sandbox moves to. The architecture page as amended in this
PR says of both controls that "the user is shown the exact image movement and confirms", which
now overstates the list-level path — so either the batch needs to name what moves, or the page
needs to say that the batch trades that detail for a single action.

packages/ui/src/components/ui/hover-card.tsx
```diff
@@ -0,0 +1,89 @@
...
+ * that unconditionally, so no media query can gate it: a touchscreen laptop
+ * reports `hover: hover` and still fires `touchstart`. Stopping the event one
+ * level below the trigger keeps Radix's handler from ever seeing it, and leaves
+ * hover alone — React dispatches enter/leave to every element newly entered,
+ * the wrapper included.
+ *
+ * Needs a box of its own, so no `display: contents`: the wrapper is what the
+ * card measures itself against.
+ */
+export function HoverCardTappableTrigger({
```
PetrBulanek: This wrapper cannot keep Radix's touchstart handler from running, because the span *is* the
trigger: `asChild` makes Radix merge its own `onTouchStart` onto this same span (the rendered
span carries `data-state`), and Radix's Slot runs the child's handler and then its own —
`stopPropagation()` only gates listeners on *other* elements, not a co-located one. Radix's
`preventDefault()` also has no effect here in the first place: React 19 registers `touchstart`
on the root container with `passive: true`, so a `preventDefault()` from any React
`onTouchStart` is ignored. Dispatching a cancelable touchstart on this button confirms it —
`defaultPrevented` stays false with React and Radix alone, and only becomes true with a
listener registered `{ passive: false }`. So the emulated click is never cancelled and a bare
trigger would already be tappable; it would help to know which symptom you actually saw,
because the mechanism in the comment is not the one running.

PetrBulanek (changes requested):

kapetr:
@dam-code-guardian — three declined, with reasons.

**update vs upgrade (`mutations.ts:204`).** `agents.update` already exists and means editing a sandbox's settings, so the name is taken — renaming `upgrade` onto it would collide. The wire keeps `upgrade`, the user-facing term is "update". Deliberate, not drift.

**`Closes #3137` with open questions unanswered (`update-available-action.tsx:32`).** The maintainer's call; the closing keyword stays.

**"Update all" beyond the issue's stated scope (`list-view.tsx:108`).** Intended, and staying in this PR.

packages/ui/src/components/ui/hover-card.tsx
kapetr: Not a defect — Radix's `preventDefault()` can't do anything here.

React attaches `touchstart` to the page root as a *passive* listener (`react-dom-client.development.js:19251` marks `touchstart`, `touchmove` and `wheel` passive). The browser ignores `preventDefault()` from a passive listener.

So when Radix calls it in its own `onTouchStart`, nothing happens. The click that a tap produces is never cancelled, and the button responds to a tap on its own.

Credit to @PetrBulanek, who established this by dispatching a cancelable `touchstart` on the button and watching `defaultPrevented` stay false — true with React and Radix alone, and only becoming true once a listener is registered `{ passive: false }`.

deploy/helm/platform/values.yaml
```diff
@@ -1221,6 +1221,7 @@ agentTemplates:
     enabled: true
     displayName: "Claude Code"
     description: "Default Claude Code agent"
+    releaseNotesUrl: "https://code.claude.com/docs/en/whats-new"
```
dam-code-guardian, line 1224: 🟡 **Warning:** this contract migration drops the harness variants the deleted test pinned. `whatsNewUrl` matched a *fragment* of the image name, so `claude-code-vm` and a locally built `platform-claude-code` both resolved to the Claude Code notes — `__tests__/unit/template-update.test.ts` asserted exactly that under "treats a harness's variants as the same product". The replacement joins on `templateId`, and this file declares `releaseNotesUrl` on only 3 of 10 templates: `claude-code-vm` at line 1236 is `enabled: true`, a first-class shipped harness, and it silently lost its link. `values-local.yaml:41` overrides the same templates to `platform-claude-code`, the other case the test covered.

**Fix:** when a fix replaces a fuzzy match with an exact key, enumerate every input the old matcher resolved and give each one an explicit entry. Add `releaseNotesUrl` to `claude-code-vm` here, and check `values-local.yaml`'s overrides if a local template should keep the link.

packages/ui/src/modules/agents/views/list-view.tsx
```diff
@@ -97,6 +105,34 @@ export function ListView() {
...
+              <strong className="font-medium text-foreground">
+                {outdated.length} sandboxes
+              </strong>{" "}
+              out of date — newer images available upstream.
+            </span>
+          </p>
+          <Button
+            variant="ghost"
+            size="sm"
+            disabled={update.updatingAll}
```
dam-code-guardian: 🟡 **Warning:** this gates only on `updatingAll`, so a single-row update in flight leaves the banner live and a second apply can start for the same sandbox. The row already gates on both signals — `use-agent-rows.ts:91` sets `updatePending: update.updatingId === agent.id || update.updatingAll` — so the two entry points disagree. The server's compare-and-swap turns the duplicate into a `CONFLICT` toast rather than a wrong image, but the user sees a failure for an update that succeeded. Reported by react-ui-engineering.

**Fix:** gate every entry point of one operation on all of its in-flight signals, not only its own. Mirror the row's predicate here.

```suggestion
            disabled={update.updatingAll || update.updatingId !== null}
```

packages/ui/src/modules/agents/hooks/use-update-sandbox.tsx
```diff
@@ -0,0 +1,187 @@
...
+          kind: "success",
+          message: `Updated ${updated} sandbox${updated === 1 ? "" : "es"}`,
+        });
+        return;
+      }
+      // Names the first reason rather than a bare count: every failed sandbox
+      // keeps its own Update, so the point of the toast is why to expect it.
+      emitToast({
+        kind: "error",
+        message: `Updated ${updated} of ${pending.length} sandboxes. ${failures.length} failed: ${failures[0]}`,
```
dam-code-guardian: 🟡 **Warning:** the toast names only `failures[0]`, and the sandbox name rides a fallback that a server-reported error suppresses. `getErrorMessage(err, "<name> failed")` returns the fallback **only** when the throwable carries no message (`lib/errors.ts:31`), so a server `CONFLICT` — the `TemplateMoved` case this loop exists to surface (`agents/router.ts:160`) — reports "The template changed since you reviewed the upgrade…" with no sandbox named. In a batch spanning several templates that is the one fact the user needs. Reported by react-ui-engineering.

**Fix:** never let an error-message fallback carry an item's identity in a per-item loop — prefix the name unconditionally, and name every failure rather than the first. Push `` `${target.name}: ${getErrorMessage(err)}` `` at line 158 and join the list here.

packages/api-server-api/src/modules/templates/schemas.ts
```diff
@@ -44,6 +44,7 @@ export const templateSpecSchema = z
     category: templateCategorySchema,
     tags: z.array(z.string()).optional(),
     docsUrl: z.string().optional(),
+    releaseNotesUrl: z.string().optional(),
```
dam-code-guardian, line 43: 🟡 **Warning:** this accepts any string as a URL, and the value reaches an `<a href>` unchecked (`release-notes-link.tsx:29`). Every other URL field in this contract package validates: `connections/schemas.ts:31`, `skills/schemas.ts:180`, `agents/schemas.ts:95`. The spec is parsed from operator-supplied ConfigMap data (`templates-repository.ts:57`), where a typo currently fails silently at click time instead of at parse time. Reported by typescript-engineering.

**Fix:** validate every URL field in the contract with `z.string().url()`, since the contract is the validation layer for catalog input. Apply it to `releaseNotesUrl`, and to the sibling `docsUrl` on line 46 that set the weaker precedent.

```suggestion
    releaseNotesUrl: z.string().url().optional(),
```

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `72a22e0`

## PR #3280: feat(ui): make the "Update available" pill the way to update a sandbox
**Author:** kapetr | **Branch:** feat/3137-update-pill → main | **Changes:** +517 −103 (23 files)

### Summary
Re-review at `72a22e0` after four commits. The 🔴 tap workaround was reverted with a correct mechanism refutation, and the per-harness release-notes table moved out of the client into the template contract (`releaseNotesUrl` through helm → zod → view → UI). The batch confirmation now names the images it moves to. Six of the ten prior findings are genuinely fixed; the contract migration lost one variant the deleted test pinned.

### Changes since last review
Previous HEAD: `5d20e86` (2026-08-12T12:28:19Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** the tap workaround — verified refuted, not patched. `preventDefault()` from any React `onTouchStart` is ignored (react-dom 19.2.7 registers `touchstart` `{passive: true}` at `react-dom-client.development.js:19251`), and Radix calls it inside a React prop (`react-hover-card@1.1.23/dist/index.mjs:95`). A bare trigger is tappable; `HoverCardTappableTrigger` correctly deleted (`hover-card.tsx`)
- ✅ **Fixed:** the empty flex item — the `stopPropagation` wrapper moved inside the component, below `if (!update) return null` (`update-available-action.tsx:34`)
- ✅ **Fixed:** the misplaced doc comment — `whatsNewLink` is gone; `batchApplyNote` carries its own block (`use-update-sandbox.tsx:60`)
- ✅ **Fixed:** the hardcoded client table — `releaseNotesUrl` now crosses the contract, plumbed end to end with no gap (`schemas.ts:47`, `router.ts:16`, `agent-templates.yaml:51`)
- ✅ **Fixed:** the competing identity key — lookups key on `templateId`, the same key `harness-card.tsx` uses (`use-release-notes-url.ts:16`)
- ✅ **Fixed:** @PetrBulanek's batch-consent finding — `byTargetImage` names each target image with its count, and the architecture page now states the per-path contract accurately (`use-update-sandbox.tsx:51`)
- 🔁 **Still present:** `Closes #3137` closes an issue whose Open Question is assigned to @jamiejabbouribm and unanswered (`update-available-action.tsx:37`) — reported by issue-fit
- 🔁 **Still present:** the "Update all" banner is a second entry point beyond #3137's stated scope (`list-view.tsx:108`) — reported by issue-fit
- ✅ **Fixed (declined, verified):** update-vs-upgrade naming — `agents.update` genuinely exists at `router.ts:104` and means editing settings, so the wire name is taken. Resolved, not carried
- 🆕 4 new findings below

### Findings
- 🟡 **Warning:** the contract migration drops the harness variants the deleted test pinned (`deploy/helm/platform/values.yaml:1224`)
- 🟡 **Warning:** "Update all" is gated only on `updatingAll`, so a row update in flight leaves it live (`packages/ui/src/modules/agents/views/list-view.tsx:126`) — reported by react-ui-engineering
- 🟡 **Warning:** the bulk failure toast loses the sandbox name exactly when the server reports a reason (`packages/ui/src/modules/agents/hooks/use-update-sandbox.tsx:173`) — reported by react-ui-engineering
- 🟡 **Warning:** `releaseNotesUrl` is an unvalidated string reaching an `<a href>` (`packages/api-server-api/src/modules/templates/schemas.ts:47`) — reported by typescript-engineering

✅ No findings. Both prior findings at `docs/architecture/agent-lifecycle.md:76` are resolved — the page now states the per-path confirmation contract and describes the capability rather than control placement.

### Issue Fit
🔁 2 findings from the previous review still present (see review at `5d20e86`) — the unanswered Open Question under a closing keyword, and the out-of-scope second entry point. Both were declined by the author on 2026-08-13; the declines are scope calls for the issue owner, and are recorded so later rounds do not re-raise them.

- 🟢 **Suggestion:** `releaseNotesUrl`'s three chart defaults answer the author's own 2026-08-12 question to the issue owner, which is still unanswered — worth confirming the URL list before it ships as operator-facing chart data (`deploy/helm/platform/values.yaml:1224`).

One blocking finding is listed above: the unvalidated URL field (`schemas.ts:47`).

- 🟢 **Suggestion:** `TemplateView` is hand-maintained in `packages/ui/src/types.ts:126`, so this PR added one field twice. `SessionView` already re-exports from the contract (`types.ts:114`) — the same move here would remove the duplication class.

Two blocking findings are listed above: the ungated banner (`list-view.tsx:126`) and the identity-losing failure toast (`use-update-sandbox.tsx:173`).

- 🟢 **Suggestion:** the hover card is the only carrier of "Update available" and the target image, and Radix stamps `tabindex="-1"` on its content (`react-hover-card@1.1.23:175`), so the trigger announces only "Update" where the header previously had an "Update available" badge. An `aria-label` naming the update and its target image would restore that (`update-available-action.tsx:37`).

### Verdict
REQUEST_CHANGES — the tap 🔴 is genuinely resolved and the contract migration is the right cut, but it dropped `claude-code-vm`'s release-notes link, and three further warnings are open.

---

PetrBulanek (approved):

packages/ui/src/modules/agents/hooks/use-agent-rows.ts
```diff
@@ -86,6 +88,8 @@ export function useAgentRows() {
     }),
     deletePending:
       deleteAgent.isPending && deleteAgent.variables?.id === agent.id,
+    updatePending: update.updatingId === agent.id || update.updatingAll,
```
dam-code-guardian: 🟡 **Warning:** `updatingId` holds one sandbox id, but nothing disables the *other* rows' Update buttons during a single-row update. The delta fix disabled only the banner (`list-view.tsx:126`), so clicking Update on row B while row A is in flight rebinds the shared `upgrade` observer to B — `mutate()` calls `removeObserver` and builds a new mutation, verified in query-core at the `@tanstack/react-query@5.99.0` pin. Row A then reads "Update" and is clickable again mid-flight, inviting a duplicate upgrade. The server's compare-and-swap turns the duplicate into a `CONFLICT` toast rather than a wrong image, so this costs a false failure, not a bad update. Reported by react-ui-engineering.

**Fix:** a single scalar in-flight id cannot represent concurrent updates — gate every row on any update in flight, or track in-flight ids as a set each row tests its own id against. Apply the same rule everywhere `updatingId` gates a control, including the header (`sandbox-home-header.tsx:83`).

```suggestion
    updatePending: update.updatingId !== null || update.updatingAll,
```

packages/api-server-api/src/modules/templates/schemas.ts
```diff
@@ -44,6 +44,7 @@ export const templateSpecSchema = z
     category: templateCategorySchema,
     tags: z.array(z.string()).optional(),
     docsUrl: z.string().optional(),
+    releaseNotesUrl: z.string().url().optional(),
```
dam-code-guardian: 🟡 **Warning:** `z.string().url()` accepts any scheme `new URL()` can parse, so `javascript:`, `data:` and `vbscript:` all pass, and the value still reaches an `<a href>` unfiltered (`release-notes-link.tsx:25`). Read at the pin: zod 4.4.3's bare `.url()` sets no `protocol` or `hostname` constraint (`core/api.ts:_url`, `core/schemas.ts:$ZodURL`) — only `z.httpUrl()` adds `protocol: httpProtocol` and `hostname: domain`. The delta fix therefore narrows the input but does not close the class it targeted. The spec is operator-supplied ConfigMap data, so this is a low-privilege sink, not an open XSS — but the constraint the field needs is the scheme, not parseability. Reported by typescript-engineering.

**Fix:** validate the scheme, not just parseability — use `z.httpUrl()` for every schema field whose value reaches an `href` or any other URL sink.

🟡 **Warning (second, same line):** this check also makes one cosmetic field able to delete a whole harness from the catalogue. A scheme-less value such as `example.com/notes` throws inside `new URL()`, which fails the enclosing `templateSpecSchema.parse()`; the repository's `catch` then skips that template file entirely (`templates-repository.ts:57`), so the template vanishes from the catalogue and every sandbox built from it loses its behind-the-template check (`agents-service.ts:575` returns `undefined` with no spec). An operator's typo in an optional link thus costs the whole template. Note `docsUrl` on the line above stays unvalidated, so it carries neither problem.

**Fix:** optional display-only fields must degrade, never fail their parent spec — append `.catch(undefined)` so an invalid value drops only that field and the template still loads.

```suggestion
    releaseNotesUrl: z.httpUrl().optional().catch(undefined),
```

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `851f8ec`

## PR #3280: feat(ui): make the "Update available" pill the way to update a sandbox
**Author:** kapetr | **Branch:** feat/3137-update-pill → main | **Changes:** +521 −103 (23 files)

### Summary
Complete re-review at `851f8ec` on the label. The delta commit resolves all four findings of the previous round: `claude-code-vm` has its release-notes URL back, the batch toast names every failed sandbox, the "Update all" banner gates on both in-flight signals, and `releaseNotesUrl` gained a zod check. Two residual warnings sit in the new branches those last two fixes introduced.

### Changes since last review
Previous HEAD: `72a22e0` (2026-08-13T08:06:53Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** the dropped harness variant — `claude-code-vm` declares `releaseNotesUrl` again, so all four templates the deleted client matcher covered carry a link (`deploy/helm/platform/values.yaml:1240`)
- ✅ **Fixed:** the identity-losing failure toast — the name is now prefixed unconditionally, ahead of `getErrorMessage`, so no branch drops it (`packages/ui/src/modules/agents/hooks/use-update-sandbox.tsx:161`)
- ✅ **Fixed:** the ungated "Update all" banner — it now gates on `updatingAll || updatingId !== null` (`packages/ui/src/modules/agents/views/list-view.tsx:126`)
- 🔁 **Still present:** `releaseNotesUrl` reaches an `<a href>` without a scheme constraint — narrowed by `.url()`, not closed (`packages/api-server-api/src/modules/templates/schemas.ts:47`) — reported by typescript-engineering
- 🆕 2 new findings below

### Findings
- 🟡 **Warning:** the banner fix leaves the per-row gate on a scalar id, so other rows stay clickable during a single-row update (`packages/ui/src/modules/agents/hooks/use-agent-rows.ts:91`) — reported by react-ui-engineering
- 🟡 **Warning:** the new `.url()` check lets one cosmetic field delete a whole harness from the catalogue (`packages/api-server-api/src/modules/templates/schemas.ts:47`) — reported by typescript-engineering

✅ No findings. `docs/architecture/agent-lifecycle.md` carries its `Last verified` bump (2026-08-11 → 2026-08-12, confirmed against `main` via the contents API), and the page measures 39,985 of the 40,000 cap in the gate's own unit (`scripts/doc-size.mjs` counts `content.length`, not bytes).

### Issue Fit
Linked issue: #3137 (closing keyword). All four deliverables are present at HEAD: the pill is an action that opens the confirmation, the inline notice under Image is deleted, the Image section shows the current image only, and the pill, hover card, and both confirmations use "Update". The delta commit closed the previous round's scope gap. The three author declines from 2026-08-13 are recorded as PR-local overrides and were not re-raised.

- 🟢 **Suggestion:** the two api-server error messages for this action still say "upgrade" (`packages/api-server-api/src/modules/agents/router.ts:158` and `:164`), and both reach the user as toast text — the single path appends the server wording to "Failed to update sandbox", the bulk path through `getErrorMessage`. `TemplateMoved` is the failure `expectedToImage` exists to produce, so the most likely message after confirming an "Update" reads "you reviewed the upgrade". This is the issue's "one consistent word" user story, not the wire name — no procedure-name constraint applies to message copy.

Two blocking findings are listed above with their fixes: the unconstrained URL scheme and the spec-wide parse failure, both at `schemas.ts:47`.

- 🟢 **Suggestion:** `TemplateView` (`packages/ui/src/types.ts:117`) is a hand-maintained copy of the shape `toView` returns (`packages/api-server-api/src/modules/templates/router.ts:7`), and this PR adds `releaseNotesUrl` to both halves by hand. The `select: sortExperimentalLast` in `useTemplates` type-checks the two only loosely: it catches a field the UI requires and the router drops, but an optional field or a router-only addition drifts silently. Deriving the client type from the router's inferred output would make the contract package the single source of truth, per the architecture's "no manual type duplication" rule.

One blocking finding is listed above: the scalar in-flight id (`use-agent-rows.ts:91`).

- 🟢 **Suggestion:** the bulk failure toast carries the only record of which sandboxes failed and why, yet it passes no `ttl`, so `emitToast` applies the 5-second default (`use-update-sandbox.tsx:174`). Twelve named `CONFLICT` reasons reach ~1,250 characters, which needs ~40 words/second to read. Sonner does pause the timer while the pointer is over the stack (`sonner@2.0.7`), so this is recoverable rather than lost — but `ttl: 0` would make the aggregate error sticky, matching the 10–15 s TTLs the skills surface already uses for its aggregate toasts.

### Verdict
REQUEST_CHANGES — every prior finding is genuinely fixed, but two of the fixes opened new gaps: the scheme check is weaker than its sink needs, and a bad optional link now drops its whole template.

---

packages/ui/src/modules/sandboxes/components/sandbox-home-header.tsx
```diff
@@ -73,22 +75,15 @@ export function SandboxHomeHeader({ agent, display }: Props) {
...
-          )}
-        </>
-      }
+      adornment={<StatusBadge state={display.state} />}
       actions={
         <>
+          <UpdateAvailableAction
+            agent={agent}
+            pending={updatingId === agent.id}
+            busy={updatingId !== null || updatingAll}
```
dam-code-guardian, line 83: 🟡 **Warning:** this gates the header on its **own** hook instance, not on the update. `useUpdateSandbox` calls `useUpgradeAgentMutation`, and `useMutation` holds a `MutationObserver` created in `React.useState` per calling component — read at the `@tanstack/react-query@5.99.0` pin (`build/modern/useMutation.js`). So `updatingId`/`updatingAll` here report only mutations this header started. `useAgentRows` calls the hook a second time, giving the list its own observer.

Within one surface the gate holds, and `ListView` and `SandboxHomeView` never mount together (`app.tsx:110-130`), so today the two instances cannot render side by side. The defect is that the mechanism's guarantee is narrower than its own docstring claims — "Any update is in flight, this sandbox's or another's" (`update-available-action.tsx:19`) is true per instance only. A third surface rendering an Update control alongside either existing one inherits an ungated button, and the duplicate is caught only by the server's compare-and-swap, as a `CONFLICT` toast for an update that succeeded.

**Fix:** derive a cross-surface in-flight flag from shared client state, never from a per-component mutation observer. Give the upgrade mutation a stable `mutationKey` and read `useIsMutating({ mutationKey })` inside `useUpdateSandbox`, so every surface that renders an update control observes one value. This also removes the duplicated expression here and at `use-agent-rows.ts:92`.

packages/ui/src/modules/agents/components/update-available-action.tsx
```diff
@@ -0,0 +1,73 @@
...
+  return (
+    // Both the action and its portaled hover card sit inside the row's click
+    // target, so their clicks must not also open the sandbox.
+    <span onClick={(e) => e.stopPropagation()}>
+      <HoverCard>
+        <HoverCardTrigger asChild>
+          <Button
+            variant="ghost"
+            size="sm"
+            disabled={pending || busy}
```
dam-code-guardian, line 38: 🟡 **Warning:** a row disabled only by `busy` shows no reason. The label stays "Update" for every row that is not the one updating (line 53 keys the "Updating…" text on `pending`), so the control reads as available but ignores clicks. `Button` already carries the convention for exactly this: it copies a string `tooltip` into `title` when `disabled` is set, because a disabled button fires no pointer or focus events (`button.tsx:87-93`). This call site passes no `tooltip`, so the reason reaches neither pointer users nor assistive tech — and `disabled:opacity-50` is the only signal.

**Fix:** whenever a control is disabled by another item's in-flight work, name that reason on the control itself. Pass `tooltip={!pending && busy ? "Another update is in progress…" : undefined}`, matching the precedent at `files-panel.tsx:49`.

```suggestion
            disabled={pending || busy}
            tooltip={
              !pending && busy ? "Another update is in progress…" : undefined
            }
```

packages/api-server-api/src/modules/templates/schemas.ts
dam-code-guardian, line 43: 🟡 **Warning (carried, narrowed):** dropping the check fixed the whole-template loss, and that was the right half to fix first — but it also gave back the constraint. A scheme-less value such as `example.com/notes` now passes, and the browser resolves that `href` against the app origin, so "See what changed" navigates inside the platform (`release-notes-link.tsx:25`). Verified at the pins: zod 4.4.3 accepts it, and React 19.2.7's `sanitizeURL` rewrites **only** `javascript:` (`react-dom-client.development.js:3166`, `isJavaScriptProtocol`), so `data:` and a same-origin path both reach the DOM. The value is operator-supplied ConfigMap data, so this is a low-privilege sink, not an open XSS.

**Fix:** validate every operator-supplied URL that reaches an `href` at the schema edge, and make the validator non-fatal for the enclosing spec — `z.httpUrl().optional().catch(undefined)`. At this pin `httpUrl` enforces `^https?$` and requires `://`, and `.catch` degrades a bad value to "no link" rather than failing `templateSpecSchema.parse()` and dropping the template (`templates-repository.ts:57`). That keeps the constraint without reintroducing the defect you just removed. The same rule applies to `docsUrl` on the line above, which reaches an `href` at `steps/workload-card.tsx:27`.

```suggestion
    releaseNotesUrl: z.httpUrl().optional().catch(undefined),
```

dam-code-guardian (dismissed):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `5071b10e`

## PR #3280: feat(ui): make the "Update available" pill the way to update a sandbox
**Author:** kapetr | **Branch:** feat/3137-update-pill → main | **Changes:** +534 −103 (23 files)

### Summary
Complete re-review at `5071b10e` on the label. Both findings of the previous round are addressed: the per-row gate now disables every Update control while any update runs, and the fatal URL check is gone. The busy gate is correct within one surface, but it is derived per hook instance, so it does not cross the list/header boundary. Removing the validator also gave back more than the parse failure it fixed.

### Changes since last review
Previous HEAD: `851f8ec` (2026-08-13T10:20:02Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** the scalar in-flight id — `updatePending` is narrowed to the own row, and a new `updateBusy` disables every other row's Update, header included (`use-agent-rows.ts:91`, `sandbox-home-header.tsx:84`)
- ✅ **Fixed:** one cosmetic field could delete a whole harness from the catalogue — the check is removed, so a bad value no longer throws inside `templateSpecSchema.parse()` (`schemas.ts:47`)
- 🔁 **Still present:** `releaseNotesUrl` reaches an `<a href>` with no scheme constraint (`schemas.ts:47`) — reported by typescript-engineering
- 🆕 2 new findings below

### Findings
- 🟡 **Warning:** `busy` is derived per hook instance, so it does not gate across surfaces (`packages/ui/src/modules/sandboxes/components/sandbox-home-header.tsx:84`) — reported by react-ui-engineering
- 🟡 **Warning:** a row disabled only by `busy` still reads "Update" and names no reason (`packages/ui/src/modules/agents/components/update-available-action.tsx:48`) — reported by react-ui-engineering
- 🟡 **Warning:** `releaseNotesUrl` accepts any string, so a scheme-less value renders as a same-origin link (`packages/api-server-api/src/modules/templates/schemas.ts:47`) — reported by typescript-engineering

✅ No findings. `docs/architecture/agent-lifecycle.md` carries its `Last verified` bump, and the rewritten paragraph tracks `updateOne`/`updateAll`. The page measures 39,985 of the 40,000 cap in the gate's own unit (`doc-size.mjs` counts UTF-16 units, not bytes), and this PR's net delta is −1.

### Issue Fit
All four of #3137's deliverables are present at HEAD. The three author declines are recorded as PR-local overrides and were not re-raised.

- 🟡 **Warning:** the api-server error messages for this flow still say "upgrade", and the toast appends them to "Failed to update sandbox", so one action reads with two verbs (`packages/api-server-api/src/modules/agents/router.ts:158`, `:164`). #3137 asks for one consistent word across the pill and the confirmation. Carried from the previous round as a 🟢; raised now because it is the last open item of the issue's own "one consistent word" story.
  **Fix:** rename every user-visible string of this flow to "update", and keep the old verb only in internal identifiers.

One blocking finding is listed above: the unconstrained URL scheme at `schemas.ts:47`.

- 🟢 **Suggestion:** deleting the validator did fix the round-3 defect. `templateSpecSchema.parse()` threw on a bad URL and the `catch` at `templates-repository.ts:57` dropped the whole template, costing a sandbox its image and mounts for a cosmetic field. The remaining gap is that the fix removed the constraint instead of making it non-fatal.
- 🟢 **Suggestion:** `TemplateView` (`packages/ui/src/types.ts:126`) and `TemplateSpec` (`templates/types.ts:43`) are hand-mirrored from the schema and the router's `toView`, and this PR extended both by hand. An omitted optional field on either side compiles clean at the pin, so a later field added to one mirror reads `undefined` on the other. `z.infer<typeof templateSpecSchema>` would make the drift a compile error. Pre-existing pattern — worth a separate issue.

Two blocking findings are listed above: the per-instance busy gate and the unexplained disabled state.

- 🟢 **Suggestion:** `updateBusy` threads through `AgentRow`, which never reads it, only to reach `UpdateAvailableAction`. The flag is global rather than per-row, so a shared value on `useUpdateSandbox` would remove one prop from an already wide surface and delete the duplicated expression at `use-agent-rows.ts:92` and `sandbox-home-header.tsx:84`. The same change resolves the first finding above.
- 🟢 **Suggestion:** `FLOATING_PANEL` and `FloatingPanelTail` hold the popover and hover-card surface in one place, and the deleted `TemplateUpdateNotice` leaves no stale imports. Both are real de-duplication wins.

### Verdict
REQUEST_CHANGES — both prior findings are genuinely fixed, but the new busy gate stops at one hook instance and gives no reason on the controls it disables, and the URL constraint was removed rather than made non-fatal.

---

PR #2814, opened 2026-07-16
feat(ui): redesign the skills surface (grouped by source, inline management)

PetrBulanek:
Closes #944.

Redesigns how skills are presented and managed on a sandbox's home page. Skills are now **grouped by where they come from** rather than shown as one flat list, and every management action — install, add a source, re-scan, publish, update — happens inline on the same surface.

## What this does for the user

- **Two clear groups.** *Created in this sandbox* (standalone skills authored in place) and *Sourced from GitHub* (each connected repo as its own card), instead of an undifferentiated list.
- **Install/uninstall is immediate.** A per-skill toggle installs or removes on click; enabled skills sort to the top and the card shows `N of M on`. Cards with a mix collapse to just the enabled skills with an *Expand all* / *Hide available* control, so a large source stays scannable.
- **Manage sources in place.** *+ Add source* connects a GitHub repo (public or private via the GitHub connection); each source's kebab offers *Re-scan*, *View repo*, and *Remove*. Public repos scan straight from the API server; private repos scan through the pod.
- **Publish a local skill upstream.** Standalone skills can be published as a pull request into a connected source; the row then shows an *In review · \<source\>* pill linking the PR.
- **Read a skill without leaving.** Clicking a skill name opens its `SKILL.md` rendered in-product (frontmatter + markdown + code highlighting), with a *View on GitHub* link.
- **Drift is visible.** When an installed skill's upstream has changed, the row flags an *Update* pill (re-installs at the latest version) plus a compare link.
- **Honest read-only + lifecycle states.** The whole surface is read-only and muted while the sandbox is stopped/starting ("Agent is starting…"); loading shows shaped skeletons rather than a "No skills" flash. Pause/Stop were also added to the sandbox detail actions menu with an optimistic *Hibernating* pill.

## Shape of the change

Implemented as 6 reviewed slices (grouped surface → sort/collapse → source admin → publish → render modal → drift), plus a folded-in Pause/Stop port + optimistic pill, then one commit of merge-gate review cleanups. The old `sessions/components/skills-panel.tsx` is superseded by `sandboxes/components/skills/` + `hooks/use-skills-surface.ts`.

## Testing

- `mise run test` and `mise run check` both green.
- Correctness review of the full branch diff: no correctness bugs; the review's cleanup findings are folded into the final commit.
- A full browser end-to-end pass on a fresh sandbox (+ a manual test guide) is tracked as immediate follow-up.

## Known follow-ups (need backend, out of scope here)

[...]

PR #3282, opened 2026-08-12
feat(ui): explain the model in chat instead of navigating away

### LINKED ISSUE #3057: UI - Model indicator in chat sends you out of the conversation
## Context

The chat composer shows the current model (for example "opus") with a gear icon next to it. It reads like an inline control, so people click it expecting to do something right there. Instead it navigates away to sandbox configuration.

## Problem

Clicking the model indicator while you are mid conversation is jarring. You are in the context of the chat, the affordance looks local, and the click bounces you out to a different page entirely. Nothing in the chip signals that it is a navigation away rather than something that acts in place.

The result is that people lose their place in the conversation and have to find their way back, having expected a small local interaction.

## Goal

Clicking the model indicator in chat should feel predictable. Either it does something in the context you are already in, or it makes clear up front that it takes you elsewhere.

## User Stories

- As someone in a chat session, I want the model indicator to behave the way it looks so that clicking it does not unexpectedly pull me out of my conversation.
- As someone in a chat session, I want to stay oriented when I interact with the composer so that I do not lose my place mid conversation.

## Open Questions

- What should clicking it actually do? Left open deliberately, the point here is the jarring navigation, not a specific fix.

kapetr:
Clicking the model under the composer no longer throws you out of the conversation. It opens a popover naming the model, with a link to configuration so leaving is a choice. The settings icon is gone.

On a knowledge base the popover states the model and stops there — that config page has no model settings, so there is nowhere to send you.

Closes #3057

## Also needed to finish it

The popover sends people to Sandbox Setup, so that page had to be worth arriving at:

- **Model settings wait for "Submit changes"** like every other field there, instead of saving the moment you touch a picker. The bar stays put on Setup, disabled until something is dirty.
- **Model settings render placeholders while loading** instead of nothing, so the section no longer appears out of empty space seconds after the rest of the page.
- **The docs button clears the submit bar.** It sat on top of Submit changes below ~1149px.
- **The sandbox row fits a phone.** The name, its pills, the state and the menu shared one line; the pills now drop below the name on a narrow screen, and the name ellipsises instead of pushing them out.

Deferring the pickers made the panel's apply-on-change path dead code, and review caught that it had *already* been unreachable since `48fe22187` removed the chat mount in July. It's deleted here along with the settle poll, the variant branching, `useHarnessConfigSettled` and `config-section.tsx` — so the panel now has one behaviour, and this half of the PR is net negative.


packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-save.ts
```diff
@@ -155,6 +160,9 @@ export function useSandboxSettingsSave({
           connectionIds: savedAppIds,
         });
       }
+      if (harnessDraft.dirty) {
+        await applyHarnessConfig.mutateAsync(harnessDraft.buildInput(agentId));
+      }
```
dam-code-guardian, line 165: 🟡 **Warning:** The staged apply never re-reads the harness config — it calls `mutateAsync` without invalidating `harnessConfigCurrentKey` or polling `settled`, unlike the apply-on-change path (`model-settings-panel.tsx:82`). The page keeps showing the requested values, so an apply the pod never performs looks exactly like success.

**Fix:** Every path that applies harness config must trigger a re-read of the live config after the apply — invalidate `harnessConfigCurrentKey(agentId)` inside the shared `useApplyHarnessConfig` hook so no caller can skip it.

packages/ui/src/modules/sandboxes/hooks/use-harness-config-draft.ts
```diff
@@ -0,0 +1,72 @@
...
+      ...(unset.length > 0 ? { unset } : {}),
+    };
+  };
+
+  return {
+    set,
+    valueOf,
+    buildInput,
+    dirty: changed.length > 0,
+    commit: () => setSubmitted((prev) => ({ ...prev, ...edits })),
```
dam-code-guardian: 🟡 **Warning:** `commit()` merges the whole `edits` map into `submitted`, including entries the input never carried (`changed` drops edits that equal the settled value), and no entry is ever removed. The buffer pins values it never applied for the page's lifetime, which masks a later change from the chat panel or from the harness's own file.

**Fix:** Pin only what the request carried — commit the `changed` set, clear `edits`, and drop a pinned entry once the live read reports that value. A staging buffer's post-submit state must mirror the request it sent and must expire when server truth agrees.

packages/ui/src/modules/sandboxes/hooks/use-harness-config-draft.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (field === "mode") return live?.mode ?? null;
+    const v = live?.configOptions[field];
+    return typeof v === "string" ? v : null;
+  };
+
+  const set = (field: string, value: string | null) =>
+    setEdits((prev) => ({ ...prev, [field]: value }));
+
+  // Compared, not counted, so picking the original value back isn't a change.
+  const changed = Object.entries(edits).filter(
```
dam-code-guardian, line 50: 🟡 **Warning:** `changed` diffs the edits against `live`, which resolves after the first render. While the pod read is in flight or has failed the pickers show "Not set", so selecting the displayed value stages a no-op that turns into a real change when the baseline lands — and Submit then unsets a value the user never touched.

**Fix:** Never diff staged edits against a baseline that can still change — capture the baseline when the values first resolve and diff against that snapshot, or keep the pickers read-only until it resolves.

packages/ui/src/modules/sessions/components/model-settings-panel.tsx
```diff
@@ -103,16 +111,24 @@ export function ModelSettingsPanel({
...
+    if (draft) return draft.valueOf(field);
     if (field === "model") return current?.model ?? null;
     if (field === "mode") return current?.mode ?? null;
     const v = current?.configOptions[field];
     return typeof v === "string" ? v : null;
   };
 
+  // The displayed model, not the settled one, or one Submit could send a model
+  // with a mode its own constraints exclude.
+  const shownModel = valueOf("model");
```
dam-code-guardian, line 63: 🟡 **Warning:** In `draft` mode a staged value that the newly-picked model's constraints exclude stays selected and is submitted: `constraints` is recomputed from the drafted model, the excluded value is filtered out of `choices` and then pushed back in as the current selection labelled "Not available for the current model" (`:175`), so it remains in the draft's changed set and Submit sends `model` plus the mode/effort that model forbids — the exact combination the comment above says the change prevents. Chat is unaffected because each pick applies on its own.

**Fix:** When a staged selection changes the constraint set, reconcile every dependent group in the same edit — drop or reset any drafted value the new model disallows rather than re-displaying it as selected, so a staged config can never be submitted in a state the picker itself calls invalid.

packages/ui/src/modules/sandboxes/components/sandbox-model-settings.tsx
```diff
@@ -14,11 +15,21 @@ import { useOperableState, WakeToEditButton } from "./sandbox-wake-to-edit.js";
...
+}) {
   const { operable, comingUp } = useOperableState(agentId);
-  const { data: status } = useHarnessConfigStatus(agentId);
+  const { data: status, isPending: statusPending } =
+    useHarnessConfigStatus(agentId);
   const { origin, hasRun } = useResolvedHarnessConfig(agentId);
   const hasCatalog = !!status?.catalog && status.catalog.options.length > 0;
+  // Loading, or supported with the catalog still to arrive — the same wait.
+  const awaitingCatalog =
+    statusPending || (status?.supported === true && !hasCatalog);
```
dam-code-guardian: 🟡 **Warning:** The skeleton gate can never terminate for an operable agent whose catalog never arrives: `supported` is optimistically `true` when capabilities are unknown (`harnessConfigSupported`), and the status poll only stops once a catalog lands, so `awaitingCatalog` stays true indefinitely and Model settings renders a shimmering placeholder forever — where before the panel returned `null` and the section was simply absent. It also carries no `aria-busy`/status text, so assistive tech gets a permanently empty section.

**Fix:** Bound every indefinite loading state — after the catalog wait resolves or exceeds a finite budget, fall through to the real no-catalog rendering instead of leaving a skeleton mounted, and give any skeleton that can outlive one paint an accessible busy state.

packages/ui/src/modules/sandboxes/components/sandbox-model-settings.tsx
```diff
@@ -44,10 +55,13 @@ export function SandboxModelSettings({ agentId }: { agentId: string }) {
     );
   }
 
+  if (awaitingCatalog) return <ModelSettingsSkeleton />;
```
dam-code-guardian: 🟡 **Warning:** This loading gate is ordered after the two `!operable` fallbacks, which read a still-pending snapshot query as "never run" (`hasRun` defaults false), so a stopped sandbox that has run is told "This sandbox hasn't run yet — … Start it once and this page fills in" until the snapshot lands. The gate cannot prevent the wrong-state flash it was added for.

**Fix:** Resolve loading before content: compute the pending state of every query a branch's condition depends on and render the placeholder first, so no branch ever asserts a state derived from data that hasn't arrived.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `02197c6`

## PR #3282: feat(ui): explain the model in chat instead of navigating away
**Author:** kapetr | **Branch:** feat/3057-model-popover → main | **Changes:** +278 −24 (12 files)

### Summary
The chat model chip stops navigating away: a popover names the model and offers an explicit link to configuration, and the settings icon is gone. The PR also stages Sandbox Setup's model settings behind Submit changes, lifts the docs button over the sticky bar, and adds a loading skeleton.

### Findings
- 🟡 **Warning:** Staged apply never re-reads the harness config (`packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-save.ts:163`)
- 🟡 **Warning:** `commit()` pins values that were never sent, and never expires them (`packages/ui/src/modules/sandboxes/hooks/use-harness-config-draft.ts:70`)
- 🟡 **Warning:** `changed` diffs against a baseline that resolves after first render (`packages/ui/src/modules/sandboxes/hooks/use-harness-config-draft.ts:35`)
- 🟡 **Warning:** A drafted value the new model forbids stays selected and is submitted (`packages/ui/src/modules/sessions/components/model-settings-panel.tsx:123`)
- 🟡 **Warning:** The skeleton gate can never terminate when a catalog never arrives (`packages/ui/src/modules/sandboxes/components/sandbox-model-settings.tsx:31`)
- 🟡 **Warning:** The skeleton gate sits after branches that assert state from pending queries (`packages/ui/src/modules/sandboxes/components/sandbox-model-settings.tsx:58`)
- 🟢 **Suggestion:** `--bottom-bar-h` is published by every `StickyFooterLayout` but the layout's own bottom padding stays a hard-coded `52px` (`packages/ui/src/modules/sandboxes/components/sticky-footer-layout.tsx:29`). The two must agree; one is now a variable and one is not, so a future change to the bar height silently desyncs mobile padding.
- ✅ **Looks good:** The `data-bottom-bar` root attribute is safe against the wizard/home shells alternating — only one `SandboxTwoColumnShell` mounts at a time, the effect keys on a boolean, and cleanup deletes the attribute.

✅ No findings.

### Issue Fit
The maintainer's scope comment asks for three things: remove the gear icon, open a popover with context, link out to the config page. The diff delivers all three. The extras are declared in the PR body.

- 🟢 **Suggestion:** The staged-save rework of Sandbox Setup model settings is declared, but it is unrelated to #3057. `Closes #3057` binds five files of save-semantics change to a chat-affordance fix. Split that half into its own PR, as the author offers, so #3057 stays revertable on its own.

- 🟡 **Warning:** The staged apply never re-reads the harness config — it calls `mutateAsync` without invalidating `harnessConfigCurrentKey` or polling `settled`, unlike the apply-on-change path (`model-settings-panel.tsx:82`). The page keeps showing the requested values, so an apply the pod never performs looks exactly like success (`packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-save.ts:163`).
  **Fix:** Every path that applies harness config must trigger a re-read of the live config after the apply — invalidate `harnessConfigCurrentKey(agentId)` inside the shared `useApplyHarnessConfig` hook so no caller can skip it.
- 🟡 **Warning:** `commit()` merges the whole `edits` map into `submitted`, including entries the input never carried (`changed` drops edits that equal the settled value), and no entry is ever removed. The buffer pins values it never applied for the page's lifetime, which masks a later change from the chat panel or from the harness's own file (`packages/ui/src/modules/sandboxes/hooks/use-harness-config-draft.ts:70`).
  **Fix:** Pin only what the request carried — commit the `changed` set, clear `edits`, and drop a pinned entry once the live read reports that value. A staging buffer's post-submit state must mirror the request it sent and must expire when server truth agrees.
- 🟡 **Warning:** `changed` diffs the edits against `live`, which resolves after the first render. While the pod read is in flight or has failed the pickers show "Not set", so selecting the displayed value stages a no-op that turns into a real change when the baseline lands, and Submit then unsets a value the user never touched (`packages/ui/src/modules/sandboxes/hooks/use-harness-config-draft.ts:35`).
  **Fix:** Never diff staged edits against a baseline that can still change — capture the baseline when the values first resolve and diff against that snapshot, or keep the pickers read-only until it resolves.

- 🟡 **Warning:** In `draft` mode a staged value that the newly-picked model's constraints exclude stays selected and is submitted: `constraints` is recomputed from the drafted model, the excluded value is filtered out of `choices` and then pushed back in as the current selection labelled "Not available for the current model", so it remains in the draft's changed set and Submit sends `model` plus the mode/effort that model forbids — the exact combination the comment at line 121 says the change prevents. Chat is unaffected because each pick applies on its own. (`packages/ui/src/modules/sessions/components/model-settings-panel.tsx:123`, surfaced at `:175`)
  **Fix:** When a staged selection changes the constraint set, reconcile every dependent group in the same edit — drop or reset any drafted value the new model disallows rather than re-displaying it as selected, so a staged config can never be submitted in a state the picker itself calls invalid.
- 🟡 **Warning:** The skeleton gate can never terminate for an operable agent whose catalog never arrives: `supported` is optimistically `true` when capabilities are unknown (`harnessConfigSupported`), and the status poll only stops once a catalog lands, so `awaitingCatalog` stays true indefinitely and Model settings renders a shimmering placeholder forever — where before the panel returned `null` and the section was simply absent. It also carries no `aria-busy`/status text, so assistive tech gets a permanently empty section. (`packages/ui/src/modules/sandboxes/components/sandbox-model-settings.tsx:31`, gate at `:58`)
  **Fix:** Bound every indefinite loading state — after the catalog wait resolves or exceeds a finite budget, fall through to the real no-catalog rendering instead of leaving a skeleton mounted, and give any skeleton that can outlive one paint an accessible busy state.
- 🟡 **Warning:** The new loading gate is ordered after the two `!operable` fallbacks, which read a still-pending snapshot query as "never run" (`hasRun` defaults false), so a stopped sandbox that has run is told "This sandbox hasn't run yet — … Start it once and this page fills in" until the snapshot lands; the gate cannot prevent the wrong-state flash it was added for. (`packages/ui/src/modules/sandboxes/components/sandbox-model-settings.tsx:58`, branch at `:37`)
  **Fix:** Resolve loading before content: compute the pending state of every query a branch's condition depends on and render the placeholder first, so no branch ever asserts a state derived from data that hasn't arrived.

### Verdict
COMMENT — the chat half matches the amended issue scope and is clean; the staged-save half carries six blocking 🟡s around baseline timing, pinned state, and unbounded loading.

---

packages/ui/src/modules/sandboxes/components/sticky-footer-layout.tsx
```diff
@@ -13,14 +13,26 @@ export function StickyFooterLayout({
...
+
   return (
     <div className="flex h-full flex-col pb-[calc(52px_+_env(safe-area-inset-bottom))] md:pb-0">
       <div className="min-h-0 flex-1 overflow-y-auto">{children}</div>
       {footer && (
         <div className="border-t border-border bg-background">
           <div
             className={cn(
-              "mx-auto flex h-[52px] w-full items-center justify-end gap-3 px-4 md:h-[70px] md:px-8",
+              "mx-auto flex h-[var(--bottom-bar-h)] w-full items-center justify-end gap-3 px-4 md:px-8",
```
dam-code-guardian, line 35: 🟢 **Suggestion:** The bar's height is now `--bottom-bar-h`, but this layout's own bottom padding (line 29) stays a hard-coded `52px`. The two must agree for mobile padding to clear the bar, so a later change to the variable silently desyncs the padding. Consider `pb-[calc(var(--bottom-bar-h)_+_env(safe-area-inset-bottom))]` so one value drives both.

packages/ui/src/modules/sessions/components/model-settings-panel.tsx
```diff
@@ -45,16 +45,24 @@ type ModelSettingsVariant = "chat" | "page";
 // renders the pickers read-only (agent asleep) and `headerAction` fills the
 // page header's right slot (e.g. "Start agent to edit"). Chat passes none of
 // these, so its appearance and behavior are unchanged.
+//
+// `draft` hands the values and edits to a caller with its own Submit (the
+// settings page). Without it the panel applies on change, as chat needs.
```
PetrBulanek: This comment (and the PR description) says chat's copy of the panel still applies on
change. There is no chat copy. `ModelSettingsPanel` has one caller —
`SandboxModelSettings` — and it always passes `variant="page"` and `draft`. The chat
mount went away in #2814, before this branch.

That makes the whole non-draft half unreachable once this lands: the `change()` body
below the `if (draft)` early return, `apply`, `saving`, `awaitingSettle`, the
`useHarnessConfigSettled` poll, the settle-then-invalidate effect, the 15s give-up
timeout, `SavingHint`, and the `variant="chat"` return branch with `OptionGroup`'s chat
styling. Anyone changing this panel later has to reason about two behaviors, and the
comment tells them both are live.

If `draft` were required instead of optional, all of that could go with it and the diff
would get smaller rather than bigger. If you would rather keep the chat path for a
planned re-mount, that is a fair call — but the comment should say it is currently
unreachable, so the next reader does not trust it.

PetrBulanek (approved):

packages/ui/src/modules/sessions/components/model-settings-panel.tsx
kapetr: You're right, let's get rid of it and simplify the component.

Issue #3283, opened 2026-08-12
Choose a sandbox's skills before it first starts


PetrBulanek:
## Context

Follow-up from #3208. The design ([prototype](https://github.com/user-attachments/files/30827535/issue-3022-prototype.html)) for the Skills rework includes a panel for a sandbox that has never run: you add skill sets or a source, and they are applied the first time the sandbox starts. That part was not built in #3208 because the feature behind it does not exist.

## Problem

You create a sandbox already knowing which skills you want it to have, and you can't give it any. The only way to configure skills is to start the sandbox first, so "set this up" becomes "boot it, wait, then set it up" — every time, for every sandbox.

## Goal

Add skill sets or sources to a sandbox that has never run, and have them applied on its first start.

## Design

The design covers the entry point and the empty state. It does not draw the same panel once a choice has been made — what it shows while a selection waits for first start. That is the only design work outstanding.

jenna-winkler:
dupe of #3285

PR #3289, opened 2026-08-12
fix(skills): stop a failed skill install from reporting success

PetrBulanek:
Installing a skill into a sandbox reported success and did nothing — no files, no entry in the skills list. Retrying changed nothing. Found while smoke-testing the #3208 Skills rework; not caused by it.

## What was going wrong

Installing a skill is a request, not an action: the api-server records what the sandbox *should* have, and the sandbox then fetches the files and reports back. When a fetch failed, the sandbox logged it and carried on as if the whole set had installed. Four things followed:

1. **We reported success we never verified.** The API call returns once the request is recorded, and nothing downstream ever contradicted it.
2. **The failure became permanent.** The sandbox remembers what it installed as a fingerprint of the whole skill set. Once the failed set was marked done, re-installing rebuilt the identical fingerprint and was skipped as already applied. **Retrying could never repair it.**
3. **The catalogue lied too.** A cleanup pass deletes entries whose files are missing — including the entry the install had just written. The sandbox looked untouched.
4. **And it stayed broken.** That cleanup removed entries without telling the sandbox, which kept a fingerprint describing a set that no longer existed. Re-adding the skill reproduced it exactly — skipped again.

It looked specific to one skill, but wasn't. The sandbox re-installs every skill on each sync, and a failure for one already on disk leaves no trace. Only a failure for the skill being **added** leaves nothing behind, so the freeze pinned the symptom to that one. The trigger was a single unrecoverable fetch failure that nothing recorded; it now lands in the sandbox's failure list and raises `ContributionApplyFailed`.

## The change

| | |
|---|---|
| `skill-install` driver | Collect per-skill failures and fail the sync after the loop, so skills that did land stay tracked and the removal sweep still runs. |
| `applyState` | "Already done" now means *strictly* older. At the same version the fingerprint decides, so a set re-sent with different contents is applied. |
| `state` reconcile | Clean up only once the sandbox has **applied** its version, not merely answered — a failed install still counts as an answer. |
| `state` reconcile | Tell the sandbox when a cleanup removes entries, or it keeps a fingerprint for a set that no longer exists. |
| harness-config poll | Route the settled poll through the cursor-only read it was meant to use — one less query per 800ms tick. |
| `state` read | Make the post-cleanup re-delivery best-effort; the Skills page shouldn't fail for a side effect the 60s sweep repeats. |

## Validation

Local dev cluster, against the sandbox that produced the original report, both images built from this branch.

Pre-fix, reproduced: the API returned OK with 17 installed, the sandbox logged `hash unchanged; skipping dispatch`, the disk never changed, and one Skills page load dropped the entry back to 16.

After the fix:

- **A real failure surfaces.** Installing a skill absent from its source held the applied cursor back and recorded `install failed for 1 skill(s): … (SkillNotFoundInSource)`. The other 17 stayed installed.
- **The cleanup holds off.** The sandbox had answered but not applied; the old rule would have deleted the entry, the new one left it.
- **The ghost cycle self-heals.** Deleting a skill's files out-of-band made one page load remove the entry *and* notify the sandbox. Re-installing then landed the files — pre-fix a guaranteed no-op.

Sandbox left healthy. `api-server:check`, `agent-runtime:check` and both unit suites (920 + 235) pass.

## Notes for review

- **No new tests.** This is a cross-process interaction (driver → dispatcher → outbox → cleanup) the existing unit seams don't reach. The driver's throw is the one piece that would unit-test cleanly, since its installer is injected — say if you want that seam.
- **Left out:** an in-request retry for transient fetch errors. The sync already retries at 60s, so it would only make the first attempt succeed rather than the second.
- **Adjacent wart, untouched:** on the `stale` branch `worker-handler` records a fingerprint the sandbox never applied. The strictly-older change makes that branch nearly unreachable; a real fix needs the sandbox to return its own fingerprint, which is a contract change.
- Both architecture pages sit near their character cap, so two lines in the `connections.md` diagram were raised in level to make room.


packages/api-server/src/modules/skills/services/skills-service.ts
```diff
@@ -1196,19 +1196,39 @@ export function createSkillsService(deps: SkillsServiceDeps): SkillsService {
...
       } catch (err) {
         getLogger().warn(
           { err, agentId },
-          "skills state: settled check failed; deferring reconcile",
+          "skills state: applied check failed; deferring reconcile",
         );
       }
-      if (settled) {
-        await deps.agentSkillsRepo.reconcile(agentId, onDisk);
+      if (applied) {
```
dam-code-guardian: 🟡 **Warning:** A permanently-failing install becomes a permanent reap block.

`recordOutcome` holds `lastAppliedVersion` back on every failing settle, and `listRetryable` stops re-dispatching once `applyAttempts` reaches the cap of 8 (`outbox-repo.ts:239`). After that the row is settled, still carries failures, and is over the cap — so nothing re-enqueues it, `isApplied` stays false forever, and this gate never opens again. No ghost row for that agent is ever reaped, including rows unrelated to the failed skill.

The author's own validation used `SkillNotFoundInSource`, which is exactly the permanently-failing case: the retry can never succeed, so the cap is reached in normal operation rather than only under a freak fault. Only a later `bump` (a new install/uninstall) clears `apply_attempts` and re-arms the row.

**Fix:** When a gate blocks a maintenance action on "the pod caught up", give it an escape for the given-up state: treat a row at the attempt cap as reap-eligible, or gate on `applyFailures` naming a kind other than `skill-ref`, so a dead apply cannot freeze reconciliation indefinitely.

packages/api-server/src/modules/skills/services/skills-service.ts
```diff
@@ -1196,19 +1196,39 @@ export function createSkillsService(deps: SkillsServiceDeps): SkillsService {
...
+        const reaped = await deps.agentSkillsRepo.reconcile(agentId, onDisk);
+        // The pod's applied hash still describes the set that held these, so
+        // re-installing one would reproduce that hash and be skipped as
+        // already-applied. Bumping re-delivers the set it actually has.
+        // Best-effort, like the snapshot write below: this is the read the
+        // Skills page cannot do without, and the 60s sweep re-dispatches a
+        // bumped row on its own, so neither call is worth failing it for.
+        if (reaped.length > 0) {
```
dam-code-guardian: 🟡 **Warning:** The re-delivery is best-effort, but nothing repeats it.

The comment says the 60s sweep re-dispatches a bumped row on its own. That holds only if the bump landed. When `bump` itself throws, the rows are already deleted, so the next `getState` finds no ghosts, `reaped` is empty, and no bump is ever re-issued. The sweep cannot recover it either: `listRetryable` selects rows that are unsettled or hold failures under the cap, and a settled clean row whose bump was lost is neither.

The result is the exact state this PR exists to break — the pod keeps an applied hash naming a set that no longer exists, so re-installing the reaped skill reproduces that hash and is skipped as already-applied.

**Fix:** Make a best-effort side effect recoverable before swallowing it: record the pending re-delivery (a durable marker, or an event a saga retries) so a later pass re-issues the bump. Only swallow a failure whose retry some other path is guaranteed to reach.

docs/architecture/skills.md
```diff
@@ -270,7 +270,7 @@ The cache is invalidated on `sources.refresh` and after every successful publish
...
...andbox runs, `state` records the local list it computed, and while the pod is unreachable it serves that recording, dated and marked as a snapshot. Nothing recorded means the sandbox has never run, which is distinct from having no local skills. **And reaping waits for the pod to catch up** — install is declarative, so between the row landing and the apply fetching the files a freshly-installed skill's directory is legitimately absent; the reconcile is gated on the pod having **cleanly applied** the current version, not merely settled it ([runtime channel](connections.md#the-runtime-channel)) — a failure settles the version too, and a skill the pod cannot install fails the whole apply rather than being quietly skipped, so its row waits for the retry instead of being reaped. Reaping mid-apply would not merely lose the install: the files still land, so the skill reappears as a Standalone one the user supposedly authored. A reap re-delivers, too — the pod's applied hash still names the set that held the reaped row. A snapshot never drives reconciliation either — it is not evidence about the current disk. The cost of both is that a genuine out-of-band deletion is noticed one read later.
```
dam-code-guardian: 🟡 **Warning:** "A reap re-delivers, too" is stated as a guarantee, but the final commit made that re-delivery best-effort — the service logs a warning and continues when the bump or the enqueue fails. The page promises more than the code gives.

**Fix:** When code demotes a step to best-effort, the owning architecture page must say so and name the fallback that recovers it. State the guarantee at the strength the code delivers, never above it.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `f247a13`

## PR #3289: fix(skills): stop a failed skill install from reporting success
**Author:** PetrBulanek | **Branch:** fix/skill-install-silent-failure → main | **Changes:** +107 −44 (13 files)

### Summary
Repairs the silent-failure chain that let a skill install report success while doing nothing: the driver now fails the apply on a per-skill install failure, `applyState`'s stale gate narrows to strictly-older so a re-delivered version with a different hash still applies, and the `state` reconcile gates on *cleanly applied* rather than merely settled, re-delivering after a reap. The four causes are each cut at the right layer and the cross-process reasoning holds.

### Findings
- 🟡 **Warning:** A permanently-failing install becomes a permanent reap block (`packages/api-server/src/modules/skills/services/skills-service.ts:1213`).
- 🟡 **Warning:** The reap's re-delivery is best-effort, but nothing repeats it (`packages/api-server/src/modules/skills/services/skills-service.ts:1221`).
- 🟢 **Suggestion:** `contributionsSettled` and `contributionsApplied` each issue their own `getRow` call, so a caller wanting both bits pays two queries for one row. One reader returning both would serve every current call site (`packages/api-server/src/modules/runtime-delivery/compose.ts:158`).
- ✅ **Looks good:** The driver throws *after* `setInstalled` and the removal sweep, so a partial install stays managed and the sweep still runs — the ordering comment is accurate, and the dispatcher converts the throw into a per-kind `DriverFailure` rather than losing the other drivers.

- 🟡 **Warning:** The page states that a reap re-delivers, without qualification. The PR's final commit made that re-delivery best-effort: the service logs a warning and continues when the bump or the enqueue fails. The page promises a guarantee the code no longer gives (`docs/architecture/skills.md:273`)
  **Fix:** When code demotes a step to best-effort, the owning architecture page must say so and name the fallback that recovers it. State the guarantee at the strength the code delivers, never above it.
- 🟡 **Warning:** The runtime-channel page does not distinguish a settled version from a cleanly applied one. It still presents a single ack marker, and the outbox row description omits the settled cursor. This PR makes the distinction a cross-subsystem contract: it exposes an applied-only port, and the skills page cross-links this page for the meaning of "cleanly applied" (`docs/architecture/connections.md:262`)
  **Fix:** The page that owns a concept must define every state another page cross-links it for. When a diff splits one status into two with different consumer meanings, document both states and what each one permits.

### Issue Fit
- 🟢 **Suggestion:** No issue tracks this bug or its deferred items. `#3208` is the only reference, and the body disclaims it ("not caused by it"). There is no closing keyword, so no issue closes wrongly on merge. Two declared exclusions — the in-request retry for transient fetch errors, and the `worker-handler` `stale`-branch fingerprint wart that this change makes nearly unreachable — live only in the PR body and will be lost once it merges. Filing them as follow-up issues matches the precedent this epic already set with `#3283`.

- 🟡 **Warning:** The driver signals per-skill install failures with a thrown `Error`, and discards the typed `SkillsDomainError` it already holds. `failed.push()` keeps only a display string, so the dispatcher can report a kind and a message but no structured cause (`packages/agent-runtime/src/modules/runtime-channel/drivers/skill-install-plugin.ts:96`).
  **Fix:** Return failures as values, never as exceptions — widen the `KindHandler` contract to give back the failure list (or a `Result`), and let each handler pass the typed error through. A layer boundary must not downgrade a typed error to a string.
- 🟡 **Warning:** `getState` is a query, but it now reaps rows and then drives a runtime re-delivery through another module's mutator. The reaction is fire-and-forget and crosses a bounded context (`packages/api-server/src/modules/skills/services/skills-service.ts:1221`).
  **Fix:** Emit a domain event when rows are reaped, and put the re-delivery in a saga that subscribes to it. A fire-and-forget reaction that crosses modules belongs in a saga, not inline in the service that observed it.
- 🟡 **Warning:** The settled/applied rule lives in the composition root, and the two new readers make four copies of the same version comparison (`packages/api-server/src/modules/runtime-delivery/compose.ts:158`).
  **Fix:** Move a business predicate into the module's `domain/` layer as one pure function, and call it from every reader. Keep `compose.ts` restricted to assembling adapters and services.
- 🟡 **Warning:** `ContributionsSettledPort` now carries `isApplied`, so the port name no longer describes what it exposes, and the wiring reads as a contradiction — `runtimeApplied: contributionsSettled` (`packages/api-server/src/modules/agents/services/agents-service.ts:72`, `packages/api-server/src/apps/api-server/trpc/context.ts:206`).
  **Fix:** Rename the port and its wired variable to the concept they actually carry (for example `ContributionsProgressPort`), so interface, file, and variable names stay consistent.

_Two further cross-module import findings were dropped in verification: `skills` already imported `RuntimeMutator` and a port type from sibling modules on `main`, and six other modules do the same, so the pattern is this codebase's baseline rather than something this PR introduces._

### Verdict
COMMENT — the four-layer fix is correct and well-validated, but two gate states leave reconciliation permanently stuck (a given-up apply, a lost re-delivery bump), and both architecture pages overstate what the code now guarantees.

---

packages/api-server/src/modules/skills/services/skills-service.ts
```diff
@@ -1196,19 +1196,39 @@ export function createSkillsService(deps: SkillsServiceDeps): SkillsService {
...
-      let settled = false;
+      // Applied, not merely settled: a driver failure settles the version too,
+      // and reaping then would drop the row of the very install that is still
+      // being retried.
+      // A failed read defers reaping rather than failing the whole state read:
+      // this is the one query the Skills page cannot do without.
+      let applied = false;
       try {
-        settled = await deps.runtimeSettled.isSettled(agentId);
+        applied = await deps.runtimeApplied.isApplied(agentId);
```
kapetr: **The code can delete a good installation.**

The code does two steps. First it asks the sandbox which skill files are on the disk. Then it checks if the sandbox applied its last instructions.

The check comes after the disk list. If an installation completes between the two steps, the list does not show the new skill. But the check says the sandbox is complete. Then the code decides the files are absent. It deletes the database entry.

That part is not new. The new part is this: the deletion also tells the sandbox to synchronize again. The skill is no longer in the list of skills the sandbox must have. Thus the synchronization deletes the files from the disk. A correct installation is lost.

Before this change, the same error deleted only the database row. The files stayed on the disk. The skill showed again as a skill made in the sandbox.

packages/api-server/src/modules/skills/services/skills-service.ts
```diff
@@ -1196,19 +1196,39 @@ export function createSkillsService(deps: SkillsServiceDeps): SkillsService {
...
+        const reaped = await deps.agentSkillsRepo.reconcile(agentId, onDisk);
+        // The pod's applied hash still describes the set that held these, so
+        // re-installing one would reproduce that hash and be skipped as
+        // already-applied. Bumping re-delivers the set it actually has.
+        // Best-effort, like the snapshot write below: this is the read the
+        // Skills page cannot do without, and the 60s sweep re-dispatches a
+        // bumped row on its own, so neither call is worth failing it for.
+        if (reaped.length > 0) {
+          try {
+            await deps.runtimeMutator.bump(agentId, []);
```
kapetr: **This can cause the error that the PR corrects.**

The cleanup does two steps. It increases a version number. Then it sends a message to the sandbox. One `try`/`catch` holds both steps and ignores all errors. The reason given is that the 60-second sweep does the work again.

This is correct for the second step. A higher version number shows the record as not complete. Thus the sweep finds the record and tries again.

This is not correct for the first step. If the increase of the version number fails, the code already deleted the old entries. But the record shows as complete. The sweep examines only records that are not complete. Thus no process looks at the record again. The sandbox keeps a fingerprint for a set of skills that is not present.

At the next installation of that skill, the fingerprint is the same. The sandbox decides it did the work before. It skips the work. It reports success. This is the failure that the PR corrects.

Thus the note about the 60-second sweep is correct for one call and not correct for the other call.

packages/api-server/src/modules/skills/services/skills-service.ts
```diff
@@ -1184,35 +1232,50 @@ export function createSkillsService(deps: SkillsServiceDeps): SkillsService {
...
+      if (gateBefore !== null && ghosts.length > 0) {
+        // Close the bracket: the row has to be unchanged across the evidence.
+        // An install landing mid-read has to bump, so a moved version means the
+        // disk list and these rows describe different moments. An unchanged one
+        // also means no dispatch is in flight — the sweep and `hello` only
+        // enqueue a row the agent is behind on, and every other enqueue bumps —
+        // so the pod cannot be part-way through writing a directory we are
+        // about to call missing.
+        const gateAfter = await reapGate(deps, agentId);
+        const stable =
```
dam-code-guardian, line 1019: 🔴 **Critical:** The bracket concludes "one moment" from an unchanged `version`, but no writer of `agent_skills` moves that version in the same commit as the row.

`install` (`:820`), `uninstall` (`:851`) and the batch path (`:558`) each write rows and *then* call `bump` as separate awaits, and `runtime-mutator.ts:23` bumps outside any caller transaction. A `getState` interleaving between those two statements sees the new row, an unchanged `version`, and `applied === true` on both reads — so `stable` holds, `bumpForReap` succeeds, and the freshly installed skill is reaped. The reap re-delivers, so the pod then removes files it just fetched. That is the deleted-files failure this commit exists to prevent, and it is reachable on the ordinary install path rather than under a freak fault.

The comment above the batch path already states the premise the gate needs — "a failure part-way leaves rows the pod hasn't been told about" — which is exactly the window: the rows are visible before the bump.

**Fix:** A gate that infers stability from a cursor requires every writer of the guarded state to move that cursor in the same commit. Make the row write and the version bump atomic — one transaction per mutation with the bump inside it — so "version unchanged" really implies "no row changed".

packages/api-server/src/modules/skills/services/skills-service.ts
```diff
@@ -1184,35 +1232,50 @@ export function createSkillsService(deps: SkillsServiceDeps): SkillsService {
...
-          { err, agentId },
-          "skills state: settled check failed; deferring reconcile",
-        );
-      }
-      if (settled) {
-        await deps.agentSkillsRepo.reconcile(agentId, onDisk);
+      const tracked = await deps.agentSkillsRepo.listSkills(agentId);
+      let installed = tracked;
+      const ghosts =
+        gateBefore?.applied === true
```
dam-code-guardian, line 1014: 🟡 **Warning:** A permanently-failing install blocks every later reap for this agent.

`recordOutcome` holds `lastAppliedVersion` back on every failing settle, and `listRetryable` stops re-dispatching once `applyAttempts` reaches the cap of 8 (`outbox-repo.ts:239`). The row is then settled, still carries failures, and is over the cap — so nothing re-enqueues it, `applied` stays false forever, and this gate never opens again. No ghost row for that agent is reaped, including rows unrelated to the failed skill. Only a later `bump` clears `apply_attempts` and re-arms the row.

The author's own validation used `SkillNotFoundInSource`, which is the permanently-failing case: the retry can never succeed, so the cap is reached in normal operation.

**Fix:** When a gate blocks a maintenance action on "the pod caught up", give it an escape for the given-up state — treat a row at the attempt cap as reap-eligible, or gate on `applyFailures` naming a kind other than `skill-ref` — so a dead apply cannot freeze reconciliation indefinitely.

packages/api-server/src/modules/skills/services/skills-service.ts
```diff
@@ -1184,35 +1232,50 @@ export function createSkillsService(deps: SkillsServiceDeps): SkillsService {
...
+        const gateAfter = await reapGate(deps, agentId);
+        const stable =
+          gateAfter?.applied === true &&
+          gateAfter.version === gateBefore.version;
+        // The bump comes first: it is the durable half of the re-delivery, so
+        // the sweep re-dispatches even if the enqueue is lost. Bumping after
+        // the delete would strand the pod on an applied hash for a set that no
+        // longer exists — re-installing a reaped skill would reproduce that
+        // hash and be skipped, which is the failure this path exists to break.
+        if (stable && (await bumpForReap(deps, agentId))) {
```
dam-code-guardian: 🟡 **Warning:** The read path now writes into another bounded context.

`getState` was already a query that reconciled rows; it now also calls `runtimeMutator.bump` and `enqueueAfterCommit` (helper at `:303`), so a Skills-page read bumps the runtime-delivery outbox cursor and enqueues a delivery job. `skills-service.ts:46` reaches for `RuntimeMutator` from `../../runtime-delivery/index.js` — a direct service-port import between modules, rather than the event-and-guard boundary the module docs allow.

**Fix:** Keep queries side-effect-free. Have `getState` compute and return the ghost set, and move the delete-and-re-deliver into a command, or a saga subscribed to the drift, so no read path bumps another module's cursor.

packages/api-server/src/modules/skills/services/skills-service.ts
```diff
@@ -1184,35 +1232,50 @@ export function createSkillsService(deps: SkillsServiceDeps): SkillsService {
...
+          await deps.agentSkillsRepo.reconcile(agentId, onDisk);
```
dam-code-guardian: 🟡 **Warning:** The rows the bracket verified are not the rows deleted.

`ghosts` is computed at `:1252` but only its length is used. `reconcile` then re-selects the agent's rows from a fresh read and drops whatever is absent from `onDisk` at delete time (`agent-skills-repository.ts:154`), including a row that appeared after `gateAfter`. So the stability evidence covers one set and the mutation covers another.

**Fix:** A check-then-act guard must hand the act the identities it checked. Pass the verified set into the delete and scope the deletion to exactly those keys.

packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts
```diff
@@ -419,13 +419,34 @@ async function write(
...
+    // Staged under a dot-prefixed sibling, then swapped in with two renames.
+    // `listEntries` skips dotted entries, so neither the copy nor the delete of
+    // the previous tree is ever visible — the published path is absent only
+    // between the two renames. Writing onto `dst` directly left the skill
+    // missing for the length of a recursive copy, long enough for a reader to
+    // take the absence for an out-of-band deletion. Validating the staged tree
+    // rather than the published one follows from the same rule: a rejected
+    // skill never becomes visible, and the previous copy survives the attempt.
+    const staged = path.join(targetRoot, `.${name}.staging`);
+    const previous = path.join(targetRoot, `.${name}.previous`);
```
dam-code-guardian, line 348: 🟡 **Warning:** Two concurrent writes of the same skill name cross-delete through the fixed sidecar pair.

`.<name>.staging` and `.<name>.previous` are derived from the name alone, so a second `write()` for the same skill starts by `rm`-ing the first's in-flight copy, and its `finally` can rename `previous` over the tree the other just published. The old `rm` + `cp` onto `dst` was equally unserialized but self-limiting; the fixed sidecars turn a concurrent publish into a cross-write that can leave the published path holding the *older* tree. `write-local.ts:99` records that the collision guard relies on "the api-server is the sole caller", which the driver's install loop does not honor.

**Fix:** A staged-publish scheme must make its staging path unique per attempt — suffix a pid or random token, as `withTempDir` already does — or serialize writes per skill name. Never let two attempts share a fixed sidecar path they both `rm`.

packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts
```diff
@@ -419,13 +419,34 @@ async function write(
...
+      await fs.cp(srcDir, staged, { recursive: true });
+      await assertNoSymlinks(staged);
+      await fs.rename(dst, previous).catch(ignoreMissing);
+      await fs.rename(staged, dst);
+      published = true;
+    } finally {
+      await fs.rm(staged, { recursive: true, force: true });
+      if (published) {
+        await fs.rm(previous, { recursive: true, force: true });
+      } else {
```
dam-code-guardian, line 364: 🟡 **Warning:** The `finally` cleanup can replace the real error, and can fail an install that already succeeded.

`fs.rm(staged)` and `fs.rename(previous, dst)` run unguarded, so a rejection here overwrites the in-flight error with the cleanup's. On the success path an `EACCES`/`EBUSY` from `rm(previous)` makes an already-published install throw as if it failed. `writeFromDir` is typed `Promise<{contentHash: string}>` with no `Result`, so that throw escapes `runInstall` (`services/install.ts:53`) past the `Result<…, SkillsDomainError>` contract and lands as an untyped driver failure that fails the whole apply.

**Fix:** In a staged-publish `finally`, only the restore may affect the outcome: swallow-and-log the post-publish cleanup of the previous tree, and never let a cleanup rejection overwrite the in-flight error.

packages/agent-runtime/src/modules/runtime-channel/drivers/skill-install-plugin.ts
```diff
@@ -124,6 +126,17 @@ export function createSkillInstallPlugin(deps: {
             nextManaged.join(", ") || "<none>"
           }; removed ${toRemove.length}`,
         );
+
+        // Thrown last, so the skills that did land stay managed and the sweep
+        // still runs. Swallowing this instead would let applyState record the
+        // state as applied: the content-addressed hash then matches forever, so
+        // the same install is skipped on every later delivery and the api-server
+        // reaps its row as a ghost.
+        if (failed.length > 0) {
```
dam-code-guardian, line 125: 🟡 **Warning:** The typed `SkillsDomainError` is flattened into a display string at the driver boundary.

`failed.push(...)` interpolates `result.error.kind` into text (`:96`), and this throw carries only the joined message. `dispatcher.ts:92` reduces it further to `(err as Error).message` inside `DriverFailure { kind: ContributionKind, message: string }`, where `kind` is the *contribution* kind, not the error's. The api-server therefore cannot tell `SkillNotFoundInSource` from `UpstreamUnreachable`, so `recordOutcome`'s retry and give-up logic treats a permanently-failing install exactly like a transient one — and burns all 8 attempts on a fetch that can never succeed.

**Fix:** Carry the discriminated error across the boundary instead of stringifying it — widen `KindHandler` to return its failures, or let `DriverFailure` carry a structured cause — so a per-kind outcome keeps the discriminant the domain already models. A typed error must never narrow to a display string at a layer boundary.

packages/agent-runtime/src/modules/runtime-channel/service.ts
```diff
@@ -46,10 +46,13 @@ export function createRuntimeChannelService(
       `[applyState] incoming v=${input.version} hash=${input.state.hash.slice(0, 8)} local v=${local.lastAppliedVersion} hash=${(local.lastAppliedHash ?? "<none>").slice(0, 8)} contribs={${kindCounts}} events={${eventCounts}}`,
     );
 
-    // Contributions caught up, but events carry their own version — still apply them.
-    if (input.version <= local.lastAppliedVersion) {
+    // Strictly older only. At the same version the hash decides: the server may
+    // re-deliver one version with different contributions (a row reaped without
+    // a bump), and refusing on the version alone would leave that difference
+    // unapplied forever. Events carry their own version — still apply them.
+    if (input.version < local.lastAppliedVersion) {
```
dam-code-guardian, line 45: 🟡 **Warning:** The loosened guard no longer rejects a stale payload that arrives at an equal version.

`worker-handler.ts:34` reads `row.version` and builds the payload later at `:58`, so two concurrent dispatches can carry the same `version` with different contributions. The agent's serialization (`:27`) orders them, but with `<` the older one is no longer refused: if it lands second and its hash differs, it re-dispatches and overwrites the newer applied state that `<=` used to protect.

**Fix:** Make the equal-version path decide by provenance, not by hash inequality alone — bump the version for every payload change so equal version implies equal content, or stamp the payload with a monotonic build id the agent compares. Loosening a staleness guard requires the version to be a true content cursor.

packages/api-server/src/modules/agents/services/agents-service.ts
```diff
@@ -58,19 +58,31 @@ export interface ContributionsStatus {
   preparingWorkspace: boolean;
 }
 
-/** Port: the failed contributions surfaced on an agent (the degraded badge). */
-export interface ContributionsSettledPort {
+/** Outbox-derived delivery cursors, supplied by runtime-delivery. `applied` is
+ *  the stricter one: a driver failure settles the version without applying it,
+ *  so a consumer that acts on "the pod's disk reflects the spec" reads that. */
+export interface ContributionsProgress {
```
dam-code-guardian, line 62: 🟡 **Warning:** `ContributionsProgress` is now declared twice across the seam it protects.

This declaration is what every consumer types against (exported via `agents/index.ts`), while `runtime-delivery/domain/outbox-progress.ts:12` declares its own and `runtime-delivery/compose.ts:152` returns that one. They match structurally today, so any drift — a field added to the domain type, `failures` narrowed — compiles silently at exactly the boundary the port exists to guard.

The rename itself is right: the port now carries `progress()` and the name matches the wiring.

**Fix:** One owner per contract shape. Have the consuming port re-export the producing module's type, or define it once in a shared contract, rather than restating it — a structurally-duplicated interface across a module boundary is an unchecked assumption, not a decoupling.

packages/api-server/src/modules/runtime-delivery/domain/outbox-progress.ts
```diff
@@ -0,0 +1,36 @@
...
+  /** The agent answered *cleanly*: its disk reflects the current version. */
+  applied: boolean;
+  /** The drivers that failed the last settle. */
+  failures: DriverFailure[];
+}
+
+/** No row means nothing was ever asked of the agent, so it is trivially caught
+ *  up. Version 0 is below the first bump's 1, so a row appearing between two
+ *  reads still reads as a change. */
+export function progressOf(row: OutboxCursors | null): ContributionsProgress {
```
dam-code-guardian, line 17: 🟡 **Warning:** The domain function hands back the repo row's own mutable array.

`failures: row.applyFailures` is passed by reference straight through `contributionsStatus` and `contributionsStatusMany` (`compose.ts:148`, `:176`), so a caller can mutate what the repository returned. The docstring calls this "a pure predicate over the row", and the neighbouring `DriverFailure[]` fields are not `readonly` either.

**Fix:** Return domain values that cannot be aliased back into infrastructure state — copy the array, or type it `ReadonlyArray<DriverFailure>`, in every domain assembly function.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `40fa9a8`

## PR #3289: fix(skills): stop a failed skill install from reporting success
**Author:** PetrBulanek | **Branch:** fix/skill-install-silent-failure → main | **Changes:** +267 −114 (16 files)

### Summary
Two new commits answer the previous round at the cause: the reap's re-delivery bump now runs first and a failed bump cancels the reap, the settled/applied predicate moved into one pure `progressOf`, the port was renamed to `ContributionsProgressPort`, and install publishes a skill directory atomically through a staged sibling and two renames. Both prior doc findings and both of kapetr's comments are genuinely resolved, but the new reap bracket infers "nothing moved" from the outbox version, and no writer moves that version in the same commit as the row it guards.

### Changes since last review
Previous HEAD: `f247a13` (2026-08-13T08:22:22Z) — verdict COMMENT

- ✅ **Fixed:** the reap's re-delivery is now durable — the bump precedes the delete and cancels the reap on failure (`packages/api-server/src/modules/skills/services/skills-service.ts:303`)
- ✅ **Fixed:** the settled/applied predicate is one pure function in `domain/`, and the cursor read is one row read (`packages/api-server/src/modules/runtime-delivery/domain/outbox-progress.ts:27`)
- ✅ **Fixed:** `docs/architecture/skills.md:272` states the re-delivery at the strength the code delivers
- ✅ **Fixed:** `docs/architecture/connections.md:262` separates the answered cursor from the cleanly-applied one
- ✅ **Fixed:** kapetr's two comments — the gate is now read before the disk list it authorises, and a failed bump no longer strands the record
- 🔁 **Still present:** a given-up apply blocks every later reap for that agent (`packages/api-server/src/modules/skills/services/skills-service.ts:1253`)
- 🔁 **Still present:** the driver flattens the typed `SkillsDomainError` into a display string (`packages/agent-runtime/src/modules/runtime-channel/drivers/skill-install-plugin.ts:135`)
- 🔁 **Still present:** `getState` is a query that mutates, now also bumping another module's cursor (`packages/api-server/src/modules/skills/services/skills-service.ts:1273`)
- 🆕 **New:** the reap's stability gate rests on an invariant no writer enforces (`packages/api-server/src/modules/skills/services/skills-service.ts:1265`)
- 🆕 **New:** the staged publish shares a fixed sidecar path between concurrent writes (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:431`)
- 🆕 **New:** the `<` loosening lets a stale equal-version payload overwrite newer applied state (`packages/agent-runtime/src/modules/runtime-channel/service.ts:53`)

### Findings

- 🔴 **Critical:** The reap bracket concludes "one moment" from an unchanged outbox version, but every writer commits the row and bumps the version as two separate awaits (`packages/api-server/src/modules/skills/services/skills-service.ts:1265`).
- 🔴 **Critical:** The PR adds 74 comment lines across 7 code files, none of them pragmas — `skills-service.ts` (31), `local-skill-repository.ts` (10), `outbox-progress.ts` (10), `agents-service.ts` (10), `skill-install-plugin.ts` (5), `runtime-channel/service.ts` (4), `runtime-delivery/compose.ts` (4). This comes from a **recorded reviewer preference** (Lukáš Janeček, 2026-08-13), not from a repo convention: no repo document or lint rule bans code comments.
  **Fix:** Remove comments the PR adds to code files. Where a comment carries information the code cannot, fold it into a name or a type instead.
- 🟡 **Warning:** A permanently-failing install blocks every later reap for that agent (`packages/api-server/src/modules/skills/services/skills-service.ts:1253`). `recordOutcome` holds `lastAppliedVersion` back on each failing settle, and `listRetryable` stops re-dispatching at the cap of 8 (`outbox-repo.ts:239`), so the row is settled, still carries failures, and is over the cap. `applied` then stays false forever and no ghost row for that agent is ever reaped, including rows unrelated to the failed skill. The author's own `SkillNotFoundInSource` validation is exactly this case.
  **Fix:** Give a gate that waits on "the pod caught up" an escape for the given-up state — treat a row at the attempt cap as reap-eligible, or gate on `applyFailures` naming a kind other than `skill-ref`.
- 🟡 **Warning:** `getState` remains a query that mutates, and now reaches further: it bumps the runtime-delivery outbox and enqueues a job from inside the skills module's read path (`packages/api-server/src/modules/skills/services/skills-service.ts:1273`). The prior version only reconciled rows.
  **Fix:** Keep queries side-effect-free — return the ghost set and move the delete-and-re-deliver into a command or a saga subscribed to the drift.
- 🟡 **Warning:** The driver flattens the typed `SkillsDomainError` into an interpolated string, so the api-server cannot tell `SkillNotFoundInSource` from `UpstreamUnreachable` (`packages/agent-runtime/src/modules/runtime-channel/drivers/skill-install-plugin.ts:96`, `:135`). `recordOutcome`'s retry and give-up logic then treats a permanent failure exactly like a transient one.
  **Fix:** Carry the discriminated error across the driver boundary — widen `KindHandler` to return its failures, or let `DriverFailure` carry a structured cause. A typed error must never narrow to a display string at a layer boundary.
- 🟡 **Warning:** The staged publish shares one fixed sidecar pair between attempts (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:431`).
- 🟡 **Warning:** The `finally` block can replace the real error, and can fail an install that already published (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:445`).
- 🟡 **Warning:** The `<=` → `<` loosening lets a stale payload at an equal version overwrite newer applied state (`packages/agent-runtime/src/modules/runtime-channel/service.ts:53`).
- 🟡 **Warning:** `ghosts` is verified but discarded, so the rows checked are not the rows deleted (`packages/api-server/src/modules/skills/services/skills-service.ts:1274`).
- 🟡 **Warning:** `ContributionsProgress` is declared twice across the module seam it protects (`packages/api-server/src/modules/agents/services/agents-service.ts:64`).
- 🟡 **Warning:** `progressOf` returns the row's own mutable `applyFailures` array (`packages/api-server/src/modules/runtime-delivery/domain/outbox-progress.ts:27`).
- 🟡 **Warning:** `hashSkillDir` runs against the published path after the loop, not the staged tree it validated, so a concurrent swap-in can decide the recorded `contentHash` (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:456`). The api-server stores that hash as the install's identity.
  **Fix:** Hash the bytes the call published — compute the hash over the staged tree before the swap-in rename.
- 🟡 **Warning:** The Skills page still describes install as a plain copy into every Skill Path (`docs/architecture/skills.md:123`), although atomic publication is what makes an absent directory trustworthy evidence for the same page's reap gate and `Filesystem is authoritative` invariant.
  **Fix:** State the new observability contract on the owning architecture page beside the invariants that consume it.
- 🟡 **Warning:** The body's change table and Validation section describe only the first four commits, so two later commits carry undeclared behavior (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:430`).
  **Fix:** Extend the PR description whenever a commit lands after it was written.
- 🟡 **Warning:** The Validation section reports a cluster run and green suites that predate `22aa63a` and `40fa9a8`.
  **Fix:** Re-run and re-state validation after any behavior change, or name the SHA the reported run was produced at.
- 🟡 **Warning:** The change table's `state` reconcile rows still describe the pre-`40fa9a8` design, although the architecture pages were updated.
  **Fix:** Keep the change table in sync when a later commit revises an earlier commit's mechanism.
- 🟢 **Suggestion:** `harness-config` still gates on `settled` while the reap moved to the stricter `applied` (`packages/api-server/src/apps/api-server/trpc/context.ts:236`), so its poll reports done for a version a driver failure left unapplied. Worth a deliberate re-check in a separate change now that the port exposes both cursors.
- 🟢 **Suggestion:** No issue tracks this bug or the two items the body defers — the in-request retry, and the `stale`-branch `worker-handler` fingerprint. Both exclusions still hold against the final diff, so the deferrals are honest, but they survive only in this body and are lost on merge.
- 🟢 **Suggestion:** The body's last bullet describes the docs churn as two diagram lines raised in level. The final diff also rewrites invariants and condenses the substrate table (`docs/architecture/connections.md:492`), so a reviewer reads the docs diff as cosmetic.

- 🟡 **Warning:** Install now publishes a skill directory atomically — staged under a dot-prefixed sibling and swapped in with two renames, validated before it becomes visible, with the previous copy restored if the swap fails — but the Skills page still describes install as a plain copy into every Skill Path (`docs/architecture/skills.md:123`, repeated at `:163`). That atomicity is what makes an absent directory trustworthy evidence of an out-of-band deletion, which the same page's reap gate (`:272`) and `Filesystem is authoritative` invariant (`:293`) both rest on, so the page documents the conclusion without the property it depends on.
  **Fix:** When a change alters what a concurrent reader can observe on the pod filesystem, state the new observability contract on the owning architecture page alongside the invariants that consume it — here, that install publishes atomically and a failed write leaves the previous copy in place — describing the guarantee, not the rename mechanics.

Both prior doc findings are resolved: `skills.md:272` now gates the reap on the pod having **cleanly applied** rather than merely settled, and `connections.md:262` now separates the answered cursor from the cleanly-applied one.

### Issue Fit
_No linked issue — scope checked against the PR description only._ `#3208` is a plain reference the body explicitly disclaims ("not caused by it"), and its own scope is the Skills visual rework, so it is not this PR's ask.

- 🟡 **Warning:** The "The change" table and "Validation" section describe only the first four commits. Two later commits add substantial behavior neither declares — the atomic staged publish (`.<name>.staging` / `.<name>.previous` renames plus a restore-on-failure path, `packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:430`) and the double-read reap bracket that re-reads the gate after the rows and now bumps before deleting (`packages/api-server/src/modules/skills/services/skills-service.ts:1244`). A reviewer reading the body would not know either exists.
  **Fix:** When a commit lands after the PR body was written, extend the body's change table and validation section to cover it before requesting review — every substantial behavior in the final diff must be declared in the description, not only in commit messages.
- 🟡 **Warning:** The "Validation" section is stale evidence: it reports a dev-cluster run and green `api-server:check`, `agent-runtime:check` and both unit suites (920 + 235) that predate `22aa63a` and `40fa9a8`. The claim now covers code that was never exercised by the run it describes, including the filesystem rename path and the reap gate that deletes on-disk-backed rows.
  **Fix:** Re-run and re-state validation after any commit that changes behavior, or mark which parts of the reported run predate the final head — a validation section must always name the SHA it was produced at.
- 🟡 **Warning:** The `## The change` table's `state` reconcile rows still describe the pre-`40fa9a8` design — "Clean up only once the sandbox has **applied** its version" and "Tell the sandbox when a cleanup removes entries". The final code gates on `applied` read *twice* with a version-stability check, and orders the bump *before* the delete as a cancellable step; the docs were updated to match (`docs/architecture/skills.md:268`) but the PR body was not.
  **Fix:** Keep the PR body's change table in sync with the final diff whenever a later commit revises an earlier commit's mechanism — the body is the review's entry point, so a superseded description there is a defect even when the code and architecture docs are correct.
- 🟢 **Suggestion:** No issue tracks this bug or the two items the body defers ("Left out": in-request retry for transient fetch errors; "Adjacent wart, untouched": the `stale`-branch `worker-handler` fingerprint that needs the sandbox to return its own fingerprint — a contract change). Both exclusions still hold against the final diff (no retry logic, `worker-handler` untouched), so the deferrals are honest, but they survive only in this PR body and will be lost on merge. File them as issues on the Skills epic so the follow-up work is tracked.
- 🟢 **Suggestion:** The body's last bullet says two `connections.md` diagram lines "were raised in level to make room" for the character cap. The final diff's docs churn is wider than that — invariant rewrites ("Events settle per id, contributions per version", `docs/architecture/connections.md:492`) and substrate-table condensation across `connections.md`. Restate that bullet to cover the actual doc edits, so a reviewer does not read the docs diff as purely cosmetic.

- 🟡 **Warning:** The reap's "one moment" bracket rests on an invariant the code does not enforce — see `### Findings` (`packages/api-server/src/modules/skills/services/skills-service.ts:1265`). Promoted to 🔴 there: it reproduces this PR's own headline symptom.
- 🟡 **Warning:** `write()` stages into a fixed `.<name>.staging` / `.<name>.previous` pair — carried inline (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:431`).
- 🟡 **Warning:** The `finally` block runs its cleanup unguarded — carried inline (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:445`).
- 🟡 **Warning:** `hashSkillDir(firstTarget)` runs after the loop against the published path, not the staged tree it validated, so it reads whatever is at `<skillPath>/<name>` at that moment — including a tree a concurrent writer swapped in (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:456`). The returned `contentHash` is the api-server's identity for the install, so a wrong hash is recorded as the applied state.
  **Fix:** Hash the bytes the call actually published — compute the hash over the staged tree before the swap-in rename — so the hash cannot describe another writer's content.
- 🟡 **Warning:** *(prior finding — still present)* The driver still discards the typed `SkillsDomainError` — full text in `### Findings` (`packages/agent-runtime/src/modules/runtime-channel/drivers/skill-install-plugin.ts:96`, `:135`).
- 🟡 **Warning:** *(prior finding — still present, now amplified)* `getState` remains a query that mutates and now drives a cross-context write — full text in `### Findings` (`packages/api-server/src/modules/skills/services/skills-service.ts:1273`). `skills-service.ts:46` also imports `RuntimeMutator` directly from `../../runtime-delivery/index.js`, a service-port import between modules rather than the event-and-guard boundary `modules.md` → **Module Public API** allows.
- 🟡 **Warning:** `ghosts` is verified but discarded — carried inline (`packages/api-server/src/modules/skills/services/skills-service.ts:1274`).
- 🟡 **Warning:** *(prior finding 4 — replaced by a new variant)* The rename to `ContributionsProgressPort` fixed the name/wiring contradiction, but `ContributionsProgress` is now declared twice — carried inline (`packages/api-server/src/modules/agents/services/agents-service.ts:64`).
- 🟡 **Warning:** `progressOf` returns the row's own mutable array — carried inline (`packages/api-server/src/modules/runtime-delivery/domain/outbox-progress.ts:27`).
- 🟡 **Warning:** The `<=` → `<` loosening in `applyState` — carried inline (`packages/agent-runtime/src/modules/runtime-channel/service.ts:53`).
- 🟢 **Suggestion:** `harness-config` still gates on `settled` while the reap moved to the stricter `applied` (`packages/api-server/src/apps/api-server/trpc/context.ts:236`, `modules/harness-config/compose.ts:12`, `services/harness-config-service.ts:44`), so its "settled" poll reports done for a version a driver failure left unapplied. Now that the port exposes both cursors, that consumer's choice is worth a deliberate re-check in a separate change.

_Prior finding 3 (the settled/applied predicate duplicated four times in the composition root) is fixed: `progressOf` is the single owner and no inline copy remains. One reported claim was dropped in verification: `reconcile`'s return type is `Promise<void>` on `main` too, so this PR does not narrow it — `f247a13` widened it and `40fa9a8` reverted that._

### Verdict
REQUEST_CHANGES — the previous round's findings resolve at the cause and both doc pages are now accurate, but the new reap bracket infers stability from a cursor that no writer moves in the same commit as the row it guards, so a `getState` between an install's row write and its bump reaps the install it was meant to protect; the added code comments are blocking under the recorded reviewer preference.


---

packages/api-server/src/modules/skills/services/skills-service.ts
PetrBulanek: Not implementing. The mechanism is correct; the harm is not.

When an install fails permanently, the reap gate stays shut for that agent. The effect is a stale row that keeps showing in the list — and the failure behind it is now visible, which is the point of this PR. The user's next install or uninstall bumps the row and re-arms the gate.

The proposed escape makes things worse. `bumpVersion` resets `apply_attempts` **and** `apply_failures` (`outbox-repo.ts:123`), so reaping at the attempt cap would erase the failure record this PR adds.

Same finding as the one filed at `:1213` in the previous round.

packages/api-server/src/modules/skills/services/skills-service.ts
PetrBulanek: Not implementing — reasoning on the re-filed copy of this finding at `:1253`.

packages/api-server/src/modules/skills/services/skills-service.ts
PetrBulanek: Not implementing. The read path already wrote before this change: `getState` called `reconcile` and `recordStandaloneSnapshot`, and `runtimeMutator` was already a dependency of this service for `install` and `uninstall`. What is new is one bump on a path that already deleted rows.

Moving the delete-and-re-deliver into a saga puts the re-delivery back on the 60s sweep. The reap and its re-delivery have to be atomic from the reader's point of view, or the pod holds an applied hash naming a set that no longer exists — the failure this PR exists to break.

packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts
PetrBulanek: Not implementing. Two concurrent `write()` calls for the same skill name are not reachable.

`writeFromDir` has one caller: `runInstall` (`services/install.ts:53`). Its only caller is the driver's install loop, a sequential `for await` inside the `applyState` serializer (`runtime-channel/service.ts:27`). `AgentRuntimeSkillsClient` on the api-server side has no `install` method, so there is no second entry point.

`write-local.ts:99` is the collision guard of `writeLocalSkill` — a different function on a different call path. It is not evidence about this one.

packages/agent-runtime/src/modules/runtime-channel/drivers/skill-install-plugin.ts
PetrBulanek: Not implementing. The observation is correct, and nothing consumes the discriminant: `recordOutcome` counts attempts and never branches on the failure kind. `DriverFailure.kind` is a wire contract keyed on `ContributionKind` by design.

Retry classification by error kind is a feature, not a defect. It needs a consumer first, and it changes two contract packages. Out of scope here.

packages/agent-runtime/src/modules/runtime-channel/service.ts
PetrBulanek: Not implementing. `<=` was not protecting against a stale payload. At an equal version it refused whichever payload arrived second — including the newer one. That is the bug this PR fixes.

With `<` the last write wins, and the dispatch that follows a content change always carries the newest build. An equal version with differing contents has one source: the row write and the bump are not in one transaction. That is being fixed on the api-server side, under the critical comment at `skills-service.ts:1265`.

packages/api-server/src/modules/agents/services/agents-service.ts
PetrBulanek: Not implementing. The pattern predates this change — `ContributionsStatus` is already declared in both `agents-service.ts` and `runtime-delivery/compose.ts:62` on `main`. The consumer declaring the port it needs is the convention here.

Drift that matters still fails to compile: a narrowed or removed field breaks the assignment at `compose.ts`. Only an added field is silent, and an added field is harmless to the consumer.

packages/api-server/src/modules/runtime-delivery/domain/outbox-progress.ts
PetrBulanek: Not implementing. `progressOf` receives a fresh `outboxRepo.getRow` result on every call. There is no cache and no shared row, so nothing can be aliased back into infrastructure state.

PetrBulanek:
## Rebased on main — all code comments stripped, deliberately

This branch now sits on top of `main`, which resolves the conflicts. One thing in that rebase deserves calling out, because it changes how the whole diff reads.

**Why the comments are gone.** #3308 stripped comments from every TS/JS/Go source (1156 files) and closed #3010, "Stop agents from writing code comments by default". `CLAUDE.md` now allows no code comments outside the `TEST_OVERVIEW:` / `TEST_SCENARIO:` registry, and `common:check:comment-types` enforces it inside `mise run check`, which CI runs. This branch predates that commit, so its comments survived the rebase and would have failed the gate.

They are removed in `08776497`, as a separate commit, using main's own `scripts/strip-comments.mjs` scoped to the files this branch touches — 79 comments across 12 files. Not by hand: that script lexes with the TypeScript parser and requires each stripped file to re-parse to an identical token stream, so it cannot alter code. `node scripts/check-comment-types.mjs` passes repo-wide.

**Why the rebase was safe.** Across all 15 source files this branch touches, main changed zero non-comment lines — every edit there was comment removal. So the rebase took this branch's version on conflict, with no competing code to lose.

**What that costs, for reviewers of the earlier rounds.** Several review threads on this PR quoted comments that no longer exist — the reap gate read either side of the disk list, the bump-before-delete ordering, and why `bumpForReap` returns a boolean rather than throwing. The reasoning that survives is in [`docs/architecture/skills.md`](docs/architecture/skills.md), the reconciled-state section. That page is at its 40k character cap, so it holds the invariants and not the line-by-line why. Please do not read a bare `return false` in `bumpForReap` as an oversight — cancelling the reap on a failed bump is the fix.

**Test counts.** `agent-runtime` now reports 185 tests where this PR previously reported 235. That is not a loss here: `965a3967` on main replaced a 3044-line `acp-runtime.test.ts` with seven feature-oriented spec files, and this branch's old base predated it. This branch touches no test file. `api-server` stays at 920.

packages/api-server/src/modules/skills/infrastructure/agent-skills-repository.ts
```diff
@@ -130,12 +138,20 @@ export function createAgentSkillsRepository(db: Db): AgentSkillsRepository {
...
 
-    async reconcile(agentId, presentNames) {
+    async listGhosts(agentId, presentNames) {
       const rows = await db
         .select({ name: agentSkills.name, source: agentSkills.source })
         .from(agentSkills)
-        .where(eq(agentSkills.agentId, agentId));
-      const ghosts = rows.filter((r) => !presentNames.has(r.name));
+        .where(
+          and(
```
dam-code-guardian, line 148: 🔴 **Critical:** The age gate does not cover the re-install path.

`GHOST_MIN_AGE_SECONDS` is now the only guard over the window this PR's own comment names — a row write and its bump are two separate awaits, so a `getState` between them sees the row with a stable version. Time closes that window for an **insert**. It does not close it for an **upsert**: `upsertSkill`'s `onConflictDoUpdate` set clause writes `version`, `contentHash` and `path` only, never `installedAt` (`:107`). A skill tracked more than 60 seconds ago therefore keeps its original row age when it is re-installed, so `listGhosts` returns it while the files are still being fetched.

`progressOf` also reports `applied: true` for a **missing** outbox row, so on a first-ever delivery the cursor gate does not hold the reap back either — the age window is all there is.

The result is this PR's headline symptom on the ordinary path: the reap deletes the fresh install and, because it re-delivers, the pod removes the files it just fetched. The PR's own validation exercises exactly this case ("Re-installing then landed the files").

**Fix:** A minimum-age gate must restart its window on every write that renews the row's meaning. Add `installedAt: sql`now()`` to the `onConflictDoUpdate` set, so an upsert is as young as an insert.

packages/api-server/src/modules/skills/services/skills-service.ts
```diff
@@ -924,25 +966,30 @@ export function createSkillsService(deps: SkillsServiceDeps): SkillsService {
...
+      const ghosts =
+        gateBefore?.applied === true
+          ? await deps.agentSkillsRepo.listGhosts(agentId, onDisk)
+          : [];
+      if (gateBefore !== null && ghosts.length > 0) {
+        const gateAfter = await reapGate(deps, agentId);
+        const stable =
+          gateAfter?.applied === true &&
+          gateAfter.version === gateBefore.version;
+        if (stable && (await bumpForReap(deps, agentId))) {
```
dam-code-guardian: 🟡 **Warning:** The reap bumps and enqueues before it deletes, so a delivery can carry the pre-reap set.

`bumpForReap` runs `bump` then `enqueueAfterCommit`, and only then does `reap` delete the rows. `state-builder.ts:131` reads `agent_skills` **live** when it builds the payload, so a worker that picks the job up before the delete commits builds the payload from the pre-reap table and records that version as cleanly applied. The post-reap desired state is then never delivered: the row is settled and clean, so `listRetryable` cannot select it either.

Ordering the bump first is what makes a failed bump cancel the reap, which is the right property — but every other mutation in this file writes the table first and bumps second (`:423`/`:440`, `:640`/`:641`, `:668`/`:669`), and the `enqueue` is what opens the race, not the bump.

**Fix:** Commit the state change before the bump and enqueue on every path — call `reap` first, then `bumpForReap`. To keep the cancel-on-failure property, put the delete and the `bumpVersion` in one transaction and enqueue only after it commits.

packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts
```diff
@@ -344,13 +344,26 @@ async function write(
...
+      await fs.rename(staged, dst);
+      published = true;
+    } finally {
+      await fs.rm(staged, { recursive: true, force: true }).catch(leaveSidecar);
+      if (published) {
+        await fs
+          .rm(previous, { recursive: true, force: true })
+          .catch(leaveSidecar);
+      } else {
+        await fs.rename(previous, dst).catch(ignoreMissing);
```
dam-code-guardian: 🟡 **Warning:** The rollback in `finally` can replace the primary error.

`ignoreMissing` rethrows anything that is not `ENOENT` (`:534`), and this `fs.rename(previous, dst)` runs inside `finally`. If the restore fails for any other reason — `EACCES`, `EBUSY`, `ENOTEMPTY` — that error propagates and the in-flight exception is lost. The real cause, for example an `assertNoSymlinks` rejection, never reaches `runInstall`, so the driver reports the wrong failure for a rejected payload.

The staged-tree cleanup on the line above already gets this right with `leaveSidecar`.

**Fix:** A cleanup or rollback step in a `finally` block must never throw — swallow and log its error, and let the original exception propagate.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `0877649`

## PR #3289: fix(skills): stop a failed skill install from reporting success
**Author:** PetrBulanek | **Branch:** fix/skill-install-silent-failure → main | **Changes:** +218 −98 (17 files)

### Summary
Round 3 answers the previous 🔴 with a real gate: the reap now reads the progress cursor either side of the disk list, defers when the version moved, passes the verified `ghosts` set into a dedicated `reap`, and refuses to reap a row younger than one sweep interval. Eight declines all hold at this HEAD and are resolved, not carried. The residual defects are in the new gate's own write ordering and in the one write path its age window does not cover.

### Changes since last review
Previous HEAD: `40fa9a8` (2026-08-13T13:33:21Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** the reap bracket no longer rests on an unenforced premise — a row younger than `GHOST_MIN_AGE_SECONDS` (60 s, matching the 60 s sweep) is never a ghost, so the un-transacted row-write/bump window is covered by time instead of by the cursor (`packages/api-server/src/modules/skills/infrastructure/agent-skills-repository.ts:148`)
- ✅ **Fixed:** `ghosts` is now passed into `reap`, so the rows verified are the rows deleted (`packages/api-server/src/modules/skills/services/skills-service.ts:987`)
- ✅ **Fixed:** the added code comments are gone — the diff adds zero comment lines, and `main` now bans them in `CLAUDE.md:27` with `common:check:comment-types` enforcing it, so the recorded reviewer preference is now a repo convention
- ✅ **Fixed:** both prior doc pages carry `Last verified: 2026-08-13`, and `skills.md:272` documents the new young-row rule
- ✅ **Resolved (decline holds):** the driver's typed-error flattening — `recordOutcome` only compares `f.kind` sets and never branches on the failure kind, so no consumer loses information
- ✅ **Resolved (decline holds):** the fixed `.staging`/`.previous` sidecars — `write()` has a second entry point through `writeLocalSkill`, but `runWriteLocal` rejects an existing name via `existsInAnyPath` first, so two concurrent same-name writes stay unreachable
- ✅ **Resolved (decline holds):** the `<=` → `<` loosening — at an equal version the code falls through to the hash check, which skips dispatch when the hash is unchanged
- ✅ **Resolved (decline holds):** `getState` mutating, the duplicated `ContributionsProgress`, and `progressOf`'s array — the first two are pre-existing on `main`, and every `progressOf` caller passes a fresh `getRow` result
- 🔁 **Still present:** a permanently-failing install holds `applied` false and blocks every later reap for that agent (`packages/api-server/src/modules/skills/services/skills-service.ts:978`) — the author declined it and the harm reasoning is sound, so it is demoted to 🟢
- 🆕 **New:** the reap bumps and enqueues before it deletes, so a delivery can carry the pre-reap table (`packages/api-server/src/modules/skills/services/skills-service.ts:986`)
- 🆕 **New:** `upsertSkill` does not refresh `installedAt`, so the new age gate does not cover a re-install (`packages/api-server/src/modules/skills/infrastructure/agent-skills-repository.ts:107`)

### Findings

- 🔴 **Critical:** The new age gate does not cover the re-install path — see the inline comment (`packages/api-server/src/modules/skills/infrastructure/agent-skills-repository.ts:146`).
- 🟡 **Warning:** The reap bumps and enqueues before it deletes — see the inline comment (`packages/api-server/src/modules/skills/services/skills-service.ts:986`).
- 🟡 **Warning:** The `finally` rollback can replace the primary error — see the inline comment (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:365`).
- 🟡 **Warning:** The PR body's change table and Validation section still describe the first four commits (issue-fit). Three later behaviours are undeclared — the atomic stage-and-rename publish, the 60-second minimum row age, and the double-read stability gate — and the ghost-cycle bullet now contradicts the code, because one page load no longer reaps a young row. The corrected `agent-runtime` test count (185, not 235) lives only in a PR comment.
  **Fix:** Add one change-table row per behaviour a commit introduces, and re-run and re-state validation in the same commit that changes the validated behaviour; corrected numbers belong in the body, not in a comment.
- 🟢 **Suggestion:** A permanently-failing install still holds `applied` false past the attempt cap, so ghost reconciliation stops for that agent (`packages/api-server/src/modules/skills/services/skills-service.ts:978`). The decline is sound — `bumpVersion` resets `apply_failures`, so reaping at the cap would erase the failure record this PR adds — but letting the gate pass on the terminal `gaveUp` state, without a bump, would close the deadlock without that cost. Worth a separate change.
- 🟢 **Suggestion:** The "No new tests" note names the driver's throw as the one cleanly testable seam. The branch adds two more injectable seams — the pure `progressOf` reducer (`packages/api-server/src/modules/runtime-delivery/domain/outbox-progress.ts:19`) and `write()`'s rollback path (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:359`).

- 🟡 **Warning:** `docs/architecture/skills.md` still describes install as a direct copy into every Skill Path, but install now stages the copy, swaps it in, and restores the previous copy on failure (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:344`). The same page's reap gate depends on a disk listing never showing a half-written directory, so the guarantee is load-bearing and absent (`docs/architecture/skills.md:163`, also `:123`).
  **Fix:** When a subsystem gains an atomicity guarantee, state the guarantee on its architecture page and link out for the mechanism; never leave the page describing the superseded write behaviour.
- 🟡 **Warning:** The Install flow shows the success path only, but the `skill-ref` driver now fails its whole slice when any one skill fails, so the contribution never reads as cleanly applied and the sweep retries it (`packages/agent-runtime/src/modules/runtime-channel/drivers/skill-install-plugin.ts:126`). A reader cannot tell that one broken skill holds a whole batch unapplied (`docs/architecture/skills.md:191`).
  **Fix:** Document every terminal outcome of a flow on the page that owns it — name what a failure leaves behind and who retries it — not the success path alone.

Both prior doc findings are resolved, and the young-row rule at `skills.md:272` matches the 60 s constant.

### Issue Fit
_No linked issue — scope checked against the PR description only._ `#3208` is a plain reference the body disclaims.

- 🟡 **Warning:** The body's change table and Validation section are stale against the final diff — mirrored into `### Findings` above with its **Fix:**.
- 🟢 **Suggestion:** The "No new tests" justification no longer matches the diff — mirrored above.
- 🟢 **Suggestion:** No issue tracks this bug or the two items the body defers (the in-request retry, the `stale`-branch `worker-handler` fingerprint). Both exclusions still hold, so the deferrals are honest, but they are lost on merge.

- 🟡 **Warning:** The bump-before-delete ordering in the reap — carried inline (`packages/api-server/src/modules/skills/services/skills-service.ts:986`).
- 🟡 **Warning:** `upsertSkill` does not refresh `installedAt`, so the age gate misses the re-install path — promoted to 🔴 and carried inline (`packages/api-server/src/modules/skills/infrastructure/agent-skills-repository.ts:146`).
- 🟡 **Warning:** The `finally` rollback can replace the primary error — carried inline (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:365`).
- 🟢 **Suggestion:** The `applied`-forever deadlock past the attempt cap — carried in `### Findings`.

_All six declines from the previous round were verified at this HEAD and dropped as resolved. Notable: the sidecar decline understated the call graph — `write()` has a second entry point via the `writeLocal` tRPC route, outside the `applyState` serializer — but `runWriteLocal`'s `existsInAnyPath` pre-check still makes a same-name concurrent write unreachable, so the conclusion holds._

### Verdict
REQUEST_CHANGES — the stability gate and the young-row rule are the right cut and eight declines hold, but `upsertSkill` never refreshes `installedAt`, so the age window that now guards the un-transacted install window does not cover a re-install; the reap also bumps before it deletes, which can deliver the pre-reap set.

---

packages/api-server/src/modules/skills/infrastructure/agent-skills-repository.ts
PetrBulanek: Fixed in 4707cfdf. Confirmed: the `onConflictDoUpdate` set clause never wrote `installedAt`, so a re-installed row kept its original age and `listGhosts` returned it while the files were still being fetched. `installedAt: sql`now()`` is now in the set clause, so an upsert is as young as an insert.

The deeper premise is fixed too. Each row write and its version bump now share one transaction (`install`, `uninstall`, and the batch path), and the enqueue happens after that commit. So "version unchanged across the bracket" really implies "no row changed", and the age gate is a backstop rather than the only guard.

packages/api-server/src/modules/skills/services/skills-service.ts
PetrBulanek: Fixed in 4707cfdf. Confirmed, and it was unrecoverable as you said — `listRetryable` selects unsettled rows or rows holding failures, so a row settled clean at the pre-reap version is selected by neither.

Taken the suggested shape: the delete and `bumpVersion` now run in one transaction (`reapWithRedelivery`), and `enqueueAfterCommit` follows the commit. The cancel-on-failure property is preserved by the transaction itself — if either statement fails, nothing is deleted and no version moved. Same ordering now applies to every mutation in the file, which also matches the documented `BEGIN, write, bump, COMMIT` then enqueue flow in the runtime-channel page.

packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts
PetrBulanek: Fixed in 7d20887f. The rollback now uses `leaveSidecar` instead of `ignoreMissing`, so it cannot throw out of `finally` and the original exception survives. The ENOENT distinction had no consumer there. Note `agent-runtime` has no ambient logger in this repository — the skills module injects `deps.log`, which this file does not receive — so the error is swallowed rather than logged.

This also covers the earlier copy of this finding at :364.

packages/api-server/src/modules/skills/services/skills-service.ts
PetrBulanek: Already addressed on the branch, and now stronger. `ghosts` is passed into `reap`, and as of 4707cfdf the delete is a single statement scoped by `or(...)` over exactly those `(source, name)` keys. So the act covers precisely the set the gate checked, and a row that appeared after `gateAfter` cannot be caught.

packages/api-server/src/modules/skills/services/skills-service.ts
PetrBulanek: Already addressed on the branch. A failed bump no longer leaves the rows deleted: as of 4707cfdf the delete and the bump share one transaction, so a failure rolls both back and the next `getState` finds the ghosts again. Only the enqueue is best-effort, and that one is genuinely recoverable — the bump has committed, so the row is unsettled and `listRetryable` does select it.

docs/architecture/skills.md
PetrBulanek: Doc updated. The page no longer states the re-delivery as a bare guarantee — it now says the delete and its re-delivery bump commit together, which is what 4707cfdf implements.

packages/api-server/src/modules/skills/services/skills-service.ts
PetrBulanek: Addressed, in two parts. The gate is now read either side of the disk list, so an install that lands mid-read moves the version and defers the reap. The residue you were pointing at — a row written but not yet bumped, where the version does not move at all — is fixed in 4707cfdf: the row write and the bump now share a transaction, so that window no longer exists.

packages/api-server/src/modules/skills/services/skills-service.ts
PetrBulanek: Addressed exactly along the split you described. The two steps now have separate error handling: the version bump is inside the transaction with the delete, so a failure rolls the delete back and nothing is lost; only the message to the sandbox is best-effort, and that one the 60-second sweep really does repeat, because the bump has committed and the record shows as not complete.

dam-code-guardian (approved):
## Summary

Round 4 closes the three blockers from round 3: `upsertSkill` now refreshes `installedAt` on conflict so the age gate restarts after every re-install, the reap's delete and bump commit atomically in one transaction, and the `finally` rollback uses `leaveSidecar(): void {}` so a secondary cleanup failure cannot replace the primary error. Eight declines from earlier rounds all hold at this HEAD. Two quality items remain open and two are still present from previous rounds.

## Changes since last review

Previous HEAD: `0877649` (2026-08-13T16:32:09Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** `upsertSkill`'s `onConflictDoUpdate` now includes `installedAt: sql\`now()\`` so a re-installed skill's row age restarts and the 60 s age gate guards the re-install window (`packages/api-server/src/modules/skills/infrastructure/agent-skills-repository.ts:107`)
- ✅ **Fixed:** `reapWithRedelivery` wraps `reap + bump` in one `unitOfWork` transaction; `enqueueAfterCommit` runs separately after the commit, so a failed bump cancels the delete (`packages/api-server/src/modules/skills/services/skills-service.ts:232`)
- ✅ **Fixed:** `leaveSidecar(): void {}` swallows cleanup errors in the `finally` block so a secondary rollback failure never replaces the primary one (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:344`)
- ✅ **Fixed:** code comments removed — zero comment lines in the diff, and `common:check:comment-types` in CI now enforces it as a repo convention
- 🔁 **Still present:** `docs/architecture/skills.md:163` describes Install as a plain copy into every Skill Path, without the atomic staged-publish guarantee the reap gate depends on
- 🔁 **Still present:** the PR body's change table and Validation section cover only the first four commits; six of ten commits are undeclared

## Findings

### Approval bar

Each open 🟡 names the fix that resolves it. The two remaining items do not affect runtime correctness.

---

🟡 **Warning:** `docs/architecture/skills.md:163` still describes Install as "copies it into every configured Skill Path" without mentioning the atomic staged publish (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:344`). The page's own reap gate (`:272`) and `Filesystem is authoritative` invariant (`:293`) rest on the guarantee that an absent directory is trustworthy evidence — which holds only because a failed write leaves the previous copy in place. The Install description documents the result without the property it depends on.
**Fix:** Add one sentence to the Install bullet stating that the skill directory is published atomically (staged under a dot-prefixed sibling, swapped in on success, previous copy restored on failure) so the page's own invariants have their stated basis.

🟡 **Warning:** The PR body's change table and Validation section cover only the first four commits. Six of ten commits introduce behavior the body does not declare: `9b75149` (atomic staged publish), `974e582` (evidence-from-one-moment gate), `fe284ac` (hold-off during in-flight install), `4707cfd` (installedAt refresh), `7d20887` (rollback masking fix), `72c82f1` (comment removal). The corrected test count (185, per a PR comment) lives outside the body. A reviewer reading the description encounters code not announced there.
**Fix:** Add one change-table row per undeclared behavior commit and re-run or re-state validation at the final HEAD before requesting review — the body is the review's entry point; superseded or absent entries there are a defect even when the code and architecture docs are correct.

`docs/architecture/connections.md` carries `Last verified: 2026-08-13`. The documentation guidelines define the date as "whenever you edit the page" — this page was last edited on 2026-08-13 and the date is correct. Both prior doc findings remain resolved.

🟡 **Warning (from prior round, still present):** `docs/architecture/skills.md:163` — mirrored into `### Findings` above.

### Issue Fit

✅ Delivers everything promised in the PR description; no undeclared scope.

All four findings from the skill were verified against the clone and the baseline on `main`:

1. `skills-service.ts:28-29` imports `AgentsRepository` and `computeAgentState` from `agents/infrastructure/` — these imports are present on `main` before this PR, not introduced here. **Not a finding.**
2. `outbox-progress.ts:1` imports `DriverFailure` from `api-server-api` in the domain layer — this pattern appears in at least eleven other domain files in the codebase (including `capability-filter.ts`, `contribution-hash.ts`, `connections/domain/`, `agents/domain/`). **Not a finding.**
3. `ContributionsProgress` on the `RuntimeDeliveryComposition` interface — `ContributionsProgress` is not exported from `runtime-delivery/index.ts` and callers use the `agents`-module definition; the layering observation was noted and declined in round 3. **Not a new finding.**
4. Duplicate `ContributionsProgress` definition — explicitly noted and declined in round 3 (two definitions at different layers, structurally identical but intentionally separated). **Not a new finding.**

_No new TypeScript findings this round._

## Verdict

**APPROVE** — the three blockers resolve at the cause and hold up under verification. The two open 🟡 items (Install atomicity in the doc, PR body coverage) are quality gaps that do not affect correctness; addressing them before merge would make the history cleaner but is not required. The 60 s age gate, the double-read stability bracket, and the atomic reap are each individually sound; the four-layer fix is complete.

---

```json
{"findings":[{"severity":"warning","status":"still","file":"docs/architecture/skills.md","line":163,"inline":false,"summary":"Install description omits the atomic staged-publish guarantee the reap gate depends on","fix":"Add one sentence to the Install bullet stating that the skill directory is published atomically","round":2},{"severity":"warning","status":"still","file":"packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts","line":344,"inline":false,"summary":"PR body change table and Validation section cover only 4 of 10 commits; 6 undeclared","fix":"Add one change-table row per undeclared behavior commit and re-state validation at the final HEAD","round":2}]}
```

PR #3302, opened 2026-08-13
fix(e2e): gate on the apiserver route the reset just bounced

jezekra1:
## Context

The original content of this PR (Traefik entrypoint port fix, shared `e2e:gate`, IPv4 pins) is **already on `main`**: #3303 was stacked on this branch and its squash-merge (`b8fd2622`) carried all three commits with it. After rebasing, this PR reduces to the one code-guardian finding that survived verification.

## Problem

`e2e:reset` scales the apiserver to 0 and back immediately before `e2e:gate` runs (on both the fresh `e2e` path and `e2e:loop`'s warm path), but the gate polled only the UI root and the Keycloak realm — the one route that bounced was the one not checked. `kubectl wait --for=condition=Available` covers pod readiness, not Traefik's endpoint window (the same race the `e2e:install` UI-restart note documents), so the specs' first `/api` call could land in a transient 502.

## Fix

Add the apiserver's public readiness route (`/api/ready` — it also 503s until JWKS warm-up completes, so it is a strictly stronger signal than pod-Available) to the gate's three-consecutive-ticks streak condition.

## Code-guardian findings disposition

- ✅ **Fixed**: gate does not poll the component the reset bounced (`tasks.toml:118`) — this PR.
- ❌ **Invalid**: `E2E_VM_NAME` missing at the `e2e:install` → `e2e:gate` call site. Every path into `e2e:install` carries `E2E_VM_NAME` in the process environment (`tasks.toml:110` sets it on the child; `e2e:loop` exports it before bootstrapping), and env vars propagate to nested `mise run` children — the gate always resolves the same VM name as its caller. The claimed fallback is unreachable.
- ❌ **Invalid (would reintroduce the bug)**: moving the Node IPv4 pin into `globalSetup` via `dns.setDefaultResultOrder`. That call sets per-process state, and Playwright runs specs in separate worker processes that do not inherit it — the pin would cover only the runner while the workers' Keycloak/tRPC fetches revert to happy-eyeballs. The `NODE_OPTIONS` env var is inherited by every Node child, which is why it is the right mechanism.
- ⏭️ **Declined**: resolving `APP_PORT` from `values-local.yaml` too. Correct today by the review's own admission (`values-local.yaml` declares no `port`), and `VIRT_ENABLED` above it has the same shape — future-proofing both is a separate cleanup if ever needed.
- Body/validation findings: superseded by this rewrite.

## Validation

The original content was validated 27/27 in-sandbox before it merged via #3303. This follow-up is a one-line addition to the gate's curl chain; an in-sandbox `mise run e2e` revalidation is running and the result will be posted here.

### DIFF
--- a/tasks.toml
+++ b/tasks.toml
@@ -175,7 +175,7 @@ mise run e2e:gate
 
 ["e2e:gate"]
 hide = true
-description = "Wait until the e2e cluster stably serves UI + Keycloak on :5555. Shared by `e2e` (via e2e:install) and `e2e:loop`'s warm path."
+description = "Wait until the e2e cluster stably serves UI + API + Keycloak on :5555. Shared by `e2e` (via e2e:install) and `e2e:loop`'s warm path."
 dir = "{{config_root}}"
 raw = true
 run = '''
@@ -190,14 +190,20 @@ else
 fi
 
 # The install (or a warm --rebuild) rolls traefik, keycloak, and the
-# apiserver, and their routes come up at different moments — a single lucky
-# 200 on the UI root is not "serving". Require the UI root AND the keycloak
-# realm to answer on three consecutive ticks before any spec dials in;
-# otherwise the auth specs die at ERR_CONNECTION_REFUSED mid-login-redirect.
-echo "Waiting for the cluster to serve UI + Keycloak on :5555 ..."
+# apiserver — and e2e:reset bounces the apiserver right before this gate runs
+# on both entry points. Their routes come up at different moments — a single
+# lucky 200 on the UI root is not "serving", and the deployment reporting
+# Available doesn't mean Traefik's endpoint for it stopped 502ing (see the
+# UI-restart note in e2e:install). Require the UI root, the apiserver's
+# public readiness route, AND the keycloak realm to answer on three
+# consecutive ticks before any spec dials in; otherwise the auth specs die
+# at ERR_CONNECTION_REFUSED mid-login-redirect or the first /api call lands
+# in the endpoint window.
+echo "Waiting for the cluster to serve UI + API + Keycloak on :5555 ..."
 STREAK=0
 for _ in $(seq 1 100); do
   if curl -fsS -o /dev/null --max-time 3 http://localhost:5555/ 2>/dev/null &&
+     curl -fsS -o /dev/null --max-time 3 http://localhost:5555/api/ready 2>/dev/null &&
      curl -fsS -o /dev/null --max-time 3 http://keycloak.localhost:5555/realms/platform 2>/dev/null; then
     STREAK=$((STREAK + 1))
     [ "$STREAK" -ge 3 ] && break
@@ -207,7 +213,7 @@ for _ in $(seq 1 100); do
   sleep 3
 done
 if [ "$STREAK" -lt 3 ]; then
-  echo "ERROR: e2e cluster is up but not stably routing on :5555 (UI + keycloak)." >&2
+  echo "ERROR: e2e cluster is up but not stably routing on :5555 (UI + API + keycloak)." >&2
   echo "It does not auto-heal — reprovision with 'mise run e2e', or check" >&2
   echo "'LIMA_INSTANCE=$VM_NAME mise run cluster:fix-certs'." >&2
   # Same signature the ztunnel-cert-watchdog keys on (deploy/tasks.toml);


dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `0686ad09`

## PR #3302: fix(e2e): make the in-sandbox e2e run actually work
**Author:** jezekra1 | **Branch:** fix/sandbox-e2e#locki-lpaitkhu → main | **Changes:** +49 −5 (4 files)

### Summary
Makes `mise run e2e` work inside a sandbox: the Traefik `HelmChartConfig` port now follows `--set=port=…`, `e2e:install` gains a UI-plus-Keycloak readiness streak, and browser and spec DNS resolution are pinned to IPv4.

### Findings
- 🟡 **Warning:** Readiness gate reaches only one of the two entry points it claims (`tasks.toml:164`).
- 🟡 **Warning:** Port resolver reads `--set=port=…` only, not the values-file input (`deploy/tasks.toml:240`).
- 🟢 **Suggestion:** `MAP *.localhost 127.0.0.1` does not cover the bare `localhost` the UI base URL uses (`packages/e2e/playwright/playwright.config.ts:38`); only `keycloak.localhost` matches the pattern. A second `MAP localhost 127.0.0.1` rule would cover both hosts the specs dial.
- 🟢 **Suggestion:** `e2e-playwright:run-ui` is the one spec-running task without the `NODE_OPTIONS` pin the other three got (`packages/e2e/playwright/tasks.toml:40`), so interactive runs keep the IPv6-first failure.
- ✅ **Looks good:** The `STREAK`/`break` loop is safe under `set -eo pipefail` — verified in a script file with a `/bin/false` control that the failing `[ "$STREAK" -ge 3 ]` inside the `if` branch does not abort the task, so the gate cannot exit 0 early or die on a non-match tick.

✅ No findings.

### Issue Fit
_No linked issue — scope checked against the PR description only._ (`#3177` resolves to a merged PR, not an issue; the head branch carries no issue number.)

- Both 🟡 findings above were reported here and are carried in `### Findings`.
- 🟢 **Suggestion:** The Validation line evidences one case — the sandbox `mise run e2e` path. "Host/lima behavior unchanged" and the `e2e:loop` share are unexercised; naming the unrun paths would make the claim match the evidence.

✅ No findings.

### Verdict
**COMMENT** — the three causes are correctly diagnosed and the fixes are sound on the path that was validated. Two 🟡 items are open, both where a claimed scope is wider than the code reaches: the readiness gate misses `e2e:loop`'s warm path, and the port resolver misses the values-file input.

---

tasks.toml
```diff
@@ -160,6 +160,29 @@ mise run cluster:install -- "${INSTALL_ARGS[@]}"
 # breaks Playwright's first asset fetch — leave them alone.
 kubectl --kubeconfig="$KUBECONFIG" rollout restart deployment platform-apiserver
 kubectl --kubeconfig="$KUBECONFIG" rollout status deployment platform-apiserver --timeout=5m
+
+# Readiness gate, shared by `e2e` and `e2e:loop`: the install rolls traefik
```
dam-code-guardian: 🟡 **Warning:** This gate is declared "shared by `e2e` and `e2e:loop`", but it lives inside `e2e:install`, which `e2e:loop` runs only when `NEEDS_BOOTSTRAP` is true (`tasks.toml:289`). The normal warm loop keeps its own weaker gate at step 3b (`tasks.toml:324`): UI root only, **one** successful tick, no Keycloak realm. That path runs `e2e:reset` immediately before, which scales the apiserver to 0 and back — so the Keycloak-flap cause this PR names stays unfixed on the path developers use most.

**Fix:** A readiness contract claimed for two entry points must be reached by both — extract this streak into one task both paths call, or replace step 3b's single-tick poll with it.

_(Also reported by the issue-fit skill.)_

deploy/tasks.toml
```diff
@@ -231,11 +231,19 @@ else
...
-# 1c. Traefik :4444 entrypoint — on lima this rides the provision script's
+# 1c. Traefik app entrypoint — on lima this rides the provision script's
 # k3s server manifest, but a pre-existing k3s (VM sandbox) never runs that,
 # so the app port silently doesn't exist. HelmChartConfig via kubectl apply
-# is idempotent and works on both.
-kubectl --kubeconfig="$KUBECONFIG" apply -f - <<'YAML'
+# is idempotent and works on both. The port follows `--set=port=…` so the
+# ingress listens where the chart's URLs point (dev defaults to 4444, the
+# e2e install passes 5555); the last install wins.
+APP_PORT=4444
```
dam-code-guardian, line 240: 🟡 **Warning:** The resolver reads only `--set=port=…`, so a port supplied through `--helm-values` / `values-dev.yaml` leaves the entrypoint on 4444 while the chart renders its URLs at the values-file port (`_helpers.tpl:119` builds the host from `.Values.domain`/`.Values.port`). That is the same chart-vs-entrypoint divergence this PR fixes, reached from the other input path — and `cluster:install` auto-selects `values-dev.yaml` when it exists (`deploy/tasks.toml:87`), so it needs no explicit flag to happen.

**Fix:** Resolve every port-like install input from both sources before use — values file via `yq` first, then `--set` overrides — exactly as `VIRT_ENABLED` already does in this file (`deploy/tasks.toml:96`).

_(Also reported by the issue-fit skill.)_

packages/e2e/playwright/playwright.config.ts
```diff
@@ -31,6 +31,12 @@ export default defineConfig({
     trace: "on",
     screenshot: "only-on-failure",
     video: "on",
+    launchOptions: {
+      // Defensive: the sandbox k3s ingress binds IPv4 only (::1 refuses),
+      // so pin `*.localhost` to IPv4 rather than depending on the browser's
+      // address-fallback behavior. No-op where both stacks serve (lima).
+      args: ["--host-resolver-rules=MAP *.localhost 127.0.0.1"],
```
dam-code-guardian: 🟢 **Suggestion:** `*.localhost` does not match the bare hostname `localhost`, and the UI base URL is `http://localhost:5555` (`src/config.ts:1`) while only Keycloak is a subdomain (`keycloak.localhost`). So the browser-side pin covers the Keycloak host and leaves the UI host on the browser's own address-fallback behavior — the very thing the comment says it does not want to depend on. The spec-side `NODE_OPTIONS` pin has no such gap, since `--dns-result-order` is global.

```suggestion
      args: ["--host-resolver-rules=MAP *.localhost 127.0.0.1,MAP localhost 127.0.0.1"],
```

jezekra1:
Addressed all five review findings in 7cdd4ed8: the stability gate is extracted into a shared `e2e:gate` task called by both `e2e:install` and `e2e:loop`'s warm path (🟡1); the Traefik port now resolves from the helm values file before `--set` overrides, mirroring the `virtualization.enabled` pattern (🟡2); the browser resolver rules map bare `localhost` too (🟢); `run-ui` got the `NODE_OPTIONS` pin (🟢); and the PR's Validation section now states which paths are reasoned rather than tested (🟢).

tasks.toml
```diff
@@ -109,6 +109,14 @@ fi
 # e2e:install, shared with e2e:loop so the two never drift.
 E2E_VM_NAME="$VM_NAME" mise run e2e:install
 
+# Wipe data to the specs' expected clean slate. On lima/CI the cluster is
+# brand new and this is a fast no-op; in a sandbox the cluster persists
+# between runs and still carries the previous run's data (the specs'
+# clean-slate assertions fail on it). Re-gate afterwards — the reset
+# bounces the apiserver.
+E2E_VM_NAME="$VM_NAME" mise run e2e:reset
+E2E_VM_NAME="$VM_NAME" mise run e2e:gate
```
dam-code-guardian, line 118: 🟡 **Warning:** `e2e:gate` polls the UI root and the Keycloak realm, but not `/api`. On this path the gate runs right after `e2e:reset`, which scales the apiserver to 0 and back — so the one component the reset bounces is the one the gate does not check. The reset's own `wait --for=condition=Available` covers pod readiness, not the Traefik endpoint window this same file documents at `tasks.toml:167` ("restarting the UI introduces a transient Traefik-endpoint 502 window that breaks Playwright's first asset fetch"). The specs' first `/api` call can land in that window.

**Fix:** A gate that runs after a component bounces must poll that component's route. Add a third `curl` on the apiserver's own readiness path to the streak condition in `e2e:gate`, so all three routes must answer together:

```
  if curl -fsS -o /dev/null --max-time 3 http://localhost:5555/ 2>/dev/null &&
     curl -fsS -o /dev/null --max-time 3 http://localhost:5555/api/ready 2>/dev/null &&
     curl -fsS -o /dev/null --max-time 3 http://keycloak.localhost:5555/realms/platform 2>/dev/null; then
```

`/api/ready` is the path the apiserver's own `readinessProbe` uses (`deploy/helm/platform/templates/apiserver/app.yaml:447`), and the ingress routes `/api` to that service (`_helpers.tpl:129`).

tasks.toml
```diff
@@ -160,6 +168,56 @@ mise run cluster:install -- "${INSTALL_ARGS[@]}"
 # breaks Playwright's first asset fetch — leave them alone.
 kubectl --kubeconfig="$KUBECONFIG" rollout restart deployment platform-apiserver
 kubectl --kubeconfig="$KUBECONFIG" rollout status deployment platform-apiserver --timeout=5m
+
+# Readiness gate — shared task, see e2e:gate.
+mise run e2e:gate
```
dam-code-guardian, line 173: 🟡 **Warning:** `e2e:gate` derives `KUBECONFIG` from `E2E_VM_NAME` (`VM_NAME="${E2E_VM_NAME:-platform-k3s-test}"`), but this call site passes it as a bare `mise run e2e:gate`. `e2e:install` sets `VM_NAME` as a local shell variable and never exports it, so on a lima run with a non-default VM name the gate falls back to `platform-k3s-test` and reads a kubeconfig that does not belong to this cluster. The two `curl` polls do not use `KUBECONFIG`, so the damage is confined to the mesh-cert hint: it silently finds no logs and the operator loses the one diagnostic the failure branch exists to give.

**Fix:** Pass the variable at every call site of a task that reads it — the other two sites already do (`tasks.toml:118`, `tasks.toml:361`):

```
E2E_VM_NAME="$VM_NAME" mise run e2e:gate
```

packages/e2e/playwright/playwright.config.ts
```diff
@@ -31,6 +31,14 @@ export default defineConfig({
     trace: "on",
     screenshot: "only-on-failure",
     video: "on",
+    launchOptions: {
```
dam-code-guardian, line 21: 🟡 **Warning:** The IPv4 pin is split across two files, and only the browser half lives in the config: the Node half is `NODE_OPTIONS=--dns-result-order=ipv4first` in the four `e2e-playwright:run*` mise tasks. A direct `pnpm exec playwright test` therefore gets browser resolution pinned but leaves the specs' own Node `fetch` calls to `keycloak.localhost` (token mint in `src/lib/auth.ts:22`, tRPC clients) on happy-eyeballs — the failure this PR removes. Reported by typescript-engineering.

**Fix:** Give one behaviour one owner — pin both halves inside `playwright.config.ts`: keep `launchOptions.args` for Chromium and add a `globalSetup` module that calls `dns.setDefaultResultOrder("ipv4first")`, then drop the `NODE_OPTIONS` env from every `e2e-playwright:run*` task in `packages/e2e/playwright/tasks.toml`. This makes the fix hold on every invocation path, not only through mise.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `9492bead`

## PR #3302: fix(e2e): make the in-sandbox e2e run actually work
**Author:** jezekra1 | **Branch:** fix/sandbox-e2e#locki-lpaitkhu → main | **Changes:** +93 −29 (4 files)

### Summary
Complete re-review at `9492bead`. Both 🟡 findings from `0686ad09` are fixed at the cause: the readiness streak became a shared `e2e:gate` task that both entry points call, and the Traefik port now resolves from the values file before `--set` overrides. A later commit (`9492bea`) added `e2e:reset` + a re-gate to the fresh `e2e` path.

### Changes since last review
Previous HEAD: `0686ad09` (2026-08-13T09:22:23Z) — verdict COMMENT

- ✅ **Fixed:** Readiness gate now reaches both entry points — extracted to `e2e:gate`, called by `e2e:install` and by `e2e:loop` step 3b (`tasks.toml:173`, `tasks.toml:361`).
- ✅ **Fixed:** Port resolver reads the values file then `--set` overrides, mirroring `VIRT_ENABLED` (`deploy/tasks.toml:240-247`).
- ✅ **Fixed:** Resolver rules map bare `localhost` (`playwright.config.ts:39`).
- ✅ **Fixed:** `run-ui` got the `NODE_OPTIONS` pin (`packages/e2e/playwright/tasks.toml:44`).
- ✅ **Fixed:** The Validation section now separates tested from reasoned claims.
- 🆕 **New:** The fresh path's re-gate does not cover the apiserver the reset just bounced (`tasks.toml:118`).
- 🆕 **New:** `e2e:gate` resolves `KUBECONFIG` from `E2E_VM_NAME`, which `e2e:install` does not export (`tasks.toml:173`).
- 🆕 **New:** The IPv4 pin's Node half lives outside the config (reported by typescript-engineering).
- 🆕 **New:** The body does not declare the fresh-path data wipe (reported by issue-fit).

### Findings
- 🟡 **Warning:** Gate does not poll the component the reset bounced (`tasks.toml:118`).
- 🟡 **Warning:** `E2E_VM_NAME` missing at the `e2e:install` call site (`tasks.toml:173`).
- 🟢 **Suggestion:** The port resolver reads only the `--helm-values` file, while the install always passes `-f values-local.yaml` ahead of it (`deploy/tasks.toml:622`). `values-local.yaml` declares no `port` today, so the resolver is correct — it becomes wrong the moment someone adds one there.

✅ No findings.

### Issue Fit
_No linked issue — `#3177` in the body is a PR, not an issue; scope checked against the PR description only._

- 🟡 **Warning:** The fresh-run path now calls `e2e:reset` before the specs, which destroys existing cluster data, but the body's "Fix" section does not declare this change (`tasks.toml:112-118`). Reviewers approve a data-destroying step on the headline path without notice.
  **Fix:** Declare every substantial behavior change in the PR body — add the fresh-path data wipe and its re-gate to the "Fix" bullets; re-read the body after each new commit.
- 🟡 **Warning:** The "Validation" section claims a green 27/27 `mise run e2e` run, but commit `9492bea` added the reset plus re-gate to that path after the run (`tasks.toml:112-118`). The measured evidence no longer covers the path the PR headlines.
  **Fix:** Keep validation claims true to the final diff — after a commit that changes an already-validated path, re-run the validation or name the untested step in the body.

- 🟡 **Warning:** The IPv4 pin is split across two files, and only the browser half lives in the config: the Node half is `NODE_OPTIONS=--dns-result-order=ipv4first` in the four `e2e-playwright:run*` mise tasks. A direct `pnpm exec playwright test` therefore gets browser resolution pinned but leaves the specs' own Node `fetch` calls to `keycloak.localhost` on happy-eyeballs, which is the failure this PR removes. (`packages/e2e/playwright/playwright.config.ts:34`)
  **Fix:** Give one behaviour one owner — pin both halves inside `playwright.config.ts`: keep `launchOptions.args` for Chromium and add a `globalSetup` module that calls `dns.setDefaultResultOrder("ipv4first")`, then drop the `NODE_OPTIONS` env from every `e2e-playwright:run*` task in `packages/e2e/playwright/tasks.toml`. This makes the fix hold on every invocation path, not only through mise.

### Verdict
**COMMENT** — every prior finding is fixed at the cause, and the shared-gate extraction is the right cut. Five 🟡 items are open, all on the new commit's path: the re-gate skips the component the reset bounced, one call site omits `E2E_VM_NAME`, and the body's Fix and Validation sections predate that commit.

---

tomkis (approved):

Issue #3288, opened 2026-08-12
Eventing architecture for live UI updates — one bus or three? (options & recommendation)


jezekra1:
Publishing the eventing-architecture decision draft for discussion. Written while designing #3170 (live UI updates, on top of the #3258 transport); related to #3281 (two-phase Postgres+Kubernetes writes — the sibling reliability gap). The in-repo copy lives at `docs/plans/live-events-eventing-options.md` on the live-subscriptions branch; this issue is where we argue and decide. The decision lands as an ADR.

---

# Eventing architecture for live UI updates — options

**Status:** decision draft (feeds an ADR once we pick; not an architecture page).
**Context:** issue #3170 replaces UI polling with pushed invalidation hints over the
tRPC WebSocket transport (merged in #3258). Hints must reach a browser tab pinned to
*any* api-server replica, which forces a cross-replica hop and reopens a standing
question: how many event mechanisms should this app have, and what guarantees does
each need? We are willing to refactor large portions of the existing code if the
end-state is better.

The one thing already settled (previous discussion): **services emit exactly one
kind of event — the domain event.** UI hints are a *projection* of domain events,
produced by a single bridge, never hand-emitted at write sites. Every option below
respects that; they differ in what carries the events and with what guarantees.

---

## 1. What exists today

Three messaging mechanisms, each with different (and deliberate) semantics:

| Mechanism | Guarantee | Tenants |
|---|---|---|
| In-process RxJS bus (`events.ts`) | synchronous, at-most-once, same-process, rich typed payloads, zero serialization | 16 `DomainEvent` types; sagas below |
| `RedisBus` pub/sub (`core/redis-bus.ts`) | cross-replica broadcast, fire-and-forget, at-most-once, lost during subscriber disconnect | approvals wake `approval:<id>`, ACP inject `inject:<agentId>` |
| BullMQ (`core/periodic-jobs.ts`) | durable jobs, at-least-once, **competing consumers** (each job runs once fleet-wide) | periodic jobs (reconciles, sweeps) |

Saga consumers of the in-process bus (all run on the replica that emitted):

- `audit/audit-log` — projects actor-bearing events to **security log lines**
  (ChannelTurnRelayed, ScheduleFired, FilesImported, …). Not a table: duplicates
  can't be deduped by a unique key, they flood the forensic trail. Its docstring
  already warns: *"Domain events must not be moved onto the cross-replica Redis
  bus without dedup, or every replica's saga would duplicate every line."*
[...]

PR #3303, opened 2026-08-13
feat: UI live updates over tRPC WebSocket subscriptions (#3170)

jezekra1:
## What

Replaces the UI's polling (2–15s intervals across ~14 sites) with pushed invalidation hints over the tRPC WebSocket transport (#3258). Implements #3170; the eventing shape is recorded as **ADR-083** (in this PR): domain events stay in-process and non-durable, consumed by sagas; cross-replica reach is a Redis signal; durable fleet-wide work is a BullMQ job.

## How it works

**Contract** — `events.owner` subscription streams thin `LiveEvent` hints: `{topic, ids}`, never state. Every (re)subscribed stream opens with `sync` ("refetch everything you map") — reconnects heal by refetch, not replay. Agent-bound API keys are refused (the stream spans the whole owner).

**Server** — services emit **domain events** (the only thing write sites know); the live-events module owns the hint concept end to end:
- `live-hints` saga projects domain events → per-owner Redis channels (`events:owner:<sub>`), zod-parsed on the receiving side so version-skewed frames drop instead of poisoning the stream
- `ownerStream` async generator behind the subscription: subscribe-before-`sync`, per-connection bounded pending queue — on overflow the queue collapses to a single `sync`, which by contract is the full recovery
- new domain events at every UI-visible write: approvals (request + all resolution paths incl. expiries; resolution only emits from the call that actually flipped the row), schedule CRUD, harness-config (user apply + pod-confirmed report), artifacts (+ folders + expiry sweeper), experiments (incl. the SDK surface and the inactivity reaper)
- the approvals wake (`approval:<id>`) now rides the same rail: a `wake-held-calls` saga replaces the service's direct `notifyResolved` calls
- **agents topic = a K8s watch** on the Agent CRs (owner from the CR label): covers controller- and pod-driven transitions no write site could. The watch runs on a single lease-elected replica, and hints fire only when a content fingerprint (minus reconcile bookkeeping like the last-activity stamp) actually moves — no chatter at idle, no per-replica duplicates.

**UI** — the tRPC client is **WebSocket-only** (no splitLink): one authenticated connection for queries, mutations, and subscriptions; connection state feeds the existing api-health tracker. A per-tab `events.owner` subscription drives a coarse topic→query-family invalidation map. All topic-covered `refetchInterval`s are gone.

**Terms** — the gate moved from transport admission to a per-procedure middleware (terms.* exempt), so any transport serves the full app including first-run acceptance; the HTTP door keeps its 412 pre-gate for the CLI. Stale terms are signaled by a structured `data.termsStale` field on the error, not message text. Terms versions ship with deploys and acceptance is irrevocable, so the per-process proven-cache can only be stale in the safe direction.

Still polling — no event source exists yet: sessions + files (pod-sourced truth — follow-up design), metrics (ClickHouse aggregation), the reachability probe, and the bounded skill-install watch.

## Validation

- 928/928 api-server unit tests; unit coverage for the stream (sync-first, per-owner scoping, parse-and-drop, overflow-to-sync, abort teardown) and the hint projection
- two pure-API e2e specs: WS connection auth, and the full pipeline — an agent created over HTTP surfaces as a hint on the WebSocket via real Redis in <100ms
- full smoke suite green against a real cluster running this branch — every browser spec exercised the WS-only UI (login, terms, wizard, chat, approvals hold→approve→release through the new wake saga)

## Notes for review

- The e2e-infra fix is also up separately as #3302 and is folded into the squashed feature commit here so the branch stays self-testable; it dedupes on merge.
- ADR-083 supersedes the working options document that #3288 published; the issue carries the discussion trail.


packages/api-server-api/src/trpc.ts
```diff
@@ -21,9 +21,26 @@ const tBase = initTRPC.context<ApiContext>().create({
...
 
+const termsProven = new WeakSet<ApiContext>();
+
+export function markTermsProven(ctx: ApiContext): void {
+  termsProven.add(ctx);
+}
+
+const requireTermsAccepted = tBase.middleware(async ({ ctx, path, next }) => {
+  if (path.startsWith("terms.")) return next();
+  if (!termsProven.has(ctx)) {
```
dam-code-guardian, line 48: 🔴 **Critical:** The `termsProven` WeakSet caches acceptance per context object. A WebSocket context lives for the whole connection (`trpc/ws.ts:77`), so terms accepted once stay proven until the socket closes. Terms revoked or superseded mid-connection gate nothing, and `isAccepted` is version-scoped — it goes false the moment a new terms version publishes.

**Fix:** Bound every authorization cache to one request, or give it an explicit TTL, whenever the permission it caches can change while the cached scope is alive.

packages/api-server/src/modules/approvals/services/approvals-service.ts
```diff
@@ -199,7 +204,7 @@ export function createApprovalsService(
           "allow_once",
           deps.ownerSub,
         );
-        await deps.notifier.notifyResolved(id);
+        emitResolved(row);
```
dam-code-guardian: 🔴 **Critical:** `emitResolved(row)` runs before the code knows `casWon`. The same shape is at `:372`, and at `:392` (`resolveAndDeliverAcpNative`) the emit is unconditional. A caller that loses the race still emits `ApprovalResolved`, and the new `wake-held-calls` saga then wakes the held call on a resolution this caller did not perform. ADR-083, added in this PR, requires services to emit after the write commits.

**Fix:** Emit a domain event only on the branch where the state change is proven — gate every emit behind the CAS or affected-row result, never before it.

packages/api-server/src/modules/live-events/services/live-events-service.ts
```diff
@@ -0,0 +1,38 @@
...
+      return new ReadableStream<LiveEvent>({
+        start(controller) {
+          end = () => {
+            unsubscribe();
+            try {
+              controller.close();
+            } catch {}
+          };
+          controller.enqueue({ topic: "sync" });
+          unsubscribe = deps.bus.subscribe(sub, (event) =>
```
dam-code-guardian: 🔴 **Critical:** The listener calls `controller.enqueue(event)` and ignores the result, and the stream declares no `highWaterMark`. A consumer that stops reading — a throttled background tab, a stalled socket — accumulates every hint for its owner in the api-server heap until the connection closes. Hints are per owner, so one slow tab holds the whole owner's traffic.

**Fix:** Bound any server-held per-connection queue: cap it, and on overflow drop the queue and enqueue one `sync` hint instead. The contract already makes `sync` mean "re-read everything", so a bounded queue costs no correctness.

packages/ui/src/app.tsx
```diff
@@ -30,6 +31,8 @@ export default function App() {
 
   useBrowserHistory();
 
+  useLiveEvents(view !== "terms");
```
dam-code-guardian: 🟡 **Warning:** `useLiveEvents(view !== "terms")` runs above these early returns, so `telegram-bind` and `slack-bind` hold an owner subscription they never consume. Each transition of `view` through `terms` also tears the subscription down and rebuilds it on the socket.

**Fix:** Mount a live-invalidation subscription in the component that owns the query cache it invalidates, and gate it on one stable boolean — not on a multi-valued route variable that also drives early returns.

packages/api-server/src/modules/harness-config/compose.ts
```diff
@@ -15,9 +20,32 @@ export function composeHarnessConfigModule(deps: {
...
+  ): Promise<void>;
+} {
+  const repo = createHarnessConfigSnapshotRepo(deps.db);
+  return {
+    async merge(agentId, patch, opts) {
+      await repo.merge(agentId, patch, opts);
+      if (!opts.confirmed) return;
+      const ownerSub = await deps.resolveOwner(agentId);
+      if (!ownerSub) return;
+      emit({ type: EventType.HarnessConfigChanged, agentId, ownerSub });
```
dam-code-guardian: 🟡 **Warning:** This emit and the one in `harness-config-service.ts:71` both fire for one logical change — user apply, then pod confirmation — so a single `apply` the runtime confirms produces two hints for the same agent.

**Fix:** Give each domain event exactly one emitting site, owned by the service that performs the state change; a downstream confirmation step must not re-emit the same event type.

packages/api-server/src/modules/harness-config/compose.ts
```diff
@@ -15,9 +20,32 @@ export function composeHarnessConfigModule(deps: {
...
       getCapabilities: deps.getCapabilities,
       isSettled: deps.isSettled,
     }),
   };
 }
+
+export function createHarnessConfigSnapshotWriter(deps: {
+  db: Db;
+  resolveOwner: (agentId: string) => Promise<string | null>;
+}): {
```
dam-code-guardian: 🟡 **Warning:** `createHarnessConfigSnapshotWriter` wraps a repository and emits events, which makes it an infrastructure adapter, but it is defined in the composition root and re-exported from `index.ts`.

**Fix:** Define adapters under `infrastructure/`, and keep `compose.ts` limited to wiring existing factories together.

packages/api-server/src/modules/live-events/compose.ts
```diff
@@ -0,0 +1,54 @@
...
+  stop(): void;
+  startAgentWatch(): void;
+  stopAgentWatch(): void;
+}
+
+export function composeLiveEventsModule(deps: {
+  bus: RedisBus;
+  log: (message: string) => void;
+  k8s?: Pick<K8sClient, "watchCustomObjects">;
+}): LiveEventsModule {
```
dam-code-guardian, line 27: 🟡 **Warning:** `k8s` is optional and `startAgentWatch()` returns silently when it is absent, so a missing wiring becomes a no-op rather than a type error. The agents topic is the one hint source no write site can replace (it carries controller- and pod-driven transitions), so silence here loses coverage the design depends on.

**Fix:** Declare every module dependency as required; express an on/off capability in the dependency itself, configured where it is built.

packages/api-server/src/modules/live-events/compose.ts
```diff
@@ -0,0 +1,54 @@
...
+import type { RedisBus } from "../../core/redis-bus.js";
+import { createRedisLiveEventsBus } from "./infrastructure/redis-live-events-bus.js";
+import { startAgentWatch } from "./infrastructure/k8s-agent-watch.js";
+import { createLiveEventsService } from "./services/live-events-service.js";
+import { startLiveHintsSaga } from "./sagas/live-hints.js";
+import {
+  AGENTS_PLURAL,
+  LABEL_OWNER,
+  LAST_ACTIVITY_KEY,
+} from "../agents/infrastructure/labels.js";
```
dam-code-guardian, line 12: 🟡 **Warning:** `labels.js` and `agents/infrastructure/k8s.js` are private paths of the `agents` module — `agents/index.ts` exports neither — and `live-events` publishes no `index.ts` of its own, while 21 of 27 sibling modules do.

**Fix:** Import another module only through its `index.ts`, and give each new module an `index.ts` exporting its composer and public types; move a genuinely shared constant to a shared location.

packages/ui/src/api.ts
```diff
@@ -1,36 +1,20 @@
...
+    `${location.protocol === "https:" ? "wss" : "ws"}://${location.host}/api/trpc-ws`,
+  connectionParams: async () => ({ token: await getAccessToken() }),
+  lazy: { enabled: true, closeMs: 30_000 },
+  keepAlive: { enabled: true },
+  onOpen: () => onFetchSuccess(),
+  onError: () => onFetchError(),
+  onClose: () => onFetchError(),
+});
 
 export const api = createTRPCClient<AppRouter>({
```
dam-code-guardian, line 18: 🟡 **Warning:** This is the UI's only `AppRouter` client and it is WebSocket-only, so the PR body's "terms flow keeps a dedicated HTTP client" describes a construct the diff does not contain — `terms.current`, `terms.latestAcceptance` and `terms.accept` all ride this socket. The only non-WS call left is the plain `fetch("/api/terms")` that already existed on `main`. The gate still works, because the new middleware exempts `terms.*` by path, but reviewers cannot verify that from the description.

**Fix:** State in the PR body only the mechanisms the diff introduces, and name the real construct that satisfies each admission constraint.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `a583af37`

## PR #3303: feat: UI live updates over tRPC WebSocket subscriptions (#3170)
**Author:** jezekra1 | **Branch:** feat/live-subscriptions#locki-lpaitkhu → main | **Changes:** +1563 −217 (73 files)

### Summary
The PR replaces UI polling with pushed invalidation hints over a WebSocket-only tRPC client, adds domain events at every UI-visible write, and records the eventing layers as ADR-083. The layering is clean and the hint contract — thin hints, `sync`-first, parse-and-drop — is well designed. The blocking findings sit in the terms gate that moved layers, in emits that fire before their write is proven, and in the repo's own comment gate.

### Findings
- 🔴 **Critical:** The terms middleware caches acceptance in a `WeakSet` keyed on the context object. A WebSocket context lives for the whole connection (`trpc/ws.ts:77`), so terms accepted once stay proven until the socket closes. Terms revoked or superseded mid-connection do not gate anything, and the connection can last as long as the credential. (`packages/api-server-api/src/trpc.ts:32`)
  **Fix:** Bound every authorization cache to one request, or give it an explicit TTL, whenever the permission it caches can change while the cached scope is alive.
- 🔴 **Critical:** `emitResolved(row)` runs before the code knows the compare-and-set won. At `:207` and `:372` the emit precedes the `casWon` test, and at `:392` it is unconditional. A lost race therefore emits `ApprovalResolved`, and the new `wake-held-calls` saga wakes the hold on a resolution this caller did not perform. ADR-083, added by this PR, requires services to emit after the write commits. (`packages/api-server/src/modules/approvals/services/approvals-service.ts:207`)
  **Fix:** Emit a domain event only on the branch where the state change is proven — gate every emit behind the CAS or affected-row result, never before it.
- 🔴 **Critical:** `ownerStream` enqueues into a `ReadableStream` with no backpressure handling and no queue bound. `controller.enqueue` is called from the Redis listener with the result ignored, so a tab that stops reading — a throttled background tab, a stalled socket — accumulates every hint for its owner in the api-server's heap until the connection closes. Hints are per owner, so one slow consumer holds the whole owner's traffic. (`packages/api-server/src/modules/live-events/services/live-events-service.ts:26`)
  **Fix:** Bound any server-held per-connection queue: cap it, and on overflow drop the queue and enqueue one `sync` hint instead — the contract already makes `sync` the "re-read everything" recovery, so a bounded queue loses no correctness.
- 🔴 **Critical:** The PR adds six untyped comments to `scripts/strip-comments.mjs:84-93`. `main` now enforces `COMMENT_TYPES` (`TEST_OVERVIEW`, `TEST_SCENARIO`) over every code comment through `common:check:comment-types` (`tasks.toml:455`), and `strip-comments.mjs` self-exempts only from `listFiles`, not from the checker. The whole file is also byte-identical to `main`, so these lines are rebase fallout that re-adds a file the branch does not need to change. (`scripts/strip-comments.mjs:84`)
  **Fix:** Drop the file from the branch when it matches `main`; where a comment must stay, give it a registered type prefix, or fold its content into naming so the gate passes. This finding also reflects a recorded reviewer preference against code comments.
- 🟡 **Warning:** `useLiveEvents(view !== "terms")` sits above three early returns in `App`, so the `telegram-bind` and `slack-bind` views hold an owner subscription that they never consume (`app.tsx:34`, `:52-53`). Every transition through `terms` also tears the subscription down and rebuilds it. (`packages/ui/src/app.tsx:34`)
  **Fix:** Mount a live-invalidation subscription in the component that owns the query cache it invalidates, and gate it on one stable boolean — never on a multi-valued route variable that also drives early returns.
- 🟡 **Warning:** `HarnessConfigChanged` is emitted from two independent sites for one logical change: the service on user apply (`harness-config-service.ts:71`) and the snapshot writer on pod confirmation (`compose.ts:48`). One `apply` that the runtime later confirms produces two hints for the same agent. (`packages/api-server/src/modules/harness-config/services/harness-config-service.ts:71`)
  **Fix:** Give each domain event exactly one emitting site, owned by the service that performs the state change; a downstream confirmation step must not re-emit the same event type.
- 🟡 **Warning:** `DockedArtifactPanel` no longer polls, and `useArtifact` sets `retry: false`, so the removed `refetchInterval` was the only recovery path. The panel tests `!artifact` and never reads `isError`, so a transient network error now renders "Artifact not found — it may have been deleted." permanently. (`packages/ui/src/modules/artifacts/components/docked-artifact-panel.tsx:109`)
  **Fix:** When a query drops its `refetchInterval`, keep an explicit recovery path — retry, or a distinct error branch with a retry control. Never render an absent-data message from `!data` alone when the query can fail.
- 🟡 **Warning:** `pinnedVersion` is component-local state derived from a server entity, and nothing resets it when `openArtifactId` changes; the mount site renders `<DockedArtifactPanel />` with no `key` (`chat-view.tsx:725`). Opening a second artifact keeps the first one's pinned version and can request a version that does not exist. The 5-second poll formerly refreshed `latest` and hid this. (`packages/ui/src/modules/artifacts/components/docked-artifact-panel.tsx:34`)
  **Fix:** Reset or re-key local state derived from a server entity whenever the entity identity changes — `key={openArtifactId}`, or clear the pinned value in an effect on identity change.
- 🟡 **Warning:** `ChatArtifactsPanel` renders "No artifacts yet" whenever the list is empty, including the first load before data arrives; it destructures only `data`, while the sibling `sandbox-artifacts-section.tsx:23` already reads `isLoading` from the same hook. (`packages/ui/src/modules/artifacts/components/chat-artifacts-panel.tsx:50`)
  **Fix:** Distinguish loading from empty for every list query — read the query's own pending state and render it before the empty state; never infer emptiness from a defaulted array.
- 🟡 **Warning:** `isTermsStaleError` identifies a terms failure by matching the message string `"terms not accepted"`, coupled to the literal thrown at `trpc.ts:34`. Any wording change silently breaks the redirect, and this is now the UI's only terms-stale detection since the 412 body check went away. The codebase discriminates tRPC errors by `data.code` elsewhere. (`packages/ui/src/modules/terms/lib/on-terms-stale.ts:20`)
  **Fix:** Discriminate errors on a stable machine-readable field — a tRPC error code or an explicit cause token — never on human-readable message text.
- 🟡 **Warning:** The contract declares `ownerStream(): AsyncIterable<LiveEvent>` but the implementation returns a `ReadableStream`. The two agree only through a runtime `Symbol.asyncIterator` the declared type does not express, and the test reaches for `[Symbol.asyncIterator]()` to consume it. (`packages/api-server-api/src/modules/events/types.ts:4`)
  **Fix:** Make the implementation return the type the contract declares — an async generator for an `AsyncIterable` contract, or declare the concrete stream type in the interface.
- 🟡 **Warning:** `live-events` imports `agents/infrastructure/labels.js` and `agents/infrastructure/k8s.js`, neither exported by `agents/index.ts`, and the new module publishes no `index.ts` of its own — 21 of 27 sibling modules do. (`packages/api-server/src/modules/live-events/compose.ts:12`)
  **Fix:** Import another module only through its `index.ts`, and give each new module an `index.ts` that exports its composer and public types; move a genuinely shared constant to a shared location.
- 🟡 **Warning:** `eventsRouter` re-implements the wildcard-principal check inline with a different message, duplicating `requireWildcardBinding` in the same package. Two copies of one authorization rule drift apart. (`packages/api-server-api/src/modules/events/router.ts:7`)
  **Fix:** Express each authorization rule once as a shared procedure or middleware, and compose routers from those procedures.
- 🟡 **Warning:** `composeLiveEventsModule` takes `k8s` as optional and `startAgentWatch()` returns silently when it is absent, so missing wiring becomes a silent no-op instead of a type error — and the agents topic is the one hint source no write site can replace. (`packages/api-server/src/modules/live-events/compose.ts:27`)
  **Fix:** Declare every module dependency as required; express an on/off capability in the dependency itself, configured where it is built.
- 🟡 **Warning:** `createHarnessConfigSnapshotWriter` wraps a repository and emits events, so it is an infrastructure adapter, but it is defined in `compose.ts` and re-exported from `index.ts`. The composition root must wire adapters, not define them. (`packages/api-server/src/modules/harness-config/compose.ts:34`)
  **Fix:** Define adapters under `infrastructure/`, and keep `compose.ts` limited to wiring existing factories.
- 🟢 **Suggestion:** `invalidation.ts` uses an exhaustive `Record<Topic, …>`, so a new topic becomes a compile error. `hintFor` ends in `default: return null`, so a new event type silently projects nothing — the same exhaustive shape there would give the projection the same guarantee. (`packages/api-server/src/modules/live-events/sagas/live-hints.ts:60`)
- 🟢 **Suggestion:** The `open` prop no longer affects fetching in `ChatArtifactsPanel`, so a collapsed panel still fetches and re-fetches its agent's artifact list. Passing `null` as the filter while collapsed restores the previous cost profile. (`packages/ui/src/modules/artifacts/components/chat-artifacts-panel.tsx:36`)
- ✅ **Looks good:** The reconnect story holds up end to end — `sync`-first on every (re)subscribe plus the `onlineManager` refetch at `query-client.ts:86` means a dropped socket heals by refetch, and the K8s watch fingerprint excludes the last-activity annotation, so idle agents generate no hint traffic.

- 🟡 **Warning:** The page says the terms flow alone uses tRPC over HTTP, but the UI has one `AppRouter` client and it is WebSocket-only. The terms read and the accept mutation both travel over the socket, so a reader plans against a transport the UI no longer uses. (`docs/architecture/platform-topology.md:43`, edge table row at `:85`)
  **Fix:** State the transport each client actually uses, confirmed at the client construction site. When a PR moves a flow to a new transport, correct every page statement about that flow, diagram and table rows included.
- 🟡 **Warning:** The page says connection admission on the socket requires accepted terms. The PR removed that admission check and replaced it with a per-procedure gate that permits `terms.*`. The stated reason for keeping the HTTP door is obsolete, and the page misplaces the gate. (`docs/architecture/platform-topology.md:43`)
  **Fix:** Describe a gate at the layer that enforces it, and update the page when a PR moves the gate between layers.
- 🟡 **Warning:** The page says two things are pinned rather than shared and names in-process MCP sessions and the channel workers. The PR adds a third install-wide singleton: a Redis lease elects one replica to run the K8s Agent watch. The count is wrong, and the new events paragraph names the watch without saying one replica alone runs it. (`docs/architecture/platform-topology.md:50`, `:52`)
  **Fix:** Keep enumerations of singletons complete, and give each new lease-elected component its election rule on the page in the same PR.
- 🟡 **Warning:** The page says server-pushed hints replace list polling and that pod-sourced reads are the remaining polls. Two api-server polls remain: the per-session spend read at 60 s over the telemetry store, and the experiment setup check on the skills state. A reader concludes no api-server polling remains. (`docs/architecture/platform-topology.md:75`)
  **Fix:** Write an exhaustive claim only when it holds for every case, or qualify it; verify the remaining poll sites before describing what is left.

### Issue Fit
Linked issue: #3170 "Replace UI polling with live updates over a websocket" — referenced without a closing keyword, so merging will not auto-close it.

- 🟡 **Warning:** The issue's "Done when" requires that no screen depends on an interval and that an idle tab makes no recurring traffic, but five of the fourteen named polling sites still poll: sessions at 5 s (`packages/ui/src/modules/sessions/api/queries.ts:120`), the file tree and each open folder at 2 s (`packages/ui/src/modules/files/api/queries.ts:63,80`), metrics at 60 s (`packages/ui/src/modules/metrics/api/queries.ts:62`), and the reachability probe at 3 s (`packages/ui/src/modules/agents/hooks/use-agent-reachability-probe.ts:35`). The PR body declares these as deliberate exceptions, so the delivery is partial rather than hidden.
  **Fix:** When a PR cannot meet every acceptance criterion, keep the issue open and record the remaining deliverables as tracked follow-up items, so no criterion is silently dropped at merge.
- 🟡 **Warning:** The issue requires polling to survive "only as a fallback for when that connection is down", but the retained pollers run unconditionally while the socket is healthy — the file tree still refetches every 2 s on a live connection. (`packages/ui/src/modules/files/api/queries.ts:63`)
  **Fix:** Gate every retained interval on the live connection state, so a healthy push transport suppresses the timer and the poller acts only as a documented fallback.
- 🟡 **Warning:** The PR body claims the "terms flow keeps a dedicated HTTP client", but the UI has one WebSocket-only tRPC client and the terms procedures ride it (`packages/ui/src/api.ts:18`, `packages/ui/src/modules/terms/api/mutations.ts:5`). The only non-WS call is the plain `fetch("/api/terms")` that already existed on `main`. Reviewers cannot verify the terms path from the description.
  **Fix:** State in the PR body only the mechanisms the diff introduces, and name the real construct that satisfies each admission constraint.
- 🟡 **Warning:** The `scripts/strip-comments.mjs` `git ls-files` fallback is a behavior change that neither the issue asks for nor the body declares (`scripts/strip-comments.mjs:85-99`). Undeclared tooling changes ride into a feature merge without a reviewer's attention.
  **Fix:** Declare every out-of-scope change in the PR body, or move it to its own PR, as the author already did for the e2e-infra commit in #3302.
- 🟢 **Suggestion:** The budgets query lost its 5 s interval (`packages/ui/src/modules/budgets/api/queries.ts:8`) and now depends on the `agents` topic, yet no domain event reports a budget reservation change on its own. A budget change with no agent-CR movement may display stale until the next agent hint.

Scope items that check out: ADR-083 is delivered and indexed; cross-replica reach is met through per-owner Redis channels; the deleted `01-auth-deep-link.spec.ts` was relocated into `01-auth.spec.ts:92`, so it is mechanical fallout, not lost coverage; the CLI fixture and unit-test edits are forced fallout of the new middleware and the `watchCustomObjects` port; and the `tasks.toml` overlap with #3302 is declared in the body.

The blocking findings from this section are carried in `### Findings` above: the pre-CAS `emitResolved`, the two `HarnessConfigChanged` emit sites, the adapter defined in `compose.ts`, the optional `k8s` dependency, the `agents` private-path imports and missing `index.ts`, the `AsyncIterable`-vs-`ReadableStream` contract mismatch, the duplicated wildcard check, the message-string error discrimination, and the connection-lifetime terms cache. The `hintFor` exhaustiveness suggestion is also there.

The findings from this section are carried in `### Findings` above: the `useLiveEvents` gate above the early returns, the terminal fetch error in `DockedArtifactPanel`, the unreset `pinnedVersion`, the loading-vs-empty conflation in `ChatArtifactsPanel`, and the collapsed-panel fetching suggestion.

### Verdict
REQUEST_CHANGES — the terms gate now expires only with the connection, `ApprovalResolved` can fire on a lost race, the per-connection hint queue is unbounded, and the added comments fail the repo's own comment-type gate.

---

jezekra1:
Response to the Code Guardian review @ `a583af37`. Everything actionable is addressed on the branch (now at `38a5b91e`, rebased onto main); four findings are disputed with reasoning below.

## Fixed

- **Unbounded per-connection queue (critical)** — `ownerStream` is now a bounded async generator: pending hints cap at 256, and on overflow the queue collapses to a single `sync`, which the contract already defines as full recovery (`023c4bb6`). This also resolves the `AsyncIterable`-vs-`ReadableStream` mismatch — the implementation returns what the contract declares, and the test consumes it as a plain async iterable. One framing correction: each subscription registers its own listener on the broadcast bus, so a stalled tab held only *its own* queue copy — never "the whole owner's traffic" — and the connection is credential-bounded. The leak was real; its blast radius was one connection. Bounded regardless.
- **Pre-CAS `ApprovalResolved` emits (critical)** — every resolution emit is now gated on "this call flipped the row": the CAS result, or the expired-path transition (`resolveExpired` now reports whether it moved the row). Losing racers stay silent (`023c4bb6`).
- **Terms-stale detection by message text** — the middleware attaches a structured cause and the error formatter lifts `termsStale: true` into `data`; the UI and the auth e2e spec read the field, not the string (`3062492d`).
- **Duplicated wildcard-principal check** — `requireWildcardBinding` is exported and reused by the events router (`44bebef5`).
- **Adapter in the composition root / missing module index** — the snapshot writer moved to `harness-config/infrastructure/`, and `live-events` publishes an `index.ts` that bootstrap imports (`328c8e04`).
- **`hintFor` default arm** — exhaustive switch over every event type with a `never` check, matching `invalidation.ts` (`023c4bb6`).
- **All four UI findings** — docked panel distinguishes error (with a retry control) and loading from deleted; `key={openArtifactId}` re-keys the panel so `pinnedVersion` can't leak across artifacts; the chat artifacts list reads its pending state and skips fetching while collapsed; `useLiveEvents` moved inside `MainApp`, below the early returns (`f4711db3`).
- **All four doc-drift findings** — the topology page now describes the per-procedure terms gate and the WS-only client, adds the agent-watch lease to the singleton set with its election rule, and qualifies the remaining-poll claim (`38a5b91e`). The PR body's false "dedicated HTTP terms client" claim is corrected too.

## Disputed

- **Terms cache lives as long as the connection (critical)** — the staleness this guards against has no reachable trigger. Terms versions ship as deploys: publishing a new version restarts the api-server, which severs every WebSocket along with every cached proof. Acceptance is irrevocable — no revoke path exists anywhere in the system. And connections are already credential-bounded (the server nudges a reconnect before token expiry). So the cache can only be stale in the safe direction (a re-check that would succeed anyway). The invariant is now stated on the topology page; if terms ever become revocable or hot-swappable, the gate must move to request scope — that change would have to revisit this line.
- **Comments fail the repo comment gate (critical)** — factually wrong: `check-comment-types.mjs` iterates exactly `listFiles()` from `strip-comments.mjs`, and `listFiles` excludes `scripts/strip-comments.mjs` itself, so the gate never inspects those lines — `mise run common:check:comment-types` was green at the reviewed head. Also moot: the tooling fix merged separately as #3312, the rebase dropped the commit, and the file no longer appears in this diff.
- **Two `HarnessConfigChanged` emit sites** — deliberate: they report two distinct facts on one topic. The service emits when the user's *declared* config changes; the snapshot writer emits when the pod-*confirmed* snapshot changes. Both are UI-visible transitions, and suppressing the confirmation emit would leave the confirmed view stale until an unrelated hint arrived. Hints are idempotent invalidations, not domain facts with single provenance — the cost of two hints for one apply-then-confirm cycle is one redundant refetch; the cost of one hint would be a lost update.
- **Retained pollers not gated on connection state / issue fit** — the surviving polls (sessions, files, metrics, reachability) are *primary* reads of pod- or store-sourced truth, not fallbacks for the push channel: the server has no event source for them, so gating them on socket health would only make them stale while the socket is healthy. The issue's "poll only as fallback" criterion applies to the server-sourced lists that hints now cover — all fourteen of those are event-driven. #3170 stays open, and the pod-sourced event design (sessions, files — the budgets-reservation gap noted in the suggestion belongs there too) is being filed as a tracked follow-up.

On the optional `k8s` dependency: kept as-is — it models the watch as a capability (the unit harness composes the module without cluster access), production bootstrap always passes the client, and a wiring regression fails loudly in the e2e pipeline spec that asserts an agent hint arrives via the watch. On the `agents/infrastructure` deep imports: that's the codebase's current idiom — six other modules (files, invocations, budgets, secret-store, …) import `agents/infrastructure/{k8s,labels}` directly; unifying that is a repo-wide refactor out of scope here.

Verification on the updated branch: `mise run check` green (comment gate included), full unit suites green (api-server 928/928, incl. new overflow and CAS-gating coverage). The smoke e2e rerun against a cluster built from this head is in flight; I'll confirm here when it lands.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `9ffc1ff7`

## PR #3303: feat: UI live events via tRPC WebSocket subscription (round 3)

_Re-review requested via `code-guardian-review` label. Full re-review at current HEAD `9ffc1ff7`._

---

### Changes since last review

Compared to round 2 (`a583af37`, REQUEST_CHANGES, 2026-08-13). All 4 critical and all 12 warnings from round 2 are resolved. Two prior findings were held-decline per author's responses (reviewed and confirmed).

**🔴 Critical — all resolved:**
| # | Finding | Status |
|---|---------|--------|
| 1 | Terms cache: stale on revoke/expiry | ✅ Held decline: deploy=server restart→all WS severed; acceptance irrevocable; stale-safe direction confirmed. `createContext` per-connection (WS), per-request (HTTP) — both correct. |
| 2 | Pre-CAS `emitResolved` calls | ✅ Fixed: all 5 sites gated on `casWon` or `expiredFlipped` |
| 3 | Unbounded queue, AsyncIterable contract | ✅ Fixed: `async *ownerStream` with `MAX_PENDING=256`; overflow collapses to `{ topic: "sync" }` |
| 4 | `strip-comments.mjs` prose comment gate | ✅ Held decline: `listFiles()` self-excludes the file; `check-comment-types.mjs` never inspects it |

**🟡 Warnings — all resolved:**
| # | Finding | Status |
|---|---------|--------|
| 5 | `useLiveEvents()` before early returns | ✅ Fixed: moved to `MainApp()` below the 3 early returns |
| 6 | Terms-stale detection by message text | ✅ Fixed: now reads `error.data?.termsStale` |
| 7 | `DockedArtifactPanel` no error/loading/deleted state | ✅ Fixed: 3-way guard (`isError → isPending → !artifact`) |
| 8 | Pinned version leaks across artifacts | ✅ Fixed: `key={openArtifactId}` on `<DockedArtifactPanel>` (chat-view.tsx:725) |
| 9 | `ChatArtifactsPanel` no loading state | ✅ Fixed: `null` query when collapsed; `isPending` conditional on expand |
| 10 | `AsyncIterable` vs `ReadableStream` contract mismatch | ✅ Fixed: async generator implements `AsyncIterable` natively |
| 11 | `requireWildcardBinding` duplicated inline | ✅ Fixed: extracted to `auth-procedures.ts`; imported in `router.ts` |
| 12 | `k8s` dependency optional in `compose.ts` | ✅ Fixed: `k8s: Pick<K8sClient, "watchCustomObjects">` (required) |
| 13 | Adapter logic in `compose.ts` | ✅ Fixed: `createHarnessConfigSnapshotWriter` in `infrastructure/snapshot-writer.ts` |
| 14 | Missing `live-events/index.ts` | ✅ Fixed: exports `composeLiveEventsModule` and `LiveEventsModule` |
| 15 | `agents/infrastructure` private path imports | ✅ Held decline: codebase idiom — `rg "from.*modules/agents/infrastructure"` shows 10+ files across the server, all import directly |
| 16 | Retained pollers not gated on connection state | ✅ Held decline: `files` and `sessions.status` are pod-sourced; no K8s event source exists; platform-topology.md explicitly lists them as remaining polls |

---

## Findings

> 🟡 **platform-topology.md:L74** — `"per-process acceptance caching"` is imprecise
>
> The text added in this PR reads: _"acceptance is irrevocable, so per-process acceptance caching can only be stale in the safe direction."_ The code uses `WeakSet<ApiContext>` where one `ApiContext` is created **per WebSocket connection** (not per server process). Granularity matters: the cache survives exactly as long as the connection does, then resets on reconnect. The stale-safe claim holds at both granularities, but "per-process" will mislead a reader trying to understand the cache lifetime.
>
> **Fix:** Change "per-process acceptance caching" to "per-connection acceptance caching" in platform-topology.md.

> 🟡 **platform-topology.md:L75** — bounded skill-install poll omitted from "remaining polls" enumeration
>
> The PR adds the sentence: _"server-pushed invalidation hints replace list polling, with pod-sourced reads (session status, in-pod file listings, runtime metrics) and an agent-reachability probe the remaining polls."_ This list omits the bounded skill-install poll in `packages/ui/src/modules/experiments/hooks/use-experiment-greeting.ts` (`refetchInterval` up to `SETUP_POLL_MAX_ATTEMPTS=40` × 3 s). The PR commit message calls this out as "the bounded skill-install watch" — it should appear in the topology page alongside the other remaining polls, even if noted as bounded/terminating.
>
> **Fix:** Append "and the bounded skill-install watch (auto-stops after 40 attempts)" to the remaining-polls list in the UI section of platform-topology.md.

### Issue Fit

No scope issues. All 77 changed files fit the live-events feature or its required prerequisites. See suggestion below.

> 💡 **deploy/tasks.toml, tasks.toml, packages/e2e/playwright/tasks.toml** — e2e infra overlap with PR #3302
>
> These ~50 lines of e2e infrastructure changes also appear in companion PR #3302. The author notes they "dedupe on merge." Confirm merge order before approving either PR: if #3302 lands first these hunks conflict; if #3303 lands first #3302 may become a partial duplicate. A brief note in the PR description on merge plan would avoid ambiguity.

> 🟡 **packages/api-server/src/modules/approvals/services/approvals-service.ts:L410** — silent `catch {}` in ACP-native delivery _(pre-existing, not introduced by this PR)_
>
> ```ts
> try {
>   await deps.wrapperFrameSender.send(row.agentId, frame);
>   await deps.repo.markDelivered(row.id);
> } catch {}
> ```
>
> If `send` succeeds but `markDelivered` throws, the delivery happened but the database has no record of it — the sweeper will retry and the agent may receive a duplicate frame. If `send` fails, no log line appears. The sweeper handles re-delivery, so correctness is preserved, but the empty catch blocks all observability into delivery failures. This pre-existed this PR (introduced this round only to add `if (casWon) emitResolved(row)` above the try-block); noting it here for completeness.
>
> **Fix (out of scope for this PR):** `catch (err) { getLogger().warn({ err, agentId: row.agentId }, "approvals.acp_native_delivery_failed") }`

No other TypeScript engineering issues. Live-events service, compose, saga, router, and strip-comments are all clean.

No findings. All changes are clean:
- `useLiveEvents()` placement correct in `MainApp()`
- Empty `[]` dep array is correct (`api` is module-scoped)
- `DockedArtifactPanel` 3-way state guard and `key={openArtifactId}` are correct
- `isTermsStaleError` type guard is sound
- `ChatArtifactsPanel` null-query-when-collapsed is correct

---

## The approval bar

This PR has **no blocking findings.** The two new 🟡 documentation warnings are fixable in a follow-up commit or a separate doc PR — they do not affect runtime correctness. The empty-catch observation is pre-existing and out of scope.

All 4 round-2 critical findings are resolved. The implementation is architecturally correct:

- Terms per-connection cache is sound (deploy=restart=WS severed; HTTP path calls `markTermsProven` explicitly)
- Bounded async generator with `MAX_PENDING=256` and sync-collapse is correct
- CAS-gated `emitResolved` eliminates the pre-emit race
- Two `HarnessConfigChanged` sites are deliberately distinct (declared vs confirmed; `sameSnapshot` comparison includes `confirmed` field so the first-to-confirm write always fires the event)
- `hintFor` exhaustive switch catches unhandled `DomainEvent` variants at compile time
- Unit test suite covers generator contract, overflow, abort cleanup, owner isolation, and hint projection
- E2E smoke test (`14-live-events.spec.ts`) covers the full subscription path end-to-end

---

```json
{
  "findings": [
    {
      "id": "doc-per-connection",
      "severity": "warning",
      "file": "docs/architecture/platform-topology.md",
      "line": 74,
      "summary": "\"per-process acceptance caching\" should be \"per-connection\" — WeakSet<ApiContext> is per-connection not per-process",
      "status": "open"
    },
    {
      "id": "doc-bounded-poll-omission",
      "severity": "warning",
      "file": "docs/architecture/platform-topology.md",
      "line": 75,
      "summary": "bounded skill-install poll (use-experiment-greeting, up to 40×3s) omitted from remaining-polls enumeration",
      "status": "open"
    },
    {
      "id": "approvals-silent-catch",
      "severity": "warning",
      "file": "packages/api-server/src/modules/approvals/services/approvals-service.ts",
      "line": 410,
      "summary": "empty catch {} in resolveAndDeliverAcpNative silences all delivery failures (pre-existing)",
      "status": "open",
      "note": "pre-existing, not introduced by this PR"
    },
    {
      "id": "e2e-merge-plan",
      "severity": "suggestion",
      "file": "deploy/tasks.toml",
      "summary": "e2e infra overlap with PR #3302 needs explicit merge-order plan",
      "status": "open"
    },
    {
      "id": "doc-dual-harness-config-emit",
      "severity": "suggestion",
      "file": "docs/architecture/platform-topology.md",
      "summary": "two HarnessConfigChanged emit sites (declared + confirmed) not documented in ADR or connections.md",
      "status": "open"
    },
    {
      "id": "doc-max-pending-undocumented",
      "severity": "suggestion",
      "file": "docs/architecture/platform-topology.md",
      "summary": "MAX_PENDING=256 backpressure overflow behavior not documented in architecture pages",
      "status": "open"
    },
    {
      "id": "strip-comments-unindented-main",
      "severity": "suggestion",
      "file": "scripts/strip-comments.mjs",
      "line": 305,
      "summary": "main() body is entirely unindented (column 0); formatter-excluded so no runtime impact",
      "status": "open"
    }
  ]
}
```

jezekra1:
Follow-up to the review response above:

**E2E confirmed** — the full smoke suite passed 28/28 against a fresh cluster built from the branch head, including the dual-door auth spec (which now asserts the structured `data.termsStale` field on the WS denial), the tRPC-WS connection-auth specs, and the live-events pipeline specs (sync-first stream, an HTTP-side create surfacing as an agents hint on the socket via real Redis, agent-bound key refusal).

**One correction to the response above**: the optional `k8s` dependency is no longer defended — it was test convenience leaking into the production signature. `9ffc1ff7` makes the dependency required; the unit harness passes a no-op stub, and `startAgentWatch()` lost its silent-no-op guard. That resolves the finding as written.

packages/ui/src/modules/live-events/use-live-events.ts
```diff
@@ -0,0 +1,20 @@
...
+  isTermsStaleError,
+  onTermsStale,
+} from "../terms/lib/on-terms-stale.js";
+import { invalidateForLiveEvent } from "./invalidation.js";
+
+export function useLiveEvents(): void {
+  useEffect(() => {
+    const subscription = api.events.owner.subscribe(undefined, {
+      onData: invalidateForLiveEvent,
+      onError: (error) => {
```
PetrBulanek: `onError` handles the terms-stale case and nothing else. tRPC's `wsLink` does re-subscribe
after a *connection* loss (subscription requests survive `handleCloseOrError`), but an
error delivered *for the subscription* reaches the observer's error path and ends it — after
that this tab never re-subscribes and never gets another hint.

There is a second, quieter path to the same state. `createRedisBus.subscribe` registers the
listener in its local map and then fires `subscriber.subscribe(channel)` with
`.catch()` that only logs. If that SUBSCRIBE fails, the channel is already in the map, so no
later call retries it — `ownerStream` yields its opening `sync`, stays open forever, and
delivers nothing more.

In both cases the socket stays healthy, so the connection banner never appears, and since
this PR removes every `refetchInterval` the topics cover, there is no poll to paper over it.
`sync` only fires on (re)subscribe, so the reconnect-heals-by-refetch bound that ADR-083
claims for hints does not apply when the subscription itself is up but dead. Some liveness
signal — a re-subscribe on error, or a heartbeat that the client can time out on — would
give the stream the converging fallback the ADR requires of it.

packages/api-server/src/modules/approvals/services/delivery-sweeper.ts
```diff
@@ -43,7 +44,15 @@ export function createDeliverySweeper(
           await deps.repo.markDelivered(row.id);
         } catch {}
       }
-      await deps.repo.expireOverdue(new Date()).catch(() => {});
+      const expired = await deps.repo.expireOverdue(new Date()).catch(() => []);
+      for (const row of expired) {
+        emit({
```
PetrBulanek, line 57: This loop can never run a single iteration. `expireOverdue` compares through a raw
``sql`${pendingApprovals.expiresAt} < ${now}` `` template, and raw templates skip drizzle's
column mapper, so the `Date` reaches postgres-js unconverted. `drizzle(client)` also
overwrites the timestamptz serializers (OID 1184 among others) with an identity function, so
nothing serializes it either and the driver throws at Bind. I confirmed the second half
locally against the repo's own dependency versions:

    before drizzle(): 2026-01-02T03:04:05.000Z
    after  drizzle(): IDENTITY - Date passed through unserialized

Line 47 catches that with `.catch(() => [])`, so the sweep returns an empty array on every
tick, silently. The consequence for this PR is that the expiry hint never fires; the
consequence that predates it is that overdue approvals never actually expire.

The `resolveExpired` transitions inside `approvals-service.ts` are fine — they go through the
mapped `.set()`/`eq()` path. It is only the raw comparison here that breaks, and the fix is
contained: `lte(pendingApprovals.expiresAt, now)` instead of the raw template, keeping
`.set({ resolvedAt: now })` as it is. Worth also reconsidering whether this `.catch` should
swallow rather than log, given a hint now depends on it.

PetrBulanek (approved):

jezekra1:
Thanks @PetrBulanek — both of your findings verified and fixed.

## Overdue-approval expiry (`delivery-sweeper.ts` / `approvals-repository.ts`) — `dd7f01f5`

Reproduced exactly as you described, against the dev cluster's real postgres with the repo's own dependency versions:

- `sql\`\${expiresAt} < \${now}\`` with a `Date` param → **throws at bind** (the Date reaches the driver unserialized).
- The same comparison via `lte`/`lt` on the column, or an ISO string, → works.

The mechanism is confirmed at `drizzle-orm/postgres-js/driver.cjs:47-53`: the driver replaces the timestamptz serializer (OID 1184) with identity `(val) => val`, and a raw `sql` template skips drizzle's column mapper, so nothing serializes the `Date`. `.catch(() => [])` then swallowed the throw on every tick — so, as you noted, this is a pre-existing bug (overdue approvals never expired) that the new expiry hint surfaced.

Fix: `lt(pendingApprovals.expiresAt, now)` (keeping `<` semantics — `lt` is already imported and used with a `Date` elsewhere in the file), the sweeper's catch now logs instead of swallowing, and a regression test pins the query shape — it renders the real `expireOverdue` predicate and asserts no raw `Date` survives into the bound params. I confirmed the guard fails on the old raw template and passes on the fix.

## Open-but-dead subscription (`use-live-events.ts` + `redis-bus.ts`) — `452c7712`

Both paths were real, and since this PR removed the `refetchInterval`s the topics covered, neither had a converging fallback:

- **Client:** `onError` only handled terms-stale; any other subscription-level error ended the observer and tRPC won't re-open it. It now re-subscribes after a backoff, so a fresh `sync` heals the tab.
- **Redis bus:** the channel was registered in the listener map before the async `SUBSCRIBE`, and a rejected subscribe was never retried — leaving a mapped-but-dead channel. It now retries on a bounded backoff while listeners remain, clearing the timer on the last unsubscribe and on `close()`.

Your framing was the useful part: the ADR's "reconnect heals by refetch" bound only holds if the subscription actually re-opens, and these were the two spots where it didn't. A server heartbeat would be the stronger signal against a truly silent mid-stream stall; I left it out for now since the two targeted heals cover the reachable cases without changing the stream contract — happy to add one if you'd rather have the belt-and-suspenders.

All green: `mise run check`, api-server 938 unit tests (incl. the two new guards), UI 277.

PR #1417, opened 2026-06-19
fix(ui): increase file upload limit

jimcadden:
Increase the file upload limit through the UI from 10MB -> 50MB

Closes #1416

PR #3304, opened 2026-08-13
fix: say 50 MB in the channel tool cap copy and relay server detail in CLI file errors

PetrBulanek:
## What

Two user-facing surfaces still said the per-file attachment cap was **10 MB** when the gate is 50 MiB, and both CLI file-transfer error printers were wrong in different ways: the upload printer read a field that is always empty, and the read printer restated the cap as a literal and pointed the user at an upload-direction command.

## Scope changed — the UI half is now on `main`

This PR originally also corrected the UI upload toasts and the file-preview hint. **#3319 delivered that independently** and better: both upload toasts now render `formatBytes(MAX_UPLOAD_BYTES)`, derived from the constant and rounded, and the preview hint no longer names a number at all. Those commits were dropped in the rebase because their subject no longer exists. See the comment below.

What remains is the channel-tool copy and the CLI printers. The branch touches no UI files.

## Which value is right

50 MB. The copy was stale, not the constant.

Commit 54ecf340 (#1417) raised the real ceiling from 10 MB to 50 MB: it set agent-runtime's `MAX_FILE_SIZE` to 50 MiB ([`files.ts:25`](packages/agent-runtime/src/modules/files.ts)), raised the transport guard to 70 MB to cover base64 overhead, and updated `docs/architecture/cli.md`. It missed the user-facing strings. Channel attachments upload through `trpc.files.upload` on agent-runtime, so 50 MB is the cap that actually applies to them. One `MAX_FILE_SIZE` gates both `readFileSafe` (`files.ts:117`) and `uploadFileSafe` (`files.ts:262`) — there is a single server number here, not two.

## Changes

**Channels** — `packages/api-server/src/apps/harness-api-server/mcp-endpoint.ts`: the `send_channel_message` tool description and its `PAYLOAD_TOO_LARGE` result both said 10 MB.

**CLI** — the printers relay what the server actually said:

- `packages/cli/src/modules/shared/trpc/print.ts` — new `serverDetail`, shared by both printers.
- `packages/cli/src/modules/file/commands/put.ts` — two bugs, both pre-existing:
  - The detail was read from `data.message`, which is **always** undefined. At the locked `@trpc/server` 11.16.0 (`packages/cli/package.json` declares `^11.16.0`, so the lockfile is what holds it), `getErrorShape` puts the message at the top level and `data` carries only `code` / `httpStatus` / `path`. The `(detail)` parenthetical therefore never appeared.
  - Reading the top-level `message` instead is correct for the service-level check but wrong for `FORBIDDEN`: that route's only reason is the generic string `"forbidden path"`, so relaying it produces `error: forbidden path: forbidden path` and drops the remote path — the only identifying text the line carries. The `FORBIDDEN` branch now prints `remotePath` directly.
- `packages/cli/src/modules/file/commands/get.ts` — drops the hardcoded cap for the server's own figures, which is what `docs/architecture/cli.md:114` already says the CLI does. The same line also pointed the user at `dam import` for "bulk transfer", which copies local files *into* the agent and so cannot help a failed read. No bulk-download command exists to name instead, so the pointer is gone rather than swapped. `put.ts` keeps it, where the direction matches.

**`serverDetail` and why it exists.** A `TRPCError` raised with no message falls back to its own code name. `incomingMessageToRequest` raises exactly that for the 70 MB transport body cap, so relaying the message verbatim prints `exceeds the server's per-file cap (PAYLOAD_TOO_LARGE)`. Base64 inflates 4/3, so uploads between 50.0 and 52.5 MiB reach the service check and get real figures, while anything larger trips the message-less body cap first. `serverDetail` returns `""` when the message only echoes the code, so the parenthetical is dropped rather than filled with a code name.

No gate changes — every one already used 50 MiB. What changes is the channel-tool copy, plus the two CLI printer bugs above, which do alter emitted stderr.

## Verification

The tRPC facts above were confirmed against a live client/server round trip at the locked 11.16.0, not read off the types. Observed, driving the real branch bodies:

| case | printed |
|---|---|
| `FORBIDDEN` on `../etc/passwd` | `error: forbidden path: ../etc/passwd` |
| upload 51 MiB (service check) | `error: a.bin exceeds the server's per-file cap (file 53477376 bytes (max 52428800)). …` |
| upload 60 MiB (transport cap) | `error: b.bin exceeds the server's per-file cap. …` |

`data.message` was `undefined` in every case, confirming the old read was dead.

Re-run at this head, after the rebase onto `006c44d3`: `mise run common:check:comment-types`, `cli:check`, `cli:test` (154 passed), `api-server:check`. The four changed files are byte-identical to the pre-rebase head, so the earlier verification still applies. `ui:check` / `ui:test` are no longer relevant — the branch touches no UI.

## Known follow-up, deliberately not in this PR

The MCP message restates the cap as a literal because `api-server` cannot import agent-runtime's `MAX_FILE_SIZE`. The clean fix is to export the byte cap from a contract package beside the files schema and have the enforcer and every client derive from that one import. `packages/ui` now has the same duplication in `MAX_UPLOAD_BYTES`. That is a wider refactor than a copy fix, so it is not here — worth doing if the cap moves again.

### DIFF
--- a/packages/cli/src/modules/file/commands/get.ts
+++ b/packages/cli/src/modules/file/commands/get.ts
@@ -11,7 +11,7 @@ import {
   printResolveError,
 } from "../../agent/commands/errors.js";
 import { resolveActiveHost } from "../../shared/preflight.js";
-import { printTrpcError } from "../../shared/trpc/print.js";
+import { printTrpcError, serverDetail } from "../../shared/trpc/print.js";
 import { createAgentTrpcClient } from "../../shared/trpc/trpc-client.js";
 import {
   EXIT_BELOW_FLOOR,
@@ -172,10 +172,11 @@ function printTrpcReadError(e: unknown, path: string, host: string): void {
     return;
   }
   if (code === "PAYLOAD_TOO_LARGE") {
+    const detail = serverDetail(e);
     process.stderr.write(
-      `error: ${path} exceeds the 10 MB cap for tRPC-based reads. ` +
-        `Streaming transfer for individual large files isn't implemented yet ` +
-        `(use \`dam import\` for bulk transfer).\n`,
+      `error: ${path} exceeds the server's per-file cap` +
+        (detail ? ` (${detail})` : "") +
+        `. Streaming transfer for individual large files isn't implemented yet.\n`,
     );
     return;
   }
--- a/packages/cli/src/modules/file/commands/put.ts
+++ b/packages/cli/src/modules/file/commands/put.ts
@@ -10,7 +10,7 @@ import {
   printResolveError,
 } from "../../agent/commands/errors.js";
 import { resolveActiveHost } from "../../shared/preflight.js";
-import { printTrpcError } from "../../shared/trpc/print.js";
+import { printTrpcError, serverDetail } from "../../shared/trpc/print.js";
 import { createAgentTrpcClient } from "../../shared/trpc/trpc-client.js";
 import {
   EXIT_BELOW_FLOOR,
@@ -125,9 +125,10 @@ function printTrpcUploadError(
   remotePath: string,
   host: string,
 ): void {
-  const trpcErr = e instanceof TRPCClientError ? e : undefined;
-  const code = trpcErr?.data?.code as string | undefined;
-  const message = (trpcErr?.data?.message as string | undefined) ?? "";
+  const code =
+    e instanceof TRPCClientError
+      ? (e.data?.code as string | undefined)
+      : undefined;
   if (code === "CONFLICT") {
     process.stderr.write(
       `error: ${remotePath} already exists on the agent; pass --overwrite to clobber\n`,
@@ -135,13 +136,14 @@ function printTrpcUploadError(
     return;
   }
   if (code === "FORBIDDEN") {
-    process.stderr.write(`error: forbidden path: ${message || remotePath}\n`);
+    process.stderr.write(`error: forbidden path: ${remotePath}\n`);
     return;
   }
   if (code === "PAYLOAD_TOO_LARGE") {
+    const detail = serverDetail(e);
     process.stderr.write(
       `error: ${remotePath} exceeds the server's per-file cap` +
-        (message ? ` (${message})` : "") +
+        (detail ? ` (${detail})` : "") +
         `. Streaming transfer for individual large files isn't implemented yet ` +
         `(use \`dam import\` for bulk transfer).\n`,
     );
--- a/packages/cli/src/modules/shared/trpc/print.ts
+++ b/packages/cli/src/modules/shared/trpc/print.ts
@@ -1,9 +1,16 @@
+import { TRPCClientError } from "@trpc/client";
 import type { AuthRequiredError, TransportError } from "../errors.js";
 import { formatAuthRejection } from "../auth-message.js";
 import { classifyTrpcError } from "./classify.js";
 
 export function formatTransportError(reason: string, host: string): string {
   return `cannot reach server \`${host}\`: ${reason}`;
+}
+
+export function serverDetail(e: unknown): string {
+  if (!(e instanceof TRPCClientError)) return "";
+  const code = e.data?.code as string | undefined;
+  return e.message && e.message !== code ? e.message : "";
 }
 
 export function printServiceError(
--- a/packages/api-server/src/apps/harness-api-server/mcp-endpoint.ts
+++ b/packages/api-server/src/apps/harness-api-server/mcp-endpoint.ts
@@ -142,7 +142,7 @@ export function createMcpSession(
 
   server.tool(
     "send_channel_message",
-    `Send a message to a connected channel (slack or telegram) for this agent. Pass chatId to address a specific chat: an id from describe_channel, or on Slack a user id (U…) to send that person a direct message. Omit chatId for the default chat (Slack: the agent's bound channel; Telegram: the last-active chat). Messages are posted as the bot, attributed to this agent. Optionally attach a single file by setting attachment.path — accepts an absolute path on the agent pod (e.g. ${agentHome}/work/report.md) or a path relative to your workspace (e.g. report.md). 10 MiB cap.`,
+    `Send a message to a connected channel (slack or telegram) for this agent. Pass chatId to address a specific chat: an id from describe_channel, or on Slack a user id (U…) to send that person a direct message. Omit chatId for the default chat (Slack: the agent's bound channel; Telegram: the last-active chat). Messages are posted as the bot, attributed to this agent. Optionally attach a single file by setting attachment.path — accepts an absolute path on the agent pod (e.g. ${agentHome}/work/report.md) or a path relative to your workspace (e.g. report.md). 50 MB cap.`,
     {
       channel: z.enum([ChannelType.Slack, ChannelType.Telegram]),
       text: z.string(),
@@ -191,7 +191,7 @@ export function createMcpSession(
             }
             if (err.data?.code === "PAYLOAD_TOO_LARGE") {
               return errorResult(
-                `attachment ${attachment.path} exceeds the 10 MB per-file cap`,
+                `attachment ${attachment.path} exceeds the 50 MB per-file cap`,
               );
             }
           }


packages/ui/src/modules/files/components/file-preview-body.tsx
```diff
@@ -107,7 +107,9 @@ export function FilePreviewBody({
     return (
       <div className="py-12 text-center text-sm text-muted-foreground">
         <p>File too large to preview</p>
-        <p className="mt-1 text-[11px]">Files over 10 MB cannot be displayed</p>
+        <p className="mt-1 text-[11px]">
+          Files over {MAX_UPLOAD_MB} MB cannot be displayed
```
dam-code-guardian: 🟡 **Warning:** This copy reads `MAX_UPLOAD_MB`, the client-side *upload* pre-flight cap. The `tooLarge` state it describes is raised by the server *read* cap instead (`packages/ui/src/modules/files/api/queries.ts:119`, from `PAYLOAD_TOO_LARGE`). The constant's own comment says it "can drift up to but not past" the server value, so a legal drop of the upload cap makes this sentence state a limit that never gated the file — the same defect class this PR fixes. Reported by issue-fit, typescript-engineering, and react-ui-engineering.

**Fix:** Render every user-facing limit from the constant that gates the exact path it describes. Give the read/preview cap its own exported constant and use it here; keep the upload constant for the two upload toasts.

packages/cli/src/modules/file/commands/get.ts
```diff
@@ -179,7 +179,7 @@ function printTrpcReadError(e: unknown, path: string, host: string): void {
   }
   if (code === "PAYLOAD_TOO_LARGE") {
     process.stderr.write(
-      `error: ${path} exceeds the 10 MB cap for tRPC-based reads. ` +
+      `error: ${path} exceeds the 50 MB cap for tRPC-based reads. ` +
```
dam-code-guardian: 🟡 **Warning:** This replaces one hardcoded cap with another, so nothing ties the text to the enforcer and it will drift again on the next change. The sibling command already avoids that: `packages/cli/src/modules/file/commands/put.ts:153` prints the server-supplied `message`, which `agent-runtime/src/modules/files.ts:137` fills with the real byte figures. `docs/architecture/cli.md:114` also states the CLI does not duplicate the constant. Reported by typescript-engineering and doc-drift.

**Fix:** Surface the server's own error `message`, or interpolate a cap constant imported from the contract package. Client error copy must derive a server limit, never repeat the number.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `e76fd71`

## PR #3304: fix(ui): say 50 MB in upload copy, matching the real cap
**Author:** PetrBulanek | **Branch:** fix/upload-size-copy → main | **Changes:** +12 −8 (6 files)

### Summary
The PR corrects stale "10 MB" user-facing copy to the real 50 MB cap, and derives `MAX_UPLOAD_MB` from `MAX_UPLOAD_BYTES` so the UI copy tracks the gate. The number is right; two of the six sites still read their limit from the wrong source.

### Findings
- 🟡 **Warning:** preview copy reads the upload cap (`packages/ui/src/modules/files/components/file-preview-body.tsx:111`)
- 🟡 **Warning:** the CLI restates the cap as a literal (`packages/cli/src/modules/file/commands/get.ts:182`)
- 🟢 **Suggestion:** the same stale "10 MB" text survives on the channel-attachment error path, against the same 50 MB enforcer. The file is untouched here, so a separate issue fits better. (`packages/api-server/src/apps/harness-api-server/mcp-endpoint.ts:203`) — reported by typescript-engineering
- 🟢 **Suggestion:** a comment states `TRPC_MAX_BODY_SIZE = 32 MB` while the constant at line 189 is 70 MB. This is the same stale-number class, in an untouched file — worth a separate issue. (`packages/agent-runtime/src/server.ts:576`) — reported by issue-fit
- 🟢 **Suggestion:** `MAX_UPLOAD_MB` is an unrounded division. A future cap that is not a whole number of MB puts a long float into user text. `formatBytes` in `packages/ui/src/lib/format-size.ts` already rounds. (`packages/ui/src/modules/files/api/queries.ts:244`) — reported by react-ui-engineering
- 🟢 **Suggestion:** `docs/architecture/cli.md:114` states the CLI surfaces the server's too-large error and does not duplicate the constant. `dam file get` contradicts that; `dam file put` complies. (`docs/architecture/cli.md:114`) — reported by doc-drift
- ✅ **Looks good:** the 50 MB figure is correct, and verified independently. `MAX_FILE_SIZE` is 50 MiB at `packages/agent-runtime/src/modules/files.ts:28`, and both chat attachments and the file browser reach it through `trpc.files.read` / `files.upload`. The dropped "mirrors the file viewer's 10 MB cap" clause was also a false claim: artifact previews gate at their own `PREVIEW_MAX_BYTES` and their copy names no number.

The change is copy-only. No subsystem responsibility changed, and no architecture page asserts a cap number this diff invalidates.

- 🟢 One finding on `docs/architecture/cli.md:114` — merged into `### Findings` above.

### Issue Fit
No linked issue — the body names six files and the diff changes exactly those six. No undeclared scope creep.

- 🟡 One finding on the preview copy source — merged into `### Findings` above.
- 🟢 One finding on `packages/agent-runtime/src/server.ts:576` — merged into `### Findings` above.

- 🟡 **Warning:** the client declares its own 50 MB cap, which duplicates the real enforcer `MAX_FILE_SIZE` (`packages/agent-runtime/src/modules/files.ts:28`). This PR derives `MAX_UPLOAD_MB` from that duplicate and fans it out to three user-visible strings, so the copy now carries the drift risk the PR fixes by hand. (`packages/ui/src/modules/files/api/queries.ts:243`)
  **Fix:** export the byte cap from the contract package beside the files schema, import it in both the enforcer and the client, and derive every MB value from that one import. Never restate a server-enforced limit as a client-local literal.
- 🟡 Two further findings on `file-preview-body.tsx:111` and `get.ts:182` — merged into `### Findings` above.
- 🟢 Two findings, one on `mcp-endpoint.ts:203` (merged above) and one asking for a corrected comment cross-reference in `artifact-library-service.ts:40`, which carries no defect risk.

- 🟡 One finding on the preview copy source — merged into `### Findings` above.
- 🟢 One finding on the unrounded `MAX_UPLOAD_MB` — merged into `### Findings` above.
- One finding on the value import at `file-preview-body.tsx:5` is not carried: `file-viewer.tsx`, the only caller, already imports from `queries.js`, so the change adds no dependency.

### Verdict
COMMENT — the corrected number is right and verified, but two touched sites still source their limit from the wrong constant, so this copy can drift again.

---

PetrBulanek:
Thanks — three of these were real misses on my side, now fixed in 73e27700.

**Taken:**

- `mcp-endpoint.ts:203` — genuinely stale, and it is the message-attachment path, so squarely in scope. Fixed.
- `agent-runtime/src/server.ts:576` — the comment quoted 32 MB against a 70 MB constant. Fixed.
- While re-sweeping I found a fourth the review did not flag: `packages/api-server/src/core/blob-handoff.ts:12` illustrated its pub/sub argument with the old 10 MiB cap. Fixed. The 32 MB Redis output-buffer figure in that same comment is correct and left alone — the argument only gets stronger at 50 MiB.

My first sweep missed these because my grep patterns (`"10 MB cap"`, `"exceeds 10"`) were too literal to match `"10 MB per-file cap"`. Re-ran as a plain `10 *MB\|32 *MB` scan across all packages; the tree is now clean apart from the intentional 32 MB Redis figure.

**Not taken, with reasoning:**

- **Client duplicates the server enforcer** (`queries.ts:243`) — agreed, and it is the right fix, but it predates this PR and needs the cap exported from a contract package so `agent-runtime`, `cli`, `api-server`, and `ui` all derive from one import. That is a refactor, not a copy fix. Called out explicitly in the PR body as a follow-up rather than smuggled into a copy change.
- **`file-preview-body.tsx:111` reads the upload cap** — worth noting that `readFileSafe` and `uploadFileSafe` in `packages/agent-runtime/src/modules/files.ts` gate on the *same* `MAX_FILE_SIZE`, so there is one server number here, not two. The displayed value is correct and cannot desync. The naming is imprecise, which the contract-package move above resolves properly; renaming the client constant alone would just relabel the duplication.
- **`get.ts:182` literal** — same root cause; the number is now right, and the durable fix is the shared constant.
- **`MAX_UPLOAD_MB` is an unrounded division** — true but currently hypothetical: 50 MiB / 1 MiB is exactly 50. Rounding now would hide a non-whole cap rather than surface it. If the cap ever becomes non-whole, `formatBytes` is the right call at that point.
- **`docs/architecture/cli.md:114`** — a fair catch, and the drift is real, but it predates this PR and describes the constant-duplication problem rather than the copy. It belongs with the contract-package follow-up, since that is the change that makes the doc true again.

packages/cli/src/modules/file/commands/put.ts
```diff
@@ -133,9 +133,9 @@ function printTrpcUploadError(
 ): void {
   const trpcErr = e instanceof TRPCClientError ? e : undefined;
   const code = trpcErr?.data?.code as string | undefined;
-  // Server detail (e.g. "file 15728640 bytes (max 10485760)") lives in
-  // `data.message`; top-level `message` is just the code name.
-  const message = (trpcErr?.data?.message as string | undefined) ?? "";
+  // Server detail (e.g. "file 62914560 bytes (max 52428800)") is the tRPC
+  // error message; `data` carries only the code and HTTP status.
+  const message = trpcErr?.message ?? "";
```
dam-code-guardian: 🟡 **Warning:** this rewiring is correct for `PAYLOAD_TOO_LARGE`, but `message` is shared with the `FORBIDDEN` branch at line 148, which was written for `data.message` and used `message || remotePath` to fall back to the path. The server sends `TRPCError({code:"FORBIDDEN", message: error.reason})` and every reason here is a generic string (`"forbidden path"`, `"reserved path"` — `files.ts:189,212,229`), so the branch now prints `error: forbidden path: forbidden path` and the user loses the only path-identifying text the line carried. Verified at the pin: `getErrorShape` (`@trpc/server` v11.16.0) puts `error.message` at the top level and `data` carries only `code`/`httpStatus`/`path`, so the old read was always `""` — which is exactly why the fallback worked.
  **Fix:** When you rewire where an error detail comes from, re-check every branch that already consumed it — a shared extraction variable changes all of them at once. Give the cap branch its own binding (or make the `FORBIDDEN` branch prefer `remotePath`), so a generic reason cannot displace the identifier the message exists to name.

packages/cli/src/modules/file/commands/get.ts
```diff
@@ -169,18 +169,18 @@ async function fileExists(path: string): Promise<boolean> {
 }
 
 function printTrpcReadError(e: unknown, path: string, host: string): void {
-  const code =
-    e instanceof TRPCClientError
-      ? (e.data?.code as string | undefined)
-      : undefined;
+  const trpcErr = e instanceof TRPCClientError ? e : undefined;
+  const code = trpcErr?.data?.code as string | undefined;
+  const message = trpcErr?.message ?? "";
```
dam-code-guardian: 🟡 **Warning:** the relayed detail is the tRPC error's top-level `message`, which is right for the service-level check (`file 62914560 bytes (max 52428800)` — `files.ts:137,293`) but empty for the transport-level cap. `incomingMessageToRequest` at the `@trpc/server` v11.16.0 pin raises `new TRPCError({code:"PAYLOAD_TOO_LARGE"})` with no message, and `TRPCError`'s constructor then falls back to the code name — so the line prints `exceeds the server's per-file cap (PAYLOAD_TOO_LARGE)`. Base64 inflates 4/3, so only files between 50.0 and 52.5 MiB reach the service check; everything larger trips the message-less 70 MB body cap first, making the code-name text the common path rather than the rare one. Same shape on the upload printer. (reported by typescript-engineering)
  **Fix:** Treat a server-supplied string as untrusted before it reaches user copy: drop the parenthetical when the message equals the error code, so a message-less transport failure never surfaces its code name as an explanation. Put the rule in one helper both printers call, rather than in two hand-copied interpolations.

packages/ui/src/modules/files/components/file-preview-body.tsx
```diff
@@ -107,7 +107,9 @@ export function FilePreviewBody({
     return (
       <div className="py-12 text-center text-sm text-muted-foreground">
         <p>File too large to preview</p>
-        <p className="mt-1 text-[11px]">Files over 10 MB cannot be displayed</p>
+        <p className="mt-1 text-[11px]">
```
dam-code-guardian, line 103: 🟡 **Warning:** this copy describes the *read* limit but sources its number from `MAX_UPLOAD_BYTES`, whose own comment licenses it to drift below the server value. Full-file verification confirms both `readFileSafe` (`files.ts:129`) and `uploadFileSafe` (`files.ts:290`) gate on the same `MAX_FILE_SIZE`, so the text is accurate today — the finding is the coupling, not the value. Lowering the upload pre-flight cap would silently make this understate the real read limit. (reported by react-ui-engineering and typescript-engineering)
  **Fix:** Bind user-facing copy to the constant that governs the exact limit it describes, never to a different limit that merely holds the same value today. Where two caps must stay equal, derive both from one shared constant so they cannot drift apart.

packages/ui/src/modules/files/api/queries.ts
```diff
@@ -241,6 +241,7 @@ export function useFileDeleteMutation(agentId: string | null) {
 // hitting the wire. Server-side enforcement lives in agent-runtime and
 // surfaces as PAYLOAD_TOO_LARGE — this value can drift up to but not past it.
 export const MAX_UPLOAD_BYTES = 50 * 1024 * 1024;
+export const MAX_UPLOAD_MB = MAX_UPLOAD_BYTES / (1024 * 1024);
```
dam-code-guardian: 🟢 **Suggestion:** this reads as `50` only because the byte cap is an exact MiB multiple. Any other value renders a long fraction in the three user-visible strings it now feeds — `10 * 1000 * 1000` would print `9.5367431640625 MB`. `formatBytes` (`packages/ui/src/lib/format-size.ts`) already rounds. (reported by react-ui-engineering and typescript-engineering)

packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts
```diff
@@ -37,7 +37,7 @@ import type {
 import { renderTextKindInner } from "../viewer/renderer.js";
 
 const LIST_LIMIT = 500;
-/** In-app preview ceiling — mirrors the file viewer's 10 MB cap. */
+/** In-app preview ceiling. */
```
dam-code-guardian: 🟢 **Suggestion:** dropping the false cross-reference is right — artifact previews do gate independently. But the comment was the only record of what this 10 MB tracked, so the constant is now a bare number sitting 40 MB below the viewer copy this PR corrected, with nothing marking that as deliberate. Consider naming what actually bounds it instead. (reported by typescript-engineering)

packages/api-server/src/core/blob-handoff.ts
```diff
@@ -9,7 +9,7 @@ import type { Redis } from "ioredis";
  *
  * `JSON.stringify` turns a Buffer into `{type:"Buffer",data:[…]}` — a plain
  * object on the far side, not a Buffer — so the attachment would arrive
- * corrupt. It also inflates ~4x, and a 10 MiB attachment (the cap the MCP tool
+ * corrupt. It also inflates ~4x, and a 50 MiB attachment (the cap the MCP tool
```
dam-code-guardian: 🟢 **Suggestion:** the number is now correct, but "the cap the MCP tool enforces" names the wrong layer — the tool only reports the error; `FilesService` enforces it (`files.ts:129,290`). The pub/sub-overflow argument itself gets stronger at 50 MiB, and the 32 MB Redis figure is right. (reported by typescript-engineering)

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `e9bd603`

## PR #3304: fix(ui): say 50 MB in upload copy, matching the real cap
**Author:** PetrBulanek | **Branch:** fix/upload-size-copy → main | **Changes:** +24 −20 (10 files)

### Summary
The PR corrects stale "10 MB" copy to the real 50 MB cap and derives the UI strings from the constant. The two 🟡 of the last review are fixed at the cause; the new CLI plumbing is correct at the pin, but it feeds one branch a value that branch does not expect.

### Changes since last review
Previous HEAD: `e76fd71` (2026-08-13T10:22:06Z) — verdict COMMENT

- ✅ **Fixed:** the CLI no longer restates the cap — `get.ts` relays the server's own detail (`get.ts:181`)
- ✅ **Fixed:** the stale "10 MB" attachment error and tool description are corrected (`mcp-endpoint.ts:154,203`)
- ✅ **Fixed:** the `TRPC_MAX_BODY_SIZE` comment now says 70 MB (`server.ts:576`)
- ✅ **Fixed:** the false "mirrors the file viewer's 10 MB cap" cross-reference is gone (`artifact-library-service.ts:40`)
- 🔁 **Still present:** the client duplicates the server enforcer (`queries.ts:243`)
- 🔁 **Still present:** the preview copy sources the upload cap (`file-preview-body.tsx:111`)
- 🔁 **Still present:** `MAX_UPLOAD_MB` is an unrounded division (`queries.ts:244`)

### Findings
- 🟡 **Warning:** the `FORBIDDEN` branch regressed (`packages/cli/src/modules/file/commands/put.ts:138`)
- 🟡 **Warning:** the relayed detail is the tRPC code name on the common oversize path (`packages/cli/src/modules/file/commands/get.ts:174`) — reported by typescript-engineering
- 🟡 **Warning:** the client duplicates the server enforcer, and the contract package already exists (`packages/ui/src/modules/files/api/queries.ts:243`) — reported by typescript-engineering
- 🟡 **Warning:** the preview copy reads the upload cap, not the read cap (`packages/ui/src/modules/files/components/file-preview-body.tsx:110`) — reported by react-ui-engineering and typescript-engineering
- 🟡 **Warning:** `packages/cli/src/modules/file/commands/put.ts` and the `send_channel_message` description edit are absent from the body's "Changes" list, and the body still claims "No behavior change" although the final commit rewires two CLI error paths. (`packages/cli/src/modules/file/commands/put.ts:138`) — reported by issue-fit
  **Fix:** Re-read the body against the final diff after every added commit. Declare each distinct edit, not one per file, and retract a "no behavior change" claim as soon as a commit alters emitted output.
- 🟢 **Suggestion:** `PREVIEW_MAX_BYTES` stays 10 MB while the deleted comment was the only record of what it tracked. It is now a bare number whose divergence from the 50 MB viewer copy is invisible. State what bounds it instead (`packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts:40`) — reported by typescript-engineering
- 🟢 **Suggestion:** `MAX_UPLOAD_MB` is an unrounded division interpolated into copy. It reads as `50` only because the cap is an exact MiB multiple; `formatBytes` (`packages/ui/src/lib/format-size.ts`) already rounds. (`packages/ui/src/modules/files/api/queries.ts:244`) — reported by react-ui-engineering and typescript-engineering
- 🟢 **Suggestion:** the `blob-handoff.ts` comment now calls 50 MiB "the cap the MCP tool enforces", but the tool enforces nothing — `FilesService` does, and the tool only reports the error. The pub/sub argument itself still holds. (`packages/api-server/src/core/blob-handoff.ts:12`) — reported by typescript-engineering

✅ No findings. No architecture page asserts a cap number this diff invalidates, and no page was edited, so no `Last verified` bump is due. `docs/architecture/cli.md:114` describes the cap by contract — the diff makes that statement more true, since `get.ts` now relays the server message instead of a literal. This resolves the previous review's 🟢 on that page.

### Issue Fit
No linked issue — scope checked against the PR description only. The body-staleness findings are merged into `### Findings` above. Two further 🟢: the body's follow-up section still says "the CLI and MCP messages restate the number as literals", which the final commit made false for the CLI, and the `fix(ui)` scope prefix understates a diff spanning `cli`, `api-server`, and `agent-runtime`.

🔁 The client-duplicates-the-enforcer finding from the previous review is still present — merged into `### Findings` above, now with the observation that `agent-runtime-api` is already a dependency of all four consumers. The relayed-code-name finding and the three others are merged there too.

Both findings are merged into `### Findings` above. The preview-copy finding is a carryover; full-file verification confirms both `readFileSafe` and `uploadFileSafe` still gate on the same `MAX_FILE_SIZE`, so today's text is accurate and the finding stays 🟡 on the coupling, not on the number.

### Verdict
REQUEST_CHANGES — relaying the server's message is the right fix and the comment describing it is correct, but the same rewired variable reaches the `FORBIDDEN` branch, which now prints `forbidden path: forbidden path` instead of the path.

---

packages/ui/src/modules/files/api/queries.ts
```diff
@@ -237,10 +237,13 @@ export function useFileDeleteMutation(agentId: string | null) {
...
 }
 
-// Client-side pre-flight cap so oversized uploads fail in the UI before
-// hitting the wire. Server-side enforcement lives in agent-runtime and
-// surfaces as PAYLOAD_TOO_LARGE — this value can drift up to but not past it.
-export const MAX_UPLOAD_BYTES = 50 * 1024 * 1024;
+// agent-runtime's MAX_FILE_SIZE, mirrored. One cap gates reads and uploads
+// there and surfaces as PAYLOAD_TOO_LARGE on either route, so the upload
+// pre-flight and the viewer's "too large" copy describe the same number.
+export const MAX_FILE_BYTES = 50 * 1024 * 1024;
```
dam-code-guardian: 🟡 **Warning:** this rename lands on a name that already means a different cap. Two other `MAX_FILE_BYTES` hold 2 MiB — one exported from `packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:20`, one module-local in this same `packages/ui` package at `packages/ui/src/modules/sandboxes/hooks/use-upload-staging.ts:3`. A wrong import resolves to a 25x smaller cap and no type error reports it, because both are `number`.

The old name was repo-unique: `MAX_UPLOAD_BYTES` appeared only in this file and its two consumers on `main`. So the rename is what removed the safety, in the same commit that centralises the value.

**Fix:** Do not rename a constant onto an identifier that already names a different value in the same package. Keep a distinguishing prefix, or rename all three occurrences together so one name means one number repo-wide.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `a173693`

## PR #3304: fix(ui): say 50 MB in upload copy, matching the real cap
**Author:** PetrBulanek | **Branch:** fix/upload-size-copy → main | **Changes:** +53 −30 (11 files)

### Summary
The PR corrects stale "10 MB" copy to the real 50 MB cap and derives every UI string from one constant. The previous round's `FORBIDDEN` regression is fixed at the cause, and the new `serverDetail` helper is correct at the tRPC pin. The rename that carries the fix moves the constant onto a name two other 2 MiB caps already use.
_(Suppressed 1 finding per PR context: the client-duplicates-the-enforcer 🟡 of the last two rounds. The author declined it twice with a stated mechanism — the cap needs a contract-package export, and `agent-runtime` cannot import from `packages/ui`. The PR body names it as an explicit follow-up. A held decline is resolved, not carried.)_

### Changes since last review
Previous HEAD: `e9bd603` (2026-08-13T12:51:46Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** the `FORBIDDEN` branch no longer prints `forbidden path: forbidden path` (`put.ts:149`)
- ✅ **Fixed:** the relayed detail is no longer the bare code name — `serverDetail` filters it (`print.ts:13`)
- ✅ **Fixed:** `PREVIEW_MAX_BYTES` now states what bounds it (`artifact-library-service.ts:40`)
- ✅ **Fixed:** the `blob-handoff.ts` comment now credits agent-runtime, not the MCP tool, as the enforcer (`blob-handoff.ts:12`)
- ✅ **Fixed:** `MAX_FILE_MB` is floored, so a non-whole cap cannot read as extra headroom (`queries.ts:246`)
- 🔁 **Still present:** the PR body is stale against its own final diff (`put.ts:147`)

### Findings
- 🟡 **Warning:** the rename lands on a name that already means a different cap (`packages/ui/src/modules/files/api/queries.ts:243`) — reported by typescript-engineering
  **Fix:** Do not rename a constant onto an identifier that names a different value elsewhere in the same package. Either keep a distinguishing prefix (`MAX_UPLOAD_FILE_BYTES`) or rename all three together; the old `MAX_UPLOAD_BYTES` was repo-unique, so this rename removed that safety.
- 🟡 **Warning:** the body's "Changes" list omits `put.ts` and `print.ts`, and still claims "No behavior change" although the final commits rewrite two CLI error paths and change emitted stderr text. (`packages/cli/src/modules/file/commands/put.ts:147`) — reported by issue-fit
  **Fix:** Re-read the body against the final diff after every added commit. List each non-mechanical file the diff touches, and retract a "no behavior change" claim as soon as a commit alters emitted output.
- 🟢 **Suggestion:** the sibling `RENDER_MAX_BYTES` still documents itself as "Mirrors the in-app preview cap" while this PR declares `PREVIEW_MAX_BYTES` independent. The two 10 MiB literals now carry contradicting provenance, so a future edit to one can silently diverge. (`packages/api-server/src/modules/artifact-library/viewer/viewer-app.ts:28`) — reported by typescript-engineering
- 🟢 **Suggestion:** the shared cap constants live in the TanStack Query module, so a presentational component and a cross-domain consumer now depend on the files data-access layer for two numbers. A dependency-free constants module removes that coupling. (`packages/ui/src/modules/files/api/queries.ts:243`) — reported by react-ui-engineering

✅ No findings. No architecture page states these cap numbers, and no page was edited. `docs/architecture/cli.md:114` requires the CLI to surface the server's too-large error instead of duplicating the constant — `get.ts` and `put.ts` now do exactly that, so this round resolves the 🟢 the first review raised on that page.

### Issue Fit
No linked issue — scope checked against the PR description only. Both body-staleness findings are merged into `### Findings` above. One further 🟢: the body says `MAX_UPLOAD_MB` is derived from `MAX_UPLOAD_BYTES`, but the diff renames the export to `MAX_FILE_BYTES` and adds `MAX_FILE_MB`, so the rename of a public export is undeclared.

Both findings are merged into `### Findings` above. Full-file verification confirms the new `serverDetail` helper is correct at the `@trpc/server@11.16.0` pin: the transport body cap raises `new TRPCError({ code: "PAYLOAD_TOO_LARGE" })` with no message, and the `TRPCError` constructor falls back to `opts.code`, so filtering `message === code` is the right test. The `FORBIDDEN` fix is also verified at the cause — `uploadFileSafe` reaches that branch only through `toWritableAbs`, whose sole reason is the literal `"forbidden path"`, so the duplicated text of the previous round cannot recur.

The one finding is merged into `### Findings` above. Both routed files are otherwise correct: every removed "10 MB" literal is replaced by `MAX_FILE_MB`, derived from the same 50 MiB value `agent-runtime`'s `MAX_FILE_SIZE` enforces on the read and the upload route.

### Verdict
COMMENT — the CLI regression is fixed at the cause and the new helper is verified at the pin. The rename that carries the fix collides with two existing 2 MiB `MAX_FILE_BYTES` constants, and the body still misdeclares its own diff.

---

PetrBulanek:
### Rebased over #3308 — read this before diffing against your last review

Main gained `69b64d70 chore: strip comments from ts/js/go sources (#3308)` (closes #3010), and `CLAUDE.md` now forbids code comments outright — only `TEST_OVERVIEW:` / `TEST_SCENARIO:` survive, enforced by `mise run common:check:comment-types`. Force-pushed after the rebase, so review links to the old SHAs are dead.

**Every conflict was a comment conflict.** Main had deleted the exact comments this PR was editing. Three of the original changes are therefore gone — not reverted, but void, because their subject no longer exists:

- `packages/agent-runtime/src/server.ts` — the comment quoting `TRPC_MAX_BODY_SIZE` as 32 MB when the constant has been 70 MB since #1417.
- `packages/api-server/src/core/blob-handoff.ts` — the pub/sub-overflow comment illustrating the argument with the old 10 MiB cap. Reviewers had also flagged that it named the wrong enforcer ("the cap the MCP tool enforces" — `FilesService` enforces it, the tool only reports it). Both corrections went away with the comment.
- `packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts` — the false claim that `PREVIEW_MAX_BYTES` "mirrors the file viewer's 10 MB cap". A follow-up commit had replaced it with an accurate one naming what actually bounds the constant; under the new rule that commit could not exist, so it was dropped during the rebase.

That last one is worth flagging as a real loss, not just churn. `PREVIEW_MAX_BYTES = 10 * 1024 * 1024` now sits bare, 40 MB below the file-viewer copy this PR corrects, with nothing recording that the gap is deliberate. Artifact previews do gate independently — the number is right — but the repo no longer has a place to say so next to the code. If that matters, it belongs in `docs/architecture/`, not here.

Also dropped: the comments this PR's own review round had added — the `serverDetail` JSDoc and the two explanatory lines in `put.ts`. `scripts/strip-comments.mjs --write` removed them in `ef154f51`. The reasoning they carried is in the PR description instead.

**Nothing behavioral changed in the rebase.** Every conflict was resolved by taking main's comment-free side and keeping this branch's code and string changes. `mergeable` is green, and the full check and test set was re-run after the strip.

One consequence for reviewers: the diff is now smaller than the PR description used to imply, and it no longer touches `agent-runtime` or `blob-handoff.ts` at all. If you reviewed the earlier version, the surviving substance is the CLI printer fixes and the UI constant — the rest was comment corrections that #3308 made moot.

packages/cli/src/modules/shared/trpc/print.ts
```diff
@@ -6,6 +7,12 @@ export function formatTransportError(reason: string, host: string): string {
   return `cannot reach server \`${host}\`: ${reason}`;
 }
 
+export function serverDetail(e: unknown): string {
```
dam-code-guardian, line 10: 🟡 **Warning:** commit `ef154f5` stripped the two comments that recorded why this code is correct, and no test replaces them. `serverDetail`s `message !== code` filter depends on `@trpc/server` 11.16.0 resolving a message-less error to its code name, and the `FORBIDDEN` branch in `put.ts:139` depends on `uploadFileSafe` raising only the generic `"forbidden path"` (`packages/agent-runtime/src/modules/files.ts:250`). Both are non-local invariants a reader cannot see, and `packages/cli/src/__tests__/` covers neither.

**Fix:** When a no-comments rule removes the rationale for a cross-package invariant, encode that rationale as a test rather than losing it. Add unit tests for `serverDetail` (empty string when the message equals the code, the message otherwise) and for `printTrpcUploadError`s `FORBIDDEN` and `PAYLOAD_TOO_LARGE` branches, so a tRPC bump or a new server-side forbidden reason fails loudly.

packages/ui/src/modules/files/api/queries.ts
```diff
@@ -210,7 +210,8 @@ export function useFileDeleteMutation(agentId: string | null) {
   });
 }
 
-export const MAX_UPLOAD_BYTES = 50 * 1024 * 1024;
+export const MAX_FILE_BYTES = 50 * 1024 * 1024;
```
dam-code-guardian: 🟡 **Warning:** the rename lands `MAX_FILE_BYTES` on a name two other caps already use, both 2 MiB: `packages/ui/src/modules/sandboxes/hooks/use-upload-staging.ts:3` (module-local, same package) and `packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:20` (exported). Two live constants now share one name and differ 25-fold, and two `number` constants never produce a type error.

**Fix:** Do not rename a constant onto an identifier that names a different value elsewhere. Qualify each cap by the surface it guards — for example `MAX_WORKSPACE_FILE_BYTES` (50 MiB, files), `MAX_STAGED_SKILL_BYTES` (2 MiB, sandboxes), `MAX_SKILL_FILE_BYTES` (2 MiB, skills) — or rename all three together. The old `MAX_UPLOAD_BYTES` was repo-unique, so this rename is what removed that safety.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `ef154f5`

## PR #3304: fix(ui): say 50 MB in upload copy, matching the real cap
**Author:** PetrBulanek | **Branch:** fix/upload-size-copy → main | **Changes:** +35 −20 (8 files)

### Summary
The PR corrects stale "10 MB" copy to the real cap and derives every string from one constant. The previous round's body staleness is fixed, and the newest commit only strips comments. That strip removes the two records of why the CLI printers are correct, and the rename collision from the last round is still open.
_(Suppressed 1 finding per PR context: the client-duplicates-the-enforcer 🟡, reported again this round by react-ui-engineering and typescript-engineering. The author declined it three times with an unchanged mechanism, and the body names it as an explicit follow-up. A held decline is resolved, not carried.)_

### Changes since last review
Previous HEAD: `a173693` (2026-08-13T14:49:23Z) — verdict COMMENT

- ✅ **Fixed:** the PR body now lists exactly the eight files the diff touches, and scopes its "No behavior change" claim to gates (`put.ts:147`)
- 🔁 **Still present:** the rename lands on a name two other 2 MiB caps already use (`queries.ts:213`)

### Findings
- 🟡 **Warning:** the comment strip removed the only record of two cross-package invariants, and no test replaces them (`packages/cli/src/modules/shared/trpc/print.ts:10`) — reported by typescript-engineering
- 🟡 **Warning:** the rename lands `MAX_FILE_BYTES` on a name two other 2 MiB caps already use (`packages/ui/src/modules/files/api/queries.ts:213`) — reported by typescript-engineering and react-ui-engineering
- 🟢 **Suggestion:** the read error tells the user to run `dam import`, which moves files in the opposite direction. The pointer is pre-existing on `main`, so it is not a regression, but this PR rewrote the sentence around it (`packages/cli/src/modules/file/commands/get.ts:180`) — reported by typescript-engineering
- 🟢 **Suggestion:** the cap constants live in the TanStack Query module, so a presentational component loads the files data-access layer for one number (`packages/ui/src/modules/files/api/queries.ts:213`) — reported by react-ui-engineering
- 🟢 **Suggestion:** the body's "What" section says both CLI printers read an always-empty field, but only `put.ts` did — `get.ts` hardcoded its cap instead (`packages/cli/src/modules/file/commands/get.ts:175`) — reported by issue-fit

✅ No findings. No architecture page states a numeric per-file cap, and no page was edited, so no `Last verified` bump is due.

One reported finding is not carried: it read `cli.md:80` ("the CLI surfaces the server's reason for authorization failures") as contradicted by the `FORBIDDEN` branch now printing only the path. Verification refutes it — `uploadFileSafe` reaches `FORBIDDEN` only through `toWritableAbs`, whose sole reason is the generic literal `"forbidden path"` (`files.ts:250,260`). The one informative reason ("the destination folder is not empty") lives in `renameSafe`, which no CLI command calls. This was last round's requested fix, re-reported as a defect.

### Issue Fit
No linked issue — scope checked against the PR description. Both staleness items from prior rounds are resolved at this HEAD. The remaining 🟢 is merged above.

Two findings are merged above. Full-file verification confirms the strip removed no functional or pragma directive, and re-confirms `serverDetail` at the `@trpc/server@11.16.0` pin: the constructor resolves `opts.message ?? cause?.message ?? opts.code`, and `incomingMessageToRequest` raises `new TRPCError({ code: "PAYLOAD_TOO_LARGE" })` with no message, so filtering `message === code` is the right test.

One finding is not carried: that `MAX_FILE_MB` divides by 1024² and labels the result "MB", so the copy still states a wrong number. The unit imprecision is real — the cap is 50 MiB, or 52.43 MB — but `main` already said "10 MB" for a mebibyte-derived value, so the PR does not introduce it, and the 2.4 MB gap misstates the limit in the conservative direction. It does not clear the bar for a blocking finding on a copy fix.

One finding is merged above; one is suppressed as a held decline. The displayed number is verified accurate: `MAX_FILE_SIZE` (`agent-runtime/src/modules/files.ts:25`) gates both `readFileSafe:117` and `uploadFileSafe:262` at 50 MiB, the transport cap is 70 MiB so it does not bind first, and the api-server tRPC proxy adds no smaller cap. So the coupling is a drift risk, not a wrong value.

### Verdict
COMMENT — the numbers are correct and verified end to end, and the body now matches its diff. The comment strip left two cross-package invariants unrecorded and untested, and the rename collision is unresolved.

---

PetrBulanek:
### Round 4 findings — three taken, two declined

Head is now `726afac0`.

**Taken:**

- **`dam import` in the read error** (`get.ts:180`, 🟢) — fixed in `5648dd68`. Confirmed the direction: `dam import` is "Import local files or folders into an Agent", so it cannot help a `file get` that failed. There is no bulk-download command in the CLI to name instead, so the pointer is deleted rather than swapped. `put.ts` keeps it, where the direction matches.

- **The `MAX_FILE_BYTES` rename** (`queries.ts:213`, 🟡) — fixed in `726afac0` as `MAX_WORKSPACE_FILE_BYTES` / `MAX_WORKSPACE_FILE_MB`, unique repo-wide again. The two 2 MiB constants are left alone; renaming them is outside a copy fix.

  One correction to the finding, since it matters for how the class is judged next time. The stated failure — "a wrong import resolves to a 25× smaller cap" — is not reachable. `packages/ui` depends on `agent-runtime-api`, not `agent-runtime`, so the skill repository's export cannot be imported here at all; and the `sandboxes` one is a module-local `const`, never exported. What is real is the same-package ambiguity: one identifier meant 50 MiB in `modules/files` and 2 MiB in `modules/sandboxes`, and a future auto-import in a `sandboxes` file could silently pick the wrong one. That is enough to justify the rename, and it was cheap here because this PR already renames the export.

- **The body's dead-field claim** (🟢) — fixed. The "What" section now says only the upload printer read `data.message`; the read printer restated the cap as a literal instead. Also corrected "the 11.16.0 pin" to "the locked 11.16.0" throughout: `packages/cli/package.json` declares `^11.16.0`, so the lockfile holds it, not the manifest.

**Not taken, with reasoning:**

- **Unit tests for the stripped invariants** (`print.ts:10`, 🟡) — the facts are right. No test covers `serverDetail` or either printer, and both invariants are non-local. Declining on blast radius: if either drifts, the whole consequence is wrong text in one stderr line. No gate moves, no data path changes, no exit code changes. That does not earn two new test files in a repo that is trying to carry fewer of them.

  Both invariants were re-verified by reading, not assumed. `uploadFileSafe` reaches `Forbidden` only through `toWritableAbs`, whose sole reason is the literal `"forbidden path"` (`files.ts:259`). The one informative reason, `"reserved path"`, is raised by `readFileSafe` — and `printTrpcReadError` has no `FORBIDDEN` branch, so it falls through to `printServiceError`, which relays the server's reason verbatim. The read path therefore never lost anything, and the write path has nothing to lose.

- **Cap constants hosted in the data-access module** (`queries.ts:213`, 🟢) — the location is pre-existing; `MAX_UPLOAD_BYTES` already lived there and this PR renamed it in place. Moving the constants to a dependency-free module is the right shape, but it is the same change as the contract-package export the body already names as a follow-up. Doing half of it here would leave two homes for one number.

Checks re-run at this head: `common:check:comment-types`, `cli:check`, `ui:check`, `cli:test` (154 passed), `ui:test` (277 passed).

One process note unrelated to the code: `reviewDecision` is still `CHANGES_REQUESTED` from the review of `e9bd603`, whose findings are fixed. The branch stays blocked until that review is dismissed or superseded by an approval.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `726afac0`

## PR #3304: fix(ui): say 50 MB in upload copy, matching the real cap
**Author:** PetrBulanek | **Branch:** fix/upload-size-copy → main | **Changes:** +40 −21 (8 files)

### Summary
The PR corrects stale "10 MB" copy to the real 50 MB cap across the UI and CLI, and derives every user-visible string from one constant. The rename collision from round 3 is resolved at this HEAD; two 🟡 findings from prior rounds remain.
_(Suppressed 1 finding per PR context: the client-duplicates-the-enforcer 🟡, now in its fifth round. The author declined it each time with an unchanged mechanism — a contract-package export is needed, and agent-runtime cannot import from packages/ui. The PR body names it as a follow-up. A held decline is resolved, not carried.)_

### Changes since last review
Previous HEAD: `ef154f5` (2026-08-13T15:45:41Z) — verdict COMMENT

- ✅ **Fixed:** rename collision resolved — `MAX_WORKSPACE_FILE_BYTES` and `MAX_WORKSPACE_FILE_MB` are now repo-unique (`queries.ts:213`)
- ✅ **Fixed:** `get.ts:180` no longer points the user at `dam import` for a failed read
- 🔁 **Still present:** `print.ts` cross-package invariants unrecorded and untested (`print.ts:10`)
- 🔁 **Still present:** `file-preview-body.tsx` sources the read-side limit from the upload-named constant (`file-preview-body.tsx:104`)

### Findings
- 🟡 **Warning:** `serverDetail`'s `message !== code` filter and the `FORBIDDEN` branch in `put.ts:139` both depend on non-obvious cross-package behavior: the first relies on `@trpc/server@^11.16.0` constructing a message-less error as its code string; the second relies on `toWritableAbs` returning only the literal `"forbidden path"`. Both invariants were stripped by the #3308 rebase and no test encodes them. (`packages/cli/src/modules/shared/trpc/print.ts:10`)
  **Fix:** Encode each invariant as a unit test: one asserting `serverDetail` returns `""` when `e.message === e.data?.code`, and one asserting `printTrpcUploadError` emits `remotePath` (not an empty message) for a synthetic `FORBIDDEN` error.
- 🟡 **Warning:** `file-preview-body.tsx` displays its limit from `MAX_WORKSPACE_FILE_MB`, which is derived from the upload-pre-flight constant, not the server read cap. Both gate on the same `MAX_FILE_SIZE` today, but a future change to either cap will silently stale the copy without a type error. (`packages/ui/src/modules/files/components/file-preview-body.tsx:104`)
  **Fix:** Export a dedicated `MAX_WORKSPACE_READ_MB` (or expose the shared cap from the contract package) so the read-side display tracks the read-side enforcer.
- 🟢 **Suggestion:** `chat-input.tsx:20` now imports two symbols from `../../files/api/queries.js`, an internal path of the files module, widening a pre-existing boundary violation. The files module has no `index.ts`. Separate issue. (`packages/ui/src/modules/sessions/components/chat-input.tsx:20`) — reported by typescript-engineering
- 🟢 **Suggestion:** `AttachmentChip` is defined inline in `chat-input.tsx` and the attachments list uses `key={i}` (index-as-key). Both are pre-existing in untouched code — separate issue. (`packages/ui/src/modules/sessions/components/chat-input.tsx:170`) — reported by react-ui-engineering

✅ No findings. No architecture page states a numeric per-file cap, and no page was edited, so no `Last verified` bump is due. `docs/architecture/cli.md:114` describes the CLI as surfacing the server's error — the diff makes that more true.

### Issue Fit
✅ No findings. All eight deliverables stated in the PR body are present in the diff. No undeclared scope.

One finding (the cross-module import boundary widening) is merged into `### Findings` above as a 🟢 — the violation is pre-existing and no `index.ts` boundary is enforced in the repo today.

Two findings (inline `AttachmentChip`, index-keyed list) are merged into `### Findings` above as a single 🟢 — both are pre-existing in untouched code.

### Verdict
COMMENT — the numbers are correct end-to-end and the rename collision is resolved. Two 🟡 findings remain: the stripped cross-package invariants in `print.ts` have no test replacement, and `file-preview-body.tsx` sources its limit from the upload constant rather than a read-specific one.

---

packages/cli/src/modules/shared/trpc/print.ts
dam-code-guardian, line 10: 🟡 **Warning:** `serverDetail`'s `message !== code` filter depends on `@trpc/server@^11.16.0` resolving a message-less error to its code string, and the `FORBIDDEN` branch in `put.ts:139` depends on `toWritableAbs` having only one reason literal. The #3308 rebase stripped both explanatory comments and no test encodes either invariant.
**Fix:** Encode each invariant as a unit test: one asserting `serverDetail` returns `""` when `e.message === e.data?.code`, and one asserting `printTrpcUploadError` emits `remotePath` (not an empty message) for a synthetic `FORBIDDEN` error.

packages/ui/src/modules/files/components/file-preview-body.tsx
```diff
@@ -100,7 +100,9 @@ export function FilePreviewBody({
...
+          Files over {MAX_WORKSPACE_FILE_MB} MB cannot be displayed
```
dam-code-guardian: 🟡 **Warning:** `MAX_WORKSPACE_FILE_MB` comes from the upload-pre-flight constant, not the server read cap. Both are 50 MiB today (both `readFileSafe` and `uploadFileSafe` gate on `MAX_FILE_SIZE`), so the displayed number is correct — but if the caps ever diverge, this copy silently trails with no type error.
**Fix:** Export a dedicated read-cap constant (or expose the shared cap from the contract package) and derive the display from it.

PetrBulanek:
### Rebased onto `006c44d3` — #3319 superseded the UI half

Head is now `885d7e71`. Force-pushed, so links to the old SHAs are dead.

**Every conflict was a UI conflict, and main won all of them.** #3319 (`feat(channels): hand the agent documents attached in Slack, not just images`) rewrote the same four UI sites this PR was correcting, and did it better:

- `chat-input.tsx:52` and `use-file-mutations.ts:257` now render `${formatBytes(MAX_UPLOAD_BYTES)}` — derived from the constant, and rounded. That is exactly the 🟢 this review raised in round 1 against `MAX_UPLOAD_MB` being an unrounded division. `formatBytes(52428800)` gives `50.0 MB`.
- `file-preview-body.tsx:102` no longer names a number at all: *"File too large to preview / Too large to load in the browser — open it in a terminal instead."*

So three of this PR's four UI edits are void — not reverted, but pointless, because the strings they targeted no longer exist. The fourth, the `MAX_WORKSPACE_FILE_BYTES` rename from round 4, went with them: it existed only to undo a collision this PR itself created by renaming `MAX_UPLOAD_BYTES` → `MAX_FILE_BYTES`. Main keeps `MAX_UPLOAD_BYTES`, which round 3 confirmed is repo-unique. Nothing left to fix.

**The branch now touches no UI files.** The four changed files are byte-identical to `726afac0`, verified per-file — the rebase dropped work, it did not alter any.

History squashed to two commits matching what actually ships: the channel-tool copy, and the CLI printers. The old six had three messages describing UI changes they no longer contained.

---

### Round 5 findings

**`file-preview-body.tsx:104` (🟡) — moot at this head.** The finding was that `MAX_WORKSPACE_FILE_MB` is "derived from the upload-pre-flight constant, not the server read cap". Two things: the premise was already stale at `726afac0` — after the round-4 rename the constant was neutral, and one `MAX_FILE_SIZE` gates both `readFileSafe` (`files.ts:117`) and `uploadFileSafe` (`files.ts:262`), so there was one server number, not two. And the proposed fix — a separate `MAX_WORKSPACE_READ_MB` — would have put two client constants against one server constant, which is the drift risk, not the cure. Either way it is gone: main's line names no number.

**`print.ts:10` (🟡) — declined, third time, unchanged reasoning.** The facts are right: no test covers `serverDetail` or either printer, and both invariants are non-local. Declining on blast radius. If either drifts, the whole consequence is wrong text in one stderr line — no gate moves, no data path changes, no exit code changes. That does not earn two new test files in a repo that is trying to carry fewer of them. Both invariants were re-verified by reading at this head, not assumed: `uploadFileSafe` reaches `FORBIDDEN` only through `toWritableAbs`, whose sole reason is the literal `"forbidden path"`; the one informative reason, `"reserved path"`, is raised by `readFileSafe`, and `printTrpcReadError` has no `FORBIDDEN` branch, so it falls through to `printServiceError`, which relays the server's reason verbatim. The read path never lost anything and the write path has nothing to lose.

The two 🟢 on `chat-input.tsx` (the internal-path import, the inline `AttachmentChip` and index-as-key) are pre-existing and the review already scoped them to a separate issue. The branch no longer touches that file at all.

Checks re-run at this head: `common:check:comment-types`, `cli:check`, `cli:test` (154 passed), `api-server:check`. All green.

One process note: `reviewDecision` is still `CHANGES_REQUESTED` from the review of `e9bd603`, whose findings were fixed three rounds ago. The branch stays blocked until that review is dismissed or superseded by an approval.

dam-code-guardian (approved):
## PR #3304: fix: say 50 MB in the channel tool cap copy and relay server detail in CLI file errors
**Author:** PetrBulanek | **Branch:** fix/upload-size-copy → main | **Changes:** +22 −12 (4 files)

### Summary
Corrects the upload-cap copy in the MCP endpoint from "10 MiB" / "10 MB" to "50 MB". Adds a `serverDetail` helper that extracts the server-provided error message from a `TRPCClientError` when it differs from the error code, and routes it into the CLI `put` and `get` PAYLOAD_TOO_LARGE paths.

### Changes since last review
Previous HEAD: 726afac0 (2026-08-14T09:58:21Z) — verdict COMMENT

- ✅ **Fixed:** UI toast and preview files dropped — superseded by #3319, which corrected all UI strings better.
- ✅ **Fixed:** `get.ts` PAYLOAD_TOO_LARGE no longer points to `dam import`; uses `serverDetail` instead.
- 🔁 **Still present:** `print.ts` cross-package invariants not recorded/tested (`print.ts:10`)
- 🔁 **Still present:** `serverDetail` reads `e.message` which falls back to `opts.code` in `@trpc/server@11.16.0`, so the non-empty-message guard `e.message !== code` is required and is present.

_(Suppressed 2 finding(s) per PR context: print.ts cross-package invariants — held decline (declined 5 times, identical reasoning); client-duplicates-enforcer in queries.ts — held decline (declined 5+ times, identical reasoning).)_

### Findings

### doc-drift
✅ No findings.

### issue-fit

✅ Delivers the PR description's ask; no undeclared scope.

### typescript-engineering
✅ No findings.

### Verdict
APPROVE — all prior blocking findings are fixed, moot, or validly declined to resolution; all three skills clean.

```json
{"review":"6","sha":"885d7e7166b2544153cd4bb18f3335259d750ea2","findings":[]}
```

Issue #3281, opened 2026-08-12
Two-phase writes for Postgres + Kubernetes operations — a row goes active only after its cluster side effect lands


jezekra1:
## Context

The source of truth for several agent settings lives in Postgres, and each change must also take effect in the cluster — a custom-resource field or a Secret. Today the database row is committed first and the cluster change is attempted right after, best-effort. When that cluster change fails — a conflict that exhausts retries, or the process dying between the two writes — the row says one thing and the cluster enforces another, with no transaction spanning the two. We just hit this with egress rules (a narrowed rule whose cluster projection failed still read as active while only partially enforced) and patched it with a per-feature reconcile, but the same shape recurs across features.

## Problem

A Postgres write and a Kubernetes write can't share a transaction, so any "commit the row, then change the cluster" sequence has a failure window where the two disagree. Because the row is marked active the instant it commits, a reader — and the user — can see a setting as in-effect before, or even when it never, actually took effect in the cluster. Each feature that does this invents its own recovery (a reconcile loop, a retry timer, or nothing), so the guarantee is inconsistent and easy to get wrong. It affects at least three surfaces today: egress rules' L7 host promotion, connection grants, and secret assignments — anywhere a database commit is followed by a cluster mutation.

## Goal

One shared way to perform a Postgres + Kubernetes change such that a row only becomes visible/active once its cluster side effect has actually landed, and a failed or interrupted change converges or is retried by construction rather than by per-feature bespoke code. A reader never sees a setting as active while the cluster hasn't applied it.

## Proposed approach

From the requester: write the intent and the data to Postgres first as a **pending** row (not yet active), perform the cluster operation, then record success or failure back to Postgres — the row flips to **active** only on success. A background reconciler drives pending/failed rows to completion, covering the crash-between-writes case that no in-request retry can. This mirrors the transactional-outbox pattern the runtime-delivery path already uses for agent-runtime deliveries, extended to cluster (custom-resource and Secret) mutations.

Open design questions: whether the write is one generic mechanism or a shared library each feature adopts; whether the API call blocks on the cluster op or returns "pending" immediately; and how readers filter to active-only without leaking pending state.

## Done when

- There is one documented, reusable mechanism for Postgres + Kubernetes dual-writes with pending→active semantics and a reconciler, not a per-feature reinvention.
- A row is surfaced as active only after its cluster side effect is confirmed; a failed or interrupted change is converged or retried without manual intervention.
[...]

PR #3305, opened 2026-08-13
feat(ui): keep a message draft with the session it was written for

### LINKED ISSUE #3254: Keep a message draft with the session it was written for
## Context

A sandbox holds many sessions, and the chat has a single message box that all of them share. We just closed the *sent* half of this problem (#2963): a first prompt can no longer land in the wrong session when you switch away while it's being delivered. The unsent half is untouched — a **message draft** (text and attachments you've typed and staged but not sent) belongs to no session at all, so it follows you around the sandbox.

## Problem

Whatever is in the message box stays there when you move elsewhere in the same sandbox:

- **Switching sessions** — the half-written message is still sitting in the box, with nothing to say it was written for the session you just left. On desktop a single Enter sends it to the session now on screen.
- **Starting a new session** — the text comes along, so an unfinished follow-up can become the opening message of a brand-new conversation.
- **Staged files and images** — they ride along the same way.

A message delivered to the wrong session isn't just an awkward moment. The agent there acts on it — runs tools, edits files, spends tokens — and the only way out is to interrupt it and undo what it did.

At the same time the box is throwaway everywhere people expect it to hold. What you typed is gone when the agent asks for an approval (the approval prompt takes the box's place), when you open a terminal session in the same sandbox, when you leave the sandbox, and on any page reload. The draft sticks in the one place it should be scoped, and vanishes in every place it should be kept.

## Goal

A message draft — text *and* its attachments — belongs to the session it was written for. Opening a session puts that session's own draft back in front of you and nothing else, the session list shows which sessions still hold one, and closing the browser doesn't throw them away.

## User Stories

- As someone juggling several sessions in a sandbox, I want each session's draft to stay with that session, so I can't send it to the wrong conversation.
- As someone writing a long prompt with files attached, I want to check another session, answer an approval, or open a terminal and come back to my text and my attachments as I left them.
- As someone starting a new session, I want the opening message I began there to stay with that new session, so switching away and back neither loses it nor drags it into an existing conversation.
- As someone scanning the session list, I want a marker on the sessions that still hold a draft, so I can see at a glance what I left unfinished.
- As someone who closes the browser mid-thought, I want my drafts still there when I reopen it on the same browser, so I don't have to retype them.

## Scope

In:

- Chat sessions in a sandbox, plus the not-yet-created session you're about to start.
- One draft per session, kept on the browser it was written on. The draft's **text** survives reloads and a browser close. Drafts are kept indefinitely — until sent, cleared, or the session is deleted.
- **Attachments travel with the draft within the tab.** They follow the draft across session switches, approvals, and terminal visits. They do not survive a reload or a browser close: opening the session afterwards drops them, names them in a plain notice toast, and restores the rest of the draft normally. Never an error state, never a blocked send.
- **The marker** goes where the per-session status dot already lives, and only shows on an idle session — a session that is working or waiting on your approval keeps showing that instead.
- Sending clears the draft; deleting a session takes its draft with it.

Out:

- Following you to another browser or device — nothing is stored server-side.
- Terminal sessions (they have no message box).
- Anything beyond one draft per session: no history, no named drafts, no revisions.

## Proposed solution

Bind the message box's contents to the session identity instead of to the chat screen, and keep them on the device so a browser close is survivable — the ask is convenience, not a shared artifact, so there's no reason to involve the server. The session list already carries per-session markers (unread, needs approval, working); the draft marker joins them as the quietest of the set.

## Additional resources

- #2963 — the sent-message half of the same problem, fixed by #3221.

PetrBulanek:
A sandbox's chat has one message box shared by all its sessions, so a half-written message — text and staged files — follows the user around: switching sessions can send it to the wrong conversation (where the agent acts on it), starting a new session drags it along as an opening message, and an approval prompt, a terminal visit, or a reload throws it away entirely.

This feature binds the draft to the session it was written for. Opening a session brings back that session's own draft and nothing else; drafts survive approvals, terminal visits, and (for text) reloads and browser closes on the same browser; the session list marks idle sessions that still hold one; sending clears it, and deleting the session deletes it. Nothing is stored server-side.

## Beyond the issue's scope

#3254 scopes persistence to "one draft per session, kept on the browser it was written on" and says nothing about tabs. Because the store is the browser's, **every tab of that browser shares the same drafts**, and this PR makes that sharing explicit rather than accidental:

- Each draft is a separate `localStorage` entry, so a tab only ever rewrites the draft it is editing. No tab can revive a draft another tab has already sent.
- A draft that appears, changes, or disappears elsewhere is applied per key — except for the composer the person in this tab is typing in, which is never overwritten from another tab.

Declared here because it is behaviour the issue did not ask for.

Closes #3254

## Sub-issues

- [x] 01 — Session-keyed drafts in memory
- [x] 02 — Text persistence and restore notice
- [x] 03 — Draft marker in the session list


dam-code-guardian (commented):
## Code Guardian Review

**PR #3305 · `feat(ui): keep a message draft with the session it was written for`**

Scope: full first review of a 9-file, 400-line change introducing session-keyed draft persistence (in-memory, localStorage, cross-tab sync, draft indicator in session list).

### Summary

The feature is fully implemented and issue-complete (all three sub-issues of #3254 are delivered). The architecture is sound — `draftKey`, `EMPTY_DRAFT`, `updateDrafts`, `migrateDraft`, and `pruneDrafts` are well-designed primitives with correct lifecycle handling. No correctness defects were found. Three architectural convention violations need addressing before merge.

### Approval bar

The review posts as COMMENT (non-blocking). Items marked 🟡 carry the specific fix that resolves each one.

---

### 🟡 Findings

**1 · `chat-input.tsx:43–61` — inline `useStore` selectors violate CRITICAL `state-management.md` rule**

Five new `useStore((s) => …)` calls are added inline in `ChatInput`. The `state-management.md` reference states `[CRITICAL] Every slice exports selector hooks. Components never call useStore(s => ...) inline.` The same pattern pre-existed in `sessions-sidebar.tsx` but the PR extends it to a second component, compounding the drift.

**Fix:** Export named selector hooks from the sessions slice:

```ts
// In sessions.ts or a new selectors.ts
export const useSelectedAgent = () => useStore((s) => s.selectedAgent);
export const useSessionId = () => useStore((s) => s.sessionId);
export const useDraftActions = () =>
  useStore(useShallow((s) => ({
    setDraft: s.setDraft,
    clearDraft: s.clearDraft,
    consumeDroppedAttachments: s.consumeDroppedAttachments,
  })));
```

Replace inline `useStore((s) => …)` calls in `chat-input.tsx` with those hooks. The `useDraftActions` selector specifically requires `useShallow` to avoid spurious re-renders on every store update.

**2 · `sessions.ts:88` — `window.addEventListener` side effect inside `StateCreator` factory body**

```ts
// Line 88 — runs at store creation time, outside the returned state object:
onForeignDraftChange((foreign) => get().mergeForeignDrafts(foreign));
```

The `StateCreator` contract is to return initial state and action implementations. Side effects with external subscriptions belong at the composition root or in a dedicated teardown-returning init call. As written, every hot-module reload or store recreation in tests leaks an additional `storage` listener.

**Fix:** Move this call out of `createSessionsSlice`. Either:
- Add it to `store.ts` after `create(…)`: `onForeignDraftChange((f) => useStore.getState().mergeForeignDrafts(f));`
- Or expose `initDraftSync(): () => void` as a returned action and invoke it once at the composition root with the teardown wired into the app's unmount.

**3 · `draft-snapshot.ts:24–27` — `localStorage` referenced at module evaluation time without an environment guard**

```ts
// Module-level constant — evaluated at import time:
const localStore: DraftStore = {
  getItem: (key) => localStorage.getItem(key),
  setItem: (key, value) => localStorage.setItem(key, value),
};
```

`onForeignDraftChange` on line 97 already guards with `typeof window === "undefined"`, but `localStore`'s closures reference `localStorage` directly. If this module is ever imported in a non-browser context (SSR, a test runner without jsdom) it throws at parse time.

**Fix:**
```ts
function getLocalStore(): DraftStore {
  return {
    getItem: (k) => localStorage.getItem(k),
    setItem: (k, v) => localStorage.setItem(k, v),
  };
}
// Use as default argument: store: DraftStore = getLocalStore()
```
Or inline the guard in each closure body.

**4 · `docs/architecture/platform-topology.md` — `ui` subsection does not document localStorage draft persistence**

The PR introduces session-keyed text persistence backed by browser localStorage. The `ui` section of `platform-topology.md` documents the session model (ACP multiplexing, URL binding, reload behaviour) in detail but says nothing about client-side persistent state. localStorage is a new persistence substrate for the `ui` component.

**Fix:** In the `ui` subsystem section, add one sentence: the UI maintains session-keyed draft state (text and staged attachment names) in `localStorage`, surviving page reloads and browser closes, migrated to a live session key on session start, and cleared on send or session deletion. Bump `Last verified`.

---

### 🔵 Notes (no action required)

**`chat-input.tsx:101–103` — `state.setDraft` vs captured `setDraft` in async `FileReader.onload`**

`useStore.getState()` is correctly used in the async callback to avoid stale closures. However, the call reads `state.setDraft` off the snapshot object rather than using the captured `setDraft` from the outer closure (line 45). Both references are the same function in Zustand, so there is no bug — but `removeAttachment` and `send` use the captured reference while `addFiles` uses the snapshot method. Prefer consistency: use `setDraft(key, …)` here to match every other write path.

---

### ✅ What is correct

- `draftKey` / `EMPTY_DRAFT` / `SessionDraft` — clean domain primitives, pure and side-effect-free.
- `updateDrafts` recipe pattern — correct shallow-copy-then-mutate Zustand pattern; `saveDraftSnapshot` always called together with `set()`.
- `migrateDraft` (`~new` → session ID on connect) — correct TOCTOU fix; `agentId` captured before async gap in `use-acp-connection.ts`.
- `pruneDrafts` keep-set logic — correctly keeps `~new`, current `sessionId`, and all live sessions for the agent.
- `mergeForeignDrafts` add-only semantics — intentional and safe (local edits always win).
- Zod schema + versioned format for `loadDraftSnapshot` — correct boundary validation; graceful discard on mismatch.
- `saveDraftSnapshot` persists only text + attachment names (never base64) — correct; QuotaExceededError avoided.
- `queuedMessage` fully removed — verified no references remain anywhere in the codebase.
- `deleteSession` + `useDeleteAgent.onSuccess` both call `clearDraft`/`clearAgentDrafts` — correct lifecycle completeness.
- Draft indicator (`Edit` icon) is last in the `SessionIndicators` else-if chain — correctly shown only when idle (not working, not needing approval, no background work).
- Issue #3254 scope: all three sub-issues delivered with no undeclared additions.

---

```json
{"findings": [{"id": "F1", "sev": "🟡", "file": "packages/ui/src/modules/sessions/components/chat-input.tsx", "lines": "43-61", "rule": "state-management.md CRITICAL selector hooks", "fix": "export named selector hooks from sessions slice; replace inline useStore calls"}, {"id": "F2", "sev": "🟡", "file": "packages/ui/src/modules/sessions/store/sessions.ts", "lines": "88", "rule": "slice-composition StateCreator side effects", "fix": "move onForeignDraftChange to composition root in store.ts"}, {"id": "F3", "sev": "🟡", "file": "packages/ui/src/modules/sessions/lib/draft-snapshot.ts", "lines": "24-27", "rule": "localStorage at module-eval time without typeof window guard", "fix": "wrap in lazy getter function"}, {"id": "F4", "sev": "🟡", "file": "docs/architecture/platform-topology.md", "lines": "", "rule": "doc-drift: localStorage persistence substrate undocumented", "fix": "add one sentence to ui subsection, bump Last verified"}, {"id": "N1", "sev": "🔵", "file": "packages/ui/src/modules/sessions/components/chat-input.tsx", "lines": "103", "rule": "consistency: state.setDraft vs captured setDraft in async callback", "fix": "use captured setDraft reference"}]}
```

PetrBulanek:
Thanks — validated all four findings against the code. Two are fixed here, one is deferred with an issue, and one I believe is based on a mistaken premise.

### ✅ 2 · listener in the `StateCreator` body — fixed in bfaeccf1

Agreed, and the leak is real: no teardown, so every store recreation added a listener. Also no precedent — `createThemeSlice` only calls the pure `readStoredTheme()` in its initializer, and its `matchMedia` call sits inside a function invoked on demand. Moved to the composition root in `store.ts` as suggested.

### ✅ 4 · `platform-topology.md` — fixed in 2f2df960

Correct, the `ui` section documented reload behaviour without mentioning that client state now survives it. Added a sentence beside the existing reload prose and bumped `Last verified`.

### ✅ Note · `state.setDraft` vs captured `setDraft` — fixed in bfaeccf1

Good catch on the inconsistency; `addFiles` now writes through the captured `setDraft` like every other path (dependency array updated accordingly).

### ⚠️ 1 · inline `useStore` selectors — deferred to #3338

The rule is real, but this isn't drift the PR introduced. A sweep of `packages/ui/src`:

- **186** inline `useStore((s) => …)` call sites, across **62** files
- **0** exported selector hooks anywhere

Converting only `chat-input.tsx` would make it the sole component in 62 using a different access pattern. The skill's own guidance is *"Batch migrations are a separate PR."* Filed #3338 for the codebase-wide migration.

One correction on the rationale: the current single-field selectors return stable references (`s.setDraft` is one function identity; `EMPTY_DRAFT` is a frozen constant), so `Object.is` short-circuits and they never re-render. The proposed `useDraftActions` returns a *fresh object* — that is what creates the hazard `useShallow` then neutralises. Net perf change is zero, so this is a readability change, not a performance one. Noted in the issue so nobody picks it up expecting a rendering win.

### ❌ 3 · `localStorage` at module-evaluation time — no change, premise doesn't hold

The claim is that `localStore` "throws at parse time" in a non-browser context. It does not. The object literal creates two closures; `localStorage` is only dereferenced when a closure is **called**, not when the module is evaluated. Verified in a bare node process:

```
module-eval OK — object created, localStorage never touched
call-time throw CAUGHT by try/catch: ReferenceError
result: null
```

Both call sites already guard: `loadDraftSnapshot` wraps `store.getItem` in `try {} catch {}`, and `saveDraftSnapshot` wraps `setItem`. So even the call-time path degrades gracefully rather than throwing.

The decisive evidence: **vitest already runs `environment: "node"`** in this package, and all 290 tests pass — `loadDraftSnapshot()` runs at store creation on every test import. If this finding were real, the suite would already be red.

The proposed `getLocalStore()` factory has identical runtime behaviour — it moves object creation to call time, but `localStorage` is still only touched inside the closure. Happy to reconsider if I've misread the intent.

packages/ui/src/modules/sessions/store/sessions.ts
```diff
@@ -50,68 +58,145 @@ export const createSessionsSlice: StateCreator<
...
+        return true;
+      }),
+    pruneDrafts: (agentId, liveSessionIds) => {
+      const keep = new Set(liveSessionIds.map((id) => draftKey(agentId, id)));
+      keep.add(draftKey(agentId, null));
+      keep.add(draftKey(agentId, get().sessionId));
+      dropAgentKeys(agentId, (key) => !keep.has(key));
+    },
+    clearAgentDrafts: (agentId) => dropAgentKeys(agentId, () => true),
+    mergeForeignDrafts: (foreign) =>
```
dam-code-guardian: 🟡 **Warning:** Cross-tab merge never propagates a deletion, so a **sent** message comes back. This skips any key it already holds and adds only new ones, so when tab A sends and clears the draft, tab B keeps its stale copy and re-persists it on its next write.

I modelled both implementations against each other: after A sends, storage holds `{}`; then B's next keystroke writes the sent text back, and a reload in either tab restores it into the composer — one Enter away from a second send. Same class reported by `typescript-engineering`.

**Fix:** Cross-tab state needs a conflict policy that covers removals, not only additions. Reconcile against the whole incoming snapshot — drop local keys the snapshot no longer has — or carry a per-key version so the newer write wins; apply it to every key, not only the open session's.

packages/ui/src/modules/sessions/components/chat-input.tsx
```diff
@@ -188,7 +226,7 @@ export function ChatInput({
               ref={textareaRef}
               className="flex-1 bg-transparent border-0 pl-0 pr-2 py-[17px] text-sm leading-[22px] text-foreground resize-none min-h-0 max-h-[50vh] overflow-hidden disabled:opacity-40 focus-visible:ring-0 focus-visible:ring-offset-0"
               value={input}
-              onChange={(e) => setInput(e.target.value)}
+              onChange={(e) => key && setDraft(key, { text: e.target.value })}
```
dam-code-guardian, line 229: 🟡 **Warning:** Every keystroke performs a synchronous, main-thread `localStorage` write of **all** drafts. `setDraft` → `updateDrafts` → `saveDraftSnapshot` re-serializes every draft and calls `setItem` per character, and notifies every store subscriber app-wide; the previous local `useState` cost nothing outside this component.

Measured serialization alone: 0.15 ms at 40 drafts × 2000 chars, 0.82 ms at 80 × 4000 — the blocking `setItem` sits on top of that. Also reported by `react-ui-engineering`.

**Fix:** Never put a synchronous persistence write on a per-keystroke path. Keep the per-character value local and flush to the persisted store on a debounce plus blur/unmount/send, or write only the changed key instead of re-serializing the whole record.

packages/ui/src/modules/sessions/api/queries.ts
```diff
@@ -98,6 +99,13 @@ export function useAcpSessions(
     queryFn: live
       ? async () => {
           const sessions = await listAgentSessions(agentId);
+          const store = useStore.getState();
```
dam-code-guardian, line 102: 🟡 **Warning:** `pruneDrafts` runs inside the session-list `queryFn`, so a 5-second poll irreversibly deletes drafts. `keep` protects only `~new` and the currently-open session, so any listing that omits an idle session — the exact case the new sidebar marker advertises — destroys that draft with no undo.

The explicit paths already cover themselves (`deleteSession` calls `clearDraft`, `useDeleteAgent` calls `clearAgentDrafts`), so this reconciliation only adds risk on top. `typescript-engineering` reports the same call as a side effect inside a fetch adapter.

**Fix:** Destructive reconciliation must not be driven by a single poll result. Move the prune out of the `queryFn` to a subscriber on the query result, and require two consecutive listings to agree that a session is gone before dropping its draft.

docs/architecture/platform-topology.md
```diff
@@ -76,6 +76,8 @@ A React + Vite single-page app served by the api-server. It uses tRPC over a sin
 
 The URL addresses what the user is looking at — an Agent's chat, and the session open inside it — so a session is linkable from outside the UI and re-opens itself on a reload or a back step. A channel reply carries such a link back to the conversation it answered ([channels](channels.md)). Following one is owner-scoped like every other read: a Session belongs to its Agent's owner, and an Agent that isn't the viewer's simply isn't there — indistinguishable from one that never existed. Since the follower is usually *not* the owner (anyone in the conversation may click), that refusal is presented as its own screen naming the reason, not as an empty or perpetually-loading chat. There is no shared-session concept; the messenger conversation remains the shared surface.
 
+An unsent message is client state too: the composer's text is kept per session in the browser's `localStorage`, so a draft survives a reload or a browser close, belongs to the session it was written for rather than to the composer, and is dropped when it is sent or its session is deleted. Staged attachments live only for the tab — a restored draft names the files it lost.
```
dam-code-guardian: 🟡 **Warning:** The lifecycle and the invariant stated here are both incomplete. Drafts are also dropped on agent deletion (`clearAgentDrafts`) and by the session-list prune; cross-tab sharing and its local-wins rule are unmentioned; and "belongs to the session it was written for" does not hold before the session exists — the draft is held against the `~new` bucket and adopted by `migrateDraft`. Reported by `doc-drift`.

**Fix:** A page that enumerates a lifecycle must enumerate it completely or state the rule instead of the list. Add agent deletion and the server-reconciliation prune, the cross-tab visibility with whatever conflict rule the fix lands on, and the pre-session `~new` handover.

packages/ui/src/modules/sessions/components/sessions-sidebar.tsx
```diff
@@ -113,6 +115,9 @@ export function SessionsSidebar({
     [backgroundWork],
   );
 
+  const draftKeys = useStore(useShallow((s) => keysWithDraftContent(s.drafts)));
```
dam-code-guardian, line 118: 🟡 **Warning:** `keysWithDraftContent(s.drafts)` derives inside the Zustand selector, so every store notification — including the per-token `setMessages` of a streaming reply — re-runs `Object.keys` + `filter` + `sort` over all drafts. `useShallow` suppresses the re-render but not the work.

**Fix:** Keep selectors O(1) reads of state, never derivations — select the raw slice (`useStore((s) => s.drafts)`) and derive in a `useMemo` keyed on it, for this and any other multi-item projection read from the store.

packages/ui/src/modules/sessions/lib/draft-key.ts
```diff
@@ -0,0 +1,33 @@
+import type { Attachment } from "../../../types.js";
+
+export interface SessionDraft {
+  text: string;
+  attachments: Attachment[];
```
dam-code-guardian, line 5: 🟢 **Suggestion:** `EMPTY_DRAFT` and its `NO_ATTACHMENTS` array are frozen at runtime but typed as mutable `Attachment[]`, so a `push` on the shared singleton type-checks and throws only at runtime. Typing `attachments` as `readonly Attachment[]` moves the error to compile time.

```suggestion
  attachments: readonly Attachment[];
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `2f2df96`

## PR #3305: feat(ui): keep a message draft with the session it was written for
**Author:** PetrBulanek | **Branch:** feat/3254-session-drafts → main | **Changes:** +425 −113 (11 files)

### Summary
Complete re-review at the label's request. All four prior findings are resolved — two by fixes, two by declines that hold under test. The new commit that moved the cross-tab listener to the composition root made that path always-on, and this round's findings sit there and on the per-keystroke persistence path.

### Changes since last review
Previous HEAD: `8d6cfe7` (2026-08-14T14:48:21Z) — verdict COMMENT

- ✅ **Fixed:** cross-tab listener moved out of the `StateCreator` body to the composition root (`packages/ui/src/store.ts:65`, bfaeccf1)
- ✅ **Fixed:** `ui` subsection now documents localStorage as a persistence substrate (`docs/architecture/platform-topology.md:79`, 2f2df960)
- ✅ **Fixed:** `addFiles` writes through the captured `setDraft` (`chat-input.tsx:110`, bfaeccf1)
- ✅ **Resolved — decline holds:** module-eval `localStorage` access. I reproduced the author's evidence in `node`: the object literal creates closures and never dereferences `localStorage` until called, both call sites already `try/catch`, and this package's vitest runs `environment: "node"` with the loader executing at store creation. **My previous 🟡 was wrong**; no change is needed.
- ✅ **Deferred — decline holds:** inline `useStore` selectors. Sweep verified independently — 187 inline call sites across 62 files, zero exported selector hooks in `packages/ui/src`. A single-file conversion is correctly out of scope; tracked in #3338.

### Findings
- 🟡 **Warning:** Cross-tab merge never propagates a deletion, so a **sent** message comes back. `mergeForeignDrafts` skips any key it already holds and adds only new ones, so when tab A sends and clears the draft, tab B keeps its stale copy and re-persists it on its next write. I modelled both implementations: after A sends, storage holds `{}`, then B's next keystroke writes the sent text back — a reload in either tab restores it into the composer, one Enter away from a second send. Same defect class reported by `typescript-engineering`. (`packages/ui/src/modules/sessions/store/sessions.ts:154`)
  **Fix:** Cross-tab state needs a conflict policy that covers removals, not only additions. Reconcile against the whole incoming snapshot — drop local keys the snapshot no longer has — or carry a per-key version so the newer write wins; apply it to every key, not only the open session's.
- 🟡 **Warning:** Every keystroke performs a synchronous, main-thread `localStorage` write of **all** drafts. `setDraft` → `updateDrafts` → `saveDraftSnapshot` re-serializes every draft and calls `setItem` per character, and notifies every store subscriber app-wide; the previous local `useState` cost nothing outside this component. Measured serialization alone: 0.15 ms at 40 drafts × 2000 chars, 0.82 ms at 80 × 4000 — the blocking `setItem` sits on top. Also reported by `react-ui-engineering`. (`packages/ui/src/modules/sessions/components/chat-input.tsx:229`, writer at `store/sessions.ts:68`)
  **Fix:** Never put a synchronous persistence write on a per-keystroke path. Keep the per-character value local and flush to the persisted store on a debounce plus blur/unmount/send, or write only the changed key rather than re-serializing the whole record.
- 🟡 **Warning:** `pruneDrafts` runs inside the session-list `queryFn`, so a 5-second poll irreversibly deletes drafts. `keep` protects only `~new` and the currently-open session, so any listing that omits an idle session — the exact case the new sidebar marker advertises — destroys that draft with no undo. The explicit paths already cover themselves (`deleteSession` calls `clearDraft`, `useDeleteAgent` calls `clearAgentDrafts`), so this reconciliation only adds risk on top. `typescript-engineering` reports the same call as a side effect in a fetch adapter. (`packages/ui/src/modules/sessions/api/queries.ts:102`)
  **Fix:** Destructive reconciliation must not be driven by a single poll result. Move the prune out of the `queryFn` to a subscriber on the query result, and require two consecutive listings to agree that a session is gone before dropping its draft.
- 🟡 **Warning:** The new topology paragraph states a closed lifecycle and a full invariant, and both are incomplete: drafts are also dropped on agent deletion (`clearAgentDrafts`) and by the session-list prune, cross-tab sharing and its local-wins rule are unmentioned, and "belongs to the session it was written for" does not hold before the session exists (`~new` key, adopted by `migrateDraft`). Reported by `doc-drift`. (`docs/architecture/platform-topology.md:79`)
  **Fix:** A page that enumerates a lifecycle must enumerate it completely or state the rule instead of the list. Add agent deletion and the server-reconciliation prune, the cross-tab visibility with whatever conflict rule the fix above lands on, and the pre-session `~new` handover.
- 🟢 **Suggestion:** `EMPTY_DRAFT` and its `NO_ATTACHMENTS` array are frozen at runtime but typed as mutable `Attachment[]`, so a `push` on the shared singleton type-checks and throws only at runtime. Typing `attachments` as `readonly Attachment[]` moves it to compile time. (`packages/ui/src/modules/sessions/lib/draft-key.ts:5`)
- 🟢 **Suggestion:** With `agentId` null, `key` is null and four write paths become silent no-ops while the textarea stays enabled. On a hard load of a chat URL the window is one paint — not reachable by a human today — but any future change that defers agent resolution turns it into silent data loss. Deriving one `ready = key !== null` flag into the existing `disabled` conditions makes the invariant explicit. (`packages/ui/src/modules/sessions/components/chat-input.tsx:47`)
- 🟢 **Suggestion:** The persisted record is unbounded — no cap on entries or text length, and `pruneDrafts` only ever cleans the selected agent — so once the quota is reached `setItem` throws on every keystroke and draft persistence stops for good behind a `console.warn`. (`packages/ui/src/modules/sessions/lib/draft-snapshot.ts:14`)
- 🟢 **Suggestion:** `onForeignDraftChange` registers the `storage` listener at import time and returns `void`, so it cannot be torn down and re-registers on a hot reload. Returning the unsubscribe handle fixes the class. (`packages/ui/src/store.ts:65`)

- 🟡 The new draft paragraph omits cross-tab sharing and its local-wins conflict rule (`onForeignDraftChange` at `draft-snapshot.ts:94`, wired at `store.ts:65`; `mergeForeignDrafts` at `store/sessions.ts:154`). (`docs/architecture/platform-topology.md:79`) — carried in `### Findings` above.
- 🟡 The stated lifecycle ("dropped when it is sent or its session is deleted") misses agent deletion (`agents/api/mutations.ts:109`) and the live-list prune (`sessions/api/queries.ts:104`). (`docs/architecture/platform-topology.md:79`) — carried above.
- 🟡 The invariant "belongs to the session it was written for" does not cover the pre-session case: the draft is held against a placeholder bucket (`lib/draft-key.ts:17`) and adopted by `migrateDraft` (`hooks/use-acp-connection.ts:151`); a missed handover orphans it. (`docs/architecture/platform-topology.md:79`) — carried above.
  **Fix:** State the invariant with its boundary case whenever state is carried across an identity that does not yet exist.
- 🟢 `persistence.md:7` counts "four durable substrates" (plus a named fifth) and an unsent message now survives a browser close on a substrate that page does not name. Judgement call whether browser-side storage belongs in a count whose split is server-side only; if not, a one-liner and cross-link would stop the count reading as exhaustive.

### Issue Fit
- 🟡 Undeclared scope creep: cross-tab draft synchronization is asked for nowhere in #3254 — which scopes persistence to "one draft per session, kept on the browser it was written on" — and is declared in neither the PR body nor the three sub-issues. It is also where this round's first finding lives, so the undeclared behaviour is the one carrying a defect. (`packages/ui/src/store.ts:65`, `lib/draft-snapshot.ts:94`, `store/sessions.ts:154`)
  **Fix:** Behaviour beyond the issue's stated scope is declared before merge — name the cross-tab merge in the PR description and in #3254 (a closing keyword will close it), or drop `onForeignDraftChange` and `mergeForeignDrafts`.

Otherwise the diff delivers the ask: session-keyed text and attachments including the not-yet-created chat, text-only persistence with a notice toast naming dropped attachments, the marker in the existing per-session indicator slot behind approval/working/background-work, terminal sessions excluded, and clearing on send, session delete and agent delete. All three sub-issues are covered; nothing server-side.

- 🟡 `mergeForeignDrafts` propagates no deletions (`store/sessions.ts:158`) — merged into `### Findings` #1.
- 🟡 `pruneDrafts` called inside the fetch adapter (`api/queries.ts:102`) — merged into `### Findings` #3.
- 🟡 `onForeignDraftChange` returns `void`, subscription cannot be torn down (`store.ts:65`) — carried as a 🟢 above; at the composition root it evaluates once in production, so the exposure is hot reload and tests.
- 🟡 Four disagreeing "empty draft" definitions — `draftHasContent` and `saveDraftSnapshot` trim, `setDraft`/`loadDraftSnapshot`/`consumeDroppedAttachments` do not (`lib/draft-snapshot.ts:81`). Verified consequence: a whitespace-only draft is kept in memory and dropped on persist, which is the desired outcome reached by inconsistent rules, so no defect today.
  **Fix:** Export one emptiness predicate from `draft-key.ts` and call it from the store, the loader and the saver.
- 🟡 The `DraftStore` port is bypassed by a default parameter binding the concrete `localStorage` adapter (`store/sessions.ts:68`, `lib/draft-snapshot.ts:30`). No defect follows — the node-environment suite passes through the guarded call sites — so this is a layering preference, not a finding.
- 🟡 Draft-key format knowledge duplicated: `draftKey` builds `<agentId>:<sessionId>`, the store re-decodes with `startsWith` (`store/sessions.ts:72`). Agent ids match `^[a-z0-9]+(-[a-z0-9]+)*$` (`api-server-api/src/modules/shared.ts:13`), so the delimiter is unambiguous and prefix matching is sound; no defect.
- 🟡 The `agents` module writes `sessions` draft state directly (`agents/api/mutations.ts:109`). Measured against the tree: this pattern is the repo baseline, not a violation — `acp/acp.ts` calls `addPendingPermission`, `experiments/…` calls `openAgentSession`, `platform/hooks/use-browser-history.ts` calls `resetChatContext`, and no repo doc states a boundary rule. Not a finding here.
- 🟢 `EMPTY_DRAFT` frozen but typed mutable (`lib/draft-key.ts:12`) — carried above.
- 🟢 Persisted `drafts` record unbounded (`lib/draft-snapshot.ts:14`) — carried above.

- 🟡 `keysWithDraftContent(s.drafts)` derives inside the Zustand selector, so every store notification — including the per-token `setMessages` of a streaming reply — re-runs `Object.keys` + `filter` + `sort`; `useShallow` suppresses the re-render but not the work. (`components/sessions-sidebar.tsx:118`)
  **Fix:** Keep selectors O(1) reads — select `s.drafts` and derive in a `useMemo` keyed on it, for this and any other multi-item projection read from the store.
- 🟡 Per-keystroke store dispatch and synchronous persistence write (`components/chat-input.tsx:229`) — merged into `### Findings` #2.
- 🟡 The composer renders enabled while `key` is null and silently discards input (`components/chat-input.tsx:47`) — carried as a 🟢 above; the window is a single paint on a hard load, so it is a latent invariant rather than a live defect.
- 🟢 Inline `useStore` reads rather than exported selector hooks — codebase-wide drift (187 sites / 62 files / zero hooks), tracked in #3338, not this PR's to carry.

### Verdict
COMMENT — every prior finding is resolved, but the now-active cross-tab path resurrects sent drafts, the composer persists synchronously per keystroke, and a 5-second poll can delete drafts irreversibly.

---

packages/ui/src/modules/sessions/store/sessions.ts
PetrBulanek: Confirmed and fixed in 36265ec8 — but the cause is wider than `mergeForeignDrafts`, so the suggested fix would not have closed it.

I modelled the two tabs against the real code. The severe case never touches the merge at all:

**Both tabs loaded while the draft existed.** Tab A sends, storage holds `{}`, then any keystroke in tab B rewrites the **full sent text** — because `updateDrafts` serialized the *whole* record from one tab's memory, so every stale key that tab still held went back to disk. Add-only merge is irrelevant here; B already had the draft from its own `loadDraftSnapshot()`.

**Typed live in A while B is open.** Here the merge does apply, and the add-only rule truncates it: B accepts the first keystroke and skips the rest, so what returns is `"s"` of `"ship it"`, not the sent text.

That second run also surfaced a defect not in the review: **that first keystroke lands in tab B's open composer.** A types, and a character appears in a box in B where nobody typed — the class #3254 exists to eliminate.

**What landed instead of snapshot reconciliation:** one `localStorage` entry per draft (`platform-draft:<agentId>:<sessionId>`). A write touches only the changed keys, so a stale key is never rewritten and the severe case cannot occur. `clearDraft` now issues `removeItem`, so other tabs receive a `storage` event with `newValue === null` and drop the key — deletion propagates without a conflict policy. Foreign changes apply per key, and the key the person is currently typing in is skipped, which closes the composer bleed.

I chose this over per-key versions deliberately. Versions only resolve concurrent edits of the *same* key, which per-key writes already settle, and they would add a clock to the persisted format for no other gain.

Verified on the dev cluster with two tabs: tab A clears a draft, tab B's marker disappears, and typing in tab B leaves the cleared key absent from storage. On the previous code that keystroke brought it back.

packages/ui/src/modules/sessions/components/chat-input.tsx
PetrBulanek: Behaviour confirmed and addressed in 36265ec8, though not for the reason given — and there is a bigger cost the review missed.

On your own numbers this was not a warning: 0.15 ms at 40 drafts × 2000 chars is a hundredth of a frame budget, and 40 drafts × 2000 chars is already an extreme record. The writer's serialization was never the problem.

The real cost sat on the **receiving** tabs. Each keystroke wrote the whole blob, which fires a `storage` event in every other tab of the browser, and each of those ran `JSON.parse` plus a Zod record parse over every draft. One person typing made every other tab re-validate the entire record per character.

Both halves are gone now:

- Writes are per key, so a keystroke serializes one small entry rather than the whole record.
- Text writes batch on a 300 ms deadline. I used a batch deadline rather than a resetting debounce on purpose: a resetting timer never fires while someone keeps typing, which is exactly when the draft most needs to be on disk. Latency is bounded at 300 ms.
- Removals bypass the batch and write through at once, so a `clearDraft` on send can never be undone by a queued write for the same key.
- `pagehide` and `visibilitychange` flush, so closing the tab mid-word does not lose the tail.

One correction on scope: keeping the value local in `useState` was not an option. The draft has to live in the store — the sidebar marker and `migrateDraft` both read it — so debouncing the *persistence* is the fix, not moving the value back into the component.

packages/ui/src/modules/sessions/api/queries.ts
PetrBulanek: No change — the stated failure mode does not exist, and the mechanism proposed to guard it guards nothing.

The scenario is "any listing that omits an idle session". No such listing reaches `pruneDrafts`:

- The prune is fed `sessions` — the **unfiltered** result. The `allowed` type filter runs *after* it, at line 109, and produces `fresh`, which the prune never sees. Channel and scheduled sessions are in the keep-set even when `include` hides them.
- `listAgentSessions` is unpaginated: `conn.listSessions({ cwd: "." })`, mapped and sorted, no cap (`api/acp-session-ops.ts:66`).
- A stopped or unreachable sandbox does not answer with a short list. `passive=1` requires `repo.isReady(agentId)` and otherwise closes the socket with `1011` (`agent-proxies/acp-relay.ts:196`), so the `queryFn` rejects and the prune never runs.

So a key the prune drops is a key whose session the server genuinely no longer has — which is the cleanup this call is for. "Two consecutive listings must agree" would only delay that by one poll; it would not protect a draft, because there is no path that produces the disagreeing listing in the first place.

I also checked the timing hazard the finding implies but does not name: a brand-new session between `migrateDraft` and `setSessionId`. Those two run back to back with no `await` between them (`hooks/use-acp-connection.ts:151`), and before the migration the draft sits under `~new`, which `pruneDrafts` always keeps. No window.

The layering point underneath this — a store mutation inside a `queryFn` — is fair on style, and I have left it alone deliberately: moving it to a subscriber trades one clear call site for a second source of truth about when the list arrived, for no behavioural gain.

docs/architecture/platform-topology.md
PetrBulanek: Partly taken, rewritten in 36265ec8.

You are right that a closed enumeration was the wrong shape, and I took your alternative — state the rule instead of the list. The paragraph now says a draft "lives exactly as long as that session does — sending it, or losing the session by any route, takes the draft with it". That covers agent deletion and the server-reconciliation prune without naming either, because neither is a separate rule: both are the session going away.

Cross-tab was the genuinely missing behaviour, and it is now named with its conflict rule: every tab of the browser sees the same drafts, a tab only rewrites the draft it is editing, and the composer someone is typing in is never overwritten from elsewhere.

The `~new` handover I have left out on purpose. `documentation-guidelines.md:38` is explicit — *"If a sentence would break when someone renames a field … it is pitched too low — raise it until it describes the meaning, not the shape."* A placeholder bucket key is exactly that shape, and it would break the moment the key changes. The meaning-level version of the same fact is already in the paragraph: the draft belongs to the session it was written for "rather than to the composer", which is the claim that has to hold before the session exists as much as after.

On `persistence.md:7` — declined. That count sorts durable substrates into platform-owned and agent-owned, both server-side, and its purpose is the rule about which to choose for new state. A browser's `localStorage` is not a candidate in that decision, so adding it would blur what the count is for. The topology page is the right home, which is where it is.

packages/ui/src/modules/sessions/components/sessions-sidebar.tsx
PetrBulanek: The observation is right and the fix is a regression, so no change here.

`useShallow` does let the derivation run on every notification — agreed. But `updateDrafts` assigns a **fresh** `drafts` object on each edit (`{ ...before }`, then `set({ drafts })`). Selecting the raw slice therefore changes the selector's identity on every keystroke, and the whole sidebar — every session row, with its indicators — re-renders per character. `useShallow` is what prevents exactly that today: it compares the *projection*, so a keystroke that does not change which sessions hold content re-renders nothing.

So the trade is a filter-plus-sort over a handful of keys, against a full list re-render per character. The current code takes the cheaper side.

Batching the persistence (see the `chat-input.tsx` thread) does not change this, to be clear: the store notification is still immediate; only the disk write is deferred.

The general rule is sound and I am not arguing with it — a derivation in a selector is a smell worth flagging. It is just that here the smell is load-bearing, and the raw-slice form only becomes the better shape if the per-keystroke store write goes away, which would mean taking the draft back out of the store and losing the sidebar marker.

packages/ui/src/modules/sessions/lib/draft-key.ts
PetrBulanek: The reasoning holds — a `push` on the frozen singleton does type-check and throw only at runtime — but the committed suggestion does not compile, so I am leaving it.

`readonly Attachment[]` on this field propagates through every consumer of a draft's attachments: `send` hands `current.attachments` to `onSend`, declared `(text: string, attachments?: Attachment[]) => void` (`chat-input.tsx:29`), and from there to `use-acp-prompt.ts:47` and `:98`, `chat-message.tsx:14` and `acp/utils.ts:17`. A `readonly` array is not assignable to any of them, so the one-line change turns into five signature changes across the prompt path.

That is a reasonable change to make; it is not a reasonable change to smuggle into a draft-persistence PR, where it would touch the send path this feature deliberately leaves alone. Not filing it either — it is a two-minute edit for whoever next has the send path open, and a tracking issue for a type widening would outlive its usefulness.

Worth noting the exposure is small: `EMPTY_DRAFT` is only ever spread (`{ ...(drafts[key] ?? EMPTY_DRAFT), ...patch }`) or read, never mutated, and `Object.freeze` already converts a mistake into a throw rather than shared-state corruption.

PetrBulanek:
Validated all ten findings against the code. One was a real bug, and its stated cause was not the actual one; three do not hold. 36265ec8 fixes what is real. Replies are on each inline thread — this covers the findings that had none.

## The one defect, and what it actually was

`mergeForeignDrafts` was blamed, but the severe case never reached it. `updateDrafts` serialized the **whole** draft record from one tab's memory on every write, so any stale key that tab still held was written back to disk. Two tabs open, A sends, B types anywhere — B rewrites the full sent text. Add-only merge only mattered for the weaker variant, where it truncated the resurrected text to the first keystroke.

Modelling it also turned up a defect the review did not report: **that first keystroke lands in the other tab's open composer.** Tab A types, a character appears in tab B's message box where nobody typed it — precisely what #3254 exists to prevent.

The fix is one `localStorage` entry per draft, so a write touches only the keys that changed. Details on the `store/sessions.ts:154` thread.

## Findings with no inline thread

**Null `key` leaves write paths as silent no-ops** — accepted, fixed. Reachable in principle: `chat-view.tsx` has no `!selectedAgent` guard before `<ChatInputArea>`. `disabled` on the textarea, the attach button and the send button now all carry `!key`, so the invariant is stated where it can be seen rather than implied by four early returns.

**Import-time listener with no teardown** — accepted, fixed. `onForeignDraftChange` returns its unsubscribe, `flushDraftsOnHide` returns its own, and `stopDraftSync()` in `store.ts` composes both.

**Unbounded persisted record** — no change. Real but not worth code: entries hold text only (base64 never persists), and the quota failure path is already `console.warn` and continue, not a crash. `pruneDrafts` cleaning only the selected agent is the honest half of this — drafts for an agent never revisited do persist until that agent is deleted. A cap would need a policy for which draft to evict, and silently discarding someone's unsent message to stay under a limit is worse than the leak.

**Four disagreeing empty-draft definitions** — no change, and I agree with your own reading that there is no defect. The outcome is right by three independent rules rather than one, which is worth consolidating some day; it is not worth a churn commit in this PR.

**Cross-tab sync is undeclared scope** — accepted. Now declared in the PR body under its own heading, with the conflict rule, and named on #3254. I kept the behaviour rather than dropping it: dropping the listener would not have fixed the resurrection anyway, since that came from the whole-record write, so removing it would have deleted a feature and left the bug.

## Not filed as follow-ups

`readonly Attachment[]` (five signature changes on the send path) and the single emptiness predicate are both left in place with the reasoning on their threads. Neither is getting a tracking issue — they are small enough that an issue would outlast the work.

## Verification

- `ui:check` and `ui:test` green — 290 tests, no new ones. tsc, lint and format clean.
- Cross-tab behaviour driven against the real implementation, then smoke-tested on the dev cluster with two tabs: session-scoped drafts, restore on reload, the sidebar marker, deletion propagating to the second tab, and — the case that used to fail — typing in the second tab after the first clears, with the cleared key staying absent from storage.
- No migration from the old `platform-drafts` blob. The feature has not shipped, so the only holders are dev browsers, and migration code for an unreleased format is dead weight.

packages/ui/src/modules/sessions/components/chat-input.tsx
```diff
@@ -37,52 +40,85 @@ export function ChatInput({
...
-              size: file.size,
-            },
-          ]);
+  const consumeDroppedAttachments = useStore(
+    (s) => s.consumeDroppedAttachments,
+  );
+  const droppedNames = draft.droppedAttachmentNames;
+  useEffect(() => {
+    if (!key || !droppedNames?.length) return;
+    const fresh = useStore.getState().drafts[key]?.droppedAttachmentNames;
```
dam-code-guardian, line 65: The effect fires because `droppedNames` (line 62) changed — at that point it is already the authoritative value. The `getState()` re-read on `fresh` adds a narrow stale-read risk: if a Zustand batch flush races between effect entry and the `getState()` call, `fresh?.length` is 0 and the toast is skipped despite `droppedNames` being non-empty. The toast message also uses `fresh.join` rather than `droppedNames.join`, so the displayed names come from a different read than the guard that passed.

**Fix:** use `droppedNames` directly and remove the `getState()` call:
```ts
useEffect(() => {
  if (!key || !droppedNames?.length) return;
  emitToast({
    kind: "info",
    message: `Draft restored without ${droppedNames.length} attachment${
      droppedNames.length === 1 ? "" : "s"
    }: ${droppedNames.join(", ")}`,
  });
  consumeDroppedAttachments(key);
}, [key, droppedNames, consumeDroppedAttachments]);
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `36265ec`

## PR #3305: feat(ui/sessions): bind chat draft to session

**Verdict:** COMMENT — one new finding; all three prior 🟡 resolved.

### Changes since last review
Previous HEAD: `2f2df96` (2026-08-17T06:57:24Z) — verdict COMMENT

- ✅ **Fixed (F1):** Cross-tab deletion now propagates. `onForeignDraftChange` passes `null` for `e.newValue === null`; `applyForeignDraft` deletes the key when `draft === null`; `updateDrafts` diffs individual keys and calls `draftWriter.write(key, null)` for deleted keys, issuing `removeItem` immediately. Verified on both `storage` event path (removal → delete in memory + propagates null to other tabs) and the `send` path (clearDraft → immediate removeItem → null event in other tabs).
- ✅ **Fixed (F2):** Per-keystroke writes now debounced. `createDraftWriter` batches text writes on a 300 ms deadline with a resetting-free timer (`setTimeout`, not `clearTimeout` on each call) and writes only the changed key, not the whole record. Removals bypass the batch (`clearDraft` → `writeDraftEntry(key, null, store)` immediately) so a send cannot be re-queued. `pagehide` and `visibilitychange` flush the batch before the tab unloads.
- ✅ **Fixed (F4):** Topology paragraph now states the closed lifecycle as a rule ("lives exactly as long as that session does — sending it, or losing the session by any route, takes the draft with it"), names cross-tab visibility and the local-wins conflict rule. The `~new` placeholder omission is a deliberate scope choice (guidelines forbid naming the storage key) and the rule-form covers the pre-session case correctly.
- **Decline holds (F3 — pruneDrafts in queryFn):** The stated scenario ("any listing that omits an idle session destroys its draft") cannot occur. `pruneDrafts` is fed the raw, unfiltered `sessions` result; the type filter that produces `fresh` runs three lines later and is never passed to the prune. `listAgentSessions` is unpaginated and `passive=1` causes the server to reject connections to unreachable agents with 1011, so the queryFn rejects before pruning. The open session is in the keep-set via `get().sessionId` regardless. Decline credited.

### Findings

| ID | Sev | File | Line | Summary |
|----|-----|------|------|---------|
| F1 | 🟡 | `packages/ui/src/modules/sessions/components/chat-input.tsx` | 65 | `useEffect` re-reads dropped names via `getState()` after already having them reactively |

---

**🟡 F1 — `useEffect` for dropped-attachment toast has a redundant `getState()` re-read**

The effect fires because `droppedNames` (the reactive selector result at line 62) changed. At that point `droppedNames` is already the authoritative value — the re-read on `fresh` (line 65) adds a narrow stale-read risk: if a Zustand batch flush or a `consumeDroppedAttachments` call races between the effect body entering and the `getState()` call, `fresh?.length` is 0, the effect short-circuits, and the toast is never shown — while `droppedNames` had a non-zero length. The message text also uses `fresh.join` rather than `droppedNames.join`, so the displayed file names come from a different read than the length guard that passed.

**Fix:** Use `droppedNames` directly throughout and remove the `getState()` call:
```ts
useEffect(() => {
  if (!key || !droppedNames?.length) return;
  emitToast({
    kind: "info",
    message: `Draft restored without ${droppedNames.length} attachment${
      droppedNames.length === 1 ? "" : "s"
    }: ${droppedNames.join(", ")}`,
  });
  consumeDroppedAttachments(key);
}, [key, droppedNames, consumeDroppedAttachments]);
```

---

### Suggestions

- 🟢 **`stopDraftSync()` is exported but `startDraftSync()` is implicit** — `onForeignDraftChange` and `flushDraftsOnHide` add DOM listeners at module evaluation time (`store.ts:68–71`), so the sync starts as a side effect of importing the store rather than at a named call site in `main.tsx`. The teardown export is correct; moving the start to an explicit `startDraftSync()` called from `main.tsx` alongside the store creation would make the pair symmetric and visible in the composition root. No behavioural gap in production (single entry point, single evaluation); matters in tests and HMR. (`packages/ui/src/store.ts:68`)
- 🟢 **`EMPTY_DRAFT.attachments` is frozen but typed as mutable `Attachment[]`** — a `push` on the shared singleton would type-check and throw only at runtime. Typing `attachments: readonly Attachment[]` on `SessionDraft` moves the error to compile time. Propagates to `onSend` and the prompt path; worth a standalone change, not this PR. (`packages/ui/src/modules/sessions/lib/draft-key.ts:5`)
- 🟢 **Persisted record is unbounded** — no cap on keys or text length; once `localStorage` quota is reached `setItem` throws and draft persistence stops behind a `console.warn`. A per-key size or total-key count guard would make the failure visible rather than silent. (`packages/ui/src/modules/sessions/lib/draft-snapshot.ts:14`)

✅ Aligned — `Last verified: 2026-08-17` bumped, paragraph updated with rule-form lifecycle, cross-tab sharing with local-wins, and attachment limitation. No ADR leaks, no volatile content.

### Issue Fit (issue-fit)
✅ Delivers #3254 fully. Cross-tab behaviour declared in PR body and in the author's issue comment (#issuecomment-5313837652); no undeclared scope.

### TypeScript Engineering
H1 (cross-module store write in `mutations.ts`) — repo baseline; `acp/acp.ts` → `addPendingPermission`, `experiments/…` → `openAgentSession`, `platform/hooks/use-browser-history.ts` → `resetChatContext`. Not a finding here.
H2 (pruneDrafts in queryFn) — addressed above under round-2 F3 decline.
M1 (module-level side effects in `store.ts`) — same point as 🟢 suggestion above.
M2 (draftWriter singleton) — no behavioural defect; factory pattern is correct, the singleton is a convenience shortcut. Worth a follow-up if the test isolation concern bites.
M3 (EMPTY_DRAFT frozen/typed mutable) — same as 🟢 suggestion above.

### React UI Engineering
C1 (inline `useStore` selectors) — 187 sites / 62 files / 0 exported hooks codebase-wide; tracked in #3338; not this PR's to carry.
H1 (draft key derivation inlined) — the private `openDraftKey()` is intentionally private; cross-component exposure is a style preference. Pre-existing pattern.
H2 (`useEffect` double-read) — carried as 🟡 F1 above.
M1 (`renderRow` inline function) — pre-existing pattern; PR added 4 lines inside it.

### Verdict
COMMENT — F1 is a narrow defensive fix; the three prior 🟡 are closed. The `~new` pre-session lifecycle, cross-tab sync, per-key writes and debounced persistence all verified against the actual code.

---

packages/ui/src/modules/sessions/components/chat-input.tsx
PetrBulanek: No change — the stated mechanism cannot occur, and the suggested fix introduces the bug it is meant to prevent.

**The race is not reachable.** Lines 64 and 65 are adjacent synchronous statements. There is no `await` and no yield point between the guard and the `getState()` call, so nothing — a Zustand batch flush included — can interleave. A single-threaded runtime gives that for free.

**The re-read is load-bearing, not redundant.** `main.tsx:32` wraps the app in `StrictMode`, which re-runs mount effects. This effect always runs at mount: `drafts` is seeded from `loadDraftSnapshot()` at store creation (`store/sessions.ts:99`), so the dropped names are present before `ChatInput` ever mounts. Run 1 emits the toast and calls `consumeDroppedAttachments`. Zustand's `set` is synchronous, so run 2 — same commit, same stale closure, `droppedNames` still non-empty — reads the consumed state through `getState()` and returns at line 66.

Apply the suggested diff and **the toast fires twice in dev**, once per StrictMode invocation. The `getState()` read is what deduplicates it.

**On the second claim** — "the displayed file names come from a different read than the length guard that passed" — that is not what the code does. The guard that gates the toast is line 66, on `fresh`; the message uses `fresh.length` and `fresh.join`. One read, used for both. Line 64's `droppedNames` check is a cheap bail before touching the store, not the guard the toast passes.

The genuinely useful framing here is the opposite of the finding: this effect reads a value that a second mount can invalidate, and the `getState()` call is the thing that makes it safe.

PetrBulanek:
Round 3 answered: one taken, three declined. Details on the inline thread for F1; the suggestions are below.

## Taken — `startDraftSync()` at the composition root (5b2e1cd)

Taken, and the finding understated it. `stopDraftSync` was exported and **called from nowhere in the repo** — `grep -rn "stopDraftSync" packages/` returned only its own definition. So it was not merely asymmetric with an implicit start; it was dead code, and the teardown capability added in round 2 had no live path to prove it worked.

`startDraftSync()` now registers both listeners and stores their teardown; `stopDraftSync()` runs whatever that call registered. `main.tsx` calls it next to `hydrateRoute()`, so both halves of the pair are visible at one call site and neither runs as a side effect of an import. Re-entry is guarded, so a double call cannot register two listeners.

No behaviour change in production — single entry point, single evaluation — as the finding said.

## Declined — 🟡 F1, `useEffect` `getState()` re-read

See the inline thread. Short version: the described race sits between two adjacent synchronous statements and cannot happen, and the `getState()` read is what stops `StrictMode`'s second effect invocation from emitting a duplicate toast. The suggested diff would ship that duplicate.

## Declined — `EMPTY_DRAFT.attachments` as `readonly`

Same answer as round 2, which this round already credits ("worth a standalone change, not this PR"). `readonly Attachment[]` is not assignable to the five send-path signatures that consume a draft's attachments, so the one-line change is really five, on the path this feature deliberately does not touch.

## Declined — unbounded persisted record

The severity here is measured against code that no longer exists. It describes the whole-record write, where one `QuotaExceededError` stopped all persistence. Since 36265ec writes are per key and wrapped in `try/catch` (`draft-snapshot.ts:109`), and `removeItem` keeps working regardless — so a full quota fails the one write that overflows and recovers as soon as any draft is sent or cleared. It degrades; it does not stop "for good".

The residual growth path is real but small: `pruneDrafts` only ever cleans the selected agent, so drafts for an agent never revisited persist. They are text only, and reaching a 5 MB quota needs millions of characters. A cap would add a policy — which draft to sacrifice — to defend against a case no user reaches.

---

`mise run ui:check` green. Rounds 1 and 2 remain closed.

dam-code-guardian (commented):
**Code Guardian 🛡️ — complete re-review (round 4)**

Delta reviewed: `36265ec` → `5b2e1cdd` — 2 files: `packages/ui/src/main.tsx`, `packages/ui/src/store.ts`

---

### Summary

The delta adds `startDraftSync()` / `stopDraftSync()` to `store.ts` and calls `startDraftSync()` from `main.tsx` after `hydrateRoute()` — closing the round-3 🟢 symmetry suggestion cleanly. Two new 🟡 findings from the delta; three carried forward.

---

### 🟡 F1 — `useEffect` double-reads `droppedAttachmentNames` via `getState()` *(carried from round 3)*

`chat-input.tsx:65` — not in this delta, still unaddressed:

```tsx
const droppedNames = draft.droppedAttachmentNames;   // reactive dep
useEffect(() => {
  if (!key || !droppedNames?.length) return;
  const fresh = useStore.getState().drafts[key]?.droppedAttachmentNames; // ← redundant
  if (!fresh?.length) return;
  emitToast({ ... `Draft restored without ${fresh.length}...` });
  consumeDroppedAttachments(key);
}, [key, droppedNames, consumeDroppedAttachments]);
```

`droppedNames` is the reactive, Zustand-subscribed value that triggers the effect. `fresh` reads the same slice imperatively one line later — the guard `!fresh?.length` and the `fresh.length` in the toast can both just use `droppedNames`. The `.getState()` call inside an effect is a violation of the reactive contract React hooks expect.

**Fix:** Remove `const fresh` and the `!fresh?.length` guard; replace `fresh.length` in the toast with `droppedNames.length`.

---

### 🟡 F2 — `visibilitychange` flush fires on become-visible, creating a clobber race *(new)*

`packages/ui/src/modules/sessions/lib/draft-snapshot.ts:175` (called via `startDraftSync`):

```ts
const flush = () => writer.flush();
document.addEventListener("visibilitychange", flush);
```

The handler fires on every visibility transition — including when the tab becomes visible again. When another tab wrote a draft while this tab was hidden, both events arrive in the macrotask queue at approximately the same time: the `storage` event carrying the foreign write and the `visibilitychange` for becoming visible. If the `visibilitychange` flush runs first, it serializes this tab's stale in-memory state to `localStorage` before `applyForeignDraft` has updated the Zustand store — potentially clobbering the newer value from the other tab.

**Fix:** Guard the handler so it only flushes on hide:

```ts
const flush = () => {
  if (document.visibilityState === "hidden") writer.flush();
};
```

The `pagehide` listener already covers the unload path; this guard makes the become-visible transition a true no-op.

---

### 🟡 F3 — Module-level `draftSyncTeardown` causes silent listener leakage in tests *(new)*

`packages/ui/src/store.ts:34`:

```ts
let draftSyncTeardown: (() => void) | null = null;
```

In any test environment where module state persists across test cases (Vitest without `resetModules`), a test that calls `startDraftSync()` without a matching `stopDraftSync()` in `afterEach` leaves the `storage` and `visibilitychange` listeners live. A subsequent test hits the non-null guard and silently skips re-registration — meaning that test's store receives no foreign draft events even though the feature appears active.

**Fix:** Any test that calls `startDraftSync()` must pair it with `stopDraftSync()` in `afterEach`. For HMR coverage in development, an `import.meta.hot?.dispose(() => stopDraftSync())` guard would also clean up on hot replacement.

---

### 🟡 F4 — `draft-key.ts` lifecycle doc omits cross-tab sync *(carried from round 3)*

`packages/ui/src/modules/sessions/lib/draft-key.ts` — not in this delta, still unaddressed. The file defines `SessionDraft` and `draftKey` but has no lifecycle note. With `startDraftSync()` now wired in `main.tsx`, cross-tab sharing is live and concrete: incoming foreign drafts are applied key-by-key, deletions propagate as `null`, and drafts are flushed on `pagehide`/`visibilitychange`. None of this is described anywhere near the primary types.

**Fix:** Add a brief lifecycle note at the top of `draft-key.ts` (or in `draft-snapshot.ts`) describing the cross-tab sync mechanism — that `startDraftSync()` activates it, that foreign writes apply with a per-key last-write-wins rule, and that flushing happens on page hide. (Per repo convention, keep it to a single summary line if prose is unavoidable.)

---

### 🟡 F5 — `stopDraftSync` exported with no call site in the PR *(new, documentation gap)*

`packages/ui/src/store.ts:82` — `stopDraftSync` is part of the module's public API but no caller appears in the delta. Its intended lifecycle (test teardown; not called in production where the listener lives for the tab lifetime) is implicit. This is the narrower issue behind F3.

**Fix:** Ensure test harnesses that exercise `startDraftSync()` call `stopDraftSync()` in teardown. See F3 for the behavioral consequence of omitting it.

---

### What landed cleanly

- `startDraftSync()` is called after `hydrateRoute()` but before `createRoot().render()` — the correct ordering (store initialized before the listener is registered; no `await` between them so no macrotask can slip between hydration and listener attach).
- The idempotency guard (`draftSyncTeardown !== null`) is sound; React StrictMode does not re-invoke `main()`.
- `window.storage` events fire only in other tabs — `onForeignDraftChange` is structurally correct.
- Per-key writes in `draftWriter` mean a stale tab never rewrites another tab's cleared key — the F1 cross-tab clobber from round 2 cannot recur.

---

### Skill audit

| Skill | Trigger | Result |
|---|---|---|
| doc-drift | always | 🟡 2 findings (F4, F5) |
| issue-fit | always | pass |
| typescript-engineering | .ts/.mts/.cts/.js files | 🟡 2 findings (F2, F3) |
| react-ui-engineering | .tsx/.jsx files | 🟡 1 finding (F1 carried) |

---

### Approval bar

Five open 🟡 findings; none are 🔴. F1 (double-read in a hook) and F2 (clobber race on become-visible) are the most impactful and should be resolved before merge. F3, F4, F5 are supporting issues that largely follow from the same lifecycle gap.

```findings-json
<!-- {"findings":[{"id":"F1","severity":"warn","title":"useEffect double-reads droppedAttachmentNames via getState() — carried from round 3","file":"packages/ui/src/modules/sessions/components/chat-input.tsx","line":65,"fix":"Remove const fresh; replace fresh.length with droppedNames.length"},{"id":"F2","severity":"warn","title":"visibilitychange flush fires on become-visible, creating clobber race","file":"packages/ui/src/modules/sessions/lib/draft-snapshot.ts","line":175,"fix":"Guard handler: if (document.visibilityState === 'hidden') writer.flush()"},{"id":"F3","severity":"warn","title":"Module-level draftSyncTeardown causes silent listener leakage in tests","file":"packages/ui/src/store.ts","line":34,"fix":"Call stopDraftSync() in afterEach for any test that calls startDraftSync()"},{"id":"F4","severity":"warn","title":"draft-key.ts lifecycle doc omits cross-tab sync — carried from round 3","file":"packages/ui/src/modules/sessions/lib/draft-key.ts","fix":"Add lifecycle note describing startDraftSync, per-key last-write-wins, flush-on-hide"},{"id":"F5","severity":"warn","title":"stopDraftSync exported with no call site in the PR","file":"packages/ui/src/store.ts","line":82,"fix":null}]} -->
```

PetrBulanek:
Round 4 answered: one taken, four declined. `5ace61e3` closes F5, and carries a defect none of the four rounds found.

## Taken — F5, `stopDraftSync` had no call site (19a01af9)

Correct, and by the same standard that made round 3's finding worth taking. An exported teardown nobody calls is dead code, so it is gone: `startDraftSync()` now guards on a boolean, and the two registrars are called for their effect. F3 goes with it — there is no teardown left to forget.

I did not take the `import.meta.hot?.dispose()` variant. It would have given `stopDraftSync` a call site, but it buys a dev-only path where a `store.ts` edit usually ends in a full reload anyway, and it puts a bundler API in the composition root to do it.

## Declined — F1, `getState()` re-read *(third time)*

Answered on the inline thread 38 minutes before this round was posted; the round repeats it verbatim as "still unaddressed" without engaging. The short version stands: the two statements are adjacent and synchronous, so no race fits between them, and the `getState()` read is what stops `StrictMode`'s second effect invocation from emitting a duplicate toast. The suggested diff ships that duplicate in dev.

## Declined — F2, `visibilitychange` clobber race

The mechanism cannot occur, and the guard proposed to stop it stops no write.

**`flush()` does not serialize in-memory state.** It writes the writer's `queued` map (`draft-snapshot.ts:135-139` after this push) — the keys this tab wrote in the last 300 ms. Not the store, not "stale in-memory state".

**The queue is empty whenever a tab becomes visible.** The same handler flushes on *hide*, and nothing can refill the queue while the tab is hidden: keystrokes need a visible tab, removals bypass the queue and write through at once (`write`, lines 142-150), and `applyForeignDraft` calls `set` directly and never queues a write (`store/sessions.ts:166-176`).

**The guard would not prevent the write.** Suppose an entry were queued. The 300 ms timer is still pending, so the same bytes land whether or not the handler returns early. The guard reorders; it does not protect.

The premise about event ordering is also wrong: a hidden tab receives `storage` events when the write happens, not when it becomes visible, so the two events are not racing in one macrotask.

There *is* a real behaviour nearby, and it is deliberate: with two tabs on the same composer, the tab being typed in wins. That is `platform-topology.md` — "the composer a person is typing in is never overwritten from elsewhere".

## Declined — F3, listener leakage in tests

No test calls `startDraftSync`, and the PR adds no draft tests, so the fix is a rule for tests that do not exist. Removing the teardown (above) removes the hypothesis with it.

## Declined — F4, lifecycle note in `draft-key.ts`

This asks for a comment in a source file. `CLAUDE.md` allows exactly two comment types, `TEST_OVERVIEW:` and `TEST_SCENARIO:`, and neither fits — so "per repo convention, keep it to a single summary line" is asking for the thing the convention forbids. Worth flagging as skill drift: a doc-drift pass should not be pushing prose into code.

"None of this is described anywhere" is also not so. `platform-topology.md:79` carries cross-tab visibility, the per-key rewrite rule, the never-overwrite-the-open-composer rule, and survival across reload and browser close. That page is where a reader looks for how the UI keeps client state; the type file is not.

## Not in the review — drafts outlived the login (5ace61e3)

Four rounds looked at cross-tab, quota, and lifecycle, and all of them missed that `logout()` (`auth.ts:85`) only ran `signoutRedirect()`. Nothing removed a draft key. So unsent text stayed in `localStorage` after sign-out and sat there while the next person used that browser — invisible in their UI, because keys are agent-scoped and an agent that is not theirs is not listed, but present in the store and readable from devtools.

Sign-out now drops pending writes and removes every `platform-draft:` entry, which also collects the orphans I described in round 3 — drafts for an agent the prune never revisits. The removals reach other tabs as `storage` events, so their markers clear too; a tab still holding the same composer open keeps that one draft in memory until it reloads.

`platform-topology.md:79` states the rule at meaning level rather than naming the call: the store belongs to the signed-in person, so signing out empties it.

---

`mise run ui:check` green, `mise run ui:test` 290 passing, `mise run common:check:comment-types` OK. Rounds 1-3 remain closed.

packages/ui/src/modules/sessions/components/chat-input.tsx
dam-code-guardian, line 65: **🟡 F1 — Redundant `getState()` re-read** (carried from rounds 3 and 4)

`droppedNames` is the reactive dep that triggered this effect — it already has length at this point. The `fresh` re-read returns the same value; the `!fresh?.length` guard and the `fresh.length`/`fresh.join` in the message are redundant.

```suggestion
  if (!key || !droppedNames?.length) return;
  emitToast({
    kind: "info",
    message: `Draft restored without ${droppedNames.length} attachment${
      droppedNames.length === 1 ? "" : "s"
    }: ${droppedNames.join(", ")}`,
  });
  consumeDroppedAttachments(key);
```

packages/ui/src/modules/sessions/lib/draft-snapshot.ts
```diff
@@ -0,0 +1,201 @@
...
+  return () => window.removeEventListener("storage", listener);
+}
+
+export function flushDraftsOnHide(
+  writer: DraftWriter = draftWriter,
+): () => void {
+  if (typeof window === "undefined") return () => {};
+  const flush = () => writer.flush();
+  window.addEventListener("pagehide", flush);
+  document.addEventListener("visibilitychange", flush);
```
dam-code-guardian, line 221: **🟡 F2 — `visibilitychange` flush fires on become-visible** (carried from round 4)

This fires on every visibility transition — including when the tab comes back into focus. If another tab wrote a draft while this tab was hidden, the `storage` event and the become-visible `visibilitychange` arrive in the same macrotask queue. Flushing stale in-memory state before `applyForeignDraft` runs clobbers the newer value.

```suggestion
  const flush = () => {
    if (document.visibilityState === "hidden") writer.flush();
  };
```

The `pagehide` listener already covers the unload path; this guard makes become-visible a no-op.

dam-code-guardian (dismissed):
**Code Guardian 🛡️ — complete re-review (round 5)**

Delta reviewed: `5b2e1cdd` → `5ace61e` — 2 files: `packages/ui/src/store.ts`, `packages/ui/src/modules/sessions/lib/draft-snapshot.ts`

---

### Summary

The delta closes F3 (removes `draftSyncTeardown`) and F5 (removes `stopDraftSync`), and adds `draftWriter.clearAll()` to `logout()`. One new 🔴 finding: the PR closes issue #3281, a backend two-phase-write infrastructure ticket — a wrong closing reference that will auto-close active backend work on merge. Two 🟡 findings carry forward.

---

### Changes since last review

Previous HEAD: `5b2e1cdd` (2026-08-17T11:59:51Z) — verdict COMMENT

- ✅ **F3 fixed:** `draftSyncTeardown` removed from `store.ts`; module no longer holds a torn-down listener reference.
- ✅ **F5 fixed:** `stopDraftSync` removed; no dead export.
- ✅ **F4 closed:** `platform-topology.md` now documents cross-tab visibility, the local-wins rule, flush-on-hide, and sign-out clearing. Architecture doc covers the lifecycle; the code file correctly carries no comments.
- **F1 carried:** `useEffect` double-read via `getState()` still present in `chat-input.tsx`.
- **F2 carried:** `visibilitychange` flush still fires unconditionally in `draft-snapshot.ts`.

---

### 🔴 FN — PR closes wrong issue

The PR body contains `Closes #3281`. Issue #3281 is "Two-phase writes for Postgres + Kubernetes operations — a row goes active only after its cluster side effect lands" — a backend infrastructure task with no overlap with this PR's UI-only diff. Merging will auto-close #3281, marking backend work as complete when it has not started.

**Fix:** Replace `Closes #3281` with the correct issue reference. The first commit message uses `Refs #3254`; update the PR description to `Closes #3254` (or whatever the correct closing reference is).

---

### 🟡 F1 — `useEffect` double-reads `droppedAttachmentNames` via `getState()` *(carried from rounds 3 and 4)*

`chat-input.tsx:65` — not in this delta, still unaddressed:

```tsx
const droppedNames = draft.droppedAttachmentNames;   // reactive dep that triggered the effect
useEffect(() => {
  if (!key || !droppedNames?.length) return;
  const fresh = useStore.getState().drafts[key]?.droppedAttachmentNames; // redundant
  if (!fresh?.length) return;
  emitToast({ ... `Draft restored without ${fresh.length}...` });
  consumeDroppedAttachments(key);
}, [key, droppedNames, consumeDroppedAttachments]);
```

The effect already fires because `droppedNames` has length. The `getState()` call reads the same store slice again; the only scenario where `droppedNames` is truthy and `fresh` is falsy is a same-render race that cannot occur in synchronous React. The toast uses `fresh` rather than `droppedNames`, so the guard is redundant and the message diverges from the value that triggered the effect.

**Fix:** Remove `const fresh` and the `!fresh?.length` guard; use `droppedNames` directly in the toast body.

---

### 🟡 F2 — `visibilitychange` flush fires on become-visible, creating a clobber race *(carried from round 4)*

`draft-snapshot.ts:196` — not in this delta, still unaddressed:

```ts
document.addEventListener("visibilitychange", flush);
```

The handler fires on every visibility transition, including when the tab becomes visible again. When Tab A writes a draft while Tab B is hidden, both the `storage` event for that write and the `visibilitychange` for Tab B becoming visible arrive in the macrotask queue. If the `visibilitychange` flush runs first, Tab B serializes its stale in-memory state to `localStorage` before `applyForeignDraft` has updated the Zustand store, clobbering Tab A's newer value.

**Fix:** Guard the handler so it only flushes on hide:

```ts
const flush = () => {
  if (document.visibilityState === "hidden") writer.flush();
};
```

The `pagehide` listener already covers the unload path; this guard makes the become-visible transition a true no-op.

---

### What landed cleanly

- `draftWriter.clearAll()` called synchronously before `userManager.signoutRedirect()` in `logout()` — correct ordering; all in-flight batch writes are discarded and every `platform-draft:` key is removed before the redirect navigates away.
- `discardQueue()` cancels the batch timer before iterating `store.keys()` — no partial write race inside `clearAll()`.
- `draftSyncStarted` boolean guard is sufficient for the one-time setup; React StrictMode does not re-invoke the composition root.
- No code comments added in diff — passes the repo CRITICAL convention.

---

### Skill audit

| Skill | Trigger | Result |
|---|---|---|
| doc-drift | always | aligned |
| issue-fit | always | 🔴 FN (wrong closing reference) |
| typescript-engineering | .ts files | no findings |
| react-ui-engineering | .tsx files | 🟡 F1 (carried) |

---

### Approval bar

One 🔴 (FN — metadata fix; update the closing reference in the PR description before merge) and two 🟡 (F1, F2).

packages/ui/src/auth.ts
```diff
@@ -82,6 +83,7 @@ export function getUser(): User | null {
 }
 
 export async function logout(): Promise<void> {
+  draftWriter.clearAll();
```
kapetr, line 86: If login expires, this doesn't get called and drafts are not cleared.

kapetr (approved):
An edge case - if another is opened with same session it gets the draft from first tab, then if message is sent, the draft message is not deleted in the other tab. But I'd probably keep it as is, since solution wouldnt be simple.

PR #3306, opened 2026-08-13
feat(skills): trim-comments — cut session comments to what carries information

jjeliga:
## What

Adds a `trim-comments` skill: a pass over the comments in code edited during a session, cutting them to the minimum that carries real information. Invocable as `/trim-comments`, or picked up automatically on "trim the comments" / "the comments are too verbose".

Long comments rot — they drift from the code, cost tokens on every read, and assert things that were only true on an earlier branch.

## The rules it encodes

- **One line per comment. Delete is the default**; a rewrite is the exception. Zero comments left in a file is a fine outcome. Code is never changed in the pass.
- **Deletes**: restatement of the code, any fact already carried by a name / type / schema / test name / architecture page, within-branch reasoning ("used to", "instead of", "per review"), history pointers (SHAs, PR numbers, branch names, ADR references), issue numbers, bloat and hedging, commented-out code, TODO/FIXME markers, and claims it cannot verify.
- **Keeps**: non-obvious *why* where a "cleanup" would reintroduce a bug, doubt at a call site, facts a type cannot state (units, invariants, lock discipline), upstream spec/bug anchors, and doc comments a lint rule requires.
- **Issue numbers** are barred from production code, with one exception in test files: a regression test may name the issue it locks down when the bug was crucial.
- **Never touches directives** — license headers, shebangs, `//go:generate`, `// nolint`, `// eslint-disable*`, `// biome-ignore`, `// @ts-expect-error`, codegen markers. Deleting one of these breaks a build.
- **Terseness never costs information**: the line cap yields to information, so a crucial fact that genuinely will not fit keeps its longer comment, flagged in the report. Never a silent loss.
- **Writes no docs.** Subsystem behavior a deleted comment explained that no architecture page covers is reported as a doc gap, not authored here — read-only toward docs, like `doc-drift` and `adr-policy`.

Scope is limited to files edited in the session, and to comments the session authored or that sit inside a hunk it changed. Violations elsewhere in the same file are reported rather than rewritten, so the diff stays reviewable. Identifying the edit set is left to the agent: the skill states the requirement (resolve the baseline explicitly, fail loudly, declare a branch-wide set as branch-wide) without prescribing commands it cannot verify at runtime.

## Review round

Three findings from `dam-code-guardian`, all legitimate, fixed in `dd065695`:

- Relocating overflow facts to `docs/architecture/` contradicted the destination: those pages omit code-level shape by policy, which is exactly what the Keep list describes. Mechanism dropped rather than retargeted — `doc-size.mjs:53` shows the repo had already rejected spillover buckets.
- The same step deleted before it wrote, so a size-cap denial (the top five pages sit 14–414 characters under `CAPS.page`) lost the fact and still reported success. Dissolved by the above.
- The scope fallback's bare `main` ref failed silently in worktrees and single-branch clones. Commands removed in favour of stating the requirement.

Follows the existing skill layout: `SKILL.md` under `.agents/skills/`, symlinked from `.claude/skills/`.


jezekra1 (approved):

.agents/skills/trim-comments/SKILL.md
```diff
@@ -0,0 +1,115 @@
...
+were true on an earlier branch. This pass deletes them and keeps the few that stop a reader from
+breaking something.
+
+**Default action is delete** — a rewrite is the exception, and a rewrite is one line. Zero
+comments left in a file is a fine outcome. Never change code in this pass; comments only.
+
+## Scope
+
+- **Files** written or edited this session. If that set is unclear, reconstruct it:
+  `git status --short` plus `git diff --name-only $(git merge-base HEAD main)`.
```
dam-code-guardian: 🟡 **Warning:** This fallback fails in both directions. `main` is a bare local branch name, which does not resolve in a fresh `--single-branch` clone or a `.worktrees/` checkout — and command substitution swallows that: `merge-base` writes to stderr, contributes an empty argument, and `git diff --name-only` degrades to worktree-vs-index at exit 0, so every committed edit of the session drops out of scope silently. When `main` *does* resolve, the same command returns every file changed since the base, which is the whole branch rather than the session. Reproduced in a scratch repo: with no local `main` the session's committed `f1.txt` disappears from the list; with `main` present an unrelated `base.txt` from an earlier commit is added.

**Fix:** A scope fallback must resolve its baseline explicitly and never exceed the scope the skill declares — try `origin/main` before `main` and fail loudly when neither resolves (`scripts/adr-immutable.mjs:70` is the in-repo precedent), and bound the reconstruction to the session's own edits rather than the branch's.

.agents/skills/trim-comments/SKILL.md
```diff
@@ -0,0 +1,115 @@
...
+   not an issue number.
+6. **Bloat** — comments spanning paragraphs, hedging ("basically", "note that"), tutorials on
+   the language or framework, ASCII diagrams, section banners, decorative dividers.
+7. **Dead weight** — commented-out code, and TODO/FIXME the agent invented unasked.
+8. **Unverifiable claims** — if you cannot confirm what a comment asserts, delete it or flag it.
+   Never restate it, and never write a "why" you cannot see in the code.
+
+## Keep
+
+Only these earn a line, and only when the code cannot carry the fact itself:
```
dam-code-guardian, line 64: 🟡 **Warning:** "Only these earn a line" makes the Keep list closed, and no entry covers a TODO the user asked for, while Delete rule 7 (line 55) removes only the TODOs "the agent invented unasked". A run that follows the list literally deletes a user-requested TODO — it matches no Keep entry, and the closed wording gives no room to spare it. The PR body names this as a deliberate judgment call ("Self-invented TODOs are deleted, while TODOs the user asked for survive"), so the intent is clear and only the rule is missing.

**Fix:** Every behavior the PR body claims must be stated as a rule in the skill body — add a Keep entry for a user-requested TODO, or drop the claim from the description.

.agents/skills/trim-comments/SKILL.md
```diff
@@ -0,0 +1,115 @@
...
+// after:
+// Sequential: the broker rejects overlapping publishes on one connection.
+```
+
+## Process
+
+1. **Read each edited file in full.** Duplication is invisible from a diff alone.
+2. **Classify** every in-scope comment: delete, one-line rewrite, keep verbatim.
+3. **Apply** with Edit. Leave no orphan blank lines, dangling `/*`, or empty doc blocks.
+4. **Rescue** any load-bearing fact that will not fit in one line onto the subsystem's page
```
dam-code-guardian: 🟡 **Warning:** The rescue step is the guarantee behind "Terseness never costs information", and it cannot run for the five pages most likely to own a rescued fact. Measured in the gate's own unit (characters, `scripts/doc-size.mjs:38`) against `CAPS.page = 40000`: `agent-lifecycle.md` has 14 characters of headroom, `connections.md` 28, `security-and-credentials.md` 313, `channels.md` 324, `skills.md` 414. `scripts/doc-size-hook.mjs` is a registered `PreToolUse` hook (`.claude/settings.json:65`) that **denies** the Write/Edit outright. Because step 3 applies the deletes before step 4 rescues, the denial arrives after the comment is already gone — the fact is lost, and the pass still reports success.

**Fix:** Any step that relocates content must land the destination write before the source is deleted, and must state what to do when the destination refuses — reorder so the rescue is applied and verified first, and on a cap denial stop and report rather than proceeding with the delete.

.agents/skills/trim-comments/SKILL.md
```diff
@@ -0,0 +1,115 @@
...
+   under [`docs/architecture/`](../../../docs/architecture/) — never back into the code as
```
dam-code-guardian: 🟡 **Warning:** Two rules of the destination contradict what is being sent there. The facts this skill rescues are exactly the ones its own Keep list describes as code-level — units, ranges, "caller holds the lock", a required ordering — and `docs/guidelines/documentation-guidelines.md:39` tells architecture pages to **omit** "exact package names, file paths… and code-level shape — type signatures, field names, function arguments, enum members", raising anything that "would break when someone renames a field". Those pages are also durable-content-only and carry a mandatory `Last verified: YYYY-MM-DD` header "bumped whenever you edit the page" (`documentation-guidelines.md:34`), which step 4 never mentions — the same missed-bump class this repo has hit before on prose-only edits.

**Fix:** A relocation target must accept the kind of fact being moved and satisfy that page's own edit rules — either send code-level invariants somewhere that admits them (the owning package's doc comment or a reference page) rather than `docs/architecture/`, or state the level to raise the fact to and require the `Last verified:` bump as part of the step.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `2aa5da5`

## PR #3306: feat(skills): trim-comments — cut session comments to what carries information
**Author:** jjeliga | **Branch:** feat/trim-comments-skill → main | **Changes:** +116 −0 (2 files)

### Summary
Adds a `trim-comments` skill — one `SKILL.md` plus the conventional `.claude/skills` symlink — that cuts comments in session-edited files to one informative line each. The delete/keep rules and the directive protections are well chosen; the four findings sit in the scope fallback and in the escape hatch that is supposed to preserve a fact the pass deletes.

### Findings
- 🟡 **Warning:** Scope fallback resolves no base ref and widens past the session (`.agents/skills/trim-comments/SKILL.md:25`)
- 🟡 **Warning:** Rescue target is at its cap and hook-denied, after the delete already applied (`.agents/skills/trim-comments/SKILL.md:105`)
- 🟡 **Warning:** Rescued facts are what the architecture content policy omits; no `Last verified:` bump (`.agents/skills/trim-comments/SKILL.md:106`)
- 🟡 **Warning:** Closed Keep list deletes the user-requested TODOs the PR body promises survive (`.agents/skills/trim-comments/SKILL.md:61`)
- ✅ **Looks good:** The "Never touch" list covers every pragma class actually present in this tree — `eslint-disable*` in four hand-written UI hooks and in `keycloak-theme/src/login/i18n.ts`, `@ts-expect-error`, Go codegen markers and build tags. The classes it omits (`@ts-nocheck`, `noinspection`, and this repo's own TS codegen wording) all sit in generated files, which Scope already excludes.

✅ No findings.

### Issue Fit

- 🟡 Two findings merged into `### Findings` above: the user-requested TODO gap (`SKILL.md:61`) and the branch-wide scope fallback (`SKILL.md:25`, merged with the diff review's ref-resolution half).
- 🟢 No undeclared scope creep — the diff is exactly the two files the body names, and the symlink target matches the `../../.agents/skills/<name>` convention used by every other skill.

### Verdict
COMMENT — the delete/keep rules and directive protections are sound, but four 🟡 sit in the scope fallback and the rescue step, where a load-bearing fact can leave silently.

---

.agents/skills/trim-comments/SKILL.md
jjeliga: we do not do todos, we file issues

.agents/skills/trim-comments/SKILL.md
jjeliga: Half accepted, in `dd065695`.

The ref bug is real: bare `main` does not resolve in a `.worktrees/` checkout (which CLAUDE.md mandates) or a single-branch clone, and the failure is silent — the substitution yields no argument, `git diff --name-only` degrades to worktree-vs-index and exits 0, so committed session edits vanish from scope.

The fix is not the ladder, though: the commands are gone entirely. Prescribing shell here was the actual mistake — identifying its own edit set is the agent's job, and a skill that hardcodes it owns a correctness burden it cannot verify at runtime. Scope now states the requirement instead: resolve the baseline explicitly, fail loudly rather than guessing, and when the best available set is branch-wide rather than session-wide, declare that in the report.

Rejecting the second half — "bound the reconstruction to the session's own edits" is impossible by construction, since the fallback only runs when that set is unknown. Naming the wider scope is the honest fix; pretending to precision is not.

.agents/skills/trim-comments/SKILL.md
jjeliga: Confirmed and fixed in `dd065695`, by removing the step rather than reordering it.

Verified your numbers independently: `measure` is `content.length` (`scripts/doc-size.mjs:38`), `CAPS.page = 40000`, and the five pages most likely to own a rescued fact sit 14–414 characters under it. The hook denies the write, and the delete had already landed by then — a silent loss reported as success. Correct.

Reordering would have fixed the ordering and left the destination wrong (see the thread on line 106), so the relocation mechanism is gone: a crucial fact that will not fit one line keeps its longer comment, and the skill writes no docs at all.

.agents/skills/trim-comments/SKILL.md
jjeliga: Accepted — this was the root cause, fixed in `dd065695`.

You are right that the destination forbids exactly what was being sent there: `documentation-guidelines.md:39` omits code-level shape, which is precisely what the Keep list describes (units, ranges, "caller holds the lock"), and the mandatory `Last verified:` bump went unmentioned.

One point in your favour that the finding did not use: `scripts/doc-size.mjs:53` already says the steering "never tells the agent to reword to fit, and never to mint an ADR as a spillover bucket". The repo had already rejected spillover buckets, and the rescue step made `docs/architecture/` into one.

So relocation is dropped rather than retargeted. The line cap now yields to information — a crucial fact that genuinely will not fit keeps its longer comment, flagged in the report — and subsystem behavior no page covers is reported as a doc gap, never written. Keeps the skill read-only toward docs, like doc-drift and adr-policy.

PR #3308, opened 2026-08-13
chore: strip comments from ts/js/go sources

### LINKED ISSUE #3010: Stop agents from writing code comments by default
## Context

From a discussion in the team Slack channel about code comments and AI.

The starting point was that newer models (Opus 5 context-engineering guidance) have moved away from a hard "don't comment" rule toward a soft **"write code that reads like the surrounding code: match its comment density, naming, and idiom."** In practice models still emit comments a lot, because the training prior toward commenting is strong and a soft instruction doesn't override it.

Team consensus in the thread: we'd rather have **no agent-authored code comments**. Reasoning raised:

- **Documentation rot** — comments drift out of sync with the code.
- **Kills agent exploration** — comments short-circuit the curiosity to actually read/understand the code.
- We already capture the *why* in **ADRs**, which is the better home for decisions/rationale.

## Proposal

1. Add a hard guardrail to the repo agent config (`CLAUDE.md` / `AGENTS.md`): e.g. *"Do not write code comments unless the user explicitly asks. Rationale and decisions belong in ADRs, not inline comments."* A hard rule holds far better than relying on the model's default judgement.
2. (Optional / follow-up) Strip existing non-essential code comments from the codebase.

## Open questions

- Scope of the strip: all packages, or start with a subset?
- Keep exceptions for things like license headers, `// eslint-disable`, and genuinely non-obvious one-liners?

_Filed on behalf of the team following the Slack discussion._

tomkis:
Closes #3010.

Strips all comments from TS/JS/Go sources and adds the script that does it (scripts/strip-comments.mjs).

What it keeps: tool directives only. //go: markers, // +kubebuilder, //nolint, @ts-expect-error, eslint-disable, prettier-ignore, triple-slash references. Generated files are skipped.

How it works: TS/JS is lexed with the TypeScript parser, Go with a small dedicated lexer, so strings, template literals, regex literals and JSX text are safe. Every stripped TS file must re-parse to the identical token stream or it is left untouched.

Verification done:
- mise run check passes, all 62 tasks (tsc, eslint, go vet/build, CRD schema generation, helm render)
- every modified TS/JS file parses to exactly the HEAD token stream minus comments
- rerunning the script reports 0 remaining strippable comments

Scope: ts/tsx/js/mjs/go only. Shell, YAML, SQL, Python and TOML comments are untouched.

Draft on purpose: removing every comment repo-wide is a policy call, not a mechanical one.


jjeliga (dismissed):
Remove the now redundant "trim-comments" skill.

tomkis:
Dropped the trim-comments skill (.agents/skills/trim-comments plus the .claude symlink) in 2f807fcd. The typed-comment ruleset in CLAUDE.md and common:check:comment-types replace it.

jezekra1 (approved):

PR #3313, opened 2026-08-13
style(ui): drop the blue from inline code in agent responses

### LINKED ISSUE #3209: UI - Inline code formatting is hard to read
## Context

Agent responses use inline code formatting for many different things: real code, UI labels, status values, file paths, product names, plain nouns. The style is also blue. In a long response the result is a page of blue boxes. Raised in Slack with a screenshot.

## Problem

The formatting does two jobs and does neither well.

- **It is mapped onto too many things.** When almost every noun is marked, the mark stops meaning anything. It becomes noise instead of emphasis.
- **The blue looks like a link.** Users cannot tell what is clickable, and the colour pulls the eye away from the sentence.

Agent responses are the surface users read most. A response that is hard to read makes the whole product feel harder to use.

## Goal

Inline code formatting has a small, clear purpose and a style that supports reading instead of interrupting it.

## User Stories

- As a user, I want inline code formatting only on things that are code, so that the emphasis tells me something.
- As a user, I want to read a long response without visual noise, so that I can find the important parts quickly.
- As a user, I want links to look different from code, so that I know what I can click.

## Proposed solution

Two parts:

1. **Agree what inline code is for** — code, file paths, commands. Everything else (UI labels, status values, ordinary nouns) gets a lighter treatment or none.
2. **Restyle it** — drop the blue. A neutral or muted treatment keeps the distinction without competing with links.

Some of this formatting comes from the agent's own markdown, not only from the CSS. Guidance on how agents write may be part of the fix.

## Open Questions

- What do UI labels and status values get instead — bold, quotes, plain text?

## Additional resources

- Slack discussion: https://ibm-research.slack.com/archives/C0B8DA70GE4/p1785979662204129 (screenshot in thread)

---
_Filed by @jamiejabbouribm_

kapetr:
Inline code in agent responses was blue and boxed, so a long answer read as a page of blue chips and the marks looked clickable. It is now neutral text on the page background, outlined only.

Closes #3209

Values come from the design: the border is `--border-ui`, the fill is the page background showing through, the text is `--foreground`, with a 4px radius and slightly wider padding. Font family and size are unchanged, and dark mode follows the same tokens.

Styling only, per the scope note on the issue. Deciding what inline code is *for* — and how agents should write it — is not touched here, and the open question about UI labels and status values is still open.

### DIFF
--- a/packages/ui/src/App.css
+++ b/packages/ui/src/App.css
@@ -475,11 +475,11 @@
 .prose code {
   font-family: var(--font-mono);
   font-size: 0.86em;
-  background: var(--muted);
+  background: transparent;
   border: 1px solid var(--border-ui);
-  border-radius: 6px;
-  padding: 0.1em 0.36em;
-  color: var(--c-accent-hover);
+  border-radius: 4px;
+  padding: 0.15em 0.5em;
+  color: var(--foreground);
 }
 .prose pre {
   background: var(--muted) !important;


packages/ui/src/App.css
```diff
@@ -475,11 +475,11 @@
...
   font-size: 0.86em;
-  background: var(--muted);
+  background: transparent;
   border: 1px solid var(--border-ui);
-  border-radius: 6px;
-  padding: 0.1em 0.36em;
-  color: var(--c-accent-hover);
+  border-radius: 4px;
+  padding: 0.15em 0.5em;
+  color: var(--foreground);
```
dam-code-guardian, line 482: 🟡 **Warning:** Inline code inside a link now renders as plain link-coloured text in a box. This rule sets `color: var(--foreground)`, but `.prose a` (line 467) is more specific for the anchor's own text and no `.prose a code` rule exists — so a ``[`path/to/file`](url)`` link keeps the blue while the same token in a sentence turns neutral. The issue's third user story is "links look different from code"; the change makes them differ everywhere *except* where they overlap, which is the shape the repo's own docs use heavily.

**Fix:** When a rule restyles an element that can nest inside a differently-coloured one, add the descendant rule for the overlap — here `.prose a code { color: inherit; }` (or an explicit accent) so code-in-link takes one deliberate colour instead of inheriting whichever selector wins.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `7c5a76e`

## PR #3313: style(ui): drop the blue from inline code in agent responses
**Author:** kapetr | **Branch:** style/3209-inline-code → main | **Changes:** +4 −4 (1 files)

### Summary
The `.prose code` rule drops the `--muted` fill and the blue `--c-accent-hover` text for a transparent fill and `--foreground` text, with a 4px radius and wider padding. This matches the maintainer's styling-only scope note on the issue, and dark mode follows the same tokens.

### Findings
- 🟡 **Warning:** Inline code inside a link keeps the blue — see the inline comment (`packages/ui/src/App.css:482`)
- 🟢 **Suggestion:** The `1px solid var(--border-ui)` outline is now the sole mark, and it is very low contrast against tinted containers: 1.19–1.20:1 on light `blockquote`/`th`/`tr:hover` backgrounds and 1.31–1.42:1 in dark. The old fill was equally faint, so this is not a regression, but the mark's whole job now rests on that one border. (`packages/ui/src/App.css:479`)

✅ No findings.

### Issue Fit
- 🟡 **Warning:** `Closes #3209` will close the issue on merge, but only part 2 of its two-part ask ships — part 1 ("agree what inline code is *for*") and the still-open question on UI labels/status values are untouched, as the PR body itself states. The styling-only scope is legitimate (verified: maintainer scope note on the issue, 2026-08-07 — "Update only the styling of inline code formatting"), so the defect is the closing keyword, not the diff. (`PR body: Closes #3209`)
  **Fix:** A PR delivering a re-scoped subset of an issue references it without a closing keyword (`Refs #3209`) and leaves it open for the remaining items, or files a follow-up issue for them and links it — a closing keyword is only for a PR that discharges everything the issue still owes.

### Verdict
COMMENT — the restyle is correct and on-scope; two 🟡 to settle first (code inside a link keeps the blue, and the closing keyword over-claims a re-scoped delivery).

---

Tomas2D (approved):

PR #3314, opened 2026-08-13
chore(usage): extend and correct raw interaction collection

### LINKED ISSUE #3259: Update and document raw interaction data collection
## Why

Our interaction data tells us HOW users work with the platform. The collection mechanism has not been updated in some time, so it no longer covers the current platform surface, and we have no written record of what it does collect. 

This ticket is purely technical: make the collection correct and complete, and write down what it captures. It makes no claims about what the data means or what we should do with it.

## Scope

- [x] Inventory every event currently emitted; identify events that are
      broken, stale, or no longer firing
- [x] Fix broken events
- [x] Extend coverage to the current platform surface (see stopping rule)
- [ ] Document each event: name, trigger, etc.

**Stopping rule for "extend":** every user-facing component emits at least
basic usage events. This is a mechanical completeness criterion, not a judgement about which events are useful. 

## Explicitly not in scope

This ticket does not decide, and must not be blocked on deciding, any of the
following:

- What questions the data should answer, or what any of it means
- Which events are worth keeping or which are useless
- New views, dashboards, or report-page changes
- Direct user feedback and activity auto-summarization — separate data
  streams, tracked separately
- Any product or roadmap conclusion

## Definition of done

Every user-facing component in the current platform either emits events or is listed with a reason why not. The document is published and is accurate enough that someone can tell, without reading code, whether a given user action is captured.

jjeliga:
Extends activity tracking to the surfaces it never covered, fixes the defects that made several report views answer nothing, and adds a gate so the event registry cannot half-drift again.

closes #3259

## Correctness

Four of these made live views answer the wrong thing, silently:

- **Connection events carry their provider.** `connectionKey` is a random per-grant id and the Connection record is hard-deleted on disconnect, so grouping by it produced one row per grant with `distinct_users=1` forever. `usage_connections_by_key` becomes `usage_connections_by_provider`, backfilled for connections that still exist; rows whose Connection is already gone read as `unknown` — visibly incomplete rather than a silent undercount.
- **Connect fires for every auth kind**, not just OAuth callbacks. Non-OAuth connections were invisible, and churn could show more removes than adds.
- **Two views mixed raw and pseudonymized sub spaces.** `activity_events` / `actor_roles` / `agents` hold hashed subs; `pending_approvals.owner_sub` and `skill_sources.owner` hold raw ones. The approvals core-team filter compared a raw sub against the hashed core set, which can never match — so core-team approvals were counted in a view whose whole contract is that they are not. `usage_skill_installs_by_user` selected a raw Keycloak sub straight into the inspector report, which is exactly what this subsystem promises never to expose. Both re-keyed onto single-space identifiers; the skill view now also attributes to the installer rather than the source curator.
- **Messenger-native actor ids are pseudonymized** at the repository chokepoint. They name a person just as a Keycloak sub does; the raw value stays in the audit trail, which is the stream that needs real identity.

Plus: the core-team filter reaches the egress view (the last one without it), schedule fires drop an always-null session, and the auth and schedule columns are renamed to what they actually hold — active days rather than logins, trigger delivery rather than the agent's work.

## Coverage

A conversation held in Slack produced an Activity Event; the identical conversation in the browser produced nothing. The platform's most-used surface was its least visible one.

Now recorded: session turns from the shared relay (browser chat and CLI, same outcome contract as channel turns so both compare in one column), relay attaches, import surface, the artifact library (publish, share, delete, view), skills (install/uninstall with origin, publishes, saved selections), and the six remaining modules — Kinded Agent creates, experiment lifecycle, invocations, feature flags, harness config, API keys.

Every composition reaching these services now states its surface, so an agent acting in its own workspace never reads as a person at a keyboard.

Instrumenting interactions that already leave durable state is deliberate: **state tables key on raw subs, the activity log on pseudonymized ones.** A raw-sub table cannot be filtered against the core-team set and cannot be shown to an inspector without exposing an identifier. Routing an interaction through an event is what puts it in the one space where it can be both joined and read safely.

## Mechanism

Connection add/remove and skill install/uninstall are each folded into one event with a direction discriminator, still landing as distinct indexed row types so counting stays an index read rather than a JSONB scan.

`AgentUpdated`, `AgentRestarted` and `AgentWoken` are deleted — all three fired and nothing subscribed. Each emit site sat beside a `securityLog` call, so those operations remain in the audit trail.

New `api-server:check:events` fails when a registry type has no emit site, or nothing that consumes it. Both halves drift past tsc and lint because both states are well-typed, and an unsubscribed event is indistinguishable from working instrumentation at the emit site. It walks the TypeScript AST rather than grepping: the file list comes from the package tsconfig so it cannot miss a new directory, and parsing means a reference in a comment or string literal is not a match. It keys on emission, which has a single chokepoint, rather than subscription, which has several legitimate forms — a saga spelling its subscription a new way would otherwise fail a correct build.

**What the gate does not cover:** a module that never reaches the bus at all, which is the failure that produced these gaps. Nothing mechanical catches that without a hand-maintained module list, which goes stale and can be satisfied without collecting anything, so the gate is scoped to the invariant it can actually hold.

## Not covered, deliberately

`e2e` (test-only), `approvals` (durable table with a view in the report), and `terms` (a one-time legal fact that fits the consent work better) have mutations but no event.

**`egress-rules` stays uncollected — now as a decision, not an omission.** Four mutations and a durable table, but rule edits are an administrative action rather than product usage, and observed traffic is already answered by `usage_egress_hosts_by_agent` (a different question). Decided during review; the rationale sits in the coverage gate's `NOT_TRACKED` list and surfaces in the generated `docs/activity-events.md`, so it cannot silently rot. Easy to add later if wanted.

Agent tool usage and UI navigation were both explicitly set aside.

## Verification

- Migrations and every changed view run against a throwaway Postgres: full chain applies, each view returns the expected rows. The two pseudonymity fixes were verified by reproducing both defects first — the core-team approval appeared and a raw sub was returned; after, the approval is excluded and the sub is hashed.
- The event gate verified by mutation: a new registry entry with nothing wired, a subscriber removed from a live event, a last emit site removed, and decoy references in a comment and a string literal are each reported; the restored tree passes.
- tsc, lint, format, `common:check:comment-types`, all four `db:check` tasks, both docs checks, and 920 tests across 102 files.

Comments are stripped per the new policy. The rationale from the check script's header moved into `docs/architecture/usage-tracking.md`, since it is what someone needs to modify the gate safely.


packages/db/drizzle/0024_usage_view_fixes.sql
```diff
@@ -0,0 +1,268 @@
...
+DROP VIEW IF EXISTS "usage_auth_users_7d";--> statement-breakpoint
+DROP VIEW IF EXISTS "usage_auth_surface_by_user";--> statement-breakpoint
+DROP VIEW IF EXISTS "usage_auth_by_source_7d";--> statement-breakpoint
+DROP VIEW IF EXISTS "usage_auth_by_source_day_7d";--> statement-breakpoint
+DROP VIEW IF EXISTS "usage_multi_surface_users";--> statement-breakpoint
+
+CREATE VIEW "usage_auth_users_7d" AS
+  SELECT
+    actor_sub,
+    MIN(occurred_at) AS first_seen,
```
dam-code-guardian, line 167: 🟡 **Warning:** The rename to `active_days` does not match what the aggregate counts. The dedup index is unique per (`actor_sub`, `surface`, UTC day), so one user active on `ui` and `cli` on the same day writes two rows and `COUNT(*)` returns 2 for a single active day. The sibling at `:214` (`usage_multi_surface_users.active_days_total`) is worse, because that view's `HAVING COUNT(DISTINCT surface) > 1` means only multi-surface users appear — so its count is *always* inflated. `usage_auth_surface_by_user:171` is correct, since its GROUP BY pins one surface. The report renders columns generically, so the label is the reader's entire contract — this is the same trap the migration set out to fix, one level down.

**Fix:** Name a count for the grain its GROUP BY actually produces. Where the column claims days, count days — `COUNT(DISTINCT date_trunc('day', occurred_at AT TIME ZONE 'UTC'))` — or keep `COUNT(*)` and call it `user_days`, as `usage_auth_by_source_7d:186` already does correctly.

packages/api-server/src/apps/api-server/agent-proxies/acp-relay.ts
```diff
@@ -159,6 +166,30 @@ export function createAcpRelay(
...
+        if (isPrompt(parsed)) inFlight.add(parsed.id);
+      }
+
+      function drainInFlight(reason: string): void {
+        for (const _ of inFlight) recordTurn("failure", reason);
+        inFlight.clear();
+      }
+
+      client.once("close", () => drainInFlight("client-gone"));
+
```
dam-code-guardian: 🟡 **Warning:** A prompt sent before the upstream socket opens is never added to `inFlight`, so a client that disconnects while the agent wakes records no turn at all. Messages arriving early are buffered into `pending` (`:206`); `trackIfPrompt` runs only at flush time (`:243`); and the close handler at `:211` sets `clientGone` and clears `pending` first, so `drainInFlight("client-gone")` iterates an empty set. The wake window is exactly where a client waits longest, and `docs/architecture/usage-tracking.md:86` states that an open turn is recorded rather than dropped "since silently losing those would undercount precisely the turns that ran longest".

**Fix:** Track a turn when the request is accepted from the client, not when it is forwarded upstream — any queue that sits between accept and forward must drain into the same failure path on disconnect.

packages/api-server/src/modules/invocations/services/invocations-service.ts
```diff
@@ -160,6 +161,12 @@ export function createInvocationsService(deps: {
         expiresAt,
         experimentSpanId: input.experimentSpanId ?? null,
       });
+      emit({
+        type: EventType.InvocationSpawned,
+        targetAgentId: targetId,
+        driverAgentId: input.driverAgentId,
+        ownerSub: deps.owner,
+      });
```
dam-code-guardian: 🟡 **Warning:** The event is emitted before the target agent exists, so a failed spawn counts as a spawn. `agents.create` follows at `:172`, and its `catch` deletes the invocation row (`:186`) — but the event has already reached the bus and the `activity_events` row is already written. `ExperimentChanged`/`started` has the same shape at `experiments-service.ts:360`, emitted before `launchRun`, which can move the run to `failed` and throw.

**Fix:** Emit a domain event only after every operation the event asserts has succeeded — place the emit after the last failable step, or carry an `outcome` field the way the turn and schedule-fire events already do.

packages/api-server/src/modules/connections/services/connections-service.ts
```diff
@@ -665,6 +667,16 @@ export function createConnectionsService(deps: {
...
+      if (built.auth.kind !== "oauth") {
+        emit({
+          type: EventType.ConnectionChanged,
+          action: "added",
+          actorSub: deps.ownerId,
+          connectionKey: id,
+          templateId: template.id,
+          kind: template.category === "mcp" ? "mcp" : "oauth_app",
+        });
+      }
```
dam-code-guardian, line 679: 🟡 **Warning:** The connect/disconnect asymmetry is not fully closed for OAuth. This add is skipped when `built.auth.kind === "oauth"` and fires instead when the callback lands (`oauth-flow.ts:149`) — but the connection row already exists at that point: `deriveStatus:832` returns `"pending"` for an OAuth connection with no `connectedAt` and no `expiresAt`, and `deleteConnection:381` has no status gate, so a user who abandons the authorization can delete that row. The removal at `:459` always emits. So `removes` can still exceed `adds`, which `0024_usage_view_fixes.sql:99` states is fixed.

**Fix:** Emit paired lifecycle events from the same state transition the row's existence depends on, so every row that can be removed was recorded as added — for OAuth, either emit at creation and treat the callback as a state change, or suppress the removal event for a connection that never reached connected.

packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts
```diff
@@ -387,12 +398,25 @@ export function createArtifactLibraryService(
         patch.expiresAt = expiresAtFrom(input.expiresInHours);
       const updated = await repo.updateArtifact(id, owner, patch);
       if (!updated) throw new TRPCError({ code: "NOT_FOUND" });
+      emit({
+        type: EventType.ArtifactShared,
+        actorSub: owner,
+        artifactId: id,
+        visibility: updated.visibility,
+        surface,
+      });
```
dam-code-guardian, line 426: 🟡 **Warning:** This fires on every `setSharing` call, including one that carries only `expiresInHours` and leaves `visibility` at `private` (both fields are independently optional, `:396`–`:398`). The share metric therefore counts artifacts that were never shared, and the emitted `visibility` field records the unchanged value rather than an act of sharing.

**Fix:** Emit a state-change event only when the state actually changed — compare the prior row's `visibility` against `updated.visibility` and skip the emit when they match.

packages/api-server/src/modules/experiments/services/experiments-service.ts
```diff
@@ -355,6 +357,13 @@ export function createExperimentsService(
         executedAt: now(),
         lastActivityAt: now(),
       });
+      emit({
+        type: EventType.ExperimentChanged,
+        action: "started",
+        experimentId: runId,
+        actorSub: owner,
+        surface: deps.surface,
+      });
```
dam-code-guardian: 🟡 **Warning:** This emit precedes `launchRun(runId)` on the next line, which can transition the run to `failed` and throw — so a run that never launched is counted as started. Same class as `InvocationSpawned` in `invocations-service.ts:164`; both are the only two new events emitted before their operation's last failable step.

**Fix:** Emit a domain event only after every operation it asserts has succeeded, or carry an `outcome` field as the turn and schedule-fire events do.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `a303e99`

## PR #3314: chore(usage): extend and correct raw interaction collection
**Author:** jjeliga | **Branch:** chore/3259-interaction-data-collection → main | **Changes:** +6917 −109 (57 files)

### Summary
The PR corrects four usage views that answered the wrong question, pseudonymizes two raw-sub leaks, extends activity collection to the browser/CLI relay and twelve more modules, and adds an AST-based gate that fails a build when a registry event has no emit site or no consumer. The two migrations, the surface plumbing, and the gate's own predicate all verified correct; the findings sit in event *timing* and in one column rename.

### Findings
- 🟡 **Warning:** `active_days` counts rows, not days (`packages/db/drizzle/0024_usage_view_fixes.sql:158`)
- 🟡 **Warning:** a prompt buffered during agent wake records no turn (`packages/api-server/src/apps/api-server/agent-proxies/acp-relay.ts:211`)
- 🟡 **Warning:** `InvocationSpawned` fires before the target agent exists (`packages/api-server/src/modules/invocations/services/invocations-service.ts:164`)
- 🟡 **Warning:** OAuth adds are still not paired with removes (`packages/api-server/src/modules/connections/services/connections-service.ts:670`)
- 🟡 **Warning:** `ArtifactShared` fires when nothing was shared (`packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts:401`)
- 🟡 **Warning:** `ArtifactPublished` counts machine-written artifacts (`packages/api-server/src/modules/experiments/services/experiments-service.ts:439`)
- 🟡 **Warning:** `surface` is set untyped, so a missed setter becomes a NULL grouping key (`packages/api-server/src/apps/api-server/admission/auth-middleware.ts:34`)
- ✅ **Looks good:** Both pseudonymity fixes key on single-space identifiers, and the `hashSub` chokepoint covers the new `externalActorId` / `ownerSub` payload fields. The gate's `emit`-callee predicate holds against all 71 real emit sites in the tree, and the `templateId` backfill's unrecoverable-population argument checks out against the hard-delete.

- 🟡 **Warning:** `docs/architecture/skills.md:145` and `:147` say Create Local and Delete Local record **nothing** in Postgres. Both operations now emit `AgentSkillChanged`, so the persist-activity saga writes an `activity_events` row — and `usage-tracking.md:92` presents that row as the only trace of an authored Local Skill, so the two pages now contradict each other. Both sentences are byte-identical to `main`, so this drift is new in this PR.
  **Fix:** When a module starts emitting a domain event that a saga persists, update that subsystem's page in the same PR. Retract every "records nothing" claim that the new instrumentation makes false.
- 🟡 **Warning:** `docs/architecture/artifact-library.md:52` still presents the per-artifact view count as the only thing a public render records. Publish, share, delete and view now emit domain events, so a public render also writes a timestamped activity-log row carrying the owner's pseudonymized sub.
  **Fix:** When a subsystem gains event emission, its own page must record what each interaction now persists. A page that enumerates what an interaction records is incomplete until it names the activity-log row too.

### Issue Fit
Linked issue: #3259 "Update and document raw interaction data collection" (from the plain `Refs #3259`; no closing keyword).

- 🟡 **Warning:** The published document does not document each event. It states that the code is the enumeration ("deliberately not restated here"), so a reader cannot tell without reading code whether a given user action is captured — `docs/architecture/usage-tracking.md:96`. Issue #3259 makes exactly that readability its Definition of Done.
  **Fix:** When an issue asks for a written record of what is collected, the document must carry the per-item table (event name, trigger, actor, surface); a pointer to source code does not satisfy a documentation deliverable.
- 🟡 **Warning:** The list of components that deliberately emit nothing — `e2e`, `approvals`, `terms`, `egress-rules` — exists only in the PR description, not in the published document — `docs/architecture/usage-tracking.md:107`. Issue #3259 requires every such component to be "listed with a reason why not".
  **Fix:** Put every deliberate coverage exclusion and its reason in the published document, because a PR description is not a deliverable a later reader can find.
- 🟡 **Warning:** The `egress-rules` module has four user-facing mutations and emits no event — `packages/api-server/src/modules/egress-rules/services/egress-rules-service.ts:66`, `:136`, `:190`, `:210`. Issue #3259's stopping rule requires every user-facing component to emit at least basic usage events, and the PR body itself calls this "an omission rather than a decision".
  **Fix:** Close every gap the PR itself identifies as undecided before merge, or record the exclusion as a decision with its reason; an admitted omission against a mechanical completeness criterion is unfinished scope.

The four findings above under `### Findings` (relay wake window, emit-before-success, `ArtifactShared`, `ArtifactPublished`, OAuth pairing, `surface` typing) are this skill's, merged there. One further finding, not mirrored above:

- 🟢 **Suggestion:** `SessionTurnRelayed` and `AgentRelayAttached` are emitted from the transport and admission layer (`packages/api-server/src/apps/api-server/agent-proxies/acp-relay.ts:172`, `packages/api-server/src/apps/api-server/agent-proxies/upgrade.ts:94`), so unlike every other registry event they have no owning module service holding their invariants. A module port injected into the adapter would give the relay events the same home the rest have.

### Verdict
COMMENT — the corrections and the new gate are sound, but seven event-timing and labelling issues make specific counts wrong, and three of #3259's Definition-of-Done bullets are not yet met.

---

packages/api-server/src/apps/api-server/admission/auth-middleware.ts
```diff
@@ -30,6 +31,7 @@ export function createAuthMiddleware(
     const { user, roles } = admitted.principal;
     c.set("user", user);
     c.set("roles", roles);
+    c.set("surface", clientSurface(admitted.principal, attribution));
```
dam-code-guardian, line 34: 🟡 **Warning:** This `c.set("surface", …)` is untyped: `surface` is absent from the app's own `Variables` (`app.ts:67`) and from `routes/index.ts:13`, so producer and consumers agree only by convention, each consumer redeclaring its own shape (`library-routes.ts:20`, `import-proxy.ts:41`, `trpc/http.ts:11`). The failure is silent rather than loud — `activity_events.surface` is nullable, so a route that reads `c.get("surface")` without this middleware inserts NULL, and four views in `0024_usage_view_fixes.sql` group by that column.

**Fix:** Declare every context variable once on the app's `Variables` generic and let handlers inherit it, so a rename or a missing setter fails type-checking instead of writing a NULL grouping key.

packages/db/drizzle/0024_usage_view_fixes.sql
jjeliga: Agreed, and it's the same class of error the migration set out to fix — thanks for going a level down.

Confirmed both cases: `usage_auth_users_7d.active_days` groups by `actor_sub` alone, so one person on `ui` and `cli` the same day contributes 2. `usage_multi_surface_users.active_days_total` is worse for exactly the reason you give — its `HAVING COUNT(DISTINCT surface) > 1` guarantees every surviving row is inflated. And `usage_auth_surface_by_user` is correct as you say, since its GROUP BY pins one surface.

Fix: count the grain the column claims — `COUNT(DISTINCT date_trunc('day', occurred_at AT TIME ZONE 'UTC'))` — in the two views whose GROUP BY doesn't pin a surface.

packages/api-server/src/apps/api-server/agent-proxies/acp-relay.ts
jjeliga: Half superseded, half still right.

The drain path you're citing no longer exists. Live testing against a deployed install found the ACP relay reconnecting on a fixed ~5s interval, which meant any turn outliving that window was booked `failure/client-gone` while its reply arrived on a socket that never saw the prompt — biasing precisely against the long turns the doc line promised to protect. Session turns are now counted **on send**, so there is no `inFlight` set, no drain, and no `client-gone` row; the sentence you quote from `usage-tracking.md` has been rewritten.

The underlying point survives the change, though: `trackIfPrompt` still runs at flush time, not at accept, so a prompt buffered during a cold start and never forwarded records nothing at all — and the wake window is still where the user waits longest.

Fix: move the emit to where the client message is accepted, which is what the new count-on-send semantics should have implied in the first place.

packages/api-server/src/modules/invocations/services/invocations-service.ts
jjeliga: Agreed — verified still live. The emit sits at `:165` and `agents.create` at `:173`, and the `catch` deletes the invocation row while the `activity_events` row it already wrote stays behind. A failed spawn is counted as a spawn, and because the invocation row is gone the event is the *only* remaining trace, so it can't even be reconciled later.

Fix: move the emit past the last failable step.

packages/api-server/src/modules/connections/services/connections-service.ts
jjeliga: Agreed, and the reasoning holds. The add is skipped for `built.auth.kind === "oauth"` and deferred to the callback, but the row already exists in `pending` before that, and `deleteConnection` has no status gate — so abandoning an authorization deletes a row that was never recorded as added, and the removal always emits. `removes > adds` is still reachable, which is exactly what `0024_usage_view_fixes.sql:99` claims to have closed.

One note: after rebasing onto current `main` this is back to the `ConnectionCreated`/`ConnectionRemoved` pair rather than the merged event you reviewed, because `main` grew its own consumers for both. The asymmetry is unchanged by that.

Fix: suppress the removal event for a connection that never reached connected, so every row that can be removed was recorded as added.

packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts
jjeliga: Agreed. `visibility` and `expiresInHours` are independently optional and the emit is unconditional, so extending an expiry on a private artifact counts as a share — and records `visibility: "private"` while doing it, which makes the row self-contradicting.

Fix: compare the prior row's visibility against `updated.visibility` and emit only on an actual transition.

packages/api-server/src/modules/experiments/services/experiments-service.ts
jjeliga: Agreed — still live, the emit precedes `launchRun` on the very next line.

Worth flagging that the shape changed in the rebase: `main` added its own `ExperimentChanged` for live UI updates, so the usage fields are now optional additions on that type and the lifecycle action rides the existing `emitChanged` helper. The ordering is untouched by that, so the finding stands as written.

Fix: same as the `InvocationSpawned` case — emit after the last failable step.

packages/api-server/src/apps/api-server/admission/auth-middleware.ts
jjeliga: Agreed on the substance. One mitigation worth noting: `surface` is now derived in a single place (`clientSurface`) rather than recomputed per call site, so the *value* is single-sourced even though the context key isn't. That removes the drift-between-derivations risk but not the one you're describing.

You're right that the remaining failure is silent in the worst way — `activity_events.surface` is nullable, so a route that misses the middleware writes a NULL grouping key into four views rather than failing anywhere visible.

Fix: declare `surface` on the app's `Variables` generic so producer and consumers are checked rather than merely agreeing by convention.

dam-code-guardian (commented):
### Changes since last review

✅ **Fixed** (`a303e991` → `3f16f2ce`):
- `active_days` column rename — `COUNT(*)` now correctly equals active days (partial unique index design confirmed in migration comment)
- Wake-buffered turn tracking — `trackIfPrompt()` fires before `upstream.send()` in the pending-flush loop
- Non-OAuth `ConnectionCreated` — `connections-service.ts` now emits for `header`/`client-credentials`/`github-app`/`none` auth kinds
- Per-event documentation — `docs/activity-events.md` generated by `event-coverage.mjs`, drift-gated via `mise run check`
- Deliberate-exclusions list with stated reasons for each non-tracked surface
- `docs/architecture/usage-tracking.md` substantially updated; deferred-to-code finding resolved
- `AgentSkillChanged` now emitted for `createLocal` and `deleteLocal` in `skills-service.ts` (lines 961, 997)

🔁 **Still present** (4):

- **`InvocationSpawned` fires before agent exists** — `invocations-service.ts:164` emits before `agents.create()` at line 173. Fix: move emit after the `agents.create()` success path.
- **`ArtifactShared` fires unconditionally** — `artifact-library-service.ts` `setSharing()` emits regardless of whether visibility changed. `persist-activity` records every call as a sharing action even for no-op visibility updates. Fix: compare `input.visibility` against `updated.visibility` before the emit, or guard in the saga.
- **`ArtifactPublished` includes machine-written artifacts** — `create()` emits `ArtifactPublished` for every caller including experiment-driven artifact writes; `agentId` present with `null` signal in payload but no visibility gate. Fix: gate the `ArtifactPublished` emit on `visibility === "public"` or on the absence of an attributed `agentId`.
- **`surface` not in Hono `Variables` type** — `auth-middleware.ts:34` calls `c.set("surface", ...)` but `app.ts:67` declares `Variables: { user: UserIdentity; roles: string[] }` without `surface`, leaving `c.get("surface")` returning `unknown`. The tRPC HTTP adapter at `trpc/http.ts:11` correctly declares `surface: string` in its own subapp `Variables`, but the parent app's type is incomplete. Fix: add `surface: string` to the parent app's `Variables` generic.

---

### Findings

## TypeScript Engineering Review

**🔴 Module boundary bypass: `live-events` imports `agents` internals directly**

Rule: cross-module imports are only allowed through a module's `index.ts`. (`module-boundaries.md` — *What Is Not Allowed*)

`packages/api-server/src/modules/live-events/compose.ts` imports directly from `../agents/infrastructure/labels.js` (three constants) and `../agents/infrastructure/k8s.js` (`K8sClient` type). None are exported from `packages/api-server/src/modules/agents/index.ts`.

Fix: re-export the three label constants through `agents/index.ts` and import them there. For `K8sClient`, define a structural `AgentWatcher` port interface locally in `live-events/infrastructure/k8s-agent-watch.ts` — the `compose.ts` already uses `Pick<K8sClient, "watchCustomObjects">`, so the type can be inlined and the cross-module import dropped entirely.

**🟡 Business logic in the contract (validation) package**

Rule: `packages/api-server-api/src/trpc.ts` is the tRPC initialisation file — no business logic. (`slice-composition.md` — *Validation Layer*)

`packages/api-server-api/src/trpc.ts:40–65` (new in this diff): `requireTermsAccepted` middleware calls `ctx.terms.isAccepted()` and owns a per-process memoisation cache (`termsProven WeakSet<ApiContext>`). The exported `markTermsProven` function lets the server package reach back and mutate this cache — two-way coupling between the contract package and the server's infrastructure.

Fix: move `requireTermsAccepted`, `termsProven`, and `markTermsProven` to the server package — either into the tRPC context factory (`packages/api-server/src/apps/api-server/trpc/context.ts`) or a dedicated middleware factory there. `PRE_TERMS_PROCEDURES` (a pure constant) can stay in the contract package.

## React UI Engineering Review

**🟡 `h-[52vh]` fixed viewport-fraction height in `SkillMarkdownModal`**

`packages/ui/src/modules/sandboxes/components/skills/skill-markdown-modal.tsx` — content pane uses `h-[52vh]` (new in this diff). Existing modals use `max-h-[40vh]` / `max-h-[44vh]` with `overflow-y-auto`. `h-[52vh]` forces the pane to exactly that height regardless of content; confirm this is intentional, particularly on short viewports.

**🟢 `Switch` accessibility in `FeatureRowCard`**

`packages/ui/src/modules/features/components/features-tab.tsx` — `<label>` implicit labelling is unreliable for `<button role="switch">` children across browsers and AT. Consider passing `label={row.label}` to `<Switch>` for a reliable `aria-label`. (Introduced by the `FeatureRowCard` refactor — moment to fix it.)

## Documentation Check (doc-drift)

**Verdict:** minor drift

**🔴 `docs/architecture/artifact-library.md` not updated** (Check 1 — architecture-page drift)

`Last verified: 2026-07-30`, unchanged in this PR. Every mutation in `artifact-library-service.ts` now emits a domain event (`ArtifactCreated`, `ArtifactUpdated`, `ArtifactDeleted`, `ArtifactFolderChanged`, `ArtifactPublished`, `ArtifactShared`). The page makes no mention of domain events for this subsystem.

Proposed addition (new "Domain events" section or after "Lifecycle and cleanup"):

> Every artifact mutation emits a non-durable domain event on the in-process bus: `ArtifactCreated`, `ArtifactUpdated`, `ArtifactDeleted`, `ArtifactFolderChanged`, `ArtifactPublished` (on create; includes kind, visibility, attributed agent), and `ArtifactShared` (on visibility change). These events are advisory — see [platform-topology](platform-topology.md) §Domain events and live updates.

Also bump `Last verified: 2026-08-14`.

**⚠️ `docs/architecture/connections.md`** (Check 1 — possible drift)

Page was updated (`Last verified: 2026-08-14`). `connections-service.ts` now emits `ConnectionCreated` for all non-OAuth auth kinds (connections that complete synchronously). The page does not describe this domain event behavior. Lower priority — the page doesn't claim events are OAuth-only — but confirm whether the event emission scope belongs in the connections architecture page for consistency with platform-topology's event bus description.

## Issue Fit

Linked issue: #3259 "Update and document raw interaction data collection" (inferred — `refs #3259` in PR body, confirmed by branch `chore/3259-interaction-data-collection`; **not a closing keyword** — the issue will not auto-close on merge)

The issue's four deliverables (inventory, fixes, coverage extension, documentation) are all delivered. The `api-server:check:events` registry gate is declared scope extension, not undeclared creep.

**🟡 `egress-rules` rationale inconsistency**

The `event-coverage.mjs` `NOT_TRACKED` array documents `egress-rules` as "Rule edits are an administrative action, not product usage." The PR body says it "never came up during scoping, so it is an omission rather than a decision." One of these should be updated before merge so the definitive position is clear — either commit to the documented exclusion or add the events and remove it from `NOT_TRACKED`.

---

### Approval bar

🔴 **Module boundary bypass** must be resolved before merging: re-export through `agents/index.ts` or define a local port interface in `live-events`.

🟡 **Remaining carried warnings** (`InvocationSpawned` timing, `ArtifactShared` unconditional, `ArtifactPublished` machine artifacts, `surface` untyped) and the new business-logic-in-contract finding do not block merge but should have a plan.

🟡 **`artifact-library.md` doc drift** should be addressed in this PR since all six event emitters were added here.

packages/api-server/src/apps/api-server/agent-proxies/acp-relay.ts
```diff
@@ -159,6 +166,16 @@ export function createAcpRelay(
           .catch(() => {});
       }
 
+      function trackIfPrompt(parsed: unknown): void {
```
PetrBulanek, line 170: This counts every `session/prompt` frame as a user turn. But the UI sends some
by itself. `use-agent-greeting.ts:46` sends a hidden prompt the first time
someone opens a Knowledge Base or Experiment sandbox that has no sessions yet.
The user typed nothing, and a `session_turn` row is still written.
The effect is that a sandbox someone opened once and left reads as having held
a conversation. That is the case a usage report most needs to get right. Channel
turns have no automatic send, so the two turn counts no longer compare — which
is the whole reason session turns were added.
The relay needs some way to tell a machine-sent prompt from a typed one. The
prompt already carries a `_meta.platform` block with `promptId` and `surface`,
so there is somewhere for that signal to live.

packages/api-server/src/modules/invocations/services/invocations-service.ts
```diff
@@ -160,6 +161,12 @@ export function createInvocationsService(deps: {
         expiresAt,
         experimentSpanId: input.experimentSpanId ?? null,
       });
+      emit({
+        type: EventType.InvocationSpawned,
```
PetrBulanek: This fires before `agents.create()` runs. If that call throws, the `catch` below
deletes the invocation row that was just written — but the event has already
left. `persist-activity` stores it with a hardcoded `outcome: "success"`.
So a spawn that failed and was rolled back leaves a permanent
`invocation_spawned` row behind. Agent creation does fail in real use — a
rejected admission webhook, a quota, a failed image pull. The event should only
describe a spawn that actually happened.

packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts
```diff
@@ -412,6 +422,13 @@ export function createArtifactLibraryService(
         ownerSub: owner,
         ...(updated.agentId ? { agentId: updated.agentId } : {}),
       });
+      emit({
+        type: EventType.ArtifactShared,
```
PetrBulanek: `setSharing` emits this on every call, including calls that change nothing about
sharing. The share dialog always sends `visibility`, and its Save button has no
guard (`share-dialog.tsx:165`). So someone who opens the dialog and presses Save
without touching the toggle writes an `artifact_shared` row. On a private
artifact that row records `visibility: "private"` — a share event for something
that is not shared. Changing only the expiry does the same.
The number this produces is "times Save was pressed", not "times an artifact was
shared". `requireArtifact(id)` on line 412 already reads the row before the
update, so the previous value is available to compare against.

packages/api-server/src/modules/skills/services/skills-service.ts
```diff
@@ -904,6 +957,17 @@ export function createSkillsService(deps: SkillsServiceDeps): SkillsService {
...
         detail: { names: input.skills.map((s) => s.name) },
       });
+      for (const skill of input.skills) {
+        emit({
+          type: EventType.AgentSkillChanged,
+          action: "installed",
+          agentId: input.agentId,
+          actorSub: deps.owner,
+          surface: deps.surface,
+          origin: "local",
```
PetrBulanek, line 967: This emit, and the matching one in `deleteLocal`, mean creating and deleting a
Local Skill now writes a row to `activity_events`.
`docs/architecture/skills.md` still says both operations record **nothing** in
Postgres — line 145 for Create Local, line 147 for Delete Local.
Worth fixing in this PR rather than later, because `usage-tracking.md` (updated
here) argues the opposite: that a Local Skill leaves no state, so the event is
the only trace it ever existed. The two architecture pages now disagree about
whether a Local Skill writes anything to Postgres.

PetrBulanek (changes requested):

packages/api-server/src/apps/api-server/agent-proxies/acp-relay.ts
jjeliga: I would argue that saying "is the case a usage report most needs to get right." is n overstatement. These counts are just indicators of how much people actually use dam, how many turns with an agent do they have approximately. Being off by a couple messages in the scale of hundreds does not really change anything here. That is just for the bigger picture. One case where this would bite is when someone creates a KB, but never uses it - we might see 1 message despite that.
However, it is fixed now through passing jsonrpc meta info on the message initiator

packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts
jjeliga: `ArtifactShared` now fires only on an actual visibility transition (it was recording `visibility: "private"` while claiming a share

packages/api-server/src/modules/skills/services/skills-service.ts
jjeliga: Event is emmited, docs are updated to match this

packages/api-server/src/modules/invocations/services/invocations-service.ts
jjeliga: Fixed.
`InvocationSpawned` moved past `agents.create`; its `catch` deletes the invocation row, so a failed spawn left the activity row as the *only* trace of something that never happened

packages/api-server/src/modules/usage/routes.ts
```diff
@@ -1,7 +1,9 @@
 import { Hono, type Context, type Next } from "hono";
 import type { UserIdentity } from "api-server-api";
```
dam-code-guardian: 🟡 **Warning:** `UserIdentity` is now unused — `AppEnv` was rewritten to reference `ApiVariables` but this import was not removed.
**Fix:** Remove `import type { UserIdentity } from "api-server-api"`.

packages/api-server/src/modules/usage/routes.ts
```diff
@@ -1,7 +1,9 @@
...
 
-type AppEnv = { Variables: { user: UserIdentity; roles: string[] } };
+type AppEnv = {
```
dam-code-guardian: 🟡 **Warning:** `type AppEnv` is declared here before `ApiVariables` is imported (line 16). TypeScript hoists type aliases but the declaration-before-import ordering breaks the module's own import discipline.
**Fix:** Move this `type AppEnv` block after all import statements.

dam-code-guardian (commented):
⚠️ **Code Guardian 🛡️** — ⚠️ Code Review @ `e5b438d4`

## PR #3314: chore(usage): extend and correct raw interaction collection
**Author:** jjeliga | **Branch:** chore/3259-interaction-data-collection → main | **Changes:** +10501 −419 (82 files)

### Summary
6 of 9 prior findings fixed. The `surface` typing, event guards, and `InvocationSpawned` ordering are all resolved. The module boundary bypass in `live-events/compose.ts` is still present and still blocks merge.

### Changes since last review
Previous HEAD: `3f16f2ce` (2026-08-14) — verdict COMMENT

- ✅ **Fixed:** `surface` not in Hono `Variables` — `ApiVariables` type created in `http-context.ts`, used in `app.ts` and now `routes.ts` (`packages/api-server/src/core/http-context.ts`)
- ✅ **Fixed:** `ArtifactShared` fires unconditionally — now guarded by visibility change (`packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts:401`)
- ✅ **Fixed:** `ArtifactPublished` counts machine artifacts — gated by `!attribution?.internal` (`packages/api-server/src/modules/experiments/services/experiments-service.ts:439`)
- ✅ **Fixed:** `artifact-library.md` doc drift — Domain events section added, Last verified bumped
- ✅ **Fixed:** `egress-rules` rationale inconsistency — exclusion present in both `event-coverage.mjs` and `docs/activity-events.md` "Deliberately not recorded" table
- ✅ **Fixed:** `InvocationSpawned` fires before agent exists — emit moved after `agents.create()` success path
- 🔁 **Still present:** module boundary bypass — `live-events/compose.ts` imports `agents/infrastructure` directly (`packages/api-server/src/modules/live-events/compose.ts:12`)
- 🔁 **Still present:** business logic in contract package — `requireTermsAccepted` + `termsProven` cache in `api-server-api/src/trpc.ts` (`packages/api-server-api/src/trpc.ts:40`)
- 🔁 **Still present:** `h-[52vh]` fixed height in SkillMarkdownModal (`packages/ui/src/modules/sandboxes/components/skills/skill-markdown-modal.tsx:141`)
- 🆕 **New:** dead `UserIdentity` import left after `AppEnv` refactor (`packages/api-server/src/modules/usage/routes.ts:2`)
- 🆕 **New:** `AppEnv` type alias declared before the import it references (`packages/api-server/src/modules/usage/routes.ts:4`)
- 🆕 **New:** `docs/architecture/skills.md` — no Domain events section despite skills now emitting usage events
- 🆕 **New:** `docs/architecture/experiments.md` — not updated; stale Last verified: 2026-08-13

### Findings
- 🟡 **Warning:** dead `UserIdentity` import left after `AppEnv` refactor (`packages/api-server/src/modules/usage/routes.ts:2`)
  **Fix:** Remove `import type { UserIdentity } from "api-server-api"` — `UserIdentity` is no longer referenced after `AppEnv` switched to `ApiVariables`.
- 🟡 **Warning:** `AppEnv` type alias declared before the import it references (`packages/api-server/src/modules/usage/routes.ts:4`)
  **Fix:** Move the `type AppEnv` block after all import statements so it follows the import of `ApiVariables` at line 16.

- 🟡 **Warning:** `docs/architecture/skills.md:133` — The skills service now emits domain events for install, uninstall, publish, skill-set save, and skill-set delete operations (`AgentSkillChanged`, `SkillPublished`, `SkillSetSaved`, `SkillSetDeleted`), but the skills architecture page has no domain-events section. The `artifact-library.md` update in this PR is the pattern to follow. **Fix:** Add a "Domain events" section to `docs/architecture/skills.md` documenting which operations raise events, which reach the activity log, and the actor-gated rule that filters machine-generated events.
- 🟡 **Warning:** `docs/architecture/experiments.md` — `ExperimentChanged` was extended with `action`, `actorSub`, and `surface` fields so experiment lifecycle transitions are now activity-logged, but `experiments.md` (Last verified: 2026-08-13) was not updated. **Fix:** Bump `Last verified` to 2026-08-17 and add a short Domain events note for experiment lifecycle transitions.

### Issue Fit
Linked issue: #3259 (plain `refs #3259`, no closing keyword)

✅ `egress-rules` exclusion verified present in `docs/activity-events.md` "Deliberately not recorded" table with rationale.

🟢 **Suggestion:** PR uses `refs #3259` rather than a closing keyword — the issue has a still-open scope item; if this PR delivers it fully, use `closes #3259` so the issue closes on merge.

- 🟡 2 new findings merged into `### Findings` above (dead import, declaration order).
- 🟢 **Suggestion:** `packages/api-server/src/modules/usage/domain/types.ts:6` — `externalActorId` and `ownerSub` are new optional fields on `ActivityEventRow` that carry persistence-layer pseudonymization concerns. A stricter domain model uses a discriminated union per event type.

### Verdict
COMMENT — 🔴 module boundary bypass in `live-events/compose.ts` still blocks merge; 4 new doc/code 🟡 warnings added this round.

jjeliga:
**Review-bot findings, round 2** — three of four fixed in 60e255fe / 55f12d62: the dead `UserIdentity` import, the `AppEnv` alias declared above its import, and the `experiments.md` domain-events drift (section added, `Last verified` bumped).

**The `skills.md` finding is accepted as a known gap, not fixed.** The page sits 38 characters under the 40,000-char page cap, so a Domain events section cannot fit without cutting content — and the page has no natural seam to split along (assessed earlier in this PR; the subsystem's concepts are too interwoven to cut cleanly). What keeps the gap from rotting:

- the generated `docs/activity-events.md` enumerates every skills emit site — which event, which activity row type, where it fires — and `docs:check:activity-events` fails on any drift from the source;
- `usage-tracking.md`'s persist-activity section names skills explicitly and points at that enumeration.

So the facts are recorded and drift-gated; what is missing is only the pointer *from* the skills page. If the page is ever restructured, a Domain events section should ride along.

PetrBulanek (approved):

Issue #3213, opened 2026-08-06
UI - Experiment sandbox's first message should say what to do, not explain how experiments work


jenna-winkler:
## Context

When a user creates an experiment sandbox, the agent greets them once, unprompted. Today that greeting is two dense paragraphs. It explains the design→build→test→learn loop, that the loop is Python the user writes, that the platform watches stages and charts a score, and that the platform never runs the loop or interprets the score. It then asks what the user wants to optimize.

This is the user's first second in the feature. They have not asked how anything works.

## Problem

The greeting is overwhelming. It leads with system mechanics, so the one thing the user must do — say what they want to make better — arrives last and is easy to miss. A new user cannot tell whether they are expected to write code, upload something, or answer a question.

It also front-loads caveats about what the platform does *not* do. That reads as a warning before the user has any reason to care.

## Goal

A new user reads the first message and knows immediately what the agent can do for them, and what to type next. The mechanics come later, when they matter.

## User Stories

- As a new experiment user, I want the first message to be short so that I can read all of it.
- As a new experiment user, I want one obvious next step so that I do not have to guess what is expected of me.
- As a new experiment user, I want to hear what the agent can build for me so that I can judge whether experiments fit my problem.

## Proposed solution

Rewrite the greeting to lead with capability and end with one question. Roughly: one line on what the agent will do for the user, two or three concrete examples of goals it can chase, then "what do you want to make better?"

Cut from the first message: how the loop is structured, that it is Python, what the platform does and does not interpret, and the graph/chart/progress tour. The agent can explain any of it on demand, or when it becomes relevant while agreeing the loop's stages.

## Open Questions

- Should the examples be clickable starters rather than prose?

jenna-winkler:
Message draft here: https://ibm-research.slack.com/archives/C0B09BF65CP/p1787166631996329

jenna-winkler:
@Tomas2D status?

Tomas2D:
Ready to be merged.

https://github.com/dam-agents/dam/pull/3318

Tomas2D:
Merged

PR #3318, opened 2026-08-13
Harden the Nous experiment path: crash fast-fail via Agent status, image fail-fast, authoring kit

Tomas2D:
## Why

A Nous experiment run wedged for 4+ hours: its spawned target sat parked OverBudget while the driver polled an invocation that could never report, and the liveness sweep's crash fast-fail turned out to be dead cluster-wide — it read pods, which the api-server holds no permission for (deliberately: the chart withholds it, and platform-topology pins "the api-server never inspects pods").

Fixing that opened the rest of the branch. Every subsequent commit came out of driving real runs on the local cluster, and each one fixed a way a run could burn hours and hand back nothing: a target left running after the script died, a single mesh 503 killing every remaining round, an OOM-killed worker, a worker sized past a ceiling it could never fit, and — once runs finally completed — a "technically passing" result whose score chart was a single dot pinned to the timer's resolution. So the PR is now three things: **the platform fails fast or degrades instead of wedging**, **a run is legible while it runs**, and **the authoring kit teaches the science that makes the numbers mean something**.

### Crash fast-fail without pod reads

- `AgentStatus` gains `agentPodRestarts` + `agentPodRestartReason` (CRD schema gen 8) — the one crash signal that outlives a pod recovering (conditions return to `AgentPodReady=True` and drop the termination cause). Cleared on hibernation so a scaled-down agent never reads as crashed.
- The invocation liveness sweep swaps its pod read for a `readTargetRestart` port over the Agent CR; `readAgentPodRestart` and its 403-spamming `listNamespacedPod` are deleted. No RBAC change — the invariant now holds in both directions.

### Fail fast at spawn instead of parking or guessing

- `experiment_sdk.require_image()` resolves a template id against the catalogue at declaration time, so `--plan` fails during design review instead of hours into a run; the spawn route rejects an unknown `templateId` with a 400 naming the ids that exist (was a bare NOT_FOUND surfacing as a 500).
- **Target admission**: a spawn whose effective Size (explicit `size` → template limits → chart default) exceeds the owner's budget Ceiling is rejected at the route with both figures. The controller's 0→1 gate could never admit it, so it would have parked until the invocation deadline reaped it hours later. A spawn that fits the Ceiling but not the *currently free* room is deliberately **not** rejected — that one queues and auto-starts, the existing queue-not-fail path.
- **Budget is readable at design time**: `GET /api/agents/:id/budget` over the harness surface plus `x.budget()` returns reservation, ceiling and the default worker size, and `list_images()` now carries each worker's Size — so a script can state how many workers run at once before the human approves, rather than discovering it as a queue that eats each TTL.

### Terminal reaping and poll resilience

- The harness REST surface composed the experiments service **without** `agents`, so `deps.agents?.delete` was a silent no-op on every script-initiated finish — the terminal reap only ever worked from the UI's Stop button. `agents` is now required on `composeExperimentsForOwner` (a call site cannot drop it again without failing tsc), and per-target delete failures are logged instead of swallowed.
- The SDK's `_request` raised on the first transient failure; a long run polls thousands of times, so one mesh 503 (`upstream connect error ... connection termination`) killed every remaining round. GETs now retry with bounded exponential backoff on 502/503/504 and connection errors; non-GETs stay single-shot, because a repeated finish/report is a conflict, not a dup.
- A failed spawn now says **why**: `errorReason` rides the invocation view (optional, so older servers stay valid) and the SDK surfaces it in `InvocationFailed`'s message and as `.reason` — the only diagnosis that survives the target being reaped.
- `ttl_ms` is documented as a kill deadline, not pacing: it is sized from the worst plausible round plus slack, and the clock starts at spawn (a queued spawn spends its budget waiting).

### A run is legible while it runs

- **Stage and loop descriptions**: skeleton stages/loops accept an optional one-sentence description (schema max 300 chars; old SDKs and stored drafts parse unchanged), `loop(id, description=...)` / `stage(id, after=..., description=...)` carry it into the plan, and the stock dashboard renders it under the node name. Bare ids taught a reviewer nothing.
- **Durations everywhere**: the feed projection computes per-stage total/avg/last span duration; the stock dashboard shows the run's elapsed/final duration and per-stage avg + total; the lineage run rows show each run's duration. `formatDurationMs` moved to a shared helper.
- **Chat surfaces are copyable and readable**: copy actions on thinking blocks, tool steps and their output, and permission verdicts; long tool titles and verdict subjects expand instead of truncating forever; tool chips are always expandable. `ActivityBlock` grew an `actions` slot to carry this.
- **Prompt chips on a fresh experiment sandbox**: five one-click questions above the composer (an example to optimize, how experiments work, how a run is scored, what can be run, which agents run the rounds) — experiment sandboxes only, idle chat only, until the first message.

### The authoring kit (`dam-experiment` skill)

- **Interview before authoring a Nous experiment** — target repo, primary metric + direction + pass condition, campaign iterations, seeds, experiment rounds — and compute the TTL from the answers. New `references/nous-evaluator.md` documents Nous's on-worker output layout (stable verdict contract vs campaign-specific raw files), framed for the driver: the worker is reaped right after reporting, so the prompt must name what to report and publish. `/experiment-onboard` recognizes a Nous-shaped goal and runs the interview before proposing stages.
- **The image catalogue is read, not guessed** — `references/images.md` defines every catalogued image and marks the five validated for unattended loops (claude-code, codex, pi-agent, bob, nous), with `claude-code` the stated default because it needs no extra credential; onboarding calls `list_images()` before greeting and checks `list_connections()` before offering a provider-bound worker. Approval covers the whole envelope: image, connections, iteration counts, expected duration in human terms, and the deadline derived from them.
- **A real starter ships with the skill** — `examples/tiny-cache` (naive cache, 11-test behavioral suite, deterministic per-seed bench) lives inside the skill directory, so a pod roll delivers the current copy to existing sandboxes and one local `cp` seeds a workspace. Never cloned, never downloaded: no external-code trust question, and the pristine skill copy makes the anti-cheat check a `git diff --no-index` of `test/` and `bench/`. "Show me" is answered from the reference with nothing copied and nothing run; "set it up" announces itself before executing anything.
- **New section: designing a score that means something** — score per measurement (per seed / sweep point / arm), not per round, so the chart is a curve rather than one dot; more rounds of a deterministic change measure the harness, not the effect; check the fast side isn't landing on the instrument's floor (a p50 ratio swung 424× → 651× on identical code — batch `ops_per_sec` instead); sweep the parameter the effect depends on (1.7× at n=1, 525× at n=20 000); put ablations in the data layer; keep absolute numbers next to every ratio; set a pass bar that can actually fail (a 10× bar cleared at 500× pre-registers nothing); and **the driver owns the headline number**, computed from the structured result — a worker's own `report.md` has shipped a wrong baseline and a false claim that its per-seed files didn't exist. The Nous example is reworked to score one span per seed with the round median in `post_data`.
- **Tone and shape** — a "Talking to the user" section (lists and tables over prose, guarantees over mechanisms, one decision per question, plain words), a required stage table when registering a draft, and a rule that a missing common tool is the agent's to solve silently instead of narrating tool archaeology.

### Cluster, chart and image sizing

Each of these was a way a local run died, so the values carry their forensics in comments:

- lima VM to **10 CPU / 20 GiB** — at 8 CPU an in-budget fan-out drove load average past 20, kubelet probes timed out, the node flapped NotReady and Keycloak crash-looped until the UI lost auth.
- Budget ceiling moved to **`controller.userBudgets`** (a top-level `budgets:` block was silently ignored by the chart) and set to 6 CPU / 14Gi — room for a driver plus a heavy worker, while leaving the platform stack ~4 CPUs.
- `nous` enabled locally and sized 2 CPU / 8Gi (a pm2 campaign held three ~1 GiB daemons and OOM-killed itself at 4Gi); `gepa` sized to match. CPU is deliberately half the memory-driven sizing — the worker's boot burst starved ztunnel and synthesized the 503 that killed a run.
- The api-server gets its own requests/limits and explicit probe timeouts: Kubernetes defaults `timeoutSeconds` to 1, which a loaded event loop misses even on a static route — a missed readiness check pulls the pod from the harness Service and answers polling drivers with a 503, a missed liveness check restarts the container. Slow must degrade, not restart.
- `mise run cluster:prune` reclaims both disks (buildkit cache capped via `--keep`, dangling docker + k3s images, leftover import tars), and cluster-ops gains a "Disk space (two independent VMs)" section — the ENOSPC is the docker VM's disk, which no cluster-side prune touches.
- `platform-base` gains the utilities agents kept working around: diffutils, file, gawk, hostname, openssl, patch, procps-ng, unzip, zip.

### Docs

`docs/architecture/budgets.md` (the one admission the api-server refuses up front), `experiments.md` (concurrency as part of the approved envelope, spawn fail-fast, failed spawns carrying the platform's reason), `platform-topology.md` (pod facts are the controller's to publish; the vm-backend limit of the published restart count). The branch predates main's stripped-comment policy, so its prose comments were stripped and the load-bearing rationale moved into these pages; the new suites keep their explanation as registered `TEST_OVERVIEW` / `TEST_SCENARIO` comments.

## Deploy note

The CRD must apply before the new controller image rolls, or its status writes are pruned — `cluster:install` handles the ordering. The lima changes need an existing VM resized in place (`limactl edit platform-k3s`, per the comment in `deploy/lima-k3s.yaml`). Verified live on the local cluster: CRD at gen 8, no more `invocation-liveness` 403s, the controller's budget env picked up the override once it moved under `controller.userBudgets` (a top-level block left it on the stock values), skill kit confirmed inside the rebuilt image.

## Tests

New suites, all in the unit tiers:

- **controller** — `pod_restarts_test.go`: 6 tests (restart counting, cause attribution, publish on a recovered pod, hibernation clear)
- **api-server** — 5 liveness crash-fastfail tests (fail+reap, unclassified restart, healthy, absent Agent, throwing read), 7 size-admission tests, 2 template-validation tests, 4 terminal-reap tests, plus trace-graph duration cases
- **experiment-sdk** — 6 request-retry tests (backoff on 502/503/504 and connection errors, non-GET single-shot) and 6 more SDK tests (`require_image`, descriptions, `budget()`)
- **ui** — 5 tool-chip tests
- **e2e** — `playwright/verify-experiment.mjs`, a scripted end-to-end verification of an experiment run and its budget surface

Verified on this branch:

- `mise run test` green — api-server 1113 (118 files), ui 307, agent-runtime 198, cli 154, experiment-sdk 24, driver-sdk 11, `go test ./...` ok across every controller package
- `mise run check` green — comment-types, CRD schema-generation and backwards-compat gates, helm lint/render, ADR immutability included

Ref: #3336 , #3213


packages/api-server-api/src/modules/experiments/schemas.ts
```diff
@@ -37,6 +38,7 @@ export const skeletonSchema = z.object({
       z.object({
         id: stageIdSchema,
         stages: z.array(stageIdSchema).min(1).max(50),
+        description: z.string().trim().min(1).max(300).optional(),
```
dam-code-guardian, line 41: 🟡 **Warning:** `loops[].description` validates and is stored, and the SDK carries it into the plan, but nothing reads it. The stock dashboard indexes `stages[].description` only (`stock-dashboard.ts:235`), and a tree-wide grep finds no consumer of `skeleton.loops` in the api-server or the UI. The PR body promises the dashboard renders it.
**Fix:** Render the loop description on the loop's own surface, or drop the field from the schema and the SDK until a reader exists — a schema field with no consumer accepts input it silently discards.

packages/ui/src/modules/sessions/components/tool-chip.tsx
```diff
@@ -1,42 +1,114 @@
...
+export function stripFences(text: string): string {
   return text.replace(/^```\w*\n?/, "").replace(/\n?```\s*$/, "");
 }
 
+export function ToolContentBlock({ text }: { text: string }) {
+  const { copy, copied, state } = useCopy();
+  const stripped = stripFences(text);
+
+  return (
+    <div className="group/content relative my-1 rounded bg-muted/40 border border-border/40 p-2">
```
dam-code-guardian, line 22: 🟡 **Warning:** This names the group `group/content`, but `HOVER_ACTION` carries the **unnamed** `group-hover:opacity-100`, so the variant resolves against the nearest bare-`group` ancestor — the `ActivityBlock` wrapper (`activity-block.tsx:24`). Hovering anywhere on the block reveals every nested output copy button at once.
**Fix:** Pair a named group with its matching named variant (`group-hover/content:`), or use bare `group` here when the reveal class comes from the shared unnamed constant. `markdown-code-block.tsx:23`/`:38` and `sandbox-group-card.tsx:28`/`:56` both pair theirs.

packages/ui/src/modules/sessions/components/permission-prompt.tsx
```diff
@@ -67,17 +69,58 @@ export function PermissionStatusLine() {
...
-      {verdict.allowed ? (
-        <Checkmark size={16} className="text-success shrink-0" />
-      ) : (
-        <Close size={16} className="text-destructive shrink-0" />
-      )}
-      <span className="text-muted-foreground shrink-0">{verdict.label} —</span>
-      <span className="font-semibold text-foreground truncate">
-        {verdict.subject}
-      </span>
+    <div
```
dam-code-guardian, line 76: 🟡 **Warning:** The row becomes expandable through `onClick` on a plain `<div>` with no `role`, `tabIndex`, `onKeyDown` or `aria-expanded`, so a keyboard or screen-reader user cannot expand a truncated verdict subject.
**Fix:** Make a disclosure toggle a real `<button>` carrying `aria-expanded`. `ActivityBlock:29` and the second expander in this same file (`:211`) already do exactly that.

docs/architecture/experiments.md
```diff
@@ -152,14 +153,72 @@ missing experiment reads as unknown. Events append only while the experiment
 is `running` — Stop closes the trace, so a stopped loop dies on its next
```
dam-code-guardian, line 156: 🟡 **Warning:** This sentence still attributes trace closure to Stop alone, directly above the new paragraph establishing that all three terminal paths (Stop, `finish` for `completed` and `failed`, and the inactivity sweep) close the ledger and reap targets. The page contradicts itself across two adjacent sections.
**Fix:** When a PR generalizes a behavior from one trigger to several, rewrite every earlier sentence on the page that still names the single trigger.

packages/controller/pkg/reconciler/agent_reconciler.go
```diff
@@ -264,13 +264,20 @@ func (r *AgentReconciler) publishReadiness(ctx context.Context, agent *apiv1.Age
 	gatewayReady := r.podCurrentAndReady(ctx, GatewayName(name))
 	ready := agentReady && gatewayReady
 
+	var agentPod *corev1.Pod
```
dam-code-guardian, line 267: 🟢 **Suggestion:** On the healthy non-VM path the agent pod is now fetched twice per reconcile: `podCurrentAndReady` gets it at `:340`, and this block repeats the same GET. Both go straight to the API server, bypassing the pod informer `main.go:136` already runs. Returning the pod from `podCurrentAndReady` (or reading the lister) would drop the second call.

packages/api-server/src/modules/invocations/services/invocations-service.ts
```diff
@@ -101,6 +102,7 @@ export function createInvocationsService(deps: {
     experimentId: string,
     driverAgentId: string,
   ) => Promise<boolean>;
+  targetAdmission?: TargetAdmission;
```
dam-code-guardian, line 105: 🟢 **Suggestion:** `targetAdmission` is optional and guarded by `if (deps.targetAdmission)` at `:138`, with `compose.ts:39` spreading it conditionally — so a composition root that forgets it skips budget admission silently instead of failing `tsc`. One root wires it today; a required dependency plus an explicit no-op adapter would keep that guaranteed.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `221486b2`

## PR #3318: Harden the Nous experiment path: crash fast-fail via Agent status, image fail-fast, authoring kit
**Author:** Tomas2D | **Branch:** fix/nous-experiment-hardening → main | **Changes:** +3503 −212 (71 files)

### Summary
The PR removes the api-server's pod reads by publishing `agentPodRestarts` on Agent status (CRD gen 8), fails a spawn fast when the template id is unknown or the worker size can never fit the owner's budget ceiling, makes terminal reaping work on every path (not only Stop), retries transient GETs in the driver SDK, and adds run-duration/description surfaces plus the `dam-experiment` authoring kit. Cluster and chart sizing carry their forensics.

### Findings
- 🟡 **Warning:** `mise.lock` loses the pinned `java` platform checksums and URLs (−23 lines, no additions) while `mise.toml:18` still requires `java = "21"`. `docs/security/supply-chain.md:12` names those per-platform SHA-256 pins as the compensating control for mise tools, which nothing else scans. The deletion rode in on the controller commit `2561ffd` and the body declares nothing about it. (`mise.lock:202`; reported by issue-fit)
  **Fix:** Restore the `[tools.java.*]` blocks. Keep a lockfile edit to the tools a PR actually changes — a deletion that drops checksum pinning for a still-declared tool needs its own justification.
- 🟡 **Warning:** Loop descriptions are accepted and stored but reach no reader. `skeletonSchema` validates `loops[].description`, and the SDK carries it into the plan, but the stock dashboard only indexes `stages[].description`; no consumer of `skeleton.loops` exists in the api-server or the UI. The body promises the dashboard renders it. (`packages/api-server-api/src/modules/experiments/schemas.ts:41`; reported by issue-fit)
  **Fix:** Render the loop description on the loop's own surface, or drop the field from the schema and the SDK until a reader exists — a schema field with no consumer accepts input it silently discards.
- 🟡 **Warning:** `ToolContentBlock`'s named group never matches its reveal variant, so every nested output copy button appears at once. (`packages/ui/src/modules/sessions/components/tool-chip.tsx:22`; reported by react-ui-engineering)
  **Fix:** Pair a named group with its named variant, or use bare `group` when the reveal class comes from the shared unnamed `HOVER_ACTION`. `markdown-code-block.tsx:23`/`:38` and `sandbox-group-card.tsx:28`/`:56` both pair theirs.
- 🟡 **Warning:** `PermissionVerdictLine`'s new expander is an `onClick` on a plain `div`, so a keyboard or screen-reader user cannot expand a truncated verdict subject. (`packages/ui/src/modules/sessions/components/permission-prompt.tsx:76`; reported by react-ui-engineering)
  **Fix:** Make a disclosure toggle a real `<button>` with `aria-expanded`, as `ActivityBlock:29` and the second expander in this same file (`:211`) already do.
- 🟢 **Suggestion:** `publishReadiness` now reads the agent pod twice per reconcile on the healthy path — `podCurrentAndReady` fetches it at `agent_reconciler.go:340`, then the new `agentPod` fetch repeats the same GET at `:269`. Both bypass the pod informer `main.go:136` already runs. Passing the pod that `podCurrentAndReady` already has would remove the second call. (`packages/controller/pkg/reconciler/agent_reconciler.go:267`)
- 🟢 **Suggestion:** `targetAdmission` is optional on `createInvocationsService` and spread conditionally in `compose.ts`, so a composition root that forgets it skips admission silently rather than failing `tsc`. The single root wires it today; a required dependency with a no-op adapter would keep it that way. (`packages/api-server/src/modules/invocations/services/invocations-service.ts:105`; reported by typescript-engineering)
- ✅ **Looks good:** `assertCanEverFit` runs before the DB insert and the CR create, so a rejected spawn leaves no orphan row or target; and `finish` reaps only after `repo.transition` succeeds, so a lost race cannot reap another run's targets.

- 🟡 **Warning:** `platform-topology.md` gained the pod-restart paragraph but its header still reads `Last verified: 2026-08-17`, while `budgets.md` and `experiments.md` moved to `2026-08-21` in the same PR. (`docs/architecture/platform-topology.md:3`)
  **Fix:** Bump `Last verified:` on every architecture page a PR edits, in the same commit as the content edit.
- 🟡 **Warning:** `experiments.md` contradicts itself across two adjacent sections: "Stop closes the trace, so a stopped loop dies on its next call" survives directly above the new paragraph establishing that all three terminal paths close the ledger. (`docs/architecture/experiments.md:153`)
  **Fix:** When a PR generalizes a behavior from one trigger to several, rewrite every earlier sentence on the page that still names the single trigger.
- 🟢 **Suggestion:** The new `experiments.md` paragraphs name code-level shapes — `hibernationTimeout: "0s"`, `require_image()`, the `templateId` field, the literal `400`, `GET /images` — which rot when a field or route is renamed. A link out to the contract as the field-level source of truth would hold better.
- 🟢 **Suggestion:** The SDK's asymmetric retry contract (GETs retry transient mesh failures, non-GETs never, because a repeated finish/report is a conflict) is a protocol invariant a loop author must know, and `experiments.md`'s ingestion section does not state it. (`packages/experiment-sdk/src/experiment_sdk.py:121`)

### Issue Fit
_No linked issue — no closing keyword, plain `#<n>` reference, or issue-numbered branch name. Scope was judged against the PR body._

Both surviving scope gaps are mirrored in `### Findings` above (the `mise.lock` java pin, the unrendered loop description). The body's remaining claims verified positively against the tree: CRD gen 8, `agentPodRestarts`/`agentPodRestartReason`, `readTargetRestart` replacing the deleted `listNamespacedPod`, `require_image()`/`budget()`/retry backoff, size admission, `agents` now required on `composeExperimentsForOwner`, `errorReason`, prompt chips, chart and lima sizing, and the declared test counts.

_Dropped one issue-fit warning: the body's `TEST_OVERVIEW`/`TEST_SCENARIO` claim against the new UI suite is repo baseline, not a regression — 0 of 31 existing `packages/ui` unit suites carry those markers._

- 🟢 **Suggestion:** `SizeNeverFitsError` is a domain error the route translates with `instanceof`, while the neighbouring `createResizeGate` throws a transport-level `TRPCError` from the same file — two mechanisms for one class of budget rejection. (`packages/api-server/src/modules/budgets/services/budgets-service.ts:64`)

_Four further findings are repo baseline, not introduced here, so they are not priced: `cancelInvocations?` was already optional on `main`; `agents/index.js` value imports and the wide export surface are the documented shared-contract pattern; and error-shaped validation plus `c.json({ error })` in a harness route matches `experiment-endpoints.ts` and `mcp-endpoint.ts`. The `targetAdmission` optionality is in `### Findings` above._

Both blocking findings are in `### Findings` above (the named-group mismatch, the `div` disclosure).

- 🟢 **Suggestion:** The copy affordance — `useCopy`, a ghost `icon-xs` `Button`, the `Checkmark`/`Warning`/`Copy` ternary and the `HOVER_ACTION` block — now repeats five times across four files. One `CopyButton` component would also let the missing `role="status" aria-live="polite"` announcement (present in `markdown-code-block.tsx:52`, absent from all three new buttons) be fixed in one place rather than nine.

### Verdict
COMMENT — the hardening is well-evidenced and its ordering is correct where it matters; the blocking items are one supply-chain pin deleted as a side effect, a schema field with no reader, two UI defects, and two doc-consistency gaps.


---

tomkis (approved):

PR #3279, opened 2026-08-12
feat(channels): hand the agent documents attached in Slack, not just images

pilartomas:
Closes #3272.

An agent in Slack could only ever read images. Every other attachment was
filtered out by its label *before* anything was downloaded, so a PDF, a
transcript or a spreadsheet reached neither the agent nor a notice back to the
sender — indistinguishable, from the channel, from the agent ignoring it. The
web UI has accepted arbitrary files all along, so the two entry points behaved
differently for no reason a user could see.

## What changes

An attachment is now one of two things. A **picture** the agent is shown still
rides the prompt as bytes, in the formats a model decodes. A **file** it is
handed is written into the agent's own workspace under
`.uploads/<conversation>/<random>-<name>` and referenced from the prompt as an
ACP `resource_link`, plus a short `<attached-files>` line tying the files to
that message. That is the same mechanism a web-UI chat upload already uses, so
a document arrives the same way whichever surface it came from.

Which of the two an attachment is follows **what the sender sent**, not what the
bytes turn out to be: a screenshot that arrives truncated is still a failed
picture, and naming the formats that work serves its sender better than filing
an image the harness cannot open.

There is **no format list** on the file side. A list is what caused this bug: it
silently swallows the next thing someone attaches. A file is withheld only when
the bytes say it never arrived, or it is too big — and either way the sender is
told, and so is the agent, so it never answers as though it had the file.

## The two open questions in the issue

**Same 50 MB cap as the web UI?** No — 20 MB, and deliberately. The web UI's
bytes go from the browser straight to the pod; a messenger's cross the
api-server, which holds the file and then hands the pod a base64 JSON body,
several times the file's size live at once, in a process whose default limit is
512Mi and which runs the channel workers for the whole install as a single
replica. The cap that matters here is that budget, not what the pod would
accept. 20 MB covers the documents people actually share. Beyond the per-file
and per-message ceilings, the worker holds a single budget across every inbound
path — one process serves every channel in an install, so a burst of attachments
in one conversation is why another's is refused — and the download itself is
bounded, the declared size being the uploading client's claim and sometimes
absent.

**Any file types to exclude on security grounds?** Not by format. The risks that
are real here are elsewhere, and are handled:

- **A refusal page written down as the file.** Slack answers a request it won't
  serve with a 200 and a sign-in page, which for a `.csv` or a transcript is
  indistinguishable from content — and the agent would answer from it. A refusal
  is read where a served page states what it is (its title or first heading, a
[...]

PR #3319, opened 2026-08-14
feat(channels): hand the agent documents attached in Slack, not just images

### LINKED ISSUE #3272: Support document uploads through Slack, not just images
## Context

Users share files with an agent in Slack — PDFs, transcripts, spec docs. Today only images reach the agent. Every other file type is silently dropped: nothing arrives and the user gets no error, so they assume the agent ignored them. This wastes a full round trip and erodes trust in the Slack path.

## Problem

An agent in Slack can only read images — PNG, JPEG, GIF, and WebP. Any other attachment is filtered out before it is even downloaded. The user sees no acknowledgement that the file was rejected. The web UI already accepts arbitrary files up to 50 MB and delivers them to the pod, so the two entry points behave differently for no reason the user can see.

The document types users actually send:

- **Documents** — .pdf, .docx, .txt, .md, .html
- **Presentations** — .ppt, .pptx
- **Spreadsheets** — .xlsx, .csv
- **Transcripts** — .vtt

## Goal

A user attaches any of these files in Slack and the agent receives it, the same as through the web UI. If a file type genuinely cannot be delivered, the user gets a clear message saying so instead of silence.

## User Stories

- As a researcher, I want to drop a PDF into a Slack thread so that the agent can read it without me switching to the web UI.
- As a researcher, I want to share a meeting transcript (.vtt) in Slack so that the agent can summarise it in place.
- As a user, I want a clear notice when a file cannot be delivered so that I am not left guessing whether the agent saw it.

## Open Questions

- Same 50 MB cap as the web UI, or a different limit for Slack?
- Any file types to exclude on security grounds?

pilartomas:
Closes #3272.

An agent in Slack could only ever read images. Every other attachment was
filtered out by its label *before* anything was downloaded, so a PDF, a
transcript or a spreadsheet reached neither the agent nor a notice back to the
sender — from the channel, indistinguishable from the agent ignoring it. The web
UI has accepted arbitrary files all along, so the two entry points behaved
differently for no reason a user could see.

## What changes

An attachment is now one of two things. A **picture** the agent is shown rides
the prompt as bytes, in the formats a model decodes. A **file** it is handed is
written into the agent's own workspace under
`.uploads/<conversation>/<random>-<name>` and referenced from the prompt as an
ACP `resource_link`, plus a short `<attached-files>` line tying the files to that
message. That is the mechanism a web-UI chat upload already uses, so a document
arrives the same way whichever surface it came from.

Which of the two an attachment is follows **what the sender sent**, not what the
bytes turn out to be: a screenshot that arrives truncated is still a failed
picture, and naming the formats that work serves its sender better than filing
an image the harness cannot open.

There is **no format list** on the file side. A list is what caused this bug — it
silently swallows the next thing someone attaches. A file is withheld only when
the bytes are the messenger refusing rather than the file, when the workspace
write fails, or when it exceeds a ceiling; each case is named to the agent, and
to the sender on addressed turns.

## The two open questions in the issue

**Same 50 MB cap as the web UI?** No — 20 MB per file, deliberately. The web
UI's bytes go from the browser straight to the pod; a messenger's cross the
api-server, which holds the file and then hands the pod a base64 JSON body,
several times its size live at once, in a process whose default limit is 512Mi
and which runs the channel workers for the whole install as a single replica. The
cap that matters here is that budget, not what the pod would accept.

Beyond the per-file and per-message ceilings, the worker holds one budget across
every inbound path: reserved at admission (a check a caller acts on later is read
across the download, so every concurrent fetch would pass the same snapshot) and
settled to what actually arrives (a declared size is the uploading client's claim
and can understate the file). `attachment-budget.ts` carries that reasoning and
its contract tests. The download itself is bounded too, since a declared size is
sometimes absent entirely.

**Any file types to exclude on security grounds?** Not by format. The risks that
are real here are elsewhere, and are handled:

- **A refusal page written down as the file.** Slack answers a request it will
  not serve with a 200 and a sign-in page, which for a `.csv` is
  indistinguishable from content — and the agent would answer from it. Markup is
  judged on being recognisably a refusal (a heading, a form or refresh target, an
  auth-shaped API error), so a genuine `.html` or `.xsl` still arrives.
- **An attachment *being* something.** In a shared channel the binding is the
  whole authorization, so the uploader is anyone the workspace admits. Files land
  behind a random prefix in a per-conversation directory, never at a path the
  harness reads on its own.
- **A filename as instructions.** The name is escaped where it is read, so it
  cannot close the prompt's framing and forge a block in the same vocabulary the
  turn contract uses, and cannot hand Slack a `<!channel>` to act on in a message
  this bot posts.

## Changes outside the linked issue

- **Three web-UI copy fixes.** Two oversized-upload toasts (chat composer, files
  pane) quoted "10 MB" against the 50 MB cap they guard; both now derive the
  number from the cap. A preview error quoted a 10 MB limit for a cap that
  component does not own, and no longer quotes a number at all.
- **A message whose images exceed the image cap** is now relayed with those
  images withheld and named, instead of being dropped whole — that cap bounds
  what one prompt may carry and says nothing about a file written to disk.
- **A picture can be withheld for the worker's memory budget**, which is new for
  the image path: a picture is held across the same cold-pod wait and rides the
  prompt encoded, so narrowing the budget to files would leave the larger cost
  unbounded.

## Excluded surface

**A read-along (ambient) turn tells the agent, not the channel.** Withheld
attachments are named to the sender on mention and DM turns; on read-along turns
nobody summoned the agent, and an ephemeral would notify someone who did not ask,
so the agent alone is told. A deliberate trade against the issue's "clear notice"
goal for that one surface.

## Known limitation, not addressed here

`.uploads` has no cumulative quota and nothing prunes it, so a channel member
could fill an agent's volume over many messages. That is pre-existing — web-UI
uploads land in the same tree and are equally permanent — but this widens who can
do it from the owner to anyone in a bound channel. A retention policy is its own
change and wants its own decision.

## Verification

- 972 unit tests in `api-server`, 970 pass; the 2 failures are pre-existing in
  `cluster-ca-probe.test.ts` (Node 26 refuses an IP as a TLS `servername`) and
  are untouched by this branch.
- `tsc`, `eslint`, `prettier`, `docs:check` and `common:check:comment-types` all
  clean; CodeQL clean on the branch's content.
- Coverage: delivery and linking, the path shape, every type the issue names with
  no format gate, HTML that is genuinely markup, a refusal page, an empty file,
  both size ceilings either side of the download, a staging failure, per-file
  audit attribution, deliver-once across a failed resume, several files on one
  message, a coalesced read-along batch, and a DM. The budget's contract has its
  own suite.
- Not verified against a live cluster — no Slack workspace was driven end to end.

## Manual verification worth doing

1. `@mention` the agent in a bound channel with a PDF and a question about it —
   the agent should read it from `/home/agent/.uploads/...`.
2. Attach a `.vtt` transcript and ask for a summary.
3. Attach something over 20 MB — expect a notice naming the limit, with the
   question still answered.
4. On an install without `files:read`, attach a `.csv` — expect the notice to
   name the missing permission rather than the agent summarising a login page.

---

_Replaces #3279, which carried five rounds of review history on a stale base.
All findings from those rounds are included here._


packages/api-server/src/modules/channels/infrastructure/agent-workspace-files.ts
```diff
@@ -0,0 +1,40 @@
+import { createTRPCClient, httpLink } from "@trpc/client";
+import type { AppRouter } from "agent-runtime-api";
+import { podBaseUrl } from "../../agents/infrastructure/k8s.js";
```
dam-code-guardian: 🔴 **Critical:** `podBaseUrl` is imported directly from `agents/infrastructure/k8s.js`, crossing the module boundary into `agents`' internal infrastructure. The `channels` module must depend only on `agents/index.ts`.

**Fix:** Export `podBaseUrl` from `agents/index.ts` (or better: accept the pod base URL or the tRPC client as a constructor argument to `createAgentWorkspaceFiles`). Either removes the direct dependency on another module's internals.

packages/api-server/src/modules/channels/inbound-file.ts
```diff
@@ -0,0 +1,101 @@
...
+const HEADING =
+  /<title[^>]*>([\s\S]{0,300}?)<\/title>|<h1[^>]*>([\s\S]{0,300}?)<\/h1>/g;
+
+const REFUSAL_HEADING =
+  /\bsign[\s-]?in\b|\blog[\s-]?in\b|not authori[sz]ed|access denied|permission denied|forbidden|^\s*slack\s*$/;
+
+const SIGN_IN_URL =
+  /^https?:\/\/(?:[a-z0-9-]+\.)*slack\.com\/(?:signin|workspace-signin)(?:[/?#]|$)/;
+
+const ATTRIBUTE = /([a-z-]+)\s*=\s*(?:"([^"]*)"|'([^']*)'|([^\s>]+))/g;
```
dam-code-guardian: 🟡 **Warning:** `ATTRIBUTE` is a module-level `/g` regex whose `lastIndex` is explicitly set inside the `targetElements` generator (line 37). Today the generator is always consumed synchronously (no `await` in `hasSignInTarget` or its callers), so concurrent calls cannot interleave. But the shared state is one accidental `await` away from a hard-to-reproduce bug where a second `looksLikeSignInPage` call resets `lastIndex` and corrupts the first iterator mid-scan.

**Fix:** Declare `ATTRIBUTE` inside `targetElements` as a local `const` — each call gets its own regex instance with its own `lastIndex`.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `62e46b84`

## PR #3319: feat(channels): hand the agent documents attached in Slack, not just images
**Author:** pilartomas | **Branch:** feat/slack-attachments → main | **Changes:** +1874 −161 (32 files)

### Summary
Extends Slack attachment handling to deliver non-image files (PDFs, spreadsheets, transcripts, etc.) to the agent's workspace as `resource_link` blocks, using the same path as web-UI uploads, with a memory budget, per-file/per-message caps, sign-in-page detection, and path sanitisation. Also aligns web-UI upload toasts to use the cap constant and delivers images withheld by the image cap separately from documents.

### Findings
- 🔴 **Critical:** Cross-module boundary violation in `agent-workspace-files.ts` — see TypeScript Engineering Review (`packages/api-server/src/modules/channels/infrastructure/agent-workspace-files.ts:3`)
  **Fix:** Export `podBaseUrl` from `agents/index.ts`, or inject the URL as a constructor argument so `channels` depends only on the agents module's public surface.
- 🟡 **Warning:** Shared `ATTRIBUTE` module-level `/g` regex mutated inside generator (`inbound-file.ts:26`) — see TypeScript Engineering Review
  **Fix:** Declare `ATTRIBUTE` inside `targetElements` as a local regex to eliminate shared-state risk.
- 🟡 **Warning:** Inline `AttachmentChip` subcomponent in parent file — see React UI Engineering Review (`chat-input.tsx:231`, pre-existing)
  **Fix:** Extract to `attachment-chip.tsx` in the same directory.
- 🟡 **Warning:** Template literals used for conditional class composition instead of `cn()` — see React UI Engineering Review (`chat-input.tsx:162`, `chat-input.tsx:215`, pre-existing)
  **Fix:** Replace both template literals with `cn(...)` calls.

✅ No findings.

### Issue Fit
🟢 Read-along turns give no sender notice when an attachment is withheld — the issue asks for clear notice, and the PR explicitly documents this as a deliberate trade-off (ambient turns: agent is told, sender is not). Scope is met for mention and DM turns; the ambient surface departure is declared.

- 🔴 **Critical:** `createAgentWorkspaceFiles` imports `podBaseUrl` directly from `../../agents/infrastructure/k8s.js`, bypassing `agents/index.ts`. (`packages/api-server/src/modules/channels/infrastructure/agent-workspace-files.ts:3`)
  **Fix:** Export `podBaseUrl` from the agents module's public index, or pass the pod base URL or a pre-built tRPC client as a constructor argument so `channels` depends on `agents`' public surface only.
- 🟡 **Warning:** Module-level `ATTRIBUTE` regex (with `/g` flag) has its `lastIndex` explicitly set inside the `targetElements` generator. A future `await` in the call path that suspends the generator between yields would let a concurrent `looksLikeSignInPage` call reset `lastIndex` and corrupt the iterator. (`packages/api-server/src/modules/channels/inbound-file.ts:26`)
  **Fix:** Declare `ATTRIBUTE` inside `targetElements` as a local `const` so each invocation gets its own stateless regex instance.

- 🟡 **Warning:** `AttachmentChip` subcomponent is defined in `chat-input.tsx` — the project convention is one exported component per file. (`packages/ui/src/modules/sessions/components/chat-input.tsx:231`, pre-existing code)
  **Fix:** Extract `AttachmentChip` to `attachment-chip.tsx` in the same directory and import it.
- 🟡 **Warning:** Template literal used for conditional class composition at lines 162 and 215 instead of the `cn()` utility. (`packages/ui/src/modules/sessions/components/chat-input.tsx:162`, `chat-input.tsx:215`, pre-existing code)
  **Fix:** Replace both `className={...}` template literals with `cn(...)` calls.
- 🟢 **Suggestion:** Attachment list uses array-index keys (`key={i}`) — stable keys based on file id or name would avoid spurious re-renders on removal. (`packages/ui/src/modules/sessions/components/chat-input.tsx:168`, pre-existing code)

### Verdict
REQUEST_CHANGES — the cross-module boundary violation in `agent-workspace-files.ts` must be resolved before merge.

---

pilartomas:
Fixed in b50a608c.

**🔴 Cross-module boundary.** Right, and the injection route rather than widening
`agents`' public surface: the adapter now takes a tRPC URL, resolved in
`bootstrap.ts` where `podBaseUrl` already lives. `channels` no longer imports
anything from `agents/infrastructure` (verified by grep across the module), and
the adapter is constructible in a test without a namespace.

**🟡 Shared `/g` regex.** Also right, and worth fixing even though the current
call path is synchronous — a module-level pattern whose `lastIndex` is assigned
inside a generator is a hazard that only shows up once someone adds an `await`.
It is local to `targetElements` now.

**The three `chat-input.tsx` items** — inline `AttachmentChip`, template literals
instead of `cn()`, and the index keys — I've left alone, and I'd rather say why
than silently skip them. All three are pre-existing (you flagged them as such),
and this branch touches that file for exactly one line: the oversized-upload
toast, which quoted 10 MB against the 50 MB cap it guards. Extracting a
component and rewriting `className` composition in a file I otherwise didn't
change would put refactoring of untouched code inside a feature PR, where a
reviewer can't tell the fix from the cleanup. Happy to file them as their own
issue — say the word and I will, or if you'd rather they ride along here, tell me
and I'll add them.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — 💬 Code Review @ `b50a608c`

## PR #3319: feat(channels): hand the agent documents attached in Slack, not just images

### Changes since last review

The two blocking findings from the first review are addressed:

**🔴 → ✅ Cross-module boundary**: `bootstrap.ts` now constructs `createAgentWorkspaceFiles` inline with the tRPC URL built from `podBaseUrl` (already imported there) and passes a factory `(agentId) => createAgentWorkspaceFiles(url)` to `createSlackWorker`. `channels/infrastructure/slack.ts` only imports from `./agent-workspace-files.js` via the `AgentWorkspaceFilesFactory` type. The boundary is clean.

**🟡 → ✅ Module-level `/g` regex**: `ATTRIBUTE` is now `const attributePattern = .../g` declared at the top of `targetElements` — a new local variable on every generator invocation. Concurrent calls no longer race on a shared `lastIndex`.

---

### Full review

#### Core implementation

The budget/reserve/settle/release lifecycle in `attachment-budget.ts` is correct: `settle()` adjusts the held count bidirectionally before the claim owner has called `release()`; the `live` flag makes `release()` idempotent and prevents `settle()`-after-`release()` from re-charging bytes with no owner. The initial `reserve(stagedFootprint(f.size || MAX_FILE_BYTES))` is conservative (overstates the hold for zero-declared-size files until download confirms), which is the right direction for a shared-across-turns budget. `HELD_BYTES_CAP = stagedFootprint(3 × TOTAL_FILE_BYTES_CAP)` accounts for both raw bytes and the base64 payload on the tRPC call body — the calculation is correct.

`inboundFilePath` sanitizes both the conversation ID and filename through `sanitizeSegment`, which strips traversal sequences and leading dots; `opts.unique` (8 hex chars from `randomUUID()`) is caller-controlled and contains only `[0-9a-f]`, so it passes safely without sanitization. The path traversal tests cover `../../etc` conversation IDs and `../../../.bashrc` filenames.

`looksLikeSignInPage`: the `attributePattern` local to `targetElements` correctly handles `>` inside quoted attribute values (`[^"]*` captures `a>b` as the value of `title`, and the break-on-`>` check only fires on the text *between* attributes, not inside them). The 8 KB head limit is appropriate; all regexes use negated character classes or bounded quantifiers with no catastrophic backtracking risk.

`deliverTurnFiles`' lazy `delivery ??= …` pattern defers workspace writes until `buildResumePrompt` or `buildFreshPrompt` is called — which happens only after the session resume attempt, preventing double-staging when a failed resume re-runs the turn. The test `"delivers the file once when a failed resume re-runs the turn"` pins this.

The ambient batch budget accounting is correct: `batchFiles` coalesces files from batched messages and drops what would exceed the per-turn cap, but `msg.release()` in the `finally` block covers all claims from each message's `fetchSlackAttachments` call — including files dropped by `batchFiles` — because `FetchAttachmentsResult.release()` iterates the full `claims[]` array from that fetch.

`bolt-slack-gateway.ts` streams the download with a running byte counter and throws `FileTooLargeError` mid-stream, immediately cancelling the body — large files are not fully read.

#### New finding

🟡 **`packages/ui/src/modules/files/components/file-preview-body.tsx:103` — incorrect attribution in copy**

> "The agent refused to read it at this size — open it in a terminal instead"

`tooLarge: true` is set in `queries.ts` when the tRPC call throws `PAYLOAD_TOO_LARGE` (HTTP 413) — the file content API refused to send the payload to the browser. The AI agent is not involved; it has already read the file successfully. In this codebase's UI vocabulary, "the agent" consistently refers to the Claude AI. A user who sees this will reasonably interpret it as the AI declining a task it could otherwise do.

**Fix:** attribute the limit to the viewer, not the agent. E.g.:
```tsx
<p className="mt-1 text-[11px]">
  Too large to display in the browser — open it in a terminal instead
</p>
```
Or, if the intent is to retain "agent" as shorthand for the system: "File too large to preview — the browser can't load this size. Open it in a terminal instead."

#### Pre-existing items (author declined, justified)

The branch touches two lines in `chat-input.tsx` (the `formatBytes` import and the toast substitution). The pre-existing `chat-input.tsx` items from the first review — inline `AttachmentChip` component, template literals for conditional `className`, and `key={i}` on the attachment list — remain as previously noted, unchanged by this commit.

#### Skill audit

| Skill | Result |
|---|---|
| doc-drift | 2 minor notes: (1) `docs/architecture/channels.md` workspace-write bullet could be read as Telegram-inclusive; (2) the scopes section doesn't call out that a missing `files:read` degrades to per-file notices rather than feature absence. Neither is a blocking doc gap — the PR updates the scope capabilities table. |
| issue-fit | Implementation matches issue #3272. Both open questions (cap size, security exclusions) are answered with rationale in the PR body. Ambient turn notice gap is the documented "Excluded surface" trade-off. |
| typescript-engineering | NONE |
| react-ui-engineering | 1 finding — `file-preview-body.tsx:103` (above) |

---

### The approval bar

One open 🟡 to address before this merges:

- 🟡 `packages/ui/src/modules/files/components/file-preview-body.tsx:103` — rephrase the too-large message to not attribute the limit to the AI agent. Fix: attribute it to the browser/preview viewer as shown above.

The `formatBytes(MAX_UPLOAD_BYTES)` substitutions in `chat-input.tsx` and `use-file-mutations.ts` are correct (both now read "19.1 MB" → actually 50 MB, whatever `MAX_UPLOAD_BYTES` evaluates to, rather than a hardcoded stale value).

```json
{
  "findings": [
    {
      "severity": "yellow",
      "location": "packages/ui/src/modules/files/components/file-preview-body.tsx:103",
      "title": "Too-large copy misattributes HTTP 413 to the AI agent",
      "fix": "Rephrase to attribute the limit to the preview/browser, not the AI agent. E.g. 'Too large to display in the browser — open it in a terminal instead'"
    }
  ]
}
```

pilartomas:
Fixed in 0b9ea23d.

**🟡 Preview copy.** You're right, and it was my wording: `tooLarge` comes from a
413 on the file-content API, which the agent has nothing to do with — it had
already read the file. Since "the agent" means the AI everywhere else in this UI,
the copy read as the AI declining something it can do. It now reads "Too large to
load in the browser — open it in a terminal instead."

Also took the first doc-drift note, since it was a real accuracy problem rather
than a style one: the attachment bullet and the new workspace-store bullet in
`channels.md` both read as though either messenger delivers files this way, when
Telegram drops an inbound attachment at its gateway before the worker sees it.
Both now say Slack. Paid for the words by shortening my own sentence — that page
sits at 39,989 of 40,000.

Left the second note (that a missing `files:read` degrades to per-file notices
rather than the feature disappearing). The scope-capabilities table already
carries `files:read`, and spelling out the degradation shape is per-file
behaviour that belongs in the code — where it is, in the refusal gate — rather
than on a page with 11 characters of headroom.

pilartomas:
Merged `main` (277b4fa6) — #3320 edited `channels.md`'s `Last verified` line, which conflicted. Merged rather than rebased on purpose: the review label was applied at 09:18Z so a pass may have been in flight, and force-pushing under a live review is the one thing the PR loop here forbids. All five architecture pages now read 2026-08-14; `channels.md` sits at 39,954 of 40,000.

dam-code-guardian (approved):
**Re-review at `277b4fa6`** (full re-review; label trigger)

Both prior rounds are fully resolved — the round-1 🔴 cross-module boundary is gone (factory-injected via `bootstrap.ts`), the round-1 🟡 shared `/g` regex is gone, and the round-2 🟡 copy on `file-preview-body.tsx:103` is gone. The implementation is clean.

---

## Findings

🟡 **`MAX_FILE_BYTES` exists in three packages with different values**
`packages/agent-runtime/…/local-skill-repository.ts:20` exports it at 2 MiB; `packages/ui/…/use-upload-staging.ts` declares it module-locally at 2 MiB; and this PR adds `packages/api-server/…/inbound-file.ts:14` at 20 MB. All three are in separate packages so TypeScript will never catch a wrong-package import, but the name reads as "the" file-size limit. Consider a more specific name for the new constant — e.g. `MAX_INBOUND_ATTACHMENT_BYTES` — to make the three values distinguishable at a glance.
**Fix:** rename `MAX_FILE_BYTES` in `inbound-file.ts` (and its two call sites in `slack.ts`) to a more scoped name.

🟡 **`channels.md` Mermaid diagram omits the new direct file-write path**
`agent-workspace-files.ts` constructs a tRPC client pointing straight to the pod without going through the proxy (`http://${podBaseUrl(agentId, namespace)}/api/trpc`). `platform-topology.md` has the new protocols row, but the Slack flow diagram in `channels.md` only shows the ACP relay path — the `SlackWorker → agent-runtime pod (tRPC, direct)` edge for file writes is missing. Worth a `SW -- file write (tRPC, direct) --> POD` edge.

🟡 **"Channels touch three stores" but only two are bulleted**
`persistence.md` Persistence touchpoints header was bumped from two to three, but the section still bullets only Postgres and the per-agent PVC. Redis (leader election) was apparently one of the original "two" but has no dedicated bullet. Either add a Redis bullet or keep the count at two for the stores the section actively describes.

🟡 **Live-cluster verification not run**
The PR body explicitly flags this. The full integration path (Slack webhook → api-server download → tRPC pod write → PVC) is new; the test suite covers it well in-process but the end-to-end path hasn't been driven. Noting it here as part of the approval bar.

🟢 **Budget contract is sound**
`settle()` is now bidirectional (round-4 finding); `settle()` after `release()` is a no-op (`live` flag); `reserve()` guards NaN/negative/Infinity; `NO_CLAIM` correctly handles zero-byte reserves.

🟢 **Claim lifecycle correct in all paths**
Every error branch in `fetchSlackAttachments` calls `claim.release()` before the `continue`. `fetched.release()` is in a `try/finally` wrapping `relaySharedTurn`. Ambient batch releases via `for (const msg of batch) msg.release()` in `finally`. The lazy `deliverFiles` (`??=`) pattern is safe for multiple callers.

🟢 **`looksLikeSignInPage` covers the common failure shapes**
Slack own sign-in (`SIGN_IN_URL`), JSON auth error codes (`AUTH_ERROR_CODE`), `<title>/<h1>` refusal headings (including "access denied", "forbidden"), and `form[action=…/signin]` are all handled. The `og:url` false-positive from round-4 is correctly excluded (only `http-equiv="refresh"` `meta` tags are tested). Non-English sign-in pages won't match, but that's an acceptable and documented gap.

🟢 **Path sanitization prevents traversal**
`sanitizeSegment` strips to `[A-Za-z0-9._-]+` and removes leading dots, so channel IDs and timestamps produce safe filesystem segments.

🟢 **`downloadFile` signature updated in all three implementations**
`bolt-slack-gateway.ts`, `fake-slack-gateway.ts`, and the `SlackGateway` interface all carry the new `maxBytes` parameter consistently.

🟢 **No code comments in the diff**
New `TEST_OVERVIEW` / `TEST_SCENARIO` JSDoc blocks use the two registered `COMMENT_TYPES` from `scripts/strip-comments.mjs:35` and pass `common:check:comment-types`. No unregistered comments in any diff-added code.

---

## Skill sections

All five edited architecture pages have `Last verified` bumped to 2026-08-14. ✅

🟢 **channels.md** — attachment bullet rewritten accurately; budget tiers (per-transfer, per-message, per-turn, worker-wide) match the code.

🟡 **channels.md** — Mermaid diagram missing `SlackWorker → agent-runtime pod (tRPC, direct)` edge (see finding above).

🟡 **persistence.md** — "three stores" header count inconsistent with two bulleted stores (see finding above).

🟢 **logging.md** — `channel.file.delivered` event description ("by name and byte count") matches `securityLog` call at the write site.

🟢 **platform-topology.md** — tRPC proxy split correctly; "no bearer, NetworkPolicy is the whole gate" accurately reflects `agent-workspace-files.ts` (no `Authorization` header).

🟢 **security-and-credentials.md** — "speaker with no platform identity is a writer to persistent state" is accurate.

---

### Issue Fit

🟢 Core ask fully delivered — non-image Slack attachments reach the agent workspace as `resource_link` ACP blocks, matching the web-UI path.

🟢 Both open questions from the issue are answered in the body with reasoned decisions (20 MB cap, no format exclusions).

🟢 All three web-UI copy fixes are declared and present.

🟡 No live-cluster verification run (explicitly declared in the PR body; end-to-end path is new).

---

🟢 **`attachment-budget.ts`** — reserve/settle/release contract sound; NaN/negative/zero guards correct; `settle()` after `release()` is no-op.

🟢 **`inbound-file.ts`** — `targetElements` generator correctly handles attribute order via `Map`; `looksLikeSignInPage` short-circuits before `canReadFiles` when sign-in page detected.

🟢 **`slack.ts`** — All `claim.release()` paths present; `deliverFiles` lazy `??=` is safe; sequential picture loop ensures settle-before-next-reserve; `heldBudget` covers both addressed and ambient paths.

🟡 **`MAX_FILE_BYTES` name collision** (see finding above).

---

🟢 `formatBytes(MAX_UPLOAD_BYTES)` is correct; `MAX_UPLOAD_BYTES` is declared once at 50 MB and both toast sites now derive from it.

🟡 `formatBytes(50 * 1024 * 1024)` renders as `"50.0 MB"` (pre-existing `.toFixed(1)` behavior in the formatter); consistent with other size displays in the UI.

🟢 `file-preview-body.tsx` copy change is accurate — the tooLarge threshold is server-controlled and decoupled from the upload limit.

---

## Approval bar

No 🔴 blocking findings. Four 🟡 items:

| # | Item | File | Resolution |
|---|------|------|-----------|
| 1 | `MAX_FILE_BYTES` name collision (3 packages, different values) | `inbound-file.ts` | Rename to more specific constant; can merge as-is if the team decides cross-package confusion is acceptable |
| 2 | `channels.md` Mermaid diagram missing direct tRPC file-write edge | `docs/architecture/channels.md` | Add diagram edge; docs-only follow-up is fine |
| 3 | "three stores" header but two bullets | `docs/architecture/persistence.md` | Fix count or add Redis bullet |
| 4 | No live-cluster end-to-end verification | — | Acknowledged in body; manual smoke test per the checklist there before merge |

**APPROVE** — no blocking findings. The four 🟡 items above are worth addressing before or shortly after merge; none of them block correctness or security.



```json findings-json
[
  {
    "id": "max-file-bytes-name-collision",
    "severity": "warning",
    "summary": "MAX_FILE_BYTES exists in 3 packages with different values (20 MB, 2 MiB, 2 MiB)",
    "file": "packages/api-server/src/modules/channels/inbound-file.ts",
    "line": 14,
    "fix": "Rename to MAX_INBOUND_ATTACHMENT_BYTES (and its 2 call sites in slack.ts) for clarity"
  },
  {
    "id": "channels-md-missing-direct-trpc-edge",
    "severity": "warning",
    "summary": "channels.md Mermaid diagram missing direct SlackWorker→agent-runtime tRPC file-write edge",
    "file": "docs/architecture/channels.md",
    "fix": "Add 'SW -- file write (tRPC, direct) --> POD' edge to the Slack flow diagram"
  },
  {
    "id": "persistence-md-store-count",
    "severity": "warning",
    "summary": "'Channels touch three stores' but only two stores are bulleted in Persistence touchpoints",
    "file": "docs/architecture/persistence.md",
    "fix": "Either add a Redis bullet or revert the count from three to two"
  },
  {
    "id": "no-live-cluster-verification",
    "severity": "warning",
    "summary": "No live-cluster end-to-end verification; full integration path (Slack→api-server→tRPC→PVC) is new",
    "fix": "Run manual smoke test per PR body checklist before merge"
  }
]
```

PR #3320, opened 2026-08-14
feat(connections): show Slack in the standard connection catalogue

### LINKED ISSUE #3310: Show the Slack connection in the standard catalog
## Context

The Slack connection is fully built: it's always present in the server-side catalog, and once connected it gives agents access to Slack's MCP server via OAuth. But the UI hides it behind the **Advanced connections** experimental flag, which is off by default and lives in a settings tab that only appears after a hidden gesture (five taps on the version string). In practice no one finds it.

## Problem

A user who wants their agents to work with Slack opens the connection catalogue and finds no Slack entry at all — not greyed out, simply absent. The same is true in the sandbox creation wizard. The feature works, but it's unreachable through normal use of the product, so to users it doesn't exist.

## Goal

Slack appears in the Apps tab of the connection catalogue for every user, with no flag involved — same place and same look as the other standard connections. A user can discover it, connect their workspace, and grant it to agents without knowing experimental flags exist.

## User Stories

- As a user, I want to see Slack in the connection catalogue so that I can connect my workspace without being told about a hidden flag first.
- As a user creating a sandbox, I want Slack offered in the connections step so that I can give a new agent Slack access from the start.

## Scope

- Only Slack graduates. The other hidden templates (Spotify, YouTube, Google services, custom client-credentials, GitHub apps) stay behind the flag.
- Existing Slack connections already render regardless of the flag; nothing changes for them.

## Open Questions

- When the operator hasn't registered a Slack OAuth app, the connect form asks the user to paste their own client id and secret. Is that experience acceptable as the default-visible path, or should we polish it as part of this?

pilartomas:
Closes #3310.

Slack was already fully built as a connection — always in the server-side catalog, always creatable via the API and the CLI — but the UI dropped it from the offered catalogue unless the user had turned on the *Advanced connections* experimental flag, which lives behind a hidden settings tab. The feature worked and nobody could find it.

## What changed

**Slack graduates out of the internal-only set.** `INTERNAL_ONLY_TEMPLATE_IDS` loses its `slack` entry. That set is the only gate: `filterOfferedTemplates` has exactly one consumer, `useCatalogGroups`, and all four catalogue surfaces route through it — the Connections settings page, the sandbox creation wizard's connections step, a sandbox's own Connections section, and a knowledge base's config view. Slack's template is `category: "app"` and has no curated provider entry, so it falls through to the generic branch and lands as a standalone card in the **Apps** tab, with the Slack mark already present in the icon registry. Everything else stays behind the flag.

**The flag's own description was wrong.** It read "(Google services, Slack, Spotify, custom client-credentials)" — it named Slack, and it had never mentioned YouTube or the two GitHub App installation templates. It now matches the hidden set exactly. "GitHub App installations" rather than "GitHub Apps", because "GitHub app" is already the catalogue subtitle for the ordinary, always-visible GitHub connection.

**A hint for users bringing their own Slack app** (see the open question below).

**A unit test** pins the split: Slack offered without the flag, the other five ids and the `google-` prefix still hidden. There was no test over this module before, so a regression would have been silent.

## The issue's open question

> When the operator hasn't registered a Slack OAuth app, the connect form asks the user to paste their own client id and secret. Is that experience acceptable as the default-visible path?

**Yes, on precedent — no code change needed for the form itself.** `github` and `github-enterprise` are OAuth templates using the identical creds-optional pattern, neither has ever been internal-only, and the chart ships `oauthAppDefaults.github.clientId: ""`. So a stock install *already* renders exactly this paste-your-own-client-id form, with the same "Register an OAuth app at the provider" callout and redirect-URI copy button, for two templates that have always been standard. Slack is not the first.

**One respect in which Slack genuinely is worse, and it's cheap to close.** A GitHub OAuth app declares no scope list — the scopes ride on the authorize request, so the existing callout's two instructions are complete. Slack declares its scopes *on the app*, and this template requests 26 MCP user-token scopes. A user who registers their own app and misses a scope gets no connect-time error: the connection lands green and fails later, mid-agent-turn, when `mcp.slack.com` refuses. That failure is notoriously illegible here — Slack answers a scope gap with a 200 and a sign-in page rather than an error. So this adds one line of copy to the existing per-template hint table, on the Client ID field, naming the step. Copy only, no behaviour change; happy to drop it if you'd rather keep this PR to the graduation alone.

## Scope and risk

Purely UI discoverability — there was never a server-side gate. `buildCatalog` always registered the Slack template, `listTemplates` never filtered, and `connections.create` never consulted the internal-only set, so anyone with API access or the flag could already create a Slack connection with identical effect. Authorization is unchanged: the credential is the user's own, created by their own deliberate act, and its contributions only materialize on sandboxes they explicitly grant it to.

No e2e impact — every Playwright catalogue selector is test-id based and confined to the custom-headers tab, and the Apps tab badge counts existing connections rather than offered templates, so it does not move.

## Docs

`connections.md` §Internal-only templates no longer names Slack. Two pre-existing drifts in that same sentence are fixed while it's open: the list never mentioned the three machine-to-machine templates that were also hidden (now covered by role rather than by id, so it stops drifting on the next addition), and the surfaces clause named two of the four surfaces the filter actually reaches, as an exhaustive "both … and" (now stated at the level of the concept). The reword nets *out* 38 characters, taking the page from 28 to 51 characters of headroom under the 40k cap.

`channels.md` loses the clause asserting Slack-as-a-Connection is "still withheld from regular users" — dropped rather than restated, since `connections.md` owns catalogue visibility and this page is the cross-link.

## Verification

`ui:check` (tsc, eslint, prettier), `ui:test` (290/290, including the new file), `docs:check` (ADR index, ADR immutability, page-size cap) and `common:check:comment-types` all pass.


dam-code-guardian (commented):
## Code Guardian 🛡️ — PR #3320 review

**feat(connections): show Slack in the standard connection catalogue**

This is a clean discoverability change: Slack moves from the `INTERNAL_ONLY_TEMPLATE_IDS` set into the default catalogue, available to all users without the `advanced-connections` feature flag.

---

### Summary

The diff removes `"slack"` from `INTERNAL_ONLY_TEMPLATE_IDS` in `internal-only.ts`, adds a `clientId` setup hint for Slack in `field-copy.ts`, updates the `features-tab.tsx` copy to reflect the new set of hidden templates, and bumps both architecture docs. A new unit test confirms the expected gate behaviour for all five remaining internal-only IDs and the `google-` prefix.

---

### Findings

🟡 **`packages/ui/src/modules/features/components/features-tab.tsx:45` — map render block exceeds extraction threshold**

The `.map()` block for `FEATURE_ROWS` is ~18 lines of JSX, crossing the repo's ~10-line threshold at which list-item rendering is extracted to its own component. The copy update in this PR touches the file, which brings it in scope. No logic defect.

**Fix:** Extract the `<label>…</label>` block to a `FeatureRowCard` component (sibling file or inline export in the same module), accepting `row: FeatureRow` and the `setFeature` mutation as props.

---

### Approval bar

🟡 above is a minor style finding. The core functional change is correct and safe.

- ✅ `"slack"` correctly removed from `INTERNAL_ONLY_TEMPLATE_IDS`; the five remaining IDs + `google-` prefix are undisturbed
- ✅ `filterOfferedTemplates` consumer (`useCatalogGroups`) threads `showInternal` from the `advanced-connections` flag; no new Slack-visible path bypasses it
- ✅ Slack template is `category: "app"` with icon registered; `hintFor("slack", "clientId")` is consumed by `template-field-input.tsx`
- ✅ `features-tab.tsx` description updated to match the actual remaining hidden set
- ✅ Both architecture docs updated; `Last verified` bumped to 2026-08-14 in both
- ✅ New unit test covers `isInternalOnlyTemplate` and `filterOfferedTemplates` for all affected IDs
- ✅ No code comments added or modified in any code file (repo convention)
- ✅ Closes #3310 with correct keyword and exactly in scope

**Merge when:** the 🟡 `features-tab.tsx` map-block extraction is addressed (in this PR or a follow-up marked as such).

---

### Skills

| Skill | Trigger | Result |
|---|---|---|
| doc-drift | always | ✅ aligned — both architecture pages correctly updated |
| issue-fit | always | ✅ 0 findings — closes #3310, in-scope only |
| typescript-engineering | `.ts` files changed | ✅ 0 findings |
| react-ui-engineering | `.tsx` files changed | 🟡 1 finding (map block extraction, above) |

---

```json
{"findings-json":[{"severity":"yellow","file":"packages/ui/src/modules/features/components/features-tab.tsx","line":45,"title":"map render block exceeds extraction threshold","fix":"Extract the label...label block to a FeatureRowCard component accepting row: FeatureRow and the setFeature mutation as props","source":"react-ui-engineering"}]}
```

pilartomas:
Addressed the 🟡 finding in 76b7278b: the `FEATURE_ROWS` map block is now a `FeatureRowCard` component taking `row`, `enabled` and `onToggle`, so the map reads as a single element. `ui:check`, `ui:test` (290/290) and `common:check:comment-types` all still pass.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `76b7278`

## PR #3320: feat(connections): show Slack in the standard connection catalogue
**Author:** pilartomas | **Branch:** feat/slack-standard-catalog → main

### Summary

Slack graduates from `INTERNAL_ONLY_TEMPLATE_IDS` to the standard catalogue. The prior 🟡 (map-block extraction in `features-tab.tsx`) is addressed: the inline `<label>` block is now a `FeatureRowCard` component.

### Changes since last review
Previous HEAD: `438cb5f` (2026-08-14T08:35:23Z) — verdict COMMENT

- ✅ **Fixed:** map render block exceeds extraction threshold (`features-tab.tsx:45`) — extracted to `FeatureRowCard`
- 🆕 **New:** 🟢 suggestion below (pre-existing a11y pattern)

### Findings

- 🟢 **Suggestion (pre-existing):** `packages/ui/src/modules/features/components/features-tab.tsx:47` — The `<label>` wrapping `Switch` without an explicit `htmlFor`/`aria-labelledby` link was already present before this PR; the refactor preserved the pattern. Screen-reader behaviour with a wrapping `<label>` and a Radix `<button role="switch">` varies. Suggest a separate follow-up to add `aria-labelledby` on `Switch` or convert to explicit `<label htmlFor>` pairing.

✅ No drift — `connections.md` and `channels.md` both updated with `Last verified: 2026-08-14`.

### Issue Fit
✅ Closes #3310 in full scope. `FeatureRowCard` extraction is a review-requested fix (see PR comment), not scope creep.

✅ No findings — three changed files are all in `packages/ui` (client package). `ReadonlySet`/`readonly` exports correct; contract-package imports follow the pattern.

🟢 Pre-existing accessibility pattern (wrapping `<label>` + `Switch` without explicit association) — see Findings above. `FeatureRowCard` extraction and description-string update are clean.

### Verdict
APPROVE — prior 🟡 fixed, no blocking findings.

---

PR #3207, opened 2026-08-06
feat(ci): stop rebuilding unchanged workload images on every deploy

xjacka:
## Motivation

Workload images (nous, openevolve, shinkaevolve, gepa, skydiscover) are slow, rarely-changing builds, yet every push to main rebuilt all of them (×2 architectures) because `RESOLVE_NO_REUSE=1` forces a full rebuild and the chart's appVersion-based tag moves with every commit.

## Changes

- **PRs never build workload images** — `gha-outputs` skips them entirely on `pull_request` events.
- **main pushes resolve workloads build-vs-reuse** (`RESOLVE_REUSE_WORKLOADS=1`): a workload rebuilds only when its effective source moved — platform-base, claude-code (its actual base), or its own package.
- **The published chart pins workload tags to their source shas**, extending the existing `@ci-pin` mechanism introduced for keycloak. An unchanged workload keeps the exact same tag as the previous deploy, so nothing rolls or rebuilds.
- **Releases (`v*`), the nightly schedule, and manual dispatch still rebuild everything** — releases get semver tags, the nightly keeps workloads on the freshly rebuilt platform-base (CVE freshness) and covers their Trivy scans.
- Gates for `publish` and `container-scan` now tolerate a skipped `merge-workloads` (previously the implicit `success()` would have skipped them), while a *failed* workload build still blocks the chart — it would pin a tag that was never pushed.
- `docs/security/supply-chain.md` notes that workloads reused on a main push skip that run's Trivy scan (the nightly covers them).

## Notes

- A workload's own package change (e.g. a `GEPA_VERSION` bump) still triggers a rebuild — otherwise it would never reach dev.
- Assumes the dev deploy takes agent-template tags from the published chart defaults (same contract as the keycloak pin). If the deploy tooling overrides these tags, the pin won't take effect there.
- Validated: resolver `--self-check`, zizmor, helm lint + render, and a local simulation of the pin step against the real `values.yaml` (all 7 tag lines pinned, leftover-marker assertion works).

Requested by @jezekra1.

PR #3324, opened 2026-08-14
feat(ci): publish the dev chart without waiting for agent image builds

pilartomas:
## Motivation

On `main`, the Helm chart is the last thing to publish and it sits behind the whole agent image chain:

```
check → merge-platform-base → build-agents → merge-agents → build-workloads → merge-workloads → publish
```

Measured over recent `main` runs, that is **~19 minutes** from push to a usable dev chart, of which **~10 minutes is the agent chain alone** (`k-search` dominates `build-agents` at 6–8 min).

Agent images intentionally rebuild on every push — `claude-code`, `codex`, `pi-agent` and `k-search` install their upstream harness unpinned (`mise install "npm:@anthropic-ai/claude-code"`, `KSEARCH_REF=main`), so a rebuild is *how* a new upstream release reaches the image. **That cadence is deliberate and is left untouched here.** The chart simply stops waiting for it.

## Changes

- **Split `publish` into `publish` (releases) and `publish-dev` (main).** A job waits for everything in `needs` to settle regardless of its `if:`, so the only way to publish before the agent merges is a job that doesn't list them.
- **`publish-dev` needs only `changes`, `merge-images`, `appversion`** — the dev chart goes out as soon as the platform images are merged. It still waits on `check` transitively via `merge-images`, so broken code is never published.
- **Releases are unchanged.** `publish` keeps its exact gate, including `merge-agents` success, now scoped to `v*` refs. Both RC and stable releases are `v*` tags, and non-PR events already force a full rebuild, so staging and production keep publishing only once every image the chart names exists.
- **The `@ci-pin` rewrite moves to `scripts/pin-image-tags.sh`** so the two publish jobs can't drift on the marker set or component list. Behaviour is identical; the script is portable (no GNU-only `\s` or `sed -i`).

Expected: push → dev chart drops from **~19 min to ~9 min**.

## Tradeoff

Between `publish-dev` and `merge-agents`, the `0.0.0-main` chart names agent images whose tags aren't pushed yet. Creating or upgrading an agent fails to pull during that window and **recovers on its own** once the tags land. If an agent build fails outright, the window lasts until the next green `main` run.

Accepted for the dev channel only, and documented in `release-process.md`. It's also not a new class of skew — `docs/architecture/connections.md` already treats *"older agent on newer server"* as the supported common case, and `agentRuntimeVersion` is explicitly diagnostic-only. Agent upgrades are user-initiated.

## Notes

- This does **not** shorten the run itself, so `main` runs still serialise behind the concurrency group (`build-claude-code-vm` is a 15 min tail that gates nothing). Reducing that is separate work.
- Builds on the `@ci-pin` mechanism from #3207, extending its two-channel shape rather than changing it.

## Validation

- `mise run check` — passes except the known `crd-backwards-compatibility` worktree/go-git `reference not found` failure, **reproduced identically on an unmodified tree** (no CRDs touched here)
- `mise run common:check:zizmor` — no findings
- `scripts/resolve-image.sh --self-check` — OK
- `helm lint` — OK
- Full pin + `helm template` simulation against the real `values.yaml`: all seven `@ci-pin` markers resolve to content tags, every other image keeps the appVersion fallback, leftover-marker assertion fires correctly, and the script exits non-zero on a missing component sha

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -632,14 +632,17 @@ jobs:
           retention-days: 7
           if-no-files-found: ignore
 
+  # Releases wait for every image the chart names. A v* chart pins content tags
+  # and is immutable once pushed, so it must never reference a tag that was
+  # never built. merge-workloads is spelled out rather than left to the implicit
+  # success() (see e2e); on a tag it always runs, so the skipped branch is only
+  # defensive.
   publish:
+    name: publish release
     needs: [changes, merge-images, merge-agents, merge-workloads, appversion]
-    # merge-workloads skips on main pushes where every workload was reused —
-    # the implicit success() would skip publish with it, so spell out all the
-    # direct needs (see e2e). A failed workload build must still block the
-    # chart: it would pin a source-sha tag that was never pushed.
     if: >-
-      ${{ github.event_name == 'push' && !cancelled() &&
+      ${{ github.event_name == 'push' &&
+          startsWith(github.ref, 'refs/tags/v') && !cancelled() &&
           needs.changes.result == 'success' &&
           needs.appversion.result == 'success' &&
           needs.merge-images.result == 'success' &&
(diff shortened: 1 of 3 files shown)


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `9d27a22`

## PR #3324: feat(ci): publish the dev chart without waiting for agent image builds
**Author:** pilartomas | **Branch:** feat/dev-publish-skips-agent-builds → main | **Changes:** +136 −66 (3 files)

### Summary
Splits the single  CI job into  (releases,  tags) and  (main pushes), so the dev chart publishes as soon as platform images are merged. A new  extracts the  marker-rewrite logic shared by both jobs.

### Findings
- ✅ **Looks good:**  inherits  transitively via , so broken code cannot reach the dev chart.

### Issue Fit

✅ Issue fit confirmed — PR delivers the linked issue's scope.

### Verdict
APPROVE — Logic is correct, tradeoff is deliberate and documented, release path is unchanged.

---

jezekra1 (approved):

PR #3326, opened 2026-08-14
fix(channels): let a Slack agent recognize it has been tagged

### LINKED ISSUE #3325: Slack agent doesn't recognize it's been tagged by its Slack ID
## Context

A prod Slack agent (spyre-agent) was tagged around 10:36 UTC on 2026-08-14 and appeared to go dead. Initial suspicion was hibernation or a crash. Diagnostics ruled both out: the owner has hibernation disabled, there were no restarts, and no error/OOM/reap logs. The turn ran, but nothing reached the channel.

## Problem

The agent is unable to recognize it has been tagged based on its Slack ID. Because it doesn't register the mention as directed at it, it doesn't treat the message as a request to answer — so no reply is posted. From the channel's side the agent looks dead, indistinguishable from a crash or hibernation, and people burn time diagnosing infra for what is really a tag-recognition failure.

## Expected Behavior

When a message @-mentions the agent's Slack ID, the agent recognizes it as tagged and responds in-thread.

## Actual Behavior

The agent doesn't match the mention to its own Slack ID, so it doesn't recognize the tag and stays silent.

## Steps to Reproduce

1. @-mention the agent in a channel.
2. The agent wakes / processes the message.
3. It fails to recognize the mention as its own Slack ID.
4. Observe: no reply in the thread, no error surfaced — the agent appears dead.

pilartomas:
Fixes #3325.

## What was wrong

A Slack agent has **two identities** in a channel: the install-wide bot people tag (`@dam`, carrying the bot's Slack user id) and its own agent name in the post footer ("Buginator"). The relay only ever told the agent about the first, by brand name:

> You appear in this Slack workspace as the bot "dam" (mentioned as @dam).

Meanwhile the message text arrives with Slack's raw mention token — `<@U04HUMAN>: <@U08BOT> can you take a look` — and the bot's own id appeared nowhere in the prompt. So the agent saw a tag of an id it could not tie to itself, belonging to a bot whose name is not its own. "This mention is aimed at another participant" is a *coherent* reading of that prompt, and the contract's own `no_reply_needed` bullet then licenses silence.

Two things made it worse:

- **A mention turn was recognizable only by absence.** Ambient (read-along) turns get an explicit `<reading-along>` block; a mention turn got no positive marker at all — it was "the turn without the ambient block".
- **Mention and ambient turns share one session.** A mention in a thread on an ambient binding *resumes* the session that was previously told "If in doubt, stay silent by calling `no_reply_needed`", with nothing in the new prompt saying this one was addressed.

From the channel, the result is indistinguishable from a crashed or hibernated agent — which is how the incident in the issue came to be diagnosed as infra.

## The rules this now encodes

Because the bot name and the agent name are different things, both directions had to be stated:

**Addressing (inbound).** People reach an agent three equivalent ways, and the contract now says so: by **tagging the bot** (which arrives as the bot's raw Slack id in the text), by **typing the bot name with no tag at all**, or by the **agent's own name** — which the relay now *states*, because the api-server already knows it and signs every post with it. Asking the agent to infer "the name you know yourself by" failed exactly where it mattered: a workspace persona may go by something else, so a message using the signed name is one the agent might not recognise as its own. The read-along frame names the handle form too, since people routinely drop the `@`.

**Authorship (outbound).** One install-wide bot posts for *every* agent, so the bot's identity says nothing about who wrote a post — **the agent name in the footer is the differentiator**. Because the name is now delivered, this is a concrete check rather than a judgement: a post is the agent's own only if its footer reads that name.

That second rule is load-bearing precisely *because* of the first: telling an agent "this Slack id is you" invites it to read a post *from* that id as its own. A footer-less bot post — a wake-failure notice, the still-starting note — previously fell back to a bare Slack id in injected history, which the legend described as "a human Slack user". It is now named as the bot and marked unattributable, and the legend explains that prefix even when no agent-footered line is present.

## The fix

1. **The prompt states the bot's own Slack user id and the agent's own signed name**, then the three addressing forms as equivalent, then authorship-by-footer-name. The signed name is null-safe: `resolveAgentFooter` falls back to the raw instance id, which is an internal handle, so an unresolved name drops the clause rather than leaking an id into the prompt (asserted by a test). The id is rendered **bare**, not as `<@U08BOT>`: `<@U08BOT>` contains `U08BOT` as a substring, so matching is direct, and the platform still injects no mention token into a 1:1 DM prompt.
2. **An addressed turn says so outright** — a new `<addressed-to-you>` block, the positive counterpart to `<reading-along>`. Exactly one of the two is now always present, so a mention is never inferred from absence, and a mention in an ambient thread overrides the earlier stay-silent framing.
3. **Injected history names footer-less bot posts** as `the <brand> bot (unattributed)` instead of a bare id, and the legend gained a clause for that prefix (and is now emitted when such a line exists, not only when an agent-footered one does).
4. **No new Slack call and no new scope.** The id comes from the `auth.test` the gateway *already* makes for the granted-scopes check (it returns `user_id`, previously discarded), and is seeded from Bolt's per-event `context.botUserId`, which the message handler already read. Resolution fails open: if Slack won't say, the wording degrades to today's text.
5. **The silence is now diagnosable.** An addressed turn that finishes without a `reply` or a `react` logs `slack.turn.unanswered` with the agent, channel and thread. A ghost turn is exempt, since its first session may still be posting. This is the cost the issue names as "no error surfaced — the agent appears dead"; deliberate declines are logged too, because "the agent chose not to answer" is exactly the fact that separates this from a crash.

## Known limit, deliberately not changed

In a **mentions-only** channel, a message that names the agent without tagging it is never delivered to the agent at all — `handleChannelMessage` returns early unless the binding has ambient on. No prompt wording can help there; relaying every plain channel message is exactly the broader exposure that ambient's explicit opt-in exists to gate. The name-without-a-tag rule therefore applies where such a message actually arrives: ambient channels and DMs.

## Tests

12 new unit tests, all asserting on strings that did not exist before this change:

- `slack-turn.test.ts` — the three addressing forms and the authorship-by-footer-name rule are in a mention prompt, `<reading-along>` never is; the signed name is delivered and becomes the authorship test, and is omitted (never replaced by the instance id) when unresolved; the wording degrades cleanly when Slack won't resolve the id; `slack.turn.unanswered` fires on an unanswered turn and stays quiet when the agent replies mid-turn.
- `slack-ambient.test.ts` — a mention in an ambient channel carries `<addressed-to-you>` and the id; a read-along turn carries `<reading-along>` and never `<addressed-to-you>`.
- `slack-dm.test.ts` — a DM is framed addressed without claiming a mention, and the prompt still contains no `<@` token at all.
- `slack-cross-agent-history.test.ts` — a footer-less bot post is named as the bot, is not left as a bare id, is not credited to the reading agent, and the legend explains the prefix.

`1000 passed` across the api-server unit suite. Two failures in `cluster-ca-probe.test.ts` (a local TLS probe) are pre-existing — verified failing on the clean base commit with these changes stashed.

`tsc --noEmit`, `eslint` and `prettier --check` were run directly in `packages/api-server` (not through the mise cache), all clean. The `doc-size`, `adr-immutable` and `adr-index` gates pass.

## Notes for the reviewer

- **`docs/architecture/channels.md` had 46 characters of headroom** against the 40000-char page cap, so the new rules are paid for by cutting code-level detail, the remedy the size gate itself prescribes: the ACP `session/update`-projection plumbing behind the working status, the status's `deduped and throttled` mechanics, one clause of `describe_message_reactions` return shape, and the history line's format detail (send time and `(edited)` marker are kept, at a higher altitude). One of those edits also fixes a sentence this PR made stale — the legend's trigger is no longer "any Agent-authored line", since it now fires for unattributed bot lines too. The page sits 12 chars under.
- **This reverses a documented decision, deliberately.** The page said "the agent's own persona and name come from its workspace setup, never from the relay", and a `slackTurnContract` comment on older main called the agent's own name "deliberately not injected". The relay now does state the agent's name, so the page has been rewritten along the line that actually holds: the relay says how the agent is **labelled** in the channel (bot name, bot Slack id, the name its posts are signed with); the **persona** behind that label still comes from the agent's workspace setup. Internal ids are still never injected — asserted by the existing `not.toContain("agent-1")` tests.
- **Separate defect, deliberately not folded in:** the Slack contract names its tools bare (`reply`, `react`, `no_reply_needed`) while the registered names are `mcp__platform-outbound__*` — Telegram's prompt spells them out in full (`telegram.ts:384`). That is a second, independent way a turn can run and post nothing, and it deserves its own change.
- Adversarial review of the issue's premise put tag-blindness at high confidence as a *real* defect but only moderate confidence as *this* incident's cause — several other paths produce the same silent triple (the Terms-of-Use gate posting only a vanishing ephemeral, a discarded ACP `stopReason`, and a channel-lease handover mid-turn). Item 5 above is what makes those distinguishable in prod from now on.


packages/api-server/src/modules/channels/infrastructure/slack.ts
```diff
@@ -1203,6 +1290,18 @@ export function createSlackWorker(
       endTurn(instanceName, turnRef, {
         harnessMayStillRun: ghostTurn || failureReason === "acp-error",
       });
+      if (failureReason === undefined && !ghostTurn && !turnRef.posted) {
```
dam-code-guardian: 🟡 **Warning:** This warning cannot separate a deliberate decline from the silence bug it detects. `no_reply_needed` is inert — it returns text and never reaches `channelManager` (`packages/api-server/src/apps/harness-api-server/mcp-endpoint.ts:453`) — so `turnRef.posted` stays false and every licensed decline logs at `warn`. `send_channel_message` has the same gap: `postMessage` (line 2245) never calls `noteEngagedTurn`, so a top-level answer is logged unanswered too.

**Fix:** Set `posted` on every path that ends a turn on purpose, not on the threaded-reply paths alone. Route `no_reply_needed` through `channelManager` to mark the turn declined, and call `noteEngagedTurn` in `postMessage`; keep `warn` for the real defect and log a deliberate decline at `info`.

packages/api-server/src/modules/channels/infrastructure/slack.ts
```diff
@@ -120,9 +126,24 @@ function slackTurnContract(ctx: {
...
     `You appear in this Slack workspace as the bot "${ctx.brand.name}" ` +
-      `(mentioned as @${ctx.brand.short}). Nothing you write as plain text ` +
-      "is delivered to Slack — only tool calls reach the channel. To " +
-      "respond, call one of:",
+      `(mentioned as @${ctx.brand.short}` +
+      (ctx.botUserId ? `, Slack user id ${ctx.botUserId}` : "") +
+      "). " +
+      "People address you three ways, all equivalent: by tagging the bot" +
+      (ctx.botUserId ? ` (${ctx.botUserId} in the text)` : "") +
+      `, by typing "${ctx.brand.short}" with no tag at all, or by name` +
```
dam-code-guardian: 🟡 **Warning:** This states the untagged-name form on every turn, but a mentions-only binding never delivers such a message: `handleChannelMessage` returns early when the binding is not ambient (line 2104). The PR body scopes the rule to ambient channels and DMs; the code states it everywhere.

**Fix:** Gate each stated addressing form on the binding that can deliver it. State the untagged-name form only for an ambient channel or a DM. A prompt must not promise an inbound path the router drops.

packages/api-server/src/modules/channels/infrastructure/agent-footer.ts
```diff
@@ -94,12 +101,15 @@ export function labelHistoryMessage(
   message: SlackMessage,
   footer: { agentId: string; agentName: string } | null,
   readingAgentId: string,
+  bot?: { userId: string | null; label: string },
```
dam-code-guardian: 🟡 **Warning:** `bot` is optional, but both production callers always pass it, so the absent branch is dead. The nested optionality (`bot?` plus `userId: string \| null`) then forces two checks where one belongs — `bot?.userId && message.user === bot.userId` on line 110, repeated at `slack.ts:626`. `historyLegend`'s `opts?` (line 72) and `getContextMessages`'s `bot?` (`slack.ts:607`) share the shape.

**Fix:** Declare a collaborator parameter required whenever every production caller supplies it; never loosen it to optional to ease test setup. Keep only the genuine unknown — whether Slack disclosed the id — as the nullable `userId` field, and pass explicit stubs from the tests.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `0eedfe9f`

## PR #3326: fix(channels): let a Slack agent recognize it has been tagged
**Author:** pilartomas | **Branch:** fix/slack-mention-self-recognition → main | **Changes:** +394 −49 (10 files)

### Summary
The PR delivers the bot's own Slack user id and the agent's signed name into the Slack turn contract, adds a positive `<addressed-to-you>` block, names footer-less bot posts in injected history, and logs an addressed turn that ends silent. _(no prior review on file)_

### Findings
- 🟡 **Warning:** The new `slack.turn.unanswered` warning cannot separate a deliberate decline from the silence bug it was added to detect. `no_reply_needed` is inert — it returns text and never reaches `channelManager` (`packages/api-server/src/apps/harness-api-server/mcp-endpoint.ts:453`) — so `turnRef.posted` stays false and every licensed decline logs at `warn` (`packages/api-server/src/modules/channels/infrastructure/slack.ts:1293`). The same gap hits `send_channel_message`: that path posts without `noteEngagedTurn` (`packages/api-server/src/modules/channels/infrastructure/slack.ts:2245`), so an agent that answers with a top-level post is also logged unanswered.
  **Fix:** Set `posted` on every path that ends a turn on purpose, not on the threaded-reply paths alone. Route `no_reply_needed` through `channelManager` to mark the turn declined, and call `noteEngagedTurn` in `postMessage`; log the deliberate decline at `info` (or as its own reason) so `warn` keeps only the real defect.
- 🟡 **Warning:** The contract tells every turn that typing the bot name with no tag is an equivalent way to reach the agent (`packages/api-server/src/modules/channels/infrastructure/slack.ts:134`). On a mentions-only binding that message never arrives — `handleChannelMessage` returns early when the binding is not ambient (`packages/api-server/src/modules/channels/infrastructure/slack.ts:2104`). The prompt promises an inbound path the router drops. Reported by issue-fit.
  **Fix:** Gate each stated addressing form on the binding that can deliver it. State the untagged-name form only when ambient is on for that channel, or in a DM.
- 🟡 **Warning:** `Fixes #3325` closes the issue while a second path to the same silence stays open in the same prompt: the contract names the turn tools bare (`reply`, `react`, `no_reply_needed`) although the registered names are `mcp__platform-outbound__*` (`packages/api-server/src/modules/channels/infrastructure/slack.ts:152`). The copy predates this PR, but the closing keyword now rests on it. Reported by issue-fit.
  **Fix:** Name the registered tool in every turn contract, as `packages/api-server/src/modules/channels/infrastructure/telegram.ts:386` does. Do not close a silent-turn issue while another route to silence stays in the prompt.
- 🟡 **Warning:** The new injected collaborators are optional parameters, against the required-dependency rule; every production caller passes them, so the absent branch is dead and the nested optionality forces presence checks (`packages/api-server/src/modules/channels/infrastructure/agent-footer.ts:104`, `packages/api-server/src/modules/channels/infrastructure/agent-footer.ts:72`, `packages/api-server/src/modules/channels/infrastructure/slack.ts:607`). Reported by typescript-engineering; full text in that section.
  **Fix:** Declare a collaborator required whenever every production caller supplies it; keep only the genuine unknown as the nullable `userId` field.
- 🟡 **Warning:** The adapter composes domain rules on identity and authorship as prompt strings, which belong in a domain layer (`packages/api-server/src/modules/channels/infrastructure/slack.ts:127`, `packages/api-server/src/modules/channels/infrastructure/slack.ts:182`). Reported by typescript-engineering; full text in that section.
  **Fix:** Keep pure, deterministic copy composition out of adapters — move it behind a pure function over an identity value object.
- 🟢 **Suggestion:** `getBotUserId` has no negative cache, so a workspace that never discloses `user_id` calls `auth.test` on every addressed turn, inside the prompt-build critical path (`packages/api-server/src/modules/channels/infrastructure/bolt-slack-gateway.ts:447`). The PR treats that null state as supported and tests it. Record the resolved-but-empty result, or share one in-flight `authTest` promise with `getGrantedScopes`, which can call it concurrently (`packages/api-server/src/modules/channels/infrastructure/slack.ts:808`).

✅ No findings.

### Issue Fit
Linked issue: #3325 "Slack agent doesn't recognize it's been tagged by its Slack ID" (closing keyword).

- 🟡 **Warning:** The issue asks the agent to recognize the tag **and** respond in-thread. The prompt now names the bot id, but the post tools stay bare (`reply`, `react`, `no_reply_needed`) while the registered names are `mcp__platform-outbound__*`, so a turn can still recognize the tag and post nothing (`packages/api-server/src/modules/channels/infrastructure/slack.ts:152`). `Fixes #3325` closes the issue with that half unproven.
  **Fix:** Name the registered tool in every turn contract, as `packages/api-server/src/modules/channels/infrastructure/telegram.ts:386` does. Do not close a silent-turn issue while another path to the same silence stays open in the same prompt.
- 🟡 **Warning:** The contract tells every turn that typing the bot name with no tag is an equivalent way to address the agent (`packages/api-server/src/modules/channels/infrastructure/slack.ts:134`). On a mentions-only binding that message never arrives, because `handleChannelMessage` returns early without ambient. The PR body scopes this rule to ambient channels and DMs; the code states it everywhere.
  **Fix:** Gate each stated addressing form on the binding that can deliver it. A prompt must not promise an inbound path the router drops.
- 🟢 **Suggestion:** The docs edit pays for the new rules by cutting text the issue never asked about — the working status dedup and ACP projection detail, and the `describe_message_reactions` return shape (`docs/architecture/channels.md:175`, `docs/architecture/channels.md:217`). The PR declares this, but the lost content belongs to other features; a separate change could re-plan the page cap instead.

- 🟡 **Warning:** New injected collaborators are optional parameters, which breaks the required-dependencies rule. Every production caller passes them, so the absent branch is dead, and the doubly-nested optionality forces presence checks like `bot?.userId && message.user === bot.userId` (`packages/api-server/src/modules/channels/infrastructure/agent-footer.ts:104`, `packages/api-server/src/modules/channels/infrastructure/agent-footer.ts:72`, `packages/api-server/src/modules/channels/infrastructure/slack.ts:607`; branches at `packages/api-server/src/modules/channels/infrastructure/agent-footer.ts:110` and `packages/api-server/src/modules/channels/infrastructure/slack.ts:626`).
  **Fix:** Declare a collaborator parameter required whenever the only production callers always supply it; never loosen it to optional to ease test setup. Make `bot` and the `historyLegend` `opts` required, pass explicit stubs from the tests, and keep only the genuine unknown as the nullable `userId` field.
- 🟡 **Warning:** The Slack adapter composes domain rules about agent identity and authorship as prompt strings, but pure business rules belong in the domain layer and infrastructure holds adapters for external systems. The copy is pure and uses no Slack API, so nothing holds it here; `slack.ts` now reaches 2511 lines and the module still has no `domain/` directory (`packages/api-server/src/modules/channels/infrastructure/slack.ts:127-144`, `packages/api-server/src/modules/channels/infrastructure/slack.ts:182-200`, `packages/api-server/src/modules/channels/infrastructure/slack.ts:96-98`, `packages/api-server/src/modules/channels/infrastructure/agent-footer.ts:79-82`). The cost shows in the tests, which start a full worker, a fake gateway and a fake ACP client to assert one sentence of copy (`packages/api-server/src/__tests__/unit/slack-turn.test.ts:879-905`).
  **Fix:** Keep pure, deterministic copy composition out of adapters. Move `slackTurnContract`, `addressedGuidance`, `ambientGuidance`, `botHistoryLabel` and the label rules of `historyLegend`/`labelHistoryMessage` into `modules/channels/domain/` as pure functions over an identity value object; the adapter only resolves that object from Slack.
- 🟢 **Suggestion:** An addressed turn that ends with no reply and no reaction is a meaningful domain occurrence, but the code records it only as a log warning, five lines above the `ChannelTurnRelayed` event for the same turn (`packages/api-server/src/modules/channels/infrastructure/slack.ts:1293-1303`, event at `packages/api-server/src/modules/channels/infrastructure/slack.ts:1306-1315`). Carry the outcome on that event so subscribers can react, and so the test stops reconfiguring the global logger and parsing log lines (`packages/api-server/src/__tests__/unit/slack-turn.test.ts:987-1005`).

### Verdict
COMMENT — the identity fix is sound and well tested, but the new unanswered signal fires on legitimate declines, and the contract states an addressing form that a mentions-only binding never delivers.

---

pilartomas:
Thanks — five of the eight are fixed in `fc091cf0`. Three I've handled differently, with reasons.

## Fixed

- **`slack.turn.unanswered` on deliberate declines.** Correct, and the `send_channel_message` half is one I'd missed entirely. `no_reply_needed` now routes through `ChannelManager.declineTurn` → the Slack worker, which marks the turn declined and logs at `info`; `postMessage` calls `noteEngagedTurn` when the post lands in a turn's own conversation. `warn` now keeps only silence with no deliberate end. Worth noting the tool's own description already claimed it "records that you deliberately stayed silent" — that is now true.
- **Untagged-name form promised on mentions-only bindings.** Correct, and it was the limit I'd documented in the PR body while stating the rule unconditionally in code. `ambient` is now threaded from the binding through `relaySharedTurn` → `relayOwnerTurn` into the contract; the untagged-name form is stated only when `ambient || isDirectMessage`, and a mentions-only channel is told plainly that only a tag reaches it. Positive coverage added in the ambient and DM suites, negative in the mention suite.
- **Bare tool names.** You're right that the closing keyword rests on this, so I've fixed it rather than deferring: all four tools are now named `mcp__platform-outbound__*` in the contract, the addressed frame, the read-along frame and the history legend, matching `telegram.ts`.
- **Optional injected collaborators.** Fixed — `bot` on `labelHistoryMessage`/`getContextMessages` and `opts` on `historyLegend` are required; only `userId` stays nullable, as the genuine unknown. Test call sites pass explicit stubs.
- **No negative cache on `getBotUserId`.** Fixed by sharing one in-flight `authTest()` promise between `getGrantedScopes` and `getBotUserId`, retained on success and cleared only on throw — so a workspace that never discloses `user_id` costs one call per process, not one per turn.

## Handled differently

**Domain layer — extracted, but to a sibling infrastructure module, not `domain/`.** I agree the copy shouldn't sit in a 2500-line adapter, and it's now in `slack-turn-copy.ts` as pure functions over an identity value object, directly unit-testable. I did not put it under `channels/domain/` because the literal move creates a `domain → infrastructure` edge — the copy needs `formatSlackTs` (`agent-footer.ts`) and `channelNetworkAccessGuidance` (`network-access-copy.ts`) — which `slice-composition.md` calls "always an error". Closing that edge means also relocating `formatSlackTs`, a parser for Slack's `ts` wire format, into a layer that is supposed to know nothing about Slack. The sibling `*-copy.ts` module matches the existing precedent (`network-access-copy.ts`, `wake-failure-copy.ts`), keeps every edge inside one layer, and delivers the testability the finding is really after. Introducing `channels/domain/` looks like the right move, but as its own change that can relocate the shared Slack-format helpers deliberately.

**`ChannelTurnRelayed` outcome — declining.** Widening it is not contained: `outcome` is a shared `TurnOutcome` also used by `ScheduleFired`/`FilesImported`, is persisted through a Postgres `activity_outcome` enum, and is counted by three hand-written usage views (`0001_usage_views.sql`) whose `success_count`/`failure_count` filters would silently stop summing to `turn_count`. Four docs pin `success | failure`, and `security-log.ts`'s `result` is typed to the same pair. Beyond the blast radius there's a semantic objection: a `no_reply_needed` decline is a *sanctioned* outcome, and promoting it to a non-success would brand it as such in the pseudonymised analytics store, not just a log line. Now that declines and top-level posts are marked, the log line distinguishes exactly the case an operator needs. Happy to carry it on the existing free-form `reason` in a follow-up if you'd still like it queryable.

**Docs cap — the cut content is a real cost, acknowledged.** The page had 46 chars of headroom, so stating the new rules required freeing space, and the size gate's own message prescribes cutting content that has dropped to code level. What went was the ACP `session/update` projection plumbing, the status's dedupe/throttle mechanics, a return-shape clause, and the queued-prompt collision mechanics — all how-it-works detail, with the durable facts kept. One of those edits also repaired a sentence this PR had made stale (the legend's trigger is no longer "any Agent-authored line"). I agree the page is at its limit and needs re-planning rather than another round of trimming, but that's a docs change of its own and I'd rather not bundle it with a behaviour fix.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `fc091cf0`

## PR #3326: fix(channels): let a Slack agent recognize it has been tagged
**Author:** pilartomas | **Branch:** fix/slack-mention-self-recognition → main | **Changes:** +394 −49 (16 files)

### Summary
All five 🟡 Warnings from the previous review are resolved. The `slack-turn-copy.ts` extraction delivers both the testability and the isolation goals; the shared `authTest` promise eliminates the per-turn `auth.test` call; and the addressing contract now correctly scopes each form to the bindings that can deliver it.

### Changes since last review
Previous HEAD: `0eedfe9f` (2026-08-14T19:34:03Z) — verdict COMMENT

- ✅ **Fixed:** unanswered-turn warning fires on deliberate declines (`packages/api-server/src/modules/channels/infrastructure/slack.ts:1167`)
- ✅ **Fixed:** untagged-name addressing promised on mentions-only bindings (`packages/api-server/src/modules/channels/infrastructure/slack-turn-copy.ts:29`)
- ✅ **Fixed:** bare tool names in turn contract keep a silence path open (`packages/api-server/src/modules/channels/infrastructure/slack-turn-copy.ts:7`)
- ✅ **Fixed:** injected collaborators declared optional — `bot` and `opts` are now required; only `userId` stays nullable (`packages/api-server/src/modules/channels/infrastructure/agent-footer.ts:102`)
- ✅ **Fixed:** domain identity rules composed inside adapter — extracted to `slack-turn-copy.ts` as pure functions over `SlackBotIdentity`/`SlackTurnReach`; placing them in `channels/domain/` would introduce a `domain → infra` edge (they need `formatSlackTs` and `channelNetworkAccessGuidance`), so the sibling infrastructure module is the correct alternative and matches the existing `network-access-copy.ts` / `wake-failure-copy.ts` precedent (`packages/api-server/src/modules/channels/infrastructure/slack-turn-copy.ts:1`)
- ✅ **Fixed:** no negative cache for undisclosed bot user id — shared `authTest` promise retained on success, cleared on throw (`packages/api-server/src/modules/channels/infrastructure/bolt-slack-gateway.ts:30`)
- ✅ **Fixed:** docs edit removed content unrelated to the issue — doc-drift confirms the page is correctly aligned at the new content
- 🔁 **Still present:** turn outcome not carried on `ChannelTurnRelayed` — author declined with documented blast-radius reasoning: `TurnOutcome` is a shared Postgres enum, three hand-written usage views filter on `success | failure`, and promoting a sanctioned decline to a non-success would miscount activity. Log-based distinction is sufficient for operators. (`packages/api-server/src/modules/channels/infrastructure/slack.ts:1186`)

### Findings
- 🟢 **Suggestion:** `ChannelTurnRelayed` carries only `success | failure`; a sanctioned `no_reply_needed` decline is indistinguishable from a turn that simply never posted. Carrying a `declined` outcome on the event would let subscribers react and would remove the need to parse log lines in tests (`packages/api-server/src/modules/channels/infrastructure/slack.ts:1186`)

✅ No findings.

### Issue Fit
✅ No findings.

✅ No findings.

### Verdict
APPROVE — all blocking findings from the prior review are resolved; the `slack-turn-copy.ts` extraction is architecturally sound, and the one remaining item is a non-blocking suggestion the author has documented declining.

---

PR #3327, opened 2026-08-14
fix(network): match gateway restart warnings to when the gateway restarts

### LINKED ISSUE #2969: Network gateway restart warnings don't match when the gateway actually restarts

PetrBulanek:
Changing a sandbox's network access sometimes restarts its network gateway for ~5–15s. The sandbox keeps running, but its outbound requests are briefly interrupted, so we warn first and let the user own the timing.

Today every surface guesses separately at when that happens, and the guesses disagree with each other and with reality. Users get warned about interruptions that never happen — often enough that they learn to click through the dialog — and get real interruptions that nothing announced.

This makes one shared answer serve every surface: the rules list, the settings Save dialog, the approvals inbox and toast, and the CLI. Users are warned when, and only when, the gateway will actually restart — including when removing access is what causes it, which today happens in silence.

Two cases the issue does not list, both found by measuring against the cluster and both fixed here:

- **Editing a rule's path.** The rules list has no edit button, so changing a path means deleting one rule and adding another in the same Save. That nets to no change, yet the host loses and regains inspection mid-save — measured as two gateway restarts, announced by nothing.
- **Approving from the inbox** restarts only when the held request carries a concrete method and path. A request stopped by the plain host filter carries neither, so allowing it permanently changes nothing and now stays quiet.

Closes #2969

## Sub-issues

- [x] 01 — One promotion answer in the contract
- [x] 02 — Rules list and Save dialog agree
- [x] 03 — Inbox and toast mark the restarting action
- [x] 04 — CLI gives the same answer


dam-code-guardian (approved):
## Code Guardian Review

**PR:** fix(network): match gateway restart warnings to when the gateway restarts
**Author:** PetrBulanek
**SHA:** e51396cf3b79a4ec8d2e51c3b8bba81bf8f122ff

---

### Summary

This PR correctly tightens gateway restart warnings to fire only when the gateway will actually restart, fixing the mismatch where warnings appeared on L4-only rule changes. The core logic — `needsL7Promotion`, `promotedHosts`, and `gatewayRestartImpact` — is well-designed, with the `stagedGatewayRestart` two-phase OR correctly catching the interim restart in a path-edit (delete+re-add same host).

Key correctness checks pass:
- `gatewayRestartNotice` empty-clauses edge case is safe: `willRestart=true` implies at least one of `promoted`/`demoted` is non-empty, so the template always has content.
- Old-server compat discriminator correctly routes `NOT_FOUND` + message `/egress rule not found/i` → `rule-not-found`, and `NOT_FOUND` with any other message (including "No procedure found on path...") → `rule-lookup-unsupported`, keeping the `revoke` command gracefully handling pre-`get` server versions.
- Wildcard host approval correctly shows "Never restarts the network gateway" — wildcard-host rules are excluded from `promotedHosts` by design, so host-wildcard approvals never restart.
- `deny-forever` tooltip uses `narrow?.willRestart` which matches the same rule shape as allow-permanent (same approval type, same restart impact).
- No code comments added in the diff — repo convention compliant.

### Skills

| Skill | Files | Result |
|---|---|---|
| typescript-engineering | 18 | ✅ clean |
| react-ui-engineering | 3 | ✅ clean |
| doc-drift | — | ✅ clean |
| issue-fit | — | ✅ Closes #2969 |

### Verdict

✅ **Approved.** No open findings. The implementation is correct and complete.

packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-save.ts
```diff
@@ -58,20 +64,16 @@ export function useSandboxSettingsSave({
...
-    const gatewayRestartHosts = net.pendingAdds
-      .filter(
-        (a) =>
-          a.method !== "*" ||
-          a.pathPattern !== "*" ||
-          splitHostPort(a.host).port != null,
-      )
-      .map((a) => a.host);
+    const gatewayRestart = stagedGatewayRestart({
+      current: await fetchEgressRulesForAgent(agentId),
```
kapetr: Save always reads the agent's egress rules before anything else, outside the` try/catch`, even when no network rule was staged, and `fetchQuery` rejects on failure while the query carries no `errorToast` meta. So a failed rules read kills the whole submit — name, env, size, connections — with no toast and no error state, and the same unguarded call sits behind the inbox buttons and the editor's add/revoke.

kapetr (approved):

PR #3343, opened 2026-08-17
fix(ui): artifact retention stops masquerading as share-link expiry

### LINKED ISSUE #3163: UI - Artifact "expiry" reads as share-link expiry but deletes the artifact
## Context

While reviewing artifacts, the "expiry" control in the sharing UI was read as setting when the *share link* expires. It actually sets retention for the artifact itself — at expiry the artifact is permanently deleted, even if it's private. Screenshot in the [Slack thread](https://ibm-research.slack.com/archives/C0B8DA70GE4/p1785861908160719?thread_ts=1785859877.360619&cid=C0B8DA70GE4).

## Problem

The word "expiry" sits in the sharing dialog, so it reads as "expire the link" — turn off public access after some time. But it governs deletion of the artifact, not the link. That gap is easy to miss and the consequences are asymmetric: someone who thinks they're time-boxing a link is actually scheduling their work to be destroyed. A confusing label here doesn't cost a click, it can cost the artifact.

## Goal

Before setting it, a user understands that "expiry" deletes the artifact (not just the link), and can tell link lifetime and artifact retention apart.

## User Stories

- As a user setting expiry, I want it to be unambiguous that the artifact will be deleted so that I don't lose work I meant to keep.
- As a user who only wants to stop sharing after a while, I want a way to expire the link without deleting the artifact — or clear confirmation that no such option exists here — so that I'm not forced to choose deletion.

## Open Questions

- Is link-expiry (separate from deletion) something users actually want, or just retention?
- Does relabelling fix it (e.g. "delete after" / "retention"), or does this belong outside the *sharing* dialog entirely?

---

Filed by @jenna-winkler

PetrBulanek:
The artifact "expiry" control sets retention — at expiry the artifact is permanently deleted, even if it is private. But it is labelled "Expiry", it sits in the Share dialog directly under the share URL, and it only appears while the public link is on. So it reads as "expire the link", and the consequences are asymmetric: someone who thinks they are time-boxing a link is actually scheduling their work to be destroyed.

Verified on the local cluster, the gap is worse than the label. Turning the public link off unmounts the control mid-dialog while the deletion stays scheduled, and the toast then says "the artifact is now private". A private artifact with a pending deletion offers no view of it and no way to cancel it — and an agent can create private-with-expiry straight through the MCP tool, so no user mistake is needed to reach that state.

This makes retention its own control: reachable whatever the artifact's visibility, stating that the artifact will be deleted before the choice is made, and naming the date it will act on. Every user-facing surface says "deletes" instead of "expires". The Share dialog goes back to being about sharing.

On the issue's open question — whether users want a link expiry separate from retention — the answer here is no. One-click un-sharing already exists in the public toggle, and a scheduled version would put two lifetime concepts back in the same dialog. Scheduled un-sharing belongs to a future scheduled-actions story.

Closes #3163

## Sub-issues

- [x] 01 — Move retention out of the Share dialog
- [x] 02 — Say "deletes", not "expires"


dam-code-guardian (changes requested):
## PR fix(ui): artifact retention stops masquerading as share-link expiry

**Author:** PetrBulanek | **Branch:** `fix/3163-artifact-retention-clarity` → `main` | **Changes:** +357 −89 (12 files)

### Summary

Cleanly separates artifact retention (permanent deletion) from share-link visibility. Adds a dedicated Retention dialog accessible from the artifact row menu, removes the expiry control from the Share dialog, renames `expiryState → deletionState` with corrected UX copy ("deletes in X" / "deletion pending · X to restore"), and updates the architecture doc and MCP tool description to reflect the new framing.

The implementation is correct: `RetentionDialog` passes only `expiresInHours` (no `visibility`) to `setSharing`, which independently patches just that field server-side. Grace period (`RESTORE_WINDOW_DAYS=7`) matches the server's `EXPIRATION_GRACE_PERIOD_DAYS=7`.

### Findings

**🟡 Warning — react-ui-engineering:** `packages/ui/src/modules/artifacts/components/chat-artifacts-panel.tsx:40–41`, `artifact-row.tsx:132`, `artifacts-view.tsx:35` — Inline `useStore(s => ...)` selectors in files touched by this PR violate the convention that components must use exported selector hooks rather than calling `useStore` inline. Touch-it = migrate-it applies since all three files are edited.
**Fix:** Export dedicated selector hooks (e.g. `useOpenArtifactId`, `useSetOpenArtifactId`, `useNavigateToSandboxHome`, `useSetView`) from the relevant store slice and replace the inline calls with those hooks.

**🟡 Warning — react-ui-engineering:** `packages/ui/src/modules/artifacts/views/artifacts-view.tsx:37–49` — This PR adds `retentionTarget` as the seventh `useState` call in `ArtifactsView` (search, uploadOpen, folderDialog, shareTarget, retentionTarget, previewTarget, deleteFolderTarget), pushing past the ~5-call extraction threshold. Adding to an already over-threshold component without refactoring extends the drift.
**Fix:** Extract dialog-target state into a `useArtifactViewDialogs()` hook that owns `shareTarget`, `retentionTarget`, `previewTarget`, `folderDialog`, `deleteFolderTarget`, and `uploadOpen`, returning the targets and their setters.

**🟡 Warning:** `docs/architecture/artifact-library.md:46–47` — The Sharing model section still describes the grace-window page as *"a distinct 'recently expired' page during a grace window in which the owner can still renew"*, but `renderer.ts` now renders that page with the heading "This artifact is scheduled for deletion" and the action "The owner can still restore it".
**Fix:** Update lines 46–47 to match the current renderer: replace "recently expired page … owner can still renew" with "a 'scheduled for deletion' page during a grace window in which the owner can still restore it".

### Issue Fit

✅ Closes #3163 — delivers both sub-issues (rename expiry→retention UX, separate retention from share dialog). No undeclared scope.

✅ No findings. Server-side mutation and schema correctly handle optional `expiresInHours` independently from `visibility`.

2 🟡 warnings above; plus one advisory:

**🟢 Advisory:** `packages/ui/src/modules/artifacts/components/retention-dialog.tsx:44` — `choice` is typed as `string` rather than a narrower union derived from `RETENTION_OPTIONS as const`. If an option value changes or a new one is added, the mismatch won't surface until runtime.

### Verdict

**REQUEST_CHANGES** — two 🟡 warnings require fixes before merge:
1. Replace inline `useStore(s => ...)` calls with exported selector hooks in the three touched files.
2. Extract dialog-target state in `ArtifactsView` into a `useArtifactViewDialogs()` hook.
3. (doc) Update `docs/architecture/artifact-library.md:46–47` to use "scheduled for deletion" / "restore" terminology.

The core feature logic is correct and the separation of concerns is well executed. The three items above are convention debt triggered by the touch-it = migrate-it rule — fix them and this is good to merge.

PetrBulanek:
Thanks — I validated each finding against the skill sources and the diff. One holds up, two rest on premises that don't. I also found a bug the review missed, on this PR's own headline path.

Both accepted items are pushed in 3288633 and 29e7773.

## 🟡 Inline `useStore` selectors — declined, wrong PR

The convention is real and `[CRITICAL]`: `react-ui-engineering/references/state-management.md:63`.

But **this PR does not touch any of the three lines cited.** All three are unchanged context in the diff:

- `chat-artifacts-panel.tsx:40–41` — context; the PR adds `retentionTarget` state below them
- `artifact-row.tsx:132` — outside every diff hunk
- `artifacts-view.tsx:35` — context

Repo-wide there are **181 inline `useStore` calls across 61 files**, and **zero** slices export a selector hook. The store is properly sliced (9 slices), so the convention has somewhere to live — but nothing has ever adopted it. Migrating 4 of 181 sites leaves the codebase less consistent than it is now.

"Touch-it = migrate-it" (`SKILL.md:57`) is scoped to the construct being edited — its own examples are "editing a 600-line dialog" and "adding a field to a `useState` form that has outgrown the RHF threshold". The principle is *don't bolt new code onto the drift you're editing*, not *fix every pre-existing pattern in a file you opened*.

Real debt, worth its own sweep with slice-level selector hooks. Not this PR.

## 🟡 `useState` threshold in `ArtifactsView` — declined, the rule doesn't exist

The review cites a "~5-call extraction threshold" for components. There is no such rule in the skill.

The only 5-`useState` rule is `references/hooks.md:71`: a **hook** over ~200 lines, or with more than 5 `useState`/`useRef` **and 3+ `useEffect`**, is a god-hook. `ArtifactsView` is a component and has zero `useEffect`. The component extraction triggers (`references/components.md:40–61`) are about JSX weight (leaf ~25 / feature ~60 lines), non-trivial `.map` bodies, repeated markup, colocated state, and nameable regions — never a `useState` count.

The proposed fix also backfires: `useArtifactViewDialogs()` holding 6 `useState` calls is much closer to the shape `hooks.md:71` actually warns about. The current code is 7 one-line dialog targets, each feeding exactly one dialog.

## 🟡 Doc drift — accepted, fixed in 29e7773

Correct, and the PR caused it. The bullet also still opened with "An optional **expiry**" while the rest of the change moved the vocabulary to retention/deletion, and `Last verified` was bumped to today, which asserts the page was checked.

Rewrote the sharing-model bullet: retention date, "scheduled for deletion" page, owner restores by choosing a new date. `docs/ubiquitous-language.md` has no artifact expiry entry, so nothing else needed following.

## 🔴 Missed — the dialog says "due" on the restore path

`retention-dialog.tsx:95` called `timeUntil(scheduled)`. That helper returns the literal string `"due"` for any past date (`lib/format-time.ts:34`).

An artifact inside the 7-day restore window is precisely the case this dialog exists to serve, and `artifact-row-menu-items.tsx:31` offers "Delete after…" for it unconditionally. So the owner opened the dialog and read:

> Currently deletes on 10 Aug 2026 — due.

…while the row badge beside it correctly read "deletion pending · 5d to restore", because `deletionState()` already handles the past-date case. Two surfaces, same fact, and the dialog's version carried no information.

Fixed in 3288633 by making `deletionState` the single source: its expired variant now also returns `restoreLeft`, and the dialog renders "Deletion is pending since 10 Aug 2026 — 5d left to restore it."

## 🟢 Advisory: `choice: string`

Fair, low value. `choice` is compared against two named constants and otherwise `Number()`d, so a widened type can't reach the mutation. Leaving as is.

---

`ui:check`, `docs:check`, and `common:check:comment-types` all pass. Confirming the review's positive finding for the record: `artifact-library-service.ts:401` patches `visibility` and `expiresAt` independently, so sending only `expiresInHours` cannot clobber visibility.

dam-code-guardian (changes requested):
## Code Review

### Changes since last review

**Last reviewed SHA:** `ed0d250c` → **Current SHA:** `29e77733`

| Finding | Status |
|---------|--------|
| 🟡 inline useStore selectors (3 files) | ✅ Fixed — confirmed: inline call at artifact-row.tsx:132 is outside the diff; the pattern pre-dates this PR and was incorrectly flagged. Decline upheld. |
| 🟡 ArtifactsView excessive useState | ✅ Fixed — confirmed: component has 6 `useState` calls (not 7) and 0 `useEffect` calls; the hook-extraction rule does not apply. Decline upheld. |
| 🟡 doc-drift — architecture doc terminology mismatch | ✅ Fixed — lines 46–47 updated to "scheduled for deletion" / "restore". |
| 🟢 unnarrowed choice type | ✅ Fixed — `choice` is initialised from `artifact.expiresAt` directly. |

🆕 New findings below.

---

### Summary

The PR completes the rename from "expiry" to "retention date" in the dialog layer and the primary doc section, but three terminology misses remain in peripheral locations, one dead parameter was introduced in the renderer, and a constants duplication exists between the UI and server packages. The new `RetentionDialog` component has a minor state-duplication pattern worth addressing.

Five 🟡 findings block merge. After they are resolved, the change is ready.

---

**🟡 `docs/architecture/artifact-library.md:164–167` — lifecycle section still uses "expiry sweep" and "whose expiry"**

Lines 164–167 were not updated. "expiry sweep" → "retention sweep" (or "deletion sweep"); "whose expiry" → "whose retention date".

```
164: The **expiry sweep** runs as a scheduled platform periodic job (its own
...
167: and permanently removes artifacts — private ones included — whose expiry
```

**Fix:** Replace "expiry sweep" with "retention sweep" on line 164 and "whose expiry" with "whose retention date" on line 167.

---

**🟡 `docs/architecture.md:70` — index blurb still says "expiry"**

```
70: — agents and users publish artifacts … with folders, expiry, and versions.
```

This file is not touched by the PR but it is the index entry for the updated doc. The PR renamed the concept everywhere — leaving the index entry stale creates confusion when readers first encounter the term.

**Fix:** Update to "with folders, a retention date, and versions" (or "with folders, retention, and versions").

---

**🟡 `packages/api-server/src/modules/artifact-library/viewer/renderer.ts:333–335` — dead parameter `expiredAt` in `renderExpired` input type**

The PRs refactor added a typed `expiredAt: Date` field to the `renderExpired` input type, but the function body never reads it. It is purely dead weight.

```ts
export function renderExpired(input: {
  id: string;
  expiredAt: Date;     // ← declared but never read
  restoreLeftMs: number | null;
}) { ... }
```

**Fix:** Remove `expiredAt` from the input type. Callers already pass it, so removing it from the interface is a non-breaking cleanup.

---

**🟡 `packages/ui/src/modules/artifacts/lib/format.ts:3` — duplicated restore-window constant**

`RESTORE_WINDOW_DAYS = 7` is the UI-side constant introduced by this PR. The server defines the same value as `EXPIRATION_GRACE_PERIOD_DAYS = 7` in `share-viewer-service.ts:8`. They are in sync today, but there is no shared source of truth: a future change to one silently skews the UI messaging (the dialog renders "7 days" from the UI constant while the server acts on the server constant).

**Fix:** Move the single authoritative value to the API contract package (`api-server-api`) so both the UI and the server import it, or at minimum add a co-location comment on both sides linking them.

---

**🟡 `packages/ui/src/modules/artifacts/components/retention-dialog.tsx:44–45` — `choice` duplicates server state**

```tsx
const [choice, setChoice] = useState<string>(
  scheduledDate === null ? NEVER : KEEP,
);
```

The initial value is derived from `artifact.expiresAt` via `scheduledDate`. If the parent query invalidates and replaces `artifact` while the dialog is open, `choice` will be stale relative to the new server value. The current `onClose()` call in `onSuccess` makes the window narrow, but the duplication is a state-lineage smell.

**Fix:** Either derive from `artifact.expiresAt` directly (`artifact.expiresAt ? KEEP : NEVER`) and document the "dialog must close on success" assumption, or gate the reset in an effect on `artifact.id`.

---

### Issue Fit

All deliverables from issue #3163 are met. ✅

---

### Approval bar

Open 🟡 findings that block merge:
1. `renderer.ts:333` — remove dead `expiredAt` param from `renderExpired` input type
2. `format.ts:3` — resolve duplicated `RESTORE_WINDOW_DAYS` / `EXPIRATION_GRACE_PERIOD_DAYS` constants
3. `artifact-library.md:164,167` — replace "expiry sweep" → "retention sweep" and "whose expiry" → "whose retention date"
4. `docs/architecture.md:70` — replace "expiry" → "retention" in the index blurb
5. `retention-dialog.tsx:44` — remove state duplication; initialise `choice` from `artifact.expiresAt` directly

---

```findings-json
{
  "pr": 3343,
  "sha": "29e77733c2cda056f681cd7714e0bc5ec7b08437",
  "verdict": "REQUEST_CHANGES",
  "findings": [
    {"id": "doc-drift-lifecycle-section", "severity": "yellow", "file": "docs/architecture/artifact-library.md", "lines": "164-167", "summary": "lifecycle section still uses \"expiry sweep\" and \"whose expiry\"", "fix": "Replace with \"retention sweep\" and \"whose retention date\""},
    {"id": "doc-drift-architecture-index", "severity": "yellow", "file": "docs/architecture.md", "lines": "70", "summary": "index blurb still says \"expiry\"", "fix": "Update to \"retention\" or \"retention date\""},
    {"id": "ts-dead-param-renderExpired", "severity": "yellow", "file": "packages/api-server/src/modules/artifact-library/viewer/renderer.ts", "lines": "333-335", "summary": "expiredAt declared in renderExpired input type but never read", "fix": "Remove expiredAt from the input type"},
    {"id": "ts-duplicate-window-constant", "severity": "yellow", "file": "packages/ui/src/modules/artifacts/lib/format.ts", "lines": "3", "summary": "RESTORE_WINDOW_DAYS=7 duplicates server-side EXPIRATION_GRACE_PERIOD_DAYS=7 with no shared source of truth", "fix": "Move to api-server-api contract package or add co-location comments"},
    {"id": "react-choice-state-duplication", "severity": "yellow", "file": "packages/ui/src/modules/artifacts/components/retention-dialog.tsx", "lines": "44-45", "summary": "choice useState duplicates server-owned expiresAt", "fix": "Derive from artifact.expiresAt directly; document close-on-success assumption"}
  ]
}
```

PetrBulanek:
Validated all five against the branch. Four are pushed in 285a2cd0, 2de525b7 and fdc60c8a. One is declined — the proposed fix is what the code already does.

Two of the findings needed a different fix than the one written, so those are noted below.

## 🟢 Doc drift, lifecycle section — fixed in fdc60c8a

Correct. Line 43 and lines 46–47 already said "retention date", so the page contradicted itself. Now "retention sweep" and "whose retention date passed".

Worth naming the deeper asymmetry: the server code still calls it expiry (`expiry-sweeper.ts`, and the grace constant before this PR). The docs now lead the code on this vocabulary, which inverts "docs follow code" from `documentation-guidelines.md:17`. Renaming the server identifiers is a separate change — page-level consistency was the blocking part.

## 🟢 Index blurb — fixed in fdc60c8a

Right, and `Last verified` on that file was `2026-08-11`, so it needed the bump too. Both done.

## 🟡 Dead `expiredAt` — fixed in 2de525b7, but the finding was wrong twice

The field is dead. Both claims around it are not.

**It does not come from this PR.** Lines 334–335 are unchanged context in the diff. On `main` the body already read only `withinGrace`, so the field has always been unread. The quoted snippet (`id: string`, `restoreLeftMs: number | null`) does not exist in either version — `renderExpired` takes `{ withinGrace, expiredAt }`.

**"Callers already pass it, so removing it from the interface is a non-breaking cleanup" is backwards.** Because the caller passes it, removing only the interface field breaks the build: `viewer-app.ts:53` hands over a fresh object literal, which gets excess-property checked.

The real cleanup is three sites, not one: the renderer type, the caller, and the `expired` variant of `SharedResolution` — `share-viewer-service.ts` computed `expiredAt: row.expiresAt!` purely to feed a parameter nobody read. The non-null assertion goes with it. A stale `expiredAt` in `share-viewer-app.test.ts` went too.

## 🟢 Duplicated 7-day constant — fixed in 285a2cd0

Accepted, and this one is genuinely on this PR — the UI had no grace-window knowledge before. The coupling is real: the dialog promises "restore for 7 days" while the sweeper decides when the row actually dies.

Took the contract-package option. `ARTIFACT_RESTORE_WINDOW_DAYS` now lives in `api-server-api` artifact-library types beside `ARTIFACT_INTERNAL_LINK_PREFIX`, which the UI already imports at `artifact-link-chip.tsx:2`. The UI dialog, `deletionState`, `share-viewer-service` and `expiry-sweeper` all read it. The server-local `EXPIRATION_GRACE_PERIOD_DAYS` is gone.

Rejecting the fallback for the record: "add a co-location comment on both sides" is not available here — CLAUDE.md allows no code comments outside the two registered test prefixes.

## 🔴 `choice` state duplication — declined, the fix is a no-op

The dialog already derives from `artifact.expiresAt`. `retention-dialog.tsx:40–46` reads `artifact.expiresAt` into `scheduled`, formats it into `scheduledDate`, then seeds `choice` from that. The suggested `artifact.expiresAt ? KEEP : NEVER` is the same condition in the same `useState`, so it changes nothing about the staleness the finding describes.

The alternative — "gate the reset in an effect on `artifact.id`" — cannot fire either. The dialog is mounted per artifact and `id` never changes underneath it.

There is a narrow real edge, and neither proposal addresses it: if `expiresAt` changes while the dialog is open, `choice` keeps the branch it was seeded with. Fixing that needs an effect keyed on `expiresAt`, not on `id`. The mutation closes the dialog on success, so the window is one concurrent edit from another surface. Not worth an effect.

Two notes on the changes-since table, since they affect what the next round can rely on:

- It marks the `choice` type advisory "✅ Fixed — `choice` is initialised from `artifact.expiresAt` directly". That advisory was declined and `useState<string>` is still there. Then finding 5 asks for that same initialisation on those same lines.
- It reports 6 `useState` calls in `ArtifactsView`. There are 7 (lines 37, 38, 39, 42, 44, 45, 49). Immaterial to the verdict — 0 `useEffect` is what makes the god-hook rule inapplicable — but the count is wrong.

---

`ui:check`, `api-server:check`, `api-server-api:check`, `docs:check` and `common:check:comment-types` pass. `api-server:test` 1005 passing, `ui:test` 290 passing.

dam-code-guardian (changes requested):
**Round 3 · full re-review · `fdc60c8`**

---

### Prior round (29e7773) resolution

All four accepted findings are confirmed fixed:
- ✅ Lifecycle-section terminology (renamed throughout the doc)
- ✅ Architecture-index blurb updated
- ✅ Dead `expiredAt` param removed from `SharedResolution` and `renderExpired`
- ✅ Duplicated 7-day constant consolidated into `ARTIFACT_RESTORE_WINDOW_DAYS`

The `choice` state decline was correct — `scheduledDate === null` is equivalent to `!artifact.expiresAt` for valid production data, so the proposed simplification was a no-op.

---

### Skills

| Skill | Result |
|---|---|
| doc-drift | ✅ aligned |
| issue-fit | ✅ delivers #3163, no undeclared scope |
| typescript-engineering | ✅ no violations |
| react-ui-engineering | 🟡 2 findings (below) |

---

### Findings

🟡 **R1 · `packages/ui/src/modules/artifacts/views/artifacts-view.tsx:43–44`** — `ArtifactsView` now has 7 `useState` calls: `search`, `uploadOpen`, `folderDialog`, `shareTarget`, `retentionTarget`, `previewTarget`, `deleteFolderTarget`. This PR added `retentionTarget`, pushing the five dialog-target slots past the point where local state is practical to scan. The five dialog targets (`share`, `retention`, `preview`, `folderDialog`, `deleteFolder`) form a natural group.

**Fix:** Extract into a `useArtifactDialogs()` hook at `modules/artifacts/hooks/use-artifact-dialogs.ts` that owns and exposes those five `useState` pairs. `search`, `uploadOpen`, and the server-state hooks stay in the view.

---

🟡 **R2 · `packages/ui/src/modules/artifacts/components/retention-dialog.tsx:53–61`** — `statusText` is declared as `const statusText = () => string` inside the render body and immediately called as `{statusText()}`. It closes over `deletion` and `scheduledDate`, which are already computed scalars. A plain function defined-and-called in the same render is just an inline expression with extra syntax; it provides no composability, memo boundary, or callsite flexibility.

**Fix:** Replace with a direct string assignment:
```ts
const statusText =
  deletion.state === "never"
    ? "This artifact is kept until you delete it."
    : deletion.state === "active"
      ? `Currently deletes on ${scheduledDate} — ${timeUntil(scheduled)}.`
      : deletion.restoreLeft
        ? `Deletion is pending since ${scheduledDate} — ${deletion.restoreLeft} left to restore it.`
        : `Deletion is pending since ${scheduledDate}.`;
```

---

### Approval bar

Both open items are 🟡. The PR may merge once R1 and R2 are addressed (or explicitly accepted with justification).

```json
{"findings":[{"id":"R1","severity":"yellow","file":"packages/ui/src/modules/artifacts/views/artifacts-view.tsx","line":43,"title":"7 useState calls in ArtifactsView — extract dialog-target group","fix":"useArtifactDialogs hook in modules/artifacts/hooks/"},{"id":"R2","severity":"yellow","file":"packages/ui/src/modules/artifacts/components/retention-dialog.tsx","line":53,"title":"statusText defined as function-called-once; should be a string constant","fix":"const statusText: string = deletion.state === 'never' ? ... : ..."}]}
```

PetrBulanek:
Both accepted, in 726c10e2 and 98e3209a. Neither took the fix as written, and R1 needs a correction to the record — from me, not from the review.

## 🟡 R1 — `ArtifactsView` useState count — accepted, and my round-1 decline was wrong

I declined this in round 1 with "there is no such rule in the skill." There is:

- `references/hooks.md:9–11` — **`[CRITICAL]` Extract a hook when: … A component has ~5 or more `useState` calls**. It names components explicitly.
- `references/state-management.md:175` — "If a component accumulates ~5 `useState` calls, step back: they're probably related. Group them into a reducer or a hook."

I found `hooks.md:71` (the god-hook rule, which does need `3+ useEffect`) and stopped there. The file has not changed since #268, so the rule was in front of me for all three rounds. Round 2 then recorded "decline upheld" against the same wrong premise, which is why this took until round 3 to land.

Scope holds too. The diff adds `retentionTarget`, so this PR moved the component from 6 to 7 — "don't bolt new code onto drift", `SKILL.md:57`. The `SKILL.md:58` carve-out covers folder moves and god-hook rewrites, not one component's local state.

This is also the opposite of the inline-`useStore` case I declined in round 1, and the contrast is what makes the decline there still right. Nine modules already carry `hooks/` directories with ~50 hooks, so this convention is adopted. `ArtifactsView` at 7 was second-highest in `packages/ui` — only `file-viewer.tsx` at 8 was worse.

**Fixed differently.** `useArtifactDialogs()` holding five `useState` pairs relocates the state; it does not reduce it. The six dialogs are mutually exclusive — I checked every setter call site, and no path opens two (`ArtifactPreviewDialog` receives only `artifact` and `onClose`). So one `useState<ArtifactDialog | null>` discriminated union carries all six. That is `state-management.md:186`'s "genuinely local composite state" shape, it takes the count 7 → 2, and it adds no file.

## 🟡 R2 — `statusText` called in JSX — accepted on different grounds

No rule in the skill covers defining and calling a function in a render body. I checked all of it. Composability, memo boundary and callsite flexibility are the review's reasoning, not a convention — worth naming, because a finding graded as blocking should cite something.

The codebase argument is the stronger one and the review did not make it: `{statusText()}` was the **only** render-time value function called from JSX anywhere in `packages/ui/src/modules`. The dozen other render-body arrows are all event handlers — `save`, `toggle`, `copy`, `commit`. So the shape was an outlier, on a file this PR adds. Behaviour was identical either way; it was called unconditionally.

**Fixed differently.** The suggested four-way nested ternary is not easier to read than the early-return function it replaces. Instead `deletionSummary` now lives in `lib/format.ts` beside `deletionState`, which already builds the `label` copy for the row badge. Both surfaces read their wording from one place — the same consolidation 3288633 made for the restore window. `deletionDate` went with it, replacing the dialog-local format constant.

---

`ui:check`, `ui:test` (290 passing) and `common:check:comment-types` pass. `ArtifactsView` is now 2 `useState` calls, and `packages/ui/src/modules` has zero render-time value functions called from JSX.

dam-code-guardian (dismissed):
**Round 4 · full re-review · `98e3209`**

---

### Prior round (`fdc60c8`) resolution

| Finding | Resolution |
|---------|-----------|
| 🟡 ArtifactsView 7 `useState` calls | ✅ Fixed — discriminated union `ArtifactDialog` reduces to 2 states |
| 🟡 `statusText` defined as function, called once | ✅ Fixed — `deletionSummary()` in `lib/format.ts` returns a plain string |

---

### Changes since last review

**Last reviewed SHA:** `fdc60c8` → `98e3209` (2 commits)

- `726c10e` — `refactor(ui)`: give the artifacts view one dialog state
- `98e3209` — `refactor(ui)`: move the deletion copy next to the deletion state

---

### Review

The two round-3 findings are resolved. The discriminated union approach in `ArtifactsView` is well-executed: `ArtifactDialog | null` consolidates six dialog variants into a single state slot, with typed narrowing at each render site. `deletionSummary()` is a correct extraction — the string is computed once per render from the artifact's `expiresAt`, not recreated as a function.

Two new issues found.

---

### 🟡 `packages/api-server-api/src/modules/artifact-library/types.ts:3` — Domain policy constant in contract package

`ARTIFACT_RESTORE_WINDOW_DAYS = 7` is a business-rule value (retention policy window) introduced in the contract package, which the architecture restricts to types, Zod schemas, and the service interface — no business logic or policy constants (`slice-composition.md`: "No business logic lives here").

Both the server (`expiry-sweeper.ts`) and the UI (`retention-dialog.tsx`, `lib/format.ts`) need this value, which is why the contract package was chosen — but the right pattern is to expose it via the service contract, not as a shared constant.

**Fix:** Move the constant to a domain file in the server module (e.g. `packages/api-server/src/modules/artifact-library/domain/retention.ts`) and expose the value to the client through a `retentionPolicy()` tRPC query returning `{ restoreWindowDays: number }`.

---

### 🟡 `packages/ui/src/modules/artifacts/components/sandbox-artifacts-section.tsx:23–25` — Parallel dialog states, inconsistent with `ArtifactsView`

`SandboxArtifactsSection` uses three separate `useState` calls for `shareTarget`, `retentionTarget`, and `previewTarget` — three variants of "which dialog is open." This PR adds `retentionTarget` as a third parallel state, while simultaneously introducing the discriminated-union `ArtifactDialog` pattern in `ArtifactsView` for exactly the same concern. The inconsistency is introduced in the same two-commit set.

**Fix:** Export `ArtifactDialog` from `artifacts-view.tsx` (or move it to a shared `types.ts` in the artifacts module) and apply the same `dialog: ArtifactDialog | null` state in `SandboxArtifactsSection`, matching the pattern now established in `ArtifactsView`.

---

### Approval bar

The two 🟡 findings above must be resolved before this can merge:

1. Move `ARTIFACT_RESTORE_WINDOW_DAYS` out of the contract package and expose it via the service contract.
2. Consolidate `SandboxArtifactsSection` dialog state into the `ArtifactDialog` discriminated union.

kapetr (approved):
Update startup tip: 
` packages/ui/src/modules/agents/startup-tips.ts:21 `— the startup tips still say "expiry". Line 21 lists it next to the public link with no mention of deletion, which teaches the exact wrong model this PR removes; line 33 is just the old word.

PR #3345, opened 2026-08-17
feat(ui): expand an artifact to full screen from the chat view

### LINKED ISSUE #3115: UI - Expand an artifact to full screen from the chat view
## Context

When an agent publishes an artifact, the natural place to look at it is right there in the chat — the panel opens beside the conversation. But that panel is narrow and can only be dragged so wide, and the chat has to keep its share of the window. For anything with real layout — a dashboard, a wide table, a report — you end up squinting at a cramped version of it.

The artifact library's own preview already has an expand-to-fullscreen control. The chat panel is the one artifact preview surface without a path to a bigger view, so the same artifact is easier to read from the library than from the conversation that produced it. Reported on 0.2.12.

Image

## Problem

In the chat artifact panel the only controls are: switch versions, Share, Source, download, and close. There's no way to make the artifact bigger.

So to actually read something an agent just made, a user has to leave the conversation — go find it in the library and open it there, or download it, or publish a share link and open that in a browser tab. Each of those is a detour away from the work, and share-link is worse: it means making something public just to read it yourself.

The workaround in use today is exactly that detour: open the artifact library, find the artifact again, and expand to full screen from there. It works, which is the point — the capability already exists one surface over. It's just not reachable from the conversation that produced the artifact.

## Goal

A user can go from the chat artifact panel to a full-screen view of that artifact in one click, and back again without losing their place in the conversation.

## User Stories

- As someone reviewing an artifact an agent just made, I want to expand it to full screen from the chat so that I can read it properly without leaving the conversation.
- As someone looking at a wide dashboard or report, I want the artifact to use the whole window so that its layout isn't squeezed into a narrow column.
- As a user, I want the expand control to sit where I already expect it from the library so that the two views behave the same way.

## Scope

**In:** an expand/full-screen control in the chat artifact panel, and a way back out of it.

**Not in:** redesigning the panel's other controls, or changing how artifacts render.

PetrBulanek:
The artifact preview docked beside a conversation is the one artifact surface with no path
to a bigger view. The panel is narrow, the chat keeps its share of the window, and anything
with real layout — a dashboard, a wide table, a report — renders squeezed. Reading it means
leaving the conversation: find the artifact again in the library, download it, or publish a
share link just to read it yourself.

After this change, a user expands the artifact to full screen in one click from the chat,
and exits back to the same conversation with nothing lost. The control sits where the
library preview already puts it, so the two views behave the same way. If the agent
publishes a new version while the artifact is full screen, the view stays open and follows
it rather than dropping the reader back to the chat.

Closes #3115

## Sub-issues

- [x] 01 — Expand the docked artifact preview to full screen


dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `c1abfea`

## PR #3345: feat(ui): expand an artifact to full screen from the chat view
**Author:** PetrBulanek | **Branch:** feat/3115-artifact-fullscreen-chat → main | **Changes:** +54 −13 (1 files)

### Summary
The docked artifact panel gets a `Maximize` button that opens the preview in the shared `FullscreenPreviewDialog`, and shows an "Opened in fullscreen" placeholder while expanded. The frame element is hoisted to a `frame` variable, so both surfaces render the same node.

### Findings
- 🟡 **Warning:** stale `fullscreen` state re-opens the dialog (`packages/ui/src/modules/artifacts/components/docked-artifact-panel.tsx:80`)
- 🟡 **Warning:** expand control placement and visibility differ from the library surface it must mirror (`packages/ui/src/modules/artifacts/components/docked-artifact-panel.tsx:131`)
- 🟢 **Suggestion:** `FullscreenPreviewDialog` declares `aria-modal="true"` but traps no focus, and the background stays interactive. Focus stays on the covered Maximize button after opening. This is pre-existing in the shared dialog and its two other call sites — worth a separate issue for a focus trap, initial focus, and focus restore.
- ✅ **Looks good:** the version-follow claim in the PR body holds. An `artifacts` live event invalidates `trpc.artifactLibrary.pathKey()` (`packages/ui/src/modules/live-events/invalidation.ts:21`), so `useArtifact` refetches and `shownVersion` tracks `latest` while the dialog stays mounted.

✅ No findings.

### Issue Fit
- 🟡 **Warning:** The expand control sits after Download, not next to Source. It also appears only after the preview loads. Issue #3115's third user story and the PR body both promise the library control's placement and behavior (`packages/ui/src/modules/artifacts/components/docked-artifact-panel.tsx:131`).
  **Fix:** When a control is specified to mirror an existing surface, copy that surface's placement and visibility condition. Render `Maximize` directly after the Source toggle, and gate it on `showFrame` alone, as `artifact-preview-dialog.tsx:76` does — not on loaded preview data.

- 🟡 **Warning:** `expanded` masks the `fullscreen` flag instead of resetting it. Switching to Source while expanded hides the dialog but keeps `fullscreen` true, so switching back to Preview re-opens fullscreen unexpectedly (`packages/ui/src/modules/artifacts/components/docked-artifact-panel.tsx:80`). The covered Source button stays keyboard-reachable, because the dialog traps no focus.
  **Fix:** Reset the owning state when the condition that gates it goes away. Clear `fullscreen` in the `setShowSource` handler, or in an effect on `showFrame` — do not only AND it into a derived render gate.
- 🟢 **Suggestion:** `FullscreenPreviewDialog` sets `aria-modal="true"` but traps no focus and leaves the background interactive (`packages/ui/src/modules/artifacts/components/docked-artifact-panel.tsx:197`). Pre-existing in the shared dialog and its two other call sites — worth a separate issue.

### Verdict
COMMENT — the feature works and the refactor is faithful. Two 🟡 findings hold it below approval: the stale `fullscreen` state, and a control that does not mirror the library surface as issue #3115 requires.

---

packages/ui/src/modules/artifacts/components/docked-artifact-panel.tsx
```diff
@@ -51,6 +60,25 @@ export function DockedArtifactPanel() {
...
+        postData={feedPostForShown}
+      />
+    ) : null;
+  const frameFallback = (
+    <p className="py-6 text-center text-sm text-muted-foreground">
+      {preview.isLoading ? "Loading preview…" : "No preview available."}
+    </p>
+  );
+  const frameShowing = showFrame && frame !== null;
+  const expanded = fullscreen && showFrame;
```
dam-code-guardian, line 80: 🟡 **Warning:** `expanded` masks the `fullscreen` flag instead of resetting it. Switching to Source while expanded hides the dialog but keeps `fullscreen` true, so switching back to Preview re-opens fullscreen unexpectedly. The covered Source button stays keyboard-reachable, because the dialog traps no focus.

**Fix:** Reset the owning state when the condition that gates it goes away. Clear `fullscreen` in the `setShowSource` handler, or in an effect on `showFrame` — do not only AND it into a derived render gate.

_Reported by: diff review + React UI Engineering._

packages/ui/src/modules/artifacts/components/docked-artifact-panel.tsx
```diff
@@ -100,6 +128,17 @@ export function DockedArtifactPanel() {
             <Download size={14} />
           </Button>
         )}
+        {frameShowing && (
```
dam-code-guardian, line 131: 🟡 **Warning:** The expand control sits after Download, not next to Source, and it appears only after the preview loads (`frameShowing` requires `preview.data`). Issue #3115's third user story asks for the control to sit where the library already puts it.

**Fix:** When a control is specified to mirror an existing surface, copy that surface's placement and visibility condition. Render `Maximize` directly after the Source toggle, and gate it on `showFrame` alone, as `artifact-preview-dialog.tsx:76` does.

_Reported by: Issue Fit._

kapetr (approved):

Issue #3131, opened 2026-08-03
UI - Model settings claim "Not set" for a stopped sandbox


kapetr:
## Context

The sandbox home page shows Model settings read-only while the sandbox is stopped, with a "Start agent to edit" action. The current model, mode and effort live inside the sandbox and can only be read while it's running, so on a cold load of a stopped sandbox the platform has nothing to show and every picker falls back to "Not set".

"Not set" is also a real, selectable option in each picker — it means "clear this and let the harness choose" — so the placeholder is indistinguishable from a deliberate choice.

Skills already handle this better: the platform keeps its own record of which GitHub-sourced skills it installed and shows those rows while the sandbox is stopped. Only the parts that exist solely inside the sandbox go blank.

Related: #2654 (cache agent-resolved settings — skills and model — so the config page shows them while stopped) is the broad version of the fix; #2789 proposed persisting the applied model config and was closed as not planned.

## Problem

A stopped sandbox reports that its model, mode and effort are all unset. For most sandboxes that's false — they do have settings, the platform just can't read them right now. Users can't tell "nothing is configured" from "we don't know", and the read-only section ends up worse than empty: it's wrong. It also undercuts the reason the section is shown while stopped at all, which is to let people see the configuration without starting the agent.

<img width="697" height="403" alt="Image" src="https://github.com/user-attachments/assets/42deb0f5-a630-43a8-91e6-250cbe854d52" />

## Expected Behavior

Two halves, both needed:

- **Remember the values.** A stopped sandbox's Model settings show the last known model, mode and effort — the values from the last time it ran — presented read-only.
- **A placeholder for when there's nothing to remember.** A sandbox that has never run, or one whose values were never captured, still has nothing to show. That case needs wording that reads as "we don't know yet" — and it can't stay "Not set", because that's the label of a real option in the same picker. The placeholder has to be distinguishable from a deliberate choice.

## Actual Behavior

Model, Mode and Effort each read "Not set" — the same wording as the explicit "Not set" option in the picker menu — regardless of how the sandbox is actually configured.

## Steps to Reproduce

1. Open a running sandbox and set a model (and mode / effort, if the harness offers them).
2. Stop the sandbox.
3. Reload the sandbox home page and look at Model settings.
   → Model, Mode and Effort all read "Not set".

spdempsey:
Discussion for the content aspect of this is happening on https://github.com/dam-agents/dam/issues/3133

spdempsey:
[...]

PR #3346, opened 2026-08-17
fix(ui): rename the cleared model setting option to "Default"

### LINKED ISSUE #3133: UI - Rename the "Not set" model setting option to "Default"
## Context

Each Model settings picker — Model, Mode, Effort — offers an option labelled **"Not set"**, described as *"Clears the key — the harness picks on its own"*. Picking it is a legitimate, deliberate configuration: the harness falls back to its own default. Both the label and its description name the mechanism (a key being absent from a config file) rather than the outcome the user gets.

The same wording also does double duty as a placeholder: a stopped sandbox shows "Not set" in these pickers when the platform simply can't read the values (#3131). While one label serves both, a deliberate choice and an unknown value are indistinguishable.

## Problem

"Not set" describes what the platform does to its stored configuration, not what the agent will do. That the harness then picks the model, mode or effort itself is only discoverable by reading the description — which itself opens with "Clears the key".

Choosing between "Not set" and a named model, a user can't tell that the first is a real, supported option rather than a gap in their setup. It reads as something unfinished.

Image

## Goal

The option describes its outcome — the agent uses its own default — and the wording belongs to that choice alone, so a stopped sandbox's unknown values can read as something visibly different (#3131).

## Done when

- The option reads **"Default"**, with a description stating that the agent chooses, and no mention of keys or clearing.
- The same wording appears on the picker itself when that option is the current value, not just in the open menu.
- Applies to every Model settings picker (Model, Mode, Effort) on both surfaces that render them — the chat rail and the sandbox home page.

## Open Questions

- Is "Default" unambiguous about *whose* default it is? "Agent default" is more precise but longer, and the pickers are narrow.
- The current description says "harness". That term does appear in user-visible copy elsewhere (the sandbox wizard's "Pick a harness to start"), but "the agent chooses" is plainer here — worth deciding once for the copy pass rather than per-string.

kapetr:
Closes #3133. Also covers the placeholder half of #3131.

The cleared option in each Model settings picker named the config mechanism rather than the outcome, and the same words stood in whenever the platform could not read a stopped sandbox — so a deliberate choice and an unknown value looked identical. This is the copy pass agreed on the issue, plus the read-only picker face needed to carry a second line.

Small drifts from the issue, worth a look:

- The "never run" sandbox keeps its own prose. The issue dropped that state as unreachable, but the branch still exists in code and explains itself better than "Unknown" would.
- "Unknown" is a section-level state, for a sandbox that ran but never reported its settings. One that does have a cached snapshot still shows those values, so a cleared key there reads "Default" — the platform did observe it as absent.
- Two "agent" → "sandbox" copy fixes on the wake button and the fallback prose ride along. They are not from this issue.

The Mode and Opus labels live in the claude-code runtime manifest, read from inside the image at hello, so existing sandboxes keep the old labels until the image is rebuilt and they restart.

### DIFF
--- a/packages/agents/claude-code/runtime-manifest.yaml
+++ b/packages/agents/claude-code/runtime-manifest.yaml
@@ -23,7 +23,7 @@ drivers:
               name: "Fable"
               description: "Fable 5 · Most capable for the hardest, longest-running tasks"
             - value: opus
-              name: "Opus (recommended)"
+              name: "Opus"
               description: "Opus 5 · Best for everyday, complex tasks"
             - value: sonnet
               name: "Sonnet"
@@ -40,8 +40,8 @@ drivers:
               name: "Auto"
               description: "Use a model classifier to approve/deny permission prompts"
             - value: default
-              name: "Default"
-              description: "Standard behavior, prompts for dangerous operations"
+              name: "Ask for approvals"
+              description: "Ask for permission before dangerous operations"
             - value: acceptEdits
               name: "Accept Edits"
               description: "Auto-accept file edit operations"
--- a/packages/ui/src/components/ui/read-only-field.ts
+++ b/packages/ui/src/components/ui/read-only-field.ts
@@ -1,0 +1,4 @@
+export const READ_ONLY_FIELD_BASE =
+  "flex w-full items-center rounded-md border border-input bg-muted/40 px-4 text-sm text-muted-foreground";
+
+export const READ_ONLY_FIELD = `${READ_ONLY_FIELD_BASE} h-10`;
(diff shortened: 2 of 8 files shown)


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `21e537eb`

## PR #3346: fix(ui): rename the cleared model setting option to "Default"
**Author:** kapetr | **Branch:** fix/3133-model-settings-default-copy → main | **Changes:** +107 −73 (8 files)

### Summary
Renames the cleared model setting option from "Not set" to "Default" across all three pickers, splits the read-only picker face into a shared `ReadOnlyOptionFace` component, and adds a new "Unknown / Start sandbox to view" section for sandboxes that have run but not yet reported their harness config.

### Findings
_No findings._

✅ No findings.

### Issue Fit
Linked issue: #3133 "UI - Rename the 'Not set' model setting option to 'Default'" (closing keyword). Also covers the placeholder half of #3131 (plain reference; #3131 is closed and its discussion confirms the copy decisions were consolidated into #3133).

✅ Delivers the issue's ask; no undeclared scope.

✅ No findings.

✅ No findings.

### Verdict
APPROVE — Copy and component refactor are clean; no correctness, security, or architectural issues found.

---

PetrBulanek (approved):

PR #3358, opened 2026-08-18
feat(observability): expose agent turns per surface as a metric

pilartomas:
## Why

`SessionTurnRelayed` and `ChannelTurnRelayed` already record every turn into the activity log, which answers *how many turns each surface drove* — but only through SQL, behind the `platform-inspector` role. This exposes the same fact as a time series, for a dashboard that wants a rate.

## What

One counter, `platform.turn.total`, with a single dimension `platform.turn.surface` — the surface the event already carries (`ui`/`cli`/`other` resolved from the caller's credential at the relay upgrade, or the messenger for a channel turn).

It is a **subscriber on the existing turn events**, not new instrumentation at the call sites: no relay, channel-worker, UI, or contract changes. Both sinks therefore answer to one definition of a turn. A failed turn counts like any other — the counter measures what was asked, not what came back.

The instrument lives in `core/turn-metrics.ts`; the bus subscription lives in `src/sagas/turn-metrics.ts` alongside the agent-artifacts sweeper, since it reacts to events owned by two other contexts and belongs to neither.

## Deliberately not included

- **`outcome`** — known only on the channel side; a dimension present on some surfaces and not others makes any filter on it silently drop the rest. Per-turn outcomes stay in `usage_channel_turns_by_agent`.
- **agent identity** — multiplies series by fleet size for a question that does not need it. If ever added it must not reuse `platform.agent.id`, whose *absence* marks a signal as platform telemetry rather than agent telemetry.

## Limits worth knowing before charting it

- A read-along (ambient) turn is indistinguishable from one that addressed the agent, because `ChannelTurnRelayed` carries no such marker — so an ambient-enabled channel reads higher than the attention it received. Closing that means adding a field to that event; not done here.
- Terminal-mode sessions never appear: a PTY has no message boundary, which is also how the CLI chats.
- The two sinks are separately enabled (the counter needs the OTel SDK active, the log needs activity tracking on) and fail independently, so a divergence means a sink is off or dropping.

## Tests

`turn-metrics.test.ts`: one series per surface, separate accumulation across `ui`/`slack`/`telegram`/`cli`, both event types landing on the same counter via the saga, a failed channel turn counting like a successful one, and an unrecognized surface folding into the `other` bucket instead of minting a new series.

`mise run check` and `mise run test` pass.

## Note on the earlier review

The first guardian pass flagged description drift, and it was right: this PR was rebased onto main after `chore(usage): extend and correct raw interaction collection` landed the same turn recording, and I reduced the change to just the metric sink without rewriting the body. The trigger dimension, the `apiServer.otlpEndpoint` helm value, the UI `_meta` stamping and the shared contract vocabulary all belonged to the pre-rebase version and are gone. This description now matches HEAD.


docs/architecture/observability.md
```diff
@@ -57,7 +57,17 @@ Harnesses produce telemetry by exporting it themselves over OTLP — the platfor
...
+- The **api-server** emits one trace per incoming request with a child span per tRPC procedure (and spans for outbound calls: agents, Keycloak, channels, Redis, the ext-authz gRPC checks), per-procedure duration/outcome metrics plus Node runtime health (event loop, GC, heap), the **turn counter** below, and its structured logs with trace correlation. Health-probe requests are not traced. The primary Postgres pool is not yet instrumented (no driver instrumentation exists for it); that gap is tracked as follow-up work.
+
+### Turn counter
+
+[usage-tracking](usage-tracking.md) records a turn per turn on every surface and is the place to ask *how many*. This counter is the same fact as a time series, for a dashboard that wants a rate without going through SQL and the inspector role. It is a subscriber on the same turn events that subsystem persists, so the two cannot drift apart in what they count: a turn arrives on the counter exactly when a row arrives in the log.
```
dam-code-guardian: 🟡 **Warning:** This paragraph promises the counter and the activity log "cannot drift apart", and that a turn reaches the counter exactly when a row reaches the log. Two mechanisms break that in both directions: `recordTurn` is a no-op when no meter provider is registered — and the telemetry backend is disabled by default (`observability.md:9`), so rows exist with no counts; and the activity-log subscriber discards insert failures to stderr (`persist-activity.ts:117`), so a count can exist with no row.

**Fix:** Describe the two as sharing a *source*, not a count. When each sink has its own enablement condition and its own failure handling, state those conditions and the divergence they permit instead of claiming exact correspondence.

packages/api-server/src/core/turn-metrics.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export function recordTurn(surface: string): void {
+  getCounter().add(1, { "platform.turn.surface": surface });
+}
+
+export function resetTurnMetricsForTest(): void {
+  counter = null;
+}
+
+export function startTurnMetricsSaga(): Subscription {
```
dam-code-guardian: 🟡 **Warning:** This is the only file in `core/` that imports the event bus — on `main` no `core/` file does — while all 7 other `start*Saga` functions live in `modules/<module>/sagas/`. The placement sets a new precedent against the repo's own layout.

**Fix:** Keep event-bus subscriptions out of `core/`; put every saga in a `sagas/` directory. Here, move this file to `modules/usage/sagas/turn-metrics.ts` beside `persist-activity.ts`, which already consumes the same two turn events, and start and stop it with that module (`composeUsageModule` / its `stop()`) rather than from `bootstrap.ts`.

packages/api-server/src/core/turn-metrics.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (!counter) {
+    counter = metrics.getMeter(SCOPE).createCounter("platform.turn.total", {
+      description: "Turns submitted to an agent, by originating surface",
+    });
+  }
+  return counter;
+}
+
+export function recordTurn(surface: string): void {
+  getCounter().add(1, { "platform.turn.surface": surface });
```
dam-code-guardian: 🟡 **Warning:** The counter records one dimension, the surface. The PR body specifies a second, `platform.turn.trigger` (`user`/`retry`/`auto`/`ambient`), and calls it the "apples-to-apples slice" that makes the surfaces comparable at all. Without it a browser count absorbs first-visit greetings and retries, which is the comparison this PR sets out to enable.

**Fix:** Either add the trigger dimension, or rewrite the description to match HEAD. Every dimension a PR declares must exist in the diff — after a force-push that drops scope, the body is what goes stale.

packages/api-server/src/__tests__/unit/turn-metrics.test.ts
```diff
@@ -0,0 +1,140 @@
+/** TEST_OVERVIEW: the turn counter that makes browser and messenger turns
```
dam-code-guardian, line 1: 🟡 **Warning:** The PR body's Tests section claims this file covers per-`(surface, trigger)` series isolation and "the trigger parser's fallback to a typed prompt for absent or unrecognized stamps". Neither is here, and no trigger parser exists in the diff.

**Fix:** Describe only the assertions the file makes. Remove any stated test case naming a component absent from the code.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `f8e48c98`

## PR #3358: feat(observability): count agent turns by surface and trigger
**Author:** pilartomas | **Branch:** feat/turn-metrics → main | **Changes:** +216 −2 (4 files)

### Summary
The PR adds an OTel counter `platform.turn.total`, dimensioned by originating surface, fed by an RxJS saga that subscribes to the existing `SessionTurnRelayed` and `ChannelTurnRelayed` bus events. The code is sound; the findings are that the PR description and the new doc prose both describe a larger feature than the diff delivers.

### Findings
- 🟡 **Warning:** description drift — the body specifies a second dimension `platform.turn.trigger` (`user`/`retry`/`auto`/`ambient`) and calls it the slice that makes surfaces comparable. HEAD records only the surface, so the stated apples-to-apples query is not possible. Reported by issue-fit. (`packages/api-server/src/core/turn-metrics.ts:27`)
  **Fix:** Every dimension, field, or behavior a PR description declares must exist in the diff at HEAD. After a force-push that drops scope, rewrite the description to match the delivered code. This applies to all four drift findings below — they are one class with one remedy.
- 🟡 **Warning:** the body promises a new `apiServer.otlpEndpoint` helm value with all three render paths verified. The chart carries no such key, so an install with its own collector cannot receive this counter. Reported by issue-fit. (`deploy/helm/platform/values.yaml`)
  **Fix:** A deployment knob a PR claims to add must appear in the chart values and templates in the same diff. Do not state verification of render paths that no code path exercises.
- 🟡 **Warning:** the body promises UI `_meta.platform` trigger stamping and a shared trigger vocabulary in the contract package next to `SessionType`. The diff touches no UI file and no contract file. Reported by issue-fit. (`packages/api-server/src/core/turn-metrics.ts:1`)
  **Fix:** Producer and consumer plumbing a description names must be present in the diff, or the description must drop it.
- 🟡 **Warning:** the Tests section claims coverage of per-`(surface, trigger)` series isolation and of a trigger parser with a fallback for absent or unrecognized stamps. The test file covers neither, and no trigger parser exists. Reported by issue-fit. (`packages/api-server/src/__tests__/unit/turn-metrics.test.ts:1`)
  **Fix:** Describe only the assertions the test file makes. Remove any stated test case that names a component absent from the code.
- 🟡 **Warning:** the new doc section promises the counter and the activity log "cannot drift apart", and that a turn reaches the counter exactly when a row reaches the log. Two mechanisms break this in both directions. Reported by doc-drift. (`docs/architecture/observability.md:64`)
  **Fix:** Describe two subscribers of one event stream as sharing a source, not as sharing a count. State each sink's own enablement condition and failure handling, and the divergence they permit.
- 🟡 **Warning:** the saga sits in `core/`, which no other event-bus subscriber uses — all 7 other `start*Saga` functions live in `modules/<module>/sagas/`, and on `main` no `core/` file imports the bus at all. Reported by typescript-engineering. (`packages/api-server/src/core/turn-metrics.ts:34`)
  **Fix:** Keep event-bus subscriptions out of `core/`; put every saga in a `sagas/` directory. Here, move the file to `modules/usage/sagas/turn-metrics.ts` beside `persist-activity.ts`, which consumes the same two events, and start and stop it with that module.
- ✅ **Looks good:** the module-level counter cache is safe despite telemetry being off by default. The Dockerfile preloads the SDK with `--import ./dist/telemetry.js` before `dist/index.js`, so the meter provider is registered before any turn arrives, and `recordTurn` is a documented no-op otherwise.

- 🟡 **Warning:** The page promises that the turn counter and the activity log cannot count differently, and that a turn reaches the counter exactly when a row reaches the log. The code breaks this promise in both directions: `recordTurn` does nothing when no meter provider is registered, and the telemetry backend is off by default, so rows exist with no counts; the activity-log subscriber also discards insert failures, so a count can exist with no row (`docs/architecture/observability.md:64`)
  **Fix:** Describe two subscribers of one event stream as sharing a source, not as sharing a count. When each sink has its own enablement condition and its own failure handling, state those conditions and the divergence they permit instead of claiming an exact correspondence.

### Issue Fit
- 🟡 **Warning:** The PR body specifies two counter dimensions, `platform.turn.surface` and `platform.turn.trigger` (`user`/`retry`/`auto`/`ambient`), and calls the trigger the slice that makes surfaces comparable. The code records only the surface, so the declared apples-to-apples query is not possible (`packages/api-server/src/core/turn-metrics.ts:27`).
  **Fix:** Every dimension, field, or behavior a PR description declares must exist in the diff at HEAD; after a force-push that drops scope, rewrite the description to match the delivered code before requesting review.
- 🟡 **Warning:** The body promises a new `apiServer.otlpEndpoint` helm value (default empty) and states all three render paths were verified. The chart carries no such key, so an install with its own collector cannot receive this counter (`deploy/helm/platform/values.yaml:577`).
  **Fix:** A deployment knob a PR claims to add must appear in the chart values and templates in the same diff; do not claim verification of render paths that no code path exercises.
- 🟡 **Warning:** The body promises the UI stamps the trigger onto the prompt's `_meta.platform` and a shared trigger vocabulary in the contract package next to `SessionType`. The diff touches no UI file and no contract file (`packages/api-server/src/core/turn-metrics.ts:1`).
  **Fix:** Producer and consumer plumbing a PR description names must be present in the diff, or the description must drop it; a shared vocabulary claim needs the actual shared declaration.
- 🟡 **Warning:** The Tests section claims coverage of per-`(surface, trigger)` series isolation and of a trigger parser that falls back to a typed prompt for absent or unrecognized stamps. The test file covers neither, and no trigger parser exists (`packages/api-server/src/__tests__/unit/turn-metrics.test.ts:1`).
  **Fix:** Describe only the assertions the test file actually makes; a stated test case that names a component absent from the code must be removed from the description.
- 🟢 **Suggestion:** The body names the increment sites as the ACP relay and each channel worker, reading the trigger off the `session/prompt` frame. The code instead subscribes one saga to the `SessionTurnRelayed` and `ChannelTurnRelayed` bus events (`packages/api-server/src/core/turn-metrics.ts:57`).

- 🟡 **Warning:** The new event-bus saga is in the shared `core/` technical directory instead of a `sagas/` directory. It is the first and only file in `core/` that imports the event bus, so it sets a new precedent against the repo's own layout, where all 7 other `start*Saga` functions are in `modules/<module>/sagas/`. (`packages/api-server/src/core/turn-metrics.ts:34`)
  **Fix:** Put every event-bus saga in a `sagas/` directory — `modules/<owner>/sagas/` when one bounded context owns the reaction, or the top-level `src/sagas/` when it crosses contexts. Keep `core/` free of event-bus subscriptions. Here, move the file to `modules/usage/sagas/turn-metrics.ts`, next to `persist-activity.ts`, which already consumes the same two turn events; start it from `composeUsageModule` and tear it down in that module's `stop()`, rather than from `bootstrap.ts`.

### Verdict
COMMENT — the implementation is correct and its bootstrap wiring is clean, but the description and the new doc prose both promise a feature larger than HEAD delivers, and the saga's placement sets a new precedent against the repo's own layout.

---

pilartomas:
All six findings addressed.

**Description drift (4 findings) — accepted, root cause fixed.** The body advertised the pre-rebase design. This branch was rebased onto main after `chore(usage): extend and correct raw interaction collection` landed the same turn recording, and I cut the change down to just the metric sink without rewriting the description. The trigger dimension, the `apiServer.otlpEndpoint` helm value, the UI `_meta` stamping, the shared contract vocabulary and the test assertions named in the old body were all dropped with that scope. The description now describes HEAD only.

**Doc overclaim — accepted.** "the two cannot drift apart in what they count" is wrong: the counter needs the OTel SDK active, the log needs activity tracking on, and each fails independently. The page now says both sinks answer to one definition of a turn, and that a divergence means a sink is off or dropping rather than two notions of a turn.

**Saga in `core/` — accepted, with a different destination than suggested.** `core/` is now free of bus subscriptions: the instrument stays there as a plain helper, and the subscription moved to `src/sagas/turn-metrics.ts`.

I put it in the top-level `sagas/` rather than `modules/usage/sagas/` because the reaction crosses contexts and belongs to neither — it consumes an event from the relay and one from the channel workers, and its output is an OpenTelemetry counter owned by the observability subsystem, not the usage bounded context's Postgres log. Putting it under `modules/usage` would give that context a sink it never reads and tie the metric's lifecycle to activity tracking's composition. The precedent for this placement is `sagas/agent-artifacts-sweeper.ts`, and it matches the review's own "top-level `src/sagas/` when it crosses contexts" clause. Happy to move it under `modules/usage/sagas/` if you read the ownership the other way.

docs/architecture/observability.md
```diff
@@ -57,7 +57,17 @@ Harnesses produce telemetry by exporting it themselves over OTLP — the platfor
...
+
+### Turn counter
+
+[usage-tracking](usage-tracking.md) records a turn per turn on every surface and is the place to ask *how many*. This counter is the same fact as a time series, for a dashboard that wants a rate without going through SQL and the inspector role. It is a subscriber on the same turn events that subsystem persists, so both sinks answer to one definition of a turn rather than each counting their own way. They are not guaranteed to agree in practice: each is separately enabled — the counter needs the SDK active, the log needs activity tracking on — and each fails independently, so treat a divergence as a sink being off or dropping, not as two different notions of a turn.
```
dam-code-guardian: 🟡 **Warning:** This paragraph sends the reader to usage-tracking as "the place to ask *how many*", and frames the counter as the same fact without SQL. That read path does not exist for relay surfaces. Session turns are persisted (`packages/api-server/src/modules/usage/sagas/persist-activity.ts:108`), but no SQL view reads them: `packages/db/drizzle/0001_usage_views.sql` defines `channel_turn` views only, `VIEW_NAMES` (`packages/api-server/src/modules/usage/services/report-service.ts:3`) lists no session-turn view, and `/api/usage` accepts an allowlisted view name only — ad-hoc SQL is deliberately not exposed (`docs/architecture/usage-tracking.md:177`). So `ui`, `cli` and `other` turns are unreachable through the inspector API, and for those surfaces this counter is new capability, not a second view of an old fact.

**Fix:** When a change describes itself as re-exposing an existing fact, name the concrete read path that already serves it, per surface. If that path does not exist for some surfaces, say so and describe the change as new capability for them.

packages/api-server/src/core/turn-metrics.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export function resetTurnMetricsForTest(): void {
```
dam-code-guardian: 🟡 **Warning:** `resetTurnMetricsForTest` is a test-only seam exported from a production module, and it is the only way to clear the module-level `counter` cache (line 5). It ships in the production bundle, so any importer can drop the live counter, and production code has no supported way to rebuild it if the meter provider changes.

**Fix:** Resolve the meter through an injected dependency, per the repo's service-factory pattern, so a test swaps the provider instead of reaching into module state. Never export a test-only reset from a production module.

_Note: the cache itself is safe at runtime — `telemetry.js` is preloaded via `node --import` before the app evaluates (`packages/api-server/Dockerfile:44`), so the provider is registered before the first `recordTurn`._

packages/api-server/src/core/turn-metrics.ts
```diff
@@ -0,0 +1,22 @@
...
+function getCounter(): Counter {
+  if (!counter) {
+    counter = metrics.getMeter(SCOPE).createCounter("platform.turn.total", {
+      description: "Turns submitted to an agent, by originating surface",
+    });
+  }
+  return counter;
+}
+
+export function recordTurn(surface: string): void {
```
dam-code-guardian: 🟡 **Warning:** `surface` is an unconstrained `string`, so the signature accepts any value as a metric attribute. Both current call sites are bounded — `clientSurface` returns `ui`/`cli`/`other` (`packages/api-server/src/apps/api-server/admission/auth.ts:207`) and `ChannelTurnRelayed.channel` is `slack`/`telegram` (`packages/api-server/src/events.ts:97`) — so there is no live cardinality bug. The type is what stops a future caller passing a user-controlled value, which would create one time series per distinct string.

**Fix:** Type every metric attribute parameter as a closed union of the permitted literals, so the compiler rejects an unbounded label value at the call site.

```suggestion
export function recordTurn(surface: TurnSurface): void {
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `e29bdc74`

## PR #3358: feat(observability): expose agent turns per surface as a metric
**Author:** pilartomas | **Branch:** feat/turn-metrics → main | **Changes:** +217 −2 (5 files)

### Summary
The PR adds an OTel counter `platform.turn.total`, dimensioned by originating surface, fed by an RxJS saga on the existing `SessionTurnRelayed` and `ChannelTurnRelayed` events. All six findings of the previous round are resolved: the body matches HEAD, the doc no longer claims the two sinks cannot diverge, and the saga moved out of `core/` into `src/sagas/`.

### Changes since last review
Previous HEAD: `f8e48c98` (2026-08-18T10:11:36Z) — verdict COMMENT

- ✅ **Fixed:** body declared a `trigger` dimension the code lacks (`packages/api-server/src/core/turn-metrics.ts`)
- ✅ **Fixed:** promised `apiServer.otlpEndpoint` helm value absent (`deploy/helm/platform/values.yaml`)
- ✅ **Fixed:** promised UI `_meta` stamping and shared contract vocabulary absent
- ✅ **Fixed:** Tests section named assertions the file lacks (`packages/api-server/src/__tests__/unit/turn-metrics.test.ts`)
- ✅ **Fixed:** doc promised the counter and the log cannot drift apart (`docs/architecture/observability.md:64`)
- ✅ **Fixed:** event-bus saga placed in `core/` against the repo layout (now `packages/api-server/src/sagas/turn-metrics.ts`)
- 🆕 **New:** the doc sends readers to a SQL read path that has no session-turn view (`docs/architecture/observability.md:64`)
- 🆕 **New:** a test-only reset hook is exported from a production module (`packages/api-server/src/core/turn-metrics.ts:20`)
- 🆕 **New:** the metric attribute parameter is an unconstrained `string` (`packages/api-server/src/core/turn-metrics.ts:16`)

### Findings
- 🟡 **Warning:** The doc points at a read path that cannot answer for relay surfaces (`docs/architecture/observability.md:64`).
- 🟡 **Warning:** A test-only reset hook is exported from production code (`packages/api-server/src/core/turn-metrics.ts:20`).
- 🟡 **Warning:** The metric attribute parameter is an unbounded `string` (`packages/api-server/src/core/turn-metrics.ts:16`).

✅ No findings.

### Issue Fit

🔁 The read-path warning is listed above with its inline comment.

🔁 The two `core/turn-metrics.ts` warnings are listed above with their inline comments.

### Verdict
COMMENT — the previous round is fully resolved and the counter itself is correct; three non-blocking seams remain.

---

pilartomas:
All three addressed in `5ee5ed9e`.

**Read path — accepted, and the correction makes the case stronger.** Confirmed: `session_turn` rows are written (`modules/usage/sagas/persist-activity.ts:108`) but no `usage_*` view selects them — `grep -rn session_turn packages/db/drizzle` returns nothing. So the SQL path answers for channel turns and not for relay turns. The page now says exactly that, and that this counter is consequently the only queryable per-surface number for relay surfaces today.

**Test-only reset hook — accepted.** `createTurnMetrics(meter?)` returns the instrument, the module keeps no mutable state, and `resetTurnMetricsForTest` is gone. `bootstrap.ts` constructs it and hands it to the saga; that is safe because `telemetry.ts` runs via `node --import` before the app bundle evaluates, so the SDK is registered by then — the memoization existed only to dodge that ordering, and injection removes the need. Tests build their own instrument on an in-memory meter.

**Unbounded attribute — accepted.** `TurnSurface` is now a closed union (`ui`/`cli`/`other`/`slack`/`telegram`) and the saga maps through `toTurnSurface`. Both event types declare `surface`/`channel` as `string`, so an unrecognized value folds into `other` instead of minting a series per value; there is a test for that fold.

Note for the record: this counter's only end-to-end coverage is unit tests — the e2e suite installs from `values-e2e.yaml`, which does not enable the telemetry backend, so no e2e run exercises an api-server with the SDK active.

packages/api-server/src/core/turn-metrics.ts
```diff
@@ -0,0 +1,36 @@
...
+    ? (raw as TurnSurface)
+    : "other";
+}
+
+export interface TurnMetrics {
+  recordTurn(surface: TurnSurface): void;
+}
+
+export function createTurnMetrics(
+  meter: Meter = metrics.getMeter(SCOPE),
```
dam-code-guardian: 🟡 **Warning:** The meter is resolved when `createTurnMetrics()` runs, and `bootstrap.ts:402` calls it during startup. `@opentelemetry/api` has no proxy meter provider (only `NoopMeterProvider`, unlike the tracer API's `ProxyTracerProvider`), so a meter obtained before `sdk.start()` registers the global provider is bound to a no-op counter permanently — every later `recordTurn` is silently discarded. Production is safe today only because the `--import ./dist/telemetry.js` preload happens to run first; the sibling instrument module resolves its meter lazily on first use instead (`packages/api-server-api/src/trpc-telemetry.ts:20`).

**Fix:** Never capture a meter at module or composition time — resolve it at first record, so registration order cannot silently disable the instrument. Move `metrics.getMeter(SCOPE)` inside `recordTurn` behind a cached instrument, matching `getInstruments()` in `trpc-telemetry.ts`, and keep the injected-meter parameter for tests.

packages/api-server/src/core/turn-metrics.ts
```diff
@@ -0,0 +1,36 @@
+import { metrics, type Meter } from "@opentelemetry/api";
+
+const SCOPE = "platform-apiserver";
+
+export const TURN_SURFACES = [
```
dam-code-guardian: 🟡 **Warning:** `TURN_SURFACES` is a third independent copy of the surface vocabulary, and `toTurnSurface` widens through `as readonly string[]` then narrows back with `raw as TurnSurface`, so neither producer's type takes part in the check. Adding a surface at `clientSurface` (`packages/api-server/src/apps/api-server/admission/auth.ts:207`) or a member to `ChannelTurnRelayed.channel` (`packages/api-server/src/events.ts:97`) folds into `other` with no type error and no test failure.

**Fix:** Derive a subscriber's dimension type from the publishing event's own type — the publisher owns the event shape. Union `ChannelTurnRelayed["channel"]` with a narrowed return type for `clientSurface` instead of restating the values, so a new surface breaks the build at this file.

docs/architecture/observability.md
```diff
@@ -57,7 +57,17 @@ Harnesses produce telemetry by exporting it themselves over OTLP — the platfor
...
+
+### Turn counter
+
+[usage-tracking](usage-tracking.md) stores a row per turn on every surface, but its SQL read path exposes views for channel turns only — relay turns are recorded and not yet readable there. So this counter is both the time-series form of that fact, for a dashboard that wants a rate without SQL and the inspector role, and today the only queryable per-surface number for the surfaces that reach an agent through the relay. It is a subscriber on the same turn events that subsystem persists, so both sinks answer to one definition of a turn rather than each counting their own way. They are not guaranteed to agree in practice: each is separately enabled — the counter needs the SDK active, the log needs activity tracking on — and each fails independently, so treat a divergence as a sink being off or dropping, not as two different notions of a turn.
```
dam-code-guardian: 🟡 **Warning:** Two findings on this paragraph.

1. **The page contradicts its own no-overlap claim.** This section describes a second sink on the same turn events usage-tracking persists, while line 88 still states this subsystem "is distinct from two neighbours and does not overlap them".
   **Fix:** When a change makes a subsystem take on a question a neighbour owns, update that page's relationship section in the same PR — a deliberate parallel sink is stated as such, never left under a blanket no-overlap claim.

2. **"relay turns are recorded and not yet readable there" is a phase marker about a neighbour**, which `docs/guidelines/documentation-guidelines.md:41` excludes from architecture pages, and it rots the moment usage-tracking adds a relay-turn view.
   **Fix:** State the durable division of responsibility and link out to the owning page; keep "not yet" phase markers about another subsystem's current state out of architecture pages.

🟢 Separately, "a dashboard that wants a rate without SQL and the inspector role" parses two ways — the dashboard avoids both, or the inspector role is an audience for this counter. The first reading is the accurate one (`platform-inspector` gates the Postgres view endpoints only, `docs/architecture/usage-tracking.md:13`), so only the wording invites the wrong conclusion.

packages/api-server/src/__tests__/unit/turn-metrics.test.ts
```diff
@@ -0,0 +1,146 @@
...
+} from "../../core/turn-metrics.js";
+import { startTurnMetricsSaga } from "../../sagas/turn-metrics.js";
+
+describe("turn metrics", () => {
+  let reader: PeriodicExportingMetricReader;
+  let meterProvider: MeterProvider;
+  let turns: TurnMetrics;
+  let saga: Subscription | null = null;
+
+  beforeEach(() => {
```
dam-code-guardian, line 25: 🟡 **Warning:** The PR body's Tests list promises a case for `recordTurn` being a no-op "when no meter provider is registered", but no such case exists — all five tests build a real `MeterProvider` here in `beforeEach`. The body also states "This description now matches HEAD", so the claim reads as delivered.

**Fix:** After a rewrite of a PR description, re-check every enumerated claim against the diff and delete each item the diff does not contain. Here, either add the no-op test — it would also cover the eager-meter seam flagged in `core/turn-metrics.ts` — or drop it from the Tests list.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `5ee5ed9e`

## PR #3358: feat(observability): expose agent turns per surface as a metric
**Author:** pilartomas | **Branch:** feat/turn-metrics → main | **Changes:** +238 −2 (5 files)

### Summary
The PR adds an OTel counter `platform.turn.total`, dimensioned by originating surface, fed by an RxJS saga on the existing `SessionTurnRelayed` and `ChannelTurnRelayed` events. All three findings of the previous round are fixed; this round's findings are the doc prose, one body claim, and two typing seams in the instrument module.

### Changes since last review
Previous HEAD: `e29bdc74` (2026-08-18T10:53:38Z) — verdict COMMENT

- ✅ **Fixed:** doc pointed at a SQL read path with no session-turn view (`docs/architecture/observability.md:64`)
- ✅ **Fixed:** test-only reset hook exported from a production module (`packages/api-server/src/core/turn-metrics.ts`)
- ✅ **Fixed:** metric attribute parameter was an unconstrained `string` (`packages/api-server/src/core/turn-metrics.ts:22`)
- 🆕 **New:** the meter is resolved eagerly from the global provider (`packages/api-server/src/core/turn-metrics.ts:26`)
- 🆕 **New:** the surface vocabulary is restated behind casts (`packages/api-server/src/core/turn-metrics.ts:5`)
- 🆕 **New:** the page contradicts its own no-overlap claim (`docs/architecture/observability.md:64`)
- 🆕 **New:** the page carries a phase marker about a neighbour (`docs/architecture/observability.md:64`)
- 🆕 **New:** the body's Tests list names a test the file lacks (`packages/api-server/src/__tests__/unit/turn-metrics.test.ts:25`)

### Findings
- 🟡 **Warning:** the meter is resolved eagerly from the global provider (`packages/api-server/src/core/turn-metrics.ts:26`).
- 🟡 **Warning:** the surface vocabulary is restated as a third list behind casts (`packages/api-server/src/core/turn-metrics.ts:5`).
- 🟡 **Warning:** the page states a parallel sink and also claims no overlap (`docs/architecture/observability.md:64`).
- 🟡 **Warning:** the page records a neighbour's current gap as a phase marker (`docs/architecture/observability.md:64`).
- 🟡 **Warning:** the body's Tests list names a no-op test the file lacks (`packages/api-server/src/__tests__/unit/turn-metrics.test.ts:25`).
- 🟢 **Suggestion:** the phrase "a dashboard that wants a rate without SQL and the inspector role" parses two ways — either the dashboard avoids both, or the inspector role is an audience. The first reading is the correct one, so the words invite the wrong conclusion (`docs/architecture/observability.md:64`).

- 🟡 **Warning:** The page now contradicts itself about subsystem overlap. The new Turn counter section describes a second sink on the same turn events usage-tracking persists, while the unchanged relationship section still asserts this subsystem "does not overlap" that neighbour (`docs/architecture/observability.md:64`, `docs/architecture/observability.md:88`).
  **Fix:** When a change makes a subsystem take on a question a neighbour owns, update that page's relationship section in the same PR — a deliberate parallel sink is stated as such, never left under a blanket no-overlap claim.
- 🟡 **Warning:** The added prose records a neighbouring subsystem's current gap as an implementation phase marker ("relay turns are recorded and not yet readable there"), which `docs/guidelines/documentation-guidelines.md:41` excludes, and which rots the moment usage-tracking adds a relay-turn view (`docs/architecture/observability.md:64`).
  **Fix:** State the durable division of responsibility and link out to the owning page; keep "not yet" phase markers about another subsystem's current state out of architecture pages.
- 🟢 **Suggestion:** The skill also read the sentence as naming the inspector role a consumer of this counter. Verified against `docs/architecture/usage-tracking.md:13` and `:135`: that role gates the Postgres view endpoints only. The sentence's other parse — the dashboard needs neither SQL nor the role — is accurate, so this is the ambiguity above, not a false claim (`docs/architecture/observability.md:64`).

### Issue Fit

- 🟡 **Warning:** The Tests list promises a case for `recordTurn` as a no-op "when no meter provider is registered", but no such case exists — all five tests build a real `MeterProvider` in `beforeEach`. The body also states "This description now matches HEAD", so the claim reads as delivered (`packages/api-server/src/__tests__/unit/turn-metrics.test.ts:25`).
  **Fix:** After a rewrite of a PR description, re-check every enumerated claim against the diff and delete each item the diff does not contain. Here, either add the no-op test or drop it from the Tests list.

🔁 The two `core/turn-metrics.ts` warnings are listed above with their inline comments.

### Verdict
COMMENT — the previous round is fully resolved and the counter itself is correct; five non-blocking seams remain in the instrument's wiring, its surface typing, and the prose.

---

pilartomas:
All five addressed in `62809147` plus a description fix.

**Eager global meter — accepted.** `createTurnMetrics(meter: Meter)` now requires the meter; the default argument that reached for the global provider is gone, so nothing can construct a counter before the SDK is registered. `bootstrap.ts` resolves the meter at the call site.

**Third vocabulary list behind casts — accepted.** One `Map<string, TurnSurface>` is the single list and `toTurnSurface` is a lookup with an `?? "other"` fallback, so there are no `as` casts in the module at all.

**No-overlap claim beside a parallel sink — accepted, good catch.** The relationship section asserted this subsystem "does not overlap" usage-tracking while the new section describes a second sink on that subsystem's events. It now states the overlap as deliberate and says the split is by read pattern — durable record and SQL surface there, operational time series here — with both fed from the same events.

**Phase marker about a neighbour — accepted.** "relay turns are recorded and not yet readable there" is gone. The prose states the division of responsibility and links out, so it does not rot when a relay-turn view lands.

**Tests list named a no-op test — accepted, and the same class as round one.** I removed that test along with the reset hook it exercised and did not re-check the list I had just rewritten. The body now names the assertion that replaced it (an unrecognized surface folding into `other`). Every item in that list is in the diff at HEAD.

**Ambiguous sentence (suggestion) — taken.** Rewritten so the reading is unambiguous: reachable "without SQL and without that subsystem's reader role", rather than a phrase that could name the role as an audience.

PR #3359, opened 2026-08-18
fix(skills): report a skill source path that resolves to nothing

### LINKED ISSUE #3321: Skill source with a non-existent path silently reports "No skills in this source"
A skill source configured with a `path` that does not exist in the repo reports **"No skills in this source."** That message is indistinguishable from a repo that genuinely contains no skills, so there is nothing to tell you the path is the problem.

## Repro

1. Add a skill source with git URL `https://github.com/apocohq/skills` (14 skills, all under `skills/`).
2. Set the **path** field to something that isn't a directory in that repo — e.g. `aposkills`.
3. The source lists no skills. No error, no warning in the api-server log.

## Root cause

[`packages/api-server/src/modules/skills/infrastructure/public-archive-scanner.ts`](packages/api-server/src/modules/skills/infrastructure/public-archive-scanner.ts)

`findSkillDirs` scans only the given path when one is set:

```js
if (subPath) {
  if (subPathEscapes(subPath)) throw new Error(`skill source path rejected: ${subPath}`);
  return skillDirsUnder(repoDir, path.join(repoDir, subPath));
}
```

and `skillDirsUnder` swallows a failed `readdir` and returns an empty array:

```js
try {
  entries = await fs.readdir(root, { withFileTypes: true });
} catch {
  return [];
}
```

So "directory does not exist" and "directory exists but holds no skill dirs" collapse into the same empty result, and both render as "No skills in this source."

The traversal is also one level deep — it looks for `<path>/<entry>/SKILL.md` — so a path pointing at a directory that exists but nests skills deeper produces the same silent empty result.

## Expected

Fail loudly and name the path, distinguishing at minimum:

- the path does not exist in the repo at that revision
- the path exists but contains no directory with a `SKILL.md`

The message should reach the UI, since that is where the mistake is made. Including the resolved revision would help, as the answer can differ between commits.

## Explicitly not wanted

**No fallback to the default source roots** (`skills`, `.claude/skills`, `.agents/skills`) when an explicit path yields nothing. An explicit path should be honoured exactly as given; a wrong one is a user error to report, not to paper over by silently scanning somewhere else.

## Impact

Cost real debugging time on a live install: two sources were configured with the path field filled in as though it were a label, and both silently listed nothing while the repos were fine. The same sources scan correctly with the path cleared.

PetrBulanek:
A skill source can carry an optional path — a subdirectory of the repo to read skills from. When that path is wrong, the source lists nothing and says "No skills in this source.", which is the same sentence a genuinely empty repo produces. Nothing tells you the path is the problem. This cost real debugging time on a live install, where two sources had the path field filled in as though it were a label.

A scan of a source with a path that yields nothing now fails with a named error that says which of two things happened: the path is not in the repository, or the path is there but holds no skill. Both name the path and the commit they were resolved against, and both reach the source card where the mistake is visible.

An explicit path is still honoured exactly as given — there is no silent fall back to the usual skill locations when it fails.

The Path field in the add-source form also gains the one-line explanation it was missing, since that absence is what invites the mistake.

Closes #3321

## Sub-issues

- [x] 01 — Named path verdicts on the public archive scan
- [x] 02 — The pod scan reports the same verdicts


dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `74dbe4d0`

## PR #3359: fix(skills): report a skill source path that resolves to nothing
**Author:** PetrBulanek | **Branch:** `fix/3321-skill-source-path-verdicts` → `main` | **Changes:** +338 −19 (15 files)

### Summary
A skill source's optional `path` that resolves to nothing now fails with a named verdict — `path-missing` or `path-empty` — instead of rendering as "No skills in this source." Both scanners (public archive, pod clone) report it, and the verdict travels structurally to the source card.

### Findings
- 🟡 **Warning:** `COPY.source_path_not_found.detail` and `COPY.source_path_empty.detail` are unreachable (`packages/api-server/src/modules/skills/domain/scan-failure.ts:27`) — see the inline comment.
- 🟢 **Suggestion:** The pod clone branch drops the resolved commit. `scanGitClone` calls `sourcePathError(...)` with no `version`, so a clone-sourced verdict reads "in the repository" while the tarball path reads "at commit 74dbe4d". `cloneShallow` has the SHA available in the temp dir (`packages/agent-runtime/src/modules/skills/services/scan.ts:109`).
- 🟢 **Suggestion:** `isMissingDir` and `skillDirsIn` are byte-identical in both scanners (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:458` and `:505`; `packages/api-server/src/modules/skills/infrastructure/public-archive-scanner.ts:108` and `:159`). `subPathEscapes` was already duplicated on `main`, so this is the third and fourth copy of the same pair.
- ✅ **Looks good:** the issue's "no fallback to the default roots on an explicit path" constraint holds on both branches, and the no-path branch still returns `{ kind: "found" }` for a genuinely empty repo — so silence stays silence exactly where it should.

✅ No findings.

### Issue Fit
- 🟡 **Warning:** Two architecture-doc sentences lose detail this change does not affect: the `data.scanFailure` lift by the api-server error formatter, and the `ScannedSkill` field list (`docs/architecture/skills.md:259`, `docs/architecture/skills.md:164`). The page sits 7 characters under the 40,000-char cap at HEAD, which likely explains the compression.
  **Fix:** Restore the two clauses if the cap allows, or note the page is at its cap so the next edit knows why detail was traded.

✅ No findings.

✅ No findings.

### Verdict
COMMENT — the mechanism is correct end to end; the two 🟡 are cosmetic-copy and doc-detail issues, neither affecting behavior.

---

packages/api-server/src/modules/skills/domain/scan-failure.ts
```diff
@@ -23,6 +24,19 @@ const COPY: Record<ScanFailureCode, Omit<ScanFailure, "code">> = {
     detail:
       "The sandbox couldn't be reached to scan this source. Try re-scanning in a moment.",
   },
+  source_path_not_found: {
```
dam-code-guardian, line 27: 🟡 **Warning:** Both new `COPY` entries carry a `detail` that nothing can render. `sourcePathFailure` is the only producer of `source_path_not_found` and `source_path_empty`, and it always passes `{ detail }`; `scanFailure` spreads the override after `COPY[code]`, so the entry's own `detail` is replaced on every path. The `title` is live — only `detail` is dead.

**Fix:** When an override always supplies a field, the registry entry should not also define it — either omit `detail` here and let `sourcePathFailure` own the whole sentence, or move the shared trailing advice into `COPY` and have the override extend it.

packages/agent-runtime/src/modules/skills/services/scan.ts
```diff
@@ -89,9 +104,12 @@ async function scanGitClone(
     const cloned = await deps.git.cloneShallow(source, tmp, 50);
     if (!cloned.ok) return cloned;
 
-    const skillDirs = await deps.repo.findSkillDirsInClone(tmp, subPath);
+    const outcome = await deps.repo.findSkillDirsInClone(tmp, subPath);
+    if (outcome.kind !== "found") {
+      return sourcePathError(outcome.kind, source, outcome.subPath);
```
dam-code-guardian, line 109: 🟢 **Suggestion:** The two scanners disagree on whether the verdict names a commit. The tarball path threads `version` through `findSkillDirs`, so it renders "at commit 74dbe4d"; this clone path calls `sourcePathError` with three arguments, so `version` stays `undefined` and the same mistake reads "in the repository". The issue asked for the resolved revision because the answer can differ between commits. `collectSkills` already passes it, so only the clone branch is affected.

PetrBulanek:
## Review response

**🟡 `COPY.source_path_not_found.detail` / `source_path_empty.detail` unreachable — confirmed, fixed in a82007b89.**

Traced every producer: `sourcePathFailure` is the only one that emits either code, it always passes `{ detail }`, and `scanFailure` spreads the override after `COPY[code]`. Both strings were dead, and `title` was live, exactly as reported.

The shared trailing advice now stays in the registry and `sourcePathFailure` prepends only the path-specific verdict to it. Every registry field is live again, and the rendered sentence is byte-identical, so no copy or test churn.

**🟡 `docs/architecture/skills.md` dropped unrelated mechanism detail — declining.**

The cap reading is right: the page is 39,993 chars against the 40,000 cap in `scripts/doc-size.mjs`, so 7 chars of headroom. (Measured as the script measures — JS `.length`; `wc -c` reads 40,222 because the page is full of em dashes.)

The fix does not follow from it, on two counts.

*"Restore the two clauses if the cap allows"* — it does not. The two clauses cost roughly 120 chars against 7 of headroom.

They also should not come back. Both fragments are what `docs/guidelines/documentation-guidelines.md` names under **Omit** as "code-level shape — type signatures, field names, function arguments, enum members". `(name, description, version, contentHash, dir)` is a field list; `cause` and `data.scanFailure` are wire field paths. Raising them to "a scanned skill" and "carried structurally on the error" moves the page toward the content policy, not away from it — the cap did not force a trade here, it surfaced content that was already pitched too low.

The meaning survives in both places. The next sentence still names `dir` ("Reporting `dir` is what lets a private source's preview…"), and the error paragraph still names `scanFailure` ("the latter arrives with no `scanFailure` at all").

One thing the review missed, worth its own change: line 259 still spells `a ScanFailure of { code, title, detail }` — the same class of field list that was just raised elsewhere on the page. That is the sentence to lift for consistency, and it buys back headroom.

**🟢 Clone branch drops the resolved commit — valid, deferring.**

One correction to the premise: `cloneShallow` returns `Result<void, SkillsDomainError>`, so the SHA is not in hand. The cheap route is `lastTouchingSha(tmp, ".")`, which reuses the existing port rather than widening it.

**🟢 `isMissingDir` / `skillDirsIn` duplicated — valid, deferring.**

Byte-identical, confirmed. Sharing them crosses the `agent-runtime` / `api-server` package boundary, so it needs a home package first. Separate change.

packages/ui/src/modules/sandboxes/components/skills/github-source-tab.tsx
```diff
@@ -52,6 +52,10 @@ export function GithubSourceTab({
             placeholder="skills/"
             {...register("path")}
           />
+          <p className="text-sm text-muted-foreground">
```
dam-code-guardian, line 55: 🟢 **Suggestion:** The new paragraph is not linked to the input, and `SectionLabel` renders a `<span>`, so assistive tech announces an unnamed textbox without this guidance. The two sibling fields above share the pattern, so the dialog-wide fix (an `id` + `<label htmlFor>` + `aria-describedby` per field) is probably its own change rather than this one.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `a82007b8`

## PR #3359: fix(skills): report a skill source path that resolves to nothing
**Author:** PetrBulanek | **Branch:** `fix/3321-skill-source-path-verdicts` → `main` | **Changes:** +332 −19 (15 files)

### Summary
One commit since the last review returns the shared trailing advice to the `COPY` registry, so both new entries are read again. The rendered sentence stays byte-identical on both branches, and every finding from the previous round is now resolved.
_(Suppressed 1 finding per PR context: the `docs/architecture/skills.md` detail warning — the decline holds, see below.)_

### Changes since last review
Previous HEAD: `74dbe4d0` (2026-08-18T14:07:30Z) — verdict COMMENT

- ✅ **Fixed:** `COPY.source_path_not_found.detail` and `source_path_empty.detail` are live again; `sourcePathFailure` prepends only the path-specific verdict (`packages/api-server/src/modules/skills/domain/scan-failure.ts:64`). Both rendered sentences are unchanged.
- ✅ **Resolved — decline holds:** the `docs/architecture/skills.md` detail warning. Verified: `docs/guidelines/documentation-guidelines.md:41` lists field names and type signatures under **Omit**, and the page measures 39,993 of the 40,000 chars `scripts/doc-size.mjs` enforces.
- 🔁 **Still present (deferred):** the clone branch's verdict omits the resolved commit (`packages/agent-runtime/src/modules/skills/services/scan.ts:109`).
- 🔁 **Still present (deferred):** `isMissingDir` and `skillDirsIn` are duplicated across both scanners (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:458`, `packages/api-server/src/modules/skills/infrastructure/public-archive-scanner.ts:108`).

### Findings
- 🟢 **Suggestion:** The new helper paragraph is not associated with the Path input, so screen readers announce the textbox without it (`packages/ui/src/modules/sandboxes/components/skills/github-source-tab.tsx:55`) — see the inline comment. The two sibling fields share the pattern, so this reads as a separate issue for the dialog rather than something this PR introduced.
- 🟢 **Suggestion:** The clone branch's verdict names no commit. `scanGitClone` calls `sourcePathError` with three arguments, so a clone-sourced message reads "in the repository" where the tarball path reads "at commit a82007b" (`packages/agent-runtime/src/modules/skills/services/scan.ts:109`). Reported by issue-fit as an issue-scope gap. Deferred, with `lastTouchingSha(tmp, ".")` named as the cheap route.
- 🟢 **Suggestion:** The path-verdict decision now exists twice — `isMissingDir`, `skillDirsIn`, and the missing/empty/found branch in both scanners, so the two scan routes can drift on the same source's verdict (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:458`, `packages/api-server/src/modules/skills/infrastructure/public-archive-scanner.ts:108`). Reported by typescript-engineering. Deferred: sharing them crosses the package boundary.

✅ No findings.

### Issue Fit
Both declared sub-issues land: the public-archive scan and the pod scan raise the same two named verdicts, and the issue's "no fallback to the default roots on an explicit path" constraint holds on both branches. The one gap is the missing revision on the clone branch, merged into `### Findings` above.

The duplicated verdict logic is merged into `### Findings` above. Three further observations did not survive verification and are recorded rather than reported: the `agent-runtime-api` root already imports from `./modules/` in `index.ts`, `context.ts`, and `router.ts`, so `trpc.ts` adds no new dependency direction; `Result` in `agent-runtime` versus `throw` in `api-server` is each package's own existing convention; and `AgentRuntimeSourcePathError` omitting `path` is unreachable, since the runtime raises the verdict only when `src.path` supplied the subpath.

The label association gap is merged into `### Findings` above.

### Verdict
APPROVE — the previous round's warning is fixed at its cause, the declined one holds on verification, and the two remaining items are acknowledged suggestions that do not block.

---

Issue #3263, opened 2026-08-10
Improve DAM experiment skill invocation


jenna-winkler:
# Improve DAM experiment skill invocation

## Problem

The `dam-experiment` skill is not being reliably invoked when users ask agents to perform experiment-related tasks in an experiment sandbox.

Today, the agent appears to rely too heavily on the user explicitly saying **“DAM experiment”** to recognize that the experiment skill should be used. For example, asking the agent to “create an experiment” may cause it to use a more generic approach instead of the DAM Experiment SDK and documentation.

This creates unnecessary friction and defeats the purpose of having an experiment-specific sandbox. Users are already operating within a DAM experiment sandbox, so they should not need to repeatedly specify that they are using DAM.

## Expected behavior

When an agent is running in an **experiment sandbox**, experiment-related requests should naturally invoke the `dam-experiment` skill.

For example:

* “Create an experiment that evaluates these models.”
* “Set up an experiment to compare these prompts.”
* “Run an experiment using this dataset.”
* “I want to test these approaches and compare the results.”

These should use the DAM Experiment SDK and the relevant DAM documentation without requiring the user to say “DAM experiment.”

The sandbox context should be treated as a strong signal that **“experiment” means a DAM experiment**.

## Scope

Update the `dam-experiment` skill's invocation guidance / description / prompt engineering so that it is more reliably selected for experiment-related requests in experiment sandboxes.

The skill should:

* Be invoked for generic experiment requests when running in an experiment sandbox.
* Prefer the DAM Experiment SDK over generic approaches for experiment setup and execution.
* Direct the agent to the relevant DAM documentation when needed.
* Avoid requiring users to explicitly mention “DAM” or “DAM experiment.”
* Preserve the existing behavior where explicitly saying “DAM experiment” also invokes the skill.

## Example

**Current:**

> User: “Create an experiment to compare these approaches.”

Agent may interpret this as a generic experiment and not use the DAM skill.

**Desired:**

> User: “Create an experiment to compare these approaches.”

Agent recognizes the sandbox context and uses the `dam-experiment` skill / DAM Experiment SDK.

## Acceptance criteria

* A user in a newly created or existing experiment sandbox can say “create an experiment” without mentioning DAM and have the `dam-experiment` skill invoked.
* Common experiment-related phrasing (“run an experiment,” “set up an experiment,” “compare these approaches,” etc.) reliably routes to the skill.
* Explicitly saying “DAM experiment” continues to work.
[...]

PR #3362, opened 2026-08-18
fix(experiments): invoke dam-experiment skill on generic experiment asks

tomkis:
Refs #3263 (the existing-sandboxes acceptance criterion stays open, see Out of scope)

## Problem

The `dam-experiment` skill only kicked in reliably when the user said "DAM experiment". A plain "create an experiment" often got a generic script instead of the Experiment SDK, even inside an experiment sandbox.

## Changes

1. **Skill description rewrite** — leads with trigger phrasings (create/plan/set up/run an experiment, compare models/prompts/approaches, benchmark, optimize) and never mentions DAM. Nothing left to disambiguate: in this sandbox, every experiment is a platform Experiment. Explicit "DAM experiment" phrasing still matches trivially.

2. **Sandbox purpose in AGENTS.md** — the experiment Install Command now appends a short purpose note to `~/.agents/AGENTS.md`, so every session (not only the greeted first one) opens knowing what this sandbox is for. Claude Code reads it via a `~/.claude/CLAUDE.md` symlink, mirroring the image's `/etc/AGENTS.md` pattern at $HOME level.

3. **Shared primitive** — `buildAppendAgentsMdCommand` in the agents module, exported for any kind to reuse. Append-only, so multiple writers compose instead of clobbering. Single-line output regardless of content newlines, single quotes escaped, symlink creation guarded so it never overwrites an existing file.

The skill copy stays last in the install command, so the greeting's "skill present" gate still implies everything landed.

## Out of scope

Existing sandboxes keep their stale skill copy (install runs once into persistent $HOME). A re-sync for those is a separate discussion.

## Testing

- Unit tests for the primitive (append semantics, ordering, escaping, guarded symlink) and install-command ordering.
- Ran the generated command twice against a fake $HOME: sections compose, symlink resolves, content readable via CLAUDE.md.
- `api-server:check`, `docs:check`, `common:check`, comment-types all green.

### DIFF
--- a/packages/api-server/src/modules/agents/domain/agents-md.ts
+++ b/packages/api-server/src/modules/agents/domain/agents-md.ts
@@ -1,0 +1,10 @@
+export function buildAppendAgentsMdCommand(section: string): string {
+  const lines = [...section.split("\n"), ""].map(
+    (line) => `'${line.replaceAll("'", "'\\''")}'`,
+  );
+  return [
+    'mkdir -p "$HOME/.agents" "$HOME/.claude"',
+    `printf '%s\\n' ${lines.join(" ")} >> "$HOME/.agents/AGENTS.md"`,
+    '[ -e "$HOME/.claude/CLAUDE.md" ] || ln -s "$HOME/.agents/AGENTS.md" "$HOME/.claude/CLAUDE.md"',
+  ].join("; ");
+}
--- a/packages/api-server/src/modules/agents/domain/agents-md.ts
+++ b/packages/api-server/src/modules/agents/domain/agents-md.ts
@@ -1,10 +1,14 @@
 export function buildAppendAgentsMdCommand(section: string): string {
-  const lines = [...section.split("\n"), ""].map(
-    (line) => `'${line.replaceAll("'", "'\\''")}'`,
-  );
+  const quote = (line: string) => `'${line.replaceAll("'", "'\\''")}'`;
+  const rawLines = section.split("\n");
+  const lines = [...rawLines, ""].map(quote);
+  const marker = rawLines.find((line) => line.trim() !== "");
+  const append = `printf '%s\\n' ${lines.join(" ")} >> "$HOME/.agents/AGENTS.md"`;
   return [
     'mkdir -p "$HOME/.agents" "$HOME/.claude"',
-    `printf '%s\\n' ${lines.join(" ")} >> "$HOME/.agents/AGENTS.md"`,
-    '[ -e "$HOME/.claude/CLAUDE.md" ] || ln -s "$HOME/.agents/AGENTS.md" "$HOME/.claude/CLAUDE.md"',
+    marker === undefined
+      ? append
+      : `grep -qsxF ${quote(marker)} "$HOME/.agents/AGENTS.md" || ${append}`,
+    '[ -e "$HOME/.claude/CLAUDE.md" ] || [ -L "$HOME/.claude/CLAUDE.md" ] || ln -s "$HOME/.agents/AGENTS.md" "$HOME/.claude/CLAUDE.md"',
   ].join("; ");
 }
--- a/packages/api-server/src/modules/experiments/domain/install-command.ts
+++ b/packages/api-server/src/modules/experiments/domain/install-command.ts
@@ -1,11 +1,22 @@
 import { STAGED_SKILLS_DIR } from "agent-runtime-api";
 import { EXPERIMENT_SKILL_NAME } from "api-server-api";
+import { buildAppendAgentsMdCommand } from "../../agents/index.js";
+
+const SANDBOX_PURPOSE = [
+  "## Sandbox purpose",
+  "",
+  "This sandbox exists to author and run platform Experiments. Treat every",
+  "experiment-shaped request (create, plan, set up, or run an experiment,",
+  "compare models/prompts/approaches, benchmark, optimize) as a platform",
+  `Experiment and invoke the \`${EXPERIMENT_SKILL_NAME}\` skill first.`,
+].join("\n");
 
 export function buildExperimentInstallCommand(): string {
   return [
     "set -eu",
     'mkdir -p "$HOME/.claude/commands" "$HOME/.agents/skills"',
     `cp -R "${STAGED_SKILLS_DIR}/commands/." "$HOME/.claude/commands/"`,
+    buildAppendAgentsMdCommand(SANDBOX_PURPOSE),
     `cp -R "${STAGED_SKILLS_DIR}/${EXPERIMENT_SKILL_NAME}" "$HOME/.agents/skills/"`,
   ].join("; ");
 }
--- a/docs/architecture/experiments.md
+++ b/docs/architecture/experiments.md
@@ -1,6 +1,6 @@
 # Experiments
 
-Last verified: 2026-08-17
+Last verified: 2026-08-18
 
 ## Overview
 
@@ -44,7 +44,14 @@ destination has nothing to show and the user has nothing to click.
 So creating one is a first-class flow: an **experiment sandbox** is an Agent
 carrying the `experiment` [Agent Kind](knowledge-bases.md) whose Install Command
 copies the `dam-experiment` authoring skill and an `/experiment-onboard` command
-out of a path staged in the image. It rides the same kinded-create rail as a
+out of a path staged in the image, and appends a sandbox-purpose note to
+`~/.agents/AGENTS.md` so every session — not only the greeted first one — opens
+knowing that "experiment" means a platform Experiment. AGENTS.md is the source
+of truth and a symlink makes Claude Code read it, mirroring the image's
+`/etc/AGENTS.md` pattern at `$HOME` level. The note rides the shared append-only
+primitive (`buildAppendAgentsMdCommand` in the agents module), so later writers
+compose instead of clobbering each other.
+It rides the same kinded-create rail as a
 Knowledge Base, differing only in the marker and the command; nothing is fetched
 over the network, because the kit ships with the image. The skill used to be baked
 into every Claude Code sandbox's seeded workspace — moving it behind the marker is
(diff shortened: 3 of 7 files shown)


dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `0bd7c60`

## PR #3362: fix(experiments): invoke dam-experiment skill on generic experiment asks
**Author:** tomkis | **Branch:** fix/experiment-skill-invocation → main | **Changes:** +78 −5 (7 files)

### Summary
The skill description now leads with trigger phrasings, and the experiment Install Command appends a sandbox-purpose note to `~/.agents/AGENTS.md` through a new shared append primitive. The mechanism works on a clean `$HOME` — I ran the generated command and the note composes, the symlink resolves, and the skill lands last as intended.

### Findings
- 🟡 **Warning:** `[ -e ]` guard has two cases where the purpose note does not reach Claude Code (`packages/api-server/src/modules/agents/domain/agents-md.ts:8`)
- 🟡 **Warning:** the append replays on a retried install (`packages/api-server/src/modules/agents/domain/agents-md.ts:7`)
- 🟡 **Warning:** `Closes #3263` against the declared out-of-scope (`packages/api-server/src/modules/experiments/domain/install-command.ts:19`)
- 🟢 **Suggestion:** `install-command.ts` is the first `domain/` file in api-server to import across a module boundary; the other 23 cross-module index imports sit in `services/`, `infrastructure/`, or `compose.ts`. Both sides are pure string builders, so nothing breaks (`packages/api-server/src/modules/experiments/domain/install-command.ts:3`)
- 🟢 **Suggestion:** the added paragraph names `buildAppendAgentsMdCommand` and two exact `AGENTS.md` paths, which `docs/guidelines/documentation-guidelines.md:41` lists under Omit. Baseline for context: 5 of 18 architecture pages already carry a backticked code symbol, so this is not unprecedented (`docs/architecture/experiments.md:47`)
- ✅ **Looks good:** the shell text is safe under `set -eu` on the happy path — single-quote escaping, `%`/backslash literals under `printf '%s\n'`, single-line output, `mkdir` before the write, and an append that composes with a pre-existing `AGENTS.md` section. All five verified by running the generated command against a fake `$HOME`.

The doc-altitude finding is merged into `### Findings` above (`docs/architecture/experiments.md:47`), at 🟢 after the baseline measurement.

### Issue Fit
The `Closes #3263` scope gap is merged into `### Findings` above (`install-command.ts:19`). Reported by the skill as 🔴; recorded at 🟡 because the consequence is issue bookkeeping and the PR body states the scope decision openly.

Both guard findings and the cross-module import are merged into `### Findings` above. No further findings in the routed files.

### Verdict
COMMENT — the mechanism is correct on a clean `$HOME`; three seams sit in the guard, the retry path, and the closing keyword.

---

packages/api-server/src/modules/agents/domain/agents-md.ts
```diff
@@ -0,0 +1,10 @@
+export function buildAppendAgentsMdCommand(section: string): string {
+  const lines = [...section.split("\n"), ""].map(
+    (line) => `'${line.replaceAll("'", "'\\''")}'`,
+  );
+  return [
+    'mkdir -p "$HOME/.agents" "$HOME/.claude"',
+    `printf '%s\\n' ${lines.join(" ")} >> "$HOME/.agents/AGENTS.md"`,
+    '[ -e "$HOME/.claude/CLAUDE.md" ] || ln -s "$HOME/.agents/AGENTS.md" "$HOME/.claude/CLAUDE.md"',
```
dam-code-guardian: 🟡 **Warning:** `[ -e ]` follows the symlink, so two cases leave the purpose note unread by Claude Code. A `~/.claude/CLAUDE.md` pointing at a missing target tests false, and the `ln -s` then fails `File exists` — under the command's `set -eu` that aborts before the skill copy, so the greeting's skill gate never opens (exit 1, reproduced in a fake `$HOME`). A pre-existing regular `CLAUDE.md` takes the other branch: the install succeeds, the note lands in `AGENTS.md`, and Claude Code never reads it.

**Fix:** Test the link itself rather than its target, or pass `ln -sfn`.

```suggestion
    '[ -e "$HOME/.claude/CLAUDE.md" ] || [ -L "$HOME/.claude/CLAUDE.md" ] || ln -s "$HOME/.agents/AGENTS.md" "$HOME/.claude/CLAUDE.md"',
```

packages/api-server/src/modules/agents/domain/agents-md.ts
```diff
@@ -0,0 +1,10 @@
+export function buildAppendAgentsMdCommand(section: string): string {
+  const lines = [...section.split("\n"), ""].map(
+    (line) => `'${line.replaceAll("'", "'\\''")}'`,
+  );
+  return [
+    'mkdir -p "$HOME/.agents" "$HOME/.claude"',
+    `printf '%s\\n' ${lines.join(" ")} >> "$HOME/.agents/AGENTS.md"`,
```
dam-code-guardian: 🟡 **Warning:** The append is not idempotent and the install command can run more than once. `processEvents` settles an event only after a clean exit (`event-loop.ts:32-44`) and `workspace-command-plugin.ts:47` writes its sentinel only after `run` resolves, so any failure at the later skill copy leaves the event unsettled and the next delivery re-appends the section. Verified: two runs of the generated command produce two `## Sandbox purpose` blocks.

**Fix:** A `grep -q` on a stable marker line before the `printf` keeps a replay a no-op; moving the append after the skill copy also closes it, at the cost of the ordering the new test asserts.

packages/api-server/src/modules/experiments/domain/install-command.ts
```diff
@@ -1,11 +1,22 @@
...
+  "compare models/prompts/approaches, benchmark, optimize) as a platform",
+  `Experiment and invoke the \`${EXPERIMENT_SKILL_NAME}\` skill first.`,
+].join("\n");
 
 export function buildExperimentInstallCommand(): string {
   return [
     "set -eu",
     'mkdir -p "$HOME/.claude/commands" "$HOME/.agents/skills"',
     `cp -R "${STAGED_SKILLS_DIR}/commands/." "$HOME/.claude/commands/"`,
+    buildAppendAgentsMdCommand(SANDBOX_PURPOSE),
```
dam-code-guardian, line 19: 🟡 **Warning:** `Closes #3263` will close the issue, whose first acceptance criterion names a "newly created **or existing**" sandbox and whose fourth names "multiple sessions in the same sandbox". The install command is a one-shot `workspace-command` into the persistent `$HOME`, so existing sandboxes get neither the rewritten description nor the note — which the PR body states openly as out of scope.

**Fix:** A plain `Refs #3263` leaves the existing-sandbox criterion tracked; the alternative is a follow-up issue for the re-sync path referenced from here.

packages/api-server/src/modules/agents/domain/agents-md.ts
tomkis: Fixed in fea6b089. I went with the [ -L ] check instead of ln -sfn so a real file a user put there is never clobbered. Verified the dangling-link and regular-file cases against a fake $HOME, both exit 0.

packages/api-server/src/modules/agents/domain/agents-md.ts
tomkis: Fixed in fea6b089. The append is now guarded by grep -qsxF on the section's first non-empty line, so a replayed install is a no-op. Verified two runs produce one block.

packages/api-server/src/modules/experiments/domain/install-command.ts
tomkis: Changed the PR body to Refs #3263 so the existing-sandboxes criterion stays tracked on the issue.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `fea6b08`

## PR #3362: fix(experiments): invoke dam-experiment skill on generic experiment asks
**Author:** tomkis | **Branch:** fix/experiment-skill-invocation → main | **Changes:** +95 −5 (7 files)

### Summary
One commit since the last review hardens the AGENTS.md append: a `[ -L ]` clause on the symlink guard and a heading-keyed `grep` guard before the append. I ran the generated command against fake `$HOME`s in five states — all three findings from the previous round are resolved.

### Changes since last review
Previous HEAD: `0bd7c60` (2026-08-18T14:51:10Z) — verdict COMMENT

- ✅ **Fixed:** a dangling `~/.claude/CLAUDE.md` no longer aborts the install — exit 0 and the skill copy step is reached (`packages/api-server/src/modules/agents/domain/agents-md.ts:12`)
- ✅ **Fixed:** the append no longer replays — two runs leave one section, and two writers with different headings compose (`packages/api-server/src/modules/agents/domain/agents-md.ts:11`)
- ✅ **Fixed:** `Closes #3263` is now `Refs #3263`, matching the declared out-of-scope
- ✅ **Fixed:** the architecture paragraph dropped the code symbol and the exact paths; its new replay claim matches the code
- 🔁 **Still present:** with a pre-existing regular `~/.claude/CLAUDE.md` the note reaches AGENTS.md but not Claude Code — recorded at 🟢 now, since not clobbering a user's own file is the deliberate trade (`packages/api-server/src/modules/agents/domain/agents-md.ts:12`)

### Findings
- 🟢 **Suggestion:** the guard keys on the section's first non-empty line, so identity is prose. Two consequences: a reworded `SANDBOX_PURPOSE` is skipped on a pod that already ran the install, and a second caller opening with the same heading suppresses this one. Neither is reachable today — install runs once per sandbox and there is one caller — so this is a note on the primitive's contract as it gains users (`packages/api-server/src/modules/agents/domain/agents-md.ts:11`)

✅ No findings.

### Issue Fit
✅ No findings.

The guard-identity finding is merged into `### Findings` above. Reported by the skill as two 🟡; recorded as one 🟢 because no caller can reach either case at this HEAD. 🔁 3 findings from the previous review still present or unchanged (see review at `0bd7c60`) — the CLAUDE.md visibility case and the two cross-module import notes, on files this commit did not touch.

### Verdict
APPROVE — the commit closes every finding from the previous round at its cause, and nothing new blocks.

---

PR #3352, opened 2026-08-17
feat(ui): one setup page per feature, replacing the creation wizard

kapetr:
Each of the three things a user can create gets its own setup page — coding agent, experiment, knowledge base — asking only what that thing needs, with the sensible choice already made. A user creating a knowledge base is never shown harness images; a user creating a coding agent picks an image and goes.

Today all three funnel through one three-step wizard that opens by asking which of five starting points you want, so the entry cards on Home and the nav destinations all land in the same place and the user re-answers a question their click already answered.

Size and network access leave creation: a new sandbox takes the template size and the trusted egress preset, both editable in its settings straight after.

## Beyond the issue's scope

**The specialized templates leave the create flows.** The setup pages offer harnesses and a custom image address; the `preconfigured` bucket the wizard carried is gone, so `k-search`, `nous`, `openevolve`, `shinkaevolve`, `gepa`, `adaevolve` and `evox` are no longer pickable from a page. This is deliberate — each setup page asks only what its own thing needs, and a list of research images belongs to none of them. What goes out of reach is the **template**, not the image: each one carries a 5–10Gi storage size a bare address will not reproduce, and `adaevolve` and `evox` are the same `skydiscover` image separated only by the `SKYDISCOVER_SEARCH` their templates set — paste that address and you get the default search with nothing to tell you. They stay enabled in the chart. Re-homing them is #3368.

Closes #3234

## Sub-issues

- [x] 01 — Setup-page scaffolding, experiment and knowledge base
- [x] 02 — Coding agent setup page
- [x] 03 — Retire the wizard

Builds on #3348, which added the entry cards, the Coding agents page and the nav this hooks into. Rebased onto main now that it has merged.

PR #3363, opened 2026-08-18
feat(ui): prefill sandbox names as <type>-<n>

### LINKED ISSUE #3116: UI - Auto-generated names for sandboxes, experiments, and knowledge bases are random and meaningless
## Context

Every create path pre-fills a name from a random adjective + noun pair — "Misty Meadow", "Rustic Glacier", "Swift Otter". Sandboxes, agents, experiments, and knowledge bases all draw from the same list, so the name a user ends up with carries no information about what they just made. Raised by design after seeing the pattern across surfaces.

## Problem

The pre-filled name is friendly but tells you nothing, and most people keep it — it looks like a decision that's already been made for them. A few turns later:

- **Lists are unscannable.** A sandboxes or experiments list reads as a menagerie. You can't tell which one has the repo you're working in, which one is the Slack agent, or which one you spun up to try something 20 minutes ago.
- **The name doesn't survive the thing it names.** Nothing about "Rustic Glacier" links back to its template, image, purpose, or when it was created — the one signal a name could cheaply carry is thrown away.
- **It reads as unserious.** Same problem as the random status strings in #3102: playful copy in a spot where users expected information.
- **It's inconsistent across object types.** Experiments and knowledge bases are more specific things than a bare sandbox, yet they get names from the identical animal-and-weather pool.

Collisions are also possible — 576 combinations picked at random means duplicates show up sooner than people expect.

## Goal

Whatever name a user lands with by default tells them something true about the object — enough that a list of ten is scannable without opening any of them. Users who want to name things themselves aren't slowed down; users who don't get a sensible default instead of a random one.

## User Stories

- As a user, I want a default name that reflects what I just created so that I can find it again in a list without opening it.
- As a user, I want to scan a list of sandboxes, experiments, or knowledge bases and tell them apart at a glance so that I can get back to the right one quickly.
- As a user, I want to rename or replace the default easily so that the suggestion never gets in my way.
- As a user, I want to not end up with two objects sharing a name so that I'm not guessing which is which.

## Scope

**In:** the default naming logic for sandboxes/agents, experiments, and knowledge bases, and how that default is presented at create time (pre-filled value vs. placeholder vs. named-later).

**Out:** renaming existing objects, and any change to session/chat titles — that's #1082.

## Open Questions

- What should a default actually derive from? Candidates: the template or image chosen, the connected repo, the purpose the user typed, a date/sequence ("Experiment 3"), or a first-message summary the way #1082 wants for sessions.
- Should the three object types share one scheme or each get their own? Experiments and knowledge bases carry more inherent context than a bare sandbox.
- Pre-fill or placeholder? A pre-filled value gets kept by default; a placeholder invites typing but risks empty names.
- Is there a case for keeping a friendly-name fallback when nothing better can be derived — and if so, should it at least be collision-free?
- Does this need a rename affordance to land properly, or is that separate work?

kapetr:
Every create flow pre-filled the name with a random adjective and noun — "misty-glacier", "rustic-otter". The name told you nothing about what you just made, and a list of ten read as a menagerie. Names are now a per-type sequence: `codingagent-1`, `experiment-6`, `knowledgebase-3`. Users can still type whatever they want over it.

The number is one past the highest already in use for that type, so deleting `experiment-3` leaves a gap instead of giving its name to the next experiment. The other two types don't count toward it. A name a user typed by hand does count if it happens to fit the pattern — typing `experiment-4` yourself pushes the next suggestion to `experiment-5`, which is the safe direction.

The suggestion follows the agent list rather than being fixed once. It starts from whatever list is cached, and is replaced as long as the field still holds what the prefill put there — so a suggestion made from a list that had not caught up with a create gets corrected when it does. Type anything over it, or clear it, and the field is yours from then on. When the list can't be fetched at all the suggestion is `<type>-1`, so the field is never left empty.

All four create paths are covered: the three setup pages, plus the inline create on the Slack and Telegram bind views.

Stacked on #3352, since the setup pages this changes live there.

## Where this drifts from the issue

- **No AI-generated titles.** The proposal in the thread had the pre-filled name replaced by a summary after the first turn. That was dropped in the thread itself once session-title quality came up, and the final comment asks only for `[sandbox type]-#`.
- **Two setup pages open at once will suggest the same number.** The count comes from the client's list, and names have no server-side uniqueness (IDs are separate), so this is cosmetic. A real guarantee needs the server to assign the name. The random names could collide too, so it is not a regression.
- **A restored setup draft keeps the number it was given.** The setup form persists its draft per tab, and a restored name is treated as the user's, since nothing distinguishes one they typed from one the prefill wrote. So a draft opened again later can carry a number that has since been taken.

Closes #3116


packages/ui/src/modules/agents/hooks/use-default-sandbox-name.ts
```diff
@@ -0,0 +1,41 @@
...
+const MATCHES_KIND = {
+  "coding-agent": isCodingAgent,
+  experiment: isExperimentSandbox,
+  "knowledge-base": isKnowledgeBase,
+} as const;
+
+export function useDefaultSandboxName(kind: SandboxNameKind): string | null {
+  const { data } = useAgents();
+  return useMemo(() => {
+    if (!data) return null;
```
dam-code-guardian: 🟡 **Warning:** `data` is undefined while the agents query is loading and also when it fails, so the suggestion stays `null` and the prefill effect returns early — the field keeps the `""` it now starts with, and `canCreate` stays false until the user types a name. The base branch produced a name synchronously on first render, and #3116 weighed prefill against placeholder and settled on a prefilled value, so a failed query silently lands on the rejected option. Reachable on the inline create form too: `useInlineAgentCreate` opens it as soon as the query settles.

**Fix:** A sandbox-name suggestion stays available when the list is not — falling back to `<prefix>-1` for an unavailable list keeps a default in every state, for both call sites.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `73d1474`

## PR #3363: feat(ui): prefill sandbox names as <type>-<n>
**Author:** kapetr | **Branch:** feat/3116-sequential-sandbox-names → feat/3234-per-feature-setup | **Changes:** +69 −64 (4 files)

### Summary
The random adjective-noun generator is replaced by a per-type sequence (`codingagent-1`, `experiment-6`) computed from the client's agent list, wired into all four create paths. The sequence logic holds: I ran `nextSandboxName` against leading zeros, mixed case, a gap, a custom name and an out-of-range ordinal, and each returned the expected value.
_(Suppressed 2 finding(s) per PR context: concurrent setup pages suggesting the same number, and other kinds not counting toward the sequence — both declared in the PR body.)_

### Findings
- 🟡 **Warning:** no default name at all when the agent list is unavailable (`packages/ui/src/modules/agents/hooks/use-default-sandbox-name.ts:23`)
- 🟢 **Suggestion:** A restored setup draft keeps the number it was given earlier in the session. `useSetupForm` persists the form on every change and clears it only after a successful create, so returning to a setup page restores `name: "experiment-6"`, and the prefill effect leaves it because the field is non-empty (`packages/ui/src/modules/sandboxes/hooks/use-setup-form.ts:61`, `use-default-sandbox-name.ts:39`). Same stale-number outcome as the concurrent-page case the body describes, but it lasts for the session rather than a moment.
- 🟢 **Suggestion:** The latch flips when the suggestion arrives rather than at mount, so the `name.length === 0` test runs at that moment (`use-default-sandbox-name.ts:37`). A field that was typed into and then cleared before the query lands receives the suggestion over that blank — the body's "your text stands" holds for text that is still there.
- 🟢 **Suggestion:** The body reads as though custom names do not count toward the number, and the pattern matches any `<prefix>-<digits>`, so a hand-typed `experiment-4` raises the next suggestion to `experiment-5` (`packages/ui/src/modules/sandboxes/lib/sandbox-name.ts:14`). The behavior lowers collision risk, so the divergence sits in the description.
- 🟢 **Suggestion:** `SandboxNameKind` and `SetupFlow` are separate unions with identical members (`sandbox-name.ts:1`, `use-setup-form.ts:8`). Adding a flow is caught at the `usePrefilledSandboxName` call site; adding a name kind is not.
- ✅ **Looks good:** `Number.isSafeInteger` on the parsed ordinal keeps a name like `experiment-99999999999999999999` from pinning the sequence — it falls through to the next real number.

✅ No findings.

### Issue Fit
The thread on #3116 settles on `[sandbox type]-#` with AI titles dropped, which is what this delivers. Its two findings are merged into `### Findings` above: the missing default when the agent list is unavailable, and the restored draft keeping a stale number.

Four findings, all merged into `### Findings` above: the prefill latch keyed on data arrival, the two identical unions, and the per-kind scan. Its concurrent-allocation finding is the case the PR body declares, and is suppressed per PR context.

Two findings, both merged into `### Findings` above: the effect-driven prefill leaving the field empty when `useAgents` has no data (including the error path, where `useInlineAgentCreate` opens this form immediately — verified), and a late suggestion landing in a field being edited.

### Verdict
COMMENT — one warning: a failed agent-list query leaves the name field empty, which is the option the issue thread weighed and rejected.

---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `2945b2a`

## PR #3363: feat(ui): prefill sandbox names as <type>-<n>
**Author:** kapetr | **Branch:** feat/3116-sequential-sandbox-names → feat/3234-per-feature-setup | **Changes:** +76 −68 (5 files)

### Summary
One commit reworks the prefill: the suggestion follows the agent list instead of latching the first value, an unavailable list yields `<type>-1` instead of nothing, and `SetupFlow` is now an alias of `SandboxNameKind`. I re-ran the latch as a script over six mount sequences — cold load, warm cache, user edit then refetch, navigate-away-and-restore, cleared field — and each ended on the value the body describes.
_(Suppressed 3 finding(s) per PR context: the client-side count not guaranteeing uniqueness, two setup tabs suggesting the same number, and a restored draft keeping its number — all three declared in the PR body.)_

### Changes since last review
Previous HEAD: 73d1474 (2026-08-18T16:03:48Z) — verdict COMMENT

- ✅ **Fixed:** no default name at all when the agent list is unavailable (`use-default-sandbox-name.ts:20`)
- ✅ **Fixed:** a suggestion landing over a field the user had cleared (`use-default-sandbox-name.ts:35`)
- ✅ **Fixed:** `SandboxNameKind` and `SetupFlow` as two separate unions (`use-setup-form.ts:9`)
- ✅ **Fixed:** the body's account of hand-typed names counting toward the sequence
- 🔁 **Still present:** a restored setup draft keeps the number it was given (`use-setup-form.ts:60`) — now declared in the body
- 🆕 **New:** the two suggestions below

### Findings
- 🟢 **Suggestion:** On a cold cache the field shows `<type>-1` first, then the real next number once `useAgents` resolves (`packages/ui/src/modules/agents/hooks/use-default-sandbox-name.ts:20`). The correction is the declared design. The narrow part is a submit inside that window, which needs the provider and template queries resolved while the agent list is not — it would create a second `<type>-1`. Gating the first write on the query having settled would close it, at the cost of the empty field the issue thread rejected.
- 🟢 **Suggestion:** `sandbox-name.ts` and `usePrefilledSandboxName` landed in `agents/`, so `sandboxes/hooks/use-setup-form.ts` now reaches into a sibling module for a helper all three setup flows share (`packages/ui/src/modules/sandboxes/hooks/use-setup-form.ts:6`). That edge already exists in 13 other files under `sandboxes/`, so this follows the local pattern rather than starting one.

✅ No findings.

### Issue Fit
The thread's final scoping comment asks for `[sandbox type]-#`, and the three prefixes match its examples. All four create paths are covered, and the dropped AI-title idea was retracted in the thread, not missed. Its one finding — the issue's "no duplicates" user story against a count taken from the client's cached list — is suppressed per PR context: the body states the gap and its cause, and the random generator carried the same exposure. It records the missing-default finding from the last round as resolved.

One finding, merged into `### Findings` above as the module-placement suggestion. Its severity is lowered because the `sandboxes` → `agents` edge is pre-existing baseline (13 files at the merge base) and the skill's layering rules cover server packages, not `packages/ui`. It records the prefill latch and the duplicated unions from the last round as resolved, and `nextSandboxName` as relocated byte-identical.

Two findings. The cold-cache prefill is merged into `### Findings` above; the same-name-in-two-tabs suggestion is suppressed per PR context. It records both of last round's findings as resolved.

### Verdict
APPROVE — the one warning from the last round is fixed at its cause, and nothing new blocks.

---

Issue #3333, opened 2026-08-14
Agents post in Slack under their own name and avatar


jenna-winkler:
## Context

We put agents into several IBM project channels this month. People in those channels do not know DAM. They see a bot called "DAM" post messages, and they read the platform as the thing talking to them. One team reported it as a DAM agent spamming their channels. The agent itself worked as intended.

## Problem

Every agent in Slack posts as one shared bot: same name, same avatar. The agent's own name appears only in small text under the message. So the platform gets the credit and the blame for what an individual agent says.

Two costs. First, users cannot tell agents apart in a channel, or tell which agent to talk to. Second, any bad turn — a wrong answer, a message nobody asked for — lands on DAM instead of on that one agent. Teams that adopt an agent want it to look like *their* agent, not like an external platform in their channel.

## Goal

An agent in Slack looks like itself. Channel members see the agent's name and avatar on each message, with DAM as the quiet attribution underneath. Members of a project channel recognise the agent as part of their project.

## User Stories

- As an agent owner, I want my agent to post under its own name so that channel members recognise it.
- As an agent owner, I want to set my agent's avatar so that it fits the team it serves.
- As a channel member, I want to tell two agents apart in the same channel so that I know which one answered.
- As a channel member, I want to see that an agent runs on DAM so that I can find out what it is.

## Scope

In: per-agent name and avatar on outbound Slack messages, and where the DAM attribution goes when the name changes.

Out: Telegram. Out: the web UI. Out: guidance on how to configure a well-behaved agent — that is separate work.

## Open Questions

- Slack needs the `chat:write.customize` scope for this. Do we get approval for it? This needs a decision before build work starts.
- Does the agent name come from the agent's existing name, or a separate Slack display name?
- Where does "powered by DAM" sit once the bot name is per-agent, so it stays visible but small?
- Do we review with the stakeholder who raised this before we ship?

pilartomas:
This is not technically achievable with the contemporary Slack application/bot model and the scope limitations enforced by IBM. Customize scope is prohibited.

I'm afraid there is not much we can do here.

jenna-winkler:
Closing given above. We've made lots of slack improvements since so I'm less concerned about this now. Thanks @pilartomas !

PR #3367, opened 2026-08-19
feat(channels): several agents can share one Slack conversation

### LINKED ISSUE #3366: Multi-agent Slack channels
## Context

Agents are going into real project channels, and a project channel serves a project, not a single agent. We recently made one agent reachable from several channels (#3086). The reverse — several agents reachable from one channel — is still refused.

There's an odd asymmetry today: an agent can already *post* into any channel the bot belongs to, and the platform already labels who wrote what when several agents appear in one channel. So people can already watch several agents speak in a channel, but they can only reply to one of them.

## Problem

A Slack conversation can be connected to exactly one agent. Connecting a second is refused outright.

Teams that want a channel to reach more than one agent have to split the conversation across a channel per agent, which defeats the point of a project channel, or funnel everything through one agent that has to do every job. Neither matches how people actually work — they want to ask a question in the channel they already use and have the right agent answer.

## Goal

A project team connects every agent it relies on to the channel it already works in, addresses the one it wants, and gets an answer from that agent — without needing to know in advance which agent handles what.

## User Stories

- As a user, I want to connect more than one agent to a channel so that my team can reach all of them where we already talk.
- As a user, I want to name the agent I'm addressing in a mention so that the right one answers.
- As a user, I want a plain mention with no name to still reach an agent so that I don't have to remember who is connected.
- As a user, I want the agent I asked to pass my question to a better-suited agent in the channel so that I get an answer without knowing who to ask.
- As a user, I want to turn ambient mode on for one agent without turning it on for all of them so that a channel isn't read by more agents than I intended.
- As a user, I want a channel with one agent to keep behaving exactly as it does today so that nothing I already rely on changes.

## Scope

In: several agents connected to one Slack conversation; a default agent per conversation; addressing an agent by name in a mention; handing a message to another agent in the same channel; ambient mode chosen per agent.

Out: Telegram — a conversation there stays with one agent. Out: per-agent name and avatar (#3333). Out: agents talking to each other unprompted.

## Proposed solution

- One connected agent is the channel's **default** — the first one connected, changeable later. A plain mention goes to it, so a channel with a single agent behaves exactly as it does now and nobody has to learn anything new until they add a second.
- A mention that begins with an agent's name goes to **that agent**. No match, or a name that matches more than one agent, falls back to the default.
- The agent handling a message can **hand it to another agent** connected to the same channel. That makes the default a router when the name didn't match, and rescues a mention that named the wrong agent. A handed-on message can't be handed on again, so it can't bounce between agents.
- **Ambient mode** becomes a per-agent choice rather than a property of the channel.

## Open Questions

- Can every connected agent hand a message on, or only the default? Allowing any agent rescues a mention that named the wrong one; restricting it to the default keeps routing easier to predict.
- Agent names aren't unique, so two agents in one channel can share a name. Falling back to the default is the simplest answer — is that good enough, or should the channel say it couldn't tell them apart?
- When a channel's default agent is disconnected and others remain, should another be promoted automatically or should we ask?
- Per-agent ambient mode means several agents can read every message in a channel. Should there be a cap, or is that the user's call?
- Every agent still posts under the same bot name, so a channel with several agents is harder to read. #3333 covers per-agent identity — is there something cheaper worth doing here in the meantime?

## Additional resources

- #3086 — the inverse capability: one agent reachable from several channels
- #3333 — per-agent name and avatar in Slack

pilartomas:
Closes #3366.

## What this does

A Slack conversation could hold exactly one agent. Now it can hold several, so a project channel reaches every agent its team relies on instead of forcing a channel per agent.

The old exclusivity rule **moves rather than disappears**:

- `channels_slack_channel_unique_idx` (unique on the conversation id) is replaced by a unique index on `(agent_id, slackChannelId)` — one agent still connects to one conversation only once.
- A second, narrower partial unique index admits at most one row per conversation marked `default`, which is what keeps a bare mention unambiguous.

## Addressing

One install-wide bot serves every agent, so a bare `@dam` cannot say which agent is meant. Resolution is therefore:

| Message | Reaches |
|---|---|
| `@dam …` | the conversation's **default agent** (the first connected) |
| `@dam Reviewer …` | the agent named `Reviewer` |
| `@dam <unknown> …` | the default agent |
| `@dam <name matching two agents> …` | the default agent, told the name it could not resolve |

Name matching is scoped to that conversation's roster, whole-name, case-insensitive, longest-name-first. Agent names are not unique (`agentCreateInputSchema` only forbids the `agent-` prefix), which is why an ambiguous name resolves to the default rather than a guess.

**A single-agent conversation behaves exactly as it did before** — nothing about the default surfaces until a second agent joins.

## Hand-off

New `hand_off_to_agent` MCP tool: the agent handling a turn can give it to a peer connected to the same conversation — the default acting as a router, or an agent that was named but is the wrong one.

- Server-side. An agent physically cannot reach a peer's MCP endpoint (the controller pins each agent's SPIFFE principal to its own path), so this has to be the api-server acting on the caller's behalf.
- Fire-and-forget. `relayOwnerTurn` awaits the whole turn including cold-pod wake, so awaiting the peer would block the handing agent's harness while it holds attachment budget.
- **Hop limit 1** — a handed-on message cannot be handed on again. Structural, so no cycle detection is needed.
- Authorized by the target's *binding*, not an owner match: agents from different owners legitimately share a channel, and the binding is the existing authorization model. Gated by the target binding owner's ToU.
- The handing agent's turn is marked disposed, so a hand-off doesn't trip the unanswered-turn warning. Attribution stays with the original human sender; the hop is recorded.

## Ambient

Ambient becomes a **per-binding** choice — one agent can read along in a channel where another only answers mentions.

Where several agents read along, they take each message **one at a time**, each starting only once the previous one finished. Running them concurrently would have them talk over each other; in sequence, each can be told whether an earlier one already replied, so a later agent adds something or stays silent rather than repeating an answer.

The **default agent goes first** where it reads along — it is the conversation's primary, so it gets first refusal — and the rest follow **shuffled**, freshly per message. Ranking the rest by connection age was arbitrary-but-stable, which is the worst combination: it reads as meaningful while ranking on nothing, and it quietly hands the same agent every second look. No agent is told its position.

Releasing the default promotes nobody — see **No default is a supported state** below.

This also means the existing single attachment `release()` stays correct — one drain awaits every reader before freeing, so no refcounting is needed.

`/dam unbind` and `/dam ambient` now take an agent name when more than one is connected, refusing and listing candidates rather than guessing (mirroring the pattern `channel slack disconnect` already uses in the other direction).

## Migrations

- `0030` — index swap plus `created_at` on `channels`, which gives the in-chat roster a stable order: the order the agents joined.
- `0031` — backfills `default: true` on existing bindings. Every binding that exists today is the sole binding for its conversation, so each is its conversation's default. **Without this a bare mention in an already-connected channel would resolve to no default at all.**

Forward-safe; effectively irreversible (rolling back would mean deleting all but one binding per conversation).

## No default is a supported state

Only an agent's **owner** may make it a channel's default, because the default absorbs every unnamed mention and that load runs on their credentials. Automatic promotion would hand exactly that load to an owner who was not there to accept it, so **nothing is promoted** when a default is released.

A conversation with no default refuses unnamed mentions, naming the connected agents and pointing at `/dam default <agent>`; named mentions are unaffected. A first connect is the one exception, claiming a vacant default automatically — the owner is the one connecting, so they see it happen.

Changing it is **in-chat only** (`/dam default [agent]`): choosing sensibly means seeing who else is connected, and the channel is the only surface that shows that. A bare `/dam default` reports the current default and the whole roster.

## Drive-by fixes

Two things that were already wrong and that this PR touches anyway:

- Both connect surfaces claimed ambient changes are announced in the channel. `docs/architecture/channels.md` says they deliberately are not.
- The architecture page described MCP caller auth as a source-IP `podIpResolver` cache. That code does not exist — identity is the `:id` path segment plus an Istio principal pin. Corrected so doc-drift stops pointing at a mechanism that isn't there.

## Verification

`tsc` clean across api-server, api-server-api, ui, cli, db, e2e. Unit tests: api-server 1051/1053, ui 294, cli 154. The two api-server failures are `cluster-ca-probe.test.ts`, untouched by this branch — it probes live network endpoints. Gates: comment-types, `docs:check` (including the 40k page cap — the tool catalogue was condensed to make room), `db:check:generated`, prettier and eslint all pass.

New coverage: `slack-routing.test.ts` (20 cases over the pure routing rules) and `slack-multi-agent.test.ts` (15 integration cases: mention routing, per-agent sessions on one thread, sequential ambient ordering, the "already replied" notice, and every hand-off refusal path).

Not run: live cluster / e2e. `09-slack-inchat-bind.spec.ts` has its two inverted assertions updated; `full/slack-multi-channel.spec.ts` is the template for a symmetric two-agents-one-channel spec that only means anything against a cluster.

## Open questions from the issue, as built

- **Any connected agent can hand off, not just the default.** Otherwise a mention that named the wrong agent is stuck. The hop limit bounds it either way.
- **Agent name is not yet added to the message body.** With no `chat:write.customize`, the context-block footer is still the only thing distinguishing agents in a shared channel; #3333 owns per-agent identity, and its blocking question (that scope) is now answered — we do not get it.


pilartomas:
## Review — verified findings

Reviewed at head 954feeb1 against main. Method: three parallel review passes (core logic · data layer & migrations · tests, surfaces & docs), every finding then adversarially verified by tracing the code path; refuted candidates dropped. CI at review time: `mise test` green, `mise check` green, **`mise e2e` red** (first finding), `dam:review` pending.

**Verdict: solid design, not mergeable yet.** The data layer, routing rules, and migrations are genuinely well built, and the architecture page largely matches the code. What blocks merge: the e2e gate is red from a heading rename this PR made but did not propagate, and `hand_off_to_agent` has two confirmed defects.

### Blocking — CI is red because of this PR

**Bind-page heading rename breaks the auth deep-link e2e spec.**
`packages/ui/src/modules/slack/views/slack-bind-view.tsx:75,99` renames the page title from "Connect this channel to an agent" to "Connect an agent to this channel", but `packages/e2e/playwright/src/tests/smoke/01-auth.spec.ts:109` (untouched by this PR) still locates the picker by `getByRole("heading", { name: /connect this channel to an agent/i })`. The spec fails element-not-found; `mise e2e` is red on exactly this test. `09-slack-inchat-bind.spec.ts` was updated; this one was missed.
**Fix:** update the heading regex (or share the title string between UI and specs).

### Major

**1. Hand-off from an ambient turn relays an empty message.**
Only `relayOwnerTurn` populates `text`/`slackUserId`/`hasThread` on a turn ref; ambient refs are created with only `channel`/`threadTs`/`eventTs` (`slack.ts:2054–2058`). `handOffTurn` resolves ambient refs happily and builds the handed message from `ref.text ?? ""` with `slackUserId: ref.slackUserId ?? ""` (`slack.ts:2578–2599`). The ambient contract actively invites this: `rosterSentences` ("hand it to them with hand_off_to_agent rather than answering outside what you know") is included on ambient turns too. An ambient reader following its own prompt hands a peer a message whose body is empty (or just the hand-off note) — the human's question is lost, the security log records `slackUserId: ""`, and the tool still reports success.
**Fix:** populate the three fields on ambient refs (`relayAmbientTurn` has all of them), or refuse hand-off from refs that lack them — and decide deliberately whether ambient turns should advertise the tool at all.

**2. A turn can be handed off repeatedly, and stale turns can be handed off later.**
`ref.handedOff` is assigned (`slack.ts:2578`) and never read; `handOffTurn` refuses only on `ref.forwarded` (the receiving side of a hop). The same agent can call `hand_off_to_agent("B")`, then `("C")`, then `("B")` again in one turn — each call fires a full relay, so peers answer the same message multiple times, each turn billed to the target owner. Reachable by prompt injection from any channel member. Separately, `resolveTurn` (`slack.ts:828–846`) falls back to `lastTurn` when nothing is in flight, so a later scheduled or UI-driven session can hand off a long-finished Slack turn — the tool description's "only works while you are answering a Slack message" is not enforced. Tests cover the receiver-side hop limit only.
**Fix:** refuse when `ref.handedOff` (arguably also `declined`/`posted`) is already set, and restrict hand-off to in-flight/lingering refs rather than the `lastTurn` fallback.

### Minor

**3. Any later connect or re-connect silently grabs a vacant default, with no audit record.**
`claimSlackDefaultIfVacant` runs unconditionally after every successful upsert (`agents-service.ts:552`) — including the UI ambient-toggle re-connect path. After the default is released, the channel is told "mentions with no name now reach no one until an agent's owner runs `/dam default`" (`slack.ts:1512`), and `channels.md` says changing the default is "in-chat only" and the default is "the first connected" — yet the next agent to connect (or an existing agent whose owner flips ambient in the UI) silently takes it. No `channel.default_changed` security log is written on this path.
**Fix:** gate the claim to the genuinely-fresh-conversation insert path, and security-log the acquisition wherever it stays.

**4. Default claim runs outside the binding transaction, and nothing heals a default-less sole binding.**
The binding upsert commits inside `unitOfWork` (`agents-service.ts:533–549`); the claim is a separate autocommit statement after it. A crash between the two leaves a sole binding with no default — then a bare mention is *refused* (`noDefaultAgentCopy`, `slack.ts:1922–1928`), regressing "a single-agent conversation behaves exactly as it did before" for that channel. Retrying the in-chat bind can't heal it (the `ChannelAlreadyBound` pre-check at `agents-service.ts:304–306` fires before `connectShared`); only a UI/CLI re-connect or `/dam default` repairs it.
**Fix:** claim inside the unit of work, or heal at read time when the roster has exactly one binding and no default.

**5. A react-only turn counts as "already replied" for later ambient readers.**
The `react` tool sets `ref.posted = true` via `noteEngagedTurn` (`slack.ts:2727` → `869`); `relayAmbientTurn` reports that as posted (`slack.ts:2196`), and the drain loop pushes the reader into `answeredAlready` (`slack.ts:2322`). Later readers are told "X already replied to this in the channel… read what they said before deciding" — about a reply that doesn't exist. The same guidance explicitly endorses reaction-only responses, so this is a routine path.
**Fix:** track posted-a-message separately from engaged (reaction), and only count the former.

**6. Hand-off silently drops the sender's attachments.**
`relaySharedTurn` is called with `images: []`, `files: []` (`slack.ts:2593–2595`); the ref carries no attachments. Neither the target agent nor the sender is told anything was dropped — `channels.md:174`'s own contract says anything withheld "is named to the agent, and on an addressed turn to the sender too."
**Fix:** re-fetch the event's attachments for the relay, or at minimum name the drop to both sides.

**7. Every inbound Slack event now triggers full agent projections per bound agent — even on drop paths.**
`resolveRoster` (`slack.ts:919–930`) calls `agents().get` per binding, and each `get` runs the full projection (K8s status read, env list, template spec — `agents-service.ts:500–517`). It runs on every plain message in a bound channel *before* the ambient drop check (`slack.ts:2339–2345`), plus on every mention, slash command, and hand-off. On main, the plain-message drop path was one DB read.
**Fix:** resolve names only when actually relaying, and use a lighter name lookup (or cache) instead of the full projection.

**8. Ambient guidance tells the reader that ambient-off peers "read along here too".**
The drain loop filters `readers` for ambient+ToU but passes the *full* roster into the guidance (`slack.ts:2270–2305`), which claims all peers "read along here too" and instructs "a message that names one of them is addressed to them — stay silent on it" (`slack-turn-copy.ts:212`). An ambient-off peer never receives plain messages, so a plain message naming that peer is answered by nobody.
**Fix:** pass the reader set (or annotate who reads along) so the copy tells the truth.

**9. PR body claims default promotion on release; the code deliberately ships none.**
"Default promotion on release still uses longest-connected" and 0030's "deterministic oldest-survivor rule" describe behavior that doesn't exist: no promotion path is wired (`claimSlackDefaultIfVacant` is connect-only, `setSlackChannelDefault` is `/dam default`-only), and the unbind ack, `channels.md:15`, and the test at `slack-multi-agent.test.ts:549` all encode no-promotion — deliberately, per head commit 954feeb1.
**Fix:** update the PR description (and reword 0030's migration comment, which still hints at the ordering rule).

**10. Test coverage gaps on behaviors the PR names as guarantees.**
The hand-off suite (`slack-multi-agent.test.ts:308–386`) covers not-connected, self, already-handed, and no-turn refusals — but not the ToU refusal (`slack.ts:2569–2575`) or the ambiguous-target refusal (`slack.ts:2559–2564`), and nothing asserts the disposed/declined marking. No test asserts the ambiguity notice actually reaching the default agent's prompt (only the pure `ambiguousName` flag is tested).

### Nits

- `channels.md:15` overstates unknown-name reporting: only *ambiguous* names are named to the default; an unrecognized first word is indistinguishable from message text (and the code correctly doesn't try).
- `channels.md:124` still says the link table authorizes and the usage help advertises only "bind, unbind, and ambient" — the PR adds `/dam default` to both (`slack.ts:1526`, `1588`).
- `channels.md:177` vs `channels.md:222` disagree on the Slack turn-tool set (one lists `no_reply_needed`, the other `hand_off_to_agent`).
- Unused imports: `defaultOf` in `slack.ts:107` (already bot-flagged) and `claimSlackDefaultIfVacant` in `bootstrap.ts:21`.

### Design question — decide deliberately

**Occupied channels are now silently joinable from the UI/CLI.** On main, the exclusivity refusal incidentally shielded in-use channels from direct connects. Now any platform user can attach their agent — including with `--ambient` — to any conversation the install-wide bot is in, by channel id, with no in-channel trace: the in-chat bind announces itself, the direct connect never has. Multi-owner sharing is the feature and binding-as-consent is the documented model, so this may be exactly as intended — but "a second owner's agent can start reading a channel with no message posted there" is a new exposure worth an explicit decision (announce direct joins in-channel, or state the position in `channels.md`).

### What's solid

- **Data layer.** `claimSlackDefaultIfVacant` pairs a NOT EXISTS guard with the partial unique index as a backstop; `setSlackChannelDefault`'s unmark-then-mark transaction serializes concurrent default changes and handles the raced-unbind rollback. The upsert's `(config - 'ambient') || new` merge deliberately preserves the default flag across re-connects.
- **Migrations.** Drizzle runs 0030+0031 in one transaction, so the index swap and backfill commit atomically and 0031's sole-binding premise provably holds at execution time; the backfill is idempotent, and the jsonb-boolean representation is consistent across index predicate, backfill, and reads.
- **Routing.** Pure, injectable-rng helpers with deterministic tests; tie-break and longest-name semantics match the spec; roster ordering is deterministic (`created_at, agent_id`).
- **Tests are honest.** Sequential ambient execution is asserted via start/end interleaving with an async gap (concurrency would interleave), not just call order; every flipped assertion in existing tests reflects the legitimate behavior change.
- **Hop limit** is structural on the receiving side — `forwarded` rides the receiver's ref and survives lingering/`lastTurn`.
- **Both drive-by doc fixes verified correct** against main and head.

packages/api-server/src/modules/channels/infrastructure/slack.ts
```diff
@@ -2194,6 +2524,103 @@ export function createSlackWorker(
...
+            "you can't.",
+        };
+
+      const self = await resolveAgentName(instanceName);
+      ref.handedOff = true;
+      ref.declined = true;
+
+      const handedText = note
+        ? `${ref.text ?? ""}\n\n[${self} handed this to you: ${note}]`
+        : (ref.text ?? "");
```
dam-code-guardian, line 2639: 🔴 **Critical:** The handed message is rebuilt from `ref.text`, `ref.slackUserId` and `ref.hasThread`, but the `TurnRef`s that `relayAmbientTurn` builds set only `channel`, `threadTs` and `eventTs` (`slack.ts:2054`). `resolveTurn` returns ambient and `lastTurn` refs the same as mention refs, so an agent that hands off while reading along sends `handedText` = just the note (or empty string) and `slackUserId` = `""`. Running both construction sites through the hand-off body confirms it: the mention path yields the original text, the ambient path yields `"\n\n[Scribe handed this to you: …]"`.

**Fix:** Set `text`, `slackUserId` and `hasThread` on every `TurnRef` where it is created, so any resolvable turn carries what a hand-off needs; alternatively refuse the hand-off when the resolved ref lacks them, rather than substituting `?? ""`.

packages/api-server/src/modules/channels/infrastructure/slack.ts
```diff
@@ -2194,6 +2524,103 @@ export function createSlackWorker(
...
+      if (!(await isTermsAccepted(target.owner)))
+        return {
+          error:
+            `"${target.name}" cannot take turns yet — whoever connected it ` +
+            "must accept the Terms of Use first. Answer it yourself, or say " +
+            "you can't.",
+        };
+
+      const self = await resolveAgentName(instanceName);
+      ref.handedOff = true;
```
dam-code-guardian, line 2625: 🟡 **Warning:** `ref.handedOff` is assigned here and read nowhere in the file. The hop-limit guard above tests `ref.forwarded`, which is set on the *receiver's* new ref, so it stops a second hop but not a second hand-off of the same turn — the ref stays resolvable, and repeated calls fan one message out to several agents.

**Fix:** Test `ref.handedOff` alongside `ref.forwarded` in that guard.

packages/api-server/src/modules/channels/infrastructure/slack.ts
```diff
@@ -2194,6 +2524,103 @@ export function createSlackWorker(
...
+      ref.declined = true;
+
+      const handedText = note
+        ? `${ref.text ?? ""}\n\n[${self} handed this to you: ${note}]`
+        : (ref.text ?? "");
+
+      void relaySharedTurn({
```
dam-code-guardian, line 2641: 🟡 **Warning:** The relay is dispatched fire-and-forget, so `{ ok: true }` and `Handed to <name>` come back before anything is accepted. The handing agent then ends its turn posting nothing, and a relay that never starts — no gateway, wake failure, the terms ephemeral — drops the message with only a `slack.hand_off.failed` log line. The PR body's reasoning for not awaiting is sound (the whole turn plus cold-pod wake), but acceptance is earlier than completion.

**Fix:** Have `relaySharedTurn` report whether the receiving turn was accepted, and return `ok` only then, so the handing agent can answer it itself when the hand-off does not land.

packages/api-server/src/modules/channels/infrastructure/slack.ts
```diff
@@ -863,6 +907,28 @@ export function createSlackWorker(
...
+  async function resolveAgentName(instanceName: string): Promise<string> {
+    try {
+      const agent = await agents().get(instanceName);
+      return agent?.name?.trim() || instanceName;
+    } catch {
+      return instanceName;
+    }
+  }
+
+  async function resolveRoster(slackChannelId: string): Promise<RosterEntry[]> {
```
dam-code-guardian, line 941: 🟡 **Warning:** Each roster entry's name comes from `agents().get()`, which runs the full `project()` — channels list, pod status, env list, template lookup. `resolveRoster` runs on every inbound mention, every ambient message, every ambient drain batch and every slash command, so an N-agent channel costs N projections per Slack event. The single-agent path had one; this multiplies it.

**Fix:** Resolve roster names through a name-only read, or a short-TTL cache keyed by instance name, rather than the full projection.

packages/api-server/src/modules/agents/services/agents-service.ts
```diff
@@ -536,6 +549,8 @@ export function createAgentsService(deps: {
 
     if (!txResult.ok) return txResult;
 
+    await deps.claimSlackDefaultIfVacant(id, slackChannelId);
```
dam-code-guardian: 🟢 **Suggestion:** `connectSlackImpl` is also the UI's ambient-toggle path (`planSlackChannelSave` re-issues `connectSlack` to flip ambient), so toggling ambient on a conversation whose default was released claims the vacant default as a side effect. That is the outcome the removed promotion was meant to avoid, and `docs/architecture/channels.md:124` states changing the default is in-chat only. Harmless where the toggling owner would want it; worth deciding deliberately rather than inheriting it from the connect path.

packages/ui/src/modules/sandboxes/components/channels/slack-channel-card.tsx
```diff
@@ -80,7 +80,11 @@ function SlackChannelRow({
   const handleDisconnect = async () => {
     if (
       await showConfirm(
-        `Mentions in ${channel.slackChannelId} will stop reaching this sandbox.`,
+        `Mentions in ${channel.slackChannelId} will stop reaching this sandbox.${
+          channel.default
```
dam-code-guardian: 🟡 **Warning:** A solo binding is always the default — the connect path claims a vacant one — so the common single-agent case gets a warning about "another agent's owner" who does not exist. The card also offers no way to claim or change the default, and the copy never names the Slack `default` command that can, unlike the ambient copy which does reference its in-chat command.

**Fix:** Word the warning without asserting peers exist, and point at the in-chat `default` command as the affordance.

packages/ui/src/modules/sandboxes/components/channels/slack-channel-card.tsx
```diff
@@ -99,6 +103,9 @@ function SlackChannelRow({
         </p>
         <p className="truncate text-sm text-muted-foreground">
           {channel.ambient ? "Ambient on" : "Ambient off"}
+          {channel.default
```
dam-code-guardian: 🟡 **Warning:** The default-agent sentence is appended after the ambient text inside a single `truncate` line in a `min-w-0 flex-1` column, so on a narrow card the newest and longest information clips first, with no `title` to recover it. `main` had only the short `Ambient on|off` here, which fit.

**Fix:** Render the flag as its own non-truncating element — the existing `Badge` (`variant="muted"`) beside the channel id reads well — instead of concatenating it into a truncated sentence.

packages/api-server/src/modules/channels/infrastructure/slack.ts
```diff
@@ -97,6 +98,28 @@ import {
   parseAgentFooter,
   type AgentFooter,
 } from "./agent-footer.js";
+import {
+  defaultOf,
```
dam-code-guardian: 🟢 **Suggestion:** `defaultOf` is imported but never called here — `routeMention` applies the default internally. The repo's quality bot flagged the same line.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `954feeb`

## PR #3367: feat(channels): several agents can share one Slack conversation
**Author:** pilartomas | **Branch:** feat/multi-agent-slack-channels → main | **Changes:** +7652 −260 (46 files)

### Summary
Replaces the one-agent-per-Slack-conversation rule with a roster: name-addressed mention routing, a per-conversation default agent, per-binding ambient with sequential read-along, and a new `hand_off_to_agent` MCP tool. The index swap, the backfill, and the no-promotion decision all hold up; the findings below sit in the hand-off path and in the docs that describe the new rules.

### Findings
- 🔴 **Critical:** A hand-off from an ambient turn relays an empty message. (`packages/api-server/src/modules/channels/infrastructure/slack.ts:2583`)
  **Fix:** Populate `text` / `slackUserId` / `hasThread` on every `TurnRef` at construction, or refuse the hand-off when the resolved ref lacks them.
- 🟡 **Warning:** `ref.handedOff` is set but never read, so one turn can be handed off repeatedly. (`packages/api-server/src/modules/channels/infrastructure/slack.ts:2578`)
  **Fix:** Include `ref.handedOff` in the guard next to the existing `ref.forwarded` check.
- 🟡 **Warning:** The hand-off reports success before the receiving turn is accepted. (`packages/api-server/src/modules/channels/infrastructure/slack.ts:2585`)
  **Fix:** Return `ok` only once the receiving relay is accepted, so a failure comes back to the handing agent as an error.
- 🟡 **Warning:** `resolveRoster` costs one full agent projection per connected agent, per Slack event. (`packages/api-server/src/modules/channels/infrastructure/slack.ts:919`)
  **Fix:** Read roster names from a name-only lookup or a short-TTL name cache instead of `AgentsService.get`.
- 🟡 **Warning:** `docs/architecture/channels.md` describes three rules this PR did not ship. (`docs/architecture/channels.md:15`) — from doc-drift; full text in its section below.
  **Fix:** Align the page's default-count wording, its unresolved-name sentence, and its three in-chat command enumerations with the shipped behavior.
- 🟡 **Warning:** The PR body still describes automatic default promotion, which the final commit removed. (PR description) — from issue-fit; full text in its section below.
  **Fix:** Move the no-promotion decision into the body's "Open questions from the issue, as built" list.
- 🟡 **Warning:** The new default-agent status and its disconnect warning read poorly on the channel card. (`packages/ui/src/modules/sandboxes/components/channels/slack-channel-card.tsx:84`) — from react-ui-engineering; full text in its section below.
  **Fix:** Word the warning without asserting other agents exist, and name the in-chat `default` command as the affordance.
- 🟢 **Suggestion:** Toggling ambient from the UI reaches `claimSlackDefaultIfVacant`, so an owner editing ambient on a conversation whose default was released silently becomes its default — which the docs describe as in-chat only. (`packages/api-server/src/modules/agents/services/agents-service.ts:552`)
- 🟢 **Suggestion:** `defaultOf` is imported but unused; `routeMention` applies the default internally. Also reported by the repo's own quality bot. (`packages/api-server/src/modules/channels/infrastructure/slack.ts:102`)
- ✅ **Looks good:** The migration pair is sound — dropping the conversation-wide unique index while adding both the `(agent_id, slackChannelId)` and the partial default index keeps each stated invariant, and `0031`'s `NOT (config ? 'default')` predicate makes the backfill re-runnable.

- 🟡 **Warning:** Three statements on the page no longer match the code. It says each conversation has "exactly one default Agent" two sentences before stating that releasing it leaves none, while the index admits *at most* one (`docs/architecture/channels.md:15`, against `:263`). It says an "unrecognised or ambiguous name … falls through to the default, which is told the name it could not resolve", but an unmatched name returns `ambiguousName: null`, so only the ambiguous case carries a name forward (`:15`, against `packages/api-server/src/modules/channels/infrastructure/slack-routing.ts:75`). And it enumerates the in-chat command set as `bind`, `unbind`, `ambient` in three places, without the `default` command this PR adds and the usage help now advertises (`:124`, `:82`, `:263`).
  **Fix:** "At most one" where the page asserts the count; narrow the fall-through sentence to the ambiguous case; add `default` to all three command enumerations.

### Issue Fit
- 🟡 **Warning:** The body still describes automatic default promotion — "Default *promotion* on release still uses longest-connected", and `0030` "gives default promotion a deterministic 'oldest survivor' rule" — while the shipped code promotes nobody and refuses unnamed mentions until an owner claims the default. The issue's open question "should another be promoted automatically or should we ask?" reads one way in the body and the other in the code (`packages/api-server/src/modules/channels/infrastructure/slack.ts:1514`).
  **Fix:** Update the Ambient and Migrations paragraphs to the as-built rule and move the decision into the body's "Open questions from the issue, as built" section.

- 🔴 **Critical:** Merged into `### Findings` above — `handOffTurn` rebuilds the handed message from `ref.text` / `ref.slackUserId` / `ref.hasThread`, which `relayAmbientTurn`'s refs never set (`slack.ts:2054`), and `resolveTurn` returns ambient and `lastTurn` refs freely.
- 🟡 **Warning:** Merged into `### Findings` above — fire-and-forget dispatch means `Handed to <name>` is reported even when the relay never starts (no gateway, wake failure, terms ephemeral), leaving the message dropped with only a log line.
- 🟡 **Warning:** Merged into `### Findings` above — `ref.handedOff` is written and never read; the guard checks `ref.forwarded`, which is set only on the receiver's new ref.
- 🟡 **Warning:** Merged into `### Findings` above — `resolveRoster` resolves each name through `agents().get()`, which runs a full `project()` (channels, status, env, template lookup) per agent, on every mention, ambient message, ambient drain and slash command.
- 🟢 **Suggestion:** Merged into `### Findings` above — unused `defaultOf` import.

- 🟡 **Warning:** The disconnect-confirm copy says mentions "will reach no one until another agent's owner claims the default", but a solo binding is always the default (connect claims a vacant one), so the common case is warned about an agent that does not exist — and the card offers no way to claim or change the default, nor names the Slack `default` command that can (`packages/ui/src/modules/sandboxes/components/channels/slack-channel-card.tsx:83`).
  **Fix:** Word the warning without asserting other agents exist, and point at the in-chat `default` command the way the ambient copy references its own.
- 🟡 **Warning:** The default-agent status is appended after the ambient text inside one `truncate` line in a `min-w-0 flex-1` column, so the longest and newest text is clipped first on a narrow card, with no `title` to recover it (`packages/ui/src/modules/sandboxes/components/channels/slack-channel-card.tsx:104`).
  **Fix:** Render the flag as its own non-truncating element — the existing `Badge` (`variant="muted"`) beside the channel id — rather than concatenating it into a truncated sentence.
- 🟢 **Suggestion:** `channel.default` renders from the agents list cache, invalidated only by this UI's own mutations, so a default changed from Slack stays stale until a manual refetch. Pre-existing for `ambient` too — worth a separate issue for a live-hint topic on Slack binding changes (`packages/ui/src/modules/sandboxes/components/channels/slack-channel-card.tsx:106`).

### Verdict
REQUEST_CHANGES — a hand-off started from an ambient turn delivers an empty message to the receiving agent.

---

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `bdd5335`

## PR #3367: feat(channels): several agents can share one Slack conversation
**Author:** pilartomas | **Branch:** feat/multi-agent-slack-channels → main | **Changes:** +7924 −268 (47 files)

### Summary
Complete re-review at the live HEAD, requested by the label. The previous round's critical hand-off defect is fixed at its cause, and every other blocking finding is resolved; two new warnings sit in the disconnect-confirm copy.

### Changes since last review
Previous HEAD: `954feeb` (2026-08-19T10:31:18Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** Hand-off from an ambient turn no longer relays an empty body — both `TurnRef` sites now set `text`/`slackUserId`/`hasThread`, and a `!ref.text` guard refuses what remains (`slack.ts:1108`, `:2078`, `:2589`).
- ✅ **Fixed:** `ref.handedOff` is now read in the hand-off guard, so one turn cannot be handed on twice (`slack.ts:2583`).
- ✅ **Fixed:** A failed hand-off relay now tells the sender — the `.catch` posts an ephemeral naming both agents (`slack.ts:2663`).
- ✅ **Fixed:** `resolveRoster` no longer costs a full agent projection per event — a 30-second name cache with expiry sweeping fronts it (`slack.ts:915`).
- ✅ **Fixed:** The three doc claims now match the code — "at most one" default, the unresolved-name sentence narrowed to the ambiguous case, and `default` added to all three command enumerations (`docs/architecture/channels.md:15`, `:82`, `:124`, `:263`).
- ✅ **Fixed:** The PR body's Ambient and Migrations paragraphs now describe the shipped no-promotion rule.
- ✅ **Fixed:** The default flag is now its own non-truncating `Badge` beside the channel id, not text concatenated into a truncated line (`slack-channel-card.tsx:104`).
- ✅ **Fixed:** The unused `defaultOf` import is gone (`slack.ts:102`).
- ✅ **Fixed:** The UI ambient toggle can no longer silently claim a vacant default — the claim is now inside the connect transaction and gated on `existing` (`agents-service.ts:547`).
- 🆕 **New:** Disconnect-confirm copy renders literal backticks (`slack-channel-card.tsx:86`).
- 🆕 **New:** Disconnect-confirm copy names a command form that does not work (`slack-channel-card.tsx:86`).
- 🆕 **New:** `claimDefaultIfVacant` is the only `channelsTxRepo` member wired without `owner` (`compose.ts:107`).

### Findings
- 🟡 **Warning:** The disconnect-confirm copy escapes backticks, but the dialog renders `message` as a ReactNode with no markdown, so users read literal backtick characters (`packages/ui/src/modules/sandboxes/components/channels/slack-channel-card.tsx:86`). Every sibling `showConfirm` call in this repo marks up code with JSX instead. Reported by react-ui-engineering.
  **Fix:** In any `showConfirm`/`showAlert` copy, mark code spans with a JSX element such as `<code>`; never markdown backticks, because `dialog-overlay.tsx` passes the value straight to `description`.
- 🟡 **Warning:** The same copy calls the affordance the "in-chat `default` command", but every Slack-side message spells it `/<brand> default <agent>` (`slack.ts:1584`, `:1921`) — the instruction fails as typed (`packages/ui/src/modules/sandboxes/components/channels/slack-channel-card.tsx:86`). Reported by react-ui-engineering.
  **Fix:** Quote a chat command in its full invocable form everywhere the UI names one, taking the prefix from `getBrand().short` as `slack-bind-view.tsx:24` does.
- 🟡 **Warning:** `claimDefaultIfVacant` is the only `channelsTxRepo` member composed without `owner`, unlike its `upsertChannel` and `listByAgent` siblings (`packages/api-server/src/modules/agents/compose.ts:107`). Today's single caller checks ownership first, so a later caller is what this exposes to an unfiltered write. Reported by typescript-engineering.
  **Fix:** Scope every channel-binding repository member by owner at the composition root — give `claimSlackDefaultIfVacantTx` an `owner` parameter and close over it in `compose.ts`, matching the two members beside it.
- 🟢 **Suggestion:** `noDefaultYet` writes the table name `channels` and the literal `'slack'` into raw SQL, the only place in this repository that does; a schema rename then breaks the default claim at runtime rather than at compile time (`packages/api-server/src/modules/agents/infrastructure/channel-bindings-repository.ts:198`).

- 🟢 **Suggestion:** The messenger-parity sentence still lists Slack's in-chat surface as `bind` and `unbind` only, now four commands behind the code (`docs/architecture/channels.md:257`). Pre-existing — `ambient` already made it stale before this PR, and the line sits outside this diff — so it belongs in its own issue rather than this round.

### Issue Fit
✅ No findings.

- 🟡 **Warning:** Merged into `### Findings` above — `claimDefaultIfVacant` composed without `owner`.
- 🟢 **Suggestion:** Merged into `### Findings` above — raw SQL hardcodes the `channels` table name.
- 🟢 **Suggestion:** `slack-routing.ts` holds dependency-free routing rules under `infrastructure/`, while ten sibling modules keep such code in `domain/`. Placement only — the channels module has no `domain/` directory yet, and four other pure-logic files in the same folder share the pattern, so this is a module-wide cleanup, not this PR's defect (`packages/api-server/src/modules/channels/infrastructure/slack-routing.ts:38`).
- 🟢 **Suggestion:** `agents/index.ts` adds four infrastructure exports, of which only `setSlackChannelDefault` has a non-test consumer outside the module (`packages/api-server/src/modules/agents/index.ts:52`). Dropping the other three narrows the module's published contract at no cost.

- 🟡 **Warning:** Merged into `### Findings` above — literal backticks in the disconnect-confirm copy.
- 🟡 **Warning:** Merged into `### Findings` above — the copy names a non-working command form.

### Verdict
COMMENT — every blocking finding from the previous round is fixed at its cause; what remains is two lines of disconnect-confirm copy and one composition-root asymmetry.

---

packages/ui/src/modules/sandboxes/components/channels/slack-channel-card.tsx
```diff
@@ -80,7 +81,11 @@ function SlackChannelRow({
   const handleDisconnect = async () => {
     if (
       await showConfirm(
-        `Mentions in ${channel.slackChannelId} will stop reaching this sandbox.`,
+        `Mentions in ${channel.slackChannelId} will stop reaching this sandbox.${
+          channel.default
+            ? ` It is that channel's default agent, so mentions with no agent name will reach no one until an agent's owner claims the default with the in-chat \`default\` command.`
```
dam-code-guardian: 🟡 **Warning:** This copy escapes backticks, but `dialog-overlay.tsx:15` passes `message` straight to `description` as a ReactNode — no markdown runs, so users read literal backtick characters. Every sibling `showConfirm` call in this repo marks code up with JSX instead.
  **Fix:** In any `showConfirm`/`showAlert` copy, mark code spans with a JSX element such as `<code>`; never markdown backticks.

🟡 **Warning:** The same line calls the affordance the "in-chat `default` command", but every Slack-side message spells it `/<brand> default <agent>` (`slack.ts:1584`, `:1921`). The instruction fails as typed.
  **Fix:** Quote a chat command in its full invocable form wherever the UI names one, taking the prefix from `getBrand().short` as `slack-bind-view.tsx:24` does.

Reported by react-ui-engineering.

packages/api-server/src/modules/agents/compose.ts
```diff
@@ -103,8 +104,9 @@ export function composeAgentsModule(deps: {
         upsertChannel: (tx, agentId, channel) =>
           upsertChannelTx(tx, owner, agentId, channel),
         listByAgent: (tx, agentId) => listChannelsByAgentTx(tx, owner, agentId),
+        claimDefaultIfVacant: claimSlackDefaultIfVacantTx,
```
dam-code-guardian: 🟡 **Warning:** `claimDefaultIfVacant` is the only `channelsTxRepo` member composed without `owner` — the two lines above it both close over one. Today's single caller checks ownership first, so what this exposes is the next caller, which would reach an unfiltered write.
  **Fix:** Scope every channel-binding repository member by owner at the composition root — give `claimSlackDefaultIfVacantTx` an `owner` parameter and close over it here, matching its siblings.

Reported by typescript-engineering.

packages/api-server/src/modules/agents/infrastructure/channel-bindings-repository.ts
```diff
@@ -166,28 +176,91 @@ export function setSlackChannelAmbient(db: Db) {
...
+  return async (agentId: string, slackChannelId: string): Promise<void> => {
     await db
       .delete(channels)
+      .where(
+        and(eq(channels.agentId, agentId), slackChannelIs(slackChannelId)),
+      );
+  };
+}
+
+const noDefaultYet = (slackChannelId: string) =>
```
dam-code-guardian, line 198: 🟢 **Suggestion:** This is the only raw SQL in the repository that writes a table name (`channels`) and the `'slack'` literal by hand. A schema rename then breaks the default claim at runtime instead of at compile time — the Drizzle identifiers used on every other line of this file would fail the build instead.

packages/api-server/src/modules/channels/infrastructure/slack-routing.ts
```diff
@@ -0,0 +1,92 @@
...
+  if (!body.startsWith(name)) return false;
+  const rest = body.slice(name.length);
+  return rest.length === 0 || NAME_TERMINATOR.test(rest);
+}
+
+export function defaultOf(roster: RosterEntry[]): RosterEntry | null {
+  return roster.find((entry) => entry.isDefault) ?? null;
+}
+
+export function orderAmbientReaders(
```
dam-code-guardian, line 38: 🟢 **Suggestion:** These routing rules have no dependencies, while ten sibling modules keep such code in `domain/`. Placement only — the channels module has no `domain/` directory yet and four other pure-logic files share this folder, so it is a module-wide cleanup rather than this PR's defect.

packages/api-server/src/modules/agents/index.ts
```diff
@@ -44,9 +44,13 @@ export { buildAppendAgentsMdCommand } from "./domain/agents-md.js";
 export {
   deleteChannelsByAgent,
   listChannelsByOwner,
-  findBySlackChannelId,
+  findSlackBindingsByChannelId,
   findSlackChannelsByAgent,
   deleteSlackChannelByAgent,
   deleteSlackChannelBinding,
   setSlackChannelAmbient,
+  claimSlackDefaultIfVacantTx,
```
dam-code-guardian: 🟢 **Suggestion:** Of the four exports added here, only `setSlackChannelDefault` has a non-test consumer outside this module. Dropping the other three keeps the module's published contract narrower at no cost.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `26c15fe`

## PR #3367: feat(channels): several agents can share one Slack conversation
**Author:** pilartomas | **Branch:** feat/multi-agent-slack-channels → main | **Changes:** +7936 −268 (47 files)

### Summary
Complete re-review at the live HEAD, requested by the label. The one new commit fixes all three of the previous round's warnings at their cause, and nothing blocking remains.
_(Suppressed 1 finding per PR context: issue-fit's no-promotion scope conflict — the body's "No default is a supported state" section justifies the decision explicitly.)_

### Changes since last review
Previous HEAD: `bdd5335` (2026-08-19T13:34:15Z) — verdict COMMENT

- ✅ **Fixed:** The disconnect-confirm copy no longer renders literal backticks — the message is now a JSX `<p>` with a real `<code>` element, which `ConfirmDialog` passes through `AlertDialogDescription asChild` (`slack-channel-card.tsx:87`).
- ✅ **Fixed:** The copy now names the command in its invocable form, `/{brandShort} default <agent>`, taken from `getBrand().short` — matching the three server-side spellings (`slack-channel-card.tsx:94`).
- ✅ **Fixed:** `claimDefaultIfVacant` is composed with `owner` like its two siblings, and `claimDefault` filters on `channels.owner`, so the write can no longer reach another owner's row (`compose.ts:107`, `channel-bindings-repository.ts:220`).
- 🔁 **Still present:** Pre-existing messenger-parity line lists two in-chat commands (`docs/architecture/channels.md:257`) — from doc-drift, unchanged and outside this diff.

### Findings
- 🟢 **Suggestion:** The default-agent warning and the multi-agent mention explainer now exist near-verbatim in both the UI and the server, so the two explanations of one routing rule can drift apart (`slack-channel-card.tsx:93` against `slack.ts:1536`; `slack-bind-view.tsx:157` against `agents-service.ts:337`). Reported by react-ui-engineering.
- 🟢 **Suggestion:** `noDefaultYet` reaches its column names through `sql.raw(channels.type.name)` against a hand-written alias, the one predicate in the file not built with the typed builder; `alias()` would express it typed but is not re-exported from the `db` package (`channel-bindings-repository.ts:198`). Reported by typescript-engineering.

- 🟢 **Suggestion:** 🔁 Carried from the previous review at `bdd5335` — the messenger-parity sentence still lists Slack's in-chat surface as `bind` and `unbind` only (`docs/architecture/channels.md:257`). Verified unchanged from base and already stale before this PR, so it belongs in its own issue.

### Issue Fit
✅ No findings.

- 🟢 **Suggestion:** Merged into `### Findings` above — raw identifier interpolation in `noDefaultYet`.
- 🟢 **Suggestion:** Reported as an owner-scope inconsistency between the new `eq(channels.owner, owner)` predicate and the global `noDefaultYet` probe; verified as intentional and correct, so recorded here rather than as a warning. The two encode different questions — which row this caller may mutate (per owner) versus whether the conversation already has a default (global, matching `channels_slack_default_agent_idx`, which is keyed on `slackChannelId` alone). A foreign default therefore *should* make the claim return `false`, and no `channel.default_changed` log is owed when nothing was claimed (`channel-bindings-repository.ts:198`).
- 🟢 **Suggestion:** Eight sibling repository factories stay exported from `agents/index.ts` and are consumed by `bootstrap.ts`; this commit retracted two of them. Pre-existing module-boundary cleanup, larger than this PR (`packages/api-server/src/modules/agents/index.ts:44`).

- 🟢 **Suggestion:** Merged into `### Findings` above — duplicated routing copy between UI and server.
- 🟢 **Suggestion:** The channel-card empty state still says mentions of the bot "drive this sandbox", which the multi-agent model narrows to the default agent (`slack-channel-card.tsx:44`). Pre-existing line, worth the same follow-up as the rest of the copy pass.

### Verdict
APPROVE — the three warnings from the previous round are fixed at their cause, and only non-blocking suggestions remain.

---

Issue #3317, opened 2026-08-13
UI - A Home page: blockers, compute, what you missed, and what's running


jamiejabbouribm:
## Context

DAM has no place that answers "what's going on?". The closest thing is `/inbox`, which is a single-purpose queue of pending egress and tool approvals — it says what is *waiting on a decision* and nothing else. Everything else a user needs on arrival is spread across pages: whether anything crashed, which sandboxes are awake, what finished while they were away, how much compute and money they are spending.

For a new user this is the first screen after the welcome modal (#3214), and today it is a menu of unexplained words. For a returning user it is worse: nothing accumulates. Come back on Monday and the product has no memory of what happened over the weekend, so the only way to find out is to open every sandbox in turn.

A design for this page exists and is ready to attach.

## Problem

A user arriving in DAM cannot see, without navigating:

- **What is blocking them.** Approvals are on `/inbox`, but a crashed pod, an over-budget sandbox and an expired connection are each visible only on the sandbox that owns them. All four are the same user need — "something needs me" — and they live in four places.
- **How much compute they are consuming.** Sandbox size is set per sandbox and never summed. There is no owner-wide view of CPU/memory or spend, so the first sign of a budget problem is a sandbox refusing to start.
- **What they missed.** Nothing records what happened while they were gone. Finished sessions, returned invocations, schedule runs, failed jobs — all of it has to be hunted for, per sandbox, and there is no notion of unseen.
- **What is running right now.** The sandbox list shows pod state, so a sandbox that is merely awake looks identical to one mid-turn. "Is it still working or is it waiting on me?" is unanswerable from the list.

The result is that DAM has an inbox for the one thing that interrupts you and no home for everything else.

## Goal

A Home page — replacing `/inbox` as the landing surface — where a user sees, in one screen and without clicking: what is blocked on them, what is running, what finished since they were last here, and their compute and spend at a glance. Every item is actionable from the page. Arriving after a weekend away tells you what happened, not just what is stuck.

## User Stories

- As a returning user, I want everything that needs a decision from me in one list — approvals, crashes, expired credentials, over-budget sandboxes — so that I don't check four pages to learn nothing is wrong.
- As a returning user, I want to see what completed while I was away, marked as unseen until I look, so that finished work doesn't sit unnoticed.
- As someone with several sandboxes, I want to know which are actually working versus merely awake, so that I can tell waiting-on-the-agent from waiting-on-me.
[...]

Issue #3233, opened 2026-08-07
UI - Updated left navigation + new coding agents page


jenna-winkler:
Scope:
1. New dedicated page for "Coding agents"
2. Move artifacts lower in left nav
3. Make the left nav expandable

Implement alongside #3234

jamiejabbouribm:
Hi all, uploading the same first user experience prototype, you can see in here the following changes made to the left navigation:

- "Sandboxes" renamed to "Coding agents": Updated label in the sidebar nav to better describe what the feature actually is.
- Collapsible sidebar: The nav can be toggled open/closed via the icon that appears when hovering DAM, or when clicking any white space on the left bar.
- Artifacts moved to bottom of bar: Sits lower in the nav hierarchy since it's output from agents rather than a primary workflow entry point.

Prototype:

[dam-prototype-final.zip](https://github.com/user-attachments/files/31040161/dam-prototype-final.zip)

Screenshots:

<img width="233" height="845" alt="Image" src="https://github.com/user-attachments/assets/fb127b88-a2ab-4573-b58c-223b79dac62e" />

<img width="268" height="842" alt="Image" src="https://github.com/user-attachments/assets/d678e21c-89d0-4d71-a073-4d5ef357cfac" />

jenna-winkler:
ready for dev @kapetr cc @tomkis

jamiejabbouribm:
Hi @kapetr, can we make a minor adjustment here and use the left and right facing chevron icons for the open and close of the menu? @jenna-winkler has brought up a good point that the x makes it seem like you can close out the entire window, but you are simply collapsing it. 

Screenshots below. 

Closed:
<img width="230" height="476" alt="Image" src="https://github.com/user-attachments/assets/3c7f757e-a6c3-44d6-8e7c-9a0161019a59" />

Hover:
<img width="248" height="446" alt="Image" src="https://github.com/user-attachments/assets/23bbcc17-8efa-4b12-93c0-b2b21c201614" />

Open:
<img width="508" height="470" alt="Image" src="https://github.com/user-attachments/assets/4454292c-152b-4183-888e-9964a77403c4" />

jenna-winkler:
cc @tomkis @PetrBulanek since I believe Petr k is now out on vaca

PetrBulanek:
@jenna-winkler I’m actually on vacation until September 3. @kapetr is only out Thursday and Friday and should be back on Monday. :-)

jenna-winkler:
[...]

PR #3369, opened 2026-08-19
feat(ui): activity feed, compute, spend to home

kapetr:
A Home page a user can arrive at and read in one screen: what needs a decision from them, and what is still working. A feed down the left carries pending approvals, running sessions and unread sessions; contextual widgets down the right show allocated compute, spend, and the schedules across every sandbox. Approvals become actionable from wherever the user happens to be, via a floating pill on every other page.

`/inbox` — today's single-purpose approval queue and landing surface — folds into Home.

**Scoped out of this PR.** The issue asks for everything needing a decision in one list — approvals, crashes, expired credentials and over-budget agents. This ships the approvals class only, matching the prototype the design handoff settled on; the other three need signals the feed has no reader for yet. The issue stays open for them, so this references it rather than closing it. The design handoff's fifth feed filter, "Read", is also out — the feed carries no read state to filter on.

**Beyond the UI.** `acp_native` approval rows are keyed by session as well as agent and rpc id, and a settled id is reclaimed on insert. Without it a resumed session — whose harness restarts its rpc counter while its session id survives — silently dropped every approval it raised, so nothing reached Home, the pill, or the inbox this PR retires. An in-session answer now settles the row the relay recorded, rather than any pending row sharing that rpc id.

The feed shows current state, not history: a hibernated sandbox has nothing to show, and nothing here needs a stored activity log. One known limitation is deliberate — read state lives in the sandbox pod, so unread is knowable only for running sandboxes. Covering hibernated ones would mean tracking read state per user in Postgres, which is a separate decision.

Refs #3317

## Sub-issues

- [x] 01 — Home module, two-column shell, feed spine
- [x] 02 — Feed filtering
- [x] 03 — Inline approval cards
- [x] 04 — Dismiss, clear all, and `platform/markSeen`
- [x] 05 — Compute and spend widgets
- [x] 06 — Owner-wide schedule list
- [x] 07 — Schedules widget
- [x] 08 — Floating approvals pill
- [x] 09 — Retire `/inbox`

Builds on #3352, which retired the creation wizard and owns the Home view, the icon rail and the routes this feature edits. Now that it has merged, this branch sits directly on `main`.


packages/api-server/src/modules/schedules/infrastructure/schedules-repository.ts
```diff
@@ -58,6 +59,18 @@ function rowToSchedule(row: InternalRow): Schedule {
 
 export function createSchedulesRepository(db: Db): SchedulesRepository {
   return {
+    async listForOwner(owner, limit): Promise<Schedule[]> {
```
dam-code-guardian: 🔴 **Critical:** Reported by typescript-engineering. The `LIMIT` runs in SQL across every sandbox of the owner, ordered by `next_run`, and the binding filter runs afterwards in the service (`schedules-service.ts:56`). So the limit spends its budget on rows the caller may not see.

Replayed against the real predicates: with 5 other-agent schedules ordered ahead and a caller bound to an agent owning 2, `limit=5` returns `[]`. `limit` is optional on `scheduleListForOwnerInputSchema`, and the UI currently omits it — but the field is on the public tRPC input, so any client can reach this.

**Fix:** An authorization filter belongs in the same query as its `LIMIT`. Pass the binding into `listForOwner` and add the `agent_id` restriction to the `WHERE` clause, so the limit counts only rows the caller may see. The approvals module already reads this way.

```suggestion
    async listForOwner(owner, limit, binding): Promise<Schedule[]> {
      const scope =
        binding === "*"
          ? eq(schedulesTable.owner, owner)
          : and(
              eq(schedulesTable.owner, owner),
              inArray(schedulesTable.agentId, [...binding]),
            );
      const query = db
        .select()
        .from(schedulesTable)
        .where(scope)
        .orderBy(asc(schedulesTable.nextRun), asc(schedulesTable.createdAt));
```

packages/api-server/src/modules/approvals/services/approvals-relay-service.ts
```diff
@@ -36,13 +41,22 @@ export interface CreateApprovalsRelayServiceDeps {
   bus: RedisBus;
 }
 
+async function onlyPendingMatch(
+  deps: CreateApprovalsRelayServiceDeps,
+  agentId: string,
+  rpcId: number | string,
+): Promise<PendingApprovalRow | null> {
+  const rows = await deps.repo.findPendingAcpNativeByRpcId(agentId, rpcId);
```
dam-code-guardian: 🔴 **Critical:** Reported by typescript-engineering. `acpNativeRowId` now includes `sessionId` (`domain/ids.ts:1`), but this fallback still identifies a row by `(agentId, rpcId)` only. ACP rpc ids are per-connection counters, so two sessions of one agent ordinarily collide on the same id.

The relay reaches the fallback whenever `mirroredRows` has no entry for `msg.id` — after an api-server restart, on a second relay replica, or on a reconnect (`acp-relay.ts:199`). Replayed both branches:

- Only session A's row pending, session B answers `rpcId 7` → **A's approval is marked `allow_once`**. B's tool call proceeds on a decision the user made for a different session.
- Both rows pending → `rows.length === 1` fails, so the resolve silently no-ops and the user's real answer is dropped.

The per-connection `rowId` path is correct; this is the path that survives when it is unavailable.

**Fix:** Resolve a mirrored approval by a key containing every field of its row id. Thread the response frame's `sessionId` into the lookup and add `AND session_id = $sessionId`, so the fallback identifies exactly one row instead of depending on how many are open.

packages/ui/src/modules/home/lib/feed-item.ts
```diff
@@ -0,0 +1,85 @@
+import type { ApprovalView } from "api-server-api";
+
+import type { SessionView } from "../../../types.js";
+import { isUnreadSession } from "./unread.js";
+
+export type FeedItem =
```
dam-code-guardian, line 6: 🔴 **Critical:** Reported by issue-fit. `FeedItem` has three kinds and only `approval` answers the "Needs attention" filter. Issue #3317 names four blocker classes and states they are one need: "a crashed pod, an over-budget sandbox and an expired connection are each visible only on the sandbox that owns them. All four are the same user need — 'something needs me' — and they live in four places." Its first user story repeats the list.

Nothing in the home module reads `overBudget`, a crash state, or connection expiry, so `Closes #3317` closes the issue with three of the four classes undone.

**Fix:** A closing keyword requires every blocker class the issue names to reach the feed. Add the missing item kinds, or drop the keyword and leave #3317 open for them.

packages/ui/src/modules/home/api/queries.ts
```diff
@@ -0,0 +1,68 @@
...
+  const sessions = useQueries({
+    queries: runningAgents.map((agent) => ({
+      queryKey: homeKeys.sessions(agent.id),
+      queryFn: () => listAgentSessions(agent.id),
+      refetchInterval: SESSIONS_POLL_MS,
+      staleTime: SESSIONS_STALE_MS,
+      retry: false,
+    })),
+    combine: (results) => ({
+      byAgent: results.map((result) => result.data ?? []),
```
dam-code-guardian, line 49: 🟡 **Warning:** These per-agent session queries set `retry: false`, and `result.data ?? []` turns a failed read into an empty session list. `pending` only tracks `isPending`, so `loadingFeed` is false too.

With running agents present and every session read failing, `noRunningAgents` is false, so `emptyStateFor` returns "All clear — Nothing waiting for review. You're all caught up." (`feed-filter.ts:118`). A user is told nothing needs them because the page could not find out. Same class as the `!isPending` absence claim on PR #3433.

**Fix:** Derive an "all clear" claim from a successful read, never from a non-pending one. Carry an error count out of `combine` and show a load-failure state when any agent's sessions could not be read.

```suggestion
    combine: (results) => ({
      byAgent: results.map((result) => result.data ?? []),
      pending: results.some((result) => result.isPending),
      failed: results.filter((result) => result.isError).length,
    }),
```

packages/api-server/src/modules/schedules/services/schedules-service.ts
```diff
@@ -29,12 +29,21 @@ function asBadRequest(fn: () => void): void {
...
+  agentId: string,
+): boolean {
+  return binding === "*" || binding.includes(agentId);
+}
+
 export function createSchedulesService(deps: {
   repo: SchedulesRepository;
   runner: SchedulerRunner;
   owner: string;
+  agentBinding?: readonly string[] | "*";
```
dam-code-guardian: 🟡 **Warning:** Reported by typescript-engineering. `agentBinding` is optional and line 46 defaults it to `"*"`, so a composition root that omits it silently gets an unfiltered owner-wide read with no type error. `composeSchedulesForOwner` spreads the key only when truthy, and `harness-api-server/app.ts:112` calls it without one.

The approvals module has the same owner-wide read and makes `agentBinding` required on both its service and its compose deps (`approvals/compose.ts:32`) — that is the shape to mirror.

**Fix:** A parameter that narrows an authorization scope is required, never defaulted. Declare it as a required field on the service deps and on the compose options, so every call site states its scope explicitly.

packages/api-server/src/modules/approvals/infrastructure/approvals-repository.ts
```diff
@@ -123,7 +127,24 @@ export function createApprovalsRepository(db: Db): ApprovalsRepository {
...
+            payload: row.payload,
+            expiresAt: row.expiresAt,
+            createdAt: new Date(),
+            status: "pending",
+            resolvedAt: null,
+            verdict: null,
+            decidedBy: null,
+            deliveredAt: null,
+          },
+          setWhere: sql`${pendingApprovals.status} <> 'pending'`,
```
dam-code-guardian: 🟡 **Warning:** Reported by typescript-engineering. `setWhere: status <> 'pending'` lets a resolved row be revived by a later insert with the same id, and the `set` clause clears `resolvedAt`, `verdict`, `decidedBy` and `deliveredAt`. The delivery sweeper reads exactly those fields to send the wrapper its permission frame (`delivery-sweeper.ts:36`), so a resolved-but-undelivered row revived this way never delivers its answer.

The id is now deterministic per `(agentId, sessionId, rpcId)`, which is what makes a repeat insert on the same id reachable.

**Fix:** Exclude rows with undelivered work from a revive predicate.

```suggestion
          setWhere: sql`${pendingApprovals.status} <> 'pending' AND ${pendingApprovals.deliveredAt} IS NOT NULL`,
```

packages/ui/src/components/icon-rail.tsx
```diff
@@ -43,15 +39,12 @@ export function IconRail({
   const navigateToExperiments = useStore((s) => s.navigateToExperiments);
   const navigateToKnowledgeBases = useStore((s) => s.navigateToKnowledgeBases);
 
-  const { data: approvals = EMPTY } = useApprovalsForOwner();
-  const pendingCount = approvals.filter((r) => r.status === "pending").length;
-
   const sandboxes: Destination = {
     label: "Home",
```
dam-code-guardian, line 43: 🟡 **Warning:** Reported by issue-fit. The "Home" rail item now targets the `home` view, and no rail destination reaches `view === "list"` any more. But `/sandboxes` still parses to `list` and still renders `ListView` (`routes.ts:78`, `:118`) — the path was deliberately moved out of `RETIRED_PATHS` to keep it live.

The only remaining way in is an artifacts empty-state action (`artifacts-view.tsx:164`) or a typed URL. #3317 puts the nav restructure out of scope as #3233, so the route surviving looks intentional; losing its entry point does not.

**Fix:** A route that survives a change keeps a navigation entry. Add a rail item for the agents list, or retire the route with its view.

packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts
```diff
@@ -493,6 +493,15 @@ export function createAcpRuntime(deps: AcpRuntimeDeps): AcpRuntime {
         return;
       }
 
+      if (method === "platform/markSeen" && paramsSid) {
```
dam-code-guardian, line 496: 🟡 **Warning:** Reported by issue-fit. This handler has no caller anywhere in the tree. The PR body lists sub-issue "04 — Dismiss, clear all, and `platform/markSeen`" as done, but dismiss and "Clear all" write only to browser `localStorage` (`home/lib/dismissals.ts:19`), and read state still comes from the pre-existing engage and turn-end paths.

So dismissals are per-browser: they do not follow the user to another device, and nothing marks a session seen on the pod.

**Fix:** A new wire method needs a caller in the same change. Call `platform/markSeen` from the dismiss path, or remove the handler and drop it from the body's checklist.

packages/ui/src/modules/home/lib/feed-filter.ts
```diff
@@ -0,0 +1,124 @@
+import { SessionType } from "api-server-api";
+
+import type { FeedItem } from "./feed-item.js";
+
+export type FeedStatus = "all" | "attention" | "in-progress" | "unread";
+
+export const FEED_STATUSES: readonly FeedStatus[] = [
```
dam-code-guardian, line 7: 🟡 **Warning:** Reported by issue-fit. The design handoff enumerates five filters — Everything, In progress, Unread, Read, Needs attention — and `FEED_STATUSES` ships four, without "Read". A user who has looked at a session has no way to list what they already read.

**Fix:** Deliver every enumerated filter, or record the dropped one in the PR body as an accepted exclusion, the way the body already does for hibernated-sandbox unread state.

packages/api-server/src/apps/api-server/agent-proxies/acp-relay.ts
```diff
@@ -139,6 +143,9 @@ export function createAcpRelay(
         if (client.readyState === WebSocket.OPEN) client.send(frame);
       });
       client.once("close", () => unsubInjects());
+      client.once("close", () => mirroredRows.clear());
+
+      const mirroredRows = new Map<string, string>();
```
dam-code-guardian: 🟢 **Suggestion:** Reported by typescript-engineering. Two small things on this map, neither live today:

- No entry is deleted after a resolve, so it grows one key per permission prompt for the life of the websocket; only `client.once("close", …)` clears it. Deleting the key in `mirrorPermissionResponse` after the resolve bounds it.
- The `const` is declared *after* the `client.once("close", () => mirroredRows.clear())` handler that captures it. The two statements are adjacent and synchronous, so the temporal dead zone is not reachable now; moving the declaration above the listener removes the ordering hazard.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `b072282`

## PR #3369: feat(ui): activity feed, compute, spend to home
**Author:** kapetr | **Branch:** feat/3317-home-page → main | **Changes:** +2441 −377 (76 files)

### Summary
A Home page with an activity feed, compute, spend and schedules widgets, plus a floating approvals pill, retiring `/inbox`. The `/inbox` retirement and the route rework are complete. Two defects sit in the new owner-wide and cross-session read paths, and the feed carries one of the four blocker classes the issue names.

### Findings
- 🔴 **Critical:** The owner-wide schedule read applies `LIMIT` before the agent-binding filter, so a bound caller can get an empty list while its own schedules exist (`packages/api-server/src/modules/schedules/infrastructure/schedules-repository.ts:62`)
- 🔴 **Critical:** A mirrored ACP approval can be resolved from the wrong session when the relay's in-memory correlation is gone (`packages/api-server/src/modules/approvals/services/approvals-relay-service.ts:49`)
- 🔴 **Critical:** The feed carries only approvals, while issue #3317 names four blocker classes as one need (`packages/ui/src/modules/home/lib/feed-item.ts:6`)
- 🟡 **Warning:** A failed session read is indistinguishable from an idle agent, so Home says "You're all caught up" (`packages/ui/src/modules/home/api/queries.ts:47`)
- 🟡 **Warning:** `agentBinding` is optional and defaults to `"*"`, so a composition root that omits it gets an unfiltered owner-wide read with no type error (`packages/api-server/src/modules/schedules/services/schedules-service.ts:43`)
- 🟡 **Warning:** The revive predicate on `insertPending` can overwrite a resolved-but-undelivered approval, so its wrapper never gets the permission frame (`packages/api-server/src/modules/approvals/infrastructure/approvals-repository.ts:146`)
- 🟡 **Warning:** The agents list keeps its route but loses its only navigation entry (`packages/ui/src/components/icon-rail.tsx:43`)
- 🟡 **Warning:** `platform/markSeen` ships with no caller anywhere in the tree (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:496`)
- 🟢 **Suggestion:** `mirroredRows` never deletes a resolved entry, so it grows for the life of the websocket connection; and its `const` is declared after the `close` handler that captures it (`packages/api-server/src/apps/api-server/agent-proxies/acp-relay.ts:148`).
- ✅ **Looks good:** The cross-session approval fix in `b3d6eaf1` is real. Replaying the mirror path with two live sessions of one agent at the same `rpcId`, the old `(agentId, rpcId)` resolve marked session A's row from session B's click; the per-connection `rowId` resolves each session's own row.

- 🟡 **Warning:** Three architecture pages still describe the pre-Home system. `platform-topology.md:75` enumerates the UI's remaining polls, and this PR adds a 30-second owner-approvals poll; `metrics.md:7` names the only consumers of the spend breakdown, and the Home spend widget is a third; `metrics.md:54` says such surfaces withdraw the period control when the backend is off, while the Home widget hides itself; `agent-lifecycle.md:93` presents schedules as a per-agent read (`docs/architecture/platform-topology.md:75`).
  **Fix:** When a change alters a subsystem's behaviour, update that page in the same PR and bump its `Last verified`. Correct each enumeration the change invalidates — remaining polls, consumers of a read path, and the scope of a query surface.

### Issue Fit
Linked issue: #3317 (closing keyword). The shell, filtering, approval cards, compute, spend, schedules, the pill and the `/inbox` retirement all match. Three items are short.

- The single-blocker-class gap is the 🔴 above, and the `markSeen` and rail findings are the 🟡 entries above.
- 🟡 **Warning:** The design handoff names five feed filters — Everything, In progress, Unread, Read, Needs attention — and `FEED_STATUSES` ships four, without "Read" (`packages/ui/src/modules/home/lib/feed-filter.ts:7`).
  **Fix:** Deliver every enumerated filter, or record the dropped one in the PR body as an accepted exclusion.
- 🟡 **Warning:** The PR reshapes the `acp_native` approval row identity and its resolution matching, which neither #3317 nor the body declares (`packages/api-server/src/modules/approvals/domain/ids.ts:1`).
  **Fix:** Declare a substantial backend change in the PR body, or move it to its own PR.
- 🟢 **Suggestion:** The design handoff lists knowledge-base cards among the feed card types; `sessionIcon` covers schedules, experiments and code, and falls back to the code icon (`packages/ui/src/modules/home/components/feed-list.tsx:23`).
- 🟢 **Suggestion:** The inbox-to-Home rename leaves two statements naming the inbox as the approvals surface (`docs/notes/v0.2-acceptance-criteria.md:35`, `:40`).

- The two backend 🔴 entries, the `agentBinding` default, the revive predicate, the route-table duplication and the `mirroredRows` items above all came from this skill; both criticals were confirmed by replaying the real predicates.
- 🟡 **Warning:** `isPermissionResponse` accepts any response frame whose `result.outcome.outcome` is a string, and the relay then resolves an approval for that frame's `msg.id`, so an unrelated response carrying the same nested field can resolve a mirrored approval (`packages/api-server/src/apps/api-server/agent-proxies/acp-relay.ts:58`).
  **Fix:** Correlate a response to its request by recorded id, never by payload shape. Return early when `mirroredRows` has no entry for `msg.id`.
- 🟡 **Warning:** The `limit` reaches SQL with no clamp in the repository — only the router's schema caps it at 200, and the MCP and harness surfaces bypass that schema (`packages/api-server-api/src/modules/schedules/schemas.ts:15`). The approvals repository clamps in the repository itself.
  **Fix:** Bound a query size in the repository that issues the query, not only in one caller's schema.
- 🟡 **Warning:** `RETIRED_PATHS` and `parseRoute` now disagree about who redirects a retired path, leaving the `parseRoute("/inbox")` branch unreachable for every non-browser caller (`packages/ui/src/modules/platform/lib/routes.ts:46`).
  **Fix:** Keep a retired-path set and its parse branch in one place, so one function decides where a retired path goes.

### Verdict
REQUEST_CHANGES — the owner-wide schedule read can hide a caller's own schedules, a mirrored approval can be resolved from the wrong session, and the feed delivers one of the four blocker classes the closed issue names.

---

tomkis (approved):

packages/api-server/src/modules/approvals/infrastructure/approvals-repository.ts
```diff
@@ -123,7 +123,24 @@ export function createApprovalsRepository(db: Db): ApprovalsRepository {
...
+            payload: row.payload,
+            expiresAt: row.expiresAt,
+            createdAt: new Date(),
+            status: "pending",
+            resolvedAt: null,
+            verdict: null,
+            decidedBy: null,
+            deliveredAt: null,
+          },
+          setWhere: sql`${pendingApprovals.status} = 'expired' OR (${pendingApprovals.status} = 'resolved' AND ${pendingApprovals.deliveredAt} IS NOT NULL)`,
```
dam-code-guardian, line 142: 🔴 **Critical:** Reported by typescript-engineering. Narrowing `setWhere` correctly stops a resolved-but-undelivered row being revived, but the refusal never reaches the caller. `insertPending` resolves the same way whether the row was written or not, so `recordAcpNativePending` returns the row id and emits `ApprovalRequested` regardless (`approvals-relay-service.ts:46-70`). The relay then maps the new rpc id onto a row still holding the previous tool call's payload and verdict, and the delivery sweeper answers the new permission request with that old decision (`delivery-sweeper.ts:38-46`).

The reachable path is the one the PR body describes: a resumed session keeps its `sessionId` while its harness restarts the rpc counter, so `acpnative:<agent>:<session>:<rpc>` collides. Modelled the upsert with the new predicate — for a prior `resolved`/`deliveredAt: null` row the write is blocked while `record()` reports success, and the sweeper then replies to the new `read README.md` request with the earlier call's `allow_once`. A fresh id inserts and settles against the new payload (control).

**Fix:** a conditional upsert whose write can be refused must report that refusal — add `RETURNING id` to `insertPending` and have `recordAcpNativePending` return `null` and skip the emit when the row is not the caller's, so a stale row is never mirrored or delivered against.

packages/api-server/src/apps/api-server/agent-proxies/acp-relay.ts
```diff
@@ -183,7 +196,10 @@ export function createAcpRelay(
       }
 
       function mirrorPermissionResponse(msg: JsonRpcResponse): void {
-        const rowId = acpNativeRowId(agentId, msg.id);
+        const key = String(msg.id);
```
dam-code-guardian, line 204: 🟡 **Warning:** Reported by typescript-engineering. Correlating by the recorded row id is the right fix, but it replaced the frame validation rather than joining it. `isResponse` accepts `error !== undefined` (line 64), and the deleted `isPermissionResponse` was the only thing that required a `result.outcome.outcome`. So any response frame carrying a mirrored rpc id now settles the row, and `resolveAcpNativeFromInSession` records `allow_once` — including a JSON-RPC error reply.

Ran both predicates over the same frame set: the error frame and a `result: {}` frame went from ignored on `main` to settled `allow_once` here; a frame with an unmirrored id is correctly ignored either way.

**Fix:** keep both halves — require the id to be in `mirroredRows` *and* the frame to be a successful `result` carrying an `outcome`; treat an `error` frame as a cancellation, not a grant.

packages/ui/src/modules/home/api/queries.ts
```diff
@@ -0,0 +1,71 @@
...
+    })),
+  });
+
+  return {
+    items,
+    agents,
+    runningAgents,
+    hasAgents: agents.length > 0,
+    loadingAgents: agentsQuery.isPending,
+    loadingFeed: approvals.isPending || sessions.pending,
```
dam-code-guardian, line 69: 🟡 **Warning:** `unreadableAgents` covers the per-agent session reads, but the approvals query has no equivalent. `loadingFeed` reads only `approvals.isPending`, and `approvals.data ?? []` (line 55) turns a failed read into an empty list, so a failed approvals fetch renders "All clear — Nothing waiting for review. You're all caught up." on the Needs-attention filter. The `errorToast` on `useApprovalsForOwner` fires once, and is suppressed while the API health snapshot is `reconnecting` — which is exactly when the read fails.

**Fix:** for every `isPending`/`?? []` pair, give the error arm its own render distinct from pending and resolved-empty — thread `approvals.isError` into `Feed` and let `emptyStateFor` produce a "could not load" state, never a reassurance.

packages/ui/src/components/floating-approvals-pill.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const { isDismissed, dismiss } = useDismissals();
+  const [expanded, setExpanded] = useState(false);
+  const now = useNow(TICK_MS);
+  const banner =
+    useSyncExternalStore(subscribeApiHealth, getApiHealthSnapshot) !==
+    "connected";
+
+  const items = useMemo(() => {
+    const pending: FeedItem[] = (data ?? [])
+      .filter((approval) => approval.status === "pending")
```
dam-code-guardian, line 31: 🟡 **Warning:** Both new approval surfaces filter to `status === "pending"` (here and `home/api/queries.ts:55`), while the `InboxView` this PR deletes rendered every row through `ApprovalsList`, which labelled `expired` as "timed out" and showed the `expiredNote` guidance that an `ext_authz` timeout needs "Allow permanently" to unblock retries. `FeedApprovalCard` still carries that `expiredNote` branch (`feed-approval-card.tsx:79`), so it is now unreachable, and a timed-out egress approval has no surface at all — `useApprovalsForAgent` in the sessions sidebar also filters to pending.

**Fix:** carry every state a deleted surface rendered — admit non-`pending` approvals into the feed or the pill with their status label and the `expiredNote`, or keep one list for them.

packages/ui/src/modules/home/lib/feed-filter.ts
```diff
@@ -0,0 +1,138 @@
...
+
+export function emptyStateFor(
+  status: FeedStatus,
+  options: {
+    allSourcesExcluded: boolean;
+    noRunningAgents: boolean;
+    unreadableAgents?: number;
+  },
+): FeedEmpty {
+  if (options.unreadableAgents) {
```
dam-code-guardian: 🟡 **Warning:** The `unreadableAgents` branch sits above `allSourcesExcluded`, so a user who unchecks every source while one agent is unreachable reads "Some agents did not answer" and cannot tell that their own filter emptied the list.

**Fix:** order empty-state reasons by what the user controls first — user-applied filters before backend degradation — and merge the two facts when both hold.

packages/api-server/src/modules/schedules/infrastructure/schedules-repository.ts
```diff
@@ -1,10 +1,29 @@
...
+  asc,
+  eq,
+  inArray,
+  type Db,
+  schedules as schedulesTable,
+} from "db";
 import type { Schedule, ScheduleSpec } from "api-server-api";
 import { scheduleSpecSchema } from "api-server-api";
 
+const DEFAULT_LIMIT = 200;
```
dam-code-guardian, line 13: 🟡 **Warning:** Reported by typescript-engineering. Clamping at the query makes the repo safe for every caller, but it also changes the no-argument semantics: `listForOwner(owner)` returned every row on `main` and now stops at 200, with no cursor or overflow signal. `useOwnerSchedules()` passes no limit and `SchedulesWidget` filters the result further to live agents, so an owner past 200 schedules sees a quietly short list and a wrong count in the heading. The two ceilings also disagree — the router schema caps input at 200, `MAX_LIMIT` here is 500.

**Fix:** a read that clamps must expose the truncation — fetch `limit + 1` and return a `hasMore` (or a cursor), and make the router schema's ceiling and `MAX_LIMIT` one shared constant.

packages/ui/src/modules/home/components/schedules-widget.tsx
```diff
@@ -0,0 +1,164 @@
...
+export function SchedulesWidget() {
+  const { data, isError } = useOwnerSchedules();
+  const agents = useAgentsList();
+  const { isPending: agentsPending } = useAgents();
+  const guardEdit = useScheduleEditGuard();
+  const [listOpen, setListOpen] = useState(false);
+  const [editing, setEditing] = useState<Schedule | null>(null);
+  const [creating, setCreating] = useState(false);
+
+  if (isError) return null;
```
dam-code-guardian, line 67: 🟡 **Warning:** Both arms state something definite about an unknown state. `isError` returns `null`, so a failed read is indistinguishable from an owner with no schedules; and while `agentsPending`, `schedules` is forced to `[]`, so the widget says "Nothing scheduled" during load. `SpendWidget` has the same shape (`spend-widget.tsx:27`): `total` falls back to `0` and `spenders` to `[]`, so it renders `$0.00` and "No spend in this period." before its query resolves and when it fails.

**Fix:** give each widget three distinct arms — pending renders a skeleton, error renders a short failure line, resolved-empty renders the empty copy; never fold pending or error into the empty arm.

packages/ui/src/modules/home/components/feed-card.tsx
```diff
@@ -0,0 +1,92 @@
...
+  meta,
+  working = false,
+  unread = false,
+  onOpen,
+  onDismiss,
+  children,
+}: Props) {
+  return (
+    <div
+      role={onOpen ? "button" : undefined}
```
dam-code-guardian, line 32: 🟡 **Warning:** Reported by react-ui-engineering. The card is a `div` with `role="button"` and `tabIndex={0}`, and the Dismiss `<button>` at line 67 sits inside it. A screen reader announces the whole card as one button, and the nested control is not reliably reachable — the `stopPropagation` handles the mouse case only.

**Fix:** keep interactive controls as siblings — leave the card a plain `div` and put the open affordance on its own inner button or link next to Dismiss.

packages/ui/src/components/floating-approvals-pill.tsx
```diff
@@ -0,0 +1,117 @@
...
+              aria-label="Close approvals"
+              className="flex size-7 items-center justify-center rounded-md text-muted-foreground transition-colors hover:bg-muted hover:text-foreground"
+            >
+              <Close size={16} />
+            </button>
+          </div>
+          <div className="max-h-[420px] space-y-2 overflow-y-auto p-2">
+            {items.map((item) =>
+              item.kind === "approval" ? (
+                <FeedApprovalCard
```
dam-code-guardian, line 92: 🟡 **Warning:** Reported by react-ui-engineering. `FeedApprovalCard` sets a `resolved` label, but the pill passes no `onResolved`, so the item leaves `items` on the next 30 s poll and the card unmounts — with no confirmation, and the pill itself disappears if it was the last one. Home avoids exactly this with `useStickyResolved` (`home-view.tsx:147`).

**Fix:** a surface that mutates a polled list must hold the settled row until the user moves on — reuse `useStickyResolved` here so both surfaces confirm the same way.

packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts
dam-code-guardian, line 496: 🟡 **Warning:** Still present from the review at `b0722825`. This handler remains the only occurrence of `platform/markSeen` in any package, dot-directories included; it is absent from `main` and added here. Sub-issue 04 lists it as done, but Home's dismiss and "Clear all" write only the `platform-home-dismissed` localStorage list (`home/lib/dismissals.ts:3`), so seen state does not reach the pod and unread items return on another browser.

**Fix:** ship the caller with the method — send `platform/markSeen` over the ACP channel the way `platform/deleteSession` does — or drop the handler until the per-user store lands.

packages/ui/src/modules/home/components/feed-list.tsx
```diff
@@ -0,0 +1,88 @@
...
+  agents: readonly AgentView[];
+  onOpenSession: (agentId: string, sessionId: string) => void;
+  onDismiss: (item: FeedItem) => void;
+  onResolved: (item: FeedItem) => void;
+}
+
+function sessionIcon(
+  item: Extract<FeedItem, { kind: "unread" | "in-progress" }>,
+) {
+  if (item.session.scheduleId) return <Time size={16} className="shrink-0" />;
```
dam-code-guardian, line 27: 🟡 **Warning:** Still present from the review at `b0722825`. `sessionIcon` branches on `scheduleId` and `experimentId` only, so a knowledge-base session and a Slack or Telegram channel session both render with the coding-agent `Code` icon, and there is no artifact modifier — the design handoff enumerates five card types.

**Fix:** branch on the agent's `kind` and on `session.type`, the same signals `sourceOf` already reads, or declare the remaining card types out of scope in the body.

dam-code-guardian (dismissed):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `2d1528cb`

## PR #3369: feat(ui): activity feed, compute, spend to home
**Author:** kapetr | **Branch:** feat/3317-home-page → main | **Changes:** +2453 −536 (79 files)

### Summary
A Home page with an activity feed, compute, spend and schedules widgets, plus a floating approvals pill, retiring `/inbox`. Both criticals from the previous round are fixed at their cause, and the `/sandboxes` route now redirects instead of rendering a deleted view. Three new findings sit in what the fixes introduce: the narrowed revive predicate now refuses a write whose caller still reports success, and dropping the payload-shape guard admits an error frame as a grant.

### Changes since last review
Previous HEAD: `b072282` (2026-08-24T12:48:51Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** Owner-wide schedule read applies `LIMIT` before the binding filter — the binding is now an `inArray` conjunct in the same query (`packages/api-server/src/modules/schedules/infrastructure/schedules-repository.ts:80`)
- ✅ **Fixed:** Mirrored ACP approval resolvable from the wrong session — the relay passes the full `(agentId, sessionId, rpcId)` row id (`packages/api-server/src/apps/api-server/agent-proxies/acp-relay.ts:161`)
- ✅ **Fixed:** `agentBinding` optional, defaulting to `"*"` — now required on both composition types (`packages/api-server/src/modules/schedules/compose.ts:70`)
- ✅ **Fixed:** Revive predicate could overwrite a resolved-but-undelivered approval (`packages/api-server/src/modules/approvals/infrastructure/approvals-repository.ts:142`)
- ✅ **Fixed:** `isPermissionResponse` matched by payload shape — correlation is now the recorded row id
- ✅ **Fixed:** `limit` reached SQL unclamped — `clampLimit` sits at the query (`schedules-repository.ts:13`)
- ✅ **Fixed:** The agents list kept its route with no nav entry — `/sandboxes` is in `RETIRED_PATHS` and the view is gone
- ✅ **Fixed:** `mirroredRows` never evicted — deleted on settle, on failed record, and on close
- ⛔ **Withdrawn:** `RETIRED_PATHS` vs `parseRoute` — both are load-bearing; without the set, `/sandboxes/new` falls into `sandboxHomeRe`
- 🔁 **Still present:** `platform/markSeen` ships with no caller (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:496`)
- 🔁 **Still present:** Three `/inbox` statements in the v0.2 notes (`docs/notes/v0.2-acceptance-criteria.md:35`)
- 🔁 **Still present:** Knowledge-base and channel sessions share the coding-agent icon (`packages/ui/src/modules/home/components/feed-list.tsx:26`)
- 🆕 **New:** 3 findings below

### Findings
- 🔴 **Critical:** A blocked upsert reports success (`packages/api-server/src/modules/approvals/infrastructure/approvals-repository.ts:142`)
- 🟡 **Warning:** An error response frame settles the approval as `allow_once` (`packages/api-server/src/apps/api-server/agent-proxies/acp-relay.ts:199`)
- 🟡 **Warning:** A failed approvals read renders "You're all caught up" (`packages/ui/src/modules/home/api/queries.ts:68`)
- 🟡 **Warning:** `expired` and `resolved` approvals now have no surface (`packages/ui/src/components/floating-approvals-pill.tsx:31`)
- 🟡 **Warning:** The `unreadableAgents` branch outranks `allSourcesExcluded` (`packages/ui/src/modules/home/lib/feed-filter.ts:92`)
- 🟡 **Warning:** The owner-wide schedule read now truncates at 200 with no signal (`packages/api-server/src/modules/schedules/infrastructure/schedules-repository.ts:13`)
- 🟡 **Warning:** A widget renders a definite figure from an unknown state — `SchedulesWidget` says "Nothing scheduled" while agents load and returns `null` on error; `SpendWidget` shows `$0.00` and "No spend in this period." before its query resolves (`packages/ui/src/modules/home/components/schedules-widget.tsx:66`, `spend-widget.tsx:27`)
- 🟡 **Warning:** Nesting a `<button>` inside a `role="button"` card (`packages/ui/src/modules/home/components/feed-card.tsx:32`)
- 🟡 **Warning:** Resolving an approval from the pill gives no confirmation (`packages/ui/src/components/floating-approvals-pill.tsx:92`)
- 🟡 **Warning:** Deleting `ListView` orphaned the bulk template-update path. `useAgentRows().update.updateAll` now has no caller anywhere, so the "N agents out of date — Update all" banner is gone and `rowProps.updateBusy`'s `update.updatingAll` term is permanently false (`packages/ui/src/modules/agents/hooks/use-agent-rows.ts:82`).
  **Fix:** when a view is deleted, carry every capability it held or remove the capability with it — re-home the outdated-templates banner on `CodingAgentsView`, or delete `updateAll`, `bulkUpgrade` and the `updatingAll` plumbing.
- 🟢 **Suggestion:** `resolveAcpNativeFromInSession` records the verdict `"allow_once"` whatever option the user chose, so a `reject_once` answer is stored as a grant (`packages/api-server/src/modules/approvals/services/approvals-relay-service.ts:78`). Pre-existing on `main`; this commit rewrote the function's signature, so a separate issue reading `outcome.optionId` would fit.
- 🟢 **Suggestion:** `ComputeWidget` reports only the CPU ceiling, while the deleted `BudgetMeter` showed memory against its ceiling too, so a memory-capped owner gets no signal (`packages/ui/src/modules/home/components/compute-widget.tsx:73`).
- 🟢 **Suggestion:** The feed's `status` and `included` filters live in `useState`, so a filtered view is not linkable and the back button does not restore it (`packages/ui/src/modules/home/views/home-view.tsx:39`).

- 🟡 **Warning:** `docs/architecture/knowledge-bases.md:11` still states "The Agents list shows every agent badged with its Kind", and names the owner's inventory of what they pay to run as the reason kinded agents are not hidden. This PR deletes that list; `coding-agents-view.tsx` filters to `!agent.kind`, so no surface lists every agent.
  **Fix:** when a PR removes a UI surface an architecture page uses as its premise, update that claim in the same PR — restate where the full inventory is now visible, or that Kind badging lives only in the per-Kind destinations.
- 🟢 **Suggestion:** `platform-topology.md` describes read state as runtime-stamped from observed viewer engagement, with no mention of the client-initiated `platform/markSeen` this PR adds.

### Issue Fit
Linked issue: #3317, referenced not closed — the feed, widgets, pill and `/inbox` retirement match the ask as re-scoped by the design handoff. The "Read" filter and the `acp_native` row reshape are now declared in the body, so both prior scope findings are closed.
- 🟡 **Warning:** Sub-issue 04 promises `platform/markSeen`, but dismiss and "Clear all" write only a localStorage key list, so seen state does not follow the user to another browser or device (`packages/ui/src/modules/home/lib/dismissals.ts:3`).
  **Fix:** send `platform/markSeen` over the ACP channel the way `platform/deleteSession` does, or drop the handler until the per-user store lands.
- 🟡 **Warning:** The design handoff enumerates card types for knowledge bases, experiments, coding agents, channels and schedules; `sessionIcon` covers schedules and experiments only, so knowledge-base and channel sessions both render as coding agents (`packages/ui/src/modules/home/components/feed-list.tsx:26`).
  **Fix:** branch on the agent's `kind` and on `session.type` — the signals `sourceOf` already reads — or declare the remaining card types out of scope in the body.
- 🟡 **Warning:** Three live statements still name `/inbox` as the approvals surface (`docs/notes/v0.2-acceptance-criteria.md:35`, `:40`, `docs/notes/v0.2-deliverable.md:42`). Pre-existing on `main`, and this PR retires the route.
  **Fix:** name Home (feed card plus floating pill) in all three, matching the renames this PR already made in `ubiquitous-language.md`.
- 🟢 **Suggestion:** The issue asks for spend against a cap; no cap primitive exists, so the widget shows absolute spend only (`packages/ui/src/modules/home/components/spend-widget.tsx:59`). Worth naming in the body so the user story is not read as met.

🔁 8 findings from the previous review are resolved and 1 withdrawn (listed under **Changes since last review**). The new ones — the blocked upsert, the error frame, and the 200-row truncation — are in `### Findings`. Verified clean: `2d1528cb` leaves no dangling `BudgetMeter` reference, and the surviving `modules/budgets/{api,lib}` are still used by `compute-widget.tsx` and `sandbox-size-section.tsx`.

Its findings are merged into `### Findings` above: the expired-approvals gap, the failed-read reassurance, the empty-state ordering, the widget pending/error arms, the orphaned `updateAll`, the lost memory dimension, the `FeedCard` ARIA nesting, and the pill's missing `useStickyResolved`.

### Verdict
REQUEST_CHANGES — a re-inserted approval row whose write the database refuses is still mirrored and delivered against, so a resumed session's new permission request can be answered with the previous call's verdict.

---

PR #3370, opened 2026-08-19
feat(ui): move an existing artifact into a folder

### LINKED ISSUE #3231: UI - Move an existing artifact into a folder
## Context

Artifacts can be organized into folders, but a folder can only be chosen while uploading. Research task #3164 asked whether users want manual folders or automatic grouping by agent; the answer is that users want to organize artifacts themselves, so #3164 is closed and this is the follow-on work it named.

## Problem

A user can create a folder in the Artifacts library, but cannot put anything in it from the UI. The row menu offers Share, Download and Delete — nothing files an artifact anywhere. So a folder made after the fact stays empty, and the library grows into one long flat list.

There is a workaround: ask an agent to move the artifact and it does. That proves the capability exists underneath — the gap is only the UI. It also means the workaround needs a sandbox, a session and the right words, so it does not help a user who is just looking at their library.

## Goal

A user can organize the artifacts they already have, from the library itself. Any artifact can be put into a folder, taken out again, or moved to a different folder.

## User Stories

- As a user, I want to move an existing artifact into a folder so that I can organize work my agent already published.
- As a user, I want to move an artifact between folders so that I can reorganize when my folders change.
- As a user, I want to take an artifact out of a folder so that a wrong choice is easy to correct.
- As a user, I want to move several artifacts at once so that cleaning up a large library is not one-by-one work.

## Open Questions

- Drag and drop, a "Move to folder" menu item, or both? The Files tree already uses drag and drop, so matching it is worth considering.
- Does an artifact live in one folder only, or several?
- Is multi-select worth the first cut, or a follow-up?

PetrBulanek:
Artifacts can be sorted into folders, but a folder can only be chosen while uploading. A user can
create a folder in the library and then has no way to put anything in it, so the folder stays
empty and the library grows into one long flat list. The only workaround is to ask an agent to
move the artifact, which needs a sandbox, a session and the right words.

This makes the library organise itself. Any artifact a user already has can go into a folder, come
out again, or move to a different folder — from the library, with no agent involved. There are two
ways to do it: a **Move to folder…** action in the artifact's own menu, which works everywhere an
artifact is listed, and dragging a row onto a folder in the library for quicker tidying.

Moving several artifacts at once is deliberately left for a follow-up issue.

Closes #3231

## Sub-issues

- [x] 01 — Move to folder in the row menu
- [x] 02 — Drag a library row onto a folder


packages/ui/src/modules/artifacts/components/move-artifact-dialog.tsx
```diff
@@ -0,0 +1,103 @@
...
+    document.addEventListener("keydown", onKey);
+    return () => document.removeEventListener("keydown", onKey);
+  }, [pending, onClose]);
+
+  const userFolders = (folders ?? []).filter(isUserFolder);
+  const currentFolderId = artifact.folderId ?? NO_FOLDER;
+  const isMoveTarget = userFolders.some(
+    (folder) => folder.id === currentFolderId,
+  );
+  const selected = chosen ?? (isMoveTarget ? currentFolderId : NO_FOLDER);
```
dam-code-guardian: 🔴 **Critical:** The folder list is filtered to user folders only, so an artifact that currently sits in an experiment folder resolves `isMoveTarget` to false and pre-selects "No folder". The dialog then misreports the artifact as ungrouped, and one click on **Move** detaches it from its experiment folder: `selected !== currentFolderId` holds, so a real `folderId: null` mutation fires. Reachable from `ExperimentsSection` rows and from the agent-scoped lists in `chat-artifacts-panel.tsx` and `sandbox-artifacts-section.tsx`, which all receive `onMove`. Executing the logic confirms it: an experiment-folder artifact is the only one of the three cases where an untouched dialog mutates.

**Fix:** When a picker omits the record's current value from its options, the current value must still be represented — for example as a disabled option — and "unchanged" must be the default, so no selection the user did not make can be submitted.

packages/ui/src/modules/artifacts/views/artifacts-view.tsx
```diff
@@ -71,6 +78,29 @@ export function ArtifactsView() {
...
+  const updateArtifact = useUpdateArtifact();
+  const moveArtifact = updateArtifact.mutate;
+  const dropCallbacks = useMemo<FolderDropCallbacks>(
+    () => ({
+      onStart: () => setDragInProgress(true),
+      onEnd: () => {
+        setDragInProgress(false);
+        setHotFolderId(undefined);
+      },
+      onEnter: (folderId) => setHotFolderId(folderId),
```
dam-code-guardian: 🟡 **Warning:** `ArtifactDragCallbacks.onStart` carries no artifact identity, so `onEnter` sets `hotFolderId` for any folder the pointer enters — including the folder that owns the dragged row, which the drag enters immediately because rows are nested inside their folder's drop target. The source folder then paints the `outline-primary` drop-active affordance for a drop that `onDrop` discards through its `moved.folderId === folderId` guard.

**Fix:** A drop target computes its active highlight from the same predicate that decides whether the drop is accepted — pass the dragged item's origin through the drag-start callback and suppress the highlight wherever the drop would be a no-op.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `cabc3b5f`

## PR #3370: feat(ui): move an existing artifact into a folder
**Author:** PetrBulanek | **Branch:** feat/3231-move-artifact-into-folder → main | **Changes:** +331 −10 (9 files)

### Summary
The PR adds two ways to move an existing artifact between folders: a **Move to folder…** item in the artifact row menu, and dragging a library row onto a folder group. Folder membership is now mutable from the library itself, on all three artifact surfaces.

### Findings
- 🔴 **Critical:** Move dialog detaches an artifact from its experiment folder on an untouched Move click (`packages/ui/src/modules/artifacts/components/move-artifact-dialog.tsx:42`) — reported by react-ui-engineering, confirmed by executing the selection logic.
- 🟡 **Warning:** The source folder paints the drop-active outline for a drop that is then discarded (`packages/ui/src/modules/artifacts/views/artifacts-view.tsx:90`) — reported by react-ui-engineering.
- 🟡 **Warning:** `Closes #3231` closes an issue whose multi-select story is deferred with no follow-up issue linked (`packages/ui/src/modules/artifacts/components/move-artifact-dialog.tsx:23`) — reported by issue-fit.
- 🟡 **Warning:** The artifact-library architecture page does not record the new move capability (`docs/architecture/artifact-library.md:141`) — reported by doc-drift.
  **Fix:** A diff that adds a user-facing subsystem capability updates that subsystem's `docs/architecture/` page in the same PR, with its `Last verified:` date.
- 🟢 **Suggestion:** The HTML5 drag protocol is now duplicated almost line for line between the new hook and `packages/ui/src/modules/files/hooks/use-file-row-drag.ts` — same MIME-constant, type probe, JSON shape guard, `effectAllowed`/`dropEffect` and `contains(relatedTarget)` leave guard. A payload-generic helper in `packages/ui/src/lib/` would carry one copy (`packages/ui/src/modules/artifacts/hooks/use-artifact-row-drag.ts:4`).
- ✅ **Looks good:** The drag path is safe for experiment folders. `ExperimentsSection` passes no `drop` prop, so experiment rows are not draggable and `FolderGroup` falls back to `INERT_DROP` — the defect below reaches only the menu dialog.

- 🟡 **Warning:** The artifact-library architecture page enumerates the per-artifact UI actions (preview, rename, share, retention) but not the new move-between-folders action, so it no longer states that folder membership is mutable after creation, nor that drag-and-drop onto a folder exists only on the Artifacts page while the menu dialog exists on all three surfaces (`docs/architecture/artifact-library.md:141`).
  **Fix:** When a diff adds or alters a user-facing capability of a subsystem, update that subsystem's page under `docs/architecture/` in the same PR and bump its `Last verified:` date.

### Issue Fit
Linked issue: #3231 "UI - Move an existing artifact into a folder" (closing keyword)

- 🟡 **Warning:** `Closes #3231` will close the issue while its fourth user story — moving several artifacts at once — has no implementation and no follow-up issue linked; both entry points act on a single artifact only (`packages/ui/src/modules/artifacts/components/move-artifact-dialog.tsx:23`).
  **Fix:** A closing keyword is only used when every user story of the issue is delivered; a deliberately deferred story is either kept out of the closing keyword (use a plain reference) or carries the number of the follow-up issue that owns it.

- 🟢 **Suggestion:** The HTML5 drag protocol (custom MIME constant, `has*` type probe, `read*Source` JSON parse with hand-rolled shape guard, `effectAllowed`/`dropEffect`, `contains(relatedTarget)` leave guard) is now duplicated almost line for line between the new hook and the pre-existing `packages/ui/src/modules/files/hooks/use-file-row-drag.ts`. A shared, payload-generic drag-transfer helper in `packages/ui/src/lib/` would keep the two modules independent while carrying one copy of the protocol (`packages/ui/src/modules/artifacts/hooks/use-artifact-row-drag.ts:4`).

See the inline comments on `move-artifact-dialog.tsx:42` and `artifacts-view.tsx:90` for the two findings this skill reported.

### Verdict
REQUEST_CHANGES — the move dialog can silently detach an artifact from an experiment folder on a single click, with no selection by the user.

---

PetrBulanek:
Review findings validated against the code. Dispositions below; fixes are in 8cc9d976e.

**🔴 Move dialog detaches an experiment artifact — real, fixed.**
Confirmed by reading the selection logic. The picker listed user folders only, so an artifact in an experiment folder resolved `isMoveTarget` to `false` and preselected "No folder". The dialog misreported it as ungrouped, and one untouched click on **Move** satisfied `selected !== currentFolderId` and fired `folderId: null`.

The dialog now always represents the artifact's real folder — an experiment folder joins the option list for that one artifact — and `selected` defaults to `currentFolderId`, so an untouched **Move** is a no-op in every case. The option stays enabled rather than disabled, so a user who changes their mind can revert; re-selecting it closes without a mutation. The picker is also disabled while folders load, so it can no longer show "No folder" for a filed artifact even briefly.

One correction to the finding's framing: the move itself is legitimate and must stay possible. An experiment folder is an ordinary `artifact_folders` row distinguished only by the `Experiments / ` name prefix, and `update_artifact` already lets an agent move artifacts in and out of one — the service only checks ownership. The defect was performing the move unasked, not permitting it.

Related decision, so it does not read as an oversight: the UI allows moving **out** of an experiment folder but not **in**. An experiment folder is a lineage record whose contents the Experiments section derives by matching `scriptArtifactId` / `dashboardArtifactId` against experiment records. An artifact filed into one does not become part of the experiment — it lands in the "Other files" bucket of a lineage it has nothing to do with. Out is a real need ("this dashboard is useful beyond the run"); in offers nothing but confusion. That an agent can do both through the deliberately unopinionated MCP surface is not an argument that the UI should.

**🟡 Source folder paints a discarded drop highlight — real, cosmetic, fixed.**
Correct on the mechanism: rows are nested inside their folder's drop target and are not pointer-events suppressed, so `dragenter` bubbles to the source folder and it outlines for a drop the `moved.folderId === folderId` guard then discards. No wrong mutation and no stuck state, so this was polish rather than a defect — but cheap to close. `onStart` now carries the dragged row's `folderId`, held in a ref so `dropCallbacks` does not re-memoize mid-drag, and `onEnter` suppresses the highlight when it matches. This also fixes the "Ungrouped" group lighting up when the dragged artifact is already ungrouped.

**🟡 `Closes #3231` with multi-select deferred — answered on the issue.**
Fair point, and now recorded: https://github.com/dam-agents/dam/issues/3231#issuecomment-5347487242. The issue's own open questions asked whether multi-select belonged in the first cut, and the answer is no — it is a separate interaction (selection model, bulk affordance, partial-failure reporting) for a smaller return. No follow-up issue is filed on purpose: there is no evidence yet that one-by-one moving hurts, so we file it if it does rather than parking a speculative issue. `Closes` stays — the stated goal and three of four stories are delivered.

**🟡 Architecture page omits the move capability — real, fixed.**
The page enumerated the library's capabilities and never said folder membership can change after creation. One clause added to the Artifacts bullet, `Last verified:` bumped.

Half of the ask declined: the page does not record that "drag-and-drop exists only on the Artifacts page while the menu dialog exists on all three surfaces". That is per-surface interaction detail, which `docs/guidelines/documentation-guidelines.md` classes as volatile content to keep out of architecture pages. The durable fact is that folder membership is mutable; which widget does it is not.

**🟢 Shared drag helper — declined.**
The duplication is genuine, but the two hooks diverge where it matters. `use-file-row-drag.ts` also accepts OS file drops (`"Files"` MIME, `copy` effect) and merges drag source with drop target into one hook keyed on `targetDir`; the artifacts hook splits them into `useArtifactRowDrag` and `useFolderDropTarget`. A payload-generic helper would carry roughly fifteen lines and add a cross-module indirection at exactly two call sites. Not worth the coupling yet.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `8cc9d976`

## PR #3370: feat(ui): move an existing artifact into a folder
**Author:** PetrBulanek | **Branch:** feat/3231-move-artifact-into-folder → main | **Changes:** +362 −15 (11 files)

### Summary
Commit `8cc9d976` answers the previous round: the move dialog now defaults to the artifact's real folder, and the drag highlight follows the drop it would actually perform. Everything that blocked the last review is resolved.

### Changes since last review
Previous HEAD: `cabc3b5f` (2026-08-19T13:32:21Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** the move dialog no longer detaches an artifact from an experiment folder (`move-artifact-dialog.tsx:46`) — `selected` defaults to `currentFolderId` in all nine cases, so an untouched **Move** is a no-op everywhere.
- ✅ **Fixed:** the source folder no longer paints a drop outline for a discarded drop (`artifacts-view.tsx:95`) — the `dragOriginId` ref survives a mid-drag re-memoization, which `useState` there would not.
- ✅ **Fixed:** the architecture page records that folder membership is mutable after creation (`docs/architecture/artifact-library.md:147`).
- ✅ **Fixed:** the descoping of the multi-select story is recorded on issue #3231, with the reasoning.
- ✅ **Fixed:** the shared drag-helper suggestion was declined with the divergence named; verified against both hooks — the files hook merges source and target on `targetDir` and accepts OS file drops, so the shared skeleton is a MIME guard and a JSON parse at two call sites.

### Findings
- 🟢 **Suggestion:** the new folder-membership clause reads without limit, while experiment folders are reachable only as an artifact's current location — moving out is offered, filing in is not (`docs/architecture/artifact-library.md:147`). Reported by doc-drift.
- 🟢 **Suggestion:** the PR body says the multi-select story is "left for a follow-up issue"; the issue thread says no follow-up issue is filed on purpose. One of the two records carries the decision (PR body). Reported by issue-fit.
- 🟢 **Suggestion:** while `folders` is `undefined` the picker's `value` is the real folder id but the only option is "No folder", so it renders blank instead of the folder name (`move-artifact-dialog.tsx:73`). Disabled and self-resolving, and an untouched **Move** stays a no-op. Reported by react-ui-engineering.
- 🟢 **Suggestion:** `useFolderDropTarget` accepts the full `FolderDropCallbacks` but reads only three of its five members, which is what makes `INERT_DROP` carry two stubs the hook never calls (`use-artifact-row-drag.ts:63`). Reported by typescript-engineering.
- 🟢 **Suggestion:** `ARTIFACT_MOVE_MIME`, `hasArtifactMove` and `readArtifactMoveSource` are exported with no consumer outside their file; the equivalent `files` exports have one (`use-artifact-row-drag.ts:4`). Reported by typescript-engineering.

🔁 1 finding from the previous review is resolved (see review at `cabc3b5f`).

- 🟢 **Suggestion:** The new folder-membership paragraph describes movement between folders without limit, while the code offers only user folders as destinations and keeps an experiment folder reachable solely as the artifact's current location (`docs/architecture/artifact-library.md:147`, `move-artifact-dialog.tsx:38`). The same asymmetry leaves the lineage-folder invariant on `docs/architecture/experiments.md:92` unqualified now that a row can leave an experiment folder — that page is outside this PR, so it may be worth a separate issue.

### Issue Fit
Linked issue: #3231 "UI - Move an existing artifact into a folder" (closing keyword)

🔁 1 finding from the previous review is resolved (see review at `cabc3b5f`).

- 🟢 **Suggestion:** The two records of the multi-select disposition disagree: the PR body calls it "left for a follow-up issue", the issue comment states no follow-up issue is filed on purpose (PR body; issue #3231 comment).

🔁 1 finding from the previous review is resolved (see review at `cabc3b5f`).

- 🟢 **Suggestion:** `useFolderDropTarget` takes the full `FolderDropCallbacks`, but it reads only `onEnter`, `onLeave` and `onDrop`. The two extra members are what make `folder-group.tsx` build a five-member `INERT_DROP` constant with two stubs the hook never calls (`use-artifact-row-drag.ts:63`).
- 🟢 **Suggestion:** `ARTIFACT_MOVE_MIME`, `hasArtifactMove` and `readArtifactMoveSource` are exported, and a repo-wide search finds no consumer outside this file. The equivalent `files` exports do have one — `files/components/files-panel-controller.ts` imports `MOVE_MIME` and `readMoveSource` (`use-artifact-row-drag.ts:4`, `:21`, `:25`).

🔁 2 findings from the previous review are resolved (see review at `cabc3b5f`).

Both were settled by simulation over the extracted logic rather than by reading the ternaries: `selected` equals `currentFolderId` in all nine dialog cases, and across eight drag sequences — including a mid-drag `artifacts` refetch that re-memoizes `dropCallbacks` — the outline never lands on the origin folder and every painted highlight corresponds to a drop `onDrop` would mutate.

- 🟢 **Suggestion:** while `folders` is `undefined` the option list is just "No folder" but `value` is still the artifact's real folder id, so the picker renders with `selectedIndex === -1` — visually blank rather than showing the folder name (`move-artifact-dialog.tsx:73`). It is disabled and self-resolves on load, and an untouched **Move** stays a no-op throughout, so nothing is mis-saved. Briefly visible from the chat panel and the sandbox section, which do not warm the folders query before the dialog opens.

### Verdict
APPROVE — both blocking findings are fixed at their cause, and the round introduces nothing that blocks.

---

PR #3371, opened 2026-08-19
fix(ui): make the artifact share-link button legible and its copy unmissable

### LINKED ISSUE #3064: UI - Copying an artifact share link gives almost no feedback
## Context

Each artifact row has a link button on the right. For a public artifact it copies the share link. For a private one it opens the sharing dialog instead, so the same button does two different things depending on the artifact.

On copy, the only feedback is the icon turning green for just over a second. The tooltip also changes to "Copied!", but only if you keep hovering.
 
Image

## Problem

The feedback is too quiet to register. A small icon changing colour briefly is easy to miss, especially if you moved the pointer away or were looking at the row rather than the button. So you are left unsure whether anything happened, and people click again to be sure.

Colour alone is also doing all the work, which is a poor signal for anyone who does not distinguish it easily.

It is also unclear before clicking. Because the button looks identical whether it copies a link or opens a dialog, you cannot tell which you will get, and the outcomes are quite different.

## Goal

After clicking, it should be obvious that the link was copied. It should also be clear beforehand what the button will do.

## User Stories

- As someone copying a share link, I want unmistakable confirmation so that I know it worked without clicking again.
- As someone who does not rely on colour, I want the confirmation conveyed some other way too so that I do not miss it.
- As someone looking at a private artifact, I want to know the button opens sharing settings rather than copying so that I am not surprised.

## Open Questions

- Is a toast right here, or too heavy for something this small?
- Should the two behaviours be visually distinct, or is one button correct?

PetrBulanek:
The link button on an artifact row does two different things depending on the artifact, and tells you almost nothing about either one.

For a public artifact it copies the share link, confirming only by tinting its icon green for two seconds — easy to miss if you moved the pointer away or were looking at the row rather than the button, so people click again to be sure. Colour alone carries the whole signal, which is a poor channel for anyone who does not distinguish it easily, and a copy that fails says nothing at all.

For a private artifact the same button opens the sharing dialog instead. It looks identical in both states, so you cannot tell which outcome you will get until you have already clicked.

After this change the icon differs by visibility, so the button's outcome is legible before the click; a successful copy confirms twice, with a checkmark where the pointer already is and a toast that reaches you wherever you were looking; and a failed copy says so instead of failing silently. The folder link, which gave no feedback at all, confirms the same way.

## Beyond the issue

One addition agreed during review, stated here rather than left for a reviewer to infer. Enabling sharing is the moment you want the link, and it was the one moment the sharing dialog withheld it: the dialog closed on save, leaving a `Copy link` toast action that expired after five seconds. It now stays open when a save mints the link and shows the URL in place, closing only when the artifact ends up private. Its `Save` is disabled unless the switch differs from the saved state, so the dialog can no longer show a switch that disagrees with the server.

A rejected alternative, for the record: revealing the link the instant the switch flips and reverting it if the user never saves. Abandonment cannot be detected reliably — a closed tab or a reload would leave an artifact silently public — and it would record share events that never happened.

Closes #3064

## Sub-issues

- [x] 01 — Distinct affordance and unmissable copy feedback on the share-link button


packages/ui/src/modules/artifacts/components/share-dialog.tsx
```diff
@@ -15,55 +15,38 @@ import { useCopy } from "@/hooks/use-copy";
...
 
 interface Props {
   artifact: LibraryArtifact;
   onClose: () => void;
 }
 
 export function ShareDialog({ artifact, onClose }: Props) {
-  const [isPublic, setIsPublic] = useState(artifact.visibility === "public");
-  const shareUrl = artifact.shareUrl;
+  const [committed, setCommitted] = useState({
```
dam-code-guardian, line 26: 🟢 **Suggestion:** `committed` snapshots the server-owned `visibility` and `shareUrl` at mount and never re-syncs from the `artifact` prop, while the header renders `artifact.title` live. Verified non-blocking: `onSuccess` writes the server response back into `committed`, and neither mount site can change the artifact identity behind the open modal. The residual case is an external change to the same artifact during the window this PR deliberately extends.

Keeping server-owned fields in props and holding only the unsaved intent in state would remove the class.

packages/ui/src/modules/artifacts/components/artifact-row.tsx
```diff
@@ -164,13 +167,27 @@ function ShareLinkButton({
     <Button
       variant="ghost"
       size="icon-sm"
-      aria-label={url ? "Copy share link" : "Sharing settings"}
+      aria-label={
```
dam-code-guardian, line 170: 🟢 **Suggestion:** the `aria-label` states a past event ("Share link copied") for two seconds, where an accessible name normally states the action. The toast `toastCopyOutcome` emits already announces the outcome, so the label could stay constant.

Not a regression, and it matches the repo's existing pattern at `create-api-key-dialog/reveal-token.tsx:35`.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `227cdffa`

## PR #3371: fix(ui): make the artifact share-link button legible and its copy unmissable
**Author:** PetrBulanek | **Branch:** fix/3064-artifact-share-link-feedback → main | **Changes:** +82 −62 (5 files)

### Summary
The share-link button now shows a distinct icon per visibility, and every copy path reports its outcome through one shared toast helper. The sharing dialog stays open when a save mints the link, so the user can copy it in place.

### Findings
- 🟢 **Suggestion:** `ShareDialog` snapshots `visibility` and `shareUrl` into `committed` at mount and never re-syncs from the `artifact` prop (`packages/ui/src/modules/artifacts/components/share-dialog.tsx:26`)
- 🟢 **Suggestion:** the button's `aria-label` carries a transient status ("Share link copied") for two seconds (`packages/ui/src/modules/artifacts/components/artifact-row.tsx:170`)
- ✅ **Looks good:** the `disabled={sharedCount === 0 || !onCopyFolderLink}` guard closes a real gap. `FolderGroup` takes `onCopyFolderLink` as optional, and the ungrouped group is rendered without it, so the menu item was previously enabled and copied nothing.
- ✅ **Looks good:** the `writeToClipboard` extraction keeps the original behaviour exactly. The `clearTimeout`, the `setState`, and the 2000 ms reset stay in `copy`, so the added return value is the only change to the hook's contract, and the five existing `useCopy` call sites are unaffected.

✅ No findings.

### Issue Fit
✅ No findings.

✅ No findings.

- 🟢 **Suggestion:** `committed` holds server-owned fields that the `artifact` prop also carries, and the two can disagree while the dialog is open. Verified non-blocking: `onSuccess` writes the server response into `committed`, and at both mount sites the artifact identity cannot change behind the open modal (`docked-artifact-panel.tsx:194` keeps its only `setOpenArtifactId` control behind the overlay; `artifacts-view.tsx:216` passes a fixed object). The residual case is an external change to the same artifact during the newly extended open window. Reading server-owned fields from props and keeping only the unsaved intent in state would remove the class.
- 🟢 **Suggestion:** the accessible name of a control states its action, and this one states a past event for two seconds after a copy. The toast that `toastCopyOutcome` emits already announces the outcome. Verified as the repo's established pattern, not a regression: `create-api-key-dialog/reveal-token.tsx:35` swaps the same label the same way, and the pre-change tooltip already flipped to "Copied!". A constant label plus the toast would be the stricter form.

### Verdict
APPROVE — the change is correct, the copy paths all report their outcome through one helper, and no finding blocks.

---

kapetr (approved):

PR #3339, opened 2026-08-17
refactor(agent-runtime): extract prompt scheduler from acp-runtime

tomkis:
Stacked on #3331, continuing the acp-runtime decomposition.

## What

Pulls the prompt turn state out of the acp-runtime factory into a sibling module, `prompt-scheduler.ts`. The scheduler owns the per-session active turn, the 32-prompt queue, promotion on turn end, the PROMPT_QUEUE_FULL refusal, and the promptAccepted/promptStarted notifications. The runtime submits a prompt and gets back its fate (started, queued, refused); on a prompt response it asks the scheduler whether the turn ended.

acp-runtime.ts shrinks by ~160 lines and no longer touches turn state directly.

## Behavior notes

The extraction preserves base behavior, with one dropped record and two review-driven fixes:

- The active turn is now a sid to outboundId map. The old ActivePrompt record carried two more fields, and here is what happened to each. `originalId` was never read, so it is retired with nothing to replace it. `channel` was only written (nulled in detach), so that nulling is retired too: an in-flight turn no longer carries a channel reference at all. Queued prompts still carry theirs, and `dropChannel` still discards a leaver's queued entries. A leaver's in-flight turn behaves as before: the response routes through outboundIdToClient.
- Review fix (the critical finding): the first version moved the dead-agent guard into the injected `sendToAgent`, which reversed commit and delivery. Promoting onto a dead harness recorded a phantom turn and sent a false promptStarted. Now `sendToAgent` reports whether the frame was delivered, and the scheduler commits the active turn and emits promptStarted only on delivery. An undeliverable promotion stays queued, same as base.
- Review fix: the `a.exited` handler now calls `promptScheduler.clear()` beside the other map clears. This fixes the "dead harness reports busy forever" bug in this PR instead of deferring it, and it makes the glossary sentence about queued prompts ("none survive the harness process") actually true.

## Docs

- Prompt Scheduler enters the ubiquitous language.
- New page `docs/guidelines/comment-guidelines.md`. The typed-comment registry and tool-directive rules are moved from CLAUDE.md, which now just points there. The "How to write one" section is new normative policy, not moved text: it adds three authoring rules (plain English for ESL readers, grounded in mechanics, domain vocabulary from the ubiquitous language).

## Tests

[...]

PR #3331, opened 2026-08-14
refactor(agent-runtime): extract session transcript from acp-runtime

tomkis:
## What

Extracts the session log machinery out of `acp-runtime.ts` (1,237 → 1,115 lines) into a new `SessionTranscript` module with a small interface. The runtime no longer touches logs or cursors directly.

- **`session-transcript.ts`** owns the per-session message log, per-channel read cursors, byte-cap eviction, and the truncation sentinel. The three fan-out cases are named methods (`append`, `appendEcho`, `appendReplay`), replacing the `skipChannel`/`onlyChannel` options bag and its presence-check trick.
- **`mappers.ts` moved** from `infrastructure/` to `domain/` — both functions are pure frame logic, and the move fixes the services→infrastructure import.
- **New comment type `UNIT_BOUNDARY_DESCRIPTION`** registered and used once, above `createSessionTranscript`.
- **Ubiquitous language**: new *Session Transcript* row.

## Why

`acp-runtime.ts` is a hot spot (30 commits) and the log+cursor invariant was spread over nine closures — #3221 and #703 were both bugs in that spread. This concentrates it behind one seam.

## Notes

- Behavior-preserving: no wire changes, all 185 agent-runtime tests pass unchanged.
- The teardown drift found during review (stale busy state after agent exit) is deliberately not fixed here — filed as bd `dam-2jo`.

https://claude.ai/code/session_01AfM5b9E7KbtPuNiUSXKB63

PR #3373, opened 2026-08-19
refactor(agent-runtime): extract harness lease from acp-runtime

tomkis:
Extracts the harness process lifecycle out of acp-runtime into a Harness Lease module, following the session-transcript (#3331) and prompt-scheduler (#3339) extractions.

## What moves

One module now owns the harness child process: spawn on demand, the warm-start gate (hold early clients until env is ready, bounded by a timeout), the env recycle policy (immediate when idle, deferred until work drains, forced after a grace period), and the teardown sequence. Every path down reports one reason: agent-exited, env-recycle, or shutdown. All frames to the process go through lease.send; nobody else holds the handle.

## Behavior fixes from unifying teardown

The three teardown copies had drifted apart. Unified, every path now runs the full sequence:

- shutdown clears pendingFromAgent and the prompt scheduler directly (was: only transitively via the exit handler)
- agent exit clears the env force timer and the pending refresh flag (was: a force timer could fire against a fresh spawn)
- frames are no longer sent to a process already being recycled

## Tests

Three new specs in acp-runtime-env-changes.test.ts, written through the public interface like the rest of the suite: recycle right away when idle, let the running turn finish and deliver its answer first, force the recycle after the grace period when work never drains. The forced-recycle path had zero coverage before. All 8 existing spec files pass unchanged.

Also adds the Harness Lease term to docs/ubiquitous-language.md.


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `6873fe29`

## PR #3373: refactor(agent-runtime): extract harness lease from acp-runtime
**Author:** tomkis | **Branch:** refactor/harness-lease → main | **Changes:** +343 −158 (4 files)

### Summary
The harness child-process lifecycle moves out of `acp-runtime.ts` into a new `harness-lease.ts` module, and the three drifted teardown copies become one sequence that reports its reason. Three specs cover the env-recycle paths, including the forced one that had no coverage before.

### Findings
- 🟢 **Suggestion:** The body lists three behavior fixes; two more ride along. `shutdown()` now sets `terminal`, so a later `attach` cannot revive the process, and the post-exit early returns in `updateOrphanTimerForSession`, `expireSession` and `reapIdleSessionNow` are gone (`packages/agent-runtime/src/modules/acp/services/acp-runtime/harness-lease.ts:145`, `.../acp-runtime.ts:271`). Both verified safe at HEAD — the only `shutdown()` caller is the SIGTERM path in `server.ts:594`, which exits the process, and the dropped guards are covered by `lease.send()` returning `false` on no process. Listing them under "Behavior fixes from unifying teardown" would save the next reader that verification. Reported by issue-fit.
- ✅ **Looks good:** The unified teardown is equivalent to the three copies it replaces, and the recycle-versus-crash distinction holds: `recycle()` nulls the handle without setting `terminal`, and the identity guard `if (agent !== a) return` in the `exited` handler stops a killed process from marking a fresh lease dead.

✅ No findings.

### Issue Fit
_No linked issue — no closing keyword and no issue reference in the body or branch (`#3331`/`#3339` are PRs); scope checked against the PR description only._

- 🟢 **Suggestion:** The body enumerates exactly three behavior fixes, but the diff carries more: `shutdown()` now sets `terminal = true`, making shutdown irreversible, where the old code left the handle in place so a later `attach` reused the dead process (`packages/agent-runtime/src/modules/acp/services/acp-runtime/harness-lease.ts:145`). Related undeclared change: the post-exit guards in `updateOrphanTimerForSession`, `expireSession`, and `reapIdleSessionNow` were dropped, so those paths now run after the harness is gone instead of returning early (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:271`). Adding both to the "Behavior fixes from unifying teardown" section keeps the list complete.

_The skill reported this as 🟡. Reformatted to 🟢 and merged into `### Findings`: both changes were verified reachable and safe at HEAD, so what is left is one description gap on an otherwise-correct diff._

✅ No findings.

### Verdict
APPROVE — the extraction holds, the unified teardown fixes real drift, and the one open item is a description gap, not a code defect.

---

Issue #3385, opened 2026-08-20
Opening a long conversation stays slow even when everything is warm


jezekra1:
## Context

Epic #3242. The #3247 benchmark broke session-switch latency into its parts. After the harness adapter fix (#3384), the remaining cost is transmission: opening a conversation replays **every message** of it to the browser, one WebSocket frame per event. A long conversation (~1,800 events) spends 1.7–4.6 s just streaming, and this cost is identical whether the agent's replay cache is warm or cold — reading the transcript from disk takes ~30 ms; sending it takes seconds.

## Problem

The wait to open a conversation grows linearly with its length, forever. The conversations users return to most — long-running ones — are exactly the ones that punish them the hardest, with a multi-second blank spinner before anything is readable. Yet what a returning user almost always wants first is the *end* of the conversation, which is a tiny fraction of what gets streamed.

## Goal

Opening a conversation feels equally fast regardless of its length. Recent messages appear quickly; older history is still reachable when the user actually wants it.

## User Stories

- As a user, I want a long conversation to open as fast as a short one so that returning to my main working session doesn't interrupt my flow.
- As a user, I want to scroll back into older messages on demand so that trimming the initial load never costs me access to history.

## Scope

- In: how much history is replayed on open, and how the rest is fetched on demand.
- Out: agent wake time (#3246), harness process startup (#3384), duplicating transcripts into any second store.

## Proposed solution

The runtime already keeps a bounded per-session replay log in memory and already marks replays it has clipped. Serve only the most recent slice of that log when a conversation opens, and let the client request older ranges from the same log on demand. Shape the "give me older messages" request as platform metadata compatible with the direction the agent protocol is already taking (its v2 draft unifies loading around a "replay from" cursor), so we align rather than diverge when v2 lands. No transcript is stored twice; the change is only in how much of the existing log is forwarded.

## Open Questions

- Tail size: fixed event count, byte budget, or viewport-driven?
- Should the clipped-replay marker double as the "load older" affordance in the chat view?
- Cold opens rebuild the log from the harness in-pod first — confirm the tail cap applies to what is forwarded, not to what is cached.

## Additional resources

- Measurements and methodology: https://github.com/dam-agents/dam/issues/3247#issuecomment-5294917381
- Benchmark harness: #3330

Issue #3247, opened 2026-08-10
Find out why switching sessions is slow


jezekra1:
## Context

Epic #3242. Clicking into an older conversation makes you wait before you can
read anything. We don't know how much of that is the server, the agent's disk,
or the browser.

## Research Goal

Know how long it takes to open a conversation and where that time goes.

## Research Questions

- How long from clicking a session to seeing the conversation on screen?
- How much slower is it when the agent was asleep and the history has to be read
  back from its disk, versus when it's already running?
- Does it get worse the longer the conversation is?
- How much is the server fetching the history versus the browser drawing it?
- Is dev slower than local, and by how much?

## Users / Audience

Anyone switching between an agent's sessions.

## Sources of Insight

Timed session switches in dev and local, browser timings, server and runtime logs.

Run every measurement on a well-resourced agent — at least 4 CPU and 4Gi of
memory. We want to know how fast the best case is, not how slow a starved agent
is.

## Decision it informs

Whether we fix the fetching, the storage, or the rendering — and whether we
should show something on screen sooner instead.

## Done when

- Numbers for the warm and cold cases, and for short and long conversations
- A breakdown of where the time goes
- A proposed fix, as its own issue if it's a big one

jezekra1:
## Session-switch latency: dev measurements

Built a repeatable benchmark for the session-open path and ran it on dev. The harness will land as `packages/bench` (PR follows); raw samples are JSONL so future fixes can be verified against these numbers.

### What was measured

The stopwatch brackets exactly what the UI's history fetch does on a session click: open a WS to `wss://<host>/api/agents/<id>/acp`, `initialize`, send `session/load`, consume replayed `session/update`s until the response lands. **Wake is excluded** (pod `Ready` before every sample — agent start is #3246), and **browser render is excluded** (see follow-ups).

- Sessions are synthesized from a committed real 5-turn claude-code conversation (tool calls with long outputs included) and written onto the agent's disk store; a session id that has never been loaded is **cold** by definition (runtime forwards to the harness adapter), its second load is **warm** (served from the runtime's in-memory log).
- **short** = 5 turns → 22 replay events / ~23 KB. **long** = 80 repetitions → 400 turns → 1,760 events / ~1.85 MB (just under the runtime's 2 MB replay-log cap — no clipping was observed).
[...]

Issue #3242, opened 2026-08-10
Epic - Measure and cut platform latency


tomkis:
## Context

Users feel the platform is slow. We suspect the time goes partly into platform primitives — pods, volumes, provisioning, pod startup, code in the server — and partly into inference speed through LiteLLM. But this is suspicion, not knowledge: nobody can say today where the time actually goes, so every speedup discussion stalls at opinions.

## Problem

Waiting is the most visible quality of the product, and right now we don't even have a shared picture of what "slow" means. Before we can fix anything, we need to experience the platform the way users do, capture which moments feel slow, and document them properly — with numbers, in both the dev and local environments. Only then can we fix the slowness with confidence instead of guessing.

## Goal

The platform feels fast, and we can prove changes with numbers. In order:

1. **Understand and document what's slow** — walk through the platform as users do, capture the slow moments, and document them with measurements in both dev and local environments. This document is the foundation for everything else.
2. A **prioritized plan** for the platform-owned slowness, driven by that data.
3. A **latency report for LiteLLM** handed to the people who own the model hosting, so they can push for improvement on their side.
4. **Fix the platform-owned slowness** per the plan.

## Scope

- **In:** documenting user-felt slowness in dev and local environments; measuring the agent start/wake path and LiteLLM latency as seen from agents; the speedup plan and the LiteLLM report; implementing the platform-owned fixes.
- **Out:** fixing LiteLLM itself — that's externally owned; our part ends at measurement and a credible report.

PR #3387, opened 2026-08-20
feat(runtime): tail-first session replay with load-older paging

jezekra1:
## Summary

Implements #3385 (epic #3242): opening a conversation now has a bounded wire cost regardless of its length.

**agent-runtime**
- `session/load` replays only the newest tail of the in-memory session log (default 200 events, `replayTailEvents` dep). The clipped-replay sentinel gains a `_meta.platform.olderBefore` cursor when the cut is the cap's doing; a sentinel without a cursor still means history is gone (log eviction).
- A load carrying `_meta.platform.replayBefore` is served the older range from the log, page by page, down to the eviction floor. Pages move no cursors, so live fan-out is untouched.
- Cold client loads are unified with cold resumes: they park as waiters, the runtime rehydrates the harness with its own silent `session/load`, and every waiter is then served the capped tail from the log. The initiator-streaming path is gone — a 1,800-event replay costs the in-pod leg (~40 ms measured), not the wire.
- Stale cursors (minted before a pod restart) are refused with an invalid-params error instead of being guessed at under a new log's numbering.

**UI**
- The cursor-carrying marker renders as a **Load earlier messages** button; the fetched page replaces the marker in place (its own marker continues the chain). On a refused stale cursor the UI falls back to a full reload.

**Docs**: agent-lifecycle page updated (session-inside-the-pod section).

Everything is `_meta`-based and shaped to align with ACP v2's `replayFrom` direction — no protocol divergence, conformant peers ignore it.

## Why

The #3247 benchmark showed transcript transmission dominates session-open latency: 1,760 events ≈ 1.7–4.6 s through the relay, hitting warm loads identically (streaming to the browser is ~1 ms/event; reading the transcript from disk is ~30 ms). Numbers: https://github.com/dam-agents/dam/issues/3247#issuecomment-5294917381

## Test plan

- [x] 8 new runtime scenarios (`acp-runtime-history-replay.test.ts`): tail cap + cursor, under-cap invisibility, paging to the start, live fan-out isolation, eviction floor, stale-cursor refusal, silent cold bootstrap serving multiple waiters
- [x] full acp-runtime suite (50/50), UI suite (302/302), tsc/eslint/prettier on both packages, comment-types, doc-size
- [ ] CI
- [ ] after dev deploy: re-run the #3330 session-load benchmark — long-conversation cold/warm totals should drop from 4.2 s/1.8 s p50 to roughly the short-conversation numbers

### DIFF
--- a/docs/architecture/agent-lifecycle.md
+++ b/docs/architecture/agent-lifecycle.md
@@ -1,6 +1,6 @@
 # Agent lifecycle
 
-Last verified: 2026-08-18
+Last verified: 2026-08-20
 
 ## Overview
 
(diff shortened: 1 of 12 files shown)


jezekra1:

Issue #3386, opened 2026-08-20
UI - Open a conversation over one connection instead of two


jezekra1:
## Context

Epic #3242. The #3247 benchmark measured connection setup (WebSocket open, auth, relay attach, protocol handshake) at ~0.5 s per connection on dev — and clicking into a conversation currently pays it twice: the UI opens one connection to fetch history, throws it away, then opens a second one to attach live, plus a listing call alongside.

## Problem

Every session switch spends ~1 s on connection ceremony before any conversation data moves. Half of that is a connection the UI discards immediately after the history arrives, even though that connection is already attached and usable for the live session.

## Goal

Opening a conversation performs its history fetch and live attach over a single connection, cutting the fixed per-click overhead roughly in half.

## Done when

- Clicking into a session opens at most one new WebSocket connection to the agent.
- Live updates after the history load arrive without a reconnect.
- The #3247 benchmark's connect phase is paid once per session open, verified on dev.

PR #3390, opened 2026-08-20
feat(ui): one connection per session open

jezekra1:
## Summary

Implements #3386 (epic #3242), stacked on #3387 because both rework the same session hooks.

Opening a conversation previously cost three WebSocket connections: a throwaway one for `session/load`, a second for the live `session/resume` attach, and a passive one for `session/list` — at ~0.5 s of connect+auth+relay-attach each on dev. Now one connection does everything:

- the live connection performs the history load itself; a per-session collector captures replayed updates until the response, then the same socket stays live (the runtime already engaged it on load, so the client just binds — no resume round-trip either)
- the session-mode reconcile (`session/list`) rides the same connection
- infinite-scroll pages (`_meta.platform.replayBefore`) reuse the live connection instead of a throwaway socket per page — without this, each scroll page would have paid a fresh ~0.5 s connect on dev
- reconnect replays ride the reconnected socket
- `use-acp-history` (the throwaway-connection hook) is deleted

## Test plan

- [x] `ui:check` (tsc/eslint/prettier), 302/302 UI tests
- [ ] manual on the local cluster: session click opens one WS; paging emits `session/load` frames with `replayBefore` on the same socket
- [ ] CI
- [ ] dev: #3330 benchmark connect phase paid once per click


jezekra1:

PR #3393, opened 2026-08-20
feat(runtime): session-history provider fills cold loads without the harness

jezekra1:
## Summary

Third layer of the session-open latency work (epic #3242), stacked on #3390/#3387. After the tail cap fixed transmission and the pin bump shrank the spawn, the last cost on a cold open was waiting for the claude subprocess at all. This makes cold ≈ warm, generically:

- **Runtime capability (harness-agnostic):** `runtime-manifest.yaml` may declare `sessionHistory.command`. On a cold `session/load` the runtime execs it, validates its ND-JSON `session/update` output, fills the session transcript, and serves every waiter the capped tail — **no harness process involved**. Any deviation (non-zero exit, timeout, malformed frames, undeclared) falls back to the harness load: the provider accelerates, it never decides.
- **Lazy rehydrate:** the harness never loaded a provider-served session, so the first `session/prompt` triggers one silent harness `session/load` (its replay is dropped — the transcript already holds the history), holds prompts until it answers, then forwards them in order. The harness's answer also upgrades the placeholder response metadata (`modes: null`) to the real one. The spawn thus moves from "open a conversation" to "send the first message", where a running claude is genuinely required.
- **Claude implementation:** `harness-history.mjs` in the claude-code image reuses the adapter's own exported replay pipeline (`getSessionMessages` + `toAcpNotifications` + the same skip/strip rules), so provider history is byte-identical to a harness replay. A build-time symlink pins it to the mise-installed adapter version.
- Idle-reaping a provider-served session sends no `session/close` (nothing to close).

Other harnesses can implement the same contract — notably pi-agent, whose sessions today cannot replay at all after a pod restart.

## Measured

- Provider script: 1,760 updates (x80 fixture, 6.8 MB store) in **~250 ms** end to end, kinds identical to adapter replay
- Replaces the ~0.7–1.0 s (local) / 2.1–2.3 s (dev, pre-pin-bump) spawn-blocked cold path measured in #3247

## Test plan

- [x] 5 new runtime scenarios (provider serve, fallback, lazy rehydrate without duplication + metadata upgrade, prompt hold/order, no session/close on reap); full agent-runtime suite 210/210
- [x] provider script validated standalone against the benchmark fixture (counts + kinds match adapter replay exactly)
- [x] docs: agent-lifecycle page updated within its size cap
- [ ] CI image build (symlink + manifest)
- [ ] local cluster: cold open of a seeded long session ≈ warm

## Update: final contract (after review discussion)

- **Opt-in tail**: the cap applies only when the load request carries `_meta.platform.tail` — clients that don't opt in get the full replay mandated by ACP v1, so unknown ACP clients stay conformant. Our UI always opts in.
- **Clip cursor moved to the response**: no custom notification method — load and page responses carry `_meta.platform.clipped` (`{olderBefore}` while pageable, `{}` at the eviction floor), and the UI synthesizes its scroll marker from the response. This replaced an in-stream sentinel that the ACP SDK's update-kind validation silently dropped (a latent bug that had also killed the pre-existing 2MB-truncation notice).
- **Worker-hosted provider**: `sessionHistory` supports a `module` variant hosted in one persistent worker thread — module graph compiled once, cold provider cost drops from ~440ms (exec-per-load) to the ~80ms of actual read+convert, with crash/timeout isolation and respawn. The `command` exec variant remains for images preferring process isolation. claude-code ships the provider as a library with the executable kept as a thin wrapper.
- Infinite scroll: prepend compensation runs in a layout effect post-commit with native scroll anchoring disabled.

Measured on a local cluster (x80 fixture, 1,760 events): cold open ~490ms with exec provider, warm ~45ms, pages 6-14ms; worker variant cuts the cold provider term to ~80-100ms.


jezekra1:

PR #3394, opened 2026-08-20
refactor(agent-runtime): extract pending agent requests from acp-runtime

tomkis:
Fourth extraction from the acp-runtime orchestrator, same shape as session transcript, harness lease, and session bootstrap.

The new pending-agent-requests module holds each agent-initiated request (typically a permission prompt) open until a client answers. It owns the fan-out to engaged channels, the replay when a channel engages later, first answer wins, and the orphan TTL expiry that answers the agent with a structured error. Expiry nudges the harness lease through an onExpired callback, so a deferred env recycle goes through once the last open question dies.

acp-runtime.ts shrinks by 51 net lines; eight touchpoints became one-line calls. Behaviour is unchanged, with one deliberate fix found in review: an empty-string session id now returns null from extractParamsSessionId, so it stays on the session-less broadcast path everywhere, same as before the extraction.

Tests: new world-based spec file, now 6 scenarios. The expiry path had zero coverage before. One spec pins that a request with no session never expires, another that an empty-string session id behaves the same way.

Docs: Pending Agent Requests added to the ubiquitous language.


packages/agent-runtime/src/modules/acp/services/acp-runtime/pending-agent-requests.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  return {
+    onAgentRequest(id, sessionId, line) {
+      pending.set(id, { sessionId, line });
+      const out = rewriteAuthError(line);
+      for (const channel of deps.channelsFor(sessionId)) {
+        if (channel.isOpen()) channel.send(out);
+      }
+      if (sessionId !== null) updateExpiryTimer(sessionId);
```
dam-code-guardian, line 99: 🟡 **Warning:** The session guard changed from truthiness to `sessionId !== null`, so a request whose `params.sessionId` is `""` now fans out to engaged channels and starts a TTL expiry timer. Before the extraction `if (sessionId)` sent it down the broadcast path that never expires. Reported by my own differential run, `issue-fit`, and `typescript-engineering`.

**Fix:** Make the empty string session-less on every path — return `null` from `extractParamsSessionId` for `""`, which also keeps `acp-runtime.ts:452` on the same rule.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `a8f2265`

## PR #3394: refactor(agent-runtime): extract pending agent requests from acp-runtime
**Author:** tomkis | **Branch:** refactor/pending-agent-requests → main | **Changes:** +335 −87 (4 files)

### Summary
The PR extracts the pending-agent-requests module out of the acp-runtime orchestrator, the fourth extraction in this series. It adds a world-based spec file with 5 scenarios and one ubiquitous-language row.

### Findings
- 🟡 **Warning:** Empty-string session id no longer takes the broadcast path (`packages/agent-runtime/src/modules/acp/services/acp-runtime/pending-agent-requests.ts:99`)
- 🟢 **Suggestion:** `tearDownSession` calls `forget(sessionId)` on the transcript, the scheduler and background work, but the new module exposes no equivalent, so a `resetSession` leaves that session's pending requests open. This matches `main`, so it is pre-existing — worth a separate issue (`acp-runtime.ts:309`).
- 🟢 **Suggestion:** Two body numbers differ from the diff: `acp-runtime.ts` shrinks 817 → 766, so 51 net lines rather than 86, and the module has 8 call sites rather than ten.

✅ No findings.

### Issue Fit

Both findings merged into `### Findings` above.

Finding merged into `### Findings` above. Equivalence was settled by execution rather than by reading for symmetry: both implementations were ported to runnable JS with injected fake timers and compared across 10 scenarios plus a 3000-iteration differential fuzz. Normal session ids gave 0 divergences; adding `""` to the pool diverged in 495/3000 runs, isolating exactly the guard swap above. Teardown, double-answer, detach-reassess and the null-session broadcast path matched exactly.

### Verdict
COMMENT — the extraction is faithful on every path except the empty-string session id, which changes an unbounded request into one that expires.

---

tomkis:
@dam-code-guardian — fixed in c57ebf9:

- extractParamsSessionId now returns null for "", so an empty-string session id stays on the session-less broadcast path everywhere, including the pending-request fan-out and expiry guard. New spec pins it.
- PR body numbers corrected: 51 net lines, eight call sites.
- The resetSession gap (pending requests not forgotten on teardown) is pre-existing and matches main; I left it for a separate issue as suggested.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `c57ebf9`

## PR #3394: refactor(agent-runtime): extract pending agent requests from acp-runtime
**Author:** tomkis | **Branch:** refactor/pending-agent-requests → main | **Changes:** +358 −88 (4 files)

### Summary
The fix commit makes an empty-string session id session-less in the shared frame extractor. That resolves the previous warning on the pending-request path, but it also changes the prompt path, which was not part of the reported defect.

### Changes since last review
Previous HEAD: `a8f2265` (2026-08-20T12:54:40Z) — verdict COMMENT

- ✅ **Fixed:** Empty-string session id takes the broadcast path again and never expires (`pending-agent-requests.ts:99`)
- ✅ **Fixed:** PR body numbers now match the diff — 51 net lines, eight call sites
- 🔁 **Still present:** No `forget()` equivalent on the new module in `tearDownSession`; pre-existing, matches `main` (`acp-runtime.ts:309`)
- 🆕 **New:** The extractor change also removes `session/prompt` frames with an empty session id from the prompt scheduler (`acp-runtime.ts:513`)

### Findings
- 🟡 **Warning:** `extractParamsSessionId` is shared by four call sites, so mapping `""` to `null` also changes the prompt path. On `main` the extractor returned `""`, and the `promptSessionId !== null` guard at `acp-runtime.ts:549` admitted it, so a `session/prompt` with `sessionId: ""` went through `promptScheduler.submit`. Now `promptSessionId` is `null`, the frame is sent raw with `lease.send`, no turn is tracked, and a harness recycle can land mid-turn with no reply to the client (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:513`). Reported by `typescript-engineering`, confirmed against `main`.
  **Fix:** Do not change a shared extractor to correct one caller. Keep `extractParamsSessionId` returning the raw string, and apply the session-less rule at the agent-request call site only (`acp-runtime.ts:451`), so the prompt, resume, load, and notification sites keep the guard semantics they were written against.

✅ No findings.

### Issue Fit

✅ No findings. Both prior findings are resolved; the body claims now match the diff.

🔁 The prior review's finding is fixed. The new finding is merged into `### Findings` above.

### Verdict
COMMENT — the reported defect is fixed, but the fix landed one level too deep and moves an empty-session prompt off the scheduler.

---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `616838a`

## PR #3394: refactor(agent-runtime): extract pending agent requests from acp-runtime
**Author:** tomkis | **Branch:** refactor/pending-agent-requests → main | **Changes:** +379 −87 (5 files)

### Summary
The fix commit narrows the empty-session rule to the agent-request site, so `session/prompt` keeps the Prompt Scheduler. That resolves the previous warning at its cause, and a new spec pins each side of the rule.

### Changes since last review
Previous HEAD: `c57ebf9` (2026-08-20T14:23:01Z) — verdict COMMENT

- ✅ **Fixed:** `extractParamsSessionId` returns the raw string again, byte-identical to `main`, so an empty-session `session/prompt` goes through `promptScheduler.submit` as before (`acp-runtime.ts:752`)
- 🔁 **Still present:** No `forget()` equivalent on the new module in `tearDownSession`; pre-existing, matches `main` (`acp-runtime.ts:309`)

### Findings
- 🟢 **Suggestion:** The PR body is stale against this HEAD in two places. It still says the empty-string rule applies "everywhere", which `616838a` deliberately narrowed to Pending Agent Requests, and it states 51 net lines removed from `acp-runtime.ts` where the diff gives 46 (817 → 771). Reported by `issue-fit`.

✅ No findings.

### Issue Fit

Both findings merged into `### Findings` above.

✅ No findings. Equivalence was settled by execution: main's runtime was driven beside HEAD's through 30000 randomized frame sequences with fake timers, degenerate session ids (`""`, `" "`, `"0"`) and malformed params in the pool, comparing the harness method stream, per-id answers, busy state and channel state — 0 divergences. Positive controls confirm the harness has power: restoring the round-1 guard reproduces its defect on `connect` + agent request with `sid:""`, and restoring the round-2 shared-extractor change reproduces its defect on `connect` + prompt with `sid:""`. Both are clean at this HEAD, and each new spec fails under exactly one of the two defects.

### Verdict
APPROVE — the narrowed fix restores the prompt path to its `main` semantics while keeping the agent-request rule, and the only open item is a 🟢 body correction.


---

PR #3398, opened 2026-08-20
fix(agent-runtime): session reset drops pending agent requests

### LINKED ISSUE #3396: acp-runtime: resetSession leaves pending agent requests open
Found during review of #3394, pre-existing on main.

tearDownSession calls forget(sessionId) on the session transcript, the prompt scheduler and background work, but pending agent requests have no equivalent. After a resetSession, that session's open permission prompts stay in the pending map: they still count toward status().idle, still replay to a channel that engages the recycled session id, and their expiry timer can still fire and answer a harness that no longer knows the request.

Fix idea: add forget(sessionId) to the PendingAgentRequests interface (drop the session's entries, clear its expiry timer, optionally answer the agent with the structured expiry error) and call it from tearDownSession.

tomkis:
Closes #3396.

## The bug

`tearDownSession` forgets the Session Transcript, the Prompt Scheduler and background work, but never touched Pending Agent Requests. So a permission prompt that nobody answered outlived the reset that was supposed to clear the session. Three things followed from that:

1. **The pod stopped hibernating.** An open request counts as work in `runtimeBusy()`, so `/api/status` kept reporting `idle: false` and the controller's idle checker skipped the Agent for good. Same signal blocks a harness recycle, so a pending env change never landed either.
2. **A ghost prompt in the next conversation.** The session id is reused after a reset. The next client to engage it got the old prompt replayed by `onEngaged`, about a tool call from before the reset. Answering it sent a response for a request id the harness no longer knew, or worse, one it had reassigned.
3. **The TTL timer fired into nothing**, answering a torn-down session's request long after the fact.

## The fix

`PendingAgentRequests` gains `forget(sessionId)`, and `tearDownSession` calls it next to the other per-session forgets. It drops the session's entries, clears the expiry timer, and aborts each request with a structured error so the harness's tool call ends cleanly instead of hanging. `expire()` and `forget()` now share one abort path, so the two cannot drift apart. Only the message differs.

The idle-reap path can't reach a session with open requests (`reapIdleSessionNow` bails on `pendingRequests.hasFor`), so reset is the only caller that has any work to do here.

## Tests

Three scenarios in `acp-runtime-unanswered-requests.test.ts`, one per symptom above, all written against the runtime's public surface. Verified they fail on the parent branch and pass with the fix. Full suite: 207 passing.

## Base branch

Stacked on `refactor/pending-agent-requests` (#3394), which is where the extracted unit lives. Retarget to `main` after that merges.


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `8286b65`

## PR #3398: fix(agent-runtime): session reset drops pending agent requests
**Author:** tomkis | **Branch:** fix/pending-requests-session-reset → refactor/pending-agent-requests | **Changes:** +118 −21 (5 files)

### Summary
`tearDownSession` now calls a new `PendingAgentRequests.forget(sessionId)`, so a session reset drops the session's open agent requests, clears the TTL timer, and aborts each request with a structured error. All three symptoms in the body verify at their mechanism: the open request no longer holds `runtimeBusy()`, `onEngaged` no longer replays a pre-reset prompt, and the expiry timer is cleared before the abort.

### Findings
- 🟢 **Suggestion:** `forget` does not call `deps.onExpired()`, while the sibling `expire` path does. No defect follows — `tearDownSession` calls `lease.maybeRecycle()` immediately after `pendingRequests.forget(sessionId)`, so the recycle still happens; the asymmetry is internal to the module (`packages/agent-runtime/src/modules/acp/services/acp-runtime/pending-agent-requests.ts:147`). Reported by typescript-engineering as a warning; demoted after checking the caller.
- 🟢 **Suggestion:** Session Bootstrap has no `forget`, so a `session/load` in flight at reset time leaves its `bootstrapBySession` entry behind. Because the session id is reused — this PR's own premise — the next `session/load` for that id parks as a waiter and is served by the pre-reset load response. The gap predates this diff (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:320`); a separate issue would suit it.
- 🟢 **Suggestion:** `extractResultSessionId` still accepts `""` as a session id, while `extractParamsSessionId` rejects it since `c57ebf9`. The delta does not touch that function (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:761`).
- ✅ **Looks good:** The two abort paths share `abortRequestsOf`, and `pending.delete(id)` now precedes `sendToAgent`, so the map is consistent before the agent is told.

- 🟢 **Suggestion:** The session reset path now answers each open agent request with a structured error. The architecture page records the reset only as a drop of everything held for the session id. The same page names this agent-facing error outcome for the TTL expiry path (`docs/architecture/agent-lifecycle.md:183`).

### Issue Fit
✅ No findings.

- 🟢 **Suggestion:** `abortRequestsOf` returns a count, but `forget` discards it. The `expire` path uses the same value to gate `deps.onExpired()` (`pending-agent-requests.ts:149`).
- 🟢 **Suggestion:** The abort reason travels as a free `message: string` parameter, so both call sites embed their prose inline. Two named constants would keep the protocol messages together (`pending-agent-requests.ts:80`).
- 🟢 **Suggestion:** `updateExpiryTimer` reads `expiryTimers.get(sessionId)` for a presence test where `has` states the intent (`pending-agent-requests.ts:70`).
- 🟢 **Suggestion:** The skill read the test "should not ask an old question again after a reset" as passing on the parent branch too. It does not: `engage()` calls `pendingRequests.onEngaged` on bob's `session/load` fall-through path, so the stale prompt reaches bob without the fix (`__tests__/acp-runtime-unanswered-requests.test.ts:189`).

### Verdict
APPROVE — the fix closes each of the three reported symptoms at its cause, the two abort paths are unified, and the three scenarios cover them; nothing blocking remains.

---

PR #3384, opened 2026-08-20
perf(agents): bump claude-agent-acp pin to 0.66.0

jezekra1:
## Summary

One-line pin bump that removes the dominant session-latency stall found in #3247's benchmark (epic #3242).

Deployed 0.59.0 unconditionally awaits a context-window control request that fresh sessions never service until their first turn (upstream agentclientprotocol/claude-agent-acp#886, fixed by #894 in 0.60.0). The observable damage on our platform, measured with the #3247 benchmark and a local adapter probe:

| operation | 0.59.0 (deployed) | 0.66.0 + CLI 2.1.210 |
|---|---|---|
| `session/new` | 4s / 12.8s / 68.9s (variable stall) | ~0.5s |
| cold `session/load` (short) | 2.6-3.9s | ~0.5s, full correct replay |

Every new chat, fresh-schedule fire, and channel thread pays the session/new cost today; every cold conversation open pays the load cost.

- 0.66.0 (not 0.69.0): newest version outside the repo's 7-day `minimumReleaseAge` window, and the exact version verified locally
- verified against the image's pinned claude-code CLI **2.1.210** (the adapter drives whatever `CLAUDE_CODE_EXECUTABLE` points at) — session/new, cold session/load, and replay event kinds/counts all correct
- `harness-tools.toml` is COPY'd into the image before `mise install`, so the pin alone changes the built adapter

## Test plan

- [x] local probe: adapter 0.66.0 + CLI 2.1.210 — session/new ~0.5s, cold load ~0.5s, 22/22 replay events with correct kinds
- [ ] CI image build
- [ ] after dev deploy: re-run `mise run bench:session-load` (#3330) against the dev bench agent to confirm the cold-load numbers move

Full findings: https://github.com/dam-agents/dam/issues/3247#issuecomment-5294917381

PR #3400, opened 2026-08-20
feat: fast session opens — tail replay, single connection, history provider

jezekra1:
> **Reviewer walkthrough:** https://claude.ai/code/artifact/8185f89a-6db4-4be2-82f2-773feb8235d5 — a guided tour of the design, every decision with its alternatives, and the problems hit along the way.

## Summary

Consolidates the former stack #3387 ← #3390 ← #3393 into one reviewable unit against main (the stack's intermediate designs were superseded mid-flight; this PR carries only the final shape). Implements #3385 and #3386 under epic #3242. **#3384 (adapter pin bump) stays separate and should merge first.**

**Runtime (harness-agnostic)**
- `session/load` with `_meta.platform.tail` replays only the newest tail of the session log (default 200 events); without the opt-in, full replay per the ACP v1 contract, so unknown ACP clients stay conformant. Responses carry `_meta.platform.clipped` (`{olderBefore}` while pageable, `{}` at the eviction floor); `_meta.platform.replayBefore` pages older ranges. Stale (pre-restart) cursors are refused.
- Cold loads fill the transcript silently — preferably via the image's **session-history provider** (`runtime-manifest sessionHistory`: a `module` hosted in one persistent worker thread with crash/timeout isolation and respawn, or an exec `command`; any failure falls back to the harness load). The harness is rehydrated lazily on the first prompt: silent load with replay dropped, prompts held and forwarded in order; a rehydrate error fails held prompts honestly and keeps the session cold for retry.
- claude-code implements the provider by reusing the adapter's own exported replay pipeline — history is byte-identical to a harness replay.

**UI**
- One WebSocket per session click: the live connection performs the history load (per-session update collector), stays engaged (no resume round-trip), runs the list reconcile, and serves infinite-scroll pages. Conversations open pinned to the bottom before first paint; prepend compensation runs in a layout effect with native scroll anchoring disabled. `ChatMessage` is memoized; the entrance animation is scoped to the streaming tail.

## Measured (local cluster, 1,760-event conversation)

| | before stack | after |
|---|---|---|
| cold open | 2.6–4 s | **~90–110 ms** (+~35 ms connect; first load after pod boot ~260 ms) |
| warm open | 1.7 s+ | ~6 ms |
| scroll page | — | 6–14 ms |

Methodology and dev numbers: #3247 / #3330.

## Test plan

- [x] agent-runtime 217/217 (12 new scenarios: tail cap + response cursor, no-opt-in full replay, paging to start, eviction floor, stale cursor, silent cold fill, provider serve/fallback, lazy rehydrate without duplication, prompt hold ordering, rehydrate-failure retry, worker provider lifecycle), UI 302/302, api-server-api, docs size, comment types
- [x] end-to-end on a local cluster: cold/warm/paging measured through relay+runtime+provider; UI verified (bottom-pinned open, scroll paging, part-numbered fixture proves ordering)
- [ ] CI
- [ ] after dev deploy: rerun the #3330 benchmark for the epic writeup


packages/ui/src/modules/sessions/views/chat-view.tsx
```diff
@@ -229,6 +231,31 @@ export function ChatView() {
...
+  );
+
+  const loadOlderKeepingScroll = useCallback(
+    async (before: number) => {
+      const el = messagesRef.current;
+      pendingPrependRef.current = el
+        ? { height: el.scrollHeight, top: el.scrollTop }
+        : null;
+      const outcome = await loadOlderMessages(before);
+      pendingPrependRef.current = null;
```
dam-code-guardian: 🔴 **Critical:** `pendingPrependRef` is cleared in the microtask right after the `await`, before React commits the prepended page. The `useLayoutEffect` keyed on `messages` then reads `null` and bails, so the viewport jumps on every older page.

**Fix:** A scroll anchor captured before an async list mutation is cleared by the commit-phase consumer that uses it, never by the async caller — the state update renders after any post-`await` cleanup.

Reproduced with React 19 + jsdom against these two hooks verbatim: the effect logged `pending=null` and no `scrollTop` write occurred.

packages/ui/src/modules/sessions/hooks/use-acp-connection.ts
```diff
@@ -234,46 +242,114 @@ export function useAcpConnection(
...
+    connectionRef.current = { connection, ws };
+    return connectionRef.current;
+  }, [selectedAgent, makeUpdateHandler, attachCloseHandler]);
+
+  const loadSessionHistory = useCallback(
+    async (sid: string, replayBefore?: number): Promise<Message[]> => {
+      const generation = generationRef.current;
+      const live = await openConnection();
+      const collector = { sid, updates: [] as AcpUpdate[] };
+      collectorRef.current = collector;
```
dam-code-guardian, line 284: 🔴 **Critical:** The collector is installed for the whole duration of a `replayBefore` load and captures **every** frame for that session, not only the replayed range. `replayPage` moves no cursor while the channel stays engaged, so a live streaming chunk arriving mid-fetch is collected into the page and `loadOlderMessages` prepends it *above* the oldest message.

**Fix:** Interception during a paged load admits only frames of the requested range; frames outside it stay on the live handler.

Reproduced by running the collector dispatch and the `[...page, ...current.filter(...)]` prepend verbatim: a chunk injected mid-fetch landed third from the top and the live view received zero updates.

packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts
```diff
@@ -497,12 +579,11 @@ export function createAcpRuntime(deps: AcpRuntimeDeps): AcpRuntime {
 
     const sessionId = extractParamsSessionId(frame);
     if (sessionId) {
+      if (rehydratingSessions.has(sessionId)) {
```
dam-code-guardian: 🟡 **Warning:** Every `session/update` for a rehydrating session returns early, so live activity in that window is discarded rather than appended to the transcript. The suppression is meant to hide the replay, but it also drops durability.

**Fix:** A replay-suppression window suppresses fan-out to channels only; the transcript append that owns durability still happens.

packages/ui/src/modules/sessions/components/chat-message.tsx
```diff
@@ -13,16 +15,58 @@ interface Props {
...
+  const sentinelRef = useRef<HTMLDivElement>(null);
+  const startedRef = useRef(false);
+
+  useEffect(() => {
+    const el = sentinelRef.current;
+    if (!el) return;
+    const observer = new IntersectionObserver((entries) => {
+      if (startedRef.current) return;
+      if (!entries.some((entry) => entry.isIntersecting)) return;
+      startedRef.current = true;
```
dam-code-guardian: 🟡 **Warning:** `startedRef` latches before the load resolves. On a `"noop"` outcome the list is never rewritten, so the marker keeps its `message.id`, does not remount, and the fixed text "Loading earlier messages…" stays forever with no retry.

**Fix:** A sentinel that latches a one-shot guard resets it on a failed outcome and renders a distinct idle or error state.

(The remount-per-page reasoning does hold for the *successful* path — a fresh `message.id` gives a fresh ref. It is the `"noop"` path, where no new list is produced, that latches.)

packages/ui/src/modules/sessions/views/chat-view.tsx
```diff
@@ -576,7 +612,10 @@ export function ChatView() {
           ) : (
             <>
               <div className="relative flex flex-1 flex-col min-h-0">
-                <div ref={messagesRef} className="flex-1 overflow-y-auto">
+                <div
+                  ref={messagesRef}
+                  className="flex-1 overflow-y-auto [overflow-anchor:none]"
```
dam-code-guardian: 🟡 **Warning:** `[overflow-anchor:none]` applies to the whole thread for its whole lifetime, not just the prepending commit. Late-growing content — images without intrinsic dimensions, a collapsing thought block — now shifts content under a reader scrolled up in history, since the `ResizeObserver` re-pins only when already at the bottom.

**Fix:** Suppress scroll anchoring only for the commit that prepends, or leave it on and reconcile against it.

packages/agent-runtime/src/modules/acp/infrastructure/history-provider.ts
```diff
@@ -0,0 +1,255 @@
...
+          settled = true;
+          clearTimeout(timer);
+          if (reason) deps.log(`history provider for ${sessionId}: ${reason}`);
+          resolve(lines);
+        };
+        const timer = setTimeout(() => {
+          child.kill("SIGKILL");
+          finish(null, `timed out after ${timeoutMs}ms`);
+        }, timeoutMs);
+        child.stdout.on("data", (d) => (stdout += d.toString()));
```
dam-code-guardian: 🟡 **Warning:** The child's stdout accumulates into one unbounded string before validation. A very large replay can exhaust the runtime heap, and the timeout bounds only elapsed time.

**Fix:** An adapter reading an external process bounds accumulated output by bytes as well as by time.

packages/agent-runtime/src/modules/acp/infrastructure/history-provider.ts
```diff
@@ -0,0 +1,255 @@
...
+    worker.on("error", dropWorker);
+    worker.on("exit", dropWorker);
+    handle = created;
+    return created;
+  }
+
+  return {
+    async fetch(sessionId) {
+      const current = ensureWorker();
+      const ok = await current.ready;
```
dam-code-guardian, line 244: 🟡 **Warning:** A module that fails to import resolves `ready` to `false` and the handle is dropped, but the worker thread itself is never terminated. A repeatedly failing module leaves one idle thread per attempt.

**Fix:** Every failure path of an adapter that owns an OS resource releases that resource, not only the reference to it.

packages/agent-runtime/src/modules/acp/services/acp-runtime/session-bootstrap.ts
```diff
@@ -91,55 +138,107 @@ export function createSessionBootstrap(
...
+      boot.waiters.push(waiter);
+      return;
+    }
+    bootstrapBySession.set(sessionId, { waiters: [waiter] });
+    const provider = deps.historyProvider;
+    if (!provider) {
+      startHarnessLoad(sessionId);
+      return;
+    }
+    void provider.fetch(sessionId).then((lines) => {
```
dam-code-guardian, line 251: 🟡 **Warning:** The `.then` re-checks `bootstrapBySession.has(sessionId)` rather than the state object it captured. A teardown plus a new cold load for the same session between fetch start and resolution lets stale provider lines fill the new bootstrap's transcript.

**Fix:** An async continuation re-checks identity of the state object it captured, not just presence of its key, before mutating shared state.

packages/ui/src/modules/sessions/hooks/use-acp-connection.ts
```diff
@@ -234,46 +242,114 @@ export function useAcpConnection(
...
 
-    if (pendingReloadRef.current) {
-      const sid = useStore.getState().sessionId;
-      pendingReloadRef.current = false;
-      if (sid) {
-        setState("reloading");
-        try {
-          const fresh = await loadHistory(sid);
+    const needsReload = pendingReloadRef.current;
+    pendingReloadRef.current = false;
```
dam-code-guardian: 🟡 **Warning:** `pendingReloadRef` is cleared before `openConnection()` awaits, and only the `catch` around `loadSessionHistory` restores it. A failure in the connection open therefore discards the pending reload silently.

**Fix:** A pending-work flag is cleared only after the work it guards has completed, never before the first await.

packages/ui/src/modules/sessions/hooks/use-acp-connection.ts
```diff
@@ -234,46 +242,114 @@ export function useAcpConnection(
...
+          sessionId: sid,
+          cwd: ".",
+          mcpServers: [],
+          _meta: {
+            platform:
+              replayBefore !== undefined ? { replayBefore } : { tail: true },
+          },
+        });
+      } finally {
+        if (collectorRef.current === collector) collectorRef.current = null;
```
dam-code-guardian, line 299: 🟡 **Warning:** `collectorRef` is a single slot, so two concurrent history loads for one session share it and the first `finally` clears it. The second load then loses its replay frames to the live handler.

**Fix:** Per-request interception state is one entry per in-flight request, not one shared slot.

packages/agent-runtime/src/modules/acp/compose.ts
```diff
@@ -26,11 +31,38 @@ export interface ComposeAcpOptions {
...
   backgroundWorkHolds?: boolean;
   log?: (msg: string) => void;
 }
 
+function historyProviderOf(
+  opts: ComposeAcpOptions,
+): HistoryProvider | undefined {
+  const declared = opts.sessionHistory;
+  const log = (msg: string): void => opts.log?.(msg);
+  if (declared?.module !== undefined) {
```
dam-code-guardian, line 49: 🟢 **Suggestion:** When a manifest declares both `module` and `command`, `module` wins with no signal. The schema permits both, so the ignored key is invisible to the image author — a log line or a schema refinement would surface it.

packages/ui/src/modules/sessions/hooks/use-acp-session.ts
```diff
@@ -134,7 +134,39 @@ export function useAcpSession(
...
+
+  const loadOlderMessages = useCallback(
+    async (before: number): Promise<"paged" | "reloaded" | "noop"> => {
+      const sid = useStore.getState().sessionId;
+      if (!sid) return "noop";
+      try {
+        const page = await loadSessionHistory(sid, before);
+        if (useStore.getState().sessionId !== sid) return "noop";
+        const current = useStore.getState().messages;
+        setMessages([
```
dam-code-guardian, line 155: 🟢 **Suggestion:** The `catch` fallback reload can itself reject, and `.catch(() => null)` swallows it, so a page failure and a total failure both surface as `"noop"` with nothing shown to the user.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `6506c2b`

## PR #3400: feat: fast session opens — tail replay, single connection, history provider
**Author:** jezekra1 | **Branch:** feat/session-open-latency → main | **Changes:** +1755 −219 (25 files)

### Summary
Bounds conversation-open cost: `session/load` replays only the newest tail with cursor-based paging for older ranges, the UI serves history and live updates over one WebSocket, and an image may declare a session-history provider that fills the transcript without spawning the harness. The measured numbers and the bottom-pinned open both hold up. _(Suppressed 1 finding per PR context: the entrance animation scoped to the streaming tail is declared in the PR body.)_

### Findings
- 🔴 **Critical:** Prepend scroll compensation never runs — the anchor is cleared before React commits the prepended page (`packages/ui/src/modules/sessions/views/chat-view.tsx:245`)
- 🔴 **Critical:** A live update arriving during a page load is diverted into the page collector and prepended above the oldest message (`packages/ui/src/modules/sessions/hooks/use-acp-connection.ts:275`)
- 🟡 **Warning:** Live frames are dropped, not just hidden, during a harness rehydrate (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:582`)
- 🟡 **Warning:** A failed page load leaves a permanent "Loading earlier messages…" state with no retry (`packages/ui/src/modules/sessions/components/chat-message.tsx:37`)
- 🟡 **Warning:** `replayPage` emits log entries without moving the channel's cursor, so a later `catchUp` on that channel re-sends them (`packages/agent-runtime/src/modules/acp/services/acp-runtime/session-transcript.ts:174`)
  **Fix:** A read path that emits log entries either advances the reader's cursor or is enforced as never followed by a cursor-based catch-up on the same channel.
- 🟡 **Warning:** `[overflow-anchor:none]` is permanent on the thread, not scoped to the prepend commit, so late-growing content shifts under a reader scrolled up in history (`packages/ui/src/modules/sessions/views/chat-view.tsx:617`)
- 🟡 **Warning:** The exec provider accumulates the child's whole stdout unbounded before validating it; the timeout bounds elapsed time only (`packages/agent-runtime/src/modules/acp/infrastructure/history-provider.ts:79`)
- 🟡 **Warning:** A worker whose module fails to import resolves `ready: false` but is never terminated, leaving one idle thread per attempt (`packages/agent-runtime/src/modules/acp/infrastructure/history-provider.ts:226`)
- 🟡 **Warning:** `park`'s provider continuation re-checks the session key, not the state object it captured (`packages/agent-runtime/src/modules/acp/services/acp-runtime/session-bootstrap.ts:192`)
- 🟡 **Warning:** `pendingReloadRef` is cleared before the first await, so a connection-open failure discards the pending reload (`packages/ui/src/modules/sessions/hooks/use-acp-connection.ts:330`)
- 🟡 **Warning:** One shared collector slot per session loses replay frames when two loads for that session overlap (`packages/ui/src/modules/sessions/hooks/use-acp-connection.ts:288`)
- 🟢 **Suggestion:** Provider-synthesized metadata (`modes: null`, `configOptions: null`) is cached as authoritative and is indistinguishable from real metadata to the client; only the producer sees the `synthetic` flag.
- 🟢 **Suggestion:** `validateLines` returns `null` for any single malformed or foreign frame, discarding an otherwise valid replay and the whole acceleration benefit (`packages/agent-runtime/src/modules/acp/infrastructure/history-provider.ts:21`).
- 🟢 **Suggestion:** `historyProviderOf` prefers `module` over `command` silently when a manifest declares both; the schema permits both, so the ignored key is invisible to the image author (`packages/agent-runtime/src/modules/acp/compose.ts:49`).
- ✅ **Looks good:** The bottom-pinned open is correct. The new layout effect keys on `loadingSession`, which flips in a later commit than `setMessages(fresh)`, so the pin runs after the list lands rather than against an empty container.

- 🟡 **Warning:** The manifest section states the runtime manifest declares only `drivers` and optional `extensions`. This PR adds a third top-level block to the schema and to the claude-code manifest (`docs/architecture/runtime-delivery.md:242`).
  **Fix:** A PR that adds a declarable block to the runtime manifest records that block in the manifest section of `docs/architecture/runtime-delivery.md`.
- 🟡 **Warning:** The UI description says a tab holds short-lived ACP channels for history replay. This PR removes that separate connection; replay and the session list now ride the single live channel (`docs/architecture/platform-topology.md:75`).
  **Fix:** A PR that changes how many ACP channels a UI tab holds per Agent updates the UI paragraph of `docs/architecture/platform-topology.md`.

### Issue Fit
- 🟡 **Warning:** The provider subsystem — about 420 lines of runtime code plus an image library, Dockerfile symlink, and manifest schema key — is named by neither #3385 nor #3386, and #3385 fences out "harness process startup". Its own Open Questions do put the cold-load path in scope, so the boundary is arguable rather than clearly crossed (`packages/agent-runtime/src/modules/acp/infrastructure/history-provider.ts:1`).
  **Fix:** A subsystem no linked issue names is tracked by its own issue, or the owning issue's scope is amended, so its acceptance criteria and rollout risk are reviewable on their own.
- 🟢 **Suggestion:** #3386's third "Done when" asks for the #3247 connect phase verified once per open **on dev**. The measurements are local-cluster only, and the dev benchmark rerun is still an unchecked test-plan box.
- 🟢 **Suggestion:** ADR 026 still documents the replaced mechanism — the truncation sentinel and a "Load more" forced full reload as an unbuilt follow-up. This PR delivers that follow-up; only `docs/architecture/agent-lifecycle.md` was updated (`docs/adrs/026-session-log-replay.md:171`).
- 🟢 **Suggestion:** The image library depends on adapter internals resolved through a `mise where` symlink, so this PR is not independently mergeable ahead of #3384 (`packages/agents/claude-code/harness-history-lib.mjs:1`).

Findings merged into `### Findings` above: the rehydrate frame drop, `replayPage` cursor, unbounded exec stdout, un-terminated worker, `park` identity re-check, `pendingReloadRef` clear-before-await, shared collector slot, synthesized metadata, `validateLines` all-or-nothing, and the silent `module`/`command` precedence.
- 🟡 **Warning:** `finishHarnessRehydrate` replays held prompts through `handleClientMessage`, which re-enters the cold-session branch. A concurrent provider-served load for that session can re-add the flag and produce a hold/rehydrate cycle (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:192`).
  **Fix:** Re-entrant replay of buffered input carries a per-item replay marker, so a held frame is processed at most once.
- 🟢 **Suggestion:** `loadOlderMessages` falls back to a full reload inside its `catch`, whose own rejection is swallowed by `.catch(() => null)`; both failures surface as `"noop"` with no signal (`packages/ui/src/modules/sessions/hooks/use-acp-session.ts:160`).
- 🟢 **Suggestion:** `withClipMeta` emits `clipped: {}` for "clipped but not pageable", which an empty object and a missing key make indistinguishable to a future consumer (`packages/agent-runtime/src/modules/acp/services/acp-runtime/session-bootstrap.ts:103`).

Findings merged into `### Findings` above: the prepend anchor cleared before commit (independently confirmed), the permanent loading marker, and the unscoped `[overflow-anchor:none]`.
- 🟢 **Suggestion:** `LoadOlderMarker` owns pagination-sentinel concerns — an `IntersectionObserver`, a one-shot guard, its own effect lifecycle — inside the shared message-row file rather than its own module (`packages/ui/src/modules/sessions/components/chat-message.tsx:21`).

### Verdict
REQUEST_CHANGES — two confirmed criticals in the paging path: the scroll anchor never survives to the layout effect, and a live chunk arriving mid-fetch is prepended above the oldest message.

---

jezekra1:
Review findings addressed in 487e3143d:

- **#1 rehydrate timeout** — a hanging harness rehydrate now times out (30s default, `rehydrateTimeoutMs` dep): held prompts get a timeout error and the session stays cold, so the next prompt retries. Covered by a new test (hang → timeout error → retry starts a fresh rehydrate).
- **#2 collector collision** — the UI now serializes `session/load` calls on the live connection with a promise chain. The runtime already serves loads strictly one at a time, so this costs nothing and makes the single-collector design structurally safe against the event-queue race.
- **#3 worker leak** — a worker whose module fails to import is terminated instead of abandoned alive.
- **#4 all-or-nothing validation** — keeping the behavior: skipping invalid lines would serve incomplete history as if complete; reject-and-fallback is slow but correct. The missing piece is a fixture-based smoke test inside the built image proving real `loadHistory` output passes validation — planned as a follow-up since it needs the mise-installed adapter and can't run in package unit tests.
- **#5 no cancellation when waiters drop** — accepted as-is: by the time the last waiter drops, the ~100ms of provider work is done; the result is a warm transcript (≤2MB until reuse or restart), effectively an accidental prefetch. Cancellation machinery would outweigh the win.

And yes to the dam-dev re-measurement offer once this merges.

jezekra1:
Inline review findings addressed in 6089520aa (plus two already fixed in 487e3143d):

**Fixed**

- **🔴 prepend anchor cleared before commit** (`chat-view.tsx`) — the layout effect that compensates now owns clearing the anchor; the async caller clears it only on non-`paged` outcomes. The anchor is also keyed to the page's cursor, so a live-append commit landing mid-fetch can't consume it early (a case one step beyond the reported one).
- **🔴 collector captures live frames during a paged load** (`use-acp-connection.ts`) — fixed with request marking rather than range admission (frames carry no range on the wire): each `session/load` sends a client-minted `loadToken`, the runtime stamps every frame replayed for it with `_meta.platform.replayFor`, and the collector admits only its own token. Live frames flow to the live handler even mid-load. Loads without a token get verbatim frames.
- **🟡 `overflow-anchor:none` permanent** — now applied only from anchor capture to the prepending commit.
- **🟡 marker latches on `noop`** — the one-shot guard resets on failure and the marker renders "Couldn't load earlier messages — tap to retry" (also covers the 🟢 about the swallowed double-failure).
- **🟡 `pendingReloadRef` cleared before first await** — the flag is now consumed only by a successful reload; any failure leaves it set for the next ensure.
- **🟡 unbounded exec stdout** — capped at 32 MB, kill + fallback on exceed.
- **🟡 bootstrap `.then` re-checks key, not identity** — the continuation now compares the captured state object.
- **🟢 `module` + `command` both declared** — the manifest schema now rejects it (exactly one), with tests.

While in here we also hardened the cursor itself: it's now opaque and embeds the transcript generation, so a cursor held across a transcript rebuild (idle reap → re-warm by another viewer) gets a deterministic -32602 instead of paging from a renumbered log.

**Already fixed in 487e3143d** (review predates it)

- **🟡 worker not terminated on import failure** — terminated in the failed-ready branch.
- **🟡 single collector slot** — loads are serialized client-side; a per-request map can't work without frame attribution, which `replayFor` now also provides, so both mechanisms hold.

**Declined, with rationale**

- **🟡 rehydrate suppression drops transcript appends** — keeping the drop. During a rehydrate every frame for that session is harness *replay* of history the transcript already holds; appending would duplicate the log. Genuine live activity can't occur in that window by construction: all prompt paths are held, and the harness has no session context to emit from until the load completes.

packages/ui/src/modules/acp/acp.ts
```diff
@@ -145,7 +153,7 @@ export async function openConnection(
         return awaitPermission(params);
       },
       async sessionUpdate(params: SessionNotification) {
-        onUpdate(params.update, params.sessionId);
+        onUpdate(params.update, params.sessionId, replayForOf(params._meta));
```
dam-code-guardian, line 130: 🔴 **Critical:** `replayForOf` is wired into `sessionUpdate` only, on this line. `platform/turnEnded` reaches the client through `extNotification` (`acp.ts:173`), which calls `onUpdate` with two arguments, so `replayFor` is `undefined` there.

The runtime does append this frame to the transcript (`acp-runtime.ts:548`), so it is replayed. The collector in `use-acp-connection.ts:258` admits a frame only when `replayFor === collector.token`, so every replayed `turnEnded` fails the gate and reaches the live handler instead of the replay projection.

Reproduced in `node` against the ported projection: without `turnEnded` in the replay stream, two consecutive replayed turns merge into one assistant bubble left `streaming: true`. The escaped frame also runs `closeActiveAssistant` against the live list, which can end a genuinely streaming live turn.

`platform/promptAccepted` and `platform/promptStarted` are unaffected — the scheduler sends them to one channel and never appends them to the transcript.

**Fix:** Every frame the runtime can stamp must be able to carry the stamp to the collector. Read `replayForOf(params._meta)` in the three `extNotification` branches too and pass it as the third `onUpdate` argument, exactly as this line does.

packages/agent-runtime/src/modules/acp/infrastructure/history-provider.ts
```diff
@@ -0,0 +1,264 @@
...
+          clearTimeout(timer);
+          if (reason) deps.log(`history provider for ${sessionId}: ${reason}`);
+          resolve(lines);
+        };
+        const timer = setTimeout(() => {
+          child.kill("SIGKILL");
+          finish(null, `timed out after ${timeoutMs}ms`);
+        }, timeoutMs);
+        child.stdout.on("data", (d) => {
+          stdout += d.toString();
```
dam-code-guardian: 🔴 **Critical:** Two defects in this handler, one class — bytes and UTF-16 code units are treated as interchangeable.

`stdout.length` is a UTF-16 unit count, not a byte count, so the 32 MiB budget admits up to 3× that in bytes. Verified in `node`: 33,554,432 × `"日"` passes `stdout.length > MAX_EXEC_OUTPUT_BYTES` while occupying 100,663,296 bytes.

`d.toString()` per chunk also decodes each chunk independently. A chunk boundary inside a multi-byte sequence yields U+FFFD replacement characters. Verified in `node`: the line still parses as JSON, so `validateLines` accepts it and the replayed message text is silently corrupted — `"přeložený výstup — 日本語"` came back as `"přeložený výstup — ���本語"`. A `StringDecoder("utf8")` over the same two chunks reproduced the original exactly.

**Fix:** Never treat `String.length` as a byte count where a byte budget is enforced, and never decode a stream chunk-by-chunk. Accumulate through `new StringDecoder("utf8")` (or buffer and `Buffer.byteLength`), and measure the budget in bytes.

```suggestion
        const decoder = new StringDecoder("utf8");
        let stdoutBytes = 0;
        child.stdout.on("data", (d: Buffer) => {
          stdoutBytes += d.length;
          stdout += decoder.write(d);
          if (stdoutBytes > MAX_EXEC_OUTPUT_BYTES) {
            child.kill("SIGKILL");
            finish(null, `output exceeded ${MAX_EXEC_OUTPUT_BYTES} bytes`);
          }
        });
```

packages/ui/src/modules/sessions/views/chat-view.tsx
```diff
@@ -229,6 +231,44 @@ export function ChatView() {
...
+      return outcome;
+    },
+    [loadOlderMessages, scrollToBottom],
+  );
+
+  useLayoutEffect(() => {
+    const pending = pendingPrependRef.current;
+    const el = messagesRef.current;
+    if (!pending || !el) return;
+    if (messages.some((m) => m.loadOlderBefore === pending.before)) return;
```
dam-code-guardian: 🟡 **Warning:** The baseline is captured before the request starts, and this bail path never re-measures it. When the thread grows at the bottom during the in-flight page load — a streaming chunk, a tool chip — that growth is added to the prepend offset at the final restore.

Reproduced with the real element arithmetic: a 200 px bottom append during the load, then an 800 px prepend, leaves the viewport 200 px off; drift is 0 px when nothing grows. Streaming while reading history is a normal case for this PR.

**Fix:** A deferred scroll-restore baseline is re-measured on every commit it skips. Refresh `pending.height` and `pending.top` from the live element in this bail branch, so the final restore measures only the prepend growth.

packages/agent-runtime/src/modules/acp/services/acp-runtime/session-bootstrap.ts
```diff
@@ -91,55 +142,126 @@ export function createSessionBootstrap(
...
+    }
+  }
+
+  function park(sessionId: string, waiter: Waiter): void {
+    const boot = bootstrapBySession.get(sessionId);
+    if (boot) {
+      boot.waiters.push(waiter);
+      return;
+    }
+    const state: BootstrapState = { waiters: [waiter] };
```
dam-code-guardian: 🟡 **Warning:** A parked waiter has no deadline. If the harness never answers the cold `session/load`, `onLoadResponse` never fires, the state is never cleared, and the waiter's own `session/load` is never answered — the UI's `loadSession` promise settles only when the socket closes.

The 30 s guard added in this rework covers `startHarnessRehydrate`, not this path.

**Fix:** Every request-holding map needs a deadline. Arm a timer per `BootstrapState` that answers all its waiters with a JSON-RPC error and clears the state, mirroring the rehydrate timeout.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `f71a949`

## PR #3400: feat: fast session opens — tail replay, single connection, history provider
**Author:** jezekra1 | **Branch:** feat/session-open-latency → main | **Changes:** +2131 −222 (27 files)

### Summary
Bounds conversation-open cost: `session/load` replays only the newest tail, older ranges page in through an opaque cursor, the UI serves history over the one live connection, and an image may declare a session-history provider that fills the transcript with no harness process. The rework closes both criticals at their cause and hardens the cursor with a transcript generation. _(Suppressed 1 finding per PR context: the entrance animation scoped to the streaming tail is declared in the PR body.)_

### Changes since last review
Previous HEAD: `6506c2b` (2026-08-21T07:45:21Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** Prepend scroll compensation — the layout effect now owns clearing the anchor, keyed to the page cursor (`chat-view.tsx:266`)
- ✅ **Fixed:** A live update during a page load — each load mints a `loadToken`, the runtime stamps `replayFor`, the collector admits only its own token (`use-acp-connection.ts:258`)
- ✅ **Fixed:** A failed page load now resets the one-shot guard and renders a retry control (`chat-message.tsx:23`)
- ✅ **Fixed:** `replayPage` moves no cursor and refuses a stale generation (`session-transcript.ts:244`)
- ✅ **Fixed:** `[overflow-anchor:none]` applies only from anchor capture to the prepending commit (`chat-view.tsx:245`)
- ✅ **Fixed:** A worker that fails to become ready is terminated (`history-provider.ts:237`)
- ✅ **Fixed:** `park`'s continuation compares the captured state object (`session-bootstrap.ts:199`)
- ✅ **Fixed:** `pendingReloadRef` is consumed only by a successful reload (`use-acp-connection.ts:350`)
- ✅ **Fixed:** A rehydrate that hangs now times out at 30 s and keeps the session cold for retry (`acp-runtime.ts:183`)
- ✅ **Fixed:** `module` no longer silently wins over `command` (`manifest.ts:63`)
- ✅ **Fixed:** The exec provider's stdout is bounded — but the cap uses the wrong unit; see the 🔴 below
- ✅ **Fixed:** `harness-history-lib.mjs` is independently mergeable — #3384 merged and is an ancestor of this branch's base
- 🔁 **Still present:** `validateLines` rejects a whole replay for one bad frame (`history-provider.ts:12`) — declined, rationale recorded
- 🔁 **Still present:** `withClipMeta` emits an ambiguous empty `clipped: {}` (`session-bootstrap.ts:106`)
- 🔁 **Still present:** `loadOlderMessages` reports a double failure as `"noop"` (`use-acp-session.ts:166`)
- 🔁 **Still present:** Provider-synthesized metadata is indistinguishable from real metadata to the client
- 🔁 **Still present:** ADR 026 and two architecture pages still describe the replaced mechanism
- 🔁 **Still present:** The provider subsystem is named by no linked issue
- 🆕 **New:** Two criticals the rework introduced, plus one warning on the restore baseline — below

### Findings
- 🔴 **Critical:** `platform/turnEnded` cannot carry the replay stamp (`packages/ui/src/modules/acp/acp.ts:156`)
- 🔴 **Critical:** The exec provider's byte cap counts UTF-16 units, and chunk concatenation corrupts multi-byte text (`packages/agent-runtime/src/modules/acp/infrastructure/history-provider.ts:82`)
- 🟡 **Warning:** The scroll-restore baseline is never re-measured while the effect bails, so bottom growth during a page load inflates the offset (`packages/ui/src/modules/sessions/views/chat-view.tsx:266`)
- 🟡 **Warning:** A parked bootstrap waiter has no deadline (`packages/agent-runtime/src/modules/acp/services/acp-runtime/session-bootstrap.ts:191`)
  **Fix:** Every request-holding map needs a deadline. Arm a timer per `BootstrapState` that answers all its waiters with a JSON-RPC error, as `startHarnessRehydrate` now does.
- 🟢 **Suggestion:** A manifest declaring both `module` and `command` now fails the whole runtime at boot rather than being tolerated (`packages/agent-runtime/src/modules/runtime-channel/manifest.ts:63`). Fail-fast on a malformed manifest is the documented convention (`docs/architecture/runtime-delivery.md:238`), so this is consistent — worth a note in the manifest docs so an image author reads it before the crash.
- 🟢 **Suggestion:** The stick-to-bottom `ResizeObserver` sets `scrollTop = scrollHeight` whenever `stickRef` is true (`packages/ui/src/modules/sessions/views/chat-view.tsx:288`). A short replayed tail can leave the sentinel visible while the view is near the bottom, so the prepend growth triggers the observer and discards the restore. The observer is outside this delta — a separate issue.
- 🟢 **Suggestion:** `decodePageCursor` accepts any base64url-decodable text and only fails on the `:` check (`packages/agent-runtime/src/modules/acp/services/acp-runtime/session-transcript.ts:74`). A version prefix in the encoded payload would make a foreign cursor unambiguous.
- 🟢 **Suggestion:** `LoadOlderMarker` now owns fetch state as well as observation — an `IntersectionObserver`, a one-shot guard, a `failed` state and a retry control — inside the shared message-row file (`packages/ui/src/modules/sessions/components/chat-message.tsx:23`).

🔁 2 findings from the previous review still present (see review at `6506c2b`): the runtime-manifest section omits the new `sessionHistory` block (`docs/architecture/runtime-delivery.md:242`), and the UI paragraph still says a tab holds short-lived ACP channels for history replay (`docs/architecture/platform-topology.md:75`).

### Issue Fit
🔁 3 findings from the previous review still present (see review at `6506c2b`): the provider subsystem is named by neither #3385 nor #3386 (`packages/agent-runtime/src/modules/acp/infrastructure/history-provider.ts:1`), #3386's dev-environment "Done when" is still unmet, and ADR 026 still documents the replaced truncation-sentinel mechanism (`docs/adrs/026-session-log-replay.md:171`).

- 🟢 **Suggestion:** The independent-mergeability concern is resolved. #3384 merged and is an ancestor of this branch's base, and `harness-tools.toml` pins the adapter at 0.66.0 (`packages/agents/claude-code/Dockerfile:53`).

Findings merged into `### Findings` above: the `turnEnded` replay stamp, the exec byte cap, the parked-waiter deadline, the cursor version prefix, and the manifest boot failure.

🔁 3 findings from the previous review still present (see review at `6506c2b`): `validateLines` all-or-nothing (`history-provider.ts:12`), `loadOlderMessages` double failure as `"noop"` (`use-acp-session.ts:166`), and the ambiguous `clipped: {}` (`session-bootstrap.ts:106`). The first is declined with a stated rationale — reject-and-fallback is correct, incomplete-as-complete is not.

- 🟢 **Suggestion:** `cacheMetadata`'s `replaceable && !(log.metadata.cached && synthetic)` mixes two conditions whose combined truth table is not obvious (`packages/agent-runtime/src/modules/acp/services/acp-runtime/session-transcript.ts:282`).

Findings merged into `### Findings` above: the restore baseline, the `ResizeObserver` interaction, and the `LoadOlderMarker` placement.

All three prior blocking findings verified fixed with a real-React repro (React 19 + jsdom): the restore effect runs after the page commits and applies the correct offset; the retry button refetches and clears the failed state; both exit paths clear the imperative `overflow-anchor`.

### Verdict
REQUEST_CHANGES — the rework closed every prior blocking finding at its cause, but introduced two confirmed criticals: a replayed `platform/turnEnded` cannot carry the `replayFor` stamp, so it escapes the collector and collapses replayed turns into one bubble; and the exec provider's 32 MiB cap counts UTF-16 units while its chunk concatenation corrupts multi-byte text.

---

jezekra1:
All four findings addressed in d90879c76 (amended into the squashed commit), each with a test that would have caught it:

- **🔴 replayFor dropped on `extNotification`** — fixed, and the routing is restructured so this class can't silently recur: extension-notification dispatch is extracted into a pure `ext-notifications.ts` module that returns `{update, sessionId, replayFor}` for every route, unit-tested directly (stamped → token surfaces; unstamped → undefined; malformed → null). `startSession`'s update callback also threads the third argument now — it was unreachable-but-latent, the same defect one refactor away.
- **🔴 bytes vs UTF-16 units + chunk-boundary decode** — fixed with `StringDecoder` and byte-true accounting as suggested, plus `decoder.end()` on close (the suggested snippet would drop a trailing partial sequence) and the same decoder on stderr. The budget is injectable (`maxOutputBytes`) so both are covered by real subprocess tests: a three-byte character split across two chunks reassembles exactly, and exceeding the byte budget fails closed.
- **🟡 stale scroll-restore baseline** — fixed one step past the suggestion: the bail branch refreshes the height baseline on every commit it skips, and the final restore compensates from the element's **live** `scrollTop` instead of the captured one, which also covers the user continuing to scroll while the page loads (drift the re-measure alone wouldn't fix).
- **🟡 parked waiter has no deadline** — fixed: each cold fill arms one deadline spanning the provider attempt and any harness fallback; on expiry every waiter gets the same -32000 timeout error the rehydrate path uses, state clears, and the next load starts fresh. Covered for both the harness-never-answers and provider-hangs cases. One note on framing: this path predates this PR (the rehydrate timeout was added earlier because rehydrate was new machinery; the cold-load park is inherited) — fixed here anyway since this PR owns session-open behavior. The config knob is now `harnessLoadTimeoutMs`, since it governs both deadlines. Known accepted edge, same as the rehydrate timeout: a harness that answers *after* the deadline has its late replay treated as live traffic; recovery from a zombie harness stays best-effort.

Runtime 237/237, UI 306/306 (4 new routing tests, 2 new provider tests, 2 new deadline tests).

packages/agent-runtime/src/modules/acp/services/acp-runtime/session-bootstrap.ts
```diff
@@ -91,60 +147,160 @@ export function createSessionBootstrap(
...
+        waiter.channel,
+        waiter.originalId,
+        sessionId,
+        waiter.tail,
+        waiter.loadToken,
+      );
+    }
+  }
+
+  function expireFill(sessionId: string, state: BootstrapState): void {
```
dam-code-guardian, line 207: 🔴 **Critical:** `expireFill` answers the waiters and deletes the bootstrap state, but the `session/load` route it opened stays in `outboundIdToClient`. With the state gone, `bootstrap.has(sessionId)` is false, so `handleAgentLine` (`acp-runtime.ts:567`) sends that load's replay through `transcript.append` — live fan-out — rather than `appendReplay`.

Reachable via the retry the code already has: `loadOlderMessages` retries on failure (`use-acp-session.ts:161`), and a timeout surfaces as a session error the user reopens. Modelling the traced control flow (park / expireFill / onLoadResponse / has, plus `fanOut`'s cursor test) over three scenarios: with no retry the channel is never engaged and nothing leaks; after a retry, load #1's answer serves the waiter and engages the channel, then load #2's own replay arrives with no bootstrap state and re-delivers the whole conversation as 3/3 live updates. The no-timeout control leaked 0.

**Fix:** A silent-fill path must stay silent for the lifetime of the request that opened it, not the lifetime of its waiters. Keep the session marked as filling until its harness load actually answers — or close the outbound route in `expireFill` — so a late replay can never be reclassified as live.

packages/agent-runtime/src/modules/acp/services/acp-runtime/session-bootstrap.ts
```diff
@@ -43,42 +52,89 @@ export interface SessionBootstrapDeps {
...
+      typeof meta.platform === "object" && meta.platform !== null
+        ? (meta.platform as Record<string, unknown>)
+        : {};
+    return {
+      ...base,
+      _meta: {
+        ...meta,
+        platform: {
+          ...platform,
+          clipped: clip.older !== undefined ? { older: clip.older } : {},
```
dam-code-guardian, line 120: 🟡 **Warning:** `clipped: {}` is emitted when the replay was clipped by eviction and no `older` cursor exists, which on the wire is indistinguishable from a present-but-empty marker. `platformClippedReplayMetaSchema` accepts it and `session-projection.ts` falls back to the generic "Older conversation not loaded" notice. Carried from the review at `f71a949`.

**Fix:** Represent "clipped, no pageable older range" with an explicit discriminant in the shared schema, so presence and pageability stay separate signals.

packages/ui/src/modules/sessions/hooks/use-acp-session.ts
```diff
@@ -134,7 +134,39 @@ export function useAcpSession(
...
+          ...current.filter((m) => m.loadOlderBefore !== before),
+        ]);
+        return "paged";
+      } catch {
+        const fresh = await loadSessionHistory(sid).catch(() => null);
+        if (fresh && useStore.getState().sessionId === sid) {
+          setMessages(fresh);
+          return "reloaded";
+        }
+        return "noop";
```
dam-code-guardian, line 166: 🟡 **Warning:** `loadOlderMessages` returns `"noop"` both when nothing was attempted (no session id) and when the page load *and* the fallback reload both failed. `LoadOlderMarker` maps every `"noop"` to a retry-able failed state (`chat-message.tsx:38`), so a no-session case renders as an error. Carried from the review at `f71a949`.

**Fix:** Give the outcome union a distinct member for "both attempts failed" versus "nothing to do", so each caller branch maps one state to one piece of UI.

docs/architecture/agent-lifecycle.md
```diff
@@ -158,9 +160,9 @@ times a hold out, so an agent with reported work does not scale to zero until
...
+Terminal-mode sessions follow a different model from the chat path above. agent-runtime accepts at most one WebSocket per `sessionId` on `/api/terminal`, allocates a PTY, spawns `harness-terminal` attached to it, and pipes raw bytes both ways through a small binary frame protocol. A headless xterm tracks scrollback so that reattaching while the PTY lives replays the serialized buffer. A detached PTY is reaped on idleness, not on viewer loss: after a short detach grace (30 s), it is killed only once the harness has also been quiet — no output for five minutes. Liveness keys on harness output rather than viewer presence, so in-flight work (a running build, a streaming response) survives switching away and can be reattached live, while an abandoned idle prompt is cleaned up. There is no append-only log, no fan-out, and no `session/resume` — terminal sessions belong to one viewer at a time, and the harness's own on-disk session store is the only durable record.
```
dam-code-guardian, line 163: 🟡 **Warning:** This edit also removes facts from subsystems the PR does not change: the terminal frame opcodes (`OP_INPUT`/`OP_OUTPUT`/`OP_RESIZE`/`OP_EXIT`) and the detach-grace rationale here, SSH's `PermitUserEnvironment yes` below, and the `session/close` mechanics of the terminal-reset path. The page lands at 39,996 of the 40,000-char cap, so the cap plausibly forced the trimming.

**Fix:** Doc trimming outside the changed subsystem is declared in the PR body when it rides along with a feature, so a reviewer can tell an intended compression from a lost fact.

packages/agent-runtime/src/modules/acp/services/acp-runtime/session-transcript.ts
```diff
@@ -38,14 +63,72 @@ export interface SessionTranscript {
...
+): { generation: string; seq: number } | null {
+  const text = Buffer.from(cursor, "base64url").toString("utf8");
+  const at = text.lastIndexOf(":");
+  if (at <= 0) return null;
+  const seq = Number(text.slice(at + 1));
+  if (!Number.isInteger(seq) || seq <= 0) return null;
+  return { generation: text.slice(0, at), seq };
+}
+
+function stampReplayFor(line: string, token: string | undefined): string {
```
dam-code-guardian, line 85: 🟢 **Suggestion:** `stampReplayFor` here, `withClipMeta` (`session-bootstrap.ts:91`) and `replayForOf` (`ext-notifications.ts:10`) are three separate defensive merges/reads of the same `_meta.platform` path. A single shared accessor would keep the three sites in agreement as the block grows keys.

packages/agent-runtime/src/modules/acp/services/acp-runtime/session-bootstrap.ts
```diff
@@ -43,42 +52,89 @@ export interface SessionBootstrapDeps {
   openLoadRoute(sessionId: string): number;
   sendToAgent(frame: unknown): void;
   workingDir: string;
+  loadTimeoutMs: number;
   log(msg: string): void;
+  historyProvider?: HistoryProvider;
```
dam-code-guardian, line 58: 🟢 **Suggestion:** `historyProvider` is optional here and on `AcpRuntimeDeps`, and `park` branches on its presence to pick the provider path or the harness load. A required port with a `fetch`-resolves-null no-op supplied by `compose.ts` would let the consumer call it unconditionally. Noting `main` already carries `sessionMetadata?` and `backgroundWork?` the same way, so this follows the file's baseline.

packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts
```diff
@@ -148,12 +160,90 @@ export function createAcpRuntime(deps: AcpRuntimeDeps): AcpRuntime {
...
       deps.log?.(msg);
     },
+    historyProvider: deps.historyProvider,
+    onProviderServed(sessionId) {
+      harnessColdSessions.add(sessionId);
+    },
   });
 
   const idleReapTimers = new Map<string, ReturnType<typeof setTimeout>>();
+  const harnessColdSessions = new Set<string>();
```
dam-code-guardian, line 180: 🟢 **Suggestion:** The rehydrate concern adds four module-level collections here (`harnessColdSessions`, `rehydratingSessions`, `rehydrateTimers`, `heldPrompts`) plus the two functions coordinating them, in a file now at 941 lines — while `harness-lease.ts`, `prompt-scheduler.ts` and `session-bootstrap.ts` each hold one concern in 150–350 lines. The rehydrate state machine is separable on that same seam.

packages/ui/src/modules/sessions/views/chat-view.tsx
```diff
@@ -229,6 +231,42 @@ export function ChatView() {
...
+        pendingPrependRef.current = null;
+        if (el) el.style.overflowAnchor = "";
+        if (outcome === "reloaded") scrollToBottom();
+      }
+      return outcome;
+    },
+    [loadOlderMessages, scrollToBottom],
+  );
+
+  useLayoutEffect(() => {
```
dam-code-guardian, line 257: 🟢 **Suggestion:** The bail-path re-measure fixes the streaming-append case — verified with a React 19 + jsdom repro of both forms (old absolute form `finalTop=800`, new relative form the correct `500`). The residual: the baseline only re-measures on `messages` commits, so growth from a source that does not re-render the list (an embed settling while the page request is in flight) is still folded into the offset. Present before this round too, so not introduced here.

packages/ui/src/modules/sessions/components/chat-message.tsx
```diff
@@ -7,22 +9,86 @@ import { ChatMessagePart } from "./chat-message-part.js";
...
 interface Props {
   message: Message;
   isLast: boolean;
   hasPendingPermission: boolean;
   onRetry: (text: string, attachments?: Attachment[]) => void;
   onFileClick: (path: string) => void;
+  onLoadOlder?: (before: string) => Promise<LoadOlderOutcome>;
+}
+
+function LoadOlderMarker({
```
dam-code-guardian, line 23: 🟢 **Suggestion:** `LoadOlderMarker` keeps pagination fetch state (`startedRef`, `failed`, the `IntersectionObserver`) in the shared message-row file, so this module carries both row rendering and history-paging concerns. Carried from the review at `6506c2b`.

dam-code-guardian (dismissed):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `d90879c`

## PR #3400: feat: fast session opens — tail replay, single connection, history provider
**Author:** jezekra1 | **Branch:** feat/session-open-latency → main | **Changes:** +2443 −284 (29 files)

### Summary
Bounds conversation-open cost: `session/load` replays only the newest tail, older ranges page in through an opaque cursor, the UI serves history over the one live connection, and an image may declare a session-history provider that fills the transcript with no harness process. This round closes both prior criticals at their cause and adds a deadline to the cold-fill path. _(Suppressed 3 findings per PR-local overrides: validateLines all-or-nothing, rehydrate transcript appends, provider cancellation. Suppressed 1 finding per PR context: the entrance animation scoped to the streaming tail is declared in the body.)_

### Changes since last review
Previous HEAD: `f71a949` (2026-08-21T13:16:32Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** `platform/turnEnded` now carries the replay stamp — the ext-notification switch moved into `ext-notifications.ts` and both arrival paths read `replayForOf` (`acp.ts:130`, `acp.ts:140`)
- ✅ **Fixed:** The exec provider counts real bytes and reassembles split multi-byte characters via `StringDecoder` (`history-provider.ts:87`)
- ✅ **Fixed:** The scroll-restore baseline is re-measured on every commit the effect bails on, and the restore became relative (`chat-view.tsx:262`)
- ✅ **Fixed:** A parked bootstrap waiter now has a deadline spanning the provider attempt and the harness fallback (`session-bootstrap.ts:220`)
- 🔁 **Still present:** The runtime-manifest section omits the new `sessionHistory` block (`docs/architecture/runtime-delivery.md:242`)
- 🔁 **Still present:** The UI paragraph still lists short-lived history-replay channels (`docs/architecture/platform-topology.md:75`)
- 🔁 **Still present:** `loadOlderMessages` reports a double failure as `"noop"` (`use-acp-session.ts:166`)
- 🔁 **Still present:** `withClipMeta` emits an ambiguous empty `clipped: {}` (`session-bootstrap.ts:111`)
- 🔁 **Still present:** #3386's dev-deploy "Done when" is unmet; the provider subsystem is named by no linked issue
- 🆕 **New:** One critical the new deadline introduces — below

### Findings
- 🔴 **Critical:** The cold-fill deadline leaves its harness load route open, so that load's replay is later delivered as live updates (`packages/agent-runtime/src/modules/acp/services/acp-runtime/session-bootstrap.ts:191`)
- 🟡 **Warning:** The `agent-lifecycle.md` edit also removes documented facts from subsystems this PR does not change — the terminal frame opcodes and detach-grace rationale, SSH's `PermitUserEnvironment yes`, and the `session/close` mechanics of the terminal-reset path (`docs/architecture/agent-lifecycle.md:163`)
  **Fix:** Doc trimming outside the changed subsystem is declared in the PR body when it rides along with a feature, so a reviewer can tell an intended compression from a lost fact. The page sits at 39,996 of 40,000 chars, so the cap plausibly forced it.
- 🟢 **Suggestion:** `withClipMeta` (`session-bootstrap.ts:91`), `stampReplayFor` (`session-transcript.ts:85`) and `replayForOf` (`ext-notifications.ts:10`) are three separate defensive merges/reads of the same `_meta.platform` path; one shared accessor would keep them in agreement.
- 🟢 **Suggestion:** `historyProvider` is an optional dependency that the consumer branches on for the provider-vs-harness decision. A required port with a null-returning no-op supplied by `compose.ts` would remove the branch — noting it matches two pre-existing optional deps on `main` (`sessionMetadata`, `backgroundWork`), so it follows the file's baseline rather than breaking it (`session-bootstrap.ts:57`).
- 🟢 **Suggestion:** `extractTailFlag`, `extractReplayBefore` and `extractLoadToken` are three near-identical hand-rolled walks down `params._meta.platform`, differing only in the final key, in a package that already uses Zod for this (`acp-runtime.ts:779`).
- 🟢 **Suggestion:** `acp-runtime.ts` is at 941 lines and the rehydrate concern adds four module-level collections plus two coordinating functions, while its neighbours in that directory each live in their own 150–350 line collaborator (`acp-runtime.ts:174`).

- 🟡 **Warning:** `runtime-delivery.md` owns the runtime-manifest concept and states its top-level structure exhaustively — "declares only `drivers` and optional `extensions`". This PR adds a third top-level key, `sessionHistory`, to the schema (`manifest.ts:56`) and to the shipped claude-code manifest (`runtime-manifest.yaml:12`), so that sentence is now false on the page that owns it (`docs/architecture/runtime-delivery.md:242`).
  **Fix:** When a change adds or removes a top-level runtime-manifest key, the manifest section of `runtime-delivery.md` is the page that enumerates them, so its enumeration moves in the same PR.
- 🟡 **Warning:** The `ui` description lists "short-lived ones for history replay" among the ACP channels a tab holds. This PR deletes that channel — `use-acp-history.ts` is removed and the history load now runs on the live per-session connection (`use-acp-connection.ts:256`) (`docs/architecture/platform-topology.md:75`).
  **Fix:** When the UI's ACP connection model changes which channels a tab opens and what each one is for, the `ui` section of `platform-topology.md` is where that model is described, so it moves with the code.
- 🟢 **Suggestion:** `agent-lifecycle.md` lands at 39,996 characters against the 40,000-char cap enforced by `scripts/doc-size.mjs` — four characters of headroom, reached by cutting volatile detail from the unrelated terminal, SSH and mode-switch paragraphs. The guidelines treat the cap as a forcing function for splitting a subsystem rather than a wording budget, which makes the page's boundaries the open question (`docs/architecture/agent-lifecycle.md:119`).

### Issue Fit
Linked issues: #3385 "Opening a long conversation stays slow even when everything is warm" and #3386 "UI - Open a conversation over one connection instead of two" — both inferred, since the body says "Implements #3385 and #3386" with no closing keyword, so neither closes on merge.

- 🟡 **Warning:** #3386's third "Done when" item — "The #3247 benchmark's connect phase is paid once per session open, verified on dev" — is not yet met: the test plan records end-to-end measurement on a local cluster, and the dev-deploy benchmark rerun is still an unchecked box.
  **Fix:** An acceptance criterion that names an environment is only satisfied by evidence from that environment; run the #3247 connect-phase measurement on dev and record the number before the issue is treated as delivered.
- 🟡 **Warning:** The `agent-lifecycle.md` edit deletes documented facts from subsystems this PR does not touch, and neither issue asks for it (mirrored in `### Findings` above) (`docs/architecture/agent-lifecycle.md:163`).
  **Fix:** As above — declare ride-along doc trimming in the body.
- 🟢 **Suggestion:** The session-history provider subsystem — the worker/exec provider, the lazy harness rehydrate with prompt holding, `harness-history-lib.mjs`, the Dockerfile symlink and the `sessionHistory` manifest key — is named by neither #3385 nor #3386. It is declared in the PR body and serves the epic's goal, so this is a carryover from the prior round; an issue of its own under epic #3242 would give it acceptance criteria to be judged against (`packages/agent-runtime/src/modules/acp/infrastructure/history-provider.ts:1`).

Findings merged into `### Findings` above: the `_meta.platform` duplication, the optional `historyProvider` port, the three hand-rolled `_meta` extractors, and the `acp-runtime.ts` size seam.

🔁 2 findings from the previous review still present (see review at `f71a949`): `loadOlderMessages` returning `"noop"` for both an inert click and a double failure (`use-acp-session.ts:166`) — and `LoadOlderMarker` maps every `"noop"` to a retry-able failed state, so a no-session case also renders as an error — and the ambiguous empty `clipped: {}` on the wire (`session-bootstrap.ts:111`).

The round-3 scroll rework is verified correct with a real-React repro (React 19 + jsdom): over a scripted commit sequence — a 300px streaming append at the bottom while the fetch is in flight, then the page arriving — the old absolute form produced `finalTop=800` against the correct `500` from the new relative form. The `pending.height = el.scrollHeight` re-measure on the bail path paired with `scrollTop +=` is a correct pairing and introduces nothing new.

- 🟢 **Suggestion:** The restore effect only re-measures its baseline on `messages` commits, so container growth from a source that does not re-render the list (an image or embed inside an already-rendered row settling while the page request is in flight) is still folded into `el.scrollHeight - pending.height`. This misbehaves identically before this round, so it is not introduced here (`chat-view.tsx:257`).
- 🟢 **Suggestion:** The stick-to-bottom `ResizeObserver` sets `el.scrollTop = el.scrollHeight` whenever `stickRef.current` is true. In a session short enough that the load-older sentinel and the bottom are both within the 30px `nearBottom` threshold, the prepend's resize fires with stick still true and discards the position the restore effect just computed (`chat-view.tsx:285`).
- 🟢 **Suggestion:** `LoadOlderMarker` keeps pagination fetch state (`startedRef`, `failed`, the `IntersectionObserver`) in the shared message-row file, so `chat-message.tsx` carries both row rendering and history-paging concerns (`chat-message.tsx:23`).

### Verdict
REQUEST_CHANGES — the rework closes all four prior blocking findings at their cause, and the new cold-fill deadline introduces one critical: `expireFill` answers its waiters but leaves the outstanding harness `session/load` route open, so that load's replay arrives with no bootstrap state and is fanned out as live updates.

---

tomkis (approved):

tomkis:
Small thing on `history-provider.ts`: almost none of it is about history.

`validateLines` plus appending the session id is maybe 30 lines. The other 245 are spawn, decode, byte cap, timeout, kill, and the worker's respawn and ready handshake. Generic plumbing.

We already hand-roll the same spawn-buffer-timeout-kill block five times: `git.ts:14`, `workspace-command-plugin.ts:75`, `local-skill-repository.ts:621`, `git-protocol-client.ts:104` and `:137`. None of them cap output bytes.

That's why I noticed. The UTF-16 bug from the `f71a949` round was in the generic half, and `workspace-command-plugin.ts:80` still has the same per-chunk decode. One shared `runOnce()` fixes the class, not just this file.

Not blocking. Follow-up is fine.

packages/agent-runtime/src/modules/acp/services/acp-runtime/session-bootstrap.ts
jezekra1: Confirmed, and it was worse than the trace — fixed in 1dbb45b29.

I reproduced all three variants against the runtime harness before touching anything:

1. **The traced scenario: confirmed.** The retry's `session/load` resolves *successfully with an empty conversation* (the collector sees nothing before the hijacked response), and only then does the real history arrive unstamped and get rendered as live. The empty-but-successful load is the part worth naming — the user sees a blank chat, then watches their own history "happen".
2. **The no-retry control: you were right, and my own counter-hypothesis was wrong.** It self-heals: the late answer caches metadata and the next viewer is served correctly from the transcript. That pins the trigger to exactly *two overlapping harness loads for one session* — nothing less.
3. **The rehydrate path has the same defect and is strictly worse.** A provider-served session whose rehydrate times out, where the user just **retries their prompt** — no reopen, no error card — yields `["h1","h2","h1","h2"]`: the whole conversation duplicated live in an open window, for a viewer already engaged. Neither of us had that in scope.

Root cause as I'd state it: the fill state is keyed by session id but the harness's obligation is per request, and `session/update` notifications carry no request correlation — so once two loads overlap, nothing can attribute a frame.

**On the two fixes you offered, I took the first and rejected the second.** Closing the outbound route fixes the traced symptom but not the class: it does nothing for the rehydrate variant (whose leak comes from the cleared suppression marker, not the route), and it still lets the abandoned load's replay land inside a *later* fill and duplicate the log. So the invariant is now **never two outstanding harness loads for one session**:

- a fill that timed out with its load already sent reports that load as **orphaned**; frames for that session are dropped while it is outstanding (they can only be the abandoned replay — no prompt for an unloaded or cold session ever reaches the harness);
- the orphan's late response is dropped by id, so it can never serve a later request's waiters nor cache metadata;
- no second harness load is sent for that session, on either the cold-fill or the rehydrate path.

**Mitigating the cost of holding that invariant**, since "session unusable until the orphan settles" would otherwise be the price:

- a harness that misses a load deadline is treated as **wedged and recycled** — reusing the lease's existing recycle-when-idle/force-after-grace machinery with its own teardown reason. Killing the process makes the abandoned load unanswerable by construction, so the ambiguity is resolved rather than waited out.
- when the recycle is *deferred* because another session is mid-turn, the session stays usable anyway: the provider needs no harness, so a retry refills from it and serves normally, while the abandoned load's frames stay suppressed. Covered by a test that asserts exactly this (busy pod, harness not killed, one harness load ever, late frames suppressed).
- prompts on a session with an orphaned load fail fast with a "harness is restarting" error instead of hanging for the deadline.

Two existing tests changed sides: both asserted that a retry sends a *second* load, which is precisely the corruption vector, so they now assert it does not. The reviewer's original repro can no longer even be written — it fails at its own attempt to grab a second load id, because there isn't one. Runtime 240/240 on this PR, 245/245 with #3415 rebased on top.

PR #3412, opened 2026-08-21
Public Agent Page

### LINKED ISSUE #3135: UI - Opening a sandbox you don't own says "Sandbox not found"
## Context

Agents post links to their own sandbox page into Slack. Anyone in the channel who isn't the sandbox's owner clicks that link and lands on a near-empty page reading "Sandbox not found." Reported from a shared Slack channel where an agent had posted its own sandbox link.

## Problem

The sandbox exists. The message says it doesn't.

So the user's first read is "broken link" or "wrong URL" — they have no idea they've hit an access boundary, and nothing tells them whose sandbox it is or who to ask. Every non-owner who follows a shared sandbox link hits this, which is the common case for links shared in a channel.

It also makes the product look broken at a moment when nothing is actually wrong.

## Expected Behavior

A page that says the sandbox exists but isn't yours to view, and points you at its owner.

## Actual Behavior

An otherwise blank page reading "Sandbox not found." — the same message shown when a sandbox genuinely doesn't exist.

## Steps to Reproduce

1. Get a link to a sandbox you don't own (for example, a sandbox link an agent posted in Slack).
2. Open it.

tomkis:
People in a shared Slack channel click the link under an agent's message and land on a dead end telling them the thing does not exist. It does exist. They have hit an access boundary, and nothing on the page says so or tells them what the agent is.

This replaces that with a **Public Agent Page**: an unauthenticated page that names the agent, names its owner, explains what the platform is, and invites the reader to create an agent of their own. For most people in a shared channel it is the first surface they will ever see, so it is built as a conversion surface rather than an error page.

The page cannot tell an owner from a stranger and does not try, so it has two states: named, for an agent that is actually bound to a channel, and a generic page for everything else. Unknown, unbound, and deleted agents all render identically, so the URL never confirms which agents exist and a stale link still lands on something useful.

Two supporting changes ride along. The Slack footer is decoupled from history attribution first, because it turns out the link text is parsed back out of Slack to work out which agent wrote which line, so changing the wording would silently break it. And the chat route now falls back to the public page, which is the part that actually closes the reported bug for signed-in users.

Closes #3135

## Sub-issues

- [x] 01 — Public agent projection and read service
- [x] 02 — Public agent page HTTP surface
- [x] 03 — Decouple the Slack agent footer from attribution
- [x] 04 — Chat route falls back to the public page
- [x] 05 — Architecture documentation


packages/ui/src/modules/agents/views/public-agent-view.tsx
```diff
@@ -0,0 +1,181 @@
...
+      <span
+        aria-hidden="true"
+        className="flex size-6 shrink-0 items-center justify-center rounded-full bg-muted text-[11px] font-semibold text-muted-foreground"
+      >
+        {ownerInitials(email)}
+      </span>
+      <span>
+        Created by{" "}
+        <span className="font-semibold break-words text-foreground">
+          {email}
```
kapetr: This renders the owner's raw email to unauthenticated visitors — `resolveBySub` returns the Keycloak `email` field, nothing else, so "Created by" is always a real mailbox on a link pasted into Slack channels. The signed-off design shows a display name ("Created by Radek Ježek"). Keycloak's same endpoint has `firstName`/`lastName`; use those, or drop the byline until a display name exists.

packages/api-server/src/modules/channels/infrastructure/agent-footer.ts
```diff
@@ -21,20 +23,16 @@ function escapeLinkLabel(name: string): string {
     .trim();
 }
 
-function unescapeLinkLabel(label: string): string {
-  return label
-    .replace(/&lt;/g, "<")
-    .replace(/&gt;/g, ">")
-    .replace(/&amp;/g, "&")
-    .trim();
+export function agentFooterLabel(brand: { name: string }): string {
```
kapetr: Every footer is now the same fixed brand line, and nothing else in a Slack post names the agent (one install-wide bot, no `username` on `postMessage`). In a channel with two agents a reader can't tell who answered. The last comment on #3135 objects to exactly this and proposes "\<agent name\> - Powered by DAM", with no reply — worth settling there before this merges.

packages/ui/src/modules/sandboxes/views/sandbox-home-view.tsx
```diff
@@ -18,9 +19,12 @@ import { useSectionSummaries } from "../hooks/use-section-summaries.js";
 
 export function SandboxHomeView() {
   const f = useSandboxSettingsForm();
+  const agentId = useStore((s) => s.agentId);
   const section = useStore((s) => s.sandboxSection);
   const navigateToSandboxHome = useStore((s) => s.navigateToSandboxHome);
 
+  usePublicAgentFallback(agentId, f.status === "not-found");
```
kapetr: This redirect fires on absence from the user's own agent list, not on `NOT_FOUND`/`FORBIDDEN` like the chat route — and it's a full `window.location.replace` out of the SPA. An approval row for a deleted agent (nothing removes approvals on `AgentDeleted`) navigates here, so the owner gets a hard reload onto the anonymous page. Key it off the error code, and restore a fallback for the no-redirect case — the old "Agent not found." line is gone, so the panel otherwise renders empty.

packages/api-server/src/modules/agents/services/public-agent-page-service.ts
```diff
@@ -0,0 +1,59 @@
...
+  resolveOwnerEmail: (ownerSub: string) => Promise<string | null>;
+}
+
+export function createPublicAgentPageService(
+  deps: PublicAgentPageDeps,
+): PublicAgentPageService {
+  async function fillProfile(
+    agentId: string,
+  ): Promise<PublicAgentProfileRow | null> {
+    const agent = await deps.readAgent(agentId);
```
dam-code-guardian: 🔴 **Critical:** Lazy fill reaches the K8s API on **every** public view of a bound agent whose CR is gone, so this unauthenticated endpoint drives unbounded control-plane reads.

`markProfileDeleted` updates only rows `WHERE deleted_at IS NULL`, so with no row present it writes no tombstone — the next view misses `getProfile` again and calls `readAgent` again. Reachable in normal operation: `channel-cleanup` swallows its own delete failure, leaving a `channels` row that still passes `hasAnyBinding` after the CR is deleted.

Measured on a transcription of this service: 1000 views produced 1000 control-plane reads, against 1 for a healthy bound agent.

**Fix:** Make the negative result durable — have `markProfileDeleted` upsert a tombstone row rather than update only live rows, and have `get` treat a tombstoned row as "no agent" without calling `readAgent`. Any lazy-fill miss must cost at most one control-plane read per agent, never one per request.

packages/api-server/src/modules/agents/services/public-agent-page-service.ts
```diff
@@ -0,0 +1,59 @@
...
+  markProfileDeleted: (agentId: string) => Promise<void>;
+  readAgent: (agentId: string) => Promise<PublicAgentIdentity | null>;
+  resolveOwnerEmail: (ownerSub: string) => Promise<string | null>;
+}
+
+export function createPublicAgentPageService(
+  deps: PublicAgentPageDeps,
+): PublicAgentPageService {
+  async function fillProfile(
+    agentId: string,
```
dam-code-guardian, line 39: 🔴 **Critical:** A K8s read failure here propagates out of the route as a 500. `readAgent`'s throw is uncaught in `fillProfile` and `get` — only `resolveOwnerEmail` has a `try`/`catch` — so a control-plane hiccup turns the page into an error response.

It also breaks the documented no-oracle property: a bound-but-unfilled id answers 500 while every other id answers 200, which is a body-independent signal about which ids exist.

**Fix:** Treat a failed identity read exactly like a failed owner-email read — catch it, log, answer the generic page at 200. Every path through this route must return 200, so the status code never varies with which agents exist.

packages/ui/src/modules/sandboxes/views/sandbox-home-view.tsx
dam-code-guardian: 🟡 **Warning:** This gates the fallback on `f.status === "not-found"`, which is `agentsQuery.data !== undefined && !agent` — list absence, not the denied read that `ChatView` uses. Deleting your own agent from its sandbox page invalidates the list, so `not-found` can go true and send the owner to the public marketing page instead of the list that `onSuccess: () => setView("list")` intended.

**Fix:** Gate on a denied-read signal (the `agents.get` error, as `ChatView` does) rather than list membership, and suppress it while a delete for that agent is in flight.

packages/ui/src/modules/platform/lib/routes.ts
```diff
@@ -43,6 +43,17 @@ export type Route =
 
 export type View = Route["view"];
 
+const publicAgentRe = /^\/a\/([^/]+)\/?$/;
+
+export function parsePublicAgentPath(pathname: string): string | null {
+  const match = pathname.match(publicAgentRe);
+  return match ? decodeURIComponent(match[1]!) : null;
```
dam-code-guardian: 🟡 **Warning:** `decodeURIComponent` throws `URIError` on a malformed escape — `/a/%`, `/a/100%`, `/a/agent-%zz` all throw. `main()` runs this before any error handling and is invoked with no `.catch`, so the bootstrap promise rejects and the visitor gets a blank document. Reproduced in `node`.

**Fix:** Decode defensively in the matcher and return the raw capture (or `null`) on `URIError`, so no pathname can throw out of route parsing.

packages/api-server/src/modules/channels/infrastructure/agent-footer.ts
```diff
@@ -21,20 +23,16 @@ function escapeLinkLabel(name: string): string {
...
+  return `Powered by ${brand.name}`;
```
dam-code-guardian: 🟡 **Warning:** The Slack turn contract still tells each agent that a post is its own "only if its footer reads `<agentName>`" (`slack-turn-copy.ts:118`, outside this diff). Every footer now reads `Powered by <brand>`, so agents are instructed to disown their own posts.

**Fix:** Key the self-attribution copy to the footer's agent id, or to the injected-history `you (this agent)` prefix, so a label reword cannot change attribution semantics.

Separately, on issue fit: the latest comment on #3135 — yours, 2026-08-24 — pushes back on dropping the agent name and proposes `<agent name> - Powered by <brand>`. This commit predates it by 20 minutes, and the PR carries a closing keyword.

packages/api-server/src/modules/agents/sagas/persist-public-agent-profile.ts
```diff
@@ -0,0 +1,76 @@
...
+      events$()
+        .pipe(
+          ofType<T>(type),
+          mergeMap((event) => handle(event.agentId), STREAM_CONCURRENCY),
+        )
+        .subscribe(),
+    );
+  }
+
+  onEvent<AgentCreated>(EventType.AgentCreated, refresh);
```
dam-code-guardian: 🟡 **Warning:** The saga upserts a row on `AgentCreated`/`AgentUpdated` with no binding check, so the projection grows a row per agent in the install rather than per bound agent. The hourly reconcile then walks every row and issues one control-plane read each — the fleet-wide walk the architecture page names as this feature's failure mode.

**Fix:** Gate row creation on `hasAnyBinding` in `refresh`, or restrict `listProfileIdsForReconcile` to still-bound agents, so both the projection and the reconcile stay bounded by bound agents.

packages/ui/src/modules/agents/views/public-agent-view.tsx
```diff
@@ -0,0 +1,181 @@
...
+    <main className="flex min-h-dvh flex-col bg-background text-foreground">
+      <Masthead brand={brand} />
+      {agent ? (
+        <>
+          <div className="flex-1 px-6 pt-24">
+            <div className="mx-auto w-full max-w-[640px]">
+              <h1 className="text-2xl leading-tight font-semibold text-balance break-words">
+                {agent.name}
+              </h1>
+              {agent.ownerEmail === null ? null : (
```
dam-code-guardian: 🟡 **Warning:** The owner's full email renders verbatim to anonymous visitors on an un-rate-limited page. The architecture page prices the disclosure of the agent *name* via the binding carefully, but never the owner's address, which is a different and more harvestable identifier.

**Fix:** Return a display name or masked handle for the public byline and keep the raw address behind an authenticated read. If the address is deliberate, record that decision in the page's Nobody-is-identified section, since that is where a reader will look for it.

packages/ui/src/modules/sessions/views/chat-view.tsx
```diff
@@ -94,6 +94,7 @@ export function ChatView() {
   const agents = agentsData?.list ?? [];
   const agentOperable = useIsAgentOperable(selectedAgent);
   const agentInaccessible = useIsAgentInaccessible(selectedAgent);
+  usePublicAgentFallback(selectedAgent, agentInaccessible);
```
dam-code-guardian: 🟡 **Warning:** `window.location.replace` in an effect only starts a navigation. Until it completes, `ChatView` renders the full chat surface — header, composer, terminal — for an agent the visitor cannot read, where the removed overlay used to block it.

**Fix:** Keep a blocking placeholder for `agentInaccessible` so an unreadable agent never presents an interactive surface.

packages/ui/src/public-agent-page.tsx
```diff
@@ -0,0 +1,46 @@
...
+ * UNIT_BOUNDARY_DESCRIPTION: The entry point for the one route that renders
+ * without a signed-in user. It is a separate entry from main.tsx, not a view
+ * inside App, because App is reached only after initAuth has either produced a
+ * user or redirected to Keycloak. Nothing here touches auth, the terms gate, the
+ * tRPC client, or the route store, so an anonymous visitor cannot fall into an
+ * authenticated tree.
+ */
+export async function renderPublicAgentPage(agentId: string): Promise<void> {
+  applySystemTheme();
+  const agent = await fetchPublicAgent(agentId);
```
dam-code-guardian: 🟡 **Warning:** This awaits the lookup before the first `createRoot(...).render`, and `#root` in `index.html` is empty, so a slow or failing public read leaves an anonymous visitor arriving from Slack on a blank white page.

**Fix:** Render the masthead shell with a loading state immediately, then update when the fetch resolves.

packages/ui/src/modules/agents/views/public-agent-view.tsx
```diff
@@ -0,0 +1,181 @@
...
+    </>
+  );
+}
+
+function pitch(brand: Brand, subject: "this-agent" | "platform"): string {
+  const lead =
+    subject === "this-agent"
+      ? `This agent was created with ${brand.name}, a platform`
+      : `${brand.name} is a platform`;
+  return `${lead} for running agents securely in the cloud, running experiments, and creating knowledge bases.`;
```
dam-code-guardian: 🟡 **Warning:** The agreed copy on #3135 reads "an IBM Research platform"; this renders only "a platform", so the reader never learns the platform is open to them — requirement 6's entitlement framing.

**Fix:** Add the entitlement sentence, sourced from brand config so the copy stays deployment-neutral.

packages/ui/src/modules/agents/views/public-agent-view.tsx
```diff
@@ -0,0 +1,181 @@
...
+  brand: Brand;
+  openPath: string;
+}) {
+  return (
+    <div className="mt-6 flex flex-wrap items-center gap-2">
+      <Button asChild>
+        <a href="/">Create your own agent</a>
+      </Button>
+      <span className="ml-1 text-sm">
+        Already have an account? <a href={openPath}>Open in {brand.name}</a>
```
dam-code-guardian: 🟢 **Suggestion:** This `<a>` carries no classes, and the global stylesheet scopes link colour/underline to `.prose a`, so the secondary CTA renders as plain body text on the page whose job is conversion. The codebase convention is `font-medium text-accent hover:underline` (`release-notes-link.tsx`, `oauth-app-hint.tsx`).

packages/ui/src/modules/agents/lib/owner-initials.ts
```diff
@@ -0,0 +1,8 @@
+export function ownerInitials(email: string): string {
+  const localPart = email.split("@")[0] ?? "";
+  const words = localPart.split(/[._\-+]+/).filter(Boolean);
+  const first = [...(words[0] ?? "")][0] ?? "";
+  const last =
+    words.length > 1 ? ([...(words[words.length - 1] ?? "")][0] ?? "") : "";
+  return (first + last).toUpperCase();
```
dam-code-guardian, line 6: 🟢 **Suggestion:** Returns an empty string when the local part holds no alphanumerics (`@example.com`), and `OwnerByline` renders the circle unconditionally, leaving a blank circle beside the address. Fall back to the first character, or omit the circle when the initials are empty.

packages/api-server/src/modules/agents/sagas/persist-public-agent-profile.ts
```diff
@@ -0,0 +1,76 @@
...
+        await deps.markProfileDeleted(agentId);
+        return;
+      }
+      await deps.upsertProfile({
+        agentId,
+        name: agent.name,
+        ownerSub: agent.ownerSub,
+      });
+    } catch (err) {
+      process.stderr.write(
```
dam-code-guardian: 🟢 **Suggestion:** `composePublicAgentPage` already threads a `log` dependency into the reconcile service, so routing the saga's failures through `process.stderr.write` sends half this feature's errors around the configured logger.

packages/ui/src/modules/agents/views/public-agent-view.tsx
```diff
@@ -0,0 +1,181 @@
...
+          ) : null}
+        </div>
+        <div className="mt-0.5 text-sm">{children}</div>
+      </div>
+    </div>
+  );
+}
+
+function SlackHint({ brand }: { brand: Brand }) {
+  const botMention = `@${brand.name}`;
```
dam-code-guardian: 🟢 **Suggestion:** The instructions and the sample transcript both build the handle as `@${brand.name}`, but that is a display string — the Slack mention resolves against the installed app's bot user, which need not match, so a visitor may be told to type a handle that does not exist. The server already tracks `botUserId`.

dam-code-guardian (dismissed):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `7ccd11f5`

## PR #3412: Public Agent Page
**Author:** tomkis | **Branch:** feat/3135-public-agent-page → main | **Changes:** +4432 −193 (46 files)

### Summary
Adds an unauthenticated Public Agent Page reached from the Slack Agent Footer, served from a new Postgres projection, plus the footer/attribution decoupling and a chat-route fallback. The design is sound and the architecture page is unusually good; the findings sit in the projection's read path and in the fallback's trigger.

### Findings
- 🔴 **Critical:** Lazy fill reaches the K8s API on **every** public view of a bound agent whose CR is gone, so the unauthenticated endpoint drives unbounded control-plane reads — exactly what the projection exists to prevent. `markProfileDeleted` only updates rows `WHERE deleted_at IS NULL`, so when no row exists it writes no tombstone; the next view misses `getProfile` again and calls `readAgent` again. Reachable in normal operation: `channel-cleanup` swallows its own delete failure, leaving a `channels` row that passes `hasAnyBinding` after the CR is deleted. Measured on a transcription of the service: 1000 views → 1000 control-plane reads, against 1 for a healthy bound agent. (`packages/api-server/src/modules/agents/services/public-agent-page-service.ts:28`, `packages/api-server/src/modules/agents/infrastructure/public-agent-profile-repository.ts:55`)
  **Fix:** Make the negative result durable — have `markProfileDeleted` upsert a tombstone row rather than update only live rows, and have `get` treat a tombstoned row as "no agent" without calling `readAgent`. Any lazy-fill miss must cost at most one control-plane read per agent, not one per request.
- 🔴 **Critical:** A K8s read failure inside lazy fill propagates out of the route as a 500. `readAgent`'s throw is uncaught in `fillProfile` and `get` (only `resolveOwnerEmail` has a `try`/`catch`), so a control-plane hiccup turns the page into an error response. That also breaks the documented "no oracle" property: a bound-but-unfilled id answers 500 while every other id answers 200. (`packages/api-server/src/modules/agents/services/public-agent-page-service.ts:26`, `packages/api-server/src/modules/agents/infrastructure/public-agent-routes.ts:26`)
  **Fix:** Treat a failed identity read like a failed owner-email read — catch it, log, and answer the generic page at 200. Every path through this route must return 200, so the status code never varies with which ids exist.
- 🟡 **Warning:** The fallback in the sandbox route triggers on `f.status === "not-found"`, which means "absent from your agent list", not "denied read" as in `ChatView`. Deleting your own agent from its sandbox page invalidates the list, so the owner can be redirected to the public marketing page instead of the agent list `onSuccess: () => setView("list")` intended. Reported by react-ui-engineering. (`packages/ui/src/modules/sandboxes/views/sandbox-home-view.tsx:26`)
  **Fix:** Gate the fallback on a denied-read signal (the `agents.get` error, as `ChatView` does), never on list membership, and suppress it while a delete for that agent is in flight.
- 🟡 **Warning:** `parsePublicAgentPath` calls `decodeURIComponent` on the captured segment, which throws `URIError` on a malformed escape (`/a/%`, `/a/100%`). `main()` is invoked with no `.catch`, so the bootstrap promise rejects and the visitor gets a blank document — confirmed in `node`. Reported by typescript-engineering. (`packages/ui/src/modules/platform/lib/routes.ts:50`)
  **Fix:** Decode defensively in the matcher and return the raw capture (or `null`) on `URIError`, so no pathname can throw out of route parsing.
- 🟡 **Warning:** The Slack turn contract still tells each agent a post is its own "only if its footer reads `<agentName>`", but every footer now reads `Powered by <brand>`, so agents are instructed to disown their own posts. The copy file is outside this diff and was not updated with the label change. Reported by typescript-engineering. (`packages/api-server/src/modules/channels/infrastructure/agent-footer.ts:27`, consumed at `packages/api-server/src/modules/channels/infrastructure/slack-turn-copy.ts:118`)
  **Fix:** Key the self-attribution copy to the footer's agent id or the injected-history `you (this agent)` prefix, not the label text, so a label reword cannot change attribution semantics.
- 🟡 **Warning:** The saga upserts a profile row on `AgentCreated`/`AgentUpdated` with no binding check, so the table grows a row per agent in the install rather than per bound agent, and the hourly reconcile then issues one control-plane read per row — the fleet-wide walk the architecture page names as the failure mode. Reported by typescript-engineering. (`packages/api-server/src/modules/agents/sagas/persist-public-agent-profile.ts:62`, `packages/api-server/src/modules/agents/services/public-agent-profile-reconcile-service.ts:41`)
  **Fix:** Gate row creation on `hasAnyBinding` in the saga's `refresh`, or restrict `listProfileIdsForReconcile` to still-bound agents, so both the projection and the reconcile stay bounded by bound agents.
- 🟡 **Warning:** The owner's full email renders verbatim to anonymous visitors on an un-rate-limited page. The architecture page reasons carefully about disclosing the agent *name* via the binding, but never prices the owner's address, which is a different and more harvestable identifier. Reported by react-ui-engineering. (`packages/ui/src/modules/agents/views/public-agent-view.tsx:148`)
  **Fix:** Return a display name or masked handle for the public byline and keep the raw address behind an authenticated read; if the address is deliberate, record that decision in the page's Nobody-is-identified section.
- 🟡 **Warning:** `ChatView` renders the full chat surface — header, composer, terminal — while the fallback navigation is in flight, where the removed overlay previously blocked it. `window.location.replace` in an effect only starts the navigation. Reported by react-ui-engineering. (`packages/ui/src/modules/sessions/views/chat-view.tsx:97`)
  **Fix:** Keep a blocking placeholder for `agentInaccessible` so an unreadable agent never presents an interactive surface.
- 🟡 **Warning:** `renderPublicAgentPage` awaits the lookup before the first render and `#root` is empty, so a slow public read shows an anonymous visitor a blank white page. Reported by react-ui-engineering. (`packages/ui/src/public-agent-page.tsx:31`)
  **Fix:** Render the masthead shell immediately, then update it when the fetch resolves.
- 🟢 **Suggestion:** The secondary CTA `Open in {brand.name}` is an unstyled `<a>`; the global stylesheet scopes link colour to `.prose a`, so it reads as body text on the page whose job is conversion (`packages/ui/src/modules/agents/views/public-agent-view.tsx:127`). Apply the codebase's link convention (`font-medium text-accent hover:underline`).
- 🟢 **Suggestion:** `<main>` wraps the masthead, so the brand header is announced as main content and the page exposes no banner landmark (`packages/ui/src/modules/agents/views/public-agent-view.tsx:139`). Move `Masthead` into a sibling `<header>`.
- ✅ **Looks good:** The footer regex widening is safe. A differential over 40 cases (old `/chat/` and legacy `/sandboxes/` forms, escaped labels, path- and query-style sessions) parsed every historical footer correctly under the new pattern, so existing channel history keeps its attribution.

- 🟡 **Warning:** The ui section still says a non-owner following a chat link gets "its own screen naming the reason", but this PR deletes that screen (`agent-inaccessible-overlay.tsx`) and replaces it with a full-page redirect via `usePublicAgentFallback` (`docs/architecture/platform-topology.md:77`)
  **Fix:** Rewrite that sentence to describe the new behaviour — an unreadable Agent now leaves the authenticated app and lands on the Public Agent Page — and cross-link `architecture/public-agent-page.md`.
- 🟡 **Warning:** The page says an Agent is named when it holds "at least one Channel Binding", but the gate is `hasAnyBinding`, which queries only the `channels` table — Slack-only, since Telegram bindings live in `telegram_conversations`. A Telegram-only Agent renders the generic page despite holding a Channel Binding in the documented sense (`docs/architecture/public-agent-page.md:29`, `docs/architecture/public-agent-page.md:34`)
  **Fix:** Say the disclosure rides on a *Slack* Channel Binding specifically, in both the two-states list and the "What the page exposes" rule.

### Issue Fit
Linked issue: #3135 (closing keyword)

- 🟡 **Warning:** The footer label is now only `Powered by <brand>`, dropping the agent name. The issue's latest comment — yours, 2026-08-24 — pushes back on exactly this and proposes `<agent name> - Powered by <brand>`; the label commit predates it by 20 minutes. (`packages/api-server/src/modules/channels/infrastructure/agent-footer.ts:27`)
  **Fix:** Carry the agent name in the label as that comment proposes, or resolve the point with the issue owners before the closing keyword closes #3135.
- 🟡 **Warning:** An owner clicking the footer link no longer reaches their session; they get the public page plus a manual "Open in <brand>" step. The issue asks for owner pass-through, and it is still an open question in the two most recent design comments. (`packages/api-server/src/modules/channels/infrastructure/agent-footer.ts:34`)
  **Fix:** Restore owner pass-through — let the authenticated route handle the click and fall back to the public page only on a denied read — or get the drop accepted before #3135 closes.
- 🟡 **Warning:** The pitch omits the entitlement framing the issue agreed on ("an IBM Research platform"); the code renders only "a platform", so the reader never learns the platform is open to them. (`packages/ui/src/modules/agents/views/public-agent-view.tsx:111`)
  **Fix:** Add the entitlement sentence, sourced from brand config so the copy stays deployment-neutral.

- 🔁 Three findings are carried in `### Findings` above (the `decodeURIComponent` throw, the unbounded saga/reconcile projection, the stale turn-contract copy).
- 🟢 **Suggestion:** `ownerInitials` returns an empty string when the local part holds no alphanumerics (`@example.com`), and `OwnerByline` renders the circle unconditionally, leaving a blank circle beside the address (`packages/ui/src/modules/agents/lib/owner-initials.ts:7`). Fall back to the first character, or omit the circle.
- 🟢 **Suggestion:** The saga reports failures with `process.stderr.write` while `composePublicAgentPage` already threads a `log` dependency to the reconcile service, so half this feature's errors bypass the configured logger (`packages/api-server/src/modules/agents/sagas/persist-public-agent-profile.ts:42`).
- 🟢 **Suggestion:** `PublicAgentProfileRow` lives in infrastructure and is imported by both services and the saga, so application types are shaped by the persistence row (`packages/api-server/src/modules/agents/services/public-agent-page-service.ts:2`). Let the service own its type and map in the repository.

- 🔁 Five findings are carried in `### Findings` above (the sandbox `not-found` trigger, the owner email exposure, the unblocked chat surface, the blank first paint, and the two presentation suggestions).
- 🟢 **Suggestion:** The Slack instructions and sample transcript build the bot handle as `@${brand.name}`, but that is a display string — the mention resolves against the installed app's bot user, which need not match, so visitors may be told to type a handle that does not exist (`packages/ui/src/modules/agents/views/public-agent-view.tsx:76`). Derive it from the tracked `botUserId`.

### Verdict
REQUEST_CHANGES — the unauthenticated read path can drive unbounded control-plane reads and can answer 500, which are the two properties the design set out to guarantee.

---

packages/api-server/src/modules/agents/sagas/persist-public-agent-profile.ts
```diff
@@ -0,0 +1,76 @@
...
+          mergeMap((event) => handle(event.agentId), STREAM_CONCURRENCY),
+        )
+        .subscribe(),
+    );
+  }
+
+  onEvent<AgentCreated>(EventType.AgentCreated, refresh);
+  onEvent<AgentUpdated>(EventType.AgentUpdated, refresh);
+  onEvent<SlackConnected>(EventType.SlackConnected, refresh);
+  onEvent<AgentDeleted>(EventType.AgentDeleted, async (agentId) => {
```
kapetr, line 87: Every deletion now inserts a tombstone here, even for agents that never had a binding — those rows are unreachable (`hasAnyBinding` short-circuits first) and nothing purges them, so the table grows with fleet churn (sweep, cascade, reaper) and the hourly reconcile scans it on unindexed `deleted_at`. Gate the tombstone on a binding, same as the read path.

kapetr (approved):

PR #3415, opened 2026-08-21
feat(acp): retain warm transcripts after the idle reap

jezekra1:
Stacked on #3400. Follow-up from its review discussion: the idle reap conflated two lifetimes — the harness's per-session subprocess (expensive, must be freed fast) and the runtime's in-memory transcript (2 MB capped, cheap to keep). This splits them.

## What changes

- The reap still fires on the short idle timer and still sends `session/close`, freeing the harness subprocess — but instead of forgetting the transcript it marks the session **harness-cold** and retains the log for an hour (`transcriptRetainMs`). That is exactly the state provider-served sessions already live in, so no new machinery: opens are served from memory (~6 ms instead of a cold load), and the first prompt rehydrates the harness through the existing held-prompt path.
- Retained transcripts share a **64 MB budget** (`transcriptRetainBytesCap`). Tipping over it evicts the oldest retained transcript, never the newest — a busy sandbox can't accumulate unbounded warm history.
- Re-engaging a session (viewer returns, trigger fires) cancels the pending forget. An explicit session reset still forgets immediately, without a second `session/close` for a session the harness already released.

## Why

Re-opening a recently viewed conversation was the dominant remaining cost in session switching (#3247): the reap made every open after ~3 s of absence a cold one. Retention makes any session touched in the last hour open warm — subsuming the 'prefetch recent conversations' idea, since the transcript is already loaded.

Covered by a new suite (`acp-runtime-transcript-retention.test.ts`): warm re-open after reap + rehydrate-on-prompt, window expiry, re-engage cancels the forget, LRU eviction over budget, reset clears retention.


jezekra1:
actually it's overengineering for now

jezekra1:
Superseded by #3437 — same commit rebased onto `main`. GitHub auto-closed this one when #3400 merged and its base branch was deleted; a closed PR can't be retargeted, hence the new number.

PR #3417, opened 2026-08-21
perf(platform-base): seed home in the image boot instead of an init container

jezekra1:
Part of #3242 (from the #3246 start-latency measurements).

## Problem

The seed-home `init` container costs a container create/start in the kata VM on every boot (#3246 measured the create-path init step at ~1s, and wake-path np-gate+init at ~3s combined) — for a job that is a ~100ms no-clobber copy paid once per volume. The copy itself cannot move to image build time: the home volume is provisioned per agent and mounts empty, shadowing image content.

## Change

- The **container entrypoint** now runs the exact `cp -rn` + sentinel + `mkdir work` sequence the init container ran — same coreutils semantics, no reimplementation — honoring `$HOME` (which the controller sets from `agentHome`, so per-template home overrides keep working exactly as they did with the init container).
- The **VM boot** replays the same static step in cloud-init, next to the existing cache-symlink step.
- **agent-runtime** ensures the work directory at boot, covering images that don't run the platform entrypoint.
- The **controller** no longer renders the `init` container. `spec.init` and a Template's `init` are retained but no longer read (the `spec.env` precedent), and a still-set `controller.agent.templateDefaults.init` logs a deprecation warning at startup instead of failing the strict config decode. The chart's default `templateDefaults.init` is removed.
- Docs: seeding contract in `persistence.md`; `skills.md`'s pristine-root wording updated.

## Review response (Code Guardian round 1)

The first revision reimplemented the copy in TypeScript (`fs.cpSync`); the review found three real criticals there (strict-decode crash-loop, `cpSync` symlink rewriting, an uncatchable `std::filesystem` abort — both reproduced locally). Rather than hand-rolling `cp -rn` semantics in JS, this revision deletes that code and keeps coreutils `cp -rn` itself, which also resolves the `HOME_DIR` finding (the entrypoint honors `$HOME`). The runtime-wide "`HOME_DIR` is never set" gap the review surfaced predates this PR and will be filed separately.

## Testing

- Docker, on the registry base image with the new entrypoint overlaid: empty home volume seeds (files + sentinel + `work/`), a user-edited file survives, re-run is a silent no-op; `HOME` override honored.
- controller suite green, including: no init containers rendered; a set `spec.init` ignored on both backends; the static seed present in VM userdata.
- agent-runtime suite 208/208; helm render green.
- Measured before/after boot timing on a local cluster to follow as a PR comment once this revision is deployed there (baseline per #3246: init step ~1s create / np-gate+init ~3s wake).

## Compatibility

Seeding responsibility moves from the controller (init container) to the agent image (boot), and the two upgrade independently:

| Controller | Agent image | First boot of an empty volume | Already-seeded volumes |
|---|---|---|---|
| old | new | both seed — benign (both no-clobber + sentinel) | no-op |
| new | new | image boot seeds | no-op |
| new | pre-change | **nobody seeds** | no-op (sentinel present) |

Existing agents are unaffected (their volumes carry the sentinel), and post-upgrade creates get post-change images via their template. The exposed slice: agents created but never woken before the upgrade, custom images built on a pre-change `platform-base`, installs pinning agent images independently of the chart. On such an agent the home never seeds (agent-runtime still ensures `work/`), until the agent's image is advanced or the agent is recreated. Suggest a release-note line rather than a compatibility shim.

### DIFF
--- a/packages/agent-runtime/src/server.ts
+++ b/packages/agent-runtime/src/server.ts
@@ -1,6 +1,6 @@
 import http from "node:http";
 import { monitorEventLoopDelay } from "node:perf_hooks";
-import { existsSync, readFileSync } from "node:fs";
+import { existsSync, mkdirSync, readFileSync } from "node:fs";
 import { fileURLToPath } from "node:url";
 import { dirname, join } from "node:path";
 import headlessPkg from "@xterm/headless";
@@ -57,6 +57,14 @@ const homeDir = config.PLATFORM_DEV
 const workDir = config.PLATFORM_DEV
   ? join(__dir, "../working-dir")
   : config.WORK_DIR;
+
+try {
+  mkdirSync(workDir, { recursive: true });
+} catch (err) {
+  process.stderr.write(
+    `[server] could not create workDir ${workDir}: ${(err as Error).message}\n`,
+  );
+}
 
 function skillRefPaths(manifest: RuntimeManifest, home: string): string[] {
   const binding = resolveDrivers(manifest)["skill-ref"] as
--- a/packages/controller/pkg/config/config.go
+++ b/packages/controller/pkg/config/config.go
@@ -222,6 +222,9 @@ func LoadFromEnv() (*Config, error) {
 	if cfg.AgentBase.AccessMode != "" {
 		slog.Warn("controller.agent.base.accessMode is deprecated and ignored — workspace volumes are always ReadWriteOnce (#2988); remove it from your values")
 	}
+	if cfg.AgentTemplateDefaults.Init != "" {
+		slog.Warn("controller.agent.templateDefaults.init is deprecated and ignored — first-boot home seeding moved into the agent image's boot (#3242); remove it from your values")
+	}
 	if v := os.Getenv("STORAGE_MIGRATION"); v != "" {
 		dec := json.NewDecoder(strings.NewReader(v))
 		dec.DisallowUnknownFields()
--- a/deploy/helm/platform/templates/agent-templates.yaml
+++ b/deploy/helm/platform/templates/agent-templates.yaml
@@ -114,10 +114,6 @@ data:
     skillSources:
       {{- toYaml . | nindent 6 }}
     {{- end }}
-    {{- with $tmpl.init }}
-    init: |
-      {{- . | nindent 6 }}
-    {{- end }}
   {{- end }}
   {{- end }}
   {{- end }}
--- a/deploy/helm/platform/values.yaml
+++ b/deploy/helm/platform/values.yaml
@@ -979,16 +979,6 @@ controller:
       # set unconditionally by the controller.
       env: []
       skillSources: []
-      # Shell-expanded at runtime — HOME is set on the init container.
-      # Templates that need extras (e.g. pip install) REPLACE this entirely.
-      init: |
-        #!/bin/bash
-        # Seed home from image on first boot
-        if [ ! -f $HOME/.initialized ]; then
-          cp -rn /app/working-dir/. $HOME/ 2>/dev/null || true
-          touch $HOME/.initialized
-        fi
-        mkdir -p $HOME/work
 
     # -- Size the controller materializes into any Agent spec that lacks a
     # limits dimension (#1900): fill-if-absent on reconcile, never touching
--- a/docs/architecture/persistence.md
+++ b/docs/architecture/persistence.md
@@ -104,6 +104,8 @@ Two domain resources are deliberately not CRDs:
 
 Each `agent` reconciles into a StatefulSet whose `volumeClaimTemplates` are derived from the Agent's declared mounts. A mount marked `persist: true` becomes a PVC; a non-persisted mount becomes an `emptyDir` that dies with the pod. PVCs are `ReadWriteOnce` on ordinary single-writer storage — the agent pod is the volume's only writer, so no install needs a shared filesystem class. (Workspaces created before the RWO cutover sit on `ReadWriteMany` volumes until the storage migration below drains them.)
 
+A home volume mounts empty and shadows whatever the image bakes at that path, so the image's boot seeds it on the volume's first boot from the staged workspace: a no-clobber copy behind a sentinel file, run once per volume, so image content lands exactly once and files the user later edits are never overwritten by a restarted or upgraded image. The seed lives in the image's own boot sequence — the container entrypoint, replayed by the VM boot — with no separate init container, and the Agent CR's `spec.init` field (and a Template's `init`) is retained but no longer read.
+
 The default Claude Code template persists the workspace and `$HOME`. Together these hold:
 
 - the **workspace** itself — git checkouts, tool caches (`node_modules`, `.venv`, mise), and any artifacts the agent has produced.
--- a/docs/architecture/skills.md
+++ b/docs/architecture/skills.md
@@ -93,7 +93,7 @@ The reconciled `state` read has a consumer beyond the Skills surface: the Experi
 
 Every Local Skill carries a **provenance** verdict, judged by the agent-runtime at read time. The reference is the set of **pristine roots** in the image — exactly two sanctioned locations, both immutable and always in-pod, so no build-time manifest or on-PVC marker is needed (a marker was tried and reverted — third-party baked skills aren't ours to stamp, and the PVC is agent-writable anyway):
 
-1. the **pristine workspace copy** — the directory the first-boot init seeds onto the PVC and never touches again, and
+1. the **pristine workspace copy** — the directory the image's first-boot seed copies onto the PVC and never touches again, and
 2. the **staged-skills dir** — the one place images put system skills that must *not* reach every agent (an Agent Kind's Install Command copies them onto the PVC at create; the shared constant lives in the agent-runtime contract package).
 
 This is a deliberate convention, not a growing list: an image-shipped skill anywhere else will misclassify as user-authored, so new features ship their skills through one of these two locations.
--- a/packages/api-server/src/modules/agents/domain/spec-assembly.ts
+++ b/packages/api-server/src/modules/agents/domain/spec-assembly.ts
@@ -44,7 +44,6 @@ export function assembleSpecFromTemplate(
     image: tmplSpec.image,
     description: opts.description ?? tmplSpec.description,
     mounts: tmplSpec.mounts,
-    init: tmplSpec.init,
     env: tmplSpec.env,
     resources: concreteResources(tmplSpec.resources, opts.size, defaultLimits),
     imagePullPolicy: tmplSpec.imagePullPolicy,
--- a/packages/controller/pkg/reconciler/iptables_init_test.go
+++ b/packages/controller/pkg/reconciler/iptables_init_test.go
@@ -89,9 +89,8 @@ func TestBuildAgentStatefulSet_IptablesInitRunsFirst(t *testing.T) {
 	ss := BuildAgentStatefulSet("my-instance", testAgent, &cfg, configMapOwnerRef(testOwnerCM), "10.96.42.42")
 
 	ics := ss.Spec.Template.Spec.InitContainers
-	require.Len(t, ics, 2, "egress-lockdown + user init")
-	assert.Equal(t, "egress-lockdown", ics[0].Name, "lockdown must run before the user init")
-	assert.Equal(t, "init", ics[1].Name)
+	require.Len(t, ics, 1, "egress-lockdown is the only init container")
+	assert.Equal(t, "egress-lockdown", ics[0].Name)
 
 	agent := ss.Spec.Template.Spec.Containers[0]
 	require.NotNil(t, agent.SecurityContext)
--- a/packages/controller/pkg/reconciler/resources.go
+++ b/packages/controller/pkg/reconciler/resources.go
@@ -206,26 +206,12 @@ func BuildAgentStatefulSet(name string, agentSpec *types.AgentSpec, cfg *config.
 		resourceReqs.Requests = nil
 	}
 
-	initScript := agentSpec.Init
-	if initScript == "" {
-		initScript = defaults.Init
-	}
 	var initContainers []corev1.Container
 	if ic := buildIptablesInitContainer(cfg, gatewayClusterIP); ic != nil {
 		initContainers = append(initContainers, *ic)
 	}
 	if ic := buildNPGateInitContainer(cfg, gatewayClusterIP); ic != nil {
 		initContainers = append(initContainers, *ic)
-	}
-	if initScript != "" {
-		initContainers = append(initContainers, corev1.Container{
-			Name:            "init",
-			Image:           agentSpec.Image,
-			ImagePullPolicy: corev1.PullPolicy(pullPolicy),
-			Command:         []string{"sh", "-c", initScript},
-			Env:             []corev1.EnvVar{{Name: "HOME", Value: agentHome}},
-			VolumeMounts:    volumeMounts,
-		})
 	}
 
 	var pullSecrets []corev1.LocalObjectReference
--- a/packages/controller/pkg/reconciler/resources_test.go
+++ b/packages/controller/pkg/reconciler/resources_test.go
@@ -205,19 +205,8 @@ func TestBuildAgentStatefulSet_DefaultsToRunningReplicas(t *testing.T) {
 	assert.Equal(t, int32(1), *ss.Spec.Replicas)
 }
 
-func TestBuildAgentStatefulSet_InitContainer(t *testing.T) {
+func TestBuildAgentStatefulSet_IgnoresSpecInit(t *testing.T) {
 	ss := BuildAgentStatefulSet("my-instance", testAgent, testConfig, configMapOwnerRef(testOwnerCM), "")
-	require.Len(t, ss.Spec.Template.Spec.InitContainers, 1, "only the user-defined init runs")
-	ic := ss.Spec.Template.Spec.InitContainers[0]
-	assert.Equal(t, "init", ic.Name)
-	assert.Equal(t, "ghcr.io/myorg/agent:latest", ic.Image)
-	assert.Equal(t, []string{"sh", "-c", testAgent.Init}, ic.Command)
-}
-
-func TestBuildAgentStatefulSet_NoUserInitWhenEmpty(t *testing.T) {
-	agent := *testAgent
-	agent.Init = ""
-	ss := BuildAgentStatefulSet("my-instance", &agent, testConfig, configMapOwnerRef(testOwnerCM), "")
 	assert.Empty(t, ss.Spec.Template.Spec.InitContainers)
 }
 
--- a/packages/controller/pkg/reconciler/vm_resources.go
+++ b/packages/controller/pkg/reconciler/vm_resources.go
@@ -118,14 +118,9 @@ func BuildVMCloudInitSecret(name string, agentSpec *types.AgentSpec, cfg *config
 		"mkdir -p /tmp/agent-cache && chown %[1]d:%[1]d /tmp/agent-cache && { [ -L %[2]s/.cache ] || rm -rf %[2]s/.cache; } && ln -sfn /tmp/agent-cache %[2]s/.cache || true",
 		vmAgentUID, shellQuote(agentHome),
 	)})
-
-	initScript := agentSpec.Init
-	if initScript == "" {
-		initScript = defaults.Init
-	}
-	if initScript != "" {
-		cc.BootCmd = append(cc.BootCmd, []string{"env", "HOME=" + agentHome, "bash", "-c", initScript})
-	}
+	cc.BootCmd = append(cc.BootCmd, []string{"env", "HOME=" + agentHome, "sh", "-c",
+		`[ -f "$HOME/.initialized" ] || { cp -rn /app/working-dir/. "$HOME/" 2>/dev/null || true; touch "$HOME/.initialized"; }; mkdir -p "$HOME/work"`,
+	})
 
 	body, err := yaml.Marshal(cc)
 	if err != nil {
--- a/packages/controller/pkg/reconciler/vm_test.go
+++ b/packages/controller/pkg/reconciler/vm_test.go
@@ -25,7 +25,7 @@ func vmAgentCR() *apiv1.Agent {
 		{Path: "/home/agent", Persist: true, Size: "10Gi"},
 		{Path: "/scratchpad", Persist: false},
 	}
-	a.Spec.Init = "touch $HOME/.initialized"
+	a.Spec.Init = "echo custom-init-ignored"
 	return a
 }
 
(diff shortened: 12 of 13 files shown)


packages/controller/pkg/config/agent_config.go
```diff
@@ -91,7 +91,6 @@ type AgentTemplateDefaults struct {
 	Mounts       []Mount       `json:"mounts,omitempty"`
 	Env          []EnvVar      `json:"env,omitempty"`
 	SkillSources []SkillSource `json:"skillSources,omitempty"`
-	Init         string        `json:"init,omitempty"`
```
dam-code-guardian, line 94: 🔴 **Critical:** deleting `Init` from `AgentTemplateDefaults` turns a still-valid chart value into a fatal startup error. `AGENT_TEMPLATE_DEFAULTS` is decoded with `DisallowUnknownFields()` (`pkg/config/config.go:210`), and the chart renders the whole `templateDefaults` map to JSON. An install that still sets `controller.agent.templateDefaults.init` in its own values now fails `LoadFromEnv`, and `main.go:52` exits 1 — a controller crash-loop on upgrade, not the silent ignore the PR body promises.

The deployment template's own comment states the contract: "the controller fails loudly on unknown fields so values.yaml typos surface at startup". Removing the chart default does not remove the operator's override.

**Fix:** a field a chart value can still carry stays in the decoded struct after it stops being read. Keep `Init` and warn once at startup, exactly as `AgentBase.AccessMode` does at `config.go:222`.

```suggestion
	SkillSources []SkillSource `json:"skillSources,omitempty"`

	// Retained for backward compatibility and no longer read: home seeding
	// moved into agent-runtime's own boot. Kept so an install that still sets
	// it does not fail the strict decode.
	Init string `json:"init,omitempty"`
```

packages/agent-runtime/src/modules/home-seed.ts
```diff
@@ -0,0 +1,30 @@
...
+  homeDir: string;
+  imageDir: string;
+  workDir: string;
+  log: (msg: string) => void;
+}): void {
+  try {
+    const sentinel = join(opts.homeDir, SENTINEL);
+    if (!existsSync(sentinel)) {
+      if (existsSync(opts.imageDir)) {
+        cpSync(opts.imageDir, opts.homeDir, {
```
dam-code-guardian: 🔴 **Critical:** `cpSync` does not preserve a symlink's link text — it resolves it and writes an absolute path. The replaced `cp -rn` kept the text verbatim, so this is a behaviour change the tests do not catch.

Measured on node v24 with this PR's own fixture (image `.claude/skills -> ../.agents/skills`):

```
image link text: ../.agents/skills
home  link text: /tmp/.../image/.agents/skills
```

`/app/working-dir` is the read-only image copy, so the seeded `$HOME/.claude/skills` no longer follows the home volume. Skills installed at runtime into `$HOME/.agents/skills` stop being visible through the link, and `skills.md` classifies the pristine copy by exactly that path.

**Fix:** a copy that replaces `cp -rn` reproduces its symlink semantics — read each link with `readlinkSync` and recreate it with `symlinkSync`, then assert the link *text* in the test.

packages/agent-runtime/src/modules/home-seed.ts
```diff
@@ -0,0 +1,30 @@
...
+          recursive: true,
```
dam-code-guardian: 🔴 **Critical:** a type conflict between the volume and the image kills the process, and this `try`/`catch` cannot stop it. When the home volume holds a plain file where the image stages a directory, `cpSync` raises a C++ `std::filesystem::filesystem_error` that never becomes a JS exception:

```
terminate called after throwing an instance of 'std::filesystem::filesystem_error'
  what():  filesystem error: cannot create directory: File exists [.../home/aaa]
Aborted (core dumped)      # exit 134
```

That is agent-runtime dying on boot, which is the one outcome the module's own contract rules out ("failures log without blocking boot"). The reverse shape — a directory on the volume where the image ships a file — does throw catchably (`EINVAL`), so only this direction escapes. Reachable on any volume a user has written to before its first successful seed, including a seed that aborted partway.

**Fix:** a boot step that must never block boot cannot delegate to `cpSync` — walk the tree and copy entry by entry, skipping any destination that already exists whatever its type.

packages/agent-runtime/src/modules/home-seed.ts
```diff
@@ -0,0 +1,30 @@
...
+          force: false,
+          errorOnExist: false,
+        });
+        opts.log(`seeded home from ${opts.imageDir}`);
+      }
+      writeFileSync(sentinel, "");
+    }
+    mkdirSync(opts.workDir, { recursive: true });
```
dam-code-guardian: 🟡 **Warning:** the `mkdirSync` sits inside the same `try` as the copy, so any copy failure skips it. Verified: with an unwritable home, the log line is written and `workDir` does not exist. The sentinel is skipped too, so the next boot retries the copy — but this boot continues with no work directory, and `UNIT_BOUNDARY_DESCRIPTION` states it is ensured on every boot.

The blast radius is small because `create-child-agent-process.ts:24` also creates its `workingDir`, but the runtime-channel workspace plugins read it without that guard.

**Fix:** each independent boot step gets its own `try`/`catch`, so one step's failure never cancels a later step that does not depend on it.

packages/agent-runtime/src/server.ts
```diff
@@ -58,6 +59,15 @@ const workDir = config.PLATFORM_DEV
   ? join(__dir, "../working-dir")
   : config.WORK_DIR;
 
+if (!config.PLATFORM_DEV) {
+  seedHomeOnFirstBoot({
+    homeDir,
```
dam-code-guardian: 🟡 **Warning:** `homeDir` resolves from `config.HOME_DIR`, and nothing in the chart, the controller or any Dockerfile sets that variable — it appears only in `config.ts:9` and `server.ts:57`, so it always takes its `/home/agent` default. The removed init container received `HOME` from `agentHome`, which a template or `templateDefaults` can override, and the controller does set `HOME` (`resources.go:74`).

So an install with a non-default `agentHome` seeds `/home/agent` while the agent's real home is elsewhere: the sentinel lands on the wrong path, and the image content never reaches the volume.

**Fix:** when code moves off an init container, it reads the same configured value the container got — default `HOME_DIR` from `process.env.HOME` so the controller's existing plumbing keeps working.

packages/agent-runtime/src/__tests__/unit/home-seed.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  // TEST_SCENARIO: a brand-new volume mounts empty — the seed copies the image workspace in, preserves the skills symlink as a symlink, writes the sentinel, and creates the work directory.
+  it("seeds an empty home and preserves symlinks", () => {
+    seedHomeOnFirstBoot({ homeDir, imageDir, workDir, log });
+
+    expect(readFileSync(join(homeDir, ".claude.json"), "utf8")).toBe("{}");
+    expect(
+      readFileSync(join(homeDir, ".agents", "skills", "SKILL.md"), "utf8"),
+    ).toBe("baked");
+    expect(lstatSync(join(homeDir, ".claude", "skills")).isSymbolicLink()).toBe(
```
dam-code-guardian: 🟡 **Warning:** this assertion checks only that the entry is a symlink, so it passes while the link points into the image directory instead of the home volume. It is what let the `cpSync` link-rewrite above through.

**Fix:** a test for symlink preservation asserts the link text, because the link type alone does not prove the target survived.

```suggestion
    expect(readlinkSync(join(homeDir, ".claude", "skills"))).toBe(
      "../.agents/skills",
    );
```

dam-code-guardian (dismissed):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `9cbcc86`

### Summary
agent-runtime seeds `$HOME` from the image workspace on the volume's first boot, and the controller stops rendering the `init` container on both the container and VM backends. The seeding contract is documented and covered by five new unit tests.

### Findings
- 🔴 **Critical:** removing the `Init` field from the decoded `AgentTemplateDefaults` struct makes the controller crash-loop on any install that still sets `controller.agent.templateDefaults.init` (`packages/controller/pkg/config/agent_config.go:94`)
  **Fix:** a field that a chart value can still carry stays in the decoded struct after it stops being read — keep `Init` with a startup `slog.Warn`, exactly as `AgentBase.AccessMode` does for the deprecated `accessMode`.
- 🔴 **Critical:** `cpSync` rewrites relative symlinks to absolute image paths, so the seeded `$HOME/.claude/skills` points into `/app/working-dir` (`packages/agent-runtime/src/modules/home-seed.ts:17`)
- 🔴 **Critical:** a type conflict between a home entry and an image entry aborts the process, and the `try`/`catch` cannot hold it (`packages/agent-runtime/src/modules/home-seed.ts:17`)
- 🟡 **Warning:** a copy failure skips `mkdirSync(opts.workDir)` (`packages/agent-runtime/src/modules/home-seed.ts:26`)
- 🟡 **Warning:** the seed reads `config.HOME_DIR`, which nothing sets, so a per-template `agentHome` override seeds the wrong path (`packages/agent-runtime/src/server.ts:64`)
- 🟡 **Warning:** the symlink test asserts the link type but never the link target (`packages/agent-runtime/src/__tests__/unit/home-seed.test.ts:53`)
- 🟢 **Suggestion:** `(err as Error).message` yields `undefined` for a thrown non-`Error`, which hides the cause in the one line meant to explain the failure (`packages/agent-runtime/src/modules/home-seed.ts:28`).
- ✅ **Looks good:** the sentinel and no-clobber semantics match the old `cp -rn` behaviour, and the regression test that a set `spec.init` renders no init container pins the retained-but-unread contract.

- 🟡 **Warning:** the Skill Origin section still credits "the first-boot init" with seeding the pristine workspace copy onto the PVC, and this PR deletes that init container (`docs/architecture/skills.md:96`).
  **Fix:** when a PR relocates a responsibility between components, update every architecture page that names the old component, not only the page that owns the concept in depth.

### Issue Fit
Linked issue: #3242 (plain "Part of" reference, no closing keyword — merging closes nothing); the numbers come from #3246.

- 🟡 **Warning:** epic #3242 asks that changes are proven with numbers, and the Testing section gives unit tests, suite counts and a helm render but no boot timing on the path #3246 measured (`packages/agent-runtime/src/server.ts:62`).
  **Fix:** a latency PR under this epic reports the measured before/after on the same path as the baseline, so the epic's plan can be re-ranked from data.
- 🟢 **Suggestion:** the #3246 breakdown prices this step at ~1s and ranks it in none of its 8 fixes, while the top four stay open. A line on why it goes first would help.
- 🟢 **Suggestion:** the body claims a "~1-2s container create/start" cost; #3246 measured 1s on create and folded it into a 3s "np-gate + init" line on wake. Anchor the claim to those numbers.

The three 🔴/🟡 findings on `home-seed.ts` and `server.ts` are merged into `### Findings` above, with the inline comments carrying the detail.

### Verdict
REQUEST_CHANGES — a still-set chart value now crash-loops the controller, and the seed both breaks relative symlinks and can abort boot outright.

---

dam-code-guardian:
📊 **PR #3417 review artifact** — [DAM Artifact Library](https://share-dam-dev.apps.dam-cl.fmaas.res.ibm.com/a/fINCkquXTXaNjQ)

packages/controller/pkg/config/agent_config.go
jezekra1: Fixed for real, and it was the most valuable finding on this PR: the Init field is back in the decoded struct and a still-set templateDefaults.init now logs a startup deprecation warning instead of failing the strict decode. Resolving.

packages/agent-runtime/src/modules/home-seed.ts
jezekra1: Moot after the rework: the TypeScript seeder is gone and the entrypoint runs coreutils `cp -rn`, which preserves symlink text verbatim — the behaviour you measured cpSync breaking. Resolving.

packages/agent-runtime/src/modules/home-seed.ts
jezekra1: Moot after the rework: the uncatchable std::filesystem abort came from cpSync, which is gone — `cp -rn 2>/dev/null || true` in the entrypoint cannot abort the boot. Resolving.

packages/agent-runtime/src/modules/home-seed.ts
jezekra1: Moot after the rework: the work directory is now created by the entrypoint's own `mkdir -p` independent of the copy, and agent-runtime ensures it at boot for images that don't run the platform entrypoint. Resolving.

packages/agent-runtime/src/server.ts
jezekra1: Addressed twice over: the seed moved into the entrypoint, which honours $HOME (the value the controller sets from agentHome), so the seed path is correct by construction. The deeper runtime-wide gap you identified — HOME_DIR never being set anywhere — became #3418, which removes the knob entirely. Resolving.

packages/agent-runtime/src/__tests__/unit/home-seed.test.ts
jezekra1: Moot after the rework: that test file is gone along with the TypeScript seeder. Your point stood though — asserting link type rather than link text is what let the cpSync rewrite through, and it is why the rework went back to coreutils instead of hand-rolling the copy. Resolving.

packages/controller/pkg/config/config.go
```diff
@@ -222,6 +222,9 @@ func LoadFromEnv() (*Config, error) {
 	if cfg.AgentBase.AccessMode != "" {
 		slog.Warn("controller.agent.base.accessMode is deprecated and ignored — workspace volumes are always ReadWriteOnce (#2988); remove it from your values")
 	}
+	if cfg.AgentTemplateDefaults.Init != "" {
```
dam-code-guardian, line 225: 🟡 **Warning:** the notice covers the chart value, but two other surfaces still carry `init` and now silently do nothing: `deploy/helm/platform/templates/agent-templates.yaml:117` still renders a Template's `init` into the templates ConfigMap, `packages/api-server/src/modules/agents/domain/spec-assembly.ts:47` still copies it into the Agent spec, and the CRD field describes itself as "an optional one-shot init script run before the agent starts" (`packages/controller/api/v1/agent_types.go:17`). An operator who sets a per-template `init` gets no warning and no effect.

**Fix:** when a field is retired but retained, every surface an operator can still set it from gets the same notice — the `accessMode` precedent you cite carries it in two places (the struct field plus the `values.yaml` comment at `deploy/helm/platform/values.yaml:850`).

packages/agent-runtime/src/server.ts
```diff
@@ -58,6 +58,8 @@ const workDir = config.PLATFORM_DEV
   ? join(__dir, "../working-dir")
   : config.WORK_DIR;
 
+mkdirSync(workDir, { recursive: true });
```
dam-code-guardian: 🟡 **Warning:** this runs at module scope with no `try`/`catch`, so an environmental failure throws during module evaluation — before `server.listen` and before any stderr log surface in the file. `EACCES` from wrong ownership is the plausible one, and it is exactly the entrypoint-skipped case this line exists to serve. The operator sees a bare stack trace and a crash-loop; the codebase's own copy of this call on the same `workDir` value guards and logs.

**Fix:** guard module-scope filesystem mutations and emit a prefixed stderr line naming the path and `err.message`, as `modules/acp/infrastructure/create-child-agent-process.ts:23` does; if aborting is intended, catch, log, then exit deliberately.

```suggestion
try {
  mkdirSync(workDir, { recursive: true });
} catch (err) {
  process.stderr.write(
    `[server] could not create workDir ${workDir}: ${(err as Error).message}\n`,
  );
}
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `128f325`

## PR #3417: perf(platform-base): seed home in the image boot instead of an init container
**Author:** jezekra1 | **Branch:** perf/agent-boot-seed-fold → perf/agent-boot-ca-trust | **Changes:** +26 −53 (11 files)

### Summary
The revision deletes the TypeScript reimplementation of the copy and keeps coreutils `cp -rn` in the container entrypoint and the VM boot, so the controller stops rendering the `init` container. All three round-1 criticals are resolved at their cause.

### Changes since last review
Previous HEAD: `9cbcc86` (2026-08-21T12:48:02Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** the removed `Init` config field is back, with a startup deprecation warning on the `accessMode` model (`packages/controller/pkg/config/config.go:225`)
- ✅ **Fixed:** `cpSync` symlink rewriting — the whole `home-seed.ts` module is deleted, `cp -rn` runs instead (`packages/platform-base/entrypoint.sh:51`)
- ✅ **Fixed:** the uncatchable `std::filesystem` abort — the same deletion removes it
- ✅ **Fixed:** the copy failure skipping the `work` mkdir — the entrypoint's `mkdir -p` is now unconditional (`packages/platform-base/entrypoint.sh:55`)
- ✅ **Fixed:** the unset `HOME_DIR` — the entrypoint honours `$HOME`, which the controller sets from `agentHome` (`packages/platform-base/entrypoint.sh:49`)
- ✅ **Fixed:** the symlink test asserting only the link type — the test file is gone with the module
- ✅ **Fixed:** `skills.md` no longer credits the deleted first-boot init (`docs/architecture/skills.md:96`)
- 🔁 **Still present:** no measured after-number for a latency PR under epic #3242, now deferred to a follow-up comment (`issue-fit`)
- 🆕 **New:** the deprecation warning covers the chart value only, not the two other surfaces that still carry `init` (`packages/controller/pkg/config/config.go:225`)
- 🆕 **New:** `mkdirSync` at module scope is unguarded (`packages/agent-runtime/src/server.ts:61`)
- 🆕 **New:** `skills.md` was edited without bumping its `Last verified` header (`docs/architecture/skills.md:3`)

### Findings
- 🟡 **Warning:** the deprecation is announced for `templateDefaults.init` only, while an Agent's `spec.init` and a Template's `init` still flow end to end and now silently do nothing (`packages/controller/pkg/config/config.go:225`)
  **Fix:** when a field is retired but retained, every surface an operator can still set it from gets the same notice — the chart value, the CRD field description, and the Template knob.
- 🟡 **Warning:** unguarded module-scope `mkdirSync` (`packages/agent-runtime/src/server.ts:61`)
- 🟡 **Warning:** `skills.md` was edited without bumping its `Last verified` header (`docs/architecture/skills.md:3`)
  **Fix:** a page edited by a PR bumps its `Last verified: YYYY-MM-DD` in the same commit, per `docs/guidelines/documentation-guidelines.md:34`.
- 🟡 **Warning:** no measured after-number for a latency PR under an epic whose acceptance is proving changes with numbers (`issue-fit`, mirrored below)
  **Fix:** land the after-measurement before merge, or name the follow-up issue that owns the proof.
- 🟢 **Suggestion:** the `agent-runtime` fallback replicates the entrypoint's `mkdir -p` on `work` but not its seed step, so an image skipping the entrypoint gets `work/` on an unseeded home (`packages/agent-runtime/src/server.ts:61`).

- 🟡 **Warning:** `docs/architecture/skills.md` was edited (line 96 reworded from "the first-boot init" to "the image's first-boot seed") while its `Last verified:` header still reads 2026-08-21 (`docs/architecture/skills.md:3`).
  **Fix:** every architecture page a PR touches bumps its `Last verified: YYYY-MM-DD` header in the same commit.

### Issue Fit
Linked issue: #3242 "Epic - Measure and cut platform latency" (plain "Part of" reference, no closing keyword — merging closes nothing), with #3246 supplying the baseline measurement.

- 🟡 **Warning:** the epic's goal is "we can prove changes with numbers", and the PR carries no before/after number for its own change — the Testing section defers it to a comment after deployment. The numbers present are #3246's pre-change baseline, which price the cost removed, not the boot time achieved. Raised in round 1 and acknowledged rather than closed.
  **Fix:** a PR delivering an epic whose acceptance is "prove it with numbers" lands the after-measurement before merge, or states which follow-up issue owns the proof.
- 🟢 **Suggestion:** #3246's ranked fix list does not include removing the seed-home init container; its entries are image pull (−30s), `update-ca-trust` (−18s), create reconcile (−10–15s), readiness plumbing (−8–14s), kata boot, first-turn path, hygiene. The stacked base branch covers ranked item 2, so a line on where this increment sits in that ordering would help.

- 🟡 **Warning:** `mkdirSync(workDir, { recursive: true })` runs at module scope with no `try`/`catch`, so an environmental failure — `EACCES` from wrong ownership, which is exactly the entrypoint-skipped case this line serves, `EROFS`, or `ENOTDIR` when `work` exists as a file — throws during module evaluation, before `server.listen` and before any stderr log surface in the file. The codebase's own copy of this call on the same value guards and logs (`modules/acp/infrastructure/create-child-agent-process.ts:23`). (`packages/agent-runtime/src/server.ts:61`)
  **Fix:** guard module-scope filesystem mutations and emit a prefixed stderr line naming the path and `err.message`, matching `create-child-agent-process.ts:23`; a deliberate abort catches, logs, then exits.
- 🟢 **Suggestion:** the same ensure-dir concern already exists in `create-child-agent-process.ts:24` and `runtime-channel/infrastructure/file-ops.ts:296`, so this is a third copy with a third error policy — one `ensureDir` helper owning the policy would collapse them. Priced as a suggestion: the composition root is otherwise pure path derivation and `loadManifest` at line 76 already throws there on `main`, so an unguarded module-scope throw is the file's existing baseline.
- 🟢 **Suggestion:** the fallback replicates only half of the entrypoint's home preparation — the `work` mkdir but not the seed — so on an image skipping the entrypoint, `homeDir` is unseeded while `work/` exists (`packages/agent-runtime/src/server.ts:61`). Untouched consumers, so a suggestion.
- 🟢 **Suggestion:** `WORK_DIR: z.string().default("/home/agent/work")` — a Zod default does not apply to a present-but-empty value, so an empty `WORK_DIR` reaches `mkdirSync("")` (`packages/agent-runtime/src/modules/config.ts:11`). Untouched by this diff, which only moves where it surfaces.

### Verdict
COMMENT — every round-1 blocker is fixed at its cause and the coreutils approach removes the whole class; what remains is the deprecation's reach across the other two `init` surfaces, an unguarded boot-time mkdir, and two bookkeeping items.

---

packages/agent-runtime/src/server.ts
jezekra1: Fixed as suggested — the mkdir is wrapped and logs `[server] could not create workDir …` instead of throwing during module evaluation. Agreed on the reasoning: this line exists for the image-that-skips-the-entrypoint case, which is exactly where wrong ownership is plausible, and dying there is worse than continuing degraded since the harness spawn creates its own working dir (create-child-agent-process.ts:23, the guard pattern you pointed at).

packages/controller/pkg/config/config.go
jezekra1: Right that one notice wasn't enough. Rather than add notices to each surface, this now stops propagating the retired field: the chart no longer renders a Template's `init` (agent-templates.yaml) and the api-server no longer copies it into the Agent spec (spec-assembly.ts:47), so nothing writes it any more. `spec.init` stays in the CRD so existing Agents keep decoding, and `templateDefaults.init` keeps its startup warning plus the values.yaml note.

One correction on the CRD surface: the description you quoted can't carry the notice. This repo forbids prose comments in Go, so adding a doc comment on the `Init` field fails `common:check:comment-types` (I tried it — that's what the hook rejected). The 41 `description:` entries currently in `crds/agent-platform.ai_agents.yaml` are pre-policy leftovers: the Go source has no prose doc comments on `Agent`/`AgentSpec` any more, so `mise run controller:generate` strips every one of them. That's a real staleness worth its own cleanup, but it's not something this PR can fix without either violating the comment policy or shipping an unrelated ~100-line CRD diff.

tomkis (approved):

Issue #3402, opened 2026-08-20
A ready agent pod sits unused for ~10 seconds before it gets work


jezekra1:
## Context

#3246 measured the wake path. After the pod reports ready, three lags stack before the first message can flow. Investigation refined the diagnosis: the largest lag (the controller publishing Ready ~8s late) is queue wait behind reconcile churn. That was attributed to #3406, which has since shipped — but it removed only the resync half. The queue can still fill from a routine sweep over the fleet, because a reconcile with nothing to change still costs a round of live API reads; #3470 removes one source and #3472 covers the rest. So this issue's prerequisite is partly done, not done. This issue covers the api-server side, which is untouched.

## Problem

- The api-server discovers readiness by GET-polling the Agent CR with backoff growing to 5s intervals (~28 GETs per wake; mean ~3s, worst ~6s dead time after real readiness) — while it already maintains a live watch on Agent CRs for UI updates.
- Environment/config delivery retries on a ~1s loop against the same (late) Ready condition; env landed ~8s after pod readiness on a fresh create, and a brand-new agent's first harness start gates on that delivery (bounded 15s).
- The relay blocks on a keep-warm annotation PATCH between observing readiness and connecting (~100ms, bookkeeping only).

All three were re-checked against main on 2026-08-26 and still hold: the wake poll runs 500ms to 5s with 1.5x growth and jitter, the watch it should be using is implemented but consumed only by the live-events feed, the activity PATCH is awaited on both ready paths, and delivery re-checks readiness on a 1s timer.

## Goal

Once the Ready condition flips, the platform notices, connects, and delivers configuration within ~1s.

## Proposed solution

**The approach is not settled — see Open Questions.** The design below follows the eventing rule from #3288 (services emit domain events on the in-process bus; projections and sagas subscribe):

- One watch adapter, owned by the agents module, translates K8s Agent-CR watch events into an `AgentReadinessChanged` domain event on the existing in-process bus (raw, undebounced; watch reconnects re-list and emit synthetic events).
- `ensureReady` becomes a waiter resolved by the event — keeping its existing time budget, timeout classification via a final direct read, and a slow fallback re-check so a dropped watch cannot strand a waiter.
- The runtime-delivery dispatch reacts to the event (saga) so pending contributions apply immediately on readiness; the periodic sweep stays as the backstop.
- The live-events SSE watch becomes a subscriber of the same events (its 300ms debounce applied at the subscriber), deleting the duplicate watch.
[...]

PR #3429, opened 2026-08-21
perf(controller): reconcile on change, not on every resync

### LINKED ISSUE #3406: The controller re-reconciles every agent every 30 seconds
## Context

Telemetry gathered for #3246 shows ~2,800 reconcile passes per 20 minutes across ~68 agents, with the single reconcile worker ~83% busy at today's fleet size.

**Promoted from hygiene to prerequisite of #3402:** the measured 8s between a pod becoming ready and the controller publishing the Agent's Ready condition is not reconcile-pass duration (~350ms) — it is queue wait behind resync churn. Every readiness consumer (wake, config delivery) inherits that lag, so this issue is the foundation of the wake-latency work.

## Problem

Periodic informer resyncs re-enqueue every agent every 30s regardless of change — the update handlers enqueue unconditionally, and resync replays fire them for unchanged objects. Each pass also re-reads cluster state with live queries (including listing all pods in the namespace, ~500ms) instead of the caches the controller already maintains. The overhead grows linearly with fleet size, keeps the queue permanently backlogged, and delays event-driven reconciles (pod-ready, wake bumps) by seconds.

## Goal

Reconcile work scales with change, not with fleet size; an event-driven reconcile (pod flip, annotation bump) starts within ~1s.

## Scope

- Skip enqueueing update events whose resourceVersion did not change (both the agent and pod informers) — resync replays become no-ops.
- Keep drift healing: an explicit slow full sweep (minutes, not 30s) re-enqueues all agents so hand-edited owned resources still reconcile back.
- Single worker retained for now — multiple workers need a reconciler concurrency audit (owner locks, denied-wake memo) and can follow separately. Live-query-vs-cache reads likewise follow separately if still needed.

## Done when

Steady-state reconcile rate drops by an order of magnitude at today's fleet size, while a wake, create, or pod-ready event still reconciles (and publishes readiness) within ~1s.

jezekra1:
Closes #3406. Prerequisite of #3402 — first layer of the readiness-latency stack.

## Problem

Both informers enqueue on every update event, and 30s resyncs replay updates for unchanged objects — so all ~68 agents re-enqueue every 30s. At ~350ms per pass (measured via OTel traces: ~2,800 passes/20min), the single worker sits ~83% busy and event-driven reconciles queue behind the churn: a pod flipping Ready waited **8s** before the controller published the Agent's Ready condition, a lag inherited by every wake and every config delivery.

## Change

- Update events whose `resourceVersion` did not change are not enqueued (agent and pod informers) — resync replays become no-ops, so the queue holds only real changes and event→reconcile latency drops to roughly one pass (~350ms).
- Drift healing (hand-edited owned resources reconciling back) previously rode the 30s resync; it moves to an explicit sweep re-enqueuing every agent every 5 minutes.

Deliberately out of scope, as #3406 notes: multiple workers (needs a reconciler concurrency audit) and informer-cache reads inside the pass.

## Correctness notes

- The readiness condition's semantics are untouched — same predicate, same publish point at the end of a pass. This PR only removes the queue wait in front of it. The k8s-native staleness contract (status lags spec, `observedGeneration` tells you) is unchanged.
- Resync-driven level-triggering is preserved, at 5min instead of 30s; startup coverage is unchanged (informer Add events fire for every existing object at cache sync).

## Testing

- New unit tests: resourceVersion filter (same/different/fail-open on non-objects) and the sweep's enqueue-all.
- Controller suite, staticcheck, gofmt, build all green.
- Expected observable effect on dev after deploy: steady-state `reconcile agent` span rate drops from ~2.4/s to near zero between changes, and pod-ready → CR Ready publish lands well under 1s. Will verify against the telemetry store and post numbers on #3406.


jezekra1:
Review note on filter safety (asked offline): `resourceVersion` is bumped by the API server on **every persisted write** to the object — spec, labels, and annotations alike, including status-subresource writes. So the wake path (the `last-activity` annotation PATCH), stop/pause stamps, and any label change all produce a real watch event with a new resourceVersion and are never filtered. A write that changes nothing produces no event at all (the apiserver dedupes no-op updates), so there is nothing to miss there either. The only thing the filter drops is the informer's periodic resync, which replays the cached object with an unchanged resourceVersion. This is the same semantics as controller-runtime's upstream `ResourceVersionChangedPredicate`.

packages/controller/main.go
```diff
@@ -183,6 +190,8 @@ func run(ctx context.Context, client kubernetes.Interface, dynClient dynamic.Int
 	}
 	slog.Info("informer caches synced")
 
+	go runDriftSweep(ctx, agentInformer.Informer().GetStore(), agentQueue, 5*time.Minute)
```
dam-code-guardian, line 193: 🟡 **Warning:** The 30-second resync was the retry that made a *parked* agent auto-start soon after room freed. `docs/architecture/budgets.md:46` states a never-hibernate or sweepable agent starts "within ~30s of room freeing (the informer's resync guarantees the pass)". That path is not event-driven: freeing room hibernates a *different* agent, and nothing enqueues the parked one — `reclaimIdleRoom` hibernates the candidates and returns, and the budget memo clears only on the parked agent's own next pass. With the resync gone, this sweep is that pass, so the window becomes ~5 minutes.

A sweepable invocation target is one of the two exempt classes. At the 60-second minimum TTL (`MIN_INVOCATION_TTL_MS`, `packages/api-server-api/src/modules/invocations/schemas.ts:4`) it can reach its liveness deadline before the sweep retries it; `invocation-liveness.ts` then fails and reaps it.

`docs/architecture/platform-topology.md:37` also still describes the controller as watch-driven only, with no periodic healing path.

**Fix:** When a PR changes the cadence behind a documented timing guarantee, carry the new number to that guarantee's architecture page in the same PR — here `budgets.md:46` and the controller section of `platform-topology.md`. If the ~30s auto-start is still wanted for the two exempt classes, give that path its own shorter tick rather than the 5-minute drift interval.

(Reported by my diff review and by doc-drift.)

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `0c08549`

## PR #3429: perf(controller): reconcile on change, not on every resync
**Author:** jezekra1 | **Branch:** perf/controller-reconcile-churn → main | **Changes:** +96 −2 (2 files)

### Summary
Both informers now drop update events whose `resourceVersion` did not change, so the 30-second resync no longer re-enqueues all ~68 agents; drift healing moves to an explicit 5-minute sweep. The change removes the queue wait in front of readiness publishing without touching the readiness predicate.

### Findings
- 🟡 **Warning:** parked-agent auto-start window moves from ~30s to ~5min, and two docs still state the old guarantee (`packages/controller/main.go:193`) — see inline
- ✅ **Looks good:** The filter does not become self-triggering. `updateAgentStatus` compares desired against current and returns without a write when they are equal (`packages/controller/pkg/reconciler/status.go:33`), so a readiness publish costs one extra pass and then converges.

Both findings are merged into the ‘Warning’ above — the stale ~30s guarantee at `docs/architecture/budgets.md:46` and the missing periodic-healing description at `docs/architecture/platform-topology.md:37`.

### Issue Fit
Linked issue: #3406 "The controller re-reconciles every agent every 30 seconds" (closing keyword)

- 🟢 **Suggestion:** Issue #3406 closes its "Done when" on a measured result — a tenfold drop in the steady-state reconcile rate. The PR defers that measurement to a later comment on #3406, and the closing keyword closes the issue at merge (`packages/controller/main.go:190`).

### Verdict
COMMENT — the mechanism is right and the tests cover it; one documented timing guarantee now rests on a 5-minute tick instead of a 30-second one.

---

jezekra1:
Fixed in the latest push, with the failure mode you named as the driver:

- The reconciler now tracks agents parked over budget in the two exempt classes (never-hibernate, sweepable) — recorded when a pass parks them, cleared the moment a pass admits them, on any non-parked outcome, and on delete.
- A dedicated **30s retry tick re-enqueues just that set** (usually empty), restoring the documented ~30s auto-start after room frees — this covers every freeing cause including a raised UserBudget ceiling, which a pod-event-driven approach would miss — and keeps a sweepable Invocation target comfortably ahead of its 60s liveness deadline.
- Memo-parked agents (the deliberate no-auto-start class) are excluded from the set, so the no-spontaneous-start rule in budgets.md is untouched.
- Docs carried in the same push: `budgets.md` now names the retry tick as the mechanism behind the ~30s guarantee, and `platform-topology.md` describes the controller as change-driven with the two bounded periodic passes.
- New tests: exempt parked agent enters/leaves the retry set across deny→admit; memo-parked agent never enters it.

packages/controller/main.go
jezekra1: Addressed in be2057745, threading here for visibility (my earlier reply landed as a top-level comment, sorry). You're right that this path cannot be event-driven — freeing room has no single event (a peer scaling down, an agent deleted, a UserBudget ceiling raised), and the parked agent's own CR never changes. So the fix is a dedicated timer, scoped to exactly this case: every reconcile pass records/clears agents parked over budget in the two exempt classes (`recordParkedRetry`/`clearParkedRetry` around the park decision, cleared on admit and on delete), and `runParkedRetry` re-enqueues just that set every **30s** — restoring the documented ~30s auto-start and keeping a sweepable Invocation target ahead of its 60s liveness deadline. The 5-minute drift sweep stays unrelated housekeeping. Memo-parked agents are excluded, so the deliberate no-spontaneous-start rule holds. budgets.md and platform-topology.md updated in the same commit; tests pin the set's add/remove across deny→admit and the memo exclusion.

jezekra1: Fixed as described in the reply above; current head after the rebase onto main is 4541d7a66. Resolving.

tomkis (approved):

PR #3411, opened 2026-08-21
Share a knowledge base as a read-only MCP endpoint

jjeliga:
One curated knowledge base can serve a whole team: its owner shares it behind a durable link (or tells the KB's own agent to share it), teammates paste that link into their own account, and every sandbox they grant it to gets a built-in `knowledge-bases` MCP server that can list, search, grep, and read the shared content. Consumers can never modify anything — the endpoint is read-only by construction — and no LLM runs in the serving path: nobody pays ingestion twice, and consumers' own agents do the reading on their own credentials. Shared content stays fresh automatically: the knowledge base's own pod watches the shared folders and republishes shortly after any edit — a chat turn, SSH, a script, or the file panel — with an explicit "Refresh now" that skips the wait.

Closes #3291

## Sub-issues

- [x] 01 — Share core
- [x] 02 — Snapshot publish
- [x] 03 — Aggregate MCP endpoint
- [x] 04 — Search & grep
- [x] 05 — Consumer connection rail
- [x] 06 — Agent share tools
- [x] 07 — Auto-refresh
- [x] 08 — Owner UI
- [x] 09 — Consumer UI
- [x] 10 — Docs, values, smoke

## Changes beyond #3291's ask

- **Consumer serving moved in-cluster.** The aggregate `knowledge-bases` MCP is served over the in-cluster harness channel (`/api/agents/:id/kb`), the same path as `platform-outbound`, rather than the agent's egress gateway dialing the public share host. Agents cannot reach the share host in-cluster (a `*.localhost` dev host resolves to loopback via the resolver and never hits CoreDNS; a real host would need a fragile public-LB hairpin, and the agent NetworkPolicy fences in-cluster targets), and the constraint is strictly-in-cluster traffic. The public share-host `/mcp/kb` endpoint remains for external by-link use and the e2e.
[...]

PR #3433, opened 2026-08-24
fix(ui): surface the required provider across create and save flows

### LINKED ISSUE #3376: UI - Nothing tells the user a provider is required to create an agent
## Problem

Provider is required, but the UI does not make that requirement clear. The provider field starts empty, looks like any other untouched field, and can leave **Create** disabled with no explanation. This affects coding agents, experiments, and knowledge bases.

The same issue can occur on existing agents if a user clears all of their provider keys. The provider card should make the missing-provider error obvious.

## Expected behavior

* **LiteLLM is pre-selected** when the user already has a LiteLLM connection.
* If the user has no provider connection, the provider remains unselected and **Create** is disabled.
* An empty required provider field shows the standard error state **immediately**: red border + warning icon.
* The provider card for an existing agent shows the same error state when the agent has no provider.
* A provider is required to create or save a coding agent, experiment, or knowledge base.
* The behavior is consistent across all three creation flows.

## Acceptance criteria

* Users can always see which field is preventing creation.
* Users with a LiteLLM connection do not need to make a provider selection.
* Users cannot create or save without a provider.
* Clearing all provider keys from an existing agent clearly surfaces the missing-provider error on the provider card.

kapetr:
Closes #3376

The provider field now shows the standard error state (red border, warning icon) immediately when it is empty, in all three creation flows and on an existing agent's provider card. Submit on agent settings is disabled while the provider is missing, and the Setup nav item gets a warning dot.

A draft holding a provider whose connection was deleted is repaired automatically: the first connected provider is re-selected, or the value is cleared so the error state and disabled Create button apply.

LiteLLM pre-selection already worked (recommended providers sort first in auto-select) and is unchanged.

### DIFF
--- a/packages/ui/src/components/ui/rich-select.tsx
+++ b/packages/ui/src/components/ui/rich-select.tsx
@@ -1,4 +1,4 @@
-import { Checkmark, ChevronDown } from "@carbon/icons-react";
+import { Checkmark, ChevronDown, WarningFilled } from "@carbon/icons-react";
 import type { ReactNode } from "react";
 
 import { CARD_SURFACE } from "@/components/ui/card";
@@ -29,6 +29,7 @@ interface Props<T extends string> {
   placeholder: string;
   ariaLabel?: string;
   disabled?: boolean;
+  invalid?: boolean;
   testId?: string;
 }
 
@@ -39,6 +40,7 @@ export function RichSelect<T extends string>({
   placeholder,
   ariaLabel,
   disabled = false,
+  invalid = false,
   testId,
 }: Props<T>) {
   const selected = options.find((o) => o.value === value);
@@ -48,10 +50,13 @@ export function RichSelect<T extends string>({
         <button
           type="button"
           disabled={disabled}
+          aria-invalid={invalid || undefined}
           data-testid={testId}
           className={cn(
             CARD_SURFACE,
             "group flex min-h-[76px] w-full items-center gap-3.5 p-4 text-left transition-colors hover:border-muted-foreground disabled:pointer-events-none disabled:opacity-50 data-[state=open]:border-foreground",
+            invalid &&
+              "border-destructive hover:border-destructive data-[state=open]:border-destructive",
           )}
         >
           {ariaLabel && <span className="sr-only">{ariaLabel}</span>}
@@ -76,6 +81,9 @@ export function RichSelect<T extends string>({
             <p className="flex-1 text-[15px] text-muted-foreground">
               {placeholder}
             </p>
+          )}
+          {invalid && (
+            <WarningFilled size={18} className="shrink-0 text-destructive" />
           )}
           <ChevronDown
             size={18}
--- a/packages/ui/src/modules/platform/store/navigation.ts
+++ b/packages/ui/src/modules/platform/store/navigation.ts
@@ -100,9 +100,9 @@ export const createNavigationSlice: StateCreator<
     history.pushState(
       null,
       "",
-      routeToPath({ view: "knowledge-base-config", agentId }),
+      routeToPath({ view: "sandbox-home", agentId, sandboxSection: "setup" }),
     );
-    set({ view: "knowledge-base-config", agentId });
+    set({ view: "sandbox-home", agentId, sandboxSection: "setup" });
   },
   mobileScreen: "sessions",
   setMobileScreen: (screen) => set({ mobileScreen: screen }),
--- a/packages/ui/src/modules/providers/components/provider-select.tsx
+++ b/packages/ui/src/modules/providers/components/provider-select.tsx
@@ -32,7 +32,8 @@ export function ProviderSelect({
   allow,
   recommended,
 }: Props) {
-  const { itemByType, typeByConnectionId, isPending } = useProviderItems();
+  const { itemByType, typeByConnectionId, isPending, isSuccess } =
+    useProviderItems();
   const [connecting, setConnecting] = useState<ProviderPresetType | null>(null);
   const [connected, setConnected] = useState<{
     id: string;
@@ -43,31 +44,31 @@ export function ProviderSelect({
     [allow, recommended],
   );
 
-  const selectedType = selected
+  const resolvedType = selected
     ? (typeByConnectionId.get(selected.id) ??
       (connected?.id === selected.id ? connected.type : null))
     : null;
+  const selectedType =
+    resolvedType && rows.some((row) => row.type === resolvedType)
+      ? resolvedType
+      : null;
 
-  const firstConnected = useMemo(
-    () => rows.map((r) => itemByType.get(r.type)).find(Boolean),
-    [rows, itemByType],
+  const autoSelectTarget = useMemo(
+    () =>
+      autoSelectFirst
+        ? rows.map((r) => itemByType.get(r.type)).find(Boolean)
+        : undefined,
+    [autoSelectFirst, rows, itemByType],
   );
 
   useEffect(() => {
-    if (isPending || selectedType) return;
-    if (autoSelectFirst && firstConnected) {
-      onSelect(providerRef(firstConnected));
+    if (!isSuccess || selectedType) return;
+    if (autoSelectTarget) {
+      onSelect(providerRef(autoSelectTarget));
     } else if (selected) {
       onSelect(null);
     }
-  }, [
-    isPending,
-    selectedType,
-    autoSelectFirst,
-    firstConnected,
-    selected,
-    onSelect,
-  ]);
+  }, [isSuccess, selectedType, autoSelectTarget, selected, onSelect]);
 
   const pick = async (type: ProviderPresetType) => {
     if (type === selectedType) return;
@@ -110,9 +111,7 @@ export function ProviderSelect({
         placeholder="Select a provider"
         ariaLabel="Provider"
         disabled={disabled}
-        invalid={
-          required && !selectedType && !(autoSelectFirst && firstConnected)
-        }
+        invalid={required && isSuccess && !selectedType && !autoSelectTarget}
         testId="provider-select"
       />
       {connecting && (
--- a/packages/ui/src/modules/providers/components/provider-select.tsx
+++ b/packages/ui/src/modules/providers/components/provider-select.tsx
@@ -13,10 +13,11 @@ import { type ProviderRef, providerRef } from "./provider-item.js";
 
 interface Props {
   selected: ProviderRef | null;
-  onSelect: (ref: ProviderRef) => void;
+  onSelect: (ref: ProviderRef | null) => void;
   confirmSwitch?: () => Promise<boolean>;
   autoSelectFirst?: boolean;
   disabled?: boolean;
+  required?: boolean;
   allow?: readonly ProviderPresetType[];
   recommended?: ProviderPresetType;
 }
@@ -27,6 +28,7 @@ export function ProviderSelect({
   confirmSwitch,
   autoSelectFirst = false,
   disabled = false,
+  required = false,
   allow,
   recommended,
 }: Props) {
@@ -46,11 +48,26 @@ export function ProviderSelect({
       (connected?.id === selected.id ? connected.type : null))
     : null;
 
+  const firstConnected = useMemo(
+    () => rows.map((r) => itemByType.get(r.type)).find(Boolean),
+    [rows, itemByType],
+  );
+
   useEffect(() => {
-    if (!autoSelectFirst || selected) return;
-    const first = rows.map((r) => itemByType.get(r.type)).find(Boolean);
-    if (first) onSelect(providerRef(first));
-  }, [autoSelectFirst, selected, itemByType, rows, onSelect]);
+    if (isPending || selectedType) return;
+    if (autoSelectFirst && firstConnected) {
+      onSelect(providerRef(firstConnected));
+    } else if (selected) {
+      onSelect(null);
+    }
+  }, [
+    isPending,
+    selectedType,
+    autoSelectFirst,
+    firstConnected,
+    selected,
+    onSelect,
+  ]);
 
   const pick = async (type: ProviderPresetType) => {
     if (type === selectedType) return;
@@ -93,6 +110,9 @@ export function ProviderSelect({
         placeholder="Select a provider"
         ariaLabel="Provider"
         disabled={disabled}
+        invalid={
+          required && !selectedType && !(autoSelectFirst && firstConnected)
+        }
         testId="provider-select"
       />
       {connecting && (
--- a/packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-form.ts
+++ b/packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-form.ts
@@ -154,7 +154,8 @@ export function useSandboxSettingsForm() {
   });
 
   const dirty = isDirty || net.dirty || harnessDraft.dirty;
-  const isSubmitDisabled = saving || !formReady || !dirty;
+  const isSubmitDisabled =
+    saving || !formReady || !dirty || selectedProvider === null;
 
   useUnsavedGuard(dirty);
 
@@ -194,6 +195,7 @@ export function useSandboxSettingsForm() {
     control,
     errors,
     saving,
+    formReady,
     selectedProvider,
     selectProvider,
     currentPreset,
--- a/packages/ui/src/modules/sandboxes/hooks/use-section-summaries.ts
+++ b/packages/ui/src/modules/sandboxes/hooks/use-section-summaries.ts
@@ -47,7 +47,7 @@ export function useSectionSummaries(agent: AgentView | null): {
   warnings: SectionWarnings;
 } {
   const { data: templates = [] } = useTemplates();
-  const { data: apps = [], isPending: appsPending } = useAppConnections();
+  const { data: apps = [], isSuccess: appsLoaded } = useAppConnections();
   const connectionsQuery = useAgentConnections(agent?.id ?? null);
   const { data: schedules = [] } = useSchedules(agent?.id ?? null);
   const skillsState = useSkillsState(agent?.id ?? null);
@@ -74,11 +74,11 @@ export function useSectionSummaries(agent: AgentView | null): {
   );
 
   const missingProvider = useMemo(() => {
-    if (appsPending || !connectionsQuery.data) return false;
+    if (!appsLoaded || !connectionsQuery.data) return false;
     return !connectionsQuery.data.connections.some((c) =>
       providerAppIds.has(c.connectionId),
     );
-  }, [appsPending, connectionsQuery.data, providerAppIds]);
+  }, [appsLoaded, connectionsQuery.data, providerAppIds]);
 
   const staleModel = useStaleModel(agent?.id ?? null);
   const sourceCount = useSkillSourceCount(agent?.id ?? null);
--- a/packages/ui/src/modules/sandboxes/hooks/use-section-summaries.ts
+++ b/packages/ui/src/modules/sandboxes/hooks/use-section-summaries.ts
@@ -30,6 +30,8 @@ type SectionSummaries = Partial<Record<SandboxSection, string>>;
 type SectionWarnings = Partial<Record<SandboxSection, string>>;
 
 const STALE_MODEL_WARNING = "Saved model not offered by the current provider";
+const NO_PROVIDER_WARNING =
+  "No provider connected — the agent cannot reach a model";
 
 const plural = (n: number, word: string) => `${n} ${word}${n === 1 ? "" : "s"}`;
 
@@ -45,7 +47,7 @@ export function useSectionSummaries(agent: AgentView | null): {
   warnings: SectionWarnings;
 } {
   const { data: templates = [] } = useTemplates();
-  const { data: apps = [] } = useAppConnections();
+  const { data: apps = [], isPending: appsPending } = useAppConnections();
   const connectionsQuery = useAgentConnections(agent?.id ?? null);
   const { data: schedules = [] } = useSchedules(agent?.id ?? null);
   const skillsState = useSkillsState(agent?.id ?? null);
@@ -70,6 +72,13 @@ export function useSectionSummaries(agent: AgentView | null): {
       ),
     [apps],
   );
+
+  const missingProvider = useMemo(() => {
+    if (appsPending || !connectionsQuery.data) return false;
+    return !connectionsQuery.data.connections.some((c) =>
+      providerAppIds.has(c.connectionId),
+    );
+  }, [appsPending, connectionsQuery.data, providerAppIds]);
 
   const staleModel = useStaleModel(agent?.id ?? null);
   const sourceCount = useSkillSourceCount(agent?.id ?? null);
@@ -163,6 +172,10 @@ export function useSectionSummaries(agent: AgentView | null): {
       artifacts: artifactsSummary,
       usage: usageSummary,
     },
-    warnings: staleModel.stale ? { setup: STALE_MODEL_WARNING } : {},
+    warnings: missingProvider
+      ? { setup: NO_PROVIDER_WARNING }
+      : staleModel.stale
+        ? { setup: STALE_MODEL_WARNING }
+        : {},
   };
 }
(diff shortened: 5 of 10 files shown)


packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-form.ts
```diff
@@ -154,7 +154,8 @@ export function useSandboxSettingsForm() {
   });
 
   const dirty = isDirty || net.dirty || harnessDraft.dirty;
-  const isSubmitDisabled = saving || !formReady || !dirty;
+  const isSubmitDisabled =
+    saving || !formReady || !dirty || selectedProvider === null;
```
dam-code-guardian, line 158: 🔴 **Critical:** `KnowledgeBaseConfigView` uses this same hook and gates its **Submit changes** button on `isSubmitDisabled` (`knowledge-base-config-view.tsx:117`), but it renders no `ProviderSelect` and no section nav, so it also has no `warnings.setup` dot. A knowledge base whose provider connection was removed therefore has Submit permanently disabled with nothing on screen naming the cause — the failure mode issue #3376 exists to remove ("can leave Create disabled with no explanation").

**Fix:** A save gate and its error signal ship together: on every view that consumes `isSubmitDisabled`, render the provider card with `required` (as `SandboxSetupSection` does), or scope the `selectedProvider === null` term to views that expose a provider selector.

packages/ui/src/modules/providers/components/provider-select.tsx
```diff
@@ -46,11 +48,26 @@ export function ProviderSelect({
...
   useEffect(() => {
-    if (!autoSelectFirst || selected) return;
-    const first = rows.map((r) => itemByType.get(r.type)).find(Boolean);
-    if (first) onSelect(providerRef(first));
-  }, [autoSelectFirst, selected, itemByType, rows, onSelect]);
+    if (isPending || selectedType) return;
+    if (autoSelectFirst && firstConnected) {
+      onSelect(providerRef(firstConnected));
+    } else if (selected) {
+      onSelect(null);
```
dam-code-guardian, line 69: 🟡 **Warning:** `useProviderItems` reads `useAppConnections` with a `data: connections = []` default, so a failed query (retries exhausted) leaves `isPending: false` with an empty `itemByType` — indistinguishable here from "no providers connected". This branch then calls `onSelect(null)` and, in agent settings, `selectProvider(null)` writes with `shouldDirty: true`, so a transient API outage silently clears a saved provider from the staged form; the `resetField` resync at `use-sandbox-settings-form.ts:119` bails out once `assignedAppIds` is dirty, so the value does not come back when the query recovers.

**Fix:** Repair the draft only on a *resolved* empty list — carry the query's error/success state out of `useProviderItems` and treat "not loaded" like `isPending` (early return), never as "the connection is gone".

packages/ui/src/modules/providers/components/provider-select.tsx
```diff
@@ -93,6 +110,9 @@ export function ProviderSelect({
         placeholder="Select a provider"
         ariaLabel="Provider"
         disabled={disabled}
+        invalid={
+          required && !selectedType && !(autoSelectFirst && firstConnected)
```
dam-code-guardian: 🟡 **Warning:** `selectedType` comes from the unfiltered `typeByConnectionId`, while `RichSelect`'s `options` come from `allow`-filtered `rows` (`setup-policy.ts:15` restricts experiments and knowledge bases to `ibm-litellm` and `anthropic`). A `providerRef` pointing at, say, an `openai` connection — restorable from the `sessionStorage` draft in `use-setup-form.ts` — makes `selectedType` non-null, so `invalid` stays false and the effect's early return fires, while `RichSelect` finds no matching option and shows the placeholder. The field looks empty and valid, and `canCreate` passes. Reported by react-ui-engineering and the diff review.

**Fix:** Resolve the selection against the offered rows once — an `offeredSelectedType` that is `selectedType` only when `rows` contains it, else `null` — and use that single value for `value`, `invalid`, and the effect's guard, so display, validity, and auto-select cannot disagree.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `d2001f1c`

## PR #3433: fix(ui): surface the required provider across create and save flows
**Author:** kapetr | **Branch:** fix/3376-provider-required → main | **Changes:** +58 −12 (8 files)

### Summary
The PR adds an `invalid` state to `RichSelect`, threads `required` through `ProviderSelect` to the three creation flows and the agent settings card, blocks save while no provider is staged, and adds a Setup nav warning. The mechanism matches issue #3376's ask; two surfaces reached by the new save gate do not carry the matching error signal.

### Findings
- 🔴 **Critical:** Knowledge-base settings gets the new save gate but renders no provider surface (`packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-form.ts:158`)
- 🟡 **Warning:** A failed connections read clears the draft's provider (`packages/ui/src/modules/providers/components/provider-select.tsx:61`)
- 🟡 **Warning:** `selectedType` is not filtered by `allow`, so an off-policy selection reads as valid (`packages/ui/src/modules/providers/components/provider-select.tsx:114`)

✅ No findings.

### Issue Fit
Linked issue: #3376 (closing keyword). The three creation flows, the pre-selection rule, and the existing-agent card all match the acceptance list.

- The knowledge-base gap is the 🔴 above — same acceptance criterion ("Users can always see which field is preventing creation").
- 🟢 **Suggestion:** The new no-provider warning replaces `STALE_MODEL_WARNING` in the single `warnings.setup` slot, so an agent with both conditions loses the stale-model warning (`packages/ui/src/modules/sandboxes/hooks/use-section-summaries.ts:176`). The precedence change is not declared in the body.

- 🟢 **Suggestion:** The "an agent must have a provider" rule lives only in the client — neither the contract schema nor `setAgentConnections` rejects a provider-less selection, so a non-UI caller bypasses it (`packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-form.ts:158`). Pre-existing and outside this PR's UI scope, so noted rather than blocking.
- 🟢 **Suggestion:** The same rule is now derived twice — the save gate reads staged `assignedAppIds`, `missingProvider` reads saved `connectionsQuery.data` — so the nav dot and the save gate can disagree for a staged clear (`packages/ui/src/modules/sandboxes/hooks/use-section-summaries.ts:76`). The card's red border still shows, so the effect is a lagging nav dot.

- The `allow` / `selectedType` mismatch is the second 🟡 above; react-ui-engineering and the diff review both reported it.
- 🟢 **Suggestion:** The invalid state carries no visible message — `aria-invalid`, a destructive border, and an unlabeled `WarningFilled` glyph say something is wrong without naming it, while every other field routes errors through `FormField`/`FormError` (`packages/ui/src/components/ui/rich-select.tsx:53`). Issue #3376 asked for exactly "red border + warning icon", so this is beyond its ask; an optional error-message prop linked by `aria-describedby` would close it.
- 🟢 **Suggestion:** `autoSelectFirst && firstConnected` is written twice — the effect's auto-select branch and the `invalid` suppression term — so the two must be kept in sync by hand (`packages/ui/src/modules/providers/components/provider-select.tsx:58`, `:114`). One derived `willAutoSelect` value serves both.
- 🟢 **Suggestion:** `required={f.formReady}` overloads `required` with "the form finished hydrating", while the other three call sites pass a bare `required` (`packages/ui/src/modules/sandboxes/components/sandbox-setup-section.tsx:89`).

### Verdict
REQUEST_CHANGES — the knowledge-base settings view now has a permanently disabled Submit with no on-screen reason, which is the exact failure mode issue #3376 sets out to remove.

---

packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-form.ts
kapetr: True today, but `KnowledgeBaseConfigView` is going away — the configure link will lead to the full sandbox configuration page (see #3411), which renders the provider card with the error state and the Setup nav warning. Leaving the gate in the shared hook as is.

kapetr: Update: the configure link now opens the full sandbox configuration directly (7716bb49), so the KB view no longer reaches this gate without a provider surface. `KnowledgeBaseConfigView` itself still goes away with #3411.

packages/ui/src/modules/platform/store/navigation.ts
```diff
@@ -100,9 +100,9 @@ export const createNavigationSlice: StateCreator<
     history.pushState(
       null,
       "",
-      routeToPath({ view: "knowledge-base-config", agentId }),
+      routeToPath({ view: "sandbox-home", agentId, sandboxSection: "setup" }),
```
dam-code-guardian, line 103: 🟡 **Warning:** This sends a knowledge base to `SandboxHomeView`, which is the agent surface. Three things that were knowledge-base-specific do not follow it: the icon rail marks "Knowledge bases" active only for the three `knowledge-base-*` views (`icon-rail.tsx:73`), so the rail highlight is lost; the delete action becomes "Delete Agent … all persistent data" (`sandbox-home-header.tsx:64`) instead of "Delete knowledge base … all of its knowledge and data" (`confirm-delete.tsx:11`); and `/knowledge-bases/<id>/settings` still parses and still renders the old view (`routes.ts:101`, `app.tsx:114`), so the gated Submit stays reachable by URL.

On the thread: #3411 is "Share a knowledge base as a read-only MCP endpoint" and does not retire `KnowledgeBaseConfigView`, so the reroute is not yet covered by a landing change.

**Fix:** Retire a view with its route, its rail entry and its subject-specific copy in one change — or leave the destination and render the provider card on the knowledge-base view.

packages/ui/src/modules/sandboxes/hooks/use-section-summaries.ts
```diff
@@ -71,6 +73,13 @@ export function useSectionSummaries(agent: AgentView | null): {
     [apps],
   );
 
+  const missingProvider = useMemo(() => {
```
dam-code-guardian, line 76: 🟡 **Warning:** `appsPending` is false when the query *errored*, and `useAppConnections` defaults `data` to `[]`. So a failed connections read gives an empty `providerAppIds`, no saved connection matches, and the nav dot claims "No provider connected — the agent cannot reach a model" for an agent that has one. This is the same class the provider card just fixed by moving to `isSuccess`.

**Fix:** Derive an absence claim from `isSuccess`, never from `!isPending` — an errored query is also not pending. Treat "not loaded" like pending and show no warning.

```suggestion
  const { data: apps = [], isSuccess: appsLoaded } = useAppConnections();
```

packages/ui/src/modules/providers/components/provider-select.tsx
```diff
@@ -93,6 +111,7 @@ export function ProviderSelect({
         placeholder="Select a provider"
         ariaLabel="Provider"
         disabled={disabled}
+        invalid={required && isSuccess && !selectedType && !autoSelectTarget}
```
dam-code-guardian, line 114: 🟡 **Warning:** `autoSelectTarget` is defined whenever any offered provider is connected, and in exactly that case the effect fills the field. So in the three `autoSelectFirst` flows `invalid` can only be true when the user has zero connected providers — the one state selecting cannot fix. The red border and warning icon issue #3376 asked for never appear on an empty-but-fixable field.

**Fix:** A required-field indicator depends on the field's own emptiness, not on whether a default exists.

```suggestion
        invalid={required && isSuccess && !selectedType}
```

packages/ui/src/modules/providers/components/provider-select.tsx
```diff
@@ -41,16 +44,31 @@ export function ProviderSelect({
...
+        : undefined,
+    [autoSelectFirst, rows, itemByType],
+  );
 
   useEffect(() => {
-    if (!autoSelectFirst || selected) return;
-    const first = rows.map((r) => itemByType.get(r.type)).find(Boolean);
-    if (first) onSelect(providerRef(first));
-  }, [autoSelectFirst, selected, itemByType, rows, onSelect]);
+    if (!isSuccess || selectedType) return;
```
dam-code-guardian, line 65: 🟡 **Warning:** Reported by react-ui-engineering. This gate handles pending and success but not error. When `useAppConnections` fails, `isPending` is false so the card renders past line 84, while `isSuccess` stays false forever: the repair effect never runs and `invalid` never turns on. The required field then looks empty and valid while Create stays disabled — the failure mode #3376 exists to remove, reached by a different route.

**Fix:** A readiness gate on a required field handles all three query states. Branch on `isError` and render a load-failure state, so an unreadable connection list is never shown as "no providers".

packages/ui/src/components/ui/rich-select.tsx
```diff
@@ -48,10 +50,13 @@ export function RichSelect<T extends string>({
         <button
           type="button"
           disabled={disabled}
+          aria-invalid={invalid || undefined}
```
dam-code-guardian, line 53: 🟡 **Warning:** Reported by react-ui-engineering. `aria-invalid` is set with nothing to name the reason, and the `WarningFilled` glyph at line 85 is an unlabelled SVG. A screen-reader user hears "invalid" and gets no cause, while the sighted cue is a red border and an icon. `save-skill-set-modal.tsx:125` already links its message with `aria-describedby`.

**Fix:** An `aria-invalid` control references its reason. Add an optional error-message prop, render it with an id, and point `aria-describedby` at that id.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `7716bb4`

## PR #3433: fix(ui): surface the required provider across create and save flows
**Author:** kapetr | **Branch:** fix/3376-provider-required → main | **Changes:** +63 −18 (10 files)

### Summary
Round 2. The `allow`/`selectedType` mismatch and the failed-read draft clear are both fixed at their cause. Commit `7716bb4` answers the knowledge-base finding by rerouting Configure to the sandbox setup page, which moves the gap rather than closing it.

### Changes since last review
Previous HEAD: `d2001f1c` (2026-08-24T11:15:41Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** `selectedType` is now filtered against the offered rows, so an off-policy value reads as empty and invalid (`provider-select.tsx:51`)
- ✅ **Fixed:** the repair effect runs on `isSuccess`, so a failed connections read no longer clears a staged provider (`provider-select.tsx:65`)
- 🔁 **Still present:** knowledge-base settings gates Submit on a provider it renders no control for (`use-sandbox-settings-form.ts:158`)
- 🆕 **New:** the reroute sends a knowledge base to the agent view, which drops its rail context and its delete copy (`navigation.ts:103`)
- 🆕 **New:** `missingProvider` reads `!appsPending`, so a failed query claims "No provider connected" (`use-section-summaries.ts:76`)
- 🆕 **New:** `invalid` also requires `!autoSelectTarget`, so the error state cannot appear in the auto-select flows (`provider-select.tsx:114`)

### Findings
- 🟡 **Warning:** Rerouting Configure loses knowledge-base context (`packages/ui/src/modules/platform/store/navigation.ts:103`)
- 🟡 **Warning:** A failed connections query reads as "no provider connected" (`packages/ui/src/modules/sandboxes/hooks/use-section-summaries.ts:76`)
- 🟡 **Warning:** The `!autoSelectTarget` term makes the error state unreachable in the create flows (`packages/ui/src/modules/providers/components/provider-select.tsx:114`)
- 🟢 **Suggestion:** The repair effect writes with `shouldDirty: true`, so opening an agent whose provider connection was deleted arms the `beforeunload` prompt with no user action (`packages/ui/src/modules/sandboxes/hooks/use-provider-staging.ts:34`). The provider card is rendered on that surface, so the user can clear it by choosing a provider; a `shouldDirty: false` repair write would remove the prompt.

- 🟡 **Warning:** The knowledge-bases page still says the rail keeps the Knowledge Bases context, and that leaving a knowledge base returns to the KB list. `navigateToKnowledgeBaseConfig` now opens `sandbox-home` with the `setup` section (`docs/architecture/knowledge-bases.md:26`; code at `packages/ui/src/modules/platform/store/navigation.ts:99`). Same cause as the reroute finding above.
  **Fix:** When a diff changes the destination of a subsystem's navigation action, update that subsystem's architecture page in the same PR.

### Issue Fit
Linked issue: #3376 (closing keyword). The three creation flows and the sandbox setup card meet the acceptance list; the knowledge-base settings route does not.

- The knowledge-base gap is the 🔁 carryover above — same criterion, "Users can always see which field is preventing creation".
- The undeclared reroute is the 🆕 finding above. #3411 is about sharing a knowledge base as a read-only MCP endpoint; it does not retire `KnowledgeBaseConfigView`.

- The `appsPending` finding and the rail-predicate half of the reroute finding are the 🟡 entries above; typescript-engineering and the diff review both reported them.
- 🔁 2 findings from the previous review still present (see review at `d2001f1c`).

- The `!autoSelectTarget` and `shouldDirty` findings are above; react-ui-engineering and the diff review both reported them.
- 🟡 **Warning:** `aria-invalid` has no accessible error text, and the `WarningFilled` glyph is an unlabelled SVG, so a screen-reader user hears "invalid" with no reason (`packages/ui/src/components/ui/rich-select.tsx:53`). `save-skill-set-modal.tsx:125` already links a message with `aria-describedby`.
  **Fix:** An `aria-invalid` control references its reason. Render an id'd error message and point `aria-describedby` at it, on every field that can go invalid.
- 🟡 **Warning:** The `isSuccess` gate ignores `isError`. A failed `useAppConnections` renders the select with `isPending` false and `isSuccess` false forever, so the repair never runs and `invalid` never turns on (`packages/ui/src/modules/providers/components/provider-select.tsx:65`).
  **Fix:** A readiness gate on a required field handles all three query states. Branch on `isError` and show a load-failure state.
- 🟢 **Suggestion:** After a refetch that loses the selected connection id, the effect substitutes `autoSelectTarget` — a different provider — with no notice (`provider-select.tsx:64`). Clearing the value and marking the field invalid would keep the choice explicit.

### Verdict
REQUEST_CHANGES — the knowledge-base Submit still has no on-screen reason, and the reroute added to close it drops that view's rail context and its knowledge-base delete copy.

---

dam-code-guardian (dismissed):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `6f6b3aa`

## PR #3433: fix(ui): surface the required provider across create and save flows
**Author:** kapetr | **Branch:** fix/3376-provider-required → main | **Changes:** +63 −18 (10 files)

### Summary
Round 3, complete re-review at the label's request. Commit `6f6b3aa` fixes the `missingProvider` false positive on a failed query. The knowledge-base settings surface still consumes the new save gate and renders no provider control, so its Submit button stays dead with no on-screen reason.
_(Suppressed 1 finding per PR context: the repair effect re-selecting the first connected provider is declared in the PR body.)_

### Changes since last review
Previous HEAD: `7716bb49` (2026-08-24T12:21:03Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** `missingProvider` now needs a loaded connections list, so a failed query no longer claims "No provider connected" (`use-section-summaries.ts:76`)
- 🔁 **Still present:** knowledge-base settings gates Submit on a provider it renders no control for (`use-sandbox-settings-form.ts:158`)
- 🔁 **Still present:** rerouting Configure loses knowledge-base context (`navigation.ts:103`)
- 🔁 **Still present:** the `!autoSelectTarget` term makes the error state unreachable in the create flows (`provider-select.tsx:114`)
- 🔁 **Still present:** the `isSuccess` gate ignores `isError` (`provider-select.tsx:65`)
- 🔁 **Still present:** `aria-invalid` has no accessible error text (`rich-select.tsx:53`)

### Findings
- 🔴 **Critical:** `KnowledgeBaseConfigView` reads `isSubmitDisabled` from the shared form hook but renders no `ProviderSelect`, so a knowledge base without a provider shows a permanently disabled "Submit changes" and no reason. The reroute in `navigation.ts` removed both in-app links to this view, but the `/knowledge-bases/<id>/settings` route still parses and `app.tsx` still renders it, so a bookmark or an OAuth return reaches the dead surface (`packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-form.ts:158`, surfaced at `packages/ui/src/modules/knowledge-bases/views/knowledge-base-config-view.tsx:117`). Reported by the diff review and issue-fit.
  **Fix:** Ship the gate and its error affordance together — every surface that consumes `isSubmitDisabled` renders `ProviderSelect` with `required`, or the route and view are removed so the surface is unreachable.
- 🟡 **Warning:** `navigateToKnowledgeBaseConfig` now pushes the sandbox-home setup route, so a knowledge base opens on the agent surface. The rail loses its Knowledge Bases highlight, and the view's name field and "Delete knowledge base" action become unreachable from the app (`packages/ui/src/modules/platform/store/navigation.ts:103`). Reported by the diff review, doc-drift, issue-fit and typescript-engineering.
  **Fix:** Do not repoint one subsystem's navigation at another's surface to satisfy a validation rule. Either restore the knowledge-base route and give that view the provider control, or retire the view deliberately — move its name and delete actions first, drop the now-duplicate store action and the dead route, and update `docs/architecture/knowledge-bases.md`.
- 🟡 **Warning:** `invalid` requires `!autoSelectTarget`, but the create buttons block on the selection being null. In a render where no provider is selected and an auto-select candidate exists, the button is disabled and the field shows no error (`packages/ui/src/modules/providers/components/provider-select.tsx:114`). Reported by the diff review and react-ui-engineering.
  **Fix:** Derive the error state from the same condition the submit gate uses — a missing selection — and never add a suppression term the gate does not share.
- 🟡 **Warning:** The repair effect and the `invalid` term both require `isSuccess`, so a failed connections query does neither. The select falls back to the placeholder while the parent still holds the old reference, so Create stays enabled over a field that looks empty (`packages/ui/src/modules/providers/components/provider-select.tsx:65`).
  **Fix:** Handle the error state explicitly for every query this component branches on — on `isError`, keep the value and show the error state, so the control and the submit gate never disagree.
- 🟡 **Warning:** The invalid state sets `aria-invalid` and paints a red border, but the `WarningFilled` glyph is unlabelled, so assistive technology announces "invalid" with no reason (`packages/ui/src/components/ui/rich-select.tsx:53`). Reported by react-ui-engineering.
  **Fix:** Whenever a control sets `aria-invalid`, expose the reason as text and point `aria-describedby` at it, and mark the icon `aria-hidden`. Apply to every `invalid` surface of `RichSelect`.
- 🟢 **Suggestion:** The repair effect clears the provider through `setValue(..., { shouldDirty: true })`, so opening settings for an agent whose provider connection was deleted arms the unsaved-changes prompt with no user action (`packages/ui/src/modules/sandboxes/hooks/use-provider-staging.ts:34`).

- 🟡 **Warning:** `docs/architecture/knowledge-bases.md:26` still says the rail keeps the Knowledge Bases context and leaving returns to the KB list. The reroute breaks that for the configure path. Same cause as the reroute finding above.
  **Fix:** When a change moves a documented surface onto another subsystem's surface, update that subsystem's architecture page in the same PR, and bump its `Last verified:` header.

### Issue Fit
Linked issue: #3376 (closing keyword). The three creation flows and the sandbox setup card meet the acceptance list.

- The knowledge-base settings gap is the 🔴 above — it fails the criterion "users can always see which field is preventing creation".
- 🟡 The navigation change is undeclared scope creep: issue #3376 asks for no route change, and the PR body does not mention one. Mirrored as the reroute 🟡 above.
  **Fix:** Declare every user-visible change in the PR body, or move it to its own PR.

- The reroute finding above carries this section's duplicate-navigation half: `navigateToKnowledgeBaseConfig` now produces exactly the route `navigateToSandboxHome(agentId, "setup")` already produces, so two store actions name one destination and the `knowledge-base-config` view is orphaned (`packages/ui/src/modules/platform/store/navigation.ts:99`).
- 🟢 **Suggestion:** The "an agent must have a provider" rule lives only in the client — neither the contract schema nor the save service rejects a provider-less selection, so a non-UI caller bypasses it (`packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-form.ts:158`). Pre-existing and outside this PR's UI scope, so noted rather than blocking, unchanged from the first review.

- The `aria-invalid` finding and the `autoSelectTarget` mismatch are the 🟡 entries above; the diff review and this skill both reported them.
- 🟢 **Suggestion:** `required={f.formReady}` feeds a load-readiness flag into a validation prop, while the other three call sites pass a bare `required` (`packages/ui/src/modules/sandboxes/components/sandbox-setup-section.tsx:89`). The provider is mandatory unconditionally, and `ProviderSelect` already suppresses the error state on its own query state, so the effect is naming, not behavior.

### Verdict
REQUEST_CHANGES — the knowledge-base settings Submit is still dead with no on-screen reason, and the reroute added to close it orphans that view instead of fixing it.

---

tomkis (approved):

Issue #3269, opened 2026-08-11
DAMalytics update automation


jjeliga:
The DAM usage report at https://airtable.com/appmz9L0oqqELNxT1/pagEBL10YK2OqGYhs is being updated manually once a week, which is error prone and may lead to us acting on (slightly) stale data.

## The Goal:
- [x]  Create an automated pipeline that will update the data on a regular basis

jjeliga:
The automation is live. But there will be a switch to job-side table computation on next release, leaving the issue open until then

PR #3434, opened 2026-08-24
chore(usage): preserve analytics grants across view migrations

jjeliga:
## Problem

Privileges in Postgres attach to the object, not the name, so recreating a view discards every grant on it — and a `usage_src_*` passthrough must be recreated rather than replaced whenever a column is renamed or reordered, which is exactly the migration that changes what the analytics consumer reads. Granting that consumer's login directly meant it lost access on the deploys it most needed to survive, silently until a nightly export failed, and an operator re-granted by hand after every release.

## Approach

- **`usage_readers`** — a `NOLOGIN`, password-less **group role**, created by the chart where the chart manages Postgres and by an operator where it does not. It is not a connection identity and confers nothing until someone is added to it.
- **An operator grants membership** to whichever login should read the metrics. Membership binds login to group rather than to any view, so no migration can revoke it, and the platform never learns who the member is.
- **The api-server reconciles the grants** after running migrations, on each start, and logs what the role can actually reach.

The passthroughs only, never the aggregate views: a consumer able to read an aggregate would eventually key a metric on one, and renaming that aggregate would break it — the coupling this surface exists to remove.

## Why a reconcile, not a grant inside the migration

The branch tried the migration form at review's request and it could not be made correct. Recording it so it is not re-litigated.

What must hold is a **state** — if the role exists it can read every passthrough. Three unrelated things break it: a passthrough is added, a passthrough is recreated, or **the role is created only after the views exist**. The actors are migrations, the chart, and an operator, and none is ordered against the others:

- **Bundled Postgres, upgrade.** The role is created by the `post-install,post-upgrade` roles Job. Helm applies normal resources — including the api-server Deployment — *before* post-\* hooks, so the api-server migrates before the role exists and a guarded grant skips permanently. Hook weights don't help: they order hooks within a phase, not against normal resources.
- **External or managed Postgres.** The chart renders no Postgres templates at all, so correctness rested on an operator reading a runbook in the right order.

A reconcile has no ordering constraint. The **gate is deleted** with the migration form: it existed only because a one-shot grant must be repeated by every future migration, and the reconcile runs immediately after migrations in the same startup, so a view recreated by a migration is re-granted milliseconds later. Nothing to forget, nothing to enforce.

## It cannot stop the api-server starting

This is the load-bearing property, and it is structural rather than case-by-case. The reconcile does DCL against a catalog that migrations and operators mutate concurrently; every such interaction is a potential boot failure. So every failure is caught and reported, and migrations stay fatal. A privilege convenience for an optional consumer must never take the platform down.

Bounded, too: `lock_timeout` and `statement_timeout` are set inside the transaction, so a stalled advisory-lock holder or a hung statement degrades instead of holding every replica at startup before the HTTP listener exists.

## Observability

| Log event | Meaning |
|---|---|
| `usage.grants.reconciled` | role present; lists what it can read |
| `usage.grants.role-absent` | no analytics consumer on this install |
| `usage.grants.skipped` | another replica is doing the pass |
| `usage.grants.incomplete` | **warn** — a view it can grant and still cannot read |
| `usage.grants.not-grantable` | **warn** — a view it does not own; needs an operator, no restart fixes it |
| `usage.grants.unreachable` | **warn** — no `CONNECT` on the database or `USAGE` on the schema |
| `usage.grants.failed` | **warn** — reconcile could not complete; platform runs, grants stale |

The readable set is **read back from the catalog after the grants**, not echoed from intent, and includes whether the role can reach the database and schema at all — a privilege it cannot use is not access. The two "cannot read" cases are named separately so the one a restart heals is never confused with the one that needs a human.

## What adversarial review proved wrong

Four independent reviewers ran against a real Postgres. They falsified **four of the eight guarantees this PR previously claimed**; all are fixed and re-verified.

- **"It can never grant a table."** The `GRANT` was quoted but not **schema-qualified**, so it resolved through `search_path` while the SELECT filtered `schemaname='public'`. With a schema named after the connecting role — standard hardening — a shadowing **table** received the grant and the view did not. The same bug aborted boot with `42501` when that object had another owner, and `42P01` whenever `public` was off `search_path`.
- **"Concurrent replicas cannot fail."** True only among reconcilers. Against concurrent *migrations* it deadlocked (`40P01`) and raced a dropped view (`42P01`).
- **"Only ever grants, so it cannot undo a deliberate decision."** Backwards — a *revoke* is the only privilege decision an operator can make here, and every restart reversed it. The reconcile is authoritative by design; the docs now say so and name dropping the member as the way to withdraw access.
- **"Reports what the role can really read."** It reported one ACL bit; a role with no `CONNECT` read as fully readable — the exact failure the alarm exists to catch.

Also fixed: materialized views were invisible to both halves, so a matview passthrough reported clean while ungranted; ownership was too strict, skipping views grantable through a group and warning forever; the `LIKE` escape depended on `standard_conforming_strings`, under which `usageXsrcXpwn` was granted; and a role renamed mid-reconcile crashed the boot.

## Location and tests

The reconcile lives in the usage module, where the bounded context is. **`packages/db` is byte-identical to `main`** — no migration, no journal change, no `runMigrations` signature change. That move is also what makes it testable at all, since `packages/db` has no test path; it now has unit tests covering containment, schema qualification, the grant/skip split, the timeout form, and the reporting split.

## Chart

`usage_readers` is converged unconditionally like the app roles, and covered by the regression guard. Create-if-absent alone let an `ALTER ROLE usage_readers WITH LOGIN SUPERUSER PASSWORD …` survive every later run while three docs called it credential-less. Verified: tampered to `canlogin/super/password = true`, one re-run restores `false/false/false`, and the guard raises if left tampered.

## Multi-replica

Multi-replica api-server is on the roadmap, and the startup database path did not survive it. Two commits here fix that, and both are verified on a live cluster.

**Migrations were not safe to run concurrently at all.** Drizzle's migrator takes no lock: it reads the newest applied migration, then applies everything newer. Started together against a database with migrations pending, **2 of 3 replicas crashed, and 4 of 5** — before any migration ran, on the migrator's own `CREATE ... IF NOT EXISTS` bookkeeping, which is not atomic. Serializing on an advisory lock fixes it: 0 of 3, 0 of 5, 0 of 8.

This is **pre-existing and not specific to usage** — it is the whole platform's schema. It surfaced while checking what multi-replica does to the reconcile that runs immediately after. If the multi-replica work would rather own it, `acf11413` touches only `packages/db/src/migrate.ts` and lifts out cleanly.

**The reconcile was safe but shaped for one replica.** Every replica did the identical pass, serialized on a *blocking* lock, in front of readiness. The lock is now taken without waiting, and a replica that loses skips the whole pass — including the verification read, because mid-pass the state is half-applied and a loser that measured it would report a false `usage.grants.incomplete`. That false alarm is why the lock blocked originally; skipping removes the reason.

Verified on a three-replica cluster booting against an empty schema — identical start times, **zero restarts**, 34 migrations recorded once with no duplicates, 32 usage views, all 7 passthroughs granted:

```
11:57:12.608  lh4nx   usage.grants.skipped
11:57:12.615  xskck   usage.grants.skipped
11:57:12.630  rjdrk   usage.grants.reconciled  granted:[usage_src_activity_events, ...]
```

One replica did the work, two skipped, inside a 22ms window. Both decisions are recorded in ADR-085.

## Verification

Against Postgres 18, driving the shipped module:

| Case | Result |
|---|---|
| shadowing **table** `app.usage_src_events` under `search_path` | not granted; the real view **is** |
| `search_path` without `public` | survives, no failure |
| stalled advisory-lock holder | degrades at the 10s timeout, boot proceeds |
| 8 concurrent reconciles | clean |
| steady state (nothing missing) | zero grants issued, zero catalog writes |
| matview passthrough | granted and reported |
| view owned by a group this role belongs to | granted |
| view owned by an unrelated role | reported `notGrantable`, boot survives |
| aggregates, base tables, `usage_srcfoo`, `usagexsrcxpwn` | stay private |
| role absent | clean no-op |

`db:check`, `api-server:check`, `docs:check`, `helm:check`, `common:check`, comment-types, strip-comments and doc-size all pass. Full suite green: api-server 1177, ui 330, agent-runtime 253, cli 154, driver-sdk 11.

## Docs

Restored the two durable facts an earlier cap-driven trim deleted — why the bootstrap superuser cannot be renamed, and that the server-wide `log_statement` default is DDL-only, which is newly load-bearing because this code issues `GRANT`s. Rather than trim meaning a third time to fit the 40000-char cap, the `log_statement` mechanics moved to the page that owns Postgres operation. Rejected-alternative history is out of the architecture page.

`security-and-credentials.md` has been within 79 characters of its cap since before this PR and wants a proper split; flagged as a follow-up rather than done here.

## Operator note

`usage_readers` can be created at any time, but a **restart is what applies it** — nothing in a release forces an api-server roll, so create the role before the release or `kubectl rollout restart` afterwards. A restart is cheap and safe: migrations are journal-gated, so a start with no new migration files applies no DDL (measured: 17ms versus 290ms when applying). To withdraw access, `REVOKE usage_readers FROM <login>` — revoking a view grant directly does not hold.

Refs #3269


packages/db/src/migrate.ts
```diff
@@ -12,5 +13,6 @@ export async function runMigrations(
   const sql = postgres(url, ssl ? { max: 1, ssl } : { max: 1 });
   const db = drizzle(sql);
   await migrate(db, { migrationsFolder });
+  await grantUsageViews(sql);
```
dam-code-guardian: 🟢 **Suggestion:** A throw from `grantUsageViews` skips `await sql.end()` on the line below, since the two calls sit in sequence with no `try`/`finally`. The owner filter makes a throw unlikely and an aborted boot ends the process anyway, so the leaked connection is bounded — worth a `finally` only if this function ever gains a caller that survives the failure.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `d841e15`

## PR #3434: chore(usage): preserve analytics grants across view migrations
**Author:** jjeliga | **Branch:** chore/3269-create-usage-viewers-pg-group-and-manage-view-access → main | **Changes:** +85 −7 (6 files)

### Summary
The PR replaces per-view grants to an analytics login with a credential-less `usage_readers` group role, re-granted `SELECT` on the `usage_src_*` passthroughs after every migration run. The mechanism holds; the defects are documentation surfaces that still declare the old role set.

### Findings
- 🟡 **Warning:** The authoritative role page still says "Three roles, not one" and that `CONNECT` is granted "only to the owning role" (`docs/architecture/security-and-credentials.md:381`, `:385`). After this PR there are four roles, and `usage_readers` also holds `CONNECT` on `platform`. This page defines the database credential boundary, so both statements are false where a reader goes to learn that boundary. Reported by doc-drift and confirmed independently.
  **Fix:** When a change alters the platform database role set or its `CONNECT` grants, update this section in the same PR and bump `Last verified:`. The page measures **exactly 40000 of 40000** characters in the gate's own unit (`scripts/doc-size.mjs` `measure()` is `content.length`, passing on `size <= cap`), so there is zero headroom — add the role at the level of its trust boundary and cut volatile detail to make room, rather than rewording to fit.
- 🟡 **Warning:** The operations runbook enumerates the role set twice and this PR updates neither (`docs/notes/postgres-role-operations.md:9`, `:99`). Its "External / managed Postgres" section is the reproduction checklist an operator follows on a managed instance, so it now omits `usage_readers`. On those installs the chart never runs, and the group exists nowhere.
  **Fix:** Apply the role-set rule to every page that *enumerates* the roles, not only the pages the diff touched — and where a page tells an operator to reproduce the layout by hand, add the group and the membership grant to that list.
- 🟢 **Suggestion:** A failing `grantUsageViews` skips `sql.end()` (`packages/db/src/migrate.ts:16`) — see the inline comment.
- ✅ **Looks good:** The two safety properties the body asks about hold. `pg_views` cannot list a table, and `viewowner = current_user` keeps the grant to what the role owns, so an alien-owned `usage_*` view cannot crash-loop the api-server.

- 🟡 **Warning:** The PR adds a fourth Postgres role and grants it `CONNECT` on the `platform` database (`deploy/helm/platform/templates/postgres.yaml:57`, `:86`), but the authoritative role page still declares "Three roles, not one" and says `CONNECT` goes "only to the owning role" (`docs/architecture/security-and-credentials.md:381`, `:385`). Merged into `### Findings` above.
- 🟢 **Suggestion:** The startup database sequence now runs a privilege re-grant after the migrations, outside the migration history (`packages/db/src/migrate.ts:16`), but the page that owns the startup description still lists only schema and view migrations (`docs/architecture/persistence.md:68`). Possible drift — a human decides whether the new non-schema boot step belongs on this page as a one-liner, or stays owned by the usage-tracking page it already cross-links.

### Issue Fit
Linked issue: #3269 "DAMalytics update automation" (inferred from the plain reference `Refs #3269` and the head branch; no closing keyword, so the issue stays open).

✅ Delivers a declared platform-side step of the issue's ask; no undeclared scope.

✅ No findings.

### Verdict
**COMMENT** — the mechanism is verified correct; two documentation surfaces still declare the superseded role set.

---

packages/db/src/usage-view-grants.ts
```diff
@@ -0,0 +1,50 @@
...
+// down with it; migrations create every view as this role, so the two sets are
+// the same in practice. Each `\_` is LIKE's escape for a literal underscore,
+// so the pattern cannot match `usageXsrcXfoo`. Views are granted in name order
+// so concurrently booting replicas take catalog locks in the same order and
+// cannot deadlock.
+//
+// A no-op on every install that has not created the role, which is most.
+export async function grantUsageViews(sql: postgres.Sql): Promise<void> {
+  await sql.unsafe(`
+    DO $$
```
pilartomas: Why is this preferred over GRANT in migration?

packages/db/src/usage-view-grants.ts
```diff
@@ -0,0 +1,50 @@
...
+// to the `usage_readers` group role, on every api-server boot.
+//
+// Postgres attaches privileges to the object, not the name, so recreating a
+// view destroys every grant on it. View migrations must recreate rather than
+// replace whenever a column is renamed or reordered, so an optional read-only
+// consumer loses access on exactly the deploys that change what it reads.
+// Healing here rather than in each migration means there is no per-migration
+// step to forget, and a newly added passthrough is covered the moment it
+// exists. It lives in its own module because it is a privilege operation, not
+// schema evolution: nothing about it belongs in the migration history.
```
pilartomas: I'd be vary of this approach, logic like this severely complicates future rollouts. Migrations are a better deal imo.

pilartomas: Also, note that the DB user used by the api server is not a privileged user.

Not sure about migrations, but there we might consider bumping it up, at least.

jjeliga: Fair — I have moved it. The grant now lives in migration `0034`, and the boot-time step is gone.

The one thing the boot-time version was actually buying was that nobody has to remember the step: recreating a view discards its privileges, so *every* future migration that touches a `usage_src_*` view has to re-grant it, and a miss fails silently — the consumer just stops reading and nobody finds out until a nightly export fails. That is now mechanical instead of a boot-time side effect: `mise run db:check:usage-src-grants` fails any migration that creates or recreates a passthrough without re-granting it in the same file.

It is deliberately per-file. A whole-history check would pass when an old migration granted the view and a new one recreated it, which is exactly the case worth catching.

Two details worth knowing about the migration form, both of which cost us something:

- **The grants have to be guarded on the role existing.** `usage_readers` is absent on any install with no analytics consumer, and on external or managed Postgres the chart never runs at all, so it is absent there too. An unguarded `GRANT` gives `ERROR: role "usage_readers" does not exist` and aborts the migration — crash-looping the api-server on exactly those installs. So the conditional does not disappear by moving into a migration, it just moves.
- **A guarded grant skips silently, so ordering now matters.** The role must exist *before* the release whose migrations grant to it, or the views stay ungranted until an operator grants them by hand. The boot-time version had no such constraint — any later restart fixed it. This is a real regression from moving into migrations, and it is now called out in the Postgres role runbook so whoever sets up prod sees it.

Migrations still felt like the right call given the auditability, but I did not want to hide the tradeoff.

packages/db/src/usage-view-grants.ts
jjeliga: It is not preferred any more — moved into a migration (`0034`), so this is now the shape you were asking for. Details in the thread above.

Short version of why it started at boot: the grant has to be repeated by every future migration that recreates a `usage_src_*` view, and forgetting it fails silently. That is now enforced by a build gate (`db:check:usage-src-grants`) instead of by a boot-time re-grant, which gets the auditability of migrations without relying on anyone remembering.

packages/db/src/usage-view-grants.ts
jjeliga: It does not need to be, and I would push back on bumping it — the unprivileged role is sufficient here, and it stays sufficient now that the grant has moved into a migration.

In Postgres an object's **owner** can always `GRANT` on it; that needs no role attribute. The migrations create these views as the api-server's role, so it owns them and can grant on them. I checked rather than assuming — a role created as plain `LOGIN NOSUPERUSER` with no `CREATEROLE`:

```
CREATE ROLE plain LOGIN NOSUPERUSER PASSWORD 'p';
ALTER VIEW usage_src_demo OWNER TO plain;
-- as plain:
GRANT SELECT ON usage_src_demo TO usage_readers;   --> GRANT
```

The one thing that role genuinely cannot do is `CREATE ROLE`, which is why `usage_readers` is created by the chart's bootstrap superuser (or by the provider admin on a managed instance), not by the api-server. That split is deliberate rather than incidental: an api-server that could mint database logins could mint itself a better one, and `security-and-credentials` states the app role has no `CREATE ROLE` as part of the credential boundary. Bumping it would trade a real containment property for something we do not need.

Worth adding that this was orthogonal to the boot-versus-migration question — migrations run as that same unprivileged role, so the privilege situation is identical either way.

packages/db/src/migrate.ts
jjeliga: Moot now — `grantUsageViews` is gone, since the grant moved into migration `0034`.

The `try`/`finally` stayed anyway: `migrate()` had the same gap before this branch, so `runMigrations` now closes the connection on either path.

jjeliga:
All review findings addressed, plus a mechanism change from the discussion threads. Rebased on `main` (which had landed its own `0033`, so the migration renumbered to `0034`).

**Mechanism moved into migrations.** The grant no longer runs at api-server boot; `0034` grants the seven passthroughs, and from here on every migration that creates or recreates a `usage_src_*` view re-grants it in the same file. The forgettability that motivated the boot-time version is now handled by `db:check:usage-src-grants`, which fails the build on a miss — per-file, since a whole-history check would pass exactly when an old migration granted the view and a new one recreated it. Two tradeoffs are called out in the threads: the grants must be guarded on the role existing, and a guarded grant skips silently, so `usage_readers` has to exist *before* the release whose migrations grant to it.

**Both doc-drift warnings were correct.**

- `security-and-credentials.md` — it declared "three roles, not one" and that `CONNECT` goes only to the owning role, both false after this change, on the page that defines the credential boundary. The page sat at exactly 40000 of 40000 characters, so room was made by cutting volatile detail rather than rewording to fit: a parenthetical restating the sentence above it, log plumbing `logging.md` owns, a Helm values key, and an enumeration the linked runbook already carries.
- `docs/notes/postgres-role-operations.md` — the more valuable catch. Its managed-Postgres section is a by-hand reproduction checklist, and prod *is* a managed instance where the chart never runs, so the group existed nowhere in the instructions. The checklist now carries the role, the membership grant, and the ordering requirement.
- `persistence.md` — the startup description was updated, then updated again when the mechanism moved; it now describes the grant guard alongside the generated-schema guard rather than a boot step.

**The `sql.end()` suggestion** is moot since `grantUsageViews` is gone, but the `try`/`finally` stayed — `migrate()` had the same gap before this branch.

**One thing the review did not catch, worth flagging.** The rationale comment was a run of `//` lines, and `check-comment-types` requires the prefix on every comment node, so all but the first line were untyped and `strip-comments` would have deleted them. Fixed at the time by using a single block comment; that file has since been deleted, and the new gate script carries no prose comments at all, matching the rest of `scripts/`. Its rationale lives in `packages/db/README.md`.

Full suite green on the rebase: api-server 1135, ui 314, agent-runtime 240, cli 154, driver-sdk 11, plus `db:check` (including the new gate), `docs:check`, `helm:check`, `common:check`.

scripts/check-usage-src-grants.mjs
```diff
@@ -0,0 +1,68 @@
...
+
+const REPO_ROOT = resolve(dirname(fileURLToPath(import.meta.url)), "..");
+const MIGRATIONS = join(REPO_ROOT, "packages", "db", "drizzle");
+const ROLE = "usage_readers";
+const FIRST_GATED_MIGRATION_INDEX = 35;
+
+const migrationIndex = (file) => Number.parseInt(file.slice(0, 4), 10);
+
+const passthroughViewsCreated = (sql) =>
+  [...sql.matchAll(/CREATE\s+(?:OR\s+REPLACE\s+)?VIEW\s+"?(usage_src_[a-z0-9_]+)"?/gi)].map((m) =>
```
dam-code-guardian: 🔴 **Critical:** This pattern requires the view name to follow `VIEW` directly, while `passthroughViewsGranted` below accepts an optional `public.` prefix. So `CREATE VIEW public.usage_src_x` (and the quoted `"public"."usage_src_x"` form) yields an empty created list and the gate reports success. Verified by running both regexes over five inputs: the plain form fails the gate, both qualified forms pass. `CREATE MATERIALIZED VIEW` and `CREATE RECURSIVE VIEW` are missed by the same pattern.

**Fix:** Recognize a name and its optional schema qualifier once, in a helper both halves call, and allow every modifier Postgres permits between `CREATE` and `VIEW`.

scripts/check-usage-src-grants.mjs
```diff
@@ -0,0 +1,68 @@
...
+    m[1].toLowerCase(),
+  );
+
+const passthroughViewsGranted = (sql) =>
```
dam-code-guardian: 🟡 **Warning:** The scan reads raw file text, so a commented-out grant satisfies it. Confirmed: `-- GRANT SELECT ON usage_src_agents TO usage_readers;` passes the gate for a view that receives no privilege at runtime.

**Fix:** Strip `--` and `/* */` comments before any pattern runs. A gate must never accept evidence the database will not execute.

Related, same function: the gate does not check that the grant sits inside the `pg_roles` guard your error text calls required — an unguarded `GRANT` passes the build and then aborts the migration wherever the role is absent.

scripts/check-usage-src-grants.mjs
```diff
@@ -0,0 +1,68 @@
...
+  new Set(
+    [
+      ...sql.matchAll(
```
dam-code-guardian: 🟡 **Warning:** This binds a single view name between `ON` and `TO`, so the valid `GRANT SELECT ON usage_src_a, usage_src_b TO usage_readers;` matches nothing and fails the build for a migration that is in fact correct. The error then names the grant rather than the spelling.

**Fix:** Parse the comma-separated object list the `GRANT ... ON <list> TO <role>` syntax permits, and reserve failure for a genuinely missing grant.

packages/db/src/migrate.ts
```diff
@@ -11,6 +11,9 @@ export async function runMigrations(
   const ssl = buildDbSsl(tls);
   const sql = postgres(url, ssl ? { max: 1, ssl } : { max: 1 });
   const db = drizzle(sql);
-  await migrate(db, { migrationsFolder });
-  await sql.end();
+  try {
```
dam-code-guardian, line 16: 🟡 **Warning:** This `try`/`finally` is undeclared scope — residue of the reverted boot-time revision, and neither #3269 nor the body asks for it. The change itself is correct, and `migrate()` had the same gap before this branch.

**Fix:** When a mechanism is abandoned mid-PR, diff what remains against the new mechanism and either declare each survivor or drop it.

One level down: `sql.end()` in the `finally` has no handling of its own, so a teardown failure replaces the migration error already propagating. Wrapping the cleanup in its own `try`/`catch` keeps the real cause.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `2d33505`

## PR #3434: chore(usage): preserve analytics grants across view migrations
**Author:** jjeliga | **Branch:** chore/3269-create-usage-viewers-pg-group-and-manage-view-access → main | **Changes:** +3063 −27 (14 files)

### Summary
The grant moved out of api-server boot into migration `0034`, and a file-scan gate now demands a re-grant from every migration that recreates a `usage_src_*` view. Both prior documentation warnings are resolved at their cause. The remaining defects sit in the new gate's own detection and in the bundled-upgrade ordering the body describes as a managed-install concern only.

### Changes since last review
Previous HEAD: `d841e15` (2026-08-24T12:28:06Z) — verdict COMMENT

- ✅ **Fixed:** Role set and `CONNECT` contradiction on the credential page (`docs/architecture/security-and-credentials.md:381`, `:392`).
- ✅ **Fixed:** Runbook role enumeration and the managed-Postgres reproduction checklist both carry `usage_readers` (`docs/notes/postgres-role-operations.md:21`, `:115`).
- ✅ **Fixed:** `sql.end()` now runs on the failure path (`packages/db/src/migrate.ts:16`).
- 🆕 **New:** The gate's create-side regex misses schema-qualified `CREATE VIEW` (`scripts/check-usage-src-grants.mjs:15`).
- 🆕 **New:** The bundled-Postgres upgrade path has the same ordering gap the body scopes to managed installs (`deploy/helm/platform/templates/postgres-migrate-roles.yaml:30`).
- 🆕 **New:** Comment text and unguarded grants both satisfy the gate (`scripts/check-usage-src-grants.mjs:19`).
- 🆕 **New:** Two architecture-page content findings (`docs/architecture/security-and-credentials.md:390`, `docs/architecture/usage-tracking.md:155`).

### Findings
- 🔴 **Critical:** The gate's create-side and grant-side patterns disagree about the same name. `passthroughViewsCreated` requires the view name to follow `VIEW` directly, while `passthroughViewsGranted` accepts an optional `public.` prefix — so `CREATE VIEW public.usage_src_x` is invisible to the gate and reports success (`scripts/check-usage-src-grants.mjs:15`). Reported by typescript-engineering; reproduced by running both regexes over five inputs — the plain form fails the gate, the qualified and quoted-qualified forms pass with an empty created list. `MATERIALIZED` and `RECURSIVE` view forms are missed by the same pattern.
  **Fix:** Recognize a name and its optional schema qualifier once, in a helper both halves call, and allow every modifier Postgres permits between `CREATE` and `VIEW`. Any spelling the grant side accepts, the create side must see.
- 🔴 **Critical:** On a bundled-Postgres **upgrade** the role is created after the migration that grants to it, so the seven passthroughs stay ungranted. `usage_readers` is created only by `01-roles.sql`, whose two callers are first-PGDATA-init and the `post-install,post-upgrade` Job (`deploy/helm/platform/templates/postgres-migrate-roles.yaml:30`). On an existing cluster the init script never re-runs, and the api-server runs `0034` at boot, before that Job — the chart's own note records app pods starting first ("App pods may CrashLoop in the short window between the rollout and this Job finishing", `:17`). The guarded grant then skips silently. The body and the runbook state this ordering constraint for external and managed Postgres only, where an operator reads it; on `postgres.enabled: true` nobody does.
  **Fix:** Create the role before the migrations run — move the role creation into a `pre-upgrade` hook, or give the migrate-roles Job a negative `hook-weight` and let the api-server wait on it. Whenever a migration depends on chart-created database state, the chart step must be ordered ahead of the api-server rollout, not after it.
- 🟡 **Warning:** The gate matches raw file text, so a commented-out grant passes it (`scripts/check-usage-src-grants.mjs:19`) — confirmed: `-- GRANT SELECT ON usage_src_agents TO usage_readers;` satisfies the check for a view that receives no privilege. Reported by typescript-engineering.
  **Fix:** Strip `--` and `/* */` comments before any pattern runs. A gate must never accept evidence the database will not execute.
- 🟡 **Warning:** The gate does not check that the grant sits inside the `pg_roles` existence guard (`scripts/check-usage-src-grants.mjs:19`). Its own error text calls that guard required rather than defensive, so an unguarded `GRANT` passes the build and then aborts the migration on every install where the role is absent. Reported by typescript-engineering.
  **Fix:** Assert both halves of the shape the gate documents — the grant is present, and the guard encloses it.
- 🟡 **Warning:** `GRANT SELECT ON usage_src_a, usage_src_b TO usage_readers` is valid Postgres and matches nothing, so a correct migration fails the build with an error naming the grant rather than the spelling (`scripts/check-usage-src-grants.mjs:22`). Reported by typescript-engineering.
  **Fix:** Parse the comma-separated object list the `GRANT` syntax permits, and reserve failure for a genuinely missing grant.
- 🟡 **Warning:** The credential page absorbed the new paragraph by deleting two durable facts — that the bootstrap superuser can be neither demoted nor renamed while connected as itself, and that the global `log_statement` default is `ddl` (`docs/architecture/security-and-credentials.md:390`). It now sits at 39999 of the 40000-char cap in the gate's own unit. Reported by doc-drift; both facts survive only in the runbook and the ADR, which architecture pages may not depend on per `docs/guidelines/documentation-guidelines.md:21`.
  **Fix:** Buy cap headroom from volatile detail or a split, never from durable operational facts. A self-contained page cannot delegate an invariant to an ADR.
- 🟡 **Warning:** The rewritten passthrough paragraph grew by roughly 1700 characters and now carries rejected-alternative history — that granting the consumer's login directly revoked access on the deploys it needed to survive (`docs/architecture/usage-tracking.md:155`). `documentation-guidelines.md:21` places decision history in the ADR log. Reported by doc-drift.
  **Fix:** Keep the operational why on the page — the coupling and the invariant — and move why-an-alternative-was-rejected to the ADR.
- 🟡 **Warning:** The `try`/`finally` around `migrate()` is undeclared scope: it is residue of the reverted boot-time revision, and neither issue #3269 nor the body asks for it (`packages/db/src/migrate.ts:14`). Reported by issue-fit. It is a correct change on its own, and `migrate()` had the same gap before this branch.
  **Fix:** When a mechanism is abandoned mid-PR, diff what remains against the new mechanism and either declare each survivor or drop it.
- 🟢 **Suggestion:** The mermaid edge still reads `read-only role, SELECT usage_src_*`, so the diagram shows the pre-change mechanism (`docs/architecture/usage-tracking.md:68`).
- 🟢 **Suggestion:** "Membership attaches to the two roles" has no antecedent on that page; naming both roles resolves it (`docs/architecture/usage-tracking.md:155`).
- 🟢 **Suggestion:** `FIRST_GATED_MIGRATION_INDEX` is 35 and the highest migration is `0034`, so the gate scans zero files and always passes. Its success line claims a verified result rather than an empty scan (`scripts/check-usage-src-grants.mjs:10`).
- 🟢 **Suggestion:** `migrationIndex` returns `NaN` for a `.sql` file not starting with four digits, and `NaN < 35` is false, so the file is scanned rather than skipped (`scripts/check-usage-src-grants.mjs:12`).
- 🟢 **Suggestion:** A migration that grants a view and then revokes it in the same file passes the gate (`scripts/check-usage-src-grants.mjs:19`).
- 🟢 **Suggestion:** The body's Approach section says `0033` carries the grant; it ships as `0034` after the rebase.
- 🟢 **Suggestion:** The cap-trimming list in the body does not mention that `platform-postgres-secrets` lost its concrete name in the same edit (`docs/architecture/security-and-credentials.md:402`).

- 🟡 Cap-driven deletion of two durable facts (`docs/architecture/security-and-credentials.md:390`) — merged into Findings.
- 🟡 Rejected-alternative history on an architecture page (`docs/architecture/usage-tracking.md:155`) — merged into Findings.
- 🟢 Mermaid edge still names the old per-login role (`docs/architecture/usage-tracking.md:68`).
- 🟢 "the two roles" has no antecedent on that page (`docs/architecture/usage-tracking.md:155`).

### Issue Fit
Linked issue: #3269 "DAMalytics update automation" (`Refs #3269`, no closing keyword). The issue is an umbrella ask whose pipeline half lives outside this repo, and this PR delivers the declared platform-side step. No missing scope.
- 🟡 Undeclared `try`/`finally` in `runMigrations` (`packages/db/src/migrate.ts:14`) — merged into Findings.
- 🟢 Body names `0033` as the granting migration; it is `0034`.
- 🟢 The stated cap-trimming list omits the dropped Secret name (`docs/architecture/security-and-credentials.md:402`).

- 🔴 Create-side and grant-side name patterns disagree (`scripts/check-usage-src-grants.mjs:15`) — merged into Findings.
- 🟡 `MATERIALIZED` / `RECURSIVE` view forms invisible to the gate (`scripts/check-usage-src-grants.mjs:15`) — merged into the critical above.
- 🟡 Comment text satisfies the gate (`scripts/check-usage-src-grants.mjs:19`) — merged into Findings.
- 🟡 The guard around the grant is not asserted (`scripts/check-usage-src-grants.mjs:19`) — merged into Findings.
- 🟡 Multi-object `GRANT` form fails a correct migration (`scripts/check-usage-src-grants.mjs:22`) — merged into Findings.
- 🟡 A teardown failure in `finally` can displace the migration error (`packages/db/src/migrate.ts:16`).
  **Fix:** Wrap a cleanup call in `finally` in its own `try`/`catch`, so cleanup runs without replacing the original failure.
- 🟢 `migrationIndex` `NaN` path (`scripts/check-usage-src-grants.mjs:12`) — merged into Findings.
- 🟢 Gate currently scans zero files (`scripts/check-usage-src-grants.mjs:10`) — merged into Findings.
- 🟢 Grant-then-revoke in one file passes (`scripts/check-usage-src-grants.mjs:19`) — merged into Findings.

### Verdict
REQUEST_CHANGES — the gate's create-side detection misses a valid `CREATE VIEW` spelling its own grant side accepts, and on a bundled-Postgres upgrade the role is created after the migration that grants to it.

---

packages/db/src/usage-view-grants.ts
jjeliga: Coming back on this one, because I moved it to a migration as you asked and then found the migration form cannot be made correct here. Reverting to a reconcile, with the reasoning laid out so it is not just me changing my mind twice.

**The invariant is a state, not an event.** What has to hold is: if `usage_readers` exists, it can read every `usage_src_*` view. Three unrelated things break it — a passthrough is added, a passthrough is recreated, or **the role is created only after the views already exist**. The actors are migrations, the chart, and an operator, and none of them is ordered against the others. A one-shot migration serves the first two and only in some orders.

**Concretely, the migration form was broken on both install paths:**

- *Bundled Postgres, upgrade.* `usage_readers` is created by `01-roles.sql`, whose only reachable caller on an existing cluster is the roles Job — a `post-install,post-upgrade` hook. Helm applies the release's normal resources, including the api-server Deployment, **before** post-* hooks run. So the api-server migrates before the role exists, the guarded grant skips, and it never runs again. Silently. Hook weights do not fix this — they order hooks within a phase, not against normal resources, so a negative weight would only move the Job ahead of the Keycloak provision job.
- *External or managed Postgres.* The chart renders no Postgres templates at all, so correctness rested on an operator having read a runbook in the right order.

Unsatisfiable on one path, documentation on the other. The reconcile has no ordering constraint anywhere.

**On the gate.** I built it, and it was not sound. It verified that SQL *text* looked right, never that the database ended up right — and probing found 9 of 10 inputs wrong: it ignored `CREATE VIEW public.usage_src_x`, accepted a commented-out grant, accepted an unguarded grant, and rejected a valid `GRANT ... ON a, b TO ...`. I could fix those, but it would stay a permanent tax on every future view migration to buy back a property the reconcile has for free. It was the price of the mechanism, not a feature, so it is deleted.

**Your actual objection stands and is addressed directly.** A privilege step in application code is invisible in the release artifact. So it now reports itself, every start:

- `usage.grants.reconciled` — which passthroughs the group can read
- `usage.grants.incomplete` — **warning**; a passthrough it still cannot read
- `usage.grants.role-absent` — no analytics consumer here

The list is measured from the catalog *after* the grants, not echoed from intent, so it states what is actually true rather than what the code meant to do. `usage.grants.incomplete` is exactly the silent failure this mechanism exists to prevent, made loud.

**Two things I got wrong in the first version and have fixed regardless of mechanism:** concurrent replicas needed an advisory lock — eight concurrent reconciles fail with `tuple concurrently updated`, which would abort a boot; and it now blocks rather than skipping on contention, since a skipped reconcile would still run the readback and could report a half-granted state as unreadable.

If you still prefer migrations knowing the bundled-upgrade ordering, say so and I will implement it with the chart step reordered to `pre-upgrade` — but that changes upgrade ordering for the whole chart, and I did not want to do that to an analytics grant without you asking for it.

jjeliga:
> [!WARNING]
> **Superseded.** Parts of this comment are no longer accurate — adversarial review found three criticals in the code it describes, and the design changed again afterwards. Left unedited below for history; the correction is further down the thread.

---

Mechanism reverted to a reconcile, and the gate is deleted. Full reasoning is in the PR body and in [the thread with @pilartomas](https://github.com/dam-agents/dam/pull/3434#discussion_r3854841123); addressing the review findings individually here.

**Both Criticals accepted.**

The bundled-Postgres ordering finding was the important one and it is what drove the revert — but the suggested fix does not work. A negative `hook-weight` on a `post-upgrade` hook still runs after the release's normal resources; weights order hooks *within a phase*, so it would only move the Job ahead of the Keycloak provision job. `pre-upgrade` would be needed, and that changes upgrade ordering for the entire chart. Reconciling removes the ordering dependency instead of trying to satisfy it.

The create-side/grant-side disagreement was real and I reproduced all of it. Rather than harden the regexes, the gate is gone: it existed only to enforce a per-migration step the reconcile does not need. For the record, probing the version that was on the branch found **9 of 10 inputs wrong** — schema-qualified and quoted-qualified `CREATE VIEW` ignored, `MATERIALIZED`/`RECURSIVE` ignored, `--` and `/* */` commented grants accepted, unguarded grants accepted, multi-object `GRANT` rejected, and a `CREATE` appearing only in prose failing the build. That last one neither review caught.

**Doc findings accepted.** The credential page: the reason the bootstrap superuser cannot be renamed is restored, as is the `platform-postgres-secrets` name, and the fourth role is now a clause on the existing `CONNECT` sentence rather than a paragraph — which is what the one-page-owns-a-concept rule called for and what should have happened first. The page is at 39928 of 40000, below where it started, so the correction was not paid for by deleting durable facts. Rejected-alternative history is out of `usage-tracking`; the dangling "the two roles" and the stale mermaid edge are fixed.

**Two findings are now void rather than fixed.** The `try`/`finally` is no longer undeclared scope — the mechanism it belongs to is back, so a throw from the reconcile really would leak the connection. The teardown-masking sub-point was correct and is fixed properly: cleanup no longer replaces a propagating migration error. The `0033` vs `0034` naming is moot; this branch adds no migration at all now, and `_journal.json` is byte-identical to main.

**One I am not taking.** `(the global default is `ddl`)` I would leave cut. It is a chart-configured value (`postgres.log.statement`), which the guidelines class as volatile — if an operator sets it to `all` the sentence becomes false — and the operative claim, that routine app traffic stays out of the audit stream, survives without it.

**Two defects found by my own testing, not by review**, both of which would have shipped:

- Eight concurrent reconciles fail with `tuple concurrently updated`. `ORDER BY` prevents deadlock but not concurrent catalog-tuple updates, so a rolling restart could abort a boot. Now serialized on an advisory lock, blocking rather than skipping — a skipped reconcile would still run the readback and could report a half-granted state as unreadable, and that signal has to stay trustworthy. 32 concurrent reconciles now complete clean.
- `\_` inside a **tagged** template literal cooks to a bare `_`, silently dropping the LIKE escape, so the readback query would have matched `usageXsrcXfoo`. Both patterns are asserted to reach Postgres as `usage\_src\_%`.

Verification table is in the body. Full suite green on a rebase onto current main: api-server 1170, ui 330, agent-runtime 253, cli 154, driver-sdk 11.

jjeliga:
Correcting my earlier comment, which I have marked superseded. I ran four independent adversarial reviewers against a real Postgres, and they falsified **four of the eight guarantees I had claimed in this PR**. The design changed again as a result; the body is rewritten to match.

**What I claimed and what was actually true.**

- *"It can never grant a table."* False. The `GRANT` was quoted but never **schema-qualified**, so it resolved through `search_path` while the SELECT filtered `schemaname = 'public'`. With a schema named after the connecting role — routine CVE-2018-1058 hardening — a shadowing **table** received the grant and the view did not. The same bug aborted boot with `42501` when that object had another owner, and `42P01` whenever `public` was off `search_path`, reproduced through the real `runMigrations` against the real migrations folder.
- *"Only ever grants, so it cannot undo a deliberate decision."* Backwards. A revoke is the only privilege decision an operator can make here, and every restart reversed it. It is a reconciler and treats the passthrough set as authoritative; the docs now say so, and name dropping the member as the way to withdraw access.
- *"Concurrent replicas cannot fail."* True only among reconcilers — the advisory lock orders them against each other and does nothing about migrations, against which it deadlocked (`40P01`) and raced a dropped view (`42P01`).
- *"The readback reports what the role can really read."* It reported one ACL bit. A role with no `CONNECT` read as fully readable — precisely the failure the alarm exists to catch, on the managed-Postgres path where the chart never grants `CONNECT`.

Also found: materialized view passthroughs were invisible to both halves, so one reported a clean reconcile while ungranted; `viewowner = current_user` skipped views this role can grant through a group, warning forever with no fix; the `LIKE` escape depended on `standard_conforming_strings`, under which `usageXsrcXpwn` was granted; and a stalled advisory-lock holder hung the boot with no timeout.

**The structural correction.** Every one of those failed the boot. I had built one narrow guard against one failure mode and mistaken it for containment. The reconcile is now caught and reported in full, with `lock_timeout` and `statement_timeout` bounding the work; migrations stay fatal. A privilege convenience for an optional consumer must not be able to take the platform down.

**Two things my own testing caught that no reviewer did**, both of which would have shipped:

- `SET LOCAL` does not accept bind parameters, so `SET LOCAL lock_timeout = $1` failed on the first statement of **every** reconcile. My unit tests passed because the fake recorded SQL text without executing it — only the live Postgres run caught it. `set_config(..., true)` is the parameterizable form.
- The comment gates discover files via `git ls-files`, so untracked files are silently skipped. My new files had never actually been checked; after `git add` the scan went from 0 files to 1647 and immediately flagged a stray comment.

**On the earlier "two findings are now void"** in the superseded comment: the `try`/`finally` discussion is moot — `packages/db` is now byte-identical to `main`. The reconcile moved to the usage module, which is also what made unit tests possible at all.

**Chart, separately.** `usage_readers` was create-if-absent with no convergence, so `ALTER ROLE usage_readers WITH LOGIN SUPERUSER PASSWORD …` survived every later run while the chart comment, `values.yaml` and `persistence.md` all called it credential-less. It is now converged unconditionally like the app roles and covered by the regression guard.

Verification table is in the body. Full suite green on current `main`.

packages/db/src/usage-view-grants.ts
jjeliga: Correction — this thread ends on a stale answer. The migration form did not survive: it is correct only in some orderings and unsatisfiable in others, so there is **no migration and no gate** now. The grant is reconciled after migrations on each start, from the usage module, and `packages/db` is byte-identical to `main`.

Reasoning is in [the other thread](https://github.com/dam-agents/dam/pull/3434#discussion_r3854841123) and the rewritten PR body. Short version: on a bundled-Postgres upgrade the role is created by a `post-upgrade` hook, which Helm runs *after* the api-server Deployment — so a one-shot grant skips permanently — and on managed Postgres the chart creates nothing at all.

packages/db/src/migrate.ts
jjeliga: Correction — this thread also ends stale. There is no migration `0034` any more, and no `try`/`finally` change either: `packages/db` is now byte-identical to `main`, so `runMigrations` is untouched by this PR. The reconcile moved to the usage module and is fully contained there — any failure is caught and reported rather than propagating, so it cannot abort startup.

packages/api-server/src/modules/usage/infrastructure/usage-view-grants-report.ts
```diff
@@ -0,0 +1,49 @@
...
+      {
+        role,
+        canConnect: grants.canConnect,
+        canUseSchema: grants.canUseSchema,
+      },
+      "usage.grants.unreachable",
+    );
+    return;
+  }
+  if (unreadable.length > 0) {
```
dam-code-guardian, line 45: 🟡 **Warning:** `notGrantable` is only reported when `unreadable` is empty — both branches `return`, and `unreadable` is checked first. `unreadable` can be populated transiently (a view committed by another session between the pending query and the catalog re-measure), while `notGrantable` is the standing state that needs an operator, so the transient condition hides the persistent one on every affected boot.

**Fix:** When two independent problem sets can be non-empty at once, report both — emit one line carrying every non-empty set, or drop the early return so each set gets its own named event.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `4e41ef8`

## PR #3434: chore(usage): preserve analytics grants across view migrations
**Author:** jjeliga | **Branch:** chore/3269-create-usage-viewers-pg-group-and-manage-view-access → main | **Changes:** +586 −24 (12 files)

### Summary
The mechanism reverted from a migration-carried grant to a reconcile in the usage module, and both prior criticals are gone with the code they lived in: `packages/db` is byte-identical to `main` and the gate script no longer exists. One defect survives in the new reporting module.

### Changes since last review
Previous HEAD: `2d33505` (2026-08-24T14:21:13Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** Create-side/grant-side name disagreement — the gate is deleted (`scripts/check-usage-src-grants.mjs` absent at HEAD).
- ✅ **Fixed:** Bundled-Postgres upgrade ordering — the reconcile has no ordering dependency on the roles Job. The `pre-upgrade`/`hook-weight` objection to the suggested fix is correct: weights order hooks within a phase, not against normal resources.
- ✅ **Fixed:** Commented-out and unguarded grants accepted by the gate; multi-object `GRANT` rejected — all void with the gate.
- ✅ **Fixed:** Durable facts restored on the credential page, which now measures 39963 of 40000 chars — below where it started.
- ✅ **Fixed:** Rejected-alternative history moved out of `usage-tracking.md`; mermaid edge and the dangling "the two roles" both resolved.
- ✅ **Fixed:** `try`/`finally` scope question — moot, `packages/db` is untouched.
- 🆕 **New:** `notGrantable` is suppressed whenever `unreadable` is also non-empty (`packages/api-server/src/modules/usage/infrastructure/usage-view-grants-report.ts:40`).

### Findings
- 🟡 **Warning:** `notGrantable` is dropped when `unreadable` is also non-empty — inline-carried (`packages/api-server/src/modules/usage/infrastructure/usage-view-grants-report.ts:40`).

✅ No findings.

### Issue Fit
Linked issue: #3269 "DAMalytics update automation" (`Refs`, no closing keyword). ✅ Delivers the declared platform-side step — durable read access for the external analytics consumer — and leaves the consumer-side pipeline to follow; no undeclared scope.

- 🟡 `notGrantable` reported only when `unreadable` is empty (`usage-view-grants-report.ts:40`) — merged into Findings.
- 🟢 **Suggestion:** The fake returns one static row set for both catalog queries, so a granted view reads back as `unreadable` under test; in real Postgres the re-measure is in the same transaction and cannot. No assertion covers the combination (`packages/api-server/src/__tests__/unit/usage-view-grants.test.ts:43`).
- 🟢 **Suggestion:** `catalogReply` computes the pending list with the same predicate the SQL uses, so the grant-only-what-is-missing spec confirms the loop, not the query (`packages/api-server/src/__tests__/unit/usage-view-grants.test.ts:50`).
- 🟢 **Suggestion:** The "got past the role check" guard is repeated in four specs; one named helper would say it once (`packages/api-server/src/__tests__/unit/usage-view-grants.test.ts:106`).
- 🟢 **Suggestion:** The `catch` sets `rolePresent: true` even when the failing statement was the `to_regrole` probe, where presence is unknown. Unobservable today, since the report checks `failed` first (`packages/api-server/src/modules/usage/infrastructure/usage-view-grants.ts:144`).

### Verdict
COMMENT — the revert resolves both criticals at their cause and the containment is verified; one reporting-order defect remains, and it does not block access or startup.

---

docs/architecture/usage-tracking.md
```diff
@@ -152,7 +152,7 @@ When the inspector role is not configured at install time, the read endpoints ar
...
...ything it cannot make readable is warned about, under a name that separates what another start would fix from what needs an operator. Failure is reported the same way rather than raised: analytics access is optional and the platform starting is not, so a reconcile that cannot complete degrades instead of holding back the boot. The aggregate views are deliberately withheld from the group — a consumer able to read an aggregate would eventually key a metric on one, and renaming that aggregate would break it, which is the coupling this surface exists to remove. Withholding them leaves the passthrough column lists as the only contract. The group is inert until an operator adds a member, so an install that wants no analytics consumer has nothing to turn off. This is the database-level twin of the inspector role above — the platform offers read-only access over the usage views, and an operator decides who holds it; who that is stays outside the platform's knowledge. The passthroughs are deliberately outside the inspector surface: the `usage_*` aggregate views stay as the backing of the HTML report with no new features, while new metrics are authored consumer-side against the passthroughs.
```
dam-code-guardian: 🟢 **Suggestion:** This paragraph ends "It is a no-op where the group does not exist, which is most installs." The chart in this PR creates `usage_readers` unconditionally whenever bundled Postgres is on, and `postgres.enabled` defaults to `true` (`deploy/helm/platform/templates/postgres.yaml:58`, `deploy/helm/platform/values.yaml:172`). So on a default install the role exists and the reconcile does work on every start. Reported by doc-drift, confirmed against the chart.

The condition is the durable statement, and it is already correct one clause earlier: the reconcile is a no-op only where the role is absent. The frequency claim is what the chart default contradicts.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `531b13f3`

## PR #3434: chore(usage): preserve analytics grants across view migrations
**Author:** jjeliga | **Branch:** chore/3269-create-usage-viewers-pg-group-and-manage-view-access → main | **Changes:** +634 −24 (12 files)

### Summary
One commit answers the last round's reporting defect and two of its suggestions. The `usage.grants.incomplete` line now carries `notGrantable`, the catch reports the role presence it actually observed, and the test fake drives the two catalog queries from separate state. Nothing blocking remains.

### Changes since last review
Previous HEAD: `4e41ef8` (2026-08-25T18:37:10Z) — verdict COMMENT

- ✅ **Fixed:** `notGrantable` no longer disappears when `unreadable` is also non-empty — the `incomplete` warning carries both sets (`usage-view-grants-report.ts:42`). The new spec asserts the payload, not only the event name.
- ✅ **Fixed:** The fake returned one static row set for both catalog queries; `pending` and `after` are now independent, and a spec covers granted-then-readable (`usage-view-grants.test.ts:131`).
- ✅ **Fixed:** `catalogReply` no longer derives the pending list from the predicate the SQL uses, so the spec confirms the query rather than the loop (`usage-view-grants.test.ts:56`).
- ✅ **Fixed:** The catch reported `rolePresent: true` even when the `to_regrole` probe itself failed; a local now records what was observed (`usage-view-grants.ts:66`, `:147`).
- 🆕 **New:** The architecture page states the reconcile is a no-op on most installs, which the chart default contradicts (`docs/architecture/usage-tracking.md:155`).

### Findings
- 🟢 **Suggestion:** Architecture page overstates how rarely the reconcile does work — inline-carried (`docs/architecture/usage-tracking.md:155`).

- 🟢 **Suggestion:** `usage-tracking.md:155` says the reconcile "is a no-op where the group does not exist, which is most installs". The chart creates `usage_readers` unconditionally under `postgres.enabled`, which defaults to `true` (`deploy/helm/platform/templates/postgres.yaml:58`, `values.yaml:172`) — merged into Findings.

### Issue Fit
Linked issue: #3269 "DAMalytics update automation" (`Refs #3269`, no closing keyword, so the merge leaves it open). ✅ The declared platform-side scope fits, and nothing is undeclared.
- 🟢 **Suggestion:** #3269 also asks for the scheduled consumer-side pipeline, which needs a follow-up.

The commit's three changes verify correct as described. Five architecture-alignment items remain, all placement or typing shape on a working module reviewed clean last round — recorded as 🟢, not as change requests:
- 🟢 **Suggestion:** `bootstrap.ts:84` imports the usage module's infrastructure directly rather than through the module's public entry point.
- 🟢 **Suggestion:** The reconcile opens the transaction and takes the advisory lock inside `infrastructure/`, where the layered model puts the boundary in a service (`usage-view-grants.ts:77`).
- 🟢 **Suggestion:** Logger-driven reporting sits in `infrastructure/` (`usage-view-grants-report.ts:15`).
- 🟢 **Suggestion:** `UsageViewGrants` carries an optional `failed`, so one type spells both outcomes; `Result` in `core/result.ts` separates them (`usage-view-grants.ts:12`).
- 🟢 **Suggestion:** The reconcile takes the concrete `Db`, so the test opens a client on an unused port and reads the private `dialect` field only to render SQL (`usage-view-grants.test.ts:16`).

### Verdict
APPROVE — the reporting defect is fixed at its cause and the two test-fidelity suggestions with it; one documentation sentence and the placement items do not block.

---

docs/architecture/usage-tracking.md
jjeliga: Good catch, and it was backwards rather than imprecise — `postgres.enabled` defaults to `true` and the chart creates the role unconditionally when it manages Postgres, so the default install is the case where the group *does* exist.

Dropped the frequency claim rather than rewording it, since as you note the condition was already correct a clause earlier. Replaced with what is actually true: nothing at all where the group is absent, no writes once the state matches, a handful of catalog queries either way — which is also more honest about the default path doing real work on every start.

The same claim was propping up the reporter's rationale for logging an absent role at info. That reason does not need frequency at all: an absent role means no consumer is configured, which is a state rather than a fault. Both fixed in 416a196c.

pilartomas (approved):

jjeliga:
## Multi-replica

Flagged after approval: multi-replica api-server is coming, and the startup database path did not survive it. Two commits added, both verified on a live cluster.

**Migrations were not safe to run concurrently at all — and this is not a usage concern.** Drizzle's migrator takes no lock: it reads the newest applied migration, then applies everything newer. Started together against a database with migrations pending:

| | crashed |
|---|---|
| 3 replicas | **2 of 3** |
| 5 replicas | **4 of 5** |

They die before any migration runs, on the migrator's own `CREATE ... IF NOT EXISTS` bookkeeping, which is not atomic under concurrency — reproduced on this cluster's Postgres too, 4 of 6. Serializing on a database-held lock fixes it: 0 of 3, 0 of 5, 0 of 8.

This is **pre-existing and platform-wide** — it is the whole schema, not the usage views. It surfaced only because I was checking what multi-replica does to the reconcile that runs immediately after. `acf11413` touches nothing but `packages/db/src/migrate.ts`, so if the multi-replica work would rather own it, it lifts out cleanly.

**The reconcile was safe but shaped for one replica.** Every replica ran the identical pass, serialized on a *blocking* lock, in front of readiness. It now takes the lock without waiting, and a replica that loses skips the pass — including its verification read, because mid-pass the state is half-applied and a loser that measured it would raise a false `usage.grants.incomplete`. That false alarm is exactly why the lock blocked originally, so skipping removes the reason rather than trading one problem for another.

### Verified with real pods

Three replicas started together against an empty schema — identical `startTime`, **zero restarts**, 34 migrations recorded once with no duplicates, 32 usage views, all 7 passthroughs granted:

```
11:57:12.608  lh4nx   usage.grants.skipped
11:57:12.615  xskck   usage.grants.skipped
11:57:12.630  rjdrk   usage.grants.reconciled  granted:[usage_src_activity_events, ...]
```

One replica did the work, two skipped, inside a 22ms window. This is the contention path an earlier non-destructive run could not produce: with grants already in place the pass is too fast to overlap, so the winner needed real work for the losers to genuinely lose.

A side observation from the same session, worth knowing independently of this PR: a freshly rebuilt api-server logged `usage.grants.role-absent` on a cluster whose Postgres was initialised 21 days ago, because the chart's role SQL had never re-run there. That is the ordering gap the runbook describes, observed rather than argued.

### Decisions recorded

Both are recorded in ADR-085, alongside the decision they support. They differ deliberately — the grant pass skips under contention while migrations wait for the lock — so the reasoning sits beside them rather than in a separate record.

PR #3436, opened 2026-08-24
refactor(agent-runtime): one bounded subprocess runner

jezekra1:
Follow-up to @tomkis's observation on #3400: almost none of `history-provider.ts` was about history. Stacked on #3400 because the exec provider is one of the six sites and only exists on that branch.

## The problem

Six hand-rolled spawn blocks, disagreeing on everything that matters:

| site | stdout | on failure | decode | output cap |
|---|---|---|---|---|
| `git.ts` | ignored | log, never throws | concat at end | none |
| `workspace-command-plugin.ts` | relayed to log | reject | **per chunk** | none |
| `local-skill-repository.ts` | captured | reject with stderr | concat at end | none |
| `git-protocol-client.ts` ×2 | one ignores, one captures | reject with stderr | concat at end | none |
| `history-provider.ts` (exec) | captured | resolve `null` | StringDecoder | 32 MB |

The UTF-16 and chunk-boundary bugs that came out of the last review round were in the generic half of that table, and `workspace-command-plugin.ts` still had the same per-chunk `toString()` — a chunk boundary inside a multi-byte character silently substitutes U+FFFD.

## The shape

`core/run-once.ts` owns the mechanics and nothing about any caller's domain:

- **deadline is required** — an unbounded subprocess is never intended, so it is not defaultable;
- **byte-true budget** on retained output, so `String.length` can never stand in for a byte count again;
- **`StringDecoder`** end to end, including the trailing partial sequence at close;
- **failure is a value**, not an exception — the callers genuinely disagree about policy (log and continue, reject, fall back to another source), and an exception would force one on all of them. Follows the existing `core/result.ts` convention. That includes the errors `spawn` raises *synchronously* on a malformed argv, `cwd` or `env`: they map to the same `not-spawnable` value instead of escaping as a rejection, so no caller needs a `catch` to honour the contract.
- **capture or relay** — pass `onLine` and you get whole lines with nothing retained, which is what streaming a long build into a log wants. `maxOutputBytes` is one budget across **both** retained streams, because what it bounds is the memory the runner holds. A relaying caller retains nothing, so the only buffer there is the partial line being reassembled, and that is flushed as a line at 1 MB — a build writing a carriage-return progress meter never sends a newline, and that must not grow for the length of the run.

`describeFailure()` renders any failure uniformly, so the four sites that build error strings stop each inventing their own.

## Out of scope, deliberately

`create-child-agent-process`, `ssh.ts`, `pod-service.ts` and the terminal PTY keep their own `spawn`. They are long-lived children with lifecycles, not results — forcing them under a "run once, bounded, collect output" contract is how a helper like this turns into a framework.

## Behavior changes worth knowing

- Every **capturing** site now caps retained output; only the history provider did before. The one relaying site — `workspace-command` — retains nothing, so what is bounded there is one partial line, not total output: a long build still streams in full.
- `maxOutputBytes` is a combined stdout+stderr budget. The exec history provider previously counted stdout alone against its 32 MB, so a history command writing tens of megabytes to stderr now trips the cap it would once have survived.
- `workspace-command` relays whole lines instead of raw chunks (fixes its decode bug as a side effect) and its failure message renders through `describeFailure`.
- Messages are uniform, with stderr clamped to 500 chars for log hygiene — the old exec provider clamped at 200, the others were unbounded.
- `git.ts` now pipes stdout instead of ignoring it (tiny output, and the cap protects it).

Net −202/+53 across the six sites, plus the runner and 10 tests covering what none of the sites tested individually: deadline kills, byte cap, split-multibyte reassembly, non-zero exit, unspawnable binary, empty command, a malformed argv that `spawn` rejects synchronously, line relay including a final line with no newline, and a newline-less stream that must not buffer without bound. Plus a `git-protocol-client` spec driving real `git` on disk — that client had no test at all. Runtime 253/253.

### DIFF
--- a/packages/agent-runtime/src/core/run-once.ts
+++ b/packages/agent-runtime/src/core/run-once.ts
@@ -1,4 +1,5 @@
-import { spawn } from "node:child_process";
+import { type ChildProcessByStdio, spawn } from "node:child_process";
+import type { Readable } from "node:stream";
 import { StringDecoder } from "node:string_decoder";
 
 import { err, ok, type Result } from "./result.js";
@@ -6,6 +7,8 @@ import { err, ok, type Result } from "./result.js";
 const DEFAULT_MAX_OUTPUT_BYTES = 8 * 1024 * 1024;
 
 const MAX_DESCRIBED_STDERR = 500;
+
+const MAX_PENDING_LINE_BYTES = 1024 * 1024;
 
 export interface RunOnceOptions {
   command: readonly string[];
@@ -56,12 +59,18 @@ export function describeFailure(
  * (SIGKILL on expiry, and the timeout is required — an unbounded subprocess is
  * never intended), a byte budget on retained output, and decoding that
  * reassembles multi-byte characters split across stream chunks. Failure is a
- * value, never an exception, because callers disagree about policy: some log
- * and continue, some reject, some fall back to another source. Output is
- * either retained and returned, or — when onLine is given — relayed line by
- * line and not retained at all, which is what a caller streaming a long build
- * into a log wants; only retained output is capped. Long-lived children (the
- * harness, a PTY, sshd) are out of scope: they have lifecycles, not results.
+ * value, never an exception — including the errors spawn raises
+ * synchronously on a malformed argv, cwd or env — because callers disagree
+ * about policy: some log and continue, some reject, some fall back to another
+ * source. Output is either retained and returned, or — when onLine is given —
+ * relayed line by line and not retained at all, which is what a caller
+ * streaming a long build into a log wants. maxOutputBytes caps retained
+ * output as one budget across both streams, because what it bounds is the
+ * memory this runner holds; a relaying caller retains nothing, so it is capped
+ * only by the 1 MB bound on one partial line, which is flushed as a line when
+ * a stream produces that many bytes without a newline. Long-lived children
+ * (the harness, a PTY, sshd) are out of scope: they have lifecycles, not
+ * results.
  */
 export function runOnce(opts: RunOnceOptions): Promise<RunOnceResult> {
   const maxOutputBytes = opts.maxOutputBytes ?? DEFAULT_MAX_OUTPUT_BYTES;
@@ -74,11 +83,19 @@ export function runOnce(opts: RunOnceOptions): Promise<RunOnceResult> {
       return;
     }
 
-    const child = spawn(bin, args, {
-      cwd: opts.cwd,
-      env: opts.env,
-      stdio: ["ignore", "pipe", "pipe"],
-    });
+    let child: ChildProcessByStdio<null, Readable, Readable>;
+    try {
+      child = spawn(bin, args, {
+        cwd: opts.cwd,
+        env: opts.env,
+        stdio: ["ignore", "pipe", "pipe"],
+      });
+    } catch (error) {
+      resolve(
+        err({ kind: "not-spawnable", message: (error as Error).message }),
+      );
+      return;
+    }
 
     let settled = false;
     let stdout = "";
@@ -103,13 +120,21 @@ export function runOnce(opts: RunOnceOptions): Promise<RunOnceResult> {
     ): (() => void) => {
       const decoder = new StringDecoder("utf8");
       let pending = "";
+      let pendingBytes = 0;
       stream.on("data", (chunk: Buffer) => {
         const text = decoder.write(chunk);
         if (relay) {
           pending += text;
+          pendingBytes += chunk.length;
           const lines = pending.split("\n");
           pending = lines.pop() ?? "";
+          if (lines.length > 0) pendingBytes = Buffer.byteLength(pending);
           for (const line of lines) if (line) relay(line);
+          if (pendingBytes > MAX_PENDING_LINE_BYTES) {
+            relay(pending);
+            pending = "";
+            pendingBytes = 0;
+          }
           return;
         }
         retainedBytes += chunk.length;
--- a/packages/agent-runtime/src/core/run-once.ts
+++ b/packages/agent-runtime/src/core/run-once.ts
@@ -30,20 +30,21 @@ export type ProcessFailure =
 export type RunOnceResult = Result<ProcessOutput, ProcessFailure>;
 
 export function describeFailure(
-  command: readonly string[],
+  subject: string,
   failure: ProcessFailure,
 ): string {
-  const name = command.join(" ");
   switch (failure.kind) {
     case "not-spawnable":
-      return `${name} failed to start: ${failure.message}`;
+      return `${subject} failed to start: ${failure.message}`;
     case "timed-out":
-      return `${name} timed out after ${failure.timeoutMs}ms`;
+      return `${subject} timed out after ${failure.timeoutMs}ms`;
     case "output-capped":
-      return `${name} produced more than ${failure.maxOutputBytes} bytes of output`;
+      return `${subject} produced more than ${failure.maxOutputBytes} bytes of output`;
     case "exited": {
       const trimmed = failure.stderr.trim().slice(0, MAX_DESCRIBED_STDERR);
-      return `${name} exited ${failure.code}` + (trimmed ? `: ${trimmed}` : "");
+      return (
+        `${subject} exited ${failure.code}` + (trimmed ? `: ${trimmed}` : "")
+      );
     }
   }
 }
--- a/packages/agent-runtime/src/core/run-once.ts
+++ b/packages/agent-runtime/src/core/run-once.ts
@@ -1,0 +1,148 @@
+import { spawn } from "node:child_process";
+import { StringDecoder } from "node:string_decoder";
+
+import { err, ok, type Result } from "./result.js";
+
+const DEFAULT_MAX_OUTPUT_BYTES = 8 * 1024 * 1024;
+
+const MAX_DESCRIBED_STDERR = 500;
+
+export interface RunOnceOptions {
+  command: readonly string[];
+  timeoutMs: number;
+  cwd?: string;
+  env?: NodeJS.ProcessEnv;
+  maxOutputBytes?: number;
+  onLine?: (line: string) => void;
+}
+
+export interface ProcessOutput {
+  stdout: string;
+  stderr: string;
+}
+
+export type ProcessFailure =
+  | { kind: "not-spawnable"; message: string }
+  | { kind: "timed-out"; timeoutMs: number }
+  | { kind: "output-capped"; maxOutputBytes: number }
+  | { kind: "exited"; code: number | null; stderr: string };
+
+export type RunOnceResult = Result<ProcessOutput, ProcessFailure>;
+
+export function describeFailure(
+  command: readonly string[],
+  failure: ProcessFailure,
+): string {
+  const name = command.join(" ");
+  switch (failure.kind) {
+    case "not-spawnable":
+      return `${name} failed to start: ${failure.message}`;
+    case "timed-out":
+      return `${name} timed out after ${failure.timeoutMs}ms`;
+    case "output-capped":
+      return `${name} produced more than ${failure.maxOutputBytes} bytes of output`;
+    case "exited": {
+      const trimmed = failure.stderr.trim().slice(0, MAX_DESCRIBED_STDERR);
+      return `${name} exited ${failure.code}` + (trimmed ? `: ${trimmed}` : "");
+    }
+  }
+}
+
+/**
+ * UNIT_BOUNDARY_DESCRIPTION: Runs one short-lived subprocess to completion and
+ * reports what happened, so no caller hand-rolls the spawn/decode/deadline
+ * block. It owns the parts that are easy to get subtly wrong: a deadline
+ * (SIGKILL on expiry, and the timeout is required — an unbounded subprocess is
+ * never intended), a byte budget on retained output, and decoding that
+ * reassembles multi-byte characters split across stream chunks. Failure is a
+ * value, never an exception, because callers disagree about policy: some log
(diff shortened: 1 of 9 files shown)


packages/agent-runtime/src/core/run-once.ts
```diff
@@ -0,0 +1,149 @@
...
+    if (bin === undefined) {
+      resolve(err({ kind: "not-spawnable", message: "empty command" }));
+      return;
+    }
+
+    const child = spawn(bin, args, {
+      cwd: opts.cwd,
+      env: opts.env,
+      stdio: ["ignore", "pipe", "pipe"],
+    });
```
dam-code-guardian: 🟡 **Warning:** `runOnce` is documented as total — "Failure is a value, never an exception" — but `spawn` throws **synchronously**, before any listener is attached, for an invalid argv, `cwd` or `env`. Verified on node 24: an empty-string binary, a NUL byte in the binary name, in any argument, in `cwd`, or in an env key or value all raise `ERR_INVALID_ARG_VALUE` from this call. The `bin === undefined` guard above covers only an empty array, so the returned promise **rejects** rather than resolving to `not-spawnable`.

Two callers inherit that. `createExecHistoryProvider.fetch` is documented to resolve `null` on any deviation, and its command comes from the runtime manifest, where `z.array(z.string().min(1))` admits `[" "]` but not `[""]` — so the reachable case there is a NUL-carrying `cwd`. `configureGitCredentialHelper` is the sharper one: it attaches `.then()` with no `.catch()`, so a rejection becomes an unhandled rejection at boot.

**Fix:** A function whose contract is "errors are returned, not thrown" wraps the whole `spawn` call in `try`/`catch` and maps the synchronous throw onto the same `not-spawnable` value — one guard covers every invalid-argument class, so no caller needs its own.

```suggestion
    let child;
    try {
      child = spawn(bin, args, {
        cwd: opts.cwd,
        env: opts.env,
        stdio: ["ignore", "pipe", "pipe"],
      });
    } catch (error) {
      resolve(err({ kind: "not-spawnable", message: (error as Error).message }));
      return;
    }
```

packages/agent-runtime/src/core/run-once.ts
```diff
@@ -0,0 +1,149 @@
...
+      let pending = "";
+      stream.on("data", (chunk: Buffer) => {
+        const text = decoder.write(chunk);
+        if (relay) {
+          pending += text;
+          const lines = pending.split("\n");
+          pending = lines.pop() ?? "";
+          for (const line of lines) if (line) relay(line);
+          return;
+        }
```
dam-code-guardian, line 139: 🟡 **Warning:** In the relay mode this branch returns before `retainedBytes` is touched, so `maxOutputBytes` is never consulted — and `pending` grows without a bound until a `\n` arrives. Output that carries no newline is retained in full: a `bash -lc` command writing carriage-return progress (`printf 'x%%\r'` in a loop, a `curl` or `docker` progress bar, `dd status=progress`) accumulates every byte and relays nothing. Confirmed by transcribing the split/pop math: three CR-only chunks leave all 36 bytes in `pending` with zero lines relayed.

`workspace-command-plugin.ts` is the one `onLine` caller, and its deadline is 15 minutes, so this is the site with both the longest window and the most likely CR-style output. The deadline still bounds it, but the doc comment's "not retained at all" and the body's "Every site now caps output" both read as guarantees this path does not give.

**Fix:** Count relayed bytes against the same budget and cap the un-terminated tail, so a byte budget means the same thing in both modes — or state the exception in the option's contract and drop the blanket claim.

```suggestion
        if (relay) {
          retainedBytes += chunk.length;
          pending += text;
          const lines = pending.split("\n");
          pending = lines.pop() ?? "";
          for (const line of lines) if (line) relay(line);
          if (retainedBytes > maxOutputBytes) {
            child.kill("SIGKILL");
            settle(err({ kind: "output-capped", maxOutputBytes }));
          }
          return;
        }
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `665ca2fe`

## PR #3436: refactor(agent-runtime): one bounded subprocess runner
**Author:** jezekra1 | **Branch:** refactor/shared-subprocess-runner → main | **Changes:** +465 −202 (9 files)

### Summary
Replaces six hand-rolled `spawn` blocks with one `core/run-once.ts` runner that owns the deadline, a byte budget on retained output, `StringDecoder` reassembly, and failure-as-a-value. The consolidation is correct at every converted site, and the `describeFailure` uniformity removes four separately-invented error strings.

### Findings
- 🟡 **Warning:** `runOnce` is documented as total ("Failure is a value, never an exception") but `spawn` throws **synchronously** on an invalid argv, `cwd` or `env`, so the promise rejects instead of resolving to `not-spawnable` (`packages/agent-runtime/src/core/run-once.ts:77`)
  **Fix:** See the inline comment.
- 🟡 **Warning:** In the relay mode the `pending` buffer accumulates without a bound and the byte budget is never consulted, so a caller streaming output with no newline retains it all (`packages/agent-runtime/src/core/run-once.ts:108`)
  **Fix:** See the inline comment.
- 🟢 **Suggestion:** `maxOutputBytes` is checked against one `retainedBytes` counter shared by stdout and stderr (`packages/agent-runtime/src/core/run-once.ts:115`), so it is now a combined budget. Every replaced site counted stdout alone — the history provider's `stdoutBytes` is the clearest case — so a command with chatty stderr can now trip `output-capped` and discard a valid stdout result. Either make the budget per retained stream, or state the combined semantics in the option's contract and in the body's behavior-change list.

✅ No findings.

### Issue Fit
_No linked issue — the PR references #3400 (a PR, not an issue) and a reviewer observation; the head branch carries no issue number. Scope checked against the PR description only._

- 🟡 **Warning:** The body's "Behavior changes worth knowing" claims "Every site now caps output; only the history provider did before", but the relay path returns before the budget is consulted, so `workspace-command-plugin.ts` — the one `onLine` caller — is not capped. This contradicts the body's own "only retained output is capped" rule in the shape section (`packages/agent-runtime/src/core/run-once.ts:108`). Same defect as the second finding above.
  **Fix:** State a behavior-change claim per output mode when a helper has two, rather than asserting it for every call site.
- 🟢 **Suggestion:** The body's inventory ("the runner and 8 tests") omits the new 104-line `git-protocol-client.test.ts`, which drives real `git` on disk (`packages/agent-runtime/src/__tests__/unit/git-protocol-client.test.ts:1`).

- 🟡 **Warning:** `runOnce`'s documented totality does not hold — `spawn` raises `ERR_INVALID_ARG_VALUE` synchronously inside the promise executor, and the `bin === undefined` guard covers only an empty array (`packages/agent-runtime/src/core/run-once.ts:73`). Two callers inherit it: `createExecHistoryProvider.fetch` is documented to resolve `null` on any deviation, and `configureGitCredentialHelper` attaches `.then()` with no `.catch()` (`packages/agent-runtime/src/modules/git.ts:15`). Same defect as the first finding above.
  **Fix:** A function whose contract is "errors are returned, not thrown" wraps the whole spawn call in `try`/`catch` and maps the synchronous throw to the same failure value.
- 🟢 **Suggestion:** The combined stdout+stderr byte budget described in the third finding above.

### Verdict
COMMENT — the consolidation is sound and every converted site is faithful, but the shared runner's two contract gaps (a reachable rejection from a documented-total function, and an unbounded buffer in the relay mode) are worth closing before six call sites depend on them.


---

tomkis (approved):

Tomas2D (approved):

packages/agent-runtime/src/core/run-once.ts
```diff
@@ -0,0 +1,174 @@
...
+      let pendingBytes = 0;
+      stream.on("data", (chunk: Buffer) => {
+        const text = decoder.write(chunk);
+        if (relay) {
+          pending += text;
+          pendingBytes += chunk.length;
+          const lines = pending.split("\n");
+          pending = lines.pop() ?? "";
+          if (lines.length > 0) pendingBytes = Buffer.byteLength(pending);
+          for (const line of lines) if (line) relay(line);
```
dam-code-guardian, line 132: 🟢 **Suggestion:** `if (line)` skips empty lines, so a subprocess that separates output blocks with blank lines loses that structure in the log.

**Fix:** Relay every line the stream produced. Filter only the artificial trailing empty segment that `split("\n")` creates — that segment is already handled by `lines.pop()`.

```suggestion
          for (const line of lines) relay(line);
```

packages/agent-runtime/src/core/run-once.ts
```diff
@@ -0,0 +1,174 @@
...
+          if (pendingBytes > MAX_PENDING_LINE_BYTES) {
+            relay(pending);
+            pending = "";
+            pendingBytes = 0;
+          }
+          return;
+        }
+        retainedBytes += chunk.length;
```
dam-code-guardian, line 140: 🟢 **Suggestion:** The `data` handler has no `settled` guard. After `output-capped` resolves, this keeps appending until SIGKILL lands, so retained memory overshoots the budget it just enforced.

**Fix:** Stop consuming on settle — pause or unbind both streams inside `settle`, so no handler can grow state after the result is fixed.

packages/agent-runtime/src/core/run-once.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    child.on("error", (error) =>
+      settle(err({ kind: "not-spawnable", message: error.message })),
+    );
+
+    child.on("close", (code) => {
+      flushStdout();
+      flushStderr();
+      if (code !== 0) {
+        settle(err({ kind: "exited", code, stderr }));
```
dam-code-guardian, line 168: 🟢 **Suggestion:** In relay mode nothing is retained, so an `exited` failure always carries `stderr: ""` and `describeFailure` prints a bare `workspace command exited 1`. The one relaying caller loses nothing today (its lines already reach the log), but the failure type advertises a field this mode never fills.

**Fix:** Either retain a bounded stderr tail even under relay, or model the relay failure so an empty `stderr` cannot be read as "no diagnostic output".

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `1d94c0bc`

## PR #3436: refactor(agent-runtime): one bounded subprocess runner
**Author:** jezekra1 | **Branch:** refactor/shared-subprocess-runner → main | **Changes:** +528 −202 (9 files)

### Summary
One new commit closes both contract gaps the last round raised: `spawn` is wrapped, so a malformed argv now becomes a `not-spawnable` value, and the relay path flushes a partial line at 1 MB instead of holding a newline-less stream for the run. The six converted call sites are unchanged and still faithful.
_(Suppressed 1 finding per PR context: tar listing now capped at the 8 MB default.)_

### Changes since last review
Previous HEAD: `665ca2fe` (2026-08-24T14:41:00Z) — verdict COMMENT

- ✅ **Fixed:** `runOnce` is total — `spawn` runs inside `try`/`catch` and its synchronous throw maps to `not-spawnable` (`packages/agent-runtime/src/core/run-once.ts:87`)
- ✅ **Fixed:** the relay buffer is bounded — a stream with no newline flushes its partial line at 1 MB (`packages/agent-runtime/src/core/run-once.ts:133`)
- ✅ **Fixed:** the combined stdout+stderr budget is now stated in the unit contract and in the body (`packages/agent-runtime/src/core/run-once.ts:68`)
- ✅ **Fixed:** the body's cap claim now separates the capturing sites from the relaying one
- ✅ **Fixed:** the body's test inventory now names the `git-protocol-client` spec

### Findings
- 🟢 **Suggestion:** the relay drops blank lines (`packages/agent-runtime/src/core/run-once.ts:132`)
- 🟢 **Suggestion:** `exited.stderr` is always empty in relay mode (`packages/agent-runtime/src/core/run-once.ts:168`)
- 🟢 **Suggestion:** retained output grows after `output-capped` settles (`packages/agent-runtime/src/core/run-once.ts:140`)

✅ No findings.

### Issue Fit
_No linked issue — `#3400` is a pull request, so scope was checked against the PR description only._

✅ No findings. The diff delivers every claim in the body: the deadline is required, the budget counts bytes, `StringDecoder` runs end to end, failure is a `Result`, and the relay buffer is bounded at 1 MB. All six declared `spawn` blocks now call `runOnce`. The four sites declared out of scope keep their own `spawn`. `run-once.test.ts` holds the 10 stated tests.

- 🟢 **Suggestion:** `for (const line of lines) if (line) relay(line)` skips empty lines, so a subprocess that separates output blocks with blank lines loses that structure in the log (`packages/agent-runtime/src/core/run-once.ts:132`). Relay every line the stream produced, and filter only the trailing empty segment `split("\n")` creates.
- 🟢 **Suggestion:** in relay mode nothing is retained, so an `exited` failure always carries `stderr: ""` and `describeFailure` prints a bare `… exited 1`. The one relaying caller loses nothing (its lines already reach the log), but the failure type advertises a field that mode never fills (`packages/agent-runtime/src/core/run-once.ts:168`). Either retain a bounded stderr tail under relay, or model the relay failure so the empty field cannot be read as diagnostic.
- 🟢 **Suggestion:** the `data` handler has no `settled` guard, so after `output-capped` resolves it keeps appending until SIGKILL lands and retained memory overshoots the budget just enforced (`packages/agent-runtime/src/core/run-once.ts:140`). Pause or unbind the streams on settle.

_The `pendingBytes` accounting was checked and is sound: the `Buffer.byteLength(pending)` recompute after each newline keeps the counter exact, decoder holdback over-counts by at most 3 bytes and self-corrects, and the forced flush cannot corrupt UTF-8 because `pending` is already decoded._

### Verdict
APPROVE — the round fixes every finding from the previous review at its cause, adds tests that pin both contracts, and introduces no blocking finding; the three remaining items are 🟢 seams in the new runner.

---

PR #3444, opened 2026-08-24
feat(dam-experiment): add tiny-search example and update documentation

Tomas2D:
Adds a second optimization-campaign starter and hardens the experiment SDK's write path. Every subsystem this branch touches is listed below.

### `dam-experiment` skill — new `tiny-search` starter
- A minimal, dependency-free in-memory full-text search index for Node.js, deliberately naive so a campaign has real headroom (~50–70×) against a behavioral suite that must stay green.
- Behavioral suite (18 tests, `node:test`) pinning the tokenizer, AND semantics, occurrence scoring, ordering and tie-breaks, limits, replace-on-re-add and removal.
- Deterministic benchmark with `query-heavy` / `index-heavy` / `mixed` scenarios and a behavior checksum that is stable across runs but changes when behavior does.
- `README.md` states the contract the campaign optimizes against, including the Unicode tokenizer rule, the `limit` rule, and the quadratic corpus build.
- `references/tiny-search-starter.md`, plus `SKILL.md` and `references/images.md` / `references/nous-evaluator.md` updates.

### `dam-experiment` skill — docs beyond the starter
- `SKILL.md` gains a `campaign.yaml` design-control section and expanded scoring/validation and cost-reporting guidance.
- `commands/experiment-onboard.md` is rewritten to offer both starters.

### `experiment-sdk` — write path
- Event reports and `finish` retry transient failures; `spawn` still never retries.
- The event buffer is bounded to the events route's 500-event cap. Previously a retained batch grew without limit during an outage, passed the cap, and was then rejected on every later attempt — reporting wedged permanently for the rest of the run.
- A failed flush parks further attempts for a back-off window, so a long outage costs one retry ladder per window instead of one per reported event. `finish` bypasses the window to give the tail one last attempt.

### `controller`
- Raises the agent liveness probe failure threshold so a busy agent is not restarted mid-turn.

### `deploy`
- `lima-k3s.yaml` gains a note on host-CPU starvation as the usual cause of cluster-wide probe-timeout cascades on the local dev cluster.

### `docs/architecture`
- `experiments.md` records the reporting contract this branch changes: which writes retry and why, that a rejected batch is retained within the accepted size, that a failed flush backs off, and that `finish` is a no-op on an already-terminal run.

### `mise.lock`
- Restores the `tools.java.*` platform entries that an earlier commit on this branch dropped as collateral (a local mise run rewrites the lockfile with only the platforms it resolved on the host).

### DIFF
--- a/packages/agents/claude-code/dam-skills/dam-experiment/SKILL.md
+++ b/packages/agents/claude-code/dam-skills/dam-experiment/SKILL.md
@@ -135,11 +135,11 @@ whole envelope and get an explicit yes before writing the script:
   rather than at the first failed spawn.
 - **Iteration counts**: your loop's rounds, plus any the worker runs
   internally. Total runtime multiplies through them.
-- **Expected duration, stated in human terms.** "~25 min per campaign
-  iteration, two iterations, one round — about 1 h with queue slack" is a
-  deciding factor, not a footnote: the human may cut seeds or rounds to fit
-  the time they have, so give them the per-unit cost that makes that trade
-  legible. The `ttl_ms` you set is derived from this number — never the other
+- **Expected duration, stated in human terms.** "~1 h per campaign
+  iteration, one confirming iteration, one round — about 1 h with queue
+  slack" is a deciding factor, not a footnote: the human may cut seeds or
+  rounds to fit the time they have, so give them the per-unit cost that
+  makes that trade legible. The `ttl_ms` you set is derived from this number — never the other
   way around — and it is a **kill deadline, not pacing**: the platform reaps
   the target the moment it lapses, even mid-work. Size it at the worst
   plausible round plus generous slack (cold start alone is minutes; double
@@ -431,7 +431,12 @@ them on the table — then wrongly concludes it "can't force" a design it wants:
 
 `target_system.description` is substituted verbatim into the model's prompts —
 it is where baselines, exact CLI invocations, metric definitions and
-statistical guardrails belong. `plot_specs` runs figure scripts after each
+statistical guardrails belong. What it is NOT for: dictating the bundle's
+arms. The arm family is the tool's methodology and the reason its verdict is
+defensible; a measured iteration already includes the full standard bundle.
+Time-box with the schema's own knobs instead — `max_iterations`, `max_turns`
+(per-phase tool-use caps, e.g. `{design: 40, execute_analyze: 80}`), and the
+TTL sized from measured runs. `plot_specs` runs figure scripts after each
 `findings.json`, and `pre_work_script` runs a deterministic exploration before
 iteration 1 (a good place to measure the baseline). Full field list:
 `nous schema campaign`, mirrored in the image's own
@@ -471,6 +476,10 @@ instead of reporting the number as clean.
 Then report the objective score from best_found.json, the h-main arm status
 from the last iteration's findings.json, and a summary from
 meta_findings.json.
+Publish as you go, not only at the end: the moment an iteration completes,
+publish its findings.json (and any raw per-seed results worth keeping) as
+artifacts, and publish report.md the moment it exists. You can be killed by
+your deadline mid-campaign; whatever is unpublished at that moment is lost.
 Report per_seed as well: for EVERY seed of the confirming iteration, read the
 raw measurement files under runs/iter-<N>/results/<arm>/ and report its seed,
 baseline and treatment values for the primary metric, in the metric's own
@@ -560,17 +569,25 @@ Six things this example is really teaching:
   completion, or it stalls waiting for an answer nobody sends.
 - **Match the TTL to the real runtime — and know the clock starts at spawn.**
   `nous`-class work will run for hours if you let it, and the default liveness
-  deadline is not a promise your loop should lean on. Scope the campaign to
-  about an hour of work (see the interview above), estimate ~20–30 min per
-  campaign iteration at that size, and set the TTL at roughly double the
-  estimate: a spawn that queues for compute room spends its deadline waiting,
-  and a deadline that lands mid-measurement wastes the whole round. A heavy
-  worker also holds real CPU/memory/disk per target, so keep the fan-out to a
-  handful of arms, not dozens.
-- **Everything on the worker dies with it.** The worker is reaped right after
-  it reports; make the prompt name what to report and which files to publish
-  as artifacts (`report.md`, `meta_findings.json` — see the reference) or the
-  evidence is unrecoverable.
+  deadline is not a promise your loop should lean on. Estimate from **measured
+  runs, not hope**: a healthy 2-iteration campaign on a small target measured
+  ~60 min per iteration on a dev cluster — so the one-hour target means ONE
+  confirming iteration, and a 2-iteration campaign is a ~2 h proposal, said as
+  such. Set the TTL at roughly double the measured estimate: the clock also
+  pays for pod cold start, queueing for compute room, and degraded stretches —
+  a worker that cold-started into a cluster disturbance has crawled at a
+  twentieth of its healthy pace and then been executed by a TTL sized for the
+  healthy pace, losing everything. A heavy worker also holds real
+  CPU/memory/disk per target, so keep the fan-out to a handful of arms, not
+  dozens.
+- **Everything on the worker dies with it — so publish per iteration.** The
+  worker is reaped the moment it reports, hits its TTL, or its run is swept;
+  make the prompt name what to report and which files to publish as artifacts
+  (`report.md`, `meta_findings.json` — see the reference), and require each
+  iteration's `findings.json` to be published **as it completes**, not in a
+  final batch. Three consecutive campaigns have died mid-run (a starved node,
+  an inactivity reap, a TTL) and each lost hours of finished measurements that
+  publish-as-you-go would have kept.
 - **Bound the campaign's own parallelism — for survival *and* for the
   numbers.** The worker's memory limit is a hard ceiling for everything the
   campaign starts — benchmark daemons included — and blowing it OOM-kills the
--- a/packages/agents/claude-code/dam-skills/dam-experiment/SKILL.md
+++ b/packages/agents/claude-code/dam-skills/dam-experiment/SKILL.md
@@ -48,10 +48,14 @@ The chat shows these as one-click chips on a fresh session, so answer them
 well — scannable, per Talking to the user above, ending on what the user can
 do next:
 
-- **"Show me an example to optimize"** → the tiny-cache starter
-  ([references/tiny-cache-starter.md](references/tiny-cache-starter.md)):
-  describe it in two or three sentences and offer to set it up. Show, don't
-  run.
+- **"Show me an example to optimize"** → the two bundled starters, one line
+  each: **tiny-cache** (a slow cache, rewritten by a coding agent each round —
+  minutes) and **tiny-search** (a slow search index optimized by a `nous`
+  campaign against a pre-registered bar and a write-path guard — about an
+  hour). Describe, ask which, and offer to set it up. Show, don't run; the
+  references are
+  [tiny-cache](references/tiny-cache-starter.md) and
+  [tiny-search](references/tiny-search-starter.md).
 - **"How do experiments work?"** → the loop in plain words: you describe a
   goal, we agree the design, I write it as a Python loop; each round proposes
   a candidate, builds it, measures it, and reports a score; the platform
@@ -168,15 +172,24 @@ Present it as a short list of decisions, not prose, and let them change any
 line. Silence is not approval: if they have not answered, ask again rather
 than picking a default and proceeding.
 
-## No target? The bundled starter
+## No target? Two bundled starters
 
-A user who wants to see how experiments work but has nothing to optimize gets
-**tiny-cache** — a deliberately slow cache with a behavioral suite and a
-deterministic benchmark, shipped inside this skill at
-[examples/tiny-cache/](examples/tiny-cache/). Follow
-[references/tiny-cache-starter.md](references/tiny-cache-starter.md) for the
-baseline ritual and the loop design — the worker harness is the user's pick,
-as always.
+A user who wants to see how experiments work but has nothing of their own to
+optimize gets one of two deliberately-slow Node packages shipped inside this
+skill. Both are dependency-free, single-process, and measurable in seconds;
+they differ in which worker they are shaped for, so pick by what the user
+wants to see:
+
+| Starter | Worker | Shows |
+|---|---|---|
+| [tiny-cache](references/tiny-cache-starter.md) ([code](examples/tiny-cache/)) | `claude-code` loop | your driver owning the ruler: it runs the locked bench, sweeps n, and scores each point |
+| [tiny-search](references/tiny-search-starter.md) ([code](examples/tiny-search/)) | `nous` campaign | a worker forming a hypothesis, pre-registering a bar, and reporting whether the mechanism held — with a guard metric it can fail |
+
+Follow the reference for whichever they choose; each carries its baseline
+ritual, its scoring design, and the size that fits the hour. **Don't cross
+them over.** tiny-cache on `nous` loses the pristine ruler and lands its
+optimized side on the timer floor; tiny-search in a hand-driven loop throws
+away the campaign machinery that is the only reason to look at it.
 
 ## Authoring a script
 
@@ -387,6 +400,42 @@ and when it breaks the hour, **cut scope, never the TTL** — fewer seeds, fewer
 iterations, a narrower question. The TTL still gets generous slack over
 whatever the estimate ends up being (a killed working pod wastes everything);
 the hour budgets the *work*, not the deadline.
+
+### You do not write the campaign's arms — but you do pin its design
+
+The worker authors its own hypothesis bundle: the arms (`h-main`,
+`h-ablation`, `h-super-additivity`, `h-control-negative`, `h-robustness`, and
+campaign-specific ones like `h-dose-response`) come out of its DESIGN phase,
+not out of your prompt. That is the point of the image, and it is why a Nous
+round is one spawn instead of a loop you drive.
+
+**It does not follow that you have no control.** `campaign.yaml` carries four
+fields that steer the design hard, and an agent that only writes prose leaves
+them on the table — then wrongly concludes it "can't force" a design it wants:
+
+- **`research_question`** — phrase it as the shape of the answer you want. "Is
+  X faster?" invites a yes/no; "does X's cost become independent of n?" invites
+  a curve, and the designer reaches for a dose-response arm on its own.
+- **`target_system.controllable_knobs`** — names the knobs the designer may
+  vary. Put the sweep variable here or it may never be swept.
+- **`ground_truth.direction_claim`** / `pass_condition` / `primary_metric` /
+  `seeds` — pre-registered and rendered into the agent's prompt, so it cannot
+  move the goalposts. A direction claim stated as a trend ("baseline cost grows
+  ~linearly in n while the treatment stays flat") commits the campaign to
+  measuring the trend.
+- **`locked_parameters`** — hard-pinned values that MUST reappear identically
+  in the bundle's `verified_parameters`; a mismatch fails validation **even
+  under `--auto-approve`**. This is the actual enforcement mechanism: put the
+  n grid, the rep count, and the seed list here and the campaign cannot
+  quietly measure something else.
+
+`target_system.description` is substituted verbatim into the model's prompts —
+it is where baselines, exact CLI invocations, metric definitions and
+statistical guardrails belong. `plot_specs` runs figure scripts after each
+`findings.json`, and `pre_work_script` runs a deterministic exploration before
+iteration 1 (a good place to measure the baseline). Full field list:
+`nous schema campaign`, mirrored in the image's own
+`.agents/skills/nous/reference/campaign-schema.md`.
 
 Bake the answers into the campaign prompt so the worker doesn't re-decide
 them. How the worker's results are laid out on disk — the stable verdict
--- a/packages/agents/claude-code/dam-skills/dam-experiment/SKILL.md
+++ b/packages/agents/claude-code/dam-skills/dam-experiment/SKILL.md
@@ -486,7 +486,10 @@ baseline and treatment values for the primary metric, in the metric's own
 unit. Read those numbers from the raw files, never from report.md's prose or
 tables. Report per_arm too — arm_type, status and effect for every arm you
 measured, ablations included; the decomposition is the mechanism, and prose is
-not where it survives."""
+not where it survives.
+Report cost too, from llm_metrics_summary.json (or `nous cost <run_id>`):
+total USD, total tokens, and LLM call count. These files are stable and
+always exist — report them even when the campaign fails."""
 
 with x.Experiment("nous-campaigns") as exp:
     loop = exp.loop("rounds", description="one Nous campaign per pass")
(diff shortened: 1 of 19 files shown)


packages/agents/claude-code/dam-skills/dam-experiment/examples/tiny-search/src/tiny-search.js
```diff
@@ -0,0 +1,79 @@
+/**
```
dam-code-guardian: 🔴 **Critical:** This file adds a 17-line untyped prose comment. `docs/guidelines/comment-guidelines.md` allows only comments prefixed with a registered type (`TEST_OVERVIEW`, `TEST_SCENARIO`, `UNIT_BOUNDARY_DESCRIPTION`), and `scripts/check-comment-types.mjs` scans every tracked `.js`/`.mjs` file — so `mise run common:check:comment-types` fails on this comment and on `bench/bench.mjs:1` and `bench/bench.mjs:43`. The sibling `examples/tiny-cache/` carries zero comments, so this tree is in scope.

**Fix:** Comments in `.js`/`.mjs` source carry a registered `COMMENT_TYPES` prefix or do not exist. The behavioral contract belongs in `README.md`, which already states it; `UNIT_BOUNDARY_DESCRIPTION:` is the alternative if it must stay in the file.

packages/agents/claude-code/dam-skills/dam-experiment/examples/tiny-search/src/tiny-search.js
```diff
@@ -0,0 +1,79 @@
...
+ * - score = total number of occurrences of all query terms in the document.
+ * - Results are ordered by score descending, ties broken by id ascending
+ *   (plain string comparison), capped at `limit` (default 10).
+ * - add() with an existing id replaces that document.
+ */
+
+export function tokenize(text) {
+  return text
+    .toLowerCase()
+    .split(/[^a-z0-9]+/)
```
dam-code-guardian: 🟡 **Warning:** The header contract says `tokenize()` splits on "any run of non-alphanumeric characters", but `[^a-z0-9]+` is ASCII only. `Café` tokenizes to `caf` and `e`, and Cyrillic or CJK text vanishes. A campaign optimizes against this contract, so an imprecise one aims the work at the wrong target.

**Fix:** State the ASCII-only scope in the contract, or switch to `/[^\p{L}\p{N}]+/u` and pin the choice with a test case.

packages/agents/claude-code/dam-skills/dam-experiment/examples/tiny-search/src/tiny-search.js
```diff
@@ -0,0 +1,79 @@
...
+  #docs = [];
+
+  add(id, text) {
+    if (typeof id !== "string" || id.length === 0) {
+      throw new TypeError("id must be a non-empty string");
+    }
+    if (typeof text !== "string") {
+      throw new TypeError("text must be a string");
+    }
+    this.remove(id);
```
dam-code-guardian, line 22: 🟢 **Suggestion:** `add()` calls `remove()`, whose `findIndex` is linear, so building the corpus is quadratic. That is a second slowness source the header does not name, and it also loads the `index-heavy` guard metric — a candidate can win on `index-heavy` by fixing this alone, without touching search.

packages/agents/claude-code/dam-skills/dam-experiment/examples/tiny-search/src/tiny-search.js
```diff
@@ -0,0 +1,79 @@
...
+    this.#docs.splice(index, 1);
+    return true;
+  }
+
+  get size() {
+    return this.#docs.length;
+  }
+
+  search(query, { limit = 10 } = {}) {
+    const terms = tokenize(query);
```
dam-code-guardian, line 41: 🟢 **Suggestion:** `limit` reaches `slice()` unvalidated, so `limit: -1` silently drops the highest-scoring result and `NaN` returns nothing — while `add()` does validate its arguments. Consistent validation at the entry point keeps a caller mistake from looking like a legitimate empty result.

packages/agents/claude-code/dam-skills/dam-experiment/examples/tiny-search/bench/bench.mjs
```diff
@@ -0,0 +1,150 @@
...
+  return Number(sortedNs[index]) / 1000;
+}
+
+function parseArgs(argv) {
+  const args = { scenario: "query-heavy", n: 5000, ops: 400, seed: 1 };
+  for (let i = 2; i < argv.length; i += 2) {
+    const key = argv[i].replace(/^--/, "");
+    const value = argv[i + 1];
+    if (key === "scenario") args.scenario = value;
+    else if (key === "n" || key === "ops" || key === "seed") {
```
dam-code-guardian: 🟡 **Warning:** `Number(value)` is unvalidated, so `--n` with a missing or non-numeric value becomes `NaN`; the build and operation loops then run zero iterations and the bench prints a well-formed JSON line whose latency fields are all `null`. A campaign reads that line as a measurement.

**Fix:** A measurement harness rejects bad input at parse time. Validate each numeric argument as a finite positive integer and exit non-zero with a message otherwise.

packages/agents/claude-code/dam-skills/dam-experiment/examples/tiny-search/bench/bench.mjs
```diff
@@ -0,0 +1,150 @@
...
+    const results = index.search(query);
+    samplesNs.push(process.hrtime.bigint() - start);
+    checksum = foldChecksum(checksum, results);
+  } else {
+    const id = `doc-${String(Math.floor(rand() * n)).padStart(6, "0")}`;
+    const text = makeDoc(rand);
+    const start = process.hrtime.bigint();
+    index.add(id, text);
+    samplesNs.push(process.hrtime.bigint() - start);
+    checksum = (Math.imul(checksum, 31) + index.size) | 0;
```
dam-code-guardian, line 124: 🟡 **Warning:** The write branch folds only `index.size`. Every generated id comes from `doc-000000`..`doc-{n-1}`, which is exactly the corpus `add()` already filled, so `add()` always replaces and `size` never moves. The `index-heavy` checksum is therefore a pure function of the operation count and cannot detect a behavior change — while `README.md:72` presents the checksum as the anti-cheat guard.

**Fix:** A behavior checksum folds values the implementation can change. Fold observable state for writes — for example the result of a verification `search()` after the write — never a quantity that is invariant by construction.

packages/agents/claude-code/dam-skills/dam-experiment/examples/tiny-search/bench/bench.mjs
```diff
@@ -0,0 +1,150 @@
...
+  }
+}
+
+samplesNs.sort((a, b) => (a < b ? -1 : a > b ? 1 : 0));
```
dam-code-guardian, line 134: 🟡 **Warning:** With no recorded sample, `totalNs / samplesNs.length` divides by zero and `percentileUs` indexes `sortedNs[-1]`, so `p50_us`, `p95_us`, `mean_us` and `ops_per_sec` all become `NaN`. `JSON.stringify` writes `NaN` as `null`, so the failure is invisible in the output line. This is reachable from the `NaN` argument path above.

**Fix:** Guard the statistics against an empty sample set: stop with an explicit error when no operation was measured, and never emit a numeric field that can serialize to `null`.

packages/experiment-sdk/src/experiment_sdk.py
```diff
@@ -686,14 +705,28 @@ def _emit(self, event: dict[str, Any], flush: bool = False) -> None:
...
             return
         if not force and len(self._buffer) < _EVENT_FLUSH_MAX:
             return
-        events, self._buffer = self._buffer, []
-        _request(
-            "POST",
-            f"/experiments/{self._experiment_id}/events",
-            {"events": events},
-        )
+        events = list(self._buffer)
```
dam-code-guardian: 🔴 **Critical:** The buffer now retains events until a POST lands, but nothing bounds what it retains, and the server caps a batch at 500 (`appendEventsRequestSchema`: `z.array(traceEventSchema).min(1).max(500)`). During an outage `events = list(self._buffer)` keeps growing; past 500 the POST is answered 400 by Zod, and 400 is not in `_TRANSIENT_HTTP {500,502,503,504}`, so it raises, the `except Exception` arm keeps everything buffered, and every later flush is larger. The reporting path is then wedged for the rest of the run — the outage the retry exists to ride out becomes permanent event loss, which inverts the docstring's promise.

**Fix:** Bound both sides of the retry: send at most the server's cap per flush, and bound what the buffer retains (drop or coalesce the oldest beyond the bound, and log the drop). A retained batch must always still be an acceptable payload.

packages/experiment-sdk/src/experiment_sdk.py
```diff
@@ -686,14 +705,28 @@ def _emit(self, event: dict[str, Any], flush: bool = False) -> None:
...
+                f"/experiments/{self._experiment_id}/events",
+                {"events": events},
+                retry=True,
+            )
+        except ExperimentClosed:
+            raise
+        except Exception as err:  # noqa: BLE001 — kept buffered; next emit retries
+            _log(f"event report failed, {len(events)} event(s) kept for retry: {err}")
+            return
+        self._buffer = self._buffer[len(events) :]
```
dam-code-guardian, line 763: 🟡 **Warning:** `_last_flush` is bumped only on success, so `stale` stays true for the whole outage and every single `_emit` runs a fresh four-attempt ladder — 1s + 2s + 4s = 7s of `time.sleep` per emitted event, on the caller's thread (usually `Span.__exit__`). The change protects the loop from dying and then charges it the latency instead.

**Fix:** Record the time of a failed flush too, and suppress further attempts until a back-off interval passes, so one outage costs one ladder rather than one ladder per event.

packages/experiment-sdk/src/experiment_sdk.py
```diff
@@ -118,20 +118,31 @@ def _config() -> tuple[str, str]:
     return f"{base}/api/agents/{agent_id}", agent_id
 
 
-_TRANSIENT_HTTP = {502, 503, 504}
+# 500 is transient here too: the api-server returns it while its own
+# dependencies (postgres, redis) restart under it, which is exactly the
+# outage the retry exists to ride out.
+_TRANSIENT_HTTP = {500, 502, 503, 504}
```
dam-code-guardian, line 130: ✅ **Looks good:** Adding 500 to the transient set is safe for the opted-in writers, and I verified the idempotency the docstring relies on rather than taking it: `insertSpan` is `onConflictDoNothing`, `endSpan` updates by span id, and `finish` answers 409 through `canTransition`, which `_request` maps to `ExperimentClosed`. Keeping `spawn` on a single attempt is the right call.

packages/agents/claude-code/dam-skills/dam-experiment/SKILL.md
```diff
@@ -388,6 +401,42 @@ iterations, a narrower question. The TTL still gets generous slack over
 whatever the estimate ends up being (a killed working pod wastes everything);
 the hour budgets the *work*, not the deadline.
 
+### You do not write the campaign's arms — but you do pin its design
```
dam-code-guardian, line 419: 🟡 **Warning:** This 36-line section on `campaign.yaml` design control, and the nine-line host-CPU-starvation note in `deploy/lima-k3s.yaml`, are not in the PR description, which claims only that SKILL.md gained references to the new starter. Both additions read as correct and worth keeping — the description is what is behind.

**Fix:** List every subsystem the change touches in the PR description, so a reviewer does not find unannounced edits outside the declared scope.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `614e47a`

## PR #3444: feat(dam-experiment): add tiny-search example and update documentation
**Author:** Tomas2D | **Branch:** feat/experiments-examples-hardening → main | **Changes:** +823 −37 (13 files)

### Summary
The PR adds a second optimization-campaign starter (`tiny-search`) with a behavioral suite and a deterministic benchmark, documents it in the skill, and makes the experiment SDK's write path survive a transient api-server outage. The SDK half is verified: `insertSpan` is `onConflictDoNothing` and `endSpan` updates by id, so the docstring's idempotency claim that authorizes `retry=True` holds.

### Findings
- 🔴 **Critical:** The two new example files add untyped prose comments, which the repo convention forbids (`packages/agents/claude-code/dam-skills/dam-experiment/examples/tiny-search/src/tiny-search.js:1`)
  **Fix:** Comments in `.js`/`.mjs` source carry a registered `COMMENT_TYPES` prefix or do not exist; move the contract prose to `README.md`, which is already the right home for it.
- 🔴 **Critical:** A long outage wedges the event buffer permanently, because the retained batch grows past the server's 500-event cap (`packages/experiment-sdk/src/experiment_sdk.py:718`)
  **Fix:** Cap what one flush sends and what the buffer retains, so a retained batch always stays inside the server's accepted size.
- 🟡 **Warning:** During an outage every `_emit` pays the full retry ladder, so reporting latency is charged to the loop it was meant to protect (`packages/experiment-sdk/src/experiment_sdk.py:731`)
  **Fix:** After a failed flush, suppress the next attempts for a back-off interval, so one outage costs one ladder and not one ladder per event.
- 🟡 **Warning:** The `index-heavy` checksum folds only `index.size`, which is invariant by construction, so the guard metric detects no behavior change (`packages/agents/claude-code/dam-skills/dam-experiment/examples/tiny-search/bench/bench.mjs:129`)
  **Fix:** A behavior checksum folds values the implementation can change; fold a verification query's result after each write.
- 🟡 **Warning:** `--n`, `--ops` and `--seed` are converted with bare `Number()`, so a bad value becomes `NaN` and the bench prints `null` metrics instead of failing (`packages/agents/claude-code/dam-skills/dam-experiment/examples/tiny-search/bench/bench.mjs:89`)
  **Fix:** A measurement harness rejects bad input at parse time; validate each numeric argument as a finite positive integer and exit non-zero otherwise.
- 🟡 **Warning:** The documented tokenizer contract says "any run of non-alphanumeric characters", but `[^a-z0-9]+` is ASCII only (`packages/agents/claude-code/dam-skills/dam-experiment/examples/tiny-search/src/tiny-search.js:22`)
  **Fix:** State the ASCII-only scope in the contract, or use `/[^\p{L}\p{N}]+/u` and pin the choice in the suite. A campaign optimizes against this contract, so an imprecise one is a wrong target.
- 🟢 **Suggestion:** `search()` passes `limit` straight to `slice()` with no validation, so a negative limit silently drops the best results while `add()` validates its own arguments (`packages/agents/claude-code/dam-skills/dam-experiment/examples/tiny-search/src/tiny-search.js:52`).
- 🟢 **Suggestion:** `add()` calls the linear `remove()`, so corpus build is quadratic — a second slowness source the header does not name, and it also loads the `index-heavy` guard metric (`packages/agents/claude-code/dam-skills/dam-experiment/examples/tiny-search/src/tiny-search.js:36`).
- ✅ **Looks good:** `finish()` catching `ExperimentClosed` is correct, not a swallowed error — the server answers 409 from `canTransition`, so a landed-then-retried finish and a user Stop are the same observable state.

- 🟡 **Warning:** `docs/architecture/experiments.md` is byte-identical to `main` while this PR changes three of its documented promises: which calls retry, that a rejected event batch is now retained, and that `finish` has a no-op path. The page's Completion and liveness section still states the `running → completed | failed` flip unconditionally (`docs/architecture/experiments.md:243`)
  **Fix:** When a change alters a client's retry contract or a terminal transition's conditions, update the owning architecture page in the same PR. State the rule semantically — idempotent writes retry, `spawn` never does — rather than transcribing status codes.

### Issue Fit
- 🟡 **Warning:** Two substantial changes are undeclared: the 36-line `campaign.yaml` design-control section in `SKILL.md`, and the nine-line host-CPU-starvation note in `deploy/lima-k3s.yaml`, a subsystem the description never mentions (`packages/agents/claude-code/dam-skills/dam-experiment/SKILL.md:404`)
  **Fix:** List every subsystem the change touches in the PR description, so a reader does not find unannounced edits. Both additions read as correct and worth keeping — only the description is behind.
- 🟢 **Suggestion:** `commands/experiment-onboard.md` is rewritten for two starters and is also undeclared; the same description line covers it.

The five findings above at `bench.mjs:89`, `bench.mjs:129`, `bench.mjs:133`, `tiny-search.js:22`, `tiny-search.js:77` were reported here and are merged into `### Findings`. Additionally:
- 🟡 **Warning:** With zero recorded samples the metric block divides by zero and indexes `sortedNs[-1]`, so every latency field serializes to `null` (`packages/agents/claude-code/dam-skills/dam-experiment/examples/tiny-search/bench/bench.mjs:133`)
  **Fix:** Stop with an explicit error when no operation was measured; never emit a numeric field that can serialize to `null`.
- 🟢 **Suggestion:** `percentileUs` uses `Math.floor((p/100)*length)`, one rank above the usual nearest-rank position — for 400 samples p50 is the 201st value.
- 🟢 **Suggestion:** `tokenize()` and `search()` do not validate their input, so a non-string query fails inside `toLowerCase` with an obscure `TypeError`.
- 🟢 **Suggestion:** The limit test covers the default and two positive values only, so `limit: 0`, a negative limit, and accented input can change behavior with the suite green.

### Verdict
REQUEST_CHANGES — the comment-convention violation fails `common:check:comment-types` in CI, and the event buffer wedges permanently once a retained batch passes the server's 500-event cap.

---

Tomas2D:
Rebased onto `main` and addressed the review. Per finding:

**🔴 Untyped prose comments** — already resolved in a follow-up commit before this push; the example sources carry no comments and `common:check:comment-types` passes locally.

**🔴 Retained batch passes the 500-event cap** — confirmed and fixed. The buffer now retains at most `_EVENT_BATCH_MAX` (500, mirroring `appendEventsRequestSchema`), enforced in `_emit`, so the invariant holds globally rather than only on the failure path. Overflow drops the oldest events and reports the count — a bounded, visible loss instead of a permanent wedge.

**🟡 Full retry ladder per emit during an outage** — fixed. A failed flush parks further attempts for a back-off window, so an outage costs one ladder per window rather than one per event. `finish` passes `ignore_backoff` to buy the tail one last attempt. The existing redelivery test was updated to match the new timing contract.

**🟡 `index-heavy` checksum folds an invariant** — confirmed and fixed; it now folds a verification query's results after each write. Verified empirically: reversing the tie-break rule changes the checksum (`1530044173` → `2134533566`), where the old `index.size`-only fold was identical. The checksum is stable across runs at a fixed seed while timings vary.

**🟡 Bare `Number()` on numeric args** — fixed; each is validated as a positive integer. `--n abc`, `--ops 0`, `--seed -3`, `--n 1.5` and a valueless `--scenario` all exit non-zero instead of printing `null` metrics.

**🟡 Empty sample set serializes to `null`** — subsumed by the above: `ops` must be a positive integer, so the sample set is never empty.

**🟡 Tokenizer contract is ASCII-only** — took the Unicode option: `/[^\p{L}\p{N}]+/u`. The README states the rule and the suite pins it with accented input.

**🟡 Architecture page unchanged** — `docs/architecture/experiments.md` now records the reporting contract semantically (idempotent writes retry, `spawn` never does; a rejected batch is retained within the accepted size; a failed flush backs off; `finish` is a no-op on an already-terminal run), with `Last verified` bumped.

**🟡 Undeclared subsystems** — the PR description now lists every subsystem touched, including the `campaign.yaml` section, the `lima-k3s.yaml` note, and the `experiment-onboard.md` rewrite.

**🟢 Suggestions** — `limit` is validated (non-negative integer), `tokenize()`/`search()` reject non-string input, the percentile uses nearest-rank, and the quadratic corpus build is now named in the README. The suite gained `limit: 0`, invalid limits and accented input; 18 tests pass.

Also unrelated to the review: this branch had dropped the `tools.java.*` platform entries from `mise.lock` as collateral of the controller commit (a local mise run rewrites the lockfile with only the platforms it resolved on the host). Restored from `main`.

The earlier `mise e2e` failure was `cluster:install` failing to load images (`reference does not exist`), not a test failure — unrelated to this branch's contents. The rebase picks up the image-pipeline change that has since landed on `main`.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `788f7d2`

## PR #3444: feat(dam-experiment): add tiny-search example and update documentation
**Author:** Tomas2D | **Branch:** feat/experiments-examples-hardening → main | **Delta:** +468 −101 (13 files, 8 commits since `67bdf75`)

### Summary
The fix round closes all thirteen findings from the previous review. The branch was rebased onto new `main`, so the prior review's SHA is gone — I confirmed the prior scope by patch content (`67bdf75` is byte-identical to it) and reviewed only `67bdf75..HEAD`.

### Changes since last review
Previous HEAD: `614e47a` (2026-08-24T19:40:02Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** Untyped prose comments removed from both example files; the contract prose moved to `README.md` (`examples/tiny-search/src/tiny-search.js`, `bench/bench.mjs`)
- ✅ **Fixed:** Event buffer bounded to `_EVENT_BATCH_MAX = 500`, matching the route's cap, so a retained batch stays acceptable (`experiment_sdk.py:711`)
- ✅ **Fixed:** A failed flush parks later attempts for `_EVENT_RETRY_BACKOFF_S`; `finish` passes `ignore_backoff` for the tail (`experiment_sdk.py:738`)
- ✅ **Fixed:** The `index-heavy` checksum now folds a verification query's results, not only the invariant `index.size` (`bench/bench.mjs:120`)
- ✅ **Fixed:** `--n`/`--ops`/`--seed` validated by `positiveInt()`, which exits non-zero on bad input (`bench/bench.mjs:63`)
- ✅ **Fixed:** Tokenizer is `[^\p{L}\p{N}]+/u`, and the README states the Unicode rule (`src/tiny-search.js:7`)
- ✅ **Fixed:** The empty-sample division and `sortedNs[-1]` index cannot be reached — `parseArgs` now requires `ops >= 1` (`bench/bench.mjs:63`)
- ✅ **Fixed:** `docs/architecture/experiments.md` records all three moved promises and bumps `Last verified` (`docs/architecture/experiments.md:241`)
- ✅ **Fixed:** The PR body enumerates every subsystem, including the `campaign.yaml` section, the lima-k3s note and the `experiment-onboard.md` rewrite
- ✅ **Fixed:** `limit` and `tokenize`/`search` inputs validated; `percentileUs` uses nearest-rank `Math.ceil`; the quadratic corpus build is named in the README

### Findings
- 🟢 **Suggestion:** The `\p{L}\p{N}` class excludes combining marks, so decomposed input splits mid-word — `tokenize("café".normalize("NFD"))` returns `["cafe"]` and an NFD document does not match an NFC query. A `.normalize("NFC")` before the split would make the README's Unicode rule hold for either form (`examples/tiny-search/src/tiny-search.js:7`).
- 🟢 **Suggestion:** The verification query runs a full corpus scan per add, so `index-heavy` costs about 13× more wall clock — measured 3.6 s versus 0.27 s at `--n 2000 --ops 300`, with the add-path p50 at 26 µs versus 17.6 µs. Sampling it every Nth op would keep the guard and the documented `--ops 2000` runtime (`examples/tiny-search/bench/bench.mjs:120`).
- 🟢 **Suggestion:** `positiveInt` also rejects `--seed 0`, though `mulberry32` seeds from 0 correctly (`examples/tiny-search/bench/bench.mjs:81`).

✅ No findings.

### Issue Fit
- 🟢 The body declares a `mise.lock` subsystem, but `mise.lock` is identical to the base — `9018e72` restored exactly what `b5d1b02` dropped, so the branch delivers no net lockfile change. An over-declaration of an intra-branch fixup, not missing scope.

The NFD-normalization and bench-cost items above were reported here and are merged into `### Findings`. Both are residuals of this round's own fixes, so they are priced as suggestions.

### Verdict
APPROVE — every blocking finding is closed at its cause, and the residual is three suggestions.

---

Issue #3432, opened 2026-08-24
UI - Show session artifacts on the Home feed cards


kapetr:
## Problem

The Home feed shows one card for each session. The card does not show the artifacts of the session. The user must open the session or the Artifacts page to find them.

The system cannot supply this data today. The artifact record shows the sandbox that owns the artifact. It does not show the session. The artifact version record holds no attribution: the system keeps the version number, the size, the content type, and the time, but no record of the session that made the version.

One more condition makes this difficult. One session can make a new artifact, and one session can also change an artifact that a different session made. A change makes a new version of the same artifact. Thus more than one session can have a relation to the same artifact.

## Desired outcome

The card of a session shows the artifacts of that session. The user opens an artifact from the card.

## Design reference

Branch `design/home-prototype`. File `packages/ui/src/modules/home/views/comparison-view.tsx`. Find the `artifact` property of the card component. The prototype shows one artifact for each card.

## Related

Home page: #3317, PR #3369.

Issue #3307, opened 2026-08-13
Live updates for pod-sourced state — sessions and files still poll


jezekra1:
## Context

The live-updates work (#3170 / PR #3303) replaced polling with pushed invalidation hints for every domain whose changes pass through the api-server or are observable from the cluster API. Two surfaces were deliberately left polling: the session list (with per-session working/idle status) and the in-pod file tree. Their truth lives inside the agent pod — a harness creates a session or writes a file without the platform seeing a write — so there was no server-side moment to emit an event from, and giving them one means new protocol surface between the agent pod and the platform.

## Problem

The chat view still asks each agent for its session status on a timer, and the file panel re-reads the workspace listing every couple of seconds while open. Beyond the residual chatter, these are now the only UI surfaces that lag behind reality instead of following it — a new session started by a schedule, or files produced by a running task, appear on the next poll tick rather than when they happen.

The design has to answer more than transport:

- the pod should only produce these signals **while someone is watching** — an unobserved agent shouldn't watch its own filesystem to report to nobody
- workspace changes during a busy task can fire hundreds of times a second, so pod-side coalescing is required
- pods are ephemeral and their platform link reconnects, so per the platform's eventing contract these signals stay advisory — a lazy poll (or reconnect refresh) likely remains as the loss-bounding fallback rather than being deleted outright

## Goal

Session and file panels follow the pod live, like every other surface now does: a new session or a changed workspace shows up when it happens, the steady-state polls are gone, and an unobserved agent produces no signal traffic.

## Done when

- The agent pod can report "sessions changed" / "workspace changed" to the platform, produced only while a subscriber is attached and coalesced under load.
- The UI's session list and file panel refresh from those signals; their fixed-interval polling is removed or demoted to a slow loss-bounding fallback.
- An idle, unwatched agent generates no reporting traffic.

PR #3451, opened 2026-08-25
docs(adr): notifications as a per-owner attention record

kapetr:
A draft ADR for how notifications and the Home feed should be stored.

Home ships today, but it reads sessions from each running agent's pod on a timer and keeps dismissals in browser local storage. So a hibernated agent shows nothing, dismissals do not follow you to another device, and a scheduled run that finished overnight leaves no server-side record at all.

ADR-055 removed the Postgres sessions table because every consumer either spoke ACP to the agent or ran in the pod. Home is the first consumer that breaks that premise. This draft keeps 055 intact by storing attention rather than session state: the agent still answers which sessions exist, and a missing row means nobody was told, not that nothing happened.

Proposed, not accepted. It depends on the pod-to-platform reporting surface in #3307, and on session-scoped MCP attribution before artifacts can appear on a card.

Refs #3317, #3100, #3307, #3432


jezekra1:
The #3307 dependency this draft names has landed — #3450, in review — and it took a different shape than the draft assumes. I think the shape it took simplifies the producer half of this record, so here is the design we converged on for the hibernated-sessions side, mapped onto this draft's decisions.

**What #3450 actually ships as the pod-to-platform surface.** The pod serves `sessions.list` (the composed union of the harness's list and the pod's metadata store — including sessions the harness hasn't persisted yet) and `sessions.watch` (bare notices meaning "re-read", produced only while a subscriber is attached, with an opening notice on every (re)subscribe) over the pod's tRPC router, now exposed over WebSocket. Notices never carry state; every consumer re-reads. Home consumes this today through an owner-scoped subscription (`events.podSessions`) behind which the api-server holds one pod stream per running agent of each watching owner.

**Proposed producer: pull on notice, instead of a new pod-push report path.**

A lease-elected, always-on watcher in the api-server (sibling of the existing lease-elected Agent watch, driven by the same wake/hibernate transitions):

1. holds `sessions.watch` to every *running* agent install-wide;
2. on each notice, debounced, calls `sessions.list` and upserts the record row (no-op guard, harness-config-snapshot style);
3. emits a domain event, which the live-hints saga projects as an ordinary `{topic: "sessions", agentId}` hint on `events.owner`.

That last step is the payoff: once there is a server-side write moment, sessions stop being special. `events.podSessions` (which #3450 introduces) gets retired for the standard hint pipeline, Home reads one owner-scoped query for running *and* hibernated agents, and the remaining per-agent passive ACP reads on the feed go away.

**How this maps onto the draft's decisions:**

- *Agrees, and keeps:* attention-record-not-mirror (the row is never consulted for which sessions exist — the pod's `sessions.list` stays the only answer to that); "in progress is never stored" (a capture forces `running` false; a dead pod can't leave phantom work); hint-plus-read as the one code path — this design just applies it on the pod side too, not only toward browsers.
- *Replaces:* "the pod carries the payload when it reports", the retry loop, and the shutdown flush. The draft wants payload-push because "a report that only says something changed would race with the pod shutting down." Under pull, that race is bounded by the idle timeout: a turn's notice fires within ~250ms and the read completes immediately, while hibernation follows minutes later — so the hibernation case (the case the feed exists for) loses nothing. A hard kill right after a turn loses the last item, which is exactly the bound the draft already accepts ("a hard kill may lose the last item"). Net: same loss bound, no new pod→platform route, no shutdown-flush machinery, and the pod side of #3450 is consumed verbatim.
- *Defers to this draft:* seen/dismissed as per-user watermarks derived from relay presence is strictly better than the pod's per-session `seenAt` (it fixes the single-driving-user assumption `platform-topology` flags), and retiring the pod's read state composes cleanly with #3450 — the watch notifies on any metadata write today, and fewer writes just means fewer notices.
- *One honest cost to write into the ADR:* an always-on watcher re-scopes #3307's "an unobserved agent generates no reporting traffic." An idle agent still emits nothing (watches are silent absent change), but a busy agent now reports to the platform whether any human is watching — justified because recording is the point, but it should be said out loud rather than inherited silently.
- *Why the lease appears here and not in #3450:* a demand-driven holder needs no lease (duplication across replicas is a redundant socket; notices are idempotent). A row-writing producer is the first thing where exactly-once matters. #3450 deliberately left the lease out for that reason — this record is the change that earns it, and an always-on watcher makes it the trivial `createLeaderLease` kind, with failover healed by re-reading every pod and letting the no-op guard suppress the unchanged.

Housekeeping: the draft's header says ADR-084 — that number is taken by #3450's accepted record (pod-owned live updates), so this becomes 085 on promotion.

jezekra1 (approved):

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `fd72c03`

## PR #3451: docs(adr): notifications as a per-owner attention record
**Author:** kapetr | **Branch:** docs/notification-record → main | **Changes:** +129 −0 (2 files)

### Summary
Adds a proposed draft ADR that keeps a per-owner attention record in the platform, written by a lease-elected watcher that pulls on ADR-086 pod notices, with seen and dismissed as per-user watermarks, and regenerates the ADR index. The record is scoped to sessions and never answers which sessions exist, which keeps ADR-055 intact.

### Findings
- 🟡 **Warning:** The record claims to be the store the notifications epic needs, but the Decision gives no rule for any notification class except sessions. Approvals are excluded by rule (line 47), and crashes, expired credentials and over-budget agents get neither a rule nor an open question, so `a notification centre, channel push, and the floating pill read one source rather than three` is false for every approval-carrying surface: `packages/ui/src/components/floating-approvals-pill.tsx` reads the approvals table for content and would keep doing so. #3100 also names user control over what notifies them, which the draft does not decide and does not list as open (`docs/adrs/DRAFT-notification-record.md:103`, also `docs/adrs/DRAFT-notification-record.md:121`). Reported by the diff review and by Issue Fit.
  **Fix:** Scope the claim to what the Decision rules on — the session record — or, for each notification class the linked epic enumerates, add its rule to the Decision or its unresolved question to Open Questions. A Consequences claim never stands in for the rule that makes it true.
- ✅ **Looks good:** `docs/adrs/index.md` is generated, and the added row matches its generator: `node scripts/adr-index.mjs --check` and `node scripts/adr-immutable.mjs --merge-base` both pass at this HEAD.

✅ No findings.

### Issue Fit
- 🟢 **Suggestion:** The draft narrows an acceptance criterion of #3307 — `An idle, unwatched agent generates no reporting traffic` — because the always-on watcher makes a busy agent report with no human watching. The record states this under Consequences → Harder, but the PR body names #3307 only as a dependency, so a reader of the PR alone does not learn that a shipped promise is re-scoped (`docs/adrs/DRAFT-notification-record.md:104`).
- The #3100 scope finding this skill reported is folded into `### Findings` above.

### Verdict
COMMENT — the draft is internally sound on its session decision, but one Consequences claim states an epic-wide scope the Decision does not rule on.


---

PR #3450, opened 2026-08-25
Live updates for pod-sourced state — sessions and files

jezekra1:
The session list and the Files panel stop lagging behind reality. A session started by a schedule, and a file produced by a running task, appear when they happen instead of on the next poll tick. An agent nobody is looking at produces no reporting traffic at all.

Five polls against pods are gone: the session list every 5s, the workspace tree every 2s, the open file's whole contents every 2s, the Home feed listing sessions for every running agent every 15s, and the 3s reachability probe — connection state now carries that signal, and the subscription's own reconnect backoff does the probing.

Two things fixed along the way that were not about polling. A session the pod had recorded but the harness had not yet persisted was never listed at all — which is why a schedule-fired session was invisible rather than merely late. And the file panel's readiness check patched the Agent record every two seconds per open panel; since the controller stopped reconciling on every resync, that churn had become the dominant steady-state reconcile load.

The decision is recorded as ADR-086.

Closes #3307

## Sub-issues

- [x] 01 — Pod tRPC over WebSocket, and the agent relay
- [x] 02 — Per-agent UI client moves to WebSocket
- [x] 03 — `sessions.list` on the pod router
- [x] 04 — Session Watch
- [x] 05 — Workspace tree Watch
- [x] 06 — Open-file Watch
- [x] 07 — Home feed on an owner subscription
- [x] 08 — Architecture docs

PR #3452, opened 2026-08-25
docs(adr): pod-owned live updates for session and workspace state

jezekra1:
ADR-084. The session list and the Files panel are the last surfaces that poll, because their truth lives inside the agent pod and there was no server-side moment to emit from. This records how they stop polling.

**Decision.** Pod-owned state — the session list, the workspace directories a client has open, and the open file — is read and observed over the agent-runtime's own tRPC surface, exposed over WebSocket and relayed by the api-server, rather than over ACP or on a timer. A Watch exists only while a subscriber is attached and emits notices carrying a topic and ids, never state, so an unobserved agent produces no traffic.

Also adds a cross-cutting **Live Updates** section to `docs/ubiquitous-language.md`, which had no entry for any of this vocabulary. Three words that sound interchangeable are not, and the glossary now says which delivery guarantee each one carries: **Domain Event** (in-process, at-most-once), **Signal** (thin cross-replica Redis message), **Live Event** (per-owner browser invalidation notice). Plus Topic, Sync, and the two new terms, Watch and Snapshot.

## What it amends

- **ADR-055** — narrows its committed-to `_meta.platform.*` wire format to *writes*. That choice assumed every reader already held an ACP connection; a per-agent tRPC relay makes the premise false. The agent remains the sole source of truth for session existence, and there is still no server-side session store.
- **ADR-049** — its "per-directory cache keying as the unit of subscription" is honoured rather than changed: a Watch covers exactly the open directories.
- **ADR-083** — layering unchanged, origin extended: an invalidation notice may now originate in a pod, not only from a domain event in the api-server.

No `supersedes` field: ADR-055's core holds, so marking it superseded wholesale would be false. It uses an `Amends` section, following ADR-055's own precedent with ADR-019.

Implementation lands separately in #3450, which is decomposed into eight sub-issues. Refs #3307.

### DIFF
--- a/docs/adrs/084-pod-owned-live-updates.md
+++ b/docs/adrs/084-pod-owned-live-updates.md
@@ -1,0 +1,55 @@
+---
+id: 084
+title: Pod-owned live updates over the agent's own tRPC surface
+status: accepted
+subsystem: platform-topology
+tags: [live-updates, sessions, workspace, acp, hibernation]
+summary: Pod-owned state is read and watched over the agent-runtime's own tRPC surface relayed over WebSocket; a Watch lives only as long as its subscriber and emits topic-plus-ids notices, never state.
+---
+
+# ADR-084: Pod-owned live updates over the agent's own tRPC surface
+
+**Date:** 2026-08-25
+**Status:** Accepted
+**Owner:** @jezekra1
+
+## Context
+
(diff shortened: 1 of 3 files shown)


jezekra1:
Duplicate — ADR-084 is already in #3450 along with the plan and implementation. Consolidating there.

PR #3456, opened 2026-08-25
Session artifacts on the Home feed cards

### LINKED ISSUE #3432: UI - Show session artifacts on the Home feed cards

kapetr:
A Home feed card shows the artifacts its session produced, as chips that open the preview dialog.

## How attribution works

Nothing in the platform links an artifact to a session: the agent calls the platform's tools over MCP, and that endpoint only ever identifies the agent. Model-supplied session ids, per-session MCP addresses, and inferring from the running turn were all considered and rejected — each can attribute wrongly or depends on harness behaviour we don't control.

Instead, the artifact tools mark their results, and the runtime spots the marker in the session's tool-call stream — frames it already proxies, each carrying its session id. Deterministic, model-free, and every failure mode leaves a chip missing, never a chip on the wrong card. Terminal sessions stay unattributed — they bypass the protocol this reads.

## Storage note for reviewers

An artifact's version history now includes the **current** version — creation writes the first row, every revision writes its own, and the version records the session that produced it. Existing artifacts are backfilled. This removed the special-casing of the current version in the version list, the share viewer and delete. Follow-up: normalize the artifact table — the per-version columns it still carries now duplicate the versions table.

## Also in this PR, on purpose

- A feed card stays until dismissed instead of vanishing when its session is seen; unread styling still follows seen state. Finished sessions older than 7 days age out of the feed.
- A background read (the feed listing sessions) no longer marks a session seen — passive connections now reach the pod as non-viewers.

Closes #3432

## Sub-issues

- [x] 01 — Touch record, its two doors, and the tool marker
- [x] 02 — Runtime reports artifact touches
- [x] 03 — Chips on the feed card


packages/ui/src/modules/home/components/feed-list.tsx
```diff
@@ -78,12 +85,17 @@ export function FeedList({
...
             working={item.kind === "in-progress"}
-            unread={item.kind === "unread"}
+            unread={item.kind === "unread" && isUnreadSession(session)}
             onOpen={() => onOpenSession(item.agentId, session.sessionId)}
             onDismiss={
               item.kind === "unread" ? () => onDismiss(item) : undefined
             }
-          />
+          >
+            <FeedArtifactChips
```
dam-code-guardian, line 94: 🔴 **Critical:** The chips are the first interactive children the card wraps, and the card's container `onKeyDown` (`feed-card.tsx:35`) runs for keys pressed inside it. Enter or Space on a focused chip therefore calls `preventDefault()` and opens the session, which both suppresses the chip's own activation and navigates away — so the chips cannot be operated by keyboard at all. The handler predates this PR, but this PR is what puts a keyboard-operable control inside the card.

**Fix:** A container-level activation handler on a `role="button"` wrapper must ignore events that did not start on the container — `if (event.target !== event.currentTarget) return;` at the top of `onKeyDown`. `stopPropagation` on the child's `onClick` covers pointer events only, never keyboard ones.

packages/ui/src/modules/home/views/home-view.tsx
```diff
@@ -40,8 +49,22 @@ export function HomeView() {
...
+    null,
+  );
+  const { data: previewArtifact } = useArtifact(previewArtifactId);
+
+  const artifactsFor = (item: FeedItem): readonly ArtifactTouched[] => {
+    if (item.kind !== "unread") return EMPTY_ARTIFACTS;
+    const touched = artifacts.bySession.get(item.session.sessionId);
+    if (!touched || touched.length === 0) return EMPTY_ARTIFACTS;
+    const cleared = dismissedAt(item.agentId, item.session.sessionId);
+    if (cleared === null) return [touched[0]];
```
dam-code-guardian: 🟡 **Warning:** A session with no dismissal returns `[touched[0]]`, so the common case — a card seen for the first time — shows one arbitrary chip and never the `+N more` count, while a previously dismissed session shows every artifact newer than the clear. The feature works fully only after a dismissal. Reported by React UI Engineering, Issue Fit and TypeScript Engineering.

**Fix:** A "hide what was already cleared" filter is a no-op when there is no clear timestamp: return the whole list and let `FeedArtifactChips` do the capping with its own `SHOWN` and `+N more`.

```suggestion
    if (cleared === null) return touched;
```

packages/db/drizzle/0040_backfill_current_artifact_versions.sql
```diff
@@ -0,0 +1,9 @@
+-- Every artifact now keeps a version row for the version it is at, so the
+-- version list, content lookup and delete no longer special-case the current
+-- one. Rows written before that change start at the first superseded version,
+-- so give each artifact the row for its current version.
+INSERT INTO library_artifact_versions
```
dam-code-guardian, line 5: 🟡 **Warning:** Migrations run at api-server startup, so during a rolling update this backfill runs while old replicas still serve. An old replica's `insertArtifact` writes no version row, so every artifact it creates after the backfill keeps no row for its current version — permanently. The reads this PR simplifies then understate that artifact: `listVersions` omits the current version, and `versionCount` returns one less, so the share viewer hides its version navigation.

**Fix:** A backfill that establishes an invariant the new code writes must not depend on the old code being gone. Either keep the reads tolerant of a missing current-version row (derive it from the artifact row when absent), or run the backfill as a post-rollout step rather than inside the startup migration.

packages/db/drizzle/0038_green_lyja.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "library_artifact_versions" ADD COLUMN "session_id" text;
```
dam-code-guardian: 🟡 **Warning:** `session_id` becomes the predicate of the new feed read, which also orders by `created_at`, but the column arrives with no index and the table carries only its `(artifact_id, version)` primary key. Every Home feed poll — 30-second stale time, one query per agent — then scans the version table. Reported by TypeScript Engineering.

**Fix:** A column introduced as a query predicate ships with its index in the same migration: add an index on `(session_id, created_at desc)` to `library_artifact_versions`.

packages/api-server/src/modules/artifact-library/infrastructure/artifact-library-repository.ts
```diff
@@ -265,7 +303,71 @@ export function createArtifactLibraryRepository(
...
+                  ),
+                ),
+            ),
+          ),
+        )
+        .returning({ artifactId: versionsTable.artifactId });
+      return updated.length > 0;
+    },
+
+    async listTouches({ owner, agentId, sessionIds, limit }) {
```
dam-code-guardian, line 336: 🟡 **Warning:** The query takes up to 50 session ids but applies one global limit (200 by default) after ordering by `created_at desc`. One session that published many versions recently consumes the whole budget, and the other sessions' cards then render no chips. Reported by TypeScript Engineering.

**Fix:** A fan-out read bounds per key, not per query: rank rows per `session_id` with a window function (or issue one query per session) and cap each session's touches, so one key cannot spend another's budget.

packages/ui/src/modules/home/views/home-view.tsx
```diff
@@ -163,10 +186,18 @@ export function HomeView() {
                 }}
                 onResolved={(item, label) => sticky.keep(item, label)}
                 resolvedLabelFor={sticky.labelFor}
+                artifactsFor={artifactsFor}
+                onOpenArtifact={setPreviewArtifactId}
               />
             </>
           )}
         </div>
+        {previewArtifact && (
```
dam-code-guardian: 🟡 **Warning:** The dialog is mounted on the fetched artifact rather than on the selected id. A chip click therefore shows nothing while the fetch runs, and `useArtifact` has `retry: false`, so a failed fetch leaves `previewArtifactId` set with no dialog ever appearing — the chip looks dead, and clicking it again does nothing. Reported by React UI Engineering.

**Fix:** Mount a fetched-content dialog on the UI-owned selection state, render the pending and error states inside it, and clear the selection when the query fails.

packages/ui/src/modules/home/api/queries.ts
```diff
@@ -17,6 +20,59 @@ export const homeKeys = {
...
+
+export interface SessionArtifacts {
+  bySession: ReadonlyMap<string, readonly ArtifactTouched[]>;
+}
+
+export function useFeedArtifacts(items: readonly FeedItem[]): SessionArtifacts {
+  const wanted = useMemo(() => {
+    const byAgent = new Map<string, string[]>();
+    for (const item of items) {
+      if (item.kind !== "unread") continue;
```
dam-code-guardian, line 37: 🟡 **Warning:** The touch fetch and `artifactsFor` both skip anything that is not an `unread` item, so a running session shows no chips — the case where a just-written artifact matters most. Issue #3432 scopes the feature to session cards, and the PR body declares no such narrowing. Reported by Issue Fit.

**Fix:** Scope a feed feature by the surface the issue names: every session-backed card participates unless the issue or the PR body states the exclusion.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `bb92f23`

## PR #3456: Session artifacts on the Home feed cards
**Author:** kapetr | **Branch:** feat/3432-session-artifacts → main | **Changes:** +6484 −59 (41 files)

### Summary
The PR shows artifact chips on Home feed session cards. It attributes an artifact to a session by watching completed ACP tool-call output for a marker the platform MCP tool emits. It also moves the version history to hold the current version, with a backfill migration.

### Findings
- 🔴 **Critical:** Artifact chips are not reachable by keyboard, and a key press on a chip opens the session instead (`packages/ui/src/modules/home/components/feed-list.tsx:94`)
  **Fix:** see the inline comment.
- 🟡 **Warning:** A card that was never dismissed shows only its first artifact (`packages/ui/src/modules/home/views/home-view.tsx:65`) — reported also by Issue Fit and TypeScript Engineering
  **Fix:** see the inline comment.
- 🟡 **Warning:** An artifact created during the deploy window keeps no version row for its current version (`packages/db/drizzle/0038_backfill_current_artifact_versions.sql:5`)
  **Fix:** see the inline comment.
- 🟡 **Warning:** The new `session_id` predicate has no index (`packages/db/drizzle/0037_stormy_mastermind.sql:1`) — reported by TypeScript Engineering
  **Fix:** see the inline comment.
- 🟡 **Warning:** One global row limit is shared by up to 50 sessions, so one busy session starves the others (`packages/api-server/src/modules/artifact-library/infrastructure/artifact-library-repository.ts:336`) — reported by TypeScript Engineering
  **Fix:** see the inline comment.
- 🟡 **Warning:** The preview dialog opens off fetched data, so a chip click looks dead while the fetch runs and stays dead if it fails (`packages/ui/src/modules/home/views/home-view.tsx:195`) — reported by React UI Engineering
  **Fix:** see the inline comment.
- 🟡 **Warning:** Chips are limited to `unread` feed items, so a running session that just wrote an artifact shows none (`packages/ui/src/modules/home/api/queries.ts:37`) — reported by Issue Fit
  **Fix:** see the inline comment.
- 🟡 **Warning:** Documentation drift on the runtime channel's shape and on the attribution mechanism — 3 findings under **Documentation Check** below (`docs/architecture/runtime-delivery.md:16`, `docs/architecture/runtime-delivery.md:330`, `docs/architecture/artifact-library.md:17`)
  **Fix:** Update every place a page states a protocol's shape, and describe a new cross-subsystem mechanism on its owning page.
- 🟢 **Suggestion:** The `unread` feed-item kind now admits every session inside the 7-day window, so the name no longer states what the variant is; `dismissals.ts:10` and `queries.ts:37` branch on the kind alone (`packages/ui/src/modules/home/lib/feed-item.ts:65`). Consider `kind: "session"` with unreadness as a derived field. Reported by TypeScript Engineering; no user-visible defect follows today, because the count and the dot both re-check `isUnreadSession`.
- 🟢 **Suggestion:** The 7-day `FEED_WINDOW_MS` cut-off is a new user-visible retention rule that the PR body does not state (`packages/ui/src/modules/home/lib/unread.ts:15`). Reported by Issue Fit.
- 🟢 **Suggestion:** The row limit has two ceilings — the input schema caps it at 200, the repository clamps to 500 — so the higher one is unreachable (`packages/api-server-api/src/modules/artifact-library/schemas.ts:120`). Reported by TypeScript Engineering.
- 🟢 **Suggestion:** `sessionId: row.sessionId ?? ""` cannot be reached, because the `inArray` predicate excludes nulls; the fallback would group artifacts under an empty session key (`packages/api-server/src/modules/artifact-library/infrastructure/artifact-library-repository.ts:362`). Reported by TypeScript Engineering.
- 🟢 **Suggestion:** Chips render in server order with no sort by `touchedAt`, and a truncated file name carries no `title`, so the first 3 slots are arbitrary and a long name is unreadable (`packages/ui/src/modules/home/components/feed-artifact-chips.tsx:16`). Reported by React UI Engineering.
- ✅ **Looks good:** `attributeVersion` writes the session only when the row holds no session or the same one, and scopes the row to the owner. A second session cannot take over an attribution, which is what the PR body's "a chip missing rather than a chip on the wrong card" claim needs on the write path.

- 🟡 **Warning:** The page contradicts itself about the runtime channel's size. The route section now says "Three tRPC routes", but the overview still says "The runtime channel is two routes", the context bullets list only `applyState` and `hello`, and the diagram shows only those two edges (also lines 13 and 20-33). (`docs/architecture/runtime-delivery.md:16`)
  **Fix:** When a PR adds a route to a documented protocol, update every place the page states that protocol's shape — overview sentence, responsibility bullets, and diagrams — not only the section that lists the routes.
- 🟡 **Warning:** The invariant "the agent's only outbound channel is the paired gateway, which routes back to the harness-API-server's `hello`" is now stale. The agent-runtime also calls `reportArtifactTouch`. (`docs/architecture/runtime-delivery.md:330`)
  **Fix:** A new agent→api-server call belongs in the page's invariant list, not only in its route list. Invariants that list the calls across a trust boundary are updated in the PR that adds one.
- 🟡 **Warning:** The attribution mechanism has no semantics on either owning page. The artifact-library page says a version "records the session that produced it when one is known" but not who makes it known, and runtime-delivery names the route without its meaning. (`docs/architecture/artifact-library.md:17`)
  **Fix:** Describe a new cross-subsystem responsibility on the owning page — who observes what, and what the message means — and carry a one-liner plus a cross-link on the other page.

### Issue Fit
Linked issue: #3432 "UI - Show session artifacts on the Home feed cards" (closing keyword).
- 🟡 **Warning:** The never-dismissed card shows one chip — merged into `### Findings` above (`packages/ui/src/modules/home/views/home-view.tsx:65`).
- 🟡 **Warning:** Chips are gated on the `unread` kind in the fetch and the render path — merged into `### Findings` above (`packages/ui/src/modules/home/api/queries.ts:37`).
- 🟢 **Suggestion:** The 7-day feed window is an undeclared retention rule — carried in `### Findings` above (`packages/ui/src/modules/home/lib/unread.ts:15`).

- 🟡 **Warning:** One global limit across up to 50 sessions — merged into `### Findings` above (`packages/api-server/src/modules/artifact-library/infrastructure/artifact-library-repository.ts:336`).
- 🟡 **Warning:** No index on the new `session_id` predicate — merged into `### Findings` above (`packages/db/drizzle/0037_stormy_mastermind.sql:1`).
- 🟡 **Warning:** The never-dismissed card shows one chip — merged into `### Findings` above (`packages/ui/src/modules/home/views/home-view.tsx:65`).
- 🟡 **Warning:** The widened `unread` discriminant — carried in `### Findings` above as a suggestion, because no user-visible defect follows from it today (`packages/ui/src/modules/home/lib/feed-item.ts:65`).
- 🟢 **Suggestion:** Two limit ceilings, 200 and 500 — carried in `### Findings` above.
- 🟢 **Suggestion:** The unreachable `?? ""` session-id fallback — carried in `### Findings` above.

- 🔴 **Critical:** Chips are not reachable by keyboard — merged into `### Findings` above (`packages/ui/src/modules/home/components/feed-list.tsx:94`).
- 🟡 **Warning:** The never-dismissed card shows one chip — merged into `### Findings` above (`packages/ui/src/modules/home/views/home-view.tsx:65`).
- 🟡 **Warning:** The preview dialog opens off fetched data — merged into `### Findings` above (`packages/ui/src/modules/home/views/home-view.tsx:195`).
- 🟢 **Suggestion:** Unsorted chips and a truncated name with no `title` — carried in `### Findings` above.

### Verdict
REQUEST_CHANGES — the chips the PR adds cannot be operated by keyboard, and seven blocking findings across the UI, the feed read, and the migration remain open.

---

dam-code-guardian (commented):
## Code Guardian review — re-review

Prior review at `bb92f23` asked for changes. Its blocker is fixed. Three warnings survive, all of them at the same seams as before.

### Changes since last review

18 commits, and the branch was rebased onto a new `main`. The migrations were renumbered: the prior review's `0037`/`0038` are now `0038_green_lyja.sql`, `0039_backfill_current_artifact_versions.sql` and `0040_outstanding_sway.sql`, so the carried findings are re-anchored below.

### Verdict

`COMMENT` — the blocker is closed, so this no longer blocks on correctness of the feature itself. The rollout window in the backfill is the one item worth resolving before merge.

### ✅ Fixed

- 🔴 Chips are keyboard-reachable and a key press no longer opens the session. `feed-card.tsx` drops `role="button"`, `tabIndex` and the `onKeyDown` handler; the title is a real `<button>` whose `after:absolute after:inset-0` pseudo-element carries the click target. Pseudo-elements take no keyboard events, so Enter on a chip fires the chip alone.
- 🟡 The `session_id` predicate now has a matching partial index (`0040_outstanding_sway.sql:1`), and `schema.ts` declares the same one.
- 🟡 The preview dialog no longer opens off fetched data — a pending and an error header render until `useArtifact` resolves (`home-view.tsx:197`).
- 🟡 A never-dismissed card now shows every artifact its session touched, not only the first.
- 🟡 The two architecture pages now describe the marker flow, the history holding the current version, and terminal sessions staying unattributed.

### 🔁 Still present

- 🟡 **Warning:** Artifacts created during the rolling deploy window keep no current-version row. `maxSurge: 1` / `maxUnavailable: 0` (`deploy/helm/platform/templates/apiserver/app.yaml:66`) keeps old pods serving after the new pod runs the migration at bootstrap, and the old `insertArtifact` writes no version row. The one-shot backfill has already passed them, so their current version is permanently absent from `listVersions`, and `versionCount` under-reports by one. (`packages/db/drizzle/0039_backfill_current_artifact_versions.sql:5`)
  **Fix:** An invariant a rolling deploy can violate needs a self-healing writer, not a one-shot backfill — have the read or the next publish insert the missing current-version row `ON CONFLICT DO NOTHING`, so the window closes itself.
- 🟡 **Warning:** `listTouches` still applies one row limit across up to 50 sessions, ordered by recency. One busy session consumes the whole budget and the other cards show no chips. (`packages/api-server/src/modules/artifact-library/infrastructure/artifact-library-repository.ts:336`)
  **Fix:** Bound a multi-key list query per key, not globally — a lateral join or `row_number()` partitioned by `session_id` gives each requested session its own slice.
- 🟡 **Warning:** Chips resolve only for `unread` feed items, so a running session's card never shows the artifacts it is producing. (`packages/ui/src/modules/home/api/queries.ts:37`, `packages/ui/src/modules/home/views/home-view.tsx:63`)
  **Fix:** Gate artifact chips on the session, not the feed-item kind — resolve touches for every session-backed card, `in-progress` included.

### 🆕 New

- 🟢 **Suggestion:** The relay now dials `?passive=1` upstream and agent-runtime marks those channels non-viewer, so passive UI reads no longer stamp read state. The topology page still says only machine-driven channels are excluded (`docs/architecture/platform-topology.md:55`).
- 🟢 **Suggestion:** Filtering chips by the session's last dismissal time hides artifacts the session made before it (`packages/ui/src/modules/home/views/home-view.tsx:68`). The behaviour looks deliberate — `sessionDismissedAt` was added for it — so this is a note, not an objection: it is worth a line in the PR body.
- 🟢 **Suggestion:** The `+{rest} more` counter is inert text, so artifacts past the third cannot be opened by mouse or keyboard (`packages/ui/src/modules/home/components/feed-artifact-chips.tsx:36`).
- 🟢 **Suggestion:** `artifactTouchIn` matches the marker in any completed tool result, not only an artifact tool's. An unrelated tool echoing that JSON attributes a version to the wrong session; the agent-and-owner check in `recordTouch` limits the blast radius (`packages/agent-runtime/src/modules/acp/domain/artifact-touch.ts:40`).
- 🟢 **Suggestion:** `toFeedItems` defaults `now` to `Date.now()` and runs unmemoized, so a session at the 7-day edge can appear and disappear between renders (`packages/ui/src/modules/home/lib/feed-item.ts:43`).
- 🟢 **Suggestion:** `useFeedArtifacts` recomputes `wanted` every render because `items` is a fresh array from `useFeed` (`packages/ui/src/modules/home/api/queries.ts:34`).
- 🟢 **Suggestion:** The Dismiss button reveals on `group-hover` only, so keyboard focus lands on an invisible control (`packages/ui/src/modules/home/components/feed-card.tsx:74`). Pre-existing, and this round only moved the line; `icon-rail.tsx` pairs the reveal with `group-focus-within:opacity-100`.
- 🟢 **Suggestion:** `TOUCH_LIMIT_MAX` equals `TOUCH_LIMIT_DEFAULT`, so `clampTouchLimit` can never raise the limit, and the same bound is stated three times with the schema's `max(200)` (`artifact-library-repository.ts:47`).
- 🟢 **Suggestion:** The `flatMap` null-guard on `row.sessionId` is unreachable — the `inArray` predicate already excludes NULL rows (`artifact-library-repository.ts:360`).
- 🟢 **Suggestion:** `(err as Error).message` logs `undefined` for a non-`Error` throw, so the give-up line can name no cause (`packages/agent-runtime/src/modules/runtime-channel/artifact-touch-reporter.ts:36`).
- 🟢 **Suggestion:** The preview failure branch offers a close button but no retry (`packages/ui/src/modules/home/views/home-view.tsx:204`).
- 🟢 **Suggestion:** `TOUCH_SESSIONS_MAX` silently drops sessions past 50 per agent, and those cards show no chips with no signal (`packages/ui/src/modules/home/api/queries.ts:46`).

### doc-drift

- 🟢 **Suggestion:** The relay now dials `?passive=1` upstream and agent-runtime marks those channels non-viewer, so passive UI reads no longer stamp session read state. The page still says only machine-driven channels are excluded (`docs/architecture/platform-topology.md:55`).

### issue-fit

Linked issue: #3432 — `Closes #3432`, three sub-issues all checked.

- 🟡 **Warning:** Chips render only for `unread` feed items, so a running session's card never shows its artifacts (`packages/ui/src/modules/home/views/home-view.tsx:63`).
  **Fix:** Gate artifact chips on the session, not the feed-item kind.
- 🟢 **Suggestion:** Dismissal time filters the chips, hiding artifacts a session made before its last dismissal (`packages/ui/src/modules/home/views/home-view.tsx:69`).
- 🟢 **Suggestion:** The overflow chip renders as plain text `+N more`, so artifacts past the third cannot be opened (`packages/ui/src/modules/home/components/feed-artifact-chips.tsx:36`).
- 🟢 **Suggestion:** `TOUCH_SESSIONS_MAX` silently drops sessions past 50 per agent (`packages/ui/src/modules/home/api/queries.ts:46`).

The three extra scopes (card stays until dismissed, 7-day age-out, passive read no longer marks seen) and the version-history change are declared in the PR body — not creep.

### typescript-engineering

- 🟡 **Warning:** `listTouches` applies one row limit across up to 50 sessions, ordered by recency, so a busy session consumes the whole budget (`packages/api-server/src/modules/artifact-library/infrastructure/artifact-library-repository.ts:336`).
  **Fix:** Bound a multi-key list query per key — `row_number()` partitioned by `session_id`, or a lateral join.
- 🟢 **Suggestion:** `TOUCH_LIMIT_MAX` equals `TOUCH_LIMIT_DEFAULT`, and the schema's `max(200)` states the same bound a third time (`artifact-library-repository.ts:47`).
- 🟢 **Suggestion:** The `flatMap` null-guard on `row.sessionId` is unreachable (`artifact-library-repository.ts:360`).
- 🟢 **Suggestion:** `(err as Error).message` logs `undefined` for a non-`Error` throw (`packages/agent-runtime/src/modules/runtime-channel/artifact-touch-reporter.ts:36`).
- 🟢 **Suggestion:** `toFeedItems` defaults `now` to `Date.now()` unmemoized, so a 7-day-edge session can flicker between renders (`packages/ui/src/modules/home/lib/feed-item.ts:43`).
- 🟢 **Suggestion:** `artifactTouchIn` matches the marker in any completed tool result, not only artifact-tool results (`packages/agent-runtime/src/modules/acp/domain/artifact-touch.ts:40`).
- 🟢 **Suggestion:** `useFeedArtifacts` recomputes `wanted` on every render (`packages/ui/src/modules/home/api/queries.ts:34`).

### react-ui-engineering

- 🟢 **Suggestion:** The Dismiss button stays `opacity-0` and reveals on `group-hover` only, so keyboard focus lands on an invisible control (`packages/ui/src/modules/home/components/feed-card.tsx:74`). Pre-existing; `icon-rail.tsx` already pairs the reveal with `group-focus-within:opacity-100`.
- 🟢 **Suggestion:** The `+{rest} more` counter is inert text (`packages/ui/src/modules/home/components/feed-artifact-chips.tsx:36`).
- 🟢 **Suggestion:** A chip's accessible name is only the file name, so nothing tells a screen reader it opens a preview (`packages/ui/src/modules/home/components/feed-artifact-chips.tsx:24`).
- 🟢 **Suggestion:** The preview failure branch offers no retry (`packages/ui/src/modules/home/views/home-view.tsx:204`).

Verified at their cause: the prior keyboard 🔴 and the prior preview-dialog 🟡 are both fixed, and dropping `stopPropagation` from Dismiss is safe — the card `<div>` has no `onClick` and the button sits at `relative z-10` above the title overlay.

---

<sub>Reviewed by Code Guardian 🛡️ · [definition](https://github.com/dam-agents/code-guardian)</sub>

kapetr:
@dam-code-guardian The two carried warnings are decisions, not gaps:
**Backfill rollout window** — accepted. We run one replica, so the window is empty in practice. If a missing current-version row ever shows up, the self-healing insert you describe is the fix.

**Running-session chips** — by design. A running turn may never touch the artifact, or may still rewrite it — so the card makes no claim until the session finishes. Then the chips show exactly what was touched.

**Row limit across sessions** — accepted for now. With the limit at 200 and three chips per card, starving another card takes 200 touches in 7 days on one agent. Not worth a lateral join yet.

docs/architecture/runtime-delivery.md
```diff
@@ -333,5 +337,7 @@ Capabilities also gate whole flows, not only payload items: `hello` carries a nu
...
 - **Events settle per id, contributions per version.** The worker stamps `dispatched_at` for the events the agent reports it ran, whatever the contribution outcome.
+- **The api-server is the only caller of `applyState` from the cluster.** The harness port admits ingress only from api-server pods; the agent's only outbound channel is the paired gateway, which routes back to the harness API server's two callbacks: `hello`, and the artifact-touch report — the agent-runtime saying which session produced an artifact version, having seen the platform tool's marked result in that session's ACP stream. The receiving side verifies the artifact belongs to the calling agent and never overwrites another session's attribution; the semantics live with [the artifact library](artifact-library.md).
+- **Capabilities are honored end-to-end.** A Contribution or Event kind not in the agent's advertised set is dropped at send time, never silently delivered. A grant that requires unsupported kinds succeeds with a UI warning; the unsupported parts simply don't appear in the agent's payload.
```
dam-code-guardian: 🔴 **Critical:** This merge appended a second copy of two invariants, and the superseded copies survive two lines below. Line 342 still names `hello` as the agent's only outbound callback, which contradicts both the bullet added here and the new `reportArtifactTouch` route this PR adds. Line 343's older "Capabilities are honored end-to-end" copy carries a flow-gating sentence that the copy added here drops, so the page states that rule at two levels of completeness.

**Fix:** Keep one invariant per topic. When a PR changes an invariant, edit the existing bullet in place rather than appending a second version; delete lines 342-343 and fold their surviving detail into the two bullets above.

packages/ui/src/modules/home/views/home-view.tsx
```diff
@@ -163,10 +188,28 @@ export function HomeView() {
...
           )}
         </div>
+        {previewArtifactId &&
+          (previewArtifact ? (
+            <ArtifactPreviewDialog
+              artifact={previewArtifact}
+              onClose={() => setPreviewArtifactId(null)}
+            />
+          ) : (
+            <Modal widthClass="w-[860px]">
```
dam-code-guardian: 🟡 **Warning:** This fallback dialog closes only through the header Close button. The shared `Modal` binds no Escape handler and no backdrop click, so a keyboard user has one single exit. Every other dialog in the module adds its own listener (`move-artifact-dialog.tsx:35`, `rename-artifact-dialog.tsx:27`, `fullscreen-preview-dialog.tsx:16`), and this one does not.

**Fix:** Every modal must close on Escape. Put the Escape listener in the shared `Modal` component, so no dialog can be built without it, and drop the per-dialog copies.

packages/agent-runtime/src/modules/acp/infrastructure/artifact-touch.ts
```diff
@@ -0,0 +1,54 @@
+import { artifactTouchPayloadSchema } from "api-server-api";
```
dam-code-guardian, line 1: 🟡 **Warning:** The `acp` domain layer imports a Zod schema from the `api-server-api` contract package. The domain layer must hold pure TypeScript with no external dependency, so a contract change now reaches the innermost layer.

**Fix:** Keep every `domain/` file free of workspace and third-party imports. Validate the payload in the service or infrastructure caller and pass the domain layer an already-parsed plain value.

packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts
```diff
@@ -83,6 +87,7 @@ export interface AcpRuntimeDeps {
   backgroundWork?: BackgroundWorkRegistry;
   backgroundWorkRecheckMs?: number;
   isTerminalSessionActive?: (sessionId: string) => boolean;
+  onArtifactTouch?: (touch: ArtifactTouch) => void;
```
dam-code-guardian: 🟡 **Warning:** `onArtifactTouch` is an optional dependency, and line 633 branches on its presence. A wiring that was forgotten then looks exactly like a deliberate opt-out, and the feature fails silently.

**Fix:** Make every service dependency required and call it unconditionally. Move the on/off decision into the dependency itself — a no-op reporter when the feature is off — configured where it is built at the composition root.

packages/agent-runtime/src/modules/acp/compose.ts
```diff
@@ -46,6 +47,7 @@ export interface ComposeAcpOptions {
   };
   isTerminalSessionActive?: (sessionId: string) => boolean;
   backgroundWorkHolds?: boolean;
+  onArtifactTouch?: (touch: ArtifactTouch) => void;
```
dam-code-guardian: 🟡 **Warning:** `ComposeAcpOptions` carries the same optional `onArtifactTouch`, so the consumer of the composition helper must know the capability is switchable.

**Fix:** Have the composition root always build a reporter — a no-op one when the feature is off — and pass it as a required dependency instead of an optional callback.

packages/agent-runtime/src/modules/runtime-channel/artifact-touch-reporter.ts
```diff
@@ -0,0 +1,50 @@
...
+    version: number;
+  }) => void;
+}
+
+export function createArtifactTouchReporter(opts: {
+  apiServerUrl: string;
+  agentId: string;
+  log?: (msg: string) => void;
+}): ArtifactTouchReporter {
+  const client = createHarnessClient({
```
dam-code-guardian: 🟡 **Warning:** `createArtifactTouchReporter` builds its own harness tRPC client although the composition root already holds one. This bypasses dependency injection and leaves two clients for the same upstream.

**Fix:** Inject external-system clients as a port through the factory `opts`. Construct concrete clients exactly once, at the composition root.

packages/api-server-api/src/modules/runtime/harness-router.ts
```diff
@@ -7,6 +8,11 @@ const v1Router = harnessT.router({
     .mutation(({ ctx, input }) =>
       ctx.runtimeDelivery.hello(ctx.agentId, input),
     ),
+  reportArtifactTouch: harnessT.procedure
```
dam-code-guardian: 🟡 **Warning:** The `runtime` contract module hosts the `reportArtifactTouch` procedure and imports the artifact-library schema for it, so the artifact-library contract is now split across two modules.

**Fix:** Keep each procedure in the module that owns its bounded context. Put artifact-touch procedures in the `artifact-library` module and compose that router into the harness root router.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `f2c5b51`

## PR #3456: Session artifacts on the Home feed cards
**Author:** kapetr | **Branch:** feat/3432-session-artifacts → main | **Changes:** +6951 −76 (41 files)

### Summary
The PR shows, on each Home feed session card, the artifacts that session touched, and it moves the version history to hold the current version. Complete re-review at the label's request: the PR's own code is unchanged since `c148b37`, but the merge of `main` left two contradictory invariant bullets in `runtime-delivery.md`.
_(Suppressed 3 finding(s) per PR context: rollout-window backfill, global row limit in `listTouches`, chips only on `unread` — the author accepted all three in [a comment](https://github.com/dam-agents/dam/pull/3456#issuecomment-5507270121) on 2026-09-02.)_

### Changes since last review
Previous HEAD: `c148b37` (2026-09-02T08:28:31Z) — verdict COMMENT

3 commits, all from `main`: a CRD doc-comment fix, the knowledge-base MCP share, and the merge itself. No change to the PR's own code, so no finding from the previous review is fixed.

- 🔁 **Still present:** rollout window leaves current-version rows absent (`packages/db/drizzle/0040_backfill_current_artifact_versions.sql:5`) — accepted by the author, suppressed
- 🔁 **Still present:** one global row limit across up to 50 sessions (`packages/api-server/src/modules/artifact-library/infrastructure/artifact-library-repository.ts:338`) — accepted by the author, suppressed
- 🔁 **Still present:** chips resolve only for `unread` items (`packages/ui/src/modules/home/api/queries.ts:37`) — accepted by the author, suppressed
- 🔁 **Still present:** Dismiss reveals on hover only (`packages/ui/src/modules/home/components/feed-card.tsx:74`)
- 🔁 **Still present:** `TOUCH_LIMIT_MAX` equals `TOUCH_LIMIT_DEFAULT` and the cap is stated three times (`packages/api-server/src/modules/artifact-library/infrastructure/artifact-library-repository.ts:48`)
- 🔁 **Still present:** the `+N more` counter is inert text (`packages/ui/src/modules/home/components/feed-artifact-chips.tsx:36`)
- 🆕 **New:** the merge duplicated two invariant bullets, and the superseded copies survive (`docs/architecture/runtime-delivery.md:341`)
- 🆕 **New:** six layering and accessibility findings first reported at this HEAD — listed in full below

### Findings
- 🔴 **Critical:** duplicated invariants in `docs/architecture/runtime-delivery.md:341` (inline)
- 🟡 **Warning:** the new fallback dialog has no Escape key and no backdrop close (`packages/ui/src/modules/home/views/home-view.tsx:204`) (inline)
- 🟡 **Warning:** the `acp` domain layer imports a Zod schema from the contract package (`packages/agent-runtime/src/modules/acp/domain/artifact-touch.ts:1`) (inline)
- 🟡 **Warning:** `onArtifactTouch` is an optional dependency in two places (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:90`, `packages/agent-runtime/src/modules/acp/compose.ts:50`) (inline)
- 🟡 **Warning:** the reporter builds its own harness client (`packages/agent-runtime/src/modules/runtime-channel/artifact-touch-reporter.ts:19`) (inline)
- 🟡 **Warning:** `reportArtifactTouch` sits in the `runtime` contract module (`packages/api-server-api/src/modules/runtime/harness-router.ts:11`) (inline)
- 🟡 **Warning:** the "UI surfaces" section of `docs/architecture/artifact-library.md:170` does not list the new Home feed surface. The Overview added by this PR describes the chips and the preview dialog, but the section a reader uses to find every surface still names only the Artifacts destination, the sandbox home view, and the chat view.
  **Fix:** Add each new library surface to the section that owns the surface list. A mention in the Overview never replaces that enumeration.
- 🟢 **Suggestion:** `item.kind === "unread" && isUnreadSession(...)` now appears at three sites (`packages/ui/src/modules/home/components/feed-list.tsx:88`, `packages/ui/src/modules/home/lib/feed-filter.ts:53`, `:76`). One later edit makes the filter, the counter and the badge disagree. Extract one predicate over `FeedItem`.
- 🟢 **Suggestion:** the unread dot and the working dots carry no text equivalent, so the card states unread and working by pixels alone (`packages/ui/src/modules/home/components/feed-card.tsx:66`). Pre-existing — this PR only re-renders it; worth a separate issue with the other feed indicators.
- 🟢 **Suggestion:** the `+{rest} more` counter is an inert `<span>`, so artifacts past the third cannot be opened from the card (`packages/ui/src/modules/home/components/feed-artifact-chips.tsx:36`).
- 🟢 **Suggestion:** `TOUCH_LIMIT_MAX` equals `TOUCH_LIMIT_DEFAULT`, so `clampTouchLimit` can never raise the limit, and the schema's `max(200)` states the same bound a third time (`packages/api-server/src/modules/artifact-library/infrastructure/artifact-library-repository.ts:48`). The typescript-engineering skill rates this 🟡; it is held at 🟢 here — unchanged since `c148b37`, where the same judgment was accepted.
- 🟢 **Suggestion:** the Dismiss button stays focusable while `opacity-0` and reveals on `group-hover` only (`packages/ui/src/modules/home/components/feed-card.tsx:74`). The react-ui-engineering skill rates this 🟡; it is held at 🟢 here — the behaviour is unchanged since `c148b37` and this round only moved the line. `icon-rail.tsx:135` already pairs the reveal with `group-focus-visible:opacity-100`.

### doc-drift

- 🔴 **Critical:** A duplicated invariant bullet survives the merge and still names `hello` as the agent's only callback. The page contradicts the same section two lines above it, and contradicts the new `reportArtifactTouch` route this PR adds (`docs/architecture/runtime-delivery.md:342`)
  **Fix:** Delete the superseded copy and keep one invariant per topic. When a PR changes an invariant, edit the existing bullet in place instead of appending a second version of it.
- 🟡 **Warning:** The "Capabilities are honored end-to-end" invariant appears twice. The added copy drops the flow-gating sentence that the original copy carries, so the page states the rule at two different levels of completeness (`docs/architecture/runtime-delivery.md:341`)
  **Fix:** Remove the added duplicate and retain the complete bullet. Check the invariants list for repeated bullet titles before you commit an architecture-page edit.
- 🟡 **Warning:** The "UI surfaces" section does not list the new Home feed surface (`docs/architecture/artifact-library.md:170`)
  **Fix:** Add each new library surface to the section that owns the surface list.

### issue-fit

Linked issue: #3432 — `Closes #3432`.

- 🟡 **Warning:** Issue #3432 asks that the card of a session shows the artifacts of that session, but only `unread` cards get chips, so a running session never shows its artifacts (`packages/ui/src/modules/home/views/home-view.tsx:63`, `packages/ui/src/modules/home/api/queries.ts:37`) — **accepted by the author** on 2026-09-02: a running turn may still rewrite the artifact, so the card makes no claim until the session finishes. Suppressed, not re-raised.
- 🟢 **Suggestion:** The card shows at most 3 chips and renders the remainder as inert `+N more` text, so a session with more artifacts gives no way to open them (`packages/ui/src/modules/home/components/feed-artifact-chips.tsx:5`).

### typescript-engineering

- 🟡 **Warning:** The `acp` domain layer imports a Zod schema from the `api-server-api` contract package. The domain layer must hold pure TypeScript with zero external dependencies (`packages/agent-runtime/src/modules/acp/domain/artifact-touch.ts:1`)
  **Fix:** Keep every `domain/` file free of workspace and third-party imports. Validate the payload in the caller and pass the domain layer an already-parsed value.
- 🟡 **Warning:** `AcpRuntimeDeps.onArtifactTouch` is optional and line 633 branches on its presence, so an absent wiring looks the same as a deliberate opt-out (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:90`)
  **Fix:** Make every service dependency required and call it unconditionally. Move the on/off decision into the dependency itself, built at the composition root.
- 🟡 **Warning:** `ComposeAcpOptions` also carries `onArtifactTouch` as optional, so the consumer knows the capability is switchable (`packages/agent-runtime/src/modules/acp/compose.ts:50`)
  **Fix:** Have the composition root always build a reporter — a no-op one when the feature is off — and pass it as a required dependency.
- 🟡 **Warning:** `createArtifactTouchReporter` builds its own harness tRPC client although the composition root already holds one (`packages/agent-runtime/src/modules/runtime-channel/artifact-touch-reporter.ts:19`)
  **Fix:** Inject external-system clients as a port through the factory `opts`. Construct concrete clients once, at the composition root.
- 🟡 **Warning:** The `runtime` contract module hosts `reportArtifactTouch` and imports the artifact-library schema for it, so the artifact-library contract is split across two modules (`packages/api-server-api/src/modules/runtime/harness-router.ts:11`)
  **Fix:** Keep each procedure in the module that owns its bounded context, and compose that router into the harness root router.
- 🟡 **Warning:** `clampTouchLimit` re-implements the 200-row cap the contract Zod schema already enforces, so the two constants can drift (`packages/api-server/src/modules/artifact-library/infrastructure/artifact-library-repository.ts:50`)
  **Fix:** Validate and clamp external input once, in the contract schema; let the repository hold only its own default. _(Held at 🟢 in `### Findings` — unchanged since `c148b37`.)_
- 🟢 **Suggestion:** The contract names the operation `ArtifactTouchService.record`, but the server implements `recordTouch`, which forces an inline adapter object in the harness wiring (`packages/api-server-api/src/modules/artifact-library/types.ts:505`, `packages/api-server/src/apps/harness-api-server/harness-router.ts:72`)
  **Fix:** Use the same method name in the contract and the implementation.

### react-ui-engineering

- 🟡 **Warning:** The Dismiss button stays focusable while `opacity-0`, and only `group-hover` reveals it, so a keyboard user focuses an invisible control (`packages/ui/src/modules/home/components/feed-card.tsx:74`)
  **Fix:** Every control that hides itself with opacity must also reveal itself on focus. Add a `focus-visible:opacity-100` or `group-focus-within:opacity-100` variant beside each `group-hover:opacity-100`. _(Held at 🟢 in `### Findings` — unchanged since `c148b37`.)_
- 🟡 **Warning:** The unread dot and the working dots carry no accessible text, so the card signals unread and working state by pixels alone (`packages/ui/src/modules/home/components/feed-card.tsx:66`)
  **Fix:** Give every state indicator a text equivalent — visually hidden text or an `aria-label` on the card title button — and mark the decorative dot `aria-hidden`. _(Held at 🟢 in `### Findings` — the line is not in this diff.)_
- 🟡 **Warning:** The new fallback Modal closes only through its header Close button. It ignores the Escape key and backdrop clicks, so keyboard users have one single exit (`packages/ui/src/modules/home/views/home-view.tsx:204`)
  **Fix:** Every modal must close on Escape. Put the Escape listener in the shared `Modal` component, so no dialog can be built without it.
- 🟢 **Suggestion:** `item.kind === "unread" && isUnreadSession(session)` is now repeated in three places (`packages/ui/src/modules/home/components/feed-list.tsx:88`). Extract one predicate over `FeedItem` and call it from all three.

### Verdict
REQUEST_CHANGES — the merge left `runtime-delivery.md` stating two contradictory versions of the same invariant, one of them denying the callback route this PR adds; six layering and accessibility warnings are open beside it.

---

jezekra1 (approved):

packages/ui/src/components/modal.tsx
```diff
@@ -8,14 +8,27 @@ import { cn } from "@/lib/utils";
...
+export function Modal({
+  widthClass = "w-[560px]",
+  onClose,
+  children,
+}: ModalProps) {
   const labelId = useId();
   const panelRef = useRef<HTMLDivElement>(null);
   useFocusTrap(panelRef);
   useBodyScrollLock();
+  useEffect(() => {
```
dam-code-guardian, line 24: 🟢 **Suggestion:** this Escape listener is global and unconditional, so a nested dialog and its parent both close on one key press. `ArtifactPreviewDialog` passes `onClose` to `Modal` and `FullscreenPreviewDialog` keeps its own `document` listener, so one Escape closes the fullscreen view and the preview dialog together.

**Fix:** A dialog Escape handler must close only the topmost dialog. Register the handler on the dialog panel element, or keep a shared modal stack and let only its last entry act.

packages/ui/src/modules/home/components/feed-artifact-chips.tsx
```diff
@@ -0,0 +1,48 @@
...
+          title={artifact.fileName}
+          onClick={() => onOpen(artifact.artifactId)}
+          className="inline-flex items-center gap-1.5 rounded-md border border-border/50 bg-muted/40 px-2.5 py-1.5 text-sm text-muted-foreground transition-all hover:border-border hover:bg-muted/70 hover:text-foreground"
+        >
+          <Document size={16} className="shrink-0" />
+          <span className="max-w-[160px] truncate">{artifact.fileName}</span>
+        </button>
+      ))}
+      {rest > 0 && (
+        <button
```
dam-code-guardian, line 38: 🟢 **Suggestion:** this button unmounts itself when clicked, because `rest` becomes 0 after the expansion. Keyboard focus then falls back to `document.body` and the user loses their place in the card list.

**Fix:** A control that removes itself on activation must move focus to a stable target. Focus the first newly revealed chip, or keep the control mounted as a collapse toggle.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `abe8d8de`

## PR #3456: Session artifacts on the Home feed cards
**Author:** kapetr | **Branch:** feat/3432-session-artifacts → main | **Changes:** +7000 −83 (45 files)

### Summary
The PR shows, on each Home feed session card, the artifacts that session touched, and it moves the version history to hold the current version. Complete re-review at the label's request: the fix commit `abe8d8de` answers all seven open findings at their cause.
_(Suppressed 3 finding(s) per PR context: rollout-window backfill, global row limit in `listTouches`, chips only on `unread` — the author accepted all three on 2026-09-02.)_

### Changes since last review
Previous HEAD: `f2c5b51` (2026-09-02T12:42:09Z) — verdict REQUEST_CHANGES

1 commit, `abe8d8de` (23 files, +87 −45).

- ✅ **Fixed:** the duplicated invariant bullets are deleted, one bullet per topic (`docs/architecture/runtime-delivery.md`)
- ✅ **Fixed:** the "UI surfaces" list names the Home feed artifact chips (`docs/architecture/artifact-library.md:172`)
- ✅ **Fixed:** `artifact-touch.ts` moved out of `domain/` into `infrastructure/` (`packages/agent-runtime/src/modules/acp/infrastructure/artifact-touch.ts`)
- ✅ **Fixed:** `onArtifactTouch` is required in both `AcpRuntimeDeps` and `ComposeAcpOptions`, called unconditionally (`acp-runtime.ts:90`, `compose.ts:50`)
- ✅ **Fixed:** the reporter takes an injected `HarnessClient` (`artifact-touch-reporter.ts:19`)
- ✅ **Fixed:** the route moved to the artifact-library contract module as `artifactLibrary.v1.reportTouch` (`packages/api-server-api/src/modules/artifact-library/harness-router.ts`)
- ✅ **Fixed:** `Modal` closes on Escape, Dismiss reveals on `focus-visible`, `+N more` is a real button, the repeated unread predicate is one `isUnreadItem`
- 🆕 **New:** seven suggestions, all on the surface the fix itself adds — listed below

### Findings
- 🟢 **Suggestion:** the new `Modal` Escape listener is global and unconditional, so one Escape closes a nested dialog and its parent together. In `ArtifactPreviewDialog` the fullscreen preview and the `Modal` each hold a listener (`packages/ui/src/components/modal.tsx:24`)
- 🟢 **Suggestion:** the `+N more` button unmounts itself when clicked, so focus falls back to `document.body` and the user loses their place in the card list (`packages/ui/src/modules/home/components/feed-artifact-chips.tsx:38`)
- 🟢 **Suggestion:** `runtime-delivery.md` still names the route `runtime.reportArtifactTouch` under the `runtime.v1.*` prefix, which this commit renamed to `artifactLibrary.v1.reportTouch` (`docs/architecture/runtime-delivery.md:13`, `:16`, `:33`, `:63`). All docs items of this round are folded here.
- 🟢 **Suggestion:** `isUnreadItem` takes a widened structural shape instead of the `FeedItem` union, so a rename of the `"unread"` kind stays a silent `false` rather than a compile error (`packages/ui/src/modules/home/lib/unread.ts:27`)
- 🟢 **Suggestion:** `createHarnessClient` and `HarnessClient` are new exports of the `runtime-channel` index, but its only consumer still imports the deep path (`packages/agent-runtime/src/modules/runtime-channel/index.ts:6`, `packages/agent-runtime/src/server.ts:35`)
- 🟢 **Suggestion:** two `HarnessClient` instances now exist for the same api-server URL and agent id — `server.ts` builds one and `composeRuntimeChannel` builds a second internally (`packages/agent-runtime/src/server.ts:94`)
- 🟢 **Suggestion:** a card lists only the artifacts touched after its last dismissal, so an earlier artifact never gets a chip when the card returns (`packages/ui/src/modules/home/views/home-view.tsx:68`)

### doc-drift
- 🟢 **Suggestion:** the page names the callback `runtime.reportArtifactTouch` and assigns it to the api-server Runtime Delivery context; `abe8d8de` moved the route to the `artifactLibrary` namespace as `reportTouch` (`docs/architecture/runtime-delivery.md:13`, `:33`).
- 🟢 **Suggestion:** the page states the channel is three tRPC routes "prefixed by protocol-major version (`runtime.v1.*`)". After `abe8d8de` only `applyState` and `hello` sit under `runtime.v1.*` (`docs/architecture/runtime-delivery.md:16`, `:63`).

### issue-fit
Linked issue: #3432 — `Closes #3432`. No missing scope and no undeclared scope creep.
- 🟢 **Suggestion:** the card lists only artifacts touched after its last dismissal, but the issue asks for "the artifacts of that session" (`packages/ui/src/modules/home/views/home-view.tsx:68`).

### typescript-engineering
- 🟢 **Suggestion:** `isUnreadItem` takes a widened structural shape rather than the `FeedItem` union, dropping discriminated-union checking at both call sites (`packages/ui/src/modules/home/lib/unread.ts:27`).
- 🟢 **Suggestion:** the new `runtime-channel` index exports `createHarnessClient` and `HarnessClient`, but no file imports them through the index (`packages/agent-runtime/src/modules/runtime-channel/index.ts:6`).
- 🟢 **Suggestion:** the `acp` service layer imports `artifactTouchIn` from `infrastructure/` directly. The same pattern already exists in `services/trigger-session-driver.ts`, so the move added one instance, not a new pattern (`acp-runtime.ts:9`).
- 🟢 **Suggestion:** after the `record` → `recordTouch` rename the harness adapter is an identity passthrough (`packages/api-server/src/apps/harness-api-server/harness-router.ts:73`).
- 🟢 **Suggestion:** two `HarnessClient` instances for the same api-server URL and agent id (`packages/agent-runtime/src/server.ts:94`).

### react-ui-engineering
- 🟢 **Suggestion:** the new `Modal` Escape listener is global, so a nested dialog and its parent both close on one Escape. Reproduced with React 19 and jsdom (`packages/ui/src/components/modal.tsx:24`).
- 🟢 **Suggestion:** the `+N more` button unmounts itself on click, so `activeElement` moves to `BODY`. Reproduced with React 19 and jsdom (`feed-artifact-chips.tsx:38`).
- 🟢 **Suggestion:** `FeedArtifactChips` expands but never collapses, so the card grows until remount (`feed-artifact-chips.tsx:15`).
- 🟢 **Suggestion:** `Modal` now owns Escape, but 26 other call sites pass no `onClose` and two dialogs keep their own effect, so Escape support is inconsistent (`packages/ui/src/components/modal.tsx:11`).
- 🔁 1 finding from the previous review still present (see review at `f2c5b51`) — the unread and working dots carry no accessible text; outside this delta.

### Verdict
APPROVE — `abe8d8de` closes all seven open findings at their cause, and every remaining item is the cost of those fixes, not a blocker.

---

dam-code-guardian:
@kapetr All three are recorded as decisions. The backfill window is the useful one: I stored that the api-server runs one replica, so I must read the replica count before the deployment strategy, and the window is a green note until a second writer can exist. The mechanism stands, and the self-healing insert is the fix if a missing row ever shows up. The running-session chips are stored as deliberate — the card makes no claim until the session finishes, and an absent claim is not incomplete state. The row limit is accepted with its arithmetic: 200 touches in 7 days on one agent, so no lateral join yet. None of these will come back in a later review.

PR #3457, opened 2026-08-25
feat: operator-configurable compute request link

kapetr:
The "Need more?" link on the home compute widget pointed at a Slack channel that does not exist (`app_redirect?channel=DAMBUDGET`).

- The link now comes from the server: Helm `links.computeRequest` → `LINKS_COMPUTE_REQUEST` env → `links.all` tRPC query, so it can be changed per install without a UI rebuild.
- When unset, the server serves `null` and the UI falls back to its bundled constant, now pointing at #dam-community.
- Label renamed from "Need more?" to "Request more" to make the action clearer.

Requested in Slack by Jamie Jabbour.

### DIFF
--- a/packages/api-server-api/src/modules/links/types.ts
+++ b/packages/api-server-api/src/modules/links/types.ts
@@ -1,7 +1,7 @@
 import { z } from "zod";
 
 export const linksSchema = z.object({
-  computeRequest: z.url().nullable(),
+  computeRequest: z.string().nullable(),
 });
 
 export type Links = z.infer<typeof linksSchema>;
--- a/packages/api-server-api/src/modules/links/types.ts
+++ b/packages/api-server-api/src/modules/links/types.ts
@@ -1,7 +1,7 @@
 import { z } from "zod";
 
 export const linksSchema = z.object({
-  computeRequest: z.url(),
+  computeRequest: z.url().nullable(),
 });
 
 export type Links = z.infer<typeof linksSchema>;
--- a/packages/api-server-api/src/modules/links/types.ts
+++ b/packages/api-server-api/src/modules/links/types.ts
@@ -1,0 +1,7 @@
+import { z } from "zod";
+
+export const linksSchema = z.object({
+  computeRequest: z.url(),
+});
+
+export type Links = z.infer<typeof linksSchema>;
--- a/deploy/helm/platform/templates/apiserver/app.yaml
+++ b/deploy/helm/platform/templates/apiserver/app.yaml
@@ -404,6 +404,10 @@ spec:
               value: {{ .Values.terms.version | quote }}
             - name: TERMS_TEXT
               value: {{ .Values.terms.text | quote }}
+            {{- if .Values.links.computeRequest }}
+            - name: LINKS_COMPUTE_REQUEST
+              value: {{ .Values.links.computeRequest | quote }}
+            {{- end }}
             {{- if .Values.clickstack.enabled }}
             # The api-server's own operational telemetry (request traces,
             # per-procedure metrics, logs) exports OTLP straight to the
--- a/deploy/helm/platform/values.yaml
+++ b/deploy/helm/platform/values.yaml
@@ -51,6 +51,14 @@ terms:
   # judgment — whitespace edits don't bump this.
   version: ""
 
+# -- Outbound links surfaced in the UI, changeable per install without a
+# rebuild (the api-server serves them over `links.all`).
+links:
+  # Where "Request more" next to the compute meter points — a chat channel,
+  # form, or ticket queue. Complete URL. Empty string uses the bundled
+  # default.
+  computeRequest: ""
+
 # -- Base domain for the app. UI and API share this host; the ingress
 # splits `/api/*` to the api-server and everything else to the UI. Keycloak
 # lives at `keycloak.{domain}`.
--- a/deploy/helm/platform/values.yaml
+++ b/deploy/helm/platform/values.yaml
@@ -55,7 +55,7 @@ terms:
 # rebuild (the api-server serves them over `links.all`).
 links:
   # Where "Request more" next to the compute meter points — a chat channel,
-  # form, or ticket queue. Complete URL. Empty string uses the bundled
+  # form, or ticket queue. Complete URL. Empty string uses the UI's bundled
   # default.
   computeRequest: ""
 
--- a/packages/api-server-api/src/context.ts
+++ b/packages/api-server-api/src/context.ts
@@ -12,6 +12,7 @@ import type { LiveEventsService } from "./modules/events/types.js";
 import type { ExperimentsService } from "./modules/experiments/types.js";
 import type { InvocationsQueryService } from "./modules/invocations/types.js";
 import type { KnowledgeBasesService } from "./modules/knowledge-bases/types.js";
+import type { Links } from "./modules/links/types.js";
 import type { FilesService } from "./modules/files/router.js";
 import type { HarnessConfigService } from "./modules/harness-config/types.js";
 import type { SchedulesService } from "./modules/schedules/types.js";
@@ -47,6 +48,7 @@ export interface ApiContext {
   features: FeaturesService;
   files: FilesService;
   harnessConfig: HarnessConfigService;
+  links: Links;
   liveEvents: LiveEventsService;
   metrics: MetricsService;
   terms: TermsService;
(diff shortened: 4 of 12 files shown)


packages/api-server-api/src/modules/links/types.ts
```diff
@@ -0,0 +1,7 @@
+import { z } from "zod";
+
+export const linksSchema = z.object({
+  computeRequest: z.url().nullable(),
```
dam-code-guardian: 🟡 **Warning:** `z.url()` here is wrong in both directions. A typo'd `LINKS_COMPUTE_REQUEST` (`slack.com/x`, no scheme) fails this `.parse()` inside `loadConfig()`, which `bootstrap()` calls unguarded — so a cosmetic link crash-loops the api-server. In the other direction `z.url()` accepts `javascript:` and `mailto:`, and `compute-widget.tsx:66` renders the served value as an `href` without the `isExternalHttpUrl` guard that `release-notes-link.tsx:13` applies to its server-driven link. Both confirmed against zod 4.4.3.

**Fix:** An optional cosmetic URL from operator config degrades, it does not abort — `z.httpUrl()` rejects non-http schemes and `.catch(null)` turns a bad value into the bundled fallback instead of a failed boot. The tradeoff is that a typo then disables the link silently rather than loudly.

```suggestion
  computeRequest: z.httpUrl().nullable().catch(null),
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `c522777`

## PR #3457: feat: operator-configurable compute request link
**Author:** kapetr | **Branch:** fix/home-request-more-link → main | **Changes:** +53 −4 (12 files)

### Summary
Moves the home compute widget's "Request more" link into server config (Helm `links.computeRequest` → `LINKS_COMPUTE_REQUEST` → new `links.all` tRPC query), with the bundled UI constant as the fallback and a corrected default channel URL. Wiring is complete and matches the PR body end to end.
_(Suppressed 1 finding per PR context: the UI treats a server `null` as "use the bundled default" — the body declares that behavior.)_

### Findings
- 🟡 **Warning:** `z.url()` at the config boundary is wrong in both directions — see the inline comment on `packages/api-server-api/src/modules/links/types.ts:4`.

✅ No findings.

### Issue Fit
- 🟢 **Suggestion:** The body says the fallback points at #dam-community, but the constant is an opaque archive URL (`packages/ui/src/constants.ts:15`), so the claim is not checkable from the diff.

Every other body promise has matching code: Helm value and env wiring, the `links.all` query, the `null` default when unset, the UI fallback, and the label rename. No undeclared scope.

The skill reported three module-shape findings as 🟡. Verified as accurate observations, recorded here as 🟢 — none creates a defect, and a service wrapper over a static config value adds no behavior:

- 🟢 `linksRouter.all` returns `ctx.links` directly; it is the only router in the package that returns raw context state instead of a service method (`packages/api-server-api/src/modules/links/router.ts:5`).
- 🟢 No `packages/api-server/src/modules/links/` counterpart exists — `links: config.links` passes config through as the context value (`packages/api-server/src/apps/api-server/trpc/context.ts:285`).
- 🟢 `links` groups by data shape rather than domain; `computeRequest` belongs to the compute/budget area that consumes it (`packages/api-server-api/src/modules/links/types.ts:3`).
- 🟢 `COMPUTE_REQUEST_URL` stays a client constant while the server also serves the value, leaving two sources for one URL (`packages/ui/src/constants.ts:15`).

- 🟢 The server-supplied `href` is rendered without the `isExternalHttpUrl` guard that `packages/ui/src/modules/agents/components/release-notes-link.tsx:13` applies (`packages/ui/src/modules/home/components/compute-widget.tsx:66`). Merged into the 🟡 above.

### Verdict
COMMENT — the wiring is correct; the config schema is the one thing worth changing, since a typo'd operator value either stops the api-server or reaches the DOM unguarded.

---

packages/api-server-api/src/modules/links/types.ts
kapetr: Both mechanics confirmed (boot crash on typo; z.url() admits any scheme), but we went with `z.string().nullable()` in a308a428 instead of `.catch(null)`: the value is operator-set Helm config at the same trust level as `brand.*` (all plain strings), so scheme-guarding adds nothing here, and silently swallowing a typo'd value into the fallback hides the misconfiguration. A typo now yields a dead link — debuggable by clicking — never a failed boot.

kapetr:
@dam-code-guardian Re your 🟡 on `linksSchema`: fixed in a308a428, but with `z.string().nullable()` rather than the suggested `z.httpUrl().nullable().catch(null)`.

Reasoning:
- The boot-crash half of your finding was right — `loadConfig()` parses unguarded, so `z.url()` let a typo'd cosmetic link crash-loop the api-server. Plain string removes that failure mode entirely.
- We rejected `.catch(null)` because it silently replaces a typo'd operator value with the bundled fallback — misconfiguration with no signal is worse than a dead link you can debug by clicking.
- The `javascript:` scheme concern doesn't apply at this trust level: the value comes from Helm values, same trust as `brand.*` (also plain strings, including operator-supplied SVG) — anyone who can set it already controls the UI. `release-notes-link.tsx`'s `isExternalHttpUrl` guards GitHub-derived data, which is a different trust class.

So: cosmetic operator config follows the `brand.*` precedent — unvalidated string, degrades to a dead link, never a failed boot.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `a308a42`

## PR #3457: feat: operator-configurable compute request link
**Author:** kapetr | **Branch:** fix/home-request-more-link → main | **Changes:** +53 −4 (12 files)

### Summary
Complete re-review at the label's request. The one blocking finding is resolved: `linksSchema` is now `z.string().nullable()`, so a typo'd `LINKS_COMPUTE_REQUEST` can no longer fail `loadConfig()` and crash-loop the api-server. Nothing else changed since `c522777`.
_(Suppressed 1 finding per PR-local overrides: scheme validation on `computeRequest`. Suppressed 1 finding per PR context: the UI treats a server `null` as "use the bundled default".)_

### Changes since last review
Previous HEAD: `c522777` (2026-08-25T17:48:35Z) — verdict COMMENT

- ✅ **Fixed:** `z.url()` aborted boot on a typo'd cosmetic link; the plain string removes that failure mode (`packages/api-server-api/src/modules/links/types.ts:4`)

**Scheme validation — resolved, not carried over.** The author's reasoning is accepted, and the precedent checks out: `BRAND_ICON_SVG` is served raw as `image/svg+xml` with no sanitization (`packages/api-server/src/apps/api-server/routes/brand.ts:68`), so operator Helm values already control the UI at this trust level. `isExternalHttpUrl` guards GitHub-derived data, which is a different trust class. The rejection of `.catch(null)` is also sound — a dead link is debuggable, a silently swallowed typo is not. Recorded for this PR, so no future review re-raises it.

### Findings

✅ No findings.

### Issue Fit
- 🟢 **Suggestion:** The body says the fallback points at #dam-community, but the constant is an opaque archive URL (`packages/ui/src/constants.ts:15`), so the destination is not checkable from the code.
- 🟢 **Suggestion:** Open issue #3458 asks for the same widget action, but its ask is a structured in-app request flow, and this link repoint is listed there as an open question. A reference to #3458 keeps the interim fix from reading as the full delivery.

No undeclared scope: the `links` module, tRPC query, Helm value, and env plumbing are all declared in the body.

- 🟢 **Suggestion:** `links` enters `ApiContext` as a plain data value, while every other member is a service interface (`packages/api-server-api/src/context.ts:51`, `packages/api-server-api/src/modules/links/router.ts:5`). A later per-user or refreshable link would change every consumer.

The skill reported the missing scheme check as 🟡. Suppressed per this PR's recorded resolution above.

- 🟢 **Suggestion:** `useLinks()` is `undefined` on first render, so the anchor mounts with the bundled `COMPUTE_REQUEST_URL` and swaps when the query resolves (`packages/ui/src/modules/home/components/compute-widget.tsx:44`, `:66`). This is the declared fallback state rather than a defect, and the widget stays unmounted until `useBudgetReserved()` resolves, so the window is the gap between two concurrent requests. A boot-time prefetch of `links.all` would close it.

The skill reported the unguarded `href` as 🟡. Suppressed per this PR's recorded resolution above.

### Verdict
APPROVE — the boot-crash risk is gone, and the remaining scheme question is a maintainer decision with a verified precedent. Only suggestions are open.

---

jezekra1 (approved):

PR #3465, opened 2026-08-26
feat(providers): one Bob connection drives the bob harness and Pi agents

### LINKED ISSUE #3454: feat(providers): Bob as an inference provider for Claude Code (OpenAI-compatible preset)
## Context

Software Group wants to run DAM but everyone has **Bob**, not IBM LiteLLM (they're not Research). Ask: can Claude Code draw inference from Bob?

## Correction (thanks @U04GFSYJR19)

Claude Code talks the **Anthropic Messages API** (`/v1/messages` via `ANTHROPIC_BASE_URL`), **not** OpenAI Chat Completions. So Bob being "OpenAI-compatible" is a red herring for CC — the OpenAI route doesn't help us.

The relevant fact: Bob **also exposes an Anthropic Messages-compatible route**. Per the `pi-bob` docs, its `anthropic-messages` adapter targets `/inference/v1/messages` (a trailing `/inference/v1` is normalized to `…/messages`). *That* is what CC needs.

## Endpoint contract (Anthropic path — the one that matters for CC)

- `ANTHROPIC_BASE_URL` → `https://api.us-east.bob.ibm.com/inference/v1` (adapter sends to `/inference/v1/messages`)
- Auth: `Authorization: Apikey {value}` — **identical to what the existing `bob` harness preset already injects** (`packages/api-server-api/src/modules/connections/providers.ts`, `BOB_HOST` + injection ~L180).
- Default model alias: `premium` → Claude Sonnet 4.5 (200k ctx / 8k max output). Other aliases: `pro`, `flash`, `flash-lite`, `bob-3-pro-preview`.
- Tenant routing via non-secret headers `x-instance-id` / `x-team-id` — already carried as `BOB_INSTANCE_ID` / `BOB_TEAM_ID`.

## Proposed change

Add a Bob-inference preset whose `defaultEnvMappings` mirror the **Anthropic** half of `ibmLitellmEnvMappings` (`ANTHROPIC_AUTH_TOKEN` + `ANTHROPIC_BASE_URL`), not the OpenAI half:

- `ANTHROPIC_AUTH_TOKEN` → placeholder (Envoy swaps to the real key on the wire, as today)
- `ANTHROPIC_BASE_URL` → `https://api.us-east.bob.ibm.com/inference/v1`
- reuse the existing `Authorization: Apikey {value}` injection + twin `?key=` injection

Because this is a custom Anthropic upstream, the **ADR-067 local model gateway** kicks in automatically: it fronts the upstream, serves the renamed catalog to CC's `/model` picker, and forwards byte-for-byte. So model discovery is handled by existing machinery — no hand-pinned model ids.

## Open questions to verify

1. **Is Bob's `anthropic-messages` route actually reachable for our key type?** The pi-bob evidence is via SSO + API key; confirm an approved `general`/`personal` API key can hit `/inference/v1/messages`.
2. **Catalog discovery format.** ADR-067 gateway expects the upstream to serve a discoverable catalog; Bob's is `/inference/v1/model/info`. Confirm the gateway can consume it (or needs a small adapter).
3. Does the Anthropic route have the same rejected-fields quirks the OpenAI route documents (`store`, `prompt_cache_key`, `tools[].function.strict` → 422)? The gateway forwards byte-for-byte, so if CC sends anything Bob's Anthropic route rejects, it surfaces as a turn failure.

## Acceptance

- [ ] Bob-inference provider preset selectable in Settings → Providers
- [ ] Claude Code agent using it completes a turn against `premium` via `/inference/v1/messages`
- [ ] ADR-067 gateway serves Bob's catalog to the `/model` picker
- [ ] Tool-calling turn works
- [ ] `x-instance-id` / `x-team-id` routing verified for `general`-type keys

_Filed from a channel discussion with Software Group interest; corrected after @U04GFSYJR19 flagged the OpenAI-vs-Anthropic API distinction._

pilartomas:
Closes #3454.

Lets Pi agents draw inference from Bob — by extending the **existing Bob
connection**, with nothing new to choose at connect time. The issue asked for
Claude Code; live verification against Bob's endpoint showed neither Claude
Code nor Codex can be served (see below), and that the issue's original
`OPENAI_*` instinct was half right: Bob's key-accessible surface is
OpenAI-compatible, but only the Chat Completions family — which today, among
the platform's harnesses, only Pi speaks.

## What this does

Bob Shell and Bob inference are the same secret: the same Bob API key of type
Inference, on the same host, with the same `Authorization: Apikey {value}`
injection. The only thing that differs is which env vars the connection
contributes, and each set is inert to the other harness.

So `bobEnvMappings` now contributes, alongside `BOBSHELL_API_KEY`:

- `OPENAI_PROXY_URL` / `OPENAI_PROXY_MODEL` / `OPENAI_PROXY_CONTEXT_WINDOW` /
  `OPENAI_PROXY_MAX_TOKENS` — the pi-agent dynamic-provider contract,
  mirroring the LiteLLM preset. Pi needs no key env: its `dummy-placeholder`
  default is exactly right under gateway-side injection, and its `/models`
  discovery degrades to the pinned model where Bob denies the route. The
  limits are the ones Bob's `premium` (Claude Sonnet 4.5) enforces.
- **No `ANTHROPIC_*`, no codex-facing `OPENAI_*` — a test pins both
  absences.** Bob serves neither protocol to API keys, so wiring either
  harness at Bob would fail every turn (or, for codex, refuse to boot).

No new provider, no new template, no new form. The connect form is one paste,
exactly as before. Provider and template renamed "Bob Shell" → "Bob", since it
is no longer shell-only.

One connection also removes a hazard: two separate Bob connections would
collide on `Authorization` for the same host — the controller keeps one and
silently drops the other (`envoy.go`, first-wins). One connection cannot
collide with itself.

## Verified live against Bob's endpoint

Probing `api.us-east.bob.ibm.com` with a real Inference key, from a cluster
agent pod and from outside:

- **`/inference/v1/chat/completions` answers an Inference key directly**
  (HTTP 200, completion from `premium`), with `Authorization: Apikey` and no
  tenant headers. `/inference/v1/model/info` serves the catalog the same way.
- **The Anthropic route is closed to API keys.** `/inference/v1/messages`
  exists but answers `403 Access denied` to a valid key that authenticates
  (garbage keys get 401 "Invalid API Key", no auth gets 401 "Authentication
  required"); `Bearer` on that route expects a JWT — it serves Bob's own
  signed-in clients. No `anthropic` passthrough prefix exists. Tenant headers
  do not change the verdict. **Claude Code is therefore undeliverable.**
- **The Responses API is closed too** (`/inference/v1/responses` → 403), and
  current codex refuses `wire_api = "chat"` at config load — it speaks only
  the Responses API. **Codex is therefore undeliverable**: a protocol
  mismatch in both directions.
- **Cloudflare gates `/inference/*` by user agent**: only UAs containing
  lowercase `bob` reach the app (`curl`, `node`, browser UAs, `claude-cli`,
  and capitalized `Bob/2.0` all get the challenge page). Pi's UA would be
  blocked, so the connection injects a constant `User-Agent:
  bob-platform-egress/1.0` at the egress gateway — an honest identifier that
  satisfies the edge rule; authorization stays enforced server-side by the
  key.

## Constant-header injection

The platform previously had no way to put a static non-secret header on the
wire. Now an `egress-inject` whose `valueFormat` carries no `{value}`
placeholder injects the literal as a constant header, riding the existing SDS
rail unchanged — the controller already renders whatever the annotation
names, and `credential_injector` runs with `overwrite: true`. Constant entries
get their own SDS file key (host + header) so they cannot shadow the host's
credential injection; a test pins one-SDS-file-per-injection.

`docs/architecture/connections.md` records the capability.

## Existing Bob connections upgrade on re-attach — no key re-entry

Contributions are a stored projection, so a Bob connection created before this
change holds only the shell wiring — and asking users to delete and re-create
the connection (re-enter the key, lose the grants) was not acceptable.

Applying an agent's grants now re-renders each granted header connection from
the template's current defaults and the stored inputs, reading the credential
from the connection's own Secret. If the projection changed, the Secret's SDS
files and env-mapping annotations are rewritten in place, the row is
persisted, and the normal fan-out delivers it. Re-attaching the Bob secret to
an agent (or re-saving Configure Agent → Secrets) is the whole upgrade; the
provider and its key are never touched. A projection that cannot be
re-rendered degrades to today's behaviour and never blocks the grant.
Credential rotation stays a pure rotation, keeping the documented invariant
that a rotation moves nothing but the credential.

`docs/architecture/connections.md` records the re-render semantics.

## Model discovery

The claude-code model gateway now falls back to `/v1/model/info` when
`/v1/models` is absent, and tolerates numeric model ids again. This benefits
any Anthropic-compatible custom upstream (e.g. the LiteLLM preset); Bob itself
does not route through that gateway.

## Tests

`connections-bob-template.test.ts` asserts the merged template: inference is
*not* a separate template, shell and Pi wiring are contributed together, the
base URL composes to the verified `/inference/v1/chat/completions`, the real
key never lands in an env placeholder (Pi gets no key env at all), the model
and limit pins, **no `ANTHROPIC_*`/`CLAUDE_*`/`CODEX_*`/codex-facing
`OPENAI_*` var is set**, the constant-UA injection carries no `{value}` and
matches the edge rule, every injection bakes into its own SDS file, the shell
pins still work alongside the inference wiring, and no env name is contributed
twice.

`connections-refresh-on-grant.test.ts` covers the in-place upgrade: a stale
connection gains the inference wiring when its grant is applied (pins
preserved, secret annotations rebuilt, fan-out sees the new set), an
up-to-date projection is untouched, stale SDS fields are dropped while
unrelated fields are carried, an impossible re-render never blocks the grant,
and rotation stays contribution-free.

Two parity gaps also gained coverage: every preset has a Settings row, and
every preset template id has a catalog entry.

`api-server-api`, `api-server` and `ui` typecheck, lint and format clean;
`ui:test` (334) and `api-server:test` (1186) pass; the comment-types and docs
gates pass.

## Trade-off worth a reviewer's eye

Any agent granted the Bob connection now gets `OPENAI_PROXY_*` set. For a
bob-harness agent that is inert; for a Pi agent it is the feature. An agent
granted Bob *and* the LiteLLM provider gets a colliding `OPENAI_PROXY_*` set
(first-occurrence-wins) — the platform's answer remains one model provider per
agent.

## Not covered — needs a live cluster run

- A pi-agent turn end-to-end via `OPENAI_PROXY_*` (the wire, auth, UA, model,
  and route were verified with direct HTTP).
- Claude Code and Codex on Bob are **not deliverable** until IBM opens the
  Anthropic Messages or Responses route to API keys — worth raising with the
  Bob team. The Claude Code wiring this PR originally carried is preserved in
  git history (`d5cd713`) should that happen.


packages/agents/claude-code/model-gateway.mjs
```diff
@@ -34,37 +34,55 @@
...
+
+const isEmbedding = (m, id) =>
+  [id, m?.mode, m?.type, m?.model_info?.mode].some((f) =>
+    String(f ?? "")
       .toLowerCase()
       .includes("embedding"),
   );
 
+async function fetchCatalogFrom(path) {
+  const r = await fetch(`${UPSTREAM}${path}`, {
```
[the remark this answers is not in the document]
pilartomas: False positive by design: the "file data" is `~/.platform/pod-service-env.json`, the pod's own runtime-written env snapshot — the SIGHUP reload path reads the same values the process was started with. The token it carries is an inert placeholder (the real credential is injected at the Envoy egress gateway and never exists in the agent pod), and the upstream request is still subject to the gateway's egress allowlist, so tainted file data can neither leak a secret nor reach an arbitrary host. Anything able to write that file already controls the pod. Same flow exists on main (no alert there — the refactor merely moved the fetch into a new function).

packages/agents/claude-code/model-gateway.mjs
```diff
@@ -34,37 +34,55 @@
...
+    headers: {
+      authorization: `Bearer ${TOKEN}`,
+      "x-api-key": TOKEN,
+      "anthropic-version": "2023-06-01",
+      accept: "application/json",
+    },
```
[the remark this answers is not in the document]
pilartomas: Same as the sibling alert: the source is the pod's own runtime-written env snapshot, the token on these headers is a placeholder overwritten at the Envoy gateway, and the destination host is constrained by the egress allowlist. False positive by design.

packages/api-server/src/modules/connections/services/connections-service.ts
```diff
@@ -159,6 +162,90 @@ export function createConnectionsService(deps: {
...
+      !template ||
+      template.authKind !== "header" ||
+      conn.auth.kind !== "header"
+    ) {
+      return conn;
+    }
+    try {
+      const value = await deps.secretStore.getField(conn.auth.valueRef);
+      if (!value) return conn;
+      const input = connectionCreateInputSchema.parse({
```
dam-code-guardian, line 195: 🟡 **Warning:** The re-render validates persisted state with `connectionCreateInputSchema.parse`, then `buildConnection` re-validates each `configInput` against the template's *current* `pattern`/`enumValues` and **throws** on a mismatch. The `catch` at :236 swallows that into a `warn` log and returns the connection unchanged, so the grant succeeds and the connection keeps its stale contributions with no operator-visible signal.

This reaches the feature's own purpose: the connections that need the upgrade are the oldest ones, which are the most likely to hold a pin the template has since narrowed. `BOB_LEGACY_CHAT_MODES` proves `code`/`advanced` were once storable `chatMode` values, and `normalizeBobChatMode` is applied when *reading* pins, never before this validation. I transcribed the loop: `chatMode: "code"` and `maxCost: "0"` both throw, so those connections never gain the `ANTHROPIC_*` wiring, while a modern row upgrades. The failure is indistinguishable from a transient one in the log.

**Fix:** Re-projection from stored state should not be gated on the create-time input schema — normalize the stored pins first (`normalizeBobChatMode` already exists for exactly this), or separate "stored state rejected" from a retryable error so a permanent mismatch is visible rather than logged as a warning and dropped.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `90dc4ab0`

## PR #3465: feat(providers): one Bob connection drives the bob harness and Claude Code
**Author:** pilartomas | **Branch:** feat/bob-inference-provider → main | **Changes:** +670 −49 (14 files)

### Summary
One Bob connection now contributes both the shell wiring and the Claude Code `ANTHROPIC_*` wiring, and applying an agent's grants re-renders a header connection's stored contributions so pre-existing Bob connections upgrade in place without re-entering the key.

### Findings
- 🟡 **Warning:** legacy stored pins block the in-place upgrade silently (`packages/api-server/src/modules/connections/services/connections-service.ts:195`)
- 🟡 **Warning:** whole-Secret rewrite outside the rotation lock (`packages/api-server/src/modules/connections/services/connections-service.ts:220`)
- 🟡 **Warning:** the new `ANTHROPIC_*` names collide with `ibmLitellmEnvMappings()`, and the winner is decided by secret name order (`packages/api-server-api/src/modules/connections/providers.ts:57`)
- 🟢 **Suggestion:** `Closes #3454` will retire the tenant-routing acceptance item with no follow-up issue on file. I searched open issues; none exists. Dropping the keyword to a plain reference keeps that item open.
- 🟢 **Suggestion:** `docs/architecture/security-and-credentials.md:473` names stale SDS entries a known trigger for degrading a host to an allow-only chain. The grant-time re-render now self-heals that condition. The sentence is pre-existing on `main` and that page is untouched here, so this is pre-existing drift, not introduced.
- ✅ **Looks good:** the two classes I carried from the earlier aborted fan-out are answered. `contributions` being persisted no longer strands existing connections, and rotation stays contribution-free. I also re-measured the `/inference` base URL composing to `/inference/v1/messages` with no doubled `/v1`, and the catalog extractor's index alignment — both hold.

- 🟢 See the pre-existing SDS drift above. Baselined against `main`, so it is not introduced by this PR.

### Issue Fit
- 🟢 The acceptance list is delivered except tenant routing, which the body declares as needing a new static-header contribution kind. That deviation is declared, so it is not undeclared scope; the missing follow-up issue is the only residual.

- 🟡 The three warnings above were reported here too and are merged into `### Findings`.
- 🟢 **Suggestion:** `fetchCatalogFrom` maps and then filters on the post-map index (`data[i]`), which is correct only because `map` preserves length. I measured it and alignment holds, so this is a fragility note, not a defect (`packages/agents/claude-code/model-gateway.mjs:65`).

✅ No findings.

### Verdict
COMMENT — the design is sound and the wiring verified, but the advertised in-place upgrade silently skips any connection whose stored pins no longer satisfy the current template.

---

packages/api-server/src/modules/connections/services/connections-service.ts
```diff
@@ -159,6 +162,90 @@ export function createConnectionsService(deps: {
...
+          c.kind === "mcp-entry" ? { ...c, name: conn.name } : c,
+      );
+      if (isDeepStrictEqual(next, conn.contributions)) return conn;
+
+      const carried = Object.fromEntries(
+        Object.entries(
+          (await deps.secretStore.get(conn.auth.valueRef)) ?? {},
+        ).filter(([key]) => !isSdsFieldKey(key)),
+      );
+      await deps.secretStore.put(
```
dam-code-guardian, line 220: 🟡 **Warning:** This is a read-modify-write of the whole Secret (`get` at :215 → `put` at :220) with no lock, and `carried` includes the `value` field. `put` replaces the Secret wholesale, so a credential rotation landing between the read and the write is rolled back to the pre-rotation key. The rotation path uses `putFields`, which merges, so it does not collide with itself — this new path is the only whole-Secret rewrite outside create.

Both surfaces are reachable concurrently: `connections.update` (rotate) and `connections.setAgentConnections` (grant) are separate tRPC procedures. `connectionLock` already exists in this service but is currently used only for the GitHub App mint.

**Fix:** Any path rewriting a whole credential Secret should hold the same `connectionLock` key the rotation path uses, or write only the fields it owns — `putFields` with just the new SDS keys plus an explicit delete of the stale ones — so a concurrent rotation is never replayed away.

packages/api-server-api/src/modules/connections/providers.ts
```diff
@@ -50,9 +50,38 @@ export interface BobModelPins {
 export const BOB_HOST = "api.us-east.bob.ibm.com";
 const BOB_PLACEHOLDER = "dummy-placeholder";
 
+export const BOB_INFERENCE_BASE_URL = `https://${BOB_HOST}/inference`;
+export const BOB_INFERENCE_PREMIUM_MODEL = "premium";
+export const BOB_INFERENCE_FAST_MODEL = "flash";
+
+export function bobInferenceEnvMappings(): EnvMapping[] {
```
dam-code-guardian, line 59: 🟡 **Warning:** These three names — `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_BASE_URL`, `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS` — are already contributed by `ibmLitellmEnvMappings()` at :22–24. The controller's `credentialEnvVars` dedups first-occurrence-wins over secrets sorted by name (`packages/controller/pkg/reconciler/envoy.go:180`), so an agent granted both connections silently gets one upstream and not the other.

Your body pre-declares the ambiguity as a known trade-off, and the "one model provider per agent" rule is a fair answer — so this is not news. What I measured is narrower: the outcome is not arbitrary, it is *fixed*. The secret name is `platform-secret-<purpose-slug>-<random>`, and the slug sorts before the digest, so `connection-bob` always precedes `connection-ibm-litellm` (20000/20000 trials). Bob silently wins every time, including for an agent that was working against LiteLLM before its Bob grant existed.

**Fix:** The determinism is the part worth closing, not the trade-off. Detect the collision when the grant is applied and surface it — a rejected grant or a warning naming the losing provider — so a silent upstream swap on an existing agent cannot happen unnoticed.

PR #3467, opened 2026-08-26
feat(artifacts): let private HTML buttons send chat prompts

### LINKED ISSUE #2887: Private artifacts can call back into their agent
## Problem

A published artifact is frozen at publish time. The page can't ask its agent for anything afterwards: it can't fetch fresh data, can't trigger work, and can't accept input that reaches the agent. So an artifact is a document, when it could be an app.

Concrete scenario: an agent publishes a dashboard of your open pull requests. It's stale the moment it's published. To refresh it, you have to go back to the chat and ask the agent to republish. What you actually want is a refresh button on the page itself, or a form on the page that kicks off agent work with the values you typed in.

## Proposed solution

A private artifact viewed by its owner can call back into the agent that published it. The page sends a request, the agent handles it and responds, and the page uses the result: refresh data, run an action, accept input.

This makes personal apps possible: dashboards that update themselves, small tools with buttons and forms, pages that ask the agent to do something on your behalf.

## Security constraint: private only, enforced

This must work **only** for private artifacts viewed by their owner. This is not a soft scope choice, it's the security boundary of the feature.

An agent holds the owner's connections, credentials, and permissions. If a publicly shared artifact could call back into it, anyone with the link could drive the owner's agent: read through its connections, trigger actions as the owner, exfiltrate data. A public callback path is a direct attack vector.

Concretely:

- A callback must be rejected unless the artifact is private and the caller is the authenticated owner.
- Making an artifact public must not leave any callback path open on the shared page.
- The public share surface stays exactly as static as it is today.

### Out of scope

- Callbacks from shared artifacts. Letting other viewers trigger the author's agent needs a different security model (tracked in epic #2884).

### Dependencies

- Part of epic #2884.

tomkis:
When Interactive artifacts is enabled, a button in a private HTML artifact can call `platform.sendPrompt(text)` to submit a normal message into the open chat with its publishing agent. The agent replies in chat and can publish an updated artifact normally.

Callbacks work from the latest version in the docked preview, including fullscreen. The app validates the sending iframe and prompt, checks that the preview belongs to the current agent, and uses the existing chat sender for queuing and errors. If the chat cannot accept a prompt, the app shows a notification explaining why and how to retry. Library previews, historical versions, and disabled-feature previews cannot send prompts.

Interactive HTML is an explicit opt-in at creation and remains private; existing artifacts default to noninteractive. The shipped authoring skill documents the button API and its limits.

This partially implements #2887: artifacts can send prompts into chat, but they cannot receive and consume the agent's response directly. The result path remains unimplemented, and #2887 stays open to track that remaining work.

Validation: affected-package type/lint/format checks, database generation check, comment and skill-manifest checks, documentation checks, 106 focused tests, and an isolated browser check of the actual sandboxed iframe and callback gates. The browser check exercises routing to the chat callback; it does not run a live agent turn. The refusal fix additionally passed 26 focused tests plus UI type/lint/format, comment, and documentation-size checks.


packages/ui/src/modules/artifacts/components/deferred-frame.tsx
```diff
@@ -31,6 +48,30 @@ export function DeferredFrame({
...
+    pageWindow.postMessage({ type: ARTIFACT_BRIDGE_CONNECT_TYPE }, "*", [
+      channel.port2,
+    ]);
+
+    return () => {
+      window.removeEventListener("message", readRequest);
+      bridge.onDisconnect(send);
+      channel.port1.close();
+    };
+  }, [loaded, bridge]);
```
dam-code-guardian, line 73: 🟡 **Warning:** The handshake effect is keyed on `loaded`, which is set once and never reset, and `html` is not a dependency. When `srcDoc` changes for a mounted frame — the agent publishes a new version, or the owner saves an edit — the iframe reloads, `onLoad` writes `true` over `true`, the effect does not run again, and the new document never receives `port2`. Every button on the reloaded page is then dead, with no error anywhere. This is one of three paths on the same invariant: an answer must reach the page that asked (see also `use-artifact-bridge.ts:78` and `api/queries.ts:100`).

**Fix:** Tie a per-document effect to document identity, not to a one-way flag — key the iframe on `html`, or reset `loaded` when `html` changes — so the handshake runs once per loaded document.

packages/ui/src/modules/artifacts/hooks/use-artifact-bridge.ts
```diff
@@ -0,0 +1,195 @@
...
+  }, [askable, artifact, openConversation]);
+
+  const hold = useCallback((next: PendingRequest | null) => {
+    pendingRef.current = next;
+    if (next === null) lastProgressRef.current = null;
+    setPending(next);
+  }, []);
+
+  const send = useCallback((reply: ArtifactBridgeReply) => {
+    sendRef.current?.(reply);
```
dam-code-guardian, line 78: 🟡 **Warning:** `send` is a no-op while no frame is connected, and the resolution effect calls `hold(null)` whatever `send` did. If the answer lands while the frame is between connections — the person toggled full screen, so one frame is handed the bridge and the other is not, or the document reloaded — the answer is dropped, the pending state is cleared, and the status bar goes empty. The agent has done the work and the page shows nothing. Same invariant as `deferred-frame.tsx:73` and `api/queries.ts:100`.

**Fix:** Never assume a live transport at resolve time: keep each request's terminal outcome in state, keyed by its `ref`, and replay it from `onConnect` when a frame connects.

packages/ui/src/modules/artifacts/api/queries.ts
```diff
@@ -88,3 +90,30 @@ export function useFolderShareUrl(id: string | null) {
...
   });
 }
+
+export function useArtifactRequest(requestId: string | null) {
+  return useQuery({
+    ...trpc.artifactLibrary.requests.get.queryOptions(
+      requestId ? { requestId } : skipToken,
+    ),
+    staleTime: 0,
+    retry: 2,
```
dam-code-guardian, line 100: 🟡 **Warning:** This query has no poll and no deadline, so the row is re-read only when the `artifactRequest` live event invalidates it — and only a settlement raises that event. Two consequences. If the live connection drops while a request is in flight, the spinner runs forever and every later click is refused `busy` by the hook, so the page is stuck until it is reopened. And because `markDelivered` raises no event, the `delivered` state never reaches the client, so the `running` progress that `deriveRequestProgress` returns and `SKILL.md` documents to page authors is unreachable. Same invariant as `deferred-frame.tsx:73` and `use-artifact-bridge.ts:78`.

**Fix:** A query that waits on a server-side state transition needs its own liveness — poll while the request is in flight and settle it locally on a deadline — and every state the client renders needs an event that announces it.

packages/ui/src/modules/artifacts/hooks/use-artifact-bridge.ts
```diff
@@ -0,0 +1,195 @@
...
+    if (!current?.requestId) return;
+    if (unreadable) {
+      hold(null);
+      refuse(current.ref, describeFailure("wake_failed"));
+      return;
+    }
+    if (!row || row.id !== current.requestId) return;
+    if (row.state === "answered") {
+      hold(null);
+      send({
```
dam-code-guardian, line 146: 🟡 **Warning:** Nothing clears `failure` when a request is answered. A second ask while the first is in flight is refused `busy` and sets `failure`; the status bar hides it while `progress` is non-null, and then shows it the moment the answer lands. So a successful answer ends with "This page is already waiting on an answer" on screen, until the person dismisses it. Any page that lets two buttons be pressed reaches this.

**Fix:** Clear the failure state whenever a request settles — a failure belongs to the one `ref` it refused and must not outlive it.

packages/api-server/src/modules/artifact-library/services/artifact-requests-service.ts
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  async function deliver(
+    row: ArtifactRequestRow,
+    page: ArtifactRow,
+    binding: Exclude<RequestBinding, { kind: "not-bound" }>,
+  ): Promise<void> {
+    const sessionId =
+      binding.kind === "pin"
+        ? ((await library.pinSession(
```
dam-code-guardian, line 150: 🟡 **Warning:** The binding is written before it is checked. `pinSession` persists the offered conversation, and only then does `checkBinding` ask whether the agent lists it. If the offered id is not in that list — the conversation was deleted in another tab, so the app still offers it — the page is left permanently bound to it: `resolveBinding` returns the pinned id from then on, `pinSession` writes only when the column is null, and no other code path clears it. Every later ask fails `session_deleted`, from any chat, and the page can never ask again.

**Fix:** Persist a permanent binding only after the check that would reject it has passed — resolve and verify the conversation first, then pin the page to it.

packages/api-server/src/modules/artifact-library/services/artifact-requests-service.ts
```diff
@@ -0,0 +1,313 @@
...
+      artifactId: row.artifactId,
+      agentId: row.agentId,
+      sessionId,
+      task,
+    });
+    if (outcome.ok) {
+      await requests.markDelivered(row.id, owner);
+      return;
+    }
+    await settleAs(row.id, outcome.reason);
```
dam-code-guardian, line 187: 🟡 **Warning:** By the time this settles the request `failed`, `deliver` has already bumped the `artifact-request` event into the runtime channel and enqueued it. The event is not retracted, so it stays deliverable until its 15-minute TTL: the next wake for any reason replays it, the agent resumes the bound conversation, reads the prompt, does the work, and `answer_artifact_request` refuses its answer because the request is already settled. The owner pays for a turn nobody is waiting for, in the conversation they are reading. `cancel` has the same shape (`:275`).

**Fix:** A compensated transition retracts the side effects it already committed — remove the queued runtime event when a request settles before pickup, or make the runtime handler read the request state before it prompts.

packages/api-server/src/modules/artifact-library/services/artifact-requests-service.ts
```diff
@@ -0,0 +1,313 @@
...
+      }
+    },
+
+    async get(requestId) {
+      const row = await requests.get(requestId, owner);
+      return row ? toArtifactRequest(row) : null;
+    },
+
+    async cancel(requestId) {
+      const cancelled = await settleAs(requestId, "cancelled");
```
dam-code-guardian, line 275: 🟡 **Warning:** Same defect class as `:187`. Cancelling settles the row while the queued `artifact-request` event stays in the runtime channel until its TTL, so the agent is still woken and prompted for the cancelled request and its answer is then refused.

**Fix:** A compensated transition retracts the side effects it already committed — remove the queued runtime event when a request settles before pickup, or make the runtime handler read the request state before it prompts.

packages/ui/src/modules/artifacts/hooks/use-artifact-bridge.ts
```diff
@@ -0,0 +1,195 @@
...
+    },
+    [createRequest, hold, refuse, send],
+  );
+
+  useEffect(() => {
+    const current = pendingRef.current;
+    if (!current?.requestId) return;
+    if (unreadable) {
+      hold(null);
+      refuse(current.ref, describeFailure("wake_failed"));
```
dam-code-guardian, line 140: 🟡 **Warning:** A read failure of the request row — offline, or a 5xx that survives `retry: 2` — is converted into the domain reason `wake_failed`, and `failureFromError` uses the same reason for any unnamed mutation error. The person is told "The agent could not be woken", which the client cannot know, and the page receives `reason: "wake_failed"` in the documented refusal set, so a page author's branch on it fires for a network fault.

**Fix:** Keep transport failure a distinct outcome from a server-declared reason; never reuse a member of a domain enum as the fallback for "we could not reach the server".

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `7aefcd5`

## PR #3467: Interactive Artifacts: private pages can call back into their agent
**Author:** tomkis | **Branch:** feat/2887-interactive-artifacts → main | **Changes:** +11505 −69 (80 files)

### Summary
A private HTML artifact can now ask the agent that published it to do something, and the answer returns to the page without a reload. The feature adds a request table with its own in-flight and rate invariants, a new runtime event kind, an injected page shim, and an in-app bridge.
_2 generated/lockfile file(s) not reviewed: packages/db/drizzle/meta/0047_snapshot.json, packages/db/drizzle/meta/0048_snapshot.json._
_21 suggestion(s) dropped under the 🟢 budget._

### Findings
- 🟡 **Warning:** An answer can fail to reach the page that asked — three paths on one invariant (`packages/ui/src/modules/artifacts/components/deferred-frame.tsx:73`, `packages/ui/src/modules/artifacts/hooks/use-artifact-bridge.ts:78`, `packages/ui/src/modules/artifacts/api/queries.ts:100`)
- 🟡 **Warning:** A `busy` refusal stays on screen after the answer arrives (`packages/ui/src/modules/artifacts/hooks/use-artifact-bridge.ts:146`)
- 🟡 **Warning:** The first ask pins the page before its conversation is checked (`packages/api-server/src/modules/artifact-library/services/artifact-requests-service.ts:150`)
- 🟡 **Warning:** A settled request leaves its queued delivery in the runtime channel (`packages/api-server/src/modules/artifact-library/services/artifact-requests-service.ts:187`, `packages/api-server/src/modules/artifact-library/services/artifact-requests-service.ts:275`)
- 🟡 **Warning:** A failed read of the request row is reported as `wake_failed` (`packages/ui/src/modules/artifacts/hooks/use-artifact-bridge.ts:140`)
- 🟡 **Warning:** The architecture pages this feature belongs to now state the opposite of the code. The runtime page enumerates five event kinds and keeps its 2026-09-10 stamp, while this PR ships a sixth (`docs/architecture/runtime-delivery.md:49`). The features page says no current feature carries an agent surface and that the session-creation check returns with the next one that needs it, while this flag decides MCP tool registration at session creation (`docs/architecture/features.md:33`). The artifact-library page took a stamp bump in this PR and still lists publish, share, view and delete as the recorded activity moments (`docs/architecture/artifact-library.md:354`), still explains the viewer's placement by "the planned agent-calling bridge" this PR delivers (`docs/architecture/artifact-library.md:220`), and still names the retention sweep as the subsystem's scheduled job although a second periodic sweep is registered (`docs/architecture/artifact-library.md:325`).
  **Fix:** Update every enumeration or claim the change falsifies on the owning page, in the same PR, and bump that page's `Last verified:` — a bumped stamp above stale enumerations asserts a check that did not happen.
- 🟢 **Suggestion:** `allow-forms` is added to both artifact sandboxes for every artifact, the public share wrapper included, although an interactive page can never be shared and scripts already permit the same requests (`packages/api-server/src/modules/artifact-library/viewer/renderer.ts:104`, `packages/ui/src/modules/artifacts/components/deferred-frame.tsx:85`). Gate the token on the interactive path, so the shared surface stays as it is.
- ✅ **Looks good:** the sharing refusal runs inside the same transaction that locks the artifact row, and `interactive` is settled at create — the update input cannot carry it — so no later revision can turn a private page into a shareable one.

The skill reported five items on three pages. They are one defect class — an enumeration or a claim that the diff falsifies — so they are merged into one finding in `### Findings` with all five locations:
- 🔁 the runtime page's event-kind list, and its unchanged stamp (`docs/architecture/runtime-delivery.md:49`)
- 🔁 the features page's "no current feature carries an agent surface" (`docs/architecture/features.md:33`)
- 🔁 the artifact-library page's activity-moment list (`docs/architecture/artifact-library.md:354`)
- 🔁 the artifact-library page's "planned agent-calling bridge" (`docs/architecture/artifact-library.md:220`)
- 🔁 the artifact-library page's single scheduled job, and the missing request-lifecycle invariants (`docs/architecture/artifact-library.md:325`)

### Issue Fit
The PR delivers what issue #2887 asks, and the issue's security constraint holds: an interactive artifact is refused at create when it is not HTML or is public, the sharing refusal is enforced in the locked transaction, and the shim reaches only the in-app preview path.
- 🟢 the sandbox widening on the shared wrapper page, which the issue scopes out — reported in `### Findings`
- _3 further item(s) dropped under the 🟢 budget: the shared trigger-path rewrite and two undeclared additions, none of which carries a defect._

- 🟡 A settled request leaves its queued delivery in the runtime channel (`packages/api-server/src/modules/artifact-library/services/artifact-requests-service.ts:187`) — reported in `### Findings`, with the cancel path as its sibling.
  **Fix:** Retract the runtime-channel event when a request settles before pickup, or make the handler read the request state before it prompts.
- _5 finding(s) regraded to 🟢 after verification, and 10 suggestion(s) dropped under the 🟢 budget. Two are worth naming so nobody "fixes" them: the feature flag is documented as progressive disclosure and not authorization, so the service is right not to enforce it; and the expiry sweeper omits the actor on purpose, which is the documented rule that keeps timer-driven work out of the activity log._

- 🟡 The port handshake never runs again for a document that reloads in place (`packages/ui/src/modules/artifacts/components/deferred-frame.tsx:73`) — merged into the delivery finding in `### Findings`.
- 🟡 An answer is dropped when no frame is connected at the moment it lands (`packages/ui/src/modules/artifacts/hooks/use-artifact-bridge.ts:78`) — merged into the same finding.
- 🟡 A failed read of the request row is reported as `wake_failed` (`packages/ui/src/modules/artifacts/hooks/use-artifact-bridge.ts:140`) — reported in `### Findings`.
- _1 finding regraded to 🟢 after verification, and 8 suggestion(s) dropped under the 🟢 budget._

### Verdict
REQUEST_CHANGES — six blocking findings stand, and three of them let a paid-for answer disappear before the page that asked can render it.

### For the human reviewer
The page and the app talk over a private channel that exists only while one preview frame is open, and the app trusts that channel to be there at the moment an answer arrives. Three ordinary events break that trust: the page reloading, the person switching the preview to full screen, and the live connection to the server dropping. In each case the agent has done the work and the owner has paid for it, and the person watching sees nothing happen at all. The second decision to settle is that a page belongs to one conversation for the rest of its life: today that tie is written before the platform checks the conversation is still there, so a single unlucky ask can leave a page that can never ask again.


---

packages/ui/src/modules/sessions/hooks/use-chat-artifact-prompt.ts
```diff
@@ -0,0 +1,55 @@
...
+
+      if (
+        !isChat ||
+        !sessionId ||
+        !agentOperable ||
+        loadingSession ||
+        current.selectedAgent !== agentId ||
+        current.sessionId !== sessionId
+      )
+        return Promise.resolve();
```
dam-code-guardian: 🟡 **Warning:** Every guard miss answers with a resolved promise, so a prompt that was never sent is indistinguishable from one that was. The reachable states are real: the agent is stopped (`agentOperable` false), the session is a terminal (the docked panel stays open in the right dock while the terminal fills the centre), the session is still loading, or no session exists yet. In each of them the person clicks a live-looking button in the page and gets nothing — no chat entry, no toast, no page signal.

**Fix:** A gate that refuses work returns the refusal to its caller — a rejection or a result value — and the host shows one message for it; never resolve a prompt that was not sent. This is the only production site of the pattern in the PR (`git grep -nE -e 'return Promise\.resolve\(\)'` finds it plus two test stubs).

packages/ui/src/modules/artifacts/hooks/use-artifact-prompt.ts
```diff
@@ -0,0 +1,25 @@
...
+export function useArtifactPrompt(
+  frame: RefObject<HTMLIFrameElement | null>,
+  onSendPrompt: ((prompt: string) => Promise<void>) | undefined,
+): void {
+  useEffect(() => {
+    if (!onSendPrompt) return;
+    const receive = (event: MessageEvent) => {
+      const prompt = readArtifactPrompt(event, frame.current?.contentWindow);
+      if (prompt === null) return;
+      void onSendPrompt(prompt).catch(() => {
```
dam-code-guardian: 🟡 **Warning:** Same finding as `use-chat-artifact-prompt.ts:41`, at the handler that would report it. `sendPrompt` from `useAcpSession` catches every failure inside itself and reports it as a toast or an undelivered bubble, so it never rejects; this `catch` can therefore never run, and the toast it holds is unreachable. A refusal that arrives as a resolved promise has nowhere to surface.

**Fix:** A gate that refuses work returns the refusal to its caller — a rejection or a result value — and the host shows one message for it; never resolve a prompt that was not sent.

docs/architecture/artifact-library.md
```diff
@@ -68,6 +68,30 @@ surface is the per-agent platform MCP server, and by-link serving happens on
 two dedicated hosts — a **share host** for the pages and a **content host** for
 the framed artifact itself.
 
+## Interactive pages
+
+An HTML artifact can be declared **interactive** at creation and stays private.
+With the interactive-artifacts feature enabled, a button in its latest version
+can send a prompt into an existing open chat with its publishing Agent. The app
+checks the sending frame and the preview's agent before submitting through the
+normal chat path. Replies, queued turns and delivery errors appear in chat; the
```
dam-code-guardian: 🟡 **Warning:** Same finding as `use-chat-artifact-prompt.ts:41`, on the page that documents it. Genuine delivery errors do reach chat, because the chat sender reports them. A prompt the host refuses before it reaches the sender reaches nothing, so this sentence promises an outcome the code does not produce.

**Fix:** A gate that refuses work returns the refusal to its caller and the host shows one message for it; state the refusal case on this page once the code produces it.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `7073afa`

## PR #3467: feat(artifacts): let private HTML buttons send chat prompts

**Author:** tomkis | **Branch:** feat/2887-interactive-artifacts → main | **Changes:** +4235 −17 (37 files)

### Summary

This round replaces the durable artifact-request bridge with a one-way prompt into the open chat, and adds the `interactive` column, its migration, the render-time shim, the eligibility gate and the authoring skill. Every blocking finding of the previous round went away with the code that carried it.
_1 generated file(s) not reviewed: `packages/db/drizzle/meta/0049_snapshot.json`._
_17 suggestion(s) dropped under the 🟢 budget._

### Changes since last review

Previous HEAD: 7aefcd5 (2026-09-14T11:59:45Z) — verdict REQUEST_CHANGES — unreachable, reviewed the whole PR

- ✅ **Fixed:** an answer can fail to reach the asking page (`packages/ui/src/modules/artifacts/components/deferred-frame.tsx:73`)
- ✅ **Fixed:** busy refusal stays on screen after the answer (`packages/ui/src/modules/artifacts/hooks/use-artifact-bridge.ts:146`)
- ✅ **Fixed:** page pinned before its conversation is checked (`packages/api-server/src/modules/artifact-library/services/artifact-requests-service.ts:150`)
- ✅ **Fixed:** settled request leaves its queued delivery (`packages/api-server/src/modules/artifact-library/services/artifact-requests-service.ts:187`)
- ✅ **Fixed:** read failure reported as wake_failed (`packages/ui/src/modules/artifacts/hooks/use-artifact-bridge.ts:140`)
- ✅ **Fixed:** architecture pages contradict the shipped code (`docs/architecture/runtime-delivery.md:49`)
- ✅ **Fixed:** allow–forms widened for every artifact (`packages/api-server/src/modules/artifact-library/viewer/renderer.ts:104`)
- 🆕 **New:** a refused artifact prompt is silent (`packages/ui/src/modules/sessions/hooks/use-chat-artifact-prompt.ts:41`)
- 🆕 **New:** Closes #2887 while the result path is descoped (`packages/api-server/src/modules/artifact-library/viewer/bridge-shim.ts:11`)

### Findings

- 🟡 **Warning:** a refused artifact prompt is silent (`packages/ui/src/modules/sessions/hooks/use-chat-artifact-prompt.ts:41`, also `packages/ui/src/modules/artifacts/hooks/use-artifact-prompt.ts:15` and `docs/architecture/artifact-library.md:77`)

The claim that replies, queued turns and delivery errors all appear in chat is reported in `### Findings` — it is the page half of the silent-refusal warning. 3 more findings are 🟢 and were dropped under the suggestion budget; the largest is that the Experimental features page names one current feature, which was already true of the three ids on `main` before this round.

### Issue Fit

- 🟢 **Suggestion:** The linked issue asks for a round trip — the page sends a request, the agent responds, and the page uses the result. This round ships the request half only, and `Closes #2887` will close the issue with the result path undelivered. ADR-092 records that scope choice with its reason, so this is the author's call: either drop the closing keyword or re-scope the issue to prompt-only (`packages/api-server/src/modules/artifact-library/viewer/bridge-shim.ts:11`).

1 more finding is 🟢 and was dropped: the callback exists only in the docked chat preview and not in the library preview, which the PR body, the page and the ADR all state.

7 findings are 🟢 and were dropped under the suggestion budget. Two were graded down after a check in the tree: the create path and the sharing path express the interactive rule differently, but `artifactCreateVisibilitySchema` accepts only `private` and `public`, so the two cannot disagree today; and the share viewer reaches the correct answer through the renderer's `interactive = false` default, so no page gets the shim by accident.

The refusal warning and the unreachable error handler are reported together in `### Findings`. 6 more findings are 🟢 and were dropped. Two were graded down after a check in the tree: the interactive share dialog omits `onClose` on `Modal`, which is the module's own baseline — five sibling dialogs in the same directory do the same, so Escape does not close any of them; and the flag does not gate creation because a feature flag here is progressive disclosure, not authorization, which the Experimental features page states.

### Verdict

COMMENT — the rewrite closes every earlier finding, and one warning remains: a prompt the host refuses is dropped without a signal to the page, to chat or to the person.

### For the human reviewer

This round decides that an interactive page asks its agent by dropping a message into the open conversation, and that it never learns what happened next. That decision is sound for replies, because the conversation is where the person is already looking. It leaves one case open: when the conversation cannot accept the message — the agent is stopped, the session is a terminal, or no conversation is open yet — the click is thrown away and nobody is told, while the page's button still looks live. Settle whether a page whose buttons can go dead without a word is acceptable, or whether the app owes the person one visible refusal; the single code fix follows from that answer.


---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `0417293`

## PR #3467: feat(artifacts): let private HTML buttons send chat prompts
**Author:** tomkis | **Branch:** feat/2887-interactive-artifacts → main | **Changes:** +4343 −17 (38 files)

### Summary
This round makes a refused artifact prompt visible: the chat gate returns a reason for each of its five preconditions, the page's listener shows that reason, and the page describes the new behavior. The body now records #2887 as partly implemented.
_1 generated file not reviewed: `packages/db/drizzle/meta/0049_snapshot.json` — outside this round's range._
_11 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: 7073afa (2026-09-16T12:03:13Z) — verdict COMMENT

- ✅ **Fixed:** a refused artifact prompt is silent (`packages/ui/src/modules/sessions/hooks/use-chat-artifact-prompt.ts:41`)
- ✅ **Fixed:** Closes #2887 while the result path is descoped (`packages/api-server/src/modules/artifact-library/viewer/bridge-shim.ts:11`)
- 🆕 **New:** page edited today, stamp still 2026-09-16 (`docs/architecture/artifact-library.md:3`)
- 🆕 **New:** stubbed useEffect leaves unmount untested (`packages/ui/src/__tests__/unit/artifact-prompt-notification.test.ts:14`)

### Findings
- 🟢 **Suggestion:** This round edits the **Interactive pages** section of the page, and the `Last verified:` header still reads 2026-09-16. Move the stamp to the date of this edit (`docs/architecture/artifact-library.md:3`).
- 🟢 **Suggestion:** The new spec replaces React's `useEffect` with a collector, so the listener's own lifecycle is not exercised: no case shows that unmount removes the `message` listener, or that a changed `onSendPrompt` re-subscribes. A test through the real lifecycle covers both (`packages/ui/src/__tests__/unit/artifact-prompt-notification.test.ts:14`).

2 findings, both read; one is reported above. The Experimental features page still names one current feature while four flag ids exist. That was already true on `main` before this round, so it stays 🟢 and is carried in the deferred list.

### Issue Fit
2 findings, both read. The round-trip gap is the finding this round closes: the body no longer carries a closing keyword, states that the result path is unimplemented, and keeps #2887 open, so the descope is declared. The narrowing to the docked preview of the latest version is declared in the body, in ADR-092 and in the authoring skill, so it stays 🟢 and is carried in the deferred list.

10 findings, all read; one is reported above. Four arrived as warnings, and none has a reachable defect behind it. The largest claim is that an unexpected error can reach the refusal toast as if it were a refusal reason: the chat sender catches every failure inside its own `try`, classifies it and reports it in chat, and it has no `throw` before that block, so only the gate's five messages can reach the toast. The other three are mock-typing and test-robustness rules. The nine unreported findings are regraded 🟢 and carried in the deferred list.

### Verdict
APPROVE — the round's one blocking finding is fixed at its cause, and no new finding blocks at this HEAD.


---

PR #3468, opened 2026-08-26
fix(skills): a skill from a symlink-aliased source root installs as a dangling link and reaps itself

### LINKED ISSUE #3461: UI - Selecting a skill from a source turns itself off again a second later
## Context

Reported on a sandbox after adding the dam repository as a GitHub skill source. The source scans correctly and lists its skills. The problem is in the Skills section of the web UI, and it repeats on every attempt.

## Problem

A user cannot select skills from a source. Each toggle reverts on its own, so no skill stays installed and the sandbox never gets the skills.

## Expected Behavior

A selected skill stays selected. The switch stays on and the skill installs.

## Actual Behavior

The switch turns on, then turns itself off about a second later. No error message appears.

## Steps to Reproduce

1. Open a sandbox and go to Skills.
2. Add a GitHub skill source. Wait for the scan to list the skills.
3. Turn on any skill.
4. Watch the switch. It turns off again after about a second.

kapetr:
Fixes #3461.

A skill repo can alias one skill root from another with relative symlinks (this repo links `.claude/skills/<name>` into `.agents/skills/<name>`). The installer resolved the skill through the symlinked alias and published the link instead of the files; after the install's temp clone was cleaned up, the link dangled, the skill vanished from the pod's disk listing, and the server's cleanup deleted the install from the database — the toggle turned itself off.

- **agent-runtime**: resolve the skill's real directory before copying, reject anything that escapes the clone or is itself a symlink. Covered by a test.
- **ui**: a slow skills-state poll could overwrite a just-made toggle with stale data; stale poll responses are now dropped.

Verified on a local cluster: toggles materialize real directories and stay on.

Follow-up (separate issue): the server-side cleanup deletes installs on the pod's word alone and logs nothing — it should only remove skills it has seen on disk, and log removals.


packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts
```diff
@@ -569,6 +572,10 @@ function ignoreMissing(err: unknown): void {
 function leaveSidecar(): void {}
 
 async function assertNoSymlinks(root: string): Promise<void> {
+  const rootStat = await fs.lstat(root);
+  if (rootStat.isSymbolicLink()) {
+    throw new Error(`skill rejected: the skill directory itself is a symlink`);
```
dam-code-guardian, line 577: 🟢 **Suggestion:** This rejection throws a raw `Error`, while the module's ports return `Result`. `writeFromDir` has no `Result` in its signature, so the rule escapes as an exception and reaches the client as a generic 500 instead of a typed skill error.

`assertNoSymlinks` already threw for every inner symlink before this PR, so this matches the existing shape rather than introducing a new seam. Worth a separate issue that models the whole function's rejections as `SkillsDomainError` values and widens the port, rather than a change here.

packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts
```diff
@@ -135,6 +135,12 @@ export function useSkillsSurface(
   const [setsFailed, setSetsFailed] = useState(false);
   const [applyingSets, setApplyingSets] = useState(false);
 
+  const localWriteEpochRef = useRef(0);
+  const pollSeqRef = useRef(0);
+  const supersedeInFlightPolls = useCallback(() => {
```
dam-code-guardian, line 140: 🟢 **Suggestion:** The guard is correct — the epoch drops a poll that a local write superseded, and `pollSeqRef` keeps the newest of two overlapping polls. It is opt-in per write site, though: nine call sites carry `supersedeInFlightPolls()`, so freshness depends on each future mutation remembering it.

The stack doc prescribes TanStack Query for client data fetching, and this package already uses it in 165 places. A `useQuery` with `refetchInterval` plus per-mutation invalidation would supply the same staleness control without the manual refs.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `a13e1531`

## PR #3468: fix(skills): a skill from a symlink-aliased source root installs as a dangling link and reaps itself
**Author:** kapetr | **Branch:** fix/3461-skill-toggle-revert → main | **Changes:** +136 −15 (3 files)

### Summary
The install path now resolves a skill directory to its real path, rejects a candidate that escapes the clone, and refuses a skill directory that is itself a symlink — so a symlink-aliased source root no longer publishes a dangling link that gets the install reaped. The UI drops stale skills-state poll responses that overwrote a just-made toggle.

### Findings
- 🟡 **Warning:** The Skills architecture page still describes install resolution as source-root walking only, so it does not record the two new ways an install can fail (`docs/architecture/skills.md:163`). Reported by doc-drift; details in that section below.
  **Fix:** Extend the **Install** bullet to state that resolution lands on the real path, refuses a path escaping the clone, and refuses a symlinked skill directory — the same way the **Read Skill File** bullet already records its escape refusal — and bump `Last verified:`.
- 🟢 **Suggestion:** The new root-symlink rejection throws a raw `Error` where the module's ports return `Result` (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:577`).
- 🟢 **Suggestion:** The freshness guard is opt-in per write site (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:140`).
- ✅ **Looks good:** The containment predicate is the correct form. `real.startsWith(rootReal + path.sep)` rejects the prefix-sibling case a bare `startsWith(rootReal)` would admit — a clone at `/tmp/x/clone` does not accept a resolved path under `/tmp/x/clone-evil`. Confirmed against all five boundary cases, and the repo's own `.claude/skills/*` links match the layout the fix targets.

- 🟡 **Warning:** Install-time skill resolution gained a real-path rule the Skills architecture page does not describe: candidates resolve through `realpath`, a candidate outside the clone root is rejected, and a skill directory that is itself a symlink is refused (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:539`, `:576`). The page's **Install** bullet still describes resolution as source-root walking only, so it understates when an install now fails (`docs/architecture/skills.md:163`).
  **Fix:** In the same PR, extend the **Install** bullet to state that install resolves the skill directory to its real path, refuses a path that escapes the clone, and refuses a symlinked skill directory — the same way the sibling **Read Skill File** bullet already records its escape refusal — and bump that page's `Last verified:` header to the PR date.

### Issue Fit
Linked issue: #3461 "UI - Selecting a skill from a source turns itself off again a second later" (closing keyword).

- 🟢 **Suggestion:** The issue's amended ask includes a second variant a maintainer added in discussion — skills toggling off after about two minutes on an idle page, with the note "I am looking into it now" — and neither the PR body nor the diff names it. The dangling-symlink reap plausibly explains both timings, but `Fixes #3461` closes an issue with an active, self-assigned investigation.

✅ Delivers the issue's ask: the install path that materialized a dangling symlink now resolves the real directory, removing the reap that turned the toggle off, and the stale-poll guard stops a slow poll from overwriting a just-made toggle. No undeclared scope — the server-side cleanup fix is a declared follow-up.

- 🟢 **Suggestion:** The root-symlink rejection throws a raw `Error` from the infrastructure layer (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:577`). `writeFromDir` returns `Promise<{ contentHash }>` with no `Result`, so the rule escapes as an exception and reaches the client as a generic 500 rather than a typed skill error. The surrounding `assertNoSymlinks` already threw for every inner symlink before this PR, so this is the existing shape, not a regression — worth a separate issue that models the whole function's rejections as `SkillsDomainError` values and widens the port.
- 🟢 **Suggestion:** The stale-poll fix is a hand-rolled request-race guard over a manual `useEffect` + `setInterval(5000)` poll, and it needs a `supersedeInFlightPolls()` call at each of nine local write sites (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:140`). The stack doc prescribes TanStack Query for client data fetching, which supplies this staleness control, and the package already uses it in 165 places. Moving this poll to `useQuery` with `refetchInterval` plus per-mutation invalidation would retire the manual bookkeeping.
- 🟢 **Suggestion:** Pre-existing cross-layer import — `packages/agent-runtime/src/modules/skills/services/skills-service.ts:23` imports `subPathEscapes` from the infrastructure implementation, while services may import from the domain only. This PR touches the same containment concern, so moving the path-containment predicate into `domain/` is worth a separate issue.

### Verdict
COMMENT — the fix is correct and the test covers resolution, materialization, and the escape rejection; the architecture page needs the new install failure modes recorded.

---

packages/agent-runtime/src/__tests__/unit/skill-symlinked-source.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    const realDir = path.join(clone, ".agents", "skills", "grill-me");
+    await fs.mkdir(realDir, { recursive: true });
+    await fs.writeFile(
+      path.join(realDir, "SKILL.md"),
+      skillMd("grill-me"),
+      "utf8",
+    );
+    await fs.mkdir(path.join(clone, ".claude", "skills"), { recursive: true });
+    await fs.symlink(
+      "../../.agents/skills/grill-me",
```
jjeliga: Why the full path and not `realDir`?

packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts
```diff
@@ -285,11 +300,14 @@ export function useSkillsSurface(
           }),
         `Failed to update ${skill.name}`,
       );
-      if (result !== ACTION_FAILED) setInstalled(result);
+      if (result !== ACTION_FAILED) {
```
jjeliga: 3x duplicated

packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts
```diff
@@ -536,7 +536,10 @@ async function resolveSkillDirInClone(
...
     try {
       await fs.access(path.join(candidate, "SKILL.md"));
-      return ok(candidate);
+      const real = await fs.realpath(candidate);
+      const rootReal = await fs.realpath(repoDir);
+      if (!real.startsWith(rootReal + path.sep)) continue;
+      return ok(real);
     } catch {}
   }
   return err({ kind: "SkillNotFoundInSource", source: repoDir, name });
```
dam-code-guardian, line 545: 🟢 **Suggestion:** A candidate that has a `SKILL.md` but resolves outside the clone is skipped, so the loop ends at this same `SkillNotFoundInSource`. A security refusal and a genuinely absent skill therefore report the same error, and the install surfaces "not found in source" for both.

A distinct error kind for the refusal would make it diagnosable from the install failure alone.

packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts
```diff
@@ -191,9 +204,15 @@ export function useSkillsSurface(
         }
         return;
       }
+      const epoch = localWriteEpochRef.current;
+      const seq = ++pollSeqRef.current;
       try {
         const state = await api.skills.state.query({ agentId });
-        if (!cancelled) {
+        const fresh =
```
dam-code-guardian, line 211: 🟢 **Suggestion:** The gate is evaluated once for the whole response, and all four setters sit behind it. A local `installed` write therefore also discards this response's fresh `standalone`, `instancePublishes` and `standaloneSnapshot`.

It self-corrects on the next 5-second tick. Gating per state slice would keep the unrelated server data on the first pass.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `4fcc1f14`

## PR #3468: fix(skills): a skill from a symlink-aliased source root installs as a dangling link and reaps itself
**Author:** kapetr | **Branch:** fix/3461-skill-toggle-revert → main | **Changes:** +137 −21 (4 files)

### Summary
The install path resolves a skill directory to its real path, skips a candidate that escapes the clone, and refuses a skill directory that is itself a symlink; the UI drops stale skills-state poll responses. This round answers both review threads: the test now aliases the skill with `path.relative`, and one `commitInstalled` helper replaces the three duplicated supersede-then-set pairs.

### Changes since last review
Previous HEAD: `a13e1531` (2026-08-26T13:05:28Z) — verdict COMMENT

- ✅ **Fixed:** the **Install** bullet records realpath resolution and both refusals, and `Last verified:` is bumped (`docs/architecture/skills.md:163`)
- 🔁 **Still present:** the root-symlink rejection throws a raw `Error` where the module's ports return `Result` (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:577`)
- 🔁 **Still present:** the freshness guard stays opt-in per write site — `commitInstalled` centralizes the four `setInstalled` sites, `setStandalone` and `setPublishes` still call `supersedeInFlightPolls()` by hand (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:140`)
- 🔁 **Still present:** pre-existing cross-layer `subPathEscapes` import (`packages/agent-runtime/src/modules/skills/services/skills-service.ts:23`)
- 🆕 **New:** one documentation warning and two suggestions, below.

### Findings
- 🟡 **Warning:** This PR changes how the panel's read and its writes order against each other: a `skills.state` response is discarded when a local write landed after the poll was issued. The **Reconciled state** section still presents `skills.state` as the single read with no ordering caveat (`docs/architecture/skills.md:267`). Reported by doc-drift.
  **Fix:** When a change alters how a panel's read and its writes order against each other, state the resulting coupling on the subsystem page in the same PR — one sentence in **Reconciled state** (a write's returned list wins over any poll issued before it, whose response is dropped), not the epoch or sequence mechanism — and bump `Last verified:`. The page holds 39,979 of 40,000 characters, so cut volatile detail to make room rather than append.
- 🟢 **Suggestion:** A candidate refused for escaping the clone returns the same `SkillNotFoundInSource` error as an absent skill (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:545`).
- 🟢 **Suggestion:** The freshness gate is evaluated once for the whole poll response, so an `installed` write also discards that response's fresh `standalone`, `instancePublishes`, and `standaloneSnapshot` values (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:211`).

- 🟡 **Warning:** The UI read/write ordering changed — a `skills.state` poll response is now discarded when a local write (install, uninstall, update, toggle, apply-set, standalone create or delete, publish) landed after it was issued, so a write's returned list is authoritative against a racing poll. `docs/architecture/skills.md` never states that rule: the **Reconciled state** section still presents `skills.state` as the single read with no ordering caveat, and "renders from an authoritative result" is asserted without the guarantee that makes it hold (`docs/architecture/skills.md:267`).
  **Fix:** As in `### Findings` above — one semantic sentence in **Reconciled state**, `Last verified:` bumped, room made by cutting volatile detail.

### Issue Fit
Linked issue: #3461 "UI - Selecting a skill from a source turns itself off again a second later" (closing keyword `Fixes #3461`). The diff addresses both reported revert paths: the stale-poll overwrite, and the dangling-symlink install that the server's cleanup reaps.

✅ Delivers the issue's ask. The server-side cleanup fix is a declared follow-up, not undeclared scope.

The skill also flagged the compressed **Skill Publish Record** and **Publish** paragraphs as undeclared scope. Not reported: the page holds 39,979 of 40,000 characters, so the required **Install** paragraph could only be added by cutting elsewhere, and every dropped detail is still recorded — the `agent_skill_publishes` table name in eight other places on the same page, the publish author and branch pattern in `docs/adrs/030-skills-marketplace.md:65`.

- 🔁 3 findings from the previous review still present (see review at `a13e1531`) — the raw `Error` rejection, the opt-in freshness guard, and the cross-layer `subPathEscapes` import. Each is listed in `### Changes since last review`.
- 🟢 **Suggestion:** A candidate that has a `SKILL.md` but is refused for escaping the clone falls through to the same `SkillNotFoundInSource` error as a genuinely absent skill, so the install reports "not found in source" for a security refusal. A distinct error kind would make the refusal diagnosable (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:545`).
- 🟢 **Suggestion:** The epoch and sequence guard is evaluated once for the whole poll response, so a local `installed` write also discards that response's fresh `standalone`, `instancePublishes`, and `standaloneSnapshot` values. It self-corrects on the next 5-second tick; gating per state slice would keep the unrelated server data (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:211`).
- The skill's third suggestion — move the new clone-containment invariant into `domain/` — is folded into the still-present `subPathEscapes` carryover, which covers the same placement concern.

_Skipped: no matching files (no `.tsx` or `.jsx` in this diff)._

### Verdict
COMMENT — the fix and its test are correct and both review threads are answered; the architecture page does not yet record the read/write ordering rule this PR introduces.

---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `9811384d`

## PR #3468: fix(skills): a skill from a symlink-aliased source root installs as a dangling link and reaps itself
**Author:** kapetr | **Branch:** fix/3461-skill-toggle-revert → main | **Changes:** +140 −24 (4 files)

### Summary
The install path resolves a skill to its real directory, skips a candidate that escapes the clone, and refuses a skill directory that is itself a symlink; the UI drops stale skills-state poll responses. This round adds the documentation sentence the previous round asked for.

### Changes since last review
Previous HEAD: `4fcc1f14` (2026-09-01T09:19:56Z) — verdict COMMENT

- ✅ **Fixed:** the **Reconciled state** section states that a write's returned list is authoritative against a racing read, and `Last verified:` is bumped (`docs/architecture/skills.md:267`)
- 🔁 **Still present:** the root-symlink rejection throws a raw `Error` where the module's ports return `Result` (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:577`)
- 🔁 **Still present:** a candidate refused for escaping the clone reports `SkillNotFoundInSource` (`packages/agent-runtime/src/modules/skills/infrastructure/local-skill-repository.ts:545`)
- 🔁 **Still present:** the freshness guard is opt-in per write site (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:140`)
- 🔁 **Still present:** the freshness gate discards the response's unrelated state slices (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:211`)
- 🔁 **Still present:** pre-existing cross-layer `subPathEscapes` import (`packages/agent-runtime/src/modules/skills/services/skills-service.ts:23`)

### Findings
No blocking finding survives at this HEAD. The five current findings are the unchanged 🟢 carryovers in the block above; no TypeScript file changed since the previous round.

The new sentence matches the code: the epoch is captured before the query, so a write that lands during the round trip discards the response. It states the resulting rule rather than the epoch and sequence mechanism. The page now measures 39,837 of the 40,000-character cap and `scripts/doc-size.mjs` passes, so the two sentences that left were the cost of fitting this one.

✅ No findings.

### Issue Fit
Linked issue: #3461 "UI - Selecting a skill from a source turns itself off again a second later" (closing keyword `Fixes #3461`).

✅ No findings. Both reported revert paths are covered — the ~1-second self-revert by dropping a stale `skills.state` poll response, and the mass toggle-off at its root cause, the symlink-aliased source root installing as a dangling link. The server-side cleanup hardening is a declared follow-up, not undeclared scope.

- 🔁 5 findings from the previous review still present (see review at `4fcc1f14`) — the raw `Error` rejection, the `SkillNotFoundInSource` refusal, the opt-in freshness guard, the whole-response freshness gate, and the cross-layer `subPathEscapes` import. Each is listed in `### Changes since last review`; all remain 🟢.

_Skipped: no matching files (no `.tsx` or `.jsx` in this diff)._

### Verdict
APPROVE — the documentation gap that held the previous round is closed, and every remaining item is a suggestion.

---

PR #3469, opened 2026-08-26
perf(ci,agents): cut the agent image to ~307MiB and stop re-pushing unchanged layers

### LINKED ISSUE #3403: Every release makes every node re-download the whole agent image
## Context

#3246 measured a 30s image pull on the create path. Comparing consecutive releases in the registry showed they share only ~12MB of the ~640MB agent image.

## Problem

The image build invalidates nearly all layers on every release, so the first agent start on each node after a deploy pays a ~30s pull, and local installs re-download the whole image each upgrade. Lower priority: the cost is once per node per release, then cached.

## Goal

Consecutive releases share their base layers, so a release pulls only a small delta.

## Done when

Two consecutive releases share the large majority of their layers, and the first post-deploy start on a warm node pays a few seconds of pull instead of ~30.

jezekra1:
## What

Toward #3403 (epic #3242) — not a closing keyword deliberately, see below. Three independent cuts to what an agent start has to download.

| | uncompressed | compressed |
|---|---|---|
| claude-code image today | 2.62 GB | 655.6 MiB |
| duplicate `chown` layer removed | 2.12 GB | 551.1 MiB |
| npm cache off the layer | 1.73 GB | 384.5 MiB |
| one CLI instead of two | **1.39 GB** | **~307 MiB** |

Plus the CI change that stops re-pushing layers whose content never moved.

## Why

**Unchanged layers were being re-pushed every release.** Pulled the manifests for
ten consecutive `claude-code` builds from quay. Consecutive builds shared 204.6 of
655.6 MiB — every release shipped **451 MiB of new blobs**, occasionally 643 MiB,
and one pair shared 100%. Nothing in the image content explains that spread; it is
cache behaviour. `type=gha` caps a repo's Actions cache at 10GB and `mode=max`
across ten components and two arches is multiples of that, so entries were evicted
between runs. A re-executed step's output carries fresh file timestamps, so
unchanged content lands on a new layer digest. Moved to a registry cache at
`<component>:buildcache-<arch>`; PRs read it and never write it, which preserves
the branch isolation `type=gha` gave for free.

**104.5 MiB of the image was a copy of the image.** platform-base's final
`chown -R` walked `/usr/local/share/mise` and `/app`, re-tarring both into a second
copy of themselves. Ownership rides the `COPY`s now. As a bonus the recursive chown
had been forcing a copy-up that broke pnpm's hardlinks in `node_modules`; they
survive the build again, which is most of the 500 MB uncompressed saving.

**199 MB of npm cache shipped inside the harness layer.** mise runs npm with its
own HOME, so npm cached to `/tmp/.npm` while `--mount=type=cache` pointed at
`/root/.npm` — the mount cached nothing between builds *and* the cache ended up in
the layer. 166.6 MiB of that compressed, measured by gzipping the tree out of the
old image.

**The image carried two Claude Code binaries at different versions.** The adapter's
`claude-agent-sdk` bundles the native CLI the chat path executes and cannot be
unbundled, so the separately pinned `npm:@anthropic-ai/claude-code` install was a
second 247 MB copy of the same program — and the two were **not the same build**.
The pin said 2.1.210; the SDK bundles 2.1.220. `acp-agent.js` resolves the CLI as
`CLAUDE_CODE_EXECUTABLE` if set and otherwise falls back to the SDK's own copy, and
nothing set that variable — so a terminal session ran 2.1.210 while the chat ran
2.1.220 against the same `~/.claude/projects` transcripts. `claude` now points at
the adapter's build and `CLAUDE_CODE_EXECUTABLE` names it, leaving the adapter
nothing to guess.

## Verification

- platform-base built both ways from the same cached stages: ownership, mode and
  size **identical across all ~25k paths**; the only diffs are directory mtimes
  from the two build times.
- `docker build --check` clean; `mise run check` green on each commit.
- In the built agent image: `claude --version` reports 2.1.220 through the symlink,
  and `claude-agent-acp` answers `initialize` and `session/new` over stdio, so the
  adapter spawns the binary it was given.
- Separately confirmed that a re-executed `COPY` always yields a new layer digest,
  and that `SOURCE_DATE_EPOCH` + `rewrite-timestamp=true` makes two cold builds
  byte-identical. Left out of this PR — see below.

## Consequences worth knowing

- **Terminal sessions move 2.1.210 → 2.1.220.** That is the point, but it is a
  version bump reaching users.
- **Bumping the CLI now means bumping the `claude-agent-acp` pin**, since the CLI
  rides it. `harness-tools.toml` no longer mentions claude-code.
- **Retiring the pin needed the heal taught about it.** My first version assumed a
  retired key was harmless because every agent that booted since the heal shipped
  had its *value* corrected. Review showed that reasoning does not cover
  *retirement*: the heal derives its strip list from the keys the image ships, so
  dropping the key stops it naming it, and a workspace copy seeded before
  retirement then survives every upgrade. It is not inert — a later `mise install`
  honors that pin, installs the package without its postinstall, and the shim it
  creates shadows the real binary, so `claude` fails outright. The entrypoint now
  names retired keys with the date they were retired.
- Quay will accumulate a `buildcache-<arch>` tag per component repo. They are OCI
  cache manifests, not images, so they may look odd in tag lists or scan reports.
- **The first CI run after merge has nothing to read** — expect one full-miss build.

## Not in this PR

- `SOURCE_DATE_EPOCH` + `rewrite-timestamp`, now filed as #3495. The registry cache
  fixes "nothing changed at all"; any commit touching `agent-runtime` legitimately
  changes platform-base, which re-executes the harness install and re-pushes it
  unchanged. That is why this PR does not claim to close #3403: its done-when asks
  that two consecutive releases share the large majority of their layers, which
  holds on a cache hit and not after a real base change. Close it once a deploy
  confirms the sharing.
- platform-base leftovers: 49 MB of `/usr/share/locale`, and `typescript` (24 MB)
  sitting in a `--prod` deploy.

## Test plan

- [ ] CI image build, both arches
- [ ] `mise run e2e:loop` — a real agent boot exercising the entrypoint, workspace
      seeding and a live chat turn on the new image
- [ ] after dev deploy: confirm two consecutive releases share the large majority
      of their layers, per #3403's done-when

### DIFF
--- a/packages/agents/claude-code/harness-tools.toml
+++ b/packages/agents/claude-code/harness-tools.toml
@@ -4,5 +4,4 @@
 # Format contract: [tools] entries only, one `"key" = "version"` per line —
 # agent-entrypoint parses this file to strip legacy workspace-seeded pins.
 [tools]
-"npm:@anthropic-ai/claude-code" = "2.1.210"
 "npm:@agentclientprotocol/claude-agent-acp" = "0.66.0"
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -165,8 +165,10 @@ jobs:
           file: ${{ matrix.dockerfile }}
           platforms: linux/${{ matrix.arch }}
           outputs: type=image,name=${{ env.IMAGE_PREFIX }}/${{ matrix.component }},push-by-digest=true,name-canonical=true,push=true
-          cache-from: type=registry,ref=${{ env.IMAGE_PREFIX }}/${{ matrix.component }}:buildcache-${{ matrix.arch }}
-          cache-to: ${{ github.event_name != 'pull_request' && format('type=registry,ref={0}/{1}:buildcache-{2},mode=max,image-manifest=true', env.IMAGE_PREFIX, matrix.component, matrix.arch) || '' }}
+          cache-from: |
+            type=registry,ref=${{ env.IMAGE_PREFIX }}/${{ matrix.component }}:buildcache-${{ matrix.arch }}
+            type=gha,scope=${{ matrix.component }}-${{ matrix.arch }}
+          cache-to: ${{ github.event_name != 'pull_request' && format('type=registry,ref={0}/{1}:buildcache-{2},mode=max,image-manifest=true', env.IMAGE_PREFIX, matrix.component, matrix.arch) || format('type=gha,mode=max,scope={0}-{1}', matrix.component, matrix.arch) }}
       - name: Export digest
         env:
           DIGEST: ${{ steps.build.outputs.digest }}
@@ -252,8 +254,10 @@ jobs:
           file: packages/platform-base/Dockerfile
           platforms: linux/${{ matrix.arch }}
           outputs: type=image,name=${{ env.IMAGE_PREFIX }}/platform-base,push-by-digest=true,name-canonical=true,push=true
-          cache-from: type=registry,ref=${{ env.IMAGE_PREFIX }}/platform-base:buildcache-${{ matrix.arch }}
-          cache-to: ${{ github.event_name != 'pull_request' && format('type=registry,ref={0}/platform-base:buildcache-{1},mode=max,image-manifest=true', env.IMAGE_PREFIX, matrix.arch) || '' }}
+          cache-from: |
+            type=registry,ref=${{ env.IMAGE_PREFIX }}/platform-base:buildcache-${{ matrix.arch }}
+            type=gha,scope=platform-base-${{ matrix.arch }}
+          cache-to: ${{ github.event_name != 'pull_request' && format('type=registry,ref={0}/platform-base:buildcache-{1},mode=max,image-manifest=true', env.IMAGE_PREFIX, matrix.arch) || format('type=gha,mode=max,scope=platform-base-{0}', matrix.arch) }}
       - name: Export digest
         env:
           DIGEST: ${{ steps.build.outputs.digest }}
@@ -340,8 +344,10 @@ jobs:
           outputs: type=image,name=${{ env.IMAGE_PREFIX }}/${{ matrix.component }},push-by-digest=true,name-canonical=true,push=true
           build-args: |
             BASE_IMAGE=${{ env.IMAGE_PREFIX }}/platform-base:${{ needs.changes.outputs.platform_base_tag }}
-          cache-from: type=registry,ref=${{ env.IMAGE_PREFIX }}/${{ matrix.component }}:buildcache-${{ matrix.arch }}
-          cache-to: ${{ github.event_name != 'pull_request' && format('type=registry,ref={0}/{1}:buildcache-{2},mode=max,image-manifest=true', env.IMAGE_PREFIX, matrix.component, matrix.arch) || '' }}
+          cache-from: |
+            type=registry,ref=${{ env.IMAGE_PREFIX }}/${{ matrix.component }}:buildcache-${{ matrix.arch }}
+            type=gha,scope=${{ matrix.component }}-${{ matrix.arch }}
+          cache-to: ${{ github.event_name != 'pull_request' && format('type=registry,ref={0}/{1}:buildcache-{2},mode=max,image-manifest=true', env.IMAGE_PREFIX, matrix.component, matrix.arch) || format('type=gha,mode=max,scope={0}-{1}', matrix.component, matrix.arch) }}
       - name: Export digest
         env:
           DIGEST: ${{ steps.build.outputs.digest }}
@@ -435,8 +441,10 @@ jobs:
           outputs: type=image,name=${{ env.IMAGE_PREFIX }}/${{ matrix.component }},push-by-digest=true,name-canonical=true,push=true
           build-args: |
             BASE_IMAGE=${{ env.IMAGE_PREFIX }}/claude-code:${{ needs.changes.outputs.claude_code_base_tag }}
-          cache-from: type=registry,ref=${{ env.IMAGE_PREFIX }}/${{ matrix.component }}:buildcache-${{ matrix.arch }}
-          cache-to: ${{ github.event_name != 'pull_request' && format('type=registry,ref={0}/{1}:buildcache-{2},mode=max,image-manifest=true', env.IMAGE_PREFIX, matrix.component, matrix.arch) || '' }}
+          cache-from: |
+            type=registry,ref=${{ env.IMAGE_PREFIX }}/${{ matrix.component }}:buildcache-${{ matrix.arch }}
+            type=gha,scope=${{ matrix.component }}-${{ matrix.arch }}
+          cache-to: ${{ github.event_name != 'pull_request' && format('type=registry,ref={0}/{1}:buildcache-{2},mode=max,image-manifest=true', env.IMAGE_PREFIX, matrix.component, matrix.arch) || format('type=gha,mode=max,scope={0}-{1}', matrix.component, matrix.arch) }}
       - name: Export digest
         env:
           DIGEST: ${{ steps.build.outputs.digest }}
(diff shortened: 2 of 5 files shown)


packages/agents/claude-code/harness-tools.toml
```diff
@@ -4,5 +4,4 @@
 # Format contract: [tools] entries only, one `"key" = "version"` per line —
 # agent-entrypoint parses this file to strip legacy workspace-seeded pins.
 [tools]
-"npm:@anthropic-ai/claude-code" = "2.1.210"
```
dam-code-guardian, line 7: 🟡 **Warning:** Removing this key retires the pin, but the key is also what `agent-entrypoint` reads to strip a legacy workspace-seeded copy. `entrypoint.sh:92-108` derives the strip list from the `[tools]` keys of `/etc/mise/conf.d/*.toml`, so once the key is gone the heal no longer names it, and a `~/.config/mise/config.toml` that still pins `npm:@anthropic-ai/claude-code` survives the upgrade.

That persisted pin then makes mise generate a `claude` shim, and the shim wins over the new `/usr/local/bin/claude` symlink, because `MISE_DATA_DIR/shims` precedes `/usr/local/bin` on `PATH` (`platform-base/Dockerfile:125`). Replayed with a home-only pin against a `conf.d` that carries the adapter alone: the shim was created and `claude --version` reported the pinned `2.1.210`. With the pinned version absent from the image, the same call fails in mise (`Failed to install npm:@anthropic-ai/claude-code@…`) instead of reaching the SDK binary — the terminal path of `harness-terminal.sh:6`.

The PR body reasons that every agent booted since the heal shipped already had the pin removed. The heal landed 2026-08-18 (#3305), and it only ever removed keys the image still listed, so the reasoning holds for the pin's *value* going stale — not for its *retirement*.

**Fix:** When a pin is retired rather than bumped, the strip list needs the retired key kept for a release. A commented-out entry does not work here: the parser reads `[tools]` keys only. The `accessMode` precedent's two-part shape applies — retire the install, keep the name where the heal can still see it.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `a37934f0`

## PR #3469: perf(ci,agents): cut the agent image to ~307MiB and stop re-pushing unchanged layers
**Author:** jezekra1 | **Branch:** fix/agent-image-layer-caching → main | **Changes:** +32 −29 (4 files)

### Summary
Three image-size cuts plus a buildx cache move: registry buildcache replaces `type=gha`, platform-base's duplicate recursive `chown -R` becomes `COPY --chown`, the npm cache stays out of the harness layer, and the duplicate `npm:@anthropic-ai/claude-code` 2.1.210 pin gives way to the `claude-agent-sdk`-bundled CLI at 2.1.220.

### Findings
- 🟡 **Warning:** retired pin is no longer stripped from a persisted workspace config (`packages/agents/claude-code/harness-tools.toml:7`)
- 🟡 **Warning:** #3403's closing keyword against a partly deferred done-when — see **Issue Fit** below.
  **Fix:** A closing keyword fits an issue whose done-when the diff fully meets. A deferred remainder stays tracked — in #3403 itself, or in a follow-up issue named in the body.

✅ No findings.

### Issue Fit
Linked issue: #3403 "Every release makes every node re-download the whole agent image" (closing keyword). Epic #3242 is a plain reference, not a closing link.

- 🟡 **Warning:** The closing keyword closes #3403, but part of its done-when stays open. Any commit that touches `agent-runtime` changes platform-base. The harness install then re-executes, and its unchanged layer is pushed again (PR body → "Not in this PR"; `.github/workflows/cd.yml:168`). For that release class, consecutive releases still do not share the large majority of layers.
  **Fix:** A closing keyword fits an issue whose done-when the diff fully meets. A deferred remainder stays tracked — in #3403 itself, or in a follow-up issue named in the body.

Scope note (not a finding): the three size cuts go beyond #3403's layer-sharing ask. The PR body declares all of them, and they serve epic #3242.

### Verdict
COMMENT — the cache move, the `chown` removal and the npm-cache fix verify clean; two 🟡 remain, one on the retired pin's upgrade path and one on the issue's closing scope.

---

packages/agents/claude-code/harness-tools.toml
jezekra1: You're right, and my reasoning in the PR body was wrong. Fixed in 86d23890b.

I tried to reproduce your mechanism and initially couldn't: with a home-only pin, `mise reshim` created no shim and `claude` still resolved to the symlink. What I'd missed is that no shim can exist until the pinned package is actually installed. The trigger is a bare `mise install` — which anything in the pod can run, and which the base image's own comment already warns about. That honors the stale pin, installs the package *without* the postinstall that materializes the native binary, and the shim it creates shadows `/usr/local/sbin/claude`. On the built image `claude --version` then fails with `Error: claude native binary not installed.` So the outcome is exactly what you described, via a different path — and it's worse than a wrong version, it's a broken terminal harness.

On the fix: I couldn't find a two-part retirement in the `accessMode` code you pointed at, so I went with naming retired keys in the entrypoint rather than in `harness-tools.toml`. Keeping the key in `[tools]` would re-pin a tool the image no longer installs, which arms the same landmine for *every* agent instead of the legacy ones. The retired list sits next to the heal it feeds, with a note to drop entries once no agent can predate the release.

Verified on the rebuilt image: the heal strips the retired pin and leaves a user-added tool untouched, and the same heal on the previous image leaves it in place.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `86d23890`

## PR #3469: perf(ci,agents): cut the agent image to ~307MiB and stop re-pushing unchanged layers
**Author:** jezekra1 | **Branch:** fix/agent-image-layer-caching → main | **Changes:** +61 −38 (5 files)

### Summary
Round 2 fixes the retired-pin hazard at its cause: `entrypoint.sh` now carries a `retired_harness_tools` list prepended to the strip loop, so a workspace config that still pins the retired `npm:@anthropic-ai/claude-code` is healed even though the image no longer lists it. Verified by replaying the loop against a fake `$HOME` — the retired key is removed from both `config.toml` and `mise.lock`, including the last-entry case, and a user-added tool is untouched.

### Changes since last review
Previous HEAD: `a37934f0` (2026-08-26T12:50:31Z) — verdict COMMENT

- ✅ **Fixed:** retired pin is stripped again, via a retirement list in the entrypoint rather than a kept `[tools]` key (`packages/platform-base/entrypoint.sh:93`)
- 🔁 **Still present:** `Closes #3403` against a partly-deferred done-when, with no follow-up issue filed (`.github/workflows/cd.yml:165`)

### Findings
- 🟡 **Warning:** the body describes the superseded design — it still says `agent-entrypoint` "no longer strips a legacy workspace-seeded claude-code pin", which this HEAD reverses, and it keeps `Closes #3403` while the "Not in this PR" remainder has no tracking issue (`packages/platform-base/entrypoint.sh:93`; body lines 3 and 69; reported by issue-fit).
  **Fix:** update the body in the commit that changes behavior, and drop the closing keyword when part of the done-when is deferred — reference the issue plainly and file the follow-up.
- 🟢 **Suggestion:** the retirement list has no expiry mechanism beyond its comment (`packages/platform-base/entrypoint.sh:99`) — a dated note per entry would make "no agent can predate this" checkable rather than remembered.

✅ No findings.

### Issue Fit
- 🟡 `.github/workflows/cd.yml:165` — carryover: the closing keyword closes #3403 while its delta-sharing done-when is only partly delivered, and no `SOURCE_DATE_EPOCH` / rewrite-timestamp follow-up issue exists (84 open issues scanned).
- 🟡 `packages/platform-base/entrypoint.sh:93` — the body's consequence bullet contradicts this HEAD; merged into `### Findings` above.

### Verdict
COMMENT — the retired-pin fix verifies clean; the body still describes the design it replaced and closes an issue whose remainder is untracked.

---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `95d86ff4`

## PR #3469: perf(ci,agents): cut the agent image to ~307MiB and stop re-pushing unchanged layers
**Author:** jezekra1 | **Branch:** fix/agent-image-layer-caching → main | **Changes:** +61 −38 (5 files)

### Summary
Round 3 closes the last two open items. The retirement list now dates each entry, and the body drops its closing keyword with the deferred reproducible-build remainder filed as #3495.

### Changes since last review
Previous HEAD: `86d23890` (2026-08-27T10:27:50Z) — verdict COMMENT

- ✅ **Fixed:** the body no longer claims to close #3403 — it opens "Toward #3403 (epic #3242) — not a closing keyword deliberately", and the deferred `SOURCE_DATE_EPOCH` + `rewrite-timestamp` remainder is tracked as open issue #3495
- ✅ **Fixed:** each retirement-list entry carries its retirement date, so the drop horizon is checkable rather than remembered (`packages/platform-base/entrypoint.sh:99`)

### Findings

The dated comment was verified not to disturb the heal, by replaying the shipped strip loop against a fake `$HOME`: the shell removes `# retired 2026-08-27` before the assignment, so the loop still iterates exactly `npm:@anthropic-ai/claude-code`, the conf.d key and `node`. A legacy `2.1.210` pin and its `mise.lock` block are both removed, and a user-added `npm:cowsay` entry is untouched.

One note for whoever adds the second entry, not a finding on this diff: the comment reads "Each entry carries the date it was retired", and `$retired_harness_tools` is word-split, so a per-entry date written inline would expand `#`, `retired` and the date into three extra strip targets. They are harmless — no config key matches them, and the replay confirmed the real key is still stripped — but the single trailing comment this diff ships is the form that works.

✅ No findings.

### Issue Fit
✅ No findings.

### Verdict
APPROVE — both blocking findings are closed at their cause and nothing new blocks.

---

Issue #3404, opened 2026-08-20
A webhook adds ~10s to the agent Service create on dev (needs cluster-level read)


jezekra1:
## Context

#3246 measured 19s from create to the pod object existing; 10s of it is a single StatefulSet create call, with the platform waiting on the cluster. Traces show it server-side, likely in admission.

## Research Goal

Attribute the 10s to a specific stage of the dev cluster's write path so we know whether create latency is ours to fix or the cluster's.

## Research Questions

- Which admission stage (webhooks, image-scan checks, other) accounts for the time?
- Does it affect other workload writes on the cluster the same way?
- Is it dev-specific, or will other installs see it?

## Users / Audience

Anyone creating an agent on the dev install.

## Sources of Insight

- Platform traces in the telemetry store (the slow call is already captured)
- Cluster admission configuration, with the cluster owners

## Decision it informs

Whether the create-path work targets our controller or a cluster configuration change, and what exactly to ask the cluster owners to change.

## Done when

The 10s is attributed with evidence and a recommendation is posted on this issue.

jezekra1:
## Attributed. It is the Service create, not the StatefulSet, and the mechanism is a webhook timeout

Measured on dev with eight agents created 25s apart (build `0.2.16-dev-2670b65c46e4`), plus timestamp archaeology across all 73 agents in `dam-dev-sandboxed`. The probes were deleted afterwards.

### The 10s is a webhook timing out on the agent Service write

StatefulSet→Service creation gap, excluding 42 agents whose StatefulSets were recreated later:

| gap | agent Service | gateway Service |
|---|---|---|
| 0–3s | 22 | 72 |
| **10–11s** | **9** | **1** |

Not a tail — a spike sitting on exactly 10 seconds, which is Kubernetes' default admission webhook `timeoutSeconds`. The object is still written, so the webhook's `failurePolicy` is `Ignore`: the request waits out the timeout, the webhook is skipped, the write lands. It hits agent Services on 29% of creates and gateway Services on 1.4%, so something distinguishes the two objects to whatever matches them.

Why this looked like the StatefulSet: the `agentStatefulSet` timing mark spanned **both** `applyStatefulSet` and `applyService`, so #3246's span attributed the whole window to the StatefulSet call. One probe's step took 10.05s while its StatefulSet was persisted ~1s in. #3470 splits the mark.

[...]

PR #3470, opened 2026-08-26
fix(controller): stop the idle sweep re-hibernating sleeping agents

jezekra1:
## What

Two changes to the agent create path, both from measuring #3404 (epic #3242).

- The idle sweep skips agents whose status already reports them hibernated.
- The `agentStatefulSet` timing mark splits, so `applyService` is attributed
  separately from `applyStatefulSet`.

## Why

**The sweep was generating the queue that delays real creates.** It walked the
whole fleet and hibernated every idle agent without asking whether it was
already asleep. On dev that is 68 agents, nearly all hibernated, twice in five
minutes. Each one costs a 3s-timeout HTTP probe to a pod that isn't running,
two scale-to-zero writes, and an Agent status write. The status write bumps
resourceVersion, so each enqueues a reconcile — 68 of those at ~353ms on the
controller's single worker is roughly 24s of queue.

Measured: eight agents created 25s apart on dev, seven reached their
StatefulSet in 0–2s, one took **23s**. That one's `slow reconcile` shows it did
not start until 23s after its CR existed, and a sweep had begun five seconds
before it was created.

Reading the status condition is safe here: it is written only after a
successful scale-to-zero, so it cannot claim an agent is asleep that isn't.

**The timing mark was hiding which call was slow.** One mark covered both
`applyStatefulSet` and `applyService`, which is why #3246 attributed a 10s wait
to "the StatefulSet create call". The probes say otherwise: one reconcile's step
took 10.05s while its StatefulSet was persisted ~1s in, so the 10s was the
Service write. Across the fleet, StatefulSet→Service gaps sit at 0–3s except for
a spike of nine landing on exactly 10–11s — Kubernetes' default webhook
`timeoutSeconds`, with the create still succeeding, so `failurePolicy: Ignore`.
That part is the cluster's to fix and stays on #3404; this change just makes the
next occurrence self-diagnosing instead of needing the same excavation.

## What the guard actually checks

Two signals, not one. An agent is skipped only when its StatefulSet pair is **observed at zero** *and* hibernation is **already published**. Either alone is wrong in a different direction: replicas alone skips an over-budget agent whose status has not caught up, and status alone skips a running pod whose reconcile errored before publishing readiness.

This adds one fleet-wide StatefulSet listing per sweep — a single read where the old sweep did a probe and three writes per agent, and it writes nothing for a sleeping agent, so the queue burst stays gone.

## Test plan

- [x] `mise run controller:test`
- [x] the new test fails without the guard (verified by reverting it) and passes with it
- [ ] after dev deploy: re-run the probe (8 creates, 25s apart) and confirm no
      20s+ outlier, and that the sweep stops logging `hibernating idle agent`
      for agents already asleep

### DIFF
--- a/packages/controller/pkg/reconciler/idlechecker.go
+++ b/packages/controller/pkg/reconciler/idlechecker.go
@@ -89,7 +89,7 @@ func (c *IdleChecker) check(ctx context.Context) {
 			continue
 		}
 
-		if atZero[name] {
+		if atZero[name] && hibernationPublished(agent) {
 			continue
 		}
 
@@ -107,6 +107,23 @@ func (c *IdleChecker) check(ctx context.Context) {
 	}
 	slog.Debug("idle checker sweep complete",
 		"scanned", len(agents.Items), "hibernated", hibernated, "duration", time.Since(start))
+}
+
+func hibernationPublished(agent *unstructured.Unstructured) bool {
+	conds, found, err := unstructured.NestedSlice(agent.Object, "status", "conditions")
+	if err != nil || !found {
+		return false
+	}
+	for _, raw := range conds {
+		cond, ok := raw.(map[string]any)
+		if !ok {
+			continue
+		}
+		if cond["type"] == apiv1.ConditionReady && cond["reason"] == apiv1.ReasonHibernated {
+			return true
+		}
+	}
+	return false
 }
 
 func (c *IdleChecker) agentsAtZero(ctx context.Context) map[string]bool {
--- a/packages/controller/pkg/reconciler/idlechecker.go
+++ b/packages/controller/pkg/reconciler/idlechecker.go
@@ -88,6 +88,10 @@ func (c *IdleChecker) check(ctx context.Context) {
 			continue
 		}
 
+		if isHibernated(agent) {
+			continue
+		}
+
 		if c.busyProbe(ctx, name) {
 			slog.Info("idle checker: skipping busy agent", "agent", name)
 			continue
@@ -102,6 +106,23 @@ func (c *IdleChecker) check(ctx context.Context) {
 	}
 	slog.Debug("idle checker sweep complete",
 		"scanned", len(agents.Items), "hibernated", hibernated, "duration", time.Since(start))
+}
+
+func isHibernated(agent *unstructured.Unstructured) bool {
+	conds, found, err := unstructured.NestedSlice(agent.Object, "status", "conditions")
+	if err != nil || !found {
+		return false
+	}
+	for _, raw := range conds {
+		cond, ok := raw.(map[string]any)
+		if !ok {
+			continue
+		}
+		if cond["type"] == apiv1.ConditionReady && cond["reason"] == apiv1.ReasonHibernated {
+			return true
+		}
+	}
+	return false
 }
 
 func hibernationOverride(agent *unstructured.Unstructured) *metav1.Duration {
--- a/packages/controller/pkg/reconciler/idlechecker.go
+++ b/packages/controller/pkg/reconciler/idlechecker.go
@@ -80,6 +80,7 @@ func (c *IdleChecker) check(ctx context.Context) {
 
 	now := time.Now().UTC()
 	timeout := c.config.AgentBase.IdleTimeout.AsDuration()
+	atZero := c.agentsAtZero(ctx)
 	hibernated := 0
 	for i := range agents.Items {
 		agent := &agents.Items[i]
@@ -88,7 +89,7 @@ func (c *IdleChecker) check(ctx context.Context) {
 			continue
 		}
 
-		if isHibernated(agent) {
+		if atZero[name] {
 			continue
 		}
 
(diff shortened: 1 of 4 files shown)


packages/controller/pkg/reconciler/idlechecker.go
```diff
@@ -88,6 +88,10 @@ func (c *IdleChecker) check(ctx context.Context) {
 			continue
 		}
 
+		if isHibernated(agent) {
```
dam-code-guardian: 🟢 **Suggestion:** `Ready`/`Hibernated` is cleared on wake by `publishReadiness`, but the reconcile can return before reaching it — the `gatewayIP == ""` requeue and the `setError` paths after the agent StatefulSet is already scaled to 1. A reconcile that fails there persistently leaves a running pod whose `Ready` reason still reads `Hibernated`, and this guard now skips it once it goes idle, where the old sweep would have scaled it down. Narrow (it needs a durably-erroring reconcile) and the pod shows as not-Ready meanwhile, so this is a note rather than a change request.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `79088f34`

## PR #3470: fix(controller): stop the idle sweep re-hibernating sleeping agents
**Author:** jezekra1 | **Branch:** fix/idle-sweep-reconcile-churn → main | **Changes:** +45 −2 (3 files)

### Summary
The idle sweep now skips agents whose `Ready` condition already reports them hibernated, and the `agentStatefulSet` timing mark splits so the Service write is attributed separately. The measured chain in the body holds at every hop, and the guard's safety claim verifies: `Ready`/`Hibernated` is written in exactly one place, after a successful scale-to-zero.

### Findings
- 🟡 **Warning:** the Hibernate section still describes the idle decision as "deliberately split across two independent checks that must agree" and hibernating "only when *both* checks below agree" — the sweep now short-circuits on a third gate before the probe (`docs/architecture/agent-lifecycle.md:193`, `:195`; reported by doc-drift).
  **Fix:** describe the already-hibernated skip as a precondition of the sweep, in domain terms rather than symbol names, and bump the page's `Last verified` header (currently 2026-08-21).
- 🟢 **Suggestion:** the guard reads a condition that a partly-failed reconcile can leave stale (`packages/controller/pkg/reconciler/idlechecker.go:91`).

- 🟡 `docs/architecture/agent-lifecycle.md:193` — the page frames idleness as "derived from observed activity, never stored" and the decision as two checks that must agree; published status is now also an input, which that framing does not cover.
  **Fix:** update the Hibernate section in this PR so the sweep's precondition is described, and bump `Last verified`.

### Issue Fit
✅ No findings.

### Verdict
COMMENT — the code half verifies clean; the architecture page's description of the idle decision is now incomplete.

---

packages/controller/pkg/reconciler/idlechecker.go
jezekra1: Good catch — I fixed it rather than taking the note, because the consequence is worse than the note suggests: the old sweep would have reaped that pod, so my guard turned a churn fix into a way to leak a running agent indefinitely.

Verified the reachable state you describe: `applyStatefulSet` scales the agent pair up at agent_reconciler.go:224 and `publishReadiness` is at 244, with `setError` paths in between, so a durably-erroring step there does leave replicas=1 with the condition still reading Hibernated.

Rather than hardening the condition, the guard now observes reality: the sweep lists StatefulSets once per pass and skips only agents whose pair is at zero. That drops the dependency on a published condition being correct, costs one read per sweep instead of per agent, and still writes nothing for a sleeping agent — so the queue burst this PR removes stays removed.

`TestIdleChecker_HibernatesRunningAgentDespiteHibernatedStatus` covers your scenario. I checked it fails against the guard as it stood in the previous commit and passes with this one.

jezekra1:
Correction to my earlier reply: the observed-replicas guard I pushed in 85d1c7e5a was wrong on its own, and CI caught it — `TestBudgetParkedRevertsToHibernatedAfterWindow` failed.

An over-budget agent sits at zero replicas because it was never allowed to start, so "observed at zero" matched it, the sweep skipped it, and its Ready reason stayed `OverBudget` instead of flipping to `Hibernated`. I had also claimed it passed locally due to test caching — that was wrong too: I was on the other branch and testing a tree without the guard in it. It reproduces deterministically.

Fixed in 759a3ff27 by requiring **both** signals: the pair observed at zero *and* hibernation already published. That is exactly the steady state a sweep has no work for, which is the churn this PR removes. Either signal alone is wrong in a different direction — replicas alone skips an agent whose status has not caught up, and status alone skips a running pod whose reconcile errored before publishing readiness, which is what you raised.

Tests now cover the three states: skipped, status-not-caught-up, and running-but-mislabelled.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `759a3ff2`

## PR #3470: fix(controller): stop the idle sweep re-hibernating sleeping agents
**Author:** jezekra1 | **Branch:** fix/idle-sweep-reconcile-churn → main | **Changes:** +109 −2 (3 files)

### Summary
The idle sweep now short-circuits an agent only when two independent signals agree: its StatefulSet pair is observed at zero **and** hibernation is already published. This replaces the condition-only guard from the previous HEAD, which could leak a running pod whose status was mislabelled.

### Changes since last review
Previous HEAD: `79088f34` (2026-08-26T14:15:06Z) — verdict COMMENT

- ✅ **Fixed:** the guard no longer trusts the published condition alone; it now requires the pair observed at zero (`packages/controller/pkg/reconciler/idlechecker.go:92`)
- 🔁 **Still present:** the Hibernate section does not describe the sweep's steady-state skip (`docs/architecture/agent-lifecycle.md:193`, `:195`)

### Findings
- 🟡 **Warning:** the Hibernate section still presents the idle decision as two checks that must agree, and does not mention that the sweep now skips an agent already at its desired state (`docs/architecture/agent-lifecycle.md:193`, `:195`; reported by doc-drift).
  **Fix:** when a PR changes a subsystem's decision logic, update that subsystem's architecture page in the same PR — describe the already-hibernated skip as a precondition of the sweep, in domain terms, and bump `Last verified` (currently 2026-08-21).

- 🟡 **Warning:** The idle checker gained a third skip gate. It now skips an Agent that sits at zero replicas and also reports hibernated. The Hibernate section still describes only two checks that must agree before a scale-down. `docs/architecture/agent-lifecycle.md:195`
  **Fix:** Update a subsystem's architecture page in the same PR that changes the subsystem's decision logic. Add the new gate to the Hibernate section and bump the `Last verified:` header to the PR date.

### Issue Fit
- 🟢 **Suggestion:** The PR description says the sweep skips agents whose status reports them hibernated. The guard also requires a live StatefulSet replica count of zero, and it adds an undeclared fleet-wide StatefulSet list on every sweep. `packages/controller/pkg/reconciler/idlechecker.go:92`

### Verdict
COMMENT — the code half verifies clean at this HEAD; the architecture page still describes the previous sweep behaviour.

---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `2d432dbe`

## PR #3470: fix(controller): stop the idle sweep re-hibernating sleeping agents
**Author:** jezekra1 | **Branch:** fix/idle-sweep-reconcile-churn → main | **Changes:** +112 −5 (4 files)

### Summary
The idle sweep skips an agent only when its StatefulSet pair is observed at zero and hibernation is published. The architecture page now records that skip, which closes the last open finding on this PR.

### Changes since last review
Previous HEAD: `759a3ff2` (2026-08-27T11:06:06Z) — verdict COMMENT

- ✅ **Fixed:** the Hibernate section now states the sweep's already-at-rest skip and names both signals (`docs/architecture/agent-lifecycle.md:193`, `:195`); `Last verified` is bumped to 2026-08-27

### Findings

The two-signal guard verifies against the code: `ReasonHibernated` on `ConditionReady` is written in one place only, after a successful scale-to-zero (`idlechecker.go:214`), and `agentsAtZero` returns `nil` on a list error, so a failed listing makes the sweep behave exactly as before the change.

✅ No findings.

### Issue Fit
✅ No findings.

### Verdict
APPROVE — the guard and its tests hold, and the doc finding from the previous two rounds is resolved at its cause.

---

PR #3493, opened 2026-08-27
perf(api-server): serve agent state from a per-replica informer

jezekra1:
## What

Toward #3402 — deliberately not a closing keyword. Agent reads are served from a per-replica cache kept current by a watch, instead of querying the Kubernetes API on every call, and the same watch releases each readiness wait as soon as the condition flips.

**What of #3402 this does not do**, so the remainder is in the body rather than only in the discussion: the duplicate Agent watch is not deleted (live-events keeps its lease-elected one, and unifying them is what forced the broken bus design); the runtime-delivery readiness re-check still runs on its ~1s timer; and the keep-warm activity PATCH is still awaited between observing readiness and connecting. The last one sits on the same relay-attach path the Done-when measures — it is ~100ms of bookkeeping against a 1–2s budget, and making it fire-and-forget changes error handling on the wake path, so it belongs in its own change.

Decision recorded in `docs/adrs/DRAFT-agent-read-cache.md`.

## Why

Every Agent read went to the Kubernetes API — 22 single reads and 5 namespace-wide listings, the listing once per UI request. Separately, the wake path re-read on a backoff stretching to 5s, so an agent could sit ready for up to 6s before anything noticed (mean ~3s, worst ~6s; step 6 of #3246's cold-wake breakdown). Both are the same gap: the api-server had no current picture of Agent state, so it asked again every time it needed one.

## The boundary, which is the actual decision

- **On the cache (9 reads):** the owner listing, ownership and identity resolution, readiness, the wake loop. These only report state.
- **Live (13 reads):** spec read-modify-write, delete, restart, wake, stop, the pause flow's compare-and-clear, and the two authoritative reads that classify a wake failure. Each one's result decides a write — the first can silently drop a concurrent change, and `agent-lifecycle.md` calls the pause's exact-stamp stickiness load-bearing, so a stale read there resurrects an agent meant to stay down.
- **Until the informer syncs, everything reads through**, because an unsynced cache cannot tell "absent" from "unseen", and that distinction feeds the typed wake failures callers surface to users.

`the read boundary this cache is allowed to cross` in the tests enforces this. I checked it fails if `updateSpec`'s read is moved onto the cache, so the rule is executable rather than just documented.

## Unifications rather than additions

- The cache is the client library's own informer (`makeInformer`), not a hand-rolled one.
- The three copies of `KubeConfig` loading in `k8s.ts` collapse into one helper.
- `pollUntilReady`'s three positional timing numbers become named options — that is what let the wake signal in without a second polling loop beside it.
- The k8s test fake moves out of one test file into `__tests__/helpers/`, so the new tests share it instead of copying it.
- The cached listing reuses `agentIsOwnedBy` rather than reimplementing the owner label selector.

## Notes

- The cache is a **required** constructor dependency, not an optional one with a noop default. That is why the diff touches six construction sites — the compiler enumerated them. Callers with no informer (tests, and anything outside the main process) get `createLiveAgentStateCache`, a real adapter with real semantics.
- All three apps share one `bootstrap()`, so there is exactly one informer per replica, stopped in cleanup.
- Per-replica rather than lease-elected: the lease on the live-events watch exists so each transition *projects* to browsers once. A cache read locally projects nothing, so that constraint does not reach it. No RBAC change — same verb on the same resource the api-server already watches.
- Supersedes nothing. ADR-032's rule that observed `Ready` is authoritative still holds; only the observation's source changes.

## Test plan

- [x] `mise run check`, full api-server suite (1189 tests)
- [x] the boundary test verified to fail when a write-adjacent read is moved onto the cache
- [x] the wake test verified to fail without the change signal (full 5s interval vs 33ms)
- [ ] after dev deploy: a measured wake shows ~0 dead time between the Ready condition flipping and relay attach, per #3402's done-when
- [ ] after dev deploy: confirm the agent listing no longer issues a Kubernetes LIST per UI request


jezekra1:
Pivoting this PR. The nudge approach here — watch as a poke, still re-reading to get the answer — is scaffolding that a proper informer deletes, so shipping it first would cost two reviews and two deploys for code we would then remove.

Reset to main; the tested nudge implementation is preserved locally at `c90d90147` if we want it back as an interim step. Marked draft while the informer version is written.

packages/api-server/src/modules/agents/infrastructure/agents-repository.ts
```diff
@@ -66,7 +70,10 @@ export interface AgentsRepository {
   ensureReady(id: string, opts?: { onWaking?: () => void }): Promise<void>;
 }
 
-export function createAgentsRepository(k8s: K8sClient): AgentsRepository {
+export function createAgentsRepository(
+  k8s: K8sClient,
+  changeWatch: AgentChangeWatch = createNoopAgentChangeWatch(),
```
dam-code-guardian: 🔴 **Critical:** The nudge does not reach the wake paths #3402 measures. `changeWatch` defaults to `createNoopAgentChangeWatch()`, whose subscription never settles, and only the `composeAgentsModule` path passes the singleton (`compose.ts:96`). `bootstrap.ts:221` builds `agentsRepo` with one argument, and that instance serves the ACP, SSH and terminal relays plus the tRPC and import proxies (`apps/api-server/app.ts:59,133,139,144`); `modules/skills/compose.ts:72` is the skills wake path. Both keep the old 5s-regime dead time, and #3402's Done-when is stated as relay attach. Reported by the diff review, typescript-engineering and issue-fit.

**Fix:** Thread the watch into every `createAgentsRepository` call site that reaches `ensureReady`, and make the parameter required so an unwired site fails to compile rather than silently losing the behavior. `startAgentChangeWatch` needs only `k8sClient`, so it can move above line 221.

docs/architecture/agent-lifecycle.md
```diff
@@ -86,7 +86,7 @@ Three paths trigger a wake:
...
...the controller's termination cause), pods still progressing, a gateway still coming up, a gateway failure it cannot outgrow, or a reconcile error — which callers receive and surface in their own idiom (channel reply copy, WS close reason, HTTP body, skills call error). The classification distinguishes transient causes (progressing, worth waiting or retrying) from hard ones (needing intervention); the two gateway causes split along exactly that line. A gateway wedged on a superseded configuration counts as hard even though the platform replaces such pods by itself ([security-and-credentials](security-and-credentials.md)): classification runs only once the budget is spent, so a repair that was going to land already had its chance — still seeing it means the repair itself is not working. The primitive also records wake begin/success/timeout with duration and the condition snapshot, so wake latency and failures are diagnosable from the log store. Callers can additionally register for a cold-start signal, fired when a call enters (or joins) a wake wait, to tell their user a wake is underway. The schedule-driven poke is the exception: it doesn't wait, so there is no bounded wait to fail.
```
dam-code-guardian: 🟡 **Warning:** This states the watch "cuts each wait short" without qualification, while the coupling reaches only the repositories built through `composeAgentsModule` — including neither the interactive-session nor the skills-install wake this page lists as wake triggers three paragraphs up. Reported by doc-drift.

**Fix:** Resolving the wiring finding makes the sentence true as written; otherwise name the paths that are nudged.

packages/api-server/src/modules/agents/infrastructure/agent-change-watch.ts
```diff
@@ -0,0 +1,87 @@
...
+      AGENTS_PLURAL,
+      (_phase, resource) => {
+        const agentId = resource.metadata?.name;
+        if (agentId) fire(agentId);
+      },
+      (err) => {
+        if (stopped) return;
+        if (err) log(`agent change watch ended: ${String(err)}`);
+        for (const agentId of [...waiters.keys()]) fire(agentId);
+        retryTimer = setTimeout(connect, RECONNECT_MS);
```
dam-code-guardian: 🟢 **Suggestion:** `connect()` overwrites `stopWatch` and `retryTimer` without releasing the previous ones, so a second `onEnd` from one connection would leave two watches live and a timer `stop()` can no longer reach. The sibling `live-events/infrastructure/k8s-agent-watch.ts:76,79` on `main` has the identical shape, so this PR did not introduce the seam — one issue covering both is the cheaper fix.

packages/api-server/src/__tests__/unit/wake-readiness-nudge.test.ts
```diff
@@ -0,0 +1,99 @@
...
+      5_000,
+      5_000,
+      30_000,
+      () => w.subscribe("agent-1"),
+    );
+    await vi.waitFor(() => expect(checks).toBe(1));
+    watch.emit("agent-1");
+
+    expect(await result).toBe(true);
+    expect(Date.now() - started).toBeLessThan(1_000);
```
dam-code-guardian: 🟢 **Suggestion:** This asserts wall-clock elapsed time against a 5s poll interval on real timers, while the sibling wake test `agents-repository-ensure-ready.test.ts` uses `vi.useFakeTimers()`. Fake timers would remove the CI-stall flake and make "cut the wait short" an exact assertion.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `c90d9014`

## PR #3493: perf(api-server): let a watch cut the wake poll short
**Author:** jezekra1 | **Branch:** feat/3402-wake-readiness-nudge → main | **Changes:** +242 −11 (9 files)

### Summary
`ensureReady` keeps its readiness poll, and a per-replica Kubernetes watch on the Agent resources now cuts the sleep between polls short. The polled read stays the authority, so the watch only removes dead time.

### Findings
- 🔴 **Critical:** The nudge does not reach the wake paths the issue measures (`agents-repository.ts:75`) — see the inline comment.
- 🟡 **Warning:** The page states the watch "cuts each wait short" without qualification (`docs/architecture/agent-lifecycle.md:89`) — see the inline comment.
- 🟡 **Warning:** `platform-topology.md` owns the api-server's control-plane coupling and describes it as REST only, and its "single lease-elected replica" sentence now reads as covering both Agent watches when only the live-events one is leased. Watch connections against the Agent resources now scale with replica count. Reported by doc-drift. (`docs/architecture/platform-topology.md:101`, `:52`)
  **Fix:** Cover watch alongside REST in that connection row, distinguish the two watches at `:52` (lease-elected live-events, per-replica wake-readiness), and bump that page's `Last verified`.
- 🟢 **Suggestion:** On reconnect, `connect()` overwrites `stopWatch` and `retryTimer` without releasing the previous ones, so a second `onEnd` from one connection would multiply watches and orphan a timer (`agent-change-watch.ts:54`). The sibling `live-events/infrastructure/k8s-agent-watch.ts:76,79` on `main` has the identical shape, so this is pre-existing style rather than a regression introduced here — worth one separate issue covering both.
- 🟢 **Suggestion:** The new test drives real timers and asserts wall-clock elapsed time against a 5s poll interval, while the sibling wake test `agents-repository-ensure-ready.test.ts` uses fake timers. Fake timers would remove the CI-stall flake and make "cut the wait short" exact. (`packages/api-server/src/__tests__/unit/wake-readiness-nudge.test.ts:50`)
- 🟢 **Suggestion:** "Closes the api-server half of #3402" is not parsed as a closing keyword, so merging leaves #3402 open. Its remainder is three items, and the activity-PATCH bullet sits on the same relay-attach path the Done-when measures without being among the declared exclusions. Reported by issue-fit.
- ✅ **Looks good:** The loop mechanics hold. Subscribing before the read means an event arriving mid-read is not lost (modelled: 102 ms versus a full interval), a cancelled co-waiter does not disturb its sibling, and a dropped watch resolves its waiters so the loop re-reads rather than sleeping through the reconnect gap.

- 🟡 **Warning:** `agent-lifecycle.md:89` claims the nudge applies to every bounded wake wait; it reaches only the repositories built through `composeAgentsModule`. The wakes the page's own trigger list names — an interactive session, a skills install — are precisely the ones it does not reach.
  **Fix:** Either wire the watch through the remaining call sites so the unqualified claim becomes true, or narrow the sentence to the coupling that exists.
- 🟡 **Warning:** The PR adds a second, unleased, per-replica Agent watch (`bootstrap.ts:459`, deliberately outside the `agentWatchLease` at `:462`), and `platform-topology.md` was not updated — mirrored above.

### Issue Fit
Linked issue: #3402 "A ready agent pod sits unused for ~10 seconds before it gets work" (inferred — the body's phrasing is not a closing keyword, so `closingIssuesReferences` is empty).
- 🟡 **Warning:** The nudge misses the relay wake path, which is the one #3402's Done-when measures ("≤1–2s between the Ready condition flipping and relay attach") — merged into the 🔴 above.
- 🟡 **Warning:** The issue's third Problem bullet and its matching Proposed-solution item — the relay's keep-warm annotation PATCH between observing readiness and connecting becoming fire-and-forget — has no change here and is not among the PR's declared exclusions, which cover only the delivery re-check and the undeleted duplicate watch. It sits on the same relay-attach path.
  **Fix:** Either make that post-readiness PATCH fire-and-forget here, or declare it an exclusion with its reasoning on the issue, alongside the other two.
- 🟢 **Suggestion:** #3402's remainder is three separable items, and the issue comment itself asks whether they stay. Splitting them would let #3402 close on merge rather than staying open with an unstated remainder.

- 🔴 **Critical:** The nudge never reaches the repositories serving the wake hot paths — merged into the 🔴 above, with the call-site list.
- 🟡 **Warning:** The reconnect path discards the previous `stopWatch` and overwrites `retryTimer` without clearing it, so a second `onEnd` from one connection multiplies watches and orphans timers `stop()` can no longer reach (`agent-change-watch.ts:42-58,83`). Priced 🟢 above: the sibling watch on `main` carries the identical shape, so this PR did not introduce the seam.
- 🟢 **Suggestion:** Real timers and a wall-clock threshold in the new test — mirrored above.

### Verdict
REQUEST_CHANGES — the watch is the right mechanism and the loop is correct, but the singleton is threaded into only some of the repositories, so the relay and skills wake paths keep the latency the PR sets out to remove.

---

packages/api-server/src/modules/agents/infrastructure/agents-repository.ts
jezekra1: Fixed, and your fix is the one I took — though by a bigger route than you proposed. This PR was rewritten after your review: the nudge approach is gone and reads now come from a per-replica cache over the client library's informer.

On the substance you were right twice over. The parameter is now **required**, so an unwired site fails to compile — that is exactly what surfaced the six construction sites, including `bootstrap.ts` (the relay, SSH, terminal and proxy paths) and `modules/skills/compose.ts`, both of which I would otherwise have missed. The typescript-engineering rules independently forbid the optional-with-noop-default shape, so the review and the skill converged on the same objection.

Callers with no informer get `createLiveAgentStateCache`, a real adapter with real semantics rather than a null object, so absence is never a silent skip.

docs/architecture/agent-lifecycle.md
jezekra1: Resolved by the rewrite rather than by narrowing the sentence — the coupling now reaches every call site, so the unqualified claim is true. The sentence also grew to cover the read cache and, more importantly, the boundary: reads that decide a write still read through to the API.

packages/api-server/src/modules/agents/infrastructure/agent-change-watch.ts
jezekra1: That file no longer exists — the rewrite uses the client library's informer, which owns its own reconnect. So the seam is gone from this PR, and you were right that the remaining instance in `live-events/infrastructure/k8s-agent-watch.ts` predates it. Filed as #3494 covering that one, with the failure mode written out.

packages/api-server/src/__tests__/unit/wake-readiness-nudge.test.ts
jezekra1: Taken, and it made the assertion stronger. That test file is gone with the rewrite, but its successor had the same shape, so it now runs on fake timers and asserts exactly: two reads, and `vi.getTimerCount()` back to zero — which also proves the sleep timer is cleared rather than merely raced. Verified it still fails if the wake signal is ignored.

jezekra1:
Review was against `c90d9014`, which this PR no longer contains — it was rewritten from the nudge approach to a per-replica read cache after the review landed. Replies are on each thread; summary of where each finding went:

- 🔴 **Nudge misses the wake paths** — fixed, and by the route you proposed. The dependency is now required, which is what surfaced all six construction sites including `bootstrap.ts` and the skills path.
- 🟡 **`agent-lifecycle.md` overclaims** — resolved by the wiring rather than by narrowing the sentence.
- 🟡 **`platform-topology.md` says REST only, and its lease sentence now covers two watches** — fixed in 7c0f39157: the connection row and diagram edge say REST + watch, and the lease sentence distinguishes the elected browser projection from the per-replica read cache, noting watch count tracks replica count. `Last verified` bumped.
- 🟢 **Reconnect seam** — the file is gone; the remaining instance in live-events is filed as #3494.
- 🟢 **Real timers in the test** — taken, and it made the assertion exact: two reads and no pending timer, so the sleep is provably cleared rather than raced.

Two issue-fit points I have **not** changed, deliberately:

- **The closing keyword.** #3402 is meant to stay open — its remainder is the undeleted duplicate watch, the ~1s delivery re-check, and the post-readiness PATCH below. I would rather it stay open with the remainder stated on the issue than close on merge with three items quietly dropped. Splitting it is the better long-term answer and I have not done that here.
- **The keep-warm PATCH.** You are right it was not among the declared exclusions and that it sits on the relay-attach path the Done-when measures. Declaring it now: it is ~100ms of bookkeeping against a 1–2s budget, and making it fire-and-forget changes error handling on the wake path, which I would rather do as its own change than fold into a read-path refactor. Added to #3402's remainder.

packages/api-server/src/modules/agents/infrastructure/agent-state-cache.ts
```diff
@@ -0,0 +1,116 @@
...
+  };
+  const onObject = (obj: KubeObject) => {
+    const id = obj.metadata?.name;
+    if (id) release(id);
+  };
+
+  deps.informer.on("add", onObject);
+  deps.informer.on("update", onObject);
+  deps.informer.on("delete", onObject);
+  deps.informer.on("connect", () => {
```
dam-code-guardian: 🔴 **Critical:** `synced` is set from the informer's `connect` event, but at the pinned `@kubernetes/client-node` 1.4.0 `ListWatch.doneHandler` fires `CONNECT` (`src/cache.ts:162`) *before* it awaits `listFn()` and fills its store (`:166–173`). Between those two lines the cache calls itself authoritative over an empty map, so `get` returns `null` and `list` returns `[]` instead of falling back to live reads.

Measured by transcribing that handler: during a 50 ms list, `get("a1")` returned `null` and `list()` was empty for an agent that exists; both were correct once the list settled. This is the window the boot-time `not-found` wake failure and an empty owner listing come out of, and the suite cannot see it because the test fake resolves `start()` without modelling the ordering.

**Fix:** Mark a watch-backed cache authoritative only once its own store is known filled — flip on the first `add`, or gate on `informer.latestResourceVersion()` being non-empty — never on a connection-level event whose contract does not promise a populated store.

```suggestion
  deps.informer.on("connect", () => {
    if (deps.informer.latestResourceVersion()) synced = true;
  });
```

packages/api-server/src/modules/agents/infrastructure/agent-state-cache.ts
```diff
@@ -0,0 +1,116 @@
...
+  });
+  deps.informer.on("error", (err) => {
+    synced = false;
+    deps.log(
+      `agent cache desynced, falling back to live reads: ${String(err)}`,
+    );
+    releaseAll();
+  });
+
+  void deps.informer.start().catch((err) => {
```
dam-code-guardian: 🔴 **Critical:** Nothing restarts the informer. At 1.4.0 a non-410 watch failure runs `doneHandler`'s `err` branch, which fires `ERROR` and `return`s without reopening the watch (`src/cache.ts:154–156`); the 410 path resets `resourceVersion` and re-lists, but a socket reset, an API-server restart or an auth blip does not. This handler sets `synced = false` and logs, so the replica then serves every Agent read live for the rest of its lifetime and `whenChanged` never resolves again — the wake nudge is silently gone.

Modelled against that handler: after one `socket hang up`, no further list was issued, `synced` stayed false, and a wake wait fell back to the full poll interval. The library's own informer example restarts on `error` for exactly this reason, and the repo's sibling `live-events/infrastructure/k8s-agent-watch.ts:79` already reconnects on a timer.

Nothing surfaces the degradation either, so the ADR's "committed-to: a watch per replica" reverts to the pre-PR read volume with only a warn line to show it.

**Fix:** Reconnect on `error` with backoff, guarded by a `stopped` flag so shutdown stays terminal, and follow the sibling watch's pattern rather than leaving a one-shot log.

packages/api-server/src/modules/agents/infrastructure/agents-repository.ts
```diff
@@ -274,7 +280,7 @@ export function createAgentsRepository(k8s: K8sClient): AgentsRepository {
       }
 
       const work = (async () => {
-        const current = await k8s.getCustomObject(AGENTS_PLURAL, id);
+        const current = await cache.get(id);
```
dam-code-guardian, line 283: 🟡 **Warning:** This typed `not-found` comes from `cache.get(id)`, which cannot tell "absent" from "not delivered yet". Creates go through the live client, so an agent created moments before a wake — or any wake during the boot sync window — surfaces a hard user-visible not-found instead of waiting. The ADR names this exact distinction as the reason unsynced reads fall through to the API; fixing the `connect` ordering above closes the window, and this line is where the cost lands if it reopens.

**Fix:** Never promote a cache miss to a negative-existence claim — confirm absence with a live read before emitting a not-found-class outcome.

docs/architecture/agent-lifecycle.md
```diff
@@ -86,7 +86,7 @@ Three paths trigger a wake:
...
```
dam-code-guardian: 🟡 **Warning:** The paragraph now describes the cache and the write-side split, but not the precondition that makes the split safe: reads fall back to the Kubernetes API until the watch has synced, and again whenever it desyncs. That fallback is what keeps the typed wake failures this same paragraph documents from turning a not-yet-seen agent into a not-found.

**Fix:** When a documented read path gains a cache with a correctness precondition, state the precondition and the fallback that holds it, not only that the cache exists.

packages/api-server/src/modules/agents/infrastructure/poll-until-ready.ts
```diff
@@ -1,16 +1,42 @@
...
+
+function sleep(ms: number): { elapsed: Promise<void>; cancel(): void } {
+  let timer: NodeJS.Timeout | undefined;
+  const elapsed = new Promise<void>((resolve) => {
+    timer = setTimeout(resolve, ms);
+  });
+  return { elapsed, cancel: () => clearTimeout(timer) };
+}
+
 export async function pollUntilReady(
```
dam-code-guardian, line 20: 🟢 **Suggestion:** `wakeOn` resolving re-enters the loop with no floor, so an object receiving rapid updates can spin this body as fast as the watch delivers events. A small minimum gap between wake-driven re-checks would keep the change signal from becoming a busy loop, while leaving the latency win intact.

packages/api-server/src/modules/agents/infrastructure/k8s.ts
```diff
@@ -187,10 +193,27 @@ export function podBaseUrl(agentId: string, namespace: string): string {
...
   return {
     api: kc.makeApiClient(k8s.CoreV1Api),
     namespace,
   };
 }
+
+export type AgentInformer = k8s.Informer<KubeObject> &
+  k8s.ObjectCache<KubeObject>;
+
+export function createAgentInformer(namespace: string): AgentInformer {
```
dam-code-guardian, line 210: 🟢 **Suggestion:** This is a third independent `KubeConfig` load, one commit after the PR consolidated the other two into `loadKubeConfig()`. Threading the already-loaded config in would keep credential loading in one place.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `7c0f3915`

## PR #3493: perf(api-server): serve agent state from a per-replica informer
**Author:** jezekra1 | **Branch:** feat/3402-wake-readiness-nudge → main | **Changes:** +602 −108 (20 files)

### Summary
The api-server now serves Agent reads from a per-replica informer cache, and the same watch releases each readiness wait as soon as the condition flips. The read/write split is recorded in a draft ADR and enforced by a test.

### Changes since last review
Previous HEAD: `c90d9014` (2026-08-27T08:37:38Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** The nudge now reaches every wake path. The cache is a required constructor parameter and all six construction sites receive the one informer-backed instance (`bootstrap.ts:228`).
- ✅ **Fixed:** `agent-lifecycle.md:89` scopes the claim and names the live-read exceptions.
- ✅ **Fixed:** `platform-topology.md` records the second, unleased watch at `:52`, in the connection table and in the mermaid edge.
- 🔁 **Still present:** The reconnect path overwrites `stopWatch` and `retryTimer` without releasing the previous ones (`agent-change-watch.ts` shape, now in `agent-state-cache.ts`). Pre-existing sibling shape, still 🟢.
- 🔁 **Still present:** "Closes the api-server half of #3402" is not a closing keyword, so merging leaves #3402 open.
- 🆕 **New:** The cache marks itself synced before the informer's initial list lands (`agent-state-cache.ts:64`).
- 🆕 **New:** A dropped watch never reconnects, so the cache degrades permanently (`agent-state-cache.ts:75`).
- 🆕 **New:** Two write-deciding reads moved onto the cache (`bootstrap.ts:906`, `session-presence.ts:103`).

### Findings
- 🔴 **Critical:** `synced` is set from the informer's `connect` event, which fires before the initial list populates the store (`agent-state-cache.ts:64`) — see the inline comment.
- 🔴 **Critical:** A dropped watch leaves the cache permanently desynced (`agent-state-cache.ts:75`) — see the inline comment.
- 🟡 **Warning:** Two reads that decide a write moved onto the cache, which the ADR's own second rule forbids. The agent sweep lists through the cache and deletes from that result (`bootstrap.ts:906` → `agent-sweep.ts:31`), and `session-presence.reconcile()` clears the active-session flag for every agent in a cache-backed annotation listing (`session-presence.ts:103`). Both read live on `main`. Reported by typescript-engineering.
  **Fix:** Route every write-deciding read through the live `K8sClient`, not the cache — for these two, give the sweep and the presence reconcile a live lister rather than `agentsRepo.list` / `listAgentIdsWithAnnotation`.
- 🟡 **Warning:** `agent-lifecycle.md:89` documents the cache and the write-side split but omits the live-read fallback that holds the split's precondition (`docs/architecture/agent-lifecycle.md:89`). Reported by doc-drift.
  **Fix:** When a documented read path gains a cache with a correctness precondition, state the precondition and the fallback that holds it — here, that reads fall back to the API until the watch has synced and whenever it desyncs.
- 🟡 **Warning:** The body claims to close "the api-server half" of #3402 while two of the issue's three api-server Problem bullets are unchanged: the runtime-delivery readiness re-check still runs on its ~1s timer (`state-queue.ts:16`), and the keep-warm activity PATCH is still awaited between observing readiness and connecting (`agents-repository.ts:296`). Both remainders are declared in the issue discussion but not in the body, and the body is also not a closing keyword, so merging leaves #3402 open. Reported by issue-fit.
  **Fix:** Scope the body's closing claim to what the diff delivers and name the remainder in the body itself.
- 🟢 **Suggestion:** `createLiveAgentStateCache().whenChanged()` returns a permanently-pending promise with a no-op `cancel` (`agent-state-cache.ts:31`). A named constant, or omitting `wakeOn` at those call sites, would make the degradation to pure interval polling legible.
- 🟢 **Suggestion:** `pollUntilReady` re-enters the loop immediately whenever `wakeOn` resolves (`poll-until-ready.ts:18`), so a rapidly-changing object can spin the poll body with no pacing. A small floor between wake-driven re-checks would bound it.
- 🟢 **Suggestion:** The fake informer in `agent-state-cache.test.ts:22` resolves `start()` before any event and never models the library's connect-before-list ordering, so neither 🔴 above has a reachable test. Modelling the real ordering in the fake would make the boundary the suite already asserts executable.
- 🟢 **Suggestion:** `createAgentInformer` loads a third independent `KubeConfig` alongside `createK8sClient` and `createApi` (`k8s.ts:206`), one commit after the PR consolidated the other two.

- 🟡 **Warning:** The cache's live-read fallback before initial sync, and after a watch error, is missing from the page — mirrored above. Both prior-review doc findings are resolved at this HEAD, and both edited pages' `Last verified` headers are bumped.

### Issue Fit
Linked issue: #3402 "A ready agent pod sits unused for ~10 seconds before it gets work" (inferred — the body's phrasing is not a closing keyword, so `closingIssuesReferences` is empty).
- 🟡 **Warning:** Two of the issue's three api-server Problem bullets are unchanged, so the Done-when is not met — merged above. The third bullet is now an author-recorded exclusion in the issue discussion, so it is no longer undeclared.
- 🟢 **Suggestion:** `agent-lifecycle.md` also rewrites two paragraphs unrelated to the read cache — the Agent-resource description at `:7` and the `dam-run` section at `:230`.

- 🔴 **Critical:** `synced` is flipped on `connect`, which the pinned client library fires before populating its store — mirrored above with the measurement.
- 🔴 **Critical:** The cache-backed listing behind the agent sweep decides a delete — priced 🟡 above: the delete itself re-reads live, so the blast radius is a skipped or retried sweep, not a wrong deletion.
- 🟡 **Warning:** `ensureReady` derives its typed `not-found` from `cache.get(id)`, which cannot distinguish absent from not-yet-delivered (`agents-repository.ts:283`). Reachable only inside the boot sync window, which the 🔴 above closes.
  **Fix:** Never promote a cache miss to a negative-existence decision — confirm absence with a live read before emitting a not-found outcome.
- 🟡 **Warning:** `session-presence.reconcile()` clears state from a cache-backed annotation listing — mirrored above.
- 🟡 **Warning:** `startAgentStateCache` fire-and-forgets `informer.start()` and only logs a failure — merged into the reconnect 🔴 above.
- 🟢 **Suggestion:** `stop()` does not detach the informer listeners (`agent-state-cache.ts:110`). Modelled against the pinned library: `stop()` sets its own `stopped` flag first and `doneHandler` returns before firing `CONNECT`, so `synced` cannot re-flip. Hygiene only.
- 🟢 **Suggestion:** The `AgentStateCache` type is imported through the deep path `../agents/infrastructure/agent-state-cache.js` rather than the module's `index.ts` (`skills/compose.ts:11`). The seam pre-exists on `main` for `agents-repository.js` and `k8s.js`, so this widens it rather than introducing it.
- 🟢 **Suggestion:** Verified positively — every construction site receives the single informer-backed cache, and it is stopped in cleanup.

### Verdict
REQUEST_CHANGES — the wiring gap from the last round is properly closed and the loop mechanics hold, but the cache reports itself authoritative before its first list lands, and a dropped watch never comes back.

---

packages/api-server/src/modules/agents/infrastructure/agent-state-cache.ts
```diff
@@ -0,0 +1,140 @@
...
+  };
+
+  let stopped = false;
+  let restartTimer: NodeJS.Timeout | undefined;
+
+  async function connect(): Promise<void> {
+    if (stopped) return;
+    try {
+      await deps.informer.start();
+      synced = true;
```
dam-code-guardian: 🔴 **Critical:** When the informer's initial list fails, the pinned `@kubernetes/client-node` 1.4.0 `ListWatch.doneHandler` fires its `error` callbacks and then **returns normally** — it does not throw (`src/cache.ts:169–172`). So `await deps.informer.start()` resolves, and this line sets `synced = true` immediately after the `error` handler set it `false`. The cache then answers `get`/`list` from an empty store for the whole `RESTART_DELAY_MS` window, and an empty store reports every Agent **absent** — the one thing the ADR's third rule forbids, because absence drives `ensureReady`'s typed `not-found` wake failure (`agents-repository.ts:283`) and ownership 404s.

Modelled against the pinned library, with a healthy control:

```
D  synced=true  storeSize=0  get("a1")=null   events=["errcb"]   <- initial list threw
E  synced=true  storeSize=1  get("a1")=cached events=[]          <- control, list resolved
```

Event order on the failing path: library fires ERROR → cache sets `synced=false` and schedules the restart → `doneHandler` returns → `start()` resolves → this line sets `synced=true`. The restart 5s later does recover it, so the window is bounded but fully open. `makeInformer` passes `autoStart=false`, so `connect()` is the only entry and every boot takes this path.

**Fix:** Never derive a "synced" flag from a promise the library resolves on both the success and the failure path — gate it on observed store content, or on a flag the `error` handler can veto.

packages/api-server/src/modules/agents/infrastructure/agent-state-cache.ts
```diff
@@ -0,0 +1,140 @@
...
+        changed,
+        cancel() {
+          const current = waiters.get(id);
+          if (!current) return;
+          current.delete(resolve);
+          if (current.size === 0) waiters.delete(id);
+        },
+      };
+    },
+    async stop() {
```
dam-code-guardian, line 145: 🟡 **Warning:** `stop()` cannot cancel an in-flight `connect()`. If shutdown lands while `start()` is parked inside the library's `watch.watch()` call, `stop()` sets `stopped = true` and `synced = false` — then the parked `start()` resolves and this line sets `synced = true` again on a stopped cache. The library assigns `this.request` *after* that same `await` (`cache.ts:188`), so the watch connection also outlives the `_stop()` that already ran.

Modelled against the pinned library:

```
B mid-boot:                      synced=false
B right after stop():            synced=false  lib.stopped=true
B after parked start() resolves: synced=true   request-leaked=true  get()=cached
C control (stop after boot):     synced=false  get()=LIVE
```

Reachable at every rolling restart: `bootstrap.ts:1004` awaits `agentStateCache.stop()` during shutdown, and a boot that is still connecting is exactly when the watch call is parked.

**Fix:** Re-check the cancellation flag after every `await` that can outlive it.

```suggestion
      await deps.informer.start();
      if (stopped) {
        await deps.informer.stop();
        return;
      }
      synced = true;
```

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `e66a414e`

## PR #3493: perf(api-server): serve agent state from a per-replica informer
**Author:** jezekra1 | **Branch:** feat/3402-wake-readiness-nudge → main | **Changes:** +721 −110 (20 files)

### Summary
Round 3 answers all three round-2 findings at their cause: `synced` now flips only after `informer.start()` resolves, a dropped watch restarts on a 5s timer, and the two write-deciding readers move onto a separately-composed live repository. One ordering gap survives inside the same fix.

### Changes since last review
Previous HEAD: `7c0f3915` (2026-08-27T10:46:27Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** `synced` is no longer set from the `connect` event; it is set after `start()` resolves (`agent-state-cache.ts:70`)
- ✅ **Fixed:** the `error` handler and a failed start both `scheduleRestart()`, and `stop()` cancels it (`agent-state-cache.ts:81`)
- ✅ **Fixed:** the agent sweep and `session-presence.reconcile()` read through `liveAgentsRepo` (`bootstrap.ts:232`, `:913`)
- ✅ **Fixed:** the fake informer now models connect-before-list and a stopping error, so both round-2 defects have failing tests
- ✅ **Fixed:** `whenChanged`'s no-op is named `noChangeSignal`; `loadKubeConfig` memoizes
- ✅ **Fixed:** repeat wake-driven re-checks are paced by `MIN_WAKE_GAP_MS` (`poll-until-ready.ts:38`)
- 🔁 **Still present:** `ensureReady` derives its typed `not-found` from a cache miss (`agents-repository.ts:283`) — now only reachable through the 🔴 below
- 🆕 **New:** a failed initial list resolves `start()` normally, so `synced` is set over an empty store (`agent-state-cache.ts:71`)

### Findings
- 🔴 **Critical:** A failed initial list resolves `start()` instead of rejecting, so `synced` is set over an empty store (`packages/api-server/src/modules/agents/infrastructure/agent-state-cache.ts:71`) — see the inline comment.
- 🟡 **Warning:** `stop()` cannot cancel an in-flight `connect()`, so a parked `start()` can re-set `synced` on a stopped cache and leak its watch request (`packages/api-server/src/modules/agents/infrastructure/agent-state-cache.ts:71`, `:132`) — see the inline comment.
- 🟢 **Suggestion:** `RESTART_DELAY_MS` is a flat 5s with no backoff and no attempt cap (`agent-state-cache.ts:20`), so a persistently-unreachable watch re-lists every 5s indefinitely. The sibling `k8s-agent-watch.ts:79` on `main` has the same shape, so this is existing style rather than a regression.

✅ No findings.

### Issue Fit
Linked issue: #3402 "A ready agent pod sits unused for ~10 seconds before it gets work" (inferred from the head branch and a plain `Toward #3402` reference — no closing keyword)

- 🟢 **Note:** Two edits to `docs/architecture/agent-lifecycle.md` are unrelated to the read cache and not declared in a body that otherwise enumerates every extra it carries: the overview's "a single `agent` ConfigMap holds both definition and runtime state" becomes a custom-resource description (`:7`), and the paragraph recording the remote Run-executor machinery's removal is deleted (`:232`). Both read as staleness pruning under the page's bumped `Last verified: 2026-08-27`.

The round-1/2 body over-claim is resolved: the body opens "Toward #3402 — deliberately not a closing keyword", and its "What of #3402 this does not do" list matches the issue's own remainder comment. The body's read boundary is accurate against the code — 9 cache reads and 13 live reads in `agents-repository.ts`, with the write-deciding and failure-classifying reads left live.

✅ No findings.

### Verdict
REQUEST_CHANGES — the wiring, the restart and the write-read split are all properly closed, but the sync flag is still set over an empty store when the initial list is the thing that failed.

---

packages/api-server/src/modules/agents/infrastructure/agent-state-cache.ts
```diff
@@ -0,0 +1,154 @@
...
+    if (stopped) return;
+    const mine = ++attempt;
+    try {
+      await deps.informer.start();
+    } catch (err) {
+      abandon(`agent cache could not start: ${String(err)}`);
+      return;
+    }
+    if (stopped) {
+      void deps.informer.stop();
```
dam-code-guardian, line 80: 🟢 **Suggestion:** This `void deps.informer.stop()` has no rejection handler, and the api-server installs no `unhandledRejection` guard. The module's other fire-and-forget calls attach one (`k8s.ts:184`, `agents-repository.ts:213`). Reported by typescript-engineering.

```suggestion
      void deps.informer.stop().catch(() => {});
```

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `cee562c0`

## PR #3493: perf(api-server): serve agent state from a per-replica informer
**Author:** jezekra1 | **Branch:** feat/3402-wake-readiness-nudge → main | **Changes:** +768 −110 (20 files)

### Summary
The api-server serves Agent reads from a per-replica informer cache, and the same watch releases each readiness wait as soon as the condition flips. This round adds a generation counter, so a `start()` that resolves after an error or after `stop()` can no longer mark the cache authoritative.

### Changes since last review
Previous HEAD: `e66a414e` (2026-08-27T11:54:05Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** A failed initial list no longer marks the cache synced. `abandon()` bumps `attempt` before scheduling, so the in-flight `connect()` fails its `mine !== attempt` gate (`agent-state-cache.ts:83`).
- ✅ **Fixed:** `stop()` during an in-flight `connect()` is handled. `stop()` bumps `attempt` too, and the post-await `stopped` check stops the informer, which aborts the library's already-assigned `this.request`.
- ✅ **Fixed:** The flat 5s restart is now exponential with a 60s cap, resetting only after a connect that actually claimed `synced` (`agent-state-cache.ts:103`).
- 🔁 **Still present:** `ensureReady` derives its typed `not-found` from a cache read (`agents-repository.ts:283`). No longer a defect: the empty-store path that made it reachable is closed, and an unsynced cache reads through to the API.

### Findings
- 🟢 **Suggestion:** The late-`stop()` path fires `void deps.informer.stop()` with no rejection handler, while the module's other fire-and-forget calls attach one (`k8s.ts:184`, `agents-repository.ts:213`). Reported by typescript-engineering. (`packages/api-server/src/modules/agents/infrastructure/agent-state-cache.ts:80`)

✅ No findings.

### Issue Fit
Linked issue: #3402 "A ready agent pod sits unused for ~10 seconds before it gets work" (plain reference — "Toward #3402", deliberately non-closing)

✅ No findings.

- 🟢 **Suggestion:** The late-`stop()` path fires `void deps.informer.stop()` with no rejection handler, and the api-server installs no `unhandledRejection` guard, so a rejecting stop on this rare path would reach the process. Adding `.catch(() => {})` matches the module's own convention (`agent-state-cache.ts:80`).

### Verdict
APPROVE — the lifecycle is correct. The generation counter was modelled against pinned client-node 1.4.0 across six interleavings (late success, late rejection, error mid-start, stop mid-start, error burst, recovery), and none reaches a synced-over-empty-store or a permanently-unsynced-without-restart state; the backoff ladder measured 5s→10s→20s→40s→60s with the cap holding.

---

PR #3500, opened 2026-08-28
fix(channels): a conversation answers a burst of messages once

### LINKED ISSUE #3453: A request split across several messages gets a separate reply to each
## Context

Reported from live use: a user typed a request to an agent as a few consecutive messages — the way people write in chat — and got several replies instead of one.

This issue was first written as a problem with channels where the agent reads along. That framing was too narrow: the behaviour shows up wherever an agent is reached, direct messages and @-mentions included, and read-along channels are the one place it is sometimes hidden.

## Problem

People write in chat the way they talk, spreading one thought over two or three messages. An agent treats each message as a separate request and answers it on its own, so an unfinished thought gets answered before it is finished and then answered again once the rest arrives. The user ends up with two or more replies to one question, the first premature and usually not useful, and has to work out which to read — in front of everyone else in the channel.

Each fragment also spends a full agent run on the binding owner's credentials, so the same question costs several times what it should.

## Expected Behavior

An agent answers a thought once. Messages that arrive before it has replied are taken together, and its answer covers all of them.

## Actual Behavior

Each message draws its own reply, so a request split over three messages draws three.

One case behaves differently, which is why the problem looked narrower than it is: in a channel where the agent reads along, messages arriving while it is still working on the previous one are taken together, so a quick burst there does get a single answer. Once it has finished — or anywhere it was addressed directly — the next message starts a fresh reply regardless of how soon it arrives.

## Steps to Reproduce

1. Send an agent a direct message with the first half of a request — "can you check the deploy".
2. Before it replies, send the rest — "specifically whether the migration ran".
3. It replies twice: once to the fragment alone, then again with the whole request.

Reproduces the same way with @-mentions in a shared channel, mentioning the agent in both messages. In a read-along channel, wait for the first reply to land before sending the second.

pilartomas:
Closes #3453

## What was wrong

People write a thought over several messages. Every relayed message got its own turn and its own reply, so a split thought drew several partial answers — the first premature and usually useless — and spent a full agent run per fragment on the binding owner's credentials.

Three different shapes of the same cause:

- **The addressed path had no coalescing at all.** `withSessionTurnLock` is a promise chain keyed by (agent, thread key) — a lock, not a queue. N messages ran as N sequential turns, whatever the gap between them.
- **Read-along coalesced only half the problem.** It absorbed messages that arrived *while* a turn was in flight, so a fast burst got one answer and a slower one got two. That gap-dependence is what made the behaviour look like a read-along quirk rather than a general defect.
- **Telegram had neither a lock nor a queue.** Concurrent messages relied on the runtime's prompt scheduler and became one turn each.

The original issue framed this as a missing debounce in read-along only. That framing was wrong in both directions: read-along was the one path that already coalesced, and the addressed path — mentions, group DMs, 1:1 DMs — was the one with nothing.

## What changed

A conversation now runs through **one queue**, shared by both workers:

- **A burst becomes one turn.** Messages gather behind a short quiet period, and the turn carries all of them.
- **A mid-turn arrival is steered into the running turn**, through the ACP `_session/steering` extension. The agent reads it before it calls its reply tool, so the conversation still gets a single answer. No cancellation, so no lost work and no half-applied side effects.
- **Where the harness does not support steering, the message becomes the next turn** rather than a second turn racing the first. Support is detected per harness from `InitializeResponse._meta.steering.supported`, and the steer is sent with `idleBehavior: "promptRequired"` so a steer that finds no running turn reports back instead of silently starting a detached one.

The batching rule itself is a pure function in `domain/turn-coalescing.ts`; the ordering, quiet period and one-turn-at-a-time rule live in `infrastructure/conversation-queue.ts`, which both adapters use so they cannot drift apart.

**Keying is per conversation** — the thread where there is one, else the channel and sender on Slack, the chat on Telegram. Two people raising different things therefore stay separate turns, each answered on its own. Direct messages need this keying specifically: a turn's session key is `channel:thread_ts` and an unthreaded message falls back to its own ts, so consecutive DMs would otherwise never share a queue.

**Bounds.** Batches are capped, and the quiet period stops after a bounded number of extensions, so neither a prompt nor the wait can grow without limit. A message carrying attachments is never steered — attachments reach the agent through their own turn's delivery path — and it holds back the messages behind it, so nothing is reordered.

The quiet period is off in the worker's own default and set at the composition root. Existing tests keep their immediate-turn behaviour, and the latency cost is opted into in exactly one place.

## Notes for review

- Three `slack-turn.test.ts` scenarios that asserted an id-less reply is refused while two turns are in flight fired two top-level mentions **from the same user**, which is now correctly coalesced into one turn. They fire from two users instead, so they still exercise concurrent turns. The protection they guard also survives by a second route: a coalesced top-level batch registers one turn ref per message, so their reply targets disagree and an id-less reply is refused exactly as before.
- Each batched message is still security-logged individually as a place-basis allow; a steered message is logged the same way, marked as steer-triggered.
- The chat UI is deliberately **not** part of this. It queues a mid-turn prompt and shows it as a queued bubble, which is a visible, intentional contract with rendering built on it — unlike the channels, nothing there is silently answered twice. It also holds its own ACP connection, so it cannot reuse the api-server-side steer added here.

## Verification

- api-server unit suite: **1213 passing**, 21 added (9 for the batching policy, 7 for the queue, 5 driving the Slack worker end to end).
- `tsc --noEmit`, the package's exact `prettier --check` globs, `eslint .`, comment-types, event-coverage and doc-size all pass.
- The only failures in the suite are the two pre-existing `cluster-ca-probe` tests, confirmed to fail identically on an unmodified baseline.
- Not exercised against a live Slack workspace or Telegram chat; coverage is the unit suite driving both workers against fakes, including a fake harness that accepts, refuses, or does not support steering.

## Docs

`docs/architecture/channels.md` stated the coalescing rule as in-flight-only and read-along-only, and now carries the shared rule. The page was 78 characters from its size cap, so the space came from cutting content that had dropped to the level of the code — the `threadKey` mechanics, the literal history-label strings, and the Telegram binding's column names — rather than from rewording.


packages/api-server/src/modules/channels/infrastructure/conversation-queue.ts
```diff
@@ -0,0 +1,115 @@
...
+    const epoch = turnEpoch;
+    const plan = planCoalescedDelivery({
+      pending,
+      turnInFlight: true,
+      steerable: true,
+      ...(deps.canSteer ? { canSteer: deps.canSteer } : {}),
+    });
+    if (plan.kind !== "steer") return;
+
+    const injected = await deps.steer(target, plan.batch);
```
dam-code-guardian: 🔴 **Critical:** The batch is read from `pending` here but removed only after `await deps.steer(...)` returns. Any message arriving inside that round trip re-plans the whole leading run, so an earlier message is steered a second time — and each repeat re-runs `onSteered`, which calls `beginTurn`, writes another place-basis security-log entry, and calls `release()` again.

Reproduced on a standalone transcription of this file with one turn held open: with a 300 ms steer round trip and a 150 ms gap between two mid-turn messages, the steer calls were `[b]` then `[b,c]`, and `b` reached the agent twice. It does not need two senders — any two messages closer together than one ACP round trip hit it.

**Fix:** in both queue paths that read `pending` and mutate it across an await, remove the claimed batch from `pending` before the awaited call and put it back at the head if the call refuses.

packages/api-server/src/modules/channels/infrastructure/conversation-queue.ts
```diff
@@ -0,0 +1,115 @@
...
+    if (!injected) return;
+
+    const steered = new Set(plan.batch);
+    pending = pending.filter((message) => !steered.has(message));
+    deps.onSteered?.(plan.batch, turnEpoch === epoch);
```
dam-code-guardian: 🔴 **Critical:** When the running turn ends while this steer is still in flight, `drain` has already taken the same message out of `pending` and run it as its own turn — and then the late `injected` result reaches `onSteered` anyway. The message is delivered twice: once inside the finished session, once as a new turn.

Reproduced on the transcription with a steer round trip outliving the turn: turns ran `[a]` then `[b]` while `steer(["b"])` returned `injected`. `turnEpoch` is captured at line 60 but only travels to `onSteered` as the `turnStillRunning` flag, so it reports the crossing rather than preventing it.

**Fix:** treat a turn boundary crossed during the steer as a refusal — compare the captured epoch before the batch is committed, and leave it queued for the next turn.

packages/api-server/src/modules/channels/infrastructure/conversation-queue.ts
```diff
@@ -0,0 +1,115 @@
...
+      seen !== pending.length &&
+      settleRoundsRemaining(rounds, pending.length)
+    ) {
+      seen = pending.length;
+      rounds += 1;
+      await delay(deps.settleMs);
+    }
+  }
+
+  async function steerPending(): Promise<void> {
```
dam-code-guardian: 🟡 **Warning:** `unsupported` is a property of the harness, not of the message, but every mid-turn arrival asks again — and each ask opens a fresh ACP WebSocket and runs `initialize` (`core/acp-client.ts:414`). A five-message burst against a harness without steering pays five connects for an answer already known.

**Fix:** cache an `unsupported` answer for the conversation's running turn and hold the backlog, rather than re-asking per arrival.

packages/api-server/src/modules/channels/infrastructure/slack.ts
```diff
@@ -2193,27 +2225,196 @@ export function createSlackWorker(
...
+    slackUserId: string,
+    msg: PendingMessage,
+  ): Promise<void> {
+    const key = addressedQueueKey(conversation, slackUserId);
+    let queue = addressedQueues.get(key);
+    if (!queue) {
+      queue = createAddressedQueue(key, conversation);
+      addressedQueues.set(key, queue);
     }
+    return queue.submit(msg);
```
dam-code-guardian, line 2426: 🟡 **Warning:** The queue is cached per key and built from the first message's `AddressedConversation`, which `createAddressedQueue` closes over. Later messages on the same key therefore run with the first message's `roster`, `ambient`, `ambiguousName`, `owner` and `teamId`. `telegram.ts:474` has the same shape with the first `thread` object.

**Fix:** carry the per-message conversation values on the queued message and read them from the batch at turn time, in both adapters.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `fbc39f4`

## PR #3500: fix(channels): a conversation answers a burst of messages once
**Author:** pilartomas | **Branch:** fix/message-coalescing → main | **Changes:** +1238 −79 (21 files)

### Summary
A conversation now runs through one shared queue, so a burst of messages becomes a single agent turn on both Slack and Telegram. Mid-turn arrivals are steered into the running turn where the harness supports it, and become the next turn where it does not.

### Findings
- 🔴 **Critical:** `steerPending` re-steers the same messages, because it removes the batch from `pending` only after the `await deps.steer(...)` round trip (`packages/api-server/src/modules/channels/infrastructure/conversation-queue.ts:69`). Any message arriving inside that round trip re-plans the whole leading run, so the earlier message is injected again.
  **Fix:** in both queue paths that read `pending` and mutate it across an await, remove the claimed batch before the awaited call and put it back at the head on refusal.
- 🔴 **Critical:** A steer that completes after its turn ended delivers the message twice — `drain` has already taken it out of `pending` and run it as its own turn (`packages/api-server/src/modules/channels/infrastructure/conversation-queue.ts:74`). `turnEpoch` is captured, but it only reaches `onSteered` as a flag; it never suppresses the delivery.
  **Fix:** treat a turn boundary crossed during the steer as a refusal — compare the captured epoch before the batch is committed, and leave it queued for the next turn.
- 🟡 **Warning:** `docs/architecture/channels.md:184` states the coalescing rule as shared by both messengers and by ambient traffic, but the ambient path still runs its own queue (`slack.ts:2756`): `pending.splice(0)` with no quiet period, no cap and no steering. The issue's read-along case therefore keeps its old gap-dependent behavior. Reported by doc-drift and issue-fit.
  **Fix:** document a rule as shared once every path uses the shared implementation; until then name the exempt path on the page.
- 🟡 **Warning:** A harness that refuses steering is re-asked once per arriving message, each call opening a fresh ACP WebSocket and running `initialize` (`conversation-queue.ts:57`, `core/acp-client.ts:414`). The `unsupported` answer is a capability fact, not a per-message one. Reported by typescript-engineering.
  **Fix:** cache an `unsupported` answer for the running turn and hold the backlog, rather than re-asking per message.
- 🟡 **Warning:** `createAddressedQueue` closes over the first message's `AddressedConversation`, so later messages on the same key run with the first one's `roster`, `ambient`, `ambiguousName` and `teamId` (`slack.ts:2406`). `telegram.ts:474` has the same shape with the first `thread`. Reported by typescript-engineering.
  **Fix:** carry the per-message conversation values on the queued message and read them from the batch at turn time.
- 🟢 **Suggestion:** `batchFiles` is now shared with the addressed queue, but its over-cap log line still reads `slack.ambient_file.over_batch_cap` (`slack.ts:2712`), so a dropped addressed attachment is recorded as an ambient event.
- ✅ **Looks good:** The three reworked `slack-turn.test.ts` scenarios do still exercise concurrent turns — firing from two users defeats the new per-sender top-level key (`slack.ts:2275`), so the id-less-reply protection is genuinely retested rather than silently dropped.

- 🟡 **Warning:** `docs/architecture/channels.md:184` — the rewritten bullets state behavior the branch does not have: ambient traffic is described as coalescing "under the same rule as addressed traffic", while `drainAmbientQueue` (`slack.ts:2756`) has no settle period, no cap and no steering. The same bullet credits the new queue with stopping concurrent first messages from minting duplicate sessions; the guard that does this is the still-present `withSessionTurnLock`, whose description the edit removed.
  **Fix:** an architecture page states only behavior the code on the branch implements; a shared rule is documented as shared once every path is migrated.
- 🟢 **Suggestion:** the size-cap trim also dropped Telegram's recording of the authorizing owner's sub as `authorized_by`, which the code still implements (`telegram-conversations-repository.ts:29`). The binder-or-owner and owner-only-default rules did survive on the page.

### Issue Fit
- 🟡 **Warning:** `slack.ts:2741` — the read-along path keeps its in-flight-only coalescer, so it stays on the gap-dependent behavior issue #3453 names ("Once it has finished … the next message starts a fresh reply regardless of how soon it arrives"). Same finding as the doc bullet above.
  **Fix:** move every path the docs claim for the shared rule onto the shared queue, or name the excluded path as out of scope.

Otherwise the fit is good: the body's reframing of the issue is argued and matches the issue's own revised text, and the chat-UI exclusion is declared with a reason.

- 🔴 **Critical:** `conversation-queue.ts:69` — re-steering across the await (merged into `### Findings` above; reproduced with 5 mid-turn arrivals producing 5 steer calls, message `b` injected 5 times).
- 🔴 **Critical:** `conversation-queue.ts:69` — double delivery when the turn ends mid-steer (merged above; reproduced as turns `a` then `b` while `steer(["b"])` returned `injected`).
- 🟡 **Warning:** `conversation-queue.ts:57` — one refused steer per arrival on an unsupported harness (merged above).
- 🟡 **Warning:** `slack.ts:2406` — stale conversation snapshot per queue (merged above).
- 🟢 **Suggestion:** `slack.ts:2712` — `slack.ambient_file.over_batch_cap` on an addressed turn (merged above).

### Verdict
REQUEST_CHANGES — the shared queue re-injects a message when a second one arrives inside the steer round trip, and delivers it twice when the turn ends mid-steer.

---

pilartomas:
All six addressed in `d8b4ac8a7`. Both criticals were real — thank you, the second one in particular I had convinced myself was handled by the epoch and it wasn't.

**🔴 Re-steering across the await** — fixed as suggested: `steerPending` now claims its batch out of `pending` before the awaited call and restores it at the head on refusal.

**🔴 Double delivery when the turn ends mid-steer** — fixed, with one deliberate difference from the suggested wording. Claiming the batch up front already removes the half where `drain` runs it as its own turn, so what remains is what to do with a steer that comes back `injected` after the boundary. I treat a crossed epoch as a refusal and requeue, as you say. Worth naming the residual: if the adapter really did inject before its turn settled, the line can appear again in the next prompt. That trade is intentional — a repeated line costs the agent a little context, a dropped message costs the user their answer — and the narrow window is bounded by the adapter answering `promptRequired` once its turn has settled. Also fixed alongside: `onEmpty` no longer fires while a steer is in flight, which would otherwise have dropped the queue from its map while a batch was still out and let a later message start a second queue for the same conversation.

Both have regression tests, each checked to discriminate: with the claim-before-await reverted, exactly those two go red and the other eight stay green.

**🟡 Read-along still on its own queue** — you and doc-drift and issue-fit were all pointing at the same thing, and the page was making a claim the branch didn't honour. Read-along now goes through the shared queue, so it gets the quiet period, the cap and one-turn-at-a-time. It is *not* steered, and I've documented that exclusion rather than papering over it: a read-along batch fans out to one turn per reading Agent, so there is no single turn to steer into, and picking one reader to steer would quietly change which agent sees what. The page now names the exclusion and its reason.

**🟡 Page credited the new queue with guarding duplicate sessions** — correct, that is `withSessionTurnLock` and my edit had removed its description. Restored on the session-key bullet.

**🟡 `unsupported` re-asked per message** — agreed, it's a harness fact. Cached for the turn; later arrivals wait instead of opening another socket to be refused. Covered by a test asserting one steer call across three arrivals.

**🟡 Stale conversation snapshot** — fixed on both adapters. `roster`, `ambient`, `ambiguousName`, `teamId` and Telegram's `thread` now ride on the queued message and are read from the batch at turn time; the queue keeps only the key-defining fields.

**🟢 `slack.ambient_file.over_batch_cap`** — now `slack.file.over_batch_cap`, since `batchFiles` serves both paths.

**🟢 doc-drift on `authorized_by`** — not restored. The page is 20 characters from its cap and that is a column name; the space it would take came from cutting exactly that kind of detail. The binder-or-owner and owner-only-default rules that matter for working here did survive, and table shapes are already delegated to the persistence page.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `d8b4ac8`

## PR #3500: fix(channels): a conversation answers a burst of messages once
**Author:** pilartomas | **Branch:** fix/message-coalescing → main | **Changes:** +1461 −174 (21 files)

### Summary
The round closes all six findings from the previous review. Both criticals are gone: I reproduced the fixes on a transcription of the queue, and the claim-before-await plus the epoch-as-refusal each deliver a steered message exactly once. Read-along now runs on the shared queue, so the architecture page's shared rule is true.

### Changes since last review
Previous HEAD: `fbc39f4` (2026-08-28T08:09:22Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** re-steering across the await — the batch is claimed before the call (`conversation-queue.ts:80`)
- ✅ **Fixed:** double delivery when the turn ends mid-steer — a crossed epoch requeues (`conversation-queue.ts:94`)
- ✅ **Fixed:** `unsupported` re-asked per message — cached for the queue (`conversation-queue.ts:66`)
- ✅ **Fixed:** stale conversation snapshot per queue — values ride on the message (`slack.ts:2328`, `telegram.ts:484`)
- ✅ **Fixed:** page claimed a shared rule the ambient path did not use (`docs/architecture/channels.md:184`)
- ✅ **Fixed:** addressed over-cap logged as an ambient event (`slack.ts:2721`)
- 🆕 **New:** a refused or epoch-crossed batch is restored behind newer messages (`conversation-queue.ts:95`)

### Findings
- 🟡 **Warning:** restore-at-head loses its place, because `submit` starts a steer per arrival (`packages/api-server/src/modules/channels/infrastructure/conversation-queue.ts:95`)
  **Fix:** see the inline comment.
- 🟢 **Suggestion:** the Telegram resume path publishes `existing.sessionId` before `sendPrompt`, so a steer arriving during a failed resume targets a dead session (`telegram.ts:425`). The outcome is a refusal and a requeue, so no message is lost. Reported by typescript-engineering.

- 🟢 **Suggestion:** the burst bullet says the queue keys on "the channel and sender on Slack". Read-along keys on the channel alone (`ambientQueueKey`, `slack.ts:2738`), so the sender clause does not hold for the read-along traffic the same sentence folds in. The ambient-routing bullet below states the per-channel scope correctly. (`docs/architecture/channels.md:184`)

### Issue Fit
✅ No findings. Read-along now shares the queue, so the gap-dependent case in #3453 is closed; the steering exclusion is stated with its reason.

- 🟡 **Warning:** `conversation-queue.ts:95` / `:139` — unserialized steers, restore lands out of order (merged into `### Findings` above; reproduced independently — 5 concurrent `deps.steer` calls for a 5-message burst, and delivery order `a, c, b`).
- 🟢 **Suggestion:** `telegram.ts:425` — steer against a session id published before a failed resume (merged above).

### Verdict
COMMENT — every previous finding is fixed and verified. One new 🟡: the restore path can reorder a split thought, which is the behavior this PR exists to protect.

---

packages/api-server/src/modules/channels/infrastructure/conversation-queue.ts
```diff
@@ -0,0 +1,142 @@
...
+    } catch (err) {
+      deps.onError?.(err);
+    } finally {
+      steersInFlight -= 1;
+    }
+
+    if (result === "unsupported") steeringUnsupported = true;
+
+    if (result !== "injected" || turnEpoch !== epoch) {
+      pending = [...batch, ...pending];
```
dam-code-guardian: 🟡 **Warning:** A batch claimed for a steer is restored at the head of `pending`, but the queue keeps advancing while the steer is out, so the restore point is no longer the head. A refusal, or a crossed epoch, therefore lands the batch **behind** messages that arrived after it.

Reproduced on a transcription of this file, each probe with a healthy control:

- One steer in flight, the turn ends, a third message arrives → delivery order `a`, `c`, `b`.
- Two overlapping steers refused in call order → the batch inverts to `c`, `b`.
- Read-along needs no network for this: its `steer` is `async () => "refused"` (`slack.ts:2814`), so three same-tick `void queue.submit(...)` calls (`slack.ts:2841`) alone gave `a`, then `d, c, b`.

Controls that stayed correct: a single steer with nothing arriving behind it, non-overlapping steers, and same-tick arrivals with no turn running.

Nothing is lost or duplicated, so this is not the previous critical returning. It does contradict the invariant the PR states in three places — "so nothing is reordered" in the body, `turn-coalescing.ts:52`, and the page's burst bullet — and the reordering is a split thought reaching the agent out of sequence, which is the case this PR exists for.

**Fix:** A batch claimed out of the queue still owns its place in the order. Hold one delivery in flight per conversation — gate `submit`'s `steerPending()` on the in-flight steer, and block a new turn while `steersInFlight > 0` — so a later arrival joins the next batch and the restore point is still the head when the refusal returns.

packages/api-server/src/modules/channels/infrastructure/telegram.ts
```diff
@@ -400,6 +422,7 @@ export function createTelegramWorker(deps: {
       const existing = await findThreadSession(agentId, thread.id);
       if (existing) {
         try {
+          onSession?.(existing.sessionId);
```
dam-code-guardian, line 425: 🟢 **Suggestion:** `onSession` publishes `existing.sessionId` to the queue before `sendPrompt` runs, so a steer arriving during a failed resume targets a session that is already gone. The outcome is a refusal and a requeue, so no message is lost. Reported by typescript-engineering.

pilartomas:
Fixed in `99e34fe72`.

**🟡 Restore-at-head loses its place** — real, and the reproduction is exactly right. `submit` called `steerPending` per arrival with nothing serialising them, so a burst put several round trips in flight, each on its own connection; restoring a refused batch to the head of a queue the others had already advanced past could then put it behind messages sent after it. Reordering a split thought is precisely the failure this PR exists to prevent, so thank you for catching it.

One steer now runs at a time, and on return it loops to take whatever gathered behind it. That keeps ordering, keeps the batch's place on a refusal, and collapses a burst to one connection rather than one per message. Three regression tests, each checked to discriminate: with the guard removed exactly those three go red and the other nine stay green — including one asserting peak concurrency of 1 across your five-message case, with delivery order `b, c, d, e, f`.

**🟢 Telegram steer against a session id from a failed resume** — accurate, and I've left it, deliberately. Publishing the id before the prompt is what makes steering possible at all during the turn: `sendPrompt` only returns once the turn is over, so waiting for it would mean never steering. The window is narrow and self-correcting — the steer is refused, the batch requeued (no message lost, as you note), and the fresh session republishes the id over the stale one within the same turn. Making it airtight would mean a way to retract a published session id, which is more machinery than the failure mode justifies.

**🟢 Burst bullet's sender clause** — right, and it was my imprecision: read-along keys on the channel alone. The sentence now scopes the sender to addressed turns. Room came from raising field names out of the sequence diagram and dropping the Telegram binding table's name, which the persistence page already owns.

packages/api-server/src/modules/channels/infrastructure/conversation-queue.ts
```diff
@@ -0,0 +1,153 @@
...
+      steering = false;
+    }
+  }
+
+  async function drain(): Promise<void> {
+    if (draining) return;
+    draining = true;
+    try {
+      await settle();
+      while (pending.length > 0) {
```
dam-code-guardian, line 119: 🔴 **Critical:** Serializing the steers closed the concurrent-claimer half of the last round's finding, but the reordering has a second path — this turn boundary. A claimed batch is out of `pending`, so `drain`'s own loop cannot see it: when the turn it targeted ends while the round trip is still outstanding, `drain` composes the next turn from what arrived meanwhile, and the restore at line 102 then lands behind it.

Reproduced on a transcription of the queue at this HEAD, `settleMs: 400`:

```
submit a        -> turn 1 = [a], sessionId = sess-1
submit b        -> claimed, steer(sess-1, [b]) in flight, pending = []
submit c        -> pending = [c]; steerPending returns at the `steering` guard
turn 1 ends     -> epoch 1; drain's while-loop sees pending = [c] -> turn 2 = [c]
steer returns   -> epoch crossed -> pending = [b]  -> turn 3 = [b]
```

Delivery `a, c, b` for messages sent `a, b, c`. It also reproduces with `draining` already false: the turn ends, `drain` exits, a fresh `submit` starts the next turn, and the restore arrives behind it.

**Fix:** Restoring a batch recovers its place among what is *still pending*; it cannot recover its place against what already went out. So a claimed batch has to stop the queue advancing past it — `drain` must not select a `start` batch while one is unresolved (`steersInFlight > 0`, or a live steer cycle), and must wait for it to be injected or restored first. The rule holds for any queue that hands a claimed batch to a fallible transport and restores it on failure: the claim window must block the ordering-sensitive consumer, not only concurrent claimers.

packages/api-server/src/modules/channels/infrastructure/conversation-queue.ts
```diff
@@ -0,0 +1,153 @@
...
+      settleRoundsRemaining(rounds, pending.length)
+    ) {
+      seen = pending.length;
+      rounds += 1;
+      await delay(deps.settleMs);
+    }
+  }
+
+  async function steerPending(): Promise<void> {
+    if (steering) return;
```
dam-code-guardian: 🟡 **Warning:** `steering` is a single latch over a resource whose identity changes each turn, so it spans turns. A round trip that outlives the turn it targeted keeps the latch closed, and the live successor turn cannot be steered into — arrivals fall through to their own later turn, which is the extra turn the queue exists to fold away. Before this round those messages did reach the harness on the same interleaving.

Reproduced at this HEAD: a `sess-1`-targeted steer hangs, turn 1 ends, turn 2 runs on `sess-2`, and a message arriving during turn 2 produces no `steer(sess-2, …)` call at all — it returns at line 70 and becomes turn 3.

**Fix:** Scope a latch to what it protects. This one enforces one-steer-at-a-time *within a turn*, so clear it when `turnEpoch` advances (or make the cycle per-turn), and a new turn always starts steerable. Generally: a latch guarding a resource whose identity can change must be invalidated when that identity changes, or its exclusion leaks onto the successor.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `99e34fe`

## PR #3500: fix(channels): a conversation answers a burst of messages once
**Author:** pilartomas | **Branch:** fix/message-coalescing → main | **Changes:** +1541 −180 (21 files)

### Summary
Serializing the steers removes the concurrency half of the previous round's ordering finding: one round trip runs at a time, and a burst now costs one connection. The ordering violation itself survives, on a second path — the turn boundary rather than a concurrent claimer — so the reordering the new block comment rules out still reproduces.

### Changes since last review
Previous HEAD: `d8b4ac8` (2026-08-28T09:16:01Z) — verdict COMMENT

- ✅ **Fixed:** a burst opened one steer per arrival — one runs at a time now, peak concurrency 1 (`conversation-queue.ts:70`)
- ✅ **Fixed:** the burst bullet's sender clause did not hold for read-along keying (`docs/architecture/channels.md:184`)
- 🆕 **New:** a claimed batch is restored behind a turn that started while it was outstanding (`conversation-queue.ts:119`)
- 🆕 **New:** `steering` outlives its turn and suppresses steering into the next one (`conversation-queue.ts:70`)

The Telegram resume-order suggestion is recorded as accepted — the refusal-and-requeue path is self-correcting within the turn, as your comment sets out. It will not come back.

### Findings
- 🔴 **Critical:** `drain` composes a new turn while a claimed batch is still out, so the restore lands behind messages sent after it (`packages/api-server/src/modules/channels/infrastructure/conversation-queue.ts:119`)
  **Fix:** see the inline comment.
- 🟡 **Warning:** `steering` is not scoped to a turn, so a round trip that outlives its turn blocks steering into the live successor (`packages/api-server/src/modules/channels/infrastructure/conversation-queue.ts:70`)
  **Fix:** see the inline comment.

- 🟢 **Suggestion:** the de-naming reached the Overview and the Telegram prose, but the Topology paragraph still carries `ChannelManager` and `telegram_conversations` literally, so the page now has two names for each. Folding those two mentions the same way also returns a little cap budget. (`docs/architecture/channels.md:48`)

### Issue Fit
✅ No findings. The delta stays inside #3453's ask, and the page now states the real keying.

- 🔴 **Critical:** `conversation-queue.ts:101` — the claimed batch is restored behind a turn that already started (merged into `### Findings` above; reproduced independently as `b1 → b3 → b2`, and a second shape as `y1 → y3 → y2` with `draining` already false).
- 🟡 **Warning:** `conversation-queue.ts:70` — the `steering` latch leaks onto the successor turn (merged above; reproduced — a message arriving in the live turn 2 was never steered and became its own turn).
- 🟢 **Suggestion:** `conversation-queue.ts:150` — the promise `submit` returns resolves at a different point per path: the message that owns the steer cycle waits for the whole loop, while one arriving during it returns at the `steering` guard. `handleAppMention` awaits `enqueueAddressed`, so the ingress back-pressure differs across a burst.

### Verdict
REQUEST_CHANGES — a claimed batch can still reach the agent after messages that were sent later, which is the ordering this queue exists to hold.

---

pilartomas:
Fixed in `c08eeec69`.

**🔴 Turn composed while a batch is still claimed** — correct, and I under-fixed last round. Serialising the steers closed the concurrent-claimer path, but the same ordering violation survived across the turn boundary: a claimed batch goes out, the turn ends under it, `drain` composes the next turn from what arrived *after* it, and the restore lands behind them. `b1 → b3 → b2`, exactly as you reproduced.

The steer cycle is now a promise the queue holds, and `drain` waits on it before composing a turn — so a claimed batch is always back in the queue before the next turn is planned, and keeps its place. Regression test checked to discriminate: without the wait, only it goes red.

**🟡 `steering` latch outliving its turn** — retired rather than re-scoped. The latch is gone; the cycle promise clears itself, and since no turn is composed while it is outstanding, a round trip can no longer overlap the successor turn at all. There is a test asserting steering still works on the turn after a slow one — it does not discriminate against the previous code (the latch cleared on cycle end there too; the leak needed the overlap that the critical fix removes), but it pins the behaviour going forward.

**🟢 `submit` resolving at different points** — taken. It returns the same cycle promise on every path now, so a burst applies uniform back-pressure to `handleAppMention` instead of resolving early for the messages that did not start the cycle.

**🟢 Two names for each in Topology** — my de-naming was half-done. Folded in the paragraph and in both diagrams, so the literals are gone from the page; that returned about 30 characters of cap budget.

One thing worth flagging for the human reviewer rather than hiding in a diff: `channels.md` now sits ~36 characters under its 40 000 cap. Anything further added there needs real content removed first — I have already taken out the code-level detail that was straightforwardly justifiable, so the next edit is a judgement call about what the page should stop covering.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `c08eeec`

### Summary
The steer cycle is now a promise the queue holds, and `drain` waits on it before it composes a turn. That closes the turn-boundary half of the ordering defect and retires the `steering` latch with it. Every path I could construct holds order.

### Changes since last review
Previous HEAD: `99e34fe` (2026-08-28T10:08:17Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** a claimed batch was restored behind a turn that started while it was outstanding (`conversation-queue.ts:119`)
- ✅ **Fixed:** the `steering` latch outlived its turn and blocked steering into the successor (`conversation-queue.ts:70`)
- ✅ **Fixed:** `submit` resolved at a different point per path (`conversation-queue.ts:150`)
- ✅ **Fixed:** the Topology paragraph and both diagrams kept the de-named literals (`docs/architecture/channels.md:48`)

Both blocking findings are gone, and the fix is the stronger of the two shapes available: `drain` waits for the cycle instead of the cycle guarding itself. I ran the module as a runnable copy and checked the paths the change opens. `b1 → b3 → b2` now delivers `b1, b2, b3`. A five-message burst with a refusal delivers `m1` then `m2…m6` in order, on one steer connection. An attachment at the head is not steered and is not overtaken. A throwing turn does not wedge the queue. `onEmpty` fires once, after the cycle clears, on each of the four turn-versus-steer orderings — so the queue cannot drop from its map while a batch is still out.

Your note on the `channels.md` cap is the right call to surface. The page measures 39 964 characters against the 40 000 cap, so 36 characters remain and the next edit there does need something removed first.

### Findings

- 🟢 **Suggestion:** the reply-target rule — each batched message registers its own target, and a reply naming none is refused — sits with ambient routing, while addressed top-level batches now follow it too (`docs/architecture/channels.md:187`). Moving the sentence into the burst bullet and cross-referencing it keeps the character count flat.
- 🟢 **Suggestion:** the withheld-attachment promise of a sender notice does not hold for the per-batch size ceiling on an addressed turn (`docs/architecture/channels.md:176`). The line is unchanged from `main`, so this is pre-existing and reads better as its own issue than as work on this PR.

The skill's third finding did not survive verification. It reads the burst bullet as promising sender-narrowed keying on Telegram; the narrowing clause attaches to the Slack channel case, and `addressedQueueKey` keys Slack top-level traffic on `top:<slackUserId>` while `enqueueTelegramTurn` keys on agent and chat, as the sentence says.

### Issue Fit
✅ No findings. The delta stays inside #3453's ask.

✅ No findings. The skill executed the queue and its domain module as a runnable copy, reproduced the repo's own ordering scenarios as a positive control, then ran randomized interleavings — 300 seeds with no lost, duplicated, or reordered message, and 250 seeds with no premature or missing `onEmpty`. It did not finish a sweep over `unsupported` results, `canSteer` blocks, throwing turns, and a live settle period; I covered those four by hand with the results above.

### Verdict
APPROVE — both blocking findings from the previous round are fixed, the ordering guarantee holds on every path I exercised, and nothing new blocks.

---

Issue #3431, opened 2026-08-24
An agent silently misses configuration its runtime image is too old to accept


pilartomas:
## Context

Agents run a pinned runtime image, and older images don't understand every kind of configuration the platform now sends. The platform is deliberately tolerant of this: it drops the parts an agent can't accept and delivers the rest. That tolerance is the right default, but it is currently silent — nothing tells the owner it happened.

This came out of investigating a control-plane logging flood (#3413). The flood turned out to be a different problem, fixed in #3414. What remained is this.

## Problem

A user grants a connection to an agent — credentials, network access — and the platform reports success. If the agent's image is too old to accept one of those parts, that part is discarded on every delivery. The agent runs without the credentials or the network permissions it was granted.

From the user's side everything looks correct: the grant is listed, no error appears, the agent looks healthy. The agent then fails at runtime in a way that points nowhere near the cause — a missing credential, a blocked host. Updating the image fixes it, but nothing suggests that is the remedy.

There is already a warning at grant time when the platform knows in advance that the agent can't support part of a grant. It doesn't cover the gap appearing later, when an agent predates the kind of configuration being sent and nobody revisits the grant.

## Goal

An owner can tell, without reading server logs, that their agent isn't receiving configuration it was granted, and knows that updating the agent's image is what fixes it.

## User Stories

- As a user, I want to be told when my agent can't accept configuration I've granted it so that I don't debug a failure whose cause is invisible.
- As a user, I want to know which parts of a grant actually reached my agent so that I can trust what the grant list shows me.
- As a user, I want to know that updating my agent's image is the fix so that I can resolve it myself.

## Scope

Covers telling the owner about the gap and pointing at the remedy. Does not cover updating agent images automatically.

## Open Questions

- Which kinds of configuration are serious enough to warrant this? Missing credentials or network permissions arguably leave an agent misconfigured; a missing optional extra may deserve only a quiet note.
- Should a serious gap mark the agent unhealthy and stop it taking work, or leave it running with a prominent warning? Stopping is safer, but turns a degraded agent into an unavailable one.
- Is this surfaced on the agent, on the grant, or both?

## Additional resources

- #3413 — the investigation this came from. Its second bucket described this as an endless retry loop; that framing was wrong, and delivery does settle.
- #3414 — fixes the unrelated logging flood found in the same investigation.

PR #3501, opened 2026-08-28
fix: old agent runtimes fall back to polling, with an update action

jezekra1:
#3450 assumed the pod serves its tRPC router over WebSocket. Images built before it don't — so an old agent's file and session panels presented as permanently unreachable while its chat worked fine, with nothing telling the user why.

This is a backward-compatibility layer, not just a notice. Old agents keep working through the paths that never left, and both panels say what's happening:

| surface | new runtime | old runtime |
|---|---|---|
| files tree / open file | WS + watch, no polls | HTTP tRPC proxy, 5s poll |
| session list | pod tRPC + watch | passive ACP read, 5s poll |
| Home feed rows | notices | 15s poll |
| owner-wide holder | pod stream | skipped — no futile dials |
| panels | — | one-line notice; carries the existing **Update** action when the agent has a pending template update |

**Detection** follows the `harnessConfig` precedent: new runtimes advertise a `liveUpdates` capability flag on `hello`, stored in the existing capabilities json (no migration) and carried through the contributions status onto the agent view. Absence means an old image — no probes, no version comparisons.

**The flag fails closed to compat mode, which is safe in both directions**: the HTTP proxy and the ACP intercept serve new pods too, so a missed or late `hello` degrades an agent to polling, never to broken.

The compat polls retire themselves: they exist only for agents whose stored capabilities lack the flag, which ends when the last old image is updated.

Refs #3307, #3431 — this is the live-updates instance of that issue's silent-degradation class: the same detection (a capability flag on hello) now drives a visible notice with the update action, and could carry the dropped-configuration cases the issue describes.


packages/ui/src/modules/sessions/api/queries.ts
```diff
@@ -91,12 +95,15 @@ export function useAcpSessions(
...
 ) {
+  const compat = useAgentLacksLiveUpdates(agentId);
   const live = !!agentId && (options?.enabled ?? true);
   return useQuery({
     queryKey: acpSessionsKeys.list(agentId, include),
     queryFn: live
       ? async () => {
-          const sessions = await listAgentSessions(agentId);
+          const sessions = compat
+            ? await listAgentSessionsOverAcp(agentId)
```
dam-code-guardian: 🔴 **Critical:** This switch is the only place `listAgentSessions` is chosen against `compat`. `listAgentSessions` itself (`acp-session-ops.ts:118`) calls `agentTrpc(agentId).sessions.list.query()` — the pod tRPC surface that arrived with #3450 — and four other callers never consult `compat`: `schedules/api/queries.ts:47`, `experiments/components/experiment-dock-panel.tsx:36`, and the resume-failure classifier and mode reconciliation at `sessions/hooks/use-acp-session.ts:30,122`.

On an old runtime the `/api/trpc-ws` upgrade hits the pod's `else socket.destroy()`, so `agent-trpc.ts`'s `isAbnormalClose` handler calls `markAgentUnreachable`. `useIsAgentOperable` reads that latch, and every compat query is gated `enabled: !!agentId && operable` — so one unmigrated call disables the fallback this PR adds. `use-agent-reachability.ts:14` clears the latch only when the agent is *not* running, so it persists for the tab's lifetime.

**Fix:** Every reader of a capability-gated primitive goes through the same switch — put the `agentLacksLiveUpdates` check inside `listAgentSessions` (delegating to `listAgentSessionsOverAcp`) rather than at one call site, so no caller can reach a transport its runtime lacks.

packages/ui/src/modules/agents/components/runtime-outdated-notice.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const agents = useAgentsList();
+  const { updateOne, updatingId, updatingAll } = useUpdateSandbox();
+  const agent = agentId ? agents.find((a) => a.id === agentId) : undefined;
+  const update = agent?.templateUpdate ?? null;
+
+  return (
+    <div className="flex items-center gap-2 border-b border-border/50 px-4 py-2 text-xs text-muted-foreground">
+      <span className="flex-1">
+        Live updates need a newer agent runtime — showing polled data.
+        {update ? "" : " Update this agent to get instant updates."}
```
dam-code-guardian: 🟡 **Warning:** This reads inverted. `update` truthy means a template update is pending and the `UpdateAvailableAction` button *is* rendered below — and that is exactly when the sentence is suppressed. When `update` is null no button renders, and the user is told to "Update this agent" with no control to do it.

**Fix:** Tie call-to-action copy to the same condition that renders its control — emit the sentence when `update` is truthy (or drop it as redundant with the button), and show a no-action-available variant otherwise.

packages/api-server/src/modules/runtime-delivery/infrastructure/outbox-repo.ts
```diff
@@ -95,6 +102,25 @@ export function createOutboxRepo(db: Db): OutboxRepo {
...
+      if (agentIds.length === 0) return result;
+      const rows = await db
+        .select({
+          id: agentsTable.id,
+          caps: agentsTable.runtimeCapabilities,
+        })
+        .from(agentsTable)
+        .where(inArray(agentsTable.id, agentIds));
+      for (const row of rows) {
+        const parsed = capabilitiesSchema.safeParse(row.caps);
```
dam-code-guardian: 🟡 **Warning:** This strict-parses the whole `capabilities` object to read one boolean, so an unrelated validation failure decides a capability. Verified against the pinned zod 4.4.3: a runtime advertising an `events` member the api-server's enum does not know fails `safeParse` wholesale → `null` → `liveUpdates: false`. A *newer* agent is then shown the outdated-runtime notice and put on polling. Unknown top-level keys are stripped safely; only the enum arrays carry this.

**Fix:** Feature detection reads only the field it needs — parse `liveUpdates` with a narrow schema over the stored json, so a neighbouring field cannot decide it.

packages/api-server/src/modules/live-events/compose.ts
```diff
@@ -31,16 +32,22 @@ export function composeLiveEventsModule(deps: {
...
     log: deps.log,
-    listRunningAgentIds: async (ownerSub) =>
-      (await deps.agentsRepo.list(ownerSub))
+    listRunningAgentIds: async (ownerSub) => {
+      const running = (await deps.agentsRepo.list(ownerSub))
         .filter(agentStreamable)
-        .map((agent) => agent.id),
+        .map((agent) => agent.id);
+      const features = await deps.runtimeFeaturesFor(running);
+      return running.filter((id) => features.get(id)?.liveUpdates);
```
dam-code-guardian, line 49: 🟡 **Warning:** This filter reads capabilities that are written only at `hello`, and `hello-handler.ts` publishes no live event. `reconcile` runs on the `agents`/`sync` topics, and the K8s agent watch publishes on a resource-fingerprint change, which a database-only capability write does not move. An agent that starts and says hello after the subscription is established stays out of the watch set until some unrelated agents-topic event fires.

**Fix:** When a stored read becomes a membership filter, the write that establishes it triggers re-evaluation — publish an agents-topic live event from the hello handler.

packages/ui/src/modules/files/components/files-panel.tsx
```diff
@@ -117,6 +118,9 @@ export function FilesPanel({
                 onCancel={controller.handleCancelNew}
               />
             )}
+            {controller.runtimeOutdated && (
```
dam-code-guardian: 🟡 **Warning:** `chat-view.tsx:549,574` mounts `SessionsSidebar` and `FilesPanel` in the same left column simultaneously, and `SidebarSection` renders its children even when collapsed. A compat agent therefore gets two identical notices and two Update buttons. Both are also inside the panels' `overflow-y-auto` container, so the notice and its action scroll out of view.

**Fix:** An agent-scoped status notice renders once per agent, not once per panel — hoist it to the container that owns the panels, and place it as a sibling above the scrolling region.

docs/architecture/platform-topology.md
```diff
@@ -73,7 +73,7 @@ A per-agent Envoy pod paired with the agent-runtime pod. Mounts the owner's cred
...
...he api-server. It uses tRPC over a single authenticated WebSocket for resource management, permission flows, and live updates — server-pushed invalidation hints replace list polling, and pod-sourced state follows the pod the same way, so the remaining polls are runtime metrics (a telemetry-store aggregation, not pod-sourced), the pod's background-work report (30 s, only while the agent is awake), skill-install state, Home's approval reads, and a compatibility poll for agents whose runtime predates live updates — those fall back to the HTTP proxy and the passive ACP list, flagged in the panels with an update action, until their image is updated — and ACP WebSockets for bidirectional agent communication. A tab holds several ACP channels per Agent at once: the live channel for the session on screen, short-lived ones for history replay and Home's one-shot session-list reads, and — while a brand-new session's first prompt is on its way — a channel of its own, which either becomes the live one or carries that turn to completion after the user has moved on. Permission prompts, tool calls, and streaming output all flow over the live ACP connection. See [`packages/ui/`](../../packages/ui/).
```
dam-code-guardian, line 76: 🟡 **Warning:** This page was edited here but its `Last verified:` header (line 3) still reads 2026-08-27, and `runtime-delivery.md:3` still reads 2026-08-24 — `docs/guidelines/documentation-guidelines.md:34` requires the bump on any edit. This paragraph also still says a Home tab makes the api-server watch the session lists of *every* running agent, which `live-events/compose.ts:49` now narrows to capable runtimes. Separately, `runtime-delivery.md:102` transcribes the literal field names `harnessConfig` and `liveUpdates`, which line 41 of the same guidelines excludes from architecture pages.

**Fix:** One doc pass per PR — bump `Last verified:` on every edited page, restate a scope the code narrowed, and name capabilities in domain vocabulary rather than by field.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `7873114`

## PR #3501: fix: old agent runtimes fall back to polling, with an update action
**Author:** jezekra1 | **Branch:** fix/old-runtime-upgrade-notice → main | **Changes:** +250 −34 (35 files)

### Summary
Adds a `liveUpdates` capability flag on `hello` and routes agents that lack it to an HTTP tRPC proxy plus interval polling, with a notice in the files and sessions panels. The compat path is complete for files and the session sidebar; the findings below are the call sites and signals it does not reach.

### Findings
- 🔴 **Critical:** `listAgentSessions` reaches the WS transport on old runtimes, latching the agent `unreachable` (`packages/ui/src/modules/sessions/api/queries.ts:105`)
- 🟡 **Warning:** Two `assembleAgent` call sites omit `features` and so report `liveUpdates: false`: `connectSlack` has `status` in scope and passes its other two fields, and `create` passes no status at all. Both return an `AgentView` the UI renders with the outdated-runtime notice. (`packages/api-server/src/modules/agents/services/agents-service.ts:601`, `:828`)
  **Fix:** A capability field is not an optional parameter with a degraded default — make `features` required on `assembleAgent` so the compiler names every construction site.
- 🟡 **Warning:** The notice's call-to-action is inverted (`packages/ui/src/modules/agents/components/runtime-outdated-notice.tsx:15`)
- 🟡 **Warning:** A whole-object strict parse decides one capability, so a *newer* runtime can be downgraded to polling (`packages/api-server/src/modules/runtime-delivery/infrastructure/outbox-repo.ts:118`)
- 🟡 **Warning:** Nothing re-reconciles the owner pod-session watch when capabilities become known (`packages/api-server/src/modules/live-events/compose.ts:49`)
- 🟡 **Warning:** The notice renders once per panel, not once per agent — two notices, two Update buttons (`packages/ui/src/modules/files/components/files-panel.tsx:121`, `packages/ui/src/modules/sessions/components/sessions-sidebar.tsx:232`)
- 🟡 **Warning:** Both edited architecture pages miss their mandatory `Last verified:` bump; one keeps a scope the code narrowed, one transcribes literal field names (`docs/architecture/platform-topology.md:76`)
- 🟢 **Suggestion:** `agentTrpcHttp` reproduces none of the WS client's reachability wiring, so an unreachable old-runtime agent still reads as operable; both client maps also grow per agent id with no eviction (`packages/ui/src/modules/agents/agent-trpc.ts:47`). Reported by typescript-engineering.
- 🟢 **Suggestion:** The PR refs #3431 but delivers only the live-updates case. The body declares the gap and there is no closing keyword, so the issue stays open. Reported by issue-fit.

- 🟡 The pod-watch scope is now capability-gated in code, but the page still says a Home tab makes the api-server watch the session lists of *every* running agent its owner has. `docs/architecture/platform-topology.md:76`
  **Fix:** When code narrows the set a subsystem acts on, restate the scope on its architecture page in the same PR.
- 🟡 Edited in this PR, but `Last verified:` still reads 2026-08-27. `docs/architecture/platform-topology.md:3`
- 🟡 Edited in this PR, but `Last verified:` still reads 2026-08-24. `docs/architecture/runtime-delivery.md:3`
  **Fix:** Any edit to an architecture page bumps its `Last verified:` header — required by `docs/guidelines/documentation-guidelines.md:34`.
- 🟡 The register-call description transcribes literal capability field names (`harnessConfig`, `liveUpdates`), which the content policy excludes from architecture pages. `docs/architecture/runtime-delivery.md:102`
  **Fix:** Describe protocol payloads semantically — say the agent reports which optional surfaces it supports, and link the contract package as the field-level source of truth.

### Issue Fit
- 🟢 The PR references #3431 ("An agent silently misses configuration its runtime image is too old to accept") but delivers none of its ask — that issue is about dropped *granted configuration*, while this PR surfaces only the live-updates degradation. Harmless: the body declares the gap and there is no closing keyword, so #3431 stays open.

- 🟡 `runtimeCapabilitiesMany` strict-parses the whole `capabilities` object to extract one boolean, so any unrelated validation failure silently downgrades a modern agent to the compat path. A runtime advertising an event kind the api-server's enum does not yet know makes `safeParse` fail wholesale → `null` → polling and a "runtime outdated" notice for an up-to-date agent. `packages/api-server/src/modules/runtime-delivery/infrastructure/outbox-repo.ts:116`
  **Fix:** Feature detection must parse only the field it needs and tolerate unknown neighbours.
- 🟡 Two `assembleAgent` call sites omit the new `features` argument and therefore report `liveUpdates: false`. `packages/api-server/src/modules/agents/services/agents-service.ts:601`, `:828`
  **Fix:** Make `features` required on `assembleAgent` so the compiler flags every construction site.
- 🟡 Nothing re-reconciles the owner pod-session watch when an agent's capabilities become known. Capabilities are written only at `hello`, and that handler publishes no live event, while reconcile runs on `agents`/`sync` topics only. `packages/api-server/src/modules/live-events/compose.ts:49`
  **Fix:** Publish an agents-topic live event from the hello handler so newly-known-capable agents get watched.
- 🟡 The compat HTTP client carries no reachability signal and is never evicted. `packages/ui/src/modules/agents/agent-trpc.ts:47`
  **Fix:** A fallback transport must reproduce the signals the primary transport provided, and per-agent clients should be evicted when the agent leaves the list.

_Four further findings from this skill were checked against the tree and are not reported: its 🔴 initial-load race, and the query-key, dual-source-of-truth, and `features?.` rollout warnings. All rest on `compat` being readable before capabilities are known. It is not — `useAgentLacksLiveUpdates`, `useIsAgentOperable` and the module-level `agentLacksLiveUpdates` read the **same** query (`agentsKeys.listWithChannels()`), so while it is pending `runState` is `undefined`, `operable` is false, and no query or subscription runs at all; when it resolves, `state` and `features` arrive in one object. `AgentView` is served by the api-server build that defines it, and the query cache is not persisted._

- 🔴 The guidance sentence is inverted: `{update ? "" : " Update this agent to get instant updates."}` shows "Update this agent…" only when `templateUpdate` is null — exactly when no Update button is rendered and the user has no way to act — and suppresses it when the actionable button is present. `packages/ui/src/modules/agents/components/runtime-outdated-notice.tsx:14`
  **Fix:** Tie call-to-action copy to the same condition that renders its control.
- 🟡 Both `SessionsSidebar` and `FilesPanel` render in the same left column at the same time, so an outdated runtime produces two identical notices — and two Update buttons — stacked in one sidebar. `packages/ui/src/modules/sessions/components/sessions-sidebar.tsx:232`, `packages/ui/src/modules/files/components/files-panel.tsx:121`
  **Fix:** An agent-scoped status notice must render once per agent, not once per panel.
- 🟡 The notice carries no `role="status"`, unlike the repo's equivalent notices (`stale-model-notice.tsx:17`, `skill-drift-banner.tsx:29`), so screen-reader users are never told the panel switched to polled data. `packages/ui/src/modules/agents/components/runtime-outdated-notice.tsx:11`
  **Fix:** Give dynamically inserted status regions `role="status"`.
- 🟡 In both panels the notice sits inside the `overflow-y-auto` container above the list, so it and its Update action scroll out of view. `packages/ui/src/modules/files/components/files-panel.tsx:121`
  **Fix:** Render persistent status banners as a sibling above the scrolling region, or sticky within it.

### Verdict
REQUEST_CHANGES — the compat path is sound where it was wired, but `listAgentSessions` reaches the WS transport on old runtimes and latches the agent `unreachable`, which disables the fallback itself.

---

packages/ui/src/modules/sessions/api/queries.ts
jezekra1: @-

packages/ui/src/modules/agents/components/runtime-outdated-notice.tsx
jezekra1: @-

packages/api-server/src/modules/runtime-delivery/infrastructure/outbox-repo.ts
jezekra1: @-

packages/api-server/src/modules/live-events/compose.ts
jezekra1: @-

packages/ui/src/modules/files/components/files-panel.tsx
jezekra1: @-

docs/architecture/platform-topology.md
jezekra1: @-

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `0e7672d`

## PR #3501: fix: old agent runtimes fall back to polling, with an update action
**Author:** jezekra1 | **Branch:** fix/old-runtime-upgrade-notice → main | **Changes:** +341 −39 (39 files)

### Summary
Complete re-review at the live HEAD. Both commits since `7873114` land: the fallback is now unbypassable at the `listAgentSessions` boundary, detection tolerates unknown neighbouring capability members, and `features` is a required parameter at every assembly site. All seven findings from the previous review are resolved.

### Changes since last review
Previous HEAD: `7873114` (2026-08-28T12:08:26Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** the WS dial on old runtimes — the `compat` switch moved into `listAgentSessions` itself, so all five callers inherit it (`packages/ui/src/modules/sessions/api/acp-session-ops.ts:119`)
- ✅ **Fixed:** `features` is now a required positional parameter, and all four `assembleAgent` sites pass it (`packages/api-server/src/modules/agents/infrastructure/agent-mappers.ts:204`)
- ✅ **Fixed:** the notice's call-to-action now matches the condition that renders the button (`packages/ui/src/modules/agents/components/runtime-outdated-notice.tsx:18`)
- ✅ **Fixed:** `runtimeFeaturesOf` parses one field over `z.looseObject` with `.catch(undefined)`, so an unknown neighbour no longer downgrades a newer runtime (`packages/agent-runtime-api/src/modules/runtime/types.ts:254`)
- ✅ **Fixed:** the hello handler emits `RuntimeHelloReceived`, projected to an `agents` hint, which drives `reconcile` (`packages/api-server/src/modules/runtime-delivery/services/hello-handler.ts:32`)
- ✅ **Fixed:** the notice renders once per agent, hoisted to the column that owns both panels, above the scrolling sections, with `role="status"` (`packages/ui/src/modules/sessions/views/chat-view.tsx:552`)
- ✅ **Fixed:** both architecture pages carry `Last verified: 2026-08-28`, the narrowed watch scope is restated, and the literal field names are gone (`docs/architecture/platform-topology.md:3`)
- 🔁 **Still present:** `agentTrpcHttp` carries no reachability signal and its client map is never evicted (`packages/ui/src/modules/agents/agent-trpc.ts:47`)

### Findings
- 🟢 **Suggestion:** Both capability readers return `false` for an agent id absent from the agents-list cache, so an unknown capability reads as "has live updates". The hook path is safe — its consumers gate on `useIsAgentOperable`, which is `false` while the same query is pending — but `useScheduleSessions` gates only on `agentId && scheduleId`, so its `listAgentSessions` call can dial the WS surface on an old runtime before the list resolves. The server-side contract asserted in `runtime-features.test.ts` is the opposite direction. (`packages/ui/src/modules/agents/api/queries.ts:40`, `:48`)
- 🟢 **Suggestion:** The notice does not follow the flex contract its two sibling notices use. Its root has no `shrink-0` in a column whose other children pin their basis via `sectionFlex`, and its text span is `flex-1` without `min-w-0`; the left column clamps to 140px. Both peers pair `min-w-0 flex-1` with `shrink-0` to keep the trailing action from being clipped (`stale-model-notice.tsx:21`, `skill-drift-banner.tsx:33`). (`packages/ui/src/modules/agents/components/runtime-outdated-notice.tsx:12`, `:16`)
- 🟢 **Suggestion:** `agentTrpcHttp` still reports no reachability, so `useIsAgentOperable` never learns a compat agent went away and the 5s poll keeps hitting a dead pod; `httpClients` also grows per agent id with no eviction. Carried over from the previous review, on code these commits did not touch. (`packages/ui/src/modules/agents/agent-trpc.ts:47`)
- 🟢 **Suggestion:** The `RuntimeHelloReceived` emit is skipped without a trace when owner resolution returns `null`, and `bootstrap.ts:303` builds `resolveAgentOwner` with `.catch(() => null)`, so a transient repository error looks like an ownerless agent. The handler already logs its snapshot-write failure; the same treatment here would make a missing hint attributable. (`packages/api-server/src/modules/runtime-delivery/services/hello-handler.ts:31`)
- 🟢 **Suggestion:** `runtimeFeaturesMany` sits on `OutboxRepo` but queries the `agents` table, which the sibling `AgentsRuntimeRepo` in the same file owns. (`packages/api-server/src/modules/runtime-delivery/infrastructure/outbox-repo.ts:51`)
- 🟢 **Suggestion:** `AgentView` restates the capability shape inline instead of using the exported `RuntimeFeatures`, which this file's package already imports from. (`packages/ui/src/types.ts:117`)

- 🟢 Three statements on the edited topology page are still absolute where the PR made the path conditional: the HTTP-proxy row says "No UI caller remains" while `agentTrpcHttp` now dials it from the browser; the session list is described as "no longer an ACP read", which `listAgentSessionsOverAcp` contradicts in compat; and the pod tRPC surface is credited with "the watches behind all three live panels", all three of which compat disables. The paragraph at line 52 got its caveat — these three did not. (`docs/architecture/platform-topology.md:55`, `:95`, `:96`)

### Issue Fit
- 🟢 The PR delivers the live-updates instance of #3431's silent-degradation class, not the dropped grant configuration the issue describes. Unchanged from the previous review, and non-blocking: the body declares the gap and there is no closing keyword, so #3431 stays open.

- 🔁 4 findings from the previous review: 3 fixed (strict capability parse, optional `features`, no watch re-reconciliation), 1 still present (see review at `7873114`). New findings this round are in `### Findings` above — the cache-miss fallback direction, the hello-emit diagnostic, the repository boundary, and the duplicated contract type.

- 🔁 All 4 findings from the previous review are fixed: the inverted call-to-action, the duplicated per-panel notice, the missing `role="status"`, and the notice scrolling out of view. One new layout finding is in `### Findings` above.

### Verdict
APPROVE — every finding from the previous round is resolved, including the critical WS dial, and nothing new blocks. The remaining items are non-blocking.

---

PR #3498, opened 2026-08-27
perf(platform-base,ci): resolve tools once per pod, and make layer digests content-only

jezekra1:
## What

Toward #3294 — deliberately not a closing keyword; see **Why #3294 stays open** below. The entrypoint resolves mise's tools onto `PATH` once per pod, so a tool call no longer pays mise's startup.

## Why

`jq` on the pod was not jq — it was a symlink to a 91MB mise binary that re-resolved its toolchain on every invocation. Measured on the pod in #3294: **282ms per call against 17ms** for the real binary, 438ms for `gh`.

It is not only the tools the issue named. `node` and `python3` were on shims too — 21ms against 12ms, and 14ms against 6ms measured here — and the harness calls both constantly.

## What mise actually recommends, since it matters here

Their container guidance is `mise exec` or shims-on-PATH, and their shims page says outright that the performance difference is *"probably not going to be noticeable"*. That is not wrong; it assumes you are comparing per-call cost against `mise activate`'s per-shell-prompt cost. A pod has no prompt, and the harness invokes tools programmatically in loops, so the cost is paid per call and never amortises. Their guidance targets short-lived CI jobs where paying 250ms a handful of times is irrelevant.

I checked for a supported fast path: 148 settings, and the only shim-related one is `windows_shim_mode`. The caching settings govern activation's env resolution, not shim startup. There is no fast-shim mode.

So resolving paths onto `PATH` is the only mechanism that removes the cost — and it is native (`mise env`), just not the variant their container page suggests.

## How

- **Boot:** `eval "$(mise env)"` in the entrypoint. Covers every installed tool, in derived images too, with no per-image step. **36ms, once.**
- **Eager:** jq, fd, rg and **gh** are installed at build — for egress rather than speed, since a lazy first use has to reach the network through the gateway; the same argument the Dockerfile already makes for python. gh moved to this tier during review: it is the tool agents reach for most, the platform already provisions its credentials into the pod, and mise downloads gh *from GitHub*, so a lazy first use had to reach GitHub through the egress gateway to obtain it. 11MB for jq, fd and rg; 12.5MB gzipped for gh.
- **Lazy:** kubectl, oc, uv, uvx and gws still install on demand. Their wrapper links the resolved binary into `tool-bin` and hands off, so only the first call pays.

## Verified on the built agent image

| | before | after |
|---|---|---|
| boot resolution | — | 36 ms once |
| jq | 12 ms/call | **1 ms** |
| python3 | 14 ms/call | **6 ms** |
[...]

PR #3514, opened 2026-08-31
perf(platform-base): link tools into tool-bin from a mise postinstall hook

JanPokorny:
Follow-up to #3498, which put the real tool binaries on PATH via `tool-bin` but populated it by hand — an `eval "$(mise env)"` at pod boot, plus an `ln -sf` inside each lazy-tool shim. mise can do this itself, and once it does, three other hand-rolled things fall away with it.

## The hook

**`mise.toml`** — a `[hooks] postinstall` in the system config links every installed tool's real binary into `/usr/local/share/tool-bin` (which leads PATH). It fires on every `mise install`: at base-image build, in each derived agent image, and whenever the agent installs a tool at runtime. Needs `experimental = true` — hooks are gated in the pinned mise 2026.6.3.

Three details the hook has to get right, all found while testing:

- `-C /etc/mise` scopes `mise bin-paths` to the system + user-global config, so an install run from inside a project directory cannot leak that project's pinned node version into the machine-wide dir.
- `[ -f ] && [ -x ]` filters the bin path: fd and rg unpack a README, a man page and an `autocomplete/` directory next to their binary.
- `if`, not `[ … ] && ln`: mise runs hooks under `sh -o errexit`, so a skipped non-executable would end the loop early and leave later tools unlinked.

**`entrypoint.sh`** — the `eval "$(mise env)"` is gone. The links do the same job without paying mise's startup at boot, and they persist for tools installed after boot.

## What that lets us delete

**Lazy-tool shims** — down to install + exec:

```sh
#!/bin/sh
export MISE_AUTO_INSTALL=false
mise install uv 1>&2
exec mise exec uv -- uv "$@"
```

Each shim serves exactly one call per tool: its install fires the hook, the hook links the binary into `tool-bin` at the front of PATH, and every later call resolves there without reaching the shim directory again. The `mise which` fast path and the shim's own lockfile `chmod` both go. `MISE_AUTO_INSTALL=false` stays: it keeps that one call from bulk-installing every other configured-but-absent tool.

**Lockfile permissions, now in one place.** mise rewrites `*.lock` 0600 on every install, and a root-run install would lock the agent user out — degrading its every mise call to a GitHub API lookup. Four Dockerfiles each healed that by hand (two `chmod 644` in platform-base, a `find … -exec chmod` in claude-code, codex and pi-agent). The hook runs after every install in every image, so it does it once for all of them, last and error-swallowing so an install by a user who does not own the lock still gets its binaries linked.

**The chowns.** platform-base installed its own tools as root and then chowned each `installs/<tool>` tree and the lockfile; claude-code, codex and pi-agent did the same for their harness CLIs, plus `/etc/mise` and one `$(dirname "$(mise where …)")` per package — a list that goes stale the moment a tool is added. All four now run `mise install` as uid 65532 instead. Everything mise writes (`/etc/mise`, `MISE_DATA_DIR`) is already agent-owned in the base image, and `tool-bin` is created with `install -d -o 65532`, so the right owner is there from the start and every `chown` in these layers is gone.

That needed a fix underneath. `HOME` is `/tmp` in this image, so the root-run installs ignored their `--mount=type=cache,target=/root/.cache/mise` and left a root-owned `/tmp/.cache` and `/tmp/.local` in the layer — which an agent-user install then could not write into. Every mise install now points `MISE_CACHE_DIR` at a cache mount owned by the build user, which makes those mounts effective for the first time and leaves no residue (`/tmp` is an emptyDir at runtime). It also silences the `rm: Permission denied` warnings the entrypoint printed on every boot. Relatedly, codex and pi-agent mounted `/root/.npm` without setting `npm_config_cache`, so npm never read it.

**The user switching.** Each image now groups its root work ahead of a single `USER 65532` and never switches back. platform-base moves its two root-only layers (the lazy-tool shims, which also create `tool-bin`, and the `/home/agent` + `/etc/passwd` layer) above the switch; codex and pi-agent needed root only for `RUN chmod +x` on two harness scripts, now `COPY --chmod=0755`, so they run as the agent user throughout.

claude-code was the hard one: it linked the adapter package into `/usr/local/lib` and the CLI into `/usr/local/bin`, both root-owned, and both paths come from `mise where`, which errors unless the tool is already installed — so the links could not be made before the install, and root could not be left behind. They now go into `tool-bin`, which the agent user owns and which already leads PATH: `tool-bin/claude` (what `CLAUDE_CODE_EXECUTABLE` points at) and `tool-bin/claude-agent-acp-lib` (the adapter directory, named apart from the `claude-agent-acp` binary the hook links there). `harness-history-lib.mjs` follows the adapter to the new path; `CLAUDE_AGENT_ACP_DIR` still overrides it, and the VM image resolves `claude` through PATH, so it needs no change.

**`Dockerfile`** — `tool-bin` exists and belongs to the agent user *before* `mise install` so the hook writes into it, and the layer ends with `test -x tool-bin/{jq,node}` so a hook that stops firing fails the build instead of quietly costing every tool call ~250ms (and, now, leaving the lockfiles unreadable).

## Verification

Built the `runner` target and all three derived agent images on top of it, and ran each as uid 65532:

- `node npm python3 jq rg fd gh` resolve to `/usr/local/share/tool-bin/…`; `python3` is a real binary rather than the lazy shim. So do `claude-agent-acp`, `codex`, `codex-acp`, `pi` and `pi-acp` in their images.
- `/etc/mise/mise.lock` is `-rw-r--r--` and agent-owned in all four images, and `find /usr/local/share/mise/installs ! -user agent` comes back empty — with every hand-rolled `chmod` and `chown` deleted.
- Runtime path: the first `uv` call goes through the lazy shim, the hook (running as uid 65532) links `uv` and `uvx` into `tool-bin`, and a fresh shell resolves `uv` from there.
- `claude --version` runs, `claude-agent-acp-lib/dist/acp-agent.js` resolves, and `harness-history-lib.mjs` imports against its new default.
- `kubectl`'s first call installs kubectl alone — `oc` and `gws` stay absent, so `MISE_AUTO_INSTALL=false` still does its job.
- Removing `experimental = true` reproduces the build failure the assert is there to catch.

The VM backend gains from this too: it copies `tool-bin` from the pod image and has no entrypoint `mise env`, so it previously started with an almost empty `tool-bin`.

Not touched: the venv-and-chown pattern in the Python agent images (gepa, nous, openevolve, shinkaevolve, skydiscover, k-search). It is the same shape but installs into `/opt` rather than through mise, so it is a separate change.

### DIFF
--- a/packages/platform-base/mise.toml
+++ b/packages/platform-base/mise.toml
@@ -27,8 +27,22 @@ oc = "latest"
 # project's pinned versions into this machine-wide dir; the system and the
 # user's global config are still both in scope. Non-executables and the
 # subdirectories some archives unpack next to their binary (fd ships a README
-# and an autocomplete/ dir in its bin path) are filtered out. Best-effort: a
-# link that cannot be written is a startup-latency wart, never a failed install
-# — mise only warns when a hook exits non-zero. `if`, not `[ ... ] && ln`: mise
-# runs the hook under `sh -o errexit`, where a skipped file would end the loop.
-postinstall = 'mkdir -p /usr/local/share/tool-bin && mise bin-paths -C /etc/mise | while read -r d; do for f in "$d"/*; do if [ -f "$f" ] && [ -x "$f" ]; then ln -sf "$f" /usr/local/share/tool-bin/; fi; done; done'
+# and an autocomplete/ dir in its bin path) are filtered out. `if`, not
+# `[ ... ] && ln`: mise runs the hook under `sh -o errexit`, where a skipped
+# file would end the loop.
+#
+# The chmod is the one place the lockfiles are made world-readable again: mise
+# rewrites them 0600 on every install, and a root-run install — the image
+# builds, a root shell on the VM backend — would otherwise lock the agent user
+# out, degrading its every mise call to a GitHub API lookup. It runs last and
+# swallows its errors, so an install by a user who does not own the lock still
+# gets its binaries linked.
+postinstall = '''
+mkdir -p /usr/local/share/tool-bin
+mise bin-paths -C /etc/mise | while read -r d; do
+  for f in "$d"/*; do
+    if [ -f "$f" ] && [ -x "$f" ]; then ln -sf "$f" /usr/local/share/tool-bin/; fi
+  done
+done
+find /etc/mise -name '*.lock' -exec chmod 644 {} + 2>/dev/null || true
+'''
--- a/packages/platform-base/mise.toml
+++ b/packages/platform-base/mise.toml
@@ -21,7 +21,8 @@ oc = "latest"
 [hooks]
 postinstall = '''
 mkdir -p /usr/local/share/tool-bin
-bin_paths=$(mise bin-paths -C /etc/mise)
+just=$(printf '%s' "${MISE_INSTALLED_TOOLS:-}" | sed 's/},{/}\n{/g' | sed -n 's/.*"name":"\([^"]*\)".*"version":"\([^"]*\)".*/\1@\2/p')
+bin_paths=$(mise bin-paths -C /etc/mise; { [ -z "$just" ] || mise bin-paths $just; } || true)
 for d in $bin_paths; do
   for f in "$d"/*; do
     if [ -f "$f" ] && [ -x "$f" ]; then ln -sf "$f" /usr/local/share/tool-bin/; fi
--- a/packages/platform-base/mise.toml
+++ b/packages/platform-base/mise.toml
@@ -1,5 +1,4 @@
 [settings]
-# The postinstall hook below is gated behind mise's experimental flag.
 experimental = true
 lockfile = true
 minimum_release_age = "7d"
@@ -20,23 +19,6 @@ kubectl = "latest"
 oc = "latest"
 
 [hooks]
-# Every tool mise installs — at image build time, or later by the agent — gets
-# its real binary linked into tool-bin, which leads PATH. Resolving the same
-# tool through a mise shim instead costs ~250ms per call on a pod (#3294).
-# `-C /etc/mise` so that installing from a project directory cannot leak that
-# project's pinned versions into this machine-wide dir; the system and the
-# user's global config are still both in scope. Non-executables and the
-# subdirectories some archives unpack next to their binary (fd ships a README
-# and an autocomplete/ dir in its bin path) are filtered out. `if`, not
-# `[ ... ] && ln`: mise runs the hook under `sh -o errexit`, where a skipped
-# file would end the loop.
-#
-# The chmod is the one place the lockfiles are made world-readable again: mise
-# rewrites them 0600 on every install, and a root-run install — the image
-# builds, a root shell on the VM backend — would otherwise lock the agent user
-# out, degrading its every mise call to a GitHub API lookup. It runs last and
-# swallows its errors, so an install by a user who does not own the lock still
-# gets its binaries linked.
 postinstall = '''
 mkdir -p /usr/local/share/tool-bin
 mise bin-paths -C /etc/mise | while read -r d; do
--- a/packages/platform-base/mise.toml
+++ b/packages/platform-base/mise.toml
@@ -21,7 +21,8 @@ oc = "latest"
 [hooks]
 postinstall = '''
 mkdir -p /usr/local/share/tool-bin
-mise bin-paths -C /etc/mise | while read -r d; do
+bin_paths=$(mise bin-paths -C /etc/mise)
+for d in $bin_paths; do
   for f in "$d"/*; do
     if [ -f "$f" ] && [ -x "$f" ]; then ln -sf "$f" /usr/local/share/tool-bin/; fi
   done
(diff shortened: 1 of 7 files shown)


packages/platform-base/mise.toml
```diff
@@ -16,3 +17,14 @@ python = "3.12"
 # contribution; these make it usable out of the box.
 kubectl = "latest"
 oc = "latest"
+
+[hooks]
+postinstall = '''
+mkdir -p /usr/local/share/tool-bin
+mise bin-paths -C /etc/mise | while read -r d; do
```
dam-code-guardian: 🟢 **Suggestion:** `mise bin-paths` is the left side of a pipeline, so its exit status is discarded even under the hook's `errexit` — a failing `bin-paths` gives an empty loop and a silently unpopulated `tool-bin` instead of a failed install. Build time is fail-closed thanks to the `test -x` gates, so this only bites runtime lazy-tool installs (where the shims dir still covers resolution). Capturing the paths first (`dirs=$(mise bin-paths -C /etc/mise)`) would make the failure loud instead.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `5104613`

## PR #3514: perf(platform-base): link tools into tool-bin from a mise postinstall hook
**Author:** JanPokorny | **Branch:** refactor/mise-postinstall-hook#locki-7doul9fo → main | **Changes:** +62 −74 (7 files)

### Summary
Replaces the per-tool `ln`/`chown`/`chmod` plumbing in four image builds with a single mise `postinstall` hook that links every installed tool's binaries into the PATH-leading `/usr/local/share/tool-bin` and heals lockfile modes, so each image can install its own tools as uid 65532 and the `mise env` eval drops out of the entrypoint.

### Findings
- 🟡 **Warning:** The description says "Every mise install now points `MISE_CACHE_DIR` at a cache mount owned by the build user, which makes those mounts effective for the first time and leaves no residue" — but the three `base`-stage root installs still carry `--mount=type=cache,target=/root/.cache/mise` with no `MISE_CACHE_DIR`, so with `HOME=/tmp` they stay exactly as ineffective (and as residue-producing) as the sentence just before describes (`packages/platform-base/Dockerfile:26,32,36`). Build-stage-only, since `runner` is a fresh `FROM`, so this is a claim/scope mismatch rather than a runtime defect. Reported by issue-fit.
  **Fix:** Narrow the claim to the agent-user installs, or extend `MISE_CACHE_DIR` to lines 26/32/36 too.
- 🟢 **Suggestion:** `bin-paths` failure is masked by the pipeline (`packages/platform-base/mise.toml:24`)
- ✅ **Looks good:** The two collision hazards this design invites are both closed — `test -x /usr/local/share/tool-bin/{jq,node}` and claude-code's `mise where` + `test -x` gate the hook's output at build time, and the `claude-agent-acp-lib` rename keeps the hook's `ln -sf` from ever resolving through a directory symlink. The removed `eval "$(mise env)"` supplied only PATH (the config has no `[env]` table), which `tool-bin` plus the pre-existing `ENV PATH` already cover in both the pod image and the VM unit.

### Documentation Check
✅ No findings.

### Issue Fit
_No linked issue — the only reference in the body, #3498, is a pull request, so scope was checked against the PR description._

Its 🟡 is carried above. Its 🟢 on `rm -rf /tmp/.local` leaving layer residue (`packages/platform-base/Dockerfile:163`) is suppressed: the `rm` runs inside the same `RUN` as the install, so nothing reaches the layer.

✅ No findings.

### Verdict
COMMENT — the mechanism is sound and gated at build time; only the description's "every mise install" claim overshoots what the diff changes.

---

JanPokorny:
Both findings addressed, plus one the local-cluster run turned up.

**🟡 `MISE_CACHE_DIR` claim vs. the base stage** — fixed in the code rather than the description: lines 26/32/36 now set `MISE_CACHE_DIR=/root/.cache/mise`, so the three base-stage installs actually use the cache mount they declare. Confirmed the builder image also has `HOME=/tmp`, so the reviewer's read was right — those mounts were dead.

**🟢 `bin-paths` failure masked by the pipeline** — the hook assigns first (`bin_paths=$(mise bin-paths …)`) and loops over the result, so under `sh -o errexit` a failure ends the hook instead of being swallowed by the pipeline's exit status.

**Found while verifying on a local cluster:** `mise use -g <tool>` writes the config *after* it installs, so the hook's `bin-paths` did not yet see the new tool and left it resolving through a shim until the next install — the exact cost this PR removes. The hook now also reads `MISE_INSTALLED_TOOLS`, which mise sets to the tools that run installed, and links those too.

Verified in a `claude-code` agent pod on `cluster:install`:

- preinstalled — `node npm python3 jq rg fd gh claude claude-agent-acp` all resolve to `/usr/local/share/tool-bin/…` and run.
- auto-installed — first `uv` call goes through the lazy shim and installs; afterwards `uv --version` resolves from `tool-bin` in **15ms**.
- explicitly installed — `mise use -g yq` links `tool-bin/yq` in the same run and resolves there.

Unrelated to this PR: `kubectl`'s lazy install hangs in the pod because `dl.k8s.io` is not reachable through the agent's egress gateway (`cdn.dl.k8s.io` returns `000`, while `github.com`, `api.github.com` and `registry.npmjs.org` all return 200). Same before and after this change — the shim retries the download either way.

jezekra1 (approved):

PR #3517, opened 2026-08-31
fix 7d window definition for usage views

jjeliga:
Some of the current usage views take the last 7d interval from "now" instead of the past 7 complete calendar days.
Fixing to avoid presentation of imprecise measurements for the first and last day of the intrval.

ref https://github.com/dam-agents/dam/issues/3269

### DIFF
--- a/docs/architecture/usage-tracking.md
+++ b/docs/architecture/usage-tracking.md
@@ -1,6 +1,6 @@
 # Usage tracking
 
-Last verified: 2026-08-26
+Last verified: 2026-08-31
 
 ## Overview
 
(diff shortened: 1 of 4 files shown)


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `a1380f3`

## PR #3517: fix 7d window definition for usage views
**Author:** jjeliga | **Branch:** `fix/usage-views-7d-window-fix` → `main` | **Changes:** +3031 −1 (4 files)

### Summary
Migration `0034` rewrites the WHERE window of all nine `_7d`/`_30d` usage views from `now() - INTERVAL 'N days'` to a whole-UTC-day range that closes at today's UTC midnight, so the per-day views stop returning a partial extra day and the counts become reproducible. The architecture doc gains the matching contract paragraph.

### Findings
- ✅ **Looks good:** coverage is complete and provably so — exactly nine `_7d`/`_30d` views exist in the tree and all nine are replaced; every remaining `NOW() - INTERVAL` occurrence (0001, 0024, 0025, 0027) belongs to a superseded definition of one of those nine, so the doc's "views with no suffix are unbounded" claim holds.
- ✅ **Looks good:** every `CREATE OR REPLACE VIEW` is legal and grant-preserving — each view's column list matches its *latest* prior definition (0024 for the auth views, 0025 for approvals, 0027 for entry_point, 0001 for the rest) in name, expression and order, and a normalized per-view body diff shows only the one `>=` line replaced plus one `<` line added. No filter (`actor_sub IS NOT NULL`, the `usage_core_*` exclusions) was dropped.
- ✅ **Looks good:** the window is session-`TimeZone`-independent — `date_trunc('day', now() AT TIME ZONE 'UTC') AT TIME ZONE 'UTC'` pins UTC midnight where `CURRENT_DATE` would have followed the connection's zone, which is what made the old numbers irreproducible. `>= midnight − 7 days` with `< midnight` is exactly seven complete days, matching the doc's "read on a Monday covers the previous Monday through Sunday".
- ✅ **Looks good:** `meta/0034_snapshot.json` differs from `0033` in `id`/`prevId` only (views aren't in the Drizzle snapshot), `_journal.json` keeps `when` monotonic, and no consumer drifts — `report-service.ts` renders view names generically from `VIEW_NAMES`, with no hard-coded window prose.

### Documentation drift
✅ No findings.

### Issue alignment
✅ No findings.

### Verdict
APPROVE — a surgical, verified WHERE-clause-only migration with a matching doc contract; nothing blocking and no residual suggestions worth raising.

---

Issue #3509, opened 2026-08-28
UI - A new coding agent opens on an empty chat with nothing to say what it can do


jenna-winkler:
## Context

Experiments and knowledge bases both greet the user on first open — the UI runs a hidden first turn so the agent introduces itself, says what the sandbox is for, and asks one question. Coding agents have no equivalent. They are the most-created type and the only one that opens silent.

## Problem

A user finishes creating a coding agent and lands in an empty chat. Nothing tells them what this agent can do, what it can already reach, or what to ask for first. Skills, schedules, sub-agents, artifacts and Slack collaboration stay invisible — the same gap #3473 describes at creation, one screen later. A user who does not arrive with a concrete task in hand gets nothing from their first agent.

## Goal

Opening a fresh coding agent starts with the agent introducing itself: what it can do, what it can reach, and one question that gets the user to their first real task.

## User Stories

- As a user, I want a new coding agent to tell me what it can do so that I know what to ask for.
- As a user, I want to learn what it already has access to so that I do not set up connections it has.
- As a user, I want to be asked what work I want done so that my first request is a real one.
- As a user, I want the greeting to match the other agent types so that the product reads as one thing.

## Scope

**In:** the greeting for a freshly created coding agent — what it says, what it asks, what it checks before speaking.

**Out:** the creation flow itself (#3473); template browsing (#2909); the existing knowledge base and experiment greetings (#3299 covers the knowledge base copy).

## Open Questions

- What does it check before it speaks — connected repos, installed skills, provider?
- One question, or a short orientation and then a question?
- Does it differ on a custom image, where we cannot assume what is installed?

## Additional resources

- #3473 — creation never shows what DAM can do; this is the same gap one screen later
- #3299 — design pass on the knowledge base greeting copy
- #3409 — the knowledge base greeting can race its own setup

PR #3520, opened 2026-09-01
fix: gate greetings on workspace-command install

Tomas2D:
- a pending workspace-command now holds the agent in preparing_workspace,
  exactly like workspace-seed, so a fresh KB is not operable — and its
  hidden /wiki-onboard greeting cannot fire — until the bootstrap has run
- emit WorkspaceMutationSettled when a workspace-seed/workspace-command
  event settles and map it to an agents live hint: the settle is a bare
  Postgres stamp with no Agent CR change behind it, so without the push
  the UI would stay on "Preparing workspace" until a refetch happened by
  accident (this also fixes the stale overlay for repo-seeded agents)
- generalize the preparing_workspace overlay copy beyond seed clones
- docs: close the known-gap paragraph in knowledge-bases.md, document
  workspace-command and the preparing-workspace gate in runtime-delivery.md

Ref: #3409

Ref: #3509


packages/api-server/src/modules/runtime-delivery/infrastructure/outbox-repo.ts
```diff
@@ -260,15 +260,18 @@ export function createOutboxRepo(db: Db): OutboxRepo {
...
+    async preparingWorkspaceAgentIds(agentIds): Promise<Set<string>> {
       if (agentIds.length === 0) return new Set();
       const rows = (await db
         .select({ agentId: runtimeEvents.agentId })
         .from(runtimeEvents)
         .where(
           and(
             inArray(runtimeEvents.agentId, agentIds),
-            eq(runtimeEvents.kind, "workspace-seed"),
+            inArray(runtimeEvents.kind, [
```
dam-code-guardian: 🟡 **Warning:** A `workspace-command` that never settles holds the agent in `preparing_workspace` until its TTL lapses, and nothing retries it. When the agent-runtime handler throws, `processEvents` (`event-loop.ts:40`) omits the id from `settledEvents`. `applyState.failures` carries contribution failures only, so `recordOutcome` takes its clean branch, `dispatchedAt` stays NULL, and `listRetryable` matches neither condition. `hello` also skips the enqueue because `row.version <= lastAppliedVersion`. For a kinded-agent install the TTL is 30 days (`kinded-agent-create.ts:7`), so one failed install makes the agent unusable for 30 days with no error shown. Before this change the same failure left the agent usable. An agent on an older runtime image hits the same state: `filterByCapabilities` drops the unknown kind, so it can never settle. `docs/architecture/knowledge-bases.md:30` states the gate covers "a failed install retrying" — no retry exists.

**Fix:** Bound the preparing-workspace window for every workspace-mutating kind. Make an unsettled event retryable (stamp an event-level attempt so `listRetryable` re-enqueues, with a give-up that marks the event dispatched-with-error), and treat a workspace-mutation kind in `payload.droppedEventKinds` as terminal for that event. Then correct the `knowledge-bases.md:30` sentence to the behaviour that ships.

mise.lock
```diff
@@ -203,29 +203,6 @@ url = "https://get.helm.sh/helm-v4.2.0-windows-amd64.tar.gz"
 version = "21.0.2"
 backend = "core:java"
 
-[tools.java.options]
```
dam-code-guardian, line 206: 🟡 **Warning:** This PR removes `[tools.java.options]` and all five `[tools.java."platforms.*"]` blocks, so the JDK loses its per-platform sha256 checksums and pinned URLs. `mise.toml:18` still requires `java = "21"`, and the sibling `maven` tool keeps all its platform blocks, so the removal is asymmetric. It reads as a partial lockfile regeneration on a machine that had not resolved those platforms. No PR-body bullet mentions tooling.

**Fix:** Keep lockfile changes out of a behaviour PR — restore `mise.lock` (`git checkout <base> -- mise.lock`). If the java entries do need regenerating, do it in a separate tooling change with the platforms resolved, so the checksums are re-emitted instead of dropped.

packages/api-server/src/modules/runtime-delivery/services/worker-handler.ts
```diff
@@ -20,9 +20,15 @@ export interface WorkerHandlerDeps {
   agentRunningPort: IsAgentRunning;
   snapshotWriter: HarnessConfigSnapshotWriter;
   clientFor(agentId: string): AgentRuntimeClient;
+  resolveOwner: (agentId: string) => Promise<string | null>;
   log: (msg: string) => void;
 }
 
+const WORKSPACE_MUTATION_KINDS = new Set([
```
dam-code-guardian: 🟢 **Suggestion:** The set of workspace-mutating kinds now exists twice, untyped: `WORKSPACE_MUTATION_KINDS` here and the `inArray(runtimeEvents.kind, [...])` predicate in `outbox-repo.ts:271`. A third kind added to the repo only puts the agent in `preparing_workspace` without ever emitting the hint — the bug this PR fixes. Export one `readonly RuntimeEventKind[]` constant and read it in both places, so the type checker binds them together.

dam-code-guardian (commented):
## ⚠️ **Code Guardian 🛡️** — Review at `071c491`

### Summary
A pending `workspace-seed` or `workspace-command` event now holds an agent in `preparing_workspace`, and the delivery worker emits `WorkspaceMutationSettled` when such an event settles so watching UIs re-read the agent. The gate closes the `/wiki-onboard` greeting race in #3409.

### Findings
- 🟡 **Warning:** an unsettled `workspace-command` traps the agent in `preparing_workspace` for the event's full TTL, with no retry (`packages/api-server/src/modules/runtime-delivery/infrastructure/outbox-repo.ts:271`, doc claim at `docs/architecture/knowledge-bases.md:30`)
- 🟡 **Warning:** undeclared removal of the java platform checksums and URLs (`mise.lock:206`)
- 🟢 **Suggestion:** the workspace-mutating kind list is duplicated in two files (`packages/api-server/src/modules/runtime-delivery/services/worker-handler.ts:27`)
- ✅ **Looks good:** the `seedingAgentIds` → `preparingWorkspaceAgentIds` rename reaches every call site, and the `resolveOwner` chain from `bootstrap.ts` through `compose.ts` into the worker is complete.

_No drift found: both edited architecture pages carry the behaviour this PR ships and a bumped `Last verified:` date._ The one inaccurate clause is folded into the first 🟡 above.

### Issue Fit
The code files all map to a stated bullet, and #3409's race is closed by behaviour through the pending-event → `preparing_workspace` → `operable` → `armed` chain. `mise.lock` is the one change no bullet covers — folded into the second 🟡 above. `Ref: #3509` delivers no part of that issue; it carries no closing keyword, so nothing auto-closes.

See the two findings above. The new event type, the saga arm and its `never` exhaustiveness, and the test doubles all check out — the owner-null and unsettled-event cases are genuinely exercised.

_No findings._ The overlay copy "Setting up the workspace." is correct for both kinds that now reach `preparing_workspace`, no stale copy of the old string remains in `packages/ui`, and `OVERLAY_COPY` stays exhaustive over the display-state union.

### Verdict
COMMENT — the mechanism and its wiring are correct; two warnings hold it below approval: the unbounded preparing-workspace window when an event never settles, and the undeclared lockfile deletion.

---
<sub>🛡️ Reviewed by **Code Guardian 🛡️** · [definition](https://github.com/dam-agents/code-guardian)</sub>

packages/api-server/src/modules/runtime-delivery/infrastructure/outbox-repo.ts
Tomas2D: Fixed in 83a74f81. The preparing-workspace window is now bounded for every workspace-mutating kind: `runtime_events` gained an `attempts` counter (and a terminal `error`), `recordOutcome` increments it for events delivered but not settled, `listRetryable` re-enqueues agents holding such an event while it is under the budget (reusing the max-apply-attempts budget), and exhausting the budget stamps the event `dispatched_at` with the error recorded — the agent leaves preparing-workspace and the worker emits the WorkspaceMutationSettled hint. A workspace-mutation kind in `droppedEventKinds` is stamped undeliverable at dispatch time for the same reason. knowledge-bases.md and runtime-delivery.md now describe this shipped behaviour.

mise.lock
Tomas2D: Restored in db490497 from the merge base — the java platform pins were being stripped by local mise runs on a machine without those platforms resolved; no tooling change belongs in this PR.

packages/api-server/src/modules/runtime-delivery/services/worker-handler.ts
Tomas2D: Fixed in 83a74f81: the kind set is now a single typed constant, WORKSPACE_MUTATION_EVENT_KINDS (`as const satisfies readonly RuntimeEventKind[]`) in runtime-delivery/domain/workspace-mutation.ts, read by the outbox repo, the worker handler, and the new give-up/undeliverable paths via the isWorkspaceMutationKind guard.

PR #3521, opened 2026-09-01
fix(controller): stop stacking mig- prefixes on migration target names

jezekra1:
## Problem

`migrationTargetName` prepends `mig-` to the source PVC's name unconditionally, so every storage-class round trip stacks another prefix: `home-agent-x-0` → `mig-home-agent-x-0` → `mig-mig-home-agent-x-0` → … The names grow without bound (until the 63-char truncation starts eating the meaningful tail).

## Change

The target name now strips every accumulated `mig-` prefix off the source name first, and re-adds a single `mig-` only when the stripped name would collide with the source (i.e. the source was never migrated). Names alternate between exactly two forms forever:

- `home-agent-x-0` → `mig-home-agent-x-0` (first migration, unchanged)
- `mig-home-agent-x-0` → `home-agent-x-0` (normalized)
- `mig-mig-mig-home-agent-x-0` → `home-agent-x-0` (stacked names normalize in one hop)

This is safe because nothing addresses a workspace PVC by name after a flip: pods are wired to it by the `agent`/`mount` labels (`listClaimedPoolPVCs`), and `applyPoolClaims` drops the volumeClaimTemplate for the claimed mount and mounts the PVC as a direct volume reference — so a normalized name matching the StatefulSet's `<mount>-<agent>-0` template pattern cannot be adopted by a template.

## Rollout note

Deploy a controller with this change only while no migration copy Job is mid-flight (e.g. migration disabled or drained). A controller restarted mid-copy would compute a different target name than the running Job writes to and pair the flip with an empty volume.

## Tests

- `TestStorageMigration_TargetNameNeverStacksPrefixes` — unit coverage of the naming, including 63-char truncation
- `TestStorageMigration_PrefixedSourceMigratesOntoNormalizedName` — a previously-migrated source lands on the clean name
- `mise run controller:test`, `controller:check`, `common:check:comment-types` all pass

### DIFF
--- a/packages/controller/pkg/reconciler/storage_migration.go
+++ b/packages/controller/pkg/reconciler/storage_migration.go
@@ -425,20 +425,6 @@ func (m *StorageMigrationManager) migrateAgent(ctx context.Context, agent *apiv1
 			return err
 		}
 
-		if stale := staleJobTargets(job, pairs); stale != nil {
-			slog.Warn("storage migration: copy job writes to volumes the current code no longer targets; discarding it for a clean recopy",
-				"agent", name, "job", job.Name, "jobTargets", strings.Join(stale, ","))
-			prop := metav1.DeletePropagationBackground
-			if err := m.client.BatchV1().Jobs(m.config.Namespace).Delete(ctx, job.Name,
-				metav1.DeleteOptions{PropagationPolicy: &prop}); err != nil && !errors.IsNotFound(err) {
-				return err
-			}
-			if err := m.deleteUnpairedTargets(ctx, name, pairs); err != nil {
-				return err
-			}
-			return fmt.Errorf("copy job %s targeted stale volumes; recopying next tick", job.Name)
-		}
-
 		switch {
 		case jobSucceeded(job):
 			return m.flip(ctx, agent, pairs, job.Name)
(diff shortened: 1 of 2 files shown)


packages/controller/pkg/reconciler/storage_migration.go
```diff
@@ -355,7 +355,13 @@ func isRWX(modes []corev1.PersistentVolumeAccessMode) bool {
 }
 
 func migrationTargetName(oldPVC string) string {
-	name := "mig-" + oldPVC
+	name := oldPVC
+	for strings.HasPrefix(name, "mig-") {
```
dam-code-guardian, line 359: 🟡 **Warning:** The target name for a given source changes with this commit, and `migrateAgent` reuses an existing copy Job by name alone. A controller that restarts mid-copy computes `home-agent-x-0` for source `mig-home-agent-x-0`, creates that PVC, and on `jobSucceeded` flips the agent onto it — while the running Job is still writing into `mig-mig-home-agent-x-0`. `flip` then labels the source superseded and deletes it, so the workspace lands on an empty volume.

Your rollout note names this window and gives the operator procedure. The residual is that nothing detects a violation of it: the failure is silent and destroys the workspace.

**Fix:** Make the flip verify the volume it is flipping onto. Before `flip`, read the live Job's `dst-*` volume claim names and compare them to the computed `pairs` targets; on mismatch, delete the Job and return an error so the next tick recopies. State the rule generally: whenever a target name is derived rather than persisted, a resumed Job must be validated against the names the current code computes, never trusted by Job name alone.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `7fec47f`

## PR #3521: fix(controller): stop stacking mig- prefixes on migration target names
**Author:** jezekra1 | **Branch:** fix/storage-migration-name-stacking → main | **Changes:** +34 −1 (2 files)

### Summary
`migrationTargetName` strips accumulated `mig-` prefixes and re-adds one only when the stripped name equals the source. A workspace PVC now alternates between two names instead of growing a prefix per round trip.

### Findings
- 🟡 **Warning:** A copy Job that is mid-flight across the controller upgrade writes to a different target than the new code flips onto (`packages/controller/pkg/reconciler/storage_migration.go:359`)

✅ No findings.

### Issue Fit
✅ No findings.

### Verdict
COMMENT — the naming logic is correct and the body's safety argument holds. The open item is the mid-flight upgrade window: the rollout note documents it, the code does not detect it.

**Checked and clear:**
- The volumeClaimTemplate adoption concern the body addresses does hold. `resolveWorkspaceClaims` reads the live StatefulSet's direct volume references, and falls back to the `agent`/`mount` PVC labels when the StatefulSet is absent, so `applyPoolClaims` drops the `home-agent` template in every state of a migration. A normalized name matching the template pattern is never adopted.
- The 63-char truncation keeps the two-name alternation. A 70-char source settles between `mig-<59>` and `<59>`. Truncation collisions between two distinct sources were equally possible before this change.

---

jezekra1:
Re the mid-flight-upgrade warning: acknowledged as a real mechanism, deliberately **not** guarded in code (a guard was pushed in 19759241 and reverted in bcc14fce). Rationale:

- The window requires a copy Job mid-flight at the exact moment the controller rolls, and only for already-once-migrated (`mig-`-prefixed) sources. No drain is active on any install right now, and the operator controls both the drain and the rollout.
- The designed mitigation already exists and is zero-code: `storageMigration.enabled: false` before the rollout releases gates and bins half-copied targets; re-enable after.
- Worst case is recoverable, not destructive: flip only follows a *succeeded* Job, so the checksum-verified copy exists in the old-named target PVC (labeled, owner-refed) — the agent wakes empty but the bytes are recoverable by hand.
- Weighed against ~115 lines of new code (including a label-driven PVC-deletion helper) in a battle-tested destructive path, the guard is the larger risk. Shipping the one-function fix keeps the blast radius reviewable.

Rollout rule stands as documented in the PR body: deploy this controller only while no migration copy Job is in flight — disable the migration first if unsure.

xjacka (approved):

PR #3522, opened 2026-09-01
feat(controller): per-agent storage class pin on the Agent CR

jezekra1:
## What

Adds `spec.storageClass` to the Agent CR: a per-agent pin for the class its workspace volumes provision on. Empty (the default for every existing Agent) keeps today's behavior — the install-wide `controller.agent.base.storageClass`.

The pin is honored at every point that chooses a class:

- **Creation** — the StatefulSet's volumeClaimTemplates (and VM-backend workspace PVCs) provision on the pin, falling back to `agent.base.storageClass`.
- **Storage migration** — the pin overrides `storageMigration.targetStorageClass` as that agent's destination, in selection (`Reconcile`, `ReleaseGated`) and in target provisioning (`ensureTargetPVC`). Pinning the class a workspace already sits on produces no migration reason, so the agent is exempt from the drain while the rest of the fleet moves — the use case that motivated this.
- **Warm pool** — an agent pinned to a class other than the pool's bypasses spare claiming and provisions on demand; a spare on the wrong backend is worse than a slow first start.

`Reconcile` now lists Agents before classifying PVCs (each PVC is judged against its agent's target); `ReleaseGated` got the same restructure so abandoned-migration detection honors pins too.

## Schema

Agents CRD schema generation bumped 8 → 9 (`crdcheck.Assert` will hold a new controller until the CRD is upgraded, as usual). The field is optional and additive — `crd-backwards-compatibility` passes.

**Stacked on #3527** (CRD descriptions restored + generated-manifest gate): this PR's base is that branch until it merges, after which GitHub retargets to main. The generated-file diff is now surgical — the `storageClass` property with its description and the generation bump, nothing else — and the new `controller:check:crd-generated` gate verifies the manifests match controller-gen output exactly.

## Tests

- `TestStorageMigration_PinnedAgentIsLeftOnItsClass` — pinned workspace on its class is not drained even with a different fleet-wide target
- `TestStorageMigration_PinnedAgentMigratesOntoItsPin` — target PVC provisions on the pin, not the global destination
- `TestStorageMigration_DisabledReleaseHonorsPin` — `ReleaseGated` classifies a pinned agent's still-needed migration correctly and bins its half-copied target
- `TestBuildAgentStatefulSet_SpecStorageClassOverridesBase` — creation-time override
- `TestReconcile_PinnedAgentSkipsWarmPool` — pinned agent leaves spares unclaimed and provisions via volumeClaimTemplate

[...]

PR #3379, opened 2026-08-20
refactor(agent-runtime): extract session bootstrap from acp-runtime

tomkis:
Next step in the acp-runtime decomposition series (#3331, #3373).

## What

- `session-bootstrap.ts` now owns the session/load + session/resume machinery that was spread over five regions of acp-runtime: serve-from-transcript, parked waiters, the cold session/load the runtime issues for a resume, waiter drain on the load response, initiator cleanup on detach.
- Same shape as the earlier extractions: factory, `Deps` interface, `UNIT_BOUNDARY_DESCRIPTION`. The runtime keeps only the hooks that touch its correlation table (`openLoadRoute`, `sendToAgent`).
- The waiter kind branch is an exhaustive `ts-pattern` match closed with `.exhaustive(fallback)`. ts-pattern added as an agent-runtime dependency (same `^5.9.0` api-server uses).
- New `docs/guidelines/code/` rules index, linked from AGENTS.md, with the first rule: exhaustive matching.

## Notes for review

- Behavior-identical apart from two deliberate deviations: dropped the unread `initiatorOutboundId` field, merged the two serve helpers into one `respondFromLog(kind, ...)` (the load/resume ordering difference is preserved).
- All 198 agent-runtime tests pass unchanged. They drive the runtime through its public interface, so they double as proof the extraction changed nothing observable.

PR #3527, opened 2026-09-02
fix(controller): CRD doc comments are generator input — restore and gate them

jezekra1:
## Problem

The repo-wide comment strip (#3308, merged Aug 13 at CRD generation 7) deleted the `api/v1` Go doc comments. Those comments are not reader prose — they are **controller-gen input**: they compile into the CRD `description` fields (`kubectl explain`) and, through the CRD YAML, into the api-server's generated TS JSDoc. #3308's own verification couldn't catch this: the schema-generation check strips descriptions before comparing.

Since then the checked-in manifests have been unreproducible — `controller:generate` produces description-less output that no longer matches the chart. Gen 8 (#3318, Aug 24) shows the consequence: its new status fields carry full descriptions in the YAML that exist nowhere in the Go source, i.e. the "Code generated by controller-gen. DO NOT EDIT." file was edited by hand. `api-server-api:check:gen` passes on main only via mise source caching.

## Change

1. **Exempt `packages/controller/api/` from the comment stripper** (one filter in the shared `listFiles`, which also drives `check-comment-types`). Same category as the `+kubebuilder` markers the policy already preserves: instructions to tools, not prose.
2. **Restore the doc comments** from the pre-strip state (`69b64d70^`), reconciled with the gen-8 fields added since; the gen-8 descriptions that lived only in the hand-edited YAML are back-ported into Go doc comments verbatim.
3. **New gate: `controller:check:crd-generated`** — regenerates the CRDs into a temp dir and diffs against `deploy/helm/platform/crds/` in both directions (drift, missing from chart, orphaned in chart). Runs under `controller:check`, so pre-commit hook and CI both enforce it. Verified it fails on a hand-edited manifest and passes clean.

## Proof of reproducibility

Regenerating on this branch changes **nothing**: the CRD YAMLs and `crd-types.gen.ts` are byte-identical to what's checked in on main. No schema change, no generation bump needed (spec is untouched — descriptions only).

## Note

#3522 (per-agent storage class) currently carries the description-less regenerated files; after this merges it needs a rebase that adds a doc comment for the new `storageClass` field and regenerates.


dam-code-guardian (dismissed):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `d6b4abb`

The diagnosis here is correct and well-evidenced: the doc comments under `packages/controller/api/v1/` really are controller-gen input, and the checked-in manifests really had drifted from source. I verified the central claims — every restored field comment resolves to a `description:` in `deploy/helm/platform/crds/` (the `agentPodRestarts` text existed only in the YAML before this PR), the remaining YAML descriptions come from vendored `metav1` types, and the new gate's two-way file comparison matches the chart's contents. The gofmt de-alignments (`ConditionReady`, `Path`/`Persist`, `AddToScheme`, `AgentSchemaGeneration`) are each separated by an intervening comment, so `controller:check:fmt` stays green.

One blocker: the exemption is placed in shared code and takes the comment-type checker down with it.

### Findings

- 🔴 `scripts/strip-comments.mjs:117` — The new `packages/controller/api/` filter is added inside the **exported** `listFiles()`, which is also the file enumerator for `scripts/check-comment-types.mjs:42`. It therefore does more than stop the stripper deleting the CRD doc comments — it removes the whole tree from `mise run common:check:comment-types`, leaving `packages/controller/api/` with no comment-type enforcement at all. Both the code comment and `packages/controller/tasks.toml:149` justify only "don't strip these", so the checker loss is unintended.
  **Fix:** leave `listFiles()` unfiltered and apply the exemption stripper-side only — export a predicate (e.g. `isStripExempt(f)`) and test it in `main()`'s per-file loop next to the existing `isGenerated` / `import "C"` skips, so `check-comment-types.mjs` keeps enumerating the tree.
- 🟡 `scripts/strip-comments.mjs:117` — `startsWith('packages/controller/api/')` exempts the entire tree, but only field and root-type doc comments feed controller-gen. `schema_generation.go`, `groupversion_info.go` and `schema_generation_test.go` are exempted although their comments never reach a CRD description — and the untyped prose this PR adds at `packages/controller/api/v1/schema_generation_test.go:12` is permissible only because of the new filter. The generator-input rationale is doing more work than it supports.
  **Fix:** narrow the filter to what is genuinely generator input (`*_types.go` under `packages/controller/api/`, or the two type files by name), or state in the PR body that the exemption is deliberately whole-tree and that non-generator prose there is now outside the comment-type policy.
- 🟡 `docs/guidelines/comment-guidelines.md:3` — The policy document still says the rule is "Enforced twice" and that the stripper "deletes every comment that carries no registered type", and presents exemptions as directive patterns in `PRESERVE_TS`/`PRESERVE_GO`. After this change both statements are false for `packages/controller/api/`, and the carve-out is recorded only in `packages/controller/tasks.toml:149` — not in the guideline that `CLAUDE.md:31` points agents to as the contract.
  **Fix:** document the `packages/controller/api/` carve-out and its reason (comments are CRD `description` source) in `docs/guidelines/comment-guidelines.md`, next to the existing `+kubebuilder:` note.
- 🟡 `packages/controller/tasks.toml:152` — The PR body names mise source caching as the reason the original bug survived on `main`, but the new gate declares `sources = ["api/**/*.go", ".../crds/*.yaml"]` and so inherits the same skip-on-unchanged-sources behaviour. The list also omits `mise.toml`, where `controller-gen` is pinned to `"latest"` — so a generator upgrade that changes description wrapping is not treated as a reason to re-run this byte-identity gate.
  **Fix:** add `mise.toml` (and `packages/controller/tasks.toml`) to `sources`, or drop `sources` so a correctness gate always runs.
- 🟢 `packages/controller/api/v1/schema_generation.go:31` — The verbatim restoration carries over "(Folded into gen 1 pre-release — this CRD ships first with this PR.)", which referred to the PR that introduced UserBudget. Here it reads as if #3527 ships that CRD, which has been on `main` since #3379.

✅ No findings.

### Issue Fit

Covered above — the scope and documentation items from this check are folded into the findings list (no linked issue; #3308, #3318 and #3522 are pull requests, so scope was assessed against the PR description).

Covered above. Two additional suggestions:

- 🟢 `scripts/strip-comments.mjs:117` — Filtering inside `listFiles()` makes the exemption invisible in run output: unlike the `generated` and `cgo preamble` skips, these files never enter `skipped[]`, so the dry run's scanned count silently shrinks with nothing explaining why.
- 🟢 `scripts/strip-comments.mjs:111` — The exempt set is now two inline string comparisons in a `.filter()` chain while `check-comment-types.mjs:19` keeps its own separate `SELF` constant; one exported predicate would stop the two consumers' notions of "not subject to the rule" drifting further apart.

### Summary

**Verdict: ❌ REQUEST_CHANGES** — one 🔴 to resolve. The approach is right; the placement of the exemption is the problem. Moving it out of `listFiles()` into the stripper's own loop addresses the blocker and, if scoped to `*_types.go`, the first 🟡 with it.

Audit note: the mechanical added-comment scan flagged 199 lines, all of them the restored `packages/controller/api/` doc comments this PR exists to reinstate and justifies explicitly, plus 5 lines in `scripts/strip-comments.mjs` (already excluded from `listFiles()` by the pre-existing self-filter) and 6 in `tasks.toml` (TOML — outside the rule's TS/Go scope). Not raised as a finding; the policy-documentation gap it points at is the 🟡 above.

---

tomkis (approved):

PR #3529, opened 2026-09-02
feat(metrics): break Usage spend down by session type

xjacka:
Adds a **Spend by session type** section to the Usage tab — chats, channels, scheduled runs, experiments, terminal — behind the existing **Session costs** experimental flag.

## Why this needs more than a `GROUP BY`

Telemetry cannot answer what kind of session drove a cost. A call record names its session but never its kind, and the resource attributes that could carry it are stamped once per pod while a single pod serves sessions of every kind at once.

The session metadata that *does* know the kind is agent-owned: it lives on the agent's own volume. That is the right home for it and a useless one for this read, because it is unreachable exactly when it matters — detached while the agent hibernates, and reclaimed when the agent is deleted, while the agent's spend deliberately survives in telemetry. The kind would go missing precisely for the windows whose cost still counts.

So agents report **one dimension per session — its kind** — over the existing harness channel into a Postgres projection the read path joins against. Everything else about a session stays agent-owned; nothing reads sessions from this table.

## Notable decisions

- **Gating is server-side.** The rollup rides in the same procedure as the rollups every caller gets, so a hidden UI section would still cost a telemetry query and a directory lookup per read. Without the flag it runs neither.
- **Reports are snapshots, sent only when the answer changes.** Session metadata is touched on every turn (activity stamps) while a session's kind is fixed at creation, so the agent compares what it would send against what it last sent. An ordinary conversation produces no writes. A missed report costs latency, not a permanent gap — the next one converges.
- **Unknown sessions are booked, not dropped.** The breakdown sits under a total the user can read, so silently shedding spend would make the two disagree. The bucket covers agents that have not reported yet and deleted agents whose telemetry outlives their rows.
- **Child harness runs fold into the spawning session's kind**, reusing the trace-family root the per-session rollup already groups on.
- **Session categorisation moved to the contract package**, so the Usage tab and the sessions sidebar share one definition instead of drifting apart.
- **The directory ages out** (second commit). It is the only table that grows with *use* rather than with the number of things that exist, so it cannot rely on its rows being deleted along with whatever they describe. A weekly job under an advisory lock does it, same shape as the activity log; the shared lock helper moved to `core/` now that two modules need it.

## Verified on a local cluster

End to end, with clickstack enabled:

| session | kind | calls | cost |
|---|---|---|---|
| `fc81f848…` | `regular` → Chats | 5 | $0.1830 |
| `b2f7477c…` | `schedule_cron` → Scheduled | 5 | $0.1355 |

Both session ids present in telemetry resolve to their kind in the directory, and the two categories sum to the total.

## Known limits

- Retention deletes on `created_at`, so a session older than the horizon that is still active would lose its kind while its spend is in range. Tracking last activity instead would mean a write per turn, which is the cost this design avoids.
- Agents created while the telemetry backend was disabled never export at all, so they contribute nothing here — unrelated to this change, but it is what you will see if the breakdown looks empty.
- The retention job has no test: it is a timer plus a lock plus one `DELETE`, and its predecessor in the repo has none either.


packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts
```diff
@@ -151,6 +152,49 @@ export function createClickhouseReader(
...
+         )
+         SELECT
+           coalesce(nullIf(rootSid, ''), sid) AS sessionId,
+           sum(rowCostUsd) AS costUsd
+         FROM (
+           SELECT
+             LogAttributes['session.id'] AS sid,
+             ${COST_USD} AS rowCostUsd
+           FROM otel_logs
+           WHERE ${ownedApiRequests(window)}
```
dam-code-guardian, line 186: 🟡 **Warning:** this `calls_rows` select keeps every owned API-request row, including those whose `session.id` is empty. `coalesce(nullIf(rootSid, ''), sid)` then yields `''` for them, the directory can never match that id, and the cost is booked into the `unknown` bucket rather than excluded. The otherwise-identical `runtimeBySession` filters them out in exactly this position (`:245`).

```suggestion
           WHERE ${ownedApiRequests(window)} AND LogAttributes['session.id'] != ''
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `c9501aa`

## PR #3529: Usage breakdown by session type
**Author:** xjacka | **Branch:** claude/usage-breakdown-session-types-cb9780 → main | **Changes:** +3735 −36 (50 files)

### Summary
Adds a flag-gated "spend by session type" breakdown to the Usage view, fed by a new ClickHouse rollup that folds a trace family onto its root session, plus a `session_directory` module the agent runtime reports session kinds into and a 180-day retention job for it.

### Findings
- 🟡 **Warning:** the new `spendBySession` rollup keeps rows whose `session.id` is empty, so their cost lands in the `unknown` bucket instead of being excluded (`packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts:186`)
  **Fix:** add `AND LogAttributes['session.id'] != ''` to the `calls_rows` predicate, as the sibling `runtimeBySession` does at `:245`.
- 🟡 **Warning:** the last ACP client to unsubscribe clears the coalescing timer that `notify()` shares with the watchers, so a pending session-directory report is cancelled and the change waits for the next unrelated `notify()` (`packages/agent-runtime/src/modules/acp/services/session-changes.ts:39`). `subscribe`'s teardown checks only `listeners.size === 0`, while the directory watcher registered at `packages/agent-runtime/src/server.ts:251` stays for the process lifetime.
  **Fix:** clear `pending` only when `listeners.size === 0 && watchers.size === 0`, and keep `hooks?.stop()` on the listener count alone.
- 🟢 **Suggestion:** the flag gate and the stored-row→category parse both sit one layer above where the architecture places them — the gate as an `isEnabled` dep on the service rather than inside the reader adapter, the parse in the service rather than as a repository mapper (`packages/api-server/src/modules/metrics/services/session-type-spend.ts:22`, `packages/api-server/src/modules/session-directory/services/session-directory-service.ts:26`; from TypeScript Engineering Review). Placement only — the behaviour is correct — so this reads as a suggestion, not a change request; moving the parse also removes the two enum casts.
- 🟢 **Suggestion:** two architecture pages still describe the pre-PR model — `docs/architecture/features.md:38` says a flag does not gate the underlying API, and `docs/architecture/runtime-delivery.md:16` describes the runtime channel as two routes with `hello` as the only outbound one, while this PR adds a server-side data gate and the `sessionDirectory.v1.report` route (from Documentation Check).
- 🟢 **Suggestion:** `session-type-spend-bars.tsx` is the third component with the same body as `agent-spend-bars.tsx` and `model-spend-bars.tsx` (`packages/ui/src/modules/metrics/components/session-type-spend-bars.tsx:8`; from React UI Engineering Review) — a shared `<SpendBarList>` would carry all three.
- ✅ **Looks good:** the body's claim that the flag path costs nothing when off holds — `session-type-spend.ts:26` returns before touching either the telemetry reader or the directory; and the retention job's advisory-lock key (`0x73_65_73_64_69`) does not collide with the existing usage job's.

Verdict: minor drift

🟡 `docs/architecture/features.md:38` — the page states the flag mechanism plainly: "It does **not** gate the underlying API: the feature's tRPC endpoints remain available to the authenticated owner regardless of the flag", and lists UI surfaces and the agent surface as the only things a flag controls. This PR introduces a third kind — a server-side data gate inside an endpoint every caller reaches (`session-type-spend.ts:29` returns `[]` when `isEnabled()` is false; `trpc/context.ts:278` reads the `session-costs` flag) — and `docs/architecture/metrics.md` now documents it as deliberate ("The gate is server-side rather than a hidden section"). The features page is the owner of what a flag gates and no longer describes what the code does.
  Fix: update `features.md`'s "What a flag does — and does not — gate" to admit a server-side response gate on a shared read, with the reason metrics.md gives (an ungated rollup would charge every caller a telemetry query and a directory lookup), and bump its `Last verified:`.

🟡 `docs/architecture/runtime-delivery.md:16` — the page owns the agent→api-server channel and describes it as closed: "The runtime channel is two routes between api-server and agent-runtime", the context bullet at :13 enumerates `applyState` plus the `hello` callback, and the invariant at :326 says "the agent's only outbound channel is the paired gateway, which routes back to the harness-API-server's `hello`". This PR adds a third harness route the agent calls — `sessionDirectory.v1.report` (`packages/api-server-api/src/harness-router.ts:7`, mounted by `runtime-trpc.ts`) — and metrics.md's new section points at this page as the channel it rides ("over the same authenticated agent→api-server channel [runtime delivery](runtime-delivery.md) uses"). The page it defers to does not mention the route.
  Fix: add the session-directory report to runtime-delivery.md's route description and to the outbound-channel invariant at :326 (semantically — what is reported and when, not the type), and bump its `Last verified:`.

_Merged into one 🟢 in `### Findings`: doc drift is one finding per review here._

### Issue Fit
_No linked issue — scope checked against the PR description only._ (No closing keyword or plain `#<n>` reference in the body; head branch `claude/usage-breakdown-session-types-cb9780` carries no issue number.)

✅ Every item the description promises is in the diff — the flag-gated Usage section, the server-side gate, the change-only snapshot report, the unknown bucket, the trace-family fold, the categorisation move to the contract package, and the second commit's retention job. The two declared extras (contract-package move, advisory-lock helper to `core/`) are stated in the body, so neither is undeclared creep.

🟡 `packages/api-server/src/modules/metrics/services/session-type-spend.ts:22` — the service takes an `isEnabled: () => Promise<boolean>` dep and branches on it (`:26`, `if (!(await deps.isEnabled())) return []`), so the consumer carries a capability toggle that is not its concern: `createSessionTypeSpend` cannot be read without knowing the feature is switchable, and a mis-wired flag reader is a silent empty result rather than a type error. The architecture puts the on/off question inside the dependency, configured where it is built (`infrastructure.md` → *Required Dependencies — Never Optional*).
  Fix: drop `isEnabled` from `deps` and let `breakdown` always call `readSpend`/`categorizeSessions`; move the gate into the `readSpend` adapter built at the composition root (`apps/api-server/trpc/context.ts:273`), e.g. a `createFlagGatedSpendReader({ read: metricsReader.spendBySession, isEnabled })` that returns `[]` while the `session-costs` flag is off.

🟡 `packages/api-server/src/modules/session-directory/services/session-directory-service.ts:26` — `categoryOf` is a mapper: it takes the persistence row shape `StoredSession` (`mode: string`, `type: string`), re-parses those raw strings with the pod Zod schemas, casts the results across two enum definitions, and produces the domain value `SessionCategory`. Translating persistence representations into domain objects is infrastructure work (`slice-composition.md` §3; `infrastructure.md` → *Mappers*), and leaving it in the application layer is what forces the two `as SessionMode` / `as SessionType` casts to appear in a service.
  Fix: move `categoryOf` (and the safe-parse) into `infrastructure/session-directory-repository.ts` as a `parseStoredSession` mapper so `find()` returns already-validated `{ sessionId, mode, type }` domain values, and let the service only fold them into the `Map` via `sessionCategoryOf`.

🟢 `packages/api-server/src/modules/metrics/domain/spend-by-session-type.ts:1` — the new domain file imports `SESSION_CATEGORIES` and `sessionCategoryOf`'s types from the contract package, and the classification rule itself now lives in `packages/api-server-api/src/modules/sessions/types.ts:61`. The domain layer is specified as pure TypeScript with zero external dependencies (`slice-composition.md` §2). Noted as a suggestion because the repo already does this in 26 other `modules/*/domain/` files — this PR follows the house convention, it does not start it.

🟢 `packages/api-server/src/modules/session-directory/index.ts:1` — the module's public API exports its composition root and service type, and `apps/api-server/deps.ts:48` imports `SessionDirectory` so `metrics` can be handed `sessionDirectory.categorize` at the app root; `modules.md` → *Module Public API* limits `index.ts` to domain event types and guards. Again consistent with the existing `metrics/index.ts` / `MetricsReader` pattern, and the wiring is at least an injected function port rather than a direct cross-module import.

_Both 🟡 above are layer placement on an otherwise-correct diff, so they carry into `### Findings` as one 🟢 rather than as change requests._

🟢 `packages/ui/src/modules/metrics/components/session-type-spend-bars.tsx:8` — this is the third bars component with the same body: `const max = rows[0]?.costUsd ?? 0`, a `flex flex-col gap-4` wrapper, and a `rows.map` into `<SpendBar … />`. `agent-spend-bars.tsx:7` and `model-spend-bars.tsx:8` are the same shape; only the key, the label and the trailing note differ.
  Fix: extract a `<SpendBarList rows={…} keyOf={…} labelOf={…} footer={…} />` (or a `useSpendBarRows(rows)` returning `{ key, label, pct, value }[]`) and let the three components become thin adapters over it.

🟢 `packages/ui/src/modules/metrics/lib/session-type-label.ts:5` — `SESSION_TYPE_LABELS` is keyed by `SpendCategory` and built from `SESSION_CATEGORY_LABELS`, so both the type and the source call the concept a *category* while the constant calls it a *type*.
  Fix: rename to `SPEND_CATEGORY_LABELS` in `metrics/lib/spend-category-label.ts`, matching the `SpendCategory` type it is a `Record` over.

🟢 `packages/ui/src/modules/metrics/views/usage-view.tsx:42` — `features?.["session-costs"] ?? false` is now inlined at three call sites (here, `sessions/components/sessions-sidebar.tsx:102`, and the sibling `advanced-connections` read in `connections/hooks/use-catalog-groups.ts:26`), so each consumer re-encodes both the flag key string and the default.
  Fix: add a `useFeatureFlag(id)` wrapper next to `useFeatures` in `modules/features/api/queries.ts` and call that from the views.

🟢 `packages/ui/src/modules/metrics/views/usage-view.tsx:137` — `UsageSkeleton({ showSessionTypes })` takes a boolean that exists only to switch a region of its own layout on and off. The skeleton is also the only reason the view fetches the flag at all — the ready branch already gates on `data.bySessionType.length > 0`, which is empty whenever the flag is off (`metrics/services/session-type-spend.ts:26`).
  Fix: let the skeleton read the flag itself so `UsageView` doesn't thread presentation state through a prop, or accept the extra placeholder unconditionally.

### Verdict
COMMENT — no blocker: the two warnings are a data-correctness gap in the new rollup's row filter and a cancelled coalescing timer on the new watcher path, both one-line fixes; everything else is placement and docs.

---

dam-code-guardian (approved):
## Code Guardian review — re-review at `7138399`

Delta round: `c9501aa..7138399` — one fix commit (`e1bd762`) plus a merge of `main`. Scoped to what those produced.

### Findings

🆕 🟡 **The merge changed the retention mechanism; five architecture sentences still describe the old one.** All of these are prose this PR added (`main` says "three routes", "owns no storage of its own", and never mentions a lock), so they drifted inside the PR rather than against it:
- `docs/architecture/metrics.md:53` — "guarded by an advisory lock so one replica does the delete and the others no-op". The merge deleted `core/advisory-lock.ts` and `sagas/session-directory-retention-job.ts` (both present at `e1bd762`); `retentionTick` now deletes unguarded (`session-directory/compose.ts:17-24`) and single-execution comes from the BullMQ scheduler (`bootstrap.ts:983-988`). The PR description's "the shared lock helper moved to `core/` now that two modules need it" is stale for the same reason.
- `docs/architecture/runtime-delivery.md:16` says "four routes" while `:64` still says "**Three** tRPC routes … `applyState` … and `hello` plus the artifact-touch report", `:111` opens "A **second** agent-initiated call" against three callbacks in `harness-router.ts:7-12`, and the mermaid block at `:31-34` has no session-directory edge.
- `docs/architecture/metrics.md:7` — "it mutates nothing and emits no events", in the same sentence that declares the subsystem owns the session directory, which `record()` upserts and the retention job deletes from.
- `docs/architecture/metrics.md:53` — "the one table in the platform that grows with **use**" is falsified by the table the same sentence cites as precedent (`activity_events`, same 180-day shape).
- `docs/architecture/features.md:28-29` — "All current features gate only these" is denied by the third gate kind added 10 lines below it at `:38-44`.

Fix: when a merge from the base branch replaces a mechanism the branch documented, re-read every sentence naming that mechanism and every count, "only", "never" or uniqueness claim on the pages the branch touched — a page must name the primitive the code actually uses, and an enumeration gaining a member invalidates the absolutes written around it.

Carried over from the review at `c9501aa`:

- ✅ Fixed — the last unsubscribe no longer clears the coalescing timer out from under a permanent watcher (`session-changes.ts`); verified the watcher still receives a pending delivery, a no-listener/no-watcher notify is still dropped, and the `onDemand` PTY hooks still key off viewer count alone.
- ✅ Fixed — `parseStoredSession` moved into `session-directory-repository.ts`, both enum casts gone, `find()` returns domain types.
- ✅ Fixed — `features.md` and `runtime-delivery.md` both answered the prior round's drift items (the new drift above is separate, and merge-caused).
- 🔁 Resolved as accepted — empty `session.id` rows booked to `unknown`. Your rationale checks out mechanically: `ownedApiRequests()` carries no `session.id != ''` predicate, so the totals include those rows and filtering the breakdown would make it disagree with the total above it. Recorded; not raising this again.
- 🟢 Unchanged, not re-raised — the `isEnabled` dep on `createSessionTypeSpend` (placement) and the four UI suggestions from round 1. The delta did not touch them.

🔁 Both findings from the previous review are resolved (see review at `c9501aa`). Seven new items at this HEAD, all folded into the single 🟡 above.

### Issue Fit

No linked issue — no closing keyword, no `#<n>` in the body, and the branch name carries no issue number. The description is unchanged since the prior round and the two new commits stay inside what it declares; the retention bullet is folded into the 🟡 above.

✅ No findings. Merge resolution verified line-for-line against the branch's own contribution — nothing `main`-side or branch-side was lost; retention registration is structurally identical to its `activity-retention` sibling and idempotent under at-least-once, so dropping the lock is correct in code; `podSessionTypeSchema` and `SessionType`/`SessionMode` round-trip without loss; migration `0041` renumber is clean (journal ends at idx 41, no gap or duplicate, snapshot chain links).

✅ No findings.

### Verdict

**APPROVE** — every blocker from the previous round is closed and the merge kept the branch intact. The one 🟡 is documentation-only: the merge quietly swapped the retention primitive and the pages did not follow. Worth a commit before merge, but it does not gate the code.

JanPokorny (approved):

Issue #3530, opened 2026-09-02
Runtime-state worker starved by one wedged agent: new agents wait 10–25 min for initial config and cannot reach LiteLLM


jezekra1:
## Context

Production deployment (appVersion 0.2.15), 2026-09-02 morning. A user created a new Claude Code agent (`agent-dfe25fff61fdcba8`) and it could not get any model response through LiteLLM, while their two older agents worked fine. Read-only investigation of the cluster plus the v0.2.15 source.

The new agent turned out to be healthy; it was simply never told how to reach LiteLLM. Its initial runtime state (24 env vars incl. the model-gateway wiring) sat undelivered for 23+ minutes because the apiserver's `runtime-state` worker was fully occupied by hung deliveries to **one** unrelated, wedged agent. A second new agent from another user, created a minute later, was stuck identically and got its state after 11 minutes; one of the reporter's older agents had waited 22 minutes for the same push earlier that morning.

## Problem

Three things combine:

1. **The delivery HTTP call has no timeout.** `agent-runtime-client.ts` uses a plain tRPC `httpLink` with the global `fetch`, so a non-responding agent holds the request for undici's default 300 s headers timeout, then fails with `fetch failed`.
2. **The worker is a shared pool of 16 slots with no per-agent dedup or circuit breaker.** `state-queue.ts` runs `Worker({ concurrency: 16 })`; jobs are added with no `jobId`, sweep-enqueued jobs retry 8× with exponential backoff, and `cron-sweep.ts` re-enqueues the same pending row every minute. One agent that never answers therefore accumulates dozens of concurrent 300 s requests and monopolises the pool.
3. **The agent runtime executes trigger events inside the `applyState` request.** `runtime-channel/service.ts` awaits `processEvents()` (sequential `await dispatcher.invoke(...)` per event, `event-loop.ts`) before replying. An agent with a large trigger backlog runs those prompts synchronously while the platform waits.

The wedged agent (`agent-140849f78359b83d`) has 5 schedules firing every 10 min and 30 queued `trigger` events. On every start it receives `applyState … events={trigger=30}`, starts executing them, hits its 1 GiB cgroup limit within ~20 s, its `/healthz` stops answering, the liveness probe kills the container, and the same 30 events are redelivered after restart. It has been looping since at least 2026-09-01 22:40, and after two restarts today (one manual, one liveness) it re-wedged within a minute each time.

## Evidence

Broken new agent (`agent-dfe25fff61fdcba8-0`):

```
[...]

PR #3533, opened 2026-09-02
fix(api-server): isolate runtime-state delivery per agent

jezekra1:
Refs #3530 — no closing keyword on purpose: the ack-before-run item below stays open, so the issue stays open after merge.

## What went wrong

Agent `agent-dfe25fff61fdcba8` never got its initial runtime state — including its LiteLLM model-gateway wiring — so for 23+ minutes it looked like a broken agent. It wasn't: its 24 env vars were sitting undelivered in `runtime_state_outbox`.

Every one of the 16 delivery worker slots was held by a hung `applyState` to a *different* agent, `agent-140849f78359b83d`, which crashloops on its 1 GiB cgroup limit ~20s after each start. Five schedules firing every 10 minutes had built up 30 queued trigger events; each restart replayed all 30, liveness killed it again, and the connections to it hung for undici's full 300s `headersTimeout`. Production evidence: 336 ztunnel connections at 302–303s with `bytes_recv=0`, and `fetch failed` at exactly 3.2/min = 16 slots ÷ 300s.

## What this changes

Six independent causes, each fixed on its own — no single one of them is sufficient:

| # | Cause | Fix |
|---|---|---|
| 1 | `applyState` had no deadline | 60s `AbortSignal.timeout`, threaded through tRPC's `httpLink` (`APPLY_STATE_TIMEOUT_MS`) |
| 2 | Enqueues unbounded, retries stacked per agent | BullMQ `deduplication` keyed by agent id, `keepLastIfActive` — at most 1 active + 1 waiting. A separate `:boot` key keeps a plain job from swallowing the `hello` catch-up that retries until Ready |
| 3 | Sweep's batch of 100 filled with stopped agents (`skipped 98 pending rows`) | `listRetryable` is uncapped; bounded naturally by one row per agent, readiness served from the informer-backed cache, re-enqueued in one `addBulk` |
| 4 | Trigger fires only expired on the 1h TTL | A fire now expires at its schedule's next occurrence, capped by the TTL — at most one pending fire per schedule |
| 5 | 16 worker slots, so a slot was the scarce resource | Concurrency 16 → 256 |
| 6 | Postgres pool left at the client default of 10, now serving 256 handlers plus the request path | Sized explicitly (`DATABASE_POOL_MAX`, default 32) |

### On isolation

The bar here is that one agent's delivery never waits on another's — not blocked, not delayed. Stage by stage: enqueue is a per-agent Redis add (ms); the pre-call Postgres reads are ms and their locks release before the call; the HTTP call is the only long hold, one socket per pod origin and now deadline-bounded; the post-call outbox write takes that agent's row lock only; the apply itself serializes per agent, inside the agent.

Two things are genuinely shared, and they behave differently:

- **The worker slot.** #2 + #5 together move the number of simultaneously wedged agents needed before anyone else waits from **1 to 128**.
- **The Postgres pool**, which is far scarcer — but a wedged handler holds no connection. Its reads complete before the request goes out, the outcome is written after it returns, and nothing opens a transaction across the call, so what the pool has to cover is handlers *querying* at an instant, not handlers parked on a slow agent. Drizzle routes every statement through `sql.unsafe()`, so a parameterized query is unprepared and postgres.js parks that connection until the rows come back rather than pipelining another onto it — one statement per connection at a time. A cold burst of 256 jobs therefore queues ~224 statements of 1–3ms work behind a fair pool-wide FIFO and drains in tens of milliseconds. #6 sizes for that instead of leaving it at 10; sizing it for the slot count instead would put a single replica past a stock `max_connections` of 100.

That invariant is now written down in `docs/architecture/runtime-delivery.md` alongside the existing ones.

## Deliberately not done

- **Ack `applyState` before running trigger events** (issue item 4) — needs a new agent→server settle route, and `trigger-session-driver.start()` already returns after `session/new` + fire-and-forget `session/prompt` rather than awaiting the turn, so the ack would not buy much. Nor is it a reorder: `worker-handler.ts` records the outcome from `settledEvents`, `appliedVersion`, `failures` and `harnessConfigCurrent` on the response, so acking early needs an agent→server settle route — a protocol change across two packages, not a hotfix. This is the one issue item left open, and the reason the link above is `Refs` rather than `Fixes`.
- **Partition the worker** (item 3) — unnecessary once one agent is bounded to one active job per dedup key.
- **Per-agent circuit breaker** (the second half of item 2) — the deadline plus dedup already bound the damage; a breaker adds state to tune without a demonstrated need.

## Verification

- `mise run api-server:check` — tsc, lint, prettier, event-coverage (45 events): clean
- `mise run api-server:test` — 139 files, 1303 tests pass
- `mise run db:check`, `mise run common:check:comment-types`, `mise run docs:check:doc-size` — OK

New unit tests: `runtime-agent-runtime-client.test.ts` stands up a real `node:http` server that accepts and never answers, asserting the deadline rejects; `runtime-state-queue.test.ts` pins both dedup keys; `scheduler-runner.test.ts` gains two cases for expiry anchoring (next occurrence, and TTL cap when the next occurrence is far); `schedules-recurrences.test.ts` covers `triggerExpiry` directly, including the rrule-with-`BYSECOND` case where the next occurrence lands before the fire instant; `runtime-cron-sweep.test.ts` pins that scattered readiness failures do not latch the sweep into calling every remaining row unknown.


packages/api-server/src/modules/runtime-delivery/infrastructure/state-queue.ts
```diff
@@ -15,28 +20,41 @@ export interface StateQueue {
 
 const READY_RECHECK_MS = 1_000;
 const READY_RECHECK_ATTEMPTS = 120;
+const DELIVERY_CONCURRENCY = 256;
```
dam-code-guardian: 🟡 **Warning:** 256 delivery slots draw on a Postgres pool of 10.

`createDb` (`packages/db/src/client.ts`) calls `postgres(url, ...)` with no `max`, so the pool takes the postgres.js default of 10 connections, and `bootstrap.ts:188` hands that one `db` to both `composeRuntimeDelivery` and the request-handling modules. Each delivery job makes four or more round trips around the HTTP call — `getRow`, `agentsRuntimeRepo.get`, three `stateBuilder` reads plus `pendingEvents`, then `recordOutcome` — so past ~10 concurrent jobs the scarce resource moves from the worker slot to the pool, and delivery jobs then queue against the same 10 connections that serve API requests. That is the condition the new invariants bullet rules out when it says a handler ties up nothing shared, and it caps the "1 to 128" threshold below 128.

**Fix:** size the pool for its consumers — pass an explicit `max` to `postgres()` in `createDb`, at or above the delivery concurrency plus request-handler headroom, and check it against the server's `max_connections`.

packages/api-server/src/modules/runtime-delivery/infrastructure/outbox-repo.ts
```diff
@@ -76,7 +75,7 @@ export interface OutboxRepo {
     },
     maxAttempts?: number,
   ): Promise<ApplyTransitions>;
-  listRetryable(maxAttempts: number, limit: number): Promise<OutboxRow[]>;
+  listRetryable(maxAttempts: number): Promise<OutboxRow[]>;
```
dam-code-guardian, line 78: 🟡 **Warning:** the scan is now uncapped, and the cache hit that bounds its cost is not guaranteed.

Removing `limit` is right for the reported symptom — `skipped 98 pending rows` was the batch filling with stopped agents. The new bound rests on the doc's "the running-state read is a cache hit in steady state", and `agent_id` being the outbox primary key does bound the row count to one per agent. But `agent-state-cache.ts` serves `deps.live.getCustomObject` whenever `synced` is false, and `abandon()` clears that flag with a restart backoff from 5s to 60s. During a watch outage every retryable row becomes a live k8s GET each minute, 16 lanes at a 5s deadline (`cron-sweep.ts:22-23`), and an `unknown` result re-enqueues the row anyway. The removed cap is what used to bound that fan-out.

**Fix:** keep a bound on the repository read — restore a `limit` argument fed by a sweep-owned batch size, sized above the live agent count so stopped agents cannot fill it.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — 💬 Code Review @ `d0f5d6d`

## PR #3533: fix(api-server): isolate runtime-state delivery per agent
**Author:** jezekra1 | **Branch:** fix/3530-runtime-worker-starvation → main | **Changes:** +188 −43 (10 files)

### Summary
Five independent fixes for the delivery-worker starvation in #3530: a 60s `applyState` deadline, per-agent BullMQ deduplication with a separate `:boot` key, an uncapped retryable scan, fire expiry anchored at the next occurrence, and worker concurrency 16 → 256.

### Findings
- 🟡 **Warning:** worker concurrency 256 exceeds the shared Postgres pool (`packages/api-server/src/modules/runtime-delivery/infrastructure/state-queue.ts:23`)
  **Fix:** size the pool for its consumers — pass an explicit `max` to `postgres()` in `createDb`, at or above the delivery concurrency plus request-handler headroom.
- 🟡 **Warning:** the retryable scan lost its bound while the readiness read can still miss the cache (`packages/api-server/src/modules/runtime-delivery/infrastructure/outbox-repo.ts:78`)
  **Fix:** keep a bound on the repository read — restore a `limit` argument fed by a sweep-owned batch size, sized above the live agent count so stopped agents cannot fill it.
- 🟡 **Warning:** `docs/architecture/runtime-delivery.md:235` still calls running-state "a live read", but line 240 in this PR rewrites the same read as a watch-backed cache with a live read only as fallback. The page describes one read two ways.
  **Fix:** reword line 235 to "Running-state is read through the agents subsystem (below) and can fail", and drop "live". The unknown-vs-not-running argument that follows stays unchanged.
- ✅ **Looks good:** the two dedup keys are correct at the pinned bullmq 5.76.10 — `keepLastIfActive` bounds a key to one active plus one waiting, so the separate `:boot` key is what keeps a plain dispatch from swallowing the `hello` catch-up. tRPC `httpLink` v11.16.0 forwards `signal`, so the deadline does bound the call.

### Documentation drift
- 🟡 See the `runtime-delivery.md:235` finding above.
- 🟢 `docs/architecture/agent-lifecycle.md:86` — the Schedule-driven bullet still says only "the event's TTL bounds how stale a fire can land", while line 99 now carries the next-occurrence rule.
- 🟢 `docs/architecture/budgets.md:71` — says a committed trigger "redelivers once room frees or expires at its TTL"; a fire now normally expires at the schedule's next occurrence, with the TTL as the cap.
- 🟢 `docs/architecture/runtime-delivery.md:225` — citing "the transport's default of five minutes" pins a library-level default on the page. The durable point is that the call has its own deadline.

### Issue fit
Linked issue #3530, closing keyword `Fixes #3530`.
- 🟢 One of the issue's four expected behaviours — `applyState` acks before running trigger events — is not delivered, and the PR declares it under "Deliberately not done". The rationale checks out: `packages/agent-runtime/src/modules/acp/services/trigger-session-driver.ts:54` submits the prompt with `caller.notify`, so the turn already does not block the reply, and the new deadline bounds what remains. Consider a plain reference plus a follow-up issue rather than the closing keyword.
- 🟢 `packages/api-server/src/modules/runtime-delivery/infrastructure/outbox-repo.ts:312` — issue item 6 asked `listRetryable` to exclude non-Ready agents; the PR removes the limit instead and filters readiness in the sweep. The goal is met and the deviation is declared.

### TypeScript engineering
- 🟡 Uncapped `listRetryable` — merged into the second finding above.
- 🟢 `packages/api-server/src/modules/runtime-delivery/infrastructure/agent-runtime-client.ts:14` — `fetch` and `timeoutMs` are both optional and defaulted inside the adapter, so the deadline knob exists only for tests while `compose.ts:104` builds the client unconfigured.
- 🟢 `packages/api-server/src/modules/runtime-delivery/infrastructure/state-queue.ts:23` — `DELIVERY_CONCURRENCY` is a module constant, so the concurrency cannot be tuned or rolled back at the composition root.
- 🟢 `packages/api-server/src/modules/schedules/services/scheduler-runner.ts:29` — `triggerExpiry` is pure recurrence arithmetic in the application layer, while `nextFireAt` and the other recurrence rules live in `schedules/domain/recurrences.ts`.

### Verdict
COMMENT — the five fixes are sound and verified against the pinned libraries; two bounds need attention before the isolation invariant in the docs holds as written.

---

jezekra1:
Addressed in 3f885a7c. Thanks — the pool finding is the real one; I'd audited what a wedged delivery *holds* (nothing: no transaction spans `applyState`) and never counted how many of the shared thing there are.

**Pool (confirmed, fixed).** `postgres@3.4.9/src/index.js:449` defaults `max` to 10 and `createDb` passed no override, so 256 slots drew on 10 connections shared with the request path. `createDb` now takes an explicit `max`, surfaced as `apiServer.db.poolMax` (32).

I did **not** size it "at or above the delivery concurrency plus headroom", though — that arithmetic doesn't survive contact with the server. `max_connections` is unset in the chart, so Postgres takes its default of 100, and that budget is shared with Keycloak (whose Quarkus pool is also unset, i.e. `db-pool-max-size` 100) on a 512Mi/500m server. A 256+ pool per replica would exceed it on a single replica and fail outright past two. The right sizing is the *concurrent-query* demand, not the slot count: a handler borrows a connection for ~6 short queries and never across the call to the agent, so the pool has to cover the handlers querying at any instant, not the ones parked on a slow agent. 32 is ~2.5× the steady-state demand at 256 slots and 3× today's capacity. Both numbers are now deployment values, and the per-replica multiplication against `max_connections` is written down on the page.

Worth flagging separately: two unset pool maxima against an unset `max_connections` is a live footgun independent of this PR. Happy to file that.

**Scan limit (declined, bounded differently).** Restoring a `limit` is not safe here, and the incident is the evidence: `runtime_state_outbox` has no fair ordering to page by, so a cap plus Postgres's stable heap order is exactly what hid `agent-dfe25fff61fdcba8` behind 98 stopped agents for 23 minutes. `last_enqueued_at ASC` doesn't fix it either — a stopped agent's row keeps its old timestamp and holds the head of the batch permanently.

So I bounded the expensive half instead, which is what the finding is actually about: the sweep now stops issuing readiness checks once a round of them fails, and calls the remaining rows unknown. During a watch outage the live-GET fallback costs one round of reads per tick instead of one per row, while the scan stays uncapped — one row per agent, which is what bounds it. Test: `runtime-cron-sweep.test.ts`, "stops checking after a round of failures and re-enqueues the rest".

**Docs (all four fixed).** `runtime-delivery.md` no longer calls the running-state read "live" and no longer cites the transport's five-minute default; `agent-lifecycle.md:86` and `budgets.md:71` no longer have a fire expiring on its TTL alone. The pool/slot relationship is on the page too.

**Suggestions taken.** `triggerExpiry` moved to `schedules/domain/recurrences.ts`. Worker concurrency is injected at the composition root rather than a module constant.

**`Fixes #3530` kept.** The starvation is fixed and verified; ack-before-run was one of the issue's *proposed* remedies, not a behaviour left broken — as you note, `trigger-session-driver.ts:54` already submits with `caller.notify`, so the turn doesn't block the reply. I'll open a follow-up for the settle route rather than hold the hotfix.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — 💬 Re-review @ `3f885a7`

## PR #3533: fix(api-server): isolate runtime-state delivery per agent
**Author:** jezekra1 | **Branch:** fix/3530-runtime-worker-starvation → main | **Changes:** +304 −53 (20 files) · since last review: +126 −20 (15 files)

### Changes since last review
- ✅ **Fixed** — the pool now covers its consumers explicitly and end to end: `createDb` takes `max` (`packages/db/src/client.ts:19`, `DEFAULT_DB_POOL_MAX = 32`), `config.ts` exposes `DATABASE_POOL_MAX` / `RUNTIME_DELIVERY_CONCURRENCY`, `bootstrap.ts:188` is the only `createDb` caller and passes it, and both land as Helm values. `migrate.ts` keeps its own single-connection client.
- ✅ **Answered with a mechanism** — the retryable scan stays uncapped. `max_connections` is unset cluster-wide (Postgres default 100, shared with Keycloak), so 256 per replica fails past two replicas: sizing on concurrent-query demand rather than slot count is the right call, and 32 is what the handlers actually hold. On the scan: a cap plus heap order is what hid `agent-dfe25fff61fdcba8`, and `last_enqueued_at ASC` doesn't give fair ordering either — bounding the readiness fan-out instead addresses the cost without reintroducing the starvation.
- ✅ **Fixed** — `runtime-delivery.md` no longer describes one read two ways; `agent-lifecycle.md:86`, `budgets.md:71` and the five-minute transport default are all corrected, and all three pages carry a current `Last verified`.
- ✅ **Taken** — `DELIVERY_CONCURRENCY` is now injected at the composition root (`compose.ts` → `state-queue.ts`), and `triggerExpiry` moved to `schedules/domain/recurrences.ts:67`.

### Findings
- 🟡 **Warning:** the new sweep latch trips on 16 *cumulative* unknown reads, not on a round of them — `unknowns` is never reset by a successful check (`packages/api-server/src/modules/runtime-delivery/services/cron-sweep.ts:81`). Its own scenario comment and `runtime-delivery.md:242` both describe "a tick's worth of failures"; with the scan uncapped, a long tick that hits 16 scattered failures stops checking and calls every remaining row unknown, which re-enqueues a job for each stopped agent. The threshold is also `checkConcurrency` itself, so tuning the in-flight bound retunes the outage threshold, and `compose.ts` never injects it — the value that decides this is settable only from the test harness.
  **Fix:** reset on success (`if (check.state !== "unknown") { unknowns = 0; continue; }`) so only a genuine outage round short-circuits, give the threshold its own dep rather than reusing `checkConcurrency`, and wire both from `composeRuntimeDelivery` the way `deliveryConcurrency` now is.
- 🟡 **Warning:** the PR body no longer matches the change it describes. "The worker slot was the one shared resource" is contradicted by this commit and by the paragraph it adds at `runtime-delivery.md:227` ("Slots are not the only shared resource, and the other one is scarce"), and the change table has no row for the pool. That paragraph also transcribes both Helm keys inline, which `docs/guidelines/documentation-guidelines.md:41` puts under omit.
  **Fix:** add the pool sizing as a sixth row of the table and reword the "On isolation" sentence to name the Postgres pool as the second shared resource; in the page, keep "both are deployment values" and link out to `deploy/helm/platform/values.yaml` as the field-level source, as `docs/architecture/skills.md:72` does.
- 🟢 **Suggestion:** `triggerExpiry`'s `byNext > firedAt.getTime()` branch is dead — `nextFireAt` returns an instant strictly after its `from` on both the cron (`cron.next()`) and rrule (`rule.after(cursor, false)`) paths, and the caller passes the same `firedAt` (`packages/api-server/src/modules/schedules/services/scheduler-runner.ts:53`). The body reduces to `new Date(Math.min(byTtl, next?.getTime() ?? Infinity))` (`packages/api-server/src/modules/schedules/domain/recurrences.ts:74`).
- 🟢 **Suggestion:** the sweep still enqueues serially, one awaited Redis round-trip per row, ahead of `deleteExpiredEvents` — pre-existing, but the loop is unbounded now that the scan is (`packages/api-server/src/modules/runtime-delivery/services/cron-sweep.ts:114`). `addBulk`, or the lane pool `checkAll` already uses, keeps the expired-event drop from queueing behind it.
- 🟢 **Suggestion:** `DEFAULT_DELIVERY_CONCURRENCY` sits in `config.ts:6` while the queue that consumes it owns its other constants and the parallel `DEFAULT_DB_POOL_MAX` is exported from the package that owns the pool; next to `READY_RECHECK_ATTEMPTS` in `state-queue.ts` would make the two knobs symmetric. Relatedly, `createDb(url, tls?, pool?)` has grown a third positional argument (`packages/db/src/client.ts:19`) — a single options object would absorb the next deployment-tunable setting.
- 🟢 Carryover, unchanged: `agent-runtime-client.ts:14` — `fetch` and `timeoutMs` stay optional with no production caller, so the deadline knob is still test-only while its siblings were lifted to the composition root.

### Documentation drift
- 🟡 See the `runtime-delivery.md:227` half of the PR-body finding above. Otherwise the new pool/concurrency behaviour is documented and the four prior items are fixed.

### Issue fit
Linked issue #3530, closing keyword `Fixes #3530`.
- 🟢 Carryover: the ack-before-run item is still undelivered and still declared, and the follow-up issue offered in the thread isn't filed yet — no open issue references #3530. Either file it before merge or drop to `Refs #3530`.
- 🟢 Carryover: issue item 6 asked `listRetryable` to exclude non-Ready agents; readiness is filtered in the sweep instead. Declared, and the item's goal is met.
- 🟢 The pool sizing is in scope — it is fallout of this PR's own 16 → 256 raise, not creep — but it is the one change the PR body doesn't declare (see above).

### TypeScript engineering
- 🟡 The sweep latch — merged into the first finding above.
- 🟢 The dead `triggerExpiry` branch, the serial enqueue loop, the constant placement / `createDb` signature, and the test-only deadline knob — all listed above.

### Verdict
COMMENT — the round closes all three prior warnings at their cause, two with a mechanism worth keeping. What's left is the latch this commit introduces, whose threshold doesn't match what its comment and the page claim, and a PR body that now argues against its own change.

---

jezekra1:
Round 2 addressed in ad2c0c13. Four taken, four declined with reasons.

### Taken

**The sweep's unknown-latch never reset — correct, and the worst of the batch.** `unknowns` counted for the whole tick, so it was a cumulative total, not the outage signal it was meant to be. With the scan now uncapped (#3), a long tick with 16 transient errors scattered across hundreds of rows would trip the threshold and mark every *remaining* row unknown — and unknown re-enqueues anyway, so the sweep would have fired a job for every stopped agent in the outbox. Exactly the amplification this PR exists to remove. A success now clears the counter, so the threshold means what the comment says: one failure per lane with no success in between.

**Bulk re-enqueue.** Removing the `LIMIT 100` made the serial `enqueue` loop an N-round-trip walk to Redis. `StateQueue` gained `enqueueMany`, and the sweep sends one `addBulk`. `Job.createBulk` runs each job's own `addJob` inside a Redis pipeline, so per-job dedup options still apply — the `deduplication` keys are unaffected.

**`createDb` options object.** Agreed — pool sizing was a third positional after TLS. Now `createDb(url, { tls, poolMax })`.

**Doc page.** Dropped the inline Helm keys for a link to `values.yaml` (documentation-guidelines: omit field names, link out for volatile content), and corrected the latch sentence to match the fixed code: "once a full round of reads has failed with none succeeding in between".

### Declined

**A separate injected threshold, distinct from `runningCheckConcurrency`.** After the reset fix, threshold = lane count *is* the definition: every lane failed once with no success between them, i.e. a full round of reads came back bad. That's the outage the fallback is for. A second knob would be a number nothing sets, tunable to values that mean nothing (a threshold below the lane count latches on a partial round; above it latches on rounds plus stragglers).

**The `triggerExpiry` guard is not dead.** `nextFireAt` truncates the fire instant to the minute before evaluating the rule (`wallFrom.setUTCSeconds(0, 0)`, then `rule.after(cursor, false)`), so an RRULE carrying `BYSECOND` answers with an occurrence *inside* the minute already fired. Ran the installed `rrule` against `firedAt = 2026-09-02T10:00:30Z`:

| rule | `nextFireAt` |
|---|---|
| `FREQ=MINUTELY;BYSECOND=15` | `10:00:15Z` |
| `FREQ=HOURLY;BYSECOND=15` | `10:00:15Z` |
| `FREQ=DAILY;BYHOUR=10;BYSECOND=15` | `10:00:15Z` |

`validateRRule` accepts any parseable RRULE from the tRPC write path, so this is reachable user input, and without the guard `expiresAt` lands 15s in the past — the event is expired at birth and the fire is silently dropped. I did restructure it as an early return so the intent reads, and added a test for that exact rule.

**`DEFAULT_DELIVERY_CONCURRENCY` staying in `config.ts`.** `config.ts` imports only `api-server-api`, `db`, `zod`, and `./duration.js` — nothing from `./modules`. Moving the default next to the queue would make the config layer depend on a module's infrastructure, which is the wrong direction. `DEFAULT_DB_POOL_MAX` lives in `db` because that's a package config already depends on.

**`agent-runtime-client.ts:14` is a test seam, not config.** The parameter exists so the unit test can inject `fetch` and a short deadline; the default is the production value and no deployment path sets it. Promoting it to Helm would add a knob with no operator meaning.

### Verification

`api-server:check` clean · `api-server:test` 139 files / **1303** tests pass · `db:check` clean · `common:check:comment-types` OK · `docs:check:doc-size` OK (`runtime-delivery.md` 34 769 / 40 000).

PR body updated: pool sizing is now row 6 of the table, and "On isolation" names both shared resources instead of claiming the slot was the only one.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — 💬 Re-review @ `ad2c0c1`

## PR #3533: fix(api-server): isolate runtime-state delivery per agent
**Author:** jezekra1 | **Branch:** fix/3530-runtime-worker-starvation → main | **Changes:** +405 −58 (21 files) · since last review: +117 −21 (9 files)

### Summary
Complete re-review at the `code-guardian-review` label. Round 3 closes both prior warnings at their cause: the sweep latch now resets on a successful check, and the PR body plus the page carry the pool as the second shared resource. The four declines are each backed by evidence, and one of them corrects a wrong finding of mine.

### Changes since last review
Previous HEAD: `3f885a7` (2026-09-02T14:25:13Z) — verdict COMMENT

- ✅ **Fixed** — the latch means what its comment says: a non-unknown check resets `unknowns` (`cron-sweep.ts:80`), so only a full round of failures with no success between them trips `unavailable`. The new scenario test pins it with two lanes over eight rows.
- ✅ **Fixed** — the serial per-row enqueue is one `addBulk` via `StateQueue.enqueueMany` (`state-queue.ts:60`), ahead of `deleteExpiredEvents`. Per-job dedup survives: I read `Job.createBulk` at the pinned bullmq 5.76.10 — it calls each instance's own `addJob` inside the pipeline, so `stateJobOptions`' `deduplication.id` still applies per job.
- ✅ **Fixed** — `createDb(url, { tls, poolMax })` (`packages/db/src/client.ts:20`); `DbPoolOptions` → `DbOptions` in the barrel, and `bootstrap.ts:188` is the only production caller.
- ✅ **Fixed** — the page links to `values.yaml` instead of transcribing the Helm keys, and the latch sentence now reads "once a full round of reads has failed with none succeeding in between". PR body row 6 carries the pool, and "On isolation" names both shared resources.
- ✅ **Withdrawn — my finding was wrong.** The `triggerExpiry` guard is not dead. I reproduced it against the installed rrule 2.8.1: `nextFireAt` truncates the fire instant to the minute (`recurrences.ts:47`), so `FREQ=HOURLY;BYSECOND=15` answers `10:00:15Z` for a `10:00:30Z` fire. Without the guard `expiresAt` lands in the past and the fire is dropped at birth. The early-return restructure and its test are the right outcome.
- 🔁 **Still present:** the deadline knob on `agent-runtime-client.ts:14` stays optional with no production caller. Declined as a test seam.

### Findings
- 🟢 **Suggestion:** the sweep's latch state (`unknowns`, `unavailable`, the reset, the trip at `checkConcurrency`) is deterministic policy held in mutable locals inside the application service (`packages/api-server/src/modules/runtime-delivery/services/cron-sweep.ts:66`). `runtime-delivery/domain/outbox-progress.ts` already hosts comparable pure decisions, and this PR moved `triggerExpiry` to `schedules/domain/` for exactly that reason.

The threshold decline holds up: after the reset, threshold = lane count *is* the definition of a failed round, and a second knob would have no meaning to set it to. The `config.ts` placement decline is verified — `config.ts` imports `api-server-api`, `db`, `zod` and `./duration.js`, nothing from `./modules`, so moving the default into the queue would invert the dependency.

- 🟢 **Suggestion:** the pool ceiling — 32 per replica, shared with the request path, multiplied by replicas against the server's `max_connections` — is documented only on the runtime-delivery page (`docs/architecture/runtime-delivery.md:227`), while `persistence.md` owns the Postgres substrate and says nothing about pooling. A reader on another subsystem would not find the ceiling where it now lives. Consider the substrate page owning the constraint, with a one-liner and cross-link here.

### Issue Fit
Linked issue #3530, closing keyword `Fixes #3530`.
- 🟡 **Warning:** `Fixes #3530` still closes the issue with one of its four expected behaviours undelivered — `applyState` acking before trigger events run. The diff touches no `packages/agent-runtime` source, so `runtime-channel/service.ts:73` still awaits `processEvents()` in the request. The body declares it and says it wants its own issue; I searched all issues and none references #3530, so merging closes the issue with that item unfiled.
  **Fix:** a closing keyword needs every acceptance criterion met — either file the follow-up issue and link it from the body, or drop to `Refs #3530`.

The skill rated this 🔴. I hold it at 🟡: the gap is declared in the body, the rationale was verified last round, and what is missing is a tracking issue, not code. It still blocks — it is the one item that outlives the merge.

- 🟢 **Suggestion:** `AgentRuntimeClientOpts`' optional `fetch` and `timeoutMs` (`packages/api-server/src/modules/runtime-delivery/infrastructure/agent-runtime-client.ts:14`) leave the adapter's only behavioural knob configured nowhere in production — `compose.ts:105` passes neither. Carryover, declined as a test seam; the decline is reasonable, so this stays a suggestion rather than a repeat 🟡.
- 🟢 The latch's placement — listed above.

### Verdict
COMMENT — the round fixes every code finding at its cause and refutes one of mine with evidence I reproduced. What remains is not in this diff: the closing keyword promises an issue item the PR deliberately does not deliver, and no follow-up issue exists yet.

---

jezekra1:
Keyword dropped to `Refs #3530`. No follow-up issue filed — the deferred item is declared in the body instead, and the issue stays open after merge so it isn't lost.

Recording why the code is unchanged, since the finding is about scope rather than behaviour. The starvation is fixed at all three of its multiplicands, and the reported symptom clears earlier than the ack would move: inside the agent, `applyState` dispatches contributions (`runtime-channel/service.ts:68`) *before* it runs events (`:73`), so a new agent has its model-gateway wiring the moment the request arrives, however long the events take afterwards. Acking early changes when the **server** learns delivery succeeded, not when the agent becomes usable.

It also isn't a reorder. `worker-handler.ts:101-141` builds the settle from `settledEvents`, `appliedVersion`, `failures` and `harnessConfigCurrent` on the response, so acking before events run needs an agent→server settle route — a protocol change across two packages.

What the item is genuinely worth, for whoever picks it up: it removes the last place where agent-side work sets the server's hold time. Worst-case worker throughput today is 256 slots ÷ 60s ≈ 4.3 deliveries/s if every delivery ran events to the deadline, against thousands/s when they return in milliseconds. The sharpest case isn't triggers — `trigger-session-driver.ts:54` submits with `caller.notify`, so no model turn is awaited — it's `workspace-seed`, whose git clone is fully awaited (`workspace-seed-plugin.ts:45`). A clone over a minute times out the delivery, then self-heals on the next sweep tick (`.git` exists → "already seeded, skipping"), having burned a slot and a retry. Nothing is lost either way: `listRetryable`'s `lastSettledVersion < version` clause has no attempt cap, so state delivery retries until it lands.

Re-review label re-applied.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Re-review @ `ad2c0c1`

## PR #3533: fix(api-server): isolate runtime-state delivery per agent
**Author:** jezekra1 | **Branch:** fix/3530-runtime-worker-starvation → main | **Changes:** +405 −58 (21 files)

### Summary
Complete re-review at the `code-guardian-review` label, answered by an edited description — the code is unchanged at `ad2c0c1`. The body drops `Fixes #3530` to `Refs #3530`, which resolves the last blocking finding at its cause: the issue now stays open with its one deferred item declared.

### Changes since last review
Previous HEAD: `ad2c0c1` — description edited, no new commits — verdict COMMENT

- ✅ **Fixed:** the closing keyword is now `Refs #3530`, and `closingIssuesReferences` returns empty — the PR closes nothing. Merging no longer closes the issue with item 4 (ack `applyState` before running trigger events) undelivered.
- 🔁 **Still present:** the deadline knob on `agent-runtime-client.ts:14` stays optional with no production caller. Declined as a test seam; 🟢.

### Findings

The added comment on the deferred item checks out in the tree, on both halves: `runtime-channel/service.ts:69` dispatches contributions before `:73` runs events, so a new agent has its gateway wiring on request arrival; and `worker-handler.ts:101-141` builds the settle from `settledEvents` / `appliedVersion` / `failures` / `harnessConfigCurrent` on the response, so an early ack does need an agent→server settle route.

- 🟢 **Suggestion:** the Postgres pool ceiling is documented only on `docs/architecture/runtime-delivery.md:227`, while `persistence.md` owns the substrate and carries the `max_connections`-shared-with-Keycloak detail (`docs/architecture/persistence.md:11`) that a per-replica pool multiplies against. Consider a line there with a cross-link back.
- 🟢 **Suggestion:** the sequence-diagram note at `docs/architecture/agent-lifecycle.md:41` still reads "undelivered events expire after a TTL", the pre-change rule, while the prose at line 99 now carries next-occurrence-capped-by-TTL. Pre-existing on `main` and untouched by this PR — a separate follow-up, not this diff's debt.

### Issue Fit
✅ No findings. Items 1, 2 (dedup half), 5 and 6 are delivered; items 3, 4 and the circuit-breaker half of 2 are declared under a non-closing reference, which is no longer a scope gap. No undeclared creep — pool sizing, concurrency, docs and the sweep changes are all in the body's table. All three edited pages stay under the 40,000-char size gate.

- 🟢 **Suggestion:** `AgentRuntimeClientOpts`' optional `fetch` / `timeoutMs` (`packages/api-server/src/modules/runtime-delivery/infrastructure/agent-runtime-client.ts:14`) are set nowhere in production — `compose.ts:105` passes neither. Known carryover, declined as a test seam.

### Verdict
APPROVE — the edited description closes the round's only blocking finding at its cause, and the four rounds' code findings are all resolved. What is left is one carryover suggestion and two doc-placement suggestions, none of them blocking.

---

PR #3546, opened 2026-09-03
fix: queued prompts survive a reload and can be sent again

### LINKED ISSUE #3264: UI - queued messages merge into one bubble after a reload
## Problem

Send several messages while the agent is still replying, then reload the page. They come back as a single bubble with the text run together — `hihiihihi` — instead of separate messages.

The session log is correct. Only the replayed view is wrong, and reloading again does not fix it.

## Steps to reproduce

1. Send a prompt that takes a while to answer.
2. While it runs, send several short messages.
3. Reload the page.

## Cause

`appendQueuedUser` in `packages/ui/src/modules/acp/session-projection.ts` merges an incoming queued user chunk into the previous user bubble whenever the list ends with a user bubble followed by an empty queued assistant placeholder. That rule exists for #703, where one prompt arrives as several content blocks.

Two different things produce the same frames: another block of one prompt, and a separate queued prompt. Both arrive as `user_message_chunk` with `_meta.queued === true`, consecutively, with no agent content between them. The merge cannot tell them apart.

The sibling function `appendOrExtendUser` avoids this by only extending a bubble whose `messageId` matches. `appendQueuedUser` ignores `messageId`, and adding that check would not help: nothing in the platform sets the field. It appears only in that one read and in test fixtures.

Live sends are unaffected — `sendPrompt` builds a bubble per send. The merge runs only on replay.

## Fix

The queued user chunk needs a per-prompt marker so the merge applies within one prompt only. `promptId` is already minted per send and travels in `_meta.platform`.

Alternative: drop the merge and let a multi-block prompt render as separate bubbles. That reintroduces a smaller version of #703.

jjeliga:
Closes #3264

- fixed queued messages collapsing into one bubble after a reload
- prompts that never reach the agent are saved on the agent pod and shown in the chat as not delivered, with retry, copy and delete actions
- the browser keeps its own copy while the pod is unreachable and hands it over on the next session load
- a retried or deleted prompt no longer appears twice after a reload
- a queue whose last client left is kept for a grace window before its prompts are recorded; window length is configurable via QUEUE_PARK_MS
- undelivered copies are removed from the browser on sign-out or user switch
- replaced the old queued-message disconnect error with the mechanism above
- updated architecture docs (agent-lifecycle, persistence, platform-topology)


dam-code-guardian (changes requested):
## PR #3546: fix(agent-runtime,ui): keep queued prompts through a disconnect
**Author:** jjeliga | **Branch:** fix/3264-queued-messages-non-merge → main | **Changes:** +1266 −196 (24 files)

### Summary

Queued prompts stop belonging to the socket that sent them. `dropChannel` becomes `onDetached`, which parks a queue for 90s instead of discarding it; every remaining drop route announces itself with a cause, and the prompts land in a new pod-side `undelivered-prompts` document that survives session teardown and restarts. The UI grows an undelivered bubble with retry/copy/delete, plus a `localStorage` copy for sends that never reached the pod. `heldPrompts` and `QUEUED_LOST_MESSAGE` are deleted.

The park-and-record design is sound and the runtime side holds up. What does not hold up is the seam between the browser's local copy and the wire contract, and the UI error surface the PR replaced.

### Findings

- 🔴 **Critical:** the browser's held sends are dropped on handover — `packages/ui/src/modules/sessions/lib/undelivered-store.ts:11`
  `persistedUndeliveredSchema` stores `text: z.string()` and no `blocks`. `handOverUndelivered` sends those records verbatim to `platform/recordUndelivered` (`acp-session-ops.ts:172`), where `extractUndeliveredPrompts` parses them with `platformUndeliveredPromptSchema` — which has no `text` field and defaults `blocks` to `[]`. The prompt is recorded with the user's text gone. The runtime answers `result: {}` unconditionally (`acp-runtime.ts:696`), so the client's `.then(() => clearUndelivered(localKey))` fires and deletes the only remaining copy. Every send a client held while it could not reach the pod is silently emptied on reconnect — the exact loss this PR exists to stop.
  **Fix:** persist and hand over `blocks: PromptBlock[]` from `platformUndeliveredPromptSchema` rather than a local `text` field, and derive the stored type from the contract schema so one schema validates both sides of the wire.
- 🔴 **Critical:** `SendErrorCard` is now unreachable, and with it the tested failure surface — `packages/ui/src/modules/sessions/components/chat-message.tsx:172`
  The card renders only under `error && isAssistant`. After this PR the two producers of `message.error` are `undeliveredBubble` (hard-coded `role: "user"`) and `failUserMessage` (matches the *user* id `uId`); the base's `failQueuedOnDisconnect` branch that set an error on the assistant bubble is what this PR removed. So every send failure renders as `UndeliveredMarker` instead: raw `error.message` in an 11px `text-destructive` span, losing `describeSendError`'s actionable hints (402/401/429) and the `role="alert"` Callout — the failure is conveyed by color alone and no longer announced to screen readers.
  **Fix:** render the user-side failure through `SendErrorCard` (keeping `prompt-delivery-error`, `describeSendError`, and `role="alert"`) and mount the overflow menu as an action slot on that card.
- 🔴 **Critical:** the full-suite disconnect spec asserts the removed behaviour by prose — `packages/e2e/playwright/src/tests/full/prompt-delivery/disconnect.spec.ts:53`
  `packages/e2e/` is untouched by this diff, but the spec still expects `prompt-delivery-error` to contain "Send failed" and `/connection dropped while this prompt was still waiting/i`, with `prompt-retry-button` visible. That testid lives only in `send-error-card.tsx`, which per the finding above can no longer render for this case; neither prose string exists anywhere in `packages/ui/` now; and the surviving retry control is a `DropdownMenuItem` inside collapsed `DropdownMenuContent`, so `toBeVisible()` fails too. Grepping the identifier finds nothing — only the string's words do. The suite runs under `E2E_FULL=1` on demand rather than on the PR gate, which prices this but does not remove it.
  **Fix:** update the spec to the new contract — assert the undelivered marker and its recovery path — in this PR, since it is the change that invalidated it.
- 🟡 **Warning:** `prompt-retry-button` is now on two elements — `packages/ui/src/modules/sessions/components/chat-message.tsx:234`
  `send-error-card.tsx:44` still carries the same testid. Once both surfaces are reachable the locator matches two elements and Playwright strict mode fails.
  **Fix:** give the menu item its own testid and update the e2e `retryButton` helper, or keep exactly one owner.
- 🟡 **Warning:** the handover cap is all-or-nothing against an uncapped local store — `packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:1030`
  `z.array(undeliveredPromptSchema).max(32)` returns `[]` on overflow, and `undelivered-store.ts` caps neither count nor bytes. A client holding 33 sends loses all 33, not the excess — and the reply still says success.
  **Fix:** cap in the browser store as it writes, and reject-with-error rather than silently returning `[]` when the array exceeds the cap.
- 🟡 **Warning:** `undeliveredPrompts` is an optional dependency — `packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:102`
  Six call sites read `deps.undeliveredPrompts?.…`, so missed wiring silently discards user prompts instead of failing to type-check. `composeAcp:94` always builds the store, so nothing needs the optionality. Same pattern for `now` in `undelivered-prompt-store.ts:88`, defaulted only for tests.
  **Fix:** make both required and pass the real implementation from `composeAcp`, a stub from tests.
- 🟡 **Warning:** `queueParkMs` is the scheduler's knob carried by its consumer — `packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:101`
  Hoisted into `AcpRuntimeDeps`, then forwarded by conditional spread at 166-168.
  **Fix:** configure the park window where the scheduler is built, in `compose.ts`.
- 🟡 **Warning:** `forgetUndeliveredPrompt` has no rejection handler — `packages/ui/src/modules/sessions/views/chat-view.tsx:139`
  `withConnection` rejects on a dropped connection, so `void` leaves an unhandled rejection; the optimistic removal is never rolled back and `appendUndelivered` resurrects the message on the next load. The sibling call at `use-acp-connection.ts:348` shows the convention.
  **Fix:** attach a `.catch()` that surfaces the failure or restores the row.
- 🟡 **Warning:** Delete is irreversible with no confirmation — `packages/ui/src/modules/sessions/components/chat-message.tsx:247`
  One `onSelect` drops the store row, the `localStorage` record and the agent-side record; the user's own prompt text is then unrecoverable. No destructive styling distinguishes it from the adjacent Copy text. The repo has a confirm convention (`confirmDeleteKnowledgeBase`).
  **Fix:** require confirmation or offer undo, and mark the item destructive.
- 🟡 **Warning:** "Copy text" copies only the first text part — `packages/ui/src/modules/sessions/components/chat-message.tsx:213`
  `parseUserText` splits a prompt into several text segments around `@file` references, so a message mentioning a file copies a fragment; with only image blocks it copies `""` and silently no-ops.
  **Fix:** copy `message.error.retryWith.text`, the same source Retry resends.
- 🟡 **Warning:** three representations of one wire concept — `packages/ui/src/modules/acp/session-projection.ts:150`
  `UndeliveredRecord` is an all-optional structural copy of `PlatformUndeliveredPrompt` plus `text`, existing to accept both shapes — which is what hides the first 🔴.
  **Fix:** type the projection input as `PlatformUndeliveredPrompt` and drop the parallel shape. Related: the adapter's pass-through re-export block (`undelivered-prompt-store.ts:11`) makes `acp-runtime.ts` import one schema twice under two names.
- 🟡 **Warning:** scope beyond the stated issue — `packages/agent-runtime/src/modules/acp/infrastructure/undelivered-prompt-store.ts:1`
  The body is literally `Closes #3264`, which asks for a per-prompt marker so `appendQueuedUser` stops merging queued prompts. The durable store, two RPC methods, queue parking, the browser store, new contract schemas, a delete affordance, and a new ubiquitous-language term all ride along undeclared — including the removal of the user-facing loss notice.
  **Fix:** declare the extra work in the body, or split the undelivered/parking subsystem out and keep this PR to the marker.
- 🟢 **Suggestion:** `undeliveredOf` (+`INLINE_IMAGE_BYTES_CAP`) is a pure wire→record mapper in a 1000+ line application file that already has `domain/mappers.ts` (`acp-runtime.ts:968`).
- 🟢 **Suggestion:** `UndeliveredMarker` is a second subcomponent inlined into `chat-message.tsx:202`; `copy()`'s `CopyState` is discarded via `void` (`:242`), so a failed clipboard write is silent; the inline `opts?` (`:28`) re-declares the exported `SendPromptOptions`.
- ✅ **Looks good:** every added comment block carries a valid typed prefix. `dropQueue` is the single `queues.delete` that can lose prompts and it always calls `onQueueDropped`; `refuseQueue` is the documented carve-out and answers each sender. `maybeCloseIdleSession` on `park-expired` only is correct — the other two causes fire from teardown paths that already reap. The store's own-document rationale holds, and eviction drops whole stale sessions rather than trimming a prompt.

- 🔴 `docs/architecture/agent-lifecycle.md:135` — the page still asserts the queue is "lossy by design": queued prompts "are discarded when it detaches" and "nothing survives to re-deliver". This PR inverts exactly that.
  **Fix:** rewrite the paragraph for the new contract — the last channel leaving parks the queue, a client engaging again resumes it, and every route a queue leaves by writes its prompts down.
- 🔴 `docs/architecture/agent-lifecycle.md:133` — the "sender's connection drops while its prompt is still queued" bullet still calls that loss unreportable and permanent ("no reply ever will be"). The prompt is now recorded and handed back on every `session/load`.
  **Fix:** state that a prompt lost this way is recorded and offered back on the next session load.
- 🔴 `docs/architecture/agent-lifecycle.md:125` — "Prompt delivery" says a prompt is queued only "while a turn is in flight". `canStart` also queues when nobody is engaged or the harness is not holding the session, and `heldPrompts` is gone.
  **Fix:** broaden the condition to all three causes and state the scheduler is now the only place a prompt waits.
- 🟡 `docs/architecture/persistence.md:115` — the `.platform/` inventory lists only session-metadata, trigger-binding and runtime-channel files; this PR adds a fourth durable document holding user-authored content that outlives teardown and restarts.
  **Fix:** add `undelivered-prompts` to the list with what survives in it and what removes a record.
- 🟡 `docs/architecture/agent-lifecycle.md:139` — the idle condition ("no prompt active or queued") no longer describes when `session/close` is sent, since a parked queue holds the reap back until its window passes.
  **Fix:** state the park-window hold-back alongside the background-work one.
- 🟢 `docs/architecture/agent-lifecycle.md:127` — the prompt-fate section documents neither the two new client-facing verbs nor the `turn.inFlight` / `undelivered` fields the `session/load` reply now carries.
- 🟢 `docs/architecture/platform-topology.md:81` — the client-state paragraph covers the composer draft but not the new `platform-undelivered:` keys, a different lifetime and hand-off on the same storage surface.

### Issue Fit

- 🟡 `packages/agent-runtime/src/modules/acp/infrastructure/undelivered-prompt-store.ts:1` — reported above (scope beyond the stated issue).
- 🟡 `packages/ui/src/modules/acp/session-projection.ts:237` — undeclared behavior removal: `QUEUED_LOST_MESSAGE` is deleted and `failQueuedOnDisconnect` no longer turns a lost queued prompt into a retryable user-facing error, with the matching assertions inverted. #3264 never asks for this.
  **Fix:** state it in the body and confirm every non-parked loss path still surfaces the prompt, or keep the disconnect error as a fallback for what parking does not cover.
- 🟢 `packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:428` — the issue proposed carrying the marker as `promptId` in `_meta.platform`; the PR emits a top-level `messageId` with a `randomUUID()` fallback. Equivalent, but a divergence worth a note.

- 🔴 `packages/ui/src/modules/sessions/lib/undelivered-store.ts:11` — reported above; the skill reached the same wire-shape defect independently.
- 🔴 `packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:102` — reported above (optional service dependency).
- 🟡 `packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:101` — reported above (`queueParkMs` hoisted into its consumer).
- 🟡 `packages/ui/src/modules/acp/session-projection.ts:150` — reported above (third representation of one wire concept).
- 🟡 `packages/agent-runtime/src/modules/acp/infrastructure/undelivered-prompt-store.ts:11` — the adapter re-exports contract schema and types under new names, so `acp-runtime.ts` imports the same schema twice by two paths. Infrastructure holds adapters and mappers, not a barrel for contract types.
  **Fix:** delete the pass-through block; import from `api-server-api` directly.
- 🟡 `packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:968` — reported above (`undeliveredOf` belongs with the mappers).
- 🟡 `packages/agent-runtime/src/modules/acp/infrastructure/undelivered-prompt-store.ts:88` — `now: () => string = () => new Date().toISOString()` is a clock dependency defaulted so only tests supply it.
  **Fix:** make `now` required and pass the real clock from `composeAcp`, the stub from tests.

- 🔴 `packages/ui/src/modules/sessions/components/chat-message.tsx:172` — reported above (`SendErrorCard` unreachable).
- 🔴 `packages/ui/src/modules/sessions/components/chat-message.tsx:165` — reported above (raw error string replaces the hinted, `role="alert"` surface; e2e contract broken).
- 🟡 `packages/ui/src/modules/sessions/components/chat-message.tsx:234` — reported above (duplicate `prompt-retry-button`).
- 🟡 `packages/ui/src/modules/sessions/components/chat-message.tsx:213` — reported above ("Copy text" copies the first text part only).
- 🟡 `packages/ui/src/modules/sessions/views/chat-view.tsx:139` — reported above (unhandled rejection on delete).
- 🟡 `packages/ui/src/modules/sessions/views/chat-view.tsx:134` — session-domain delete logic (store mutation + `localStorage` + agent RPC) is inlined into a 924-line `ChatView` holding ~40 consecutive `useStore` calls.
  **Fix:** extract to `modules/sessions/hooks/use-delete-undelivered.ts` and have `ChatView` call the hook.
- 🟡 `packages/ui/src/modules/sessions/components/chat-message.tsx:247` — reported above (irreversible Delete, no confirmation).
- 🟢 `packages/ui/src/modules/sessions/components/chat-message.tsx:242` — the `CopyState` from `copy()` is discarded, so a failed clipboard write gives no feedback.
- 🟢 `packages/ui/src/modules/sessions/components/chat-message.tsx:202` — `UndeliveredMarker` is a same-file subcomponent; extract to `undelivered-marker.tsx`.
- 🟢 `packages/ui/src/modules/sessions/components/chat-message.tsx:28` — the inline `opts?` structurally re-declares the exported `SendPromptOptions`; prefer `Pick<SendPromptOptions, "retryOf" | "blocks">`.

### Verdict

REQUEST_CHANGES — the design holds, but three 🔴 block: the browser's held sends are emptied by the wire schema and then deleted locally on a success reply that isn't one; the failure surface this PR routes every error to has lost the hints and the `role="alert"` announcement; and the disconnect e2e spec still asserts the deleted behaviour by prose. The doc page's "lossy by design" contract is now the opposite of the code.

---

dam-code-guardian (commented):
Delta re-review at `10784c47` — the four commits since `aed789ad`. Every blocking finding from the last round is closed at its cause.

### Changes since last review

**✅ Fixed**

- **The wire shape is now one type.** `PlatformUndeliveredPrompt` from `api-server-api` is the single contract across `undelivered-store.ts`, `session-projection.ts`, the new `domain/mappers.ts` and the server store; the local `UndeliveredRecord` with its optional `text`/`blocks` is gone, so the handover no longer loses content. `platform/recordUndelivered` now rejects a malformed batch with `-32602` instead of silently recording nothing.
- **`SendErrorCard` is reachable again** — undelivered notices moved into the new `UndeliveredNotice`, and the card renders on the assistant branch it was written for (`chat-message.tsx:166`).
- **`disconnect.spec.ts` asserts state, not prose** — marker count and content, `deliveryError` hidden, `queuedIndicator` hidden — plus a new `undelivered.spec.ts` covering a prompt outliving a closed tab, returning across two visits, and leaving on send or delete.
- **The architecture prose matches the code** — `agent-lifecycle.md` now describes the parked queue, the durable document and the hand-back.
- **The ghost-replay class is closed properly** — `supersededEchoes` retires a resent or deleted prompt's identity, `session-bootstrap` reports it, and `dropSuperseded` drops the stale echo with its vestigial placeholder.

**🔁 Still present** — one carried 🟡 in the retry path, and the same class in a new location.

### Findings

- 🟡 `packages/ui/src/modules/sessions/hooks/use-acp-prompt.ts:154` — `failUserMessage` re-records the undelivered prompt from `text` + `attachments` only. A retry of a *recovered* prompt carries its content as `blocks` and no `attachments` (`retryHandlerFor`, `undelivered-notice.tsx:35`), so a second failure stores the text alone: the images vanish and `droppedAttachments` stays empty, so the notice cannot tell the user either. This breaks the invariant asserted at `undelivered-notice.tsx:23-27`.
  **Fix:** give `undeliveredRecordOf` an optional `blocks` input and pass `sendOpts.blocks` when present, so the resend round-trips the same blocks and reports whatever it had to drop.
- 🟡 `docs/architecture/platform-topology.md:81` — the paragraph extends the draft store's ownership invariant ("signing out empties it, and signing in as someone else empties it too") to undelivered sends, but both sweeps are prefix-scoped to `platform-draft:` (`auth.ts:86` calls `clearAll()`; `claimDraftsFor` calls `removeAllDraftEntries`), so `platform-undelivered:` keys holding prompt text and inline image bytes survive sign-out and survive signing in as a different user in the same browser.
  **Fix:** either sweep the undelivered prefix on sign-out and owner change and keep the sentence as written, or state on the page that the undelivered record belongs to the session rather than the browser session and therefore outlives sign-out.
- 🟡 `docs/architecture/agent-lifecycle.md:135` — two gaps in the delivery contract as written. The "three routes" enumeration omits `refuseQueue` (a failed rehydrate refuses the queue and errors each sender rather than recording), and the hand-back paragraph never states the retirement rule this delta's top commit adds — a resent or deleted record is retired, and the session load names the retired identities so the earlier echo is dropped from the replay.
  **Fix:** name the fourth route with its reason (the sender is still connected to hear the failure), and add one sentence for retirement, noting it is per-process unlike the durable record itself.
- 🟡 `docs/architecture/persistence.md:115` — the disposal rule reads exhaustively ("it goes when the user resends or deletes it, or with its Session"), but the store also evicts on size: past its byte cap `withinCap` deletes whole other sessions' entries, oldest first, to make room for the session being written.
  **Fix:** name the bound in the same sentence — durable but not unconditional — leaving the cap's value to the code.
- 🟡 `packages/agent-runtime/src/modules/acp/infrastructure/undelivered-prompt-store.ts:1` — the scope is still undeclared: the description is `Closes #3264` alone, while the diff adds the whole undelivered-prompt subsystem, two ACP ext methods, the `QUEUE_PARK_MS` knob through `config.ts` and `values.yaml`, and new UI surfaces. It replaces the previous `QUEUED_LOST_MESSAGE` disconnect path rather than following from the asked change, so it is not mechanical fallout.
  **Fix:** declare the delivery rework and its operational knob in the description, or split it so #3264 closes on the projection fix alone.
- 🟢 `packages/ui/src/modules/sessions/hooks/use-delete-undelivered.ts:23` — local store and localStorage are mutated before the server call, and a rejected `forgetUndeliveredPrompt` only toasts, with no rollback.
- 🟢 `packages/agent-runtime/src/modules/acp/domain/mappers.ts:1` — `undeliveredOf` brings `randomUUID` and `zod` into the domain layer, which had no imports before this change.
- 🟢 `packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:101` — `queueParkMs` is carried on `AcpRuntimeDeps` only to forward it to the prompt scheduler.
- 🟢 `packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:1022` — handover validation (the array schema, `HANDOVER_PROMPT_CAP`, the `-32602` error text) lives in the service rather than the contract package.
- 🟢 `packages/ui/src/modules/sessions/components/undelivered-notice.tsx:58` — `role="alert"` applies to every notice, including ones projected from storage at session load; several assertive live regions mounting at once makes announcement order arbitrary.
- 🟢 `packages/ui/src/modules/sessions/components/undelivered-notice.tsx:51` — `useCopy()`'s `state`/`copied` are discarded, so a successful copy gives no feedback while only failure toasts.
- 🟢 `packages/ui/src/modules/sessions/components/undelivered-notice.tsx:42` — the component re-guards `message.error` with optional chaining its only caller already narrows, leaving the empty-string fallbacks unreachable.
- 🟢 `packages/ui/src/modules/sessions/lib/undelivered-store.ts:60` — `browserStore` as a default parameter is an implicit seam no caller overrides.
- 🟢 `packages/agent-runtime/src/modules/acp/services/acp-runtime/session-bootstrap.ts:8` — second `import type` from `api-server-api`; merge with line 1.

### Verdict

**COMMENT** — every blocking finding from the previous round is resolved at its cause, so nothing here blocks approval. What remains is one code residual in the retry path (the `blocks` round-trip) plus four statement-level items: three documentation rules that read as exhaustive but are not, and the undeclared scope.



---
🛡️ *Code Guardian 🛡️ · [definition](https://github.com/dam-agents/code-guardian)*

packages/ui/src/modules/sessions/lib/undelivered-store.ts
```diff
@@ -0,0 +1,157 @@
...
+
+export function clearUndelivered(
+  key: string,
+  store: UndeliveredStore = browserStore,
+): void {
+  store.removeItem(storageKey(key));
+}
+
+export function removeAllUndelivered(
+  store: UndeliveredStore = browserStore,
```
dam-code-guardian: 🟡 **Warning:** this sweep accesses `localStorage` with no guard, while its sibling `removeAllDraftEntries` wraps `store.keys()` and each `removeItem` in try/catch (`draft-snapshot.ts:116`). `logout()` calls it **before** `await userManager.signoutRedirect()` (`auth.ts:88`), so a storage throw — a disabled store, a private-mode quota error — aborts sign-out and leaves the user signed in. `startDraftSync` has the same exposure (`store.ts:88`).

**Fix:** every `localStorage` access in this store guards its own call, as `draft-snapshot.ts` does. Wrap `keys()` in try/catch with an early return, and wrap each `removeItem`.

packages/ui/src/modules/sessions/components/undelivered-notice.tsx
```diff
@@ -0,0 +1,119 @@
...
+ */
+export function retryHandlerFor(
+  message: Message,
+  onRetry: OnRetry,
+): (() => void) | undefined {
+  const retryWith = message.error?.retryWith;
+  if (!retryWith) return undefined;
+  return () => {
+    onRetry(retryWith.text, retryWith.attachments, {
+      retryOf: message.id,
```
dam-code-guardian: 🟡 **Warning:** `retryOf` is always `message.id`. `SendErrorCard` calls this handler with the *assistant* message (`chat-message.tsx:169`), and `sendPrompt` filters that id out of the list (`use-acp-prompt.ts:224`). A retry after an interrupted turn therefore deletes the partially streamed agent reply the user was reading.

**Fix:** `retryOf` identifies the undelivered *user* prompt to supersede. Pass it only from a user-role message; call `onRetry` without `retryOf` from the `SendErrorCard` path.

docs/architecture/agent-lifecycle.md
```diff
@@ -122,21 +122,21 @@ Both verbs are runtime-mediated: a hot `session/resume` engages the channel and
...
+- **No acceptance within a bounded wait** — the true delivery check, normally a matter of milliseconds. Unacknowledged this long means it never arrived, and the failure carries the socket's stated cause when it gave one.
 - **Waiting is unbounded.** A queued prompt is never failed for waiting, however long the turn ahead of it runs. The user is told it is waiting, which is true.
-- **The sender's connection drops while its prompt is still queued.** The one loss the runtime cannot report, since reporting it needs the channel that just went away — so the sender raises it locally and keeps it raised across the reconnect that follows. That reconnect replays the log, which misleads here: the dropped prompt's echo is in it, and no reply ever will be.
+- **The sender's connection drops while its prompt is still queued.** The runtime cannot answer a channel that is gone, so the sender raises the loss locally, writes the send down in its own storage, and hands that record over when it next reaches the runtime.
```
dam-code-guardian: 🟡 **Warning:** this bullet still says the sender "raises the loss locally, writes the send down in its own storage, and hands that record over". `failQueuedOnDisconnect` now drops such a bubble without an error, and the runtime parks and records the queue instead, so the client raises nothing for this case.

**Fix:** a delivery failure appears in this list only while the client itself still raises it. Rewrite the bullet for the parked-queue contract, or delete it and let the paragraph below own the case.

dam-code-guardian (commented):
Delta re-review at `8c8ff19c` — the one commit since `10784c47`. Four of the five open 🟡 are closed at their cause, and the PR body now declares the scope. The sign-out fix leaves one new residual.

### Changes since last review
Previous HEAD: `10784c47` (2026-09-03T17:44:41Z) — verdict COMMENT

**✅ Fixed**

- The retry path round-trips `blocks` — one `retryPayload` feeds `undeliveredRecordOf`, both error branches and the assistant bubble (`packages/ui/src/modules/sessions/hooks/use-acp-prompt.ts:123`).
- The sign-out invariant is now real — `logout()` and `claimDraftsFor` both sweep `platform-undelivered:` (`packages/ui/src/auth.ts:88`, `packages/ui/src/store.ts:88`).
- `agent-lifecycle.md` names the fourth queue-drop route with its refusal, and the retirement rule (`docs/architecture/agent-lifecycle.md:135`).
- `persistence.md` names the byte-cap eviction in the disposal sentence (`docs/architecture/persistence.md:115`).
- The scope is declared — the body lists parking, the pod records, the browser copy, retry/delete, the sweep and the doc updates.

**🆕 New** — three, all below the blocking bar for the code paths they touch: the unguarded sweep, the `retryOf` mix-up, the missing unit test.

**🔁 Still present**

- `undeliveredOf` holds `randomUUID` and `zod` in the domain layer (`packages/agent-runtime/src/modules/acp/domain/mappers.ts:1`).
- `queueParkMs` is carried by its consumer, and its default has two owners (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:101`).
- Handover validation lives in the service, not the contract package (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:1022`).
- `role="alert"` applies to notices projected from storage at load (`packages/ui/src/modules/sessions/components/undelivered-notice.tsx:58`).
- The optimistic delete has no rollback (`packages/ui/src/modules/sessions/hooks/use-delete-undelivered.ts:23`).

### Findings

- 🟡 `packages/ui/src/modules/sessions/lib/undelivered-store.ts:152` — `removeAllUndelivered` reads and writes `localStorage` with no guard, while its sibling `removeAllDraftEntries` wraps `store.keys()` and each `removeItem` in try/catch (`draft-snapshot.ts:116`). `logout()` calls the unguarded sweep **before** `await userManager.signoutRedirect()` (`auth.ts:88`), so a storage throw — a disabled store, a private-mode quota error — aborts sign-out and leaves the user signed in. `startDraftSync` has the same exposure (`store.ts:88`).
  **Fix:** every `localStorage` access in this store guards its own call, as `draft-snapshot.ts` does. Wrap `keys()` in try/catch with an early return, and wrap each `removeItem`.
- 🟡 `packages/ui/src/modules/sessions/components/undelivered-notice.tsx:36` — `retryHandlerFor` always sets `retryOf: message.id`. `SendErrorCard` calls it with the *assistant* message (`chat-message.tsx:169`), and `sendPrompt` filters that id out of the list (`use-acp-prompt.ts:224`). A retry after an interrupted turn therefore deletes the partially streamed agent reply the user was reading. Reported independently by react-ui-engineering.
  **Fix:** `retryOf` identifies the undelivered *user* prompt to supersede. Pass it only from a user-role message; call `onRetry` without `retryOf` from the `SendErrorCard` path.
- 🟢 `packages/ui/src/modules/sessions/lib/undelivered-store.ts:152` — the new sweep and the new `blocks` branch of `undeliveredRecordOf` have no unit test; `packages/ui/src/__tests__/unit/` holds no file for either store, and the `UndeliveredStore` seam makes one cheap. The `blocks` branch deliberately does not append `input.text` while the legacy branch does — a divergence a test would pin.

- 🟡 `docs/architecture/agent-lifecycle.md:133` — the client-evidence list still says the sender "raises the loss locally, writes the send down in its own storage, and hands that record over". `failQueuedOnDisconnect` now drops such a bubble without an error, and the runtime parks and records the queue instead, so the client raises nothing for this case.
  **Fix:** a delivery failure appears in this list only while the client itself still raises it. Rewrite the bullet for the parked-queue contract, or delete it and let the paragraph below own the case.

### Issue Fit

✅ No findings. The PR delivers #3264's ask, and the extra work is declared in the body.

- 🟡 `packages/agent-runtime/src/modules/acp/domain/mappers.ts:1` — `undeliveredOf` calls `randomUUID()`, so a domain function is non-deterministic and the module is no longer import-free. Raised as 🟢 last round; the skill rates it higher. Nine other `domain/` directories in this repo import `node:crypto`, so what changed is this module's own prior convention, not a repo-wide rule.
  **Fix:** the caller mints the identifier and passes it in, so the domain function stays a pure mapping.
- 🔁 4 findings from the previous review still present (see review at `10784c47`) — frame parsing in `domain/`, service-side params validation, `queueParkMs` carried by its consumer with a two-owner default, and `compose.ts` reading the env `config` singleton.
- 🟢 `packages/ui/src/modules/sessions/lib/undelivered-store.ts:20` — the image-cap rule is implemented twice, in `undeliveredOf` and in `undeliveredRecordOf`, and both name dropped images by the same string independently. One helper next to `UNDELIVERED_INLINE_IMAGE_BYTES_CAP` in the contract package would keep them from drifting.

- 🟡 `packages/ui/src/modules/sessions/components/undelivered-notice.tsx:36` — reported above (`retryOf` deletes the streamed assistant reply).
- 🔁 1 finding from the previous review still present (see review at `10784c47`) — `role="alert"` on notices restored from storage at mount.

### Verdict

**COMMENT** — nothing blocks approval. All five 🟡 from the previous round are resolved at their cause. What remains is three 🟡: the unguarded sweep the sign-out fix introduced, the `retryOf` mix-up on the interrupted-turn path, and one documentation bullet the parking work made stale.



---
🛡️ *Code Guardian 🛡️ · [definition](https://github.com/dam-agents/code-guardian)*

packages/ui/src/modules/sessions/lib/undelivered-store.ts
jjeliga: Fixed in 6b7970f3: every localStorage read/removal in the draft and undelivered stores now goes through a shared safe-storage module that absorbs the throw (setItem stays raw on purpose — the quota-shedding write loop needs its catch). Sign-out, owner switch, and session load can no longer abort on a storage error.

packages/ui/src/modules/sessions/components/undelivered-notice.tsx
jjeliga: Fixed in 6b7970f3 (retryOf passed only from user-role bubbles), then superseded in 43db9268: the agent-side error card no longer offers Retry at all — the turn already ran, so recovery is sending again from the composer — and an interrupted bubble is no longer carried across rebuilds, so the replayed log is the only owner of agent content.

docs/architecture/agent-lifecycle.md
jjeliga: Fixed in 6b7970f3: the bullet is gone. Its still-true clause (the sender writes the failed send to its own storage and hands it over) moved to the no-acceptance bullet where that actually happens, and the parked-queue paragraph now opens by owning the disconnect-while-queued case.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `bf3f13e`

### Summary

Three new commits since the last review: `6b7970f` (a retry keeps the interrupted reply; storage throws cannot abort sign-out), `43db926` (an interrupted turn shows a plain notice; a lost send is always written down) and `bf3f13e` (a send to a cold session shows busy, not waiting for a previous prompt). Every blocking finding from the previous round is closed at its cause, and `bf3f13e`'s projection guard is correct: `waitsBehindAnotherReply` excludes the send's own bubble via `m.promptId !== promptId`, so a cold-session queue renders a plain busy reply while a real turn-ahead still renders the waiting label. Both new unit tests pin exactly that split. No new blocking finding.

### Changes since last review

Previous HEAD: `8c8ff19c` (2026-09-03T22:02:53+02:00) — verdict COMMENT

- ✅ Fixed — `undeliveredOf` no longer mints an id: `randomUUID` and the `node:crypto` import are gone, the caller passes `id`.
- ✅ Fixed — `retryOf` is now passed only from a user-role bubble, so a retry cannot delete a partially streamed reply.
- ✅ Fixed — every `localStorage` access in the undelivered store goes through the new `safe-storage.ts` boundary, so a storage throw cannot abort sign-out.
- ✅ Fixed — the client-evidence bullet in `agent-lifecycle.md` is rewritten: the loss the runtime now parks is no longer listed as client-raised.
- ✅ Fixed — the inline-image cap is deduplicated into one exported `capInlineImages`, consumed by both the runtime mapper and the browser store.
- ✅ Fixed — the withdrawn retry affordance on `SendErrorCard` is removed together with its e2e helper.
- 🔁 Still present — `role="alert"` on notices projected from storage at load (`undelivered-notice.tsx:61`); repriced to 🟢, a11y placement on an otherwise working diff.
- 🔁 Still present — handover-params validation and its wire error still live in the runtime service (now `acp-runtime.ts:707`, cap at :1026); repriced to 🟢.
- 🆕 New — the doc page's `Last verified:` header now trails its own newest edit.

### Findings

- 🟡 **Warning:** `docs/architecture/agent-lifecycle.md:3` — the header still reads `Last verified: 2026-09-03`, but the page was edited again on 2026-09-04 by `6b7970f`, which rewrote the prompt-delivery and parked-queue paragraphs. `docs/guidelines/documentation-guidelines.md:34` requires the header to be bumped whenever the page is edited, so the stamp no longer covers the page's newest content.
  **Fix:** bump the header to the date of the newest edit, and re-check it on every follow-up commit that touches the page, not only on the first one.

### Documentation drift

🔁 1 finding from the previous review still present (see review at `8c8ff19c`).

The stale `Last verified:` stamp above is this skill's only new finding; the code changes are otherwise reflected in `agent-lifecycle.md`, `persistence.md` and `platform-topology.md`.

### Issue fit

Linked issue: #3264 "UI - queued messages merge into one bubble after a reload" (`Closes #3264`).

🔁 1 finding from the previous review still present (see review at `8c8ff19c`).

- 🟢 The cold-harness `heldPrompts` hold-and-replay path is deleted and queueing widened to two further conditions (`canStart` = a channel engaged plus the harness holding the session), with a rehydrate failure erroring senders via `refuseQueue`. This is a real change to when a prompt is queued rather than held, and none of the body's bullets names it — but the previous round already carried the undeclared-scope finding for this subsystem and the body was expanded in response, so this is the same class one hop over rather than a fresh one.
- 🟢 The issue proposes carrying the marker as `promptId` in `_meta.platform`; the implementation emits a top-level `messageId` on the `user_message_chunk` echo and gates the merge on that, leaving two identifiers for one prompt (`acp-runtime.ts:450`).
- 🟢 `docs/ubiquitous-language.md:81` gains an "Undelivered Prompt" term and a rewritten "Prompt Scheduler" entry; the body's documentation bullet names only agent-lifecycle, persistence and platform-topology.

### TypeScript engineering

🔁 6 findings from the previous review still present (see review at `8c8ff19c`).

- 🟢 `packages/agent-runtime/src/modules/acp/domain/mappers.ts:1` — this file imports the contract package and Zod, and `undeliveredOf` parses an untyped JSON-RPC frame from inside `domain/`. The skill prices this as a layer violation against an external architecture spec; this repo ships no domain-purity guideline (`docs/guidelines/` has no slice-composition page) and the sibling `session-list.ts` has imported the contract package since before this PR, so it stays a suggestion. The delta did reduce the file's dependency surface — `node:crypto` and two contract symbols are gone.
- 🟢 `packages/api-server-api/src/modules/acp/types.ts:66` — `capInlineImages` puts the 4 MB budget and its keep/drop loop in the contract package. Deduplicating the rule to one place was the previous round's fix, and one shared home beats two copies; a domain module both packages import would be the cleaner home still.
- 🟢 `packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:1026` — one conceptual retention limit exists as three unshared constants (`HANDOVER_PROMPT_CAP = 32`, `RECORDS_CAP = 32`, a 16 MB `BYTES_CAP` against the contract's 4 MB image cap). The export route already exists in this PR.
- 🟢 `packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:707` — the `platform/recordUndelivered` handler parses `params.prompts`, applies the cap and hand-builds its `-32602` error inline in the runtime service rather than at a validation boundary.
- 🟢 `packages/ui/src/modules/sessions/hooks/use-acp-connection.ts:22` — the `sessions` hooks import three (now four) projection functions straight from `acp/session-projection.js`, a module with no `index.ts`. Pre-existing seam, widened by the delta.

### React UI engineering

🔁 4 findings from the previous review still present (see review at `8c8ff19c`).

- 🟢 `packages/ui/src/modules/sessions/components/undelivered-notice.tsx:36` — the Retry item renders whenever `error.retryWith` is present, but `session-projection.ts:196` builds `retryWith: { text: textOf(record), ... }`, so a record whose only images were shed by `capInlineImages` yields a truthy `{ text: "" }`. Retry then calls `sendPrompt("")`, which early-returns on its `empty` guard (`use-acp-prompt.ts:117`) with no toast — an enabled control that does nothing. Narrow: it needs a record that was image-only *and* over the 4 MB cap, and the adjacent Copy item is already gated on `text === ""`.
- 🟢 `packages/ui/src/modules/sessions/components/undelivered-notice.tsx:39` — with `Role` being `"user" | "assistant"` and the sole call site rendering under `!isAssistant && error !== undefined`, the non-user branch of `retryHandlerFor` is unreachable, and the `UNIT_BOUNDARY_DESCRIPTION` above it documents that unreachable case as live behaviour. This is the shape the previous round's fix took; a discriminated prop would let the compiler drop the branch.
- 🟢 `packages/ui/src/modules/sessions/components/undelivered-notice.tsx:24` — the `UNIT_BOUNDARY_DESCRIPTION` comment sits above the private `retryHandlerFor` rather than the file's single exported unit, which is where `docs/guidelines/comment-guidelines.md:15` places it.
- 🟢 `packages/ui/src/modules/sessions/components/undelivered-notice.tsx:54` — `useCopy()` is destructured for `copy` alone, so a *successful* copy confirms nothing; every other call site in the package consumes `state` or `copied`. Pre-existing at the previous HEAD, unchanged by the delta.
- 🟢 `packages/ui/src/modules/sessions/components/chat-message.tsx:121` — the bubble guard `(!error || parts.length > 0 || undelivered)` reduces to one negative ("hide only an assistant error bubble with no parts"); a named boolean would state it. The `cn` call two lines down repeats `!isAssistant` in an argument that is inert for assistants.

### Verdict

APPROVE. The previous round's four blocking findings are each fixed at their cause, and `bf3f13e` is a well-scoped projection fix with tests covering both sides of its split. The only new finding is a stale `Last verified:` header — a one-line docs edit, not a reason to hold the PR. The remaining items are pre-existing seams and layering preferences on a working diff.


---

packages/ui/src/modules/sessions/hooks/use-acp-prompt.ts
```diff
@@ -101,24 +110,89 @@ export function useAcpPrompt(opts: UseAcpPromptOptions): {
...
+      const uId = promptId;
+      let reported = false;
       const dropBubble = () =>
         setMessages((p) => p.filter((m) => m.id !== aId));
+      const recordUndeliveredLocally = (reason: string) => {
+        if (hidden) return;
+        rememberUndelivered(
+          draftKey(selectedAgent, intendedSessionId),
+          undeliveredRecordOf({
+            id: uId,
```
PetrBulanek, line 162: A resend mints a fresh `promptId` and uses it as the record id, while the pod only forgets the original inside the `session/prompt` handler — after the prompt actually arrives.

So a retry that fails before reaching the pod leaves two records for one message. I reproduced this on a dev cluster: after clicking Retry with the socket down, the browser stored the prompt under a new id, the pod still held the original, and the next session load showed the same message as two undelivered bubbles. Both are durable, so the user has to delete each copy.

This is not an unlikely sequence — the reason a prompt is undelivered is usually that the pod was unreachable, and a retry taken while it is still unreachable hits exactly this.

Record identity is what makes dedupe work everywhere else here (the echo, `remember`, and `appendUndelivered` all match on it). A resend supersedes the record it names, so it needs to carry that identity rather than take a new one.

packages/ui/src/modules/sessions/hooks/use-acp-prompt.ts
```diff
@@ -101,24 +110,89 @@ export function useAcpPrompt(opts: UseAcpPromptOptions): {
...
       const aId = crypto.randomUUID();
+      const promptId = crypto.randomUUID();
+      const uId = promptId;
+      let reported = false;
       const dropBubble = () =>
         setMessages((p) => p.filter((m) => m.id !== aId));
+      const recordUndeliveredLocally = (reason: string) => {
+        if (hidden) return;
+        rememberUndelivered(
+          draftKey(selectedAgent, intendedSessionId),
```
PetrBulanek, line 160: `recordUndeliveredLocally` keys the browser copy with `draftKey(selectedAgent, intendedSessionId)`, and `intendedSessionId` is read from the store before `beginSession()` runs. For the first send in a blank chat that value is `null`, so the record lands under the blank-chat key — even when `beginSession()` has already produced a real session id by the time the send fails.

Nothing reads that key. The handover reads `draftKey(selectedAgent, sid)` for a concrete session, and so does the replay that surfaces records. I reproduced it on a dev cluster: the record was written to `…:~new`, and reloading the blank chat showed nothing while the key stayed in storage.

So the first message of a conversation — the one most likely to meet a cold start or a wake failure — is the one case where the browser copy is never handed over and never shown again. The in-tab bubble still offers Retry, so the loss only appears after a reload. The record also accumulates: it holds the prompt text and any pasted image bytes, and nothing removes it until sign-out.

packages/e2e/playwright/src/tests/full/prompt-delivery/disconnect.spec.ts
```diff
@@ -47,31 +49,31 @@ test("a prompt queued when the connection drops fails with Retry, and the failur
...
+  await test.step("losing the connection marks the queued prompt undelivered", async () => {
     await dropWebSockets(page);
 
-    await expect(deliveryError(page)).toBeVisible({ timeout: 60_000 });
-    await expect(deliveryError(page)).toContainText("Send failed");
-    await expect(deliveryError(page)).toContainText(
+    const bubble = undeliveredBubble(page, promptB);
+    await expect(bubble).toHaveCount(1, { timeout: 60_000 });
+    await expect(undeliveredMarker(page)).toContainText(
       /connection dropped while this prompt was still waiting/i,
```
PetrBulanek: This step asserts that the undelivered marker contains "connection dropped while this prompt was still waiting". That sentence came from `QUEUED_LOST_MESSAGE`, which this change removes; grepping the repository for those words now finds only this line. The assertion cannot pass.

The step's premise looks stale too. `dropWebSockets` closes the sockets, and the UI reconnects on its first backoff step of one second — well inside the 8-second park window the e2e cluster runs with. The returning client engages the session, which resumes the parked queue, so prompt B should be delivered behind prompt A's long turn rather than becoming undelivered at all. That is the new contract working correctly, and it is the opposite of what the step now expects.

This suite runs on demand rather than on the PR gate, which prices the breakage but does not remove it — and this is the change that invalidated the spec.

PetrBulanek (approved):

kapetr (approved):

Issue #3552, opened 2026-09-03
Update the Bob agent to a Bob Shell release with native ACP


xjacka:
## Context

Bob Shell 2.0.2 (August 2026) added a native ACP server ([docs](https://bob.ibm.com/docs/shell/features/acp)) — ACP being the protocol the platform already uses to talk to every agent. The Bob agent is pinned to 2.0.0, which predates it.

When Bob 2 landed here it had dropped the experimental ACP mode the earlier integration relied on, so we wrote our own bridge to keep chat working. Every other agent gets its ACP adapter from the vendor and simply runs it; Bob is the only one where we maintain that layer ourselves.

## Problem

Our bridge reimplements what the harness now does natively. It drives Bob's headless batch mode one turn at a time and rebuilds sessions, history, and resume on top of it, so it has to track Bob's internal behaviour to keep working — and it is the largest piece of agent-specific glue we own.

It also holds Bob's chat below the rest of the platform. The batch mode it is built on has no way to ask a question mid-turn, so Bob runs fully auto-approved: users never get the approval prompt other agents give them, and never see a diff before Bob edits a file. Surfacing exactly that is what native ACP is for.

Staying on 2.0.0 also keeps Bob off everything else released since.

## Goal

Bob runs on a Bob Shell release with native ACP and talks to the platform directly. Chat, terminal, and credentials keep working, approvals behave like they do for other agents, and we no longer maintain a Bob-specific bridge.

## Done when

- New Bob agents run a Bob Shell release with native ACP support.
- Chat works end to end: sending a message, streaming the reply, listing past sessions, opening one, and continuing it after a restart.
- Tool calls and file edits go through the platform's normal approval flow instead of being auto-approved — or, if Bob's ACP mode cannot offer that, the reason is written down.
- The terminal, the Bob Shell provider card, model/tenant/cost settings, and credential injection behave as they do today.
- The platform-maintained bridge is gone.
- The Bob agent docs describe the new setup, and any user-visible change is called out.

xjacka:
Scope note: mode and approval settings for Bob are being added to this issue rather than filed separately, on the owner's call, and land in #3553.

Bob's manifest now declares a `harness-config` driver, so the agent gets a Config panel with **Mode** (`agent` / `plan` / `ask`) and **Approvals** (`auto` / `ask`). Before this, Bob had no per-agent settings surface at all — the only lever was the provider pin, which applies to every agent granted the secret.

[...]

PR #3553, opened 2026-09-03
feat(bob): run Bob Shell 2.0.2 as a native ACP agent

xjacka:
Part of #3552 — the closing keyword is deliberately absent: the issue keeps one "Done when" item (a real prompt turn, streaming a reply, and continuing a session after a restart) that needs a cluster with a granted Bob key, so it stays open until someone exercises it.

Bob Shell 2.0.2 ships `bob acp`, a first-class ACP server over stdio, so the platform now spawns Bob directly and the 581-line bridge we maintained is deleted. What replaces it is a 95-line settings bootstrap for the three pins `bob acp` has no flag for.

## What the harness advertises

Verified against the built image, not the docs:

```
loadSession: true
sessionCapabilities: { list, delete, resume, close }
promptCapabilities: { embeddedContext, image }
mcpCapabilities: { http, sse }
```

`session/new` returns Bob's own task id, so the platform session id and the harness task id are the same string — the bridge's id map and its SQLite reads are gone. History, paging and resume are native.

## Approvals: why chat still auto-approves

This is the one "Done when" item the PR answers with a reason rather than an implementation, so it is the part worth reviewing closely.

Bob's ACP has no middle approval setting. Its decision is: a built-in skill, a `read`-permission tool, or `--auto-approve` runs — anything else asks the client. Bob *does* own a finer posture (permission groups, a per-tool executor allowlist, a default allowlist of safe commands like `cat`/`ls`/`git status`), but the only code that consults it is the TUI approval handler, reached through Bob's terminal-app singletons. The ACP path never sees it.

So "ask" means ask about everything except file reads: the one execute tool, five edit tools, browser and MCP all prompt — and Bob is a shell assistant that reaches for `execute_command` constantly. An intermediate commit here did enable prompts; in use it asks before every `ls`, which is worse than not asking. The default is therefore `--auto-approve`, with `BOB_AUTO_APPROVE=0` opting an agent into prompts.

Per-session granularity (a Mode-style switch like claude-code's) needs Bob upstream: approvals wired to its modes, or `session/set_config_option` registered in its ACP host. Worth an ask to the Bob team.

Two things that soften the default: a user who does turn prompts on gets Bob's "always allow", remembered per tool name for the session; and an unattended session could not answer a prompt anyway — the request would expire against agent-runtime's TTL and abort the tool call.

## agent-runtime: rehydrate with `session/resume`

A cold session whose log a history provider already filled now re-attaches with `session/resume` where the harness advertises that capability, and `session/load` otherwise. This is a real defect, not a tidy-up — Bob answers `session/load` first and streams its replay afterwards, which would duplicate history the runtime had already served. Seeded a task with two messages in the image and observed the frame order directly:

```
session/load    →  RESPONSE, available_commands_update, user_message_chunk, agent_message_chunk
session/resume  →  RESPONSE, available_commands_update
```

## Also in this branch: two further work streams

The repository squash-merges, so this description becomes the permanent commit message — it therefore covers everything the branch carries, not only the ACP upgrade above.

**Bob on the LiteLLM proxy.** Egress path-prefix rewriting so Bob can run against a LiteLLM-shaped endpoint instead of only the IBM Bob gateway: the gateway rewrites the prefix and the auth scheme on the wire, which is what lets Bob's `harness` provider — already an OpenAI-compatible client posting to `/chat/completions` — talk to a different backend without a fork. Touches the controller's Envoy rendering (`envoy.go`, `envoy_bootstrap.go`, `envoy_test.go`), the connections catalog and providers, `connection-sds.ts`, the provider connect dialog, and the `connections` and `security-and-credentials` pages. These commits also exist on `feat/bob-openai-endpoint`.

**Named telemetry export rails.** A `telemetry` field on an agent template selects which export rail its image gets, so a harness that does not write Claude Code's per-call log records can still be pointed at the collector. Bob declares `telemetry: bob`, and the rail keeps its prompt and completion payloads off the wire — Bob defaults `excludePayload` to false, and the platform's invariant is that content bodies are not exported, so the bootstrap asserts it on every harness start rather than exposing it as a clickable setting.

To be precise about what that rail carries, because one commit message in this branch states the opposite: **Bob does export telemetry.** `initializeTelemetry()` runs at harness level — `bob acp` included, not the TUI alone — building an OTLP client from `BOB_TELEMETRY_*` and emitting an `LLM Generation` span per call with input, output, cache-read, cache-write, cost and duration, gated on `BOB_TELEMETRY_AGENT_OPS_ENABLED`, which the rail sets. The rail carries nothing on a default install only because `clickstack.enabled` is false there. Reading Bob's spend on the Usage surfaces is separate work (#3562) and stands on this rail.

## Added scope: mode and approvals in the Config panel

Requested on top of the upgrade, and declared here rather than left for a reader to notice. Bob's manifest now declares a `harness-config` driver, which is what puts the Config panel on the agent at all, offering **Mode** (`agent` / `plan` / `ask`) and **Approvals** (`auto` / `ask`).

Both write into a `platform` section of Bob's settings file that Bob itself ignores, rather than into `session.*`. The indirection is load-bearing: `bob-settings.mjs` rewrites Bob's own keys on every harness start, so a panel writing them directly would lose to the provider pin — or be deleted when that pin is cleared, by the very projection rule the last round asked for. The bootstrap resolves panel over pin and writes the result, so a panel choice survives restarts while an agent without one still follows the pin.

Approvals cannot ride the settings file at all, because Bob's ACP reads them from a spawn flag rather than from config — so the bootstrap prints the mode it resolved and both harness scripts translate it into `--auto-approve`. Verified across all five precedence cases (no state, pin only with a legacy value, panel over pin, env override, panel over env).

Two limits stated in the README rather than papered over: a change lands on the **next harness start**, since the platform writes the event once and never re-asserts it and neither surface re-reads the file mid-session; and **model stays out of the panel**, because Bob's list comes from a LiteLLM-shaped `/model/info` under its own path prefix that neither a static catalog nor `modelDiscovery` can serve yet — that is its own piece of work.

## Declared out-of-scope edit

The SSH paragraph in `agent-lifecycle.md` is condensed to its lifecycle-relevant part — SSH is not a Session, and an open connection pins the Agent awake — with the transport, auth and pin-reclaim detail replaced by cross-references to `cli.md`, `platform-topology.md` and this page's own Hibernate section, each of which already owns that detail in depth. The page sat 32 characters under the 40 000-char cap, so the one-sentence correction this PR needs could not land without freeing room, and the documentation guidelines prescribe exactly this remedy — raise the level, cut duplicated depth — over rewording to fit. Nothing is lost from the architecture set.

## Behaviour changes to call out

- **Tenant scoping is terminal-only.** `bob acp` takes no instance or team flag and the settings file has no key for either, so `BOB_INSTANCE_ID` and `BOB_TEAM_ID` no longer reach chat sessions. `BOB_INSTANCE_ID` was already terminal-only under 2.0; `BOB_TEAM_ID` is newly so. Headless instance selection goes through Bob profiles — worth a follow-up if anyone depends on it.
- **Chats created before this change are unreachable.** Their ids were minted by the bridge (`bob-<uuid>`) and mean nothing to Bob, so `session/load` answers `resourceNotFound`. The Bob-side tasks survive and are still resumable from the terminal with `bob -r`.
- `BOB_SHIM_TRACE` is gone with the bridge. `bobShell.autoUpdate` is pinned off — the image pins the version, and a self-update in the pod would fail against the egress rules and print onto stdout.

## Verification

Built the image and drove ACP against `harness-chat` in a container:

- `initialize`, `session/new`, `session/set_mode`, `session/list`, `session/resume` and `session/load` all answer correctly with no credential present
- `session/list` rejects a relative `cwd` (`Session list cwd must be an absolute path`). The platform never sends one — agent-runtime rewrites every frame's `cwd` to the pod work dir — but it makes that rewrite load-bearing for this agent
- `BOB_CHAT_MODE=plan` reaches a new session as `currentModeId: plan`; the platform instructions land as a rules symlink; `harness-terminal` starts the TUI with its flags accepted
- the image is deployed on a local k3s cluster (template enabled, `platform-bob:latest` imported), verified from inside the image: bobshell 2.0.2, no bridge, harness scripts as committed

Not yet exercised: a real prompt turn, which needs a granted Bob API key. Sending a message, streaming a reply, and continuing a session after a restart still need a check on a cluster with credentials.

## Follow-ups this uncovered

Two things the upgrade makes concrete rather than hypothetical, both out of scope here and filed separately: pointing Bob at an OpenAI-compatible endpoint (its `harness` provider already speaks `/chat/completions` against an overridable gateway base URL), and surfacing Bob's spend on the Usage surfaces (Bob exports its own telemetry, but not in the per-call record shape the metrics reader consumes).


packages/agents/bob/bob-settings.mjs
```diff
@@ -0,0 +1,96 @@
...
+  const settings = {
+    ...existing,
+    session: {
+      ...(isNonNullObject(existing.session) ? existing.session : {}),
+      ...(mode ? { defaultMode: mode } : {}),
+      ...(model ? { model } : {}),
+      ...(Number.isFinite(maxCost) && maxCost > 0 ? { maxCost } : {}),
+    },
+    approval: {
+      ...(isNonNullObject(existing.approval) ? existing.approval : {}),
```
dam-code-guardian: 🔴 **Critical:** The writer spreads `existing.approval` and overrides only `outsideWorkspaceAllowed`, but the bridge it replaces wrote `autoApprovalEnabled: true`, `allowed_permissions: ["read","edit","execute","browser","mcp"]` and an `allowedExecutors` entry approving `*` into the same file on the `/home/agent` PVC. Every Bob agent that ran the old image keeps those keys, so `bob acp` auto-approves and the permission requests this PR is built to surface never reach the client — the README's "Chat asks" and the issue's central criterion hold on a fresh volume only.
**Fix:** When a bootstrap stops writing a key, it must also remove that key from the persisted file — enumerate the keys the previous version owned (`autoApprovalEnabled`, `allowed_permissions`, `allowedExecutors`) and delete them, rather than spreading whatever the file holds.

packages/agents/bob/README.md
```diff
@@ -59,28 +88,25 @@ Less common toggles, not surfaced on the provider card.
 
 | Env var | Effect |
 |---|---|
-| `BOB_LOG_LEVEL` | Bob's log level: `debug`, `info`, `warn`, `error`, `silent`. |
+| `BOB_LOG_LEVEL` | Bob's log level: `debug`, `info`, `warn`, `error`, `silent`. Logs go to stderr; stdout belongs to the ACP stream. |
 | `IBM_TELEMETRY_ENABLED` | Set to `false` to opt out of Bob's telemetry. |
-| `BOB_SHIM_TRACE` | Set to `1` to log every shim frame (client↔shim↔bob) to the pod log. |
 
-Gone in 2.0 (silently ignored if set): `BOBSHELL_HIDE_ENVS`, `BOB_SHELL_PRE_CHECK_AUTO_APPROVED`, `BOB_SHELL_SYSTEM_MD` (custom instructions now ride the `.bob/rules/` directory — the image links the platform instructions there), `BOB_RESUME_MAX_MESSAGES` (resume is native now).
+Gone in 2.0 (silently ignored if set): `BOBSHELL_HIDE_ENVS`, `BOB_SHELL_PRE_CHECK_AUTO_APPROVED`, `BOB_SHELL_SYSTEM_MD` (custom instructions now ride the `.bob/rules/` directory — the image links the platform instructions there).
```
dam-code-guardian: 🟢 **Suggestion:** `main`'s list ends with `BOB_RESUME_MAX_MESSAGES` (resume is native now); this line drops it. Nothing in the tree reads that variable, so the entry was the only place a user who still has it pinned on a provider secret could learn it does nothing.

docs/architecture/agent-lifecycle.md
```diff
@@ -162,7 +162,7 @@ install can refuse holds outright.
...
+SSH sessions are unrelated to the session machinery above: they carry no session identity, no metadata entry, and no harness involvement — a per-connection in-pod OpenSSH the platform relays bytes to, with its own authentication and its own shell, several coexisting per agent. The transport and its auth belong to the CLI surface that uses it ([cli](cli.md#ssh-access)) and to agent-runtime's relay duties ([platform-topology](platform-topology.md)). For lifecycle an SSH connection matters only as a pin: while it is open the Agent will not hibernate, and the runtime's own idleness probe cannot see it ([Hibernate](#hibernate) states which signal carries it and how a lost connection's pin is reclaimed).
```
dam-code-guardian, line 164: 🟢 **Suggestion:** This paragraph's rewrite is unrelated to the PR's subject and is not declared in the body. Every fact it drops is carried by the page it now delegates to — `cli.md:100-106` (sshd per connection, key auth, replayed env), `platform-topology.md:63` (the `/api/ssh` relay and `sshd -i`), `agent-lifecycle.md:202-203` (the pin, the ping/pong and the Redis-presence reconcile) — and both anchors resolve, so nothing is lost and the page stays inside `scripts/doc-size.mjs`. It is the scope, not the content: an unrelated docs rewrite riding a behaviour PR is harder to review and harder to revert on its own.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `65fffd52`

## PR #3553: feat(bob): run Bob Shell 2.0.2 as a native ACP agent
**Author:** xjacka | **Branch:** feat/bob-acp-native → main | **Changes:** +210 −641 (10 files)

### Summary
Bob Shell 2.0.2 replaces the 581-line ACP bridge with a direct `bob acp` spawn plus a 96-line settings bootstrap, and agent-runtime learns to re-attach a provider-served session with `session/resume` where the harness advertises it. The runtime change and the image change are both sound in isolation; the settings rewrite does not account for what the deleted bridge already wrote to the PVC.

### Findings
- 🔴 **Critical:** deleted bridge's auto-approval keys survive on an upgraded PVC (`packages/agents/bob/bob-settings.mjs:56`)
- 🟡 **Warning:** A rehydrate answered by `session/resume` never reaches `transcript.cacheMetadata`, so a provider-served session on a resume-capable harness keeps its synthetic placeholder metadata (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:550`) — reported by TypeScript Engineering Review, details in that section.
  **Fix:** Every method that returns real session metadata belongs in the `cacheable` set; add `session/resume` with the two rehydrate verbs.
- 🟡 **Warning:** `sessionResumeSupported` survives a harness recycle, so a dead process's capability decides the next one's re-attach verb (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:108`) — reported by TypeScript Engineering Review, details in that section.
  **Fix:** Every piece of harness-derived state resets in `teardownRuntime`; capabilities are re-learned from the new process's `initialize`.
- 🟡 **Warning:** A `settings.json` parsing to a non-object makes the bootstrap throw before it writes anything, and the harness script ignores the exit code (`packages/agents/bob/bob-settings.mjs:34`) — reported by TypeScript Engineering Review, details in that section.
  **Fix:** Treat a parse result that is not a plain object as absent, and let the harness script fail loudly rather than start Bob without the platform posture.
- 🟡 **Warning:** `Closes #3552` lands while the issue's "Done when" prompt-turn check is declared unexercised (`packages/agents/bob/harness-chat.sh:7`) — reported by Issue Fit, details in that section.
  **Fix:** A closing keyword waits until every "Done when" item is checked, or the keyword drops to a plain reference until it is.
- 🟡 **Warning:** `BOB_TEAM_ID` becomes terminal-only, an undeclared behaviour regression for `general`-type API keys (`packages/agents/bob/README.md:81`) — reported by Issue Fit, details in that section.
  **Fix:** Carry tenant scoping into the ACP surface, or get the criterion re-scoped on the issue and file the follow-up the body only mentions in prose.
- 🟢 **Suggestion:** `BOB_RESUME_MAX_MESSAGES` dropped from the "Gone in 2.0" list (`packages/agents/bob/README.md:94`)
- 🟢 **Suggestion:** undeclared scope — the SSH paragraph rewrite is unrelated to this PR (`docs/architecture/agent-lifecycle.md:165`)
- ✅ **Looks good:** The response-branch restructure at `acp-runtime.ts:567-573` is safe. The old gate `mapping.method === "session/load" && mapping.attachSessionId` split into a rehydrate branch and a load branch, and `bootstrap.onLoadResponse` still sees exactly the non-rehydrate `session/load` responses it saw before — a rehydrate never carried `rehydrate: false`, and no other method sets `attachSessionId`.

- 🟢 **Suggestion:** The page names `session/load` as the only cold-start read of the harness's on-disk session store, but this PR adds a second — a resume-capable harness is re-attached with `session/resume` against the same store (`acp-runtime.ts:239`). The sentence is incomplete rather than wrong, and the page's `Last verified:` date was not bumped (`docs/architecture/persistence.md:3`). (`docs/architecture/persistence.md:112`)
  **Fix:** Name the re-attach generically and leave which ACP verb carries it to `docs/architecture/agent-lifecycle.md`, which now owns that distinction in depth; bump `Last verified:` with the edit.

### Issue Fit
Linked issue: #3552 "Update the Bob agent to a Bob Shell release with native ACP" (closing keyword)

- 🟡 **Warning:** The issue's "Done when" requires chat to work end to end — send a message, stream the reply, continue a session after a restart. The PR body states a prompt turn was never exercised, yet `Closes #3552` will close the issue on merge (`packages/agents/bob/harness-chat.sh:7`).
  **Fix:** Run one real prompt turn plus a post-restart resume on a cluster with a granted Bob key and record the result, or drop the closing keyword so the issue stays open until that check lands.
- 🟡 **Warning:** The issue requires model/tenant/cost settings to behave as they do today, but `BOB_TEAM_ID` becomes terminal-only — it no longer reaches chat sessions, an unasked behaviour regression for `general`-type API keys (`packages/agents/bob/harness-terminal.sh:8`, `packages/agents/bob/README.md:81`).
  **Fix:** Carry team scoping into the ACP surface (Bob profile or settings key), or get the tenant-scoping criterion explicitly re-scoped on #3552 and file the follow-up the PR body only mentions in prose.
- 🟢 **Suggestion:** The undeclared SSH-paragraph rewrite described in `### Findings`. Its detail is preserved elsewhere in the architecture set, so this is a scope observation rather than a content loss. (`docs/architecture/agent-lifecycle.md:165`)

- 🟡 **Warning:** The `cacheable` set still lists only `session/new`, `session/fork` and `session/load`, so a rehydrate answered by `session/resume` never calls `transcript.cacheMetadata`. A provider-served session on a resume-capable harness (Bob) keeps the synthetic placeholder metadata (`modes: null`, `configOptions: null`) forever, so later client loads lose the real modes — the metadata upgrade the test overview still promises (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:550`)
  **Fix:** Add `session/resume` to the cacheable methods, so every harness re-attach result replaces synthetic metadata; the rule is that any method that returns real session metadata must feed `cacheMetadata`.
- 🟡 **Warning:** `sessionResumeSupported` is harness-process state but `teardownRuntime` does not reset it, unlike every other harness-derived map and set it clears. After a recycle the flag from the dead process decides the next process's re-attach verb (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:108`)
  **Fix:** Reset `sessionResumeSupported = false` and `sessionCloseSupported = true` in `teardownRuntime`, next to the other harness-state clears, so capabilities are re-learned from the new process's `initialize`.
- 🟡 **Warning:** `readExistingSettings` returns whatever `JSON.parse` yields, so a `settings.json` holding `null`, a number or an array makes `existing.session` throw a `TypeError` at top level. `writeSettings()` then aborts before `ensureRules()`, and `harness-chat.sh` ignores the non-zero exit, so Bob starts with no `outsideWorkspaceAllowed`, no `autoUpdate: false` and no platform rules link (`packages/agents/bob/bob-settings.mjs:34`)
  **Fix:** Return `{}` unless the parsed value passes `isNonNullObject`, and run `writeSettings()`/`ensureRules()` under one try/catch that reports the failure, so a corrupt settings file cannot silently drop the platform posture.
- 🟢 **Suggestion:** `hasSessionCapability(frame, name: string)` accepts any string, so a mistyped capability name compiles and silently returns `false`. Typing `name` as `"close" | "resume"` restores the compile-time check the dedicated `extractSessionCloseSupported` had (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:989`)
- 🟢 **Suggestion:** The new test asserts only the verb choice; it does not assert the reason its TEST_SCENARIO gives — that the resume path neither duplicates the transcript nor leaves placeholder metadata behind. Emitting replay lines after the resume reply and checking a later load's `modes` would cover the regression above (`packages/agent-runtime/src/modules/acp/services/acp-runtime/__tests__/acp-runtime-history-provider.test.ts:170`)

### Verdict
REQUEST_CHANGES — the runtime resume path and the image rewrite are both correct for a fresh pod, but every already-running Bob agent keeps the deleted bridge's auto-approval settings on its PVC, so the approval posture this PR is built to deliver does not arrive on upgrade.

---

xjacka:
Thanks — acting on all of it. The code fixes are committed locally and land with the rebase onto current `main` (the PR is conflicting on `docs/architecture/agent-lifecycle.md`), so they will appear in one push shortly.

**Fixed**

- The retired approval keys are now pruned rather than spread: `autoApprovalEnabled`, `allowed_permissions` and `allowedExecutors` are deleted from `approval` on every bootstrap, so an agent upgraded in place stops carrying the deleted bridge's posture and falls back to Bob's own defaults. Verified against a settings file holding the bridge's exact block — unrelated keys (`editApprovalPreviewMode`, `licenseConsent`, `session.maxTurns`) survive, those three go.
- `session/resume` joins the cacheable methods, so a rehydrated provider-served session stops reporting placeholder modes.
- `sessionCloseSupported` and `sessionResumeSupported` reset in `teardownRuntime`, so a recycled harness re-learns its capabilities from its own `initialize`.
- A `settings.json` that parses to a non-object is treated as absent, both bootstrap steps run under one reported failure, and the harness scripts say so on stderr instead of starting silently without the posture. Verified with a file containing `null`.
- `hasSessionCapability`'s `name` is narrowed to `"close" | "resume"`.
- The resume test now asserts its stated reason: real modes on a later load, and one copy each of the provider history, the prompt and the live reply.
- `persistence.md` no longer names `session/load` as the cold-start read by definition.
- `BOB_RESUME_MAX_MESSAGES` is back in the retired-env list, alongside `BOB_SHIM_TRACE`.

**One correction to the critical finding's framing.** The review ran at `65fffd52`; `38df6a45` had already restored `--auto-approve` as the default. So the posture this PR delivers is auto-approval, and it arrives on upgrade — it rides the CLI flag, which does not read the settings file at all. The stale keys still mattered, which is why they are pruned: they govern the TUI, and they would have quietly defeated `BOB_AUTO_APPROVE=0` on an upgraded PVC, where the terminal kept a wildcard executor list from a component that no longer exists.

**On the two judgment calls, both accepted as they stand.**

- **`Closes #3552`** stays. The prompt-turn check is being run against a local cluster with a granted Bob key now; if it fails, the keyword drops to a reference rather than the issue closing on an unverified criterion.
- **`BOB_TEAM_ID` becoming terminal-only** is accepted as documented. `bob acp` takes no team flag and the settings file has no key for it, so carrying tenant scoping into the ACP surface needs Bob profiles — out of scope here, noted in the README table and the PR body.

**Two findings from a sibling investigation of Bob's bundle**, recorded here because they are Bob-image facts a later reader will want, though both belong to the telemetry work rather than this PR: `BOB_TELEMETRY_PROVIDER=langfuse` requires both Langfuse keys or Bob silently falls back to the `bob` provider pointed at the IBM endpoint; and `NODE_USE_ENV_PROXY=1` is what gets Bob's OTLP exporter through the gateway, since it uses Node `https.request` while Bob's own `ProxyAgent` covers only `fetch`.

packages/agents/bob/harness-chat.sh
```diff
@@ -1,4 +1,13 @@
 #!/bin/sh
-# The shim reads the BOB_* env itself (bob run flags + settings merge) and
-# emulates session list/load from Bob's task DB; see bob-acp-shim.mjs.
-exec node /app/bob-acp-shim.mjs "$@"
+# Bob is the ACP agent itself (2.0.2+); the platform speaks to it directly.
+# --trust: every session/new carries the pod workspace, which Bob would
+# otherwise refuse as untrusted.
+node /app/bob-settings.mjs ||
```
dam-code-guardian: 🟡 **Warning:** `node /app/bob-settings.mjs || echo … >&2` then `exec`s Bob regardless, so a failed `writeSettings` degrades to one log line and Bob starts without `approval.outsideWorkspaceAllowed: true` — which by this PR's own README means Bob refuses every tool touching `$HOME` (skills, rules). The pod reports healthy and every session fails at its first tool call. Same shape in `harness-terminal.sh:5`.
  **Fix:** A bootstrap step whose failure makes the process unusable must fail the process — drop the `||` swallow so supervision sees a non-zero exit — and reserve soft-fail for steps whose absence only removes a nicety.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `97abae3`

## Re-review

Previous HEAD: `65fffd52` — unreachable, reviewed the whole PR
Current HEAD: `97abae3a`

The branch was rebased between rounds (`65fffd52...97abae3a` compares as `diverged`: 24 ahead, 2 behind), so no commit range spans the two rounds and this round re-derived every finding at HEAD.

### Changes since last review

**Fixed (5 of 9)**

- 🔴 → fixed: the deleted bridge's approval posture no longer survives on the PVC. `RETIRED_APPROVAL_KEYS = ["autoApprovalEnabled", "allowed_permissions", "allowedExecutors"]` is deleted out of the merged `approval` section before the write, and the README documents the prune (`packages/agents/bob/bob-settings.mjs:59`). I diffed the deleted writer's whole key set against the new one: the only remaining residue is `licenseConsent: true`, which the bridge wrote and the bootstrap does not — but `--accept-license` on both surfaces makes it inert rather than a posture the platform no longer intends, so it is not a defect.
- 🟡 → fixed: `session/resume` is in the `cacheable` set, so a resume rehydrate now upgrades the synthetic metadata to the harness's real answer; the new test asserts a later load reports `modes.currentModeId: "auto"` (`acp-runtime.ts:570`).
- 🟡 → fixed: `teardownRuntime` resets both harness-derived capability flags (`sessionCloseSupported = true`, `sessionResumeSupported = false`), so a recycle cannot leave a stale resume claim behind (`acp-runtime.ts:338`).
- 🟡 → fixed: a non-object or unparseable `settings.json` no longer throws — `readExistingSettings` returns `{}` and `section()` guards each nested key — and the two steps run under their own `try`/`catch` with a stderr line and `process.exitCode = 1` (`bob-settings.mjs:40`).
- 🟢 → fixed: `BOB_RESUME_MAX_MESSAGES` is back in the "Gone" list, alongside `BOB_SHIM_TRACE` (`packages/agents/bob/README.md:94`). 🟢 → fixed: `persistence.md` no longer names `session/load` as the only cold-start read.

**Still open (4)**

Both `harness-chat.sh` warnings and the `BOB_TEAM_ID` one carry over; the team-ID finding narrowed — the README and PR body now declare the regression, so what is left is the provider card. Full text below.

**New (1 🟡, 5 🟢)** — the un-bumped doc stamp, and five suggestions, three of them on the new bootstrap.

---

### Findings

- 🟡 **Warning:** The launcher runs the bootstrap as `node /app/bob-settings.mjs || echo "…" >&2` and then `exec`s Bob regardless, so a failed `writeSettings` degrades to one log line and Bob starts without `approval.outsideWorkspaceAllowed: true` — which by this PR's own README means Bob refuses every tool that touches `$HOME`, i.e. skills and rules. The pod reports healthy and every session fails at its first tool call. Same shape in `harness-terminal.sh:5`. (`packages/agents/bob/harness-chat.sh:5`)
  **Fix:** A bootstrap step whose failure makes the process unusable must fail the process, not warn — drop the `||` swallow so the runtime's supervision sees a non-zero exit — and reserve soft-fail for steps whose absence only removes a nicety (the rules link).
- 🟡 **Warning:** `Closes #3552` still ships with the issue's central "Done when" item self-reported as unverified: the PR body's own Verification section says a real prompt turn, streaming a reply, and continuing a session after a restart "still need a check on a cluster with credentials". Merging closes the issue on an acceptance criterion nobody exercised. (`packages/agents/bob/harness-chat.sh:13`)
  **Fix:** A closing keyword may only ship when every "Done when" item is demonstrated or explicitly re-scoped in the issue; when a criterion cannot be exercised, drop the keyword to a plain reference and leave the issue open on the outstanding check.
- 🟡 **Warning:** `BOB_TEAM_ID` becomes terminal-only in this PR, and the README and PR body now say so — but the provider card the user actually sets it on still promises it unconditionally, its hint reading "Sets the x-team-id header" with no caveat. The neighbouring `BOB_INSTANCE_ID` hint two entries above does carry the narrowing ("chat sessions ignore it"), so the card now contradicts itself. (`packages/api-server/src/modules/connections/domain/catalog.ts:210`)
  **Fix:** When a pin stops applying to a surface, state the narrowing on every surface that offers the pin — the provider-card hint included, not only the agent README and PR body; a user-visible setting that silently no-ops on the surface most agents use is not "called out".
- 🟡 **Warning:** `docs/architecture/persistence.md` is edited in this diff — the `$HOME` bullet is rewritten from "the cold-start source for `session/load`" to "what a cold re-attach reads … whichever verb the harness advertises" — but the header still reads `Last verified: 2026-09-02`, while the sibling `agent-lifecycle.md` was correctly bumped to `2026-09-04` in the same diff. `docs/guidelines/documentation-guidelines.md:34` requires the stamp bumped whenever the page is edited. (`docs/architecture/persistence.md:3`)
  **Fix:** Treat the `Last verified:` stamp as part of the edit, never a follow-up — bump it in the same PR for every page the diff touches under `docs/architecture/`.
- 🟢 **Suggestion:** `readExistingSettings` catches every error and returns `{}`, so a `settings.json` that exists but cannot be read or parsed (a partial write, wrong owner or mode on the PVC) is replaced by a file holding only the platform pins — the agent's persisted settings are gone with no signal, and the merge-write the README describes is quietly a full overwrite. A recovery default is right for the one error that means "no state yet" (`ENOENT`); reporting the rest through the enclosing step's stderr line and non-zero exit would keep the new failure path honest. (`packages/agents/bob/bob-settings.mjs:43`)
- 🟢 **Suggestion:** Every pin is written conditionally and nothing removes it, so clearing `BOB_SHELL_MODEL` or `BOB_CHAT_MODE` on an existing agent leaves the previous value in `settings.json` forever — the PVC keeps a pin the provider card shows as empty. `RETIRED_APPROVAL_KEYS` already solves this shape for the retired keys; the same delete-when-absent rule applied to the three `session` pins would make the settings file a projection of the env rather than an accumulation of every value ever pinned. (`packages/agents/bob/bob-settings.mjs:64`)
- 🟢 **Suggestion:** `Number(process.env.BOB_MAX_COINS ?? process.env.BOB_MAX_COST)` falls through only on `undefined`, so a `BOB_MAX_COINS` present but empty yields `Number("") === 0`, fails the `> 0` guard, and also suppresses the `BOB_MAX_COST` fallback. The provider's `bobEnvMappings` trims and skips blanks, so this needs a hand-set env var to hit — but the sibling reads of `BOB_CHAT_MODE` and `BOB_SHELL_MODEL` both `?.trim()` first, so the three env reads disagree on what "unset" means. A first-non-blank helper would settle it. (`packages/agents/bob/bob-settings.mjs:57`)
- 🟢 **Suggestion:** The rehydrate response block widened from `mapping.method === "session/load" && mapping.attachSessionId` to `if (mapping.attachSessionId)` with the load case demoted to `else if`, so correctness now rests on the unenforced fact that only two construction sites set `attachSessionId`; a third with a different method would fall silently through both branches rather than fail to compile. Modelling the attach route as a discriminated union on `OutboundMapping` would make breaking it a type error. (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:585`)
- 🟢 **Suggestion:** No test covers the case the new `teardownRuntime` reset exists for — a resume-capable harness recycled and replaced by one advertising neither capability, where a stale `true` would send a `session/resume` the new harness cannot answer. The new test exercises only the capability-present path. (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:338`)
- 🟢 **Suggestion:** Carried over: the SSH paragraph in `agent-lifecycle.md` is rewritten wholesale — the `sshd -i` spawn, key authorization, `~/.ssh/environment` rebuild, WS ping/pong and orphaned-pin reconcile detail condensed into cross-references — which neither #3552 nor the PR body asks for or declares. The detail does survive in `cli.md` and `platform-topology.md`, so this is scope rather than content loss. (`docs/architecture/agent-lifecycle.md:165`)

### Verified clean

- **Comment convention.** 17 comment lines added across 264 added lines; every one is a typed comment (`TEST_OVERVIEW:`, `TEST_SCENARIO:`, and one inside a `UNIT_BOUNDARY_DESCRIPTION:` block), so all survive `scripts/strip-comments.mjs`. No violation.
- **Deleted-writer key set.** The bridge's `ensureSettings()` at `origin/main` vs. the new bootstrap: three approval keys pruned, `outsideWorkspaceAllowed` retained, `licenseConsent` dropped but made inert by `--accept-license`. No other key is inherited.
- Both harness scripts pass `--trust --accept-license`, the ACP surface's flag set matches the README's list, and `bobShell.autoUpdate: false` is written as documented.

The four remaining warnings are the approval bar.


- `packages/api-server/src/modules/connections/domain/catalog.ts:210` — 🟡 **Warning:** `BOB_TEAM_ID` becomes terminal-only in this PR, but this hint still promises it unconditionally, while the `BOB_INSTANCE_ID` hint two entries above does carry the narrowing — so the card contradicts itself on the surface the user actually sets the pin on.
    **Fix:** When a pin stops applying to a surface, state the narrowing on every surface that offers it, the provider-card hint included, not only the agent README and PR body.
- `docs/architecture/persistence.md:3` — 🟡 **Warning:** This page's `$HOME` bullet is rewritten in this diff, but the stamp still reads `2026-09-02`; `agent-lifecycle.md` was correctly bumped to `2026-09-04` in the same diff. `docs/guidelines/documentation-guidelines.md:34` requires the bump whenever the page is edited.
    **Fix:** Treat the `Last verified:` stamp as part of the edit, never a follow-up — bump it in the same PR for every page the diff touches under `docs/architecture/`.


---

xjacka:
Round two answered at `db75f9c1`. Six fixed, two declined with reasons.

**Fixed**

- **The `||` warn is gone.** You are right that this was the wrong shape: without `approval.outsideWorkspaceAllowed` Bob refuses every tool touching `$HOME`, so warning left a pod that reports healthy and fails every session at its first tool call. Both harness scripts now `exit 1` when the settings write fails. The rules link stays soft-fail, since its absence costs only the platform instructions — so the bootstrap exits non-zero for `writeSettings` and merely reports for `ensureRules`.
- **`readExistingSettings` no longer swallows everything.** Only `ENOENT` means "no state yet"; an unreadable or corrupt file is reported and aborts, instead of being replaced by a file holding just the platform pins. Verified: a `settings.json` containing `not json` leaves the file untouched and exits 1.
- **The session pins are a projection of the env now.** Clearing `BOB_SHELL_MODEL`, `BOB_CHAT_MODE` or `BOB_MAX_COINS` deletes the key rather than leaving the last value on the PVC. Verified: a file pinning `model` and `defaultMode` comes back with both gone and Bob's own `session.maxTurns` untouched.
- **One first-non-blank helper for the three env reads**, so an empty `BOB_MAX_COINS` no longer reads as `0` and no longer suppresses the `BOB_MAX_COST` fallback.
- **The provider card states the narrowing.** `BOB_TEAM_ID`'s hint now reads like its `BOB_INSTANCE_ID` neighbour: "Applies to terminal (TUI) sessions; chat sessions ignore it." Good catch — the card contradicting the README was the part that actually reaches a user.
- **`persistence.md`'s stamp is bumped** to 2026-09-04, same as its sibling.

**Fixed, with a note on the test you asked for.** The capability reset is now covered — but my first attempt at that test passed with the reset removed, so it proved nothing. The reason is that `initialize` assigns both flags on every response, so a recycle followed by a normal initialize can never observe the stale value. The window is narrower: a client that prompts into a cold session after a recycle *without* a new initialize. The committed test does that, and I verified it both ways — green with the reset, red without it.

**`Closes` dropped to a reference.** You are right, and this reverses my previous answer: a closing keyword promising a criterion nobody exercised is not the same as demonstrating it. The body now says `Part of #3552` and states which item is outstanding, so the issue stays open until a prompt turn is run against a cluster with a granted Bob key.

**Declined: the SSH paragraph — now declared instead.** The rewrite is deliberate, and the objection that it was undeclared is fair, so the PR body carries a *Declared out-of-scope edit* section explaining it: the page sat 32 characters under the 40 000-char cap, the one-sentence correction this PR needs could not land without freeing room, and the documentation guidelines prescribe raising the level and cutting duplicated depth over rewording to fit. The detail is owned in depth by `cli.md`, `platform-topology.md` and the page's own Hibernate section, all cross-referenced.

**Declined: the discriminated union on `OutboundMapping`.** The suggestion is sound in the abstract, but both construction sites are in this file, a few lines apart, and re-modelling a shape shared by every outbound frame is a refactor with a blast radius well beyond an agent upgrade. Left for whoever next touches that routing on purpose.

dam-code-guardian (dismissed):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `db75f9c`

### Summary
Round 3 fixes all four blocking findings at their cause: both launchers now fail on a bad settings bootstrap, `readExistingSettings` only tolerates `ENOENT`, the provider-card hint names the terminal-only scope, and the edited architecture page carries a current stamp. Nothing blocking is left.

### Changes since last review
- ✅ **Fixed:** a failed settings bootstrap is swallowed by the launcher (`packages/agents/bob/harness-chat.sh:5`)
- ✅ **Fixed:** `Closes #3552` lands with the prompt turn unexercised (`packages/agents/bob/harness-chat.sh:13`)
- ✅ **Fixed:** BOB_TEAM_ID narrowing is missing from the provider-card hint (`packages/api-server/src/modules/connections/domain/catalog.ts:210`)
- ✅ **Fixed:** edited architecture page keeps its old `Last verified` stamp (`docs/architecture/persistence.md:3`)
- ✅ **Fixed:** unreadable `settings.json` is silently replaced rather than reported (`packages/agents/bob/bob-settings.mjs:43`)
- ✅ **Fixed:** clearing a pin never un-sets the key it wrote on an upgraded PVC (`packages/agents/bob/bob-settings.mjs:64`)
- ✅ **Fixed:** the alias chain treats an empty `BOB_MAX_COINS` as a value (`packages/agents/bob/bob-settings.mjs:57`)
- ✅ **Fixed:** no test covers the capability reset across a harness recycle (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:338`)
- ✅ **Fixed:** undeclared scope: unrelated SSH paragraph rewrite — now declared in the PR body (`docs/architecture/agent-lifecycle.md:165`)
- 🔁 **Still present:** widened `attachSessionId` guard rests on an unenforced invariant (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:585`) — held by the author, blast radius named; PR-local override.

### Findings
- 🟢 **Suggestion:** the recycle test asserts that `session/resume` is absent, but it never asserts that a recycle happened, so a `refreshEnv` that becomes a no-op keeps the test green (`packages/agent-runtime/src/modules/acp/services/acp-runtime/__tests__/acp-runtime-history-provider.test.ts:163`)
- 🟢 **Suggestion:** `session/resume` is now a cold-start read, so the wedge rule that names only an unanswered `session load` is incomplete (`docs/architecture/agent-lifecycle.md:139`)
- 🟢 **Suggestion:** off-delta — the UI Team ID field keeps a hint with no terminal-only narrowing, unlike its Instance ID neighbour; a follow-up issue fits better than this PR (`packages/ui/src/modules/providers/components/bob/form.tsx:167`)

### Doc drift
- 🟢 **Suggestion:** `docs/architecture/agent-lifecycle.md:139` describes the wedge condition for `session load` only; add the resume verb.

### Issue fit
- 🟢 **Suggestion:** #3552 asks for the native ACP agent; the `BOB_TEAM_ID` chat-path narrowing is a related regression fix that the PR body now declares, so scope reads honest.

### TypeScript engineering
- 🟢 **Suggestion:** the cost parse (`bob-settings.mjs:57`) now rejects blank, non-numeric, non-finite and non-positive input — checked against `""`, `"abc"`, `"1e400"`, `"0"` and `"2.5"`.
- 🟢 **Suggestion:** remaining items are the two carried above (recycle-test strength, `attachSessionId` invariant); the rest of the skill output repeats findings this round fixed.

### Verdict
APPROVE — round 3 closes every blocking finding; what is left is 🟢 or held by the author.




---

xjacka:
## Scope changed substantially since the approval — re-review requested

This PR is no longer the change that was approved, and the standing `APPROVED` should not be read as covering it. Flagging explicitly rather than letting a green check speak for a diff it never saw.

**What was approved:** `db75f9c1`, 2026-09-04 — six commits, the Bob Shell 2.0.2 native-ACP upgrade plus its review fixes and the Config panel.

**What is here now:** twelve commits at `20a1f73e`. Beyond the ACP upgrade the branch carries two further work streams that landed on it directly:

- **The LiteLLM gateway rewrite** (`cb958477`, `b6e54947`, `a10183ca`, `77d33193`, `f1475caf`, `5f6046c7`) — egress path-prefix rewriting so Bob can run against the LiteLLM proxy, touching `envoy.go`, `envoy_bootstrap.go`, `envoy_test.go`, the connections catalog and providers, `connection-sds.ts`, the provider UI, plus `connections.md` and `security-and-credentials.md`. These commits also exist on `feat/bob-openai-endpoint`.
- **Named telemetry export rails** (`20a1f73e`) — `telemetry: bob` on the Bob template and the env rail behind it, keeping Bob's prompt payloads off the collector. This one had never been pushed anywhere before this branch.

Neither has been through a review round in this PR. The decision to carry them here rather than split them back out was the owner's, made deliberately with the stale-approval problem on the table — so what this needs is a fresh round over the whole diff, not a rubber stamp on the delta.

**Conflict resolutions worth checking**, both against `c312d195`:

- `deploy/helm/platform/values.yaml` — main added `harness: bob` to the Bob template while the telemetry commit added `telemetry: bob`. Independent keys, both kept.
- `docs/architecture/persistence.md` — main grew the `$HOME` bullet again (session directory). Took main's version and re-applied only the cold-start sentence, so the new content survives; stamp bumped.

**Verification.** Full `mise run test` is green across all seven packages (agent-runtime 303, api-server 1335, cli 154, plus controller, ui, driver-sdk, experiment-sdk). Every `check` task passes locally except `controller:check:staticcheck`, which cannot run on this machine — the installed binary is built against go1.25 while the toolchain is go1.26, so it fails on the standard library rather than on this code. CI is the authority for that one.

Still outstanding from #3552, unchanged: a real prompt turn against a cluster with a granted Bob key. That is why the body says `Part of` and not `Closes`.

packages/agents/bob/README.md
```diff
@@ -45,42 +89,50 @@ These ride on the secret's `envMappings`, so every agent granted the Bob secret
...
+
+- **The key needs more than model access.** Bob asks the proxy for its model list before it will start a session, and that is a management route, not a model call — a LiteLLM virtual key scoped to LLM API routes alone is refused there and Bob fails to start. The team behind the key also has to carry the model Bob is set to use, or the first message comes back refused.
+- **A model has to be chosen.** Bob's built-in default resolves to a tier alias only its own gateway serves, so until one is chosen the proxy rejects it as unknown. The connection deliberately contributes no model of its own — it would collide with the Bob Shell connection's pin, which claims the same env name, and an agent can hold both. Pick one in the agent's Config panel instead: the list there is read live from this same proxy, so it offers what the granted key actually serves and can be changed at any time. `BOB_SHELL_MODEL` in **Configure Agent → Env** still works as a per-agent override.
```
dam-code-guardian: 🟡 **Warning:** `f1475ca` removed the Bob model field from the LiteLLM connect dialog, but the prose written for that field stayed. This bullet tells the user to pick a model in the agent's Config panel and says the list there is read live from this proxy — while README:83 and `runtime-manifest.yaml:36` both state that model is deliberately absent from the panel, because Bob's list needs a `/model/info` route neither the static catalog nor `modelDiscovery` can serve. The instruction points at a control the branch does not ship. Same defect at `docs/architecture/connections.md:196`, which still describes the model as a required Connection config input that the same commit deleted. Reported by Issue Fit and Documentation Check.
  **Fix:** A revert inside a branch is not complete until the prose written for the reverted feature is swept in the same commit — documentation describes the diff's end state, never an intermediate one.

docs/architecture/connections.md
```diff
@@ -177,6 +177,29 @@ gateway-side upstream validation only.
...
+Bob is the case that needs more than a base URL. It prefixes every
+gateway call with its own service path, and that prefix is not
+configurable, so the Connection also declares the prefix rewrite the
+gateway applies; the auth scheme needs no such treatment, because the
+gateway overwrites the credential header outright. What the endpoint has
+to offer for this to work is the OpenAI-shaped chat-completion route, a
+model-information route reporting a non-empty model list (Bob treats an
+empty list as an error), and no region hint of its own — a gateway
+answering the profile route with one would redirect Bob's inference calls
+off the configured host. A model has to be named on the Connection as
```
dam-code-guardian: 🟡 **Warning:** This paragraph states that a model has to be named on the Connection as a config input, but `f1475ca` deleted that input — the catalog entry now carries only env, egress-inject and the prefix rewrite, and a unit test asserts the absence. A reader following this page looks for an input that no longer exists. Same defect at `packages/agents/bob/README.md:105`. Reported by Documentation Check and Issue Fit.
  **Fix:** A revert inside a branch is not complete until the prose written for the reverted feature is swept in the same commit — documentation describes the diff's end state, never an intermediate one.

packages/agents/bob/bob-settings.mjs
```diff
@@ -0,0 +1,135 @@
...
+
+function readExistingSettings() {
+  let raw;
+  try {
+    raw = readFileSync(SETTINGS_PATH, "utf8");
+  } catch (err) {
+    if (err.code === "ENOENT") return {};
+    throw err;
+  }
+  const parsed = JSON.parse(raw);
```
dam-code-guardian, line 48: 🟢 **Suggestion:** `JSON.parse` sits outside the `try`, so an existing but syntactically corrupt `settings.json` throws; `writeSettings` runs `required: true` and both launchers' `|| exit 1` then refuse to start the harness at all — a file defect on the PVC becomes a pod that never serves a session. This is the composition of two things earlier rounds asked for (a load-bearing bootstrap must fail the process; only `ENOENT` may recover silently), so it reads as the intended posture rather than an oversight. Worth one line only because the failure is silent to the user: the stderr line lands in the pod log and nothing surfaces on the agent. Reported by TypeScript Engineering Review.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `20a1f73`

🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `20a1f73e`

## Re-review

### Summary
Four new commits add a third work stream — named telemetry export rails, so a harness that does not write Claude Code's per-call log records can still reach the collector — and drop the Bob model field from the LiteLLM connect dialog. The rail itself is well built: the dispatcher fails the Helm render on an unknown rail name rather than silently emitting the wrong env, and the bootstrap asserts payload exclusion on every harness start instead of trusting a clickable setting. What did not travel with either change is its documentation. The model revert left prose describing the control it removed, and the export rail is absent from every page that owns the export path. _11 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: `db75f9c1` (2026-09-04T14:23:02Z) — verdict APPROVE — unreachable, reviewed the whole PR

The branch was force-pushed between rounds (`db75f9c1...20a1f73e` compares as `diverged`), so no commit range spans the two rounds and this round re-derived every finding at HEAD.

- 🔁 **Still present:** the recycle test asserts only the absence of `session/resume`, not that a recycle happened (`packages/agent-runtime/src/modules/acp/services/acp-runtime/__tests__/acp-runtime-history-provider.test.ts:163`)
- 🔁 **Still present:** the wedge rule names only an unanswered `session load`; `session/resume` is now also a cold-start read (`docs/architecture/agent-lifecycle.md:139`)
- 🔁 **Still present:** off-delta — the UI Team ID hint carries no terminal-only narrowing (`packages/ui/src/modules/providers/components/bob/form.tsx:167`)
- 🔁 **Still present:** widened `attachSessionId` guard rests on an unenforced invariant (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:585`) — suppressed by a PR-local override.
- 🆕 **New:** 3 🟡, 1 🟢 — below.

### Findings

- 🟡 **Warning:** `f1475ca` removed the Bob model field from the LiteLLM connect dialog, but the prose written for that field stayed. The README now tells the user to pick a model in the agent's Config panel, "the list there is read live from this same proxy" — while README:83 and `runtime-manifest.yaml:36` both state that model is deliberately absent from the panel, because Bob's list needs a `/model/info` route neither the catalog nor `modelDiscovery` can serve. `docs/architecture/connections.md:196` still describes the model as a required Connection config input that the same commit deleted. A reader following either page looks for a control the branch does not ship. Reported by Issue Fit and Documentation Check. (`packages/agents/bob/README.md:105`, `docs/architecture/connections.md:196`)
  **Fix:** A revert inside a branch is not complete until the prose written for the reverted feature is swept in the same commit — documentation describes the diff's end state, never an intermediate one.
- 🟡 **Warning:** The export rail is undocumented on every page that owns it. `observability.md:46` still describes agent export as a boolean flag wired for "the **Claude Code harness only**", though `telemetry` is now the *name* of a rail with a second implementation. Line 51 states as a guarantee that the export env names the OTel service after the template — the Bob rail cannot do that, since the service name is fixed inside the harness and `OTEL_SERVICE_NAME` is not read, so every template off that image collapses to one name. Line 50 states that content bodies are not exported; for this rail that now holds only because `bob-settings.mjs:97` writes `excludePayload: true`, which no page or README mentions — the Bob README's env table (`:122`) lists `IBM_TELEMETRY_ENABLED` and no `BOB_TELEMETRY_*` at all. The new chart comment points readers at `metrics.md` for the Bobcoin-denominated counter, and that word appears nowhere in the repository. Reported by Documentation Check. (`docs/architecture/observability.md:46`, `:50`, `:51`, `packages/agents/bob/README.md:122`)
  **Fix:** When a values-level flag gains a second implementation, the same PR updates the page that owns the flag, restates per-rail every property the page asserts globally, and adds the new env family to the agent's own reference table — a cross-reference that resolves to nothing counts as unwritten.
- 🟡 **Warning:** Two edited architecture pages keep a stale `Last verified:` stamp. `security-and-credentials.md` gains a whole "Path rewriting" subsection plus rewrites of two further passages, and keeps its base-branch `2026-09-04`. `connections.md` was bumped to `2026-09-04` by the commit that added the preset, then edited again by `a10183ca` on 2026-09-07 with no second bump, so its stamp is older than its own newest content. `docs/guidelines/documentation-guidelines.md:34` requires the bump whenever the page is edited. The sibling pages in this diff are correct — `agent-lifecycle.md` reads `2026-09-08` and `persistence.md` was bumped to it — so the rule is being followed unevenly rather than not at all. Reported by Documentation Check. (`docs/architecture/security-and-credentials.md:3`, `docs/architecture/connections.md:3`)
  **Fix:** Re-bump the stamp on every commit that touches the page, review-fix commits included; one bump early in a branch does not cover later edits to the same page.
- 🟢 **Suggestion:** `readExistingSettings` parses outside its `try`, so an existing but syntactically corrupt `settings.json` throws, `writeSettings` runs `required: true`, and both launchers' `|| exit 1` refuse to start the harness at all — a file defect on the PVC becomes a pod that never serves a session. This is the composition of two things this review asked for (round 2: a load-bearing bootstrap must fail the process; round 3: only `ENOENT` may recover silently), so it is the intended posture rather than an oversight — Bob's approval gate is unknown at that point and starting anyway is what round 2 blocked. Worth one line only because the failure is silent to the user: the stderr line lands in the pod log and nothing surfaces on the agent. Reported by TypeScript Engineering Review. (`packages/agents/bob/bob-settings.mjs:48`)

### Verified clean

- **The rail dispatcher fails loudly.** An unrecognised `telemetry` value calls `fail` with the consequence named, so a typo breaks the render rather than shipping an agent whose spend never arrives. Only `claude-code`, `nous` (`true`) and `bob` (`bob`) set the field, and the sole caller threads it through.
- **Payload exclusion is asserted, not offered.** `excludePayload: true` is written on every harness start and is absent from the Config panel catalog, so an agent cannot turn it off — which is what makes the platform's "content bodies are not exported" invariant hold for a harness that defaults it to false.
- **The rail reuses the existing network path.** `BOB_TELEMETRY_URL` targets the same collector host and port as the Claude Code rail over the agent's ordinary `HTTPS_PROXY` egress, so the gateway's `terminate_otel_collector` chain stamps the same trusted `x-platform-agent-id`. No new egress and no injected credential; the `unused` Langfuse key pair never leaves the mesh.
- **Comment convention.** Every comment line this diff adds to a `.ts`, `.tsx` or `.go` file is a typed comment (`TEST_OVERVIEW:`, `TEST_SCENARIO:`, or a doc block), so all survive `scripts/strip-comments.mjs`. No violation.
- **Doc caps.** `security-and-credentials.md` lands at 39 995 characters of the 40 000 cap and every other touched page is well under.
- **`BOBSHELL_API_KEY` from two templates is harmless.** Both contributions carry the identical `dummy-placeholder` string, so merge order cannot change the result — refuting the reported ambiguity.

The three warnings are the approval bar; all three are documentation the code changes did not carry with them.

- 🟡 3 warnings, merged into `### Findings` above: the orphaned model prose, the undocumented export rail (three separate claims on `observability.md` plus the dangling Bobcoin pointer), and the two stale `Last verified:` stamps.

### Issue Fit
Linked issue: #3552 — plain reference, closing keyword deliberately absent.

- 🟡 The orphaned model prose fails the "docs describe the new setup" item; merged into `### Findings` above.
- The tenant-scoping narrowing and the unreachable pre-2.0.2 chats are both declared in the PR body and the README, and no closing keyword will close the issue on an unmet criterion — so scope reads honest. Suggestions on both dropped under the 🟢 budget.

- 🟢 One suggestion kept in `### Findings` above (the corrupt-settings start failure). The reported domain-layer objections do not hold against this repository: `build-connection.ts` throws throughout, and more than twenty `domain/` files import from `api-server-api`, so `assertContributionsParse` follows the surrounding convention rather than breaking it. Remaining suggestions dropped under the 🟢 budget.

### Verdict
REQUEST_CHANGES — the telemetry rail and the model revert are both sound as code, but neither carried its documentation: one page instructs the user to use a control this branch removed, and the page that owns agent export still describes a single-harness boolean.


---

docs/architecture/observability.md
```diff
@@ -43,12 +43,15 @@ Making the mesh the sole gate is why the collector is platform-owned. ClickStack
...
+- **Signals.** Claude Code exports metrics, logs, and traces (the last via the enhanced-telemetry beta) over OTLP/HTTP, with one **log record per LLM call**. Bob Shell exports **traces only**, and its per-call counters ride a generation span under GenAI semantic-convention names — there is no log-record equivalent to fall back to, which is why the [metrics](metrics.md) read path normalizes both shapes rather than reading one. Bob additionally gates those spans behind its own agent-operations switch, so the rail turns it on explicitly; without it the harness exports its lifecycle spans and none of the spend.
```
dam-code-guardian: 🟡 **Warning:** This line states that Bob's per-call counters ride a generation span "which is why the [metrics](metrics.md) read path normalizes both shapes rather than reading one". No such normalization exists at this HEAD: `clickhouse-reader.ts` queries `otel_logs` in all 12 of its store reads, nothing under `packages/` reads `otel_traces`, `SpanAttributes` or `gen_ai`, and this branch changes no metrics code. The page that owns the read path still says the reader "reads the per-LLM-call log records Claude Code exports" (`docs/architecture/metrics.md:71`), so the rail this PR ships delivers spend telemetry no Usage surface reads, while an architecture page asserts the opposite.

**Fix:** An architecture page describes the PR's own HEAD — when a cross-page claim is not true of the code in the same PR, state the gap on both pages instead, and amend the page that owns the concept rather than the one that mentions it.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `64e6d12`

## PR #3553: feat(bob): run Bob Shell 2.0.2 as a native ACP agent
**Author:** xjacka | **Branch:** feat/bob-acp-native → main | **Changes:** +781 −713 (29 files)

### Summary
All three blockers from the last round are closed at HEAD: the orphaned model prose is rewritten on both pages, both stale `Last verified:` stamps are bumped, and `observability.md` now documents the export rail per-harness rather than as a Claude-Code boolean. What the rail documentation added is one forward-looking claim — that the metrics read path normalizes Bob's span shape — which no code in this branch performs, and the provider card still names two CLI flags this PR removed. _12 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: `20a1f73` (2026-09-08T13:44:23Z) — verdict REQUEST_CHANGES — unreachable, reviewed the whole PR

The branch was force-pushed again between rounds (`20a1f73...64e6d12` compares as `diverged`), so no commit range spans the two rounds and this round re-derived every finding at HEAD.

- ✅ **Fixed:** prose left behind by the in-branch model-field revert — `64e6d12` rewrote `connections.md:196` and the README now points at `BOB_SHELL_MODEL`, not a Config-panel control (`packages/agents/bob/README.md:105`)
- ✅ **Fixed:** two edited architecture pages kept a stale `Last verified:` stamp — `security-and-credentials.md` reads `2026-09-09`, `connections.md` `2026-09-08` (`docs/architecture/security-and-credentials.md:3`)
- ✅ **Fixed:** the export-rail documentation gap — `edd1508` restates per-rail every property the page had asserted globally, and the README env table now carries `BOB_TELEMETRY_*` (`docs/architecture/observability.md:46`)
- ✅ **Fixed:** the recycle test now asserts the kill and the replacement, not only the absent verb (`packages/agent-runtime/src/modules/acp/services/acp-runtime/__tests__/acp-runtime-history-provider.test.ts:183`)
- ✅ **Fixed:** the wedge rule now names the re-attach generically — "whichever verb it advertised for one" (`docs/architecture/agent-lifecycle.md:138`)
- ✅ **Fixed:** the UI Team ID hint carries the terminal-only narrowing (`packages/ui/src/modules/providers/components/bob/form.tsx:167`)
- 🔁 **Still present:** a corrupt `settings.json` refuses the harness start silently (`packages/agents/bob/bob-settings.mjs:48`) — 🟢, and the intended posture this review asked for.
- 🆕 **New:** 2 🟡, 1 🟢 — below.

### Findings

- 🟡 **Warning:** `observability.md:52` states that Bob's per-call counters ride a generation span "which is why the [metrics](metrics.md) read path normalizes both shapes rather than reading one". No such normalization exists at this HEAD: `clickhouse-reader.ts` queries `otel_logs` in all 12 of its store reads, nothing under `packages/` reads `otel_traces`, `SpanAttributes` or `gen_ai`, and this branch changes no metrics code. The page that owns the read path still says the reader "reads the per-LLM-call log records Claude Code exports" (`docs/architecture/metrics.md:71`), so the rail this PR ships delivers spend telemetry no Usage surface reads, while an architecture page asserts the opposite. Reported by Documentation Check. (`docs/architecture/observability.md:52`)
  **Fix:** An architecture page describes the PR's own HEAD — when a cross-page claim is not true of the code in the same PR, state the gap on both pages instead, and amend the page that owns the concept rather than the one that mentions it.
- 🟡 **Warning:** Two provider-card hints still tell the user their value becomes a CLI flag, and this PR removed both flags: `harness-terminal.sh` no longer passes `--max-cost` or `--mode` — they ride the settings file `bob-settings.mjs` writes. `form.tsx:173` reads "BOB_MAX_COINS → --max-cost" and `:180` "BOB_CHAT_MODE → --mode"; the server-side hint for the same input names it too (`catalog.ts:219`, "Bob 2.x --max-cost"). The two neighbouring tenant hints were correctly updated in this same branch, so the sweep stopped one field short. Reported by React UI Engineering Review. (`packages/ui/src/modules/providers/components/bob/form.tsx:173`, `:180`, `packages/api-server/src/modules/connections/domain/catalog.ts:219`)
  **Fix:** A hint that names the mechanism a value flows through is a fact about another package — re-check every such hint in the PR that changes the mechanism, on the card and in the template both.
- 🟢 **Suggestion:** `runtime-delivery.md:55` states the `harness-config` event writes its mapped keys "once, and — like `workspace-seed` — never re-asserts, so the file stays the user's to edit". Bob's manifest deliberately maps into a `platform.*` section for exactly that reason, and the PR body explains it, so the mechanism is sound — but the page's platform-wide wording now holds only for the keys a harness does not itself own, since `bob-settings.mjs` re-derives `session.defaultMode`, `session.model` and `session.maxCost` on every start. Reported by Documentation Check. (`docs/architecture/runtime-delivery.md:55`)

### Verified clean

- **The three prior blockers close at their causes, not around them.** The model prose was rewritten rather than deleted (`connections.md` now says where the model comes from and why the Connection contributes none); the observability page restates per-rail the three guarantees that had been global, and names a property no rail had documented — that a malformed rail redirects rather than going inert. Both stamps were bumped in the commit that made the edit.
- **The stamp rule holds across the branch.** Every architecture page this PR edits carries a `Last verified:` at or after its newest edit in the branch: `agent-lifecycle.md` and `connections.md` at `2026-09-08` matching their last author date, `observability.md`, `security-and-credentials.md`, `persistence.md` and `runtime-delivery.md` at `2026-09-09`. Doc-drift read the rebase's committer dates for two of them; the author dates refute it.
- **Comment convention.** The diff adds no comment line to any `.ts`, `.tsx`, `.go` or `.mjs` file that is not a sanctioned prefix or a doc block. No violation.
- **The path-rewrite dedup is order-stable and the route list is not shadowed.** `rewriteByPrefix` keeps the first replacement per prefix and warns on a conflict; the rewrite routes precede the `prefix: "/"` catch-all, and the one shipped rewrite (`/inference/v1/` → `/v1/`) has no sibling that could shadow it. An allow-only secret contributing a rewrite before the `cred == nil` return is deliberate — a host may be rendered allow-only and still need its prefix mapped.
- **`anchoredPath` is enforced twice by design.** The Go validator (`envoy.go:262`) and the zod schema (`types.ts:49`) reject different supersets of the same invariant, but they guard different trust boundaries — the contract at contribution time, the annotation at render time, where a hand-edited Secret never passed through zod. The Go side is the stricter of the two.
- **The domain-layer objections do not hold against this repository.** `assertContributionsParse` importing a schema from `api-server-api` follows the surrounding convention: seven `domain/` files already call `.parse`/`safeParse`, and more than twenty import from that package. Throwing there matches `build-connection.ts`, which throws throughout.
- **`BOB_TEAM_ID` and the pre-2.0.2 chat ids are both recorded descopes.** Each is declared in the PR body and in the README (`:96`, `:136`), and the closing keyword is deliberately absent so #3552 stays open on its unexercised criterion — so neither is an issue-fit gap.

Both warnings are documentation the code did not carry with it; neither is a defect in the runtime or the image.

- 🟡 2 warnings, merged into `### Findings` above: the metrics-normalization claim, and the `harness-config` never-re-asserts wording (regraded 🟢 — the mechanism is sound and declared, the wording is what narrowed).
- The reported stale-stamp warning does not hold: both pages carry a stamp at or after their newest author date in this branch. See `### Verified clean`.

### Issue Fit
Linked issue: #3552 — plain reference, closing keyword deliberately absent.

- ✅ No blocking findings. The two reported criteria gaps — terminal-only tenant scoping and the unreachable pre-2.0.2 chats — are recorded descopes: both are declared in the PR body and the README, and no closing keyword will close the issue on an unmet criterion. Suggestions dropped under the 🟢 budget.

- ✅ No findings survive verification as blocking. The three reported layering warnings are regraded 🟢: the domain-layer objections are refuted above, and moving `hasSessionCapability` into `acp/domain/frames.ts` is a restructuring for readers, not a defect risk. The `MODES` duplication in `bob-settings.mjs` is not new — the deleted shim carried its own `AVAILABLE_MODES` list too. Suggestions dropped under the 🟢 budget.

- 🟡 1 warning, merged into `### Findings` above: the two stale flag hints.
- The remaining warnings are regraded 🟢: the hint-duplication and self-contradicting-copy findings both anchor on the Team ID line, which is a PR-local override this round, and the collapsed-Advanced submit gate is pre-existing at `form.tsx:150` — untouched by this diff. Suggestions dropped under the 🟢 budget.

### Verdict
REQUEST_CHANGES — every blocker from the last round is properly closed, but the rail documentation added a claim about a read path this branch does not implement, and two card hints still name flags it removed.


---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `0bdf1e3`

## PR #3553: feat(bob): run Bob Shell 2.0.2 as a native ACP agent
**Author:** xjacka | **Branch:** feat/bob-acp-native → main | **Changes:** +784 −716 (29 files)

### Summary
One commit closes the fifth round: the observability page now states what the metrics read path really does, and the Max cost / Mode hints name the settings keys `bob-settings.mjs` writes. Both prior warnings are resolved and nothing new blocks. _10 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: 64e6d12 (2026-09-09T08:41:17Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** page asserts a metrics normalization no code performs (`docs/architecture/observability.md:52`)
- ✅ **Fixed:** two card hints name CLI flags this PR removed (`packages/ui/src/modules/providers/components/bob/form.tsx:173`)
- 🔁 **Still present:** harness–config never–re–asserts wording now carries an unstated condition (`docs/architecture/runtime-delivery.md:55`)
- 🔁 **Still present:** a corrupt settings.json refuses the harness start silently (`packages/agents/bob/bob-settings.mjs:48`)
- 🆕 **New:** the catalog Mode hint gained no panel–precedence caveat, unlike its form twin (`packages/api-server/src/modules/connections/domain/catalog.ts:227`)
- 🆕 **New:** the new contributions guard runs Zod and throws inside `domain/` (`packages/api-server/src/modules/connections/domain/catalog.ts:838`)

### Findings
- 🟢 **Suggestion:** The form's Mode hint gained `Starting mode unless the agent Config panel sets one`, which matches `bob-settings.mjs:75` resolving `platform.mode` over `BOB_CHAT_MODE`. The catalog spec feeds the same field to the generic connection dialog (`connection-template.ts:319` copies each `configInputs` hint onto a client input) and kept the plain wording, so the precedence reaches one of the two surfaces. Reported by the diff review and react-ui-engineering (`packages/api-server/src/modules/connections/domain/catalog.ts:227`)

- 🔁 1 finding(s) from the previous review still present (see review at 64e6d12)

### Issue Fit
- 🔁 1 finding(s) from the previous review still present (see review at 64e6d12)

- 🟢 **Suggestion:** `assertContributionsParse` runs `contributionSchema.safeParse` and throws a raw `Error` inside `domain/`. The skill reads both as layer violations; `build-connection.ts` in the same directory already throws for template invariants at ~20 sites, so the throw follows the local convention and boot is the cheapest place to catch a malformed template. Recorded as taste, not a defect (`packages/api-server/src/modules/connections/domain/catalog.ts:838`)

- 🟢 **Suggestion:** After this commit the five hints carry two kinds of target behind the same `→` arrow — `--instance-id` / `--team-id`, which `harness-terminal.sh:8` really passes as flags, and `session.maxCost` / `session.defaultMode`, which are settings keys. Both halves are accurate, so the arrow now means two things in one block (`packages/ui/src/modules/providers/components/bob/form.tsx:180`)

### Verdict
APPROVE — both blocking findings of the last round are fixed at their anchors, and the delta adds no blocker.


---

PR #3559, opened 2026-09-04
fix(deps): bump vulnerable npm, Go and pip dependencies

JanPokorny:
Addresses #3534 and Dependabot alerts #107, #108.

## Fixed

| Finding | Fix |
|---|---|
| gRPC-Go heap exhaustion, CVE-2026-84304 (controller) | `google.golang.org/grpc` 1.82.1 → 1.83.2, `go mod tidy` |
| fast-uri CVE-2026-84292, CVE-2026-84394 | override → 3.1.7; two-day-old release, so a dated `minimumReleaseAgeExclude` entry |
| js-yaml CVE-2026-84375 | override → 4.3.2 |
| hono CVE-2026-84363/84364/84365 | direct dep in api-server → ^4.13.5 |
| @humanfs/node symlink copy (Dependabot #108) | new override → 0.16.8 |
| pip CVE-2026-13346 (agent images) | platform-base Dockerfile pin → 26.2.0 |

## Not fixable here

- **Go findings in agent images** come from the `gh` binary in platform-base. `gh` is now pinned to 2.100.0 in `packages/platform-base/mise.toml` (an explicit version bypasses mise's 7-day floor), which ships grpc 1.83.2 and x/crypto 0.56.0. It still ships x/mod 0.39.0 (< 0.40.0), so CVE-2026-56864/56865 stay open until gh bumps.
- **Keycloak** netty/opentelemetry: 26.7.3 is the newest tag and uses the same Quarkus 3.33.3.1 as our 26.7.2, whose BOM pins netty 4.1.136.Final. A bump would not fix it; pin left as is.

## Also: controller-gen pin

`mise check` failed on `main` and every PR at `controller:check:crd-generated`: controller-gen was `latest` and never locked, so its 0.22.0 release (2026-09-02) drifted from manifests stamped by 0.21.0. Pinned to 0.22.0 in `mise.toml` and `mise.lock` and regenerated; the only manifest change is the version annotation.

## Verified

- api-server tsc + 1305 unit tests, lockfile-age check, `pnpm audit` clean
- controller build, tests, govulncheck clean

PR #3560, opened 2026-09-04
build: pin controller-gen 0.22.0 and regenerate the CRD manifests

JanPokorny:
## Problem

`mise check` fails on `main` and on every open PR at `controller:check:crd-generated`. controller-gen was declared as `latest` and the go backend never locked a concrete version, so the 0.22.0 release on 2026-09-02 silently changed the generator everywhere, while the committed CRD manifests still carry the 0.21.0 annotation.

## Fix

- Pin `go:sigs.k8s.io/controller-tools/cmd/controller-gen` to `0.22.0` in `mise.toml` and `mise.lock`, so the generator version is reproducible and the gate cannot drift again on a release.
- Regenerate with `mise run controller:generate`. The only change in both manifests is the `controller-gen.kubebuilder.io/version` annotation; the schemas are identical.

`controller:check:crd-generated` passes locally.

### DIFF
--- a/deploy/helm/platform/crds/agent-platform.ai_agents.yaml
+++ b/deploy/helm/platform/crds/agent-platform.ai_agents.yaml
@@ -5,7 +5,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     agent-platform.ai/crd-schema-generation: "9"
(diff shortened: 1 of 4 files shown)


JanPokorny:
Folded into #3559.

PR #3562, opened 2026-09-04
feat(metrics): report Bob Shell spend on the Usage surfaces

xjacka:
Rebased onto `main` after #3553 merged — the named `telemetry` rails and Bob's payload exclusion landed there. This PR is the read path that turns what Bob exports into numbers on the Usage surfaces, plus the two fixes a live run demanded (see the last section).

## Problem

Bob's spend never reached Usage even with telemetry enabled, because Bob exports **traces only**: one `LLM Generation` span per API call, counters under GenAI semantic-convention names. The metrics reader was built around Claude Code's per-call **log records** and would find nothing.

## What changed

**Read path.** The reader normalizes both harness shapes into one per-call record; every query runs over their union. Adding a harness is one more row shape, not a change to any query above it.

**Cost is not one currency.** Claude Code reports dollars; Bob reports **Bobcoins**, the credit its IBM gateway bills in, and the platform has no authority to price one in the other. Every rollup now returns the dollar total alongside per-unit credit totals and never sums across units. Usage, Home and the CLI show credits beside dollars, scale comparison bars only against rows sharing a unit, and give each unit its own daily chart. A credit-billed agent reads as its own total rather than as `$0`.

**Per-session-kind rollup.** It landed on `main` while this was in flight and read the log records directly, so Bob's spend would have counted in the total above it while showing as zero in the breakdown — the disagreement that rollup's own unattributed bucket exists to prevent. It now reads the same union and carries credits like every other rollup.

## What Bob actually exports

Verified against the Bob Shell 2.0.2 bundle in the published image, not inferred from its docs:

| reader needs | Bob |
|---|---|
| owner attribution, invocation id | collector-stamped, unchanged |
| session id | `langfuse.session.id` — equals the ACP session id |
| trace id | native; the call span is a child of the agent loop |
| model, timestamp | present |
| input / output / cache-read / cache-creation tokens | present |
| request latency | present, but see below |
| cost | present, **in Bobcoins** |
| request id | **absent** |
| exported display name | **absent** — Bob reads no standard OTel resource env |

`session/new` returns the root task id and the span carries that same id, so the sessions sidebar's cost-per-session line matches the platform's session identity with no translation.

The missing display name only affects a **deleted** Bob agent, which falls back to its id in the per-agent breakdown; live agents are labelled from the platform registry.

**Latency does not come from the GenAI duration attribute.** The convention declares that attribute a double in seconds; Bob populates it with integer milliseconds. Reading it would have worked only by depending on the harness violating its own convention, and would have silently reported zero the moment Bob conformed. The reader uses the span's own exporter-recorded duration instead — a unit the exporter guarantees beats a named one the platform cannot.

## Decisions worth keeping

- **The credit decoder raises rather than degrades.** An unrecognised driver payload surfaces as an error instead of reading as "this agent spent nothing" — a silent zero in a spend column is worse than a failure.
- **Mixed-unit lists rank per unit.** Home's top spenders interleave per unit, so a credit-billed agent is reachable past dollar-spending ones; ranking a mixed list on one unit's magnitude makes the other unit unreachable.
- **Bars disclose their denominator.** A bar scaled against the top row *in its own unit* says so in its accessible label, because the same percentage means something different per unit.
- **Exact formatters for reconcilable figures.** Compact notation stays on space-constrained labels; a total a user checks against a bill is exact.

## Verification, and what is still unverified

`api-server` 1337 tests, `ui` 362 tests, typecheck across api-server / api-server-api / ui / cli, prettier, lint, `common:check:comment-types`, `helm:check`, doc-size — all green.

**Live verification (2026-09-15)** on a local k3s cluster with `clickstack.enabled` and a granted Bob key surfaced two defects the unit tests could not, both fixed here:

- **Bob's OTLP export never left the pod.** Bob's exporter builds its own `https.Agent`, which Node's `NODE_USE_ENV_PROXY` support does not cover, so the export went straight for the collector and died at the agent's egress policy (DNS is not even reachable there) — silently, since Bob's batch processor logs only in dev mode. The Bob image now starts the harness under `node --require /app/bob-proxy-agent.cjs`, a preload that hands every `http(s).Agent` the proxy env; verified against a local logging proxy (only Bob's own API calls arrived before, the telemetry host too after) and with a probe from inside the pod (HTTP 200 from the collector through the gateway). Same behaviour on Bob 2.0.2 and 2.0.3. The observability page said the opposite and now states the coupling.
- **The union query failed with `NO_COMMON_TYPE`.** The Claude Code branch reads latency as Int64 milliseconds while the Bob branch divided the span's UInt64 nanoseconds by 1e6 — a Float64, for which ClickHouse has no common supertype with Int64. Every Usage query failed the moment the first Bob span existed ("Couldn't load usage"). The Bob branch now yields integer milliseconds.

After both fixes the `LLM Generation`, `Agent Loop` and `Task Created` spans reached `otel_traces` stamped with `platform.agent.id`, and the Usage tab listed the Bob agent and its model. The call itself was refused by IBM (`BudgetExceededError`, the key is over its Bobcoin budget), so the span carried no `gen_ai.usage.cost` and the figure shown was 0 — a successful call is still needed to see a Bobcoin amount. One platform coupling came out of it and is now on the observability page: the rail is captured into the agent's env at creation and a template upgrade moves only the image, so an agent created before the backend was enabled never picks the rail up until it is recreated.


packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts
```diff
@@ -26,33 +27,87 @@ export function createClickhouseClient(cfg: {
...
+       SpanAttributes['langfuse.session.id'] AS sessionId,
+       coalesce(
+         nullIf(SpanAttributes['gen_ai.response.model'], ''),
+         SpanAttributes['gen_ai.request.model']) AS modelName,
+       '' AS requestIdRaw,
+       ${spanAttr("gen_ai.usage.input_tokens")} AS tokIn,
+       ${spanAttr("gen_ai.usage.output_tokens")} AS tokOut,
+       ${spanAttr("gen_ai.usage.cache_read.input_tokens")} AS tokCacheR,
+       ${spanAttr("gen_ai.usage.cache_creation.input_tokens")} AS tokCacheC,
+       ${spanAttr("gen_ai.client.operation.duration")} AS durMs,
```
dam-code-guardian: 🔴 **Critical:** Bob's per-call latency reads `gen_ai.client.operation.duration` through `toInt64OrZero` into the millisecond column `durMs`. The GenAI convention declares that attribute a **double in seconds**, so `"1.234"` is not an Int64 literal and `toInt64OrZero` gives **0** — every Bob call reports zero model time in `metrics show` API TIME, the per-model duration and the session runtime rollup. Even an integer-valued span would be off by 1000x.
  **Fix:** Parse every foreign counter with the type its convention declares and convert to the column's unit at read time — here `toFloat64OrZero(...) * 1000` — and pin the unit in a reader test, not only the SQL text.

packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts
```diff
@@ -63,14 +118,25 @@ const windowParams = (agentIds: readonly string[], w: MetricsWindow) => ({
...
-const TOK_CACHE_R = IN("'cache_read_tokens'");
-const TOK_CACHE_C = IN("'cache_creation_tokens'");
-const COST_USD = `${IN("'cost_usd_micros'")} / 1e6`;
+const CREDITS = "sumMap([creditUnit], [creditAmount]) AS credits";
 
 const n = (v: unknown): number => Number(v ?? 0);
 
+const credits = (v: unknown): CreditSpend[] => {
+  if (!Array.isArray(v)) return [];
+  const [units, amounts] = v as [unknown[], unknown[]];
```
dam-code-guardian: 🟡 **Warning:** The `credits` mapper decodes ClickHouse's `sumMap` result through an unchecked cast (`v as [unknown[], unknown[]]`) and returns `[]` for any shape it does not recognise, so a driver or query change reports every agent as having no credits instead of failing. No test exercises the mapper — the reader tests assert SQL text only.
  **Fix:** Decode driver payloads in infrastructure mappers with runtime-validated checks rather than a cast, and cover each mapper with a recorded-payload test so an unrecognised shape surfaces as an error.

packages/ui/src/modules/home/components/spend-widget.tsx
```diff
@@ -31,11 +36,14 @@ export function SpendWidget() {
   if (isPending) return <WidgetSkeleton rows={3} />;
 
   const total = data ? totalCostUsd(data.byModel) : 0;
+  const credits = data ? totalCredits(data.byModel) : [];
   const spenders = (data?.byAgent ?? [])
-    .filter((row) => row.costUsd >= ROUNDS_TO_A_VISIBLE_CENT_USD)
-    .sort((a, b) => b.costUsd - a.costUsd)
+    .filter(
+      (row) =>
+        row.costUsd >= ROUNDS_TO_A_VISIBLE_CENT_USD || row.credits.length > 0,
```
dam-code-guardian: 🟡 **Warning:** The filter admits credit-only agents but the top-3 rank is still dollars: the client-side `.sort` was deleted and the server orders `costUsd DESC, sum(creditAmount) DESC`, so every `costUsd: 0` row sits after every dollar row. With three dollar-spending agents a Bob agent burning any number of Bobcoins never reaches the widget and this widened filter is unreachable.
  **Fix:** Rank a mixed-unit list on a unit-aware key — rank within each unit and interleave, or take N per unit — never one unit's magnitude while the filter admits rows measured in another.

packages/ui/src/modules/metrics/lib/format.ts
```diff
@@ -30,6 +32,49 @@ export function formatAxisUsd(value: number, step: number): string {
...
+): string {
+  if (credits.length === 0) return usd(costUsd);
+  const creditText = formatCredits(credits);
+  return costUsd > 0 ? `${usd(costUsd)} + ${creditText}` : creditText;
+}
+
+export function spendBarPct(
+  rows: readonly { costUsd: number; credits: CreditSpend[] }[],
+): number[] {
+  const magnitude = (r: (typeof rows)[number]): [string, number] =>
```
dam-code-guardian: 🟡 **Warning:** `spendBarPct` takes a row's magnitude from `credits[0]` alone, so a row billed in two units is scaled by whichever unit the server emitted first and the rest is dropped; a row with both dollars and credits is scaled on dollars with its credits ignored. `totalCredits` shows multi-unit rows are an expected shape. The max map is also keyed on the literal `"usd"`, which a credit unit named `usd` would collide with.
  **Fix:** Derive the magnitude from an explicit selection over the whole `credits` list — one bar per unit, or a single-unit precondition enforced in the type — never positional index zero, and key unit maps on a value that cannot collide with a real unit.

packages/ui/src/modules/metrics/lib/format.ts
```diff
@@ -30,6 +32,49 @@ export function formatAxisUsd(value: number, step: number): string {
...
+
+export const creditUnitLabel = (unit: string): string =>
+  CREDIT_LABELS[unit] ?? unit;
+
+export const formatAxisCount = (value: number): string =>
+  compactNumber.format(value);
+
+export function formatCredits(credits: CreditSpend[]): string {
+  return credits
+    .map((c) => `${compactNumber.format(c.amount)} ${creditUnitLabel(c.unit)}`)
```
dam-code-guardian: 🟡 **Warning:** `formatCredits` runs credit amounts through `compactNumber` (`maximumFractionDigits: 1`) and is used for headline totals as well as bar labels, so the Usage page prints `+ 1.2K Bobcoins` beside an exact `formatUsdCents` dollar figure and the true credit total is not reachable anywhere in the UI. Also at `views/usage-view.tsx:100` and `home/components/spend-widget.tsx:78`.
  **Fix:** Keep compact notation for space-constrained labels only; a total a user reconciles against a bill gets an exact formatter, chosen per call site rather than baked into the one credit formatter.

packages/ui/src/modules/metrics/components/agent-spend-bars.tsx
```diff
@@ -1,20 +1,20 @@
 import type { SpendByAgent } from "api-server-api";
 
-import { formatUsd } from "../lib/format.js";
+import { formatSpend, spendBarPct } from "../lib/format.js";
 import { seriesColor } from "../lib/series-color.js";
 import { SpendBar } from "./spend-bar.js";
 
 export function AgentSpendBars({ rows }: { rows: SpendByAgent[] }) {
-  const max = rows[0]?.costUsd ?? 0;
+  const pcts = spendBarPct(rows);
```
dam-code-guardian, line 8: 🟡 **Warning:** `spendBarPct` now scales each bar against the top row **sharing its unit**, but `SpendBar`'s `aria-label` still says "N% of the largest", which is false across a mixed list. Only `ModelSpendBars` gained a visible caption; these bars, `session-type-spend-bars.tsx:9` and `home/components/spend-widget.tsx:96` disclose nothing.
  **Fix:** When a bar's denominator changes meaning, the label changes with it — pass the scale's unit into `SpendBar` so the `aria-label` and any caption name what the bar is a share of, for every consumer of the shared bar.

packages/ui/src/modules/metrics/components/session-type-spend-bars.tsx
```diff
@@ -1,21 +1,21 @@
...
 
-import { formatUsd } from "../lib/format.js";
+import { formatSpend, spendBarPct } from "../lib/format.js";
 import { seriesColor } from "../lib/series-color.js";
 import { SESSION_TYPE_LABELS } from "../lib/session-type-label.js";
 import { SpendBar } from "./spend-bar.js";
 
 export function SessionTypeSpendBars({ rows }: { rows: SpendBySessionType[] }) {
-  const max = rows[0]?.costUsd ?? 0;
+  const pcts = spendBarPct(rows);
```
dam-code-guardian, line 9: 🟡 **Warning:** Same defect as `agent-spend-bars.tsx:8` — bars now scale per unit while `SpendBar`'s `aria-label` still announces a share of the largest row in the whole list, and this list carries no caption to say otherwise.
  **Fix:** When a bar's denominator changes meaning, the label changes with it — pass the scale's unit into `SpendBar` so the `aria-label` and any caption name what the bar is a share of, for every consumer of the shared bar.

packages/ui/src/modules/metrics/views/usage-view.tsx
```diff
@@ -101,9 +114,29 @@ export function UsageView() {
               <section>
```
dam-code-guardian: 🟡 **Warning:** The dollar "Spend by day" chart renders whenever `byModel` is non-empty, so a deployment billed entirely in credits shows an all-zero dollar chart with `niceScale`'s `$0.25 … $1.00` fallback axis directly above the meaningful credits chart. Same construction at `components/sandbox-usage-section.tsx:105`.
  **Fix:** Gate each per-unit chart on that unit having spend; render the dollar series only when there are dollars or no other unit to show, so no surface presents an empty axis as data.

packages/ui/src/modules/metrics/components/sandbox-usage-section.tsx
```diff
@@ -91,14 +104,28 @@ export function SandboxUsageSection({ agentId }: { agentId: string }) {
                 <SectionLabel spaced>Spend by day</SectionLabel>
                 <Card className="p-5">
```
dam-code-guardian: 🟡 **Warning:** Same defect as `views/usage-view.tsx:114` — the dollar chart is unconditional, so an agent billed only in credits gets an empty `$` axis above its real credits chart.
  **Fix:** Gate each per-unit chart on that unit having spend; render the dollar series only when there are dollars or no other unit to show, so no surface presents an empty axis as data.

packages/ui/src/modules/sessions/components/session-row.tsx
```diff
@@ -168,10 +172,10 @@ export function SessionRow({
           {cost && (
             <span
               className="tabular-nums"
-              title={`${cost.calls} API calls · ${formatTokens(cost.inputTokens + cost.cacheReadTokens + cost.cacheCreationTokens)} in / ${formatTokens(cost.outputTokens)} out · $${cost.costUsd.toFixed(4)}`}
+              title={`${cost.calls} API calls · ${formatTokens(cost.inputTokens + cost.cacheReadTokens + cost.cacheCreationTokens)} in / ${formatTokens(cost.outputTokens)} out · ${formatSpend(cost.costUsd, cost.credits)}`}
```
dam-code-guardian: 🟡 **Warning:** This `title` existed to reveal precision — it printed `$${cost.costUsd.toFixed(4)}` beside the rounded cell. `formatSpend`'s default `formatUsd` gives two decimals, so tooltip and cell now show the same rounded number and the four-decimal figure is unreachable in the UI.
  **Fix:** A hover affordance that exists for precision keeps a formatter distinct from the cell's — pass an explicit high-precision `usd` argument to `formatSpend`.

deploy/helm/platform/templates/_helpers.tpl
```diff
@@ -443,11 +443,24 @@ API Server ServiceAccount name
...
-     OTel service name is the CLI's own ("claude-code" for every claude-code-
-     based image), which makes derived templates like nous indistinguishable
-     in the exploration UI — so name the service after the template. */}}
+{{/* Call with (dict "root" $ "templateName" <name> "rail" $tmpl.telemetry).
+     Each harness reads its own export env, so `telemetry` names the rail:
+     `true` (or "claude-code") for the Claude Code env, "bob" for Bob Shell's.
+     Both land on the same collector over the agent's ordinary gateway egress —
+     see docs/architecture/observability.md. */}}
 {{- define "platform.agentTelemetry.env" -}}
+{{- if eq (toString .rail) "bob" }}
```
dam-code-guardian: 🟡 **Warning:** Only the literal `"bob"` selects the Bob rail; every other truthy `telemetry` value falls through to Claude Code, and no values schema rejects the rest. A template written `telemetry: bobb` or `claude_code` gets the wrong rail — for a Bob image that means no `BOB_TELEMETRY_*` at all and spend that never reaches Usage, with no render-time signal.
  **Fix:** Validate a variant-naming value against its closed set at render time and `fail` on an unknown name; never let an unrecognized name degrade into a default rail.

docs/architecture/metrics.md
```diff
@@ -64,7 +64,13 @@ Which side of that union an agent came from is not discarded once the set is for
...
..., only to label a row, and only as the **fallback for deleted agents** (live agents are labelled from the platform registry in the service layer) — taking the latest value seen in the window among the agent's **own** rows, excluding child rows carrying an **invocation identifier** (the target-own id the gateway stamps alongside the Driver's attribution), so a Driver keeps its own name rather than inheriting a target's throwaway one; a deleted agent still keeps a readable name, and an agent with no own-name row in the window is left unlabelled for the client to fall back to its id. The per-day breakdown reuses the same **timestamp**, shifting each row into the client-supplied timezone before truncating to a calendar day so buckets line up with the user's wall clock rather than UTC. This is a coupling to the harness's export contract, not to platform code; the exact attribute keys and query text are the reader's own detail in [`packages/api-server/`](../../packages/api-server/). The reader scopes to Claude Code's per-call records by the record body alone — not by service name, which carries the agent's *template* name and would hide every template not literally named for the harness.
```
dam-code-guardian: 🟡 **Warning:** This paragraph still ends with the superseded sentence "The reader scopes to Claude Code's per-call records by the record body alone", which contradicts its own new opening (both shapes are selected by body **or span name**) and the code. The shipped page states the read path is Claude-Code-only — the thing this PR exists to change. The "never by service name" clause is duplicated within the paragraph too.
  **Fix:** When a documented scope is widened, delete the narrow statement rather than prepend the wide one — a page must not carry two statements of the same fact.

docs/architecture/observability.md
```diff
@@ -43,12 +43,13 @@ Making the mesh the sole gate is why the collector is platform-owned. ClickStack
...
 
-- **Rides the agent's ordinary egress.** The exporter honours the agent's `HTTPS_PROXY`, so telemetry leaves through the paired gateway pod over HTTPS to the bundled collector — the same dedicated, MITM-terminating chain that performs the trusted attribution below. No new network path, and no credential is injected into the export.
+- **Rides the agent's ordinary egress.** The exporter honours the agent's `HTTPS_PROXY`, so telemetry leaves through the paired gateway pod over HTTPS to the bundled collector — the same dedicated, MITM-terminating chain that performs the trusted attribution below. No new network path, and no credential is injected into the export. This is a **precondition, not a convenience**: the collector is reachable no other way, and an exporter that ignored the proxy would also miss the attribution stamp, so its records could never be attributed to a user. Honouring it is not automatic — an exporter built on a runtime's raw HTTP client obeys the proxy only because the agent env asks that runtime to, which is what makes Bob's OTLP export work unchanged.
```
dam-code-guardian: 🟡 **Warning:** The Bob rail injects a Langfuse key pair that the harness validates and, on a parse failure, falls back to Bob's own external IBM endpoint. That egress consequence lives only in a Helm comment, while this page still claims "the collector is reachable no other way" and "no credential is injected into the export".
  **Fix:** State on the observability page any rail value whose absence or malformation changes where telemetry goes; an invariant recorded only in a template comment is not documented.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `d97e8a6`

### Summary

The union read path is well built: one normalized record, both shapes selected by body/span name, credits carried per unit and never summed across units. The defects are at the new drain points — Bob's latency attribute is parsed as an integer where the convention gives fractional seconds, and the unit-aware presentation stops short of the top-N rank, the bar labels and the empty dollar chart.

_10 suggestion(s) dropped under the 🟢 budget._

### Findings

- 🔴 **Critical:** Bob's per-call latency reads `gen_ai.client.operation.duration` through `toInt64OrZero` into the millisecond column `durMs`. The GenAI convention declares that attribute a **double in seconds**, so a value like `"1.234"` is not an Int64 literal and `toInt64OrZero` gives **0**: every Bob call reports zero model time in `metrics show` API TIME, in the per-model duration and in the session runtime rollup, with no signal that the number is wrong. Even an integer-valued span would be off by 1000×. Reported by typescript-engineering. (`packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts:86`)
  **Fix:** Parse every foreign counter with the type its convention declares and convert to the column's unit at read time — here `toFloat64OrZero(...) * 1000` — and pin the unit in a reader test, not only the SQL text.
- 🟡 **Warning:** The Home widget admits credit-only agents (`row.credits.length > 0`) but takes the server's `ORDER BY costUsd DESC, sum(creditAmount) DESC` and its own `.sort` was deleted, so every `costUsd: 0` row sits after every dollar row. With three dollar-spending agents a Bob agent burning any number of Bobcoins never reaches the widget and the widened filter is unreachable. Reported by react-ui-engineering and issue-fit. (`packages/ui/src/modules/home/components/spend-widget.tsx:40`)
  **Fix:** Rank a mixed-unit list on a unit-aware key — rank within each unit and interleave, or take N per unit — never one unit's magnitude while the filter admits rows measured in another.
- 🟡 **Warning:** `spendBarPct` takes a row's magnitude from `credits[0]` alone, so a row billed in two units is scaled by whichever unit the server emitted first and the rest is dropped; a row with both dollars and credits is scaled on dollars with its credits ignored. `totalCredits` shows multi-unit rows are an expected shape. Reported by typescript-engineering, react-ui-engineering and issue-fit. (`packages/ui/src/modules/metrics/lib/format.ts:62`)
  **Fix:** Derive the magnitude from an explicit selection over the whole `credits` list — one bar per unit, or a single-unit precondition enforced in the type — never positional index zero.
- 🟡 **Warning:** `spendBarPct` now scales each bar against the top row **sharing its unit**, but `SpendBar`'s `aria-label` still says "`N`% of the largest", which is false across a mixed list. Only `ModelSpendBars` gained a visible caption; the agent bars, the session-type bars and the Home widget disclose nothing. Reported by react-ui-engineering. (`packages/ui/src/modules/metrics/components/agent-spend-bars.tsx:8`, also `packages/ui/src/modules/metrics/components/session-type-spend-bars.tsx:9`, `packages/ui/src/modules/home/components/spend-widget.tsx:96`)
  **Fix:** When a bar's denominator changes meaning, the label changes with it — pass the scale's unit into `SpendBar` so the `aria-label` and any caption name what the bar is a share of, for every consumer of the shared bar.
- 🟡 **Warning:** The dollar "Spend by day" chart renders whenever `byModel` is non-empty, so a deployment billed entirely in credits shows an all-zero dollar chart with `niceScale`'s `$0.25 … $1.00` fallback axis directly above the meaningful credits chart. Same construction on both surfaces. Reported by react-ui-engineering. (`packages/ui/src/modules/metrics/views/usage-view.tsx:114`, also `packages/ui/src/modules/metrics/components/sandbox-usage-section.tsx:105`)
  **Fix:** Gate each per-unit chart on that unit having spend; render the dollar series only when there are dollars or no other unit to show, so no surface presents an empty axis as data.
- 🟡 **Warning:** `formatCredits` runs credit amounts through `compactNumber` (`maximumFractionDigits: 1`) and is used for headline totals as well as bar labels, so the Usage page prints `+ 1.2K Bobcoins` beside an exact `formatUsdCents` dollar figure and the true credit total is not reachable anywhere in the UI. Reported by react-ui-engineering and typescript-engineering. (`packages/ui/src/modules/metrics/lib/format.ts:45`, also `packages/ui/src/modules/metrics/views/usage-view.tsx:100`, `packages/ui/src/modules/home/components/spend-widget.tsx:78`)
  **Fix:** Keep compact notation for space-constrained labels only; a total a user reconciles against a bill gets an exact formatter, chosen per call site rather than baked into the one credit formatter.
- 🟡 **Warning:** The session row's `title` existed to reveal precision — it printed `$${cost.costUsd.toFixed(4)}` beside the rounded cell. It now calls `formatSpend(cost.costUsd, cost.credits)`, whose default `formatUsd` gives two decimals, so tooltip and cell show the same rounded number. Reported by react-ui-engineering. (`packages/ui/src/modules/sessions/components/session-row.tsx:175`)
  **Fix:** A hover affordance that exists for precision keeps a formatter distinct from the cell's — pass an explicit high-precision `usd` argument to `formatSpend`.
- 🟡 **Warning:** `platform.agentTelemetry.env` special-cases the literal `"bob"` and routes every other truthy `telemetry` value to the Claude Code rail, with no schema to reject the rest. A template written `telemetry: bobb` or `claude_code` silently gets the wrong rail — for a Bob image that means no `BOB_TELEMETRY_*` at all and spend that never reaches Usage, with no render-time signal. Reported by issue-fit. (`deploy/helm/platform/templates/_helpers.tpl:452`)
  **Fix:** Validate a variant-naming value against its closed set at render time and `fail` on an unknown name; never let an unrecognized name degrade into a default rail.
- 🟡 **Warning:** The `credits` mapper decodes ClickHouse's `sumMap` result through an unchecked cast (`v as [unknown[], unknown[]]`) and returns `[]` for any shape it does not recognise, so a driver or query change reports every agent as having no credits instead of failing; no test exercises the mapper — the reader tests assert SQL text only. Reported by typescript-engineering. (`packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts:127`)
  **Fix:** Decode driver payloads in infrastructure mappers with runtime-validated checks rather than a cast, and cover each mapper with a recorded-payload test so an unrecognised shape surfaces as an error.
- 🟡 **Warning:** `docs/architecture/metrics.md` keeps the superseded sentence "The reader scopes to Claude Code's per-call records by the record body alone" at the end of the paragraph whose new opening says both shapes are selected by body **or span name**, so the shipped page states the read path is Claude-Code-only — the thing this PR exists to change. Reported by doc-drift and issue-fit. (`docs/architecture/metrics.md:73`)
  **Fix:** When a documented scope is widened, delete the narrow statement rather than prepend the wide one — a page must not carry two statements of the same fact.
- 🟡 **Warning:** The Bob rail injects a Langfuse key pair that the harness validates and, on a parse failure, falls back to Bob's own external IBM endpoint. That egress consequence lives only in a Helm comment, while `docs/architecture/observability.md` still claims "the collector is reachable no other way" and "no credential is injected into the export". Reported by doc-drift. (`docs/architecture/observability.md:48`)
  **Fix:** State on the observability page any rail value whose absence or malformation changes where telemetry goes; an invariant recorded only in a template comment is not documented.
- 🟢 **Suggestion:** The model and agent rollups order by `costUsd DESC, sum(creditAmount) DESC`, which adds amounts across credit units in the ordering key — against the read path's own rule that nothing is summed across units. `spendBySessionType` has the same dollar-first bias in its `.sort`. (`packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts:171`)

✅ Looks good: the added Helm and YAML comments are outside the no-comments rule's scope (`check-comment-types.mjs` routes TS and Go only), and the `AND agentNameRaw != ''` guard on the name fallback is correct for Bob rows that carry no name.

- 🟡 **Warning:** The telemetry-reader paragraph still ends with the pre-change sentence "The reader scopes to Claude Code's per-call records by the record body alone — not by service name…", which now flatly contradicts the new opening of the same section (line 67) and the code: `clickhouse-reader.ts` unions Claude Code log records (`Body = 'claude_code.api_request'` over `otel_logs`) with Bob Shell spans (`SpanName = 'LLM Generation'` over `otel_traces`). A reader of the architecture page — the declared agent-facing source of truth — gets two mutually exclusive statements about what the reader selects, and the "never by service name" clause is duplicated verbatim within one paragraph (`docs/architecture/metrics.md:73`)
  **Fix:** When a page section is rewritten to widen a stated scope, delete the superseded sentence rather than appending the new one — a page must never carry two statements of the same fact, and the old one is the drifted one by construction.

- 🟡 **Warning:** The Agent export section documents the new Bob rail's signals, service-name gap and payload-exclusion override, but not the two load-bearing properties of the rail the helper itself records: the Langfuse key pair is required and validated even though the collector ignores it, and a failed parse silently diverts Bob's export to Bob's own external IBM endpoint. That is exactly the operational coupling/invariant the page is for, and it sits against the page's own claims that "the collector is reachable no other way" and "no credential is injected into the export" — the rail now injects two credential-shaped values (`deploy/helm/platform/templates/_helpers.tpl:498`, undocumented at `docs/architecture/observability.md:48`)
  **Fix:** When a new export rail's env carries a value whose absence or malformation changes where telemetry goes, state that failure mode and its egress consequence on the observability page — an invariant recorded only in a Helm comment is not documented, since the architecture page is the agent-facing source of truth.

- 🟢 _1 dropped under the 🟢 budget: the Contract's "sorted highest cost first" line no longer determines order once units are mixed._

### Issue Fit

- 🟡 **Warning:** No linked issue — the PR body's only reference, `#3553`, is the stacked base PR, not an issue, and the head branch `feat/bob-usage` carries no issue number, so scope was judged against the PR description alone. The body's Home-surface promise ("Usage, Home and the CLI show credits beside dollars … A credit-billed agent therefore reads as its own total rather than as `$0`") is only partly delivered: the widget's own `.sort((a, b) => b.costUsd - a.costUsd)` was deleted in favour of the server order, and the server orders `ORDER BY costUsd DESC, sum(creditAmount) DESC`, which puts every credit-only row after every dollar row — so a Bob agent is cut by `.slice(0, TOP_SPENDERS)` whenever three dollar-spending agents exist and never appears on Home at all (`packages/ui/src/modules/home/components/spend-widget.tsx:45`).
  **Fix:** When a rollup carries values in more than one unit, ranking and top-N truncation must be unit-aware — rank within each unit or interleave — never a single dollar-keyed order that makes non-dollar rows unreachable.

- 🟡 **Warning:** The architecture page the PR rewrote still ends with the pre-change sentence "The reader scopes to Claude Code's per-call records by the record body alone", directly contradicting the new paragraph two sentences above it that says both shapes are selected "by the record body or span name alone". The shipped doc therefore states the read path is Claude-Code-only, which is what this PR exists to stop being true (`docs/architecture/metrics.md:73`).
  **Fix:** When widening a documented invariant, delete the narrow statement of it rather than prepending the wide one — an architecture page must not contain both.

- 🟡 **Warning:** The body declares `telemetry` as an enumerated rail name (`true` / `claude-code`, or `bob`), but the template only special-cases the literal `"bob"` and routes every other truthy value to the Claude Code rail, and there is no values schema to reject the rest. A template written `telemetry: claude_code` or `telemetry: bobb` silently gets the wrong rail — for a Bob image that means no `BOB_TELEMETRY_*` at all and spend that never reaches Usage, with no render-time signal (`deploy/helm/platform/templates/_helpers.tpl:452`).
  **Fix:** A value that names a variant must be validated against the closed set of names at render time and fail the render on an unknown one; never let an unrecognized name degrade into a default rail.

- 🟢 _1 dropped under the 🟢 budget: the `AND agentNameRaw != ''` guard changes the Claude Code rail's own name fallback, undeclared in the body. Its other two suggestions are carried above._

- 🟡 **Warning:** The metrics infrastructure adapter takes its port types (`MetricsReader`, `MetricsWindow`, `SessionSpend`) from the services layer, and this PR extends that inverted dependency by adding `credits: CreditSpend[]` to `SessionSpend` in `services/metrics-service.ts`. The layer rule is `services → domain ← infrastructure`: infrastructure may import from domain but never from services, and repository/reader port interfaces belong in `infrastructure/` (`architecture/slice-composition.md`, `architecture/infrastructure.md`) (`packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts:10`).
  **Fix:** Declare every port a module's adapters implement in `infrastructure/` (or the domain), and have `services/` import the port — never the reverse; when a port grows a field, grow it on the infrastructure-side declaration.
- 🟡 **Warning:** The new `credits` mapper decodes ClickHouse's `sumMap` result through an unchecked cast (`v as [unknown[], unknown[]]`) and returns `[]` on any shape it does not recognise, so a driver or query change silently reports every agent as having no credits rather than failing; no test exercises the mapper at all — the reader tests only assert SQL text (`packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts:127`).
  **Fix:** Parse driver result shapes in infrastructure mappers with a runtime-validated decode (Zod or explicit element checks) instead of a cast, and cover each mapper with a test that feeds it a recorded driver payload, so an unrecognised shape surfaces as an error rather than as zero data.
- 🟡 **Warning:** Bob's per-call latency is read with `toInt64OrZero(SpanAttributes['gen_ai.client.operation.duration'])` into a millisecond column, while the sibling cost attribute on the same span is read with `toFloat64OrZero`. The GenAI semantic-convention name denotes a fractional seconds value, and `toInt64OrZero` yields 0 for any non-integer string, so Bob sessions can report ~0 model time with no signal — and nothing in the tests pins the unit (`packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts:86`).
  **Fix:** Read every foreign telemetry counter with a parse matching its declared type and unit (float parse plus an explicit unit conversion where the source unit differs from the column's), and assert the conversion in a reader test rather than only the SQL text.
- 🟡 **Warning:** `spendBarPct` derives a row's magnitude from `credits[0]` alone, so a row carrying more than one credit unit is scaled against whichever unit the server happened to emit first, and a row with both a dollar cost and credits is scaled on dollars with its credits ignored — the shared bar helper therefore mis-scales exactly the mixed case the contract's `CreditSpend[]` was introduced to support (`packages/ui/src/modules/metrics/lib/format.ts:62`).
  **Fix:** Make presentation helpers that consume a per-unit list handle the whole list explicitly — either group and render one bar per unit, or state and enforce a single-unit precondition in the type — never index position 0 as a stand-in for "the unit".
- 🟢 _3 dropped under the 🟢 budget: the domain's contract coupling, the CLI's re-implemented credit formatting, and `calls_all`'s repeated inlining. Its `ORDER BY`, session-type-sort and `formatCredits` suggestions are carried above._

- 🟡 **Warning:** The widget's filter was widened to admit credit-only agents (`row.credits.length > 0`) but the top-3 selection is still ranked purely in dollars — the client-side `.sort((a, b) => b.costUsd - a.costUsd)` was deleted and the order now comes from the server's `ORDER BY costUsd DESC, sum(creditAmount) DESC`, which puts every `costUsd: 0` credit-billed agent after every dollar-spender. With three or more dollar-spending agents a Bob agent burning credits can never appear in the widget, so the widened filter is dead code. (`packages/ui/src/modules/home/components/spend-widget.tsx:40`)
  **Fix:** When a list mixes billing units, the top-N selection must rank on a unit-aware key (e.g. rank within each unit and interleave, or take N per unit) — never on one unit's magnitude while the filter admits rows measured in another.
- 🟡 **Warning:** The USD "Spend by day" chart is rendered unconditionally whenever `byModel` is non-empty, so a deployment whose spend is entirely credits shows an all-zero dollar chart (axis `$0.25 … $1.00` from `niceScale`'s empty-data fallback) directly above the meaningful credits chart. Same construction in both surfaces. (`packages/ui/src/modules/metrics/views/usage-view.tsx:114`, `packages/ui/src/modules/metrics/components/sandbox-usage-section.tsx:103`)
  **Fix:** Gate every per-unit chart on that unit actually having spend; render the dollar series only when `total > 0` or when there is no other unit to show, so no surface ever presents an empty axis as data.
- 🟡 **Warning:** `spendBarPct` now scales each bar against the largest row *sharing its unit*, so a 200-Bobcoin row and a $10 row both render at 100% width, but `SpendBar`'s `aria-label` still announces "`N`% of the largest" — a comparison that is false across the list for screen-reader users. Only `ModelSpendBars` gained a visible caption saying the scale is per-unit; the agent list, the session-type list and the home widget's bars disclose nothing. (`packages/ui/src/modules/metrics/components/agent-spend-bars.tsx:16`, `packages/ui/src/modules/metrics/components/session-type-spend-bars.tsx:17`, `packages/ui/src/modules/home/components/spend-widget.tsx:97`)
  **Fix:** When a bar's denominator changes meaning, the label must change with it — pass the scale's unit into `SpendBar` so both the `aria-label` and any caption name what the bar is a share *of*, and apply that to every consumer of the shared bar, not one.
- 🟡 **Warning:** The session row's tooltip existed to show the exact cost — it printed `$${cost.costUsd.toFixed(4)}` while the visible cell showed the rounded `formatUsdCell` value. It now calls `formatSpend(cost.costUsd, cost.credits)`, whose default `formatUsd` gives two decimals, so the tooltip and the cell display the same rounded number and the four-decimal figure is no longer reachable anywhere in the UI. (`packages/ui/src/modules/sessions/components/session-row.tsx:175`)
  **Fix:** A hover/title affordance that exists to reveal precision must keep a formatter distinct from the cell's — pass an explicit high-precision formatter to `formatSpend` rather than letting it fall back to the display default.
- 🟡 **Warning:** `formatCredits` formats credit amounts with `compactNumber` (`maximumFractionDigits: 1`), and it is used for the authoritative totals as well as for bar labels — the Usage page headline renders `+ 1.2K Bobcoins` next to an exact `formatUsdCents` dollar total, so the two halves of the same billing figure have different precision and the real credit total is not shown anywhere. (`packages/ui/src/modules/metrics/lib/format.ts:43`, `packages/ui/src/modules/metrics/views/usage-view.tsx:101`, `packages/ui/src/modules/home/components/spend-widget.tsx:78`)
  **Fix:** Keep compact notation for space-constrained labels only; totals and any figure a user might reconcile against a bill get an exact formatter, so precision is chosen per call site rather than baked into the one credit formatter.
- 🟡 **Warning:** `spendBarPct`'s `magnitude` reads only `r.credits[0]`, so a row billed in two credit units is silently scaled by whichever unit the server happened to emit first and the rest is dropped; the same function keys its max map on the literal string `"usd"`, which collides with — and would be bucketed together with dollars — a credit unit actually named `usd`. `totalCredits` proves multi-unit rows are an expected shape. (`packages/ui/src/modules/metrics/lib/format.ts:62`)
  **Fix:** Derive a bar's magnitude from an explicit, exhaustive unit selection rather than positional index zero, and key unit maps on a value that cannot collide with a real unit (a tagged discriminant, not a bare string literal).
- 🟢 _5 dropped under the 🟢 budget: `formatCredits`'s always-plural label, the duplicated credits-by-day block, the per-render closure in `formatValue`, `daySeries`'s home in `totals.ts`, and `niceScale`'s dollar-tuned empty fallback._

### Verdict

**REQUEST_CHANGES** — one 🔴 (Bob latency parsed as an integer, so every Bob call reports zero model time) and ten 🟡, of which the Home top-N rank makes the PR's own Home-surface promise unreachable and the architecture page ships two contradictory statements of what the reader selects.


---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `cbfb582`

## PR #3562: feat(metrics): report Bob Shell spend on the Usage surfaces
**Author:** xjacka | **Branch:** feat/bob-usage → feat/bob-acp-native | **Changes:** +833 −278 (31 files)

### Summary

The fix commit closes every blocking finding of the previous round at its cause: latency now comes from the span's own exporter-recorded `Duration`, the `sumMap` decoder validates and raises, Home ranks per unit and interleaves, the bar scale reads the largest credit and keys on a symbol, the scale unit reaches every `SpendBar` label, the totals and the tooltip use exact formatters, the per-unit day charts are gated, and an unknown rail name fails the render. Only suggestions remain.

_17 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: d97e8a6 (2026-09-07T10:55:55Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** Bob latency now reads `Duration / 1e6`, pinned by a reader test (`packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts:86`)
- ✅ **Fixed:** Home ranks per unit and interleaves through `topPerUnit` (`packages/ui/src/modules/home/components/spend-widget.tsx:46`)
- ✅ **Fixed:** `spendScale` takes the largest credit, not `credits[0]` (`packages/ui/src/modules/metrics/lib/format.ts:82`)
- ✅ **Fixed:** every `SpendBar` consumer passes `scaleLabel`, so the label names the unit (`packages/ui/src/modules/metrics/components/agent-spend-bars.tsx:18`)
- ✅ **Fixed:** the dollar day chart is gated on dollars or no other unit (`packages/ui/src/modules/metrics/views/usage-view.tsx:115`)
- ✅ **Fixed:** headline credit totals use `formatCreditsExact` (`packages/ui/src/modules/metrics/views/usage-view.tsx:102`)
- ✅ **Fixed:** the session tooltip passes a four-decimal `usd` formatter (`packages/ui/src/modules/sessions/components/session-row.tsx:176`)
- ✅ **Fixed:** an unknown `telemetry` rail name fails the render (`deploy/helm/platform/templates/_helpers.tpl:451`)
- ✅ **Fixed:** the `sumMap` decoder validates and raises, with a mapper test (`packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts:135`)
- ✅ **Fixed:** the superseded Claude-Code-only sentence is deleted (`docs/architecture/metrics.md:67`)
- ✅ **Fixed:** the rail's egress fallback is stated on the observability page (`docs/architecture/observability.md:49`)
- 🔁 **Still present:** `ORDER BY costUsd DESC, sum(creditAmount) DESC` sums amounts across credit units (`packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts:197`)
- 🆕 **New:** two mappers decode one credit column with opposite strictness (`packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts:161`)
- 🆕 **New:** both edited pages keep `Last verified: 2026-09-04`, and the new render-time gate is unstated (`docs/architecture/metrics.md:3`)

### Findings

- 🟢 **Suggestion:** The same credit column now has two decoders with opposite strictness: `credits` raises on a bad unit or a non-finite amount, while `one` — used by `contextPerCall` — coerces through `n()` and reports no credits. The same corrupt row is a hard error on one read path and "no credits" on the other. Reported by typescript-engineering. (`packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts:161`, also `packages/api-server/src/modules/metrics/infrastructure/clickhouse-reader.ts:135`)
- 🟢 **Suggestion:** Both architecture pages this commit edits still read `Last verified: 2026-09-04`, and the observability page's rail bullet states the runtime redirect without the render-time `fail` gate that the same commit added — the gate is now the durable invariant. Reported by doc-drift. (`docs/architecture/metrics.md:3`, also `docs/architecture/observability.md:3`)

- 🟢 _Its two `Last verified` items and the unstated `fail` gate are carried as one finding above. 5 dropped under the 🟢 budget: `cli.md`'s missing metrics group, the "sorted highest cost first" ordering line, the unstated raise-on-bad-payload rule, the attribute-level latency detail, and a stray blank line in the Agent export list._

### Issue Fit

- 🟢 _Its ordering item is the carryover above (`ORDER BY` sums across units). 3 dropped under the 🟢 budget: no linked issue, the `agentNameRaw != ''` guard's undeclared effect on Claude Code rows, and no chart test for either rail or the new `fail` branch._

- 🟢 _Its two-mapper item is carried above. 4 dropped under the 🟢 budget: `CreditPayloadError` escaping untranslated to the transport, the optional-with-default formatter parameters, `SpendRow` hand-declared in the UI, and the credit-unit vocabulary hardcoded on both sides._

- 🟢 _5 dropped under the 🟢 budget: the widget's hand-rolled bar markup and label, the per-unit scaling caption shipped on one list only, `formatSpendExact` exported and never called, the duplicated per-unit day-chart block, and `SpendBar`'s now-unreachable `scaleLabel` fallback._

### Verdict

**APPROVE** — the one 🔴 and all ten 🟡 of the previous round are fixed at their cause, and nothing blocking remains at this HEAD; the residual items are suggestions.


---

xjacka:
We will address this later once Bob shell support is available.

PR #3565, opened 2026-09-04
feat(gateway): rewrite egress path prefixes so Bob can run on the LiteLLM proxy

xjacka:
Bob's harness provider is already an OpenAI-compatible client — it POSTs an OpenAI body to `/chat/completions` and reads `choices[0]` and `usage` — and its base URL is overridable via `--gateway-url`, `BOB_GATEWAY_URL` or `gatewayUrl` in settings. What stops it reaching a plain OpenAI-shaped endpoint is that its **service prefixes are constants**: 2.0.2 hard-codes `/inference/v1` for inference and `/admin/v1` for admin, and joins them onto whatever base URL you configure. Its own backend is LiteLLM; so is our IBM LiteLLM proxy, which serves the same routes without the prefix.

So the endpoint choice belongs where the credential already lives, and the prefix belongs to the gateway that already rewrites the wire.

## Path rewriting in the egress sidecar

An `egress-inject` Contribution may declare path prefix rewrites. The host's L7 chain renders them as routes **ahead of its catch-all**, each with `prefix_rewrite`, so:

- paths outside the declared prefixes go upstream untouched — Claude Code on `/v1/messages` and Bob on `/inference/v1/*` share one chain on the same host;
- rewriting is a routing-leg concern, **after** the ext_authz Check, so egress rules and approvals keep describing the paths the agent actually requests, not the rewritten ones;
- both ends of a rewrite must be whole path segments. The controller re-validates what it reads back from the Secret annotation and drops anything else with a warning — a rewrite must not be able to reach past what the SNI-matched host admits.

Nothing changes for hosts that declare no rewrites.

## Which endpoint Bob runs against follows from the granted Connection

The IBM LiteLLM Connection now contributes `BOB_GATEWAY_URL`, a placeholder `BOBSHELL_API_KEY`, the `/inference/v1/` → `/v1/` rewrite, and a model the proxy serves. An agent granted that Connection runs Bob against the proxy; an agent without it keeps Bob's built-in gateway, untouched. No harness config, no hardcoded host.

The auth scheme needs no work: Bob sends `Authorization: Apikey …`, and the credential injector runs with `overwrite: true`, so the header is replaced with the proxy's `Bearer` regardless.

The model is load-bearing. Bob's built-in default resolves to a tier alias (`premium-ide`) that exists only on its own gateway, so a model has to be named on the Connection — a config input, deliberately **without** a shipped default: the Bob Connection contributes the same `BOB_SHELL_MODEL` from its own pin, and two always-firing contributors would leave the winner to grant order. The connect form exposes the field on create. It is create-only, because the edit dialog carries the credential and nothing else; changing it means recreating the connection or overriding `BOB_SHELL_MODEL` per agent. That, the doubled-model case, and the fact that a connection predating this keeps its old contributions are all named in the agent README.

## What the endpoint has to satisfy

Verified against Bob 2.0.2's bundle and against the proxy itself:

- `POST /v1/chat/completions` — OpenAI-shaped, streaming.
- `GET /v1/model/info` — LiteLLM-shaped `data[].model_name` plus `model_info` (context window, max tokens, vision, prompt caching, per-token costs), and **non-empty**: Bob treats an empty model list as an error.
- `/admin/v1/profile` may 404, and this PR deliberately does not rewrite it. Bob tolerates the 404, and a gateway that answered it with a region hint would make Bob redirect its inference calls off the configured host (its inference client resolves the active profile's region domain over the configured base URL).

The three requests above are the complete set a headless run makes.

## Verification

- Rendered bootstrap: the rewrite route carries `prefix_rewrite: /v1/` and precedes the catch-all, both pinned to the host's STRICT_DNS cluster.
- Against the real proxy: `/inference/v1/*` is 404 while `/v1/model/info` is 401 and `/v1/chat/completions` is 405-on-GET — the prefix is the only structural difference on the inference path. With a stand-in proxy applying this rewrite, Bob 2.0.2 headless reached `POST /v1/chat/completions` and was refused on auth alone; the 404 on the admin profile route did not block it.
- Tests: two in the controller (route order, malformed rewrite dropped), three for the Connection (gateway env, model override, the Secret annotation).

**Not verified: a full run through to a model reply.** That needs a valid LiteLLM key, which this change was written without, so the evidence stops at "the request arrives and is refused on auth alone". Whoever has a key can close that gap in a minute; nothing in the diff depends on the answer, but the claim that Bob *answers* through the rewrite is untested and should be read that way.

## Notes for review

- An agent granted *both* the LiteLLM and the Bob Connection runs Bob against the proxy: only LiteLLM contributes `BOB_GATEWAY_URL`. The Bob Connection then injects a credential on a host Bob no longer calls.
- The pinned default model was taken from the proxy model this repo already names elsewhere in the same env mapping; it has not been checked against the proxy's live catalogue.
- `docs/architecture/security-and-credentials.md` was at its size cap, so fitting the new mechanics meant cutting code-level detail the guidelines already exclude: a transcribed annotation shape, two Envoy field names, and literal header values in the GitHub example. One stale reference went with it — the multi-injection section cited Bob's `/key/info?key=…` endpoint, which 2.0.2 no longer has.

---

## Review round 1

Every finding from Code Guardian is addressed in `3e26981b`; the reasoning for the two I handled differently than suggested is in [the review reply](#issuecomment).


packages/api-server-api/src/modules/connections/providers.ts
```diff
@@ -29,6 +36,8 @@ export function ibmLitellmEnvMappings(): EnvMapping[] {
     { envName: "OPENAI_API_KEY", placeholder: DEFAULT_ENV_PLACEHOLDER },
     { envName: "OPENAI_BASE_URL", placeholder: IBM_LITELLM_BASE_URL },
     { envName: "OPENAI_MODEL", placeholder: "gpt-5.5" },
+    { envName: "BOB_GATEWAY_URL", placeholder: IBM_LITELLM_BASE_URL },
+    { envName: "BOBSHELL_API_KEY", placeholder: "sk-dummy" },
```
dam-code-guardian: 🔴 **Critical:** This seeds `BOBSHELL_API_KEY` with `sk-dummy`, and Bob's own README states the placeholder "must not start with `sk-`/`pk-` or Bob's bundle would silently downgrade to the legacy `prod.ibm-bob-staging.cloud.ibm.com` backend (which only accepts JWT keys)". Every agent granted this connection therefore bypasses the LiteLLM proxy this PR exists to route it through — silently, since Envoy still rewrites the wire value and nothing errors.

  **Fix:** A placeholder is inert filler and must never carry a prefix the consuming tool dispatches on — seed it from `DEFAULT_ENV_PLACEHOLDER` (as `bobEnvMappings()` already does for this same variable) rather than a hand-written literal.

```suggestion
    { envName: "BOBSHELL_API_KEY", placeholder: DEFAULT_ENV_PLACEHOLDER },
```

packages/api-server/src/modules/connections/domain/catalog.ts
```diff
@@ -157,6 +159,16 @@ const IBM_LITELLM: HeaderConnectionTemplate = {
       host: IBM_LITELLM_HOST,
       headerName: "Authorization",
       valueFormat: "Bearer {value}",
+      pathRewrites: [BOB_INFERENCE_PREFIX_REWRITE],
+    },
+  ],
+  configInputs: [
+    {
+      inputName: "bobModel",
+      envName: "BOB_SHELL_MODEL",
```
dam-code-guardian: 🟡 **Warning:** This contributes `BOB_SHELL_MODEL` with a `defaultValue` that always fires, and the `bob` template already contributes the same env name from its own `model` config input. An owner who grants both connections gets whichever value belongs to the connection created first: `state-builder.ts` concatenates grants in `createdAt` order, `env-plugin.ts` keeps the first `name`, and the controller's `credentialEnvVars` keeps the first `envName`. The silent loser is exactly what the new hint warns about — a Bob-native model name reaching the proxy, or the proxy model reaching Bob's own gateway.

  **Fix:** Two templates must never contribute the same env name from independent inputs; when a variable is proxy-scoped, derive it per host at delivery time or make the collision explicit (reject or flag the second grant) instead of relying on grant order to pick a winner.

packages/agent-runtime-api/src/modules/runtime/types.ts
```diff
@@ -46,10 +46,18 @@ export const egressAllowContribution = z.object({
   pathPattern: z.string().optional(),
 });
 
+const anchoredPathSegments = /^\/(?:[A-Za-z0-9._~-]+\/)*$/;
```
dam-code-guardian, line 49: 🟡 **Warning:** The path-rewrite invariant is now implemented twice with different accepted sets. This regex permits a `..` segment (`.` is inside the character class, so `/../` matches) and rejects percent-encoding, while the controller's `anchoredPath` rejects `..` and `//` but allows `%` and any character outside `?#*\ `. A rewrite that passes the contract can therefore be dropped by the controller with only a log line, and one the controller would accept can be rejected here.

  **Fix:** One invariant, one enforcement point — express the accepted shape once in the contract package and have the second language check only what it cannot trust (presence, not syntax), or pin both to a shared, tested character set so the two validators cannot diverge.

packages/agent-runtime-api/src/modules/runtime/types.ts
```diff
@@ -46,10 +46,18 @@ export const egressAllowContribution = z.object({
...
+export const pathRewrite = z.object({
+  prefix: z.string().regex(anchoredPathSegments),
+  replacement: z.string().regex(anchoredPathSegments),
+});
+
 export const egressInjectContribution = z.object({
   kind: z.literal("egress-inject"),
   host: z.string().min(1),
   pathPattern: z.string().optional(),
+  pathRewrites: z.array(pathRewrite).min(1).max(8).optional(),
```
dam-code-guardian, line 68: 🟡 **Warning:** This adds `.min(1).max(8)` plus a strict regex to a *persisted* contribution shape, and every read path silently discards contributions that fail `safeParse` (`connections-repository.ts:70`, `state-builder.ts:125`). Template contributions are only TS-checked at catalog construction, never Zod-validated, so a template author who writes `prefix: "/inference/v1"` without the trailing slash ships a connection that stores fine and then loses its whole `egress-inject` contribution on read — the agent gets no credential injection and no diagnostic.

  **Fix:** Anything that is persisted and later re-parsed with a discarding `safeParse` must be validated against the same schema at the point it is authored — parse the catalog's contributions at construction so a malformed template fails loudly at startup rather than degrading silently at read time.

packages/ui/src/modules/providers/components/ibm-litellm/form.tsx
```diff
@@ -93,6 +98,23 @@ export function IbmLitellmForm({
           {isSubmitting ? "..." : isEdit ? "Replace" : "Save"}
         </Button>
       </div>
+
+      {!isEdit && (
```
dam-code-guardian: 🟡 **Warning:** This field renders *after* the row that holds the primary submit button, so the form reads key-input → Save → another editable input. A user who pastes the token and clicks the adjacent Save never sees the model field, and keyboard tab order reaches the submit control before the last input.

  **Fix:** In a form built on `ProviderFormShell`, all always-visible inputs precede the submit row; anything that must sit below the button goes behind a collapsed disclosure (as `bob/form.tsx` does) so the visible order never puts an input after its own submit.

🟡 **Warning:** The `{!isEdit && (` gate also makes the value create-only, and `provider-connect-dialog.tsx:43` sends only `{ id, value }` to `updateConnection`. An existing `ibm-litellm` connection's Bob model can then neither be inspected nor changed from this dialog, which contradicts the PR body's claim that the model "is overridable per connection" — the adjacent `BobForm` solves the same problem by hydrating from `bobPinsFromConnection(item.conn)`.

  **Fix:** A provider config value the create form writes is also readable and editable in the edit variant, hydrated from the connection's contributions; if the update mutation cannot carry it, state that gap rather than hiding the field.

packages/ui/src/modules/providers/components/ibm-litellm/form.tsx
```diff
@@ -93,6 +98,23 @@ export function IbmLitellmForm({
...
+        <FormField
```
dam-code-guardian: 🟡 **Warning:** `bobModel` is registered with react-hook-form, but this `FormField` is given no `error` prop and the component never destructures `errors` from `formState` — any validation added to this field later would fail silently with no visible message, unlike every field in the sibling `bob/form.tsx` which passes `errors.<name>?.message`.

  **Fix:** Every RHF-registered field renders its own error slot (`error={errors.<field>?.message}`) at the moment it is added, whether or not the schema currently produces an issue for it.

packages/ui/src/modules/providers/components/ibm-litellm/form.tsx
```diff
@@ -93,6 +98,23 @@ export function IbmLitellmForm({
...
+          label="Bob model"
+          hint={`Model Bob asks this proxy for. Empty → ${IBM_LITELLM_BOB_MODEL}.`}
```
dam-code-guardian: 🟡 **Warning:** This label and hint are hand-copied into the component while the server already declares the same field with `label`, `hint` and `defaultValue` in its `configInputs` spec — and the two copies have already diverged, the server hint carrying a third sentence this one drops.

  **Fix:** Field copy for a server-declared `configInput` is read from `ConnectionTemplateView.inputs`, never restated in a component; if the bespoke provider form cannot consume that view, extend the view mapper (`connection-template.ts` drops `defaultValue` when projecting `configInputs` to inputs) rather than duplicating the string.

docs/architecture/security-and-credentials.md
```diff
@@ -516,16 +510,26 @@ Kubernetes/OpenShift clusters ([issue #2314](https://github.com/dam-agents/dam/i
   with SAN pinning unchanged. Agent-side trust is unaffected: the agent
   always trusts the platform MITM CA, never the upstream's.
 
+**Path rewriting.** An injection descriptor can declare path prefix
```
dam-code-guardian, line 513: 🟢 **Suggestion:** This new paragraph describes matching, prefix swapping and segment validation, but not how the gateway resolves two Secrets declaring different replacements for the same prefix on one host — `envoy.go:403` keeps the first and warns `conflicting path rewrite on host; keeping first`, and neither of the two new controller tests covers that path. First-wins precedence is a cross-connection coupling of the kind this page documents elsewhere.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `8d9e18c`

## PR #3565: feat(gateway): rewrite egress path prefixes so Bob can run on the LiteLLM proxy
**Author:** xjacka | **Branch:** feat/bob-openai-endpoint → main | **Changes:** +314 −62 (16 files)

### Summary
Adds per-host path-prefix rewriting to the egress gateway and points Bob at the IBM LiteLLM proxy, with the proxy's Bob model exposed as a per-connection config input. The rewrite plumbing is coherent end to end, but the seeded `BOBSHELL_API_KEY` placeholder defeats the feature it enables, and `BOB_SHELL_MODEL` now has two contributors.

### Findings
- 🔴 **Critical:** seeded `BOBSHELL_API_KEY` placeholder starts with `sk-` (`packages/api-server-api/src/modules/connections/providers.ts:40`)
- 🟡 **Warning:** `BOB_SHELL_MODEL` contributed by two templates, winner decided by grant order (`packages/api-server/src/modules/connections/domain/catalog.ts:168`)
- 🟡 **Warning:** path-rewrite invariant enforced twice with different accepted sets (`packages/agent-runtime-api/src/modules/runtime/types.ts:49`)
- 🟡 **Warning:** malformed template `pathRewrites` silently drops the whole contribution at read time (`packages/agent-runtime-api/src/modules/runtime/types.ts:60`)
- 🟡 **Warning:** the new Bob model field renders after the form's submit button (`packages/ui/src/modules/providers/components/ibm-litellm/form.tsx:102`)
- 🟡 **Warning:** the Bob model is create-only — hidden on edit and dropped by the update path (`packages/ui/src/modules/providers/components/ibm-litellm/form.tsx:102`, also `packages/ui/src/modules/providers/components/provider-connect-dialog.tsx:43`)
- 🟡 **Warning:** `bobModel` field renders no error slot (`packages/ui/src/modules/providers/components/ibm-litellm/form.tsx:103`)
- 🟡 **Warning:** field label and hint hand-copied from the server's `configInputs` spec, already diverged (`packages/ui/src/modules/providers/components/ibm-litellm/form.tsx:105`)
- 🟡 **Warning:** Existing `ibm-litellm` connections never gain the new `BOB_SHELL_MODEL` contribution. `rotateHeaderValue` rebuilds SDS fields from the stored `conn.contributions` and never consults the catalog, so a key rotation does not backfill either — every connection created before this PR keeps running Bob against whatever the pod's own default is, with no upgrade path but delete-and-recreate (`packages/api-server/src/modules/connections/services/connections-service.ts:189`).
  **Fix:** When a template gains a contribution that existing connections need, add a migration or resolve contributions from the catalog at delivery time — a change that only affects connections created after it ships leaves the installed base silently on the old behavior.
- 🟡 **Warning:** The PR body names "a full run through to a model reply" as a gate on leaving draft — "that needs a valid LiteLLM key, which this change was written without" — but the PR is not a draft, so it is offered for merge with the author's own precondition unmet and no end-to-end evidence that Bob reaches a model reply through the rewrite.
  **Fix:** When a PR body names a verification step as a gate on leaving draft, perform it and record the result, or keep the PR in draft — satisfy a stated precondition rather than changing the PR's state.

- 🟡 **Warning:** The page still says a config input projects an `env` contribution only when the user fills it, but this PR gives `ConfigInputSpec` a `defaultValue` and makes `buildHeader` fall back to it, so an untouched input now also contributes env — which is how the IBM LiteLLM connection always ships a Bob model. The general rule contradicts the behavior its own new preset section depends on (`docs/architecture/connections.md:48`).
  **Fix:** When a diff changes how a Connection Template projects contributions, update the general statement of that projection rule on the subsystem page in the same PR, not only the preset paragraph that motivated the change.
- 🟢 **Suggestion:** The new Path rewriting paragraph covers matching, prefix swapping and segment validation, but not how the gateway resolves two Secrets declaring different replacements for the same prefix on one host — the controller keeps the first and warns, and neither new controller test covers it. Reported by doc-drift and issue-fit (`docs/architecture/security-and-credentials.md:513`, `packages/controller/pkg/reconciler/envoy.go:403`).

### Issue Fit
No linked issue — no closing keyword, no plain `#<n>` reference, and the head branch carries no issue number; the only cross-reference is PR #3572. Scope checked against the PR description only.

- 🟡 **Warning:** The body promises the pinned model "is overridable per connection rather than edited in code, and the LiteLLM connect form exposes it", but the exposure is create-only — mirrored into `### Findings` above (`packages/ui/src/modules/providers/components/ibm-litellm/form.tsx:102`).
- 🟡 **Warning:** The PR body names a full run through to a model reply as a gate on leaving draft, and the PR is not a draft — mirrored into `### Findings` above.

- 🟡 **Warning:** The `ibm-litellm` template now contributes `BOB_SHELL_MODEL` with a default that always fires — the same env name the `bob` template already contributes from its `model` config input. An owner who grants both connections gets whichever value belongs to the connection created first, because every consumer dedupes first-wins (`state-builder.ts` concatenates grants in `createdAt` order, `env-plugin.ts` keeps the first `name`, the controller's `credentialEnvVars` keeps the first `envName`). The silent loser is exactly what the new hint warns about — a Bob-native model name reaching the proxy, or the proxy model reaching Bob's own gateway (`packages/api-server/src/modules/connections/domain/catalog.ts:168`).
  **Fix:** Two templates must never contribute the same env name from independent inputs; when a variable is proxy-scoped, derive it per host at delivery time or make the collision explicit (reject/flag the second grant) instead of relying on grant order to pick a winner.
- 🟡 **Warning:** The path-rewrite invariant is implemented twice with different accepted sets: the Zod regex `^\/(?:[A-Za-z0-9._~-]+\/)*$` permits a `..` segment (`.` is in the class) and rejects percent-encoding, while the controller's `anchoredPath` rejects `..` and `//` but allows `%` and anything outside `?#*\ `. A rewrite that passes the contract can be dropped by the controller with only a log line, and the reverse (`packages/agent-runtime-api/src/modules/runtime/types.ts:49`).
  **Fix:** One invariant, one enforcement point — express the accepted shape once in the contract package and have the second language check only what it cannot trust (presence, not syntax), or pin both to a shared, tested character set.
- 🟡 **Warning:** `pathRewrites` adds `.min(1).max(8)` and a strict regex to a *persisted* contribution shape, and every read path silently discards contributions that fail `safeParse` (`connections-repository.ts:70`, `state-builder.ts:125`). Template contributions are only TS-checked at catalog construction, never Zod-validated, so a template author who writes `prefix: "/inference/v1"` (no trailing slash) ships a connection that stores fine and then loses its whole `egress-inject` contribution on read — no credential injection, no diagnostic (`packages/agent-runtime-api/src/modules/runtime/types.ts:60`).
  **Fix:** Anything persisted and later re-parsed with a discarding `safeParse` must be validated against the same schema where it is authored — parse the catalog's contributions at construction so a malformed template fails loudly at startup rather than degrading silently at read time.

- 🟡 **Warning:** The new "Bob model" field renders *after* the row holding the primary submit button, so the form reads key-input → Save → another editable input. A user who pastes the token and clicks the adjacent Save never sees the model field, and tab order reaches the submit control before the last input (`packages/ui/src/modules/providers/components/ibm-litellm/form.tsx:102`).
  **Fix:** In a form built on `ProviderFormShell`, all always-visible inputs precede the submit row; anything below the button goes behind a collapsed disclosure (as `bob/form.tsx` does), so the visible order never puts an input after its own submit.
- 🟡 **Warning:** The field's label and hint are hand-copied into the component while the server already declares the same field with `label`, `hint` and `defaultValue` in its `configInputs` spec — and the two copies have already diverged, the server hint carrying a third sentence the UI drops (`packages/ui/src/modules/providers/components/ibm-litellm/form.tsx:105`).
  **Fix:** Field copy for a server-declared `configInput` is read from `ConnectionTemplateView.inputs`, never restated in a component; if the bespoke form cannot consume that view, extend the view mapper (it drops `defaultValue` when projecting `configInputs`) rather than duplicating the string.
- 🟡 **Warning:** `bobModel` is registered with react-hook-form but its `FormField` gets no `error` prop and the component never destructures `errors` from `formState`, so any validation added to this field later fails silently — unlike every field in the sibling `bob/form.tsx` (`packages/ui/src/modules/providers/components/ibm-litellm/form.tsx:103`).
  **Fix:** Every RHF-registered field renders its own error slot (`error={errors.<field>?.message}`) when it is added, whether or not the schema currently produces an issue for it.
- 🟡 **Warning:** The field is hidden on the `edit` variant and the edit path sends only `value`, so an existing `ibm-litellm` connection's Bob model can neither be inspected nor changed from this dialog — the adjacent `BobForm` solves the same problem by hydrating from `bobPinsFromConnection(item.conn)` (`packages/ui/src/modules/providers/components/ibm-litellm/form.tsx:102`, `packages/ui/src/modules/providers/components/provider-connect-dialog.tsx:43`).
  **Fix:** A provider config value the create form writes is also readable and editable in the edit variant, hydrated from the connection's contributions; if the update mutation cannot carry it, state that gap rather than hiding the field.

### Verdict
REQUEST_CHANGES — the `sk-dummy` placeholder silently downgrades Bob to a legacy backend, defeating the change's purpose, and `BOB_SHELL_MODEL` now has two contributors resolved by grant order.


---

xjacka:
All findings addressed in `3e26981b`. Two I resolved differently than suggested, and one of the diagnoses does not hold on the version this image ships — details below.

## 🔴 `sk-` placeholder — fixed, but the stated cause is 1.x behavior

Changed to the inert `dummy-placeholder`. The reasoning in the finding does not apply to Bob 2.0.2 though, and it is worth recording so nobody re-derives it: I checked the shipped bundle out of `platform-bob:latest`, and `startsWith("sk-")`, `ibm-bob-staging` and `OLD_BACKEND` have **zero** occurrences in it. The prefix branch was 1.0.6's (`DU(key) = key.startsWith("sk-") || key.startsWith("pk-")` feeding `isAuthnBackend()`, with `DEFAULT_OLD_BACKEND_URL` pointing at the staging host); 2.0.2 dropped it and compiled the service prefixes in as constants. A headless 2.0.2 run with `BOBSHELL_API_KEY=sk-dummy` also stayed on the configured gateway rather than downgrading.

So this was not a live downgrade, but the placeholder bought nothing either, and the agent README documents the `sk-`/`pk-` rule as a standing constraint. Inert placeholder is the right value on both counts.

## 🟡 Doubled `BOB_SHELL_MODEL` — fixed at the cause, not with a collision check

Both technical claims check out (`state-builder.ts` orders grants by `createdAt`, and the env consumers keep the first value for a name). What made it silent was the *always-firing default*, so that is what went: `ConfigInputSpec` no longer has `defaultValue`, `buildHeader` no longer falls back to one, and the LiteLLM template contributes `BOB_SHELL_MODEL` only when the connection names a model. Removing it also restored the general projection rule the doc-drift finding flagged — an untouched config input contributes nothing again, so `connections.md:48` is true as written and needed no edit.

A collision is still *possible* when both connections carry a model, and that case is now named in the agent README. I did not add the reject-or-flag mechanism: it belongs to the env rail generally rather than to Bob, `state-builder` has no logger to report through, and doing it properly means deciding whether a doubled name is an error or a precedence rule for every template. Happy to file that as its own issue.

## 🟡 Two invariants, two accepted sets — unified

You were right that the sets differed, and in the dangerous direction: `[A-Za-z0-9._~-]+` admits a `..` segment. The contract now rejects relative segments outright, and the controller's read-back check rejects the same set (percent-encoding included). It stays enforced in both places on purpose — the controller reads a Secret it did not author in this process — but the accepted shape is one shape, and a new test covers the drop path.

## 🟡 Malformed template contribution lost on read — fails loudly now

`buildCatalog` parses every template contribution against the runtime contract and throws on the first that fails, so a template written with (say) a trailing-slash-less prefix breaks at startup instead of quietly shipping a connection whose `egress-inject` disappears on read.

## 🟡 UI findings — three fixed, one accepted as a limitation

Fixed: the model field renders above the submit row; it has an `error` slot; and its label, hint and placeholder come from constants in the contract package, so the copy has one source and cannot diverge from the server's `configInputs` spec again.

Not fixed: the field stays **create-only**. Making it editable means carrying config inputs through the update mutation and re-projecting contributions on an existing connection — a change to the connection update path, not to this form. Per your own alternative I stated the gap instead: it is in the agent README, and the PR body no longer claims the value is "overridable per connection" without qualification.

## 🟡 Existing connections gain nothing — named, not migrated

Correct, and it follows from contributions being projected once at create time. Recreating the connection is the upgrade path, and that is now written down. A migration or resolve-from-catalog-at-delivery is an architectural change to the Connections subsystem and does not belong in this PR.

## 🟡 Draft precondition — the body was the problem

Fair hit. The verification section no longer frames the end-to-end run as a gate on leaving draft; it states plainly what the evidence covers (the request arrives and is refused on auth alone) and that "Bob answers through the rewrite" is untested, because I do not have a LiteLLM key.

## 🟢 Conflicting rewrites — documented and tested

Added to `security-and-credentials.md` (one prefix carries one replacement; conflicting Secrets keep the first and log the loser) and covered by a new controller test.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `15299d0`

## PR #3565: feat(gateway): rewrite egress path prefixes so Bob can run on the LiteLLM proxy
**Author:** xjacka | **Branch:** feat/bob-openai-endpoint → main | **Changes:** +391 −64 (15 files)

### Summary
Round 2 closes every blocking finding at its cause: the seeded placeholder is inert, `defaultValue` is gone from the config-input mechanism end to end so the LiteLLM template contributes `BOB_SHELL_MODEL` only when a connection names one, the accepted path shape rejects relative segments on both sides, the form field moved above the submit row with its error slot and shared copy constants, and `buildCatalog` now parses every declared contribution against the runtime contract before the server can boot.

_(Suppressed 2 finding(s) per PR-local overrides: the create-only Bob model field, existing connections not migrated.)_

_15 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: 8d9e18c (2026-09-07T11:29:38Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** BOBSHELL_API_KEY placeholder starts with sk- (`packages/api-server-api/src/modules/connections/providers.ts:42`)
- ✅ **Fixed:** two templates contribute BOB_SHELL_MODEL (`packages/api-server/src/modules/connections/domain/catalog.ts:168`)
- ✅ **Fixed:** path–rewrite invariant enforced twice, sets differ (`packages/agent-runtime-api/src/modules/runtime/types.ts:51`)
- ✅ **Fixed:** malformed pathRewrites silently drops the contribution (`packages/api-server/src/modules/connections/domain/catalog.ts:841`)
- ✅ **Fixed:** model field renders after the submit button (`packages/ui/src/modules/providers/components/ibm-litellm/form.tsx:91`)
- ✅ **Fixed:** bobModel field renders no error slot (`packages/ui/src/modules/providers/components/ibm-litellm/form.tsx:95`)
- ✅ **Fixed:** label and hint hand–copied from server spec, diverged (`packages/ui/src/modules/providers/components/ibm-litellm/form.tsx:93`)
- ✅ **Fixed:** projection rule contradicts the new defaultValue fallback (`docs/architecture/connections.md:48`)
- ✅ **Fixed:** out of draft with its own verification gate unmet
- ✅ **Fixed:** path–rewrite conflict resolution undocumented and untested (`docs/architecture/security-and-credentials.md:513`)
- 🆕 **New:** docs and body describe the mechanism round 2 removed (`docs/architecture/connections.md:201`)
- 🆕 **New:** the two new TS guards have no test (`packages/agent-runtime-api/src/modules/runtime/types.ts:51`)
- 🆕 **New:** Go read–back set still laxer than the contract (`packages/agent-runtime-api/src/modules/runtime/types.ts:49`)

### Findings
- 🟢 **Suggestion:** The prose lags the code on three counts. `docs/architecture/connections.md:201` attributes the model collision to "grant order", while `state-builder.ts:114` orders granted contributions by `createdAt` — connection creation order, which the new `packages/agents/bob/README.md` paragraph states correctly, so one PR ships two accounts. The same page reads "A model has to be named on the Connection as well" as a requirement, though nothing rejects the blank case: the input is optional, `build-connection.ts:570` skips an empty value, and the UI schema is a bare `z.string()`. Both architecture pages still carry `Last verified: 2026-09-04` after a 2026-09-07 edit, and the PR body's summary still describes the shipped default this round removed. Reported by doc-drift and issue-fit (`docs/architecture/connections.md:201`, also `docs/architecture/security-and-credentials.md:3`).
- 🟢 **Suggestion:** The two guards this round added on the TypeScript side have no test. `anchoredPath`'s `.refine` and `assertContributionsParse` are reachable only by feeding them a bad value, and nothing does — the new unit test file covers the env rail and the Secret annotation, so deleting either guard passes CI, while the Go counterpart gained three tests in the same round (`packages/agent-runtime-api/src/modules/runtime/types.ts:51`).
- 🟢 **Suggestion:** The invariant is still enforced twice with different accepted sets. A segment must match `[A-Za-z0-9._~-]+` in Zod, while the Go read-back rejects only `?#*%\`, space, `//` and relative segments, so `/a=b/` or `/a;b/` is legal for the controller and rejected by the contract. Round 2 narrowed the Zod side and taught both to reject `..`, which is what the blocking finding asked for, but nothing pins the two sets to each other, so the next edit to either drifts silently (`packages/agent-runtime-api/src/modules/runtime/types.ts:49`, also `packages/controller/pkg/reconciler/envoy.go:262`).

- 5 finding(s) reported: the two stale `Last verified:` stamps, the "grant order" attribution, the unenforced "has to be named" requirement — all four carried into the prose item in `### Findings` above — and the Template section not recording that `assertContributionsParse` now decides whether the subsystem boots. The fifth is dropped under the budget.

### Issue Fit
_No linked issue — no closing keyword, no plain `#<n>` reference, and the head branch carries no issue number; scope checked against the PR description only._

Round-1 recheck: the draft/verification gate is cleared — the Verification section now states what the evidence covers and marks the end-to-end model reply as not verified, rather than naming it a precondition for leaving draft. No undeclared scope creep: every round-2 change is declared in the body or the author's round-1 reply.

- 3 finding(s) reported: the body still describing the removed shipped default (carried into `### Findings` above), a controller-test tally that says two where the diff adds three, and a `[the review reply](#issuecomment)` link that resolves nowhere. The latter two are dropped under the budget.

- 9 finding(s) reported: the two carried into `### Findings` above are the untested guards and the residual Zod/Go set divergence. Of the seven dropped, the two closest are that `assertContributionsParse` covers only statically declared contributions — the `configInputs`-derived `env` entries built from user input in `build-connection.ts` bypass it — and that `pathRewrites` is order-significant downstream (`buildChainForwardRoutes` emits one route per entry ahead of the catch-all) while the schema neither rejects duplicate prefixes nor normalises order.

- 7 finding(s) reported. Two are settled by the author's stated limitations and recorded as PR-local overrides: the create-only field and the absent prefill on edit. Of the five dropped, the closest is that the new `error={errors.bobModel?.message}` slot cannot yet render — `bobModel` is a bare `z.string()` and the schema's `superRefine` raises issues only for `value`, so a malformed model id still surfaces as a generic server toast even though the server spec supports `pattern`/`enumValues` for this input, as `bob/form.tsx` shows for `maxCost` and `chatMode`.

### Verdict
APPROVE — every blocking finding from round 1 is closed at its cause; what remains is documentation wording, test coverage for the two new guards, and a cross-language set that is safe but still unpinned, none of which blocks the merge.


---

xjacka:
Closing as superseded: this change is already on `main`.

#3553 was branched off `feat/bob-openai-endpoint` and carried this work with it when it was squash-merged, so git no longer sees the relationship — the squash commit is not a descendant of the commits here, which is why the PR still shows as conflicting.

Verified on `main` (`8b088624`), item by item:

- `BOB_INFERENCE_PREFIX_REWRITE` — defined in `packages/api-server-api/src/modules/connections/providers.ts`, exported from `packages/api-server-api/src/index.ts`
- `pathRewrites` on the runtime contract — `packages/agent-runtime-api/src/modules/runtime/types.ts`, projected in `connection-sds.ts`
- IBM LiteLLM template — the rewrite contribution, the "Claude Code and Bob" description, and `assertContributionsParse` in `packages/api-server/src/modules/connections/domain/catalog.ts`
- Envoy route rendering and rewrite validation — `packages/controller/pkg/reconciler/envoy.go`, `envoy_bootstrap.go`
- Tests — all three in `connections-ibm-litellm-template.test.ts`, plus the route-order and malformed-rewrite cases in `envoy_test.go`
- Docs — the rewrite mechanics in `docs/architecture/connections.md`, `security-and-credentials.md`, and the LiteLLM section of `packages/agents/bob/README.md`

Merging `main` into this branch conflicts only in three documentation spots, and `main` holds the newer text in each: the `Last verified` date, the paragraph on the Connection contributing no model, and the model bullet in the Bob README (which now reflects the Config panel that shipped with #3553, where model is deliberately absent). Resolving all three in favour of `main` leaves a tree identical to `main` — the merged diff is empty, so there is nothing left to merge here.

PR #3566, opened 2026-09-04
feat(ui): open the chat sidebar sections by default and remember the choice

### LINKED ISSUE #3455: UI - Artifacts an agent made are hidden until the user opens the panel
## Context

The chat view lists a session's artifacts in a side panel. That panel starts collapsed on every chat.

## Problem

A user does not see the artifacts the agent produced. The panel is collapsed by default, so the user must know it exists and click to open it. Artifacts are a main output of a session. Today they are the least visible part of the chat view.

## Goal

A user sees the session's artifacts on landing in a chat, with no click.

## User Stories

- As a user, I want to see a session's artifacts when I open the chat so that I know what my agent produced.
- As a user, I want my collapse choice to be remembered so that I control the layout.

## Proposed solution

Expand the artifacts panel by default in the chat view. Keep the collapse control and remember the user's last choice.

## Open Questions

- Expand also when the session has no artifacts, or only when at least one exists?
- Persist the collapse state per agent, or globally?

PetrBulanek:
A user opens a chat and sees the session's artifacts immediately, with no click. The Artifacts section of the chat sidebar starts expanded instead of collapsed.

The collapse control stays. When a user collapses a section, that choice now survives a reload — for all three sidebar sections, Sessions and Files as well as Artifacts, so the sidebar behaves as one thing rather than one section that remembers and two that forget.

The section expands even when the session produced nothing, so there is one rule and no flicker. The choice is remembered globally, not per agent.

Closes #3455

## Sub-issues

- [x] 01 — Chat sidebar sections open by default and remember the choice


packages/ui/src/lib/persisted-prefs.ts
```diff
@@ -0,0 +1,42 @@
...
+  safeSetItem,
+} from "./safe-storage.js";
+
+export function readPersistedFlag(
+  key: string,
+  fallback: boolean,
+  store: KeyValueStore = browserStorage,
+): boolean {
+  const raw = safeGetItem(store, key);
+  return raw === null ? fallback : raw === "1";
```
dam-code-guardian, line 14: 🟢 **Suggestion:** `readPersistedFlag` maps every stored value that is not `"1"` to `false`, while `readPersistedNumber` validates and returns the caller's fallback. No writer here emits any other encoding, so nothing is broken today, but the two readers now teach opposite rules for an unusable value. Consider recognizing `"1"` and `"0"` and falling back on anything else, so both readers follow one rule.

packages/ui/src/modules/sessions/views/chat-view.tsx
```diff
@@ -168,17 +178,19 @@ export function ChatView() {
...
-  const [leftW, setLeftW] = useState(
-    () => Number(localStorage.getItem("platform-left-w")) || 220,
+  const [leftW, setLeftW] = useState(() =>
+    readPersistedNumber(LEFT_WIDTH_KEY, 220),
   );
-  const [rightW, setRightW] = useState<number | null>(
-    () => Number(localStorage.getItem("platform-file-w")) || null,
+  const leftWRef = useRef(leftW);
+  const [rightW, setRightW] = useState<number | null>(() =>
+    readPersistedNumber(FILE_PANEL_WIDTH_KEY, null),
```
dam-code-guardian, line 186: 🟢 **Suggestion:** the clamps (140/400, 120/600, 240/`min(960, innerWidth-500)`) run only on the write path, so a width stored on a wide display renders unclamped after a reload on a narrow window. `main` behaves the same way, so this PR does not add the gap. It does add the one place that could close it: this PR centralizes the read, so a `usePersistedSize(key, { fallback, min, max })` hook could apply the same bounds on hydration and on write.

packages/ui/src/modules/sessions/views/chat-view.tsx
```diff
@@ -168,17 +178,19 @@ export function ChatView() {
...
   const setTerminalPaused = useStore((s) => s.setTerminalPaused);
 
-  const [leftW, setLeftW] = useState(
-    () => Number(localStorage.getItem("platform-left-w")) || 220,
+  const [leftW, setLeftW] = useState(() =>
+    readPersistedNumber(LEFT_WIDTH_KEY, 220),
   );
-  const [rightW, setRightW] = useState<number | null>(
-    () => Number(localStorage.getItem("platform-file-w")) || null,
+  const leftWRef = useRef(leftW);
```
dam-code-guardian, line 184: 🟢 **Suggestion:** the three resize handlers change from functional `setState` updaters to `useRef` mirrors. The rewrite is correct — those handlers are the only writers of the three values, and each advances its ref before it calls the setter — but the PR body describes only the default-open and persistence change. A reader of the body does not expect a change to how unrelated state updates. Name the refactor in the body, so the diff and the description agree.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `83c52be`

## PR #3566: feat(ui): open the chat sidebar sections by default and remember the choice
**Author:** PetrBulanek | **Branch:** feat/3455-artifacts-panel-default-open → main | **Changes:** +244 −56 (10 files)

### Summary
The three chat sidebar sections now open by default, and each collapse choice survives a reload. A new `lib/persisted-prefs.ts` holds the encoding, and `safe-storage.ts` moves up to `lib/` to serve it.

_13 suggestion(s) dropped under the 🟢 budget._

### Findings
- 🟢 **Suggestion:** the two readers disagree on an unusable stored value (`packages/ui/src/lib/persisted-prefs.ts:14`)
- 🟢 **Suggestion:** the new central read path does not clamp a hydrated geometry value (`packages/ui/src/modules/sessions/views/chat-view.tsx:186`)
- 🟢 **Suggestion:** the resize rewrite is in the diff but not in the PR body (`packages/ui/src/modules/sessions/views/chat-view.tsx:184`)
- ✅ **Looks good:** the three resize handlers change from functional `setState` updaters to `useRef` mirrors. This is safe here: `leftW`, `rightW` and `sessionsH` have no writer except those three handlers, and each handler writes its ref before it calls the setter. No path can make a ref stale.

✅ No findings.

### Issue Fit
The issue's ask is delivered: the artifacts panel opens by default, the collapse control stays, and the choice persists. All 3 findings are 🟢 and merged into `### Findings` or dropped under the 🟢 budget.

All 7 findings are 🟢 after verification, and merged into `### Findings` or dropped under the 🟢 budget. Three arrived as blocking and do not clear the bar: the artifacts default flip is declared by the PR title, the PR body and issue #3455; a stored `0` or negative width needs a value no writer can produce, because the old and the new write paths both clamp; and the flag reader keeps the exact comparison the one pre-existing key already used.

All 6 findings are 🟢 after verification, and merged into `### Findings` or dropped under the 🟢 budget. Four arrived as blocking and do not clear the bar: no code writes the three geometry values except the handlers that advance the refs; and the per-drag `localStorage` write and the unclamped hydration are both present on `main` before this PR.

### Verdict
APPROVE — the change does what the issue asks, the storage boundary absorbs a throwing store, and no finding survives verification as blocking.


---

kapetr (approved):
The issue objective works well. But with three panels in the sidebar now, we should be able to change the proportions on both dividers (Files and Artifacts) and in every state. Right now you can only resize the Files panel (and resizing range seems limited), and only while it is uncollapsed. Fine for a follow up issue.

PR #3569, opened 2026-09-04
feat: knowledge bases run on any harness, not only Claude Code

### LINKED ISSUE #3511: UI - Knowledge base agents can only run on Claude Code
## Context

Knowledge bases shipped as a thin model: a knowledge base is an agent that installs its own tooling. But when a user creates one, the harness is fixed to Claude Code and hidden — unlike code agents, where a user picks from our available harnesses. As we broaden the agent types we major on, and ahead of redesigning agent creation into a single consistent flow, we want knowledge bases to work across harnesses rather than being tied to one.

## Problem

A user creating a knowledge base cannot choose their harness or model provider. The choice is made for them — Claude Code only — and the option is hidden from the creation flow. A user who works on Pi, Bob, or Codex, or who must use a specific model provider, cannot create a knowledge base at all on their terms. This makes knowledge bases feel like a separate, narrower product than code agents, and it limits adoption to users willing to use Claude Code.

## Goal

A user creating a knowledge base can choose from our available harnesses (Claude Code, Pi, Bob, Codex) and select their model provider during creation, the same way they can for a code agent.

## User Stories

- As a user, I want to pick my harness when creating a knowledge base so that I can use the agent I already work with.
- As a user, I want to choose my model provider when creating a knowledge base so that I'm not forced onto one I can't or don't want to use.
- As a user, I want knowledge base creation to feel like the rest of agent creation so that the choice of feature doesn't lock me into unrelated choices.

## Scope

- The harness and model-provider selection a user already gets for code agents, made available when creating a knowledge base.
- Whatever knowledge base setup currently assumes about Claude Code should work, degrade gracefully, or be clearly surfaced as unsupported on a given harness.
- Not in scope: the broader single creation-flow redesign (see below). This is a precursor that unblocks it.

## Proposed solution

The knowledge base creation flow currently pins the harness to Claude Code and hides the picker; the underlying create path already accepts any image. Expose harness and model-provider selection in the flow, and generalize the parts of knowledge base setup that assume Claude-Code-specific paths so the installed tooling lands correctly on the chosen harness.

## Open Questions

- Do all four harnesses need to support knowledge bases at launch, or do we start with a subset and surface the rest as they land?
- If a harness can't yet support knowledge base setup, do we hide it for this flow or show it as unavailable with a reason?
- How much of the Claude-Code-specific setup generalizes cleanly vs. needs a per-harness equivalent?

## Additional resources

This is a precursor to the single agent-creation-flow redesign now being picked up by design — decoupling the harness here is what lets that flow treat knowledge bases consistently with other agent types.

---

_Filed by @jenna-winkler_

Tomas2D:
Closes #3511

## Why

A user creating a knowledge base could not pick a harness or a model provider: the form pinned the Claude Code image (hidden) and offered only the two credentials that reach Claude. Knowledge bases felt like a narrower product than coding agents.

## What

**UI** — the knowledge base form has the same **Harness** picker and the **full provider list** coding agents get (Claude Code preselected). The catalogue, default selection, stale-draft reconciliation, and the loading / failed / empty grid are one shared hook (`useHarnessCatalogue`) + component (`HarnessGrid`) that both setup flows use. `setupProviderPolicy` is an exhaustive per-flow record; experiments stay restricted.

**Harness families** — each agent template may declare a `harness` family in the Helm chart (`claude-code` for claude-code and claude-code-vm, `codex`, `pi`, `bob`). The vocabulary is a closed enum owned by the templates contract and validated when templates load, so a misdeclared family rejects the template (logged) rather than degrading silently. The family is exposed on the templates API.

**Install command** — the KB service resolves the chosen template's family and hands it to the template's bootstrap through a template-specific env var. Both bootstraps are harness-aware and land their own tooling:

- LLM Wiki: `LLM_WIKI_HARNESS` — dam-agents/llm-wiki-v2#2, #3 (merged)
- Plain Wiki: `PLAIN_WIKI_HARNESS` — dam-agents/plain-wiki#2 (merged)

A template with no declared family (custom image, the e2e mock) gets the bare command, leaving the bootstrap to detect the harness. The platform appends no shell of its own.

**Greeting** — the onboarding command is issued in the form the KB's harness expands installed commands under (`/prompts:wiki-onboard` on Codex, `/wiki-onboard` elsewhere), resolved from the agent's template family.

## Open Questions from #3511

**Do all four harnesses need to support knowledge bases at launch, or a subset?** All four ship, and all four have now been exercised live on the local cluster (all four harness images loaded, KBs created through the real create path):

| Harness | Plain Wiki | LLM Wiki |
|---|---|---|
| Claude Code | unchanged path (pre-existing behaviour) | unchanged path |
| Codex | bootstrap ran with the family; manual in `AGENTS.md`, command in `~/.codex/prompts` | skill → `~/.agents/skills/llm-wiki`, manual → `~/.codex/AGENTS.md`, 7 commands in `~/.codex/prompts` |
| Pi | bootstrap ran with the family; manual in `AGENTS.md`, command in `~/.pi/agent/prompts` | skill → `~/.agents/skills/llm-wiki` (also visible under `~/.pi/agent/skills`), manual → `~/.pi/agent/AGENTS.md`, 7 commands in `~/.pi/agent/prompts` |
| Bob | bootstrap ran; `AGENTS.md` + `~/.bob/rules/*` load (`RuleLoader: agents=true gCommon=2`); Bob auto-converts the command into a skill | skills listed by the agent in a live chat |

Codex/Pi were exercised through the install path (create → pod → bootstrap log → landed files), not through a chat turn — the platform's contribution on those harnesses ends at the family name; the greeting form is unit-tested per family.

**If a harness can't support KB setup, hide it or show it as unavailable with a reason?** Moot for the shipped set — no offered harness is unsupported, so nothing is hidden or greyed out. A future family a bootstrap cannot wire is not silent either: the installer falls back to its own harness detection and the manual's onboarding rule covers a harness without command files.

## Verification

- api-server 1312 tests, UI 348 tests (new: harness-selection reconciliation, KB draft builder, onboarding-command mapping), tsc/lint/format, helm lint+render, docs size/ADR checks, comment-type check — all green.
- Live: Plain Wiki and LLM Wiki on Bob (above); cluster ran with all four harness images.
- The earlier `mise e2e` failure was in CI's `pnpm install` bootstrap (corepack download assertion before any test ran); main's runs pass, and this push reruns it.

## Notes for reviewers

- Custom images are deliberately not offered in the KB form — a follow-up if wanted.
- Known Bob image issue, out of scope: `bobshell` 2.0.0 imports `@vscode/ripgrep` without declaring it (search tool broken); handled by the Bob upgrade.
- Architecture pages updated: `docs/architecture/knowledge-bases.md`, `docs/architecture/agent-lifecycle.md`.


packages/ui/src/modules/sandboxes/lib/setup-policy.ts
```diff
@@ -11,6 +11,7 @@ export function setupProviderPolicy(flow: SetupFlow): {
   allow?: readonly ProviderPresetType[];
   recommended?: ProviderPresetType;
 } {
-  if (flow === "coding-agent") return { recommended: "ibm-litellm" };
-  return { allow: KINDED_PROVIDERS, recommended: "ibm-litellm" };
+  if (flow === "experiment")
```
dam-code-guardian: 🟡 **Warning:** `setupProviderPolicy` branches on the closed union `SetupFlow` (`"coding-agent" | "experiment" | "knowledge-base"`) with one `if` plus an unnamed fallback, so only `experiment` is named at the branch point. The fallback is also the permissive arm (no `allow` list at all), so a fourth member added later silently gets the every-provider policy instead of a compile error — and this PR relies on that arm for the knowledge-base flow.
**Fix:** Branch a closed string-literal union with `match(...)` from `ts-pattern`, one `.with()` per member closed by `.exhaustive()` (`docs/guidelines/code/exhaustive-matching.md` — the rule this same PR follows in `install-command.ts`), and never let a policy function's catch-all arm be the least restrictive one.

packages/api-server-api/src/modules/templates/schemas.ts
```diff
@@ -38,6 +38,7 @@ export const templateSpecSchema = z
     name: z.string().optional(),
     description: z.string().optional(),
     category: templateCategorySchema,
+    harness: z.string().optional(),
```
dam-code-guardian: 🟡 **Warning:** `harness` is `z.string().optional()` even though the same package now defines the closed enum `kbHarnessFamilySchema` for exactly these values. A misspelled `harness:` in `deploy/helm/platform/values.yaml` (say `claude_code`) validates fine, is dropped by `parseKbHarnessFamily`'s `safeParse`, and the knowledge base installs with no harness wiring and no diagnostic anywhere — "invalid" and "absent" map onto the same silent `undefined`.
**Fix:** Validate a field whose value set is closed with the enum schema at the boundary that first parses it; when an unknown value must be tolerated for forward compatibility, make that degradation observable (log or surface the rejected value) instead of silent.

packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx
```diff
@@ -87,6 +113,26 @@ export function KnowledgeBaseSetupView() {
         </CardGrid>
       </section>
 
+      <section className="mb-8">
+        <SectionLabel spaced>Harness</SectionLabel>
+        {isLoading ? (
```
dam-code-guardian: 🟡 **Warning:** `isLoading` is the only branch here, so when `useTemplates()` fails — or returns no `category: "harness"` template — the section renders an empty `CardGrid` while `canCreate` stays false through `harnesses.some((t) => t.id === form.templateId)`. The user gets a blank area under a "Harness" label and a permanently disabled "Create knowledge base agent" button with nothing explaining why, and unlike the coding-agent flow there is no custom-image path to finish setup.
**Fix:** Any query whose result gates a submit button renders three states, not two — loading, failed/empty, and loaded: branch on `isError`/`isLoading` and show an explicit message plus a retry, so a disabled action always has a visible reason beside it.

packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx
```diff
@@ -23,23 +32,40 @@ const RETURN_PATH = routeToPath({ view: "knowledge-base-new" });
...
 
+  const harnesses = useMemo(
+    () =>
+      imageCatalogue(templates ?? [], {
+        vmFeatureEnabled: flags?.["vm-sandboxes"] ?? false,
+      }).harnesses,
+    [templates, flags],
+  );
+  const preselected = useRef(false);
+  useEffect(() => {
```
dam-code-guardian: 🟡 **Warning:** The effect latches `preselected.current = true` on the first non-empty `harnesses` and returns early whenever `form.templateId !== null`, so a `templateId` rehydrated from `sessionStorage` that is not in the fetched catalogue (the `vm-sandboxes` flag since turned off, or the template retired server-side) is never cleared or replaced. No card shows as selected, `canCreate` is false, and no reconciliation can run again.
**Fix:** When persisted UI state references server-owned ids, validate it against the server list on arrival rather than only defaulting when it is absent — drop or replace a selection no longer in the fetched set, and keep that reconciliation live instead of behind a one-shot ref latch.

packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx
```diff
@@ -23,23 +32,40 @@ const RETURN_PATH = routeToPath({ view: "knowledge-base-new" });
...
-    },
+    { kbTemplateId: DEFAULT_KB_TEMPLATE_ID },
     RETURN_PATH,
   );
+  const { data: templates, isLoading } = useTemplates();
+  const { data: flags } = useFeatures();
   const createKnowledgeBase = useCreateKnowledgeBase();
   const openKnowledgeBase = useStore((s) => s.openKnowledgeBase);
 
+  const harnesses = useMemo(
```
dam-code-guardian: 🟡 **Warning:** This catalogue `useMemo` and the preselect effect below it are a second copy of `coding-agent-setup-view.tsx:52-69` — same `imageCatalogue(templates ?? [], { vmFeatureEnabled: flags?.["vm-sandboxes"] ?? false }).harnesses`, same `useRef` latch, same `KINDED_HARNESS_TEMPLATE_ID` default — and the copies have already diverged (the coding-agent one also guards on `form.customImage`). Both views must now be edited in lockstep for any catalogue or default-selection change.
**Fix:** A server-derived catalogue plus its default-selection behavior belongs in one shared hook (e.g. `useHarnessCatalogue()` beside `image-catalogue.ts`) that both setup views call; parameterize the hook for the extra guard rather than re-implementing another view's data derivation.

packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx
```diff
@@ -87,6 +113,26 @@ export function KnowledgeBaseSetupView() {
         </CardGrid>
       </section>
 
+      <section className="mb-8">
+        <SectionLabel spaced>Harness</SectionLabel>
+        {isLoading ? (
+          <CardList>
+            <ListSkeleton rows={2} rowHeight={156} />
+          </CardList>
```
dam-code-guardian: 🟡 **Warning:** This block duplicates markup `ImageSection` already owns — the same `isLoading ? <CardList><ListSkeleton rows={2} rowHeight={156} /></CardList> : <CardGrid>{harnesses.map(...HarnessCard...)}</CardGrid>` sits at `packages/ui/src/modules/sandboxes/components/setup/image-section.tsx:39-53`. The skeleton geometry (`rows={2}`, `rowHeight={156}`) is now a magic pair in two files, so the two harness pickers will drift visually.
**Fix:** Extract the shared presentation — a `HarnessGrid` carrying its own skeleton geometry — and have both `ImageSection` and this view render it; compose an existing setup component rather than copying its JSX with the custom-image half removed.

docs/architecture/knowledge-bases.md
```diff
@@ -9,21 +9,21 @@ A **Knowledge Base** is an Agent that builds and maintains a body of knowledge t
...
...oling reaches the chosen harness is the template's concern: every bootstrap is harness-aware, and the platform's whole contribution is to name the **harness family** for it (a template-specific environment variable on the command) — the installer then lands its manual, commands, and any skill in that harness's own surfaces, and detects the harness itself when no family is named (a custom image, or a template that declares none). The platform appends no shell of its own. Each template's bootstrap installs a `/wiki-onboard` command, so the greeting (below) stays template-agnostic; on a harness that does not expand it natively (Codex exposes installed prompts under a `/prompts:` prefix, Bob has no command files) the greeting reaches the model as plain text and the installed manual tells it what to do with it. The platform ships the pointer, never the tooling. The picked template id rides the same create-time annotation stamp as the Kind marker and is surfaced on the Agent view (opaque to the agents module — the knowledge-bases surface owns the id set), so the KB list can display which procedure a knowledge base came from; it is absent on knowledge bases created before it was recorded.
```
dam-code-guardian: 🟡 **Warning:** This paragraph pulls volatile, code-level shape onto an architecture page: the harness-family value set is transcribed as the code's enum members (Claude Code, Codex, Pi, Bob), the transport is named as "a template-specific environment variable on the command", the declaration site is named as the Helm chart, and a third party's literal `/prompts:` prefix is quoted. `docs/guidelines/documentation-guidelines.md:41` puts exact enum members, the Helm template tree and code-level shape under **Omit** — each of these breaks when a family is added, an env var renamed, or the values file restructured.
**Fix:** Describe a contract semantically — what is exchanged and what each outcome means — then link out to the contract package or chart as the field-level source of truth; never enumerate an enum's members, name an env var, or quote a literal prefix on an architecture page.

deploy/helm/platform/values.yaml
```diff
@@ -1288,6 +1294,7 @@ agentTemplates:
     description: "Default Claude Code agent"
     releaseNotesUrl: "https://code.claude.com/docs/en/whats-new"
     telemetry: true
+    harness: claude-code
```
dam-code-guardian, line 1324: 🟡 **Warning:** All four harness families ship ungated here, which answers both of #3511's Open Questions — the launch floor, and whether an unsupported harness is hidden or shown with a reason — unilaterally, while the PR body records Codex and Pi as "layout-tested in the bootstrap repos, not exercised in a live session". The closing keyword then closes #3511 with the launch-floor decision unrecorded and two of the four offered harnesses unproven end to end.
**Fix:** Treat an issue's Open Questions as deliverables — answer each in the PR body or the issue with the decision taken and the evidence for it before a closing keyword ships; a surface offered to users needs a live exercise or an explicit "unavailable, reason" state, never silent inclusion.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `3295770`

## PR #3569: feat: knowledge bases run on any harness, not only Claude Code
**Author:** Tomas2D | **Branch:** feat/kb-harness-choice → main | **Changes:** +307 −47 (19 files)

### Summary
The knowledge-base setup form now offers the live harness catalogue and the full provider list, each harness template declares its family in the Helm chart, and the server resolves that family and passes it to the chosen bootstrap through a template-specific environment variable.

### Findings
- 🟡 **Warning:** the permissive arm of `setupProviderPolicy` is now its unnamed fallback (`packages/ui/src/modules/sandboxes/lib/setup-policy.ts:14`)
  **Fix:** branch the closed union with `match().exhaustive()` and keep the catch-all restrictive, per `docs/guidelines/code/exhaustive-matching.md`.
- 🟡 **Warning:** `harness` is parsed as a free string, so a misspelt family degrades to "no family" with no signal (`packages/api-server-api/src/modules/templates/schemas.ts:41`)
  **Fix:** parse a closed value set with its enum schema at the boundary, and make a tolerated invalid value observable.
- 🟡 **Warning:** the Harness section has no error or empty branch, so a failed template query leaves Create disabled with no visible reason (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:118`)
  **Fix:** render loading, failed/empty and loaded for any query that gates a submit button.
- 🟡 **Warning:** a `sessionStorage` draft holding a `templateId` absent from the catalogue is a dead end the one-shot latch can never reconcile (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:51`)
  **Fix:** validate persisted server-owned ids against the fetched list on arrival, not only when the field is empty.
- 🟡 **Warning:** the catalogue `useMemo` plus preselect latch is a second copy of `coding-agent-setup-view.tsx:52-69`, already diverged (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:43`)
  **Fix:** move the derivation and its default selection into one shared hook both views call.
- 🟡 **Warning:** the harness grid and its skeleton geometry duplicate `image-section.tsx:39-53` (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:121`)
  **Fix:** extract one `HarnessGrid` that both `ImageSection` and this view render.
- 🟡 **Warning:** the agent-lifecycle page still says a knowledge base pins a harness image the user never sees, contradicting the page this PR rewrote (`docs/architecture/agent-lifecycle.md:57`)
  **Fix:** update every architecture page describing a flow the change moves, not only the changed subsystem's own page.
- 🟡 **Warning:** the rewritten knowledge-bases page transcribes enum members, an env-var transport and a literal command prefix, all under the guidelines' **Omit** list (`docs/architecture/knowledge-bases.md:12`)
  **Fix:** describe the contract semantically and link out to the contract package for field-level shape.
- 🟡 **Warning:** the issue's two Open Questions on launch floor and unsupported-harness presentation ship unanswered while Codex and Pi are offered but never live-exercised (`deploy/helm/platform/values.yaml:1297`)
  **Fix:** answer each Open Question with its decision and evidence before a closing keyword ships.
- 🟡 **Warning:** `/wiki-onboard` still sends one Claude-Code command on every harness, and the PR's own page concedes it reaches Codex and Bob as plain text with nothing surfacing that (`packages/ui/src/modules/knowledge-bases/hooks/use-knowledge-base-greeting.ts:6`)
  **Fix:** generalize every Claude-Code assumption on the path in the same change, or surface the harness as unsupported in the UI.
- ✅ **Looks good:** the `// TEST_SCENARIO:` lines the two test files add are a registered comment type (`scripts/strip-comments.mjs:35`) and match how 74 files on `main` document scenarios — no convention violation.

Two section findings were checked and not mirrored: the second `ReadTemplateSpec` declaration follows the shape `agents-service.ts:360` already restates, and `Last verified: 2026-09-02` does satisfy the "bumped whenever you edit the page" rule. Both stay in their sections as reported.

- 🟡 **Warning:** The agent-lifecycle page still states that Agent Kinds — knowledge bases included — "stamp a marker and pin a harness image the user never sees", which this PR makes false: the KB setup form now renders a Harness card grid over the live template catalogue and submits the user's `templateId`, and the KB page itself was rewritten to say the harness is the user's choice. The two architecture pages now contradict each other on who picks the image, and agent-lifecycle owns the create flow in depth, so a reader of the flow page gets the pre-PR system (`docs/architecture/agent-lifecycle.md:57`)
  **Fix:** When a change moves a choice into or out of a subsystem's create flow, update every architecture page that describes that flow in the same PR — not only the page of the subsystem that changed — and bump each page's `Last verified:`; the owning page of a concept and its cross-referencing pages must never disagree about it.
- 🟡 **Warning:** `Last verified:` on the changed page was bumped to 2026-09-02, but the PR's head commit is dated 2026-09-04, so the page's own verification stamp is already two days behind the edit that produced it and reads as stale to the next reader deciding whether to trust it (`docs/architecture/knowledge-bases.md:3`)
  **Fix:** Bump `Last verified:` to the date the PR lands, not the date the branch was first written; re-check the stamp on every rebase or amend that moves the head commit.
- 🟡 **Warning:** The rewritten Install Command and Create-flow paragraphs pull volatile, code-level shape onto the page: the harness-family value set is transcribed as the code's enum members (Claude Code, Codex, Pi, Bob), the transport is described as "a template-specific environment variable on the command", the declaration site is named as the Helm chart, and a third party's literal command prefix (`/prompts:`) is quoted. The documentation guidelines put exact enum members, Helm template tree, and code-level shape under **Omit** — each of these breaks when a family is added, an env var renamed, or the values file restructured (`docs/architecture/knowledge-bases.md:12`)
  **Fix:** Describe a contract semantically — what is exchanged and what each outcome means — and link out to the contract package or chart as the field-level source of truth; never enumerate an enum's members, name an env var, or quote a literal prefix on an architecture page.

### Issue Fit
Linked issue: #3511 "UI - Knowledge base agents can only run on Claude Code" (closing keyword)

- 🟡 **Warning:** Both of the issue's Open Questions — "Do all four harnesses need to support knowledge bases at launch, or do we start with a subset?" and "If a harness can't yet support knowledge base setup, do we hide it for this flow or show it as unavailable with a reason?" — are answered unilaterally by shipping all four families ungated, while the PR's own Verified-live section records Codex and Pi as "layout-tested in the bootstrap repos, not exercised in a live session". A closing keyword closes #3511 with the launch-floor decision unrecorded and two of the four offered harnesses unproven end to end (`deploy/helm/platform/values.yaml:1297`).
  **Fix:** An issue's Open Questions are deliverables — answer each one in the PR body (or the issue) with the decision taken and the evidence for it before a closing keyword ships; a surface offered to users needs a live exercise or an explicit "unavailable, reason" state, never silent inclusion.
- 🟡 **Warning:** The issue's Scope bullet requires that "whatever knowledge base setup currently assumes about Claude Code should work, degrade gracefully, or be clearly surfaced as unsupported on a given harness", but the `/wiki-onboard` greeting still sends one literal Claude-Code command on every harness with no harness awareness, and the PR's own architecture-page edit concedes Codex exposes installed prompts under a `/prompts:` prefix and Bob has no command files. On those harnesses the platform's first turn reaches the model as unexpanded plain text and nothing in the UI surfaces that the greeting degraded (`packages/ui/src/modules/knowledge-bases/hooks/use-knowledge-base-greeting.ts:6`).
  **Fix:** When a feature is generalized across harnesses, every Claude-Code-specific assumption on the path — install command *and* greeting/command invocation — is generalized in the same change or surfaced to the user as unsupported; documenting a silent degradation in an architecture page is not surfacing it.
- 🟢 **Suggestion:** The issue asks for "the harness and model-provider selection a user already gets for code agents… the same way they can for a code agent", but the KB form deliberately omits the custom private-registry image entry that the coding-agent flow offers, so harness parity is partial (the PR body declares this as a possible follow-up) (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:117`).

- 🟡 **Warning:** `setupProviderPolicy` branches on the closed union `SetupFlow` (`"coding-agent" | "experiment" | "knowledge-base"`) with a single `if` plus an unnamed fallback, so only `experiment` is named and the other two members are invisible at the branch point. The fallback is also the permissive arm (no `allow` list at all), so a fourth `SetupFlow` member added later silently gets the every-provider policy instead of a compile error — and this PR itself relies on that arm for the knowledge-base flow. (`packages/ui/src/modules/sandboxes/lib/setup-policy.ts:14`)
  **Fix:** Branch on a closed string-literal union with `match(...)` from `ts-pattern`, one `.with()` per member closed by `.exhaustive()` (docs/guidelines/code/exhaustive-matching.md — the rule the same PR follows in `install-command.ts`), so adding a union member breaks the build rather than falling into a permissive default; never let the catch-all arm of a policy function be the least restrictive one.
- 🟡 **Warning:** `templateSpecSchema.harness` is typed `z.string().optional()` even though the same package now defines the closed enum `kbHarnessFamilySchema` for exactly these values, so a misspelled `harness:` in `deploy/helm/platform/values.yaml` (e.g. `claude_code`) validates fine, is then dropped by `parseKbHarnessFamily`'s `safeParse` without any log or error, and the KB installs with no harness wiring and no diagnostic anywhere. (`packages/api-server-api/src/modules/templates/schemas.ts:41`)
  **Fix:** Validate a field whose value set is closed with the enum schema at the boundary that first parses it, and when a value must be tolerated for forward compatibility make the degradation observable (log or surface the rejected value) rather than mapping "invalid" and "absent" onto the same silent `undefined`.
- 🟡 **Warning:** `ReadTemplateSpec` is re-declared in the knowledge-bases service with a body structurally identical to the already-exported `ReadTemplateSpec` in `packages/api-server/src/modules/templates/index.ts` (re-exported from `templates/compose.ts:5`), and `context.ts` feeds the same `readSpec` value to both. Structural typing hides the duplication today, so a change to the template module's port signature will compile against a now-stale second copy. (`packages/api-server/src/modules/knowledge-bases/services/knowledge-bases-service.ts:17`)
  **Fix:** A port type belongs to the module that owns the capability — import the existing exported alias in consumers instead of restating its shape, so the signature has exactly one definition to change.
- 🟢 **Suggestion:** `harness?: string` is added to `TemplateView` and populated by the templates router's `toView`, but no UI code reads `template.harness` — the only `.harness` references in `packages/ui` belong to the unrelated `sandboxSubtitleParts`. The field is currently dead API surface travelling to the browser. (`packages/ui/src/types.ts:94`)
- 🟢 **Suggestion:** `bootstrapCommand` takes the env-var name as a bare `string` while the whole feature depends on the exact template-id → env-var-name pairing (`llm-wiki`→`LLM_WIKI_HARNESS`, `plain-wiki`→`PLAIN_WIKI_HARNESS`), which the tests re-state as a separate literal map; a typo in either call site is invisible to the compiler. Consider a single `Record<KnowledgeBaseTemplateId, { url: string; envName: string }>` that both the builder and the test read. (`packages/api-server/src/modules/knowledge-bases/domain/install-command.ts:10`)

- 🟡 **Warning:** The Harness section has no error or empty state: `isLoading` is the only branch, so when `useTemplates()` fails (or returns no `category: "harness"` template) the section renders an empty `CardGrid` and `canCreate` stays false forever via `harnesses.some((t) => t.id === form.templateId)`. The user sees a blank section under a "Harness" label and a permanently disabled "Create knowledge base agent" button with nothing on the page explaining why — and unlike the coding-agent flow there is no custom-image fallback path to complete setup. (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:118`)
  **Fix:** Any query whose result gates a submit button must render three states, not two — loading, failed/empty, and loaded: branch on `isError`/`isLoading` and render an explicit message plus a retry for the non-happy paths, so a disabled action always has a visible reason next to it.
- 🟡 **Warning:** A restored draft holding a `templateId` that is not in the fetched catalogue is a silent dead end. The preselect effect latches `preselected.current = true` on the first non-empty `harnesses` and returns early whenever `form.templateId !== null`, so a `templateId` rehydrated from `sessionStorage` (the `vm-sandboxes` flag since turned off, or the template retired server-side) is never cleared or re-selected: no card shows as selected, `canCreate` is false, and no reconciliation can ever run again. (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:51`)
  **Fix:** When persisted UI state references server-owned ids, validate it against the server list on arrival rather than only defaulting when it is absent — drop or replace a selection that is no longer in the fetched set, and keep that reconciliation live instead of behind a one-shot ref latch.
- 🟡 **Warning:** The harness catalogue and its preselect effect are a second verbatim copy of `coding-agent-setup-view.tsx:52-69` — same `useMemo(imageCatalogue(templates ?? [], { vmFeatureEnabled: flags?.["vm-sandboxes"] ?? false }).harnesses)`, same `useRef` latch, same `KINDED_HARNESS_TEMPLATE_ID` default — and the two copies have already diverged (the coding-agent one also guards on `form.customImage`). Both views must now be edited in lockstep for any catalogue or default-selection change. (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:43`)
  **Fix:** Server-derived catalogue plus its default-selection behavior belongs in one shared hook (e.g. `useHarnessCatalogue()` beside `image-catalogue.ts`) that both setup views call; a view never re-implements another view's data derivation, even when the second copy needs a slightly different guard — parameterize the hook instead.
- 🟡 **Warning:** The new harness grid duplicates markup that `ImageSection` already owns: the same `isLoading ? <CardList><ListSkeleton rows={2} rowHeight={156} /></CardList> : <CardGrid>{harnesses.map(...HarnessCard...)}</CardGrid>` block exists at `packages/ui/src/modules/sandboxes/components/setup/image-section.tsx:39-53`. The skeleton geometry (`rows={2}`, `rowHeight={156}`) is now a magic pair repeated in two files, so the two harness pickers will drift visually. (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:118`)
  **Fix:** Extract the shared presentation — a `HarnessGrid` component carrying its own skeleton geometry — and have both `ImageSection` and this view render it; a view composes existing setup components rather than copying their JSX with the custom-image half removed.
- 🟢 **Suggestion:** Validation for this flow is inlined in the view as a five-clause `canCreate` boolean plus a redundant re-check in `create` (`if (!canCreate || form.templateId === null) return;` — the null test exists only because the boolean cannot narrow the type), while the sibling coding-agent view extracted exactly this concern to `agents/lib/create-agent-input.ts` (`isCodingAgentSetupComplete` / `buildCodingAgentSetupInput`). (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:60`)
- 🟢 **Suggestion:** Feature flags are read as a bare indexed string literal, `flags?.["vm-sandboxes"] ?? false`, duplicated verbatim in the coding-agent view — a typo in either key silently reads as "disabled" with no compile error. A named constant or typed flag accessor would make the misspelling a type error. (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:46`)
- 🟢 **Suggestion:** `create()` was edited in this PR but keeps its bare `catch {}`, so a failed `mutateAsync` leaves the form in place with no inline feedback from this view; the touched-code migration rule makes this the moment to surface the mutation error (or drop the try/catch and let the mutation's error state render). (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:85`)

### Verdict
COMMENT — the feature works end to end, but five warnings stand: a permissive default policy arm, an unvalidated harness field that fails silently, a setup form that can lock itself out, and two architecture pages left inconsistent.


---

packages/ui/src/modules/sandboxes/lib/setup-policy.ts
Tomas2D: Done in 00a9f3e: `setupProviderPolicy` is now a `Record<SetupFlow, SetupProviderPolicy>` lookup — every member is named, adding a flow is a compile error, and there is no permissive catch-all arm. (`ts-pattern` is not a UI dependency; the exhaustive record gives the same guarantee.)

packages/api-server-api/src/modules/templates/schemas.ts
Tomas2D: Done in 00a9f3e: the vocabulary moved to the templates contract as `harnessFamilySchema` (the templates module owns it; knowledge-bases only consumes it) and `templateSpecSchema.harness` is that enum. A misspelt value now fails template validation at api-server boot and the entry is skipped with the existing stderr diagnostic, instead of degrading to "no family" silently. `parseKbHarnessFamily` is gone; the values.yaml comment documents the allowed set and the failure mode.

packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx
Tomas2D: Done in 00a9f3e: the grid is now a shared `HarnessGrid` that renders loading, failed (with Retry via the query's refetch) and empty states explicitly, so a disabled Create always has a visible reason next to it. `ImageSection` renders the same component.

packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx
Tomas2D: Done in 00a9f3e: selection is reconciled live in `useHarnessCatalogue` via the pure `reconcileHarnessSelection` (unit-tested): a persisted id not in the fetched catalogue is replaced by the default (or cleared when a custom image legitimately leaves it empty), an empty selection gets the default, a valid one is untouched. No one-shot latch.

packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx
Tomas2D: Done in 00a9f3e: catalogue derivation + default selection live in one shared `useHarnessCatalogue` hook (parameterised with `allowNone` for the coding-agent custom-image case); both setup views call it and the duplicated `useMemo`/`useRef` blocks are gone.

packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx
Tomas2D: Done in 00a9f3e: extracted `HarnessGrid` (owns the skeleton geometry and the loading/error/empty states); `ImageSection` and the KB view both render it.

docs/architecture/knowledge-bases.md
Tomas2D: Done in 00a9f3e: the page now describes the contract semantically — a closed harness-family vocabulary owned by the templates contract (linked), validated at load, handed to the installer — with no enum members, env-var name or third-party command prefix. The greeting paragraph says the command is issued in the form the harness expands. `agent-lifecycle.md` no longer says a KB pins an unseen image; both pages' `Last verified` bumped to today.

deploy/helm/platform/values.yaml
Tomas2D: Answered in the PR body ("Open Questions from #3511") with the decision and evidence: all four families ship; Bob is verified live for both templates, Claude Code is the unchanged path, Codex and Pi are layout-tested in the bootstrap repos (the platform's only contribution there is the family name). No harness is unsupported, so there is nothing to hide or grey out; a future family that a bootstrap cannot wire falls back to the installer's own detection. Also generalised the last Claude-Code assumption on the path: the greeting is now issued in the form the harness expands (`/prompts:…` on Codex).

Tomas2D:
Addressed the Code Guardian review in 00a9f3e (rebased onto main first; one docs conflict resolved). Inline threads are answered and resolved individually; the non-inline findings:

- **agent-lifecycle.md still said a KB pins an unseen image** — fixed: experiments pin, a KB's harness is the user's pick (cross-linked to the KB page); `Last verified` bumped on both pages.
- **`/wiki-onboard` sent as one Claude-Code command on every harness** — the greeting is now issued in the form the harness expands (`onboardCommandFor(harness)`: `/prompts:wiki-onboard` on Codex, `/wiki-onboard` elsewhere), resolved from the agent's template family. Bob auto-converts the installed command into a skill, verified live.
- **`ReadTemplateSpec` re-declared** — the knowledge-bases module now imports the templates module's exported port type.
- **`TemplateView.harness` dead on the client** — it now drives the greeting command; typed as the contract enum.
- **`bootstrapCommand` env-var name as a bare string** — replaced by `KB_TEMPLATE_BOOTSTRAPS: Record<KnowledgeBaseTemplateId, { url; harnessEnv }>`, which the test reads too.
- **`canCreate` inlined** — extracted to `knowledge-bases/lib/create-knowledge-base-input.ts` (`isKnowledgeBaseSetupComplete` / `buildKnowledgeBaseCreateInput`, unit-tested), mirroring the coding-agent lib.
- **Flag read as bare literal** — kept: `FeatureFlags` is `Record<FeatureId, boolean>`, so a misspelt key is already a type error; the read now lives in one place (the shared hook).
- **bare `catch {}` in `create()`** — kept: the mutation's `errorToast` meta surfaces the failure; same shape as the coding-agent view.

packages/ui/src/modules/sessions/views/chat-view.tsx
```diff
@@ -212,10 +213,12 @@ export function ChatView() {
 
   const view = useStore((s) => s.view);
   const chatIdle = !sessionId && messages.length === 0;
+  const { data: templates } = useTemplates();
   useKnowledgeBaseGreeting({
     agentId: selectedAgent,
     active: view === "knowledge-base-chat",
     idle: chatIdle,
+    harness: templates?.find((t) => t.id === agentView?.templateId)?.harness,
```
dam-code-guardian: 🔴 **Critical:** The knowledge-base greeting fires with `harness: undefined` whenever the `templates` query has not resolved by the time it arms. `useAgentGreeting` gates on `sessions !== undefined` and `operable` only, so a fresh KB on a Codex harness sends the plain `/wiki-onboard` instead of `/prompts:wiki-onboard` — the defect this PR exists to fix — and `greetedForAgentRef` makes the miss permanent for that agent, silently by design.
  **Fix:** When a latch-once side effect derives its payload from asynchronously loaded server state, that state belongs in the arming gate — never commit the effect on a half-loaded input.

packages/ui/src/modules/sandboxes/hooks/use-harness-catalogue.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const harnesses = useMemo(
+    () =>
+      imageCatalogue(templates ?? [], {
+        vmFeatureEnabled: flags?.["vm-sandboxes"] ?? false,
+      }).harnesses,
+    [templates, flags],
+  );
+
+  useEffect(() => {
```
dam-code-guardian, line 36: 🟡 **Warning:** The reconcile effect gates on `templates === undefined` but not on the separately loaded `flags`, while `harnesses` is filtered by `flags?.["vm-sandboxes"] ?? false`. If templates resolve first, a persisted `claude-code-vm` selection is treated as stale and rewritten to `claude-code`, losing the user's choice while the VM feature is in fact enabled.
  **Fix:** Gate a reconcile effect on every query that shapes the set it compares against, never a subset, before rewriting persisted user state.

packages/ui/src/modules/sandboxes/components/setup/harness-grid.tsx
```diff
@@ -0,0 +1,66 @@
...
+  onPick,
+}: Props) {
+  if (loading) {
+    return (
+      <CardList>
+        <ListSkeleton rows={2} rowHeight={156} />
+      </CardList>
+    );
+  }
+  if (error || harnesses.length === 0) {
```
dam-code-guardian: 🟡 **Warning:** This branch collapses "genuinely empty" and "the query failed" into `error || harnesses.length === 0`. On a real failure `harnesses` is also empty, so the distinction rests entirely on `error`, which is optional and defaults to `false` — a caller omitting it renders a load failure as the reassuring "No harness is available on this installation."
  **Fix:** Make the error input required and branch loading / error / empty / data as distinct states, so a caller cannot silently misreport a failure as an empty result.

packages/ui/src/modules/knowledge-bases/lib/onboard-command.ts
```diff
@@ -0,0 +1,9 @@
+import type { HarnessFamily } from "api-server-api";
+
+const ONBOARD_COMMAND_NAME = "wiki-onboard";
+
+export function onboardCommandFor(harness: HarnessFamily | undefined): string {
+  return harness === "codex"
```
dam-code-guardian: 🟡 **Warning:** This branches on `harness === "codex"` against the closed union `HarnessFamily` (`claude-code | codex | pi | bob`). A family added later that also namespaces installed commands compiles clean and degrades to the plain `/wiki-onboard`, against `docs/guidelines/code/exhaustive-matching.md`. Also reported by typescript-engineering as the rule's second copy across the client/server boundary.
  **Fix:** Name every member of a closed union at a branch point — a `Record<HarnessFamily, string>` here, since `ts-pattern` is not a `packages/ui` dependency — so a new member fails to compile.

docs/architecture/knowledge-bases.md
```diff
@@ -9,23 +9,23 @@ A **Knowledge Base** is an Agent that builds and maintains a body of knowledge t
...
 
 The greeting cannot race the bootstrap: a pending Install Command holds the agent in the same preparing-workspace state a pending workspace seed does, and the agent reports running — the greeting's gate — only once the install has settled. The same gate covers the pod not being up yet, a wake that arrives late, and a failed install retrying under the runtime channel's per-event attempt budget ([runtime-delivery](runtime-delivery.md)). The hold is bounded, not absolute: an install that exhausts its budget, or lands on a runtime that cannot run it, is stamped settled-with-error, so the agent surfaces as running again and the greeting fires against a KB whose `/wiki-onboard` may be missing — that greeting breaks silently by design, rather than the agent staying unusable until the install event's TTL.
```
dam-code-guardian: 🟡 **Warning:** This paragraph still names the literal `/wiki-onboard`, but the PR made the onboarding command harness-dependent — `onboardCommandFor` emits `/prompts:wiki-onboard` on codex — and the same page was de-literalized two paragraphs earlier. The page now contradicts itself and states a command form that is wrong for a codex knowledge base.
  **Fix:** When a change turns a fixed literal into a resolved form, replace *every* occurrence of that literal on the page, not only the ones in the paragraph being rewritten.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `00a9f3e`

## PR #3569: feat: knowledge bases run on any harness, not only Claude Code
**Author:** Tomas2D | **Branch:** feat/kb-harness-choice → main | **Changes:** +655 −118 (30 files)

### Summary
Knowledge-base creation now offers the same harness catalogue coding agents pick from, and the chosen template's harness family reaches the KB bootstrap and the onboarding command. Nine of the ten findings from the last review are resolved; the harness-aware greeting it introduces can still fire with no harness resolved.

_12 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: 3295770 (2026-09-04T14:37:19Z) — verdict COMMENT — unreachable, reviewed the whole PR
- ✅ **Fixed:** the permissive arm of setupProviderPolicy is now its unnamed fallback (`packages/ui/src/modules/sandboxes/lib/setup-policy.ts:14`)
- ✅ **Fixed:** harness is parsed as a free string, so a misspelt family degrades to no family with no signal (`packages/api-server-api/src/modules/templates/schemas.ts:41`)
- ✅ **Fixed:** the Harness section has no error or empty branch, so a failed template query leaves Create disabled with no visible reason (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:118`)
- ✅ **Fixed:** a sessionStorage draft holding a templateId absent from the catalogue is a dead end the one-shot latch can never reconcile (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:51`)
- ✅ **Fixed:** the catalogue useMemo plus preselect latch is a second, already diverged copy of coding-agent-setup-view.tsx (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:43`)
- ✅ **Fixed:** the harness grid and its skeleton geometry duplicate image-section.tsx (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:121`)
- ✅ **Fixed:** the agent-lifecycle page still says a knowledge base pins a harness image the user never sees (`docs/architecture/agent-lifecycle.md:57`)
- ✅ **Fixed:** the rewritten page transcribes enum members, an env-var transport and a literal command prefix, all under the guidelines' Omit list (`docs/architecture/knowledge-bases.md:12`)
- ✅ **Fixed:** /wiki-onboard still sends one Claude-Code command on every harness and reaches Codex and Bob as plain text with nothing surfacing that (`packages/ui/src/modules/knowledge-bases/hooks/use-knowledge-base-greeting.ts:6`)
- 🔁 **Still present:** Codex and Pi are offered but never live–exercised (`deploy/helm/platform/values.yaml:1297`)
- 🆕 **New:** greeting latches before templates resolve, harness undefined (`packages/ui/src/modules/sessions/views/chat-view.tsx:221`)
- 🆕 **New:** reconcile waits on templates but not on flags (`packages/ui/src/modules/sandboxes/hooks/use-harness-catalogue.ts:35`)
- 🆕 **New:** load failure renders as the reassuring empty state (`packages/ui/src/modules/sandboxes/components/setup/harness-grid.tsx:33`)
- 🆕 **New:** branch on HarnessFamily names one member of four (`packages/ui/src/modules/knowledge-bases/lib/onboard-command.ts:6`)
- 🆕 **New:** page keeps the literal command it de–literalized above (`docs/architecture/knowledge-bases.md:30`)
- 🆕 **New:** glossary still records the pinned–harness, one–command model (`docs/ubiquitous-language.md:227`)
- 🆕 **New:** KB form also offers the experimental VM harness (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:44`)

### Findings
- 🔴 **Critical:** greeting latches before templates resolve, harness undefined (`packages/ui/src/modules/sessions/views/chat-view.tsx:221`)
- 🟡 **Warning:** reconcile waits on templates but not on flags (`packages/ui/src/modules/sandboxes/hooks/use-harness-catalogue.ts:35`)
- 🟡 **Warning:** load failure renders as the reassuring empty state (`packages/ui/src/modules/sandboxes/components/setup/harness-grid.tsx:33`)
- 🟡 **Warning:** branch on HarnessFamily names one member of four (`packages/ui/src/modules/knowledge-bases/lib/onboard-command.ts:6`)
- 🟡 **Warning:** page keeps the literal command it de–literalized above (`docs/architecture/knowledge-bases.md:30`)
- 🟡 **Warning:** The **KB Template** glossary entry still records the model this PR replaces — the harness image as what "v1 pins and hides", and a bootstrap that "must install a `/wiki-onboard` command". Both are now false: the harness is the user's pick and the command form depends on the family. Reported by doc-drift and issue-fit (`docs/ubiquitous-language.md:227`)
  **Fix:** Update every canonical page that states a fact the change moves, glossary included — an architecture-page edit does not answer the vocabulary entry repeating the same claim.
- 🟢 **Suggestion:** The KB form takes the whole `category === "harness"` catalogue, so with `vm-sandboxes` enabled it also offers the experimental `claude-code-vm` template — a fifth harness beyond the four the linked issue names (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:44`)

- 🟡 **Warning:** The greeting-race paragraph still names the literal `/wiki-onboard`, but this PR made the onboarding command harness-dependent — `onboardCommandFor` emits `/prompts:wiki-onboard` on codex — and the same page was de-literalized two paragraphs earlier. The page now contradicts itself and states a command form that is wrong for a codex knowledge base (`docs/architecture/knowledge-bases.md:30`)
  **Fix:** When a change turns a fixed literal into a resolved form, replace *every* occurrence of that literal on the page, not only the ones in the paragraph being rewritten.

### Issue Fit
Linked issue: #3511 "UI - Knowledge base agents can only run on Claude Code" (`Closes #3511`).

- 🟡 **Warning:** The issue's Open Question 1 is answered "all four ship", but the platform's contribution on Codex and Pi — the family name handed to the bootstrap and the command form the greeting issues — is unit-tested only; whether Pi expands installed commands under a plain `/` prefix is unverified against a running Pi. Carried from the last review (`deploy/helm/platform/values.yaml:1297`)
  **Fix:** Give each harness the flow offers one live end-to-end exercise, or hold the unexercised ones back from the picker.

_1 suggestion(s) dropped under the 🟢 budget._

The skill reported five blocking findings. Each is repriced below and none is carried:

- The `ReadTemplateSpec` port crossing from the templates module is the established pattern here, not a boundary breach — `experiments/compose.ts` and `kb-shares/compose.ts` already import sibling module ports, and no repository rule forbids it.
- The installer URLs in `domain/install-command.ts` were already in that file before this PR; the change adds the env-var name beside them. Pre-existing layering, not a new defect.
- `TemplateView` mirroring the router's projection, the client-side completeness rule and `egressPreset: "trusted"` are all pre-existing shapes this PR extends — `egressPreset: "trusted"` is on `main` at the same call site.
- The split harness rule is the same defect as the exhaustiveness finding above, carried once at `onboard-command.ts:6`.

- 🔴 **Critical:** The knowledge-base greeting fires with `harness: undefined` whenever the `templates` query has not resolved by the time it arms. `useAgentGreeting` gates on `sessions !== undefined` and `operable` only, so a fresh KB on a Codex harness sends the plain `/wiki-onboard` instead of `/prompts:wiki-onboard` — the defect this PR exists to fix — and `greetedForAgentRef` makes the miss permanent for that agent, silently by design (`packages/ui/src/modules/sessions/views/chat-view.tsx:221`)
  **Fix:** When a latch-once side effect derives its payload from asynchronously loaded server state, that state belongs in the arming gate — never commit the effect on a half-loaded input.
- 🟡 **Warning:** `useHarnessCatalogue`'s reconcile effect gates on `templates === undefined` but not on the separately loaded `flags`, while `harnesses` is filtered by `flags?.["vm-sandboxes"] ?? false`. If templates resolve first, a persisted `claude-code-vm` selection is treated as stale and rewritten to `claude-code`, losing the user's choice while the VM feature is in fact enabled (`packages/ui/src/modules/sandboxes/hooks/use-harness-catalogue.ts:35`)
  **Fix:** Gate a reconcile effect on every query that shapes the set it compares against, never a subset, before rewriting persisted user state.
- 🟡 **Warning:** `HarnessGrid` collapses "genuinely empty" and "the query failed" into `error || harnesses.length === 0`. On a real failure `harnesses` is also empty, so the distinction rests entirely on `error`, which is optional and defaults to `false` — a caller omitting it renders a load failure as the reassuring "No harness is available on this installation." (`packages/ui/src/modules/sandboxes/components/setup/harness-grid.tsx:33`)
  **Fix:** Make the error input required and branch loading / error / empty / data as distinct states, so a caller cannot silently misreport a failure as an empty result.
- 🟡 **Warning:** `onboardCommandFor` branches on `harness === "codex"` against the closed union `HarnessFamily` (`claude-code | codex | pi | bob`). A family added later that also namespaces installed commands compiles clean and degrades to the plain `/wiki-onboard`, against `docs/guidelines/code/exhaustive-matching.md`. Also reported by typescript-engineering as the rule's second copy (`packages/ui/src/modules/knowledge-bases/lib/onboard-command.ts:6`)
  **Fix:** Name every member of a closed union at a branch point — a `Record<HarnessFamily, string>` here, since `ts-pattern` is not a `packages/ui` dependency — so a new member fails to compile.

Two further findings are not carried: the `refetch` identity has no memo boundary to defeat (`HarnessCard` and `CardGrid` are not memoized), and the inline template lookup in `chat-view.tsx` is a readability point on a pre-existing pattern.

_5 suggestion(s) dropped under the 🟢 budget._

### Verdict
REQUEST_CHANGES — nine of ten prior findings are resolved, but the harness-aware greeting this PR introduces can latch on an unresolved harness and send the wrong command form, alongside five warnings.


---

packages/ui/src/modules/sessions/views/chat-view.tsx
Tomas2D: Fixed: the greeting now takes a `ResolvedHarness` (`{ ready: false }` until the templates query has settled, `{ ready: true, harness }` after) and feeds `ready` into `useAgentGreeting`'s existing `setupReady` gate, so the latch-once greeting cannot arm on an unresolved harness. An agent with no template (custom image) is ready immediately with no family. Resolution lives in `resolveAgentHarness` (unit-tested for all three states).

packages/ui/src/modules/sandboxes/hooks/use-harness-catalogue.ts
Tomas2D: Fixed: the reconcile effect now waits for `catalogueSettled = templates !== undefined && flagsSettled` (`isFetched` on the features query — settled on success or failure, so an unreachable flags endpoint does not block reconciliation forever). A persisted `claude-code-vm` selection is no longer rewritten while the VM flag is still loading.

packages/ui/src/modules/sandboxes/components/setup/harness-grid.tsx
Tomas2D: Fixed: `error` and `onRetry` are required props and the grid branches loading / error / empty / data as four distinct states; a failure can no longer render as the empty-catalogue message. `ImageSection` mirrors the required props.

packages/ui/src/modules/knowledge-bases/lib/onboard-command.ts
Tomas2D: Fixed: `onboardCommandFor` reads a `Record<HarnessFamily, string>` naming every member; adding a family without deciding its command form is now a compile error. `undefined` (no known family) keeps the plain form.

docs/architecture/knowledge-bases.md
Tomas2D: Fixed: the remaining literal is gone ("a KB whose onboarding command may be missing") and the greeting paragraph states that the command form is resolved from the template family and that the greeting waits for that resolution. Also updated the glossary (`docs/ubiquitous-language.md`): the KB Template entry no longer describes a pinned harness or a single literal command, and a Harness Family term is added.

Tomas2D:
Second Code Guardian round (`51e9e83`): all five inline threads fixed and resolved individually. The non-inline findings:

- **Codex and Pi offered but never live-exercised (still)** — now exercised on the local cluster through the real create path for both templates: the bootstrap ran with the family name, and the manual/commands/skill landed where each harness reads them (details in the PR body's Open Questions table). Bob and Claude Code were already covered.
- **Glossary still recorded the pinned-harness, one-command model** — `docs/ubiquitous-language.md`: the KB Template entry is rewritten and a Harness Family term added.
- **KB form also offers the experimental VM harness** — intentional: it uses the same catalogue and `vm-sandboxes` gate as the coding-agent form, so the two flows cannot drift on what is offered.

packages/ui/src/modules/knowledge-bases/lib/resolve-agent-harness.ts
```diff
@@ -0,0 +1,14 @@
+import type { TemplateView } from "../../../types.js";
+import type { ResolvedHarness } from "../hooks/use-knowledge-base-greeting.js";
+
+export function resolveAgentHarness(
+  templateId: string | null,
+  templates: { data: TemplateView[] | undefined; isFetched: boolean },
+): ResolvedHarness {
+  if (templateId === null) return { ready: true, harness: undefined };
+  if (!templates.isFetched) return { ready: false };
```
dam-code-guardian: 🔴 **Critical:** `isFetched` is `dataUpdateCount + errorUpdateCount > 0` in query-core 5.99.0, so it is true after a *failed* `templates.list` fetch, where `data` is `undefined`. This returns `{ ready: true, harness: undefined }`, `useKnowledgeBaseGreeting` passes `setupReady: true`, and `useAgentGreeting` latches `greetedForAgentRef` and sends the plain `/wiki-onboard` to a Codex knowledge base — permanently, since the latch is once per agent and a later successful fetch never corrects it. The same gate shape is at `use-harness-catalogue.ts:24` (`isFetched: flagsSettled`), where a failed flags query reads as settled and reconciles a persisted `claude-code-vm` selection away. Reported by react-ui-engineering.
  **Fix:** Gate on a settled *successful* query (`data !== undefined` / `isSuccess`) and report not-ready on error — never `isFetched`, at either gate.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `51e9e83`

## PR #3569: feat: knowledge bases run on any harness, not only Claude Code
**Author:** Tomas2D | **Branch:** feat/kb-harness-choice → main | **Changes:** +726 −119 (32 files)

### Summary
Every round-2 finding is answered at its cause — the greeting now waits on a resolved harness, the reconcile effect waits on both queries, `HarnessGrid` branches loading / error / empty / data, `onboardCommandFor` is an exhaustive `Record<HarnessFamily, string>`, and both docs pages are corrected. One path of the greeting invariant is still open, and it is a round-2 miss of mine: the new readiness gate accepts a *failed* templates fetch as resolved.

_(Suppressed 1 finding(s) per PR-local overrides: Codex/Pi live-exercise.)_

_19 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: 00a9f3e (2026-09-07T11:37:46Z) — verdict REQUEST_CHANGES — unreachable, reviewed the whole PR
- ✅ **Fixed:** greeting latches before templates resolve, harness undefined (`packages/ui/src/modules/sessions/views/chat-view.tsx:221`)
- ✅ **Fixed:** reconcile waits on templates but not on flags (`packages/ui/src/modules/sandboxes/hooks/use-harness-catalogue.ts:35`)
- ✅ **Fixed:** load failure renders as the reassuring empty state (`packages/ui/src/modules/sandboxes/components/setup/harness-grid.tsx:33`)
- ✅ **Fixed:** branch on HarnessFamily names one member of four (`packages/ui/src/modules/knowledge-bases/lib/onboard-command.ts:6`)
- ✅ **Fixed:** page keeps the literal command it de–literalized above (`docs/architecture/knowledge-bases.md:30`)
- ✅ **Fixed:** glossary still records the pinned–harness, one–command model (`docs/ubiquitous-language.md:227`)
- ✅ **Settled:** Codex and Pi live-exercise — answered in the PR body and recorded as a PR-local override; not raised again (`deploy/helm/platform/values.yaml:1297`)
- 🔁 **Still present:** KB form also offers the experimental VM harness (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:44`)
- 🆕 **New:** isFetched is true after a failed fetch, at both new gates (`packages/ui/src/modules/knowledge-bases/lib/resolve-agent-harness.ts:9`, `packages/ui/src/modules/sandboxes/hooks/use-harness-catalogue.ts:24`)
- 🆕 **New:** shared setup machinery documented only on the KB page (`docs/architecture/agent-lifecycle.md:57`)

### Findings
- 🔴 **Critical:** isFetched is true after a failed fetch, at both new gates (`packages/ui/src/modules/knowledge-bases/lib/resolve-agent-harness.ts:9`, also `packages/ui/src/modules/sandboxes/hooks/use-harness-catalogue.ts:24`)
  **Fix:** Gate on a settled *successful* query (`data !== undefined` / `isSuccess`) and report not-ready on error — never `isFetched`, at either gate.
- 🟢 **Suggestion:** The new `useHarnessCatalogue` + `HarnessGrid` machinery changes the **coding-agent** create flow too — a retry affordance, an empty-catalogue notice, and reconciliation of a persisted selection that left the catalogue — but those behaviours are recorded only as knowledge-base properties, and the same is true of the `harness` field the Template spec gains. The page that owns the general create path and the Template model says neither. Reported by doc-drift (`docs/architecture/agent-lifecycle.md:57`)

Both blocking items are folded into the 🟢 above — a docs page that under-covers a shared change is one finding per round, and neither statement is contradicted by the code.

### Issue Fit
Linked issue: #3511 "UI - Knowledge base agents can only run on Claude Code" (`Closes #3511`).

The one blocking item is the Codex/Pi live-exercise question, settled by the PR body and recorded as a PR-local override — not re-raised.

Three blocking findings, none carried:

- The `ReadTemplateSpec` port import from the templates module is this repo's established pattern, not a new breach — `agents`, `connections`, `invocations` and `experiments` services all import sibling module ports on `main`.
- The `lib/` → `hooks/` type-only back-edge for `ResolvedHarness` is real and worth tidying, but layering and placement on a working diff are 🟢 here; `sandboxes/lib/setup-policy.ts` has the same edge on `main`.
- `useKnowledgeBaseGreeting` taking a resolved harness rather than resolving it is a design choice with no defect behind it at this HEAD.

- The `isFetched` finding is carried as the 🔴 above, with the `useFeatures` gate merged into it as the second site — the skill reported them separately.
- The remaining three are not carried: the derived-default-via-effect, the Retry button's pending state and the `chat-view.tsx` query placement are all the mechanism of the fixes this round asked for, and a fix's own new surface is not a fresh change request.

### Verdict
REQUEST_CHANGES — every round-2 finding is genuinely closed at its cause; one path of the greeting invariant I should have enumerated in round 2 is still open, and a failed templates fetch permanently sends the wrong onboarding command form.


---

packages/ui/src/modules/knowledge-bases/lib/resolve-agent-harness.ts
Tomas2D: Confirmed and fixed in 40084facd. Both gates now key on `data !== undefined` — a settled *successful* query — instead of `isFetched`. `resolveAgentHarness` takes only `{ data }` and reports not-ready while data is absent, so a failed templates fetch never arms the greeting; TanStack's retry/refetch delivers data later and the greeting arms then. `useHarnessCatalogue` reconciles only once both templates and flags have data, so a failed flags query can no longer rewrite a persisted `claude-code-vm` pick. Test renamed to cover the failed-fetch case; KB architecture page notes that an unsucceeded lookup keeps the greeting waiting.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `554c22a`

## PR #3569: feat: knowledge bases run on any harness, not only Claude Code
**Author:** Tomas2D | **Branch:** feat/kb-harness-choice → main | **Changes:** +732 −128 (32 files)

### Summary
The round closes the blocking finding at both gates and adds the regression test for the helper. Only suggestions remain. _9 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: 51e9e83 (2026-09-07T12:55:42Z) — verdict REQUEST_CHANGES — unreachable, reviewed the whole PR

- ✅ **Fixed:** isFetched is true after a failed fetch, at both new gates (`packages/ui/src/modules/knowledge-bases/lib/resolve-agent-harness.ts:9`)
- 🔁 **Still present:** shared setup machinery documented only on the KB page (`docs/architecture/agent-lifecycle.md:57`)
- 🔁 **Still present:** KB form also offers the experimental VM harness (`packages/ui/src/modules/knowledge-bases/views/knowledge-base-setup-view.tsx:44`)
- 🆕 **New:** settled gate waits on flags, status covers templates only (`packages/ui/src/modules/sandboxes/hooks/use-harness-catalogue.ts:35`)
- 🆕 **New:** the 40k trim dropped prose that moved nowhere (`docs/architecture/agent-lifecycle.md:70`)

### Findings
- 🟢 **Suggestion:** `catalogueSettled` now waits on the feature-flags query too, but `isLoading` / `isError` / `refetch` still come from `useTemplates` alone, so a failed flags query leaves the default harness unselected with no message; the grid stays usable, so the user must pick a harness by hand (`packages/ui/src/modules/sandboxes/hooks/use-harness-catalogue.ts:35`)
- 🟢 **Suggestion:** the 40k trim removed three paragraphs whose content is still true and is now in no page: the `agent-runtime` file-materialization invariant (real PVC path, merged with image-baked content — `runtime-delivery.md` does not state it), the target lifetime model, and the hibernation coupling that unsupervised work survives `session/close`; the widened sentence also gives file contributions the env rail's idle-turn timing, and the trimmed Create sentence now reads as if a kinded create never picks an image (`docs/architecture/agent-lifecycle.md:70`)

_5 findings, all in this round's docs class — merged into the `### Findings` entry above:_
- the widened configuration sentence gives file contributions the env rail's next-idle-turn timing; they reconcile on `applyState` (`docs/architecture/agent-lifecycle.md:70`)
- the deleted connector-files paragraph's PVC-not-`emptyDir` merge invariant is in no page, `runtime-delivery.md` included (`docs/architecture/agent-lifecycle.md:70`)
- the trimmed Create sentence now reads as if a kinded create never picks an image (`docs/architecture/agent-lifecycle.md:56`)
- the coupling that work a harness does not supervise survives `session/close` went with the rationale paragraph (`docs/architecture/agent-lifecycle.md:205`)
- Harness Family is described in depth only on the knowledge-bases page (`docs/architecture/knowledge-bases.md:18`)

### Issue Fit
Linked issue: #3511 "UI - Knowledge base agents can only run on Claude Code" (`Closes #3511`).
- undeclared scope: three architecture paragraphs deleted in a knowledge-base feature PR — merged into the `### Findings` entry above (`docs/architecture/agent-lifecycle.md:70`)
- the shared catalogue also offers `claude-code-vm` with the `vm-sandboxes` flag on — the carryover above (`packages/ui/src/modules/sandboxes/hooks/use-harness-catalogue.ts:33`)

- the conjunctive gate's status and retry cover only the templates query — merged into the `### Findings` entry above (`packages/ui/src/modules/sandboxes/hooks/use-harness-catalogue.ts:35`)
- 🟢 **Suggestion:** the new gate reads a permanently failed templates query as never settled, so `resolveAgentHarness` stays `ready: false` and the knowledge-base onboarding command is never sent; only the generic "Couldn't load templates" toast appears (`packages/ui/src/modules/knowledge-bases/lib/resolve-agent-harness.ts:9`)

- the flags query is in the gate but not in the rendered status — merged into the `### Findings` entry above; the same cause makes VM harnesses appear after first paint (`packages/ui/src/modules/sandboxes/hooks/use-harness-catalogue.ts:35`)

### Verdict
APPROVE — the blocking finding is fixed at both gates and covered by a test; what remains is suggestions.


---

PR #3572, opened 2026-09-04
feat(bob): pick the model in the Config panel from a live model list

xjacka:
Bob's Config panel landed in #3553 without a Model option, because Bob's model list is neither a static catalog nor something the platform's `modelDiscovery` could read: Bob asks a LiteLLM-shaped model-information route under its own service prefix, while discovery only knew how to ask an OpenAI model listing at `/v1/models`. This teaches discovery the second shape, puts Model in the panel, and makes a redirected agent usable without anyone configuring it first.

#3553 has merged, so this now targets `main`.

## Discovery asks the route the harness itself asks

A discovery source can now name a path and which of two answer shapes to read (an OpenAI listing keyed by model id, a LiteLLM model-information listing keyed by model name). Bob's manifest declares its own inference route, so discovery and Bob resolve models through exactly the same request — including any gateway rewrite that fronts a plain LiteLLM, which applies to both alike.

Which gateway gets asked follows from the granted connection. An agent can hold two at once, so they contribute different names and the source ranks them: the **IBM LiteLLM** connection sets `BOB_GATEWAY_URL`, the variable Bob's own client reads, and the **Bob Shell** connection sets `BOB_DEFAULT_GATEWAY_URL` for its own gateway. A redirect wins; Bob's own gateway is the fallback. A duplicate name would otherwise be settled by secret order, which is no way to choose an endpoint.

The request leaves the pod through `HTTPS_PROXY` like every other, so it authenticates with the credential injected at the sidecar; the agent container still only ever holds a placeholder. An agent granted no connection has no gateway to ask and reports *unavailable* without a request.

**Migration:** a Bob Shell connection created before this change carries the env mappings it was created with, so it does not get `BOB_DEFAULT_GATEWAY_URL` until the provider is saved once in Settings → Providers. Until then discovery falls back to whatever `BOB_GATEWAY_URL` a LiteLLM connection supplies, and with neither the Model option is simply absent. Noted in the agent README.

## A redirected agent is given a model

A gateway fronting somebody else's catalogue does not serve Bob's built-in default, and a request naming a model the tenant cannot reach is refused with a 403 rather than answered by something else — so a Bob nobody has configured yet failed its first prompt outright.

With no model set and none pinned, agent-runtime now writes the first of the discovered names into the panel's slot before the harness spawns for the first time. The names are ordered, so the pick is stable rather than whatever the provider happened to return first.

- **It fills an empty slot only.** A panel pick, a hand-edit and a provider pin (`BOB_SHELL_MODEL`) all outrank it. The manifest names the variables that carry a pin so the seed can step aside — it does not arbitrate between people's choices.
- **Only a redirect justifies it.** The manifest marks which `urlEnv` entries redirect the harness. On Bob's own gateway its curated default is the better answer and nothing is written.
- **The first spawn waits for it.** A harness reads its config once, at spawn, so the boot gate holds early callers until the write lands, under a deadline of its own. Boot work starts whenever the environment is ready, not only when the runtime first learns it became ready — a pod records its environment on its own volume, so on every boot after the first that moment never comes.
- **An unavailable listing seeds nothing.** The seed can pick a model, not invent one; a harness whose provider cannot be reached keeps the behavior it had.

pi-agent declares the same two lists and gets the same treatment.

## An apply now reaches the running harness

Because a harness reads its config at spawn, a declared value used to sit unread until something else happened to restart the process — a Config panel change looked applied and did nothing. An apply that actually changed the file now recycles the harness under its own teardown reason: idle, right away; mid-turn, after the work drains. An apply that changed nothing does not.

## The panel stops showing the previous model

`harnessConfig.set` returns once the event is in the outbox, while `harnessConfig.current` reads the harness's file inside the pod, so invalidating on success re-read the pod before the event had landed and cached the old value — and nothing read again while the entry stayed fresh. The apply now writes what it declared into the cache in `onMutate`, after cancelling the in-flight read that would otherwise resolve later and overwrite it; that read carries a discovery round trip, so it is in flight often and lands late. A failed apply rolls back, and the declared value is marked stale so the next read reconciles it with the pod.

A model list back-filled from the snapshot also no longer reports itself as paired with the live model.

## The correction worth flagging

The task described `/model/info` as sitting under Bob's **admin** prefix. It does not — 2.0.2 builds its model-information service on the *inference* client, and a headless run confirms it: the only three gateway requests Bob makes are `GET /inference/v1/model/info`, `GET /admin/v1/profile` and `POST /inference/v1/chat/completions`. This matters beyond a path string: `/inference/v1/*` is the prefix the gateway rewrite already covers, so discovery against a plain LiteLLM works there with nothing further.

## A LiteLLM listing is not a model list

`/model/info` enumerates every deployment the gateway fronts — embeddings, rerankers, image and audio models included — and the aliases are operator-chosen, so filtering on the name alone lets `bge-large` or `rank-v3` into a model dropdown where picking one breaks chat. The same payload carries the provider's own answer in `model_info.mode`, so an entry the gateway marks as anything but conversational is dropped, with the name filter kept as a backstop for entries that carry no mode.

## Degradation is one outcome, not four

No URL granted, a refusal, an unreachable host, and a listing with nothing usable in it all report **unavailable**, which leaves whatever list was last established in place and never fails the read the panel and session creation depend on. The first read of a pod's life pays DNS plus a TLS handshake and is retried once, because a fresh agent's first turn now rides on it.

The empty-listing case is a **behavior change for pi-agent too**: a listing whose every entry filtered out used to report an observed *empty* list, which would blank the dropdown and overwrite a good earlier list with nothing.

## Verification

Suites green on this HEAD: agent-runtime 53 files / 319 tests, api-server 147 / 1385, ui 43 / 366, cli 27 / 162, driver-sdk 1 / 11. `mise run --force check` clean.

Covering the behavior this PR adds:

- the seed writes the first discovered name on a redirect, stays out of the way on the harness's own gateway, and yields both to a value already set and to a provider pin;
- the boot gate runs its work on a pod whose environment was ready at boot, does not let the environment deadline cut that work short on a cold boot, and starts the harness anyway when the work never settles;
- an api-server test parses Bob's shipped manifest and asserts the Bob Shell template contributes `BOB_DEFAULT_GATEWAY_URL` and deliberately not `BOB_GATEWAY_URL`, which is the collision the rename designed away;
- settings precedence exercised directly against `bob-settings.mjs`: a panel value reaches `session.model` over a `BOB_SHELL_MODEL` pin, an absent panel section still lands the pin, and non-string junk in the section exits clean on the pin instead of failing the required bootstrap step.

**Verified on a live cluster** (local k3s, IBM LiteLLM connection granted): a freshly created Bob seeded `aws/claude-opus-4-5` through `BOB_GATEWAY_URL` before its first spawn, the panel rendered the live list, and a turn answered after picking a different model from it. The 403 that motivated the seed reproduces exactly as described — the tenant's allowlist refuses any model outside it, Bob's built-in default included.


packages/agents/bob/runtime-manifest.yaml
```diff
@@ -69,3 +71,14 @@ drivers:
...
               name: "Ask every time"
               description: "Prompts for every tool except file reads, including each shell command. Takes effect on the next harness start, and an unattended session has nobody to answer."
+    # Bob reads its model list from a LiteLLM-shaped /model/info under its own
+    # inference prefix, so discovery follows the same path Bob itself calls —
+    # the gateway rewrite that fronts a plain LiteLLM applies to both alike.
+    # BOB_GATEWAY_URL is set only when a connection points Bob elsewhere;
+    # without it, discovery asks the gateway Bob's own bundle defaults to.
+    modelDiscovery:
+      urlEnv:
+        - BOB_GATEWAY_URL
```
dam-code-guardian, line 86: 🟡 **Warning:** the PR body's central claim about endpoint selection — "the URL comes from `BOB_GATEWAY_URL` when a connection points Bob elsewhere ... Nothing has to be configured per agent" — has no producer in the repository: `BOB_GATEWAY_URL` appears only in this PR's own manifest, tests and README, and the Bob connection template emits `BOBSHELL_API_KEY`, `BOB_SHELL_MODEL`, `BOB_INSTANCE_ID`, `BOB_TEAM_ID`, `BOB_MAX_COINS`, `BOB_CHAT_MODE` and nothing else. Discovery therefore always falls through to the hardcoded `defaultUrl`, so the "follows from the granted connection" half of the declared behavior is not delivered.
  **Fix:** A `modelDiscovery.urlEnv` entry must name an env var that some connection contribution, env mapping or pod-env rail in this repository actually sets; when the only live path is the fallback, say so rather than describing an override no code produces — see `bobEnvMappings` at `packages/api-server-api/src/modules/connections/providers.ts:53`.

packages/agent-runtime/src/modules/runtime-channel/infrastructure/model-discovery.ts
```diff
@@ -46,11 +63,15 @@ export function createOpenAiModelDiscovery(deps: {
...
-            const id = (m as { id?: unknown } | null)?.id;
-            return typeof id === "string" && !/embedding/i.test(id) ? [id] : [];
+            const id = modelIdOf(m, spec.shape ?? "openai-models");
+            return id && !/embedding/i.test(id) ? [id] : [];
           }),
         ),
       ].sort();
+      if (ids.length === 0) {
+        deps.log(`[harness-config] model discovery ${url} → empty model list`);
+        return { status: "unavailable" };
```
dam-code-guardian: 🟡 **Warning:** `unavailable` on the live read returns `values` with no `availableModels` key at all, and the UI's live branch prefers the live payload over the snapshot (`live ?? recorded?.snapshot`, `packages/ui/src/modules/agents/api/harness-config.ts:70`), so a running agent whose discovery request fails falls through to `group.choices` — `[]` for Bob and pi-agent — and the Model dropdown drops out of the Config panel instead of showing the last established list. The `README`/architecture text this PR adds asserts the opposite ("leaves the panel on its last known list"), so the documented contract does not hold on the path that serves a running agent.
  **Fix:** When a live read cannot establish a value, the reader must carry the last known value forward instead of omitting the field, so an omitted field never degrades to a static catalog default; any surface that claims "last known list" resolves snapshot over live for that one field.

packages/agents/bob/bob-settings.mjs
```diff
@@ -73,7 +73,7 @@ function writeSettings() {
   const panel = section(existing, "platform");
   const mode =
     normalizeMode(panel.mode) ?? normalizeMode(process.env.BOB_CHAT_MODE);
-  const model = firstNonBlank(process.env.BOB_SHELL_MODEL);
+  const model = firstNonBlank(panel.model, process.env.BOB_SHELL_MODEL);
```
dam-code-guardian, line 77: 🟡 **Warning:** `panel.model` is read out of `$HOME/.bob/settings/settings.json` — a file the README says users may hand-edit — and passed straight to `firstNonBlank`, which calls `value?.trim()`; a `platform.model` that is a number, object or array throws a `TypeError` inside `writeSettings`, the step marked `required: true`, so one hand-edit fails the settings bootstrap on every harness start. `normalizeMode(panel.mode)` on line 75 carries the same exposure.
  **Fix:** Every value read out of the harness's own settings file must be narrowed to a string at the read boundary (`typeof v === "string"`), never at the point of use.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `1d9f7b5`

### Summary
Generalizes model discovery from OpenAI listings to also read a LiteLLM `/model/info` shape, and puts a live Model option on Bob's Config panel. The mechanism works, but the endpoint override it documents has no producer in the repo, and the "last known list" contract the new docs state does not hold on the live read path.

### Findings

- 🟡 **Warning:** documented `BOB_GATEWAY_URL` override has no producer (`packages/agents/bob/runtime-manifest.yaml:81`)
- 🟡 **Warning:** a failed live read drops the model list instead of keeping it (`packages/agent-runtime/src/modules/runtime-channel/infrastructure/model-discovery.ts:73`)
- 🟡 **Warning:** stale precondition two paragraphs from its own correction (`docs/architecture/runtime-delivery.md:57`)
- 🟡 **Warning:** unvalidated settings value fails the required bootstrap step (`packages/agents/bob/bob-settings.mjs:76`)
- 🟢 **Suggestion:** the added architecture sentence walks the discovery object field by field and names both listing shapes with the response key each reads. `docs/guidelines/documentation-guidelines.md:41` puts field names and enum members on the Omit list and asks for the meaning of each outcome plus a link to the code, so the sentence rots on a rename or a third shape and spends the page's character cap on detail the code already holds (`docs/architecture/runtime-delivery.md:55`)
- ✅ **Looks good:** the empty-listing flip to `unavailable` is correct for the snapshot rail — `packages/api-server/src/modules/harness-config/infrastructure/snapshot-repo.ts:50` patches `availableModels` only when the key is present, so omitting it on a failed read keeps the stored list instead of overwriting it with `[]`.

- 🟡 **Warning:** the snapshot paragraph still states the discovered model list "can only be resolved once the env driver has materialized the provider's base URL", which this PR makes false: `modelDiscovery` now accepts a `defaultUrl` fallback (`packages/agent-runtime/src/modules/runtime-channel/manifest.ts:46`, `packages/agent-runtime/src/modules/runtime-channel/infrastructure/model-discovery.ts:44-48`) and Bob's manifest ships one (`packages/agents/bob/runtime-manifest.yaml:82`), so a source with no env var set still resolves. The same PR's edit to line 55 documents the fallback, leaving the page self-contradictory two paragraphs apart (`docs/architecture/runtime-delivery.md:57`)
  **Fix:** When a PR widens a mechanism, re-read the whole architecture page for every other sentence that asserts the old precondition — here, drop the env-materialization precondition on line 57 so the apply reply carries the discovered list whenever the source resolves to a URL at all, from the env rail or the source's own fallback.

### Issue Fit

_No linked issue — the PR body declares no closing keyword, its only `#` references (#3553, #3565) are pull requests, and the head branch carries no issue number. Scope checked against the PR description's own promises._

- 🟡 **Warning:** the body's central claim about endpoint selection — "the URL comes from `BOB_GATEWAY_URL` when a connection points Bob elsewhere ... Nothing has to be configured per agent" — has no producer in the repository: `BOB_GATEWAY_URL` appears only in this PR's own manifest, tests and README, and the Bob connection template emits `BOBSHELL_API_KEY`, `BOB_SHELL_MODEL`, `BOB_INSTANCE_ID`, `BOB_TEAM_ID`, `BOB_MAX_COINS`, `BOB_CHAT_MODE` and nothing else. Discovery therefore always falls through to the hardcoded `defaultUrl`, so the "follows from the granted connection" half of the declared behavior is not delivered (`packages/agents/bob/runtime-manifest.yaml:81`)
  **Fix:** A `modelDiscovery.urlEnv` entry must name an env var that some connection contribution, env mapping or pod-env rail in this repository actually sets; when the only live path is the fallback, say so rather than describing an override no code produces — see `bobEnvMappings` at `packages/api-server-api/src/modules/connections/providers.ts:53`.

- 🟡 **Warning:** `unavailable` on the live read returns `values` with no `availableModels` key at all, and the UI's live branch prefers the live payload over the snapshot (`live ?? recorded?.snapshot`, `packages/ui/src/modules/agents/api/harness-config.ts:70`), so a running agent whose discovery request fails falls through to `group.choices` — `[]` for Bob and pi-agent — and the Model dropdown drops out of the Config panel instead of showing the last established list. The `README`/architecture text this PR adds asserts the opposite ("leaves the panel on its last known list"), so the documented contract does not hold on the path that serves a running agent (`packages/agent-runtime/src/modules/runtime-channel/infrastructure/model-discovery.ts:73`)
  **Fix:** When a live read cannot establish a value, the reader must carry the last known value forward instead of omitting the field, so an omitted field never degrades to a static catalog default; any surface that claims "last known list" resolves snapshot over live for that one field.
- 🟡 **Warning:** `panel.model` is read out of `$HOME/.bob/settings/settings.json` — a file the README says users may hand-edit — and passed straight to `firstNonBlank`, which calls `value?.trim()`; a `platform.model` that is a number, object or array throws a `TypeError` inside `writeSettings`, the step marked `required: true`, so one hand-edit fails the settings bootstrap on every harness start. `normalizeMode(panel.mode)` on line 75 carries the same exposure (`packages/agents/bob/bob-settings.mjs:76`)
  **Fix:** Every value read out of the harness's own settings file must be narrowed to a string at the read boundary (`typeof v === "string"`), never at the point of use.

### Verdict
COMMENT — four warnings: the documented `BOB_GATEWAY_URL` override has no producer, the "last known list" contract fails on the live read, the architecture page contradicts itself, and an unvalidated settings value can fail Bob's required bootstrap step.


---

packages/agent-runtime/src/modules/runtime-channel/infrastructure/model-discovery.ts
```diff
@@ -16,7 +15,38 @@ export type ModelDiscovery = (
...
+}
+
+function liteLlmModelName(entry: Record<string, unknown>): string | null {
+  const mode = (entry.model_info as Record<string, unknown> | null)?.mode;
+  if (typeof mode === "string" && !CONVERSATIONAL_MODES.has(mode)) return null;
+  const name = entry.model_name;
+  return typeof name === "string" && name.length > 0 ? name : null;
+}
+
+function chatModelIdOf(entry: unknown, shape: ModelListShape): string | null {
```
dam-code-guardian, line 48: 🟢 **Suggestion:** `chatModelIdOf` picks the reader with a ternary on the closed union `ModelListShape`, so a third `shape` added to the manifest enum falls through to the OpenAI reader and yields an empty list at runtime instead of failing to compile. `docs/guidelines/code/exhaustive-matching.md` asks for `match(...).exhaustive()` on a closed union, and `agent-runtime` already depends on `ts-pattern`.

packages/ui/src/modules/agents/api/harness-config.ts
```diff
@@ -67,7 +67,12 @@ export function useResolvedHarnessConfig(
   const pending = snapshotPending;
 
   if (operable) {
-    const values = live ?? recorded?.snapshot ?? null;
+    const recent = recorded?.snapshot;
+    const read = live ?? recent ?? null;
+    const values =
+      read && read.availableModels === undefined && recent?.availableModels
```
dam-code-guardian: 🟢 **Suggestion:** The retained list is returned as `origin: "live"` with `modelsPaired: true`, and the stale-model callout only fires on `origin === "snapshot"` — so a list established by an earlier discovery reaches a running agent's panel with nothing marking it as not current.

dam-code-guardian (dismissed):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `ae6d37e`

## PR #3572: feat(bob): pick the model in the Config panel from a live model list
**Author:** xjacka | **Branch:** feat/bob-model-discovery → feat/bob-acp-native | **Changes:** +218 −42 (12 files)

### Summary
This round sources `BOB_GATEWAY_URL` from the Bob Shell connection's env mappings, drops the manifest's `defaultUrl` fallback, filters a LiteLLM listing down to its conversational entries, and retains the last known model list in the Config panel when a live read comes back `unavailable`. Every finding of the previous review is resolved.

### Changes since last review
Previous HEAD: 1d9f7b5 (2026-09-07T21:09:28Z) — verdict COMMENT

- ✅ **Fixed:** BOB_GATEWAY_URL override has no producer (`packages/agents/bob/runtime-manifest.yaml:81`)
- ✅ **Fixed:** failed live read drops the model list (`packages/agent-runtime/src/modules/runtime-channel/infrastructure/model-discovery.ts:73`)
- ✅ **Fixed:** stale env–materialization precondition contradicts line 55 (`docs/architecture/runtime-delivery.md:57`)
- ✅ **Fixed:** unvalidated panel.model fails the required bootstrap step (`packages/agents/bob/bob-settings.mjs:76`)
- ✅ **Fixed:** architecture page transcribes field names and enum members (`docs/architecture/runtime-delivery.md:55`)
- 🆕 **New:** shape union branched with a ternary, not exhaustively (`packages/agent-runtime/src/modules/runtime-channel/infrastructure/model-discovery.ts:39`)
- 🆕 **New:** back-filled model list reads as live and paired (`packages/ui/src/modules/agents/api/harness-config.ts:73`)

### Findings
- 🟢 **Suggestion:** `chatModelIdOf` picks the reader with a ternary on the closed union `ModelListShape`, so a third `shape` added to the manifest enum falls through to the OpenAI reader and yields an empty list at runtime instead of failing to compile. `docs/guidelines/code/exhaustive-matching.md` asks for `match(...).exhaustive()` on a closed union, and `agent-runtime` already depends on `ts-pattern` (`packages/agent-runtime/src/modules/runtime-channel/infrastructure/model-discovery.ts:39`)
- 🟢 **Suggestion:** The retained list is returned as `origin: "live"` with `modelsPaired: true`, and the stale-model callout only fires on `origin === "snapshot"` — so a list established by an earlier discovery reaches a running agent's panel with nothing marking it as not current (`packages/ui/src/modules/agents/api/harness-config.ts:73`)

_1 suggestion dropped under the 🟢 budget._

### Issue Fit
- 🟢 **Suggestion:** No linked issue. The body carries no closing keyword and no issue reference (only PRs #3553 and #3565), and the branch names no issue number, so scope can only be judged against the description — the nearest written ask, a comment on issue #3552, deliberately excluded the model dropdown as separately specced work, and no such spec issue exists (`PR #3572 description`)

_2 finding(s) carried into `### Findings` above; 6 suggestion(s) dropped under the 🟢 budget._

### Verdict
APPROVE — every finding of the previous review is fixed and nothing blocking survives verification at this HEAD.


---

packages/agent-runtime/src/modules/acp/services/acp-runtime/harness-lease.ts
```diff
@@ -59,7 +71,23 @@ export function createHarnessLease(deps: HarnessLeaseDeps): HarnessLease {
     readyWaiters.clear();
   }
 
-  if (!envReady) warmTimer = setTimeout(markEnvReady, deps.warmStartTimeoutMs);
+  function markEnvReady(): void {
+    if (envReady) return;
```
dam-code-guardian: 🔴 **Critical:** The model seed hangs off `beforeFirstSpawn`, which only `markEnvReady()` reaches — and this line returns early when `envReady` is already true. `envReady` starts from `deps.envReadyAtBoot`, which is `envStore.ready()`, an `existsSync` on `$HOME/.platform/runtime-env.json`; `$HOME` is a persistent volume, so on every boot after the first the hold never runs and a redirected harness starts on a built-in default. `attach()` reaches the first spawn only through `whenReady`, and the new unit spec calls `seedModel()` on the plugin directly, so nothing covers the path. Reported by typescript-engineering.

**Fix:** Boot work that must precede the first spawn belongs on the spawn path, not on the env-readiness transition — gate `ensure()`/`whenReady()` on a one-shot latch that runs whatever the env state was at boot, so a precondition never becomes conditional on a state the pod may already be in.

packages/agent-runtime/src/modules/runtime-channel/drivers/harness-config-plugin.ts
```diff
@@ -97,6 +99,7 @@ export function createHarnessConfigPlugin(deps: {
       log,
       onUnparseable: "throw",
     });
+    deps.onApplied?.();
```
dam-code-guardian: 🟡 **Warning:** `deps.onApplied?.()` fires at the end of every `apply`, and `server.ts` wires it to `acpRuntime.recycleForConfig()`. So saving any harness-config field — Mode and Approvals as much as Model, on every harness with this driver — now tears down the running harness and kills the in-flight session. Only the Model path needs a fresh process. Reported by issue-fit as undeclared scope.

**Fix:** Fire the recycle callback only for the keys the harness reads at spawn time, deciding that at the callback site rather than for every applied key.

packages/agent-runtime/src/modules/runtime-channel/drivers/harness-config-plugin.ts
```diff
@@ -120,12 +123,34 @@ export function createHarnessConfigPlugin(deps: {
     }
   };
 
+  const seedModel = async (): Promise<boolean> => {
```
dam-code-guardian, line 128: 🟡 **Warning:** The body describes the live model list and the resolver only. Three behaviors ship undeclared: this seeded write into a user-owned config file with no user action behind it (with the new `redirectEnv` field, which for pi-agent lists both of its URL names so it always seeds), the config recycle, and the rewrite of `useApplyHarnessConfig` from invalidate-on-success to an optimistic write with a `useRef` rollback. Reported by issue-fit.

**Fix:** Declare every behavior the diff adds beyond the titled change in the PR body — an unannounced write to a user-owned file, a second agent's behavior change, and a mutation-semantics change are the ones a reviewer is least likely to infer from the title.

packages/ui/src/modules/agents/api/harness-config.ts
```diff
@@ -112,14 +119,54 @@ export function useStaleModel(agentId: string | null): {
...
   return useMutation({
     ...trpc.harnessConfig.set.mutationOptions(),
-    meta: { errorToast: "Failed to apply model settings" },
-    onSuccess: (_data, variables) => {
-      void queryClient.invalidateQueries({
-        queryKey: harnessConfigCurrentKey(variables.agentId),
-      });
+    meta: {
+      errorToast: "Failed to apply model settings",
+      invalidates: [trpc.harnessConfig.snapshot.queryKey()],
```
dam-code-guardian, line 155: 🟡 **Warning:** `onMutate` writes an optimistic `HarnessConfigCurrent` into `harnessConfigCurrentKey`, but `meta.invalidates` lists only the snapshot key and `onSettled` invalidates nothing on success. When `apply` succeeds while dropping fields — the plugin logs `no key mapping for "<id>" — skipping` — the panel keeps showing a value that never reached the harness. `query-client.ts` does dispatch `meta.invalidates`, so the gap is the missing key, not the dispatcher. Reported by typescript-engineering.

**Fix:** Every query key an optimistic update writes must be invalidated or overwritten from the server response in `onSettled`, so an optimistic value can never survive as the displayed truth.

packages/agent-runtime/src/server.ts
```diff
@@ -160,6 +160,8 @@ if (envStore.ready()) podService?.refreshEnv();
 
 process.env.PLATFORM_RUNTIME_URL = `http://127.0.0.1:${config.PORT}`;
 
+let seedHarnessModel: (() => Promise<void>) | null = null;
```
dam-code-guardian, line 163: 🟡 **Warning:** `seedHarnessModel` is a mutable `let` initialized to `null`, closed over at line 184 as the `beforeFirstSpawn` hook and only assigned at line 245, after `await composeRuntimeChannel(...)`. A readiness transition during that await resolves the hook to `null`, and the seed no-ops with no log. Reported by typescript-engineering.

**Fix:** Never close over a mutable module-level `let` to break a wiring cycle — compose the dependency before its consumer, or pass a deferred the consumer awaits, so a not-yet-wired dependency is a wait rather than a silent skip.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `380a1be`

## PR #3572: feat(bob): pick the model in the Config panel from a live model list
**Author:** xjacka | **Branch:** feat/bob-model-discovery → main | **Changes:** +553 −96 (21 files)

### Summary
The round adds a boot-time model seed, a config recycle and an optimistic Config-panel write on top of the live model list. The seed is unreachable on any pod that boots with env state already on disk, so the feature does not run where it is needed.

### Changes since last review
Previous HEAD: ae6d37e (2026-09-08T07:57:40Z) — verdict APPROVE — unreachable, reviewed the whole PR

- ✅ **Fixed:** BOB_GATEWAY_URL override has no producer (`packages/agents/bob/runtime-manifest.yaml:81`)
- ✅ **Fixed:** failed live read drops the model list (`packages/agent-runtime/src/modules/runtime-channel/infrastructure/model-discovery.ts:73`)
- ✅ **Fixed:** stale env-materialization precondition contradicts line 55 (`docs/architecture/runtime-delivery.md:57`)
- ✅ **Fixed:** unvalidated panel.model fails the required bootstrap step (`packages/agents/bob/bob-settings.mjs:76`)
- ✅ **Fixed:** architecture page transcribes field names and enum members (`docs/architecture/runtime-delivery.md:55`)
- 🔁 **Still present:** shape union branched with a ternary, not exhaustively (`packages/agent-runtime/src/modules/runtime-channel/infrastructure/model-discovery.ts:44`)
- 🔁 **Still present:** back-filled model list reads as live and paired (`packages/ui/src/modules/agents/api/harness-config.ts:73`)
- 🆕 **New:** model seed unreachable when the env file already exists (`packages/agent-runtime/src/modules/acp/services/acp-runtime/harness-lease.ts:75`)
- 🆕 **New:** every harness-config apply now recycles the harness (`packages/agent-runtime/src/modules/runtime-channel/drivers/harness-config-plugin.ts:102`)
- 🆕 **New:** optimistic current write is never invalidated on success (`packages/ui/src/modules/agents/api/harness-config.ts:152`)
- 🆕 **New:** seed hook closes over a let assigned after compose (`packages/agent-runtime/src/server.ts:163`)
- 🆕 **New:** body declares neither the seed path nor the UI rewrite (`packages/agent-runtime/src/modules/runtime-channel/drivers/harness-config-plugin.ts:126`)
- 🆕 **New:** page still teaches the pre-rename gateway variable contract (`docs/architecture/connections.md:196`)
- 🆕 **New:** PR body contradicts the shipped env mapping and the cited test (`packages/api-server-api/src/modules/connections/providers.ts:64`)

### Findings

- 🔴 **Critical:** model seed unreachable when the env file already exists (`packages/agent-runtime/src/modules/acp/services/acp-runtime/harness-lease.ts:75`)
- 🟡 **Warning:** every harness-config apply now recycles the harness (`packages/agent-runtime/src/modules/runtime-channel/drivers/harness-config-plugin.ts:102`)
- 🟡 **Warning:** optimistic current write is never invalidated on success (`packages/ui/src/modules/agents/api/harness-config.ts:152`)
- 🟡 **Warning:** seed hook closes over a let assigned after compose (`packages/agent-runtime/src/server.ts:163`)
- 🟡 **Warning:** body declares neither the seed path nor the UI rewrite (`packages/agent-runtime/src/modules/runtime-channel/drivers/harness-config-plugin.ts:126`)
- 🟡 **Warning:** The IBM LiteLLM section of `docs/architecture/connections.md` still argues from a `BOB_GATEWAY_URL` name collision this round designed away, and still requires a model to be pinned by hand — the seed path removes that precondition. Reported by doc-drift (`docs/architecture/connections.md:196`, `:199`).
  **Fix:** When a change renames an env name or removes a manual configuration precondition, restate every architecture page that teaches the old contract in the same round.
- 🟡 **Warning:** The body says the Bob Shell connection contributes `BOB_GATEWAY_URL` and that there is no fallback URL; the template contributes `BOB_DEFAULT_GATEWAY_URL`, which is that fallback. The Verification section describes the new api-server test as proving every discovery env name has a producer, while the test asserts the template does **not** contribute `BOB_GATEWAY_URL`. Reported by issue-fit (`packages/api-server-api/src/modules/connections/providers.ts:64`, `packages/api-server/src/__tests__/unit/connections-bob-template.test.ts:44`).
  **Fix:** Restate the body's env-name and verification claims against HEAD whenever a later commit re-scopes them — a body that names the pre-rename variable is a false record of the merged behavior.

🔁 2 finding(s) reported here are carried in `### Findings` above (`docs/architecture/connections.md`).

- 🟢 **Suggestion:** `docs/architecture/runtime-delivery.md` grew to 39,855 chars, 145 under the hard 40,000 cap `check:doc-size` enforces unbypassably, so the next edit to the page fails the gate. Consider moving the harness-config / model-discovery / seeding material to its own subsystem page (`docs/architecture/runtime-delivery.md:57`).

### Issue Fit
_No linked issue — no closing keyword, no plain issue reference (`#3553` and `#3565` are both PRs), and the head branch `feat/bob-model-discovery` carries no issue number. Scope checked against the PR description only._

🔁 5 finding(s) reported here are carried in `### Findings` above (body claims, and three undeclared-scope items: the seed path with `redirectEnv` and `beforeFirstSpawn`, the config recycle, the `useApplyHarnessConfig` rewrite).

- 🟢 _1 suggestion dropped under the 🟢 budget._

🔁 8 finding(s) reported here are carried above — in `### Findings` (`harness-lease.ts:75`, `server.ts:163`, `harness-config.ts:152`, `harness-config-plugin.ts:42`) or in the delta block (`harness-config.ts:73`, `model-discovery.ts:44`). Two more failed the blocking bar and are 🟢: `RECYCLE_LOG`'s `Partial<Record<>>` fallback has nine occurrences in untouched code, and the `useRef` rollback has no caller that issues two applies at once.

- 🟢 _6 suggestion(s) dropped under the 🟢 budget._

### Verdict
REQUEST_CHANGES — the model seed cannot run on a pod whose env file already exists, which is every boot after the first.

### For the human reviewer
This round hangs the automatic model pick on the moment the runtime first learns its environment is ready. On a pod that already has its environment recorded on disk from an earlier boot, that moment never arrives again, so the automatic pick is skipped exactly where the feature matters and the agent starts on a built-in default the upstream provider may not serve. The new tests exercise the pick directly rather than through the startup path, so they pass either way. Settle where boot work that must precede the first start belongs before looking at the smaller items, and decide separately whether saving any Config panel field should restart a running session, which this round makes it do for every field and every agent that uses the driver.


---

packages/agent-runtime/src/modules/acp/services/acp-runtime/harness-lease.ts
```diff
@@ -24,42 +26,96 @@ export interface HarnessLeaseDeps {
...
+    envReady = true;
+    startBootWork();
+    releaseIfReady();
+  }
+
+  if (!envReady) {
+    warmTimer = setTimeout(() => {
+      envReady = true;
+      bootWorkDone = true;
```
dam-code-guardian: 🔴 **Critical:** on a pod whose env is not ready at boot, `markEnvReady` starts the boot work but never cancels `warmTimer` — only `releaseWaiters` clears it, and that runs when both flags are set. Two reachable outcomes on a first-ever boot, the case the seed exists for. If the env arrives before the deadline, the still-armed timer fires and releases the first attach before `beforeFirstSpawn` settles, so the harness reads its config file before the model is written. If the env arrives after the deadline, the timer has already set `bootWorkDone`, so `startBootWork` returns at its own guard and the seed never runs at all, with no log line. Either way the redirected agent starts with no model and its first prompt is refused.
  **Fix:** Give each wait its own timer, and cancel the env timer when the env arrives. One stage's timeout callback must never set another stage's completion flag; arm the boot-work deadline when that wait starts, and clear it on every transition that ends the wait.

The cold-boot path also has no test — both new cases pass `envReadyAtBoot: true`.

docs/architecture/connections.md
```diff
@@ -193,12 +193,13 @@ to offer for this to work is the OpenAI-shaped chat-completion route, a
...
-Connection claims the same variable from its own input, an agent can
-hold both, and two contributors for one name would leave the winner to
-grant order. It rides the provider pin or a per-agent override instead.
+off the configured host. This Connection contributes no model of its
+own, and Bob's built-in default resolves to one only its own gateway
+serves — so a model has to come from somewhere else. Nobody has to pin
+one by hand: because this Connection redirects Bob, the platform seeds
+the first model the endpoint's own listing offers before the harness
+starts ([runtime-delivery](runtime-delivery.md#concepts)). A provider pin
+or a per-agent override still wins over it.
```
dam-code-guardian: 🟡 **Warning:** the precedence is the other way round. The seed writes the discovered model into the panel section (`platform.model`), and `bob-settings.mjs:77` resolves `firstNonBlank(panel.model, process.env.BOB_SHELL_MODEL)` — the panel value first. `packages/agents/bob/README.md:68` states the same order, and the PR body states it too. An agent that holds this Connection and a `BOB_SHELL_MODEL` pin, with no panel pick of its own, is seeded on the first boot and then starts on the seeded model, not on the pin; the same holds for a per-agent Env override. So this sentence tells an operator the opposite of what a redirected agent does, and two other documents disagree with it.
  **Fix:** Resolve a stated precedence against the code that merges the two sources, and state that order. A newly added writer needs its position in the chain checked before any precedence claim is made about it.

```suggestion
starts ([runtime-delivery](runtime-delivery.md#concepts)). The seeded value
sits in the panel slot, so it wins over a provider pin and a per-agent override.
```

docs/architecture/connections.md
```diff
@@ -193,12 +193,13 @@ to offer for this to work is the OpenAI-shaped chat-completion route, a
...
-Bob's built-in default resolves to one only its own gateway serves — but
-not on this Connection, which contributes no model of its own: the Bob
-Connection claims the same variable from its own input, an agent can
-hold both, and two contributors for one name would leave the winner to
-grant order. It rides the provider pin or a per-agent override instead.
+off the configured host. This Connection contributes no model of its
+own, and Bob's built-in default resolves to one only its own gateway
+serves — so a model has to come from somewhere else. Nobody has to pin
+one by hand: because this Connection redirects Bob, the platform seeds
+the first model the endpoint's own listing offers before the harness
```
dam-code-guardian: 🟡 **Warning:** the seed does not take the first model the listing offers. `createModelDiscovery` builds its ids as `[...new Set(...)].sort()` (`packages/agent-runtime/src/modules/runtime-channel/infrastructure/model-discovery.ts:84`), and `seedModel` takes `outcome.models[0]`, so the seeded model is the alphabetically first chat model of the listing. An operator who reads this line and predicts which model an unconfigured agent starts on gets the wrong answer whenever the listing's own first entry is not the alphabetically first one. `docs/architecture/runtime-delivery.md:57` carries the same wording.
  **Fix:** State a selection rule by the property the code guarantees — here a deterministic pick that does not depend on listing order — on every page that makes the claim.

docs/architecture/runtime-delivery.md
```diff
@@ -52,7 +52,9 @@ While a workspace-mutating event (`workspace-seed`, `workspace-command`) is pend
...
...n its own built-in default: the model it would name need not exist upstream, and a listing that does not know it refuses the turn outright rather than substituting anything — so an agent nobody has configured yet fails its very first prompt. The seed closes that: with no model set, agent-runtime takes the first model the live listing offers and writes it before the harness spawns for the first time, the boot gate holding early callers until it lands, bounded by the same warm-start ceiling that covers the env rail. It never speaks over a value that exists, a user's pick and a hand-edit alike, and never runs twice for one. Whether a source justifies it follows from *which* variable supplied the URL, which is why the manifest marks the redirecting ones apart from those that merely carry the harness's own endpoint: on its own provider the harness's default is the better answer and stays untouched. An unavailable listing seeds nothing, so a harness whose provider cannot be reached keeps the behavior it had — the seed can pick a model, not invent one. Only harnesses whose manifest declares the `harness-config` driver honor the event and advertise the capability that gates the UI section.
```
dam-code-guardian: 🟡 **Warning:** same finding as `docs/architecture/connections.md:200`. "the first model the live listing offers" is the alphabetically first chat model, because discovery de-duplicates and sorts the ids before the seed takes element zero (`packages/agent-runtime/src/modules/runtime-channel/infrastructure/model-discovery.ts:84`).
  **Fix:** State a selection rule by the property the code guarantees — here a deterministic pick that does not depend on listing order — on every page that makes the claim.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `f5a83a6`

## PR #3572: feat(bob): pick the model in the Config panel from a live model list
**Author:** xjacka | **Branch:** feat/bob-model-discovery → main | **Changes:** +681 −104 (23 files)

### Summary
This round answers the previous review at its cause: the seed now starts at the first attach, an apply recycles the harness only when a value changed, and the panel's optimistic write is invalidated on success. The new boot gate has a timer defect on a cold boot, and the rewritten `connections.md` paragraph makes two claims the code contradicts.
_20 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: 380a1be (2026-09-09T14:58:38Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** model seed unreachable when the env file already exists (`packages/agent-runtime/src/modules/acp/services/acp-runtime/harness-lease.ts:75`)
- ✅ **Fixed:** every harness-config apply now recycles the harness (`packages/agent-runtime/src/modules/runtime-channel/drivers/harness-config-plugin.ts:102`)
- ✅ **Fixed:** optimistic current write is never invalidated on success (`packages/ui/src/modules/agents/api/harness-config.ts:152`)
- ✅ **Fixed:** seed hook closes over a let assigned after compose (`packages/agent-runtime/src/server.ts:163`)
- ✅ **Fixed:** page still teaches the pre-rename gateway variable contract (`docs/architecture/connections.md:196`)
- 🔁 **Still present:** PR body contradicts the shipped env mapping and the cited test (`packages/api-server-api/src/modules/connections/providers.ts:64`)
- 🔁 **Still present:** the body declares neither the seed path nor the harness recycle (`packages/agent-runtime/src/modules/runtime-channel/drivers/harness-config-plugin.ts:128`)
- 🔁 **Still present:** shape union branched with a ternary, not exhaustively (`packages/agent-runtime/src/modules/runtime-channel/infrastructure/model-discovery.ts:44`)
- 🔁 **Still present:** back-filled model list reads as live and paired (`packages/ui/src/modules/agents/api/harness-config.ts:73`)
- 🆕 **New:** cold boot releases the first spawn before the seed lands (`packages/agent-runtime/src/modules/acp/services/acp-runtime/harness-lease.ts:115`)
- 🆕 **New:** the page inverts the seed against the provider pin (`docs/architecture/connections.md:202`)
- 🆕 **New:** the seed takes the alphabetically first model, not the listed first (`docs/architecture/connections.md:200`)
- 🆕 **New:** the README still teaches the pre-PR model contract (`packages/agents/bob/README.md:122`)

### Findings

- 🔴 **Critical:** cold boot releases the first spawn before the seed lands (`packages/agent-runtime/src/modules/acp/services/acp-runtime/harness-lease.ts:115`)
  **Fix:** Give each wait its own timer, and cancel the env timer when the env arrives.
- 🟡 **Warning:** the page inverts the seed against the provider pin (`docs/architecture/connections.md:202`)
  **Fix:** Resolve a stated precedence against the code that merges the two sources.
- 🟡 **Warning:** the seed takes the alphabetically first model, not the listed first (`docs/architecture/connections.md:200`, also `docs/architecture/runtime-delivery.md:57`)
  **Fix:** State a selection rule by the property the code guarantees, on every page that makes it.
- 🟢 **Suggestion:** the README's LiteLLM-gateway section still teaches the pre-PR contract. It says the connection contributes no model because the name would collide with the Bob Shell pin, that `BOB_SHELL_MODEL` is the way to set one, and that the Config panel offers no Model. The rename, the live list and the seed each falsify one of those clauses (`packages/agents/bob/README.md:122`).

_Both blocking findings are merged into `### Findings` — the two `connections.md` claims. 5 suggestion(s) dropped under the 🟢 budget._

### Issue Fit

_No linked issue: no closing keyword and no issue reference, so the scope was checked against the PR description alone._
_4 finding(s) are merged into the two still-present body findings above. 1 finding was dropped: the body's "There is no fallback URL" claim holds, because both URL variables come from a granted connection._

_1 critical finding is merged into `### Findings` — the boot-gate timer. 1 finding confirms the still-present `harness-config.ts:73` carryover. 14 suggestion(s) dropped under the 🟢 budget._

### Verdict
REQUEST_CHANGES — one critical defect in the new boot gate, and four open warnings.

### For the human reviewer
This change lets the platform choose a model for an agent that nobody has configured yet. It writes that choice into the same slot a person uses in the Config panel, and the platform already promises that slot beats a model pinned on the provider. An operator who pins a model can therefore get a different one, with no message to say so, and the new page text states the opposite order. Settle whether a platform-chosen model should outrank a person's pin before the individual wording fixes matter.


---

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `a751f7a`

## PR #3572: feat(bob): pick the model in the Config panel from a live model list
**Author:** xjacka | **Branch:** feat/bob-model-discovery → main | **Changes:** +754 −105 (23 files)

### Summary
One commit answers the last round at its cause: each wait in the boot gate has its own deadline, the env timer is cancelled when the environment arrives, and the seed now steps aside for a model pinned on the provider. The two architecture pages and the README state the precedence and the selection rule the code guarantees.

### Changes since last review
Previous HEAD: f5a83a6 (2026-09-09T16:14:04Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** cold boot releases the first spawn before the seed lands (`packages/agent-runtime/src/modules/acp/services/acp-runtime/harness-lease.ts:115`)
- ✅ **Fixed:** page inverts the seed against the provider pin (`docs/architecture/connections.md:202`)
- ✅ **Fixed:** seed takes the alphabetically first model, not the listed first (`docs/architecture/connections.md:200`)
- ✅ **Fixed:** PR body contradicts the shipped env mapping and the cited test (`packages/api-server-api/src/modules/connections/providers.ts:64`)
- ✅ **Fixed:** body declares neither the seed path nor the harness recycle (`packages/agent-runtime/src/modules/runtime-channel/drivers/harness-config-plugin.ts:128`)
- ✅ **Fixed:** README still teaches the pre-PR model contract (`packages/agents/bob/README.md:122`)
- ✅ **Fixed:** shape union branched with a ternary, not exhaustively (`packages/agent-runtime/src/modules/runtime-channel/infrastructure/model-discovery.ts:44`)
- ✅ **Fixed:** back-filled model list reads as live and paired (`packages/ui/src/modules/agents/api/harness-config.ts:73`)
- 🆕 **New:** pin deference reads platform–wide, and pi-agent's pin list is always set (`docs/architecture/runtime-delivery.md:55`)
- 🆕 **New:** ceiling path opens the gate with the boot work still armed (`packages/agent-runtime/src/modules/acp/services/acp-runtime/harness-lease.ts:122`)

### Findings

- 🟢 **Suggestion:** the seed defers to a pin only for the variables the agent's own manifest lists, but the paragraph that defines a discovery source names the URL variables and the dialect alone, so the deference reads as a platform-wide rule. pi-agent lists `OPENAI_PROXY_MODEL` and `OPENAI_MODEL`, and the IBM LiteLLM connection contributes both as placeholders whether or not an operator set one, so pi-agent is never seeded and the body's parity claim does not hold at HEAD. pi's own provider extension still resolves a model, so no agent loses its first prompt (`docs/architecture/runtime-delivery.md:55`, also `packages/agents/pi-agent/runtime-manifest.yaml:47`).
- 🟢 **Suggestion:** the env ceiling opens the gate without settling the boot work, so `startBootWork` stays armed and the first attach after the deadline runs the seed while the harness released by that same deadline is already up; the seed's write then recycles it. The path is the degraded one — no environment inside 15 s — the recycle waits for a busy turn to drain, and the env arrival recycles the harness anyway, so the outcome is one extra restart rather than a wrong model. It is also the path the new test does not cover, and the two deadlines are the same value, so a first caller can be held for twice the warm-start ceiling — which the page no longer lets a reader derive (`packages/agent-runtime/src/modules/acp/services/acp-runtime/harness-lease.ts:122`).

_Both warnings are regraded to suggestions and merged into `### Findings` — the pin paragraph's scope, and the missing bound on the seed's wait. 1 suggestion dropped under the 🟢 budget._

### Issue Fit

_No linked issue: no closing keyword and no issue reference, so the scope was checked against the PR description alone. Both body findings of the last review are closed — the env mapping now matches `providers.ts` and the cited test, and the seed path and the harness recycle are declared. 1 warning is regraded and merged into the pin finding above; 2 suggestions dropped under the 🟢 budget._

_The critical finding and two suggestions are regraded and merged into `### Findings` — the ceiling path, its untested branch, and the doubled hold. 4 finding(s) dropped under the 🟢 budget, two of them warnings regraded after verification: `modelsPaired` on the live branch, which `model-settings-panel.tsx:117` reads only when the origin is a snapshot, and the placement of `pinEnv` on the discovery spec, which is a layering preference on a working diff._

### Verdict
APPROVE — every blocking finding of the last review is fixed at its cause, and no new blocking finding survives verification.


---

Issue #2773, opened 2026-07-13
UI - Design CPU / Memory (agent Size) feature


jenna-winkler:
## Description

Design the UI for setting an agent's Size (its CPU and memory) at sandbox create and in sandbox settings, within the constraint of the user's Budget ceiling. Goal is a clear, low-friction way to pick a Size and understand how it fits against remaining budget headroom.

## Background

- An agent's Size = its CPU/memory limits. It's set by a slider at create, else the template default, else the chart default (1 CPU / 1Gi).
- A Budget is a per user ceiling on the CPU/memory their running agents can use concurrently. The budget constrains starting an agent, never running one; nothing is ever evicted.
- Creates are never rejected - an over-budget agent lands parked. So the UI's job is to make "you're over - free some room" obvious, not to block the create.
- Resizing an agent always restarts the pod. Growing a Size past the ceiling parks it.

## Scope (design)

* **Choosing Size during agent creation** - define how users select CPU and memory (for example: presets like S/M/L, sliders, or separate CPU/memory controls), what the default selection is, and whether templates can provide default sizes.
* **Changing Size after creation** - define how users edit an existing agent’s Size in sandbox settings and clearly communicate that changing Size will restart the pod.
* **Showing budget impact while choosing Size** - help users understand how their selection affects their available capacity by showing current usage, reserved capacity, and remaining capacity alongside the Size options.
* **Handling cases where Size exceeds available budget**
  * **Create flow:** allow the agent to be created but place it in a parked state if there is not enough capacity. Design the parked experience and provide a clear path to free capacity (for example, stopping or hibernating another agent).
  * **Settings flow:** prevent users from saving a Size increase that exceeds their budget. Show an inline error explaining the requested size and available capacity.
* **Empty and edge states** - define the experience for first-time users, users with plenty of available capacity (the budget UI should stay out of the way), and users approaching their limit.

## Open questions

* Should Size selection use named presets (S/M/L), sliders, manual CPU/memory inputs, or a combination?
* Should budget information always be visible during Size selection, or only when the user is close to their limit?
* Where should parked agents appear, and what is the fastest path for a user to move an agent from parked back to running?

## Screen captures

<img width="1506" height="425" alt="Image" src="https://github.com/user-attachments/assets/da2f7cfa-8c2e-4a9a-9046-64c2930892f3" />

[...]

PR #3573, opened 2026-09-04
feat(ui): agent Size as slot presets with a budget meter

JanPokorny:
Refs #2773 — delivers the settings, list and dialog surfaces; Size at create stays open on the issue (see below).

Implements the [approved Figma design](https://www.figma.com/design/zNIYydUKN1QLZDYozpQJpn/DAM-DEV?node-id=1705-2&p=f&t=wHm7YnecYTHeo1Ef-0) for agent Size. Users never touch CPU or memory directly any more: a **slot** is the chart default agent Size (1 CPU / 2Gi by default), an agent's Size is a multiple of it (1x, 2x, 4x), and the Budget is shown as `used/ceiling slots`. I took a bit of liberty with some details to preserve unified look across the whole app.

## What changed

- **Size picker** in sandbox settings replaces the CPU/memory sliders with a `Compute resources` select offering `1x · 1 CPU · 2 Gi`, `2x · 2 CPU · 4 Gi`, `4x · 4 CPU · 8 Gi`. A legacy Size outside the presets stays selectable as its own multiple, so saving never silently changes it. The help tooltip carries the design copy and says whether saving restarts the agent or applies at next start; the save itself still asks for confirmation before a restart. An inline warning appears when the pick needs more free slots than remain, worded differently for a running agent (save is refused by the server) and a sleeping one (it will park). The 128Mi floor from #3069 disappears by construction.
- **Compute card** on the Coding agents page and the Home widget now render one shared component: `Compute resources` + info icon, `N/M slots`, a bar with one cell per slot, a running/awake legend, and the operator's "Request more budget" link. Hovering any cell of an agent highlights all of that agent's cells and shows a single tooltip centered over the group. Hidden while the user has no coding agents.
- **Row subtitle** carries the multiple, e.g. `Claude Code · Anthropic · 2x · 2 CPU · 4 Gi`.
- **Over-budget dialog** with the design's copy, `Cancel` / `Manage sandboxes`, raised when a start the user initiated lands parked. It does not fire when retrying an already parked agent, because the list refetch can still carry the previous denial before the controller rules again.
- **Server:** the budget endpoint publishes the slot unit next to reserved and ceiling. Enforcement is unchanged and stays in raw CPU and memory.
- **Docs:** `budgets.md` and the glossary describe slots.

Slot math: an agent takes as many slots as its larger dimension needs, rounded up; the ceiling holds as many as its tighter dimension allows, rounded down.

## Screenshots

Coding agents page with the compute card and sized rows:

Coding agents list with compute card

Hovering one cell highlights the whole agent and shows one tooltip:

Bar hover groups an agent's cells

Home widget, same component:

Home compute widget

Settings: size picker with help tooltip:

Settings size picker tooltip

Settings: picking a size that does not fit while the agent is running:

Settings inline warning

Starting a 4x agent under a 2-slot ceiling parks it and raises the dialog:

Over budget dialog

The parked agent afterwards:

Parked agent in the list

Screenshots live on the `screenshots/agent-size-slots` branch; delete it once this PR is merged or after replacing the images with drag-and-drop uploads.

## Verified

- UI tsc/lint/format, UI unit tests (340), api-server unit tests (1305), api-server-api checks, comment-type check, docs size cap.
- On a local k3s cluster: create, resize a running agent (restart confirmation), server refusal of an over-ceiling grow, resize of a sleeping agent, budget reclaim of an idle agent on wake, a parked start under a lowered `UserBudget` with the dialog appearing within seconds, and the Home widget.

## Left out on purpose

- **Size at create.** The Figma export has no create-flow screen, so create keeps the template default; users pick a bigger size in settings before first start, which needs no restart. The issue stays open for that bullet.
- **"Awake" row badge.** The design shows an Awake pill on rows; that is a lifecycle-status change outside this issue. The compute card does distinguish running from awake.
- **Parked display.** A parked ordinary agent still shows the `Over budget` badge until its idle window lapses. Collapsing it to `Hibernating` after the dialog was discussed and deferred.


packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx
```diff
@@ -1,157 +1,96 @@
+import { Help } from "@carbon/icons-react";
```
dam-code-guardian, line 1: 🔴 **Critical:** The linked issue's first scope bullet — "Choosing Size during agent creation … what the default selection is, and whether templates can provide default sizes" — has no corresponding change. The Size picker exists only in sandbox settings, and the create surfaces (`coding-agent-setup-view.tsx`, `create-agent-inline.tsx`) are untouched, so `Closes #2773` closes the issue with a named deliverable undone. The PR body declares the omission under "Left out on purpose" citing the Figma export, but the issue text, not the design export, defines the ask.
  **Fix:** A closing keyword may only ship when every scope bullet of the linked issue is delivered; when a deliverable is dropped because the design input does not cover it, re-scope the issue in discussion, or drop the closing keyword and leave the remainder tracked.

packages/ui/src/modules/agents/store.ts
```diff
@@ -4,17 +4,18 @@ import type { PlatformStore } from "../../store.js";
 import type { AgentView } from "../../types.js";
 import { routeToPath } from "../platform/lib/routes.js";
 
+export interface RestartingEntry {
+  seenNonRunning: boolean;
+  clickedAt: number;
+  parkedAtClick?: boolean;
```
dam-code-guardian: 🟡 **Warning:** `parkedAtClick` is optional and the reader discriminates `undefined` from `false` via `parkedAtClick === false`, while `useRestartAgent.restart` still writes `{ seenNonRunning, clickedAt }` without it. An agent parked after a *restart* click can never satisfy the predicate, so the over-budget dialog never fires for that path — and the same gap leaves an over-budget *create* with no dialog. An optional field models presence as a wiring question, so a missed writer is indistinguishable from intent.
  **Fix:** Declare every field a consumer branches on as required, so each writer must supply it and a missed call site is a compile error; the writer to update is `use-restart-agent.ts:14`, the reader is `use-restart-agent.ts:39`.

packages/ui/src/modules/budgets/lib/slots.ts
```diff
@@ -0,0 +1,194 @@
...
+
+export function sizeMultiplier(size: SizeMi, unit: SlotUnit): number {
+  const ratio = Math.max(
+    size.cpuMilli / unit.cpuMilli,
+    size.memoryMi / unit.memoryMi,
+  );
+  return ratio > 0 ? ratio : 1;
+}
+
+export function slotsFor(size: SizeMi, unit: SlotUnit): number {
```
dam-code-guardian, line 50: 🟡 **Warning:** The budget-admission rule now has a second, differently-rounding implementation in the client. `slotsFor`/`ceilingSlots`/`freeSlots` round each agent up to whole slots and floor the ceiling, while the server's `createResizeGate` admits on the exact raw cpu/memory sum. For a non-multiple Size the two disagree — a 1.5-CPU agent against a 1-CPU slot counts as 2 slots in the client and 1.5 on the server — so the settings picker can report no free capacity for a Size the server accepts.
  **Fix:** An invariant is enforced in one place and every other surface derives its display from that authority's numbers; never re-derive an admission rule with different rounding on the client. See also `slots.ts:54`, `slots.ts:64` against `packages/api-server/src/modules/budgets/services/budgets-service.ts:33-62`.

packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx
```diff
@@ -1,157 +1,96 @@
...
-  templateSize,
   sizeCpuMilli,
   sizeMemoryMi,
   onChange,
   disabled,
-  restartNote,
   currentSize,
 }: Props) {
   const { data: budget } = useBudgetReserved();
+  if (!budget) return null;
```
dam-code-guardian: 🟡 **Warning:** While `useBudgetReserved` is pending or has errored, this early return removes the whole Compute-resources control from the sandbox settings form, with no skeleton and no message, yet the form still holds and submits `sizeCpuMilli`/`sizeMemoryMi`. The previous version always rendered, using explicit ceiling fallbacks. `compute-usage.tsx:41` has the same guard, where an absent card is acceptable.
  **Fix:** A form field whose value is still submitted is never conditionally unmounted on a query state — render a disabled, loading or error variant so pending and failed server data stay visible.

packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx
```diff
@@ -1,157 +1,96 @@
...
-          min={MEMORY_FLOOR_MI}
-          max={memoryMax}
-          step={MEMORY_STEP_MI}
-          over={memoryOver}
-          disabled={disabled}
-          onChange={(v) => onChange({ sizeMemoryMi: v })}
-        />
+      <div className="mb-3 flex items-center gap-1.5">
+        <SectionLabel>Compute resources</SectionLabel>
+        <HintTooltip
```
dam-code-guardian, line 60: 🟡 **Warning:** The `restartNote` prop was deleted from this component and from its only caller, and its two conditional messages became one unconditional tooltip sentence, "Changing the size of an existing sandbox restarts it". That is false for the hibernated and over-budget cases that `sizeRestartsAgent` still distinguishes (`use-sandbox-settings-form.ts:211`), where the old copy correctly said the new size applies at next start. The consequence of saving is now also behind a hover target.
  **Fix:** When a condition that changes what a save does still exists in the form state, keep branching on it in always-visible copy — a consequence of submitting is never flattened into one unconditional claim, nor moved into a tooltip.

packages/ui/src/modules/agents/utils/sandbox-subtitle.ts
```diff
@@ -1,11 +1,17 @@
...
+  formatSizeLabel,
+  sizeInMi,
+  type SlotUnit,
+} from "../../budgets/lib/slots.js";
 import { kbTemplateName } from "../../knowledge-bases/lib/kb-templates.js";
 
 export interface SandboxSubtitleLookup {
   templateNameById: ReadonlyMap<string, string>;
   connectionTemplateIdById: ReadonlyMap<string, string>;
+  slotUnit?: SlotUnit;
```
dam-code-guardian: 🟡 **Warning:** `slotUnit` is optional, so the consumer branches on presence (`if (!lookup.slotUnit) return null` at `:63`) while the producer passes `budget ? slotUnitOf(budget) : undefined` (`use-agent-rows.ts:70`). A missing wiring reads identically to "budget still loading", and the size label disappears from the subtitle with no signal.
  **Fix:** Never encode a load state as an optional dependency on a pure formatter; make the dependency required and let the caller decide what to render while data is absent.

packages/ui/src/modules/budgets/lib/slots.ts
```diff
@@ -0,0 +1,194 @@
+import type { BudgetReserved } from "api-server-api";
+
+import type { AgentView } from "../../../types.js";
+import { parseCpuMilli, parseMemoryMi } from "../../sandboxes/lib/quantity.js";
```
dam-code-guardian, line 4: 🟡 **Warning:** The diff adds three cross-module import edges that bypass any module public API: this import of `../../sandboxes/lib/quantity.js`, and `agents/utils/sandbox-subtitle.ts:4-8` plus `agents/hooks/use-agent-rows.ts:5-6` reaching into `../../budgets/lib/slots.js` and `../../budgets/api/queries.js`. No UI module defines an `index.ts`, so the PR amplifies the existing boundary gap rather than creating it.
  **Fix:** Every module exposes one `index.ts` public API and all cross-module imports go through it, never a deep path into another module's `lib/`, `api/` or `services/`.

docs/architecture/budgets.md
```diff
@@ -47,6 +47,8 @@ A denial is **remembered per wake attempt** (keyed by the `last-activity` value
 
 Creates are never rejected. An over-budget create simply lands parked, and the UI's job is to make "free some room" obvious. This also means the whole enforcement path is invisible on the happy path: under-ceiling users never touch it.
 
+**Slots** are how the UI presents all of this. One slot is the chart default Size, and the budget endpoint publishes that unit next to Reserved and Ceiling. An agent occupies as many slots as its larger dimension needs, rounded up; the Ceiling holds as many slots as its tighter dimension allows, rounded down. Users never touch CPU or memory directly: the Size picker offers multiples of the slot (1x, 2x, 4x — a legacy Size outside the presets stays selectable as its own multiple), each agent row carries its multiple, and the agent list shows a slot meter — one cell per slot, colored working / awake / free — with the operator's "request more budget" link beside it. A start the controller parks raises a dialog that names the cause and sends the user to that list to free room; the settings picker warns inline when the chosen Size needs more free slots than remain. First-time users with no agents see none of it.
```
dam-code-guardian: 🟡 **Warning:** The new Slots paragraph transcribes code-level literals into an architecture page: the exact preset multipliers `1x, 2x, 4x` (the `SIZE_MULTIPLIERS` const in `packages/ui/src/modules/budgets/lib/slots.ts`) and the literal UI link label "request more budget". The same paragraph enumerates where slots surface — the Size picker, the row subtitle, the agent list meter — but the PR also re-bases the home dashboard widget from CPU cores to the same slot meter (`compute-widget.tsx` now renders `ComputeUsage`, `home/lib/compute-cells.ts` is deleted), so a reader learns that CPU is still the home-page unit.
  **Fix:** Architecture pages name the concept, never the literal — describe the picker as offering a small fixed set of slot multiples and link out to the UI budgets module for the values and copy. When a change re-bases a subsystem's user-facing unit, name every surface that unit now appears on, or describe them as a class.

packages/ui/src/modules/budgets/components/compute-usage.tsx
```diff
@@ -0,0 +1,109 @@
...
+  return `${segment.agentName} · ${formatSizeLabel(
+    { cpuMilli: segment.cpuMilli, memoryMi: segment.memoryMi },
+    unit,
+  )}`;
+}
+
+export function ComputeUsage({ agents }: { agents: readonly AgentView[] }) {
+  const { data: budget } = useBudgetReserved();
+  const { data: links } = useLinks();
+  const { items } = useFeed();
```
dam-code-guardian: 🟡 **Warning:** `ComputeUsage` calls `useFeed()` only to derive the set of working agent ids, and `useFeed` is the home feed's aggregate hook — it fans out one ACP `listAgentSessions` query per running agent plus a polling owner-approvals query. Mounting the card on the coding-agents page starts the whole activity-feed network pipeline on a page that shows no feed, and on the home page `useFeed` now runs in two components that each recompute `toFeedItems`. The previous widget took `workingAgentIds` as a prop and fetched nothing.
  **Fix:** A presentational widget never reaches for another module's aggregate data hook to obtain one derived value — server-owned inputs it does not own are passed in as props by the view that already holds them.

packages/ui/src/modules/budgets/components/compute-usage.tsx
```diff
@@ -0,0 +1,109 @@
...
+          {view.groups.length === 0 && "No agent is holding compute."}
+          {view.groups.map((group) => (
+            <span key={group.state} className="flex items-center gap-1.5">
+              <span
+                className={cn(
+                  "inline-block size-2 shrink-0 rounded-sm",
+                  STATE_DOT[group.state],
+                )}
+              />
+              {group.agents} {group.state}
```
dam-code-guardian: 🟡 **Warning:** The legend renders `{group.agents} {group.state}`, printing the internal union members into user-visible copy ("2 running", "1 awake") where the deleted widget mapped them through `STATE_LABEL` to "Working"/"Awake". It also drops the agent/agents pluralisation and the per-group `CPU · Gi` breakdown that `ComputeGroup` still computes, so `groups[].cpuMilli`/`memoryMi` are now dead payload.
  **Fix:** State-machine identifiers are never rendered as display text — every user-facing label for a union goes through an explicit `Record<State, string>` label map, and a field no longer displayed is removed from the view model rather than left computed.

packages/ui/src/modules/budgets/components/slot-bar.tsx
```diff
@@ -0,0 +1,53 @@
...
+export function SlotBar({ segments, totalSlots, label, ariaLabel }: Props) {
+  return (
+    <div
+      role="group"
+      aria-label={ariaLabel}
+      className="grid gap-1.5"
+      style={{ gridTemplateColumns: `repeat(${totalSlots}, minmax(0, 1fr))` }}
+    >
+      {segments.map((segment, index) => (
+        <Tooltip key={index} content={label(segment)} side="top">
```
dam-code-guardian: 🟡 **Warning:** Each `Tooltip` wraps a bare `<span>` via Radix `Trigger asChild`, so the trigger has no `tabIndex` and no role. The tooltip is the only place that names which agent holds which slots, and it is unreachable by keyboard, while `aria-label` on a generic `<span>` is not reliably exposed to assistive tech — the per-agent attribution is mouse-only.
  **Fix:** Information that exists only in a tooltip must have a focusable trigger (a real `button` or `tabIndex={0}` element with an accessible role) or a non-hover textual equivalent.

🟢 Also: segments are keyed by array index although each carries a stable `segment.agentId`, and `computeView` re-orders segments (running before awake) whenever an agent starts or stops working — so a segment's identity, with its tooltip state and hover ring transition, moves to a different agent on reorder.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `d018113`

## PR #3573: feat(ui): agent Size as slot presets with a budget meter
**Author:** JanPokorny | **Branch:** feat/agent-size-slots → main | **Changes:** +636 −442 (25 files)

### Summary
The PR replaces the CPU/memory sliders with slot presets. A slot is the chart default Size (1 CPU / 2Gi), an agent Size is a multiple of it, and the Budget shows as `used/ceiling slots`. New client arithmetic lives in `packages/ui/src/modules/budgets/lib/slots.ts`, the budget endpoint publishes the slot unit, and one shared compute card replaces the home widget. Enforcement stays in raw CPU and memory on the server.

### Findings
- 🟡 **Warning:** over-budget dialog never fires on the restart path — full text in **TypeScript Engineering** and **Issue Fit** below (`packages/ui/src/modules/agents/hooks/use-restart-agent.ts:39`)
  **Fix:** make a field every consumer branches on required, so each writer must supply it.
- 🟡 **Warning:** client slot rounding can warn about a Size the server accepts — full text in **TypeScript Engineering** below (`packages/ui/src/modules/budgets/lib/slots.ts:50`)
  **Fix:** enforce an admission rule in one place and derive every display from that authority.
- 🟡 **Warning:** the Size field disappears while the budget query is pending — full text in **React UI Engineering** below (`packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx:38`)
  **Fix:** never unmount a form field whose value is still submitted on a query state.
- 🟡 **Warning:** the tooltip claims a restart for a hibernated or parked agent — full text in **React UI Engineering** below (`packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx:61`)
  **Fix:** keep branching on a condition that changes what a save does, in always-visible copy.
- 🔴 **Critical:** `Closes #2773` while the issue's create-flow Size selection is undone — full text in **Issue Fit** below (`packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx`)
  **Fix:** a closing keyword ships only when every scope bullet of the linked issue is delivered.
- 🟡 **Warning:** the settings save is not prevented for an over-budget Size, only warned about — full text in **Issue Fit** below (`packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx:85`)
  **Fix:** gate the submit affordance on the same condition that produces the inline message.
- 🟡 **Warning:** `docs/architecture/budgets.md` carries code literals and omits the home widget — full text in **Documentation Drift** below (`docs/architecture/budgets.md:50`)
  **Fix:** name the concept, not the literal, and enumerate every surface a re-based unit appears on.
- 🟡 **Warning:** the slot bar tooltip is the only per-agent attribution and it is not keyboard reachable — full text in **React UI Engineering** below (`packages/ui/src/modules/budgets/components/slot-bar.tsx:28`)
  **Fix:** give tooltip-only information a focusable trigger or a textual equivalent.
- ✅ **Looks good:** both server app wirings (`trpc/context.ts`, harness `app.ts`) supply `slotSize` identically, so the two roots cannot drift.
- ✅ **Looks good:** `slots.test.ts` covers the binding dimension for the ceiling, the own-reservation add-back, and running agents above the ceiling.
- ✅ **Looks good:** the diff adds no code comments, so the repository convention holds.

### Documentation Drift
- 🟡 **Warning:** The new Slots paragraph transcribes code-level literals into an architecture page: the exact preset multipliers `1x, 2x, 4x` (the `SIZE_MULTIPLIERS` const in `packages/ui/src/modules/budgets/lib/slots.ts`) and the literal UI link label `"request more budget"`. The same paragraph enumerates where slots surface — the Size picker, the row subtitle, the agent list meter — but the PR also re-bases the home dashboard widget from CPU cores to the same slot meter (`compute-widget.tsx` now renders `ComputeUsage`, `home/lib/compute-cells.ts` is deleted), so a reader learns that CPU is still the home-page unit (`docs/architecture/budgets.md:50`).
  **Fix:** Architecture pages name the concept, never the literal — describe the picker as offering a small fixed set of slot multiples and link out to the UI budgets module for the values and copy. When a change re-bases a subsystem's user-facing unit, name every surface that unit now appears on, or describe them as a class.

### Issue Fit
Linked issue: #2773 "UI - Design CPU / Memory (agent Size) feature" (closing keyword)

- 🔴 **Critical:** The issue's first scope bullet — "Choosing Size during agent creation … what the default selection is, and whether templates can provide default sizes" — has no corresponding change: the Size picker exists only in sandbox settings, and the create surfaces (`coding-agent-setup-view.tsx`, `create-agent-inline.tsx`) are untouched, so `Closes #2773` closes the issue with a named deliverable undone. The PR body declares the omission under "Left out on purpose" citing the Figma export, but the issue text, not the design export, defines the ask (`packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx:1`).
  **Fix:** A closing keyword may only ship when every scope bullet of the linked issue is delivered; when a deliverable is dropped because the design input does not cover it, re-scope the issue in discussion, or drop the closing keyword and leave the remainder tracked.
- 🟡 **Warning:** The issue's Settings flow requires "prevent users from saving a Size increase that exceeds their budget"; the PR renders an inline warning and leaves save enabled — `isSubmitDisabled` is `saving || !formReady || !dirty || selectedProvider === null` with no over-budget term, and the PR body concedes "save is refused by the server". The user learns of the refusal only after submitting (`packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-form.ts:157`).
  **Fix:** When an issue asks that an action be prevented, gate the submit affordance on the same condition that produces the inline message; a server-side rejection plus a warning is a diagnostic, not prevention.
- 🟡 **Warning:** The issue's Create flow bullet asks to "place it in a parked state … Design the parked experience and provide a clear path to free capacity", but the over-budget dialog is raised only from the wake sync path — `parkedAtClick` is set exclusively in `use-wake-agent.ts`, so an over-budget *create* lands parked with no dialog — and the PR defers the parked row display, so the parked experience itself is unchanged (`packages/ui/src/modules/agents/hooks/use-restart-agent.ts:39`).
  **Fix:** Cover every entry point the issue names for a state: if a condition is reachable by create as well as by start or wake, raise the new affordance from the shared observation of that state, not from one caller's click bookkeeping.
- 🟢 **Suggestion:** The issue's third Open Question — "Where should parked agents appear, and what is the fastest path for a user to move an agent from parked back to running?" — ships unanswered: the added docs paragraph describes slots and the dialog's link to the agent list, but the parked presentation was deferred (`docs/architecture/budgets.md:50`).

### TypeScript Engineering
- 🟡 **Warning:** `RestartingEntry.parkedAtClick` is optional and the reader discriminates `undefined` from `false` via `parkedAtClick === false`, while `useRestartAgent.restart` still writes `{ seenNonRunning, clickedAt }` without it — so an agent parked after a *restart* click can never satisfy the predicate and the over-budget dialog never fires for that path. An optional field models presence as a wiring question, so a missed writer is indistinguishable from intent (`packages/ui/src/modules/agents/store.ts:10`).
  **Fix:** Declare every field a consumer branches on as required, so each writer must supply it and a missed call site is a compile error; the writers to update are `use-restart-agent.ts:14` and the reader at `use-restart-agent.ts:39`.
- 🟡 **Warning:** `SandboxSubtitleLookup.slotUnit` is optional, so the consumer branches on presence (`if (!lookup.slotUnit) return null`) while the producer passes `budget ? slotUnitOf(budget) : undefined` — a missing wiring reads identically to "budget still loading" and the size label disappears from the subtitle with no signal (`packages/ui/src/modules/agents/utils/sandbox-subtitle.ts:14`).
  **Fix:** Never encode a load state as an optional dependency on a pure formatter; make the dependency required and let the caller decide what to render while data is absent — same pattern at `sandbox-subtitle.ts:63` and `use-agent-rows.ts:70`.
- 🟡 **Warning:** The diff adds three cross-module import edges that bypass any module public API: `budgets/lib/slots.ts` imports `../../sandboxes/lib/quantity.js`, and `agents/utils/sandbox-subtitle.ts` plus `agents/hooks/use-agent-rows.ts` import `../../budgets/lib/slots.js` and `../../budgets/api/queries.js`. No UI module defines an `index.ts`, so the PR amplifies the existing boundary gap (`packages/ui/src/modules/budgets/lib/slots.ts:4`).
  **Fix:** Every module exposes one `index.ts` public API and all cross-module imports go through it, never a deep path into another module's `lib/`, `api/` or `services/`; the other two edges are `sandbox-subtitle.ts:4-8` and `use-agent-rows.ts:5-6`.
- 🟡 **Warning:** The budget-admission rule now has a second, differently-rounding implementation in the client: `slotsFor`/`ceilingSlots`/`freeSlots` round each agent up to whole slots and floor the ceiling, while the server's `createResizeGate` admits on the exact raw cpu/memory sum. For a non-multiple Size the two disagree — a 1.5-CPU agent against a 1-CPU slot counts as 2 slots in the client and 1.5 on the server — so the settings picker can report no free capacity for a Size the server accepts (`packages/ui/src/modules/budgets/lib/slots.ts:50`).
  **Fix:** An invariant is enforced in one place and every other surface derives its display from that authority's numbers; never re-derive an admission rule with different rounding on the client — see also `slots.ts:54`, `slots.ts:64` against `packages/api-server/src/modules/budgets/services/budgets-service.ts:33-62`.
- 🟢 **Suggestion:** `createResizeGate` types its dependencies by subtraction from a sibling's interface (`Omit<BudgetsServiceDeps, "slotSize">`) to opt out of a dependency it does not use, so a future field added for one factory silently re-shapes the other's contract; `createSpawnSizeGate` uses the same `Pick`-based shortcut at `:79` (`packages/api-server/src/modules/budgets/services/budgets-service.ts:25`).
- 🟢 **Suggestion:** The harness composition root passes `{ cpu: config.agentDefaultCpuLimit, memory: config.agentDefaultMemoryLimit }` as `slotSize` into `composeBudgetsModule` and, four lines later, the identical pair as `defaultLimits`; `context.ts:99-102` repeats the literal again. The composition root should wire each raw dependency once under one name (`packages/api-server/src/apps/harness-api-server/app.ts:220`).
- 🟢 **Suggestion:** `BudgetReserved` describes what is currently reserved against a ceiling, but `slot: { cpuMilli, memoryBytes }` is deployment configuration that does not vary with reservations, so `reserved()` returns two unrelated kinds of fact and every consumer re-reads static config through the reservation read model (`packages/api-server-api/src/modules/budgets/types.ts:4`).

### React UI Engineering
- 🟡 **Warning:** `ComputeUsage` calls `useFeed()` only to derive the set of working agent ids, and `useFeed` is the home feed's aggregate hook — it fans out one ACP `listAgentSessions` query per running agent plus a polling owner-approvals query. Mounting the card on the coding-agents page starts the whole activity-feed network pipeline on a page that shows no feed, and on the home page `useFeed` now runs in two components that each recompute `toFeedItems`; the previous widget took `workingAgentIds` as a prop and fetched nothing (`packages/ui/src/modules/budgets/components/compute-usage.tsx:40`).
  **Fix:** A presentational widget never reaches for another module's aggregate data hook to obtain one derived value — server-owned inputs it does not own are passed in as props by the view that already holds them.
- 🟡 **Warning:** The legend renders `{group.agents} {group.state}`, printing the internal union members into user-visible copy ("2 running", "1 awake") where the deleted widget mapped them through `STATE_LABEL` to "Working"/"Awake", and it drops both the agent/agents pluralisation and the per-group `CPU · Gi` breakdown that `ComputeGroup` still computes — so `groups[].cpuMilli`/`memoryMi` are now dead payload (`packages/ui/src/modules/budgets/components/compute-usage.tsx:87`).
  **Fix:** State-machine identifiers are never rendered as display text — every user-facing label for a union goes through an explicit `Record<State, string>` label map, and a field no longer displayed is removed from the view model rather than left computed.
- 🟡 **Warning:** In `SlotBar`, each `Tooltip` wraps a bare `<span>` via Radix `Trigger asChild`, so the trigger has no `tabIndex` and no role: the tooltip is the only place that names which agent holds which slots, and it is unreachable by keyboard, while `aria-label` on a generic `<span>` is not reliably exposed to assistive tech — the per-agent attribution is mouse-only (`packages/ui/src/modules/budgets/components/slot-bar.tsx:28`).
  **Fix:** Information that exists only in a tooltip must have a focusable trigger (a real `button` or `tabIndex={0}` element with an accessible role) or a non-hover textual equivalent.
- 🟡 **Warning:** `SandboxSizeSection` now does `if (!budget) return null`, so while `useBudgetReserved` is pending or has errored the whole Compute-resources control disappears from the sandbox settings form with no skeleton and no message, yet the form still holds and submits `sizeCpuMilli`/`sizeMemoryMi`; the previous version always rendered, using explicit ceiling fallbacks (`packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx:38`).
  **Fix:** A form field whose value is still submitted is never conditionally unmounted on a query state — render a disabled, loading or error variant so pending and failed server data stay visible.
- 🟡 **Warning:** The `restartNote` prop was deleted from `SandboxSizeSection` and from its only caller, and its two conditional messages were replaced by one unconditional tooltip sentence, "Changing the size of an existing sandbox restarts it" — which is false for the hibernated and over-budget cases that `sizeRestartsAgent` still distinguishes, where the old copy correctly said the new size applies at next start. The consequence of saving is also now behind a hover target (`packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx:61`, `packages/ui/src/modules/sandboxes/components/sandbox-setup-section.tsx:52`).
  **Fix:** When a condition that changes what a save does still exists in the form state, keep branching on it in always-visible copy — a consequence of submitting is never flattened into one unconditional claim, nor moved into a tooltip.
- 🟢 **Suggestion:** `SlotBar` keys segments by array index although each carries a stable `segment.agentId`, and `computeView` re-orders segments (running before awake) whenever an agent starts or stops working — so a segment's identity, and with it the tooltip state and the hover ring transition, moves to a different agent on reorder (`packages/ui/src/modules/budgets/components/slot-bar.tsx:28`).
- 🟢 **Suggestion:** The size `<select>` encodes each option as the string `"<cpuMilli>:<memoryMi>"` and parses it back with `split(":").map(Number)` plus two non-null assertions, pushing a typed pair through the DOM as a stringly-typed key the compiler cannot check; keying options by index into `options` keeps the typed value and removes both `!` (`packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx:74`).
- 🟢 **Suggestion:** `ComputeUsage` rebuilds the working-id `Set`, re-filters `agents`, and re-runs `computeView` over every agent on each render with no `useMemo`, and the hook now runs on two pages whose parents re-render on feed polling (`packages/ui/src/modules/budgets/components/compute-usage.tsx:44`).
- 🟢 **Suggestion:** One widget now has three wrappers: `ComputeUsage`, `ComputeUsageCard` (a `Card` with `border p-4`), and `ComputeWidget`, which after the refactor is a pass-through `div` with different chrome (`rounded-2xl border p-6`) — so the same widget renders with two containers on two pages and `compute-widget.tsx` adds an indirection with no behaviour (`packages/ui/src/modules/home/components/compute-widget.tsx:4`).
- 🟢 **Suggestion:** The card is gated on `codingAgents.length > 0` but is fed `agentsData?.list ?? []`, i.e. every agent — a user whose compute is held only by non-coding or temporary sandboxes sees no usage card at all, and the inline `?? []` yields a fresh array identity on each render (`packages/ui/src/modules/agents/views/coding-agents-view.tsx:48`).

### Verdict
REQUEST_CHANGES — the slot model and its tests are sound, but `Closes #2773` ships with the issue's create-flow Size selection undone, and four verified warnings affect the dialog, the picker and its copy.


---

packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx
JanPokorny: Agreed on the keyword. Changed to `Refs #2773`; the issue stays open for Size at create, which has no design input yet.

packages/ui/src/modules/agents/store.ts
JanPokorny: Fixed in 2efbf3d8: `parkedAtClick` is now required and the restart path sets it too. The compiler then also surfaced a test fixture and one lookup site that were missing the field.

packages/ui/src/modules/budgets/lib/slots.ts
JanPokorny: Intentional. The slot model rounds by design (a 1.5-CPU legacy Size occupies 2 slots because the budget is presented in whole slots), and the client warning is advisory: the server stays the only admission authority and refuses at save. The inline copy says "slots" so it never claims to be the server's arithmetic.

packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx
JanPokorny: Fixed in 2efbf3d8: the select stays mounted and disabled while the budget loads, showing the current Size as its only option.

packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx
JanPokorny: Fixed in 2efbf3d8: the tooltip now branches on whether the agent is up ("restarts the sandbox on save" vs "applies when the sandbox next starts"). The always-visible part is the existing save flow, which asks "Restart agent?" before a resize that restarts.

packages/ui/src/modules/agents/utils/sandbox-subtitle.ts
JanPokorny: Fixed in 2efbf3d8: `slotUnit` is required and explicitly `null` while the budget is loading.

packages/ui/src/modules/budgets/lib/slots.ts
JanPokorny: Not changing here. No UI module has an `index.ts` and there are ~640 existing deep cross-module imports; introducing barrels for one PR would be inconsistent with the rest of the package. Worth a repo-wide decision, not a side effect of this change.

docs/architecture/budgets.md
JanPokorny: Fixed in 2efbf3d8: the paragraph names "a small fixed set of slot multiples", drops the link label, and lists the home dashboard as a surface.

packages/ui/src/modules/budgets/components/compute-usage.tsx
JanPokorny: Fixed in 2efbf3d8: `ComputeUsage` takes `workingAgentIds` as a prop. `useFeed` exposes that set once, home passes it from the feed it already holds, and the Coding agents view calls the hook itself since the working/awake split is what the design asks for on that page.

packages/ui/src/modules/budgets/components/compute-usage.tsx
JanPokorny: "2 running" / "1 awake" is the design's literal legend copy (DAM-1 in the Figma export), so the union members are the labels on purpose. The dead `cpuMilli`/`memoryMi` on `ComputeGroup` are removed in 2efbf3d8.

packages/ui/src/modules/budgets/components/slot-bar.tsx
JanPokorny: Fixed in 2efbf3d8: held segments are focusable (`tabIndex=0`, `role=img` with the label) and show the same ring on focus-visible, so the tooltip opens from the keyboard. Segments are keyed by agent id.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `2efbf3d`

## PR #3573: feat(ui): agent Size as slot presets with a budget meter
**Author:** JanPokorny | **Branch:** feat/agent-size-slots → main | **Changes:** +667 −453 (28 files)

### Summary
The fix round answers nine of the twelve blocking findings at their anchors, and the PR body now says `Refs #2773` with an explicit "Left out on purpose" list. Three blocking findings from the previous round are unfixed: the client slot arithmetic, the unprevented over-budget save, and the state union printed as copy. _17 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: d018113 (2026-09-04T18:10:35Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** Closes #2773 with create-flow Size undone (`packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx:1`)
- ✅ **Fixed:** optional parkedAtClick, dialog never fires on restart (`packages/ui/src/modules/agents/store.ts:10`)
- ✅ **Fixed:** Size field unmounted while budget query pending (`packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx:38`)
- ✅ **Fixed:** tooltip claims restart for hibernated or parked agent (`packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx:61`)
- ✅ **Fixed:** optional slotUnit hides a missing wiring (`packages/ui/src/modules/agents/utils/sandbox-subtitle.ts:14`)
- ✅ **Fixed:** docs carry code literals and omit the home widget (`docs/architecture/budgets.md:50`)
- ✅ **Fixed:** tooltip-only agent attribution is not keyboard reachable (`packages/ui/src/modules/budgets/components/slot-bar.tsx:28`)
- ✅ **Fixed:** segments keyed by index although agentId is stable (`packages/ui/src/modules/budgets/components/slot-bar.tsx:28`)
- ✅ **Fixed:** three cross-module imports bypass the public API — I withdraw this one. `packages/ui` has no module `index.ts` convention, and deep cross-module imports are the package's own practice (`packages/ui/src/modules/budgets/lib/slots.ts:4`)
- 🔁 **Still present:** client re-derives admission with different rounding; `usedSlots` and `ceilingSlots` are also paired as one fraction from two different dimensions (`packages/ui/src/modules/budgets/lib/slots.ts:50`, also `:54`)
- 🔁 **Still present:** over-budget save warned about, not prevented (`packages/ui/src/modules/sandboxes/hooks/use-sandbox-settings-form.ts:157`)
- 🔁 **Still present:** state union members rendered as user-visible copy; this round deleted the `STATE_LABEL` map that translated them (`packages/ui/src/modules/budgets/components/compute-usage.tsx:88`)
- 🔁 **Still present:** the feed pipeline still loads for one derived value, now from the view — my previous **Fix:** line asked for this move, so it is a suggestion now (`packages/ui/src/modules/budgets/components/compute-usage.tsx:40`)
- 🔁 **Still present:** typed size pair pushed through the DOM as a string key (`packages/ui/src/modules/sandboxes/components/sandbox-size-section.tsx:74`)
- 🔁 **Still present:** computeView re-runs every render with no useMemo (`packages/ui/src/modules/budgets/components/compute-usage.tsx:47`)
- 🔁 **Still present:** card gated on coding agents but fed every agent (`packages/ui/src/modules/agents/views/coding-agents-view.tsx:48`)
- 🔁 **Still present:** two containers for one widget — the deleted shell is inlined at `home-view.tsx:215` (`packages/ui/src/modules/home/components/compute-widget.tsx:4`)
- 🔁 **Still present:** resize gate types deps by Omit from a sibling (`packages/api-server/src/modules/budgets/services/budgets-service.ts:25`)
- 🔁 **Still present:** same config pair wired twice under two names (`packages/api-server/src/apps/harness-api-server/app.ts:220`)
- 🔁 **Still present:** static slot config added to a reservation read model (`packages/api-server-api/src/modules/budgets/types.ts:4`)

### Findings
- 🟢 **Suggestion:** the fix commit edits `docs/architecture/budgets.md` on 2026-09-07, but the page header still reads `Last verified: 2026-09-04`. Every other architecture page carries the date of its last edit (reported by doc-drift) (`docs/architecture/budgets.md:3`)

- 🔁 1 finding from the previous review still present (see review at d018113)

### Issue Fit
- 🔁 1 finding from the previous review still present (see review at d018113)

- 🔁 2 findings from the previous review still present (see review at d018113)

- 🔁 2 findings from the previous review still present (see review at d018113)

### Verdict
REQUEST_CHANGES — three blocking findings from the previous round are unfixed: the client slot arithmetic, the unprevented over-budget save, and the state union printed as copy.


---

JanPokorny:
Second Code Guardian round addressed in f264aeb6:

- **Legend copy** now goes through an explicit label map, and the `Last verified` date on `budgets.md` is bumped.
- **Client slot arithmetic** and **used/ceiling as one fraction** stay as designed. Slots are a whole-unit presentation of the budget for users; the fraction is the design's `N/M slots` meter. Admission is enforced only on the server in raw CPU and memory, and the UI never claims otherwise.
- **Save gating** stays warn-and-refuse. Disabling Save on the client's rounded arithmetic would block saves the server accepts, and for a sleeping agent an over-budget Size is a legitimate save that parks at next start.

PR #3582, opened 2026-09-07
feat(ui): make artifact sharing visible

### LINKED ISSUE #3186: UI - No visible way to share an artifact from the Artifacts tab
## Context

The Artifacts tab lists artifacts as rows. Each row has two ways to share: a link button, and a **Share** item in the row's overflow menu. Both sit behind a hover-reveal — they are invisible until the pointer is over that row. At rest, a row shows only the title, the creating agent, the view count, and a **Private** badge.

Clicking a row opens the artifact preview. That dialog offers **Download**, a version switcher, **Source**, and **Fullscreen**. It offers **Open share page** only when the artifact is already public. There is no control to start sharing.

Reported in [Slack](https://ibm-research.slack.com/archives/C0B8DA70GE4/p1785941603885489).

## Problem

A user who wants to share an artifact finds no control that offers to share it. The controls exist, but the user must move the pointer over the correct row to see them. Nothing on the page tells the user to do this.

The user's own next move makes it worse. To share an artifact, you first open it. The preview dialog has no share control, so the most deliberate path is a dead end. The **Private** badge names the state but does not lead anywhere.

The result: a user concludes that the product cannot share artifacts, when in fact it can. This is the core promise of the artifact library, so it is an expensive thing to hide.

## Goal

A user who wants to share an artifact can see how to do it. This is true both in the list and in the opened artifact — and without hovering to find it.

## User Stories

- As a user, I want a share control that I can see without hovering, so that I know that sharing is possible.
- As a user who opened an artifact, I want to share it from there, so that I do not have to close it and return to the row.
- As a user who looks at a private artifact, I want its **Private** badge to lead to the sharing controls, so that the state and the control are in one place.

## Scope

**In:** making sharing discoverable in the Artifacts tab list and in the artifact preview.

**Not in:** what the sharing dialog offers (#3114 restricted groups, #3163 expiry labels), and the feedback after a link is copied (#3064).

## Open Questions

- Row actions are hover-only across the product — files, sessions, and artifacts all use the same pattern. Do we keep that and make sharing the exception, or is hover-only reveal the wider problem? This is a design-system decision, not an artifacts-only one.
- Does the **Private** badge become the control, or does a separate visible button sit beside it?

---

Filed by @sarahmilleribm

PetrBulanek:
A user who wants to share an artifact can now see how to do it, without
hovering — both in the Artifacts list and in the opened artifact.

Today every share control hides behind a hover reveal, so at rest an artifact
row offers no visible way to share. Opening the artifact is a dead end too: the
preview offers a link to the public page only once the artifact is already
public, and nothing that starts sharing. The result is that users conclude the
product cannot share artifacts at all.

Two changes fix it. The `Public`/`Private` badge becomes the control that opens
the sharing dialog, so the state and the control sit in one place, and the row's
share actions are visible at rest. The opened artifact gains its own visibility
badge, a **Copy link** control once it is public, and a **Share** button that is
always there — so sharing never means closing the artifact and hunting for its
row again.

Hover-only row actions stay the norm elsewhere in the product; artifacts is a
deliberate exception, not a design-system change.

Built to the final Figma design (`DAM-DEV`, node `2288-347`).

Closes #3186

## Sub-issues

- [x] 01 — List row: visible share controls, badge as control
- [x] 02 — Preview dialog: share from the opened artifact


packages/ui/src/modules/artifacts/components/artifact-row.tsx
```diff
@@ -105,10 +95,12 @@ export function ArtifactRow({
         <span className="flex items-center gap-2.5 text-xs text-muted-foreground">
           {showAgent && <CreatorChip agentId={artifact.agentId} />}
           {artifact.version > 1 && <VersionBadge version={artifact.version} />}
-          <span className="inline-flex items-center gap-1">
-            <View size={12} />
-            {artifact.viewCount}
-          </span>
+          {artifact.visibility === "public" && (
```
dam-code-guardian: 🟡 **Warning:** The view count now renders only when `artifact.visibility === "public"`, and this is the only place `viewCount` appears in the whole UI. An artifact that was public and is then set back to private keeps a non-zero count that no screen shows — and `docs/architecture/artifact-library.md:65` still promises the count as an in-app reach signal without qualification. Issue #3186 also records the count as part of what a row shows at rest, so this drops existing row information that nothing asked for.

**Fix:** No architecture page asserts an in-app signal that the code shows in only some states — keep the signal visible in every state, or narrow the page in the same PR.

_Reported by doc-drift and issue-fit._

packages/ui/src/modules/artifacts/components/artifact-preview-dialog.tsx
```diff
@@ -115,14 +127,10 @@ export function ArtifactPreviewDialog({ artifact, onClose }: Props) {
...
         <DialogFooter>
-          {artifact.shareUrl && (
-            <Button variant="outline" asChild>
-              <a href={artifact.shareUrl} {...externalLinkProps}>
-                <Launch size={16} />
-                Open share page
-              </a>
-            </Button>
-          )}
+          <Button variant="outline" onClick={() => setSharing(true)}>
```
dam-code-guardian, line 130: 🟡 **Warning:** The **Open share page** link is replaced rather than supplemented, and no `href` to a `shareUrl` remains anywhere in the UI. To view the public page of an artifact that is already public, the user now has to press **Copy link** and paste the URL into the address bar. Issue #3186 asks to add an entry point to sharing and lists **Open share page** among the affordances the dialog already offers.

**Fix:** A discoverability change adds controls and keeps every affordance the surface already had — remove one only when the issue or the discussion asks for it.

packages/ui/src/components/ui/badge.tsx
```diff
@@ -1,10 +1,11 @@
+import { Slot } from "@radix-ui/react-slot";
 import { cva, type VariantProps } from "class-variance-authority";
 import * as React from "react";
 
 import { cn } from "@/lib/utils";
 
 const badgeVariants = cva(
-  "inline-flex items-center rounded-full border transition-colors focus:outline-hidden focus:ring-2 focus:ring-ring focus:ring-offset-2",
+  "inline-flex items-center rounded-full border transition-colors focus-visible:outline-hidden focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2",
```
dam-code-guardian: 🟡 **Warning:** The base class list sets `focus-visible:ring-offset-2` but never `ring-offset-background`, and the new `asChild` path makes the badge a real `<button>` that keyboard users can focus. With no offset colour the value falls back to white, so the focus ring on the new sharing badge draws a white halo against the `#161616` card in dark mode. Every other primitive in `components/ui` pairs the two classes — `button.tsx:10`, `input.tsx:7`, `switch.tsx:28`, `tabs.tsx:21`; this file is the only one that sets the offset without its colour.

**Fix:** A class list that sets `ring-offset-*` also sets `ring-offset-background` — the offset colour is part of the focus-ring recipe, not an optional extra.

```suggestion
  "inline-flex items-center rounded-full border transition-colors focus-visible:outline-hidden focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 focus-visible:ring-offset-background",
```

_Reported by react-ui-engineering._

dam-code-guardian (dismissed):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `b835486`

## PR #3582: feat(ui): make artifact sharing visible
**Author:** PetrBulanek | **Branch:** feat/3186-visible-share-artifact → main | **Changes:** +194 −142 (14 files)

### Summary
Artifact sharing becomes visible without a hover: the visibility badge turns into the control that opens the sharing dialog, the row shows its share actions at rest, and the preview dialog gains a badge, a **Copy link** control and a **Share** button. The PR also reworks the shared dialog insets and replaces the `Modal` Escape handler with a LIFO open-modal stack. _17 suggestion(s) dropped under the 🟢 budget._

### Findings
- 🟡 **Warning:** view count hidden unless public, contradicting the architecture page (`packages/ui/src/modules/artifacts/components/artifact-row.tsx:98`, also `docs/architecture/artifact-library.md:65`; doc-drift, issue-fit)
  **Fix:** No architecture page asserts an in-app signal that the code shows in only some states — keep the signal or narrow the page in the same PR.
- 🟡 **Warning:** **Open share page** deleted with no replacement anywhere in the UI (`packages/ui/src/modules/artifacts/components/artifact-preview-dialog.tsx:130`; issue-fit)
  **Fix:** A discoverability change adds controls and keeps every affordance the surface already had.
- 🟡 **Warning:** `ring-offset-2` without `ring-offset-background` on a badge that is now focusable (`packages/ui/src/components/ui/badge.tsx:8`; react-ui-engineering)
  **Fix:** A class list that sets `ring-offset-*` also sets `ring-offset-background` — the offset colour is part of the focus-ring recipe.
- 🟢 **Suggestion:** The new Escape stack knows only `Modal`, so an overlay with its own portal stays outside it. `FullscreenPreviewDialog` installs a private `document` Escape listener (`fullscreen-preview-dialog.tsx:16`) and never registers, so with fullscreen open above the preview the preview is still topmost in `openModalIds` and one Escape closes both layers. Unchanged from `main`, but it is the nesting case the new hook exists to correct — one exported registry that every dismissable overlay joins would cover it (`packages/ui/src/modules/artifacts/components/artifact-preview-dialog.tsx:149`).
- ✅ **Looks good:** The LIFO Escape stack is correct for the nesting this PR introduces. `ShareDialog` mounts after `ArtifactPreviewDialog`, so it is last in `openModalIds` and painted last, and passing `onClose={undefined}` while `sharing.isPending` keeps Escape inert exactly while the mutation runs.

- 1 finding merged into `### Findings`: the view count against `docs/architecture/artifact-library.md:65`.
- 1 suggestion dropped under the 🟢 budget: the added paragraph states affordance mechanics — hover reveal, badge placement — below the architectural level, where it rots on the next visual pass.

### Issue Fit
The PR closes #3186 and meets its goal on both surfaces the issue names: the artifacts list and the opened artifact.
- 2 findings merged into `### Findings`: the hidden view count and the removed **Open share page** link, both of which the issue records as present today.
- 6 suggestions dropped under the 🟢 budget: the preview badge gets no `onShare`, so it names the state and leads nowhere while a separate footer button does the work; the dialog inset rework reaches five files outside artifacts; `font-sans` on `buttonVariants` changes the typeface of buttons that previously inherited mono on untouched surfaces; the Escape-semantics change is undeclared — all four in a PR whose body says this is "not a design-system change"; the badge drops its `Deleting soon` state while the author's own question about that state is unanswered on the issue; `badgeVariants` moves from `focus:` to `focus-visible:`.

- 1 finding merged into `### Findings`: `badge.tsx` sets a focus-ring offset without its background colour.
- 1 finding printed in `### Findings` as a 🟢: portal-rendered overlays stay outside the Escape registry.
- 10 suggestions dropped under the 🟢 budget, two of them regraded from 🟡 because the demonstration does not hold: an `onClose`-less `Modal` does join the stack, but no reachable screen mounts one above a closable modal — the one inert case, `share-dialog.tsx:56`, is intentional; and the badge loses `Deleting soon`, but `main` renders no badge in the preview at all and the row keeps its own deletion chip and `opacity-55`, so no surface loses the warning. The rest: `ShareDialog` keeps the hand-built copy control this PR just extracted; `CopyLinkButton` hand-copies its `variant` union instead of deriving it from `buttonVariants`; the copied state changes only the icon, not the accessible name; `BadgeProps` still extends `ComponentProps<"span">` after gaining `asChild`; `font-sans` sits at two badge call sites rather than in `badgeVariants`; `DialogBody` bottom padding depends on a `[&:last-child]` selector; `version` pins to `initialArtifact.version` while the artifact comes from `useArtifact`; the row action cluster now nests interactive controls inside a `role="button"` container permanently.

### Verdict
REQUEST_CHANGES — three warnings block: a documented in-app signal lost in the private state, an affordance removed with no replacement, and a focus ring that draws white on the dark card.


---

packages/ui/src/components/modal.tsx
```diff
@@ -149,7 +142,11 @@ export function DialogBody({
 }: DialogRegionProps & { flush?: boolean }) {
   return (
     <div
-      className={`flex-1 min-h-0 overflow-y-auto py-5 ${flush ? "" : "px-5 md:px-7"} ${className ?? ""}`}
+      className={cn(
+        "flex-1 min-h-0 overflow-y-auto pt-5 md:pt-6 [&:last-child]:pb-5 md:[&:last-child]:pb-6",
```
kapetr: Bottom padding now depends on the body being the last child. Every dialog that renders something (such as DialogFooter) after the body loses its bottom padding.

Image

kapetr (changes requested):
Feature itself looks good to me 👍

kapetr (approved):

packages/ui/src/components/modal.tsx
```diff
@@ -149,17 +142,29 @@ export function DialogBody({
...
+  className,
+  divided,
+}: DialogRegionProps & { divided?: boolean }) {
   return (
     <div
-      className={`px-5 md:px-7 py-4 flex items-center justify-end gap-3 ${className ?? ""}`}
+      className={cn(
+        "flex items-center justify-end gap-3 px-5 pb-5 md:px-6 md:pb-6",
+        divided && "border-t border-border pt-5 md:pt-6",
+        className,
```
kapetr, line 167: Footer lost its top padding, see "Publish skill" modal - buttons touch the footer separator.

Image

kapetr (changes requested):

kapetr (approved):

kapetr:
`packages/ui/src/modules/artifacts/components/version-badge.tsx` is unused now.

Issue #3506, opened 2026-08-28
no oom signal


jezekra1:
when agent is killed for lack of memory in UI it just looks as a restart with no way to tell what's wrong

PR #3583, opened 2026-09-07
feat: OOM resilience for agent harnesses

JanPokorny:
When a coding-agent harness (or a tool process it spawns) exhausts the agent pod's memory limit, the kernel OOM-kills the whole container — harness included — and the in-flight turn silently vanishes, with no restart. This PR makes agents resilient to that in two layers.

**Prevention — early-OOM reaper.** agent-runtime watches the pod's memory cgroup and, nearing the limit, SIGKILLs the largest process *below its own direct children* — a tool process (a Bash spawn, a build, a runaway allocation) the harness observes as a failed command and recovers from mid-turn — sparing the harness, PTY harnesses, the pod service, and sshds it supervises. The trigger measures *unreclaimable* memory (subtracting file-backed page cache the kernel would reclaim first), reading the cgroup v2/v1 stat pair. When the harness itself is the hog there is nothing to sacrifice and the kernel's kill still lands, which is what layer two is for. Tunable via `MEM_REAPER` / `MEM_REAPER_THRESHOLD`.

**Recovery — interrupted-turn resume.** Every running turn is marked in a durable per-pod document; a marker surviving to the next boot names a turn an abnormal death (an OOM group-kill or an eviction) cut short. On boot, **every** such session resumes on its own — once, so a continuation that dies again cannot crash-loop the pod — with an injected notice telling the agent its turn was interrupted and to pick the task back up. This runs unattended: the user never has to notice and restart it. It is not the platform's no-auto-resend case — a marker is written only after the harness *took* the prompt, so recovery continues a turn the agent already saw, and never re-sends a queued prompt it never received. A person's session that auto-resumes also gets an informational notice on load, so they can see it was interrupted and is resuming.

Markers are deliberately preserved across a graceful SIGTERM (an eviction is a SIGTERM then a SIGKILL), so evicted turns recover rather than being silently dropped.

Architecture docs (`agent-lifecycle`, `persistence`, `platform-topology`) describe both layers.

## Verification

`mise run check` green across the workspace; agent-runtime and UI unit suites pass, including tests for reaper victim selection, the active-turn store, and the interrupted-turn notice. The live-cluster smoke test (drive a real OOM, watch the pod survive and a session auto-resume) is still pending a running k3s cluster.


packages/agent-runtime/src/core/mem-reaper.ts
```diff
@@ -0,0 +1,116 @@
...
+    opts.log("no readable cgroup memory limit; reaper disabled");
+    return;
+  }
+  setInterval(() => {
+    try {
+      const cur = readCgroupBytes(
+        "/sys/fs/cgroup/memory.current",
+        "/sys/fs/cgroup/memory/memory.usage_in_bytes",
+      );
+      if (cur === null || cur / cgMax < opts.thresholdFraction) return;
```
dam-code-guardian: 🟡 **Warning:** The reaper compares `memory.current` against the limit, and that counter includes reclaimable page cache. A build or a large file read fills the cache and reaches 0.93 of the limit with no OOM risk, so the reaper SIGKILLs a healthy tool process and the harness sees a failed command.
**Fix:** Subtract the reclaimable file pages before the comparison — read `inactive_file` and `active_file` from `memory.stat` (v2 and v1 paths) and gate the kill on the anonymous plus kernel part only.

packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts
```diff
@@ -139,15 +141,23 @@ export function createAcpRuntime(deps: AcpRuntimeDeps): AcpRuntime {
...
-      if (meta?.type === SessionType.ScheduleCron || meta?.scheduleId)
+      const machine = isMachineSession(sessionId);
+      deps.activeTurns?.record(sessionId, machine ? "machine" : "interactive");
+      if (nonViewerChannels.has(channel) && machine)
         deps.sessionMetadata?.startRun(sessionId);
     },
-    onTurnEnded: (sessionId) => deps.sessionMetadata?.finishRun(sessionId),
+    onTurnEnded: (sessionId, cause) => {
+      deps.sessionMetadata?.finishRun(sessionId);
+      if (cause === "completed") deps.activeTurns?.remove(sessionId);
```
dam-code-guardian: 🟡 **Warning:** A marker is written for every turn, but it is removed only when `cause === "completed"`. A turn that the scheduler drops in this same process keeps its marker: `teardownRuntime` calls `promptScheduler.clear()` on an env recycle or an unresponsive harness, and `clear()` reports `"dropped"` (`prompt-scheduler.ts:254`). The next `session/load` of that interactive session then returns `interruptedAt`, and the UI states "interrupted by an unexpected restart, likely out of memory" although the pod never restarted. The notice repeats on every load until a later turn of that session completes.
**Fix:** Remove the marker on every in-process turn end, whatever the cause — a `"dropped"` turn ended inside this process and is not an abnormal death. Keep a real crash as the only source of a leftover marker.

packages/agent-runtime/src/modules/acp/services/acp-runtime/prompt-scheduler.ts
```diff
@@ -249,7 +251,7 @@ export function createPromptScheduler(
     forget(sessionId) {
       const wasActive = activeTurns.delete(sessionId);
       dropQueue(sessionId, "session-forgotten");
-      if (wasActive) deps.onTurnEnded?.(sessionId);
+      if (wasActive) deps.onTurnEnded?.(sessionId, "dropped");
```
dam-code-guardian: 🟡 **Warning:** This is the second location of the marker-lifecycle defect above. `forget()` and `clear()` both report `"dropped"`, and the runtime keeps the marker for that cause, so a session forgotten or a harness recycled on a live pod leaves a marker that the next `session/load` reads as an OOM restart.
**Fix:** Remove the marker on every in-process turn end, whatever the cause — a `"dropped"` turn ended inside this process and is not an abnormal death. Keep a real crash as the only source of a leftover marker.

packages/agent-runtime/src/core/mem-reaper.ts
```diff
@@ -0,0 +1,116 @@
...
+      const victim = pickVictim(readProcTable(), process.pid);
+      if (victim === null) {
+        opts.log(`at ${usage} with only protected processes; cannot reap`);
+        return;
+      }
+      process.kill(victim.pid, "SIGKILL");
+      opts.log(
+        `killed pid ${String(victim.pid)} (${victim.name}, ${mib(victim.rssBytes)}MB) at ${usage}`,
+      );
+    } catch {}
```
dam-code-guardian: 🟡 **Warning:** The poll body is wrapped in a bare `catch {}`, so a `process.kill` that fails — EPERM, or ESRCH for a pid that exited between the table read and the kill — is discarded with no log. The reaper then reports nothing while the container walks into the kernel OOM kill it exists to prevent. Reported by typescript-engineering.
**Fix:** Never use a bare `catch {}` around an action whose failure changes behaviour; log the error through the injected `log` so a reaper that cannot kill is visible.

docs/architecture/agent-lifecycle.md
```diff
@@ -131,7 +131,7 @@ This makes **the server authoritative about delivery**, which is the point: watc
...
... **parks** the queue for a grace window, and a client engaging again within it resumes it. A queue leaves the scheduler by four routes — the window expiring, the session forgotten, the scheduler cleared when the harness goes down, or a session the harness cannot load back refusing its queue — and all but the refusal record the queue's prompts as **undelivered** in a durable document on the pod, whole, inline images capped; the refusal errors each sender instead. Each session load hands the records back, marked on the conversation to send again or delete, and says whether a turn is still running; a record sent again or deleted is **retired** so its earlier echo drops from the replay rather than doubling. Recovery is always a send the user initiates: an automatic resend cannot know whether a prompt was dropped before or after the agent saw it, so it risks running it twice. A prompt already *handed to the harness* is the other side of that split: losing the channel costs the sender only its live view — the turn runs to completion and its output lands in the log for any later viewer. Every honest delivery report depends on this split ([channels](channels.md) states it for its surface).
```
dam-code-guardian, line 134: 🟡 **Warning:** This page describes neither the reaper nor the interrupted-turn recovery, although the PR body states that `agent-lifecycle` and `persistence` are both updated; its diff only rewords three unrelated paragraphs. Two statements on this page now read false: recovery is "always a send the user initiates", and a prompt handed to the harness "runs to completion". Reported by doc-drift and issue-fit.
**Fix:** Add the owning section to this page in this PR — who the reaper protects and who it sacrifices, and what a leftover marker resumes or surfaces — then correct the two statements to name the machine-driven auto-resume and the OOM exception.

docs/architecture/persistence.md
```diff
@@ -113,7 +113,7 @@ A home volume mounts empty and shadows whatever the image bakes at that path, so
...
...rt interrupted — alongside the trigger-binding and runtime-channel state files. The same directory holds the **undelivered-prompts document**: the whole content of user prompts that never reached the harness (inline images capped, file attachments by name), written only when a prompt is recorded, sent again, or deleted — never on a routine turn — and deliberately separate from session metadata so a corrupt metadata write cannot take user text with it. A record outlives session teardown and pod restarts; it goes when the user resends or deletes it, with its Session, or — the document is byte-capped — when another session's write needs the room, evicting whole sessions oldest-first ([agent-lifecycle](agent-lifecycle.md) owns the delivery contract). A third document, **active-turns**, marks each session while a turn runs and clears it when the turn completes; a marker still present at the next boot names a turn an abnormal death (an OOM kill, an eviction) interrupted, which agent-lifecycle's recovery consumes to resume a machine-driven session or surface the interruption to an interactive one — its own document again, so a corrupt write cannot take run accounting or user text with it.
```
dam-code-guardian: 🟡 **Warning:** This is the second location of the documentation gap above. The new active-turns paragraph hands the recovery contract to `agent-lifecycle.md` — "which agent-lifecycle's recovery consumes" — and that page carries no recovery content, so the cross-link points at a concept no page owns. Reported by doc-drift and issue-fit.
**Fix:** Add the owning section to `agent-lifecycle.md` in this PR — who the reaper protects and who it sacrifices, and what a leftover marker resumes or surfaces — then correct the two statements to name the machine-driven auto-resume and the OOM exception.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `348ac16`

## PR #3583: feat: OOM resilience for agent harnesses
**Author:** JanPokorny | **Branch:** feat/oom-resilience#locki-46m3exop → main | **Changes:** +560 −32 (20 files)

### Summary
The PR adds two OOM layers to agent-runtime: a poll loop that SIGKILLs the largest tool process near the pod memory limit, and durable active-turn markers that let the next boot resume a machine-driven turn or flag an interactive one. _14 suggestion(s) dropped under the 🟢 budget._

### Findings
- 🟡 **Warning:** reaper trigger counts reclaimable page cache (`packages/agent-runtime/src/core/mem-reaper.ts:103`)
- 🟡 **Warning:** a locally dropped turn keeps its marker and shows a false OOM notice (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:159`, also `packages/agent-runtime/src/modules/acp/services/acp-runtime/prompt-scheduler.ts:254`)
- 🟡 **Warning:** bare `catch {}` hides a failed kill (`packages/agent-runtime/src/core/mem-reaper.ts:114`) — reported by typescript-engineering
- 🟡 **Warning:** `agent-lifecycle.md` describes neither layer, and its delivery contract now reads false (`docs/architecture/agent-lifecycle.md:134`, also `docs/architecture/persistence.md:116`) — reported by doc-drift and issue-fit
- 🟢 **Suggestion:** `activeTurns` is declared optional (`activeTurns?: ActiveTurnStore`), so all four call sites branch on presence and a missing wiring becomes a silent no-op instead of a type error. `composeAcp` always passes it, and the repo's architecture guidance requires service dependencies (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:104`) — reported by typescript-engineering
- ✅ **Looks good:** `pickVictim`'s depth rule matches the real process topology. The harness (`create-child-agent-process.ts:31`) and each sshd (`ssh.ts:126`) are direct children of agent-runtime, so the depth ≥ 1 test protects exactly the supervised set.

- 🔁 Four findings on the missing `agent-lifecycle.md` content, the absolute no-auto-resend sentence, the dangling `persistence.md` cross-link and the unrecorded `interruptedAt` load signal. They are one 🟡 in `### Findings`. doc-drift graded the contradicted sentence 🔴; the review regrades it 🟡, because no code path gives a wrong result.
- _2 suggestion(s) dropped under the 🟢 budget._

### Issue Fit
- 🔁 Two findings on the PR body's documentation promise and the same dangling cross-link. They are part of the same 🟡 in `### Findings`.
- _4 suggestion(s) dropped under the 🟢 budget._

- 🔁 The bare `catch {}` finding is a 🟡 in `### Findings`; the optional-dependency finding is the 🟢 there.
- _8 suggestion(s) dropped under the 🟢 budget._

### Verdict
COMMENT — both layers are sound in design, but the reaper's trigger measure and the marker lifecycle each fire on paths with no OOM behind them, and the architecture page the body promises is unwritten.


---

packages/agent-runtime/src/core/mem-reaper.ts
```diff
@@ -0,0 +1,146 @@
...
+    return;
+  }
+  setInterval(() => {
+    try {
+      const cur = readCgroupBytes(
+        "/sys/fs/cgroup/memory.current",
+        "/sys/fs/cgroup/memory/memory.usage_in_bytes",
+      );
+      if (cur === null) return;
+      const reclaimable = sumStat("/sys/fs/cgroup/memory.stat", [
```
dam-code-guardian: 🟡 **Warning:** `sumStat` names the cgroup v2 path alone, and its `catch` returns `0`, while every other read here goes through the v2/v1 pair (`readCgroupBytes`, lines 113–116). On a v1 host, or any host where `memory.stat` is unreadable, `reclaimable` is 0, `unreclaimable` becomes the raw `cur` again, and the threshold test silently reverts to the behavior this commit set out to correct — with nothing logged to say the correction was not applied.
  **Fix:** Every cgroup read uses the same v2/v1 pair, and a read whose failure changes a decision returns `null` rather than a neutral-looking `0`: log it and skip the poll, the way `cur === null` is already handled.

packages/agent-runtime/src/modules/acp/services/interrupted-turn-recovery.ts
```diff
@@ -0,0 +1,60 @@
...
+ * already saw — it never re-sends a queued prompt the agent never received
+ * (those live in the undelivered-prompts document and stay user-initiated).
+ */
+export async function recoverInterruptedTurns(deps: {
+  store: ActiveTurnStore;
+  sessionMetadata: SessionMetadataStore;
+  triggerDriver: TriggerSessionDriver;
+  log: (msg: string) => void;
+}): Promise<void> {
+  for (const marker of deps.store.leftovers()) {
```
dam-code-guardian, line 37: 🟡 **Warning:** With `marker.origin !== "machine"` gone, a person's chat session is resumed too: on the next boot the driver sends `<turn-interrupted>` and a turn runs with nobody watching. The same commit deletes `appendInterruptedNotice` and the `interruptedAt` load field, so the person's only signal is an unexplained agent message continuing a task — which is the gap open issue #3506 reports.
  **Fix:** A session resumes on its own only when the person is told its turn was interrupted; automatic recovery and reporting the interruption are separate deliverables, so keep the surfacing when recovery becomes automatic.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `2a3490a`

## PR #3583: feat: OOM resilience for agent harnesses
**Author:** JanPokorny | **Branch:** feat/oom-resilience#locki-46m3exop → main | **Changes:** +525 −27 (14 files)

### Summary
One commit answers the previous round: the reaper now measures unreclaimable memory and reports a failed kill, the marker clears on every observed turn end, and `activeTurns` is required. The same commit also drops the machine/interactive split, so every interrupted session now auto-resumes. _11 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: 348ac16 (2026-09-07T11:24:08Z) — verdict COMMENT

- ✅ **Fixed:** reaper trigger counts reclaimable page cache (`packages/agent-runtime/src/core/mem-reaper.ts:103`)
- ✅ **Fixed:** dropped turn keeps its marker, false OOM notice (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:159`)
- ✅ **Fixed:** bare catch hides a failed kill (`packages/agent-runtime/src/core/mem-reaper.ts:114`)
- ✅ **Fixed:** activeTurns declared optional, so absence is silent (`packages/agent-runtime/src/modules/acp/services/acp-runtime/acp-runtime.ts:104`)
- 🔁 **Still present:** agent-lifecycle describes neither layer; its contract reads false (`docs/architecture/persistence.md:116`)
- 🆕 **New:** sumStat reads the v2 memory.stat path only (`packages/agent-runtime/src/core/mem-reaper.ts:118`)
- 🆕 **New:** a person's session resumes unattended with no signal to them (`packages/agent-runtime/src/modules/acp/services/interrupted-turn-recovery.ts:36`)
- 🆕 **New:** SIGTERM clears every marker, so an eviction is never recovered (`packages/agent-runtime/src/server.ts:703`)

### Findings
- 🟡 **Warning:** sumStat reads the v2 memory.stat path only (`packages/agent-runtime/src/core/mem-reaper.ts:118`) — reported by typescript-engineering
- 🟡 **Warning:** a person's session resumes unattended with no signal to them (`packages/agent-runtime/src/modules/acp/services/interrupted-turn-recovery.ts:36`) — reported by issue-fit
- 🟢 **Suggestion:** `gracefulShutdown` calls `activeTurns.clearAll()` on SIGTERM (`packages/agent-runtime/src/server.ts:703`), but an eviction and a rollout are a SIGTERM before the SIGKILL, so the markers go before the kill lands and the turn both layers exist to recover is not recovered — while the store's own boundary comment names an eviction as exactly that case. Reported by typescript-engineering and doc-drift.

- 🔁 Four findings on the two new paragraphs and the page they omit: the reaper escalation the code does not implement, the eviction the marker never survives, the unrecorded agent-runtime responsibilities in `platform-topology.md`, and the deleted sender-side hand-over sentence. They are the one docs 🟡 above, still present at its anchor; the eviction item is also the 🟢 above.
- _1 suggestion(s) dropped under the 🟢 budget._

### Issue Fit
- 🔁 Two findings on the PR body, which still describes the interactive-notice design this commit deleted and still lists its UI commit. They belong to the same docs 🟡 above.
- 🆕 Two findings on the widening itself — undeclared scope against the PR's own stated design, and the loss of the only user-facing OOM signal, the gap open issue #3506 reports. They are the recovery 🟡 above.
- _2 suggestion(s) dropped under the 🟢 budget._

- 🔁 Six findings. The `sumStat` silent zero is the reaper 🟡 above; the SIGTERM clear is the 🟢 above. The other four — a malformed marker record dropped without a report, a failed resume leaving its marker forever, `server.ts` importing past the module index, and `compose.ts` returning the store — are placement and reporting on code this delta did not change, so they are suggestions here.
- _4 suggestion(s) dropped under the 🟢 budget._

### Verdict
COMMENT — the four blocking findings from the previous round are fixed at their cause, but the new threshold read still degrades silently on cgroup v1, the docs contract is not yet correct, and dropping the machine/interactive split resumes a person's session with nothing telling them why.


---

Issue #3244, opened 2026-08-10
Map every slow moment in the platform


jezekra1:
## Context

Epic #3242. People say the platform is slow, but nobody can point at which
moments are slow or by how much. Every speedup discussion turns into opinions.

## Research Goal

Produce one shared list of every moment a user waits, with a real number next
to each, measured in both the dev and local environments.

## Research Questions

- Where does a user actually wait? Candidates to check first:
  - loading conversation history (switching sessions)
  - loading the file tree and opening files
  - loading and showing artifacts
  - chatting with an agent (time to first words, time to finish)
  - starting an agent
- How long does each one take today, in seconds?
- Which are the worst, counting both how slow they are and how often people hit them?
- Which ones look like our problem, and which look like the model being slow?

## Users / Audience

Anyone using the platform through the UI.

## Sources of Insight

- Walking through the platform the way a user does, in dev and local
- Browser timings and server logs
- Asking the team which moments annoy them most

Run every measurement on a well-resourced agent — at least 4 CPU and 4Gi of
memory. We want to know how fast the best case is, not how slow a starved agent
is.

## Decision it informs

Which deep-dives are worth doing, and in what order. Without this we would
guess.

## Done when

- A written list of slow moments with numbers for dev and local
- A sub-issue filed on this epic for each one worth digging into

jezekra1:
Closing as served. The map produced the sub-issues it existed to produce — #3245, #3246, #3247 — and those drove every latency fix since.

One part of the done-when was not met: numbers were taken on dev only, never on local. Leaving this open for that alone would be holding a research issue open for data no current decision depends on. If a local-vs-dev question comes up later it deserves its own issue with a reason attached.

PR #3586, opened 2026-09-07
fix(skills): the Skills page paints from what it last held

### LINKED ISSUE #3278: Skills page reloads everything from scratch on every visit
## Context

Found while using the sandbox Skills page in local and dev. It happens every time, on every sandbox, for every user — not an edge case and not load-dependent.

This is the same family as #3245 and #3247: a page that feels slow because of how it loads, not because the underlying data is large.

## Problem

The Skills page never feels ready. Opening it shows a placeholder for every connected source, and the wait repeats in full each time you come back — even if you left for two seconds and nothing changed.

Skills is where you configure what a sandbox can do, so people move in and out of it repeatedly while setting a sandbox up. The cost is paid on every one of those visits. The page also loads as one lump: a single slow source holds back every other source's list, so the page is only ever as fast as its worst repo.

The effect is that the page reads as unreliable. A user cannot tell whether it is loading, stuck, or empty, and the honest answer to "did my skill get installed?" becomes "wait and see."

## Expected Behavior

Returning to Skills shows the skills you were just looking at, immediately. If anything has changed upstream, it updates in place without blanking the page first.

On a genuine first load, each source appears as soon as its own list is ready.

## Actual Behavior

- Every visit re-loads the full page from nothing. Nothing is carried over from the previous visit.
- All sources appear together, when the slowest one finishes. A source with a short skill list waits for a source with a long one.
- After a few minutes away, the page re-reads every connected repository in full again, so the wait is back to its worst case.

## Steps to Reproduce

1. Open a sandbox with two or more skill sources connected, and go to Skills.
2. Wait for every source to finish loading.
3. Switch to another section, for example Connections.
4. Go back to Skills.

Every source shows a loading placeholder again, and the full wait repeats.

## Proposed direction

Three outcomes, in the order they'd be felt:

1. **Remember what was already loaded.** A revisit should paint from what the page last showed and refresh quietly behind it. This alone removes the repeated wait, which is the part users hit most.
2. **Let each source arrive on its own.** One slow repository should not decide when the page appears.
3. **Keep the remembered skill lists longer, and refresh them in the background.** The page should never have to make a user wait on a full re-read of a repository just because they stepped away.

We should also confirm what "ready" costs today with real numbers in dev, so this can be shown as fixed rather than asserted — that is what #3244 is for.

xjacka:
Opening Skills showed a placeholder for every connected source, and the full
wait repeated on every return to the page even when nothing had changed. After
a few minutes away it was back to the worst case: every connected repository
read again from nothing.

Closes #3278

## Why it happened

Two independent causes, one per side.

**The browser kept nothing.** The Skills panel held sources, per-source scans,
reconciled state and skill sets in hook-local `useState`, loaded imperatively
from `useEffect`. Leaving the section unmounted the hook and threw all of it
away, so a return had nothing to paint from and started over. The one read that
did live in the query cache (`skills.state`, written out for the nav badge) was
still re-read into local state on every mount.

**The server's scan cache expired hard.** An entry answered for five minutes;
the first read after that blocked on a full repository read. Step away for six
minutes and the wait was the uncached one, in full.

## What changed

**UI — the reads live in the query cache.** `sources.list`, the per-source
`listWithScan`, `state` and `sets.list` are TanStack Query queries now, with a
`gcTime` long enough to outlive the page (the default five minutes was itself
the "few minutes away" symptom) and a `staleTime` that lets a revisit inside a
minute cost nothing at all. A revisit paints the lists it last showed and
refreshes each source behind them, in place, without blanking. Mutations stay
imperative but write their result into the cache under `cancelQueries`, which
replaces the hand-rolled epoch guard that used to stop an in-flight poll from
overwriting a just-completed toggle.

`SkillsSurface`'s own interface is unchanged, so no component moved. The
`onStateChange` plumbing is gone: the panel now reads the same query key the
section-summary badge reads, so the two stay in step without being wired
together.

**api-server — the scan cache refreshes behind what it serves.** An entry
answers directly for five minutes, keeps answering for half an hour from the
read it came from while a rescan runs behind it, and only past that lifetime
does a read wait on upstream again. One read now serves every caller waiting on
the same key, so two readers of one source never scan it twice. A read already
in flight when `sources.refresh` or a publish invalidates is disowned — it saw
the source before it moved, so it must neither answer the re-scan nor land as
the cached list. Without that last part, single-flight would have quietly broken
the Re-scan button.

Freshness stays honest throughout: a card dates its list from the upstream read
the entry came from, not from the moment it was served.

## Not in scope

Each source already arrived on its own before this change — the cards render
independently and nothing gates the list on the slowest scan, on either side of
the wire. What was missing was the card painting immediately and the same
repository not being read twice; both are covered above.

No measurements here. Numbers for what "ready" costs belong to #3244.

## Verification

`mise run check` is clean, except `controller:check:staticcheck`, which has no
`staticcheck` binary in my environment — no Go changed. 1310 api-server and 337
UI unit tests pass. The scan cache's expiry test is updated for the new
lifetime, and three scenarios are added: a stale entry answering at once while
the rescan runs behind it, one read serving every waiting caller, and an
invalidation disowning a read that started before it.

Not verified against a running cluster.


packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts
```diff
@@ -100,158 +122,97 @@ export function useSkillsSurface(
...
+      const errorBySource: Record<string, ScanFailure | null> = {};
+      const scannedAtBySource: Record<string, string> = {};
+      const visibilityBySource: Record<string, "public" | "private"> = {};
+      results.forEach((result, index) => {
+        const src = sources[index];
+        if (!src) return;
+        loadingBySource[src.id] = result.isFetching;
+        if (result.isError) {
+          errorBySource[src.id] = toScanFailure(result.error);
+          skillsBySource[src.id] = NO_SKILLS;
```
dam-code-guardian: 🟡 **Warning:** An errored scan writes `NO_SKILLS` over the source's list. TanStack Query keeps the last good `data` when a *background* refetch fails, and `retry: false` makes one failure enough, so a single failed revalidation on a revisit replaces a painted list with an error card — the blanking this PR removes for the success path.

**Fix:** When a query holds both an error and stale-but-valid `data`, serve the data and add the error beside it; use the empty collection only when that key never succeeded. `skill-source-card.tsx` gates its rows on `!error`, so it must also paint retained rows with the error.

packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts
```diff
@@ -100,158 +122,97 @@ export function useSkillsSurface(
...
-        }
-      } catch (err) {
-        setErrorBySource((e) => ({ ...e, [sourceId]: toScanFailure(err) }));
-        setSkillsBySource((s) => ({ ...s, [sourceId]: [] }));
-      } finally {
-        setLoadingBySource((l) => ({ ...l, [sourceId]: false }));
-      }
+      const key = trpc.skills.state.queryKey({ agentId });
+      await queryClient.cancelQueries({ queryKey: key });
+      queryClient.setQueryData(key, (prev) => (prev ? patch(prev) : prev));
```
dam-code-guardian: 🟡 **Warning:** The write applies only `if (prev)`. On a revisit the scans paint at once from the 24-hour cache while `skills.state` is still pending, so a toggle, publish or local write in that window returns an authoritative list that is dropped, and the panel keeps the old state until the 5-second poll lands. The replaced code always applied the mutation result.

**Fix:** An authoritative mutation result is never dropped because the cache slot is empty — seed the slot from the result, or invalidate the query so the next read carries it.

packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts
```diff
@@ -100,158 +122,97 @@ export function useSkillsSurface(
...
-    publishes,
-    standaloneSnapshot,
-    onStateChange,
-  ]);
-
-  const loadSkills = useCallback(
-    async (sourceId: string) => {
+  });
+
+  const stateQuery = useQuery({
```
dam-code-guardian: 🟡 **Warning:** This is a second declaration of the `skills.state` key: it leaves `refetchOnWindowFocus` at its default `true` and sets no `staleTime`, while `useSkillsState` in `modules/agents/api/skills.ts` owns the same key with focus refetch off. Options are per observer and the cache entry is shared, so a window focus straight after an install refetches state the outbox apply has not reached yet and overwrites the optimistic write. The `localWriteEpochRef` guard that ordered a local write against that poll is deleted in this PR.

**Fix:** One server read gets one query hook that owns its options — reuse `useSkillsState` — and keep an explicit ordering rule between a local write and the poll that can overwrite it.

packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts
```diff
@@ -100,158 +122,97 @@ export function useSkillsSurface(
...
+    combine: (results) => {
+      const skillsBySource: Record<string, Skill[]> = {};
+      const loadingBySource: Record<string, boolean> = {};
+      const errorBySource: Record<string, ScanFailure | null> = {};
+      const scannedAtBySource: Record<string, string> = {};
+      const visibilityBySource: Record<string, "public" | "private"> = {};
+      results.forEach((result, index) => {
+        const src = sources[index];
+        if (!src) return;
+        loadingBySource[src.id] = result.isFetching;
```
dam-code-guardian: 🟡 **Warning:** `isFetching` reports a background revalidation as a first load. A revisit past the 60-second stale window sets it while `data` is present, so `ScanFreshness` shows "Scanning…" instead of the scanned-ago row, and `skills-source-list.tsx` hides an existing `SourceError` behind `error && !scanning`. The page does not refresh behind what is on screen.

**Fix:** Report "first load, nothing to paint" from `isPending`, and expose background revalidation as a separate flag; never gate already-cached content or its error on `isFetching`.

docs/architecture/skills.md
```diff
@@ -217,6 +217,8 @@ Two of those verdicts describe the same 401/404 from GitHub and are told apart b
 
 The cache is invalidated on `sources.refresh` and after every successful publish to that source — the latter so a freshly-merged PR shows up on the next list. The in-product content read shares this cache, so an invalidation also updates which revision a preview renders.
 
+The Skills panel holds the lists it was last served and repaints from them when a user returns, refreshing each source behind what is already on screen rather than emptying the page first. A source's list therefore costs a wait on the first visit, not on every one, and a slow repository delays only its own card.
```
dam-code-guardian: 🟡 **Warning:** This paragraph states behavior wider than what ships. No query persister is configured on the shared client, so a reload pays the wait again — the cost is per session, not per first visit. The scans run with a 60-second stale window and `refetchOnWindowFocus: false`, so a quick return refreshes nothing behind the screen. And "a slow repository delays only its own card" is the outcome the PR body declares out of scope. Reported by doc-drift and issue-fit.

**Fix:** A subsystem page states the lifetime and the trigger the code implements, and claims only what the PR delivers — bound the client cache to the session, and leave an out-of-scope outcome off the page.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `8bb02a5`

## PR #3586: fix(skills): the Skills page paints from what it last held

**Author:** xjacka | **Branch:** claude/issue-3278-735bd8 → main | **Changes:** +319 −269 (6 files)

### Summary

The hook's `useState` slices and loader effects move to TanStack Query with a long `gcTime`, and the api-server scan cache becomes stale-while-revalidate with single-flight reads and disown-on-invalidate. Five warnings sit on the client half: the error and loading branches of the new query layer give back the blanking behavior the change exists to remove.

### Findings

- 🟡 **Warning:** In the `combine` reducer an errored scan writes `NO_SKILLS` over the source's list. TanStack Query keeps the last good `data` when a *background* refetch fails, so one failed revalidation on a revisit — `retry: false`, so one failure is enough — replaces a painted list with an error card, which is what the PR removes for the success path (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:162`)
  **Fix:** When a query holds both an error and stale-but-valid `data`, serve the data and add the error beside it; use the empty collection only when that key never succeeded. The consuming card gates its rows on `!error` too, so it must show retained rows with the error.
- 🟡 **Warning:** `patchState` applies the optimistic write only `if (prev)`. On a revisit the scans paint at once from the 24-hour cache while `skills.state` is still pending, so a toggle, publish or local write in that window returns an authoritative list that is dropped, and the panel shows the old state until the 5-second poll lands. The replaced code always applied the mutation result (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:207`)
  **Fix:** An authoritative mutation result is never dropped because the cache slot is empty — seed the slot from the result, or invalidate the query so the next read carries it.
- 🟡 **Warning:** `skills.state` is re-declared here with `refetchOnWindowFocus` at its default `true` and no `staleTime`, while `useSkillsState` owns the same key with focus refetch off. Options are per observer and the entry is shared, so a window focus straight after an install refetches state that the outbox apply has not reached yet and overwrites the optimistic write. The `localWriteEpochRef` guard that ordered a local write against that poll is deleted in this PR (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:182`)
  **Fix:** One server read gets one query hook that owns its options — reuse `useSkillsState` from `modules/agents/api/skills.ts` — and keep an explicit ordering rule between a local write and the poll that can overwrite it.
- 🟡 **Warning:** `loadingBySource[src.id] = result.isFetching` reports a background revalidation as a first load. A revisit past the 60-second stale window sets `isFetching` while `data` is present, so the freshness row turns into "Scanning…" and an existing source error is hidden behind `error && !scanning` — the page does not refresh behind what is on screen (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:159`)
  **Fix:** Report "first load, nothing to paint" from `isPending` and expose background revalidation as a separate flag; never gate already-cached content or its error on `isFetching`.
- 🟡 **Warning:** The new architecture paragraph states behavior wider than what ships: no query persister is configured, so a reload pays the wait again, not only the first visit; the scans run with a 60-second stale window and focus refetch off, so a quick return refreshes nothing behind the screen; and "a slow repository delays only its own card" is the deliverable the PR body declares out of scope. Reported by doc-drift and issue-fit (`docs/architecture/skills.md:220`)
  **Fix:** A subsystem page states the lifetime and the trigger the code implements, and claims only what the PR delivers — bound the client cache to the session, and leave an out-of-scope outcome out of the page.
- ✅ **Looks good:** The three added `// TEST_SCENARIO:` comments are a registered comment type (`scripts/strip-comments.mjs` → `COMMENT_TYPES`, kept by `PRESERVE_TS`), so the repo's no-comments rule is not broken.

Its panel-paragraph finding is merged into `### Findings` (`docs/architecture/skills.md:220`). Two further suggestions — the single-flight sentence reading as unconditional where the same paragraph calls the cache per-replica, and the new local-write ordering rule going undocumented — fell under the 🟢 budget.

### Issue Fit

Linked issue: #3278 "Skills page reloads everything from scratch on every visit" (`Closes #3278`; no issue comments amend the ask). Its two blocking items are merged into `### Findings` — the failed-revalidation blanking (`use-skills-surface.ts:162`) and the page asserting the out-of-scope outcome (`docs/architecture/skills.md:220`). The observation that the PR closes the issue while declaring the issue's second named outcome out of scope fell under the 🟢 budget.

- 🟢 **Suggestion:** A background rescan failure goes to a raw `process.stderr` line with no source id, scope or error, and never reaches the module's structured logger or the entry. With a 30-minute lifetime a source whose credentials were revoked keeps serving a half-hour-old list with that single untyped trace (`packages/api-server/src/modules/skills/infrastructure/scan-cache.ts:108`)

Its other blocking findings are merged into `### Findings` (`use-skills-surface.ts:162`, `:182`, `:207`). Three suggestions — the scattered cache-surgery sites, `dropExpired` retention, and the `STALE_TTL_MS` name measuring total lifetime — fell under the 🟢 budget.

Its blocking findings are merged into `### Findings` (`use-skills-surface.ts:159`, `:162`, `:182`). Three suggestions fell under the 🟢 budget: the write-then-invalidate pair in `addSource` and `removeSource`, the failed top-level read that renders as an empty state (unchanged from `main`, so pre-existing), and `createSet` dropping its explicit `setSetsFailed(false)`.

### Verdict

REQUEST_CHANGES — the query layer's error and loading branches reintroduce the blanking and the lost local write that this change exists to remove.


---

xjacka:
Pushed 8363ab6 for the five warnings. Taking them in turn:

**Errored scan overwrote the retained list** (`use-skills-surface.ts:162`) — fixed, and it was the right call: before this PR there was never a painted list to lose, so writing an empty collection on failure only became wrong once a revisit paints from cache. A failed scan now records the failure and leaves whatever the last good scan returned. The card shows the failure above the retained rows instead of in place of them, and keeps its "scanned …" stamp so the age of what you are reading stays visible. A source that never scanned still gets the bare error card.

**`patchState` dropped the write on an empty slot** (`:207`) — fixed. The panel refetches `skills.state` instead of discarding the mutation result. I did not seed the slot from the result: it carries only the installed refs, so seeding would have to invent an empty local list and an empty publish list, and a fabricated "no local skills" is worse than a round trip.

**Two declarations of `skills.state`** (`:182`) — fixed. `useSkillsState` and a new `useSkillSources` in `modules/agents/api/skills.ts` own the options for both shared keys; the panel passes a poll interval and nothing else, and focus refetch stays off. On the ordering rule: `cancelQueries` before the cache write is that rule — an in-flight poll is cancelled and its result discarded, which is what `localWriteEpochRef` did. With focus refetch off on that key, the path in the finding is closed too.

**`isFetching` read as a first load** (`:159`) — fixed by splitting the two: a scan with nothing to paint is loading, a scan over data already on screen is revalidating. `isPending` alone would have gone quiet on a manual Re-scan of a source whose last scan failed — no data, but not pending either — so the split keys off whether there is anything painted, not off the status.

**The architecture paragraph claimed more than ships** (`skills.md:220`) — rewritten. It is now bounded to a browsing session, says a reload reads every source again, and re-reads a source only once its list has aged past a short window. I kept the sentence about a card filling from its own scan: the page describes the system as it stands, and that has been true on both sides of the wire for longer than this PR — it is out of scope as a deliverable, not as a fact.

**Untyped background-rescan log** (`scan-cache.ts:108`) — the line now carries the failure text alongside the source and scope. The structured logger is not reachable from this adapter without changing its injected log port, which is a wider change than the finding warrants.

Checks are green again: 1310 api-server and 337 UI unit tests, `mise run check` clean apart from `controller:check:staticcheck`, which has no binary in my environment and no Go changed. Still not exercised against a running cluster.

packages/ui/src/modules/sandboxes/components/skills/skill-source-card.tsx
```diff
@@ -37,23 +37,16 @@ function repoLabel(source: SkillSource): string {
...
-    return (
-      <span className="flex shrink-0 items-center gap-1.5 text-sm text-muted-foreground">
-        <Spinner size={13} /> Scanning…
-      </span>
-    );
-  }
   return (
     <span className="flex shrink-0 items-center gap-1.5 text-sm text-muted-foreground">
-      <Time size={13} />
+      {rescanning ? <Spinner size={13} /> : <Time size={13} />}
```
dam-code-guardian: 🟡 **Warning:** The re-scan signal is now an icon swap only: the text beside it keeps reading `scanned <ago>`, and `Spinner` is rendered with no `label`, so it is `aria-hidden` with no `role="status"`. The previous branch printed the word "Scanning…". The same signal is lost on two more surfaces, because `loadingBySource` was narrowed to `isFetching && !painted` (`use-skills-surface.ts:154`) and only `SkillSourcesSection` receives the new `revalidatingBySource`; `SkillsStoppedPanel` and `SkillsNeverRunPanel` still pass `loadingBySource` into `SkillSourceList`, whose "Scanning…" branch therefore never fires for a source that already painted. Reported by react-ui-engineering.
  **Fix:** A narrowed shared flag reaches every consumer in the same change, and an in-flight state is told by text or a labelled live region, never by an icon alone — pass `label` to `Spinner`, and thread `revalidatingBySource` into `SkillSourceList` as well.

packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts
```diff
@@ -100,158 +119,93 @@ export function useSkillsSurface(
...
+      const loadingBySource: Record<string, boolean> = {};
+      const revalidatingBySource: Record<string, boolean> = {};
+      const errorBySource: Record<string, ScanFailure | null> = {};
+      const scannedAtBySource: Record<string, string> = {};
+      const visibilityBySource: Record<string, "public" | "private"> = {};
+      results.forEach((result, index) => {
+        const src = sources[index];
+        if (!src) return;
+        const painted = result.data !== undefined;
+        loadingBySource[src.id] = result.isFetching && !painted;
```
dam-code-guardian, line 154: 🟡 **Warning:** This narrowing is the second site of the same finding. `loadingBySource` now means "fetching with nothing painted", but `skills-surface.tsx:106` and `:158` still hand it to `SkillsStoppedPanel` and `SkillsNeverRunPanel`, which map it to `SkillSourceList`'s `scanning` prop. A re-scan of a painted source therefore shows nothing on those two panels. Reported by react-ui-engineering.
  **Fix:** A narrowed shared flag reaches every consumer in the same change — thread `revalidatingBySource` into `SkillSourceList` too, or keep the old field's meaning and name the new distinction separately.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `8363ab6`

## PR #3586: fix(skills): the Skills page paints from what it last held

**Author:** xjacka | **Branch:** claude/issue-3278-735bd8 → main | **Changes:** +352 −289 (9 files)

### Summary

The round answers all six findings at their cause: the errored scan keeps its retained list, `patchState` refetches instead of dropping the write, both shared keys get one declaration in `modules/agents/api/skills.ts`, first load and revalidation are separate flags, and the architecture paragraph is bounded to the session. One warning is left — the split flag reaches only one of the three surfaces that showed a re-scan.

_(Suppressed 2 finding(s) per PR context: react-ui-engineering stale freshness stamp, issue-fit undeclared failure rendering.)_

_14 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: 8bb02a5 (2026-09-07T12:37:56Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** errored scan overwrites retained list with empty (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:162`)
- ✅ **Fixed:** patchState drops the write when the slot is empty (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:207`)
- ✅ **Fixed:** second skills.state declaration re-enables focus refetch (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:182`)
- ✅ **Fixed:** isFetching reports revalidation as first load (`packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:159`)
- ✅ **Fixed:** panel paragraph claims wider behavior than ships (`docs/architecture/skills.md:220`)
- ✅ **Fixed:** background rescan failure logged untyped to stderr (`packages/api-server/src/modules/skills/infrastructure/scan-cache.ts:108`)
- 🆕 **New:** the re-scan in-flight signal is lost on three surfaces (`packages/ui/src/modules/sandboxes/components/skills/skill-source-card.tsx:49`)
- 🆕 **New:** re-scan button stays enabled during its own re-scan (`packages/ui/src/modules/sandboxes/components/skills/skill-source-card.tsx:53`)

### Findings

- 🟡 **Warning:** The re-scan in-flight signal is lost on three surfaces (`packages/ui/src/modules/sandboxes/components/skills/skill-source-card.tsx:49`, `packages/ui/src/modules/sandboxes/hooks/use-skills-surface.ts:154`) — reported by react-ui-engineering
  **Fix:** A narrowed shared flag reaches every consumer in the same change, and an in-flight state is told by text or a labelled live region, never by an icon alone.
- 🟢 **Suggestion:** The re-scan `Button` is not disabled while `rescanning` is true, so each further click fires another `sources.refresh` mutation plus an invalidation. The previous branch replaced the whole freshness row with "Scanning…", so the control could not be clicked twice (`packages/ui/src/modules/sandboxes/components/skills/skill-source-card.tsx:53`)

Its finding — the page does not state that a card can show a retained list and a failure verdict together — fell under the 🟢 budget. The rewritten paragraph now bounds the client cache to the session and states the re-read trigger, which is what the previous round asked for.

### Issue Fit

Linked issue: #3278 "Skills page reloads everything from scratch on every visit" (`Closes #3278`). Its finding that the failure rendering is undeclared is suppressed: the round's own comment describes it in full ("the card shows the failure above the retained rows instead of in place of them"). Two further findings — the closing keyword landing with no run of the issue's browser repro, and the "not in scope" outcome being contradicted on the issue rather than with the reporter — fell under the 🟢 budget.

Its three blocking findings are regraded: the poll-versus-`patchState` race is not reachable, because `applyBatch` commits the rows before it returns and `state` reads them back, and the author states `cancelQueries` as the ordering rule; the `sets.list` options gap is unchanged from the previous HEAD; and the import of `modules/agents/api/skills.js` follows about twenty existing edges into that module. They and four suggestions fell under the 🟢 budget.

Its spinner finding is merged into `### Findings` (`skill-source-card.tsx:49`), and its `revalidatingBySource` finding into the same entry (`use-skills-surface.ts:154`). The freshness-stamp finding is suppressed: the author states the stamp is kept on purpose, so the age of the retained list stays visible. The re-scan button finding is regraded and printed as the review's suggestion; the write affordances over a stale list, and three suggestions, fell under the 🟢 budget.

### Verdict

COMMENT — every previous finding is closed at its cause; one warning is left, where the new revalidation flag reaches one surface of three.


---

xjacka:
Pushed da55267 for the remaining warning and the suggestion.

**The re-scan signal reached only one of three surfaces** (`skill-source-card.tsx:49`, `use-skills-surface.ts:154`) — correct, and it was my regression: narrowing `loadingBySource` to "fetching with nothing painted" silenced the two read-only source panels, which never got the new flag and so stopped saying anything during a re-scan over a list they already had. Both panels take `revalidatingBySource` now and read "Scanning…" from either flag, as they did before the narrowing. Those rows carry no skill list, so text in place of the stamp is still the right shape there — it is only the card, which keeps its rows, that needs the stamp to survive.

On the icon: the card's spinner now passes a label, so it renders as a `role="status"` region with text rather than a bare spinning ring. That is the existing `Spinner` contract, so nothing new had to be invented for it.

**Re-scan stayed enabled during its own re-scan** (`:53`) — fixed, and widened a little: the button and the kebab item are both disabled while a scan is in flight, on the card and in the panels. The card is where the old text-swap used to make a second click impossible, but the panels have the same two entry points into `sources.refresh` and the same double-fire.

Checks green: 1310 api-server and 337 UI unit tests, `mise run check` clean apart from `controller:check:staticcheck` (no binary in my environment; no Go changed). Still not exercised against a running cluster — the freshness row, the disabled states and the panel text are verified by reading, not by driving the page.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `da55267`

🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `da55267`

## PR #3586: fix(skills): the Skills page paints from what it last held

**Author:** xjacka | **Branch:** claude/issue-3278-735bd8 → main | **Changes:** +382 −293 (12 files)

### Summary

The round threads `revalidatingBySource` from the surface into both read-only panels and the source list, labels the two card spinners, and disables the Re-scan button and both kebab items while a scan is in flight. The last warning is closed at its cause on all three surfaces; nothing blocking is left.

_(Suppressed 4 finding(s) per PR context: the issue's second outcome, recorded as out of scope in the body; the error-plus-retained-list rendering, described in the round's own comment, twice; the list showing "Scanning…" in place of the stamp, stated as deliberate for rows that carry no list.)_

### Changes since last review
Previous HEAD: 8363ab6 (2026-09-07T21:32:26Z) — verdict COMMENT

- ✅ **Fixed:** the re–scan in–flight signal is lost on three surfaces (`packages/ui/src/modules/sandboxes/components/skills/skill-source-card.tsx:49`)
- ✅ **Fixed:** re–scan button stays enabled during its own re–scan (`packages/ui/src/modules/sandboxes/components/skills/skill-source-card.tsx:53`)
- 🆕 **New:** disabled tracks isFetching, not the refresh mutation (`packages/ui/src/modules/sandboxes/components/skills/skill-source-card.tsx:62`)
- 🆕 **New:** list error banner hides on every background revalidation (`packages/ui/src/modules/sandboxes/components/skills/skills-source-list.tsx:114`)
- 🆕 **New:** Last verified stamp moves backwards against main (`docs/architecture/skills.md:3`)

### Findings

- 🟢 **Suggestion:** Both new `disabled` props read the query's `isFetching` (`rescanning`, `loading || revalidating`), not the re-scan action. `refreshSource` awaits `sources.refresh` before anything sets a flag, so the control is live for that window, and a routine background revalidation past the 60-second stale window disables it and announces "Re-scanning this source" for a scan nobody asked for (`packages/ui/src/modules/sandboxes/components/skills/skill-source-card.tsx:62`)
- 🟢 **Suggestion:** In the list row the error banner is gated on `error && !scanning`, and `scanning` now includes revalidation, so an unreadable source's failure text disappears and returns on each automatic refetch (`packages/ui/src/modules/sandboxes/components/skills/skills-source-list.tsx:114`)
- 🟢 **Suggestion:** `docs/architecture/skills.md` keeps `Last verified: 2026-09-07` from the merge base while `main` already reads `2026-09-09`, so merging moves the stamp backwards on a page this branch rewrites (`docs/architecture/skills.md:3`)

Its `Last verified` finding is printed above as a suggestion — docs drift is one item per round, and the page's content is otherwise consistent with what ships.

### Issue Fit

Linked issue: #3278 "Skills page reloads everything from scratch on every visit" (`Closes #3278`). Its finding that the issue's second outcome is closed without being delivered is suppressed: the body records the descope and states the outcome already held before the change. Its error-rendering finding is the previous round's, suppressed on the same grounds as then. One suggestion — the disabled states and spinner labels going unmentioned in the change list — fell under the 🟢 budget.

Its `isFetching`-versus-mutation finding is printed above as a suggestion (`skill-source-card.tsx:62`), and its error-banner finding as the second (`skills-source-list.tsx:114`). The `!error` guard finding is the previous round's and stays suppressed. The finding that the list swaps the stamp for "Scanning…" where the card keeps it is suppressed: the author states the shape is deliberate, since those rows carry no skill list. Two suggestions — collapsing the six per-source records into one, and modelling the card's four status inputs as a discriminated union — fell under the 🟢 budget.

### Verdict

APPROVE — the narrowed flag now reaches every surface that showed a re-scan, and the residual is the new disabled state's precision, which is a suggestion.


---


---

Issue #3555, opened 2026-09-03
UI - An artifact is a dead end when the user wants to keep working on it


jenna-winkler:
## Context

Artifacts are where finished agent work lands. Users often want the next version, not the current one. Today the library treats an artifact as a final object: the row actions are share, rename, move, download, and delete. Peer products keep the artifact connected to the agent that made it, so the user can ask for a change from the artifact itself.

## Problem

A user looking at an artifact has no way to continue it. To iterate, the user must remember which agent made the artifact, open that agent, start a session, and describe the artifact from memory. Most users do not do this. The artifact stays as it is.

## Goal

From any artifact, a user can start work on it in one click, with the agent that made it and the artifact already in context.

## User Stories

- As a user, I want to continue working on an artifact from the artifacts page so that I do not have to find its agent myself.
- As a user, I want the new session to already know which artifact I mean so that I do not have to describe it again.
- As a user, I want the same affordance wherever an artifact appears so that the behaviour is predictable.

## Scope

**In:** an action on the artifact row in the artifacts page that opens a new session on the creating agent, with the artifact referenced in the message input.

**Not in:** artifacts a user uploaded (no creating agent), and the interactive-artifact direction under #2884.

## Proposed solution

Add a chat icon next to the existing share and overflow actions on the artifact row. It reuses the chat icon already used to open an agent's chat elsewhere in the app, so the meaning carries over without new wording. The click opens a new session on the creating agent and pre-fills the input with the artifact reference and a short line about iterating on it. The user reads the pre-filled text, edits it, and sends — nothing runs before the user sends.

## Open Questions

- Icon only, or an explicit label? A label states the intent, but every other row action is icon-only.
- What should the pre-filled text say? It must name the artifact and invite an edit, without reading as a command the user did not write.
- What happens for an artifact with no creating agent — hide the action, or let the user pick an agent?

laraneich12:
Design can be found here: https://www.figma.com/design/zNIYydUKN1QLZDYozpQJpn/DAM-DEV?node-id=2540-3825&t=gf1elNfI1hMXNlQ9-1

jenna-winkler:
Just one minor comment but looks great @laraneich12 !

Ready for dev

PR #3596, opened 2026-09-07
feat(artifacts): a user can edit the text of an artifact they own

### LINKED ISSUE #3556: UI - A user cannot change the text of an artifact they own
## Context

Files in a sandbox are editable in the UI. Artifacts are not, though they are the versioned, shareable copy.

## Problem

A user cannot change the text of an artifact they own. The row offers share, rename, move, download and delete; opening one gives a read-only preview.

Two workarounds, both bad:

- Ask the agent to redo it. Everything else in the document can change too.
- Download, edit locally, upload. Produces a separate artifact, so any link already shared keeps pointing at the old text.

Why it matters:

- Users read their own work closely right before sharing it, which is when they find the thing they want changed.
- Artifacts already version, and a new version keeps the same link, so the store already supports an edit. Only the UI is missing.
- It affects #3286 and #3277

## Goal

A user can change the text of an artifact they own, from where they read it. Identity, link and previous versions survive.

## Scope

**In:** editing artifact text; making the resulting version legible.

**Not in:** non-text artifacts (images, binaries, uploads). Sandbox file editing, which works. Handing an artifact back to its agent, which is #3555.

## Proposed solution

Add editing to the artifact preview, where the user already is. The write path exists: an artifact update already accepts new content, bumps the version and preserves the link, so this is a UI affordance over an API that works. The sandbox file editor is the precedent to reuse.

## Open Questions

- Does #3555 (hand the artifact back to its agent) cover this instead? A typo does not need an agent; a rewrite might.
- Should a version record who wrote it? Versions are steppable today but nothing distinguishes the owner's edit from the agent's publish.
- What happens when an agent publishes a new version over an owner's edit: overwrite, branch, or refuse?
- Does editing belong with rename, since both are the owner changing their own artifact?

PetrBulanek:
An owner can change the text of an artifact they own, from the preview they read
it in. Saving publishes a new version, so the artifact keeps its identity, its
share link and every earlier version — a link already shared keeps working and now
serves the corrected text.

Today the only ways to fix a line are to ask the agent to redo the whole document,
or to download it, edit it and upload it again, which mints a separate artifact and
leaves every shared link pointing at the old text.

Two things follow from letting owners write versions. A save states the version it
changed and is refused when an agent has published a newer one, so an owner never
silently replaces the agent's work — they are told, and may then overwrite
deliberately. And a version records who wrote it, so the history says which version
came from the owner and which from the agent.

Editing is offered wherever an owner already reads an artifact: the preview dialog,
the docked panel beside a chat, and an Edit entry in each row's menu. Only the
current version is editable, and only text small enough for one inline write —
never an image, and never a document too large to send back whole.

Not in scope: non-text artifacts, sandbox file editing, and handing an artifact
back to its agent (#3555).

Closes #3556

## Sub-issues

- [x] 01 — Version authorship and the stale-save guard
- [x] 02 — Editable source view and the preview dialog
- [x] 03 — Editing from the chat panel and the row menu
- [x] 04 — Version history with author and time


packages/ui/src/modules/artifacts/components/chat-artifacts-panel.tsx
```diff
@@ -121,12 +135,9 @@ export function ChatArtifactsPanel({
...
                     artifact={artifact}
                     active={artifact.id === openArtifactId}
-                    onClick={() =>
-                      setOpenArtifactId(
-                        artifact.id === openArtifactId ? null : artifact.id,
-                      )
-                    }
+                    onClick={() => void openRow(artifact.id)}
                     drag={dropCallbacks}
+                    onEdit={(a) => setOpenArtifactId(a.id, { edit: true })}
```
dam-code-guardian: 🔴 **Critical:** This calls the raw store setter, bypassing the `openRow` discard guard directly above it, and `setOpenArtifactId` unconditionally resets `openArtifactDirty: false`. Because the docked panel is keyed on `openArtifactId`, choosing Edit on any row while an unsaved draft is open destroys that draft with no prompt. `artifact-link-chip.tsx:43` reaches the same setter and is a second loss path now that a dirty buffer can exist.
**Fix:** Route every change of `openArtifactId` through one opener that awaits the discard confirmation, so no UI path reaches the raw setter while a draft is dirty.

packages/ui/src/modules/artifacts/components/artifact-row-menu-items.tsx
```diff
@@ -25,6 +28,11 @@ export function ArtifactRowMenuItems({
 
   return (
     <>
+      {isTextKind(artifact.kind) && (
```
dam-code-guardian: 🟡 **Warning:** Edit is offered for every non-binary kind, but editing needs the loaded text to fit `INLINE_CONTENT_MAX_BYTES` (2 MB) while the preview cap is 10 MB. For a text artifact between the two, `couldEdit` skips the content fetch, the `initialEdit` effect returns on `!content`, and the surface opens read-only with no explanation.
**Fix:** Compute the affordance from the rule that admits the action — gate the entry on kind and `sizeBytes` — or have the opened surface state the refusal; no entry in this feature may be offered and then do nothing.

packages/ui/src/modules/artifacts/hooks/use-artifact-editor.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  useEffect(() => {
+    if (editing) return;
+    setDraft(content?.content ?? "");
+    setBaseVersion(artifact?.version);
+  }, [content?.content, artifact?.version, editing]);
+
+  useEffect(() => {
+    if (!initialEdit || !content) return;
+    if (editable) setEditing(true);
```
dam-code-guardian, line 54: 🟡 **Warning:** The other half of the dead-affordance finding: when the artifact is text but over the editable size, this effect consumes the intent and never enters edit mode, so the menu's Edit silently does nothing.
**Fix:** Gate the affordance on the same predicate, or report the refusal here.

packages/ui/src/modules/artifacts/hooks/use-artifact-editor.ts
```diff
@@ -0,0 +1,126 @@
...
+  const [draft, setDraft] = useState(content?.content ?? "");
+  const [baseVersion, setBaseVersion] = useState(artifact?.version);
+
+  useEffect(() => {
+    if (editing) return;
+    setDraft(content?.content ?? "");
+    setBaseVersion(artifact?.version);
+  }, [content?.content, artifact?.version, editing]);
+
+  useEffect(() => {
```
dam-code-guardian, line 47: 🟡 **Warning:** `ArtifactPreviewDialog` passes no `onEditConsumed`, so `initialEdit` stays `true` for the dialog's whole life while this effect depends on `content`. The save mutation's own `getContent` invalidation, or any live-event refetch, re-runs it and forces `setEditing(true)` again after the user saved or cancelled.
**Fix:** Consume a one-shot intent exactly once and make it unobservable afterwards — seed `useState` from it, or require `onEditConsumed` on every caller.

packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts
```diff
@@ -313,22 +316,26 @@ export function createArtifactLibraryService(
...
+          kind,
+          contentType: stored.contentType,
+          fileName,
+          storageRef: stored.storageRef,
+          sizeBytes: stored.sizeBytes,
+          version: 1,
+          visibility: input.visibility ?? "private",
+          expiresAt: expiresAtFrom(input.expiresInHours),
+        },
+        agentId ? "agent" : "user",
```
dam-code-guardian, line 383: 🟡 **Warning:** `create` derives the author as `agentId ? "agent" : "user"` while `update` (line 376) yields `null` for the same missing attribution, and `attribution` is optional on both. The platform's own bookkeeping writes take both paths — experiment dashboards and script clones are created with `{ agentId, internal: true }` and later updated with no attribution — so one artifact's history reads "Agent" for v1 and blank for v2, while `docs/architecture/artifact-library.md:29` asserts such versions "name nobody".
**Fix:** Make `attribution` required on `create` and `update` with one discriminated type and one shared derivation, so every caller including the internal ones declares an author, then align the architecture page with what the code records.

packages/api-server-api/src/modules/artifact-library/types.ts
```diff
@@ -118,7 +124,11 @@ export interface ArtifactLibraryService {
     input: ArtifactCreateInput,
     attribution?: { agentId: string; internal?: boolean },
   ): Promise<LibraryArtifact>;
-  update(id: string, input: ArtifactUpdateInput): Promise<LibraryArtifact>;
+  update(
+    id: string,
+    input: ArtifactUpdateInput,
+    attribution?: ArtifactWriteAttribution,
```
dam-code-guardian, line 137: 🟡 **Warning:** The other location of the authorship finding: `attribution` optional here means the four programmatic `update` callers in `experiments-service.ts` compile unchanged and write `author: null`, so the type checker cannot tell an intentionally unattributed write from unwired plumbing.
**Fix:** Make the parameter required on both `create` and `update` with one discriminated attribution type, so every call site declares an author.

packages/ui/src/modules/artifacts/hooks/use-artifact-editor.ts
```diff
@@ -0,0 +1,126 @@
...
+interface Options {
+  artifact: LibraryArtifact | null | undefined;
+  content: ArtifactContent | null | undefined;
+  isHeadVersion: boolean;
+  initialEdit?: boolean;
+  onEditConsumed?: () => void;
+}
+
+function isConflict(err: unknown): boolean {
+  return err instanceof TRPCClientError && err.data?.code === "CONFLICT";
```
dam-code-guardian, line 21: 🟢 **Suggestion:** This is the third copy of the "edit a text blob and save with optimistic concurrency" flow — the preview dialog and the docked panel duplicate the toolbar and gating, and `packages/ui/src/modules/files/components/file-viewer.tsx:47` detects a conflict by regex-matching the error message where this reads `err.data?.code === "CONFLICT"`. One hook plus one toolbar component parameterized over the save mutation would leave a single home for the next change.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `be391c9`

## PR #3596: feat(artifacts): a user can edit the text of an artifact they own
**Author:** PetrBulanek | **Branch:** feat/3556-artifact-text-editing → main | **Changes:** +4000 −143 (28 files)

### Summary
In-app text editing for owned artifacts, with an `expectedVersion` stale-save guard, per-version authorship (`user` / `agent` / unattributed) and a version-history popover. The concurrency guard and the conflict-then-overwrite flow are sound; the unsaved-draft protection has a hole, and authorship is recorded inconsistently across the write paths.

### Findings

- 🔴 **Critical:** The sidebar Edit entry calls the raw store setter `setOpenArtifactId(a.id, { edit: true })`, which bypasses the `openRow` discard guard directly above it and unconditionally resets `openArtifactDirty: false`; because the docked panel is keyed on `openArtifactId`, choosing Edit on any row while an unsaved draft is open destroys that draft with no prompt (`packages/ui/src/modules/artifacts/components/chat-artifacts-panel.tsx:140`). The unchanged `artifact-link-chip.tsx:43` reaches the same setter and becomes a second loss path now that a dirty buffer can exist.
  **Fix:** Route every change of `openArtifactId` through one opener that awaits the discard confirmation, so no UI path reaches the raw setter while a draft is dirty.
- 🟡 **Warning:** The row menu offers Edit for every non-binary kind, but editing needs the loaded text to fit `INLINE_CONTENT_MAX_BYTES` (2 MB) while the preview cap is 10 MB; for a text artifact between the two, `couldEdit` skips the content fetch, the effect returns on `!content`, and the surface opens read-only with no explanation (`packages/ui/src/modules/artifacts/components/artifact-row-menu-items.tsx:31`, also `packages/ui/src/modules/artifacts/hooks/use-artifact-editor.ts:48`). Reported by issue-fit and react-ui-engineering.
  **Fix:** Compute the affordance from the rule that admits the action — gate the menu entry on kind and `sizeBytes` — or have the opened surface state the refusal; no entry in this feature may be offered and then do nothing.
- 🟡 **Warning:** `ArtifactPreviewDialog` passes no `onEditConsumed`, so `initialEdit` stays `true` for the dialog's life while the consuming effect depends on `content`; the save mutation's own `getContent` invalidation, or any live-event refetch, re-runs it and forces `setEditing(true)` again after the user saved or cancelled (`packages/ui/src/modules/artifacts/hooks/use-artifact-editor.ts:46`).
  **Fix:** Consume a one-shot intent exactly once and make it unobservable afterwards — seed `useState` from it, or require `onEditConsumed` on every caller.
- 🟡 **Warning:** Authorship is derived two ways for one concept: `create` uses `agentId ? "agent" : "user"` while `update` yields `null` when no attribution is passed, and `attribution` is optional on both. The platform's own bookkeeping writes take both paths — experiment dashboards and script clones are created with `{ agentId, internal: true }` (author `agent`) and later updated with no attribution (author `null`) — so one artifact's history reads "Agent" for v1 and blank for v2, and `docs/architecture/artifact-library.md:29` asserts such versions "name nobody" (`packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts:337`, also `packages/api-server-api/src/modules/artifact-library/types.ts:130`). Reported by doc-drift and typescript-engineering.
  **Fix:** Make `attribution` required on `create` and `update` with one discriminated type and one shared derivation, so every caller including the internal ones declares an author, then align the architecture page with what the code records.

The page is updated and its `Last verified:` date matches HEAD. The authorship paragraph contradicts the create path for platform-internal writes — merged into the authorship finding in `### Findings`. Its remaining suggestions fell to the 🟢 budget.

### Issue Fit
Linked issue: #3556 "UI - A user cannot change the text of an artifact they own" (closing keyword). The issue is met for text within the inline cap; the dead Edit affordance is the gap — merged into `### Findings`.

Optional `attribution` leaves the four programmatic `update` callers writing `author: null` — merged into the authorship finding in `### Findings`. Its remaining suggestions fell to the 🟢 budget.

The unguarded Edit call site and the replayed `initialEdit` are in `### Findings`.
- 🟢 **Suggestion:** The editing flow now exists three times — the preview dialog and the docked panel carry near-identical toolbar, content-gate and version bookkeeping, and `packages/ui/src/modules/files/components/file-viewer.tsx:47` is a third copy that detects a conflict by regex-matching the error message where this one reads `err.data?.code === "CONFLICT"`. One hook plus one toolbar component parameterized over the save mutation would leave a single home for the next change.

### Verdict
REQUEST_CHANGES — the sidebar Edit entry discards an open unsaved draft with no prompt, and three warnings remain around the dead Edit affordance, the replayed edit intent and inconsistent version authorship.


---

packages/ui/src/modules/artifacts/components/chat-artifacts-panel.tsx
PetrBulanek: Fixed in e408eefa. Every change of `openArtifactId` now goes through one opener, `useOpenArtifact`, which awaits the discard confirmation — but only when the id actually changes, so re-opening the artifact already held never claims to discard something it keeps. The row click, this Edit entry, `artifact-link-chip.tsx` and the two experiment call sites all route through it.

There was a second half to this that the finding does not name. Choosing Edit on the *already open* row keeps the key, so nothing remounts — but `setOpenArtifactId` still reset `openArtifactDirty` to `false`, and the sync effect in the docked panel did not re-run because its `editor.dirty` dep had not changed. The flag then stayed `false` while the buffer was dirty, so the guard was dead for every later row click too. The store now keeps the flag when the id is unchanged.

Two raw calls remain, both by intent: `closePanel` awaits `editor.confirmDiscard()`, which tests the live buffer more precisely than the store flag, and `use-artifact-deletion` has already asked a destructive confirm — a second "discard changes?" there would be noise.

packages/ui/src/modules/artifacts/components/artifact-row-menu-items.tsx
PetrBulanek: Fixed in e408eefa. The rule that admits the save now lives in one place — `isEditableArtifact` in `lib/editable.ts`, over kind and `sizeBytes`. This menu entry asks it, and so does `couldEdit` in both editing surfaces, which also removes the two duplicated inline expressions. The 2 MB–10 MB text band no longer offers an Edit that does nothing.

packages/ui/src/modules/artifacts/hooks/use-artifact-editor.ts
PetrBulanek: Fixed in e408eefa. The hook now holds the intent in its own state and clears it as it consumes it, so no caller can replay it. It still arms on a rising `initialEdit`, which the docked panel needs for Edit on the row it already holds.

One correction to the severity reasoning: this was not only a *possible* refetch. A save bumps the version, which changes the `getContent` query key, and `useSaveArtifactContent` invalidates that query as well. So the dialog re-entered edit mode after every save from the row-menu Edit path — the main path of sub-issue 03 for the artifacts view and the sandbox section.

packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts
PetrBulanek: Addressed in e408eefa, though the doc sentence was the wrong half of the pair.

The four platform `update` calls in `experiments-service.ts` now pass `{ agentId: driverAgentId }`. That is the honest value: the artifact's own `agentId` column already names that agent, and its v1 already reads "Agent". A dashboard's history no longer mixes a named author with a blank one. The architecture page now says the platform's bookkeeping writes name the agent they were made for, instead of claiming they name nobody.

I did not make `attribution` required on `create` and `update`. Every `update` caller now passes it, so the type change would lock in the convention rather than fix behaviour, and it costs four test edits in `artifact-library.test.ts` — say the word if you want the compiler holding that line. On `create` the case is different again: absent attribution means "user", which is correct for the UI upload path, so the default there carries meaning rather than hiding a gap.

packages/ui/src/modules/artifacts/hooks/use-artifact-editor.ts
PetrBulanek: Left as is for now — noted rather than dismissed. The fragile half is the one you name: `file-viewer.tsx:91` matches `/conflict|changed on disk/i` against the message text, where this hook reads `err.data?.code`. If the three flows get one home, that regex is the part to delete, not the part to carry over.

packages/ui/src/modules/artifacts/hooks/use-artifact-editor.ts
PetrBulanek: Covered by the reply on `artifact-row-menu-items.tsx`. The affordance and this effect now ask one predicate, `isEditableArtifact`, so the unreachable branch is gone rather than made to explain itself.

packages/api-server-api/src/modules/artifact-library/types.ts
PetrBulanek: Covered by the reply on `artifact-library-service.ts`. Every `update` caller now passes attribution, so no call site writes `author: null` by omission. I left the parameter optional deliberately — the reasoning is in that reply.

packages/ui/src/modules/artifacts/components/artifact-preview-dialog.tsx
```diff
@@ -28,67 +40,119 @@ import { VersionSwitcher } from "./version-switcher.js";
...
+    artifact: live ?? artifact,
+    content: content.data,
+    isHeadVersion: pinnedVersion === null,
+    initialEdit,
+  });
+  const wantSource = !renderable || showSource || editor.editing;
 
   return (
     <>
       <Modal widthClass="w-[860px]" onClose={onClose}>
```
dam-code-guardian: 🔴 **Critical:** this dialog mounts the same editor as the docked panel, but it guards no exit from it. `onClose` goes straight to the header close button and to `Modal`'s Escape handler (`modal.tsx:27`), so Escape or the X discards a dirty draft with no prompt — while the identical draft in `DockedArtifactPanel` is protected by `confirmDiscard`.

  **Fix:** every exit from a surface that holds a draft asks the editor first — route this dialog's close and its Escape key through `editor.confirmDiscard` before it dismisses, so no dismissal path can drop a dirty draft.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `e408eef`

## PR #3596: feat(artifacts): a user can edit the text of an artifact they own
**Author:** PetrBulanek | **Branch:** feat/3556-artifact-text-editing → main | **Changes:** +4084 −170 (33 files)

### Summary
This round answers the previous review at its cause: one guarded opener
(`useOpenArtifact`) owns every artifact swap, one predicate
(`isEditableArtifact`) owns the Edit affordance, the edit intent is consumed
once, and the four experiment writes state their agent. One path of the
unsaved-draft invariant stays open, on surfaces the previous round did not name.
_20 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: be391c9 (2026-09-08T13:57:39Z) — verdict REQUEST_CHANGES

- ✅ **Fixed:** sidebar Edit bypasses the unsaved-draft guard (`packages/ui/src/modules/artifacts/components/chat-artifacts-panel.tsx:140`)
- ✅ **Fixed:** Edit offered for text over the inline cap does nothing (`packages/ui/src/modules/artifacts/components/artifact-row-menu-items.tsx:31`)
- ✅ **Fixed:** initialEdit is never consumed in the dialog and replays (`packages/ui/src/modules/artifacts/hooks/use-artifact-editor.ts:46`)
- 🔁 **Still present:** create and update disagree on unattributed authorship (`packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts:337`)
- 🔁 **Still present:** third copy of the edit–and–save flow, divergent conflict test (`packages/ui/src/modules/artifacts/hooks/use-artifact-editor.ts:21`)
- 🆕 **New:** two more exits drop an unsaved artifact draft (`packages/ui/src/modules/artifacts/components/artifact-preview-dialog.tsx:82`)
- 🆕 **New:** page describes the docked preview as read–only (`docs/architecture/artifact-library.md:200`)

### Findings
- 🔴 **Critical:** two more exits drop an unsaved artifact draft (`packages/ui/src/modules/artifacts/components/artifact-preview-dialog.tsx:82`, also `packages/ui/src/modules/files/store.ts:62`) — full text in **React UI Engineering Review** and inline.

I own this as a round–1 miss. The invariant is "no path removes a dirty draft
without asking"; round 1 named one entry point instead of every path. The paths
are the sidebar Edit (fixed), the panel close (guarded), the dialog dismissal,
and the eviction of the artifact panel by the file panel.

- 🟢 **Suggestion:** `docs/architecture/artifact-library.md` describes the docked preview as a surface that "renders the selected artifact and follows new versions" (line 200), states the edit capability with no size bound (line 175), and asserts an attribution invariant the contract does not carry (line 28). The docked preview now hosts the editor, editability is gated on the inline–update cap (2 MiB) and not on the read cap (10 MiB), and `update`'s attribution stays optional. The PR body and issue #3556 also put uploads out of scope, but the gate tests text–ness and not provenance (issue-fit).

### Issue Fit
- 🔁 Folded into the documentation item above — the scope claim and the gate disagree on uploads.

- 🔁 2 finding(s) from the previous review still present (see review at be391c9): the optional attribution argument on `update`, and the duplicated edit–and–save flow with its shared `CONFLICT` test.
- Its other items are suggestions on code this delta does not change; they are dropped under the 🟢 budget.

- 🔴 **Critical:** the dialog mounts the same editor as the docked panel, but it guards no exit from it: `onClose` goes straight to the header close button and to `Modal`'s Escape handler (`modal.tsx:27`), so Escape or the X discards a dirty draft with no prompt. The same draft dies a second way — `setOpenFilePath` clears `openArtifactId` unconditionally, so opening any file in the tree unmounts the keyed `DockedArtifactPanel` (`packages/ui/src/modules/artifacts/components/artifact-preview-dialog.tsx:82`, also `packages/ui/src/modules/files/store.ts:62`).
  **Fix:** every exit from a surface that holds a draft asks the editor first — route the dialog's close and its Escape key through `confirmDiscard`, and gate the eviction of one docked panel by another on the evicted panel's dirty flag, so the guarded opener is the only way in and the only way out.
- Its remaining items are suggestions; they are dropped under the 🟢 budget.

### Verdict
REQUEST_CHANGES — the guarded opener is correct, but two more exit paths of the
same invariant discard a draft, and the attribution warning stays open.


- `packages/ui/src/modules/files/store.ts:62` — 🔴 **Critical:** the same defect class, second location. `setOpenFilePath` clears `openArtifactId` unconditionally, and the two docked panels share one slot, so opening any file in the tree unmounts the keyed `DockedArtifactPanel` and discards an unsaved artifact draft with no prompt. `useFileTree` prompts on `openFileDirty` only, so the guard `useOpenArtifact` centralizes is bypassed from the other lineage.
  
    **Fix:** every exit from a surface that holds a draft asks the editor first — gate this eviction on `openArtifactDirty` exactly as `useOpenArtifact` gates on it, or let one guard cover both slots.


---

packages/ui/src/modules/artifacts/components/artifact-preview-dialog.tsx
PetrBulanek: Fixed in 61ca72e2. Both exits now ask.

The dialog routes its close button and its Escape through `editor.confirmDiscard` before it dismisses. One wrinkle the finding does not name, which the obvious fix walks into: the confirm is a Radix `AlertDialog`, so Escape closes *that* and then keeps bubbling to the `Modal`'s own `window` keydown listener, which asks again. `showConfirm` writes a single dialog slot, so the second ask would have replaced the first dialog and orphaned its promise — Escape would have become a confirm the user could not dismiss with Escape. `requestClose` ignores a re-entrant call for that reason.

For the second location I took the `let one guard cover both slots` half of your fix rather than the narrower one. `useDockDraftGuard` reads both dirty flags, and both lineages ask it — `useOpenArtifact` and `useFileTree`. Gating only the file side on `openArtifactDirty` would have left the mirror bug standing: `setOpenArtifactId` clears `openFilePath` as well, and nothing checked `openFileDirty`, so opening an artifact dropped an unsaved *file* draft with no prompt. That half predates this PR — it is on main today — but the one-slot dock is what makes both directions possible, so one guard closes both rather than one bug being fixed and its reflection left.

Each slot setter now also clears the evicted slot's dirty flag. Without that the shared guard would eventually prompt to discard a draft that no longer exists, since nothing but the panel's unmount cleanup was resetting the flag.

`docs/architecture/artifact-library.md` now states the shared slot and the invariant, rather than describing the docked preview as a surface that only renders.

packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts
PetrBulanek: Still holding, and this round I can give the argument the last one missed.

The read side and the write side are different questions, and the finding treats them as one. Migration `0043` adds `author` as a nullable column with no backfill, so every version written before this PR reads `null` permanently. `ArtifactVersionInfo.author` therefore has to stay `ArtifactVersionAuthor | null` whatever the write signature says — `version-list.tsx` already renders that case by omitting the author rather than guessing one. Making `attribution` required on `create` and `update` would not remove a single `null` from any history the UI will ever show.

On the write side the behaviour you asked for is already in place. All six callers pass attribution: the four in `experiments-service.ts`, `mcp-tools.ts:279`, and `router.ts:69` with `{ user: true }`. No production path writes `author: null` by omission. What is left is the compiler holding a convention that no caller currently breaks, and it costs four test edits to buy.

One genuine inconsistency near this finding, if you would rather spend the round there: `create` derives the author by truthiness (`agentId ? "agent" : "user"`, line 337) while `update` derives it by key presence (`"agentId" in attribution`, line 379). Two tests for one concept. They agree today only because the union declares `agentId: string`, so I have left it alone in this PR — but that is the part that deserves one shared derivation, rather than the optionality.

dam-code-guardian: The decline is accepted and recorded for this PR: `attribution` stays optional on `create` and `update`, and future reviews will not raise it again. The migration argument settles it — `0043` adds a nullable `author` with no backfill, so the read type must accept `null` for all earlier versions, and the six write callers already pass attribution.

The inconsistency you name is recorded in its place: `create` tests the author by truthiness at line 337, and `update` tests it by key presence at line 379. If a later round spends a finding here, it is the shared derivation.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `0a8ce3d`

## PR #3596: feat(artifacts): a user can edit the text of an artifact they own
**Author:** PetrBulanek | **Branch:** feat/3556-artifact-text-editing → main | **Changes:** +4185 −183 (37 files)

### Summary
This round closes the last open path of the unsaved-draft invariant: both dialog
exits route through the editor's confirm, and one guard now covers both dock
slots in both directions. No blocking finding survives — the residual is
suggestions on code this round did not touch.
_18 suggestion(s) dropped under the 🟢 budget._
_(Suppressed 1 finding(s) per PR-local overrides: `types.ts:134`.)_

### Changes since last review
Previous HEAD: e408eef (2026-09-08T16:03:58Z) — verdict REQUEST_CHANGES — unreachable, reviewed the whole PR

- ✅ **Fixed:** two more exits drop an unsaved artifact draft (`packages/ui/src/modules/artifacts/components/artifact-preview-dialog.tsx:82`)
- ✅ **Fixed:** create and update disagree on unattributed authorship (`packages/api-server/src/modules/artifact-library/services/artifact-library-service.ts:337`)
- ✅ **Fixed:** third copy of the edit–and–save flow, divergent conflict test (`packages/ui/src/modules/artifacts/hooks/use-artifact-editor.ts:21`)
- ✅ **Fixed:** page describes the docked preview as read–only (`docs/architecture/artifact-library.md:200`)
- 🆕 **New:** page claims every dismissal asks; in-app navigation does not (`docs/architecture/artifact-library.md:276`)
- 🆕 **New:** dialog header keeps the pre-save size and file name (`packages/ui/src/modules/artifacts/components/artifact-preview-dialog.tsx:105`)
- 🆕 **New:** an agent version is named after the artifact's current agent (`packages/ui/src/modules/artifacts/components/version-list.tsx:22`)

### Findings
- 🟢 **Suggestion:** The new dock paragraph states that "every dismissal of a surface holding [an unsaved edit]" asks first, but `resetChatContext` clears both slots and both dirty flags unconditionally, and every in-app navigation calls it (`agents/store.ts:99,113,127,143`, `use-browser-history.ts:30,44`); `useUnsavedGuard` covers browser unload only. Narrow the sentence to the exits the code guards. The PR body's "sandbox file editing" scope line has the same shape — the shared dock guard does change the file viewer's discard behaviour, which the page itself states (doc-drift, issue-fit) (`docs/architecture/artifact-library.md:276`, also `packages/ui/src/modules/sessions/store/sessions.ts:205`).
- 🟢 **Suggestion:** The dialog re-reads `live` from the query cache and hands it to the editor, but every field it renders — title, file name, size — still comes from the `artifact` prop the parent holds in `useState` and never refreshes, so after an in-dialog save the header keeps the pre-save byte count; `couldEdit` at `:66` reads the same stale `sizeBytes` (react-ui-engineering) (`packages/ui/src/modules/artifacts/components/artifact-preview-dialog.tsx:105`, also `:56`).
- 🟢 **Suggestion:** Every `author === "agent"` row is labelled with the artifact's current `agentId`, but `ArtifactVersionInfo` carries the actor kind only, so a version written by another agent reads as this one. The generic "Agent" label the code already falls back to is the safe render until the payload carries the writer (react-ui-engineering) (`packages/ui/src/modules/artifacts/components/version-list.tsx:22`, also `packages/api-server-api/src/modules/artifact-library/types.ts:62`).

- 🔁 Folded into the first item in **Findings** — the page's universal dismissal claim is wider than the guarded exits.

### Issue Fit
- 🔁 Folded into the same item — the body's scope line and the shared dock guard disagree about the file viewer.

- 🔁 The authorship items are settled for this PR: the optional `attribution` argument is closed under a PR-local override, and its stated replacement — `create` deriving the author by truthiness while `update` derives it by key presence — is a placement item on a working diff, so it is a suggestion this round and dropped under the 🟢 budget.
- Its remaining items are suggestions on code this round does not change (the read-path cast, the plain `text` column, the positional repository arguments); they are dropped under the 🟢 budget.

- 🟢 2 of its items are carried in **Findings** above: the dialog's stale prop reads and the version author label.
- Its other items are suggestions on code this round does not change; they are dropped under the 🟢 budget.

### Verdict
APPROVE — every blocking finding of the previous rounds is closed at its cause, and nothing that survives verification blocks.


---

PR #3599, opened 2026-09-07
fix(api-server): remove every Agent-scoped Postgres record on delete, on any path

### LINKED ISSUE #3545: Deleting an agent directly in the cluster leaves its bindings, shares and skills behind
## Context

Operators sometimes remove agents at the cluster level rather than through the UI or API: stale agents of departed users, environment cleanups, incident response. A staging cleanup today removed nine hibernated agents that way. The cluster side was clean afterwards: workspaces, pods, secrets and jobs were all removed with the agent. The platform side was not. Related: #244 fixed the same class of leak for Envoy secrets.

## Problem

The platform runs two kinds of cleanup when an agent is deleted. One is a periodic sweep that removes leftover records for agents that no longer exist. It covers egress rules, pending approvals, registry pull secrets, connection grants, environment variables and invocations, and it works for any deletion path. The other is an "agent deleted" event that fires only when the deletion goes through the API. Everything hanging off that event is skipped when the agent is removed directly, and none of it is covered by the sweep:

- **channel bindings** — a Slack or Telegram channel stays bound to an agent that no longer exists, including a workspace's default-agent slot
- **knowledge-base shares** owned by the agent
- **skill records** attached to the agent
- **public agent profile**
- **usage reporting** never records the deletion, so the agent stays counted as live

Two ways to close it: extend the periodic sweep to these record kinds, which also heals whatever is already orphaned in every environment, or emit the deletion event when the platform observes a deletion it did not initiate.

## Expected Behavior

Deleting an agent ends in the same state whether it is deleted through the UI, the API, the lifetime reaper, or directly in the cluster.

## Actual Behavior

Only the API path runs the event-driven cleanups. A direct deletion leaves channel bindings, shares, skill records and the public profile pointing at a dead agent, and usage never learns the agent is gone.

## Steps to Reproduce

1. Create an agent and bind it to a channel, share a knowledge base from it, or make its profile public.
2. Delete the agent's custom resource directly in the cluster.
3. The agent disappears from the UI, but the binding, share and profile records remain, and they are still there after the half-hourly sweep has run.

### LINKED ISSUE #3597: Deleting an agent leaves its schedules, runtime-delivery records and API-key scopes behind on every deletion path
## Context

Follow-up to #3545, which covers cleanups that run only when the deletion goes through the API. This issue is the other half: records that nothing removes on any deletion path, API included. Deleting an agent cleans the cluster side (workspace, pod, secrets) and the platform removes some of the agent's Postgres records with it, but the list of what gets cleaned was assembled one table at a time and several tables were never added.

Nothing in Postgres references an agent by foreign key. The cluster object is the authoritative record, so the database cannot cascade on its own and every cleanup is hand-written per table. Agent names are also reusable, so leftovers are not just disk: a later agent with the same name inherits them.

## Problem

Deleting an agent, by any path, leaves these behind:

- **schedules** — the agent's schedules stay and keep firing on their recurrence indefinitely. Every fire fails because there is no agent to wake, but not before writing a fresh runtime-delivery record and a queued event for the dead agent, so orphaned schedules manufacture more orphaned data on every tick. The docs acknowledge that schedules survive deletion as orphans; nothing acts on it.
- **runtime-delivery bookkeeping** — the per-agent version counter and applied-config cursor are never removed, so there is one permanent row for every agent ever created. Events already delivered to an agent are never removed either, deleted or not. A new agent with the same name starts from the old agent's cursor instead of a clean slate.
- **experiments** driven by the agent, with their spans, are never closed out or removed.
- **API-key scoping** — a key scoped to the deleted agent keeps that agent's name in its scope, so a later agent with the same name is covered by the key without anyone granting it.
- **usage record snapshots** — the deleted agent's usage row rightly stays for billing, but it keeps the full harness-config and skills snapshots, although the skills docs say deletion clears them.

## Expected Behavior

Deleting an agent removes everything that existed only because of that agent, on every deletion path, and a new agent created under the same name starts with no inherited state. What stays behind on purpose (usage history for billing, library artifacts, the activity log) is a deliberate, documented list.

## Actual Behavior

Schedules, runtime-delivery records, experiments and API-key scopes for the deleted agent remain. The schedules keep firing and failing forever, and a same-named agent inherits all of it.

## Steps to Reproduce

1. Create an agent, add a recurring schedule to it, and create an API key scoped to it.
2. Delete the agent through the UI or API.
3. The schedule keeps firing on its recurrence; each fire fails and writes a new runtime-delivery row and queued event for the deleted agent. The key still lists the agent in its scope.
4. Create a new agent with the same name. The old schedule fires into it, and the old key covers it.

jezekra1:
Deleting an agent cleaned the cluster side but left Postgres behind in two ways. Cleanups that only ran on the API path (channel bindings, skills, knowledge-base shares, the usage mirror) were skipped when the custom resource was deleted directly in the cluster (#3545). And several tables were never cleaned on any path: schedules kept firing forever and re-created runtime-delivery rows for the dead agent on every tick, the runtime-delivery outbox row lived forever, running experiments were never closed, and API-key scopes kept the dead agent's name (#3597). Agent ids are generated unless the API caller supplies one, so a later Agent reusing an id is an edge case, but when it happened it inherited all of this.

## What changed

- **One declared list is the whole cleanup contract.** Every record kind says how to list its agent ids and how to clean up. The delete-time hooks are that list's cleanups, run in order inside the API delete. The 30-minute orphan sweep reads the same list, diffs each kind's ids against the live custom resources, and re-runs every cleanup for ids that no longer resolve. Sixteen kinds, one type, no hand-mirrored second list.
- **The cleanup sagas are gone.** Channel bindings, Telegram conversations, skills, knowledge-base shares, the usage mirror and the public profile used to clean up by reacting to the deletion event, which only fired on the API path and had no retry. Each module now exports its cleanup as a function and it sits in the list with the rest. Module ownership is unchanged; only the caller is. Sixteen kinds in total.
- **The deletion event carries no cleanup.** Both paths still emit `AgentDeleted`, but only for runtime reactions: the UI hint and the Slack worker stop. The sweep resolves the owner from a surviving channel, schedule or share row before emitting. `cleanupHooks` is now a required dependency of the agents service.
- **New cleanups:** schedules (queued fires cancelled), runtime-delivery outbox and events, running experiments (failed with a reason, open spans ended, invocations shed, results artifact minted) and draft experiments (removed), API-key scopes. Ending open spans lives in the reap helper the inactivity sweep shares, so an experiment reaped for inactivity now also gets its spans closed; before, they stayed `running` forever on that path too.
- **The sweep re-reads each candidate's CR right before reaping.** A row written for an agent created between the CR list and the table scan is never reaped from a live agent. The image-pull Secret is written before the CR exists and sits outside this guarantee; that race is pre-existing and narrowed here.
- **Usage mirror hygiene.** A created agent that reuses a row resets the row's runtime and config snapshots, and marking deleted clears them too and is idempotent, so a same-named agent starts clean and the deletion timestamp cannot drift under repeated sweeps.

## Deliberately not touched

- The session directory (`agent_sessions`) is kept on purpose so spend stays attributable after deletion. An earlier draft of this change removed it; reverted.
- Dispatched runtime events for live agents are still never pruned. That is a general retention gap, not an agent-deletion one.
- A schedule fire that is in flight at the moment of deletion can re-create the outbox row once; the queued job then drops on "not found" and the sweep reaps the rows within 30 minutes.
- A two-step cleanup whose first step changes the rows the sweep reads (share revoke, then snapshot purge) is not retried if the second step fails. Making deletion cleanup durable belongs to the eventing decision in #3288, not to a mechanism of its own.

## Docs

The persistence page owns the cleanup contract (Lifetime section); agent-lifecycle, runtime-delivery, experiments, skills, usage-tracking, channels and security-and-credentials carry one-liners and cross-links.

Closes #3545
Closes #3597


packages/api-server/src/sagas/agent-artifacts-sweeper.ts
```diff
@@ -47,9 +53,11 @@ export function createAgentArtifactsSweeper(
           );
         }
       }
+      emit({ type: EventType.AgentDeleted, agentId });
```
dam-code-guardian: 🟡 **Warning:** This emit omits `ownerSub`, but the API delete path sends it (`agents-service.ts:967`). `hintFor` returns `null` for an `AgentDeleted` with no `ownerSub` (`live-hints.ts:22`), so a sweep-detected deletion publishes no `agents` live hint and the owner's UI keeps the dead agent until a manual reload. That contradicts the PR body's claim that every subscriber reacts exactly as on an API delete, and `persistence.md:147` states the same equivalence unqualified.

**Fix:** Make `ownerSub` a required field of `AgentDeleted` and resolve the owner in the sweep before it emits, so no emitter can publish a partially filled copy of the event.

packages/api-server/src/sagas/agent-artifacts-sweeper.ts
```diff
@@ -1,17 +1,20 @@
...
 import { AGENTS_PLURAL } from "../modules/agents/infrastructure/labels.js";
 
 export interface AgentArtifactsSweeper {
   tick(): Promise<void>;
 }
 
+export interface AgentCleanupSource {
+  name: string;
+  listAgentIds: () => Promise<string[]>;
+  cleanup?: (agentId: string) => Promise<void>;
```
dam-code-guardian: 🟡 **Warning:** `cleanup` was required on `main` and is optional here, and both consumers branch on its presence: the sweep does `if (!source.cleanup) continue` below, and bootstrap derives the delete-time hooks with `flatMap((s) => s.cleanup ? [s.cleanup] : [])`. A source added to `agentCleanupSources` with a forgotten `cleanup` compiles clean and silently disappears from both paths — the exact drift this PR exists to remove. "A kind can no longer be in one and not the other" now holds by convention, not by type. The same shape repeats in `experiment-driver-cleanup.ts:12` and `experiments/compose.ts:139,188,207`.

**Fix:** Dependencies are required, never optional (`architecture/infrastructure.md:169`): give the sweep two separately typed inputs — cleaning sources with a required `cleanup`, and detect-only sources with no such field — and make `onReaped`, `now`, `artifactLibraryFor` and `agentsFor` required, with no-ops injected at the composition root.

packages/api-server/src/modules/experiments/services/reap-running-experiment.ts
```diff
@@ -0,0 +1,26 @@
...
+  ExperimentsRepository,
+} from "../infrastructure/experiments-repository.js";
+
+export async function reapRunningExperiment(
+  repo: ExperimentsRepository,
+  row: ExperimentRow,
+  error: string,
+  at: Date,
+): Promise<boolean> {
+  const flipped = await repo.transition(row.id, "running", "failed", {
```
dam-code-guardian, line 13: 🟡 **Warning:** This transition flips the experiment row only. `experiment_spans` cascades on a deleted row, so drafts are covered, but a reaped *running* experiment keeps its span rows at `status = 'running'` forever, and the trace feed renders them as live (`experiments-service.ts:301`). Issue #3597 names spans with this deliverable ("**experiments** driven by the agent, with their spans, are never closed out or removed"), and the PR's "Deliberately not touched" list does not mention them, so `Closes #3597` closes the issue with a named element unhandled.

**Fix:** Close a parent record's children in the same path — end every `running` span of a reaped experiment — or name them in the not-touched list with the reason.

packages/api-server/src/modules/experiments/services/experiment-driver-cleanup.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export interface ReapedExperiment {
+  id: string;
+  owner: string;
+  driverAgentId: string;
+}
+
+export interface CreateExperimentDriverCleanupDeps {
+  repo: ExperimentsRepository;
+  onReaped?: (row: ReapedExperiment) => Promise<void>;
```
dam-code-guardian: 🟡 **Warning:** Same defect class as `agent-artifacts-sweeper.ts:12`. `onReaped?` and `now?` are optional and called conditionally (`await deps.onReaped?.({…})`), so the service cannot tell a deliberately follow-up-less wiring from a forgotten one, and the terminal follow-up — invocation shedding, pin clear, dashboard snapshot — is silently skipped when the field is missing.

**Fix:** Dependencies are required, never optional (`architecture/infrastructure.md:169`): declare `onReaped` and `now` as required and call them unconditionally, injecting a no-op at the composition root when a caller wants no follow-up.

packages/api-server/src/modules/experiments/compose.ts
```diff
@@ -125,65 +130,99 @@ export function composeExperimentsForOwner(opts: {
...
+      repo,
+      reason: "experiment reaped for inactivity",
+      logTag: "experiment-inactivity",
+    }),
   });
 }
 
+export function createExperimentsCleanupHook(opts: {
+  db: Db;
+  artifactLibraryFor?: (owner: string) => ArtifactLibraryServiceImpl;
```
dam-code-guardian: 🟡 **Warning:** Same defect class as `agent-artifacts-sweeper.ts:12`. `artifactLibraryFor?` and `agentsFor?` are optional here and at lines 139 and 188, and `createReapFollowUp` resolves them with presence checks (`opts.agentsFor?.(owner)`, `opts.artifactLibraryFor ? … : null`). The capability toggles — does this reap cancel invocations, does it snapshot a dashboard — are hoisted into the consumer's `deps` instead of being owned by the injected dependency.

**Fix:** Dependencies are required, never optional (`architecture/infrastructure.md:169`): make both ports required and let each own its enabled or disabled behavior where it is built, so the follow-up code has no presence checks.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `67d0cdf`

## PR #3599: fix(api-server): remove every Agent-scoped Postgres record on delete, on any path
**Author:** jezekra1 | **Branch:** fix/agent-delete-cleanup → main | **Changes:** +721 −149 (33 files)

### Summary
One declared source list now drives both the delete-time cleanup hooks and the orphan sweep, and the sweep emits `AgentDeleted` for each confirmed orphan, so a cluster-side deletion ends in the same state as an API delete. The list is complete for the tables the two issues name, except experiment spans.

_4 suggestion(s) dropped under the 🟢 budget._

### Findings

- 🟡 **Warning:** The sweep emits `AgentDeleted` without `ownerSub`, but the API delete path sends it (`agents-service.ts:967`). `hintFor` returns `null` for an `AgentDeleted` with no `ownerSub`, so a sweep-detected deletion publishes no `agents` live hint and the owner's UI keeps the dead agent until a manual reload. This breaks the PR body's own claim that "every subscriber reacts exactly as on an API delete", and `persistence.md:147` states the same equivalence unqualified (`packages/api-server/src/sagas/agent-artifacts-sweeper.ts:56`)
  **Fix:** Make `ownerSub` a required field of `AgentDeleted` and resolve the owner in the sweep before it emits, so no emitter can publish a partially filled copy of the event.
- 🟡 **Warning:** Issue #3597 names spans with the experiments deliverable ("**experiments** driven by the agent, with their spans, are never closed out or removed"), but the driver cleanup only flips the row `running` → `failed` and deletes drafts. `experiment_spans` cascades on a deleted experiment row, so drafts are covered, but a reaped running experiment keeps its span rows at `status = 'running'` forever, and the trace feed renders them as live (`experiments-service.ts:301`). The PR's "Deliberately not touched" list does not mention them, so `Closes #3597` closes the issue with a named element unhandled (`packages/api-server/src/modules/experiments/services/reap-running-experiment.ts:13`)
  **Fix:** Close a parent record's children in the same path — end every `running` span of a reaped experiment — or name them in the not-touched list with the reason.
- 🟡 **Warning:** `AgentCleanupSource.cleanup` became optional in this PR (it is required on `main`), and both consumers branch on its presence: the sweep does `if (!source.cleanup) continue` and bootstrap derives the delete-time hooks with `flatMap((s) => s.cleanup ? [s.cleanup] : [])`. A source added to `agentCleanupSources` with a forgotten `cleanup` compiles clean and silently disappears from both paths, which is the exact failure mode this PR exists to remove — "a kind can no longer be in one and not the other" holds only by convention now, not by type. The same optional-dependency shape repeats in the new driver-cleanup service and its composers (`packages/api-server/src/sagas/agent-artifacts-sweeper.ts:12`)
  **Fix:** Dependencies are required, never optional (`.agents/skills/typescript-engineering/architecture/infrastructure.md:169`): give the sweep two separately typed inputs — cleaning sources with a required `cleanup`, and detect-only sources with no such field — and make `onReaped`, `now`, `artifactLibraryFor` and `agentsFor` required, with no-ops injected at the composition root.
- 🟡 **Warning:** `security-and-credentials.md` owns the identity-and-auth records (`persistence.md:65`), and still calls the API-key binding "an optional agent allowlist" with no lifetime statement, although this PR mutates that allowlist on every deletion path. Five sibling pages received a one-liner in this same diff; the page that owns the record did not, so its reader cannot learn the key's scope set changes under it (`docs/architecture/security-and-credentials.md:164`)
  **Fix:** When a PR gives an existing record kind a new lifetime rule, update the architecture page that owns that record, not only the page that owns the cleanup mechanism.
- 🟢 **Suggestion:** `compose.ts` now carries orchestration rather than wiring: `createReapFollowUp` sequences invocation cancellation, pin clearing and dashboard snapshotting with its own try/catch and stderr reporting, beside two more logic-bearing exports. A composition root is meant to construct and pass; this use case belongs in `services/` (`packages/api-server/src/modules/experiments/compose.ts:141`)
- ✅ **Looks good:** The pre-reap `getCustomObject` re-read closes the create-between-list-and-scan race the sweep would otherwise lose, and it matters more than it looks: `markAgentDeleted` soft-deletes the usage row, so without the re-read a newly created agent's billing row could be marked deleted. `markAgentDeleted` is also now idempotent (`isNull(deletedAt)` guard), so a repeated sweep cannot drift the timestamp.

### Documentation Check

- 🟡 **Warning:** `security-and-credentials.md` still describes the API-key model as carrying "an optional agent allowlist" with no lifetime statement, although a delete now strips the agent id from every key's `agent_ids` (`removeAgentIdFromScopes`, wired into both the hooks and the sweep). See `### Findings` (`docs/architecture/security-and-credentials.md:164`)
  **Fix:** When a PR gives an existing record kind a new lifetime rule, update the architecture page that owns that record, not only the page that owns the cleanup mechanism.

`channels.md` attributing deletion cleanup to the API path only was also reported. The PR body states the page sits at its size cap and documents the backstop on `persistence.md` instead, which `persistence.md:147` confirms. Not a finding.

### Issue Fit

- 🟡 **Warning:** Issue #3597 names spans as part of the experiments deliverable, but a reaped running experiment keeps its `experiment_spans` rows at `status = 'running'`, and the not-touched list does not mention them. See `### Findings` (`packages/api-server/src/modules/experiments/services/reap-running-experiment.ts:13`)
  **Fix:** Close a parent record's children in the same path, or name them in the not-touched list with the reason.

Both issues' other named tables are covered: schedules, runtime-delivery outbox and events, API-key scopes, and the usage snapshot reset are each in the source list, on both paths.

- 🟡 **Warning:** `AgentCleanupSource.cleanup` is optional and every consumer branches on its presence, so a source with a forgotten `cleanup` silently vanishes from the delete path and the sweep. The same shape repeats in `experiment-driver-cleanup.ts:12` (`onReaped?`, `now?`) and in `compose.ts:139`, `compose.ts:188` and `compose.ts:207` (`artifactLibraryFor?`, `agentsFor?`). See `### Findings` (`packages/api-server/src/sagas/agent-artifacts-sweeper.ts:12`)
  **Fix:** Dependencies are required, never optional: split cleaning and detect-only sources into two types, and make the follow-up ports required with no-ops injected at the composition root.
- 🟡 **Warning:** The sweep emits `AgentDeleted` with no `ownerSub`, so `live-hints.ts:22` publishes no hint and subscribers that need the field degrade silently. See `### Findings` (`packages/api-server/src/sagas/agent-artifacts-sweeper.ts:56`)
  **Fix:** Make `ownerSub` required on `AgentDeleted` and resolve the owner in the sweep before it emits.
- 🟢 **Suggestion:** `compose.ts` carries application logic rather than wiring — `createReapFollowUp` orchestrates invocation cancellation, pin clearing and dashboard snapshotting. See `### Findings` (`packages/api-server/src/modules/experiments/compose.ts:141`)

### Verdict
COMMENT — the delete-and-sweep architecture is sound, but spans stay `running` on a reaped experiment, the sweep's event drops the owner so the UI never refreshes, and the optional `cleanup` field re-opens the drift this PR closes.


---

jezekra1:
Addressed the review and rebased onto main.

- **Owner on the sweep's `AgentDeleted`.** The sweep now resolves the owner before it cleans anything, from whichever of the agent's channel, schedule or share rows still exists, and puts it on the event. The field stays optional: an orphan can have no owner-bearing row left at all, and dropping the event in that case would skip the cleanups that matter more than the hint. Worth noting the UI was never stale here: the live-events agent watch already publishes the `agents` hint on a `DELETED` custom resource, owner taken from the CR's labels.
- **Spans.** Reaping a running experiment now ends its open spans as `error`, on both the inactivity and the driver-deleted path.
- **Types.** Cleaning sources and detect-only detectors are separate types; a source without a cleanup no longer compiles, and the hooks derive from the cleaning list with a plain `map`. The reap follow-up's pin, artifact-library and agents ports are required. The driver-deleted path injects a no-op pin, which is also a correctness fix: clearing an annotation on a deleted CR throws, and that would have skipped the dashboard snapshot.
- **Compose.** The reap follow-up and invocation cancellation moved out of `compose.ts` into a service.
- **Docs.** `security-and-credentials` states the API-key allowlist lifetime. Two pages needed tightening to stay under the size cap; `agent-lifecycle` on main is at exactly the cap.

packages/api-server/src/sagas/agent-artifacts-sweeper.ts
```diff
@@ -36,8 +44,11 @@ export function createAgentArtifactsSweeper(
 
     if (orphans.size === 0) return;
 
-    const toDelete = [...orphans].slice(0, deps.batchSize);
-    for (const agentId of toDelete) {
+    let reaped = 0;
+    for (const agentId of [...orphans].slice(0, deps.batchSize)) {
+      if (await deps.k8s.getCustomObject(AGENTS_PLURAL, agentId)) continue;
+      reaped++;
+      const ownerSub = await deps.resolveOwner(agentId);
```
dam-code-guardian, line 47: 🟡 **Warning:** `resolveOwner` reads channels, schedules and kb-shares. An orphan detected through any other kind — api-keys, agent-env, egress-rules, runtime-delivery, usage, telegram-conversations — gets an `AgentDeleted` with no `ownerSub`, and `hintFor` returns `null` for that event (`modules/live-events/sagas/live-hints.ts:21`), so the owner's UI gets no live hint. `docs/architecture/persistence.md:57` states the opposite: the sweep "emits the deletion event, so every subscriber reacts as on an API delete".
  **Fix:** Resolve the owner from a source that covers every kind that can detect an orphan — the soft-deleted `agents` mirror row holds one for any agent the platform ever saw — and state in the docs which consumers a partial event does not reach.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `d409f7c`

## PR #3599: fix(api-server): remove every Agent-scoped Postgres record on delete, on any path
**Author:** jezekra1 | **Branch:** fix/agent-delete-cleanup → main | **Changes:** +880 −195 (36 files)

### Summary
The cleanup contract is now one list of sources that both the delete path and the orphan sweep read, the sweep re-reads the CR before it reaps, and all five prior findings are resolved. Three warnings remain: the reconstructed deletion event carries no owner for most kinds, and two pages describe a reach the code does not have. _3 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: 67d0cdf (2026-09-07T18:09:00Z) — verdict COMMENT

- ✅ **Fixed:** sweep emits AgentDeleted without ownerSub, no live hint (`packages/api-server/src/sagas/agent-artifacts-sweeper.ts:56`)
- ✅ **Fixed:** reaped experiment keeps its spans at status running (`packages/api-server/src/modules/experiments/services/reap-running-experiment.ts:13`)
- ✅ **Fixed:** optional cleanup hides a forgotten wiring from the type checker (`packages/api-server/src/sagas/agent-artifacts-sweeper.ts:12`)
- ✅ **Fixed:** API-key page states no lifetime for the agent allowlist (`docs/architecture/security-and-credentials.md:164`)
- ✅ **Fixed:** compose.ts carries reap orchestration, not wiring (`packages/api-server/src/modules/experiments/compose.ts:141`)
- 🆕 **New:** the sweep resolves an owner from three kinds only, so its event stays weaker than an API delete (`packages/api-server/src/sagas/agent-artifacts-sweeper.ts:51`)
- 🆕 **New:** the channels page still names the cleanup saga as the only deletion path (`docs/architecture/channels.md:136`)
- 🆕 **New:** page edited in the 2026-09-08 commit keeps its 2026-09-07 Last verified (`docs/architecture/agent-lifecycle.md:3`)
- 🆕 **New:** the extracted follow-up service builds its own repositories from a raw Db (`packages/api-server/src/modules/experiments/services/reap-follow-up.ts:43`)

### Findings
- 🟡 **Warning:** the sweep resolves an owner from three kinds only, so its event stays weaker than an API delete (`packages/api-server/src/sagas/agent-artifacts-sweeper.ts:51`, also `docs/architecture/persistence.md:57`)
- 🟡 **Warning:** the channels page still names the cleanup saga as the only deletion path (`docs/architecture/channels.md:136`)
- 🟡 **Warning:** `agent-lifecycle.md` is edited by `d409f7c` (2026-09-08), which rewrites the Postgres-row sentence in the Delete section, but its `Last verified:` header still reads 2026-09-07 — the value on `origin/main`. `security-and-credentials.md`, edited in the same commit, was bumped to 2026-09-08. Reported by doc-drift. (`docs/architecture/agent-lifecycle.md:3`)
  **Fix:** Bump `Last verified:` on every architecture page an edit touches, on the last commit that touches it — a page edited twice in a PR carries the later date.
- 🟢 `reap-follow-up.ts` sits in `services/` but takes a raw `db: Db` and builds `createInvocationsRepository` and `createDashboardSnapshotter` itself, and imports the invocations repository factory across a module boundary. Its sibling `dashboard-snapshot.ts` already does this on `main`, so the extraction follows the module's shape; taking built ports would keep every adapter in `compose.ts`. Reported by typescript-engineering. (`packages/api-server/src/modules/experiments/services/reap-follow-up.ts:43`)

🔁 3 finding(s), all in `### Findings` above.

### Issue Fit
_2 finding(s) not upheld: the span close on the inactivity path is what the previous review required, and the API-key paragraph rewording is prose scope, not a defect._

🔁 1 finding(s), in `### Findings` above.

### Verdict
REQUEST_CHANGES — every prior finding is resolved, but the reconstructed deletion event carries no owner for most kinds, and the pages that describe the new contract state a reach the code does not have.


- `docs/architecture/channels.md:136` — 🟡 **Warning:** Channel bindings and Telegram conversations are now sweep detectors (`packages/api-server/src/bootstrap.ts:895`), so a deletion done in the cluster clears their rows through the sweep. This line still names the channel-cleanup saga as the only path, and the page's `Last verified` stays 2026-09-04. Reported by doc-drift.
    **Fix:** When a change gives a subsystem a second path into an existing behavior, update the page that owns that concept and bump its `Last verified`, not only the cross-cutting page.


---

jezekra1:
On the second-round findings:

- **Owner reach of the sweep's event.** The suggested source does not work: the usage mirror's `owner_sub` is HMAC-pseudonymized at the write boundary (see usage-tracking, "Determinism is load-bearing"), and the live-hint bus is keyed by the raw Keycloak sub, so a hint published to that value reaches nobody. The raw owner survives deletion only in kinds that carry it: channels, schedules, KB shares, experiments, invocations, API keys. The first three cover the saga-backed subscribers, which are the ones the event exists for. The only owner-dependent subscriber is the UI hint, and the live-events agent watch already publishes it when the custom resource disappears, owner taken from the CR's labels, so no user-visible gap remains. Took the review's alternative fix: persistence now states exactly when the event names an owner and which consumer that affects.
- **channels page** now says rows are cleared on any deletion path, with its `Last verified` bumped. The page sits 17 characters under its size cap, so the mechanism detail had to go rather than grow.
- **agent-lifecycle `Last verified`** bumped to 2026-09-08.
- **Follow-up service building its own repositories** mirrors its sibling `dashboard-snapshot.ts` on main; left as is to keep the module's shape consistent. Happy to lift both to built ports in a follow-up.

dam-code-guardian (dismissed):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `b301e0d`

## PR #3599: fix(api-server): remove every Agent-scoped Postgres record on delete, on any path
**Author:** jezekra1 | **Branch:** fix/agent-delete-cleanup → main | **Changes:** +884 −199 (37 files)

### Summary
Round 3 closes every blocking finding of round 2: the sweep's `AgentDeleted` now carries `ownerSub` where an event names one and the docs say so, and the three stale doc claims are corrected. Nothing blocks. _5 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: d409f7c (2026-09-08T07:44:52Z) — verdict REQUEST_CHANGES
- ✅ **Fixed:** the sweep resolves an owner from three kinds only, so its event stays weaker than an API delete (`packages/api-server/src/sagas/agent-artifacts-sweeper.ts:51`)
- ✅ **Fixed:** the channels page still names the cleanup saga as the only deletion path (`docs/architecture/channels.md:136`)
- ✅ **Fixed:** page edited in the 2026-09-08 commit keeps its 2026-09-07 Last verified (`docs/architecture/agent-lifecycle.md:3`)
- ✅ **Fixed:** the extracted follow-up service builds its own repositories from a raw Db (`packages/api-server/src/modules/experiments/services/reap-follow-up.ts:43`)
- 🆕 **New:** two edited architecture pages keep their base-branch Last verified (`docs/architecture/experiments.md:3`)
- 🆕 **New:** agent_case_studies absent from the kept-on-purpose list (`docs/architecture/persistence.md:147`)
- 🆕 **New:** cleanup reach is published as table queries, not one named port (`packages/api-server/src/modules/usage/index.ts:1`)

### Findings
- 🟢 **Suggestion:** `experiments.md` and `skills.md` gain a body paragraph in this PR but keep the `Last verified: 2026-09-07` of the base branch, while the six sibling pages moved to `2026-09-08`. The documentation guidelines ask for a bump on every edited page (`docs/guidelines/documentation-guidelines.md:34`), so the stamp no longer says which pages were re-read against the final state (`docs/architecture/experiments.md:3`, also `docs/architecture/skills.md:3`). Reported by doc-drift.
- 🟢 **Suggestion:** `agent_case_studies` rows are agent-scoped and survive deletion on purpose, which `case-studies.md:58` states, but the new kept-on-purpose list does not name them. The list is the page a reader consults for what stays behind (`docs/architecture/persistence.md:147`). Reported by issue-fit.
- 🟢 **Suggestion:** each module now publishes its cleanup reach as separate table queries (`listUsageAgentIds`, `listApiKeyAgentIds`, `listKbShareAgentIds`, `findChannelOwnerByAgent`, …) and bootstrap assembles them, with `schedulesBoot.repo` and `runtimeDelivery.outboxRepo` called directly. One named `listAgentIds` / `findOwnerByAgent` / `cleanup` port per module would let a new record kind join the sweep without a new export (`packages/api-server/src/modules/usage/index.ts:1`, also `packages/api-server/src/bootstrap.ts:858`). Reported by typescript-engineering.

✅ No findings beyond the one merged into `### Findings`.

### Issue Fit
✅ No findings beyond the one merged into `### Findings`. Both `Closes` issues are covered: #3545's delete-time cleanup per record kind, and #3597's documented deliberate-keeps list.

✅ No findings beyond the one merged into `### Findings`. The service-owns-its-repository reading of `reap-follow-up.ts` is the shape `dashboard-snapshot.ts` already has on `main`, and a detector whose subscriber fails still lists its rows on the next tick, so neither is a defect here.

### Verdict
APPROVE — every blocking finding of the previous rounds is resolved, and the remaining notes are documentation and boundary suggestions.


---

jezekra1:
Fourth commit unifies the two cleanup mechanisms. The channel, skills, knowledge-base share and usage deletion sagas are removed; each module exports its cleanup as a function and it joins the single source list, so the API delete and the orphan sweep run exactly the same code for every kind. The sweep's detect-only "detectors" type introduced in the second commit is gone with them.

`AgentDeleted` is still emitted on both paths, but it carries no cleanup any more: its consumers are the UI hint, the Slack worker stop and the public profile projection. That resolves the second-round concern about the event's reach in a different way: nothing that must happen depends on the event.

PR description updated to match.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `32cbe57`

## PR #3599: fix(api-server): remove every Agent-scoped Postgres record on delete, on any path
**Author:** jezekra1 | **Branch:** fix/agent-delete-cleanup → main | **Changes:** +895 −331 (42 files)

### Summary
Round 4 finishes the inversion: the three `AgentDeleted` cleanup sagas and the `markAgentDeleted` arm of `persist-agents` are gone, so all fifteen kinds are cleaned from the one declared list on both paths and the event only notifies. Both prior suggestions are closed; the skills page now contradicts itself about which mechanism reaps the publish rows. _3 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: b301e0d (2026-09-08T08:47:41Z) — verdict APPROVE
- ✅ **Fixed:** agent_case_studies absent from the kept-on-purpose list (`docs/architecture/persistence.md:147`)
- ✅ **Fixed:** cleanup reach is published as table queries, not one named port (`packages/api-server/src/modules/usage/index.ts:1`)
- 🔁 **Still present:** three edited architecture pages keep their base-branch Last verified while the five siblings edited in the same change moved to the PR date (`docs/architecture/experiments.md:3`)
- 🆕 **New:** the skills page still names the deleted cleanup saga as what reaps the publish rows, four lines above its own corrected bullet (`docs/architecture/skills.md:138`)
- 🆕 **New:** the shared reap helper makes the inactivity path also flip open spans to error, a behavior change the PR body does not declare (`packages/api-server/src/modules/experiments/services/reap-running-experiment.ts:18`)

### Findings

- 🟡 **Warning:** The Delete Local bullet still says `agent_skill_publishes` rows are "reaped only by the `AgentDeleted` cleanup saga", but this commit deleted `modules/skills/sagas/skills-cleanup.ts` and the event now carries no cleanup at all. The page's own Cleanup bullet four lines below was rewritten to the delete-by-agent form, so the same page states both mechanisms and a reader cannot tell which one holds (`docs/architecture/skills.md:138`)
  **Fix:** When a mechanism named on an architecture page is removed, grep the whole page for every reference to its name and update all of them in the same commit — correcting the bullet that defines it is not enough while another section still names it.
- 🟢 **Suggestion:** `experiments.md`, `skills.md` and `runtime-delivery.md` all carry a body edit in this PR but keep the base branch's `Last verified: 2026-09-07`, while the five sibling pages edited in the same change moved to `2026-09-08`. The stamp is what tells a reader which pages were re-read against the final state (`docs/architecture/experiments.md:3`, also `docs/architecture/skills.md:3`, `docs/architecture/runtime-delivery.md:3`). Reported by doc-drift and by issue-fit.

- 🟡 **Warning:** `skills.md:138` still attributes the publish-row reaping to the deleted `AgentDeleted` cleanup saga while the Cleanup bullet below it names the new mechanism. See `### Findings` (`docs/architecture/skills.md:138`)
  **Fix:** Grep the whole page for the removed mechanism's name, not only the section that defines it.

Two further reports do not survive verification as stated. `usage-tracking.md:29` — the Mermaid node label `agent CM created / deleted` feeds the bus, not the saga; the bus still carries `AgentDeleted` (the Slack worker and the public profile subscribe), so the edge is accurate and only the saga's own bullet needed the correction it received. `channels.md:138` — the sentence describes Slack **worker registration**, which `channel-manager.ts:508` still drives from `AgentDeleted`; that reaction survives this PR unchanged. The `persistence.md:147` enumeration is complete against the fifteen declared sources: `agent-env` appears as "environment variables".

### Issue Fit

One undeclared widening was reported and dropped under the 🟢 budget: sharing `reapRunningExperiment` means an experiment reaped for inactivity now also has its open spans flipped to `error` with an `endedAt` (`reap-running-experiment.ts:18`), which the PR body does not mention. The report's second half does not hold — `bootstrap.ts:928` already passed `pin`, `artifactLibraryFor` and `agentsFor` on `main`, so no follow-up became unconditional.

The `agent_case_studies` scenario does not reproduce: agent ids come from `generateK8sName` (random hex, `configmap-mappers.ts:3`), so a recreated agent never inherits the row or collides on the week index — the residual is an orphaned row, not wrong data, and the kept-on-purpose list now names it. The public-profile gap is closed by the `usage-agents` source: `persist-agents` writes an `agents` row for every `AgentCreated` unconditionally, and `listUsageAgentIds` lists every row with `deletedAt IS NULL`, so an orphan is always detected and the emitted `AgentDeleted` reaches `retireProfile` regardless of whether a channels row survives. Both `Closes` issues are covered.

One type-surface note was reported and dropped under the 🟢 budget: `endOpenSpans(experimentId, status: SpanStatus, endedAt)` accepts `"running"` where the neighbouring `endSpan` narrows to `Exclude<SpanStatus, "running">`, so the new method can write a span carrying both an `endedAt` and a `running` status (`experiments-repository.ts:135`).

Three blocking reports do not survive verification. `bootstrap.ts:552` — the `systemAgents` instance composed without `cleanupHooks` has exactly two consumers, `slackWorker` (649) and `telegramWorker` (684), and neither calls `.delete`; the sweep and the API path both go through `harnessAgentsServiceFor` (1003), which does pass the hooks. `bootstrap.ts:987` — the `periodicJobs.start()`-before-declaration window is the ordering `origin/main` already has (`start()` 921, `harnessAgentsServiceFor` 945, closed over by `invocationsCleanupHook` at 797), so it is off-delta, and BullMQ's immediate-tick premise is not read off this code. `reap-running-experiment.ts:19` — a throw in `endOpenSpans` does leave the row `failed` with spans open, but that is the same two-write shape `main` has between `transition` and its emit; the transaction argument is a design note here, not a defect this PR introduces. The `reap-follow-up.ts` boundary report is settled by the author's dismissal on this PR.

### Verdict
COMMENT — the cleanup inversion is complete and both prior notes are closed, but the skills page now names two different mechanisms for the same rows four lines apart.


- `docs/architecture/skills.md:138` — 🟡 **Warning:** This bullet still says `agent_skill_publishes` rows are "reaped only by the `AgentDeleted` cleanup saga", but this commit deleted `modules/skills/sagas/skills-cleanup.ts` and the event now carries no cleanup. The Cleanup bullet four lines below already names the new mechanism, so the page states both.
  
  **Fix:** When a mechanism named on an architecture page is removed, grep the whole page for every reference to its name and update all of them in the same commit.


---

jezekra1:
Round 4 response, in the fifth commit:

- **skills page** no longer names the deleted cleanup saga anywhere; the Delete Local bullet now says the publish rows are reaped by the agent cleanup on deletion.
- **Last verified** bumped on experiments, skills and runtime-delivery, the three pages that carried a body edit with the base-branch date.
- **Span end status** narrowed to `Exclude<SpanStatus, "running">`, matching the neighbouring `endSpan`, so an ended span can no longer carry `running`.
- **Inactivity path widening** is now declared in the PR body: sharing the reap helper means an experiment reaped for inactivity also gets its open spans closed. Before, they stayed `running` forever on that path as well.

One correction to the PR body prompted by the round-4 verification: agent ids are generated unless the API caller supplies one, so a later Agent reusing an id is an edge case rather than the norm. The inheritance fixes still hold for that case.

docs/architecture/persistence.md
```diff
@@ -141,10 +141,10 @@ Workspace volumes created before the ReadWriteOnce cutover live on shared-writab
...
... against the live custom resources and re-runs every cleanup for each id that no longer resolves. The deletion event carries no cleanup; both paths emit it only to notify projections and runtime reactions — the UI hint, the Slack worker registration, the public profile. The sweep names the owner on it only when a surviving channel, schedule or share row still records one; the UI hint is the one consumer that needs it, and the agent watch already delivers that hint when the custom resource disappears. It re-reads the candidate's CR right before reaping, so a row written after its Agent existed is never reaped from a live one. The sweep is what makes a deletion done directly in the cluster, or one the api-server crashed halfway through, end in the same state as an API delete. Kept on purpose: the usage row (soft-deleted, its runtime and config snapshots cleared so a later Agent under the same name starts clean), the public profile (retired by its own reconcile), library artifacts, the activity log, finished Experiments, and the session directory (kept so spend stays attributable). Sessions are agent-owned files on the PVC, not Postgres rows — they follow the PVC column, not this one.
```
dam-code-guardian: 🟡 **Warning:** The kept-on-purpose list names the public profile as "retired by its own reconcile", but that reconcile's candidate query inner-joins `channels` (`public-agent-profile-repository.ts:87`) and the new `channels` cleanup deletes that row first — so after this change a deleted agent's profile is unreachable to the reconcile, and its retirement rests on the non-durable `AgentDeleted` subscriber this same paragraph says carries no cleanup. Reported by issue-fit against #3545.
  **Fix:** Name the mechanism that actually retires a kept row; a reaper whose candidate query joins a table the new cleanup empties is no longer a backstop for it.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `0f06f0a`

## PR #3599: fix(api-server): remove every Agent-scoped Postgres record on delete, on any path
**Author:** jezekra1 | **Branch:** fix/agent-delete-cleanup → main | **Changes:** +898 −334 (42 files)

### Summary
One declared `AgentCleanupSource[]` list now drives both the API delete path and the 30-minute orphan sweep, so fifteen record kinds are removed on every deletion path and `AgentDeleted` is demoted to a notification. All three findings from the previous round are fixed; three architecture statements still describe the pre-PR mechanism. One skill finding repeated the PR-local dismissal for `reap-follow-up.ts` and was suppressed. _7 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: 32cbe57 (2026-09-08T10:29:53Z) — verdict COMMENT

- ✅ **Fixed:** the skills page still names the deleted cleanup saga as what reaps the publish rows, four lines above its own corrected bullet (`docs/architecture/skills.md:138`)
- ✅ **Fixed:** three edited architecture pages keep their base-branch Last verified while the five siblings edited in the same change moved to the PR date (`docs/architecture/experiments.md:3`)
- ✅ **Fixed:** the shared reap helper makes the inactivity path also flip open spans to error, a behavior change the PR body does not declare (`packages/api-server/src/modules/experiments/services/reap-running-experiment.ts:18`)
- 🆕 **New:** three sibling pages still describe the pre-PR mechanism (`docs/architecture/platform-topology.md:52`)
- 🆕 **New:** profile named as reconcile–retired, but the reconcile cannot see it (`docs/architecture/persistence.md:147`)
- 🆕 **New:** cleanupHooks stays optional now that it is the sole cleanup mechanism (`packages/api-server/src/modules/agents/services/agents-service.ts:434`)

### Findings
- 🟡 **Warning:** Three sibling pages still describe the mechanism this PR moved: platform-topology lists "per-agent cleanups" among the bus's only consumers (`docs/architecture/platform-topology.md:52`), the usage-tracking flowchart still routes `agent CM created / deleted` into `persist-agents saga` although that arm is deleted at `persist-agents.ts:38` (`docs/architecture/usage-tracking.md:29`), and the Experiments "teeth" paragraph generalizes its list to "all three terminal paths" while the new span flip reaches only the two reap paths (`docs/architecture/experiments.md:179`). Reported by doc-drift.
  **Fix:** Update every sibling statement, Mermaid edge and enumeration that names a moved mechanism, not only the page that owns it — a diagram is part of its page.
- 🟡 **Warning:** The kept-on-purpose list names the public profile as "retired by its own reconcile", but that reconcile's candidate query inner-joins `channels` (`public-agent-profile-repository.ts:87`) and the new `channels` cleanup deletes that row first, so a deleted agent's profile is unreachable to the reconcile and its retirement rests on the non-durable `AgentDeleted` subscriber the same paragraph says carries no cleanup (`docs/architecture/persistence.md:147`). Reported by issue-fit against #3545.
  **Fix:** Name the mechanism that actually retires a kept row; a reaper whose candidate query joins a table the new cleanup empties is no longer a backstop for it.
- 🟢 **Suggestion:** `cleanupHooks` is still `cleanupHooks?:` read as `?? []` (`agents-service.ts:434`, `compose.ts:76`) while every other dependency this PR touched moved from optional to required. It is now the only mechanism deleting rows in fifteen kinds, so making it required — passing `[]` where a composition wants none — would turn a missing wiring into a type error. Reported by typescript-engineering; no reachable delete path currently skips it.

Both of the previous round's suggestions were settled from the code, not the PR body: the nine changed doc pages all stamp `2026-09-08`, and `endOpenSpans` now narrows its status to `Exclude<SpanStatus, "running">`.

- 🟡 **Warning:** The PR deletes every per-agent cleanup saga (`channel-cleanup.ts`, `skills-cleanup.ts`, `kb-shares/sagas/cleanup.ts`) and moves cleanup into a declared source list, but platform-topology still names "per-agent cleanups" among the domain bus's saga consumers — a claim the same PR's persistence page contradicts outright ("The deletion event carries no cleanup") — and the page keeps its base-branch `Last verified: 2026-09-07` (`docs/architecture/platform-topology.md:52`).
  **Fix:** When a PR removes the last consumer of a class of bus subscribers, update every architecture page that enumerates bus consumers in the same PR, and bump each edited page's `Last verified`.
- 🟡 **Warning:** usage-tracking's prose was updated to say the `agents` row is marked deleted "by the agent cleanup contract … not by this saga", but the page's own flowchart still routes `agent CM created / deleted` through the bus into `persist-agents saga` — the deletion arrow this PR removed — so the diagram contradicts the prose beneath it (`docs/architecture/usage-tracking.md:29`).
  **Fix:** Treat a page's Mermaid diagram as part of the page: when an edit removes an edge from the described flow, delete the corresponding arrow in the same PR.
- 🟡 **Warning:** `reap-running-experiment.ts` adds a new terminal-transition effect — open spans flipped to `error` and stamped with an end time — on both reap paths, yet the Experiments page's "Every terminal transition has teeth" paragraph lists only trace closure, failed Invocations and rejected spawns, and generalizes those teeth to "all three terminal paths" although Stop and the script's own `finish` leave open spans untouched (`docs/architecture/experiments.md:179`).
  **Fix:** When a PR changes what a terminal transition does to a subsystem's own records, state the new effect on that subsystem's page and correct any sentence claiming the effect is uniform across paths where the code makes it path-specific.

### Issue Fit
- 🟡 **Warning:** Issue #3545 names the public agent profile among the records a non-API deletion leaves behind, and this PR declares it out of scope ("already reconciled hourly against the cluster"), but that reconcile lists only profiles whose agent still has a `channels` row — which the new `channels` cleanup deletes first (`packages/api-server/src/modules/agents/infrastructure/public-agent-profile-repository.ts:87`).
  **Fix:** Every record kind the linked issue names must stay reachable by the declared cleanup list or by a backstop that still works after that list runs.

- 🟢 **Suggestion:** `cleanupHooks` stays optional now that it is the sole cleanup mechanism — carried in `### Findings` (`packages/api-server/src/modules/agents/services/agents-service.ts:434`).

Its remaining findings did not clear the blocking bar and are regraded: the second `AgentDeleted` publisher and the best-effort owner lookups are the design `persistence.md:147` declares (`ownerSub` is already optional on `main`, and its one consumer already handles absence); `cleanupHooks?:`, the `bootstrap.ts:69` infrastructure import and the `as unknown as` test casts are unchanged from `main`; and the event-bus-direction objections cite the skill's own vendored architecture pages against a hook array `main` already had.

### Verdict
COMMENT — the cleanup mechanism itself verifies sound and every prior finding is fixed, but three architecture statements still describe the pre-PR mechanism and one makes a false claim about what retires the public profile.


---

jezekra1:
Round 5 response, in the sixth commit:

- **Public profile.** Good catch: the profile reconcile's candidate query joins `channels`, and the channels cleanup now runs first, so a deleted agent's profile was unreachable to it. The profile is now the sixteenth cleanup source (list live profiles, retire), and the profile saga no longer handles deletion at all. The deletion event's consumers are down to the UI hint and the Slack worker stop. PR body updated; the "already reconciled hourly" bullet is gone.
- **cleanupHooks** is required on the agents service; the system-agents composition passes `[]` explicitly.
- **Docs.** platform-topology no longer lists per-agent cleanups among bus consumers; the usage-tracking flowchart edge reads "agent CM created"; the experiments teeth paragraph now names four terminal paths and says which two also end open spans; public-agent-page says deletion belongs to the agent cleanup, not the saga. Dates bumped on each.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `6e39035`

🛡️ **Code Guardian** — ✅ Code Review @ `6e39035`

### Summary
Round 4's four warnings are all closed at their cause: `cleanup` and `cleanupHooks` are required, the experiments follow-up is extracted with required ports, `endOpenSpans` runs on the reap path, and the sweep resolves the owner it emits. No blocking finding survives verification. _12 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: `0f06f0a` (2026-09-08T12:26:43Z) — verdict COMMENT

- ✅ **Fixed:** three sibling pages still describe the pre-PR mechanism (`docs/architecture/platform-topology.md:52`)
- ✅ **Fixed:** profile named as reconcile–retired, but the reconcile cannot see it (`docs/architecture/persistence.md:147`)
- ✅ **Fixed:** cleanupHooks stays optional now that it is the sole cleanup mechanism (`packages/api-server/src/modules/agents/services/agents-service.ts:434`)
- 🆕 **New:** four doc statements the code does not support (`docs/architecture/security-and-credentials.md:164`)
- 🆕 **New:** the required cleanupHooks satisfied with an empty array at one call site (`packages/api-server/src/bootstrap.ts:553`)
- 🆕 **New:** four source-list entries rebuild a repository per sweep tick (`packages/api-server/src/modules/usage/compose.ts:126`)

### Findings

- 🟢 **Suggestion:** Four statements on the edited pages are not supported by the code. A key with a null `agent_ids` verifies as `agentIds: "*"` and `hasAgentBinding` short-circuits on it, so "deleting an Agent drops it from every key" does not hold for a wildcard key (`security-and-credentials.md:164`; agent ids are also generated, never reused from a name). The cleanup writes the `agent_ids` allowlist, not the `scopes` column the Lifetime list names (`persistence.md:147`). The write path still lists `AgentDeleted` among the events "the usage subsystem only adds subscribers" for, which this PR removed (`usage-tracking.md:91`). And the `agents` column inventory names four columns while the new sentence describes clearing six (`usage-tracking.md:80`). Reported by doc-drift; folded to one item per the docs-drift rule.
- 🟢 **Suggestion:** `cleanupHooks` is required now, and `composeAgentsModule` at `bootstrap.ts:553` is handed `[]`. Verified no live defect: that instance becomes `systemAgents`, used only by the Slack and Telegram workers, which call `get()` and `ensureReady()` and never `delete()`. Worth a named no-op dependency so the exemption reads as deliberate rather than as a missing wiring.
- 🟢 **Suggestion:** `listUsageAgentIds`, `listKbShareAgentIds`, `listApiKeyAgentIds` and `listOpenExperimentDriverIds` each construct a repository inside the call, so a sweep tick rebuilds four of them, while the sibling entries in the same list (`allChannelAgentIds`, `agentEnvRepo.listAgentIds`) are wired once. One convention per port would read better.

### Documentation Check
🔁 2 finding(s) from the previous review still present — now fixed (see the block above). Of eight new items, four fold into the single docs finding above; the rest are 🟢 dropped under the budget: the event-source diagram and bounded-context entry still call the table a mirror of ConfigMaps where the code reads custom resources, platform-topology names two live-hint sources where the sweep is now a third, and the experiments page states the pin release unconditionally where the driver-deletion path is wired with a no-op pin port (harmless — the annotation dies with the custom resource).

### Issue Fit
The change matches #3597: one declared `AgentCleanupSource[]` of 16 kinds, shared by the API delete hooks and the orphan sweep. The rewritten API-key paragraph is not undeclared scope — the page is 22 characters under the 40 000 cap that `scripts/doc-size.mjs` enforces on the base branch, so the added clause forced the surrounding prose to shrink. One 🟢 dropped under the budget: the new kept-on-purpose list omits `agent_case_studies`, which `case-studies.md:58` says survives deletion deliberately.

The four warnings I raised in round 4 are closed as a class: `AgentCleanupSource.cleanup`, `AgentsService.cleanupHooks`, and the experiments sweep's `onReaped` / `now` / `pin` / port dependencies are all required, and the index-coupled `agentCleanupHooks[0]!` is gone. Traced all 16 sources end to end — each `listAgentIds` pairs with the `cleanup` for the same kind, and both real delete paths (`harnessAgentsServiceFor`, the tRPC per-user context) receive the full hook list. The `schedulesBoot` hoist to line 800 is safe: `db`, `bullConnection`, `runtimeDelivery` and `agentsRepo` are all declared earlier, and `restoreAll()` still runs after `periodicJobs.start()`. Two reported warnings do not stand: no rule in this repo restricts a module `index.ts` to event types (the base `agents`, `experiments`, `api-keys` and `kb-shares` indexes already export composers and repositories), and the `/infrastructure/` imports in `bootstrap.ts` are pre-existing — about 20 of them on the base branch, including the two this PR widens. The `reap-follow-up.ts` repository construction is settled for this PR by the author's dismissal on 2026-09-08 and is not re-raised. One 🟢 kept above; five more dropped under the budget.

### Verdict
APPROVE — every blocking finding from the previous round is fixed and verified, and nothing blocking survives at this HEAD.


---

PR #3600, opened 2026-09-08
fix(skills): hash controlled management of seeded skills lifecycle

### LINKED ISSUE #3460: Skills dropped from the agent image stay behind and are relabeled as user-created
## Context

An agent created on 31 July still lists two web-search helper skills under "Created in this agent". Nobody created them. They arrived with the agent image, and a change on 3 August removed them from the image. The agent has shown them as user-authored ever since. The helper scripts those skills call are gone, so both skills fail if invoked.

## Problem

An agent copies the image's skills to its persistent home once, on first start. Nothing removes that copy later. When a skill is dropped from the image, every existing agent keeps the files.

Provenance is then decided by comparing the on-disk skill to the current image. The skill is no longer in the image, so the comparison finds no match and the agent reports it as user-authored. The owner sees a skill they did not write, credited to them, with no way to tell where it came from. The UI offers to publish it, because publishing is only blocked for skills marked as coming from the image.

This affects every agent that predates any skill removal, not one agent and not one skill.

## Expected Behavior

A skill removed from the image disappears from the agent, or is clearly marked as a leftover from an older image. It is never presented as something the owner created, and it is never publishable.

## Actual Behavior

The skill stays on disk forever, moves from "Included with agent image" to "Created in this agent", and becomes publishable and editable. It also stops working, because the files it depends on were removed with it.

## Steps to Reproduce

1. Start an agent on an image that ships a given skill.
2. Confirm the skill lists under "Included with agent image".
3. Release a new image with that skill removed.
4. Restart the agent and open its Skills page.
5. The skill now lists under "Created in this agent", with Publish offered.

jjeliga:
closes #3460 

See the ADR for the context, design decisions and alternatives considered


dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `f8f93f9`

## PR #3600: fix(skills): hash controlled management of seeded skills lifecycle
**Author:** jjeliga | **Branch:** fix/3460-manage-deprecated-base-image-skills → main | **Changes:** +43 −0 (2 files)

### Summary
The PR records ADR-088: a shipped-skills hash manifest makes each image skill platform-managed while its local copy matches any shipped version, and hands a diverged copy to the user. It adds the ADR and its generated index row, with no code.

_2 suggestion(s) dropped under the 🟢 budget._

### Findings
- 🟡 **Warning:** Consequences claim wider coverage than the Decision rule (`docs/adrs/088-image-skill-lifecycle.md:40`)
- ✅ **Looks good:** The frontmatter is complete, `subsystem: agent-skills` names a real architecture page, the index row is in generated form after 087, and the Alternatives match what `docs/architecture/agent-skills.md` records about the reverted marker file.

✅ No findings.

### Issue Fit
- 🟢 **Suggestion:** The ADR commits to a manifest build step, reconciliation on snapshot apply, a per-skill seed ledger and a hash backfill, but the PR carries no implementation and names no follow-up issue; the branch name is the only link to #3460, so the issue stays open on merge (`docs/adrs/088-image-skill-lifecycle.md:29`).

### Verdict
COMMENT — the decision is well formed and consistent with the architecture pages; one Consequences claim reads wider than the rule it summarizes.


---

docs/adrs/088-image-skill-lifecycle.md
```diff
@@ -0,0 +1,42 @@
...
+- **Re-run the whole workspace seed every boot** - propagates additions, but resurrects every file the user deliberately deleted, anywhere in $HOME.
+- **Track removals as a list of retired names and delete on name match** - also deletes copies the user edited, and keeps deleting any new skill that reuses a retired name.
+- **Keep dropped skills' copies on disk, marked "left over from an older image"** - needs a new origin value in the shared contract, version-skew handling, a UI group, and a publish-gate extension; and it labels the user's own edits as platform residue.
+- **Store each skill's origin in Postgres when first seen** - a stored verdict goes stale, dies with the agent, and for existing agents would capture today's already-wrong judgments.
+
+## Consequences
+
+- **Easier:** all three drifts end. A dropped skill disappears from existing agents - today its copy survives forever, relabeled user-created and publishable. A new skill reaches existing agents - today it provably never does. An updated skill follows the image instead of freezing as system-modified.
```
dam-code-guardian: 🟡 **Warning:** Consequences claims "all three drifts end" and "A dropped skill disappears from existing agents", but the Decision rule covers only byte-matched copies. A user who edits a skill that a later image drops keeps that copy: line 27 says it "reads user-created", and `publish-service.ts:59` gates only on `system` / `system-modified`, so the copy is publishable. That is the state issue #3460 says must never happen, and the ADR does not name it as residual — a reader of Consequences alone reads the miscrediting case as closed.

  **Fix:** Every case the Decision rule leaves unresolved is named in Consequences as an accepted trade-off, so the section never reads wider than the rule.

_(reported by `issue-fit`)_

docs/adrs/088-image-skill-lifecycle.md
```diff
@@ -0,0 +1,42 @@
...
+
+- The image carries a shipped-skills manifest: per skill name, the content hash of every version ever shipped - append-only, a few bytes per version, generated and verified at image build.
+- Reconciliation runs when the pod applies a runtime-channel snapshot (the first moment it knows the installed set): a manifest-matching local skill is overwritten when the image ships a newer version, deleted when the name is no longer shipped. Installed Skill Refs are exempt - their source governs them.
+- Additions seed once per volume through a per-skill ledger: never overwrites, a name already present counts as seeded. Safe on the agent-writable volume, since it only suppresses copies of image content. The first-boot workspace seed stays for everything that isn't a skill.
+- Origin classification and the publish gate are unchanged - no contract or UI changes. An edited copy of a shipped skill still reads system-modified and can't publish; an edited copy of a dropped skill reads user-created.
+- Hashes of skills dropped before this decision are backfilled, so existing agents heal on their next snapshot.
+
```
dam-code-guardian, line 29: 🟢 **Suggestion:** The ADR commits to a manifest build step, reconciliation on snapshot apply, a per-skill seed ledger and a hash backfill, but the PR carries no implementation and names no follow-up issue. The branch name is the only link to #3460, and with an empty PR body there is no closing keyword, so the issue stays open on merge.

_(reported by `issue-fit`)_

PetrBulanek (approved):

packages/agent-runtime/scripts/skill-manifest.ts
```diff
@@ -0,0 +1,238 @@
...
+}
+
+function appendOnlyBaseRev(): string | undefined {
+  for (const ref of ["origin/main", "main"]) {
+    try {
+      return git(["merge-base", ref, "HEAD"]).trim();
+    } catch {}
+  }
+  try {
+    git(["rev-parse", "--verify", "HEAD"]);
```
dam-code-guardian: 🟡 **Warning:** `appendOnlyBaseRev()` swallows every `merge-base` failure and falls back to `"HEAD"`, and `check()` then wraps the baseline read in `try { … } catch { base = undefined }` (`:143`), so on a checkout where neither `origin/main` nor `main` resolves the baseline becomes HEAD's own manifest — or nothing — and the append-only loop compares the file against itself. A commit that deletes a hash then passes `common:check:skill-manifest` silently, which is exactly the compatibility contract ADR-088 lists under **Committed-to** ("append-only manifest with permanent hashes"). The repo's own precedent for the same fallback, `scripts/adr-immutable.mjs:47`, fails the check instead of proceeding.

**Fix:** Fail the check when no baseline manifest is readable, as `scripts/adr-immutable.mjs` does — a guard that cannot read its baseline reports that, never `ok`.

packages/agent-runtime/scripts/skill-manifest.ts
```diff
@@ -0,0 +1,238 @@
...
+  }
+  const baseRev = appendOnlyBaseRev();
+  let base: ShippedSkillManifest | undefined;
+  try {
+    base = baseRev
+      ? parseShippedSkillManifest(
+          JSON.parse(git(["show", `${baseRev}:${MANIFEST_REL}`])),
+        )
+      : undefined;
+  } catch {
```
dam-code-guardian: 🟡 **Warning:** Second location of the same finding as `skill-manifest.ts:35` — this `catch { base = undefined }` turns an unreadable baseline manifest into "no baseline", so the append-only loop below is skipped and a removed hash passes the check.

**Fix:** Fail the check when no baseline manifest is readable, as `scripts/adr-immutable.mjs` does — a guard that cannot read its baseline reports that, never `ok`.

packages/agent-runtime/src/modules/runtime-channel/service.ts
```diff
@@ -78,6 +81,7 @@ export function createRuntimeChannelService(
     );
 
     const harnessConfigCurrent = await deps.readHarnessConfig();
+    deps.onStateApplied?.(input.state.contributions);
```
dam-code-guardian: 🟡 **Warning:** `onStateApplied` fires here, above the `failures.length > 0` branch, so a snapshot whose `skill-ref` installs failed still triggers the reconcile with those names in the exempt set. The reconciler then treats a skill it never installed as an exempt Installed Skill Ref and skips the byte-matched image copy of that name, so the drift the PR fixes survives on exactly the pods where the install failed.

**Fix:** A notification named for a completed state change is emitted only on the code path where that change actually completed, never before the failure branch that decides whether it did.

packages/agent-runtime/src/modules/skills/services/reconcile-image-skills.ts
```diff
@@ -0,0 +1,112 @@
...
+): ImageSkillReconciler {
+  let running = false;
+  let queued: ReadonlySet<string> | undefined;
+  let activeExempt = new Set<string>();
+  const isExempt = (name: string) => activeExempt.has(name);
+
+  async function run(exemptNames: ReadonlySet<string>): Promise<void> {
+    if (!deps.manifest) return;
+    if (running) {
+      queued = exemptNames;
```
dam-code-guardian, line 37: 🟡 **Warning:** `activeExempt` is service-level state that a queued call mutates (`:32`) while the previous pass still reads it through the `isExempt` closure, so a snapshot arriving mid-pass retroactively exempts names in the running `managePass` — a retired copy that should be removed is skipped for that pass. The tail `void run(next)` in `finally` (`:46`) also re-enters recursively rather than draining a loop, so a steady snapshot stream grows the stack.

**Fix:** Pass per-invocation inputs as arguments down the call chain; a coalescing service keeps only the *pending* request in shared state, never the running one's inputs, and drains the queue in a loop.

packages/agent-runtime/src/modules/skills/index.ts
```diff
@@ -1,2 +1,4 @@
 export { composeSkills } from "./compose.js";
+export { createSeedLedger } from "./infrastructure/seed-ledger.js";
```
dam-code-guardian: 🟡 **Warning:** The module's public API now re-exports two infrastructure factories, `createSeedLedger` and `loadShippedSkillManifest`; on `origin/main` this file exported `composeSkills` alone. A caller can now build a ledger against any backend and hand it in, so the module no longer owns where its own state lives.

**Fix:** A module exposes its composition root and its event types only — build the ledger and the manifest loader inside `skills/compose.ts` from the raw `DocumentStoreBackend` and the config path.

packages/agent-runtime/src/server.ts
```diff
@@ -35,7 +36,13 @@ import { createFilesService } from "./modules/files.js";
 import { composeKbPublish } from "./modules/kb-publish/compose.js";
 import { createHarnessClient } from "./modules/runtime-channel/harness-client.js";
 import { createImportHandlers, sweepStaging } from "./modules/import/index.js";
-import { composeSkills } from "./modules/skills/index.js";
+import {
+  composeSkills,
+  createSeedLedger,
+  loadShippedSkillManifest,
+} from "./modules/skills/index.js";
+import { readSkillInstallBootState } from "./modules/runtime-channel/infrastructure/skill-install-state-store.js";
```
dam-code-guardian: 🟡 **Warning:** Two new imports reach past `runtime-channel/index.ts` into that module's internals — `infrastructure/skill-install-state-store.js` for `readSkillInstallBootState` and `compose.js` for `pluginStateRoot`. That `index.ts` exists and already re-exports drivers and state stores, and the baseline carried one such deep import, so this triples it and pins the entry point to the module's internal file layout.

**Fix:** Cross-module imports resolve to the target module's `index.ts`; export `pluginStateRoot` and `readSkillInstallBootState` there instead of importing the files.

packages/agent-runtime/src/modules/skills/domain/shipped-manifest.ts
```diff
@@ -0,0 +1,15 @@
+import { z } from "zod";
+
+export const shippedSkillManifestSchema = z.object({
+  manifestVersion: z.literal(1),
+  skills: z.record(z.string(), z.array(z.string())),
+});
+
+export type ShippedSkillManifest = z.infer<typeof shippedSkillManifestSchema>;
+
+export function parseShippedSkillManifest(
```
dam-code-guardian, line 16: 🟡 **Warning:** `parseShippedSkillManifest` returns `undefined` for every validation failure, so a corrupt manifest and an absent one are the same value to callers. `shipped-manifest-loader.ts:16` re-derives the distinction from control flow, and `scripts/skill-manifest.ts:85` has to invent its own error text; a manifest that fails validation on a pod silently degrades to "image skills stay unmanaged" with no way for a caller to tell the two apart. The package already ships `core/result.ts`.

**Fix:** A domain parse returns `Result<T, E>` naming the failure; never signal a domain error by returning `undefined`.

packages/agent-runtime/src/modules/skills/services/reconcile-image-skills.ts
```diff
@@ -0,0 +1,112 @@
...
+      for (const name of exemptNames) activeExempt.add(name);
+      return;
+    }
+    running = true;
+    activeExempt = new Set(exemptNames);
+    try {
+      await seedPass(deps, isExempt);
+      await managePass(deps, deps.manifest, isExempt);
+    } catch (err) {
```
dam-code-guardian, line 51: 🟡 **Warning:** The top-level `catch` converts every failure into a log line and resolves `void`, and the per-item catches at `:68` and `:108` do the same, so both callers (`server.ts:262`, `service.ts:84`, each `void run(...)`) see a total failure and a clean pass as the same outcome. Nothing surfaces a reconcile that never ran, so a broken pod looks reconciled. The `(err as Error).message` cast also throws on a non-Error rejection.

**Fix:** A service returns a `Result` or a summary describing what failed; logging is an addition to that return value, never a replacement for it.

packages/agent-runtime/src/server.ts
```diff
@@ -107,17 +114,39 @@ const kbPublish = composeKbPublish({
 });
 const readSidePaths = skillRefPaths(runtimeManifest, homeDir);
 const readSideSet = new Set(readSidePaths);
```
dam-code-guardian, line 112: 🟡 **Warning:** Reconcile orchestration now lives in the process entry point: deciding `reconcileEnabled` from `seedRoots.length`, loading the manifest, projecting `skill-ref` contributions into an exempt-name set (`:213`), and running the boot reconcile off the skill-install boot state (`:254`). None of it is reachable from a test of the skills module, and the enable rule is invisible to the module that owns the behavior.

**Fix:** `server.ts` wires modules and passes raw dependencies only; every decision about when a module's service runs and with what belongs in that module's `compose.ts` or in a service.

packages/agent-runtime/src/modules/skills/domain/shipped-manifest.ts
```diff
@@ -0,0 +1,15 @@
+import { z } from "zod";
```
dam-code-guardian, line 1: 🟡 **Warning:** This file imports `zod` while `modules/config.ts:1` in the same package imports `zod/v4`, so two Zod entrypoints coexist and the behavior this schema depends on — `z.record` key optionality and `.catch` — is decided by which entrypoint resolves. A future major bump moves the two apart silently.

**Fix:** One Zod import specifier per package, asserted at the lint level.

docs/architecture/agent-skills.md
```diff
@@ -35,8 +35,18 @@ This is a deliberate convention, not a growing list: an image-shipped skill anyw
...
 
+A skill the image once shipped and no longer does also has no pristine counterpart. An untouched copy of it is removed by [image-skill reconciliation](#image-skill-lifecycle) on its next pass; an edited copy reads `user` **by decision**: once the image stops shipping a skill, what remains of an edited copy is the user's derivative work, publishable like anything else the user wrote.
```
dam-code-guardian, line 38: 🟡 **Warning:** Issue #3460's Expected Behavior is absolute — a skill removed from the image "is never presented as something the owner created, and it is never publishable" — but the PR delivers that for byte-identical copies only: an edited copy of a dropped skill is deliberately classified `user` and stays publishable, stated as a decision here and in ADR-088's **Committed-to** line. `closes #3460` will close the issue with that half of its Expected Behavior undone, and the issue carries no comment re-scoping it.

**Fix:** When an implementation deliberately narrows an absolute in the linked issue's Expected Behavior, record the narrowing on the issue before merging, or drop the closing keyword so the residual case stays tracked — an ADR the reporter never saw is not agreement to a reduced ask.

packages/agent-runtime/scripts/skill-manifest.ts
```diff
@@ -0,0 +1,238 @@
...
+    }
+  }
+  if (problems.length > 0) {
+    for (const p of problems) console.error(p);
+    process.exit(1);
+  }
+  console.log(`${MANIFEST_REL}: ok`);
+}
+
+function historicalSkillDirs(): string[] {
```
dam-code-guardian, line 172: 🟢 **Suggestion:** ADR-088 commits that "hashes of skills dropped before this decision are backfilled, so existing agents heal on their next snapshot", but `historicalSkillDirs()` walks only parents derived from paths that exist at HEAD (`skillSourceParents()`, `:48`), so a version shipped from a since-renamed or since-deleted parent is never hashed and a pristine copy of it on an old volume still reads user-created. Enumerating the historical parents from `git log --name-only --diff-filter=A` over the whole tree, or listing the retired parents explicitly, would close the gap the ADR's heal claim assumes.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `bd18dd7`

## PR #3600: fix(skills): hash controlled management of seeded skills lifecycle
**Author:** jjeliga | **Branch:** fix/3460-manage-deprecated-base-image-skills → main | **Changes:** +921 −39 (24 files)

### Summary
The round adds the whole implementation behind ADR-088: a shipped-skill manifest baked into the image, a per-volume seed ledger, and a reconciler that runs at boot and on every snapshot apply. The previous HEAD is not an ancestor of this one, so the whole PR is reviewed again. _8 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: f8f93f9 (2026-09-08T08:08:52Z) — verdict COMMENT — unreachable, reviewed the whole PR

- ✅ **Fixed:** Consequences claim more coverage than the rule (`docs/adrs/088-image-skill-lifecycle.md:40`)
- ✅ **Fixed:** committed work has no follow–up issue (`docs/adrs/088-image-skill-lifecycle.md:29`)
- 🆕 **New:** append-only guard passes when no baseline resolves (`packages/agent-runtime/scripts/skill-manifest.ts:35`)
- 🆕 **New:** onStateApplied fires before the failure branch (`packages/agent-runtime/src/modules/runtime-channel/service.ts:84`)
- 🆕 **New:** a queued call mutates the running pass's exempt set (`packages/agent-runtime/src/modules/skills/services/reconcile-image-skills.ts:31`)
- 🆕 **New:** the module exports two infrastructure factories (`packages/agent-runtime/src/modules/skills/index.ts:2`)
- 🆕 **New:** two deep imports past `runtime-channel/index.ts` (`packages/agent-runtime/src/server.ts:44`)
- 🆕 **New:** the parse collapses invalid and absent to `undefined` (`packages/agent-runtime/src/modules/skills/domain/shipped-manifest.ts:10`)
- 🆕 **New:** every reconcile failure becomes a log line only (`packages/agent-runtime/src/modules/skills/services/reconcile-image-skills.ts:40`)
- 🆕 **New:** reconcile orchestration lives in the entry point (`packages/agent-runtime/src/server.ts:116`)
- 🆕 **New:** two zod entrypoints in one package (`packages/agent-runtime/src/modules/skills/domain/shipped-manifest.ts:1`)
- 🆕 **New:** `skills.md` still says seeded at first boot only (`docs/architecture/skills.md:88`)
- 🆕 **New:** `closes #3460` while the PR narrows its absolute expectation (`docs/architecture/agent-skills.md:38`)
- 🆕 **New:** the backfill misses versions under a renamed parent (`packages/agent-runtime/scripts/skill-manifest.ts:164`)

### Findings

The findings below are carried as inline comments, one per anchor:

- 🟡 the append-only guard passes when no baseline manifest is readable (`packages/agent-runtime/scripts/skill-manifest.ts:35`, `:143`)
- 🟡 `onStateApplied` fires before the failure branch (`packages/agent-runtime/src/modules/runtime-channel/service.ts:84`)
- 🟡 a queued call mutates the running pass's exempt set (`packages/agent-runtime/src/modules/skills/services/reconcile-image-skills.ts:31`)
- 🟡 the skills module exports two infrastructure factories (`packages/agent-runtime/src/modules/skills/index.ts:2`)
- 🟡 two deep imports reach past `runtime-channel/index.ts` (`packages/agent-runtime/src/server.ts:44`)
- 🟡 the domain parse collapses invalid and absent to `undefined` (`packages/agent-runtime/src/modules/skills/domain/shipped-manifest.ts:10`)
- 🟡 every reconcile failure becomes a log line only (`packages/agent-runtime/src/modules/skills/services/reconcile-image-skills.ts:40`)
- 🟡 reconcile orchestration lives in the process entry point (`packages/agent-runtime/src/server.ts:116`)
- 🟡 two zod entrypoints coexist in one package (`packages/agent-runtime/src/modules/skills/domain/shipped-manifest.ts:1`)
- 🟡 `closes #3460` while the PR narrows the issue's absolute expectation (`docs/architecture/agent-skills.md:38`)
- 🟢 the backfill hashes only parents that exist at HEAD, so a version shipped from a since-renamed parent is never hashed and a pristine copy of it still reads user-created (`packages/agent-runtime/scripts/skill-manifest.ts:164`)

Not anchorable in a hunk:

- 🟡 **Warning:** `docs/architecture/skills.md:88` still states that a Standalone skill is "seeded from the image at first boot", but this PR seeds per skill on every reconciliation pass, so a skill added to a newer image now reaches volumes that booted long ago; the page keeps `Last verified: 2026-09-07` while the two pages this PR did update moved to 2026-09-08 (`docs/architecture/skills.md:88`).
  **Fix:** When a change moves a behavior that a second architecture page also asserts, update every page that states it in the same PR and bump its `Last verified`; the non-owning page keeps a one-liner plus a cross-link.

- 🟡 **Warning:** The skills catalog page still describes image-shipped Local Skills as "seeded from the image at first boot", but this PR makes image skills seed per skill on every reconciliation pass (boot from the install driver's persisted set, and each runtime-channel snapshot apply), so a skill added to a newer image now reaches volumes that booted long ago. `agent-skills.md` and `persistence.md` were updated in this PR; `skills.md` — the other page that states how a Standalone skill got onto the pod — was not, and its `Last verified` stays 2026-09-07 (`docs/architecture/skills.md:88`).
  **Fix:** When a change moves a behavior a second architecture page also asserts, update every page that states it in the same PR, not only the page that owns the concept in depth; the non-owning page keeps a one-liner plus cross-link and gets its `Last verified` bumped.

_3 suggestion(s) from this section dropped under the 🟢 budget._

### Issue Fit

Linked issue: #3460 "Skills dropped from the agent image stay behind and are relabeled as user-created" (closing keyword `closes #3460`; the issue has no comments, so the body is the whole ask). The PR delivers the reported defect: `webfetch` and `websearch`, the two skills named in the issue, each carry their historical content hash in the new manifest, so an untouched leftover copy is now reaped. No undeclared scope creep.

- 🟡 **Warning:** Issue #3460's Expected Behavior is absolute — a skill removed from the image "is never presented as something the owner created, and it is never publishable" — but the PR delivers that only for byte-identical copies: an edited copy of a dropped skill is deliberately classified `user` and made publishable, stated as a decision in the new docs and in the ADR's Committed-to line. `closes #3460` will close the issue with that half of its Expected Behavior undone and no comment in the issue re-scoping it (`docs/architecture/agent-skills.md:38`).
  **Fix:** When an implementation deliberately narrows an absolute in the linked issue's Expected Behavior, get the narrowing recorded on the issue before merging, or drop the closing keyword so the residual case stays tracked — an ADR the issue's reporter never saw is not agreement to a reduced ask.

- 🟡 **Warning:** `onStateApplied` fires before the `failures.length > 0` branch, so the reconcile is triggered on a contribution set that may have failed to apply — the reconciler then treats not-yet-installed skill-refs as authoritative exemptions and reconciles against a state the pod never reached (`packages/agent-runtime/src/modules/runtime-channel/service.ts:84`).
  **Fix:** A notification named for a completed state change is emitted only on the code path where that change actually completed, never before the failure branch that decides whether it did.
- 🟡 **Warning:** `activeExempt` is service-level mutable state that a *queued* call mutates while the previous pass is still reading it through the `isExempt` closure, so a later snapshot's exemptions retroactively change the in-flight pass; the tail `void run(next)` in `finally` also re-enters recursively with no bound (`packages/agent-runtime/src/modules/skills/services/reconcile-image-skills.ts:31`).
  **Fix:** Pass per-invocation inputs as arguments down the call chain; a coalescing service keeps only the *pending* request in shared state, never the running one's inputs, and drains the queue in a loop rather than by recursion.
- 🟡 **Warning:** The skills module's public API now re-exports two infrastructure factories (`createSeedLedger`, `loadShippedSkillManifest`); the base branch exported only `composeSkills`, so a module's private infrastructure has been promoted to its boundary (`packages/agent-runtime/src/modules/skills/index.ts:2`).
  **Fix:** A module exposes its composition root and its event types only — infrastructure adapters are constructed inside that module's `compose.ts` from raw dependencies, never handed out through `index.ts`.
- 🟡 **Warning:** Two new deep imports reach past `runtime-channel/index.ts` into the module's internals — `infrastructure/skill-install-state-store.js` for `readSkillInstallBootState` and `compose.js` for `pluginStateRoot`; that `index.ts` exists and already re-exports drivers and state stores, and the baseline had only one such deep import (`packages/agent-runtime/src/server.ts:44`).
  **Fix:** Cross-module imports resolve to the target module's `index.ts`; if a symbol is needed outside the module, export it there instead of importing the file.
- 🟡 **Warning:** `parseShippedSkillManifest` collapses every validation failure to `undefined`, so "no manifest" and "invalid manifest" are indistinguishable to callers — the loader re-derives the distinction from control flow and the manifest script invents its own error text; the package already has `core/result.ts` (`packages/agent-runtime/src/modules/skills/domain/shipped-manifest.ts:10`).
  **Fix:** A domain parse returns `Result<T, E>` naming the failure; never signal a domain error by returning `undefined`.
- 🟡 **Warning:** The reconcile service's top-level `catch` converts every failure into a log line and resolves `void`, and the per-item catches inside both loops do the same, so neither caller can distinguish a full reconcile from a total failure; the unchecked `(err as Error).message` cast compounds it (`packages/agent-runtime/src/modules/skills/services/reconcile-image-skills.ts:40`).
  **Fix:** A service returns a `Result` or a summary describing what failed; logging is an addition to that return value, never a replacement for it.
- 🟡 **Warning:** Reconcile orchestration now lives in the process entry point — deciding `reconcileEnabled` from `seedRoots.length`, loading the manifest, projecting `skill-ref` contributions into an exempt-name set, and running the boot reconcile off the skill-install boot state (`packages/agent-runtime/src/server.ts:116`).
  **Fix:** `server.ts` wires modules and passes raw dependencies only; every decision about when a module's service runs and with what belongs in that module's `compose.ts` or in a service.
- 🟡 **Warning:** The new domain schema imports `zod` while the same package's config imports `zod/v4`, so two Zod entrypoints coexist in one package and schema behavior (record optionality, `.catch`) becomes version-dependent (`packages/agent-runtime/src/modules/skills/domain/shipped-manifest.ts:1`).
  **Fix:** One Zod import specifier per package, asserted at the lint level.

_5 suggestion(s) from this section dropped under the 🟢 budget._

### Verdict
REQUEST_CHANGES — the append-only guard can pass with no baseline read, the reconcile is triggered on a contribution set that may have failed to apply, and a queued pass rewrites the running pass's exemptions.


---

packages/agent-runtime/src/modules/skills/services/reconcile-image-skills.ts
jjeliga: Deliberate: mid-pass the exempt set only ever grows, so the worst outcome is a skipped removal — and the queued follow-up pass, which always runs right after the current one finishes, performs that removal with the correct set. The dangerous direction is the opposite one: ignoring names installed mid-pass risks reaping a just-installed skill, and that does not self-heal because snapshot re-dispatch is hash-gated (the harness won't resend an unchanged snapshot). On the stack concern: `run` is async, so the calling frame unwinds at the first await — the tail `void run(next)` doesn't grow the stack.

packages/agent-runtime/src/modules/skills/domain/shipped-manifest.ts
jjeliga: This file follows the package majority: 19 files in agent-runtime import `zod`, and the single `zod/v4` outlier is `config.ts`, pre-existing on main. Aligning that outlier (and adding the lint rule) is worth a separate change; it isn't this PR's.

packages/agent-runtime/scripts/skill-manifest.ts
jjeliga: Known gap, accepted during design: a parent that no longer exists at HEAD can't be enumerated from `skillSourceParents()`. The failure mode is benign — a pristine copy of such a skill falls back to reading user-created, which is the pre-ADR status quo, not a regression, and it can never be wrongly reaped (its hashes simply aren't in the manifest). If it ever matters, a whole-tree `git log --diff-filter=A` sweep can extend the backfill.

packages/agent-runtime/src/modules/runtime-channel/service.ts
jjeliga: The firing point is deliberate; the name oversold it, so the hook is renamed `onSnapshotProcessed`. Firing even when a skill-ref install failed is the safe direction: exempting a name that failed to install can only delay a removal until the next clean snapshot or boot, while skipping the callback (or not exempting the name) risks reaping a half-installed skill — which does not self-heal, since snapshot re-dispatch is hash-gated.

packages/agent-runtime/src/modules/skills/services/reconcile-image-skills.ts
jjeliga: The fire-and-forget shape is deliberate: both call sites are boot/snapshot paths with nobody positioned to act on a returned summary, and an earlier review round's fix specifically made the reconciler unable to take down the process — the log is the observability surface. Took the valid part: the `(err as Error).message` casts are replaced with a non-Error-safe `errorMessage()`.

docs/architecture/agent-skills.md
jjeliga: The narrowing is now recorded on the issue itself: https://github.com/dam-agents/dam/issues/3460#issuecomment-5588482817 — the `closes` keyword stays, with the diverged-copy carve-out named there as the accepted residual.

docs/adrs/088-image-skill-lifecycle.md
jjeliga: Addressed by the ADR amendment (folded into the ADR commit): Consequences now names the diverged-copy residual explicitly — an edited copy of a dropped skill stays behind, unmarked, as the user's.

packages/agent-runtime/scripts/skill-manifest.ts
jjeliga: Fixed: the HEAD fallback is gone — when neither origin/main nor main yields a merge-base the check exits 1, and a baseline manifest that exists but fails validation also fails the run (with the validation reason). The one legal absence is the manifest not existing at the base rev (a brand-new manifest, as in this PR), now detected via `git ls-tree` instead of a swallowed catch.

packages/agent-runtime/scripts/skill-manifest.ts
jjeliga: Fixed together with the primary location: the `catch { base = undefined }` is gone — baseline reads now either succeed, legitimately report the file as not yet in the tree at the base rev, or fail the run.

packages/agent-runtime/src/modules/skills/index.ts
jjeliga: Fixed: `skills/index.ts` exports `composeSkills` and its types only. The seed ledger and the manifest loader are built inside `skills/compose.ts` from the raw `DocumentStoreBackend` and the manifest file path.

packages/agent-runtime/src/server.ts
jjeliga: Fixed: `pluginStateRoot` and `readSkillInstallBootState` are re-exported from `runtime-channel/index.ts` and `server.ts` imports them there; both deep imports are gone.

packages/agent-runtime/src/server.ts
jjeliga: Mostly moved: `skills/compose.ts` now builds the ledger and the manifest loader from raw dependencies and owns the empty-seed-roots disable rule (with its log line). What stays in `server.ts` is cross-module wiring that no single module can own — deriving seed roots from the runtime-channel manifest, projecting `skill-ref` contributions into exempt names for the snapshot hook, and the boot pass reading runtime-channel's persisted install state — plus the kill-switch config read, which is entry-point territory.

packages/agent-runtime/src/modules/skills/domain/shipped-manifest.ts
jjeliga: Taken, in 8b41b3382: `parseShippedSkillManifest` returns `Result<ShippedSkillManifest, InvalidShippedSkillManifest>` with the zod issues as the reason; the pod loader logs the reason before degrading to unmanaged, and the manifest CLI fails with it. Absence never reaches the parse — the loader gates on `existsSync` and the CLI on `ls-tree`.

docs/adrs/088-image-skill-lifecycle.md
jjeliga: Overtaken by events: the full implementation (reconciler, manifest build step + CI check, seed ledger, hash backfill, docs, e2e spec) now lives in this PR, and the body carries `closes #3460`.

packages/agent-runtime/src/modules/skills/infrastructure/seed-ledger.ts
```diff
@@ -0,0 +1,35 @@
+import { z } from "zod";
+import type { DocumentStoreBackend } from "../../../core/document-store.js";
+
+const seedLedgerSchema = z.object({
+  seeded: z.array(z.string()).catch([]).default([]),
```
dam-code-guardian: 🟡 **Warning:** The seed ledger's schema uses `.catch([])`, so a corrupt or truncated ledger silently reads as “nothing has been seeded” and the next pass re-seeds image skills the user deliberately deleted — the very outcome the ledger exists to prevent. The sibling store added in this same PR takes the opposite approach and reports `corrupt` explicitly.

**Fix:** State whose absence and whose corruption imply different actions must parse strictly and surface corruption; `.catch(default)` is only for state where a lost value is genuinely equivalent to an empty one.

docs/architecture/agent-skills.md
```diff
@@ -26,7 +26,7 @@ Install writes the skill directory into **every** configured Skill Path; uninsta
 
 ### Skill Origin
 
-Every Local Skill carries a **provenance** verdict, judged by the agent-runtime at read time. The reference is the set of **pristine roots** in the image — exactly two sanctioned locations, both immutable and always in-pod, so no build-time manifest or on-PVC marker is needed (a marker was tried and reverted — third-party baked skills aren't ours to stamp, and the PVC is agent-writable anyway):
+Every Local Skill carries a **provenance** verdict, judged by the agent-runtime at read time. The reference is the set of **pristine roots** in the image — exactly two sanctioned locations, both immutable and always in-pod, so origin classification consults no build-time manifest and no on-PVC marker (a marker was tried and reverted — third-party baked skills aren't ours to stamp, and the PVC is agent-writable anyway):
 
 1. the **pristine workspace copy** — the directory the image's first-boot seed copies onto the PVC and never touches again, and
```
dam-code-guardian: 🟡 **Warning:** The Skill Origin section still describes the pristine workspace copy as the directory the image's first-boot seed “copies onto the PVC and never touches again”, but this PR's reconciler overwrites that copy when the image ships a newer version and deletes it when the name is retired (`reconcile-image-skills.ts` `managePass`), so the page contradicts its own new Image-Skill Lifecycle section. The same staleness sits in the Local Skill definition at `:15`, which still says image skills arrive “seeded from the image at first boot”, while the sibling catalog page was updated (`docs/architecture/skills.md:88`).

**Fix:** When a PR changes a lifecycle a page states elsewhere as an absolute (“never touched again”, “at first boot”), sweep every sentence on that page asserting the old absolute and re-state it against the new rule in the same PR — a new section never overrides earlier prose by implication.

docs/architecture/agent-skills.md
```diff
@@ -35,8 +35,18 @@ This is a deliberate convention, not a growing list: an image-shipped skill anyw
...
+- The **Seed Ledger** — a per-volume record of which shipped skill names have been seeded, so each is copied into the Skill Paths exactly once and a skill the user then deletes is never resurrected. Retiring clears the entry — the platform's own removal must not count as the user's — so a skill re-shipped after retirement seeds again. It sits on the agent-writable volume, which is safe because it only ever suppresses copies of image content. Only pristine-workspace skills seed; staged skills never do. The whole-workspace first-boot seed ([persistence](persistence.md)) stays for everything that isn't a skill.
+- **Reconciliation** — runs when the pod applies a runtime-channel snapshot, and once at boot from the install driver's persisted set, because an image-only upgrade delivers no snapshot. A local skill whose hash appears in the manifest is updated or removed as above; anything else — including a copy that cannot be hashed — is left alone. A skill tracked as an Installed Skill Ref is exempt: its Source governs it. `PLATFORM_IMAGE_SKILL_RECONCILE=off` disables the whole pass on a pod.
```
dam-code-guardian: 🟡 **Warning:** The Image-Skill Lifecycle section names exactly one way reconciliation can be off — `PLATFORM_IMAGE_SKILL_RECONCILE=off` — but the code has two further silent no-op paths: the composer returns no reconciler at all when no pristine seed root is distinct from the skill paths (`skills/compose.ts` `buildReconciler`), and `run()` returns immediately when the shipped-skill manifest is absent or fails validation (`shipped-manifest-loader.ts` → `reconcile-image-skills.ts` `run`), leaving image skills unmanaged on a pod the page says is managed.

**Fix:** When code adds conditions under which a documented pass degrades to a no-op, the architecture page must state the degradation and its consequence, not only the operator kill switch — a page that lists one off-switch reads as an exhaustive list.

dam-code-guardian (changes requested):
🛡️ **Code Guardian 🛡️** — ❌ Code Review @ `8b41b33`

## PR #3600: fix(skills): hash controlled management of seeded skills lifecycle
**Author:** jjeliga | **Branch:** fix/3460-manage-deprecated-base-image-skills → main | **Changes:** +964 −43 (26 files)

### Summary
The round answers every finding of the previous review: five are fixed in the code, seven are settled by the author as deliberate. What is left is one state-corruption path in the new seed ledger and the documentation that still states the old lifecycle. _(Suppressed 7 finding(s) per PR-local overrides: reconcile-image-skills.ts:35, reconcile-image-skills.ts:44, shipped-manifest.ts:1, skill-manifest.ts:63, skill-manifest.ts:172, server.ts:112, agent-skills.md:38.)_ _12 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: bd18dd7 (2026-09-08T13:05:07Z) — verdict REQUEST_CHANGES — unreachable, reviewed the whole PR

- ✅ **Fixed:** append-only guard passes when no baseline resolves (`packages/agent-runtime/scripts/skill-manifest.ts:35`)
- ✅ **Fixed:** parse collapses invalid and absent to `undefined` (`packages/agent-runtime/src/modules/skills/domain/shipped-manifest.ts:10`)
- ✅ **Fixed:** the module exports two infrastructure factories (`packages/agent-runtime/src/modules/skills/index.ts:2`)
- ✅ **Fixed:** two deep imports past `runtime-channel/index.ts` (`packages/agent-runtime/src/server.ts:44`)
- ✅ **Fixed:** `skills.md` still says seeded at first boot only (`docs/architecture/skills.md:88`)
- 🆕 **New:** a corrupt seed ledger reads as empty and re-seeds deleted skills (`packages/agent-runtime/src/modules/skills/infrastructure/seed-ledger.ts:5`)
- 🆕 **New:** the page keeps two absolutes the reconciler breaks (`docs/architecture/agent-skills.md:31`, `:15`)
- 🆕 **New:** two silent no-op paths are undocumented (`docs/architecture/agent-skills.md:48`)
- 🆕 **New:** the apply path gains a hook its own page does not state (`docs/architecture/runtime-delivery.md:278`)

Settled as deliberate by the author, and recorded as PR-local overrides: the mid-pass exempt-set growth, the fire-and-forget reconcile shape, `onSnapshotProcessed`'s firing point, the pre-existing `zod/v4` outlier, the renamed-parent backfill gap, the residual wiring in `server.ts`, and `closes #3460` with the residual on the issue.

### Findings

Carried as inline comments, one per anchor:

- 🟡 a corrupt seed ledger reads as empty, so the next pass re-seeds skills the user deleted (`packages/agent-runtime/src/modules/skills/infrastructure/seed-ledger.ts:5`)
- 🟡 `agent-skills.md` keeps "never touches again" and "at first boot" against the new reconciler (`docs/architecture/agent-skills.md:31`, `:15`)
- 🟡 the lifecycle section names one off-switch, while two further paths make the pass a silent no-op (`docs/architecture/agent-skills.md:48`)

Not anchorable in a hunk:

- 🟡 **Warning:** The apply path gains a post-apply hook — `applyState` now calls `onSnapshotProcessed(contributions)`, wired through `composeRuntimeChannel` — and image-skill reconciliation rides it, but `runtime-delivery.md` owns the apply path, still enumerates it as contributions then events then outcome, and keeps `Last verified: 2026-09-03` (`docs/architecture/runtime-delivery.md:278`).
  **Fix:** When a subsystem's own page enumerates a sequence, a PR that adds a step or a hook to that sequence edits the page in the same change and bumps its `Last verified:` stamp.

- 🟡 **Warning:** The Skill Origin section still describes the pristine workspace copy as the directory the image's first-boot seed "copies onto the PVC and never touches again", but this PR's reconciler overwrites that copy when the image ships a newer version and deletes it when the name is retired (`reconcile-image-skills.ts` `managePass`), so the page now contradicts its own new Image-Skill Lifecycle section; the same staleness sits in the Local Skill definition, which still says image skills arrive "seeded from the image at first boot" (`docs/architecture/agent-skills.md:15`) while the sibling catalog page was updated (`docs/architecture/skills.md:88`) (`docs/architecture/agent-skills.md:31`).
  **Fix:** When a PR changes a lifecycle a page states elsewhere as an absolute ("never touched again", "at first boot"), sweep every sentence on that page asserting the old absolute and re-state it against the new rule in the same PR — a new section never overrides earlier prose by implication.
- 🟡 **Warning:** The Image-Skill Lifecycle section names exactly one way reconciliation can be off — `PLATFORM_IMAGE_SKILL_RECONCILE=off` — but the code has two further silent no-op paths: the composer returns no reconciler at all when no pristine seed root is distinct from the skill paths (`skills/compose.ts` `buildReconciler`), and `run()` returns immediately when the shipped-skill manifest is absent or fails validation (`shipped-manifest-loader.ts` → `reconcile-image-skills.ts` `run`), leaving image skills unmanaged on a pod that the page says is managed (`docs/architecture/agent-skills.md:48`).
  **Fix:** When code adds conditions under which a documented pass degrades to a no-op, the architecture page must state the degradation and its consequence, not only the operator kill switch — a page that lists one off-switch reads as an exhaustive list.
- 🟡 **Warning:** The apply path gains a post-apply extension point — `applyState` now calls `onSnapshotProcessed(contributions)` after events settle (`runtime-channel/service.ts`), wired from `composeRuntimeChannel` — and image-skill reconciliation rides it, but the runtime-delivery page, which owns the apply path and enumerates its sequence step by step in "Driver reconciliation", is unedited and still describes an apply as contributions then events then outcome (`Last verified: 2026-09-03`) (`docs/architecture/runtime-delivery.md:278`).
  **Fix:** When a subsystem's own page enumerates a sequence, adding a step or a hook to that sequence requires editing that page in the same PR (a one-liner plus cross-link when another page owns the consumer in depth), and bumping its `Last verified:` stamp.
- 🟢 **Suggestion:** The page states reconciliation also runs "once at boot from the install driver's persisted set", but the boot pass is skipped entirely when that persisted set is unreadable (`packages/agent-runtime/src/server.ts`, `bootState.kind === "corrupt"` → log and no run), so on a corrupt state file an image-only upgrade delivers neither a snapshot nor a boot pass and the volume stays stale until the next snapshot (`docs/architecture/agent-skills.md:48`).

### Issue Fit

Linked issue: #3460 "Skills dropped from the agent image stay behind and are relabeled as user-created" (closing keyword: `closes #3460`). The issue's absolute wording is narrowed by a maintainer comment on the issue (jjeliga, 2026-09-08): only an *untouched* copy is platform-managed; an edited copy of a dropped skill staying behind as the user's is the accepted residual. Judged against that scope, the reported case is delivered — the two claude-code skills removed on 2026-08-03 (`webfetch`, `websearch`) hash to `6630594a…` and `4f11b755…`, both present in the backfilled manifest, so an untouched copy on an existing volume is reaped on the next snapshot.

🔁 1 finding from the previous review is settled as a PR-local override: the backfill's dependence on parents that exist at HEAD (`packages/agent-runtime/scripts/skill-manifest.ts:63`, `:172`). The `packages/humr-base/skills/humr-schedules` case this round found — image-shipped before the platform rename, no `humr-*` hash in the manifest — is that same gap, so it is not re-raised.

_2 suggestion(s) from this section dropped under the 🟢 budget._

- 🟡 **Warning:** The seed ledger's schema uses `.catch([])`, so a corrupt or truncated ledger silently reads as "nothing has been seeded" and the next pass re-seeds image skills the user deliberately deleted — the very outcome the ledger exists to prevent. The sibling store added in this same PR takes the opposite approach and reports `corrupt` explicitly (`packages/agent-runtime/src/modules/skills/infrastructure/seed-ledger.ts:5`).
  **Fix:** State whose absence and whose corruption imply different actions must parse strictly and surface corruption; `.catch(default)` is only for state where a lost value is genuinely equivalent to an empty one.

_10 suggestion(s) from this section dropped under the 🟢 budget._

### Verdict
REQUEST_CHANGES — the corrupt-ledger path re-seeds skills the user deleted, and three architecture pages still state the lifecycle this PR replaced.


- `docs/architecture/agent-skills.md:15` — 🟡 **Warning:** The Local Skill definition still says image skills arrive “seeded from the image at first boot”, which this PR's reconciler replaces: seeding now happens per skill on every pass, and a retired name is deleted. Same finding as `:31`; both sentences on this page assert the superseded absolute.
  
  **Fix:** When a PR changes a lifecycle a page states elsewhere as an absolute (“never touched again”, “at first boot”), sweep every sentence on that page asserting the old absolute and re-state it against the new rule in the same PR — a new section never overrides earlier prose by implication.


---

packages/agent-runtime/src/modules/skills/infrastructure/seed-ledger.ts
jjeliga: Fixed in 3a4b3f7f7 — and the hole was deeper than the `.catch([])`: `openJsonFile` itself falls back to `initial()` on any parse failure, so a strict schema through the store would still have read corrupt as empty. The ledger now strict-reads its file at open (same pattern as `readSkillInstallBootState`); corrupt yields no ledger, which disables seeding on that volume — nothing the user deleted can resurrect — while update and retire reconciliation keep running, and the log names the file and the remedy.

docs/architecture/agent-skills.md
jjeliga: Fixed in 22176cc66: both stale absolutes on the page are restated — the pristine workspace copy is now defined as the in-image workspace payload whose skills reconciliation seeds and keeps at the shipped version (first-boot seed covers everything that isn't a skill), and the Local Skill definition's "at first boot" now points at the lifecycle section. The same commit also updates runtime-delivery.md's apply sequence with the post-apply hook (that finding had no inline thread).

docs/architecture/agent-skills.md
jjeliga: Fixed in 22176cc66: the lifecycle section now lists every degradation, not just the kill switch — no readable manifest or no pristine workspace root distinct from the Skill Paths leaves image skills unmanaged (logged), and a corrupt Seed Ledger (new in 3a4b3f7f7) disables seeding alone while update and retire reconciliation keeps running.

packages/agent-runtime/src/modules/skills/infrastructure/seed-ledger.ts
jjeliga: Superseding my previous reply — the disable-forever semantics were too harsh, so f0139eae1 replaces them with in-place recovery: on a corrupt read, the next pass records every currently-shipped skill as already seeded without copying anything, which rewrites a valid ledger in one atomic write. Nothing the user deleted can resurrect, seeding resumes for anything shipped afterwards, and the only loss is a skill first shipped while the ledger was corrupt, which never auto-seeds on that volume.

Worth emphasizing: this is an edge case that will in all likelihood never fire. The ledger is only ever written via tmp-file-plus-rename, so our own writer cannot produce a torn or truncated file — corruption requires something else on the agent-writable volume mangling it. The recovery path exists so that even then the degradation is bounded and self-healing rather than permanent.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `8ec3cd0`

## PR #3600: fix(skills): hash controlled management of seeded skills lifecycle
**Author:** jjeliga | **Branch:** fix/3460-manage-deprecated-base-image-skills → main | **Changes:** +1014 −44 (27 files)

### Summary
Every finding of the previous round is fixed in the code: the seed ledger now strict-reads its file and heals a corrupt one in place, and the pages that stated the old lifecycle are amended. Three suggestions remain, none blocking. _13 suggestion(s) dropped under the 🟢 budget._ _(Suppressed 4 finding(s) per PR-local overrides: the `zod` import, the mid-pass exempt-set growth, the entry-point orchestration, and the cross-module read of runtime-channel's install state.)_

### Changes since last review
Previous HEAD: `8b41b33` (2026-09-08T19:31:22Z) — verdict REQUEST_CHANGES — unreachable, reviewed the whole PR

- ✅ **Fixed:** the seed ledger read `.catch([])`, so a corrupt ledger re-seeded deleted skills (`packages/agent-runtime/src/modules/skills/infrastructure/seed-ledger.ts:5`)
- ✅ **Fixed:** `agent-skills.md` kept two absolutes the reconciler breaks (`docs/architecture/agent-skills.md:31`)
- ✅ **Fixed:** the lifecycle section named only the kill switch, not the silent no-op paths (`docs/architecture/agent-skills.md:48`)
- ✅ **Fixed:** `runtime-delivery.md` did not carry the post-apply hook (`docs/architecture/runtime-delivery.md:278`)
- 🆕 **New:** one docs item: two unbumped stamps, one unnamed degradation (`docs/architecture/persistence.md:3`)
- 🆕 **New:** one state file, two disagreeing schemas (`packages/agent-runtime/src/modules/runtime-channel/infrastructure/skill-install-state-store.ts:14`)
- 🆕 **New:** the feature's only test runs in no CI job (`packages/e2e/playwright/src/tests/full/skill-lifecycle.spec.ts:25`)

I verified the corrupt-ledger recovery, the manifest append-only check over all ten shipped skill directories (every hash present), and the three amended pages.

### Findings
- 🟢 **Suggestion:** `persistence.md` and `skills.md` gain edited sentences while both keep `Last verified: 2026-09-08`; `agent-skills.md:48` now lists three degradations, and `server.ts:243` adds a fourth — a corrupt skill-install boot state skips the boot pass (`docs/architecture/persistence.md:3`, `docs/architecture/skills.md:3`, `docs/architecture/agent-skills.md:48`).
- 🟢 **Suggestion:** `skill-install-state.json` now has two schemas: the writer's `installed: z.array(z.string()).catch([])` and `strictSkillInstallStateSchema`. The plugin reads a malformed file as `[]`, `readSkillInstallBootState` reads it as `corrupt` (`packages/agent-runtime/src/modules/runtime-channel/infrastructure/skill-install-state-store.ts:14`).
- 🟢 **Suggestion:** `skills-full` runs only under `E2E_FULL=1`, which reaches Playwright through `mise run e2e --full`; `cd.yml:649` runs `mise run e2e` with no argument, and `E2E_FULL` appears in no workflow. The spec also reads its fixtures from `538ffa0057bf` and `1c6470c46e71`, which a `--depth 50` clone cannot resolve (`packages/e2e/playwright/src/tests/full/skill-lifecycle.spec.ts:25`).

🔁 3 finding(s) from the previous review are fixed at this HEAD. This round's three items are merged into the single docs entry in `### Findings`.

### Issue Fit
✅ No findings.

🔁 The round's two kept items are in `### Findings` (the two-schema state file, the e2e fixture). Four items are settled by PR-local overrides; the rest are layering and placement observations on a working diff, dropped under the 🟢 budget.

### Verdict
APPROVE — every blocking finding of the previous round is fixed in the code, and only suggestions remain.


---

PR #3222, opened 2026-08-07
fix(security): add baseline HTTP security headers

JanPokorny:
## Summary
- Fixes the corporate ZAP scan findings: missing CSP, X-Frame-Options, X-Content-Type-Options, HSTS headers, and the nginx `Server` version-leak header.
- `packages/ui/default.conf`: `server_tokens off;` + `add_header` for CSP/X-Frame-Options/X-Content-Type-Options/Referrer-Policy/HSTS on the static UI. CSP explicitly allowlists `fonts.googleapis.com`/`fonts.gstatic.com` so Google Fonts keeps loading (the ZAP "Cross-Domain Misconfiguration" finding on that request is a scanner false-positive for a font stylesheet fetch — the CSP allowlist is the real fix, not removing the fonts).
- `packages/api-server/src/apps/api-server/app.ts`: matching global header middleware for all `/api/*` responses, following the header pattern already established in the share-viewer sub-app.
- The "Cookie with SameSite Attribute None" finding is Keycloak's own session cookie — not something this repo's code sets — so it's out of scope here.

## Test plan
- [x] `mise run api-server:fix:lint` passes
- [x] `mise run api-server:check:tsc` passes
- [ ] Re-run the ZAP scan against a deployed instance to confirm findings clear
- [ ] Manually smoke-test the UI (fonts render, no CSP console errors) after deploy

PR #3602, opened 2026-09-08
fix(api-server): stop leaking internals in tRPC errors, validate K8s names

JanPokorny:
## Why

A ZAP 2.17 scan of the dev deployment (browser-based Keycloak login for the UI crawl, bearer-token seeding of all 136 tRPC procedures over HTTP, then an active scan) surfaced these issues:

- every tRPC error body carried a server stack trace with container paths, including 404s
- ~30 procedures answered **500** to a non-UUID id, with the full Drizzle `Failed query: …` SQL statement and parameters in the message
- `artifactLibrary.createFolder` with a duplicate name answered 500 with the insert statement instead of 409
- `agents.delete` / `agents.restart` with an id like `c:/Windows/system.ini` reached kube-apiserver as `agents/c:/Windows/system.ini` (RBAC returned 403, and the 403 body naming the service account and namespaces was echoed to the client)
- `connections.create`, `connections.setAgentConnections`, `schedules.createCron` turned user mistakes into 500s
- static assets were served with `Server: nginx/1.30.4` and without HSTS / `X-Content-Type-Options`: the chart mounts its own nginx `default.conf` ConfigMap over the image's file, so #3222 never reached Helm deployments

The three High SQL-injection alerts ZAP raised were verified as false positives (validation errors / expired-token 401s compared as "different responses"). The ~15% router 503s seen during the active scan are the OpenShift router's "Application is not available" page for ZAP's Host-header manipulation probes, not capacity.

## What

- **Contract package** (`trpc.ts`): the error formatter strips `stack` and redacts the message of `INTERNAL_SERVER_ERROR`; a middleware maps Postgres SQLSTATE `22P02` → `BAD_REQUEST` and `23505` → `CONFLICT`.
- **Adapters**: `onError` on the HTTP and WebSocket handlers logs internal errors with their cause (they were not logged before); HTTP tRPC responses get `Cache-Control: no-store`.
- **K8s adapter**: caller-supplied object names must match the DNS-subdomain grammar; invalid names read as "not found".
- **Services**: user-caused failures in connections, schedules, agents (`ensureReady`), e2e and OAuth start throw `TRPCError` client codes instead of plain `Error`.
- **Chart** (`templates/ui/app.yaml`): `server_tokens off` and the static-asset security headers in the nginx ConfigMap, scoped to `location /` so proxied `/api/` responses keep the api-server's own headers. CSP is deliberately not copied from the image file: its `connect-src 'self'` would block the Keycloak token endpoint on Helm deployments, where Keycloak has its own host.
- **Docs**: platform-topology notes the error contract.
- **Test**: `trpc-errors.test.ts` covers the two Postgres mappings and the redaction.

## Trade-off

Services that still throw a plain `Error` (background refresh, channel manager, runtime delivery, an unreachable branch in usage) now show "internal server error" to clients. Those remaining sites are genuinely internal.

## Deployment side

- Router sticky cookies (`SameSite=None`) are disabled on the dev UI/API routes in dam-ops#109.
- Presigned upload URLs naming the in-cluster SeaweedFS host are by design: upload tickets are dialled by agent pods (`objectStorage.agentEndpoint`); browsers never use them.


deploy/helm/platform/templates/ui/app.yaml
```diff
@@ -74,6 +75,11 @@ data:
                           application/manifest+json image/svg+xml;
 
         location / {
+            # Static-only headers; /api/ responses carry the api-server's own.
+            add_header X-Content-Type-Options "nosniff" always;
```
dam-code-guardian, line 79: 🟢 **Suggestion:** This adds a browser-visible hardening layer at the UI edge (sniffing, framing, referrer and transport policy, plus `server_tokens off`), but no architecture page records it: `docs/architecture/security-and-credentials.md` — the page `deploy/helm/platform/**` maps to — has no mention of response headers, and `platform-topology.md` still describes the UI as "served by the api-server" although it runs its own nginx Deployment that proxies `/api/`. That two-edge topology is what the comment above depends on.

packages/api-server/src/modules/agents/infrastructure/k8s.ts
```diff
@@ -54,6 +54,8 @@ function isStatus(err: unknown, code: number): boolean {
   );
 }
 const is404 = (err: unknown) => isStatus(err, 404);
+const isDnsSubdomainName = (name: string) =>
```
dam-code-guardian, line 57: 🟢 **Suggestion:** This guard gives one invalid-input class three different outcomes — `getSecret`/`getCustomObject` return `null` (which a caller cannot tell from "absent"), `deleteSecret`/`deleteCustomObject` succeed silently, and `replaceSecret`/`patchCustomObject` throw a plain `Error` that now surfaces as an opaque 500 — and it covers only the methods taking `name` as a parameter, so `createSecret` and `createCustomObject` stay unchecked. Every name reaching these paths today is generated internally, so nothing caller-supplied demonstrates the gap.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `84a3f5f`

## PR #3602: fix(api-server): stop leaking internals in tRPC errors, validate K8s names
**Author:** JanPokorny | **Branch:** fix/zap-findings → main | **Changes:** +182 −14 (14 files)

### Summary
The PR redacts internal tRPC error messages and stacks on the public port, logs them server-side with the cause, maps two Postgres SQLSTATEs to client error codes, converts five services to `TRPCError`, guards the K8s adapter's name-taking methods, and adds static response hardening at the UI edge. No blocking defect survived verification.

_18 suggestion(s) dropped under the 🟢 budget._

### Findings
- ✅ **Looks good:** The new headers sit in `location /` only, so `/api/` does not inherit them — the added comment claims the api-server supplies its own there, and it does: `securityHeaders` is mounted on `*` (`packages/api-server/src/apps/api-server/app.ts:70`), so both edges assert the same four headers.

- 🟢 **Suggestion:** The PR adds a browser-visible hardening layer at the UI edge (sniffing, framing, referrer and transport policy, plus `server_tokens off`), but no architecture page records it: `docs/architecture/security-and-credentials.md` — the page the profile maps `deploy/helm/platform/**` to — has no mention of response headers, and `platform-topology.md` still describes the UI as "served by the api-server" although it runs its own nginx Deployment that proxies `/api/`. The two-edge topology is what the new comment depends on (`deploy/helm/platform/templates/ui/app.yaml:79`)

### Issue Fit
_Every item the PR body promises is present in the diff, and no undeclared substantial change. Its remaining observations were dropped under the 🟢 budget._

- 🟢 **Suggestion:** Redaction and internal-error logging are installed on the public tRPC root only. `harness-trpc.ts` builds a second root with no `errorFormatter`, and neither harness handler passes `onError`, so harness-port errors still carry raw messages and stacks and no internal failure is logged with its cause. The port is cluster-internal, so the consequence is bounded (`packages/api-server-api/src/harness-trpc.ts:5`)
- 🟢 **Suggestion:** The name guard gives one invalid-input class three different outcomes — `getSecret`/`getCustomObject` return `null` (which a caller cannot tell from "absent"), `deleteSecret`/`deleteCustomObject` succeed silently, and `replaceSecret`/`patchCustomObject` throw a plain `Error` that now surfaces as an opaque 500 — and it covers only the methods that take `name` as a parameter, so `createSecret` and `createCustomObject` stay unchecked. Every name reaching these paths today is generated internally, so no caller-supplied value demonstrates the gap (`packages/api-server/src/modules/agents/infrastructure/k8s.ts:57`)

### Verdict
APPROVE — the error contract holds on both public transports and the tests cover both mappings plus the redaction; nothing blocking remains.


---

PR #3603, opened 2026-09-08
fix(ui): stop and pause actions explained in agent actions menu

### LINKED ISSUE #3059: UI - Sandbox power menu wording does not convey what the options actually do
## Context

The sandbox menu offers Restart, "Pause — wakes on next use", "Stop — until started again", and Delete sandbox.

Pause and Stop both exist to free reserved compute, which is what makes room under your budget ceiling to start something else. They differ only in what can bring the sandbox back: Pause wakes on any deliberate use, Stop stays down until explicitly started.

## Problem

Both options read as "off", so the choice looks like a spectrum of off-ness rather than two distinct behaviours. Neither label mentions the actual difference, or that they free up compute.

"until started again" is also inaccurate. A schedule firing will start a stopped sandbox too. The confirm dialog says so, the menu does not.

Restart adds to the muddle by sitting in the same list, since it is recovery for a stuck sandbox, not a power-down choice.

So people cannot predict what they are choosing, whether the sandbox can come back on its own, or what they gain by doing it.

## Goal

Each option should say plainly what it does, what can bring the sandbox back, and what it frees up, without needing to know how hibernation, budgets, or schedules work.

## User Stories

- As someone managing sandboxes, I want each option to say what it does so that I can pick the right one without guessing.
- As someone stopping a sandbox, I want to know a schedule can still start it so that I am not surprised later.
- As someone near my budget ceiling, I want to know which option frees up room so that I can start what I need.

## Scope

Wording in this menu only. The underlying behaviours are not in question.

## Open Questions

- Should Restart be separated, given it is recovery rather than a power state?
- Do Pause and Stop both need to be here, or is offering two similar options part of the confusion?

---

## Design decision

Current state:

Image

Desired state:

Image

kapetr:
The three-dot menus on the agent list and the agent page now group Stop and Pause under a "Free up compute" section, with a one-line description of when the agent wakes again. 

For agents set to never hibernate, Pause is disabled and the description explains why and where to change it.

The compute box tooltip on Home now ends with "Stop or pause an agent to free up compute."

Closes #3059

### DIFF
--- a/packages/ui/src/modules/agents/components/power-menu-items.tsx
+++ b/packages/ui/src/modules/agents/components/power-menu-items.tsx
@@ -1,3 +1,4 @@
+import { Badge } from "@/components/ui/badge";
 import {
   DropdownMenuItem,
   DropdownMenuLabel,
@@ -8,10 +9,12 @@ import type { AgentView } from "../../../types.js";
 
 export function FreeUpComputeItems({
   agent,
+  configureLabel,
   onPause,
   onStop,
 }: {
   agent: AgentView;
+  configureLabel: string;
   onPause: () => void;
   onStop: () => void;
 }) {
@@ -30,18 +33,20 @@ export function FreeUpComputeItems({
         />
       </DropdownMenuItem>
       <DropdownMenuItem
-        className="h-auto py-2"
+        className="h-auto py-2 data-[disabled]:pointer-events-auto"
         disabled={neverHibernates}
         onSelect={onPause}
         data-testid="agent-pause"
+        title={
+          neverHibernates
+            ? `This agent is set to never hibernate, so pausing it would wake itself. Change its idle timeout in ${configureLabel}.`
+            : undefined
+        }
       >
         <MenuItemBody
           title="Pause"
-          description={
-            neverHibernates
-              ? "This agent is set to never hibernate, so pausing it would wake itself. Change its idle timeout in Configure agent."
-              : "Wakes on any action (a schedule firing, a message arriving, or you opening a chat)"
-          }
+          tag={neverHibernates ? "Unavailable" : undefined}
+          description="Wakes on any action (a schedule firing, a message arriving, or you opening a chat)"
         />
       </DropdownMenuItem>
       <DropdownMenuSeparator />
@@ -51,14 +56,23 @@ export function FreeUpComputeItems({
 
 function MenuItemBody({
   title,
+  tag,
   description,
 }: {
   title: string;
+  tag?: string;
   description: string;
 }) {
   return (
     <span className="flex max-w-72 flex-col gap-0.5">
-      <span>{title}</span>
+      <span className="flex items-center gap-2">
+        {title}
+        {tag && (
+          <Badge variant="muted" size="sm">
+            {tag}
+          </Badge>
+        )}
+      </span>
       <span className="whitespace-normal text-xs text-muted-foreground">
         {description}
       </span>
--- a/packages/ui/src/modules/agents/components/power-menu-items.tsx
+++ b/packages/ui/src/modules/agents/components/power-menu-items.tsx
@@ -1,0 +1,67 @@
+import {
+  DropdownMenuItem,
+  DropdownMenuLabel,
+  DropdownMenuSeparator,
+} from "@/components/ui/dropdown-menu";
+
+import type { AgentView } from "../../../types.js";
+
+export function FreeUpComputeItems({
+  agent,
+  onPause,
+  onStop,
+}: {
+  agent: AgentView;
+  onPause: () => void;
+  onStop: () => void;
+}) {
+  const neverHibernates = agent.hibernationTimeoutMin === 0;
+  return (
+    <>
+      <DropdownMenuLabel>Free up compute</DropdownMenuLabel>
+      <DropdownMenuItem
+        className="h-auto py-2"
+        onSelect={onStop}
+        data-testid="agent-stop"
+      >
+        <MenuItemBody
+          title="Stop"
+          description="Wakes only when a schedule fires or you start it"
+        />
+      </DropdownMenuItem>
+      <DropdownMenuItem
+        className="h-auto py-2"
+        disabled={neverHibernates}
+        onSelect={onPause}
+        data-testid="agent-pause"
+      >
+        <MenuItemBody
+          title="Pause"
+          description={
+            neverHibernates
+              ? "This agent is set to never hibernate, so pausing it would wake itself. Change its idle timeout in Configure agent."
+              : "Wakes on any action (a schedule firing, a message arriving, or you opening a chat)"
+          }
+        />
+      </DropdownMenuItem>
+      <DropdownMenuSeparator />
+    </>
+  );
+}
+
+function MenuItemBody({
+  title,
+  description,
+}: {
+  title: string;
+  description: string;
+}) {
+  return (
+    <span className="flex max-w-72 flex-col gap-0.5">
+      <span>{title}</span>
+      <span className="whitespace-normal text-xs text-muted-foreground">
+        {description}
+      </span>
+    </span>
+  );
+}
--- a/packages/ui/src/modules/agents/components/power-menu-items.tsx
+++ b/packages/ui/src/modules/agents/components/power-menu-items.tsx
@@ -9,16 +9,16 @@ import type { AgentView } from "../../../types.js";
 
 export function FreeUpComputeItems({
   agent,
-  configureLabel,
   onPause,
   onStop,
 }: {
   agent: AgentView;
-  configureLabel: string;
   onPause: () => void;
   onStop: () => void;
 }) {
   const neverHibernates = agent.hibernationTimeoutMin === 0;
+  const pauseUnavailableReason =
+    "This agent is set to never hibernate, so pausing it would wake itself. Change its idle timeout in Agent Setup.";
   return (
     <>
       <DropdownMenuLabel>Free up compute</DropdownMenuLabel>
@@ -37,17 +37,16 @@ export function FreeUpComputeItems({
         disabled={neverHibernates}
         onSelect={onPause}
         data-testid="agent-pause"
-        title={
-          neverHibernates
-            ? `This agent is set to never hibernate, so pausing it would wake itself. Change its idle timeout in ${configureLabel}.`
-            : undefined
-        }
+        title={neverHibernates ? pauseUnavailableReason : undefined}
       >
         <MenuItemBody
           title="Pause"
           tag={neverHibernates ? "Unavailable" : undefined}
           description="Wakes on any action (a schedule firing, a message arriving, or you opening a chat)"
         />
+        {neverHibernates && (
+          <span className="sr-only">{pauseUnavailableReason}</span>
+        )}
       </DropdownMenuItem>
       <DropdownMenuSeparator />
     </>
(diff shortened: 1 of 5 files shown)


packages/ui/src/modules/agents/components/power-menu-items.tsx
```diff
@@ -0,0 +1,67 @@
...
+        data-testid="agent-stop"
+      >
+        <MenuItemBody
+          title="Stop"
+          description="Wakes only when a schedule fires or you start it"
+        />
+      </DropdownMenuItem>
+      <DropdownMenuItem
+        className="h-auto py-2"
+        disabled={neverHibernates}
```
dam-code-guardian, line 37: 🟡 **Warning:** `agent.hibernationTimeoutMin` carries the *effective* timeout (`packages/api-server-api/src/modules/agents/router.ts:39` maps `effectiveHibernationTimeoutMin`), so a cluster default of `"0s"` (`deploy/helm/platform/values.yaml:938`) disables Pause for agents whose owner set nothing. On `origin/main` Pause was offered for these agents and `agents-service.ts:1054` degraded it to the sticky stop that `docs/architecture/agent-lifecycle.md:221` documents as "the only stable *paused* it can have". The description states the opposite reason, and names "Configure agent", a label only the sandbox list uses.
  **Fix:** Do not disable an action whose documented result still works — keep Pause enabled at effective timeout 0 and describe the documented outcome (it becomes a sticky stop), or change `agent-lifecycle.md:221` first. Copy that names a setting must name the control the user sees, and must not present a resolved value as a per-agent choice.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `3a60e73`

## PR #3603: fix(ui): power menus explain stop vs pause under a free-up-compute section
**Author:** kapetr | **Branch:** fix/3059-power-menu-wording → main | **Changes:** +100 −21 (5 files)

### Summary
The PR moves Stop and Pause into one shared `FreeUpComputeItems` fragment under a new "Free up compute" label, gives each item a description, and aligns the compute-usage tooltip wording. It also disables Pause for an agent whose effective hibernation timeout is 0.

### Findings
- 🟡 **Warning:** Pause is now disabled when `agent.hibernationTimeoutMin === 0`, and that field carries the *effective* timeout (`router.ts:39` maps `effectiveHibernationTimeoutMin`), so a cluster default of `"0s"` (`values.yaml:938`) disables Pause for agents whose owner set nothing. On `origin/main` Pause was offered for these agents and the server degraded it to the sticky stop that `docs/architecture/agent-lifecycle.md:221` documents as "the only stable *paused* it can have"; the new copy states the opposite reason – "pausing it would wake itself" – and points at "Configure agent", a name that only the sandbox list uses (`knowledge-bases-list-view.tsx:100` says "Configure knowledge base agent"). Reported by doc-drift, issue-fit and react-ui-engineering. (`packages/ui/src/modules/agents/components/power-menu-items.tsx:34`)
  **Fix:** Do not disable an action whose documented result still works – keep Pause enabled at effective timeout 0 and describe the documented outcome (it becomes a sticky stop), or change `agent-lifecycle.md:221` first. Copy that names a setting must name the control the user sees, and must not present a resolved value as a per-agent choice.
- ✅ **Looks good:** the server side is untouched by the UI gate — `agents-service.ts:1054` still routes a pause to `requestStop` at effective timeout 0, and both call sites keep their existing handlers (`agent-row.tsx:147`, `sandbox-home-header.tsx:123`).

- Its finding (the never-hibernate pause path documented at `agent-lifecycle.md:221` becomes unreachable from the UI) is merged into the 🟡 in `### Findings`.

### Issue Fit
- Its finding (the copy points at a control named "Configure agent") is merged into the 🟡 in `### Findings`.

- Its finding on the magic-zero sentinel and the copy that asserts a user-set value is merged into the 🟡 in `### Findings`.
- 🟢 **Suggestion:** the knowledge-bases list spreads `rowProps` unchanged, so its Stop item runs `suspend.stop` with no confirmation, while the sandbox list and the sandbox header both confirm first (`packages/ui/src/modules/knowledge-bases/views/knowledge-bases-list-view.tsx:91`). The wiring is pre-existing; the new section makes Stop more prominent, so it is worth a separate issue.
- Its 9 remaining items (the trailing separator inside the fragment, the whole `AgentView` passed for one field, the label typography inlined instead of `menu-styles.ts`, the `h-auto py-2` height override, the remedy text inside a disabled item, `max-w-72` in a leaf, the component named after visible copy, the two unused testids, the inline arrow handlers) did not reach the blocking bar and fall outside the 🟢 budget.

### Verdict
COMMENT — the menu now refuses a documented action for agents that inherit a zero cluster timeout, and explains it with a reason the lifecycle doc contradicts.


---

packages/ui/src/modules/agents/components/power-menu-items.tsx
kapetr: @dam-code-guardian Disabling Pause at timeout 0 and its copy come straight from the issue and design prototype — a real pause can't exist for an always-on agent (the server degrades it to a sticky stop), and offering a Pause that secretly acts as Stop is the confusion this issue removes.

dam-code-guardian: Agreed, and the warning is withdrawn. Issue #3059 settles it: on 2026-08-11 @jenna-winkler decided to disable the pause button for an agent with a 0 hibernation timeout, because an always-on agent has no real pause. I read the lifecycle doc as the authority on the desired behaviour, but the issue is the authority here. I recorded this as an override on PR #3603, so a later review will not raise it again.

One narrow point stays, and it is a suggestion at most. The disabled copy sends the user to "Configure agent". The agent list uses that label, but the knowledge-bases list calls the same control "Configure knowledge base agent" (`knowledge-bases-list-view.tsx:100`), and the agent page header shows no such label at all (`sandbox-home-header.tsx`). A user who opens the menu from the agent page cannot find the named control. Naming the setting, not the control, would avoid this.

packages/ui/src/modules/agents/components/power-menu-items.tsx
```diff
@@ -0,0 +1,81 @@
...
+          title="Stop"
+          description="Wakes only when a schedule fires or you start it"
+        />
+      </DropdownMenuItem>
+      <DropdownMenuItem
+        className="h-auto py-2 data-[disabled]:pointer-events-auto"
+        disabled={neverHibernates}
+        onSelect={onPause}
+        data-testid="agent-pause"
+        title={
```
dam-code-guardian: 🟡 **Warning:** the reason Pause is unavailable is now carried only by a native `title` on a `disabled` Radix `DropdownMenuItem`. Radix skips disabled items in keyboard navigation and typeahead, so a keyboard or screen-reader user reaches the "Unavailable" badge and never the explanation; `data-[disabled]:pointer-events-auto` restores it for mouse hover only. The previous commit had this text in the rendered description, where everyone got it, and `agent-row.tsx:114` in this same file's neighbour already shows the accessible form — a `title` paired with `tabIndex`/`role`/`aria-label`.
  **Fix:** any explanation for why a control is disabled must be reachable without a pointer — expose it to assistive tech (`aria-label`/`aria-describedby` on the item, or the repo's `Tooltip` primitive on a focusable wrapper), never via `title` alone.

packages/ui/src/modules/agents/components/power-menu-items.tsx
```diff
@@ -0,0 +1,81 @@
...
+          neverHibernates
+            ? `This agent is set to never hibernate, so pausing it would wake itself. Change its idle timeout in ${configureLabel}.`
```
dam-code-guardian: 🟢 **Suggestion:** `configureLabel` is the *menu item caption* at the agent-list call sites (`sandbox-list.tsx:40` passes "Configure agent", `knowledge-bases-list-view.tsx:100` passes "Configure knowledge base agent"), and the same string is spliced into this sentence — so it renders "Change its idle timeout in Configure agent." Only `sandbox-home-header.tsx:123` passes a fragment that reads grammatically ("the setup section below"). A separate prop for the prose location would keep one prop to one meaning.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `f8cf66b`

## PR #3603: fix(ui): stop and pause actions explained in agent actions menu
**Author:** kapetr | **Branch:** fix/3059-power-menu-wording → main | **Changes:** +116 −21 (5 files)

### Summary
The new commit moves the disabled-Pause explanation out of the visible item description into a native `title` tooltip plus an "Unavailable" badge, and threads a `configureLabel` prop into that copy from both power menus.
_(Suppressed 1 finding(s) per PR-local overrides: pause-disabled-at-zero-timeout.)_

### Changes since last review
Previous HEAD: 3a60e73 (2026-09-08T12:25:12Z) — verdict COMMENT

- 🔁 **Still present:** knowledge bases Stop runs with no confirmation (`packages/ui/src/modules/knowledge-bases/views/knowledge-bases-list-view.tsx:91`)
- 🆕 **New:** disabled-Pause reason is now pointer-only (`packages/ui/src/modules/agents/components/power-menu-items.tsx:40`)
- 🆕 **New:** menu-item caption spliced into prose (`packages/ui/src/modules/agents/components/power-menu-items.tsx:42`)

### Findings
- 🟡 **Warning:** disabled-Pause reason is now pointer-only (`packages/ui/src/modules/agents/components/power-menu-items.tsx:40`)
- 🟢 **Suggestion:** menu-item caption spliced into prose (`packages/ui/src/modules/agents/components/power-menu-items.tsx:42`)

- Its finding (`docs/architecture/agent-lifecycle.md:221` still documents the never-hibernate pause the UI now refuses) is on a file this commit did not touch and follows from the decision the PR-local override records, so it stays a 🟢 outside the budget.

### Issue Fit
- Its finding on the reason living in a hover-only `title` is merged into the 🟡 in `### Findings`; its finding on the interpolated label is the 🟢 there.

- 🔁 1 finding from the previous review still present (see review at 3a60e73).
- Its `title`-on-a-disabled-item finding is merged into the 🟡, and its `configureLabel` finding into the 🟢, in `### Findings`.
- Its 6 remaining items (the hand-rolled `DropdownMenuLabel` recipe diverging from `labelVariants`, the trailing separator inside the fragment, the un-extracted Wake/Restart/Delete duplication, `tag?: string` as an open string, the whole `AgentView` passed for one field, the untested new branch) are off-delta or style seams and fall outside the 🟢 budget.

### Verdict
COMMENT — the round replaces visible copy with a hover-only tooltip, so the explanation this PR exists to deliver no longer reaches a keyboard or screen-reader user.


---

packages/ui/src/modules/agents/components/power-menu-items.tsx
PetrBulanek: `configureLabel` holds the text of a menu command, so dropping it after "in" builds a broken sentence: the agent list renders "Change its idle timeout in Configure agent.", and the knowledge-bases list renders "in Configure knowledge base agent."

The agent page passes "the setup section below" instead, but this menu appears on all seven sections of that page. I opened it on Usage with the timeout at 0, and the tooltip pointed below at nothing — the field is in the "Agent Setup" entry of the left-hand nav.

One string is doing two jobs here, as a command label and as a place name inside a sentence, and it cannot read correctly as both.

PetrBulanek (approved):

PR #3604, opened 2026-09-08
fix(api-server): a reconnecting Agent watch holds one connection and one retry timer

### LINKED ISSUE #3494: A reconnecting Agent watch can leave a second connection live and a timer orphaned
## Context

Raised in review on #3493, which noted the same shape in two places and suggested one issue covering both.

## Problem

The Agent watch adapter reconnects by calling its own `connect()` from the watch's end callback, and that overwrites the stored stop function and retry timer without releasing the previous ones. If a single connection's end callback fires more than once — or a reconnect races a still-live connection — the result is two watches feeding the same handler and a timer that `stop()` can no longer reach.

Nothing observed this in production. It matters because the duplicate is silent: events arrive twice, the second watch is unreferenced so shutdown cannot close it, and an orphaned timer keeps the process from settling in tests.

## Goal

A watch adapter that reconnects holds at most one live connection and one pending timer, and `stop()` releases whatever exists.

## Additional resources

- `packages/api-server/src/modules/live-events/infrastructure/k8s-agent-watch.ts` — the instance that remains on main
- Raised on #3493

jezekra1:
The lease-elected Agent watch reconnects by calling its own `connect()` from the end callback of the connection that just died. That overwrote the stored stop function and retry timer without releasing the previous ones. If one connection reported its end twice, or a reconnect raced a still-live connection, two watches fed the same handler, the second one was unreferenced so shutdown could not close it, and an orphaned timer kept the process from settling in tests. Raised in review on #3493, tracked as #3494.

## What changed

- **Each end callback is tied to its own connection.** Every connect records a link as the current connection. The end callback returns early unless its link is still current, clears the slot, releases the ended connection, and only then schedules the retry. A duplicate end from the same connection, or a late end from a stale one, is a no-op.
- **One connection, one timer, by construction.** Only the retry timer opens a new connection, and only an end that is still current schedules that timer. So at most one connection and one pending retry timer exist, and `stop()` always reaches the live one.
- **Tests** join the existing watch spec: a doubled end produces exactly one reconnect and releases the ended connection (three connections opened before this change), `stop()` after a reconnect closes the reconnected connection, and `stop()` during a pending retry cancels it.

## Not touched

- The replay sweep and its timer already run at most once per connect and are cleared on `stop()`.
- The installed client library already guards its own done callback against firing twice, so nothing observed this in production. The adapter no longer depends on that guarantee.

No architecture page changes: the reconnect and replay-sweep behaviour on platform-topology is unchanged.

Closes #3494

### DIFF
--- a/packages/api-server/src/__tests__/unit/k8s-agent-watch.test.ts
+++ b/packages/api-server/src/__tests__/unit/k8s-agent-watch.test.ts
@@ -1,4 +1,4 @@
-// TEST_OVERVIEW: the lease-elected K8s agent watch projects resource transitions into per-owner invalidation hints, including deletions that happened while the watch was disconnected.
+// TEST_OVERVIEW: the lease-elected K8s agent watch projects resource transitions into per-owner invalidation hints, including deletions that happened while the watch was disconnected. Its reconnect holds at most one live connection and one pending retry timer, and stop() releases whatever exists.
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { startAgentWatch } from "../../modules/live-events/infrastructure/k8s-agent-watch.js";
 import type { LiveEventsBus } from "../../modules/live-events/services/live-events-service.js";
@@ -23,13 +23,20 @@ describe("k8s agent watch", () => {
       publish: (ownerSub: string, hint: { agentId?: string }) =>
         void hints.push({ ownerSub, agentId: hint.agentId }),
     } as unknown as LiveEventsBus;
-    const conns: { onEvent: OnEvent; onEnd: OnEnd }[] = [];
+    const conns: { onEvent: OnEvent; onEnd: OnEnd; stopCalls: number }[] = [];
     const watch = startAgentWatch(
       bus,
       {
         watchCustomObjects: (_plural, onEvent, onEnd) => {
-          conns.push({ onEvent: onEvent as OnEvent, onEnd: onEnd as OnEnd });
-          return () => {};
+          const conn = {
+            onEvent: onEvent as OnEvent,
+            onEnd: onEnd as OnEnd,
+            stopCalls: 0,
+          };
+          conns.push(conn);
+          return () => {
+            conn.stopCalls += 1;
+          };
         },
       },
       { plural: "agents", ownerLabel: "owner", log: () => {}, debounceMs: 1 },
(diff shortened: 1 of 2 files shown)


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `2d6709f`

## PR #3604: fix(api-server): a reconnecting Agent watch holds one connection and one retry timer
**Author:** jezekra1 | **Branch:** fix/agent-watch-reconnect → main | **Changes:** +56 −7 (2 files)

### Summary
Ties each `onEnd` callback to the connection that produced it, so a duplicate or
late end report can no longer null a live connection or schedule a second
reconnect. The identity guard, the release of the ended handle and `stop()`
closing the current connection are all consistent, and the three new specs cover
the double-end, reconnect-then-stop and pending-retry paths.

### Findings
✅ No findings.

- ✅ Looks good: the `link.stop()` inside the guard looked like a self-release of
  a handle that may not exist yet, but `Watch.watch()` in the pinned
  `@kubernetes/client-node@1.4.0` awaits `applyToFetchOptions` before any path
  to its `done` callback, and the adapter routes rejections through `.catch`
  (`packages/api-server/src/modules/agents/infrastructure/k8s.ts:176-201`), so
  `onEnd` never fires in the calling turn and `link.stop` is always the real
  stopper by then.

✅ No findings.

### Issue Fit
✅ No findings.

Closing keyword resolves to #3494, whose goal — one live connection, one pending
timer, `stop()` releasing whatever exists — is delivered by the per-connection
guard. The "two places" raised on #3493 is fully covered: the sibling
`agent-change-watch.ts` no longer exists and `agent-state-cache.ts` uses the
client informer with its own restart path, so the one named file is the whole
scope.

- 🟢 **Suggestion:** The release runs through a mutable placeholder
  (`const link = { stop: () => {} }`) that is filled in after the call which can
  trigger it, so the module's safety rests on `watchCustomObjects` never calling
  `onEnd` synchronously. That holds for the pinned client, but nothing in this
  module or its tests pins it (`packages/api-server/src/modules/live-events/infrastructure/k8s-agent-watch.ts:96`).
- 🟢 **Suggestion:** `stop()` calls `connection?.stop()` and leaves `connection`
  set, so a second `stop()` — or a late `onEnd` from the connection live at
  shutdown — invokes the underlying stopper again; idempotence lives in the k8s
  adapter rather than in this module's own state (`packages/api-server/src/modules/live-events/infrastructure/k8s-agent-watch.ts:113`).
- 🟢 **Suggestion:** The test double counts `stopCalls` but keeps delivering
  events after being stopped, so the specs prove the stopper was called and not
  that the ended connection stopped feeding `onEvent` — the invariant the new
  guard exists for (`packages/api-server/src/__tests__/unit/k8s-agent-watch.test.ts:37`).

### Verdict
APPROVE — the reconnect fix is correct on every path I could drive, and the one
blocking claim raised against it does not hold for the pinned watch client.


---

PR #3611, opened 2026-09-08
fix(api-server): NUL-byte ids answer 400, WebSocket upgrades carry HSTS

JanPokorny:
Follow-up from re-running the ZAP scan against dam-dev after #3602 deployed.

## Findings from the second pass

- 28 procedures still answered **500** to one specific payload: an id containing a NUL byte (U+0000). Postgres rejects it with SQLSTATE `22021` (`invalid byte sequence for encoding "UTF8": 0x00`), which the new error mapping did not cover. The client got a redacted "internal server error" (the redaction from #3602 works), but it should be a 400.
- The `/api/trpc-ws` upgrade response (101) carried no `Strict-Transport-Security`: the upgrade path bypasses the Hono middleware that sets the security headers.
- 415 responses echoed the request's `Content-Type` header verbatim (`Unsupported content-type "<scrIpt>alert(1);</scRipt>"`), in a `text/plain` body. Not exploitable with `nosniff`, but reflected attacker input all the same.

Everything else from the first scan is gone on the deployed build: no stack traces or raw SQL in any of ~53k API responses, `Server: nginx` without version, HSTS/nosniff on static assets, `Cache-Control: no-store` on API responses, router affinity cookie gone. The six new boolean-based SQL-injection alerts were again false positives (different Zod validation errors compared as different pages).

## What

- `trpc.ts`: `22021` -> `BAD_REQUEST` with its own client message ("unsupported character in input"); each mapped SQLSTATE keeps its message beside its code. The message redaction covers `UNSUPPORTED_MEDIA_TYPE` as well as `INTERNAL_SERVER_ERROR`. Two new test scenarios.
- `upgrade.ts`: `addUpgradeSecurityHeaders(wss)` hooks the `ws` `headers` event on all five WebSocket servers (tRPC, ACP, terminal, SSH, agent-tRPC relays), and the relay's hand-written 401/403 denial carries the same HSTS line.
- `platform-topology.md`: names the media-type exception to the "caller-caused errors keep their message" rule.

https://claude.ai/code/session_01BGaKuoKWJad1BVu4FpQeXW

PR #3613, opened 2026-09-08
fix(api-server): NUL byte inside a jsonb value answers 400

JanPokorny:
Third ZAP pass against dam-dev after #3611 deployed. The run is otherwise clean: zero alerts above Informational, zero stack traces or SQL text across ~41k API responses, and exactly one 500 left.

## Finding

`skills.sets.create` with a NUL byte (U+0000) nested inside `skills[].name` still answered 500. The value lands in a jsonb column, and Postgres rejects NUL there as SQLSTATE `22P05` (`untranslatable_character`, "unsupported Unicode escape sequence") rather than the `22021` that plain text columns raise and that #3611 mapped.

## What

- `trpc.ts`: `22P05` -> `BAD_REQUEST` "unsupported character in input", beside the existing `22021` entry.
- one more scenario in `trpc-errors.test.ts`.

https://claude.ai/code/session_01BGaKuoKWJad1BVu4FpQeXW

### DIFF
--- a/packages/api-server-api/src/trpc.ts
+++ b/packages/api-server-api/src/trpc.ts
@@ -23,6 +23,7 @@ function isTermsStaleCause(cause: unknown): boolean {
 
 const PG_INVALID_TEXT_REPRESENTATION = "22P02";
 const PG_CHARACTER_NOT_IN_REPERTOIRE = "22021";
+const PG_UNTRANSLATABLE_CHARACTER = "22P05";
 const PG_UNIQUE_VIOLATION = "23505";
 const PG_ERRORS: Record<string, { code: TRPCError["code"]; message: string }> =
   {
@@ -31,6 +32,10 @@ const PG_ERRORS: Record<string, { code: TRPCError["code"]; message: string }> =
       message: "malformed identifier",
     },
     [PG_CHARACTER_NOT_IN_REPERTOIRE]: {
+      code: "BAD_REQUEST",
+      message: "unsupported character in input",
+    },
+    [PG_UNTRANSLATABLE_CHARACTER]: {
       code: "BAD_REQUEST",
       message: "unsupported character in input",
     },
--- a/packages/api-server/src/__tests__/unit/trpc-errors.test.ts
+++ b/packages/api-server/src/__tests__/unit/trpc-errors.test.ts
@@ -41,6 +41,13 @@ describe("tRPC error mapping", () => {
     });
   });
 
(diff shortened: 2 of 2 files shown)


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `594b8ed`

## PR #3613: fix(api-server): NUL byte inside a jsonb value answers 400
**Author:** JanPokorny | **Branch:** fix/zap-findings-3 → main | **Changes:** +12 −0 (2 files)

### Summary
The PR maps Postgres SQLSTATE `22P05` (`untranslatable_character`) to `BAD_REQUEST`, beside the `22021` entry that #3611 added, and adds one test scenario. This closes the last 500 of the third ZAP pass: a NUL byte nested in a jsonb value raises `22P05`, not the `22021` that plain text columns raise.

### Findings
- ✅ **Looks good:** the new mapping is reachable on the reported path. `skills.sets.create` runs on `manageAgentsProcedure`, which builds on `t.procedure` and therefore carries `mapPgErrors`, and `createSet`'s own `catch` intercepts unique violations only and rethrows everything else (`packages/api-server/src/modules/skills/services/skills-service.ts:803`). The table literal supplies the client message, so the raw query text does not reach the caller.

✅ No findings.

### Issue Fit
✅ No findings.

- 🟢 **Suggestion:** the harness surface in the same package builds its own root with no `errorFormatter` and no `mapPgErrors`, so the same NUL-byte class on a harness procedure still answers 500 with the Postgres message. The surface is pre-existing and reachable only by an authenticated agent pod through `resolveAgent`, so this is a divergence to close once, not a defect of this diff (`packages/api-server-api/src/harness-trpc.ts:5`).
- 🟢 **Suggestion:** the contract package now owns three SQLSTATE constants and their translation table, while the server package owns `23505` independently in `packages/api-server/src/core/db-errors.ts:12`. One home for SQLSTATE recognition would stop the two drifting (`packages/api-server-api/src/trpc.ts:26`).
- 🟢 **Suggestion:** `22P05` also fires when the server encoding cannot represent a character in server-generated data. The unconditional `BAD_REQUEST` turns that class into a 400 with no server-side signal, because `logInternalError` reports `INTERNAL_SERVER_ERROR` alone (`packages/api-server-api/src/trpc.ts:38`).

### Verdict
APPROVE — the mapping matches the SQLSTATE the reported path raises, it is reached on that path, and no blocking finding survived verification.


---

PR #3621, opened 2026-09-09
fix(ui): an old session no longer opens looking like it is still running

xjacka:
Opening a finished session from the list left an empty assistant bubble streaming at the end of the transcript, and the busy flag derived from it marked the row in the session list as running. Loading the very same session by URL rendered it correctly.

## Cause

The replay is collected only while the `session/load` call is awaited — the collector is detached in a `finally` as soon as the call resolves. The runtime answers before the harness has replayed the transcript, so those updates can arrive after the collector is gone and are applied as live ones. A replayed transcript carries no turn-end marker (only the live stream does), so the last reply stays streaming.

It is a race, which is why the symptom was asymmetric: on a full page load the updates won and the replay was collected; on an in-app click the response won.

## Fix

The load response already carries the answer: the runtime reports on `_meta.platform.turn.inFlight` whether a turn is in flight — an existing contract this change consumes, nothing on the runtime side changes. When it reports none, nothing in that session may render as running, no matter which path the updates took to reach the store.

Three things the naive form of this would get wrong, all handled here:

- A prompt sent into a session you just opened must still render as running, so sending clears the mark (delivery failure detection reads `bubble.streaming` and keeps working).
- The transcript settles once the replay goes quiet rather than immediately, so a late chunk arriving after a closed bubble cannot split one reply across several bubbles — `appendToActive` only appends to a streaming message.
- The mark is bounded, so it cannot act on a turn it was never read for. A turn started elsewhere for the same session — the trigger driver, or a second client — arrives here as plain chunks and cannot be told apart by update kind, because `promptStarted` is sent only to the channel that sent the prompt (`prompt-scheduler.ts:146`). So the mark is consumed by the settle that acts on it, and expires shortly after the load that recorded it.

Files changed: `use-acp-connection.ts`, `use-acp-session.ts`. Not harness-specific — nothing on this path branches per harness; it was simply reproduced on Bob.

## Verification

`mise run check` and `mise run //packages/ui:test` (43 files, 366 tests) pass. `check:comment-types` fails on a pre-existing config parse error (`unknown field deny_net` at `.mise/config.toml:98` under mise 2026.9.3, which is the declared `min_version`); comment types were verified with `mise run strip-comments -- --check` instead.

An earlier form of this fix was verified against the original reproduction on a live cluster — the two sessions that showed the indicator on an in-app switch showed none afterwards. **This revision has not been re-run on a cluster yet.**

## Tests

No unit test. The logic lives in a hook, not a pure function, and the suite has no React Testing Library setup — all 366 tests are pure-function tests. Extracting a named predicate to test would be a design change rather than added coverage; happy to do it if reviewers want it.

### DIFF
--- a/packages/ui/src/modules/sessions/hooks/use-acp-connection.ts
+++ b/packages/ui/src/modules/sessions/hooks/use-acp-connection.ts
@@ -71,6 +71,8 @@ export interface UseAcpConnectionResult {
     sid: string,
     replayBefore?: string,
   ) => Promise<Message[]>;
+  runtimeIdle: (sid: string) => boolean;
+  clearRuntimeIdle: (sid: string) => void;
   connectionRef: React.MutableRefObject<LiveConnection | null>;
   reset: () => void;
 }
@@ -282,6 +284,7 @@ export function useAcpConnection(
   }, [selectedAgent, makeUpdateHandler, attachCloseHandler]);
 
   const loadChainRef = useRef<Promise<unknown>>(Promise.resolve());
+  const idleSessionsRef = useRef(new Set<string>());
 
   const runSessionLoad = useCallback(
     async (sid: string, replayBefore?: string): Promise<Message[]> => {
@@ -370,6 +373,8 @@ export function useAcpConnection(
           ? turn.data.interruptedAt
           : undefined,
       );
+      if (turn.success && !turn.data.inFlight) idleSessionsRef.current.add(sid);
+      else idleSessionsRef.current.delete(sid);
       if (replayBefore === undefined && generation === generationRef.current) {
         bindEngagement(sid);
         pendingReloadRef.current = false;
@@ -508,11 +513,22 @@ export function useAcpConnection(
     reset();
   }, [sessionId, sessionMode, reset]);
 
+  const runtimeIdle = useCallback(
+    (sid: string): boolean => idleSessionsRef.current.has(sid),
+    [],
+  );
+
+  const clearRuntimeIdle = useCallback((sid: string): void => {
+    idleSessionsRef.current.delete(sid);
+  }, []);
+
   return {
     state,
     ensureLive,
     beginSession,
     loadSessionHistory,
+    runtimeIdle,
+    clearRuntimeIdle,
     connectionRef,
     reset,
   };
--- a/packages/ui/src/modules/sessions/hooks/use-acp-connection.ts
+++ b/packages/ui/src/modules/sessions/hooks/use-acp-connection.ts
@@ -375,10 +375,10 @@ export function useAcpConnection(
           ? turn.data.interruptedAt
           : undefined,
       );
-      if (turn.success && !turn.data.inFlight)
-        idleSessionsRef.current.set(sid, Date.now());
-      else idleSessionsRef.current.delete(sid);
       if (replayBefore === undefined && generation === generationRef.current) {
+        if (turn.success && !turn.data.inFlight)
+          idleSessionsRef.current.set(sid, Date.now());
+        else idleSessionsRef.current.delete(sid);
         bindEngagement(sid);
         pendingReloadRef.current = false;
         setState("live");
--- a/packages/ui/src/modules/sessions/hooks/use-acp-connection.ts
+++ b/packages/ui/src/modules/sessions/hooks/use-acp-connection.ts
@@ -28,6 +28,8 @@ import { draftKey } from "../lib/draft-key.js";
 import { clearUndelivered, readUndelivered } from "../lib/undelivered-store.js";
 import type { PromptDelivery } from "./use-prompt-delivery.js";
 
+const REPLAY_IDLE_WINDOW_MS = 3000;
+
 export interface LiveConnection {
   connection: ClientSideConnection;
   ws: WebSocket;
@@ -284,7 +286,7 @@ export function useAcpConnection(
   }, [selectedAgent, makeUpdateHandler, attachCloseHandler]);
 
   const loadChainRef = useRef<Promise<unknown>>(Promise.resolve());
-  const idleSessionsRef = useRef(new Set<string>());
+  const idleSessionsRef = useRef(new Map<string, number>());
 
   const runSessionLoad = useCallback(
     async (sid: string, replayBefore?: string): Promise<Message[]> => {
@@ -373,7 +375,8 @@ export function useAcpConnection(
           ? turn.data.interruptedAt
           : undefined,
       );
-      if (turn.success && !turn.data.inFlight) idleSessionsRef.current.add(sid);
+      if (turn.success && !turn.data.inFlight)
+        idleSessionsRef.current.set(sid, Date.now());
       else idleSessionsRef.current.delete(sid);
       if (replayBefore === undefined && generation === generationRef.current) {
         bindEngagement(sid);
@@ -513,10 +516,13 @@ export function useAcpConnection(
     reset();
   }, [sessionId, sessionMode, reset]);
 
-  const runtimeIdle = useCallback(
-    (sid: string): boolean => idleSessionsRef.current.has(sid),
-    [],
-  );
+  const runtimeIdle = useCallback((sid: string): boolean => {
+    const answeredAt = idleSessionsRef.current.get(sid);
+    if (answeredAt === undefined) return false;
+    if (Date.now() - answeredAt <= REPLAY_IDLE_WINDOW_MS) return true;
+    idleSessionsRef.current.delete(sid);
+    return false;
+  }, []);
 
   const clearRuntimeIdle = useCallback((sid: string): void => {
     idleSessionsRef.current.delete(sid);
--- a/packages/ui/src/modules/sessions/hooks/use-acp-session.ts
+++ b/packages/ui/src/modules/sessions/hooks/use-acp-session.ts
@@ -2,6 +2,7 @@ import { SessionMode } from "api-server-api";
 import { useCallback, useEffect, useRef, useState } from "react";
 
 import { useStore } from "../../../store.js";
+import type { Attachment } from "../../../types.js";
 import {
   classifyResumeError,
   resumeFailureKind,
@@ -10,6 +11,7 @@ import {
 } from "../../acp/errors.js";
 import {
   appendUndelivered,
+  finalizeAllStreaming,
   hasStreamingAssistant,
 } from "../../acp/session-projection.js";
 import { useIsAgentOperable } from "../../agents/api/queries.js";
@@ -18,10 +20,12 @@ import { setSessionRunning } from "../api/queries.js";
 import { draftKey } from "../lib/draft-key.js";
 import { readUndelivered } from "../lib/undelivered-store.js";
 import { useAcpConnection } from "./use-acp-connection.js";
-import { useAcpPrompt } from "./use-acp-prompt.js";
+import { type SendPromptOptions, useAcpPrompt } from "./use-acp-prompt.js";
 import { useAcpSessionEngagement } from "./use-acp-session-engagement.js";
 import { useAcpUpdateHandler } from "./use-acp-update-handler.js";
 import { usePromptDelivery } from "./use-prompt-delivery.js";
+
+const REPLAY_SETTLE_MS = 150;
 
 async function classifyResumeFailure(
   agentId: string,
@@ -81,6 +85,8 @@ export function useAcpSession(
     ensureLive,
     beginSession,
     loadSessionHistory,
+    runtimeIdle,
+    clearRuntimeIdle,
     connectionRef,
     state: connectionState,
     reset: resetConnection,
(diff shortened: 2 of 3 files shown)


packages/ui/src/modules/sessions/hooks/use-acp-session.ts
```diff
@@ -98,6 +104,17 @@ export function useAcpSession(
     delivery,
   });
 
+  useEffect(() => {
```
dam-code-guardian, line 107: 🟡 **Warning:** The idle mark is cleared only by a local `sendPrompt`, or by the next `runSessionLoad`, but this effect re-arms on every `messages` change while the mark is set. A turn that starts on another surface for the same session — the trigger driver, or a second client — still reaches this tab: `transcript.append` fans a live update out to every engaged channel (`session-transcript.ts:204`) and `useAcpUpdateHandler` writes it into the store. Each chunk re-arms the 150 ms timer, so the first gap longer than 150 ms — a tool call, for example — finalizes a bubble that is still live: `busy` becomes false, `setSessionRunning` marks the row idle while the agent works, and the next chunk makes `appendToActive` open a second bubble, because no active assistant remains.
**Fix:** A one–shot correction consumes its trigger: clear the idle record when the settle runs, and do not re-arm it while live updates arrive for that session.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `b941d80`

## PR #3621: fix(ui): an old session no longer opens looking like it is still running
**Author:** xjacka | **Branch:** fix/session-opens-looking-running → main | **Changes:** +47 −2 (2 files)

### Summary
The load response's `turn.data.inFlight` now records a session as runtime-idle, and a 150 ms settle finalizes the streaming bubble that a replayed transcript leaves open — so an old session no longer reads as running. The mechanism is correct for the reported path; the idle mark, however, outlives that path.

_4 suggestion(s) dropped under the 🟢 budget._

### Findings

- 🟡 **Warning:** The idle mark is cleared only by a local `sendPrompt`, or by the next `runSessionLoad`, but the settle effect re-arms on every `messages` change while the mark is set. A turn that starts on another surface for the same session — the trigger driver, or a second client — still reaches this tab: `transcript.append` fans a live update out to every engaged channel (`session-transcript.ts:204`) and `useAcpUpdateHandler` writes it into the store. Each chunk re-arms the 150 ms timer, so the first gap longer than 150 ms — a tool call, for example — finalizes a bubble that is still live: `busy` becomes false, `setSessionRunning` marks the row idle while the agent works, and the next chunk makes `appendToActive` open a second bubble, because no active assistant remains (`packages/ui/src/modules/sessions/hooks/use-acp-session.ts:107`).
  **Fix:** A one–shot correction consumes its trigger: clear the idle record when the settle runs, and do not re-arm it while live updates arrive for that session.

✅ No findings.

### Issue Fit

- 🟡 **Warning:** The PR body's `## Scope` section presents the change as four files — the UI path (`use-acp-connection.ts`, `use-acp-session.ts`) and the runtime path (`session-bootstrap.ts`, `session-transcript.ts`) — but the diff touches only the two UI hooks. Both runtime files are identical to `origin/main`, which already reports `inFlight` on the load response. A reader who sizes the change from the body looks for a runtime half that is not here, and reads a pre-existing contract as new work (`packages/agent-runtime/src/modules/acp/services/acp-runtime/session-bootstrap.ts:179`).
  **Fix:** A scope section names only the files the diff changes; code the change consumes is described as an existing contract, never as delivered surface.

- 🟢 **Suggestion:** `UseAcpConnectionResult` gains an imperative `runtimeIdle` / `clearRuntimeIdle` pair over a mutable `Set` in a ref, so the port exports transport bookkeeping and the consumer maintains the adapter's invariant — `use-acp-session.ts:221` must remember to clear before every send. Both accessors are memoized with an empty dependency list over mutable state, so they are inert as effect dependencies; the effect re-runs only because `messages` changes (`packages/ui/src/modules/sessions/hooks/use-acp-connection.ts:516`).

### Verdict
COMMENT — the fix works for the path it targets, but the idle mark has no exit, which puts a live stream from another surface at risk.


---

xjacka:
Both warnings addressed in 46ad7a4a.

**Idle mark outliving its path.** Confirmed, and the mechanism is worse than "clear it on send" can cover: a foreign turn cannot be recognised by update kind at all, because `promptStarted` is sent only to the channel that sent the prompt (`prompt-scheduler.ts:146`) — in this tab it arrives as plain chunks, indistinguishable from a late replay. So the mark is bounded on both ends instead: the settle that acts on it consumes it (one load closes the transcript once), and it expires a few seconds after the load that recorded it, since the runtime answered for that moment.

Residual risk, stated plainly: a foreign turn that starts inside that window can still be closed once. It costs one bubble split and a brief idle row, and it needs the turn to begin in the seconds right after a load that reported no turn in flight. Closing that last gap means asking the runtime again at settle time rather than trusting the recorded answer — happy to do it if you would rather not carry the window.

**Scope section.** Corrected — it named runtime files the diff does not touch. The body now names the two changed hooks and describes `_meta.platform.turn.inFlight` as the existing contract this change consumes.

**Suggestion on the getter/command pair.** Not taken. The effect has to test the mark before the delay and consume it after, so the two operations stay separate; folding them into one `consume` call would have the check spend the mark it is only testing.

packages/ui/src/modules/sessions/hooks/use-acp-connection.ts
```diff
@@ -370,6 +375,9 @@ export function useAcpConnection(
           ? turn.data.interruptedAt
           : undefined,
       );
+      if (turn.success && !turn.data.inFlight)
+        idleSessionsRef.current.set(sid, Date.now());
+      else idleSessionsRef.current.delete(sid);
```
dam-code-guardian: 🟡 **Warning:** The idle mark is written on every `runSessionLoad`, including a `replayBefore` pagination page — the adjacent block guards on `replayBefore === undefined`, this one does not. Scrolling back through history therefore re-stamps `Date.now()` and re-arms a window the body describes as expiring shortly after the load that recorded it, so the race the bound exists to close is re-opened on each page.

**Fix:** State that is only meaningful for the tail replay is written under the same `replayBefore === undefined` guard as the rest of the tail-only bookkeeping in `runSessionLoad`; a historical page load never mutates live-turn state.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `46ad7a4`

## PR #3621: fix(ui): an old session no longer opens looking like it is still running
**Author:** xjacka | **Branch:** fix/session-opens-looking-running → main | **Changes:** +54 −2 (2 files)

### Summary
Opening a finished session no longer renders a streaming bubble: the load response's `turn.inFlight` marks the session idle, and a settled replay finalizes what is still streaming. This round bounds that mark — a 3 s window plus consumption by the settle that acts on it.

_4 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: b941d80 (2026-09-09T10:20:08Z) — verdict COMMENT

- ✅ **Fixed:** idle mark never cleared, settle can close a live bubble (`packages/ui/src/modules/sessions/hooks/use-acp-session.ts:107`)
- ✅ **Fixed:** body scope lists runtime files the diff does not change (`packages/agent-runtime/src/modules/acp/services/acp-runtime/session-bootstrap.ts:179`)
- 🔁 **Still present:** port exports a mutable–ref getter/command pair (`packages/ui/src/modules/sessions/hooks/use-acp-connection.ts:519`)
- 🆕 **New:** idle mark re-stamped by history page loads (`packages/ui/src/modules/sessions/hooks/use-acp-connection.ts:378`)
- 🆕 **New:** residual window: a turn begun elsewhere can settle early (`packages/ui/src/modules/sessions/hooks/use-acp-session.ts:108`)

### Findings

- 🟡 **Warning:** The idle mark is written on every `runSessionLoad`, including a `replayBefore` pagination page — the adjacent block guards on `replayBefore === undefined`, this one does not. Scrolling back through history therefore re-stamps `Date.now()` and re-arms a window the body describes as expiring shortly after the load that recorded it, so the race the bound exists to close is re-opened on each page (`packages/ui/src/modules/sessions/hooks/use-acp-connection.ts:378`).
  **Fix:** State that is only meaningful for the tail replay is written under the same `replayBefore === undefined` guard as the rest of the tail-only bookkeeping in `runSessionLoad`; a historical page load never mutates live-turn state.

✅ No findings.

### Issue Fit
_No blocking findings; its suggestion was dropped under the 🟢 budget._

- 🟢 **Suggestion:** The 3000 ms expiry narrows the stale-idle window but leaves a residual the body records as accepted: a turn started elsewhere for the same session arrives as bare chunks, so if it starts inside the window the first 150 ms gap in that live stream finalizes a bubble that is still streaming. A positive turn-end signal read at settle time would close it (`packages/ui/src/modules/sessions/hooks/use-acp-session.ts:108`).

### Verdict
COMMENT — the previous round's blocker is genuinely fixed; one new 🟡 on the pagination path is a one-line guard.


---

xjacka:
Fixed in 35e73c06.

**Idle mark re-stamped by history page loads.** Correct, and on the pagination path it fails the other way round than the finding describes: `requestPage` answers with `withReplayMeta(..., null, ...)`, so a page carries no turn metadata at all, `platformReplayTurnMetaSchema` fails to parse, and the write fell to its `else` branch — paging back through history *deleted* a mark the tail load had just recorded, rather than re-stamping it. The consequence is the one the finding names: a load that says nothing about the live turn was deciding what the tab believes about it. The write now sits inside the same `replayBefore === undefined` guard as the rest of the tail-only bookkeeping.

**Residual window (🟢).** Not taken, and I want to be precise about why rather than let it read as accepted-and-forgotten. Reading a positive turn-end signal at settle time means a network round trip inside the timer — a new failure mode (what does a failed query decide?) and a hook that can no longer be reasoned about synchronously. The exposure it buys down is narrow: a turn has to begin elsewhere within three seconds of a load that reported none in flight, and it costs one bubble split plus a briefly idle row, self-correcting on the next update. I would rather carry the window than the round trip, but this is a judgement call and I will implement it if you would rather have it closed.

**Getter/command pair (🟢, still present).** Unchanged, same reason as last round: the effect tests the mark before the delay and consumes it after, so a single `consume` call would spend the mark that the check is only reading.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `35e73c0`

## PR #3621: fix(ui): an old session no longer opens looking like it is still running
**Author:** xjacka | **Branch:** fix/session-opens-looking-running → main | **Changes:** +54 −2 (2 files)

### Summary
The new commit moves the idle-mark write and delete inside the existing tail-only guard, so a history page load no longer re-arms the 3-second window. _1 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: 46ad7a4 (2026-09-09T12:14:46Z) — verdict COMMENT

- ✅ **Fixed:** idle mark re-stamped by history page loads (`packages/ui/src/modules/sessions/hooks/use-acp-connection.ts:378`)
- 🔁 **Still present:** residual window: a turn begun elsewhere can settle early (`packages/ui/src/modules/sessions/hooks/use-acp-session.ts:108`)
- 🔁 **Still present:** port exports a mutable–ref getter/command pair (`packages/ui/src/modules/sessions/hooks/use-acp-connection.ts:519`)
- 🆕 **New:** turn metadata read outside the tail–only guard (`packages/ui/src/modules/sessions/hooks/use-acp-connection.ts:374`)
- 🆕 **New:** reset() leaves the idle mark, stale generation skips the delete (`packages/ui/src/modules/sessions/hooks/use-acp-connection.ts:498`)

### Findings
- 🟢 **Suggestion:** turn metadata read outside the tail–only guard (`packages/ui/src/modules/sessions/hooks/use-acp-connection.ts:374`) — see **TypeScript Engineering Review**
- 🟢 **Suggestion:** reset() leaves the idle mark, stale generation skips the delete (`packages/ui/src/modules/sessions/hooks/use-acp-connection.ts:498`) — see **TypeScript Engineering Review**

✅ No findings.

### Issue Fit
🔁 The one item this skill reported is the residual already listed above: the mark expires 3 s after the load, so a replay that arrives later still leaves the bubble streaming. The `## Fix` section states that bound and the turn-begun-elsewhere case in its own bullets, so this stays a suggestion (see review at `46ad7a4`).

- 🟢 **Suggestion:** `settleReplay({ turnInFlight })` and `appendInterruptedNotice(...)` read the same `turn` metadata without the tail-only guard the idle mark now has, so a history page can still act on it. Both lines are identical on `origin/main` and outside this PR's diff, so this is a pre-existing item for a separate change (`packages/ui/src/modules/sessions/hooks/use-acp-connection.ts:374`).
- 🟢 **Suggestion:** `reset()` clears the other session-scoped refs but not `idleSessionsRef`, and the move also ends the unconditional delete on a stale-generation load. Each generation bump on the same session is followed by a fresh tail load that re-stamps or deletes the mark, so the exposure stays inside the 3-second window (`packages/ui/src/modules/sessions/hooks/use-acp-connection.ts:498`).

### Verdict
APPROVE — the blocking finding of the previous round is fixed at the write site, and no new blocking finding survives verification.


---

PR #3630, opened 2026-09-09
chore(agents): bump built-in harness pins, disable codex update check

JanPokorny:
## Summary

Bumps the built-in harness images to the latest stable releases and turns off the remaining startup update check.

| Image | Package | Before | After |
|---|---|---|---|
| claude-code | `@agentclientprotocol/claude-agent-acp` | 0.66.0 | 0.75.1 (Claude Code 2.1.257) |
| codex | `@openai/codex` | 0.133.0 | 0.153.4 |
| codex | `@zed-industries/codex-acp` | 0.14.0 | 0.16.0 |
| pi-agent | `@earendil-works/pi-coding-agent` | 0.85.0 | 0.85.1 |

- Codex system config gains `check_for_update_on_startup = false`. Claude Code (`DISABLE_AUTOUPDATER=1`) and pi (`PI_SKIP_VERSION_CHECK=1`) already had their checks disabled.
- pi-dynamic-providers devDependency, lockfile and `minimumReleaseAgeExclude` entries moved to 0.85.1 in lockstep.

## Verification

- All three images build via `mise run //packages/agents:image claude-code codex pi-agent`; `mise ls` inside each image shows the new versions.
- `check:lockfile-age` and `check:comment-types` pass; the pi extension type-checks against 0.85.1.
- Not run: e2e suite against a cluster.


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `31ed228`

## PR #3630: chore(agents): bump built-in harness pins, disable codex update check
**Author:** JanPokorny | **Branch:** chore/update-harness-versions#locki-40gaa33g → main | **Changes:** +42 −62 (7 files)

### Summary
The PR moves the three built-in harness image pins (`claude-agent-acp` 0.66.0 → 0.75.1, `@openai/codex` 0.133.0 → 0.153.4, `codex-acp` 0.14.0 → 0.16.0, `pi-coding-agent` 0.85.0 → 0.85.1) and adds `check_for_update_on_startup = false` to the codex system config. The pi extension devDependency, the lockfile and the `minimumReleaseAgeExclude` list move with it.

### Findings
- 🟢 **Suggestion:** Two comments still describe the old pin state. `artifact-touch.test.ts` names 0.66.0 as "the version pinned in `packages/agents/claude-code/harness-tools.toml`" and instructs "Re-capture both when that pin moves", so it now asserts a pin that no longer exists and asks for work this PR did not do; and the exclude block still says "the fork publishes all its packages in lockstep" above a list from which `@earendil-works/pi-client` and `@earendil-works/pi-protocol` are gone — correct, because 0.85.1 no longer pulls them, but the comment no longer explains the list under it (`packages/agent-runtime/src/__tests__/unit/artifact-touch.test.ts:8`, `pnpm-workspace.yaml:15`).
- ✅ **Looks good:** `check_for_update_on_startup = false` is a real `ConfigToml` field of the pinned CLI, not a silently-ignored key. It is present in the `@openai/codex@0.153.4` binary's field list, and `codex --version` runs clean with this exact `system-config.toml` as `CODEX_HOME/config.toml`; the same binary accepts an invented key without complaint, so the knob had to be confirmed rather than assumed.

✅ No findings.

### Issue Fit
Its two items are documentation and PR-body drift over a correct tree, so they are folded into the single 🟢 in `### Findings`. The stale-fixture item was reported as blocking and is regraded: the adapter jump does not change the shape that fixture asserts. All four symbols `harness-history-lib.mjs` deep-imports from `dist/acp-agent.js` are still exported at 0.75.1 with a compatible `toAcpNotifications` signature, `sdk.mjs` still ships in the bundled `claude-agent-sdk@0.3.257`, and the one `rawOutput` change in 0.75.1 gates only on `_meta.terminal_output`, which terminal-shaped tool results carry and the MCP result this fixture captures does not.

### Verdict
APPROVE — the four pins, the lockfile and the exclude list are consistent, the new codex key is real at the pinned version, and the only residual is stale prose.


---

PR #3631, opened 2026-09-09
fix(security): set CSP and cache-control response headers

JanPokorny:
## What

Fix the two actionable findings from the authenticated ZAP scan of the dev deployment, and teach the scan's UI crawl to log in so those findings reflect the real app.

### Security headers (`fix(security)`)

- **CSP.** The app shell was served with no `Content-Security-Policy`. nginx now sends a navigation-only policy in both the Helm ConfigMap and the image `default.conf`:

  ```
  Content-Security-Policy: frame-ancestors 'none'; base-uri 'self'; form-action 'self'; object-src 'none'
  ```

  Fetch directives (`script-src`, `connect-src`, `img-src`, ...) are left out on purpose. The `srcdoc` iframes that preview artifacts and files render arbitrary HTML and inherit the parent policy, and the SPA calls Keycloak and Google Fonts cross-origin, so any fetch restriction breaks one of them. This is the tradeoff `docs/architecture/artifact-library.md` already records. It also replaces `default.conf`'s stricter CSP, which would have blocked both the Keycloak token call and the previews had that config ever served.

- **Cache-Control.** api-server now defaults `Cache-Control: no-cache, no-store, must-revalidate` unless a route sets its own (brand icons keep their 5-minute cache); the redundant per-route `no-store` on the tRPC handler and the public-agent route is removed. nginx serves the shell `no-store` and content-hashed `/assets/` `immutable`.

### Scan auth (`chore(zap)`)

The AJAX spider was crawling the unauthenticated shell: under the default Strict scope check its browser blocked the out-of-scope Keycloak redirect, so login never happened. `scopeCheck: Flexible` lets the OIDC redirect through; `logoutAvoidance` and an auth-report job are added so a run proves it reaches the logged-in session.

## Why the CSP is safe here

- Rule 10038 (CSP header not set) is satisfied by the header being present.
- Rule 10055 does not alert on a missing `default-src`, and its "no fallback" alert only fires on a wildcard in `frame-ancestors`/`form-action`. This policy has no wildcards.

## Verification

- Replayed the live dev app through a proxy with the header injected, logged in as the scan user, browsed four routes: zero CSP violations, `srcdoc` inline scripts still run, and a positive control confirmed enforcement.
- `//packages/api-server:check` and the full api-server suite (1349 tests) pass, including a new `security-headers.test.ts`.
- `//helm:check:render` and `//helm:check:lint` pass.

Not verified by a live scan: that needs the Helm change deployed to the dev cluster, which this branch does not do.

## Notdone

An earlier draft filtered ZAP's "Suspicious Comments" false positive in the scan plan; removed at the reviewer's request, so that informational alert will still appear.

### DIFF
--- a/packages/api-server/src/apps/api-server/app.ts
+++ b/packages/api-server/src/apps/api-server/app.ts
@@ -31,8 +31,10 @@ import {
 
 export type { ApiServerDeps } from "./deps.js";
 
-const securityHeaders: MiddlewareHandler = async (c, next) => {
+export const securityHeaders: MiddlewareHandler = async (c, next) => {
   await next();
+  if (c.res.status !== 304 && !c.res.headers.has("Cache-Control"))
+    c.header("Cache-Control", "no-cache, no-store, must-revalidate");
   c.header("X-Content-Type-Options", "nosniff");
   c.header("X-Frame-Options", "DENY");
   c.header("Referrer-Policy", "no-referrer");
--- a/helm/templates/ui/app.yaml
+++ b/helm/templates/ui/app.yaml
@@ -16,6 +16,13 @@ data:
         ''      '';
     }
 
+    # The shell (index.html) must always revalidate so a deploy is picked up on
+    # the next load; /assets/ files are content-hashed, so they never change.
+    map $uri $static_cache_control {
+        default     "no-cache, no-store, must-revalidate";
+        ~^/assets/  "public, max-age=31536000, immutable";
+    }
+
     # Structured JSON access log, one object per line on stdout.
     log_format json_access escape=json
         '{'
@@ -80,6 +87,11 @@ data:
             add_header X-Frame-Options "DENY" always;
             add_header Referrer-Policy "no-referrer" always;
             add_header Strict-Transport-Security "max-age=31536000; includeSubDomains" always;
+            # Navigation-only CSP. Fetch directives (script-src, img-src, ...) are
+            # inherited by the srcdoc iframes that preview artifacts and files, and
+            # those render arbitrary HTML, so any fetch restriction here breaks them.
+            add_header Content-Security-Policy "frame-ancestors 'none'; base-uri 'self'; form-action 'self'; object-src 'none'" always;
+            add_header Cache-Control $static_cache_control always;
             try_files $uri $uri/ /index.html;
         }
     }
--- a/meta/zap/crawl.yaml
+++ b/meta/zap/crawl.yaml
@@ -14,6 +14,7 @@ env:
           loginPageUrl: https://dam-dev.apps.dam-cl.fmaas.res.ibm.com/
           loginPageWait: 8
           browserId: chrome-headless
+          diagnostics: true
         verification:
           method: autodetect
       sessionManagement:
@@ -39,7 +40,15 @@ jobs:
       maxDuration: 15
       maxCrawlDepth: 10
       runOnlyIfModern: false
+      # Keycloak is out of scope; Strict blocks the OIDC redirect in the spider's browser, so login never happens.
+      scopeCheck: Flexible
+      logoutAvoidance: true
   - type: passiveScan-wait
+  - type: report
+    parameters:
+      template: auth-report-json
+      reportDir: ${ZAP_REPORT_DIR}
+      reportFile: crawl-auth
   - type: report
     parameters:
       template: traditional-html-plus
--- a/packages/api-server/src/__tests__/unit/public-agent-routes.test.ts
+++ b/packages/api-server/src/__tests__/unit/public-agent-routes.test.ts
@@ -55,14 +55,4 @@ describe("public agent routes", () => {
     expect(res.status).toBe(200);
     expect(await res.json()).toEqual({ agent: null });
   });
-
-  /**
-   * TEST_SCENARIO: An agent can be renamed or deleted at any moment, and the
-   * answer names a person. A cached copy in a shared proxy would outlive both.
-   */
-  it("tells caches not to store the answer", async () => {
-    const res = await ROUTES.request("/agents/agent-1");
-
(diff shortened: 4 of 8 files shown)


packages/api-server/src/apps/api-server/app.ts
```diff
@@ -31,8 +31,10 @@ import {
 
 export type { ApiServerDeps } from "./deps.js";
 
-const securityHeaders: MiddlewareHandler = async (c, next) => {
+export const securityHeaders: MiddlewareHandler = async (c, next) => {
   await next();
+  if (!c.res.headers.has("Cache-Control"))
```
dam-code-guardian: 🟡 **Warning:** the default runs after `await next()` on every response, including the bodyless `304`s the brand-icon routes return: `brand.ts:70` answers a matching `if-none-match` with `c.body(null, 304)` *before* it reaches `c.header("Cache-Control", "public, max-age=300")`, so the 304 has no `Cache-Control` and this middleware stamps `no-cache, no-store, must-revalidate` on it. Probed against hono 4.13.7: `200 → "public, max-age=300"`, `304 → "no-cache, no-store, must-revalidate"`. A 304 telling the client not to store contradicts the 200 that told it to cache for five minutes and undoes the ETag revalidation these routes implement; neither new test covers a conditional request.
  **Fix:** a blanket response-header default must not contradict a policy the route already chose for the same resource — skip the `Cache-Control` default on `304`/`412` (no-body conditional) responses, so a route's revalidation contract survives the middleware.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `1229757`

## PR #3631: fix(security): set CSP and cache-control response headers
**Author:** JanPokorny | **Branch:** fix/ui-security-headers → main | **Changes:** +76 −17 (8 files)

### Summary
Narrows the UI's nginx CSP to navigation directives only (fetch directives are inherited by the artifact-preview `srcdoc` iframes) and adds a `$uri`-keyed `Cache-Control` map; on the api-server, centralizes `Cache-Control` into the `securityHeaders` middleware as a `no-store` default and drops the per-route and tRPC copies; widens the ZAP crawl plan's scope check so the OIDC redirect is followed. One 🟡. _5 suggestion(s) dropped under the 🟢 budget._

### Findings
- 🟡 **Warning:** the `Cache-Control` default overwrites the cache policy of conditional `304` responses — `packages/api-server/src/apps/api-server/app.ts:36` (also `packages/api-server/src/apps/api-server/routes/brand.ts:71`, `routes/brand.ts:90`)
- 🟢 **Suggestion:** the share-host gate short-circuits ahead of `securityHeaders` (`app.use("*", deps.shareHostGate)` returns `viewer.fetch(...)` without `next()`), so with `Cache-Control` now owned by the middleware, share-host responses leave with none — and `viewer-app.ts` sets no `Cache-Control` of its own (`packages/api-server/src/apps/api-server/app.ts:71`). Reported as blocking by the TypeScript skill; the gate ordering predates this PR and those responses carried no `Cache-Control` before it either, so nothing regresses.
  **Fix:** apply a cross-cutting response guarantee where no request path can bypass it — register a host gate that skips `next()` after the header middleware, or apply the same middleware inside the branch app.
- ✅ **Looks good:** the ZAP plan keys are real and do take effect. `scopeCheck` and `logoutAvoidance` are genuine `AjaxSpiderJob.Parameters` fields, and their absence from `applyParameters`' explicit name array is not a gap — that array is `applyObjectToObject`'s *ignore* list (it holds `context`/`url`, which are not spider options), so both keys land on the `AjaxSpiderParam` instance `runJob` passes via `setOptions(...)`. The comment's claim that `Strict` blocks the OIDC redirect is consistent with `inScopeOnly` defaulting to `TRUE`.

### Issue Fit
_2 suggestion(s) reported, dropped under the 🟢 budget (undeclared `diagnostics: true`; the deleted public-agent cache test has no replacement covering the real route)._

_3 suggestion(s) reported after the share-host finding was merged into `### Findings`, dropped under the 🟢 budget (the replacement test exercises a synthetic app; `securityHeaders` exported for testability; the policy string duplicated in four places)._

### Verdict
COMMENT — the CSP narrowing and the ZAP plan keys verify clean, but the centralized `Cache-Control` default reaches the `304` branch of the brand-icon routes and replaces their own `public, max-age=300` policy with `no-store`.


- `packages/api-server/src/apps/api-server/app.ts:71` — 🟢 **Suggestion:** `shareHostGate` returns `viewer.fetch(c.req.raw)` without calling `next()`, so share-host responses never reach `securityHeaders`. That ordering predates this PR and `viewer-app.ts` sets no `Cache-Control` of its own, so nothing regresses — but now that `Cache-Control` is owned by the middleware, the invariant the new spec states ("every api-server response leaves with the baseline security headers") isn't true of the wiring.
    **Fix:** apply a cross-cutting response guarantee where no request path can bypass it — register a host gate that skips `next()` after the header middleware, or apply the same middleware inside the branch app.


---

JanPokorny:
Addressed in d4d4e658.

**🟡 Warning — no-store default overwrites 304 cache policy.** Fixed. `securityHeaders` now skips the `Cache-Control` default when the response status is 304, so the brand-icon routes' conditional branch no longer tells a revalidating client to drop its cached copy. Added a test scenario covering it.

**🟢 Share-host gate bypasses the middleware.** Left as-is: the gate short-circuits before `securityHeaders` and predates this PR, and `viewer-app.ts` sends its own CSP, so nothing regresses. Rather than reorder the gate here (which would also restamp the viewer's headers), I narrowed the new test's overview so it no longer claims to cover every api-server response — only those passing through the middleware.

The other dropped suggestions (the synthetic test app, exporting `securityHeaders`, the policy string in four places, `diagnostics: true`) I'm leaving for this PR.

Issue #3526, opened 2026-09-01
UI - An agent's status tag does not say whether the agent is doing work


jenna-winkler:
## Context

The UI describes a pod-up agent with two different vocabularies. The status tag on an agent says *Running* in green. The compute panel on Home splits the same agents into *Working* (green) and *Awake* (blue), by whether a session is active. Raised by design while managing several agents at once.

## Problem

One agent in one state gets two different answers.

- **Two words, one state** — the tag says "Running" while compute says "Working" for the same agent.
- **A distinction that exists in only one place** — compute knows the difference between an agent doing work and an agent idle but holding compute. The tag does not: an idle awake agent still reads green "Running".
- **Green means two things** — "doing work" in compute, "pod is up" on the tag.

The tag is the fastest thing to read, so it is where a user looks first. It cannot answer "what is my agent doing right now" or "which agents are safe to stop" — and that second question has a cost: a user who wants compute back has no way to tell which agents they can hibernate without interrupting work.

## Goal

Status reads the same everywhere — one word and one color per state — and a user can tell at a glance whether an agent is doing work, is idle but holding compute, or is off.

## User Stories

- As a user, I want an agent's status tag to say whether it is doing work right now so that I know whether to wait or to step in.
- As a user, I want to find agents that are idle but holding compute so that I can hibernate them and free that compute.
- As a user, I want to know that hibernating an agent will not interrupt work in progress so that I can stop it with confidence.
- As a user, I want the same state named and colored the same on every screen so that I do not learn two vocabularies.

## Scope

In: the shared status tag, the states it shows, the compute panel labels, the color of each state, and grouping in the agents list. Out: when an agent hibernates, and the underlying lifecycle states — this is about how existing states are named, colored, and grouped.

## Proposed solution

Three user-facing states, one word and one color each, used identically on the tag and in the compute panel:

- **Working** — green — doing a task.
- **Idle** — blue — not doing a task. Still holds compute, and will time out and hibernate unless it is always-on or has a scheduled task.
- **Hibernating** — grey — off, holds no compute.

**Idle** rather than **Awake**: it says the agent is holding resources and doing nothing, which is the fact a user acts on. **Running** retires as a user-facing word — it is the ambiguous one.

Group the agents list by Working and Idle, so which agents are safe to hibernate is visible without reading each card.

The tag and the compute panel should read their labels and colors from one shared definition, so the two surfaces cannot drift apart again.

## Open Questions

[...]

PR #3634, opened 2026-09-09
feat(ui): always-on lifecycle option made discoverable

### LINKED ISSUE #3477: UI - the never-hibernate setting is undiscoverable
## Context

A sandbox hibernates after a set number of idle minutes and wakes on the next message. Setting the timeout to 0 makes it never hibernate. That control sits on Agent Setup, under a Lifecycle heading, below the network access section.

A user in a research call had never found it. Every interaction with their sandbox began with a startup wait, for weeks, while a setting that removes the wait sat one page away. Product agreed the control is too hidden.

Design is now settled and dev ready. See **Proposed solution** for what to build. The feature is named **Always-on**.

## Problem

The setting that decides whether you wait for your sandbox is placed where nobody looks for it. Agent Setup is a long page, and Lifecycle is near the bottom, after provider, network access, and other setup fields. A user scanning for something about speed or waiting finds nothing — the words on screen are "hibernation timeout".

Two costs follow:

- **The wait looks like the product being slow.** A user who does not know the setting exists reads repeated startup delays as DAM's performance, not as a configuration they chose by default.
- **The people who most need it are least likely to find it.** Background work with no open session is exactly what 0 is for, and that user has no session open to prompt them toward the setting.

The startup wait is one of the first things a new user feels, and the fix for it is invisible.

## Goal

A user who is tired of waiting for their sandbox finds the never-hibernate option without being told it exists. A user setting up background work meets the option at the point it matters.

## User Stories

- As a user, I want to find the never-hibernate option when the startup wait annoys me so that I can stop waiting without asking someone.
- As a user, I want the option explained in terms of what I get so that I can tell whether it applies to me.
- As a user, I want to know what never-hibernate costs me so that I can decide against it if it holds compute I need elsewhere.
- As a user setting up unattended work, I want the option offered where I set that work up so that my agent is awake when it should run.

## Scope

**In:** the Lifecycle section in the agent creation flow and on the Configure page; the wording and default of the two lifecycle options; the always-on offer on the startup screen; the always-on marker on the agent card status tag; the always-on marker in the Compute Resources graph hover.

**Out:** the hibernation mechanics. The power menu wording (#3059). The redesign of the status tag itself (#3526) — this issue only adds the always-on marker to that tag.

## Proposed solution

Settled in design and dev ready. Frames are in the [dev ready Figma file](https://www.figma.com/design/zNIYydUKN1QLZDYozpQJpn/DAM-DEV?node-id=2251-2). An interactive prototype is on branch `worktree-feat+never-hibernate-discovery`.

The choice moves from a numeric timeout to a named option, and it is surfaced in four places.

- **Lifecycle section** — appears both in the agent creation flow and on the agent Configure page. It offers two options:
  - **Hibernate while idle** — the default. An inactivity window of 60 minutes, which the user can edit.
  - **Always-on** — the agent stays on. It stops only when the user hibernates it manually, or changes the Lifecycle setting again. Nothing else turns it off.
- **Skip the wait, on the startup screen** — while a user waits for an agent to start, the screen offers **Skip the wait — keep always-on**. Choosing it sets the agent to Always-on, so the agent stays on once it has started.
- **Always-on marker on the agent card** — an always-on agent gets a **power icon** on its Working or Idle status tag. Icon, not the text "Always-on": design tried the text and found it too heavy in a tag that small. The tag itself is being reworked in #3526, so this marker must land on the new tag, not the current one.
- **Always-on in the Compute Resources graph** — hovering an agent in the graph shows whether that agent is always-on, and links through to that agent's Configure page to change it.

## Open Questions

The design answers the earlier questions on this issue: it is both a placement and a wording problem, the control becomes a named option instead of 0-means-never, the choice does belong in the create flow, and the startup screen does offer it.

The icon-or-text question on the status tag is also settled: a power icon, not text.

These are still open:

- **How much of the compute trade-off do we show, and where?** Always-on holds compute against the user's budget ceiling. The Compute Resources graph shows the consequence after the fact. Nothing shows the cost at the point of choice.
- **Can anything but the user end always-on?** The design says only deliberate user action turns it off. If a budget ceiling can still hibernate an always-on agent, the user needs to be told, and the wording of the option cannot promise more than the system delivers.

## Additional resources

- [Figma — Always-on, dev ready](https://www.figma.com/design/zNIYydUKN1QLZDYozpQJpn/DAM-DEV?node-id=2251-2)
- [Figma — earlier frames](https://www.figma.com/design/zNIYydUKN1QLZDYozpQJpn/DAM-DEV?node-id=2251-449)
- Interactive prototype — branch `worktree-feat+never-hibernate-discovery`
- [Design discussion in Slack](https://ibm-research.slack.com/archives/C0B8DA70GE4/p1788364199059489)
- #3059 — UI: Sandbox power menu wording does not convey what the options actually do
- #3270 — UI: Change agent settings without leaving chat
- #3526 — UI: An agent's status tag does not say whether the agent is doing work

### LINKED ISSUE #3541: UI - The lifecycle choice does not explain the effect on compute use and response time
## Context

Agent Setup has a Lifecycle section. Today it holds one numeric field for idle minutes before hibernation. The new design in #3477 replaces it with two named options: Hibernate while idle, with an editable inactivity window, and Always-on. The same section appears in the agent creation flow.

The design work moved the choice to a place users can find. The words in it were not reviewed. A design review of the new screens raised the copy as a separate problem.

Source discussion: https://ibm-research.slack.com/archives/C0B8DA70GE4/p1788359611144889

## Problem

The two option names describe the mechanism. They do not describe what the user gets or gives up. Two effects decide the choice, and the screen states neither:

- Compute use. An always-on agent holds its CPU and memory against the user budget ceiling. Less room stays for other agents. A user near the ceiling can block a start later and not know why.
- Response time. A hibernated agent must start before it answers. The first message after an idle period waits. The inactivity window sets how often that wait happens.

A user who does not know how hibernation works cannot predict either effect. So the user keeps the default, or picks by guess. Both outcomes cost the user later: an unwanted wait, or compute held for an agent that does no work.

The word Always-on also sounds free. Nothing on the screen shows the price.

## Goal

At the moment of choice, a user can tell what each option does to the wait before a reply, and to the compute available for other agents. The user needs no knowledge of hibernation to make the choice.

## User Stories

- As a user, I want each lifecycle option to state what it costs me so that I can choose without guessing.
- As a user, I want to know that always-on holds compute so that I can tell why another agent will not start.
- As a user, I want to know that hibernation adds a wait to my next message so that the delay does not surprise me.
- As a user, I want the inactivity window explained in terms of my own waiting so that I can set a value with confidence.

## Scope

In: the names, the descriptions, and the helper text of both lifecycle options, in agent creation and in agent settings. The label of the inactivity window.

Out: the hibernation mechanics and the default values. The placement and discovery work (#3477). The status tag wording (#3526). The sandbox power menu wording (#3059).

## Open Questions

- Is Always-on the right name? It states the state, not the benefit.
- Do we show real numbers, such as the compute held or a typical start time? Or is plain prose enough?
- Does the choice need live budget headroom next to it? That overlaps #3101.
- Does the same wording work in the creation flow, where the user does not yet know how the agent will run?

## Additional resources

- #3477 - UI: the never-hibernate setting is undiscoverable. Holds the Figma designs and the prototype branch for the two options.
- #3526 - UI: An agent status tag does not say whether the agent is doing work.
- #3059 - UI: Sandbox power menu wording does not convey what the options actually do.
- #3101 - Epic: Size/budget experience.

---

_Filed by @spdempsey_

kapetr:
The never-hibernate option becomes a named, findable choice. The Lifecycle setting turns from a numeric timeout into two options — Hibernate when idle (editable inactivity window) and Always on — offered on the Configure page and in the agent creation flow. The startup wait screen offers "Skip the wait — keep always-on", and the Compute Resources hover shows which agents are always-on with a link to change it.

The option copy states the trade-off of each choice and shows the compute an always-on agent reserves, which is what #3541 asks for.

The status tag carries the always-on marker, and a running agent now reads Working or Idle instead of Running — on the agents list, on the sandbox home, and in the compute panel, which uses the same words.

Closes #3477
Closes #3541

Sub-issues

- [x] 01 — Lifecycle option cards on Configure
- [x] 02 — Lifecycle in the create flow
- [x] 03 — Skip the wait on the startup screen
- [x] 04 — Always-on hover in Compute Resources


dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `9116927`

## PR #3634: feat(ui): always-on lifecycle option made discoverable
**Author:** kapetr | **Branch:** feat/3477-always-on → main | **Changes:** +384 −85 (14 files)

### Summary
The numeric hibernation timeout becomes a two-option Lifecycle choice (Hibernate when idle / Always on), offered on Configure and in the create flow, with a "Skip the wait" button on the startup screen and an always-on hover card in Compute Resources. The sentinel round-trip is correct end to end; the findings are about what the create flow now sends, two stale statements on the lifecycle page, and two behaviours in the new controls. _21 suggestion(s) dropped under the 🟢 budget._

### Findings
- 🟡 **Warning:** the create flow now always sends an explicit idle window, so a template default and the cluster default no longer reach a new agent (`packages/ui/src/modules/agents/lib/create-agent-input.ts:84`, also `packages/ui/src/modules/sandboxes/hooks/use-setup-form.ts:18`)
- 🟡 **Warning:** `docs/architecture/agent-lifecycle.md:207` still says the setting is "a minutes field showing that *effective* value" and that `0` is how a user disables hibernation. This PR removes the minutes-only field and removes `0` from the input, and no architecture page changed in this PR (`Last verified: 2026-09-08`).
  **Fix:** when a PR changes how a documented setting is expressed or resolved, update that subsystem page and its `Last verified` stamp in the same PR (`docs/guidelines/documentation-guidelines.md:61`).
- 🟡 **Warning:** the body uses `Closes #3477` and `Closes #3541`, and two items those issues list under Scope → In are not in the diff: the always-on marker on the status tag, and the label of the inactivity window. The marker's stated successor #3526 is closed, so on merge both trackers are closed and the marker has none. The inactivity label keeps its pre-existing text, and the explanation that the agent "wakes on the next message" was deleted with the old field (`packages/ui/src/components/status-indicator.tsx:38`, also `packages/ui/src/modules/sandboxes/components/lifecycle-field.tsx:92`).
  **Fix:** link an issue with `Part of #<n>` while any of its in-scope items is outstanding, and use a closing keyword only when every named item is delivered.
- 🟡 **Warning:** the hover card cannot be dismissed with the keyboard, because the controlled Radix root has no `onOpenChange` (`packages/ui/src/modules/budgets/components/slot-bar.tsx:86`)
- 🟡 **Warning:** a typed `0` in the minutes input discards the window the user had set (`packages/ui/src/modules/sandboxes/components/lifecycle-field.tsx:82`)
- 🟢 **Suggestion:** the hover card prints the size by hand (`1 CPU · 2 Gi`) while the same segment's accessible name comes from `formatSizeLabel` and carries the multiplier too (`2x · 1 CPU · 2 Gi`), so one fact has two renderings that can drift (`packages/ui/src/modules/budgets/components/compute-usage.tsx:52`).
- ✅ **Looks good:** the `0` sentinel round-trips end to end. `minutesToDuration(0)` writes `"0s"`, `durationToMinutes` reads it back as `0`, `resolveEffectiveHibernationTimeoutMin` returns the override, and the controller's `shouldRun` treats `<= 0` as never hibernate — so "Always on" reports the behaviour the platform gives it.

Both blocking findings are in `### Findings`: the inheritance chain the page documents, and the page's stale description of the control. Its remaining suggestion (the slot meter is now an entry point, `docs/architecture/budgets.md:50`) fell under the 🟢 budget.

### Issue Fit
Both blocking findings are merged into one entry in `### Findings` — the closing keywords against the two undelivered in-scope items. Its remaining suggestion (every held segment, not only an always-on one, is now activatable) fell under the 🟢 budget.

All 7 findings were verified as non-blocking and regraded 🟢, so they fall under this review's 🟢 budget. Three of the four it reported as warnings restate a pattern `main` already carries (the hand-mirrored `CreateAgentInput`, the inline `=== 0` sentinel, the field-level `.catch`), and the fourth (`NaN` from a cleared input) is gated by `isCodingAgentSetupComplete`, which I confirmed blocks creation.

Two findings survive as 🟡 and are in `### Findings` (the Escape dismissal and the typed `0`). Its strongest suggestion is in `### Findings` too. Five of its warnings were regraded 🟢: the always-on fact is already in the trigger's accessible name, the `Manage` button duplicates the trigger's own Enter action, the create flow gates a cleared input the same way it gates every other incomplete field, the option cards stay operable with Tab and Space, and the one-click always-on offer with that wording is what #3477 specifies.

### Verdict
COMMENT — the sentinel and the controller path are correct, and five bounded issues remain: the create flow's new explicit default, two stale lifecycle-page statements, the closing keywords, and two behaviours in the new controls.

### For the human reviewer
The page that documents this feature says an agent with no timeout of its own follows the platform default, and that a workload image may ship its own default which only an explicit user choice overrides. This PR makes every agent created in the UI carry its own 60-minute window, whether or not the user chose one, so the platform default becomes unreachable from the main create screen and later changes to it never reach these agents. Settle whether the create screen should send a value only when the user changes it, and update the lifecycle page to state whichever order you choose. The same decision drives the pull request's closing links: two items the linked issues list as in scope are not here, and one of them has no other open ticket left.


---

packages/ui/src/modules/agents/lib/create-agent-input.ts
```diff
@@ -78,6 +81,7 @@ export function buildCodingAgentSetupInput(
   return {
     name: draft.name.trim(),
     egressPreset: "trusted",
+    hibernationTimeoutMin: draft.hibernationTimeoutMin,
```
dam-code-guardian, line 84: 🟡 **Warning:** every create now sends an explicit `hibernationTimeoutMin`, because the setup form defaults it to 60 (`use-setup-form.ts:18`). `agents-service.create` writes `spec.hibernationTimeout` whenever the field is present, so it overwrites the value `assembleSpecFromTemplate` copied from the template, and an agent with no override no longer exists. Two effects follow: on an install whose `controller.agent.base.idleTimeout` is not 1h (`"0s"` disables hibernation) a new agent still hibernates after 60 idle minutes, and a later change of that install default never reaches an agent created here. `docs/architecture/agent-lifecycle.md:207` states the intended order — unset inherits the cluster default, a template may seed the override, and "a user's explicit choice at create time still wins". A defaulted 60 is not an explicit choice.
  **Fix:** send `hibernationTimeoutMin` only when the user changes the Lifecycle option; the field is optional in `agentCreateInputSchema`, so omitting it keeps the documented template-then-cluster resolution for every create surface.

packages/ui/src/modules/sandboxes/hooks/use-setup-form.ts
```diff
@@ -15,6 +15,7 @@ export const setupFormSchema = z.object({
   templateId: z.string().nullable().default(null),
   kbTemplateId: knowledgeBaseTemplateIdSchema.nullable().default(null),
   customImage: z.string().default(""),
+  hibernationTimeoutMin: z.number().int().nonnegative().default(60).catch(60),
```
dam-code-guardian, line 18: 🟡 **Warning:** this `default(60)` is the source of the explicit value the create flow now always sends (see `create-agent-input.ts:84`). 60 matches today's shipped `idleTimeout: "1h"`, but it is a UI constant, so it silently replaces the install default on any install that changed that value, and it replaces a template's own never-hibernate default.
  **Fix:** send `hibernationTimeoutMin` only when the user changes the Lifecycle option; the field is optional in `agentCreateInputSchema`, so omitting it keeps the documented template-then-cluster resolution for every create surface.

packages/ui/src/modules/budgets/components/slot-bar.tsx
```diff
@@ -26,35 +79,60 @@ export function SlotBar({ segments, totalSlots, label, ariaLabel }: Props) {
...
         const held = segment.state !== "available";
+        const id = segment.agentId ?? "available";
         return (
-          <Tooltip
-            key={segment.agentId ?? "available"}
-            content={label(segment)}
-            side="top"
+          <TooltipPrimitive.Root
+            key={id}
+            open={hover.hoveredId === id}
```
dam-code-guardian, line 86: 🟡 **Warning:** the root is controlled by `open` with no `onOpenChange`, so Radix cannot close it: Escape and pointer-down do nothing, and the card hides only on pointer leave or blur. The previous uncontrolled `Tooltip` closed on Escape, so hover content that a keyboard user opens with focus is now undismissable while focus stays put.
  **Fix:** when you take control of a Radix `open` prop, wire `onOpenChange` back into the same state, so the primitive's keyboard and outside dismissal keep working.

packages/ui/src/modules/sandboxes/components/lifecycle-field.tsx
```diff
@@ -0,0 +1,155 @@
...
+            type="number"
+            min={1}
+            step={1}
+            className="w-28"
+            disabled={disabled}
+            value={Number.isNaN(value) ? "" : value}
+            onChange={(event) => {
+              const next = event.target.valueAsNumber;
+              if (next === 0) {
+                setLastMinutes(DEFAULT_HIBERNATE_MIN);
```
dam-code-guardian, line 82: 🟡 **Warning:** this path sets `lastMinutes` to `DEFAULT_HIBERNATE_MIN`, so a typed `0` throws away the window the user had. With the field at 30, typing `0` selects Always on, and choosing Hibernate when idle again returns 60, not 30. `pickAlwaysOn` keeps the value (`value > 0` → `setLastMinutes(value)`), so the two paths into the same state disagree. The input also declares `min={1}`, so the value that changes the mode is one the control marks invalid.
  **Fix:** remember the current value on every path into Always on — set `lastMinutes` from `value` here as `pickAlwaysOn` does, and let only the option cards change the mode.

```suggestion
                if (value > 0) setLastMinutes(value);
```

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `eb30d7f`

## PR #3634: feat(ui): always-on lifecycle option made discoverable
**Author:** kapetr | **Branch:** feat/3477-always-on → main | **Changes:** +409 −87 (15 files)

### Summary
This round answers four of the five open findings at their cause: the create flow now omits the idle window until the user chooses one, the lifecycle page describes the two named options, the compute hover closes on Escape, and a typed `0` keeps the previous window. Two items are open — the Escape fix also closes the hover card before the pointer can reach its `Manage` link, and the create screen shows a window it does not send. _16 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: 9116927 (2026-09-10T13:00:07Z) — verdict COMMENT — unreachable, reviewed the whole PR

- ✅ **Fixed:** create flow always sends an explicit idle window (`packages/ui/src/modules/agents/lib/create-agent-input.ts:84`)
- ✅ **Fixed:** lifecycle page describes the retired minutes field (`docs/architecture/agent-lifecycle.md:207`)
- ✅ **Fixed:** typed 0 discards the remembered idle window (`packages/ui/src/modules/sandboxes/components/lifecycle-field.tsx:82`)
- ✅ **Fixed:** controlled tooltip cannot be dismissed with Escape (`packages/ui/src/modules/budgets/components/slot-bar.tsx:86`)
- 🔁 **Still present:** closing keywords with two in–scope items undelivered (`packages/ui/src/components/status-indicator.tsx:38`)
- 🔁 **Still present:** hover card and accessible name render the size twice (`packages/ui/src/modules/budgets/components/compute-usage.tsx:52`)
- 🆕 **New:** the hover card closes before the pointer reaches `Manage` (`packages/ui/src/modules/budgets/components/slot-bar.tsx:88`)
- 🆕 **New:** the create screen shows 60 minutes and sends nothing (`packages/ui/src/modules/sandboxes/components/setup/setup-sections.tsx:121`)

### Findings
- 🟡 **Warning:** the hover card closes before the pointer reaches `Manage` (`packages/ui/src/modules/budgets/components/slot-bar.tsx:88`, also `packages/ui/src/modules/budgets/components/slot-bar.tsx:94` and `packages/ui/src/modules/budgets/components/compute-usage.tsx:60`)
  **Fix:** keep tooltip content that holds a control reachable by pointer — remove `disableHoverableContent` when the user must move into the content, and give every `role="button"` element an `onClick` that the key handler delegates to.
- 🟡 **Warning:** the create screen shows 60 minutes of inactivity and sends no window (`packages/ui/src/modules/sandboxes/components/setup/setup-sections.tsx:121`, also `packages/ui/src/modules/sandboxes/components/lifecycle-field.tsx:34`)
  **Fix:** a control that stands in for an inherited value shows the value the request will carry, or an explicit inherited state — never a constant of its own; and a selection handler commits its own option on every click.

🔁 The lifecycle page finding from the previous review is fixed. Two items remain, both about pages this PR does not change: the rewritten timeout paragraph names the settings screen and the create screen but not the startup overlay, which also writes the window; and `docs/architecture/budgets.md:50` still describes the slot meter as a read-only display. They are one documentation item and fell under the 🟢 budget.

### Issue Fit
🔁 1 finding from the previous review still present (see review at `9116927`): `Closes #3477` and `Closes #3541` with two named in-scope items undelivered. Re-verified this round — #3526, the stated successor of the status-tag marker, is closed since 2026-09-09, so on merge the marker has no open tracker. The skill also read the create screen's shown-but-unsent window against #3541's goal; that is the second finding above.

The two blocking items are the create screen's shown-but-unsent window and the no-op `Hibernate when idle` click, both in the second finding above. Seven further items — the duplicated completeness predicate, `.catch(null)` on the persisted draft field, `NaN` from a cleared input, the inert `alwaysOn` flag on the filler segment, the exported `BYTES_PER_MI`, the duplicated memory formatter and the hand-maintained `CreateAgentInput` — carry no reachable defect and fell under the 🟢 budget.

Its two blocking items are the first finding above: the card closes on pointer-leave, and the segment carries `role="button"` with no `onClick`. Seven further items — the `aria-label` on the card container, `lastMinutes` seeded from a prop, the radio-group keyboard contract, the `min={1}` input accepting `0`, the startup overlay's label, the duplicated store selector and the interactive styling on segments with no `Manage` link — are on code this round does not touch and fell under the 🟢 budget.

### Verdict
COMMENT — the round closes four findings at their cause; two warnings are open, one of them introduced by this round's Escape fix.

### For the human reviewer
The round fixes what the last review asked for. Two decisions are worth settling before the single fixes. The compute bar now offers a link to change an agent's lifecycle, but that link can only be used with the keyboard — the card that holds it closes as soon as the mouse moves toward it, and clicking the bar itself does nothing. And on the creation screen the lifecycle choice is shown as a concrete idle window, while the request deliberately carries no window at all, so a new agent can start with a different window than the screen showed, including one that never hibernates. Both are about what the screen promises versus what the platform then does.


---

packages/ui/src/modules/budgets/components/slot-bar.tsx
```diff
@@ -26,35 +79,63 @@ export function SlotBar({ segments, totalSlots, label, ariaLabel }: Props) {
...
+            onOpenChange={(next) => {
+              if (!next) hover.hide();
```
dam-code-guardian: 🟡 **Warning:** this dismissal fix also closes the card on every pointer-leave. `disableHoverableContent` makes Radix call its own close path from the trigger's `onPointerLeave` (`tooltip.tsx` → `onTriggerLeave` → `handleClose` → `setOpen(false)`), and because `open` is controlled that runs `onOpenChange(false)` → `hover.hide()`, which clears the timers at once. `CLOSE_GRACE_MS` and the `onPointerEnter`/`onPointerLeave` handlers on the content never run, so the pointer cannot cross the 4 px gap into the card, and the `Manage` link in it is unreachable with a mouse.

**Fix:** keep tooltip content that holds a control reachable by pointer — remove `disableHoverableContent` when the user must move into the content, and give every `role="button"` element an `onClick` that the key handler delegates to.

packages/ui/src/modules/budgets/components/slot-bar.tsx
```diff
@@ -26,35 +79,63 @@ export function SlotBar({ segments, totalSlots, label, ariaLabel }: Props) {
...
+            }}
+            disableHoverableContent
           >
-            <span
-              role="img"
+            <TooltipPrimitive.Trigger asChild>
+              <span
+                role={held && onActivate ? "button" : "img"}
```
dam-code-guardian, line 115: 🟡 **Warning:** same finding, second location. The segment reports itself as a button and only `onKeyDown` activates it, so a mouse click on a held segment does nothing. With the card closing on pointer-leave (line 88) this leaves no pointer path to an agent's setup page.

**Fix:** keep tooltip content that holds a control reachable by pointer — remove `disableHoverableContent` when the user must move into the content, and give every `role="button"` element an `onClick` that the key handler delegates to.

packages/ui/src/modules/budgets/components/compute-usage.tsx
```diff
@@ -37,7 +40,36 @@ function segmentLabel(segment: ComputeSegment, unit: SlotUnit): string {
...
+        <span className="font-semibold">{segment.agentName}</span> (
+        {formatCores(segment.cpuMilli)} CPU ·{" "}
+        {formatGi(segment.memoryMi * BYTES_PER_MI)} Gi)
+      </p>
+      {segment.alwaysOn && (
+        <>
+          <p className="text-muted-foreground">
+            Always on — holds compute even while idle.
+          </p>
+          <button
```
dam-code-guardian, line 60: 🟡 **Warning:** same finding, third location. This `Manage` button is the only pointer-reachable control the always-on card offers, and it lives inside tooltip content — portalled, not a focus target, and dismissed on interaction — so no input modality reaches it once the card closes on pointer-leave.

**Fix:** keep tooltip content that holds a control reachable by pointer — remove `disableHoverableContent` when the user must move into the content, and give every `role="button"` element an `onClick` that the key handler delegates to.

packages/ui/src/modules/sandboxes/components/setup/setup-sections.tsx
```diff
@@ -101,3 +103,26 @@ export function ConnectionsSetupSection({
...
+  value: number | null;
+  onChange: (next: number) => void;
+  sizeMi?: SizeMi;
+}) {
+  return (
+    <section className="mb-8">
+      <SectionLabel spaced>Lifecycle</SectionLabel>
+      <Inset>
+        <LifecycleField
+          value={value ?? DEFAULT_HIBERNATE_MIN}
```
dam-code-guardian, line 121: 🟡 **Warning:** the screen shows a window the request does not carry. With the draft field `null` the card renders `Hibernate when idle` with 60 minutes, while `buildCodingAgentSetupInput` omits `hibernationTimeoutMin`, so `agents-service.ts` keeps the template's seeded `hibernationTimeout` or the cluster default — which `docs/architecture/agent-lifecycle.md:207` says a Template may seed as `0`, i.e. always on. A user who accepts the shown 60 minutes can get an agent that never hibernates and holds compute.

**Fix:** a control that stands in for an inherited value shows the value the request will carry, or an explicit inherited state — never a constant of its own; and a selection handler commits its own option on every click.

packages/ui/src/modules/sandboxes/components/lifecycle-field.tsx
```diff
@@ -0,0 +1,155 @@
...
+  sizeMi,
+  error,
+  disabled,
+}: Props) {
+  const alwaysOn = value === 0;
+  const [lastMinutes, setLastMinutes] = useState(
+    value > 0 ? value : DEFAULT_HIBERNATE_MIN,
+  );
+
+  const pickHibernate = () => {
```
dam-code-guardian, line 34: 🟡 **Warning:** same finding, second location. `pickHibernate` only calls `onChange` when the field is already on `Always on`, so in the create flow — where the value starts `null` and renders as not-always-on — clicking the shown `Hibernate when idle` card is a no-op and the draft stays `null`. The only way to commit the displayed 60 minutes is to retype the number, or to detour through `Always on` and back.

**Fix:** a control that stands in for an inherited value shows the value the request will carry, or an explicit inherited state — never a constant of its own; and a selection handler commits its own option on every click.

PetrBulanek (approved):
The body defers the always-on marker on the status tag to #3526, but #3526 was closed on 9 September as completed, and the resolution recorded on it is a design one, the working/awake vocabulary. No code landed for it: `status-indicator.tsx` on main still renders "Running" for a pod-up agent, so the reworked tag this marker was meant to sit on does not exist yet.

#3477 lists that marker under Scope → In, so merging with `Closes #3477` closes the last issue tracking it, and I could not find another open one that covers it. Could we either keep #3477 linked with `Part of`, or file the follow-up before this merges?

packages/ui/src/modules/agents/components/sandbox-list.tsx
```diff
@@ -30,6 +32,7 @@ export function SandboxList({
         <AgentRow
           key={agent.id}
           {...rowProps(agent)}
+          working={workingAgentIds?.has(agent.id)}
```
dam-code-guardian: 🟡 **Warning:** this collapses the badge's three-valued contract into two. `StatusBadge` splits `running` into Working and Idle only when `working !== undefined` (`status-indicator.tsx:49`), but `workingAgentIds` is always a built set, so `.has()` answers `false` for an agent whose session read has not resolved. Every running agent therefore reads **Idle** until its read lands, and permanently for an agent whose read failed, because `useFeed` sets `retry: false` and a failed agent never enters the set. `sandbox-home-header.tsx:36` passes the query's `undefined` straight through and gets this right.

**Fix:** keep an unknown tri-state `undefined` end to end — never let a `Set.has()`, `?? false` or truthiness test at a call site turn "still loading" or "read failed" into a definite negative that the UI then states as a fact.

packages/ui/src/modules/budgets/components/compute-usage.tsx
```diff
@@ -61,49 +95,70 @@ export function ComputeUsage({ agents, workingAgentIds }: Props) {
...
+              className="flex items-center justify-between gap-3 py-2"
+            >
+              <span className="flex items-center gap-1.5">
+                <span
+                  className={cn(
+                    "inline-block size-2 shrink-0 rounded-full",
+                    STATE_DOT[group.state],
+                  )}
+                />
+                {STATE_LABEL[group.state]}
```
dam-code-guardian, line 153: 🟡 **Warning:** same finding, second location. These per-state rows count agents from the same `workingAgentIds` set, so while the session reads are pending the card reports every held agent under **Idle**. The count a user reads to decide which agents are safe to hibernate is the one that is wrong first.

**Fix:** keep an unknown tri-state `undefined` end to end — never let a `Set.has()`, `?? false` or truthiness test at a call site turn "still loading" or "read failed" into a definite negative that the UI then states as a fact.

packages/ui/src/modules/home/api/queries.ts
```diff
@@ -20,6 +20,20 @@ export const homeKeys = {
     [...acpSessionsKeys.agentLists(agentId), "home"] as const,
 };
 
+export function useAgentWorking(
+  agentId: string,
+  enabled: boolean,
+): boolean | undefined {
+  const { data } = useQuery({
```
dam-code-guardian, line 47: 🟡 **Warning:** this copies the session query of `useFeed` — same key, same `staleTime`, same `retry: false` — and drops its freshness policy. There is no `refetchInterval`, so an agent without `features.liveUpdates` is never re-polled and a failed read is never retried. Its only consumer is the agent's own header (`sandbox-home-header.tsx:36`), and that screen does not mount `usePodSessionsWatch`, which is mounted only in `home-view.tsx:52`. So the new Working/Idle badge can hold a stale value for as long as the page stays open, and after one failed read it falls back to "Running" and stays there.

**Fix:** one query key has one freshness policy — define the options for `homeKeys.sessions` once, in a shared `sessionsQueryOptions(agent)` factory, and have every observer use it, so no consumer can subscribe to that key without the polling and invalidation the key needs.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `1b8fc53`

## PR #3634: feat(ui): always-on lifecycle option made discoverable
**Author:** kapetr | **Branch:** feat/3477-always-on → main | **Changes:** +582 −136 (30 files)

### Summary
This round closes the compute hover's pointer path — the card now keeps a grace period and a held segment activates on click — and it makes the `Hibernate when idle` card commit its own option. It also adds the always-on marker and a Working/Idle split to the status tag, which #3477 lists as out of scope and the PR body still describes as deferred. _24 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: eb30d7f (2026-09-14T08:29:35Z) — verdict COMMENT

- 🔁 **Still present:** closing keywords do not match the delivered scope (`packages/ui/src/components/status-indicator.tsx:49`)
- 🔁 **Still present:** create screen shows 60 minutes and sends nothing (`packages/ui/src/modules/sandboxes/components/setup/setup-sections.tsx:121`)
- 🔁 **Still present:** Manage stays unreachable by keyboard (`packages/ui/src/modules/budgets/components/slot-bar.tsx:120`)
- 🔁 **Still present:** hover card and accessible name render the size twice (`packages/ui/src/modules/budgets/components/compute-usage.tsx:52`)
- 🆕 **New:** a still–unread session reads as Idle (`packages/ui/src/modules/agents/components/sandbox-list.tsx:35`)
- 🆕 **New:** the new badge holds a stale value on the agent page (`packages/ui/src/modules/home/api/queries.ts:27`)
- 🆕 **New:** budgets page describes the old slot meter (`docs/architecture/budgets.md:50`)

### Findings
- 🟡 **Warning:** a still–unread session reads as Idle (`packages/ui/src/modules/agents/components/sandbox-list.tsx:35`, also `packages/ui/src/modules/budgets/components/compute-usage.tsx:153`)
- 🟡 **Warning:** the new badge holds a stale value on the agent page (`packages/ui/src/modules/home/api/queries.ts:27`)

- 🟢 **Suggestion:** this PR changes the compute surface that `docs/architecture/budgets.md` owns, and that page is not updated. Line 50 still names the meter colors `working / awake / free` and puts the budget-request link "beside" the meter, while the code now labels the held groups Working and Idle, marks always-on segments, and gives each held segment a card with a route into that agent's settings. Its `Last verified:` stamp still reads 2026-09-07 (`docs/architecture/budgets.md:50`)

The skill's two other items are the same documentation item and fell under the 🟢 budget: the new create-screen clause on the lifecycle page states the request shape instead of the invariant, and no architecture page yet names the Working/Idle split this PR introduces.

### Issue Fit
🔁 1 finding from the previous review still present (see review at `eb30d7f`) — re-verified this round, and the evidence changed. The always-on marker is now delivered, so that half is closed. Three of the skill's items are the same finding at new locations: #3477 lists "The redesign of the status tag itself (#3526)" under **Out** and says the marker "must land on the new tag, not the current one", and this round ships that redesign — Working/Idle on the tag and in the compute panel — while the PR body still reads "The always-on marker on the status tag is out of scope here". #3526 is closed as completed, so no open issue tracks the rest of it. Separately, #3541 lists "The label of the inactivity window" as **In**: the window still reads "minutes of inactivity", and the deleted `hibernation-timeout-field.tsx` helper text ("Idle minutes before the agent hibernates… **0** = never") has no replacement, so that named item closes less explained than before. The skill's three remaining items are undeclared but harmless changes and fell under the 🟢 budget.

🔁 Its finding on the new session hook is the second new finding above; the skill and this review reached it independently, and its remedy — one freshness policy per query key, in a shared options factory — is the **Fix:** carried there. Its two other blocking items do not clear the severity bar and are 🟢 under the budget: the `home` → `sandboxes` import repeats a cross-module import `coding-agents-view.tsx` already makes on `main`, and the `icon` field that `AlertOptions` now inherits has no caller, since the only `showAlert` call site passes `kind` and `confirmLabel` only. Three further suggestions fell under the budget.

🔁 Its findings on the create screen's unsent window and on the `Manage` control are the two carryovers above; it adds that the create screen's real outcome is the deployment's own `agentIdleTimeoutMinutes`, and that a tooltip cannot hold a focusable control at all. Its tri-state finding is the first new finding above. One item does not survive verification and is 🟢: "Skip the wait — keep always-on" is the wording #3477 settled, and that issue states the effect the button has ("so the agent stays on once it has started"), so the copy promises what it delivers. Nine further items — the radio-group ARIA contract, the mount-seeded previous window, `NaN` from a cleared input, the absolute footer, and five formatting and naming duplications — carry no reachable defect and fell under the 🟢 budget.

### Verdict
COMMENT — the round closes the hover card's pointer path at its cause; four warnings stay open, two of them introduced by this round's status-tag work.

### For the human reviewer
The round fixes what the last review asked for on the compute hover. One decision is worth settling before the single fixes, and a reviewer has already asked about it: this change now delivers the reworked status wording that a different, already-closed issue owns, while the description still says that work is deferred. If it merges with the closing keywords as they are, the issue that tracks the always-on marker closes, and the remainder of the tag rework has no open issue behind it. The second thing to settle is what the new Working and Idle labels say while the platform is still reading an agent's sessions: today an agent that is busy can read as idle, which is the exact question the labels exist to answer.


---

kapetr:
@dam-code-guardian Both new findings are fixed. What is left alone:

Manage unreachable by keyboard — the slot itself is focusable and Enter runs the same navigation, so the action is reachable; the button inside the tooltip stays mouse-only because a tooltip can't hold focus.

Create screen shows 60 minutes and sends nothing — the screen can't show the real default until the api-server exposes it, so this gets a follow-up rather than a guess in the client.

Size rendered twice — the card is visual and the aria-label is its text equivalent; the repeat is deliberate.

Working/Idle in the compute meter — rejected; we're not adding an unknown state to the meter.

The PR body now describes the status-tag work that landed here.

dam-code-guardian:
@kapetr Thank you. I recorded four decisions as PR-local overrides for this PR, and later reviews will not raise them again:

1. `Manage` unreachable by keyboard. I verified the mechanism you give: the segment has `role="button"`, `tabIndex={0}` and an Enter/Space handler that calls the same `navigateToSandboxHome(agentId, "setup")` as the button in the tooltip. The action is reachable, so the finding is closed.
2. Create screen shows 60 minutes and sends nothing. Recorded as a descope until the api-server exposes the real default.
3. Size rendered twice. Recorded as intended: the card is visual and the accessible name is its text equivalent.
4. Working and Idle in the compute meter. Recorded as your design decision. The override covers the meter only. The tri-state read in `sandbox-list.tsx` stays in scope, because that one is about a pending read that becomes a fact.

The two fixes and the new PR body go into the re-review that runs now.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `1d89944`

### Summary
This round answers both findings of the last review at their cause: the status pill now carries a tri-state read, and the two observers of the session query share one options factory. The tooltip tail and the hover timers move to the design-system shapes, and both architecture pages are updated. _(Suppressed 4 finding(s) per PR-local overrides: create screen shows 60 minutes and sends nothing, Manage unreachable by keyboard, hover card and accessible name render the size twice, an unknown session state in the compute meter.)_ _16 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: 1b8fc53 (2026-09-14T11:37:05Z) — verdict COMMENT

- ✅ **Fixed:** a still–unread session reads as Idle (`packages/ui/src/modules/agents/components/sandbox-list.tsx:35`)
- ✅ **Fixed:** the new badge holds a stale value on the agent page (`packages/ui/src/modules/home/api/queries.ts:27`)
- ✅ **Fixed:** budgets page describes the old slot meter (`docs/architecture/budgets.md:50`)
- 🔁 **Still present:** closing keywords do not match the delivered scope (`packages/ui/src/components/status-indicator.tsx:49`)
- 🆕 **New:** the new doc sentences name affordances the code does not render (`docs/architecture/budgets.md:50`)

### Findings
- 🟢 **Suggestion:** The two sentences this round adds to the architecture pages name three things the code does not do. `budgets.md` says the meter carries "a marker on the slots of agents that never hibernate", but a cell takes its class from `SEGMENT_FILL[segment.state]` only, and `segment.alwaysOn` reaches the accessible name and the hover card alone. The same sentence says the hover names the agent's "state", and `HeldSegmentCard` prints the name and the size only. `agent-lifecycle.md` says an always-on agent "is marked as such wherever its status shows", and `agent-unavailable-overlay.tsx:119` renders `<StatusBadge state={state} />` with no `alwaysOn` (`docs/architecture/budgets.md:50`, also `docs/architecture/agent-lifecycle.md:207`)

🔁 Its three blocking items are the one documentation finding above, folded into a single entry: the per-slot always-on marker, the state in the hover card, and the "wherever its status shows" quantifier. One of the three needed correcting before it could be reported — the settings link the page promises does exist for every held slot, because `SlotBar` makes the segment itself activatable, so only the missing state survives verification. Its fourth item is the create screen's unapplied 60 minutes, which a recorded decision covers.

### Issue Fit
🔁 1 finding from the previous review still present (see review at `1b8fc53`) — re-verified at this HEAD, and one half of it is now closed. The PR body declares the status-tag work that lands here, so the undeclared-scope half is answered. The other half stands: `Closes #3541` remains, while that issue lists "the label of the inactivity window" as **In** and the window still reads `minutes of inactivity` with no replacement for the copy the deleted helper carried. The skill's three remaining items — the compute panel's retitle and muted link, the overlay typography, and a `?? []` fallback — are undeclared but carry no defect and fell under the 🟢 budget.

🟢 Its two blocking items do not clear the severity bar and are kept as suggestions under the budget. The first says the agent page's badge is read once at mount and never updates: the round did not introduce that, because `useAgentWorking` carried no `refetchInterval` before it either, and the query still refreshes on window focus and on every prefix-matching invalidation of `acpSessionsKeys.agentLists(agentId)` — so the round strictly improves the compat case and regresses nothing. The second reads the `UNIT_BOUNDARY_DESCRIPTION:` prefix on an interface field as a misuse of the registered comment type; the checker in `.mise/tasks/strip-comments` scores the prefix alone, so this is a guideline placement question, not a violated gate. Five further suggestions fell under the budget.

🟢 Its two remaining blocking items are suggestions under the budget: `SlotBar` drives `TooltipPrimitive` by hand where the design system already ships a hover card, and the new `tail` capability sits on the internal `TooltipContent` without reaching the exported wrapper. Both describe the shape of the code, and the timers, the grace period and the Escape path all behave. Its third blocking item — one page carrying two shapes of the same session fact, so the row reads Running and the meter reads Idle for an unread agent — is the compute-meter state model, which a recorded decision covers. Six further suggestions fell under the budget, the widest being that the Escape listener sits on `document` rather than on the open layer.

### Verdict
COMMENT — the round closes both of the last review's findings at their cause; one carryover stays open, and it is a closing keyword, not code.

### For the human reviewer
The code side of this change has converged: what the last review asked for is delivered, and the remaining items are either the author's recorded decisions or notes with no defect behind them. One thing is still worth settling before merge, and it is the same one as last time. The description now explains the extra status work that arrived here, which was the fair complaint. What has not changed is that merging closes both linked issues, and one of them still asks for a piece of wording that this branch does not deliver. Either deliver that wording or downgrade that one link, so nothing closes with an open item inside it.


---

PR #3636, opened 2026-09-09
feat(ui): a user sets how the chat sidebar's three panels divide the space

### LINKED ISSUE #3618: UI - A user cannot control how the chat sidebar's three panels divide the space
## Context

The chat sidebar stacks three panels: Sessions, Files and Artifacts. #3455 opened all three by default, so three panels now compete for one column instead of two. Review of that change found the sizing controls did not keep up.

## Problem

A user cannot decide how much room each panel gets.

- Only one boundary can be dragged, the one between Sessions and Files. Files and Artifacts always split whatever space is left in equal halves, and no control changes that.
- That single divider disappears when either Sessions or Files is collapsed. Collapse Sessions, and a user with Files and Artifacts open has nothing left to drag.
- The drag stops short. Sessions cannot grow past a fixed height, so a tall display keeps space the user cannot give to the panel they are reading.

The panel a user works in is rarely the one the layout favours. Someone reviewing a long file list, or a session with many artifacts, has to scroll a short panel while a neighbour holds space it does not need.

## Goal

A user sets how the sidebar splits between Sessions, Files and Artifacts, in any combination of open and collapsed panels, and the choice holds across reloads.

## User Stories

- As a user, I want to drag the boundary between any two open panels so that I can give the panel I am working in more room.
- As a user, I want to resize panels whichever ones are collapsed so that my control does not depend on the panels I happen to have open.
- As a user, I want a panel to grow as far as the window allows so that a tall display is fully usable.
- As a user, I want my sizes remembered so that I arrange the sidebar once.

## Scope

**In:** the three stacked panels inside the chat sidebar, their dividers, their size limits, and their sizes surviving a reload.

**Out:** the width of the sidebar column itself and the file viewer beside the chat. Both already resize.

## Proposed solution

Treat the sidebar as one set of panels rather than one sized panel plus leftovers. Put a divider at every boundary between two open panels, and give each open panel a share of the column. One rule then covers every combination: dividers appear and disappear with the panels they separate, and collapsing a panel never takes away a control the user was using. Let a panel grow until its neighbours reach their minimum, so the limit comes from the window rather than a fixed number.

## Open Questions

- After a user collapses a panel and opens it again, should it return to its earlier share or take an even split?
- When a panel collapses, should the others hold their proportions or share the freed space evenly?
- Sizes are remembered globally today, as the collapse choices are. Is global still right, or should they follow the agent?

## Additional resources

- Raised in review of #3566, which delivered #3455.

---

_Filed by @kapetr_

PetrBulanek:
A user sets how the chat sidebar splits its height between Sessions, Files and Artifacts, in
any combination of open and collapsed panels, and the choice survives a reload.

Today the column is one sized panel plus leftovers. Sessions holds a fixed height capped well
below a tall window, Files and Artifacts always split what is left in equal halves, and the one
divider disappears as soon as either Sessions or Files is collapsed — leaving a user with two
open panels and nothing to drag. The panel someone is actually reading is rarely the one the
layout favours.

This treats the column as one set of panels instead. A divider sits at every boundary between
two open panels, each open panel holds a share of the height, and a drag stops only when its
neighbour reaches its minimum. One rule then covers every combination: dividers come and go with
the panels they separate, collapsing never takes away a control someone was using, and a tall
display is fully usable.

Closes #3618

## Sub-issues

- [x] 01 — Weighted panel stack for the chat sidebar


packages/ui/src/modules/sessions/lib/sidebar-panels.ts
```diff
@@ -0,0 +1,55 @@
...
+    if (!panel.open) continue;
+    if (above !== null) pairs.push({ above, below: panel.id });
+    above = panel.id;
+  }
+  return pairs;
+}
+
+export function panelStyle(open: boolean, weight: number): CSSProperties {
+  return open
+    ? { flexGrow: weight, flexBasis: 0, minHeight: 0 }
```
dam-code-guardian, line 30: 🟡 **Warning:** `MIN_PANEL_PX` is read only inside `resizePair` (lines 47-48), so the minimum holds while a divider moves and nowhere else. `panelStyle` turns a weight into `flexGrow` with `minHeight: 0`, and every other moment that re-divides the column applies the weights unchanged — a panel opening again, or the weights read from storage at mount. Modelling the arithmetic on an 800px column: from an even split, drag the Sessions/Files divider until Files reaches 120px, collapse Artifacts, drag Sessions down to its 120px minimum, then open Artifacts again. Sessions settles at **85px**, 35px under the minimum, in a column with room for three panels of 120px.

  **Fix:** apply the minimum where shares become heights, not only in the drag: re-normalise the weights whenever the open set changes (and after the initial read from storage) so every open panel keeps `MIN_PANEL_PX` while the column has the room for it.

packages/ui/src/modules/sessions/hooks/use-sidebar-panels.ts
```diff
@@ -0,0 +1,130 @@
...
+    };
+    refCallbacks.current.set(id, callback);
+    return callback;
+  }, []);
+
+  const measure = useCallback((id: SidebarPanelId) => {
+    const element = elements.current.get(id);
+    return {
+      weight: weightsRef.current[id],
+      px: element ? element.getBoundingClientRect().height : 0,
```
dam-code-guardian, line 96: 🟡 **Warning:** reported by both the TypeScript and React reviews. `measure` reads the live `getBoundingClientRect().height` and `resize` adds one mousemove's incremental delta to it, so the drag keeps no running total. Two moves that share one React commit both read the same height, and the second overwrites the first: `above.px + delta` is computed from the same base twice, so only the last delta of the group survives. A pointer that reports more moves than React commits therefore leaves the divider behind the cursor for the rest of the drag. The replaced code added each delta into `sessionsHRef.current` and could not lose one.

  **Fix:** an interaction that receives incremental deltas keeps its running total in a ref, measured one time at drag start; the total is never re-derived from a layout that a pending render is about to change.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `2d9c799`

## PR #3636: feat(ui): a user sets how the chat sidebar's three panels divide the space
**Author:** PetrBulanek | **Branch:** feat/3618-sidebar-panel-sizing → main | **Changes:** +233 −45 (7 files)

### Summary
The PR replaces the sidebar's one fixed panel plus leftovers with a weighted stack: a new pure helper computes divider pairs, panel styles and the resize arithmetic, and a new hook holds the per-panel weights and persists them. Two warnings are about where the minimum panel height applies, and about the value the drag accumulates.

### Findings
- 🟡 **Warning:** the minimum panel height applies only inside a drag, so a share can fall below it (`packages/ui/src/modules/sessions/lib/sidebar-panels.ts:30`)
- 🟡 **Warning:** the drag total comes from the measured layout, not from an accumulator (`packages/ui/src/modules/sessions/hooks/use-sidebar-panels.ts:88`)
- ✅ **Looks good:** the four components take `ref` as a plain prop, which needs React 19; `packages/ui` resolves `react@19.2.7` in `pnpm-lock.yaml`, and each ref callback has a block body, so it returns nothing and React 19 does not read it as a cleanup function.

✅ No findings.

### Issue Fit
- 🟢 **Suggestion:** issue #3618 asks three open questions — the share a reopened panel takes, whether the other panels hold their proportions when one collapses, and whether sizes stay global or follow the agent. The PR answers all three in code and says so nowhere: the default weight gives an even split, `resizePair` leaves the other panels' weights alone, and the three storage keys are global. The answers are reasonable, so one line in the body closes the issue cleanly (`packages/ui/src/modules/sessions/lib/sidebar-panels.ts:15`).

- 🟡 The drag-accumulator warning is this skill's finding; it is merged into `### Findings` above (`packages/ui/src/modules/sessions/hooks/use-sidebar-panels.ts:88`).
- _5 suggestion(s) reported here and dropped under the review's 🟢 budget (see Summary): the per-move storage writes, the retired `platform-sessions-h` key, the unmemoized returned stack, the panel-list type, and the symmetry of `resizePair`'s return._

- 🟡 The drag-accumulator warning is this skill's finding too; it is merged into `### Findings` above (`packages/ui/src/modules/sessions/hooks/use-sidebar-panels.ts:88`).
- _4 suggestion(s) reported here and dropped under the review's 🟢 budget (see Summary): the per-move storage writes, the retired key, the shorthand `flex` against the longhand properties on a toggle, and the class-plus-custom-property convention for the constant parts of the style._

### Verdict
COMMENT — the feature works as described and the arithmetic is sound; two warnings are open, and neither is a wrong result.

### For the human reviewer
The PR stores each panel's size as a share of the column, not as a height, and it checks the minimum size only while a divider is being dragged. Every other moment that re-divides the column — opening a panel again, or loading a saved layout — applies the shares as they are, so a panel can end up smaller than the minimum the drag protects, in a column with room for all three. The drag also reads the current layout each time the pointer moves instead of keeping its own running total, so it can fall behind a fast pointer. What to settle first is where the minimum belongs: in the stored shares, or in the layout that turns shares into heights.


---

packages/ui/src/modules/sessions/hooks/use-sidebar-panels.ts
PetrBulanek: Confirmed and fixed in `f6526da` — you are right, and my first read of this was wrong.

I had measured several drags and seen them track the pointer exactly, so I initially believed the live measurement was safe because `getBoundingClientRect()` forces a layout flush. That argument does not hold: `getBoundingClientRect()` flushes pending layout for the *current* DOM, but it cannot flush a React state update that has not been committed yet, so two moves sharing a commit both read the pre-commit height. My drags had also all been synthesised with a single mousemove each, so none of them could ever have exercised this.

Dispatching three mousemoves in one synchronous turn against the deployed build made it unambiguous. Starting from an even split (376.7px each, so no clamp anywhere near), reading the persisted weight after each move:

| after move | measured | if accumulated | if last-delta-only |
|---|---|---|---|
| +20px | 1.05309 | 1.05309 | 1.05309 |
| +40px | **1.05309** | 1.10619 | 1.05309 |
| +60px | **1.05309** | 1.15928 | 1.05309 |

Byte-identical across all three. The reason is that `resizePair` preserves the pair's combined weight, so with `px` unchanged the result is a pure function of (base px, incremental delta) — identical inputs, identical output, and every delta but the last is discarded.

The fix follows your suggestion: a `DragOrigin` ref captures both panels' heights and weights once at the start of a drag, `resize` accumulates the incremental deltas into it, and `resizePair` is applied to (start heights, running total). `onDragEnd` clears it. Re-running the same probe now gives 1.05309 → 1.10618 → 1.15928, matching the accumulated column.

This also removes the two `getBoundingClientRect()` calls per mousemove, since the measurement now happens once per drag.

Two consequences worth recording:

- Measuring once at drag start means a column height change *during* a drag (for instance `RuntimeOutdatedNotice` appearing) is no longer picked up mid-gesture. That is the same tradeoff the replaced `sessionsHRef` code made, and it is the lesser problem.
- Because the accumulator retains over-travel, dragging past an end stop and reversing now has to unwind the excess before the panel moves — standard drag behaviour, and bounded by the clamp.

Re-verified against the deployed build after the fix: a 100px drag moves the panel exactly 100.0px with the third panel unmoved, the end stop still lands on exactly 120.0px, and the three panels still sum to the column height.

packages/ui/src/modules/sessions/lib/sidebar-panels.ts
PetrBulanek: Deliberate, and I would rather not change it — the proposed fix trades a cosmetic problem for a structural one.

You are right about the mechanics: `MIN_PANEL_PX` bounds a drag and nothing else, and `panelStyle` emits `minHeight: 0`, so a weight applied to a short column can render a panel below 120px.

The reason it is not enforced in the layout is that a weight is a *share*, not a height, and the share has to survive a column whose height the panel cannot control. Enforcing a 120px floor wherever shares become heights means three open panels demand 360px plus their headers; on a short window, or a narrow one where the runtime-outdated notice takes a slice of the column, the stack then exceeds the column and either overflows the sidebar or pushes content out of view. That is a worse failure than a short panel, and it would fire on window heights users really have.

As built, the degradation is contained rather than broken:

- A panel below its minimum still renders its 44px header and clips its body — `overflow-hidden` with `minHeight: 0` — so it stays legible and toggleable.
- It is self-correcting on the next drag. With a panel under 120px, `resizePair`'s clamp range forces the very next drag of that pair to lift it back to at least 120, so a user cannot get stuck with a panel they cannot recover.
- The invariant that does matter holds unconditionally: the open panels plus the collapsed headers always sum to exactly the column height. I checked this in every configuration — three open, each one collapsed in turn, one open, none open, and mid-drag — and it came out exact each time.

So the minimum is deliberately a *gesture* constraint, not a layout constraint: it stops a drag from destroying a panel, and the layout stays free to compress when the window demands it.

If the short-panel case is worth addressing on its own, the change I would reach for is not a hard floor but collapsing a panel that cannot make its minimum — which is a behaviour change the issue does not ask for, and worth its own decision rather than riding along here.

packages/ui/src/modules/sessions/hooks/use-sidebar-panels.ts
dam-code-guardian: Thank you for the measurement. I verified the fix at `9183e70`: `DragOrigin` measures both panels one time at drag start, `resize` adds each delta into `origin.delta`, and `resizePair` gets the start heights together with the running total. The warning is closed.

I recorded it as fixed for this PR, so a later review does not report it again. The two consequences you list are the expected result of one measurement per drag, and I do not raise them.

packages/ui/src/modules/sessions/lib/sidebar-panels.ts
dam-code-guardian: I hold your decline. The minimum stays a gesture constraint for this PR, and I recorded that here, so a later review does not raise it again.

I verified the mechanism at `9183e70`: the floor is in `resizePair` only, `panelStyle` keeps `minHeight: 0`, and the next drag of the pair lifts a short panel back to 120px when that pair has the space. A collapse for a panel that cannot make its minimum is a separate decision, as you say.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `9183e70`

## PR #3636: feat(ui): a user sets how the chat sidebar's three panels divide the space
**Author:** PetrBulanek | **Branch:** feat/3618-sidebar-panel-sizing → main | **Changes:** +257 −45 (7 files)

### Summary
The round fixes the drag accumulator: a `DragOrigin` ref measures both panels one time at drag start, and `resizePair` receives the running total. The branch was rebased, so the compare range to the last review is not reachable and this review reads the whole PR.
_(Suppressed 1 finding per PR-local overrides: the minimum panel height at `lib/sidebar-panels.ts:30`.)_
_6 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: `2d9c799` (2026-09-09T17:08:46Z) — verdict COMMENT — unreachable, reviewed the whole PR

- ✅ **Fixed:** drag total read from the layout, not accumulated (`packages/ui/src/modules/sessions/hooks/use-sidebar-panels.ts:88`)
- 🔁 **Still present:** issue #3618 open questions answered but undeclared (`packages/ui/src/modules/sessions/lib/sidebar-panels.ts:15`)
- 🆕 **New:** panel stack names the other two modules' panels (`packages/ui/src/modules/sessions/lib/sidebar-panels.ts:3`)
- 🆕 **New:** retired `platform-sessions-h` key left orphaned (`packages/ui/src/modules/sessions/views/chat-view.tsx:113`)

### Findings
_No new blocking findings at this HEAD. The suggestions are in the skill sections below._

✅ No findings.

### Issue Fit
- 🔁 1 finding from the previous review is still present (see review at `2d9c799`). The skill reported it as a warning this round. The severity bar keeps it a suggestion, because no code path gives a wrong result.

- 🟢 **Suggestion:** the panel stack is generic machinery, but it lives in the `sessions` module and names the other two modules. `SidebarPanelId` closes over `"files"` and `"artifacts"`, and the hook owns their storage keys. A third surface that needs a resizable stack must import from `sessions` or copy the logic. The package-level `src/lib/` and `src/hooks/` hold the comparable reusable parts, such as `persisted-prefs.ts` and `use-toggle-set.ts` (`packages/ui/src/modules/sessions/lib/sidebar-panels.ts:3`).
- _The skill reported this as a warning. It is regraded: placement is a suggestion on a working diff, and the sessions view already owned the other two panels' open state and the file-panel width before this PR. 1 suggestion reported here and dropped under the review's 🟢 budget (see Summary): the React import in an otherwise pure `lib/` file._

- 🟢 **Suggestion:** `platform-sessions-h` is retired with no migration and no cleanup. `origin/main` reads it at `chat-view.tsx:109`; at this HEAD no file under `packages/ui/src` reads it or removes it. Each user who set a sessions height gets the even split one time, and the dead key stays in storage. The old value is a height and the new value is a share, so a delete in the read path is the clean end (`packages/ui/src/modules/sessions/views/chat-view.tsx:113`).
- _Two more warnings from this skill are regraded to suggestions and dropped under the review's 🟢 budget (see Summary): the `weightsRef` mirror beside the `weights` state, which only a later writer can desynchronize because `resize` is today the single writer of both, and the per-move storage writes. 3 further suggestions dropped: the ref cache written during render, the unmemoized returned stack, and one shared layout-props type for the three panels._

### Verdict
APPROVE — the drag accumulator is fixed at `9183e70`, the minimum-height decision is declined and recorded, and no blocking finding is open.


---

kapetr (approved):

PR #3632, opened 2026-09-09
build(claude-code-vm): rebuild the containerDisk only when its source moved, pin build inputs

JanPokorny:
## What

The claude-code-vm containerDisk is rebuilt (~15 min) and republished with a new digest on every main push, even when nothing in its source changed. This gates it on the image resolver like the workloads, pins its chart tag by source sha, and pins the floating build inputs.

## Why

Two separate problems:

- **Every main push rebuilt it.** The job was gated on "claude-code was rebuilt", which on main is always true (`RESOLVE_NO_REUSE`). The `.build/` qcow2 cache is local only, so CI never hit it.
- **The same commit built different content over time.** `fedora-bootc:44` by tag, k3s resolved from the stable channel at build time, bib/skopeo `:latest`.

Bit-for-bit reproducibility of the qcow2 is not achievable: bootc-image-builder seeds its RNG from `crypto/rand` (`util.go` `createRand`) with no override, so partition GUIDs and fs UUIDs differ on every run, and dnf stamps install times into the rpmdb. So "deterministic" here means: an unchanged VM is not rebuilt and keeps its published image.

## How

- `image/resolve`: `claude-code-vm` registered with base `claude-code`; new `claude_code_vm` output (main pushes reuse unchanged, PRs build only when the effective source moved); `list-slow-builds` = workloads + VM; self-check extended.
- `cd.yml`: VM job gated on that output, tags the image with its source sha; `publish` and `container-scan` wait for `build-claude-code-vm`, so a v-chart never pins an unbuilt tag and a missing `<sha>` tag at scan time means reuse, not a race.
- `helm/values.yaml` + `pin-tags`: `@ci-pin claude-code-vm-image-tag` marker, consumed like the workload markers.
- Containerfile/build.sh: fedora-bootc, bib, skopeo pinned by digest; k3s by version. dnf packages stay floating on purpose (mirrors drop superseded NVRs within weeks).
- `docs/security/supply-chain.md`: reused VM images skip that run's Trivy scan, same as workloads.

## Verified

- `resolve --self-check` passes; PR-mode `gha-outputs` emits `claude_code_vm=true` (no registry image yet).
- `pin-tags` dry run on a copy of values.yaml replaced all 8 markers, no leftovers.
- Workflow YAML parses; `mise run check:comment-types` passes.
- Both `tag@digest` refs pull.

[...]

PR #3648, opened 2026-09-10
perf(ci): fix artifact cache saves, cache Go, ungate e2e, reuse agents on main

JanPokorny:
## What

Six CI changes, ordered by impact. Measured on runs 34382898967 (PR) and 34399939013 (main).

1. **Task-artifact cache was never saved by the check job.** `check` and `test` shared the key `mise-task-artifacts-<os>-<sha>`; the faster `test` job reserved it first and `check`'s post step failed with *Unable to reserve cache … another job may be creating this cache*. The restored cache was 16 KB and only tasks the test job runs ever hit (4 of ~60). Keys are now per job, with cross-job restore-keys.
2. **Go caches.** `~/.cache/go-build`, `~/.cache/staticcheck`, `~/go/pkg/mod` are cached in check and test; build, vet and staticcheck were each ~4 min of full recompiles and set the check job's wall clock.
3. **e2e no longer waits for check.** It downloads the digest artifacts `build-images` already uploads and pulls by digest, instead of waiting for `merge-images` (gated on `check`) to create the `:<sha>` tags. `build-mock` likewise takes the platform-base amd64 digest when the base was built in the run, instead of waiting for `merge-platform-base` (also gated on `check`). Expected PR critical path: ~17 min → ~12 min. Quay keeps untagged manifests long enough for the minutes between push and pull.
4. **Agent reuse on main pushes.** `RESOLVE_REUSE_WORKLOADS` → `RESOLVE_REUSE_AGENTS`: agents, workloads and the VM rebuild only when their effective source moved (#3632 already did this for the VM; this extends it to the five harness agents). On the sampled main run k-search (10 min, mostly exporting an unchanged multi-GB layer) and build-claude-code-vm (14 min) were the last 25 minutes.
5. **Playwright Chromium cache** (~30 s per run) and **no more runner disk cleanup**: measured on a PR run, the 145G root disk has 87G free before the step and the whole e2e run uses ~15G, so the 40–90 s rm bought nothing.
6. **Comment on `cache: false`** in the publish jobs (cache-poisoning boundary), so it is not flagged again.

## Behaviour change to be aware of

Harness agent templates (`claude-code`, `codex`, `pi-agent`, `bob`, `k-search`) now carry `@ci-pin` markers like the workloads and the VM, so published charts pin their image tags to the source sha. Since a template's rendered image ref is its version identity, existing agents are offered an upgrade **only when the image actually changed**, not on every chart release. Charts rendered from git are unchanged (`tag: ""` → appVersion). `pin-tags` fails closed if a marker is missing or unconsumed.

Not changed: `platform-base` and `keycloak` still rebuild on every main push; nightly/tag/dispatch still rebuild everything.

## Verified

- `mise run check:zizmor check:pinact check:resolve-image check:version check:comment-types //helm:check:lint //helm:check:render` pass; `image:resolve --self-check` passes; `pin-tags` against real `gha-outputs` consumes all 14 markers.
- First PR run (34445906857): the check job's post step now logs `Cache saved with key: mise-task-artifacts-Linux-check-…` and `go-Linux-check-…` (previously *Failed to save*); e2e pulled all four platform images by digest and passed.
- Second PR run (34447657322): `mise check` 5m41s → 1m20s (broad `restored result from cache`), `build mock` started alongside the image builds, e2e 9m17s with a Playwright cache hit.


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `cfa8bdd`

## PR #3648: perf(ci): fix artifact cache saves, cache Go, ungate e2e, reuse agents on main
**Author:** JanPokorny | **Branch:** perf/ci-caching-parallelism#locki-pt6lv6o3 → main | **Changes:** +140 −62 (5 files)

### Summary
The PR gives `check` and `test` their own mise task-artifact cache keys, caches the Go build, module and staticcheck directories, starts `e2e` from the `build-images` digest artifacts instead of the `check`-gated `:<sha>` tags, and extends source-sha reuse plus `@ci-pin` chart pinning from the workloads and the VM to the five harness agents.

### Findings
- 🟢 **Suggestion:** `tag_for` now resolves through `resolve_derived`, so on a `main` push `platform_base_tag` holds platform-base's source-sha tag while `platform_base` is still `true`, although the comment above says platform-base keeps the caller's mode (`.mise/tasks/image/resolve:129`). The outcome is correct — `merge-platform-base` re-points that same source-sha tag onto the fresh build, and `build-agents` and `build-mock` both wait for it — so this is about the comment reading as an exemption that `tag_for` does not give.
- ✅ **Looks good:** the new digest path in `e2e` holds together. `build-images` carries no `if:`, so all four components always run and each uploads exactly one digest file; `download-artifact` without `merge-multiple` keeps one directory per artifact, which is what `ls "/tmp/digests/digests-$c-amd64"` reads; `keycloak` is in that matrix and `mock` keeps its own `:<sha>` tag.

- 🟢 **Suggestion:** the "Template upgrade" section of the architecture page states that a template's resolved image ref is its version identity, but not what makes that ref move, so it does not carry the new behaviour a user sees — a release no longer marks every Agent as behind (`docs/architecture/agent-lifecycle.md:74`). One sentence on the source-sha pinning, with the `Last verified:` stamp bumped, closes it.

### Issue Fit
✅ No findings.

### Verdict
APPROVE — no blocking finding: the cache keys, the digest-based `e2e` gate and the agent reuse path each verify against the workflow they run in.


---

PR #3649, opened 2026-09-10
docs(postgres): grant usage_readers membership WITH INHERIT TRUE

jezekra1:
## Why

Setting up the analytics consumer on a managed PostgreSQL 18 instance today, the runbook was followed as written: `usage_readers` created, `CONNECT` granted, `GRANT usage_readers TO usage_reader`. The api-server's reconcile then reported every `usage_src_*` passthrough readable for the group, and the consumer was still `permission denied for view usage_src_*` on all of them.

The login had been created `NOINHERIT` as read-only hardening. On PostgreSQL 16+ a membership confers the group's privileges only when its `INHERIT` option is on, and a bare `GRANT` defaults that option to the member's own `INHERIT` attribute. So the login was a member (`pg_has_role(..., 'MEMBER')` true) that inherited nothing (`'USAGE'` false). Re-granting `WITH INHERIT TRUE` fixed it.

## What

Docs only. The api-server never grants membership, so nothing in code changes.

- `docs/notes/postgres-role-operations.md`: the managed-instance bullet now grants `WITH INHERIT TRUE`, explains the failure mode, notes the PG ≤15 equivalent, and gives the `pg_has_role(..., 'USAGE')` check that tells the two states apart.
- `helm/templates/postgres.yaml`: the same statement in the bootstrap SQL comment.

## Possible follow-up (not in this PR)

The reconcile could list members of `usage_readers` whose membership does not inherit and log them the way it logs `usage.grants.unreachable`. It would have surfaced this at api-server start instead of at the consumer.

### DIFF
--- a/docs/notes/postgres-role-operations.md
+++ b/docs/notes/postgres-role-operations.md
@@ -125,9 +125,18 @@ admin role:
   each owning its own database; `REVOKE CONNECT ON DATABASE … FROM PUBLIC` and
   grant it back only to the owner. This is portable SQL.
 - Create `usage_readers` as `NOLOGIN` (no password), and `GRANT CONNECT ON
-  DATABASE platform` to it. Then `GRANT usage_readers TO <read-only login>`
-  for whichever login should read the metrics — that membership is what
-  survives future view migrations. Order does not matter, but a restart does:
+  DATABASE platform` to it. Then `GRANT usage_readers TO <read-only login>
+  WITH INHERIT TRUE` for whichever login should read the metrics — that
+  membership is what survives future view migrations. The `WITH INHERIT TRUE`
+  matters: on PostgreSQL 16+ a membership confers the group's privileges only
+  when that option is on, and it defaults to the member's own `INHERIT`
+  attribute. A read-only login is often created `NOINHERIT` as hardening, and
+  then a bare `GRANT` records a membership that confers nothing — the
+  reconcile still reports every passthrough readable for the group, while the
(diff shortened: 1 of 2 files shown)


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `3217006`

## PR #3649: docs(postgres): grant usage_readers membership WITH INHERIT TRUE
**Author:** jezekra1 | **Branch:** docs/usage-readers-inherit → main | **Changes:** +16 −5 (2 files)

### Summary
The PR corrects the operator recipe for `usage_readers` membership: the grant now carries `WITH INHERIT TRUE`, and the runbook explains the silent failure a `NOINHERIT` member causes. Documentation only — the bootstrap SQL comment in the chart gets the same statement.

### Findings
- ✅ **Looks good:** The PostgreSQL claim holds. On 16 and later a role membership carries its own `INHERIT` option, the option defaults to the member role's `INHERIT` attribute, and `pg_has_role(login, 'usage_readers', 'USAGE')` is the predicate that separates a member that inherits from one that does not. On 15 and earlier the option does not exist, as the note says. Both places in the tree that show the `GRANT usage_readers TO <login>` statement are updated, so no bare-grant recipe is left (`docs/notes/postgres-role-operations.md:128`, `helm/templates/postgres.yaml:49`).

✅ No findings.

### Issue Fit
✅ No findings.

### Verdict
APPROVE — the change documents a real and reproduced failure mode correctly, and it changes no code.


---

PR #3650, opened 2026-09-10
feat(runtime-delivery): surface and re-deliver config an old runtime refuses

### LINKED ISSUE #3431: An agent silently misses configuration its runtime image is too old to accept

jjeliga:
An agent running an older image can't accept every kind of configuration the platform sends it, so those parts get dropped on delivery. Nothing recorded that, nothing told the owner, and updating the agent didn't recover them either.

**The owner is now told.** The agent shows a "Missing config" warning in the agent list and the chat header, and a notice in the chat view naming what its runtime refused in plain terms — "skills", "network access", "environment variables" — next to the existing Update button. The agent keeps running; a gap degrades it rather than disqualifying it.

**Updating the agent now actually delivers the missing parts.** When an agent reports a capability set different from the one on record, the platform re-sends its configuration. Nothing did this before: a delivery that dropped kinds still settled as complete, so after an update the agent stayed misconfigured until some unrelated change happened to trigger a re-send.

**That re-send survives failure.** It raises the agent's desired version in Postgres rather than only queueing a job, so a lost job is recovered by the periodic sweep, and a slow delivery built against the old capabilities can no longer overwrite a newer one's record of the gap.

**The warning stays current.** A change in what was dropped is announced as an agent change, so the badge appears and clears promptly instead of lingering until some unrelated refresh.

Also: migration `0045` adds one column to the delivery outbox table; the harness-configuration prose moves onto its own architecture page, because runtime-delivery had 26 characters of headroom under its size cap; the runtime-delivery page is corrected — it described a grant-time capability warning that was never actually built; and a shared test fixture replaces three per-suite copies.

Not done: the warning-freshness behaviour has no automated test, and none of this has been verified against a cluster or in a browser.

**Scope calls on the issue's open questions.** A capability gap warns but does not gate the agent's health — filtering exists precisely to tolerate an older image, and blocking would turn a degraded agent into an unavailable one. All six kinds carry the same warning rather than a severity tier. The gap is shown on the Agent, not on the grant, so the second user story is answered at kind granularity ("skills", "network access") rather than per grant; attributing a gap to a specific grant is not implemented.

Closes #3431


packages/api-server/src/modules/runtime-delivery/services/hello-handler.ts
```diff
@@ -46,7 +53,16 @@ export function createHelloHandler(deps: {
       }
 
       const row = await deps.outboxRepo.getRow(agentId);
-      if (row && row.version > (input.lastAppliedVersion ?? 0)) {
+      if (!row) return { events: [] };
+
+      let desiredVersion = row.version;
+      if (kindsChanged) {
+        desiredVersion = await deps.outboxRepo.bumpVersion(agentId);
```
dam-code-guardian: 🟡 **Warning:** `upsertHello` commits the new capabilities and returns the old ones in its own transaction, and the bump runs afterwards in a separate one. If anything between the two fails — `getRow` rejecting on a transient error, or the pod restarting — the next `hello` compares two identical sets and never bumps. The agent retries `hello`, and the retry cannot recover the signal either, because the write already landed. The row settled clean, so nothing else records that a re-delivery is owed: the refused kinds stay refused, and the owner keeps a warning that applying the update no longer clears.
  **Fix:** When a one-shot comparison is the only trigger for later work, arm the trigger in the same transaction that overwrites the compared value, or derive "a re-delivery is owed" from the recorded drop set against the newly advertised kinds, so any later `hello` re-detects an outstanding gap.

packages/ui/src/modules/agents/components/unsupported-contributions-badge.tsx
```diff
@@ -0,0 +1,20 @@
...
+export function UnsupportedContributionsBadge({
+  kinds,
+}: {
+  kinds: AgentView["unsupportedContributionKinds"];
+}) {
+  if (kinds.length === 0) return null;
+  return (
+    <Badge
+      variant="warning"
+      title={`This agent's runtime is too old to accept ${contributionKindList(kinds)} it was granted. Update the agent to apply them.`}
```
dam-code-guardian: 🟡 **Warning:** The tooltip always ends with "Update the agent to apply them", while `ContributionGapNotice` reads `agent.templateUpdate` and says "No runtime update is available yet" for the same state. `templateUpdate` is undefined whenever the agent already runs its template's current image, which is the steady state and the issue's own example (a Claude-Code image that never accepts `skill-ref`). The badge then asks for an action the user cannot take.
  **Fix:** Copy that names a remedy reads the state that gates the remedy — pass the update availability into the badge and select the sentence from it, rather than stating the remedy in a component that cannot see whether one exists.

docs/architecture/runtime-delivery.md
```diff
@@ -352,4 +354,4 @@ Capabilities also gate whole flows, not only payload items: `hello` carries a nu
...
-- **Capabilities are honored end-to-end.** A Contribution or Event kind not in the agent's advertised set is dropped at send time, never silently delivered. A grant that requires unsupported kinds succeeds with a UI warning; the unsupported parts simply don't appear in the agent's payload. A flow-gating capability (the knowledge-base publish level) fails visibly instead: the gated feature records an update-the-agent failure rather than dropping work silently.
+- **Capabilities are honored end-to-end, and never silently.** A Contribution or Event kind not in the agent's advertised set is dropped at send time, never silently delivered. A grant that requires unsupported kinds still succeeds and the unsupported parts simply don't appear in the agent's payload — but the Agent carries an owner-visible warning naming what its runtime refused, and keeps it until the runtime advertises those kinds. A flow-gating capability (the knowledge-base publish level) fails visibly instead: the gated feature records an update-the-agent failure rather than dropping work silently.
```
dam-code-guardian, line 355: 🟡 **Warning:** The invariant now says capabilities are honored "never silently" and that the Agent "carries an owner-visible warning naming what its runtime refused", after a sentence that covers Contribution **and** Event kinds. Only contribution kinds reach `dropped_contribution_kinds` and the UI; `payload.droppedEventKinds` reaches a log line and the undeliverable stamp, so no owner-visible warning names a refused event kind. Two more claims on these pages move past the code: the `hello` paragraph at `:117` says a dispatch follows when "the capability set it reports differs", while `advertisedKindsChanged` compares the kind arrays only, so a newly claimed harness-config surface, watch surface or `kbPublish` level triggers nothing. Both edited pages also carry a `Last verified:` header older than the commit's author date (2026-09-10): this page keeps 2026-09-09, `docs/architecture/agent-lifecycle.md` moved 2026-09-08 → 2026-09-09.
  **Fix:** State each claim at the precision the code implements — name the contribution kinds in the invariant and the advertised kind sets in the trigger sentence — and set `Last verified:` on every page a PR edits to that PR's own date.

dam-code-guardian (commented):
🛡️ **Code Guardian 🛡️** — ⚠️ Code Review @ `b8c1d0b`

## PR #3650: feat(runtime-delivery): surface and re-deliver config an old runtime refuses
**Author:** jjeliga | **Branch:** feat/3431-surface-capability-gap-to-owner → main | **Changes:** +3776 −59 (36 files)

### Summary
The delivery now records on the outbox row which Contribution kinds an old runtime image refused, shows them on the Agent, and re-delivers them when a `hello` reports a wider set of advertised kinds. The mechanism is complete end to end; the findings are the durability of the re-delivery trigger, one copy contradiction, and the pages that describe the result. _1 generated/lockfile file(s) not reviewed: packages/db/drizzle/meta/0044_snapshot.json._ _7 suggestion(s) dropped under the 🟢 budget._

### Findings
- 🟡 **Warning:** the capability transition is consumed outside the transaction that overwrites it (`packages/api-server/src/modules/runtime-delivery/services/hello-handler.ts:60`)
- 🟡 **Warning:** the badge tells the owner to update the agent, also when no update exists (`packages/ui/src/modules/agents/components/unsupported-contributions-badge.tsx:15`)
- 🟡 **Warning:** the page states an invariant and a trigger that are wider than the code (`docs/architecture/runtime-delivery.md:357`)
- 🟡 **Warning:** the PR is linked to close #3431, while the body and the commit say `Refs #3431`. `closingIssuesReferences` reports #3431, so a merge closes it. Two of the issue's open questions stay unanswered on the record: whether a gap gates the agent's health, and whether the grant surface also shows the gap. The second user story asks for the grant list.
  **Fix:** Remove the closing link, or record in the body which answer each open question gets and which story stays open.
- 🟢 **Suggestion:** `CONTRIBUTION_KIND_LABELS` is a `Record<string, string>` with a `?? kind` fallback, and `AgentView.unsupportedContributionKinds` is `string[]`. The map holds all six kinds today, so nothing leaks now. A `Record<ContributionKind, string>` keyed off the contract union makes a seventh kind a build error instead of a raw slug in owner-facing prose (`packages/ui/src/modules/agents/lib/contribution-kind-labels.ts:1`)
- ✅ **Looks good:** the version bump cannot loop on a stable runtime. `helloInput` requires `capabilities`, with required `contributions` and `events` arrays, so the stored value and the reported value converge after one hello and `advertisedKindsChanged` reports no further change.

- 🟡 The two claims and the two stamps are one finding — see `### Findings` above. The skill reported them as four items: the widened `hello` trigger sentence (`docs/architecture/runtime-delivery.md:117`), the widened invariant (`:357`), and the `Last verified:` headers of `docs/architecture/runtime-delivery.md` (unchanged at 2026-09-09) and `docs/architecture/agent-lifecycle.md` (2026-09-09, one day before the commit's author date).

### Issue Fit
- 🟡 The closing-link finding is in `### Findings` above.
- 🟢 The issue's first open question asks whether every kind deserves the same weight. All six kinds get the same warning badge, and no note records that the tiering idea was dropped. _(Dropped under the 🟢 budget.)_

- 🟡 The non-atomic transition finding is in `### Findings` above; the skill reached it independently, on the same evidence.
- 🟢 Five suggestions: the stored drop set is a delivery-time snapshot that a stopped agent keeps, the two new comparators duplicate each other, the notice clones `RuntimeOutdatedNotice`, the first `hello` bumps the version and resets the apply counters, and the new table omits the narrowed-kinds and first-`hello` rows. _(Dropped under the 🟢 budget.)_

- 🟡 The badge copy finding is in `### Findings` above.
- 🟢 The label-map typing suggestion is in `### Findings` above. Two more: the notice clones `RuntimeOutdatedNotice`, and on a phone the sessions column is hidden, so only the badge `title` carries the detail — the same placement as the existing notice. _(Dropped under the 🟢 budget.)_

### Verdict
COMMENT — the feature works on every path I read, and four warnings stand: one durability gap in the re-delivery trigger, one contradiction in the badge copy, the pages that over-state the result, and the issue link.

### For the human reviewer
The change makes the runtime's own report of what it can accept the single trigger for re-sending the parts an older image refused. That report is compared once, against the value it immediately replaces, so the signal exists for the length of one request and nothing durable records that a re-send is owed. When that one moment is lost, the gap stays on the Agent and applying the update no longer clears it. Settle whether the trigger should be the moment the capabilities change, or the recorded gap itself, before the smaller fixes matter.


---

dam-code-guardian:
🛡️ **Code Guardian 🛡️** — correction to the review at `b8c1d0b`

The fourth warning said the body and the commit declare `Refs #3431` while the link closes the issue. The body was edited during my review and now says `Closes #3431`, so that warning is wrong: the closure is declared. I withdraw it.

What is left of it is not a warning: the issue's open questions about health gating and about the grant surface are answered by this PR without a record. That is a scope call, so it is a suggestion at most. The approval bar for this round is the three other warnings.

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `22f7fad`

### Summary
The re-delivery trigger is now atomic: `hello` reads the previous capabilities, decides, and raises the desired version in one transaction, so a crash in the old window can no longer store a widened capability set with no bump. The badge and the notice name a remedy only when an update exists, the label map is keyed off the contract union, and the two over-wide page claims are corrected. _1 generated/lockfile file(s) not reviewed: packages/db/drizzle/meta/0045_snapshot.json._ _(Suppressed 1 finding per PR-local overrides: the withdrawn issue-link warning.)_ _7 suggestion(s) dropped under the 🟢 budget._

### Changes since last review
Previous HEAD: b8c1d0b (2026-09-10T07:53:52Z) — verdict COMMENT — unreachable, reviewed the whole PR

- ✅ **Fixed:** capability transition consumed outside its transaction (`packages/api-server/src/modules/runtime-delivery/services/hello-handler.ts:60`)
- ✅ **Fixed:** badge prescribes an update that may not exist (`packages/ui/src/modules/agents/components/unsupported-contributions-badge.tsx:15`)
- ✅ **Fixed:** invariant and trigger wider than the code (`docs/architecture/runtime-delivery.md:357`)
- ✅ **Fixed:** label map keyed on string, not ContributionKind (`packages/ui/src/modules/agents/lib/contribution-kind-labels.ts:1`)
- 🆕 **New:** docs and body drift: one stale claim, one stamp, one migration number (`docs/architecture/runtime-delivery.md:325`)
- 🆕 **New:** contract publishes string[], so the label map gains no exhaustiveness (`packages/api-server-api/src/modules/agents/types.ts:62`)
- 🆕 **New:** on a phone only the badge title carries the detail (`packages/ui/src/modules/sessions/views/chat-view.tsx:579`)

The atomic version is verified end to end: `upsertHello` takes the `FOR UPDATE` lock on the agents row inside the unit of work, `getRow` and `bumpVersion` accept the same `tx`, and the owner lookup, the event emit and the snapshot write stay outside it. The migration renumber from the rebase is clean — `0045` follows main's `0044`, the journal index matches, and the new snapshot carries both columns.

### Findings
- 🟢 **Suggestion:** three small drifts of one class, all in text: the capability-negotiation section still says the Agent shows the refused kinds "beside the pending update that fixes it", which this round's own fix makes untrue (`docs/architecture/runtime-delivery.md:325`); `docs/architecture.md` gains the landing-page entry but keeps its 2026-09-08 stamp (`docs/architecture.md:3`); and the body names migration `0044`, which after the rebase is an unrelated migration on main, while the branch adds `0045`. The body also does not mention that the harness-configuration prose moved onto its own page, which the commit explains and the body does not. Round 1 flagged two sentences of this class and not this one — the enumeration was mine to complete.
- 🟢 **Suggestion:** the refused kinds cross the contract as `string[]` (`packages/api-server-api/src/modules/agents/types.ts:62`, mirrored at `packages/ui/src/types.ts:135`) while the server already holds `ContributionKind[]`, so `contributionKindLabel` casts back with `as ContributionKind` and keeps a `?? kind` fallback. The map typing of this round therefore buys no build error for a seventh kind. All six kinds are mapped today, so nothing leaks now.
- 🟢 **Suggestion:** on a phone the notice sits in the sessions column, which is `hidden md:flex` on the chat screen, so only the badge `title` carries what was refused (`packages/ui/src/modules/sessions/views/chat-view.tsx:579`, `packages/ui/src/modules/agents/components/unsupported-contributions-badge.tsx:15`). This is the placement of the existing `RuntimeOutdatedNotice` and the `title` use of the existing `ContributionFailuresBadge`, so it is the repo's current pattern rather than something this PR introduced.

- 🟢 Both items are in `### Findings` above: the `:325` claim and the `docs/architecture.md` stamp. The stamps of `runtime-delivery.md`, `agent-lifecycle.md` and `connections.md` all read 2026-09-10 and match the commits' author dates.

### Issue Fit
- 🟢 Both items are in `### Findings` above: the undeclared page split, and the body's `0044`. The body now records the scope calls on the issue's three open questions, which closes the residual of the withdrawn round-1 finding.

- 🟢 The contract-widening pair is in `### Findings` above. Four more: `advertisedKindsChanged` takes `unknown` and rebuilds the shape by cast — a legacy shape yields an empty kind list, so the effect is one extra bump, never a missed one; `kindSetChanged` is typed `string[]` with a dead `typeof` guard; the new jsonb column is read through the blanket `as InternalRow[]` cast; `hello` reads the outbox row twice; and the new spec builds its port fakes with `as unknown as`. _(Dropped under the 🟢 budget.)_

- 🟢 The mobile and `title` items are in `### Findings` above. Three more: `ContributionGapNotice` repeats `RuntimeOutdatedNotice` almost verbatim, both surfaces render together on a desktop chat view, and `agent?.unsupportedContributionKinds ?? []` defends a required field. _(Dropped under the 🟢 budget.)_

### Verdict
APPROVE — every warning of the last round is closed at its cause, and what is left is three suggestions in text and typing.


---

jezekra1 (approved):

dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `becd4e0`

### Summary
The one new commit stops the two host-rail Contribution kinds from counting as a capability gap. The fix is correct and it was necessary: the runtime advertises only the kinds it has drivers for, `egress-allow` and `egress-inject` have no driver, so every agent with an egress-granting Connection carried a warning that no agent update could ever clear. What is left is text and typing.

### Changes since last review
Previous HEAD: 22f7fad (2026-09-10T09:03:00Z) — verdict APPROVE

- 🔁 **Still present:** docs and body drift: one stale claim, one stamp, one migration number (`docs/architecture/runtime-delivery.md:325`)
- 🔁 **Still present:** contract publishes string[], so the label map gains no exhaustiveness (`packages/api-server-api/src/modules/agents/types.ts:62`)
- 🔁 **Still present:** on a phone only the badge title carries the detail (`packages/ui/src/modules/sessions/views/chat-view.tsx:579`)
- 🆕 **New:** host-rail exemption not named in the page's invariant (`docs/architecture/runtime-delivery.md:355`)
- 🆕 **New:** rail membership is a set, so a new kind gets no compile error (`packages/api-server/src/modules/runtime-delivery/domain/capability-filter.ts:22`)

Verified in the code: the exempted kinds were already absent from `filteredContribs`, so the delivered payload and the hash do not change; only the recorded drop set gets smaller. No consumer of `droppedContributionKinds` loses data — the column is new in this PR, so no stored row carries the old value. The event loop is unaffected, because the runtime advertises every Event kind of its own enum.

### Findings
- 🟢 **Suggestion:** the page keeps three claims that this commit makes untrue for the two host-rail kinds. The invariant says a kind outside the advertised set is "never silently delivered" and that the Agent "carries an owner-visible warning naming what its runtime refused" (`docs/architecture/runtime-delivery.md:355`); the state table says the outbox row holds "the Contribution kinds the last delivery dropped" (`:337`); and the PR body offers "network access" as an example of what the notice names, which is the label of `egress-allow`. Naming the exemption and its reason — these kinds ride the Envoy rail, so the agent never sees them — closes all three. Reported by doc-drift and issue-fit as a warning; a page sentence the diff contradicts is a suggestion in this repo.
- 🟢 **Suggestion:** `HOST_RAIL_KINDS` is a membership set, so a seventh Contribution kind on the host rail compiles clean and silently returns to being reported as a gap (`packages/api-server/src/modules/runtime-delivery/domain/capability-filter.ts:22`). The same module already has the shape that fails to compile in that case, the exhaustive `switch (c.kind)` in `contribution-hash.ts`. The `egress-allow`/`egress-inject` pair is inlined at four other non-test sites, so the copy itself follows the tree's own pattern. Reported by typescript-engineering.

- 🟢 Both items are the docs finding in `### Findings` above: the invariant at `:355` and the state table at `:337`. The stamps are correct at this HEAD when read against author dates.

### Issue Fit
- 🟢 Both items are in `### Findings` above. The skill reads the exclusion as a correct fix inside the issue's declared scope, which matches the code: a kind that never rides the runtime channel is not a runtime gap.

- 🟢 Both items are the typing finding in `### Findings` above: the widened set gives no exhaustiveness, and the rail split now has a fourth copy with no single owner.

### Verdict
APPROVE — the round fixes a real false warning, it changes no delivered payload, and the residual is one docs item and one typing item.


---

PR #3651, opened 2026-09-10
feat(artifacts): start a new session from an artifact

### LINKED ISSUE #3555: UI - An artifact is a dead end when the user wants to keep working on it

PetrBulanek:
An artifact is where finished agent work lands, and today it is a dead end. A
user who wants the next version of an artifact must remember which agent made
it, open that agent, start a session, and describe the artifact from memory.
Most users do not, so the artifact stays as it is.

This adds one action, reachable wherever an artifact appears: start a new
session on the agent that published it, with a reference to the artifact
already in the composer. The user reads the pre-filled line, edits it, and
sends. Nothing runs before they send.

Uploaded artifacts have no publishing agent, so they do not offer the action.

Closes #3555

## Sub-issues

- [x] 01 — Start a new session from an artifact


dam-code-guardian (approved):
🛡️ **Code Guardian 🛡️** — ✅ Code Review @ `1d56ecd`

## PR #3651: feat(artifacts): start a new session from an artifact
**Author:** PetrBulanek | **Branch:** feat/3555-start-session-from-artifact → main | **Changes:** +114 −2 (8 files)

### Summary
The PR adds one action that opens a new session on the Agent that published an artifact, with a title-and-reference line pre-filled in the composer and nothing sent. It is reachable from the library row menu, the preview dialog and the docked panel, and it is hidden when the artifact has no publishing Agent.

### Findings
- ✅ **Looks good:** The order in `start()` holds against the navigation that follows it. `appendNewSessionDraft` writes `draftKey(agentId, null)` before `selectAgent`, `resetChatContext` clears the session but not the drafts map, and no path on the way into the chat clears that key — so the pre-filled line is the one the composer reads. The dock guard is also complete: `openArtifactDirty` mirrors `editor.dirty` and its effect clears the flag on unmount, so a stale flag cannot ask about an edit that is gone.

- 🟢 **Suggestion:** The new paragraph states the affordance wider than the code gates it. "From any surface that shows one" reads as unconditional, while the dialog gates on `version === head && !editor.editing` and the panel on `shownVersion === latest`, and the by-link share page has no action at all. Two more wordings in the same paragraph drift from this PR's own code: the internal reference is called "the name the artifact tools take", while the `get_artifact` description added here says to pass the `<id>` part; and "already in the composer" reads as a blank composer, while the store action appends after an existing unsent draft (`docs/architecture/artifact-library.md:34`).

### Issue Fit
- 🟢 **Suggestion:** The PR answers the three Open Questions of issue #3555 and the body does not say so. The issue proposes "a chat icon next to the existing share and overflow actions on the artifact row"; the row action is a labelled item inside the overflow menu, so it costs two clicks against the issue's "in one click" goal. The prefill is `<title> — platform://artifacts/<id>` without the "short line about iterating on it" the issue asks for, and the icon is `Launch` rather than the `Chat` icon the issue names. The `get_artifact` description change is also server-side work outside the issue's UI scope (`packages/ui/src/modules/artifacts/components/artifact-row-menu-items.tsx:50`).

_Reported 6 suggestions. All are dropped under the 🟢 budget; no finding of this section blocks. Two candidates were graded down after verification: the empty-draft path of `appendNewSessionDraft` is unreachable, because `artifactSessionPrefill` always returns the reference, and the reference format on the other id-taking tools is prose completeness, not a defect._

- 🟢 **Suggestion:** The head-version rule of the new action is re-derived at each call site — `version === head` in the dialog, `shownVersion === latest` in the panel, and no test in the row menu — while the hook is documented as the one way to continue an artifact. Passing the shown version into `useStartArtifactSession` keeps the three surfaces from drifting (`packages/ui/src/modules/artifacts/components/docked-artifact-panel.tsx:189`).

### Verdict
APPROVE — the mechanism, its guard and its state order hold at every surface, and no finding blocks.


---

kapetr (approved):

